PR #8369, opened 2017-05-03
Add no-default-browser-check to chrome runner.

brendandahl:


yurydelendik:
Thank you for the patch.

PR #8370, opened 2017-05-03
Replace unnecessary `bind(this)` and `var self = this` statements with arrow functions in remaining `src/` files

Snuffleupagus:
Easier reviewing with https://github.com/mozilla/pdf.js/pull/8370/files?w=1.


yurydelendik (approved):

yurydelendik:
Thank you for the patch.

PR #8375, opened 2017-05-04
Always reset the document properties dialog when it's closed (issue 8371)

### LINKED ISSUE #8371: Wrong Title in document properties
Configuration:
- PDF.js version: 1.7.225
- Is an extension: No

1. Open a PDF file with title specified
2. Open file properties window
3. Open a PDF file with no title specified
4. Open file properties window

When viewing document properties, the title should be -.

What went wrong?
The title is the first document's title. (When step 2 is omitted, then step 4 will behave as expected.)

Snuffleupagus:
As far as I can tell, the issue with the document properties dialog showing data for a previous PDF file opened in the default viewer, goes back all the way to PR #4149. The reason that no one caught this issue before is probably because the `Open file` button in the default viewer isn't used *that* heavily when testing.

To me, the solution to this problem is nice and simple. Since we're already updating every *available* property each time that the document properties dialog is opened, this patch simply ensures that we reset *all* of them when the dialog is closed.[1]

Fixes #8371.

---

[1] I suppose that the most *correct* solution would be to only update the document properties dialog *once* for a given PDF file, and then just ensure that it's reset when the PDF file is closed.
However, that would have meant a much larger refactoring effort, so in the interest of getting a long standing edge-case fixed I decided against that for now. Besides, since I can't imagine that users generally open the document properties *many* times for a given PDF file, I don't think it matters too much if it's updated on every open/close.

### DIFF
--- a/web/pdf_document_properties.js
+++ b/web/pdf_document_properties.js
@@ -17,6 +17,8 @@ import { getPDFFileNameFromURL, mozL10n } from './ui_utils';
 import { createPromiseCapability } from './pdfjs';
 import { OverlayManager } from './overlay_manager';
 
+const DEFAULT_FIELD_CONTENT = '-';
+
 /**
  * @typedef {Object} PDFDocumentPropertiesOptions
  * @property {string} overlayName - Name/identifier for the overlay.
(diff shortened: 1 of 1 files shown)


yurydelendik (commented):
I would move this code to the beginning of the open() method. The _dataAvailableCapability is not reset when document closed, so probably more simple solution will be to move _getProperties in the setDocumentAndUrl, cleanup when "clear document" is called (and remove dataAvailableCapability complexity)

yurydelendik:
I see this is maybe our optimization against getting more than-need-information. Probably we need capability after all.

Snuffleupagus:
Closing, since based on the review comments I've done a more extensive re-factoring of this code, which I'll submit in a new PR.

PR #8377, opened 2017-05-04
Replace unnecessary `bind(this)` statements with arrow functions in `web/` files

Snuffleupagus:
By using `let`, which is block-scoped, instead of `var` in a couple of places we're able to get rid of additional `bind` calls.

### DIFF
--- a/web/pdf_viewer.js
+++ b/web/pdf_viewer.js
@@ -383,8 +383,8 @@ var PDFViewer = (function pdfViewer() {
             return;
           }
           var getPagesLeft = pagesCount;
-          for (var pageNum = 1; pageNum <= pagesCount; ++pageNum) {
-            pdfDocument.getPage(pageNum).then(function(pageNum, pdfPage) {
+          for (let pageNum = 1; pageNum <= pagesCount; ++pageNum) {
+            pdfDocument.getPage(pageNum).then((pdfPage) => {
               var pageView = this._pages[pageNum - 1];
               if (!pageView.pdfPage) {
                 pageView.setPdfPage(pdfPage);
@@ -393,7 +393,7 @@ var PDFViewer = (function pdfViewer() {
               if (--getPagesLeft === 0) {
                 pagesCapability.resolve();
               }
-            }.bind(this, pageNum));
+            });
           }
         });
 
--- a/web/app.js
+++ b/web/app.js
@@ -833,10 +833,10 @@ var PDFViewerApplication = {
         }
         this.loadingBar.show();
 
-        this.disableAutoFetchLoadingBarTimeout = setTimeout(function () {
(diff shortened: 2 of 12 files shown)


web/pdf_viewer.js
```diff
@@ -383,8 +383,8 @@ var PDFViewer = (function pdfViewer() {
             return;
           }
           var getPagesLeft = pagesCount;
-          for (var pageNum = 1; pageNum <= pagesCount; ++pageNum) {
-            pdfDocument.getPage(pageNum).then(function(pageNum, pdfPage) {
+          for (let pageNum = 1; pageNum <= pagesCount; ++pageNum) {
```
yurydelendik: this was just converted into "var" for translated code, do we need to move let inside for-block?

yurydelendik: let make it `for(var i = 1; i <= pagesCount; i++) { let pageNum = i + 1;`

Snuffleupagus: Unless I'm missing something, this shouldn't be a problem. Note that Babel converts a loop with `let` into a loop over a *function*, thus creating the necessary closure; see e.g. [this simple example](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015&targets=&browsers=&builtIns=false&debug=false&experimental=true&loose=false&spec=false&code=var%20pageCount%20%3D%2010%3B%0Afor%20(let%20pageNum%20%3D%201%3B%20pageNum%20%3C%3D%20pagesCount%3B%20%2B%2BpageNum)%20%7B%0A%20Promise.resolve(()%20%3D%3E%20%7B%0A%20console.log(pageNum)%3B%0A%20%7D)%3B%0A%7D).

The complete code, taken from the preview above, thus becomes:
```javascript
onePageRenderedCapability.promise.then(function () {
  if (_pdfjs.PDFJS.disableAutoFetch) {
    pagesCapability.resolve();
    return;
  }
  var getPagesLeft = pagesCount;

  var _loop = function _loop(_pageNum) {
    pdfDocument.getPage(_pageNum).then(function (pdfPage) {
      var pageView = _this._pages[_pageNum - 1];
      if (!pageView.pdfPage) {
        pageView.setPdfPage(pdfPage);
      }
      _this.linkService.cachePageRef(_pageNum, pdfPage.ref);
      if (--getPagesLeft === 0) {
        pagesCapability.resolve();
      }
    });
  };

  for (var _pageNum = 1; _pageNum <= pagesCount; ++_pageNum) {
    _loop(_pageNum);
  }
});
```

yurydelendik: okay, nvm, I looked at wrong place

yurydelendik:
Thank you for the patch.

PR #8355, opened 2017-05-01
Replace unnecessary `bind(this)` and `var self = this` statements with arrow functions in `src/core/evaluator.js`

Snuffleupagus:
Note that by using `let` instead of `var` in `PartialEvaluator.setGState` and `TranslatedFont.loadType3Data`, we can get rid of further `bind` usages since `let` is block-scoped.
Also, the fact that `bind` wasn't used in the `Font` case inside of `setGState` is actually a bug which has been present ever since PR #5205, where a closure was replaced by a standard loop.[1]

---
[1] I'm not aware of any bugs caused by this, but that is probably more a happy accident than anything else, since e.g. just removing the `bind` from the `SMask` case without using block-scoped variables causes test failures.

*Edit:* Slightly smaller diff with https://github.com/mozilla/pdf.js/pull/8355/files?w=1.

PR #8381, opened 2017-05-05
Re-factor `PDFDocumentProperties` such that it's properly reset when a new PDF file is opened (issue 8371)

### LINKED ISSUE #8371: Wrong Title in document properties

Snuffleupagus:
This patch contains the following improvements:
 - Only fetch the various document properties *once* per PDF file opened, and cache the result (in a frozen object).
 - Always update the *entire* dialog at once, to prevent inconsistent UI state (issue 8371).
 - Ensure that the dialog, and all its internal properties, are reset when `PDFViewerApplication.close` is called.
 - Inline, and re-factor, the `getProperties` method in `open`, since that's the only call-site.
 - Always overwrite the fileSize with the value obtained from `pdfDocument.getDownloadInfo`, to ensure that it's correct.
 - ES6-ify the code that's touched in this patch.

Fixes #8371.

### DIFF
--- a/web/pdf_document_properties.js
+++ b/web/pdf_document_properties.js
@@ -13,9 +13,11 @@
  * limitations under the License.
  */
 
-import { getPDFFileNameFromURL, mozL10n } from './ui_utils';
+import { cloneObj, getPDFFileNameFromURL, mozL10n } from './ui_utils';
 import { createPromiseCapability } from './pdfjs';
 import { OverlayManager } from './overlay_manager';
+
+const DEFAULT_FIELD_CONTENT = '-';
 
 /**
  * @typedef {Object} PDFDocumentPropertiesOptions
@@ -29,21 +31,16 @@ class PDFDocumentProperties {
   /**
    * @param {PDFDocumentPropertiesOptions} options
    */
-  constructor(options) {
-    this.overlayName = options.overlayName;
-    this.fields = options.fields;
-    this.container = options.container;
+  constructor({ overlayName, fields, container, closeButton, }) {
+    this.overlayName = overlayName;
+    this.fields = fields;
+    this.container = container;
 
-    this.rawFileSize = 0;
-    this.url = null;
-    this.pdfDocument = null;
+    this._reset();
 
-    // Bind the event listener for the Close button.
-    if (options.closeButton) {
-      options.closeButton.addEventListener('click', this.close.bind(this));
+    if (closeButton) { // Bind the event listener for the Close button.
+      closeButton.addEventListener('click', this.close.bind(this));
     }
-    this._dataAvailableCapability = createPromiseCapability();
-
     OverlayManager.register(this.overlayName, this.container,
                             this.close.bind(this));
   }
@@ -52,9 +49,51 @@ class PDFDocumentProperties {
    * Open the document properties overlay.
    */
   open() {
+    let freezeFieldData = (data) => {
+      Object.defineProperty(this, 'fieldData', {
+        value: Object.freeze(data),
+        writable: false,
+        enumerable: true,
+        configurable: true,
+      });
+    };
+
     Promise.all([OverlayManager.open(this.overlayName),
                  this._dataAvailableCapability.promise]).then(() => {
-      this._getProperties();
+      // If the document properties were previously fetched (for this PDF file),
+      // just update the dialog immediately to avoid redundant lookups.
+      if (this.fieldData) {
+        this._updateUI();
+        return;
+      }
+      // Get the document properties.
+      this.pdfDocument.getMetadata().then(({ info, metadata, }) => {
+        freezeFieldData({
+          'fileName': getPDFFileNameFromURL(this.url),
+          'fileSize': this._parseFileSize(this.maybeFileSize),
+          'title': info.Title,
+          'author': info.Author,
+          'subject': info.Subject,
+          'keywords': info.Keywords,
+          'creationDate': this._parseDate(info.CreationDate),
+          'modificationDate': this._parseDate(info.ModDate),
+          'creator': info.Creator,
+          'producer': info.Producer,
+          'version': info.PDFFormatVersion,
+          'pageCount': this.pdfDocument.numPages,
+        });
+        this._updateUI();
+
+        // Get the correct fileSize, since it may not have been set (if
+        // `this.setFileSize` wasn't called) or may be incorrectly set.
+        return this.pdfDocument.getDownloadInfo();
+      }).then(({ length, }) => {
+        let data = cloneObj(this.fieldData);
+        data['fileSize'] = this._parseFileSize(length);
+
+        freezeFieldData(data);
+        this._updateUI();
+      });
     });
   }
 
@@ -66,19 +105,6 @@ class PDFDocumentProperties {
   }
 
   /**
-   * Set the file size of the PDF document. This method is used to
-   * update the file size in the document properties overlay once it
-   * is known so we do not have to wait until the entire file is loaded.
-   *
-   * @param {number} fileSize - The file size of the PDF document.
-   */
-  setFileSize(fileSize) {
-    if (fileSize > 0) {
-      this.rawFileSize = fileSize;
-    }
-  }
-
-  /**
    * Set a reference to the PDF document and the URL in order
    * to populate the overlay fields with the document properties.
    * Note that the overlay will contain no information if this method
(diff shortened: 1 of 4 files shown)


web/pdf_document_properties.js
```diff
@@ -29,32 +31,69 @@ class PDFDocumentProperties {
...
-      this._getProperties();
+      if (OverlayManager.active !== this.overlayName) {
+        return;
+      }
+      // Get the document properties.
+      this.pdfDocument.getMetadata().then(({ info, metadata, }) => {
+        if (OverlayManager.active !== this.overlayName) {
+          return;
+        }
+        let content = {
```
timvandermeij: I think this is a more general question and not completely related to this patch per se, but what is the advantage of using `let` here over regular `var`? As I understood it, `let` is useful for block scoping, but is that necessary here?

web/pdf_document_properties.js
```diff
@@ -87,71 +113,71 @@ class PDFDocumentProperties {
    * @param {Object} pdfDocument - A reference to the PDF document.
    * @param {string} url - The URL of the document.
    */
-  setDocumentAndUrl(pdfDocument, url) {
+  setDocument(pdfDocument, url) {
+    if (this.pdfDocument) {
```
timvandermeij: I think there is no real need to check for this here since `this._reset()` does the check too.

Moreover, shouldn't `this.pdfDocument` be private as well, i.e., `this._pdfDocument`, as it may only be set using this setter (like `this._url`)?

web/pdf_document_properties.js
```diff
@@ -87,71 +113,71 @@ class PDFDocumentProperties {
...
-      for (var identifier in content) {
-        this._updateUI(this.fields[identifier], content[identifier]);
+  /**
+   * @private
+   */
+  _reset() {
+    if (this.pdfDocument) {
+      // Reset the dialog when a new PDF file is opened (issue 8371).
+      for (let id in this.fields) {
+        this._updateField(this.fields[id]);
```
timvandermeij: I would expect this to be `this._clearField(this.fields[id])` or `this._updateField(this.fields[id], DEFAULT_FIELD_CONTENT)` instead because using the same method for both updating and clearing seems a bit counterintuitive given the name and the signature of the method. In this case, it's not entirely clear to me from reading this line that it's performing a reset.

Moreover, shouldn't `this.fields` be private as well?

timvandermeij (commented):
In general this patch looks good, but I've got some minor questions.

web/pdf_document_properties.js
```diff
@@ -29,32 +31,69 @@ class PDFDocumentProperties {
...
-
     OverlayManager.register(this.overlayName, this.container,
                             this.close.bind(this));
   }
 
   /**
    * Open the document properties overlay.
+   *
+   * Note that since the document properties are fetched asynchronously,
+   * the dialog will *only* be updated if it's still open when the data arrives.
```
Snuffleupagus: > If the dialog is closed (not visible), then it will still be updated in the background for the next time you open it.

That should be the opposite of what the code does, both currently and with this patch.
Since we always re-fetch the data, and update the dialog every time it's opened, there's no point to updating anything if the dialog has been closed by the time e.g. `this.pdfDocument.getMetadata()` is resolved. (Keep in mind that all the operations involved here are asynchronous.)

So basically, I was trying to provide a general comment for these checks:
```javascript
if (OverlayManager.active !== this.overlayName) {
  return;
}
```

> Open https://mozilla.github.io/pdf.js/web/viewer.html, wait two seconds and open the document properties and you'll immediately see the data.

Noting gets loaded until the dialog is successfully opened, so what you describe as "immediately" just *seem* instantaneous since two seconds after loading the `tracemonkey.pdf` file, there's probably not many heavy operations still occurring. So the asynchronous fetching of the various document properties thus becomes very fast.

> Or is that not what you meant here?

No, that's not really what I meant :-)
So any suggestions on making this easier to understand would be helpful!
(I could also move, or even just remove, the comment if doesn't help make sense of the code.)

web/pdf_document_properties.js
Snuffleupagus: > As I understood it, let is useful for block scoping, but is that necessary here?

It's not technically necessary *here*, but you could also flip the question on its head and ask: "Why would you (generally) use `var` when `let` now is a thing?" :-)
Basically, using `let` can actually help avoid some nasty bugs[1] related to e.g. hoisting of `var`.[2]

There's a lot of useful information in this MDN article: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let.

---
[1] See e.g. https://github.com/mozilla/pdf.js/pull/8355#issue-225492050, which shows how `var` can easily do something other than what you'd expect/intend.
[2] There's obviously cases where you want `var` specifically, but in most cases it doesn't matter, and then I think I'd favour `let` in new code. (The general Firefox frontend code-base also seem to be trending in this general direction, unless I'm not mistaken.)

web/pdf_document_properties.js
Snuffleupagus: > I think there is no real need to check for this here since this._reset() does the check too.

There's a very good reason, and it's that usually `setFileSize` is called before `setDocument`. If we'd just unconditionally call `reset` here, then that would complete remove the point of `setFileSize`.
Maybe some sort of comment might be in order here!?

> Moreover, shouldn't this.pdfDocument be private as well, i.e., this._pdfDocument, as it may only be set using this setter (like this._url)?

Hmm, I'm guessing that I probably didn't change that since `this.pdfDocument` is used in *a lot* of `web/` files. Actually, I'm wondering if I shouldn't just revert the naming in this patch to its prior state, since I'm not sure if marking these as "private" properties really adds anything?

web/pdf_document_properties.js
Snuffleupagus: > `this._updateField(this.fields[id], DEFAULT_FIELD_CONTENT)`

That'd be completely redundant, since `_updateField` will always fallback to `DEFAULT_FIELD_CONTENT` when no (valid) content is provided (there's a comment to that effect in the method).
Having *two* different methods for this just seem unnecessary to me; would a better comment help here?

web/pdf_document_properties.js
timvandermeij: Good point! I understand what you mean now. The comment itself looks clear after reading it again with this in mind, but I would indeed suggest to move it above the first `OverlayManager` check as the comment describes why we need such a check. Other than moving it, the wording seems fine to me.

web/pdf_document_properties.js
timvandermeij: I agree. That article is indeed really helpful and I see why `let` can be preferred over `var` now, even though it's not strictly necessary in this case. All is fine here then, thanks!

Snuffleupagus:
@timvandermeij I've just pushed a new version of the patch, which re-works quite a bit more of this code.

Please take a look at the updated https://github.com/mozilla/pdf.js/pull/8381#issue-226592775, where the first *two* points are a significant change from the first version of the patch (and the current `master`).

I've tried to take your questions/comments into account when writing, and commenting, the new code.
Unfortunately this largely invalidates your previous reviewing, for which I apologise, but I hope that this new approach makes sense!

web/pdf_document_properties.js
timvandermeij: The checks are fine then, but a small comment could indeed be helpful here to avoid any confusion.

I'm fine with reverting the naming changes. It will make the diff smaller and for now it does not add much. We can change all of them at a later time if we want, but then we should indeed do all of them instead of just a few.

web/pdf_document_properties.js
timvandermeij: You're right, but to me it just looked strange to update a field without providing a value. The fallback is implicit if you just read that line without going to the actual `updateField` method. Adding a comment above this line saying that the fallback value is used will indeed help to avoid confusion when reading this file.

timvandermeij:
This new version looks really good and appears to resolve my comments. I'll do a full review later today.

Snuffleupagus:
I've tweaked a couple of comments slightly, to fix a couple of typos and missing words.

/botio-linux preview

Snuffleupagus:
@timvandermeij Sorry about the back and forth in this patch, but I realized that there were a number of opportunities to clean-up the code I've written.
I think that I'm now finally happy with the patch, since I've been able to simplify the code quite a bit compared to prior versions. Notably, the diff size for `pdf_document_properties.js` is almost cut in half.

timvandermeij (approved):

timvandermeij:
Nice improvement; thank you for fixing and simplifying this!

PR #8382, opened 2017-05-05
Convert part of the files in the `src/core/` folder to ES6 modules

Snuffleupagus:
This patch converts only `worker.js`, and the files it directly depends on, to ES6 modules.

*Edit:* Much smaller, and more readable, diff with https://github.com/mozilla/pdf.js/pull/8382/files?w=1.

### DIFF
--- a/src/core/network.js
+++ b/src/core/network.js
@@ -13,603 +13,589 @@
  * limitations under the License.
  */
 
-'use strict';
+import {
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
At least the unit-tests passed now, but Chrome (on Linux) still timed out during the reference tests.
Closing for now, since even this smaller PR unfortunately doesn't seem to work :-(

PR #8361, opened 2017-05-02
Export LoopbackPort

tbtlr:
As discussed on IRC, necessary changes to load PDF.js in a .jsm module and initializing a fake worker port externally.

PR #8385, opened 2017-05-06
Fix typo in LoopbackPort export

tbtlr:
There was a typo in #8361, sorry for that.


Snuffleupagus:
(There's no point in running tests, since the `src/pdf.js` file isn't used in testing.)

/botio-linux preview

Snuffleupagus:
Thanks for the patch!

PR #8364, opened 2017-05-02
Adds babel caching for system.js.

yurydelendik:
Makes developer viewer loading faster.

PR #8387, opened 2017-05-06
amends Babel cache (#8364) implementation to also work on Safari

### LINKED ISSUE #8386: Master @ 50d026f Broken? Viewer no longer loads on Safari but works on Firefox
```
2928578164fc076ec46d74c6de6e43187fa38105 — Works
50d026fbda4858e46b13914360dfa0bc2ddf3ddb — Does Not Work
bbb2cc000eecd7dbe43128152bcff769dfccf120 — First Bad Commit (according to `git bisect`)
```

Configuration:
- Web browser and its version: Safari 10.1 (12603.1.30.0.34) NG, Firefox 53.0.2 (64-bit) OK
- Operating system and its version: macOS 10.12.4 (16E195)
- PDF.js version: master @ 50d026f
- Is an extension: No

1. Bump PDF.js repo to latest `master` commit 50d026f.
2. Run `gulp server`
3. Check one of the PDFs in the embedded viewer

What is the expected behaviour?
Viewer should load and render document in all major browsers

What went wrong?
Viewer loads, but if using Safari, it does not render document

I’ve also tried:
- `npm prune` then `npm install`. No luck
- Cloning the entire repo again from GitHub, then `npm install`. No luck: same issue in Safari

evadne:
Fixes a problem with Safari having `crypto.webkitSubtle` instead of `crypto.subtle`, which can cause digest generation for cached files to fail, and the viewer eventually not load in development mode.

Fixes #8386

### DIFF
--- a/systemjs.config.js
+++ b/systemjs.config.js
@@ -36,7 +36,8 @@
 
   var isCachingPossible = typeof indexedDB !== 'undefined' &&
                           typeof TextEncoder !== 'undefined' &&
-                          typeof crypto !== 'undefined';
+                          typeof crypto !== 'undefined' &&
+                          typeof crypto.subtle !== 'undefined';
 
   SystemJS.config({
     packages: {


Snuffleupagus:
I'll leave the decision to @yurydelendik, but given https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle#Browser_compatibility, I'm wondering if it would be better to change [`isCachingPossible`](https://github.com/mozilla/pdf.js/blob/master/systemjs.config.js#L37-L39) to something like this:
```javascript
var isCachingPossible = typeof indexedDB !== 'undefined' &&
                        typeof TextEncoder !== 'undefined' &&
                        typeof crypto !== 'undefined' &&
                        typeof crypto.subtle !== 'undefined';
```

evadne:
Looks like Safari and IE are both prefixed, I’ll amend.

evadne:
I’ve added a couple fixes since then: caching can still be enabled for Safari, as long as it is not turned on for the worker. Did not look at getting IE to behave at the moment, though.

Snuffleupagus:
Personally I think the latest version of this PR is *unnecessarily* complicated, keeping in mind that the code should be maintainable, since this is basically trying to hack around one (or more) browser specific bugs.
Hence I'd like to re-iterate the simple suggestion from https://github.com/mozilla/pdf.js/pull/8387#issuecomment-299643196, that simply disables caching instead.

Also, please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

yurydelendik:
I agree with disabling caching for Safari. The developer version of demo viewer is not intended to be used on all supported browsers, but only on the limited set of browsers supporting new EcmaScript syntax and established set of HTML5 apis.

Deploying viewers in production environments  without first building it is not/will not be supported.

timvandermeij:
I agree here. I think the solution in https://github.com/mozilla/pdf.js/pull/8387#issuecomment-299643196 is the only acceptable solution here.

evadne:
Right but I actually develop using Safari 😬 added one more commit incorporating your feedback, thank you!

NB: I did find in my testing that caching works for Safari as long as it is not enabled in worker. It would seem that with caching enabled, the base/worker communication ceases to work. Now with caching disabled for Safari entirely the problem is not triggered but I would foresee it coming back again in the near future once WebKit drops prefix.

systemjs.config.js
```diff
@@ -34,9 +34,10 @@
     'node_modules/systemjs-plugin-babel/systemjs-babel-browser.js';
   var PluginBabelCachePath = 'external/systemjs/plugin-babel-cached.js';
 
-  var isCachingPossible = typeof indexedDB !== 'undefined' &&
-                          typeof TextEncoder !== 'undefined' &&
-                          typeof crypto !== 'undefined';
+  var isCachingPossible = (typeof indexedDB !== 'undefined') &&
+                          (typeof TextEncoder !== 'undefined') &&
+                          (typeof crypto !== 'undefined') &&
+                          (typeof crypto.subtle !== 'undefined');
```
Snuffleupagus: Nit: There's really no need to add parenthesis on every line here, it's quite readable anyway.
(Also, in this case, it unnecessarily changes the blame for unrelated lines.)

evadne: Nixed your nit, hopefully

yurydelendik:
> Right but I actually develop using Safari 😬 added one more commit incorporating your feedback, thank you!

Thank you for addressing the feedback.

> It would seem that with caching enabled, the base/worker communication ceases to work. ... I would foresee it coming back again in the near future once WebKit drops prefix.

If WebKit drops prefix, I hope they will support full [Crypto interface](https://www.w3.org/TR/WebCryptoAPI/#crypto-interface) , which must be supported in Window and Worker objects. Could you explain concern/problem with worker side and why it will not work?

yurydelendik (approved):
Thank you for the patch.

evadne:
Thanks for merging. I did not get to do root cause analysis on the problem with worker + caching but will keep this in mind should situation avail itself.

PR #8388, opened 2017-05-07
Cache JPEG images, just as we do for other image formats, in `evaluator.js` (issue 8380)

### LINKED ISSUE #8380: Horrible performance with a pdf document
https://www.mbank.pl/download/firma/Dyspozycja-zmiany-typu-rachunku-biecego.pdf?noredir -- this pdf takes over 25 seconds to display.

Configuration:
- Web browser and its version: Nightly 55.0a1 (2017-05-05) (32-bit)
- Operating system and its version: Windows 7 SP1
- PDF.js version: 1.8.320
- Is an extension: yes

Snuffleupagus:
For some reason, we're putting all kind of images *except* JPEG into the `imageCache` in `evaluator.js`.[1]
This means that in the PDF file in issue #8380, we'll keep sending the *same* two small images[2] to the main-thread and decoding them over and over. This is obviously hugely inefficient!

As can be seen from the discussion in the issue, the performance becomes *extremely* bad if the user has the addon "Adblock Plus" installed. However, even in a clean Firefox profile, the performance isn't that great.

This patch not only addresses the performance implications of the "Adblock Plus" addon together with that particular PDF file, but it *also* improves the rendering times considerably for *all* users.
Locally, with a clean profile, the rendering times are reduced from `~2000 ms` to `~500 ms` for my setup!

Obviously, the general structure of the PDF file and its operator sequence is still hugely inefficient, however I'd say that the performance with this patch is good enough to consider the issue (as it stands) resolved.[3]

Fixes #8380.

---
[1] Not technically true, since inline images are cached from `parser.js`, but whatever :-)

[2] The two JPEG images have dimensions 1x2, respectively 4x2.

[3] To make this even more efficient, a new state would have to be added to the `QueueOptimizer`. Given that PDF files this stupid fortunately aren't too common, I'm not convinced that it's worth doing.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -473,9 +473,14 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           NativeImageDecoder.isSupported(image, this.xref, resources)) {
         // These JPEGs don't need any more processing so we can just send it.
         operatorList.addOp(OPS.paintJpegXObject, args);
-        this.handler.send('obj',
-          [objId, this.pageIndex, 'JpegStream',
-           image.getIR(this.options.forceDataSchema)]);
+        this.handler.send('obj', [objId, this.pageIndex, 'JpegStream',
+                                  image.getIR(this.options.forceDataSchema)]);
+        if (cacheKey) {
+          imageCache[cacheKey] = {
+            fn: OPS.paintJpegXObject,
+            args,
+          };
+        }
         return;
       }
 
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
One thing that I forgot to mention in https://github.com/mozilla/pdf.js/pull/8388#issue-226850147, is that if you *disable* the `NativeImageDecoder` then all images format, including JPEG, are already cached as `OPS.paintImageXObject` with the current `master`.

What this patch does is simply to ensure that we *also* cache JPEG images, as `OPS.paintJpegXObject`, when the `NativeImageDecoder` is *enabled* (which is the default).

yurydelendik:
Thank you for the patch.

Snuffleupagus:
Sorry; I've just restored the branch, let's try again.

/botio-linux makeref

PR #8391, opened 2017-05-07
Convert the hand tool, interfaces and secondary toolbar to ES6 syntax

timvandermeij:
*Easier reviewing with https://github.com/mozilla/pdf.js/pull/8391/files?w=1.*

### DIFF
--- a/web/hand_tool.js
+++ b/web/hand_tool.js
@@ -20,20 +20,17 @@ import { localized } from './ui_utils';
  * @typedef {Object} HandToolOptions
  * @property {HTMLDivElement} container - The document container.
  * @property {EventBus} eventBus - The application event bus.
+ * @property {BasePreferences} preferences - Object for reading/writing
+ *                                           persistent settings.
  */
 
-/**
- * @class
- */
-var HandTool = (function HandToolClosure() {
+class HandTool {
   /**
-   * @constructs HandTool
    * @param {HandToolOptions} options
    */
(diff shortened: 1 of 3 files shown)


web/hand_tool.js
```diff
@@ -46,12 +43,12 @@ var HandTool = (function HandToolClosure() {
 
     this.eventBus.on('togglehandtool', this.toggle.bind(this));
 
-    Promise.all([localized,
-                 preferences.get('enableHandToolOnLoad')]).then((values) => {
+    let enableOnLoad = preferences.get('enableHandToolOnLoad');
+    Promise.all([localized, enableOnLoad]).then((values) => {
```
Snuffleupagus: Note that you can actually use array destructuring here (and I'm not sure that the temporary `enableOnLoad` is necessary for readability); how about the following instead?
```javascript
Promise.all([
  localized,
  preferences.get('enableHandToolOnLoad')
]).then(([, value]) => {
  if (value === true) {
    this.handTool.activate();
  }
}).catch(function(reason) {});
```

Snuffleupagus (approved):
I've got one small suggestion, but apart from that this looks really good; thank you!

web/hand_tool.js
yurydelendik: I had trouble with `([, value]) =>` with construct. Let's keep it as is -- I have something in mind to remove this statement later.

PR #8393, opened 2017-05-09
Replace unnecessary `var self = this` statements with arrow functions in `web/app.js`

Snuffleupagus:
Also replaces `var` with `let` in the functions/methods that are touched in the patch. Please note that this should be completely safe, for two separate reasons, since trying to access `let` in a scope where it's not defined is first of all a runtime error and second of all an ESLint error (thanks to the `no-undef` rule).

*Edit:* Slightly more manageable diff with https://github.com/mozilla/pdf.js/pull/8393/files?w=1.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -258,131 +258,131 @@ var PDFViewerApplication = {
    * @private
    */
   _initializeViewerComponents() {
-    var self = this;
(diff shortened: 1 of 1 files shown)


yurydelendik:
Thank you for the patch.

PR #8394, opened 2017-05-09
Wraps mozL10n to async calls; splits firefox and generic l10n libs.

yurydelendik:


yurydelendik:
Needed for #8402 too

web/app.js
```diff
@@ -1582,10 +1618,12 @@ function webViewerTextLayerRendered(e) {
   if (typeof PDFJSDev !== 'undefined' &&
       PDFJSDev.test('FIREFOX || MOZCENTRAL') &&
       e.numTextDivs > 0 && !PDFViewerApplication.supportsDocumentColors) {
-    console.error(mozL10n.get('document_colors_not_allowed', null,
+    PDFViewerApplication.l10n.get('document_colors_not_allowed', null,
       'PDF documents are not allowed to use their own colors: ' +
       '\'Allow pages to choose their own colors\' ' +
-      'is deactivated in the browser.'));
+      'is deactivated in the browser.').then((msg) => {
```
Snuffleupagus: Nit: Let's indent the above three lines with two more spaces, to improve readability here.

web/app.js
```diff
@@ -254,6 +258,24 @@ var PDFViewerApplication = {
     ]).catch(function (reason) { });
   },
 
+  _initializeL10n() {
+    if ((typeof PDFJSDev === 'undefined' ||
+        !PDFJSDev.test('FIREFOX || MOZCENTRAL')) &&
+        PDFViewerApplication.viewerPrefs['pdfBugEnabled']) {
```
Snuffleupagus: The above condition isn't exactly equal to the current ones, since in `master` it's possible to change the `#locale` in dev-mode (i.e. when using `gulp server`) without having the 'pdfBugEnabled' pref set.
I think it'd be nice if we could keep that behaviour, so this needs to look something like:
```javascript
if (((typeof PDFJSDev === 'undefined' || !PDFJSDev.test('PRODUCTION')) ||
     this.viewerPrefs['pdfBugEnabled']) &&
    (typeof PDFJSDev === 'undefined' ||
     !PDFJSDev.test('FIREFOX || MOZCENTRAL'))) {
```

Obviously that isn't as readable as one would like, but it may be possible to simplify the condition somewhat.

web/app.js
```diff
@@ -1446,15 +1476,15 @@ function webViewerInitialized() {
...
-      PDFJS.disableFontFace = true;
-      console.warn(mozL10n.get('web_fonts_disabled', null,
-        'Web fonts are disabled: unable to use embedded PDF fonts.'));
-    }
+  if (typeof PDFJSDev !== 'undefined' &&
+      PDFJSDev.test('FIREFOX || MOZCENTRAL') &&
+      !PDFViewerApplication.supportsDocumentFonts) {
+    PDFJS.disableFontFace = true;
+    PDFViewerApplication.l10n.get('web_fonts_disabled', null,
+        'Web fonts are disabled: unable to use embedded PDF fonts.').
```
Snuffleupagus: Nit: Let's indent this line with two more spaces, to improve readability here.

web/password_prompt.js
```diff
@@ -62,15 +64,18 @@ class PasswordPrompt {
...
+        promptString = this.l10n.get('password_invalid', null,
           'Invalid password. Please try again.');
+      } else {
+        promptString = this.l10n.get('password_label', null,
+          'Enter the password to open this PDF file.');
       }
 
-      this.label.textContent = promptString;
+      promptString.then((msg) => {
+        this.label.textContent = promptString;
```
Snuffleupagus: Needs to be `this.label.textContent = msg;`.

Snuffleupagus (commented):
Generally this looks really good, thank you!

However, given the size of the PR and that there's a couple of parameter/naming issues which cause UI glitches, I'd like to look at/test this again once the comments are addressed.

yurydelendik:
@Snuffleupagus thanks for the review

/botio-windows preview

web/pdf_viewer.js
```diff
@@ -899,16 +902,19 @@ var PDFViewer = (function pdfViewer() {
      * @param {HTMLDivElement} pageDiv
      * @param {PDFPage} pdfPage
      * @param {boolean} renderInteractiveForms
+     * @param {L10n} l10n
```
Snuffleupagus: Shouldn't the type be `{IL10n}` here, since that's used everywhere else?

web/annotation_layer_builder.js
```diff
@@ -96,15 +96,18 @@ class DefaultAnnotationLayerFactory {
    * @param {HTMLDivElement} pageDiv
    * @param {PDFPage} pdfPage
    * @param {boolean} renderInteractiveForms
+   * @param {L10n} l10n
```
Snuffleupagus: Shouldn't the type be `{IL10n}` here, since that's used everywhere else?

web/app.js
```diff
@@ -254,6 +258,25 @@ var PDFViewerApplication = {
     ]).catch(function (reason) { });
   },
 
+  _initializeL10n() {
+    // Locale can be changed only when special debugging flags is present in
+    // the hash section of the URL, or development version of viewer is used.
+    // It is not possible to change locale for Firefox extension builds.
+    if (typeof PDFJSDev === 'undefined' || !PDFJSDev.test('PRODUCTION') ||
+        (!PDFJSDev.test('FIREFOX || MOZCENTRAL') &&
+         PDFViewerApplication.viewerPrefs['pdfBugEnabled'])) {
```
Snuffleupagus: This can be `this.viewerPrefs['pdfBugEnabled']`, given the current scope here (and you're already using `this.externalServices` below).

web/pdf_thumbnail_view.js
```diff
@@ -30,6 +30,7 @@ var THUMBNAIL_CANVAS_BORDER_WIDTH = 1; // px
  * @property {boolean} disableCanvasToImageConversion - (optional) Don't convert
  *   the canvas thumbnails to images. This prevents `toDataURL` calls,
  *   but increases the overall memory usage. The default value is false.
+ * @property {L10n} l10n - Localization service.
```
Snuffleupagus: Shouldn't the type be `{IL10n}` here, since that's used everywhere else?

web/pdf_thumbnail_view.js
```diff
@@ -64,6 +65,7 @@ var PDFThumbnailView = (function PDFThumbnailViewClosure() {
   /**
    * @constructs PDFThumbnailView
    * @param {PDFThumbnailViewOptions} options
+   * @param {IL10n} l10n
```
Snuffleupagus: Please remove this line, since it's passed in with the options now.

web/toolbar.js
```diff
@@ -170,7 +174,12 @@ var Toolbar = (function ToolbarClosure() {
     },
 
     _updateUIState: function Toolbar_updateUIState(resetNumPages) {
-      function selectScaleOption(value, scale) {
+      if (!this._wasLocalized) {
+        // Don't update UI state until we will localize the toolbar.
+        return;
+      }
+
+      let selectScaleOption = (value, scale, l10n) => {
```
Snuffleupagus: Please remove the `l10n` parameter, since it's now unused.

Snuffleupagus (approved):
I've left a couple of final comments; and please also add `createL10n` to `DefaultExternalServices` in https://github.com/mozilla/pdf.js/blob/master/web/app.js#L67.

Please note that when rebasing this, the changes to the `hand_tool.js` and `secondary_toolbar.js` files are no longer needed!

This looks really good, thanks for working on this!

external/webL10n/l10n.js
```diff
@@ -998,7 +998,6 @@ document.webL10n = (function(window, document, undefined) {
       loadLocale(lang, function() {
         if (callback)
           callback();
-        translateFragment();
```
timvandermeij: Mention this change at https://github.com/yurydelendik/pdf.js/blob/6f76510a74f086de4fdb9a21a677649de9618c56/external/webL10n/l10n.js#L24 as well. If we update `webL10n` in the future, this change may be accidentally reverted if it's not mentioned there.

PR #8284, opened 2017-04-13
Convert the files in the `/src/core` folder to ES6 modules

Snuffleupagus:
*Depends on PR #8364.*

Please note that the `glyphlist.js` and `unicode.js` files are converted to CommonJS modules instead, since Babel cannot handle files that large and they are thus excluded from transpilation.

Much smaller diff with https://github.com/mozilla/pdf.js/pull/8284/files?w=1.

PR #8396, opened 2017-05-10
Adds streams-lib polyfill and exports ReadableStream from shared/util.

mukulmishra18:
This is the first PR of project **Streams API in PDF.js**. This allow PDF.js to use Streams API in the process of networking and rendering. For that purpose, this PR:

- Fork Streams API [ref-implementation](https://github.com/whatwg/streams/tree/master/reference-implementation) and convert it to commonjs module.

- Import/Export ReadableStream from shared/util.js.

- Adds test for ReadableStream to check it's working.

Increase in file size(in kb) after adding this code, in pdf.js and pdf.worker.js:

- pdf.js: 468.8(after) - 343.7(before) = ~125

- pdf.worker.js: 1400(after) - 1300(before) = 100

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1012,8 +1012,7 @@ gulp.task('lib', ['buildnumber'], function () {
   };
   var licenseHeader = fs.readFileSync('./src/license_header.js').toString();
   var preprocessor2 = require('./external/builder/preprocessor2.js');
-
-  return merge([
+  var buildLib = merge([
     gulp.src([
       'src/{core,display}/*.js',
       'src/shared/{compatibility,util}.js',
@@ -1027,6 +1026,11 @@ gulp.task('lib', ['buildnumber'], function () {
     gulp.src('test/unit/*.js', {base: '.'}),
   ]).pipe(transform(preprocess))
     .pipe(gulp.dest('build/lib/'));
+  return merge([
+    buildLib,
+    gulp.src('external/streams/streams-lib.js', {base: '.'})
+      .pipe(gulp.dest('build/')),
+  ]);
 });
 
 gulp.task('web-pre', ['generic', 'extension', 'jsdoc']);
@@ -1320,6 +1324,8 @@ gulp.task('dist-repo-prepare', ['dist-pre'], function () {
     createStringSource('bower.json', JSON.stringify(bowerManifest, null, 2));
 
   return merge([
+    gulp.src('external/streams/streams-lib.js', {base: '.'})
+      .pipe(gulp.dest('build/dist/')),
     packageJsonSrc.pipe(gulp.dest(DIST_DIR)),
     bowerJsonSrc.pipe(gulp.dest(DIST_DIR)),
     vinyl.src('external/dist/**/*',
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,6 +8,7 @@ external/webL10n/
 external/cmapscompress/
 external/builder/fixtures/
 external/builder/fixtures_esprima/
+external/streams/
 src/shared/cffStandardStrings.js
 src/shared/fonts_utils.js
 test/tmp/
--- a/external/streams/LICENSE.md
+++ b/external/streams/LICENSE.md
@@ -1,0 +1,21 @@
+# WHATWG Streams Reference Implementation Licensing
+
+For the reference implementation and tests for the WHATWG Streams Standard.
+
+This code is dual-licensed under CC0 and the MIT license. You can choose which one you want to use.
+
+## CC0
+
+To the extent possible under law, the authors have dedicated all copyright and related and neighboring rights to this software to the public domain worldwide. This software is distributed without any warranty.
+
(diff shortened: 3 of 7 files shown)


Snuffleupagus (commented):
First of all, please note that Travis has *a lot* of complaints about the code. In this case, since it concerns external code, I'd suggest simply adding ` external/streams/` to https://github.com/mozilla/pdf.js/blob/master/.eslintignore. ***Edit:*** Remember that running `gulp lint` helps catch these kind of issues locally.

Second of all, this will get bundled into *all* build targets by default. I imagine that we don't want to do that, at least not at the present time, since it would cause e.g. the `FIREFOX`/`MOZCENTRAL` builds to become quite a lot larger (for code that we currently don't use in production).

mukulmishra18:
> I'd suggest simply adding external/streams/ to https://github.com/mozilla/pdf.js/blob/master/.eslintignore

Thanks for you suggestion. I will add the file.

> this will get bundled into all build targets by default.

No idea what we have to do for this. Maybe we can compress the polyfill. Have to ask @yurydelendik .

yurydelendik:
> the FIREFOX/MOZCENTRAL builds to become quite a lot larger 

We need to follow https://bugzilla.mozilla.org/show_bug.cgi?id=1272697 for that. I'm expecting landing of streamable getTextContent as soon as we land this PR, so the code will be using streams api in production.

timvandermeij:
It looks like Travis does not see the library as a module. Is it a valid CommonJS module?

Moreover, the file looks built. I suspect one of the next steps will be to fetch the original from the repository and build it ourselves? In case the upstream library changes, it should be easy for us to rebuild our version from source.

yurydelendik:
TODO:
- [ ] replace `const assert = require('assert');` -- it adds some useless for us node.js stubs
- [ ] on `gulp lib` and `gulp dist` copy "external/streams/" to the build/external/streams/ and build/dist/external/streams/ folder
- [ ] use LICENSE.md from reference implementation in the external/streams/

gulpfile.js
```diff
@@ -1014,19 +1014,26 @@ gulp.task('lib', ['buildnumber'], function () {
...
 
   return merge([
+    gulp.src('external/streams/streams-lib.js', {base: '.'})
+      .pipe(gulp.dest('build/')),
     gulp.src([
       'src/{core,display}/*.js',
       'src/shared/{compatibility,util}.js',
       'src/{pdf,pdf.worker}.js',
-    ], {base: 'src/'}),
+    ], {base: 'src/'})
```
yurydelendik: too many irrelevant changes. let's make original marge as `var buildLib = merge([...]);` and then add
```
return merge([
   buildLib,
   gulp.src('external/streams/streams-lib.js', {base: '.'})
     .pipe(gulp.dest('build/'))
  ]);
```

gulpfile.js
```diff
@@ -1372,7 +1379,10 @@ gulp.task('dist-repo-git', ['dist-repo-prepare'], function () {
   console.log();
 });
 
-gulp.task('dist', ['dist-repo-git']);
+gulp.task('dist', ['dist-repo-git'], function () {
+  gulp.src('external/streams/streams-lib.js', {base: '.'})
+    .pipe(gulp.dest('build/dist/'));
```
yurydelendik: this belong in the dist-repo-prepare task

gulpfile.js
mukulmishra18: Okay, sure. Looks more clear. Thanks for suggestion.

Snuffleupagus:
While I'd wish that the size impact of this patch was a bit smaller, I've got no formal objections here :-)
So, as far as I'm concerned, we can land this PR (with passing tests of course)!

/botio test

Snuffleupagus:
It unfortunately seems that loading all of this new code has made the tests even more unreliable :-(
Even the unit-tests now timeout frequently, whereas there are fine when run against `master`.

mukulmishra18:
> It unfortunately seems that loading all of this new code has made the tests even more unreliable :-(
> Even the unit-tests now timeout frequently, whereas they're fine when run against master.

Not sure why this is happening(as all unit-tests are passing on my local machine) and what will be the potential solution of this :-(

But i think we can reduce the size of `streams-lib` file by removing all except ReadableStream(as we only need this for now) from [streams.lib.entry.js](https://github.com/mukulmishra18/streams-lib/blob/master/streams.lib.entry.js) file, as we are using this file to bundle streams-lib.

Snuffleupagus:
> Not sure why this is happening(as all unit-tests are passing on my local machine) and what will be the potential solution of this :-(

This is a pre-existing issue affecting *only* the Chrome browser on the Linux bot, and one that has thus far prevented landing for example PR #8284.
So while it's not the fault of this patch itself, it will probably still make landing this difficult at the moment (until a solution/workaround for the timeouts have been found).

Note that e.g. compared to PR #8284, the current PR seems to cause even worse timeouts on the Linux bot.

PR #8397, opened 2017-05-10
Simplify `extractText`, and replace unnecessary `var self = this` statements with arrow functions, in `web/pdf_find_controller.js`

Snuffleupagus:
- Simplify `extractText` in `web/pdf_find_controller.js`
   Currently this method first uses a loop to build a temporary array to hold Promises, which are then resolved from a recursive helper function once the textContent is fetched for each page.

   To me, this is unncessarily complicated, since we can do everything within one loop by simply chaining the asynchronous calls to retrieve the textContent. (Note that this guarantees that the textContent of the pages is still fetched sequentially.)

 - Replace unnecessary `var self = this` statements with arrow functions in `web/pdf_find_controller.js`

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -13,6 +13,7 @@
  * limitations under the License.
  */
 
+import { createPromiseCapability } from './pdfjs';
 import { scrollIntoView } from './ui_utils';
 
 var FindStates = {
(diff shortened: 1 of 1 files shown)


yurydelendik (approved):
Looks good

PR #8398, opened 2017-05-11
Convert the `DOMCMapReaderFactory` and `DOMCanvasFactory` to ES6 classes

Snuffleupagus:
- Convert the `DOMCMapReaderFactory` to an ES6 class

   Given that we only create *one* instance of this class per `getDocument` call, this shouldn't matter performance wise.

 - Convert the `DOMCanvasFactory` to an ES6 class

   For consistency, also updates the `pdf2png.js` example to use the slightly less verbose `canvasAndContext` parameter name.

Smaller diff with https://github.com/mozilla/pdf.js/pull/8398/files?w=1.

### DIFF
--- a/examples/node/pdf2png/pdf2png.js
+++ b/examples/node/pdf2png/pdf2png.js
@@ -29,22 +29,22 @@ NodeCanvasFactory.prototype = {
     };
   },
 
-  reset: function NodeCanvasFactory_reset(canvasAndContextPair, width, height) {
-    assert(canvasAndContextPair.canvas, 'Canvas is not specified');
+  reset: function NodeCanvasFactory_reset(canvasAndContext, width, height) {
+    assert(canvasAndContext.canvas, 'Canvas is not specified');
     assert(width > 0 && height > 0, 'Invalid canvas size');
-    canvasAndContextPair.canvas.width = width;
-    canvasAndContextPair.canvas.height = height;
+    canvasAndContext.canvas.width = width;
+    canvasAndContext.canvas.height = height;
   },
 
-  destroy: function NodeCanvasFactory_destroy(canvasAndContextPair) {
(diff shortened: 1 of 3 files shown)


yurydelendik (changes requested):
I would like to keep factory methods non-static.

Snuffleupagus:
> I would like to keep factory methods non-static.

I can certainly change it back; but if you don't mind explaining, why is that?

Do you simply want to ensure that we have flexibility, if in the future we want to initialize this factory with certain parameters; or is there another reason?
For future reference, I simply want to ensure that I understand your line of reasoning here :-)

yurydelendik:
> Do you simply want to ensure that we have flexibility, if in the future we want to initialize this factory with certain parameters; or is there another reason?

Our factory does not contain state, however it may (e.g. use). The static method are not normal object methods and by doing this refactoring you are promoting using its type as an object not an instance. Notice that JavaScript not really best language to see this difference, but you will not be able to do such thing in other languages (I don't think it's possible in TypeScript too), so it looks weird to me and will make a reader that knows the factory pattern (e.g. http://www.oodesign.com/factory-pattern.html) to read the code. More format explanation will be that rendering is using ICanvasFactory interface and it is logical that a class will implement this interface, but in your case you force type of a class to have such interface. Basic way to express what you just did it is via singleton pattern.

Also it's breaking API change.

Snuffleupagus:
@yurydelendik Thanks for the in-depth answer, I've updated the PR.

yurydelendik (approved):
Can you check if `constructor({ baseUrl = null, isCompressed = false, }) {` works (and above) better?

Snuffleupagus:
> Can you check if constructor({ baseUrl = null, isCompressed = false, }) { works (and above) better?

Sure, that works just fine locally; thank you for the review!

PR #8399, opened 2017-05-11
Wait until `PDFDocumentLoadingTask.destroy` is resolved before calling `done` in the API unit-tests

Snuffleupagus:
Try to ensure that everything, including workers, are properly destroyed before running the next test-case.

*Please note:* Even though this didn't magically fix the timeouts in PR #8284, it's still conceptually the right thing to do which is why I figured that I'd submit this PR.


yurydelendik:
Thank you for the patch.

PR #8401, opened 2017-05-11
Adds gulp dist-install command; using pdfjs-dist package in examples.

yurydelendik:
It will simplify example usage (e.g. now examples can be copied and used as is in the custom project that has pdfjs-dist installed) and avoid git-tag error when `gulp dist` run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,8 @@ The repository contains a hello world example that you can run locally:
 
 + [examples/helloworld/](https://github.com/mozilla/pdf.js/blob/master/examples/helloworld/)
 
+More examples can be found at the examples folder. Some of them are using the pdfjs-dist package, which can be built and installed in this repo directory via `gulp dist-install` command.
+
 For an introduction to the PDF.js code, check out the presentation by our
 contributor Julian Viereck:
 
--- a/examples/learning/helloworld.html
+++ b/examples/learning/helloworld.html
@@ -11,9 +11,9 @@
 <canvas id="the-canvas" style="border:1px  solid black"></canvas>
 
 <!-- for legacy browsers add compatibility.js -->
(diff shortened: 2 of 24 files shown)


README.md
```diff
@@ -106,6 +106,8 @@ The repository contains a hello world example that you can run locally:
 
 + [examples/helloworld/](https://github.com/mozilla/pdf.js/blob/master/examples/helloworld/)
 
+More examples can be found at the examples folder. Some of them are using the pdfjs-dist package, which can be built and installed in this repo direcotry via `gulp dist-install` command.
```
timvandermeij: Nit: `direcotry` should be `directory`

timvandermeij:
Could you update this patch to resolve the conflict?

examples/learning/helloworld.html
```diff
@@ -11,9 +11,9 @@
 <canvas id="the-canvas" style="border:1px  solid black"></canvas>
 
 <!-- for legacy browsers add compatibility.js -->
-<!--<script src="../compatibility.js"></script>-->
+<!--<script src="../../node_modules/pdfjs-dist/web/compatibility.js"></script>-->
 
-<script src="../../build/generic/build/pdf.js"></script>
+<script src="../../node_modules/pdfjs-dist/pdf.js"></script>
```
timvandermeij: I think this should end with `/build/pdf.js` instead. Now it throws `ReferenceError: PDFJS is not defined`.

timvandermeij (changes requested):

examples/learning/helloworld.html
yurydelendik: Fixed, thanks.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #8406, opened 2017-05-14
Gracefully ignore transaction failure in plugin-babel-cached

Rob--W:
Before this PR, any DB transaction error prevents the plugin from completing. Since the cache is a best-efforts attempt, I fix this by ignoring the error (instead of rethrowing the error).


timvandermeij:
Looks good, thank you!

PR #8409, opened 2017-05-15
[DON'T MERGE] Enable logging for chrome.

brendandahl:


timvandermeij:
Closing since the issues with the bots are fixed. Both the Windows and Linux bots now run Firefox and Chrome. I updated the FAQ wiki page as well.

PR #8411, opened 2017-05-16
Replace unnecessary `var self = this` statements with arrow functions in `web/pdf_page_view.js` and `web/pdf_thumbnail_view.js`

Snuffleupagus:
Also replaces `var` with `let` in code that's touched in the patch. Please note that this should be completely safe, for two separate reasons, since trying to access let in a scope where it's not defined is first of all a runtime error and second of all an ESLint error (thanks to the `no-undef` rule).


timvandermeij (approved):

timvandermeij:
Thank you for the patch!

PR #7911, opened 2016-12-22
When exporting pdf to canvas, can pass a background color parameter

dennisrjohn:
When exporting pdf to canvas, can pass a background parameter (rgb or rgba) to set the canvas background color to something other than white.

We are exporting pdfs to canvas and need the ability to change the background color of the canvas, or make it transparent.

This changes allows a background parameter be sent to the render method, which then sets the background color of the exported canvas:
page.render({canvasContext: context, viewport: viewport, background: rgba(255, 255, 255, 0})

If no background is sent, the canvas background defaults to white.

PR #8413, opened 2017-05-16
[api-minor] Adds additional parameter so background color of canvas can be set

cgreening:
Adds an additional parameter that lets you set the background color of the canvas when rendering pdf.

e.g. for a transparent background:

```js
page.render({ canvasContext, viewport, backgroundColor: 'rgba(0,0,0,0)' });
```

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -717,6 +717,11 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
  * @property {Object} canvasFactory - (optional) The factory that will be used
  *                    when creating canvases. The default value is
  *                    {DOMCanvasFactory}.
+ * @property {Object} background - (optional) Background to use for the canvas.
+ *                    Can use any valid canvas.fillStyle: A DOMString parsed as
+ *                    CSS <color> value, a CanvasGradient object (a linear or
+ *                    radial gradient) or a CanvasPattern object (a repetitive
+ *                    image). The default value is 'rgb(255,255,255)'.
  */
 
 /**
@@ -2128,7 +2133,12 @@ var InternalRenderTask = (function InternalRenderTaskClosure() {
                                     this.objs, this.canvasFactory,
                                     params.imageLayer);
 
-      this.gfx.beginDrawing(params.transform, params.viewport, transparency);
+      this.gfx.beginDrawing({
+        transform: params.transform,
+        viewport: params.viewport,
+        transparency,
+        background: params.background,
+      });
       this.operatorListIdx = 0;
       this.graphicsReady = true;
       if (this.graphicsReadyCallback) {
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -705,8 +705,8 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
 
   CanvasGraphics.prototype = {
 
-    beginDrawing: function CanvasGraphics_beginDrawing(transform, viewport,
-                                                       transparency) {
+    beginDrawing({ transform, viewport, transparency,
+                   background = null, }) {
       // For pdfs that use blend modes we have to clear the canvas else certain
       // blend modes can look wrong since we'd be blending with a white
       // backdrop. The problem with a transparent backdrop though is we then
@@ -716,7 +716,7 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       var height = this.ctx.canvas.height;
 
       this.ctx.save();
-      this.ctx.fillStyle = 'rgb(255, 255, 255)';
+      this.ctx.fillStyle = background || 'rgb(255, 255, 255)';
       this.ctx.fillRect(0, 0, width, height);
       this.ctx.restore();
 
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -14,6 +14,9 @@
  */
 
 import {
+  buildGetDocumentParams, NodeFileReaderFactory, TEST_PDFS_PATH
+} from './test_utils';
+import {
   createPromiseCapability, FontType, InvalidPDFException, isNodeJS,
   MissingPDFException, PasswordException, PasswordResponses, StreamType,
   stringToBytes
@@ -24,28 +27,7 @@ import {
 import {
   getDocument, PDFDocumentProxy, PDFPageProxy
 } from '../../src/display/api';
-import { NodeFileReaderFactory } from './test_utils';
 import { PDFJS } from '../../src/display/global';
-
-const TEST_PDFS_PATH = {
-  dom: '../pdfs/',
-  node: './test/pdfs/',
-};
-
-function buildGetDocumentParams(filename, options) {
-  let params = Object.create(null);
-  if (isNodeJS()) {
-    params.data = NodeFileReaderFactory.fetch({
-      path: TEST_PDFS_PATH.node + filename,
-    });
-  } else {
-    params.url = new URL(TEST_PDFS_PATH.dom + filename, window.location).href;
-  }
-  for (let option in options) {
-    params[option] = options[option];
-  }
-  return params;
-}
 
 describe('api', function() {
   let basicApiFileName = 'basicapi.pdf';
--- a/test/unit/custom_spec.js
+++ b/test/unit/custom_spec.js
@@ -1,0 +1,110 @@
+/* Copyright 2017 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { buildGetDocumentParams } from './test_utils';
+import { DOMCanvasFactory } from '../../src/display/dom_utils';
+import { getDocument } from '../../src/display/api';
+import { isNodeJS } from '../../src/shared/util';
+import { PDFJS } from '../../src/display/global';
+
+function getTopLeftPixel(canvasContext) {
+  let imgData = canvasContext.getImageData(0, 0, 1, 1);
+  return {
+    r: imgData.data[0],
+    g: imgData.data[1],
+    b: imgData.data[2],
+    a: imgData.data[3],
+  };
+}
+
+describe('custom canvas rendering', function() {
+  let transparentGetDocumentParams = buildGetDocumentParams('transparent.pdf');
+
+  let CanvasFactory;
+  let loadingTask;
+  let page;
+
+  beforeAll(function(done) {
+    if (isNodeJS()) {
+      PDFJS.pdfjsNext = true;
+      // NOTE: To support running the canvas-related tests in Node.js,
+      // a `NodeCanvasFactory` would need to be added (in test_utils.js).
+    } else {
+      CanvasFactory = new DOMCanvasFactory();
+    }
+    loadingTask = getDocument(transparentGetDocumentParams);
+    loadingTask.promise.then(function(doc) {
+      return doc.getPage(1);
+    }).then(function(data) {
+      page = data;
+      done();
+    }).catch(function (reason) {
+      done.fail(reason);
+    });
+  });
+
+  afterAll(function(done) {
+    CanvasFactory = null;
+    page = null;
+    loadingTask.destroy().then(done);
+  });
+
+  it('renders to canvas with a default white background', function(done) {
+    if (isNodeJS()) {
+      pending('TODO: Support Canvas testing in Node.js.');
+    }
+    var viewport = page.getViewport(1);
+    var canvasAndCtx = CanvasFactory.create(viewport.width, viewport.height);
+
+    page.render({
+      canvasContext: canvasAndCtx.context,
+      viewport,
+    }).then(function() {
+      var { r, g, b, a } = getTopLeftPixel(canvasAndCtx.context);
+      CanvasFactory.destroy(canvasAndCtx);
+      expect(r).toEqual(255);
+      expect(g).toEqual(255);
+      expect(b).toEqual(255);
+      expect(a).toEqual(255);
+      done();
+    }).catch(function (reason) {
+      done(reason);
+    });
+  });
+
+  it('renders to canvas with a custom background', function(done) {
+    if (isNodeJS()) {
+      pending('TODO: Support Canvas testing in Node.js.');
+    }
+    var viewport = page.getViewport(1);
+    var canvasAndCtx = CanvasFactory.create(viewport.width, viewport.height);
+
+    page.render({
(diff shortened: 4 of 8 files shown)


Snuffleupagus:
Isn't this basically a duplicate of PR #7911?

> e.g. for a transparent background:

Based on [this comment](https://github.com/mozilla/pdf.js/blob/08f8b68f12bd6a8e5479777c6f0f61b6feba02c7/src/display/canvas.js#L712-L713), it seems that one side-effect of transparent canvas would be worse text rendering!?

cgreening:
Didn't realise someone had already tried to get this in. 

I don't completely understand why it was rejected.

yurydelendik:
@cgreening there are multiple reason, but main one (from my point of view) is we are hesitant to accept something without tests. Since we are doing refactoring something, we might break some functionality we don't use. By accepting a patch, we usually committed to support it in the future, without testing it is just a burden :)

src/display/canvas.js
```diff
@@ -706,7 +706,8 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
   CanvasGraphics.prototype = {
 
     beginDrawing: function CanvasGraphics_beginDrawing(transform, viewport,
```
yurydelendik: refactor beginDrawing to destructure parameters:

```
beginDrawing({transform, viewport, transparency, backgroundColor})
```

yurydelendik (changes requested):
That said. Let's add a tests for that, e.g. create a new file "custom_spec.js" to the unit tests that will read canvas pixels.

src/display/api.js
```diff
@@ -2128,7 +2128,10 @@ var InternalRenderTask = (function InternalRenderTaskClosure() {
                                     this.objs, this.canvasFactory,
                                     params.imageLayer);
 
-      this.gfx.beginDrawing(params.transform, params.viewport, transparency);
+      this.gfx.beginDrawing(params.transform,
+                            params.viewport,
+                            transparency,
+                            params.backgroundColor);
```
yurydelendik: Add documentation about new parameter https://github.com/MRMBRAND/pdf.js/blob/f52376929d7584d41f38b2c741b6ed29b17fa181/src/display/api.js#L700

cgreening:
Will update the code with requested changes and add some tests.

src/display/canvas.js
cgreening: Not sure this can be done as the transparency argument doesn't seem to come from params:

```js
      this.gfx.beginDrawing(params.transform,
                            params.viewport,
                            transparency,
                            params.backgroundColor);
```

src/display/api.js
```diff
@@ -717,6 +717,8 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
  * @property {Object} canvasFactory - (optional) The factory that will be used
  *                    when creating canvases. The default value is
  *                    {DOMCanvasFactory}.
+ * @property {string} backgroundColor - Background color to use for the canvas.
+ *                    The default value is 'rgb(255,255,255)'
```
Snuffleupagus: Please mark this as `(optional)`, refer e.g. to the `canvasFactory` just above. Also, the `The default value is 'rgb(255,255,255)'` sentence is missing a period at the end.

Finally, when all review comments have been addressed, please squash the commits into one; see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

src/display/api.js
cgreening: Done - let me know if it has enough details

cgreening:
Tests and documentation added - let me know if you need anything else.

src/display/api.js
cgreening: done

test/unit/api_spec.js
```diff
@@ -12,7 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+import { buildGetDocumentParams, NodeFileReaderFactory, TEST_PDFS_PATH } from './test_utils';
```
Snuffleupagus: This line is too long, it should be:
```javascript

import {
  buildGetDocumentParams, NodeFileReaderFactory, TEST_PDFS_PATH
} from './test_utils';
```

And please keep *one* blank line between the copyright header and the first `import`.

test/unit/test_utils.js
```diff
@@ -57,4 +78,6 @@ class NodeCMapReaderFactory {
 export {
   NodeFileReaderFactory,
   NodeCMapReaderFactory,
+  buildGetDocumentParams,
+  TEST_PDFS_PATH
```
Snuffleupagus: Nit: Add a colon at the end here, i.e. `TEST_PDFS_PATH,`.

test/unit/test_utils.js
```diff
@@ -23,6 +23,27 @@ class NodeFileReaderFactory {
...
+  } else {
+    params.url = new URL(TEST_PDFS_PATH.dom + filename, window.location).href;
+  }
+  for (let option in options) {
+    params[option] = options[option];
+  }
+  return params;
+}
+
+
```
Snuffleupagus: Nit: Please remove one of the blank lines here, one is enough.

test/unit/custom_spec.js
```diff
@@ -0,0 +1,99 @@
+import { buildGetDocumentParams } from './test_utils';
```
Snuffleupagus: This file is missing the required license header, see e.g. https://github.com/mozilla/pdf.js/blob/master/src/license_header.js.

test/unit/custom_spec.js
```diff
@@ -0,0 +1,99 @@
...
+import { isNodeJS } from '../../src/shared/util';
+import { PDFJS } from '../../src/display/global';
+
+function getTopLeftPixel(canvasContext) {
+  var imgData = canvasContext.getImageData(0, 0, 1, 1);
+  var r = imgData.data[0];
+  var g = imgData.data[1];
+  var b = imgData.data[2];
+  var a = imgData.data[3];
+  return { r, g, b, a};
```
Snuffleupagus: I'm not sure that the temporary variables adds much value here, let's just make it:
```javascript
  return {
    r: imgData.data[0],
    g: imgData.data[1],
    b: imgData.data[2],
    alpha: imgData.data[3],
  };
```

test/unit/custom_spec.js
```diff
@@ -0,0 +1,99 @@
...
+}
+
+describe('custom canvas rendering', function() {
+  var transparentFileName = 'transparent.pdf';
+  var transparentGetDocumentParams = buildGetDocumentParams(transparentFileName);
```
Snuffleupagus: The `transparent.pdf` file isn't added to the patch, which cause failures in all the tests below. Also, since this line is too long, let's just do this instead:
```javascript
var transparentGetDocumentParams = buildGetDocumentParams('transparent.pdf');
```

test/unit/custom_spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+  var CanvasFactory;
+  var loadingTask;
+  var pdfDocument, page;
```
Snuffleupagus: I don't think that `pdfDocument` needs to be defined here.

test/unit/custom_spec.js
```diff
@@ -0,0 +1,99 @@
...
+      PDFJS.pdfjsNext = true;
+      // NOTE: To support running the canvas-related tests in Node.js,
+      // a `NodeCanvasFactory` would need to be added (in test_utils.js).
+    } else {
+      CanvasFactory = new DOMCanvasFactory();
+    }
+    loadingTask = getDocument(transparentGetDocumentParams);
+    loadingTask.promise.then(function(doc) {
+      pdfDocument = doc;
+      return pdfDocument.getPage(1);
```
Snuffleupagus: `return doc.getPage(1);` should work just as well here, and removes the otherwise unnecessary `pdfDocument` variable.

test/unit/custom_spec.js
```diff
@@ -0,0 +1,99 @@
...
+      page = data;
+      done();
+    }).catch(function (reason) {
+      done.fail(reason);
+    });
+  });
+
+  afterAll(function(done) {
+    CanvasFactory = null;
+    done();
```
Snuffleupagus: Please clean up the `loadingTask` properly here, i.e. `loadingTask.destroy().then(done):`.

test/unit/custom_spec.js
```diff
@@ -0,0 +1,99 @@
...
+      expect(r).toEqual(255);
+      expect(g).toEqual(255);
+      expect(b).toEqual(255);
+      expect(a).toEqual(255);
+    });
+    promiseDone.then(function() {
+      done();
+    }).catch(function (reason) {
+      done.fail(reason);
+    });
```
Snuffleupagus: Nit: The above can probably be simplified to simply (applies in the test below as well):
```javascript
page.render({
  canvasContext: canvasAndCtx.context,
  viewport,
}).then(function() {
  var { r, g, b, a } = getTopLeftPixel(canvasAndCtx.context);
  CanvasFactory.destroy(canvasAndCtx);
  expect(r).toEqual(255);
  expect(g).toEqual(255);
  expect(b).toEqual(255);
  expect(a).toEqual(255);

  done();
}).catch(function (reason) {
  done.fail(reason);
});
```

src/display/canvas.js
Snuffleupagus: I'm assuming that @yurydelendik meant that you should change the method signature like this (and also change `api.js` as indicated elsewhere):
```javascript
beginDrawing({ transform, viewport, transparency,
               backgroundColor = null, }) {
```

src/display/api.js
Snuffleupagus: With the changed signature of `beginDrawing`, this becomes:
```javascript
this.gfx.beginDrawing({
  transform: params.transform,
  viewport: params.viewport,
  transparency,
  backgroundColor: params.backgroundColor,
});
```

src/display/api.js
```diff
@@ -717,6 +717,8 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
  * @property {Object} canvasFactory - (optional) The factory that will be used
  *                    when creating canvases. The default value is
  *                    {DOMCanvasFactory}.
+ * @property {string} backgroundColor - (optional) Background color to use for the canvas.
+ *                    The default value is 'rgb(255,255,255).'
```
Snuffleupagus: Also, please keep the line lengths in mind, i.e. this needs to be
```
* @property {string} backgroundColor - (optional) Background color to use for
*                    the canvas. The default value is 'rgb(255,255,255)'.
```

test/unit/custom_spec.js
```diff
@@ -0,0 +1,99 @@
...
+    }).then(function(data) {
+      page = data;
+      done();
+    }).catch(function (reason) {
+      done.fail(reason);
+    });
+  });
+
+  afterAll(function(done) {
+    CanvasFactory = null;
```
Snuffleupagus: Please add `page = null;` here too.

Snuffleupagus (commented):
Thanks for adding unit-tests!
However, none of them currently work, since they reference a PDF file that isn't included in the patch.

Also, please use `let` rather than `var` when you're adding *new* code.

cgreening:
Should be all the comments addressed - let me know if I've missed one.

test/unit/custom_spec.js
cgreening: done

test/unit/custom_spec.js
cgreening: done

src/display/canvas.js
```diff
@@ -705,8 +705,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
 
   CanvasGraphics.prototype = {
 
-    beginDrawing: function CanvasGraphics_beginDrawing(transform, viewport,
-                                                       transparency) {
+    beginDrawing: function CanvasGraphics_beginDrawing({
+      transform, viewport, transparency, backgroundColor = null,
+    }) {
```
Snuffleupagus: This is overly verbose, taking full advantage of ES6 notation this can be simplified to:
```javascript
beginDrawing({ transform, viewport, transparency,
               backgroundColor = null, }) {
```
Refer to https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015.

test/unit/custom_spec.js
```diff
@@ -0,0 +1,111 @@
...
+    page = null;
+    loadingTask.destroy().then(done);
+    done();
```
Snuffleupagus: This line needs to be removed, since the point of `loadingTask.destroy().then(done);` just above is to wait for everything to be terminated before continuing with the tests.

src/display/canvas.js
cgreening: That seems to be the style of all the functions in the canvas.js file. I'm happy to clean that up as part of this pull request - but it might be better in a new one?

test/unit/custom_spec.js
cgreening: done

src/display/canvas.js
yurydelendik: We slowly (since not all the code covered by tests) refactoring that to new style. In sake of this PR, we prefer to change only the affected code.

src/display/api.js
```diff
@@ -717,6 +717,8 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
  * @property {Object} canvasFactory - (optional) The factory that will be used
  *                    when creating canvases. The default value is
  *                    {DOMCanvasFactory}.
+ * @property {string} backgroundColor - (optional) Background color to use for
+ *                    the canvas. The default value is 'rgb(255,255,255)'.
```
yurydelendik: nit: can you make a note that the color is specified in the CSS format? Actually it can be string, CanvasGradient, or CanvasPattern -- someone creative might apply paper texture as a background, no?

yurydelendik (approved):

src/display/api.js
cgreening: Have added more docs and changes the name of the parameter to "background" as it's not necessarily a color.

src/display/canvas.js
```diff
@@ -705,8 +705,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
 
   CanvasGraphics.prototype = {
 
-    beginDrawing: function CanvasGraphics_beginDrawing(transform, viewport,
-                                                       transparency) {
+    beginDrawing({
+      transform, viewport, transparency, background = null,
+    }) {
```
Snuffleupagus: This looks kind of weird, compared to existing code (e.g. the placement of the `{` bracket is strange); I think it ought to be:
```javascript
beginDrawing({ transform, viewport, transparency,
               background = null, }) {
```

Snuffleupagus: This looks kind of weird, compared to existing method signatures with ES6 parameter destructuring (e.g. the placement of the `{` bracket is strange); it should be:
```javascript
beginDrawing({ transform, viewport, transparency,
               background = null, }) {
```

cgreening: done

Snuffleupagus:
Thank you for the patch!

timvandermeij:
I'm a bit confused: are the Windows failures expected here?

Snuffleupagus:
*Please note:* The test "failures" in Firefox on the Windows bot in the *latest* test run isn't caused by this patch, since I can replicate all these of those changes locally (on Windows) against `master`.

PR #8415, opened 2017-05-17
Add "npm run" commands for "gulp" commands

pardoman:
# Use case
I don't like having gulp installed globally.
Instead, I prefer having it as a dev-dependency, which can be located and run by npm.

# The Change
The change is simple, add a couple entries to `package.json` which allows typing `npm run server` to do the same as `gulp server`. Similarly, it adds `npm run generate` for `gulp generate`.

That's it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,9 @@
     "yargs": "^3.14.0"
   },
   "scripts": {
-    "test": "gulp lint unittestcli externaltest"
+    "test": "gulp lint unittestcli externaltest",
+    "server": "gulp server",
+    "generic": "gulp generic"
   },
   "repository": {
     "type": "git",


Snuffleupagus:
> I don't like having gulp installed globally.
> Instead, I prefer having it as a dev-dependency, which can be located and run by npm.

I suppose that the question is if we want to support such a "personalized" work-flow, for something that is completely unused for standard PDF.js development. (It's easy to add code, but then it has to be maintained in the future.)

@yurydelendik What is your option on this patch?

---
> Similarly, it adds npm run generate for gulp generate.

This doesn't look correct, *if* this patch is accepted the commit message should probably be changed to actually refer to `generic` instead.

yurydelendik:
> I prefer having it as a dev-dependency

> What is your option on this patch?

I install most of in non-global configuration. Personally I don't really feel a need for such shortcuts, commands like`node_modules/.bin/gulp`, PATH with `npm bin` or aliases work well for me, since I'm "not following install global"-pattern, but that my choice.

The list of available commands (server, generic) is incomplete, and I feel we need to define full list of main commands or don't do it at all.

pardoman:
Hey guys, thanks for taking a look at the PR so quickly!

> since I'm "not following install global"-pattern, but that my choice.

That's my sentiment as well, thanks for expressing it clearer.


> The list of available commands (server, generic) is incomplete

Good to know. The two npm commands included in this PR are based on what I read in `README.md` after I cloned the repo. I'm still not familiarized with most of the codebase; I bet I can find the full list of commands as I get more acquainted with it.

Cheers.

timvandermeij:
I think the only good solution here would be to use a wildcard, i.e., redirecting `npm run *` to `gulp *`. Personally, I'm not really seeing the value of this otherwise, as it would either only work for some commands (causing confusion as to which work and which do not) or we have to enumerate all of them, which is bad for maintenance as we would have to update the commands in two places (most likely forgetting this sometime). Moreover, if you don't want to install Gulp globally, why not just use a (Bash) alias?

pardoman:
Agreed that having to enumerate all of them twice wouldn't scale.

I'll close the PR given that it's not as straight forward as I initially though. Thanks!

PR #8416, opened 2017-05-17
Refactor and ES6-ify `PDFLinkService.navigateTo`

Snuffleupagus:
This patch replaces a `var self = this;` statement with arrow functions, and `var` with `let` in `PDFLinkService.navigateTo`.

Furthermore, when I started looking at this method, it quickly became clear that its code is somewhat of a mess. Since I'm one of the persons that have touched this code over the years, I figured that it'd be a good idea to try and clean it up a bit.

### DIFF
--- a/web/pdf_link_service.js
+++ b/web/pdf_link_service.js
@@ -85,71 +85,76 @@ var PDFLinkService = (function PDFLinkServiceClosure() {
     },
 
     /**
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:

PR #8417, opened 2017-05-17
Use eslint-plugin-mozilla and the webextensions environment to avoid defining globals in each file.

Standard8:
eslint-plugin-mozilla has various tools to make it easier to find globals - this patch enables some of them for `extensions/firefox`.

Additionally, I also specify the `webextensions` environment for `extensions/chrome` and also use some importing globals rules to simplify the global definitions there.

eslint-plugin-mozilla does have a recommended configuration, however, give the discussion in #7957 I've left that alone for now.

### DIFF
--- a/extensions/chromium/.eslintrc
+++ b/extensions/chromium/.eslintrc
@@ -3,11 +3,20 @@
     ../../.eslintrc
   ],
 
+  "env": {
+    "webextensions": true
+  },
+
+  "plugins": [
+    "mozilla"
+  ],
+
   "parserOptions": {
     "sourceType": "script"
   },
 
   "rules": {
+    "mozilla/import-globals": "error",
     "object-shorthand": "off",
   },
 }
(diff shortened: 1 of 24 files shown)


Snuffleupagus (approved):
Thank you for the patch!

Snuffleupagus:
> eslint-plugin-mozilla does have a recommended configuration, however, give the discussion in #7957 I've left that alone for now.

I highly appreciate that this PR didn't enforce a default ESLint config, since I'm personally *not convinced* that PDF.js need/should be forced into that; for a couple of reasons:
 - First of all, please note that mozilla-central is basically a downstream consumer of this GitHub repo, and that *generally* PDF.js code shouldn't be modified directly in mozilla-central.
 - Secondly, we're currently in the process of slowly transitioning the code to ES6, which means that some rules simply cannot be enforced.[1]
 - Thirdly, some cutting-edge JavaScript features may not be well supported by Babel yet (we're using Babel to produce certain ES5 friendly build).
 - Finally, the PDF.js source files are bundled with Webpack, which means that some of the rules enforced at the source level might not translate all that well to the resulting `pdf.js`/`pdf.worker.js/`viewer.js` files.

---
[1] Thus far, enforcing ESLint rules has also been used as an opportunity to cleanup/modernize surrounding code too; hence using *only* automatic fixups of lint issues of are IMHO not always the best idea.

PR #8323, opened 2017-04-22
Change the `importl10n` script to use the Beta channel, since Aurora is being removed

Snuffleupagus:
Since the Aurora channel is being removed, and is no longer going to be receiving updates (with the possible exception of security bugs), we need to change the target for the `gulp importl10n` command.

For more information, please refer to:
 - https://groups.google.com/forum/#!topic/firefox-dev/q4Eb5nEcfSg
 - https://hacks.mozilla.org/2017/04/simplifying-firefox-release-channels/
 - https://mozilla.github.io/process-releases/draft/development_overview/
 - http://release.mozilla.org/firefox/release/2017/04/17/Dawn-Project-FAQ.html

PR #8421, opened 2017-05-18
Change the `importl10n` script to use the Nightly channel

Snuffleupagus:
It seems that PR #8323 unfortunately didn't pick the right channel for pulling translations from, according to the information in:
 - https://groups.google.com/forum/#!topic/mozilla.dev.l10n/vzmbHLPo3rg
 - https://github.com/mozilla-l10n/localizer-documentation/blob/master/misc/aurora_faqs.md

My apologies for the unnecessary churn in this code!


yurydelendik:
Thank you for the patch.

PR #8430, opened 2017-05-19
Adds sendWithStream method in MessageHandler.

mukulmishra18:
This is second PR for the project [Streams API in PDF.js](https://github.com/mozilla/pdf.js/projects/4).

This PR adds a method `sendWithStream` in `MessageHandler` at shared/utils. We can use this method to pass data between main and worker thread using Streams API. This method returns a ReadableStream.
The behaviour of stream is just like the standard [ReadableStream Constructor](https://streams.spec.whatwg.org/#rs-constructor).

This PR also contains tests for the functionalities of stream underlyingSource(pull, cancel) and streamSink(read data with blocking promise, error).

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -1214,24 +1214,50 @@ var createObjectURL = (function createObjectURLClosure() {
   };
 })();
 
+function resolveCall(fn, args, thisArg = null) {
+  if (!fn) {
+    return Promise.resolve(undefined);
+  }
+  return new Promise((resolve, reject) => {
+    resolve(fn.apply(thisArg, args));
+  });
+}
+
+function resolveOrReject(capability, success, reason) {
+  if (success) {
+    capability.resolve();
+  } else {
+    capability.reject(reason);
+  }
+}
+
+function finalize(promise) {
+  return Promise.resolve(promise).catch(() => {});
+}
+
 function MessageHandler(sourceName, targetName, comObj) {
   this.sourceName = sourceName;
   this.targetName = targetName;
   this.comObj = comObj;
-  this.callbackIndex = 1;
+  this.callbackId = 1;
+  this.streamId = 1;
   this.postMessageTransfers = true;
-  var callbacksCapabilities = this.callbacksCapabilities = Object.create(null);
-  var ah = this.actionHandler = Object.create(null);
+  this.streamSinks = Object.create(null);
+  this.streamControllers = Object.create(null);
+  let callbacksCapabilities = this.callbacksCapabilities = Object.create(null);
+  let ah = this.actionHandler = Object.create(null);
 
   this._onComObjOnMessage = (event) => {
-    var data = event.data;
+    let data = event.data;
     if (data.targetName !== this.sourceName) {
       return;
     }
-    if (data.isReply) {
-      var callbackId = data.callbackId;
+    if (data.stream) {
+      this._processStreamMessage(data);
+    } else if (data.isReply) {
+      let callbackId = data.callbackId;
       if (data.callbackId in callbacksCapabilities) {
-        var callback = callbacksCapabilities[callbackId];
+        let callback = callbacksCapabilities[callbackId];
         delete callbacksCapabilities[callbackId];
         if ('error' in data) {
           callback.reject(data.error);
@@ -1242,13 +1268,13 @@ function MessageHandler(sourceName, targetName, comObj) {
         error('Cannot resolve callback ' + callbackId);
       }
     } else if (data.action in ah) {
-      var action = ah[data.action];
+      let action = ah[data.action];
       if (data.callbackId) {
-        var sourceName = this.sourceName;
-        var targetName = data.sourceName;
+        let sourceName = this.sourceName;
+        let targetName = data.sourceName;
         Promise.resolve().then(function () {
           return action[0].call(action[1], data.data);
-        }).then(function (result) {
+        }).then((result) => {
           comObj.postMessage({
             sourceName,
             targetName,
@@ -1256,7 +1282,7 @@ function MessageHandler(sourceName, targetName, comObj) {
             callbackId: data.callbackId,
             data: result,
           });
-        }, function (reason) {
+        }, (reason) => {
           if (reason instanceof Error) {
             // Serialize error to avoid "DataCloneError"
             reason = reason + '';
@@ -1269,6 +1295,8 @@ function MessageHandler(sourceName, targetName, comObj) {
             error: reason,
           });
         });
+      } else if (data.streamId) {
+        this._createStreamSink(data);
       } else {
         action[0].call(action[1], data.data);
       }
@@ -1289,9 +1317,9 @@ MessageHandler.prototype = {
   },
   /**
    * Sends a message to the comObj to invoke the action with the supplied data.
-   * @param {String} actionName Action to call.
-   * @param {JSON} data JSON data to send.
-   * @param {Array} [transfers] Optional list of transfers/ArrayBuffers
+   * @param {String} actionName - Action to call.
+   * @param {JSON} data - JSON data to send.
+   * @param {Array} [transfers] - Optional list of transfers/ArrayBuffers
    */
   send(actionName, data, transfers) {
     var message = {
@@ -1304,14 +1332,14 @@ MessageHandler.prototype = {
   },
   /**
    * Sends a message to the comObj to invoke the action with the supplied data.
-   * Expects that other side will callback with the response.
-   * @param {String} actionName Action to call.
-   * @param {JSON} data JSON data to send.
-   * @param {Array} [transfers] Optional list of transfers/ArrayBuffers.
+   * Expects that the other side will callback with the response.
+   * @param {String} actionName - Action to call.
+   * @param {JSON} data - JSON data to send.
+   * @param {Array} [transfers] - Optional list of transfers/ArrayBuffers.
    * @returns {Promise} Promise to be resolved with response data.
    */
   sendWithPromise(actionName, data, transfers) {
-    var callbackId = this.callbackIndex++;
+    var callbackId = this.callbackId++;
     var message = {
       sourceName: this.sourceName,
       targetName: this.targetName,
@@ -1329,9 +1357,214 @@ MessageHandler.prototype = {
     return capability.promise;
   },
   /**
+   * Sends a message to the comObj to invoke the action with the supplied data.
+   * Expect that the other side will callback to signal 'start_complete'.
+   * @param {String} actionName - Action to call.
+   * @param {JSON} data - JSON data to send.
+   * @param {Object} queueingStrategy - strategy to signal backpressure based on
+   *                 internal queue.
+   * @param {Array} [transfers] - Optional list of transfers/ArrayBuffers.
+   * @return {ReadableStream} ReadableStream to read data in chunks.
+   */
+  sendWithStream(actionName, data, queueingStrategy, transfers) {
+    let streamId = this.streamId++;
+    let sourceName = this.sourceName;
+    let targetName = this.targetName;
+
+    return new ReadableStream({
+      start: (controller) => {
+        let startCapability = createPromiseCapability();
+        this.streamControllers[streamId] = {
+          controller,
+          startCall: startCapability,
+        };
+        this.postMessage({
+          sourceName,
+          targetName,
+          action: actionName,
+          streamId,
+          data,
+          desiredSize: controller.desiredSize,
+        });
+        // Return Promise for Async process, to signal success/failure.
+        return startCapability.promise;
+      },
+
+      pull: (controller) => {
+        let pullCapability = createPromiseCapability();
+        this.streamControllers[streamId].pullCall = pullCapability;
+        this.postMessage({
+          sourceName,
+          targetName,
+          stream: 'pull',
+          streamId,
+          desiredSize: controller.desiredSize,
+        });
+        // Returning Promise will not call "pull"
+        // again until current pull is resolved.
+        return pullCapability.promise;
+      },
+
+      cancel: (reason) => {
+        let cancelCapability = createPromiseCapability();
+        this.streamControllers[streamId].cancelCall = cancelCapability;
+        this.postMessage({
+          sourceName,
+          targetName,
+          stream: 'cancel',
+          reason,
+          streamId,
+        });
+        // Return Promise to signal success or failure.
+        return cancelCapability.promise;
+      },
+    }, queueingStrategy);
+  },
+
+  _createStreamSink(data) {
+    let self = this;
+    let action = this.actionHandler[data.action];
+    let streamId = data.streamId;
+    let desiredSize = data.desiredSize;
+    let sourceName = this.sourceName;
+    let targetName = data.sourceName;
+    let capability = createPromiseCapability();
+
+    let sendStreamRequest = ({ stream, chunk, success, reason, }) => {
+      this.comObj.postMessage({ sourceName, targetName, stream, streamId,
+                                chunk, success, reason, });
+    };
+
+    let streamSink = {
+      enqueue(chunk, size = 1) {
+        let lastDesiredSize = this.desiredSize;
+        this.desiredSize -= size;
+        // Enqueue decreases the desiredSize property of sink,
+        // so when it changes from positive to negative,
+        // set ready as unresolved promise.
+        if (lastDesiredSize > 0 && this.desiredSize <= 0) {
+          this.sinkCapability = createPromiseCapability();
+          this.ready = this.sinkCapability.promise;
+        }
+        sendStreamRequest({ stream: 'enqueue', chunk, });
+      },
+
+      close() {
+        sendStreamRequest({ stream: 'close', });
+        delete self.streamSinks[streamId];
+      },
+
+      error(reason) {
+        sendStreamRequest({ stream: 'error', reason, });
+      },
+
+      sinkCapability: capability,
+      onPull: null,
+      onCancel: null,
+      desiredSize,
+      ready: null,
+    };
+
+    streamSink.sinkCapability.resolve();
+    streamSink.ready = streamSink.sinkCapability.promise;
+    this.streamSinks[streamId] = streamSink;
+    resolveCall(action[0], [data.data, streamSink], action[1]).then(() => {
+      sendStreamRequest({ stream: 'start_complete', success: true, });
+    }, (reason) => {
+      sendStreamRequest({ stream: 'start_complete', success: false, reason, });
+    });
+  },
+
+  _processStreamMessage(data) {
+    let sourceName = this.sourceName;
+    let targetName = data.sourceName;
+    let streamId = data.streamId;
+
+    let sendStreamResponse = ({ stream, success, reason, }) => {
+      this.comObj.postMessage({ sourceName, targetName, stream,
+                                success, streamId, reason, });
+    };
+
+    let deleteStreamController = () => {
+      // Delete streamController only when start, pull and
+      // cancel callbacks are resolved, to avoid "TypeError".
+      Promise.all([
+        this.streamControllers[data.streamId].startCall,
+        this.streamControllers[data.streamId].pullCall,
+        this.streamControllers[data.streamId].cancelCall
+      ].map(function(capability) {
+        return capability && finalize(capability.promise);
+      })).then(() => {
+        delete this.streamControllers[data.streamId];
+      });
+    };
+
+    switch (data.stream) {
+      case 'start_complete':
+        resolveOrReject(this.streamControllers[data.streamId].startCall,
+                        data.success, data.reason);
+        break;
+      case 'pull_complete':
+        resolveOrReject(this.streamControllers[data.streamId].pullCall,
+                        data.success, data.reason);
+        break;
+      case 'pull':
+        // Ignore any pull after close is called.
+        if (!this.streamSinks[data.streamId]) {
+          sendStreamResponse({ stream: 'pull_complete', success: true, });
+          break;
+        }
+        // Pull increases the desiredSize property of sink,
+        // so when it changes from negative to positive,
+        // set ready property as resolved promise.
+        if (this.streamSinks[data.streamId].desiredSize <= 0 &&
+            data.desiredSize > 0) {
+          this.streamSinks[data.streamId].sinkCapability.resolve();
+        }
+        // Reset desiredSize property of sink on every pull.
+        this.streamSinks[data.streamId].desiredSize = data.desiredSize;
+        resolveCall(this.streamSinks[data.streamId].onPull).then(() => {
+          sendStreamResponse({ stream: 'pull_complete', success: true, });
+        }, (reason) => {
+          sendStreamResponse({ stream: 'pull_complete',
+                               success: false, reason, });
+        });
+        break;
+      case 'enqueue':
+        this.streamControllers[data.streamId].controller.enqueue(data.chunk);
+        break;
+      case 'close':
+        this.streamControllers[data.streamId].controller.close();
+        deleteStreamController();
+        break;
+      case 'error':
+        this.streamControllers[data.streamId].controller.error(data.reason);
+        deleteStreamController();
+        break;
+      case 'cancel_complete':
+        resolveOrReject(this.streamControllers[data.streamId].cancelCall,
+                        data.success, data.reason);
+        deleteStreamController();
+        break;
+      case 'cancel':
+        resolveCall(this.streamSinks[data.streamId].onCancel,
+                    [data.reason]).then(() => {
+          sendStreamResponse({ stream: 'cancel_complete', success: true, });
+        }, (reason) => {
+          sendStreamResponse({ stream: 'cancel_complete',
+                               success: false, reason, });
+        });
+        delete this.streamSinks[data.streamId];
+        break;
+      default:
+        throw new Error('Unexpected stream case');
+    }
+  },
+
+  /**
    * Sends raw message to the comObj.
    * @private
-   * @param message {Object} Raw message.
+   * @param {Object} message - Raw message.
    * @param transfers List of transfers/ArrayBuffers, or undefined.
    */
   postMessage(message, transfers) {
--- a/test/unit/util_spec.js
+++ b/test/unit/util_spec.js
@@ -20,46 +20,46 @@ import {
 describe('util', function() {
   describe('stringToPDFString', function() {
     it('handles ISO Latin 1 strings', function() {
-      var str = '\x8Dstring\x8E';
+      let str = '\x8Dstring\x8E';
       expect(stringToPDFString(str)).toEqual('\u201Cstring\u201D');
     });
 
     it('handles UTF-16BE strings', function() {
-      var str = '\xFE\xFF\x00\x73\x00\x74\x00\x72\x00\x69\x00\x6E\x00\x67';
+      let str = '\xFE\xFF\x00\x73\x00\x74\x00\x72\x00\x69\x00\x6E\x00\x67';
       expect(stringToPDFString(str)).toEqual('string');
     });
 
     it('handles empty strings', function() {
       // ISO Latin 1
-      var str1 = '';
+      let str1 = '';
       expect(stringToPDFString(str1)).toEqual('');
 
       // UTF-16BE
-      var str2 = '\xFE\xFF';
+      let str2 = '\xFE\xFF';
       expect(stringToPDFString(str2)).toEqual('');
     });
   });
 
   describe('removeNullCharacters', function() {
     it('should not modify string without null characters', function() {
-      var str = 'string without null chars';
+      let str = 'string without null chars';
       expect(removeNullCharacters(str)).toEqual('string without null chars');
     });
 
     it('should modify string with null characters', function() {
-      var str = 'string\x00With\x00Null\x00Chars';
+      let str = 'string\x00With\x00Null\x00Chars';
       expect(removeNullCharacters(str)).toEqual('stringWithNullChars');
     });
   });
 
   describe('ReadableStream', function() {
     it('should return an Object', function () {
-      var readable = new ReadableStream();
+      let readable = new ReadableStream();
       expect(typeof readable).toEqual('object');
     });
 
     it('should have property getReader', function () {
-      var readable = new ReadableStream();
+      let readable = new ReadableStream();
       expect(typeof readable.getReader).toEqual('function');
     });
   });
--- a/test/unit/util_stream_spec.js
+++ b/test/unit/util_stream_spec.js
@@ -1,0 +1,378 @@
+/* Copyright 2017 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createPromiseCapability, MessageHandler } from '../../src/shared/util';
+
+describe('util_stream', function () {
+  // Temporary fake port for sending messages between main and worker.
+  class FakePort {
+    constructor() {
+      this._listeners = [];
+      this._deferred = Promise.resolve(undefined);
+    }
+
+    postMessage(obj) {
+      let event = { data: obj, };
+      this._deferred.then(() => {
+        this._listeners.forEach(function (listener) {
+          listener.call(this, event);
+        }, this);
+      });
+    }
+
+    addEventListener(name, listener) {
+      this._listeners.push(listener);
+    }
+
+    removeEventListener(name, listener) {
+      let i = this._listeners.indexOf(listener);
+      this._listeners.splice(i, 1);
+    }
+
+    terminate() {
+      this._listeners = [];
+    }
+  }
+
+  // Sleep function to wait for sometime, similar to setTimeout but faster.
+  function sleep(ticks) {
+    return Promise.resolve().then(() => {
+      return (ticks && sleep(ticks - 1));
+    });
+  }
+
+  describe('sendWithStream', function () {
+    it('should return a ReadableStream', function () {
+      let port = new FakePort();
+      let messageHandler1 = new MessageHandler('main', 'worker', port);
+      let readable = messageHandler1.sendWithStream('fakeHandler');
+      // Check if readable is an instance of ReadableStream.
+      expect(typeof readable).toEqual('object');
+      expect(typeof readable.getReader).toEqual('function');
+    });
+
+    it('should read using a reader', function (done) {
+      let log = '';
+      let port = new FakePort();
+      let messageHandler1 = new MessageHandler('main', 'worker', port);
+      let messageHandler2 = new MessageHandler('worker', 'main', port);
+      messageHandler2.on('fakeHandler', (data, sink) => {
+        sink.onPull = function () {
+          log += 'p';
+        };
+        sink.onCancel = function (reason) {
+          log += 'c';
+        };
+        sink.ready.then(() => {
+          sink.enqueue('hi');
+          return sink.ready;
+        }).then(() => {
+          sink.close();
+        });
+        return sleep(5);
+      });
+      let readable = messageHandler1.sendWithStream('fakeHandler', {}, {
+        highWaterMark: 1,
+        size() {
+          return 1;
+        },
+      });
+      let reader = readable.getReader();
+      sleep(10).then(() => {
+        expect(log).toEqual('');
+        return reader.read();
+      }).then((result) => {
+        expect(log).toEqual('p');
+        expect(result.value).toEqual('hi');
+        expect(result.done).toEqual(false);
+        return sleep(10);
+      }).then(() => {
+        return reader.read();
+      }).then((result) => {
+        expect(result.value).toEqual(undefined);
+        expect(result.done).toEqual(true);
+        done();
+      });
+    });
+
+    it('should not read any data when cancelled', function (done) {
+      let log = '';
+      let port = new FakePort();
+      let messageHandler2 = new MessageHandler('worker', 'main', port);
+      messageHandler2.on('fakeHandler', (data, sink) => {
+        sink.onPull = function () {
+          log += 'p';
+        };
+        sink.onCancel = function (reason) {
+          log += 'c';
+        };
+        log += '0';
+        sink.ready.then(() => {
+          log += '1';
+          sink.enqueue([1, 2, 3, 4], 4);
+          return sink.ready;
+        }).then(() => {
+          log += '2';
+          sink.enqueue([5, 6, 7, 8], 4);
+          return sink.ready;
+        }).then(() => {
+          log += '3';
+          sink.close();
+        }, () => {
+          log += '4';
+        });
+      });
+      let messageHandler1 = new MessageHandler('main', 'worker', port);
+      let readable = messageHandler1.sendWithStream('fakeHandler', {}, {
+        highWaterMark: 4,
+        size(arr) {
+          return arr.length;
+        },
+      });
+
+      let reader = readable.getReader();
+      sleep(10).then(() => {
+        expect(log).toEqual('01');
+        return reader.read();
+      }).then((result) => {
+        expect(result.value).toEqual([1, 2, 3, 4]);
+        expect(result.done).toEqual(false);
+        return sleep(10);
+      }).then(() => {
+        expect(log).toEqual('01p2');
+        return reader.cancel();
+      }).then(() => {
+        expect(log).toEqual('01p2c');
+        done();
+      });
+    });
+
+    it('should not read when errored', function(done) {
+      let log = '';
+      let port = new FakePort();
+      let messageHandler2 = new MessageHandler('worker', 'main', port);
+      messageHandler2.on('fakeHandler', (data, sink) => {
+        sink.onPull = function () {
+          log += 'p';
+        };
+        sink.onCancel = function (reason) {
+          log += 'c';
+        };
+        sink.ready.then(() => {
+          sink.enqueue([1, 2, 3, 4], 4);
+          return sink.ready;
+        }).then(() => {
+          log += 'error';
+          sink.error('error');
+        });
+      });
+      let messageHandler1 = new MessageHandler('main', 'worker', port);
+      let readable = messageHandler1.sendWithStream('fakeHandler', {}, {
+        highWaterMark: 4,
+        size(arr) {
+          return arr.length;
+        },
+      });
+
+      let reader = readable.getReader();
+
+      sleep(10).then(() => {
+        expect(log).toEqual('');
+        return reader.read();
+      }).then((result) => {
+        expect(result.value).toEqual([1, 2, 3, 4]);
+        expect(result.done).toEqual(false);
+        return reader.read();
+      }).then(() => {
+      }, (reason) => {
+        expect(reason).toEqual('error');
+        done();
+      });
+    });
+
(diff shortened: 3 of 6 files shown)


src/shared/util.js
```diff
@@ -1219,16 +1219,111 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+        case 'pull_complete':
+          if (data.success === true || data.success === 'true') {
+            streamControllers[data.streamId].pullCall.resolve();
+          } else {
+            streamControllers[data.streamId].pullCall.reject();
+          }
+          break;
+        case 'pull':
+          // pull increases the desiredSize property of sink, so when it
+          // changes from -ve to +ve set ready property as resolved promise.
```
Snuffleupagus: Unless one already knows what `ve` means, then this comment is unfortunately not very helpful. (This applies to another comment below as well.)

src/shared/util.js
```diff
@@ -1269,6 +1364,75 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+      } else if (data.streamId) {
+        let streamId = data.streamId;
+        let desiredSize = data.desiredSize;
+        let sourceName = this.sourceName;
+        let targetName = data.sourceName;
+        let capability = createPromiseCapability();
+        let streamSink = {
+          enqueue(chunk, size) {
+            var lastDesiredSize = this.desiredSize;
+            this.desiredSize -= size !== undefined ? size : 1;
```
Snuffleupagus: Why not change the signature to `enqueue(chunk, size = 1) {` instead, since this line can then be simplified to `this.desiredSize -= size`?

src/shared/util.js
```diff
@@ -1329,6 +1493,64 @@ MessageHandler.prototype = {
     return capability.promise;
   },
   /**
+   * Sends a message to the comObj to invoke the action with the supplied data.
+   * Expect that other side will callback to signal start_complete.
```
Snuffleupagus: Nit: `Expects that the other side will callback to signal 'start_complete'.`

src/shared/util.js
```diff
@@ -1329,6 +1493,64 @@ MessageHandler.prototype = {
...
+   * @param {String} actionName Action to call.
+   * @param {JSON} data JSON data to send.
+   * @param {Object} queueingStrategy strategy to signal backpressure based on
+   *                 internal queue.
+   * @param {Array} [transfers] Optional list of transfers/ArrayBuffers.
+   * @return {ReadableStream} ReadableStream to read data in chunks.
```
Snuffleupagus: Nit: Let's make this more consistent with how we generally write JSDoc comments (note that the other JSDoc comments in `MessageHandler` aren't entirely correct):
```
* @param {string} actionName - Action to call.
* @param {JSON} data - JSON data to send.
* @param {Object} queueingStrategy - strategy to signal backpressure based on
*   internal queue.
* @param {Array} [transfers] - Optional list of transfers/ArrayBuffers.
* @return {ReadableStream} ReadableStream to read data in chunks.
```

src/shared/util.js
```diff
@@ -1329,6 +1493,64 @@ MessageHandler.prototype = {
...
+   */
+  sendWithStream(actionName, data, queueingStrategy, transfers) {
+      var streamId = this.streamId++;
```
Snuffleupagus: Please reduce the indentation to only two spaces within this method.

test/unit/util_spec.js
```diff
@@ -14,52 +14,52 @@
  */
 
 import {
-  ReadableStream, removeNullCharacters, stringToPDFString
+   ReadableStream, removeNullCharacters, stringToPDFString
```
Snuffleupagus: Unnecessary change, please revert.

test/unit/util_stream_spec.js
```diff
@@ -0,0 +1,311 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { MessageHandler } from '../../src/shared/util';
+
+describe('util stream', function () {
```
Snuffleupagus: Nit: Please make this `util_stream` instead, since that's consistent with existing tests.

test/unit/util_stream_spec.js
```diff
@@ -0,0 +1,311 @@
...
+  // Temporary fake port for sending messages between main and worker.
+  class FakePort {
+    constructor() {
+      this._listeners = [];
+      this._deferred = Promise.resolve(undefined);
+    }
+
+    postMessage(obj) {
+      let e = {data: obj};
```
Snuffleupagus: Please use a more descriptive variable name than `e`.

Snuffleupagus (commented):
I've left a couple of comments, based on a *very* cursory look at the code.

Also, I'm wondering if there's any reason that the new unit-tests cannot be run on Travis?
Otherwise, let's add the new test file to https://github.com/mozilla/pdf.js/blob/master/test/unit/clitests.json.

src/shared/util.js
```diff
@@ -1219,16 +1219,111 @@ function MessageHandler(sourceName, targetName, comObj) {
   this.targetName = targetName;
   this.comObj = comObj;
   this.callbackIndex = 1;
+  this.streamId = 1;
```
mukulmishra18: Yes, sure. I will change it to `streamIndex`

src/shared/util.js
```diff
@@ -1219,16 +1219,111 @@ function MessageHandler(sourceName, targetName, comObj) {
...
       return;
     }
-    if (data.isReply) {
+    if (data.stream) {
+      let sourceName = this.sourceName;
+      let targetName = data.sourceName;
+      let streamId = data.streamId;
+      switch (data.stream) {
+        case 'start_complete':
+          if (data.success === true || data.success === 'true') {
```
mukulmishra18: Just to be at safer side, i thought maybe somehow while sending data between main and worker, `success` will change to string. But i am not sure.
Thanks for your suggestion, i will change this to `if (data.success) {`. This looks more clear.

src/shared/util.js
mukulmishra18: Okay, i will make it `negative`/`positive` instead.

src/shared/util.js
```diff
@@ -1219,16 +1219,111 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+          }
+          break;
+        case 'pull':
+          // pull increases the desiredSize property of sink, so when it
+          // changes from -ve to +ve set ready property as resolved promise.
+          if (streamSinks[data.streamId].desiredSize <= 0 &&
+              data.desiredSize > 0) {
+            streamSinks[data.streamId].sinkCapability.resolve();
+          }
+          // reset desiredSize property of sink on every pull.
```
mukulmishra18: Yeah, sure. Thanks.

src/shared/util.js
mukulmishra18: Great, this is short and more clear. We will do this way.

src/shared/util.js
mukulmishra18: > note that the other JSDoc comments in MessageHandler aren't entirely correct

I referred those comments to write this :)
Okay, i will do as you suggested, and also change other comments in `MessageHandler` to follow this.

src/shared/util.js
mukulmishra18: Okay, sure.

test/unit/util_stream_spec.js
mukulmishra18: Okay, Thanks for your suggestions.

src/shared/util.js
Snuffleupagus: > Yes, sure. I will change it to `streamIndex`

Thanks, but I now notice that this may become problematic with respect to the line length limit; it may be simpler to keep `this.streamId` and just change `this.callbackIndex` -> `this.callbackId` instead :-)

Looking at the existing code, the `callbackId` name is actually already used in most places, except here: https://github.com/mozilla/pdf.js/blob/master/src/shared/util.js#L1313.

src/shared/util.js
Snuffleupagus: > Just to be at safer side, i thought maybe somehow while sending data between main and worker, success will change to string. But i am not sure.

That shouldn't be a problem, since the [structured cloning algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) shouldn't change types for something as simple as a Boolean.

If you want to check for yourself, look at the value/type of `data.isReply` in the current code.

src/shared/util.js
mukulmishra18: > but I now notice that this may become problematic with respect to the line length limit

"_line length limit_"? I think i am not getting this point, can you please elaborate?

src/shared/util.js
mukulmishra18: > That shouldn't be a problem, since the structured cloning algorithm maintains types (and throws on unsupported types).

Okay right, we are using _structured cloning algorithm_ in `postMessage`.
Thanks for clarification.

src/shared/util.js
Snuffleupagus: I just noticed that there's a couple of fairly long lines in the patch already (for example [this one](https://github.com/mozilla/pdf.js/pull/8430/commits/9444da8061e4bfc2c7a827f34752000bbc089fd5#diff-8d2b578d58d4d8c8a2975020fbbc84d9R1263)), and it's possible that using `streamIndex` instead of `streamId` everywhere could perhaps unnecessarily go over the maximum limit of 80 characters per line.

However, it seems that we already use `callbackId` in most places in the existing code, hence I'm suggesting to keep `this.streamId` in this patch and just change the *two* existing occurrences of `this.callbackIndex` to `this.callbackId` instead.

Sorry about given confusing/conflicting advice here!

mukulmishra18: > I'm suggesting to keep `this.streamId` in this patch and just change the two existing occurrences of `this.callbackIndex` to `this.callbackId` instead.

Okay, sure. Thanks for clarification here.

mukulmishra18:
Okay, but why it is failing the tests :-(

`gulp unittestcli` is just running fine on my local machine, with result:
`412 specs, 0 failures, 11 pending specs`

Snuffleupagus:
It seems that one of the tests fail on Travis, and I've noticed that while all tests pass locally, there still seem to be some errors regarding accessing removed properties.

Note that you can easily run all (or part) of the unit-tests locally in a browser, by doing `gulp server` and then pointing your browser to http://localhost:8888/test/unit/unit_test.html. It's also possible to run part of the tests as well, using e.g. http://localhost:8888/test/unit/unit_test.html?spec=util_stream.
Doing that locally, I get two `TypeError`s as indicated below.
```
TypeError: streamControllers[data.streamId] is undefined[Learn More] system.js%20line%204%20%3E%20eval:1206:19
TypeError: streamSinks[data.streamId] is undefined[Learn More] system.js%20line%204%20%3E%20eval:1215:1
```
This suggests that currently there's code trying to access certain properties *after* a `delete ...` expression has been run. Those errors need to be dealt with, since it indicates that the ordering of the various stream events isn't the expected one. (Fixing this might also fix the failure on Travis, since that error looks similar.)

mukulmishra18:
> This suggests that currently there's code trying to access certain properties after a delete ... expression has been run.

Yes, you are right. Actually, as `desiredSize` property of sink in greater than sum of size of all data chunks, so it is not blocking the stream(we are using `sink.ready` property to block the stream when queue is full). Due to this, it is calling `sink.close` [here](https://github.com/mozilla/pdf.js/pull/8430/files#diff-9241beeb85135518feb02f194291f04cR285), before calling pull. Decreasing the size of `desiredSize` is solving this problem.

Thanks for your help here :)

Snuffleupagus:
Testing the latest version of the patch locally (on Windows), I get different behaviour in Firefox vs Chrome:
 - In Firefox Nightly (version 55), there's a `Expected '012p' to equal '012'` failure in the `should read data with blocking promise and buffer whole data into stream` test.
 - In Chrome Beta (version 59), the new unit-tests pass.

This unfortunately suggests that there're issues either in the `streams` polyfill, the `MessageHandler` implementation, the unit-tests; or perhaps some combination of them.
(It might just be a timing issue with the e.g. the `sleep` function used in the unit-tests, or it might point to a deeper issue with regards to cross-browser compatibility.)

mukulmishra18:
> Testing the latest version of the patch locally (on Windows), I get different behaviour in Firefox vs Chrome:
> - In Firefox Nightly (version 55), there's a Expected '012p' to equal '012' failure in the should read data with blocking promise and buffer whole data into stream test.
> - In Chrome Beta (version 59), the new unit-tests pass.

Yes, you are right, i have Chrome(version 52) locally, and it is failing on this also.

But in my opinion, `012` would be the correct value for `log`, as you see in above tests(eg [here](https://github.com/mozilla/pdf.js/pull/8430/files#diff-9241beeb85135518feb02f194291f04cR240)) `log` does not contains any `p`(as we didn't called pull yet).

mukulmishra18:
> It might just be a timing issue with the e.g. the `sleep` function used in the unit-tests, or it might point to a deeper issue with regards to cross-browser compatibility.

I am not sure, but potentially this seems to be a timing problem. Running the test with different timeout value(e.g. `sleep(0)`, or `sleep(1)`[1]), gives different `log` values. I also think, this is not a fault of `sleep` function itself(as instead of `sleep`, i tried `setTimeout` function also).

[1]  **note:** `sleep(1)` and `sleep(10)` has no difference on `log` values

test/unit/util_stream_spec.js
```diff
@@ -0,0 +1,324 @@
...
+      let reader = readable.getReader();
+
+      sleep(10).then(() => {
+        expect(log).toEqual('');
+        return reader.read();
+      }).then((result) => {
+        expect(result.value).toEqual([1, 2, 3, 4]);
+        expect(result.done).toEqual(false);
+        return sleep(10);
+      }).then(() => {
```
yurydelendik: add `reason` param here and use it instead of log for expect below.

yurydelendik (commented):
Please also rebase to new master.

test/unit/util_stream_spec.js
```diff
@@ -0,0 +1,324 @@
...
+        sink.onCancel = function (reason) {
+          log += 'c';
+        };
+        sink.ready.then(() => {
+          sink.enqueue('hi');
+          return sink.ready;
+        }).then(() => {
+          sink.close();
+        });
+      });
```
mukulmishra18: done

test/unit/util_stream_spec.js
mukulmishra18: Okay sure, done in new commit

test/unit/util_stream_spec.js
```diff
@@ -0,0 +1,386 @@
...
+        return reader.read();
+      }).then((result) => {
+        expect(result.value).toEqual(undefined);
+        expect(result.done).toEqual(true);
+        done();
+      });
+    });
+
+    it('should ignore any pull after close is called', function (done) {
+      function createPromiseCapability() {
```
Snuffleupagus: Why are you duplicating this here, please import if from `shared/util.js` instead!

src/shared/util.js
```diff
@@ -1218,20 +1218,108 @@ function MessageHandler(sourceName, targetName, comObj) {
...
     }
-    if (data.isReply) {
-      var callbackId = data.callbackId;
+    if (data.stream) {
+      let sourceName = this.sourceName;
+      let targetName = data.sourceName;
+      let streamId = data.streamId;
+      let sendStreamResponse = ({ stream, success }) => {
+        comObj.postMessage({ sourceName, targetName, stream,
+        success, streamId });
```
Snuffleupagus: Nit: The indentation is off here, please alight the parameters like this:
```javascript
comObj.postMessage({ sourceName, targetName, stream,
                     success, streamId });
```

src/shared/util.js
```diff
@@ -1269,6 +1357,51 @@ function MessageHandler(sourceName, targetName, comObj) {
...
         });
+      } else if (data.streamId) {
+        let streamId = data.streamId;
+        let desiredSize = data.desiredSize;
+        let sourceName = this.sourceName;
+        let targetName = data.sourceName;
+        let capability = createPromiseCapability();
+        let sendStreamRequest = ({ stream, chunk, reason, success }) => {
+          comObj.postMessage({ sourceName, targetName, stream, streamId,
+            chunk, reason, success });
```
Snuffleupagus: Similar to https://github.com/mozilla/pdf.js/pull/8430/files#r119478584.

test/unit/util_stream_spec.js
mukulmishra18: Okay, sure. Thanks.

timvandermeij:
I have a question that is related to, but more general than this pull request. The unit tests are clear and help understand what the Streams API does (because I'm not familiar with it yet myself), but I think it would really help (since this is going to be one of the core components) to have some documentation about the terminology, such as "high watermark", "sink" and "queueing strategy". I think you have a really good reference with your write-up at https://docs.google.com/document/d/1yTPvPBhGkFn7XUntAt8wdVPckxHovYylNAIJOGnGRXc/edit and just extending that should already suffice. This is not something that needs to be done right away, but I just wanted to let you know that I think that it would help a lot with understanding the code.

Other than that, this looks promising. Keep up the nice work!

src/shared/util.js
mukulmishra18: done

src/shared/util.js
mukulmishra18: doen

src/shared/util.js
```diff
@@ -1218,20 +1218,108 @@ function MessageHandler(sourceName, targetName, comObj) {
...
   this._onComObjOnMessage = (event) => {
-    var data = event.data;
+    let data = event.data;
     if (data.targetName !== this.sourceName) {
       return;
     }
-    if (data.isReply) {
-      var callbackId = data.callbackId;
+    if (data.stream) {
+      let sourceName = this.sourceName;
```
yurydelendik: extract entire if (data.stream) into separate method e.g. '_processStreamMessage(data)'

src/shared/util.js
```diff
@@ -1269,6 +1357,51 @@ function MessageHandler(sourceName, targetName, comObj) {
             error: reason
           });
         });
+      } else if (data.streamId) {
```
yurydelendik: extract entire if (data.streamId) into separate method e.g. '_createStreamSink(data)'

src/shared/util.js
```diff
@@ -1218,20 +1218,108 @@ function MessageHandler(sourceName, targetName, comObj) {
...
     if (data.targetName !== this.sourceName) {
       return;
     }
-    if (data.isReply) {
-      var callbackId = data.callbackId;
+    if (data.stream) {
+      let sourceName = this.sourceName;
+      let targetName = data.sourceName;
+      let streamId = data.streamId;
+      let sendStreamResponse = ({ stream, success }) => {
```
yurydelendik: here and at sendStreamRequest add 'reason' parameter after success -- it shall be populated/used when success == false. reject on xxx_complete side shall accept this reason as argument.

src/shared/util.js
```diff
@@ -1218,20 +1218,108 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+        comObj.postMessage({ sourceName, targetName, stream,
+                             success, streamId });
+      };
+      let deleteStreamController = () => {
+        Promise.all([streamControllers[data.streamId].startCall,
```
yurydelendik: nit: startCall on the new line

add .cancelCall promise here too

mukulmishra18:
>  but I think it would really help (since this is going to be one of the core components) to have some documentation about the terminology, such as "high watermark", "sink" and "queueing strategy".

Yes you are right, I am also thinking about the same. Having documentation about the code and terminology will help to understand it better, even i can understand later what i have done :-D
I will try to update the [Streams API for PDF.js](https://docs.google.com/document/d/1yTPvPBhGkFn7XUntAt8wdVPckxHovYylNAIJOGnGRXc/edit ) proposal as we proceed with the project. I think for documenting the terminology related to Streams API, i can take reference from [Steams Spec](https://streams.spec.whatwg.org/).

> Other than that, this looks promising. Keep up the nice work!

Thank You :)

timvandermeij:
I did not notice the specification yet, but it's surprisingly good to read and understand. It's a really helpful resource (that I see you already mentioned in your proposal) and I think extracting the most important terminology from it is a good idea. Thank you for considering that; I'll leave it for now (to not digress from the topic of this particular PR) and watch how to proposal progresses.

src/shared/util.js
mukulmishra18: Okay, done in new commit.

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done, thanks for suggestions.

src/shared/util.js
```diff
@@ -1218,20 +1218,164 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+  this.streamId = 1;
   this.postMessageTransfers = true;
-  var callbacksCapabilities = this.callbacksCapabilities = Object.create(null);
-  var ah = this.actionHandler = Object.create(null);
+  let streamSinks = this.streamSinks = Object.create(null);
+  let streamControllers = this.streamControllers = Object.create(null);
+  let callbacksCapabilities = this.callbacksCapabilities = Object.create(null);
+  let ah = this.actionHandler = Object.create(null);
+
+  let _processStreamMessage = (data) => {
```
yurydelendik: looks good, but I wonder if we can move _processStreamMessage and _createStreamSink as a method of MessageHandler (we have this.comObj, this.streamSinks and this.streamControllers available to us)

src/shared/util.js
```diff
@@ -1218,20 +1218,164 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+        delete streamControllers[data.streamId];
+      });
+    };
+    switch (data.stream) {
+      case 'start_complete':
+        if (data.success) {
+          streamControllers[data.streamId].startCall.resolve();
+        } else {
+          streamControllers[data.streamId].startCall.reject();
+        }
```
yurydelendik: add utility function `let resolveOrReject = (capability, success, reason) => { if (success) { capability.resolve(); ...` and use it at xxxx_complete branches

src/shared/util.js
```diff
@@ -1218,20 +1218,164 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+        // set ready property as resolved promise.
+        if (streamSinks[data.streamId].desiredSize <= 0 &&
+            data.desiredSize > 0) {
+          streamSinks[data.streamId].sinkCapability.resolve();
+        }
+        // Reset desiredSize property of sink on every pull.
+        streamSinks[data.streamId].desiredSize = data.desiredSize;
+        let pullResult = streamSinks[data.streamId].onPull !== null ?
+          streamSinks[data.streamId].onPull() : undefined;
+        Promise.resolve(pullResult).then(() => {
```
yurydelendik: add `let resolveCall = (fn, args) => { if (!fn) return Promise.resolve(undefined); return new Promise( (resolve, reject) => { resolve(fn.apply(null, args)); }); }` and use it in pull, cancel and start cases.

src/shared/util.js
```diff
@@ -1218,20 +1218,164 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+      case 'enqueue':
+        streamControllers[data.streamId].controller.enqueue(data.chunk);
+        break;
+      case 'close':
+        streamControllers[data.streamId].controller.close();
+        deleteStreamController();
+        break;
+      case 'error':
+        streamControllers[data.streamId].controller.error(data.reason);
+        break;
```
yurydelendik: call deleteStreamController when error

src/shared/util.js
```diff
@@ -1218,20 +1218,164 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+      ready: null,
+    };
+    streamSink.sinkCapability.resolve();
+    streamSink.ready = streamSink.sinkCapability.promise;
+    streamSinks[streamId] = streamSink;
+    let actionResult = action[0].call(action[1], data.data, streamSink);
+    Promise.resolve(actionResult).then(() => {
+      sendStreamRequest({ stream: 'start_complete', success: true });
+    }, () => {
+      sendStreamRequest({ stream: 'start_complete', success: false });
```
yurydelendik: add reason here

src/shared/util.js
```diff
@@ -1218,20 +1218,164 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+  let _createStreamSink = (data) => {
+    let action = ah[data.action];
+    let streamId = data.streamId;
+    let desiredSize = data.desiredSize;
+    let sourceName = this.sourceName;
+    let targetName = data.sourceName;
+    let capability = createPromiseCapability();
+    let sendStreamRequest = ({ stream, chunk, success, reason }) => {
+      comObj.postMessage({ sourceName, targetName, stream, streamId,
+                           chunk, reason, success });
```
yurydelendik: nit: switch 'success' and 'reason' places

src/shared/util.js
```diff
@@ -1218,20 +1218,164 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+    let streamId = data.streamId;
+    let sendStreamResponse = ({ stream, success, reason }) => {
+      comObj.postMessage({ sourceName, targetName, stream,
+                           success, streamId });
+    };
+    let deleteStreamController = () => {
+      // Delete streamController only when start, pull and
+      // cancel callbacks are resolved, to avoid "TypeError".
+      Promise.all([
+        streamControllers[data.streamId].startCall,
```
yurydelendik: add `let finalize = (promise) => { return Promise.resolve(promise).catch(() => {}); };` utility function and use it to wrap streamControllers[data.streamId].xxxCall here

yurydelendik (commented):
To improve readability and error handling

src/shared/util.js
mukulmishra18: I think i am not getting this utility function properly, can you please explain a bit.
We are going to use this function something like: `finalize([streamControllers[data.streamId].startCall, streamControllers[data.streamId].pullCall, streamControllers[data.streamId].cancelCall]).then(() => {...})` ?

src/shared/util.js
mukulmishra18: Can we do this like: `let resolveOrReject = ({ capability, success, reason }) => { if (success) { capability.resolve(); ...`, as i think we don't have reason for `success == true`

src/shared/util.js
mukulmishra18: We require to call `action[0]` with `this = action[1]` in start case [here](https://github.com/mozilla/pdf.js/pull/8430/files#diff-8d2b578d58d4d8c8a2975020fbbc84d9R1360). So maybe this function shall be `let resolveCall = ({ fn, _this, args}) => { if (!fn) return Promise.resolve(undefined); return new Promise( (resolve, reject) => { resolve(fn.apply(_this, args)); }); }` instead.

src/shared/util.js
yurydelendik: it's fine as 3 args, you can add `undefined` for success == true or just don't supply it

src/shared/util.js
yurydelendik: sure, you make make its signature `(fn, args, thisArg)` and for pull/cancel call use thisArg == null.

src/shared/util.js
mukulmishra18: Done in new commit.

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
Snuffleupagus: I might be totally misunderstanding, but I'm interpreting https://github.com/mozilla/pdf.js/pull/8430#discussion_r119618166 as suggesting that they be moved into `MessageHandler.prototype` instead (since updated patch doesn't really change anything otherwise)!?

mukulmishra18: > I might be totally misunderstanding, but I'm interpreting #8430 (comment) as suggesting that they be moved into MessageHandler.prototype instead (since the updated patch doesn't really change anything otherwise)!?

I think i misunderstood [#8439 (comment)](https://github.com/mozilla/pdf.js/pull/8430#discussion_r119618166), sorry for that. I will move `_processStreamMessage` and `_createStreamSink` to `MessageHandler.prototype`

mukulmishra18: I just thought earlier that maybe doing something like: `let _createStreamSink = this._createStreamSink = () => {...}` will make `_createStreamSink` a method of `MessageHandler`.

yurydelendik: it will, but from performance point of view this methods without closure (in this case members of the prototype) will be more effective and easier to debug.

src/shared/util.js
```diff
@@ -1218,20 +1218,172 @@ function MessageHandler(sourceName, targetName, comObj) {
...
+  this.callbackId = 1;
+  this.streamId = 1;
   this.postMessageTransfers = true;
-  var callbacksCapabilities = this.callbacksCapabilities = Object.create(null);
-  var ah = this.actionHandler = Object.create(null);
+  let streamSinks = this.streamSinks = Object.create(null);
+  let streamControllers = this.streamControllers = Object.create(null);
+  let callbacksCapabilities = this.callbacksCapabilities = Object.create(null);
+  let ah = this.actionHandler = Object.create(null);
+  let resolveCall = (fn, args, thisArg = null) => {
```
Snuffleupagus: It doesn't look like this needs to be an arrow function (since there's no `this` reference in the function), so please just do `function resolveCall(fn, args, thisArg = null) {` instead.
Also, for readability, please add a newline before this function definition.

mukulmishra18: Okay, sure. Thanks for suggestion.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+   * @param {String} actionName - Action to call.
+   * @param {JSON} data - JSON data to send.
+   * @param {Object} queueingStrategy - strategy to signal backpressure based on
+   *                 internal queue.
+   * @param {Array} [transfers] - Optional list of transfers/ArrayBuffers.
+   * @return {ReadableStream} ReadableStream to read data in chunks.
+   */
+  sendWithStream(actionName, data, queueingStrategy, transfers) {
+    let streamId = this.streamId++;
+    let that = this;
```
Snuffleupagus: If you're not able to avoid aliasing `this` (by using e.g. arrow functions below), at least use `let self = this;` instead since that's the convention in the code-base.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+    let sourceName = this.sourceName;
+    let targetName = this.targetName;
+    return new ReadableStream({
+      start(controller) {
```
Snuffleupagus: `start: (controller) => {`, and similar below, should work and removes the need for binding `this` manually.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+    if (!fn) {
+      return Promise.resolve(undefined);
+    }
+    return new Promise((resolve, reject) => {
+      resolve(fn.apply(thisArg, args));
+    });
+  },
+
+  _createStreamSink(data) {
+    let that = this;
```
Snuffleupagus: If you're not able to avoid aliasing this (by using e.g. arrow functions below), at least use let self = this; instead since that's the convention in the code-base.

mukulmishra18: Can we use [#8430 (comment)](https://github.com/mozilla/pdf.js/pull/8430#discussion_r119932914), here also?
As i am using `that` inside `streamSink` object, and we can use something like `close: () => {...}`

Snuffleupagus: Maybe not, you'll have to check, but please use the name `self` instead.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+   * @param {Object} queueingStrategy - strategy to signal backpressure based on
+   *                 internal queue.
+   * @param {Array} [transfers] - Optional list of transfers/ArrayBuffers.
+   * @return {ReadableStream} ReadableStream to read data in chunks.
+   */
+  sendWithStream(actionName, data, queueingStrategy, transfers) {
+    let streamId = this.streamId++;
+    let that = this;
+    let sourceName = this.sourceName;
+    let targetName = this.targetName;
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+          sourceName,
+          targetName,
+          action: actionName,
+          streamId,
+          data,
+          desiredSize: controller.desiredSize
+        });
+        // Return Promise for Async process, to signal success/failure.
+        return startCapability.promise;
+      },
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+          sourceName,
+          targetName,
+          stream: 'pull',
+          streamId,
+          desiredSize: controller.desiredSize
+        });
+        // Returning Promise will not call "pull"
+        // again until current pull is resolved.
+        return pullCapability.promise;
+      },
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+  },
+
+  _createStreamSink(data) {
+    let that = this;
+    let action = this.actionHandler[data.action];
+    let streamId = data.streamId;
+    let desiredSize = data.desiredSize;
+    let sourceName = this.sourceName;
+    let targetName = data.sourceName;
+    let capability = createPromiseCapability();
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+    let sendStreamRequest = ({ stream, chunk, success, reason }) => {
+      this.comObj.postMessage({ sourceName, targetName, stream, streamId,
+                                chunk, success, reason });
+    };
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+        this.desiredSize -= size;
+        // Enqueue decreases the desiredSize property of sink,
+        // so when it changes from positive to negative,
+        // set ready as unresolved promise.
+        if (lastDesiredSize > 0 && this.desiredSize <= 0) {
+          this.sinkCapability = createPromiseCapability();
+          this.ready = this.sinkCapability.promise;
+        }
+        sendStreamRequest({ stream: 'enqueue', chunk });
+      },
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+      close() {
+        sendStreamRequest({ stream: 'close' });
+        delete that.streamSinks[streamId];
+      },
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+      error(reason) {
+        sendStreamRequest({ stream: 'error', reason });
+      },
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+      sendStreamRequest({ stream: 'start_complete', success: true });
+    }, (reason) => {
+      sendStreamRequest({ stream: 'start_complete', success: false, reason });
+    });
+  },
+
+  _processStreamMessage(data) {
+    let sourceName = this.sourceName;
+    let targetName = data.sourceName;
+    let streamId = data.streamId;
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+    let sendStreamResponse = ({ stream, success, reason }) => {
+      this.comObj.postMessage({ sourceName, targetName, stream,
+                                success, streamId, reason });
+    };
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+    let resolveOrReject = (capability, success, reason) => {
+      if (success) {
+        capability.resolve();
+      } else {
+        capability.reject(reason);
+      }
+    };
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+    let finalize = (promise) => {
+      return Promise.resolve(promise).catch(() => {});
+    };
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+
+  _processStreamMessage(data) {
+    let sourceName = this.sourceName;
+    let targetName = data.sourceName;
+    let streamId = data.streamId;
+    let sendStreamResponse = ({ stream, success, reason }) => {
+      this.comObj.postMessage({ sourceName, targetName, stream,
+                                success, streamId, reason });
+    };
+    let resolveOrReject = (capability, success, reason) => {
```
Snuffleupagus: You can do `function resolveOrReject(capability, success, reason) {` instead here.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+      if (success) {
+        capability.resolve();
+      } else {
+        capability.reject(reason);
+      }
+    };
+    let finalize = (promise) => {
```
Snuffleupagus: You can do `function finalize(promise) {` instead here.

src/shared/util.js
```diff
@@ -1329,9 +1336,221 @@ MessageHandler.prototype = {
...
+      },
+      error(reason) {
+        sendStreamRequest({ stream: 'error', reason });
+      },
+      sinkCapability: capability,
+      onPull: null,
+      onCancel: null,
+      desiredSize,
+      ready: null,
+    };
```
Snuffleupagus: Nit: Adding a newline here would aid readability of the code.

src/shared/util.js
mukulmishra18: Yes, you are right. We can't use that logic. I tried and it is not working.
Okay, i will change it to `self`.

src/shared/util.js
mukulmishra18: Done in new commit. Thanks.

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done, thanks for the suggestion. I will take care of that.

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
mukulmishra18: Done

src/shared/util.js
```diff
@@ -1329,9 +1336,233 @@ MessageHandler.prototype = {
...
+          reason,
+          streamId
+        });
+        // Return Promise to signal success or failure.
+        return cancelCapability.promise;
+      }
+    }, queueingStrategy);
+  },
+
+  resolveCall(fn, args, thisArg = null) {
```
yurydelendik: resolveCall does not need to be a member of MessageHandler. Can you move it out of the class (e.g. just before MessageHandler)?

yurydelendik: resolveOrReject and finalize can go there too

yurydelendik (approved):
looks good (with resolveOrReject, finalize and resolveCall before MessageHandler constructor)

src/shared/util.js
mukulmishra18: Done in new commit. Thanks.

src/shared/util.js
```diff
@@ -1214,24 +1214,50 @@ var createObjectURL = (function createObjectURLClosure() {
...
+
+function resolveOrReject(capability, success, reason) {
+  if (success) {
+    capability.resolve();
+  } else {
+    capability.reject(reason);
+  }
+}
+
+function finalize(promise) {
```
Snuffleupagus: Unless I'm missing something, you're *not* actually passing `Promise`s to this function (at line https://github.com/mozilla/pdf.js/pull/8430/files#diff-8d2b578d58d4d8c8a2975020fbbc84d9R1492 and below).

Rather, it seems that what you're passing is actually `promiseCapabilities`, which means that this function currently doesn't really make sense to me; what am I missing here?

src/shared/util.js
```diff
@@ -1329,9 +1357,212 @@ MessageHandler.prototype = {
...
+    let sendStreamResponse = ({ stream, success, reason, }) => {
+      this.comObj.postMessage({ sourceName, targetName, stream,
+                                success, streamId, reason, });
+    };
+
+    let deleteStreamController = () => {
+      // Delete streamController only when start, pull and
+      // cancel callbacks are resolved, to avoid "TypeError".
+      Promise.all([
+        finalize(this.streamControllers[data.streamId].startCall),
```
yurydelendik: here and below must be  `finalize(this.streamControllers[data.streamId].startCall.promise),`

yurydelendik: sorry. let's map all capabilities to promises `Promise.all([this.streamControllers[data.streamId].startCall, this.streamControllers[data.streamId].pullCall,....].map((capability) => { return capability && finalize(capability.promise); }))`.

src/shared/util.js
mukulmishra18: Oh, yes. You are right. Thanks for pointing out.

src/shared/util.js
mukulmishra18: Done in new commit. Thanks.

Snuffleupagus (approved):
r=me, modulo passing tests.

yurydelendik:
Thank you for the patch.

PR #8431, opened 2017-05-19
Replace a couple of leftover `make.js` references with `gulpfile.js` in docs/comments

Snuffleupagus:
Also updates a `console.log` statement in the `gulp importl10n` command (since I forgot it in a previous patch, and it didn't seem necessary with a separate patch for it).


timvandermeij (approved):

timvandermeij:
Thanks!

PR #8433, opened 2017-05-22
Remove the reference to `compatibility.js` from the README, since it's included by default in `GENERIC` builds

Snuffleupagus:
Please note that the FAQ (in the Wiki) was already updated to mention that the standalone `compatibility.js` file isn't necessary anymore.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #8441, opened 2017-05-24
Ensure that `TilingPattern`s have valid (non-zero) /BBox arrays (issue 8330)

### LINKED ISSUE #8330: Part of pdf is not rendering [Invalid canvas size error in console]
Link to PDF file (or attach file here
[FailingPDF.pdf](https://github.com/mozilla/pdf.js/files/954710/FailingPDF.pdf)

Configuration:
- Chrome (57.0.2987.133)
- MAC OS (10.10.5)
- PDF.js version:1.8.252
- Is an extension: No

1. Kindly upload the file attached above in the https://mozilla.github.io/pdf.js/web/viewer.html
2. When the first page is rendered it throw an error on the console telling invalid canvas size

A part of the pdf present in the first page was not loaded, and the loading is stuck.
Below is the error, and the loading sign on the pdf first page
error
loadingsign

Ideally it should have loaded the part of the pdf.

Snuffleupagus:
Fixes #8330.

*Please note:* This is a tentative patch, since I'm not entirely certain if it's the best solution; but it does work!

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -480,30 +480,40 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                                    stateManager.state.clone());
     },
 
-    handleTilingType:
-        function PartialEvaluator_handleTilingType(fn, args, resources,
-                                                   pattern, patternDict,
-                                                   operatorList, task) {
+    handleTilingType(fn, args, resources, pattern, patternDict, operatorList,
+                     task) {
       // Create an IR of the pattern code.
-      var tilingOpList = new OperatorList();
+      let tilingOpList = new OperatorList();
       // Merge the available resources, to prevent issues when the patternDict
       // is missing some /Resources entries (fixes issue6541.pdf).
-      var resourcesArray = [patternDict.get('Resources'), resources];
-      var patternResources = Dict.merge(this.xref, resourcesArray);
+      let resourcesArray = [patternDict.get('Resources'), resources];
+      let patternResources = Dict.merge(this.xref, resourcesArray);
 
       return this.getOperatorList({
         stream: pattern,
         task,
         resources: patternResources,
         operatorList: tilingOpList,
-      }).then(function () {
-        // Add the dependencies to the parent operator list so they are
-        // resolved before sub operator list is executed synchronously.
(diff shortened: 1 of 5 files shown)


Snuffleupagus:
*Checking if the test "failures" in Chrome on Windows is intermittent.*

/botio-windows test

src/core/evaluator.js
```diff
@@ -480,30 +480,32 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
-        }, patternDict, args));
+          argsArray: tilingOpList.argsArray,
+        }, patternDict, args);
+      }).then(function(tilingPatternIR) {
+        // Add the dependencies to the parent operator list so they are resolved
+        // before the sub operator list is executed synchronously.
+        operatorList.addDependencies(tilingOpList.dependencies);
+        operatorList.addOp(fn, tilingPatternIR);
+      }, function(reason) {
+        warn(`handleTilingType - ignoring pattern: "${reason}".`);
```
Snuffleupagus: I'm not sure if we should send an `UnsupportedFeature` notification here, or if it doesn't matter!?

src/core/pattern.js
```diff
@@ -795,12 +795,17 @@ Shadings.Dummy = (function DummyClosure() {
...
+  let matrix = dict.getArray('Matrix');
+  let bbox = Util.normalizeRect(dict.getArray('BBox'));
+  let xstep = dict.get('XStep');
+  let ystep = dict.get('YStep');
+  let paintType = dict.get('PaintType');
+  let tilingType = dict.get('TilingType');
+
+  // Ensure that the pattern has a non-zero width and height, to prevent errors
+  // in `pattern_helper.js` (fixes issue8330.pdf).
+  assert((bbox[2] - bbox[0]) !== 0 && (bbox[3] - bbox[1]) !== 0,
```
yurydelendik: (it's pre-existing practice but) let's change this to `if..throw new Error`. I'll will be opening issue to re-inspect assert usage (see #8506)

yurydelendik (approved):
Looks good with change below.

src/core/pattern.js
Snuffleupagus: Sure, I've updated the patch; thank you for the review!

PR #8448, opened 2017-05-25
[don't merge] Checking if --disable-hang-monitor helps

yurydelendik:


timvandermeij:
It looks like Windows and Chrome might work together again. It's strange that Linux and Chrome have an intermittent issue, but if we can switch Chrome testing to Windows (and open a tracking issue for the Linux bot), then I think we can at least continue for now.

PR #8119, opened 2017-02-28
Convert the overlay manager to a class

timvandermeij:
This work makes the code similar to the other components and reduces the number of dependencies for those components because the overlay manager is no longer a global.

I kept the `FIXME` comments equal to the already existing ones as the way we amend configuration objects should be improved in a follow-up. This way it is at least clear which parts should be refactored.

*Much easier to review with https://github.com/mozilla/pdf.js/pull/8119/files?w=1.*

PR #6233, opened 2015-07-18
Improve support for viewing PDFs from file://-URLs in the Chrome extension

Rob--W:
Up to now, if a user did not check the "Allow access to file URLs" checkbox at the extension settings page, then they could not view local PDFs in Chromium.

With this patch, such navigations are intercepted, and the extension will ask the user to enable the setting when it detects that it has no access to file:-URLs.

When the user selects "Allow access to file URLs", the extension reloads and Chrome closes all existing extension tabs (https://crbug.com/511670). Since this is disruptive, I've added a work-around (save tab state before page unload, restore tab state when extension loads).

To test:
1. node make chromium
2. chromium --load-extension=build/chromium
3. Visit chrome://extensions and uncheck the 'Allow access to file URLs" at the PDF.js extension (this was automatically checked because you've loaded the extension in unpacked mode).
4. Open a local PDF in the browser.
5. Verify that the PDF.js extension prompts for file access permissions.
6. Follow the instructions:
   - Click on the chrome://extensions link.
   - Click on the "Allow access to file URLs" checkbox at the extensions page.
7. Navigate back
8. Verify that the local PDF is displayed by the extension.<br><br>
9. Go to chrome://extensions in a new tab (so the PDF viewer is still open).
10. Click on "Allow access to file URLs" (to untick the checkbox).
11. Observe that Chrome closes the PDF viewer tab, but verify that the PDF Viewer extension re-opens the tab (and prompts to re-enable file access).

Screenshot:

Chrome extension asking for access to local files

The displayed string adapts itself to the browser. Start Chrome/Chromium with the `LANG=nl_NL.UTF8` environment variable, and you will see a Dutch browser interface and the Dutch translation of "Allow access to file URLs":

Chrome extension asking for access to local files (LANG=nl_NL.UTF8)

PR #8450, opened 2017-05-26
Convert the overlay manager to ES6 syntax

timvandermeij:
This pull request is a redo of #8119.

You can test that the `web/chromecom.js` change works using the steps from #6233. The other overlays (document properties, password prompt and print) can be tested using the viewer itself. For convenience, you can use http://async5.org/moz/passwordOU.pdf (password: 123456) to test the password prompt.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -119,6 +119,8 @@ var PDFViewerApplication = {
   store: null,
   /** @type {DownloadManager} */
   downloadManager: null,
+  /** @type {OverlayManager} */
+  overlayManager: null,
   /** @type {Preferences} */
   preferences: null,
   /** @type {Toolbar} */
@@ -261,6 +263,8 @@ var PDFViewerApplication = {
     let appConfig = this.appConfig;
 
     return new Promise((resolve, reject) => {
+      this.overlayManager = new OverlayManager();
+
       let eventBus = appConfig.eventBus || getGlobalEventBus();
       this.eventBus = eventBus;
 
@@ -329,16 +333,15 @@ var PDFViewerApplication = {
 
       this.pdfViewer.setFindController(this.findController);
 
-      // FIXME better PDFFindBar constructor parameters
+      // TODO: improve `PDFFindBar` constructor parameter passing
       let findBarConfig = Object.create(appConfig.findBar);
       findBarConfig.findController = this.findController;
       findBarConfig.eventBus = eventBus;
       this.findBar = new PDFFindBar(findBarConfig);
 
-      this.overlayManager = OverlayManager;
-
       this.pdfDocumentProperties =
-        new PDFDocumentProperties(appConfig.documentProperties);
+        new PDFDocumentProperties(appConfig.documentProperties,
+                                  this.overlayManager);
 
       this.pdfCursorTools = new PDFCursorTools({
         container,
@@ -361,7 +364,8 @@ var PDFViewerApplication = {
         });
       }
 
-      this.passwordPrompt = new PasswordPrompt(appConfig.passwordOverlay);
+      this.passwordPrompt = new PasswordPrompt(appConfig.passwordOverlay,
+                                               this.overlayManager);
 
       this.pdfOutlineViewer = new PDFOutlineViewer({
         container: appConfig.sidebar.outlineView,
@@ -375,7 +379,7 @@ var PDFViewerApplication = {
         downloadManager,
       });
 
-      // FIXME better PDFSidebar constructor parameters
+      // TODO: improve `PDFSidebar` constructor parameter passing
       let sidebarConfig = Object.create(appConfig.sidebar);
       sidebarConfig.pdfViewer = this.pdfViewer;
       sidebarConfig.pdfThumbnailViewer = this.pdfThumbnailViewer;
@@ -1911,7 +1915,7 @@ function webViewerClick(evt) {
 }
 
 function webViewerKeyDown(evt) {
-  if (OverlayManager.active) {
+  if (PDFViewerApplication.overlayManager.active) {
     return;
   }
 
--- a/web/chromecom.js
+++ b/web/chromecom.js
@@ -17,7 +17,6 @@
 import { DefaultExternalServices, PDFViewerApplication } from './app';
 import { BasePreferences } from './preferences';
 import { DownloadManager } from './download_manager';
-import { OverlayManager } from './overlay_manager';
 import { PDFJS } from './pdfjs';
 
 if (typeof PDFJSDev === 'undefined' || !PDFJSDev.test('CHROME')) {
@@ -56,10 +55,11 @@ ChromeCom.request = function ChromeCom_request(action, data, callback) {
 /**
  * Resolves a PDF file path and attempts to detects length.
  *
- * @param {String} file Absolute URL of PDF file.
- * @param {Function} callback A callback with resolved URL and file length.
+ * @param {String} file - Absolute URL of PDF file.
+ * @param {OverlayManager} overlayManager - Manager for the viewer overlays.
+ * @param {Function} callback - A callback with resolved URL and file length.
  */
-ChromeCom.resolvePDFFile = function ChromeCom_resolvePDFFile(file, callback) {
+ChromeCom.resolvePDFFile = function(file, overlayManager, callback) {
   // Expand drive:-URLs to filesystem URLs (Chrome OS)
   file = file.replace(/^drive:/i,
       'filesystem:' + location.origin + '/external/');
@@ -110,7 +110,7 @@ ChromeCom.resolvePDFFile = function ChromeCom_resolvePDFFile(file, callback) {
         if (isAllowedAccess) {
           callback(file);
         } else {
-          requestAccessToLocalFile(file);
+          requestAccessToLocalFile(file, overlayManager);
         }
       });
     });
@@ -155,7 +155,7 @@ function reloadIfRuntimeIsUnavailable() {
 }
 
 var chromeFileAccessOverlayPromise;
-function requestAccessToLocalFile(fileUrl) {
+function requestAccessToLocalFile(fileUrl, overlayManager) {
   var onCloseOverlay = null;
   if (top !== window) {
     // When the extension reloads after receiving new permissions, the pages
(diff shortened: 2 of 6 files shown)


web/app.js
```diff
@@ -329,16 +332,18 @@ var PDFViewerApplication = {
...
 
-      this.overlayManager = OverlayManager;
-
+      // TODO: improve `PDFDocumentProperties` constructor parameter passing
+      let documentPropertiesConfig =
+        Object.create(appConfig.documentProperties);
+      documentPropertiesConfig.overlayManager = this.overlayManager;
       this.pdfDocumentProperties =
-        new PDFDocumentProperties(appConfig.documentProperties);
+        new PDFDocumentProperties(documentPropertiesConfig);
```
Snuffleupagus: What you could perhaps do instead (here and elsewhere in the patch), to avoid adding even more of these hacks, is something similar to what PR #8394 does; see https://github.com/mozilla/pdf.js/pull/8394/files#diff-529d1853ee1bba753a0fcb40ea778723R372.

web/app.js
```diff
@@ -119,6 +119,9 @@ var PDFViewerApplication = {
   store: null,
   /** @type {DownloadManager} */
   downloadManager: null,
+  /** @type {OverlayManager} */
+  // TODO: move initialization to `_initializeViewerComponents` if possible
```
Snuffleupagus: Nit: missing a period at the end.

web/chromecom.js
```diff
@@ -56,10 +55,12 @@ ChromeCom.request = function ChromeCom_request(action, data, callback) {
...
  *
- * @param {String} file Absolute URL of PDF file.
- * @param {Function} callback A callback with resolved URL and file length.
+ * @param {String} file - Absolute URL of PDF file.
+ * @param {OverlayManager} overlayManager - Manager for the viewer overlays.
+ * @param {Function} callback - A callback with resolved URL and file length.
  */
-ChromeCom.resolvePDFFile = function ChromeCom_resolvePDFFile(file, callback) {
+ChromeCom.resolvePDFFile =
+    function ChromeCom_resolvePDFFile(file, overlayManager, callback) {
```
Snuffleupagus: This fits on one line if you just use:
`ChromeCom.resolvePDFFile = function(file, overlayManager, callback) {`

web/chromecom.js
```diff
@@ -338,8 +339,9 @@ class ChromePreferences extends BasePreferences {
 
 var ChromeExternalServices = Object.create(DefaultExternalServices);
 ChromeExternalServices.initPassiveLoading = function (callbacks) {
-  var appConfig = PDFViewerApplication.appConfig;
-  ChromeCom.resolvePDFFile(appConfig.defaultUrl,
+  let appConfig = PDFViewerApplication.appConfig;
+  let overlayManager = PDFViewerApplication.overlayManager;
```
Snuffleupagus: This can be `let { appConfig, overlayManager, } = PDFViewerApplication;` instead.

Snuffleupagus (commented):
Note that I've not actually tested this yet, but generally it looks good!
However, I'm adding a couple of smaller comments based on a quick look at the code.

web/overlay_manager.js
```diff
@@ -13,132 +13,139 @@
  * limitations under the License.
  */
 
-var OverlayManager = {
-  overlays: {},
-  active: null,
+const ESC_KEYCODE = 27;
```
Snuffleupagus: I'm not sure that adding this as a constant really brings much value; it's not like the keycode of `Esc` is going to change!?

web/app.js
timvandermeij: Done in the new commit.

web/app.js
timvandermeij: Good idea! Done in the new commit.

web/chromecom.js
timvandermeij: Done in the new commit.

web/chromecom.js
timvandermeij: Done in the new commit.

web/overlay_manager.js
timvandermeij: I think that's exactly the reason why it should be a constant, both for readability and because it does not change. However, I can see that that is a personal choice, so I reverted the change in the new commit.

Snuffleupagus:
> I left a TODO for the initialization of the overlay manager. We cannot move that (yet) because the viewer is usually not yet initialized when e.g., web/chromecom.js needs it.

Looking at the code, I'm cannot help wondering if this problem wasn't limited to the prior PR, but actually isn't an issue any more in *this* PR!?
In this patch, you're now accessing `PDFViewerApplication.overlayManager` from inside of the `externalServices.initPassiveLoading` function. So tracing through the code pertaining to the `CHROME` build target, this is what I see:
 1. `externalServices.initPassiveLoading` is called from https://github.com/mozilla/pdf.js/blob/master/web/app.js#L484, within the `PDFViewerApplication.initPassiveLoading` method (see https://github.com/mozilla/pdf.js/blob/master/web/app.js#L481).
 2. `PDFViewerApplication.initPassiveLoading` is called from https://github.com/mozilla/pdf.js/blob/master/web/app.js#L1523, which is inside of the `webViewerOpenFileViaURL` function (see https://github.com/mozilla/pdf.js/blob/master/web/app.js#L1493).
 3. `webViewerOpenFileViaURL` is called from https://github.com/mozilla/pdf.js/blob/master/web/app.js#L1486, which is the *last* thing that `webViewerInitialized` does (see https://github.com/mozilla/pdf.js/blob/master/web/app.js#L1344).
 4. `webViewerInitialized` is called from https://github.com/mozilla/pdf.js/blob/master/web/app.js#L392, once the `PDFViewerApplication.initialize` method has *resolved*.
 5. `PDFViewerApplication.initialize` (see https://github.com/mozilla/pdf.js/blob/master/web/app.js#L149), will resolve *only* once `PDFViewerApplication._initializeViewerComponents` has successfully been run, see https://github.com/mozilla/pdf.js/blob/master/web/app.js#L156.
 6. Finally, note how `ChromeExternalServices.initPassiveLoading` accesses `PDFViewerApplication.appConfig` in https://github.com/mozilla/pdf.js/blob/master/web/chromecom.js#L341. Unless the current code is broken, we can thus at the *very least* be sure that regardless of all of the above, we've reached line https://github.com/mozilla/pdf.js/blob/master/web/app.js#L153 when `initPassiveLoading` is called.

Unless I've made some sort of fundamental error in the analysis above, it should be possible to place the `this.overlayManager = new OverlayManager();` call after line https://github.com/mozilla/pdf.js/blob/ed661e70631571047b753bfef5c12c84d3476533/web/app.js#L153 instead.

timvandermeij:
You're absolutely right! Everything works as expected with the changes you described. Here is the interdiff of the change:

```diff
diff --git a/web/app.js b/web/app.js
index 58653bfe..0b99c732 100644
--- a/web/app.js
+++ b/web/app.js
@@ -120,8 +120,7 @@ var PDFViewerApplication = {
   /** @type {DownloadManager} */
   downloadManager: null,
   /** @type {OverlayManager} */
-  // TODO: Move initialization to `_initializeViewerComponents` when possible.
-  overlayManager: new OverlayManager(),
+  overlayManager: null,
   /** @type {Preferences} */
   preferences: null,
   /** @type {Toolbar} */
@@ -264,6 +263,8 @@ var PDFViewerApplication = {
     let appConfig = this.appConfig;
 
     return new Promise((resolve, reject) => {
+      this.overlayManager = new OverlayManager();
+
       let eventBus = appConfig.eventBus || getGlobalEventBus();
       this.eventBus = eventBus;
 
diff --git a/web/pdf_print_service.js b/web/pdf_print_service.js
index 3b44da12..dde9d447 100644
--- a/web/pdf_print_service.js
+++ b/web/pdf_print_service.js
@@ -18,7 +18,7 @@ import { PDFPrintServiceFactory, PDFViewerApplication } from './app';
 import { PDFJS } from './pdfjs';
 
 let activeService = null;
-let overlayManager = PDFViewerApplication.overlayManager;
+let overlayManager = null;
 
 // Renders the page to the canvas of the given print service, and returns
 // the suggested dimensions of the output page.
@@ -310,6 +310,7 @@ if ('onbeforeprint' in window) {
 var overlayPromise;
 function ensureOverlay() {
   if (!overlayPromise) {
+    overlayManager = PDFViewerApplication.overlayManager;
     overlayPromise = overlayManager.register('printServiceOverlay',
       document.getElementById('printServiceOverlay'), abort, true);
     document.getElementById('printCancel').onclick = abort;
```

web/pdf_print_service.js
```diff
@@ -310,7 +310,8 @@ if ('onbeforeprint' in window) {
 var overlayPromise;
 function ensureOverlay() {
   if (!overlayPromise) {
-    overlayPromise = OverlayManager.register('printServiceOverlay',
+    overlayManager = PDFViewerApplication.overlayManager;
```
Snuffleupagus: Given that this file binds itself to `window.print`, I'm wondering if there's any way that a user might be able to trigger printing before `PDFViewerApplication` is completely initialized?

I've not been successful in triggering this myself; but we might want to add an additional check here just for good measure, to avoid an uncaught `Error` if the `overlayManager` isn't defined yet!? Something like this, perhaps:
```javascript
if (!overlayManager) {
  throw new Error('OverlayManager not initialized.'); // Might need a better error message...
}  
```

web/pdf_print_service.js
```diff
@@ -217,8 +217,8 @@ window.print = function print() {
   } finally {
     if (!activeService) {
       console.error('Expected print service to be initialized.');
-      if (OverlayManager.active === 'printServiceOverlay') {
-        OverlayManager.close('printServiceOverlay');
+      if (overlayManager.active === 'printServiceOverlay') {
```
Snuffleupagus: In line with comment below, and the existing code in `PDFPrintService.destroy`, we might want change this to the follow instead (to avoid an unlikely uncaught `Error`):
```javascript
ensureOverlay().then(function () {
  if (overlayManager.active !== 'printServiceOverlay') {
    return; // overlay was already closed
  }
  overlayManager.close('printServiceOverlay');
});
```
*Note:* I'd actually consider this a pre-existing issue with the code!

Snuffleupagus (approved):
I've tested all the current overlays (document properties, password prompt, printing, Chrome local file access), and everything still works as it should :-)

However, looking through `pdf_print_service.js` one last time, there's a couple of cases where we may want to pro-actively avoid possible, but *very* unlikely, errors.

Aside from that, this looks *great*; thank you for persisting with the class conversion of the OverlayManager!

web/pdf_print_service.js
timvandermeij: I have not been able to trigger that as well, but for safety I added this check in the new commit.

web/pdf_print_service.js
timvandermeij: Indeed, `ensureOverlay` should have been there. I added it in the new commit.

Issue #8451, opened 2017-05-26
pdfjs-dist/webpack.js and IE9 support


FranckFreiburger:
I successfully tested [pdfjs-dist/webpack.js](https://github.com/mozilla/pdfjs-dist/blob/master/webpack.js) for bundling the pdf worker properly.
Everything works fine except for IE9 (used with compatibility.js).
Since IE9 has no Worker class, [worker-loader](https://github.com/webpack-contrib/worker-loader/blob/5225c54e56e77e5ba50f2d983a5bf725d91137bf/index.js#L15) cannot be used properly.
Is it possible to set `PDFJS.workerPort` according to the `Worker` class availability ? 
eg.
```
if ( typeof window !== 'undefined' && 'Worker' in window ) {

	pdfjs.PDFJS.workerPort = new PdfjsWorker();
}
```

yurydelendik:
There is a little we can do for IE9. Please don't use pdfjs-dist/webpack in this case and continue using setting of PDFJS.workerSrc method. Closing as won't fix.

yurydelendik:
How is the worker-loader is used with IE9 anyway?

FranckFreiburger:
I think that worker-loader don't care about browsers that do not have the `Worker` class, it just generates the sting `new Worker(${workerPublicPath})`.

yurydelendik:
> I think that worker-loader don't care about browsers that do not have the Worker class,

Correct. We always promoted PDFJS.workerSrc method since it has proper non-Worker fallback.  pdfjs-dist/webpack is added to people who use worker-loader.

FranckFreiburger:
Since I already use webpack for my project, I would prefer to also use worker-loader, but except for IE9.
The current setup works fine:
```
var PDFJS = require('pdfjs-dist');

var PdfjsWorker = require('worker-loader!pdfjs-dist/build/pdf.worker.js');
if ( typeof window !== 'undefined' && 'Worker' in window )
	PDFJS.PDFJS.workerPort = new PdfjsWorker();
```

On IE9 console, I get `Warning: Setting up fake worker.` (what is expected), and on other browsers, no fake worker and pdf are loading properly.

yurydelendik:
This might work. Add else branch with `PDFJS.PDFJS.disableWorker=true;` and submit as a PR.

FranckFreiburger:
ok, I will create the PR.
[...]

PR #8452, opened 2017-05-26
Update webpack.js to support browsers that do not have the Web Workers API

FranckFreiburger:
Allow to use `pdfjs-dist/webpack.js` even if the final bundle is used by browsers that do not support the Web Workers API (eg. IE9).
See issue #8451

### DIFF
--- a/external/dist/webpack.js
+++ b/external/dist/webpack.js
@@ -16,6 +16,11 @@
 
 var pdfjs = require('./build/pdf.js');
 var PdfjsWorker = require('worker-loader!./build/pdf.worker.js');
-pdfjs.PDFJS.workerPort = new PdfjsWorker();
+
+if (typeof window !== 'undefined' && 'Worker' in window) {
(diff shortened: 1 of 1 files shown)


external/dist/webpack.js
```diff
@@ -16,6 +16,11 @@
 
 var pdfjs = require('./build/pdf.js');
 var PdfjsWorker = require('worker-loader!./build/pdf.worker.js');
-pdfjs.PDFJS.workerPort = new PdfjsWorker();
+
+if ( typeof window !== 'undefined' && 'Worker' in window ) {
```
yurydelendik: spaces after `(` and before `)` needs to be removed. please also amend/squash commits.

timvandermeij:
Looks good. Could you squash the commits (https://github.com/mozilla/pdf.js/wiki/Squashing-Commits)?

timvandermeij (approved):

timvandermeij:
Thank you!

PR #8455, opened 2017-05-28
Convert the page view to ES6 syntax

timvandermeij:
*Easier reviewing with https://github.com/mozilla/pdf.js/pull/8455/files?w=1.*

### DIFF
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -24,7 +24,7 @@ import {
 import { getGlobalEventBus } from './dom_events';
 import { RenderingStates } from './pdf_rendering_queue';
 
-var TEXT_LAYER_RENDER_DELAY = 200; // ms
+const TEXT_LAYER_RENDER_DELAY = 200; // ms
 
 /**
  * @typedef {Object} PDFPageViewOptions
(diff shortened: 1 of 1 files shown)


web/pdf_page_view.js
```diff
@@ -90,125 +78,143 @@ var PDFPageView = (function PDFPageViewClosure() {
...
+    div.appendChild(this.loadingIconDiv);
+  }
 
-      this.loadingIconDiv = document.createElement('div');
-      this.loadingIconDiv.className = 'loadingIcon';
-      div.appendChild(this.loadingIconDiv);
-    },
+  update(scale, rotation) {
+    this.scale = scale || this.scale;
+    this.rotation = rotation || this.rotation;
```
Snuffleupagus: This isn't correct, since it will effectively prevent setting `this.rotation` to zero, if the rotation has previously been set to a non-zero value; please revert this change!

web/pdf_page_view.js
```diff
@@ -44,41 +44,29 @@ var TEXT_LAYER_RENDER_DELAY = 200; // ms
...
-    var renderingQueue = options.renderingQueue;
-    var textLayerFactory = options.textLayerFactory;
-    var annotationLayerFactory = options.annotationLayerFactory;
-    var enhanceTextSelection = options.enhanceTextSelection || false;
-    var renderInteractiveForms = options.renderInteractiveForms || false;
-
-    this.id = id;
-    this.renderingId = 'page' + id;
-    this.pageLabel = null;
+  constructor(options) {
```
Snuffleupagus: A more general question here: Should the signature be changed to `constructor({ ... }) {` instead, or would that be overly verbose considering the amount of parameters involved?

Snuffleupagus (changes requested):
A couple of quick comments thus far; but setting r- for now, just so that we don't accidentally land this while rotation is broken.

web/pdf_page_view.js
timvandermeij: I chose to not do that because it would indeed be too verbose. I usually keep a limit at five or six parameters: if we have more, than such an expansion does not provide much more insight in my opinion.

web/pdf_page_view.js
timvandermeij: Good point! I'm not entirely sure how I missed that, but I reverted the change and added a small comment to make it more clear that this structure is intended.

web/pdf_page_view.js
```diff
@@ -44,41 +44,29 @@ var TEXT_LAYER_RENDER_DELAY = 200; // ms
...
  * @implements {IRenderableView}
  */
-var PDFPageView = (function PDFPageViewClosure() {
+class PDFPageView {
   /**
-   * @constructs PDFPageView
    * @param {PDFPageViewOptions} options
    */
-  function PDFPageView(options) {
-    var container = options.container;
```
Snuffleupagus: I think it'd be a little bit clearer if we keep `let container = options.container;` here, considering that otherwise you're handling all the other parameters *except* this one at the top of the constructor.

web/pdf_page_view.js
```diff
@@ -44,41 +44,29 @@ var TEXT_LAYER_RENDER_DELAY = 200; // ms
...
-    var id = options.id;
-    var scale = options.scale;
-    var defaultViewport = options.defaultViewport;
```
Snuffleupagus: Similarily, let's keep `let defaultViewport = options.defaultViewport;` such that we can maintain the slightly simpler `this.pdfPageRotate = defaultViewport.rotation;` below.

web/pdf_page_view.js
```diff
@@ -217,449 +225,421 @@ var PDFPageView = (function PDFPageViewClosure() {
...
+        cont();
+      },
+      cancel() {
+        renderTask.cancel();
+      },
+    };
+
+    let viewport = this.viewport;
+    let canvas = document.createElement('canvas');
+    canvas.id = 'page' + this.id;
```
Snuffleupagus: Nit: Pre-existing issue, which I've been annoyed with for a while but haven't bothered to fix: Can we do `canvas.id = this.renderingId;` instead.

Snuffleupagus (approved):
Good to land with final comments addressed, thanks for the patch!

web/pdf_page_view.js
timvandermeij: Done in the new commit.

web/pdf_page_view.js
timvandermeij: Done in the new commit.

web/pdf_page_view.js
timvandermeij: Done in the new commit.

PR #8458, opened 2017-05-30
Refactor the `getTempCanvas` function in `pdf_thumbnail_view.js` to a factory, in preparation for ES6 conversion of the thumbnail related code

Snuffleupagus:
This patch intends to simplify future ES6 refactoring of the thumbnail code, since the current code isn't going to work well together with proper `Class`es.

*Edit:* Somewhat simpler reviewing with https://github.com/mozilla/pdf.js/pull/8458/files?w=1.

### DIFF
--- a/web/pdf_thumbnail_view.js
+++ b/web/pdf_thumbnail_view.js
@@ -32,35 +32,52 @@ var THUMBNAIL_CANVAS_BORDER_WIDTH = 1; // px
  *   but increases the overall memory usage. The default value is false.
  */
 
+const TempImageFactory = (function TempImageFactoryClosure() {
+  let tempCanvasCache = null;
+
+  return {
+    getCanvas(width, height) {
+      let tempCanvas = tempCanvasCache;
+      if (!tempCanvas) {
+        tempCanvas = document.createElement('canvas');
+        tempCanvasCache = tempCanvas;
+      }
+      tempCanvas.width = width;
+      tempCanvas.height = height;
+
+      // Since this is a temporary canvas, we need to fill it with a white
+      // background ourselves. `_getPageDrawContext` uses CSS rules for this.
+      if (typeof PDFJSDev === 'undefined' ||
+          PDFJSDev.test('MOZCENTRAL || FIREFOX || GENERIC')) {
+        tempCanvas.mozOpaque = true;
+      }
+
+      let ctx = tempCanvas.getContext('2d', { alpha: false, });
+      ctx.save();
+      ctx.fillStyle = 'rgb(255, 255, 255)';
+      ctx.fillRect(0, 0, width, height);
+      ctx.restore();
+      return tempCanvas;
+    },
+
+    destroyCanvas() {
+      let tempCanvas = tempCanvasCache;
+      if (tempCanvas) {
+        // Zeroing the width and height causes Firefox to release graphics
+        // resources immediately, which can greatly reduce memory consumption.
+        tempCanvas.width = 0;
+        tempCanvas.height = 0;
+      }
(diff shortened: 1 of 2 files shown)


web/pdf_thumbnail_view.js
```diff
@@ -32,35 +32,52 @@ var THUMBNAIL_CANVAS_BORDER_WIDTH = 1; // px
  *   but increases the overall memory usage. The default value is false.
  */
 
+const TempImageFactory = (function TempImageFactoryClosure() {
```
timvandermeij: Is there a particular reason to not make this a class like the other factories?

web/pdf_thumbnail_view.js
```diff
@@ -405,11 +423,13 @@ var PDFThumbnailView = (function PDFThumbnailViewClosure() {
     },
   };
 
+  PDFThumbnailView.cleanup = function() {
```
timvandermeij: I'm a bit confused about why we need this construction (and why we even have it at the moment). How I see it is that `PDFThumbnailViewer` manages all `PDFThumbnailView` instances. Wouldn't it be much easier and more logical to move this cleanup logic to `PDFThumbnailViewer` instead? It would go over all `PDFThumbnailView` instances, destroy them and when they are all destroyed, call the factory's `destroyCanvas` method to delete the shared canvas.

timvandermeij (commented):
I have some questions about this, but I like the idea in general. Coincidentally I had been fiddling with this last weekend, but I couldn't figure out why it had to be so complex, so I left it in favor of the overlay manager PR. I'm hoping you can provide the answer for that concern :-)

web/pdf_thumbnail_view.js
Snuffleupagus: Yes, I wanted to keep the actual canvas private, and that cannot be done using classes (at least not as they are specified in ES6, but hopefully that changes in a future EcmaScript version).

Also, please keep in mind that currently all `PDFThumbnailView`s share just *one* temporary canvas, so a class would be overkill here.

web/pdf_thumbnail_view.js
Snuffleupagus: Since the temporary canvas is only used by `PDFThumbnailView`, the `PDFThumbnailViewer` really shouldn't know anything about them (the current code isn't correct in my opinion).

What you're commenting on is a static method, i.e. one that is shared between all `PDFThumbnailView`s, so there's nothing to iterate over here!

web/pdf_thumbnail_view.js
timvandermeij: Fair enough. Let's keep it this way then.

web/pdf_thumbnail_view.js
timvandermeij: You're right. Maybe I'm worried for nothing; it looks a bit strange, but it's because the one canvas is shared among `PDFThumbnailView` instances. If there would be one per instance, it would be a lot simpler, but we don't want that for memory reasons. I'm fine with this then. Sorry for the noise!

Snuffleupagus: No worries; I suppose that I can see why this looks a bit strange on first sight!

With ES6 classes, this particular code becomes slightly less verbose, since static methods are placed among "regular" ones; using the `static cleanup() { ... }` notation.

timvandermeij: Good, I think having that method explicitly defined as static among the other ones would resolve the concern I had here in terms of code readability.

timvandermeij (approved):

timvandermeij:

PR #8459, opened 2017-05-30
Using pre-built code for testing.

yurydelendik:
(... in hope to fix Chrome failures on test servers)

- Unit test will use pre-built worker code when API code is testing
- Ref tests will use pre-built main and worker code

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1055,22 +1055,22 @@ gulp.task('publish', ['generic'], function (done) {
     });
 });
 
-gulp.task('test', function () {
+gulp.task('test', ['generic'], function () {
   return streamqueue({ objectMode: true },
     createTestSource('unit'), createTestSource('browser'));
 });
 
-gulp.task('bottest', function () {
+gulp.task('bottest', ['generic'], function () {
   return streamqueue({ objectMode: true },
     createTestSource('unit'), createTestSource('font'),
     createTestSource('browser (no reftest)'));
 });
 
-gulp.task('browsertest', function () {
+gulp.task('browsertest', ['generic'], function () {
   return createTestSource('browser');
 });
 
-gulp.task('unittest', function () {
+gulp.task('unittest', ['generic'], function () {
   return createTestSource('unit');
 });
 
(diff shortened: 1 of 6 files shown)


gulpfile.js
```diff
@@ -1055,34 +1055,34 @@ gulp.task('publish', ['generic'], function (done) {
...
   return createTestSource('browser');
 });
 
-gulp.task('unittest', function () {
+gulp.task('unittest', ['generic'], function () {
   return createTestSource('unit');
 });
 
-gulp.task('fonttest', function () {
+gulp.task('fonttest', ['generic'], function () {
```
Snuffleupagus: Unless you also update `test/font/jasmine-boot.js`, I don't think this change is necessary!?

yurydelendik: fixed

yurydelendik:
Let's update ref images.

/botio-windows makeref

Snuffleupagus:
Hmm, looking at the end of the output from the Windows bot in either in https://github.com/mozilla/pdf.js/pull/8459#issuecomment-304925517 (see http://54.215.176.217:8877/87ca7b61808c5a3/output.txt) or https://github.com/mozilla/pdf.js/pull/8459#issuecomment-304938909 (see http://54.215.176.217:8877/34da9c465d2c659/output.txt), it seems that not all test actually ran in *Firefox*!

For Firefox, in the first case https://github.com/mozilla/pdf.js/blob/master/test/test_manifest.json#L3409 is the last test in the log; in the second case https://github.com/mozilla/pdf.js/blob/master/test/test_manifest.json#L3395 is the last test.

Are the logs just incomplete, or didn't all tests get run in Firefox on the Windows bot?

yurydelendik:
I think it's just hitting 200kb maxBuffer limit (https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) -- it shall be fine when "no-reference" messages will go away.

Snuffleupagus:
Landed as part of PR #8284 (obviously with the correct author info).

PR #8461, opened 2017-05-30
Update pdf2svg.js

### LINKED ISSUE #8419: pdf2svg segfaults
Steps to reproduce (attached PDF):
[coredump.pdf](https://github.com/mozilla/pdf.js/files/1008487/coredump.pdf)

````
node examples/node/pdf2svg.js /tmp/coredump.pdf
````

Console:
````
# Document Loaded
Number of Pages: 2

# Page 1
Size: 308.88x606.72

Warning: Unimplemented graphic state BM
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
# Page 2
Size: 308.88x606.72

Warning: Unimplemented graphic state BM
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency
Warning: Unimplemented operator dependency

<--- Last few GCs --->

[30883:0x33e2030]    67959 ms: Scavenge 1373.3 (1414.7) -> 1373.3 (1414.7) MB, 0.3 / 0.0 ms  allocation failure incremental marking delaying mark-sweep
[30883:0x33e2030]    67977 ms: Mark-sweep 1373.3 (1414.7) -> 1147.4 (1176.9) MB, 18.8 / 0.0 ms  (+ 15.1 ms in 19 steps since start of marking, biggest step 5.8 ms, walltime since start of marking 4861 ms) last resort 
[30883:0x33e2030]    68032 ms: Mark-sweep 1147.4 (1176.9) -> 1147.1 (1152.4) MB, 54.9 / 0.0 ms  last resort 


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x1c67e52266a1 <JS Object>
    1: DoJoin(aka DoJoin) [native array.js:~97] [pc=0x2d223be6039b](this=0xbcbfd402311 <undefined>,q=0xfd62042e9f9 <JS Array[36]>,r=36,F=0xbcbfd4023b1 <true>,B=0xbcbfd402471 <String[0]: >,A=0xbcbfd402421 <false>)
    2: Join(aka Join) [native array.js:~122] [pc=0x2d223be6dfc2](this=0xbcbfd402311 <undefined>,q=0xfd62042e9f9 <JS Array[36]>,r=36,B=0xbcbfd402471 <String[0]: >,A=0xbcbfd402421 <false...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [node]
 2: 0x12b640c [node]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [node]
 5: v8::internal::Factory::NewRawTwoByteString(int, v8::internal::PretenureFlag) [node]
 6: v8::internal::Runtime_StringBuilderConcat(int, v8::internal::Object**, v8::internal::Isolate*) [node]
 7: 0x2d223bb043a7
Aborted (core dumped)
````

### PR DESCRIPTION
**Pull request by yshahin**
Changing file write to sync solves the problem.
Looks like promise chaining with a file write at the end with out waiting for the write to happen causes the issue.

Fixes #8419

### DIFF
--- a/examples/node/pdf2svg.js
+++ b/examples/node/pdf2svg.js
@@ -21,7 +21,7 @@ var data = new Uint8Array(fs.readFileSync(pdfPath));
 function writeToFile(svgdump, pageNum) {
   var name = getFileNameFromPath(pdfPath);
   if(!fs.existsSync('./svgdump/')) {
-    fs.mkdirSync('./svgdump/')
+    fs.mkdirSync('./svgdump/');
   }
 
   try {
--- a/examples/node/pdf2svg.js
+++ b/examples/node/pdf2svg.js
@@ -20,7 +20,7 @@ var data = new Uint8Array(fs.readFileSync(pdfPath));
 // Dumps svg outputs to a folder called svgdump
 function writeToFile(svgdump, pageNum) {
   var name = getFileNameFromPath(pdfPath);
-  if(!fs.existsSync('./svgdump/')) {
+  if (!fs.existsSync('./svgdump/')) {
     fs.mkdirSync('./svgdump/');
   }
 
--- a/examples/node/pdf2svg.js
+++ b/examples/node/pdf2svg.js
@@ -23,7 +23,13 @@ function writeToFile(svgdump, pageNum) {
   if(!fs.existsSync('./svgdump/')) {
     fs.mkdirSync('./svgdump/')
   }
-  fs.writeFileSync('./svgdump/' + name + "-" + pageNum + '.svg', svgdump, 'utf-8');
+
+  try {
+    fs.writeFileSync('./svgdump/' + name + '-' + pageNum + '.svg', svgdump, 'utf-8');
+    console.log('Page: ' + pageNum);
+  } catch(err) {
+    console.log('Error: ' + err);
+  }
 }
 
 // Get filename from the path
@@ -62,7 +68,7 @@ pdfjsLib.getDocument({
           writeToFile(svgDump, pageNum);
         });
       });
-    })
+    });
   };
 
   for (var i = 1; i <= numPages; i++) {
--- a/examples/node/pdf2svg.js
+++ b/examples/node/pdf2svg.js
@@ -20,18 +20,10 @@ var data = new Uint8Array(fs.readFileSync(pdfPath));
 // Dumps svg outputs to a folder called svgdump
 function writeToFile(svgdump, pageNum) {
   var name = getFileNameFromPath(pdfPath);
-  fs.mkdir('./svgdump/', function(err) {
-    if (!err || err.code === 'EEXIST') {
-      fs.writeFile('./svgdump/' + name + "-" + pageNum + '.svg', svgdump,
-        function(err) {
-          if (err) {
-            console.log('Error: ' + err);
-          } else {
-            console.log('Page: ' + pageNum);
-          }
-        });
-    }
-  });
+  if(!fs.existsSync('./svgdump/')) {
+    fs.mkdirSync('./svgdump/')
+  }
+  fs.writeFileSync('./svgdump/' + name + "-" + pageNum + '.svg', svgdump, 'utf-8');
 }
 
 // Get filename from the path

### CODE REVIEW

**Comment by yurydelendik** [comment]
I wonder if there is a logical explanation to this fix -- node.js people recommend to use async callbacks.

> Looks like promise chaining with a file 

callback chaining

File: examples/node/pdf2svg.js [hunk 1]
```diff
@@ -20,18 +20,10 @@ var data = new Uint8Array(fs.readFileSync(pdfPath));
...
-        function(err) {
-          if (err) {
-            console.log('Error: ' + err);
-          } else {
-            console.log('Page: ' + pageNum);
-          }
-        });
-    }
-  });
+  if(!fs.existsSync('./svgdump/')) {
```
timvandermeij: Add a space after `if`, so that it becomes `if (!fs....`.

examples/node/pdf2svg.js
```diff
@@ -20,18 +20,10 @@ var data = new Uint8Array(fs.readFileSync(pdfPath));
...
+    fs.mkdirSync('./svgdump/')
```
timvandermeij: You need a semicolon at the end of the line.

examples/node/pdf2svg.js
```diff
@@ -20,18 +20,10 @@ var data = new Uint8Array(fs.readFileSync(pdfPath));
...
+  }
+  fs.writeFileSync('./svgdump/' + name + "-" + pageNum + '.svg', svgdump, 'utf-8');
```
timvandermeij: Change the double quotes to single quotes on this line. Moreover, the prints are now gone. Is there a way to add them back?

timvandermeij (changes requested):

timvandermeij:
I'm also wondering about the explanation for this. I'm fine with the sync approach if that works, but I do not understand why the async approach does not work properly.

yshahin:
I think the problem is not with the async calls but probably with when the call gets executed.
The "End Document" print statement gets printed before the write happens, which I think might be the cause.
By blocking the promise till the write happens this solves the problem.
There might be a way to do it with async calls but I did not figure it out yet.

Will continue working on it.

yshahin:
I tried different different things but only the sync file write works. I think we need to consult a node export here to tell us what the issue is. From the previous link it seems like a memory leak problem.

yurydelendik:
> By blocking the promise till the write happens this solves the problem.

I tried adding callback to block a promise -- the same problem.

> I tried different different things but only the sync file write works.

I'm worry it's just coincidence and the patch can be just masking the real problem, which can surface later.

yurydelendik:
Blocked by https://github.com/nodejs/node/issues/13337

yurydelendik:
node.js experts say it is a memory leak, so switching to sync might only fix the issue only for this case and can reappear in the future. I recommend to wait until leak will be found.

yshahin:
@yurydelendik take a look at this https://github.com/petkaantonov/bluebird/issues/1068 it seems like an inherit problem in promises in node.
The solution presented requires bluebird lib, I can try it out and see if it works.

yurydelendik:
@yshahin okay, pdf.js also have polyfill in compatibility.js, see if "forcing" it will help

yurydelendik:
@yashsriv the https://github.com/yurydelendik/nodesegfault/commit/2bb6f26c09309d942901c3d66fca3e6d3fdaf6f4 fixes the issue, so you are right it's due to Promises implementation.

yurydelendik:
Blocked by https://github.com/nodejs/help/issues/652

yshahin:
@yurydelendik how do you want to handle this?
Use the your fix? or change it to sync calls?

timvandermeij:
If it works with our promise polyfill and not with the native Node.js one, I'm not sure why the Node.js people claim the memory leak is on our side. Even if there is a memory leak, it shouldn't matter if you use sync/async writing.

Personally I'm fine with the sync approach here to solve this issue, but I'll let @yurydelendik make the call.

Rob--W:
I'm closing this PR in favor of mine at #8540, because moving from async to sync is unnecessary.

The reason that moving from async to sync "solves" the problem is that it forces Node.js to suspend execution until the file writing finishes. Consequently, pages are effectively processed sequentially.
My patch makes also changes the pdf2svg code to process the pages sequentially, but without blocking execution.

In this thread, it has been said that using our Promise polyfill solves the issue as well. I think that this is a mere coincidence. Normally promises are scheduled as a microtask, but with our [polyfill it is scheduled using setTimeout](https://github.com/mozilla/pdf.js/blob/8e9b4b5ff2be605895904db86be938d86a123016/src/shared/compatibility.js#L897-L967), which causes the promise to resolve less quickly. Consequently, when the SVG generation process (with many "promises") finishes, some file writing requests have likely finished as well, allowing Node.js to reclaim the memory for the string that is passed to `fs.writeFile` ([in pdf2svg.js](https://github.com/mozilla/pdf.js/blob/8e9b4b5ff2be605895904db86be938d86a123016/examples/node/pdf2svg.js#L21-L25)).

PR #8462, opened 2017-05-31
Removes last UMDs from the modules.

yurydelendik:


src/main_loader.js
```diff
@@ -13,60 +13,20 @@
...
-    factory((root.pdfjsMainLoader = {}), root.pdfjsDisplayAPI,
-      root.pdfjsDisplayAnnotationLayer, root.pdfjsDisplayTextLayer,
-      root.pdfjsDisplayDOMUtils, root.pdfjsSharedUtil, root.pdfjsDisplaySVG,
-      root.pdfjsDisplayGlobal);
-  }
-}(this, function (exports, displayAPI, displayAnnotationLayer,
-                  displayTextLayer, displayDOMUtils, sharedUtil,
-                  displaySVG, displayGlobal) {
-
-  // Sync the exports below with ./pdf.js file/template.
```
Snuffleupagus: Unless I'm missing something, shouldn't we should keep this comment?

Snuffleupagus (approved):
Besides the inline question, is there any reason for not updating https://github.com/mozilla/pdf.js/blob/master/src/pdf.js to the same format as is now used in `src/main_loader.js`?

This looks good to me, thank you!

src/main_loader.js
yurydelendik: We can remove main_loader.js

PR #8463, opened 2017-05-31
Allow specifying the `PDFJS_NEXT` build flag via an environment variable when running the various `gulp` commands

Snuffleupagus:
After PR #8459, the run-time of the various `gulp` test commands has regressed quite badly on Windows. For me, `gulp test` now takes approximately *twice* as long when run locally on Windows.
The problem seems to be the Babel transpilation step, which takes well over five minutes to run.[1]

For someone like me, who runs tests a lot locally, this slowdown is really hurting the overall development experience.
To get around this I tested setting `PDFJS_NEXT = true` in `gulpfile.js`, since the transpilation step isn't necessary when testing in a modern browser.

However, having to edit `gulpfile.js` every time that I need to run tests isn't very practical. Hence this patch, which adds an environment variable that allows you to disable the transpilation simply by using e.g. `PDFJS_NEXT=true gulp test`.

I hope that this can be considered an acceptable solution, such that I don't need to maintain this patch locally (or worse, edit `gulpfile.js` locally before testing).

---
[1] This can also be observed on the Windows bot, but it seems fine on Linux.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -131,6 +131,8 @@ function createWebpackConfig(defines, output) {
   var licenseHeader = fs.readFileSync('./src/license_header.js').toString();
   var enableSourceMaps = !bundleDefines.FIREFOX && !bundleDefines.MOZCENTRAL &&
                          !bundleDefines.CHROME;
+  var pdfjsNext = bundleDefines.PDFJS_NEXT ||
+                  process.env['PDFJS_NEXT'] === 'true';
 
   return {
     output: output,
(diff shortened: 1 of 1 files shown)


yurydelendik (approved):
Looks good. Can you split that into two lines to add `var pdfjsNext =` before return?

Snuffleupagus:
> Looks good. Can you split that into two lines to add var pdfjsNext = before return?

That's a lot better, fixed now, thanks!

/botio-linux lint

PR #8467, opened 2017-05-31
Bug 1353029 - Pass PdfJs.enabled into child on change.

brendandahl:
Syncing changes from https://bugzilla.mozilla.org/show_bug.cgi?id=1353029


timvandermeij:
This is already merged upstream, so could you rebase onto the current master and resolve the ESLint failures so we can land it here as well?

PR #8472, opened 2017-06-01
Adds Streams API in getTextContent.

mukulmishra18:
Streams API in getTextContent, fetch data incrementally(in small chunks) instead of fetching bulky data.

This PR adds Streams API in getTextContent, to fetch textContent incrementally instead of fetch whole data at once, without changing it's core functionality. To fetch data incrementally, `textContent` is enqueued into the streams internal queue directly at `getTextContent` in `PartialEvaluator` whenever we have any.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1652,6 +1652,21 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
               // Enqueue textContent chunk before call for sub-form.
               enqueueChunk();
+              let sinkWrapper = {
+                cachedChunks = [],
+                enqueue(chunk, size) {
+                  sink.enqueue(chunk, size);
+                  this.cachedChunks.push(chunk);
+                },
+
+                get desiredSize() {
+                  return sink.desiredSize;
+                },
+
+                get ready() {
+                  return sink.ready;
+                }
+              };
               next(self.getTextContent({
                 stream: xobj,
                 task,
@@ -1659,7 +1674,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 stateManager: xObjStateManager,
                 normalizeWhitespace,
                 combineTextItems,
-                sink
+                sinkWrapper
               }).then(function (formTextContent) {}));
               return;
             case OPS.setGState:
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1166,7 +1166,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
     },
 
     getTextContent({ stream, task, resources, stateManager = null,
-                     normalizeWhitespace = false, combineTextItems = false, }) {
+                     normalizeWhitespace = false, combineTextItems = false,
+                     sink, }) {
       // Ensure that `resources`/`stateManager` is correctly initialized,
       // even if the provided parameter is e.g. `null`.
       resources = resources || Dict.empty;
@@ -1406,11 +1407,20 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
         textContentItem.str.length = 0;
       }
 
+      function enqueueChunk() {
+        if (textContent.items.length > 0) {
+          sink.enqueue(textContent, textContent.items.length);
+          textContent.items = [];
+          textContent.styles = Object.create(null);
+        }
+      }
+
       var timeSlotManager = new TimeSlotManager();
 
       return new Promise(function promiseBody(resolve, reject) {
         var next = function (promise) {
-          promise.then(function () {
+          enqueueChunk();
+          Promise.all([promise, sink.ready]).then(function () {
             try {
               promiseBody(resolve, reject);
             } catch (ex) {
@@ -1640,6 +1650,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 xObjStateManager.transform(matrix);
               }
 
+              // Enqueue textContent chunk before call for sub-form.
+              enqueueChunk();
               next(self.getTextContent({
                 stream: xobj,
                 task,
@@ -1647,12 +1659,13 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 stateManager: xObjStateManager,
                 normalizeWhitespace,
                 combineTextItems,
+                sink
               }).then(function (formTextContent) {
-                Util.appendToArray(textContent.items, formTextContent.items);
-                Util.extendObj(textContent.styles, formTextContent.styles);
+                // Util.appendToArray(textContent.items, formTextContent.items);
+                // Util.extendObj(textContent.styles, formTextContent.styles);
 
-                xobjsCache.key = name;
-                xobjsCache.texts = formTextContent;
+                // xobjsCache.key = name;
+                // xobjsCache.texts = formTextContent;
               }));
               return;
             case OPS.setGState:
@@ -1676,20 +1689,27 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               }
               break;
           } // switch
+          if (textContent.items.length >= sink.desiredSize) {
+            // Wait for ready, if we reach highWaterMark.
+            stop = true;
+            break;
+          }
         } // while
         if (stop) {
           next(deferred);
           return;
         }
         flushTextContentItem();
-        resolve(textContent);
+        enqueueChunk();
+        resolve();
       }).catch((reason) => {
         if (this.options.ignoreErrors) {
           // Error(s) in the TextContent -- allow text-extraction to continue.
           warn('getTextContent - ignoring errors during task: ' + task.name);
 
           flushTextContentItem();
-          return textContent;
+          enqueueChunk();
+          return;
         }
         throw reason;
       });
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1660,13 +1660,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 normalizeWhitespace,
                 combineTextItems,
                 sink
-              }).then(function (formTextContent) {
-                // Util.appendToArray(textContent.items, formTextContent.items);
-                // Util.extendObj(textContent.styles, formTextContent.styles);
-
-                // xobjsCache.key = name;
-                // xobjsCache.texts = formTextContent;
-              }));
+              }).then(function (formTextContent) {}));
               return;
             case OPS.setGState:
               flushTextContentItem();
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -874,30 +874,35 @@ var WorkerMessageHandler = {
       });
     }, this);
 
-    handler.on('GetTextContent', function wphExtractText(data) {
+    handler.on('GetTextContent', function wphExtractText(data, sink) {
       var pageIndex = data.pageIndex;
-      return pdfManager.getPage(pageIndex).then(function(page) {
+      sink.onPull = function (desiredSize) { };
+      sink.onCancel = function (reason) { };
+
+      pdfManager.getPage(pageIndex).then(function(page) {
         var task = new WorkerTask('GetTextContent: page ' + pageIndex);
         startWorkerTask(task);
 
         var pageNum = pageIndex + 1;
         var start = Date.now();
-        return page.extractTextContent({
+        page.extractTextContent({
           handler,
           task,
+          sink,
           normalizeWhitespace: data.normalizeWhitespace,
           combineTextItems: data.combineTextItems,
-        }).then(function(textContent) {
+        }).then(function() {
           finishWorkerTask(task);
 
           info('text indexing: page=' + pageNum + ' - time=' +
                (Date.now() - start) + 'ms');
-          return textContent;
+          sink.close();
         }, function (reason) {
           finishWorkerTask(task);
           if (task.terminated) {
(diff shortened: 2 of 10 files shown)


src/core/evaluator.js
```diff
@@ -1640,19 +1650,22 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
               }).then(function (formTextContent) {
-                Util.appendToArray(textContent.items, formTextContent.items);
-                Util.extendObj(textContent.styles, formTextContent.styles);
+                // Util.appendToArray(textContent.items, formTextContent.items);
+                // Util.extendObj(textContent.styles, formTextContent.styles);
 
-                xobjsCache.key = name;
-                xobjsCache.texts = formTextContent;
+                // xobjsCache.key = name;
+                // xobjsCache.texts = formTextContent;
```
Snuffleupagus: First of all, please don't add commented out code in a PR.

Second of all, this code cannot just be removed, since the whole point of using a cache here is to avoid having to parse a particular `XObject` more than once!

src/core/evaluator.js
```diff
@@ -1676,13 +1689,19 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+          }
         } // while
         if (stop) {
           next(deferred);
           return;
         }
         flushTextContentItem();
-        resolve(textContent);
+        enqueueChunk();
+        resolve();
```
Snuffleupagus: Similar changes needs to be made in the `if (this.options.ignoreErrors) {` block of the `catch` below, since otherwise you're definitely going to break text-extraction in some PDF files.

*Edit:* The `return textContent;` line probably needs to replaced with
```javascript
enqueueChunk();
return;
```

src/core/evaluator.js
```diff
@@ -1676,13 +1689,19 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               }
               break;
           } // switch
+          if (textContent.items.length >= sink.desiredSize) {
+            // wait for ready, if we reach highWaterMark.
```
Snuffleupagus: Nit: Capitalization.

src/core/evaluator.js
```diff
@@ -1640,19 +1650,22 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 xObjStateManager.transform(matrix);
               }
 
+              // enqueue textContent chunk before call for sub-form.
```
Snuffleupagus: Nit: Capitalization.

src/core/worker.js
```diff
@@ -874,30 +874,33 @@ var WorkerMessageHandler = {
...
         var task = new WorkerTask('GetTextContent: page ' + pageIndex);
         startWorkerTask(task);
 
         var pageNum = pageIndex + 1;
         var start = Date.now();
-        return page.extractTextContent({
-          handler,
-          task,
+        page.extractTextContent({
+          handler, task, sink,
```
Snuffleupagus: Nit: Since all the existing properties are on their own lines, please don't mix styles here; just add `sink,` on its own line.

src/display/api.js
```diff
@@ -957,10 +957,41 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
      */
     getTextContent: function PDFPageProxy_getTextContent(params) {
       params = params || {};
-      return this.transport.messageHandler.sendWithPromise('GetTextContent', {
+      var readableStream = this.transport.messageHandler
+                                           .sendWithStream('GetTextContent', {
```
Snuffleupagus: Weird indentation here, do e.g. this instead:
```javascript
let readableStream =
    this.transport.messageHandler.sendWithStream('GetTextContent', {
  pageIndex: this.pageNumber - 1,
  ...
```

src/display/api.js
```diff
@@ -957,10 +957,41 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
         pageIndex: this.pageNumber - 1,
         normalizeWhitespace: (params.normalizeWhitespace === true),
         combineTextItems: (params.disableCombineTextItems !== true),
+      }, {
+        highWaterMark: 5,
```
Snuffleupagus: Please avoid the magical number here, and replace it with a suitably named constant instead.

Snuffleupagus (commented):
Really sorry, but I have to ask: Did this PR actually pass all reference tests locally (particularly the `text` ones)?

mukulmishra18:
>Really sorry, but I have to ask: Did this PR actually pass all reference tests locally (particularly the `text` ones)?

No, this PR is failing `text` test. I have no idea why it is failing, that's why i submitted PR. I think submitting PR will help understand better what really the problem is. Hope you will understand.

src/core/evaluator.js
mukulmishra18: > Second of all, this code cannot just be removed, since the whole point of using a cache here is to avoid having to parse a particular XObject more than once!

Yes, you are right. But we are sending data to other side(main side) whenever we have any, so no need to cache.

Snuffleupagus: > [...] so no need to cache.

Yes there is!
Since we don't want to potentially parse the *same* `XObject` over and over again, which is what your patch forces us to do, the current code caches its textContent and returns it immediately if it's already available; please see https://github.com/mozilla/pdf.js/blob/593dec1bb7aec1802abf8268137b0f7adab2ae32/src/core/evaluator.js#L1607-L1614.

This optimization should remain intact, even with the new code, since the point isn't whether we send the data immediately to the main thread. The point is that we don't want to parse any `XObject` more than once!

src/core/evaluator.js
yurydelendik: resolve() actually resolves the outer promise -- just removing it will block its resolution

src/core/evaluator.js
yurydelendik: I think I added this caching at one point but not sure it addressed any issue. Let's find a test that actually regressed by removing this optimization.

yurydelendik:
I would like to see here also:
  - some changes to text_layer_builder to use stream reader + interrupting reading on page view cleanup.
  - some measurements that simulates human scrolling

src/core/evaluator.js
Snuffleupagus: > resolve() actually resolves the outer promise -- just removing it will block its resolution

I'm not sure that I follow here.
It might not have been clear from my comment, since I couldn't add it on the line in question, but I'd assume that we *also* need to update line https://github.com/mozilla/pdf.js/blob/593dec1bb7aec1802abf8268137b0f7adab2ae32/src/core/evaluator.js#L1692.

yurydelendik: ah, yes, at `this.options.ignoreErrors`

src/core/evaluator.js
mukulmishra18: Done

src/core/evaluator.js
mukulmishra18: Done

src/core/worker.js
mukulmishra18: Done

src/display/api.js
mukulmishra18: Looks nice. Done in new commit.

src/core/evaluator.js
Snuffleupagus: I've looked around quickly, but have thus far not found a case where caching the `textContent` helps.
*However*, that doesn't mean that the caching is without value; please see below!

For documents with a *huge* number of `XObject`s (such as #8380), despite *non* of its `XObject`s being of the `/Type /Form` kind, the cache still helps a lot.
The explanation is that [evaluator.js#L1608](https://github.com/mozilla/pdf.js/blob/593dec1bb7aec1802abf8268137b0f7adab2ae32/src/core/evaluator.js#L1608) means that we can bail out in [evaluator.js#L1613](https://github.com/mozilla/pdf.js/blob/593dec1bb7aec1802abf8268137b0f7adab2ae32/src/core/evaluator.js#L1613), before having to do a bunch of redundant `Dict` lookups in [evaluator.js#L1616-L1623](https://github.com/mozilla/pdf.js/blob/593dec1bb7aec1802abf8268137b0f7adab2ae32/src/core/evaluator.js#L1616-L1623).

I've tried disabling the caching, for the PDF file in #8380, and the run-time of `getTextContent` *tripled* (from ~200ms to ~600ms).

So, to try and summarize: I'd agree that we might not need to cache previous `textContent` as we currently do. However, I'm thinking that we should still cache (the names of) non-`Form` XObjects so that we can avoid unnecessary repeated `Dict` lookups.

---
Sorry, the above became a bit long; but I hope it makes sense!

src/core/evaluator.js
mukulmishra18: Done in new commit. Added: `enqueueChunk(); return;` in `this.options.ignoreErrors` block.

src/core/evaluator.js
yurydelendik: It makes sense. I just wanted to find a test case first, thanks. All we need to do is wrap sink and pass wrapper for inner processing: wrapper will redirect data to the original sink and store passed items. When cache needs to be used we need to have a simple loop that will write items into the stream.

src/display/api.js
yurydelendik: ReadableStream needs to know its "buffer size". I'm thinking to keep it around 100 as a starter, but it can go lower. `TEXT_CONTENT_CHUNK_SIZE` as a name?

src/display/api.js
```diff
@@ -957,10 +957,41 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
+        size(textContent) {
+          return textContent.items.length;
+        }
+      });
+      return new Promise((resolve, reject) => {
+        function pump() {
```
Snuffleupagus: You can probably simplify this function somewhat, by doing e.g. something like this:
```javascript
function pump() {
  reader.read().then(function({ done, data, }) {
    if (done) {
      resolve(textContent);
    } else {
      for (var key in data.styles) {
        textContent.styles[key] = data.styles[key];
      }
      textContent.items = textContent.items.concat(data.items);
      pump();
    }
  }, reject);
```

(Again, remember that unless you access `this`, there's no point to using an arrow function.)

src/display/api.js
```diff
@@ -957,10 +957,41 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
      */
     getTextContent: function PDFPageProxy_getTextContent(params) {
       params = params || {};
-      return this.transport.messageHandler.sendWithPromise('GetTextContent', {
+      var readableStream =
+        this.transport.messageHandler.sendWithStream('GetTextContent', {
```
Snuffleupagus: Nit: As indicated in https://github.com/mozilla/pdf.js/pull/8472#discussion_r119935306, please indent *this* line with two additional spaces (and let's use `let` instead when adding *new* code.)

src/display/api.js
```diff
@@ -957,10 +957,41 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
+            }
+          }, (error) => {
+            reject(error);
+          });
+        }
+
+        var reader = readableStream.getReader();
+        var textContent = {
+          items: [],
+          styles: {}
```
Snuffleupagus: Perhaps use `Object.create(null)` instead of `{}`?

mukulmishra18: > Perhaps use Object.create(null) instead of {}?

Okay, sure.
We are preferring `Object.create(null)` over `{}` because `Object.create(null)` creates a completely empty Object, and doing `{}` creates an Object having `__proto__` property?
Is it something related to memory Optimization?

src/core/evaluator.js
mukulmishra18: >  All we need to do is wrap sink and pass wrapper for inner processing: wrapper will redirect data to the original sink and store passed items.

So, wrapper for sink maybe something like: `let sinkWrapper = { sink: sink, items = [] }` ?
And instead of passing `sink` to `getTextContent()`, we can pass `sinkWrapper`.

yurydelendik: let sinkWrapper = {
  cachedChunks: [],
  get ready() { return sick.ready; },
  get desiredSize() { return sink.desiredSize; },
  enqueue(chunk, size) { sink.enqueue(chunk, size); this.cachedChunks.push(chunk); },
 // ... close (I don't think we are using onPull or onCancel though)

src/display/api.js
```diff
@@ -957,10 +957,41 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
+      return new Promise((resolve, reject) => {
+        function pump() {
+          reader.read().then((result) => {
+            if (result.done) {
+              resolve(textContent);
+            } else {
+              for (var key in result.value.styles) {
+                textContent.styles[key] = result.value.styles[key];
+              }
+              textContent.items = textContent.items.concat(result.value.items);
```
Snuffleupagus: Also, as mentioned in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat, `concat` creates a *new* array.
Hence, a better approach is probably to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push, which just extends the existing array instead.

Hence it's probably not a bad idea to use `Util.appendToArray` instead here!

mukulmishra18: > Hence it's probably not a bad idea to use Util.appendToArray instead here!

Then probably it better to have `Util.extendObj` instead of `textContent.styles[key]` for consistency?

PR #8482, opened 2017-06-04
Convert the thumbnail view/viewer to ES6 syntax

timvandermeij:
*Easier reviewing with https://github.com/mozilla/pdf.js/pull/8482/files?w=1.*

### DIFF
--- a/web/pdf_thumbnail_view.js
+++ b/web/pdf_thumbnail_view.js
@@ -19,8 +19,9 @@ import {
 import { getOutputScale, NullL10n } from './ui_utils';
 import { RenderingStates } from './pdf_rendering_queue';
 
-var THUMBNAIL_WIDTH = 98; // px
-var THUMBNAIL_CANVAS_BORDER_WIDTH = 1; // px
+const MAX_NUM_SCALING_STEPS = 3;
+const THUMBNAIL_CANVAS_BORDER_WIDTH = 1; // px
+const THUMBNAIL_WIDTH = 98; // px
 
 /**
  * @typedef {Object} PDFThumbnailViewOptions
@@ -31,7 +32,7 @@ var THUMBNAIL_CANVAS_BORDER_WIDTH = 1; // px
  * @property {PDFRenderingQueue} renderingQueue - The rendering queue object.
  * @property {boolean} disableCanvasToImageConversion - (optional) Don't convert
  *   the canvas thumbnails to images. This prevents `toDataURL` calls,
- *   but increases the overall memory usage. The default value is false.
+ *   but increases the overall memory usage. The default value is `false`.
  * @property {IL10n} l10n - Localization service.
  */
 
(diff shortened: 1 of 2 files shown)


web/pdf_thumbnail_view.js
```diff
@@ -156,296 +148,282 @@ var PDFThumbnailView = (function PDFThumbnailViewClosure() {
...
-      if (!noCtxScale && outputScale.scaled) {
-        ctx.scale(outputScale.sx, outputScale.sy);
-      }
-      return ctx;
-    },
+  /**
+   * @private
+   */
+  _convertCanvasToImage() {
+    if (!this.canvas || this.renderingState !== RenderingStates.FINISHED) {
```
Snuffleupagus: Given that these two conditions are quite different, I'm wondering if having two different `if`s (as in the current code) actually aid readability somewhat here!?

web/pdf_thumbnail_view.js
```diff
@@ -156,296 +148,282 @@ var PDFThumbnailView = (function PDFThumbnailViewClosure() {
...
+      finishRenderTask(null);
+    }, function(error) {
+      finishRenderTask(error);
+    });
+    return renderCapability.promise;
+  }
+
+  setImage(pageView) {
+    let img = pageView.canvas;
+    if (!img || this.renderingState !== RenderingStates.INITIAL) {
```
Snuffleupagus: Given that these two conditions are quite different, I'm wondering if having two different `if`s (as in the current code) actually aid readability somewhat here!?

web/pdf_thumbnail_viewer.js
```diff
@@ -30,201 +30,193 @@ var THUMBNAIL_SCROLL_MARGIN = -19;
...
-    setDocument: function PDFThumbnailViewer_setDocument(pdfDocument) {
-      if (this.pdfDocument) {
-        this._cancelRendering();
-        this._resetView();
-      }
+  /**
+   * @private
+   */
+  _resetView() {
+    this.thumbnails = [];
```
Snuffleupagus: Since you're touching this code anyway, mind changing it to  @`this._thumbnails` instead throughout the file?
This would address an inconsistency between `PDFViewer`/`PDFThumbnailViewer` (where the former uses `this._pages`), since `this.getThumbnail` should be used from outside of the class anyway.

Snuffleupagus (approved):
Just a couple of questions, feel free to push back if you don't agree!

Looks good, thank you for doing this!

web/pdf_thumbnail_view.js
timvandermeij: Personally I'm fine with both. To reduce the size of the diff, I reverted this change in the new commit.

web/pdf_thumbnail_view.js
timvandermeij: Personally I'm fine with both. To reduce the size of the diff, I reverted this change in the new commit.

web/pdf_thumbnail_viewer.js
timvandermeij: Done in the new commit.

PR #8483, opened 2017-06-04
Convert the find controller to ES6 syntax

timvandermeij:
Moreover, rename `FindStates` to `FindState` since enumeration names are usually not in plural, for readability and consistency with the ones in `src/shared/util.js`.

*Easier reviewing with https://github.com/mozilla/pdf.js/pull/8483/files?w=1.*

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -16,17 +16,18 @@
 import { createPromiseCapability } from 'pdfjs-lib';
 import { scrollIntoView } from './ui_utils';
 
-var FindStates = {
-  FIND_FOUND: 0,
-  FIND_NOTFOUND: 1,
-  FIND_WRAPPED: 2,
-  FIND_PENDING: 3,
+const FindState = {
+  FOUND: 0,
+  NOT_FOUND: 1,
+  WRAPPED: 2,
+  PENDING: 3,
 };
 
-var FIND_SCROLL_OFFSET_TOP = -50;
-var FIND_SCROLL_OFFSET_LEFT = -400;
+const FIND_SCROLL_OFFSET_TOP = -50;
+const FIND_SCROLL_OFFSET_LEFT = -400;
+const FIND_TIMEOUT = 250; // ms
 
-var CHARACTERS_TO_NORMALIZE = {
+const CHARACTERS_TO_NORMALIZE = {
   '\u2018': '\'', // Left single quotation mark
   '\u2019': '\'', // Right single quotation mark
   '\u201A': '\'', // Single low-9 quotation mark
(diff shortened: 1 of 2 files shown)


web/pdf_find_controller.js
```diff
@@ -16,17 +16,18 @@
 import { createPromiseCapability } from 'pdfjs-lib';
 import { scrollIntoView } from './ui_utils';
 
-var FindStates = {
-  FIND_FOUND: 0,
-  FIND_NOTFOUND: 1,
-  FIND_WRAPPED: 2,
-  FIND_PENDING: 3,
+const FindState = {
```
Snuffleupagus: I completely agree with the name change :-)

web/pdf_find_controller.js
```diff
@@ -41,447 +42,440 @@ var CHARACTERS_TO_NORMALIZE = {
...
-            break;
-          }
-          if (prevElem.match + prevElem.matchLength >=
-              currentElem.match + currentElem.matchLength) {
-            currentElem.skipped = true;
-            return true;
-          }
-        }
-        return false;
+  // Helper for multi-term search that fills the `matchesWithLength` array
```
Snuffleupagus: Nit: Thanks for improving the comment, but shouldn't it use JSDoc formatting since it's documenting a method?

web/pdf_find_controller.js
```diff
@@ -41,447 +42,440 @@ var CHARACTERS_TO_NORMALIZE = {
...
+    this.updateUIState(FindState.PENDING);
+
+    this._firstPagePromise.then(() => {
+      this.extractText();
+
+      clearTimeout(this.findTimeout);
+      if (cmd === 'find') {
+        // Trigger the find action with a small delay to avoid starting the
+        // search when the user is still typing (saving resources).
+        this.findTimeout = setTimeout(this.nextMatch.bind(this), FIND_TIMEOUT);
```
Snuffleupagus: Nice to see this replaced with a constant!

web/pdf_find_controller.js
```diff
@@ -41,447 +42,440 @@ var CHARACTERS_TO_NORMALIZE = {
...
 
-    nextPageMatch: function PDFFindController_nextPageMatch() {
-      if (this.resumePageIdx !== null) {
-        console.error('There can only be one pending page.');
+  nextPageMatch() {
+    if (this.resumePageIdx !== null) {
+      console.error('There can only be one pending page.');
+    }
+
+    let pageIdx = null, matches = null;
```
Snuffleupagus: Unless I'm mistaken, you *only* need to define `matches` here since it's used outside of the loop. However, `let pageIdx = ...` can go inside of the loop since that is limited to its scope.

Snuffleupagus (commented):
While this does look really good, I won't have time to test this patch properly until tomorrow[1]; sorry about that!

---
[1] Since we've got no tests for this code, I just want to make sure to do a proper round of testing *before* this lands.

web/pdf_find_controller.js
timvandermeij: Done in the new commit.

web/pdf_find_controller.js
timvandermeij: Done in the new commit.

Snuffleupagus (approved):
Looks good, thank you!

Issue #3120, opened 2013-04-19
Scrolling for PDF file is slow


cyroxx:
http://www.pharmazeutische-zeitung.de/fileadmin/pdf/PZ-Beilage-ApBetrO_2012_21.pdf

This looks odd and produces a bunch of errors. Actually, it is just those two:

```
downloadable font: table 'CFF ': failed to parse table (font-family: "g_font_p3_4" style:normal weight:normal stretch:normal src index:0)
```

and

```
downloadable font: rejected by sanitizer (font-family: "g_font_p3_4" style:normal weight:normal stretch:normal src index:0)
```

The other errors are in fact only siblings of the two mentioned above, as only the font family is different.

Note that the font is provided as a data url. Please also note that the document contains German umlauts.

**EDIT**
Tested under Firefox 20.0.1.

fkaelberer:
The original issue is still (or again?) there: 
Its broken in FF Dev Edition 36.0a2 (Win), FF nightly 37.0a1 (Win&Linux), Chrome 39 (Linux), 
It works fine in FF 34 (Win&Linux)
Here is page 11 in FF 36 and 34:
![zwischenablage01](https://cloud.githubusercontent.com/assets/4624768/5606647/a1d20254-9438-11e4-902c-291c88f124ce.png)

fkaelberer:
I also consider the performance issue as not fixed, even with #5586: try to scroll down to page 25 or some other page with a lot of small text, and the large number of div elements will slow Firefox down significantly (there's one div for every two letters or so). If its not slow, try scrolling up an down for a while.

Snuffleupagus:
@fkaelberer Since the font issue was actually fixed at one point (PR #4259), would you mind running [`git bisect`](https://github.com/mozilla/pdf.js/wiki/Bisecting-a-Regression) to find out where it regressed again?

fkaelberer:
Last good commit: 2803c31... Merge pull request #5129 from Snuffleupagus/issues-4875-4881-tests
([viewer](http://107.21.233.14:8877/01aecf0715f0d71/web/viewer.html) from previous commit, #5127)
First bad commit: 6865c28... Merge pull request #5111 from nnethercote/better-cidchars 
([viewer](http://107.21.233.14:8877/75f92af42a46fc1/web/viewer.html) from following commit, #5136)

fkaelberer:
[...]

PR #8488, opened 2017-06-05
Streams get text content

mukulmishra18:
Continuing PR #8472 

Aim of this PR, also mentioned in #8472 :
- [x] streamTextContent() API method
- [x] some changes to text_layer_builder to use stream reader + interrupting/canceling reading on page view cleanup (see renderTextLayer()).
- [x] some measurements that simulates human scrolling
- [x] caching of sub-extractTextContent calls (via sink wrapping)

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -271,7 +271,7 @@ var Page = (function PageClosure() {
     },
 
     extractTextContent({ handler, task, normalizeWhitespace,
-                         combineTextItems, }) {
+                         sink, combineTextItems, }) {
       var contentStreamPromise = this.pdfManager.ensure(this,
                                                         'getContentStream');
       var resourcesPromise = this.loadResources([
@@ -299,6 +299,7 @@ var Page = (function PageClosure() {
           resources: this.resources,
           normalizeWhitespace,
           combineTextItems,
+          sink,
         });
       });
     },
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1176,7 +1176,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
     },
 
     getTextContent({ stream, task, resources, stateManager = null,
-                     normalizeWhitespace = false, combineTextItems = false, }) {
+                     normalizeWhitespace = false, combineTextItems = false,
+                     sink, seenStyles = Object.create(null), }) {
       // Ensure that `resources`/`stateManager` is correctly initialized,
       // even if the provided parameter is e.g. `null`.
       resources = resources || Dict.empty;
@@ -1214,7 +1215,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
       // The xobj is parsed iff it's needed, e.g. if there is a `DO` cmd.
       var xobjs = null;
-      var xobjsCache = Object.create(null);
+      var skipEmptyXObjs = Object.create(null);
 
       var preprocessor = new EvaluatorPreprocessor(stream, xref, stateManager);
 
@@ -1225,7 +1226,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           return textContentItem;
         }
         var font = textState.font;
-        if (!(font.loadedName in textContent.styles)) {
+        if (!(font.loadedName in seenStyles)) {
+          seenStyles[font.loadedName] = true;
           textContent.styles[font.loadedName] = {
             fontFamily: font.fallbackName,
             ascent: font.ascent,
@@ -1416,11 +1418,21 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
         textContentItem.str.length = 0;
       }
 
+      function enqueueChunk() {
+        let length = textContent.items.length;
+        if (length > 0) {
+          sink.enqueue(textContent, length);
+          textContent.items = [];
+          textContent.styles = Object.create(null);
+        }
+      }
+
       var timeSlotManager = new TimeSlotManager();
 
       return new Promise(function promiseBody(resolve, reject) {
-        var next = function (promise) {
-          promise.then(function () {
+        let next = function (promise) {
+          enqueueChunk();
+          Promise.all([promise, sink.ready]).then(function () {
             try {
               promiseBody(resolve, reject);
             } catch (ex) {
@@ -1615,11 +1627,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               }
 
               var name = args[0].name;
-              if (xobjsCache.key === name) {
-                if (xobjsCache.texts) {
-                  Util.appendToArray(textContent.items, xobjsCache.texts.items);
-                  Util.extendObj(textContent.styles, xobjsCache.texts.styles);
-                }
+              if (name in skipEmptyXObjs) {
                 break;
               }
 
@@ -1633,8 +1641,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               assert(isName(type), 'XObject should have a Name subtype');
 
               if (type.name !== 'Form') {
-                xobjsCache.key = name;
-                xobjsCache.texts = null;
+                skipEmptyXObjs[name] = true;
                 break;
               }
 
@@ -1650,6 +1657,26 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 xObjStateManager.transform(matrix);
               }
 
+              // Enqueue the `textContent` chunk before parsing the /Form
+              // XObject.
+              enqueueChunk();
+              let sinkWrapper = {
+                enqueueInvoked: false,
+
+                enqueue(chunk, size) {
+                  this.enqueueInvoked = true;
+                  sink.enqueue(chunk, size);
+                },
+
+                get desiredSize() {
+                  return sink.desiredSize;
+                },
+
+                get ready() {
+                  return sink.ready;
+                },
+              };
+
               next(self.getTextContent({
                 stream: xobj,
                 task,
@@ -1657,12 +1684,12 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 stateManager: xObjStateManager,
                 normalizeWhitespace,
                 combineTextItems,
-              }).then(function (formTextContent) {
-                Util.appendToArray(textContent.items, formTextContent.items);
-                Util.extendObj(textContent.styles, formTextContent.styles);
-
-                xobjsCache.key = name;
-                xobjsCache.texts = formTextContent;
+                sink: sinkWrapper,
+                seenStyles,
+              }).then(function() {
+                if (!sinkWrapper.enqueueInvoked) {
+                  skipEmptyXObjs[name] = true;
+                }
               }));
               return;
             case OPS.setGState:
@@ -1686,20 +1713,27 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               }
               break;
           } // switch
+          if (textContent.items.length >= sink.desiredSize) {
+            // Wait for ready, if we reach highWaterMark.
+            stop = true;
+            break;
+          }
         } // while
         if (stop) {
           next(deferred);
           return;
         }
         flushTextContentItem();
-        resolve(textContent);
+        enqueueChunk();
+        resolve();
       }).catch((reason) => {
         if (this.options.ignoreErrors) {
           // Error(s) in the TextContent -- allow text-extraction to continue.
           warn('getTextContent - ignoring errors during task: ' + task.name);
 
           flushTextContentItem();
-          return textContent;
+          enqueueChunk();
+          return;
         }
         throw reason;
       });
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -952,15 +952,51 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
 
     /**
      * @param {getTextContentParameters} params - getTextContent parameters.
+     * @return {ReadableStream} ReadableStream to read textContent chunks.
+     */
+    streamTextContent(params = {}) {
+      const TEXT_CONTENT_CHUNK_SIZE = 100;
+      return this.transport.messageHandler.sendWithStream('GetTextContent', {
+        pageIndex: this.pageNumber - 1,
+        normalizeWhitespace: (params.normalizeWhitespace === true),
+        combineTextItems: (params.disableCombineTextItems !== true),
+      }, {
+        highWaterMark: TEXT_CONTENT_CHUNK_SIZE,
+        size(textContent) {
+          return textContent.items.length;
+        },
+      });
+    },
+
+    /**
+     * @param {getTextContentParameters} params - getTextContent parameters.
      * @return {Promise} That is resolved a {@link TextContent}
      * object that represent the page text content.
      */
     getTextContent: function PDFPageProxy_getTextContent(params) {
       params = params || {};
-      return this.transport.messageHandler.sendWithPromise('GetTextContent', {
-        pageIndex: this.pageNumber - 1,
-        normalizeWhitespace: (params.normalizeWhitespace === true),
-        combineTextItems: (params.disableCombineTextItems !== true),
+      let readableStream = this.streamTextContent(params);
+
+      return new Promise(function(resolve, reject) {
+        function pump() {
+          reader.read().then(function({ value, done, }) {
+            if (done) {
+              resolve(textContent);
+              return;
+            }
+            Util.extendObj(textContent.styles, value.styles);
+            Util.appendToArray(textContent.items, value.items);
+            pump();
+          }, reject);
+        }
+
+        let reader = readableStream.getReader();
+        let textContent = {
+          items: [],
+          styles: Object.create(null),
+        };
+
+        pump();
       });
     },
 
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -20,14 +20,20 @@ import { CustomStyle, getDefaultSetting } from './dom_utils';
  * Text layer render parameters.
  *
  * @typedef {Object} TextLayerRenderParameters
- * @property {TextContent} textContent - Text content to render (the object is
- *   returned by the page's getTextContent() method).
+ * @property {TextContent} textContent - (optional) Text content to render
+ *   (the object is returned by the page's getTextContent() method).
+ * @property {ReadableStream} textContentStream - (optional) Text content
+ *   stream to render (the stream is returned by the page's
+ *   streamTextContent() method).
  * @property {HTMLElement} container - HTML element that will contain text runs.
  * @property {PageViewport} viewport - The target viewport to properly
  *   layout the text runs.
  * @property {Array} textDivs - (optional) HTML elements that are correspond
  *   the text items of the textContent input. This is output and shall be
  *   initially be set to empty array.
+ * @property {Array} textContentItemsStr - (optional) Strings that correspond
+ *   the `str` property of the text items of textContent input. This is output
+ *   and shall be initially be set to empty array.
  * @property {number} timeout - (optional) Delay in milliseconds before
  *   rendering of the text  runs occurs.
  * @property {boolean} enhanceTextSelection - (optional) Whether to turn on the
@@ -122,6 +128,9 @@ var renderTextLayer = (function renderTextLayerClosure() {
       }
     }
     task._textDivProperties.set(textDiv, textDivProperties);
+    if (task._textContentStream) {
+      task._layoutText(textDiv);
+    }
 
     if (task._enhanceTextSelection) {
       var angleCos = 1, angleSin = 0;
@@ -157,7 +166,6 @@ var renderTextLayer = (function renderTextLayerClosure() {
     if (task._canceled) {
       return;
     }
-    var textLayerFrag = task._container;
     var textDivs = task._textDivs;
     var capability = task._capability;
     var textDivsLength = textDivs.length;
@@ -170,50 +178,12 @@ var renderTextLayer = (function renderTextLayerClosure() {
       return;
     }
 
-    // The temporary canvas is used to measure text length in the DOM.
-    var canvas = document.createElement('canvas');
-    if (typeof PDFJSDev === 'undefined' ||
-        PDFJSDev.test('FIREFOX || MOZCENTRAL || GENERIC')) {
-       canvas.mozOpaque = true;
+    if (!task._textContentStream) {
+      for (var i = 0; i < textDivsLength; i++) {
+        task._layoutText(textDivs[i]);
+      }
     }
-    var ctx = canvas.getContext('2d', { alpha: false, });
 
-    var lastFontSize;
-    var lastFontFamily;
-    for (var i = 0; i < textDivsLength; i++) {
-      var textDiv = textDivs[i];
-      var textDivProperties = task._textDivProperties.get(textDiv);
-      if (textDivProperties.isWhitespace) {
-        continue;
-      }
-
-      var fontSize = textDiv.style.fontSize;
-      var fontFamily = textDiv.style.fontFamily;
-
-      // Only build font string and set to context if different from last.
-      if (fontSize !== lastFontSize || fontFamily !== lastFontFamily) {
-        ctx.font = fontSize + ' ' + fontFamily;
-        lastFontSize = fontSize;
-        lastFontFamily = fontFamily;
-      }
-
-      var width = ctx.measureText(textDiv.textContent).width;
-      textLayerFrag.appendChild(textDiv);
-
-      var transform = '';
-      if (textDivProperties.canvasWidth !== 0 && width > 0) {
-        textDivProperties.scale = textDivProperties.canvasWidth / width;
-        transform = 'scaleX(' + textDivProperties.scale + ')';
-      }
-      if (textDivProperties.angle !== 0) {
-        transform = 'rotate(' + textDivProperties.angle + 'deg) ' + transform;
-      }
-      if (transform !== '') {
-        textDivProperties.originalTransform = transform;
-        CustomStyle.setProp('transform', textDiv, transform);
-      }
-      task._textDivProperties.set(textDiv, textDivProperties);
-    }
     task._renderingDone = true;
     capability.resolve();
   }
@@ -499,19 +469,27 @@ var renderTextLayer = (function renderTextLayerClosure() {
    * @param {boolean} enhanceTextSelection
    * @private
    */
-  function TextLayerRenderTask(textContent, container, viewport, textDivs,
-                               enhanceTextSelection) {
+  function TextLayerRenderTask({ textContent, textContentStream, container,
+                                 viewport, textDivs, textContentItemsStr,
+                                 enhanceTextSelection, }) {
     this._textContent = textContent;
+    this._textContentStream = textContentStream;
     this._container = container;
     this._viewport = viewport;
     this._textDivs = textDivs || [];
+    this._textContentItemsStr = textContentItemsStr || [];
+    this._enhanceTextSelection = !!enhanceTextSelection;
+
+    this._reader = null;
+    this._layoutTextLastFontSize = null;
+    this._layoutTextLastFontFamily = null;
+    this._layoutTextCtx = null;
     this._textDivProperties = new WeakMap();
     this._renderingDone = false;
     this._canceled = false;
     this._capability = createPromiseCapability();
     this._renderTimer = null;
     this._bounds = [];
-    this._enhanceTextSelection = !!enhanceTextSelection;
   }
   TextLayerRenderTask.prototype = {
     get promise() {
@@ -519,6 +497,10 @@ var renderTextLayer = (function renderTextLayerClosure() {
     },
 
     cancel: function TextLayer_cancel() {
+      if (this._reader) {
+        this._reader.cancel();
+        this._reader = null;
+      }
       this._canceled = true;
       if (this._renderTimer !== null) {
         clearTimeout(this._renderTimer);
@@ -527,21 +509,100 @@ var renderTextLayer = (function renderTextLayerClosure() {
       this._capability.reject('canceled');
     },
 
-    _render: function TextLayer_render(timeout) {
-      var textItems = this._textContent.items;
-      var textStyles = this._textContent.styles;
-      for (var i = 0, len = textItems.length; i < len; i++) {
-        appendText(this, textItems[i], textStyles);
+    _processItems(items, styleCache) {
+      for (let i = 0, len = items.length; i < len; i++) {
+        this._textContentItemsStr.push(items[i].str);
+        appendText(this, items[i], styleCache);
+      }
+    },
+
+    _layoutText(textDiv) {
+      let textLayerFrag = this._container;
+
+      let textDivProperties = this._textDivProperties.get(textDiv);
+      if (textDivProperties.isWhitespace) {
+        return;
       }
 
-      if (!timeout) { // Render right away
-        render(this);
-      } else { // Schedule
-        this._renderTimer = setTimeout(() => {
-          render(this);
-          this._renderTimer = null;
-        }, timeout);
+      let fontSize = textDiv.style.fontSize;
+      let fontFamily = textDiv.style.fontFamily;
+
+      // Only build font string and set to context if different from last.
+      if (fontSize !== this._layoutTextLastFontSize ||
+          fontFamily !== this._layoutTextLastFontFamily) {
+        this._layoutTextCtx.font = fontSize + ' ' + fontFamily;
+        this._lastFontSize = fontSize;
+        this._lastFontFamily = fontFamily;
       }
+
+      let width = this._layoutTextCtx.measureText(textDiv.textContent).width;
+
+      let transform = '';
+      if (textDivProperties.canvasWidth !== 0 && width > 0) {
+        textDivProperties.scale = textDivProperties.canvasWidth / width;
+        transform = 'scaleX(' + textDivProperties.scale + ')';
+      }
+      if (textDivProperties.angle !== 0) {
+        transform = 'rotate(' + textDivProperties.angle + 'deg) ' + transform;
+      }
+      if (transform !== '') {
+        textDivProperties.originalTransform = transform;
+        CustomStyle.setProp('transform', textDiv, transform);
+      }
+      this._textDivProperties.set(textDiv, textDivProperties);
+      textLayerFrag.appendChild(textDiv);
+    },
+
+    _render: function TextLayer_render(timeout) {
+      let capability = createPromiseCapability();
+      let styleCache = Object.create(null);
+
+      // The temporary canvas is used to measure text length in the DOM.
+      let canvas = document.createElement('canvas');
+      if (typeof PDFJSDev === 'undefined' ||
+          PDFJSDev.test('FIREFOX || MOZCENTRAL || GENERIC')) {
+         canvas.mozOpaque = true;
+      }
+      this._layoutTextCtx = canvas.getContext('2d', { alpha: false, });
+
+      if (this._textContent) {
+        let textItems = this._textContent.items;
+        let textStyles = this._textContent.styles;
+        this._processItems(textItems, textStyles);
+        capability.resolve();
+      } else if (this._textContentStream) {
+        let pump = () => {
+          this._reader.read().then(({ value, done, }) => {
+            if (done) {
+              capability.resolve();
+              return;
+            }
+
+            Util.extendObj(styleCache, value.styles);
+            this._processItems(value.items, styleCache);
+            pump();
+
+          }, capability.reject);
+        };
+
+        this._reader = this._textContentStream.getReader();
+        pump();
+      } else {
+        throw new Error('Neither "textContent" nor "textContentStream"' +
+          ' parameters specified.');
+      }
+
+      capability.promise.then(() => {
+        styleCache = null;
+        if (!timeout) { // Render right away
+          render(this);
+        } else { // Schedule
+          this._renderTimer = setTimeout(() => {
+            render(this);
+            this._renderTimer = null;
+          }, timeout);
+        }
+      }, this._capability.reject);
     },
 
     expandTextDivs: function TextLayer_expandTextDivs(expandDivs) {
@@ -610,11 +671,15 @@ var renderTextLayer = (function renderTextLayerClosure() {
    * @returns {TextLayerRenderTask}
    */
   function renderTextLayer(renderParameters) {
-    var task = new TextLayerRenderTask(renderParameters.textContent,
-                                       renderParameters.container,
-                                       renderParameters.viewport,
-                                       renderParameters.textDivs,
-                                       renderParameters.enhanceTextSelection);
+    var task = new TextLayerRenderTask({
+      textContent: renderParameters.textContent,
+      textContentStream: renderParameters.textContentStream,
+      container: renderParameters.container,
+      viewport: renderParameters.viewport,
+      textDivs: renderParameters.textDivs,
+      textContentItemsStr: renderParameters.textContentItemsStr,
+      enhanceTextSelection: renderParameters.enhanceTextSelection,
+    });
     task._render(renderParameters.timeout);
     return task;
   }
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -1378,6 +1378,7 @@ MessageHandler.prototype = {
         this.streamControllers[streamId] = {
           controller,
           startCall: startCapability,
+          isClosed: false,
         };
         this.postMessage({
           sourceName,
@@ -1409,6 +1410,7 @@ MessageHandler.prototype = {
       cancel: (reason) => {
         let cancelCapability = createPromiseCapability();
         this.streamControllers[streamId].cancelCall = cancelCapability;
+        this.streamControllers[streamId].isClosed = true;
         this.postMessage({
           sourceName,
           targetName,
@@ -1532,9 +1534,15 @@ MessageHandler.prototype = {
         });
         break;
       case 'enqueue':
-        this.streamControllers[data.streamId].controller.enqueue(data.chunk);
+        if (!this.streamControllers[data.streamId].isClosed) {
+          this.streamControllers[data.streamId].controller.enqueue(data.chunk);
+        }
         break;
       case 'close':
+        if (this.streamControllers[data.streamId].isClosed) {
+          break;
+        }
+        this.streamControllers[data.streamId].isClosed = true;
         this.streamControllers[data.streamId].controller.close();
         deleteStreamController();
         break;
@@ -1548,6 +1556,9 @@ MessageHandler.prototype = {
         deleteStreamController();
         break;
       case 'cancel':
+        if (!this.streamSinks[data.streamId]) {
+          break;
+        }
         resolveCall(this.streamSinks[data.streamId].onCancel,
                     [data.reason]).then(() => {
           sendStreamResponse({ stream: 'cancel_complete', success: true, });
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -24,8 +24,6 @@ import {
 import { getGlobalEventBus } from './dom_events';
 import { RenderingStates } from './pdf_rendering_queue';
 
-const TEXT_LAYER_RENDER_DELAY = 200; // ms
-
 /**
  * @typedef {Object} PDFPageViewOptions
  * @property {HTMLDivElement} container - The viewer element.
@@ -444,12 +442,11 @@ class PDFPageView {
     let resultPromise = paintTask.promise.then(function() {
       return finishPaintTask(null).then(function () {
         if (textLayer) {
-          pdfPage.getTextContent({
+          let readableStream = pdfPage.streamTextContent({
             normalizeWhitespace: true,
-          }).then(function textContentResolved(textContent) {
-            textLayer.setTextContent(textContent);
-            textLayer.render(TEXT_LAYER_RENDER_DELAY);
           });
+          textLayer.setTextContentStream(readableStream);
+          textLayer.render();
         }
       });
     }, function(reason) {
--- a/web/text_layer_builder.js
+++ b/web/text_layer_builder.js
@@ -41,6 +41,8 @@ var TextLayerBuilder = (function TextLayerBuilderClosure() {
     this.textLayerDiv = options.textLayerDiv;
     this.eventBus = options.eventBus || getGlobalEventBus();
     this.textContent = null;
+    this.textContentItemsStr = [];
+    this.textContentStream = null;
     this.renderingDone = false;
     this.pageIdx = options.pageIndex;
     this.pageNumber = this.pageIdx + 1;
@@ -79,7 +81,7 @@ var TextLayerBuilder = (function TextLayerBuilderClosure() {
      *   for specified amount of ms.
      */
     render: function TextLayerBuilder_render(timeout) {
-      if (!this.textContent || this.renderingDone) {
+      if (!(this.textContent || this.textContentStream) || this.renderingDone) {
         return;
       }
       this.cancel();
@@ -88,9 +90,11 @@ var TextLayerBuilder = (function TextLayerBuilderClosure() {
       var textLayerFrag = document.createDocumentFragment();
       this.textLayerRenderTask = renderTextLayer({
         textContent: this.textContent,
+        textContentStream: this.textContentStream,
         container: textLayerFrag,
         viewport: this.viewport,
         textDivs: this.textDivs,
+        textContentItemsStr: this.textContentItemsStr,
         timeout,
         enhanceTextSelection: this.enhanceTextSelection,
       });
@@ -113,6 +117,11 @@ var TextLayerBuilder = (function TextLayerBuilderClosure() {
       }
     },
 
+    setTextContentStream(readableStream) {
+      this.cancel();
+      this.textContentStream = readableStream;
+    },
+
     setTextContent: function TextLayerBuilder_setTextContent(textContent) {
       this.cancel();
       this.textContent = textContent;
@@ -122,8 +131,8 @@ var TextLayerBuilder = (function TextLayerBuilderClosure() {
                                                              matchesLength) {
       var i = 0;
       var iIndex = 0;
-      var bidiTexts = this.textContent.items;
-      var end = bidiTexts.length - 1;
+      let textContentItemsStr = this.textContentItemsStr;
+      var end = textContentItemsStr.length - 1;
       var queryLen = (this.findController === null ?
                       0 : this.findController.state.query.length);
       var ret = [];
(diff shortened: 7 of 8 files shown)


src/core/evaluator.js
```diff
@@ -1606,9 +1616,11 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
               var name = args[0].name;
               if (xobjsCache.key === name) {
-                if (xobjsCache.texts) {
-                  Util.appendToArray(textContent.items, xobjsCache.texts.items);
-                  Util.extendObj(textContent.styles, xobjsCache.texts.styles);
+                xobjsCache.cachedChunks = sink.cachedChunks;
```
yurydelendik: this line belong below after self.getTextContent call as `xobjsCache.cachedChunks = sinkWrapper.cachedChunks;`

yurydelendik:
Improvements for textContent style: with every chunk we are building new style object (and it's a regression since since we are passing more style). We need to pass an object/map/parameter (let's name it `seenStyles`) and use it at https://github.com/mukulmishra18/pdf.js/blob/f0c4c155bda9d3693c100f803aec61dc7f250b6f/src/core/evaluator.js#L1219. When we add a style to seenStyles we will add style at textContent.style. That will fix a regression and will also improve performance if we will pass seenStyles to the getTextContent sub-call.

Also, sinkWrapper can collect only items (without styles) -- styles were already sent. We asked to send cached items we need to send only portion of sink.desiredSize and block on sick.ready to send rest of the cached items.

src/core/evaluator.js
```diff
@@ -1406,11 +1407,20 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
         textContentItem.str.length = 0;
       }
 
+      function enqueueChunk() {
+        if (textContent.items.length > 0) {
```
Snuffleupagus: Nit: To avoid recomputing the length, I'd suggest adding e.g. `let length = textContent.items.length;` and using that here and below.

src/core/evaluator.js
```diff
@@ -1640,19 +1651,34 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+                },
+
+                get desiredSize() {
+                  return sink.desiredSize;
+                },
+
+                get ready() {
+                  return sink.ready;
+                },
+              };
```
Snuffleupagus: Nit: Newline here.

src/core/evaluator.js
```diff
@@ -1640,19 +1651,34 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
                 task,
                 resources: xobj.dict.get('Resources') || resources,
                 stateManager: xObjStateManager,
                 normalizeWhitespace,
                 combineTextItems,
-              }).then(function (formTextContent) {
-                Util.appendToArray(textContent.items, formTextContent.items);
-                Util.extendObj(textContent.styles, formTextContent.styles);
-
+                sinkWrapper,
```
Snuffleupagus: The object shorthand notation *only* works when the property and variable names matches. That isn't the case here, so this line will definitely break things; hence you need to change it to `sink: sinkWrapper,`

src/core/evaluator.js
```diff
@@ -1640,19 +1651,34 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 xObjStateManager.transform(matrix);
               }
 
+              // Enqueue textContent chunk before call for sub-form.
```
Snuffleupagus: Nit: What does "sub-form" mean here?
It might be better to just write something along the lines of:
```javascript
// Enqueue the `textContent` chunk before parsing the /Form XObject.
```

src/display/api.js
```diff
@@ -957,10 +957,41 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
+        size(textContent) {
+          return textContent.items.length;
+        },
+      });
+
+      return new Promise(function(resolve, reject) {
+        function pump() {
+          reader.read().then(function({ value, done, }) {
+            if (done) {
+              resolve(textContent);
```
Snuffleupagus: Just `return;` after this line and remove the `else` branch instead.

src/display/api.js
```diff
@@ -957,10 +957,41 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
+            } else {
+              Util.extendObj(textContent.styles, value.styles);
+              Util.appendToArray(textContent.items, value.items);
+              pump();
+            }
+          }, function(error) {
```
Snuffleupagus: `}, reject);` should work just as well here.

src/core/evaluator.js
mukulmishra18: > That isn't the case here, so this line will definitely break things; hence you need to change it to sink: sinkWrapper,

Oh, yes. That is why `text` test is failing. Silly mistake :-)
Thanks for pointing out.

src/core/evaluator.js
mukulmishra18: Done in new commit. Thanks.

src/core/evaluator.js
mukulmishra18: Done

src/core/evaluator.js
mukulmishra18: Done

src/core/evaluator.js
mukulmishra18: Sure, seems more clear. Done in new commit. Thanks for suggestion.

src/display/api.js
mukulmishra18: Done

src/display/api.js
mukulmishra18: Done.

mukulmishra18:
> Improvements for textContent style: with every chunk we are building new style object (and it's a regression since since we are passing more style). We need to pass an object/map/parameter (let's name it seenStyles) and use it at https://github.com/mukulmishra18/pdf.js/blob/f0c4c155bda9d3693c100f803aec61dc7f250b6f/src/core/evaluator.js#L1219.

So maybe we can change the signature of `getTextContent` to `getTextContent({...., seenStyles = [ ], })` and use `seenStyles` at https://github.com/mukulmishra18/pdf.js/blob/f0c4c155bda9d3693c100f803aec61dc7f250b6f/src/core/evaluator.js#L1219 like:
```javascript
if (!(font.loadedName in seenStyles)) {
  seenStyles.push(font.loadName);
  textContent.styles[font.loadedName] = { ... };
}
```

yurydelendik:
> So maybe we can change the signature of getTextContent to getTextContent({...., seenStyles = [ ], })

Correct, but it shall be a map (instead of array) -- it might be easier to create `Object.create(null)` during outer most call (without making seenStyles optional). Notice that `in` does not check if item is present -- it checks if property exists in the map/object.

mukulmishra18:
> Correct, but it shall be a map (instead of array) -- it might be easier to create Object.create(null) during outer most call (without making seenStyles optional).

If we are going to use map, then we have to store something like this: `seenStyles[font.loadedName] = value`, `value` maybe this: https://github.com/mozilla/pdf.js/blob/master/src/core/evaluator.js#L1219 also.
But do we need to store value, as we are using `seenStyles` only for lookup?

yurydelendik:
You can store `true` value in the seenStyles map.

mukulmishra18:
> we need to send only portion of sink.desiredSize and block on sick.ready to send rest of the cached items.

Okay, that means while sending back cached `textContent` from here: https://github.com/mozilla/pdf.js/pull/8488/files#diff-0b94c2e77a5259f7a728122fdbf9f46aR1623, we have to take into account `sink.ready` property. Maybe like:
```javascript
if (xobjsCache.cachedItems)
  xobjsCache.cachedItems.forEach(function(chunk) {
    sink.ready.then(function() {
      sink.enqueue({ items: chunk, styles: {}, }, chunk.length);
    });
  });
}
```

src/core/evaluator.js
```diff
@@ -1640,19 +1653,37 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
               }
 
+              // Enqueue the `textContent` chunk before parsing the /Form
+              // XObject.
+              enqueueChunk();
+              let sinkWrapper = {
+                cachedItems: [],
+                enqueue(chunk, size) {
+                  sink.enqueue(chunk, size);
+                  this.cachedItems.push(chunk.items);
```
yurydelendik: It's better to have continuous set of items, so `Util.appendToArray(this.cachedItems, chunk.items);` might be better

src/core/evaluator.js
```diff
@@ -1606,9 +1618,10 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
               var name = args[0].name;
               if (xobjsCache.key === name) {
-                if (xobjsCache.texts) {
-                  Util.appendToArray(textContent.items, xobjsCache.texts.items);
-                  Util.extendObj(textContent.styles, xobjsCache.texts.styles);
+                if (xobjsCache.cachedItems) {
+                  xobjsCache.cachedItems.forEach(function(chunk) {
```
yurydelendik: Since it's just items use `sink.enqueue({ items: xobjsCache.cachedItems, styles: {}, }, xobjsCache.cachedItems.length);` (no iterating)

yurydelendik (commented):
> Okay, that means while sending back cached textContent from here: https://github.com/mozilla/pdf.js/pull/8488/files#diff-0b94c2e77a5259f7a728122fdbf9f46aR1623, we have to take into account sink.ready property.

I found a pre-existing bug: we are sending the same text items, but theoretically they have to have different transform.

To fix https://github.com/mozilla/pdf.js/pull/8472#discussion_r119953235 , we just need to replace entire xobjsCache logic with `skipEmptyXobj` which will have XObject name if getTextContent sub-call does not invoke any enqueue method -- sink wrapper will be just modified to track that. Also, we don't have any test that have several same objects in a row with text content (only images, see PDF from referred discussion). @Snuffleupagus sounds like a good plan?

Snuffleupagus:
> I found a pre-existing bug: we are sending the same text items, but theoretically they have to have different transform.

That's a really nice find!

> To fix #8472 (comment) , we just need to replace entire xobjsCache logic with skipEmptyXobj which will have XObject name if getTextContent sub-call does not invoke any enqueue method -- sink wrapper will be just modified to track that. Also, we don't have any test that have several same objects in a row with text content (only images, see PDF from referred discussion). @Snuffleupagus sounds like a good plan?

Yes, that approach looks good to me!
If you don't mind a bit of bike-shedding on the name, how about `skipNonFormXObjs` instead, to make it clearer what type of `XObject`s that are put in the cache?
Also, with that cache now only storing names, couldn't we just use a `WeakMap` rather than an object?

yurydelendik:
> Also, with that cache now only storing names, couldn't we just use a WeakMap rather than an object?

WeakMap is good for objects only. For string map/object properties are fine. I was suggesting to use only one last name, but we can store all empty-text-items we can find.

mukulmishra18:
>  we just need to replace entire xobjsCache logic with `skipEmptyXobj` which will have XObject name if getTextContent sub-call does not invoke any enqueue method -- sink wrapper will be just modified to track that.

Then maybe we can add extra property in `sinkWrapper` to track wether enqueue is called or not, like:

```javascript
sinkWrapper = {
  enqueueInvoked: false,
  cachedItems: [ ],
  enqueue(chunk, size) {
    enqueueInvoked = true;
    ....
  },
}
```

and add `name` to `skipNonFormXobjs` after the sub-getTextContent call if `sinkWrapper.enqueueInvoked === true`, after this: https://github.com/mozilla/pdf.js/pull/8488/commits/e27eb6964363f58c831f905bec229d7b293fab60#diff-0b94c2e77a5259f7a728122fdbf9f46aR1686

We  don't need `xobjsCache` now? How we are going to use `skipNonFormXobjs`?

yurydelendik:
> `cachedItems: [ ],`

You don't need cachedItems.

> We don't need xobjsCache now? How we are going to use skipNonFormXobjs?

You don't need xobjsCache. The skipNonFormXobjs will be used similar to seenStyles, but it can be local to the function scope (no need to pass it as a parameter).

src/core/evaluator.js
mukulmishra18: Yes, right. Done in new commit. Thanks

src/core/evaluator.js
mukulmishra18: Done

src/core/document.js
```diff
@@ -299,6 +299,8 @@ var Page = (function PageClosure() {
           resources: this.resources,
           normalizeWhitespace,
           combineTextItems,
+          sink,
+          seenStyles: Object.create(null),
```
Snuffleupagus: Is there actually any reason to do this, can't you just use `seenStyles = Object.create(null),` in the method signature in `evaluator.js` instead?

src/core/evaluator.js
```diff
@@ -1204,7 +1205,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
       // The xobj is parsed iff it's needed, e.g. if there is a `DO` cmd.
       var xobjs = null;
-      var xobjsCache = Object.create(null);
+      var skipNonFormXobjs = Object.create(null);
```
Snuffleupagus: Nit: Since it's XObject (note the upper-case O), how about `skipNonFormXObjs` instead?

src/core/evaluator.js
```diff
@@ -1623,8 +1632,6 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               assert(isName(type), 'XObject should have a Name subtype');
 
               if (type.name !== 'Form') {
-                xobjsCache.key = name;
```
Snuffleupagus: Isn't the whole point here to XObject that does *not* have `/Subtype = /Form`?
Hence you need to replace this with `skipNonFormXObjs[name] = true;` here instead.

src/core/evaluator.js
```diff
@@ -1640,19 +1647,38 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
-                Util.appendToArray(textContent.items, formTextContent.items);
-                Util.extendObj(textContent.styles, formTextContent.styles);
-
-                xobjsCache.key = name;
-                xobjsCache.texts = formTextContent;
+                sink: sinkWrapper,
+                seenStyles,
+              }).then(function () {
+                if (sinkWrapper.enqueueInvoked === false) {
+                  skipNonFormXobjs[name] = true;
```
Snuffleupagus: Similar to https://github.com/mozilla/pdf.js/pull/8488/commits/1e9429e8f925ad74ce119a7ec503f5ae6d434539#r120729155, this line needs to be removed, since you're not suppose to cache these XObjects any more.

src/core/evaluator.js
```diff
@@ -1605,11 +1617,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               }
 
               var name = args[0].name;
-              if (xobjsCache.key === name) {
-                if (xobjsCache.texts) {
-                  Util.appendToArray(textContent.items, xobjsCache.texts.items);
-                  Util.extendObj(textContent.styles, xobjsCache.texts.styles);
-                }
+              // Skip if /Form XObject don't have text.
```
Snuffleupagus: This comment makes no sense, since what you're caching now is XObjects that does *not* have `/Subtype = /Form` set; let's just remove it.

src/core/document.js
mukulmishra18: Yes, sure. We can do this way also. I think i get confused by the https://github.com/mozilla/pdf.js/pull/8488#issuecomment-306547988

src/core/evaluator.js
mukulmishra18: Done in new commit. Thanks

src/core/evaluator.js
mukulmishra18: Done

src/core/evaluator.js
mukulmishra18: Done

src/core/evaluator.js
mukulmishra18: eh, Sorry. Removed in new commit.

src/core/evaluator.js
```diff
@@ -1604,55 +1616,65 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+              let matrix = xobj.dict.getArray('Matrix');
               if (isArray(matrix) && matrix.length === 6) {
                 xObjStateManager.transform(matrix);
               }
 
+              // Enqueue the `textContent` chunk before parsing the /Form
+              // XObject.
+              enqueueChunk();
+              let sinkWrapper = {
+                enqueueInvoked: false,
```
Snuffleupagus: This now appears to be totally unused, so I'm assuming that it can be removed?

mukulmishra18: Not sure, but it is mentioned in https://github.com/mozilla/pdf.js/pull/8488#pullrequestreview-42607568, that `skipNonFormXObj` should have `name` of XObject for which enqueue is not called, maybe that's why we need the code here: https://github.com/mozilla/pdf.js/pull/8488#pullrequestreview-42717620
Am i missing something?

yurydelendik: use at lines below `this.enqueueInvoked = true;` and there suppose to be then for `next(self.getTextContent({` which will `skipNonFormXObjs[name] = this.enqueueInvoked;`

src/core/evaluator.js
```diff
@@ -1604,55 +1616,65 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+              };
+
               next(self.getTextContent({
                 stream: xobj,
                 task,
                 resources: xobj.dict.get('Resources') || resources,
                 stateManager: xObjStateManager,
                 normalizeWhitespace,
                 combineTextItems,
-              }).then(function (formTextContent) {
```
yurydelendik: `missing skipNonFormXObjs[name] = sinkWrapper.enqueueInvoked` inside then()

yurydelendik: actually missing `if (sinkWrapper.enqueueInvoked)  {skipNonFormXObjs[name] = true;}`

src/core/evaluator.js
yurydelendik: nit: new line

src/core/evaluator.js
Snuffleupagus: Now I'm *really* confused here, since I thought that we agreed that we should only cache XObjects that have `Subtype !== Form`.
Did I completely misunderstand your intention with https://github.com/mozilla/pdf.js/pull/8488#pullrequestreview-42607568, and that we should *also* cache some XObjects with `Subtype === Form`?

mukulmishra18: > actually missing if (sinkWrapper.enqueueInvoked) {skipNonFormXObjs[name] = true;}

Is it not be like: `if (!sinkWrapper.enqueueInvoked) { skipNonFormXObjs[name] = true; }` instead?

yurydelendik: > that we should also cache some XObjects with Subtype === Form

it does not matter if XObject is Form or Not, but if it does not have text content, we shall not call getTextContent again

yurydelendik: > Is it not be like: `if (!sinkWrapper.enqueueInvoked) `

Yes, it's correct, sorry

Snuffleupagus: > it does not matter if XObject is Form or not, but if it does not have text content, we shall not call getTextContent again

OK, got it; thanks!

@mukulmishra18 Really sorry about the confusion caused here; let's update this according to the latest comments made by Yury.

src/core/evaluator.js
Snuffleupagus: Based on the latest round of comments in https://github.com/mozilla/pdf.js/pull/8488#discussion_r120747114, the initially suggested name `skipEmptyXObjs` probably makes a lot more sense.

Again, sorry about causing unnecessary confusion here!

src/core/evaluator.js
mukulmishra18: Okay, sure. Done in new commit. Thanks.

src/core/evaluator.js
mukulmishra18: Done in new commit. Thanks.

src/display/api.js
```diff
@@ -951,17 +951,60 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
     },
 
     /**
+     * @param {boolean} normalizeWhitespace - replaces all occurrences of
+     * whitespace with standard spaces (0x20). The default value is `false`.
+     * @param {boolean} disableCombineTextItems - do not attempt to combine
+     * same line {@link TextItem}'s. The default value is `false`.
```
Snuffleupagus: The method doesn't take parameters, but it takes an optional object, so the above comments should be replaced with what's used below:
`* @param {getTextContentParameters} params - getTextContent parameters.`

src/display/api.js
```diff
@@ -951,17 +951,60 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
+     * @return {ReadableStream} ReadableStream to read textContent chunks.
+     */
+    streamTextContent({ normalizeWhitespace = false,
+                        disableCombineTextItems = false, }) {
```
Snuffleupagus: Currently, it's optional to provide an options object to `getTextContent` and it'd be a really good idea if the new method has a similar signature!
I think that the following should work to make the object optional (otherwise just use the `params` notation as below):
```javascript
streamTextContent({ normalizeWhitespace, disableCombineTextItems, } = {}) {
```

src/display/api.js
```diff
@@ -951,17 +951,60 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
+      const TEXT_CONTENT_CHUNK_SIZE = 100;
+      return this.transport.messageHandler.sendWithStream('GetTextContent', {
+        pageIndex: this.pageNumber - 1,
+        normalizeWhitespace,
+        disableCombineTextItems,
```
Snuffleupagus: Please use the same validation as below, to ensure that what we're passing to the worker is sanitized!
```javascript
  normalizeWhitespace: normalizeWhitespace === true,
  combineTextItems: disableCombineTextItems !== true,
```

src/display/api.js
mukulmishra18: Done in new commit. Thanks.

src/display/api.js
mukulmishra18: Then using `params` will be better. Thanks. Updated in new commit.

src/display/api.js
mukulmishra18: Okay, sure. Thanks.

src/display/api.js
```diff
@@ -952,15 +952,51 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
     /**
      * @param {getTextContentParameters} params - getTextContent parameters.
+     * @return {ReadableStream} ReadableStream to read textContent chunks.
+     */
+    streamTextContent(params) {
+      const TEXT_CONTENT_CHUNK_SIZE = 100;
+      return this.transport.messageHandler.sendWithStream('GetTextContent', {
+        pageIndex: this.pageNumber - 1,
+        normalizeWhitespace: (params.normalizeWhitespace === true),
+        disableCombineTextItems: (params.disableCombineTextItems !== true),
```
Snuffleupagus: This isn't correct, it *needs* to be `combineTextItems: ...`, please see https://github.com/mozilla/pdf.js/pull/8488/files#diff-0ecad279ed2252e3eb47a4d96ec1463cL963 below.

src/display/api.js
```diff
@@ -952,15 +952,51 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
 
     /**
      * @param {getTextContentParameters} params - getTextContent parameters.
+     * @return {ReadableStream} ReadableStream to read textContent chunks.
+     */
+    streamTextContent(params) {
```
Snuffleupagus: `streamTextContent(params = {}) {`, otherwise `params` won't be optional (as it is for backwards compatibility reasons in `getTextContent` below).

src/display/api.js
mukulmishra18: Sorry, fixed in new commit.

src/display/api.js
mukulmishra18: Fixed, Thanks.

mukulmishra18:
Added some changes to use stream reader to stream `textContent` chunk in viewer. Viewer is loading just fine along with text selection with new code.
But i think we need some changes to use `cancel` method [here](https://github.com/mozilla/pdf.js/pull/8488/commits/e5d6050d45602c82c9c1f46768cae17feb7db6b2#diff-986df9c532f6b9ab6cf81a23ce26e446R526). As we are not passing `textContentStream` in `renderTextLayer`, `readableStream` is never initialized [here](https://github.com/mozilla/pdf.js/pull/8488/commits/e5d6050d45602c82c9c1f46768cae17feb7db6b2#diff-986df9c532f6b9ab6cf81a23ce26e446R507).(`renderTextLayer` is called from `text_layer_builder.js`).
So, the potential solution of this is, we can create another method `setTextContentStream` just like `setTextContent` [here](https://github.com/mozilla/pdf.js/blob/master/web/text_layer_builder.js#L116), and pass readableStream from `pdf_page_view.js` [here](https://github.com/mozilla/pdf.js/pull/8488/commits/e5d6050d45602c82c9c1f46768cae17feb7db6b2#diff-c880b70cf7d6b97863eeb665a736c4baR485). Something like:
```javascript
let readableStream = pdfPage.streamTextContent({
  normalizeWhitespace: true,
});
textLayer.setTextContentStream(readableStream);
```

Not sure if this is correct solution, require experts guidance @yurydelendik, @Snuffleupagus .

yurydelendik:
>  i think we need some changes to use cancel method

Notice that `cancel` is a method of the reader (and not of the stream) -- you need to keep opened reader reference to cancel it.

> create another method setTextContentStream

Having setTextContentStream sounds good to me

src/display/text_layer.js
```diff
@@ -528,20 +534,48 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
+      let capability = createPromiseCapability();
+      let pump;
+      if (this._textContent === undefined) {
+        pump = () => {
+          reader.read().then(({ value, done, }) => {
+            if (done) {
+              capability.resolve();
+              return;
+            }
+            Util.appendToArray(this._textContent.items, value.items);
```
yurydelendik: We don't need to collect all items from all chunks. We can process items as they coming in. Create e.g. `processItems(items, stylesCache)` and use it for a chunk and "this._textContent" case: move appendText-loop logic there. For chunks you have to maintain stylesCache with all styles, for this._textContent you can just use `this._textContent.styles`.

src/display/text_layer.js
```diff
@@ -528,20 +534,48 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
+            Util.extendObj(this._textContent.styles, value.styles);
+            pump();
+          });
```
yurydelendik: add `.catch(capability.reject);` to catch error and cancel request. Also below, properly handle the rejection by rejecting this._capability of the task.

src/display/text_layer.js
mukulmishra18: > We can process items as they coming in.

Not sure if i am getting you properly here. Do we need to have all styles cached before we start processing items, or we can build stylesCache on the fly during processing of items, something like:

```javascript
pump = () => {
  this._reader.read().then(({ value, done, }) => {
  if (done) {
    return;
  }

  Util.extendObj(styleCache, value.styles);
  this._processItems(value.items, stylesCache);
  pump();

  }, () => {
    this._capability.reject('Error in reading chunks');
  });
};
```

> Create e.g. `processItems(items, stylesCache)` and use it for a chunk

Can you please elaborate a bit `processItems(items, stylesCache)` working.

yurydelendik: `Util.extendObj(styleCache, value.styles);` look good for caching. processItems is whatever you see in the appendText-loop logic https://github.com/mukulmishra18/pdf.js/blob/e5d6050d45602c82c9c1f46768cae17feb7db6b2/src/display/text_layer.js#L566-L568.

yurydelendik:
Found #3120 with many text divs, let's measure improvement against this PDF

src/display/text_layer.js
```diff
@@ -527,21 +536,55 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
+
+      if (this._textContent) {
+        let textItems = this._textContent.items;
+        let textStyles = this._textContent.styles;
+        for (let i = 0, len = textItems.length; i < len; i++) {
+          appendText(this, textItems[i], textStyles);
+        }
+        capability.resolve();
+      } else {
+        this._styleCache = Object.create(null);
```
yurydelendik: make `styleCache` a function local.

src/display/text_layer.js
```diff
@@ -527,21 +536,55 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
-        appendText(this, textItems[i], textStyles);
+      let capability = createPromiseCapability();
+      let pump;
+
+      if (this._textContent) {
+        let textItems = this._textContent.items;
+        let textStyles = this._textContent.styles;
+        for (let i = 0, len = textItems.length; i < len; i++) {
+          appendText(this, textItems[i], textStyles);
+        }
```
yurydelendik: use `this._processItems(textItems, textStyles)`

web/pdf_page_view.js
```diff
@@ -452,11 +452,41 @@ class PDFPageView {
 
     let resultPromise = paintTask.promise.then(function() {
       return finishPaintTask(null).then(function () {
+        let pump;
```
Snuffleupagus: I really don't understand any of these changes; what's the point of streaming if you're still going to fetch all of the textContent here before rendering the textLayer?

As far as can tell, isn't the point that you just want to pass a `readableStream` via `web/text_layer_builder.js` onto `src/display/text_layer.js` where it's then used to actually get the textContent!? 

I.e. shouldn't this all of this code just be replaced with:
```javascript
if (textLayer) {
  pdfPage.streamTextContent({
    normalizeWhitespace: true,
  }).then(function(readableStream) {
    textLayer.setTextContentStream(readableStream);
    textLayer.render(TEXT_LAYER_RENDER_DELAY);
  });
}
```

mukulmishra18: > fetch all of the textContent here before rendering the textLayer

I think we need `textContent` in text_layer_builder.js files, [here](https://github.com/mozilla/pdf.js/blob/master/web/text_layer_builder.js#L125), not calling `setTextContent` gives an error.

Snuffleupagus: That just suggest, at least to me, that further changes may be required (perhaps in `text_layer_builder.js`).
However, if it turns out to be absolutely necessary to fetch *all* textContent here anyway, I really don't understand why all this code is needed here. In that case you could just keep the existing `pdfPage.getTextContent` call here (in addition to the `streamTextContent` one), to aid readability of this code!

src/display/text_layer.js
mukulmishra18: Done in new commit.

src/display/text_layer.js
mukulmishra18: Done. Thanks.

web/pdf_page_view.js
```diff
@@ -260,6 +260,10 @@ class PDFPageView {
    */
   updatePosition() {
     if (this.textLayer) {
+      let readableStream = this.pdfPage.streamTextContent({
+        normalizeWhitespace: true,
+      });
+      this.textLayer.setTextContentStream(readableStream);
```
Snuffleupagus: Considering that there were no existing `setTextContent` call here, I'm sceptical that you need to add this code here.

This would basically force *a lot* of duplicate streams being created at *every* scroll event, which seems like it would be really bad for performance. I'd assume that issue observed actually lies elsewhere, rather than in *this* particular method.

mukulmishra18:
New commit fixes two issues, that we discussed:

-  Updates `textContentStream` at [updatePosition](https://github.com/mozilla/pdf.js/blob/master/web/pdf_page_view.js#L261), to fix error during `readableStream.getReader`

-  Extra check at `enqueue` [case](https://github.com/mozilla/pdf.js/blob/master/src/shared/util.js#L1534), to not enqueue more chunks when cancel is called.

I tried to run viewer with this code and now there is no error in the console related to Streams API, and text selection is also working. But unfortunately, it is bit slow to load pages while scrolling and memory usage is also high (as compare to master) :(

web/text_layer_builder.js
```diff
@@ -79,22 +81,23 @@ var TextLayerBuilder = (function TextLayerBuilderClosure() {
      *   for specified amount of ms.
      */
     render: function TextLayerBuilder_render(timeout) {
-      if (!this.textContent || this.renderingDone) {
+      if (!this.textContentStream || this.renderingDone) {
```
Snuffleupagus: This is basically making this whole file unusable together with `setTextContent`.
I'm assuming that you what you really want is something like this:
```javascript
if (!(this.textContent || this.textContentStream) || this.renderingDone) {
```

web/text_layer_builder.js
```diff
@@ -79,22 +81,23 @@ var TextLayerBuilder = (function TextLayerBuilderClosure() {
...
         return;
       }
       this.cancel();
 
       this.textDivs = [];
       var textLayerFrag = document.createDocumentFragment();
       this.textLayerRenderTask = renderTextLayer({
-        textContent: this.textContent,
+        textContentStream: this.textContentStream,
```
Snuffleupagus: Unless we completely remove the ability to use `setTextContent`, I think that you need to have both of these here:
```javascript
  textContent: this.textContent,
  textContentStream: this.textContentStream,
```

web/text_layer_builder.js
```diff
@@ -79,22 +81,23 @@ var TextLayerBuilder = (function TextLayerBuilderClosure() {
...
         container: textLayerFrag,
         viewport: this.viewport,
         textDivs: this.textDivs,
         timeout,
         enhanceTextSelection: this.enhanceTextSelection,
       });
       this.textLayerRenderTask.promise.then(() => {
+        this.textContentItemsStr = this.textLayerRenderTask.textContentItemsStr;
```
Snuffleupagus: Isn't it better to follow the same pattern as is used with `this.textDivs` for this property as well?

src/display/text_layer.js
```diff
@@ -610,11 +653,14 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
-                                       renderParameters.textDivs,
-                                       renderParameters.enhanceTextSelection);
+    var task = new TextLayerRenderTask({
+                textContent: renderParameters.textContent,
+                textContentStream: renderParameters.textContentStream,
+                container: renderParameters.container,
+                viewport: renderParameters.viewport,
+                textDivs: renderParameters.textDivs,
+                enhanceTextSelection: renderParameters.enhanceTextSelection,
+              });
```
Snuffleupagus: The indentation is completely off here, it needs to be this:
```javascript
var task = new TextLayerRenderTask({
  textContent: renderParameters.textContent,
  ...
});
```

src/display/text_layer.js
```diff
@@ -20,8 +20,10 @@ import { CustomStyle, getDefaultSetting } from './dom_utils';
  * Text layer render parameters.
  *
  * @typedef {Object} TextLayerRenderParameters
- * @property {TextContent} textContent - Text content to render (the object is
- *   returned by the page's getTextContent() method).
+ * @property {TextContent} textContent - (optional) Text content to render
+ *   (the object is returned by the page's getTextContent() method).
+ * @property {ReadableStream} textContentStream - Text content stream to render.
```
Snuffleupagus: Nit: This should probably also be marked as optional.

src/display/text_layer.js
```diff
@@ -499,9 +501,12 @@ var renderTextLayer = (function renderTextLayerClosure() {
    * @param {boolean} enhanceTextSelection
    * @private
    */
-  function TextLayerRenderTask(textContent, container, viewport, textDivs,
-                               enhanceTextSelection) {
+  function TextLayerRenderTask({ textContent, textContentStream, container,
+                                 viewport, textDivs, enhanceTextSelection, }) {
+    this.textContentItemsStr = [];
```
Snuffleupagus: Similar to another comment in `web/text_layer_builder.js`, I think this should be handled similar to the `textDivs` instead.

src/display/text_layer.js
```diff
@@ -527,21 +536,55 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
+      }
+    },
+
     _render: function TextLayer_render(timeout) {
-      var textItems = this._textContent.items;
-      var textStyles = this._textContent.styles;
-      for (var i = 0, len = textItems.length; i < len; i++) {
-        appendText(this, textItems[i], textStyles);
+      let capability = createPromiseCapability();
+      let pump;
```
Snuffleupagus: Nit: I don't think that you need to define this variable here, since you're only using it in one branch below.

src/display/text_layer.js
```diff
@@ -527,21 +536,55 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
+      let styleCache = Object.create(null);
+
+      if (this._textContent) {
+        let textItems = this._textContent.items;
+        let textStyles = this._textContent.styles;
+        this._processItems(textItems, textStyles);
+        capability.resolve();
+      } else {
```
Snuffleupagus: It's probably a good idea to make this `} else if (this._textContentStream) {` instead; see comment below as well.

yurydelendik:
I'm getting `this.streamControllers[data.streamId] is undefined` in the `case 'enqueue':`

src/display/text_layer.js
```diff
@@ -527,21 +536,55 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
+            Util.extendObj(styleCache, value.styles);
+            this._processItems(value.items, styleCache);
+            pump();
+
+          }, capability.reject);
+        };
+
+        this._reader = this._textContentStream.getReader();
+        pump();
       }
```
Snuffleupagus: It's probably a good idea to ensure that either `textContent` *or* `textContentStream` was provided, so inline with the comment just above let's do:
```javascript
} else {
  throw new Error('Neither "textContent" nor "textContentStream" parameters specified.');
}
```

src/shared/util.js
```diff
@@ -1531,10 +1532,18 @@ MessageHandler.prototype = {
         });
         break;
       case 'enqueue':
-        this.streamControllers[data.streamId].controller.enqueue(data.chunk);
+        stream = this.streamControllers[data.streamId].controller
+                  ._controlledReadableStream;
```
Snuffleupagus: Nit: Please leave the dot on the previous line instead.

yurydelendik:
> it is bit slow to load pages while scrolling and memory usage is also high

Please publish automatic script that performs scrolling

Snuffleupagus (commented):
I've added a number of suggestions/comments.

Besides that, I'd ask that you please *do not* replace `var`s in this patch *unless* you are already touching the line in question.
This is a fairly big patch, that touches a lot of different code, a lot of it which is important for a well functioning viewer; and all the "unrelated" changes are making it more difficult to focus on the important changes.

web/pdf_page_view.js
yurydelendik: We just re-requesting data here since we are not storing entire text content anymore. getReader() can be requested once. We can cache entire textContent content after the read, however I want to try to get away without caching.

Snuffleupagus: > We just re-requesting data here since we are not storing entire text content anymore. getReader() can be requested once. We can cache entire textContent content after the read, however I want to try to get away without caching.

Sure, but considering that [`PDFPageView.updatePosition`](https://github.com/mozilla/pdf.js/blob/82f3145a5d8630378f378ef8f4f8596204273a6b/web/pdf_page_view.js#L261) is called from [`PDFViewer._scrollUpdate`](https://github.com/mozilla/pdf.js/blob/82f3145a5d8630378f378ef8f4f8596204273a6b/web/pdf_viewer.js#L455) which is called on [*every* scroll event](https://github.com/mozilla/pdf.js/blob/82f3145a5d8630378f378ef8f4f8596204273a6b/web/pdf_viewer.js#L129), couldn't all the (potentially thousands) of `streamTextContent` calls slow us down considerably here?

I'm even thinking that it may explain https://github.com/mozilla/pdf.js/pull/8488#issuecomment-308158072: 
> But unfortunately, it is bit slow to load pages while scrolling and memory usage is also high (as compare to master) :(

Couldn't we at least avoid doing this if the textLayer has already been completely rendered?

yurydelendik: oh, wow. we are doing text layer update for each time we scroll? hmm....

Snuffleupagus: We do, and I think it has something to do with the find code; at least [PDFFindController.updatePage]https://github.com/mozilla/pdf.js/blob/82f3145a5d8630378f378ef8f4f8596204273a6b/web/pdf_find_controller.js#L280 seems to me that it could be connected!?

We might want to try and make sense of the find code, and refactor that first?

yurydelendik: okay, plan is to change if above to `if (this.textLayer && !this.textLayer.renderingDone) {`

web/text_layer_builder.js
mukulmishra18: Done, Thanks.

web/text_layer_builder.js
mukulmishra18: Done.

web/text_layer_builder.js
mukulmishra18: Sure, this looks more logical. Done in new commit. Thanks

src/display/text_layer.js
mukulmishra18: Done

src/display/text_layer.js
mukulmishra18: Done

src/display/text_layer.js
mukulmishra18: Done

src/display/text_layer.js
mukulmishra18: Sure, thanks.

src/display/text_layer.js
mukulmishra18: Sorry, done in new commit.

src/shared/util.js
mukulmishra18: Removed this code in new commit(rather added an extra property `isClosed` in streamsController to check if stream is closed). Also removed unnecessary `let`s in the patch. Thanks for all your suggestions.

web/pdf_page_view.js
```diff
@@ -258,11 +258,7 @@ class PDFPageView {
   /**
    * Called when moved in the parent's container.
    */
-  updatePosition() {
-    if (this.textLayer) {
-      this.textLayer.render(TEXT_LAYER_RENDER_DELAY);
-    }
-  }
+  updatePosition() { }
```
yurydelendik: The `updatePosition` was removed in #8531

yurydelendik:
- [ ] extract https://github.com/mukulmishra18/pdf.js/blob/06db7e9b54f2f777022538c2581a587ea839593e/src/display/text_layer.js#L186-L217 logic into separate method (e.g. _layoutText) and use it from the appendText().

src/display/text_layer.js
```diff
@@ -527,21 +537,57 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
-          this._renderTimer = null;
-        }, timeout);
-      }
+        } else { // Schedule
+          this._renderTimer = setTimeout(() => {
+            render(this);
+            this._renderTimer = null;
+          }, timeout);
+        }
+      }, () => {
```
yurydelendik: no need for function here: use this._capability.reject directly

web/pdf_page_view.js
```diff
@@ -453,12 +449,11 @@ class PDFPageView {
...
         if (textLayer) {
-          pdfPage.getTextContent({
+          let readableStream = pdfPage.streamTextContent({
             normalizeWhitespace: true,
-          }).then(function textContentResolved(textContent) {
-            textLayer.setTextContent(textContent);
-            textLayer.render(TEXT_LAYER_RENDER_DELAY);
           });
+          textLayer.setTextContentStream(readableStream);
+          textLayer.render(TEXT_LAYER_RENDER_DELAY);
```
yurydelendik: let's try to remove TEXT_LAYER_RENDER_DELAY here

src/display/text_layer.js
mukulmishra18: > no need for function here: use this._capability.reject directly

I tried to reject _capability without function, and `_capability` is always rejected irrespective of it's state. But using with function just working fine. Is i am doing something wrong, or there is any confusion here.

web/pdf_page_view.js
mukulmishra18: Rebased to new master. Thanks.

web/pdf_page_view.js
mukulmishra18: Done with new commit. Thanks.

src/display/text_layer.js
Snuffleupagus: Replacing this line with `}, this._capability.reject);` instead *ought* to work just fine, is that what you tried?

web/text_layer_builder.js
```diff
@@ -122,8 +131,8 @@ var TextLayerBuilder = (function TextLayerBuilderClosure() {
                                                              matchesLength) {
       var i = 0;
       var iIndex = 0;
-      var bidiTexts = this.textContent.items;
-      var end = bidiTexts.length - 1;
+      let bidiTextsStr = this.textContentItemsStr;
```
Snuffleupagus: If you're going to change the local variable name anyway, how about using consistent names instead (applies below as well)?
I.e. `let textContentItemsStr = this.textContentItemsStr;`.

src/display/text_layer.js
```diff
@@ -20,8 +20,10 @@ import { CustomStyle, getDefaultSetting } from './dom_utils';
  * Text layer render parameters.
  *
  * @typedef {Object} TextLayerRenderParameters
- * @property {TextContent} textContent - Text content to render (the object is
- *   returned by the page's getTextContent() method).
+ * @property {TextContent} textContent - (optional) Text content to render
+ *   (the object is returned by the page's getTextContent() method).
+ * @property {ReadableStream} textContentStream - (optional) Text Content stream
```
Snuffleupagus: Spelling nit: `content` with lowercase c instead?

src/display/text_layer.js
```diff
@@ -20,8 +20,10 @@ import { CustomStyle, getDefaultSetting } from './dom_utils';
...
+ *   to render(the stream is returned by the page's streamTextContent() method).
```
Snuffleupagus: Nit: Missing space between `render` and `(`.

Also, there needs to be a JSDoc comment for `textContentItemsStr` as well (similar to `textDivs` below).

src/display/text_layer.js
```diff
@@ -499,9 +465,16 @@ var renderTextLayer = (function renderTextLayerClosure() {
    * @param {boolean} enhanceTextSelection
    * @private
    */
-  function TextLayerRenderTask(textContent, container, viewport, textDivs,
-                               enhanceTextSelection) {
+  function TextLayerRenderTask({ textContent, textContentStream, container,
+                                 viewport, textDivs, textContentItemsStr,
+                                 enhanceTextSelection, }) {
+    this._layoutTextLastFontSize = null;
```
Snuffleupagus: Can we please try to order the properties in a more logical way here, by first initializing the ones from the parameters and then the rest after?
```javascript
this._textContent = textContent;
this._textContentStream = textContentStream;
this._container = container;
this._viewport = viewport;
this._textDivs = textDivs || [];
this._textContentItemsStr = textContentItemsStr || [];
this._enhanceTextSelection = !!enhanceTextSelection;

this._reader = null;
// All other existing (and new) properties here...
```

src/display/text_layer.js
mukulmishra18: >Replacing this line with `}, this._capability.reject);` instead ought to work just fine

Yes, you are right. This is working just fine.

But i am trying something like `}, this._capability.reject('Error while processing chunks'));`

Snuffleupagus:
I've just tested the latest version of the patch locally, and while text-selection still seems to work, when scrolling quickly there's errors printed in the console (the following is just an excerpt):
```
PDF 110dd61fd57444010b1ab5ff38782f0f [1.4 pdfeTeX-1.21a / TeX] (PDF.js: -) app.js:1828:13
TypeError: The stream (in closed state) is not in the readable state and cannot be enqueued to system.js%20line%204%20%3E%20eval:2402:15
TypeError: The stream (in closed state) is not in the readable state and cannot be closed system.js%20line%204%20%3E%20eval:2384:15
TypeError: this.streamSinks[data.streamId] is undefined system.js%20line%204%20%3E%20eval:1505:15
```

mukulmishra18:
> TypeError: The stream (in closed state) is not in the readable state and cannot be enqueued to

Sorry, i got why this error is occurring. [Here](https://github.com/mukulmishra18/pdf.js/commit/38a2a78cb649cae4e4a2a1e345517b199a6b7f5c#diff-8d2b578d58d4d8c8a2975020fbbc84d9R1537), I am accidentally checking `isClosed` as a property of `controller`, but it should be a property of `streamsController` Object. Replacing `if (!this.streamControllers[data.streamId].controller.isClosed) {` with `if (!this.streamControllers[data.streamId].isClosed) {` fixes the issue.

> TypeError: The stream (in closed state) is not in the readable state and cannot be closed
TypeError: this.streamSinks[data.streamId] is undefined system.js%20line%204%20%3E%20eval:1505:15

Not sure about these issues, but these should be some race conditions, when more close events are called for the same stream.

src/display/text_layer.js
Snuffleupagus: > But i am trying something like }, this._capability.reject('Error while processing chunks'));

Sorry, but that simply won't work! Why are you attempting to do that in the first place?

Please just use `}, this._capability.reject);` as suggested in both https://github.com/mozilla/pdf.js/pull/8488#discussion_r122292008 and https://github.com/mozilla/pdf.js/pull/8488#discussion_r122509772, since that will propagate the underlying error which is all that's really needed here.

mukulmishra18: Okay, i got it. By `}, this._capability.reject);`, we are just propagating down the same error.
Thanks, done with new commit.

web/text_layer_builder.js
mukulmishra18: Done with new commit. Thanks.

src/display/text_layer.js
mukulmishra18: Done. Thanks.

src/display/text_layer.js
mukulmishra18: Okay, added JSDoc comment for `textContentItemsStr` in new commit. Thanks.

src/display/text_layer.js
mukulmishra18: Done with new commit. Thanks for your guidance here.

mukulmishra18:
### Performance measurement

***

Steps for measuring the performance:

- Open the viewer in master branch, and paste [this](https://gist.github.com/mukulmishra18/831e74fc0d49398f3172c5af0c6c4afc) code in console to take data for automation of scrolling.
- Paste above taken data into the console(as `rr = data`), and run [this](https://gist.github.com/mukulmishra18/19e53e31fb89ce704866a7e2ad06d3a5) code in console to perform automatic viewer scrolling.
- Measure the memory usage using devtools or run [this](https://gist.github.com/mukulmishra18/18635de4c82522bde3e7c4c0761f3365) code as `python name_of_file.py <pID>`, where pID is process ID of viewer. This will give peak memory used, when process ends.

**Measurement Configuration**
- Operating system: Ubuntu 14.04 LTS
- Browser: Chrome (version: 58.0.3)

**Output:**

| Branch\measurement[1] | 1 | 2 | 3 |
| ---- | --- | --- | --- |
| master | 238.70 | 231.23 | 230.75 |
| streams | 236.15 | 234.24 | 233.78 |

[1]: Peak memory value in MB.

yurydelendik (approved):
Really good. Looks like we don't regress performance of the viewer, but entire text layer becomes more responsive to cancelling and potentially uses less memory at once.

yurydelendik:
Thank you for the patch. Good work!

PR #7904, opened 2016-12-20
Further adjust the heuristics used to detect OpenType font files with CFF data, to ensure that all Type0 fonts are handled the same way regardless of font Subtype (issue 7901)

Snuffleupagus:
Changing this particular code makes me somewhat nervous about regressions, since PR #5770 necessitated the follow-up PR #6270.
However, the patch passes all tests added in those PRs (and obviously all other tests). Furthermore, I've manually checked all the issues/bugs referenced in PRs #5770 and #6270 without finding any issues.

**Please note:** This patch fixes *only* the font bug, not the SVG conversion, present on pages two and three of the PDF file in issue #7901.

PR #6270, opened 2015-07-25
Adjust the heuristics used to detect OpenType font file with CFF data (bug 1186827, bug 1182130, issue 6264)

Snuffleupagus:
_This is a tentative patch._

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1186827.
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1182130.
Fixes #6264.

@brendandahl This is the best I'm able to come up with, on short notice, to fix the above regressions without breaking already existing test-cases. Could you please review this patch?
(Also, if this doesn't seem like a good solution, would you have time to work on those issues yourself?)

PR #5770, opened 2015-02-28
Correctly detect OpenType font files with CFF data

Snuffleupagus:
Fixes #5334.
Fixes #215.
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1093246.
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1125614.

According to the specification, http://www.microsoft.com/typography/otspec/otff.htm, OpenType font files with CFF data should have `OTTO` in the header.

PR #8507, opened 2017-06-09
Only special-case OpenType fonts with `CFF` data if it's both a composite (i.e. Type0) font and also has a non-default CID to GID map (issue 8480)

### LINKED ISSUE #8480: Font not displaying characters properly (standard char set)
Link to PDF file (or attach file here): https://www.ordinate.buzz/sample.pdf

Configuration:
- Web browser and its version: Issue occurs on FF and Chrome. Works on IE
- PDF.js version: //mozilla.github.io/pdf.js/build/pdf.js
- Is an extension:

My code (standard code pasted from samples) is below. The file displays 'normally' in a PDF viewer, but with PDF.JS on Chrome/Firefox, it doesn't. 

What went wrong? 

Various letters of the font in the top line of text don't get rendered properly. The font in question is called 'GE Inspira Sans'

See Screenshot.

capture

<script>	
	var url = 'sample.pdf';

// Disable workers to avoid yet another cross-origin issue (workers need
// the URL of the script to be loaded, and dynamically loading a cross-origin
// script does not work).
// PDFJS.disableWorker = true;

// The workerSrc property shall be specified.
PDFJS.workerSrc = '//mozilla.github.io/pdf.js/build/pdf.worker.js';

// Asynchronous download of PDF
var loadingTask = PDFJS.getDocument(url);
  console.log('PDF loaded');
  
  // Fetch the first page
  var pageNumber = 1;
  pdf.getPage(pageNumber).then(function(page) {
    console.log('Page loaded');
    
    var scale = 1.5;
    var viewport = page.getViewport(scale);

    // Prepare canvas using PDF page dimensions
    var canvas = document.getElementById('the-canvas');
    var context = canvas.getContext('2d');
    canvas.height = viewport.height;
    canvas.width = viewport.width;

    // Render PDF page into canvas context
    var renderContext = {
      canvasContext: context,
      viewport: viewport
    };
    var renderTask = page.render(renderContext);
    renderTask.then(function () {
      console.log('Page rendered');
    });
  });
}, function (reason) {
  // PDF loading error
  console.error(reason);
});	
</script>	
<canvas id="the-canvas" style=" position: absolute;top:0;bottom: 0;left: 0;right: 0;margin:auto;"></canvas>

Snuffleupagus:
*As mentioned the last time that I touched this particular part of the font code, I'm sincerely hope that this doesn't cause any regressions!*

However, the patch passes all tests added in PRs #5770, #6270, and #7904 (and obviously all other tests as well). Furthermore, I've manually checked all the issues/bugs referenced in those PRs without finding any issues.

Fixes #8480.


brendandahl (approved):
Looks good!

PR #8508, opened 2017-06-10
Fixes duplicate creation of PDFWorker for the same port.

### LINKED ISSUE #8246: Error: Cannot resolve callback 1 when opening multiple pdfs

[pdf-sample.pdf](https://github.com/mozilla/pdf.js/files/903097/pdf-sample.pdf)
[compressed.tracemonkey-pldi-09.pdf](https://github.com/mozilla/pdf.js/files/903096/compressed.tracemonkey-pldi-09.pdf)

Configuration:
- Web browser and its version: Chrome 56.0.2924.87
- Operating system and its version: OSX 10.11.6
- PDF.js version: pdfjs-dist 1.7.414
- Is an extension: No

Using following React component:

```jsx
import React, { Component } from 'react';
import PDFJS from 'pdfjs-dist/webpack';

class App extends Component {
  constructor() {
    super();
    this.state = {pdfs: []};
    this.handleInputChange = this.handleInputChange.bind(this);
  }

  handleInputChange(event) {
    const file = event.target.files[0];
    const reader = new FileReader();
    const self = this;
    reader.onload = function() {
      const typedArray = new Uint8Array(this.result);
      PDFJS.getDocument(typedArray).then(pdf => {
        self.setState({pdfs: self.state.pdfs.concat(pdf)});
      });
    };
    reader.readAsArrayBuffer(file);
  }

  render() {
    return (
      <div>
        <input type="file" onChange={this.handleInputChange}/>
        <ul>
          {this.state.pdfs.map((pdf, i) => <li key={i}>{pdf.numPages}</li>)}
        </ul>
      </div>
    );
  }
}
```

1. Upload first attached PDF
2. Upload second attached PDF

What is the expected behavior? Both PDFs' page counts are printed:

image

What went wrong? The expected behavior occurs, but PDF.js throws the following error when a second PDF is opened, and will throw the error again whenever a new PDF is opened:

```
Error: Cannot resolve callback 1
pdf.js:340     at error (http://localhost:3000/static/js/bundle.js:33026:18)
    at MessageHandler.messageHandlerComObjOnMessage (http://localhost:3000/static/js/bundle.js:33732:10)
```

(specifically, from line https://github.com/mozilla/pdf.js/blob/59392fd5442d819879dc7ec0a0a8ce9a2fbd2f71/src/shared/util.js#L1291)

This appears unrelated to the contents of the PDFs, this occurs with any sequence of multiple PDFs I've tried other than opening the same file twice.

If I call `pdf.destroy()` on the already-open PDF before opening another, this does not occur.

yurydelendik:
Fixes #8246

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -264,7 +264,7 @@ function getDocument(src, pdfDataRangeTransport,
     // Worker was not provided -- creating and owning our own. If message port
     // is specified in global settings, using it.
     var workerPort = getDefaultSetting('workerPort');
-    worker = workerPort ? new PDFWorker(null, workerPort) : new PDFWorker();
+    worker = workerPort ? PDFWorker.fromPort(workerPort) : new PDFWorker();
     task._worker = worker;
   }
   var docId = task.docId;
@@ -1160,7 +1160,7 @@ class LoopbackPort {
  * @class
  */
 var PDFWorker = (function PDFWorkerClosure() {
-  var nextFakeWorkerId = 0;
+  let nextFakeWorkerId = 0;
 
   function getWorkerSrc() {
     if (typeof workerSrc !== 'undefined') {
@@ -1177,7 +1177,7 @@ var PDFWorker = (function PDFWorkerClosure() {
     error('No PDFJS.workerSrc specified');
   }
 
-  var fakeWorkerFilesLoadedCapability;
+  let fakeWorkerFilesLoadedCapability;
 
   // Loads worker code into main thread.
   function setupFakeWorkerGlobal() {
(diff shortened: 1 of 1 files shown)


src/display/api.js
```diff
@@ -1153,6 +1153,8 @@ class LoopbackPort {
   }
 }
 
+var pdfWorkerPorts = new WeakMap();
```
Snuffleupagus: Nit: Can we use `let` instead when adding new code?

Snuffleupagus (approved):
Seems fine to me, thanks!

Feel free to land this, with passing tests (and optional nit addressed).

PR #8515, opened 2017-06-12
Adds special case for origin of blob to the compatibility URL.

### LINKED ISSUE #7089: Problem streaming PDF on IPad
Hello,

I'm have an application that is streaming a pdf from server to client and then displaying it using pdf.js.  I finally got it working on desktop browsers but on the Ipad I get the error "file origin does not match viewwer's".  Here is the javascript code which I am using to pull the pdf stream (FileStreamResult) from the server and then load it to pdf.js:

```
    var xhr = new XMLHttpRequest();
    xhr.open('GET', '/DataProfile/GetPdf?projectId=M2419', true);
    xhr.responseType = 'blob';

    xhr.onload = function (e) {
        if (this.status == 200) {
            var blob = new Blob([this.response], { type: 'application/pdf' });
            var url = URL.createObjectURL(blob);

            var viewerUrl = '/pdfviewer/viewer.html?file=' + encodeURIComponent(url);
            $("#frmMain").attr("src", viewerUrl);
        }
    };

    xhr.send();
```

I'm guessing that for some reason the createObjectURL method is generating a url with a different origin even though it works just fine on all desktop browsers.  Any help would be appreciated.

yurydelendik:
Fixes #7089

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -1812,6 +1812,14 @@ PDFJS.compatibilityChecked = true;
         case 'javascript':
         case 'mailto':
           return 'null';
+        case 'blob':
+          // Special case of blob: -- returns valid origin of _schemeData.
+          try {
+            return new JURL(this._schemeData).origin || 'null';
+          } catch (_) {
+            // Invalid _schemeData origin -- ignoring errors.
+          }
+          return 'null';
       }
       host = this.host;
       if (!host) {


src/shared/compatibility.js
```diff
@@ -1812,6 +1812,14 @@ PDFJS.compatibilityChecked = true;
         case 'javascript':
         case 'mailto':
           return 'null';
+        case 'blob':
+          // Special case of blob: -- returns valid origin of _schemaData.
+          try {
+            return new JURL(this._schemaData).origin || 'null';
```
Snuffleupagus: Don't you mean this._schemeData here, since the above property is always undefined (applies in the comment below as well)!?

Snuffleupagus (approved):
This looks perfectly reasonable to me, but I haven't got the necessary hardware to test that it actually fixes the referenced issue.

@yurydelendik If you've successfully tested this, please feel free to land the patch!

yurydelendik:
> but I haven't got the necessary hardware to test that it actually fixes the referenced issue.

I tried using polyfill with blob URL on IE9, and it worked as expected.

PR #7394, opened 2016-06-09
Skip inherited props in object iteration

alexander-dmitriev:
Caused 'Error during font loading: this._map[charCode].charCodeAt is not a function'

Sample file: https://yadi.sk/d/xgAl_jmCsN3fA

Issue #7064, opened 2016-03-04
JS error when loading fonts in IE10/11


tonyjin:
Issue: Some documents are not properly loading in IE 10/11. Test file: https://isotropic.org/papers/chicken.pdf. I could not reproduce on the generic viewer, but could when using a separate app built on PDF.js.

Seems to be caused by:
https://github.com/mozilla/pdf.js/blob/8402c791711a1dd0a66021cc02e97e455194d8b1/src/core/fonts.js#L234

the "for (var charCode in this._map)" is not checking for hasOwnProperty when iterating. Debugging in IE10 / IE11 shows an attempt to check this.map["find"].charCodeAt(0).

tonyjin:
Doesn't look like the ToUnicodeMap code changed recently - maybe the contents or type of the map it's iterating over changed and the "for...in" no longer performs expected behavior.

tonyjin:
Found the issue - we are polyfilling Array.prototype.find. I'll submit a pull request to add a hasOwnProperty check which seems like a good idea regardless.

yurydelendik:
`find` shall not be enumerable on the Array objects. Are you using some library that make this property enumerable?

(Adding hasOwnProperty check will not be acceptable due to amount of item ToUnicodeMap has to deal with)

Snuffleupagus:
> I'll submit a pull request to add a hasOwnProperty check which seems like a good idea regardless.

That was rejected previously, for performance reasons, please see https://github.com/mozilla/pdf.js/pull/5411#issuecomment-59098223.

yurydelendik:
>  we are polyfilling Array.prototype.find

When you are polyfilling, use `Object.defineProperty(Array.prototype, "find", { value: ..., enumeratable: false, configurable: true });` instead of `Array.prototype.find = ...`

tonyjin:
Ah, thanks for the tip. I'll close this. We were using this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Polyfill

PR #8516, opened 2017-06-12
Fix blank font loading issue.

githubBrandon:
If a pdf (particularly in the form of an array buffer) has missing fonts, no fonts will display. This fixes that.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -225,7 +225,8 @@ var ToUnicodeMap = (function ToUnicodeMapClosure() {
 
     forEach(callback) {
       for (var charCode in this._map) {
-        callback(charCode, this._map[charCode].charCodeAt(0));
+        if (typeof this._map[charCode].charCodeAt === 'function')
+          callback(charCode, this._map[charCode].charCodeAt(0));
       }
     },
 


src/core/fonts.js
```diff
@@ -225,7 +225,8 @@ var ToUnicodeMap = (function ToUnicodeMapClosure() {
 
     forEach(callback) {
       for (var charCode in this._map) {
-        callback(charCode, this._map[charCode].charCodeAt(0));
+        if (typeof this._map[charCode].charCodeAt === 'function')
```
yurydelendik: why not `typeof this._map[charCode] === 'string'` ?

yurydelendik (changes requested):
It's hard to understand the fix. The patch is missing the test and description does not provide an explanation for the fix. Things to be addressed: provide a comment why `=== 'function'` check is needed and add a test.

Snuffleupagus:
> It's hard to understand the fix. The patch is missing the test and description does not provide an explanation for the fix. Things to be addressed: provide a comment why === 'function' check is needed and add a test.

Besides the above, won't this suffer from similar performance issues as issues/PRs #4183, #5411, #7064, #7394 (and probably more); which we have previously always rejected?

src/core/fonts.js
githubBrandon: The error is `This._map[charCode].charCodeAt is not a function`. When seeing that error, no fonts are loaded. When this line is added, the pdf renders perfectly. I thought dealing with the error directly would be better, but I can test if `typeof this._map[charCode] === 'string'` works.

githubBrandon:
@Snuffleupagus I'm not sure in what way this relates to those. A simple type check should not have a noticeable impact on performance. This isn't changing prototypes for calling any helper functions.

yurydelendik:
> A simple type check should not have a noticeable impact on performance

Extra property check in the hot loops add a performance penalty. There is no reason to do that, unless there is a really good reason.

Snuffleupagus:
> I'm not sure in what way this relates to those. A simple type check should not have a noticeable impact on performance. This isn't changing prototypes for calling any helper functions.

In the issues/PRs I refereed to, the problem has been the use of JavaScript Libraries/Frameworks which adds custom properties/methods to Arrays in such a way that they become `enumerable` (which is *not* correct).
Hence these kind of added checks *should not* be necessary, and they will definitely have a non-zero performance cost!

I think the relevant question here is what the values of `charCode` and `this._map[charCode]` are when things break?

src/core/fonts.js
yurydelendik: > When seeing that error, no fonts are loaded... I thought dealing with the error directly would be better

I'm thinking dealing with the error directly will slow down font parsing and will make PDF.js slower.

timvandermeij:
Closing since the pull request cannot be merged in the current state. No example PDF file has been provided that is fixed by this patch, the performance may be degraded by the patch and it's not clear why we even need such a check. If more information can be provided regarding these problems, we can consider reopening it.

Issue #8476, opened 2017-06-02
Strange PDF display behavior 


Todai88:
## Important **EDIT**:

I have now tried picking up the .PDF from a url inside of the project. Error still occurs...

## Here is a SO thread about the problem with all code:
[https://stackoverflow.com/questions/44326665/strange-behavior-when-displaying-pdf-in-html](https://stackoverflow.com/questions/44326665/strange-behavior-when-displaying-pdf-in-html)

## Configuration:
- Web browser and its version: **Chrome Version 58.0.3029.110 (64-bit)**
- Operating system and its version: ** Windows 10 **
- PDF.js version:  Latest Stable
- Is an extension:

## Steps to reproduce the problem:

As this is using a unsecured REST API I can't really tell you how to reproduce the error.

But basically what's happening is that I'm using AngularJS $http library to grab a JSON Array from said API. One of the fields of the JSONObjects in the Array is a `varbinary(max)` reprentation of a PDF which in Angular gets decoded into a string (according to `typeof`). 

**Code:**

This is how I pull out the data using Angular: 

    $http.post("http://some-example.com/api/PDF/Get", JSON.stringify(vm.content))
    	.then(function (response) {
    		//success
    
    		vm.data = response.data;
    		$sce.trustAsResourceUrl(vm.data[0].Pdf);
    	},
    	function (error) { 
    
    	})
    	.finally(function () {
    		let pdf = atob(vm.data[0].Pdf);
    		let loadingTask = PDFJS.getDocument({ data: pdf }); 
    
    		loadingTask.promise.then(function (pdf) { 
    
    			let pageNumber = 1;
    			pdf.getPage(pageNumber).then(function (page) {
    				let scale = 1;
    				let viewport = page.getViewport(scale);
    
    				let canvas = document.getElementById('test-canvas');
    				let context = canvas.getContext('2d');
    				canvas.height = viewport.height;
    				canvas.width = viewport.width;
    
    				let renderContext = {
    					canvasContext: context,
    					viewport: viewport
    				};
    				let renderTask = page.render(renderContext);
    				renderTask.then(function () { 
    					// Wait for rendering to finish
    					renderTask.promise.then(function () {
    						pageRendering = false;
    						if (pageNumPending !== null) {
    							// New page rendering is pending
    							renderPage(pageNumPending);
    							pageNumPending = null;
    						}
    					});
    				}); 
    			});
    		});
    	});

[...]

PR #8519, opened 2017-06-12
Preventing from using the same canvas for multiple render()

### LINKED ISSUE #5636: Error when the same canvas is used for two render operations
To illuminate the common mistake: throw an exception if the canvas/context is used to start the rendering when other one with the same context is in progress

... and probably modify hello work example to e.g. print console message to indicate when rendering is finished.

yurydelendik:
Attempt to make life less confusing (#8476)

Fixes #5636

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -901,7 +901,7 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
         stats.time('Rendering');
         internalRenderTask.initializeGraphics(transparency);
         internalRenderTask.operatorListChanged();
-      }, complete);
+      }).catch(complete);
 
       return renderTask;
     },
@@ -2103,6 +2103,7 @@ var RenderTask = (function RenderTaskClosure() {
  * @ignore
  */
 var InternalRenderTask = (function InternalRenderTaskClosure() {
+  let canvasInRendering = new WeakMap();
 
   function InternalRenderTask(callback, params, objs, commonObjs, operatorList,
                               pageNumber, canvasFactory) {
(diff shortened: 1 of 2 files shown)


src/display/api.js
```diff
@@ -2125,13 +2126,25 @@ var InternalRenderTask = (function InternalRenderTaskClosure() {
...
+    this._canvas = params.canvasContext.canvas;
   }
 
   InternalRenderTask.prototype = {
 
     initializeGraphics:
         function InternalRenderTask_initializeGraphics(transparency) {
 
+      if (this._canvas) {
+        console.log(this._canvas);
```
timvandermeij: Nit: remove the left-over debug print

test/unit/api_spec.js
```diff
@@ -1041,6 +1041,32 @@ describe('api', function() {
...
+        viewport,
+      });
+      var renderTask2 = page.render({
+        canvasContext: canvasAndCtx.context,
+        viewport,
+      });
+
+      Promise.all([
+        renderTask1.promise,
+        renderTask2.then(() => {
```
Snuffleupagus: Nit: `renderTask2.promise.then` for consistency?

src/display/api.js
```diff
@@ -2125,13 +2126,24 @@ var InternalRenderTask = (function InternalRenderTaskClosure() {
...
 
   InternalRenderTask.prototype = {
 
     initializeGraphics:
         function InternalRenderTask_initializeGraphics(transparency) {
 
+      if (this._canvas) {
+        if (canvasInRendering.has(this._canvas)) {
+          throw new Error(
+            'Cannot use the same canvas during multiple render() operations.' +
```
Snuffleupagus: Nit: add space at the end of this string, to avoid it reading `... operations.Use different...`.

Snuffleupagus (approved):

PR #8522, opened 2017-06-13
Fixes WeakMap polyfill (and improves PDFWorker port check).

yurydelendik:
Regressed on IE9 after #8508 -- WeakMap polyfill did not work per spec.

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -1176,10 +1176,14 @@ PDFJS.compatibilityChecked = true;
   }
   WeakMap.prototype = {
     has(obj) {
+      if ((typeof obj !== 'object' && typeof obj !== 'function') ||
+          obj === null) {
+        return false;
+      }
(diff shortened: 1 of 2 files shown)


src/shared/compatibility.js
```diff
@@ -1176,10 +1176,13 @@ PDFJS.compatibilityChecked = true;
   }
   WeakMap.prototype = {
     has(obj) {
+      if (typeof obj !== 'object' || obj === null) {
```
Snuffleupagus: Unless I'm misunderstanding the examples at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#Examples, it appears that it should also work if `obj` is of type `function`.
So, do we need to add a `typeof obj !== 'function'` check as well here?

Snuffleupagus (approved):
Looks good to me, thank you!

PR #8535, opened 2017-06-17
Reset `pageRotation` and `downloadComplete` in `PDFViewerApplication.close`

Snuffleupagus:
Currently, these properties are reset in what appears to be somewhat arbitrary locations (within the `load` and `open` methods respectively). The explanation is probably that both of these properties predates the existence of any centralized clean-up code in the viewer.

Hence I think that it makes sense to move the resetting of these properties to the `close` method, since that improves the overview of what's actually cleaned-up/reset when changing documents in the viewer.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -136,6 +136,7 @@ var PDFViewerApplication = {
   l10n: null,
   pageRotation: 0,
   isInitialViewSet: false,
+  downloadComplete: false,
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good. Nice clean-up!

PR #8536, opened 2017-06-17
Refactor `ObjectLoader` to use `Dict`s correctly, rather than abusing their internal properties

Snuffleupagus:
These patches address something that I think is incorrect, or at least unfortunate, in the current `ObjectLoader` implementation. *Please refer to the individual commit messages for additional details.*


timvandermeij (approved):

timvandermeij:
Awesome refactoring work, thanks!

PR #8539, opened 2017-06-18
Remove a redundant `PDFViewer.currentScale` call from `PDFViewerApplication.load`

Snuffleupagus:
Since this call occurs *before* the `PDFViewer.setDocument` call, it won't actually cause any scale change.
Furthermore, moving it should not be necessary, since the `scale` is already used as the fallback case in `PDFViewerApplication.setInitialView` (provided it's non-zero, which isn't even the case in the default viewer).

Hence this patch should cause no functional changes at all, since it simply removes a piece of unnecessary code.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -907,7 +907,6 @@ var PDFViewerApplication = {
     this.pdfDocumentProperties.setDocument(pdfDocument, this.url);
 
     let pdfViewer = this.pdfViewer;
-    pdfViewer.currentScale = scale;
     pdfViewer.setDocument(pdfDocument);
     let firstPagePromise = pdfViewer.firstPagePromise;
     let pagesPromise = pdfViewer.pagesPromise;


timvandermeij (approved):

timvandermeij:
Nice find!

Snuffleupagus:
It unfortunately appears that removing this one line uncovered existing issues elsewhere the code :-(

For example, if you load [http://www.texample.net/media/pgf/builds/pgfmanualCVS2012-11-04.pdf#disableStream=true&disableAutoFetch=true](http://www.texample.net/media/pgf/builds/pgfmanualCVS2012-11-04.pdf#disableStream=true&disableAutoFetch=true).[1]
And then *before* the first page has begun rendering, either open the sidebar *or* change the scale to one of the predefined ones (`page-fit`/`page-width`/`page-actual`).

The pre-existing issue is that `PDFViewer` allows you to set various parameters; e.g. [`currentPageNumber`](https://github.com/mozilla/pdf.js/blob/054fe1392078078e9d2d71879ed6a73d2202ed50/web/pdf_viewer.js#L168-L171), [`currentScale`](https://github.com/mozilla/pdf.js/blob/054fe1392078078e9d2d71879ed6a73d2202ed50/web/pdf_viewer.js#L245-L249), [`currentScaleValue`](https://github.com/mozilla/pdf.js/blob/054fe1392078078e9d2d71879ed6a73d2202ed50/web/pdf_viewer.js#L264-L268), and [`pagesRotation`](https://github.com/mozilla/pdf.js/blob/054fe1392078078e9d2d71879ed6a73d2202ed50/web/pdf_viewer.js#L288-L290); *before* the `setDocument` method has been called.
In particular, calling `PDFViewer.currentScaleValue` before setting the `pdfDocument` is especially problematic. For example, resizing the viewer (which is what opening the sidebar amounts to), may result in `PDFViewer.currentScaleValue` calls before the `pdfDocument` has loaded.

 1. Calling `PDFViewer.currentPageNumber` before `setDocument` first of all cannot run the [necessary validation](https://github.com/mozilla/pdf.js/blob/054fe1392078078e9d2d71879ed6a73d2202ed50/web/pdf_viewer.js#L188-L192) of the value provided. Secondly, depending of whether or not the `defaultRenderingQueue` is used, there's a real risk that `currentPageNumber` won't be consistent with the actual rendered page after `setDocument` is called.
I'm not certain that there's any simply solution to the above, and it's also a real risk that trying to add `this._currentPageNumber` validation code to `setDocument` could become quite messy!

 2. Calling `PDFViewer.pagesRotation` before `setDocument` will currently be ignored when creating `PDFPageView` instances, which means that the rotation in `PDFViewer` would be *inconsistent* with the one in `PDFPageView`.
This is the only point that would be really easy to address, I've even got a WIP patch (see https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:viewer-components-rotation).

3. Calling `PDFViewer.currentScale` before `setDocument` is the only thing that currently works.

4. Calling `PDFViewer.currentScaleValue` before `setDocument` does not, and cannot work, since it means that `_currentScale` and `_currentScaleValue` will *not* be consistent.
Normally, when calling `PDFViewer.currentScaleValue` the [`_setScale`](https://github.com/mozilla/pdf.js/blob/054fe1392078078e9d2d71879ed6a73d2202ed50/web/pdf_viewer.js#L511) method will compute the correct `scale` corresponding to a predefined scale string. However, that cannot be done if `pdfDocument` is undefined, since it relies on the pages already being available, and we *need* to know the `currentScale` when creating the `PDFPageView` instances, in [web/pdf_viewer.js#L354](https://github.com/mozilla/pdf.js/blob/054fe1392078078e9d2d71879ed6a73d2202ed50/web/pdf_viewer.js#L354).

So, to try and summarize, there's a couple of methods that simply does not work if called *before* `setDocument`.
What I'm actually wondering is why we even need to support setting those properties at all, since if we just did `if (!this.pdfDocument) { return; }` in `currentPageNumber`/`currentScale`/`currentScaleValue`/`pagesRotation`, none of these issues would exist!?

Also, in the [`simpleviewer` example](https://github.com/mozilla/pdf.js/blob/054fe1392078078e9d2d71879ed6a73d2202ed50/examples/components/simpleviewer.js#L53-L54), we're showing how e.g. the scale can be changed *after* `setDocument` is called (since the `pagesinit` event is dispatched [at the end of `setDocument`](https://github.com/mozilla/pdf.js/blob/054fe1392078078e9d2d71879ed6a73d2202ed50/web/pdf_viewer.js#L403)).

A compromise would perhaps be to only allow, `pagesRotation` and `currentScale` if `this.pdfDocument` isn't defined yet, but the inconsistency doesn't seem very user friendly to me!

---

@timvandermeij, @yurydelendik How do you think that we should proceed here, to solve the underlying issue this PR uncovered?
Can we simply `return` early in the various setters discussed above if `setDocument` hasn't been called, or do need to try and support at least some of them?

---
[1] The debugging parameters are merely intended to slow down loading somewhat, such that you have more time to do the next step.

timvandermeij:
I personally think we should return early and not allow setting any of these properties before `setDocument` is called. Doing so doesn't make sense in my opinion and it complicates the code a lot, leading to these kind of subtle issues that are hard to find beforehand. I'm not sure why one would want to set these properties before actually setting a document to apply them on, so I'm more than fine with disallowing that in favor of cleaner and more predictable code.

PR #8540, opened 2017-06-18
Reduce memory requirements of pdf2svg.js example to avoid OOM 

### LINKED ISSUE #8534: SVG: 'FATAL ERROR: JavaScript heap out of memory'
Steps to reproduce (file has 60MB - can not be attached to this issue):
http://doppelbauer.name/FatalProcessOutOfMemory.pdf

````
npm install
gulp dist-install
node examples/node/pdf2svg.js /tmp/FatalProcessOutOfMemory.pdf
````

Console (simplified):
````
# Document Loaded
Number of Pages: 302

# Page 1
Size: 793x594

Warning: Unimplemented operator dependency

[...snip...]

# Page 128
Size: 793x594

Warning: Unimplemented operator dependency

<--- Last few GCs --->

   81662 ms: Mark-sweep 1389.1 (1431.5) -> 1388.7 (1432.5) MB, 66.2 / 0.0 ms [allocation failure] [GC in old space requested].
   81735 ms: Mark-sweep 1388.7 (1432.5) -> 1388.7 (1399.5) MB, 73.6 / 0.0 ms [last resort gc].
   81797 ms: Mark-sweep 1388.7 (1399.5) -> 1388.7 (1399.5) MB, 62.1 / 0.0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x12951c0cfb39 <JS Object>
    1: xmlEncode [/tmp/pdf.js/examples/node/domstubs.js:~4] [pc=0xfd18a29a38f] (this=0x23692a476751 <JS Global Object>,s=0x25e56998d669 <Very long string[5286062]>)
    2: setAttributeNS(aka DOMElement_setAttributeNS) [/tmp/pdf.js/examples/node/domstubs.js:~80] [pc=0xfd18a2416f3] (this=0x25e56998e911 <a DOMElement with map 0x3340bbd0b479>,NS=0x2c0bba00f919 <String[28]: http://www.w3.org/1999/xl...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [node]
 2: 0x10b835c [node]
 3: v8::Utils::ReportApiFailure(char const*, char const*) [node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [node]
 5: v8::internal::Factory::NewRawOneByteString(int, v8::internal::PretenureFlag) [node]
 6: v8::internal::String::SlowFlatten(v8::internal::Handle<v8::internal::ConsString>, v8::internal::PretenureFlag) [node]
 7: v8::internal::String::Flatten(v8::internal::Handle<v8::internal::String>, v8::internal::PretenureFlag) [node]
 8: v8::internal::Runtime_StringCharCodeAtRT(int, v8::internal::Object**, v8::internal::Isolate*) [node]
 9: 0xfd189f079a7
````

### PR DESCRIPTION
**Pull request by Rob--W**
See the individual commit messages for more details. I verified that the output of the first 9 pages (before and after my patches) are identical:

```
$ gdb -ex run --args \
 node --max_old_space_size=200 \
 examples/node/pdf2svg.js /tmp/FatalProcessOutOfMemory.pdf

$ md5sum svgdump/FatalProcessOutOfMemory-[1-9].svg
8d3b9b5ff4bffef8423346f3b964a96f  svgdump/FatalProcessOutOfMemory-1.svg
0c00d37793523580bb45af64cf0b0510  svgdump/FatalProcessOutOfMemory-2.svg
2efeb696ec79a5c2e1318ee8d6c26d78  svgdump/FatalProcessOutOfMemory-3.svg
898f49bbda623137f5edc985bfafcac3  svgdump/FatalProcessOutOfMemory-4.svg
00628f7ebc47afbbc2992b85da7a0828  svgdump/FatalProcessOutOfMemory-5.svg
dec693597bc61286d10b4f0af8b468ea  svgdump/FatalProcessOutOfMemory-6.svg
6c047f815efd2816ace541a23487f48f  svgdump/FatalProcessOutOfMemory-7.svg
f702367a6b8903273a42277a12d40d2d  svgdump/FatalProcessOutOfMemory-8.svg
cd1867d8cb8b88b5d16e3c31b51ef2b5  svgdump/FatalProcessOutOfMemory-9.svg
```

(note that if you wait until the PDF-to-SVG conversion finishes for the given example from #8534, that you end up with a svgdump directory of 1.5G.)

Fixes #8534

### DIFF
--- a/examples/node/domstubs.js
+++ b/examples/node/domstubs.js
@@ -91,24 +91,27 @@ DOMElement.prototype = {
   },
 
   toString: function DOMElement_toString() {
-    var attrList = [];
-    for (i in this.attributes) {
-      attrList.push(i + '="' + xmlEncode(this.attributes[i]) + '"');
+    var buf = [];
+    buf.push('<' + this.nodeName);
+    if (this.nodeName === 'svg:svg') {
+      buf.push(' xmlns:xlink="http://www.w3.org/1999/xlink"' +
+               ' xmlns:svg="http://www.w3.org/2000/svg"');
+    }
+    for (var i in this.attributes) {
+      buf.push(' ' + i + '="' + xmlEncode(this.attributes[i]) + '"');
     }
 
+    buf.push('>');
+
     if (this.nodeName === 'svg:tspan' || this.nodeName === 'svg:style') {
-      var encText = xmlEncode(this.textContent);
-      return '<' + this.nodeName + ' ' + attrList.join(' ') + '>' +
-             encText + '</' + this.nodeName + '>';
-    } else if (this.nodeName === 'svg:svg') {
-      var ns = 'xmlns:xlink="http://www.w3.org/1999/xlink" ' +
-               'xmlns:svg="http://www.w3.org/2000/svg"'
-      return '<' + this.nodeName + ' ' + ns + ' ' + attrList.join(' ') + '>' +
-             this.childNodes.join('') + '</' + this.nodeName + '>';
+      buf.push(xmlEncode(this.textContent));
     } else {
-      return '<' + this.nodeName + ' ' + attrList.join(' ') + '>' +
-             this.childNodes.join('') + '</' + this.nodeName + '>';
+      this.childNodes.forEach(function(childNode) {
+        buf.push(childNode.toString());
+      });
     }
+    buf.push('</' + this.nodeName + '>');
+    return buf.join('');
   },
 
   cloneNode: function DOMElement_cloneNode() {
--- a/examples/node/pdf2svg.js
+++ b/examples/node/pdf2svg.js
@@ -18,7 +18,7 @@ var pdfPath = process.argv[2] || '../../web/compressed.tracemonkey-pldi-09.pdf';
 var data = new Uint8Array(fs.readFileSync(pdfPath));
 
 // Dumps svg outputs to a folder called svgdump
-function writeToFile(svgdump, pageNum) {
+function writeToFile(svgdump, pageNum, callback) {
   var name = getFileNameFromPath(pdfPath);
   fs.mkdir('./svgdump/', function(err) {
     if (!err || err.code === 'EEXIST') {
@@ -29,7 +29,10 @@ function writeToFile(svgdump, pageNum) {
           } else {
             console.log('Page: ' + pageNum);
           }
+          callback();
         });
+    } else {
+      callback();
     }
   });
 }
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by doublex** [comment]
@Rob--W 
Works great! Do you want $50 bounty?

**Comment by Rob--W** [comment]
@doublex Sure, why not. I already appreciate the good bug reports from you, thanks for those!

File: examples/node/domstubs.js [hunk 1]
```diff
@@ -91,24 +91,26 @@ DOMElement.prototype = {
...
-      return '<' + this.nodeName + ' ' + attrList.join(' ') + '>' +
-             encText + '</' + this.nodeName + '>';
-    } else if (this.nodeName === 'svg:svg') {
-      var ns = 'xmlns:xlink="http://www.w3.org/1999/xlink" ' +
-               'xmlns:svg="http://www.w3.org/2000/svg"'
-      return '<' + this.nodeName + ' ' + ns + ' ' + attrList.join(' ') + '>' +
-             this.childNodes.join('') + '</' + this.nodeName + '>';
+      str += xmlEncode(this.textContent);
     } else {
-      return '<' + this.nodeName + ' ' + attrList.join(' ') + '>' +
```
yurydelendik: I'm thinking the issue was only here: we where doubling size of `this.childNodes.join('')`. I'm thinking that replacing `str += ` with `buf.push()` and `bug.join('')` at the end will free more memory.

yurydelendik (commented):
Now, that's is confusing. Less memory garbage is generated when `buf.push();` and `buf.join()` is used instead of `s += ...` (that's true in other languages too e.g. Java/C#)

examples/node/domstubs.js
Rob--W: The issue is both in the attributes and the childNodes list. When I first changed `attrList` to `+=`, I was able to get 9 additional pages, whereas doing the full conversion gave only one more (on top of the 9). I haven't tested `childNodes` in isolation, but that is probably not too relevant.

A JS string in V8 can internally be represented in multiple ways. One of the representations is a sequence of substrings, which is especially useful in use cases involving concatenation, like this one. This string will be flattened at some point (e.g. when you read individual characters of the string).

Another reason for the current state of the patch (besides the boost in (peak) memory usage) is that it becomes easier to implement svg-to-file streaming (like a readable stream). It is conceptually as simple as replacing `+= extraData` with `.write(extraData)`.

yurydelendik: > It is conceptually as simple as replacing += extraData with .write(extraData)

Yeah, the extent of the current changes is good. I just want to replace `str +=` to `buf.push()` verbatim, which will fit the agenda above.

> One of the representations is a sequence of substrings...

I understand. But we can produce really large string and I don't want JS JITs to make the call how to handle this data internally.

Rob--W: I applied the changes. Now only 19 pages are rendered before crashing (instead of 20 with `+=`), but the intent is more clear now.

yurydelendik:
Thank you for the patch.

PR #8542, opened 2017-06-19
Move svg:clipPath generation from clip to endPath

### LINKED ISSUE #8527: SVG: TypeError: Cannot read property 'cloneNode' of undefined
Steps to reproduce, attached file: 
[cloneNode-undefined.pdf](https://github.com/mozilla/pdf.js/files/1074501/cloneNode-undefined.pdf)

````
npm install
gulp dist-install
node examples/node/pdf2svg.js /tmp/cloneNode-undefined.pdf
````

Console:
````
# Document Loaded
Number of Pages: 65

# Page 1
Size: 787.9433x525.2961

Error: TypeError: Cannot read property 'cloneNode' of undefined
````

### PR DESCRIPTION
**Pull request by Rob--W**
See the commit message for a detailed explanation.

I think that we also need to check whether it is necessary to add more `.endPath` calls to src/display/svg.js, similar to the many `consumePath` calls in src/display/canvas.js. However I don't understand that part of the code well enough to tell whether it's needed.

Fixes #8527 
Seems to improve #8496 (https://github.com/mozilla/pdf.js/issues/8496#issuecomment-309432026)

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -361,6 +361,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
     this.extraStack = [];
     this.commonObjs = commonObjs;
     this.objs = objs;
+    this.pendingClip = null;
     this.pendingEOFill = false;
 
     this.embedFonts = false;
@@ -389,6 +390,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
       this.transformMatrix = this.transformStack.pop();
       this.current = this.extraStack.pop();
 
+      this.pendingClip = null;
       this.tgrp = null;
     },
 
@@ -894,9 +896,10 @@ SVGGraphics = (function SVGGraphicsClosure() {
       current.setCurrentPoint(x, y);
     },
 
-    endPath: function SVGGraphics_endPath() {},
-
-    clip: function SVGGraphics_clip(type) {
+    endPath: function SVGGraphics_endPath() {
+      if (!this.pendingClip) {
+        return;
+      }
       var current = this.current;
       // Add current path to clipping path
       var clipId = 'clippath' + clipCount;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/display/svg.js [hunk 1]
```diff
@@ -264,6 +264,7 @@ var SVGExtraState = (function SVGExtraStateClosure() {
     this.dependencies = [];
 
     // Clipping
+    this.pendingClip = null;
```
timvandermeij: It does not look like you store the pending clip in the `SVGExtraState` object (but instead as a property of `SVGGraphics`), so I think this should be moved to above https://github.com/mozilla/pdf.js/pull/8542/files?w=1#diff-84f753e2bb39669e059ce3447e5a0057R365, just like in `canvas.js`.

src/display/svg.js
```diff
@@ -894,36 +896,41 @@ SVGGraphics = (function SVGGraphicsClosure() {
       current.setCurrentPoint(x, y);
     },
 
-    endPath: function SVGGraphics_endPath() {},
+    endPath: function SVGGraphics_endPath() {
+      if (this.pendingClip) {
```
timvandermeij: I think we should use this instead:

```javascript
if (!this.pendingClip) {
  return;
}
```

That way we avoid a level of indenting, which improves readability.

timvandermeij (changes requested):

src/display/svg.js
Rob--W: You're right, fixed.

src/display/svg.js
Rob--W: Done too.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #8552, opened 2017-06-20
Ensure canvas is really hidden when used with pdfjs-dist.

yurydelendik:
Avoids black background flickering in such examples as components/simpleviewer.html


Snuffleupagus (approved):
Confirmed locally, thank you for fixing this!

PR #8555, opened 2017-06-21
Don't allow setting various properties, such as `currentPageNumber`/`currentScale`/`currentScaleValue`/`pagesRotation`, before `{PDFViewer, PDFThumbnailViewer}.setDocument` has been called

Snuffleupagus:
Currently a number of these properties do not work correctly if set *before* calling `setDocument`; please refer to the discussion starting in https://github.com/mozilla/pdf.js/pull/8539#issuecomment-309706629.

Rather than trying to have *some* of these methods working, but not others, it seems much more consistent to simply always require that `setDocument` has been called.

/cc @timvandermeij Since your comment in https://github.com/mozilla/pdf.js/pull/8539#issuecomment-309897649 seems to support doing this.
/cc @yurydelendik I know that https://github.com/mozilla/pdf.js/pull/8539#issuecomment-309706629 is a long read, but can you please check that you're OK with that reasoning and this PR?

### DIFF
--- a/web/pdf_thumbnail_viewer.js
+++ b/web/pdf_thumbnail_viewer.js
@@ -95,8 +95,15 @@ class PDFThumbnailViewer {
   }
 
   set pagesRotation(rotation) {
+    if (!(typeof rotation === 'number' && rotation % 90 === 0)) {
+      throw new Error('Invalid thumbnails rotation angle.');
+    }
+    if (!this.pdfDocument) {
+      return;
+    }
     this._pagesRotation = rotation;
-    for (let i = 0, l = this._thumbnails.length; i < l; i++) {
+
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):
Nice work! I think this makes the code much more consistent and predictable (avoiding subtle bugs). Feel free to merge this if @yurydelendik approves the patch too.

Snuffleupagus:
This PR contains a proper fix to the underlying issues uncovered by #8539, please refer to https://github.com/mozilla/pdf.js/pull/8539#issuecomment-309706629 for all the gory details.
@yurydelendik Since you originally added this code, back when the viewer split into components, I just wanted to make sure you're on-board with this change before we land the PR.

yurydelendik (approved):

PR #8556, opened 2017-06-21
Stop tracking the rotation in `PDFViewerApplication` and directly use `PDFViewer.pagesRotation` instead

Snuffleupagus:
Part of the rotation handling code, in what's now `web/app.js`, hasn't really changed since before the viewer was split into multiple files/components.

Similar to other properties, such as current page/scale, we should probably avoid tracking state in multiple places. Hence I'm suggesting that we don't store the rotation in `PDFViewerApplication`, and access the value in `PDFViewer` instead.

Since `PDFViewerApplication.pageRotation` has existed for a very long time, a getter was added to avoid outright breaking third-party code that may depend on it.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -134,7 +134,6 @@ var PDFViewerApplication = {
   eventBus: null,
   /** @type {IL10n} */
   l10n: null,
-  pageRotation: 0,
   isInitialViewSet: false,
   downloadComplete: false,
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #8561, opened 2017-06-22
Don't use a hidden canvas when constructing the `zoomLayer` in `PDFPageView.update`

Snuffleupagus:
*This is an existing issue that I noticed while testing PR #8552.*

When zooming or rotation occurs, we'll try to use the current canvas as a (CSS transformed) preview until the page has been completely re-drawn.
If you manage to change the scale (or rotation) *very* quickly, it's possible that `PDFPageView.update` can be called *before* a previous `render` operation has progressed far enough to remove the `hidden` property from the canvas.

The result is thus that a page may be *entirely* black during zooming or rotation, which doesn't look very good. This effect can be a bit difficult to spot, but it does manifest even in the default viewer.

### DIFF
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -230,7 +230,7 @@ class PDFPageView {
         });
         return;
       }
-      if (!this.zoomLayer) {
+      if (!this.zoomLayer && !this.canvas.hasAttribute('hidden')) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
I verified that the black page does happen before the patch and does not happen anymore after the patch. Thank you!

PR #8567, opened 2017-06-23
Fix error when using custom CMapReaderFactory

ghetolay:
When using a source object on `getDocument()` with a custom `CMapReaderFactory` we end up sending that factory to the worker as part of the message. This causes `postMessage()` to throw because it can't clone a function.

So the param object passed to `_fetchDocument()` shouldn't have that factory as part of his properties.

It's really trivial so I didn't add test but I could build something if you want.


yurydelendik:
Thank you for the patch.

PR #8571, opened 2017-06-24
Mention in the "Getting the Code" section of README.md that an ES6 compatible browser is required when using the default viewer with `gulp server`

Snuffleupagus:
The https://github.com/mozilla/pdf.js/wiki/Contributing article has been updated to explicitly mention that an ES6 browser is now a minimum requirement for development.
Since we recently have seen a couple of issues filed which seemed to indicate that people tried to use PDF.js in browsers without full ES6 support, it's probably a good idea to mention this more prominently in the README as well.


timvandermeij:
Good idea!

Issue #7764, opened 2016-10-28
PDF to SVG opacity


alafr:
**Attach file here:**  [file.pdf](https://github.com/mozilla/pdf.js/files/559611/file.pdf)
**Web browser and its version:** all
**PDF.js version:** 1.6.274, https://mozilla.github.io/pdf.js/build/pdf.js

**What went wrong?**
All paths and shapes with opacity < 1 are rendered opaque
I get in the console the message: `Warning: Unimplemented graphic state ca (and CA)`

I can see there is no use of "fill-opacity", "stroke-opacity" or "opacity" attributes in the `src/display/svg.js` source code. Will this be implemented?

doublex:
+1

timvandermeij:
This pull request fixes the opacity issues tracked in this issue. Note that the only difference with the canvas rendering is then the stroke width, which is tracked in #7767.

PR #8574, opened 2017-06-25
SVG: move setting the stroke properties to the `stroke` method

### LINKED ISSUE #7767: PDF to SVG stroke properties
- Attach file here: [bug-stroke.pdf](https://github.com/mozilla/pdf.js/files/560532/bug-stroke-uncompressed.pdf)
- Web browser and its version: probably all - I use chrome 54.0.2840.71 m (64-bit)
- PDF.js version: 1.6.274 from https://mozilla.github.io/pdf.js/build/pdf.js

What went wrong?
The width, dasharray, linejoin, linecap of the paths and lines are not rendered in the SVG.
I added breakpoints in the javascript debugger and I could see that the missing styling is correctly understood by the [`executeOpTree()`](https://github.com/mozilla/pdf.js/blob/master/src/display/svg.js#L523-L543) function but is not yet defined when needed by [`constructPath()`](https://github.com/mozilla/pdf.js/blob/master/src/display/svg.js#L884-L892) function. Maybe these lines should be moved in the [`stroke()`](https://github.com/mozilla/pdf.js/blob/master/src/display/svg.js#L992) function ?

timvandermeij:
In general, we may not know the stroke properties when path construction happens. Since we must know the properties when we apply the stroke, we should set the properties at that point. Note that we already do that for the color and opacity, but not yet for the other properties.

Fixes #7767 (and makes the PDF file from the linked issue #7764 render perfectly)

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -882,16 +882,6 @@ SVGGraphics = (function SVGGraphicsClosure() {
         }
       }
       current.path.setAttributeNS(null, 'd', d.join(' '));
-      current.path.setAttributeNS(null, 'stroke-miterlimit',
-                                  pf(current.miterLimit));
-      current.path.setAttributeNS(null, 'stroke-linecap', current.lineCap);
-      current.path.setAttributeNS(null, 'stroke-linejoin', current.lineJoin);
-      current.path.setAttributeNS(null, 'stroke-width',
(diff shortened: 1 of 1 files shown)


Snuffleupagus (approved):
Looks good, thank you!

theinhtunthein:
Myanmar

PR #8576, opened 2017-06-26
Remove the `selectScaleOption` helper function, in `Toolbar._updateUIState`, and simply inline its code instead

Snuffleupagus:
After PR #8510, we now always lookup the localized `page_scale_percent` string to prevent any possible ordering issues. Since the scaleSelect dropdown is updated asynchronous, there's really no point in having a helper function any more, hence this code can rather be placed inline in `Toolbar._updateUIState`.

### DIFF
--- a/web/toolbar.js
+++ b/web/toolbar.js
@@ -96,9 +96,8 @@ class Toolbar {
   }
 
   _bindListeners() {
-    let eventBus = this.eventBus;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #8580, opened 2017-06-28
Fix how we detect and handle missing glyph data.

### LINKED ISSUE #7696: PDF file containing hyeroglyphical characters: cmap Adobe-Japan1-UCS2 (8.001) does not work but 5.001 do.
Link to PDF file (or attach file here): http://s000.tinyupload.com/index.php?file_id=00088937975359643663

Configuration:
- Web browser and its version: Google Chrome 53.0.2785.143 m (the latest for now), Mozilla Firefox 49.0.1
- Operating system and its version: Windows 7 SP1 x64 
- PDF.js version: 1.6.210
- Is an extension: no

Not sure if this is a really PDF.js issue, probably it is related to the Adobe cmap files. The workaround of the issue is placed at the end if the issue description.

1. Open the attached file in the PDF.js viewer.
2a. The viewer throw the following error in the JS console: "DOMException: Invalid font data inArrayBuffer."
2b. The file displays incorrectly (characters are different in compare with the file opened in the Adobe Acrobat).

The PDF file opened in the Adobe Acrobat: https://snag.gy/0uApct.jpg
The PDF file opened in the PDF.js viewer (malformed): https://snag.gy/eVcl21.jpg
Developer console with error: https://snag.gy/ZoWXuA.jpg

IMPORTANT NOTES (WORKAROUND)
1. The viewer tries to use the Adobe-Japan1-UCS2.bcmap cmap that is in the original PDF.js distribution. 
2. If I set the cMapPacked property to false and add the unpacked cmap file from the Adobe Acrobat distribution, the behavior is the same. The version of the cmap file is 8.001
3. I've found the older cmap file version (5.001) and generated the appropriate packed cmap with the cmapscompress tool (https://github.com/mozilla/pdf.js/tree/master/external/cmapscompress). Everything started working fine for the PDF file: https://snag.gy/GwfJVE.jpg.

Plain text cmaps:
- 8.001 (from the Adobe Acrobat distro, not working): http://s000.tinyupload.com/index.php?file_id=17713344732879219876
- 5.001 (working): http://s000.tinyupload.com/index.php?file_id=91820851318135254049

Packed cmaps:
- probably 8.001 (from the PDF.js distro, not working): http://s000.tinyupload.com/index.php?file_id=57887949807560618862
- 5.001 (self generated, working): http://s000.tinyupload.com/index.php?file_id=01658083714704364959

### LINKED ISSUE #8255: v1.6.380+ Regression: Glyphs’ migration to PUA can fail, get missed when rendered
Link to PDF file (or attach file here): ~confidential but can send over secure channel~ reduced test case available, see comment

Configuration: Any
- Web browser and its version: Any
- Operating system and its version: Any
- PDF.js version: Any version more recent than 1.6.380
- Is an extension: No

Explanation:

Starting from build 1.6.380, presumably with commit 8d036faf403fb71fd106f7f605251268feb6a658, symbolic fonts are having their glyphs moved entirely to the Private Use Area in order to work around a canvas rendition problem.

I have seen this cause an issue with one of my documents where Hangul glyphs using Arial Unicode MS do not render properly in PDF.js on or after said release (1.6.380). I have traced the problem through the codebase and think I know what the problem is.

The problem is that there are three Private Use Areas…

1. Basic Multilingual Plane, U+E000 - U+F8FF, 6400 glyphs
2. Plane 15, U+F0000 - U+FFFFD, 65534 glyphs
3. Plane 16, U+100000 - U+10FFFD, 65534 glyphs

…and PDF.js only moves glyphs into the first area which supposedly gives it capacity for 6400 glyphs, but I have a document that triggers a problem where certain glyphs do not render, because it is using a font marked as Symbolic and has more than that amount of glyphs, with some of them already in the U+E000 - U+F8FF range. In practice, this causes certain documents to render without all glyphs being seen and is reproducible both in Firefox Developer Edition and a custom viewer I built by pulling components from `pdfjs-dist`.

I have currently worked around this issue in my own system by setting that font as not symbolic, and left the following comments for your reflection —

> This is probably because that the Private Use Area in question, while explicitly left undefined by the Unicode Consortium, has already been filled by vendors like Apple (which uses the last glyph 0xF8FF as an Apple icon) and Microsoft (in various Core Fonts, even Arial Unicode MS, whose version 1.01 contains 43 characters in the Private Use Area).
> Therefore at this time the most immediate solution is to work around this issue by ensuring that if Arial Unicode MS is marked as Symbolic by whatever application producing a PDF, it is unmarked as Symbolic here within $Service.

It would be nice if we can either patch `PRIVATE_USE_OFFSET_START` and `PRIVATE_USE_OFFSET_END` (in `src/core/fonts.js`) or make PDF.js use all three Private Use Areas and ensure that it does not touch any glyph already in these areas.

I may be able to contribute that patch given enough guidance.

### LINKED ISSUE #8570: Some Japanese characters do not render
[boxViewerで平仮名カタカナ句読点が抜けるPDF_Chrome印刷でPDF (1).pdf](https://github.com/mozilla/pdf.js/files/1098673/boxViewer.PDF_Chrome.PDF.1.pdf)

Configuration:
- Web browser and its version: Chrome Version 59.0.3071.104 
- Operating system and its version: Mac OS Sierra 10.12.5
- PDF.js version: Latest (in viewer.html demo)
- Is an extension: No

1. Go to pdf.js web viewer demo (https://mozilla.github.io/pdf.js/web/viewer.html)
2. View attached PDF

Should look like printed version of website: https://www.konicaminolta.jp/about/research/future/3dlr/

Only Chinese characters show

Additional notes:
Re-printing the web page as PDF seems to work in some situations in pdf.js, but the attached file does render properly in Mac Preview & Chrome's PDF viewer - is there a missing cmap or something?

brendandahl:
I had started looking at #8255 awhile back, but after seeing @Snuffleupagus pr #8572 I realized what was going on. 

This fixes a couple of things:
 - If dupFirstEntry was used in sanitizeGlyphLocations then the glyph count was too high and we ended up reading beyond loca table and disabling glyph that could actually be there.
- We didn't always mark missinglyphs as missing, now we always check the glyph length.
- We tried to adjust the mapping of missing glyphs. Often there are thousand of missing glyphs in regions that would be re-mapped which caused us to run out of private use locations.

Fixes #8255.
Fixes #8570.
Fixes #7696.


brendandahl:
I didn't see any failure locally (on macOS), but it looks like things have changed (for the worse) on windows. I'm guessing more glyphs are in the private use area not using the CJK shaper, but I'll have to look into it.

Snuffleupagus:
> I didn't see any failure locally (on macOS), but it looks like things have changed (for the worse) on windows. I'm guessing more glyphs are in the private use area not using the CJK shaper, but I'll have to look into it.

*One additional data-point, in case it's helpful:* I ran the full test-suite locally on Windows 7 (in Firefox Nightly 56), and I didn't notice any failures locally either.
However, I've got HWA (hardware acceleration) *enabled* locally, and I'm assuming that the bot runs with HWA *disabled*, which may help explain/pinpoint the issue!?

src/core/fonts.js
```diff
@@ -834,6 +834,7 @@ var Font = (function FontClosure() {
       if ((usedFontCharCodes[fontCharCode] !== undefined ||
            isProblematicUnicodeLocation(fontCharCode) ||
            (isSymbolic && !hasUnicodeValue)) &&
+          !missingGlyphs[glyphId] &&
```
Snuffleupagus: Just asking, and it might be a totally pointless micro-optimization, but would it make sense (at all) to check `!missingGlyphs[glyphId]` *first* in this condition? I.e. instead do:

```javascript
if (!missingGlyphs[glyphId] &&
    (usedFontCharCodes[fontCharCode] !== undefined ||
     isProblematicUnicodeLocation(fontCharCode) ||
     (isSymbolic && !hasUnicodeValue)) &&
    nextAvailableFontCharCode <= PRIVATE_USE_OFFSET_END) { // Room left.
```

The reason that I'm asking is that we could then avoid pointless `isProblematicUnicodeLocation` function calls when we *know* that a `glyph` is missing, since I'm assuming that it's cheaper to lookup `missingGlyphs[glyphId]` than it's calling a function?

Again, this might a completely unnecessary change, but I figured it can't hurt to ask!

brendandahl:
Looks like the windows failures may be a firefox update since the refs failed over in https://github.com/mozilla/pdf.js/pull/8585#issuecomment-311963567

src/core/fonts.js
brendandahl: I think it depends on a number of things and would really need to be benchmarked, but I don't thing this code is particularly hot, so it's not really worth worrying about it. I've updated it to be first though.

Snuffleupagus (approved):
This is a really nice improvement, not to mention simplification, of the way we detect missing glyphs.
This makes a lot more sense than the old hacks (that I added); thanks for doing this!

PR #8582, opened 2017-06-28
Convert the text layer builder to ES6 syntax

timvandermeij:
To enable the enhanced text selection, run `PDFViewerApplication.preferences.set('enhanceTextSelection', true);` in the console.

*Easier reviewing with https://github.com/mozilla/pdf.js/pull/8582/files?w=1.*

### DIFF
--- a/web/text_layer_builder.js
+++ b/web/text_layer_builder.js
@@ -16,7 +16,7 @@
 import { getGlobalEventBus } from './dom_events';
 import { renderTextLayer } from 'pdfjs-lib';
 
-var EXPAND_DIVS_TIMEOUT = 300; // ms
+const EXPAND_DIVS_TIMEOUT = 300; // ms
 
 /**
  * @typedef {Object} TextLayerBuilderOptions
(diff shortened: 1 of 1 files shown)


web/text_layer_builder.js
```diff
@@ -30,378 +30,376 @@ var EXPAND_DIVS_TIMEOUT = 300; // ms
...
+    let selectedMatchIdx = (this.findController === null ?
+                            -1 : this.findController.selected.matchIdx);
+    let highlightAll = (this.findController === null ?
+                        false : this.findController.state.highlightAll);
+    let infinity = {
+      divIdx: -1,
+      offset: undefined,
+    };
+
+    let beginText = function(begin, className) {
```
Snuffleupagus: I don't follow why was this function, and the one just below, was changed from `function funcName(...)` to `let funcName = function(...)`.
I'm probably missing something obvious here, mind explaining it to me?

timvandermeij: I just did that to be consistent with e.g., https://github.com/mozilla/pdf.js/blob/master/web/pdf_cursor_tools.js#L88 (and other places where such a format is used). The current notation works as well. I don't have a strong preference for either one, so I went with consistency here, but let me know if you want me to revert it.

Snuffleupagus: > I just did that to be consistent with e.g., https://github.com/mozilla/pdf.js/blob/master/web/pdf_cursor_tools.js#L88 (and other places where such a format is used).

Thanks for the explanation! However, isn't there a difference here, since in the case you're referring to we're using an arrow function to be able to access `this` without having to use `bind` or `self`?

I really don't know enough to say if there's (generally) any difference in performance between function declarations and function expressions, so do what you want here :-)

Snuffleupagus: If you decide to keep the new code (using function expressions), you probably want to switch the order of `beginText` and `appendTextToDiv` for good measure, since function *expressions* are not hoisted (as opposed to function *declarations*).

I.e. the following (simplified) example will not work:
```javascript
let beginText = function(...) {
  // Code here
  appendTextToDiv(...);
};
beginText({ ... }, '');

let appendTextToDiv = function(...) {
  // Code here
};
```
But this will example would work:
```javascript
let appendTextToDiv = function(...) {
  // Code here
};

let beginText = function(...) {
  // Code here
  appendTextToDiv(...);
};
beginText({ ... }, '');
```

timvandermeij: I decided to revert to the original function declarations to reduce the size of the change and because, as you mentioned, the function expression don't add much here.

Snuffleupagus (approved):
Looks good, thank you!

PR #8583, opened 2017-06-29
ES6-ify the code in `web/app.js` and `web/viewer.js` 

Snuffleupagus:
The changes consist mostly of changing `var` to `let`/`const`, and using shorthand method signatures.

*Easier reviewing with https://github.com/mozilla/pdf.js/pull/8583/files?w=1.*

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -44,8 +44,8 @@ import { SecondaryToolbar } from './secondary_toolbar';
 import { Toolbar } from './toolbar';
 import { ViewHistory } from './view_history';
 
-var DEFAULT_SCALE_DELTA = 1.1;
-var DISABLE_AUTO_FETCH_LOADING_BAR_TIMEOUT = 5000;
+const DEFAULT_SCALE_DELTA = 1.1;
+const DISABLE_AUTO_FETCH_LOADING_BAR_TIMEOUT = 5000;
 
 function configure(PDFJS) {
   PDFJS.imageResourcesPath = './images/';
@@ -64,7 +64,7 @@ function configure(PDFJS) {
   }
 }
 
-var DefaultExternalServices = {
+const DefaultExternalServices = {
   updateFindControlState(data) {},
   initPassiveLoading(callbacks) {},
   fallback(data, callback) {},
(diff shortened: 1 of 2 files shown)


web/app.js
```diff
@@ -87,7 +87,7 @@ var DefaultExternalServices = {
   },
 };
 
-var PDFViewerApplication = {
+let PDFViewerApplication = {
```
Snuffleupagus: Please note that using `let` here won't change the ability to access `PDFViewerApplication` through the console (or similar), since [viewer.js#L183](https://github.com/mozilla/pdf.js/blob/d6a6f390ae9a39554b73a942860f67badbef1b4b/web/viewer.js#L183) and [viewer.js#L187](https://github.com/mozilla/pdf.js/blob/d6a6f390ae9a39554b73a942860f67badbef1b4b/web/viewer.js#L187) already ensures that it's defined globally.

web/app.js
```diff
@@ -718,42 +716,36 @@ var PDFViewerApplication = {
...
-    }
-
-    this.pdfDocument.getData().then(
-      function getDataSuccess(data) {
-        var blob = createBlob(data, 'application/pdf');
-        downloadManager.download(blob, url, filename);
-      },
-      downloadByUrl // Error occurred try downloading with just the url.
-    ).then(null, downloadByUrl);
+    this.pdfDocument.getData().then(function getDataSuccess(data) {
```
timvandermeij: I think we can use `function(data)` (anonymous function) here too.

web/viewer.js
```diff
@@ -16,26 +16,26 @@
 
 'use strict';
 
-var DEFAULT_URL = 'compressed.tracemonkey-pldi-09.pdf';
+let DEFAULT_URL = 'compressed.tracemonkey-pldi-09.pdf';
```
timvandermeij: Should this be `const`, or did you do it this way so external code can edit it?

timvandermeij (approved):
Looks good! r=me with or without the comments addressed since they're minor.

web/app.js
Snuffleupagus: That's a fair point, will fix that tomorrow!

web/viewer.js
Snuffleupagus: Sorry, I really should have let a comment about this as well, but unfortunately we cannot do that :-(
Apparently `CHROME` builds need to mess with it, see https://github.com/mozilla/pdf.js/blob/master/web/viewer.js#L27, and I didn't bother looking into if that can be re-factored somehow.

PR #8585, opened 2017-06-29
Check that the `MessageHandler` isn't already terminated in the `onFailure` handler in `src/core/worker.js` (issue 8584)

### LINKED ISSUE #8584: error in console when calling destroy() method
Related to https://github.com/mozilla/pdf.js/issues/3485

Link to Test file (or attach file here):
[test.zip](https://github.com/mozilla/pdf.js/files/1111691/test.zip)

Links to files used in test:
https://github.com/mozilla/pdfjs-dist/blob/master/build/pdf.combined.js
https://github.com/mozilla/pdfjs-dist/blob/master/web/compatibility.js
https://github.com/mozilla/pdf.js/blob/master/web/compressed.tracemonkey-pldi-09.pdf

Configuration:
Firefox 54.0
Chrome Version 59.0.3071.115 (Official Build) (64-bit)
Windows 10 version 1703 build 15063.413
- PDF.js version:
1.7.225

1. extract test.zip to www under Apache
2. open test.html in Firefox or Chrome

loading is stopped without errors

What went wrong?
**In Firefox**
13:36:22.790 TypeError: handler is null 1 pdf.combined.js:27936:11

**In Chrome**
Uncaught (in promise) TypeError: Cannot read property 'send' of null
    at onFailure (pdf.combined.js:27936)
    at <anonymous>

Snuffleupagus:
All other code-paths already checks that the `MessageHandler` isn't terminated, but apparently `onFailure` was missing that check (compare e.g. with the `onSuccess` function).
From what I can tell, this is only an issue if workers are *disabled*, hence why I didn't bother adding a unit-test.

Fixes #8584.

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -654,6 +654,8 @@ var WorkerMessageHandler = {
       }
 
       function onFailure(e) {
+        ensureNotTerminated();
+
         if (e instanceof PasswordException) {
           var task = new WorkerTask('PasswordException: response ' + e.code);
           startWorkerTask(task);


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #7959, opened 2017-01-16
Display a notification on the `sidebarToggle` button for PDF documents with outline/attachments

Snuffleupagus:
*This is something that I've had lying around for quite a while locally, and I figured that I might as well just submit it to see if others think there's value in it!*

Fixes #5961
Fixes #7025

---

 - Display a notification on the `sidebarToggle` button for PDF documents with outline/attachments

   > A longstanding issue with the viewer is that you cannot tell if a PDF document includes an outline and/or attachments without actually opening the sidebar.
This patch contains a suggested solution for that, by displaying an hide-on-interaction notification on the `sidebarToggle` button (and the relevant sidebar view buttons). Note that this was inspired by e.g. the update notification that is displayed on the menu button in Firefox.
   >
   > For an initial implementation, I've tried to do this in such a way that the notification isn't too distracting. Without being an UX expert, I don't think that we'd want something too in-your-face, in order to keep the viewer toolbars reasonable clean. (We probably do *not* want e.g. an entire notification bar in these situations, since that would take up unnecessary screen space and require actions from the user to close.)
   >
   > However it's certainly possible that the current notification might simply be *too* inconspicuous to be truly helpful to users, but we could probably iterate on that if the feature itself is deemed useful.
One idea could be to e.g. use a modified tooltip for the `sidebarToggle`, e.g. "Toggle Sidebar (outline and/or attachments available)", when the notification is used.

 - Append the contents of `FileAttachment` annotations to the attachments view of the sidebar, for easier access to the embedded files

   > Other PDF viewers, e.g. Adobe Reader, seem to append `FileAttachment`s to their attachments views.
One obvious difference in PDF.js is that we cannot append all the annotations on document load, since that would require parsing *every* page. Despite that, it still seems like a good idea to add `FileAttachment`s, since it's thus possible to access all the various types of attachments from a single place.
   >
   > *Note:* With the previous patch we display a notification when a `FileAttachment` is added to the sidebar, which thus makes appending the contents of these annotations to the sidebar slightly more visible/useful.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/pdf.js/7959)

PR #8594, opened 2017-06-30
Change `var` to `let`, and use object destructuring, in a couple of previously class converted `web/*.js` files

Snuffleupagus:
Note that these files were among the first to be converted to ES6 classes, so it probably makes sense to do another pass to bring them inline with the most recent ES6 conversions.

*Edit:* If there would have been any scoping errors from the `var` -> `let` conversion, the `no-undef` ESLint rule would have caught that, which should provide some assurance when reviewing.

### DIFF
--- a/web/pdf_attachment_viewer.js
+++ b/web/pdf_attachment_viewer.js
@@ -27,19 +27,19 @@ import {
 
 /**
  * @typedef {Object} PDFAttachmentViewerRenderParameters
- * @property {Array|null} attachments - An array of attachment objects.
+ * @property {Object|null} attachments - A lookup table of attachment objects.
  */
 
 class PDFAttachmentViewer {
   /**
    * @param {PDFAttachmentViewerOptions} options
    */
-  constructor(options) {
+  constructor({ container, eventBus, downloadManager, }) {
     this.attachments = null;
 
-    this.container = options.container;
-    this.eventBus = options.eventBus;
-    this.downloadManager = options.downloadManager;
+    this.container = container;
+    this.eventBus = eventBus;
+    this.downloadManager = downloadManager;
 
     this._renderedCapability = createPromiseCapability();
     this.eventBus.on('fileattachmentannotation',
@@ -79,12 +79,12 @@ class PDFAttachmentViewer {
       throw new Error('bindPdfLink: ' +
                       'Unsupported "PDFJS.disableCreateObjectURL" value.');
     }
-    var blobUrl;
+    let blobUrl;
     button.onclick = function() {
       if (!blobUrl) {
         blobUrl = createObjectURL(content, 'application/pdf');
       }
-      var viewerUrl;
+      let viewerUrl;
       if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
         // The current URL is the viewer, let's use it and append the file.
         viewerUrl = '?file=' + encodeURIComponent(blobUrl + '#' + filename);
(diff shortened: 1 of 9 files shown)


web/pdf_attachment_viewer.js
```diff
@@ -116,13 +116,12 @@ class PDFAttachmentViewer {
   /**
    * @param {PDFAttachmentViewerRenderParameters} params
    */
-  render(params = {}) {
-    var attachments = params.attachments || null;
-    var attachmentsCount = 0;
+  render({ attachments, keepRenderedCapability = false, }) {
+    attachments = attachments || null;
```
timvandermeij: I think we can remove this line and just set `attachments = null` in the destructured object above.

web/pdf_attachment_viewer.js
```diff
@@ -116,13 +116,12 @@ class PDFAttachmentViewer {
   /**
    * @param {PDFAttachmentViewerRenderParameters} params
    */
-  render(params = {}) {
-    var attachments = params.attachments || null;
-    var attachmentsCount = 0;
+  render({ attachments, keepRenderedCapability = false, }) {
```
timvandermeij: `keepRenderedCapability` is not in the documentation for `PDFAttachmentViewerRenderParameters` above. Could you add it for consistency?

web/pdf_outline_viewer.js
```diff
@@ -156,9 +156,9 @@ class PDFOutlineViewer {
   /**
    * @param {PDFOutlineViewerRenderParameters} params
    */
-  render(params = {}) {
-    var outline = params.outline || null;
-    var outlineCount = 0;
+  render({ outline, }) {
+    outline = outline || null;
```
timvandermeij: Same here, can't we remove this line and use `outline = null` above instead?

timvandermeij (approved):
Nice work! r=me with the comments addressed.

web/pdf_attachment_viewer.js
Snuffleupagus: There's a comment in https://github.com/mozilla/pdf.js/blob/9e4dcf156b6d31926b8d48112ee6625b674b1ddc/web/pdf_attachment_viewer.js#L55-L59, which was added as a result of the discussion in https://github.com/mozilla/pdf.js/pull/7959#discussion_r96517486.

Since that parameter should *not* be used unless you know what you're doing, I didn't want to add a too visible comment; do you no longer think that this was the right call here?

web/pdf_attachment_viewer.js
Snuffleupagus: I thought about this while writing the patch, and ended up purposely leaving this code intact (given that default parameters are only applied when `param === undefined`).
Please note that the `attachments` parameter should *not* be regarded as optional, which is what a default parameter suggests!

In this case, what we pass to `render` in https://github.com/mozilla/pdf.js/blob/9e4dcf156b6d31926b8d48112ee6625b674b1ddc/web/app.js#L1046 comes directly from the API. Currently we'll return `null`, coming from https://github.com/mozilla/pdf.js/blob/9e4dcf156b6d31926b8d48112ee6625b674b1ddc/src/core/obj.js#L318, when there're no attachments present. But I figured that it makes sense for `web/pdf_attachment_viewer.js` to not assume that, and always fallback to `null` here.

Hence it might be most correct to actually do something like `attachments = isArray(attachments) ? attachments : null;` instead here!?

web/pdf_outline_viewer.js
Snuffleupagus: Please refer to https://github.com/mozilla/pdf.js/pull/8594#discussion_r125156197, since that applies here as well.

timvandermeij:
Thank you for landing this! I agree with the comments you added above.

PR #8132, opened 2017-03-04
Viewer: enable find functionality for small devices

timvandermeij:
The commit messages provide more information about the intent of each patch.

This PR is a re-do of #6274, so please refer to that PR for previous discussion on this subject. This patch aims to resolve the issues mentioned there while keeping the code changes minimal.

Fixes #6191.
Supersedes #6274.

*Reviewing is easier with https://github.com/mozilla/pdf.js/pull/8132/files?w=1.*

PR #8609, opened 2017-07-05
Ensure that `PDFFindBar._adjustWidth` is called in all situations where the width of the findbar might have changed

Snuffleupagus:
- Actually reset the `findResultsCount` label, in addition to hiding it, when no matches are found

   Currently we're *only* hiding the label, but not actually resetting it until a new match is found.
Obviously it's being hidden, but it seems that it really ought to be completely reset as well (since e.g. `PDFFindBar.reset` won't technically reset *all* state otherwise).

 - Ensure that `PDFFindBar._adjustWidth` is called in all situations where the width of the findbar might have changed

   After PR #8394, where the l10n service was converted to be asynchronous, we're no longer calling `_adjustWidth` after updating the `findMsg` label. Hence it's currently possible that the width of the findbar won't be correct. The solution is simple though, just call `_adjustWidth` after the `findMsg` label has been (asynchronously) updated.

   Another existing issue, which was an oversight in PR #8132, is that `PDFFindBar.updateResultsCount` may be called directly from `PDFFindController`. In that case, we're not calling `_adjustWidth` at all, which means that the findbar may also not have the correct width.
The simple solution here is to always call `_adjustWidth` at the end of `updateResultsCount` (which is why we no longer need the `_adjustWidth` call at the end of `updateUIState`).

### DIFF
--- a/web/pdf_find_bar.js
+++ b/web/pdf_find_bar.js
@@ -151,15 +151,15 @@ class PDFFindBar {
       return; // No UI control is provided.
     }
 
-    // If there are no matches, hide the counter.
     if (!matchCount) {
+      // If there are no matches, hide and reset the counter.
       this.findResultsCount.classList.add('hidden');
-      return;
+      this.findResultsCount.textContent = '';
+    } else {
+      // Update and show the match counter.
+      this.findResultsCount.textContent = matchCount.toLocaleString();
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

Issue #7613, opened 2016-09-07
Interactive form (AcroForm) support


timvandermeij:
_This is a tracking issue only, so this is not the place for any other questions or discussions. Open a new issue for that._

This is a meta issue for interactive form (AcroForm) support according to Chapter 12.7 of the PDF reference (http://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G11.2110737). This includes all form elements except for signature fields, which are tracked in #1076. The objective is to get https://github.com/mozilla/pdf.js/blob/master/test/pdfs/f1040.pdf.link to render completely, but also to resolve other open issues and PRs.

**General**
- [x] Prepare core and display layer for implementing form elements (#7596)
- [x] Reference testing (#7602)
- [x] Preference (#7602)
- [x] Remove global `PDFJS.renderInteractiveForms` usage (#7640)
- [x] Refactor field name construction code in `WidgetAnnotation` (#7775)
- [x] Refactor or clarify where annotations are rendered
  - Mostly in the display layer, but text widget annotations with appearance streams are rendered in the core layer, which causes confusion...
- [x] Appearances
- [x] Storing entered values for when the page is destroyed when it is not visible
- [x] Printing entered values
  - Either print the HTML elements or render the contents onto the canvas (use `appendToOperatorList`)
- [x] Enable by default
- [x] Update the example (#8030)
- [x] Add Firefox pref to enable/disable forms (https://bugzilla.mozilla.org/show_bug.cgi?id=1652145)

**Text widgets**
- [x] Rendering of single-line fields (#7602)
- [x] Handle maximum length (#7622)
- [x] Handle flags: multiline and read-only (#7633)
- [x] Handle flags: comb (#7649)
- [x] Handle justification (#7622)
- [x] Sanitize `maxLen` and `textAlignment` in the core layer and unit tests for this (#7629)

**Choice widgets**
- [x] Rendering of combo boxes (#7671)
- [x] Rendering of list boxes (#7671)

**Button widgets**
- [x] Rendering of pushbuttons (#9191)
- [x] Rendering of checkboxes (#7898)
- [x] Rendering of radio buttons (#7898)

Snuffleupagus:
[...]

PR #8616, opened 2017-07-05
Issue 7613, Parse AcroForm dictionary

dmitryskey:
https://github.com/mozilla/pdf.js/issues/7613, Parse AcroForm dictionary and 'DR' appearance and set annotation wizard font properties.

Appearances
It seems like the current font code for text widgets is dead as fontRefName is never set...
Parse AcroForm dictionary


dmitryskey:
Tests failed

PR #8617, opened 2017-07-05
Adds Streams API support for networking task of PDF.js project.

mukulmishra18:
This PR aims to add Streams API support for networking task of PDF.js. For the same, `network.js` file is moved to the main thread and `PDFNetwokStream` is implemented at worker to talk to main via `sendWithStream` and/or `sendWithPromise` method of `MessageHandler`.

Worker will ask for data(when needed) by sending stream messages to main side. PDF data is returned from the handlers(main to worker) in small chunks.

**TODO:**
- [x] enqueue needs to access tranfers parameter

- [x]  check how size and highwatermark work (for now highwatermark = 1 and every chunk size = 1)

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -198,217 +198,119 @@ IPDFStreamRangeReader.prototype = {
 
 /** @implements {IPDFStream} */
 var PDFWorkerStream = (function PDFWorkerStreamClosure() {
-  function PDFWorkerStream(params, msgHandler) {
-    this._queuedChunks = [];
-    var initialData = params.initialData;
-    if (initialData && initialData.length > 0) {
-      this._queuedChunks.push(initialData);
-    }
+  function PDFWorkerStream(msgHandler) {
     this._msgHandler = msgHandler;
-
-    this._isRangeSupported = !(params.disableRange);
-    this._isStreamingSupported = !(params.disableStream);
-    this._contentLength = params.length;
-
+    this._contentLength = null;
     this._fullRequestReader = null;
-    this._rangeReaders = [];
-
-    msgHandler.on('OnDataRange', this._onReceiveData.bind(this));
-    msgHandler.on('OnDataProgress', this._onProgress.bind(this));
+    this._rangeRequestReaders = [];
   }
   PDFWorkerStream.prototype = {
-    _onReceiveData: function PDFWorkerStream_onReceiveData(args) {
-       if (args.begin === undefined) {
-         if (this._fullRequestReader) {
-           this._fullRequestReader._enqueue(args.chunk);
-         } else {
-           this._queuedChunks.push(args.chunk);
-         }
-       } else {
-         var found = this._rangeReaders.some(function (rangeReader) {
-           if (rangeReader._begin !== args.begin) {
-             return false;
-           }
-           rangeReader._enqueue(args.chunk);
-           return true;
-         });
-         assert(found);
-       }
-    },
-
-    _onProgress: function PDFWorkerStream_onProgress(evt) {
-       if (this._rangeReaders.length > 0) {
-         // Reporting to first range reader.
-         var firstReader = this._rangeReaders[0];
-         if (firstReader.onProgress) {
-           firstReader.onProgress({ loaded: evt.loaded, });
-         }
-       }
-    },
-
-    _removeRangeReader: function PDFWorkerStream_removeRangeReader(reader) {
-      var i = this._rangeReaders.indexOf(reader);
-      if (i >= 0) {
-        this._rangeReaders.splice(i, 1);
-      }
-    },
-
-    getFullReader: function PDFWorkerStream_getFullReader() {
+    getFullReader() {
       assert(!this._fullRequestReader);
-      var queuedChunks = this._queuedChunks;
-      this._queuedChunks = null;
-      return new PDFWorkerStreamReader(this, queuedChunks);
+      this._fullRequestReader = new PDFWorkerStreamReader(this._msgHandler);
+      return this._fullRequestReader;
     },
 
-    getRangeReader: function PDFWorkerStream_getRangeReader(begin, end) {
-      var reader = new PDFWorkerStreamRangeReader(this, begin, end);
-      this._msgHandler.send('RequestDataRange', { begin, end, });
-      this._rangeReaders.push(reader);
+    getRangeReader(begin, end) {
+      let reader = new PDFWorkerStreamRangeReader(begin, end, this._msgHandler);
+      this._rangeRequestReaders.push(reader);
       return reader;
     },
 
-    cancelAllRequests: function PDFWorkerStream_cancelAllRequests(reason) {
+    cancelAllRequests(reason) {
       if (this._fullRequestReader) {
         this._fullRequestReader.cancel(reason);
       }
-      var readers = this._rangeReaders.slice(0);
-      readers.forEach(function (rangeReader) {
-        rangeReader.cancel(reason);
+      let readers = this._rangeRequestReaders.slice(0);
+      readers.forEach(function (reader) {
+        reader.cancel(reason);
       });
     },
   };
 
   /** @implements {IPDFStreamReader} */
-  function PDFWorkerStreamReader(stream, queuedChunks) {
-    this._stream = stream;
-    this._done = false;
-    this._queuedChunks = queuedChunks || [];
-    this._requests = [];
-    this._headersReady = Promise.resolve();
-    stream._fullRequestReader = this;
+  function PDFWorkerStreamReader(msgHandler) {
+    this._msgHandler = msgHandler;
 
-    this.onProgress = null; // not used
+    this._contentLength = null;
+    this._isRangeSupported = false;
+    this._isStreamingSupported = false;
+
+    let readableStream = this._msgHandler.sendWithStream('GetReader');
+
+    this._reader = readableStream.getReader();
+
+    this._headersReady = this._msgHandler.sendWithPromise('ReaderHeadersReady').
+        then((data) => {
+      this._isStreamingSupported = data.isStreamingSupported;
+      this._isRangeSupported = data.isRangeSupported;
+      this._contentLength = data.contentLength;
+    });
   }
   PDFWorkerStreamReader.prototype = {
-    _enqueue: function PDFWorkerStreamReader_enqueue(chunk) {
-      if (this._done) {
-        return; // ignore new data
-      }
-      if (this._requests.length > 0) {
-        var requestCapability = this._requests.shift();
-        requestCapability.resolve({ value: chunk, done: false, });
-        return;
-      }
-      this._queuedChunks.push(chunk);
-    },
-
     get headersReady() {
       return this._headersReady;
     },
 
-    get isRangeSupported() {
-      return this._stream._isRangeSupported;
+    get contentLength() {
+      return this._contentLength;
     },
 
     get isStreamingSupported() {
-      return this._stream._isStreamingSupported;
+      return this._isStreamingSupported;
     },
 
-    get contentLength() {
-      return this._stream._contentLength;
+    get isRangeSupported() {
+      return this._isRangeSupported;
     },
 
-    read: function PDFWorkerStreamReader_read() {
-      if (this._queuedChunks.length > 0) {
-        var chunk = this._queuedChunks.shift();
-        return Promise.resolve({ value: chunk, done: false, });
-      }
-      if (this._done) {
-        return Promise.resolve({ value: undefined, done: true, });
-      }
-      var requestCapability = createPromiseCapability();
-      this._requests.push(requestCapability);
-      return requestCapability.promise;
-    },
-
-    cancel: function PDFWorkerStreamReader_cancel(reason) {
-      this._done = true;
-      this._requests.forEach(function (requestCapability) {
-        requestCapability.resolve({ value: undefined, done: true, });
+    read() {
+      return this._reader.read().then(function({ value, done, }) {
+        if (done) {
+          return { value: undefined, done: true, };
+        }
+        // `value` is wrapped into Uint8Array, we need to
+        // unwrap it to ArrayBuffer for further processing.
+        return { value: value.buffer, done: false, };
       });
-      this._requests = [];
+    },
+
+    cancel(reason) {
+      this._reader.cancel(reason);
     },
   };
 
   /** @implements {IPDFStreamRangeReader} */
-  function PDFWorkerStreamRangeReader(stream, begin, end) {
-    this._stream = stream;
-    this._begin = begin;
-    this._end = end;
-    this._queuedChunk = null;
-    this._requests = [];
-    this._done = false;
-
+  function PDFWorkerStreamRangeReader(begin, end, msgHandler) {
+    this._msgHandler = msgHandler;
     this.onProgress = null;
+
+    let readableStream = this._msgHandler.sendWithStream('GetRangeReader',
+                                                         { begin, end, });
+
+    this._reader = readableStream.getReader();
   }
   PDFWorkerStreamRangeReader.prototype = {
-    _enqueue: function PDFWorkerStreamRangeReader_enqueue(chunk) {
-      if (this._done) {
-        return; // ignore new data
-      }
-      if (this._requests.length === 0) {
-        this._queuedChunk = chunk;
-      } else {
-        var requestsCapability = this._requests.shift();
-        requestsCapability.resolve({ value: chunk, done: false, });
-        this._requests.forEach(function (requestCapability) {
-          requestCapability.resolve({ value: undefined, done: true, });
-        });
-        this._requests = [];
-      }
-      this._done = true;
-      this._stream._removeRangeReader(this);
-    },
-
     get isStreamingSupported() {
       return false;
     },
 
-    read: function PDFWorkerStreamRangeReader_read() {
-      if (this._queuedChunk) {
-        return Promise.resolve({ value: this._queuedChunk, done: false, });
-      }
-      if (this._done) {
-        return Promise.resolve({ value: undefined, done: true, });
-      }
-      var requestCapability = createPromiseCapability();
-      this._requests.push(requestCapability);
-      return requestCapability.promise;
+    read() {
+      return this._reader.read().then(function({ value, done, }) {
+        if (done) {
+          return { value: undefined, done: true, };
+        }
+        return { value: value.buffer, done: false, };
+      });
     },
 
-    cancel: function PDFWorkerStreamRangeReader_cancel(reason) {
-      this._done = true;
-      this._requests.forEach(function (requestCapability) {
-        requestCapability.resolve({ value: undefined, done: true, });
-      });
-      this._requests = [];
-      this._stream._removeRangeReader(this);
+    cancel(reason) {
+      this._reader.cancel(reason);
     },
   };
 
   return PDFWorkerStream;
 })();
-
-/** @type IPDFStream */
-var PDFNetworkStream;
-
-/**
- * Sets PDFNetworkStream class to be used as alternative PDF data transport.
- * @param {IPDFStream} cls - the PDF data transport.
- */
-function setPDFNetworkStreamClass(cls) {
-  PDFNetworkStream = cls;
-}
 
 var WorkerMessageHandler = {
   setup(handler, port) {
@@ -536,16 +438,9 @@ var WorkerMessageHandler = {
         return pdfManagerCapability.promise;
       }
 
-      var pdfStream;
+      var pdfStream, cachedChunks = [];
       try {
-        if (source.chunkedViewerLoading) {
-          pdfStream = new PDFWorkerStream(source, handler);
-        } else {
-          if (!PDFNetworkStream) {
-            throw new Error('./network module is not loaded');
-          }
-          pdfStream = new PDFNetworkStream(data);
-        }
+        pdfStream = new PDFWorkerStream(handler);
       } catch (ex) {
         pdfManagerCapability.reject(ex);
         return pdfManagerCapability.promise;
@@ -553,18 +448,6 @@ var WorkerMessageHandler = {
 
       var fullRequest = pdfStream.getFullReader();
       fullRequest.headersReady.then(function () {
-        if (!fullRequest.isStreamingSupported ||
-            !fullRequest.isRangeSupported) {
-          // If stream or range are disabled, it's our only way to report
-          // loading progress.
-          fullRequest.onProgress = function (evt) {
-            handler.send('DocProgress', {
-              loaded: evt.loaded,
-              total: evt.total,
-            });
-          };
-        }
-
         if (!fullRequest.isRangeSupported) {
           return;
         }
@@ -580,6 +463,15 @@ var WorkerMessageHandler = {
           disableAutoFetch,
           rangeChunkSize: source.rangeChunkSize,
         }, evaluatorOptions, docBaseUrl);
+        // There may be a chance that `pdfManager` is not initialized
+        // for first few runs of `readchunk` block of code. Be sure
+        // to send all cached chunks, if any, to chunked_stream via
+        // pdf_manager.
+        for (let i = 0; i < cachedChunks.length; i++) {
+          pdfManager.sendProgressiveData(cachedChunks[i]);
+        }
+
+        cachedChunks = [];
         pdfManagerCapability.resolve(pdfManager);
         cancelXHRs = null;
       }).catch(function (reason) {
@@ -587,7 +479,7 @@ var WorkerMessageHandler = {
         cancelXHRs = null;
       });
 
-      var cachedChunks = [], loaded = 0;
+      var loaded = 0;
       var flushChunks = function () {
         var pdfFile = arraysToBytes(cachedChunks);
         if (source.length && pdfFile.length !== source.length) {
@@ -969,7 +861,6 @@ if (typeof window === 'undefined' && !isNodeJS() &&
 }
 
 export {
-  setPDFNetworkStreamClass,
   WorkerTask,
   WorkerMessageHandler,
 };
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -15,7 +15,7 @@
 /* globals requirejs, __non_webpack_require__ */
 
 import {
-  createPromiseCapability, deprecated, getVerbosityLevel, globalScope,
+  assert, createPromiseCapability, deprecated, getVerbosityLevel, globalScope,
   info, InvalidPDFException, isArray, isArrayBuffer, isInt, isSameOrigin,
   loadJpegStream, MessageHandler, MissingPDFException, NativeImageDecoding,
   PageViewport, PasswordException, StatTimer, stringToBytes,
@@ -28,6 +28,7 @@ import {
 import { FontFaceObject, FontLoader } from './font_loader';
 import { CanvasGraphics } from './canvas';
 import { Metadata } from './metadata';
+import { PDFDataTransportStream } from './transport_stream';
 
 var DEFAULT_RANGE_CHUNK_SIZE = 65536; // 2^16 = 65536
 
@@ -78,6 +79,17 @@ if (typeof PDFJSDev !== 'undefined' &&
       callback(worker.WorkerMessageHandler);
     });
   }) : null;
+}
+
+/** @type IPDFStream */
+var PDFNetworkStream;
+
+/**
+ * Sets PDFNetworkStream class to be used as alternative PDF data transport.
+ * @param {IPDFStream} cls - the PDF data transport.
+ */
+function setPDFNetworkStreamClass(cls) {
+  PDFNetworkStream = cls;
 }
 
 /**
@@ -281,8 +293,20 @@ function getDocument(src, pdfDataRangeTransport,
       if (task.destroyed) {
         throw new Error('Loading aborted');
       }
+
+      let networkStream;
+      if (rangeTransport) {
+        networkStream = new PDFDataTransportStream(params, rangeTransport);
+      } else if (!params.data) {
+        networkStream = new PDFNetworkStream({
+          source: params,
+          disableRange: getDefaultSetting('disableRange'),
+        });
+      }
+
       var messageHandler = new MessageHandler(docId, workerId, worker.port);
-      var transport = new WorkerTransport(messageHandler, task, rangeTransport,
+      messageHandler.postMessageTransfers = worker.postMessageTransfers;
+      var transport = new WorkerTransport(messageHandler, task, networkStream,
                                           CMapReaderFactory);
       task._transport = transport;
       messageHandler.send('Ready', null);
@@ -317,7 +341,6 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
   return worker.messageHandler.sendWithPromise('GetDocRequest', {
     docId,
     source,
-    disableRange: getDefaultSetting('disableRange'),
     maxImageSize: getDefaultSetting('maxImageSize'),
     disableFontFace: getDefaultSetting('disableFontFace'),
     disableCreateObjectURL: getDefaultSetting('disableCreateObjectURL'),
@@ -1238,9 +1261,7 @@ var PDFWorker = (function PDFWorkerClosure() {
     // pdf.worker.js file is needed.
     if (typeof PDFJSDev === 'undefined' || !PDFJSDev.test('PRODUCTION')) {
       if (typeof SystemJS === 'object') {
-        Promise.all([SystemJS.import('pdfjs/core/network'),
-                     SystemJS.import('pdfjs/core/worker')]).then((modules) => {
-          var worker = modules[1];
+        SystemJS.import('pdfjs/core/worker').then((worker) => {
           WorkerMessageHandler = worker.WorkerMessageHandler;
           fakeWorkerFilesLoadedCapability.resolve(WorkerMessageHandler);
         });
@@ -1254,7 +1275,6 @@ var PDFWorker = (function PDFWorkerClosure() {
       }
     } else if (PDFJSDev.test('SINGLE_FILE')) {
       var pdfjsCoreWorker = require('../core/worker.js');
-      require('../core/network.js');
       WorkerMessageHandler = pdfjsCoreWorker.WorkerMessageHandler;
       fakeWorkerFilesLoadedCapability.resolve(WorkerMessageHandler);
     } else {
@@ -1285,6 +1305,7 @@ var PDFWorker = (function PDFWorkerClosure() {
 
     this.name = name;
     this.destroyed = false;
+    this.postMessageTransfers = true;
 
     this._readyCapability = createPromiseCapability();
     this._port = null;
@@ -1381,6 +1402,7 @@ var PDFWorker = (function PDFWorkerClosure() {
               this._port = worker;
               this._webWorker = worker;
               if (!data.supportTransfers) {
+                this.postMessageTransfers = false;
                 isPostMessageTransfersDisabled = true;
               }
               this._readyCapability.resolve();
@@ -1513,11 +1535,10 @@ var PDFWorker = (function PDFWorkerClosure() {
  * @ignore
  */
 var WorkerTransport = (function WorkerTransportClosure() {
-  function WorkerTransport(messageHandler, loadingTask, pdfDataRangeTransport,
+  function WorkerTransport(messageHandler, loadingTask, networkStream,
                            CMapReaderFactory) {
     this.messageHandler = messageHandler;
     this.loadingTask = loadingTask;
-    this.pdfDataRangeTransport = pdfDataRangeTransport;
     this.commonObjs = new PDFObjects();
     this.fontLoader = new FontLoader(loadingTask.docId);
     this.CMapReaderFactory = new CMapReaderFactory({
@@ -1528,6 +1549,10 @@ var WorkerTransport = (function WorkerTransportClosure() {
     this.destroyed = false;
     this.destroyCapability = null;
     this._passwordCapability = null;
+
+    this._networkStream = networkStream;
+    this._fullReader = null;
+    this._lastProgress = null;
 
     this.pageCache = [];
     this.pagePromises = [];
@@ -1564,10 +1589,10 @@ var WorkerTransport = (function WorkerTransportClosure() {
       waitOn.push(terminated);
       Promise.all(waitOn).then(() => {
         this.fontLoader.clear();
-        if (this.pdfDataRangeTransport) {
-          this.pdfDataRangeTransport.abort();
-          this.pdfDataRangeTransport = null;
+        if (this._networkStream) {
+          this._networkStream.cancelAllRequests();
         }
+
         if (this.messageHandler) {
           this.messageHandler.destroy();
           this.messageHandler = null;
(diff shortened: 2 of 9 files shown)


src/core/worker.js
```diff
@@ -197,6 +197,115 @@ IPDFStreamRangeReader.prototype = {
 }
 
 /** @implements {IPDFStream} */
+class PDFNetworkStream {
```
Snuffleupagus: Given the IRC discussion yesterday, I'm assuming that we shouldn't use `class` here (or elsewhere), since this code is in `src/core`!?

src/core/worker.js
```diff
@@ -197,6 +197,115 @@ IPDFStreamRangeReader.prototype = {
...
+  constructor(data, msgHandler) {
+    this._data = data;
```
Snuffleupagus: Nit: Underscore is usually used to denote a property, or method, that should be considered "private". However, in both `PDFNetworkStreamReader` and `PDFNetworkStreamRangeReader` below, you're *directly* accessing `_data` (and* `_msgHandler`). To me, this feels like abusing notation, and it'd be better it you change the properties to `this.data = data` (and `this.msgHandler`) here instead. (Another option would be to keep them "private" here, and add getters.)

src/core/worker.js
mukulmishra18: Hmm..., I see. Okay, I will change it to the way we implement `class` in PDF.js, like: [IPDFStream](https://github.com/mozilla/pdf.js/blob/master/src/core/worker.js#L62). Although, I still think `class` looks nice :-)

src/core/worker.js
mukulmishra18: Sorry, I will change it to `this.data = data`.

src/shared/util.js
```diff
@@ -1433,13 +1433,19 @@ MessageHandler.prototype = {
     let targetName = data.sourceName;
     let capability = createPromiseCapability();
 
-    let sendStreamRequest = ({ stream, chunk, success, reason, }) => {
+    let sendStreamRequest = ({ stream, chunk, transfers,
+                               success, reason, }) => {
+      if (transfers) {
```
Snuffleupagus: I have to ask: Why isn't is sufficient to simply always call
```javascript
this.postMessage({ sourceName, targetName, stream, streamId,
                   chunk, success, reason, }, transfers);
```
and just let the `postMessage` method deal with it?
Also, don't you need to add similar code in `_processStreamMessage` below as well?

mukulmishra18: > and just let the postMessage method deal with it?

Yeah, right. Seems like `this.postMessage` is doing same thing, that I am trying here. Okay, calling just `this.postMessage` sounds like a good idea. Thanks.

> Also, don't you need to add similar code in _processStreamMessage below as well?

`sendStreamResponse` mostly used for signaling `*_complete` messages e.g. `pull_complete`(to resolve callbacks), and I don't think we are going to send _transfers_ in these types of messages, at least for now. Maybe we require this in future, but not sure.  @yurydelendik, do we require this?

yurydelendik: I don't understand the concern, but sendStreamRequest will have optional transfers, and I recommend just call `this.postMessage({...}, transfers)` as mentioned above

yurydelendik:
1. display/api.js has one more `pdfjs/core/network` reference that can be removed.

2. transfers can be used now

```diff
-            sink.enqueue(result.value);
+            sink.enqueue(new Uint8Array(result.value), 1, [result.value]);
```

3. and notice, the result.value array buffer is wrapped into Uint8Array, which needs to be unwrapped:

```diff
   read() {
-    return this._reader.read();
+    return this._reader.read().then(({value, done}) => {
+      if (done) { return {value: undefined, done: true}; }
+      return {value: value.buffer, done: false};
+    });;
   },
```

src/display/api.js
```diff
@@ -1601,6 +1606,62 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
+            isRangeSupported: this._fullReader.isRangeSupported,
+            contentLength: this._fullReader.contentLength,
+          });
+        });
+
+        return headersCapability.promise;
+      }, this);
+
+      messageHandler.on('GetRangeReader', function(data, sink) {
+        this._PDFNetworkStream = new PDFNetworkStream(data.data);
```
yurydelendik: no need to create new PDFNetworkStream (and data field can be removed)

mukulmishra18: Done in new commit. Thanks.

src/shared/util.js
mukulmishra18: Just thinking, do we really need same changes in [sendStreamResponse](https://github.com/mozilla/pdf.js/blob/master/src/shared/util.js#L1486)? like:

```javascript
let sendStreamResponse = ({ stream, transfers, success, reason, }) => {
  this.postMessage({ sourceName, targetName, stream, 
                     success, streamId, reason, }, transfers);
};
```

src/core/worker.js
```diff
@@ -197,6 +197,126 @@ IPDFStreamRangeReader.prototype = {
 }
 
 /** @implements {IPDFStream} */
+function PDFNetworkStream(data, msgHandler) {
+  this.data = data;
+  let source = data.source;
+  this._contentLength = source.length;
+  this.msgHandler = msgHandler;
```
Snuffleupagus: Nit: Let's place `this.msgHandler = msgHandler;` right after the `this.data = data;` line, to group the properties that are set from the provided parameters.

src/core/worker.js
```diff
@@ -197,6 +197,126 @@ IPDFStreamRangeReader.prototype = {
...
+  this._msgHandler = networkStream.msgHandler;
+  this._data = networkStream.data;
+  this.onProgress = null;
+
+  this._contentLength = null;
+  this._isRangeSupported = false;
+  this._isStreamingSupported = false;
+
+  let readableStream = this._msgHandler.sendWithStream('GetReader',
+                         this._data);
```
Snuffleupagus: Nit: Indentation, let's align this with `'GetReader'` instead.

src/core/worker.js
```diff
@@ -197,6 +197,126 @@ IPDFStreamRangeReader.prototype = {
...
+
+  this._reader = readableStream.getReader();
+
+  this._headersReady = this._msgHandler.sendWithPromise('ReaderHeadersReady').
+    then((data) => {
```
Snuffleupagus: Nit: Two more spaces of indentation on this line please.

src/core/worker.js
```diff
@@ -197,6 +197,126 @@ IPDFStreamRangeReader.prototype = {
...
+  },
+};
+
+/** @implements {IPDFStreamRangeReader} */
+function PDFNetworkStreamRangeReader(begin, end, networkStream) {
+  this._msgHandler = networkStream.msgHandler;
+  this.onProgress = null;
+
+  let readableStream = this._msgHandler.sendWithStream('GetRangeReader',
+                        { begin, end, });
```
Snuffleupagus: Nit: Indentation, let's align this directly under 'GetRangeReader' instead.

src/display/api.js
```diff
@@ -1231,8 +1232,8 @@ var PDFWorker = (function PDFWorkerClosure() {
     // pdf.worker.js file is needed.
     if (typeof PDFJSDev === 'undefined' || !PDFJSDev.test('PRODUCTION')) {
       if (typeof SystemJS === 'object') {
-        Promise.all([SystemJS.import('pdfjs/core/network'),
-                     SystemJS.import('pdfjs/core/worker')]).then((modules) => {
+        Promise.resolve(SystemJS.import('pdfjs/core/worker')).
+          then((modules) => {
```
Snuffleupagus: Nit: Please indent *this* line with two more spaces, and also change it to `then(worker) => {`.

src/display/api.js
```diff
@@ -1601,6 +1605,63 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
+        sink.onPull = () => {
+          this._fullReader.read().then(function({ value, done, }) {
+            if (done) {
+              sink.close();
+              return;
+            }
+            // Enqueue data chunk into sink, and transfer it
+            // to other side as `Transferable` object.
+            sink.enqueue(new Uint8Array(value), 1, [value]);
+          }).catch(function(reason) {
```
Snuffleupagus: No need for an intermediate function here, `}).catch(sink.error);` should work just as well.

src/display/api.js
```diff
@@ -1601,6 +1605,63 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
+          this._PDFNetworkStream.getRangeReader(data.begin, data.end);
+
+        sink.onPull = () => {
+          this._rangeReader.read().then(function({ value, done, }) {
+            if (done) {
+              sink.close();
+              return;
+            }
+            sink.enqueue(new Uint8Array(value), 1, [value]);
+          }).catch(function(reason) {
```
Snuffleupagus: No need for an intermediate function here, }).catch(sink.error); should work just as well.

src/shared/util.js
```diff
@@ -1433,13 +1433,14 @@ MessageHandler.prototype = {
     let targetName = data.sourceName;
     let capability = createPromiseCapability();
 
-    let sendStreamRequest = ({ stream, chunk, success, reason, }) => {
-      this.comObj.postMessage({ sourceName, targetName, stream, streamId,
-                                chunk, success, reason, });
+    let sendStreamRequest = ({ stream, chunk, transfers,
+                               success, reason, }) => {
+      this.postMessage({ sourceName, targetName, stream, streamId,
+                           chunk, success, reason, }, transfers);
```
Snuffleupagus: Nit: alignment.

src/core/worker.js
mukulmishra18: Done in new commit. Thanks.

src/core/worker.js
mukulmishra18: Done. Thanks

src/core/worker.js
mukulmishra18: Done.

src/core/worker.js
mukulmishra18: Done. Thanks.

src/display/api.js
mukulmishra18: Done in new commit. Thanks.

src/display/api.js
mukulmishra18: Sure, Thanks for pointing out.

src/display/api.js
mukulmishra18: Done. Thanks.

src/shared/util.js
mukulmishra18: Sorry, done with new commit. Thanks.

src/worker_loader.js
```diff
@@ -28,7 +28,6 @@ importScripts('./shared/compatibility.js');
 importScripts('../node_modules/systemjs/dist/system.js');
 importScripts('../systemjs.config.js');
 
-Promise.all([SystemJS.import('pdfjs/core/network'),
-             SystemJS.import('pdfjs/core/worker')]).then(function () {
+Promise.resolve(SystemJS.import('pdfjs/core/worker')).then(function () {
```
yurydelendik: Here and in api.js, you can directly use `SystemJS.import` without wrapping with `Promise.resolve`:
```
SystemJS.import('pdfjs/core/worker').then(function () {...
```

src/core/worker.js
```diff
@@ -197,6 +197,138 @@ IPDFStreamRangeReader.prototype = {
...
+/** @implements {IPDFStreamReader} */
+function PDFNetworkStreamReader(networkStream) {
+  this._msgHandler = networkStream.msgHandler;
+  this._data = networkStream.data;
+  this.onProgress = null;
+
+  this._contentLength = null;
+  this._isRangeSupported = false;
+  this._isStreamingSupported = false;
+  let queueingStrategy = { highWaterMark: 100,
```
yurydelendik: let's remove queueingStrategy from here and range reader, or make explicit highWaterMark:1 and size:()=>1

src/display/api.js
```diff
@@ -1601,6 +1604,59 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
+            isRangeSupported: this._fullReader.isRangeSupported,
+            contentLength: this._fullReader.contentLength,
+          });
+        });
+
+        return headersCapability.promise;
+      }, this);
+
+      messageHandler.on('GetRangeReader', function(data, sink) {
+        this._rangeReader =
```
yurydelendik: change to `let _rangeReader` it must be unique for every new range request.

src/worker_loader.js
mukulmishra18: Done in new commit. Thanks.

src/core/worker.js
mukulmishra18: Removed in new commit. Thanks.

src/display/api.js
mukulmishra18: Done. Thanks.

src/display/network.js
```diff
@@ -592,8 +592,209 @@ PDFNetworkStreamRangeRequestReader.prototype = {
...
 
+/** @implements {IPDFStream} */
+var PDFWorkerStream = (function PDFWorkerStreamClosure() {
+  function PDFWorkerStream(params, msgHandler) {
+    this._queuedChunks = [];
+    var initialData = params.initialData;
+    if (initialData && initialData.length > 0) {
+      this._queuedChunks.push(initialData);
+    }
+    this._msgHandler = msgHandler;
```
yurydelendik: we don't need this._msgHandler and msgHandler

mukulmishra18: Sorry, done with new commit.

src/display/api.js
```diff
@@ -1577,30 +1579,97 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
       }
 
+      messageHandler.on('GetReader', function(data, sink) {
+        data.disableRange = true;
+        data.disableStream = true;
+        if (data.chunkedViewerLoading) {
+          this._networkStream = new PDFWorkerStream(data.source);
+        } else {
+          this._networkStream = new PDFNetworkStream(data);
+        }
```
yurydelendik: move this code to the WorkerTransport constructor and remove this.networkStreamCapability promise

mukulmishra18: Done with new commit. Thanks.

src/display/api.js
```diff
@@ -1522,6 +1522,14 @@ var WorkerTransport = (function WorkerTransportClosure() {
     this.destroyCapability = null;
     this._passwordCapability = null;
 
+    this._networkStream = null;
+    this._fullReader = null;
+    if (pdfDataRangeTransport /* === true = chunkedViewerLoading */) {
```
Snuffleupagus: Nit: The comment isn't completely easy to read at first, and furthermore I'm not sure how much value it adds!?

mukulmishra18: Removed with new commit. Thanks.

src/display/transport_stream.js
```diff
@@ -0,0 +1,241 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { assert, createPromiseCapability } from '../shared/util';
+
+ /** @implements {IPDFStream} */
+var PDFDataTransportStream = (function PDFDataTransportStreamClosure() {
+  function PDFDataTransportStream(params, pdfDataRangeTransport) {
+    this.transportReadyCapability = createPromiseCapability();
```
yurydelendik: we don't need transportReadyCapability, since we moved stream to the main thread and PDFDataTransportStream can cache data as soon as addProgressiveReadListener bound. If I'm not mistaken, we can safely remove this guard.

src/display/transport_stream.js
```diff
@@ -0,0 +1,241 @@
...
+
+    this._pdfDataRangeTransport = pdfDataRangeTransport;
+    this._isRangeSupported = !(params.disableRange);
+    this._isStreamingSupported = !(params.disableStream);
+    this._contentLength = params.length;
+
+    this._fullRequestReader = null;
+    this._rangeReaders = [];
+
+    if (this._pdfDataRangeTransport) {
```
yurydelendik: it shall always be true, can you add `assert(pdfDataRangeTransport)` as the top of the function.

src/display/transport_stream.js
```diff
@@ -0,0 +1,241 @@
...
+    cancelAllRequests:
+        function PDFDataTransportStream_cancelAllRequests(reason) {
+      if (this._fullRequestReader) {
+        this._fullRequestReader.cancel(reason);
+      }
+      var readers = this._rangeReaders.slice(0);
+      readers.forEach(function (rangeReader) {
+        rangeReader.cancel(reason);
+      });
+    },
```
yurydelendik: add `this.pdfDataRangeTransport.abort();` here

src/display/transport_stream.js
mukulmishra18: Actually I am using `this.transportReadyCapability` to signal `this._pdfDataRangeTransport.transportReady()` [here](https://github.com/mozilla/pdf.js/pull/8617/commits/e18b71d617a561e819087ed334c52884de56402e#diff-3089181c2189d3ef27f3dbd49ca08965R50), `this.transportReadyCapability` is resolved only when we get [PDFManagerReady](https://github.com/mozilla/pdf.js/pull/8617/commits/e18b71d617a561e819087ed334c52884de56402e#diff-0ecad279ed2252e3eb47a4d96ec1463cL1731) message from worker.

I am doing something wrong here?

yurydelendik: you can signal `transportReady` in the constructor after event binding and keep PDFManagerReady body empty.

mukulmishra18: Done in new commit. Thanks for clarification here.

src/display/transport_stream.js
mukulmishra18: Done with new commit. Thanks.

src/display/transport_stream.js
mukulmishra18: Done.

src/display/api.js
```diff
@@ -281,8 +283,18 @@ function getDocument(src, pdfDataRangeTransport,
       if (task.destroyed) {
         throw new Error('Loading aborted');
       }
+      let data = { source: params,
+                   disableRange: getDefaultSetting('disableRange'), };
```
yurydelendik: we don't need to create data temporate, let's use `params` directly with PDFDataTransportStream. For PDFNetworkStream inline the object literal as parameter.

src/display/network.js
```diff
@@ -17,7 +17,6 @@ import {
   assert, createPromiseCapability, globalScope, isInt, MissingPDFException,
   UnexpectedResponseException
 } from '../shared/util';
-import { setPDFNetworkStreamClass } from './worker';
```
yurydelendik: add setPDFNetworkStreamClass to the api.js and import it from there

src/pdf.worker.js
```diff
@@ -21,9 +21,4 @@ var pdfjsBuild = PDFJSDev.eval('BUNDLE_BUILD');
 
 var pdfjsCoreWorker = require('./core/worker.js');
 
-if (typeof PDFJSDev === 'undefined' ||
-    !PDFJSDev.test('FIREFOX || MOZCENTRAL')) {
-  require('./core/network.js');
-}
```
yurydelendik: Move this code to the src/pdf.js

yurydelendik (commented):
Some minor changes

yurydelendik:
- [ ] reduce amount of information sent via GetDocRequest
- [ ] check if we can remove ReaderDocProgress round trip
- [ ] test in different configurations (including IE9)

Snuffleupagus:
Are the two current commits in this PR intended to be squashed before landing? If so, then this question can be ignored, otherwise I'm wondering:

In the "Adds Streams API support for networking task of PDF.js project." commit you add [`PDFNetworkStream`](https://github.com/mozilla/pdf.js/pull/8617/commits/995197212b755f988c6a91bc315f50df7f10d2d5#diff-780c33139a29f60c0cc45b80fb05f900R200), [`PDFNetworkStreamReader`](https://github.com/mozilla/pdf.js/pull/8617/commits/995197212b755f988c6a91bc315f50df7f10d2d5#diff-780c33139a29f60c0cc45b80fb05f900R234), and [`PDFNetworkStreamRangeReader`](https://github.com/mozilla/pdf.js/pull/8617/commits/995197212b755f988c6a91bc315f50df7f10d2d5#diff-780c33139a29f60c0cc45b80fb05f900R290) to the `src/core/worker` file. Then in the "Moves PDFWorkerStream to main side." commit that code is immediately removed from said file.
Adding a bunch of code in one commit, just to remove it in the next seem to cause unnecessary churn in the code. So, if the commits aren't intended to be squashed, can we try to avoid this?

src/core/worker.js
```diff
@@ -198,218 +198,123 @@ IPDFStreamRangeReader.prototype = {
 
 /** @implements {IPDFStream} */
 var PDFWorkerStream = (function PDFWorkerStreamClosure() {
-  function PDFWorkerStream(params, msgHandler) {
-    this._queuedChunks = [];
-    var initialData = params.initialData;
-    if (initialData && initialData.length > 0) {
-      this._queuedChunks.push(initialData);
-    }
+  function PDFWorkerStream(data, msgHandler) {
```
mukulmishra18: Actually, we don't need to pass `source` or `data` to `PDFWorkerStream` here, this can changed to accept `msgHandler` as it's single parameter, something like: 
 `function PDFWorkerStream(msgHandler) {...`.

src/display/network.js
mukulmishra18: Just want to ask, why we are using `setPDFNetworkStreamClass` function to set/import(kind of) `PDFNetworkStream` here, and not directly importing it, like `import { PDFNetworkStream } from './network';`

mukulmishra18:
> Are the two current commits in this PR intended to be squashed before landing?

Yeah, you are right. All commits will finally squashed into a single commit before merging. I just want to make sure that the changes for each TODO are distinguishable. Different commit contains changes for different TODO.

src/core/worker.js
yurydelendik: sounds good to me

src/core/worker.js
```diff
@@ -458,6 +363,7 @@ var WorkerMessageHandler = {
     // This context is actually holds references on pdfManager and handler,
     // until the latter is destroyed.
     var pdfManager;
+    var fullRequest;
```
yurydelendik: save only onReaderDocProgress, if needed, but I prefer to not do a round trip from main thread, to just send the same data back.

src/display/network.js
yurydelendik: The network.js must not be included in the firefox/mozcentral builds of pdf.js (as we did in pdf.worker.js) to reduce size. Notice that `gulp firefox`'s pdf.worker.js did not have any network.js code.

src/core/worker.js
```diff
@@ -538,24 +444,19 @@ var WorkerMessageHandler = {
...
-          pdfStream = new PDFNetworkStream(data);
-        }
+        pdfStream = new PDFWorkerStream(data, handler);
       } catch (ex) {
         pdfManagerCapability.reject(ex);
         return pdfManagerCapability.promise;
       }
 
-      var fullRequest = pdfStream.getFullReader();
+      fullRequest = pdfStream.getFullReader();
```
mukulmishra18: Done.

src/display/api.js
mukulmishra18: Done, Thanks.

src/pdf.worker.js
mukulmishra18: Done, thanks.

src/core/worker.js
mukulmishra18: Done with new commit. Also removed `ReaderDocProgress` round trip. Thanks.

src/core/worker.js
```diff
@@ -555,12 +452,7 @@ var WorkerMessageHandler = {
             !fullRequest.isRangeSupported) {
           // If stream or range are disabled, it's our only way to report
           // loading progress.
-          fullRequest.onProgress = function (evt) {
-            handler.send('DocProgress', {
-              loaded: evt.loaded,
-              total: evt.total,
-            });
-          };
+          handler.send('DocProgress');
```
Snuffleupagus: How can this work if you're not actually sending any data with the message!?
I've quickly tested the viewer, and the progress bar is now broken, which I'm assuming might be connected to this!

src/display/api.js
```diff
@@ -281,8 +293,19 @@ function getDocument(src, pdfDataRangeTransport,
       if (task.destroyed) {
         throw new Error('Loading aborted');
       }
+
+      let networkStream;
+      if (rangeTransport) {
+        networkStream = new PDFDataTransportStream(params, rangeTransport);
+      } else {
+        networkStream =
```
Snuffleupagus: Nit: The indentation looks weird here, why not simply the following instead?
```javascript
networkStream = new PDFNetworkStream({
  source: params,
  disableRange: ...,
});
```

src/core/worker.js
mukulmishra18: I am trying to fix a round trip of progress data(from main to worker, back to main) here. As we don't have any progress data at worker side, this message doesn't contain any data, it is just to notify `DocProgress` event.

src/display/api.js
mukulmishra18: Done with new commit. Thanks.
Also tried fixing viewer progress bar, can you please check if it works for you?

src/core/worker.js
mukulmishra18: Removed with new commit, as we don't need this anymore. As we are moving all networking related tasks to main side, we can easily notify `DocProgress` at main side in [ReaderHeaderReady](https://github.com/mozilla/pdf.js/pull/8617/commits/4531fea269ca195d6a72a4ab45ff00629b00fe58#diff-0ecad279ed2252e3eb47a4d96ec1463cR1625), and eliminates round trip of progress data.

src/core/worker.js
```diff
@@ -538,31 +440,14 @@ var WorkerMessageHandler = {
...
         pdfManagerCapability.reject(ex);
         return pdfManagerCapability.promise;
       }
 
       var fullRequest = pdfStream.getFullReader();
       fullRequest.headersReady.then(function () {
-        if (!fullRequest.isStreamingSupported ||
-            !fullRequest.isRangeSupported) {
-          // If stream or range are disabled, it's our only way to report
-          // loading progress.
```
yurydelendik: I'm thinking this comment will be useful in the ReaderHeadersReady

src/display/api.js
Snuffleupagus: > Also tried fixing viewer progress bar, can you please check if it works for you?

Yes, it appears that the progress bar is now updating correctly in all configurations that I've tested (with/without `disableRange`, `disableStream`, `disableAutoFetch`, and combinations of them).

src/display/api.js
```diff
@@ -1574,32 +1594,72 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
+        };
+      }, this);
+
+      messageHandler.on('ReaderHeadersReady', function(data) {
+        let headersCapability = createPromiseCapability();
+        let fullReader = this._fullReader;
+
+        fullReader.headersReady.then(() => {
+          if (!fullReader.isStreamingSupported ||
+              !fullReader.isRangeSupported) {
```
Snuffleupagus: Nit: You can probably reduced the overall indentation a bit by using an early return here, i.e.
```javascript
  if (fullReader.isStreamingSupported && fullReader.isRangeSupported) {
    return;
  }
  fullReader.onProgress = (evt) => {
    ...
```

Snuffleupagus:
@mukulmishra18 Can you please rebase this PR on the *latest* `master` (including PR #8643), since otherwise this cannot be tested in Internet Explorer.

src/display/api.js
```diff
@@ -1574,32 +1593,74 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
+          _rangeReader.read().then(function({ value, done, }) {
+            if (done) {
+              sink.close();
+              return;
+            }
+            sink.enqueue(new Uint8Array(value), 1, [value]);
+          }).catch(sink.error);
+        };
+
+        sink.onCancel = (reason) => {
```
Snuffleupagus: Nit: `sink.onCancel = _rangeReader.cancel` instead?

yurydelendik: nope, _rangeReader needs to be bound.

mukulmishra18: Okay, I didn't understand this. What is the problem in `sink.onCancel = _rangeReader.cancel`? Can you please explain it?

mukulmishra18:
@Snuffleupagus, PR is rebased to the new `master`. Thanks for your help with #8643 here.

src/core/worker.js
mukulmishra18: Done.

Snuffleupagus:
I've done some quick testing in Internet Explorer, and these are the results[1]:
 - IE 11; the viewer and the PDF document loads successfully and appears to work correctly.
 - IE 10; the viewer loads, but the document does *not*.
  In the console, the following message is printed: `Unhandled rejection: DataCloneError`. This may indicate that a `postMessage` is attempting to send for example an `Error` object directly, which doesn't work since the structured cloning algorithm cannot handle that.[2]
 - IE 9; the viewer and the PDF document loads successfully and appears to work correctly (albeit very slowly, which is similar to the current `master`).

---
[1] All testing was done using IE 11. The prior versions, i.e. IE 9 and 10, were tested using the emulation feature in the IE 11 devtools.

[2] Note that in the existing code, there's special handling of the `Error` case; see https://github.com/mozilla/pdf.js/blob/95093a5276d9e627717aa25411284727f312fad5/src/shared/util.js#L1285-L1288.
However, I'd suggest figuring out *why* this happens with the current patch *first*, rather than just trying to add similar checks to every `postMessage` call in the Stream-related `MessageHandler` methods/functions.

src/display/transport_stream.js
```diff
@@ -0,0 +1,239 @@
...
+      }
+      var readers = this._rangeReaders.slice(0);
+      readers.forEach(function (rangeReader) {
+        rangeReader.cancel(reason);
+      });
+      this._pdfDataRangeTransport.abort();
+    },
+  };
+
+  /** @implements {IPDFStreamReader} */
```
timvandermeij: Are these interfaces defined somewhere? If not, it may be good to make a file similar to https://github.com/mukulmishra18/pdf.js/blob/dfc67021597f7ce57476f478274a9929c2bd5931/web/interfaces.js that contains their definitions.

Snuffleupagus: > Are these interfaces defined somewhere?

Yes, please see https://github.com/mozilla/pdf.js/blob/ca3c08f12b1b1e70523d06b54232d26e07244f43/src/core/worker.js#L55-L197.

timvandermeij: Perfect! Thanks.

mukulmishra18:
Okay, so after some testing I found that we are not sending the first data chunk from [worker.js file](https://github.com/mukulmishra18/pdf.js/blob/network-streaming/src/core/worker.js#L510). This issue arises because, at the first run of [readChunk](https://github.com/mukulmishra18/pdf.js/blob/network-streaming/src/core/worker.js#L490-#L532) block, we didn't have `pdfManager` initialized, so data `chunk` is pushed into `cachedChunks` and hence never sent to the _chunked_stream_, results in data loss.

A potential solution of this is, we should check for any `cachedChunk` in [if(pdfManager) {...](https://github.com/mukulmishra18/pdf.js/blob/network-streaming/src/core/worker.js#L510) and send it to _chunked_stream_, if we have any.

src/core/worker.js
```diff
@@ -623,6 +508,16 @@ var WorkerMessageHandler = {
...
+              // There may be a chance that pdfManager is not initialized
+              // for some previous runs of this(readchunk) block of code.
+              // Be sure to send all cached chunks, if any, to chunked_stream
+              // via pdf_manager.
+              if (cachedChunks.length !== 0) {
+                for (let i = 0; i < cachedChunks.length; i++) {
+                  pdfManager.sendProgressiveData(cachedChunks[i]);
+                }
+                cachedChunks = [];
+              }
```
yurydelendik: Can we move that to the https://github.com/mozilla/pdf.js/pull/8617/files#diff-780c33139a29f60c0cc45b80fb05f900R443  (just after we created pdfManager)?

mukulmishra18: Done. Thanks. Also I think we don't need to check for `if (cachedChunks.length !== 0) {..`, as it will not invoke `pdfManager.sendProgressiveData` if `cachedChunks.length === 0`

src/display/api.js
```diff
@@ -1297,6 +1298,7 @@ var PDFWorker = (function PDFWorkerClosure() {
 
     this.name = name;
     this.destroyed = false;
+    this.postMessageTransfers = false;
```
yurydelendik: Let's make it `true` by default

src/display/api.js
```diff
@@ -1389,6 +1391,7 @@ var PDFWorker = (function PDFWorkerClosure() {
             }
             var supportTypedArray = data && data.supportTypedArray;
             if (supportTypedArray) {
+              this.postMessageTransfers = data.supportTransfers;
```
yurydelendik: can you move it below in the `if (!data.supportTransfers) {` block as `this.postMessageTransfers = false;`

src/display/api.js
mukulmishra18: Done. Thanks.

src/display/api.js
mukulmishra18: Done with new commit. Thanks.

mukulmishra18:
Running unit test locally, I am getting `Uncaught Error: Unknown action from worker: RequestDataRange`(worker to main). But this is really strange, as there is no place in the code where I am sending `RequestDataRange` message.

src/shared/util.js
```diff
@@ -1222,11 +1222,25 @@ function resolveCall(fn, args, thisArg = null) {
   });
 }
 
+function wrapReason(reason) {
```
yurydelendik: let's wrap only if reason is object, otherwise return reason as is

src/shared/util.js
```diff
@@ -1222,11 +1222,25 @@ function resolveCall(fn, args, thisArg = null) {
...
+  let name = reason.name;
+  let message = reason.message;
+
+  switch (name) {
+    case 'MissingPDFException':
+      return new MissingPDFException(message);
+    case 'UnexpectedResponseException':
+      return new UnexpectedResponseException(message);
```
yurydelendik: status is needed here

src/shared/util.js
```diff
@@ -1222,11 +1222,25 @@ function resolveCall(fn, args, thisArg = null) {
...
+    default:
+      return new UnknownErrorException(message);
```
yurydelendik: details parameter is needed here

src/shared/util.js
```diff
@@ -1222,11 +1222,25 @@ function resolveCall(fn, args, thisArg = null) {
...
+  }
+}
+
 function resolveOrReject(capability, success, reason) {
   if (success) {
     capability.resolve();
   } else {
-    capability.reject(reason);
+    capability.reject(wrapReason(reason));
```
yurydelendik: let's not wrap it here

src/shared/util.js
```diff
@@ -1482,6 +1498,7 @@ MessageHandler.prototype = {
     let streamId = data.streamId;
 
     let sendStreamResponse = ({ stream, success, reason, }) => {
+      reason = reason === undefined ? reason : wrapReason(reason);
```
yurydelendik: since we are adding reason check at wrapReason, let's remove it from here and add to callsites instead

src/shared/util.js
mukulmishra18: Done with new commit. Thanks.

src/shared/util.js
mukulmishra18: Done. Thanks.

src/shared/util.js
mukulmishra18: Done. Thanks.

src/shared/util.js
mukulmishra18: Done.

src/shared/util.js
mukulmishra18: Done with new commit. Thanks for your reviews.

mukulmishra18:
This seems to be an intermittent issue `TypeError: this.streamControllers[data.streamId] is undefined thrown `, and not thrown locally while testing.

But with a quick look at [_processStreamMessage](https://github.com/mozilla/pdf.js/blob/master/src/shared/util.js#L1479) of _MessageHandler_, I find out that, we are loosely checking(directly checking `streamController[data.streamId].isClosed`) or not checking for existence of `this.streamController[data.streamId]` in [enqueue](https://github.com/mozilla/pdf.js/blob/master/src/shared/util.js#L1535), [error](https://github.com/mozilla/pdf.js/blob/master/src/shared/util.js#L1548) and [close](https://github.com/mozilla/pdf.js/blob/master/src/shared/util.js#L1540) stream cases. As there maybe a chance(high), that we will be calling these three stream cases even after _streamController_ is deleted.

This issue can be reproduced by simply modifying one of util_stream_spec.js tests, something like:

```javascript
messageHandler2.on('fakeHandler', function (data, sink) {
  sink.onPull = function () { log += 'p'; };
  sink.onCancel = function (reason) { log += 'c'; };
  log += '0';
  sink.ready.then(() => {
    log += '1';
    sink.enqueue([1, 2, 3, 4], 4);
    return sink.ready;
  }).then(() => {
    log += 'c';
    sink.close();
    return sleep(10);
  }).then(() => {
    // sink.error(); or sink.close(); can also be called.
    sink.enqueue([1, 2, 1], 3);
  });
});
```

Adding extra checks for `this.streamController[data.streamId]` in `enqueue`, `error` and `close` stream cases should potentially resolve this issue.
@yurydelendik can you please check if this looks reasonable to you?

yurydelendik:
> Adding extra checks for this.streamController[data.streamId] in enqueue, error and close stream cases should potentially resolve this issue.

Based on your example/test, it is incorrect behavior of the sink consumer since enqueue is delivering after close. By ignoring it at enqueue only hide the problem. So adding check to just ignore command will be a bad idea.

yurydelendik:
One minor issue to fix, when `#disableRange=true&disableStream=true` is used with extension, the load fails since we don't have access to PDFNetworkStream and we have only data. I wonder if we need to fallback to PDFDataTransportStream in this case.

src/shared/util.js
```diff
@@ -1489,6 +1509,7 @@ MessageHandler.prototype = {
     let streamId = data.streamId;
 
     let sendStreamResponse = ({ stream, success, reason, }) => {
+      reason = wrapReason(reason);
```
yurydelendik: no need for this

src/shared/util.js
```diff
@@ -1431,13 +1445,15 @@ MessageHandler.prototype = {
     let targetName = data.sourceName;
     let capability = createPromiseCapability();
 
-    let sendStreamRequest = ({ stream, chunk, success, reason, }) => {
-      this.comObj.postMessage({ sourceName, targetName, stream, streamId,
-                                chunk, success, reason, });
+    let sendStreamRequest = ({ stream, chunk, transfers,
+                               success, reason, }) => {
+      reason = wrapReason(reason);
```
yurydelendik: not need for this

yurydelendik (commented):
wrapReason does not need to be present in the sendStreamRequest, sendStreamResponse or resolveOrReject. Move wrapReason() call to the places where `data.reason` is used/retrieved.

yurydelendik:
Thank you for the patch.

Snuffleupagus:
It seems that this patch may have caused some weird behaviour in the Firefox addon with `disableAutoFetch = true`, as evident by strange loading bar behaviour when opening [http://mirrors.ctan.org/info/lshort/english/lshort.pdf#disableStream=true&disableAutoFetch=true](http://mirrors.ctan.org/info/lshort/english/lshort.pdf#disableStream=true&disableAutoFetch=true).[1]

Looking at data received through `onProgress`, it seems that with this patch the entire PDF file is fetched at *once* thus bypassing the `disableAutoFetch` setting. When doing skip-cache reloads, I'm getting `percent` equal to (in order) `101`, `100`, `101` in [PDFViewerApplication.progress](https://github.com/mozilla/pdf.js/blob/343b4dc2b61c88f9b549f212b5d91a8b7acd4209/web/app.js#L849).

---
[1] It may be necessary to do "skip-cache" reloads, i.e. <kbd>Ctrl</kbd>+<kbd>F5</kbd> to ensure that the PDF data isn't cached.

yurydelendik:
> It seems that this patch may have caused some weird behaviour in the Firefox addon

I was encountering with behavior before: the initial XHR fetches quickly entire file when cached. I think it's normal for small files. We just need to pay attention if data and events is handled properly in chunks and streams.

Snuffleupagus:
> I was encountering with behavior before: the initial XHR fetches quickly entire file when cached.

Huh, I'm seeing this when doing skip-cache reloads (as indicated above), so I don't think caching should be relevant then!?

yurydelendik:
Not sure I'm getting the same behavior locally: PDFViewerApplication.progress reports proper percent values in ascending order.

Snuffleupagus:
> Not sure I'm getting the same behavior locally: PDFViewerApplication.progress reports proper percent values in ascending order.

I can confirm that it works locally (with `gulp server`), but I'm seeing the weird behaviour with the *Firefox addon*.

yurydelendik:
> but I'm seeing the weird behaviour with the Firefox addon.

FWIW I'm testing on Nightly (e10s) with current development addon.

PR #8618, opened 2017-07-06
Remove the scale-not-initialized hack from `webViewerResize` (in app.js)

Snuffleupagus:
- Remove the scale-not-initialized hack from `webViewerResize` (in app.js)

   Since we no longer, after PR #8555, allow changing the scale until the document is loaded, that hack is no longer necessary. Furthermore, no part of that event handling function needs to run unless a document is loaded.

   The reason that this hack was initially added, is that previously the `ViewHistory` might be updated *before* `PDFViewerApplication.setInitialView` had run (in some cases leading to incorrect inital document scale). Since that is no longer possible, this is now dead code.

- Ensure that the document is rendered on load, no matter what happens, by always calling `PDFViewer.update` *after* the initial position has been set

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1790,19 +1790,18 @@ function webViewerUpdateViewarea(evt) {
 }
 
 function webViewerResize() {
-  let currentScaleValue = PDFViewerApplication.pdfViewer.currentScaleValue;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #8621, opened 2017-07-06
Bump minimum eslint version to 3.19.0

Rob--W:
The previously referenced eslint version is not compatible with our code base.
For example, rule "prefer-promise-reject-errors" requires 3.14.0, while package.json specifies `^3.11.1`.


Snuffleupagus (approved):
Looks good, thanks!

PR #8627, opened 2017-07-07
Fallback on font widths if CFF data is broken

### LINKED ISSUE #8591: CFF font widths not parsed properly
When reading the widths in a CFF font we need to fill the width array using default and nominal width (see the spec on how this should work).

yurydelendik:
Fixes #8591

### DIFF
--- a/src/core/cff_parser.js
+++ b/src/core/cff_parser.js
@@ -268,12 +268,14 @@ var CFFParser = (function CFFParserClosure() {
       cff.charset = charset;
       cff.encoding = encoding;
 
-      var charStringsAndSeacs = this.parseCharStrings(
-                                  charStringIndex,
-                                  topDict.privateDict.subrsIndex,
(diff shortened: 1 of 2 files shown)


src/core/cff_parser.js
```diff
@@ -644,7 +648,9 @@ var CFFParser = (function CFFParserClosure() {
                                        globalSubrIndex);
         }
         if (state.width !== null) {
```
brendandahl: I thought it was strange that the widths were broken in TaroUTR50SortedList112.pdf so I looked in acrobat. Font "QEGUTH+KozMinPro-Regular" is a cid font so we should also look in the sub font dict's private dict for the nominal width and default width. In acrobat there are valid values there. There's some code above here that gets the private dict see `fdArray[fdIndex].privateDict`

brendandahl (changes requested):

PR #8628, opened 2017-07-08
Use the Width/Height of image dict in JpegStream

### LINKED ISSUE #8614: not possible to render (Error during JPEG image loading)
Link to PDF file (or attach file here): [OBW-OVK.pdf](https://github.com/mozilla/pdf.js/files/1125123/OBW-OVK.pdf)

Configuration:

- Webbrowser and its version: Chrome 59.0.3071.115

- Operatingsystem and its version: Microsoft Windows 7 Enterprise SP1

- PDF.js version: 1.8.516

- Is an extension: no

1. Load attached pdf-file
2.

Pdf has been rendered as blank pages. Error trace:

 Error: Invalid XRef stream header
util.js:296     at error (https://mozilla.github.io/pdf.js/build/pdf.worker.js:355:17)
    at XRef_readXRef [as readXRef] (https://mozilla.github.io/pdf.js/build/pdf.worker.js:21322:29)
    at XRef_parse [as parse] (https://mozilla.github.io/pdf.js/build/pdf.worker.js:20992:28)
    at PDFDocument_setup [as setup] (https://mozilla.github.io/pdf.js/build/pdf.worker.js:29900:17)
    at PDFDocument_parse [as parse] (https://mozilla.github.io/pdf.js/build/pdf.worker.js:29789:12)
    at _util.Util.inherit.ensure (https://mozilla.github.io/pdf.js/build/pdf.worker.js:39550:28)
    at Promise (<anonymous>)
    at LocalPdfManager_ensure [as ensure] (https://mozilla.github.io/pdf.js/build/pdf.worker.js:39545:14)
    at LocalPdfManager.BasePdfManager_ensureDoc [as ensureDoc] (https://mozilla.github.io/pdf.js/build/pdf.worker.js:39497:19)
util.js:282 Warning: Indexing all PDF objects
app.js:1056 PDF f5a022ce9c1cfd977eedadde6a39bd36 [1.2 - / -] (PDF.js: 1.8.516)
util.js:282 Warning: Error during JPEG image loading
util.js:282 Warning: Dependent image isn't ready yet
util.js:282 Warning: Error during JPEG image loading
util.js:282 Warning: Dependent image isn't ready yet

Rob--W:
This patch provides a work-around that fixes #8614.

### DIFF
--- a/src/core/stream.js
+++ b/src/core/stream.js
@@ -871,6 +871,45 @@ var PredictorStream = (function PredictorStreamClosure() {
  * DecodeStreams.
  */
 var JpegStream = (function JpegStreamClosure() {
+  /**
+   * Ensures that the width/height of the JPEG file match the given parameters.
+   *
+   * The relevant parts of the JPEG format are documented at section B2.2 of
+   * https://www.w3.org/Graphics/JPEG/itu-t81.pdf
+  */
+  function setJpegDimensions(bytes, width, height) {
+    function getUint16(offset) {
+      return (bytes[offset] << 8) | bytes[offset + 1];
+    }
+    function setUint16(offset, value) {
+      bytes[offset] = (value >> 8) & 0xFF;
+      bytes[offset + 1] = value & 0xFF;
+    }
+    if (getUint16(0) !== 0xFFD8) {
+      return;  // Invalid JPEG header.
+    }
+    var i = 2;  // The segments start right after the leading 0xFFD8.
+    // The minimum size of a SOF frame is 10.
+    var maxIndex = bytes.length - 10;
+    while (i < maxIndex) {
+      if (bytes[i] !== 0xFF) {
+        return;  // Invalid start of SOF marker.
+      }
+      var marker = bytes[i + 1];
+      if (marker === 0xC0 || marker === 0xC2) { // SOF0 or SOF2
+        if (getUint16(i + 5) !== height) {
+          setUint16(i + 5, height);
+        }
+        if (getUint16(i + 7) !== width) {
+          setUint16(i + 7, width);
+        }
+      }
+      // Advance to the next frame marker.
+      i += getUint16(i + 2) + 2;
+    }
+    return bytes;
+  }
+
   function JpegStream(stream, maybeLength, dict, params) {
     // Some images may contain 'junk' before the SOI (start-of-image) marker.
     // Note: this seems to mainly affect inline images.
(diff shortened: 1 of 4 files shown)


Rob--W:
Oops, misspelled file name of new test in test_manifest.json. There aren't any regressions with our current set of tests, which looks good.

yurydelendik:
Not sure what to do yet, but fixing bytes did not work really well in the past with font data. Few questions: can Image detect if bad image is loaded? If yes, will it be better if we will send error back to worker, and fallback to our decoder to read it to the best of its "ability"?

yurydelendik:
(from https://github.com/mozilla/pdf.js/issues/8614#issuecomment-313899388) if origin of image is TIFF, it's probably only height which is corrupted. It will be probably nice to have recovery to incomplete image in JPEG anyway to see partial image.

Rob--W:
Indeed, only the height has been corrupted. How did you derive "it's only the height which is corrupted" from the assumption "origin of image is TIFF"? II don't see an obvious relation between the two.

> Not sure what to do yet, but fixing bytes did not work really well in the past with font data (too brittle).

I am specifically fixing the issue of width-height mismatch, and perform strict validation to make sure that I only change JPEG data that are otherwise structurally valid.

> Few questions: can Image detect if bad image is loaded?

Yes, the image will fail to load in the main thread (in browsers only, not in our Node.js code where domutils assumes that any image assigned to the `Image` constructor is valid).

> If yes, will it be better if we will send error back to worker, and fallback to our decoder to read it to the best of its "ability"?

Possibly. I guess that we have to add special logic to fix up the image. What other kind of errors do you want to catch with this approach?

yurydelendik:
> What other kind of errors do you want to catch with this approach?

If possible, incomplete image in terms of byte stream and error in the middle of byte stream.

Snuffleupagus:
I should have responded (a lot) sooner here given that I was flagged for review, but considering the comments made by @yurydelendik forgot to do so; sorry about that!

> [...] but fixing bytes did not work really well in the past with font data (too brittle).

I completely agree with this assessment, since I too would worry about breaking "good" images by doing this.

The assumption that the width/height as specified in the image dictionary must agree *exactly* with the width/height encoded in the actual image data might not be generally justifiable. Looking at the [relevant section of the PDF specification](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G6.1877438), the following passage even seem to confirm this (emphasis mine):
> During encoding, several parameters shall control the algorithm and the information loss. The values of these parameters, **which include the dimensions of the image** and the number of components per sample, **are entirely under the control of the encoder and shall be stored in the encoded data**. [...]

---

> Few questions: can Image detect if bad image is loaded? If yes, will it be better if we will send error back to worker, and fallback to our decoder to read it to the best of its "ability"?

Falling back to our built-in JPEG decoder is a really nice suggestion, to address the more general case of broken native image decoding, and it turns out that implementing this actually isn't that difficult.
Doing so uncovered the fact that the JPEG images, in the PDF file in question, contains [DNL markers](https://www.w3.org/Graphics/JPEG/itu-t81.pdf#page=49) which `src/core/jpg.js` currently doesnt support. Given the way that the decoder is currently implemented this particular marker wasn't entirely straightforward to support, but it's still doable

I've attempted to address these points in PR #9427, which thus contains an alternative solution to the issue.

Rob--W:
Closing in favor of #9427.

PR #8630, opened 2017-07-09
Remove __pdfjsdev_webpack__, use webpack options

### LINKED ISSUE #8622: __pdfjsdev_webpack__ appears in built library
`__pdfjsdev_webpack__` appears in the generated build artifacts. Can we get rid of it?* Either by not generating this string in the output, or by removing it from the input source files.

E.g. in `build/dist/build/pdf.js` (snippet originates from `src/shared/util.js`):

```javascript
function isNodeJS() {
  if (typeof __pdfjsdev_webpack__ === 'undefined') {
    return (typeof process === 'undefined' ? 'undefined' : _typeof(process)) === 'object' && process + '' === '[object process]';
  }
  return false;
}
```

\* I guess that the custom plugin was introduced to prevent webpack from stubbing "process", "require", etc. This can be disabled with the `node` option in webpack's build configuration (`node: false` since webpack 3, or an object with every value set to `false` - see https://webpack.js.org/configuration/node/).
To use `require` without it getting replaced, we can use `__non_webpack_require__`, but **only if PDF.js must always be built with webpack before it can be used** (since we use modules and a preprocessor, I guess that this is the case?).

/cc @yurydelendik

Rob--W:
Fixes #8622.

`__pdfjsdev_webpack__` was used to skip evaluating part of an AST, in order to not mangle some `require` symbols.

This PR removes `__pdfjsdev_webpack__`, and:

- Uses `__non_webpack_require__` when one wants the output to contain `require` instead of `__webpack_require__`.
- Adds options to the webpack config to prevent "polyfills" for some Node.js-specific APIs to be added.
- Use `// eslint-disable-next-line no-undef` instead of `/* globals ... */` for variables that are not meant to be used globally.

The comparison of the old and new code is shown below. It looks like an improvement:

- `__pdfjsdev_webpack__` has now disappeared from the code as expected
- An unnecessary polyfill for `global` has disappeared too!

```diff
--- build/dist/build/pdf.js	2017-07-09 15:40:49.243287314 +0200
+++ build-dist-build-pdf.js-with-patch	2017-07-09 15:26:24.824066429 +0200
@@ -88,7 +88,7 @@
 /******/ 	__w_pdfjs_require__.p = "";
 /******/
 /******/ 	// Load entry module and return exports
-/******/ 	return __w_pdfjs_require__(__w_pdfjs_require__.s = 15);
+/******/ 	return __w_pdfjs_require__(__w_pdfjs_require__.s = 14);
 /******/ })
 /************************************************************************/
 /******/ ([
@@ -96,7 +96,7 @@
 /***/ (function(module, exports, __w_pdfjs_require__) {
 
 "use strict";
-/* WEBPACK VAR INJECTION */(function(global) {
+
 
 Object.defineProperty(exports, "__esModule", {
   value: true
@@ -105,9 +105,9 @@
 
 var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };
 
-__w_pdfjs_require__(14);
+__w_pdfjs_require__(13);
 
-var _streamsLib = __w_pdfjs_require__(10);
+var _streamsLib = __w_pdfjs_require__(9);
 
 var globalScope = typeof window !== 'undefined' && window.Math === Math ? window : typeof global !== 'undefined' && global.Math === Math ? global : typeof self !== 'undefined' && self.Math === Math ? self : undefined;
 var FONT_IDENTITY_MATRIX = [0.001, 0, 0, 0.001, 0, 0];
@@ -900,10 +900,7 @@
   return ch === 0x20 || ch === 0x09 || ch === 0x0D || ch === 0x0A;
 }
 function isNodeJS() {
-  if (typeof __pdfjsdev_webpack__ === 'undefined') {
-    return (typeof process === 'undefined' ? 'undefined' : _typeof(process)) === 'object' && process + '' === '[object process]';
-  }
-  return false;
+  return (typeof process === 'undefined' ? 'undefined' : _typeof(process)) === 'object' && process + '' === '[object process]';
 }
 function createPromiseCapability() {
   var capability = {};
@@ -1444,7 +1441,6 @@
 exports.utf8StringToString = utf8StringToString;
 exports.warn = warn;
 exports.unreachable = unreachable;
-/* WEBPACK VAR INJECTION */}.call(exports, __w_pdfjs_require__(6)))
 
 /***/ }),
 /* 1 */
@@ -2400,11 +2396,11 @@
 
 var _dom_utils = __w_pdfjs_require__(1);
 
-var _font_loader = __w_pdfjs_require__(12);
+var _font_loader = __w_pdfjs_require__(11);
 
-var _canvas = __w_pdfjs_require__(11);
+var _canvas = __w_pdfjs_require__(10);
 
-var _metadata = __w_pdfjs_require__(7);
+var _metadata = __w_pdfjs_require__(6);
 
 function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }
 
@@ -2415,7 +2411,7 @@
 var pdfjsFilePath = typeof document !== 'undefined' && document.currentScript ? document.currentScript.src : null;
 var fakeWorkerFilesLoader = null;
 var useRequireEnsure = false;
-if (typeof __pdfjsdev_webpack__ === 'undefined') {
+{
   if (typeof window === 'undefined') {
     isWorkerDisabled = true;
     if (typeof require.ensure === 'undefined') {
@@ -3853,8 +3849,8 @@
 }();
 var version, build;
 {
-  exports.version = version = '1.8.524';
-  exports.build = build = '7b4887dd';
+  exports.version = version = '1.8.525';
+  exports.build = build = '0300fb01';
 }
 exports.getDocument = getDocument;
 exports.LoopbackPort = LoopbackPort;
@@ -5415,26 +5411,6 @@
 "use strict";
 
 
-var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };
-
-var g;
-g = function () {
-  return this;
-}();
-try {
-  g = g || Function("return this")() || (1, eval)("this");
-} catch (e) {
-  if ((typeof window === "undefined" ? "undefined" : _typeof(window)) === "object") g = window;
-}
-module.exports = g;
-
-/***/ }),
-/* 7 */
-/***/ (function(module, exports, __w_pdfjs_require__) {
-
-"use strict";
-
-
 Object.defineProperty(exports, "__esModule", {
   value: true
 });
@@ -5509,7 +5485,7 @@
 exports.Metadata = Metadata;
 
 /***/ }),
-/* 8 */
+/* 7 */
 /***/ (function(module, exports, __w_pdfjs_require__) {
 
 "use strict";
@@ -5876,7 +5852,7 @@
 exports.WebGLUtils = WebGLUtils;
 
 /***/ }),
-/* 9 */
+/* 8 */
 /***/ (function(module, exports, __w_pdfjs_require__) {
 
 "use strict";
@@ -5895,7 +5871,7 @@
 
 var _annotation_layer = __w_pdfjs_require__(2);
 
-var _metadata = __w_pdfjs_require__(7);
+var _metadata = __w_pdfjs_require__(6);
 
 var _text_layer = __w_pdfjs_require__(5);
 
@@ -5907,8 +5883,8 @@
 }
 var PDFJS = _util.globalScope.PDFJS;
 {
-  PDFJS.version = '1.8.524';
-  PDFJS.build = '7b4887dd';
+  PDFJS.version = '1.8.525';
+  PDFJS.build = '0300fb01';
 }
 PDFJS.pdfBug = false;
 if (PDFJS.verbosity !== undefined) {
@@ -6015,7 +5991,7 @@
 exports.PDFJS = PDFJS;
 
 /***/ }),
-/* 10 */
+/* 9 */
 /***/ (function(module, exports, __w_pdfjs_require__) {
 
 "use strict";
@@ -9035,7 +9011,7 @@
 }]));
 
 /***/ }),
-/* 11 */
+/* 10 */
 /***/ (function(module, exports, __w_pdfjs_require__) {
 
 "use strict";
@@ -9048,9 +9024,9 @@
 
 var _util = __w_pdfjs_require__(0);
 
-var _pattern_helper = __w_pdfjs_require__(13);
+var _pattern_helper = __w_pdfjs_require__(12);
 
-var _webgl = __w_pdfjs_require__(8);
+var _webgl = __w_pdfjs_require__(7);
 
 var MIN_FONT_SIZE = 16;
 var MAX_FONT_SIZE = 100;
@@ -10771,7 +10747,7 @@
 exports.CanvasGraphics = CanvasGraphics;
 
 /***/ }),
-/* 12 */
+/* 11 */
 /***/ (function(module, exports, __w_pdfjs_require__) {
 
 "use strict";
@@ -11068,7 +11044,7 @@
 exports.FontLoader = FontLoader;
 
 /***/ }),
-/* 13 */
+/* 12 */
 /***/ (function(module, exports, __w_pdfjs_require__) {
 
 "use strict";
@@ -11081,7 +11057,7 @@
 
 var _util = __w_pdfjs_require__(0);
 
-var _webgl = __w_pdfjs_require__(8);
+var _webgl = __w_pdfjs_require__(7);
 
 var ShadingIRs = {};
 ShadingIRs.RadialAxial = {
@@ -11440,11 +11416,11 @@
 exports.TilingPattern = TilingPattern;
 
 /***/ }),
-/* 14 */
+/* 13 */
 /***/ (function(module, exports, __w_pdfjs_require__) {
 
 "use strict";
-/* WEBPACK VAR INJECTION */(function(global) {
+
 
 var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };
 
@@ -12941,19 +12917,18 @@
     globalScope.URL = JURL;
   })();
 }
-/* WEBPACK VAR INJECTION */}.call(exports, __w_pdfjs_require__(6)))
 
 /***/ }),
-/* 15 */
+/* 14 */
 /***/ (function(module, exports, __w_pdfjs_require__) {
 
 "use strict";
 
 
-var pdfjsVersion = '1.8.524';
-var pdfjsBuild = '7b4887dd';
+var pdfjsVersion = '1.8.525';
+var pdfjsBuild = '0300fb01';
 var pdfjsSharedUtil = __w_pdfjs_require__(0);
-var pdfjsDisplayGlobal = __w_pdfjs_require__(9);
+var pdfjsDisplayGlobal = __w_pdfjs_require__(8);
 var pdfjsDisplayAPI = __w_pdfjs_require__(3);
 var pdfjsDisplayTextLayer = __w_pdfjs_require__(5);
 var pdfjsDisplayAnnotationLayer = __w_pdfjs_require__(2);

```


yurydelendik (approved):
Thank you for the patch

PR #8634, opened 2017-07-10
Update inconsistent names

### LINKED ISSUE #7452: Inconsistent naming: Fit Page/Full Width vs Page Fit/Page Width
Review from the Chrome Web Store (but this bug applies to all builds):

> First of all, THANK YOU SO MUCH! This extension is freaking awesome! Anyway, I discovered a naming inconsistency in the extension. In the Options menu, you have the "Default zoom level" option. One of the options is "Full Width." But when you actually open a PDF, the zoom level that corresponds to that is called "Page Width." You may wish to consider calling both of these the same thing to prevent confusion. I would actually recommend changing both to say "Fit Width" instead, so it uses the same terminology as Adobe.

"Page Fit" and "Page Width" was renamed to "Fit Page" and "Full Width" in 3d435768385e567c908b201fd9d894b85b07c275, in viewer.html. However, the translations never changed. When I created `options.html`, I based the strings on viewer.html, hence the inconsistency spotted by a very sharp user.

There are two options here:
- Update l10n files to use "Fit Page" and "Full Width".
- Update web/viewer.html and `extensions/chromium/options/options.html` to use "Page Fit" and "Page Width".

nish17:
I'm sorry for last time, the text editor I'm using refactored the code and increased the indentation to 4 spaces. That's why it was showing too many changes.

Fixes #7452.

### DIFF
--- a/extensions/chromium/options/options.html
+++ b/extensions/chromium/options/options.html
@@ -50,8 +50,8 @@ body {
     <select>
       <option value="auto" selected="selected">Automatic Zoom</option>
       <option value="page-actual">Actual Size</option>
-      <option value="page-fit">Fit Page</option>
-      <option value="page-width">Full Width</option>
+      <option value="page-fit">Page Fit</option>
+      <option value="page-width">Page Width</option>
       <option value="custom" class="custom-zoom" hidden></option>
       <option value="50">50%</option>
       <option value="75">75%</option>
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good! Thank you for fixing this.

(Note that I moved the "Fixes #..." bit into the description of the pull request so that GitHub closes the issue automatically.)

PR #8636, opened 2017-07-11
Convert `PDFViewer` to an ES6 class

Snuffleupagus:
*Much more managable diff with https://github.com/mozilla/pdf.js/pull/8636/files?w=1.*

### DIFF
--- a/web/pdf_viewer.js
+++ b/web/pdf_viewer.js
@@ -26,14 +26,14 @@ import { PDFPageView } from './pdf_page_view';
 import { SimpleLinkService } from './pdf_link_service';
 import { TextLayerBuilder } from './text_layer_builder';
 
-var PresentationModeState = {
+const PresentationModeState = {
   UNKNOWN: 0,
   NORMAL: 1,
   CHANGING: 2,
   FULLSCREEN: 3,
 };
 
-var DEFAULT_CACHE_SIZE = 10;
+const DEFAULT_CACHE_SIZE = 10;
 
 /**
  * @typedef {Object} PDFViewerOptions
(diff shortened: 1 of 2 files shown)


web/pdf_viewer.js
```diff
@@ -135,833 +133,825 @@ var PDFViewer = (function pdfViewer() {
...
+      return;
+    }
 
-      for (let i = 0, ii = this._pages.length; i < ii; i++) {
-        let pageView = this._pages[i];
-        pageView.update(pageView.scale, rotation);
-      }
+    if (!(0 < val && val <= this.pagesCount)) {
+      console.error(
+        'PDFViewer._setCurrentPageNumber: "${val}" is out of bounds.');
```
timvandermeij: This whole string should be in backticks instead of regular quotes to actually evaluate the template literal `"${val}"`.

web/pdf_viewer.js
```diff
@@ -135,833 +133,825 @@ var PDFViewer = (function pdfViewer() {
...
+          // *unless* the page would thus become too wide to fit horizontally.
+          let horizontalScale = isLandscape ?
+            Math.min(pageHeightScale, pageWidthScale) : pageWidthScale;
+          scale = Math.min(MAX_AUTO_SCALE, horizontalScale);
           break;
         default:
-          console.error('PDFViewer_scrollPageIntoView: \'' + dest[1].name +
-                        '\' is not a valid destination type.');
+          console.error(
+            'PDFViewer._setScale: "${value}" is an unknown zoom value.');
```
timvandermeij: This whole string should be in backticks instead of regular quotes to actually evaluate the template literal "${value}".

timvandermeij (approved):
Nice work! Good to merge with the comments addressed.

web/pdf_viewer.js
Snuffleupagus: D'oh, this is embarrassing!
Thank you for spotting this; and as usual thanks for the review!

web/pdf_viewer.js
Snuffleupagus: D'oh, this is embarrassing!
Thank you for spotting this; and as usual thanks for the review!

PR #6984, opened 2016-02-13
Update `JpegImage.getData` to support `forceRGBoutput` for images with `numComponents === 1` (issue 6066)

Snuffleupagus:
_A more robust solution for issue #6066._

As a temporary work-around for (the upstream) [bug 1164199](https://bugzilla.mozilla.org/show_bug.cgi?id=1164199), we parsed _all_ images in the Firefox addon during a short time.
Doing so uncovered an issue with our image handling (see #6066), for JPEG images with a `DeviceGray` ColorSpace _and_ `bpc !== 1` (bits per component).

As long as we let the browser handle image decoding in this case, this isn't going to be an issue, but I do think that we should proactively fix this to avoid future issues if we change where the images are decoded (in `jpg.js` vs in browser).
Also, we currently don't seem to have a test-case for that kind of image data.

PR #8637, opened 2017-07-11
[DO NOT MERGE] Use `src/core/jpg.js` for all JPEG image decoding

Snuffleupagus:
This is only intended to compare the functionality/quality of `jpg.js` with the native browser decoding of JPEG images, to see if there's any obvious features missing from `jpg.js`.

/cc @brendandahl This is inspired by your comment in https://github.com/mozilla/pdf.js/pull/6984#issuecomment-184778220.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -358,6 +358,7 @@ var Driver = (function DriverClosure() { // eslint-disable-line no-unused-vars
           PDFJS.getDocument({
             url: absoluteUrl,
             password: task.password,
+            nativeImageDecoderSupport: 'none',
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
The failures seem to fall into three major categories:
 - One outright error, in test `issue4926`, where it from a cursory look appears that we fail to detect that colour conversion is necessary.
 - A fair number of cases where the images look blurry, compared to the native rendering, for example `bug946506` and `pdkids-page2`.
 - Finally a large amount of images where the differences are virtually imperceivable to the naked eye.

PR #8638, opened 2017-07-11
In `src/core/jpg.js`, ensure that the Adobe JPEG marker always takes precedence, even when the color transform code is zero

Snuffleupagus:
- Add support for the `nativeImageDecoderSupport` parameter, to force JPEG image decoding using `src/core/jpg.js`, when running the reference tests

 - In `src/core/jpg.js`, ensure that the Adobe JPEG marker always takes precedence, even when the color transform code is zero

   According to the PDF specification, please see http://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G6.2394361, if an Adobe JPEG marker is present it should always take precedence. This even seem to be consistent with the existing comment that is present in the code.
Hence it seems reasonable to interpret `transformCode === 0` as no color conversion being necessary.

   Fixes the rendering of page 1 in `issue-4926` (from the test-suite), when the built-in `src/core/jpg.js` image decoder is used.

### DIFF
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -958,12 +958,13 @@ var JpegImage = (function JpegImageClosure() {
       return data;
     },
 
-    _isColorConversionNeeded: function isColorConversionNeeded() {
-      if (this.adobe && this.adobe.transformCode) {
-        // The adobe transform marker overrides any previous setting
-        return true;
-      } else if (this.numComponents === 3) {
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #8640, opened 2017-07-12
Check for undefined

cmyksvoll:
`new URL(file, window.location.href)` throws the following error in IE11 + iPad Safari when `file` is undefined:
Unable to get property 'replace' of undefined or null reference


web/app.js
```diff
@@ -1404,6 +1404,9 @@ if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
     'http://mozilla.github.io', 'https://mozilla.github.io'];
   validateFileURL = function validateFileURL(file) {
     try {
+      if (file === undefined) {
```
brendandahl: Let's move this up out of the try/catch.

brendandahl (changes requested):

PR #8642, opened 2017-07-12
ensure a secure version of Node.js is used

Pomax:
this PR makes sure that a secure version of the LTS version of Node.js gets used (see https://nodejs.org/en/blog/vulnerability/july-2017-security-releases)

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "yargs": "^3.14.0"
   },
   "engines": {
-    "node": "^6.11.1",
+    "node": "^6.11.1"
   },
   "scripts": {
     "test": "gulp lint unittestcli externaltest"
--- a/package.json
+++ b/package.json
@@ -34,6 +34,9 @@
     "wintersmith": "^2.0.0",
     "yargs": "^3.14.0"
   },
+  "engines": {
+    "node": "^6.11.1",
+  },
   "scripts": {
     "test": "gulp lint unittestcli externaltest"
   },


Rob--W:
[`engines`](https://docs.npmjs.com/files/package.json#engines) is not for what version of Node is recommended, but what version of Node the library is compatible with.

Historically PDF.js was written to run in any modern browser (we specifically have compatibility.js to add missing functionality in some cases). To the best of my knowledge, we don't provide explicit support for a particular Node version (a very limited subset of the functionality is tested via Travis on the latest stable Node.js release). However, the examples are written using ES5 so that it works reasonably well if the user wants to experiment with PDF.js even if they don't have the latest and greatest version of Node.js

Pomax:
@Rob--W many CI and deployment tools will look at `engines` to determine what they need to install. Heroku, for instance, absolutely consults your package.json for an `engines` field to determine what it needs to install before it tries to run your code.

Rob--W:
It seems that you are conflating "minimum supported version" and "desired/secure version".
IMO, engines in package.json in a library specifies the minimum runtime that is known to be compatible, with the implication that a lower version is likely unsupported & incompatible. This is not the case here, so I don't want to add the `engines` field with an unnecessarily high value.

And since you mention Heroku specifically, here is how they use "engines':

> If you don’t specify a version, the latest Long-Term-Support release will be used.  
> -- https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

And although not documented in that article, what I found in practice is that if the `engines.node` field specifies a range/minimum, e.g. `>=4.0.0`, then heroku will install a recent version.

PR #8643, opened 2017-07-13
Add `Number.isNaN` and `Number.isInteger` polyfills in compatibility.js, since the Streams polyfill relies on them

Snuffleupagus:
Without this, the Streams polyfill will fail in Internet Explorer when the code-paths containing these methods are used.

*Please note:* This is something that I noticed when testing PR #8617 in IE (all versions). While, in the context of that PR, it would be sufficient to polyfill `Number.isNaN` I did a quick code search in the Streams polyfill and found `Number.isInteger` being used as well.

***Edit:*** The patch was updated to remove a spurious blank line, however no functional code changes were made.

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -836,6 +836,29 @@ PDFJS.compatibilityChecked = true;
   };
 })();
 
+// Provides support for Number.isNaN in legacy browsers.
+// Support: IE.
+(function checkNumberIsNaN() {
+  if (Number.isNaN) {
+    return;
+  }
+  Number.isNaN = function(value) {
+    return typeof value === 'number' && isNaN(value);
+  };
+})();
+
+// Provides support for Number.isInteger in legacy browsers.
+// Support: IE.
+(function checkNumberIsInteger() {
+  if (Number.isInteger) {
+    return;
+  }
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
*The patch was updated to remove a spurious blank line, no functional code changes were made. Hence the preview above should still be accurate.*

yurydelendik (approved):
Looks good. Can we also replace isInt by Number.isInteger and isNaN by Number.isNaN in our  src/ code?

Snuffleupagus:
> Looks good. Can we also replace isInt by Number.isInteger and isNaN by Number.isNaN in our src/ code?

That ought to be possible, especially the part about replacing our existing `isInt` helper function.
However, given that [`isNaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN) has slightly different behaviour than [`Number.isNaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) for non-numbers, this part will most likely require *very* careful analysis of each case to not accidentally alter the logic.

Since this is a blocker for PR #8617, I'm landing this as-is, and we can look into doing the conversion later on. Thank you for the review!

Issue #7041, opened 2016-02-29
Poorly rendered image quality


nicholassinggih:
Hi all,

I have tried using the latest 1.3.91 version for this. All I did was modify the viewer.js to point to my own PDF. This is the result using the fresh out of the box viewer.html :
![image](https://cloud.githubusercontent.com/assets/13591425/13384259/fd9dd7a2-dee7-11e5-9598-05e8baae48af.png)

The left one is the actual PDF, and the right one is rendered by viewer.html. Same thing in IE and Chrome. I did not try it in FF because my users use either IE or Chrome. 

I've tried searching everywhere I could and have been unsuccessful in finding a solution to this. Here's the PDF file from the picture. 
[.0002.pdf](https://github.com/mozilla/pdf.js/files/150290/default.0002.pdf)

Rob--W:
What exactly is wrong with the PDF from viewer.html?
(the zoom levels are different by the way, the one at the left uses 100% zoom, the one on the right 80%).

nicholassinggih:
Hi Rob,

Thanks for your reply. 

The problem is the text quality (I'm not sure if it's strictly font rendering or not) makes it a bit harder to read the words on it. Check out the word Billing at number 12, for example. The one on the right looks like it says "Bllllng". My users need to be able to see at least 8 pages at the same time, as they go through thousands every day. This means they're going to view the pdfs with an even smaller zoom level. 

The one on the left is not at 100%. It's actually at a smaller percentage than 80%. Just look at the sizes. The one on the left showed number 1-28 on the same screen, while the right one only showed 25.

Rob--W:
Could you paste a screenshot with exactly the same zoom sizes?

The "text rendering" is not a text rendering issue. What you see is an image. The question is whether the quality deteriotated after scaling (and of course, whether it can be improved).

nicholassinggih:
Let me clarify that I've tried playing with the scaling for both. I've tried looking at the PDF at 50% in Adobe Reader and Foxit. The quality doesn't get affected much by scaling. 

Here's from the viewer at 100%. Notice that the first 'i' in the word Billing still looks like an 'l'.
![image](https://cloud.githubusercontent.com/assets/13591425/13407566/16a688d0-df7d-11e5-968d-6af49437f70f.png)

And this next one is at 100% in Adobe Reader.
![image](https://cloud.githubusercontent.com/assets/13591425/13407640/9591e00e-df7d-11e5-9e82-3526f0f59083.png)

[...]

PR #8650, opened 2017-07-14
Performance optimizations

wojtekmaj:
Hey, that's my first Pull Request here, so *please do* hate me :)

I dug into the code a little, analyzing performance and trying to find potential bottlenecks.

Here's what I changed:
* **PageViewport**
    * I refactored function by reducing the number of calculations within the function. I simplified some calculations on basic arithmetic level: `((b + a) / 2) - a` => `(b- a) / 2`, `a * n + b * n - c * n` => `(a + b - c) * n` etc.
    * After code analysis, I've optimize the order of calculations (for example, offsetCanvasX was different for different rotation solely because width was switching with height when the viewport was rotated by 90 od 270 degrees) I've managed to reduce the code necessary to output what this function is supposed to.

    That resulted in **20-40% performance increase** in my tests. (~80 => 50 milliseconds per 10000 runs).
* **log2**
    * I replaced custom solution with Math functions.
    * I placed a warning that log2 function is not exactly doing what the name suggests. Didn't change the name because I didn't want to make a commotion.

    While I didn't measure how the change affected the performance in pdf.js, I did a [simple test on jsperf](https://jsperf.com/log2-pdfjs/1) which pointed native Math.log as a clear winner with **500% performance increase**.
    I've decided to use `Math.ceil(Math.log(x) * Math.LOG2E)` instead of `Math.ceil(Math.log2(x));` for compatibility reasons; virtually every browser supports `Math.log`, while `Math.log2` is only supported in modern browsers, while giving no performance bonus whatsoever.
* **CanvasGraphics_paintInlineImageXObject**
    * I rewrote the function painting image so that it would not draw a temporary image on a canvas each time it goes through a loop which purpose is to divide widthScale and heightScale by 2 as long as they are bigger than 2.

    I've measured the performance after implementing this change and the difference is big. In my test file, the function was called twice. Before the change, on average runs took 347 ms and 127 ms. After the change, they took 229 and 95 ms respectively. That's **32% performance increase**. Woo! 🎉

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -2107,12 +2107,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       var paintWidth = width, paintHeight = height;
       var tmpCanvasId = 'prescale';
       // Vertial or horizontal scaling shall not be more than 2 to not loose the
-      // pixels during drawImage operation, painting on the temporary canvas(es)
-      // that are twice smaller in size
-      while (
-        (widthScale > 2 && paintWidth > 1) ||
-        (heightScale > 2 && paintHeight > 1)
-      ) {
+      // pixels during drawImage operation
+      while ((widthScale > 2 && paintWidth > 1) ||
+             (heightScale > 2 && paintHeight > 1)) {
         var newWidth = paintWidth, newHeight = paintHeight;
         if (widthScale > 2 && paintWidth > 1) {
           newWidth = Math.ceil(paintWidth / 2);
@@ -2127,14 +2124,16 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
         paintHeight = newHeight;
       }
 
+      // painting on the temporary canvas
       tmpCanvas = this.cachedCanvases.getCanvas(tmpCanvasId,
                                                 newWidth, newHeight);
       tmpCtx = tmpCanvas.context;
       tmpCtx.clearRect(0, 0, newWidth, newHeight);
       tmpCtx.drawImage(imgToPaint, 0, 0, width, height,
-                                  0, 0, newWidth, newHeight);
+                                   0, 0, newWidth, newHeight);
       imgToPaint = tmpCanvas.canvas;
 
+      // final painting
       ctx.drawImage(imgToPaint, 0, 0, paintWidth, paintHeight,
                                 0, -height, width, height);
 
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -2105,12 +2105,14 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       }
 
       var paintWidth = width, paintHeight = height;
-      var tmpCanvasId = 'prescale1';
+      var tmpCanvasId = 'prescale';
       // Vertial or horizontal scaling shall not be more than 2 to not loose the
       // pixels during drawImage operation, painting on the temporary canvas(es)
       // that are twice smaller in size
-      while ((widthScale > 2 && paintWidth > 1) ||
-             (heightScale > 2 && paintHeight > 1)) {
+      while (
+        (widthScale > 2 && paintWidth > 1) ||
+        (heightScale > 2 && paintHeight > 1)
+      ) {
         var newWidth = paintWidth, newHeight = paintHeight;
         if (widthScale > 2 && paintWidth > 1) {
           newWidth = Math.ceil(paintWidth / 2);
@@ -2120,17 +2122,19 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
           newHeight = Math.ceil(paintHeight / 2);
           heightScale /= paintHeight / newHeight;
         }
-        tmpCanvas = this.cachedCanvases.getCanvas(tmpCanvasId,
-                                                  newWidth, newHeight);
-        tmpCtx = tmpCanvas.context;
-        tmpCtx.clearRect(0, 0, newWidth, newHeight);
-        tmpCtx.drawImage(imgToPaint, 0, 0, paintWidth, paintHeight,
-                                     0, 0, newWidth, newHeight);
-        imgToPaint = tmpCanvas.canvas;
+
         paintWidth = newWidth;
         paintHeight = newHeight;
-        tmpCanvasId = tmpCanvasId === 'prescale1' ? 'prescale2' : 'prescale1';
       }
+
+      tmpCanvas = this.cachedCanvases.getCanvas(tmpCanvasId,
+                                                newWidth, newHeight);
+      tmpCtx = tmpCanvas.context;
+      tmpCtx.clearRect(0, 0, newWidth, newHeight);
+      tmpCtx.drawImage(imgToPaint, 0, 0, width, height,
+                                  0, 0, newWidth, newHeight);
+      imgToPaint = tmpCanvas.canvas;
+
       ctx.drawImage(imgToPaint, 0, 0, paintWidth, paintHeight,
                                 0, -height, width, height);
 
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -2105,32 +2105,35 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       }
 
       var paintWidth = width, paintHeight = height;
+      var tmpWidth = paintWidth, tmpHeight = paintHeight;
       var tmpCanvasId = 'prescale';
       // Vertial or horizontal scaling shall not be more than 2 to not loose the
       // pixels during drawImage operation
       while ((widthScale > 2 && paintWidth > 1) ||
              (heightScale > 2 && paintHeight > 1)) {
-        var newWidth = paintWidth, newHeight = paintHeight;
+        tmpWidth = paintWidth;
+        tmpHeight = paintHeight;
+
         if (widthScale > 2 && paintWidth > 1) {
-          newWidth = Math.ceil(paintWidth / 2);
-          widthScale /= paintWidth / newWidth;
+          tmpWidth = Math.ceil(paintWidth / 2);
+          widthScale /= paintWidth / tmpWidth;
         }
         if (heightScale > 2 && paintHeight > 1) {
-          newHeight = Math.ceil(paintHeight / 2);
-          heightScale /= paintHeight / newHeight;
+          tmpHeight = Math.ceil(paintHeight / 2);
+          heightScale /= paintHeight / tmpHeight;
         }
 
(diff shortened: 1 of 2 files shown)


src/display/canvas.js
```diff
@@ -2105,12 +2105,14 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       }
 
       var paintWidth = width, paintHeight = height;
-      var tmpCanvasId = 'prescale1';
+      var tmpCanvasId = 'prescale';
```
timvandermeij: The comment below must be adjusted as well since we're not painting on temporary canvases anymore, but only on one in the end.

src/display/canvas.js
```diff
@@ -2105,12 +2105,14 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
       // Vertial or horizontal scaling shall not be more than 2 to not loose the
       // pixels during drawImage operation, painting on the temporary canvas(es)
       // that are twice smaller in size
-      while ((widthScale > 2 && paintWidth > 1) ||
-             (heightScale > 2 && paintHeight > 1)) {
+      while (
```
timvandermeij: Please revert this change as the existing style is the one that is used everywhere throughout the code base.

src/display/canvas.js
```diff
@@ -2120,17 +2122,19 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
         paintHeight = newHeight;
-        tmpCanvasId = tmpCanvasId === 'prescale1' ? 'prescale2' : 'prescale1';
       }
+
+      tmpCanvas = this.cachedCanvases.getCanvas(tmpCanvasId,
+                                                newWidth, newHeight);
+      tmpCtx = tmpCanvas.context;
+      tmpCtx.clearRect(0, 0, newWidth, newHeight);
+      tmpCtx.drawImage(imgToPaint, 0, 0, width, height,
+                                  0, 0, newWidth, newHeight);
```
timvandermeij: This line needs an extra space for indenting.

src/display/canvas.js
```diff
@@ -2120,17 +2122,19 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
-        tmpCtx.drawImage(imgToPaint, 0, 0, paintWidth, paintHeight,
-                                     0, 0, newWidth, newHeight);
-        imgToPaint = tmpCanvas.canvas;
+
         paintWidth = newWidth;
         paintHeight = newHeight;
-        tmpCanvasId = tmpCanvasId === 'prescale1' ? 'prescale2' : 'prescale1';
       }
+
+      tmpCanvas = this.cachedCanvases.getCanvas(tmpCanvasId,
```
timvandermeij: I think this change makes sense, but I can also recall that painting on the temporary canvases happened for a reason. @yurydelendik Could you perhaps comment on this, i.e., is there a specific reason why we're constantly painting on temporary canvases instead of only once in the end? Does it improve quality somehow?

src/shared/util.js
```diff
@@ -591,13 +591,9 @@ function string32(value) {
                              (value >> 8) & 0xff, value & 0xff);
 }
 
+// NOTE: This name is misleading as it rounds the result up
```
timvandermeij: Does this match the previous behavior? If not, why do we round up in the new version?

src/shared/util.js
```diff
@@ -907,58 +903,37 @@ var PageViewport = (function PageViewportClosure() {
...
+      rotation = rotation + 360;
     }
-
-    if (dontFlip) {
-      rotateC = -rotateC; rotateD = -rotateD;
-    }
-
-    var offsetCanvasX, offsetCanvasY;
+    var sideways = rotation % 180;
+    var a = rotation < 180 ? 1 : -1;
```
timvandermeij: Can we use a more descriptive name for this? I'm not sure what it represents.

timvandermeij:
The default PDF file (Tracemonkey) now gives the following in the console:

`Error: invalid canvas size`

The second page doesn't render completely anymore because of it. I assume that this is related to the `CanvasGraphics_paintInlineImageXObject` change. Could you check that?

src/display/canvas.js
wojtekmaj: That was my first thought. I checked that. There is a difference between the outputs of these two. It now produces sharper images. But I mean *sharper* sharper, and not *nearest neighbour crap* sharper. Here's a sample (5390px => 300px), left before and right after.

image

src/shared/util.js
wojtekmaj: I'm not sure either! It's just here to build rotate array (previously rotateA, rotateB, rotateC, rotateD). rotateParameter? transformParameter? transformFactor...?

wojtekmaj:
Hmmmm, yeah, I suspect the same. But I'm clicking furiously through it and I can't catch this error (but I caught several others. Do you know any more details? Does it happen on a specific page or with a specific scale?

timvandermeij:
It's triggered 100% for me when opening http://54.67.70.0:8877/6116cb2945c1db9/web/viewer.html and scrolling to page 2. It happens with any scale.

wojtekmaj:
Yeah, I'm on it. It happens when scale is already at 1 so it won't ever enter this loop. I'm gonna push fixes very soon.

src/display/canvas.js
fkaelberer: The reason for the iterative downsampling is that most browsers do a terrible job of downsampling by factors larger than 2 (I guess, they do use nearest neighbors). PDF documents often contain (scanned) text and drawings, rather than fotos, so the browser's downsampling does in fact severely reduce image quality. 
See https://github.com/mozilla/pdf.js/issues/7041#issuecomment-192496865 and below for some examples in which I compare iterative downsampling vs. browser downsampling.

src/shared/util.js
timvandermeij: I don't have an immediate suggestion here. Maybe @yurydelendik has an idea what this represents and how we can call it?

src/display/canvas.js
wojtekmaj: I understand your point, but having a look at the example above it is clear to me that there are cases in which we're losing both quality and performance. Maybe we should do some additional research on which browsers screw up which types of images, and leave the loop for them only?

src/shared/util.js
```diff
@@ -591,13 +591,9 @@ function string32(value) {
...
 
+// NOTE: This name is misleading as it rounds the result up
 function log2(x) {
-  var n = 1, i = 0;
-  while (x > n) {
-    n <<= 1;
-    i++;
-  }
-  return i;
+  return Math.ceil(Math.log(x) * Math.LOG2E);
```
fkaelberer: For inputs <= 0, the old version returned 0, whereas your version returns infinity.

fkaelberer: ... whereas the old version loops infinitely if `x > (1<<30)`

Snuffleupagus:
If I were to guess, I'd say that you probably didn't run any tests locally with these patches applied!?
Just doing a quick round of testing locally, it seems that this PR causes a number of rendering regressions in existing test-cases. Please refer to https://github.com/mozilla/pdf.js/wiki/Contributing, especially the parts about "Generating reference images" and "Run testing", for more information on the subject.

The title of the PR is "Performance optimizations", but the "Optimize performance of CanvasGraphics_paintInlineImageXObject function by not creating temporary images in while loop" commit isn't just an optimization, but actually *alters* the logic of the code. Most likely that this commit explains the regressions, and in my opinion it should be removed from the PR (https://github.com/mozilla/pdf.js/pull/8650#discussion_r127568469 explains perfectly well why the code looks like it currently does)!

Furthermore, I'm finding the "Optimize performance of PageViewport by reducing the number of calculations within the function" to make the code somewhat difficult to follow. I'll leave the actual review to @yurydelendik, but I cannot help wonder if needs to be *this* complicated?

src/shared/util.js
fkaelberer: My Proposal: 
```
    var scaleX = rotation < 180 ? scale : -scale;
    var scaleY = dontFlip ? scaleX : -scaleX;
    var rotate = sideways ? [0, scaleX, -scaleY, 0] : [scaleX, 0, 0, scaleY];

...

    this.transform = rotate.concat(
      offsetX + (width / 2 - rotate[0] * centerX - rotate[2] * centerY),
      offsetY + (height / 2 - rotate[1] * centerX - rotate[3] * centerY)
    );
```
which you get by the following refactoring steps:
1. multiply `a` by `scale`, and remove all ` * scale` in the `this.transform = ...` statement
2. multiply `flip` by `a`, remove `a *` in the `var rotate = ...` statement
3. rename `a` to `scaleX`
4. rename `flip` to `scaleY`
5. flip sign of `scaleY`
6. move `offsetX/Y +` to the beginning of the respective lines to have more visual alignement

fkaelberer: You could also write down the `transform` matrices explicitely, which would further reduce the number of arithmetic operations: 
```
...
    } else {
      width = scale * Math.abs(viewBox[2] - viewBox[0]);
      height = scale * Math.abs(viewBox[3] - viewBox[1]);
    }

    // creating transform for the following operations:
    // translate(-centerX, -centerY), rotate and flip vertically,
    // scale, and translate(offsetCanvasX, offsetCanvasY)
    var centerX = scaleX * (viewBox[2] + viewBox[0]) / 2;
    var centerY = scaleY * (viewBox[3] + viewBox[1]) / 2;

    if (sideways) {
        this.transform = [0, scaleX, -scaleY, 0, 
                          offsetX + centerY + width / 2, 
                          offsetY - centerX + height / 2];
    } else {
        this.transform = [scaleX, 0, 0, scaleY,
                          offsetX - centerX + width / 2,
                          offsetY - centerY + height / 2];
    }

    this.width = width;
    this.height = height;
...
```

src/shared/util.js
wojtekmaj: I tracked down all the paths log2() output might go and it seems like the only place when I'm not certain that input > 0 is where the result, saved as symbolCodeLength, is ultimately passed to decodeIAID as a third argument. There are only two places like this, so we could, just to be sure, change

`var symbolCodeLength = log2(inputSymbols.length)`

to

`var symbolCodeLength = inputSymbols.length && log2(inputSymbols.length)`

timvandermeij:
I'm going to close this pull request because `log2` is already fixed in #9256, the canvas graphics commit causes regressions and the page viewport changes are currently a bit hard to follow and needs more explanation as to why it works. If you want, you can revisit the page viewport changes in a new pull request, which will also make it easier to review. Thank you.

PR #8652, opened 2017-07-14
Optimize performance of CanvasGraphics_showText by merging glyphs

wojtekmaj:
This may sound weird, but intelligently merging glyphs before rendering them (add numbers to each other, merge glyph objects by adding fontChar, unicode and width properties) gives a **render performance boost** of about **26%**. On my test text document, before optimization I could definitely see the bottom of the page flashing in white for a split second. After optimization, rendering fits in 1-2 frames, reducing white flash to zero.

Here are the results from my simple test suite.

**Before:**

image

**After:**

image

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -1417,17 +1417,51 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       return shadow(this, 'isFontSubpixelAAEnabled', enabled);
     },
 
-    showText: function CanvasGraphics_showText(glyphs) {
+    showText: function CanvasGraphics_showText(unmergedGlyphs) {
       var current = this.current;
       var font = current.font;
       if (font.isType3Font) {
-        return this.showType3Text(glyphs);
+        return this.showType3Text(unmergedGlyphs);
       }
 
       var fontSize = current.fontSize;
       if (fontSize === 0) {
         return;
       }
+
+      // If glyphs placed next to each other are of the same type,
+      // we can merge them and reduce calls to fillText/drawChar.
+      var glyphs = unmergedGlyphs.reduce((glyphList, glyph, index) => {
+        if (index < 1) {
+          glyphList.push(glyph);
+          return glyphList;
+        }
+
+        var previousGlyph = glyphList[glyphList.length - 1];
+
+        if (isNum(previousGlyph) !== isNum(glyph) ||
+            glyph.accent ||
+            glyph.isInFont !== previousGlyph.isInFont ||
+            glyph.isSpace !== previousGlyph.isSpace) {
+          glyphList.push(glyph);
+          return glyphList;
+        }
+
+        if (isNum(glyph)) {
+          glyphList.splice(glyphList.length - 1, 1, previousGlyph + glyph);
+        } else {
+          glyphList.splice(glyphList.length - 1, 1, {
+            accent: previousGlyph.accent,
+            fontChar: previousGlyph.fontChar + glyph.fontChar,
+            isInFont: previousGlyph.isInFont,
+            isSpace: previousGlyph.isSpace,
+            unicode: previousGlyph.unicode + glyph.unicode,
+            width: previousGlyph.width + glyph.width,
+          });
+        }
+
+        return glyphList;
+      }, []);
 
       var ctx = this.ctx;
       var fontSizeScale = current.fontSizeScale;


src/display/canvas.js
```diff
@@ -1417,18 +1417,54 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
 
+      // If glyphs placed next to each other are of the same type,
+      // we can merge them and reduce calls to fillText/drawChar.
+      var glyphs = unmergedGlyphs.reduce((arr, glyph, index) => {
+        if (index < 1) {
+          arr.push(glyph);
+          return arr;
+        }
+
+        var prevGlyph = arr[arr.length - 1];
```
timvandermeij: Can we call this `previousGlyph` for readability?

src/display/canvas.js
```diff
@@ -1417,18 +1417,54 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
       }
 
       var fontSize = current.fontSize;
       if (fontSize === 0) {
         return;
       }
 
+      // If glyphs placed next to each other are of the same type,
+      // we can merge them and reduce calls to fillText/drawChar.
+      var glyphs = unmergedGlyphs.reduce((arr, glyph, index) => {
```
timvandermeij: I think `arr` is not really a descriptive name. Perhaps `glyphList` or something like that would be better?

src/display/canvas.js
```diff
@@ -1417,18 +1417,54 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
+        if (index < 1) {
+          arr.push(glyph);
+          return arr;
+        }
+
+        var prevGlyph = arr[arr.length - 1];
+
+        if (
```
timvandermeij: For consistency with existing code, please format this as follows:

```javascript
if (isNum(prevGlyph) !== isNum(glyph) ||
    glyph.accent ||
    glyph.isInFont !== prevGlyph.isInFont ||
    glyph.isSpace !== prevGlyph.isSpace) {
  arr.push(glyph);
  return arr;
}
```

wojtekmaj:
@timvandermeij Fixed everything you wrote, thank you!

src/display/canvas.js
```diff
@@ -1417,18 +1417,53 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
+          return glyphList;
+        }
+
+        var previousGlyph = glyphList[glyphList.length - 1];
+
+        if (isNum(previousGlyph) !== isNum(glyph) ||
+            glyph.accent ||
+            glyph.isInFont !== previousGlyph.isInFont ||
+            glyph.isSpace !== previousGlyph.isSpace
+        ) {
```
timvandermeij: Please move this line back to the previous line as well, so we get `glyph.isSpace !== previousGlyph.isSpace) {` on one line.

timvandermeij (commented):
Looks good to me, but since this touches core canvas logic, I'd like @yurydelendik to review this as well. Thanks!

wojtekmaj:
I'd love to hear his opinion. My main concern is that while in my case on every file I tested I could reduce the number of render calls by 30-40%, but if we stumble upon a horribly unoptimized PDF the performance won't be improved by attempting to merge the glyphs.

On the bright side, that opens us a possibility to optimize on a whole new level. In my case, glyphs[] looked kinda like this:

```js
[
  { fontChar: 'E', width: 500 ... },
  3,
  { fontChar: 'x', width: 460 ... },
  { fontChar: 'a', width: 476 ... },
  1,
  { fontChar: 'm', width: 460 ... },
  { fontChar: 'p', width: 476 ... },  
  { fontChar: 'l', width: 476 ... },
  4,
  { fontChar: 'e', width: 460 ... },
  ...
]
```

Now, this is how the text "Example" looked in my document. Actual render size of this word was about... 40 pixels? In other words, these numbers in between did nothig to the rendering.
What we could do is drop them completely and pass *entire words* to `fillText`. Or perhaps add those numbers to previous glyph width? I think it would not affect the rendering and we could end up with:

```js
[
  { fontChar: 'E', width: 503 ... },
  { fontChar: 'xa', width: 937 ... },
  { fontChar: 'mpl', width: 1416 ... },
  { fontChar: 'e', width: 460 ... },
  ...
]
```

and that would be reeeally something.

yurydelendik:
We tried to do it in the past -- personally I'm skeptical. We found that fonts might not contain proper width (our bugs and just corrupted PDF fonts), or different browser font (layout) engines not pull right glyph width or change distance between glyphs, and it's pretty important with PDF.

Let's see if problem still exist for windows and linux.

/botio test

brendandahl:
One thing to keep in mind when benchmarking is the browser will cache glyph runs, so if you're repeatedly re-rendering the same text it will definitely be faster to have chunks of text. However, if you're only rendering it once there's not really any performance boost. IIRC, I saw speed improvements on re-rendering doing something similar to this PR, but first render was unaffected.

wojtekmaj:
Uuuuh, this miserably failed firefox-issue4801-page1 reftest. Well, it's been fun and very educational trip!

PR #7635, opened 2016-09-15
Unify handling of various cursor tools, e.g. the current Hand Tool and a possible future Zoom Tool, in a new `PDFCursorTools` module

Snuffleupagus:
With the current way that the `HandTool` is implemented, if someone would try to also add a Zoom tool (as issue 1260 asks for) that probably wouldn't work very well given that you'd then have two cursor tools which may not play nice together.
Hence this patch, which attempts to refactor things so that it should be simpler to add e.g. a Zoom tool as well (given that that issue is marked as "good-beginner-bug", and I'm not sure if that really applies considering the current state of the code).

Note that I personally have no interest in implementing a Zoom tool (similar to Adobe Reader) since I wouldn't use it, but I figured that it can't hurt to make this code a bit more future proof.

PR #8653, opened 2017-07-15
Add UI for the cursorToolOnLoad pref in the Chrome extension + migration logic

Rob--W:
While preparing for a release, I encountered the "Don\'t know how to handle cursorToolOnLoad!" error in the console of the options page. After an investigation, I found that the hand tool preference was converted from a boolean to an integer in #7635.

I fixed the UI and added migration logic - see the commit message for more details.

Test:

1. Run `gulp chromium`
2. Go to `chrome://extensions` in Chrome, enable Developer mode and load the extension from `build/chromium/`.
3. Click on Inspect background page of the extension, and run the following code:

    ```javascript
    chrome.storage.local.set({enableHandToolOnLoad: true}, function() {location.reload();});
    ```
4. Run the following code to confirm that the preference has been migrated:

    ```javascript
    chrome.storage.local.get('enableHandToolOnLoad', console.log); // Should be an empty {}.
    ```
5. Click on the "Options" link and confirm that the "Cursor tool on load" dropdown is set to "Hand Tool".

Optional (to check that the migration logic does not interfere with a normal functioning of the migrated pref):

6. Change the dropdown to "Text selection tool", close the options page, reload the extension and open the options page.
7. Confirm that the dropdown value is still at "Text selection".
8. Repeat step 6 and 7, but change from "Text selection" to "Hand tool" instead.

### DIFF
--- a/extensions/chromium/options/migration.js
+++ b/extensions/chromium/options/migration.js
@@ -29,6 +29,8 @@ limitations under the License.
     storageLocal.get(storageKeys, function(values) {
       if (!values || !Object.keys(values).length) {
         // No local storage - nothing to migrate.
+        // ... except possibly for a renamed preference name.
+        migrateRenamedStorage();
         return;
       }
       migrateToSyncStorage(values);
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
I verified this and it works as intended. Thank you for fixing this!

Issue #8188, opened 2017-03-23
Missing references in internal links in pageviewer example


damiantp:
Hi,

I'm trying different components of pdfjs to build my own viewer. My idea is based on the "pageviewer" component example because I want to load only a page at a time. Using this example is working ok, but I don't get the internal links to other pages (although I only load one each time).

I'm using the following pdf (in page 3) as an example: http://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/pdf_open_parameters.pdf

I have added two statements to try to mimick what the simpleviewer example is doing with the linkservice, but it's not adding the page reference in the links. 
```
var pdfLinkService = new PDFJS.PDFLinkService(); // <--- ADDED
pdfLinkService.setDocument(pdfDocument, null); // <--- ADDED
```
(next is the modified version with these two lines)

My question is, is there a configuration that allows the pageviewer example generate links with page references?

**Modified versión of "pageviewer" example:**

```
'use strict';

if (!PDFJS.PDFViewer || !PDFJS.getDocument) {
  alert('Please build the pdfjs-dist library using\n' +
        '  `gulp dist`');
}

// The workerSrc property shall be specified.
//
PDFJS.workerSrc = '../../build/dist/build/pdf.worker.js';

var DEFAULT_URL = '../../web/pdf_open_parameters.pdf';
var PAGE_TO_VIEW = 3;
var SCALE = 1.0;

var container = document.getElementById('pageContainer');
var pdfLinkService = new PDFJS.PDFLinkService(); // <--- ADDED

// Loading document.
PDFJS.getDocument(DEFAULT_URL).then(function (pdfDocument) {

  pdfLinkService.setDocument(pdfDocument, null); // <--- ADDED

  // Document loaded, retrieving the page.
  return pdfDocument.getPage(PAGE_TO_VIEW).then(function (pdfPage) {
    // Creating the page view with default parameters.
    var pdfPageView = new PDFJS.PDFPageView({
      container: container,
      id: PAGE_TO_VIEW,
      scale: SCALE,
      defaultViewport: pdfPage.getViewport(SCALE),
      // We can enable text/annotations layers, if needed
      textLayerFactory: new PDFJS.DefaultTextLayerFactory(),
      annotationLayerFactory: new PDFJS.DefaultAnnotationLayerFactory()
    });
    // Associates the actual page with the view, and drawing it
    pdfPageView.setPdfPage(pdfPage);
    return pdfPageView.draw();
  });
});
[...]

PR #8664, opened 2017-07-17
Fixes Issue #8188 Missing references in internal links

nish17:


yurydelendik (changes requested):
It's not what #8188 is looking for -- we need component similar to PDFViewer (not example). However additional example will be needed to show how PDFSinglePageViewer works.

nish17:
I'm sorry, I didn't understand it fully. What exactly is the component and How do I implement the interface?

nish17:
I still don't understand that part of building UI for PDFSinglePageViewer How to do it? and one more thing If I put the code of simpleviewer.js file into some other file by keeping rest everything same, then why it's not working? Web browser is only able to load the UI and not the pdf File. Why is it happening?

nish17:
and I don't know why Travis build fails, I didn't change anything in the build folder.

yurydelendik:
The [comment](https://github.com/mozilla/pdf.js/pull/8664#pullrequestreview-50716736) above still stands. I'm expecting class created at web/ that has interface similar to what implemented in web/pdf_viewer.js. This class needs to be properly included in pdfjs-dist and example in examples/compoments/ provided.

timvandermeij:
Closing since a new PR is opened for this.

PR #8666, opened 2017-07-19
Fix TypeError that occurs in colorspace.js on accidentally passing an 'Array' instead of 'TypedArray'

apoorv-mishra:
Hi, 
**(Assumption: The base color space is set to be DeviceRGB)**
If an `Array`(not as `TypedArray` or `String`) is passed in `IndexedCS` constructor(https://github.com/mozilla/pdf.js/blob/master/src/core/colorspace.js#L509) as `lookup`, **TypeError** occurs on calling `fillRgb(...)`(on `IndexedCS` object), passing `lookup` as `comps`(https://github.com/mozilla/pdf.js/blob/master/src/core/colorspace.js#L116-L118). The error occurs because of `src.subarray(...)` in https://github.com/mozilla/pdf.js/blob/master/src/core/colorspace.js#L635 since `src` is expected to be a `TypedArray`. I made a small change to fix this.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -538,7 +538,7 @@ var IndexedCS = (function IndexedCSClosure() {
                                               dest, destOffset) {
       var numComps = this.base.numComps;
       var start = src[srcOffset] * numComps;
-      this.base.getRgbItem(this.lookup, start, dest, destOffset);
+      this.base.getRgbBuffer(this.lookup, start, 1, dest, destOffset, 8, 0);
     },
     getRgbBuffer: function IndexedCS_getRgbBuffer(src, srcOffset, count,
                                                   dest, destOffset, bits,


Snuffleupagus:
Thanks for the patch, but I've got a couple of questions/suggestions below.

Did you successfully run the reference tests locally, as explained in https://github.com/mozilla/pdf.js/wiki/Contributing, with this patch?
Looking at the IRC logs, in particular http://logs.glob.uno/?c=mozilla%23pdfjs&s=18+Jul+2017&e=18+Jul+2017#c64549, it mentions `TAMReview.pdf` as an example of Indexed ColorSpace usage. It appears that this patch unfortunately *breaks* at least that file, and thus quite possibly others as well.

From looking at the history of the code, see PR #3334, it appears that the *only* reason for supporting `Array` at all was for IE9 compatibility. Note that the `lookup` parameter, as coming from a PDF file, should *never* actually be a regular `Array`.
Wouldn't a better solution be to change [colorspace.js#L528-L529](https://github.com/mozilla/pdf.js/blob/4a74cc418ccb0b14dbbb50e0b964f32a41d16647/src/core/colorspace.js#L528-L529) to e.g. the following instead?
```javascript
} else if (lookup instanceof Uint8Array) {
  this.lookup = lookup;
} else if (lookup instanceof Array) {
  this.lookup = new Uint8Array(lookup);
```

Also, please update the commit message to something a little more descriptive, since it's really impossible to tell what the patch does based on just `TypeError fix in colorspace.js` without also reading the PR description. (In this case, even looking at the code itself doesn't really explain the reason for the change.)

apoorv-mishra:
Hi @Snuffleupagus , there is one more issue with making it as `Uint8Array`. Have a look at http://logs.glob.uno/?c=mozilla%23pdfjs&s=18+Jul+2017&e=18+Jul+2017#c64496.

yurydelendik (changes requested):
Looks like you need to change getRgbItem -> getRgbBuffer call in the IndexedCS_getRgbItem above.

src/core/colorspace.js
```diff
@@ -525,8 +525,10 @@ var IndexedCS = (function IndexedCSClosure() {
       for (var i = 0; i < length; ++i) {
         this.lookup[i] = lookup.charCodeAt(i);
       }
-    } else if (lookup instanceof Uint8Array || lookup instanceof Array) {
+    } else if (lookup instanceof Uint8Array) {
       this.lookup = lookup;
+    } else if (lookup instanceof Array) {
+      this.lookup = new Uint8Array(lookup);
```
yurydelendik: Can you revert the changes above? For IE9, it will not matter.

yurydelendik (approved):
Looks good (with just getRgbBuffer change)

src/core/colorspace.js
apoorv-mishra: Sure.

timvandermeij:
Merging with r=yurydelendik. Thank you for the patch!

Issue #8670, opened 2017-07-19
Text layer divs are not split upon changes in char/word spacing


audunhalland:
Link to PDF file (or attach file here):
[test.pdf](https://github.com/mozilla/pdf.js/files/1158969/test.pdf)

Configuration:
- Web browser and its version: Chrome 59
- Operating system and its version: macOS 10.2.5
- PDF.js version: pdfjs-dist 1.8.508
- Is an extension: ?

Steps to reproduce the problem:
1. Load the attached file "test.pdf" in pdf.js viewer
2. inspect the ".textLayer > div" rule in dev mode, and comment out "color: transparent"
3. text layer now shows as a transparent overlay, and reveals the bug.

Alternative way to reproduce, without messing with any css:
1. Load the attached file "test.pdf" in pdf.js viewer
2. Enter find mode, and type "oversikt" (Norwegian).
3. The highlighted section in the document extends beyond the size of that word. The whole div is stretched because of the extended character spacing of the following periods. The word and the dots should have been two separate divs.

What is the expected behavior? (add screenshot)
<img width="696" alt="screen shot 2017-07-19 at 13 55 03" src="https://user-images.githubusercontent.com/5767935/28365722-ed1b9c76-6c89-11e7-898c-de740bf90aeb.png">

What went wrong? (add screenshot)
<img width="691" alt="screen shot 2017-07-19 at 13 53 37" src="https://user-images.githubusercontent.com/5767935/28365684-c7885ddc-6c89-11e7-9106-1a0b0fe70b4e.png">

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

audunhalland:
Created pull request: https://github.com/mozilla/pdf.js/pull/8671

timvandermeij:
Closing since this is fixed now.

PR #8671, opened 2017-07-19
Split text content items upon word/char spacing operator.

audunhalland:
Fix for the following issue: https://github.com/mozilla/pdf.js/issues/8670

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1545,9 +1545,11 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 args[4], args[5]);
               break;
             case OPS.setCharSpacing:
+              flushTextContentItem();
               textState.charSpacing = args[0];
               break;
             case OPS.setWordSpacing:
+              flushTextContentItem();
               textState.wordSpacing = args[0];
               break;
             case OPS.beginText:


yurydelendik:
We need increase of DOM nodes analysis attached to this PR. It looks like this patch will increase amount of nodes in text layer.

audunhalland:
Yes, it definitely increases the amount of DOM nodes, there appears to be no way around that. Do you have a formal way to perform such an analysis against existing test PDFs?

yurydelendik:
> Do you have a formal way to perform such an analysis against existing test PDFs

There is no formal way. You can modify node's [getinfo](https://github.com/mozilla/pdf.js/blob/master/examples/node/getinfo.js) example to return amount of items returned by getTextContent call for some subset of test PDFs.

audunhalland:
For my test file (same as attached to #8670), [test.pdf](https://github.com/mozilla/pdf.js/files/1159560/test.pdf), I did a comparison of the number of children of the textLayer. Before: 134, after: 154.

If you want me to perform this comparison with other reference files, I'd be happy to do so.

yurydelendik:
> test.pdf... Before: 134, after: 154.

I consider this file to be the best case for PDF.js text layer. I would like to see number to bigger subset of PDF from different sources, e.g. from test/pdfs/. It will be nice to see mean and deviation change across bigger set.

> If you want me to perform this comparison with other reference files, I'd be happy to do so.

Yes, please

audunhalland:
I made a little script that scans all parseable files and pages in test/pdfs/*.pdf, using the output from page.streamTextContent({ normalizeWhitespace: true }), which I understand to be analogous to the number of divs in the text layer. Result:

Total divs all pages, before: 41036 after: 42912, which is an increase of 4.5%.

The mean of deviation is +5.6 divs per page where diff is not 0. Mean of deviation for _all_ pages is +2.28.

The largest increase ratio is very large and happens to be 1300%. This happens to be "issue3427.pdf" (page 2), increased from just 20 divs to 260. The document in question apparently includes spacing operators between every word and looks completely unnecessary by just looking at the rendered result. I have not studied this in detail, but perhaps my fix could include extra checks for avoiding splits where PDF spacing operators are just trivial and do not really change the state.

yurydelendik:
Additional list of most regressed pages by this patch:

- [ ] issue1419.pdf, page 4, 30=>2912 (9607%)
- [ ] artofwar.pdf, page 108, 43=>800 (1761%)
- [ ] issue3427.pdf, page 2, 20=>260 (1200%)
- [ ] geothermal.pdf, page 134, 171=>232 (36%)
- [ ] bug867484.pdf, page 12, 155=>196 (27%)
...

Notice that listed only unique document. IMHO regressions more than 10% is not acceptable in our case.

yurydelendik:
> The document in question apparently includes spacing operators between every word and looks completely unnecessary by just looking at the rendered result. 

We cannot control how documents are produced. This type of documents can be found on the web and we cannot regress PDF.js behavior since we cannot speculate what portion of documents will be affected.

As is this patch cannot be accepted. Thank you for the effort. I'll close this PR, but it can be reopened if "smarter" heuristic can be proposed.

PR #8673, opened 2017-07-19
[api-minor] Add support for PageMode in the API and viewer (issue 8657)

### LINKED ISSUE #8657: /PageMode /UseAttachments should be honored
[This uncompressed PDF document](https://github.com/mozilla/pdf.js/files/1151246/attachment-label.pdf) is automatically displayed by _Acrobat Reader_ (old version, the latest available for _Linux_) such as in:

attachment-open

The attachment pane is automatically opened by this code in _Acrobat_:

```
52 0 obj
<<
  /Type /Catalog
  /Pages 24 0 R
  /Metadata 25 0 R
  /Names 27 0 R
  /PageLabels <<
    /Nums [ 0 <<
          /S /D
          /St 1
        >> ]
  >>
  /PageMode /UseAttachments
  /Version /1.7
>>
endobj
```

Possible [values for `/PageMode`](https://wwwimages2.adobe.com/content/dam/Adobe/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=82) are:

* `/UseNone`
* `/UseOutlines`
* `/UseThumbs`
* `/FullScreen`
* `/UseOC`
* `/UseAttachments`

I’m not sure whether `/FullScreen` or `/UseOutlines` are implemented, but `/UseAttachments` is not. This page mode contains essential essential information (attachments that the user might not notice as part of the PDF document).

Many thanks for your help.

Snuffleupagus:
If we want to support this property, the following patches provides a way of doing so.

Fixes #8657.

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -102,6 +102,10 @@
       "type": "boolean",
       "default": false
     },
+    "disablePageMode": {
+      "type": "boolean",
+      "default": false
+    },
     "disableTelemetry": {
       "title": "Disable telemetry",
       "type": "boolean",
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -313,6 +313,24 @@ var Catalog = (function CatalogClosure() {
       return pageLabels;
     },
 
+    get pageMode() {
+      let obj = this.catDict.get('PageMode');
+      let pageMode = 'UseNone'; // Default value.
+
+      if (isName(obj)) {
+        switch (obj.name) {
+          case 'UseNone':
+          case 'UseOutlines':
+          case 'UseThumbs':
+          case 'FullScreen':
+          case 'UseOC':
+          case 'UseAttachments':
+            pageMode = obj.name;
+        }
+      }
+      return shadow(this, 'pageMode', pageMode);
+    },
+
     get attachments() {
       var xref = this.xref;
       var attachments = null, nameTreeRef;
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -770,6 +770,10 @@ var WorkerMessageHandler = {
       }
     );
 
+    handler.on('GetPageMode', function wphSetupGetPageMode(data) {
+      return pdfManager.ensureCatalog('pageMode');
+    });
+
     handler.on('GetAttachments',
       function wphSetupGetAttachments(data) {
         return pdfManager.ensureCatalog('attachments');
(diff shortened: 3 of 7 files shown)


timvandermeij (changes requested):
The code looks really good and this works for the basic API document, but for some reason it does not for the document provided in the issue. With e.g., the preview viewer, the sidebar does open, but not at the attachments view, but at the outline view instead. Any idea what causes that?

Snuffleupagus:
> [...] but for some reason it does not for the document provided in the issue. With e.g., the preview viewer, the sidebar does open, but not at the attachments view, but at the outline view instead. Any idea what causes that?

It seems that the PDF file in the issue doesn't actually have any attachments, since the `Root` -> `Names` -> `EmbeddedFiles` -> `Names` is empty. However, it does have a number of `/FileAttachment` *annotations*, which we for convenience (and consistency with e.g. Adobe Reader) append to the "Attachments" sidebar after the page containing them has been rendered.[1]

Since we disable empty sidebar views once the initial `getOutline`/`getAttachments` calls has resolved, see https://github.com/mozilla/pdf.js/blob/master/web/pdf_sidebar.js#L407-L434, this ought to explain the issue you're describing.
In general, there probably isn't a way to solve this, since that would require parsing the annotations of *every* single page before rendering the document. This is obviously *not* going to happen, since it would require fetching the entire PDF file before any rendering occurs, effectively breaking the ranged/chunked loading in the viewer.

Thus, I'd tend to consider this a pre-existing condition, but I agree that it's somewhat annoying/confusing!

What we could *maybe* try, is to delay the disabling of an empty Attachment view in the sidebar until e.g. the first page has rendered. Although considering that a `/FileAttachment` annotation could occur on *any* page of a document, that isn't going to help much in the general case.

@timvandermeij How do you think that we should proceed here?

---
[1] To me, the PDF file is abusing certain properties of the Adobe Reader implementation, since appending `/FileAttachment` annotations to the sidebar should not be a behaviour that a PDF file relies upon in various PDF viewers.

timvandermeij (approved):
I verified this with other PDF files (use `grep -r "UseOutline" *` and `grep -r "UseThumbs" *` in the `test/pdfs` folder to find them) and this patch works fine. The one in the original issue indeed relies on behavior that we cannot support, but at the very least the sidebar is now opened for that PDF file, giving the user a clue that there is content to be found there (and the green indicator on the attachments tab should help the user as well).

timvandermeij:
Really good, thanks!

PR #8678, opened 2017-07-20
Fix a typo, in the `Catalog.numPages` getter, than prevents shadowing from working correctly

Snuffleupagus:
Looking at the blame, it seems that this typo was present even before PR #700 (almost six years ago).
The result of using `'num'`, rather than the *correct* `'numPages'` string, is that the `Catalog.numPages` getter isn't actually being shadowed.


timvandermeij (approved):

timvandermeij:
Good find, thanks!

PR #8681, opened 2017-07-21
Fix several issues with glyph id mappings (issue 8668, bug 1383504)

### LINKED ISSUE #8668: [Regression] Missing glyph in the `issue8187` reference test
Unfortunately it appears that https://github.com/mozilla/pdf.js/blob/master/test/pdfs/issue8187.pdf regressed in commit https://github.com/mozilla/pdf.js/commit/eff257b8204c1816b8bfb660a3c5446bc06581b3 (PR #8580), which is the commit *just* after it was fixed by https://github.com/mozilla/pdf.js/commit/9f5c1550ede84bb4e489f5a11cafcf07ea55da1d (PR #8592).

This is what it looks like, with the current `master`:

This is what it *ought* to look like: 
1_ref

The obvious question becomes, why didn't the regression tests catch this in PR #8580?
Looking at the timestamps, it seems that the answer is that the *final* round of testing in PR https://github.com/mozilla/pdf.js/pull/8580#issuecomment-312725647 occurred *before* PR https://github.com/mozilla/pdf.js/pull/8592#event-1148823255 landed.
When I did the last round of reviewing, before makeref-ing and merging, I unfortunately didn't notice this discrepancy. Thus, the tests in PR #8580 didn't run after PR #8592 had landed, and the regression went unnoticed and the incorrect rendering is now what the bots consider "correct".

@brendandahl I've been unable to come up with a quick solution that doesn't break other test-cases, and I don't really have time to work on this right now, so I'm hoping that you can take a look!

brendandahl:
The initial issue with #8255 was I added a missing glyphs
check to adjustMapping, but this caused us to skip re-mapping
a glyph if the fontCharCode was a missingGlyph which in turn
caused us to overwrite a valid glyph id with an invalid one. While
fixing this, I also added a warning if the private use area is full since
this also accidentally happened when I made a different mistake.

This brought to light a number of issues where we map
missing glyphs to notdef, but often the notdef is actually defined
and then ends up being drawn. Now the glyphs don't get
mapped in toFontChar and so they are not drawn by the canvas.

Fixing the above brought up another issue though in bug1050040.pdf.
In this PDF, the font fails to load by the browser and before we were still
drawing the glyphs because it looked like the font had them, but with the fixes
above the glyphs showed up as missing so we didn't attempt draw them. To
fix this, I now throw an error when the loca table is in really bad shape and
we fall back to trying to use a system font. We now also use this fall back if
there are any format errors during converting fonts.

I expect this to fix:

firefox-issue1049-page2
firefox-issue1049-page3
firefox-issue8187-page1
firefox-bug1354114-page30

---
Fixes #8668.
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1383504.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -481,7 +481,7 @@ var ProblematicCharRanges = new Int32Array([
  */
 var Font = (function FontClosure() {
   function Font(name, file, properties) {
-    var charCode, glyphName, unicode;
+    var charCode;
 
     this.name = name;
     this.loadedName = properties.loadedName;
@@ -498,6 +498,7 @@ var Font = (function FontClosure() {
     var type = properties.type;
     var subtype = properties.subtype;
     this.type = type;
+    this.subtype = subtype;
 
     this.fallbackName = (this.isMonospace ? 'monospace' :
                          (this.isSerifFont ? 'serif' : 'sans-serif'));
@@ -512,6 +513,7 @@ var Font = (function FontClosure() {
     this.descent = properties.descent / PDF_GLYPH_SPACE_UNITS;
     this.fontMatrix = properties.fontMatrix;
     this.bbox = properties.bbox;
+    this.defaultEncoding = properties.defaultEncoding;
 
     this.toUnicode = properties.toUnicode;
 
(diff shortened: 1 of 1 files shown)


src/core/fonts.js
```diff
@@ -532,88 +534,14 @@ var Font = (function FontClosure() {
...
     }
-    var glyphsUnicodeMap;
+
     if (!file || file.isEmpty) {
       if (file) {
         // Some bad PDF generators will include empty font files,
         // attempting to recover by assuming that no file exists.
         warn('Font file is empty in "' + name + '" (' + this.loadedName + ')');
       }
-
```
brendandahl: All of this is extracted into fallbackToSystemFont().

Snuffleupagus:
Besides fixing issue #8668, this patch also fixes [bug 1383504](https://bugzilla.mozilla.org/show_bug.cgi?id=1383504).

brendandahl:
For some reason linux doesn't seem to like it when we don't have cmap entries for all the type1 fonts. 
/botio test

brendandahl:
Hmm....well now the following fixes don't work, but the original regression, issue8187, is at least fixed.

firefox-issue1049-page2
firefox-issue1049-page3
firefox-bug1354114-page30

Snuffleupagus:
> Hmm....well now the following fixes don't work, but the original regression, issue8187, is at least fixed.
> 
> firefox-issue1049-page2
> firefox-issue1049-page3
> firefox-bug1354114-page30

Do you want to try and fix the above files in this PR, or would a follow-up PR be better?
It would be really nice to get this landed, since it fixes issue #8668 and [bug 1383504](https://bugzilla.mozilla.org/show_bug.cgi?id=1383504) which are regressions!

brendandahl:
@Snuffleupagus let's land as is and I'll do follow ups to re-fix the above

Snuffleupagus (approved):
Seems fine to me, thank you!

(I'm updating the title to include the issue/bug that's fixed by this PR, since that will allow RyanVM to mark the Bugzilla dependencies when doing the next update.)

PR #8684, opened 2017-07-21
Remove most `assert()` calls (issue 8506) 

### LINKED ISSUE #8506: Remove error() and re-inspect assert() calls.
There is lots of confusion how error() is used. It also may mislead people when reading source code -- error() hides `throw new Error()`, so it's not easy to see control flow for human and JS analyzers cannot complain about dead code.

The same issue can be found with assert(). Normally assert() is used to define precondition for the following code, means that code shall never have its check `false`. (In other languages it's usually defined as hard stop for the program). By hiding checks and throw inside assert(), we also potentially removing JIT ability to make some assumptions.

- [x] Introduce InvalidPDFError() to raise when PDF is malformed instead of generic Error()
- [x] We need to review code and change `error(_)` => `throw new Error(_)`
- [ ] We need to review code and change `assert(_,?)` => `if (!_) throw new Error(_)` for non-assert-like calls (test for it is: it's impossible to remove this check/assert without changing logic of the program and affect the testing)

Snuffleupagus:
This replaces `assert` calls with `throw new FormatError()`/`throw new Error()`.
In a few places, throwing an `Error` (which is what `assert` meant) isn't correct since the enclosing function is supposed to return a `Promise`, hence some cases were changed to `Promise.reject(...)` and similarily for `createPromiseCapability` instances.

Fixes #8506.

@yurydelendik Based on https://github.com/mozilla/pdf.js/issues/8506#issuecomment-307455596 in particular: I wasn't really sure if we should leave some, and if so *which* ones, of the current `assert` calls in place.
~~This patch is currently replacing *all* of the them, but I figured that might not be what you actually want here which is why I've marked the PR as `work-in-progress` and would very much appreciate feedback on this!~~

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -673,7 +673,9 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
       var fontRef, xref = this.xref;
       if (font) { // Loading by ref.
-        assert(isRef(font));
+        if (!isRef(font)) {
+          throw new Error('The "font" object should be a reference.');
+        }
         fontRef = font;
       } else { // Loading by name.
         var fontRes = resources.get('Font');
@@ -866,7 +868,9 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       resources = resources || Dict.empty;
       initialState = initialState || new EvalState();
 
-      assert(operatorList, 'getOperatorList: missing "operatorList" parameter');
+      if (!operatorList) {
+        throw new Error('getOperatorList: missing "operatorList" parameter');
+      }
 
       var self = this;
       var xref = this.xref;
@@ -928,10 +932,14 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
               var xobj = xobjs.get(name);
               if (xobj) {
-                assert(isStream(xobj), 'XObject should be a stream');
+                if (!isStream(xobj)) {
+                  throw new FormatError('XObject should be a stream');
+                }
 
                 var type = xobj.dict.get('Subtype');
-                assert(isName(type), 'XObject should have a Name subtype');
+                if (!isName(type)) {
+                  throw new FormatError('XObject should have a Name subtype');
+                }
 
                 if (type.name === 'Form') {
                   stateManager.save();
@@ -1087,10 +1095,14 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
             case OPS.shadingFill:
               var shadingRes = resources.get('Shading');
-              assert(shadingRes, 'No shading resource found');
+              if (!shadingRes) {
+                throw new FormatError('No shading resource found');
+              }
 
               var shading = shadingRes.get(args[0].name);
-              assert(shading, 'No shading object found');
+              if (!shading) {
+                throw new FormatError('No shading object found');
+              }
 
               var shadingFill = Pattern.parseShading(shading, null, xref,
                 resources, self.handler);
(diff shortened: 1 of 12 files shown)


src/core/evaluator.js
```diff
@@ -765,7 +767,9 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
         }
         this.fontCache.put('id_' + fontID, fontCapability.promise);
       }
-      assert(fontID, 'The "fontID" must be defined.');
+      if (!fontID) {
```
yurydelendik: looks like fontID generated by internal logic. if we can tell that out logic produces non-empty id, we can keep assert.

src/core/image.js
```diff
@@ -525,7 +523,9 @@ var PDFImage = (function PDFImageClosure() {
           }
           if (this.needsDecode) {
             // Invert the buffer (which must be grayscale if we reached here).
-            assert(kind === ImageKind.GRAYSCALE_1BPP);
```
yurydelendik: looks like a good candidate for assert, if needsDecode == true is driven only by internal logic (and not use data)

src/core/network.js
```diff
@@ -291,7 +291,9 @@ PDFNetworkStream.prototype = {
   },
 
   getFullReader: function PDFNetworkStream_getFullReader() {
-    assert(!this._fullRequestReader);
```
yurydelendik: internal logic, let's keep it

src/core/parser.js
```diff
@@ -157,7 +157,9 @@ var Parser = (function ParserClosure() {
         } else if (state === 1) {
           state = (ch === I) ? 2 : 0;
         } else {
-          assert(state === 2);
```
yurydelendik: internal logic, let's keep it

src/core/pattern.js
```diff
@@ -403,7 +405,9 @@ Shadings.Mesh = (function MeshClosure() {
 
   function buildFigureFromPatch(mesh, index) {
     var figure = mesh.figures[index];
-    assert(figure.type === 'patch', 'Unexpected patch mesh figure');
```
yurydelendik: mostly internal logic, I think we can keep it

src/core/worker.js
```diff
@@ -232,7 +232,9 @@ var PDFWorkerStream = (function PDFWorkerStreamClosure() {
            rangeReader._enqueue(args.chunk);
            return true;
          });
-         assert(found);
```
yurydelendik: internal logic, let's keep it

src/core/worker.js
```diff
@@ -254,7 +256,9 @@ var PDFWorkerStream = (function PDFWorkerStreamClosure() {
     },
 
     getFullReader: function PDFWorkerStream_getFullReader() {
-      assert(!this._fullRequestReader);
```
yurydelendik: internal logic, let's keep it

src/display/font_loader.js
```diff
@@ -155,7 +155,9 @@ if (typeof PDFJSDev === 'undefined' || !PDFJSDev.test('MOZCENTRAL')) {
   FontLoader.prototype.queueLoadingCallback =
       function FontLoader_queueLoadingCallback(callback) {
     function LoadLoader_completeRequest() {
-      assert(!request.end, 'completeRequest() cannot be called twice');
```
yurydelendik: internal logic, let's keep it

src/shared/util.js
```diff
@@ -1634,7 +1635,6 @@ export {
   FormatError,
   arrayByteLength,
   arraysToBytes,
-  assert,
```
yurydelendik: The asserts in this file looks like internal logic, let's keep them... and export

yurydelendik (commented):
Here is my suggestions for acceptable asserts.

Snuffleupagus:
@yurydelendik Thanks a lot for your feedback, I've revert all of the `assert`s that you suggested.

/botio test

timvandermeij (approved):

timvandermeij:
Nice work, thanks!

PR #8688, opened 2017-07-23
Parse  content disposition  header

### LINKED ISSUE #4284: pdf.js relies on urls to contain the 'pdf' extension
When the server doesn't provide the Content-Disposition header, pdf.js relies on urls to contain the 'pdf' extension. But URLs are locators, not names.
Steps to reproduce:

```
mv web/compressed.tracemonkey-pldi-09.pdf web/compressed.tracemonkey-pldi-09
sed -i 's/compressed.tracemonkey-pldi-09.pdf/compressed.tracemonkey-pldi-09/g' web/viewer.js
firefox web/viewer.html
```

Now click on download. You will be offered a 'document.pdf' file. The name should be something more meaningful.
The bug also happens when I leave out the pdf extension on my apache web server.

Proposed solution:
Use the title of the pdf. (as this [viewer.js](https://github.com/mozilla/pdf.js/blob/master/web/viewer.js#L1043) code). The title is also used by firefox for the File -> "save page as" functionality when displaying a HTML page like http://en.wikipedia.org/wiki/Internet_media_type .

Not every html web page ends in .html. Instead by the extension, a document's type is specified by its MIME-type.
However, most pdf files have the pdf extension, and most pdfs online also have a good-to-store name in the url.
I don't know whether the new retrieval method should overwrite the url retrieval, or be a fallback to it.

See also #3455.

### LINKED ISSUE #6396: Content-Disposition filename ignored somehow
Hello, currently I render my Browser documents with PDFJS however I have patched the viewer.js to support my Authorization header i just added httpHeaders to paramters before PDFJS.getdocument so nothing special, however PDFJS ignore's my Content-Disposition Header which has a filename=
it always gives me document.pdf

I tested against a CORS version and non cors currently this is my content-disposition:

```
 Content-Disposition:inline; filename="Angebot_412356.pdf"
```

Is there a setting that it should use this filename or do I miss something?

Currently it isn't related to https://github.com/mozilla/pdf.js/issues/2407 since that will work on save + s.
However on my side either Firefox/Chrome nobody gets the correct content-disposition i tried:

```
 Content-Disposition: inline; filename=name.pdf
 Content-Disposition: attachment; filename=name.pdf
 Content-Disposition: inline; filename="name.pdf"
 Content-Disposition: attachment; filename="name.pdf"
```

Also that's the Code I added in viewer.js starting after Line 6373 and the PDFJS line is already there:

```
 var token = window.localStorage.getItem('token');
 var typ = window.localStorage.getItem('type');
 parameters.httpHeaders = {'Authorization': typ + ' ' + token};
 PDFJS.getDocument(parameters, pdfDataRangeTransport, passwordNeeded,
```

btw. I could also set the filename by myself, I think of the possibility that I know the filename already.

jsperezg:
Good afternoon.

This is my small contribution to this project. My  patch uses the file name received in the content disposition  header instead of infer it from the URL. 

In case no content disposition is received it will work as usual.

Fixes #4284.
Fixes #6396.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2001,8 +2001,10 @@ var WorkerTransport = (function WorkerTransportClosure() {
         return {
           info: results[0],
           metadata: (results[1] ? new Metadata(results[1]) : null),
+          contentDispositionFileName: (this._fullReader ?
+                                       this._fullReader.fileName : null),
         };
-      });
+      }.bind(this));
     },
 
     getStats: function WorkerTransport_getStats() {
--- a/src/display/network.js
+++ b/src/display/network.js
@@ -15,7 +15,8 @@
 
 import { assert, createPromiseCapability, stringToBytes } from '../shared/util';
 import {
-  createResponseStatusError, validateRangeRequestCapabilities
+  createResponseStatusError, extractFilenameFromHeader,
+  validateRangeRequestCapabilities
 } from './network_utils';
 import globalScope from '../shared/global_scope';
 
@@ -340,6 +341,7 @@ function PDFNetworkStreamFullRequestReader(manager, source) {
   this._requests = [];
   this._done = false;
   this._storedError = undefined;
+  this._fileName = null;
 
   this.onProgress = null;
 }
@@ -350,11 +352,13 @@ PDFNetworkStreamFullRequestReader.prototype = {
     var fullRequestXhrId = this._fullRequestId;
     var fullRequestXhr = this._manager.getRequestXhr(fullRequestXhrId);
 
+    const getResponseHeader = (name) => {
+      return fullRequestXhr.getResponseHeader(name);
+    };
+
     let { allowRangeRequests, suggestedLength, } =
       validateRangeRequestCapabilities({
-        getResponseHeader: (name) => {
-          return fullRequestXhr.getResponseHeader(name);
-        },
+        getResponseHeader,
         isHttp: this._manager.isHttp,
         rangeChunkSize: this._rangeChunkSize,
         disableRange: this._disableRange,
@@ -379,6 +383,11 @@ PDFNetworkStreamFullRequestReader.prototype = {
       // server should not be returning that it can support range
       // requests.
       networkManager.abortRequest(fullRequestXhrId);
+    }
+
+    // Content-Disposition: attachment; filename=Naïve file.txt
+    if (networkManager.isPendingRequest(fullRequestXhrId)) {
+      this._fileName = extractFilenameFromHeader(getResponseHeader);
     }
 
     this._headersReceivedCapability.resolve();
@@ -427,6 +436,10 @@ PDFNetworkStreamFullRequestReader.prototype = {
         total: data.lengthComputable ? data.total : this._contentLength,
       });
     }
+  },
+
+  get fileName() {
+    return this._fileName;
   },
 
   get isRangeSupported() {
--- a/web/app.js
+++ b/web/app.js
@@ -153,6 +153,7 @@ let PDFViewerApplication = {
   baseUrl: '',
   externalServices: DefaultExternalServices,
   _boundEvents: {},
+  contentDispositionFileName: null,
 
   // Called once when the document is loaded.
   initialize(appConfig) {
@@ -675,6 +676,7 @@ let PDFViewerApplication = {
     this.store = null;
     this.isInitialViewSet = false;
     this.downloadComplete = false;
+    this.contentDispositionFileName = null;
 
     this.pdfSidebar.reset();
     this.pdfOutlineViewer.reset();
@@ -792,7 +794,8 @@ let PDFViewerApplication = {
     let url = this.baseUrl;
     // Use this.url instead of this.baseUrl to perform filename detection based
     // on the reference fragment as ultimate fallback if needed.
-    let filename = getPDFFileNameFromURL(this.url);
+    let filename = this.contentDispositionFileName ||
+      getPDFFileNameFromURL(this.url);
     let downloadManager = this.downloadManager;
     downloadManager.onerror = (err) => {
       // This error won't really be helpful because it's likely the
@@ -1144,9 +1147,11 @@ let PDFViewerApplication = {
       });
     });
 
-    pdfDocument.getMetadata().then(({ info, metadata, }) => {
+    pdfDocument.getMetadata().then(
+        ({ info, metadata, contentDispositionFileName, }) => {
       this.documentInfo = info;
       this.metadata = metadata;
+      this.contentDispositionFileName = contentDispositionFileName;
 
       // Provides some basic debug information
       console.log('PDF ' + pdfDocument.fingerprint + ' [' +
@@ -1170,6 +1175,10 @@ let PDFViewerApplication = {
 
       if (pdfTitle) {
         this.setTitle(pdfTitle + ' - ' + document.title);
+      }
+
+      if (!pdfTitle && contentDispositionFileName) {
+        this.setTitle(contentDispositionFileName);
       }
 
       if (info.IsAcroFormPresent) {
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "uglify-es": "^3.1.2",
     "vinyl": "^2.1.0",
     "vinyl-fs": "^2.4.4",
-    "webpack": "^3.6.0",
+    "webpack": "^3.10.0",
     "webpack-stream": "^4.0.0",
     "wintersmith": "^2.4.1",
     "yargs": "^9.0.1"
--- a/src/display/fetch_stream.js
+++ b/src/display/fetch_stream.js
@@ -17,8 +17,8 @@ import {
   AbortException, assert, createPromiseCapability
 } from '../shared/util';
 import {
-  createResponseStatusError, validateRangeRequestCapabilities,
-  validateResponseStatus
+  createResponseStatusError, extractFilenameFromHeader,
+  validateRangeRequestCapabilities, validateResponseStatus
 } from './network_utils';
 
 function createFetchOptions(headers, withCredentials) {
@@ -67,6 +67,7 @@ class PDFFetchStream {
 class PDFFetchStreamReader {
   constructor(stream) {
     this._stream = stream;
+    this._fileName = null;
     this._reader = null;
     this._loaded = 0;
     let source = stream.source;
@@ -100,11 +101,13 @@ class PDFFetchStreamReader {
       this._reader = response.body.getReader();
       this._headersCapability.resolve();
 
+      const getResponseHeader = (name) => {
+        return response.headers.get(name);
+      };
+
       let { allowRangeRequests, suggestedLength, } =
         validateRangeRequestCapabilities({
-          getResponseHeader: (name) => {
-            return response.headers.get(name);
-          },
+          getResponseHeader,
           isHttp: this._stream.isHttp,
           rangeChunkSize: this._rangeChunkSize,
           disableRange: this._disableRange,
@@ -112,6 +115,7 @@ class PDFFetchStreamReader {
 
       this._contentLength = suggestedLength;
       this._isRangeSupported = allowRangeRequests;
+      this._fileName = extractFilenameFromHeader(getResponseHeader);
 
       // We need to stop reading when range is supported and streaming is
       // disabled.
@@ -129,6 +133,10 @@ class PDFFetchStreamReader {
 
   get contentLength() {
     return this._contentLength;
+  }
+
+  get fileName() {
+    return this._fileName;
   }
 
   get isRangeSupported() {
--- a/src/display/network_utils.js
+++ b/src/display/network_utils.js
@@ -16,6 +16,7 @@
 import {
   assert, MissingPDFException, UnexpectedResponseException
 } from '../shared/util';
+import { getFilenameFromUrl } from './dom_utils';
 
 function validateRangeRequestCapabilities({ getResponseHeader, isHttp,
                                             rangeChunkSize, disableRange, }) {
@@ -65,8 +66,23 @@ function validateResponseStatus(status) {
   return status === 200 || status === 206;
 }
 
+function extractFilenameFromHeader(getResponseHeader) {
+  const contentDisposition = getResponseHeader('Content-Disposition');
+
+  if (contentDisposition) {
+    let parts =
+      /.+;\s*filename=(?:"|')(.+\.pdf)(?:"|')/gi.exec(contentDisposition);
+    if (parts !== null && parts.length > 1) {
+      return getFilenameFromUrl(parts[1]);
+    }
+  }
+
+  return null;
+}
+
 export {
   createResponseStatusError,
   validateRangeRequestCapabilities,
   validateResponseStatus,
+  extractFilenameFromHeader,
 };
--- a/src/display/node_stream.js
+++ b/src/display/node_stream.js
@@ -22,7 +22,9 @@ let url = __non_webpack_require__('url');
 import {
   AbortException, assert, createPromiseCapability
 } from '../shared/util';
-import { validateRangeRequestCapabilities } from './network_utils';
+import {
+  extractFilenameFromHeader, validateRangeRequestCapabilities
+} from './network_utils';
 
 class PDFNodeStream {
   constructor(source) {
@@ -72,6 +74,7 @@ class BaseFullReader {
     this._done = false;
     this._errored = false;
     this._reason = null;
+    this._fileName = null;
     this.onProgress = null;
     let source = stream.source;
     this._contentLength = source.length; // optional
@@ -105,6 +108,10 @@ class BaseFullReader {
 
   get isStreamingSupported() {
     return this._isStreamingSupported;
+  }
+
+  get fileName() {
+    return this._fileName;
   }
 
   read() {
@@ -282,13 +289,15 @@ class PDFNodeStreamFullReader extends BaseFullReader {
       this._headersCapability.resolve();
       this._setReadableStream(response);
 
+      const getResponseHeader = (name) => {
+        // Make sure that headers name are in lower case, as mentioned
+        // here: https://nodejs.org/api/http.html#http_message_headers.
+        return this._readableStream.headers[name.toLowerCase()];
+      };
+
       let { allowRangeRequests, suggestedLength, } =
       validateRangeRequestCapabilities({
-        getResponseHeader: (name) => {
-          // Make sure that headers name are in lower case, as mentioned
-          // here: https://nodejs.org/api/http.html#http_message_headers.
-          return this._readableStream.headers[name.toLowerCase()];
-        },
+        getResponseHeader,
         isHttp: stream.isHttp,
         rangeChunkSize: this._rangeChunkSize,
         disableRange: this._disableRange,
@@ -299,6 +308,9 @@ class PDFNodeStreamFullReader extends BaseFullReader {
       }
       // Setting right content length.
       this._contentLength = suggestedLength;
+
+      // Setting the file name from the response header
+      this._fileName = extractFilenameFromHeader(getResponseHeader);
     };
 
     this._request = null;
--- a/test/unit/network_utils_spec.js
+++ b/test/unit/network_utils_spec.js
@@ -14,8 +14,8 @@
  */
 
 import {
-  createResponseStatusError, validateRangeRequestCapabilities,
-  validateResponseStatus
+  createResponseStatusError, extractFilenameFromHeader,
+  validateRangeRequestCapabilities, validateResponseStatus
 } from '../../src/display/network_utils';
 import {
   MissingPDFException, UnexpectedResponseException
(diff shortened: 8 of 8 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@ Makefile
 node_modules/
 examples/node/svgdump/
 examples/node/pdf2png/*.png
+.idea
```
timvandermeij: Revert this change (and therefore this commit). We don't keep editor-specific lines in the `.gitignore` file; it is the responsibility of the developer to setup these directories elsewhere or exclude them when committing. Moreover, we try to keep the changes in a patch limited to only the intended functionality to make review easier.

src/core/network.js
```diff
@@ -414,6 +415,17 @@ PDFNetworkStreamFullRequestReader.prototype = {
       networkManager.abortRequest(fullRequestXhrId);
     }
 
+    // Content-Disposition: attachment; filename=Naïve file.txt
+    var fullRequestXhr = networkManager.getRequestXhr(fullRequestXhrId);
+    var contentDisposition = fullRequestXhr
```
timvandermeij: In the other parts of the codebase, we format this line as follows:

```javascript
var contentDisposition =
  fullRequestXhr.getResponseHeader('Content-Disposition');
```

Please do that here as well.

timvandermeij (commented):
This patch already looks good! I have added some small review comments, but someone with more in-depth knowledge about the network code should do a full review.

Please address these comments and after that squash the commits into one. Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits if you're unfamiliar with that. A rebase would also remove the merge commit that you currently have.

*(I updated the description to include the issues that this PR fixes).*

src/core/network.js
```diff
@@ -414,6 +415,17 @@ PDFNetworkStreamFullRequestReader.prototype = {
...
+                              .getResponseHeader('Content-Disposition');
+    if (contentDisposition) {
+      var parts = /.+; filename="(.+)"/gi.exec(contentDisposition);
```
timvandermeij: This patch also needs some unit tests in https://github.com/mozilla/pdf.js/blob/master/test/unit/network_spec.js. Since this line is the most essential part of the patch, I would suggest to perhaps extract these four lines into a separate function and test that in the unit test.

web/app.js
```diff
@@ -560,7 +560,10 @@ let PDFViewerApplication = {
   setTitleUsingUrl(url) {
     this.url = url;
     this.baseUrl = url.split('#')[0];
-    let title = getPDFFileNameFromURL(url, '');
+    let title = this.attachedFileName;
+    if (title === null) {
+      title = getPDFFileNameFromURL(url, '');
+    }
```
Snuffleupagus: Why not simply `let title = this.attachedFileName || getPDFFileNameFromURL(url, '');` instead?
That way, you'd also avoid issues with the title being empty if the file name found in the content disposition header is e.g. an empty string.

web/app.js
```diff
@@ -728,7 +736,10 @@ let PDFViewerApplication = {
     let url = this.baseUrl;
     // Use this.url instead of this.baseUrl to perform filename detection based
     // on the reference fragment as ultimate fallback if needed.
-    let filename = getPDFFileNameFromURL(this.url);
+    let filename = this.attachedFileName;
+    if (filename === null) {
+      filename = getPDFFileNameFromURL(this.url);
+    }
```
Snuffleupagus: Similar to https://github.com/mozilla/pdf.js/pull/8688/files#r128925411, this can be `let filename = this.attachedFileName || getPDFFileNameFromURL(this.url);` instead.

src/core/network.js
jsperezg: Completelly agree!

I'll do.

jsperezg:
Thank you all for your comments. I'm working on it in my spare time. I hope to invest some time on it next weekend.

jsperezg:
Good  morning all.

Following your valuable feedback I've added tests for the function  that extracts the file name from the Content Disposition header. I've also fixed the code styling issues that you mentioned.

Finally I've left out of the .gitignore the IDE files and squashed the changes.

Please, tell me if there is anything more that I can do.

PD: With two kids I've few time to contribute so It's my first contribution to a big project. At this point I must say that it's being a pleasure. You as community are very kind and easy going. Thank you very much for this pleasant experience.

src/core/network.js
```diff
@@ -414,9 +415,27 @@ PDFNetworkStreamFullRequestReader.prototype = {
...
+      fullRequestXhr.getResponseHeader('Content-Disposition');
+    this._fileName = this._extractNameFrom(contentDisposition);
+
     this._headersReceivedCapability.resolve();
   },
 
+  _extractNameFrom:
+      function PDFNetworkStreamFullRequestReader_extractNameFrom(cDisposition) {
+    if (cDisposition) {
+      var parts = /.+; filename="(.+)"/gi.exec(cDisposition);
```
Snuffleupagus: *I've not really reviewed the entire patch, but one thing that I can't help but wondering here is:* Why don't you also ensure that the filename actually ends with `.pdf`?

If the file format is wrong, then using the filename found in the content disposition header doesn't really make a lot of sense (especially since various places within the viewer depends on it).

jsperezg: This is a good point.

The patch is pretty simple. It parses the content disposition header in network layer and sends up to the viewer UI the result. The UI uses it to compose the title and the download file name.

Adding some kind of validation that prevents the viewer from loading invalid files and making it fail in a facy way would be certainly an improvement.

I can do it, but obviously it will take some extra  time since I'm not familiar with the procedures that you follow to raise exceptions  from the network layer to the UI. 

At this point I would suggest to split the work in two. First finish integrating this pull request in the main branch with the current functionality. It will solve a couple of issues and add a feature long awaited. 

Then we can create an issue to add the feature that you suggest. I'm not member of the official project, and I don't want to write permissions to the main repo. I'm a newbie that will need some time to have a general vision of all the project, but, if it's feasible, you can asssign the issue to me and I will work on it in a new branch after this one is acceped and merged.

In case you consider this pull request should not be merged without this feature it would be nice if you could give me some advice about how to handle it. A file name with a line number would be enough: "Hi Juan, look at xxxx.js line yyy. There we do something similar to what you need to achieve".

Kind regards.

Snuffleupagus: There's a reason that I added a comment on *this* particular line :-)
You ought to be able to simply extend the regular expression to only match on filenames ending with `.pdf`; so wouldn't using a regexp like `/^Content-Disposition\:.*\;?\bfilename=(?:'|")([^'"]*\.pdf)(?:'|")/i` get the job done here (while also doing some basic validation)?

stemwinder: I'm not a contributor, just patiently waiting for this pull.

I think the point @jsperezg is making is that if validating the file name falls within the scope of his work, then it should also raise an exception so the user is not confused. I could imagine a developer beating their head against a wall for a good many hours because a validation event on user input is failing silently.

I don't really think validating the file name is within scope, or really necessary at all, but like I said I'm not a contributor - my opinion doesn't matter in that regard.

Thanks for the good work @jsperezg!

jsperezg: That's exactly what I was assuming @stemwinder . 

@Snuffleupagus Adding this validation in the method that extracts the file name from the headers in practice means that it will be ignored (it will return nil since it don't matches the input string). If you try to load something that is not a pdf it will fail elsewhere because of the file content.

If we want to add  a validation  on the file name we should go one step forward and stop the loading process if the received file name do not seems to be a valid pdf.

Anyway, if all this discussion is about adding a '\.pdf' in the regexp and writting a couple of tests, I'll do. It's not a big challenge at all.

Snuffleupagus: I'm afraid that you might be missing the point here, since as-is this patch would allow a completely bogus filename to be used in various places the viewer; one example being when downloading a file.

Note how in [web/app.js](https://github.com/mozilla/pdf.js/blob/master/web/app.js), we're using the [`getPDFFileNameFromURL`](https://github.com/mozilla/pdf.js/blob/master/web/ui_utils.js#L394) function to obtain a `*.pdf` filename (with a fallback to `document.pdf`).
So I definitely think that some form of validation of a Content-Disposition filename is necessary, if not on the API side, then at the *very* least not using it unconditionally in the viewer (like the patch does)!
Without validation on the API side, then [web/app.js#L731](https://github.com/mozilla/pdf.js/blob/master/web/app.js#L731) would probably need to look something like this:
```javascript
let filename;
if (this.attachedFileName) {
  filename = getPDFFileNameFromURL(this.attachedFileName, '');
}
filename = filename || getPDFFileNameFromURL(this.url);
```

jsperezg: Good morning @Snuffleupagus 

I've already added it but I need some assistance from somebody with more experience than me:

I've been also working in the branch conflicts. I've discovered that I made some newbie mistakes during the branch process: Basically I didn't executed this step:

git remote add upstream git://github.com/mozilla/pdf.js.git

I tried to do it on the branch for this pull request and it resulted in a small mess.

Then I decided to start again, following the steps in the contribution  guide and adding all the changes to a new branch.

Now I have everything working, in just one commit.

What I don't know how to do is replace the invalid branch of this pull request  with the right one. How can I do it? Or should I create a new pull request?

PD: I'm not pushed the changes yet.

yurydelendik (changes requested):
IMHO _extractNameFrom and tests are looking fine.

Notice that network.js was moved and you may not need worker.js changes anymore. So rebasing will not work for you -- I recommend by exporting your changes into some file, reset branch (via `git checkout -B parse_attachment_file_name upstream/master`) and force pushing less changes.

I would like to remove onAttachedFileName and return this filename in the transport's getMetadata -- you shall have direct access to stream there.

jsperezg:
Good afternoon @yurydelendik .

I've fixed the problems I had with the previous branch. I just need to work a bit on:

"I would like to remove onAttachedFileName and return this filename in the transport's getMetadata -- you shall have direct access to stream there.".

I'll take 7 days of holidays starting tomorrow so I hope to continue with it next week.

jsperezg:
@yurydelendik 

Hi  all. I'm back from my holidays. As I promised now the file name is returned in transport's getMetadata.

Best regards.

timvandermeij:
Thank you for working on this again! I think you forgot to push the updated patch since GitHub does not show that anything changed. If you rebased, you may need to force-push the commit.

timvandermeij:
@jsperezg Are you still working on this? It looks like some changes still need to be made. If not, contributor @himanish-star showed interest in finishing it if that's okay. Could you let us know if you have time to finish it or not?

jsperezg:
@timvandermeij I did the changes requested by @yurydelendik, but it was some months ago. I don't know why  it has not been reviewed since then.

Anyway IMHO this PR is close to be ready to be merged so I have no problem in having @himanish-star involed in order to push it forward.

Please, let me know if there is something that I can do.

timvandermeij:
@jsperezg I completely missed the updated commit. Last time the new commit did not show up on GitHub for some reason, which explains https://github.com/mozilla/pdf.js/pull/8688#issuecomment-321665752, but I did not notice that it appeared in the meantime. Sorry, my bad!

You indeed made the changes and I think the patch looks quite good. I see one thing that needs to be done and that's to remove the merge commit. Now you have two commits in this PR while it should only be one. To fix that, I think Yury's comment at https://github.com/mozilla/pdf.js/pull/8688#pullrequestreview-53215716 explains it well: it may be the easiest to apply your changes on a clean branch that is up-to-date with the current master and force-push that to this branch. If you need assistance with that, you can always contact us on IRC or ask here. Once that is done, we can do a final review and hopefully merge it then. Thank you!

src/display/api.js
```diff
@@ -1494,6 +1494,7 @@ var PDFWorker = (function PDFWorkerClosure() {
         var isTypedArraysPresent = Uint8Array !== Float32Array;
         var port = new LoopbackPort(isTypedArraysPresent);
         this._port = port;
+        this._webWorker = port;
```
timvandermeij: It looks like this is unused and can be removed.

src/display/api.js
```diff
@@ -2053,11 +2054,13 @@ var WorkerTransport = (function WorkerTransportClosure() {
     },
 
     getMetadata: function WorkerTransport_getMetadata() {
+      let fullReader = this._fullReader;
       return this.messageHandler.sendWithPromise('GetMetadata', null).
         then(function transportMetadata(results) {
         return {
           info: results[0],
           metadata: (results[1] ? new Metadata(results[1]) : null),
+          attachedFileName: fullReader.fileName,
```
timvandermeij: I'm a bit confused about the name `attachedFileName` because it has nothing to do with attachments in PDF files, yet that's what I think about first when I read it. Would `headerFileName` or `contentDispositionFileName` perhaps be better? If so, do not forget to change this elsewhere in the patch too.

src/display/network.js
```diff
@@ -414,9 +415,27 @@ PDFNetworkStreamFullRequestReader.prototype = {
       networkManager.abortRequest(fullRequestXhrId);
     }
 
+    // Content-Disposition: attachment; filename=Naïve file.txt
+    var fullRequestXhr = networkManager.getRequestXhr(fullRequestXhrId);
```
timvandermeij: Change `var` to `let` here and at the other occurrences in this patch.

src/display/network.js
```diff
@@ -414,9 +415,27 @@ PDFNetworkStreamFullRequestReader.prototype = {
...
+    var contentDisposition =
+      fullRequestXhr.getResponseHeader('Content-Disposition');
+    this._fileName = this._extractNameFrom(contentDisposition);
+
     this._headersReceivedCapability.resolve();
   },
 
+  _extractNameFrom:
```
timvandermeij: Let's change this line and the next line to simply `_extractNameFrom(contentDisposition)`. We use anonymous functions a lot and it makes it easier to read (also writing `contentDisposition` instead of `cDisposition` is more readable).

I think it may also be more clear to rename this to `_extractNameFromHeader` to be more explicit about where it's extracted from.

test/unit/network_spec.js
```diff
@@ -176,4 +176,62 @@ describe('network', function() {
       done.fail(reason);
     });
   });
+
+  describe('PDFNetworkStreamFullRequestReader_extractNameFrom', function () {
```
timvandermeij: `describe('header filename extraction', ...` is a bit shorter and easier to read here.

test/unit/network_spec.js
```diff
@@ -176,4 +176,62 @@ describe('network', function() {
...
+        url: pdf1,
+        rangeChunkSize: 65536,
+        disableStream: true,
+      },
+      disableRange: true,
+    });
+
+    var requestReader = stream.getFullReader();
+
+    it('returns null when no content disposition', function() {
```
timvandermeij: `when no content disposition header is present`

test/unit/network_spec.js
```diff
@@ -176,4 +176,62 @@ describe('network', function() {
...
+      expect(requestReader._extractNameFrom(null)).toBeNull();
+      expect(requestReader._extractNameFrom(undefined)).toBeNull();
+    });
+
+    it('Response header for the main body', function () {
```
timvandermeij: `gets the filename from the response header from the main body`. The same applies to the descriptions below, i.e., try to make the descriptions flow better with the code for readability.

test/unit/network_spec.js
```diff
@@ -176,4 +176,62 @@ describe('network', function() {
...
+      )).toBe('filename.pdf');
+    });
+
+    it('Header of multipart body', function () {
+      expect(requestReader._extractNameFrom(
+        'Content-Disposition: form-data'
+      )).toBeNull();
+
+      expect(requestReader._extractNameFrom(
+        'Content-Disposition: form-data; name="fieldName"'
```
timvandermeij: Could you add a test case where `name="filename"`, just to make sure that the regex does not pick that up as a valid filename?

jsperezg:
Good morning @timvandermeij 

I will fix the code according to your comments and rebase the PR squashing the changes.

Best regards.

test/unit/network_spec.js
jsperezg: Good point. I'll do

test/unit/network_spec.js
jsperezg: consider it done

jsperezg:
Good morning @yurydelendik 

I've changed the source according to your comments and fixed all issues that I've found with linter and tests, please let me know if there is necessary to make any other adjustments.

Finally, there is something that I want to share with you and the rest of the pdf.js team:

I've had problems running the tests. I work as consultant. Due to this fact my node version is screwed to certain version that seemed to be the cause of these problems. Recently I started using docker-compose/docker in order to have an isolated development environment for each project.

In order to be able to run the tests locally I've built a Dockerfile and a docker-compose.yml files that install all requirements. Do you thing it would be interesting to add them to pdf.js project in a separate PR? If so I would be pleased to share them with the rest of the community.

timvandermeij:
I think we should leave out the Docker and `package.json` changes since they're unrelated to this pull request. Moreover, it looks like indentation changes have been made that make the diff considerably harder to read and they should not be necessary. Please revert those to make the change as small as possible.

src/display/network.js
```diff
@@ -414,9 +415,31 @@ PDFNetworkStreamFullRequestReader.prototype = {
       networkManager.abortRequest(fullRequestXhrId);
     }
 
+    // Content-Disposition: attachment; filename=Naïve file.txt
+    if (networkManager.isPendingRequest(fullRequestXhrId)) {
+      let fullRequestXhr = networkManager.getRequestXhr(fullRequestXhrId);
+      let contentDisposition = fullRequestXhr
```
timvandermeij: Format this like:

```javascript
let contentDisposition =
  fullRequestXhr.getResponseHeader('Content-Disposition');
```

src/display/network.js
```diff
@@ -414,9 +415,31 @@ PDFNetworkStreamFullRequestReader.prototype = {
...
+                               .getResponseHeader('Content-Disposition');
+      this._fileName = this._extractNameFromHeader(contentDisposition);
+    }
+
     this._headersReceivedCapability.resolve();
   },
 
+  _extractNameFromHeader: function
```
timvandermeij: Just make this `_extractNameFromHeader(contentDisposition) {` (one line). No need for the long name since we use anonymous functions more and more now.

jsperezg:
@timvandermeij I'll do

test/unit/network_spec.js
```diff
@@ -176,4 +176,63 @@ describe('network', function() {
       done.fail(reason);
     });
   });
+
+  describe('header filename extraction', function () {
+    var stream = new PDFNetworkStream({
```
timvandermeij: `let stream`

test/unit/network_spec.js
```diff
@@ -176,4 +176,63 @@ describe('network', function() {
...
+      source: {
+        url: pdf1,
+        rangeChunkSize: 65536,
+        disableStream: true,
+      },
+      disableRange: true,
+    });
+
+    var requestReader = stream.getFullReader();
```
timvandermeij: `let requestReader`

timvandermeij:
This patch has really good unit tests. Functionally this is looking really good. The comment above are more to adhere to the coding style that we use throughout the project and to keep the patch small. The former is for consistency, the latter for ease of reviewing and reducing the amount of code that we need to touch in this patch (both for reviewing and for avoiding regressions).

jsperezg:
@timvandermeij Since it's my first contribution to this project your comments regarding to coding style are welcome. I will fix them as soon as I have some time: usually during weekends.

Kind regards.

test/unit/network_spec.js
```diff
@@ -176,4 +176,63 @@ describe('network', function() {
       done.fail(reason);
     });
   });
+
+  describe('header filename extraction', function () {
```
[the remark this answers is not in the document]
Snuffleupagus: Furthermore, on the subject of tests, let's update the API unit-tests with
```javascript
expect(metadata.contentDispositionfileName ).toEqual(null);
```
placed just before https://github.com/mozilla/pdf.js/blob/5a52ee0a795ca41f27ede587dc494cd304dc81e1/test/unit/api_spec.js#L797

web/app.js
```diff
@@ -617,6 +618,7 @@ let PDFViewerApplication = {
     this.findBar.reset();
     this.toolbar.reset();
     this.secondaryToolbar.reset();
+    this.contentDispositionFileName = null;
```
Snuffleupagus: To maintain better grouping between functionality here, let's move this line to just after https://github.com/mozilla/pdf.js/pull/8688/files#diff-529d1853ee1bba753a0fcb40ea778723R611 instead.

web/app.js
```diff
@@ -1070,69 +1073,80 @@ let PDFViewerApplication = {
...
-                  (!PDFJS.disableWebGL ? ' [WebGL]' : '') + ')');
-
-      let pdfTitle;
-      if (metadata && metadata.has('dc:title')) {
-        let title = metadata.get('dc:title');
-        // Ghostscript sometimes return 'Untitled', sets the title to 'Untitled'
-        if (title !== 'Untitled') {
-          pdfTitle = title;
+    pdfDocument.getMetadata().then(
+      ({ info, metadata, contentDispositionFileName, }) => {
```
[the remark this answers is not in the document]
Snuffleupagus: Furthermore, I'd guess that we also want to use this filename (if it exists) when displaying the document properties dialog as well; please see https://github.com/mozilla/pdf.js/blob/5a52ee0a795ca41f27ede587dc494cd304dc81e1/web/pdf_document_properties.js#L74-L84

Snuffleupagus (commented):
As explained in comments below, the network code is now spread across a couple of different files (to support `fetch` and Node.js).

I've left a few comments below that hopefully helps with bringing this patch more inline with the current state of the networking code.

*Please note:* Make sure that you rebase this patch against the *latest* master *before* attempting to address these comments (since it won't work otherwise).

src/display/network.js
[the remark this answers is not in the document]
jsperezg: I've rabased bu I can't find the network_utils.js that you mention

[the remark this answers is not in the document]
Snuffleupagus: > I've rabased bu I can't find the network_utils.js that you mention

This is probably a stupid question, but are you sure that you fetched the latest upstream code *before* rebasing?
Assuming that you're following the workflow outlined in https://github.com/mozilla/pdf.js/wiki/Contributing, you'd then run the following git commands:
```
git checkout parse_attachment_file_name
git fetch upstream
git rebase upstream/master
```
After that, you should find be able to find the `src/display/network_utils.js` file.

Snuffleupagus:
> All in all my impression is that this is going to nowhere.

At this point, this is actually quite close to landing (from a functionality standpoint)!
However, very shortly after this PR was originally opened, the `fetch_stream.js` and `node_stream.js` implementations landed. Hence there's unfortunately a need to update this patch such that the interfaces of the different network `fullReader` implementations match.

> I do it in my spare time. 

It's not really a good excuse, but most of us (myself included) work on PDF.js in our spare time as well, hence things sometimes fall through the cracks; sorry about that!

> but new features, support for new API, and so forth.

I wouldn't call any of the final round of review comments new features/API, it's just that we need to take into account the `fetch_stream.js` and `node_stream.js` files as well.

> I would really appreciate if you could review my last patch and we coud prepare a kind of roadmap of what must, be done, what is desirable to be done. Any help with the changes would be welcome as well.

The latest round of review comments is the final one, really!
Once we've gotten consistent interfaces between the `fullReader` implementations in [network.js](https://github.com/mozilla/pdf.js/blob/master/src/display/network.js), [fetch_stream.js](https://github.com/mozilla/pdf.js/blob/master/src/display/fetch_stream.js), and [node_stream.js](https://github.com/mozilla/pdf.js/blob/master/src/display/fetch_stream.js) this should be good to land.

I did try to provide some details in the inline comments, but I'm not sure if they're clear enough (if you're not already familiar code). What sort of additional details would you find helpful here?

src/display/api.js
```diff
@@ -2053,11 +2053,19 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
+        let contentDispositionfileName = null;
+
+        if (fullReader !== null) {
+          contentDispositionfileName = fullReader.fileName;
+        }
+
         return {
           info: results[0],
           metadata: (results[1] ? new Metadata(results[1]) : null),
+          contentDispositionFileName: contentDispositionfileName,
```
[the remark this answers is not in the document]
jsperezg: I had to bind the function context as well, Otherwise the tests fail (after rebasing the code)

src/display/network.js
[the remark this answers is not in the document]
jsperezg: what changes are supposed to go in fetch_stream.js and node_stream.js files. Do you mean that the patch needs to be reimplemented in these two files?

jsperezg:
I've just commited my last contribution guys. My last post exposed my concerns, not excuses. Simply becasue It's not my responsibility and I do it for pleasure I don't need them. Merge it or dump it to the thrash bin. Meanwhile I'm working with a branched version of the viewer that solves my necessities.

timvandermeij:
Merged in #9379. Thank you for implementing this!

PR #8690, opened 2017-07-23
adds Interface. reference pull request #8664  

nish17:
Is this what was expected? if this is correct then I'll add it into examples/components directory. Just let me know whether its correct or not? Adding interface does that mean adding .css file, .component.js file etc ?


nish17:
@timvandermeij Wil you please review this and if possible, Will you please explain this issue(https://github.com/mozilla/pdf.js/issues/8188) a bit more in detail?

timvandermeij:
Closing since a new PR is opened for this.

PR #8691, opened 2017-07-23
Implement support for square and circle annotations

timvandermeij:
The commit descriptions provide more information about the individual changes.


timvandermeij:
Does somebody have time to review this patch? It will make it easier to implement more annotation types and it is the first step towards supporting annotations without appearance streams (since the invisible annotations that we draw over the canvas can also be made visible and styled in case of missing appearance streams more easily with this approach).

timvandermeij:
I have updated the patches to resolve conflicts and to convert the code to ES6 syntax after the recent introduction of ES6 syntax for the annotation layer.

src/display/annotation_layer.js
```diff
@@ -798,7 +796,105 @@ class LineAnnotationElement extends AnnotationElement {
...
+    let data = this.data;
+    let width = data.rect[2] - data.rect[0];
+    let height = data.rect[3] - data.rect[1];
+    let svg = DOMSVGFactory.create(width, height);
+
+    // The browser draws half of the borders inside the square and half of
+    // the borders outside the square by default. This behavior cannot be
+    // changed programmatically, so correct for that here.
+    let borderWidth = data.borderStyle.width;
+    let square = document.createElementNS(SVG_NS, 'svg:rect');
```
yurydelendik: Add `createElement` to the DOMSVGFactory.

src/display/annotation_layer.js
```diff
@@ -798,7 +796,105 @@ class LineAnnotationElement extends AnnotationElement {
...
+    square.setAttributeNS(null, 'x', borderWidth / 2);
```
yurydelendik: I wonder if just setAttribute('x', boderWidth / 2); will work?

src/display/dom_utils.js
```diff
@@ -109,6 +110,23 @@ class DOMCMapReaderFactory {
   }
 }
 
+class DOMSVGFactory {
+  static create(width, height) {
+    if (width <= 0 || height <= 0) {
```
yurydelendik: If we will keep DOMSVGFactory internal to the library, `assert()` is preferred.

yurydelendik (approved):
Looks good to me.

I'm thinking DOMSVGFactory must be created Abstract factory. This means that it will not have static member and be an object that we can pass around. It will have a nice effect: we can replace node polyfill with NodeSVGFactory and have it in the library.

timvandermeij:
Thank you for the review! I incorporated all the improvements you mentioned and verified that everything still works.

PR #8692, opened 2017-07-24
Add a reduced test-case for issue 7696

Snuffleupagus:
Issue #7696 was one of the issues fixed by PR #8580. The other ones were all cases of missing glyphs, however in this particular one glyphs did render but every single one was incorrect.
Hence it probably cannot hurt to have a small, reduced, reference test for that PDF file as well.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #8696, opened 2017-07-24
Adds ready capability rejection logic for stream sink.

mukulmishra18:
This PR adds rejection logic for stream sink ready capability. Also adds `isCancelled` property in stream sink, so that we can ignore any `enqueue` and `close` operations after stream sink is closed.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -1438,6 +1438,9 @@ MessageHandler.prototype = {
 
     let streamSink = {
       enqueue(chunk, size = 1) {
+        if (this.isCancelled) {
+          return;
+        }
         let lastDesiredSize = this.desiredSize;
         this.desiredSize -= size;
         // Enqueue decreases the desiredSize property of sink,
@@ -1451,6 +1454,9 @@ MessageHandler.prototype = {
       },
 
       close() {
+        if (this.isCancelled) {
+          return;
+        }
         sendStreamRequest({ stream: 'close', });
         delete self.streamSinks[streamId];
       },
@@ -1462,6 +1468,7 @@ MessageHandler.prototype = {
       sinkCapability: capability,
       onPull: null,
       onCancel: null,
+      isCancelled: false,
       desiredSize,
       ready: null,
     };
@@ -1564,6 +1571,8 @@ MessageHandler.prototype = {
           sendStreamResponse({ stream: 'cancel_complete',
                                success: false, reason, });
         });
+        this.streamSinks[data.streamId].sinkCapability.reject(data.reason);
+        this.streamSinks[data.streamId].isCancelled = true;
         delete this.streamSinks[data.streamId];
         break;
       default:
--- a/test/unit/util_stream_spec.js
+++ b/test/unit/util_stream_spec.js
@@ -154,7 +154,7 @@ describe('util_stream', function () {
         expect(log).toEqual('01p2');
         return reader.cancel();
       }).then(() => {
-        expect(log).toEqual('01p2c');
+        expect(log).toEqual('01p2c4');
         done();
       });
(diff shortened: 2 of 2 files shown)


yurydelendik:
Thank you for the patch.

Snuffleupagus:
This cased issues with cancelling `getTextContent` in the default viewer, when zooming in/out *very* quickly, warnings of the following format are now printed the console:
```
Warning: getTextContent - ignoring errors during task: GetTextContent: page 0     util.js:286:11
Warning: getTextContent - ignoring errors during task: GetTextContent: page 1     util.js:286:11
...
```

The problem seems to be line https://github.com/mozilla/pdf.js/pull/8696/files#diff-8d2b578d58d4d8c8a2975020fbbc84d9R1574, since we're now rejecting indiscriminately.
I might be missing something, but why aren't we doing something like this instead:
```diff
diff --git a/src/shared/util.js b/src/shared/util.js
index 26ac9ba6..ded6eb16 100644
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -1593,12 +1593,15 @@ MessageHandler.prototype = {
         resolveCall(this.streamSinks[data.streamId].onCancel,
                     [wrapReason(data.reason)]).then(() => {
           sendStreamResponse({ stream: 'cancel_complete', success: true, });
+
+          this.streamSinks[data.streamId].sinkCapability.resolve();
         }, (reason) => {
           sendStreamResponse({ stream: 'cancel_complete',
                                success: false, reason, });
+
+          this.streamSinks[data.streamId].sinkCapability.reject(
+            wrapReason(reason));
         });
-        this.streamSinks[data.streamId].sinkCapability.
-          reject(wrapReason(data.reason));
         this.streamSinks[data.streamId].isCancelled = true;
         delete this.streamSinks[data.streamId];
         break;
```

/cc @mukulmishra18, @yurydelendik

mukulmishra18:
We are rejecting `sinkCapability` for `cancel` event(not for rejection of resolveCall), so I think in both state of resolveCall(means resolve and reject), we have to reject the `sinkCapability`. Also we have to reject `sinkCapability` with `data.reason` not `reason` of resolveCall rejection.

So, maybe this will be better?
```javascript
resolveCall(this.streamSinks[data.streamId].onCancel,
            [wrapReason(data.reason)]).then(() => {
  sendStreamResponse({ stream: 'cancel_complete', success: true, });
  this.streamSinks[data.streamId].sinkCapability.
    reject(wrapReason(data.reason));
}, (reason) => {
  sendStreamResponse({ stream: 'cancel_complete',
                       success: false, reason, });
  this.streamSinks[data.streamId].sinkCapability.
    reject(wrapReason(data.reason));
});
```

yurydelendik:
@Snuffleupagus I prefer to propagate the original reason whenever we can. So having reject() might be a better choice here.

I tried https://github.com/yurydelendik/pdf.js/commit/e9aec154dd2c80ec2b669d59dbdb6c818a3bcdb0, I still seldom getting warns -- maybe we need to add checks somewhere else.

Snuffleupagus:
> I tried yurydelendik/pdf.js@e9aec15, I still seldom getting warns -- maybe we need to add checks somewhere else.

I can confirm that the above patch reduces the frequency of the warnings, but does *not* prevent them completely.
Furthermore, I suppose that it'd be nice if it's (at all) possible to come up with a slightly more generally solution, rather than having to remember to add `if (sink.isCancelled) { return; }` checks at various places where streaming is used.

yurydelendik:
> possible to come up with a slightly more general solution

I think it's possible, if we add e.g. `CancelTask` error and gracefully abandon transactions by the reason type check.

PR #8698, opened 2017-07-25
Add a fallback for non-embedded SegoeUISymbol font (issue 8697)

### LINKED ISSUE #8697: PDF Does not render correctly
Link to PDF file:  https://drive.google.com/file/d/0ByuqjywXVZmhbUttYXpqeHZzR1k/view?usp=sharing

Configuration:
- Web browser and its version: Firefox 54.0.1 (64-bit)
- Operating system and its version: Arch Linux
- PDF.js version: 1.7.348
- Is an extension: no / I don't think so?

1. View the PDF
2. Scroll to the second page

2017-07-25-122116_993x720_scrot
 
What went wrong? 
2017-07-25-122031_1362x671_scrot

Snuffleupagus:
The PDF file uses a non-embedded SegoeUISymbol font, which is *not* a standard font (and is mainly used by Microsoft, see https://en.wikipedia.org/wiki/Segoe).

Fixes #8697.

### DIFF
--- a/src/core/standard_fonts.js
+++ b/src/core/standard_fonts.js
@@ -57,6 +57,7 @@ var getStdFontMap = getLookupTableFactory(function (t) {
   t['Helvetica-BoldOblique'] = 'Helvetica-BoldOblique';
   t['Helvetica-Italic'] = 'Helvetica-Oblique';
   t['Helvetica-Oblique'] = 'Helvetica-Oblique';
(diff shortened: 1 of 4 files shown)


yurydelendik (approved):
Thank you for the patch.

timvandermeij:
Thank you!

PR #8712, opened 2017-07-27
Adds node.js logic for networking tasks for PDF.js

mukulmishra18:
This PR implements [IPDFStream](https://github.com/mozilla/pdf.js/blob/master/src/core/worker.js#L62) interface in node.js environment.  `fs.createReadStream` is used to create `ReadableStream`, that can be used to stream PDF data incrementally.

### DIFF
--- a/src/display/network.js
+++ b/src/display/network.js
@@ -14,7 +14,7 @@
  */
 
 import {
-  assert, createPromiseCapability, isInt, MissingPDFException,
+  assert, createPromiseCapability, MissingPDFException,
   UnexpectedResponseException
 } from '../shared/util';
 import globalScope from '../shared/global_scope';
@@ -352,18 +352,16 @@ function PDFNetworkStreamFullRequestReader(manager, options) {
 }
 
 PDFNetworkStreamFullRequestReader.prototype = {
-  getResponseHeader(name) {
-    let fullRequestXhrId = this._fullRequestId;
-    let fullRequestXhr = this._manager.getRequestXhr(fullRequestXhrId);
-
-    return fullRequestXhr.getResponseHeader(name);
-  },
-
   _onHeadersReceived:
       function PDFNetworkStreamFullRequestReader_onHeadersReceived() {
+    var fullRequestXhrId = this._fullRequestId;
+    var fullRequestXhr = this._manager.getRequestXhr(fullRequestXhrId);
+
     let { allowRangeRequests, suggestedLength, } =
       validateRangeRequestCapabilities({
-        getResponseHeader: this.getResponseHeader.bind(this),
+        getResponseHeader: (name) => {
+          return fullRequestXhr.getResponseHeader(name);
+        },
         isHttp: this._manager.isHttp,
         rangeChunkSize: this._rangeChunkSize,
         disableRange: this._disableRange,
@@ -377,7 +375,6 @@ PDFNetworkStreamFullRequestReader.prototype = {
     }
 
     var networkManager = this._manager;
-    var fullRequestXhrId = this._fullRequestId;
     if (networkManager.isStreamingRequest(fullRequestXhrId)) {
       // We can continue fetching when progressive loading is enabled,
       // and we don't need the autoFetch feature.
--- a/src/display/network.js
+++ b/src/display/network.js
@@ -18,6 +18,7 @@ import {
   UnexpectedResponseException
 } from '../shared/util';
 import globalScope from '../shared/global_scope';
+import { validateRangeRequestCapabilities } from './network_utils';
 
 if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('FIREFOX || MOZCENTRAL')) {
   throw new Error('Module "./network" shall not ' +
@@ -351,51 +352,27 @@ function PDFNetworkStreamFullRequestReader(manager, options) {
 }
 
 PDFNetworkStreamFullRequestReader.prototype = {
-  _validateRangeRequestCapabilities: function
-      PDFNetworkStreamFullRequestReader_validateRangeRequestCapabilities() {
+  getResponseHeader(name) {
+    let fullRequestXhrId = this._fullRequestId;
+    let fullRequestXhr = this._manager.getRequestXhr(fullRequestXhrId);
 
-    if (this._disableRange) {
-      return false;
-    }
-
-    var networkManager = this._manager;
-    if (!networkManager.isHttp) {
-      return false;
-    }
-    var fullRequestXhrId = this._fullRequestId;
-    var fullRequestXhr = networkManager.getRequestXhr(fullRequestXhrId);
-    if (fullRequestXhr.getResponseHeader('Accept-Ranges') !== 'bytes') {
-      return false;
-    }
-
-    var contentEncoding =
-      fullRequestXhr.getResponseHeader('Content-Encoding') || 'identity';
-    if (contentEncoding !== 'identity') {
-      return false;
-    }
-
-    var length = fullRequestXhr.getResponseHeader('Content-Length');
-    length = parseInt(length, 10);
-    if (!isInt(length)) {
-      return false;
-    }
-
-    this._contentLength = length; // setting right content length
-
-    if (length <= 2 * this._rangeChunkSize) {
-      // The file size is smaller than the size of two chunks, so it does
-      // not make any sense to abort the request and retry with a range
-      // request.
-      return false;
-    }
-
-    return true;
+    return fullRequestXhr.getResponseHeader(name);
   },
 
   _onHeadersReceived:
       function PDFNetworkStreamFullRequestReader_onHeadersReceived() {
+    let { allowRangeRequests, suggestedLength, } =
+      validateRangeRequestCapabilities({
+        getResponseHeader: this.getResponseHeader.bind(this),
+        isHttp: this._manager.isHttp,
+        rangeChunkSize: this._rangeChunkSize,
+        disableRange: this._disableRange,
+      });
 
-    if (this._validateRangeRequestCapabilities()) {
+    // Setting right content length.
+    this._contentLength = suggestedLength || this._contentLength;
+
+    if (allowRangeRequests) {
       this._isRangeSupported = true;
     }
 
--- a/src/display/network.js
+++ b/src/display/network.js
@@ -18,7 +18,6 @@ import {
   UnexpectedResponseException
 } from '../shared/util';
 import globalScope from '../shared/global_scope';
-import { setPDFNetworkStreamClass } from './api';
 
 if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('FIREFOX || MOZCENTRAL')) {
   throw new Error('Module "./network" shall not ' +
@@ -593,8 +592,6 @@ PDFNetworkStreamRangeRequestReader.prototype = {
     this._close();
   },
 };
-
-setPDFNetworkStreamClass(PDFNetworkStream);
 
 export {
   PDFNetworkStream,
--- a/src/display/network_utils.js
+++ b/src/display/network_utils.js
@@ -13,10 +13,11 @@
  * limitations under the License.
  */
 
-import { isInt } from '../shared/util';
+import { assert, isInt } from '../shared/util';
 
 function validateRangeRequestCapabilities({ getResponseHeader, isHttp,
                                             rangeChunkSize, disableRange, }) {
+  assert(rangeChunkSize > 0);
   let returnValues = {
     allowRangeRequests: false,
     suggestedLength: undefined,
--- a/src/display/network_utils.js
+++ b/src/display/network_utils.js
@@ -1,0 +1,56 @@
+/* Copyright 2012 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { isInt } from '../shared/util';
+
+function validateRangeRequestCapabilities({ getResponseHeader, isHttp,
+                                            rangeChunkSize, disableRange, }) {
+  let returnValues = {
+    allowRangeRequests: false,
+    suggestedLength: undefined,
+  };
+  if (disableRange || !isHttp) {
+    return returnValues;
+  }
+  if (getResponseHeader('Accept-Ranges') !== 'bytes') {
+    return returnValues;
+  }
+
+  let contentEncoding = getResponseHeader('Content-Encoding') || 'identity';
+  if (contentEncoding !== 'identity') {
+    return returnValues;
+  }
+
+  let length = getResponseHeader('Content-Length');
+  length = parseInt(length, 10);
+  if (!isInt(length)) {
+    return returnValues;
+  }
+
+  returnValues.suggestedLength = length;
+  if (length <= 2 * rangeChunkSize) {
+    // The file size is smaller than the size of two chunks, so it does
+    // not make any sense to abort the request and retry with a range
+    // request.
+    return returnValues;
+  }
+
+  returnValues.allowRangeRequests = true;
+  return returnValues;
+}
+
+export {
+  validateRangeRequestCapabilities,
+};
--- a/src/display/node_stream.js
+++ b/src/display/node_stream.js
@@ -29,7 +29,8 @@ class PDFNodeStream {
     this.url = url.parse(this.source.url);
     this.isHttp = this.url.protocol === 'http:' ||
                   this.url.protocol === 'https:';
-    this.isFsUrl = !this.url.host;
+    // Check if url refers to filesystem.
+    this.isFsUrl = this.url.protocol === 'file:' || !this.url.host;
     this.httpHeaders = (this.isHttp && this.source.httpHeaders) || {};
 
     this._fullRequest = null;
@@ -71,10 +72,19 @@ class BaseFullReader {
     this._errored = false;
     this._reason = null;
     this.onProgress = null;
-    this._length = stream.source.length;
+    this._contentLength = stream.source.length; // optional
     this._loaded = 0;
 
-    this._fullRequest = null;
+    this._disableRange = stream.options.disableRange || false;
+    this._rangeChunkSize = stream.source.rangeChunkSize;
+    if (!this._rangeChunkSize && !this._disableRange) {
+      this._disableRange = true;
+    }
+
+    this._isStreamingSupported = !stream.source.disableStream;
+    this._isRangeSupported = !stream.options.disableRange;
+
+    this._readableStream = null;
     this._readCapability = createPromiseCapability();
     this._headersCapability = createPromiseCapability();
   }
@@ -84,7 +94,7 @@ class BaseFullReader {
   }
 
   get contentLength() {
-    return this._length;
+    return this._contentLength;
   }
 
   get isRangeSupported() {
@@ -104,7 +114,7 @@ class BaseFullReader {
         return Promise.reject(this._reason);
       }
 
-      let chunk = this._fullRequest.read();
+      let chunk = this._readableStream.read();
       if (chunk === null) {
         this._readCapability = createPromiseCapability();
         return this.read();
@@ -113,16 +123,52 @@ class BaseFullReader {
       if (this.onProgress) {
         this.onProgress({
           loaded: this._loaded,
-          total: this._length,
+          total: this._contentLength,
         });
       }
-      return Promise.resolve({ value: chunk, done: false, });
+      // Ensure that `read()` method returns ArrayBuffer.
+      let buffer = new Uint8Array(chunk).buffer;
+      return Promise.resolve({ value: buffer, done: false, });
     });
   }
 
   cancel(reason) {
-    this._fullRequest.close(reason);
-    this._fullRequest.destroy(reason);
+    // Call `this._error()` method when cancel is called
+    // before _readableStream is set.
+    if (!this._readableStream) {
+      this._error(reason);
+      return;
+    }
+    this._readableStream.destroy(reason);
+  }
+
+  _error(reason) {
+    this._errored = true;
+    this._reason = reason;
+    this._readCapability.resolve();
+  }
+
+  _setReadableStream(readableStream) {
+    this._readableStream = readableStream;
+    readableStream.on('readable', () => {
+      this._readCapability.resolve();
+    });
+
+    readableStream.on('end', () => {
+      // Destroy readable to minimize resource usage.
+      readableStream.destroy();
+      this._done = true;
+      this._readCapability.resolve();
+    });
+
+    readableStream.on('error', (reason) => {
+      this._error(reason);
+    });
+
+    // Destroy ReadableStream if already in errored state.
+    if (this._errored) {
+      this._readableStream.destroy(this._reason);
+    }
   }
 }
 
@@ -133,13 +179,15 @@ class BaseRangeReader {
     this._errored = false;
     this._reason = null;
     this.onProgress = null;
-    this._length = stream.source.length;
     this._loaded = 0;
+    this._readableStream = null;
     this._readCapability = createPromiseCapability();
+
+    this._isStreamingSupported = !stream.source.disableStream;
   }
 
   get isStreamingSupported() {
-    return false;
+    return this._isStreamingSupported;
   }
 
   read() {
@@ -151,80 +199,88 @@ class BaseRangeReader {
         return Promise.reject(this._reason);
       }
 
-      let chunk = this._read();
+      let chunk = this._readableStream.read();
       if (chunk === null) {
         this._readCapability = createPromiseCapability();
         return this.read();
       }
       this._loaded += chunk.length;
       if (this.onProgress) {
-        this.onProgress({
-          loaded: this._loaded,
-          total: this._length,
-        });
+        this.onProgress({ loaded: this._loaded, });
       }
-      return Promise.resolve({ value: chunk, done: false, });
+      // Ensure that `read()` method returns ArrayBuffer.
+      let buffer = new Uint8Array(chunk).buffer;
+      return Promise.resolve({ value: buffer, done: false, });
     });
   }
+
+  cancel(reason) {
+    // Call `this._error()` method when cancel is called
+    // before _readableStream is set.
+    if (!this._readableStream) {
+      this._error(reason);
+      return;
+    }
+    this._readableStream.destroy(reason);
+  }
+
+  _error(reason) {
+    this._errored = true;
+    this._reason = reason;
+    this._readCapability.resolve();
+  }
+
+  _setReadableStream(readableStream) {
+    this._readableStream = readableStream;
+    readableStream.on('readable', () => {
+      this._readCapability.resolve();
+    });
+
+    readableStream.on('end', () => {
+      // Destroy readableStream to minimize resource usage.
+      readableStream.destroy();
+      this._done = true;
+      this._readCapability.resolve();
+    });
+
+    readableStream.on('error', (reason) => {
+      this._error(reason);
+    });
+
+    // Destroy readableStream if already in errored state.
+    if (this._errored) {
+      this._readableStream.destroy(this._reason);
+    }
+  }
+}
+
+function createRequestOptions(url, headers) {
+  return {
+    protocol: url.protocol,
+    auth: url.auth,
+    host: url.hostname,
+    port: url.port,
+    path: url.path,
+    method: 'GET',
+    headers,
+  };
 }
 
 class PDFNodeStreamFullReader extends BaseFullReader {
   constructor(stream) {
     super(stream);
 
-    this._disableRange = stream.options.disableRange || false;
-    this._rangeChunkSize = stream.source.rangeChunkSize;
-    if (!this._rangeChunkSize && !this._disableRange) {
-      this._disableRange = true;
-    }
-
-    this._isStreamingSupported = !stream.source.disableStream;
-    this._isRangeSupported = false;
-
-    let options = {
-      host: this._url.host,
-      path: this._url.path,
-      method: 'GET',
-      headers: stream.httpHeaders,
-    };
-
     let handleResponse = (response) => {
       this._headersCapability.resolve();
-      this._fullRequest = response;
+      this._setReadableStream(response);
 
-      response.on('readable', () => {
-        this._readCapability.resolve();
-      });
-
-      response.on('end', () => {
-        // Destroy response to minimize resource usage.
-        response.destroy();
-        this._done = true;
-        this._readCapability.resolve();
-      });
-
-      response.on('error', (reason) => {
-        this._errored = true;
-        this._reason = reason;
-        this._readCapability.resolve();
-      });
-    };
-
-    this._request = this._url.protocol === 'http:' ?
-      http.request(options, handleResponse) :
-      https.request(options, handleResponse);
-
-    this._request.on('error', (reason) => {
-      this._errored = true;
-      this._reason = reason;
-      this._headersCapability.reject(reason);
-    });
-    this._request.end();
-
-    this._headersCapability.promise.then(() => {
       let { allowRangeRequests, suggestedLength, } =
       validateRangeRequestCapabilities({
-        getResponseHeader: this.getResponseHeader.bind(this),
+        getResponseHeader: (name) => {
+          // Make sure that headers name are in lower case, as mentioned
+          // here: https://nodejs.org/api/http.html#http_message_headers.
+          return this._readableStream.headers[name.toLowerCase()];
+        },
         isHttp: stream.isHttp,
         rangeChunkSize: this._rangeChunkSize,
         disableRange: this._disableRange,
@@ -233,12 +289,28 @@ class PDFNodeStreamFullReader extends BaseFullReader {
       if (allowRangeRequests) {
         this._isRangeSupported = true;
       }
-      this._length = suggestedLength;
-    });
-  }
+      // Setting right content length.
+      this._contentLength = suggestedLength;
+    };
 
-  getReasponseHeader(name) {
-    return this._fullRequest.headers[name];
+    this._request = null;
+    if (this._url.protocol === 'http:') {
+      this._request = http.request(createRequestOptions(
+        this._url, stream.httpHeaders), handleResponse);
+    } else {
+      this._request = https.request(createRequestOptions(
+        this._url, stream.httpHeaders), handleResponse);
+    }
+
+    this._request.on('error', (reason) => {
+      this._errored = true;
+      this._reason = reason;
+      this._headersCapability.reject(reason);
+    });
+    // Note: `request.end(data)` is used to write `data` to request body
+    // and notify end of request. But one should always call `request.end()`
+    // even if there is no data to write -- (to notify the end of request).
+    this._request.end();
   }
 }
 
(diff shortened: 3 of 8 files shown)


src/display/node_stream.js
```diff
@@ -0,0 +1,173 @@
...
+
+  cancel(reason) {
+    this._fullRequest.close(reason);
+  },
+
+  onProgress: null,
+};
+
+function PDFNetworkStreamRangeReader(path, begin, end) {
+  this._rangeRequest = fs.createReadStream(path, { begin, end, });
```
yurydelendik: change begin=>start `fs.createReadStream(path, { start, end: 99 });`

src/display/node_stream.js
```diff
@@ -0,0 +1,173 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+let fs = require('fs');
+let assert = require('assert');
+
+function createPromiseCapabilities() {
```
Snuffleupagus: Please don't duplicate this functionality, just import it from `src/shared/util.js` instead.

src/display/node_stream.js
```diff
@@ -0,0 +1,173 @@
...
+      if (chunk === null) {
+        this._readCapability = createPromiseCapabilities();
+        return this.read();
+      }
+      return Promise.resolve({ value: chunk, done: false, });
+    });
+  },
+
+  cancel(reason) {
+    this._fullRequest.cancel(reason);
```
yurydelendik: `this._rangeRequeste.destroy(reason);`

src/display/node_stream.js
```diff
@@ -0,0 +1,173 @@
...
+
+  this._rangeRequest.on('error', (reason) => {
+    this._errored = true;
+    this._reason = reason;
+    this._readCapability.resolve();
+  });
+}
+PDFNetworkStreamRangeReader.prototype = {
+  get isStreamingSupported() {
+
```
yurydelendik: `return true;`

src/display/node_stream.js
```diff
@@ -0,0 +1,173 @@
...
+      if (chunk === null) {
+        this._readCapability = createPromiseCapabilities();
+        return this.read();
+      }
+      return Promise.resolve({ value: chunk, done: false, });
+    });
+  },
+
+  cancel(reason) {
+    this._fullRequest.close(reason);
```
yurydelendik: `this._fullRequest.destroy(reason);`

src/display/node_stream.js
```diff
@@ -0,0 +1,173 @@
...
+  get contentLength() {
+
+  },
+
+  get isRangeSupported() {
+
+  },
+
+  get isStreamingSupported() {
+
```
yurydelendik: `return true;`

src/display/node_stream.js
```diff
@@ -0,0 +1,173 @@
...
+  });
+}
+PDFNetworkStreamFullReader.prototype = {
+
+  get contentLength() {
+
+  },
+
+  get isRangeSupported() {
+
```
yurydelendik: `return true;`

src/display/node_stream.js
```diff
@@ -0,0 +1,173 @@
...
+  this._fullRequest.on('error', (reason) => {
+    this._errored = true;
+    this._reason = reason;
+    this._readCapability.resolve();
+  });
+}
+PDFNetworkStreamFullReader.prototype = {
+
+  get contentLength() {
+
```
yurydelendik: add fs.Stats are received, we will know stream length

yurydelendik (commented):
few more changes:
  - `fs.lstat` can be used to get length -- it is async so you can use it as a part of headersReady
  - rename PDFNetworkStream to PDFNodeStream (and class syntax can be used here)
 - at "src/pdf.js", import setPDFNetworkStream and depending on [`isNodeJS`](https://github.com/mozilla/pdf.js/blob/343b4dc2b61c88f9b549f212b5d91a8b7acd4209/src/shared/util.js#L1094), use it to set PDFNetworkStream or PDFNodeStream.

src/display/node_stream.js
mukulmishra18: Done with new commit.

src/display/node_stream.js
mukulmishra18: Done. Thanks.

src/display/node_stream.js
mukulmishra18: Done.

src/display/node_stream.js
mukulmishra18: Done. Thanks.

src/display/node_stream.js
mukulmishra18: Done. Thanks.

src/display/node_stream.js
mukulmishra18: Done.

src/display/node_stream.js
mukulmishra18: Done.

src/display/node_stream.js
mukulmishra18: Done.

src/display/network.js
```diff
@@ -351,51 +351,24 @@ function PDFNetworkStreamFullRequestReader(manager, options) {
...
+    return fullRequestXhr.getResponseHeader(name);
   },
 
   _onHeadersReceived:
       function PDFNetworkStreamFullRequestReader_onHeadersReceived() {
-
-    if (this._validateRangeRequestCapabilities()) {
+    // Setting right content length.
+    this._contentLength =
+      parseInt(this.getResponseHeader('Content-Length'), 10);
```
Snuffleupagus: Won't just assigning a value, without doing any validation, mean that if the Content-Length header isn't present then `this._contentLength` will be `NaN` which probably isn't very good (compare with the old code which uses a `!isInt(...)` check)?

*The following might not matter, but asking anyway since I'm not sure:*
With this patch, you're always setting `this._contentLength` whereas before that was only done when range requests were supported; does this matter at all?

yurydelendik: In refactoring patches we want to preserve initial logic as much as possible, so I recommend to return multiple values/object from validateRangeRequestCapabilities, e.g. `{allowRangeRequests: ..., suggestedLength: ... }`

mukulmishra18: > `this._contentLength` will be `NaN` which probably isn't very good (compare with the old code which uses a !isInt(...) check)?

Yeah, you are right. Sorry I didn't noticed that. I will add check for `isInt`. Thanks.

>  you're always setting this._contentLength whereas before that was only done when range requests were supported

This commit is not intended to change any functionality of the network.js file(just to extract the logic for range request support). I think setting `this._contentLength` inside `if (validateRangeRequestCapability) {...` would be better. Thanks for pointing out.

src/display/node_stream.js
```diff
@@ -0,0 +1,353 @@
...
+import { assert, createPromiseCapability } from '../shared/util';
+import { validateRangeRequestCapabilities } from './network_utils';
+
+class PDFNodeStream {
+  constructor(options) {
+    this.options = options;
+    this.source = options.source;
+    this.url = url.parse(this.source.url);
+    this.isHttp = this.url.protocol === 'http' ||
+                  this.url.protocol === 'https';
```
Rob--W: the `protocol` property of a parsed URL ends with a colon, so replace `'http'` with `'http:'` (similarly for `'https'`).

src/display/node_stream.js
```diff
@@ -0,0 +1,353 @@
...
+    this._isStreamingSupported = !stream.source.disableStream;
+    this._isRangeSupported = false;
+
+    let options = {
+      host: this._url.host,
+      path: this._url.path,
+      method: 'GET',
+      headers: stream.httpHeaders,
+    };
+    this.request = http.request(options, (response) => {
```
Rob--W: You have to check whether `this._url.protocol` is `https:`, and if so, use the https.request, and http.request otherwise. Similarly for other places where you use `http.request`.

src/display/node_stream.js
```diff
@@ -0,0 +1,353 @@
...
+    let rangeStr = start + '-' + (end - 1);
+    stream.httpHeaders['Range'] = 'bytes=' + rangeStr;
+
+    let options = {
+      host: this._url.host,
+      path: this._url.path,
+      method: 'GET',
+      headers: stream.httpHeaders,
+    };
+    this.request = http.request(options, (response) => {
```
Rob--W: `this.request` should be private, `this._request` at the very least.

And also see my previous comment about `http.request.`

src/display/node_stream.js
```diff
@@ -0,0 +1,353 @@
...
+          loaded: this._loaded,
+          total: this._length,
+        });
+      }
+      return Promise.resolve({ value: chunk, done: false, });
+    });
+  }
+
+  cancel(reason) {
+    this._rangeRequest.cancel(reason);
```
Rob--W: `_rangeRequest` should not be a property of the base class, because the base class does not need to know about the implementation detail, so you should remove the use of `_rangeRequest` here. Looking at the implementation, you should probably not directly expose `_rangeRequest`, but possibly store a reference to the `.read` method, maybe as `_read`.  
Besides, the `_rangeRequest` object is not the same in the subclasses:

The argument of the callback to `http.request` is an IncomingMessage, which does not have a `cancel` method - https://nodejs.org/api/http.html#http_class_http_incomingmessage

In PDFNodeStreamFsRangeReader, the type is a ReadableStream, which does not have a `cancel` method either.

src/display/node_stream.js
```diff
@@ -0,0 +1,353 @@
...
+    this._url = stream.url;
+    this._done = false;
+    this._errored = false;
+    this._reason = null;
+    this.onProgress = null;
+    this._length = stream.source.length;
+    this._loaded = 0;
+    this._readCapability = createPromiseCapability();
+
+    this._rangeRequest = null;
```
Rob--W: This should be removed, see my note below about `_rangeRequest`.

src/display/network.js
```diff
@@ -593,8 +569,6 @@ PDFNetworkStreamRangeRequestReader.prototype = {
   },
 };
 
-setPDFNetworkStreamClass(PDFNetworkStream);
```
Rob--W: Why this change?

Rob--W (changes requested):
I did a cursory review, and found some issues.

Some observations:

- It seems that you have based the logic on `src/display/transport_stream.js`. This is not necessarily a bad thing, but I wonder whether this indeed makes most sense in Node.js.

- The implementation does not try to minimize resource usage. At the very least, once a request has finished and the response has yielded, it is probably sufficient to discard the underlying handle to the data (e.g. the IncomingMessage).

- There are no unit tests that demonstrates the new functionality, let alone that the functionality works as expected. This could have caught some of the bugs that I mentioned below. Please add them. If you need inspiration: Currently we have one Node.js-specific test, namely https://github.com/mozilla/pdf.js/blob/5b5781b45d234666241bf3354c0d390315c31d1a/test/unit/display_svg_spec.js.

src/display/network.js
mukulmishra18: In [src/pdf.js](https://github.com/mukulmishra18/pdf.js/commit/712de53d0ac469348a4d05285e4be42758e5cb8f#diff-3841ca9d0e4702635b24c39b593d0610R34), we are trying to detect the environment(using `isNodeJS`), and then setting `PDFNodeStream` or `PDFNetworkStream`, see one of the TODO here: https://github.com/mozilla/pdf.js/pull/8712#pullrequestreview-52920496

src/display/node_stream.js
mukulmishra18: > but possibly store a reference to the .read method, maybe as _read.

Sorry, but I really don't understand your concern here. Should we remove `cancel` from `BaseRangeReader` class, and also not call `this._rangeReder.read()` in `read()` method ?
Can you please explain it a bit?

mukulmishra18: > but possibly store a reference to the `.read` method, maybe as `_read`.

As far as I understand, do you mean to store reference of `.read` method of `this._rangeRequest` in `_read` property of `PDFNodeStreamRangeReader`(same for `PDFNodeStreamFsRangeReader`), somthing like: `this._read = this._rangeRequest.read`, and then use `this._read` inside `read()` of base class?

src/display/network.js
```diff
@@ -351,51 +351,27 @@ function PDFNetworkStreamFullRequestReader(manager, options) {
...
-
-    return true;
+    return fullRequestXhr.getResponseHeader(name);
   },
 
   _onHeadersReceived:
       function PDFNetworkStreamFullRequestReader_onHeadersReceived() {
+    let { allowRangeRequests, suggestedLength, } =
+      validateRangeRequestCapabilities({
+        getResponseHeader: this.getResponseHeader.bind(this),
```
yurydelendik: don't bind() -- create a function here e.g. `getResponseHeader: (name) => { return fullRequestXhr.getResponseHeader(name); },`

src/display/node_stream.js
```diff
@@ -0,0 +1,374 @@
...
+          total: this._length,
+        });
+      }
+      return Promise.resolve({ value: chunk, done: false, });
+    });
+  }
+
+  cancel(reason) {
+    this._fullRequest.close(reason);
+    this._fullRequest.destroy(reason);
```
yurydelendik: not sure if we need two of these. can we use either close() or destroy()?

src/display/node_stream.js
```diff
@@ -0,0 +1,374 @@
...
+    this._errored = false;
+    this._reason = null;
+    this.onProgress = null;
+    this._length = stream.source.length;
+    this._loaded = 0;
+    this._readCapability = createPromiseCapability();
+  }
+
+  get isStreamingSupported() {
+    return false;
```
yurydelendik: I prefer to have it set to true here, if it's enabled in options.

src/display/node_stream.js
```diff
@@ -0,0 +1,374 @@
...
+    super(stream);
+
+    this._disableRange = stream.options.disableRange || false;
+    this._rangeChunkSize = stream.source.rangeChunkSize;
+    if (!this._rangeChunkSize && !this._disableRange) {
+      this._disableRange = true;
+    }
+
+    this._isStreamingSupported = !stream.source.disableStream;
+    this._isRangeSupported = false;
```
yurydelendik: I prefer to have the above options applied for Fs readers, can we move them into base reader?

src/display/node_stream.js
```diff
@@ -0,0 +1,374 @@
...
+        isHttp: stream.isHttp,
+        rangeChunkSize: this._rangeChunkSize,
+        disableRange: this._disableRange,
+      });
+
+      if (allowRangeRequests) {
+        this._isRangeSupported = true;
+      }
+      this._length = suggestedLength;
+    });
```
yurydelendik: Move this entire function body into the handleResponse -- notice that this._headersCapability.resolve() still has to be called.

src/display/node_stream.js
mukulmishra18: Done. Thanks for pointing out.

src/display/node_stream.js
mukulmishra18: Done.

src/display/node_stream.js
mukulmishra18: Done. Thanks for the advice.

src/display/node_stream.js
mukulmishra18: Done. Thanks.

src/display/network.js
Rob--W: But in non-Node.js, the stream class is no longer set. That's not good.

It is safe to repeatedly call `setPDFNetworkStreamClass`, so restore this line.

src/display/node_stream.js
```diff
@@ -0,0 +1,374 @@
...
+  constructor(stream) {
+    this._url = stream.url;
+    this._done = false;
+    this._errored = false;
+    this._reason = null;
+    this.onProgress = null;
+    this._length = stream.source.length;
+    this._loaded = 0;
+
+    this._fullRequest = null;
```
Rob--W: This name is a bit confusing, because you use the same name elsewhere in the file.
Furthermore, `_fullRequest` is not an accurate description of the value.

How about renaming `_fullRequest` of the *FullReader classes to `_readableStream`? This would also make the code more understandable for those who are already familiar with Node.js streams.

src/display/node_stream.js
```diff
@@ -0,0 +1,374 @@
...
+import { validateRangeRequestCapabilities } from './network_utils';
+
+class PDFNodeStream {
+  constructor(options) {
+    this.options = options;
+    this.source = options.source;
+    this.url = url.parse(this.source.url);
+    this.isHttp = this.url.protocol === 'http:' ||
+                  this.url.protocol === 'https:';
+    this.isFsUrl = !this.url.host;
```
Rob--W: Is `isFsUrl` supposed to say whether a URL refers to the filesystem?
If so, this is not the right way to say so. The `file:` scheme is commonly used to refer to local files, and it can have a host part too - see e.g. https://en.wikipedia.org/wiki/File_URI_scheme

What are the expected values for url that you intend to support here?

Rob--W (changes requested):
This looks better than before. I am still expecting some unit tests to ascertain that the introduced logic works as expected (see my previous review for more details).

src/display/node_stream.js
mukulmishra18: > What are the expected values for url that you intend to support here?

I am checking here if url refers to the filesystem or not, so that we can choose between `PDFNodeStream***Reader` or `PDFNodeStreamFs***Reader`.

I am trying to parse URL using `url.parse`, and then checking for host. In most of cases I found that `url`s are of the form `../....path/to/some/pdf.pdf` or if `url` is absolute `home/userName/path/to/some/pdf.pdf` or `file:///home/user/path/to/some/pdf.pdf`-- (and all these cases works fine with above check). The only case that I found failing is `file://localhost/path/to/some/pdf.pdf`, and I think this can be fix by checking for `protocol === file:`.

Is it sounds reasonable to you, or you have something else in your mind? If not, can you suggest me something concrete?

Rob--W: The canonical URL of a local file is a `file:`-URL, so I suggest to check whether the scheme is `file:`.

src/display/node_stream.js
mukulmishra18: > How about renaming `_fullRequest` of the *FullReader classes to `_readableStream`?

Sounds like a good idea. Do you think we should also rename `_rangeRequest` with `_readableStream`(?) in *RangeReader?

src/display/network.js
mukulmishra18: Done. Thanks.

src/display/node_stream.js
mukulmishra18: Removed `this._fullRequest.close(reason);`. Thanks.

src/display/node_stream.js
mukulmishra18: Done.

src/display/node_stream.js
mukulmishra18: Done.

src/display/node_stream.js
mukulmishra18: Done. Thanks.

mukulmishra18:
I am trying to add some tests for `node_stream` logic[1]. While writing test I find out that following https://github.com/mozilla/pdf.js/pull/8712#discussion_r131277938, breaks `rangeReader.read()` operation, so we are not getting anything while reading [here](https://github.com/mozilla/pdf.js/pull/8712/commits/6e9f924bd1af2c1b4e51571b8b17768e5808f348#diff-b7612f90e11fe611d3880c5a22b3fe27R159). To fix this we have to shift `read()` in `**RangeReader`s class from `BaseRangeReader` class, and call `this._rangeReader.read()` instead.

Although running `node_stream_spec` separately I am getting all output as expected, but I am facing some problems, mainly [here](https://github.com/mozilla/pdf.js/pull/8712/commits/6e9f924bd1af2c1b4e51571b8b17768e5808f348#diff-b7612f90e11fe611d3880c5a22b3fe27R28).

-  When running `gulp unittest`, I am getting `__non_webpack_require__ is not defined`.
- When running `gulp unittestcli`, I am getting different values for `pdf2` as excepted(Getting: `home/user/pdf.js/build/lib/test/pdfs/tracemonkey.pdf`, expected: `home/user/pdf.js/test/pdfs/tracemonkey.pdf`).

[1]: Note that I am following `network_spec` file to write tests, as `node_stream` is just adding `IPDFStream` for node.js environment.

src/display/network.js
```diff
@@ -14,10 +14,11 @@
  */
 
 import {
-  assert, createPromiseCapability, globalScope, isInt, MissingPDFException,
+  assert, createPromiseCapability, globalScope, MissingPDFException,
   UnexpectedResponseException
 } from '../shared/util';
 import { setPDFNetworkStreamClass } from './api';
```
yurydelendik: remove setPDFNetworkStreamClass import from this file

mukulmishra18: I think there is some confusion at https://github.com/mozilla/pdf.js/pull/8712#discussion_r131554047. Done with new commit, Thanks.

src/display/network.js
yurydelendik: webpack can package node_stream and network in one file and initialize these in some random order. so it's better to call setPDFNetworkStreamClass from api.js and remove this line from the network.js

src/display/node_stream.js
```diff
@@ -288,8 +260,29 @@ class PDFNodeStreamRangeReader extends BaseRangeReader {
...
+  read() {
+    return this._readCapability.promise.then(() => {
+      if (this._done) {
+        return Promise.resolve({ value: undefined, done: true, });
+      }
+      if (this._errored) {
+        return Promise.reject(this._reason);
+      }
+
+      let chunk = this._rangeRequest.read();
```
Rob--W: This is duplication of code. The two implementations of `read` are identical.

What stops you from putting `read` in the base class?
If -for some reason- you cannot use `this._rangeRequest.read()` in the base class, then you can create a method in the subclass (e.g. `_read() { return this._rangeRequest.read(); }` and call `this._read();` from the base class instead.

And also, I thought that you had replaced "_rangeRequest" with "_readableStream", but I can not find that change (any more?).

Rob--W:
> Although running `node_stream_spec` separately I am getting all output as expected, but I am facing some problems, mainly [here](https://github.com/mozilla/pdf.js/pull/8712/commits/6e9f924bd1af2c1b4e51571b8b17768e5808f348#diff-b7612f90e11fe611d3880c5a22b3fe27R28).
>
> -  When running `gulp unittest`, I am getting `__non_webpack_require__ is not defined`.

It seems that we are not passing the right `babelOptions` to the transpiler when running tests in the browser.

But even if we do, then the error should still be "require is not defined" or something like "require("fs") is not a module", because it does not make any sense to run the Node.js-only test in the browser.

At minimum, skip the test when you detect that the environment is not Node, e.g. as done here: https://github.com/mozilla/pdf.js/blob/ba5dbc96326518ad716158ef040f61794cc72202/test/unit/display_svg_spec.js#L112-L114

Rob--W:
I've submitted a patch for the `__non_webpack_require__` issue at #8781. Note that an error is still expected, the true solution is to not load the module in non-Node.js environments.

src/display/node_stream.js
mukulmishra18: > What stops you from putting read in the base class?

Sorry, but not sure why when I am doing something like https://github.com/mozilla/pdf.js/pull/8712#discussion_r131416557, I am unable to read from the stream, see https://github.com/mozilla/pdf.js/pull/8712#issuecomment-320933800.

Also `_read() { return this._rangeReques.read() }` creates the same issue.

mukulmishra18:
> At minimum, skip the test when you detect that the environment is not Node.

Okay, sure. Tests for `node_stream` is now removed from `gulp unittest` and only covered in `gulp unittestcli`.

src/display/node_stream.js
Rob--W: The context is missing from the comment that you've linked. I think that the breakage is unrelated to the method being in the base class. You now literally have two (actually, after looking again THREE, `BaseFullReader.prototype.read` is almost identical, except with the "length" property in `onProgress`) identical methods `read` in the sub classes.

The inheritance tree looks like this:

- BaseRangeReader
- PDFNodeStreamRangeReader extends BaseRangeReader
- PDFNodeStreamFsRangeReader extends BaseRangeReader

Your current code duplicates the `read` method in the latter two classes (the implementation is byte for byte identical). The runtime behavior is identical regardless of where you put the implementation of the "read" method, so you should put the implementation in the base class (BaseRangeReader), to improve the maintainability of the code.

src/display/node_stream.js
```diff
@@ -0,0 +1,384 @@
...
+  }
+}
+
+class PDFNodeStreamRangeReader extends BaseRangeReader {
+  constructor(stream, start, end) {
+    super(stream);
+
+    this._readableStream = null;
+    let rangeStr = start + '-' + (end - 1);
+    stream.httpHeaders['Range'] = 'bytes=' + rangeStr;
```
yurydelendik: create new object for httpHeaders (before 'Range' is changed) -- we don't want to pollute original httpHeaders object.

yurydelendik: use `bytes=${start}-${end - 1}` format

src/display/node_stream.js
```diff
@@ -0,0 +1,384 @@
...
+
+    let options = {
+      host: this._url.host,
+      path: this._url.path,
+      method: 'GET',
+      headers: stream.httpHeaders,
+    };
```
yurydelendik: Move options creating into separate function (we have similar code above) and add `auth` and `port` field, also probably `agent`.

src/display/node_stream.js
```diff
@@ -0,0 +1,384 @@
...
+        response.destroy();
+        this._done = true;
+        this._readCapability.resolve();
+      });
+
+      response.on('error', (reason) => {
+        this._errored = true;
+        this._reason = reason;
+        this._readCapability.resolve();
+      });
```
yurydelendik: move all `.on('readable'/'end'/'error',` into  e.g. private `_setReadableStream` into base class -- it can be use by Fs constructor as well.

The same thing needs to be done for "full" readers too.

src/display/node_stream.js
```diff
@@ -0,0 +1,384 @@
...
+        this._readCapability = createPromiseCapability();
+        return this.read();
+      }
+      this._loaded += chunk.length;
+      if (this.onProgress) {
+        this.onProgress({ loaded: this._loaded, });
+      }
+      return Promise.resolve({ value: chunk, done: false, });
+    });
+  }
```
yurydelendik: agree with moving read() and cancal() into base class

src/display/node_stream.js
```diff
@@ -0,0 +1,384 @@
...
+      this._readCapability.resolve();
+    });
+  }
+}
+
+class PDFNodeStreamFsRangeReader extends BaseRangeReader {
+  constructor(stream, start, end) {
+    super(stream);
+
+    this._readableStream = fs.createReadStream(this._url.path, { start, end, });
```
yurydelendik: as mentioned above, use `this._setReadableStream(fs.createReadStream(this._url.path));`

src/display/network_utils.js
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { isInt } from '../shared/util';
+
+function validateRangeRequestCapabilities({ getResponseHeader, isHttp,
+                                            rangeChunkSize, disableRange, }) {
```
yurydelendik: can you add `assert(rangeChunkSize > 0);` here

src/display/node_stream.js
```diff
@@ -0,0 +1,384 @@
...
+
+class PDFNodeStreamFullReader extends BaseFullReader {
+  constructor(stream) {
+    super(stream);
+
+    this._disableRange = stream.options.disableRange || false;
+    this._rangeChunkSize = stream.source.rangeChunkSize;
+    if (!this._rangeChunkSize && !this._disableRange) {
+      this._disableRange = true;
+    }
```
yurydelendik: move this_disableRange and this._rangeChunkSize logic into base class.

mukulmishra18: Done. Thanks.

src/display/node_stream.js
mukulmishra18: Done in new commit. Thanks,

mukulmishra18: Done.

src/display/node_stream.js
mukulmishra18: Created `createRequestOptions` function that returns `options` object. Also tried to add a simple `http.Agent`, please check if it looks good to you. Thanks.

src/display/node_stream.js
mukulmishra18: Done in new commit.

**Note:** I have moved `this._readableStream` property in `BaseRangeReader` from concrete classes and it works just fine for me.

src/display/node_stream.js
mukulmishra18: Done in new commit. Thanks.

src/display/node_stream.js
mukulmishra18: Done.

src/display/network_utils.js
mukulmishra18: Done. Thanks.

src/display/node_stream.js
```diff
@@ -0,0 +1,351 @@
...
+  constructor(stream) {
+    super(stream);
+
+    let handleResponse = (response) => {
+      this._headersCapability.resolve();
+      this._setReadableStream(response);
+
+      let { allowRangeRequests, suggestedLength, } =
+      validateRangeRequestCapabilities({
+        getResponseHeader: this.getResponseHeader.bind(this),
```
yurydelendik: remove getResponseHeader function below and inline the response.headers access here

src/display/node_stream.js
```diff
@@ -0,0 +1,351 @@
...
+    });
+  }
+}
+
+function createRequestOptions(url, headers) {
+  let agentOptions = {
+    keepAlive: true,
+    maxSockets: 10,
+  };
+  let agent = url.protocol === 'http:' ? new http.Agent(agentOptions) :
```
yurydelendik: too complicated, let's remove agent for now.

src/display/node_stream.js
```diff
@@ -0,0 +1,351 @@
...
+          loaded: this._loaded,
+          total: this._length,
+        });
+      }
+      return Promise.resolve({ value: chunk, done: false, });
+    });
+  }
+
+  cancel(reason) {
+    this._readableStream.destroy(reason);
```
yurydelendik: here and at the range base reader, we need to think "what if cancel is called before _readableStream is set". I recommend to extract `on('error'` body into _error(reason) method and call it when this._readableStream is not set. in _setReadableStream we can check if `this._errored = true` we can quickly destroy it.

mukulmishra18: Done in new commit. Thanks.

src/display/node_stream.js
mukulmishra18: Done.

src/display/node_stream.js
mukulmishra18: Done. Thanks.

yurydelendik:
I see `UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Bad end offset: 65537` when running getinfo.js example.

yurydelendik:
If I pass HTTP URL is the getinfo.js I get the following error:

```
Error: Cannot find module '/Users/yury/Work/pdf.js/examples/node/http:/mozilla.github.io/pdf.js/web/compressed.tracemonkey-pldi-09.pdf'
```

mukulmishra18:
> Error: Cannot find module '/Users/yury/Work/pdf.js/examples/node/http:/mozilla.github.io/pdf.js/web/compressed.tracemonkey-pldi-09.pdf'

Strange, seems like we are getting wrong url in `node_stream.js`. Although on running `getinfo.js` locally, I find out that I am getting some initial chunk of data, but not all. I am doing something like:

```javascript
var pdfjsLib = require('../../build/lib/pdf.js');

pdfjs.getDocument('http://mozilla.github.io/pdf.js/web/compressed.tracemonkey-pldi-09.pdf').
  then((doc) => {...});
```
Although, earlier running the above version of `getinfo.js` gives me correct output, seems like using `read()` and `this._readableStream` in `BaseRangeReader` causing this error.

mukulmishra18:
Okay, I find out the problem. The headers received by `http.request`in node.js environment is in lower case(also mentioned here: https://nodejs.org/api/http.html#http_message_headers), but not in browser(when requesting using XHR or fetch) Example: `Accept-Ranges` in XHR or fetch whereas `accept-ranges` in node.js.

This create problem here: https://github.com/mukulmishra18/pdf.js/commit/3493f145257a288ecc03a43d11d58e4450acbceb#diff-35bcb37ef640f9436dd67f5077c08b55R271 -- (returns `undefined` when trying to access `Accept-Ranges`). Changing the requested header to lower case solves the problem. 

`getinfo.js` example is now working fine for me on my local system. @yurydelendik can you please check, if it works for you?

test/unit/node_stream_spec.js
```diff
@@ -0,0 +1,191 @@
...
+ * limitations under the License.
+ */
+/* globals __non_webpack_require__ */
+
+import { isNodeJS } from '../../src/shared/util';
+import { PDFNodeStream } from '../../src/display/node_stream';
+
+describe('node_stream', function() {
+  let pdf1 =
+    'http://mozilla.github.io/pdf.js/web/compressed.tracemonkey-pldi-09.pdf';
```
Rob--W: This results in an external network request, and slow down the unit test if network is slow, or even cause the unit test to fail if network is somehow unavailable.

Please use a local file server in Node for testing. Don't forget to close the server regardless of test failure. To do so you can use `http.createServer in a `beforeAll` hook (and start the server on a random free port using `.listen(0)`, and call `.close()` on the server in an `afterAll` hook. The return value of `.listen` has an `address` method that contains the random port in the `port` property. After starting the server, you can use something like `http://127.0.0.1:${port}/tracemonkey.pdf`.


And for the browser... Why does this test have a special case `!isNode()` if the API that you are testing is only supposed to work on Node? Am I missing something?

test/unit/node_stream_spec.js
```diff
@@ -0,0 +1,191 @@
...
+  let pdfLength = 1016315;
+  let pdf2;
+  if (!isNodeJS()) {
+    pdf2 = new URL('../pdfs/tracemonkey.pdf', window.location).href;
+  } else {
+    let path = __non_webpack_require__('path');
+    let url = __non_webpack_require__('url');
+    pdf2 =
+      url.parse(path.join(process.cwd(), './test/pdfs/tracemonkey.pdf')).href;
```
Rob--W: In this case, `pdf2` looks like `/path/to/pdf.js/test/pdfs/tracemonkey.pdf`.
I was expecting a `file://` prefix, because otherwise it would not be a URL. Does the test really pass? Regardless, can you make sure that the url really looks like a URL and not a path?

Rob--W (changes requested):

src/display/node_stream.js
```diff
@@ -0,0 +1,367 @@
...
+    });
+  }
+}
+
+class PDFNodeStreamFsRangeReader extends BaseRangeReader {
+  constructor(stream, start, end) {
+    super(stream);
+
+    this._setReadableStream(
+      fs.createReadStream(this._url.path, { start, end, }));
```
yurydelendik: I still have `(node:90529) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Bad end offset: 65537`. Node.js operates with `end` being inclusive. Make it `end: end - 1,`

test/unit/node_stream_spec.js
```diff
@@ -0,0 +1,191 @@
...
+
+    // Skipping fullReader results, requesting something from the PDF end.
+    let tailSize = (pdfLength % rangeSize) || rangeSize;
+
+    let range11Reader = stream1.getRangeReader(pdfLength - tailSize - rangeSize,
+                                             pdfLength - tailSize);
+    let range12Reader = stream1.getRangeReader(pdfLength - tailSize, pdfLength);
+
+    let range21Reader = stream2.getRangeReader(pdfLength - tailSize - rangeSize,
+                                             pdfLength - tailSize - 1);
```
yurydelendik: Must be `pdfLength - tailSize` without `- 1` based on our API definition.

test/unit/node_stream_spec.js
yurydelendik: this script is not added to jasmine-bootstrap.js -- let's also add `assert(isNode());` just in case somebody will decide to add it.

yurydelendik (approved):

yurydelendik (approved):
looks good, with Rob's requested changes and off-by-one fix in the fs range reader

yurydelendik:
let's remove fs.readFileSync() from the getinfo.js example to show that we implemented this functionality and it works

test/unit/node_stream_spec.js
mukulmishra18: > Why does this test have a special case !isNode() if the API that you are testing is only supposed to work on Node? 

Sorry, removed with new commit, as we don't that any more.

> let's also add assert(isNode()); just in case somebody will decide to add it.

Done. Thanks.

test/unit/node_stream_spec.js
mukulmishra18: > I was expecting a file:// prefix, because otherwise it would not be a URL.

Then I think adding `file://` with `process.cwd()` will help, something like : `url.parse(path.join('file://' + process.cwd(), './test/pdfs/tracemonkey.pdf'))`

src/display/node_stream.js
mukulmishra18: Done in new commit. Thanks.

test/unit/node_stream_spec.js
mukulmishra18: Done. Thanks.

test/unit/node_stream_spec.js
```diff
@@ -0,0 +1,234 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* globals __non_webpack_require__ */
+
+import { assert, isNodeJS } from '../../src/shared/util';
+import { PDFNodeStream } from '../../src/display/node_stream';
+
+// Make sure that we are not running this script is Node.js environment.
```
Snuffleupagus: Nit: Hasn't this comment accidentally been inverted, since I'm assuming that it should be e.g. `Make sure that we only running this script in Node.js environments.`?

mukulmishra18: Opps..., Thanks for pointing out.

yurydelendik:
Thank you for the patch.

PR #8715, opened 2017-07-28
Use streams-lib as polyfill

yurydelendik:
Chrome natively supports [ReadableStream](https://bugs.chromium.org/p/chromium/issues/detail?id=503491) and Firefox will soon have it (for now via `dom.streams.enabled`). Changing util.js to allow conditionally include ReadableStream as polyfill.

### DIFF
--- a/src/shared/streams_polyfill.js
+++ b/src/shared/streams_polyfill.js
@@ -1,0 +1,24 @@
+/* Copyright 2017 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


src/shared/streams_polyfill.js
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+if (typeof ReadableStream !== 'undefined') {
+  exports.ReadableStream = ReadableStream;
+} else {
+  if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('CHROME')) {
+    throw new Error('ReadableStream polyfill is not found for Chrome bundle');
```
timvandermeij: Why do we need this, i.e., why don't we unconditionally use the polyfill in this case like before? Even though Chrome now provides it natively, older versions may not have it, so in that case falling back to the polyfill should help avoid errors.

yurydelendik:
I'm assuming here that people who use the extension will not be using very old version of chrome

timvandermeij (approved):

timvandermeij:
Looks good. Thank you!

PR #8717, opened 2017-07-28
Move the `XRefMock` in the unit tests to a central location

timvandermeij:
This patch helps to avoid code duplication for this mock since more unit tests are depending on it.


Snuffleupagus (approved):
This is a really nice clean-up, but there's two additional things that I'd like to see here:
 - Please just place the `XRefMock` in the already existing https://github.com/mozilla/pdf.js/blob/master/test/unit/test_utils.js file instead.
 - Please convert the `XRefMock` in https://github.com/mozilla/pdf.js/blob/master/test/unit/evaluator_spec.js to use the new utility class as well.

With the above addressed, r=me. Thanks for doing this!

PR #8718, opened 2017-07-29
Ensure that the loadingBar isn't displayed again when the entire file has already been fetched

Snuffleupagus:
This could only (potentially) happen when `disableAutoFetch` is used.

*Edit:* This at least avoids the weird loadingBar behaviour described in https://github.com/mozilla/pdf.js/pull/8617#issuecomment-318646822, since based on the discussion there it seems that the underlying behaviour may be intermittent (and I really don't have the time to try and get to the bottom of it).

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -846,6 +846,11 @@ let PDFViewerApplication = {
   },
 
   progress(level) {
+    if (this.downloadComplete) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

PR #8719, opened 2017-07-29
Added theme-color tagl

xxczaki:
Please check https://developers.google.com/web/updates/2014/11/Support-for-theme-color-in-Chrome-39-for-Android


timvandermeij:
Since this only works for Chrome on Android and we want to keep the examples minimal and have not browser-specific parts, I'm closing this PR. Developers using the mobile example as a base can of course add this if they want, but it's out of scope for a minimal example. Thanks.

PR #8721, opened 2017-07-29
Reduces the amount of data send via GetDocRequest.

mukulmishra18:
This PR reduces the amount of data sent via `GetDocRequest`. As mentioned [here](https://github.com/mozilla/pdf.js/pull/8617#issuecomment-314561873), this fulfills one of the TODOs of #8617 .

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -340,7 +340,14 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
   }
   return worker.messageHandler.sendWithPromise('GetDocRequest', {
     docId,
-    source,
+    source: {
+      data: source.data,
+      url: source.url,
+      password: source.password,
+      disableAutoFetch: source.disableAutoFetch,
+      rangeChunkSize: source.rangeChunkSize,
+      length: source.length,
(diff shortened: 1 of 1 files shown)


src/display/api.js
```diff
@@ -338,9 +338,17 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
...
+    url: source.url,
+    password: source.password,
+    disableAutoFetch: source.disableAutoFetch,
+    rangeChunkSize: source.rangeChunkSize,
+    length: source.length,
+  };
   return worker.messageHandler.sendWithPromise('GetDocRequest', {
     docId,
-    source,
+    source: reducedSource,
```
Snuffleupagus: Nit: Any reason to not just simply inline the object here directly, like so:
```javascript
source: {
  data: source.data,
  url: source.url,
  // And so on...
},
```

mukulmishra18: Nope, not any specific reason. Thanks, done in new commit.

Snuffleupagus (approved):
Looks good to me, thank you!

PR #8722, opened 2017-07-29
Fix the helloworld example by importing the `network.js` file (PR 8617 follow-up)

Snuffleupagus:
Fixes https://stackoverflow.com/q/45393309/5317012


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #8716, opened 2017-07-28
adds PDFSinglePageViewer.js file and example

nish17:
Please leave a comment in case you find anything wrong with this patch. I'll try my best to figure the issue and try to fix it.

Fixes #8188.

PR #8724, opened 2017-07-30
Implement a `PDFSinglePageViewer` class (issue 8188)

Snuffleupagus:
This PR contains a suggested approach for addressing only https://github.com/mozilla/pdf.js/issues/8188#issuecomment-288729532, but doesn't include a `components` example.[1]
For testing purposes, the second *temporary* commit uses the new `PDFSinglePageViewer` class in the default viewer (all normal viewer functionality appears to work in my quick tests).

---
[1] *Please note:* I initially intended to try and provide feedback on PR #8716, but the new example provided there didn't even run. Fixing that locally, I couldn't help but noticing that the new class there looks *and* also behaves just like a *verbatim* copy of the existing `web/pdf_viewer.js` file; which I cannot imagine is what we want, since then why not simply use `PDFViewer` directly?

It thus seemed to me that the suggested solution in #8716 was too far from workable that it'd be difficult to provide meaningful review comments; and I figured that I'd see how difficult it would be to implement this.


timvandermeij:
I think this would be a really good approach! Duplicating and having to maintain a verbatim copy of the code does not sound like the way to go, whereas this is much easier to manage.

yurydelendik:
Cool, let try it this way. May I suggest splitting into several files as well: rename to BaseViewer, then move the code into separate pdf_viewer.js and pdf_single_page_viewer.js files.

web/pdf_viewer.js
```diff
@@ -953,7 +929,176 @@ class PDFViewer {
...
+    if (pageNumber) {
+      // Ensure that `this._currentPageNumber` is always correct.
+      this._setCurrentPageNumber(pageNumber);
+    }
+    let scrolledDown = this._ensurePageViewVisible();
+
+    scrollIntoView(pageDiv, pageSpot);
+
+    // Always update the scroll direction, since it most likely isn't correct.
+    this.scroll.down = scrolledDown;
```
Snuffleupagus: This is really an ugly hack that doesn't even work in all cases, so a more robust solution for tracking scroll direction seems necessary (taking into account first of all the page-switching direction *and* also the user scroll direction somehow, I'll look into it sometime during the weekend).

Snuffleupagus: Since the code currently relies on `this.scroll` when determining which page should be pre-rendered, and I didn't feel like refactoring all of that here too, this patch is still updating `this.scroll.down` according to the page switching direction.
However it's now done in such a way that it at least should be consistent, since it waits for the next scroll event; and unless someone complains violently I'd lean towards keeping this for now.

Snuffleupagus:
https://github.com/mozilla/pdf.js/issues/8188#issuecomment-288729532 mentions the possibility of using the `PDFSinglePageViewer` in presentation mode, which seems like a really nice idea!
I don't know if we should try to fix that in this PR as well, to try and reduce churn in the code, since I suppose that it may require some further changes?

What's not clear to me is exactly how that would work though, since just replacing a `PDFViewer` instance with a `PDFSinglePageViewer` one probably isn't entirely straightforward. Considering that we have a number of `setViewer` methods that are used to pass around references to the `PDFViewer` instance, we'd need to update all of those when switching to presentation mode. Not to mention all of the state, such as current page/scale/rotation and more, that we'd need to propagate.

If we want to try and address the above in the scope of this PR, then I'd appreciate ideas for how that can be achieved. (If we come up with a good approach it perhaps also be used to a basis for addressing issue #2638.)

Snuffleupagus:
@yurydelendik Do you have any feedback on the questions/ideas in https://github.com/mozilla/pdf.js/pull/8724#issuecomment-320434726, or should we land this as-is and iterate on that in follow-ups instead?

yurydelendik:
> should we just land this as-is and iterate on that in follow-ups instead?

yes. If there is no breaking api changes, let's do just that.

Snuffleupagus:
> If there is no breaking api changes,

Since this patch only touches code in the `web/` folder, I'm assuming that the question was meant to ask if the interface of `PDFViewer` somehow changed with these patches!?
The answered should be a definite *no*, unless I managed to make some sort of typo/error in the code :-)

Since I'm not sure if anyone has really reviewed this, besides quickly skimming though the diff, I'm don't want to land this myself.

timvandermeij:
Nice work! I'll file follow-up issues for the other ideas.

PR #8727, opened 2017-07-31
Fixes polyfill for Safari's URL

### LINKED ISSUE #8726: Error using URL object in safari
Configuration:
- Web browser and its version: Safari Version 10.1.1 (12603.2.4)
- Operating system and its version: macOS Sierra Version 10.12.5
- PDF.js version: 1.8.593
- Is an extension: No

1. Include pdf.js on any html file and load that file
2. Open developer console and paste the following code: 
`new URL(location).searchParams.get('param')`

The browser should check if the provided URL has the specified parameter (screenshot from Google Chrome).
screen shot 2017-07-31 at 12 44 03

What went wrong?
On safari, it changes the URL and redirects it, breaking the page.
screen shot 2017-07-31 at 12 44 40

yurydelendik:
Fixes #8726

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -1705,6 +1705,7 @@ PDFJS.compatibilityChecked = true;
   // Does not process domain names or IP addresses.
   // Does not handle encoding for the query parameter.
   function JURL(url, base /* , encoding */) {
+    url = String(url)
     if (base !== undefined && !(base instanceof JURL)) {
       base = new JURL(String(base));
     }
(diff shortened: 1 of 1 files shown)


src/shared/compatibility.js
```diff
@@ -1705,6 +1705,7 @@ PDFJS.compatibilityChecked = true;
   // Does not process domain names or IP addresses.
   // Does not handle encoding for the query parameter.
   function JURL(url, base /* , encoding */) {
+    url = String(url)
```
caub: jslint needs a ;

GinjiBan:
@yurydelendik 

Why is this pull request closed ?
I'm experiencing the problem, and the changes seems to fix it.

yurydelendik:
> Why is this pull request closed ?

The solution is incomplete and break other stuff. The https://github.com/mozilla/pdf.js/issues/8726 is still opened for grabs

PR #8738, opened 2017-08-02
Delete workerPort to PDFWorker cache after PDFWorker destroy

ChenMachluf:
I investigate that issue: 
FranckFreiburger/vue-pdf#9

Found out that even after PDFWorker is destroy, it not deleted from pdfWorkerPorts , what cause problems when workerPort and reusing pdf.js several times.

PDFDocumentLoadingTask destroy function call to to _worker(PDFWorker) destory.
When we gonna get another document we get a destroyed PDFWorker.

this should fix that

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1519,6 +1519,7 @@ var PDFWorker = (function PDFWorkerClosure() {
         this._webWorker.terminate();
         this._webWorker = null;
       }
+      pdfWorkerPorts.delete(this._port);
       this._port = null;
       if (this._messageHandler) {
         this._messageHandler.destroy();


timvandermeij:
Please squash the commits into one commit. Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for how to do that easily.

yurydelendik:
I need small test case to see how it works. Can you also ensure we are not breaking https://github.com/mozilla/pdf.js/issues/8246#issuecomment-304379386 ?

ChenMachluf:
@yurydelendik  I take the example repo from there and fork it.
fork it https://github.com/ChenMachluf/webpack-pdfjs-repro and commit the dist of my version to the node_modules dir there. it seems to work well :)

yurydelendik (approved):
I cannot think of any side-effects. Change looks good.

yurydelendik:
Thank you for the patch.

PR #8745, opened 2017-08-03
Properly cancel streams and guard at getTextContent.

yurydelendik:
Addresses warnings described at https://github.com/mozilla/pdf.js/pull/8696#issuecomment-319036696

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -14,10 +14,10 @@
  */
 
 import {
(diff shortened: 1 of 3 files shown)


Snuffleupagus (approved):
I can confirm that this gets rid of the console spam when streams are cancelled. Thanks for fixing this!

(Since the test failure in https://github.com/mozilla/pdf.js/pull/8745#issuecomment-320194855 didn't repeat in any of the other tests, I'm assuming that it's intermittent and hence we should be good to land this.)

PR #8746, opened 2017-08-03
More robust getPage() error handling.

yurydelendik:
Attempt to address https://bugzilla.mozilla.org/show_bug.cgi?id=1368927

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -251,6 +251,11 @@ class PDFFindController {
           // Store the pageContent as a string.
           this.pageContents[i] = strBuf.join('');
           extractTextCapability.resolve(i);
+        }, (reason) => {
+          console.error(`Unable to get page ${i + 1} text content`, reason);
+          // Page error -- assuming no text content.
+          this.pageContents[i] = '';
+          extractTextCapability.resolve(i);
         });
       });
     }
--- a/web/pdf_viewer.js
+++ b/web/pdf_viewer.js
@@ -341,7 +341,7 @@ class PDFViewer {
 
     // Fetch a single page so we can get a viewport that will be the default
     // viewport for all pages
-    return firstPagePromise.then((pdfPage) => {
+    firstPagePromise.then((pdfPage) => {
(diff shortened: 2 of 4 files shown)


web/pdf_find_controller.js
```diff
@@ -251,6 +251,10 @@ class PDFFindController {
           // Store the pageContent as a string.
           this.pageContents[i] = strBuf.join('');
           extractTextCapability.resolve(i);
+        }, (reason) => {
```
Snuffleupagus: In most places in the patch you're using `}).catch((reason) => {`, but not here[1].
So just to satisfy my own curiosity: Is there a particular reason for using this format here?

---
[1] I'm aware that there's a difference in behaviour between the two formats, since `}, (reason) => {` will only catch errors from the original promise whereas `}).catch((reason) => {` will *also* catch errors from the `then` above

web/pdf_viewer.js
```diff
@@ -387,6 +387,11 @@ class PDFViewer {
             if (--getPagesLeft === 0) {
               pagesCapability.resolve();
             }
+          }, (reason) => {
```
Snuffleupagus: The question in https://github.com/mozilla/pdf.js/pull/8746/files#diff-485f8990604b45b959a20e90265d8044R254 applies here as well.

web/pdf_find_controller.js
yurydelendik: Not exact science. Mostly I'm doing it by observation of then-handler: 1) if there is a logic that will cause exception or return a promise -- `.catch()` will be used; 2) however if logic have some simple finalization code thus no reason for exception, I'm trying to fit into then-catch-handler, it will avoid to perform additional micro task spawning.

brendandahl:
Can we add console.error logs to all the places where we consume the rejections/exceptions? Previously, you'd at least see an unhandled promise rejection error to know something went wrong.

yurydelendik:
> Can we add console.error logs to all the places where we consume the rejections/exceptions? 

Added few console.error

Issue #5704, opened 2015-02-06
Ø (LATIN CAPITAL LETTER O WITH STROKE) doesn't appear


mikemorr:
In this PDF: http://www.jazzbooks.com/mm5/download/FREE-nomenclature.pdf
the character Ø (U+00D8, LATIN CAPITAL LETTER O WITH STROKE) doesn't appear. Issue occurs on both Windows and Mac. Screenshots from http://mozilla.github.io/pdf.js/web/viewer.html and Adobe Reader:
![pdfjs](https://cloud.githubusercontent.com/assets/2184268/6082595/97430140-adf0-11e4-962d-311fa3bed203.png)
![adobereader](https://cloud.githubusercontent.com/assets/2184268/6082597/9aaf0ab8-adf0-11e4-9ca9-cf04827587dc.png)

Snuffleupagus:
This is a regression from PR #4259.

PR #8747, opened 2017-08-04
Fix two cmap related issues.

### LINKED ISSUE #8707: Spaces are render as dots
Link to PDF file (or attach file here): [07_2017_managery (21).pdf](https://github.com/mozilla/pdf.js/files/1177525/07_2017_managery.21.pdf)

Configuration:
- Web browser and its version: Chrome 59.0.3071.115
- Operating system and its version: macOS Sierra v10.2.5
- PDF.js version: 1.8.579
- Is an extension: No

1. Generate the PDF

The PDF should render normally with all the spaces.
screen shot 2017-07-26 at 19 30 47

Every spaces are replaced by a dot.

screen shot 2017-07-26 at 18 11 34

I tested with an old version of PDF.js (`1.7.235`) and it works well. So I guess it's a regression.

brendandahl:
In issue #8707, there's a char code mapped to a non-
existing glyph which shouldn't be drawn. However, we
saw it was missing and tried to then use the post table and
end up mapping it incorrectly.

This illuminated a problem with issue #5704 and bug
893730 where glyphs disappeared after above fix.  This was
from the cmap returning the wrong glyph id. Which in turn was
caused because the font had multiple of the same type of cmap
table and we were choosing the last one. Now, we instead
default to the first one. I'm unsure if we should instead be
merging the multiple cmaps, but using only the first one works.

Fixes #8707


brendandahl:
Fixes #8707

Snuffleupagus (approved):
Looks good, thank you!

mikelweb:
This didn't solve my problem: Spaces are rendered as glyphs (big dots)

brendandahl:
@mikelweb please file a new issue your pdf.

PR #8749, opened 2017-08-04
Version 1.9

yurydelendik:


timvandermeij:
I think it's good to release a new version, both because the current one is from quite some time ago and because I would like to have a stable version before we proceed with the `PDFHistory` refactoring (since it's basically a complete rewrite, it would be good to allow some time to fix regressions, if there are any).

brendandahl:
Do we have a list of api-minor changes?

timvandermeij:
I think this is the list since the latest release on April 13:

[api-minor] Add support for PageMode in the API and viewer (issue 8657)
[api-minor] Adds additional parameter so background color of canvas can be set
[api-minor] Fix svg image render
[api-minor] Always allow e.g. rendering to continue even if there are errors, and add a `stopAtErrors` parameter to `getDocument` to opt-out of this behaviour (issue 6342, issue 3795, bug 1130815)

*(from https://github.com/mozilla/pdf.js/pulls?utf8=%E2%9C%93&q=is%3Apr%20%5Bapi-minor%5D)*

PR #8756, opened 2017-08-07
Remove leftover `args[0].code` checks from the `OPS.paintXObject` cases in evaluator.js

Snuffleupagus:
From looking at blame, it seems that these checks became obsolete with PR #692 (which landed close to six years ago). Note how, after that PR, there's no longer anything being assigned to the `code` property of an Object.


timvandermeij:
Nice find!

*(The failed regression tests is a known intermittent one.)*

PR #8761, opened 2017-08-08
Fallback to plain object for globalScope.

tbtlr:
When loading PDF.js in non-browser,non-node,strict-mode environments (e.g. loading within a jsm module), globalScope is set to undefined. Should fallback to a plain object instead.

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -24,7 +24,8 @@ if ((typeof PDFJSDev === 'undefined' ||
 var globalScope =
   (typeof window !== 'undefined' && window.Math === Math) ? window :
   (typeof global !== 'undefined' && global.Math === Math) ? global :
-  (typeof self !== 'undefined' && self.Math === Math) ? self : this;
+  (typeof self !== 'undefined' && self.Math === Math) ? self :
+  (typeof this !== 'undefined' && this.Math === Math) ? this : {};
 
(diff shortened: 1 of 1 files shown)


src/shared/compatibility.js
```diff
@@ -24,7 +24,8 @@ if ((typeof PDFJSDev === 'undefined' ||
 var globalScope =
   (typeof window !== 'undefined' && window.Math === Math) ? window :
   (typeof global !== 'undefined' && global.Math === Math) ? global :
-  (typeof self !== 'undefined' && self.Math === Math) ? self : this;
+  (typeof self !== 'undefined' && self.Math === Math) ? self :
+  (typeof this !== 'undefined' && this.Math === Math) ? {};
```
Snuffleupagus: This fails linting, see https://travis-ci.org/mozilla/pdf.js/builds/262068404, since this line seems to contain a SyntaxError. I'm assuming that what you actually want here is `(typeof this !== 'undefined' && this.Math === Math) ? this : {};` instead :-)

Please note that you can easily run linting locally (using `gulp lint`) to catch these kind of issues, and please remember to squash the commit when fixing this.

Snuffleupagus (changes requested):

tobytailor:
Thats what happens if you use the GitHub web editor to make a change :). Fix follow soon.

Snuffleupagus (approved):
Thanks for the patch!

PR #8762, opened 2017-08-08
Replace the `coded` property with `isType3Font` when building the font `properties` object in `PartialEvaluator.translateFont`

Snuffleupagus:
This appears to simply have been forgotten in the re-factoring in PR #4815, where the `coded` property was renamed to the *much* more descriptive `isType3Font` property.


timvandermeij (approved):

timvandermeij:
Good find, thanks!

PR #8768, opened 2017-08-10
Adds fetch stream logic for networking part of PDF.js

### LINKED ISSUE #6126: Implement Fetch API for streaming PDF loads on Chrome
Chrome has now shipped the Fetch API ( https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/35_QSL1ABTY ) in Chrome 43. I'm not sure yet if other browser vendors will implement it, but seeing as PDF.js has support for the firefox-only moz-chunk-arraybuffer, it seems reasonable to implement this for Chrome. We often see poor network performance loading PDFs, especially when using range requests, and hopefully this can help.

I'm planning to start working on this over the next few days, any initial thoughts?

mukulmishra18:
This PR adds [Fetch](https://fetch.spec.whatwg.org/) support in networking part of PDF.js project. This PR implements [IPDFStream](https://github.com/mozilla/pdf.js/blob/21cc2c020026341586024fb6a9d41602bddc3ed5/src/core/worker.js#L62) interfaces using Fetch API to stream PDF data. Using Fetch can give us streamable response, that streams PDF data in small chunks.

P.S. fixes #6126

### DIFF
--- a/src/display/fetch_stream.js
+++ b/src/display/fetch_stream.js
@@ -1,0 +1,222 @@
+/* Copyright 2012 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { assert, createPromiseCapability } from '../shared/util';
+import {
+  createResponseStatusError, validateRangeRequestCapabilities,
+  validateResponseStatus
+} from './network_utils';
+
+function createFetchOptions(headers, withCredentials) {
+  return {
+    method: 'GET',
+    headers,
+    mode: 'cors',
+    credentials: withCredentials ? 'omit' : 'include',
+    redirect: 'follow',
+  };
+}
+
+class PDFFetchStream {
+  constructor(options) {
+    this.options = options;
+    this.source = options.source;
+    this.isHttp = /^https?:/i.test(this.source.url);
+    this.httpHeaders = (this.isHttp && this.source.httpHeaders) || {};
+
+    this._fullRequestReader = null;
+    this._rangeRequestReaders = [];
+  }
+
+  getFullReader() {
+    assert(!this._fullRequestReader);
+    this._fullRequestReader = new PDFFetchStreamReader(this);
+    return this._fullRequestReader;
+  }
+
+  getRangeReader(begin, end) {
+    let reader = new PDFFetchStreamRangeReader(this, begin, end);
+    this._rangeRequestReaders.push(reader);
+    return reader;
+  }
+
+  cancelAllRequests(reason) {
+    if (this._fullRequestReader) {
+      this._fullRequestReader.cancel(reason);
+    }
+    let readers = this._rangeRequestReaders.slice(0);
+    readers.forEach(function(reader) {
+      reader.cancel(reason);
+    });
+  }
+}
+
+class PDFFetchStreamReader {
+  constructor(stream) {
+    this._stream = stream;
+    this._reader = null;
+    this._loaded = 0;
+    this._withCredentials = stream.source.withCredentials;
+    this._contentLength = this._stream.source.length;
+    this._headersCapability = createPromiseCapability();
+    this._disableRange = this._stream.options.disableRange;
+    this._rangeChunkSize = this._stream.source.rangeChunkSize;
+    if (!this._rangeChunkSize && !this._disableRange) {
+      this._disableRange = true;
+    }
+
+    this._isRangeSupported = !this._stream.options.disableRange;
+    this._isStreamingSupported = !this._stream.source.disableStream;
+
+    this._headers = new Headers();
+    for (let property in this._stream.httpHeaders) {
+      let value = this._stream.httpHeaders[property];
+      if (typeof value === 'undefined') {
+        continue;
+      }
+      this._headers.append(property, value);
+    }
+
+    let url = this._stream.source.url;
+    fetch(url, createFetchOptions(this._headers, this._withCredentials)).
+        then((response) => {
+      if (!validateResponseStatus(response.status, this._stream.isHttp)) {
+        throw createResponseStatusError(response.status, url);
+      }
+      this._headersCapability.resolve();
+      this._reader = response.body.getReader();
+
+      let { allowRangeRequests, suggestedLength, } =
+        validateRangeRequestCapabilities({
+          getResponseHeader: (name) => {
+            return response.headers.get(name);
+          },
+          isHttp: this._stream.isHttp,
+          rangeChunkSize: this._rangeChunkSize,
+          disableRange: this._disableRange,
+        });
+
+      this._contentLength = suggestedLength;
+      this._isRangeSupported = allowRangeRequests;
+    }).catch(this._headersCapability.reject);
+
+    this.onProgress = null;
+  }
+
+  get headersReady() {
+    return this._headersCapability.promise;
+  }
+
+  get contentLength() {
+    return this._contentLength;
+  }
+
+  get isRangeSupported() {
+    return this._isRangeSupported;
+  }
+
+  get isStreamingSupported() {
+    return this._isStreamingSupported;
+  }
+
+  read() {
+    return this._headersCapability.promise.then(() => {
+      return this._reader.read().then(({ value, done, }) => {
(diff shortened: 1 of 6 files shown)


timvandermeij:
Does this fix #6126?

mukulmishra18:
> Does this fix #6126?

Yes, this patch is intended to add support for fetch stream. Currently it is using `fetch` if browser has native implementation, although I am thinking if we can add [github polyfill](https://github.com/github/fetch) for fetch, if browser doesn't have support.

While testing this patch I get to know that we don't have any property named `body` in [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) Object in firefox(but defined in chrome), so https://github.com/mozilla/pdf.js/pull/8768/commits/8370f1defc514fdc7ac6afaece5d4ddff9ca2de9#diff-ffa21364c93bc3328ab1eb76226f8a71R84 might throw error. I think we have to detect it and use `response.arrayBuffer()`(?) instead.

mukulmishra18:
While testing this PR locally, I find out that `this._headers.get(name)` is not returning value for `Accept-Ranges` http response header from https://github.com/mozilla/pdf.js/pull/8768/commits/63b42869ad0ebf2f227e80f22de9e34da2826021#diff-ffa21364c93bc3328ab1eb76226f8a71R87.  Although it is present in the response header(confirmed from the network tab in web console).

src/pdf.js
```diff
@@ -34,7 +34,10 @@ if (typeof PDFJSDev === 'undefined' ||
   if (pdfjsSharedUtil.isNodeJS()) {
     var PDFNodeStream = require('./display/node_stream.js').PDFNodeStream;
     pdfjsDisplayAPI.setPDFNetworkStreamClass(PDFNodeStream);
-  } else {
+  } else if (typeof Response !== 'undefined' && 'body' in Response.prototype) {
+    var PDFFetchStream = require('./display/fetch_stream.js').PDFFetchStream;
+    pdfjsDisplayAPI.setPDFNetworkStreamClass(PDFFetchStream);
+   } else {
```
yurydelendik: @Rob--W do you know if we can enable *only* fetch_stream.js for Chrome extension?

yurydelendik:
@mukulmishra18 Can you rebase the commit to the current master?

src/pdf.js
Rob--W: What do you mean?

1. Whether `fetch_stream` is disabled for every target except for the Chrome extension build, or
2. Whether `fetch_stream` can be unconditionally be used in the Chrome extension, and `display/network` be removed?

I can't answer 1, but if you meant 2:

- The streaming feature requires Chrome 43 - https://github.com/mozilla/pdf.js/issues/6126#issuecomment-130572300
- I looked at [the telemetry of the Chrome extension](https://github.com/Rob--W/pdfjs-telemetry) for August 2017 (this month), filtered by users with the latest version of the extension (1.8.557, released July 2017, 17th), with a user agent that contains Chrome/digits.

Based on 1.4M unique data points (i.e. pings where the client cannot be distinguished from another client in the same data set; this already happens when the user upgrades their browser version), the usage statistics are as follows:

```
Chrome 62 0.91%
Chrome 61 0.87%
Chrome 60 56.85% <-- Current stable (officially released July 2017, 25th)
Chrome 59 32.11%
Chrome 58 2.08%
Chrome 57 1.17%
Chrome 56 0.77%
Chrome 55 0.97%
Chrome 54 0.20%
Chrome 53 0.34%
Chrome 52 0.60%
Chrome 51 0.33%
Chrome 50 0.19%
Chrome 49 0.91% <-- last supported Chrome for Windows XP
Chrome 48 0.07%
Chrome 47 0.17%
Chrome 46 0.08%
Chrome 45 0.07%
Chrome 44 0.04%
Chrome 43 0.13% <-- First version where streams API is supported
Chrome 42 0.04%
Chrome 41 0.06%
Chrome 40 0.05%
Chrome 39 0.15%
Chrome 38 0.01%
Chrome 37 0.02%
Chrome 36 0.01%
Chrome 35 0.38% <-- mostly 35.0.1878.0, win7. Doesn't look like an official build.
Chrome 34 0.37% <-- Ubuntu https://apps.ubuntu.com/cat/applications/chromium-browser/
Chrome 33 0.02%
Chrome 32 0.01%
Chrome 31 0.01%
Chrome 100 0.01% <-- I don't know what this is...
```

1.13% of the users (16k pings) use a Chrome version that does not support the streams API (Chrome 42 and earlier). For now, I'd like to keep supporting these users.

yurydelendik: Thanks for response and analysis. I was asking about (2) -- fetch_stream can be unconditionally be used in the Chrome extension. You answered my question, so we will handle Chrome extension no different than generic viewer.

yurydelendik:
We are trying to "serialize" the reason we are not suppose too. Add makeReasonSerializable:

```
index 9bd0ff1b..15b3f8ff 100644
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -1247,6 +1247,17 @@ function wrapReason(reason) {
   }
 }
 
+function makeReasonSerializable(reason) {
+  if (!(reason instanceof Error) ||
+      reason instanceof AbortException ||
+      reason instanceof MissingPDFException ||
+      reason instanceof UnexpectedResponseException ||
+      reason instanceof UnknownErrorException) {
+    return reason;
+  }
+  return new UnknownErrorException(reason.message, reason.toString());
+}
+
 function resolveOrReject(capability, success, reason) {
   if (success) {
     capability.resolve();
@@ -1307,16 +1318,12 @@ function MessageHandler(sourceName, targetName, comObj) {
             data: result,
           });
         }, (reason) => {
-          if (reason instanceof Error) {
-            // Serialize error to avoid "DataCloneError"
-            reason = reason + '';
-          }
           comObj.postMessage({
             sourceName,
             targetName,
             isReply: true,
             callbackId: data.callbackId,
-            error: reason,
+            error: makeReasonSerializable(reason),
           });
         });
       } else if (data.streamId) {
```

src/display/fetch_stream.js
```diff
@@ -0,0 +1,211 @@
...
+    let url = this._stream.source.url;
+    fetch(url, {
+      method: 'GET',
+      headers: this._headers,
+      mode: 'cors',
+      credentials: this._withCredentials ? 'omit' : 'include',
+      redirect: 'follow',
+    }).then((response) => {
+      if (response.status !== 200) {
+        throw validateResponseStatus(response.status, url);
```
yurydelendik: Change name to createResponseStatusError.

src/display/fetch_stream.js
```diff
@@ -0,0 +1,211 @@
...
+
+    let url = this._stream.source.url;
+    fetch(url, {
+      method: 'GET',
+      headers: this._headers,
+      mode: 'cors',
+      credentials: this._withCredentials ? 'omit' : 'include',
+      redirect: 'follow',
+    }).then((response) => {
+      if (response.status !== 200) {
```
yurydelendik: you need to validate status based on protocol, for http it is 200 for other protocols it's 0. I recommend to add/have/change validateResponseStatus(status, isHttp) so it will return boolean that will tell is status is valid. Use this function in the range reader as well.

mukulmishra18: Done with new commit. Thanks.

src/display/fetch_stream.js
mukulmishra18: Done. Thanks.

src/display/network_utils.js
```diff
@@ -52,6 +53,24 @@ function validateRangeRequestCapabilities({ getResponseHeader, isHttp,
...
+    'Unexpected server response (' + status +
+    ') while retrieving PDF "' + url + '".', status);
+}
+
+function validateResponseStatus(status, isHttp) {
+  if (status === 404 || status === 0 && !isHttp) {
+    return false;
+  }
+  return true;
+}
```
yurydelendik:
```
if (!isHttp) {
  return status === 0;
}
return status === 200 || status === 206;
```

src/display/fetch_stream.js
```diff
@@ -0,0 +1,219 @@
...
+      this._headers.append(property, value);
+    }
+
+    let rangeStr = begin + '-' + (end - 1);
+    this._headers.append('Range', 'bytes=' + rangeStr);
+    let url = this._stream.source.url;
+    fetch(url, {
+      method: 'GET',
+      headers: this._headers,
+    }).then((response) => {
```
yurydelendik: define the same options as in full reader. extract the `createFetchOptions(headers)` method?

yurydelendik (approved):
Looks good, with the validate and options changes.

src/display/network_utils.js
mukulmishra18: Done. Thanks.

src/display/fetch_stream.js
mukulmishra18: created `createFetchOptions` with new commit. Thanks.

yurydelendik:
filed https://bugzilla.mozilla.org/show_bug.cgi?id=1395220

yurydelendik:
Thank you for the patch.

PR #6272, opened 2015-07-26
Re-implement `PDFHistory`

Snuffleupagus:
_Please refer to the individual commit messages._

@timvandermeij I know that this is a large PR, but would you be willing to try and review it?
I'd suggest that you completely ignore the combined diff, and instead look at each patch on its own.

Also, we should make sure this is landed at beginning of a development-cycle ([see future dates](https://wiki.mozilla.org/RapidRelease/Calendar#Future_branch_dates)), to allow more time for follow-ups.

Issue #5927, opened 2015-04-10
Viewer doesn't remember last rotation angle


xlc:
The viewer does remember last scroll position but not rotation angle. After reload the viewer will restore the last scroll position but because the rotation angle is not restored, the page position may be wrong.

mbbaig:
I'd like to get started working on this. Any suggestions on the devs want to go about doing this?

timvandermeij:
I would think that we can do this the same way as we store the last scroll position. I think the relevant part is located here: https://github.com/mozilla/pdf.js/blob/e4e9a2adea1b19f962dcebaca7c8144104549173/web/viewer.js#L897

TZanke:
In relation to #8743 , will this be merged in master / new version anytime?

soubhikchatterjee:
The rotation works fine in the web view, but the original file remains unchanged in the disk. Is there anyway we can save the changes to the original file on the disk?

PR #8775, opened 2017-08-13
Re-write `PDFHistory` from scratch

Snuffleupagus:

Supersedes PRs #6272 and #8753; hopefully third time's the charm :-)

*Edit:* Based on this PR, we should also be able to finally address issue #5927 properly; see WIP [patches available here](https://github.com/Snuffleupagus/pdf.js/compare/Snuffleupagus:7613cfad7ed63dcb513ef55bc84f14a9cf22d678...Snuffleupagus:83a50ff7c2f22b3a09b9d5fb3734c5f85cb453a1).

*Edit2:* Rebased to fix merge conflicts, no code changes were made.


web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+    if (!this._destination.page) {
+      // `this._destination` was set through the user changing the hash of
+      // the document. Do not add `this._position` to the browser history,
+      // to avoid "flooding" it with lots of (nearly) identical entries,
+      // since we cannot ensure that the document position has changed.
+      return;
+    }
+
+    let forceReplace = false;
+    if (this._destination.page === position.first ||
```
timvandermeij: What is this `first` entry exactly? Is there a more descriptive name for it, or could you comment it?

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+    let state = window.history.state;
+
+    this.initialized = true;
+    this.initialBookmark = null;
+
+    this._popStateInProgress = false;
+    this._blockHashChange = 0;
+    this._currentHash = getCurrentHash();
+
+    this._currentUid = this._uid = 0;
```
timvandermeij: Could you describe what this represents or what it's used for?

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+    }
+    if (state.fingerprint !== this.fingerprint) {
+      // This should only occur in viewers with support for opening more than
+      // one PDF document, e.g. the GENERIC viewer.
+      return false;
+    }
+    if (!Number.isInteger(state.uid) || state.uid < 0) {
+      return false;
+    }
+    if (typeof state.destination !== 'object' || state.destination === null) {
```
timvandermeij: Similar to the other comment, let's swap these conditions for readability.

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+   */
+  _updateViewarea({ location, }) {
+    this._position = {
+      hash: this._isViewerInPresentationMode ?
+        `page=${location.pageNumber}` : location.pdfOpenParams.substring(1),
+      page: this.linkService.page,
+      first: location.pageNumber,
+    };
+
+    if (this._popStateInProgress) {
```
timvandermeij: This appears to have no effect since the method ends here anyway. Should this be above setting `this._position` instead?

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+    this._popStateInProgress = true;
+
+    if (hashChanged) {
+      // When the hash changed, implying that the 'popstate' event will be
+      // followed by a 'hashchange' event, then we do *not* want to update the
+      // browser history when handling the 'hashchange' event (in web/app.js)
+      // since that would *overwrite* the new destination navigated to below.
+      //
+      // To avoid accidentally disabling all future user-initiated hash changes,
+      // if there's e.g. another 'hashchange' listeners that stops the event
```
timvandermeij: Nit: listener

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+      // propagation, we make sure to always force-reset `this._blockHashChange`
+      // after `HASH_CHANGE_TIMEOUT` milliseconds has passed.
```
timvandermeij: Nit: have passed

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+      this._blockHashChange++;
+      waitOnEventOrTimeout({
+        target: window,
+        name: 'hashchange',
+        delay: HASH_CHANGE_TIMEOUT,
+      }).then((type) => {
```
timvandermeij: This is now an unused variable. Usually this is denoted by replacing `type` with `_` (indicating we don't care about what it is).

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+   * @private
+   */
+  _bindEvents() {
+    let { _boundEvents, eventBus, } = this;
+
+    _boundEvents.updateViewarea = this._updateViewarea.bind(this);
+    _boundEvents.popState = this._popState.bind(this);
+    _boundEvents.pageHide = (evt) => {
+      // Attempt to push the `this._position` into the browser history when
+      // navigating away from the document, if the history is currently empty.
```
timvandermeij: Why only if the history is empty? If there are history items, I change the position and navigate away from the document, I would expect the last position to be retained when I navigate back.

web/pdf_history.js
```diff
@@ -354,6 +375,30 @@ class PDFHistory {
     if (this._popStateInProgress) {
       return;
     }
+
+    if (UPDATE_VIEWAREA_TIMEOUT > 0) {
+      // When closing the browser, a 'pagehide' event will be dispatched which
+      // *should* allow us to push the current position to the browser history.
+      // In practice, it seems that the event is arriving to late in order for
```
timvandermeij: Nit: too late

web/pdf_history.js
```diff
@@ -405,14 +450,22 @@ class PDFHistory {
...
     // This case corresponds to navigation backwards in the browser history.
     if (state.uid < this._currentUid && this._position && this._destination) {
-      if (this._destination.page &&
-          this._destination.page !== this._position.first &&
-          this._destination.page !== this._position.page) {
+      let shouldGoBack = false;
+
+      if (this._destination.temporary) {
+        // Always replace a previous *temporary* position, before moving back.
+        this._pushOrReplaceState(this._position);
```
timvandermeij: Should we force-replace here, since it *always* needs to happen?

web/pdf_history.js
```diff
@@ -63,12 +65,16 @@ class PDFHistory {
 
     this._boundEvents = Object.create(null);
     this._isViewerInPresentationMode = false;
+    this._isPagesLoaded = false;
 
-    // Ensure that we don't miss a 'presentationmodechanged' event, by
-    // registering the listener immediately.
+    // Ensure that we don't miss either a 'presentationmodechanged' or
```
timvandermeij: Nit: or -> or a

web/pdf_history.js
```diff
@@ -376,6 +386,19 @@ class PDFHistory {
...
+    if (POSITION_UPDATED_THRESHOLD > 0 && this._isPagesLoaded &&
+        this._destination && !this._destination.page) {
+      // If the current destination was set through the user changing the hash
+      // of the document, we will usually not try to push the current position
+      // to the browser history; see `this._tryPushCurrentPosition()`.
+      //
+      // To prevent `this._tryPushCurrentPosition()` from effectively being
+      // reduced to a no-op in this case, we will assume that the position
+      // *did* in fact change if the 'updateviewarea' event was dispatched
+      // more than `POSITION_UPDATED_THRESHOLD` times.
```
timvandermeij: I'm not entirely sure what this means or what this commit aims to achieve. Why is it bad if `_tryPushCurrentPosition` is reduced to a no-op if the hash is invalid, and where does the heuristic come from? Could you elaborate a bit on what problem this commit solves and how it works (approximately)?

timvandermeij (changes requested):
Phew, that was quite a bit of code to review :)

In general these patches look really good! As is to be expected with such a large pull request, I have some questions and suggestions for improvement, but most of them are minor. After these comments have been adressed, I'll do more testing (initial testing did not yield any unexpected behavior yet) and then we can ship it. With approximately one month to go before a new Firefox release, this should give us enough time in case we need to tweak things.

test/unit/ui_utils_spec.js
```diff
@@ -259,4 +260,119 @@ describe('ui_utils', function() {
...
+      });
+      // Do *not* dispatch the event, and wait for the timeout.
+
+      buttonClicked.then(function(type) {
+        expect(type).toEqual(WaitOnType.TIMEOUT);
+        done();
+      }, done.fail);
+    });
+
+
```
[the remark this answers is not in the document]
Snuffleupagus: Sure; also we might want to change the ESLint rule `no-multiple-empty-lines` to only allow *one* empty to avoid these issues.

web/pdf_history.js
```diff
@@ -30,6 +30,43 @@ class PDFHistory {
   forward() {}
 }
 
+function isDestsEqual(firstDest, secondDest) {
+  function isEntryEqual(first, second) {
+    if (typeof first !== typeof second) {
+      return false;
+    }
+    if (first instanceof Array) {
```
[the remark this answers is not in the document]
Snuffleupagus: I figured that the `typeof` checks just above should handle that, but considering that `typeof [] === 'object'` you're obviously correct. There should be a check for `second` too, really good call!

web/pdf_history.js
```diff
@@ -30,6 +30,43 @@ class PDFHistory {
...
+      return false;
+    }
+    if (typeof first === 'object' && first !== null) {
```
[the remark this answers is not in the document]
Snuffleupagus: > Can we swap these conditions?

Yes, there's no particular reason for the ordering, so if you find that more readable that can be done.

> Don't we also need to check these conditions for second?

Oh, we should probably check at least `second !== null` here too. That should be sufficient, sicne the initial `typeof first !== typeof second` ought to handle the rest.

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+    this._currentHash = getCurrentHash();
+
+    this._currentUid = this._uid = 0;
+    this._destination = null;
+    this._position = null;
+
+    if (!this._isValidState(state) || resetHistory) {
+      let { hash, page, } = parseCurrentHash(this.linkService);
+
+      if (!hash || reInitialized || resetHistory) {
```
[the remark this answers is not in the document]
Snuffleupagus: > but should we also take this path if !page, or is that valid?

If you enforced `!page` here, then you'd break the initialization for documents loading with a named destination in the hash; e.g. http://ftp.acc.umu.se/mirror/CTAN/info/lshort/english/lshort.pdf#chapter.1

That should definitely be valid, but note how you cannot extract a `page` from the hash (but the `page` will be correctly set once `PDFLinkService.navigateTo` runs). This is in contrast to the kind of hash that you get when clicking on the viewBookmark button, e.g. http://ftp.acc.umu.se/mirror/CTAN/info/lshort/english/lshort.pdf#page=10&zoom=auto,-106,760

I hope that this helps highlight why *this* particular code looks the way it is (there's unfortunately a lot of edge-cases to consider in this code).

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+      // Ensure that the browser history is initialized correctly when
+      // the document hash is present on PDF document load.
+      this._pushOrReplaceState({ hash, page, }, /* forceReplace = */ true);
+      return;
+    }
+
+    // The browser history contains a valid entry, ensure that the history is
+    // initialized correctly on PDF document load.
+    let destination = state.destination;
+    this._resetSomeInternalState(destination, state.uid);
```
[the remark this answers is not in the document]
Snuffleupagus: Sure, if you think that's more descriptive. (I went back-and-forth a bit with that particular name.)

web/pdf_history.js
Snuffleupagus: Nope, it should *not* be moved, but it will become relevant in the very next commit.

web/pdf_history.js
Snuffleupagus: I'll just remove it completely, i.e. `}).then(() => {` instead.

web/pdf_history.js
Snuffleupagus: It's the *first* visible page (in the viewer), as provided by the `updateviewarea` event. Please refer to https://github.com/mozilla/pdf.js/pull/8775/commits/1f41a595995baedfd9265ad93e226ba0a5a89c0c#diff-3b4ebbe223f93c9123a754b4845844e7R347.
 
If you look a the code in `PDFViewer`, see https://github.com/mozilla/pdf.js/blob/f54dfc63dcb116c1f97577a786372bf4c2fd0ef3/web/pdf_viewer.js#L704 you'll notice that the `pageNumber` parameter there corresponds to the *first* page.

Since, in the viewer, the first visible page often differs from the *most* visible page (i.e. the one shown in the pageNumber input). Compare what's displayed in the pageNumber input with the page number used by the viewBookmark button as you scroll around in the viewer.

I want to prevent issues with multiple (almost) identical history entries being added to the browser history, by only updating it when it's definitely clear that page we're now on is *not* the same as the one currently in the browser history.

Does this help? Based on the above, how do we make this situation clearer in code?

web/pdf_history.js
Snuffleupagus: By the time the `pagehide` event fires, it's too late to make (most) modifications of the history. That is also the whole reason for the existence of commit https://github.com/mozilla/pdf.js/pull/8775/commits/281070b20c3ed17979b85482922d9245c8eee623, refer to the big comment starting at https://github.com/mozilla/pdf.js/pull/8775/commits/281070b20c3ed17979b85482922d9245c8eee623#diff-3b4ebbe223f93c9123a754b4845844e7R380.

At this point in time, since the viewer is already being unloaded, it's impossible to push a *new* entry to the browser history. What would happen instead here, is that the current browser history entry would be overwritten (which is definitely *not* what you want here).
The only case where that doesn't matter is if the history is empty, and for all other cases we'll now rely on the temporary position instead here.

I'll try to amend this comment a bit, to make this clearer.

web/pdf_history.js
Snuffleupagus: Doing so would break things *very* badly, since it would overwrite the *current* history entry.

Since we get where when the user goes back in the history, the current browser history entry (`window.history.state`) now points to the new destination we should be navigating to.
By force-replacing in this case, you'd get stuck in the *current* position in the document, with all sorts of breakage (history non-functional as so on).

This is quite complicated to wrap your head around at first (don't ask me how I know :-), so if you'd like a can try and explain in a bit more detail!

web/pdf_history.js
Snuffleupagus: Consider loading the following document http://ftp.acc.umu.se/mirror/CTAN/info/lshort/english/lshort.pdf#pagemode=thumbs
In this case, the hash is defined, but it will obviously not cause a destination to be navigated to (which would set the `page` of its browser history entry). Furthermore, `parseCurrentHash` can obviously not extract a `page` either; https://github.com/mozilla/pdf.js/pull/8775#discussion_r135468381 might be somewhat relevant here too.

Given https://github.com/mozilla/pdf.js/pull/8775/commits/1f41a595995baedfd9265ad93e226ba0a5a89c0c#diff-3b4ebbe223f93c9123a754b4845844e7R288, `tryPushCurrentPosition` won't ever update the browser history with the current position (unless `push` is called). This could thus mean that despite commit https://github.com/mozilla/pdf.js/pull/8775/commits/281070b20c3ed17979b85482922d9245c8eee623, the browser history might not accurately capture the current position when the viewer is navigated away from or the browser closed.

Hence this heuristic, that attempts to "unlock" the adding of entries to the browser history when the position has updated enough times that it should be safe to assume that the position has in fact changed.
You could argue that this is a bit magical, especially since the value of the constant was basically pulled out of a hat, but it's my attempt to avoid permanently disabling certain parts of the implementation in e.g. the edge-case outlined above.
Note that in most cases, this heuristic won't even be necessary, but I figured that it shouldn't hurt to try and avoid this situation.

Does this help, if not ask again!

web/pdf_history.js
Snuffleupagus: It's used for two things:
First of all, to be able to tell if we're moving backwards of forwards in the history, during `_popState` event handler; see https://github.com/mozilla/pdf.js/pull/8775/commits/1f41a595995baedfd9265ad93e226ba0a5a89c0c#diff-3b4ebbe223f93c9123a754b4845844e7R407.

Second of all, it's used in the `back` and `forward` methods, to navigate through the history without accidentally leaving the viewer/PDF document. With the way that the history is implemented (in browsers), there's no way to access anything except the *current* entry (which I assume is for security reasons). Hence if we just unconditionally just called `window.history.{back, forward}`, we could easily navigate away from the document.

Since I cannot image that it'd be a good idea if the [`GoBack`/`GoForward`](https://github.com/mozilla/pdf.js/blob/f54dfc63dcb116c1f97577a786372bf4c2fd0ef3/web/pdf_link_service.js#L290-L300) named actions could navigate away from the document, those parameters help avoid that.

Is this clearer, if not please ask!

web/pdf_history.js
```diff
@@ -13,21 +13,455 @@
...
+ * @property {Array} explicitDest - The explicit destination array.
+ * @property {number} pageNumber - The page to which the destination points.
+*/
+
+function getCurrentHash() {
+  return document.location.hash;
+}
+
+function parseCurrentHash(linkService) {
+  let hash = unescape(document.location.hash).substring(1);
```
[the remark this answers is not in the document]
Snuffleupagus: Sure, if you think that's better.

Snuffleupagus:
> Phew, that was quite a bit of code to review :)

Thank you for having the persistence, despite the size[1], to wade through all of it!!!

> As is to be expected with such a large pull request, I have some questions and suggestions for improvement, but most of them are minor.

There's a number of good questions asked and I've tried my best to answer them[2], so please check if you're satisfied with the responses or if you think further clarifications are necessary!

In general, did you find that the existing comments helped, or were they to verbose?

> I'll do more testing (initial testing did not yield any unexpected behavior yet)

That's very reassuring to hear, since the whole point of this re-write was to make the code more robust (and of course more maintainable too).

Again, thanks for taking the time to do such a thorough review :-) 

---
[1] I really tried to break this up into somewhat smaller commits, but unfortunately that didn't work as well as I would have liked :-(

[2] It's sometimes quite difficult to know what parts will be confusing or difficult to understand for someone else, considering that you've got a lot of context that the reviewer might not have.

timvandermeij:
Thank you for the clear responses! In general the comments really helped to understand the code, in particular the edge cases you're addressing, so I'm really glad that it's documented so well! You're right that having the context is key here, so a review is a good way to find out which parts may need further clarification.

The responses look good so far. For some of them I'll need some more time to compare them with the code and to really understand, which I'll do this week including the testing, so we can hopefully merge it at the end of the week if all goes well. Feel free to amend the commits in the meantime if you want; I'll check each comment above in the amended version after that.

timvandermeij (approved):

timvandermeij:
Nice work! Looks really good and it's a massive improvement over the old implementation.

Snuffleupagus:
> Nice work! Looks really good and it's a massive improvement over the old implementation.

Thanks so much for helping getting this reviewed (with lots of good points/questions) and landed! :-D

Snuffleupagus:
> Nice work! Looks really good and it's a massive improvement over the old implementation.

Thanks so much for helping with getting this reviewed (with lots of good points/questions) and landed! :-D

PR #8778, opened 2017-08-14
Use `Uint8ClampedArray`, when returning data, and remove manual clamping in `src/core/jpg.js` (issue 4901)

### LINKED ISSUE #4901: Reduce or improve calls to clamp0to255 in jpg js
The following picture shows a profiling from IE10.
slow-pdfjs-ie
it shows that clamping is quite frequent. And because it is IE, it is super slow. 
http://jsperf.com/clamp-functions/12
confirms that IE is way behind all modern browsers.

The question I am asking myself:
From the perspective of the algorithm, is the clamping actually needed?
and if so, is it needed in both directions?
Would a | 0 truncation be sufficient?

Maybe @p01 or @bthorben can give me a hint or help me out here.

Snuffleupagus:
This patch removes the `clamp0to255` helper function, as well as manual clamping code in `src/core/jpg.js`.
The adjusted constants in `_convertCmykToRgb` were taken from CMYK to RGB conversion code found in `src/core/colorspace.js`.

*Please note:* There will be some very slight movement in a number of existing test-cases, since `Uint8ClampedArray` appears to use `Math.round` (or equivalent) and the old code used (basically) `Math.floor`.

Fixes #4901.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "babel-loader": "^6.4.0",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.23.0",
     "babel-preset-es2015": "^6.24.1",
+    "core-js": "^2.5.0",
     "escodegen": "^1.8.0",
     "eslint": "^4.2.0",
     "eslint-plugin-mozilla": "^0.4.0",
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -600,10 +600,6 @@ var JpegImage = (function JpegImageClosure() {
     return component.blockData;
   }
 
-  function clamp0to255(a) {
-    return a <= 0 ? 0 : a >= 255 ? 255 : a;
-  }
-
   function findNextFileMarker(data, currentPos, startPos) {
     function peekUint16(pos) {
       return (data[pos] << 8) | data[pos + 1];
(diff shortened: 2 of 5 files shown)


yurydelendik:
I wonder if we shall further clean up our polyfills in util.js and compatibility.js and also remove from global space (like we did with streams). So for typed arrays we can add typedarray_polyfill.js and re-export it in util.js. In files that are using Uint8ClampedArray we will start using imported from util.js constructor.

Snuffleupagus:
In this PR, I figured that it'd make sense to only make the minimal amount of changes required to support `Uint8ClampedArray`.

> I wonder if we shall further clean up our polyfills in util.js and compatibility.js and also remove from global space (like we did with streams). So for typed arrays we can add typedarray_polyfill.js and re-export it in util.js. In files that are using Uint8ClampedArray we will start using imported from util.js (or directly from typedarray_polyfill.js) constructor.

If we want to change this, could that perhaps be done in a follow-up[1] instead to avoid too much scope creep here?

---
[1] It would probably even qualify for the `5-good-beginner-bug` label.

yurydelendik:
> could that perhaps be done in a follow-up

Fair enough. Make sure that this polyfill will not appear in mozilla-central build. Other than that, it looks good

Snuffleupagus:
> Make sure that this polyfill will not appear in mozilla-central build.

Locally, I get the results as below (all sizes in bytes) for the relevant builds:

|     | `gulp firefox` | `gulp mozcentral` |
| --- | --- | --- |
| **master** | 6 353 174 | 3 431 002 |
| **patch** | 6 352 666 | 3 430 577 |

timvandermeij (approved):

timvandermeij:
Really nice!

PR #8781, opened 2017-08-15
__non_webpack_require__ -> require in SystemJS

Rob--W:
When running browser tests, e.g. via `gulp unittest`, the test files are not processed by babel, and neither by the "unittestcli" gulp target.

This commit copies the babelPluginReplaceNonWebPackRequire plugin from the unittestcli target to the SystemJS config so that `__non_webpack_require__` is replaced with `require` for all build targets, and adds a unit test to ensure that this indeed works as expected.

### DIFF
--- a/systemjs.config.js
+++ b/systemjs.config.js
@@ -39,6 +39,23 @@
                           typeof crypto !== 'undefined' &&
                           typeof crypto.subtle !== 'undefined';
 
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good. Thank you for the patch!

PR #8783, opened 2017-08-16
Use `Uint8ClampedArray` when returning image data in `src/core/jbig2.js` and `src/core/jpg.js`

Snuffleupagus:
*The following was done in a separate commit, to make it easy to revert just in case it's not a desired change:* Remove manual clamping code in `src/core/jpx.js`

Since we're now using `Uint8ClampedArray`, rather than `Uint8Array`, doing manual clamping shouldn't be necessary given that that is now handled natively.

This shouldn't have any measurable performance impact, but just to sanity check that I've done some quick benchmarking with the following manifest file:
```json
[
    {  "id": "S2-eq",
       "file": "pdfs/S2.pdf",
       "md5": "d0b6137846df6e0fe058f234a87fb588",
       "rounds": 100,
       "type": "eq"
    }
]
```
which gave the following results against the current `master` (repeated benchmark runs didn't result in any meaningful differences):
```
firefox | Overall      |   100 |          592 |         592 |   1 |  0.12 |
firefox | Page Request |   100 |            3 |           3 |   0 | -9.88 |
firefox | Rendering    |   100 |          588 |         589 |   1 |  0.18 |
```

### DIFF
--- a/src/core/image.js
+++ b/src/core/image.js
@@ -253,7 +253,7 @@ var PDFImage = (function PDFImageClosure() {
     // in this thread can be relying on its contents.
     if (inverseDecode) {
       for (i = 0; i < actualLength; i++) {
-        data[i] = ~data[i];
+        data[i] ^= 0xFF;
       }
     }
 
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good. Thank you!

PR #7023, opened 2016-02-24
Only draw glyphs on canvas if they are in the font or the font file is missing.

brendandahl:
Fixes #6721 

I could have sworn there were more PDF's with this issue, but I'm not finding any.

Issue #6721, opened 2015-12-03
à character rendered where it should not be rendered


timvandermeij:
The PDF file at https://www.dropbox.com/s/ygmic028oykprkw/respond_document_print.pdf?dl=1 has random `à` characters rendered where they should not be.

Snuffleupagus:
Interestingly enough, even though Adobe Reader doesn't render the `à` character, it's still present on copy-and-paste.

mjstyle:
Thanks guys!

mjstyle:
Hi guys, just wanted to check in to see if you guys were able to find a fix to this issue.

Thank you so much!

Regards,

timvandermeij:
Not yet, for these kinds of issues finding the cause of the problem is unfortunately not trivial, so this might take some time.

mjstyle:
@timvandermeij Happy New Years everyone! Just checking in to see if anyone added a fix for this yet :)

Thanks!

Regards,

mjstyle:
@timvandermeij Just following up to see if there will be a planned fix soon?

Thanks!

timvandermeij:
When a fix becomes available, it will be reported here.

mjstyle:
Hi there,

Just checking in to see if this fix is out on production?

Thanks!

Regards,
Jack

timvandermeij:
The fix has been merged here, so it's most likely out for production in the current or the next Firefox release.

mjstyle:
Hi Tim,

I'm just following up to see if this is out on production yet?

Thanks!

Regards,
Jack

Snuffleupagus:
> [...] if this is out on production yet?

What does "production" actually mean in this case?
This issue was fixed in PR #7023, which is included in the `1.5.188` release of PDF.js; see https://github.com/mozilla/pdf.js/releases/tag/v1.5.188.
The PR was merged to mozilla-central in https://bugzilla.mozilla.org/show_bug.cgi?id=1253498, and the `Target Milestone` field says Firefox 47 (i.e. it's included in that version). As can be seen in https://wiki.mozilla.org/RapidRelease/Calendar#Past_branch_dates, the current Firefox release is version 48.

mjstyle:
Understood thank you!

PR #5863, opened 2015-03-19
Hook the SVG backend into the viewer and implement testing

timvandermeij:
**Heavily WIP. Do not merge!**

The idea of this patch is to explore working with two backends (canvas and SVG) and to run the tests for the SVG backend. It is based on @yurydelendik's work in #5159. Please feel free to provide suggestions such that we can come up with a solution that can actually be implemented. The goal of this PR is to explore where the difficulties are and to try to overcome them. A follow-up patch can then provide a pretty solution.

Current problems:
- [ ] SVG to PNG does not keep font data. A possible solution is to render the SVG on a canvas and save that.
- [ ] We can do with less code edits (e.g., text test removal), but for now this was the fastest. The test driver needs to be refactored to make certain code paths depend on `PDFJS.displayBackend` to solve this. There are no text tests for the SVG backend as there is no text layer, but SVG text instead.

PR #8785, opened 2017-08-16
SVG: Don't render missing glyphs

Rob--W:
This bug is similar to the canvas bug of #6721, which was fixed by #7023

See commit description for more info and test cases.

New reduced test case introduced with this PR:
- File: https://github.com/mozilla/pdf.js/files/1229507/pcdatainvalidchar.pdf
- Expected rendered result: "hardware performance"
- Actual SVG source (can be generated with `node examples/node/pdf2svg.js pcdatainvalidchar.pdf`): "hardware\x03performance"

(I did not add an automated test because we don't have any ref test framework for SVG (e.g. painting SVG on a canvas, and then doing an image comparison, as done in #5863 - somehow that issue was closed without follow-up...).

The test case from #6721 can be used here too:
- File: https://github.com/mozilla/pdf.js/files/52205/issue6721_reduced.pdf
- Expected: "Issue   6721"
- Actual rendered result (before this patch): "Issue ààà6721"

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -742,15 +742,23 @@ SVGGraphics = (function SVGGraphicsClosure() {
           x += -glyph * fontSize * 0.001;
           continue;
         }
-        current.xcoords.push(current.x + x * textHScale);
 
         var width = glyph.width;
         var character = glyph.fontChar;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #8788, opened 2017-08-17
Attempt to delay disabling of the attachment view until FileAttachment annotations of the *initial* page has been parsed

Snuffleupagus:
As discussed in PR https://github.com/mozilla/pdf.js/pull/8673#issuecomment-317241023, we cannot solve the general issue (since that would require parsing every single page). However, we can mitigate the effect somewhat, by waiting for the FileAttachment annotations of the initially rendered page.


timvandermeij (approved):

timvandermeij:
Works just fine, thank you!

PR #8789, opened 2017-08-17
Convert `ColorSpace` and `PDFImage` to use `Uint8ClampedArray`s as output from its various methods

Snuffleupagus:

This allows us to remove manual clamping/rounding code, and instead rely on `Uint8ClampedArray` handling that natively.

*Please note:* Because of different rounding behaviour, basically `Math.round` in `Uint8ClampedArray` respectively `Math.floor` in the old code, there will be very slight movement in quite a few existing test-cases. However, the changes should be imperceivable to the naked eye, given that the absolute difference is *at most* `1` for each RGB component when comparing `master` and this patch (see also the updated expectation values in the various unit-tests).

---
Marking as "work-in-progress" for now, since I would very much appreciate feedback on the general approach of the PR!

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -254,6 +254,7 @@ class Annotation {
 
   /**
    * Set the color and take care of color space conversion.
+   * The default value is black, in RGB color space.
    *
    * @public
    * @memberof Annotation
@@ -262,7 +263,7 @@ class Annotation {
    *                        4 (CMYK) elements
    */
   setColor(color) {
-    let rgbColor = new Uint8Array(3); // Black in RGB color space (default)
+    let rgbColor = new Uint8ClampedArray(3);
     if (!Array.isArray(color)) {
       this.color = rgbColor;
       return;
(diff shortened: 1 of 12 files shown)


timvandermeij:
This looks really good to me. I like how no manual clamping is needed anymore, not only cleaning up the code but also letting the browser solve it natively (which is most likely faster as other benchmarks have shown). I agree with not enforcing clamped arrays for the general case, but only when a caller sets it.

Snuffleupagus:
@timvandermeij Since you seemed generally positive to these changes, according to https://github.com/mozilla/pdf.js/pull/8789#issuecomment-323523793, is this something that'd be willing (time permitting of course) to review if I rebased/updated the patches?

The most time-consuming part of the review would probably be checking all the test "failures", to ensure that no actual regressions occurred as a result of the changes.

timvandermeij:
No problem! Feel free to ping me once the updated version is pushed.

PR #7793, opened 2016-11-08
Rename the `ViewHistory` localStorage (and sessionStorage) key from `database` to `pdfjs.history`, and migrate existing data on read (issue 7760)

Snuffleupagus:
For consistency, I also renamed the `FIREFOX/MOZCENTRAL` sessionStorage key, but given that sessionStorage is a lot less permanent than localStorage it didn't seem necessary to migrate any existing values.

Fixes #7760.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/pdf.js/7793)

PR #8794, opened 2017-08-19
Remove the migration code for old localStorage data in `ViewHistory`

Snuffleupagus:
This was added in PR #7793, which has now been present in a number of PDF.js releases (from version `v1.7.225`, see https://github.com/mozilla/pdf.js/releases/tag/v1.7.225).
Hence we should be able to remove it now, considering that the migration code was only intended as a best effort solution to avoid wiping out all existing user data at once. Also, keep in mind that `ViewHistory` is already limited with regards to the number of documents it will simultaneous store data for.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #8796, opened 2017-08-19
Implement text rise for the SVG back-end

timvandermeij:
The property and the setter for text rise were already present, but they were never used or called. This patch completes the implementation by calling the setter when the operator is encountered and by using the text rise value when rendering text.

To verify the current behavior:
1. Open https://mozilla.github.io/pdf.js/web/viewer.html
2. Run `PDFViewerApplication.preferences.set('renderer', 'svg');` in the console and refresh.
3. Download https://www.pdflib.com/fileadmin/pdflib/Cookbook/pdf/footnotes_in_text.pdf and open it using the Open File button in the viewer. Notice that the `1` and `2` in the footnotes are not in superscript and that the console contains a message about the unimplemented operator `setTextRise`.

To verify the new behavior:
1. Open the preview viewer.
2. Run `PDFViewerApplication.preferences.set('renderer', 'svg');` in the console and refresh.
3. Open the `footnotes_in_text.pdf` fule using the Open File button in the viewer. Notice that the `1` and `2` in the footnotes are in superscript and that the console message is gone.

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -562,6 +562,9 @@ SVGGraphics = (function SVGGraphicsClosure() {
             this.setTextMatrix(args[0], args[1], args[2],
                                args[3], args[4], args[5]);
             break;
+          case OPS.setTextRise:
+            this.setTextRise(args[0]);
+            break;
           case OPS.setLineWidth:
             this.setLineWidth(args[0]);
             break;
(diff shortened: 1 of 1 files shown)


timvandermeij:
@Rob--W Since you're also working on the SVG-backend, would you perhaps have time to review this?

/botio-windows test

src/display/svg.js
```diff
@@ -782,9 +785,16 @@ SVGGraphics = (function SVGGraphicsClosure() {
         current.tspan.setAttributeNS(null, 'fill', current.fillColor);
       }
 
+      // Include the text rise in the matrix since the `pm` function creates
+      // the SVG element's `translate` entry. Work on a copy to not alter the
+      // original text matrix.
+      let matrix = current.textMatrix.slice();
+      if (current.textRise !== 0) {
+        matrix[5] += current.textRise;
```
Rob--W: How often is this called? If `setTextRise` is called quite often, then you could relieve the GC by only making a copy when it is needed:

```javascript
let matrix = current.textMatrix;
if (current.textRise) {
    matrix = matrix.slice();
    matrix[5] += current.textRise;
}
```

timvandermeij: Good point! Fixed in the new commit.

Rob--W:
Why is the unit test in Chrome failing on Windows?

```
TEST-UNEXPECTED-FAIL | gets destinations, from /Names (NameTree) dictionary | Error: close should have stream controller thrown 
```

timvandermeij:
This is a known intermittent after the stream API changes landed, but somehow I forgot to file it before. I just did that in #8816. It's not related to this patch.

timvandermeij:
The intermittent unit test failure is fixed, so it should be fine now.

Rob--W:
Thanks for the patch. I wish that we also had reftesting for the SVG backend.

PR #8800, opened 2017-08-21
Try to recover if we reach the end of the stream when searching for the `EI` marker of an inline image (issue 8798)

### LINKED ISSUE #8798: Some pdf contents are pictures,and header and footer could not be show
the problem pdf is,use pdf tool show is fine
[output_34.pdf](https://github.com/mozilla/pdf.js/files/1241137/output_34.pdf)
i use the pdf.js example demo code show result like thise
image

Configuration:
- Web browser and its version: chrome 60.0.3112.101 
- Operating system and its version:win7
- PDF.js version:pdfjs-1.8.188-dist
- Is an extension:

1. both open the problem pdf use pdfjs example and other pdf tools,compare the diffrents
2. under the search，i found that the problem pdf contents are pictures,and thier had been combined
show the header and footer

couldn't show header and footer

Snuffleupagus:
Fixes #8798.

### DIFF
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -147,10 +147,9 @@ var Parser = (function ParserClosure() {
      * Find the end of the stream by searching for the /EI\s/.
      * @returns {number} The inline stream length.
      */
-    findDefaultInlineStreamEnd:
-        function Parser_findDefaultInlineStreamEnd(stream) {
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #8801, opened 2017-08-21
Add a set of jpeg markers to be ignored

luizhartke:
There are some jpeg representations that can contain some dummy markers and they will be ignored


timvandermeij:
Do you have a PDF file that fails to render because of this that you could share here? It's hard to verify the fix without having an example file.

Where to these markers come from? Are they documented somewhere, i.e., in a specification? Right know they're kind of magic numbers to me.

Please revert the `.gitignore` and `pdfjs.config` changes; they're not related to the patch.

Finally, the lines are too long. Format it as follows:

```javascript
let dummyMarkersSet = new Set([
  '0', '1', '2', ..., '101', '102', (until you reach 80 characters in width)
  '203', '277', ...
]);
```

timvandermeij:
Closing for now since it's unknown what this fixes exactly and we cannot merge it in the current state. This can be reopened once the comments are adressed and the tests pass. Thanks.

PR #7440, opened 2016-06-25
 Prevent destinations with bad left/top values from scrolling the wrong page into view (bug 874482)

Snuffleupagus:
There are PDF generators which create destinations with e.g. too large top values, which cause the wrong page to be scrolled into view because the offset becomes negative.
By ignoring negative offsets, we can prevent this issue, and get a similar behaviour as in Adobe Reader.

However, since we're also using `PDFViewer_scrollPageIntoView` in more cases than just when links (in the document/outline) are clicked, the patch adds a way to allow the caller to opt-out of this behaviour.
In e.g. the following situations, I think that we still want to be able to allow negative offsets: when restoring a position from the `ViewHistory`, when the `viewBookmark` button is used to obtain a link to the current position, or when maintaining the current position on zooming.

Rather than adding another parameter to `PDFViewer_scrollPageIntoView`, I've changed the signature to take an parameter object instead. To maintain backwards compatibility, I've added fallback code enclosed in a `GENERIC` preprocessor tag.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=874482.

**Note:** This PR includes the patch from PR #6903  which is enough to fixes the bug, but with the addition of the second patch we also avoid the issues described in that PR.

/cc @timvandermeij

PR #8807, opened 2017-08-22
Remove handling of fallback arguments from `PDFViewer.scrollPageIntoView`

Snuffleupagus:
The method signature was improved in PR #7440, which has now been present in a number of releases (starting with `v1.6.210`).
Hence we should be able to remove this now, and just print an error message if the old format is used.


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #8808, opened 2017-08-22
Fix color of image masks inside uncolored patterns

### LINKED ISSUE #3226: Colour in pdf not rendered
Fig. 1 in http://arxiv.org/pdf/0705.4298.pdf is not displayed properly in PDF Viewer 0.8.1

### LINKED ISSUE #8741: Dotted areas rendered as black.
http://soap.plansystem.dk/pdfarchive/20_1057347_APPROVED_1193694109382.pdf

Configuration:
        FireFox 54.0.1 (64-bit) [Does not work]
        Chrome Version 59.0.3071.115 (Official Build) (64-bit) [Works fine]
        macOS Sierra 10.12.6 (16G29) 
        (Problem is also reproducible on Windows computers)
- PDF.js version:
        1.8.593
- Is an extension:
We are seeing this problem when opening the document directly in FireFox 54 as well as on Chrome when using PDF.JS (both version 1.8.254 and 1.8.593) in our Angular 2 project.
Opening the document directly in Chrome, using the build in viewer, renderes the PDF correctly.

1. Open the pdf referenced above in FireFox, use internal viewer (PDF.JS)
2. Zoom in on the grey areas show that they are in fact dotted.

Images from Chrome 
image
image

image
image

### LINKED ISSUE #8860: Translucent effect has wrong color
Link to PDF file (or attach file here): https://assets.cdn.sap.com/sapcom/docs/2016/05/827c26ba-717c-0010-82c7-eda71af511fa.pdf

Configuration:
- Web browser and its version: Firefox 55
- Operating system and its version: Linux
- PDF.js version: (as included with Firefox 55)
- Is an extension: no

1. open file
2. look at first slide

What is the expected behavior? (screenshot from Evince)
evince

What went wrong? (screenshot of PDF.js/Firefox)
pdf.js

janpe2:
Fixes #8741 
Fixes #3226
Fixes #8860

This is a patch for uncolored (`/PaintType 2`) tiling patterns that contain image masks. When uncolored patterns are used as a fill or stroke, the operator `scn` or `SCN` specifies the color of the pattern. If the pattern contains an image mask, PDF.js incorrectly draws it using black instead of the color specified by `scn` or `SCN`. 

In my patch, I assign the proper color to the `fillColor` property of the `CanvasExtraState` object. That property is used as the color of image masks in functions `paintImageMaskXObject...()` in canvas.js.

### DIFF
--- a/src/display/pattern_helper.js
+++ b/src/display/pattern_helper.js
@@ -360,7 +360,7 @@ var TilingPattern = (function TilingPatternClosure() {
       var graphics = canvasGraphicsFactory.createCanvasGraphics(tmpCtx);
       graphics.groupLevel = owner.groupLevel;
 
-      this.setFillAndStrokeStyleToContext(tmpCtx, paintType, color);
+      this.setFillAndStrokeStyleToContext(graphics, paintType, color);
 
       this.setScale(width, height, xstep, ystep);
       this.transformToScale(graphics);
@@ -401,17 +401,23 @@ var TilingPattern = (function TilingPatternClosure() {
     },
 
     setFillAndStrokeStyleToContext:
-      function setFillAndStrokeStyleToContext(context, paintType, color) {
+      function setFillAndStrokeStyleToContext(graphics, paintType, color) {
+        let context = graphics.ctx, current = graphics.current;
         switch (paintType) {
           case PaintType.COLORED:
             var ctx = this.ctx;
             context.fillStyle = ctx.fillStyle;
             context.strokeStyle = ctx.strokeStyle;
+            current.fillColor = ctx.fillStyle;
+            current.strokeColor = ctx.strokeStyle;
             break;
           case PaintType.UNCOLORED:
             var cssColor = Util.makeCssRgb(color[0], color[1], color[2]);
             context.fillStyle = cssColor;
             context.strokeStyle = cssColor;
+            // Set color needed by image masks (fixes issues 3226 and 8741).
+            current.fillColor = cssColor;
+            current.strokeColor = cssColor;
             break;
           default:
             throw new FormatError(`Unsupported paint type: ${paintType}`);
--- a/test/pdfs/pr8808.pdf.link
+++ b/test/pdfs/pr8808.pdf.link
@@ -1,0 +1,1 @@
+http://soap.plansystem.dk/pdfarchive/20_1057347_APPROVED_1193694109382.pdf
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -698,6 +698,14 @@
        "lastPage": 1,
        "type": "load"
     },
+    {  "id": "pr8808",
+       "file": "pdfs/pr8808.pdf",
+       "md5": "bdac6051a98fd8dcfc5344b05fed06f4",
+       "rounds": 1,
+       "link": true,
+       "lastPage": 1,
+       "type": "eq"
+    },
     {  "id": "issue5599",
        "file": "pdfs/issue5599.pdf",
        "md5": "529a4a9409ac024aeb57a047210280fe",


src/display/pattern_helper.js
```diff
@@ -360,7 +360,7 @@ var TilingPattern = (function TilingPatternClosure() {
       var graphics = canvasGraphicsFactory.createCanvasGraphics(tmpCtx);
       graphics.groupLevel = owner.groupLevel;
 
-      this.setFillAndStrokeStyleToContext(tmpCtx, paintType, color);
+      this.setFillAndStrokeStyleToContext(tmpCtx, paintType, color, graphics);
```
yurydelendik: let's change signature to setFillAndStrokeStyleToGraphics(graphic, paintType, color) -- `context` is accessible via graphics.ctx.

src/display/pattern_helper.js
```diff
@@ -412,6 +413,8 @@ var TilingPattern = (function TilingPatternClosure() {
             var cssColor = Util.makeCssRgb(color[0], color[1], color[2]);
             context.fillStyle = cssColor;
             context.strokeStyle = cssColor;
+            // Set color needed by image masks (fixes issues 3226 and 8741).
+            graphics.current.fillColor = cssColor;
```
yurydelendik: do we need to do that for `this.current.strokeColor =`? do we need to do the same for PaintType.COLORED case?

janpe2:
> do we need to do that for this.current.strokeColor =?

I don't know if any PDF would need this but I added it, just to be sure.

> do we need to do the same for PaintType.COLORED case?

I added lines similar to the other case. Colored tiling patterns are a bit different: now the `color` argument in `setFillAndStrokeStyleToContext()` is meaningless (or null) because the PDF operators scn and SCN have no color operands.

Snuffleupagus:
> Fixes #8741, #3226.

Would it be possible to include the PDF file from either of those issues as a *linked* test-case, to prevent future regressions?

janpe2:
I added the test case.

By the way, my patch seems to fix also issue #8860.

Snuffleupagus:
> I added the test case.

Thank you; one question though, wouldn't it be enough to just test the first page though?
If so, please add `"lastPage": 1,` to entry in `test_manifest.json`

Snuffleupagus:
@janpe2 Looking at the test logs above, it appears that the server prevented the PDF file from being downloaded:
```
Downloading http://arxiv.org/pdf/0705.4298.pdf to pdfs/issue3226.pdf...
Error during downloading of http://arxiv.org/pdf/0705.4298.pdf: HTTP 403
WARNING: File was not downloaded. See "pdfs/issue3226.pdf.error" file.
Unable to verify the checksum for the files that are used for testing.
Please re-download the files, or adjust the MD5 checksum in the manifest for the files listed above.
```

Would it be possible to link to the PDF file through e.g. http://archive.org/web/ instead, or do we need to use another PDF file for the test-case?

janpe2:
What are those failed regression tests? Should I be worried about them?

timvandermeij:
The problem is that the old file is not replaced. @yurydelendik Could you remove the old `issue3226.pdf` file from both bots so we can try again?

Snuffleupagus:
> The problem is that the old file is not replaced. @yurydelendik Could you remove the old (cached) issue3226.pdf file from both bots so we can try again?

I don't think that's really the issue at play here, since the (Internet Archive) link does *not* actually point to a PDF file.
If you open https://web.archive.org/web/20170830175223/http://arxiv.org/pdf/0705.4298.pdf directly in a browser, you'll get the same thing that the test-suite gets when downloading the file: a HTML page which references a PDF file.

It seems to me that linking to this particular PDF file probably will not work, and that the easiest solution here would be to try and use a PDF file from one of the other referenced issues as a test-case instead.

Snuffleupagus (approved):

PR #8810, opened 2017-08-22
Cache PdfJs's enabled setting for delayed init

squarewave:
See Bug 1389443 for more information. This allows us to register
PdfJs without waiting for file IO in nsHandlerService to finish.
Once that file IO is finished, we can set everything up properly
and double-check the registration.


brendandahl:
Looks good. I hope some day we can simplify all this enabled code!

PR #8812, opened 2017-08-22
Moves global scope out of shared/util.

yurydelendik:
Follow up on #8761

Using `this` in modules is tricky. (It will be replaced by babel to `undefined`). Moving globalScope out of shared/utils into separate file -- this will allow to create single object for compatibility and util usages.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1040,7 +1040,7 @@ gulp.task('lib', ['buildnumber'], function () {
   var buildLib = merge([
     gulp.src([
       'src/{core,display}/*.js',
-      'src/shared/{compatibility,util,streams_polyfill}.js',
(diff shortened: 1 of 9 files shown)


Snuffleupagus:
Strangely enough, when testing this locally using `gulp server`, I see the following printed in the console
```
Warning: Setting up fake worker.   util.js:282:11
ReferenceError: require is not defined   compatibility.js:24:5
```

yurydelendik:
Not all compatibility.js references where removed  in https://github.com/mozilla/pdf.js/pull/8778/files#diff-a72ef320c3225c98b2163b89a3e8ff62L48

Snuffleupagus (approved):

PR #8813, opened 2017-08-22
Use Array.isArray in the LoopbackPort.

yurydelendik:
For better serialization of Arrays that belong to different sandboxes.

See also http://logs.glob.uno/?c=mozilla%23pdfjs#c66373

/cc @tobytailor

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -16,7 +16,7 @@
 
 import {
   assert, createPromiseCapability, deprecated, getVerbosityLevel,
-  info, InvalidPDFException, isArray, isArrayBuffer, isInt, isSameOrigin,
+  info, InvalidPDFException, isArrayBuffer, isInt, isSameOrigin,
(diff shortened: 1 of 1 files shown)


timvandermeij:
Indeed, wouldn't just changing https://github.com/mozilla/pdf.js/blob/master/src/shared/util.js#L1097 suffice here? For the polyfill we can use `core-js` since we already use it for clamped arrays.

yurydelendik:
> Can we use Array.isArray everywhere?

Maybe. There are too many places to check if they will work as expected when changed. For now I'm just addressing regression (and hope will not add another one :).

yurydelendik:
> For IE9, a polyfill can be added to compatibility.js.

Per MDN, Array.isArray is implemented in IE9

Rob--W:
Even better. Then we don't need a polyfill at all.

Snuffleupagus (approved):
This seems fine, but it now needs a rebase :-)

r=me, with passing tests.

PR #8815, opened 2017-08-23
Confirm `eval` available before compiling PostScript. Fixes #5573

### LINKED ISSUE #5573: PostScriptCompiler compiled result violates the Content security policy (no unsafe-eval)
https://github.com/mozilla/pdf.js/blob/45ca95370513c3c5edfefa802e6e6152bec56a84/src/core/function.js#L406

is using  return new Function('src', 'srcOffset', 'dest', 'destOffset', compiled);

This does not work when using a CSP for security reasons which does not allow unsafe-eval for script-src.

PDFJS.version = '1.0.907';

TheKidCoder:
Compiling PS functions is only available if eval is available. Some sites might have `eval` disable via CSP. This includes the eval check.
This fixes: https://github.com/mozilla/pdf.js/issues/5573

### DIFF
--- a/src/core/function.js
+++ b/src/core/function.js
@@ -399,15 +399,17 @@ var PDFFunction = (function PDFFunctionClosure() {
       var range = IR[2];
       var code = IR[3];
 
-      var compiled = (new PostScriptCompiler()).compile(code, domain, range);
-      if (compiled) {
+      if (!!PDFJS.isEvalSupported) {
         // Compiled function consists of simple expressions such as addition,
         // subtraction, Math.max, and also contains 'var' and 'return'
         // statements. See the generation in the PostScriptCompiler below.
         // eslint-disable-next-line no-new-func
-        return new Function('src', 'srcOffset', 'dest', 'destOffset', compiled);
+        var compiled = new PostScriptCompiler().compile(code, domain, range);
+        if (compiled) {
+          return new Function('src', 'srcOffset', 'dest', 'destOffset', compiled);
+        }
       }
-
+  
       info('Unable to compile PS function');
 
       var numOutputs = range.length >> 1;


timvandermeij:
Doesn't this break the functionality then? In other words, is there no way to avoid that by e.g., replacing `eval` in some way?

src/core/function.js
```diff
@@ -399,15 +399,17 @@ var PDFFunction = (function PDFFunctionClosure() {
...
         // statements. See the generation in the PostScriptCompiler below.
         // eslint-disable-next-line no-new-func
-        return new Function('src', 'srcOffset', 'dest', 'destOffset', compiled);
+        var compiled = new PostScriptCompiler().compile(code, domain, range);
+        if (compiled) {
+          return new Function('src', 'srcOffset', 'dest', 'destOffset', compiled);
+        }
       }
-
+
```
timvandermeij: Remove this whitespace.

src/core/function.js
```diff
@@ -399,15 +399,17 @@ var PDFFunction = (function PDFFunctionClosure() {
       var range = IR[2];
       var code = IR[3];
 
-      var compiled = (new PostScriptCompiler()).compile(code, domain, range);
-      if (compiled) {
+      if (!!PDFJS.isEvalSupported) {
```
timvandermeij: Travis CI has some complains about this line; could you check them?

timvandermeij (changes requested):

TheKidCoder:
I don't fully understand what the benefit of the compiler is, but from some related comments I've come across it seems to only be there to improve speed. In the PDFs that were broken and now work, there is no visual difference. I'll fix the travis errors.

timvandermeij:
If that is indeed the case, then I'm fine with the change. It's a deliberate choice to disable `eval`, so users can expect less performance if they choose to do so.

@yurydelendik Could you chime in on if it's indeed only a performance loss and not a functionality loss?

yurydelendik:
It can be disabled and this will be a performance loss for some PDFs. There are some PDFs that use PS functions for colorspaces, that will make then slower -- based on #5134 4-5 times. But I think it's okay to disable the compilation when isEvalSupported = false -- not a problem for demo/generic viewer.

timvandermeij:
Thank you! In that case, this patch looks good with the comments addressed.

src/core/function.js
yurydelendik: PDFJS.isEvalSupported does not exist -- we removed all globals from PDFJS. This setting needs to be passed some other way.

Snuffleupagus: I'm assuming that we may also want to do something similar to https://github.com/mozilla/pdf.js/blob/23a4174163cba5709f05b04b371ee67b1d0148ca/src/display/font_loader.js#L401 and https://github.com/mozilla/pdf.js/blob/23a4174163cba5709f05b04b371ee67b1d0148ca/src/display/font_loader.js#L332 here as well?

timvandermeij:
Closing since this is superseded by #8909.

PR #8817, opened 2017-08-24
Set this.isCancelled in close method of streamSink.

### LINKED ISSUE #8816: Unit tests intermittently raise "Error: close should have stream controller thrown"
This happened a few times now after the stream API changes landed. The latest one is https://github.com/mozilla/pdf.js/pull/8796#issuecomment-323880702. 

/cc @mukulmishra18: You mentioned a possible fix in http://logs.glob.uno/?c=mozilla%23pdfjs&s=28+Jul+2017&e=28+Jul+2017&h=should+have+stream+controller+thrown#c65580. Could you verify if that is indeed the problem and create a PR if so?

mukulmishra18:
Fixes #8816 . This PR tries to fix the intermittent error `Error: close should have stream controller thrown` raised during unit tests.

I cannot see where we are setting `sink.isCancelled = true`, when close() method of streamSink is called. So, I think adding `this.isCancelled = true;` in close() method itself, may fix this intermittent error.

@timvandermeij can you please check if this fixes the problem, as I cannot reproduce this error locally.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -1483,6 +1483,7 @@ MessageHandler.prototype = {
         if (this.isCancelled) {
           return;
         }
+        this.isCancelled = true;
         sendStreamRequest({ stream: 'close', });
         delete self.streamSinks[streamId];
       },


timvandermeij:
Looks good. Thank you!

PR #8824, opened 2017-08-24
Prevent an infinite loop in `XRef.readXRef` by keeping track of already parsed tables (bug 1393476)

Snuffleupagus:
*Please note that while this passes all tests locally, I don't know if it's necessarily the best solution.*

With this patch, not only is the infinite loop prevented, but we're also able to actually render the file (which e.g. Adobe Reader isn't able to).

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1393476.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -1186,10 +1186,21 @@ var XRef = (function XRefClosure() {
 
     readXRef: function XRef_readXRef(recoveryMode) {
       var stream = this.stream;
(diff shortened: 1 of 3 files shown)


yurydelendik (approved):
Looks good, thank you!

timvandermeij:
I like the simplicity of the patch and the fact that we can actually render this. Awesome!

Issue #5089, opened 2014-07-25
Feature Request: Support hyperlinks in Document Outline


nibblesnbits:
Some PDFs I'm displaying with the viewer contain items in the document outline that, when rendered in Adobe Reader (haven't checked others), are actually hyperlinks.  Would it be possible to support this in pdf.js?  I'm looking at `readDocumentOutline` in `obj.js` right now and found what will hopefully lead me in the right direction, but if I'm off track let me know and I'll see if I can make the changes myself.

kvk123456:
is there any fix for this issue in next available release?

timvandermeij:
(Related to #3214)

kvk123456:
Since 3214 also closed can someone please answer me 

Any resolution for this? can you please let us know the hyperlink issue in bookmark will be fixed in next release?

Snuffleupagus:
I'm attaching the PDF file from issue #3214 here, as an additional example: [bookmark-unsupported-features.pdf](https://github.com/mozilla/pdf.js/files/69030/bookmark-unsupported-features.pdf).

timvandermeij:
@kvk123456 As far as I know nobody is working on this. If someone would like to fix this, feel free to submit a pull request. We are just closing the other issues because they are duplicates of this one.

kvk123456:
Thanks timvandermeij, here is the link I reported in 6788.

https://fbutools-stage.rightmix.com/Viewer.aspx?file=%2fViewer%2fHosted%2fGetDocument.ashx%3ffile%3dhttp%3a%2f%2fstage.efinancialdocs.com%2fdocuments%2fDocData%2fFBU2001_DOC_S_201532112645_FBU2001_LOAD.PDF

Rob--W:
FYI, I developed a patch (not published) that recovers a missing outlune from marked content (#6269). It does not work for many documents (because marked content is not that common), so I stopped working on it.

Snuffleupagus:
Here's a quick, sort of hacky, patch that adds basic support for URLs in the outline (I don't know if it's in good enough shape to bother with submitting a PR): https://github.com/Snuffleupagus/pdf.js/commit/c7d3060a614978aea6c2db04a6a6b75547613186.

kvk123456:
when is the next release, whiich will have this fix?

gsampath127:
The code is working when we the open pdf from local system.
When the pdf document is opened as query string to the viewer. The issue still exists.
Please suggest any better solution for this.

[...]

PR #8825, opened 2017-08-25
Account for broken outlines/annotations, where the destination dictionary contains an invalid `/Dest` entry

Snuffleupagus:
According to the specification, see http://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#page=377, a `Dest` entry in an outline item should *not* contain a dictionary.
Unsurprisingly there's PDF generators that completely ignore this, treating is an `A` entry instead.

The patch also adds a little bit more validation code in `Catalog.parseDestDictionary`.

Fixes https://github.com/mozilla/pdf.js/issues/5089#issuecomment-324871440.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -640,20 +640,32 @@ var Catalog = (function CatalogClosure() {
 
     var destDict = params.destDict;
     if (!isDict(destDict)) {
-      warn('Catalog_parseDestDictionary: "destDict" must be a dictionary.');
+      warn('parseDestDictionary: "destDict" must be a dictionary.');
       return;
     }
     var resultObj = params.resultObj;
     if (typeof resultObj !== 'object') {
-      warn('Catalog_parseDestDictionary: "resultObj" must be an object.');
+      warn('parseDestDictionary: "resultObj" must be an object.');
       return;
     }
     var docBaseUrl = params.docBaseUrl || null;
 
     var action = destDict.get('A'), url, dest;
+    if (!isDict(action) && destDict.has('Dest')) {
+      // A /Dest entry should *only* contain a Name or an Array, but some bad
+      // PDF generators ignore that and treat it as an /A entry.
+      action = destDict.get('Dest');
+    }
+
     if (isDict(action)) {
-      var linkType = action.get('S').name;
-      switch (linkType) {
+      let actionType = action.get('S');
+      if (!isName(actionType)) {
+        warn('parseDestDictionary: Invalid type in Action dictionary.');
+        return;
+      }
+      let actionName = actionType.name;
+
+      switch (actionName) {
         case 'URI':
           url = action.get('URI');
           if (isName(url)) {
@@ -748,11 +760,10 @@ var Catalog = (function CatalogClosure() {
           }
           /* falls through */
         default:
-          warn('Catalog_parseDestDictionary: Unrecognized link type "' +
-               linkType + '".');
+          warn(`parseDestDictionary: Unsupported Action type "${actionName}".`);
           break;
       }
-    } else if (destDict.has('Dest')) { // Simple destination link.
+    } else if (destDict.has('Dest')) { // Simple destination.
       dest = destDict.get('Dest');
     }
 
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -689,6 +689,36 @@ describe('annotation', function() {
       expect(data.dest).toEqual([{ num: 17, gen: 0, }, { name: 'XYZ', },
                                  0, 841.89, null]);
     });
+
+    it('should correctly parse a Dest, which violates the specification ' +
+       'by containing a dictionary', function() {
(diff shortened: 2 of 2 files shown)


src/core/obj.js
```diff
@@ -748,12 +760,16 @@ var Catalog = (function CatalogClosure() {
...
-          warn('Catalog_parseDestDictionary: Unrecognized link type "' +
-               linkType + '".');
+          warn(`parseDestDictionary: Unsupported Action type "${actionName}".`);
           break;
       }
-    } else if (destDict.has('Dest')) { // Simple destination link.
-      dest = destDict.get('Dest');
+    } else if (destDict.has('Dest')) { // Simple destination.
+      let simpleDest = destDict.get('Dest');
+      if (isDict(simpleDest)) {
```
Rob--W: How about directly assigning to `dest` instead of a temporary variable? In case `simpleDest` is invalid, you are already returning early, so not using a temporary variable should be fine.

test/unit/annotation_spec.js
```diff
@@ -689,6 +689,34 @@ describe('annotation', function() {
       expect(data.dest).toEqual([{ num: 17, gen: 0, }, { name: 'XYZ', },
                                  0, 841.89, null]);
     });
+
+    it('should correctly parse a Dest, which violates the specification ' +
+       'by containing a dictionary', function() {
```
Rob--W: Can you add a comment to the line where the invalid dictionary is constructed, for those who don't know the PDF spec by heart?

Rob--W:
Looks good to me, can be merged after feedback on the above nits.

test/unit/annotation_spec.js
Snuffleupagus: Sorry, I don't completely follow what sort of comment you're asking for here. Are you simply asking for a reference to the particular section of the PDF spec, or something else?

> [...] where the invalid dictionary is constructed,

*I might be reading too much into this, but just to clarify:*
Please note that there's no way for a `Dest` entry to contain a "valid" dictionary, the mere fact that one is even present makes it invalid according to the spec.

Rob--W: I was thinking of a comment before line 703, something like this (edit as needed):

```javascript
// Dest must be a Name or Array (PDF27000 section x.y.z), but there are PDFs that store a dictionary.
```

src/core/obj.js
Snuffleupagus: This was done because the (first version of the) patch only checked if an `A` entry existed, and not that it was actually a dictionary when doing the "is there a `Dest` entry and if so is it a dictionary" checks above. Hence it possible to end up here with a simple destination containing a dictionary.

I've updated the patch to do the proper checks above, so that these changes could be reverted.

Snuffleupagus:
@Rob--W Thanks for the review! 
For easier reviewing, I'm including an interdiff of the changes below:
```
diff --git a/src/core/obj.js b/src/core/obj.js
index 2e1228de..95924a6f 100644
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -651,7 +651,7 @@ var Catalog = (function CatalogClosure() {
     var docBaseUrl = params.docBaseUrl || null;
 
     var action = destDict.get('A'), url, dest;
-    if (!action && destDict.has('Dest')) {
+    if (!isDict(action) && destDict.has('Dest')) {
       // A /Dest entry should *only* contain a Name or an Array, but some bad
       // PDF generators ignore that and treat it as an /A entry.
       action = destDict.get('Dest');
@@ -764,12 +764,7 @@ var Catalog = (function CatalogClosure() {
           break;
       }
     } else if (destDict.has('Dest')) { // Simple destination.
-      let simpleDest = destDict.get('Dest');
-      if (isDict(simpleDest)) {
-        warn('parseDestDictionary: The Dest entry should not be a dictionary.');
-        return;
-      }
-      dest = simpleDest;
+      dest = destDict.get('Dest');
     }
 
     if (isString(url)) {
diff --git a/test/unit/annotation_spec.js b/test/unit/annotation_spec.js
index 024355a8..faf3293d 100644
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -700,6 +700,8 @@ describe('annotation', function() {
       let annotationDict = new Dict();
       annotationDict.set('Type', Name.get('Annot'));
       annotationDict.set('Subtype', Name.get('Link'));
+      // The /Dest must be a Name or an Array, refer to ISO 32000-1:2008
+      // section 12.3.3, but there a PDF files where it's a dictionary.
       annotationDict.set('Dest', destDict);
 
       let annotationRef = new Ref(798, 0);
```

src/core/obj.js
Snuffleupagus: I updated the patch to do the proper checks above, so that we'll no longer reach this branch in the `destDict.get('Dest') instanceof Dict` case.

Rob--W:
Thanks for the patch and review response!

PR #8826, opened 2017-08-26
Enable the `for-direction` ESLint rule

Snuffleupagus:
See https://eslint.org/docs/rules/for-direction; helps avoid typos that would cause infinite `for` loops.

Also, updates `eslint` and `eslint-plugin-mozilla` to the latest available versions.


timvandermeij (approved):

timvandermeij:
Nice addition!

PR #8828, opened 2017-08-26
Improve the annotation code by converting to ES6 syntax and removing duplicate code

timvandermeij:
The annotation code is not performance-sensitive, so we may convert this code to ES6 syntax (see http://logs.glob.uno/?c=mozilla%23pdfjs&s=23+Aug+2017&e=23+Aug+2017#c66446). Refer to the individual commits for more information about each change.

*Using https://github.com/mozilla/pdf.js/pull/8828/files?w=1 helps reviewing tremendously.*

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,8 @@
     "typogr": "^0.6.6",
     "uglify-js": "^2.6.1",
     "vinyl-fs": "^2.4.4",
-    "webpack": "^2.2.1",
-    "webpack-stream": "^3.2.0",
+    "webpack": "^3.5.5",
+    "webpack-stream": "^4.0.0",
     "wintersmith": "^2.0.0",
     "yargs": "^3.14.0"
   },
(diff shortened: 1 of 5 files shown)


Snuffleupagus (commented):
Really nice diff stats, that's a lot of boilerplate code being removed at once :-D

One thing though: The IRC discussion only seem to explicitly mention `src/display/`, and this PR would be the first case where we're using ES6 classes in `src/core`.
Since the annotations, if present, will be always be parsed during the `getOperatorList` call, I'm not sure if that's a cause for (theoretical) concern?

Although, looking at https://bugzilla.mozilla.org/show_bug.cgi?id=1167472, a number of performance improvements have already landed.
Given that the number of annotations present on any one page is usually not very large, I'd be surprised if these changes made any difference at all (performance wise).

So, tentatively `r=me`; but I'd like @yurydelendik to confirm that he's OK with this before we land it.

yurydelendik (approved):
Really good, thanks.

PR #8829, opened 2017-08-27
Attempt to improve the `EI` detection heuristics, for inline images, in streams containing `NUL` bytes (issue 8823)

### LINKED ISSUE #8823: Flaw in the "EI" detecting strategy for inline images
(#8798 followup)
[PDFJS-8798-test2.pdf](https://github.com/mozilla/pdf.js/files/1249537/PDFJS-8798-test2.pdf)

Configuration:
- Web browser and its version: FF.0.2
- Operating system and its version: W7 64bit
- PDF.js version: latest

1. open viewer
2. load file

clipboard01

clipboard02

The difference to the file of #8798 is that this one has several inline images. The hex 0 I inserted are after the first.

(Not part of this issue: the inline images are rendered blurry, see #8245)

Snuffleupagus:
*Please note that while this passes all tests locally, I don't know if it's necessarily the best solution.*

Since this patch will now treat (some) `NUL` bytes as "ASCII", the number of `followingBytes` checked are thus increased to (hopefully) reduce the risk of introducing new false positives.

Fixes #8823.


timvandermeij (approved):

timvandermeij:
The change looks good and it's good to have a regression test for this now. Thanks!

PR #8830, opened 2017-08-27
adding separators to README.md

ghost:
this PR does basically aim at:
- adding separators to README.md file;

---

**the main goal of such does basically consist of:**
- *beautifying/stylizing the file* 
- *making the file easier simple and more direct to be*:
    - read and understood

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ PDF.js is community-driven and supported by Mozilla Labs. Our goal is to
 create a general-purpose, web standards-based platform for parsing and
 rendering PDFs.
 
+---
+
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Personally I'm not convinced that this necessary (nor that it looks particularity good), considering that the use of different sized heading in the README file already provides separation between the sections.

tiagmoraismorgado:
i respect your decision in case you decide not to merge this PR. however i think this is relevant

timvandermeij:
I agree with @Snuffleupagus here. For me, the headers provide a clear enough distinction between the different sections and I find the extra borders a bit distracting. Closing for this reason, but thank you for creating the PR; it's always good to post an idea and have a discussion.

PR #8831, opened 2017-08-27
Update the `gulp minified` command to use uglify-es

Snuffleupagus:
By updating to uglify-es, rather than uglify-js, the minifier *itself* now supports ES6 code. This means that it's now possible to minify code built with `PDFJS_NEXT = true` set, i.e. with Babel transpilation disabled, which wasn't the case previously.

Note that uglify-es is based on the API of uglify-js v3, which differs from the one that we previously used.
Of particular importance is the fact that it's no longer possible to provide a path to a file for minification, but one must instead directly provide the source of the file.

For more information, please see https://github.com/mishoo/UglifyJS2/tree/harmony

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -702,25 +702,27 @@ gulp.task('minified-pre', ['buildnumber', 'locale'], function () {
 });
 
 gulp.task('minified-post', ['minified-pre'], function () {
-  var viewerFiles = [
-    MINIFIED_DIR + BUILD_DIR + 'pdf.js',
-    MINIFIED_DIR + '/web/viewer.js'
-  ];
+  var pdfFile = fs.readFileSync(MINIFIED_DIR + '/build/pdf.js').toString();
+  var pdfWorkerFile =
+    fs.readFileSync(MINIFIED_DIR + '/build/pdf.worker.js').toString();
+  var viewerFiles = {
+    'pdf.js': pdfFile,
+    'viewer.js': fs.readFileSync(MINIFIED_DIR + '/web/viewer.js').toString(),
+  };
 
   console.log();
   console.log('### Minifying js files');
 
-  var UglifyJS = require('uglify-js');
+  var UglifyES = require('uglify-es');
   // V8 chokes on very long sequences. Works around that.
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -702,25 +702,27 @@ gulp.task('minified-pre', ['buildnumber', 'locale'], function () {
...
+  var viewerFiles = {
+    'pdf.js': pdfFile,
+    'viewer.js': fs.readFileSync(MINIFIED_DIR + '/web/viewer.js').toString(),
+  };
 
   console.log();
   console.log('### Minifying js files');
 
-  var UglifyJS = require('uglify-js');
+  var UglifyJS = require('uglify-es');
```
timvandermeij: I think this should be `UglifyES`. If you want to avoid that, you can also choose `uglifier` as a more generic name.

gulpfile.js
```diff
@@ -702,25 +702,27 @@ gulp.task('minified-pre', ['buildnumber', 'locale'], function () {
 });
 
 gulp.task('minified-post', ['minified-pre'], function () {
-  var viewerFiles = [
-    MINIFIED_DIR + BUILD_DIR + 'pdf.js',
-    MINIFIED_DIR + '/web/viewer.js'
-  ];
+  var pdfFile = fs.readFileSync(MINIFIED_DIR + '/build/pdf.js').toString();
+  var pdfWorkerFile =
+    fs.readFileSync(MINIFIED_DIR + '/build/pdf.worker.js').toString();
```
timvandermeij: I think we should use `BUILD_DIR` here and in the other parts of this patch to not hardcode that and to be consistent with other places in the Gulpfile.

timvandermeij (changes requested):

gulpfile.js
Snuffleupagus: I was just moving existing code around, but I suppose it wouldn't hurt to change this :-)

gulpfile.js
Snuffleupagus: Sure, will do.

timvandermeij (approved):

timvandermeij:
Works fine, thank you!

PR #8835, opened 2017-08-28
Ensure that `PDFViewerApplication.error` outputs proper messages in FIREFOX/MOZCENTRAL builds

Snuffleupagus:
*It appears that this accidentally broke with PR #8394.*

Currently, the following will be printed in the console:
```
An error occurred while loading the PDF.
[object Promise],[object Promise]
```

With this patch we'll again get proper output, e.g. something with this format:
```
An error occurred while loading the PDF.
PDF.js v? (build: ?)
Message: unknown encryption method
```

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -841,7 +841,9 @@ let PDFViewerApplication = {
         errorMoreInfo.value = parts.join('\n');
       });
     } else {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good find, thanks!

PR #8845, opened 2017-08-30
Fixes autofetch and firefox nightly fetch streams

yurydelendik:
* Fixes fetch (#8768) and node behavior when disableAutoFetch and disableStream
* Disables fetch when ReadableStream is not available.


yurydelendik:
Trivial

Issue #8804, opened 2017-08-21
Chrome: Text rendered with SVG backend looks like garbage & text disappears while scrolling


Rob--W:
Chromium (60.0.3112.90, Linux) seems to have a difficulty with rendering a PDF with the SVG backend: https://www.pdflib.com/fileadmin/pdflib/Cookbook/pdf/footnotes_in_text.pdf (from #8796). (looks fine in Firefox 55)

Steps to reproduce the problem:
1. Open the viewer: https://mozilla.github.io/pdf.js/web/viewer.html?file=https://cors-anywhere.herokuapp.com/https://www.pdflib.com/fileadmin/pdflib/Cookbook/pdf/footnotes_in_text.pdf
2. Change rendering backend to SVG: Run `PDFViewerApplication.preferences.set('renderer', 'svg');` in the console.
3. Reload the page (this opens the viewer with the above PDF. Alternatively, if you haven't set `file=`, you can save the PDF, and use the "Open file" button to open the file).

Look at the result. There are two bugs:

1. Text looks like garbage.
2. Scroll up and down, and the text seems to flash away and back beyond a certain point.

Both bugs go away if you change the browser zoom, window size, or the zoom level in PDF.js.

![chrome-60-pdf-svg](https://user-images.githubusercontent.com/1365071/29541906-7930aff0-86d7-11e7-9d1e-31e79903edd7.gif)

For bug 2, I guess that it might have something to do with compositing layers. I enabled the "Composited render layer borders" flag at `chrome://flags` (visit `chrome://flags/#composited-layer-borders` ) and indeed, the point where the text disappears/appears is exactly the edge of a compositing layer:

![chrome-60-pdf-svg-with-frames](https://user-images.githubusercontent.com/1365071/29542085-f5edad94-86d8-11e7-962a-09ee1450c072.gif)

I think that this might be a bug in Chrome, since it looks good in Firefox 55.

Rob--W:
This seems like a regression in Chromium:

- 53.0.2785.116 - bug-free
- 54.0.2840.90 - bug 1 (text looks like garbage)
- 55.0.2883.75 - bug 1 (text looks like garbage)
- 56.0.2924.87 - bug 1 (text looks like garbage)
- 57.0.2987.110 - both bugs are present
- 60.0.3112.9 - both bugs are present (this is the current stable version of Chrome)

For bug 1, I did a bisect on Chrome using bisect-builds.py. The result does not show any obvious offenders...:
[...]

PR #8848, opened 2017-08-31
Correctly set the `credentials` of a fetch request, when the `withCredentials` parameter was passed to `getDocument`

Snuffleupagus:
Skimming through https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Sending_a_request_with_credentials_included, it looks to me like the `credentials` option was accidentally inverted.

@yurydelendik If I've completely misunderstood this code, and the MDN docs, please just close this PR (and sorry about wasting your time in that case).

### DIFF
--- a/src/display/fetch_stream.js
+++ b/src/display/fetch_stream.js
@@ -26,7 +26,7 @@ function createFetchOptions(headers, withCredentials) {
     method: 'GET',
     headers,
     mode: 'cors',
-    credentials: withCredentials ? 'omit' : 'include',
+    credentials: withCredentials ? 'include' : 'omit',
     redirect: 'follow',
   };
 }


Rob--W:
The logic is good, but tests are needed to avoid regressions. I noticed this regression when I wanted to run the test case for #8804 (which worked before the fetch feature landed).

I'll continue with writing the test case (since I started already) and see how I can attach the commit to this PR (I've seen other repositories with multiple commits from different authors, so that should be possible, I guess).

Rob--W:
I'm going to merge this because I don't see a way to open a PR that depends on this, or attach another commit to this PR (I would need write access to your branch).

Rob--W:
Until #8855 is merged, a manual test is to open the test URL from #8804 and see if the load succeeds.

PR #8850, opened 2017-08-31
CORS with credentials fix.

pushkar8723:
Because of incorrect ternary condition, 
```
The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.
```
 error is shown and file fails to load even though no credential is not required.

### DIFF
--- a/src/display/fetch_stream.js
+++ b/src/display/fetch_stream.js
@@ -26,7 +26,7 @@ function createFetchOptions(headers, withCredentials) {
     method: 'GET',
     headers,
     mode: 'cors',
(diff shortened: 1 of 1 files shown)


Rob--W:
I spotted this too, and was working on a patch with tests to ensure that this functionality does not regress. Do you want to take over and write the test yourself?

Snuffleupagus:
Thanks for submitting a patch, however I'm closing this as a duplicate since the same change has already been proposed in PR #8848.

PR #8855, opened 2017-08-31
Add test for withCredentials option

Rob--W:
Add unit test to avoid regressing on #8848.

Locally, I verified that without #8848, some of the new tests fail.
With #8848 (which has been merged), all added tests pass.

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -858,6 +858,85 @@ describe('api', function() {
         done.fail(reason);
       });
     });
+
+    describe('Cross-origin', function() {
+      var loadingTask;
+      function _checkCanLoad(expectSuccess, filename, options) {
+        if (isNodeJS()) {
+          pending('Cannot simulate cross-origin requests in Node.js');
+        }
+        var params = buildGetDocumentParams(filename, options);
+        var url = new URL(params.url);
+        if (url.hostname === 'localhost') {
+          url.hostname = '127.0.0.1';
+        } else if (params.url.hostname === '127.0.0.1') {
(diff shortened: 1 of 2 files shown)


test/unit/api_spec.js
```diff
@@ -858,6 +858,85 @@ describe('api', function() {
...
+        testCannotLoad('basicapi.pdf').then(done);
+      });
+      it('server allows cors without credentials, default withCredentials',
+          function(done) {
+        testCanLoad('basicapi.pdf?cors=withoutCredentials').then(done);
+      });
+      it('server allows cors without credentials, and withCredentials=false',
+          function(done) {
+        testCanLoad('basicapi.pdf?cors=withoutCredentials', {
+          withoutCredentials: false,
```
Snuffleupagus: Nit: Is this supposed to read `withCredentials: false,` instead, given the test name?

Rob--W: Fixed, thanks. The current default is also `withCredentials: false`, so the misspelled name doesn't affect the test result.

yurydelendik (approved):
Looks good, thank you for the patch.

Rob--W:
This a unittest-only change, so the failure in the non-unit tests are obviously not caused by this PR.

Most of my recent test runs for other patches failed tests on Windows. The noise is a bit distracting, is something being done against that?

timvandermeij:
@Rob--W Most of them are fixed, but only the person with dog PDF remains intermittent on Windows. I'd say just ignore that one for now; if it last a long time, we can always decide to temporarily disable it.

Thank you for providing this unit test!

PR #8857, opened 2017-08-31
Avoid some redundant type checks in `XRef.fetchUncompressed`

Snuffleupagus:
When looking briefly at using `Number.isInteger`/`Number.isNan` rather than `isInt`/`isNaN` (re: https://github.com/mozilla/pdf.js/pull/8643#issuecomment-315129972), I noticed that there's a couple of not entirely straightforward cases to consider.

At first I really couldn't understand why `parseInt` is being used like it is in `XRef.fetchUncompressed`, since the `num` and `gen` properties of an object reference should *always* be integers.
However, doing a bit of code archaeology pointed to PR https://github.com/mozilla/pdf.js/pull/4348#issuecomment-36367149, and it thus seem that this was a very deliberate change. Since I didn't want to inadvertently introduce any regressions, I've kept the `parseInt` calls intact but moved them to occur *only* when actually necessary.[1]

Secondly, I noticed that there's a redundant `isCmd` check for an edge-case of broken operators. Since we're throwing a `FormatError` if `obj3` isn't a command, we don't need to repeat that check.

In practice, this patch could perhaps be considered as a micro-optimization, but considering that `XRef.fetchUncompressed` can be called *many* thousand times when loading larger PDF documents these changes at least cannot hurt.

---
[1] I even ran all tests locally, with an added `assert(Number.isInteger(obj1) && Number.isInteger(obj2));` check, and everything passed with flying colours.
However, since it appears that this was in fact necessary at one point, one possible explanation is that the failing test-case(s) has now been replaced by reduced ones.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -1346,16 +1346,21 @@ var XRef = (function XRefClosure() {
       var obj1 = parser.getObj();
       var obj2 = parser.getObj();
       var obj3 = parser.getObj();
-      if (!isInt(obj1) || parseInt(obj1, 10) !== num ||
-          !isInt(obj2) || parseInt(obj2, 10) !== gen ||
-          !isCmd(obj3)) {
+
+      if (!Number.isInteger(obj1)) {
+        obj1 = parseInt(obj1, 10);
+      }
+      if (!Number.isInteger(obj2)) {
+        obj2 = parseInt(obj2, 10);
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good. Thank you!

PR #8865, opened 2017-09-02
Hide unsupported `LinkAnnotation`s (issue 3897)

### LINKED ISSUE #3897: Unrecognized link types
- [x] https://dl.dropboxusercontent.com/u/12899057/jfp-bug.pdf (Unrecognized link type: JavaScript)
- [x] http://ftp.uni-erlangen.de/mirrors/CTAN/fonts/opensans/doc/fonts/opensans/opensans.pdf (Unrecognized link type: Launch)
- [ ] `file://` links (we should disable these)

Snuffleupagus:
Rather than displaying links that does *nothing* when clicked, it probably makes more sense to simply not render them instead. Especially since it turns out that, at least at this point in time, this is *very* easy to both implement and test.

Fixes #3897.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -251,7 +251,9 @@ class AnnotationElement {
 
 class LinkAnnotationElement extends AnnotationElement {
   constructor(parameters) {
-    super(parameters, /* isRenderable = */ true);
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Thank you; I'm really happy that this is now so easy to fix!

PR #8866, opened 2017-09-02
Correctly validate the response status for non-HTTP fetch requests (PR 8768 follow-up)

### LINKED ISSUE #8870: Latest version of pdf.js does not work with blobs
We had to revert version back from
"pdfjs-dist": "^1.9.484",

to

"pdfjs-dist": "1.9.476",

because of error 200.

Snuffleupagus:
It seems that the status check, for non-HTTP loads, has been inverted which causes the default viewer to *refuse* to open local PDF files. *Follow-up to PR #8768.*

***STR:***
 1. Make sure that fetch support is enabled in the browser. In Firefox Nightly, set `dom.streams.enabled = true` and `javascript.options.streams = true` in `about:config`. *Edit:* Just to clarify, please note that the issue is reproducible in Google Chrome too.
 2. Open https://mozilla.github.io/pdf.js/web/viewer.html.
 3. Click on the "Open file" button, and open a new PDF file.

***ER:***
 A new PDF file should open in the viewer.

***AR:***
 The PDF file fails to open, with an error message of the following format:
`Message: Unexpected server response (200) while retrieving PDF "blob:https://mozilla.github.io/a4fc455f-bc05-45b5-b6aa-2ecff3cb45ce".`

*Edit:* Fixes #8870.

### DIFF
--- a/src/display/fetch_stream.js
+++ b/src/display/fetch_stream.js
@@ -94,7 +94,7 @@ class PDFFetchStreamReader {
     let url = this._stream.source.url;
     fetch(url, createFetchOptions(this._headers, this._withCredentials)).
         then((response) => {
-      if (!validateResponseStatus(response.status, this._stream.isHttp)) {
+      if (!validateResponseStatus(response.status)) {
         throw createResponseStatusError(response.status, url);
       }
       this._reader = response.body.getReader();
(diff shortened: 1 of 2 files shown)


yurydelendik (changes requested):
Let's remove entire if(!isHttp) statement. Not sure if fetch() for file: protocol is returning 0 for success.

Snuffleupagus:
> Let's remove entire if(!isHttp) statement.

Fixed now, thanks!

> Not sure if fetch() for file: protocol is returning 0 for success.

According to https://fetch.spec.whatwg.org/#scheme-fetch, it would seem that `file://` isn't even in the specification.

yurydelendik (approved):

yurydelendik:
Thank you for the patch.

mukulmishra18:
> According to https://fetch.spec.whatwg.org/#scheme-fetch, it would seem that file:// support isn't even properly defined in the specification.

`file:` protocol is not supported by the native fetch implementation and always throws `TypeError`, see https://github.com/github/fetch/issues/186 .

PR #8868, opened 2017-09-04
Store the rotation in the `ViewHistory`/`PDFHistory` (issue 5927)

### LINKED ISSUE #5927: Viewer doesn't remember last rotation angle

Snuffleupagus:
Please note that compared to previous attempts at fixing this, the commits in this PR adds rotation support to the history implementation as well.

Fixes #5927.

### DIFF
--- a/test/unit/ui_utils_spec.js
+++ b/test/unit/ui_utils_spec.js
@@ -14,8 +14,8 @@
  */
 
 import {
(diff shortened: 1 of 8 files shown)


timvandermeij (approved):

timvandermeij:
Nice to have this; thank you!

PR #8874, opened 2017-09-05
Fix the helloworld example by setting the PDFNetworkStream class (PR 8712 follow-up)

wenxcs:
The problem like https://stackoverflow.com/q/45393309/5317012 (PR#8722) is still there, but the reason now is different.

### DIFF
--- a/examples/helloworld/hello.js
+++ b/examples/helloworld/hello.js
@@ -6,7 +6,9 @@ Promise.all([System.import('pdfjs/display/api'),
              System.import('pdfjs/display/network'),
              System.resolve('pdfjs/worker_loader')])
        .then(function (modules) {
-  var api = modules[0], global = modules[1];
+  var api = modules[0], global = modules[1], network = modules[2];
+  api.setPDFNetworkStreamClass(network.PDFNetworkStream);
+
   // In production, change this to point to the built `pdf.worker.js` file.
   global.PDFJS.workerSrc = modules[3];
 


examples/helloworld/hello.js
```diff
@@ -6,9 +6,10 @@ Promise.all([System.import('pdfjs/display/api'),
              System.import('pdfjs/display/network'),
              System.resolve('pdfjs/worker_loader')])
        .then(function (modules) {
-  var api = modules[0], global = modules[1];
+  var api = modules[0], global = modules[1], network = modules[2];
   // In production, change this to point to the built `pdf.worker.js` file.
   global.PDFJS.workerSrc = modules[3];
+  api.setPDFNetworkStreamClass(network.PDFNetworkStream)
```
Snuffleupagus: Please add a semi-colon at the end of this line. Also, let's move this line to just below https://github.com/mozilla/pdf.js/pull/8874/files#diff-073ee7f864077e4b3968f8df3ba154e9R9 instead.

Snuffleupagus (changes requested):
Please update the commit message (and PR title) to something more descriptive than just "A fix to make hello world runs okay.", since that provides no information about what the patch *actually* does.
I'd suggest e.g. "Fix the helloworld example by setting the `PDFNetworkStream` class (PR 8712 follow-up)" instead.

Don't forget to squash the commits when you're done, please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

zztoy:
Thanks for @Snuffleupagus 's guidance. I renamed the PR title and move the new statement to better position.

Snuffleupagus:
> I renamed the PR title [...]

Thanks, but please also update the *commit message* to the same as the PR title.

zztoy:
Commit message updated.

Snuffleupagus:
> Commit message updated.

Wait, why is PR #8722 mentioned in the commit message (and PR title), since that's *not* where this actually broke!?
It did work after that PR, but was unfortunately broken again in PR #8712, which is why https://github.com/mozilla/pdf.js/pull/8874#pullrequestreview-60594139 suggested:
> "Fix the helloworld example by setting the PDFNetworkStream class (PR 8712 follow-up)"

zztoy:
Sorry, it's my bad. I miss-understand the "PR title" 's meaning.

Snuffleupagus (approved):

PR #7664, opened 2016-09-22
Acroform checkbox and radio controls

benweet:
Added fake checkboxes and radio buttons with a pinch of css.
Tested against http://www.cic.gc.ca/english/passport/forms/pdf/pptc153.pdf on Chrome and Firefox:

screen shot 2016-09-22 at 16 19 06

PR #8877, opened 2017-09-06
Provide `checked` styles for button widget annotations

### LINKED ISSUE #8875: Checkbox in interactive form not showing up with renderInteractiveForms enabled
[f1040 copy.pdf](https://github.com/mozilla/pdf.js/files/1279220/f1040.copy.pdf)

Configuration:
- Web browser and its version: Chrome 60.3112.113, Safari 10.1.2, Firefox 55.0.3
- Operating system and its version: macOS Sierra 10.12.6
- PDF.js version: 1.9.450
- Is an extension: No

1. Use pdf.js, turn on `renderInteractiveForms` and load the file. Or visit https://cloud.box.com/s/snwp6gyhdk6wh2m9x46p81ubs73829vj
2. See that the checkbox for `Single` under Filing Status is not checked

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): https://cloud.box.com/s/snwp6gyhdk6wh2m9x46p81ubs73829vj

Looks like this issue only happens when `renderInteractiveForms` is enabled. Other non-checkbox type fields are filled in. When `renderInteractiveForms` is turned off, checkboxes render correctly. 

This looks potentially related to https://github.com/mozilla/pdf.js/issues/8159, but that might refer to rendering without interactive forms?

timvandermeij:
Fixes #8875.

Use `test/pdfs/annotation-button-widget.pdf` (existing testcase) to verify this, together with `PDFViewerApplication.preferences.set('renderInteractiveForms', true);`.

### DIFF
--- a/test/annotation_layer_test.css
+++ b/test/annotation_layer_test.css
@@ -63,6 +63,10 @@
   width: 100%;
 }
 
+.annotationLayer .buttonWidgetAnnotation.radioButton input {
+  border-radius: 50%;
+}
+
 .annotationLayer .textWidgetAnnotation textarea {
   font: message-box;
   font-size: 9px;
@@ -78,6 +82,38 @@
   border: 1px solid transparent;
 }
 
+.annotationLayer .buttonWidgetAnnotation.checkBox input:checked:before,
+.annotationLayer .buttonWidgetAnnotation.checkBox input:checked:after,
+.annotationLayer .buttonWidgetAnnotation.radioButton input:checked:before {
+  background-color: #000;
+  content: '';
+  display: block;
+  position: absolute;
+}
+
+.annotationLayer .buttonWidgetAnnotation.radioButton input:checked:before {
+  border-radius: 50%;
+  height: 50%;
+  left: 30%;
+  top: 20%;
+  width: 50%;
+}
+
+.annotationLayer .buttonWidgetAnnotation.checkBox input:checked:before,
+.annotationLayer .buttonWidgetAnnotation.checkBox input:checked:after {
+  height: 80%;
+  left: 45%;
+  width: 1px;
+}
+
+.annotationLayer .buttonWidgetAnnotation.checkBox input:checked:before {
+  transform: rotate(45deg);
+}
+
+.annotationLayer .buttonWidgetAnnotation.checkBox input:checked:after {
+  transform: rotate(-45deg);
+}
+
 .annotationLayer .textWidgetAnnotation input.comb {
   font-family: monospace;
   padding-left: 2px;
@@ -90,6 +126,7 @@
   -moz-appearance: none;
   -ms-appearance: none;
   appearance: none;
+  padding: 0;
 }
 
 .annotationLayer .popupAnnotation {
--- a/web/annotation_layer_builder.css
+++ b/web/annotation_layer_builder.css
@@ -56,6 +56,10 @@
   width: 100%;
 }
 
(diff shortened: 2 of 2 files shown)


Snuffleupagus (commented):
To me, using this kind of CSS trickery seems unnecessarily convoluted and difficult to wrap your head around.
Looking a the test results, it also seems that the vertical alignment in (particularly) the checkboxes isn't that great.

This kind of reminds me somewhat of the approach in PR #7664, where I believe we agreed that using too much CSS where not conducive to readability and maintainability of the code. (Obviously this PR attempts to style the correct elements, but nonetheless.)

Furthermore, it seems to me that all of this is just temporary anyway, since I'd imagine that what you *ultimately* what is to somehow parse the Appearance Streams of various form controls, and use that to render the *correct* checkbox/radio controls in the UI layer somehow (as I believe was discussed on IRC at one time: http://logs.glob.uno/?c=mozilla%23pdfjs&s=16+Dec+2016&e=16+Dec+2016#c55484).

With all this said, given that forms support is nowhere near ready for prime-time, couldn't we just remove the `appearance: none;` overrides for checkbox/radio controls and fallback to rendering native browser for now to simplify things!?

Please note that I'm not saying outright "no" to this PR, but I cannot help wondering if a simpler solution wouldn't suffice at the present time :-)

timvandermeij:
The regression tests are a bit weird, especially in Chrome, because it renders the same in Firefox and Chrome in my local testing. I am happy that the regression tests at least show the checked indicators again whereas before they didn't.

The advantage of this PR is that it adresses the current problems and that it makes rendering mostly equal to Adobe Reader/Acrobat. The disadvantage is slightly more CSS, but I think it's manageable. It would indeed be even better to use the appearance streams, but not only do I not really see that happening soon, I also think we'll end up with somewhat the same situation where we have to style UI elements using information in the appearance stream, so I'm not sure if it's really going to be different then.

Yury suggested on IRC to generate images from the appearance stream, but that would only work for button widgets. For choice widgets, we do the styling manually because there is no way we can use images for that later on. Hence I thought we should stick to one solution for now, CSS, for consistency.

I do see your point though. I'm just not sure if not styling only the button widgets at all is a good alternative. Appearance streams will not be implemented in the short term, so I don't think it will be too temporary and it would be strange to have styled choice widgets and unstyled button widgets... My feeling therefore is to proceed with this PR and think about better alternatives afterwards, but if you're really opposed to that, we can consider not styling the button widgets at all, though that will not be great UX. What do you think?

Snuffleupagus:
> Yury suggested to generate images from the appearance stream, but that would only work for button widgets. For choice widgets, we do the styling manually because there is no way we can use images for that. Hence I thought we should stick to one solution for now, CSS, for consistency.

I'd imagine that you might actually be able to use buttons for those as well (with `aria` labels to indicate its function), which would allow you to checkboxes/radiobuttons here as well.

> Appearance streams will not be implemented in the short term, so I don't think it will be too temporary and it would be strange to have styled choice widgets and unstyled button widgets...

Fair point, and I don't really care *that* much (just thinking about maintainability here).

---

Another thing that I think would be nice to look into, in a follow-up of course, is if we can reduced the amount CSS duplication between `web/annotation_layer_builder.css` and ` test/annotation_layer_test.css`.
Could we have the test CSS file `@import(...)`ing the regular one, and provide overrides where necessary?

Snuffleupagus (approved):

PR #8885, opened 2017-09-07
Address a couple of edge-cases in the new `PDFHistory` implementation (PR 8775 follow-up)

Snuffleupagus:

Despite careful testing there's a couple of edge-case issues that managed to sneak into the new `PDFHistory` implementation (see PR #8775), refer in particular to the *last* commit, which is thus being addressed here.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #8889, opened 2017-09-09
Replace `value === (value | 0)` checks with `Number.isInteger(value)`

Snuffleupagus:
Rather than doing what (at first) may seem like a fairly obscure comparison, using `Number.isInteger` will clearly indicate the intent of the code.


timvandermeij (approved):

timvandermeij:
Good improvement!

PR #8893, opened 2017-09-10
Combine the common styles and overrides for the annotation layer reference tests

timvandermeij:
This patch allows us to use the common styles as used by the viewer as a baseline for the annotation layer reference tests. They are extended with a small set of overrides to ensure that all elements are visible during the test.

The overrides file now only contains the absolutely necessary rules to make all elements visible and is therefore no longer an almost verbatim copy of the common styles.

@Snuffleupagus Since you came up with this nice suggestion, would you be willing to review this?


Snuffleupagus (approved):
Too bad that we don't load the stylesheets in a way that supports `@import`, but still much nicer than having to maintain two copies of the same CSS rules :-)

Looks good, thank you!

PR #8900, opened 2017-09-12
Use the `SimpleLinkService` when running "annotations" reference tests

Snuffleupagus:
Rather than (basically) duplicating the `SimpleLinkService` in `test/driver.js`, with potential test failuires if you forget to update the test mock, it seems much nicer to just re-use the viewer component.

Note that `SimpleLinkService` is already bundled into the `build/components/pdf_viewer.js` file. Hence we only need to expose it similar to the other viewer components in that file, and make sure that the `gulp components` command runs as part of the test-setup.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1085,22 +1085,22 @@ gulp.task('publish', ['generic'], function (done) {
     });
 });
 
-gulp.task('test', ['generic'], function () {
+gulp.task('test', ['generic', 'components'], function () {
   return streamqueue({ objectMode: true, },
     createTestSource('unit'), createTestSource('browser'));
 });
 
-gulp.task('bottest', ['generic'], function () {
+gulp.task('bottest', ['generic', 'components'], function () {
   return streamqueue({ objectMode: true, },
     createTestSource('unit'), createTestSource('font'),
     createTestSource('browser (no reftest)'));
 });
 
-gulp.task('browsertest', ['generic'], function () {
+gulp.task('browsertest', ['generic', 'components'], function () {
   return createTestSource('browser');
 });
 
-gulp.task('unittest', ['generic'], function () {
+gulp.task('unittest', ['generic', 'components'], function () {
   return createTestSource('unit');
 });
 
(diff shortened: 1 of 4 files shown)


timvandermeij:
While this commit in itself looks good, I'm worried about the increase in time on the Windows bot. I saw that while it takes only a few seconds to run Webpack on the Linux bot it takes up to 10 minutes on the Windows bot. This commit adds an extra Webpack steps, thereby increasing the time with such an amount. I think we should figure out what the problem is on the Windows bot first. I had hoped updating Webpack to the latest version would help, but unfortunately it did not change anything.

@yurydelendik Would you be able to help out here? Do you know what may cause the Windows bot to be so slow with Webpack?

Snuffleupagus:
> This commit adds an extra Webpack step, thereby increasing the time with such an amount.

I'm not convinced that the added `gulp components` step is the culprit here, since according to the logs: `[13:33:02] Finished 'components' after 7.06 s`.
Looking further below, you can see `[13:39:57] Finished 'generic' after 6.92 min`, which is obviously slower than one would like but still on par with other (older) test runs on Windows.

Furthermore, it's not actually Webpack that's slow on Windows. Instead, it's Babel which takes a long time to run. Locally on Windows, I'm running all tests with `PDFJS_NEXT=true gulp {test name here}`, which disables the transpilation step and is thus very quick. Perhaps we want to skip Babel when running *tests* on the bots as well?

timvandermeij:
> Perhaps we want to skip Babel when running tests on the bots as well?

I agree with that. The bots run the most recent Firefox and Chrome versions, so there should be no need for the transpilation step. Could you add a commit to disable that so we can do a new test run?

Snuffleupagus:
> Could you add a commit to disable that so we can do a new test run?

The changes would need to be done at https://github.com/mozilla/botio-files-pdfjs, and then pushed to the bots manually. I'll see about submitting a PR for that tomorrow!

In the meantime, I'll trigger a new round of testing to check that the Windows bot wasn't just temporarily slow when the tests ran earlier today.

Snuffleupagus:
> In the meantime, I'll trigger a new round of testing to check that the Windows bot wasn't just temporarily slow when the tests ran earlier today.

It seems to have been just a temporary problem, since the runtime of the Windows bot is now back to its usual value.

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

Good to know that the runtime is not a new issue. The other patch will hopefully improve the runtime.

PR #8901, opened 2017-09-12
CCITTFaxStream problem when EndOfBlock is false

janpe2:
I have been working on the JBIG2 decoder to add support for the Huffman coding. When a JBIG2 region uses the Huffman option, its bitmaps (1-bit images) are compressed using the MMR coding. MMR is exactly the same algorithm as the PDF filter `CCITTFaxDecode` with `/K -1`. Thus I would like to use the existing decoder `CCITTFaxStream` that PDF.js already has.

MMR bitmaps in JBIG2 typically have no end-of-data codes (EOFB), so I must set the parameter `EndOfBlock` to false when using `CCITTFaxStream`. This seems to cause a problem: the last pixel row of decoded images ends prematurely.

In the JBIG2 decoder, bitmaps have no alpha channel, so the incomplete row appears as a nasty black line. Here is a screenshot of a Huffman coded JBIG2 image whose text symbols were decoded using `CCITTFaxStream`:
jbig2_mmr_black_stripes

I created a PDF test file and I found out that the same problem affects also `CCITTFaxDecode` images in PDFs when they have `/EndOfBlock false`.
[ccitt_EndOfBlock_false.pdf](https://github.com/mozilla/pdf.js/files/1296345/ccitt_EndOfBlock_false.pdf)
In each pair of images the top one has `/EndOfBlock false`, so the last row is incomplete.

The parameter `/EndOfBlock false` is very rare in real-world PDFs (default value is true). The files that I have found have no problems in PDF.js. It is impossible to see any missing pixels because the images have a high resolution and their last row is white against a white page background.

In `CCITTFaxStream` the property `eoblock` corresponds to the PDF parameter `EndOfBlock`. When its value is false, the decoding must end when all rows have been decoded (and EOFB is not expected), as specified in the PDF standard. The function `lookChar()` in `CCITTFaxStream` does exactly that and sets `this.eof` to true when the last row has been decoded. However, the function returns only a few bits from the start of the last pixel row. `lookChar()` is called only once in the last row because setting `this.eof = true` stops the whole decoding process in `readBlock()`. The rest of the decoded pixels of the last row remain in the array `codingLine` but they are never returned because `lookChar()` is not called anymore.

In this PR, I add a new property `rowsDone`. When `EndOfBlock` is false and the CCITT data of the last row has been decoded, `rowsDone` is set to true and `eof` remains false. The value `eof` is not set to true until all the buffered contents of `codingLine` have been consumed too. That happens when `this.outputBits` becomes zero.


src/core/stream.js
```diff
@@ -1858,6 +1859,10 @@ var CCITTFaxStream = (function CCITTFaxStreamClosure() {
     var refPos, blackPixels, bits, i;
 
     if (this.outputBits === 0) {
+      if (this.rowsDone) {
+        this.eof = true;
+        return null;
```
Snuffleupagus: Nit: Given the condition on the very next line, do we really need to `return null;` here as well?

Snuffleupagus (commented):
Since you created a nice reduced test-case, can you please attach it to the PR as well? :-)

Snuffleupagus (approved):
Thanks for the very clear description of the problem/solution, which made it much easier to understand the changes!

Since you created the test file yourself, could you please add it directly to the PR instead (so that we can avoid an unnecessary linked test-case)?

With that change, and the commits squashed, r=me.
Thank you for the patch!

PR #8904, opened 2017-09-12
Provide a stub for `setAttribute` in order to use the SVG back-end with Node.js

### LINKED ISSUE #8902: error running node pdf2svg.js

Configuration:
- Web browser and its version: node
- Operating system and its version: Ubuntu 16.04
- PDF.js version: 23cbe294d578085c295e700963c092816b5bf7e3
- Is an extension: no

1.  cd examples/node
2. node pdf2svg.js

convert pdf to svg

```
Error: TypeError: svg.setAttribute is not a function
```

timvandermeij:
This patch fixes a regression from PR #8691 where we switched to using `setAttribute` instead of `setAttributeNS` if no namespace is provided.

To avoid further issues, a stub for `getAttribute` is also created.

Fixes #8902.

### DIFF
--- a/examples/node/domstubs.js
+++ b/examples/node/domstubs.js
@@ -58,6 +58,12 @@ function DOMElement(name) {
 }
 
 DOMElement.prototype = {
+  getAttribute: function DOMElement_getAttribute(name) {
+    if (name in this.attributes) {
+      return this.attributes[name];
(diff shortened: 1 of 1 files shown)


yurydelendik (approved):
Thank you for the patch.

PR #8909, opened 2017-09-14
Check `isEvalSupported`, and test that `eval` is actually supported, before attempting to use the `PostScriptCompiler` (issue 5573)

### LINKED ISSUE #5573: PostScriptCompiler compiled result violates the Content security policy (no unsafe-eval)

Snuffleupagus:
Currently `PDFFunction` is implemented (basically) like a class with only `static` methods. Since it's used directly in a number of different `src/core/` files, attempting to pass in `isEvalSupported` would result in code that's *very* messy, not to mention difficult to maintain (since *every* single `PDFFunction` method call would need to include a `isEvalSupported` argument).

Rather than having to wait for a possible re-factoring of `PDFFunction` that would avoid the above problems by design, it probably makes sense to at least set `isEvalSupported` globally for `PDFFunction`.

Please note that there's one caveat with this solution: If `PDFJS.getDocument` is used to open multiple files simultaneously, with *different* `PDFJS.isEvalSupported` values set before each call, then the last one will always win.
However, that seems like enough of an edge-case that we shouldn't have to worry about it. Besides, since we'll also test that `eval` is actually supported, it should be fine.

Fixes #5573.

Supersedes PR #8815 (since that one hasn't seen any activity for weeks, and based on the findings above I'm no longer entirely convinced that it would have been a simple beginner bug).

*Edit:* Slightly smaller diff with https://github.com/mozilla/pdf.js/pull/8909/files?w=1.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -24,6 +24,7 @@ import { OperatorList, PartialEvaluator } from './evaluator';
 import { AnnotationFactory } from './annotation';
 import { calculateMD5 } from './crypto';
 import { Linearization } from './parser';
+import { PDFFunction } from './function';
 
 var Page = (function PageClosure() {
 
@@ -532,6 +533,9 @@ var PDFDocument = (function PDFDocumentClosure() {
         },
       };
       this.catalog = new Catalog(this.pdfManager, this.xref, pageFactory);
+
+      let evaluatorOptions = this.pdfManager.evaluatorOptions;
+      PDFFunction.setIsEvalSupported(evaluatorOptions.isEvalSupported);
     },
     get numPages() {
       var linearization = this.linearization;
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Looks good. Given the current state of that code, I agree that this is the nicest solution for now.

yurydelendik:
Worker can handle multiple documents and handling isEvalSupported this way may cause confusion. When need to 1) make PDFFunction factory, which is constructed based on isEvalSupported passed to worker, 2) or, pass it similar to verbosity option.

Snuffleupagus:
> Worker can handle multiple documents and handling isEvalSupported this way may cause confusion. When need to 1) make PDFFunction factory, which is constructed based on isEvalSupported passed to worker,

Sure, I'm currently looking into re-factoring `PDFFunction` and related code to fix this properly.

PR #8910, opened 2017-09-14
Fix #8888: Change behaviour of fetch to make it compatible with XHR.

### LINKED ISSUE #8888: v1.9.528 doesn't include credentials on chrome
We had this code to get pdf through authenticated url 
`PDFJS.getDocument(url)`
that breaks with version 1.9.528 on Chrome because it doesn't include CSRF-TOKEN in the request. However, it works on firefox and safari. Now I have to fix it by enforce withCredentials: true. Does new version set the default of withCredentials to false?

mukulmishra18:
Fixes #8888 .

### DIFF
--- a/src/display/fetch_stream.js
+++ b/src/display/fetch_stream.js
@@ -26,7 +26,7 @@ function createFetchOptions(headers, withCredentials) {
     method: 'GET',
     headers,
     mode: 'cors',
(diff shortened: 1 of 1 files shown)


yurydelendik:
Thank you for the patch.

PR #8912, opened 2017-09-15
[api-minor] Replace `DOMParser` with `SimpleXMLParser`

### LINKED ISSUE #8903: Replace DOMParser with custom parser
The https://github.com/mozilla/pdf.js/blob/master/src/display/metadata.js is using DOMParser. The DOMParser is missing in node.js environment. Also checking the code, we still has to "fix" the XML data. There is also concern that using generic DOMParser might be overkill or less-secure.

Let's replace it with simple implementation e.g. from https://github.com/mozilla/pdf.js/blob/master/examples/node/domparsermock.js .

timvandermeij:
The `DOMParser` is most likely overkill and may be less secure. Moreover, it is not supported in Node.js environments.

This patch replaces the `DOMParser` with a simple XML parser. This should be faster and gives us Node.js support for free. The simple XML parser is a port of the one that existed in the examples folder with improved regexes to make the parsing work properly.

The unit tests are extended for increased test coverage of the metadata code. The new method `getAll` is provided so the example does not have to access internal properties of the object anymore.

Fixes #8903.


src/display/metadata.js
```diff
@@ -13,14 +13,130 @@
...
+      let node = new SimpleDOMNode('#text', text);
+      nodes.push(node);
+      return length + ',';
+    });
+
+    let lastLength;
+    do {
+      lastLength = nodes.length;
+      data = data.replace(
+        /<([\w\:]+)((?:[\s\w:=]|'[^']*'|"[^"]*")*)(?:\/>|>([\d,]*)<\/[^>]+>)/g,
```
Snuffleupagus: Rather than creating a regular expression over and over in a loop, would it make sense to just define it *once* instead?

Snuffleupagus (commented):
Looks like a nice simplification!
One question though: Would it maybe make more sense to place the `SimpleXMLParser` (and related code) in a utility file instead (such as `/src/display/dom_utils.js`)?

src/display/metadata.js
timvandermeij: Done in the new commit, together with the code move to the utility file. Good points, thank you!

src/display/metadata.js
```diff
@@ -13,43 +13,50 @@
...
+
+class Metadata {
+  constructor(data) {
+    if (typeof data === 'string') {
+      // Ghostscript may produce invalid metadata, so try to repair that first.
+      data = this._repair(data);
+
+      // Convert the string to a DOM `Document`.
+      let parser = new SimpleXMLParser();
+      data = parser.parseFromString(data, 'application/xml');
```
Snuffleupagus: Nit: It appears that the `parseFromString` method (of the `SimpleXMLParser`) only takes one argument, so presumably we no longer need to pass in `'application/xml'` unless I'm missing something here!?

timvandermeij: D'oh! You're right, since it's now specialized to an XML parser.

src/display/dom_utils.js
```diff
@@ -131,6 +131,122 @@ class DOMSVGFactory {
...
+  hasChildNodes() {
+    return this.childNodes && this.childNodes.length > 0;
+  }
+}
+
+class SimpleXMLParser {
+  parseFromString(data) {
+    let nodes = [];
+
+    data = data.replace(/<\?[\s\S]*?\?>|<!--[\s\S]*?-->/g, '').trim();
```
yurydelendik: comment before: `// Remove all comments and processing instructions`

src/display/metadata.js
```diff
@@ -58,36 +65,41 @@ Metadata.prototype = {
...
-  },
+  get(name) {
+    return this._metadata[name] || null;
+  }
 
-  has: function Metadata_has(name) {
-    return typeof this.metadata[name] !== 'undefined';
-  },
-};
+  getAll() {
```
yurydelendik: keep 'metadata', just have deprecated() and getAll() calls inside the getter.

src/display/dom_utils.js
```diff
@@ -131,6 +131,122 @@ class DOMSVGFactory {
...
+    data = data.replace(/>([^<][\s\S]*?)</g, (all, text) => {
```
yurydelendik: comment before: `// Extract all text nodes and replace them with numeric index in the nodes.`

src/display/dom_utils.js
```diff
@@ -131,6 +131,122 @@ class DOMSVGFactory {
...
+      let length = nodes.length;
+      let node = new SimpleDOMNode('#text', this._decodeXML(text));
+      nodes.push(node);
+      if (node.textContent.trim().length === 0) {
+        return '><'; // Ignore whitespace.
+      }
+      return '>' + length + ',<';
+    });
+    data = data.replace(/<!\[CDATA\[([\s\S]*?)\]\]>/g,
```
yurydelendik: `// Extract all CDATA nodes`

src/display/dom_utils.js
```diff
@@ -131,6 +131,122 @@ class DOMSVGFactory {
...
+        function(all, text) {
+      let length = nodes.length;
+      let node = new SimpleDOMNode('#text', text);
+      nodes.push(node);
+      return length + ',';
+    });
+
+    let regex =
```
yurydelendik: comment before: `// Until such nodes without '<' and '>' content are present, replace these with numeric index in the nodes.`

src/display/dom_utils.js
```diff
@@ -131,6 +131,122 @@ class DOMSVGFactory {
...
+            children.push(childNode);
+          });
+        }
+
+        node.childNodes = children;
+        nodes.push(node);
+        return length + ',';
+      });
+    } while (lastLength < nodes.length);
+
```
yurydelendik: `// We shall have only one root index left, which will be last in the nodes.`

src/display/metadata.js
```diff
@@ -13,43 +13,50 @@
...
+  constructor(data) {
+    if (typeof data === 'string') {
+      // Ghostscript may produce invalid metadata, so try to repair that first.
+      data = this._repair(data);
+
+      // Convert the string to a DOM `Document`.
+      let parser = new SimpleXMLParser();
+      data = parser.parseFromString(data);
+    } else if (!(data instanceof Document)) {
+      throw new Error('Metadata: input is not a string or `Document`');
```
yurydelendik: do we need `instanceof Document` check? if not let's replace `if (typeof data === 'string'` with assert.

yurydelendik (approved):
Thank you for the patch.

PR #8917, opened 2017-09-16
Implement support for stamp annotations

timvandermeij:
This patch includes support for stamp annotations, which make use of the existing framework.

@Snuffleupagus Could you review this? It's very similar to the previous patches.


Snuffleupagus (approved):

PR #8919, opened 2017-09-17
Ignore 'change' events that didn't originate in the viewer (issue 8915)

### LINKED ISSUE #8915: Is it possible to have another input file chooser that will not trigger PDF.JS loading it as PDF?
PDF.JS has a file chooser icon to load another PDF without refreshing the site

i have a different file chooser that does something else, but whenever the file gets selected, PDF.JS will load that file

is there a way to override this?

Snuffleupagus:
Rather that registering a 'change' event listener on the `window`, which will thus (unnecessarily) fire in *a number* of other situations such as e.g. when the user changes the pageNumber or the current search term, we could/should just register it directly on the dynamically created `fileInput` DOM element instead.
I can see no really compelling reason why we actually need to listen for `file` changes on the `window` itself, and this way we're also able to keep the `fileInput` related code confined to one part of the code which should aid readability.
Furthermore, in custom deployments, there's less risk that we're going to interfere with "outside" code this way.

Finally, preprocessor guards were added to the `webViewerOpenFile` function, since that code doesn't make sense in e.g. the extension builds.

Fixes #8915.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1330,18 +1330,6 @@ let PDFViewerApplication = {
     window.addEventListener('hashchange', _boundEvents.windowHashChange);
     window.addEventListener('beforeprint', _boundEvents.windowBeforePrint);
     window.addEventListener('afterprint', _boundEvents.windowAfterPrint);
-    if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

Issue #6289, opened 2015-07-30
16bit image with TIFF predictor not displayed


THausherr:
16bit image with TIFF predictor in test file from the Ghent Workgroup ( http://www.gwg.org/ ) is not displayed:
https://issues.apache.org/jira/secure/attachment/12685775/GWG181_16Bit_CMYK_X4.pdf

(This is probably lowest priority, as this is not a file "from the wild")

timvandermeij:
Fixed by pull request #8921.

PR #8913, opened 2017-09-15
For /Filter entries containing `Name`s, ignore the /DecodeParms entry if it contains an Array (issue 8895)

Snuffleupagus:
Fixes #8895.

PR #8921, opened 2017-09-17
support tiff predictor for 16bit

THausherr:
(for issue #6289)
This does the same for 16 bit as the existing 8 bit tiff predictor code, an addition of the last word to this word.

The last two "& 0xFF" may or may not be needed, I see this isn't done in the 8 bit code, but I'm not a JS developer.

### DIFF
--- a/src/core/stream.js
+++ b/src/core/stream.js
@@ -735,6 +735,19 @@ var PredictorStream = (function PredictorStreamClosure() {
         buffer[pos] = buffer[pos - colors] + rawBytes[i];
         pos++;
       }
+    } else if (bits === 16) {
+      var bytesPerPixel = colors * 2;
+      for (i = 0; i < bytesPerPixel; ++i) {
+        buffer[pos++] = rawBytes[i];
+      }
+      for (; i < rowBytes; i += 2) {
+        var sum = ((rawBytes[i] & 0xFF) << 8) +
+                  (rawBytes[i + 1] & 0xFF) +
+                  ((buffer[pos - bytesPerPixel] & 0xFF) << 8) +
+                  (buffer[pos - bytesPerPixel + 1] & 0xFF);
+        buffer[pos++] = ((sum >> 8) & 0xFF);
+        buffer[pos++] = (sum & 0xFF);
+      }
     } else {
       var compArray = new Uint8Array(colors + 1);
       var bitMask = (1 << bits) - 1;
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -706,6 +706,14 @@
        "lastPage": 1,
        "type": "eq"
     },
+    {  "id": "issue6289",
+       "file": "pdfs/issue6289.pdf",
+       "md5": "0869f3d147c734ec484ffd492104095d",
+       "rounds": 1,
+       "link": true,
+       "lastPage": 1,
+       "type": "eq"
+    },
     {  "id": "issue5509",
        "file": "pdfs/issue5509.pdf",
        "md5": "1975ef8db7355b1d691bc79d0749574b",
--- a/test/pdfs/issue6289.pdf.link
+++ b/test/pdfs/issue6289.pdf.link
@@ -1,0 +1,1 @@
+https://issues.apache.org/jira/secure/attachment/12685775/GWG181_16Bit_CMYK_X4.pdf


src/core/stream.js
```diff
@@ -735,6 +735,16 @@ var PredictorStream = (function PredictorStreamClosure() {
         buffer[pos] = buffer[pos - colors] + rawBytes[i];
         pos++;
       }
+    } else if (bits === 16) {
+      var bytesPerPixel = colors * 2;
+      for (i = 0; i < bytesPerPixel; ++i) {
+        buffer[pos++] = rawBytes[i];
+      }
+	    for (; i < rowBytes; i += 2) {
```
timvandermeij: Nit: needs one extra space for indenting

src/core/stream.js
```diff
@@ -735,6 +735,16 @@ var PredictorStream = (function PredictorStreamClosure() {
...
+        var sum = ((rawBytes[i] & 0xFF) << 8) + (rawBytes[i + 1] & 0xFF) + (((buffer[pos - bytesPerPixel] & 0xFF) << 8) + (buffer[pos - bytesPerPixel + 1] & 0xFF));
```
timvandermeij: This line is too long. Could you split it into multiple lines/variables?

THausherr: Sorry, there was a tab that came in somehow.

src/core/stream.js
```diff
@@ -735,6 +735,19 @@ var PredictorStream = (function PredictorStreamClosure() {
...
+    } else if (bits === 16) {
+      var bytesPerPixel = colors * 2;
+      for (i = 0; i < bytesPerPixel; ++i) {
+        buffer[pos++] = rawBytes[i];
+      }
+      for (; i < rowBytes; i += 2) {
+        var sum = ((rawBytes[i] & 0xFF) << 8) +
+                  (rawBytes[i + 1] & 0xFF) +
+                  (((buffer[pos - bytesPerPixel] & 0xFF) << 8) +
+                  (buffer[pos - bytesPerPixel + 1] & 0xFF));
```
timvandermeij: Could you add one more space the indenting of this line? Because of the many parentheses, it's easier to see that way that this line is a part of the expression on the previous line.

src/core/stream.js
```diff
@@ -735,6 +735,19 @@ var PredictorStream = (function PredictorStreamClosure() {
         buffer[pos] = buffer[pos - colors] + rawBytes[i];
         pos++;
       }
+    } else if (bits === 16) {
```
timvandermeij: In the issue you have a test file for this. I don't know what its license is, but are we free to include that in the test suite? If so, please add it as a regular test case (see https://github.com/mozilla/pdf.js/pull/8917/files#diff-441ddcbb46641da096fad809396ea452 for an example: add the file, a `.gitignore` entry and a test manifest entry). If not, please add it as a linked test case (see https://github.com/mozilla/pdf.js/pull/8913/files#diff-9506708851a1dcc6c972f0831003f3d5 for an example: add the file in a `.link` file and add a test manifest entry). This is to make sure we don't regress this in the future.

timvandermeij (commented):
Looks good to me with these final two comments adressed. Please make sure to squash the commits after this (see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits if you're not familiar with it).

src/core/stream.js
THausherr: Which line? The first? Or all except the first? Or add spaces so that "rawbytes" and "buffer" are aligned? The green block here has 4 lines.

timvandermeij: Line 747. GitHub adds the comment at the correct line, but shows other lines for context. You would end up with:

```javascript
        var sum = ((rawBytes[i] & 0xFF) << 8) +
                  (rawBytes[i + 1] & 0xFF) +
                  (((buffer[pos - bytesPerPixel] & 0xFF) << 8) +
                   (buffer[pos - bytesPerPixel + 1] & 0xFF));
```

timvandermeij: Now that I look at it again, do we actually need those last two parentheses? I think this would be just fine:

```javascript
        var sum = ((rawBytes[i] & 0xFF) << 8) +
                  (rawBytes[i + 1] & 0xFF) +
                  ((buffer[pos - bytesPerPixel] & 0xFF) << 8) +
                  (buffer[pos - bytesPerPixel + 1] & 0xFF);
```

THausherr: The copyright is here at the last page:
http://www.gwg.org/wp-content/uploads/GWG_PDF-Output-Suite4_ReadMeFirst2.pdf

Copyright © 2014, Ghent Workgroup (http://www.gwg.org). All Rights Reserved. The Ghent 
Workgroup hereby grants permission to use this test suite and its documentation as described 
in the associated documentation, subject to the following conditions. This legal notice must be 
included in all copies containing the whole or substantial portions of the Ghent PDF Output 
Suite. Without express written permission of the Ghent Workgroup it is not permitted to use this 
PDF Output Suite for anything but its intended purpose of testing workflow setup. The Ghent 
PDF Output Suite cannot be sold or used in any commercial context without previous written 
permission by the Ghent Workgroup.

So it's a bit tricky, the copyright notice would have to be added too somehow, so I guess it's better to use the link.

timvandermeij: Yes, let's indeed use a linked test case.

src/core/stream.js
THausherr: Re squashing commits, this will take some time. I agree that you want to keep your commit history clean of "oops"-type commits. If I understand it properly, this isn't possible from the website. I'll first try from my IDE, and if that doesn't do it, I'll have to install a command line client :-(

THausherr: Surprisingly this didn't take hours or days. Thanks for the link and for your help.

test/test_manifest.json
```diff
@@ -706,6 +706,14 @@
        "lastPage": 1,
        "type": "eq"
     },
+    {  "id": "issue6289",
+       "file": "pdfs/issue6289.pdf",
+       "md5": "0869F3D147C734EC484FFD492104095D",
```
timvandermeij: This should be `0869f3d147c734ec484ffd492104095d` according to the test run logs (i.e., no capital letters).

timvandermeij (approved):
Good to merge after the MD5 change (make sure the commit is squashed again after the change).

timvandermeij:

PR #8928, opened 2017-09-19
Fix #8907: Decode URL to get correct path in node_stream.

### LINKED ISSUE #8907: ENOENT: no such file or directory, lstat
Since #8712 was introduced, opening a local file with spaces generates an error when PDFJS is executed in an Node environment (Electron in my case).

Configuration:
Electron 1.7.5
Fedora 25
- PDF.js version:
1.9.562

1. Open a locale file with spaces `file:///home/njakob/Foo%20Bar/document.pdf`

What went wrong?
The error `ENOENT: no such file or directory, lstat` is raised.

After digging into the code, when `url.parse` is [used](https://github.com/mozilla/pdf.js/blob/master/src/display/node_stream.js#L31), the path is still encoded which certainly do not please `fs` [calls](https://github.com/mozilla/pdf.js/blob/master/src/display/node_stream.js#L364).

### LINKED ISSUE #8946: Failure to open url with "C:\" in url
Configuration:
Electron 1.6
Windows 10
- PDF.js version:
(These are pdfjs-dist)
Works: v1.9.426
Doesn't work: Latest - eg: 1.9.583
- Is an extension: No

1. Try to open a file url: "file://C:\Users\John Doe\test.pdf"
PDFJS.getDocument( "file://C:\Users\John Doe\test.pdf")... 
2.ENOENT: no such file or directory, lstat 'C:\C:\Users\John%20Doe\test.pdf"

I've spent a bunch of time on this, and I think there are two issues:
1) the encoded space, which is fixed in 8907
2) the file://C:\... gets transformed to file:///C:\... (extra slash), and then it later sees the / for the filesystem, and adds another c:\ to the front.

Note - this whole file://c:\... works as expected in v1.9.426 (even with the encoded space).

mukulmishra18:
Fixes #8907
Fixes #8946

### DIFF
--- a/src/display/node_stream.js
+++ b/src/display/node_stream.js
@@ -360,8 +360,9 @@ class PDFNodeStreamRangeReader extends BaseRangeReader {
 class PDFNodeStreamFsFullReader extends BaseFullReader {
   constructor(stream) {
     super(stream);
+    let path = decodeURI(this._url.path);
 
-    fs.lstat(this._url.path, (error, stat) => {
+    fs.lstat(path, (error, stat) => {
       if (error) {
         this._errored = true;
         this._reason = error;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
The form failures on the Linux bot are finally resolved after an update. Making new reference images, but that has nothing to do with this patch directly.

/botio-linux makeref

timvandermeij:
Thank you!

PR #8931, opened 2017-09-19
Re-factor `PDFFunction` to be class-like, and introduce a `classFactory` in `PDFDocument` that lazily creates an instance of `PDFFunction`

Snuffleupagus:
*Follow-up to PR https://github.com/mozilla/pdf.js/pull/8909#issuecomment-330251733.*

This requires us to pass around `classFactory` to quite a lot of existing code, however I don't see another way of handling this while still guaranteeing that we can access `PDFFunction` as freely as in the old code.

Please note that the patch passes all tests locally (unit, font, reference), and I *very* much hope that we have sufficient test-coverage for the code in question to catch any typos/mistakes in the re-factoring.

/cc @yurydelendik

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -36,8 +36,8 @@ var Page = (function PageClosure() {
            (intent === 'print' && annotation.printable);
   }
 
-  function Page(pdfManager, xref, pageIndex, pageDict, ref, fontCache,
-                builtInCMapCache) {
+  function Page({ pdfManager, xref, pageIndex, pageDict, ref, fontCache,
+                  builtInCMapCache, classFactory, }) {
     this.pdfManager = pdfManager;
     this.pageIndex = pageIndex;
     this.pageDict = pageDict;
@@ -45,6 +45,7 @@ var Page = (function PageClosure() {
     this.ref = ref;
     this.fontCache = fontCache;
     this.builtInCMapCache = builtInCMapCache;
+    this.classFactory = classFactory;
     this.evaluatorOptions = pdfManager.evaluatorOptions;
     this.resourcesPromise = null;
 
@@ -215,6 +216,7 @@ var Page = (function PageClosure() {
         fontCache: this.fontCache,
         builtInCMapCache: this.builtInCMapCache,
         options: this.evaluatorOptions,
+        classFactory: this.classFactory,
       });
 
       var dataPromises = Promise.all([contentStreamPromise, resourcesPromise]);
@@ -290,6 +292,7 @@ var Page = (function PageClosure() {
           fontCache: this.fontCache,
           builtInCMapCache: this.builtInCMapCache,
           options: this.evaluatorOptions,
+          classFactory: this.classFactory,
         });
 
         return partialEvaluator.getTextContent({
@@ -361,6 +364,19 @@ var PDFDocument = (function PDFDocumentClosure() {
     this.pdfManager = pdfManager;
     this.stream = stream;
     this.xref = new XRef(stream, pdfManager);
+
+    let pdfFunction = null;
+    this.classFactory = {
+      getPDFFunction() {
+        if (!pdfFunction) {
+          let evaluatorOptions = pdfManager.evaluatorOptions;
+          pdfFunction = new PDFFunction({
+            isEvalSupported: evaluatorOptions.isEvalSupported,
+          });
+        }
+        return pdfFunction;
+      },
+    };
   }
 
   function find(stream, needle, limit, backwards) {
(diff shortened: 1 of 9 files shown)


Snuffleupagus:
I just noticed that fair number of function calls in `ColorSpace` were accidentally missing the `PDFFunctionFactory` parameter, which has now been addressed.
However looking at the patch, in particular the code in `src/core/colorspace.js`, I'm starting to wonder if we shouldn't actually do a similar re-factoring of `ColorSpace` (by creating it lazily using a factory) to avoid having to clutter the code in this way?

yurydelendik:
> However looking at the patch, in particular the code in src/core/colorspace.js, I'm starting to wonder if we should attempt a similar re-factoring of ColorSpace (by creating it lazily using a factory) to avoid having to clutter the code in this way?

How likely this will cause a regression? Shall we uplift the patch?

Can you change PDFFunctionFactory to pdfFunctionFactory ?

Snuffleupagus:
> How likely this will cause a regression?

Having quickly looked at actually doing this, I think that the answer is that it seems too high a risk to do all at once (considering the size and scope of the necessary changes).

Shall we uplift the patch?

If you mean to Firefox 57, then I'd say that it's not necessary at all. Considering that this PR fixes a small (potential) inconsistency when multiple documents are opened, which doesn't happen in FIREFOX/MOZCENTRAL, we should be good here.

> Can you change PDFFunctionFactory to pdfFunctionFactory ?

Since I still would want to look into doing the similar conversion for `ColorSpace` later on, I figured that using the name `classFactory` could be good. Currently it only contains a `getPDFFunction` method, but in the future it could also hold `getColorSpace` (and possibly more).

Snuffleupagus:
*Please note:* The failures above are known intermittents.

@yurydelendik Gentle review ping; as described in https://github.com/mozilla/pdf.js/pull/8931#issuecomment-331534588, the only changes made since you last looked at this is:

> > Can you change PDFFunctionFactory to pdfFunctionFactory ?

> Since I still would want to look into doing a similar conversion for ColorSpace later on, I figured that using the name classFactory instead would be good. Currently it only contains a getPDFFunction method, but in the future it could also hold getColorSpace (and possibly more).

yurydelendik (commented):
classFactory does not really look like a factory (due to the singelton/function). I recommend to change interface to be more factory-ish, e.g.

```
class PDFFunctionFactory {
  create(fn) { via .parse(fn) }
  createFromArray(array) { via .parseArray(array) }
  createFromIR(ir) { via .fromIR(ir) }
}
```

src/core/document.js
```diff
@@ -37,14 +37,15 @@ var Page = (function PageClosure() {
   }
 
   function Page(pdfManager, xref, pageIndex, pageDict, ref, fontCache,
```
yurydelendik: can we change `classFactory` to `pdfFunctionFactory` and also pass parameters as an object

Snuffleupagus:
> classFactory does not really look like a factory (due to the singelton/function). I recommend to change interface to be more factory-ish, e.g.
> 
> class PDFFunctionFactory {
>   create(fn) { via .parse(fn) }
>   createFromArray(array) { via .parseArray(array) }
>   createFromIR(ir) { via .fromIR(ir) }
> }

I'm not entirely sure that I understand completely what you're asking for here, and it feels like I'm just having problems "connecting the dots" so to speak.
To try and aid the discussion, I've pushed a *partial* and thus non-working `WIP` commit (to avoid doing a lot of work in the wrong direction), and I'd ask that you please see if I got the gist of what you asked for. If not, I'd *very much* appreciate a few additional pointers!

yurydelendik:
Sorry, by looking at our PDFFunction,  I can see why it can be confusing -- it's play kind of role of a factory, but it has not clear definition. So we can refactor/rename PDFFunction to be PDFFunctionFactory and its method can have 'createXXXX' form as I mentioned above. The `new PDFFunctionFactory(isEvalSupported)` will create an instance of this factory, and this can be directly used as pdfFunctionFactory (without _getInstance). I hope this explains that a bit.

Snuffleupagus:
@yurydelendik Is https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:PDFFunctionFactory-2 closer to what you want?

yurydelendik:
>  Is master...Snuffleupagus:PDFFunctionFactory-2 closer to what you had in mind?

It is what I had in mind. Thank you for making this happen.

Snuffleupagus:
> It is what I had in mind. Thank you for making this happen.

Thanks for all the feedback :-)
Closing as superseded by PR #8968.

PR #8932, opened 2017-09-19
JBIG2 symbol offsets

janpe2:
JBIG2 text regions have a symbol positioning problem:
symbol_pos_problem
The top pixel row of symbols is missing. The row goes outside the text region because symbols are positioned one pixel too high. Here is another one:
symbol_pos_problem_2
[jbig2_symbol_offset.pdf](https://github.com/mozilla/pdf.js/files/1314680/jbig2_symbol_offset.pdf)

In JBIG2 the `REFCORNER` value specifies which of the four corners of a symbol bitmap is positioned at the desired point (x, y). In practice (and in PDF.js) bitmaps are drawn starting from the left-top corner. If `REFCORNER` is other than left-top, some offsets are needed. Currently the JBIG2 decoder uses offsets that are one pixel too large. Thus the issue appears. Most often `REFCORNER` is left-bottom, like in my PDFs above, so only the vertical position is wrong.

When symbols are drawn, the correct offsets are _one less_ than the symbol width and height. For example, the vertical offset when `REFCORNER` is left-bottom:
symbol_offset

Typically JBIG2 images have a high resolution, so one pixel errors in symbol positioning can't be easily seen. (And symbols are rarely so near the edge of the text region that they would be clipped.)

This PR may cause JBIG2-related regression tests to fail because symbols will be moved one pixel. The new rendering is the correct one.


Snuffleupagus:
> Conflicting files
> test/pdfs/.gitignore
> test/test_manifest.json

Could you please resolve these merge conflicts, so that we can test this patch? :-)
*A small piece of advice (based on experience):* To avoid having to rebase patches a lot because of merge conflict in those two files in particular, try adding new test entries in more "random" positions rather that at the *end* of the files.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #8938, opened 2017-09-20
Use font's default width even when 0. (bug 1392647)

brendandahl:
Bug 1392647 has a PDF where the default width of the font
is 0. It draws some charcodes that don't have glyphs, but
we were wrongly using the 1000 default width for these
charcodes causing some text to be overlapping.


Snuffleupagus (approved):
That was delightfully simple, thanks for fixing this!
r=me, assuming all tests pass of course :-)

Issue #8935, opened 2017-09-19
Multiple spaces in PDF being squashed to single space in iOS during SVG conversion


SnailCoil:
Link to PDF file (or attach file here): 
[double_spaces.pdf](https://github.com/mozilla/pdf.js/files/1315807/double_spaces.pdf)

Configuration:
- Web browser and its version: AppleWebKit/603.3.8 Version/10.0 Mobile/14G60 Safari/602.1
- Operating system and its version: iOS 10.3.3
- PDF.js version: pdfjs-dist@1.9.575
- Is an extension: I'm not sure the answer to this. I'm using a small wrapper library in my web app that's specifically for Angular 2+, https://github.com/VadimDez/ng2-pdf-viewer

Steps to reproduce the problem:
1. Convert the attached PDF, or any PDF that contains multiple spaces in a row, to SVG
2. Observe multiple spaces being squashed to a single space

What is the expected behavior? (add screenshot)
![chrome_desktop_spaces](https://user-images.githubusercontent.com/379022/30615466-ed2bc606-9d43-11e7-9e24-7a275e5b4446.png)
Here is the PDF generated on desktop. The double spaces in the PDF are mirrored here.

What went wrong? (add screenshot)
![safari_ios_spaces](https://user-images.githubusercontent.com/379022/30615782-dd339f52-9d44-11e7-9294-f0a2a091e1fa.png)

iOS is automatically squashing multiple spaces to a single space. I verified that the SVG converter generated the correct glyphs, but when the correct string ("double&nbsp;&nbsp;spaces&nbsp;&nbsp;double&nbsp;&nbsp;spaces") was assigned to the svg element's text content, the double spaces disappeared. The reason for this is logical: multiple spaces in HTML are considered to be no different than a single space. 

However, desktop chrome conveniently ignores this, and renders multiple spaces with no problem. Is there a way that we can trick iOS into inserting the multiple spaces? I wondered if we could use non-breaking spaces in place of spaces to get the same functionality...

yurydelendik:
> multiple spaces in HTML are considered to be no different than a single space.

SVG is an XML format. `xml:space="preserve"` shall preserve the spaces.

SnailCoil:
Perhaps it could be a symptom of https://bugs.webkit.org/show_bug.cgi?id=112032

yurydelendik:
> Perhaps it could be a symptom of https://bugs.webkit.org/show_bug.cgi?id=112032

[...]

PR #8943, opened 2017-09-21
work around webkit bug that coalesces multiple spaces

AaronSherwoodForest:
https://bugs.webkit.org/show_bug.cgi?id=112032
https://github.com/mozilla/pdf.js/issues/8935

I ran lint and the tests locally, but 1/465 unit tests is failing whether I use "master" or my feature branch.


Snuffleupagus (commented):
Generally we do not add browser/environment specific compatibility hacks in core code. The reason is first of all that this would often have a negative impact on performance, even in browsers where the bug isn't present. Secondly, it often also hurts readability/maintainability of the code in question.

In a custom implementation of PDF.js you can obviously do this, but personally I'm not convinced that the patch would be a good idea in the PDF.js repo (at least not in its current form).

yurydelendik:
Is this solution complete? Are we sure that safari only removes double spaces and not touching leading spaces?

SnailCoil:
Thanks yury, I think your concern was spot on. I updated the logic.

Snuffleupagus, I completely understand that reasoning. I'm content to close the PR and linked issue, and people will be able to follow the issue to see this specific workaround, since this should be fixed in webkit. However, you mentioned "(at least not in its current form)"--is there a reasonable way to rework this PR to make it optional for people who need to work around the webkit bug?

timvandermeij:
I'm going to close this pull request, but we can keep the issue open for tracking purposes. Indeed we do not wish to have browser-specific hacks in the core code. Custom deployments may add this, but I don't think we should add it in PDF.js repository.

PR #8950, opened 2017-09-23
Implement support for polyline and polygon annotations

timvandermeij:


Snuffleupagus (approved):
Seems fine, thank you for the patch!

Issue #2072, opened 2012-09-04
Outline (bookmark) pane should be resizable.


zhyi:
The outline pane always has a fixed width. Though a horizontal scroll bar is provided, it is still somehow inconvenient.

Furthermore, there is currently no way to fully display a long bookmark name. I think it can be solved by showing a tip (something like title="") when the mouse is hovered above such a bookmark.

MetaMemoryT:
:+1: this is a very important issue

ppradeepk:
:+1: 

I am not sure why this doesn't have the attention it  deserves.

Daraius:
Yes, this would be my #1 Feature Request.  Resizable Bookmarks pane.
Wrapping the text allows reading it but makes the list super long to scroll around in.  So perhaps an option to wrap or not.  But resizable is highly desirable.
Thanks - Daraius

nortexoid:
Still no word on this? Would be a very nice improvement to the current viewer, especially for folk using Chrome OS where there aren't really any other options available.

le-jzr:
Any word on this? Seriously, I've taken to just editing the CSS properties by hand on most documents I open. Nobody gives a rat's ass about prettiness of animations when the current state is just flat out unusable.

timvandermeij:
The viewer is now more responsive (and the code more modular) than it was at the time this issue was filed, so this should be a bit easier to fix now. If someone wants to work on this, please feel free to do so. I'm marking this as a good beginner bug because it does not require in-depth PDF.js knowledge to fix and so it gets more attention from developers that want to start working on PDF.js.

@le-jzr I really want to emphasize that most contributors are working on PDF.js in their spare time and therefore issues have to be prioritized. Please either help out by a) contributing a patch, b) providing constructive feedback or c) upvoting the initial comment of the issue to help raise its priority. Bumping the issue with such a comment is not going to fix the issue any sooner. All contributors do whatever they can to improve PDF.js, so please keep that in mind when writing comments. Thank you.

le-jzr:
Sorry for my lack of manners. While I don't have any experience whatsoever in matters of browser layouts and JavaScript, I would like to suggest that it's not necessary to start with completely interactive solution.

Say, a query string like `?sidebar=500px` might be a good start.

timvandermeij:
Closing since this is implemented by the pull request above.

Issue #4052, opened 2013-12-22
Tweak the vertical position of the sidebar


Snuffleupagus:
Currently the sidebar is positioned differently vertically depending on the width of the viewer.
- When the width of the viewer is _above_ 770 pixels:
  ![large](https://f.cloud.github.com/assets/2692120/1796925/940b7302-6a96-11e3-8830-e0bac5de87cb.png)
- When the width of the viewer is _below_ 770 pixels:
  ![small](https://f.cloud.github.com/assets/2692120/1796927/9fa4ec48-6a96-11e3-8aad-17a79125460b.png)

I think that it would be much more consistent if the sidebar was placed in the same vertical position, regardless of the size of the viewer. I.e. looking something like this at widths above 770 pixels (compare with the first image above):
![new](https://f.cloud.github.com/assets/2692120/1796967/4aa9e6f4-6a9a-11e3-9370-e5342044b4e4.png)

It has always bothered me somewhat that the width of the toolbar changes when the sidebar is toggled.
As an added bonus, _if_ this change is made, the CSS rules that control which buttons are visible in the toolbar could be simplified a bit.

Also, changing the vertical placement of the sidebar would probably address [bug 850591](https://bugzilla.mozilla.org/show_bug.cgi?id=850591) at the same time.

/cc @brendandahl, @yurydelendik What are your opinions on this?

**EDIT:** I assume that if this should be changed, we would likely need sign-off from UI/UX as well.

Snuffleupagus:
WIP patch, which implements the above and cleans up some related CSS, available at https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:sidebar-constant-position.

Snuffleupagus:
WIP patch, which implements the above and cleans up some related CSS, available at https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:sidebar-constant-position.
I think that we need to do this *first*, before attempting to implement any sidebar resizing since otherwise trying to dynamically update all the affected media queries would be a total mess!

PR #8957, opened 2017-09-26
Resize sidebar

ironyman:
Addresses https://github.com/mozilla/pdf.js/issues/2072.


timvandermeij:
I like the direction of this patch and it already works quite well. Some comments from playing with this a bit:

- Let's restrict the sidebar width to 50% of the total screen width. Now it's possible to even resize until beyond 100%, which should not happen. After more than 50%, the PDF file itself becomes too small, so I think 50% would be a good maximum value.

- I think the timeout until the view re-renders after resizing is a bit too large, which makes the resizing feel slow. If you resize the sidebar quickly, the cursor can end up behind the page itself until it's re-rendered.

ironyman:
I addressed both of those issues.
But I noticed another problem. 
2017-09-26-222312_1037x146_scrot
toolbarViewerMiddle gets pushed off the toolbar when there's not enough space. What should we do about this?

Snuffleupagus:
In addition to https://github.com/mozilla/pdf.js/pull/8957#issuecomment-332336558, there's also a couple of other issues:
 - Most importantly, since it breaks things quite badly, is that if you resize the sidebar then the viewer will look *completely* broken after the sidebar has been closed.

 - I'm finding the code somewhat difficult to follow, and my worry is that this would be very hard to actually maintain in the future. Considering that the added JS code is making *a lot* of assumptions about the structure of the HTML/CSS, making any changes down the line there would thus be difficult.

(Presumably the reason why this functionality hasn't been implemented yet, is that the way e.g. the sidebar CSS is currently written makes this very difficult to fix in a good way.)

ironyman:
Are we OK with rewriting the sidebar CSS using flexbox? That should eliminate the need for hard coded width on every sub element of sidebarContainer.

web/pdf_sidebar.js
```diff
@@ -454,6 +460,81 @@ class PDFSidebar {
...
+
+  resize(width) {
+    if (width < 200 || width > this.mainContainer.offsetWidth) {
+      return;
+    }
+
+    // This is the media query in viewer.css, which
+    // sets #mainContainer left to 0 on small screens.
+    // We don't want to override that.
+    var mql = window.matchMedia('all and (max-width: 840px');
```
yurydelendik: Per http://caniuse.com/#search=matchMedia most of browsers we are supporting are not mentioned.

yurydelendik:
> Are we OK with rewriting the sidebar CSS using flexbox?

Maybe. But keep in mind old browser support, in worst case we need to gracefully degrade the functionality.

Is there a way to generate "global" CSS to override default size of the sidebar? And resizer may just "control" this CSS.

IMHO the sidebar shall not depend and control its parents; a resizer alone may notify (via dispatch event) that it's position changed and app will adjust its UI then.

yurydelendik (changes requested):
move pdf_sidebar_resizer into its own file.

Snuffleupagus:
> Are we OK with rewriting the sidebar CSS using flexbox?

Considering that we've not yet officially rescinded support for older IE versions, that may not work too well; see e.g. the following compatibility data https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes#Browser_compatibility.

Also, looking again at the screen-shot in https://github.com/mozilla/pdf.js/pull/8957#issuecomment-332390324,  I do wonder if we'd need to try and reach consensus on and implement #4052 first. Otherwise I'm not sure how you'd ever be able to dynamically update all of the media-queries involved!?

brendandahl:
> Are we OK with rewriting the sidebar CSS using flexbox?

It's probably about time we re-visit our supported browser/version. The can-i-use for flex box is around 97%.

FWIW: the original mock-up of the PDF viewer was done in flexbox and it was SO much cleaner than what we ended up with. Unfortunately, the site doesn't work any more http://people.mozilla.org/~shorlander/mockups/PDF.js/PDFjs-Tour-01/index.html

brendandahl:
Shorlander moved the original mock http://design.firefox.com/people/shorlander/pdfjs/prototype/pdfjs-ui-demo-i01.html

ironyman:
I noticed some weirdness in the code, the naming here is pretty confusing, should we fix this?

https://github.com/mozilla/pdf.js/blob/master/web/viewer.js#L60

timvandermeij:
Closing since this is superseded by #9026. This solution is a bit more complicated than that one. Nevertheless, thank you for working on this as it provided us with ideas on how to improve the current situation and implement this feature more quickly!

Issue #8937, opened 2017-09-20
Trouble with opening PDFs with non-ASCII characters


wojtekmaj:
Link to PDF file (or attach file here):
[23.pdf](https://github.com/mozilla/pdf.js/files/1318569/23.pdf)

Configuration:
- Web browser and its version: Any
- Operating system and its version: Windows 10
- PDF.js version: 1.9.577, likely does not matter
- Is an extension: No

What is the expected behavior? (add screenshot)

I have a package that makes a bridge between pdf.js and React, [react-pdf](https://github.com/wojtekmaj/react-pdf). Recently I was informed that the file attached above does not render correctly.
I tested the file in my project and in pdf.js online viewer to confirm. It did not work on my project, worked correctly on online viewer. However, I've noticed that online viewer makes additional server requests while opening the file, which fail on my project.

* What is the file it's trying to load?
* Can I somehow make it work with Webpack without the need to force my users to manually add the files to their projects?
* And finally, why baseUrl was null in the first place and why the request was not stopped because of it?

What went wrong? (add screenshot)

![image](https://user-images.githubusercontent.com/5426427/30657405-ce25d892-9e37-11e7-99f3-7fb897991655.png)

![image](https://user-images.githubusercontent.com/5426427/30657391-c6a790f6-9e37-11e7-8b6b-6fada73d33af.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

http://projekty.wojtekmaj.pl/react-pdf/test

yurydelendik:
Here is what we do to configure demo viewer https://github.com/mozilla/pdf.js/blob/master/web/app.js#L51-L64

Try specifying cMapUrl/cMapPacked globals.

yurydelendik:
> What is the file it's trying to load?

External CMap files that can be referred from some PDFs (similar to non-embedded fonts). They are not included into the main code due to their size

>  Can I somehow make it work with Webpack without the need to force my users to manually add the files to their projects?

The files needs to be present, so PDF.js can lazily fetch them when need. Sorry, we don't have JSONP option yet (from CDN). Somebody welcome to work on it.

> And finally, why baseUrl was null in the first place and why the request was not stopped because of it?

No reason. Adding the null-check before issuing a request or, even better, using "always fails" CMapReaderFactory will be a good addition to the PDF.js core library.

jamiebikies:
[...]

PR #8962, opened 2017-09-28
Check that `this.baseUrl` is defined before attempting to fetch any data in `DOMCMapReaderFactory`/`NodeCMapReaderFactory`

Snuffleupagus:
*Excerpt from https://github.com/mozilla/pdf.js/issues/8937#issuecomment-330941932:*

> > And finally, why baseUrl was null in the first place and why the request was not stopped because of it?

> No reason. Adding the null-check before issuing a request or, even better, using "always fails" CMapReaderFactory will be a good addition to the PDF.js core library.

This PR implements the additional validation suggested above, since that seems like a good idea regardless of if/when a specialized factory is added.

### DIFF
--- a/src/display/dom_utils.js
+++ b/src/display/dom_utils.js
@@ -68,6 +68,10 @@ class DOMCMapReaderFactory {
   }
 
   fetch({ name, }) {
+    if (!this.baseUrl) {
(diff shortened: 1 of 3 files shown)


yurydelendik:
Thank you for the patch.

PR #8968, opened 2017-09-28
Split the existing `PDFFunction` in two classes, a private `PDFFunction` and a public `PDFFunctionFactory``, and utilize the latter in `PDFDocument` to allow various code to access the methods of `PDFFunction`

Snuffleupagus:
*Follow-up to PR #8909.*

This requires us to pass around `pdfFunctionFactory` to quite a lot of existing code, however I don't see another way of handling this while still guaranteeing that we can access `PDFFunction` as freely as in the old code.

Please note that the patch passes all tests locally (unit, font, reference), and I *very* much hope that we have sufficient test-coverage for the code in question to catch any typos/mistakes in the re-factoring.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -15,7 +15,6 @@
 
 import { FormatError, info, isString, shadow, warn } from '../shared/util';
 import { isDict, isName, isStream } from './primitives';
-import { PDFFunction } from './function';
 
 var ColorSpace = (function ColorSpaceClosure() {
   /**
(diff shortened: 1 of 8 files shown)


src/core/obj.js
```diff
@@ -28,9 +28,10 @@ import { CipherTransformFactory } from './crypto';
 import { ColorSpace } from './colorspace';
 
 var Catalog = (function CatalogClosure() {
-  function Catalog(pdfManager, xref, pageFactory) {
+  function Catalog(pdfManager, xref, pageFactory, pdfFunctionFactory) {
     this.pdfManager = pdfManager;
     this.xref = xref;
+    this.pdfFunctionFactory = pdfFunctionFactory;
```
yurydelendik: I cannot spot why we need this property here.

yurydelendik (approved):
Looks good with the question below answered.

src/core/obj.js
Snuffleupagus: > I cannot spot why we need this property here.

We don't, that's just a case of a bit too much copy-pasting. Removed now, thanks!

PR #8976, opened 2017-09-29
Simplify the disabling of Node polyfills, in `gulpfile.js`, since we're now using Webpack 3.x

Snuffleupagus:
Webpack was updated a while back, but we apparently forgot to update `gulpfile.js` as well.


timvandermeij (approved):

timvandermeij:
Thank you for fixing this oversight!

PR #8977, opened 2017-09-29
Update all packages to the most recent version

### LINKED ISSUE #8970: Consider updating to worker-loader 1.0
Hey,

just a quick heads-up - worker-loader, one of your dependencies, has just released version 1.0.0. You might want to use it to avoid package duplication for folks that are keeping their dependencies up to date.

Have a good day!

timvandermeij:
To support this, the following changes have to be made as well:

- Ignore `package-lock.json` since NPM creates it, but we should not have it in the repository.
- Switch from `babel-preset-es2015` to `babel-preset-env` to resolve the deprecation warning in the test logs. The latter is more recent and flexible, but should be the same functionality-wise.
- `transform` now needs to have the `utf-8` encoding option provided. If not given, it will call the callback with a `Buffer` object, which results in an unhandled promise rejection since what is returned from the callback is a string, not a `Buffer`.

Multiple test log warnings are resolved by this patch.

Fixes #8970


gulpfile.js
```diff
@@ -152,7 +152,7 @@ function createWebpackConfig(defines, output) {
           loader: 'babel-loader',
           exclude: /src\/core\/(glyphlist|unicode)/, // babel is too slow
           options: {
-            presets: pdfjsNext ? undefined : ['es2015'],
+            presets: pdfjsNext ? undefined : ['env'],
```
Snuffleupagus: The same change must be done at https://github.com/timvandermeij/pdf.js/blob/98b9d5ec37398bcd1eeb784f2b220f57bb9cabb8/gulpfile.js#L1002 as well.

timvandermeij: Done in the new commit.

Snuffleupagus (approved):
This seems to work fine, when testing quickly locally.
r=me, with the following question answered:

Did you run a couple of different `gulp` commands (e.g. generic/firefox/mozcentral/...) locally with/without this patch, just to check that nothing obvious broke with the new package versions?

*I'm assuming that you did, but just wanted to double check :-)*

Feel free to land this, thank you!

timvandermeij:
Thank you for the review! Yes, the different `gulp` commands produced the exact same output before and after the patch.

PR #8979, opened 2017-09-30
Transform Web Archive URLs to avoid downloading an HTML page instead of the PDF file

### LINKED ISSUE #8412: Errors during running makeref

Configuration:
- Web browser and its version: NA
- Operating system and its version: MacOS X 
- PDF.js version: master
- Is an extension: no

1. Follow the contributing instructions
2. Run makeref

All pdfs are downloaded successfully?

```
WARNING: File was not downloaded. See "pdfs/issue3999.pdf.error" file.
WARNING: File was not downloaded. See "pdfs/hmm.pdf.error" file.
WARNING: File was not downloaded. See "pdfs/wdsg_fitc.pdf.error" file.
WARNING: File was not downloaded. See "pdfs/geothermal.pdf.error" file.
WARNING: File was not downloaded. See "pdfs/issue5752.pdf.error" file.
WARNING: File was not downloaded. See "pdfs/issue6117.pdf.error" file.
WARNING: File was not downloaded. See "pdfs/issue2442.pdf.error" file.
WARNING: File was not downloaded. See "pdfs/bug887152.pdf.error" file.
```

### LINKED ISSUE #8920: All Internet Archive links, used in reference testing, are broken
Apparently the [Internet Archive](http://archive.org/web/), which we depend on for a *very* large number of (linked) reference test-cases, has recently changed how they serve PDF files.

Previously, a URL such as http://web.archive.org/web/20160112115354/http://www.fao.org/fileadmin/user_upload/tci/docs/2_About%20Stacks.pdf would return a PDF file directly. However, now a HTML file is returned instead (which then points to the actual PDF file).
For someone cloning the PDF.js repo, and attempting to set-up testing for the first time, this means that *all* linked test-cases will now fail. Furthermore, it also means that we cannot use the Internet Archive when adding new test-cases.

Since the HTML file returned *does* contain a direct link to the PDF file, embedded in an `<iframe>` tag, we could perhaps add special-casing for Internet Archive URLs in [test/downloadutils.js](https://github.com/mozilla/pdf.js/blob/master/test/downloadutils.js), such that the HTML file is first downloaded and parsed to obtain a direct PDF link.

timvandermeij:
Moreover, adjust one linked test case that did not conform to the standard Web Archive URL format and adjust one linked test case because the link was dead.

Fixes #8920.
Fixes #8412.

### DIFF
--- a/test/downloadutils.js
+++ b/test/downloadutils.js
@@ -22,7 +22,22 @@ var crypto = require('crypto');
 var http = require('http');
 var https = require('https');
 
+function rewriteWebArchiveUrl(url) {
+  // Web Archive URLs need to be transformed to add `if_` after the ID.
+  // Without this, an HTML page containing an iframe with the PDF file
+  // will be served instead (issue 8920).
+  var webArchiveRegex =
+    /(^https?:\/\/web\.archive\.org\/web\/)(\d+)(\/https?:\/\/.+)/g;
+  var urlParts = webArchiveRegex.exec(url);
+  if (urlParts) {
+    return urlParts[1] + (urlParts[2] + 'if_') + urlParts[3];
+  }
+  return url;
+}
+
 function downloadFile(file, url, callback, redirects) {
+  url = rewriteWebArchiveUrl(url);
+
   var completed = false;
   var protocol = /^https:\/\//.test(url) ? https : http;
   protocol.get(url, function (response) {
--- a/test/pdfs/geothermal.pdf.link
+++ b/test/pdfs/geothermal.pdf.link
@@ -1,1 +1,1 @@
-http://web.archive.org/web/20150212141833/http://geothermal.inel.gov/publications/future_of_geothermal_energy.pdf
+https://web.archive.org/web/20170930174755/https://www.pdf-archive.com/2017/09/30/future-of-geothermal-energy/future-of-geothermal-energy.pdf
--- a/test/pdfs/issue8169.pdf.link
+++ b/test/pdfs/issue8169.pdf.link
@@ -1,1 +1,1 @@
-http://web.archive.org/save/_embed/http://210.243.166.143/prob1.pdf
+https://web.archive.org/web/20170930161657/http://210.243.166.143/prob1.pdf


SamyCookie:
Hi, I tried tests locally just now and got this : 
```
Downloading http://web.archive.org/save/_embed/http://210.243.166.143/prob1.pdf to pdfs/issue8169.pdf...
/home/samuel/src/pdf.js/test/downloadutils.js:34
    var id = regex.exec(url)[0];
                            ^
```

test/downloadutils.js
```diff
@@ -25,6 +25,16 @@ var https = require('https');
 function downloadFile(file, url, callback, redirects) {
   var completed = false;
   var protocol = /^https:\/\//.test(url) ? https : http;
+
+  if (url.indexOf('web.archive.org') !== -1) {
+    // Web Archive URLs need to be transformed to add `if_` after the ID.
+    // Without this, an HTML page containing an iframe with the PDF file
+    // will be served instead (issue 8920).
+    var regex = /web\.archive\.org\/web\/(\d+)/g;
```
SamyCookie: minor optimization, make `regex` global outside `downloadFile` to prevent to compile it each time ?

Snuffleupagus: Won't this regex match already rewritten URLs as well?
Also, I'd like us to be a bit more specific about what URL format we expect; more elsewhere.

test/downloadutils.js
```diff
@@ -25,6 +25,16 @@ var https = require('https');
...
+    var id = regex.exec(url)[0];
+    url = url.replace(regex, id + 'if_');
```
Snuffleupagus: Isn't it sufficient to just do `exec`, i.e. couldn't you just assemble the URL manually based on the result from a more "complete" regex?

test/downloadutils.js
```diff
@@ -25,6 +25,16 @@ var https = require('https');
 function downloadFile(file, url, callback, redirects) {
   var completed = false;
   var protocol = /^https:\/\//.test(url) ? https : http;
+
+  if (url.indexOf('web.archive.org') !== -1) {
+    // Web Archive URLs need to be transformed to add `if_` after the ID.
+    // Without this, an HTML page containing an iframe with the PDF file
+    // will be served instead (issue 8920).
+    var regex = /web\.archive\.org\/web\/(\d+)/g;
+    var id = regex.exec(url)[0];
```
Snuffleupagus: If the regex doesn't match anything, this will break (because `regex.exec(url) === null`).

Snuffleupagus (commented):
Thanks for doing this, but I'd like us to be a bit more explicit about what URLs we attempt to rewrite (and have a bit more validation as well). Could we e.g. do something along these lines?
```diff
diff --git a/test/downloadutils.js b/test/downloadutils.js
index ef145416..f1b977fd 100644
--- a/test/downloadutils.js
+++ b/test/downloadutils.js
@@ -22,7 +22,19 @@ var crypto = require('crypto');
 var http = require('http');
 var https = require('https');
 
+function rewriteInternetArchiveUrl(url) {
+  var urlParts =
+    /(^https?:\/\/web\.archive\.org\/web\/)(\d+)(\/https?:\/\/.+)/g.exec(url);
+  if (urlParts) {
+    return urlParts[1] + (urlParts[2] + 'if_') + urlParts[3];
+  }
+  return url;
+}
+
 function downloadFile(file, url, callback, redirects) {
+  // Comment here...
+  url = rewriteInternetArchiveUrl(url);
+
   var completed = false;
   var protocol = /^https:\/\//.test(url) ? https : http;
   protocol.get(url, function (response) {
```

timvandermeij:
The comments are fixed in the new commit, but I'm still doing a full run for verification.

SamyCookie:
No JS error for me now, but some "PDF file" downloaded are still HTML pages :(

timvandermeij:
Yes, I notice that too. For example `jai.pdf`, which gives `null` for `urlParts` even though it works properly if I execute it manually in Firefox... I'm looking into it.

timvandermeij:
I printed all the transformed URLs and noticed that every second URL is not edited. Here is a part of the output:

<pre>
https://web.archive.org/web/20151019020657if_/https://www.erowid.org/archive/rhodium/chemistry/3base/piperonal.pepper/piperine.pepper/465e03piperine.pdf
http://web.archive.org/web/20150922201828/http://w2.eff.org/legal/cases/betamax/betamax_oral_argument2.pdf
http://web.archive.org/web/20160125220711if_/http://www.star.bnl.gov/public/daq/HARDWARE/21264_data_sheet.pdf
http://web.archive.org/web/20120204112920/http://h20000.www2.hp.com/bc/docs/support/SupportManual/bpl13210/bpl13210.pdf
https://web.archive.org/web/20150915102004if_/http://www.cplusplus.com/files/tutorial.pdf
http://web.archive.org/web/20150212141833/http://geothermal.inel.gov/publications/future_of_geothermal_energy.pdf
https://web.archive.org/web/20111027045653if_/http://agb.traviangames.com/Travian_AR_Terms.pdf
http://web.archive.org/web/20160213124006/http://www.leon.pl/userfiles/file/Zapytanie%20ofertowe%201-2014.pdf
https://web.archive.org/web/20170107214304if_/https://www.pdf-archive.com/2017/01/07/issue6071/issue6071.pdf
http://web.archive.org/web/20160112115354/http://www.fao.org/fileadmin/user_upload/tci/docs/2_About%20Stacks.pdf
http://web.archive.org/web/20160414174617if_/http://cache.lego.com/bigdownloads/buildinginstructions/6030672.pdf
https://web.archive.org/web/20170507102908/https://www.mbank.pl/download/firma/Dyspozycja-zmiany-typu-rachunku-biecego.pdf?noredir
</pre>

If I move the regex inside the function, everything works correctly. It looks like Node either mutates the regex or does not cache it correctly.

timvandermeij:
I just made a fresh clone and downloaded all the files with the new commit applied. All downloaded files passed the hash checks, so this patch is ready for another review.

Snuffleupagus (approved):
Works just fine in my testing (it just took a while to re-download all files to verify); thank you for fixing this!
Also, nice to have those two links updated!

PR #8980, opened 2017-09-30
Add CDN links

LukasDrgon:
I added [jsDelivr CDN links](https://www.jsdelivr.com/package/npm/pdfjs-dist) to your site and readme for easier and faster usage. jsDelivr is the fastest opensource CDN available and built specifically for production usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,13 @@ or to build it from source. We supply pre-built versions for usage with NPM and 
 the `pdfjs-dist` name. For more information and examples please refer to the
 [wiki page](https://github.com/mozilla/pdf.js/wiki/Setup-pdf.js-in-a-website) on this subject.
 
+## Including via a CDN
+
+PDF.js is hosted on several free CDNs:
+ - https://www.jsdelivr.com/package/npm/pdfjs-dist
+ - https://cdnjs.com/libraries/pdf.js
+ - https://unpkg.com/pdfjs-dist/
+
 ## Learning
 
 You can play with the PDF.js API directly from your browser using the live demos below:
--- a/docs/contents/getting_started/index.md
+++ b/docs/contents/getting_started/index.md
@@ -57,6 +57,13 @@ $ cd pdf.js
   </div>
 </div>
 
+## Including via a CDN
+
+PDF.js is hosted on several free CDNs:
+ - https://www.jsdelivr.com/package/npm/pdfjs-dist
+ - https://cdnjs.com/libraries/pdf.js
+ - https://unpkg.com/pdfjs-dist/
+
 ## File Layout Overview
 
 ### Prebuilt


Snuffleupagus:
I'm not sure that we, first of all, should advertise a *specific* CDN in this very visible way!?
Secondly, even if we do this, I don't really think that it belongs in https://mozilla.github.io/pdf.js/getting_started, since it should contain just the bare minimum.
Finally, if added to `README.md`, it shouldn't be placed in such a way that it takes precedence over the existing "Getting the Code" section (since that is the only thing we can really guarantee work).

LukasDrgon:
>I'm not sure that we, first of all, should (effectively) endorse a specific CDN in this very visible way!?

That is, of course, completely up to you. Most JS projects do provide a CDN link in their README/documentation because it makes using the code easier (and using a CDN is usually better than serving files from your server).

From user's POV, I can say that the getting started page is exactly where I would expect this, in addition to the full zip packages, but I'll be happy to modify this PR if you decide you actually want this *somewhere*.

jimaek:
Interestingly enough Mozilla has covered jsDelivr before :) https://hacks.mozilla.org/2014/03/jsdelivr-the-advanced-open-source-public-cdn/

yurydelendik:
What jsDeliver provides that cd.js or npmcdn does not? Does jsDeliver allow cors to access cmaps?

MartinKolarik:
@yurydelendik 
 - Multiple CDN providers, including one in China, which results in better load times.
 - Also multiple DNS providers, origin server locations, and backup storage system - it's visually explained [here](https://www.jsdelivr.com/network/infographic) and means that even if half of the services we use went down at the same time, jsDelivr would still continue to work.
 - Support for on-demand minification and bundling of multiple JS/CSS files. 
 - Usage statistics per project/version/file (some developers actually use this to drive certain development decisions).
 - Similarly to npmcdn, instant updates from npm with zero config, and semver version ranges for auto-updating URLs, such as `/npm/pdfjs-dist@1`

>Does jsDeliver allow cors to access cmaps?

Yes, cors is enabled for all resources.

yurydelendik:
> > Does jsDeliver allow cors to access cmaps?

> Yes, cors is enabled for all resources.

I test the https://gist.github.com/yurydelendik/016016dcb126ee5c84b4a9c3cc734d8e example. The CORS at jsDeliver does not work. I tested npmcdn.com it works there.

README.md
```diff
@@ -45,6 +45,13 @@ PDF.js is built into version 19+ of Firefox, however, one extension is still ava
 Chrome, go to `Tools > Extension` and load the (unpackaged) extension from the
 directory `build/chromium`.
 
+### Including via a CDN
+
+To use PDF.js via a CDN you need to include this file from [jsDelivr](https://www.jsdelivr.com/package/npm/pdfjs-dist):
+```
+<script src="https://cdn.jsdelivr.net/npm/pdfjs-dist@1/build/pdf.min.js"></script>
```
yurydelendik: We cannot exclusively list only jsdelivr, please list npmcdn.com prior jsdelivr.net since some finality (e.g. cmaps) does not work with the former.

docs/contents/getting_started/index.md
```diff
@@ -37,6 +37,12 @@ Before downloading PDF.js please take a moment to understand the different layer
 
 ## Download
 
+You can use PDF.js by including this file from [jsDelivr](https://www.jsdelivr.com/package/npm/pdfjs-dist):
```
yurydelendik: Remove the text from the download section, list it in the separate section below.

yurydelendik (changes requested):
We want to provide an alternative CDNs and don't want to be exclusive to the jsDelivr. Please rephrase text in neutral language and list available/popular CDNs that host PDF.js at this moment.

MartinKolarik:
I think you just used a wrong link there. https://www.jsdelivr.com/ is the website, CDN files are on a different domain - https://cdn.jsdelivr.net/npm/pdfjs-dist/cmaps/

yurydelendik:
> I think you just used a wrong link there. https://www.jsdelivr.com/ is the website, 

Sorry, I just copied what is/was provided at https://github.com/mozilla/pdf.js/pull/8980/commits/a8d73975d976d0d96bc1a2486489fbfbb9c108f7#diff-04c6e90faac2675aa89e2176d2eec7d8R50

LukasDrgon:
@yurydelendik Let me know if you want to change anything else.

yurydelendik (approved):

README.md
```diff
@@ -88,6 +88,13 @@ This will generate `pdf.js` and `pdf.worker.js` in the `build/generic/build/` di
 Both scripts are needed but only `pdf.js` needs to be included since `pdf.worker.js` will
 be loaded by `pdf.js`. The PDF.js files are large and should be minified for production.
 
+## Including via a CDN
```
yurydelendik: Can you move this section after `## Using PDF.js in a web application`?

timvandermeij:
@LukasDrgon Also, please squash the commits into one commit (see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits) once you made the change requested above.

LukasDrgon:
All done 👍

timvandermeij (approved):

timvandermeij:
Thank you for the patch!

PR #8982, opened 2017-10-01
[api-major] Remove deprecated code

timvandermeij:
This code has been deprecated with a visible message for two years, so we can safely remove it now.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -15,11 +15,11 @@
 /* globals requirejs, __non_webpack_require__ */
 
 import {
-  AbortException, assert, createPromiseCapability, deprecated,
-  getVerbosityLevel, info, InvalidPDFException, isArrayBuffer, isSameOrigin,
-  loadJpegStream, MessageHandler, MissingPDFException, NativeImageDecoding,
-  PageViewport, PasswordException, StatTimer, stringToBytes,
-  UnexpectedResponseException, UnknownErrorException, Util, warn
+  AbortException, assert, createPromiseCapability, getVerbosityLevel, info,
+  InvalidPDFException, isArrayBuffer, isSameOrigin, loadJpegStream,
+  MessageHandler, MissingPDFException, NativeImageDecoding, PageViewport,
+  PasswordException, StatTimer, stringToBytes, UnexpectedResponseException,
+  UnknownErrorException, Util, warn
 } from '../shared/util';
 import {
   DOMCanvasFactory, DOMCMapReaderFactory, getDefaultSetting,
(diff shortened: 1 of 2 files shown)


src/display/api.js
```diff
@@ -933,12 +933,6 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
       intentState.renderTasks.push(internalRenderTask);
       var renderTask = internalRenderTask.task;
 
-      // Obsolete parameter support
-      if (params.continueCallback) {
```
Snuffleupagus: If this is removed, then the following JSDoc comment should also be removed: https://github.com/mozilla/pdf.js/blob/50b1a91c0273a8b419a1525827dc43e9e3f92ba0/src/display/api.js#L759-L762

Snuffleupagus (approved):
Looks good to me, with the comment addressed (and sign-off from other devs).

@yurydelendik Are you OK with removing these deprecated API methods at this point?

timvandermeij:
I changed the title to include `[api-major]`. Should we create a separate branch for 2.0 or can we just merge this in the current master?

yurydelendik:
Oh, branch is a good idea.

src/display/api.js
timvandermeij: Done in the new commit.

src/display/api.js
```diff
@@ -262,13 +258,8 @@ function getDocument(src, pdfDataRangeTransport,
...
 
-  if (params.disableNativeImageDecoder !== undefined) {
-    deprecated('parameter disableNativeImageDecoder, ' +
-      'use nativeImageDecoderSupport instead');
-  }
   params.nativeImageDecoderSupport = params.nativeImageDecoderSupport ||
-    (params.disableNativeImageDecoder === true ? NativeImageDecoding.NONE :
-      NativeImageDecoding.DECODE);
+                                     NativeImageDecoding.DECODE;
   if (params.nativeImageDecoderSupport !== NativeImageDecoding.DECODE &&
```
Snuffleupagus: *This could probably be considered slightly out of scope, but I have to ask.*

Do we really need to print a warning message for this *particular* API parameter, considering that we don't do that for any other ones (and doing it everywhere would require adding *a lot* of code all over the place)?

Hence I'm wondering it wouldn't suffice to replace the current code with something along the following lines?
```javascript
if (!params.nativeImageDecoderSupport ||
    !(params.nativeImageDecoderSupport === NativeImageDecoding.DECODE ||
      params.nativeImageDecoderSupport === NativeImageDecoding.NONE ||
      params.nativeImageDecoderSupport === NativeImageDecoding.DISPLAY)) {
  params.nativeImageDecoderSupport = NativeImageDecoding.DECODE;
}
```

Snuffleupagus (approved):
Looks good to me (just one perhaps off-topic question); nice code-removal, thank you!

src/display/api.js
timvandermeij: I would personally be fine with that, but let's do that in a different PR because I'd like to keep this one limited to only removing deprecated code. Nevertheless, this is a good point and can be picked up later!

Issue #8395, opened 2017-05-09
PDF renders as blank in Chrome extension viewer


diracdeltas:
Link to PDF file (or attach file here):
https://npgallery.nps.gov/pdfhost/docs/NHLS/Text/70000736.pdf

Configuration:
- Web browser and its version: Chrome 58
- Operating system and its version: Mac
- PDF.js version: 1.6.293 version of the chrome extension
- Is an extension: Yes

Steps to reproduce the problem:
1. Load the PDF
2. It's blank

What is the expected behavior? (add screenshot)
Should not be blank; it displays properly in chrome's built-in PDF viewer

What went wrong? (add screenshot)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

Snuffleupagus:
Closing as duplicate of issue #2073.

Issue #8145, opened 2017-03-09
Random pages blank in this document


berricks:
Link to PDF file (or attach file here): [The study of the effect of shock and vibration in the distributio.pdf](https://github.com/mozilla/pdf.js/files/832584/The.study.of.the.effect.of.shock.and.vibration.in.the.distributio.pdf)

Configuration:
- Web browser and its version:  Firefox 52
- Operating system and its version:  Mac OS 10.12.3
- PDF.js version: 1.5.188  Also tried this on the pdf.js demo
- Is an extension: no

Steps to reproduce the problem:
1. Open the pdf. Scroll down the pages and notice many pages are blank
2. Go to page 21 and notice this page is rendered. 

What is the expected behavior? (add screenshot)  all pages should be rendered. 

What went wrong? (add screenshot)
not sure.  Some of the blank pages do see to have a text layer that's selectable.  All pages are viewable in Mac Preview.  Chrome pdf view does not correctly render the pages that are blank in pdf.js, here's an example of page 17.
![servefile](https://cloud.githubusercontent.com/assets/18216868/23776147/9ebae9ae-04e9-11e7-8109-90a401b20c44.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

Snuffleupagus:
Closing as duplicate of issue #2073.

PR #8991, opened 2017-10-03
Implement Huffman coding in JBIG2

### LINKED ISSUE #2073: huffman is not supported
While I've tried to upload a PDF/A document which contains a compressed image, I got the following error:

PDF.JS Build: b0d8d56
Message: JBIG2 error: huffman is not supported

janpe2:
Fixes #2073 (and the closed duplicates #2837, #3092, #8145, #8395)
Fixes [Bugzilla 1131384](https://bugzilla.mozilla.org/show_bug.cgi?id=1131384)

I have implemented Huffman and MMR coding in the JBIG2 image decoder.

To implement MMR decoding, I have used the existing `CCITTFaxStream` because MMR is the same compression algorithm as the PDF filter `CCITTFaxDecode` with `/K -1`. However, I got a circular dependency problem because jbig2.js must import stream.js and vice versa. I resolved the problem by moving the class Jbig2Stream from stream.js to jbig2.js. Now I was able to remove one of the imports.

I have not implemented "symbol refinement with Huffman" because I can't find any files that would use that feature. And it is poorly documented in the JBIG2 standard.

I have added two test cases. The first one is from issue #2837 (page 7) and it contains Huffman coded SymbolDictionaries and TextRegions and Tables segments. The second one shows the test images from the JBIG2 standard (ITU T.88, Annex H) and they contain also an MMR coded GenericRegion, PatternDictionary and HalftoneRegion.

### DIFF
--- a/src/core/jbig2.js
+++ b/src/core/jbig2.js
@@ -13,10 +13,9 @@
  * limitations under the License.
  */
 
-import {
-  log2, readInt8, readUint16, readUint32, shadow
-} from '../shared/util';
+import { log2, readInt8, readUint16, readUint32, shadow } from '../shared/util';
 import { ArithmeticDecoder } from './arithmetic_decoder';
+import { CCITTFaxDecoder } from './ccitt';
 
 let Jbig2Error = (function Jbig2ErrorClosure() {
   function Jbig2Error(msg) {
@@ -206,7 +205,9 @@ var Jbig2Image = (function Jbig2ImageClosure() {
   function decodeBitmap(mmr, width, height, templateIndex, prediction, skip, at,
                         decodingContext) {
     if (mmr) {
-      throw new Jbig2Error('MMR encoding is not supported');
+      let input = new Reader(decodingContext.data, decodingContext.start,
+        decodingContext.end);
+      return decodeMMRBitmap(input, width, height, false);
     }
 
     // Use optimized version for the most common case
@@ -413,9 +414,9 @@ var Jbig2Image = (function Jbig2ImageClosure() {
                                   numberOfNewSymbols, numberOfExportedSymbols,
                                   huffmanTables, templateIndex, at,
                                   refinementTemplateIndex, refinementAt,
-                                  decodingContext) {
-    if (huffman) {
-      throw new Jbig2Error('huffman is not supported');
+                                  decodingContext, huffmanInput) {
+    if (huffman && refinement) {
+      throw new Jbig2Error('symbol refinement with Huffman is not supported');
     }
 
     var newSymbols = [];
@@ -424,17 +425,29 @@ var Jbig2Image = (function Jbig2ImageClosure() {
 
     var decoder = decodingContext.decoder;
     var contextCache = decodingContext.contextCache;
+    let tableB1, symbolWidths;
+    if (huffman) {
+      tableB1 = getStandardTable(1); // standard table B.1
+      symbolWidths = [];
+      symbolCodeLength = Math.max(symbolCodeLength, 1); // 6.5.8.2.3
+    }
 
     while (newSymbols.length < numberOfNewSymbols) {
-      var deltaHeight = decodeInteger(contextCache, 'IADH', decoder); // 6.5.6
+      var deltaHeight = huffman ?
+        huffmanTables.tableDeltaHeight.decode(huffmanInput) :
+        decodeInteger(contextCache, 'IADH', decoder); // 6.5.6
       currentHeight += deltaHeight;
-      var currentWidth = 0;
+      let currentWidth = 0, totalWidth = 0;
+      let firstSymbol = huffman ? symbolWidths.length : 0;
       while (true) {
-        var deltaWidth = decodeInteger(contextCache, 'IADW', decoder); // 6.5.7
+        var deltaWidth = huffman ?
+          huffmanTables.tableDeltaWidth.decode(huffmanInput) :
+          decodeInteger(contextCache, 'IADW', decoder); // 6.5.7
         if (deltaWidth === null) {
           break; // OOB
         }
         currentWidth += deltaWidth;
+        totalWidth += currentWidth;
         var bitmap;
         if (refinement) {
           // 6.5.8.2 Refinement/aggregate-coded symbol bitmap
@@ -451,7 +464,7 @@ var Jbig2Image = (function Jbig2ImageClosure() {
                                       0, // OR operator
                                       huffmanTables,
                                       refinementTemplateIndex, refinementAt,
-                                      decodingContext);
+                                      decodingContext, 0, huffmanInput);
           } else {
             var symbolId = decodeIAID(contextCache, decoder, symbolCodeLength);
             var rdx = decodeInteger(contextCache, 'IARDX', decoder); // 6.4.11.3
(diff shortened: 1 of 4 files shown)


src/core/jbig2.js
```diff
@@ -1306,6 +2103,62 @@ var Jbig2Image = (function Jbig2ImageClosure() {
...
+    this.buffer = data;
+    this.bufferLength = dataLength;
+    this.eof = true;
+  };
+
+  return Jbig2Stream;
+})();
+
 export {
-  Jbig2Image,
```
Snuffleupagus: Since there's people that use the image decoders in PDF.js in other projects, it would probably be a good idea to keep exporting `Jbig2Image` here (and we may also want to update https://github.com/notmasteryet/jpgjs/ in the future).

test/pdfs/jbig2_huffman_1.pdf.link
```diff
@@ -0,0 +1 @@
+https://bugs.freedesktop.org/attachment.cgi?id=11143
```
Snuffleupagus: Please link to the Internet Archive instead, to avoid the file changing or becoming unavailable: https://web.archive.org/web/20171003213334/https://bugs.freedesktop.org/attachment.cgi?id=11143

test/pdfs/jbig2_huffman_2.pdf.link
```diff
@@ -0,0 +1 @@
+http://library.sciencemadness.org/library/books/gergel_isopropyl_bromide.pdf
```
Snuffleupagus: Please link to the Internet Archive instead, to avoid the file changing or becoming unavailable: https://web.archive.org/web/20170924053030/http://library.sciencemadness.org/library/books/gergel_isopropyl_bromide.pdf

src/core/jbig2.js
```diff
@@ -14,6 +14,12 @@
  */
 
 import {
+  CCITTFaxStream, DecodeStream, Stream
+} from './stream';
+import {
+  Dict, isDict, isStream
+} from './primitives';
+import {
```
Snuffleupagus: Nit: It looks like all of these `import` statements would fit on a single line, i.e.
```javascript
import { CCITTFaxStream, DecodeStream, Stream } from './stream';
import { Dict, isDict, isStream } from './primitives';
import { log2, readInt8, readUint16, readUint32, shadow } from '../shared/util';
```

Snuffleupagus (commented):
This looks like a really impressive piece of work! Unfortunately I'm not really familiar with the JBig2 decoder, so I'm hoping that @yurydelendik has time to do the actual review.

> I resolved the problem by moving the class Jbig2Stream from stream.js to jbig2.js. Now I was able to remove one of the imports.

Would it perhaps make sense to move the Stream class to another (e.g. its own) file, rather than placing it in the image decoder file (since it's used outside of PDF.js, e.g. in https://github.com/notmasteryet/jpgjs/)?

janpe2:
> Would it perhaps make sense to move the Stream class to another (e.g. its own) file, rather than placing it in the image decoder file

So maybe I'll create a new file `core/jbig2_stream.js` and I move Jbig2Stream there. Did I understand you correctly?

If I create a new source file, should I update also some makefile or unit test files etc?

Snuffleupagus:
> So maybe I'll create a new file core/jbig2_stream.js and I move Jbig2Stream there. Did I understand you correctly?

Yes, that was my idea. However, since opinions may vary, you could perhaps hold off on making this particular change until a more complete review has be done.

> If I create a new source file, should I update also some makefile or unit test files etc?

As long as you update all the places that currently import `JBig2Stream` (seems to be limited to `src/core`), no other changes need to be made.

test/test_manifest.json
```diff
@@ -2408,6 +2408,21 @@
       "link": true,
       "type": "eq"
     },
+    {  "id": "jbig2_huffman_1",
+       "file": "pdfs/jbig2_huffman_1.pdf",
+       "md5": "93ccd85c5686bea27d3e6c3b41921b21",
+       "lastPage": 1,
+       "rounds": 1,
```
Snuffleupagus: You need to add `"link": true,` since otherwise the file won't be downloaded, which means that testing fails.

test/test_manifest.json
```diff
@@ -2408,6 +2408,21 @@
...
+       "type": "eq"
+    },
+    {  "id": "jbig2_huffman_2",
+       "file": "pdfs/jbig2_huffman_2.pdf",
+       "md5": "f019e2722bd64684e2093a0933e390f4",
+       "firstPage": 7,
+       "lastPage": 7,
+       "rounds": 1,
```
Snuffleupagus: You need to add `"link": true,` since otherwise the file won't be downloaded, which means that testing fails.

Snuffleupagus:
@yurydelendik If you have time, could you please review this PR?

yurydelendik:
> If you have time, could you please review this PR?

I'll try. I'm a little bit scared about pulling stream.js into jbig2.js. Can we refactor, e.g. in the separate PR , to move out CCITT and JBIG2 streams from the stream.js (e.g. ccitt_stream.js and jbig2_stream.js)? As well, move "pure" CCITT decoding into separate file as well (e.g. ccitt.js). The jbig2.js will depend on ccitt.js, but never on stream.js and ccitt_stream.js -- this will remove need to fake Stream and Dict to decode the data. I'll review JBIG2 logic more in the future.

Snuffleupagus:
*I've added this to the `Version 2.0` project, since I definitely think that this is the kind of improvement that we should ensure gets included in the next release.*

src/core/jbig2.js
```diff
@@ -1293,8 +1394,718 @@ var Jbig2Image = (function Jbig2ImageClosure() {
...
+      customTables[currentSegment] = decodeTablesSegment(data, start, end);
+    },
   };
 
+  function HuffmanLine(lineData) {
+    if (lineData.length === 2) {
+      // OOB line.
+      this.isOOB = true;
+      this.prefixLength = lineData[0];
+      this.prefixCode = lineData[1];
```
yurydelendik: initialize all fields (rangeLow, rangeLength, isLowerRange) here and in the same order as below. add `this.isOOB = false;` below too

yurydelendik (approved):
In general it looks good -- I added few comments. There are many places to optimize the performance, e.g. readUncompressedBitmap, readBits, or HuffmanLine refactoring, but these can be addressed later.

Sorry, it took me so long. It's really hard to review this at once. I did not review Huffman algorithm usage too, but everything looks in place and I did not find any possible infinite loops. We can fix broken images later. 

Thank you.

src/core/jbig2.js
janpe2: Done.

src/core/jbig2.js
```diff
@@ -1293,8 +1394,718 @@ var Jbig2Image = (function Jbig2ImageClosure() {
...
+
+    for (y = 0; y < height; y++) {
+      row = new Uint8Array(width);
+      bitmap.push(row);
+      shift = -1;
+      for (x = 0; x < width; x++) {
+        if (shift < 0) {
+          currentByte = decoder.readNextChar();
+          if (currentByte === -1) {
+            eof = true;
```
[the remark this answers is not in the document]
janpe2: This is an unexpected error case and I think it's best to fill the image with zeros, which is white in JBIG2. We must continue decoding so that the number of rows in the image will be equal to `height`.

Snuffleupagus:
@janpe2 Thanks a lot for implementing this, it's great to finally have Huffman support for JBig2 images!

PR #8992, opened 2017-10-04
[api-major] When rendering is cancelled, always reject with `RenderingCancelledException`

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Looks good!

PR #8994, opened 2017-10-05
Fix a regression that (effectively) makes `PDFHistory.forward` a no-op

Snuffleupagus:
- Remove the `this._currentUid` property from `PDFHistory`, since it's no longer needed

    Commit 938dffb, in PR #8885, removed the only actual usage of `this._currentUid` and it can thus be removed.

---

- Fix a regression that (effectively) makes `PDFHistory.forward` a no-op

  *It appears that this accidentally broken in PR #8775.*

  Note that `PDFHistory.forward` is only used with certain named actions, and these aren't that commonly used, which ought to explain why this error managed to sneak in.

  Steps to reproduce the issue (and verify the fix):
   1. Navigate to e.g. http://mirrors.ctan.org/info/lshort/english/lshort.pdf
   2. Click on a couple of links, or outline items, such that the history is populated with a few entries.
   3. In the console, execute `PDFViewerApplication.pdfHistory.back()` one or more times, thus navigating back to a previous viewer position.
   4. In the console, execute `PDFViewerApplication.pdfHistory.forward() one or more times.

  At the last step above, no (forward) navigation happens with the current `master`; now compare with this patch.

### DIFF
--- a/web/pdf_history.js
+++ b/web/pdf_history.js
@@ -109,7 +109,7 @@ class PDFHistory {
     this._currentHash = getCurrentHash();
     this._numPositionUpdates = 0;
 
-    this._uid = 0;
+    this._uid = this._maxUid = 0;
     this._destination = null;
     this._position = null;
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Works fine, thank you for fixing this!

PR #9001, opened 2017-10-06
Prevent the `annotationLayer` from, in some cases, becoming duplicated on the first page when the document loads

Snuffleupagus:
I don't know if this is a regression, but I noticed earlier today that depending on the initial scale *and* sidebar state, the `annotationLayer` of the first rendered page may end up duplicated; please see screen-shot below.

double-annotations

I can reproduce this reliable with e.g. https://arxiv.org/pdf/1112.0542v1.pdf#zoom=page-width&pagemode=bookmarks.
When the document loads, rendering of the first page begins immediately. When the sidebar is then opened, that forces re-rendering which thus aborts rendering of the first page.
Note that calling `PDFPageView.draw()` will always, provided an `AnnotationLayerFactory` instance exists, call `AnnotationLayerBuilder.render()`. Hence the events described above will result in *two* such calls, where the actual annotation rendering/updating happens asynchronously.

For reasons that I don't (at all) understand, when multiple `pdfPage.getAnnotations()` promises are handled back-to-back (in `AnnotationLayerBuilder.render()`), the `this.div` property seems to not update in time for the subsequent calls.
This thus, at least in Firefox, result in double rendering of all annotations on the first page.

Obviously it'd be good to find out why it breaks, since it *really* shouldn't, but this patch at least provides a (hopefully) acceptable work-around by ignoring `getAnnotations()` calls for `AnnotationLayerBuilder` instances that we're destroying (in `PDFPageView.reset()`).

*Edit:* The patch was updated slightly to ensure that we *also* won't attempt to render an `annotationLayer` during closing of a PDF document, i.e. after calling `PDFViewerApplication.close()`.

### DIFF
--- a/web/annotation_layer_builder.js
+++ b/web/annotation_layer_builder.js
@@ -41,6 +41,7 @@ class AnnotationLayerBuilder {
     this.l10n = l10n;
 
     this.div = null;
+    this._cancelled = false;
   }
 
   /**
@@ -49,6 +50,10 @@ class AnnotationLayerBuilder {
    */
   render(viewport, intent = 'display') {
     this.pdfPage.getAnnotations({ intent, }).then((annotations) => {
+      if (this._cancelled) {
+        return;
+      }
+
       let parameters = {
         viewport: viewport.clone({ dontFlip: true, }),
         div: this.div,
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this! I think being able to cancel rendering here as well (in addition to the various other viewer components) is a good addition.

PR #9003, opened 2017-10-06
Add downloading of the artifact ff build for mochitests

yurydelendik:
To run `gulp prepare-mochitest` and then `gulp mochitest`

Experimental:
- [ ] check if works on Linux
- [ ] check if works on Windows


timvandermeij:
@yurydelendik Is this still relevant to have? I think upstream is running these tests anyway, right?

timvandermeij:
Closing until it's determined that we need to do something with this.

PR #9006, opened 2017-10-07
Translater

Jul10l1r4:


Snuffleupagus:
Unfortunately we cannot accept this patch, since as explained in [CONTRIBUTING.md](https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md) (see excerpt below, we're importing the translations from Firefox.

If you think this is incorrect, please suggest the change upstream first.

PR #9007, opened 2017-10-07
Import compatibility.js

Jul10l1r4:
The file master/src/shared/compatibility.js was not being imported.

### DIFF
--- a/web/viewer.html
+++ b/web/viewer.html
@@ -45,6 +45,7 @@ See https://github.com/adobe-type-tools/cmap-resources
 <!--#endif-->
 
 <!--#if !PRODUCTION-->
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
That file is *purposely* not loaded in the viewer in non-production mode; please refer to commit https://github.com/mozilla/pdf.js/commit/e49dfe4ed7700a8742457adf49a3d4635df5d13c, in PR #8778, for additional info.
As explained there, it's already being imported from the `src/shared/util.js` file, and furthermore this patch will *actually* throw a `TypeError` when encountering `require(...)` statements in the `src/shared/compatibility.js` file.

Closing as invalid, since we unfortunately cannot accept the patch.

PR #9008, opened 2017-10-07
Mispelled isEvalSupported property at FontFaceObject() creation.

diegocr:
Hello,

While looking at the code for available options..i've found a typo when creating the `FontFaceObject()` instances, the provided `isEvalSuported` option should actually be `isEvalSupported` (note the extra `p`)

Not sure if that would be causing any real issue since there is a fallback logic, but anyway i thought you may want that fixed, fwiw it got introduced over two years ago on commit 1d12aed5cac6a0cac

Cheers.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1818,7 +1818,7 @@ var WorkerTransport = (function WorkerTransportClosure() {
               };
             }
             var font = new FontFaceObject(exportedData, {
-              isEvalSuported: getDefaultSetting('isEvalSupported'),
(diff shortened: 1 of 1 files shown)


Snuffleupagus (approved):
Really nice catch! Especially considering that the typo actually causes worse performance, since it prevents the fast-path in `FontFaceObject.getPathGenerator` from being used.

r=me, with passing tests.
Thank you for spotting, and fixing, this long-standing bug!

timvandermeij:
Nice find!

PR #9013, opened 2017-10-09
Fix a `PDFHistory` regression with document hashes of the `nameddest=...` form

Snuffleupagus:
Unfortunately I've just found out that this isn't working entirely correct; my apologies for accidentally breaking this in PR #8775.

Compare e.g. this link: http://mirrors.ctan.org/info/lshort/english/lshort.pdf#page.157, with this one: http://mirrors.ctan.org/info/lshort/english/lshort.pdf#nameddest=page.157.

Notice how in the *second* case, the history stops working correctly.

*The various edge-case regressions in the new `PDFHistory` code is reminding my why I put off the rewrite for so long :-(*

### DIFF
--- a/test/unit/pdf_history_spec.js
+++ b/test/unit/pdf_history_spec.js
@@ -13,10 +13,36 @@
  * limitations under the License.
  */
 
-import { isDestsEqual } from '../../web/pdf_history';
+import { isDestArraysEqual, isDestHashesEqual } from '../../web/pdf_history';
 
 describe('pdf_history', function() {
-  describe('isDestsEqual', function() {
+  describe('isDestHashesEqual', function() {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

> The various edge-case regressions in the new PDFHistory code is reminding my why I put off the rewrite for so long :-(

The rewrite did bring us modern and more understandable/maintainable code that is also partly unit tested. I see this as a major improvement over the old situation that had hard-to-fix issues. Big rewrites always have a higher risk for regressions, but in this case I think there are relatively few of them and they are easily fixable, so don't worry about it too much! :-)

Snuffleupagus:
@timvandermeij As always, thanks a lot for reviewing :-)

PR #9026, opened 2017-10-14
Implement sidebar resizing for modern browsers, by utilizing CSS variables (issue 2072)

### LINKED ISSUE #2072: Outline (bookmark) pane should be resizable.

Snuffleupagus:
*After having seen how complicated issue #2072 would be to implement, given the current state of our existing CSS rules, I started wondering if a simpler solution wouldn't be possible.
This PR is the result of a bit of experimentation, and contains an alternative approach for addressing issue #2072. ~~It currently depends on, and for testing purposes include, the patch from PR #8993.~~*

---

By making use of modern CSS features, in this case [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables), implementing sidebar resizing is actually quite simple. Not only will the amount of added code be fairly small, but it should also be easy to maintain since there's no need for complicated JavaScript hacks in order to update the CSS. Another benefit is that the JavaScript code doesn't need to make detailed assumptions about the exact structure of the HTML/CSS code.

Obviously this will not work in older browsers, such as IE, that lack support for CSS variables. In those cases sidebar resizing is simply disabled (via feature detection), and the resizing DOM element hidden, and the behaviour is thus *identical* to the current (fixed-width) sidebar.
However, considering the simplicity of the implementation, I really don't see why limiting this feature to "modern" browsers is a problem.

Finally, note that a few edge-cases meant that the patch is a bit larger than what the basic functionality would dictate. Among those is first of all proper RTL support, and secondly (automatic) resizing of the sidebar when the width of the *entire* viewer changes. Another, pre-existing, issue fixed here is the incomplete interface of `NullL10n`.

*Please note:* This patch has been successfully tested in both LTR and RTL viewer locales, in recent versions of Firefox and Chrome.

Fixes #2072.

### DIFF
--- a/web/pdf_sidebar_resizer.js
+++ b/web/pdf_sidebar_resizer.js
@@ -1,0 +1,184 @@
+/* Copyright 2017 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { NullL10n } from './ui_utils';
+
+const SIDEBAR_WIDTH_VAR = '--sidebar-width';
+const SIDEBAR_MIN_WIDTH = 200; // pixels
+const SIDEBAR_RESIZING_CLASS = 'sidebarResizing';
+
+/**
+ * @typedef {Object} PDFSidebarResizerOptions
+ * @property {HTMLDivElement} outerContainer - The outer container
+ *   (encasing both the viewer and sidebar elements).
+ * @property {HTMLDivElement} resizer - The DOM element that can be dragged in
+ *   order to adjust the width of the sidebar.
+ */
+
+class PDFSidebarResizer {
+  /**
+   * @param {PDFSidebarResizerOptions} options
+   * @param {EventBus} eventBus - The application event bus.
+   * @param {IL10n} l10n - Localization service.
+   */
+  constructor(options, eventBus, l10n = NullL10n) {
+    this.enabled = false;
+    this.isRTL = false;
+    this.sidebarOpen = false;
+    this.doc = document.documentElement;
+    this._width = null;
+    this._outerContainerWidth = null;
+    this._boundEvents = Object.create(null);
+
+    this.outerContainer = options.outerContainer;
+    this.resizer = options.resizer;
+    this.eventBus = eventBus;
+    this.l10n = l10n;
+
+    if (typeof CSS === 'undefined' || typeof CSS.supports !== 'function' ||
+        !CSS.supports(SIDEBAR_WIDTH_VAR, `calc(-1 * ${SIDEBAR_MIN_WIDTH}px)`)) {
+      console.warn('PDFSidebarResizer: ' +
+                   'The browser does not support resizing of the sidebar.');
+      return;
+    }
+    this.enabled = true;
+    this.resizer.classList.remove('hidden'); // Show the resizer DOM element.
+
+    this.l10n.getDirection().then((dir) => {
+      this.isRTL = (dir === 'rtl');
+    });
+    this._addEventListeners();
+  }
+
+  /**
+   * returns {number}
+   */
+  get outerContainerWidth() {
+    if (!this._outerContainerWidth) {
+      this._outerContainerWidth = this.outerContainer.clientWidth;
+    }
+    return this._outerContainerWidth;
+  }
+
+  /**
+   * @private
+   * returns {boolean} Indicating if the sidebar width was updated.
+   */
+  _updateWidth(width = 0) {
+    if (!this.enabled) {
+      return false;
+    }
+    // Prevent the sidebar from becoming too narrow, or from occupying more
+    // than half of the available viewer width.
+    const maxWidth = Math.floor(this.outerContainerWidth / 2);
+    if (width > maxWidth) {
+      width = maxWidth;
+    }
+    if (width < SIDEBAR_MIN_WIDTH) {
+      width = SIDEBAR_MIN_WIDTH;
+    }
+    // Only update the UI when the sidebar width did in fact change.
+    if (width === this._width) {
+      return false;
+    }
+    this._width = width;
+    this.doc.style.setProperty(SIDEBAR_WIDTH_VAR, `${width}px`);
+    return true;
+  }
+
+  /**
+   * @private
+   */
+  _mouseMove(evt) {
+    let width = evt.clientX;
+    // For sidebar resizing to work correctly in RTL mode, invert the width.
+    if (this.isRTL) {
+      width = this.outerContainerWidth - width;
+    }
+    this._updateWidth(width);
+  }
+
+  /**
+   * @private
+   */
+  _mouseUp(evt) {
+    // Re-enable the `transition-duration` rules when sidebar resizing ends...
+    this.outerContainer.classList.remove(SIDEBAR_RESIZING_CLASS);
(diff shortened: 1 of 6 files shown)


web/pdf_sidebar_resizer.js
```diff
@@ -0,0 +1,149 @@
...
+const SIDEBAR_WIDTH_VAR = '--sidebar-width';
+const SIDEBAR_MIN_WIDTH = 200; // pixels
+const SIDEBAR_RESIZING_CLASS = 'sidebarResizing';
+
+/**
+ * @typedef {Object} PDFSidebarResizerOptions
+ * @property {HTMLDivElement} outerContainer - The outer container
+ *   (encasing both the viewer and sidebar elements).
+ * @property {HTMLDivElement} sidebarContainer - The sidebar container
+ *   (in which the sidebar view element are placed).
```
timvandermeij: Nit: `sidebar view elements are`

web/pdf_sidebar_resizer.js
```diff
@@ -0,0 +1,149 @@
...
+    this.sidebarContainer = options.sidebarContainer;
+    this.resizer = options.resizer;
+    this.eventBus = eventBus;
+    this.l10n = l10n;
+
+    if (typeof CSS === 'undefined' || typeof CSS.supports !== 'function' ||
+        !CSS.supports(SIDEBAR_WIDTH_VAR, `calc(-1 * ${SIDEBAR_MIN_WIDTH}px)`)) {
+      console.warn('PDFSidebarResizer: ' +
+                   'The browser does not support resizing of the sidebar.');
+      this.resizer.classList.add('hidden'); // Hide the resizer DOM element.
```
timvandermeij: It's hidden by default in the HTML, so I think there is no need to add it again.

web/pdf_sidebar_resizer.js
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * @private
+   */
+  _updateWidth(width = 0) {
+    if (!this.enabled) {
+      return;
+    }
+    // Prevent the sidebar from becoming too narrow, or from occupying more
+    // than half of the available viewer width.
+    const maxWidth = Math.floor(this.outerContainer.clientWidth / 2);
```
timvandermeij: Can't we precalculate this in the constructor so we don't have to on every call to `_updateWidth`?

web/viewer.css
```diff
@@ -146,6 +150,7 @@ select {
   top: 32px;
   bottom: 0;
   width: 200px;
+  width: var(--sidebar-width);
```
timvandermeij: For one of these occurrences, can you add a comment why `width` is duplicated? I think it's done so that if browsers don't support CSS variables they fall back to the old `200px` width, but I can imagine that this CSS looks a bit weird if it's not known.

timvandermeij (commented):
Minor comments, but in general this looks really good and much easier than I expected!

web/pdf_sidebar_resizer.js
Snuffleupagus: > It's hidden by default in the HTML, so I think there is no need to add it again.

You're absolutely correct, this line is redundant!

> I think you added it just in case someone forgets to add it in the HTML, so if you want that then this can stay,

Originally I didn't (for some inexplicable reason) think to add the `hidden` class in the HTML, so this is basically just a left-over.
I've removed this line now, since nothing should break if the DOM element is visible without resizing being supported. Thanks for spotting this!

web/viewer.css
Snuffleupagus: > I think it's done so that if browsers don't support CSS variables they fall back to the old 200px width,

That's correct!
I've added a comment to the first of these occurrences, please check if you think clear enough.

web/pdf_sidebar_resizer.js
Snuffleupagus: > Can't we precalculate this so we don't have to on every call to _updateWidth?

Sure, that is a really nice idea!
I've added a getter for this value, and I don't think it's making the code unnecessarily complex.

Snuffleupagus:
@timvandermeij Thanks for good review comments!
Please find an interdiff of the changes below:
```diff
diff --git a/web/pdf_sidebar_resizer.js b/web/pdf_sidebar_resizer.js
index 6d8cdecf..d949b39b 100644
--- a/web/pdf_sidebar_resizer.js
+++ b/web/pdf_sidebar_resizer.js
@@ -24,7 +24,7 @@ const SIDEBAR_RESIZING_CLASS = 'sidebarResizing';
  * @property {HTMLDivElement} outerContainer - The outer container
  *   (encasing both the viewer and sidebar elements).
  * @property {HTMLDivElement} sidebarContainer - The sidebar container
- *   (in which the sidebar view element are placed).
+ *   (in which the sidebar view elements are placed).
  * @property {HTMLDivElement} resizer - The DOM element that can be dragged in
  *   order to adjust the width of the sidebar.
  */
@@ -39,6 +39,7 @@ class PDFSidebarResizer {
     this.enabled = false;
     this.isRTL = false;
     this.doc = document.documentElement;
+    this._outerContainerWidth = null;
     this._boundEvents = Object.create(null);
 
     this.outerContainer = options.outerContainer;
@@ -51,7 +52,6 @@ class PDFSidebarResizer {
         !CSS.supports(SIDEBAR_WIDTH_VAR, `calc(-1 * ${SIDEBAR_MIN_WIDTH}px)`)) {
       console.warn('PDFSidebarResizer: ' +
                    'The browser does not support resizing of the sidebar.');
-      this.resizer.classList.add('hidden'); // Hide the resizer DOM element.
       return;
     }
     this.enabled = true;
@@ -64,6 +64,16 @@ class PDFSidebarResizer {
   }
 
   /**
+   * returns {number}
+   */
+  get outerContainerWidth() {
+    if (!this._outerContainerWidth) {
+      this._outerContainerWidth = this.outerContainer.clientWidth;
+    }
+    return this._outerContainerWidth;
+  }
+
+  /**
    * @private
    */
   _updateWidth(width = 0) {
@@ -72,7 +82,7 @@ class PDFSidebarResizer {
     }
     // Prevent the sidebar from becoming too narrow, or from occupying more
     // than half of the available viewer width.
-    const maxWidth = Math.floor(this.outerContainer.clientWidth / 2);
+    const maxWidth = Math.floor(this.outerContainerWidth / 2);
     if (width > maxWidth) {
       width = maxWidth;
     }
@@ -87,9 +97,9 @@ class PDFSidebarResizer {
    */
   _mouseMove(evt) {
     let width = evt.clientX;
-    // For sidebar resizing to work corrently in RTL mode, invert the width.
+    // For sidebar resizing to work correctly in RTL mode, invert the width.
     if (this.isRTL) {
-      width = this.outerContainer.clientWidth - width;
+      width = this.outerContainerWidth - width;
     }
     this._updateWidth(width);
   }
@@ -132,6 +142,9 @@ class PDFSidebarResizer {
       // When the *entire* viewer is resized, such that it becomes narrower,
       // ensure that the sidebar doesn't end up being too wide.
       if (evt && evt.source === window) {
+        // Always reset the cached width when the viewer is resized.
+        this._outerContainerWidth = null;
+
         this.outerContainer.classList.add(SIDEBAR_RESIZING_CLASS);
         // Attempt to re-apply the current sidebar width.
         this._updateWidth(this.sidebarContainer.clientWidth);
diff --git a/web/viewer.css b/web/viewer.css
index ecb25695..1d2f4f8d 100644
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -149,7 +149,8 @@ select {
   position: absolute;
   top: 32px;
   bottom: 0;
-  width: 200px;
+  width: 200px; /* Here, and elsewhere below, keep the constant value for compatibility
+                   with older browsers that lack support for CSS variables. */
   width: var(--sidebar-width);
   visibility: hidden;
   z-index: 100;
```

timvandermeij (approved):

timvandermeij:
Thank you for implementing this! Really nice to have.

jeenu-arm:
Great! When I can we expect this in the Chrome extension please? I'm on 1.8.557.

PR #9027, opened 2017-10-15
Replace our `WeakMap` polyfill with the one from core-js

Snuffleupagus:
Since we're already using core-js elsewhere in `compatibility.js`, we can reduce the amount of code we need to maintain ourselves.

https://github.com/zloirock/core-js#weakmap


timvandermeij (approved):

timvandermeij:
Nice work!

PR #9028, opened 2017-10-15
[api-major] Remove all remaining `deprecated` functions/methods

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Looks good!

*(I'm hoping that the forms were a one-time failure and not becoming a new intermittent, but we'll see.)*

PR #9029, opened 2017-10-15
Enable the `--report-unused-disable-directives` ESLint command line option

Snuffleupagus:
This option was added in [version `4.8.0` of ESLint](https://github.com/eslint/eslint/releases/tag/v4.8.0), which is already listed as the minimum version in our `package.json` file; please refer to https://eslint.org/docs/user-guide/command-line-interface#--report-unused-disable-directives for additional details.

Despite the caveat listed in the link above, I still think that using this option makes sense since it will help ensure that no longer necessary disable statements are removed.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1160,7 +1160,8 @@ gulp.task('lint', function (done) {
   console.log('### Linting JS files');
 
   // Ensure that we lint the Firefox specific *.jsm files too.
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Good idea. Thanks!

PR #9033, opened 2017-10-15
[api-major] Remove the `PDFJS_NEXT` option

timvandermeij:
Nothing uses this option anymore, so setting it is a no-op now. We can safely remove it.

Use `SKIP_BABEL` (instead of `PDFJS_NEXT`) now if you want to skip Babel translation for a build.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -81,7 +81,7 @@ var DEFINES = {
   SINGLE_FILE: false,
   COMPONENTS: false,
   LIB: false,
-  PDFJS_NEXT: false,
+  SKIP_BABEL: false,
 };
 
 function safeSpawnSync(command, parameters, options) {
@@ -130,8 +130,8 @@ function createWebpackConfig(defines, output) {
   var licenseHeader = fs.readFileSync('./src/license_header.js').toString();
   var enableSourceMaps = !bundleDefines.FIREFOX && !bundleDefines.MOZCENTRAL &&
                          !bundleDefines.CHROME;
-  var pdfjsNext = bundleDefines.PDFJS_NEXT ||
-                  process.env['PDFJS_NEXT'] === 'true';
+  var skipBabel = bundleDefines.SKIP_BABEL ||
+                  process.env['SKIP_BABEL'] === 'true';
 
   return {
     output: output,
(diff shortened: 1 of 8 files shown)


Snuffleupagus (commented):
> Nothing uses this option anymore, so setting it is a no-op now. We can safely remove it.

Did we actually decide that PDF.js `2.0` will *only* support *fully* ES6 compatible browsers (personally I wouldn't mind that at all :-)?
Because if not, then we probably still need to be able to produce Babel translated builds (e.g. for `gulp generic` and/or `gulp dist`).

gulpfile.js
```diff
@@ -152,7 +149,6 @@ function createWebpackConfig(defines, output) {
           loader: 'babel-loader',
           exclude: /src\/core\/(glyphlist|unicode)/, // babel is too slow
           options: {
-            presets: pdfjsNext ? undefined : ['env'],
```
Snuffleupagus: Please don't forget about https://github.com/timvandermeij/pdf.js/blob/b95d3edba297228f63ce0b57acd62ae81e41d119/gulpfile.js#L998

timvandermeij:
It is actually not decided yet, so that was an oversight on my part. I updated the commit to put Babel translation back so we can proceed with this. For tracking purposes, I did make a note on the project to check if we can only support fully ES6 compatible browsers, but for now that's not planned (although I wouldn't mind at all either).

timvandermeij:
Do you think we should have an option like `SKIP_BABEL` in the Gulpfile so we can still make builds without Babel translation?

Snuffleupagus:
> I did make a note on the project to check if we can only support fully ES6 compatible browsers,

I'd assume that the answer might be "no" for now, given that when I asked about removing the Promise polyfill on IRC (http://logs.glob.uno/?c=mozilla%23pdfjs&s=2+Oct+2017&e=2+Oct+2017#c68001), @yurydelendik seemed to want to keep it (http://logs.glob.uno/?c=mozilla%23pdfjs&s=2+Oct+2017&e=2+Oct+2017#c68002).
If we cannot assume that Promises are available, then I unfortunately cannot see how we would assume an ES6 browser.

Snuffleupagus (approved):
Thanks for updating the patch, so that Babel translation still works!

For bookkeeping purposes, could you please (quickly) mention in the commit message that `SKIP_BABEL` should be used instead of `PDFJS_NEXT`?
r=me with that fixed, thank you for the patch!

timvandermeij:
I amended the commit message to include the `SKIP_BABEL` change.

PR #9034, opened 2017-10-15
[api-major] Change `getJavaScript` to return `null`, rather than an empty Array, when no JavaScript exists

Snuffleupagus:
Other API methods already return `null`, rather than empty Arrays/Objects, hence it makes sense to change `getJavaScript` to be consistent.


timvandermeij (approved):

timvandermeij:
Indeed good for consistency. Thanks!

PR #9037, opened 2017-10-16
Re-factor how parameters are passed to the network streams

Snuffleupagus:
*This patch is the result of me starting to look into moving parameters from `PDFJS` into `getDocument` and other API methods.*

When familiarizing myself with the code, the signatures of the various network streams seemed to be unnecessarily cumbersome since `disableRange` is currently handled separately from other parameters.
I'm assuming that the explanation for this is probably "for historical reasons", as is often the case. Hence I'd like to clean this up *before* we start the larger, and more invasive, `PDFJS` parameter re-factoring.

*Edit:* Slightly smaller diff with https://github.com/mozilla/pdf.js/pull/9037/files?w=1.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -251,10 +251,7 @@ function getDocument(src) {
       if (rangeTransport) {
         networkStream = new PDFDataTransportStream(params, rangeTransport);
       } else if (!params.data) {
-        networkStream = new PDFNetworkStream({
-          source: params,
-          disableRange: getDefaultSetting('disableRange'),
-        });
+        networkStream = new PDFNetworkStream(params);
       }
 
       var messageHandler = new MessageHandler(docId, workerId, worker.port);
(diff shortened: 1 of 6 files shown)


src/display/api.js
```diff
@@ -286,9 +283,9 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
   let apiVersion =
     typeof PDFJSDev !== 'undefined' ? PDFJSDev.eval('BUNDLE_VERSION') : null;
 
+  source.disableRange = getDefaultSetting('disableRange');
   source.disableAutoFetch = getDefaultSetting('disableAutoFetch');
   source.disableStream = getDefaultSetting('disableStream');
-  source.chunkedViewerLoading = !!pdfDataRangeTransport;
```
Snuffleupagus: Removing this line isn't completely relevant to the rest of the patch, but since I was touching code nearby it seemed like a nice clean-up given that the last usage of this property was removed in PR #8617.

timvandermeij (approved):

timvandermeij:
Looks better this way, thanks!

PR #9046, opened 2017-10-19
Extract the actual decoding in `CCITTFaxStream` into a new `CCITTFaxDecoder` "class", which the new `CCITTFaxStream` depends on

Snuffleupagus:
*Please note:* This PR attempts to implement https://github.com/mozilla/pdf.js/pull/8991#issuecomment-337712016, in an effort to help with (quickly) unblocking the actual JBig2 changes proposed in PR #8991.

@yurydelendik Is this approximately what you had in mind?


src/core/ccitt_stream.js
```diff
@@ -0,0 +1,71 @@
...
+    let ccittFaxDecoder = new CCITTFaxDecoder({
+      K: this.params.get('K'),
+      EndOfLine: this.params.get('EndOfLine'),
+      EncodedByteAlign: this.params.get('EncodedByteAlign'),
+      Columns: this.params.get('Columns'),
+      Rows: this.params.get('Rows'),
+      EndOfBlock: this.params.get('EndOfBlock'),
+      BlackIs1: this.params.get('BlackIs1'),
+    });
+    let data = ccittFaxDecoder.parse(this.bytes, this.maybeLength);
```
yurydelendik: Let's change this to pass a "source" of data, and this interface will have a single next() methods (which will return -1 for EOF).

src/core/ccitt.js
```diff
@@ -0,0 +1,1024 @@
...
+    let eoblock = options['EndOfBlock'];
+    if (eoblock === null || eoblock === undefined) {
+      eoblock = true;
+    }
+    this.eoblock = eoblock;
+    this.black = options['BlackIs1'] || false;
+  }
+
+  CCITTFaxDecoder.prototype = {
+    parse(data, maybeMinBufferLength = 0) {
```
yurydelendik: Let's remove this method and expose e.g. readNextChar() (former lookChar)

src/core/ccitt_stream.js
```diff
@@ -0,0 +1,71 @@
...
+
+  Object.defineProperty(CCITTFaxStream.prototype, 'bytes', {
+    get() {
+      // If this.maybeLength is null, we'll get the entire stream.
+      return shadow(this, 'bytes', this.str.getBytes(this.maybeLength));
+    },
+    configurable: true,
+  });
+
+  CCITTFaxStream.prototype.ensureBuffer = function() {
```
yurydelendik: This looks wrong usage/override of ensureBuffer -- ensureBuffer make sure that this.buffer has enough memory allocated.

src/core/ccitt_stream.js
```diff
@@ -0,0 +1,71 @@
...
+    });
+    let data = ccittFaxDecoder.parse(this.bytes, this.maybeLength);
+
+    this.buffer = data;
+    this.bufferLength = data.length;
+    this.eof = true;
+  };
+
+  CCITTFaxStream.prototype.readBlock = function() {
+    this.ensureBuffer();
```
yurydelendik: Let's make this method as in previous version for now.

src/core/jbig2_stream.js
```diff
@@ -0,0 +1,79 @@
...
+
+  Object.defineProperty(Jbig2Stream.prototype, 'bytes', {
+    get() {
+      // If this.maybeLength is null, we'll get the entire stream.
+      return shadow(this, 'bytes', this.stream.getBytes(this.maybeLength));
+    },
+    configurable: true,
+  });
+
+  Jbig2Stream.prototype.ensureBuffer = function() {
```
yurydelendik: Looks like this was a hack: can we add `req` parameter and probably a comment that we just ignoring interface and moving all parsed data into the buffer.

yurydelendik (commented):
I'm thinking we need to preserve a streaming nature of the streams and not consume all data as once, when it's possible.

src/core/ccitt.js
Snuffleupagus: Sure, I've changed the name as suggested in the new version of the PR.

src/core/ccitt_stream.js
Snuffleupagus: Agreed; this was removed in the new version.

src/core/ccitt_stream.js
Snuffleupagus: > Let's change this to pass a "source" of data,

*Hopefully I didn't completely misunderstand what you're asking for here :-)*

I've change the interface of `CCITTFaxDecoder` such that you can either initialize `CCITTFaxDecoder` with a stream (used from `CCITTFaxStream`), or pass in a `Uint8Array` instead (which seems like what you'd want to do in PR #8991 when decoding Huffman data).

> and this interface will have a single next() methods

Used the `readNextChar` name (as suggested above).

> (which will return -1 for EOF).

Good point, fixed now!

src/core/ccitt_stream.js
Snuffleupagus: I've reverted (most of) this, with the exception of the `lookChar` -> `readNextChar` rename and the handling of `-1` return values.

src/core/jbig2_stream.js
Snuffleupagus: Since this is a pre-existing issue, which is shared between `JpegStream`, `JpxStream`, and `Jbig2Stream`, could we perhaps do that in a follow-up instead?

The interface of all of those streams look kind of weird, and I'm actually a bit surprised that there hasn't been any errors because of it. For example: None of them actually implement `readBlock` methods, and it seems more luck that anything else that we're not calling `getBytes()` (without providing a length) for those streams, since that would trigger a code-path in [`getBytes`]( https://github.com/mozilla/pdf.js/blob/d71a576b302c6eeaa66b2affc055c796b4bd9399/src/core/stream.js#L193-L216) that assumes `readBlock` to exist.

One simple solution might to just replace `ensureBuffer` with `readBlock` for the `JpegStream`, `JpxStream`, and `Jbig2Stream` streams, which seems more correct to me!?
Anyway, since this is an old issue, I'm hoping this point can be deferred to another PR.

src/core/ccitt.js
```diff
@@ -0,0 +1,1047 @@
...
+  ];
+
+  /**
+   * @param {Stream|Uint8Array} source - The data that should be decoded.
+   * @param {Object} options - (optional) Decoding options.
+   */
+  function CCITTFaxDecoder(source, options = {}) {
+    if (isStream(source)) {
+      this.source = source;
+    } else if (source instanceof Uint8Array) {
```
yurydelendik: We can remove this branch of code for now (and isStream check) and keep only this.source = source; in this patch.

src/core/ccitt_stream.js
```diff
@@ -0,0 +1,60 @@
...
+
+    DecodeStream.call(this, maybeLength);
+  }
+
+  CCITTFaxStream.prototype = Object.create(DecodeStream.prototype);
+
+  CCITTFaxStream.prototype.readBlock = function() {
+    while (!this.eof) {
+      let c = this.ccittFaxDecoder.readNextChar();
+      if (c === -1) {
```
yurydelendik: This "if" is not present in the original code. Do you know what is changed?

src/core/ccitt_stream.js
```diff
@@ -0,0 +1,60 @@
...
+
+var CCITTFaxStream = (function CCITTFaxStreamClosure() {
+  function CCITTFaxStream(str, maybeLength, params) {
+    this.str = str;
+    this.dict = str.dict;
+
+    if (!isDict(params)) {
+      params = Dict.empty;
+    }
+    this.ccittFaxDecoder = new CCITTFaxDecoder(str, {
```
yurydelendik: I was thinking we could wrap `str` into `source` above, e.g. `const source = { next() { return str.getByte(); } };`, but this will work fine for now. Next CCITTFaxDecoder user need to follow the same interface of `{ getByte() { ... }}`

yurydelendik (approved):
Yes, that's what I had in mind. It's good to go after comments addressed.

janpe2:
I had a reason why I "faked" a Stream object in my PR #8991: I need to know the byte offset at which the CCITTFax decoding ends. When decoding MMR-encoded HalftoneRegions, there are many MMR images in a stream and they are separated only by EOFB codes. The JBIG2 data doesn't specify the start and end offset of each MMR halftone image. When I use the same Stream object to decode all the MMR bitmaps of a HalftoneRegion, the Stream remembers the end offset of each MMR image, so the decoding of the next one can continue right after it.

Would it be possible to retain this behavior in the new `CCITTFaxDecoder`? Or do you suggest the JBIG2 decoder should pass a `Uint8Array` to the `CCITTFaxDecoder` constructor. This makes it difficult to return the offset at which the decoding ended.

src/core/ccitt.js
Snuffleupagus: Sure, fixed now.

src/core/ccitt_stream.js
Snuffleupagus: > I was thinking we could wrap str into source above, e.g. const source = { next() { return str.getByte(); } };, but this will work fine for now.

That seems like a nice idea, so I figured that it cannot hurt to implement that while we're at it.

src/core/ccitt_stream.js
Snuffleupagus: I think that the old code was actually wrong here, since `lookChar` used to return `null` when EOF was reached. Hence we would actually attempt to insert `null` into `this.buffer` here, since the `!this.eof` check didn't occur until the *next* loop iteration.

In the new code, we first of all need to set `this.eof = true;` in the `CCITTFaxStream`, since various stream methods expect that property to exist. Secondly, with the refactoring `readNextChar` will return `-1` in the EOF case, and we *really* don't want insert that into `this.buffer`. (Considering that setting a `uint8array[i] = -1` will lead to `uint8array[i} === 255`.)

Snuffleupagus:
@janpe2 Regarding https://github.com/mozilla/pdf.js/pull/9046#issuecomment-339002243:

The latest version of the PR now implements https://github.com/mozilla/pdf.js/pull/9046#discussion_r146559017, see also https://github.com/mozilla/pdf.js/pull/9046/files#diff-15fc1c0e1016b6c65b03effdef7cc5d9R17.
Does this help, in any way, with addressing the use-case you describe (since it should allow you to keep track position in the `source` object that you provide)?

janpe2:
Yes, looks good. Thank you.

PR #9050, opened 2017-10-21
Enable "read with streaming" unit test for browsers that support streaming

### LINKED ISSUE #8851: Enable "read with streaming" test in test/unit/network_spec.js
The "read with streaming" test looks like this:
https://github.com/mozilla/pdf.js/blob/a4cc85fc5f5cb144b6ee7a7978dc83e906677233/test/unit/network_spec.js#L67-L77

There are two issues here:
- Lines 74 - 76 should use `pending('Streaming not supported by user agent');` instead of an incorrect `done()` call.
- With #8768, it seems that the test can be enabled for other browsers (Chrome).

@yurydelendik Can the test be enabled for other browsers?

aishwarya-sankhla:
Fixes #8851.

### DIFF
--- a/test/unit/network_spec.js
+++ b/test/unit/network_spec.js
@@ -72,7 +72,7 @@ describe('network', function() {
     var m = /Mozilla\/5.0.*?rv:(\d+).*? Gecko/.exec(userAgent);
     if (!m || m[1] < 9) {
       expect(true).toEqual(true);
-      done();
+      pending('Streaming not supported by user agent');
       return;
     }
 
--- a/test/unit/network_spec.js
+++ b/test/unit/network_spec.js
@@ -69,8 +69,9 @@ describe('network', function() {
     // The test is valid for FF only: the XHR has support of the
     // 'moz-chunked-array' response type.
     // TODO enable for other browsers, e.g. when fetch/streams API is supported.
-    var m = /Mozilla\/5.0.*?rv:(\d+).*? Gecko/.exec(userAgent);
-    if (!m || m[1] < 9) {
+    var isFirefoxWithMozChunkedEncodingSupport = /Mozilla\/5.0.*?rv:(\d+).*? Gecko/.exec(userAgent);
+    var isFetchWithStreamSupport = typeof ReadableStream;
+    if (!isFirefoxWithMozChunkedEncodingSupport && isFetchWithStreamSupport == 'undefined') {
       pending('Streaming not supported by user agent');
     }
 
--- a/test/unit/network_spec.js
+++ b/test/unit/network_spec.js
@@ -71,9 +71,7 @@ describe('network', function() {
     // TODO enable for other browsers, e.g. when fetch/streams API is supported.
     var m = /Mozilla\/5.0.*?rv:(\d+).*? Gecko/.exec(userAgent);
     if (!m || m[1] < 9) {
-      expect(true).toEqual(true);
       pending('Streaming not supported by user agent');
-      return;
     }
 
     var stream = new PDFNetworkStream({


Snuffleupagus:
> I have replaced done with pending as stated in the first part of the bug.

This isn't quite right, since quoting from https://github.com/mozilla/pdf.js/issues/8851#issue-254271886:
> Lines 74 - 76 should use `pending('Streaming not supported by user agent');` instead of an incorrect done() call.

That explicitly states that you should be replacing lines https://github.com/mozilla/pdf.js/blob/56c14e27e648d87375e89c9c4ee895c93630f3fe/test/unit/network_spec.js#L74-L76

---

> But was unable to do the second part. Could you please elaborate further on it, particularly on the second variable-ReadableStream?

Did you carefully read https://github.com/mozilla/pdf.js/issues/8851#issuecomment-326300936, since that refers to a commit (see e.g. cd95b42) which should help?
> For the second part, we need to update the if condition to logic along the lines of "if (!isFirefoxWithMozChunkedEncodingSupport && !isFetchWithStreamSupport)` (choose shorter but understandable variable if you wish). The first variable name is based on the current content of the if-statement (by the way, that has a flaw too. If Firefox 90 becomes available, the condition is incorrect). The second variable should indicate whether ReadableStream is supported (see e.g. cd95b42).

---

Finally, once this is ready for review, please remember to squash the commit (see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits). Also please make sure that the commit message (and PR title) actually describe what's changed and why, since "Replace done with pending" isn't really help for understanding the changes without looking at the code.

aishwarya-sankhla:
var isFirefoxWithMozChunkedEncodingSupport = /Mozilla\/5.0.*?rv:(\d+).*? Gecko/.exec(userAgent);
    var isFetchWithStreamSupport = typeof ReadableStream;
    if (!isFirefoxWithMozChunkedEncodingSupport && isFetchWithStreamSupport == 'undefined') {
      pending('Streaming not supported by user agent');
    }

I wanted to make sure the code is right before I made pull request.
So please let me know if the above code is correct or if there are any other changes to be made.

timvandermeij:
Could you fix the issues mentioned by Travis CI below and after that squash all commits into one commit? Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for easy steps if you're not familiar with it.

test/unit/network_spec.js
```diff
@@ -69,11 +69,10 @@ describe('network', function() {
     // The test is valid for FF only: the XHR has support of the
```
timvandermeij: These comments can now be removed, including the TODO you fixed with this patch.

Snuffleupagus:
> Could you fix the issues mentioned by Travis CI below and after that squash all commits into one commit? Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for easy steps if you're not familiar with the process.

I'm not sure if the only thing missing here is just a squash, since the patch still doesn't look quite right unfortunately. It's not really clear to me how to respond to https://github.com/mozilla/pdf.js/pull/9050#issuecomment-339073027, without basically writing the entire patch through review comments.

test/unit/network_spec.js
```diff
@@ -69,11 +69,10 @@ describe('network', function() {
...
     // 'moz-chunked-array' response type.
     // TODO enable for other browsers, e.g. when fetch/streams API is supported.
-    var m = /Mozilla\/5.0.*?rv:(\d+).*? Gecko/.exec(userAgent);
-    if (!m || m[1] < 9) {
-      expect(true).toEqual(true);
-      done();
-      return;
+    var isFirefoxWithMozChunkedEncodingSupport = /Mozilla\/5.0.*?rv:(\d+).*? Gecko/.exec(userAgent);
```
Snuffleupagus: The use of `is` in the variable name suggest that it should contain a boolean, however that isn't what's happening here. What you probably want is to adjust which regexp function you're using:
```javascript
let isFirefoxWithMozChunkedEncodingSupport =
    /Mozilla\/5.0.*?rv:(\d+).*? Gecko/.exec(userAgent);
```

test/unit/network_spec.js
```diff
@@ -69,11 +69,10 @@ describe('network', function() {
...
+    var isFetchWithStreamSupport = typeof ReadableStream;
```
Snuffleupagus: Again, this variable doesn't actually contain a boolean. Furthermore, this condition doesn't look right (compare with https://github.com/mozilla/pdf.js/commit/cd95b426c730108a529fbf6ab48b3192e22922dd). Hence you'd want something like this:

```javascript
let isFetchWithStreamSupport = (typeof Response !== 'undefined' &&
                                'body' in Response.prototype &&
                                typeof ReadableStream !== 'undefined');

File: test/unit/network_spec.js [hunk 4]
```diff
@@ -69,11 +69,10 @@ describe('network', function() {
...
+    if (!isFirefoxWithMozChunkedEncodingSupport && isFetchWithStreamSupport == 'undefined') {
```
Snuffleupagus: `if (!isFirefoxWithMozChunkedEncodingSupport && !isFetchWithStreamSupport) {`

timvandermeij:
Closing since this is superseded by #9268.

PR #9052, opened 2017-10-21
Attempt to clarify the meaning of "extension" in the ISSUE_TEMPLATE

Snuffleupagus:
Based on a number of opened issues, it seems that the "Is an extension" field might be causing some confusion as to its meaning. Without providing too much detail, I'm still thinking that we could attempt to clarify that it's referring to either of the *browser* extensions.


timvandermeij:
Indeed that does look better. Thanks.

PR #9054, opened 2017-10-21
Replace our `Promise` polyfill with the one from core-js

Snuffleupagus:
Since we're already using core-js elsewhere in `compatibility.js`, we can reduce the amount of code we need to maintain ourselves.

https://github.com/zloirock/core-js#ecmascript-6-promise


timvandermeij (approved):

timvandermeij:
Nice find!

PR #9056, opened 2017-10-21
Removed the curves from previous and next page buttons

ghost:
As they were looking like redo and undo buttons therefore causing confusion to users


thisisabhinay:
@Snuffleupagus I am new to open source projects. I saw you added a label **4-work-in-progress**. What does it actually mean? Is my commit incomplete?

Snuffleupagus:
Personally, I've never found the existing images to be confusing. That said, let's see what the general consensus is among the devs, so that we can decide if we want to make this change.

However, even if it's decided to change this, there's two things that needs to be addressed:
 - The existing images are compressed with `zopflipng`, and the new ones should be as well.
 - Since the new images looks to be symmetrical, I'm assuming that there's no need for separate images for RTL locales any more!?

thisisabhinay:
@Snuffleupagus Regarding the first point - No I haven't compressed it with zopflipng. And regarding the second point --
Yes, there is no need for separate images for RTL locales anymore. What I did was - I replaced both icons of LTR and RTL with the same image for both up and down arrows. I didn't change references for both types of images in code as it would have increased the chances of conflict.

timvandermeij:
I don't really have the confusion either, but I'm also fine with the new icons. However, aside from the `zopflipng` and LTR/RTL points mentioned, I do see some color differences in the patched area of the image. It's more clear if you zoom to 200% in your browser (which is equivalent to HiDPI screens).

timvandermeij:
I'm going to close this since I also don't really see the confusion, the current arrows are the original design and there was no further response here.

PR #9062, opened 2017-10-24
Move char codes from high surrogate pair range into private use.

### LINKED ISSUE #2884: PDF text missing
http://www.chotatsu.e-aichi.jp/portal/pdf/1/4.6.1.pdf
This is broken by 78213e8. Why is this heuristic added?

brendandahl:
Fixes #2884 

In the test PDF, the glyph that draws '4' has a unicode value of 0xD860 (which is not a valid unicode value). This value seems to have issues with Firefox and Chrome rendering it in that position.

I was hoping to stop playing whack-a-mole with the problematic unicode values and instead come up with a white list, but this is going to need more investigation. Stepping through Firefox's fillText canvas command I couldn't find an obvious spot where Firefox decided if it will draw a glyph at a certain char code. I'm thinking it's partially Firefox and the various font backends, but I'm not sure. 

As another way to try and come up with a white list, I wrote a [test script](http://brendandahl.github.io/sandbox/draw_all_char_codes/) that uses [a font](http://blogs.adobe.com/CCJKType/2012/05/breaking-64k-glyph-barrier.html) that maps all char codes to one glyph and then attempted to draw all char codes from 0 to 0xFFFF. Unfortunately, Firefox and Chrome have different behavior. Firefox seems to have issues with lots of ranges, while Chrome doesn't have any issues. For Chrome, what's really strange is 0xD860 seems to work fine with the test script, but it also fails to render the attached PDF with pdf.js, so what Chrome draws may be dependent on font type.


brendandahl:
Looks like I need to fix some unit tests, but in the mean time...

/botio test

Snuffleupagus:
> I was hoping to stop playing whack-a-mole with the problematic unicode values and instead come up with a white list, but this is going to need more investigation. 

@brendandahl Do you want to fix the test failures, such that we can land this PR as a temporary solution?
Or should we close this for now, and wait for a more complete solution instead (one based on the findings in http://brendandahl.github.io/sandbox/find_matching_glyphs/)?

brendandahl:
@Snuffleupagus I'll fix failures as a temporary solution. I started working on the full solution, but when I start moving more glyphs to the private use area it causes more problems. If a glyph has an id of 0 (usually notdef, but in some cases a valid glyph) and it gets moved the private use area it no longer will be drawn by the browser. What we really need to do is shift all the glyphs and insert a fake notdef, but that's a non-trivial amount of work.

Snuffleupagus:
@brendandahl Ping :-)

test/unit/fonts_spec.js
```diff
@@ -83,7 +83,7 @@ var checkProblematicCharRanges = function checkProblematicCharRanges() {
 
 describe('Fonts', function() {
   it('checkProblematicCharRanges', function() {
-    var EXPECTED_PERCENTAGE = 45;
+    var EXPECTED_PERCENTAGE = 100;
```
Snuffleupagus: Is it strictly necessary to make this a full 100 percent, for the tests to pass with this patch?

The reason that I'm asking is that with this change, there's nothing stopping us from adding enough entries to `ProblematicCharRanges` that there'd be no room over for re-mapping glyphs that would otherwise end up at a previously mapped location. Basically, wouldn't we always hit https://github.com/mozilla/pdf.js/blob/518da6c6a9df7e388e6bc4523f21f4d2044b8459/src/core/fonts.js#L787-L790 in that case, where as if we keep the value lower than 100 this test-case would ensure that there's atleast *some* space left in the PUA?

Snuffleupagus (approved):
Looks good to me, with one question; thanks for the patch!

test/unit/fonts_spec.js
brendandahl: I think it's now at 70%. I could lower it, but it seems odd to have a test that we have to change every time the ranges change. We would only run out of space in the PUA if the font had tons of values without unicode values and lots of values with problematic unicodes (or a very large symbolic font).

Snuffleupagus: Sure, let's just do this for now!
Hopefully, we'd be able to (eventually) replace the current black-list approach with a white-list instead.

PR #9072, opened 2017-10-26
Use `stringToBytes` in more places

Snuffleupagus:
Rather than having (basically) verbatim copies of `stringToBytes` in a few places, we can simply use the helper function directly instead.


timvandermeij (approved):

timvandermeij:
Good find, thanks!

PR #9073, opened 2017-10-26
Fix the interface of `JpegStream`/`JpxStream`/`Jbig2Stream` to agree with the other `DecodeStream`s

Snuffleupagus:
Addresses https://github.com/mozilla/pdf.js/pull/9046#discussion_r146310316; please refer to the individual commit messages for additional details.

### DIFF
--- a/src/core/jpeg_stream.js
+++ b/src/core/jpeg_stream.js
@@ -1,0 +1,112 @@
+/* Copyright 2012 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createObjectURL, shadow } from '../shared/util';
+import { DecodeStream } from './stream';
+import { isDict } from './primitives';
+import { JpegImage } from './jpg';
+
+/**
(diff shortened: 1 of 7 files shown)


src/core/jpeg_stream.js
```diff
@@ -0,0 +1,117 @@
...
+
+    jpegImage.parse(this.bytes);
+    let data = jpegImage.getData(this.drawWidth, this.drawHeight,
+                                 this.forceRGB);
+    this.buffer = data;
+    this.bufferLength = data.length;
+    this.eof = true;
+  };
+
+  JpegStream.prototype.getBytes = function(length) {
```
Snuffleupagus: I'm actually not sure what purpose this methods serves, since I successfully ran all tests locally with it commented out. Furhermore, I also ran tests with an added `if (length && length !== this.bufferLength) { throw new Error('length mismatch'); }` check, and didn't get a single test failure in that case either.

Looking at the history, it seems that this code originated back in PR #4528, but as far as I can see there's no mention in either commit messages nor PR comments of why it was necessary to add a "special" `getBytes` function for the `JpegStream`.
I'm assuming that there's a good reason why this method was added, e.g. to address a *specific* regression in one of the reference tests. However, I did check out commit https://github.com/mozilla/pdf.js/commit/58f697f977f4a60974d91b09b2da67133227837c locally and ran tests with this method commented out, and there didn't seem to be any image-related failures in that case either!?

**In summary:** Do we want to simply remove this method (and accept the small risk of a potential regression in a case where we're possible lacking test-coverage)?

yurydelendik: For code moving type of refactoring I recommend to not perform any optimizations: easier on reviewer. As for streams, in general, we need to refactor stream interface to remove reading of primitive types from it: read only chunk of available bytes, and reader class for reading bytes or int32.

Snuffleupagus: Sure, that makes a lot of sense!
I've left that method intact for now, and the PR should be ready for review now :-)

timvandermeij (approved):

timvandermeij:
Looks good!

PR #9078, opened 2017-10-29
Update ESLint and enable the `lines-between-class-members` rule

Snuffleupagus:
This rule will help aid readability in `class`es, please see https://eslint.org/docs/rules/lines-between-class-members.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #9079, opened 2017-10-29
Docs - README.md - tweaks (thumbnail, separators, styling)

ghost:
## Docs - README.md - tweaks (thumbnail, separators, styling)

**this PR does basically aim at introducing the following tweaks to README.md file:**
- *thumbnail*
- *separator*
- *styling*

---

**main motivation behind such:**
- *improving overall document accessebility*

---

</img> </img>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,16 @@
 # PDF.js
 
-PDF.js is a Portable Document Format (PDF) viewer that is built with HTML5.
+**PDF.js** is a **Portable Document Format (PDF)** viewer that is built with **HTML5**.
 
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
First of all, there was consensus in PR #8830 that we didn't want to add separators between sections.
Secondly, I'm not sure what the purpose of adding a bunch of (only tangentially relevant) images at the end of the README file really is (and *if* we'd want an image, it probably ought to be the PDF.js logo placed inline with PDF.js title of the file instead).
Finally, I'm finding the use of bold text all over the place like this quite distracting and in many cases even fairly confusing. 

Closing, based on the above.

PR #9082, opened 2017-10-30
Overwrite glyphs contour count if it's less than -1.

### LINKED ISSUE #7562: How to know when PDF rendering has partially failed
Is there a way to catch a case in which PDF rendering has partially failed? I can catch exceptions, but in some cases PDF.js only logs warnings into console and even those cases can be quite fatal for end user experience.

For example, we have one invoice PDF which gives the following warning in console.
`Warning: Failed to load font "g_d0_f2": SyntaxError: Invalid font data in ArrayBuffer.`

A screenshot of the outcome rendered in PDF.js version 1.5.188 in Chrome 52.0.2743.116 (64-bit) on Ubuntu 14.04 looks like this:
http://public.humap.com/images/pdf-font-loading-failed.png

Clearly the PDF is somewhat flawed. Still, some viewers can show the PDF properly (such as Chrome built-in PDF viewer and Ubuntu Document Viewer).

Instead of showing the flawed PDF in browser I would like to show an error message to user and provide just a link to download the PDF file. So how could I do that?

The original PDF file contains secure information, but I can deliver it to developers in case someone wants to take a look at this particular file.

brendandahl:
The test pdf has a contour count of -70, but OTS doesn't
like values less than -1.

Fixes #7562.


brendandahl:
Off by one!
/botio test

Snuffleupagus (approved):
Really nice find; especially since it brings this code nicely in line with [the specification](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6glyf.html), which clearly states that `contoursCount` should be a *signed* integer and even recommends using `-1` for compound glyphs!

r=me, with the `test_manifest.json` file updated to actually include the new test :-)

PR #9088, opened 2017-10-31
For non-embedded fonts, map softhyphen (0x00AD) to regular hyphen (0x002D) (issue 9084)

### LINKED ISSUE #9084: hyphen / minus sign not displayed
The attached pdf file contains text with several hyphen or minus signs:
[hyphen_not_displayed.pdf](https://github.com/mozilla/pdf.js/files/1427458/hyphen_not_displayed.pdf)

When I open the pdf in firefox (version 56.0.2) or in any browser via an html page with pdf.js viewer (version 1.9.426), the hyphen is not displayed.
The hyphen is correctly displayed in:
- Adobe Reader XI version 11.0.22
- Chrome version  61.0.3163.100

When I open the pdf in an html page with a pdfjs viewer in chrome, I can see in the developer tools that the hyphen is converted to `&shy;` (soft hyphen).

There are several places on the first page where the hyphen is not displayed:
1. Customer reference number  0046071415-­0060/74883
2. Steel grade  SS 2541-­03 M
3. CHARPY V2/-­20 C EN ISO 148­-1
4. TENSILE TEST EN ISO 6892­-1

Snuffleupagus:
In the PDF file, the `ToUnicode` data first maps the hyphen correctly, and then *overwrites* it to point to the softhyphen instead. That one cannot be rendered in browsers, and an empty space thus appear instead.

Fixes #9084.


brendandahl (approved):

timvandermeij:
Thank you for fixing this!

PR #9089, opened 2017-10-31
Extracts OperatorList class and prepares for streaming

yurydelendik:
Now OperatorList will be optimized all the time and no need to optimize it before we decide to send with large chunk, it can be send in smaller portions (e.g. based on time slice).


yurydelendik:
Somewhat expected this type of failures -- different operator list painting for e.g. paintInlineImageXObjectGroup . We need to check if we can avoid intermittent failures when streaming will be used.

brendandahl:
/botio test

Checking to see if test failures stay the same...

brendandahl:
Looks like there was some flakyness on windows:

! | chrome-issue4926-page1
! | chrome-issue4926-built-in-jpg-page1

brendandahl (approved):

PR #9091, opened 2017-10-31
Added Drawing rectangle and Zoomin

### LINKED ISSUE #1260: Marquee Zoom Feature
A marquee zoom tool would be a useful feature.  Having a zoom selection style similar to how "Print Edit 7.1" does their selection box is visually appealing.

vikasmahato:
Fixes #1260
This PR implements the Marquee Zoom Feature. However, implementations for setting the scrollbar position is currently missing. Any help with that would be greatly appreciated.


timvandermeij:
This hasn't really been on our radar unfortunately and it's not in a mergeable state. For now there is not really priority for this anymore, but we will keep the associated issue open. Thank you.

PR #9095, opened 2017-11-01
Split the existing `WebGLUtils` in two classes, a private `WebGLUtils` and a public `WebGLContext`, and utilize the latter in the API to allow various code to access the methods of `WebGLUtils`

Snuffleupagus:
This patch is one (small) step on the way to reduce the general dependency on a global `PDFJS` object, for PDF.js version `2.0`.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -30,6 +30,7 @@ import { CanvasGraphics } from './canvas';
 import globalScope from '../shared/global_scope';
 import { Metadata } from './metadata';
 import { PDFDataTransportStream } from './transport_stream';
+import { WebGLContext } from './webgl';
 
 var DEFAULT_RANGE_CHUNK_SIZE = 65536; // 2^16 = 65536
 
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
@yurydelendik Review ping; can you please check if the interface of the new `WebGLFactory` looks OK?

Snuffleupagus:
It looks like the "usual" intermittent errors on Windows, but let's check once more...

/botio-windows test

yurydelendik (approved):
Overall looks good, thanks for refactoring. IMHO "Factory" is not appropriate name, e.g. "ZZZFactory" have to create "ZZZ" object. Will `WebGLContext` be a better name choice here?

Snuffleupagus:
> IMHO "Factory" is not appropriate name, e.g. "ZZZFactory" have to create "ZZZ" object. Will WebGLContext be a better name choice here?

Good point, I've change the name as suggested; thanks for the review!

PR #9097, opened 2017-11-02
[WIP] Move viewer related option from the `PDFJS` object to their respective viewer components

Snuffleupagus:
*Please note that this is currently blocked on PR #9095 landing (for the `WebGL` changes).*

This PR is a *first* step on the way to reducing our dependency on the global `PDFJS` object, in order to see how feasible this is to implement, and to determine if the general format seems reasonable thus far.

***Edit:*** The more I think about this, the more unsure of the current approach I become. It seems to me that doing it this way could very well make the default viewer *more* difficult to configure (which wasn't really the goal), especially when you also consider doing a similar conversion for the API parameters!?

While it probably won't solve the general (i.e. API configuration) problem, one idea would perhaps be to utilize the `Preferences` to an even larger degree, in an attempt to simplify the mess in [`PDFViewerApplication._readPreferences`](https://github.com/mozilla/pdf.js/blob/0052dc2b0dfec6d7784e215376ae231f72dda420/web/app.js#L195) w.r.t. what settings actually takes precedence!?

Another thing, which may or may not be an issue, is that in `master` it's currently possible to fairly easily *override* values set in `src/shared/compatibility.js`. With the way this patch is written, that's no longer possible. Of course, one could argue that this isn't a problem and that we shouldn't allow users to shoot themselves in the foot like that, but I'd assume that there're situations where it may be useful nonetheless.

So, in summary: I'd appreciate other ideas for reducing our dependency on the global `PDFJS` object.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -701,6 +701,8 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
  *                                calling of PDFPage.getViewport method.
  * @property {string} intent - Rendering intent, can be 'display' or 'print'
  *                    (default value is 'display').
+ * @property {boolean} enableWebGL - (optional) Enables WebGL accelerated
+ *   rendering for some operations. The default value is `false`.
  * @property {boolean} renderInteractiveForms - (optional) Whether or not
  *                     interactive form elements are rendered in the display
  *                     layer. If so, we do not render them on canvas as well.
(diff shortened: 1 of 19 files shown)


Snuffleupagus:
I've started prototyping another way of doing the viewer configuration, that hopefully will help avoid some of problems with the approach in this PR.
Time permitting, I'm hoping to have something working sometime next week; hence closing this for now.

src/display/api.js
```diff
@@ -701,6 +701,8 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
  *                                calling of PDFPage.getViewport method.
  * @property {string} intent - Rendering intent, can be 'display' or 'print'
  *                    (default value is 'display').
+ * @property {boolean} enableWebGL - (optional) Enables WebGL accelerated
```
wojtekmaj: This might come in handy: http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

PR #7972, opened 2017-01-19
Enable the `no-unused-vars` ESLint rule

Snuffleupagus:
Please see http://eslint.org/docs/rules/no-unused-vars; note that this patch purposely uses the same rule options as in `mozilla-central`, such that it fixes part of issue #7957.

It wasn't, in my opinion, entirely straightforward to enable this rule compared to the already existing rules. In many cases a `var descriptiveName = ...` format was used (more or less) to document the code, and I choose to place the old variable name in a trailing comment to not lose that information.
~~In other cases I simply commented out the "unused" lines instead of outright removing them, since that's already done in some places in the code-base to preserve the logic for future use. (Since comments are stripped by the preprocessor anyway, that shouldn't be a problem IMHO.)~~

I welcome feedback on these changes, since it wasn't always entirely easy to know what changes made the most sense in every situation.

PR #9098, opened 2017-11-03
Fix issues found by lgtm

xiemaisi:
This fixes a few minor JavaScript issues found by [lgtm](https://lgtm.com); for more details see https://lgtm.com/projects/g/mozilla/pdf.js/.

Most of the fixes are quite straightforward and basically cosmetic, except for the one in the last commit where I took a (hopefully) educated guess at the correct fix from context.

Among the remaining alerts, [this one](https://lgtm.com/projects/g/mozilla/pdf.js/alerts/?mode=tree&severity=error&rule=7890078) is perhaps worth a closer look. I couldn't tell whether it is actually wrong, and if so how to fix it.

(Full disclosure: I work on lgtm's JavaScript analysis.)

### DIFF
--- a/src/core/cmap.js
+++ b/src/core/cmap.js
@@ -588,7 +588,7 @@ var BinaryCMapReader = (function BinaryCMapReaderClosure() {
             stream.readHex(start, dataSize);
             stream.readHexNumber(end, dataSize);
             addHex(end, start, dataSize);
-            code = stream.readNumber();
+            stream.readNumber(); // code
             // undefined range, skipping
             for (i = 1; i < subitemsCount; i++) {
               incHex(end, dataSize);
@@ -596,7 +596,7 @@ var BinaryCMapReader = (function BinaryCMapReaderClosure() {
               addHex(start, end, dataSize);
               stream.readHexNumber(end, dataSize);
               addHex(end, start, dataSize);
-              code = stream.readNumber();
+              stream.readNumber(); // code
               // nop
             }
             break;
--- a/src/core/crypto.js
+++ b/src/core/crypto.js
@@ -860,7 +860,7 @@ var AES128Cipher = (function AES128CipherClosure() {
   }
 
   function encrypt128(input, key) {
-    var t, u, v, k;
+    var t, u, v, j, k;
     var state = new Uint8Array(16);
     state.set(input);
     for (j = 0; j < 16; ++j) {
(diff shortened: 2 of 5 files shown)


Snuffleupagus (commented):
> except for the one in the last commit where I took a (hopefully) educated guess at the correct fix from context.

Rather than putting the *entire* responsibility of testing/verification solely on the reviewer, please test this yourself to ensure that there's no unforeseen side-effects. Please refer to https://github.com/mozilla/pdf.js/wiki/Debugging-pdf.js for information about the debugging tools.

See also https://github.com/mozilla/pdf.js/wiki/Contributing for information about running the various test-suites locally, which is always a good idea for patches that touch code in the `/src` folder.

> Among the remaining alerts, this one is perhaps worth a closer look. I couldn't tell whether it is actually wrong, and if so how to fix it.

That one is correct, since that is *not* a normal variable, but a getter/setter:
https://github.com/mozilla/pdf.js/blob/6521d2fd941f26a1b7ae9b97f71842e49f5ff241/web/base_viewer.js#L244-L259

xiemaisi:
Thanks for your feedback.

> Rather than putting the entire responsibility of testing/verification solely on the reviewer, please test this yourself to ensure that there's no unforeseen side-effects.

Understood. I've removed the final commit for now, I'll make a separate PR when I get time to test it.

> That one is correct, since that is not a normal variable, but a getter/setter:

Ah ha! I'll improve the analysis to pick up on that. (We already whitelist self-assignments for accessors of object literals, I just had neglected to consider accessor methods.)

> Rather than losing that, I thus think that we should do as in PR #7972 and add the variable name as a trailing comment; i.e.

Done (in both places).

Snuffleupagus:
> Address review comments.

Please note that, in order to keep the history cleaner, we don't use separate commits for addressing review feedback; hence "Address review comments." needs to be folded into the commit where it belongs.

xiemaisi:
>  hence Address review comments. needs to be folded into the commit where it belongs.

Done.

Snuffleupagus (approved):
Thanks for the patch!

xiemaisi:
My pleasure!

Perhaps you would like to take a look at our [CI integration](https://lgtm.com/docs/lgtm/using-lgtm-analysis-continuous-integration)? If you turn it on, every GitHub pull request is analysed and information about new and fixed alerts is displayed as part of the PR status. Here is an example of what it looks like in practice: https://github.com/Open-MBEE/MDK/pull/132#issuecomment-331294008

PR #9099, opened 2017-11-03
Allows 'dontFlip' as third arg in PDFPageProxy.getViewport

tiriana:
Right now if you wanted to have Viewport with 'dontFlip' you need to get the viewport from page with `getViewport(scale, rotate)` and then clone it with `.clone({ dontFlip: true })`. 

This PR allows to do it like this:

```
page.getViewport(scale, rotate, dontFlip)
```

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -783,14 +783,12 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
      * @param {number} scale The desired scale of the viewport.
      * @param {number} rotate Degrees to rotate the viewport. If omitted this
      * defaults to the page rotation.
+     * @param {boolean} dontFlip (optional) If true, axis Y will not be flipped.
      * @return {PageViewport} Contains 'width' and 'height' properties
      * along with transforms required for rendering.
      */
-    getViewport: function PDFPageProxy_getViewport(scale, rotate) {
-      if (arguments.length < 2) {
-        rotate = this.rotate;
-      }
-      return new PageViewport(this.view, scale, rotate, 0, 0);
+    getViewport(scale, rotate = this.rotate, dontFlip = false) {
+      return new PageViewport(this.view, scale, rotate, 0, 0, dontFlip);
     },
     /**
      * @param {GetAnnotationsParameters} params - Annotation parameters.
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -979,6 +979,18 @@ describe('api', function() {
       expect(viewport.width).toEqual(1262.835);
       expect(viewport.height).toEqual(892.92);
     });
+    it('gets viewport respecting "dontFlip" argument', function () {
+      const scale = 1;
+      const rotation = 135;
+      let viewport = page.getViewport(scale, rotation);
+      let dontFlipViewport = page.getViewport(scale, rotation, true);
+
+      expect(dontFlipViewport).not.toEqual(viewport);
+      expect(dontFlipViewport).toEqual(viewport.clone({ dontFlip: true, }));
+
+      expect(viewport.transform).toEqual([1, 0, 0, -1, 0, 841.89]);
+      expect(dontFlipViewport.transform).toEqual([1, 0, -0, 1, 0, 0]);
+    });
     it('gets annotations', function (done) {
       var defaultPromise = page.getAnnotations().then(function (data) {
         expect(data.length).toEqual(4);


src/display/api.js
```diff
@@ -786,11 +786,11 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
      * @return {PageViewport} Contains 'width' and 'height' properties
      * along with transforms required for rendering.
      */
-    getViewport: function PDFPageProxy_getViewport(scale, rotate) {
-      if (arguments.length < 2) {
-        rotate = this.rotate;
+    getViewport: function PDFPageProxy_getViewport(scale, rotate, dontFlip) {
```
Snuffleupagus: Since we now live in an ES6 world, can't we simplify things by using the following method signature instead?
```javascript
getViewport(scale, rotate = this.rotate, dontFlip = false) {
```

Snuffleupagus (changes requested):
When adding a new parameter, you'll need to update the JSDoc comment accordingly.

Please also add a *new* unit-test in https://github.com/mozilla/pdf.js/blob/master/test/unit/api_spec.js that uses the new parameter.

Finally, when updating the patch, don't forget to [squash the commits](https://github.com/mozilla/pdf.js/wiki/Squashing-Commits). And please make sure that the *commit message* makes sense, since there seems to be words missing in `Allows as third arg in PDFPageProxy.getViewport` (when comparing with the PR title).

src/display/api.js
tiriana: We can, I just didn't know if I should not stay with es5 here.

- [ ] use es6 default params here.

tiriana:
@Snuffleupagus Will do so. Thanks for feedback.

tiriana:
Just for the record, I still remember about this PR, just could find any time to finish it. I plan to do this shortly.

tiriana:
@Snuffleupagus Fixes pushed

src/display/api.js
```diff
@@ -783,14 +783,15 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
      * @param {number} scale The desired scale of the viewport.
      * @param {number} rotate Degrees to rotate the viewport. If omitted this
      * defaults to the page rotation.
+     * @param {boolean} dontFlip if true, axis Y will not be flipped.
```
Snuffleupagus: Apparently the existing JSDoc looks a bit weird here, but let's fix the *new* one added here:
```javascript
@param {boolean} dontFlip - (optional) If true, axis Y will not be flipped.
```

src/display/api.js
```diff
@@ -783,14 +783,15 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
      */
-    getViewport: function PDFPageProxy_getViewport(scale, rotate) {
-      if (arguments.length < 2) {
-        rotate = this.rotate;
-      }
-      return new PageViewport(this.view, scale, rotate, 0, 0);
+    getViewport: function PDFPageProxy_getViewport(
+        scale,
+        rotate = this.rotate,
+        dontFlip = false) {
```
Snuffleupagus: We don't use that kind of formatting anywhere; so as previously suggested, please just replace the *entire* method signature like this (which is valid in ES6):
```javascript
getViewport(scale, rotate = this.rotate, dontFlip = false) {
```

test/unit/api_spec.js
```diff
@@ -979,6 +979,13 @@ describe('api', function() {
       expect(viewport.width).toEqual(1262.835);
       expect(viewport.height).toEqual(892.92);
     });
+    it('gets viewport respecting "dontFlip" argument', function () {
+      var flipped = page.getViewport(1.5, 135);
```
Snuffleupagus: Nit: Please use `let` instead of `var` in new code.

test/unit/api_spec.js
```diff
@@ -979,6 +979,13 @@ describe('api', function() {
...
+      var notFlipped = page.getViewport(1.5, 135, true);
+
+      expect(flipped.transform[2]).toEqual(-notFlipped.transform[2]);
+      expect(flipped.transform[3]).toEqual(-notFlipped.transform[3]);
```
Snuffleupagus: Perhaps a bit clearer, and more complete test coverage, if you explicitly compare against the expected values (similar to what's done in the test above):
```javascript
expect(flipped.transform).toEqual([/* values here... */]);
expect(notFlipped.transform).toEqual([/* values here... */]);
```

Snuffleupagus (changes requested):
Thanks for updating the patch!

src/display/api.js
tiriana: ok

- [ ] update jsDOC

src/display/api.js
tiriana: Linter will fail on this line as it extends 80 characters. Should I add some comment to make linter ignore this line?

test/unit/api_spec.js
tiriana: ok

- [ ] switch to es6

test/unit/api_spec.js
tiriana: ok
- [ ] compare whole transform arrays

src/display/api.js
Snuffleupagus: > Linter will fail on this line as it extends 80 characters.

No it won't, if you actually change it *exactly* as asked :-)
For the method signature, your diff should contain something like this (not including other relevant changes):
```diff
diff --git a/src/display/api.js b/src/display/api.js
index bd4d404b..1fdeb803 100644
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -783,7 +783,7 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
      * @return {PageViewport} Contains 'width' and 'height' properties
      * along with transforms required for rendering.
      */
-    getViewport: function PDFPageProxy_getViewport(scale, rotate) {
+    getViewport(scale, rotate = this.rotate, dontFlip = false) {
       if (arguments.length < 2) {
         rotate = this.rotate;
       }

```

tiriana: ahh, right. I left function name `PDFPageProxy_getViewport` in the first place. Ok, I'll change it.

- [ ] Change function notation

tiriana:
Just for the record - I'm still on it. It's just that end of the year is pretty crazy, and it slips my mind to do these last polishing. 

I'll try to do this this week.

tiriana:
Just pushed 8fc25f8a793d5ceb129db4a185049fc8d1f564a1 - I took a bit different approach in tests that the one proposed, but I think it's the right one, so be sure that if works exactly the same as method used to this time - to clone viewport with 'dontFlip' param

tiriana:
forgot to squash - now it's done.

test/unit/api_spec.js
```diff
@@ -979,6 +979,15 @@ describe('api', function() {
...
     });
+    it('gets viewport respecting "dontFlip" argument', function () {
+      const scale = 1.5;
+      const rotation = 135;
+      const flipped = page.getViewport(scale, rotation);
+      const unflipped = flipped.clone({ dontFlip: true, });
+      const notFlippedFromTheStart = page.getViewport(scale, rotation, true);
+
+      expect(notFlippedFromTheStart.transform).toEqual(unflipped.transform);
+    });
```
Snuffleupagus: Personally I'd actually like to see somewhat more complete tests here, to ensure that what the API returns is *exactly* the expected result (especially since this unit-test is the only place the code-path used), given that you're *only* checking the `transform` property and nothing else here.

Hence I'd prefer something along the following lines instead.
```javascript
let viewport = page.getViewport(scale, rotation);
let dontFlipViewport = page.getViewport(scale, rotation, true);

expect(viewport.transform).toEqual([/* explicit transform array here */]);
expect(dontFlipViewport.transform).toEqual([/* explicit transform array here */]);

expect(dontFlipViewport).toEqual(viewport.clone({ dontFlip: true, }));
```

tiriana: Pls check now, I pushed changes.

tiriana:
@Snuffleupagus I pushed tests changes.

Snuffleupagus (approved):
Thanks for the patch!

PR #9106, opened 2017-11-05
Enable the `no-var` ESLint rule in the `/web` folder

Snuffleupagus:
- Use ES6 notation, and replace `var` with `let`, in `web/pdf_viewer.component.js` and `web/pdfjs.js`

 - Enable the `no-var` ESLint rule in the `/web` folder

    https://eslint.org/docs/rules/no-var

    Please note that two files were excluded:
    1. `web/debugger.js`, since there's code in other files that currently depend on the global availability of code in `web/debugger.js`. Furthermore, since that file isn't used in production, doing a ES6 conversion probably isn't a priority.

   2. `web/grab_to_pan.js`, since that file could be considered to be "external" code. We have made smaller changes to that file over the years, however doing a full ES6 `class` conversion might be a step too far!?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -166,6 +166,7 @@
     "no-useless-computed-key": "error",
     "no-useless-constructor": "error",
     "no-useless-rename": "error",
+    "no-var": "off",
     "object-shorthand": ["error", "always", {
       "avoidQuotes": true,
     }],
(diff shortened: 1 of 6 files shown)


timvandermeij (approved):

timvandermeij:
Nice conversion; thanks!

PR #9110, opened 2017-11-09
Accumulate streamed PDF data into array of buffers.

brendandahl:
While investigating [bug 1410381](https://bugzilla.mozilla.org/show_bug.cgi?id=1410381) I noticed the [PDF](http://de.free.aero/contents/light2017de.pdf) took much longer to load in the Firefox extension than in the plain web version of PDF.js. After running gecko profiler, I found we spent a lot of time in the stream converter doing append(). What was happening was Firefox was loading the entire PDF before the PDF.js viewer was loaded, which means we were always following [this](http://searchfox.org/mozilla-central/rev/c99d035f00dd894feff38e4ad28a73fb679c63a6/browser/extensions/pdfjs/content/PdfStreamConverter.jsm#163) code path, where we create a new typed array to handle the old data + the new chunk and copy it over. For the above PDF that is 27MB, this was called roughly 200 times until the array was expanded to the full 27MB!

With this new code I see the above PDF (hosted locally),  go from 15s to 3s and peak-rss go from 1GB to 300MB.

@yurydelendik We had talked about using blobs, but I don't really see an advantage to that since we have to message across the data anyways. After you review, I'll ping mossop to review as well.

### DIFF
--- a/extensions/firefox/content/PdfStreamConverter.jsm
+++ b/extensions/firefox/content/PdfStreamConverter.jsm
@@ -154,7 +154,7 @@ function getLocalizedString(strings, id, property) {
 // PDF data storage
 function PdfDataListener(length) {
   this.length = length; // less than 0, if length is unknown
-  this.buffer = null;
+  this.buffers = [];
   this.loaded = 0;
 }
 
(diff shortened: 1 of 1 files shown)


yurydelendik (approved):
Looks good.

extensions/firefox/content/PdfStreamConverter.jsm
```diff
@@ -154,33 +154,42 @@ function getLocalizedString(strings, id, property) {
...
+      combinedLength += buffer.length;
+    }
+    let combinedArray = new Uint8Array(combinedLength);
+    let writeOffset = 0;
+    while (this.buffers.length) {
+      let buffer = this.buffers.shift();
+      combinedArray.set(buffer, writeOffset);
+      writeOffset += buffer.length;
+    }
+    return combinedArray;
```
Mossop: Is readData called more than once? If so should we collapse this.buffers to [combinedArray] here so that later calls don't have to recombine all the existing buffers again?

Mossop: Oh yes I see, thanks

Mossop (approved):

PR #9118, opened 2017-11-09
Added missing network stream for svgviewer example

wassupben:
The example svgviewer project currently does not render an SVG and returns an error in the console:
Uncaught (in promise) TypeError: PDFNetworkStream is not a constructor

I have forked and fixed this issue by adding in the PDFNetworkStream.


Snuffleupagus (commented):
Can you please update the *commit message* to same as the PR title?
When using e.g. the Git command line, it's otherwise difficult to tell what the patch does without actually looking at the code.

When addressing this, please don't forget to [squash the commits](https://github.com/mozilla/pdf.js/wiki/Squashing-Commits).

Snuffleupagus (approved):
Thanks for the patch!

PR #9126, opened 2017-11-13
Extract parsing of debugging hash parameters into its own method in `PDFViewerApplication`

Snuffleupagus:
In order to move viewer related options from the global `PDFJS` object and into the initialization of the relevant components, we'll need to parse the hash parameters *before* calling `PDFViewerApplication._initializeViewerComponents`.


timvandermeij (approved):

timvandermeij:
Clean solution, thanks!

PR #9140, opened 2017-11-16
Remove the `console` polyfills

Snuffleupagus:
All browsers that we intend to support with PDF.js version 2.0 already supports `console` natively.

@yurydelendik I'm hoping that this covers everything you had in mind here!


timvandermeij (approved):

timvandermeij:
Nice clean-up!

Issue #9127, opened 2017-11-13
Add an embeddable version of the default viewer


konstin:
Please add an embeddable version of the default viewer or a viewer with comparable functionality and deploy it with npm. With embeddable I mean that the viewer is some `<div>` that you can include using your favorite template engine.

I am well aware of the reasoning in #5609. I don't think that this applies to an embedded viewer as it is still clearly part of a website and not of firefox or chrome. Various websites are using the default viewer anyway as there's no real alternative. Even projects as big as nextcloud [include the default viewer](https://github.com/nextcloud/files_pdfviewer/blob/master/templates/viewer.php) instead of creating their own viewer.

Due to the lack of an official embedable viewer most sites vendor the package from the website. This means every update is manual extra work and a lot of outdated viewers, which is clearly undesirable. There is even a [project](https://github.com/legalthings/pdf.js-viewer) which attempts to deploy an embeddable version themselves, but it seems to be abandoned.

For those reasons I think it is desirable for pdf.js to deploy an easily embeddable default viewer or a comparable viewer.

Snuffleupagus:
> I don't think that this applies to an embedded viewer as it is still clearly part of a website and not of firefox or chrome.

It very much applies here, since it's essentially impossible for a "regular" user to tell if what they're seeing in an `<iframe>` is the Firefox built-in PDF Viewer or rather a version of the default viewer hosted by the web-site itself; please see http://mozilla.github.io/pdf.js/getting_started/#introduction (emphasis mine):
> The viewer is built on the display layer and is the UI for PDF viewer in Firefox and the other browser extensions within the project. It can be a good starting point for building your own viewer. ***However, we do ask if you plan to embed the viewer in your own site, that it not just be an unmodified version. Please re-skin it or build upon it.***

---

Given that there's the [viewer components examples](https://github.com/mozilla/pdf.js/tree/master/examples/components), and that the default viewer isn't really intended to be used as-is any way (as mentioned above), should this issue simply be closed as WONTFIX?

timvandermeij:
Yes, I would agree with the reasoning above. If there is any way to improve the existing examples to make usage easier, feel free to open an issue with a suggestion or create a pull request; we'll be happy to review those.

Issue #7203, opened 2016-04-15
pdfjs-dist repository does not contain every file generated by 'gulp generic'


keabard:
In pdfjs-dist repository (https://github.com/mozilla/pdfjs-dist), there are many files missing:

- `viewer.html`
- `viewer.js`
- `l10n.js`
- `debugger.js`
- locale files
- etc...

The resulting repository is therefore unusable, even with a custom viewer.html.

Is there something missing in `make.js` or am I totally missing the point ?

The guy from PDFObject uses `viewer.html` from pdf.js repository, along with other missing files, to build his example with pdf.js fallback (http://pdfobject.com/examples/pdfjs.html).

How can I do the same without the missing files ?

yurydelendik:
We are not redistributing viewer.html to avoid (visual?) confusion of third party deployments with Firefox and Chrome extensions UI, as well as our demo viewer. There is no easy way to customize viewer.html/viewer.js, so it's not in a library yet. However there is set of components that will let you build customized viewer, see pdfjs-dist/web/pdf_viewer.js file and examples at https://github.com/mozilla/pdf.js/tree/master/examples/components. Once we will confident that our viewer can be easy customized (we are slowing working towards that) and not mistaken for the extensions, we will may include the listed files in some form. Closing as answered.

yurydelendik:
See also http://mozilla.github.io/pdf.js/getting_started/

Issue #5609, opened 2015-01-02
Viewer.html is missing in pre-built version pdfs-dist


binishkaspar:
Viewer.html is missing in pre-built version pdfs-dist https://github.com/mozilla/pdfjs-dist

yurydelendik:
@binishkaspar it's on purpose. There is a concern that end users mistake Firefox PDF Viewer with custom web site deployment, which will add support load to us. While we temporary package viewer.html in the release package (http://mozilla.github.io/pdf.js/getting_started/#download), we want to make it more customizable, different from demo/Firefox viewer, and simple to use.

We are slowly refactoring the viewer to provide set of embeddable APIs, so custom viewer can be easy deployed on a custom web site, also moving burden of the UI customization on web site developers. See web/pdf_viewer files and examples at https://github.com/mozilla/pdf.js/tree/master/examples/components and #5552 

Closing as answered.

binishkaspar:
@yurydelendik Thanks for the update.

sassanh:
@yurydelendik I understand your concerns about not liking pdf-viewers on websites look like Firefox's pdf viewer, but can we have it until the "slow refactoring" is done some day?

It took me 2 hours to find out the `pdfjs-dist` is incomplete I should download that zip file instead just to have that html file. Meanwhile I gave up pdf.js few times and tried to find an alternative because a pdf viewer library that forces you to implement and maintain the code for the pdf toolbar is not a really a pdf viewer library, it's more like a pdf renderer library. I mean there should be an easy way for users to have pdf.js work as a full pdf viewer out of the box (it's there just hidden in that zip file and "purposefully" removed from `pdfjs-dist`.)

jmle:
> @yurydelendik I understand your concerns about not liking pdf-viewers on websites look like Firefox's pdf viewer, but can we have it until the "slow refactoring" is done some day?
> 
> It took me 2 hours to find out the `pdfjs-dist` is incomplete I should download that zip file instead just to have that html file. Meanwhile I gave up pdf.js few times and tried to find an alternative because a pdf viewer library that forces you to implement and maintain the code for the pdf toolbar is not a really a pdf viewer library, it's more like a pdf renderer library. I mean there should be an easy way for users to have pdf.js work as a full pdf viewer out of the box (it's there just hidden in that zip file and "purposefully" removed from `pdfjs-dist`.)

[...]

PR #9144, opened 2017-11-16
Add the viewer to the npm package

konstin:
To avoid projects vendoring pdf.js because of the missing default viewer in the npm packge, I've added the viewer to the npm package. This also is a workaround for #9127.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1352,6 +1352,10 @@ gulp.task('dist-pre',
       SINGLE_FILE_DIR + 'build/pdf.combined.js.map',
       SRC_DIR + 'pdf.worker.entry.js',
     ]).pipe(gulp.dest(DIST_DIR + 'build/')),
+    gulp.src([
+      MINIFIED_DIR + '*',
+      MINIFIED_DIR + '**/*',
+    ]).pipe(gulp.dest(DIST_DIR + 'viewer/')),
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
The default viewer, i.e. the one found at https://mozilla.github.io/pdf.js/web/viewer.html, is purposely *not* included in https://github.com/mozilla/pdfjs-dist; please refer to https://github.com/mozilla/pdf.js/issues/5609#issuecomment-68530552 and https://github.com/mozilla/pdf.js/issues/7203#issuecomment-210510569.

Please keep the following, as mentioned in e.g. https://mozilla.github.io/pdf.js/getting_started/ and https://github.com/mozilla/pdf.js/wiki/Setup-pdf.js-in-a-website, in mind:
> However, we do ask if you plan to embed the viewer in your own site, that it not just be an unmodified version. Please re-skin it or build upon it.

Hence we don't want to make it too simple to use it directly. Instead there's a number of viewer components (which are also used by the default viewer) available, that can be used to create a custom viewer.
Please find examples of using the viewer components at https://github.com/mozilla/pdf.js/tree/master/examples/components, and that the necessary files for a custom viewer is already being distributed in https://github.com/mozilla/pdfjs-dist/tree/master/web (i.e. available through e.g. NPM).

Closing, for the above reasons.

PR #9145, opened 2017-11-16
Fix TypeError: Cannot read property 'getDestinationHash' of undefined

wojtekmaj:
If someone happened to call AnnotationLayer.render(params) without giving valid linkService in params, clicking internal link was causing said TypeError.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -310,6 +310,9 @@ class LinkAnnotationElement extends AnnotationElement {
    * @memberof LinkAnnotationElement
    */
   _bindLink(link, destination) {
+    if (!this.linkService) {
+      return;
+    }
     link.href = this.linkService.getDestinationHash(destination);
     link.onclick = () => {
       if (destination) {


Snuffleupagus (changes requested):
> If someone happened to call AnnotationLayer.render(params) without giving valid linkService in params, clicking internal link was causing said TypeError.

Given that this PR is modifying the ` web/pdf_outline_viewer.js` file, I'm not sure how the above is relevant here!?
Besides, in both `PDFOutlineViewer` and `AnnotationLayer.render`, the `linkService` is actually a required parameter.

wojtekmaj:
Oh, sorry, this line appeared twice in the code. Applied fix where it should have been. My mistake.

Snuffleupagus:
> Oh, sorry, this line appeared twice in the code. Applied fix where it should have been. My mistake.

But how does failing *silently* in just *one* particular case help in anyway here?

wojtekmaj:
Say you have several annotations on a layer, first being internal link and the rest external. As far as I know, external links do not require linkService, thus have no problem rendering without it. Silently failing rendering internal link could help rendering the rest of them.

Snuffleupagus:
> As far as I know, external links do not require linkService, thus have no problem rendering without linkService.

If you use `AnnotationLayer.{render, update}`, you *must* pass in either a `PDFLinkService` (or a `SimpleLinkService`) instance (since the parameter isn't marked as `optional`).
If you search for `this.linkService` in https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js, you'll see that it's used in a number of different situations. I'm really not keen on adding similar checks all over the place, and failing silently does not seem like a good idea to me.

> Silently failing rendering internal link could help rendering the rest of them.

Maybe, but that would also provide no indication whatsoever to the user that something actually went wrong!

wojtekmaj:
Hmmm... I understand and agree. It would be nice if we could consider adding some less cryptic error in such situations though.

PR #9147, opened 2017-11-17
Updating viewer.properties

SehyunPark:
document_properties_creation_date=생성일: --> document_properties_creation_date=생성 시간:
document_properties_modification_date=수정일: --> document_properties_modification_date: 수정 시간:
(translation problem solved
--> korean word '생성일' and '수정일' mean the exact day of the specified month (e.g 13th March) and these words don't include the specified time (hour, minute and second) 
Therefore, I edited the file using '생성 시간' and '수정 시간' which include the exact time (ranging from the exact second)

### DIFF
--- a/l10n/ko/viewer.properties
+++ b/l10n/ko/viewer.properties
@@ -80,8 +80,8 @@ document_properties_title=제목:
 document_properties_author=저자:
 document_properties_subject=주제:
 document_properties_keywords=키워드:
-document_properties_creation_date=생성일:
-document_properties_modification_date=수정일:
+document_properties_creation_date=생성 시간:
+document_properties_modification_date=수정 :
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Closing, since as explained in https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md we don't maintain the translations ourselves, but rather import them from Firefox.

SehyunPark:
Thanks for your reply, but as a korean user, koreans who are using this pdf.js file might get confused because of incorrect interpretation when it comes to using this browser...

Snuffleupagus:
If you want to suggest changes to the Korean localization of Firefox, I'd recommend that you get in contact with the Korean localization team. https://wiki.mozilla.org/L10n:Teams:ko looks like a good place to start.

PR #9156, opened 2017-11-18
Update README.md

SehyunPark:
Clarifying what IRC stands for in order to make people understand clearly who want to use relay chat in irc.mozilla.org


timvandermeij:
Thank you!

PR #9158, opened 2017-11-18
Sanitize the display value for choice widget annotations

### LINKED ISSUE #8947: DisplayValue gets ugly strings
Perhaps we should wrap the code in stringToPDFString at least for the displayValue?
Otherwise I get ugly strings like "þÿAustria" instead of "Austria" in the combo box
(see attached PDF)
[OoPdfFormExample.pdf](https://github.com/mozilla/pdf.js/files/1326416/OoPdfFormExample.pdf)

E.g.
this.data.options[i] = {
          exportValue: isOptionArray ? xref.fetchIfRef(option[0]) : option,
          displayValue: isOptionArray ? xref.fetchIfRef(option[1]) : stringToPDFString(option),
        };

https://github.com/mozilla/pdf.js/blob/f01f0fdaa78bc04dbf154dab5102cc5cdabcd4ac/src/core/annotation.js#L794

timvandermeij:
Moreover, change the formatting of two other test cases to be more compact and more in line with the other test cases.

To verify this patch, use the following steps:
1. Open https://mozilla.github.io/pdf.js/web/viewer.html.
2. Open the console and enter `PDFViewerApplication.preferences.set('renderInteractiveForms', true);`.
3. Refresh the page and open the file from the issue. Notice that the dropdown has unsanitized strings.
4. Repeat these steps with the link in the preview below. Notice that the dropdown has sanitized strings.

Fixes #8947.

### DIFF
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -1092,14 +1092,8 @@ describe('annotation', function() {
 
       var options = [['foo_export', 'Foo'], optionOneRef];
       var expected = [
-        {
-          exportValue: 'foo_export',
-          displayValue: 'Foo',
-        },
-        {
-          exportValue: 'bar_export',
-          displayValue: 'Bar',
-        }
+        { exportValue: 'foo_export', displayValue: 'Foo', },
+        { exportValue: 'bar_export', displayValue: 'Bar', },
       ];
 
       choiceWidgetDict.set('Opt', options);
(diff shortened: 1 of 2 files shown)


test/unit/annotation_spec.js
```diff
@@ -1179,6 +1167,31 @@ describe('annotation', function() {
       expect(data.options).toEqual(expected);
     });
 
+    it('should sanitize display values in option arrays (issue 8947)',
+        function() {
+      // The option value is a UTF-16BE string. The display value should be
+      // sanitized, but the export value should remain the same since that
+      // may the used as a unique identifier when exporting form values.
```
Snuffleupagus: Nit: `may be used...` instead :-)

Snuffleupagus (approved):
Looks good, with one tiny grammar nit.
Thank you for the patch!

test/unit/annotation_spec.js
timvandermeij: Whoops! Fixed now.

timvandermeij:
Firefox updated from version 57 to version 58 on the Windows bot, so we need to make new reference images.

/botio-windows makeref

PR #9170, opened 2017-11-21
Remove more polyfills for old browsers

Snuffleupagus:
Note that there's a number of additional polyfills which we ought to be able to remove, but we'd need to make a final decision regarding the exact IE and Android versions we'll support with PDF.js `2.0`.

If we decide that IE 11 is the lowest supported version (which seem reasonable given https://en.wikipedia.org/wiki/Internet_Explorer_version_history#Release_history_for_desktop_Windows_OS_version), and similarly for Android 5 (given https://en.wikipedia.org/wiki/Android_version_history), then we could remove even more code.


timvandermeij (approved):

timvandermeij:

PR #9179, opened 2017-11-22
Add mgol's name to AUTHORS, add .mailmap

mgol:
In this way mgol's past contributions are mapped correctly.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,1 @@
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,6 +16,7 @@ Jonas Jenwald <jonas.jenwald@gmail.com>
 Julian Viereck
 Justin D'Arcangelo <justindarc@gmail.com>
 Kalervo Kujala
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>
 Ophir Lojkine <@lovasoa>
 Rob Wu <rob@robwu.nl>
 Shaon Barman <shaon.barman@gmail.com>


.mailmap
```diff
@@ -0,0 +1 @@
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>
```
Snuffleupagus: I'm confused, why is this necessary since this string appear to be identical to the one used below?

mgol: `.mailmap` is used by `git shortlog` (see e.g. what `git shortlog -nse` prints) & `git blame`. It affects what lots of apps outside of the Web GitHub UI present (mainly because Git respects it in many places). For example, it changes what WebStorm shows when you use its "Annotate" feature - a visual representation of `git blame` inside of the editor.

Without it I'm visible as two distinct people in Git history of some repositories because a Unicode weirdness where `ę` can be written in two ways: as a single symbol or as `e` + [Unicode Combining Ogonek](http://www.fileformat.info/info/unicode/char/0328/index.htm) ("ogonek" means "a little tail" in Polish). Apparently different tools normalize to different versions which creates the split.

If you'd like I can spend some time to map other people that have multiple "identities" according to Git, see e.g. what [I did in the AngularJS repo](https://github.com/angular/angular.js/commit/8166dd40e4a1b8bc3d682261eede798d0e0ad9b8).

Snuffleupagus: I wasn't asking what the file itself is for, I was asking if this entry was actually correct since it looked for all intents and purposes (when viewed in GitHub) like the two strings were in fact *identical*. Hence I just wondered if there was a simple typo somewhere (before checking the patch locally).

> If you'd like I can spend some time to map other people that have multiple "identities" according to Git, see e.g. what I did in the AngularJS repo.

If we do decide to do that, I don't think that it should shouldn't be part of *this* PR.

mgol: Those entries are identical on purpose... I'm afraid I don't understand the question then.

Issue #6779, opened 2015-12-18
Refactor PDFJS global var usage and move it initialization into single file


yurydelendik:
for v1.x we are going to keep supporting accessing PDFJS.xxxx members, however for future (e.g. v2.0) we shall probably rely only on module exports keeping that single file as compatibility feature.

TODO:
- [x] move all preferences from api.js into parameter object for related methods (e.g. disableRange is parameter for getDocument, and externalLinkTarget for AnnotationLayer.render)
- [x] refactor global PDFJS usage/assignments from all core/, shared/ and display/ , keep them in the single file e.g. src/display/globals.js

shikhar-scs:
Is this issue still up @yurydelendik ?

timvandermeij:
You're right, both tasks should be done already when the last pull request above is merged.

shikhar-scs:
Thanks @timvandermeij for confirming. I'll look up for something else :+1:

Snuffleupagus:
Closing, since the remaining TODO was fixed by PR #9493.

Issue #4919, opened 2014-06-11
fix webgl parameter


CodingFabian:
I just spent 20 minutes figuring out why my webgl doesn't work.
First I tried "disableWebGL=false" because in api.js this is the name of the property and it is default initialized with true.
Then i found that this value is actually overwritten by reading from Preferences.
in preferences a value "enableWebGL" is stored. once initialized the value is negated and assigned to disableWebGL.
So I tried using "enableWebGL=true". This also did not work.
Then i figured out that the Preferences are prefilled with a value read from the URL. the value read from the url is named "webgl"

Now finally after setting "webgl=true" the preference "enableWebGL" is set to true, which on init, will set API.js PDFJS.disableWebGL, which by default is inited to true, to not the value of enableWebGL, so any check for WebGLUtils.isEnabled will no longer bail out on PDFJS.disableWebGL.

I hope you get the point :-)

PR #9185, opened 2017-11-23
[api-major] Move options from the global `PDFJS` object into the relevant viewer components and API methods

Snuffleupagus:
This should address the remaining TODO in issue #6779 (and #4919, given the renaming and that it's no longer a global parameter).

*Note:* It seems that the commits appear in a somewhat random order here in the PR; please refer to https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:rm-PDFJS-options for a properly sorted list.


Snuffleupagus:
@yurydelendik Ping; any change that you can provide feedback/review this soon?
(Given the size and scope of the changes, this PR bit-rots quickly and requires a lot of frequent updates.)

Snuffleupagus:
To try and expedite the review/landing of this PR (since I need to get this off my plate) and to get us closer to releasing PDF.js version `2.0`, I'll see if I can perhaps split up this patch series into multiple PRs to hopefully make reviewing this a bit less daunting.

Given how invasive the necessary changes are, the final PR will probably still be quite large, but I'll try to begin by first isolating the viewer component options re-factoring in one PR.
Most likely this will result in more churn in the code, and a less than desirable intermediate state until everything is done, but that's probably OK to make some sort of progress here.

PR #9187, opened 2017-11-24
Update eslint-plugin-mozilla, and enable mozilla/use-services rule.

Standard8:
I'm slowly rolling out the [mozilla/use-services](https://firefox-source-docs.mozilla.org/tools/lint/linters/eslint-plugin-mozilla.html#use-services) rule across the Firefox tree.

As part of this, I found a couple of cases in the pdfjs add-on that could be updated.

For good measure, I've also updated eslint-plugin-mozilla (which now requires eslint-plugin-no-unsanitized).


extensions/firefox/.eslintrc
```diff
@@ -1,8 +1,4 @@
 {
-  "extends": [
-    ../../.eslintrc
```
Standard8: Note: ESLint inherits automatically from eslint config files higher up the tree, so this extends isn't necessary.

Snuffleupagus (approved):

PR #9190, opened 2017-11-26
Remove the `pageRotation` getter in `PDFViewerApplication`

Snuffleupagus:
This was added, during the refactoring in PR #8556, to avoid outright breaking third-party users of the default viewer.
With PDF.js version `2.0`, where we're making API changes that aren't backwards compatible, we ought to be able to remove this piece of viewer code as well.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #9191, opened 2017-11-26
Button widget annotations: implement support for pushbuttons

### LINKED ISSUE #4872: Link in PDF is not clickable
At the top of http://www.cs.utexas.edu/users/EWD/ewd02xx/EWD288.PDF there is a link to an HTML version. Clicking it works in Apple Preview, but not in Firefox Nightly.

### LINKED ISSUE #7538: Hyperlinks not extracted for particular PDF
Hyperlinks are not getting extracted for this particular PDF. Other PDF viewers, including Chrome, Mac Preview & Acrobat, do render hyperlinks correctly.

Link to PDF file: https://cdn.relayto.com/media/files/ftV2Quw9QaWGCh7vZZbZ_BearingPoint_Annual_Report_2015.pdf

Configuration:
- Web browser and its version: Chrome 52
- Operating system and its version: Mac OS
- PDF.js version: http://mozilla.github.io/pdf.js/web/viewer.html (12.08.2016) 
- Is an extension: No

1. Upload the PDF into http://mozilla.github.io/pdf.js/web/viewer.html
2. Get 0 hyperlinks

timvandermeij:
While reading the specification, I realized that pushbuttons with action dictionaries are very similar to link annotations. Therefore, this patch implements support for pushbuttons with action dictionaries by inheriting most logic from link annotations.

Note that this patch does not handle appearance streams (which is another part of #7613) and pushbuttons that are specifically for form actions (which is also another part of #7613). However, it provides enough support to resolve the issues below (and there are *many* more PDF files at https://www.cs.utexas.edu/users/EWD/indexChron.html that are now handled correctly) and, by creating a separate class, allows us to extend this later on.

Fixes #4872
Fixes #7538
Fixes a part of #7613

*Easier reviewing with https://github.com/mozilla/pdf.js/pull/9191/files?w=1.*

### DIFF
--- a/web/annotation_layer_builder.css
+++ b/web/annotation_layer_builder.css
@@ -17,7 +17,8 @@
   position: absolute;
 }
 
-.annotationLayer .linkAnnotation > a {
+.annotationLayer .linkAnnotation > a,
+.annotationLayer .buttonWidgetAnnotation.pushButton > a {
   position: absolute;
   font-size: 1em;
   top: 0;
@@ -26,11 +27,16 @@
   height: 100%;
 }
 
-.annotationLayer .linkAnnotation > a /* -ms-a */  {
+.annotationLayer .linkAnnotation > a /* -ms-a */ {
   background: url("data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7") 0 0 repeat;
 }
 
-.annotationLayer .linkAnnotation > a:hover {
+.annotationLayer .buttonWidgetAnnotation.pushButton > a /* -ms-a */ {
+  background: url("data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7") 0 0 repeat;
+}
+
+.annotationLayer .linkAnnotation > a:hover,
+.annotationLayer .buttonWidgetAnnotation.pushButton > a:hover {
   opacity: 0.2;
   background: #ff0;
   box-shadow: 0px 2px 10px #ff0;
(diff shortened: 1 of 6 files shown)


web/annotation_layer_builder.css
```diff
@@ -26,11 +27,16 @@
   height: 100%;
 }
 
-.annotationLayer .linkAnnotation > a /* -ms-a */  {
+.annotationLayer .linkAnnotation > a /* -ms-a */ {
   background: url("data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7") 0 0 repeat;
 }
 
-.annotationLayer .linkAnnotation > a:hover {
+.annotationLayer .buttonWidgetAnnotation.pushButton > a /* -ms-a */ {
```
Snuffleupagus: I'm assuming that using the following format didn't work with the CSS preprocessor?
```CSS
.annotationLayer .linkAnnotation > a /* -ms-a */,
.annotationLayer .buttonWidgetAnnotation.pushButton > a /* -ms-a */ {
```

Anyway, I think that this is actually yet another reason to explicitly remove support for IE10 as well (in addition to IE9), since then we'd no longer need these hacks[1] :-)

---
[1] Given issue #3579.

Snuffleupagus (approved):
Looks good to me, thank you!

web/annotation_layer_builder.css
timvandermeij: Yes, I think it doesn't support that, but indeed, we should get rid of that soon anyway, hopefully.

timvandermeij:
There is an open issue for the intermittent form reference test failures.

/botio makeref

PR #9192, opened 2017-11-26
Build a fallback `ToUnicode` map for simple fonts (issue 8229)

### LINKED ISSUE #8229: Unrecognized (garbled) text when copying
I am using PDF.js to recognize text from marked areas of PDF document using slightly customized `page.getTextContent` function. This function works fine, but now I got a PDF file that is unreadable. Its text can be read in `evince` or in Chrome's internal PDF viewer (copy&paste works there), but not in PDF.js.

The font subset is embedded into PDF, it is TrueType with custom encoding:
~~~
name                                 type              encoding         emb sub uni object ID
------------------------------------ ----------------- ---------------- --- --- --- ---------
KDGXKF+TTE1A07870t00                 TrueType          Custom           yes yes yes     12  0
HXVYRX+TTE2337DB8t00                 TrueType          Custom           yes yes yes     14  0
~~~

Link to PDF file: http://files.zelenaposta.sk/pdf.js/vzor.pdf

Configuration:
- Web browser and its version: Chrome 57.0.2987.98
- Operating system and its version: Debian 8, Arch Linux, Windows 10
- PDF.js version: master (57d9a64c140283ed1e4c4e2b932c357fd1aeff32 - after 1.7.225)
- Is an extension: No

1. `gulp server`
2. Download the linked PDF into `examples` directory.
3. Open the linked PDF in Chrome: http://localhost:8888/web/viewer.html?file=%2Fexamples%2Fvzor.pdf
4. Mark and copy the first word `Rozúčtování` text into the clipboard.
5. Paste the text into any text editor.

The pasted text should be the same as shown in the viewer: `Rozúčtování`.

The pasted text is unreadable: `  `.

The result is the same in the viewer hosted on `mozilla.github.io/pdf.js`.

Snuffleupagus:
In some fonts, the included `ToUnicode` data is incomplete causing text-selection to not work properly. For simple fonts that contain encoding data, we can manually build a `ToUnicode` map to attempt to improve things.

Please note that since we're currently using the `ToUnicode` data during glyph mapping, in an attempt to avoid rendering regressions, I purposely didn't want to amend to original `ToUnicode` data for this text-selection edge-case.
Instead, I opted for the current solution, which will (hopefully) give slightly better text-extraction results in PDF file with incomplete `ToUnicode` data.

According to the PDF specification, see [section 9.10.2](http://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G8.1873172):

> A conforming reader can use these methods, in the priority given, to map a character code to a Unicode value.
> ...

Reading that paragraph literally, it doesn't seem too unreasonable to use *different* methods for different charcodes.

Fixes #8229.


brendandahl:
Seemed to be a bunch of zombie chrome's on linux. I killed them, so let's try again....

/botio-linux test

brendandahl:
Looks good. Out of curiosity, have you tried using the fallback map in adjustMapping?

One nit: I see some var->let conversions on code that wasn't really modified. In the future let's put those in another commit.

/botio makeref

brendandahl (approved):

Snuffleupagus:
> Looks good. Out of curiosity, have you tried using the fallback map in adjustMapping?

No, didn't try that. Since this was a text-selection only issue, I didn't want to risk introducing glyph mapping and thus rendering regressions elsewhere.

> One nit: I see some var->let conversions on code that wasn't really modified. In the future let's put those in another commit.

Fair point, I probably was a little overzealous with the conversion here.

Thanks for the review!

PR #9197, opened 2017-11-27
Fix for Issue 8726 (Safari's usage of searchParams)

ghost:
This commit addresses Issue #8726 

Two fixes were required to satisfy Safari compatibility for searchParams().

1) [The usage of replace() in the JURL's constructor resulted in undefined being returned from the url parameter that Safari passes.](https://github.com/mozilla/pdf.js/blob/cc47ef56ec434d6352279d58eddbcab0dec12b9f/src/shared/compatibility.js#L873) This is because Safari passes a non-String object to the JURL constructor when performing _new URL(document.location).searchParams.get("x")_ - which lacks the replace() method. Thus when doing replace() on it, url would become undefined and result in redirecting the browser to the URL with the supplied regular expression (i.e http://localhost:8888/^[%20/t/r/n/f]+|[%20/t/r/n/f]+$/g). 

The fix applied for this was to make url a String object if running on Safari browsers. This allows replace() to succeed as the method exists for the String object.

2) [When the native URL object is replaced with the JURL object on the global scope, any future URL instances will result in a JURL instance](https://github.com/mozilla/pdf.js/blob/cc47ef56ec434d6352279d58eddbcab0dec12b9f/src/shared/compatibility.js#L1033). This backfires on Safari as it does not seem to fallback to the native URL object when it is unable to find a method within JURL (i.e searchParams). Other modern browsers (i.e FireFox, Chrome) seem to correctly fallback to the native URL object when attempting to look for a method not present on the JURL object.

The fix applied for this was to make a get searchParams function for the JURL object. It creates a URLSearchParams object with the appropriate search parameters and returns it to the caller.

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -867,6 +867,11 @@ PDFJS.compatibilityChecked = true;
       base = new JURL(String(base));
     }
 
+    // Issue 8726: Safari needs url to be a String object.
+    // Otherwise, replace() below will produce a blank string.
+    if(isSafari) {
+      url = String(url);
+    }
     this._url = url;
     clear.call(this);
 
@@ -1014,6 +1019,11 @@ PDFJS.compatibilityChecked = true;
         return '';
       }
       return this._scheme + '://' + host;
+    },
+    
+    // Issue 8726: Safari does not fallback to the original URL object upon undefined methods.
+    get searchParams() {
+      return new URLSearchParams(this.search);
     },
   };
 
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -869,7 +869,7 @@ PDFJS.compatibilityChecked = true;
 
     // Issue 8726: Safari needs url to be a String object.
     // Otherwise, replace() below will produce a blank string.
-    if(isSafari) {
+    if (isSafari) {
       url = String(url);
     }
     this._url = url;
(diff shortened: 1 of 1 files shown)


src/shared/compatibility.js
```diff
@@ -867,6 +867,11 @@ PDFJS.compatibilityChecked = true;
       base = new JURL(String(base));
     }
 
+    // Issue 8726: Safari needs url to be a String object.
+    // Otherwise, replace() below will produce a blank string.
+    if (isSafari) {
+      url = String(url);
```
Snuffleupagus: For reference: This part looks fairly similar to PR #8727, which was closed based on https://github.com/mozilla/pdf.js/pull/8727#issuecomment-328908500.
@yurydelendik What were the problems?

---
Also, please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

src/shared/compatibility.js
```diff
@@ -1015,6 +1020,11 @@ PDFJS.compatibilityChecked = true;
       }
       return this._scheme + '://' + host;
     },
+
+    // Issue 8726: Safari does not fallback to the original URL object.
+    get searchParams() {
+      return new URLSearchParams(this.search);
```
Snuffleupagus: Also, given https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams#Browser_compatibility, I'm not sure how this helps in general considering that this won't work in other browsers that need to `URL` polyfill (e.g. IE11).

timvandermeij:
I'm going to close this pull request. The comment above indicate why this won't work in general. The polyfill must keep working for IE.

PR #9201, opened 2017-11-28
Add `btoa` back to domstubs.js

ydfzgyj:
`btoa` polyfill was removed in compatibility.js, but in node.js environment we still need it.


brendandahl:
It's too bad we don't have any tests exercising the node svg code. That'd be great if you added some.

PR #9207, opened 2017-11-28
Switch to using mozilla/recommended configuration for extensions/firefox.

Standard8:
This switches to using the mozilla/recommended configuration for extensions/firefox. The configuration is supplied via eslint-plugin-mozilla, already included via npm in this repository.

The benefits are that this is closer to the Firefox ESLint configuration (which is better when exporting to mozilla-central), and as eslint-plugin-mozilla is updated, you'll get any new rules more-or-less automatically.

I've de-duplicated the configuration file, the only thing I didn't remove was the parser ecmaVersion - currently it is set to 6, for mozilla-central it is currently 8. I wasn't sure if this could/should be changed or not.

### DIFF
--- a/extensions/firefox/.eslintrc
+++ b/extensions/firefox/.eslintrc
@@ -1,47 +1,42 @@
 {
+  // Note: The root .eslintrc file will define the base rules,
+  // but mozilla/recommended will override them for the rules it sets. Finally,
+  // the rules in this file will take precedence.
+  "extends": [
+    "plugin:mozilla/recommended",
+  ],
+
   "parserOptions": {
+    // Note: Remove the ecmaVersion line when Gecko 52 is no longer supported
+    // to pick up the eslint-plugin-mozilla version (or at least update to
+    // ecmaVersion 8).
     "ecmaVersion": 6,
-    "sourceType": "script"
+    "ecmaFeatures": {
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
> I've de-duplicated the configuration file, the only thing I didn't remove was the parser ecmaVersion - currently it is set to 6, for mozilla-central it is currently 8. I wasn't sure if this could/should be changed or not.

As mentioned in https://github.com/mozilla/pdf.js#firefox-and-seamonkey, we have (historically) tried to keep the Firefox addon compatible with the current ESR release (i.e. 52 as of this writing).
So provided that bumping the `ecmaVersion` doesn't enable support for features not in Firefox ESR, then I personally don't see an issue with it.

timvandermeij:
I'm also fine with this. By the way, did you remove `getBoolPref` and `log` because they are unused?

Standard8:
I took a look back, and it looks like we didn't start using ecmaVersion generally until sometime after 52. For now, I've updated the commit to add a reference to update the ecmaVersion when support for 52 is dropped.

I've also turned off experimentalObjectRestSpread as that was only implemented in FF 55.

`getBoolPref` and `log` were indeed unused in that file. I think there's versions of the same functions in other files if I remember correctly.

Snuffleupagus (approved):

PR #9213, opened 2017-11-29
Update index.md

SehyunPark:
adding comma since making it better to understand while reading the index file 
(dividing a sentence into the parts)


Snuffleupagus:
That change would actually make the sentence almost impossible to understand, since "transformation matrix" is a concept originating in the PDF specification itself.

SehyunPark:
I see! I understand what you're saying. Thank you. 
Then is it understandable for writing 
"When the viewport is created, an initial transportation matrix will also be created ..."?

PR #9219, opened 2017-11-30
correctly recognize fonts with a `-bd` suffix in their name as bold

cemerick:
Example of a such a font name in the wild is `ABZSMO+HelveticaNeueLTStd-Bd`.

I am unfortunately unable to share a document that contains that font, but it's readily googleable as an existing bolded font variant.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -1216,7 +1216,8 @@ var Font = (function FontClosure() {
         !!(nonStdFontMap[fontName] && stdFontMap[nonStdFontMap[fontName]]);
       fontName = stdFontMap[fontName] || nonStdFontMap[fontName] || fontName;
 
-      this.bold = (fontName.search(/bold/gi) !== -1);
+      this.bold = ((fontName.search(/bold/gi) !== -1) ||
+                   (fontName.search(/-bd/gi) !== -1));
       this.italic = ((fontName.search(/oblique/gi) !== -1) ||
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
> I am unfortunately unable to share a document that contains that font, but it's readily googleable as an existing bolded font variant.

Especially for font issues, given how easy it's to regress something while attempting to fix other bugs, it's imperative that we add tests.
Can you please create a reduced test-case instead (which is what we prefer for font issues anyway)?

The following tools should be helpful:
 - http://brendandahl.github.io/pdf.js.utils/browser/
This tool allows you to browse the structure of PDF files, and will this allow you to find the Ref of the font in question (and also a suitable sequence of text rendering operators in the file).
 - http://brendandahl.github.io/pdf.js.utils/maker/
This tool was designed specifically as a simple way to create reduced test-cases for font issues. Here you just select the original PDF file, input font Ref (as found above), and finally add the text rendering operators you want (again, these can be found using the tool above).

timvandermeij:
@cemerick Would it be possible for you to use the tools above to create a reduced PDF file for this patch? The patch itself looks good, but as mentioned since font handling is complex, we do need to have a test case so it can't regress.

timvandermeij:
Closing since no more information is provided. Without a test case it's not possible to verify the fix. This PR can be reopened if a test case is provided.

PR #9227, opened 2017-12-02
Only resize the sidebar with the left mouse button

Snuffleupagus:
*A small follow-up to PR #9026.*


timvandermeij (approved):

timvandermeij:
Good catch!

PR #9228, opened 2017-12-02
Implement unit tests for the utility functions `bytesToString` and `stringToBytes`

timvandermeij:


test/unit/util_spec.js
```diff
@@ -14,10 +14,37 @@
...
 
 describe('util', function() {
+  describe('bytesToString', function() {
+    it('handles non-array arguments', function() {
+      expect(function() {
+        bytesToString(null);
+      }).toThrow(new Error('Invalid argument for bytesToString'));
+    });
+
+    it('handles array arguments', function() {
```
Snuffleupagus: Since there's two code-paths in `bytesToString`, if it's not too much trouble could you please add an additional test here covering the length `> 8192` case?

Snuffleupagus (approved):
Looks good, with one minor question; thanks for adding more unit-tests :-)

test/unit/util_spec.js
timvandermeij: Done in the new commit. Good catch!

PR #9229, opened 2017-12-03
Implement unit tests for utility functions that perform type/value checks

timvandermeij:
The objective is to improve unit test coverage of shared code.

### DIFF
--- a/test/unit/util_spec.js
+++ b/test/unit/util_spec.js
@@ -14,8 +14,8 @@
  */
 
 import {
-  bytesToString, ReadableStream, removeNullCharacters, stringToBytes,
-  stringToPDFString
+  bytesToString, isArrayBuffer, isBool, isEmptyObj, isNum, isSpace, isString,
+  ReadableStream, removeNullCharacters, stringToBytes, stringToPDFString
 } from '../../src/shared/util';
 
 describe('util', function() {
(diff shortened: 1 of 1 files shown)


test/unit/util_spec.js
```diff
@@ -50,6 +50,90 @@ describe('util', function() {
...
+      expect(isEmptyObj({})).toEqual(true);
+    });
+
+    it('handles non-empty objects', function() {
+      expect(isEmptyObj({ foo: 'bar', })).toEqual(false);
+    });
+  });
+
+  describe('isNum', function() {
+    it('handles numeric values', function() {
```
Snuffleupagus: Please add an additional test-case with a floating point value, to test that `isNum` works in that case as well (to ensure that `isNum` is different from the native `Number.isInteger` function).

test/unit/util_spec.js
```diff
@@ -50,6 +50,90 @@ describe('util', function() {
     });
   });
 
+  describe('isArrayBuffer', function() {
+    it('handles array buffer values', function() {
```
Snuffleupagus: Since `isArrayBuffer` also works with typed arrays, let's add a test-case using e.g. `Uint8Array` here as well.

test/unit/util_spec.js
```diff
@@ -50,6 +50,90 @@ describe('util', function() {
...
+      expect(isArrayBuffer(new ArrayBuffer(0))).toEqual(true);
```
Snuffleupagus: Since `isArrayBuffer` also works with typed arrays, let's add a test-case using e.g. `Uint8Array` here as well.

test/unit/util_spec.js
```diff
@@ -50,6 +50,90 @@ describe('util', function() {
...
+    });
+
+    it('handles non-empty objects', function() {
+      expect(isEmptyObj({ foo: 'bar', })).toEqual(false);
+    });
+  });
+
+  describe('isNum', function() {
+    it('handles numeric values', function() {
+      expect(isNum(1)).toEqual(true);
```
Snuffleupagus: Please add an additional test-case using a floating point value here, to test that `isNum` works in that case as well (to ensure that `isNum` is different from the native `Number.isInteger` function).

test/unit/util_spec.js
Snuffleupagus: Since `isArrayBuffer` also works with typed arrays, let's add a test-case using e.g. `Uint8Array` here as well.

test/unit/util_spec.js
Snuffleupagus: Please add an additional test-case using a floating point value here, to test that `isNum` works in that case as well (to ensure that `isNum` is different from the native `Number.isInteger` function).

Snuffleupagus (approved):
Looks good, with two additional test-cases added; thank you!

test/unit/util_spec.js
timvandermeij: Done in the new commit.

test/unit/util_spec.js
timvandermeij: Done in the new commit. Good points; thanks!

PR #9230, opened 2017-12-03
Add basic support for non-embedded Calibri fonts (issue 9195)

### LINKED ISSUE #9195: All characters from attached file are displayed incorrectly insie Firefox browser and plugin
Attached PDF file here:
[testfile.pdf](https://github.com/mozilla/pdf.js/files/1506940/testfile.pdf)

Configuration:
- Web browser and its version: Firefox 57.0 PL (64-bit)
- Operating system and its version: Windows 10 Fall Creators Update
- PDF.js version: 1.9.583
- Is a browser extension: No

1. Simply open attached pdf file in browser. 
2. All is displayed wrongly

The expected behaviour can be observer in any other browser, chrome or edge or standalone PDF viewer.
Screenshot:
screen good

What went wrong?
On Firefox it looks like this:
screen bad

Snuffleupagus:
There's a number of issues with the fonts in the referenced PDF file. First of all, they contain broken `ToUnicode` data (`NUL` bytes all over the place). However even if you skip those, the `ToUnicode` data appears to contain nothing but a `IdentityH` CMap which won't help provide a proper glyph mapping.

The real issue actually turns out to be that the PDF file uses the "Calibri" font[1], but doesn't include any font files. Since that one isn't a standard font, and uses a fairly different CID to GID map compared to the standard fonts, we're not able to render the file even remotely correct.
To work around this, I'm thus proposing that we include a (incomplete) glyph map for Calibri, and fallback to the standard Helvetica font. Obviously this isn't going to look perfect, but it's really the best that we can hope to achieve given that the PDF file is missing the necessary font data.

Finally, please note that none of the PDF readers I've tried (Adobe Reader, PDFium in Chrome) were able to extract the text (which isn't very surprising, given the broken `ToUnicode` data).

Fixes #9195.

---

[1] According to Wikipedia, see https://en.wikipedia.org/wiki/Calibri, Calibri is (primarily) a Windows font.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -28,7 +28,7 @@ import {
 } from './encodings';
 import {
   getGlyphMapForStandardFonts, getNonStdFontMap, getStdFontMap,
-  getSupplementalGlyphMapForArialBlack
+  getSupplementalGlyphMapForArialBlack, getSupplementalGlyphMapForCalibri
 } from './standard_fonts';
 import {
   getUnicodeForGlyph, getUnicodeRangeFor, mapSpecialUnicodeValues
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
Let's see if the Linux bot wants to cooperate this time...

/botio-linux test

brendandahl (approved):

PR #9244, opened 2017-12-05
[Firefox addon] Temporarily disable the ESLint `semi` rule, around the `DEFAULT_PREFERENCES` declaration, to prevent mozilla-central test errors

Snuffleupagus:
While this doesn't actually fix the underlying issue, it should prevent the ESLint errors and thus make future PDF.js updates easier.
Compared to updating (and testing) the preprocessor, this seems like a reasonable workaround given its simplicity.

/cc @rvandermeulen

### DIFF
--- a/extensions/firefox/bootstrap.js
+++ b/extensions/firefox/bootstrap.js
@@ -45,11 +45,13 @@ function log(str) {
 }
 
 function initializeDefaultPreferences() {
+  /* eslint-disable semi */
   var DEFAULT_PREFERENCES =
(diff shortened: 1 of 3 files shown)


timvandermeij:
Could you perhaps give a bit more context for this patch, i.e., what is failing now, why is it happening just now and why does this solve (or work-around for) it?

brendandahl:
This about https://bugzilla.mozilla.org/show_bug.cgi?id=1421048#c3

PR #9245, opened 2017-12-06
[api-major] Only create a `StatTimer` for pages when `enableStats == true` (issue 5215)

### LINKED ISSUE #5215: StatTimer created for every page
For all pages of a document page proxies are created, even when not used. 
Each of these page proxies gets a StatTimer instance in api.js#L431.

While not having a big impact on memory and CPU it seems to be a waste to do that when they are not enabled via pdfbug. However the current way allows just to call methods on it without checking.

A refactored version could perhaps use a single timer which records its times in an per page data structure which is only created when actually used.

Snuffleupagus:
Unless the debugging tools (i.e. `PDFBug`) are enabled, or the `browsertest` is running, the `PDFPageProxy.stats` aren't actually used for anything.
Rather than initializing unnecessary `StatTimer` instances, we can simply re-use *one* dummy class (with static methods) for every page. Note that by using a dummy `StatTimer` in this way, rather than letting `PDFPageProxy.stats` be undefined, we don't need to guard *every* single stats collection callsite.

Since it wouldn't make much sense to attempt to use `PDFPageProxy.stats` when stat collection is disabled, it was instead changed to a "private" property (i.e. `PDFPageProxy._stats`) and a getter was added for accessing `PDFPageProxy.stats`. This getter will now return `null` when stat collection is disabled, making that case easy to handle.

For benchmarking purposes, the test-suite used to re-create the `StatTimer` after loading/rendering each page. However, modifying properties on various API code from the outside in this way seems very error-prone, and is an anti-pattern that we really should avoid at all cost. Hence the `PDFPageProxy.cleanup` method was modified to accept an optional parameter, which will take care of resetting `this.stats` when necessary, and `test/driver.js` was updated accordingly.

Finally, a tiny bit more validation was added on the viewer side, to ensure that all the code we're attempting to access is defined when handling `PDFPageProxy` stats.

Fixes #5215.

---
*Edit:* It would probably be good a idea to add a few unit-tests to check that `PDFPageProxy.stats` behaves reasonable with/without stats collection enabled. However, given that all options are currently global, the tests will be so much easier to write once PR #9185 has landed, which is why I didn't add any tests here.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -15,15 +15,14 @@
 /* globals requirejs, __non_webpack_require__ */
 
 import {
-  assert, createPromiseCapability, getVerbosityLevel, info,
-  InvalidPDFException, isArrayBuffer, isSameOrigin, loadJpegStream,
-  MessageHandler, MissingPDFException, NativeImageDecoding, PageViewport,
-  PasswordException, StatTimer, stringToBytes, UnexpectedResponseException,
-  UnknownErrorException, Util, warn
+  assert, createPromiseCapability, getVerbosityLevel, info, InvalidPDFException,
+  isArrayBuffer, isSameOrigin, loadJpegStream, MessageHandler,
+  MissingPDFException, NativeImageDecoding, PageViewport, PasswordException,
+  stringToBytes, UnexpectedResponseException, UnknownErrorException, Util, warn
 } from '../shared/util';
 import {
   DOMCanvasFactory, DOMCMapReaderFactory, getDefaultSetting,
-  RenderingCancelledException
+  RenderingCancelledException, StatTimer
 } from './dom_utils';
 import { FontFaceObject, FontLoader } from './font_loader';
 import { CanvasGraphics } from './canvas';
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -21,7 +21,7 @@ import {
   stringToBytes, UnexpectedResponseException, UnknownErrorException, Util, warn
 } from '../shared/util';
 import {
-  DOMCanvasFactory, DOMCMapReaderFactory, getDefaultSetting,
+  DOMCanvasFactory, DOMCMapReaderFactory, DummyStatTimer, getDefaultSetting,
   RenderingCancelledException, StatTimer
 } from './dom_utils';
 import { FontFaceObject, FontLoader } from './font_loader';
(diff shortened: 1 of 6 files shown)


src/display/api.js
```diff
@@ -1046,6 +1047,9 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
       }, this);
       this.objs.clear();
       this.annotationsPromise = null;
+      if (resetStats && this._stats instanceof StatTimer) {
+        this._stats = new StatTimer();
```
timvandermeij: Wouldn't it be better to implement a `cleanup` method on the `StatTimer`  so we can avoid creating a new instance here and re-use the existing one?

src/display/dom_utils.js
```diff
@@ -461,6 +461,72 @@ function isExternalLinkTargetSet() {
...
+    for (let i = 0, ii = times.length; i < ii; ++i) {
+      let span = times[i];
+      let duration = span.end - span.start;
+      out += `${span['name'].padEnd(longest)} ${duration}ms\n`;
+    }
+    return out;
+  }
+}
+
+class DummyStatTimer {
```
timvandermeij: Could you add a small comment to this class as to why it exists? I think you described it pretty well in the pull request description, but let's add it here too for reference.

timvandermeij (commented):
Overall I'm really liking this change; it looks much better. If these comments are addressed, let's do a test run and I'll review again.

src/display/dom_utils.js
Snuffleupagus: Sure, I've tried to add a comment; please check if it makes sense!

src/display/api.js
Snuffleupagus: While I'm not sure if creating a new instance is that much of an issue in practice, since it will only occur during testing, it's nonetheless a really nice idea!

I've updated the patch with `reset` methods in `StatTimer`/`DummyStatTimer` (in the first case, it's also being re-used by the `constructor`).

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

src/display/dom_utils.js
timvandermeij: Yes, that's exactly how I thought it should look. Nice!

src/display/api.js
timvandermeij: The `reset` method is good and the re-use is also nice.

PR #9253, opened 2017-12-08
Handle PDF files with missing 'endobj' operators, by scanning through the contents of objects found in `XRef.indexObjects` (issue 9105)

### LINKED ISSUE #9105: Invalid root reference when parsing pdf
Trying to read a pdf file using PDFJS.getDocument function. The pdf worker is throwing

UnknownErrorException {
 name: "UnknownErrorException", 
 message: "Invalid root reference", 
 details: "FormatError: Invalid root reference"
}

The error can be easily reproduced calling PDFJS.getDocument(pdfUrl) using the attached pdf
[110229_10312017_6810306.pdf](https://github.com/mozilla/pdf.js/files/1444166/110229_10312017_6810306.pdf)

Snuffleupagus:
Unfortunately this will, especially for larger documents, cause the parsing of PDF files with *corrupt* XRef data to become (slightly) slower. However, that's probably unavoidable if we want to support PDF generators that appear to treat 'endobj' like an optional operator.

Fixes #9105.

*Edit:* I've looked briefly at reducing the amount of parsing needed in order to fix the issue. However, I've been unable to come up with something better myself here :-(

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -1149,6 +1149,27 @@ var XRef = (function XRefClosure() {
           var contentLength = skipUntil(buffer, position, endobjBytes) + 7;
           var content = buffer.subarray(position, position + contentLength);
 
+          // Scan through the 'obj' content, to ensure that we won't miss
+          // any 'obj' operators in corrupt files (fixes issue9105.pdf).
+          let contentPos = token.length;
+          while (contentPos < contentLength) {
+            let contentToken = readToken(content, contentPos);
+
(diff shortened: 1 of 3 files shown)


timvandermeij:
Is it possible to use the file in https://github.com/mozilla/pdf.js/issues/9105#issuecomment-342817956 as a reduced test case instead of the linked one here? The linked one is fine too, but I always prefer a reduced one if possible.

Issue #8076, opened 2017-02-16
Page in PDF not loading completely in Firefox, missing image in Chromium


pcworld:
Link to PDF file: https://www.personalausweisportal.de/SharedDocs/Downloads/DE/Flyer-und-Broschueren/eID_Broschuere.pdf?__blob=publicationFile&v=1 (sha512sum 62e4a6a96257f219d1f1fc1c644eea86b17c056486a91a159e4cc0085347b3bc3c3d1a4e06eae204e1dce92f507afa5d55c21604584ffa852efb7fb303f9f846)

Configuration:
- Web browser and its version: Chromium 53.0.2785.143-0ubuntu0.14.04.1.1145, Firefox 51.0.1+build2-0ubuntu0.14.04.2
- Operating system and its version: Ubuntu 14.04
- PDF.js version: https://mozilla.github.io/pdf.js/web/viewer.html
- Is an extension: Issues happen in the Chromium pdf.js extension, Firefox's native pdf.js and the aforementioned online version of pdf.js

Steps to reproduce the problem:
1. Open the PDF in pdf.js and scroll to the last page.

What is the expected behavior?
The last page renders like in the following screenshot of the page in Evince:
![Last page in Evince](https://cloud.githubusercontent.com/assets/1506734/23024514/f8cffa64-f45a-11e6-8329-89f921ad7ad3.png)

What went wrong?
In Firefox: Page doesn't finish loading (loading animation doesn't disappear), page is only partially loaded:
![Last page in pdf.js in Firefox](https://cloud.githubusercontent.com/assets/1506734/23024575/315df3fe-f45b-11e6-847e-8111c26da84e.png)
In Chromium: Loading animation disappears, but the QR code is blank (white):
![Last page in pdf.js in Chromium](https://cloud.githubusercontent.com/assets/1506734/23024624/630f7a30-f45b-11e6-8b39-fd5281cfbcd5.png)

The PDF appears to have been created by "Adobe InDesign CS5.5 (7.5.3)".

yurydelendik:
Hmm, that's QR code size is 16667x16667, it's like 8000dpi image.

yurydelendik:
@pcworld can you check if you can reduce size of the image to acceptable dpi --perhaps 300dpi. I'm sure other PDF readers and network providers will say thanks :)

pcworld:
@yurydelendik I am not the creator of this PDF.
[...]

PR #9255, opened 2017-12-09
[api-minor] Downsize large image masks before rendering

apoorv-mishra:
Hi,

This PR addresses #8076 . The problem was the large image mask resolution. The solution was to scale it down in `image.js` and then send the reduced image mask data for rendering via `evaluator.js`.

For downscaling, I used 'point sampling' algorithm which does something like [this](https://www.imagemagick.org/Usage/filter/#point). Since we're dealing with image mask(b&w palette), this would serve the purpose well. Compare [this](https://i.imgur.com/KrTFwR3.png)(pdfium) vs [this](https://i.imgur.com/ICGlJr2.png).

Thanks,
Apoorv

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -61,11 +61,14 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
     isEvalSupported: true,
   };
 
-  function PartialEvaluator({ xref, handler, pageIndex, idFactory, fontCache,
-                              builtInCMapCache, options = null,
-                              pdfFunctionFactory, }) {
+  function PartialEvaluator({ pdfManager, xref, handler, pageIndex, idFactory,
+                              fontCache, builtInCMapCache, options = null,
+                              pdfFunctionFactory,
+                              combinedInitialTransform = null, }) {
+    this.pdfManager = pdfManager;
     this.xref = xref;
     this.handler = handler;
+    this.combinedInitialTransform = combinedInitialTransform;
     this.pageIndex = pageIndex;
     this.idFactory = idFactory;
     this.fontCache = fontCache;
@@ -295,7 +298,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
     async buildPaintImageXObject({ resources, image, isInline = false,
                                    operatorList, cacheKey, imageCache,
-                                   forceDisableNativeImageDecoder = false, }) {
+                                   forceDisableNativeImageDecoder = false,
+                                   ctm, }) {
       var dict = image.dict;
       var w = dict.get('Width', 'W');
       var h = dict.get('Height', 'H');
@@ -326,10 +330,22 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                                       /* forceClamped = */ true);
         var decode = dict.getArray('Decode', 'D');
 
+        let resizeWidth = width, resizeHeight = height;
+        if (this.combinedInitialTransform) {
+          const combinedTransform =
+            Util.transform(this.combinedInitialTransform, ctm);
+          const scaledDimensions =
+            Util.singularValueDecompose2dScale(combinedTransform);
+          resizeWidth = scaledDimensions[0];
+          resizeHeight = scaledDimensions[1];
+        }
+
         imgData = PDFImage.createMask({
           imgArray,
           width,
           height,
+          resizeWidth,
+          resizeHeight,
           imageIsFromDecodeStream: image instanceof DecodeStream,
           inverseDecode: (!!decode && decode[0] > 0),
         });
@@ -982,6 +998,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                     operatorList,
                     cacheKey: name,
                     imageCache,
+                    ctm: stateManager.state.ctm,
                   }).then(resolveXObject, rejectXObject);
                   return;
                 } else if (type.name === 'PS') {
@@ -1038,6 +1055,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 operatorList,
                 cacheKey,
                 imageCache,
+                ctm: stateManager.state.ctm,
               }));
               return;
             case OPS.showText:
@@ -2677,6 +2695,7 @@ var TranslatedFont = (function TranslatedFontClosure() {
       type3Options.nativeImageDecoderSupport = NativeImageDecoding.NONE;
       var type3Evaluator = evaluator.clone(type3Options);
       type3Evaluator.parsingType3Font = true;
+      type3Evaluator.combinedInitialTransform = null;
 
       var translatedFont = this.font;
       var loadCharProcsPromise = Promise.resolve();
--- a/src/core/image.js
+++ b/src/core/image.js
@@ -79,6 +79,73 @@ var PDFImage = (function PDFImageClosure() {
     return dest;
   }
 
+  /**
+   * Downsizes an image mask.
+   * @param {TypedArray} src - The source buffer.
+   * @param {Number} w1 - Original width.
+   * @param {Number} h1 - Original height.
+   * @param {Number} w2 - New width.
+   * @param {Number} h2 - New height.
+   * @returns {Object} Downsized image mask buffer.
+   */
+  function downsizeImageMask(src, w1, h1, w2, h2) {
+    const MIN_WIDTH_THRESHOLD = 4096;
+    const MIN_HEIGHT_THRESHOLD = 4096;
+
+    w2 = (w1 > MIN_WIDTH_THRESHOLD) ?
+          Math.max(MIN_WIDTH_THRESHOLD, w2) : w1;
+    h2 = (h1 > MIN_HEIGHT_THRESHOLD) ?
+         Math.max(MIN_HEIGHT_THRESHOLD, h2) : h1;
+
+    const resizeFactorX = Math.ceil(w1 / w2);
+    const resizeFactorY = Math.ceil(h1 / h2);
+    if (resizeFactorX === 1 && resizeFactorY === 1) {
+      return null;
+    }
+    const packedWidth = (w1 + 7) >> 3;
+    const xHop = resizeFactorX >> 3;
+    const yHop = resizeFactorY;
+    const bitHop = resizeFactorX & 7;
+
+    const resizedWidth = Math.ceil(w1 / resizeFactorX);
+    const resizedHeight = Math.ceil(h1 / resizeFactorY);
+
+    const destWidth = (resizedWidth + 7) >> 3;
+    const destHeight = resizedHeight;
+    const dest = new Uint8ClampedArray(destWidth * destHeight);
+
+    let destIndex = 0;
+    for (let i = 0; i < h1; i += yHop) {
+      let val = 0, destMask = 128;
+      let srcIndex = i * packedWidth;
+      let srcMask = 0x8080; // hi and low bytes are the same
+      for (let j = 0; j < resizedWidth; j++) {
+        val |= (src[srcIndex] & srcMask) ? destMask : 0;
+        destMask >>= 1;
+        if (!destMask) {
+          dest[destIndex++] = val;
+          destMask = 128;
+          val = 0;
+        }
+        srcMask >>= bitHop;
+        // Mask was shifted to the next byte?
+        if (srcMask < 0x100) {
+          srcIndex++;
+          // Duplicate low mask to the high byte
+          srcMask |= srcMask << 8;
+        }
+        srcIndex += xHop;
+      }
+      // Add padding
+      if (destMask < 128) {
+        val |= destMask | (destMask - 1);
+        dest[destIndex++] = val;
+      }
+    }
+
+    return { dest, resizedWidth, resizedHeight, };
+  }
+
   function PDFImage({ xref, res, image, isInline = false, smask = null,
                       mask = null, isMask = false, pdfFunctionFactory, }) {
     this.image = image;
@@ -264,12 +331,14 @@ var PDFImage = (function PDFImageClosure() {
   };
 
   PDFImage.createMask = function({ imgArray, width, height,
-                                   imageIsFromDecodeStream, inverseDecode, }) {
+                                   imageIsFromDecodeStream, inverseDecode,
+                                   resizeWidth, resizeHeight, }) {
     if (typeof PDFJSDev === 'undefined' ||
         PDFJSDev.test('!PRODUCTION || TESTING')) {
       assert(imgArray instanceof Uint8ClampedArray,
              'PDFImage.createMask: Unsupported "imgArray" type.');
     }
+
     // |imgArray| might not contain full data for every pixel of the mask, so
     // we need to distinguish between |computedLength| and |actualLength|.
     // In particular, if inverseDecode is true, then the array we return must
@@ -302,6 +371,18 @@ var PDFImage = (function PDFImageClosure() {
     if (inverseDecode) {
       for (i = 0; i < actualLength; i++) {
         data[i] ^= 0xFF;
+      }
+    }
+
+    if (resizeWidth < width || resizeHeight < height) {
+      const resizedImageObj = downsizeImageMask(data, width, height,
+                                              resizeWidth, resizeHeight);
+      if (resizedImageObj) {
+        return {
+          data: resizedImageObj.dest,
+          width: resizedImageObj.resizedWidth,
+          height: resizedImageObj.resizedHeight,
+        };
       }
     }
 
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -20,7 +20,7 @@ import {
   InvalidPDFException, isArrayBuffer, isSameOrigin, MissingPDFException,
   NativeImageDecoding, PasswordException, setVerbosityLevel, shadow,
   stringToBytes, UnexpectedResponseException, UnknownErrorException,
-  unreachable, URL, warn
+  unreachable, URL, Util, warn
 } from '../shared/util';
 import {
   deprecated, DOMCanvasFactory, DOMCMapReaderFactory, DummyStatTimer,
@@ -891,6 +891,9 @@ class PDFDocumentProxy {
  *                    CSS <color> value, a CanvasGradient object (a linear or
  *                    radial gradient) or a CanvasPattern object (a repetitive
  *                    image). The default value is 'rgb(255,255,255)'.
+ * @property {boolean} resizeLargeImages - (optional) Whether or not large
+ *                     images will be resized before rendering. The default
+ *                     value is true.
  */
 
 /**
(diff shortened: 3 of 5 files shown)


src/display/api.js
```diff
@@ -820,12 +820,22 @@ var PDFPageProxy = (function PDFPageProxyClosure() {
...
       var renderingIntent = (params.intent === 'print' ? 'print' : 'display');
       var canvasFactory = params.canvasFactory || new DOMCanvasFactory();
+
       let webGLContext = new WebGLContext({
         // TODO: When moving this parameter from `PDFJS` to {RenderParameters},
         //       change its name to `enableWebGL` instead.
         enable: !getDefaultSetting('disableWebGL'),
       });
 
+      var combinedInitialTransform;
```
Snuffleupagus: While it might be OK to have this new resizing behaviour be the default, it must be possible for an API consumer to opt-out of it (since it's a new feature).
Hence I'd suggest adding a new render parameter (with accompanying JSDoc), for example named  `resizeLargeImages`, which defaults to `true` but can be used to disable resizing.

So something like the following:
```javascript
let combinedInitialTransform = null;
if (params.resizeLargeImages !== false) {
  if (params.transform) {
    combinedInitialTransform = Util.transform(params.transform,
                                              params.viewport.transform);
  } else {
    combinedInitialTransform = params.viewport.transform;
  }
}
```

src/core/image.js
```diff
@@ -75,6 +75,74 @@ var PDFImage = (function PDFImageClosure() {
...
+      bitHop = 1;
+    }
+    var destWidth = (Math.ceil(packedWidth / xHop) * (8 / bitHop) + 7) >> 3;
+    var destHeight = Math.ceil(h1 / yHop);
+    var dest = new Uint8Array(destWidth * destHeight);
+    var resizedWidth = destWidth << 3;
+    var resizedHeight = destHeight;
+    if (resizeFactorX === 1 && resizeFactorY === 1) {
+      dest = src;
+      return { dest, resizedWidth, resizedHeight, };
```
Snuffleupagus: If you're passing the `src` through unchanged, I'd assume that you need to also return the *original* width/height as well to prevent rendering errors!?

src/core/image.js
```diff
@@ -75,6 +75,74 @@ var PDFImage = (function PDFImageClosure() {
...
+      bitHop = 2;
+    } else {
+      bitHop = 1;
+    }
+    var destWidth = (Math.ceil(packedWidth / xHop) * (8 / bitHop) + 7) >> 3;
+    var destHeight = Math.ceil(h1 / yHop);
+    var dest = new Uint8Array(destWidth * destHeight);
+    var resizedWidth = destWidth << 3;
+    var resizedHeight = destHeight;
+    if (resizeFactorX === 1 && resizeFactorY === 1) {
```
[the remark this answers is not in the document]
Snuffleupagus: Another option could perhaps be to simply return `null` in this case, e.g.
```javascript
function downsizeImageMask(src, w1, h1, w2, h2) {
  let resizeFactorX = Math.round(w1 / w2);
  let resizeFactorY = Math.round(h1 / h2);
  if (resizeFactorX === 1 && resizeFactorY === 1) {
     return null;
  }
  
  // The rest of the code here...
}
```
And then change the [call-site](https://github.com/mozilla/pdf.js/pull/9255/files#diff-7105ab0e122d5ff915e2f8e96347d136R344) to do something like this instead:
```javascript
  let resizedImageObj = downsizeImageMask(data, width, height,
                                          resizeWidth, resizeHeight);
  if (resizedImageObj) {
    return {
      data: resizedImageObj.dest,
      width: resizedImageObj.resizedWidth,
      height: resizedImageObj.resizedHeight,
    };
  }
}
```

Snuffleupagus (commented):
Currently, this patch only seem to resize large image *masks*, but *normal* images can and often are much larger than necessary too. Don't we need to resize regular images as well?

I've not looked in detail at the resizing algorithm implemented in the patch, however I've left a few inline comments regarding the general structure.
Please also keep the following in mind: When adding new code, please use `let` rather than `var`; and please avoid unnecessarily changing the blame on lines where it's not necessary.

apoorv-mishra:
Hi @Snuffleupagus , you are indeed right. We need to resize regular images too. But my main motive here was to fix the issue at hand. Also, navigating through the codebase, I clearly saw sort of 'separation of concerns' in handling regular images and image masks(one example being, `createImageMask` and `createImageData`). So, this patch cleanly fixes the issue without breaking other things. Also, there would not be a problem later to handle regular images(due to this separation of concerns). Though, in that case, we would be needing more sophisticated algorithms, something like(I guess) interpolation coupled with box sampling for which I would be needing some time to study. I would be highly obliged to extend this to handle regular images, if I get the time to do so. I would be pushing the changes suggested by you shortly.

Snuffleupagus:
> We need to resize regular images too. But my main motive here was to fix the issue at hand.

Thanks, I just wanted to ask since this was something I noticed while looking through the code.
Unless @yurydelendik thinks otherwise, that can probably be implemented in a follow-up instead.

src/core/image.js
```diff
@@ -75,6 +75,73 @@ var PDFImage = (function PDFImageClosure() {
...
+          mask >>= 1;
+          elemMask >>= bitHop;
+        }
+        if (!mask) {
+          dest[destIndex++] = val;
+        }
+      }
+      // Add padding
+      if (mask) {
+          while (mask) {
```
Snuffleupagus: Nit: Please remove two spaces of indentation on this line.

src/display/api.js
```diff
@@ -713,6 +713,9 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
  *                    CSS <color> value, a CanvasGradient object (a linear or
  *                    radial gradient) or a CanvasPattern object (a repetitive
  *                    image). The default value is 'rgb(255,255,255)'.
+ * @property {boolean} resizeLargeImages - (optional) Whether or not large
+ *                     images need to be resized before rendering. The default
```
Snuffleupagus: Replace `need to` with `will` here, since that seems more accurate.

src/core/image.js
```diff
@@ -75,6 +75,73 @@ var PDFImage = (function PDFImageClosure() {
...
+    let destIndex = 0;
+
+    for (let i = 0; i < h1; i += yHop) {
+      let val, mask;
+      for (let j = 0; j < packedWidth; j += xHop) {
+        if (!mask) {
+          mask = 128;
+          val = 0;
+        }
+        let elemMask = 128;
```
yurydelendik: Rename `elemMask` to `srcMask`

src/core/image.js
```diff
@@ -75,6 +75,73 @@ var PDFImage = (function PDFImageClosure() {
...
+        let elem = src[i * packedWidth + j];
+        while (elemMask) {
+          val += (elemMask & elem) ? mask : 0;
+          mask >>= 1;
+          elemMask >>= bitHop;
+        }
+        if (!mask) {
+          dest[destIndex++] = val;
```
yurydelendik: Move this to the beginning of the loop if (at line 118).

src/core/image.js
```diff
@@ -75,6 +75,73 @@ var PDFImage = (function PDFImageClosure() {
...
+        }
+      }
+      // Add padding
+      if (mask) {
+        while (mask) {
```
yurydelendik: replace the loop with `val += mask | (mask - 1)`;

src/core/evaluator.js
```diff
@@ -382,10 +385,20 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
         var imgArray = image.getBytes(bitStrideLength * height);
         var decode = dict.getArray('Decode', 'D');
 
+        let resizeWidth = null, resizeHeight = null;
```
yurydelendik: default `resizeWidth` to `width` and `resizeHeight` to `height`

src/core/image.js
```diff
@@ -271,6 +339,19 @@ var PDFImage = (function PDFImageClosure() {
       }
     }
 
+    if ((resizeWidth > 0 && resizeWidth < width) &&
```
yurydelendik: we will require the resizeWidth and resizeHeight to have a value, see comment above. let's use `resizeWidth < width || resizeHeight < height`to decide if we need to downsize.

yurydelendik (approved):
Looks good with changes below. I did not analyze the performance or quality of the algorithm -- we can improve that later.

brendandahl:
@apoorv-mishra looks like this PR is close, can you address the above comments?

apoorv-mishra:
@brendandahl So sorry, due to other commitments, I haven't been able to look at this. You're correct, this is close. I'll take a closer look at this during the coming weekend.

redlum94:
@apoorv-mishra Do you already have any news on this PR?

apoorv-mishra:
Hi @brendandahl and @yurydelendik , apologies for keeping this PR pending for so long. I have addressed the above comments pointed out by @yurydelendik and resubmitted the patch. Please have a look.

src/core/image.js
```diff
@@ -79,6 +79,68 @@ var PDFImage = (function PDFImageClosure() {
...
+    } else if (resizeFactorX >= 4) {
+      bitHop = 4;
+    } else if (resizeFactorX >= 2) {
+      bitHop = 2;
+    } else {
+      bitHop = 1;
+    }
+    let destWidth = (Math.ceil(packedWidth / xHop) * (8 / bitHop) + 7) >> 3;
+    let destHeight = Math.ceil(h1 / yHop);
+    let dest = new Uint8Array(destWidth * destHeight);
```
Snuffleupagus: This must be `Uint8ClampedArray` instead, since all image handling code has been refactored to return data in that format. (If you rebase the patch against `master`, you'll most likely get even more failures unless the typed array is fixed here.)

src/core/image.js
```diff
@@ -79,6 +79,68 @@ var PDFImage = (function PDFImageClosure() {
...
+   */
+  function downsizeImageMask(src, w1, h1, w2, h2) {
+    let resizeFactorX = Math.round(w1 / w2);
+    let resizeFactorY = Math.round(h1 / h2);
+    if (resizeFactorX === 1 && resizeFactorY === 1) {
+      return null;
+    }
+    let packedWidth = (w1 + 7) >> 3;
+    let xHop = (resizeFactorX >> 3);
+    xHop = xHop ? xHop : 1;
```
Snuffleupagus: Can't you simply fold the two lines above together as `let xHop = (resizeFactorX >> 3) || 1;`, rather than first initializing it and then using a ternary condition?

src/core/image.js
```diff
@@ -79,6 +79,68 @@ var PDFImage = (function PDFImageClosure() {
...
+    let yHop = resizeFactorY;
+    let bitHop;
```
Snuffleupagus: Nit: Could initialize this to `1` and remove the `else` case below.

Snuffleupagus (commented):
Looking at the test-results this seems to, among other things, break Type3 fonts quite badly. One possible, somewhat hacky, solution for the font related regressions might be to add `type3Evaluator.combinedInitialTransform = null;` just after line https://github.com/mozilla/pdf.js/blob/master/src/core/evaluator.js#L2654 to fix that.

apoorv-mishra:
@Snuffleupagus , Are you sure that's the font issue? Those are image masks as far as I can tell by debugging. Can you be specific about the pdf file? I debugged with the file named fit11-talk.pdf and it turned out the problem was with image masks. I didn't look at the other files though. Anyways, I'll try making the change suggested by you above and see what happens.

yurydelendik:
- [ ] we need to sort out caching before landing it -- change in combined transforms invalidates cached image

timvandermeij:
@apoorv-mishra Do you perhaps have time to finish this? I see that @yurydelendik posted some final remarks above and then we should be able to merge this.

src/core/evaluator.js
```diff
@@ -378,10 +381,22 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                                       /* forceClamped = */ true);
         var decode = dict.getArray('Decode', 'D');
 
+        let resizeWidth = width, resizeHeight = height;
+        if (this.combinedInitialTransform) {
+          let combinedTransform =
```
timvandermeij: This can be `const` instead of `let`.

src/core/evaluator.js
```diff
@@ -378,10 +381,22 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+            Util.transform(this.combinedInitialTransform, ctm);
+          let scaledDimensions =
```
timvandermeij: This can be `const` instead of `let`.

src/core/image.js
```diff
@@ -260,12 +319,14 @@ var PDFImage = (function PDFImageClosure() {
   };
 
   PDFImage.createMask = function({ imgArray, width, height,
-                                   imageIsFromDecodeStream, inverseDecode, }) {
+                                   imageIsFromDecodeStream, inverseDecode,
+                                   resizeHeight, resizeWidth, }) {
```
timvandermeij: It looks a bit better to reverse width and height here, just like in the rest of the patch, i.e., first width, then height.

src/core/image.js
```diff
@@ -301,6 +362,18 @@ var PDFImage = (function PDFImageClosure() {
       }
     }
 
+    if (resizeWidth < width || resizeHeight < height) {
+      let resizedImageObj = downsizeImageMask(data, width, height,
```
timvandermeij: This can be `const` instead of `let`.

src/core/evaluator.js
apoorv-mishra: Done.

src/core/evaluator.js
apoorv-mishra: Done.

src/core/image.js
apoorv-mishra: Done.

src/core/image.js
apoorv-mishra: Done.

Snuffleupagus:
>     * [ ]  we need to sort out caching before landing it -- change in combined transforms invalidates cached image

This part of the puzzle probably isn't very "beginner friendly", given all of the cases to consider and that caching occurs on both worker *and* main threads. I'll attempt, time permitting of course, to look into this during the weekend.

apoorv-mishra:
We also need to decide on the "min" threshold value of `width` and `height` for which to downsize the image mask, like don't downsize `if (width < 1000 && height < 1000)`. Some of the tests are failing because we haven't decided on this `MIN_THRESHOLD` check.

apoorv-mishra:
Any progess on this @Snuffleupagus ?

Snuffleupagus:
> Any progess on this

Yes, I've unfortunately been side-tracked by other issues, but now I'm slowly getting there :-)

Fixing the  `imageCache` wasn't as straightforward as I'd hoped and my initial attempt ended up requiring way too much re-factoring. I'm now trying a slightly less invasive approach; hopefully I'll have a patch ready later in the week.

timvandermeij:
The reference tests show that many image masks are downsized too much and therefore become blurry. This should be fixed first before figuring out how to do caching, because that's of no use if the functionality is not complete yet. Could you check why the reference images are regressing? You can generate the reference images locally for testing; see https://github.com/mozilla/pdf.js/wiki/Contributing.

apoorv-mishra:
> We also need to decide on the "min" threshold value of `width` and `height` for which to downsize the image mask, like don't downsize `if (width < 1000 && height < 1000)`. Some of the tests are failing because we haven't decided on this `MIN_THRESHOLD` check.

@timvandermeij This is what I suggest. What do you think?

timvandermeij:
I think a resize threshold of 1000 should be fine. If it's a constant we can always tweak it. I would say that a value that makes most reference tests look good should be OK, but note that I'm not too familiar with the image decoding/resizing code to say for sure.

apoorv-mishra:
Hello @timvandermeij , can we run ref tests on these changes? I can't run locally, browser abruptly stops when I try to run `gulp makeref`(checked out on latest master). I keep getting something like `Exiting due to channel error`

Snuffleupagus:
> can we run ref tests on these changes?

That's really of no use at this point, since the patch is still trivially broken.
The added dimension checks will only prevent issues with square images, but does absolutely nothing for the *much* more common case of (non-square) rectangular images. For example consider an image that is `100 x 1100`, in which cause you'd obviously only want to resize *one* of the dimensions.

apoorv-mishra:
Ah, sorry for being sloppy @Snuffleupagus. Honestly speaking, I thought about it, but missed it while submitting the patch. Made the corrections now. Unfortunately, still can't get around to run refs locally, so directly submitted here.

apoorv-mishra:
> For example consider an image that is 100 x 1100, in which case you'd obviously only want to resize one of the dimensions.
As I doubted, resizing one dimension doesn't seem to be a good idea @Snuffleupagus . Some of the refs fail due to that.

Snuffleupagus:
> Unfortunately, still can't get around to run refs locally, so directly submitted here.

Unless you actually run tests locally, a patch like this which touches code in the `core/` folder is basically impossible to work on effectively (as you're no double aware of at this point in time).

> As I doubted, resizing one dimension doesn't seem to be a good idea

How could you reasonably have doubted that, when you didn't run any tests locally!?
Besides, just because that didn't immediately work doesn't mean that the idea itself is incorrect, since the latest update will still perform *really* poorly on extremely narrow and large images. (You may simply have "lucked out" on choosing a limit which masks the problem for now, but it doesn't solve anything as-is.)

There's also the question if the downsizing algorithm is even correct and/or good, but I've got neither the time/interest to look into this further (given the time already spent).

apoorv-mishra:
> Unless you actually run tests locally, a patch like this which touches code in the core/ folder is basically impossible to work on effectively (as you're no double aware of at this point in time).

I understand. But, you freshly clone the repo, you run `gulp makeref`, you successfully generate refs. Then you run `gulp test`(yes friend, on `master` synced) and you see `OHNOES! Some tests failed` . I can't understand why that happens. Quite deceptive.

> As I doubted, resizing one dimension doesn't seem to be a good idea 
May be I was wrong about it considering the fact that this would somehow screw up the aspect ratio. But it doesn't work this way. I strongly feel that algorithm works correctly(even for resizing just one dimension) and does what it's supposed to do. My original version was overhauled by @yurydelendik himself. So you may consider the algo coming from him more than me. The only problem was that it "aggressively" downsizes masks. Now, that's a problem because some images are so sensitive that resizing their dimensions even by a factor of 0.75 shows perceptible crooked edges. So, I forced it to downsize by not more than 4096x4096.

> but I've got neither the time/interest to look into this further (given the time already spent).
I understand that your time is important. I am a relatively modest programmer compared to many others working on pdfjs and the likes of you. But this kind of statement(especially the "interest" part) discourages new contributors. Even if I leave the issue, other new contributors would probably hesitate to pick it up, if we write such statements.

Snuffleupagus:
Closing for now, since there's not been any updates for a couple of years and as-is this patch cannot land given that there's too many regressions in existing test-cases; issue #8076 remains open to track this.

PR #9256, opened 2017-12-09
Use native `Math` functions in the custom `log2` function

timvandermeij:
It is quite confusing that the custom function is called `log2` while it actually returns the ceiling value and handles zero and negative values differently than the native function.

To resolve this, we add a comment that explains these differences and make the function use the native `Math` functions internally instead of using our own custom logic. To verify that the function does what we expect, we add unit tests.

All browsers except for IE support `Math.log2` for quite a long time already (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/log2). For IE, we use the core-js polyfill.

According to the microbenchmark at https://jsperf.com/log2-pdfjs/1, using the native functions should also be faster, in my testing almost six times as fast.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -600,13 +600,14 @@ function string32(value) {
                              (value >> 8) & 0xff, value & 0xff);
 }
 
+// Calculate the base 2 logarithm of the number `x`. This differs from the
+// native function in the sense that it returns the ceiling value and that it
+// returns 0 instead of `Infinity`/`NaN` for `x` values smaller than/equal to 0.
 function log2(x) {
-  var n = 1, i = 0;
-  while (x > n) {
-    n <<= 1;
-    i++;
+  if (x <= 0) {
+    return 0;
   }
-  return i;
+  return Math.ceil(Math.log2(x));
 }
 
 function readInt8(data, start) {
--- a/test/unit/util_spec.js
+++ b/test/unit/util_spec.js
@@ -15,7 +15,7 @@
 
 import {
   bytesToString, isArrayBuffer, isBool, isEmptyObj, isNum, isSpace, isString,
-  ReadableStream, removeNullCharacters, stringToBytes, stringToPDFString
+  log2, ReadableStream, removeNullCharacters, stringToBytes, stringToPDFString
 } from '../../src/shared/util';
 
 describe('util', function() {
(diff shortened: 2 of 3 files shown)


Snuffleupagus:
Unfortunately `Math.ceil(Math.log2(x))` still isn't entirely compatible with the old helper function, compare e.g. the result of `x = 0` for the old/new code :-(

timvandermeij:
I have updated the patch to make sure that there are no difference in the implementations anymore.

src/shared/util.js
```diff
@@ -600,13 +600,14 @@ function string32(value) {
                              (value >> 8) & 0xff, value & 0xff);
 }
 
+// Calculate the base 2 logarithm of the number `x`. This differs from
+// the native function in the sense that it returns the ceiling value
+// and that `Infinity`/`NaN` values are treated as 0.
```
Snuffleupagus: This line is really confusing me, since as far as I can tell `log2(Infinity) == Infinity` and `log2(NaN) == NaN` rather than `0`. What's correct here, the code or the comment?

timvandermeij:
*I have updated the commit to adjust the comment above `log2`. No functional changes have been made.*

src/shared/util.js
timvandermeij: Yes, now that I read it again I can see how it's a bit confusing. I rewrote it in the new commit, so hopefully it's better like that.

Snuffleupagus: That's a lot clearer, thanks!

test/unit/util_spec.js
```diff
@@ -136,6 +136,19 @@ describe('util', function() {
...
+  describe('log2', function() {
+    it('handles values smaller than/equal to zero', function() {
+      expect(log2(0)).toEqual(0);
+      expect(log2(-1)).toEqual(0);
+    });
+
+    it('handles values larger than zero', function() {
+      expect(log2(1)).toEqual(0);
+      expect(log2(2)).toEqual(1);
+      expect(log2(3)).toEqual(2);
```
Snuffleupagus: Since our `log2` function doesn't behave like you would expect, based on the name, how about adding a test-case where the input is a floating point number?

Snuffleupagus (approved):
Seems fine to me, with one minor comment; thank you for the patch (and for adding more unit-tests)!

test/unit/util_spec.js
timvandermeij: Good suggestion! Done in the new commit.

Issue #9237, opened 2017-12-04
#findInput::placeholder style is now low-contrast in Chrome - consider setting color


alice:
Per the image below, the placeholder text in `#findInput` is now low contrast because the Chrome UA style changed to improve the placeholder contrast against a white background.
![image](https://user-images.githubusercontent.com/95208/33579847-386b848e-d99f-11e7-92c1-d13f9cec0de9.png)

`#cecece` or lighter would meet AA contrast ratio requirements.

`::placeholder` style is defined here:
https://github.com/mozilla/pdf.js/blob/04b93cf57e36f46a2cbb13c839a9b32fa76ff4e5/web/viewer.css#L444

amit7010:
Hi,
Can i do it?

timvandermeij:
Sure, just keep in mind that a solution for this must work in all browsers, i.e., not be Chrome-specific. Thanks!

amit7010:
Hi,
i used `color:#e0e0e0;` , a lighter shade then `#cecece` because it was little darker for Mozilla Firefox.
Below is the Preview of Mozilla Firefox Browser.
Firefox Quantum
Version: | 57.0.2 (64-bit) 

![pdf_mozilla](https://user-images.githubusercontent.com/22027941/33899441-812e231e-df91-11e7-8de4-15fd8b99ba9a.PNG)

Other Browsers:
Opera
Version: | 49.0.2725.47

![opera_pdf](https://user-images.githubusercontent.com/22027941/33899761-825c2fc8-df92-11e7-8d46-98b4a89eeec4.PNG)

Chrome
Version: | 63.0.3239.84 (Official Build) (64-bit)

![chrome_pdf](https://user-images.githubusercontent.com/22027941/33900757-9ea9a978-df95-11e7-95bd-71c41dea8e1e.PNG)

Microsoft Edge 
Version: | 38.14393.1066.0

I don't know why pdf s are not opening in my edge browser but saw that Input placeholder and not happy about it.

![edge_pdf](https://user-images.githubusercontent.com/22027941/33900635-283a20f6-df95-11e7-98b2-92f67ae0a73a.PNG)

Open for suggestions .. 
Guide me what to do next 
Thank You

PR #9265, opened 2017-12-12
improving visibility of the ghost text which appears in the find bar

himanish-star:
this change was made with respect to #9237 issue

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="Eslint" enabled="true" level="ERROR" enabled_by_default="true" />
+  </profile>
+</component>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -442,6 +442,7 @@ html[dir='rtl'] .findbar {
 }
 #findInput::placeholder {
   font-style: italic;
+  color: #cecece;
 }
 #findInput[data-status="pending"] {
   background-image: url(images/loading-small.png);


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="Eslint" enabled="true" level="ERROR" enabled_by_default="true" />
+  </profile>
+</component>
```
Snuffleupagus: Please remove this file, since it has nothing to do with the PR.

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
```
Snuffleupagus: Please remove this file, since it has nothing to do with the PR.

web/viewer.css
```diff
@@ -442,6 +442,7 @@ html[dir='rtl'] .findbar {
 }
 #findInput::placeholder {
   font-style: italic;
+  color: #cecece;
```
Snuffleupagus: Did you test this in Firefox, since there this change actually causes the placeholder text to be come less legible!?

Given that the issue you're attempting to fix is Chrome only, and that this change has a negative impact elsewhere, let's move this to this to here instead: https://github.com/mozilla/pdf.js/blob/6ac9e1c5ed0d5f067872b8482724c171c79566b2/web/viewer.css#L434

Also, let's use a colour in Chrome that's closer to the one used in Firefox, hence let's try `hsl(0, 0%, 75%)` instead (since the [standard colour](https://github.com/mozilla/pdf.js/blob/6ac9e1c5ed0d5f067872b8482724c171c79566b2/web/viewer.css#L1107) is defined using `hsl` notation).

Snuffleupagus (changes requested):
Please make sure that you remove the unnecessary files that were added!

Also, please update the commit message (and PR title) to something like the following instead:
`Improve visibility of the findInput placeholder in Chrome`,
since "ghost text" isn't really that helpful in understanding the actual change.

Finally, when addressing the comments, please make sure that you squash the commits; see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

.idea/inspectionProfiles/Project_Default.xml
himanish-star: ya sure ,sorry

web/viewer.css
himanish-star: @Snuffleupagus , I'm new to open source , so I made a mistake by closing this pull request as I should have made changes to this only. I realised this mistake of mine and will keep this in mind for further contributions. As of now I have made the changes you had requested for and have opened a new pull request.

PR #9267, opened 2017-12-12
Improve visibility of the findInput placeholder in Chrome

himanish-star:
this pull request is with reference to the issue #9237 and I have closed my previous pull request of #9265 . I have also made the changes requested by @Snuffleupagus


Snuffleupagus (approved):
Thanks for the patch!

Issue #9184, opened 2017-11-23
Text selection color is ugly


fitojb:
I find the text highlight color highly unsatisfactory; it reminds me of Windows 95's default theme. Can we change it to a nicer blue? I recommend #b9eaff or something similar.

bansalnitish:
@timvandermeij I would like to work on this issue. Can you just suggest what color are you looking for ?

timvandermeij:
Let's first talk this over on IRC with `yury` or `bdahl` to see if this is actually something we want to fix. Personally I find the current color just fine.

fitojb:
I won’t be able to use IRC with you guys, but a more permanent medium (such as a GitHub issue _wink_ would be great, as it’s not constrained by timezones and stuff. I bet this will get reworked anyway if you ever implement more Quantum-like visuals.

Snuffleupagus:
> Personally I find the current color just fine.

Agreed, and calling the existing colour "ugly" is highly subjective anyway; closing as WONTFIX?

PR #9270, opened 2017-12-13
improve-text-selection-color

himanish-star:
This PR is for issue #9184 . As mentioned by @fitojb , I have changed color to rgb(185,234,255) which is the same as #b9eaff

### DIFF
--- a/web/text_layer_builder.css
+++ b/web/text_layer_builder.css
@@ -60,8 +60,8 @@
   background-color: rgb(0, 100, 0);
 }
 
-.textLayer ::selection { background: rgb(0,0,255); }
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Did you actually test this, since this makes it *extremely* to spot the difference between selected/non-selected text? The suggested colour is simply way too light to be usable, even for people with good eyesight.

Furthermore, given that there's no actual discussion in the issue, it's not (at all) clear to me if we first of all even want to change the text-selection colour and secondly what a good value would be in a change is deemed necessary.
Hence I'm closing this PR, since it's unfortunately not in a merge-able state.

PR #9271, opened 2017-12-13
Handle broken, Ghostscript generated, Metadata that contains HTML character names (bug 1424938)

Snuffleupagus:
Please note that while this could be considered a regression in user-facing behaviour, I'm not convinced that it's really a regression as such since prior to PR #8912 the Metadata would fail to parse (with an XML error) and thus be ignored when setting the viewer title.
With the refactored Metadata parsing we're now able to parse this, which uncovered issues with a subset of broken Ghostscript Metadata that uses HTML character names.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1424938


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #9272, opened 2017-12-13
Replaced occurence of `throw new Error` with `unreachable` 

### LINKED ISSUE #8846: Replace occurrences of `throw new Error(...)` with `unreachable(...)` where applicable
Especially in abstract methods, we would like to use `unreachable(...)` because it's a more explicit way to indicate that a fail state of the application is reached. Let's ensure that all abstract methods use `unreachable(...)` instead of throwing an error.

Note: we use `unreachable(...)` only for code paths that cannot happen by design (such as directly calling an abstract method). Errors in parsing can happen if the PDF is corrupted for example, so since that is dependent on the input, we throw a regular error there.

nveenjain:
Fixes #8846 
Changes made in all the files where abstract methods might exist.

### DIFF
--- a/src/core/cmap.js
+++ b/src/core/cmap.js
@@ -14,7 +14,8 @@
  */
 
 import {
-  CMapCompressionType, FormatError, isString, MissingDataException, Util, warn
+  CMapCompressionType, FormatError, isString, MissingDataException,
+  unreachable, Util, warn
 } from '../shared/util';
 import { isCmd, isEOF, isName, isStream } from './primitives';
 import { Lexer } from './parser';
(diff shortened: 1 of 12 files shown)


Snuffleupagus (commented):
This looks like a good start, but I think that there's a few additional cases that can be converted. A quick, and possibly incomplete, list would be:
https://github.com/mozilla/pdf.js/blob/master/src/core/font_renderer.js#L631
https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L1477
https://github.com/mozilla/pdf.js/blob/master/src/core/pdf_manager.js#L26
https://github.com/mozilla/pdf.js/blob/master/src/display/dom_utils.js#L530

nveenjain:
@Snuffleupagus Thanks for additional cases. I implemented all of them along with some others. Please have a look now. Thanks :smiley:

Snuffleupagus (approved):
Looks good to me, thanks!

@timvandermeij Want to give it a quick look to double-check that there's no cases missed?

timvandermeij (approved):

nveenjain:
Thanks @Snuffleupagus @timvandermeij, this is my first PR. looking forward to many more :smiley:

Issue #9263, opened 2017-12-11
Some fonts not shown in pdf


gHensges:
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version: Chrome 62
- Operating system and its version: Windows 7
- PDF.js version: Actual demo project 
- Is a browser extension: No
[Rechnungsnummer_16173091.PDF](https://github.com/mozilla/pdf.js/files/1548284/Rechnungsnummer_16173091.PDF)

Steps to reproduce the problem:
1. 
2.

What is the expected behavior? (add screenshot)
![image](https://user-images.githubusercontent.com/29802177/33838547-1d3ed17a-de90-11e7-8bb1-996c72dde36f.png)

What went wrong? (add screenshot)
![image](https://user-images.githubusercontent.com/29802177/33838498-fd0b770a-de8f-11e7-9fcd-3736cd6392ef.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

Snuffleupagus:
```
PDF aafc35539e0207ffb0ed3cc48b25f7fa [1.4 3005ci KX / KMPD6399.DLL Version 5.2] (PDF.js: 2.0.203)   viewer.js:1607:7
Warning: Error during font loading: invalid array length   pdf.js:340:5
```

Snuffleupagus:
Closing as duplicate of issue #7074.

PR #9276, opened 2017-12-14
Fix loca table when offsets aren't in ascending order.

### LINKED ISSUE #7074: Text not displaying at all, images are fine
I've tried multiple browsers and the latest version of pdf.js. The text in the PDF does not show at all, but images and shapes appear fine.

Here is the PDF:

[EE-Summer-2015-for-web Original.pdf](https://github.com/mozilla/pdf.js/files/164771/EE-Summer-2015-for-web.Original.pdf)

And a screenshot from the pdf.js example viewer:

screen shot 2016-03-09 at 08 56 27

The PDF as shown in Preview on a Mac:

screen shot 2016-03-09 at 08 56 41

Sorry that I don't have any insight at all at the moment! Please let me know if you need anything else.

Thanks,

Dan

brendandahl:
Fixes #7074 and #9263 

note: I shouldn't have pushed this branch to pdf.js, but it's there so I'll delete after.


src/core/fonts.js
```diff
@@ -1759,6 +1759,11 @@ var Font = (function FontClosure() {
         var locaCount = dupFirstEntry ? numGlyphs - 1 : numGlyphs;
         for (i = 0, j = itemSize; i < locaCount; i++, j += itemSize) {
           var endOffset = itemDecode(locaData, j);
+          // The spec says the offsets should be in ascending order, however
+          // some fonts use the offset of 0 to mark a glyph as missing.
+          if (endOffset < startOffset) {
```
Snuffleupagus: Just from reading the comment above, I'd sort of expect this condition to read `if (endOffset === 0) {` instead, which seems like it would fix the issues without the regression (although I've not run tests).

brendandahl: I was hoping this would be a more general fix, but turns out there are other bad loca tables. In issue1257 it goes 598,74,74

Snuffleupagus (approved):
Looks good, thanks!

PR #9287, opened 2017-12-16
PDFjs now compatible with Librejs

### LINKED ISSUE #3775: PDF.js is not compatible with LibreJS
Being that pdf.js is free software, it would be in the best interests of the pdf.js team to tag the licenses of all Javascript files part of the project in a way that computers can read and confirm that users are in fact allowed to change the code. A practical implication of this is that pdf.js will no longer be blocked by the LibreJS browser addon.

For an example of how this has been done in a similar web application, see https://github.com/ether/etherpad-lite/pull/1066. For instructions by the LibreJS team, see http://www.gnu.org/software/librejs/free-your-javascript.html.

himanish-star:
I hope this fixes #3775 .

I have made the changes in the files mentioned by @yurydelendik 
1. src/pdf.js
2. web/pdf_page_view.js
3. web/viewer.js
**web/compatibility.js (I couldn't find this file. I may have gone wrong on my part. So if someone could locate this file for me I would fix this too)**

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -127,7 +127,8 @@ function createWebpackConfig(defines, output) {
     BUNDLE_VERSION: versionInfo.version,
     BUNDLE_BUILD: versionInfo.commit,
   });
-  var licenseHeader = fs.readFileSync('./src/license_header.js').toString();
+  var licenseHeaderLibre =
+    fs.readFileSync('./src/license_header_libre.js').toString();
   var enableSourceMaps = !bundleDefines.FIREFOX && !bundleDefines.MOZCENTRAL &&
                          !bundleDefines.CHROME;
   var skipBabel = bundleDefines.SKIP_BABEL ||
@@ -136,7 +137,7 @@ function createWebpackConfig(defines, output) {
   return {
     output: output,
     plugins: [
-      new webpack2.BannerPlugin({ banner: licenseHeader, raw: true, }),
+      new webpack2.BannerPlugin({ banner: licenseHeaderLibre, raw: true, }),
     ],
     resolve: {
       alias: {
@@ -995,7 +996,7 @@ gulp.task('lib', ['buildnumber'], function () {
     content = content.replace(removeCjsSrc, function (all, prefix, suffix) {
       return prefix + suffix;
     });
-    return licenseHeader + content;
+    return licenseHeaderLibre + content;
   }
   var babel = require('babel-core');
   var versionInfo = getVersionJSON();
@@ -1012,7 +1013,8 @@ gulp.task('lib', ['buildnumber'], function () {
       'pdfjs-lib': '../pdf',
     },
   };
-  var licenseHeader = fs.readFileSync('./src/license_header.js').toString();
+  var licenseHeaderLibre =
+    fs.readFileSync('./src/license_header_libre.js').toString();
   var preprocessor2 = require('./external/builder/preprocessor2.js');
   var buildLib = merge([
     gulp.src([
--- a/src/license_header_libre.js
+++ b/src/license_header_libre.js
@@ -1,0 +1,21 @@
+/**
+ * @licstart The following is the entire license notice for the
+ * Javascript code in this page
+ *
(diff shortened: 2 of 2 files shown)


src/pdf.js
```diff
@@ -1,4 +1,8 @@
-/* Copyright 2012 Mozilla Foundation
+/*
+ * @licstart  The following is the entire license notice for the
+ *  JavaScript code in this page.
```
Snuffleupagus: Why is the word `JavaScript` indented with one space here?

Snuffleupagus (changes requested):
First of all, the list in https://github.com/mozilla/pdf.js/issues/3775#issuecomment-157149664 is more than two years old, hence it's no longer current.

*Edit:* I'm also guessing that you didn't actually test these changes by building PDF.js, e.g. using `gulp generic` locally, and check the resulting files (since the licence headers you modified are all stripped out during the build step)!?

In general, I do *not* think that we should be editing the licence header of individual files like this, since it for one doesn't scale well.
A much better solution would be to add this during the build step instead!

I'm not sure if we should change the https://github.com/mozilla/pdf.js/blob/master/src/license_header.js file. It might be better to add a new one e.g. `license_header_libre.js` and use that in https://github.com/mozilla/pdf.js/blob/master/gulpfile.js instead.

himanish-star:
@Snuffleupagus , I have tried to make the changes requested. I have replaced the existing usages of license_header.js with license_header_libre.js. I made these replacements instead of using both as using both will lead to two license headers. Thus I have replaced the former one with the latter one.

gulpfile.js
```diff
@@ -127,7 +127,9 @@ function createWebpackConfig(defines, output) {
     BUNDLE_VERSION: versionInfo.version,
     BUNDLE_BUILD: versionInfo.commit,
   });
-  var licenseHeader = fs.readFileSync('./src/license_header.js').toString();
+  var licenseHeaderLibre = fs
+                           .readFileSync('./src/license_header_libre.js')
+                           .toString();
```
Snuffleupagus: This isn't how we usually break lines in PDF.js, let's use to following format instead:
```javascript
  var licenseHeaderLibre =
    fs.readFileSync('./src/license_header_libre.js').toString();
```

gulpfile.js
```diff
@@ -1012,7 +1014,9 @@ gulp.task('lib', ['buildnumber'], function () {
       'pdfjs-lib': '../pdf',
     },
   };
-  var licenseHeader = fs.readFileSync('./src/license_header.js').toString();
+  var licenseHeaderLibre = fs
+                           .readFileSync('./src/license_header_libre.js')
+                           .toString();
```
Snuffleupagus: The formatting comment above applies here as well.

Snuffleupagus (approved):
That looks much better, thanks!

@yurydelendik, @timvandermeij Would you agree that this is a reasonable way to address the issue?

gulpfile.js
himanish-star: Yup , I get it . Working on the indentation now

timvandermeij (approved):
Looks like a good solution to me. If we are not using the regular `license_header.js` file anymore, I'm also fine with changing that one instead of creating a new one, but either way is fine by me. I'll let @yurydelendik decide on that one.

Snuffleupagus:
> I'm also fine with changing that one instead of creating a new one, but either way is fine by me.

Since every file should have a licence header, and we're linking to that file from https://github.com/mozilla/pdf.js/wiki/Style-Guide, it thus seemed like a good idea to keep it around.

timvandermeij:
Nice, I had not noticed that one yet. It's indeed good to have that in the repository. Let's keep it this way then. If Yury is also fine with the current diff, we can land it.

himanish-star:
@Snuffleupagus , @timvandermeij any updates ?

timvandermeij:
I have looked into this a bit more and this is indeed all logic we use for the license headers in built files, so let's merge this.

timvandermeij:
Thank you for the patch!

PR #9288, opened 2017-12-16
Handle PDF files with missing 'endobj' operators, by searching for the "obj" string rather than "endobj" in `XRef.indexObjects` (issue 9105)

### LINKED ISSUE #9105: Invalid root reference when parsing pdf

Snuffleupagus:
This patch refactors the searching for 'endobj', to try and find the next occurance of "obj" and then check if it was in fact an 'endobj' and continue searching otherwise.
This approach is used to avoid having to first find 'endobj', and then re-check the entire contents of the object and having to run (potentially expensive) regular expressions on arbitrary long strings.

Fixes #9105.

---
Replaces PR #9253, since the approach used there would cause needless re-parsing of data and potentially bad regular expression performance.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -1102,10 +1102,14 @@ var XRef = (function XRefClosure() {
         return skipped;
       }
       var objRegExp = /^(\d+)\s+(\d+)\s+obj\b/;
+      const endobjRegExp = /\bendobj[\b\s]$/;
+      const nestedObjRegExp = /\s+(\d+\s+\d+\s+obj[\b\s])$/;
+      const CHECK_CONTENT_LENGTH = 25;
+
       var trailerBytes = new Uint8Array([116, 114, 97, 105, 108, 101, 114]);
       var startxrefBytes = new Uint8Array([115, 116, 97, 114, 116, 120, 114,
                                           101, 102]);
-      var endobjBytes = new Uint8Array([101, 110, 100, 111, 98, 106]);
+      const objBytes = new Uint8Array([111, 98, 106]);
       var xrefBytes = new Uint8Array([47, 88, 82, 101, 102]);
 
       // Clear out any existing entries, since they may be bogus.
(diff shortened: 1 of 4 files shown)


src/core/obj.js
```diff
@@ -1147,8 +1148,33 @@ var XRef = (function XRefClosure() {
...
           }
-          var contentLength = skipUntil(buffer, position, endobjBytes) + 7;
-          var content = buffer.subarray(position, position + contentLength);
+          let contentLength, startPos = position + token.length;
+
+          // Find the next "obj" string, rather than "endobj", to ensure that
+          // we won't skip over a new 'obj' operator in corrupt files where
+          // 'endobj' operators are missing (fixes issue9105_reduced.pdf).
+          while (startPos < buffer.length) {
+            let endPos = startPos + skipUntil(buffer, startPos, objBytes) + 4;
```
timvandermeij: What does this number 4 stand for? Can we make a constant out of it or add a comment?

src/core/obj.js
```diff
@@ -1147,8 +1148,33 @@ var XRef = (function XRefClosure() {
...
+            contentLength = endPos - position;
+
+            let tokenStr = bytesToString(buffer.subarray(endPos - 8, endPos));
```
timvandermeij: What does this number 8 stand for? Can we make a constant out of it or add a comment?

src/core/obj.js
```diff
@@ -1147,8 +1148,33 @@ var XRef = (function XRefClosure() {
...
+            if (endobjRegExp.test(tokenStr)) {
+              break; // Found 'endobj'.
+            } else {
+              let objStartPos = Math.max(endPos - 25, startPos);
```
timvandermeij: What does this number 25 stand for? Can we make a constant out of it or add a comment?

src/core/obj.js
```diff
@@ -1147,8 +1148,33 @@ var XRef = (function XRefClosure() {
...
+              let objStr = bytesToString(buffer.subarray(objStartPos, endPos));
+
+              let objToken = /\d+\s+\d+\s+obj[\b\s]$/.exec(objStr);
```
timvandermeij: There is an `objRegExp` above which looks very similar. Shouldn't we use that? If not, can we cache this regex?

src/core/obj.js
Snuffleupagus: There's existing magic numbers in this code, e.g. https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L1150, which is why I figured it wouldn't hurt if more were added.

I'll have to think about this, since comments or meaningful constant names could end up bloating this code quite a bit :-(

src/core/obj.js
Snuffleupagus: The operative work is "similar", not identical ;-)
So no, we cannot use that one since we need to check when *end* of a string here (as opposed to the beginning above), and I'm not sure if caching really matters since this particular code-patch should almost never be hit in practice.

src/core/obj.js
Snuffleupagus: > What does this number 4 stand for?

`skipUntil` gives us the starting point of `objBytes`, so we need to get the position after it; compare with https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L1150.

src/core/obj.js
Snuffleupagus: Since a `1 0 obj` (or similar) sequence can contain arbitrary whitespace, we need to check enough of the data to ensure that we won't miss e.g. `1     0     obj`.

In this case though, I'd agree that a constant is probably better especially considering that this number is somewhat randomly chosen.

src/core/obj.js
Snuffleupagus: > What does this number 8 stand for?

It's just to get the end of the stream such that we can check that if e.g. "&nbsp;endobj&nbsp;" was found; i.e. the endobj operator.

timvandermeij (approved):

timvandermeij:
Thank you for the fix!

PR #9294, opened 2017-12-19
Remove even more polyfills for old, and now unsupported, browsers

Snuffleupagus:
This PR makes the following changes to our (implicit) browser support:
 - IE 10 and below will now be *explicitly* unsupported; please see https://en.wikipedia.org/wiki/Internet_Explorer_version_history#Release_history_for_desktop_Windows_OS_version
 - Android 4 and below will now be *explicitly* unsupported; please see https://en.wikipedia.org/wiki/Android_version_history
 - Safari 7 and below will now be *explicitly* unsupported; please see https://en.wikipedia.org/wiki/Safari_version_history

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -111,31 +111,6 @@ PDFJS.compatibilityChecked = true;
   });
 })();
 
-// Provides correct document.readyState value for legacy browsers.
-// Support: IE9,10.
-(function checkDocumentReadyState() {
-  if (!hasDOM) {
-    return;
-  }
-  if (!document.attachEvent) {
-    return;
-  }
-  var documentProto = document.constructor.prototype;
-  var readyStateProto = Object.getOwnPropertyDescriptor(documentProto,
-                                                        'readyState');
-  Object.defineProperty(documentProto, 'readyState', {
-    get() {
-      var value = readyStateProto.get.call(this);
-      return value === 'interactive' ? 'loading' : value;
-    },
-    set(value) {
-      readyStateProto.set.call(this, value);
-    },
-    enumerable: true,
-    configurable: true,
-  });
-})();
-
 // Provides support for ChildNode.remove in legacy browsers.
 // Support: IE.
 (function checkChildNodeRemove() {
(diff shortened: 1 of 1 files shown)


timvandermeij:
I think we should update the table on the FAQ page with all explicitly supported versions once this lands. I do like how this patch not only cleans up old code, but also helps us to figure out which exact versions we no longer support. Before landing this I would like to know if @yurydelendik is OK with the browser support that follows from this.

yurydelendik:
> Before landing this I would like to know if @yurydelendik is OK with the browser support that follows from this.

I agree with removing. I'm just curious about IE11 -- I don't want to remove support for it just yet by accident.

We probably need to reference the version of capability.js (using SHA) that has these polyfills in FAQ, just in case if somebody will want to experiment.

Snuffleupagus:
> I think we should update the table on the FAQ page with all explicitly supported versions once this lands.

One thing that I suppose we might want to consider, is to only *explicitly support* currently maintained releases of various browsers (that we've tested) to be able to remove old compatibility code quicker in the future. And then perhaps a list of *explicitly unsupported* browsers, where we know from PRs such as this one, that it cannot work and no issues/patches will be accepted either.
Finally, a "kind-of-supported" category where e.g. performance could be bad (such as. IE11), but it should work with `compatibility.js` and issues/PRs are accepted.

Snuffleupagus:
> I agree with removing. I'm just curious about IE11 -- I don't want to remove support for it just yet by accident.

IE11 should still be supported just fine, since I took care to only remove polyfills with `IE<11` and have also successfully tested the preview in https://github.com/mozilla/pdf.js/pull/9294#issuecomment-352766020 with a couple of different files in IE11.[1]

---
 [1] I purposely removed each polyfill in a separate commit, to make it easier to if necessary bisect exactly where something broke, on the off-change that it'd every be necessary.

timvandermeij (approved):

timvandermeij:
Looks good. Thank you for cleaning this up!

timvandermeij:
I have updated the project after this change. I left the note about the browser support because we still need to update the documentation now that this has been merged. I like the idea in https://github.com/mozilla/pdf.js/pull/9294#issuecomment-353021820 for that because then it's more explicit what we support and what we do not support.

PR #9299, opened 2017-12-20
Attempt to actually resolve ColourSpace names in accordance with the specification (issue 9285)

### LINKED ISSUE #9285: pdf completely rendered blank

Configuration:
Chrome Version 63
Windows 7
- PDF.js version:
Your actual Demo Project
- Is a browser extension:

1. 
2.

image

image

Snuffleupagus:
Please refer to the PDF specification, in particular http://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G7.3801570, where the *second* point is the most relevant here:

> A colour space shall be specified in one of two ways:
>  - Within a content stream, the CS or cs operator establishes the current colour space parameter in the graphics state. The operand shall always be name object, which either identifies one of the colour spaces that need no additional parameters (DeviceGray, DeviceRGB, DeviceCMYK, or some cases of Pattern) or shall be used as a key in the ColorSpace subdictionary of the current resource dictionary (see 7.8.3, "Resource Dictionaries"). In the latter case, the value of the dictionary entry in turn shall be a colour space array or name. A colour space array shall never be inline within a content stream.
>
> - Outside a content stream, certain objects, such as image XObjects, shall specify a colour space as an explicit parameter, often associated with the key ColorSpace. In this case, the colour space array or name shall always be defined directly as a PDF object, not by an entry in the ColorSpace resource subdictionary. ***This convention also applies when colour spaces are defined in terms of other colour spaces.***

This passes all tests locally, and hopefully there won't be PDF generators that violate this part of the specification.

Fixes #9285, courtesy of implementing the emphasised part above.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -258,8 +258,8 @@ var ColorSpace = (function ColorSpaceClosure() {
     }
   };
 
-  ColorSpace.parseToIR = function(cs, xref, res, pdfFunctionFactory) {
-    if (isName(cs)) {
+  ColorSpace.parseToIR = function(cs, xref, res = null, pdfFunctionFactory) {
+    if (res && isName(cs)) {
       var colorSpaces = res.get('ColorSpace');
       if (isDict(colorSpaces)) {
         var refcs = colorSpaces.get(cs.name);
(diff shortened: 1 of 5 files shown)


Snuffleupagus:
> This convention also applies when colour spaces are defined in terms of other colour spaces.

Looking at the specification text once more, I'm now unsure if the above can actually be assumed to hold *generally* or if it's rather only applicable to the "Outside a content stream," case (given the placement of that sentence)!?

The patch is currently assuming that it holds generally, and if that's not the case then we'd probably need to implement some sort of `parsedColourSpaces` cache to fix the issue at hand.
So despite all tests passing, there's a very real risk the patch as-is could cause regressions in situations where we're currently lacking test-coverage.

PR #9303, opened 2017-12-21
[api-major] Remove the unused `encrypted` property from the `pdfInfo` object sent from the worker via the `GetDoc` message

Snuffleupagus:
I recall being confused as to the purpose of the `encrypted` property all the way back when working on PR #4750.

Looking at the history, this property was added in PR #1698 when password support was added to the API/viewer. However, its only purpose seem to have been to facilitate the addition of a `isEncrypted` function in the API. That function never, as far as I can tell, saw any use and was unceremoniously removed in PR #4144.

Since we want to avoid sending all non-essential data early during initial document loading (e.g. PR #4750), it seems correct to get rid of the `encrypted` property. Especially since it hasn't even been exposed in the API for over three years, with no complaints that I'm aware of.

Finally note that the `encrypt` property on the `XRef` instance isn't tied to the code that's being removed here. Given that we're calling `PDFDocument.parse` during `createDocumentHandler` in the worker which, via `PDFDocument.setup`, calls `XRef.parse` where the `Encrypt` data (if it exists) is always parsed.

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -402,19 +402,15 @@ var WorkerMessageHandler = {
       var loadDocumentCapability = createPromiseCapability();
 
       var parseSuccess = function parseSuccess() {
-        var numPagesPromise = pdfManager.ensureDoc('numPages');
-        var fingerprintPromise = pdfManager.ensureDoc('fingerprint');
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good find!

PR #9309, opened 2017-12-22
Keep track of parsed (array) ColourSpaces in `ColorSpace.parseToIR` to prevent infinite loops in corrupt PDF files (issue 9285)

### LINKED ISSUE #9285: pdf completely rendered blank

Snuffleupagus:
Since (array) ColourSpaces often aren't simply accessed via References alone, we need to check their actual content to be able to keep track of already parsed ones in `ColorSpace.parseToIR`.

Given the fairly varied data structures that can occur in different types of ColourSpaces, the only simple way that I could come up with was to utilize `MurmurHash3_64` to compute a hash of each ColourSpace.
Obviously doing this isn't free, but since `MurmurHash3_64` is fairly efficient the performance impact should be small. I temporarily added `console.time/timeEnd` around the code doing the hash computations and cache checks, and manually tested a few PDF files. On average, it seems that this patch incurs an additional run-time cost of `< 0.5 milli-seconds`, which does seem acceptable at least to me.

Fixes #9285.

---

*Edit:* Supersedes PR #9299.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -258,8 +258,8 @@ var ColorSpace = (function ColorSpaceClosure() {
     }
   };
 
-  ColorSpace.parseToIR = function(cs, xref, res, pdfFunctionFactory) {
-    if (isName(cs)) {
+  ColorSpace.parseToIR = function(cs, xref, res = null, pdfFunctionFactory) {
+    if (res && isName(cs)) {
       var colorSpaces = res.get('ColorSpace');
       if (isDict(colorSpaces)) {
         var refcs = colorSpaces.get(cs.name);
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -14,9 +14,59 @@
  */
 
 import {
-  FormatError, info, isString, shadow, unreachable, warn
+  assert, FormatError, info, isArrayBuffer, isNum, isString, shadow,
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

brendandahl:
It looks like the recursion comes from a colorspace named 'DeviceRGB'. Couldn't we just change it so if one of the reserved word color space names is used it will always return that colorspace e.g. do not look up "DeviceRGB" in the colorspaces resources dictionary?

Snuffleupagus:
> It looks like the recursion comes from a colorspace named 'DeviceRGB'. Couldn't we just change it so if one of the reserved word color space names is used it will always return that colorspace e.g. do not look up "DeviceRGB" in the colorspaces resources dictionary?

I was initially a bit worried that doing this could cause problems with badly generated PDF file, where using "overridden" standard ColourSpace names would be required for correct rendering (although this is *not* the case with the referenced issue).
However, given that all tests pass when using the (better) approach suggested above, let's try that instead since it's both simpler and also seems more inline with the actual wording of the specification.

@brendandahl Thanks a lot for your feedback here; can you please check if PR #9352 looks better here.

PR #9311, opened 2017-12-23
Implement unit tests for the network utility code

timvandermeij:
This should provide 100% coverage for the file.

### DIFF
--- a/test/unit/network_utils_spec.js
+++ b/test/unit/network_utils_spec.js
@@ -1,0 +1,178 @@
+/* Copyright 2017 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


test/unit/network_utils_spec.js
```diff
@@ -0,0 +1,178 @@
...
+  describe('validateRangeRequestCapabilities', function() {
+    const defaultValues = {
+      allowRangeRequests: false,
+      suggestedLength: undefined,
+    };
+
+    it('rejects range chunk sizes that are not strictly positive', function() {
+      expect(function() {
+        validateRangeRequestCapabilities({ rangeChunkSize: 0, });
+      }).toThrow(new Error());
```
Snuffleupagus: This is an existing issue, but can you please add a suitable error message to the `assert` in the core code, such that we can test that the "correct" error is thrown here?

Snuffleupagus (approved):
Looks good to me, with one small question, thank you!

*Edit:* Merry Christmas!

test/unit/network_utils_spec.js
timvandermeij: Done in the new commit.

Thank you for the review, and I wish you a merry Christmas too!

PR #9313, opened 2017-12-24
Add a fallback for non-embedded LucidaSans-Demi fonts (issue 9291)

### LINKED ISSUE #9291: Bold font not rendering properly
[PDF to debug.pdf](https://github.com/mozilla/pdf.js/files/1566831/PDF.to.debug.pdf)

Configuration:
- Web browser and its version: Firefox Quantum - 57.0.1 (64-bit)
- Operating system and its version: macOS High-Sierra 
- PDF.js version: PDF version which is bundled in the browser
- Is a browser extension: NO 

1. Open the pdf in native firefox pdf viewer which uses pdf js
2. The pdf has some bold fonts in page 2 which is not rendered in pdf js

screen shot 2017-12-14 at 12 02 20 pm

screen shot 2017-12-14 at 12 02 39 pm

Snuffleupagus:
The PDF file in the issue uses a number of *embedded* versions of Lucida fonts, but for some reason does *not* embed the LucidaSans-Demi font. According to https://en.wikipedia.org/wiki/Lucida#Usages that one should be bold, so we can at least improve rendering here (even though it won't look perfect).

Fixes #9291.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #8589, opened 2017-06-29
Fix missing notdef in expert encoding.

brendandahl:
Fixes #8586

PR #9314, opened 2017-12-24
Implement unit tests for the encodings and fix missing items

timvandermeij:
Initially I just implemented the unit tests, but quickly found that they were failing my expectation of having a size of 256 items. Some of them did contain 256 items and some did not. I looked up various resources and figured that they indeed all need to have 256 items. One of the good
resources is https://github.com/davidben/poppler/blob/master/poppler/FontEncodingTables.cc

Aside from some missing `notdef` (empty string) entries at the end of the arrays, which I assume causes issues since it may cause out-of-bounds array access which in JavaScript gives `undefined`, there was a `notdef` entry missing in the `MacExpertEncoding`, causing the entries after that to be shifted. This fix for this is similar to the one in #8589.

The unit tests verify that, for known encoding names, the return value is not only an array, but that it is also of the right length and contains only strings.

### DIFF
--- a/src/core/encodings.js
+++ b/src/core/encodings.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-var ExpertEncoding = [
+const ExpertEncoding = [
   '', '', '', '', '', '', '', '', '', '', '', '', '', '', '',
   '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '',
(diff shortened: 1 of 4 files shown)


Snuffleupagus (approved):
> [...] there was a notdef entry missing in the MacExpertEncoding, causing the entries after that to be shifted. This fix for this is similar to the one in #8589.

That's a really nice catch; it's interesting that we, to my knowledge, haven't seen any bugs caused by this yet.

Thank you for the patch!

Issue #6737, opened 2015-12-08
Title of the document is not visible with proper color 


SanvidManiyar:
Hi All,

we are using PDF.js and observed that following types of documents are not rendering correctly in IE11, Chrome and FF. it lost its color and header tittle.
[Florida Industrial MarketView Q3 2015.pdf](https://github.com/mozilla/pdf.js/files/55597/Florida.Industrial.MarketView.Q3.2015.pdf)

I have open the document from the link http://mozilla.github.io/pdf.js/web/viewer.html and it gives me the same result. find the attached screen.

![image](https://cloud.githubusercontent.com/assets/6779439/11658632/9ed870c2-9de9-11e5-9107-c3bd353268ba.png)

is there any solution for this issue  ?

Regards,
Sanvid

Snuffleupagus:
It seems that PR #6606 would improve this quite a bit.

SanvidManiyar:
Hi All,

We have tried replacing files "pattern_helper.js" with the latest version but it gives javascript error while loading a document. 

Is there any other approach to get rid of this problem ?

Regards,
Sanvid

SanvidManiyar:
Hi All,

Could you let us know that which JS is responsible to render the header content is it "pattern_helper.js"  ? So, that I can also try to improve at our my end.

Regards
Sanvid

timvandermeij:
I don't think just updating `pattern_helper.js` is enough. You will need to update all files.

SanvidManiyar:
Hi,

Thanks For your reply. I have also change the other files mentioned with the latest version of PDF.js but it seems that the base version which is used to develop a solution of #6606 is different from the latest version. 

Replacing other mentioned files with pattern_helper.js gives an error while loading the document.

Is it possible to merge this change with the latest version ? so that we can have this improvement in latest version of DDF.js.

Regards,
Sanvid

timvandermeij:
Yes, the #6606 has merge conflicts, so the original file has changed since that patch. Because that PR has not been merged and is not ready yet, it is not recommended to use it yet, but if you really need it you can apply the changes from https://github.com/mozilla/pdf.js/pull/6606/files manually.

SanvidManiyar:
Hi,

Thanks you for your reply.

With a view to apply the changes, Could you suggest which version should I take as base version as the latest one is not supporting change files ?

Regards,
Sanvid

PR #6606, opened 2015-11-07
Improve performance and correctness of Tiling Patterns

Rob--W:
This PR significantly improves the performance of PDFs with tiling patterns (which also fixes a bunch of blurry images), and fixes a correctness issue caused by a missing transform.

Fixes #6496
Fixes #5698
Fixes #1434
Fixes #2825 (the rendering issues have been fixed, but there are still spikes in memory usage (which is very problematic on low-memory devices), this is probably caused by https://crbug.com/448150, and possibly the same issue as https://github.com/bp74/StageXL/issues/214?. I think that we can work around this memory spike by introducing an artificial delay in our task scheduler to give the browser a chance to GC the temporary CanvasGradient patterns - this could be added in a separate PR).

Fixes #6737
Fixes #7731
Fixes #8264
Fixes #8579
Fixes #9922
Fixes #9927
Fixes #9928
Fixes #7946
Fixes #8083
Fixes #8373
Fixes #9770
Fixes #9665
Fixes #9763
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1467424

PR #9320, opened 2017-12-27
Avoid rendering errors by passing in the `webGLContext` when creating a new `CanvasGraphics` in `getColorN_Pattern` (PR 9095 follow-up)

Snuffleupagus:
This was an oversight in PR #9095, which unfortunately breaks rendering in some PDF files (e.g. the one from issue #6737).

It thus appears that we don't have any test-coverage for this code-path, and given the relative complexity of the PDF files affected by this bug I wasn't able to easily create a reduced test-case.
*Please note:* The linked test-case included in this patch is currently *not* rendered correctly (that'd be the PR #6606), but it at least gives us some test-coverage here.


timvandermeij (approved):

timvandermeij:
Good fix, thank you!

PR #9321, opened 2017-12-27
Remove the `background` compatibility hacks for `linkAnnotation` in old versions of IE

Snuffleupagus:
With PDF.js version `2.0` we'll no longer support IE 9 and 10, hence these hacks can now be removed.


timvandermeij (approved):

timvandermeij:
Thank you!

Issue #9020, opened 2017-10-11
Failure to open local file in windows platform


juncaixinchi:
Configuration:

- Web browser and its version: 
Electron 1.7.8 ( use webpack and set `target: 'electron-renderer'`)
- Operating system and its version:
Windows 10
- PDF.js version:
pdfjs-dist: 1.9.640
- Is an extension:
no

Steps to reproduce the problem:
1. try to open `C:/User/1234/Desktop/a.pdf`
2. window.location transform it to `file:///C:/User/1234/Desktop/a.pdf`
3. PDFNodeStreamFsFullReader get `/C:/User/1234/Desktop/a.pdf`
4. ENOENT: no such file or directory, lstat `C:\C:\User\1234\Desktop\a.pdf`

What went wrong?

the pr https://github.com/mozilla/pdf.js/pull/8928 is not working

path before `decodeURI`: `/C:/User/1234/Desktop/a.pdf`. Howerver, after `decodeURI`, path is still the same.

see also 
https://github.com/mozilla/pdf.js/issues/8946

pcad:
@juncaixinchi - We had a similar issue. PDFJS switched in the 1.9 release to the fetch api.  Fetch doesn't work the same way as XHR wrt the file:// protocol.

https://github.com/electron/electron/issues/9474#issuecomment-301810762

juncaixinchi:
@pcad if you use webpack, you can set `target: 'electron-renderer'` or `target: 'electron` in `webpack.config.js`, then PDFJS will using Node Stream to load local file rather than fetch api.

However, due to the bug I refered in this issue, it's only working in Linux or Mac platform, failed in windows.
About fetch, `file:// protocol` was refused to add to fetch api.  ╮(￣ . ￣)╭

https://github.com/github/fetch/pull/92#issuecomment-140665932

matteodem:
So what's the solution to this problem?

juncaixinchi:
@matteodem  A PR(https://github.com/mozilla/pdf.js/pull/9323) has been added to solve this problem, you can try the latest release, it should be OK.

matteodem:
@juncaixinchi I get following error:

``` 
fetch_stream.js:96 Fetch API cannot load file:///C:/Users/demichma/Downloads/ZH_BD_CD_07773.pdf. URL scheme must be "http" or "https" for CORS request.
```

see http://jsfiddle.net/9engc9mw/2601/

Does the build version have to be updated too?

[...]

PR #9323, opened 2017-12-28
Get correct path in node_stream on windows platform( issue #9020)

juncaixinchi:
Fixes https://github.com/mozilla/pdf.js/issues/9020

Get the correct path from urls like `file:///C:/User/1234/Desktop/a.pdf`

### DIFF
--- a/src/display/node_stream.js
+++ b/src/display/node_stream.js
@@ -24,6 +24,8 @@ import {
 } from '../shared/util';
 import { validateRangeRequestCapabilities } from './network_utils';
 
+const fileUriRegex = /^file:\/\/\/[a-zA-Z]:\//;
+
 class PDFNodeStream {
   constructor(source) {
     this.source = source;
@@ -362,7 +364,13 @@ class PDFNodeStreamRangeReader extends BaseRangeReader {
 class PDFNodeStreamFsFullReader extends BaseFullReader {
   constructor(stream) {
     super(stream);
+
     let path = decodeURI(this._url.path);
+
+    // Remove the extra slash to get right path from url like `file:///C:/`
+    if (fileUriRegex.test(this._url.href)) {
+      path = path.replace(/^\//, '');
+    }
 
     fs.lstat(path, (error, stat) => {
       if (error) {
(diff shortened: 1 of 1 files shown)


juncaixinchi:
anyone to review the code ？ @timvandermeij  @Snuffleupagus

src/display/node_stream.js
```diff
@@ -362,8 +362,13 @@ class PDFNodeStreamRangeReader extends BaseRangeReader {
 class PDFNodeStreamFsFullReader extends BaseFullReader {
   constructor(stream) {
     super(stream);
+
     let path = decodeURI(this._url.path);
 
+    if (/^file:\/\/\/[a-zA-Z]:\//.test(this._url.href)) {
```
timvandermeij: This regex looks like the same one below. Can we cache it at the top of the as a constant, e.g., something like:

```javascript
const fileUriRegex = /^file:\/\/\/[a-zA-Z]:\//
```

and then use `fileUriRegex` here?

src/display/node_stream.js
```diff
@@ -362,8 +362,13 @@ class PDFNodeStreamRangeReader extends BaseRangeReader {
...
+      path = path.replace(/^\//, '');
```
timvandermeij: Can you add a comment above the `if` statement for why this is necessary, i.e., what the exact problem on Windows is that this solves?

src/display/node_stream.js
```diff
@@ -384,8 +389,14 @@ class PDFNodeStreamFsRangeReader extends BaseRangeReader {
...
 
+    let path = decodeURI(this._url.path);
+
+    if (/^file:\/\/\/[a-zA-Z]:\//.test(this._url.href)) {
+      path = path.replace(/^\//, '');
+    }
+
     this._setReadableStream(
-      fs.createReadStream(decodeURI(this._url.path), { start, end: end - 1, }));
+      fs.createReadStream(decodeURI(path), { start, end: end - 1, }));
```
timvandermeij: You're now decoding the path twice because `path` is already decoded above. You should be able to remove the `decodeURI` call here.

timvandermeij (commented):
Squash the commits into one after you've made these changes. Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits if you're not familiar with the process. Thank you!

timvandermeij:
Looks good to me this way.

@mukulmishra18 Do you have time to take a second look at this?

mukulmishra18 (approved):
Looks good to me. @timvandermeij feel free to merge it.

timvandermeij (approved):

timvandermeij:
Thank you!

PR #9325, opened 2017-12-29
Adjust `decodeACSuccessive` in src/core/jpg.js to improve the rendering quality of (progressive) JPEG images

Snuffleupagus:
I've been looking into the remaining point in #8637 about blurry images, to see if we could perhaps improve the rendering quality slightly there. After quite a bit of debugging, it seems that the issue is limited to certain progressive JPEG images.

As mentioned previously, I've got no detailed knowledge of the JPEG format, but this patch does seem to improve things quite a bit for the images in question.
Squinting at https://searchfox.org/mozilla-central/rev/6c33dde6ca02b389c52e8db3d22494df8b916f33/media/libjpeg/jdphuff.c#492-639, it seems reasonable that we should take the sign of the data into account. Furthermore, looking at the specification in https://www.w3.org/Graphics/JPEG/itu-t81.pdf#page=118, the "F.2.4.3 Decoding the binary decision sequence for non-zero DC differences and AC coefficients" section even contains a description of this (even though I cannot claim to really understand the details).

---
**Current master, with *native* browser image decoding:**
master

**Current master, with the `src/core/jpg.js` decoder:**
master_built-in-jpeg-decoder

**This patch, with the updated `src/core/jpg.js` decoder:**
patch_built-in-jpeg-decoder


src/core/jpg.js
```diff
@@ -235,7 +235,8 @@ var JpegImage = (function JpegImageClosure() {
       var s;
       var rs;
       while (k <= e) {
-        var z = dctZigZag[k];
+        let offsetZ = offset + dctZigZag[k];
+        let sign = component.blockData[offsetZ] < 0 ? -1 : 1;
```
Snuffleupagus: I suppose that we could have re-added `Util.sign`, i.e. revert commit https://github.com/mozilla/pdf.js/commit/5b67c7594cce0ff48e73420f2a1d08b12e3da0ba, but that just didn't seem worthwhile given that this is only used in *one single* spot in the code.

Snuffleupagus: I suppose that we could have re-added `Util.sign`, i.e. revert commit https://github.com/mozilla/pdf.js/commit/5b67c7594cce0ff48e73420f2a1d08b12e3da0ba, but it just didn't seem worthwhile given that this is only used in *one* single spot in the code.

test/test_manifest.json
```diff
@@ -3589,6 +3589,14 @@
        "type": "eq",
        "nativeImageDecoderSupport": "none"
     },
+    {  "id": "decodeACSuccessive",
+       "file": "pdfs/decodeACSuccessive.pdf",
+       "md5": "7749c032624fe27ab8e8d7d5e9a4a93f",
+       "rounds": 1,
+       "link": true,
```
timvandermeij: This should be removed because the test case is in the repository.

timvandermeij: This should be removed; the test case is in this commit itself.

timvandermeij (approved):
Looks good with this final comment fixed. Thank you for looking into this!

src/core/jpg.js
timvandermeij: Good choice. I don't think this is worth a utility function, not only because it's really simple, but also because it's only used in one place.

test/test_manifest.json
Snuffleupagus: D'oh; good eye, fixed now!

PR #9327, opened 2017-12-29
Avoid truncating JPEG images with DeviceGray ColourSpaces when using the `src/core/jpg.js` built-in decoder

Snuffleupagus:
The bug that this patch fixes is limited to the built-in JPEG decoder, and was unearthed by PR #9260. The underlying issue has existed since PR #6984, where the contents of this patch ought to have been included (if it weren't for the fact that we had no *easy* way to test `src/core/jpg.js` back then).

*Please note:* The slight movement in the reference test is a result of using the `src/core/jpg.js` decoder, rather than the native browser one.

### DIFF
--- a/src/core/image.js
+++ b/src/core/image.js
@@ -546,14 +546,21 @@ var PDFImage = (function PDFImageClosure() {
           }
           return imgData;
         }
-        if (this.image instanceof JpegStream && !this.smask && !this.mask &&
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #9329, opened 2017-12-30
Stop using the deprecated `gulp-util` module

timvandermeij:
The `gulp-util` module is now deprecated and authors are asked to stop using it (refer to https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5 for more information).

PDF.js does not rely on it that much, fortunately, so it's relatively easy for us to remove the dependency. This patch does that by making the following changes:

- Require `gulp-zip` version 4.1.0 or higher since they already removed their `gulp-util` dependency in that version.
- Replace `gulp-util.log` with the `fancylog` module as recommended in the article above.
- Replace `gulp-util.File` with the `Vinyl` module as recommended in the article above.

The only change I had to make for Vinyl is removing the `base` and `cwd` lines since they may not be empty strings anymore. This way we fall back to the defaults Vinyl provides, which for us doesn't matter since we move the file afterwards anyway. Moreover, I used `vfs` for `vinyl-fs` in the `Gulpfile` to avoid confusion with `vinyl` (which is also how the documentation names the variable).

This is all we can do on our side; the other modules that still use `gulp-util` must be updated upstream.

Note that this is necessary in order to be able to support the upcoming Gulp 4 (https://github.com/gulpjs/gulp-util/issues/143).


demurgos (approved):

Snuffleupagus (approved):
Thank you!

PR #9330, opened 2017-12-31
Remove the `CustomStyle` class

timvandermeij:
It is only used in a few places to handle prefixing style properties if necessary. However, we used it only for `transform`, `transformOrigin` and `borderRadius`, which according to Can I Use are supported natively (unprefixed) in the browsers that PDF.js 2.0 supports. Therefore, we can remove this class, which should help performance too since this avoids extra function calls in parts of the code that are called often.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -14,8 +14,8 @@
  */
 
 import {
-  addLinkAttributes, CustomStyle, DOMSVGFactory, getDefaultSetting,
-  getFilenameFromUrl, LinkTarget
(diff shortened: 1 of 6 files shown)


Snuffleupagus (approved):
Really nice find!
Over the years I've wondered if we actually needed that abstraction, and it's great to finally get rid of it now.

> [...] which according to Can I Use are supported natively (unprefixed) in the browsers that PDF.js 2.0 supports.

*Just for brevity:* It appears that especially the non-prefixed `transform` puts e.g. Safari compatibility at version 9 and above, which should be totally fine given https://en.wikipedia.org/wiki/Safari_version_history.

Thank you for the patch!

PR #9331, opened 2018-01-02
local num used instead of global pageNum (#9328)

### LINKED ISSUE #9328: examples/learning/prevnext.html use local num instead of global pageNum
This is just a logic enhancement in the `examples/learning/prevnext.html` code, as well as the JSFiddle gh-pages branch corresponding to same file.

So:

https://github.com/mozilla/pdf.js/blob/9e0559264ce8a782b5a82c6e479e6c3903cdb2cc/examples/learning/prevnext.html#L85

changed to

```nodejs
document.getElementById('page_num').textContent = num;
```

There `pageNum` is used, but since `renderPage` received a `num` parameter, that could be used instead of the outside variable. I guess it doesn't make much of a difference anyway, since `pageNum` is kept in sync all the time, but then `renderPage` could just as easily use `pageNum` from outside instead of receiving a parameter.

It's definitely a `5-good-beginner-bug` tag. It's clear to me that a PR branched off master could modify `examples/learning/prevnext.html` file, but I'm not clear how to update JSFiddle, since the url from [gh-pages has `/pdfjs/` in it](https://github.com/mozilla/pdf.js/blob/gh-pages/examples/index.html#L87), so who has permission could update that jsfiddle directly w/o needing to change the link in github pages? Or regenerate it so it keeps `pdfjs` in url `//jsfiddle.net/pdfjs/<newcode>`.

himanish-star:
fixes: #9328 

I have made the changes as desired. Am I also supposed to do something on the gh-pages branch too ? As per the directions mentioned in the issue I guess I don't have to do anything with the gh-pages branch. Please tell me if anything is required to be done.

### DIFF
--- a/examples/learning/prevnext.html
+++ b/examples/learning/prevnext.html
@@ -82,7 +82,7 @@
     });
 
     // Update page counters
-    document.getElementById('page_num').textContent = pageNum;
+    document.getElementById('page_num').textContent = num;
   }
 
   /**


Snuffleupagus:
> Am I also supposed to do something on the gh-pages branch too ? As per the directions mentioned in the issue I guess I don't have to do anything with the gh-pages branch.

The `gh-pages` branch should *never* be updated manually, since that's done automatically by a bot when code is merged.
*However*, depending on the outcome of these questions from https://github.com/mozilla/pdf.js/issues/9328#issue-285218143:
> [...] but I'm not clear how to update JSFiddle,
> Or regenerate it so it keeps pdfjs in url //jsfiddle.net/pdfjs/<newcode>

this PR may (or may not) need to *also* update the URL in https://github.com/mozilla/pdf.js/blob/master/docs/contents/examples/index.md#previousnext-example

@yurydelendik Can you please update [this JSFiddle](jsfiddle.net/pdfjs/wagvs9Lf/embed/js,html,result), or generate a new one, with the change in this PR present?

Snuffleupagus (changes requested):
Besides https://github.com/mozilla/pdf.js/pull/9331#issuecomment-354742681, this patch also needs a slightly better commit message. I.e. one that actually mentions that the change concerns the "prevnext" example, since otherwise it's impossible to tell what the patch does without looking at the code.

himanish-star:
@Snuffleupagus , I have edited the commit message

yurydelendik:
> @yurydelendik Can you please update this JSFiddle, or generate a new one, with the change in this PR present?

Updated affected line.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this! The JSFiddle is updated, so no further code changes are necessary for that.

PR #9336, opened 2018-01-03
Correctly extract component data from "Image and tile size" (SIZ) markers in JPEG 2000 images

Snuffleupagus:
This is something that I noticed while attempting to debug https://bugzilla.mozilla.org/show_bug.cgi?id=1374945.
Just looking at the code, the `YRsiz` parameter seemed immediately wrong and the fact that every component used the *same* data also looked strange.
Comparing with the specification, see https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-T.800-200208-S!!PDF-E&type=items#page=37, confirmed that this is indeed incorrect.

Note that I haven't got any example of a PDF file that is fixed by this patch, but that might be more luck than anything else. Manually checking a couple of files with included JPEG 2000 images, the `Csiz`/`XRsiz`/`YRsiz` parameters were `1` which could explain why this hasn't been an issue before.

Obviously we shouldn't generally make changes to `core` code without adding tests, but in this case I'm simply not sure how to obtain/create one. However, since the existing code doesn't make sense this patch could hopefully be deemed acceptable anyway.

### DIFF
--- a/src/core/jpx.js
+++ b/src/core/jpx.js
@@ -185,8 +185,9 @@ var JpxImage = (function JpxImageClosure() {
                   precision: (data[j] & 0x7F) + 1,
                   isSigned: !!(data[j] & 0x80),
                   XRsiz: data[j + 1],
-                  YRsiz: data[j + 1],
+                  YRsiz: data[j + 2],
                 };
+                j += 3;
                 calculateComponentDimensions(component, siz);
                 components.push(component);
               }


timvandermeij (approved):

timvandermeij:
I completely agree with the reasoning here. Even though it's unfortunate that there is no test file, the current code is indeed wrong. Thank you for spotting this!

Issue #9279, opened 2017-12-14
PDF rendered using pdf.js shows square like garbage


subhashlg26:
[document_Redacted-rev-6751.pdf](https://github.com/mozilla/pdf.js/files/1560105/document_Redacted-rev-6751.pdf)
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version: Chrome 62.0.3202.94 (Official Build) (64-bit)
- Operating system and its version: Mac 10.12.6 (16G29)
- PDF.js version: pdfjs-1.9.426-dist
- Is a browser extension: No

Steps to reproduce the problem:
1. Use latest version pdf.js to render attached document.

What is the expected behavior? 
Text should have been displayed as it is directly opened from chrome.
![image](https://user-images.githubusercontent.com/657554/34004418-d9b3acce-e11d-11e7-929d-e7e2ea5a700f.png)

What went wrong? (add screenshot)
Square like garbage is displayed
![image](https://user-images.githubusercontent.com/657554/34004462-f7745894-e11d-11e7-86a6-8b0b85aa502d.png)

Snuffleupagus:
Possibly a duplicate of issue #9278.

THausherr:
"Warning: Ran out of space in font private use area."

Here's a reduced file with the same problem
[PDFJS-9279-reduced.pdf](https://github.com/mozilla/pdf.js/files/1562314/PDFJS-9279-reduced.pdf)

timvandermeij:
Fixed by #9340.

Issue #9278, opened 2017-12-14
Google Noto Serif CJK JP font is not rendered properly in Mozilla firefox


arunbhati:
Attach (recommended) or Link to PDF file here: [FirefoxBugDocument.pdf](https://github.com/mozilla/pdf.js/files/1559178/FirefoxBugDocument.pdf)

Configuration:
- Web browser and its version:
 Firefox (57.0.1)
- Operating system and its version:
 Mac OS X (52.4.1)
- PDF.js version:
 1.6.315
- Is a browser extension:
 No

Steps to reproduce the problem:
1. Open attached pdf (FirefoxBugDocument.pdf) in firefox browser and This pdf file won't render properly.

What is the expected behavior? (add screenshot)
It should show "This text should be in bold" text in **bold**.
<img width="896" alt="properlyrenderedpdf" src="https://user-images.githubusercontent.com/6111411/33993026-dab657b6-e0fa-11e7-8961-f6d2f6cba66e.png">

What went wrong? (add screenshot)
Firefox console throws exception while rendering PDF.
```
downloadable font: OS/2: first char index 65535 > last char index 65534 in os2 (font-family: "g_d0_f1" style:normal weight:normal stretch:normal src index:0) source:   document.pdf

downloadable font: rejected by sanitizer (font-family: "g_d0_f1" style:normal weight:normal stretch:normal src index:0) source:   document.pdf

downloadable font: font load failed (font-family: "g_d0_f1" style:normal weight:normal stretch:normal src index:0) source:  document.pdf
```

timvandermeij:
Fixed by #9340.

Issue #8795, opened 2017-08-19
Font rendering broken due to overfull PUA


timvandermeij:
The font rendering for http://doppelbauer.name/core-dump.pdf is broken. The console message `Warning: Ran out of space in font private use area.` is printed over 4200 times and sanitizer errors appear.

DemiMarie:
May I ask what you mean by “sanitizer”?

If you mean ASAN, MSAN or the like, then that is a potential security bug in the host application.

Snuffleupagus:
> May I ask what you mean by “sanitizer”?

OpenType Sanitizer, see https://github.com/khaledhosny/ots/.

> then that is a potential security bug in the host application.

Not here, since fonts rejected by OTS (in Firefox or Chrome) are dropped and thus *not* passed on to the OS.

DemiMarie:
@Snuffleupagus That’s a borked PDF then.  I suggest closing as Not a Bug.

Snuffleupagus:
> That’s a borked PDF then.

Based on this particular PDF.js warning message, you simply *cannot* conclude that the PDF file *itself* is broken without any further analysis.
In this case, it's *much* more likely to be caused by limitations in the (current) PDF.js font parsing code.

> I suggest closing as Not a Bug.

Considering that other PDF viewers have no trouble rendering this file, that suggestions does *not* seem like a good course of action in this case! Or in general really, since most PDF viewers (including PDF.js) will try to render even corrupt PDF files (which I don't think we're dealing with here, btw).

timvandermeij:
Fixed by #9340.

Issue #8234, opened 2017-04-04
Letter ä not showing in pdf


pchtri:
Link to PDF file (or attach file here):
[FF52_CR_ä_missing.pdf](https://github.com/mozilla/pdf.js/files/892611/FF52_CR_a_missing.pdf)

Configuration:
- Web browser and its version: FireFox 52.0.1
- Operating system and its version: Windows 7 and 10
- PDF.js version: 1.7.406
- Is an extension:

Steps to reproduce the problem:
1. Open the attached file in https://mozilla.github.io/pdf.js/web/viewer.html or in FireFox 52.0.1
2.

What is the expected behavior? (add screenshot)
![image](https://cloud.githubusercontent.com/assets/26895763/24650700/a73256a4-1933-11e7-853c-529ebf4f31c3.png)
The same pdf works fine in Acrobat Reader, Chrome and all other viewers I've tested.

What went wrong? (add screenshot)
![image](https://cloud.githubusercontent.com/assets/26895763/24650791/09272272-1934-11e7-9d0f-f4d83b6e36bb.png)
The small letter ä is shown in the same color as the background (white). If you select the text and paste it to notepad ä is there, but it doesnt show in the pdf.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):
https://mozilla.github.io/pdf.js/web/viewer.html

THausherr:
[PDFJS-8234_reduced.pdf](https://github.com/mozilla/pdf.js/files/900112/PDFJS-8234_reduced.pdf)
Reduced version of the file. The font does not have does not have a glyph name for code 32, but it has a ToUnicode entry (ä) and a glyph (ä). Likely a bug in the subsetter of the library that produced this file.

timvandermeij:
Fixed by #9340.

PR #9340, opened 2018-01-04
Map all glyphs to the private use area and duplicate the first glyph.

brendandahl:
This was all spurred from a discussion with Jonathan Kew about our issues glyphs when the are mapped to certain unicode values. In an ideal world we'd have a canvas API to do something like drawGlyphID(), but he agreed that for now it's best to always use the private use area.

...

There have been lots of problems with trying to map glyphs to their unicode
values. It's more reliable to just use the private use areas so the browser's
font renderer doesn't mess with the glyphs.

Using the private use area for all glyphs did highlight other issues that this
patch also had to fix:

  * small private use area - Previously, only the BMP private use area was used
    which can't map many glyphs. Now, the (much bigger) PUP 16 area can also be
    used.

  * glyph zero not shown - Browsers will not use the glyph from a font if it is
    glyph id = 0. This issue was less prevalent when we mapped to unicode values
    since the fallback font would be used. However, when using the private use
    area, the glyph would not be drawn at all. This is illustrated in one of the
    current test cases (issue #8234) where there's an "ä" glyph at position
    zero. The PDF looked like it rendered correctly, but it was actually not
    using the glyph from the font. To properly show the first glyph it is always
    duplicated and appended to the glyphs and the maps are adjusted.

  * supplementary characters - The private use area PUP 16 is 4 bytes, so
    String.fromCodePoint must be used where we previously used
    String.fromCharCode. This is actually an issue that should have been fixed
    regardless of this patch.

Fixes:
- [bug 1425312](https://bugzilla.mozilla.org/show_bug.cgi?id=1425312)
- issue #9279 
- issue #8795 
- issue #9278
- issue #8234 This is already in our test suite, but the glyph is now the correct ‘ä’ (compare to acrobat)


brendandahl:
Looks like I have some issues on Linux (besides slight glyph movement).

Snuffleupagus:
> Looks like I have some issues on Linux (besides slight glyph movement).

You're probably already aware of it, but even on Windows there appears to be some issues still. Note that in both `issue1049` and `bug1354114` the previously undefined glyphs are replaced by actual glyphs, but unfortunately not the correct ones (so it's better, but still not quite right).

src/core/fonts.js
```diff
@@ -2279,13 +2208,13 @@ var Font = (function FontClosure() {
       }
 
       var dupFirstEntry = false;
```
Snuffleupagus: Unless I'm mistaken, this can be removed now since it's no longer used for anything.

src/core/fonts.js
```diff
@@ -37,10 +37,11 @@ import { FontRendererFactory } from './font_renderer';
 import { Stream } from './stream';
 import { Type1Parser } from './type1_parser';
 
-// Unicode Private Use Area
-var PRIVATE_USE_OFFSET_START = 0xE000;
-var PRIVATE_USE_OFFSET_END = 0xF8FF;
-var SKIP_PRIVATE_USE_RANGE_F000_TO_F01F = false;
+ // Unicode Private Use Areas:
+var PRIVATE_USE_AREAS = [
```
Snuffleupagus: Nit: Use `const` instead here.

src/core/cff_parser.js
```diff
@@ -885,6 +877,30 @@ var CFF = (function CFFClosure() {
 
     this.isCIDFont = false;
   }
+  CFF.prototype = {
+    duplicateFirstGlyph: function CFF_duplicateFirstGlyph() {
+      // Browsers will not display a glyph at position 0. Typically glyph 0 is
+      // notdef, but a number of fonts put a valid glyph so it must be
```
Snuffleupagus: Grammar nit: Shouldn't it perhaps be a "there" after "... valid glyph", i.e.
`// notdef, but a number of fonts put a valid glyph there so it must be`?

src/core/cff_parser.js
```diff
@@ -1556,7 +1583,43 @@ var CFFCompiler = (function CFFCompilerClosure() {
...
+        case 3:
+          let start = 0;
+          let lastFD = fdSelect.fdSelect[0];
+          let ranges = [
+            format,
+            0, // nRanges place holder
+            0, // nRanges place holder
+            (start >> 8) & 0xFF,
+            start & 0xFF,
+            lastFD];
```
Snuffleupagus: Nit: For improved readability, the `];` should be moved to the next line instead.

Snuffleupagus:
In addition to the issues/bugs listed in https://github.com/mozilla/pdf.js/pull/9340#issue-286137592, the following ones also appear to be fixed:
 - Issue #5377 (almost, there's just *one* missing glyph at the start now)
 - Issue #5476
 - Issue #6397
 - Issue #8680

test/pdfs/bug1425312.pdf.link
```diff
@@ -0,0 +1 @@
+https://www.unicode.org/charts/PDF/U11A50.pdf
```
Snuffleupagus: How about linking to the Internet Archive here, to avoid the file becoming unavailable, i.e. using the following link instead https://web.archive.org/web/20171116124714/https://www.unicode.org/charts/PDF/U11A50.pdf.

test/test_manifest.json
```diff
@@ -3327,6 +3339,12 @@
       "rounds": 1,
       "type": "eq"
     },
+    {  "id": "bug1425312",
```
Snuffleupagus: This entry needs `"link": true,` added, to fix the following from the test logs:
```
WARNING: Unable to open file for reading "Error: ENOENT: no such file or directory, open 'pdfs/bug1425312.pdf'".
Unable to verify the checksum for the files that are used for testing.
Please re-download the files, or adjust the MD5 checksum in the manifest for the files listed above.
```

Also, it'd probably suffice to only test the second page here.

brendandahl:
The above changes don't fix everything, just the windows issues and nits. I'm still looking into the linux issues.

brendandahl:
For reference, the code that causes glyph 0 to not be mapped/drawn: 
https://searchfox.org/mozilla-central/rev/7fb999d1d39418fd331284fab909df076b967ac6/gfx/thebes/gfxFontUtils.cpp#207

brendandahl:
Small update and some troubles: 

Issue #2881 was failing on linux because the platform was rejecting the font (OTS was happy). After downloading and building a debug build of freetype, I used ftlint to find the font was being rejected because the number of charstrings was greater than the number of charset values. I then wrote a function to write out an updated cff charset (which fixed the issue). However, after this, endchar.pdf started failing. This is because it uses the `enchchar\seac` operator which requires the charset to be set to a predefined charset, whereas we were now writing out a custom format 0 charset.

So far I don't have a great plan to handle fixing both issues. Some possible ideas though:

- Don't duplicate glyph 0, but instead use our js font renderer if we're trying to draw glyph 0. Cons: a few glyphs may look different relative to others.
- Move glyph 0 to it's own font. Cons: lots of work to move a glyph and all it supporting stuff and we'd have to add another path to change the font in show text.
- Merge endchar/seac into a single glyph. Not entirely sure if this is possible, but it may be the cleanest situation.

I'm open to ideas though @Snuffleupagus @yurydelendik

brendandahl:
Try that again with the correct changes...
/botio test

brendandahl:
Always do seac analysis for type1 fonts too...
/botio test

brendandahl:
@Snuffleupagus @yurydelendik This is ready for review.

Snuffleupagus (commented):
Unfortunately I'm slightly out of my depth, and don't really have time, to do an exhaustive review here; sorry about that!
However, I've read through the patch carefully and the general approach looks really good to me. Re-mapping all glyphs to PUAs is a great idea, since it ought to help prevent future bugs too.

Besides a couple of small comments/questions below, please remember to fold the `To Squash: ...` commit into its parent commit :-)

src/core/fonts.js
```diff
@@ -2883,12 +2823,14 @@ var Font = (function FontClosure() {
         var seac = this.seacMap[charcode];
         fontCharCode = seac.baseFontCharCode;
         accent = {
-          fontChar: String.fromCharCode(seac.accentFontCharCode),
+          fontChar: String.fromCodePoint(seac.accentFontCharCode),
```
Snuffleupagus: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint#Browser_compatibility this (unsurprisingly) isn't supported by IE11.

Please include the necessary `string/from-code-point` polyfill, using [core-js](https://github.com/zloirock/core-js#ecmascript-string-and-regexp), in the `src/shared/compatibility.js` file.

src/core/font_renderer.js
```diff
@@ -122,7 +122,7 @@ var FontRendererFactory = (function FontRendererFactoryClosure() {
   }
 
   function lookupCmap(ranges, unicode) {
-    var code = unicode.charCodeAt(0), gid = 0;
+    var code = unicode.codePointAt(0), gid = 0;
```
Snuffleupagus: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt#Browser_compatibility this (unsurprisingly) isn't supported by IE11.

Please include the necessary `string/code-point-at` polyfill, using [core-js](https://github.com/zloirock/core-js#ecmascript-string-and-regexp), in the `src/shared/compatibility.js` file.

src/core/fonts.js
```diff
@@ -1905,8 +1840,6 @@ var Font = (function FontClosure() {
           glyf.data.set(newGlyfData.subarray(0, firstEntryLength), writeOffset);
           itemEncode(loca.data, locaData.length - itemSize,
                      writeOffset + firstEntryLength);
-        } else {
-          glyf.data = newGlyfData.subarray(0, writeOffset);
```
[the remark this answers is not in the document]
brendandahl: I've waited too long to look at this again and now I'm not exactly sure. I think at some point I was going to get rid of the dupFirstEntry option since it is pretty much always true unless there's not room to duplicate the glyph. Let me try adding it back.

Snuffleupagus:
Following-up on https://github.com/mozilla/pdf.js/pull/9340#pullrequestreview-142513504: I'm not thinking that given the very large number of issues/bugs that this PR fixes (or significantly improves), that it probably makes the most sense to simply land this (*after* squashing commits) and deal with any remaining issues in future PRs instead.

Snuffleupagus:
Following-up on https://github.com/mozilla/pdf.js/pull/9340#pullrequestreview-142513504: I'm now thinking that given the very large number of issues/bugs that this PR fixes (or significantly improves), that it probably makes the most sense to simply land this (*after* squashing commits) and deal with any remaining issues in future PRs instead.

test/unit/jasmine-boot.js
```diff
@@ -60,7 +60,6 @@ function initializePDFJS(callback) {
     'pdfjs-test/unit/dom_utils_spec',
     'pdfjs-test/unit/encodings_spec',
     'pdfjs-test/unit/evaluator_spec',
-    'pdfjs-test/unit/fonts_spec',
```
Snuffleupagus: I forgot to mention it earlier, but you'll also need to remove the following line: https://github.com/mozilla/pdf.js/blob/436d2efa8a89edef3a05f70592e3a70fd0c7d11e/test/unit/clitests.json#L21

timvandermeij:
Given the number of issues that this pull request fixes, I fully agree that we should land this now that multiple people looked at it (I also looked at the code and the reference test differences) and it also introduces more test coverage. It would also be great to have this for version 2.0.

/botio makeref

timvandermeij:
Nice work!

PR #9343, opened 2018-01-06
Restore the `btoa`/`atob` polyfills for Node.js

Snuffleupagus:
These were removed in PR #9170, since they were unused in the browsers that we'll support in PDF.js version `2.0`.
However looking at the output of Travis, where a subset of the unit-tests are run using Node.js, there's warnings about `btoa` being undefined. This doesn't appear to cause any errors, which probably explains why we didn't notice this before (despite PR #9201).

*Edit:* See e.g. the previous Travis run https://travis-ci.org/mozilla/pdf.js/builds/325585865#L600

### DIFF
--- a/examples/node/domstubs.js
+++ b/examples/node/domstubs.js
@@ -197,10 +197,6 @@ DOMElementSerializer.prototype = {
   },
 };
 
-function btoa (chars) {
-  return Buffer.from(chars, 'binary').toString('base64');
-}
-
 const document = {
   childNodes : [],
 
@@ -245,7 +241,6 @@ Image.prototype = {
   }
 }
 
-exports.btoa = btoa;
 exports.document = document;
 exports.Image = Image;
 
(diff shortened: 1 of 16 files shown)


timvandermeij:
I don't really understand why we need all the `isNodeJS() -> isNodeJS` changes and why it has to live in a separate file. Why can't we just use the one in `src/shared/util` anymore?

Snuffleupagus:
> I don't really understand why we need all the isNodeJS() -> isNodeJS changes and why it has to live in a separate file. Why can't we just use the one in src/shared/util anymore?

Because it seemed nice to be able to re-use it in `compatibility.js`, and since that file imported by `src/shared/util.js` trying to access the existing `isNodeJS` function from `compatibility.js` would cause a circular reference. Hence I figured that following the `globalScope` pattern could be a reasonable solution, but I'm certainly open to ideas here!

timvandermeij (approved):

timvandermeij:
Thank you for fixing this! I forgot about this patch after the last comment; sorry for that. Your solution is clear to me now after the explanation and I do like the re-use here.

Issue #8263, opened 2017-04-10
Don't guess file name from non-http/file URLs


yurydelendik:
From e-mail:

> pdf.js tries to find a nice filename for the document. this takes an incredible amount of time when a data-uri document is displayed

data: or blob: schemas doesn't any useful info to guess title, so viewer.js:1709-1710 / getPDFFileNameFromURL may skip regex if non-http/file URL is used.

Snuffleupagus:
> data: or blob: schemas doesn't any useful info to guess title, so viewer.js:1709-1710 / getPDFFileNameFromURL may skip regex if non-http/file URL is used.

Wouldn't that basically break the use-case that PR #8024 added, since it depends on the filename being appended to a `blob:` URL for e.g. the download prompt to be able to display the correct filename, i.e. does this mean that we need to back out that PR?

yurydelendik:
> Wouldn't that basically break the use-case that PR #8024 added

I'm sorry, but I'm not quite following this. I'm talking about use in https://github.com/mozilla/pdf.js/blob/31f88756148475440c94c3cc2ad5da25139b101b/web/app.js#L578 and https://github.com/mozilla/pdf.js/blob/31f88756148475440c94c3cc2ad5da25139b101b/web/app.js#L578. If there is a problem with logic in #8024, can we improve the attachment opening in some way or reduce the proposed above logic to not exclude such attachment URLs?

Snuffleupagus:
Looking at https://github.com/mozilla/pdf.js/blob/d6dfc26d4e578ce11ecb28132f166022edb3439f/web/pdf_attachment_viewer.js#L112, which for PDF attachments (in FIREFOX/MOZCENTRAL builds) is then used by `PDFViewerApplication.download` to determine the file name at https://github.com/mozilla/pdf.js/blob/d6dfc26d4e578ce11ecb28132f166022edb3439f/web/app.js#L751.

It might not be an issue, but I simply wanted to mention that particular use-case, such that we don't inadvertently break the file name detection there.

Snuffleupagus:
Unless I'm mistaken, this problem described in the first post *mainly* concerns `data:` URLs, since they can be arbitrarily long. For a `blob:` URL on the other hand, since the identifier part has a fixed length, it shouldn't be any worse than the length of e.g. a `http:` or `file:` URL.

[...]

PR #9345, opened 2018-01-08
Download button is now hidden for PDFs which are loaded from 'file://'

### LINKED ISSUE #9342: Hide Download Button when loading PDF from 'file://'
Configuration:
- Web browser and its version: Firefox 57
- PDF.js version: 1.9.583
- Is a browser extension: Yes

1. Open a local PDF file with Firefox

Hidden Download Button as the file I am looking at is already on my computer.

What went wrong?
Download Button is visible.

himanish-star:
fixes : #9342 

@timvandermeij , I have tried to solve this but I'm not sure if this is the best possible solution. I used `gulp firefox` to build a firefox extension. I can see my changes in the `build/firefox/content/web/viewer.js` but I tried a lot to also install this extension into my firefox browser using `pdf.js.xpi` file but it doesn't get installed as it is not compatible with the newer edition. Can you suggest me how to check this.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -15,9 +15,10 @@
 /* globals PDFBug, Stats */
 
 import {
-  animationStarted, DEFAULT_SCALE_VALUE, getPDFFileNameFromURL, isValidRotation,
-  MAX_SCALE, MIN_SCALE, noContextMenuHandler, normalizeWheelEventDelta,
-  parseQueryString, PresentationModeState, ProgressBar, RendererType
+  animationStarted, DEFAULT_SCALE_VALUE, getPDFFileNameFromURL, isFileSchema,
+  isValidRotation, MAX_SCALE, MIN_SCALE, noContextMenuHandler,
+  normalizeWheelEventDelta, parseQueryString, PresentationModeState,
+  ProgressBar, RendererType
 } from './ui_utils';
 import {
   build, createBlob, getDocument, getFilenameFromUrl, InvalidPDFException,
@@ -675,6 +676,8 @@ let PDFViewerApplication = {
     this.store = null;
     this.isInitialViewSet = false;
     this.downloadComplete = false;
+    this.url = '';
+    this.baseUrl = '';
 
     this.pdfSidebar.reset();
     this.pdfOutlineViewer.reset();
@@ -735,6 +738,12 @@ let PDFViewerApplication = {
         }
         parameters[prop] = args[prop];
       }
+    }
+
+    if (this.url && isFileSchema(this.url)) {
+      let appConfig = this.appConfig;
+      appConfig.toolbar.download.setAttribute('hidden', 'true');
+      appConfig.secondaryToolbar.downloadButton.setAttribute('hidden', 'true');
     }
 
     let loadingTask = getDocument(parameters);
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -381,6 +381,14 @@ function noContextMenuHandler(evt) {
   evt.preventDefault();
 }
 
+function isFileSchema(url) {
+  let i = 0, ii = url.length;
+  while (i < ii && url[i].trim() === '') {
+    i++;
+  }
+  return url.substr(i, 7).toLowerCase() === 'file://';
+}
+
 function isDataSchema(url) {
   let i = 0, ii = url.length;
   while (i < ii && url[i].trim() === '') {
@@ -665,6 +673,7 @@ export {
   SCROLLBAR_PADDING,
   VERTICAL_PADDING,
   isValidRotation,
+  isFileSchema,
   cloneObj,
   PresentationModeState,
   RendererType,


web/viewer.js
```diff
@@ -177,6 +177,11 @@ function getViewerConfiguration() {
 
 function webViewerLoad() {
   let config = getViewerConfiguration();
+  let locationHref = window.location.href;
+  let downloadButton = config.toolbar.download;
```
Hirse: Isn't there also a `secondaryToolbar` for the overflow menu?

web/viewer.js
```diff
@@ -177,6 +177,11 @@ function getViewerConfiguration() {
...
+  if (locationHref.startsWith('file://')) {
```
Hirse: This is checking the url, which is correct for my use-case, but has side effects when serving the `viewer.html` itself from `file`.

web/viewer.js
himanish-star: Yeah you're right.  I'll fix that too once I get a confirmation that the changes I have made is correct

web/viewer.js
himanish-star: OK,  I will look into this also once I'm back from college

Snuffleupagus: Furthermore you cannot use `String.prototype.startsWith()` without a polyfill since it's not properly supported across browsers, as shown in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith#Browser_compatibility.

Snuffleupagus (changes requested):
As far as I'm concerned the issue you're attempting to solve here is very much an edge-case, considering that at least Firefox will prevent linking to `file://` URLs and that it's thus limited to files opened manually by the user, hence I'm just not sure if it's necessary to fix this!?

One thing that we'll need to be really careful about here is how the URL is matched, such that we won't accidentally re-introduce issue #8263 in a different form.
That said, if other reviewers think that we *should* do this, I'd be fine with it provided that the implementation is correct.
(As already pointed out in an inline comment, this isn't currently checking the URL of the PDF file, but rather the URL of the viewer itself which simply cannot work. And note that the file you'd need to edit is actually `web/app.js`.)

himanish-star:
@Snuffleupagus , Ok I will look into this

web/viewer.js
himanish-star: @Hirse , Could you please tell me all the possible cases of rendering `pdf.js`. Actually right now I'm trying to deeply understand the code. So if you tell me all the rendering cases and how can I try out those cases that would help me to understand the code better. By rendering cases what I mean is 

- opening viewer.html manually
- firefox extension. @Snuffleupagus , How do I test the firefox extension generated by me using `gulp firefox`
- what does the generic build do

Hirse: @himanish-star I don't know all possibilities, but you don't need all.
It is safe to assume that using `window.location` is not the best solution. You can look at the code where the PDF is loaded and check its URL there.

himanish-star: @Hirse , thanks. I may take few more days but I'll surely try to fix this.

himanish-star:
After proceeding further I can see that `PDFJSDev` is used in a number of places. So is this basically used to partition code for different builds, as I can see the `if` conditions and they show comparisons like `MOZCENTRAL`/`FIREFOX`/`CHROME`. So assuming this thing to be true , I have a question
https://github.com/mozilla/pdf.js/blob/5a52ee0a795ca41f27ede587dc494cd304dc81e1/web/app.js#L1610

The above declaration of the function `webViewerOpenFileViaURL` is thus considered only in the generic build. So in the generic build file is opened manually by the user. Then I believe that no URLs would start with `'file:'`. Then why this check
https://github.com/mozilla/pdf.js/blob/5a52ee0a795ca41f27ede587dc494cd304dc81e1/web/app.js#L1611

himanish-star:
@Snuffleupagus , I have made some changes now do you think they are correct

Snuffleupagus:
As outlined in https://github.com/mozilla/pdf.js/pull/9345#pullrequestreview-87163803, I'm not really sure that we need/should fix this issue.
@timvandermeij What's your opinion here?

> I have made some changes now do you think they are correct

Unfortunately, I don't think so. Provided that we actually want to fix this, in order to cover all possible cases you probably want to do something along the following lines instead.

 - Add a `isFileSchema` helper function similar to
https://github.com/mozilla/pdf.js/blob/ba0a3aebd05fe85f5eef42260b1346d5856b6ba1/web/ui_utils.js#L384

 - Around line [web/app.js#L739](https://github.com/mozilla/pdf.js/blob/ba0a3aebd05fe85f5eef42260b1346d5856b6ba1/web/app.js#L739) do something like this:
```javascript
if (this.url && isFileSchema(this.url)) {
  // Hide buttons...
```

> This is occurring because of the polyfill. Then what should I do ? Shall I implement this polyfill in another way

In this case, a solution not requiring a polyfill would be a much better choice here. (Furthermore, for future reference, please note that polyfills aren't allowing in regular code but should go in https://github.com/mozilla/pdf.js/blob/master/src/shared/compatibility.js instead.)

---

Regarding the installing of unsigned Firefox addons, you'll need to use the Nightly version of Firefox and then set the `xpinstall.signatures.required` preference to `false` in [about:config](https://support.mozilla.org/kb/about-config-editor-firefox).

himanish-star:
@Snuffleupagus , thanks for the help. I'll make the changes requested.  I'll wait for your confirmation of fixing this issue.

timvandermeij:
I think it would be nice to have it fixed, but only if the fix is small, so not more than a few lines of code, preferably without any polyfills. If strictly necessary, use one from `core-js` like we already do for other polyfills (there is an ES6 string `startsWith` polyfill available).

himanish-star:
Yeah sure @timvandermeij

himanish-star:
Is there any other way of testing the extensions  as this doesn't work
> Regarding the installing of unsigned Firefox addons, you'll need to use the Nightly version of Firefox and then set the xpinstall.signatures.required preference to false in about:config.

himanish-star:
@Snuffleupagus , I have made the changes, please have a look

web/ui_utils.js
```diff
@@ -381,6 +381,15 @@ function noContextMenuHandler(evt) {
   evt.preventDefault();
 }
 
+function isFileSchema(url) {
+  if (!url.__proto__.startsWith) {
```
timvandermeij: Since this is polyfill code, it shouldn't be here but in `src/shared/compatibility.js` instead. It should be in the same format as https://github.com/mozilla/pdf.js/blob/master/src/shared/compatibility.js#L189 with the `core-js/fn/string/starts-with` function (see https://github.com/zloirock/core-js/blob/master/fn/string/starts-with.js).

himanish-star: @timvandermeij , sure I'll make the changes

himanish-star: @timvandermeij , please have a look now.

web/ui_utils.js
```diff
@@ -381,6 +381,10 @@ function noContextMenuHandler(evt) {
   evt.preventDefault();
 }
 
+function isFileSchema(url) {
+  return url.startsWith('file://');
```
Snuffleupagus: I'm not sure if you actually read *all* of https://github.com/mozilla/pdf.js/pull/9345#issuecomment-357431768, so please take a closer look at it! Quoting from it:
> Add a `isFileSchema` helper function similar to https://github.com/mozilla/pdf.js/blob/ba0a3aebd05fe85f5eef42260b1346d5856b6ba1/web/ui_utils.js#L384
> [...]
> 
> In this case, a solution not requiring a polyfill would be a much better choice here.

So, please just follow the *exact same* pattern as the `isDataSchema` function just below for this helper function and remove the polyfill.

src/shared/compatibility.js
```diff
@@ -172,6 +172,13 @@ PDFJS.compatibilityChecked = true;
   Number.isInteger = require('core-js/fn/number/is-integer');
 })();
 
+(function checkStringStartsWith() {
```
Snuffleupagus: As mentioned in https://github.com/mozilla/pdf.js/pull/9345#issuecomment-357431768, please implement a solution not requiring a polyfill and remove any changes to this file.

(Also, the polyfill is wrong since it should've been `String.prototype.startsWith`, but that's beside the point here since the polyfill isn't necessary.)

web/app.js
```diff
@@ -737,6 +738,12 @@ let PDFViewerApplication = {
       }
     }
 
+    if (this.url && isFileSchema(this.url)) {
+      let appConfig = PDFViewerApplication.appConfig;
```
Snuffleupagus: Why `PDFViewerApplication.appConfig` here? You could, and even should, simply use `this.appConfig` here.

Snuffleupagus (changes requested):
> Is there any other way of testing the extensions as this doesn't work
> > Regarding the installing of unsigned Firefox addons, you'll need to use the Nightly version of Firefox and then set the xpinstall.signatures.required preference to false in about:config.

Works for me (in Firefox Nightly), so without *detailed* information about what "this doesn't work" actually means we cannot suggest what might have gone wrong.

himanish-star:
>  so without detailed information about what "this doesn't work" actually means we cannot suggest what might have gone wrong.

I am very sorry on my part for not providing details. So I have made a GIF below which explains the procedure I have used to setup the extension.
simplescreenrecorder-2018-01-16_16 12 03

___________________________________________________________________________________
> As mentioned in #9345 (comment), please implement a solution not requiring a polyfill and remove any changes to this file.

Done
___________________________________________________________________________________
> Why PDFViewerApplication.appConfig here? You could, and even should, simply use this.appConfig here.

Done
___________________________________________________________________________________
> I'm not sure if you actually read all of #9345 (comment), so please take a closer look at it! 

Made a mistake while reading it. Very sorry again

Snuffleupagus:
> So I have made a GIF below which explains the procedure I have used to setup the extension.

Thanks for that; unfortunately I'm not sure what to suggest since those steps works for me on Windows (but I don't think that the OS should be relevant here).
The only thing that I did notice is that you don't actually get a "addon unsigned" error, but the addon rather seem to be rejected because of the [`maxVersion`](https://github.com/mozilla/pdf.js/blob/master/extensions/firefox/install.rdf#L17) being too low. Anyway, I'm out of ideas here :-(

---

A pre-existing issue, but somewhat more relevant now, is that the `this.url`/`this.baseUrl` properties on `PDFViewerApplication` aren't properly reset when closing a document.
Can you please reset them both to empty strings (which is how they're [initialized](https://github.com/mozilla/pdf.js/blob/f774abc8d335712fb14ebc6dc262eb389363607b/web/app.js#L152-L153)), just after line https://github.com/mozilla/pdf.js/blob/f774abc8d335712fb14ebc6dc262eb389363607b/web/app.js#L677

With that fixed, I suppose that we can try this patch.

himanish-star:
@Snuffleupagus , sure I'll make the changes requested

Snuffleupagus (approved):
This seems to work as expected now, so let's try this.
Thanks for the patch!

Snuffleupagus:
As evident from e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1450689 and https://mozilla.logbot.info/pdfjs/20180528#c14817333, the fact that the download button isn't available for local files is apparently causing some user confusion.

Hence we should probably consider simply backing out this PR, with the *exception* of [these lines](https://github.com/himanish-star/pdf.js/blob/06b3bb8214ee364c351dbd3e4e668e2ccb8cbad0/web/app.js#L679-L680) in `PDFViewerApplication.close()`.

@timvandermeij Could you help back this out?

---
[1] Note that in https://github.com/mozilla/pdf.js/pull/9345#pullrequestreview-87163803, I did express some reservation regarding whether or not PR was actually a good idea.

timvandermeij:
Yes, I'll take care of backing this out given the confusion I can understand.

PR #9349, opened 2018-01-09
Add composer.json to make easy for PHP developers keep track of pdf.js changes

fabiomontefuscolo:
Hello!

First of all, thanks for pdf.js, it is really amazing!

I'm proposing you a `composer.json` to be submitted to packagist.org, so any PHP developer using `composer` can keep track of pdf.js.

There are some additional step to submit pdf.js on packagist, but I can't do it. Only maintainers can do the steps described at https://packagist.org/about#how-to-update-packages. I think it takes no more than 10 minutes.

After that, packagist will automatically update package when you push or release changes on your repository.

Can you accept this pull request?

Thank you again,
Fabio Montefuscolo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,21 @@
 language: node_js
 node_js:
-  - node
+- node
 cache:
   directories:
-    - node_modules
+  - node_modules
 install:
-  - npm install -g npm@latest
-  - npm install -g gulp-cli
-  - npm install
-  - npm update
+- npm install -g npm@latest
+- npm install -g gulp-cli
+- npm install
+- npm update
+- gulp publish
+deploy:
+  provider: releases
+  api_key:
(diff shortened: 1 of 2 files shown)


yurydelendik:
This tasks need to be performed when pdfjs-dist is generated. By reading the https://packagist.org/about#how-to-update-packages, the PDF.js sources (along with repo) is not suitable to be published to the packagist. Please update the solution to:

- [ ] do file generate during build step
- [ ] keep only Mozilla Foundation in the authors list

We can add the service hook for pdfjs-dist later. https://github.com/mozilla/pdfjs-dist is already using proper version naming.

timvandermeij:
Closing since the change isn't in a mergable state at the moment and there has been no more follow-up. However, we can reconsider a new pull request if the suggestions from Yury are incorporated.

PR #9350, opened 2018-01-09
Implement `closeEOFillStroke` in SVG backend

janpe2:
That's the PDF operator `b*`.

Here's a reduced PDF:
[path_painting_ops.pdf](https://github.com/mozilla/pdf.js/files/1616137/path_painting_ops.pdf)


Snuffleupagus (approved):
Looks good to me, thank you for the patch!

PR #9352, opened 2018-01-10
Attempt to actually resolve ColourSpace names in accordance with the specification (issue 9285)

### LINKED ISSUE #9285: pdf completely rendered blank

Snuffleupagus:
Please refer to the PDF specification, in particular http://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G7.3801570

> A colour space shall be specified in one of two ways:
>  - Within a content stream, the CS or cs operator establishes the current colour space parameter in the graphics state. The operand shall always be name object, which either identifies one of the colour spaces that need no additional parameters (DeviceGray, DeviceRGB, DeviceCMYK, or some cases of Pattern) or shall be used as a key in the ColorSpace subdictionary of the current resource dictionary (see 7.8.3, "Resource Dictionaries"). In the latter case, the value of the dictionary entry in turn shall be a colour space array or name. A colour space array shall never be inline within a content stream.
>
> - Outside a content stream, certain objects, such as image XObjects, shall specify a colour space as an explicit parameter, often associated with the key ColorSpace. In this case, the colour space array or name shall always be defined directly as a PDF object, not by an entry in the ColorSpace resource subdictionary. This convention also applies when colour spaces are defined in terms of other colour spaces.

Fixes #9285.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -258,17 +258,7 @@ var ColorSpace = (function ColorSpaceClosure() {
     }
   };
 
-  ColorSpace.parseToIR = function(cs, xref, res, pdfFunctionFactory) {
-    if (isName(cs)) {
-      var colorSpaces = res.get('ColorSpace');
-      if (isDict(colorSpaces)) {
-        var refcs = colorSpaces.get(cs.name);
-        if (refcs) {
-          cs = refcs;
-        }
-      }
-    }
-
+  ColorSpace.parseToIR = function(cs, xref, res = null, pdfFunctionFactory) {
     cs = xref.fetchIfRef(cs);
     if (isName(cs)) {
       switch (cs.name) {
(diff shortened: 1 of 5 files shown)


brendandahl:
/botio-linux test
(looks like this may be a new intermittent)

brendandahl (approved):

Snuffleupagus:
> (looks like this may be a new intermittent)

It should be since that test-case is actually the one included in *this* PR, and it's also in the closed PR #9309. In that one, `makeref` was run *before* https://github.com/mozilla/pdf.js/pull/9309#issuecomment-356467898 which caused me to completely abandon that approach in favour of the current PR.
However, between closing PR #9309 and running tests in *this* one, the Linux bot was updated (see PR #9353), which should thus explain the movement here.

@brendandahl Thanks for the review!

PR #9356, opened 2018-01-11
Two small `PDFHistory` fixes, concerning the `this._maxUid` property and the 'pagehide' event

Snuffleupagus:
- Ensure that the `PDFHistory._maxUid` property is correctly updated when initializing/navigating the history
   
   This is a follow-up to commit e772124, in PR #8994, to cover a couple of cases missed there.

 - Allow overwriting temporary, in addition to empty, history entries on 'pagehide'

   When navigating away from the viewer, there's no good reason for disallowing replacement of a *temporary* history entry (in addition to empty ones).
   
   Given that the current position is temporarily added to the browser history using a (short) timeout, the history entry will most likely already be correct when the 'pagehide' event fires. However, if the user is quick enough that might not always be the case, in which case the adjusted logic may help.

### DIFF
--- a/web/pdf_history.js
+++ b/web/pdf_history.js
@@ -133,6 +133,9 @@ class PDFHistory {
     let destination = state.destination;
     this._updateInternalState(destination, state.uid,
                               /* removeTemporary = */ true);
+    if (this._uid > this._maxUid) {
+      this._maxUid = this._uid;
+    }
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #9362, opened 2018-01-13
Fix incorrect use of unofficial stream.close API

acchou:
When following the contributing instructions (https://github.com/mozilla/pdf.js/wiki/Contributing) the `gulp makeref` step causes the following error:

```
events.js:180
    throw new errors.TypeError('ERR_INVALID_ARG_TYPE', 'listener', 'Function');
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "listener" argument must be of type Function
    at _addListener (events.js:180:11)
    at WriteStream.addListener (events.js:240:10)
    at WriteStream.close (fs.js:2302:10)
    at WriteStream.<anonymous> (/Users/achou/Code/pdf.js/test/downloadutils.js:81:14)
    at WriteStream.emit (events.js:164:20)
    at finishMaybe (_stream_writable.js:605:14)
    at afterWrite (_stream_writable.js:456:3)
    at onwrite (_stream_writable.js:446:7)
    at fs.js:2246:5
    at FSReqWrap.wrapper [as oncomplete] (fs.js:703:5)
```

This may be Node-version specific (I'm using v9.2.1). In any case it appears the code in `downloadutils.js` calls `stream.close()`, when the correct API appears to be `stream.end()` (https://nodejs.org/api/stream.html#stream_writable_end_chunk_encoding_callback).

This pull request changes the API call and appears to solve the issue.

On a separate note, is there an official code formatter for pdf.js? I've noticed that when using vscode the default formatter deals with whitespace a little differently.


.gitignore
```diff
@@ -8,3 +8,4 @@ node_modules/
 examples/node/svgdump/
 examples/node/pdf2png/*.png
 package-lock.json
+.vscode
```
timvandermeij: Remove this since it's unrelated to the pull request. We decided not to include editor-specific entries in our `.gitignore` because it would be an endless list. Please add it to your editor's global igore list instead.

test/downloadutils.js
```diff
@@ -43,7 +43,7 @@ function downloadFile(file, url, callback, redirects) {
   protocol.get(url, function (response) {
     var redirectTo;
     if (response.statusCode === 301 || response.statusCode === 302 ||
-        response.statusCode === 307 || response.statusCode === 308) {
+      response.statusCode === 307 || response.statusCode === 308) {
```
timvandermeij: Revert all whitespace changes in this patch. We try to enforce as much as possible with `eslint`, but these are unfortunately not caught. The code style guide is at https://github.com/mozilla/pdf.js/wiki/Style-Guide.

timvandermeij (commented):
This patch looks good if only the API call change remains in the diff. Make sure to end up with one commit as well, so either squash the current commits after making the changes (see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits) or create a new commit with your change after first pulling the most recent master branch. Thank you!

acchou:
I'll submit a new pull request with the modifications requested.

PR #9363, opened 2018-01-13
Limit PDFFetchStream to http(s) in the Chrome extension

### LINKED ISSUE #9361: Cannot view local files in Chrome extension with fetch API
I have built the latest version of the extension, and tried to open a local PDF file. It won't open, the console shows the following error:

> pdf.js:8302 Fetch API cannot load file:///path/to/file.pdf. URL scheme must be "http" or "https" for CORS request.

To fix this, the Chrome extension must only use PDFFetchStream for http(s) schemes, and PDFNetworkStream for other schemes (file, ftp).

Rob--W:
The `fetch` API is only supported for http(s), even in Chrome extensions. Because of this limitation, we should use the `XMLHttpRequest` API when the requested URL is not a http(s) URL.

Fixes #9361

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -83,15 +83,25 @@ if (typeof PDFJSDev !== 'undefined' &&
   }) : null;
 }
 
-/** @type IPDFStream */
-var PDFNetworkStream;
+/**
+ * @typedef {function} IPDFStreamFactory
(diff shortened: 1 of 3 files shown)


Snuffleupagus (approved):
Looks good to me, thanks for the patch!

PR #9364, opened 2018-01-15
changed decodeURI to decodeURIComponent: Fixes #8987

### LINKED ISSUE #8987: ENOENT: no such file or directory, lstat with # encoded in filename
#8928 actually resolved the issue #8907 I open couple days ago. However, when filenames contain an encoded `#` character (`%23`), current decoding process do not handled it.

Configuration:

* Web browser and its version:
  Electron 1.7.8
* Operating system and its version:
   Fedora 25
* PDF.js version:
  1.9.617

1. Open a locale file with an `#` encoded character `file:///home/njakob/Foo%20Bar/document%23.pdf`

What went wrong?
* The error `ENOENT: no such file or directory, lstat` is raised.

Issue is related to the [`decodeURI`](https://github.com/mozilla/pdf.js/blob/e4c09c7cba9cb21db447e1e8d00051e97b40f7d8/src/display/node_stream.js#L363) usage which ignore decoding of `%23`.

Solution should also cover cases were the `#` is inside the path such as `file:///home/njakob/Foo%23/document.pdf`.

shikhar-scs:
pdf.js node extension now opens file containing the character `#` in the file name or the path of the file.

Read the following snippet from [here](https://stackoverflow.com/questions/747641/what-is-the-difference-between-decodeuricomponent-and-decodeuri) 
```
The main difference between `decodeURI` and `decodeURIComponent` is that:
     -decodeURI is intended for use on the full URI.
     -decodeURIComponent is intended to be used on URI components, that is, any part that lies between separators (; / ? : @ & = + $ , #).
So, in decodeURIComponent these separators are decoded also because they are regarded as text and not special characters.
 ```
Solves #8987 

Thank you @mukulmishra18 for your help.

Please review.

### DIFF
--- a/src/display/node_stream.js
+++ b/src/display/node_stream.js
@@ -365,7 +365,7 @@ class PDFNodeStreamFsFullReader extends BaseFullReader {
   constructor(stream) {
     super(stream);
 
-    let path = decodeURI(this._url.path);
+    let path = decodeURIComponent(this._url.path);
 
     // Remove the extra slash to get right path from url like `file:///C:/`
     if (fileUriRegex.test(this._url.href)) {
@@ -392,7 +392,7 @@ class PDFNodeStreamFsRangeReader extends BaseRangeReader {
   constructor(stream, start, end) {
     super(stream);
 
-    let path = decodeURI(this._url.path);
+    let path = decodeURIComponent(this._url.path);
 
     // Remove the extra slash to get right path from url like `file:///C:/`
     if (fileUriRegex.test(this._url.href)) {


mukulmishra18 (approved):
Looks good to me. @timvandermeij can you please have a look and run the tests to see if we are good to go.

shikhar-scs:
@timvandermeij you have any idea why are these tests failing ??

timvandermeij (approved):

timvandermeij:
Thank you for fixing this! Note that the reference test failure is a known intermittent problem and can be ignored for this patch.

shikhar-scs:
oh thanx a lot @timvandermeij @mukulmishra18 :smile:

PR #9368, opened 2018-01-16
end() is the official way to release a Writable stream

acchou:
See previous pull request #9362.


acchou:
I don't know why these tests would fail given the nature of this change.

Snuffleupagus (approved):
Thank you!

timvandermeij:
The reference test failure is a known intermittent one by the way, so no worries about that one.

Issue #7909, opened 2016-12-22
Feature request: Add type definitions for Typescript and make it available on NPM.


DannyBoyNg:
I would like to request "type definitions" for Typescript for this library. It would be great to use this library in my Typescript projects.

Thanks.

timvandermeij:
Could you elaborate a bit about what this is exactly? Why can't you use it in TypeScript project right now and how large is this change?

yurydelendik:
It's possible to re-create definition files from display/api.js, and it will be easier for person who already done that. We shall adapt style from https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html . Problem will be to keep api.js and definition files in sync but will be easier after the latter is created.

yurydelendik:
Notice that pdf.d.ts shall be created for the pdfjs-dist/build/pdf.js file and not files in the src/ folder.

dannyboyng:
I'm not sure how much work it is to make "type declaration files" because I have no experience myself creating "type declaration files". 

I think you can use third party Javascript libraries in Typescript without "type declaration files", but when "type declaration files" are available, you can get rich intellisense while using third party libraries.

https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html is a good resource to start learning about "type declaration files".

dannyboyng:
I would prefer if you could also make the "type declaration files" available on NPM. It makes it easy to include them in a project. 

For example by using to following command: npm install @types/pdfjs-dist --save-dev

yurydelendik:
Publishing them via "@types/" is one way to do it and usually made by third-party contributors. If we are committed to include support for .d.ts files, then they can be included in pdfjs-dist's package.json manifest (there not reason to place them into separate npm package afaik)

dannyboyng:
Apparently, you can bundle "type declaration files" together with your already existing NPM package. I forgot about that. No need to publish them to @types/

dmiller9911:
There a d.ts file in the DefinitelyTyped repo for this:  [DefinitelyTyped pdf.js](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4869992bc079b88280b9ff91213528904109e8ae/pdf/index.d.ts)  

I have not checked how complete or accurate it is, but it might be a good starting point.

StevenLiekens:
I tried using the typings from DefinitelyTyped but they are so bad it's laughable. (laugh so you don't cry)

[...]

PR #9369, opened 2018-01-16
Add v1 of typescript definition file

acchou:
This is a WIP typescript definition file (`index.d.ts`). There are several issues to resolve before this is ready:

- [x] Determine where in the repository the file should live. I've placed it in the root for now.
- [x] Modify gulpfile.js to copy `index.d.ts` to the root of the distribution directory `build/dist`. If the file is named `index.d.ts` there is no need for further configuration, it is optional but suggested by the TypeScript docs to add a reference to `package.json`.
- [x] Add some tests. Testing the types will require having TypeScript. I am hesitant to add TypeScript as a devDependency, but perhaps this is the best way to proceed... Alternatively we could omit tests or add tests in a separate repo. Some guidance here would be helpful.
- [x] Cull down the type file to the desired public interface. I've erred on the side of completeness, a fair amount of culling should be done to remove types not intended for external consumption. Some guidance would be helpful here, as I'm not entirely sure what was intended to be in the public API.

### DIFF
--- a/external/types/core.d.ts
+++ b/external/types/core.d.ts
@@ -1,0 +1,227 @@
+/* Copyright 2017 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// TypeScript type definitions for PDFJS
+
+// The _pdfjs namespace is a convenient place for making type
+// declarations. The actual exports are at the end of the file,
+// and pick symbols from this namespace to export in a shape that
+// resembles the API for PDFJS.
+declare namespace _pdfjs {
+  //
+  // Convenience types
+  //
+
+  /**
+   * 6 element transform matrix: [a, b, c, d, e, f]
+   * The transformation between two coordinate systems can be
+   * represented by a 3x3 transformation matrix written as follows:
+   *
+   *   | a b 0 |
+   *   | c d 0 |
+   *   | e f 1 |
+   *
+   * The transformation matrix has only six elements that can be changed,
+   * so it is usually represented by a six element array [a, b, c, d, e, f].
+   *
+   * Coordinate transformations are expressed as matrix multiplications:
+   *
+   *                             | a b 0 |
+   *   [x', y', 1] = [x, y, 1] x | c d 0 |
+   *                             | e f 1 |
+   */
+  type Transform = number[];
+
+  /**
+   * A generic 3d matrix transform M on a 3-vector [X, Y, Z]:
+   *
+   *   | a b c |   | X |
+   *   | d e f | x | Y |
+   *   | g h i |   | Z |
+   *
+   * M is assumed to be serialized as a 9 element array [a,b,c,d,e,f,g,h,i],
+   */
+  type Transform3d = number[];
+
+  // 2 dimensional point [x, y]
+  type Point = [number, number];
+
+  // 2 dimensional point object
+  interface Vertex {
+    x: number;
+    y: number;
+  }
+
+  // 3 dimensional point [x, y, z]
+  type Point3d = [number, number, number];
+
+  // Rectangle with 4 elements: [x1, y1, x2, y2]
+  type Rect = [number, number, number, number];
+
+  type PageIndex = number;
+  type PageLabels = string;
+
+  //
+  // src/core/obj.js
+  //
+  interface FileSpec {
+    filename: string;
+    content: Uint8Array;
+  }
+
+  interface Attachments {
+    [file: string]: FileSpec;
+  }
+
+  type DestinationTarget = any[];
+
+  type Destination = {
+    [destination: string]: any[];
+  };
+
+  //
+  // src/core/primitives.js
+  //
+  interface Ref {
+    num: number;
+    gen: number;
+  }
+
+  // Interning names
+  interface Name {
+    name: string;
+  }
+
+  interface StringMap {
+    [key: string]: string;
+  }
+
+  //
+  // src/core/annotation.js
+  //
+  interface Annotation {
+    annotationFlags: AnnotationFlag;
+    annotationType: AnnotationType;
+    borderStyle: AnnotationBorderStyle;
+    color: Color | null;
+    hasAppearance: boolean;
+    id: string;
+    rect: Rect;
+    subtype: AnnotationSubtype | null;
+
+    // For popup annotations
+    hasPopup?: boolean;
+    title?: string;
+    contents?: string;
+  }
+
+  /**
+   * Contains all data regarding an annotation's border style.
+   */
+  class AnnotationBorderStyle {
+    width: number;
+    style: AnnotationBorderStyleType;
+    dashArray: number[];
+    horizontalCornerRadius: number;
+    verticalCornerRadius: number;
+  }
+
+  interface WidgetAnnotation extends Annotation {
+    annotationType: AnnotationType.WIDGET;
+    fieldName: string;
+    fieldType: string;
+    fieldValue: string | string[] | null;
+    alternativeText: string;
+    defaultAppearance: string;
+    fieldFlags: number;
+    readOnly: boolean;
+  }
+  interface TextWidgetAnnotation extends WidgetAnnotation {
+    fieldValue: string;
+    textAlignment: number | null;
+    maxLen: number | null;
+    multiLine: boolean;
+    comb: boolean;
+  }
+  interface ButtonWidgetAnnotation extends WidgetAnnotation {
+    checkBox: boolean;
+    radioButton: boolean;
+    pushButton: boolean;
+    fieldValue: string | null;
+    buttonValue?: string | null;
+  }
+  interface ChoiceWidgetAnnotation extends WidgetAnnotation {
+    options: ChoiceOption[];
+    fieldValue: string[];
+    combo: boolean;
+    multiSelect: boolean;
+  }
+  interface ChoiceOption {
+      exportValue: string;
+      displayValue: string;
+  }
+  interface TextAnnotation extends Annotation {
+    annotationType: AnnotationType.TEXT;
+    name: string;
+  }
+  interface LinkAnnotation extends Annotation {
+    annotationType: AnnotationType.LINK;
+  }
+  interface PopupAnnotation extends Annotation {
+    annotationType: AnnotationType.POPUP;
+    parentType: string | null;
+    parentId: string;
+    title: string;
+    contents: string;
+    color: Color | null;
+  }
+  interface LineAnnotation extends Annotation {
+    annotationType: AnnotationType.LINE;
+    lineCoordinates: Rect;
+  }
+  interface SquareAnnotation extends Annotation {
+    annotationType: AnnotationType.SQUARE;
+  }
+  interface CircleAnnotation extends Annotation {
+    annotationType: AnnotationType.CIRCLE;
+  }
+  interface PolylineAnnotation extends Annotation {
+    annotationType: AnnotationType.POLYLINE;
+    vertices: Vertex[];
+  }
+  interface PolygonAnnotation extends Annotation {
+    annotationType: AnnotationType.POLYGON;
+    vertices: Vertex[];
+  }
+  interface HighlightAnnotation extends Annotation {
+    annotationType: AnnotationType.HIGHLIGHT;
+  }
+  interface UnderlineAnnotation extends Annotation {
+    annotationType: AnnotationType.UNDERLINE;
+  }
+  interface SquigglyAnnotation extends Annotation {
+    annotationType: AnnotationType.SQUIGGLY;
+  }
+  interface StrikeOutAnnotation extends Annotation {
+    annotationType: AnnotationType.STRIKEOUT;
+  }
+  interface StampAnnotation extends Annotation {
+    annotationType: AnnotationType.STAMP;
+  }
+  interface FileAttachmentAnnotation extends Annotation {
+    annotationType: AnnotationType.FILEATTACHMENT;
+    file: FileSpec;
+  }
+}
--- a/external/types/core.d.ts
+++ b/external/types/core.d.ts
@@ -205,6 +205,10 @@ declare namespace _pdfjs {
     annotationType: AnnotationType.POLYGON;
     vertices: Vertex[];
   }
+  interface InkAnnotation extends Annotation {
+    annotationType: AnnotationType.INK;
+    inkLists: Vertex[][];
+  }
   interface HighlightAnnotation extends Annotation {
     annotationType: AnnotationType.HIGHLIGHT;
   }
--- a/external/types/pdf.d.ts
+++ b/external/types/pdf.d.ts
@@ -1,0 +1,109 @@
+/* Copyright 2017 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// <reference path="dom.d.ts" />
+/// <reference path="shared.d.ts" />
+/// <reference path="core.d.ts" />
+/// <reference path="display.d.ts" />
+/// <reference path="web.d.ts" />
+
+declare module 'pdfjs-dist' {
+  //
+  // src/pdf.js
+  //
+  export import build = _pdfjs.build;
+  export import version = _pdfjs.version;
+  export import getDocument = _pdfjs.getDocument;
+  export import LoopbackPort = _pdfjs.LoopbackPort;
+  export import PDFDataRangeTransport = _pdfjs.PDFDataRangeTransport;
+  export import PDFWorker = _pdfjs.PDFWorker;
+  export import renderTextLayer = _pdfjs.renderTextLayer;
+  export import AnnotationLayer = _pdfjs.AnnotationLayer;
+  export import createPromiseCapability = _pdfjs.createPromiseCapability;
+  export import PasswordResponses = _pdfjs.PasswordResponses;
+  export import InvalidPDFException = _pdfjs.InvalidPDFException;
+  export import MissingPDFException = _pdfjs.MissingPDFException;
+  export import SVGGraphics = _pdfjs.SVGGraphics;
+  export import NativeImageDecoding = _pdfjs.NativeImageDecoding;
+  export import UnexpectedResponseException = _pdfjs.UnexpectedResponseException;
+  export import OPS = _pdfjs.OPS;
+  export import VerbosityLevel = _pdfjs.VerbosityLevel;
+  export import UNSUPPORTED_FEATURES = _pdfjs.UNSUPPORTED_FEATURES;
+  export import createValidAbsoluteUrl = _pdfjs.createValidAbsoluteUrl;
+  export import createObjectURL = _pdfjs.createObjectURL;
+  export import removeNullCharacters = _pdfjs.removeNullCharacters;
+  export import shadow = _pdfjs.shadow;
+  export import createBlob = _pdfjs.createBlob;
+  export import Util = _pdfjs.Util;
+  export import RenderingCancelledException = _pdfjs.RenderingCancelledException;
+  export import getFilenameFromUrl = _pdfjs.getFilenameFromUrl;
+  export import LinkTarget = _pdfjs.LinkTarget;
+  export import addLinkAttributes = _pdfjs.addLinkAttributes;
+  export import GlobalWorkerOptions = _pdfjs.GlobalWorkerOptions;
+  export import apiCompatibilityParams = _pdfjs.apiCompatibilityParams;
+
+  //
+  // Convenience types that are used often.
+  //
+  export import PDFDocumentProxy = _pdfjs.PDFDocumentProxy;
+  export import DocumentInitParameters = _pdfjs.DocumentInitParameters;
+  export import PDFPageProxy = _pdfjs.PDFPageProxy;
+  export import PDFDocumentLoadingTask = _pdfjs.PDFDocumentLoadingTask;
+  export import PageIndex = _pdfjs.PageIndex;
+  export import Outline = _pdfjs.Outline;
+  export import Transform = _pdfjs.Transform;
+  export import Point = _pdfjs.Point;
+  export import Rect = _pdfjs.Rect;
+  export import PDFObjects = _pdfjs.PDFObjects;
+  export import PDFOperatorList = _pdfjs.PDFOperatorList;
+  export import PageViewport = _pdfjs.PageViewport;
+  export import Annotation = _pdfjs.Annotation;
+  export import RenderParameters = _pdfjs.RenderParameters;
+  export import PDFViewerOptions = _pdfjs.PDFViewerOptions;
+  export import DocumentLoadingProgress = _pdfjs.DocumentLoadingProgress;
+
+  //
+  // All other types. Warning: this should only be used to access types, not
+  // functions or variables.
+  //
+  export import types = _pdfjs;
+}
+
+declare module 'pdfjs-dist/web/pdf_viewer' {
+  //
+  // web/pdf_viewer.component.js
+  //
+  export import PDFViewer = _pdfjs.PDFViewer;
+  export import PDFSinglePageViewer = _pdfjs.PDFSinglePageViewer;
+  export import PDFPageView = _pdfjs.PDFPageView;
+  export import PDFLinkService = _pdfjs.PDFLinkService;
+  export import SimpleLinkService = _pdfjs.SimpleLinkService;
+  export import TextLayerBuilder = _pdfjs.TextLayerBuilder;
+  export import DefaultTextLayerFactory = _pdfjs.DefaultTextLayerFactory;
+  export import AnnotationLayerBuilder = _pdfjs.AnnotationLayerBuilder;
+  export import DefaultAnnotationLayerFactory = _pdfjs.DefaultAnnotationLayerFactory;
+  export import PDFHistory = _pdfjs.PDFHistory;
+  export import PDFFindController = _pdfjs.PDFFindController;
+  export import EventBus = _pdfjs.EventBus;
+  export import DownloadManager = _pdfjs.DownloadManager;
+  export import ProgressBar = _pdfjs.ProgressBar;
+  export import GenericL10n = _pdfjs.GenericL10n;
+  export import NullL10n = _pdfjs.NullL10n;
+}
+
+// Webpack version.
+declare module 'pdfjs-dist/webpack' {
+  export * from 'pdfjs-dist';
+}
--- a/external/types/shared.d.ts
+++ b/external/types/shared.d.ts
@@ -16,516 +16,523 @@ interface HTMLAnchorElement {}
 interface ReadableStream {}
 
 declare namespace _pdfjs {
-    //
-    // src/shared/util.js
-    //
+  //
+  // src/shared/util.js
+  //
 
-    enum VerbosityLevel {}
-    namespace VerbosityLevel {
-        const ERRORS: VerbosityLevel;
-        const WARNINGS: VerbosityLevel;
-        const INFOS: VerbosityLevel;
-    }
+  enum VerbosityLevel {}
+  namespace VerbosityLevel {
+    const ERRORS: VerbosityLevel;
+    const WARNINGS: VerbosityLevel;
+    const INFOS: VerbosityLevel;
+  }
 
-    let shadow: typeof Object.defineProperty;
-    function createBlob(data: any, contentType: string): Blob;
-    function createObjectURL(
-        data: any,
-        contentType: string,
-        forceDataSchema?: boolean
-    ): string;
-    let isLittleEndian: boolean;
-    function removeNullCharacters(str: string): string;
+  let shadow: typeof Object.defineProperty;
+  function createBlob(data: any, contentType: string): Blob;
+  function createObjectURL(
+    data: any,
+    contentType: string,
+    forceDataSchema?: boolean
+  ): string;
+  let isLittleEndian: boolean;
+  function removeNullCharacters(str: string): string;
 
-    class PasswordException extends Error {
-        constructor(msg: string, code: PasswordResponses);
-        code: PasswordResponses;
-    }
+  class PasswordException extends Error {
+    constructor(msg: string, code: PasswordResponses);
+    code: PasswordResponses;
+  }
 
-    class UnknownErrorException extends Error {
-        constructor(msg: string, details: string);
-        details: string;
-    }
+  class UnknownErrorException extends Error {
+    constructor(msg: string, details: string);
+    details: string;
+  }
 
-    class UnexpectedResponseException extends Error {
-        constructor(msg: string, status: number);
-        status: number;
-    }
+  class UnexpectedResponseException extends Error {
+    constructor(msg: string, status: number);
+    status: number;
+  }
 
-    class InvalidPDFException extends Error {
-        constructor(msg: string);
-    }
+  class InvalidPDFException extends Error {
+    constructor(msg: string);
+  }
 
-    class MissingPDFException extends Error {
-        constructor(msg: string);
-    }
+  class MissingPDFException extends Error {
+    constructor(msg: string);
+  }
 
-    interface CloneArgs {
-        scale: number;
-        rotation: number;
-    }
+  interface CloneArgs {
+    scale: number;
+    rotation: number;
+  }
 
-    class PageViewport {
-        viewBox: Rect;
-        scale: number;
-        rotation: number;
-        offsetX: number;
-        offsetY: number;
+  class PageViewport {
+    viewBox: Rect;
+    scale: number;
+    rotation: number;
+    offsetX: number;
+    offsetY: number;
 
-        // creating transform for the following operations:
-        // translate(-centerX, -centerY), rotate and flip vertically,
-        // scale, and translate(offsetCanvasX, offsetCanvasY)
-        transform: Transform;
+    // creating transform for the following operations:
+    // translate(-centerX, -centerY), rotate and flip vertically,
+    // scale, and translate(offsetCanvasX, offsetCanvasY)
+    transform: Transform;
 
-        width: number;
-        height: number;
-        fontScale: number;
-
-        /**
-         * Clones viewport with additional properties.
-         * @param args {Object} (optional) If specified, may contain the 'scale' or
-         * 'rotation' properties to override the corresponding properties in
-         * the cloned viewport.
-         * @returns {PageViewport} Cloned viewport.
-         */
-        clone(args?: CloneArgs): PageViewport;
-
-        /**
-         * Converts PDF point to the viewport coordinates. For examples, useful for
-         * converting PDF location into canvas pixel coordinates.
-         * @param x {number} X coordinate.
-         * @param y {number} Y coordinate.
-         * @returns {Object} Object that contains 'x' and 'y' properties of the
-         * point in the viewport coordinate space.
-         * @see {@link convertToPdfPoint}
-         * @see {@link convertToViewportRectangle}
-         */
-        convertToViewportPoint(x: number, y: number): Point;
-
-        /**
-         * Converts PDF rectangle to the viewport coordinates.
-         * @param rect {Array} xMin, yMin, xMax and yMax coordinates.
-         * @returns {Array} Contains corresponding coordinates of the rectangle
-         * in the viewport coordinate space.
-         * @see {@link convertToViewportPoint}
-         */
-        convertToViewportRectangle(rect: Rect): Rect;
-
-        /**
-         * Converts viewport coordinates to the PDF location. For examples, useful
-         * for converting canvas pixel location into PDF one.
-         * @param x {number} X coordinate.
-         * @param y {number} Y coordinate.
-         * @returns {Object} Object that contains 'x' and 'y' properties of the
-         * point in the PDF coordinate space.
-         * @see {@link convertToViewportPoint}
-         */
-        convertToPdfPoint(x: number, y: number): Point;
-    }
+    width: number;
+    height: number;
+    fontScale: number;
 
     /**
-     * Promise Capability object.
-     *
-     * @typedef {Object} PromiseCapability
-     * @property {Promise} promise - A promise object.
-     * @property {function} resolve - Fulfills the promise.
-     * @property {function} reject - Rejects the promise.
+     * Clones viewport with additional properties.
+     * @param args {Object} (optional) If specified, may contain the 'scale' or
+     * 'rotation' properties to override the corresponding properties in
+     * the cloned viewport.
+     * @returns {PageViewport} Cloned viewport.
      */
-    interface PromiseCapability<T> {
-        promise: Promise<T>;
-        resolve: (val: T) => void;
-        reject: (reason?: any) => void;
-    }
+    clone(args?: CloneArgs): PageViewport;
 
     /**
-     * Creates a promise capability object.
-     * @alias createPromiseCapability
-     *
-     * @return {PromiseCapability} A capability object contains:
-     * - a Promise, resolve and reject methods.
+     * Converts PDF point to the viewport coordinates. For examples, useful for
+     * converting PDF location into canvas pixel coordinates.
+     * @param x {number} X coordinate.
+     * @param y {number} Y coordinate.
+     * @returns {Object} Object that contains 'x' and 'y' properties of the
+     * point in the viewport coordinate space.
+     * @see {@link convertToPdfPoint}
+     * @see {@link convertToViewportRectangle}
      */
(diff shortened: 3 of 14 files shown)


timvandermeij:
If complete, this fixes #7909. I'm wondering how the maintenance of this will go. Should we maintain it manually on can we generate this file somehow?

yurydelendik:
let's place it in e.g. "external/types/" folder and also split into several files, e.g. display.ts/core.ts/shared.ts

acchou:
I manually created this file, and I recommend that it be manually updated. There are many cases where there's a stylistic choice on how to specify the types, and in this code base in particular I think it serves as useful documentation, even for developers who are not using TypeScript.

The best way to help maintain it is probably to create a test file that exercises some of the most common APIs. When the APIs change, the test will fail and prompt the developer who changed the code to update the TypeScript definition and tests. 

I have taken a crack at this with an external test file (not included in this pull request), but I'm having trouble setting up the test scaffolding. Some issues:
- Where should the test go
- How to specify the dependency on TypeScript (e.g. is it ok to add TypeScript as a devDependency)
- How to set up the test itself. I'm not familiar with the test framework that PDFJS uses. If one of you could create an empty test file that is already working, I can add the tests themselves much easier.

acchou:
@yurydelendik It is possible to split the types into separate files, but it introduces several issues:

- The type files will need to be refactored if definitions move around, or if files are added or removed (which have a public API)
- It requires TypeScript references, which are a bit less well understood (at least by me) and will require some additional configuration. I'm worried this may cause issues for users down the road.
- I'm not sure about how well symbol lookup will work with IDEs.

I think the best solution would be to convert PDFJS itself to TypeScript, then the types would be part of the source files and no parallel type definitions or files would be needed. But this is probably too ambitious :)  However I'll say that this type definition file goes a long way towards that goal, if it's something the team would consider.

For now I'd suggest we put the effort into adding tests, then we can consider whether refactoring to split the types makes sense. Does that sound ok?

acchou:
I'll also add that as I mentioned in the original PR, there's much culling that could probably be done to make the type file smaller. Doing that first might result in a small enough file that there's less of a feeling that it needs to be split up.

Snuffleupagus:
> The best way to help maintain it is probably to create a test file that exercises some of the most common APIs. When the APIs change, the test will fail and prompt the developer who changed the code to update the TypeScript definition and tests.

I suppose that my general question/objection here is if people using TypeScript will actually help maintain these definitions, since it seems somewhat unfair to put that burden on every PDF.js contributor considering that the library is only using regular JavaScript!?

> Cull down the type file to the desired public interface. I've erred on the side of completeness, a fair amount of culling should be done to remove types not intended for external consumption. Some guidance would be helpful here, as I'm not entirely sure what was intended to be in the public API.

In general, the public interface is the one found in https://github.com/mozilla/pdf.js/blob/master/src/display/api.js and perhaps in some of the other files in https://github.com/mozilla/pdf.js/tree/master/src/display (i.e. what ends up in the built `pdf.js` file).
Probably it also makes sense to document e.g. the viewer components, e.g. as listed in https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.component.js.

However, note that all code found in https://github.com/mozilla/pdf.js/tree/master/src/core is running in a web worker, and is thus not even accessible much less part of any public interface.
Finally, note that as part of upcoming version `2.0` of PDF.js, see https://github.com/mozilla/pdf.js/projects/5, we're also aiming to remove the global `PDFJS` object.

yurydelendik:
>  The type files will need to be refactored if definitions move around, or if files are added or removed (which have a public API)

We already tracking changes to public API. So modifying/syncing with .d.ts might not be an issue

>   It requires TypeScript references, which are a bit less well understood (at least by me) and will require some additional configuration. I'm worried this may cause issues for users down the road.

When TS builds .ts, it generates file that references other files. So it's pretty common practice.

>   I'm not sure about how well symbol lookup will work with IDEs.

As above, IDE will support stuff generated by standard TS  compiler.

> convert PDFJS itself to TypeScript,

There is also alternative: Flow. I fonder how easy to convert TS to Flow (or other way around).

Snuffleupagus:
> I think the best solution would be to convert PDFJS itself to TypeScript, then the types would be part of the source files and no parallel type definitions or files would be needed. But this is probably too ambitious :)

We're still in the process of slowly transitioning the code-base to ES6, and generally more modern JavaScript usage. However, I'd seriously question if it's worth the time and effort required to migrate everything to TypeScript (especially considering the size of the code-base and the number of existing contributors).

acchou:
@Snuffleupagus I understand that the intended API is in api.js, but the actual interface exposed includes other types that are required to support those in api.js. Those types in turn expose access to other types. I took the closure of accessible types in this manner to create the type definition file. This is the set of files the types actually come from:

```
src/display/global.js
src/shared/util.js
src/display/api.js
src/display/dom_utils.js
src/display/annotation_layer.js
src/display/text_layer.js
src/display/metadata.js
src/display/svg.js
src/core/obj.js
src/core/primitives.js
src/core/annotation.js
web/interfaces.js
web/pdf_link_service.js
web/ui_utils.js
web/pdf_rendering_queue.js
web/base_viewer.js
web/text_layer_builder.js
web/pdf_find_controller.js
web/annotation_layer_builder.js
web/download_manager.js
web/pdf_page_view.js
web/pdf_viewer.js
web/pdf_thumbnail_viewer.js
web/pdf_thumbnail_view.js
web/pdf_history.js
src/display/canvas.js
src/core/fonts.js
src/core/font_renderer.js
src/core/cmap.js
src/core/stream.js
src/display/global.js
src/pdf.js
```

@yurydelendik This is another thing to consider, do we really want to add a `.ts` file for every one of these? We should cull this list first, then decide if splitting into separate files is needed.

To your other question, unfortunately Flow types and TypeScript types aren't compatible at the moment, though it would be a very nice feature to have. My understanding is that TS is getting more traction, and in practice my understanding is that there are more requests for TS than Flow types for pdf.js from the user community.

acchou:
If you'd like to see which types come from which files, they're documented in the source code: https://github.com/acchou/pdf.js/blob/41205e3a8dd09e49f710760658a0c9d2168b0381/index.d.ts

Snuffleupagus:
> [...] but the actual interface exposed includes other types that are required to support those in api.js. Those types in turn expose access to other types.

Please keep in mind here that what's returned through the API is *not* at all the same types that exist in the worker. When data is passed from a web worker to the main thread (or vice versa), via `postMessage`, the ["structured cloning algorithm"](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) is being used.

What this means in practice is that if you have e.g. an object on the *worker* side which is `instanceof Name`, see https://github.com/mozilla/pdf.js/blob/master/src/core/primitives.js#L19, then such an object once sent to the main thread (i.e. the API) consists of nothing more than a *plain* object of the format `{ name: 'some_string', }`.

The above is just *one* very simply example out of many more, but as you can see the types used in the worker doesn't really make sense on the main/API side of things. What's returned by various API methods could probably be thought of more like a proxy to the underlying worker data instead.

acchou:
Ok, this was something that I didn't understand. When the types are corrected, they should serve as helpful documentation of what the public API actually looks like.

I'll take some time to make this correction; what would be most helpful is to understand how to create a test for this, then I can write actual tests for these types. I need help with the test scaffolding/setup. Where should I put tests for types, and how should I set it up?

acchou:
It should be easy to convert: just change classes to interfaces, and remove methods. This is for all data objects returned by `postMessage`, which should help cull down the type file a lot.

For example, I could modify `Name` as:

```
  interface Name {
    name: string;
  }
```

acchou:
Let's consider `PDFPageProxy` as an example.  It has a method:
```
    getAnnotations(params?: GetAnnotationsParameters): Promise<Annotation[]>;
```
In general it looks like in the pdfjs API, things returned as `Promise<T>` are posted from workers using structured cloning, so I'm assuming that `Annotation[]` will be a cloned object.

Currently I have `Annotation` typed like this:
```
class Annotation {
    data: {
      annotationFlags: number;
      borderStyle: AnnotationBorderStyle;
      color: Color | null;
      hasAppearance: boolean;
      id: string;
      rect: Rect;
      subtype: AnnotationSubtype | null;
    };

    viewable: boolean;
    printable: boolean;

    /**
     * Check if a provided flag is set.
     *
     * @public
     * @memberof Annotation
     * @param {number} flag - Hexadecimal representation for an annotation
     *                        characteristic
     * @return {boolean}
     * @see {@link shared/util.js}
     */
    hasFlag(flag: AnnotationFieldFlag): boolean;
  }
```
There are a few things I could change:
- Turn `Annotation` into an `interface` instead of `class`.
- Remove the `hasFlag` method because it won't be accessible in the API.
- Because `hasFlag` is the only place that mentions the enum type `AnnotationFieldFlag`, I can remove that too. So the following enum is not accessible to the public API:
```
 enum AnnotationFieldFlag {
    READONLY,
    REQUIRED,
    NOEXPORT,
    MULTILINE,
    PASSWORD,
    NOTOGGLETOOFF,
    RADIO,
    PUSHBUTTON,
    COMBO,
    EDIT,
    SORT,
    FILESELECT,
    MULTISELECT,
    DONOTSPELLCHECK,
    DONOTSCROLL,
    COMB,
    RICHTEXT,
    RADIOSINUNISON,
    COMMITONSELCHANGE
  }
```
Therefore the final type for `Annotation` would be:
```
  interface Annotation {
    data: {
      annotationFlags: number;
      borderStyle: AnnotationBorderStyle;
      color: Color | null;
      hasAppearance: boolean;
      id: string;
      rect: Rect;
      subtype: AnnotationSubtype | null;
    };

    viewable: boolean;
    printable: boolean;
  }
```
Does that sound correct?

Snuffleupagus:
> Does that sound correct?

Not quite, since the only thing that's returned for annotations is the `data` object, as can be seen in https://github.com/mozilla/pdf.js/blob/96c573ad387696c6efd24d080a44d592b50da703/src/core/document.js#L310-L319

---

As a piece of general advice: I'd suggest playing with the API locally to check what's *actually* returned from its various methods, since it might be difficult to reason about that theoretically if you're not already familiar with that particular part of the code-base.

acchou:
I'll have to do that some more. It's often not enough though, because dynamically it's hard to know if you're seeing everything - there are plenty of optional keys in various types that don't always show up every time, for example. 

I'll try adding a unit test, that should help document things and also provide dynamic verification of the types.

timvandermeij:
The `StatTimer` class is used in our debugging tools, so it's not part of the API.

acchou:
Ok thanks. Another issue: `AnnotationType` doesn't appear to be exported, which means the `annotationType` field of `Annotation` will be an opaque `number` to clients. Am I missing something or should `AnnotationType` be exported?

acchou:
Is `PDFWorker` meant to be opaque, or are its methods intended for public consumption? Here's the typing I have so far, but it's unclear what is intended. `api_spec.js` uses these some of these methods.

```
    class PDFWorker {
        constructor(name: string, port?: MessagePort);
        name: string;
        destroyed: boolean;
        postMessageTransfers: boolean;
        promise: Promise<void>;
        port: MessagePort;
        messageHandler: MessageHandler;
        destroy(): void;
        fromPort(port: MessagePort): PDFWorker;
    }
```

acchou:
I just pushed several changes:
- Separate types into multiple files in types/*.d.ts, as suggested by @yurydelendik 
- As suggested by @Snuffleupagus, verified static types by examining runtime types. Made some fixes to the static types as a result.
- Added types for viewer components
- Add tests for the most common APIs to verify runtime type matches type declarations. These tests live in examples/typescript. The tests have a dependency on pdf files from test/pdfs.
- Modify dist target in gulpfile to copy type files to build/dist/types
- Modify generated package.json to add reference to toplevel type declaration file pdf.d.ts. This means TypeScript consumers of the pdfjs-dist module automatically get the correct typings.

For the viewer components, I exported them by default into the `PDFJS` object. I'm not sure I've set that part up correctly.

acchou:
Also, the tests are not run as part of any other gulp task, they are separate at the moment.

timvandermeij:
I have not done an in-depth review, but I did notice two things. Most of the files are just type definitions, but why does `examples/typescript/src/domstubs.ts` contain actual logic, i.e., is it not possible to use the existing logic in any way since we'd really prefer not to duplicate it? Moreover, since we're in the process of deprecating the global `PDFJS` object (see https://github.com/mozilla/pdf.js/blob/master/src/display/global.js#L36), we should find another solution for that here, but I'm not entirely sure what a good approach for this would be.

I notice that there may be some unanswered questions above, but maybe in the meantime you already found some answers. If not, could you compile a list of remaining questions in one comment so it's easier for us to answer them?

For reviewing, it would help to rebase onto the current master and to squash the commits (see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits if you're not familiar with it). Now there are 21 commits and 24 changes files, which makes it hard to get a good overview of the changes since some changes are just there because of merge commits. After rebasing and squashing, there should be no more merge commits and one (or perhaps two if you want to split the testing from the actual type definitions) functional commits. Thank you!

timvandermeij:
For Gulp, I think the best place is https://github.com/mozilla/pdf.js/blob/a7cb560f170ed6ace1fb758129972eb0daed3eb4/gulpfile.js#L1469. The `externaltest` target already runs some separate test files for code in the `external` folder, so you can add your tests there too.

acchou:
For the external test, my test requires an `npm install` to pull in some dependencies. I'm not super familiar with gulp and I'm not sure it's a good idea to blindly do an `npm install` and `npm test`. Open to suggestions on that.

The only remaining question I have is how to structure the viewer types. These are currently collected in `types/web.d.ts` and exported in `types/pdf.d.ts`. But this means an import of `"pdfjs-dist"` automatically pulls the viewer's exported types into the `PDFJS` variable as well (following what's done by https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.component.js. Perhaps if `PDFJS` is going away then there is a more modular approach.

On the topic of removing `PDFJS`, this should be straightforward to accommodate. Everything is defined in the `_pdfjs` namespace, and re-exported in `pdf.d.ts`. To reshape the API just move things around in the latter file to re-export appropriately.

I'll squash the commits shortly.

acchou:
Squashed.

acchou:
Regarding `domstubs.ts`, I could work on removing it. It will mean there's a dependency between `examples/typescript` and `examples/node`. Maybe `domstubs.js` should live somewhere else? Or we can just accept the sibling dependency.

acchou:
Ok, I've removed `domstubs.ts` and made some small changes to `domstubs.js` to work with the typescript tests.

Snuffleupagus:
From a cursory look, there still appears to be cases where https://github.com/mozilla/pdf.js/pull/9369#issuecomment-358589858 hasn't been taken into account.

One such example is https://github.com/mozilla/pdf.js/pull/9369/files#diff-e5a3b04eccbbf519a4c1a76116253f13R317, which doesn't make sense once the font data has been sent to the main-thread.
That data-structure appear to be reference on the main-thread, e.g. in https://github.com/mozilla/pdf.js/pull/9369/files#diff-ff0f20673748cb2bc6d19776ad0ee4feR79, where it doesn't really make sense. In this particular case, you can check the contents of https://github.com/mozilla/pdf.js/blob/b6c57d90887c1a936e2638e83cf88576aea8ca36/src/display/api.js#L1753
to see what the font data actually looks like once it arrives on the main-thread/API side. 

Again, please keep in mind that all code residing in https://github.com/mozilla/pdf.js/tree/master/src/core is running in a worker, and the code in https://github.com/mozilla/pdf.js/tree/master/src/display is running on the main-thread and is what's exposed through the API.

acchou:
Ah yes, I didn't test the SVG api. I'll fix that.

acchou:
@Snuffleupagus I just committed some changes that sidestep this a bit. After looking at SVG a bit more, the API appears to be very small.  So I removed most of the fields and boiled it down to only what's needed, and moved the remaining definition to `display.d.ts` instead of breaking out SVG into its own definition file. This eliminated the last direct reference to Font and some other types, which I removed from `core.d.ts`.

This is a bit of a cop out because Fonts and Glyphs are accessible outside -- but only indirectly, for example through `commonObj`, which have type `PDFObjects`:

```
interface PDFObjects {
    objs: { [objId: string]: PDFObject };
    ...
}
```

And I've chosen to type the data within `PDFObject` as `any`:

```
  interface PDFObject {
    data: any;
    ...
  }
```

I could be more specific about the type here, but it seems this is a particularly dynamic field that has both dynamic keys and dynamic values, which TypeScript cannot type much more specifically than what I've shown above.

In light of this limitation, I thought it would be best to avoid publishing a type for `Font`, `Glyph`, and others that are only needed for SVG internals, or only accessible through an `any`-typed field. In the latter case the user would have to manually cast the field from `any` to a known type. I feel this is probably a relatively rare use case and in such situations the developer would probably need to consult the runtime types anyway.

In summary, some types were left out, and the remaining ones in `core.d.ts` appear correct from what I can tell. There are some runtime tests to verify them, though the tests for `Annotation` don't cover all subclasses.

Let me know if this sounds ok, or if we should take another approach.

acchou:
I'd like to add a little motivation to this pull request to hopefully help get some attention towards reviewing and merging it.

**Should pdf.js invest in type declaration files?** Type declarations can get complex and are not free to maintain, so it's a reasonable question to ask. I'm not in a position to know the resources available to this project to judge cost/benefit, but I can comment on the benefit side of the equation:
- It serves as useful documentation of known types and can help users understand the public API.
- Editors like vscode are increasingly able to use type declaration files for autocomplete and in-editor documentation [even if you use only plain JS](https://blog.mariusschulz.com/2014/05/19/using-typescripts-type-definition-files-to-get-tooling-support-for-plain-javascript
)
- Users of TypeScript you can get type checking which reduces errors and increases productivity. This greatly reduces the friction when adopting pdf.js for TypeScript developers.

**Why TypeScript, as opposed to some other type system like Flow?**
- It's not mutually exclusive. If someone wants to contribute type definitions for Flow then they can do so.
- If we must choose, consider that by several metrics TypeScript is more popular. TS has 30k github stars vs. 15k for Flow.
- TypeScript has a much broader library of type definitions. [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) has over 5.5k contributors. [flow-typed](https://github.com/flowtype/flow-typed) has fewer than 500.

Pdf.js is a large code base, and the documentation for its API is [very limited](https://mozilla.github.io/pdf.js/api/). A type declaration file can help by providing information that would otherwise require digging deep into the source code and/or examining incomplete examples at runtime.

My experience writing type definitions for pdf.js has demonstrated that the API is far from simple, and in many places it is very unclear what constitutes a public API vs. some accidentally leaked internal representation. This is not a criticism. Considering the size of the code base and the lack of static checking that is inherent in JavaScript, the fact that the code base is as coherent as it is, is nothing short of heroic.

Nevertheless, writing types forces more precision and overt choices. Putting in the effort makes the code base more accessible to good tooling, more accessible to developers who want to contribute to the code base, and more maintainable. And clients who use pdf.js with these types will find it much easier to keep up to date with changes to the API.

nicolas-schreiber:
@acchou How can I create a PDFViewer using these definition files and pdfjs-dist? 
It seems PDFViewer is still undefined when simply importing pdfjs-dist.

acchou:
If I have the types right, it should be as easy as:

```
import { PDFViewer } from "pdfjs-dist";

const viewer = new PDFViewer({
    container: div
});
```

Of course, this will only work once this pull request is merged. Or you can build pdf.js from my fork.

brendandahl:
Are there any tools to warn us when definitions are out of sync?

As somewhat of an aside, I would be open to the idea of someone converting/re-writing our "public" api to TypeScript. Having used plain JS and TypeScript in rather large projects now I would never start something in plain JS again.

acchou:
I've added tests that dynamically check some subset of the most commonly used types (see `examples/typescript`). More work could be done there. There are also projects that attempt to add dynamic checks but I think they're too early stage and/or require intrusive changes at the moment.

This type declaration file should go a long way towards making the public API based on TypeScript. Let me know if you have a specific interface in mind... if the project can come to consensus on this then I might take it on depending on how big the project is.

nicolas-schreiber:
@acchou I did build from your fork but I get following error: 

> pdfjs_dist_1.PDFViewer is not a constructor

This might be due to the PDFViewer residing in a different folder and file.
Should I create an Issue or something? I am not sure here is the right place to discuss this.

nicolas-schreiber:
And by the way, these are missing in display.d.ts and in pdf.d.ts:
```
  let showPreviousViewOnLoad: boolean;
  let defaultZoomValue: string;
  let sidebarViewOnLoad: number;
  let enableHandToolOnLoad: boolean;
  let cursorToolOnLoad: number;
  let enableWebGL: boolean;
  let pdfBugEnabled: boolean;
  let disableTextLayer: boolean;
  let useOnlyCssZoom: boolean;
  let enhanceTextSelection: boolean;
  let renderer: string;
  let renderInteractiveForms: boolean;
  let enablePrintAutoRotate: boolean;
  let disablePageMode: boolean;
  let disablePageLabels: boolean;
```

acchou:
@nicolas-schreiber Feel free to create an issue on my fork for this. I think the issue is that I added PDFViewer to the toplevel export to make the type accessible easily, but it cannot be constructed through that variable. Instead construct it through the PDFJS variable:

```typescript
import { PDFJS } from "pdfjs-dist";

const viewer = new PDFJS.PDFViewer({
    container: div
});
```
Re: the missing variables, how exactly are these set? If you could provide a PR on my fork that would be helpful.

nicolas-schreiber:
There is no option to create Issues on your fork. And I tried removing that PDFJS variable. Still the same behaviour.
I looked into it a bit more and the PDFJS Object was actually fine but all the stuff from `web/pdf_viewer.js` was missing. The only way I could fix it was to manually import it.

I created a fork and made two commits with separate additions (only to the typings, not the pdf_viewer):
https://github.com/nicolas-schreiber/pdf.js/commit/08321877b298d78afc6ddf33e7e3044f10134a22
https://github.com/nicolas-schreiber/pdf.js/commit/508a674a7174f7beb5495388b3cba662e38bf382

acchou:
Modified my fork to allow issues.

I haven't used the viewer; can you provide an example of how you included it? I'm not sure how it's packaged and how to modify the type definitions to reflect that. If as I suspect it's intended to be included in a script tag, then I think the existing typing is the best we can do?

acchou:
Looking at `src/examples/components` it looks like the viewer variables are added to `PDFJS` by a script tag like this:

```html
  <script src="../../node_modules/pdfjs-dist/build/pdf.js"></script>
  <script src="../../node_modules/pdfjs-dist/web/pdf_viewer.js"></script>
```
So `PDFViewer` and the other viewer type declarations are not included by default when importing `pdfjs-dist`, by design. When a script tag is used to include `pdf_viewer.js`, the definitions it provides are added to the global `PDFJS` variable. I'm not exactly where this behavior falls in the TypeScript scheme of things (see https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html).

The type definition file currently (after the [small change](https://github.com/mozilla/pdf.js/pull/9369/commits/60a03f763e0f9ef5fa7c95ec734eaa9167ebb5a5) yesterday) exports a more complete `PDFJS` variable that includes types as if pdf_viewer.js is included. I'm not sure how to structure the type definitions to say "only include the type for `PDFViewer` if pdf_viewer.js is added via script tag". Perhaps the best we can do is improve the documentation by adding some comments about this situation to the type declaration file...

I added a skeleton `PDFFindController` declaration per your suggestion. `PDFFindController` is a good example of a class where a type declaration could make things much clearer for a client of pdf.js. I currently declare it as:

```
  interface FindCommandState {
    query: string,
    phraseSearch: boolean,
    caseSensitive: boolean,
    highlightAll: boolean,
    findPrevious: boolean,
  }

  class PDFFindController {
    constructor (options: PDFFindControllerOptions);
    reset(): void;
    nextMatch(): void;
    executeCommand(cmd: string, state: FindCommandState): void;
  }
```

This is just my attempt at clarifying the API, but there are many questions remaining:
- there are many other non-underscore methods in `PDFFindController`, should they be part of the public API?
- `highlightAll` is never accessed in the implementation of `PDFFindController`, though there are examples of it around the code base. Should it be there?
- are any of the fields in `FindCommandState` optional? If left undefined, is that ok usage for the client?

Clarifying these questions and refining the type declaration would provide users with a much clearer idea of how to use this API and avoid mistakes.

As for https://github.com/nicolas-schreiber/pdf.js/commit/08321877b298d78afc6ddf33e7e3044f10134a22, I'd like to clarify that this is correct usage of pdf.js:

```
// Assuming pdf_viewer.js is loaded via script tag
import { showPreviousViewOnLoad } from "pdfjs-dist";
showPreviousViewOnLoad = true;
```

If that's correct (a test would be helpful though that would require some additional scaffolding effort) then I'll add a comment indicating these variables are accessible only with pdf_viewer.js.

acchou:
Squashed again to make it easier to review.

external/types/web.d.ts
```diff
@@ -0,0 +1,924 @@
...
+   * @property {IL10n} l10n - Localization service.
+   */
+  interface PDFThumbnailViewerOptions {
+    container: HTMLDivElement;
+    linkService: IPDFLinkService;
+    renderingQueue: PDFRenderingQueue;
+    l10n: IL10n;
+  }
+
+  class PDFThumbnailViewer {
```
Snuffleupagus: This isn't exposed, please see https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.component.js for a list of the viewer components that are.

external/types/web.d.ts
```diff
@@ -0,0 +1,924 @@
...
+    linkService: IPDFLinkService;
+    renderingQueue: PDFRenderingQueue;
+    disableCanvasToImageConversion?: boolean;
+    l10n: IL10n;
+  }
+
+  //
+  // web/pdf_thumbnail_view.js
+  //
+  class PDFThumbnailView {
```
Snuffleupagus: This isn't exposed, please see https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.component.js for a list of the viewer components that are.

external/types/web.d.ts
acchou: Right, the only things that are exposed directly are defined in `pdf.d.ts` here: https://github.com/mozilla/pdf.js/pull/9369/files/74331b9822b16f59456667b935b3ea430b85b29c#diff-1fc4703ffd3b1030ba834559462f6879R80

A client can't reference `PDFThumbnailViewer` directly; they will have to get one through an access path. (Well, there is one exception which is through the [`types` namespace](https://github.com/mozilla/pdf.js/pull/9369/files/74331b9822b16f59456667b935b3ea430b85b29c#diff-1fc4703ffd3b1030ba834559462f6879R151), which is just an alias for the internal `_pdfjs` namespace. I added `types` because it can be convenient to declare variables, arguments, and return values using types that are not necessarily exposed directly. There is warning that this is not intended to access non-types, because they won't work.)

Some of the things that *are* exposed are `PDFViewer`, `PDFSinglePageViewer`, and `PDFPageView`, all of which have an access path that reaches to `PDFThumbnailViewer` and `PDFThumbnailView`. For example:

`PDFViewer` => `PDFRenderingQueue`:
https://github.com/mozilla/pdf.js/pull/9369/files/74331b9822b16f59456667b935b3ea430b85b29c#diff-4e0cc7eb4db8e8e1efac0bc43f50e0b5R685
`PDFRenderingQueue` => `PDFThumbnailViewer`:
https://github.com/mozilla/pdf.js/pull/9369/files/74331b9822b16f59456667b935b3ea430b85b29c#diff-4e0cc7eb4db8e8e1efac0bc43f50e0b5R230

I just confirmed dynamically that the field `pdfThumbnailViewer` is accessible (though set to null) in the `simpleviewer` example through the access path: `pdfViewer.renderingQueue.pdfThumbnailViewer`. It's `null`, so *if* `pdfThumbnailViewer` is ever set, it will be accessible dynamically there.

For `PDFViewer` the path is `PDFViewer` => `BaseViewer` => `PDFViewerOptions` => `PDFRenderingQueue`: https://github.com/mozilla/pdf.js/pull/9369/files/74331b9822b16f59456667b935b3ea430b85b29c#diff-4e0cc7eb4db8e8e1efac0bc43f50e0b5R424

`PDFThumbnailView` is only accessible through `PDFThumbnailViewer`, which is only accessible through `PDFRenderingQueue`, so perhaps one of the following is the case:
- `PDFRenderingQueue` is not supposed to be an accessible API?
- `PDFRenderingQueue` is supposed to be opaque, with no accessible internals?
- `PDFThumbnailViewer` is not supposed to be an accessible API, specifically from `PDFRenderingQueue`?
- `PDFThumbnailViewer` is accessible but opaque?
- `PDFThumbnailViewer` actually *is* supposed to be accessible, but only indirectly through access paths that go through `PDFRenderingQueue`?

external/types/pdf.d.ts
```diff
@@ -78,6 +78,8 @@ declare module 'pdfjs-dist' {
 
   //
   // web/pdf_viewer.component.js
+  // Warning: These functions are only accessible with web/pdf_viewer.js.
+  //          It is NOT automatically imported with 'pdfjs-dist'.
```
Snuffleupagus: Unfortunately this is simply *not* correct, since the variables added below are not exposed *anywhere*! And certainly not on the, [soon to be removed](https://github.com/mozilla/pdf.js/projects/5), global `PDFJS` object.

The variables added here correspond to https://github.com/mozilla/pdf.js/blob/master/web/default_preferences.json, which are *only* ever used in the default viewer (i.e. the one found at https://mozilla.github.io/pdf.js/web/viewer.html) and nowhere else. These are *not* used/exposed either through the PDF.js API, nor through https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.component.js. Hence the default viewer, i.e. file `web/app.js`, shouldn't be part of any "public" API.

Snuffleupagus: Unfortunately this is simply *not* correct, since the variables added below are not exposed *anywhere*! And certainly not on the, [soon to be removed](https://github.com/mozilla/pdf.js/projects/5), global `PDFJS` object.

The variables added here correspond to https://github.com/mozilla/pdf.js/blob/master/web/default_preferences.json, which are *only* ever used in the default viewer (i.e. the one found at https://mozilla.github.io/pdf.js/web/viewer.html) and nowhere else. These are *not* used/exposed either through the PDF.js API, nor through https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.component.js. Hence the default viewer, i.e. file `web/app.js`, shouldn't be part of any "public" API.

external/types/web.d.ts
```diff
@@ -921,4 +921,120 @@ declare namespace _pdfjs {
 
   // Constructor function
   let GenericL10n: new (lang: string) => GenericL10n;
+
+  //
+  // web/app.js
```
Snuffleupagus: The variables added here correspond to https://github.com/mozilla/pdf.js/blob/master/web/default_preferences.json, which are only ever used in the default viewer (i.e. the one found at https://mozilla.github.io/pdf.js/web/viewer.html) and nowhere else. These are not used/exposed either through the PDF.js API, nor through https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.component.js. Hence the default viewer, i.e. file web/app.js, shouldn't be part of any "public" API.

external/types/pdf.d.ts
nicolas-schreiber: I am using some of these variables from the PDFJS object, so they have to be exposed.
Some are for example used in your examples:
https://github.com/mozilla/pdf.js/tree/master/examples/mobile-viewer

Snuffleupagus: > Some are for example used in your examples:
> https://github.com/mozilla/pdf.js/tree/master/examples/mobile-viewer

That example is *not* in any way utilizing https://github.com/mozilla/pdf.js/blob/master/web/default_preferences.json, either directly or indirectly.
I'm not sure exactly which ones that you claim is used, but if you mean https://github.com/mozilla/pdf.js/blob/ea3d8450d218dd53619c72c934547fe57f22530f/examples/mobile-viewer/viewer.js#L24-L25 then those are already covered by
https://github.com/mozilla/pdf.js/blob/ea3d8450d218dd53619c72c934547fe57f22530f/web/ui_utils.js#L76-L81
and
https://github.com/mozilla/pdf.js/blob/ea3d8450d218dd53619c72c934547fe57f22530f/web/ui_utils.js#L98-L103

nicolas-schreiber: You said

> the variables added below (in this commit) are not exposed anywhere!

`useOnlyCssZoom` and `disableTextLayer` are part of those variables.
I don't get how `ui_utils.js` covers these. It only makes sure that the variables are not undefined.

nicolas-schreiber: Ok, I might just have understood something. Can it be that some of those variables should not be accessible but somehow still are? The TextLayer, for example, should be activated through creating a new TextLayerBuilder and adding it to the PDFViewer?

If that is so, I just had the biggest aha moment.
And this would mean good Definition Files would really help to document all that stuff.

external/types/web.d.ts
acchou: I just undid this merge and squashed again, so we are back where we were before the pull request.

acchou:
After PR #9479 settles, I'll take a stab at modifying the TypeScript declaration files to reflect the new API shape. This will be a good test for how easy (or difficult) the declaration files will be to maintain.

swftvsn:
What's the status on types currently? I see that #9479 has been merged, so this work could continue? Is there anything I can help with considering my limited knowledge?

acchou:
Thanks for the question. The work will continue, I plan on incorporating changes reflecting #9479 shortly.

external/types/pdf.d.ts
```diff
@@ -0,0 +1,112 @@
...
+  export import DocumentLoadingProgress = _pdfjs.DocumentLoadingProgress;
+
+  //
+  // All other types. Warning: this should only be used to access types, not
+  // functions or variables.
+  //
+  export import types = _pdfjs;
+}
+
+declare module 'pdfjs-dist' {
```
nicolas-schreiber: I would call this `pdfjs-dist/web/pdf_viewer`.
This conforms with the `webpackUniversalModuleDefinition` in the built `pdf_viewer.js`.
I can actually use it with a `pdfjs-dist/web/pdf_viewer/webpack` module and Webpack to automatically import the file.

external/types/web.d.ts
```diff
@@ -0,0 +1,992 @@
...
+    eventBus: EventBus;
+    externalLinkTarget?: LinkTarget;
+    externalLinkRel?: string;
+  }
+
+  /**
+   * @param {PDFLinkServiceOptions} options
+   * Constructor for PDFLinkService
+   */
+  let PDFLinkService: new (options: PDFLinkServiceOptions) => PDFLinkService;
```
nicolas-schreiber: I do not know how up-to-date this example is but it does not supply any parameter to the constructor. https://github.com/mozilla/pdf.js/blob/master/examples/components/simpleviewer.js

external/types/pdf.d.ts
acchou: @Snuffleupagus Could you comment on what module the viewer components belong in, if any? It looks like @nicolas-schreiber is right that in `dist/web/pdf_viewer.js` there is a module definition for `pdfjs-dist/web/pdf_viewer` but the official download (https://mozilla.github.io/pdf.js/getting_started/#download) doesn't have this file at all, instead it has `web/viewer.js` and no module definition for `pdf_viewer`.

I'm unsure what the right thing is here.

external/types/web.d.ts
acchou: You're right.

acchou:
After looking at pdfjs-dist some more, I believe @nicolas-schreiber is right. I modified the module declaration for the viewer to `pdfjs-dist/web/pdf_viewer` and updated the constructor function type for `PDFLinkService`, and squashed again.

acchou:
Note that the declaration for the `pdfjs-dist/web/pdf_viewer` requires use as a module, not as a global `pdfjsViewer` variable in a script. Given that TypeScript clients are more likely to use pdf.js as a module than a script, I think this is ok.

I did play around with the TypeScript methods of declaring UMD modules but couldn't figure it out. Maybe someone else who's more familiar with it can issue a future PR if that's an important use case for them.

In the meantime, the viewer is accessible as a module.

SylvesterMachielse:
Does anyone know the status?

acchou:
It's largely complete (as of 3/27 - I haven't checked more recent updates), but the lack of feedback from maintainers means it's basically in limbo... Maybe someone from the team could select one of these options:

- [ ] we can't review this now, maybe by XYZ date
- [ ] sorry we're just not interested in adding typescript type declarations to pdf.js
- [ ] we have some issues with the type declaration file and/or its quality, style, or process
- [ ] we'd like to try out flow or some other system
- [ ] adding this is too low priority compared with other things on the roadmap
- [ ] this belongs outside pdf.js proper, perhaps it should go in @types/pdfjs-dist
- [ ] other?

I might submit this to @types/pdfjs-dist instead if there is no further response.

timvandermeij:
@yurydelendik Could you chime in for the above?

timvandermeij:
I looked at this again and I still have concerns regarding maintenance, mainly because PDF.js itself does not use any TypeScript and keeping this in sync with the code will be challenging. Therefore, to answer your question above, I personally think it's better to place it in `@types/pdfjs-dist` since that contains type definitions for a lot of other libraries too and it seems that people are used to having their type definitions for non-TypeScript projects there.

brendandahl:
Sorry for the delay, Yury and I haven't had much time to work on pdf.js lately. We discussed this a bit recently though and we'd like to accept this and then work on removing duplicate comments and letting the typescript definitions be the documentation for the public API. I think we can accept as is, then do the cleanup.

acchou:
Sounds good. Let me know if I can help with any cleanup here.

FraserKillip:
Sorry to be that guy.. but any update on the merge for this? The current @types are completely wrong and it makes working with this lib very hard :/

timvandermeij:
I'm also fine with merging this after the 2.0 stable release, which should happen soon, if it's updated as we have added a few more annotation types in the meantime. Since this doesn't touch any core files, I'm also fine with doing even more cleanup at a later stage.

acchou:
If you can point me in the general direction of where these new annotation types are located, I could update the type annotations.

external/types/shared.d.ts
```diff
@@ -0,0 +1,531 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+interface HTMLAnchorElement {}
+interface ReadableStream {}
+
+declare namespace _pdfjs {
+    //
```
timvandermeij: Indenting in this file should be two spaces, just like the other files.

external/types/core.d.ts
```diff
@@ -0,0 +1,227 @@
...
+    annotationType: AnnotationType.SQUARE;
+  }
+  interface CircleAnnotation extends Annotation {
+    annotationType: AnnotationType.CIRCLE;
+  }
+  interface PolylineAnnotation extends Annotation {
+    annotationType: AnnotationType.POLYLINE;
+    vertices: Vertex[];
+  }
+  interface PolygonAnnotation extends Annotation {
```
timvandermeij: The annotation layer changes are in https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js (core) and https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js (display). I think `InkAnnotation` is missing, but maybe there is more.

acchou: Done, see b74c5fd.

swftvsn:
Hi, sorry to be that guy again but, any update on this? @timvandermeij @acchou ?

timvandermeij:
I'm still a bit on the fence here. Even though this PR looks good, there is an alternative in https://github.com/mozilla/pdf.js/pull/10575 that aims to generate the TypeScript files from type annotations in the docstrings. To me this seems much more maintainable since we don't have to keep multiple files in sync. What do you think? I think providing TypeScript definitions is good, but since it isn't our main focus it should really be easy to maintain.

brendandahl:
I was recently made aware of typescripts `--checkJs` mode which works off of jsdoc comments. If we can get that working on travis, I think that's the path going forward.

burtonator:
May I recommend just moving forward with even a manually updated typescript definition? 

This ticket was updated 1y10m ago... This is REALLY needed for external developers because without it the API is very very very opaque.  

Even just copying @types/nodejs-dist over and running with that would be good because we can manually merge the types in the mean time as they evolve.  

Sending a 1 line PR is a lot easier than trying to build an auto-generating system.  I can't imagine the core API changes that much in 1y10m where we couldn't have updated it manually.

brendandahl:
I wasn't hoping to build an auto generating system. I was thinking we could have all our typescript definition files and then use things like `@implements SomeInterface`. However, I played around more with `checkJs` and it's not as powerful as I was hoping, as it does not support interfaces and [doesn't look like it ever will](https://github.com/Microsoft/TypeScript/issues/16142).

tamuratak:
If the approach of #10575 is preferred,  what prevents #10575 to be merged? To my [testing](https://github.com/tamuratak/pdf.js/tree/o_dts), rebasing the branch of  #10575 to the current master is not difficult. 

With `npx gulp dts`, it generates `build/dts/types.d.ts`.

wojtekmaj:
If there are no TS experts in PDF.js team I'd suggest to move these typings to e.g. definitely-typed. Otherwise, we will have a hard time maintaining these typings, with not many people willing & able to review the changes when necessary. Also, this would require contributors to know TypeScript whenever any API change is done.

acchou:
Agreed, that seems to be the right thing to do at this point.

One thing before I do that though... has anyone tried to use the new --declaration flag with --allowJs (https://www.typescriptlang.org/docs/handbook/release-notes/overview.html#--declaration-and---allowjs)? This seems to allow the typescript compiler to natively generate declaration files from jsdoc annotated javascript source. This seems to be the preferred approach for the team from PR #10575.

timvandermeij:
@acchou I don't think anyone tried that yet, but it would be good to look into. I think we've come to the conclusion that manually maintaining the TypeScript bindings here is just not feasible and I agree with @wojtekmaj's analysis here. Let's close this PR and focus our efforts either on getting #10575 landed. We have already merged most of the documentation improvements from that patch in separate PRs, so we can either choose to rebase that PR or make a new PR with the solution where TypeScript natively generates the declaration files from the JSDoc comments. In any case, if we are to provide the bindings, they must be generated in some form to avoid a maintenance burden. Thank you!

PR #9379, opened 2018-01-18
[api-minor] Extract the Content-Disposition filename

Snuffleupagus:
This PR includes the patch, rebased against the latest `master`, from PR #8688 (obviously maintaining correct author info), together with a follow-up patch to address some final comments to keep all `IPDFStreamReader` implementations consistent.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -154,7 +154,7 @@ let PDFViewerApplication = {
   baseUrl: '',
   externalServices: DefaultExternalServices,
   _boundEvents: {},
-  contentDispositionFileName: null,
+  contentDispositionFilename: null,
 
   // Called once when the document is loaded.
   initialize(appConfig) {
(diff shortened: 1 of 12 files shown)


web/app.js
```diff
@@ -1178,7 +1183,10 @@ let PDFViewerApplication = {
       }
 
       if (pdfTitle) {
-        this.setTitle(pdfTitle + ' - ' + document.title);
+        this.setTitle(
+          `${pdfTitle} - ${contentDispositionFilename || document.title}`);
```
yurydelendik: `document.title` contains HTML viewer title. I think we just want to see following instead this patch:

    if (!pdfTitle && contentDispositionFilename) pdfTitle = contentDispositionFilename;

yurydelendik (approved):
Looks good to me.

web/app.js
yurydelendik: Actually, it's more complicated than I tought, having `document.title` to hold application variable is totally confusing. The change is good.

timvandermeij:
Thank you!

PR #7554, opened 2016-08-20
Support for Content-Disposition

anoopelias:
Fixes #6396 

Some notes for the reviewer,
1. Copied considerable part of the code (both implementation and tests) from [content-disposition](https://www.npmjs.com/package/content-disposition) npm module. I couldn't find an easy way to reuse a node module in browser here.
2. Behavior of parsing content-disposition has been modified from the corresponding node module here in pdf.js for the below case,
   - If there is an invalid encoding in utf8, the npm module will put a replacement character (`\ufffd`) and parses the rest. However, here in pdf.js, we will throw an error. The reason being, in npm module it uses Buffer class while here we use `decodeURIComponent(escape(str));` (Throwing any error during parsing will cause pdf.js to just ignore the entire header.)
3. Only unit tests are checked and updated. I couldn't get the other tests running because of the mismatch in hashes (I think).

Please let me know any feedback.

Rgds,

PR #9383, opened 2018-01-19
Better content disposition parser

Rob--W:
The current implementation of the filename-from-content-disposition logic from #9379 is

https://github.com/mozilla/pdf.js/blob/75dc2bbd359990ebb1c1484f204acf66c3cb8221/src/display/network_utils.js#L59-L62
  
and it has several issues. Examples where the method gives the wrong answer:

- `attachment; filename=file.pdf` (failing to detect file.pdf because of missing quotes)
- `attachment; filename="foo.pdf"; filename*=UTF-8''bar.pdf` (the result is `foo.pdf"; filename*=UTF-8''bar.pdf` because of the greedy `.+`)
- `attachment; filename*=UTF-8''file.pdf'` (no result because `filename*=` from RFC 6266 / RFC 5987 is not supported)

This PR uses a Content-Disposition parser that I have written and tested before (at https://github.com/Rob--W/open-in-browser/commit/6f3bbb8bbfc1e3e943200fffdb68d35075e82ddd).

### DIFF
--- a/src/display/network_utils.js
+++ b/src/display/network_utils.js
@@ -16,7 +16,9 @@
 import {
   assert, MissingPDFException, UnexpectedResponseException
 } from '../shared/util';
-import { getFilenameFromUrl } from './dom_utils';
+import {
+  getFilenameFromContentDispositionHeader
+} from './content_disposition';
 
 function validateRangeRequestCapabilities({ getResponseHeader, isHttp,
                                             rangeChunkSize, disableRange, }) {
@@ -56,10 +58,9 @@ function validateRangeRequestCapabilities({ getResponseHeader, isHttp,
 function extractFilenameFromHeader(getResponseHeader) {
   const contentDisposition = getResponseHeader('Content-Disposition');
   if (contentDisposition) {
-    let parts =
-      /.+;\s*filename=(?:'|")(.+\.pdf)(?:'|")/gi.exec(contentDisposition);
-    if (parts !== null && parts.length > 1) {
-      return getFilenameFromUrl(parts[1]);
+    let filename = getFilenameFromContentDispositionHeader(contentDisposition);
+    if (/\.pdf$/i.test(filename)) {
+      return filename;
     }
   }
   return null;
--- a/.eslintrc
+++ b/.eslintrc
@@ -120,7 +120,10 @@
     "keyword-spacing": ["error", { "before": true, "after": true, }],
     "linebreak-style": ["error", "unix"],
     "lines-between-class-members": ["error", "always"],
-    "max-len": ["error", 80],
+    "max-len": ["error", {
+      "code": 80,
+      "ignoreUrls": true
+    }],
     "new-cap": ["error", { "newIsCap": true, "capIsNew": false, }],
     "new-parens": "error",
     "no-array-constructor": "error",
--- a/src/display/content_disposition.js
+++ b/src/display/content_disposition.js
@@ -1,0 +1,211 @@
+/* Copyright 2017 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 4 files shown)


src/display/network_utils.js
```diff
@@ -56,15 +55,199 @@ function validateRangeRequestCapabilities({ getResponseHeader, isHttp,
...
 
+// The logic in getFilenameFromContentDispositionHeader is adapted from
+// https://github.com/Rob--W/open-in-browser/blob/9f5fcae11cf6d99c503a15894f22efdfcd2075b7/extension/content-disposition.js
+// with the following changes:
+// - modified to conform to PDF.js's coding style.
+// - Support UTF-8 decoding when TextDecoder is unsupported.
+// - Replace "startsWith" with other string method for compat with old browsers.
+// - Move return to the end of the function to prevent Babel from dropping the
+//   function declarations.
+function getFilenameFromContentDispositionHeader(contentDisposition) {
```
Snuffleupagus: Since this function is basically "external" code, similar to other imported code (e.g. streams/webL10n), can't we just place it in the [external/ directory](https://github.com/mozilla/pdf.js/tree/master/external) instead, e.g. `external/content-disposition/` or similar?

Snuffleupagus (approved):
> The current implementation of the filename-from-content-disposition logic from #9379 [...] has several issues.

For the initial implementation, I think that we went with a purposely simple implementation given e.g. then comment at the end of https://github.com/mozilla/pdf.js/pull/7554#issuecomment-241726667. 

However, considering that the existing code apparently gets it wrong in some cases and that the implementation here seems pretty clean, I think that this patch seems reasonable.

r=me, with the inline question addressed and with all tests passing. Thank you!

src/display/network_utils.js
Rob--W: The `external/` directory doesn't seem to be used for direct inclusion of code, except for the `streams` library. Even in this case, there is extra code in `gulpfile.js` to manage the inclusion.

If you prefer a separate file, I can put it in a new file content-disposition.js.

The helper functions need to be in the scope of the main function, because the `needsEncodingFixup` variable needs to be updated:
https://github.com/Rob--W/pdf.js/blob/a59a81ad1fe0041f4cb45b4477079193f917b190/src/display/network_utils.js#L75
If I had to refactor the code to move the maintenance of this variable to a separate function parameter and/or return value, the code would look more messier. I would then have to pass an extra parameter to `textdecode` (and all of its callers), and either update this parameter (i.e. box the boolean in an object), or change the return value to return a pair (the boolean and the original return value).

Given this explanation, do you want any more changes before merging?

Snuffleupagus: > The helper functions need to be in the scope of the main function, because the needsEncodingFixup variable needs to be updated:

I overlooked that, sorry!

> If you prefer a separate file, I can put it in a new file content-disposition.js.
> [...]
> Given this explanation, do you want any more changes before merging?

I do like this idea, so please let's move the new code (verbatim) to a new `src/display/content-disposition.js` file; and then run the full test-suite before landing.

Rob--W: Done. `gulp unittestcli` passes locally.

`unittestcli` is also run by Travis, so if the latest comit has a green checkmark, you can merge it.

Snuffleupagus: Couldn't the "Improve correctness of Content-Disposition parser" and "Move content disposition logic to separate file" commits be merged, to reduce the churn in the affected code a bit?

Rob--W: Done.

vitaliy-patsay:
Hi there, thanks for the parser. However I have a question - why do you expect filenames don't have whitespaces?

Rob--W:
@vitaliy-patsay Thanks for the question. It is an oversight and I have submitted a fix - see #9435.

vitaliy-patsay:
@Rob--W  Thanks Rob!

PR #9385, opened 2018-01-20
[api-major] Remove the `SINGLE_FILE` build target and the `PDFJS.disableWorker` option

Snuffleupagus:
*Please refer to the commit messages for additional details.*

This PR is an initial attempt at implementing https://mozilla.logbot.info/pdfjs/20180102#c14068811-c14068833; feedback welcome!

 - [x] Update the examples. Given the performance implications of loading the worker file in the main-thread, do we actually want to mention this in the examples or should we simply not advertise it at all?

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1233,7 +1233,6 @@ var PDFWorker = (function PDFWorkerClosure() {
 
   // Loads worker code into main thread.
   function setupFakeWorkerGlobal() {
-    var WorkerMessageHandler;
     if (fakeWorkerFilesLoadedCapability) {
       return fakeWorkerFilesLoadedCapability.promise;
     }
@@ -1245,26 +1244,24 @@ var PDFWorker = (function PDFWorkerClosure() {
       fakeWorkerFilesLoadedCapability.resolve(mainWorkerMessageHandler);
       return fakeWorkerFilesLoadedCapability.promise;
     }
-    // In the developer build load worker_loader which in turn loads all the
+    // In the developer build load worker_loader.js which in turn loads all the
     // other files and resolves the promise. In production only the
     // pdf.worker.js file is needed.
     if (typeof PDFJSDev === 'undefined' || !PDFJSDev.test('PRODUCTION')) {
       if (typeof SystemJS === 'object') {
         SystemJS.import('pdfjs/core/worker').then((worker) => {
-          WorkerMessageHandler = worker.WorkerMessageHandler;
-          fakeWorkerFilesLoadedCapability.resolve(WorkerMessageHandler);
+          fakeWorkerFilesLoadedCapability.resolve(worker.WorkerMessageHandler);
         });
       } else if (typeof require === 'function') {
-        var worker = require('../core/worker.js');
-        WorkerMessageHandler = worker.WorkerMessageHandler;
-        fakeWorkerFilesLoadedCapability.resolve(WorkerMessageHandler);
+        let worker = require('../core/worker.js');
+        fakeWorkerFilesLoadedCapability.resolve(worker.WorkerMessageHandler);
       } else {
         throw new Error(
           'SystemJS or CommonJS must be used to load fake worker.');
       }
     } else {
-      var loader = fakeWorkerFilesLoader || function (callback) {
-        Util.loadScript(getWorkerSrc(), function () {
+      let loader = fakeWorkerFilesLoader || function(callback) {
+        Util.loadScript(getWorkerSrc(), function() {
           callback(window.pdfjsDistBuildPdfWorker.WorkerMessageHandler);
         });
       };
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1216,6 +1216,19 @@ var PDFWorker = (function PDFWorkerClosure() {
     throw new Error('No PDFJS.workerSrc specified');
   }
 
+  function getMainThreadWorkerMessageHandler() {
+    if (typeof window === 'undefined') {
+      return null;
+    }
+    if (typeof PDFJSDev === 'undefined' || !PDFJSDev.test('PRODUCTION')) {
+      return (window.pdfjsNonProductionPdfWorker &&
+              window.pdfjsNonProductionPdfWorker.WorkerMessageHandler);
+    }
+    // PRODUCTION
+    return (window.pdfjsDistBuildPdfWorker &&
+            window.pdfjsDistBuildPdfWorker.WorkerMessageHandler);
+  }
+
   let fakeWorkerFilesLoadedCapability;
 
   // Loads worker code into main thread.
@@ -1225,6 +1238,13 @@ var PDFWorker = (function PDFWorkerClosure() {
       return fakeWorkerFilesLoadedCapability.promise;
     }
     fakeWorkerFilesLoadedCapability = createPromiseCapability();
+
+    let mainWorkerMessageHandler = getMainThreadWorkerMessageHandler();
+    if (mainWorkerMessageHandler) {
+      // The worker was already loaded using a `<script>` tag.
+      fakeWorkerFilesLoadedCapability.resolve(mainWorkerMessageHandler);
+      return fakeWorkerFilesLoadedCapability.promise;
+    }
     // In the developer build load worker_loader which in turn loads all the
     // other files and resolves the promise. In production only the
     // pdf.worker.js file is needed.
@@ -1316,7 +1336,7 @@ var PDFWorker = (function PDFWorkerClosure() {
       // Right now, the requirement is, that an Uint8Array is still an
       // Uint8Array as it arrives on the worker. (Chrome added this with v.15.)
       if (typeof Worker !== 'undefined' && !isWorkerDisabled &&
-          !getDefaultSetting('disableWorker')) {
+          !getMainThreadWorkerMessageHandler()) {
         var workerSrc = getWorkerSrc();
 
         try {
@@ -1427,7 +1447,7 @@ var PDFWorker = (function PDFWorkerClosure() {
     },
 
     _setupFakeWorker: function PDFWorker_setupFakeWorker() {
-      if (!isWorkerDisabled && !getDefaultSetting('disableWorker')) {
+      if (!isWorkerDisabled) {
         warn('Setting up fake worker.');
         isWorkerDisabled = true;
       }
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -46,8 +46,7 @@ var pdfjsFilePath =
 
 var fakeWorkerFilesLoader = null;
 var useRequireEnsure = false;
-if (typeof PDFJSDev !== 'undefined' &&
-    PDFJSDev.test('GENERIC && !SINGLE_FILE')) {
+if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
   // For GENERIC build we need add support of different fake file loaders
   // for different  frameworks.
   if (typeof window === 'undefined') {
@@ -1243,10 +1242,6 @@ var PDFWorker = (function PDFWorkerClosure() {
         throw new Error(
           'SystemJS or CommonJS must be used to load fake worker.');
       }
-    } else if (PDFJSDev.test('SINGLE_FILE')) {
-      var pdfjsCoreWorker = require('../core/worker.js');
-      WorkerMessageHandler = pdfjsCoreWorker.WorkerMessageHandler;
-      fakeWorkerFilesLoadedCapability.resolve(WorkerMessageHandler);
     } else {
       var loader = fakeWorkerFilesLoader || function (callback) {
         Util.loadScript(getWorkerSrc(), function () {
@@ -1320,9 +1315,8 @@ var PDFWorker = (function PDFWorkerClosure() {
       // all requirements to run parts of pdf.js in a web worker.
       // Right now, the requirement is, that an Uint8Array is still an
       // Uint8Array as it arrives on the worker. (Chrome added this with v.15.)
-      if ((typeof PDFJSDev === 'undefined' || !PDFJSDev.test('SINGLE_FILE')) &&
-          !isWorkerDisabled && !getDefaultSetting('disableWorker') &&
-          typeof Worker !== 'undefined') {
+      if (typeof Worker !== 'undefined' && !isWorkerDisabled &&
+          !getDefaultSetting('disableWorker')) {
         var workerSrc = getWorkerSrc();
 
         try {
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1485,6 +1485,10 @@ var PDFWorker = (function PDFWorkerClosure() {
     return new PDFWorker(null, port);
   };
 
+  PDFWorker.getWorkerSrc = function() {
+    return getWorkerSrc();
+  };
+
   return PDFWorker;
 })();
 
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -24,7 +24,7 @@ import {
   DOMCanvasFactory, RenderingCancelledException
 } from '../../src/display/dom_utils';
 import {
-  getDocument, PDFDocumentProxy, PDFPageProxy
+  getDocument, PDFDocumentProxy, PDFPageProxy, PDFWorker
 } from '../../src/display/api';
 import isNodeJS from '../../src/shared/is_node';
 import { PDFJS } from '../../src/display/global';
@@ -404,6 +404,11 @@ describe('api', function() {
       }).catch(function (reason) {
         done.fail(reason);
       });
+    });
+    it('gets current workerSrc', function() {
+      let workerSrc = PDFWorker.getWorkerSrc();
+      expect(typeof workerSrc).toEqual('string');
+      expect(workerSrc).toEqual(PDFJS.workerSrc);
     });
   });
   describe('PDFDocument', function() {
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -122,16 +122,14 @@ describe('api', function() {
             path: TEST_PDFS_PATH.node + basicApiFileName,
           });
         } else {
-          var nonBinaryRequest = PDFJS.disableWorker;
-          var request = new XMLHttpRequest();
+          let nonBinaryRequest = false;
+          let request = new XMLHttpRequest();
           request.open('GET', TEST_PDFS_PATH.dom + basicApiFileName, false);
-          if (!nonBinaryRequest) {
-            try {
-              request.responseType = 'arraybuffer';
-              nonBinaryRequest = request.responseType !== 'arraybuffer';
-            } catch (e) {
-              nonBinaryRequest = true;
-            }
+          try {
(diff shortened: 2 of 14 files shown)


web/app.js
```diff
@@ -1512,6 +1513,36 @@ if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
...
+      script.src = PDFJS.workerSrc;
+      script.onload = function() {
+        resolve();
+      };
+      script.onerror = function () {
+        reject(new Error(`Cannot load fake worker at: ${script.src}`));
+      };
+      (document.getElementsByTagName('head')[0] || document.body).
+        appendChild(script);
+      }
```
timvandermeij: Nit: this bracket is misindented.

web/app.js
```diff
@@ -1512,6 +1513,36 @@ if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
...
+        reject(new Error(
+          'SystemJS or CommonJS must be used to load fake worker.'));
+      }
+    } else {
+      let script = document.createElement('script');
+      script.src = PDFJS.workerSrc;
+      script.onload = function() {
+        resolve();
+      };
+      script.onerror = function () {
```
timvandermeij: Nit: remove space after `function` to be consistent with the one above.

timvandermeij (commented):
This patch looks really good! Aside from the two minor comments below, I think that we should indeed not advertise disabling workers in the examples.

web/app.js
Snuffleupagus: Whoops, too much copy-pasting of existing code here (and above as well); good eye!

web/app.js
```diff
@@ -1512,6 +1513,36 @@ if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
...
+    } else {
+      let script = document.createElement('script');
+      script.src = PDFJS.workerSrc;
+      script.onload = function() {
+        resolve();
+      };
+      script.onerror = function() {
+        reject(new Error(`Cannot load fake worker at: ${script.src}`));
+      };
+      (document.getElementsByTagName('head')[0] || document.body).
```
Rob--W: This can be simplified to `document.head || document.body`. Only use `document.head` is only undefined in IE8-, and supported by all other browsers: https://developer.mozilla.org/en-US/docs/Web/API/Document/head

I usually do `document.head || document.documentElement` (since `<head>` should exist; if it does not, then `document.body` may also be empty, but the root element should always exist, so that's a safe fallback).

web/chromecom.js
```diff
@@ -65,7 +64,9 @@ let ChromeCom = {
     file = file.replace(/^drive:/i,
         'filesystem:' + location.origin + '/external/');
 
-    if (/^filesystem:/.test(file) && !PDFJS.disableWorker) {
+    const disableWorker = (window.pdfjsDistBuildPdfWorker &&
```
Rob--W: Under which circumstances is `window.pdfjsDistBuildPdfWorker` unset? When the worker script is loaded in the main thread, right? Is it possible for the code to reach this place before the worker script has had a chance to initiate?

In either case, please add a brief comment here to explain what this line is doing..

Rob--W (commented):
Just a question, how am I supposed to test PDF.js without workers in the Chrome extension build target?

web/app.js
Snuffleupagus: I was just copying the existing pattern used in https://github.com/mozilla/pdf.js/blob/master/web/app.js#L1515-L1534; should we simply change both them?

Rob--W: You can do that in a separate commit (to keep the refactoring separate from the actual feature).

web/chromecom.js
Snuffleupagus: > Under which circumstances is window.pdfjsDistBuildPdfWorker unset? When the worker script is loaded in the main thread, right?

It's unset by default, i.e. when workers *are* being used. Hence it should only ever be defined when the `pdf.worker.js` file is loaded on the main-thread.

> Is it possible for the code to reach this place before the worker script has had a chance to initiate?

That shouldn't happen, since the `resolvePDFFile` function is being called from
https://github.com/mozilla/pdf.js/blob/fe5102a27fac7235e14334664cfc7e6dcaa04a92/web/chromecom.js#L353-L355
which in turn is called from https://github.com/mozilla/pdf.js/blob/fe5102a27fac7235e14334664cfc7e6dcaa04a92/web/app.js#L591-L596
which is called from
https://github.com/mozilla/pdf.js/blob/fe5102a27fac7235e14334664cfc7e6dcaa04a92/web/app.js#L1652-L1654
`webViewerOpenFileViaURL` is then called from
https://github.com/mozilla/pdf.js/blob/fe5102a27fac7235e14334664cfc7e6dcaa04a92/web/app.js#L1536
which is only *after* the viewer initialization code has run, see
https://github.com/mozilla/pdf.js/blob/fe5102a27fac7235e14334664cfc7e6dcaa04a92/web/app.js#L502-L504

Finally, note that we always wait on the hash parameters (among other things) to be initialized in `PDFViewerApplication.initialize`
https://github.com/mozilla/pdf.js/blob/fe5102a27fac7235e14334664cfc7e6dcaa04a92/web/app.js#L166-L172

---

> In either case, please add a brief comment here to explain what this line is doing..

Sure, will do.

*Ninja-edit:* Actually, do we still need this code-path?
This code seem to have been originally added in PR #4598, but the referenced upstream bug http://crbug.com/362061 seem to have been fixed over 3 years ago.
I'd thus assume that the affected versions of Chrome is no longer supported since quite some time, so can we (in a separate PR) remove this code-path and thus not have to worry about it here?

Snuffleupagus:
> how am I supposed to test PDF.js without workers in the Chrome extension build target?

I don't know exactly how you're currently testing this configuration, so the following might be too simple, but wouldn't it be possible to do something like the following:
Set the `pdfBugEnabled` preference, and then append `disableWorker=true` to the viewer (or PDF) URL.

web/chromecom.js
Rob--W: I try to not break backcompat unless necessary. I should take a look at my telemetry (https://github.com/Rob--W/pdfjs-telemetry) to see whether it is viable to rip out code that supports old Chrome versions (there is lots of other code besides this line).

Rob--W:
> Set the `pdfBugEnabled` preference, and then append `#disableWorker=true` to the viewer (or PDF) URL.

I used to use the `#disableWorker=true` parameter, but when that was not supported by default I just put a breakpoint at the end of the library and then set `PDFJS.disableWorker=true` . Your suggestion looks a bit neater, thanks for that.

web/app.js
Snuffleupagus: > I usually do `document.head || document.documentElement`

This suggesting has now been implemented (only in the newly added code); thank you!

Snuffleupagus: > I usually do `document.head || document.documentElement`

This suggestion has now been implemented (only in the newly added code); thank you!

web/chromecom.js
Snuffleupagus: A comment has now been added here.

Snuffleupagus:
@yurydelendik Since this PR attempts to implement the idea you proposed at https://mozilla.logbot.info/pdfjs/20180102#c14068811-c14068833, do you have time to review (or provide feedback) here?

wojtekmaj:
I'm an author of React-PDF and I'm concerned about this change.

How does it impact users using Webpack, who don't want to use worker? Previously,  I just set up disableWorker=true option and pdf.js requested pdf.worker.js via require() and created a fake worker. How that procedure will change? Will I need to require the worker file myself to disable worker?

Rob--W:
@wojtekmaj Your current use case of `PDFJS.disableWorker` has a bad impact on performance/UX. You traded performance (something that is important to users) for ease of code distribution (which is only convenient for developers).

I guess that it's technically possible to serialize the worker script and then use `URL.createObjectURL` to load a worker from a single file, but that may be broken by overly aggressive code optimizers (e.g. by having moved a function outside of the scope of the serialized function). It may also break when a strict Content Security Policy is enforced.

You should keep the PDF.js worker in a separate file, and assign the location of it to `PDFJS.workerSrc` (if it is at a non-standard location):

https://github.com/mozilla/pdf.js/blob/b6c57d90887c1a936e2638e83cf88576aea8ca36/src/display/global.js#L140-L148

The logic of worker script resolution is here:
https://github.com/mozilla/pdf.js/blob/b6c57d90887c1a936e2638e83cf88576aea8ca36/src/display/api.js#L41-L45
https://github.com/mozilla/pdf.js/blob/b6c57d90887c1a936e2638e83cf88576aea8ca36/src/display/api.js#L1205-L1218

wojtekmaj:
Thanks for information @Rob--W. I actually realize the benefits of service worker here and I discourage my users from using a version without worker, but I do provide them with such ability as for many beginners setting it up is an huge problem. I did struggle a lot with this myself.

Also, sometimes bundle size matters. Currently how PDF.js plays with Webpack is a mess. Sadly, unless provided with additional configuration (which I failed to figure out), Webpack includes pdf.worker.js code **twice** because of using two separate loaders on them - it's required via worker-loader in webpack.js entry file, and then *again* without using worker-loader in pdf.js itself. Because of using separate loaders Webpack is not smart enough to dedupe them.

Rob--W:
The code changes LGTM.

The code path for fake workers seems to not be covered by unit tests. Wouldn't it make sense to add a single test that customized `PDFJS.workerSrc` and test whether the fake worker can be loaded? Or will we rely on manual tests only for this case?

Snuffleupagus:
> The code path for fake workers seems to not be covered by unit tests.

Given that workers are always disabled in Node.js, we should have *implicit* fake worker test-coverage thanks to the (subset of) API unit-tests running on Travis.
https://github.com/mozilla/pdf.js/blob/fd242ad2c2191f5de32f6b158e15db8e012f9467/src/display/api.js#L54-L55
Slightly off topic for this PR, but we should see if some of the currently `pending` API tests could be enabled on Node.js/Travis since we now have [node_stream.js](https://github.com/mozilla/pdf.js/blob/master/src/display/node_stream.js) available.

> Wouldn't it make sense to add a single test that customized `PDFJS.workerSrc` and test whether the fake worker can be loaded?

While we might be able to add a Node.js-only explicit test for that, I'm not sure that we'd be able to easily test this when the unit-tests run normally (i.e. in browser) without risk of negatively impacting subsequent tests.
The reason being that once we've setup the fake workers, we'll no longer attempt to use real workers any more; refer to
https://github.com/mozilla/pdf.js/blob/fd242ad2c2191f5de32f6b158e15db8e012f9467/src/display/api.js#L1436-L1439
and
https://github.com/mozilla/pdf.js/blob/fd242ad2c2191f5de32f6b158e15db8e012f9467/src/display/api.js#L1323-L1325

> Or will we rely on manual tests only for this case?

Given the above, I'm slightly on the fence about adding a fake worker test here!

Rob--W:
> > The code path for fake workers seems to not be covered by unit tests.
> 
> Given that workers are always disabled in Node.js, we should have implicit fake worker test-coverage thanks to the (subset of) API unit-tests running on Travis:

I am referring to the injected script tag in your new `loadFakeWorker` function from this PR. There is probably no test coverage for that.

> Given the above, I'm slightly on the fence about trying to add a fake worker test here! Opinions?

How about a custom parameter to disable workers in the unit test? It doesn't even need to be a PDF.js API, it could also be something that does `window.Worker = null`.

Snuffleupagus:
> How about a custom parameter to disable workers in the unit test? It doesn't even need to be a PDF.js API, it could also be something that does window.Worker = null.

Sorry, but I really don't understand how this would help. As mentioned above, as soon as [`PDFWorker._setupFakeWorker`](https://github.com/mozilla/pdf.js/blob/fd242ad2c2191f5de32f6b158e15db8e012f9467/src/display/api.js#L1435) has executed *once* then workers become permanently disabled.

Rob--W:
> > How about a custom parameter to disable workers in the unit test? It doesn't even need to be a PDF.js API, it could also be something that does window.Worker = null.
>
> Sorry, but I really don't understand how this would help. As mentioned above, as soon as `PDFWorker._setupFakeWorker` has executed once then workers become permanently disabled.

When I wrote that, I had in mind that it would not be run by default until someone manually edits the URL and adds something like `&disableworker=true`. But I guess that since it is already a manual test, one can as well add a breakpoint with a debugger and manually do `window.Worker = null` for testing. Or enable debugging in the viewer and add `&disableworker=true` and see if the viewer works. So let's ignore my previous request.

One last thing: The `Worker` code path uses `getWorkerSrc()` whereas your newly introduced `<script>` tag uses `PDFJS.workerSrc` only. This could potentially result in an unexpected behavioral difference, where the Worker loads successfully, but the non-worker fails. I'll defer to your judgement on whether this should be fixed.

Snuffleupagus:
> The Worker code path uses getWorkerSrc() whereas your newly introduced <script> tag uses PDFJS.workerSrc only. This could potentially result in an unexpected behavioral difference, where the Worker loads successfully, but the non-worker fails.

That's a very good catch, thanks for spotting this!
I agree that this inconsistency seem bad, so I've addressed it by adding a new method to `PDFWorker` that allows (indirect) access to the internal `getWorkerSrc` helper function.

src/display/api.js
```diff
@@ -1491,6 +1502,14 @@ var PDFWorker = (function PDFWorkerClosure() {
     return new PDFWorker(null, port);
   };
 
+  PDFWorker.getWorkerSrc = function() {
+    try {
```
Rob--W: Why try-catch? `getWorkerSrc` is not expected to throw (unless there is no way to determine the worker source.

test/unit/api_spec.js
```diff
@@ -405,6 +403,12 @@ describe('api', function() {
         done.fail(reason);
       });
     });
+    it('gets current workerSrc', function(done) {
+      let workerSrc = PDFWorker.getWorkerSrc();
+      expect(typeof workerSrc).toEqual('string');
+      expect(workerSrc).toEqual(PDFJS.workerSrc);
+      done();
```
Rob--W: Since the test is synchronous, remove `done();` and `done` in the `function(done){`. `done` is only needed for async tests.

Rob--W (changes requested):

src/display/api.js
Snuffleupagus: While you're correct that it *shouldn't* throw, than may still happen in some cases [given this code](https://github.com/mozilla/pdf.js/blob/55e3f97aa97ad648ef955203524bbdb5593671d1/src/display/api.js#L1217)

In e.g. Node.js, where workers are disabled by default and `workerSrc` isn't set in the examples, if that function is called a (perhaps unexpected) error would be thrown. With the `getWorkerSrc` function exposed to the outside, I simply wanted to avoid "surprising" anyone using this function.

Snuffleupagus: While you're correct that it *shouldn't* throw, than may still happen in some cases [given this code](https://github.com/mozilla/pdf.js/blob/55e3f97aa97ad648ef955203524bbdb5593671d1/src/display/api.js#L1217).

In e.g. Node.js, where workers are disabled by default and `workerSrc` isn't set in the examples, if that function is called a (perhaps unexpected) error would be thrown.
With the `getWorkerSrc` function now exposed to the outside, it seemed nice to avoid "surprises" for an API consumer who uses that function.

test/unit/api_spec.js
Snuffleupagus: Done :-)

src/display/api.js
Snuffleupagus: While you're correct that it *shouldn't* throw, that could still happen [given this code](https://github.com/mozilla/pdf.js/blob/55e3f97aa97ad648ef955203524bbdb5593671d1/src/display/api.js#L1217).
Hence it seemed nice to try and avoid surprises for an API consumer that uses this function not expecting an `Error` to be thrown. Are you saying that you think this is a completely unwarranted concern, and that should return the internal `getWorkerSrc` value as-is?

Rob--W: Yes, I suggest to not catch the error. `workerSrc` doesn't make sense in Node.js, and when it does, we have to provide a suitable implementation instead of silently returning nothing.

Snuffleupagus: I've make the requested changes; thanks for the review comments!

Rob--W (approved):

Snuffleupagus:
@yurydelendik Ping; since this PR implements your idea from https://mozilla.logbot.info/pdfjs/20180102#c14068811-c14068833, did you want to look at it before we land this?

Kerumen:
@Snuffleupagus This PR was released on `pdfjs-dist` as a patch update ([v2.0.305](https://github.com/mozilla/pdfjs-dist/tree/v2.0.305/build) to [v2.0.310](https://github.com/mozilla/pdfjs-dist/tree/v2.0.310/build)).
Unfortunately, it broke the lib react-pdf (https://github.com/wojtekmaj/react-pdf/issues/149) and maybe others. It was tagged as `api-major`, why did you release this as a patch update?

Snuffleupagus:
*Regarding https://github.com/mozilla/pdf.js/pull/9385#issuecomment-363030784:*

We're currently working towards the release of PDF.js version `2.0`; that work is being tracked in https://github.com/mozilla/pdf.js/projects/5. As can be seen, there's still a few outstanding things to address, which is why it's not officially released yet; the official releases are always available at https://github.com/mozilla/pdf.js/releases.

Since we wanted to be able to refactor/cleanup old code, the *major* version number was thus increased to signal that there's changes that are by design *not* backwards compatibility.
Hence upgrading from PDF.js version `1.x.y` to version `2.x.y` cannot be expected to be a drop-in replacement, without suitable adjustments to the code using the PDF.js library.

In closing, the problem seem to be twofold:
First of all, users of `pdfjs-dist` didn't pin the version number to `1.x.y` until PDF.js version `2.0` has been officially released. Secondly, some users (might have) overlooked the significance of the increased *major* version number.

Finally, the actual breaking change in `pdfjs-dist` was thus a lot earlier than stated above, more specifically in https://github.com/mozilla/pdfjs-dist/tree/v2.0.87.

Kerumen:
@Snuffleupagus I don't understand. The breaking change (`pdf.combined.js` removed) was from `2.0.305` to `2.0.310`. I was already using the branch `2.x` without problems.

wojtekmaj:
I think the misunderstanding is that usually on GitHub if something is pre-release it's marked as so, e.g. 2.0.305-beta or something like that. Mozilla just throws pdfjs-dist on npm "as is" with every merge, without e.g. releasing using a tag "next" or similar. So, if you take pdfjs-dist@latest you are actually getting the beta without knowing about it.

Snuffleupagus:
> I was already using the branch 2.x without problems.

That can probably be attributed to pure luck, given how you're currently using/integrating PDF.js in your application, rather than anything else.
As mentioned, PDF.js version `2.0` will be a breaking change w.r.t. to the API, and shouldn't be considered final until released officially at https://github.com/mozilla/pdf.js/releases.

Rob--W:
@wojtekmaj Your analysis is correct, we should improve the documentation. I filed #9440 for that.

PR #9401, opened 2018-01-25
Make the test framework more resilient to errors.

brendandahl:
I was running all tests as 'text' tests and I noticed that some failed to be converted to valid utf8 for xml which caused the test framework to hang. For example: issue2537 failed to be rasterized.

I also tried to fix up a number of other places where we aren't passing on promise rejections or using onerror handlers. There are still more, but I think if we want to fix them all we should just go to async/await to clean this all up.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -43,7 +43,7 @@ var rasterizeTextLayer = (function rasterizeTextLayerClosure() {
 
   function rasterizeTextLayer(ctx, viewport, textContent,
                               enhanceTextSelection) {
-    return new Promise(function (resolve) {
+    return new Promise(function (resolve, reject) {
       // Building SVG with size of the viewport.
       var svg = document.createElementNS(SVG_NS, 'svg:svg');
       svg.setAttribute('width', viewport.width + 'px');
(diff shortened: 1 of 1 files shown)


test/driver.js
```diff
@@ -126,12 +129,15 @@ var rasterizeAnnotationLayer = (function rasterizeAnnotationLayerClosure() {
...
     for (let key in styles) {
-      styles[key].promise = new Promise(function(resolve) {
+      styles[key].promise = new Promise(function(resolve, reject) {
         let xhr = new XMLHttpRequest();
         xhr.open('GET', styles[key].file);
         xhr.onload = function() {
           resolve(xhr.responseText);
         };
+        xhr.onerror = function(e) {
+          reject(new Error('Error fetching annoation style ' + e));
```
Snuffleupagus: Spelling nit: `annoation` -> `annotation`.

test/driver.js
```diff
@@ -220,6 +234,9 @@ var rasterizeAnnotationLayer = (function rasterizeAnnotationLayerClosure() {
             ctx.drawImage(img, 0, 0);
             resolve();
           };
+          img.onerror = function(e) {
+            reject(new Error('Error rasterizing annotations layer ' + e));
```
Snuffleupagus: Nit: `annotations layer` -> `annotation layer`.

Snuffleupagus (approved):

timvandermeij:
@brendandahl Do you have time to fix this up and merge it? It would be good to have this to prevent any errors.

PR #9409, opened 2018-01-26
Fix Chrome 49 compatibility via polyfills

Rob--W:
I suggest that you look at the individual patches for review.


Snuffleupagus (approved):
Looks good, thanks!

Issue #9397, opened 2018-01-23
Evaluation of support for old Chrome versions


Rob--W:
The Chrome extension has been designed to support all Chrome versions since the initial release.

We need to evaluate whether there are commonly used versions that we wish to support, and maybe set a [`minmum_chrome_version` in `manifest.json`](https://developer.chrome.com/extensions/manifest/minimum_chrome_version) to explicitly drop support for older versions.

We can make a data=driven decision since I've been collecting information about browser usage statistics via pdfjs.robwu.nl (https://github.com/Rob--W/pdfjs-telemetry).
In November 2017, I compiled a list of Chrome versions: https://github.com/mozilla/pdf.js/pull/8768#discussion_r135390035

I looked at the data again for this month (2018, 1 jan - 22 jan):

- There are 1.6M unique data points. There has not been a major update of Chrome this month, so the number of double-counted users across two major versions is probably low. The unique user ID is reset when the browser has a major version update, so some users may be counted more than once. The number of unique IDs in the dataset is 1.2M, so about 400k users are counted twice in the statistics below (as a result, the numbers may be slightly skewed towards recent versions, since only recent Chrome versions are automatically updated with patches while keeping the same major version).

```
Chrome 66 0.02%
Chrome 65 0.66%
Chrome 64 0.74% <-- Current beta, will be promoted to stable today (23 jan 2018)
Chrome 63 74.77% <-- Current stable
Chrome 62 13.81%
Chrome 61 3.57%
Chrome 60 1.40%
Chrome 59 0.73%
Chrome 58 0.33%
Chrome 57 0.28%
Chrome 56 0.25%
Chrome 55 0.59%
Chrome 54 0.10%
Chrome 53 0.16%
Chrome 52 0.31%
Chrome 51 0.35%
Chrome 50 0.07%
Chrome 49 0.78% <-- Last supported version of Windows XP
Chrome 48 0.06%
Chrome 47 0.14%
Chrome 46 0.04%
Chrome 45 0.12% <-- Mostly Ubuntu users
Chrome 44 0.02%
Chrome 43 0.12% <-- Mostly Ubuntu users
Chrome 42 0.03%
Chrome 41 0.05%
Chrome 40 0.03%
Chrome 39 0.12%
Chrome 38 0.01%
Chrome 37 0.02%
Chrome 36 0.01%
Chrome 35 0.06%
[...]

PR #9410, opened 2018-01-26
Cleanup code to drop support for Chrome < 49 in the Chrome extension

Rob--W:
This finalizes the work for #9397 .

@Snuffleupagus In particular, this commit removes a snippet that referenced `PDFJS.disableWorker`, so you don't need to account for it any more in #9385.


Snuffleupagus (commented):
I believe that the following line needs to removed to fix the linting failure: https://github.com/mozilla/pdf.js/blob/28c87ce9c61fecbdbfe56edb982a6a5cd9a0ed64/extensions/chromium/pdfHandler.js#L16

Rob--W:
Done.

Snuffleupagus (approved):
I've successfully tested the Chrome extension locally, using Google Chrome `Version 64.0.3282.119 (Official Build) beta (64-bit)`.

According to https://github.com/mozilla/pdf.js/issues/9397#issue-290812728:
> We need to evaluate whether there are commonly used versions that we wish to support, and maybe set a minmum_chrome_version in manifest.json to explicitly drop support for older versions.

Should that be done in the current PR as well?
If the answer is "no", you can land this PR as-is.
Otherwise, please feel free to land once the mentioned `manifest.json` entry has been added. 

Thanks for removing all this old code!

Rob--W:
Among the logged user agents are also custom browsers. I'm not sure about the impact of setting a minimum version in the manifest on these forks, so I'll just omit the field.

Rob--W:
Thanks for the quick review @Snuffleupagus

PR #9412, opened 2018-01-27
Implement the `AESBaseCipher` class and let the `AES128Cipher` and `AES256Cipher` classes extend it

### LINKED ISSUE #5155: Refactor AESXXXCipher in src/core/crypto.js
While developing #5154 , I noticed that there is a huge amount of code duplication in the AES128 and AES256 cipher implementation in src/core/crypto.js. Other than a few constants and two slightly different methods, the rest of the code is equivalent: http://pastebin.mozilla.org/5870349. This should be refactored.

timvandermeij:
More information is provided in the commit messages. Unit tests are already available and still pass. I highly recommend to review this per commit instead of looking at the full diff and reviewing the second commit with the `?w=1` flag.

The diff of the two classes after the first commit can be found at https://www.diffchecker.com/fBgVIwRm. This clearly shows the amount of code duplication and that basically everything can live in the base class except for `_expandKey`. The `decryptBlock` differences are resolved by taking the one for AES-256 in the base class and setting the `iv` parameter to `null` by default so that for AES-128 nothing changes.

Fixes #5155.


src/core/crypto.js
```diff
@@ -995,189 +954,171 @@ var AES128Cipher = (function AES128CipherClosure() {
...
+      this.iv = iv;
+    } else {
       for (i = 0; bufferLength < 16 && i < sourceLength; ++i, ++bufferLength) {
         buffer[bufferLength] = data[i];
       }
-      if (bufferLength < 16) {
-        // need more data
+      if (bufferLength < 16) { // Need more data.
         this.bufferLength = bufferLength;
         return new Uint8Array([]);
```
Snuffleupagus: Pre-existing nit: This looks weird, let's use `new Uint8Array(0);` here and elsewhere instead.

src/core/crypto.js
```diff
@@ -629,173 +629,131 @@ var NullCipher = (function NullCipherClosure() {
...
+  }
+
+  _expandKey(cipherKey) {
+    throw new Error('Cannot call `_expandKey` on the base class');
   }
 
-  function _decrypt(input, key) {
-    var state = new Uint8Array(16);
+  _decrypt(input, key) {
+    let t, u, v, i, j, k;
```
Snuffleupagus: Nit: Since we're now using `let` rather than `var`, thanks for that by the way :-), it doesn't seem entirely necessary to define the *loop* variables in this "method/function global" way any more!?

This probably applies elsewhere too...

Snuffleupagus (commented):
Thanks for taking this on, since there's a lot of duplication removed here!

Given that we've thus far avoided using ES6 `class`es in `src/core/` code, since the performance might still not be on-par with the old way of simulating classes, would it be possible to find a PDF file (e.g. in the test-suite) which make use of either of these algorithms and [benchmark](https://github.com/mozilla/pdf.js/wiki/Benchmarking-your-changes) the impact of this PR?

*Edit:* One thing that's probably missing from those instructions, and which I forgot myself once, is that you need to remember to manually re-run `gulp generic` *before* each benchmarking step (since we're now relying on a built `pdf.worker.js` file when running tests).

I realize that it may perhaps be difficult to find such a file, but I figured that it cannot hurt to ask.

timvandermeij:
Thank you for the review! I have made all changes and I did a benchmark of the code.

I created the following `benchmark.json` file with two files from the test suite that use the AES-128 cipher (found using `grep -r "AESV2" test/*` and confirmed by placing print statements in the class and opening the files in the viewer):

```
[
  {
    "id": "i9",
    "file": "pdfs/i9.pdf",
    "md5": "ba7cd54fdff083bb389295bc0415f6c5",
    "rounds": 20,
    "type": "load"
  },
  {
    "id": "ichiji.pdf",
    "file": "pdfs/ichiji.pdf",
    "md5": "66b645802d33513cd598886e017392b8",
    "rounds": 20,
    "type": "load"
  }
]
```

Then I ran all benchmarking steps from https://github.com/mozilla/pdf.js/wiki/Benchmarking-your-changes, which I updated to be more step-by-step and to include the `gulp generic` step, which was a really good point from you. Doing so I obtained the following table:

```
Firefox | Overall      |   120 |           55 |          54 |  -1 | -1.69 |              
Firefox | Page Request |   120 |            3 |           3 |   0 | -4.26 |              
Firefox | Rendering    |   120 |           52 |          52 |  -1 | -1.59 |
```

Therefore, we can conclude that this change has no negative performance impact.

Snuffleupagus (approved):
Looks good, thanks for cleaning this up!

Also, thanks's for taking the time to benchmark the changes.

PR #9416, opened 2018-01-28
fix example helloworld

valentin-p:
The example helloworld was using an obsolete way of importing the pdf setPDFNetworkStreamClass, i changed it to setPDFNetworkStreamFactory.


timvandermeij:
Closing in favor of #9436 since that one also makes the SVG viewer example work again and keeps using the network stream like we currently do, because the fetch stream does not work at least locally using Firefox. Thank you for bringing this to our attention though!

PR #9417, opened 2018-01-29
[api-minor] Refactor annotation rendering

dmitryskey:
As it was discussed in the https://github.com/mozilla/pdf.js/pull/8625 this is a first commit from the series of changes were I'm adding rendering of font properties and color features for input wizard annotations. This commit doesn't contain any functional changes rather code refactoring. As it was already discussed in more details we need an additional worker to parse annotation AcroForm dictionary. I created a dummy one called AnnotationWorkerTask in order to reduce number of potential regressions, but it might be possible to re-utilize the standard worker although it doesn't seem like we have any performance penalties with additional one.

This commit also contains updated unit-tests.


src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+      return Promise.resolve(this);
+    }
+
+    let data = this.data;
+    let self = this;
+
+    let opList = new OperatorList();
+    let appearanceStream = new Stream(stringToBytes(data.defaultAppearance));
+    let formFonts = params.pdfManager.pdfDocument.acroForm.get('DR');
+    opList.acroForm = params.pdfManager.pdfDocument.acroForm;
```
timvandermeij: We should not put additional member variables on the operator list. I don't really see why it should be necessary and it breaks the expected structure of the object.

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+    return params.evaluator.getOperatorList({
+      stream: appearanceStream,
+      task: params.task,
+      resources: formFonts,
+      operatorList: opList,
+    }).then(() => {
+      let a = opList.argsArray;
+      let i;
```
timvandermeij: Remove this line and put the `let` in the `for` loop instead: `for (let i = 0, ii = opList.fnArray.length; i < ii; i++) {` is usually how we write those loops (since it caches the length too).

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+    let appearanceStream = new Stream(stringToBytes(data.defaultAppearance));
+    let formFonts = params.pdfManager.pdfDocument.acroForm.get('DR');
+    opList.acroForm = params.pdfManager.pdfDocument.acroForm;
+    return params.evaluator.getOperatorList({
+      stream: appearanceStream,
+      task: params.task,
+      resources: formFonts,
+      operatorList: opList,
+    }).then(() => {
+      let a = opList.argsArray;
```
timvandermeij: Can we give this a more descriptive name? Perhaps `args`?

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+      let i;
+      for (i = 0; i < opList.fnArray.length; i++) {
+        let fn = opList.fnArray[i];
+        switch (fn | 0) {
+          case OPS.setFont:
+            data.annotationFonts = opList.acroForm.annotationFonts;
```
timvandermeij: I think we should look for another way to pass this information because changing existing object seems like an antipattern. Since we have the `task` available, I think a better (but not optimal) solution would be to store this information in the `task` instead, since that is also the one fetching the information. In any case, I think we should not pollute the `opList` and `pdfManager` objects.

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+            data.fontRefName = a[i][0];
+            data.fontSize = a[i][1];
+            break;
+          case OPS.setGrayFill:
+            if (a[i].length >= 1) {
```
timvandermeij: This should never happen, right (same for the check in the operator below)? If we do want to check this, please log a warning if this `if` evaluates to `false` so we can see it in the console.

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+              let gray = Math.round(a[i][0] * 0x100);
+              data.fontColor = Util.makeCssRgb(gray, gray, gray);
```
timvandermeij: Do this operator and the one below really operate on the font? Their names indicate to me that they work on the rectangle instead, providing gray/RGB fill colors, but I may be wrong here!

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+            }
+
```
timvandermeij: Remove the newlines before the `break` (below too).

src/core/annotation.js
```diff
@@ -772,10 +845,12 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     if (!isDict(appearanceStates)) {
       return;
     }
+
```
timvandermeij: Revert the newline changes here and a few lines below. They are unrelated to your patch.

src/core/document.js
```diff
@@ -14,11 +14,10 @@
  */
 
 import { Catalog, ObjectLoader, XRef } from './obj';
+import { createPromiseCapability, error, info, isArrayBuffer, isNum,
```
timvandermeij: Format this in the same way as before, so `import {` on one line and the actual imports start on the next line. Finally, `} from ...` should also be on a new line.

src/core/document.js
```diff
@@ -33,8 +32,8 @@ var Page = (function PageClosure() {
   var LETTER_SIZE_MEDIABOX = [0, 0, 612, 792];
 
   function isAnnotationRenderable(annotation, intent) {
-    return (intent === 'display' && annotation.viewable) ||
-           (intent === 'print' && annotation.printable);
+    return (intent === 'display' && annotation && annotation.viewable) ||
```
timvandermeij: This should not be possible, because `annotation` should always be defined. The annotations in the list returned by the annotations API must always contain valid `Annotation` objects, so what is the reason for this change?

src/core/document.js
```diff
@@ -255,7 +254,7 @@ var Page = (function PageClosure() {
         for (i = 0, ii = annotations.length; i < ii; i++) {
           if (isAnnotationRenderable(annotations[i], intent)) {
             opListPromises.push(annotations[i].getOperatorList(
-              partialEvaluator, task, renderInteractiveForms));
+              partialEvaluator, task, true));
```
timvandermeij: Revert this change. I presume it was for testing your change, but we do want to use the preference here. Tip: when you're developing, open the console and run `PDFViewerApplication.preferences.set('renderInteractiveForms', true);` (and refresh the page) to enable this preference locally so you can easily test it.

src/core/document.js
```diff
@@ -308,29 +307,112 @@ var Page = (function PageClosure() {
...
         }
-      }
-      return annotationsData;
+        return annotationsData;
+      });
     },
 
     get annotations() {
-      var annotations = [];
+      var AnnotationWorkerTask = (function AnnotationWorkerTaskClosure() {
```
timvandermeij: I think it would be better to import the existing `WorkerTask` class from `src/core/worker.js` for this instead since the code itself looks exactly the same.

src/core/document.js
```diff
@@ -308,29 +307,112 @@ var Page = (function PageClosure() {
...
+            if (this.terminated) {
+              throw new Error('Annotation worker task was terminated');
+            }
+          },
+        };
+
+        return AnnotationWorkerTask;
+      })();
+
+      // create a blank annotation fonts array
```
timvandermeij: I think we can avoid this step completely if we put this in the `task` instead.

src/core/document.js
```diff
@@ -308,29 +307,112 @@ var Page = (function PageClosure() {
...
+        handler,
+        pageIndex: this.pageIndex,
+        idFactory: this.idFactory,
+        fontCache: this.fontCache,
+        builtInCMapCache: this.builtInCMapCache,
+        options: this.evaluatorOptions,
+      });
+
       var annotationRefs = this.getInheritedPageProp('Annots') || [];
+      var annotationsPromise = [];
```
timvandermeij: `annotationPromises`

src/core/document.js
```diff
@@ -308,29 +307,112 @@ var Page = (function PageClosure() {
...
       for (var i = 0, n = annotationRefs.length; i < n; ++i) {
         var annotationRef = annotationRefs[i];
-        var annotation = AnnotationFactory.create(this.xref, annotationRef,
-                                                  this.pdfManager,
-                                                  this.idFactory);
-        if (annotation) {
-          annotations.push(annotation);
-        }
+          annotationsPromise.push(AnnotationFactory.create(
```
timvandermeij: Create an intermediate variable for readability, so first `let annotationPromise = AnnotationFactory.create(...);` and then `annotationPromises.push(annotationPromise);`.

src/core/document.js
```diff
@@ -308,29 +307,112 @@ var Page = (function PageClosure() {
...
+          this.xref,
+          annotationRef,
+          this.pdfManager,
+          this.idFactory,
+          partialEvaluator,
+          task
```
timvandermeij: End with a trailing comma.

src/core/document.js
```diff
@@ -308,29 +307,112 @@ var Page = (function PageClosure() {
...
+        ));
       }
-      return shadow(this, 'annotations', annotations);
+
+      return shadow(this, 'annotations',
+        Promise.all(annotationsPromise)).then(function (annotations) {
```
timvandermeij: I don't think we need to do all the logging for this worker task. Let's keep things simple here and remove it.

timvandermeij (commented):
Thank you for splitting the work into multiple smaller pull requests, and nice work so far! This makes it easier to review and understand.

Since this is the base work for other pull requests, I want to make this as good as possible. I added quite a bit of comments, but a lot of them are also stylistic to make the code more consistent with the rest of the codebase. Overall the patch is looking good, so please don't consider the number of comments a negative thing! I think the comments will help to simplify the implementation quite a bit and I try to give some tips to make developing easier. If you have any questions, please feel free to comment on them.

I would also suggest to amend the commit message to e.g., `Implement initial default appearance parsing for widget annotations`. This makes it easier to see in the history what changed exactly.

src/core/document.js
Snuffleupagus: That still wouldn't address the underlying problem with `WorkerTask`s in this PR, since the purpose of `WorkerTask` is to allow waiting for the completion of various running tasks in `src/core/worker.js` to avoid e.g. intermittent issues in tests. Note how a worker won't be terminated until all running `WorkerTask`s have finished:
https://github.com/mozilla/pdf.js/blob/55e3f97aa97ad648ef955203524bbdb5593671d1/src/core/worker.js#L832-L846

As is, this PR (and the previous one) is completely breaking that assumption!

---

It seems to me that this PR still suffers from the same issue as the original one: It's simply *way* too large, with lots of different changes, making it very difficult to follow.
I'd thus suggest splitting this up a lot more, for example let's start with a PR that *only* makes the annotation parsing async, then one to parse the `AcroForm` dictionary[1], then one to use that dictionary when parsing annotations, and so on...

[1] Also, correct parsing of `AcroForm` dictionaries should also fix the (default) `canvas` based rendering of form elements.

src/core/annotation.js
```diff
@@ -50,36 +56,46 @@ class AnnotationFactory {
...
+          default:
+            warn('Unimplemented widget field type "' + fieldType + '", ' +
+               'falling back to base field type.');
+            break;
         }
-        warn('Unimplemented widget field type "' + fieldType + '", ' +
-             'falling back to base field type.');
-        return new WidgetAnnotation(parameters);
+
+        return widget.parseAppearance(parameters);
```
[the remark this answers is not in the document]
Snuffleupagus: In the context of the *current* PR, I'm not even sure if `parseAppearance` does anything nor if it even works!?
Considering that it attempts to access non-existent properties of object, e.g. `acroForm.annotationFonts`, which doesn't actually exist yet. As mentioned elsewhere, I think that this PR needs to be split up a lot more.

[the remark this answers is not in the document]
Snuffleupagus: In the context of the *current* PR, I'm not even sure if `parseAppearance` does anything nor if it even works!?
Considering that it attempts to access non-existent properties of object, e.g. `acroForm.annotationFonts`, which doesn't actually exist yet. As mentioned elsewhere, I think that this PR needs to be split up a lot more.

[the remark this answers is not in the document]
Snuffleupagus: In the context of the *current* PR, I'm not even sure if `parseAppearance` does anything nor if it even works!?
Considering that it attempts to access non-existent properties of objects, e.g. `acroForm.annotationFonts`, which doesn't actually exist yet. As mentioned elsewhere, I think that this PR needs to be split up a lot more.

src/core/annotation.js
```diff
@@ -31,11 +31,17 @@ class AnnotationFactory {
    * @param {Object} idFactory
    * @returns {Annotation}
    */
-  static create(xref, ref, pdfManager, idFactory) {
+  static create(xref, ref, pdfManager, idFactory, evaluator, task) {
+    if (!xref) {
```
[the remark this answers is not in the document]
dmitryskey: First of all, thank you very much for your review. I'm kinda busy with my current work, so I'll respond pretty slow + I need time to re-run tests after changes etc.

Off the top of my head I added this validation when I was working on unit tests. I removed it and it seems like no regression.

src/core/annotation.js
```diff
@@ -31,11 +31,17 @@ class AnnotationFactory {
...
+      return;
+    }
+
     let dict = xref.fetchIfRef(ref);
+
```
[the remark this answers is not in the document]
dmitryskey: Ok

src/core/annotation.js
```diff
@@ -50,36 +56,46 @@ class AnnotationFactory {
...
 
       case 'Text':
-        return new TextAnnotation(parameters);
+        return Promise.resolve(new TextAnnotation(parameters));
 
       case 'Widget':
         let fieldType = Util.getInheritableProperty(dict, 'FT');
         fieldType = isName(fieldType) ? fieldType.name : null;
 
+        let widget = new WidgetAnnotation(parameters);
```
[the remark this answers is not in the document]
dmitryskey: Here is a deal. Now instead of annotation itself I return promise. The reason I do it because DR dictionary should be parsed asynchronously, see also my comments below.

Also I just realized that I merged the code incorrectly for new annotations like Stamp, I'll add Promise.resolve to them as well.

src/core/annotation.js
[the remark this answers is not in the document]
dmitryskey: So, let try to explain my logic and I'd really appreciate your ideas and suggestions. AcroForm DR dictionary currently is not parsed with all sub-elements that contain an info regarding font style, background color, border color etc. I see two ways to do that - extend the current worker to do that or add another cycle of parsing. Personally I'd prefer the former, but I'm just not sure how to implement it without regressions. The latter seems working fine but it's definitely seems like 'plug-in' code. parseAnnotation does exactly what I mentioned - it returns another promise that is resolved in the second worker and parse 'DR' dictionary.

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
     return !!(this.data.fieldFlags & flag);
   }
 
+  /**
+   * Parse widget annotation appearance and extract font information.
+   *
+   * @public
+   * @memberof WidgetAnnotation
+   * @param {number} flag - Hexadecimal representation for an annotation
```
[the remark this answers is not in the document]
dmitryskey: Ok, cool, sounds great. I just not 100% sure for private vs public because this method returns promise and I call it after widget annotation creation.

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+   *                        field characteristic
+   * @return {Promise}
+   */
+  parseAppearance(params) {
```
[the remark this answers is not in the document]
dmitryskey: I renamed it setDefaultAppearance, see my comment above

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+    if (!params.pdfManager || !params.pdfManager.pdfDocument) {
+      return Promise.resolve(this);
+    }
+
+    let data = this.data;
+    let self = this;
```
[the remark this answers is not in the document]
dmitryskey: Thank you to point me out. I totally forgot about that after converting annotation.js to ES6 and kept ES5 code.

src/core/annotation.js
dmitryskey: Here I think I need your recommendation, I just not familiar enough with all details of the PDF.JS code. Is it fine to extend OperatorList object with new property acroForm? In this case I tried to keep changes local in order before calling evaluator...

src/core/annotation.js
dmitryskey: Sounds great, args works for me

src/core/annotation.js
dmitryskey: Ok, cool, thank you

src/core/annotation.js
dmitryskey: sounds great, I'll add warning instead

src/core/annotation.js
dmitryskey: In this case PDF specs are vague, I didn't find if font color is always specified as RGB or it's possible that gray might be used. I'd rather to keep it just in case.

src/core/annotation.js
dmitryskey: Thank you, done

src/core/annotation.js
dmitryskey: Done, thank you

src/core/document.js
dmitryskey: Got it, thank you. I wasn't sure for such formatting rules and relied upon lint validation

src/core/document.js
dmitryskey: Here is a deal. There is a test file called annotation-text-without-popup.pdf that failed without this validation. Let me take a look once again and try to find why empty annotation was returned for this form

src/core/document.js
dmitryskey: This is 100% true, thank you!

src/core/document.js
dmitryskey: Done, thank you

src/core/document.js
dmitryskey: Done

src/core/document.js
dmitryskey: I'm not sure for that, this is function call, do we need this comma after task (if you meant that)

src/core/document.js
dmitryskey: Ok, done, thank you

src/core/document.js
dmitryskey: Thank you very much for the worker import hint, I replaced it and it worked just fine.

I'm 100% ok to split it further, but could you guys explain a little bit problem you referred to? Should I add the worker task I use for parsing annotations to the WorkerTasks collection?

src/core/annotation.js
```diff
@@ -661,6 +677,63 @@ class WidgetAnnotation extends Annotation {
...
+   * Parse widget annotation appearance and extract font information.
+   *
+   * @public
+   * @memberof WidgetAnnotation
+   * @param {number} flag - Hexadecimal representation for an annotation
+   *                        field characteristic
+   * @return {Promise}
+   */
+  parseAppearance(params) {
+    if (!params.pdfManager || !params.pdfManager.pdfDocument) {
```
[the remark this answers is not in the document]
dmitryskey: This is validation required because of unit-tests. We don't create AcroForm dictionary for them and I tried to minify unit-tests modifications. Normally this validation does nothing, and in this sense it's 100% correct

src/core/document.js
dmitryskey: So I took a look at this PDF in RUSP and as you can see it contains null annotation. That's why I added such a condition. Formally this is malformed document but I'd prefer to improve the annotation parsing a little bit

image

src/core/annotation.js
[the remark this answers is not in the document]
timvandermeij: The issue here is not that that it returns a promise, but rather that a new instance is created here *unconditionally*, and most likely overwritten a few lines below. That means that this object is created unnecessarily in most cases, so if you move the creation of the object into the `default` case, you'll end up with the same behavior as before, but no intermediate object that is overwritten later.

src/core/annotation.js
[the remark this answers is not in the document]
timvandermeij: If you implement the comment above, this can become private because you would not need to call it manually anymore. Instead, it would be called in the constructor of the object itself, therefore the other code does not need to know about it.

src/core/document.js
timvandermeij: > I'd thus suggest splitting this up a lot more, for example let's start with a PR that only makes the annotation parsing async, then one to parse the AcroForm dictionary[1], then one to use that dictionary when parsing annotations, and so on...

I think that sounds good. The smaller the patch, the easier it is to verify. I'm not too familiar with how `WorkerTask`s work, unfortunately, so I too could use some additional feedback here. The only reason we're even dealing with a `WorkerTask` is because it's a necessary parameter. What should we do differently with it?

src/core/document.js
timvandermeij: Yes, a comma after `task`. The reason to end with a trailing comma is that if in the future we add more parameters, you don't need to touch this line again. It makes it a bit easier in the Git logs.

src/core/annotation.js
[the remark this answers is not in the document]
dmitryskey: Ok, now I see, thank you!

src/core/annotation.js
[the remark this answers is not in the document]
dmitryskey: But how can I return promise right from the constructor? Can you show me some code snippet, maybe I just not very well familiar with promises syntax

src/core/document.js
dmitryskey: I'm still a little bit confused. Here is the function call, if I add the comma after task Lint generates an error.

static create(xref, ref, pdfManager, idFactory, evaluator, task);

var annotationPromise = AnnotationFactory.create(
          this.xref,
          annotationRef,
          this.pdfManager,
          this.idFactory,
          partialEvaluator,
          task
        );

Here is the current code from master, there is no command after the last parameter as well:

var annotation = AnnotationFactory.create(this.xref, annotationRef,
                                                  this.pdfManager,
                                                  this.idFactory);

timvandermeij: I thought this was an object and did not see it was a function call. You can disregard my comment; sorry for the noise!

src/core/annotation.js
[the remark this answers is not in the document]
timvandermeij: Now I see the problem here. Indeed, returning a promise from the constructor is not a good pattern. Your way of doing this is better then, but I'm still a bit on the fence about letting other code trigger the fetching of the appearance data; it seems like something the class should handle itself, but I'm not sure if that's possible. For now you can leave the code like you have it already; perhaps if the updated patch is pushed (with should already be smaller given that most of the comments are adressed) we'll get new ideas or we'll just leave it like this.

dmitryskey:
Hi!

I simplified this commit as much as possible and rebased this commit. Now it contains asynchronous annotation parsing as it was proposed above.

QQ - can be keep reviewing in this thread or it'd better to close this request and open a new one?

dmitryskey:
Hi, All!

Is there any news on code review?

test/unit/annotation_spec.js
```diff
@@ -70,12 +70,13 @@ describe('annotation', function() {
         { ref: annotationRef, data: annotationDict, }
       ]);
 
-      var annotation = AnnotationFactory.create(xref, annotationRef,
-                                                pdfManagerMock, idFactoryMock);
-      var data = annotation.data;
-      expect(data.annotationType).toEqual(AnnotationType.LINK);
+      AnnotationFactory.create(xref, annotationRef,
+        pdfManagerMock, idFactoryMock).then((annotation) => {
```
timvandermeij: Here and below: indent this line with two more spaces so it becomes:

```javascript
AnnotationFactory.create(xref, annotationRef,
    pdfManagerMock, idFactoryMock).then((annotation) => {
  var data = annotation.data;
  expect(data.annotationType).toEqual(AnnotationType.LINK);
  expect(data.id).toEqual('10R');
});

This is just our code style for multi-line statements so that we can keep the body indented with two spaces.
```

test/unit/annotation_spec.js
```diff
@@ -90,16 +91,19 @@ describe('annotation', function() {
...
-      var annotation1 = AnnotationFactory.create(xref, annotationDict,
-                                                 pdfManagerMock, idFactory);
-      var annotation2 = AnnotationFactory.create(xref, annotationDict,
-                                                 pdfManagerMock, idFactory);
-      var data1 = annotation1.data, data2 = annotation2.data;
-      expect(data1.annotationType).toEqual(AnnotationType.LINK);
-      expect(data2.annotationType).toEqual(AnnotationType.LINK);
+      AnnotationFactory.create(xref, annotationDict,
+                pdfManagerMock, idFactory).then((annotation1) => {
+        var data1 = annotation1.data;
```
timvandermeij: Since it's scoped, there is no need to call this `data1` and `annotation1`. I think you can just use `data` and `annotation` here if you use `let` instead of `var` (the latter applies in your other changes in this file too).

test/unit/annotation_spec.js
```diff
@@ -937,29 +943,40 @@ describe('annotation', function() {
       var flags = AnnotationFieldFlag.COMB + AnnotationFieldFlag.MULTILINE +
                   AnnotationFieldFlag.PASSWORD + AnnotationFieldFlag.FILESELECT;
 
-      for (var i = 0, ii = invalidFieldFlags.length; i <= ii; i++) {
+      var xref;
```
timvandermeij: I don't really understand why this test has to be changed other than making it asynchronous. Why the additional `AnnotationFactory.create`?

timvandermeij (commented):
First of all, sorry for the delay here! I've been a bit limited in my time and we were also busy getting PDF.js 2.0 closer to a release.

This looks much better already! However, as @Snuffleupagus commented, it would be better to split the changes a bit more:

> I'd thus suggest splitting this up a lot more, for example let's start with a PR that only makes the annotation parsing async, then one to parse the AcroForm dictionary[1], then one to use that dictionary when parsing annotations, and so on...

Let's do the first in this PR, i.e., only make the parsing asynchronous. It will make the patch much simpler to understand and easier to merge. I'll indicate which parts you can remove from this PR. If that is done, the patch should be much smaller and mergeable. Thank you!

src/core/annotation.js
```diff
@@ -31,7 +31,7 @@ class AnnotationFactory {
    * @param {Object} idFactory
    * @returns {Annotation}
    */
-  static create(xref, ref, pdfManager, idFactory) {
+  static create(xref, ref, pdfManager, idFactory, evaluator, task) {
```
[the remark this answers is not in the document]
dmitryskey: Ok

src/core/annotation.js
```diff
@@ -50,66 +50,76 @@ class AnnotationFactory {
       subtype,
       id,
       pdfManager,
+      evaluator,
```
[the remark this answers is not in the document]
dmitryskey: Ok.

src/core/annotation.js
```diff
@@ -50,66 +50,76 @@ class AnnotationFactory {
...
 
       case 'Text':
-        return new TextAnnotation(parameters);
+        return Promise.resolve(new TextAnnotation(parameters));
 
       case 'Widget':
         let fieldType = Util.getInheritableProperty(dict, 'FT');
         fieldType = isName(fieldType) ? fieldType.name : null;
 
+        let widget = new WidgetAnnotation(parameters);
```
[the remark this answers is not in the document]
dmitryskey: Ok

src/core/annotation.js
```diff
@@ -50,66 +50,76 @@ class AnnotationFactory {
...
+          default:
+            warn('Unimplemented widget field type "' + fieldType + '", ' +
+               'falling back to base field type.');
+            break;
         }
-        warn('Unimplemented widget field type "' + fieldType + '", ' +
-             'falling back to base field type.');
-        return new WidgetAnnotation(parameters);
+
+        return widget.setDefaultAppearance(parameters);
```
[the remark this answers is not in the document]
dmitryskey: Ok

src/core/annotation.js
```diff
@@ -661,6 +671,18 @@ class WidgetAnnotation extends Annotation {
     return !!(this.data.fieldFlags & flag);
   }
 
+  /**
```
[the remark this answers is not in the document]
dmitryskey: Ok

src/core/document.js
```diff
@@ -33,8 +33,8 @@ var Page = (function PageClosure() {
   var LETTER_SIZE_MEDIABOX = [0, 0, 612, 792];
 
   function isAnnotationRenderable(annotation, intent) {
-    return (intent === 'display' && annotation.viewable) ||
-           (intent === 'print' && annotation.printable);
+    return (intent === 'display' && annotation && annotation.viewable) ||
```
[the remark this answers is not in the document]
dmitryskey: Ok, sounds reasonable. The issue I addressed is definitely existing, I just wasn't sure what is your internal policy for PR

src/core/worker.js
```diff
@@ -735,22 +743,29 @@ var WorkerMessageHandler = {
         var task = new WorkerTask('RenderPageRequest: page ' + pageIndex);
         startWorkerTask(task);
 
+        var annotationTask = new WorkerTask(
```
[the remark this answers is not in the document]
dmitryskey: Ok. Also I removed task parameter from get_AnnotationsData method

test/unit/annotation_spec.js
dmitryskey: Yes, I see.

test/unit/annotation_spec.js
dmitryskey: I'm not 100% positive I understand this unit test, but here is a deal. With synchronous execution it runs for combobox flag + invalid flags and check the condition expect(data.comb).toEqual(false) and then for the combobox only with condition  expect(data.comb).toEqual(true). In this mode one can have the same loop and calculate true/false parameter. With asynchronous model I was forced to split loop into two parts - combobox flag + invalid flags and then combobox only.

I feel the code might be simplified, so any suggestions are very appreciated.

timvandermeij: The unit test checks if the "comb" mode for text widget annotations is only set when the field flags are valid. The "comb" mode splits the text field into equal regions of a fixed amount of characters. For example, a field to enter a serial number that consists of three blocks of four characters will be rendered as as 1234 | 5678 | 9012, equally spaced in the text field. It checks this by setting all invalid field flags and removing them one by one in a loop until all invalid field flags are gone. The "comb" mode should only be set if the invalid field flags are all removed, which is what we assert.

However, since you're only making the annotation API asynchonous, no other changes should be necessary for this unit test as far as I can tell. In other words, I think the following should work for your purposes:

```javascript
    it('should only accept comb fields when the flags are valid', function() {
      let invalidFieldFlags = [
        AnnotationFieldFlag.MULTILINE,
        AnnotationFieldFlag.PASSWORD,
        AnnotationFieldFlag.FILESELECT,
      ];

      // Start with all invalid flags set and remove them one by one.
      // The field may only use combs when all invalid flags are unset.
      let flags = AnnotationFieldFlag.COMB + AnnotationFieldFlag.MULTILINE +
                 AnnotationFieldFlag.PASSWORD + AnnotationFieldFlag.FILESELECT;

      for (let i = 0, ii = invalidFieldFlags.length; i <= ii; i++) {
        textWidgetDict.set('MaxLen', 20);
        textWidgetDict.set('Ff', flags);

        let textWidgetRef = new Ref(93, 0);
        let xref = new XRefMock([
          { ref: textWidgetRef, data: textWidgetDict, }
        ]);

        AnnotationFactory.create(xref, textWidgetRef, pdfManagerMock,
            idFactoryMock).then((annotation) => {
          let data = annotation.data;
          expect(data.annotationType).toEqual(AnnotationType.WIDGET);

          let valid = (invalidFieldFlags.length === 0);
          expect(data.comb).toEqual(valid);
        });

        // Remove the last invalid flag for the next iteration.
        if (!valid) {
          flags -= invalidFieldFlags.splice(-1, 1);
        }
      }
    });
```

Hopefully this helps to understand the test a bit better. If not, or the test turns out to be more complicated than I thought here, feel free to let me know and I'll try to provide more explanation or look into any difficulties!

dmitryskey: The problem with this code (and as far as I remember it was exactly what I tried to address by splitting loop into 2 parts) is the 'valid' variable. It's initiated asynchronously and as result you can't use it in the condition 'if (!valid)' directly, unit test fails with the message "ReferenceError: valid is not defined". Plus if invalidFieldFlag is spliced synchronously I'm not sure if you can rely on it in the asynchronous l-function, it might cause a race condition.

timvandermeij: Yes, I now understand why you made that choice. For now I don't really see a better way, so let's keep it like that. Thank you for explaining this!

dmitryskey:
Hi, updated commit is pushed into branch

timvandermeij:
Thank you! I'll ty to re-check this during the weekend. I think that will result in the final set of comments and then we should be good to merge this.

src/core/document.js
```diff
@@ -308,29 +309,40 @@ var Page = (function PageClosure() {
...
         }
-      }
-      return annotationsData;
+        return annotationsData;
+      });
     },
 
-    get annotations() {
-      var annotations = [];
+    getAnnotations: function Page_getAnnotations() {
```
timvandermeij: Add the following right below this line:

```javascript
if (this._annotations) {
  return this._annotations;
}
```

This handles the caching for you in a private member variable. If `getAnnotations` is called for the first time, the code below this will be executed, but when it's called a second time, the cached variable will be returned.

src/core/document.js
```diff
@@ -308,29 +309,40 @@ var Page = (function PageClosure() {
...
+          annotationRef,
+          this.pdfManager,
+          this.idFactory
+        );
+
+        annotationsPromises.push(annotationPromise);
       }
-      return shadow(this, 'annotations', annotations);
+
+      return shadow(this, 'annotations',
```
timvandermeij: No need for the shadowing anymore now that this is not a getter anymore. You can change it to `this._annotations = Promise.all(...). then(...);` and then `return this._annotations;`.

src/core/worker.js
```diff
@@ -724,8 +724,12 @@ var WorkerMessageHandler = {
     );
 
     handler.on('GetAnnotations', function wphSetupGetAnnotations(data) {
-      return pdfManager.getPage(data.pageIndex).then(function(page) {
-        return pdfManager.ensure(page, 'getAnnotationsData', [data.intent]);
+      var pageIndex = data.pageIndex;
```
timvandermeij: No need for this intermediate variable; just keep it like before since it should fit on one line.

test/unit/annotation_spec.js
```diff
@@ -90,16 +91,19 @@ describe('annotation', function() {
...
 
-      var annotation1 = AnnotationFactory.create(xref, annotationDict,
-                                                 pdfManagerMock, idFactory);
-      var annotation2 = AnnotationFactory.create(xref, annotationDict,
-                                                 pdfManagerMock, idFactory);
-      var data1 = annotation1.data, data2 = annotation2.data;
-      expect(data1.annotationType).toEqual(AnnotationType.LINK);
-      expect(data2.annotationType).toEqual(AnnotationType.LINK);
+      AnnotationFactory.create(xref, annotationDict,
+                pdfManagerMock, idFactory).then((annotation) => {
```
timvandermeij: Indent this like you did here: https://github.com/mozilla/pdf.js/pull/9417/files#diff-04ad2b4c022ab1eae0d58931b8fe0b7dR74

test/unit/annotation_spec.js
```diff
@@ -90,16 +91,19 @@ describe('annotation', function() {
...
+        var data = annotation.data;
+        expect(data.annotationType).toEqual(AnnotationType.LINK);
+        expect(data.id).toEqual('annot_p0_1');
+      });
 
-      expect(data1.id).toEqual('annot_p0_1');
-      expect(data2.id).toEqual('annot_p0_2');
+      AnnotationFactory.create(xref, annotationDict,
+                pdfManagerMock, idFactory).then((annotation) => {
```
timvandermeij: Indent this like you did here: https://github.com/mozilla/pdf.js/pull/9417/files#diff-04ad2b4c022ab1eae0d58931b8fe0b7dR74

test/unit/annotation_spec.js
```diff
@@ -448,15 +454,15 @@ describe('annotation', function() {
...
-                                                pdfManager, idFactoryMock);
-      var data = annotation.data;
-      expect(data.annotationType).toEqual(AnnotationType.LINK);
-
-      expect(data.url).toEqual(
-        'http://www.example.com/0013/001346/134685E.pdf#4.3');
-      expect(data.unsafeUrl).toEqual('../../0013/001346/134685E.pdf#4.3');
-      expect(data.dest).toBeUndefined();
+      AnnotationFactory.create(xref, annotationRef,
+                pdfManager, idFactoryMock).then((annotation) => {
```
timvandermeij: Indent this like you did here: https://github.com/mozilla/pdf.js/pull/9417/files#diff-04ad2b4c022ab1eae0d58931b8fe0b7dR74

test/unit/annotation_spec.js
```diff
@@ -506,17 +512,17 @@ describe('annotation', function() {
...
-
-      expect(data.url).toEqual(new URL('http://www.example.com/test.pdf#' +
+      AnnotationFactory.create(xref, annotationRef,
+          pdfManagerMock, idFactoryMock).then((annotation) => {
+        var data = annotation.data;
+        expect(data.annotationType).toEqual(AnnotationType.LINK);
+        expect(data.url).toEqual(new URL('http://www.example.com/test.pdf#' +
                                  '[14,{"name":"XYZ"},null,298.043,null]').href);
-      expect(data.unsafeUrl).toEqual('http://www.example.com/test.pdf#' +
+        expect(data.unsafeUrl).toEqual('http://www.example.com/test.pdf#' +
```
timvandermeij: Add two extra spaces of indenting to the line below to align it nicely with this line.

test/unit/annotation_spec.js
```diff
@@ -546,16 +552,16 @@ describe('annotation', function() {
...
-      var data = annotation.data;
-      expect(data.annotationType).toEqual(AnnotationType.LINK);
-
-      expect(data.url).toEqual(
-        new URL('http://www.example.com/test/pdfs/Part II/Part II.pdf').href);
-      expect(data.unsafeUrl).toEqual('Part II/Part II.pdf');
-      expect(data.dest).toBeUndefined();
-      expect(data.newWindow).toEqual(true);
+      AnnotationFactory.create(xref, annotationRef,
+                pdfManager, idFactoryMock).then((annotation) => {
```
timvandermeij: Indent this like you did here: https://github.com/mozilla/pdf.js/pull/9417/files#diff-04ad2b4c022ab1eae0d58931b8fe0b7dR74

test/unit/annotation_spec.js
```diff
@@ -677,15 +682,15 @@ describe('annotation', function() {
...
-      expect(data.url).toBeUndefined();
-      expect(data.unsafeUrl).toBeUndefined();
-      expect(data.dest).toEqual([{ num: 17, gen: 0, }, { name: 'XYZ', },
+      AnnotationFactory.create(xref, annotationRef,
+          pdfManagerMock, idFactoryMock).then((annotation) => {
+        var data = annotation.data;
+        expect(data.annotationType).toEqual(AnnotationType.LINK);
+        expect(data.url).toBeUndefined();
+        expect(data.unsafeUrl).toBeUndefined();
+        expect(data.dest).toEqual([{ num: 17, gen: 0, }, { name: 'XYZ', },
```
timvandermeij: Add two extra spaces of indenting to the line below to align it nicely with this line.

test/unit/annotation_spec.js
```diff
@@ -1295,12 +1313,12 @@ describe('annotation', function() {
...
       ]);
 
-      var annotation = AnnotationFactory.create(xref, lineRef, pdfManagerMock,
-                                                idFactoryMock);
-      var data = annotation.data;
-      expect(data.annotationType).toEqual(AnnotationType.LINE);
-
-      expect(data.lineCoordinates).toEqual([1, 2, 3, 4]);
+      AnnotationFactory.create(xref, lineRef, pdfManagerMock,
+                idFactoryMock).then((annotation) => {
```
timvandermeij: Indent this like you did here: https://github.com/mozilla/pdf.js/pull/9417/files#diff-04ad2b4c022ab1eae0d58931b8fe0b7dR74

timvandermeij (approved):
Looks good to me after these final changes are made and the tests pass.

dmitryskey:
First of all, thank you very much for such a throughout code review, it helped me to understand my changes a way better :-)

All proposed changes are committed, 2 commits are rebased. I ran all the tests successfully including the rendering the file 'annotation-text-without-popup.pdf' which had a null annotation element.

dmitryskey:
Hi, guys!

What is the next step for this PR? Should it be merged first or I can add another commit?

timvandermeij:
@dmitryskey Sorry, I forgot about this since we're working hard to finish PDF.js version 2.0. You now have a merge commit above, which is usually what we try to avoid. You should use Git rebase instead so you can rebase onto the latest master without a separate merge commit. Other than that, I think code-wise the patch looks good. If the merge commit is gone, I would like to do some final testing an then we can merge it.

@yurydelendik Since this is an API change, should we target it for inclusion in version 2.0 or should we merge it afterwards?

dmitryskey:
Hi!

I merged with the latest code in the master and combined my changes into single commit. All tests were passed.

strichter:
Is this landing at any time soon?

brendandahl:
Now that 2.0 is in pre-release, let's put this in a 2.1 release.
@timvandermeij do you want to finish review and merge?

timvandermeij:
Yes, I'll probably have time for this during this weekend.

dmitryskey:
Can I close this pull request because of merge?

PR #9420, opened 2018-01-30
Take the dictionary, and not just the image data, into account when caching inline images (issue 9398)

### LINKED ISSUE #9398: Regression in issue #8496 with canvas backend
Issue #8496 is about the SVG backend but now also the **canvas** backend has problems with the same PDF. I have tested a few versions: 1.7.225 and 1.8.188 have no problems but newer versions like 1.9.583 (built-in viewer in Firefox), 1.10.96 and 2.0.288 (current demo viewer) don't render the file correctly.

PDF attached in issue #8496

Configuration:
- Web browser and its version: Firefox 57.0.4
- PDF.js version: see above
- Is a browser extension:

1. Open the mentioned PDF with the default canvas backend enabled.

Version 1.7.225 renders all texts fine:
kuva

What went wrong? 
Current demo viewer 2.0.288 has some parts of texts missing:
kuva

Snuffleupagus:
The reason for the bug is that we're only computing a checksum of the image data itself, but completely ignore the inline dictionary. The latter is important, since in practice it's not uncommon for inline images to be identical but use e.g. different ColourSpaces.

There's obviously a couple of different ways that we could compute a hash/checksum of the dictionary.
Initially I tried using `MurmurHash3_64` to compute a hash of the keys/values in the dictionary. Unfortunately this approach turned out to be *way* too slow in practice, especially for PDF files with a huge number of inline images; in particular issue #2618 would regresses quite badly with this solution.

The solution that is instead implemented in this patch, is to compute a checksum of the dictionary contents. While this is a much simpler, not to mention a lot more efficient, solution there's one drawback associated with it:
If the contents of inline image dictionaries are ordered differently, they will not be considered equal with this approach which could thus lead to failures to cache repeated inline images. In practice this doesn't seem to be a problem in any of the PDF files I've tested, and generally I'd rather err on the side of *not* caching given that too aggressive caching can easily lead to rendering bugs.
One small, but somewhat annoying, complication is that by the time `Parser.makeInlineImage` is called, we no longer know the *exact* stream position where the inline image dictionary starts. Having access to that information is crucial here, and the easiest solution I could come up with is to track this in the current `Lexer` instance.[1]

With the patch, we're thus able to fix the referenced issues without incurring large regressions in problematic cases such as issue #2618.

Fixes #9398; also improves/fixes the `issue8823` reference test.

---

[1] Obviously I'd have preferred if this patch could be limited to `Parser.makeInlineImage`, without the need for this "hack", but I'm not sure what that'd look like here.

### DIFF
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -29,7 +29,22 @@ import { Jbig2Stream } from './jbig2_stream';
 import { JpegStream } from './jpeg_stream';
 import { JpxStream } from './jpx_stream';
 
-var MAX_LENGTH_TO_CACHE = 1000;
+const MAX_LENGTH_TO_CACHE = 1000;
+const MAX_ADLER32_LENGTH = 5552;
+
+function computeAdler32(bytes) {
+  let bytesLength = bytes.length;
+  if (bytesLength >= MAX_ADLER32_LENGTH) {
+    throw new Error('computeAdler32: The input is too large.');
+  }
+  let a = 1, b = 0;
+  for (let i = 0; i < bytesLength; ++i) {
+    // No modulo required in the loop if `bytesLength < 5552`.
+    a += bytes[i] & 0xFF;
+    b += a;
+  }
+  return ((b % 65521) << 16) | (a % 65521);
+}
 
 var Parser = (function ParserClosure() {
   function Parser(lexer, allowStreams, xref, recoveryMode) {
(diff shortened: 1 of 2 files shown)


brendandahl:
Do we need the test case if issue 8823 seems to test it as well?

Snuffleupagus:
> Do we need the test case if issue 8823 seems to test it as well?

I suppose not; it's been removed now :-)

Snuffleupagus:
@brendandahl Ping; your comment in https://github.com/mozilla/pdf.js/pull/9420#issuecomment-363622700 has been addressed, any chance that you have time to review this again?

brendandahl (approved):

timvandermeij:
Nice work!

PR #9426, opened 2018-02-01
Attempt to find the next valid marker when encountering invalid image data in `JpegImage.parse` (issue 9425)

### LINKED ISSUE #9425: Unable to decode image: JpegError: JPEG error
File:
[Test.pdf](https://github.com/mozilla/pdf.js/files/1682471/Test.pdf)

Configuration:
  Firefox 58 x64
  Chrome 64 x64

  Windows 10 x64

- PDF.js version: 
  Last one. Pulled from github today ( 2018-01-31).
  Tested also on v1.10.88.
  
- Is a browser extension: No

1. Try to visualize the attached PDF.

What went wrong?
The warnings:
```
PDF 5289ace025eea668263a7252d6500fc5 [1.3 SP C232SF (093960) / -] (PDF.js: -)
util.js:27 Warning: Unable to decode image: JpegError: JPEG error: unknown marker 100
VM28 extensions::schemaUtils:27 Warning: Unable to decode image: JpegError: JPEG error: unknown marker 100
util.js:27 Warning: Dependent image isn't ready yet
util.js:27 Warning: Unable to decode image: JpegError: JPEG error: unknown marker 100
VM28 extensions::schemaUtils:27 Warning: Unable to decode image: JpegError: JPEG error: unknown marker 100
```

Snuffleupagus:
In the JPEG images in the referenced PDF file, the DHT (Define Huffman Tables) segments contain more data than expected based on the length parameter.

Fixes #9425.

### DIFF
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -369,7 +369,7 @@ var JpegImage = (function JpegImageClosure() {
       // Some bad images seem to pad Scan blocks with e.g. zero bytes, skip past
       // those to attempt to find a valid marker (fixes issue4090.pdf).
       if (fileMarker && fileMarker.invalid) {
-        warn('decodeScan - unexpected MCU data, next marker is: ' +
+        warn('decodeScan - unexpected MCU data, current marker is: ' +
              fileMarker.invalid);
         offset = fileMarker.offset;
       }
@@ -389,7 +389,7 @@ var JpegImage = (function JpegImageClosure() {
     // Some images include more Scan blocks than expected, skip past those and
     // attempt to find the next valid marker (fixes issue8182.pdf).
     if (fileMarker && fileMarker.invalid) {
-      warn('decodeScan - unexpected Scan data, next marker is: ' +
+      warn('decodeScan - unexpected Scan data, current marker is: ' +
            fileMarker.invalid);
       offset = fileMarker.offset;
     }
@@ -601,12 +601,12 @@ var JpegImage = (function JpegImageClosure() {
     return component.blockData;
   }
 
-  function findNextFileMarker(data, currentPos, startPos) {
+  function findNextFileMarker(data, currentPos, startPos = currentPos) {
     function peekUint16(pos) {
       return (data[pos] << 8) | data[pos + 1];
     }
 
-    var maxPos = data.length - 1;
+    const maxPos = data.length - 1;
     var newPos = startPos < currentPos ? startPos : currentPos;
 
     if (currentPos >= maxPos) {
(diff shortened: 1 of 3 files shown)


src/core/jpg.js
```diff
@@ -874,6 +874,13 @@ var JpegImage = (function JpegImageClosure() {
               offset -= 3;
               break;
             }
+            let nextFileMarker = findNextFileMarker(data, offset - 2);
+            if (nextFileMarker && nextFileMarker.invalid) {
+              warn('JpegImage.parse - unexpected image data, next marker is: ' +
+                   nextFileMarker.invalid);
```
timvandermeij: If I'm reading https://github.com/Snuffleupagus/pdf.js/blob/f2fed5d67181285a265b05f41e57d99fb2bda249/src/core/jpg.js#L631 correctly, `nextFileMarker.invalid` is actually the old marker (`currentMarker`), correct? Don't you want to print https://github.com/Snuffleupagus/pdf.js/blob/f2fed5d67181285a265b05f41e57d99fb2bda249/src/core/jpg.js#L632 (`newMarker`) here instead?

Snuffleupagus: I was simply following the same pattern as used elsewhere in the code, e.g.
https://github.com/mozilla/pdf.js/blob/29d77dedad84dafdfed04c8ca7d07b680c6cfce3/src/core/jpg.js#L369-L373
and
https://github.com/mozilla/pdf.js/blob/29d77dedad84dafdfed04c8ca7d07b680c6cfce3/src/core/jpg.js#L389-L393

For debugging purposes, I think it's much more interesting to know what the current *incorrect* marker is here (which was the purpose of these warnings).
However, your question made me realize that it's probably a poor choice of words here. So would replacing `next` with `current` here, and elsewhere in the file, perhaps be a better choice?

timvandermeij: Yes, I think that `unexpected data, current marker is:` would be better because I think that's what's being printed.

Snuffleupagus: I've made the changes at all the affected locations; thank you!

Re-reading https://github.com/mozilla/pdf.js/pull/9426#discussion_r165816331, I realize that I probably didn't answer all of your questions. If there something about `findNextFileMarker`, or its usage, that's unclear I'd be happy to try and explain!

timvandermeij: Looks better this way. Now that those warnings are changed, the term `invalid` makes sense to me again: `invalid` is the current invalid marker, and `marker`/`offset` are the new valid marker and its offset in the data stream.

timvandermeij (approved):

timvandermeij:
/botio makeref

(The font rendering on Linux changed after an update, which has nothing to do with this patch.)

PR #9427, opened 2018-02-01
Fallback to the built-in JPEG decoder when browser decoding fails, and attempt to handle JPEG images with DNL (Define Number of Lines) markers (issue 8614)

### LINKED ISSUE #8614: not possible to render (Error during JPEG image loading)

Snuffleupagus:

Fixes #8614.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -80,11 +80,10 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       var colorSpace = dict.get('ColorSpace', 'CS');
       colorSpace = ColorSpace.parse(colorSpace, this.xref, this.resources,
                                     this.pdfFunctionFactory);
-      var numComps = colorSpace.numComps;
-      var decodePromise = this.handler.sendWithPromise('JpegDecode',
-        [image.getIR(this.forceDataSchema), numComps]);
-      return decodePromise.then(function (message) {
-        var data = message.data;
+
+      return this.handler.sendWithPromise('JpegDecode', [
+        image.getIR(this.forceDataSchema), colorSpace.numComps
+      ]).then(function({ data, width, height, }) {
         return new Stream(data, 0, data.length, image.dict);
       });
     },
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -349,7 +349,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
     },
 
     buildPaintImageXObject({ resources, image, isInline = false, operatorList,
-                             cacheKey, imageCache, }) {
+                             cacheKey, imageCache,
+                             forceDisableNativeImageDecoder = false, }) {
       var dict = image.dict;
       var w = dict.get('Width', 'W');
       var h = dict.get('Height', 'H');
@@ -419,28 +420,47 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
         return Promise.resolve();
       }
 
-      var nativeImageDecoderSupport = this.options.nativeImageDecoderSupport;
+      const nativeImageDecoderSupport = forceDisableNativeImageDecoder ?
+        NativeImageDecoding.NONE : this.options.nativeImageDecoderSupport;
       // If there is no imageMask, create the PDFImage and a lot
       // of image processing can be done here.
       var objId = 'img_' + this.idFactory.createObjId();
-      operatorList.addDependency(objId);
-      args = [objId, w, h];
 
       if (nativeImageDecoderSupport !== NativeImageDecoding.NONE &&
           !softMask && !mask && image instanceof JpegStream &&
           NativeImageDecoder.isSupported(image, this.xref, resources,
                                          this.pdfFunctionFactory)) {
         // These JPEGs don't need any more processing so we can just send it.
-        operatorList.addOp(OPS.paintJpegXObject, args);
-        this.handler.send('obj', [objId, this.pageIndex, 'JpegStream',
-                                  image.getIR(this.options.forceDataSchema)]);
-        if (cacheKey) {
-          imageCache[cacheKey] = {
-            fn: OPS.paintJpegXObject,
-            args,
-          };
-        }
-        return Promise.resolve();
+        return this.handler.sendWithPromise('obj', [
+          objId, this.pageIndex, 'JpegStream',
+          image.getIR(this.options.forceDataSchema)
+        ]).then(function() {
+          // Only add the dependency once we know that the native JPEG decoding
+          // succeeded, to ensure that rendering will always complete.
+          operatorList.addDependency(objId);
+          args = [objId, w, h];
+
+          operatorList.addOp(OPS.paintJpegXObject, args);
+          if (cacheKey) {
+            imageCache[cacheKey] = {
+              fn: OPS.paintJpegXObject,
+              args,
+            };
+          }
+        }, (reason) => {
+          warn('Native JPEG decoding failed -- trying to recover: ' +
+               (reason && reason.message));
+          // Try to decode the JPEG image with the built-in decoder instead.
+          return this.buildPaintImageXObject({
+            resources,
+            image,
+            isInline,
+            operatorList,
+            cacheKey,
+            imageCache,
+            forceDisableNativeImageDecoder: true,
+          });
+        });
       }
 
       // Creates native image decoder only if a JPEG image or mask is present.
@@ -456,6 +476,10 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           pdfFunctionFactory: this.pdfFunctionFactory,
         });
       }
+
+      // Ensure that the dependency is added before the image is decoded.
+      operatorList.addDependency(objId);
+      args = [objId, w, h];
 
       PDFImage.buildImage({
         handler: this.handler,
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -357,12 +357,12 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
       if (!(w && isNum(w)) || !(h && isNum(h))) {
         warn('Image dimensions are missing, or not numbers.');
-        return;
+        return Promise.resolve();
       }
       var maxImageSize = this.options.maxImageSize;
       if (maxImageSize !== -1 && w * h > maxImageSize) {
         warn('Image exceeded maximum allowed size and was removed.');
-        return;
+        return Promise.resolve();
       }
 
       var imageMask = (dict.get('ImageMask', 'IM') || false);
@@ -396,7 +396,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
             args,
           };
         }
-        return;
+        return Promise.resolve();
       }
 
       var softMask = (dict.get('SMask', 'SM') || false);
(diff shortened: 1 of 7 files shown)


src/core/evaluator.js
```diff
@@ -419,28 +420,47 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
-          };
-        }
-        return Promise.resolve();
+        return this.handler.sendWithPromise('obj', [
+          objId, this.pageIndex, 'JpegStream',
+          image.getIR(this.options.forceDataSchema)
+        ]).then(function() {
+          // Only add the dependency once we know that the native JPEG decoding
+          // succeeded, to ensure that rendering will always complete.
+          operatorList.addDependency(objId);
```
Rob--W: I don't know this code, so correct me if I say something stupid.

Does the move of `addDependency` introduce any side effects? E.g. is it possible that by deferring the dependency, that at some point the operator list does not observe any dependencies and finishes too early?
If so, is it feasible to ensure that the dependency is resolved, even on failure?

src/core/evaluator.js
```diff
@@ -419,28 +420,47 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
 
-      var nativeImageDecoderSupport = this.options.nativeImageDecoderSupport;
+      const nativeImageDecoderSupport = forceDisableNativeImageDecoder ?
+        NativeImageDecoding.NONE : this.options.nativeImageDecoderSupport;
       // If there is no imageMask, create the PDFImage and a lot
       // of image processing can be done here.
-      var objId = 'img_' + this.idFactory.createObjId();
-      operatorList.addDependency(objId);
-      args = [objId, w, h];
+      const objId = 'img_' + this.idFactory.createObjId();
```
Rob--W: This change is unrelated to the commit. Can you undo it? It might be better to consistently convert the whole function (or even file) from `var` to `let` / `const` in one commit or PR (e.g. for consistency, I would expect `w` and `h` to also be constant, at the very least).

src/core/jpg.js
```diff
@@ -635,7 +655,8 @@ var JpegImage = (function JpegImageClosure() {
   }
 
   JpegImage.prototype = {
-    parse: function parse(data) {
+    parse(data, options = {}) {
+      const dnlScanLines = options.dnlScanLines || null;
```
Rob--W: Why not:

```javascript
parse(data, { dnlScanLines = null }) {
```

src/core/jpg.js
```diff
@@ -839,6 +862,13 @@ var JpegImage = (function JpegImageClosure() {
             break;
 
           case 0xFFDA: // SOS (Start of Scan)
+            numScanSegments++;
+            // A DNL marker (0xFFDC), if it exists, is only allowed at the end
+            // of the first scan segment and may only occur once in an image.
+            // Furthermore, to prevent an infinite loop, do *not* attempt to
+            // check DNL markers during re-parsing of the first scan segment.
+            const checkDNLMarker = numScanSegments === 1 && !dnlScanLines;
```
Rob--W: `dnlScanLines !== null` to avoid infinite recursion when the value is `0`.

src/core/jpg.js
```diff
@@ -839,6 +862,13 @@ var JpegImage = (function JpegImageClosure() {
             break;
 
           case 0xFFDA: // SOS (Start of Scan)
+            numScanSegments++;
```
Rob--W: Do you really need to count? Can't you use a boolean `hasSeenSOS` ;)?

src/core/jpg.js
```diff
@@ -28,6 +28,18 @@ let JpegError = (function JpegErrorClosure() {
   return JpegError;
 })();
 
+let DNLMarkerError = (function DNLMarkerErrorClosure() {
+  function DNLMarkerError(scanLines) {
```
Rob--W: This is an atypical Error constructor. Where is `message`?

Rob--W (changes requested):
Thanks for this PR!
Have you tried jpeg-width-height-mismatch.pdf from #8628 ? If I recall correctly, that is the minimal test case that I created based on the original bug report.

src/core/evaluator.js
Snuffleupagus: This change was reverted.

src/core/evaluator.js
Snuffleupagus: > Does the move of addDependency introduce any side effects?

It shouldn't since we're still adding the dependency *before* the operator here; more below...

> E.g. is it possible that by deferring the dependency, that at some point the operator list does not observe any dependencies and finishes too early?

As far as I can tell, that shouldn't be a problem since the (now async) `buildPaintImageXObject` method won't be resolved until *either* this code-path or the fallback just below has completed successfully. This should thus ensure that the `operatorList` cannot be finished before the image resources are fully parsed and handled.

src/core/jpg.js
Snuffleupagus: I've made the suggested change; however I also added some validation here, to ensure that the parameter is consistent with the JPEG specification.

src/core/jpg.js
Snuffleupagus: (In this case, given that this `Error` is completely internal, it's perhaps not strictly needed.)

Fair point, I've updated the signature of `DNLMarkerError` to address this.

src/core/jpg.js
Snuffleupagus: I think you meant `dnlScanLines === null` here!?
Otherwise it will have the opposite effect, regarding the infinite recursion issue.

Given that zero isn't a valid value, which is now being enforced, I left this comparison intact.

src/core/jpg.js
Snuffleupagus: Perhaps not, but I do find the current state slightly clearer w.r.t. the specification.
With the proposed change you'd need to compare `hasSeenSOS === false` below (before setting `hasSeenSOS = true;`), which seems slightly strange when we've in fact actually encountered a SOS marker by entering this block.

src/display/dom_utils.js
```diff
@@ -480,16 +480,17 @@ class DummyStatTimer {
...
-  img.onerror = (function loadJpegStream_onerrorClosure() {
-    objs.resolve(id, null);
-    warn('Error during JPEG image loading');
+function loadJpegStream(imageUrl) {
+  return new Promise((resolve, reject) => {
+    const img = new Image();
+    img.onload = function() {
+      resolve(img);
+    };
+    img.onerror = function(error) {
```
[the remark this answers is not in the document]
Snuffleupagus: Good point, it's removed now.

Also, in the new version of the PR this code was moved/re-factored such that https://github.com/mozilla/pdf.js/pull/9427#discussion_r165848545 is addressed as well.

Snuffleupagus (commented):
Thank you for the review; I've attempted to address/answer all the comments/questions, please take a look.

> Have you tried jpeg-width-height-mismatch.pdf from #8628 ? If I recall correctly, that is the minimal test case that I created based on the original bug report.

I've tested that one, but it doesn't render in any of the other PDF readers that I tried: Adobe Reader, PDFium in Chrome, nor PDF.js (with this PR).
Comparing that PDF file with the original one, there isn't a DNL marker in that JPEG image so I'm not sure if it's really a good reduced test-case!?

src/display/api.js
```diff
@@ -1818,8 +1817,18 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
+              img.onload = function() {
+                resolve(img);
+              };
+              img.onerror = function() {
+                reject(new Error('Error during JPEG image loading'));
+              };
+              img.src = imageData;
+            }).then((img) => {
+              pageProxy.objs.resolve(id, img);
+            });
```
Rob--W: In case of an error, `pageProxy.objs.resolve` is not called any more. Is that change intentional?

src/core/jpg.js
```diff
@@ -635,7 +657,13 @@ var JpegImage = (function JpegImageClosure() {
...
 
   JpegImage.prototype = {
-    parse: function parse(data) {
+    parse(data, { dnlScanLines = null, } = {}) {
+      if (dnlScanLines !== null && !Number.isInteger(dnlScanLines) &&
+          !(dnlScanLines >= 0x0001 && dnlScanLines <= 0xFFFF)) {
+        warn('JpegImage.parse - invalid "dnlScanLines" parameter: ' +
+             dnlScanLines);
+        dnlScanLines = null;
+      }
```
Rob--W: Since `dnlScanLines` is an internal parameter, and you already enforce that it is greater than 0, is it really necessary to strictly validate the value of `dnlScanLines` ?

Snuffleupagus: I suppose that we can do without this, I'll remove it.

src/display/api.js
Snuffleupagus: Yes, this was intentional considering that when the native loading/decoding fails here we're now falling back to the built-in JPEG decoder. Hence we'll instead add the image to `pageProxy.objs` in the `case 'Image':` branch below, once the built-in decoding has happened.

Rob--W: I see. Can you add a comment after `reject(new Error('Error during JPEG image loading'));` that briefly explains that the error is handled by the fallback logic in `core/evaluator.js`?

Rob--W (approved):
Changes look good to me, with one more nit.

src/display/api.js
Snuffleupagus: Sure, a comment has been added; thank you!

Rob--W:
Since I'm not very familiar with the image processing code, I'd like another pair of eyes before merging it.

@yurydelendik Can you take a look?

Snuffleupagus:
> Since I'm not very familiar with the image processing code, I'd like another pair of eyes before merging it.
> 
> @yurydelendik Can you take a look?

@yurydelendik Ping; do you have time to provide a final round of review here, so that this can land?

yurydelendik (approved):
Look good. Thank you for the patch.

timvandermeij:
Thank you!

Issue #5264, opened 2014-09-04
Highlight showing over text in IE, but works in Chrome/Firefox


msladev:
There is a text that was highlighted but in IE the highlight is showing over the text and hiding the original text. However, in Chrome and Firefox the highlight works as expected. Screenshots are shown below.

In IE:
![hightlight](https://cloud.githubusercontent.com/assets/8659209/4154582/b9649a0c-3462-11e4-9e41-979c03b142fa.PNG)

In Chrome:
![highlight chrome](https://cloud.githubusercontent.com/assets/8659209/4154652/88afe910-3463-11e4-97fc-a940e14a480d.PNG)

Thanks,
MSLA-IT

tbasse:
@msladev, your solution worked perfectly for us until we stumbled upon some PDFs which rendered other image and shape elements with a 50% transparency due to this change.

Example PDF: https://dl.dropboxusercontent.com/u/1652601/ht-1-handbook.pdf

We then moved the workaround to `CanvasGraphics_beginAnnotation()` and only apply it for IE so it only affects the text highlight and only when it needs to.

![screen shot 2015-04-25 at 10 42 43 am](https://cloud.githubusercontent.com/assets/291433/7330805/07fb91d8-eb39-11e4-84c2-bfef0a815d2d.png)

lochstar:
Thanks @tbasse, works great.

ArnoldFabasoft:
We had the same issue in our integration of pdf.js. We implemented the workaround of tbasse for internet explorer. Thanks @tbasse 
Will this fix or an other implementation find a way into pdf.js? I want to modify the original code as less as possible.
Thx Stefan

jeremypress:
@yurydelendik, would you be open to a contribution that implements this workaround? I highly doubt that Microsoft will be addressing this on their end.

timvandermeij:
This workaround is not acceptable for in the core, but it is may be accepted if it can be feature-detected and placed in https://github.com/mozilla/pdf.js/blob/master/src/shared/compatibility.js.

timvandermeij:
Closing IE11/Edge (non-Chromium-based) issues in response to #11211. Please note that, as outlined in https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support, only bugs which completely prevent the library and/or the default viewer from running will be accepted.

PR #9431, opened 2018-02-02
Manually blend when encountering unsupported blend modes

jeremypress:
This PR attempts to improve (but not fully solve) blending modes on unsupported browsers. Notably IE11. This PR is based off of @tbasse 's work in https://github.com/mozilla/pdf.js/issues/5264. 

I wanted to open this up to suggestions before a full regression/unit test pass.

Example 1
 When blend modes are supported:
screen shot 2018-02-02 at 10 01 37 am 

When blend modes are not supported:
annotations no blend

When blend modes are not supported, with my change:
annotations

Example 2
 When blend modes are supported:
screen shot 2018-02-02 at 10 01 30 am

When blend modes are not supported:
blend no blend

When blend modes are not supported, with my change:
blend modes

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -934,6 +934,12 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
             break;
           case 'BM':
             this.ctx.globalCompositeOperation = value;
+            // We only manually blend if the mode is not 'source-over'
+            // which is the only supported blend mode does no blending
+            // See issue https://github.com/mozilla/pdf.js/issues/5264
+            this.blendModeFallback =
+              this.ctx.globalCompositeOperation !== value &&
+              value !== 'source-over';
             break;
           case 'SMask':
             if (this.current.activeSMask) {
@@ -955,6 +961,11 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
             this.tempSMask = null;
             break;
         }
+      }
+
+      if (this.blendModeFallback) {
+        this.ctx.globalAlpha = 0.5;
+        this.blendModeFallback = false;
       }
     },
     beginSMaskGroup: function CanvasGraphics_beginSMaskGroup() {
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -934,12 +934,6 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
             break;
           case 'BM':
             this.ctx.globalCompositeOperation = value;
-            // We only manually blend if the mode is not 'source-over'
-            // which is the only supported blend mode does no blending
-            // See issue https://github.com/mozilla/pdf.js/issues/5264
-            this.blendModeFallback =
-              this.ctx.globalCompositeOperation !== value &&
-              value !== 'source-over';
             break;
           case 'SMask':
             if (this.current.activeSMask) {
@@ -963,9 +957,13 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
         }
       }
 
-      if (this.blendModeFallback) {
-        this.ctx.globalAlpha = 0.5;
-        this.blendModeFallback = false;
+      if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
+        // See checkBlendModeSupport() in compatiblity.js
+        if (this.ctx.blendModeFallback) {
+          // 0.5 is chosen to allow equal visibility for top and bottom content
+          this.ctx.globalAlpha = 0.5;
+          this.ctx.blendModeFallback = false;
+        }
       }
     },
     beginSMaskGroup: function CanvasGraphics_beginSMaskGroup() {
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -958,9 +958,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       }
 
       if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
-        // See checkBlendModeSupport() in compatiblity.js
-        if (this.ctx.blendModeFallback) {
-          // 0.5 is chosen to allow equal visibility for top and bottom content
+        // See checkBlendModeSupport() in compatiblity.js.
+        if (this.ctx.blendModeFallback && this.ctx.globalAlpha === 1) {
+          // 0.5 is chosen to allow equal visibility for top and bottom content.
           this.ctx.globalAlpha = 0.5;
           this.ctx.blendModeFallback = false;
         }
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -851,3 +851,35 @@ PDFJS.compatibilityChecked = true;
 })();
 
 }
+
+// Provides a partial workaround when blend modes are not supported
+// See issue https://github.com/mozilla/pdf.js/issues/5264
+// NOTE: This is not a blend mode polyfill, but a way to see blended content
+// that would otherwise be unviewable
+// Support: IE11
+(function checkBlendModeSupport() {
+  if (!hasDOM) {
+    return;
+  }
+  const canvas = document.createElement('canvas');
+  const ctx = canvas.getContext('2d');
+  ctx.globalCompositeOperation = 'multiply';
+
+  // Check to see if the current browser supports globalCompositeOperation
+  if (ctx.globalCompositeOperation === 'multiply') {
+    return;
+  }
+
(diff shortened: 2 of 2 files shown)


src/display/canvas.js
```diff
@@ -934,6 +934,12 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
             break;
           case 'BM':
             this.ctx.globalCompositeOperation = value;
+            // We only manually blend if the mode is not 'source-over'
+            // which is the only supported blend mode does no blending
+            // See issue https://github.com/mozilla/pdf.js/issues/5264
+            this.blendModeFallback =
+              this.ctx.globalCompositeOperation !== value &&
```
jeremypress: If the mode is unsupported, `ctx. globalCompositeOperation` won't update with that value. This allows us to avoid a browser specific check. If this is considered too brittle, I can move this check to compatibility and use the method described in the issue.

src/display/canvas.js
```diff
@@ -956,6 +962,11 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
             break;
         }
       }
+
+      if (this.blendModeFallback) {
+        this.ctx.globalAlpha = 0.5;
```
jeremypress: Since we aren't supporting the actual blend mode here, 50% was chosen to allow equal visibility of the top and bottom content

src/display/canvas.js
Snuffleupagus: > This allows us to avoid a browser specific check.

That's sort of beside the point here, at least in my opinion. Browser detection is usually used *only* as a last resort in `src/shared/compatibility.js`, in rare cases where feature detection won't work correctly.

This PR would still mean adding compatibility hacks to the regular code-base, which is something we always want to avoid for a couple of reasons.
First of all, it adds overhead (even though it's probably small in this case) to all modern browsers, even those that have proper blend mode support.
Secondly, these sort of hacks increase the maintenance burden of the code in question, which we generally want to avoid.

Snuffleupagus: > This allows us to avoid a browser specific check.

That's sort of beside the point here, at least in my opinion. Browser detection is usually used *only* as a last resort in `src/shared/compatibility.js`, in rare cases where feature detection won't work correctly.

This PR would still mean adding compatibility hacks to the regular code-base, which is something we always want to avoid for a couple of reasons.
First of all, it adds a certain amount of overhead to all modern browsers, even those that have proper blend mode support.
Secondly, these sort of hacks increase the maintenance burden of the code in question, which we generally want to avoid.

jeremypress: hey @Snuffleupagus, thanks for the feedback.

My intention was that in many cases where blend modes are used, even viewing critical content is not possible, (see example 2). It is definitely not correct behavior, but I believe that it will make the document more readable in many cases, at the expense of a neutral change in other cases. I think browser overhead is minimal here, as I've only added one `if` check and an operation (setting global alpha) that is done in many other places.

If you're not comfortable with this kind of fallback, would writing a polyfill for blend modes on unsupported browsers be feasible? To me, it sounds like even more overhead. I would still need to do feature/browser detection, and I would be adding a lot more math based and risky code to the code base. My thought process for that kind of a fix would be:

1. Select the area of whatever asset/shape/image we are about to add on our existing canvas
2. Calculate the new color for that area of the canvas using the formula for that mode
3. Render the new asset and apply the color to the area.

Snuffleupagus: > I think browser overhead is minimal here, as I've only added one if check and an operation (setting global alpha) that is done in many other places.

It's not simply about run-time overhead, it's in equal measure relevant to the mental overhead these hacks will add to the main code-base.

---

Please note that you might want to wait for feedback from of others contributors here, since I suppose opinions might vary as to if/how we want to address this.

However, even if the consensus would be that we're OK with the current hack, I think that there's a couple of changes that would be required here:
 - First of all, the added code needs to be enclosed in preprocessor tags to prevent it from being bundled into anything except `GENERIC` builds; i.e.
```javascript
if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
  // The new code goes here...
}
```
 - Secondly, you'd need to *very clearly* state in the comments that this is a polyfill for IE11 (in addition to the issue link).
 - Thirdly, do the cheapest comparison first, i.e. `this.blendModeFallback = value !== 'source-over' && this.ctx.globalCompositeOperation !== value;`.

src/display/canvas.js
timvandermeij: Please add that as a comment here.

src/display/canvas.js
timvandermeij: I agree with the above. If it is possible to place it in `compatibility.js` I'd highly prefer that, but if that's not possible/makes the code much more difficult, then this is acceptable provided that the above is done, i.e., there is a good comment and it's not bundled unnecessarily.

jeremypress: Moved the check to compatibility.js as efficiently as I could. I now override the setter for the property in question, and this check is only done once.

src/display/canvas.js
jeremypress: added

jeremypress:
Hey @timvandermeij and @Snuffleupagus, 
Any thoughts on the new approach?

Snuffleupagus (changes requested):
Looking closer at the surrounding code, I'm no longer sure that adding this kind of "hack" (rather than a proper polyfill) is a particularly good idea here.
Also I still cannot escape the feeling that trading one kind of rendering error for *another* isn't conceptually correct, and that it may simply result in a different class of rendering bugs being reported.

As mentioned inline, always overwriting the `globalAlpha` value may end up *causing* rendering regressions in some PDF files (obviously only in IE11, but still).
Please keep in mind that we only run tests in Firefox and Chrome, hence we'd have no way of knowing if this causes regressions.

src/shared/compatibility.js
```diff
@@ -883,3 +883,35 @@ PDFJS.compatibilityChecked = true;
 })();
 
 }
+
+// Provides a partial workaround when blend modes are not supported
+// See issue https://github.com/mozilla/pdf.js/issues/5264
+// NOTE: This is not a blend mode polyfill, but a way to see blended content
+// that would otherwise be unviewable
+// Support: IE11
+(function checkBlendModeSupport() {
```
Snuffleupagus: Also, this code won't be bundled into `GENERIC` builds as-is, it would need to be placed *before* the `URL` polyfill above.

src/display/canvas.js
```diff
@@ -956,6 +956,15 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
             break;
         }
       }
+
+      if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
```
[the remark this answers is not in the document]
timvandermeij: I'm also a bit nervous about this. Would checking if `globalAlpha` is set to a non-default value before overwriting it address this concern? I do think that it would be best to consolidate this workaround in `compatiblity.js` and not have any changes in `canvas.js` to avoid any risks.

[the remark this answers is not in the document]
jeremypress: @Snuffleupagus, The reason why I override the alpha value in `canvas.js` as opposed to the compatibility file is to avoid my value being overriden by a later `ca` instruction. I noticed that the order of instruction type is not guaranteed. 

I agree with @timvandermeij that I should include a check for a current global alpha value before setting hard coding it myself. An example of this problem would be `BM = 'overlay'` and `ca = 0.25`. My current implementation would overwrite this to 0.5, but since all content would be visible it should be left as is. 

I'll attempt to access this from the compatibility file but it may only be accessible after we have processed all of our instructions in `canvas.js` (where I am currently overriding the alpha value now). I'll include a new commit with the suggested style fixes + my proposed safety change.

src/shared/compatibility.js
```diff
@@ -883,3 +883,35 @@ PDFJS.compatibilityChecked = true;
...
+
+  // Check to see if the current browser supports globalCompositeOperation
+  if (ctx.globalCompositeOperation === 'multiply') {
+    return;
+  }
+
+  /* eslint-disable accessor-pairs, object-shorthand */
+  Object.defineProperty(CanvasRenderingContext2D.prototype,
+    'globalCompositeOperation', {
+    set: function(value) {
```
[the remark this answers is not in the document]
jeremypress: When I switched to the shorthand notation the function isn't being called for some reason. I was following the `mdn` example from here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set.

jeremypress:
Closing this as it's not a complete solution. Thanks for the discussion!

PR #9432, opened 2018-02-04
Add webpack.overwrite.js to change webpack.config

rchaser53:
I want the easy way to change webpack.config for 'gulp generic'.
So I add webpack.overwrite.js to change webpack.config. It's a useful for CI.

Use like below.
```
// webpack.overwrite.js
exports.overwrite = {
  devtool: false  // suppress to emit sourcemap
}
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -131,7 +131,7 @@ function createWebpackConfig(defines, output) {
   var skipBabel = bundleDefines.SKIP_BABEL ||
                   process.env['SKIP_BABEL'] === 'true';
 
-  return {
+  return Object.assign({
     output: output,
     plugins: [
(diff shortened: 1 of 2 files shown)


timvandermeij:
I'm not convinced that we should have this extra way to pass in configuration. The advantage of the Gulpfile is that it's easy to read/edit and the configuration is in one place. This extra file would make it harder to find out where configuration is set. Custom deployments should just edit the file; if there are options that really need to be edited often by a lot of people, another approach could for example be to make a command line parameter such as `gulp generic --no-source-maps` since that also makes it easier to script. Given these reasons, I'm closing this PR.

rchaser53:
you may know gulp cannot use option such as  ``` --no-source-maps ```.
so what you says needs to change build tool from gulp to something.
i think it's not a smart idea.

PR #9433, opened 2018-02-04
[Firefox addon] Change the minimum supported version to Firefox Nightly, and remove no longer needed fallback code

Snuffleupagus:

/cc @rvandermeulen Since this blocks the next PDF.js update.


timvandermeij (approved):

timvandermeij:
Thank you for doing this! I completely agree that it's just not worth the effort anymore to keep the extension compatible with older Firefox versions. This makes maintenance easier and unblocks the upstream PDF.js update, and if the extension is converted to a WebExtension things will become easier too.

PR #9436, opened 2018-02-04
Use `setPDFNetworkStreamFactory` in the `helloworld` and `svgviewer` examples

timvandermeij:
This patch fixes a regression from #9363, causing the examples not to load anymore.


timvandermeij:
Trivial.

PR #9444, opened 2018-02-06
locally hosted viewer should allow any file location

AlexejStukov:
Included 'qrc://' for Qt resource files

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1482,7 +1482,7 @@ let PDFViewerApplication = {
 
 let validateFileURL;
 if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
-  const HOSTED_VIEWER_ORIGINS = ['null',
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Please refer to PR #6916 for background on `HOSTED_VIEWER_ORIGINS`, and the related URL validation.

The default set of `HOSTED_VIEWER_ORIGINS` are the bare minimum that are necessary for https://mozilla.github.io/pdf.js/web/viewer.html to work. For any sort of custom implementation of the default viewer, you should edit `HOSTED_VIEWER_ORIGINS` as required yourself depending on the situation.

Hence I don't think that this is a change that we need/should make in PDF.js library itself, which is why I'm closing the PR.

AlexejStukov:
Sorry did not see that. The comment  [// Hosted or local viewer, allow for any file locations](https://github.com/mozilla/pdf.js/blob/master/web/app.js#L1494) had me thinking the current behaviour was unintentional.

PR #9454, opened 2018-02-07
gracefully handling exceptions during annotation creation

Mobiletainment:
without handling exceptions during annotation creation, a single malformed annotation (e.g. http:myurl.com) leads to an error on Edge, resulting in the document not being rendered at all.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -323,11 +323,16 @@ var Page = (function PageClosure() {
       var annotationRefs = this.getInheritedPageProp('Annots') || [];
       for (var i = 0, n = annotationRefs.length; i < n; ++i) {
         var annotationRef = annotationRefs[i];
-        var annotation = AnnotationFactory.create(this.xref, annotationRef,
-                                                  this.pdfManager,
-                                                  this.idFactory);
-        if (annotation) {
-          annotations.push(annotation);
+        
+        try {
+          var annotation = AnnotationFactory.create(this.xref, annotationRef,
+                                                    this.pdfManager,
+                                                    this.idFactory);
+          if (annotation) {
+            annotations.push(annotation);
+          }
+        } catch (ex) {
+          console.error('PDFDocument: cannot create annotation', ex);
         }
       }
       return shadow(this, 'annotations', annotations);


Snuffleupagus:
Please provide a PDF file that can be used to evaluate/test the proposed solution, and can be used as a basis for either a unit or reference test. (As it stands, the patch looks more like a work-around than a proper way of handling the error.)

Mobiletainment:
Thank you for your feedback! I wanted to provide a PDF in the first place, but unfortunately I don't have the rights for it. This particular document was created with Adobe Acrobat 9.0 Paper Capture Plug-in using OCR regnotion, hence the invalid URI. I'll try to create a reference PDF which causes the same issue and will get back to you.

Mobiletainment:
I was able to re-create a document with an invalid URI. It should cause an error when viewing with PDF.js using Microsoft Edge:
[This is a test document with an invalid URL.pdf](https://github.com/mozilla/pdf.js/files/1703137/This.is.a.test.document.with.an.invalid.URL.pdf)

Snuffleupagus:
> I was able to re-create a document with an invalid URI. It should cause an error when viewing with PDF.js using Microsoft Edge:
> This is a test document with an invalid URL.pdf

That file works fine in both Firefox and Chrome, so it would appear that this is limited to Edge.
Executing `new URL('http:Wesastianstr.9a')` in the dev-tools of either browsers works just fine without errors, and a valid URL object is returned. Note that this is equivalent to what PDF.js does, see https://github.com/mozilla/pdf.js/blob/43f1f96b101c1fa8a7b839659a2e505275a82e42/src/shared/util.js#L346-L351

According to https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#Browser_compatibility, the `URL` constructor should be supported in Edge and [otherwise we polyfill it](https://github.com/mozilla/pdf.js/blob/43f1f96b101c1fa8a7b839659a2e505275a82e42/src/shared/compatibility.js#L225).

Furthermore, as can been seen above, this code already runs inside of a `try ... catch` block in PDF.js which should thus be catching any potential errors from the `URL` constructor. If that doesn't work, that rather seems to point to a bug in Edge. Please note that as a general rule, we don't want to add compatibility hacks to the main code-base.

Based on the above, I'm going to close this PR.

PR #9459, opened 2018-02-07
Respect workerSrc if set

### LINKED ISSUE #9443: Issues with patch that removed `disableWorker`
Thanks for all the great work!

I'm running into a slight issue with my particular setup. The recent patch to remove `PDFJS.disableWorker` (https://github.com/mozilla/pdf.js/commit/56a8c934dde6cce806b1d74fa80c181eb30bbb96) updated one of the checks for a fake worker to use

```
 if (typeof Worker !== 'undefined' && !isWorkerDisabled &&
          !getMainThreadWorkerMessageHandler()) {
```

I prefetch the worker file while users are not on a page that requires PDF preview, and when they do come to a document preview, I point the workerSrc to that same file that I prefetched earlier. Because of browser caching, the worker can be loaded from the cache. With this patch, `getMainThreadWorkerMessageHandler()` will return a handler because I prefetched the worker file at an earlier time and then this block returns false and goes into `_setupFakeWorker()`. Can this last handler check be entirely removed?

On a related, but slightly different note, would you folks be open to a PR that swaps these two in https://github.com/mozilla/pdf.js/blob/7fbeeebbffe68787117aade4b3c00c4728aac81c/src/display/api.js#L1204?

```
if (typeof workerSrc !== 'undefined') {
      return workerSrc;
}
if (getDefaultSetting('workerSrc')) {
     return getDefaultSetting('workerSrc');
}
```

The library I build can sometimes be run in an application that uses RequireJS (in which `workerSrc` will be set to requirejs.url(...something)). I'd explicitly like to keep using the workerSrc I define using PDFJS.workerSrc = 'someUrl'.

tonyjin:
Respect user-defined `workerSrc` over internal overrides.

Resolves #9443.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1202,11 +1202,11 @@ var PDFWorker = (function PDFWorkerClosure() {
   let nextFakeWorkerId = 0;
 
   function getWorkerSrc() {
-    if (typeof workerSrc !== 'undefined') {
-      return workerSrc;
-    }
     if (getDefaultSetting('workerSrc')) {
       return getDefaultSetting('workerSrc');
+    }
+    if (typeof workerSrc !== 'undefined') {
+      return workerSrc;
     }
     if (typeof PDFJSDev !== 'undefined' &&
         PDFJSDev.test('PRODUCTION && !(MOZCENTRAL || FIREFOX)') &&


timvandermeij:
This change looks reasonable to me, but since I'm not too familiar with the worker logic, I would like a second pair of eyes here just to be sure I'm not missing anything.

Snuffleupagus (approved):
The current situation, where the `workerSrc` option is being silently ignored in *some* cases seems far from ideal to me.

I suppose that the only caveat here would be if there's API consumers that managed to set an *invalid* `workerSrc`, and where the current logic thus "saved" them.
However I'm not too worried about that, and besides my impression from questions/issues over the years is that it's *a lot* more common for users to have trouble because they did *not* set the `workerSrc` option at all.

The above is thus a somewhat long-winded way of saying that I think this change is fine. Thanks for the patch!

tonyjin:
Thanks! I appreciate the context.

PR #9465, opened 2018-02-09
Attempt to unify the `disableRange`/`contentLength` handling in the various network streams

Snuffleupagus:
First of all, note how in both `fetch_stream.js` and `node_stream.js` we always overwrite the `this._contentLength` property even when the response headers doesn't actually contain any (valid) length information. This could thus result in the `length` parameter, as passed to the network stream, being completely ignored despite having no better information available.
Secondly, in `node_stream.js` the `this._isRangeSupported` property wasn't always updated correctly based on the response headers.

### DIFF
--- a/src/display/node_stream.js
+++ b/src/display/node_stream.js
@@ -304,11 +304,9 @@ class PDFNodeStreamFullReader extends BaseFullReader {
           disableRange: this._disableRange,
         });
 
-      if (allowRangeRequests) {
-        this._isRangeSupported = true;
-      }
+      this._isRangeSupported = allowRangeRequests;
       // Setting right content length.
-      this._contentLength = suggestedLength;
+      this._contentLength = suggestedLength || this._contentLength;
 
       this._filename = extractFilenameFromHeader(getResponseHeader);
     };
--- a/src/display/fetch_stream.js
+++ b/src/display/fetch_stream.js
@@ -74,14 +74,14 @@ class PDFFetchStreamReader {
     this._withCredentials = source.withCredentials;
     this._contentLength = source.length;
     this._headersCapability = createPromiseCapability();
-    this._disableRange = source.disableRange;
+    this._disableRange = source.disableRange || false;
     this._rangeChunkSize = source.rangeChunkSize;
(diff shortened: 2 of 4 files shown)


src/display/node_stream.js
```diff
@@ -304,11 +304,9 @@ class PDFNodeStreamFullReader extends BaseFullReader {
           disableRange: this._disableRange,
         });
 
-      if (allowRangeRequests) {
-        this._isRangeSupported = true;
-      }
+      this._isRangeSupported = allowRangeRequests;
```
timvandermeij: This change should also be done in `src/display/network.js` above.

Snuffleupagus: That's not strictly necessary, since there's a subtle difference between `network.js` respectively `fetch_stream.js`/`node_stream.js` regarding the initialization of the properties.
In `network.js` everything is explicitly disabled until we've determined that support exist, however in e.g. `fetch_stream.js` you could probably assume that stream support exists (by design). This is a bit counter-intuitive at first, but I'm not sure if we want to change that particular instance!?

For `network.js`, note these definitions https://github.com/mozilla/pdf.js/blob/9767b8a9f1ae7cf5250da648deb62b5f1c4d7292/src/display/network.js#L337-L338 and then how there're explicitly set in https://github.com/mozilla/pdf.js/blob/9767b8a9f1ae7cf5250da648deb62b5f1c4d7292/src/display/network.js#L370-L380
Now compare with e.g. `fetch_stream.js`, where these definitions are used https://github.com/mozilla/pdf.js/blob/9767b8a9f1ae7cf5250da648deb62b5f1c4d7292/src/display/fetch_stream.js#L83-L84 and then https://github.com/mozilla/pdf.js/blob/9767b8a9f1ae7cf5250da648deb62b5f1c4d7292/src/display/fetch_stream.js#L116-L123.

timvandermeij (approved):

timvandermeij:
Thank you for the patch!

src/display/node_stream.js
timvandermeij: I see. Let's indeed not change that now.

PR #9453, opened 2018-02-07
Upstream the changes from: Bug 1432992 - Remove definitions of Ci, Cr, Cc, and Cu

Snuffleupagus:
Please see:
 - https://bugzilla.mozilla.org/show_bug.cgi?id=1432992
 - https://hg.mozilla.org/integration/autoland/rev/b38d59f71915

Fixes #9449; ~~currently blocked on a new version of `eslint-plugin-mozilla`, see https://github.com/mozilla/pdf.js/issues/9449#issuecomment-363709486.~~

/cc @rvandermeulen

PR #9434, opened 2018-02-04
Upstream the changes from: Bug 1339461 - Convert foo.indexOf(...) == -1 to foo.includes() and implement an eslint rule to enforce this

Snuffleupagus:
- Polyfill `String.prototype.includes` using core-js

   See https://github.com/zloirock/core-js#ecmascript-6-string.

 - Upstream the changes from: Bug 1339461 - Convert foo.indexOf(...) == -1 to foo.includes() and implement an eslint rule to enforce this

   Yet another case where PDF.js code was modified in `mozilla-central` without the changes happening in the GitHub repo first; *sigh*.
If we don't upstream at least the changes in `extensions/firefox/`, any future update of PDF.js in `mozilla-central` will be blocked.

   Please see:
    - https://bugzilla.mozilla.org/show_bug.cgi?id=1339461
    - https://hg.mozilla.org/mozilla-central/rev/d5a5ad1dbbf2

/cc @rvandermeulen Since this blocks the next PDF.js update.

PR #9032, opened 2017-10-15
Simplify the check, and remove the warning, for the `nativeImageDecoderSupport` API parameter

Snuffleupagus:
As discussed in PR https://github.com/mozilla/pdf.js/pull/8982#discussion_r142392423.

**TODO**
 - [x] Check if we possibly need to include [core-js](https://github.com/zloirock/core-js) polyfills for `Object.values` and/or `Array.includes`.

PR #8138, opened 2017-03-08
Get rid of `element.removeChild(element.firstChild)` usage (bug 1345253)

Snuffleupagus:
Instead of just upstreaming the changes from [bug 1345253](https://bugzilla.mozilla.org/show_bug.cgi?id=1345253) as-is, it seemed better to simply get rid of the loops altogether and use the same approach as in `PDFViewer`/`PDFThumbnailViewer`.

The PR also fixes an existing issue with `PDFOutlineViewer.reset` not actually resetting *all* the state it should.

Fixes #8139.

PR #9468, opened 2018-02-10
Enable the `mozilla/avoid-removeChild` and `mozilla/use-includes-instead-of-indexOf` ESLint rules globally

Snuffleupagus:
~~*Currently depends on the `eslint-plugin-mozilla` version upgrade included in PR #9453.*~~

 - Enable the `mozilla/avoid-removeChild` ESLint rule globally

   This rule is available from https://www.npmjs.com/package/eslint-plugin-mozilla, and is enforced in mozilla-central. Note that we have a polyfill for `ChildNode.remove()` and that most cases have already been fixed, see PRs #8056 and #8138.

 - Enable the `mozilla/use-includes-instead-of-indexOf` ESLint rule globally

   This rule is available from https://www.npmjs.com/package/eslint-plugin-mozilla, and is enforced in mozilla-central. Note that we have the necessary `Array`/`String` polyfills and that most cases have already been fixed, see PRs #9032 and #9434.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,6 +24,7 @@
   "rules": {
     // Plugins
     "mozilla/avoid-removeChild": "error",
+    "mozilla/use-includes-instead-of-indexOf": "error",
 
     // Possible errors
     "for-direction": "error",
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,6 +4,10 @@
     "sourceType": "module",
   },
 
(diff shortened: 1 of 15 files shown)


timvandermeij:
This patch can be rebased now that the other patch is merged.

timvandermeij (approved):

timvandermeij:
Nice code simplification!

PR #9470, opened 2018-02-11
Ensure that `JpegImage.getData` returns the correct data length when `forceRGBoutput == true` (issue 4888)

### LINKED ISSUE #4888: jpg.js returns incorrect data arrays from conversion
I was trying to understand how jpg.js does the conversion when i noticed they all overwrite the input data array. 
Sounds like a good optimization idea for me, but then I noticed that
`convertYcckToRgb` and `convertCmykToRgb` do the same,
however the result array should be only 3/4th of the original data array length.

shouldn't they before returning data do a data.length = offset to truncate the array?

Snuffleupagus:
With PDF.js version `2.0` we'll only support browsers with built-in `TypedArray` functionality, hence there doesn't seem to be any good reason not to implement this now.

Fixes #4888.

*Edit:* Please note that I've also run the full test-suite locally, using `nativeImageDecoderSupport: 'none',` instead of https://github.com/mozilla/pdf.js/blob/be2674a0e8d4bf239e08eb33a42d1f4d15fde7ee/test/driver.js#L347 to ensure that everything still works correctly with this patch.

### DIFF
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -1083,7 +1083,8 @@ var JpegImage = (function JpegImageClosure() {
                0.116935020465145) +
           k * (-0.000343531996510555 * k + 0.24165260232407);
       }
-      return data;
+      // Ensure that only the converted RGB data is returned.
+      return data.subarray(0, offset);
     },
 
     _convertYcckToCmyk: function convertYcckToCmyk(data) {
@@ -1140,7 +1141,8 @@ var JpegImage = (function JpegImageClosure() {
                193.58209356861505) -
           k * (22.33816807309886 * k + 180.12613974708367);
       }
-      return data;
+      // Ensure that only the converted RGB data is returned.
+      return data.subarray(0, offset);
     },
 
     getData: function getData(width, height, forceRGBoutput) {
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
@Rob--W Is this something that you'd be willing, time permitting of course, to review?
Please keep in mind that "no" is a perfectly valid answer :-)

Rob--W:
Can you add a regression test?

Snuffleupagus:
> Can you add a regression test?

There's existing test-cases, one example is https://github.com/mozilla/pdf.js/blob/master/test/pdfs/cmykjpeg.pdf which is an [`eq` test](https://github.com/mozilla/pdf.js/blob/ea3d8450d218dd53619c72c934547fe57f22530f/test/test_manifest.json#L1195-L1201), that hits this code-path and are still rendered correctly.
What sort of additional test coverage did you have in mind here?

Rob--W:
Without a test that fails in the current master and passes with the PR, how do we know and can we ensure that this PR actually fixes something?

Snuffleupagus:
@Rob--W Ping; is the added unit-test the sort of additional test coverage you were after here?

Rob--W:
I confirmed that the test fails before the PR, and passes thereafter. Moreover, the `subarray` function has low overhead because it does not copy the buffer (only the view is modified), so it is unlikely for this change to reduce the performance.

Thanks for the fix and test, I'm going to merge it!

PR #9472, opened 2018-02-11
Improve the interface for `PDFFindController`

timvandermeij:
In pull request #9369, I noticed the following question:

> There are many other non-underscore methods in `PDFFindController`; should they be part of the public API?

Looking into this, I found that the private methods in this file are indeed not properly indicated. This patch series fixes that in order to clarify the interface.


Snuffleupagus (approved):
Looks good to me; thank you!

PR #9473, opened 2018-02-12
Remove the non-`a.click()` case from the `DownloadManager`

Snuffleupagus:
With PDF.js version `2.0`, all the browsers we intend to support implements `a.click()`; see https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click#Browser_compatibility.

*Edit:* Smaller diff with https://github.com/mozilla/pdf.js/pull/9473/files?w=1.


timvandermeij (approved):

timvandermeij:
Good find, thanks!

PR #9040, opened 2017-10-17
Remove the `enableHandToolOnLoad` preference migration code in `web/pdf_cursor_tools.js`

Snuffleupagus:

PR #9475, opened 2018-02-13
Remove the actual `enableHandToolOnLoad` preference (PR 9040 follow-up)

Snuffleupagus:
This should have been removed as part of PR #9040, but was simply overlooked.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #9479, opened 2018-02-13
[api-major] Refactor viewer components initialization to reduce their dependency on the global `PDFJS` object

### LINKED ISSUE #4919: fix webgl parameter

### LINKED ISSUE #9322: Remove undefined PDFJS.isValidUrl property
The `isValidUrl` function in `src/dom_utils.js` was deprecated in 71a781ee5c9e548c16d8e612f41fbc45f1082fc5 and removed in 0f90d5130c9b4c847d5efecac9a727c151156883.
However, the source code still contains references to it in display/global.js:

https://github.com/mozilla/pdf.js/blob/fe30d4d468237864eb5a0609658a8372bfaa3ca0/src/display/global.js#L18-L19

https://github.com/mozilla/pdf.js/blob/fe30d4d468237864eb5a0609658a8372bfaa3ca0/src/display/global.js#L70

This `isValidUrl` import and `PDFJS.isValidUrl` property should be removed.

Snuffleupagus:
Since PR #9185 is, at this point in time, the main remaining blocker for PDF.js `2.0`, I've tried to split that one up a bit to hopefully be able to move forward here. As indicated in https://github.com/mozilla/pdf.js/pull/9185#issuecomment-365308082 this PR will result in a less than ideal intermediate state, but hopefully that's acceptable.

Fixes #4919.
Fixes #9322.

@timvandermeij Sorry to ask, but I'm hoping that you might be willing (time permitting) to review this?


yurydelendik (approved):
Sorry for long response. I looked the overall approach and I like it. I did not do any nitpicking, if @timvandermeij agrees/glances too, we can land it. Thank you for the patch.

timvandermeij:
Really good. I'll take a close look in the coming days.

timvandermeij (approved):

timvandermeij:
Thank you for improving this! It's quite a bit to review, but I really like how this simplifies things. The subtle improvements for the text layer are also good to have.

Snuffleupagus:
@yurydelendik, @timvandermeij Thank you both for helping review this!

Rob--W:
I coincidentally came across this change, and I believe that the decision to not migrate preferences is taken too lightly. The Chrome extension allows administrators to control PDF.js via the preferences, and it is important that these identifiers are stable. Even if you don't want to write migration code, I'd appreciate if you ping me whenever there is a backwards-incompatible preference change. This is also necessary so I can mention the new preference name in release notes.

This is the second time that I notice breaking changes to preferences. For the previous one (hand tool -> hand+cursor tool), I wrote this migration code: https://github.com/mozilla/pdf.js/commit/19549bb7d6191cc6e79cc806b0fbf9c0598db192 (similar logic needs to be applied here).
@Snuffleupagus Do you want to write migration code for your change here, or should I do it?

Snuffleupagus:
> [...] and I believe that the decision to not migrate preferences is taken too lightly.

Generally I'd agree that we should always try to migrate preferences, however I think there's extenuating circumstances in this case!
Given that these changes will be release with PDF.js `2.0`, which is an increase in the *major* version number, backwards incompatible changes ought to be allowed/expected. Otherwise, as far as I'm concerned, there doesn't seem to be much reason to increase the major version number.
Furthermore, please note that a number of backwards incompatible API changes have already been made, and there's more to come as well; see https://github.com/mozilla/pdf.js/projects/5.

> I'd appreciate if you ping me whenever there is a backwards-incompatible preference change. This is also necessary so I can mention the new preference name in release notes.

That's a fair point, sorry about that!

> Do you want to write migration code for your change here, or should I do it?

Unfortunately I don't have time, and my sole focus right now is to get https://github.com/mozilla/pdf.js/projects/5 over the line to avoid holding up future releases any longer.
Also, as indicated above, I'm not totally convinced that migrating is necessary here.

Rob--W:
The Chrome extension is built off master, so it's already using 2.x.
As the maintainer of the Chrome extension, internal API changes are not really that significant. What matters is the external, user-facing changes, such as preferences.

I'll create a new issue and assign it to me.

Rob--W:
Eventually these changes will also make it into Firefox, won't it? Firefox is also already using 2.x, shouldn't these preferences be migrated in Firefox too?

Snuffleupagus:
> Eventually these changes will also make it into Firefox, won't it?

That's correct.

> Firefox is also already using 2.x, shouldn't these preferences be migrated in Firefox too?

Considering that the preferences are *not* that easily accessible in the UI, but rather available through `about:config`, I wouldn't worry too much about it since nothing will outright "break" here.
Assuming that a user manually set `disableTextLayer`, the worst thing that will happen is that the text is now selectable/searchable again. As for `enhanceTextSelection`, that is/was still somewhat experimental and consequently I wouldn't worry about it.

PR #9480, opened 2018-02-14
[api-major] Move the worker related options from the global `PDFJS` object

Snuffleupagus:
*This PR extracts the worker related changes from PR #9185.*

Please note that, as mentioned in the first commit message, it unfortunately seems to me that we cannot convert *all* of the worker options to parameters on `getDocument`/`PDFWorker` instead (since we probably need to maintain a reasonably easy way to configure workers).
However the `workerPort`/`workerSrc` options can at least be moved from the `PDFJS` object, which does seem worthwhile to me since it gets us closer to be able to remove `PDFJS` completely.

*Edit:* Once this lands I'll rebase/rewrite PR #9185 on top of it, which should thus take care of the major outstanding work required for version `2.0`.

@yurydelendik, @timvandermeij Sorry to bother you again, but any chance that (either of) you could review this PR to unblock the final removal of all options from the global `PDFJS` object (in #9185)?
Also, based on a quick look, it might not actually be that difficult to remove the global `PDFJS` object completely once this and then #9185 lands.

*Edit 2:* Based on this PR, the final round of changes would be https://github.com/Snuffleupagus/pdf.js/compare/refactor-worker-options...Snuffleupagus:refactor-api-options, which converts all remaining options *and* removes `PDFJS` completely.

### DIFF
--- a/examples/acroforms/acroforms.js
+++ b/examples/acroforms/acroforms.js
@@ -15,7 +15,8 @@
 
 'use strict';
 
-PDFJS.workerSrc = '../../node_modules/pdfjs-dist/build/pdf.worker.js';
+PDFJS.GlobalWorkerOptions.workerSrc =
+  '../../node_modules/pdfjs-dist/build/pdf.worker.js';
 
 var DEFAULT_URL = '../../test/pdfs/f1040.pdf';
 var DEFAULT_SCALE = 1.0;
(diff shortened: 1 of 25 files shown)


timvandermeij (approved):

timvandermeij:
Awesome work! This is a really clean solution and I'm glad that we're progressing so well with this. Thank you for working on this!

Snuffleupagus:
@timvandermeij Thanks so much for reviewing this; we're getting really close to releasing version `2.0` now!
I did try to split the work into many separate commits, which I hope did helped (at least somewhat) when reviewing.

PR #9485, opened 2018-02-15
Ability to handle URI annotations

dlnsk:
Added URI annotation for links where schema is different to 'http' (like 'media://' etc.). It allow to handle clicks on those links and do something in viewer (display media player for example).

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -45,7 +45,12 @@ class AnnotationElementFactory {
 
     switch (subtype) {
       case AnnotationType.LINK:
-        return new LinkAnnotationElement(parameters);
+        let data = parameters.data;
+        if (data.unsafeUrl !== undefined && data.url === undefined && data.dest === undefined) {
+          return new URIAnnotationElement(parameters);
+        } else {
+          return new LinkAnnotationElement(parameters);
+        }
 
       case AnnotationType.TEXT:
         return new TextAnnotationElement(parameters);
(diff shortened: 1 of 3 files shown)


Snuffleupagus (commented):
I'm not at all convinced that we should have this behaviour in the standard PDF.js library, since it will cause unsupported/non-working links to become visible in the viewer by default; please note that *not* rendering links that does nothing when clicked was a very conscious decisioned in PR #8865.

For a custom implementation, it's probably easier to handle this use-case by either modifying the following helper function to cover the needed protocol https://github.com/mozilla/pdf.js/blob/4f5fb782378e321f50c8047f3109310359c904c0/src/shared/util.js#L320
or by changing the rendering condition
https://github.com/mozilla/pdf.js/blob/4f5fb782378e321f50c8047f3109310359c904c0/src/display/annotation_layer.js#L269-L270

Furthermore given that this won't ever be used in the default viewer, it would mean adding (basically) unused code which which we always want to avoid (given that code needs to be maintained).

timvandermeij:
I fully agree with the comment above. Moreover, `URIAnnotation` is not an annotation type as defined in the PDF specification, which we want to follow closely, and `unsafeUrl` should not be used in the core code (it's only made available for custom deployments to use). I'm therefore closing this PR because of these reasons, but if you want to use this in a custom deployment I can suggest to follow the steps above to make it easier. Thanks.

Issue #9440, opened 2018-02-05
Document that pdfjs-dist is a "nightly" build


Rob--W:
https://github.com/mozilla/pdfjs-dist and https://www.npmjs.com/package/pdfjs-dist does not clearly show that the release is a continuous build. The documentation, sourced from https://github.com/mozilla/pdf.js/blob/a7c35025fed8beb8f9b93688fff40497c7ad2de0/external/dist/README.md should probably be changed to make this more clear. And https://mozilla.github.io/pdf.js/getting_started/ does not clearly state that pdfjs-dist is not the stable release build.

Here is an example of confusion that was caused by this: https://github.com/mozilla/pdf.js/pull/9385#issuecomment-363030784

brendandahl:
Or we could send npm releases instead.

wojtekmaj:
npm releases, please! I'd use `next` tag for all the beta ones and no tag for stable releases. Pretty much industry standard at the moment.

brendandahl:
I did some work on this over in https://github.com/mozilla/botio-files-pdfjs/pull/22.

One thing I'm not sure how to handle is fixing our current npm setup. I'm thinking I'll retag the 2.0 branch as `next` and publish an older 1.0 version as stable until we release. This may lead to some weirdness for anyone who's already pulled a 2.0 release from npm though. Open to suggestions!

wojtekmaj:
Can't answer for the whole community but in my opinion, the change is very much worth this one-time commotion.

wojtekmaj:
Oh yeah. We can just tag non-final releases with `next` instead of `latest` starting with official 2.0. So:

1. `^1.x.xxx` users will still to be on newest 1.x pre-release released - no change
2. `^2.x.xxx` users will receive final stable 2.0 and will stop receiving pre-releases
3. `latest` users will receive final stable 2.0 and will stop receiving pre-releases

So that's pretty sweet IMO.

We could also give people from point 1 stuck on 1.x pre-release a hint using [npm deprecate](https://docs.npmjs.com/cli/deprecate) by deprecating all 1.x pre-releases with a message like "You're likely unknowingly using unstable PDF.js release. Please downgrade to the newest stable version, 1.9.xxx or upgrade to all-new 2.x version. Upgrade guide here: http://example.com/pdfjsupgradeguide".

timvandermeij:
The bot change is now merged.

[...]

PR #9493, opened 2018-02-18
[api-major] Move the remaining options from, and completely remove, the global `PDFJS` object

Snuffleupagus:
This is the final (major) piece of work required before we can release PDF.js version `2.0`, the other remaining TODOs are basically of the nice-to-have kind.
I've tried to split it into as many commits as possible to aid reviewing; in some cases ignoring whitespace changes with `?w=1` should also help.

@yurydelendik, @timvandermeij Please check if the new way of handling (default) viewer options makes sense here!


timvandermeij:
I think I'll have time during the weekend to review this.

src/display/api.js
```diff
@@ -26,6 +26,7 @@ import {
   RenderingCancelledException, StatTimer
 } from './dom_utils';
 import { FontFaceObject, FontLoader } from './font_loader';
```
timvandermeij: It looks like the commit message of this commit is copy-pasted from the first one ;-)

src/display/api.js
```diff
@@ -169,6 +169,9 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *   of PDF files. When enabled, and if the server supports partial content
  *   requests, then the PDF will be fetched in chunks.
  *   The default value is `false`.
```
timvandermeij: It looks like the commit message of this commit is copy-pasted from the first one ;-)

src/display/api.js
```diff
@@ -178,6 +178,9 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
...
  *   NOTE: It is also necessary to disable streaming, see above,
  *         in order for disabling of pre-fetching to work correctly.
```
timvandermeij: It looks like the commit message of this commit is copy-pasted from the first one ;-)

examples/components/singlepageviewer.js
```diff
@@ -36,16 +36,16 @@ var SEARCH_FOR = ''; // try 'Mozilla';
...
-var pdfSinglePageViewer = new PDFJS.PDFSinglePageViewer({
+var pdfSinglePageViewer = new pdfjsDistWebPdfViewer.PDFSinglePageViewer({
   container: container,
   linkService: pdfLinkService,
 });
 pdfLinkService.setViewer(pdfSinglePageViewer);
 
 // (Optionally) enable find controller.
-var pdfFindController = new PDFJS.PDFFindController({
+var pdfFindController = new pdfjsDistWebPdfViewer.PDFFindController({
```
timvandermeij: For consistency with the other examples, could you add a trailing comma on the next line?

web/pdf_viewer.component.js
```diff
@@ -25,31 +26,28 @@ import { DownloadManager } from './download_manager.js';
...
-PDFJS.DefaultAnnotationLayerFactory = DefaultAnnotationLayerFactory;
-PDFJS.PDFHistory = PDFHistory;
-PDFJS.PDFFindController = PDFFindController;
-PDFJS.EventBus = EventBus;
-
-PDFJS.DownloadManager = DownloadManager;
-PDFJS.ProgressBar = ProgressBar;
-PDFJS.GenericL10n = GenericL10n;
-PDFJS.NullL10n = NullL10n;
+const pdfjsVersion = PDFJSDev.eval('BUNDLE_VERSION');
```
timvandermeij: I'm not entirely sure why these two constants are added here because they weren't in the code before and they are not exported. What am I missing here?

timvandermeij (approved):
Looks good to me with the comments addressed. I'm really impressed with how much work has been put in this patch! The new way of handling app options is much more consistent in my opinion.

web/app_options.js
```diff
@@ -13,13 +13,110 @@
...
+    kind: OptionKind.VIEWER,
+  },
+  enableWebGL: {
+    /** @type {boolean} */
+    value: false,
+    kind: OptionKind.VIEWER,
+  },
+  externalLinkRel: {
+    /** @type {string} */
+    value: null,
```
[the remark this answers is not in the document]
Snuffleupagus: > Moreover, since this is a string type, I would at least expect `''`.

You're absolutely correct that this should be a string, however an empty string wouldn't work here; see: https://github.com/mozilla/pdf.js/blob/5f98f9b8f10d68df6cbd331b5e135e167c71829b/src/display/dom_utils.js#L317

I've changed this particular line to explicitly list the "default" link rel instead; but see below as well.

> If this is to avoid a dependecy on code from src, can we add a comment somewhere in this file to indicate that?

Correct, I view this (new) file somewhat similar to https://github.com/mozilla/pdf.js/blob/master/web/default_preferences.json. Even though we can, as opposed to a `.json` file, import constants/types from elsewhere here I nonetheless wanted to avoid that pattern. I've tried to add a comment to that effect.

web/app_options.js
```diff
@@ -122,6 +122,29 @@ const defaultOptions = {
...
+
+  workerPort: {
+    /** @type {Object} */
+    value: null,
+    kind: OptionKind.WORKER,
+  },
+  workerSrc: {
+    /** @type {string} */
+    value: (typeof PDFJSDev === 'undefined' || !PDFJSDev.test('PRODUCTION') ?
+           '../src/worker_loader.js' : '../build/pdf.worker.js'),
```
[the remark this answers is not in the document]
Snuffleupagus: Fixed.

web/app_options.js
```diff
@@ -13,13 +13,110 @@
...
+    kind: OptionKind.VIEWER,
+  },
+  locale: {
+    /** @type {string} */
+    value: (typeof navigator !== 'undefined' ? navigator.language : 'en-US'),
+    kind: OptionKind.VIEWER,
+  },
+  maxCanvasPixels: {
+    /** @type {number} */
+    value: viewerCompatibilityParams.maxCanvasPixels || 16777216,
```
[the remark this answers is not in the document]
Snuffleupagus: I'd very much prefer to *not* do that, for a couple of reasons.

First of all, the "source of truth" in the new code is the JSDocs added to the viewer components and API methods, rather than this file itself; again compare with https://github.com/mozilla/pdf.js/blob/master/web/default_preferences.json.
In this particular case, the documentation is already present in https://github.com/mozilla/pdf.js/blob/5f98f9b8f10d68df6cbd331b5e135e167c71829b/web/base_viewer.js#L59-L61 and https://github.com/mozilla/pdf.js/blob/5f98f9b8f10d68df6cbd331b5e135e167c71829b/web/base_viewer.js#L59-L61

Second of all, if we start adding that kind of comments in this file we'll probably end up doing it for *a lot* more options. That would add maintenance burden, since each JSDoc comment would then be duplicated twice in the code-base. Furthermore, this file would become much longer and quite unwieldy in my opinion.

web/app.js
```diff
@@ -1512,6 +1515,9 @@ if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
 
 function loadFakeWorker() {
   return new Promise(function(resolve, reject) {
+    if (!GlobalWorkerOptions.workerSrc) {
```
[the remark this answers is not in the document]
Snuffleupagus: With the new way of setting the API/worker options, in `PDFViewerApplication.open`, we haven't actually set the `workerSrc` by the time this code runs (which happens during initialization of the viewer *itself*).

Previously the `workerSrc` was set earlier in the viewer initialization, see: https://github.com/mozilla/pdf.js/blob/5f98f9b8f10d68df6cbd331b5e135e167c71829b/web/app.js#L53
However, with all this refactoring I wanted to unify how/when the API/worker options are set, which necessitated this change.
If your question is why we don't just set the `workerSrc` directly here, and simply skip this check, it's to be slightly more accommodating to a custom deployment where `globalWorkerOptions.workerSrc` was manually set.

src/display/api.js
Snuffleupagus: I've tried to trim the commit messages a bit.

src/display/api.js
Snuffleupagus: I've tried to trim the commit messages a bit.

examples/components/singlepageviewer.js
Snuffleupagus: Fixed.

web/pdf_viewer.component.js
Snuffleupagus: This was probably not entirely relevant to the rest of the patch, however it addresses an in my opinion existing shortcoming.
Note how both https://github.com/mozilla/pdf.js/blob/master/src/pdf.js and https://github.com/mozilla/pdf.js/blob/master/src/pdf.worker.js contains this information (without exporting it). That can be very handy if you're looking at the built files and want to know which version/build it corresponds to.

Snuffleupagus:
@timvandermeij Thanks a lot for taking the time to review this, much appreciated :-)
I've attempted to address and/or answer your review comments, care to take another look?

timvandermeij (approved):

timvandermeij:
It's really good to finally get rid of the global `PDFJS` object and making PDF.js version 2.0 ready for release. Thank you so much!

agsingletrack:
Hi,

I couldn't raise an issue on the pdfjs-dist repo.

We were on version 2.0.376 of pdfjs-dist and a later patch release broke the build for us as it removed the PDFJS export — we had an import line of `import {PDFJS as PDFJSReader} from 'pdfjs-dist/web/pdf_viewer'`. 

How does the versioning of pdfjs-dist work? Why was this breaking change made in a patch release?

brendandahl:
@agsingletrack see #9440

PR #9494, opened 2018-02-18
Windows drive letter

### LINKED ISSUE #9482: Windows path changes drive letter...
## Configuration:
- Web browser and its version: Node 8.9.x
- Operating system and its version: Windows Server 2012
- PDF.js version: pdfjs-dist@^2.0.269
- Is a browser extension: No

1. Working directory on `E:\`
2. `await PDF.getDocument("C:\\Windows\\TEMP\\b6f1c1cd-f811-4b84-8116-5adfe4bd973b")`

Loads the PDF document.

## What went wrong?

Note: drive letter is now E:\
```
ENOENT: no such file or directory, lstat 'E:\\Windows\\TEMP\\b6f1c1cd-f811-4b84-8116-5adfe4bd973b'
```

## Notes

I don''t know if this is corrected with `pdfjs-dist@2.0.374` which seems to have been published today.

I was able to work around the issue by encoding to a `file://` url.  But wanted to submit the problem upstream.

```
// work around drive change issue with windows on the server... C:\ became E:/
if (process.platform === 'win32' && /^[a-z]\:/i.test(filePath)) {
  filePath = `file:///${filePath
    .split(/[\\\/]+/g)
    .map(encodeURIComponent)
    .join('/')}`;
}
```

pardeepshokeen:
fixes : #9482 
adding 'file:///' before windows absolute path, like changing  'D:/pdfFile.pdf' to  'file:///D:/pdfFile.pdf'

### DIFF
--- a/src/display/node_stream.js
+++ b/src/display/node_stream.js
@@ -28,14 +28,30 @@ import {
 
 const fileUriRegex = /^file:\/\/\/[a-zA-Z]:\//;
 
+function parseUrl(sourceUrl) {
+  let parsedUrl = url.parse(sourceUrl);
+  if (parsedUrl.protocol === 'file:' || parsedUrl.host) {
+    return parsedUrl;
+  }
+  // Prepending 'file:///' to Windows absolute path.
+  if (/^[a-z]:[/\\]/i.test(sourceUrl)) {
+    return url.parse(`file:///${sourceUrl}`);
+  }
+  // Changes protocol to 'file:' if url refers to filesystem.
+  if (!parsedUrl.host) {
+    parsedUrl.protocol = 'file:';
+  }
+  return parsedUrl;
+}
+
 class PDFNodeStream {
   constructor(source) {
     this.source = source;
-    this.url = url.parse(source.url);
+    this.url = parseUrl(source.url);
     this.isHttp = this.url.protocol === 'http:' ||
                   this.url.protocol === 'https:';
     // Check if url refers to filesystem.
-    this.isFsUrl = this.url.protocol === 'file:' || !this.url.host;
+    this.isFsUrl = this.url.protocol === 'file:';
     this.httpHeaders = (this.isHttp && source.httpHeaders) || {};
 
     this._fullRequest = null;


timvandermeij:
The regular expression should be cached at the top of the file like https://github.com/pardeepshokeen/pdf.js/blob/c11d8efd7ff486b884caf41d97095b11a599193f/src/display/node_stream.js#L29.

@Rob--W Should this patch revert the changes from the previous patch or are those still good to have?

src/display/node_stream.js
```diff
@@ -30,6 +30,10 @@ const fileUriRegex = /^file:\/\/\/[a-zA-Z]:\//;
 
 class PDFNodeStream {
   constructor(source) {
+    // prepend 'file:///' to windows absolute path
+    if (source.url.match(/^[a-z]:[/\\]/i)) {
+      source.url = `file:///${source.url}`;
+    }
```
Rob--W: Don't overwrite the `source.url` variable, use a separate variable instead. It might be a good idea to add a helper function that runs your logic, and then calls `url.parse`. Then replace the `url.parse` call below with a call to your function.

src/display/node_stream.js
```diff
@@ -30,6 +30,10 @@ const fileUriRegex = /^file:\/\/\/[a-zA-Z]:\//;
 
 class PDFNodeStream {
   constructor(source) {
+    // prepend 'file:///' to windows absolute path
```
Rob--W: "Prepend" and "Windows" should be capitalized, and the sentence should end with a period.

Rob--W:
Caching the RegExp is not really necessary, it's not going to be a bottleneck here.

Also the code from the previous patch should be reverted indeed.

src/display/node_stream.js
```diff
@@ -30,6 +30,10 @@ const fileUriRegex = /^file:\/\/\/[a-zA-Z]:\//;
...
+    if (source.url.match(/^[a-z]:[/\\]/i)) {
```
Rob--W: Use `/^[a-z]:[/\\]/i.test(source.url)` instead of `source.url.match(/.../)`.

Rob--W (changes requested):

timvandermeij:
Good. The change to be reverted is https://github.com/mozilla/pdf.js/commit/8e278d9a458d8d65a963025bceb8381238f545cc#diff-35bcb37ef640f9436dd67f5077c08b55, but it looks like we need to keep https://github.com/mozilla/pdf.js/commit/32080f10816e7671d85330ef827722f9e3c8b1c6#diff-35bcb37ef640f9436dd67f5077c08b55.

pardeepshokeen:
The problem here is, when source.url is like "file:///C:/file.pdf" then after url.parse its path becomes '/C:/file.pdf' and first slash need to be removed as done  in #9323 .

but here when source.url is like 'C;/file.pdf' then after parsing path become /file.pdf which converted to '(working dir ) :/file.pdf' . 

So, I think  instead of prepending 'file:///' to url, adding drive letter to path (which is protocol after url.parse) should be fine.  and there will be no need of reverting previous path.

Rob--W:
Ah I see. The two issues are different after all. There is no need to revert anything, just make the requested changes and it's good to merge.

Since you're making a helper function anyway, let's try to enforce that the URL is always meaningful, i.e. in the helper function detect that if the host is missing, that the protocol should be switched to a `file:///`-URL (this would replace the regexp in your current patch: first you try to parse the input as an URL, and if the host is empty, prepend `file:///` and retry):

Then remove the empty-host check from:

https://github.com/mozilla/pdf.js/blob/99060e2485cacb8513e7bc50d88b70101069b1c9/src/display/node_stream.js#L38

(FWIW, Node.js supports WHATWG URLs with `file://`, but that is still experimental, so let's not use that for now.)

Rob--W (approved):

Rob--W:
Looks good to me. Please squash all commits so there is only one commit to merge.

I think that the current and new implementation does not support UNC paths (`\\hostname\path\to\file.pdf`), but let's not worry about that. If anyone cares they will open a new ticket.

src/display/node_stream.js
```diff
@@ -28,14 +28,30 @@ import {
 
 const fileUriRegex = /^file:\/\/\/[a-zA-Z]:\//;
 
+function parseUrl(sourceUrl) {
+   var parsedUrl = url.parse(sourceUrl);
```
timvandermeij: Please make sure that the indenting in this function is two spaces, not three. Also, use `let` instead of `var`.

pardeepshokeen:
hi @Rob--W . I have done requested changes and squashed the commits. Will you please review this.

Rob--W:
Thanks for the changes and squashing the commits. This looks good!

I suggest that you choose a more descriptive commit message in the future. "helper function to parse url" is not wrong, but doesn't provide much information about the commit's purpose. Something like "Support paths with drive letters on Windows" would be more descriptive, for example.

In any case, thanks again for this fix!

Issue #9498, opened 2018-02-20
ie11 file origin does not match viewer's origin


Barzou:
Hi,

PDF.JS : latest release (v1.9.426) 
Tested on IE11 (11.192.16299.0)
Use case : I get pdf js sources and I do a **gulp minified.**

When I open my viewer, it works on all Browser except Internet explorer.
I got this error : "file origin does not match viewer's origin"

Inside **pdf.viewer.js** script (yeah, i read minified js =P), the part
`if(new URL(e,window.location.href).origin!==t)throw new Error("file origin does not match viewer's")`
**Throw an error** 
because
**new URL(window.location.href).origin is undefined**

I saw on an issue where you explained that you use a polyfill for javaScript URL object (https://github.com/mozilla/pdf.js/issues/9358) but it doesn't seems to work on internet explorer 11.

To help you, I tried this one https://github.com/lifaon74/url-polyfill/blob/master/url-polyfill.js which seems to work great on IE11.

I'll do a temporary work around to avoid this error on my own project (literally skip throw error).

Rob--W:
The URL polyfill in src/shared/compatibility.js overrides the `URL` constructor if `origin` is not supported.
The default demo viewer works fine for me when I open the viewer in IE 11.1000.17074.0 (and also an older version, IE11.0.9600.17728).

This is therefore not an issue with the default viewer.

I also ran `gulp minified` from the master branch, and simulated a broken URL API with the following test case:

```html
<script>delete URL.prototype.origin</script>
<script src=pdf.js></script>
<script>document.write(new URL(location.href).origin)</script>
```

and the test passed (localhost was being printed, instead of `undefined`).

Did you load pdf.js *before* pdf.viewer.js?

Barzou:
You're right.

I tested my use case with pdf.js included and it works perfectly.
I wasn't sure about what scripts should be included, because some script are loaded inside library and it worked with other browser.

But.. the online demo https://mozilla.github.io/pdf.js/web/viewer.html include pdf.js and I didn't notice.

Mea culpa.

Barzou:
Still doesn't work on IE11, I have a second problem. Yesterday I tried :
`new URL(window.location.href).origin`
And this line works fine. 

But when a parameter is added inside pdf.viewer.js, it doesn't seems to work.
I added some logs to pdf.viewer.js : 
``` JS
if (new URL(e, window.location.href).origin !== t){
[...]

PR #9505, opened 2018-02-23
Don't block origin-less blob:-URLs in hosted viewer

Rob--W:
This is to work around a bug in IE11 (and IE10, earlier versions of IE do not support `URL.createObjectURL` for creating `blob:`-URLs anyway) - #9498 .

I could have checked for an empty or null `origin`, but since the browser's Same-origin policy already blocks cross-origin `blob:`-URL requests, I decided to have simpler code and just rely on the browser's same-origin policy enforcement for access control of `blob:`-URLs.


yurydelendik (approved):
looks good to me. thanks

timvandermeij:
Thank you for improving this!

Issue #6990, opened 2016-02-14
add paper size to document information


ousia:
Would it be possible that you add paper size to document information?

This is important to know whether the document uses letter or A4 paper size (among others).

berenddeschouwer:
And can it include portrait/landscape?

timvandermeij:
What is the expected behavior if multiple page sizes/orientations are used, i.e., some pages are A4 portrait, but some are A3 landscape for example?

yurydelendik:
and also we need to deal with internationalization and localization, e.g.
 - "A4" or "Letter" needs to be translated
 - 10cm×15cm shall be in inches for US

berenddeschouwer:
Re: rotation.  I'm going to check what #8043 does for me.  That may fix my problem.  It looks like I really need to read #8043 and #6103 properly, and make sure I'm on a late release with those fixes applied.


Re: multi-size.  Dunno if any print backends support that without splitting the printout into multiple print jobs.  If you don't split the print job, you have "print on biggest" or "scale to smallest" as sane options.

Cups includes an media size header with every print job, so Linux and MacOS are limited to one-size per print-job, as far as I know.


Re: localisation.
The printer has a list of possible and available paper sizes.
The OS print dialog knows the locale of the user, and a list of printers.
The OS print dialog doesn't know the original size.
The pdf viewer knows the locale, but not a list of printers, so not a list of available paper.
The pdf viewer knows the original PDF size.
The pdf generator knows the server and client locales.
The user hates "pc load letter"

yurydelendik:
>Re: rotation. I'm going to check what PR 8043 does for me. 
>Re: multi-size. Dunno if any print backends support that without splitting the printout i..
>Re: localisation... The printer has a list of possible and available paper sizes...

@berenddeschouwer I don't think any of the above related to OP

ousia:
> What is the expected behavior if multiple page sizes/orientations are used, i.e., some pages are A4 portrait, but some are A3 landscape for example? How do other viewers do that?

@timvandermeij, SumatraPDF displays the size of the browsed page (the one that has the number in the toolbar).

>and also we need to deal with internationalization and localization, e.g.
>
> * "A4" or "Letter" needs to be translated
> * 10cm×15cm shall be in inches for US

@yurydelendik, I don’t think `A4` should be translated.

And as far as i know, page dimensions are specified in points inside the PDF document. So PDF.js should use the default dimension unit from OS.

[...]

PR #9508, opened 2018-02-24
Add paper size to document information/properties

pal03377:
First step fixing issue #6990

As discussed there, it makes sense to have the paper size in the document properties dialog.

Here is a basic start that displays the width and height of the first page in the PDF in the document properties. It does not convert this to a local unit (mm, cm, inch, ...) and does not try to find a matching standard document size like A4. Still, it is a start and works.

As suggested in #6990, I initially wanted the paper size of the currently viewed page to be shown. There were two problems with this approach:

1. I couldn't find an easy and elegant way to get the current page in [pdf_document_properties.js](https://github.com/mozilla/pdf.js/blob/master/web/pdf_document_properties.js) that didn't feel 'hacky'.
2. The document properties are loaded once (when they are first opened) and then cached (not changed in the HTML anymore). To avoid the issue that when opening the document properties on page 1, closing them and scrolling to page 2, then opening them again, the page size of page 1 would still be shown, I would have needed to disable the caching. This is something I think needs further discussion as I don't know exactly how this would affect the performance (my guess: not much).

So now it shows the paper size of the first page. If someone wants to implement the current page view, the code is easily expandable.

Next steps to a beautiful "page size experience" (not in this PR):

1. Translate the string "Page size" to other languages
2. Convert width and height to inches/cm/mm/...? depending on the localisation settings
3. find out the usual page sizes in the user's country, then try to find a match (e.g. A4, CD, ...)

I tried to make my code as clean and documented as possible and to adapt my own coding style to the other code. However, since this is my first PR, it's certainly possible that I'm doing something that is not matching the usual workflow in open source. I would be happy to receive feedback about what I could do better in the future. Thank you.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -627,6 +627,23 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
       return this.transport.getMetadata();
     },
     /**
+     * @param {number} pageNumber The page number to get the page size from.
+     * The first page is 1, which is also the default page used.
+     * @return {Promise} A promise that is resolved with an dict containing the
+     * width and height in inches.
+     */
+    getPageSizeInches(pageNumber) {
+      pageNumber = pageNumber || 1;
+      return this.getPage(pageNumber).then((page) => {
+        const [x1, y1, x2, y2] = page.view;
+        // convert values from user units to inches
+        return {
+          width: (x2 - x1) / 72 * page.userUnit,
+          height: (y2 - y1) / 72 * page.userUnit,
+        };
+      });
+    },
+    /**
      * @return {Promise} A promise that is resolved with a TypedArray that has
      * the raw data from the PDF.
      */
--- a/web/pdf_document_properties.js
+++ b/web/pdf_document_properties.js
@@ -79,9 +79,14 @@ class PDFDocumentProperties {
           contentDispositionFilename || getPDFFileNameFromURL(this.url),
           this._parseFileSize(this.maybeFileSize),
           this._parseDate(info.CreationDate),
-          this._parseDate(info.ModDate)
+          this._parseDate(info.ModDate),
+          this.pdfDocument.getPageSizeInches().then((pageSizeInches) => {
+            return this._parsePageSize(pageSizeInches);
+          }),
+
         ]);
-      }).then(([info, metadata, fileName, fileSize, creationDate, modDate]) => {
+      }).then(([info, metadata, fileName, fileSize,
+                creationDate, modDate, pageSize]) => {
         freezeFieldData({
           'fileName': fileName,
           'fileSize': fileSize,
@@ -95,6 +100,8 @@ class PDFDocumentProperties {
           'producer': info.Producer,
           'version': info.PDFFormatVersion,
           'pageCount': this.pdfDocument.numPages,
+          'pageSizeInch': pageSize.inch,
+          'pageSizeMM': pageSize.mm,
         });
         this._updateUI();
 
@@ -210,6 +217,33 @@ class PDFDocumentProperties {
       size_mb: (+(kb / 1024).toPrecision(3)).toLocaleString(),
       size_b: fileSize.toLocaleString(),
     }, '{{size_mb}} MB ({{size_b}} bytes)');
+  }
+
+  /**
+   * @private
+   */
+  _parsePageSize(pageSizeInches) {
+    if (!pageSizeInches) {
+      return Promise.resolve([undefined, undefined]);
+    }
+    const sizes_two_units = {
+      'width_in': Math.round(pageSizeInches.width * 100) / 100,
+      'height_in': Math.round(pageSizeInches.height * 100) / 100,
+      // 1in = 25.4mm; no need to round to 2 decimals for mm
+      'width_mm': Math.round(pageSizeInches.width * 25.4 * 10) / 10,
+      'height_mm': Math.round(pageSizeInches.height * 25.4 * 10) / 10,
+    };
+    return Promise.all([
+      this.l10n.get('document_properties_page_size_in',
+        sizes_two_units, '{{width_in}} in × {{height_in}} in'),
+      this.l10n.get('document_properties_page_size_mm',
+        sizes_two_units, '{{width_mm}} mm × {{height_mm}} mm'),
+    ]).then(([parsedPageSizeInches, parsedPageSizeMM]) => {
+      return Promise.resolve({
+        inch: parsedPageSizeInches,
+        mm: parsedPageSizeMM,
+      });
+    });
   }
 
   /**
--- a/l10n/en-US/viewer.properties
+++ b/l10n/en-US/viewer.properties
@@ -89,6 +89,13 @@ document_properties_creator=Creator:
 document_properties_producer=PDF Producer:
 document_properties_version=PDF Version:
 document_properties_page_count=Page Count:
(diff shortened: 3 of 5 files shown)


pal03377:
How are translations usually added in open source projects?
For example, if I want the German translation "Seitenformat" to be added, what do I usually do? Or is there some other system that collects translations from the community and then adds them in one bigger PR?

pal03377:
example screenshot: page size

Here is an example screenshot. ;-)

web/pdf_document_properties.js
```diff
@@ -79,9 +79,11 @@ class PDFDocumentProperties {
           contentDispositionFilename || getPDFFileNameFromURL(this.url),
           this._parseFileSize(this.maybeFileSize),
           this._parseDate(info.CreationDate),
-          this._parseDate(info.ModDate)
+          this._parseDate(info.ModDate),
+          this.pdfDocument.getPageSize().then(this._parsePageSize.bind(this)),
```
yurydelendik: don't use bind() here, use arrow function

web/pdf_document_properties.js
```diff
@@ -212,6 +215,19 @@ class PDFDocumentProperties {
...
+   * @private
+   */
+  _parsePageSize(pageSize) {
+    if (!pageSize) {
+      return Promise.resolve(undefined);
+    }
+    return this.l10n.get('document_properties_page_size_pt', {
+      width: pageSize[0],
+      height: pageSize[1],
+    }, '{{width}}pt × {{height}}pt');
```
yurydelendik: Most of users don't know what "pt" is, make it dual "in" and "mm" instead

src/display/api.js
```diff
@@ -626,6 +626,21 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
...
+     * The first page is 1, which is also the default page used.
+     * @return {Promise} A promise that is resolved with an array of the
+     * width and height in the user space units - [width, height].
+     */
+    getPageSize(pageNumber) {
+      pageNumber = pageNumber || 1;
+      return this.getPage(pageNumber).then((page) => {
+        return [
+          page.pageInfo.view[2] - page.pageInfo.view[0], // width
+          page.pageInfo.view[3] - page.pageInfo.view[1]  // height
```
yurydelendik: Don't use internal `pageInfo`, use `view` directly. I recommend hoisting it above, e.g. via `const [x1,y1,x2,y2] = page.view;`. Also, [user unit](https://github.com/mozilla/pdf.js/blob/master/src/display/api.js#L866) is not used.

pal03377:
@yurydelendik Thanks for your feedback.
I implemented your suggestions. Now the user unit is used, and everything can be in inches or mm.

yurydelendik:
> and everything can be in inches or mm.

It's better to show in both units, since you can recognize  something from one unit but not from another, examples:

- It easy to tell that "8.5 ⨉ 11.0 in" is a letter, but "215.9 ⨉ 279.4 mm" meaningless
- It easy to tell that "210 ⨉ 297 mm" is A4, but "8.3 ⨉ 11.7 in" meaningless

And it does not matter which country/locale you in.

pal03377:
Interesting approach... OK, I'll rewrite it to be like that.
(Both Adobe Reader and Sumatra PDF only display one unit.)
How should that look?
`210 ⨉ 297 mm / 8.3 ⨉ 11.7 in`
or `8.3 ⨉ 11.7 in / 210 ⨉ 297 mm`
or `8.3 ⨉ 11.7 in, 210 ⨉ 297 mm`
or `210 ⨉ 297 mm, 8.3 ⨉ 11.7 in`
or something completely different?
And, should the units be displayed after each number (like in `210 mm ⨉ 297 mm / 8.3 in ⨉ 11.7 in`) or only after the second number (like in the examples above)?

yurydelendik:
> or something completely different?

it can be on two lines, two localization strings.

> should the units be displayed after each number

yes, that's fine

> Adobe Reader and Sumatra PDF only display one unit

They have access to OS internationalization settings (and also named paper sizes). I don't know if JavaScript has access to OS units of length.

pal03377:
Here you are. Now it is split into inches and mm and displayed ond under another. :-)

src/display/api.js
```diff
@@ -626,6 +626,23 @@ var PDFDocumentProxy = (function PDFDocumentProxyClosure() {
...
+     */
+    getPageSizeInches(pageNumber) {
+      pageNumber = pageNumber || 1;
+      return this.getPage(pageNumber).then((page) => {
+        const [x1, y1, x2, y2] = page.view;
+        // convert values from user units to inches
+        return [
+          (x2 - x1) / 72 * page.userUnit, // width
+          (y2 - y1) / 72 * page.userUnit  // height
+        ];
```
yurydelendik: no comments needed, return an object here: `return { width: ..., height: ... };`

web/pdf_document_properties.js
```diff
@@ -79,9 +79,14 @@ class PDFDocumentProperties {
...
-          this._parseDate(info.ModDate)
+          this._parseDate(info.ModDate),
+          this.pdfDocument.getPageSizeInches().then((pageSizeInches) => {
+            return this._parsePageSize(pageSizeInches);
+          }),
+
         ]);
-      }).then(([info, metadata, fileName, fileSize, creationDate, modDate]) => {
+      }).then(([info, metadata, fileName, fileSize,
+                creationDate, modDate, pageSize]) => {
```
yurydelendik: use `[ pageSizeInch, pageSizeMM ]` instead of `pageSize`; or deconstruct in such way latter.

yurydelendik (commented):
Please squash comments for the final review. (Can you also add a final screenshot?)

pal03377:
screenshot inches and millimeters page size

Screenshot as requested

pal03377:
@yurydelendik I fear that I don't really understand how to do a squash. I did research and found what it is useful for and I also already tried the described steps to do it, but it doesn't look right and I fear to do something wrong here. What can I do? (I'll keep trying...)

yurydelendik:
@pal03377 Standard `git rebase -i c894fadb~1` worked for me -- I updated you branch, pull it locally. One comments was not addressed https://github.com/mozilla/pdf.js/pull/9508#discussion_r172311179 -- access `[0]` or `[1]` looks too magical.

pal03377:
I hope I got what you wanted me to do right; now I'm using an JS object with properties `inch` and `mm` instead of a "magical" array.

pal03377:
@yurydelendik What are the next steps here?

yurydelendik:
> What are the next steps here?

Sorry, I was waiting for commits to be squashed. Do you need help with that?

pal03377:
Oh, I may be the one to be sorry. I didn't notice there were 2 commits again. 
I can try to do it myself and hope not to mess anything up.  ;-)

pal03377:
hmm, now there is an additional "merge" commit there...

pal03377:
Is it OK like this?

yurydelendik:
Thank you for the patch!

pal03377:
Thank you for helping me out!
Very cool. 

🎉

timvandermeij:

> How are translations usually added in open source projects?

You don't have to do anything for that. If you add the English strings, the translation teams from Mozilla will take care of translating the strings. There is a system that shows them that new English strings have been added and are waiting for a translation.

PR #9509, opened 2018-02-25
Implement a single `getInheritableProperty` utility function

timvandermeij:
This function combines the logic of two separate methods into one. The loop limit is also a good thing to have for the calls in `src/core/annotation.js`.

Moreover, since this is important functionality, a set of unit tests and documentation is added.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -15,7 +15,8 @@
 
 import {
   AnnotationBorderStyleType, AnnotationFieldFlag, AnnotationFlag,
-  AnnotationType, OPS, stringToBytes, stringToPDFString, Util, warn
+  AnnotationType, getInheritableProperty, OPS, stringToBytes, stringToPDFString,
+  Util, warn
 } from '../shared/util';
 import { Catalog, FileSpec, ObjectLoader } from './obj';
 import { Dict, isDict, isName, isRef, isStream } from './primitives';
@@ -60,7 +61,7 @@ class AnnotationFactory {
         return new TextAnnotation(parameters);
 
       case 'Widget':
-        let fieldType = Util.getInheritableProperty(dict, 'FT');
+        let fieldType = getInheritableProperty({ dict, key: 'FT', });
         fieldType = isName(fieldType) ? fieldType.name : null;
 
         switch (fieldType) {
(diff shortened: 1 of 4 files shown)


src/core/document.js
```diff
@@ -62,33 +62,16 @@ var Page = (function PageClosure() {
...
+     * @private
+     */
+    _getInheritableProperty(key, getArray = false) {
+      let value = getInheritableProperty(this.pageDict, key, getArray,
+                                         /* stopWhenFound = */ false);
+      if (!Array.isArray(value)) {
+        return value;
       }
-      return Dict.merge(this.xref, valueArray);
+      return isDict(value[0]) ? Dict.merge(this.xref, value) : value[0];
```
Snuffleupagus: Please note that `Dict.merge` is meaningless, and will only slow things down (albeit probably not much), when `value.length === 1`.

src/shared/util.js
```diff
@@ -644,6 +644,53 @@ function isEvalSupported() {
...
+ * @param {string} key - The key of the property to find the value for.
+ * @param {boolean} getArray - Whether or not the value should be fetched as an
+ *   array. The default value is `false`.
+ * @param {boolean} stopWhenFound - Whether or not to stop the traversal when
+ *   the key is found. If set to `false`, we always walk up the entire parent
+ *   chain, for example to be able to find `\Resources` placed on multiple
+ *   levels of the tree. The default value is `true`.
+ */
+function getInheritableProperty(dict, key, getArray = false,
+                                stopWhenFound = true) {
```
Snuffleupagus: Since there's cases where you might want to use `stopWhenFound` without `getArray`, wouldn't it perhaps be a bit nicer to have this helper function take an object instead!?

```javascript
function getInheritableProperty({ dict, key, getArray = false,
                                  stopWhenFound = true, }) {
```

timvandermeij: Good point! I'll try to make the changes during the weekend and also re-check your pull request for version 2.0.

Snuffleupagus: > also re-check your pull request for version 2.0.

Thanks; being this close I just want to make sure that it lands *before* anything else at this point, to reduce additional risk in the release.

src/shared/util.js
```diff
@@ -644,6 +644,53 @@ function isEvalSupported() {
...
+      }
+      values.push(value);
+    }
+    if (++loopCount > LOOP_LIMIT) {
+      warn(`getInheritableProperty: maximum loop count exceeded for "${key}"`);
+      break;
+    }
+    dict = dict.get('Parent');
+  }
+  return values ? values : undefined;
```
Snuffleupagus: Is there any reason why we cannot just initialize `let values = undefined` above (instead of using `null`), and then simply use `return values;` here?

src/core/document.js
timvandermeij: Done in the new commit.

src/shared/util.js
timvandermeij: Done in the new commit.

src/shared/util.js
timvandermeij: Done in the new commit. Note that I changed it to just `let values;` because otherwise ESLint would raise the following error: `It's not necessary to initialize 'values' to undefined (no-undef-init)`.

src/core/annotation.js
```diff
@@ -814,7 +820,7 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
     // inherit the options from a parent annotation (issue 8094).
     this.data.options = [];
 
-    let options = Util.getInheritableProperty(params.dict, 'Opt');
+    let options = getInheritableProperty({ dict, key: 'Opt', getArray: true, });
```
Snuffleupagus: `getArray` shouldn't be necessary here, since we're already iterating through *and* de-referencing the array below at https://github.com/mozilla/pdf.js/pull/9509/files#diff-18ca06e2bd5be4c9132b51f78388b8f0R827.

The only effect of using `getArray` here is that we're now having to iterate through the array *twice*, once in `Dict.getArray` and then once below.

timvandermeij: Good catch! I've removed that and updated the commit/pull request message.

Snuffleupagus (approved):
Looks good to me, thank you!

(I'm not sure if we should wait for the `2.0` release to *actually* occur, before landing this.)

PR #9520, opened 2018-02-28
Attempt to fix the `Array.prototype.includes` and `String.prototype.includes` polyfills (issue 9514, 9516)

### LINKED ISSUE #9514: Warning: FormatError: Required "loca" table is not found in Internet Explorer
http://edelman.amsterdam/wp-content/uploads/Terms_Conditions_NL.pdf

Configuration:
- Web browser and its version: Internet Explorer 11 version 11.248.16299.0
- Operating system and its version: Windows 10 Home version 1709 build 16299.248
- PDF.js version: 2.0.391
- Is a browser extension: no

1. Render the pdf in internet explorer

The pdf renders like it does in Chrome
Image of pdf in Chrome

The pdf doesn't render the font correctly and renders all sorts of random characters.
Image of pdf in Internet Explorer

If the information provided isn't enough I'm willing to create an example.

### LINKED ISSUE #9516: wrong polyfill with IE11 ?
schermata 2018-02-27 alle 15 11 26

Configuration:
- Web browser and its version: IE11
- PDF.js version: not sure, but probably >= 2.0.0
- Is a browser extension: no

1. open https://mozilla.github.io/pdf.js/web/viewer.html
2. try to use Array.prototype.includes()

[1,2].includes(1) should return true, not false

Snuffleupagus:
I don't understand why the previous way importing the polyfills didn't work, and I don't have time to try and figure it out, however this patch seems to fix things.

Fixes #9514.
Fixes #9516.

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -12,7 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint-disable no-extend-native, mozilla/use-includes-instead-of-indexOf */
+/* eslint-disable mozilla/use-includes-instead-of-indexOf */
 /* globals PDFJS */
 
 // Skip compatibility checks for the extensions and if we already ran
@@ -131,7 +131,7 @@ PDFJS.compatibilityChecked = true;
   if (String.prototype.includes) {
     return;
   }
-  String.prototype.includes = require('core-js/fn/string/includes');
+  require('core-js/fn/string/includes');
 })();
 
 // Provides support for Array.prototype.includes in legacy browsers.
@@ -140,7 +140,7 @@ PDFJS.compatibilityChecked = true;
   if (Array.prototype.includes) {
     return;
   }
-  Array.prototype.includes = require('core-js/fn/array/includes');
+  require('core-js/fn/array/includes');
 })();
 
 // Provides support for Math.log2 in legacy browsers.


src/shared/compatibility.js
```diff
@@ -131,7 +131,7 @@ PDFJS.compatibilityChecked = true;
   if (String.prototype.includes) {
     return;
   }
-  String.prototype.includes = require('core-js/fn/string/includes');
```
Rob--W: Just remove the `String.prototype.includes =` part and this should work.

When you import `core-js/fn/string/includes`, then `String.prototype.includes` is already overwritten, which can be seen using the following example:

```
$ node -e 'delete String.prototype.includes; console.log("".includes); require("core-js/fn/string/includes"); console.log("abc".includes("a"))'
undefined
true
```

The exported function from `require('core-js/fn/string/includes')` does not use `this`, but expects the first parameter to be the string, and the other parameters to be the parameters for the string method. I.e. `require('core-js/fn/string/includes')("abc", "b");`. With the current (wrong) code, it becomes require('core-js/fn/string/includes').call("abc", "b")`, which becomes (since `this` is ignored) `require('core-js/fn/string/includes')("b")`, i.e. equivalent to `"b".includes(undefined)`.

The documentation at https://github.com/zloirock/core-js#usage also shows that the intended usage for the default mode is to just call `require` and not use its return value.

Rob--W (changes requested):

src/shared/compatibility.js
Snuffleupagus: Thank you; I've made the suggested changes and everything seems to work just fine now!

Rob--W (approved):

Rob--W:
Squashed two bugs in one PR, good one!

PR #9541, opened 2018-03-08
[CRX] Disable fetch in Chrome 60-

Rob--W:
A user reported that they were not able to open many PDF files, and after some investigation I found that this is caused by the lack of credentials (cookies) in requests using `fetch`.

From the user's perspective, this is a regression in PDF.js (https://github.com/Rob--W/pdf.js/releases/tag/chrome-2.0.301).

See the commit message for more details + test case.

### DIFF
--- a/src/pdf.js
+++ b/src/pdf.js
@@ -52,8 +52,21 @@ if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
 } else if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('CHROME')) {
   let PDFNetworkStream = require('./display/network.js').PDFNetworkStream;
   let PDFFetchStream;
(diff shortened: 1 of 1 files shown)


timvandermeij:
The code itself looks good, but I'm only wondering if it should be here or if we can move it to the compatibility file (to keep browser-specific checks in core at a minimum). The latter would be better, but I'm not sure if it's doable given that this code runs early on when loading a PDF file and the compatibility file may not be loaded yet.

Rob--W:
`compatibility.js` contains polyfills. This patch does not provide a polyfill, but code that is only specific to the Chrome extension.

We could add more layers of indirection, but that increases the cost of maintenance (now the stream selection logic is in one place).

timvandermeij (approved):

timvandermeij:
Thank you for the patch!

PR #9542, opened 2018-03-08
[CRX] Fix error that broke the Chrome extension

Rob--W:
PR #9493 moved from `appConfig.defaultUrl` to `AppOptions.get('defaultUrl')`.
However, it forgot to replace `appConfig.defaultUrl` in chromecom.js, and as a result the extension is not able to open any PDF file.

That is quite a serious bug, so I would merge this before releasing 2.0.
/cc @Snuffleupagus

### DIFF
--- a/web/chromecom.js
+++ b/web/chromecom.js
@@ -15,6 +15,7 @@
 /* globals chrome */
 
 import { DefaultExternalServices, PDFViewerApplication } from './app';
+import { AppOptions } from './app_options';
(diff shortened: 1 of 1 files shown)


timvandermeij:
Looks good to me. Could you resolve the failure on Travis?

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

Issue #9510, opened 2018-02-25
PDF.js Installed in SeaMonkey But Does Not Work


DavidRoss:
Windows 7 Ultimate SP1 x64
Mozilla/5.0 (Windows NT 6.1; WOW64; rv:52.0) Gecko/20100101 SeaMonkey/2.49.2
pdf.js 2.0.391

See closed issue #9506 for pdf.js 2.0.389.  That issue also affects pdf.js 2.0.391.  

The install.rdf file contained within pdf.xpi 2.0.391 has the following:  
    `
    <em:targetApplication>
      <Description>
       <em:id>{92650c4d-4b8e-4d2a-b7eb-24ecf4f6b63a}</em:id>
       <em:minVersion>2.46</em:minVersion>
       <em:maxVersion>2.51a1</em:maxVersion>
     </Description>
    </em:targetApplication>`

This means pdf.js 2.0.391 can be installed and used with SeaMonkey 2.49.2, which is false.  It does indeed insall but cannot be used.  Either the stated range of SeaMonkey versions needs to be updated, or else that section of install.rdf needs to be removed.

dilworks:
@DavidRoss  could you upload your last working version of pdf.js.xpi? I've been just hit by this on SM2.49, and I neglected to backup my last working version, so I'm now unable to read PDFs on my browser :/

DavidRoss:
You can get pdf.js version 2.0.305 at
<http://www.rossde.com/test/pdf.js.xpi>.  This is the last version I
installed.

dilworks:
Oh, that one works just fine here, thanks for it!

Considering how slow has gotten the Seamonkey stable version release map, I wonder if pdf.js will regain compatibility with it soon...

timvandermeij:
Closing since the extension code is removed from our repository. It is maintained upstream in `mozilla-central` now, so file a bug in Bugzilla instead if Seamonkey support is still wished.

PR #9563, opened 2018-03-15
Remove SeaMonkey support from the install.rdf

IsaacSchemm:
Since pull request #9433 was merged, the pdf.js extension no longer supports SeaMonkey, so it should be removed from the install.rdf.


timvandermeij:
This is related to #9510, but it may become obsolete by #9566.

timvandermeij:
Closing since the other pull request will be merged instead, which also contains this change. Thank you.

PR #9564, opened 2018-03-15
Use 7-Zip on Windows if possible

IsaacSchemm:
The "zip" utility is not installed by default on Windows.

If the user is running Windows, this code will check the standard 7-Zip installation directory on the current drive. If 7z.exe is found, it will run that instead of zip.

Potential future issues:
* 7z and zip use different command line arguments
* 7-Zip might be installed in a different folder, or on a different drive

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -790,10 +790,13 @@ gulp.task('firefox', ['firefox-pre'], function (done) {
     env: { 'TZ': 'UTC', },
   };
 
-  exec('zip -r ' + FIREFOX_EXTENSION_NAME + ' ' +
+  var exe = process.platform === 'win32' && fs.existsSync('/Program Files/7-Zip/7z.exe')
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
> The "zip" utility is not installed by default on Windows.

The "correct" solution here, rather than adding a bunch of platform dependent (and program specific code) in `gulpfile.js`, would probably be to try and use the `gulp-zip` package for this gulp command too since it's already used elsewhere in this file.

IsaacSchemm:
@Snuffleupagus I would agree with that.

timvandermeij:
This may also become obselete by #9566.

yurydelendik:
> The "zip" utility is not installed by default on Windows.
> 7-Zip might be installed in a different folder

Zip is a dependency (similar to node) and required to perform task that is obsolete and is not performed on windows. If somebody decide to use it on window, they just need to install "zip", as they would install "7-zip", which is not a standard utility too. Closing as won't fix.

PR #9573, opened 2018-03-16
New XML parser

yurydelendik:
Improves performance of the RegExp based SimpleXMLParser.

The code for XMLParserBase copied from
https://github.com/mozilla/shumway/blob/16451d8836fa85f4b16eeda8b4bda2fa9e2b22b0/src/avm2/natives/xml.ts


src/display/xml_parser.js
```diff
@@ -0,0 +1,360 @@
...
+      }
+      switch (entity) {
+        case 'lt':
+          return '<';
+        case 'gt':
+          return '>';
+        case 'amp':
+          return '&';
+        case 'quot':
+          return '\"';
```
Snuffleupagus: The old code also handled `apos`, see https://github.com/mozilla/pdf.js/pull/9573/files#diff-3cb21ac6c8ce15118704e14df3dcc9b2L256; is that no longer necessary here?

yurydelendik: It's HTML thing, but we can add it there too

src/display/xml_parser.js
```diff
@@ -0,0 +1,362 @@
...
+};
+
+function isWhitespace(s, index) {
+  const ch = s[index];
+  return ch === ' ' || ch === '\n' || ch === '\r' || ch === '\t';
+}
+
+function isWhitespaceString(s) {
+  for (let i = 0; i < s.length; i++) {
+    const ch = s[i];
```
timvandermeij: Let's just call `isWhitespace(s[i])` here so we don't have to copy the logic. Also, let's cache `s.length` using `for (let i = 0, ii = s.length; i < ii; i++)` like we do elsewhere in the codebase.

src/display/metadata.js
```diff
@@ -23,13 +23,15 @@ class Metadata {
     // Ghostscript may produce invalid metadata, so try to repair that first.
     data = this._repair(data);
 
-    // Convert the string to a DOM `Document`.
+    // Convert the string to a XML document.
```
timvandermeij: Nit: `an XML document`

src/display/xml_parser.js
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    while (pos < s.length && !isWhitespace(s, pos) &&
+           s[pos] !== '>' && s[pos] !== '/') {
+      ++pos;
+    }
+    name = s.substring(start, pos);
+    skipWs();
+    while (pos < s.length && s[pos] !== '>' &&
+    s[pos] !== '/' && s[pos] !== '?') {
```
timvandermeij: This should to be aligned with `pos` for readability.

src/display/xml_parser.js
```diff
@@ -0,0 +1,362 @@
...
+                  this.onError(
+                    XMLParserErrorCode.UnterminatedDoctypeDeclaration);
+                  return;
+                }
+                complexDoctype = true;
+              }
+              const doctypeContent =
+                s.substring(j + 8, q + (complexDoctype ? 1 : 0));
+              this.onDoctype(doctypeContent);
+              // XXX pull entities ?
```
timvandermeij: What does this mean? Can we remove it?

src/display/xml_parser.js
```diff
@@ -0,0 +1,362 @@
...
+              return;
+            }
+            this.onBeginElement(content.name, content.attributes, isClosed);
+            j += content.parsed + (isClosed ? 2 : 1);
+            break;
+        }
+      } else {
+        do {
+          /* skipping text items */
+        } while (j++ < s.length && s[j] !== '<');
```
timvandermeij: Do we need a `do .. while` loop here instead of a regular `while` loop as used above?

src/display/xml_parser.js
```diff
@@ -0,0 +1,362 @@
...
+      return child.textContent;
+    }).join('');
+  }
+
+  hasChildNodes() {
+    return this.childNodes && this.childNodes.length > 0;
+  }
+}
+
+export class SimpleXMLParser extends XMLParserBase {
```
timvandermeij: Usually we put the exports at the bottom of the file. Can we do that here too for consistency (compare https://github.com/mozilla/pdf.js/pull/9573/files#diff-3cb21ac6c8ce15118704e14df3dcc9b2L405)?

src/display/xml_parser.js
```diff
@@ -0,0 +1,362 @@
...
+  constructor() {
+    super();
+    this._currentFragment = null;
+    this._stack = null;
+    this._errorCode = XMLParserErrorCode.NoError;
```
timvandermeij: `parseFromString` sets this too, so should we just set it too `null` here too like the other two members?

timvandermeij (commented):
In general this looks good to me (since it already worked in Shumway, I glanced over the details of the XML parsing), but there are some final comments I'd like to see addressed before approving this. Thanks!

src/display/xml_parser.js
yurydelendik: I just don't want to change shape of the object by keep _errorCode a number. I can add XMLParserErrorCode.Unknown value?

timvandermeij: I see. In that case, let's keep it like how it is now for simplicity. Thanks!

Rob--W:
Why is a custom XML parser used instead of the standard `DOMParser` API ( https://developer.mozilla.org/en-US/docs/Web/API/DOMParser )? All web browsers, even IE 9+ support this API.

Snuffleupagus:
> Why is a custom XML parser used instead of the standard DOMParser API ( https://developer.mozilla.org/en-US/docs/Web/API/DOMParser )? All web browsers, even IE 9+ support this API.

Please see issue #8903; and also https://bugzilla.mozilla.org/show_bug.cgi?id=1386676.

PR #9574, opened 2018-03-17
Update translations

timvandermeij:
This required changing the import script in two ways:

- we should use the `default` branch and not the `tip` tag since the latter may refer to another branch than `default` (this is the case for the `vi` locale, which caused in the files to be overwritten with incorrect contents since `tip` referred to the `THUNDERBIRD600b1_2018031614_RELBRANCH` branch);
- we should check if the response code is indeed 200 because recently a script removed all empty localization files upstream (refer to https://bugzilla.mozilla.org/show_bug.cgi?id=1443175).

*Smaller diff with https://github.com/mozilla/pdf.js/pull/9574/files?w=1.*


Snuffleupagus (approved):
Looks good to me, thank you!

PR #9576, opened 2018-03-17
Update packages

timvandermeij:
Jasmine had a major version bump and required a few minor changes in our booting code. Most notably, using `pending` in a `describe` block is no longer supported, so we can only return early there. On the positive side, the unit tests now run in a random order by default, which eliminates any dependencies between unit tests.

Note that upgrading to Webpack 4 is out of scope for this patch since the bots cannot work well with the newly generated bundles (both browsers on both bots do not react within 120 seconds). Webpack 4 is not faster for us than Webpack 3, so for now there is no need to upgrade.


test/unit/jasmine-boot.js
```diff
@@ -124,9 +124,9 @@ function initializePDFJS(callback) {
     },
   });
 
-  var catchingExceptions = queryString.getParam('catch');
-  env.catchExceptions(typeof catchingExceptions === 'undefined' ?
-                      true : catchingExceptions);
+  var stoppingOnSpecFailure = queryString.getParam('failFast');
+  env.stopOnSpecFailure(typeof stoppingOnSpecFailure === 'undefined' ?
+                        true : stoppingOnSpecFailure);
```
Snuffleupagus: From a cursory look at https://github.com/jasmine/jasmine/commit/e15f273f06c6b8c7902050fe500486c056802ed1, which necessitated this change, shouldn't the default now be `false` instead here?

test/font/jasmine-boot.js
```diff
@@ -76,9 +76,9 @@ function initializePDFJS(callback) {
     },
   });
 
-  var catchingExceptions = queryString.getParam('catch');
-  env.catchExceptions(typeof catchingExceptions === 'undefined' ?
-                      true : catchingExceptions);
+  var stoppingOnSpecFailure = queryString.getParam('failFast');
+  env.stopOnSpecFailure(typeof stoppingOnSpecFailure === 'undefined' ?
+                        true : stoppingOnSpecFailure);
```
Snuffleupagus: From a cursory look at https://github.com/jasmine/jasmine/commit/e15f273f06c6b8c7902050fe500486c056802ed1, which necessitated this change, shouldn't the default now be `false` instead here?

test/font/jasmine-boot.js
```diff
@@ -94,8 +94,9 @@ function initializePDFJS(callback) {
   // Reporters
   var htmlReporter = new jasmine.HtmlReporter({
     env,
-    onRaiseExceptionsClick() {
-      queryString.navigateWithNewParam('catch', !env.catchingExceptions());
+    onStopExecutionClick() {
+      queryString.navigateWithNewParam('failFast',
+                                       !env.stoppingOnSpecFailure());
```
Snuffleupagus: Re above, shouldn't this also be inverted?

test/unit/jasmine-boot.js
```diff
@@ -142,8 +142,9 @@ function initializePDFJS(callback) {
   // Reporters
   var htmlReporter = new jasmine.HtmlReporter({
     env,
-    onRaiseExceptionsClick() {
-      queryString.navigateWithNewParam('catch', !env.catchingExceptions());
+    onStopExecutionClick() {
+      queryString.navigateWithNewParam('failFast',
+                                       !env.stoppingOnSpecFailure());
```
Snuffleupagus: Re above, shouldn't this also be inverted?

test/unit/jasmine-boot.js
timvandermeij: I have fixed these comments in the new commit. Good catch!

Snuffleupagus:
> Most notably, using `pending` in a `describe` block is no longer supported, so we can only return early there.

In that case, I can't help wondering if the best solution wouldn't be to move those `pending` instances into each of the affected `it` blocks instead?
Obviously having to do that is a bit unfortunate, but I think it's conceptually the "right" solution since just `return`ing doesn't provide a good indication that tests were skipped.

test/unit/api_spec.js
```diff
@@ -1192,7 +1193,8 @@ describe('api', function() {
   });
   describe('Multiple PDFJS instances', function() {
```
Snuffleupagus: Also while you're here, let's update the name to `Multiple getDocument instances` (which I overlooked when removing the global `PDFJS` object).

package.json
```diff
@@ -2,41 +2,41 @@
...
+    "babel-loader": "^7.1.4",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
     "babel-preset-env": "^1.6.1",
     "core-js": "^2.5.3",
-    "escodegen": "^1.9.0",
-    "eslint": "^4.18.1",
-    "eslint-plugin-mozilla": "^0.9.0",
+    "escodegen": "^1.9.1",
+    "eslint": "^4.19.0",
+    "eslint-plugin-mozilla": "^0.9.1",
```
Snuffleupagus: There's apparently an even newer version, `0.10.0` of this package now.

Snuffleupagus (approved):
Looks good to me with the final comments, especially https://github.com/mozilla/pdf.js/pull/9576#issuecomment-374047346, addressed; thank you!

timvandermeij:
Firefox on Linux had an update, but there are no visual differences.

/botio-linux makeref

PR #9577, opened 2018-03-17
[api-minor] Fix various issues related to pageSize, and display the size for the active page in the document properties dialog

Snuffleupagus:
Glancing quickly at the recent implementation of pageSize, for the document properties dialog, there were a couple of things that seemed to benefit from some refactoring; please see the commit messages for details.

*Edit:* Can be tested with e.g. https://github.com/mozilla/pdf.js/blob/master/test/pdfs/sizes.pdf.

PR #9586, opened 2018-03-19
Ensure that `PDFPageProxy.pageSizeInches` handles non-default /Rotate entries correctly

Snuffleupagus:
*Follow-up to PRs #9577 and #9508.*

Without this patch, the pageSize will be incorrectly reported for some PDF files.
*Edit:* One such example is https://github.com/mozilla/pdf.js/blob/master/test/pdfs/rotation.pdf

### DIFF
--- a/web/pdf_document_properties.js
+++ b/web/pdf_document_properties.js
@@ -13,7 +13,9 @@
  * limitations under the License.
  */
 
-import { cloneObj, getPDFFileNameFromURL, NullL10n } from './ui_utils';
+import {
+  cloneObj, getPageSizeInches, getPDFFileNameFromURL, NullL10n
+} from './ui_utils';
 import { createPromiseCapability } from 'pdfjs-lib';
 
 const DEFAULT_FIELD_CONTENT = '-';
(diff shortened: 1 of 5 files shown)


yurydelendik:
I missed that pageSizeInches was added to src/display/api.js, can we move it to the ui_utils as a function? and tests to the ui_utils_spec?

yurydelendik:
something like this ^ , sorry

timvandermeij:
I'm wondering if it's only used in one place, namely the document properties file, shouldn't it just be placed there instead of in the general/shared `ui_utils` file?

Snuffleupagus:
We should probably land this before releasing version `2.0`. @yurydelendik do you want to address https://github.com/mozilla/pdf.js/pull/9586#issuecomment-374389858 so that we can land this PR?

Snuffleupagus:
@timvandermeij Sorry to bother you, but given added information in https://github.com/mozilla/pdf.js/pull/9586#issuecomment-374867882, would you be OK with landing this as-is?

web/pdf_document_properties.js
```diff
@@ -13,7 +13,12 @@
  * limitations under the License.
  */
 
-import { cloneObj, getPDFFileNameFromURL, NullL10n } from './ui_utils';
+import {
```
timvandermeij: Can we put these behind each other like we do in all other places (for example https://github.com/mozilla/pdf.js/pull/9586/files#diff-c7f5d75f7b6b0f13392f9d4c49a0c990L17)? It's just to keep a consistent style.

timvandermeij:
The change itself looks good to me. I'm fine with putting this in `ui_utils`, at least for now. Can you fix the comment above and perhaps squash the two commits into one so we can land it?

Snuffleupagus:
> Can you fix the comment above and perhaps squash the two commits into one so we can land it?

Both points have been taken care of; thanks!

timvandermeij (approved):

timvandermeij:
Good find!

PR #9587, opened 2018-03-19
Move the `cursorToolOnLoad` preference handling into `AppOptions` (PR 9493 follow-up)

Snuffleupagus:
- Move the `cursorToolOnLoad` preference handling into `AppOptions` (PR 9493 follow-up)

   Since no other viewer component is currently reading preferences itself, this patch thus unifies the behaviour across the viewer.

 - Remove the deprecated `mozL10n` and `localized` properties from `web/ui_utils.js`

   With PDF.js version `2.0`, these can now be removed. ***Edit:** It looks like we missed this previously when removing other deprecated code.*

/cc @timvandermeij

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -179,6 +179,9 @@ let PDFViewerApplication = {
       preferences.get('sidebarViewOnLoad').then(function resolved(value) {
         AppOptions.set('sidebarViewOnLoad', value);
       }),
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Good catch; thanks!

Issue #8489, opened 2017-06-06
examples/node/pdf2png example fails when using any other file other than the helloworld.pdf example file


nicholaswmin:
### Configuration:
- Node.js v7.6.0
- OSX Sierra
- pdf.js v1.8.188

### Steps to reproduce the problem:

1. Add attached `gal-trace.pdf` file in `examples/helloworld/` folder
2. Change `pdfURL` in `examples/node/pdf2png/pdf2png` to point to `examples/helloworld/gal-trace.pdf`
2. Run the `pdf2png.js` example

### What is the expected behavior? 

- It creates 14 images - one image for each page in `gal-trace.pdf`

### What went wrong? (add screenshot)

- No images are created
- Terminal logs the following error:

```
# PDF document loaded.
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): ReferenceError: document is not defined
(node:16114) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): ReferenceError: document is not defined
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): ReferenceError: document is not defined
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 4): ReferenceError: document is not defined
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 5): ReferenceError: document is not defined
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 6): ReferenceError: document is not defined
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 7): ReferenceError: document is not defined
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 8): ReferenceError: document is not defined
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 9): ReferenceError: document is not defined
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 10): ReferenceError: document is not defined
(node:16114) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 11): ReferenceError: document is not defined
[...]

PR #9588, opened 2018-03-20
Improve node.js support

swftvsn:
This change fixes "Unhandled rejection ReferenceError: HTMLElement is not defined" issue that is discussed in more detail in #8489.

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -2112,8 +2112,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       var heightScale = Math.max(Math.sqrt(c * c + d * d), 1);
 
       var imgToPaint, tmpCanvas;
-      // instanceof HTMLElement does not work in jsdom node.js module
-      if (imgData instanceof HTMLElement || !imgData.data) {
+      // typeof check is needed due to node.js support, see issue #8489
+      if ((typeof HTMLElement === 'function' &&
+           imgData instanceof HTMLElement) || !imgData.data) {
         imgToPaint = imgData;
       } else {
         tmpCanvas = this.cachedCanvases.getCanvas('inlineImage',


brendandahl (changes requested):
You have some lint failures.

swftvsn:
Second commit should fix that. On a side note, we all have wide screens even on laptops, so imho capping the line length to 80 chars only makes the code harder to read in most cases. But that's just a matter of personal preference of course, I use 200 on my projects and still have ample space on any display.

src/display/canvas.js
```diff
@@ -2112,8 +2112,10 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       var heightScale = Math.max(Math.sqrt(c * c + d * d), 1);
 
       var imgToPaint, tmpCanvas;
-      // instanceof HTMLElement does not work in jsdom node.js module
-      if (imgData instanceof HTMLElement || !imgData.data) {
+      // typeof check is needed due to node.js support, see issue #8489
+      if ((typeof HTMLElement === 'function' &&
+           imgData instanceof HTMLElement) ||
+           !imgData.data) {
```
Snuffleupagus: This line is indented with one space too many; let's re-format it like this instead:
```javascript
if ((typeof HTMLElement === 'function' &&
     imgData instanceof HTMLElement) || !imgData.data) {
```

Also, please squash the commits; refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

swftvsn:
@Snuffleupagus The if is now formatted per spec you gave, and the commits are squashed. Thanks for the excellent squash tutorial! (https://github.com/mozilla/pdf.js/wiki/Squashing-Commits)

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #9594, opened 2018-03-21
Add stricter validation in `Catalog.readPageLabels`

Snuffleupagus:
The current PageLabel dictionary validation code won't catch some (unlikely) forms of corruption. For example: a `Type`/`S` entry being `null`/`0`/empty string, a `P`/`St` entry being `null`/`0`.

Please note: I'm not aware of any bugs caused by the old code, but I've had this patch sitting locally for some time and figured it couldn't hurt to submit it.

**Edit:** When cleaning out my local PDF.js repository clone, I found this patch lying around.
In all likelihood, this will be my *last* PDF.js patch for the foreseeable future :-)

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -263,33 +263,45 @@ var Catalog = (function CatalogClosure() {
 
       for (var i = 0, ii = this.numPages; i < ii; i++) {
         if (i in nums) {
-          var labelDict = nums[i];
+          const labelDict = nums[i];
           if (!isDict(labelDict)) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this, and of course for all the other patches you submitted over the years! It's really exciting that we'll soon be ready for version 2.0 and it wouldn't be possible without your countless contributions, so know that we're very thankful for that!

PR #9565, opened 2018-03-15
Rename the globals to shorter names.

brendandahl:
pdfjsDistBuildPdf=pdfjsLib
pdfjsDistWebPdfViewer=pdfjsViewer
pdfjsDistBuildPdfWorker=pdfjsWorker

See #9545

PR #9611, opened 2018-03-29
Clean-up `getMainThreadWorkerMessageHandler` for non-PRODUCTION mode

Snuffleupagus:
*This is a final piece of clean-up of code that I recently wrote, after which I'm done :-)*

When the `getMainThreadWorkerMessageHandler` function was added, in PR #9385, it did so by basically introducing a `web/app.js` dependency in `src/display/api.js` through the `window.pdfjsNonProductionPdfWorker` property[1]. Even though this is limited to non-`PRODUCTION` mode, i.e. `gulp server`, it still seems unfortunate to have that sort of viewer dependency in the API code itself.

With the new, much nicer and shorter, names introduced in PR #9565 we can remove this non-`PRODUCTION` hack and just use `window.pdfjsWorker` in both the viewer and the API regardless of the build mode.

---

[1] It didn't seem correct to piggy-back on the `window.pdfjsDistBuildPdfWorker` property in non-`PRODUCTION` mode.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1317,13 +1317,7 @@ var PDFWorker = (function PDFWorkerClosure() {
     if (typeof window === 'undefined') {
       return null;
     }
-    if (typeof PDFJSDev === 'undefined' || !PDFJSDev.test('PRODUCTION')) {
-      return (window.pdfjsNonProductionPdfWorker &&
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

Issue #6685, opened 2015-11-24
Default viewer doesn't use 'retina' toolbar icons on iOS


jasongisstl:
When using the default viewer provided with PDF.js on iOS (retina), the toolbar icons appear blurry.

This seems to be because the following media query in _viewer.css_ isn't being matched:
`@media screen and (min-resolution: 2dppx)`

Changing this to include a webkit-specific prefix seems to solve this:
`@media screen and (min-resolution: 2dppx), (-webkit-min-device-pixel-ratio: 2)`
(My CSS knowledge isn't great, so apologies if there are problems with this)

I realise that users are expected to build upon the default viewer, but it would be nice if this worked out of the box.

(Tested using iOS 9.1 Safari & UIWebview)

yurydelendik:
Safari on desktop affected as well, the following shall be fine for our preprocessor:

```
@media
  screen and (-webkit-min-device-pixel-ratio: 2),
  screen and (min-resolution: 2dppx) {
```

prohtex:
I'm wondering if any progress has been made on this one. I am implementing pdfjs for clients who exclusively use tablets and mobile devices, and many comment that the icons are "blurry." On a desktop with a high-resolution display, the issue is not as critical as the icons are much smaller relative to the view area. But on a mobile device, it is odd.

![img_0829](https://user-images.githubusercontent.com/12946938/35244375-c340a3f0-ff74-11e7-8cbb-dff7ec4970be.PNG)

yurydelendik:
> I'm wondering if any progress has been made on this one.

Nope. Any ideas why mobile safari is doing it?

prohtex:
Isn't the issue that no one has created high-resolution assets to be served to the browser? I'm also wondering why the desktop viewer always displays a sans-serif font while iOS uses what looks like Times Roman!

yurydelendik:
> Isn't the issue that no one has created high-resolution assets to be served to the browser? 

Are these need to be in Safari specific format? WFM on Firefox and Chorme. Also text looks good as well.

prohtex:
I'm using Safari mobile on iOS devices. I don't have access to an android phone rn to check, but I assumed the assets simply weren't large enough to display properly.

timvandermeij:
Every image has a `@2x` variant for hi-dpi screens and this is served when the media query matches, so from our perspective this should be working (and has been tested on actual hi-dpi screens). Most likely Safari mobile on iOS is doing something strange that causes the media query not to match.

[...]

Issue #9392, opened 2018-01-22
Default font for UI elements on iOS


prohtex:
iOS devices display the toolbar elements (page count, zoom level) in Times New Roman rather than the sans-serif default seen on Desktop.

![img_0829](https://user-images.githubusercontent.com/12946938/35244911-a6d41024-ff76-11e7-902d-429b49cc609c.PNG)

shishirkulkarni:
Hello,
My name is Shishir. I am looking for projects where I can make some open source contributions. I was hoping to take up this issue. This issue also affects the current view option available on smaller screens. The fix that I am thinking of is adding a font-family declaration with value `Arial, Sans-Serif`. If this sounds alright and no one is looking at the issue, Can I take this up?

prohtex:
Hi @shishirkulkarni, that would be great!

Snuffleupagus:
Given that it works correctly in Mozilla Firefox (and also in e.g. Google Chrome), this unfortunately looks like a platform/browser specific issue rather than a PDF.js one; hence closing this as WONTFIX.

---

Also, please note http://mozilla.github.io/pdf.js/getting_started/#introduction (emphasis mine):
> The viewer is built on the display layer and is the UI for PDF viewer in Firefox and the other browser extensions within the project. It can be a good starting point for building your own viewer. ***However, we do ask if you plan to embed the viewer in your own site, that it not just be an unmodified version. Please re-skin it or build upon it.***

PR #9629, opened 2018-04-03
Implement autoprefixer

wojtekmaj:
Hey there!
I've implemented autoprefixer as a proof of concept. It was a breeze, actually!

Likely will help with: #6685

The only caveat worth mentioning is that if there were some browser-specific hacks like adding `!important` to `:-ms-something` only. See comments in the actual code for details.

**Action:**
I need to ask you for help regarding specific configuration on which browsers to target. For now, it's all on default settings and we don't configure anything.

I carefully compared the files using Mergely, and most of the differences make sense, but you **will** want to hardcode the configuration so that you'll be sure there are no changes without your knowledge. 

This config would be in form of `.browserslistrc` file. Sample file could look like this:

```
# Browsers that we support 
 
> 1%
Last 2 versions
IE 10
```

which means that we would like to support anything with more than 1%, market share, all last 2 versions of everything, no exceptions, and IE10.

Here's all the fun you can do: https://www.npmjs.com/package/browserslist

I checked *Allow edits from maintainers* and you can feel invited to use this option :)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,9 +18,11 @@
 
 'use strict';
 
+var autoprefixer = require('autoprefixer');
 var fancylog = require('fancy-log');
 var fs = require('fs');
 var gulp = require('gulp');
+var postcss = require('gulp-postcss');
 var rename = require('gulp-rename');
 var replace = require('gulp-replace');
 var transform = require('gulp-transform');
@@ -577,6 +579,7 @@ gulp.task('generic', ['buildnumber', 'locale'], function () {
     preprocessHTML('web/viewer.html', defines)
         .pipe(gulp.dest(GENERIC_DIR + 'web')),
     preprocessCSS('web/viewer.css', 'generic', defines, true)
+        .pipe(postcss([autoprefixer()]))
         .pipe(gulp.dest(GENERIC_DIR + 'web')),
 
     gulp.src('web/compressed.tracemonkey-pldi-09.pdf')
@@ -602,6 +605,7 @@ gulp.task('components', ['buildnumber'], function () {
     createComponentsBundle(defines).pipe(gulp.dest(COMPONENTS_DIR)),
     gulp.src(COMPONENTS_IMAGES).pipe(gulp.dest(COMPONENTS_DIR + 'images')),
     preprocessCSS('web/pdf_viewer.css', 'components', defines, true)
+        .pipe(postcss([autoprefixer()]))
         .pipe(gulp.dest(COMPONENTS_DIR)),
   ]);
 });
@@ -629,6 +633,7 @@ gulp.task('minified-pre', ['buildnumber', 'locale'], function () {
     preprocessHTML('web/viewer.html', defines)
         .pipe(gulp.dest(MINIFIED_DIR + 'web')),
     preprocessCSS('web/viewer.css', 'minified', defines, true)
+        .pipe(postcss([autoprefixer()]))
         .pipe(gulp.dest(MINIFIED_DIR + 'web')),
 
     gulp.src('web/compressed.tracemonkey-pldi-09.pdf')
@@ -719,6 +724,7 @@ gulp.task('mozcentral-pre', ['buildnumber', 'locale'], function () {
     preprocessHTML('web/viewer.html', defines)
         .pipe(gulp.dest(MOZCENTRAL_CONTENT_DIR + 'web')),
     preprocessCSS('web/viewer.css', 'mozcentral', defines, true)
+        .pipe(postcss([autoprefixer()]))
         .pipe(gulp.dest(MOZCENTRAL_CONTENT_DIR + 'web')),
 
     gulp.src(FIREFOX_EXTENSION_DIR + 'locale/en-US/*.properties')
@@ -766,6 +772,7 @@ gulp.task('chromium-pre', ['buildnumber', 'locale'], function () {
     preprocessHTML('web/viewer.html', defines)
         .pipe(gulp.dest(CHROME_BUILD_CONTENT_DIR + 'web')),
     preprocessCSS('web/viewer.css', 'chrome', defines, true)
+        .pipe(postcss([autoprefixer()]))
         .pipe(gulp.dest(CHROME_BUILD_CONTENT_DIR + 'web')),
 
     gulp.src('LICENSE').pipe(gulp.dest(CHROME_BUILD_DIR)),
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -71,6 +71,20 @@ var builder = require('./external/builder/builder.js');
 var CONFIG_FILE = 'pdfjs.config';
 var config = JSON.parse(fs.readFileSync(CONFIG_FILE).toString());
 
+// Default Autoprefixer config used for generic, components, minifed-pre
+var AUTOPREFIXER_CONFIG = {
+  browsers: [
+    'last 2 versions',
+    'Chrome >= 49', // Last supported on Windows XP
+    'Firefox >= 52', // Last supported on Windows XP
+    'Firefox ESR',
+    'IE >= 11',
+    'Safari >= 8',
+    '> 0.5%',
+    'not dead',
+  ],
+};
+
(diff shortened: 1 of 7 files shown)


web/viewer.css
```diff
@@ -53,36 +53,9 @@ select {
...
-  background-color: #000;
-  width: 100%;
-  height: 100%;
-  overflow: hidden;
-  cursor: none;
-  -moz-user-select: none;
-}
-
 #viewerContainer.pdfPresentationMode:-ms-fullscreen {
   top: 0px !important;
```
wojtekmaj: This is the line I was mentioning in the description. The thing is, now that we have separate rules:

1. :-ms-fullscreen
2. :fullscreen

they get expanded by autoprefixer to:

1. :-ms-fullscreen
2. :-webkit-full-screen
3. :-moz-full-screen
4. :-ms-fullscreen
5. :fullscreen

As you can see, 1. was untouched, but 2 got expanded into all rules for all the browsers. So now, -ms-specific overwrites are before all the others, so if any other !important would occur in generic code, we can get these rules overwritten. This is not the case, thankfully, but it's something we should be aware of for the future.

gulpfile.js
```diff
@@ -766,6 +772,7 @@ gulp.task('chromium-pre', ['buildnumber', 'locale'], function () {
     preprocessHTML('web/viewer.html', defines)
         .pipe(gulp.dest(CHROME_BUILD_CONTENT_DIR + 'web')),
     preprocessCSS('web/viewer.css', 'chrome', defines, true)
+        .pipe(postcss([autoprefixer()]))
```
Rob--W: This version only needs to be compatible with Chrome, so use:

```javascript
autoprefixer({ browsers: ['chrome >= 49'], })
```

(Chrome 49, because it is the last supported Chrome version by WIndows XP).

gulpfile.js
```diff
@@ -719,6 +724,7 @@ gulp.task('mozcentral-pre', ['buildnumber', 'locale'], function () {
     preprocessHTML('web/viewer.html', defines)
         .pipe(gulp.dest(MOZCENTRAL_CONTENT_DIR + 'web')),
     preprocessCSS('web/viewer.css', 'mozcentral', defines, true)
+        .pipe(postcss([autoprefixer()]))
```
Rob--W: This is only used in the latest version of Firefox, so use:

```javascript
autoprefixer({ browsers: ['last 1 firefox versions'], })
```

Rob--W (commented):
Can you show the diff between the CSS files generated before, and after this PR?

wojtekmaj:
Absolutely, here are the files:

### `gulp generic`
Before: [viewer.css](https://github.com/mozilla/pdf.js/files/1875767/viewer.txt)
After: [viewer.css](https://github.com/mozilla/pdf.js/files/1875770/viewer.txt)

### `gulp mozcentral-pre`
Before: [viewer.css](https://github.com/mozilla/pdf.js/files/1875789/viewer.txt)
After: [viewer.css](https://github.com/mozilla/pdf.js/files/1875775/viewer.txt)

### `gulp chromium-pre`
Before: [viewer.txt](https://github.com/mozilla/pdf.js/files/1875786/viewer.txt)
After: [viewer.css](https://github.com/mozilla/pdf.js/files/1875782/viewer.txt)

Rob--W:
Generated viewer.css for chromium and mozcentral looks good to me.
I haven't checked the `generic` build yet.

The browser support table is at https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support . Can you add a browserlist so that the autoprefixer generates the expected prefixes?

wojtekmaj:
We went for config passed directly to `autoprefixer()` function approach, so I'd follow that approach and do it again for `gulp generic` for consistency and to avoid any side effects.

This table is a little too general for me to build a config. Here's my best guess. How does that sound?

```
last 4 versions
IE 11
Safari 8
iOS 8
```

Rob--W:
`Chrome 49` is OK because it is only listed because of Windows XP support (if XP support is desired, perhaps it's worth including "Firefox 52" too - this is currently Firefox ESR, and the last to support Win XP: https://support.mozilla.org/en-US/kb/end-support-windows-xp-and-vista ).
`IE 11` is OK because it is the last IE version.
`Safari >= 8` should be used to account for newer versions of Safari; similarly for ios.

Firefox ESR is missing from the list. Perhaps you can use `defaults` (see https://github.com/browserslist/browserslist#queries) and extend that with the explicitly supported browser versions.

wojtekmaj:
`Safari 8` <==> `Safari >= 8` unless Safari would regress in development <sub><sup>(I wouldn't be surprised 🙄)</sup></sub>
iOS is covered by `last 2 versions` and in my opinion taking their [fabulous update distribution speed](https://developer.apple.com/support/app-store/) I wouldn't bother much more, but up to you.
Regarding Windows XP support, I added Chrome 49 only because of `chromium-pre` config, maybe that was a mistake.

So to sum up, we could go for:

Option 1: Old iOS support (>99,5% market share), XP support
```
last 2 versions
Firefox 52 # last supported on Windows XP
Chrome 49 # last supported on Windows XP
IE 11
iOS last 3 versions
Safari 8 # desktop Safari
```

Option 2: iOS default (>95% market share), XP support
```
last 2 versions
Firefox 52 # last supported on Windows XP
Chrome 49 # last supported on Windows XP
IE 11
Safari 8 # desktop Safari
```

Option 3: No XP support, iOS default
```
last 2 versions
Firefox ESR
IE 11
Safari 8 # desktop Safari
```

Rob--W:
> `Safari 8` <==> `Safari >= 8` unless Safari would regress in development <sub><sup>(I wouldn't be surprised 🙄)</sup></sub>

The intention is to support Safari 8 and later, so `Safari >= 8` is more accurate. I haven't checked whether there would be any difference (I would not be surprised either way), but it is possible to have a new prefix in a later version. A historical example of prefix madness is Flexbox.

How about this:

```
last 2 versions
Chrome 49 # Last supported on Windows XP
Firefox 52 # Last supported on Windows XP
Firefox ESR
IE 11
Safari >= 8
> 0.5%
not dead
```

(based on the `defaults` as documented at https://github.com/browserslist/browserslist#queries)
(92.16% coverage according to http://browserl.ist/?q=last+2+versions%2C+Chrome+49%2C+Firefox+52%2C+Firefox+ESR%2C+IE+11%2C+Safari+%3E%3D+8%2C+%3E+0.5%25%2C+not+dead)

How did you determine the browser support? I'm checking browserl.ist and the results are difference. E.g. option 1 has 76.64% coverage according to http://browserl.ist/?q=last%202%20versions,Firefox%2052,Chrome%2049,IE%2011,last%203%20iOS%20versions,Safari%208 (note: I fixed the syntax by changing "iOS last 3 versions" to "last 3 iOS versions").

wojtekmaj:
Browser support was only applying to iOS versions, I thought I made it clear, sorry if it wasn't. This config looks great to me!

timvandermeij:
Nice work! The integration indeed looks easier than I would have expected. I'll review this in the coming days (mainly checking the various diffs), but quickly looking at the diffs tells me that this is a good idea; it cleans up the CSS quite a bit and it's much less error-prone since I'm seeing some (un)prefixed rules that we currently just miss.

wojtekmaj:
I'm adding what @Rob--W suggested (no reason not to, it doesn't cost us anything to support some more prefixes), but in a form of `autoprefixer()` config in gulpfile to have it consistent, if you don't mind. Another reason is that .browserlistrc can be read by several other libraries, too, and that could introduce some side effects I can't predict.

timvandermeij:
Yes, that sound fine to me.

timvandermeij (approved):

timvandermeij:
Thank you for implementing this! It cleans up the CSS files quite a bit.

wojtekmaj:
Super glad I could help. ❤️ to Mozilla!

prohtex:
Any chance we could address #9392 as well?

PR #9660, opened 2018-04-12
Support running the tests headlessly.

brendandahl:
Now you can add "headless: true" to the browser_manifest.json. I can't get headless to work on Chrome with MacOS though.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -140,7 +140,9 @@ function examineRefImages() {
   startServer();
   var startUrl = 'http://' + server.host + ':' + server.port +
                  '/test/resources/reftest-analyzer.html#web=/test/eq.log';
-  var browser = WebBrowser.create(sessions[0].config);
+  var config = Object.assign({}, sessions[0].config);
+  config['headless'] = false;
+  var browser = WebBrowser.create(config);
   browser.start(startUrl);
 }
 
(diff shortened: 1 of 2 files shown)


timvandermeij:
Nice work! I was thinking about this some time ago because I hoped that we could eventually run the reference tests on Travis CI too, and not requiring a display would make that a bit easier.

Rob--W:
@brendandahl Can you provide log files (maybe via IRC?)

brendandahl:
I was getting:
```
ERROR:gpu_process_transport_factory.cc(1007)] Lost UI shared context.
```
I ended up finding https://github.com/testem/testem/issues/1117#issuecomment-356534997 and it now seems to work locally.

timvandermeij:
@brendandahl Could you fix up this PR so we can merge it?

I'm thinking that we could use this to perhaps run more tests on Travis CI instead of on the bots. The only problem is that we need a place to store reference images and PDFs, but according to Travis CI documentation a simple Amazon S3 bucket can be used for persistent storage.

brendandahl:
Note: with chrome I'm currently running into https://bugs.chromium.org/p/chromium/issues/detail?id=848210 on MacOS.

timvandermeij (approved):

timvandermeij:
I tried this locally and it worked just fine. Thank you for implementing this!

Issue #9663, opened 2018-04-13
once the search bar is closed it should be clear the next time it's opened.


shradha-khapra:
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version:
- Operating system and its version:
- PDF.js version:1.4
- Is a browser extension:

Steps to reproduce the problem:
1. search for any text in the pdf-viewer
2.close the search bar and re-open it

What is the expected behavior? (add screenshot)

What went wrong? (add screenshot)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

shradha-khapra:
Hey @timvandermeij , I would like to work on this issue . Can this issue be assigned to me?

shradha-khapra:
@timvandermeij  can you please tell me the location in the repository where I can find the JS code written for the search button? Because I was unable to find it in the viewer.js file.

shradha-khapra:
Never mind. I found the JS file.

Snuffleupagus:
> once the search bar is closed it should be clear the next time it's opened.

This is *not* how the native findbar works in Firefox, so why should the PDF.js (built-in) findbar behave any different here!? Having different behaviours for the native/built-in findbars seems very confusing, and will result in a bad UX; hence I'd suggest that WONTFIX is an appropriate resolution here.

Snuffleupagus:
> once the search bar is closed it should be clear the next time it's opened.

This is *not* how the native findbar works in Firefox, so why should the PDF.js (built-in) findbar behave any different here!? Having different behaviours for the native/built-in findbars seems very confusing, and will result in a bad UX; hence I'd suggest that WONTFIX is an appropriate resolution here.

timvandermeij:
Good point. Now that I'm looking at this again and having tried both the native find bar and the PDF.js find bar, it indeed wouldn't make sense for them to differ in behavior, so let's close this.

PR #9668, opened 2018-04-16
input-space of search button fixed

shradha-khapra:


shradha-khapra:
This pull request is for issue #9663 .Please respond if I made some mistake.

web/pdf_find_bar.js
```diff
@@ -189,6 +189,7 @@ class PDFFindBar {
 
   toggle() {
     if (this.opened) {
+      PDFFindController.reset();  // To reset the Input Bar when Search Button closed
```
timvandermeij: Let's handle this in the `close()` method instead since it should also happen if the dialog is closed in other ways than via `toggle()`.

shradha-khapra: okay. I'll try that.

timvandermeij:
Closing since we decided not to fix #9663.

Issue #8682, opened 2017-07-21
view pdf as svg  GUI is not same as PDF file


hurrry:
Link to PDF file (or attach  here):
http://techexcel.devsuite.net/appweb/k6.pdf
Configuration:
- Web browser and its version:
- Operating system and its version:
- PDF.js version:
- Is an extension:

Steps to reproduce the problem:
1.  view this pdf file as svg is not good!
2. please see two screenshot in below.

What is the expected behavior? (add screenshot)
[http://techexcel.devsuite.net/appweb/2.png](url)
What went wrong? (add screenshot)
[http://techexcel.devsuite.net/appweb/1.png](url)
![Uploading 1.png…]()
![Uploading 2.png…]()

Snuffleupagus:
> Link to PDF file (or attach here):
> http://techexcel.devsuite.net/appweb/k6.pdf

The PDF file is no longer available, and additionally the `Configuration` section is completely empty; closing as INCOMPLETE for now?

timvandermeij:
The PDF file is available in the Web Archive at https://web.archive.org/web/20180403195550/https://techexcel.devsuite.net/appweb/k6.pdf.

timvandermeij:
This is mostly fixed now. Only blend modes are missing and the calendar icon at the top of each page misses a line.

Snuffleupagus:
Given that blend modes support for the SVG back-end is already being tracked in issue #10935, we can probably close this issue as a duplicate of that one (also given the *unsupported* status of the SVG back-end itself).

Duplicate of #10935

PR #9673, opened 2018-04-17
SVG clip intersections and operators

### LINKED ISSUE #8496: SVG example generates svg holding lots of small black boxes -> the DOM looks weird.
[error_prone_file.pdf](https://github.com/mozilla/pdf.js/files/1057922/error_prone_file.pdf)
It also happens in other pages of the pdf file, but I have only provided one page to reproduce the problem.

Configuration:
- Chrome - 58.0.3029.110 (64-bit)
- macOS Sierra - 10.12.5
- PDF.js version: "betaVersion": "1.8.188"

- Just try the added pdf file with pdf.js's own provided svgviewer example

https://www.dropbox.com/s/c2gknbtlo6vchua/Screenshot%202017-06-07%2014.34.57.png?dl=0

What went wrong?
https://www.dropbox.com/s/082uf4z73v19zrs/Screenshot%202017-06-07%2014.35.26.png?dl=0

--
I the earlier versions of pdf.js this problem was occurring in many of my pdf's -> many have been fixed over time, but it still seems to occur sometimes.

--
kaarebs

janpe2:
Fixes #8496

When a new clipping path is set, the result should be the *intersection* of the new path with the previous clipping path. Currently the SVG backend incorrectly just *replaces* the previous clipping path with the new one.

SVG can automatically compute the intersection. It's easiest to add a `clip-path` attribute to a `clipPath` element. The [SVG Recommendation](https://www.w3.org/TR/SVG11/masking.html#EstablishingANewClippingPath) says:
> If a valid 'clip-path' reference is placed on a 'clipPath' element, the resulting clipping path is the intersection of the contents of the 'clipPath' element with the referenced clipping path.

In issue #8496 a text-shaped clipping path is first set and then a larger rectangular path. This is repeated in a number of texts on the page.
[issue8496_reduced.pdf](https://github.com/mozilla/pdf.js/files/1920753/issue8496_reduced.pdf)
The reduced PDF should show a black letter 'a', not a rectangle.

I have also fixed all the painting operators `S s f f* B B* b b* n` that should generate clipping paths. Previously only `n` has worked. Here is a file to test these:
[clip_operators.pdf](https://github.com/mozilla/pdf.js/files/1920754/clip_operators.pdf)

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -995,6 +995,8 @@ SVGGraphics = (function SVGGraphicsClosure() {
         this.extraStack.forEach(function (prev) {
           prev.clipGroup = null;
         });
+        // Intersect with the previous clipping path.
+        clipPath.setAttributeNS(null, 'clip-path', current.activeClipUrl);
       }
       current.activeClipUrl = 'url(#' + clipId + ')';
 
@@ -1069,6 +1071,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
       if (current.element) {
         current.element.setAttributeNS(null, 'fill', current.fillColor);
         current.element.setAttributeNS(null, 'fill-opacity', current.fillAlpha);
+        this.endPath();
       }
     },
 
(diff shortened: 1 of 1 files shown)


timvandermeij:
Moreover, this fixes the majority of #8682 (only some text/images are missing, but all clipping is now correct).

timvandermeij (approved):

timvandermeij:
Good find; thank you for improving this!

PR #9681, opened 2018-04-20
Pass imageLayer to PDFPageView

websirnik:
This pull request allows passing a custom `imageLayer` when creating a new PDFPageView.

```
var pdfPageView = new pdfjsViewer.PDFPageView({
    container: container,
    id: PAGE_TO_VIEW,
    scale: SCALE,
    defaultViewport: pdfPage.getViewport(SCALE),
    imageLayer: {
        beginLayout: function (){},
        endLayout: function (){},
        appendImage: function(img){}
    }
});
```

### DIFF
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -87,6 +87,7 @@ class PDFPageView {
     this.renderingQueue = options.renderingQueue;
     this.textLayerFactory = options.textLayerFactory;
     this.annotationLayerFactory = options.annotationLayerFactory;
+    this.imageLayer = options.imageLayer;
     this.renderer = options.renderer || RendererType.CANVAS;
     this.enableWebGL = options.enableWebGL || false;
     this.l10n = options.l10n || NullL10n;
@@ -576,6 +577,7 @@ class PDFPageView {
       viewport: this.viewport,
       enableWebGL: this.enableWebGL,
       renderInteractiveForms: this.renderInteractiveForms,
+      imageLayer: this.imageLayer
     };
     let renderTask = this.pdfPage.render(renderContext);
     renderTask.onContinue = function (cont) {


timvandermeij:
Could you elaborate a bit on the use case for this? I'm not really seeing why we should have this as a parameter in the library (it sounds like something that may be useful for a custom version, but not really for the library itself).

web/pdf_page_view.js
```diff
@@ -576,6 +577,7 @@ class PDFPageView {
       viewport: this.viewport,
       enableWebGL: this.enableWebGL,
       renderInteractiveForms: this.renderInteractiveForms,
+      imageLayer: this.imageLayer
```
timvandermeij: Missing trailing comma.

web/pdf_page_view.js
```diff
@@ -87,6 +87,7 @@ class PDFPageView {
     this.renderingQueue = options.renderingQueue;
     this.textLayerFactory = options.textLayerFactory;
     this.annotationLayerFactory = options.annotationLayerFactory;
+    this.imageLayer = options.imageLayer;
```
timvandermeij: The documentation for `PDFPageViewOptions` above needs to be updated too.

Snuffleupagus:
> Could you elaborate a bit on the use case for this? I'm not really seeing why we should have this as a parameter in the library (it sounds like something that may be useful for a custom version, but not really for the library itself).

Isn't another problem also that the `imageLayer` functionality is both completely unused *and* untested as-is. Before surfacing this functionality more prominently, shouldn't e.g. unit-tests be added *first* such that the relevant code-paths are actually exercised!?

(Also, if a new option is added in `PDFPageView`, it really should be added to `BaseViewer` as well.)

websirnik:
imageLayer implementation was partly already in the codebase but was never finished. The use case for me was to collect images on a page via `appendImage(img)` and create a custom image zoom-in experience after.

Since the is no API call to get the images on a page, imageLayer feels like a good solution. I can add documentation & tests. Should I proceed?

timvandermeij:
Thank you for sharing the use case. I'm personally fine with this as long as there are (unit) tests for it so that we can make sure that it doesn't break in a future update.

Issue #9675, opened 2018-04-18
Error: Requesting object that isn't resolved yet on PDF even with disableFontFace


RafaPolit:
Failing PDF:
[15240917076692yt54e2iickoemgikbw486w29.pdf](https://github.com/mozilla/pdf.js/files/1926035/15240917076692yt54e2iickoemgikbw486w29.pdf)

Configuration:
- Version: **node v6.9.5**
- Operating system and its version: **Ubuntu 16.04LTS**

- PDF.js version: **2.0.0**

Steps to reproduce the problem:
1. Run the demo code found in https://github.com/mozilla/pdf.js/blob/master/examples/node/pdf2png/pdf2png.js
2. Console outputs:
```
# PDF document loaded.
(node:1211) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Requesting object that isn't resolved yet Verdana_path_C
```

What is the expected behavior? 

**A successful conversion**

What went wrong?

```
# PDF document loaded.
(node:1211) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Requesting object that isn't resolved yet Verdana_path_C
```
This is happening even though the disableFontFace: true flag has been set, which seems wrong.

Thanks for any insight.  Best regards,
Rafa.

RafaPolit:
I believe that the error can be caught (instead of an UnhandledPromiseRejection) replacing the **_scheduleNext** function in https://github.com/mozilla/pdf.js/blob/master/src/display/api.js#L2452 with a catch.

What is causing the error is less clear at this point.

```
  _scheduleNext: function InternalRenderTask__scheduleNext() {
      if (this.useRequestAnimationFrame && typeof window !== 'undefined') {
        window.requestAnimationFrame(this._nextBound);
      } else {
        Promise.resolve(undefined).then(this._nextBound).catch(err => { this.callback(err); } );
      }
    }
```

timvandermeij:
Not yet. We've been busy with other pull requests first, but I do agree that the error handling should at least be improved here. The cause of the underlying issue is not yet clear.

Snuffleupagus:
> The cause of the underlying issue is not yet clear.

This is simply (another) duplicate of issue #4244, and should probably be closed as such.

(Also, it seems to me if a work-around is implemented, it really needs to apply *generally* and not only for a part of the problem. Hence the proposed PR does *not* strike me as a good solution.)

[...]

PR #9689, opened 2018-04-24
Fixed critical unhandled promise that prevented error catching using API

RafaPolit:
The affected line causes an unhandled rejection if any code inside _nextBound throws an error.

This makes using the API impossible, as there is no way of catching a failing PDF.

For this issue, the offending PDF and the underlying problem throwing the error, please refer to: #9675 

This PR fixes the unhandled rejection error only.  This is not a fix for the error that is being thrown.  It's just an indispensable catch to the promise so that the error can bubble up to where it should in order for the error to be catched.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2453,7 +2453,7 @@ var InternalRenderTask = (function InternalRenderTaskClosure() {
       if (this.useRequestAnimationFrame && typeof window !== 'undefined') {
         window.requestAnimationFrame(this._nextBound);
       } else {
-        Promise.resolve(undefined).then(this._nextBound);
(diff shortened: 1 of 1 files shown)


RafaPolit:
If this PR has no conflicts, even after all the changes of late, please consider merging it.  It's just one line and really allows for error handling in a lot of situations where we are just stuck with no way of catching errors.

We would really hope to use the main package instead of running temporary forks, so if you would venture a timeframe of implementation, that would be great.

Thanks for all the hard work! Best regards,
Rafa.

brendandahl:
Thanks for the patch!

PR #9699, opened 2018-05-02
Update README.md

cloojure:
Global install requires root privilege.


Rob--W:
No, please do not. It is a terrible idea to run something like `npm` as root.

You can set `prefix=PUT YOUR CUSTOM USER-WRITABLE PATH HERE` in `~/.npmrc` to change the installation location of `npm -g`. For documentation, see https://docs.npmjs.com/files/npmrc

PR #9701, opened 2018-05-03
Add abort functionality in fetch stream

### LINKED ISSUE #9698: PDF.js should abort fetch requests during teardown
1. Fetch a document with streaming enabled.
2. Destroy the document loading task
3. Observe that the fetch request is not aborted.

Fetch request should be aborted when cancel is called on `PDFFetchStreamReader` and `PDFFetchStreamRangeReader`.

**Context**
It seems when fetch was implemented (https://github.com/mozilla/pdf.js/issues/6126#issuecomment-130572300) `abort` on fetch was still in flux. 

It seems now that has settled and chromium has also implemented aborting fetch requests in 66 (https://bugs.chromium.org/p/chromium/issues/detail?id=750599)

mukulmishra18:
Fixes #9698

### DIFF
--- a/src/display/fetch_stream.js
+++ b/src/display/fetch_stream.js
@@ -21,10 +21,11 @@ import {
   validateRangeRequestCapabilities, validateResponseStatus
 } from './network_utils';
 
-function createFetchOptions(headers, withCredentials) {
+function createFetchOptions(headers, withCredentials, abortController) {
   return {
     method: 'GET',
     headers,
+    signal: abortController && abortController.signal,
     mode: 'cors',
     credentials: withCredentials ? 'include' : 'same-origin',
     redirect: 'follow',
@@ -80,6 +81,9 @@ class PDFFetchStreamReader {
       this._disableRange = true;
     }
 
+    if (typeof AbortController !== 'undefined') {
+      this._abortController = new AbortController();
+    }
     this._isStreamingSupported = !source.disableStream;
     this._isRangeSupported = !source.disableRange;
 
@@ -93,8 +97,8 @@ class PDFFetchStreamReader {
     }
 
     let url = source.url;
-    fetch(url, createFetchOptions(this._headers, this._withCredentials)).
-        then((response) => {
+    fetch(url, createFetchOptions(this._headers, this._withCredentials,
+        this._abortController)).then((response) => {
       if (!validateResponseStatus(response.status)) {
         throw createResponseStatusError(response.status, url);
       }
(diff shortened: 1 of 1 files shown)


mukulmishra18:
I can see node `Aborted (core dump)` error with an error message:

`../src/node_contextify.cc:631:static void node::contextify::ContextifyScript::New(const v8::FunctionCallbackInfo<v8::Value>&): Assertion 'args[1]->IsString()' failed.` 

I think there is some issue with gulp and node v10.0.0. It is something related to internal contextify script of nodeJS. See https://github.com/nodejs/node/issues/20281 node issue for more details.

mukulmishra18:
@Rob--W Can you please review this PR?

src/display/fetch_stream.js
```diff
@@ -80,6 +81,10 @@ class PDFFetchStreamReader {
       this._disableRange = true;
     }
 
+    if (window.AbortController) {
```
Snuffleupagus: Drive-by nit: It's probably best to *not* reference `window` in `/src` code unless actually necessary, so `typeof AbortController !== 'undefined'` or `typeof AbortController === 'function'` would seem like a much better way of feature-testing here.

src/display/fetch_stream.js
```diff
@@ -184,6 +193,11 @@ class PDFFetchStreamRangeReader {
     this._readCapability = createPromiseCapability();
     this._isStreamingSupported = !source.disableStream;
 
+    if (window.AbortController) {
```
Snuffleupagus: Drive-by nit: See https://github.com/mozilla/pdf.js/pull/9701/files#r189454633.

src/display/fetch_stream.js
mukulmishra18: Seems reasonable. Done in new commit. Thanks.

src/display/fetch_stream.js
mukulmishra18: Done.

src/display/fetch_stream.js
```diff
@@ -80,6 +81,10 @@ class PDFFetchStreamReader {
       this._disableRange = true;
     }
 
+    if (window.AbortController) {
+      this._abortController = new AbortController();
+      this._abortSignal = this._abortController.signal;
```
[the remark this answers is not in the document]
mukulmishra18: Passing `abortSignal` directly to `createFetchOptions` seems a good idea. Done in commit. Thank You.

src/display/fetch_stream.js
```diff
@@ -171,6 +176,10 @@ class PDFFetchStreamReader {
     if (this._reader) {
       this._reader.cancel(reason);
     }
+    // Abort fetch if it is abortable.
```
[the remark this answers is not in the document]
mukulmishra18: Removed in new commit. Thanks.

Rob--W (approved):
LGTM, thanks for the patch.

PR #9566, opened 2018-03-15
Remove the Firefox extension building code.

brendandahl:
Firefox no longer supports this legacy extension and it is a pain to
sync changes from here and mozilla central.

There's still a bit more cleanup we could do for locale's, but I'd like to leave that for a follow up bug.

PR #9726, opened 2018-05-15
Add SeaMonkey extension

IsaacSchemm:
I've added a SeaMonkey extension for pdf.js. The extension is essentially identical to the last version that was compatible with Firefox 52, frozen in time, but with the newest version of pdf.js included.

The extension works with SeaMonkey 2.49 through 2.53. It should work with the "current" version of SeaMonkey throughout 2018, since SeaMonkey 2.54-2.56 will be skipped and SeaMonkey 2.57 (based on Firefox 60 ESR) isn't expected to be ready before the end of the year.

I understand if this is too narrow of a case to merit being merged into the main code repository, but I thought I should at least let you know I was working on it. The only changes to files outside of the extensions/seamonkey folder are in gulp.js (to build the SeaMonkey extension) and firefoxcom.js (to fix a bug when using the find tool.)


Snuffleupagus:
*I'm no longer active in the PDF.js project, but what this PR suggest strikes me as the wrong thing to do.*

Please note the Firefox addon code was purposely removed from this repository, in PR #9566, just so that it wouldn't be necessary to keep maintaining it both here and in `mozilla-central`.
Adding that code back would thus force the PDF.js maintainers to, once again, worry about keeping it up to date with changes made in `mozilla-central`. (More on that below, please read on.)

> The extension is essentially identical to the last version that was compatible with Firefox 52, frozen in time,

Based on the quoted text, particularly the "frozen in time" bit, I'm assuming the intent when submitting this PR was that "new" Seamonkey addon code wouldn't need to be maintained.
If that was the intention, then that poses a *huge* problem if/when security vulnerabilities are discovered. That isn't just a theoretical concern either, since security vulnerabilities were found (and fixed) not long ago[1].

If this (outdated) addon code is re-added to the PDF.js Github repository, it would thus force the maintainers to apply any security patches to this code as well. This would thus add an unnecessary step, which the move of the Firefox specific code into `mozilla-central` sought to avoid, which just seems like a waste of developer time/effort to me at this point.


> I understand if this is too narrow of a case to merit being merged into the main code repository, but I thought I should at least let you know I was working on it.

It definitely seems much too narrow of a use-case, since this PR proposes adding a few thousand lines of unused[2] and untested code. Not to mention that it may also contain known security vulnerabilities (which actually happens to be the case).
Hence it seems, to me, that this Seamonkey addon is a better fit for a repository forked from the official one. (A separate repository also means less risk of users perceiving the Seamonkey addon as officially supported.)

> [...] and firefoxcom.js (to fix a bug when using the find tool.)

FYI, that shouldn't be necessary and it suggests that `supportsIntegratedFind` (in PdfStreamConverter.jsm) isn't working as expected in Seamonkey.

---
[1] Hopefully I'm not divulging any sensitive details by linking to https://www.mozilla.org/en-US/security/advisories/mfsa2018-11/#CVE-2018-5157, but that website is publicly available so it ought to be OK.

[2] At least from the perspective of the general PDF.js project.

IsaacSchemm:
> FYI, that shouldn't be necessary and it suggests that supportsIntegratedFind (in PdfStreamConverter.jsm) isn't working as expected in Seamonkey.

As far as I can tell, supportsIntegratedFind is working just fine - the problem is that SeaMonkey _itself_ doesn't support integrated find.

PR #9729, opened 2018-05-16
Add a `gulp image_decoders` command to package the image decoders (i.e. jpg.js, jpx.js, jbig2.js) separately, and publish them in pdfjs-dist

### LINKED ISSUE #8966: Separating JPG decoder
Hi, I am looking for a good JPG decoder in Javascript.

There are several projects, that "cut out" the JPEG decoder from PDF.js and publish it on GitHub as a separate repository. E.g. this one: https://github.com/notmasteryet/jpgjs .

Could you separate the JPEG decoder from pdf.js into a separate repository? PDF.js would then use it as an external "module".  The maintenance would be easier (smaller amount of code in repositories, issues would be split into multiple repositories etc.).

### LINKED ISSUE #9388: Determine if the `parseJbig2` function should be reimplemented
In PR #9386 this code is being removed since it's completely unused, however this issue exists so as to maintain a record of the code. We should either reimplement it properly or close this issue if it's not relevant anymore.

Snuffleupagus:
*I said I was done, but this PR is simply me trying to make a positive contribution (since the discussion there is going nowhere) to issue #8966.*

Since extracting `jpg.js` into a separate repository would have a number of issues[1], a solution that wouldn't add much (if any)  overhead is to automatically publish the PDF.js image decoders separately[2] from the main PDF.js library in https://github.com/mozilla/pdfjs-dist.

**Edit:** Note that by skipping a couple of the (unused) larger polyfills, the size of the built `pdf.image_decoders.js` file is thereby reduced by 36% (when compared to bundling all polyfills).

Fixes #9388.
Fixes #8966.

---
[1] *Among the problems:* It would risk regressing performance for the PDF use-cases, it would cause additional maintenance burden for the few regular PDF.js contributors left, and finally would also result is really bad or non-existent test-coverage for the code (since the JPEG code is currently tested indirectly through the PDF.js reference tests).
[2] This would effectively replace the severely outdated bundle found at https://github.com/notmasteryet/jpgjs/blob/master/jpg.js


timvandermeij:
I notice quite a few commits that do not appear to be related to the publishing of the image decoders. Would it perhaps be good to split those off into a separate pull request for clarity?

Snuffleupagus:
> I notice quite a few commits that do not appear to be related to the publishing of the image decoders.

That's a totally fair point! (I was just sort of hoping not having to bother with a bunch of different PRs.)

> Would it perhaps be good to split those off into a separate pull request for clarity?

Half of the commits have now been removed. Please note that I purposely left the commits that move `PageViewport`/`MessageHandler`, since those help reduce the size of `src/shared/util.js` quite a bit.

The "Explicitly list the ignored files in the `src/shared/` folder, for the `gulp lib` build target" commit is also kept, since it meant that it's unnecessary to touch `gulpfile.js` when moving the `MessageHandler`.

All in all, there's now three "support" commits and four "actual" functionality commits left here now :-)

yurydelendik (approved):
Okay, let's do it. Thank you for the patch.

photopea:
So where exactly can we find these jpg.js , jpx.js etc ? I can't see them in https://github.com/mozilla/pdfjs-dist

timvandermeij:
@photopea This becomes available in `pdfjs-dist` when the final 2.0 release is made. We were asked to only update that repository when a stable release happens, so it's no longer updated after every commit like it was before because of possible unstabilities.

Issue #9679, opened 2018-04-19
pdf rendered completely blank


gHensges:
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version: chrome 65
- Operating system and its version: windows 10
- PDF.js version: actual demo project
- Is a browser extension: no
[K BIS.PDF](https://github.com/mozilla/pdf.js/files/1929531/K.BIS.PDF)

Steps to reproduce the problem:
1. 
2.

What is the expected behavior? (add screenshot)

What went wrong? (add screenshot)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

timvandermeij:
The image cannot be loaded: `Warning: Unable to decode image: JpegError: JPEG error: unexpected marker ffd9`

timvandermeij:
Closing since this is fixed by the pull request above.

PR #9730, opened 2018-05-16
Ignore the rest of the image when encountering an EOI (End of Image) marker while parsing Scan data (issue 9679)

Snuffleupagus:
When submitting the "separate image-decoders" PR, I happened to glance briefly at other open JPEG issues and noticed that issue #9679 looked really simple, so...

One caveat here is that I've not run *any* tests locally, so I suppose that this could fail spectacularly :-)

*Edit:* The Windows failures looks intermittent (the `forms` tests), or unrelated to the built-in JPEG decoder. The Linux "failures" are only in `text` tests, and should thus be unrelated as well; they look like the old "Firefox was updated on the bots, and `makeref` is thus needed".

### DIFF
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -41,6 +41,18 @@ let DNLMarkerError = (function DNLMarkerErrorClosure() {
   return DNLMarkerError;
 })();
 
+let EOIMarkerError = (function EOIMarkerErrorClosure() {
+  function EOIMarkerError(message) {
+    this.message = message;
+  }
+
+  EOIMarkerError.prototype = new Error();
+  EOIMarkerError.prototype.name = 'EOIMarkerError';
+  EOIMarkerError.constructor = EOIMarkerError;
+
+  return EOIMarkerError;
+})();
+
 /**
  * This code was forked from https://github.com/notmasteryet/jpgjs.
  * The original version was created by GitHub user notmasteryet.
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
Let's see if this still works...

/botio test

timvandermeij:
It looks like Webpack is requiring more and more memory lately, and I think this is also what I saw when I tried to upgrade PDF.js to Webpack 4, which worked fine on my machine but not on the test bots. There are options to increase allowed memory usage (refer to https://stackoverflow.com/questions/26094420/fatal-error-call-and-retry-last-allocation-failed-process-out-of-memory), but instead of doing that I'd rather find the cause of the excessive memory usage. Indeed it doesn't seem related to this patch at all.

Snuffleupagus:
> [...] but instead of doing that I'd rather find the cause of the excessive memory usage. Indeed it doesn't seem related to this patch at all.

Since finding/fixing the Webpack issues might take some time/effort, could we simply land this patch as-is given that the Linux bot ran successfully (since its failures are unrelated to image decoding)?

timvandermeij:
@yurydelendik @brendandahl It looks like the Windows bot is down. Could you restart it?

Snuffleupagus:
Slightly off-topic, but given that the Windows bot always seem to fail with `FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory` this probably explains why neither `gh-pages` nor `pdfjs-dist` has been updated for a month now.
I'm assuming that's already tracked elsewhere, even though quick searching didn't find anything.

timvandermeij:
Yes, I think the failures here, the `gh-pages` not being updated and Webpack 4 not working all share the same cause. It looks like Webpack requires too much memory on the bots. I'll see if I can find a way to reduce Webpack's memory usage, since I read a few articles that state that the default settings are perhaps a bit too simplistic.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this (and for the triage of other issues)! Good to have the bots working again too.

PR #9740, opened 2018-05-19
Use Dict.getArray, instead of Dict.get, when getting the 'Size' in constructSampled in src/core/function.js

### LINKED ISSUE #9734: Replace `Dict.get` with `Dict.getArray` at one spot in src/core/function.js
*I was recently sorting through and throwing out old notes, when I found this and figured it wouldn't hurt to at least submit an issue.*

Please refer to pull request #7295 for the background. Basically that PR missed one spot, so at https://github.com/mozilla/pdf.js/blob/7ba50448fdef660d22b799b5814af36ad63407c4/src/core/function.js#L189
`get` should be replaced with `getArray` (which could help avoid future bugs).

Since this issue ought to be a good candidate for the `5-good-beginner-bug` label, I'm including a link to https://github.com/mozilla/pdf.js/wiki/Contributing for good measure.

ptmxyz:
Use Dict.getArray, instead of Dict.get, when getting the 'Size' in constructSampled on line 189 of src/core/function.js

Fixes #9734.

### DIFF
--- a/src/core/function.js
+++ b/src/core/function.js
@@ -186,7 +186,7 @@ var PDFFunction = (function PDFFunctionClosure() {
       domain = toMultiArray(domain);
       range = toMultiArray(range);
 
-      var size = toNumberArray(dict.get('Size'));
+      var size = toNumberArray(dict.getArray('Size'));
       var bps = dict.get('BitsPerSample');
       var order = dict.get('Order') || 1;
       if (order !== 1) {


Snuffleupagus:
Drive by comment: Please provide a better commit message! Reading just the commit message, and not the code, it isn't really possible to understand the change. Furthermore the use of the word "last" isn't correct, since there's a number of `Dict.get` occurences in that file (but those should definitely *not* be changed).

A better suggestion might be e.g:
> Use `Dict.getArray`, instead of `Dict.get`, when getting the 'Size' in `constructSampled` in src/core/function.js

pedrotp:
@Snuffleupagus thanks for the feedback! I updated my commit message and title accordingly.

Snuffleupagus:
> I updated my commit message and title accordingly.

It doesn't look like the commit message was updated, only the PR title; did you forget to force push your changes with `git push -f ...`?
Besides there's now a typo in the commit message, since there's no file called "functions.js" (note the extra "s"). Furthermore, the commit message is still unnecessarily imprecise, so why not simply use the suggestion above verbatim!?

Since you'll need to actually push the changes, please update the *commit message* to:
"Use `Dict.getArray`, instead of `Dict.get`, when getting the 'Size' in `constructSampled` in src/core/function.js (PR 7295 follow-up)"

timvandermeij:
@pedrotp Hi, did you manage to update the commit message so we can try to merge it? You can do this locally with `git commit --amend`, entering the new commit message as outlined above and then using `git push -f` to push the new commit. Thank you!

pedrotp:
@timvandermeij @Snuffleupagus sorry, first commit. edited the commit message, description and title, should be good to go now. thanks again for the feedback.

timvandermeij (approved):

timvandermeij:

*(Note that the failing tests on the Windows bot are known intermittent failures, so they are unrelated to this patch.)*

PR #9741, opened 2018-05-20
Build improvements

### LINKED ISSUE #9000: Any reason for not committing the lock file?
Hi!
This is not a bug issue, just a question. Is there any special reason not to be using a lock file? I guess it'd be `package-lock.json`, since the instructions mention using `npm`.

### LINKED ISSUE #9688: Mark as webpack 4 compatible
The [webpack version](https://github.com/mozilla/pdfjs-dist/blob/master/package.json#L19) in pdfjs-dist is currently  "^2.0.0 || ^3.0.0" though it works well with "^4.0.0" as well. 

Can you please mark this as 4.x compatible ? 

Webpack 4 has been out for a while and my applications currently using webpack 4.x are using pdfjs-dist without any problems but when I run npm install I get a spurious peer dependency version mismatch warning which I'd like to get rid of.

timvandermeij:
The objective is to modernize our builds. The first commit updates all packages to their latest versions, except for Webpack. The second commit performs the upgrade to Webpack 4. The third commit includes the `package-lock.json` file in the repository since this is recommended and it allows for reproducible builds using pinned versions.

Fixes #9688.
Fixes #9000.


timvandermeij:
@yurydelendik @brendandahl Unfortunately this doesn't solve the OOM errors on the bots. I think the only solution left is to increase the maximum allowed memory usage on the bots. For that, we need to set the environment variable `NODE_OPTIONS="--max-old-space-size=4096"` for a maximum of 4GB (which I hope is enough) on the bots. Could you make that persist on both bots or should be set it somewhere in the `botio-files-pdfjs` repository?

brendandahl:
@timvandermeij do you want me to bump the default for max_old_space_size on the bots?

brendandahl:
I also had to bump the node version to 8 to use NODE_OPTIONS. Let's see how it does...

brendandahl:
Try with the bot on this time...
/botio-windows test

timvandermeij:
I'm afraid we need to do something about Webpack's excessive memory usage and perhaps the bundle size first before we're able to land this...

brendandahl:
What's strange is your branch seems to build fine for me locally on MacOS, but the windows bot hangs forever (it doesn't run out of memory).

brendandahl:
A bundle of master does eventually finish. It takes 16 minutes though!

timvandermeij:
Wow, that's not great. It works just fine locally on Linux for me too, so I'm not sure why the bots are struggling with this so much...

timvandermeij:
I simplified the commits and rebased onto the current master branch. Here we go...

/botio test

timvandermeij:
Fortunately there is some progress here. The bots are now passing the Webpack stage, which is where things broke previously. The unit tests and font tests also work now, but the regression tests still hang for both bots and browsers.

@brendandahl Can you see what's causing the timeout on the bots? Do they both use at least the latest LTS version of Node.js (8.11.2)? If possible, we can also upgrade to the latest stable version (10.3.0). Or perhaps both bots just need a reboot.

brendandahl:
In the browser console:

```
ReferenceError: pdfjsLib is not defined[Learn More] driver.js:274:5
Driver
http://127.0.0.1:61499/test/driver.js:274:5
<anonymous>
http://127.0.0.1:61499/test/test_slave.html:35:18
```

gulpfile.js
```diff
@@ -146,8 +146,15 @@ function createWebpackConfig(defines, output) {
   var skipBabel = bundleDefines.SKIP_BABEL ||
                   process.env['SKIP_BABEL'] === 'true';
 
+  // Required to expose e.g., the `window` object.
+  output.globalObject = 'this';
+
   return {
+    mode: 'production',
```
Snuffleupagus: As can be seen in https://webpack.js.org/concepts/mode/#usage, this mode apparently enables a number of optimizations. Note in particular `UglifyJsPlugin`, which implies that code will now be minified by default!

Since there's a separate `gulp minified` command already, with necessary options being passed, it seems that you want to *explicitly* opt-out of this new Webpack default by specifying `optimization: { minimize: false, },` below (and probably in the examples/webpack/webpack.config.js file too).

timvandermeij: Done. I noticed that `gulp dist-install` already minifies the bundle, so you're right about that this can also be removed in the example (not sure why it was there before).

timvandermeij:
Yay! Does anyone have time to review this? I have amended the commits to have the latest versions of every package.

brendandahl:
Thanks for the updates!

termux-io:
[`d81ead787607028d954356fa89cfce6a9ba53f1f](Https://jekyllmio.io)`

PR #9742, opened 2018-05-20
Replace deprecated new Buffer(string) constructor with Buffer.from(string)

### LINKED ISSUE #9739: Replace deprecated `new Buffer(string)` constructor with `Buffer.from(string)`
We use `new Buffer(content)` here (where `content` is a string):

https://github.com/mozilla/pdf.js/blob/7ba50448fdef660d22b799b5814af36ad63407c4/gulpfile.js#L127

The use of the `Buffer` constructor has been deprecated, and since Node.js 6.x the `Buffer.from` method is available as a replacement: https://nodejs.org/api/all.html#buffer_buffer_from_buffer_alloc_and_buffer_allocunsafe

Last month, Node.js 4.x became unsupported, so we can switch switch to `Buffer.from` (and drop support for *building* PDF.js since 4.x and earlier are now officially unsupported Node.js versions): https://medium.com/the-node-js-collection/april-2018-release-updates-from-the-node-js-project-71687e1f7742

AbhimanyuVashisht:
Fixes #9739.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -124,7 +124,7 @@ function createStringSource(filename, content) {
   source._read = function () {
     this.push(new Vinyl({
       path: filename,
-      contents: new Buffer(content),
+      contents: Buffer.from(content),
     }));
     this.push(null);
   };


timvandermeij:
I would suggest to amend the commit message (use `git commit --amend`) to the title of the issue instead, e.g.: "Replace deprecated `new Buffer(string)` constructor with `Buffer.from(string)`". This makes it easier to see what the patch fixed when looking through the Git logs. Otherwise, this patch looks good to me.

AbhimanyuVashisht:
Thanks for the feedback. Actually i am new to opensource contributions.

Commit updated, please do check

Yours Sincerely
Abhimanyu Vashisht

PR #9756, opened 2018-05-28
Remove usage of `makeSubStream` from `Type1Parser.extractFontProgram` in src/core/type1_parser.js (issue 9735)

### LINKED ISSUE #9735: Attempt to remove usage of `makeSubStream` in src/core/type1_parser.js
*I was recently sorting through and throwing out old notes, when I found this and figured it wouldn't hurt to at least submit an issue.*

In `Type1Parser`, there's currently two usages of `makeSubStream` that seem completely unnecessary; see https://github.com/mozilla/pdf.js/blob/7ba50448fdef660d22b799b5814af36ad63407c4/src/core/type1_parser.js#L562-L566 and https://github.com/mozilla/pdf.js/blob/7ba50448fdef660d22b799b5814af36ad63407c4/src/core/type1_parser.js#L585-L589

Using `makeSubStream` in that fashion makes the code look more convoluted than necessary, and should also be less efficient since it causes a new `Stream` to be initialized when all that's really needed here is to get a known number of bytes from the already available stream.

At least in theory, although testing is *obviously* required, it should be possible to replace each of those blocks with the following code instead
```javascript
lenIV = program.properties.privateData['lenIV'];
encoded = this.readCharStrings(stream.getBytes(length), lenIV);
```

Since this issue could potentially be a candidate for the `5-good-beginner-bug` label, I'm including a link to https://github.com/mozilla/pdf.js/wiki/Contributing for good measure. Especially the sections "Generating reference images" and "Run testing" is of particular importance, for anyone wanting to work on this.

Snuffleupagus:
This avoids the initialization of, potentially thousands of, unnecessary `Stream` objects, by getting the required number of bytes directly instead.
Given the special behaviour, when `length === 0`, of the `getBytes`/`skip` methods, it's also necessary to handle that particular case to prevent errors when encountering empty CharStrings.

Fixes #9735.

### DIFF
--- a/src/core/type1_parser.js
+++ b/src/core/type1_parser.js
@@ -559,11 +559,9 @@ var Type1Parser = (function Type1ParserClosure() {
               var glyph = this.getToken();
               length = this.readInt();
               this.getToken(); // read in 'RD' or '-|'
-              data = stream.makeSubStream(stream.pos, length);
+              data = (length > 0 ? stream.getBytes(length) : new Uint8Array(0));
               lenIV = program.properties.privateData['lenIV'];
-              encoded = this.readCharStrings(data.getBytes(), lenIV);
-              // Skip past the required space and binary data.
-              stream.skip(length);
+              encoded = this.readCharStrings(data, lenIV);
               this.nextChar();
               token = this.getToken(); // read in 'ND' or '|-'
               if (token === 'noaccess') {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you! I took care of running all tests locally and everything passed.

PR #9758, opened 2018-05-28
Use `fs.unlinkSync` instead of `fs.unlink` when removing `eq.log`

timvandermeij:
*Another issue I found when hardening the test infrastructure.*

This is necessary because Node.js crashes when `fs.unlink` is called without a callback function.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -207,7 +207,7 @@ function startRefTest(masterMode, showRefImages) {
     }
 
     if (fs.existsSync(eqLog)) {
-      fs.unlink(eqLog);
+      fs.unlinkSync(eqLog);
     }
     if (fs.existsSync(testResultDir)) {
       testUtils.removeDirSync(testResultDir);


timvandermeij:
Trivial (and verified locally because the bots don't work reliably yet).

Snuffleupagus:
In https://github.com/mozilla/pdf.js/blob/master/test/font/ttxdriver.js there's a couple of occurrences of `fs.unlink` without callback as well. Could fixing those too perhaps help avoid a bunch of `(node:5308) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.` lines when the font-tests, see e.g. the following (Windows) bot output: http://54.215.176.217:8877/fd76f7812d1afe5/output.txt

timvandermeij:
Yes, it looks like it. I think I somehow excluded the `test` directory in my `grep` for this patch. I'll take care of fixing this; thanks for bringing it up!

PR #9759, opened 2018-05-30
Fix webpack exclude path on windows.

### LINKED ISSUE #9737: gulp generic cannot create build/pdf.js and pdfwork.js
image
image

can not create build/generic/build

brendandahl:
Fixes #9737


timvandermeij (approved):

timvandermeij:
Nice find, thank you for helping out with this!

I have good hopes that this will help for the build improvements PR too. I now have some more ideas to optimize that.

PR #9769, opened 2018-06-01
Reduce the amount of errors logged, primarily in Node.js/Travis, when running the unit-tests

Snuffleupagus:
*Split off from PR #9729, as requested in https://github.com/mozilla/pdf.js/pull/9729#issuecomment-393698402.* /cc @timvandermeij 

Compare the output without these patches https://travis-ci.org/mozilla/pdf.js/builds/386523654, to the output *with* the patches https://travis-ci.org/mozilla/pdf.js/builds/386612114.

### DIFF
--- a/src/display/api_compatibility.js
+++ b/src/display/api_compatibility.js
@@ -15,6 +15,8 @@
 
 let compatibilityParams = Object.create(null);
 if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
+  const isNodeJS = require('../shared/is_node');
+
   const userAgent =
     (typeof navigator !== 'undefined' && navigator.userAgent) || '';
   const isIE = /Trident/.test(userAgent);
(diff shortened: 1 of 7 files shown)


src/display/api_compatibility.js
```diff
@@ -42,9 +44,15 @@ if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
       compatibilityParams.disableStream = true;
     }
   })();
+
+  // Support: Node.js
+  (function checkFontFaceAndImage() {
+    // Node.js is missing native support for `@font-face` and `Image`.
+    if (isNodeJS()) {
+      compatibilityParams.disableFontFace = true;
```
timvandermeij: Now that this is set in the library, I think we can simplify:

https://github.com/mozilla/pdf.js/blob/57e8676180752c375e0d364234565d6240a9a3ec/examples/node/pdf2png/pdf2png.js#L62-L66

to:

```javascript
pdfjsLib.getDocument({ data: rawData, }).then(function (pdfDocument) {
```

timvandermeij:
Aside from the comment above, this looks good to me. I'll try to review this and the other PR tomorrow.

timvandermeij (commented):
I finished reviewing this. If the example comment is adressed, this is good to go.

src/display/api_compatibility.js
Snuffleupagus: > [...] so please check if the compatibility introduced here is indeed included in the library bundle.

Given that https://github.com/mozilla/pdf.js/blob/master/src/display/api_compatibility.js is imported by the API, this should be totally fine; see also https://github.com/mozilla/pdf.js/blob/2921cc0eba053896860e62a4bd1ffb5d29be51e5/src/display/api.js#L29

Furthermore, that fact that the `document` errors are now gone from the Travis output also shows that including these new Node.js specific compatibility options works fine.

Snuffleupagus:
There's apparently still some warnings being printed in the Node.js/Travis output, coming from unit-testing `web/` files (since only the `src/` files use `info`/`warn`).
While the viewer shouldn't be using those helper functions, it probably wouldn't be *that* difficult to add (new) viewer-specific helper functions such that the viewer files wouldn't cause console "spam" either.

timvandermeij (approved):

timvandermeij:
Nice work, thank you!

PR #9771, opened 2018-06-02
Remove the `gulp extension` build target

### LINKED ISSUE #9768: Remove the `gulp extension` build target
After PR https://github.com/mozilla/pdf.js/pull/9566, the `gulp extension` build target is now basically just an alias for `gulp chromium`; see https://github.com/mozilla/pdf.js/blob/e39541ac18d6307c4441aca7b4b345425eeb5c42/gulpfile.js#L416

Hence it seems, at least to me, that the `gulp extension` build target could simply be removed now.

---

Note that it's currently used at *one* other spot in `gulpfile.js`, see https://github.com/mozilla/pdf.js/blob/e39541ac18d6307c4441aca7b4b345425eeb5c42/gulpfile.js#L911
That line can actually be simplified to just `gulp.task('web-pre', ['generic', 'jsdoc']);`, since the Chromium extension isn't included in the [`gh-pages` branch](https://github.com/mozilla/pdf.js/tree/gh-pages) and building it is thus a waste of time/resources in that case.

timvandermeij:
Fixes #9768.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -413,8 +413,6 @@ gulp.task('default', function() {
   });
 });
 
-gulp.task('extension', ['chromium']);
-
 gulp.task('buildnumber', function (done) {
   console.log();
   console.log('### Getting extension build number');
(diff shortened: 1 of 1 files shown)


timvandermeij:
Trivial (over a minute faster to build previews and lighter on the bot resources).

PR #9773, opened 2018-06-03
Fix multibyte decoding in content_disposition.js

### LINKED ISSUE #9616: Chrome extension fails to render properly attachment file name in encoded utf-8

Configuration:
Chrome 65.0.3325.181
OSX 10.13.3
- PDF.js version:
2.0.466
- Is a browser extension:
Chrome extension

1. Make server return properly encoded file name in a header. For example, use this value for header `Content-Disposition=attachment; filename="test-a-test.pdf"; filename*=UTF-8''test-%C3%A4-%D1%82%D0%B5%D1%81%D1%82.pdf`
2. Open the file in a browser tab.

File name should be properly decoded and showed correctly in browser tab header or after file has been download using 'Download' button. For example from above, file name in the browser tab should be "test-ä-тест.pdf".

What went wrong?
File name is wrongly decoded:
image
image

But if file is saved from the chrome incognito tab (no plugin used), name will be decoded properly
image

Rob--W:
I made some mistakes when trying to make the content_disposition.js compatible with non-modern browsers (IE/Edge). As a result text decoding was mostly broken (the original at https://github.com/Rob--W/open-in-browser/blob/55c71eb44e0ad71a3bb443457666fd48421612ac/extension/content-disposition.js works as expected).

I fixed and verified that it works as expected (see the commit description).
The tests in `test-content-disposition.js` contain multiple UTF-8 decoding tests, but just to make sure that #9616 is really fixed, I inserted the following too:

```javascript
console.log('Test from https://github.com/mozilla/pdf.js/issues/9616');
check("attachment; filename=\"test-a-test.pdf\"; filename*=UTF-8''test-%C3%A4-%D1%82%D0%B5%D1%81%D1%82.pdf",
      "test-ä-тест.pdf");
```

Fixes #9616

### DIFF
--- a/src/display/content_disposition.js
+++ b/src/display/content_disposition.js
@@ -78,24 +78,27 @@ function getFilenameFromContentDispositionHeader(contentDisposition) {
   }
   function textdecode(encoding, value) {
     if (encoding) {
-      if (!/^[^\x00-\xFF]+$/.test(value)) {
+      if (!/^[\x00-\xFF]+$/.test(value)) {
         return value;
       }
       try {
         let decoder = new TextDecoder(encoding, { fatal: true, });
         let bytes = new Array(value.length);
         for (let i = 0; i < value.length; ++i) {
-          bytes[i] = value.charCodeAt(0);
+          bytes[i] = value.charCodeAt(i);
         }
         value = decoder.decode(new Uint8Array(bytes));
         needsEncodingFixup = false;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me. Thank you for fixing this!

PR #9774, opened 2018-06-03
Use `fs.unlinkSync` instead of `fs.unlink` when removing files in the font tests

timvandermeij:
*Another issue found when upgrading the test infrastructure. Refer to https://github.com/mozilla/pdf.js/pull/9758#issuecomment-394187950 for the original report, of which this patch is a follow-up.*

This should prevent `[DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated` in the logs.


timvandermeij:
Trivial (the font tests still work and no more deprecation warnings are logged).

PR #9781, opened 2018-06-06
Miscellaneous API clean-up

Snuffleupagus:
This patch series contains various (small) clean-up of API related code. I just recently found this locally and initially thought I'd submitted it as part of the PDF.js version `2.0` project, however I'd apparently missed to do so; better late than never I suppose :-)

/cc @timvandermeij 

---

It'd probably be a good idea to file a follow-up to fix the pre-existing issues with the Error handling in the `node_stream.js` file, as outlined in the last commit message:
> Somewhat unrelated to this patch, but `src/display/node_stream.js` is currently *not* handling errors in a consistent or even correct way; compared with `src/display/network.js` and `src/display/fetch_stream.js`.
Obviously using the `createResponseStatusError` utility function, from `src/display/network_utils.js`, might not make much sense in a Node.js environment. However at the *very* least it seems that `MissingPDFException`, or `UnknownErrorException` when one cannot tell that the PDF file is "missing", should be manually thrown.
> 
> As is, the API (i.e. `getDocument`) is not returning the *expected* errors when loading fails in Node.js environments (as evident from the `pending` API unit-test).


timvandermeij (approved):

timvandermeij:

PR #9793, opened 2018-06-09
Remove `Util.appendToArray` and `Util.prependToArray`

timvandermeij:
The former may be replaced by regular JavaScript array concatenation and the latter is unused. This avoids unnecessary function calls/imports.

### DIFF
--- a/src/core/stream.js
+++ b/src/core/stream.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import { FormatError, isSpace, stringToBytes, Util } from '../shared/util';
+import { FormatError, isSpace, stringToBytes } from '../shared/util';
 import { isDict } from './primitives';
 
 var Stream = (function StreamClosure() {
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
> This avoids unnecessary function calls/imports.

While that's certainly correct, it will however result in more allocations since the existing array won't be re-used; please refer to [this MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat):
> The concat() method is used to merge two or more arrays. This method does not change the existing arrays, but instead returns a new array.

Snuffleupagus:
> This avoids unnecessary function calls/imports.

While that's certainly correct, it will however result in more allocations[1] since the existing array won't be re-used; please refer to [this MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat):
> The concat() method is used to merge two or more arrays. This method does not change the existing arrays, but instead returns a new array.

---
[1] Whether or not this would be a problem, I don't know. Using `push` instead of `concat` might work!?

timvandermeij:
Good point! With the spread syntax, it's now possible to use the regular `push` operation instead of `concat`, so this both avoids additional method calls and reuses the original array. I have pushed the new version. Could you check if it's what you had in mind?

Snuffleupagus:
> Could you check if it's what you had in mind?

That looks great; and it's nice that more (old) code can be removed now!

---

A good follow-up might be to try and replace `Util.extendObj` with (the native) `Object.assign` too. It'd basically just be a search-and-replace exercise, but you'd probably need to add a (core-js) polyfill though.

timvandermeij:
Thank you for checking this!

PR #9795, opened 2018-06-10
Replace `Util.extendObj` by `Object.assign`

timvandermeij:
Follow-up to the suggestion from #9793.

/cc @Snuffleupagus


Rob--W (approved):
LGTM

PR #9800, opened 2018-06-11
checking if the user's input (page number)  is NaN

wonjunior:
Instead of checking if `pdfViewer.currentPageLabel` is equal to himself (with !==) i.e `pdfViewer.currentPageLabel` is `NaN`, use `Number.isNaN(pdfViewer.currentPageLabel)`.


timvandermeij:
I don't think this is correct. While it looks like we're comparing the variable to itself, `currentPageNumber` is actually a quite elaborate setter (see https://github.com/mozilla/pdf.js/blob/7ba50448fdef660d22b799b5814af36ad63407c4/web/base_viewer.js#L212), so it does more than just an assignment. After the logic is complete, we want to check if the condition holds. Unless there is a visible bug (and if so, please open an issue for it), I think this is fine.

PR #9807, opened 2018-06-13
npm install

shinriyo:
it is important


timvandermeij:
This is already mentioned in the file at https://github.com/mozilla/pdf.js#using-pdfjs-in-a-web-application. This version is just another way of using PDF.js, but not the main one, so it's purposely not outlined at the top of the file.

shinriyo:
@timvandermeij sorry. I didn't find it...

PR #8922, opened 2017-09-17
Allow `getOperatorList`/`getTextContent` to skip errors when parsing broken XObjects (issue 8702, issue 8704)

Snuffleupagus:
This patch makes use of the existing `ignoreErrors` property in `src/core/evaluator.js`, see PRs #8240 and #8441, thus allowing us to attempt to recovery as much as possible of a page even when it contains broken XObjects.

Fixes #8702.
Fixes #8704.

*Edit:* More manageable diff with https://github.com/mozilla/pdf.js/pull/8922/files?w=1.

PR #8240, opened 2017-04-05
[api-minor] Always allow e.g. rendering to continue even if there are errors, and add a `stopAtErrors` parameter to `getDocument` to opt-out of this behaviour (issue 6342, issue 3795, bug 1130815)

Snuffleupagus:
Other PDF readers, e.g. Adobe Reader and PDFium (in Chrome), will attempt to render as much of a page as possible even if there are errors present.
Currently we just bail as soon the first error is hit, which means that we'll usually not render anything in these cases and just display a blank page instead.

*NOTE:* This patch changes the default behaviour of the PDF.js API to always attempt to recover as much data as possible, even when encountering errors during e.g. `getOperatorList`/`getTextContent`, which thus improve our handling of corrupt PDF files and allow the default viewer to handle errors slightly more gracefully.
In the event that an API consumer wishes to use the old behaviour, where we stop parsing as soon as an error is encountered, the `stopAtErrors` parameter can be set at `getDocument`.

Fixes, inasmuch it's possible since the PDF files are corrupt, e.g. issue #6342, issue #3795, and [bug 1130815](https://bugzilla.mozilla.org/show_bug.cgi?id=1130815) (and probably others too).

*Note:* Slightly smaller diff with https://github.com/mozilla/pdf.js/pull/8240/files?w=1.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/pdf.js/8240)

PR #9809, opened 2018-06-13
Allow `FontFaceObject.getPathGenerator` to ignore non-embedded fonts during rendering

Snuffleupagus:
This fixes, first of all, the inconsistent error handling that PR #9689 introduced in `InternalRenderTask._scheduleNext`. With the code currently in `master`, a "regular" canvas rendering operation is still left in a perpetually pending state if an error is thrown at any point during rendering (leading to a permanent spinner in the viewer).

Furthermore, rather than stopping all rendering when attempting to draw *non-embedded* fonts as paths, the `FontFaceObject.getPathGenerator` method is now able to ignore errors[1] and let rendering continue with a warning/notification.

*Please note:* While this provides a (perhaps) slightly less jarring behaviour for issue #4244 (and its duplicates), keep in mind that the only "real" solution would be to start shipping standard fonts with the PDF.js library.

---
[1] Similar to PRs #8240 and #8922, this is all controlled with the `stopAtErrors` parameter; please see https://github.com/mozilla/pdf.js/blob/2030d1718f4fb07656ac8ada99dbf7f6f2e4f320/src/display/api.js#L154-L157

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1884,6 +1884,7 @@ var WorkerTransport = (function WorkerTransportClosure() {
             var font = new FontFaceObject(exportedData, {
               isEvalSupported: params.isEvalSupported,
               disableFontFace: params.disableFontFace,
+              ignoreErrors: params.ignoreErrors,
               fontRegistry,
             });
             var fontReady = (fontObjs) => {
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1885,6 +1885,7 @@ var WorkerTransport = (function WorkerTransportClosure() {
               isEvalSupported: params.isEvalSupported,
               disableFontFace: params.disableFontFace,
               ignoreErrors: params.ignoreErrors,
+              onUnsupportedFeature: this._onUnsupportedFeature.bind(this),
               fontRegistry,
             });
             var fontReady = (fontObjs) => {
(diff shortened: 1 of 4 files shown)


timvandermeij:
Something I noticed while looking at this quickly: why do we need the unsupported feature commit exactly? Either we're throwing errors or printing warnings, but in both cases they should end up in the console, which is where the unsupported feature message will be too AFAICT, so wouldn't that result in two messages for the same exception?

Snuffleupagus:
> why do we need the unsupported feature commit exactly?

For the *native* Firefox notification bar, please see https://github.com/mozilla/pdf.js/blob/b590519ea28de84ad7c9d119a85a2cabf842da4d/web/app.js#L760-L761 and then the `MOZCENTRAL` specific code-path in https://github.com/mozilla/pdf.js/blob/b590519ea28de84ad7c9d119a85a2cabf842da4d/web/app.js#L825-L844

Hence `PDFViewerApplication.fallback` is a no-op in non-Firefox builds, so there shouldn't be any duplicate messages. Both the code as written and actual testing agrees with me here :-)

timvandermeij (approved):

timvandermeij:
Really good that rendering now continues; thanks!

PR #5525, opened 2014-12-03
Disable PresentationMode for certain problematic configurations

Snuffleupagus:
Instead of trying to hack around various browser defects, let's just disable PresentationMode in the affected browsers. This patch:
- Disables PresentationMode in IE11+ when the viewer is embedded; fixes #4711.

Issue #4711, opened 2014-04-29
Only first page can be seen in presentation mode under IE11


srees:
IE11 successfully renders PDF inside iframe, but when presentation mode is selected only the first page can be seen. Can't seem to find a way to get to the next page.

djivko:
Most probably this defect is related to this one:
https://connect.microsoft.com/IE/feedback/details/838286/ie-11-incorrectly-reports-dom-element-sizes-in-fullscreen-mode-when-fullscreened-element-is-within-an-iframe

PR #9815, opened 2018-06-14
Remove the `disableFullscreen` hack for embedded IE 11 (issue 9585)

### LINKED ISSUE #9585: Why is fullscreen mode disabled for IE11 when documents are embedded?
I asked about fullscreen mode support for IE11 in https://github.com/mozilla/pdf.js/issues/9572, and was told that it is disabled when documents are embedded. Can you please give me more information about why this is necessary? I'd like to try to fix the issue (if possible) and it would be helpful to know what bugs to look out for that caused fullscreen mode to be disabled for embedded documents IE11. In my initial exploration of this issue, I didn't see any obvious problems when I enabled fullscreen mode for some documents rendered in an iframe in IE11. Thank you!

Snuffleupagus:
It appears that Microsoft silently fixed the problem that required disabling of fullscreen mode, in e.g. `iframe`s, in IE 11; please see issue #4711 and PR #5525 for historical context.

Unfortunately my Google-fu isn't strong enough to find any *official* information regarding the fixing of the browser bug in IE. However testing of the default viewer in IE 11, with this patch applied, it now appears that Presentation Mode is working correctly even in an `iframe` in IE 11.
Further anecdotal evidence that the bug is in fact fixed, is for example that jQuery previously contained a work-around for the IE bug. However, that's removed over two years ago now; see https://github.com/jquery/jquery/commit/ff1a0822f72d2b39fac691dfcceab6ede5623b90 and the issues referenced there.

Given that the default viewer isn't intended to be used as-is anyway (in custom deployments), it didn't seem necessary to keep the `disableFullscreen` option around since it was *only* ever added for compatibility purposes.

Fixes #9585.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -553,10 +553,6 @@ let PDFViewerApplication = {
         support = false;
       }
     }
-    if (support && AppOptions.get('disableFullscreen') === true) {
-      support = false;
-    }
-
     return shadow(this, 'supportsFullscreen', support);
   },
 
--- a/web/app_options.js
+++ b/web/app_options.js
@@ -43,11 +43,6 @@ const defaultOptions = {
     value: '',
     kind: OptionKind.VIEWER,
   },
-  disableFullscreen: {
-    /** @type {boolean} */
-    value: viewerCompatibilityParams.disableFullscreen || false,
-    kind: OptionKind.VIEWER,
-  },
   disableHistory: {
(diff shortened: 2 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good! Not sad to see this go at all.

PR #9819, opened 2018-06-15
Fix regression affecting the horizontal `viewerContainer` offset for small window widths (PR 9816 follow-up)

Snuffleupagus:
*The danger of fixing one bug is that it can, sometimes too easily, cause another one in the process; sorry for not catching this when testing PR #9816 locally.*

When the *entire* viewer becomes narrow enough, as controlled by the `@media all and (max-width: 840px)` media query, the sidebar should (semi-transparently) overlay the `viewerContainer` instead of moving it horizontally. Unfortunately the changes made in PR #9816 caused the relevant CSS rules to be skipped, because of how the inheritance model works in CSS.

I'm well aware that `!important` is usually advised against, since it "breaks" the CSS inheritance model. However in this case it seemed reasonable to use it, to not only fix the bug at hand but to also prevent similar bugs from occurring in the future.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -1847,10 +1847,10 @@ html[dir='rtl'] #documentPropertiesOverlay .row > * {
   }
 
   html[dir='ltr'] #outerContainer.sidebarOpen #viewerContainer {
-    left: 0px;
+    left: 0px !important;
   }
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good catch, thanks!

PR #9821, opened 2018-06-17
Set the correct document title when opening a new file in the `GENERIC` default viewer

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Looks good. Thank you for finding this!

PR #9822, opened 2018-06-17
[api-minor] Refactor the annotation code to be asynchronous

timvandermeij:
This commit is the first step towards implementing parsing for the appearance streams of annotations.

Supersedes #9417.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -26,13 +26,25 @@ import { Stream } from './stream';
 
 class AnnotationFactory {
   /**
+   * Create an `Annotation` object of the correct type for the given reference
+   * to an annotation dictionary. This yields a promise that is resolved when
+   * the `Annotation` object is constructed.
+   *
    * @param {XRef} xref
    * @param {Object} ref
    * @param {PDFManager} pdfManager
    * @param {Object} idFactory
-   * @returns {Annotation}
+   * @return {Promise} A promise that is resolved with an {Annotation} instance.
    */
   static create(xref, ref, pdfManager, idFactory) {
+    return pdfManager.ensure(this, '_create',
+                             [xref, ref, pdfManager, idFactory]);
+  }
+
+  /**
+   * @private
+   */
+  static _create(xref, ref, pdfManager, idFactory) {
     let dict = xref.fetchIfRef(ref);
     if (!isDict(dict)) {
       return;
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -16,7 +16,7 @@
 import {
   assert, FormatError, getInheritableProperty, info, isArrayBuffer, isNum,
   isSpace, isString, MissingDataException, OPS, shadow, stringToBytes,
-  stringToPDFString, Util
+  stringToPDFString, Util, warn
 } from '../shared/util';
 import { Catalog, ObjectLoader, XRef } from './obj';
 import { Dict, isDict, isName, isStream, Ref } from './primitives';
@@ -226,8 +226,7 @@ var Page = (function PageClosure() {
 
       // Fetch the page's annotations and add their operator lists to the
       // page's operator list to render them.
-      var annotationsPromise = this.pdfManager.ensure(this, 'annotations');
-      return Promise.all([pageListPromise, annotationsPromise]).then(
+      return Promise.all([pageListPromise, this._parsedAnnotations]).then(
           function ([pageOpList, annotations]) {
         if (annotations.length === 0) {
           pageOpList.flush(true);
@@ -292,30 +291,44 @@ var Page = (function PageClosure() {
       });
     },
 
-    getAnnotationsData: function Page_getAnnotationsData(intent) {
-      var annotations = this.annotations;
-      var annotationsData = [];
-      for (var i = 0, n = annotations.length; i < n; ++i) {
-        if (!intent || isAnnotationRenderable(annotations[i], intent)) {
-          annotationsData.push(annotations[i].data);
+    getAnnotationsData(intent) {
+      return this._parsedAnnotations.then(function(annotations) {
+        let annotationsData = [];
+        for (let i = 0, ii = annotations.length; i < ii; i++) {
+          if (!intent || isAnnotationRenderable(annotations[i], intent)) {
+            annotationsData.push(annotations[i].data);
+          }
         }
-      }
-      return annotationsData;
+        return annotationsData;
+      });
     },
 
     get annotations() {
-      var annotations = [];
-      var annotationRefs = this._getInheritableProperty('Annots') || [];
-      for (var i = 0, n = annotationRefs.length; i < n; ++i) {
-        var annotationRef = annotationRefs[i];
-        var annotation = AnnotationFactory.create(this.xref, annotationRef,
-                                                  this.pdfManager,
-                                                  this.idFactory);
-        if (annotation) {
-          annotations.push(annotation);
-        }
-      }
-      return shadow(this, 'annotations', annotations);
+      return shadow(this, 'annotations',
+                    this._getInheritableProperty('Annots') || []);
+    },
+
+    get _parsedAnnotations() {
+      const parsedAnnotations =
+        this.pdfManager.ensure(this, 'annotations').then(() => {
+          const annotationRefs = this.annotations;
+          const annotationPromises = [];
+          for (let i = 0, ii = annotationRefs.length; i < ii; i++) {
+            annotationPromises.push(AnnotationFactory.create(
+              this.xref, annotationRefs[i], this.pdfManager, this.idFactory));
+          }
+
+          return Promise.all(annotationPromises).then(function(annotations) {
+            return annotations.filter(function isDefined(annotation) {
+              return !!annotation;
+            });
+          }, function(reason) {
+            warn(`_parsedAnnotations: "${reason}".`);
+            return [];
+          });
+        });
+
+      return shadow(this, '_parsedAnnotations', parsedAnnotations);
     },
   };
 
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -723,9 +723,9 @@ var WorkerMessageHandler = {
       }
     );
 
-    handler.on('GetAnnotations', function wphSetupGetAnnotations(data) {
-      return pdfManager.getPage(data.pageIndex).then(function(page) {
-        return pdfManager.ensure(page, 'getAnnotationsData', [data.intent]);
+    handler.on('GetAnnotations', function({ pageIndex, intent, }) {
+      return pdfManager.getPage(pageIndex).then(function(page) {
+        return page.getAnnotationsData(intent);
       });
     });
 
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -43,22 +43,22 @@ describe('annotation', function() {
     }
   }
 
-  function IdFactoryMock(params) {
-    var uniquePrefix = params.prefix || 'p0_';
-    var idCounters = {
-      obj: params.startObjId || 0,
-    };
-    return {
-      createObjId() {
-        return uniquePrefix + (++idCounters.obj);
-      },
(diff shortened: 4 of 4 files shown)


timvandermeij:
What gives, Chrome? First it fails on both bots, then only on Linux. Firefox thinks everything is fine.

@yurydelendik @brendandahl Can you see why Chrome hangs, or reboot the bots?

timvandermeij:
Chrome still failed :(

brendandahl:
Hmm..seems to fail in the same spot after bug850854. Is this happening on other branches or just here?

timvandermeij:
It looks like it only happens for this commit and only in Chrome. I'll check if I can add more diagnostic information to the test runner to find out exactly which test fails and then I can hopefully try to reproduce it locally in Chrome. Even if I can't, it will be useful to let the test runner output which test actually caused the timeout for future failures too.

timvandermeij:
@brendandahl I made the test runner output the name of the failing test:

```
TEST-UNEXPECTED-FAIL | test wdsg_fitc failed; chrome has not responded in 120s
```

However, I don't have issues with this test locally. It only seems to happen on the Linux bot, which has gotten almost 10 minutes slower since some time ago and which also shows other weird logs (see the message directly above). Could you perhaps reboot the machine, since I have no idea what else it could be?

dmitryskey:
After debugging I found that the problem happens when PDF.js parses link annotations in this document, the stack is the following:

ChunkedStream_ensureByte (chunked_stream.js:110)
ChunkedStream_getByte (chunked_stream.js:163)
Lexer_nextChar (parser.js:714)
Lexer (parser.js:658)
XRef_fetchUncompressed (obj.js:1414)
XRef_fetch (obj.js:1393)
XRef_fetchIfRef (obj.js:1367)
create (annotation.js:40)
getAnnotations (document.js:316)
getOperatorList (document.js:182)
(anonymous) (worker.js:743)
Promise.then (async)
wphSetupRenderPage (worker.js:736)
MessageHandler._onComObjOnMessage (message_handler.js:128)
Worker.postMessage (async)
postMessage (message_handler.js:435)
send (message_handler.js:158)
PDFPageProxy_render (api.js:944)
paintOnCanvas (pdf_page_view.js:580)
draw (pdf_page_view.js:466)
(anonymous) (acroforms.js:46)
Promise.then (async)
(anonymous) (acroforms.js:33)
Promise.then (async)
(anonymous) (acroforms.js:32)
Promise.then (async)
PDFDocumentLoadingTask_then (api.js:495)
(anonymous) (acroforms.js:27)

It only happens in Chrome, the exception was thrown in the code, by some reason 'this.loadedChunks' is undefined

    ensureByte: function ChunkedStream_ensureByte(pos) {
      var chunk = Math.floor(pos / this.chunkSize);
      if (chunk === this.lastSuccessfulEnsureByteChunk) {
        return;
      }

      if (!this.loadedChunks[chunk]) {
        throw new MissingDataException(pos, pos + 1);
      }
      this.lastSuccessfulEnsureByteChunk = chunk;
    },

brendandahl:
Usually if there's a MissingDataException that's unhandled it means that we have some code path that needs to handle it and retry once there is more data.

See the BasePdfManager's [ensure](https://github.com/mozilla/pdf.js/blob/646d81cd09eb3c1880ec26c9305a855d02550bd2/src/core/pdf_manager.js#L158) function and the various things that call it.

dmitryskey:
Yes, it's definitely possible as well, one can add the try/catch block in the document.js:316 and create an annotation promise after the MissingDataException once again.

But is it possible to check also that 128K chunk range buffer helps for this Linux bot in Chrome? The 64K buffer size exists from the very beginning in 2013 and I think it's safe to double that, also it's configurable since 2015. The total amount of memory consumption should be the same but the number of workers will be reduced. I'm not familiar with V8 engine details, but I can imagine that when this patch was added, the additional annotation workers exceeded some limit for the slow Linux bot.

brendandahl:
I'm not sure I follow what you mean by workers, in pdf.js we only spawn one worker ever. Do you mean HTTP requests?

Increasing the limit just hides the issue, we need to address the real issue of not handling the exception correctly.

dmitryskey:
Yea, I was talking about worker.js:435 with 'parseStartXRef', my interpretation of this code was that we create a new worker for parsing XRef reference element. Sorry for the misunderstanding.

I think that just handling an exception hides the issue too, this is a definitely a reproducible regression for this change.

Snuffleupagus:
Looking at the patch, there's a couple of (more-or-less obvious) problems with the implementation:

 - First and foremost, when investigating the bug causing the failures in Chrome, you'll notice that this patch doesn't actually make *any* attempt at loading missing Annotation data (which is evident from the `MissingDataException`s being thrown). Instead, the new code will happily assume that all Annotation data was already loaded, and if/when that's not the case trouble ensues.

   In the old (synchronous) code, loading of Annotation data was done via `PDFManager.ensure()`, but that appears to have been completely removed in this patch. However, simply adding that back isn't sufficient now that the code is supposed to be asynchronous. Instead it seems that the Annotation data needs to be loaded similar to e.g. the page resources (by using the `ObjectLoader`).
  Implementing this fixes the bug, and it also makes conceptual sense when comparing with other (similar) asynchronous code.

 - Secondly, almost all of the Annotation unit-tests are now *completely* broken.
  Running the tests (manually) in a browser, via http://localhost:8888/test/unit/unit_test.html, there's now *loads* of `SPEC HAS NO EXPECTATIONS` warnings all over the place. (Unfortunately the unit-test framework, i.e. Jasmine, apparently cannot be configured to simply fail in that case.)

   All the unit-tests touched in this patch thus needed to be modified to *actually wait* for the now asynchronous Annotation code to complete. Furthermore, by chaining Promises, the "should only accept comb fields when the flags are valid" test-case can be restored to slightly saner looking code.

---

The following patch should address all of the (known) issues, as outlined above:
https://gist.githubusercontent.com/Snuffleupagus/5f1404c8527c03d3925734b37a4de6aa/raw/d6e50c24c792d2e2c2e936e5cd71efe7dc2eeeb8/0001-Ensure-that-all-Annots-resources-are-actually-loaded.patch

(I'd appreciate if my patch is *not* just folded into other one, but instead appended to PR such that the author information is preserved.)

Snuffleupagus:
Looking at the patch, there's a couple of (more-or-less obvious) problems with the implementation:

 - First and foremost, when investigating the bug causing the failures in Chrome, you'll notice that this patch doesn't actually make *any* attempt at loading missing Annotation data (which is evident from the `MissingDataException`s being thrown). Instead, the new code will happily assume that all Annotation data was already loaded, and if/when that's not the case trouble ensues.

   In the old (synchronous) code, loading of Annotation data was done via `PDFManager.ensure()`, but that appears to have been completely removed in this patch. However, simply adding that back isn't sufficient now that the code is supposed to be asynchronous. Instead it seems that the Annotation data needs to be loaded similar to e.g. the page resources (by using the `ObjectLoader`).
  While implementing that fixed the bug, there's a potential risk that using the `ObjectLoader` this way may cause issues with "too much" of the file being forced to load as explained below; **careful analysis would be required here**.
https://github.com/mozilla/pdf.js/blob/646d81cd09eb3c1880ec26c9305a855d02550bd2/src/core/obj.js#L1812-L1815

 - Secondly, almost all of the Annotation unit-tests are now *completely* broken.
  Running the tests (manually) in a browser, via http://localhost:8888/test/unit/unit_test.html, there's now *loads* of `SPEC HAS NO EXPECTATIONS` warnings all over the place. (Unfortunately the unit-test framework, i.e. Jasmine, apparently cannot be configured to simply fail in that case.)

   All the unit-tests touched in this patch thus needed to be modified to *actually wait* for the now asynchronous Annotation code to complete. Furthermore, by chaining Promises, the "should only accept comb fields when the flags are valid" test-case can be restored to slightly saner looking code.

---

The following patch is an attempt at addressing the most severe (known) issues, as outlined above (however note that the `ObjectLoader` usage would need checking):
https://gist.githubusercontent.com/Snuffleupagus/5f1404c8527c03d3925734b37a4de6aa/raw/d6e50c24c792d2e2c2e936e5cd71efe7dc2eeeb8/0001-Ensure-that-all-Annots-resources-are-actually-loaded.patch

Snuffleupagus:
Looking at the patch, there's a couple of (more-or-less obvious) problems with the implementation:

 - First and foremost, when investigating the bug causing the failures in Chrome, you'll notice that this patch doesn't actually make *any* attempt at loading missing Annotation data (which is evident from the `MissingDataException`s being thrown). Instead, the new code will happily assume that all Annotation data was already loaded, and if/when that's not the case trouble ensues.

   In the old (synchronous) code, loading of Annotation data was done via `PDFManager.ensure()`, but that appears to have been completely removed in this patch. However, simply adding that back isn't sufficient now that the code is supposed to be asynchronous. Instead it seems that the Annotation data needs to be loaded similar to e.g. the page resources (by using the `ObjectLoader`).

   While implementing that fixed the bug, there's a potential risk that using the `ObjectLoader` this way may cause issues with "too much" of the file being forced to load as explained below; **careful analysis would be required here**.
https://github.com/mozilla/pdf.js/blob/646d81cd09eb3c1880ec26c9305a855d02550bd2/src/core/obj.js#L1812-L1815

 - Secondly, almost all of the Annotation unit-tests are now *completely* broken.
  Running the tests (manually) in a browser, via http://localhost:8888/test/unit/unit_test.html, there's now *loads* of `SPEC HAS NO EXPECTATIONS` warnings all over the place. (Unfortunately the unit-test framework, i.e. Jasmine, apparently cannot be configured to simply fail in that case.)

   All the unit-tests touched in this patch thus needed to be modified to *actually wait* for the now asynchronous Annotation code to complete. Furthermore, by chaining Promises, the "should only accept comb fields when the flags are valid" test-case can be restored to slightly saner looking code.

---

The following patch is an attempt at addressing the most severe (known) issues, as outlined above (however note that the `ObjectLoader` usage would need checking):
https://gist.githubusercontent.com/Snuffleupagus/5f1404c8527c03d3925734b37a4de6aa/raw/d6e50c24c792d2e2c2e936e5cd71efe7dc2eeeb8/0001-Ensure-that-all-Annots-resources-are-actually-loaded.patch

Snuffleupagus:
Looking at the patch, there's a couple of (more-or-less obvious) problems with the implementation:

 - First and foremost, when investigating the bug causing the failures in Chrome, you'll notice that this patch doesn't actually make *any* attempt at loading missing Annotation data (which is evident from the `MissingDataException`s being thrown). Instead, the new code will happily assume that all Annotation data was already loaded, and if/when that's not the case trouble ensues.

   In the old (synchronous) code, loading of Annotation data was done via `PDFManager.ensure()`, but that appears to have been completely removed in this patch. However, simply adding that back isn't sufficient on its own, now that the code is supposed to be asynchronous, without *also* making sure that `AnnotationFactory.create` is able to load any missing Annotation data.

 - Secondly, almost all of the Annotation unit-tests are now *completely* broken.
  Running the tests (manually) in a browser, via http://localhost:8888/test/unit/unit_test.html, there's now *loads* of `SPEC HAS NO EXPECTATIONS` warnings all over the place. (Unfortunately the unit-test framework, i.e. Jasmine, apparently cannot be configured to simply fail in that case.)

   All the unit-tests touched in this patch thus needed to be modified to *actually wait* for the now asynchronous Annotation code to complete. Furthermore, by chaining Promises, the "should only accept comb fields when the flags are valid" test-case can be restored to slightly saner looking code.

---

The following patch is an attempt at addressing the most severe (known) issues, as outlined above: https://gist.githubusercontent.com/Snuffleupagus/7eda081ba84de47ff196278fa13e2eeb/raw/e882fcd2cb9dc217ea07c3069fdb5261146365cf/0001-Ensure-that-all-Annots-resources-are-actually-loaded.patch

timvandermeij:
Thank you @dmitryskey and @Snuffleupagus for making this change possible!

PR #9832, opened 2018-06-21
Fix a number of regressions/inefficiencies introduced by adding Scroll/Spread modes to the viewer (PR 9208 follow-up)

Snuffleupagus:
This PR contains a couple of *very* quick patches, to address a number of regressions/inefficiencies introduced in the viewer by the Scroll/Spread modes PR.[1]
In addition, most of the new code isn't needed (at all) in `PDFSinglePageView` instances, and this PR also tries to clean that up as well.

Given the regressions, the `2.0.550` pre-release might need to be replaced once this PR lands.

**Edit:** Please note that the viewer is currently *unusable* in older browsers (such as IE11), refer to the first commit.

---
[1] There's probably room for additional clean-up, but someone else is more than welcome to do that :-)

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -597,11 +597,11 @@ class BaseViewer {
       if (!currentPage) {
         return;
       }
-      let hPadding = (this.isInPresentationMode || this.removePageBorders) ?
-        0 : SCROLLBAR_PADDING;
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this! I have updated the project board so we don't forget to replace the pre-release.

PR #9837, opened 2018-06-23
Replace `NotImplementedException` with `unreachable`

### LINKED ISSUE #9836: Replace `NotImplementedException` with `unreachable` in src/core/pdf_manager.js
The `NotImplementedException`, [as defined here](https://github.com/mozilla/pdf.js/blob/34594a5b02394a531c1512ff10b44148225abe67/src/shared/util.js#L446-L456), is currently only being used in a single file; please see [src/core/pdf_manager.js](https://github.com/mozilla/pdf.js/blob/34594a5b02394a531c1512ff10b44148225abe67/src/core/pdf_manager.js).

The rest of the code-base is now using `unreachable`, [as defined here](https://github.com/mozilla/pdf.js/blob/34594a5b02394a531c1512ff10b44148225abe67/src/shared/util.js#L287-L289), for code that by *design* should never be accessed (e.g. abstract methods).
Hence, in `src/core/pdf_manager.js`, all occurrences of `throw new NotImplementedException();` could simply be replaced with `unreachable(/* message here */);` instead. For example, https://github.com/mozilla/pdf.js/blob/34594a5b02394a531c1512ff10b44148225abe67/src/core/pdf_manager.js#L52 would be replaced by `unreachable('Abstract method onLoadedStream');` and similarly for the rest of the affected methods. 

Finally, the `NotImplementedException` code itself can be removed once it's completely unused.

---

Since this issue could potentially be a candidate for the `5-good-beginner-bug` label, I'm including a link to https://github.com/mozilla/pdf.js/wiki/Contributing for good measure. Especially the section "[4. Run lint and testing](https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing)" is of importance, for anyone wanting to work on this.

timvandermeij:
Moreover, convert `src/code/pdf_manager.js` to ES6 classes/syntax while we're here for more compact, readable and modern code.

Fixes #9836.

*I recommend to review this per commit and use the `?w=1` flag especially for the second commit.*

### DIFF
--- a/src/core/pdf_manager.js
+++ b/src/core/pdf_manager.js
@@ -14,216 +14,193 @@
  */
 
 import {
-  createPromiseCapability, createValidAbsoluteUrl, MissingDataException,
-  shadow, unreachable, Util, warn
+  createValidAbsoluteUrl, MissingDataException, shadow, unreachable, warn
 } from '../shared/util';
 import { ChunkedStreamManager } from './chunked_stream';
 import { PDFDocument } from './document';
 import { Stream } from './stream';
 
-var BasePdfManager = (function BasePdfManagerClosure() {
-  function BasePdfManager() {
-    unreachable('Cannot initialize BaseManagerManager');
+class BasePdfManager {
+  constructor() {
+    if (this.constructor === BasePdfManager) {
+      unreachable('Cannot initialize BasePdfManager.');
+    }
   }
 
-  BasePdfManager.prototype = {
-    get docId() {
-      return this._docId;
-    },
+  get docId() {
+    return this._docId;
+  }
 
-    get password() {
-      return this._password;
-    },
+  get password() {
+    return this._password;
+  }
 
-    get docBaseUrl() {
-      var docBaseUrl = null;
-      if (this._docBaseUrl) {
-        var absoluteUrl = createValidAbsoluteUrl(this._docBaseUrl);
-        if (absoluteUrl) {
-          docBaseUrl = absoluteUrl.href;
-        } else {
-          warn('Invalid absolute docBaseUrl: "' + this._docBaseUrl + '".');
-        }
+  get docBaseUrl() {
+    let docBaseUrl = null;
+    if (this._docBaseUrl) {
+      const absoluteUrl = createValidAbsoluteUrl(this._docBaseUrl);
+      if (absoluteUrl) {
+        docBaseUrl = absoluteUrl.href;
+      } else {
+        warn(`Invalid absolute docBaseUrl: "${this._docBaseUrl}".`);
       }
-      return shadow(this, 'docBaseUrl', docBaseUrl);
-    },
+    }
+    return shadow(this, 'docBaseUrl', docBaseUrl);
+  }
 
-    onLoadedStream: function BasePdfManager_onLoadedStream() {
-      unreachable('Abstract method `onLoadedStream` called');
-    },
+  onLoadedStream() {
+    unreachable('Abstract method `onLoadedStream` called');
+  }
 
-    ensureDoc: function BasePdfManager_ensureDoc(prop, args) {
-      return this.ensure(this.pdfDocument, prop, args);
-    },
+  ensureDoc(prop, args) {
+    return this.ensure(this.pdfDocument, prop, args);
+  }
 
-    ensureXRef: function BasePdfManager_ensureXRef(prop, args) {
-      return this.ensure(this.pdfDocument.xref, prop, args);
-    },
+  ensureXRef(prop, args) {
+    return this.ensure(this.pdfDocument.xref, prop, args);
+  }
 
-    ensureCatalog: function BasePdfManager_ensureCatalog(prop, args) {
-      return this.ensure(this.pdfDocument.catalog, prop, args);
-    },
+  ensureCatalog(prop, args) {
+    return this.ensure(this.pdfDocument.catalog, prop, args);
+  }
 
-    getPage: function BasePdfManager_getPage(pageIndex) {
-      return this.pdfDocument.getPage(pageIndex);
-    },
+  getPage(pageIndex) {
+    return this.pdfDocument.getPage(pageIndex);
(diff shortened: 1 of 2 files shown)


src/core/pdf_manager.js
```diff
@@ -15,152 +15,149 @@
...
+        } else {
+          result = value;
         }
-      });
-    },
+        resolve(result);
+      } catch (e) {
+        reject(e);
+      }
+    });
```
Snuffleupagus: This is more old code that could be simplified to e.g.
```javascript
return new Promise(function() {
  const value = obj[prop];
  let result;
  if (typeof value === 'function') {
    result = value.apply(obj, args);
  } else {
    result = value;
  }
  return result;
});
```

src/core/pdf_manager.js
```diff
@@ -15,152 +15,149 @@
...
-  function NetworkPdfManager(docId, pdfNetworkStream, args, evaluatorOptions,
-                             docBaseUrl) {
     this._docId = docId;
     this._password = args.password;
     this._docBaseUrl = docBaseUrl;
     this.msgHandler = args.msgHandler;
     this.evaluatorOptions = evaluatorOptions;
 
-    var params = {
+    const params = {
```
Snuffleupagus: You can probably just remove this temporary object, by doing e.g.
```javascript
this.streamManager = new ChunkedStreamManager(pdfNetworkStream, {
  msgHandler: args.msgHandler,
  url: args.url,
  length: args.length,
  disableAutoFetch: args.disableAutoFetch,
  rangeChunkSize: args.rangeChunkSize,
});
```

src/core/pdf_manager.js
```diff
@@ -15,152 +15,149 @@
...
     this._docId = docId;
     this._password = password;
     this._docBaseUrl = docBaseUrl;
     this.evaluatorOptions = evaluatorOptions;
-    var stream = new Stream(data);
+
+    const stream = new Stream(data);
     this.pdfDocument = new PDFDocument(this, stream);
     this._loadedStreamCapability = createPromiseCapability();
     this._loadedStreamCapability.resolve(stream);
```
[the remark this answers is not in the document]
timvandermeij: Done in the new commit.

src/core/pdf_manager.js
timvandermeij: Done in the new commit.

src/core/pdf_manager.js
timvandermeij: I tried doing this, but I keep on getting unit test failures with this change:

```
TEST-UNEXPECTED-FAIL | creates pdf doc from URL | Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
```

I don't really see why both would be different, but apparantly not explicitly rejecting the promise causes different behaviour?!

Snuffleupagus: That's really strange; anyway I'm sorry for wasting your time with bad advice!

src/core/pdf_manager.js
```diff
@@ -14,216 +14,203 @@
...
 import { PDFDocument } from './document';
 import { Stream } from './stream';
 
-var BasePdfManager = (function BasePdfManagerClosure() {
-  function BasePdfManager() {
-    unreachable('Cannot initialize BaseManagerManager');
+class BasePdfManager {
+  constructor() {
+    if (this.constructor === BasePdfManager) {
+      throw new Error('Cannot initialize BasePdfManager.');
```
Rob--W: Why did you replace `unreachable` with `throw new Error`?

src/core/pdf_manager.js
Rob--W: Jonas' remark on simplicity is correct, in the sense that you don't have to explicitly catch errors and reject errors (that is the default behavior). However, you have to call `resolve(result)` instead of `return result;` in the end.

We don't use async/await anywhere in the codebase (yet?), but if we were, then the code could also be simplified to (without `try-catch` and without the explicit `Promise` constructor):

```javascript
async ensure(obj, prop, args) {
  const value = obj[prop];
  if (typeof value === 'function') {
    return value.apply(obj, args);
  }
  return value;
}
```

src/core/pdf_manager.js
```diff
@@ -14,216 +14,203 @@
...
-    },
+  requestRange(begin, end) {
+    return Promise.resolve();
+  }
 
-    requestLoadedStream: function LocalPdfManager_requestLoadedStream() {
-      return;
-    },
+  requestLoadedStream() {
+    return;
```
Rob--W: This was already in the old code; can we remove the unnecessary `return` statement (here and on the following lines too)?

src/core/pdf_manager.js
timvandermeij: Whoops, I copy-pasted that from a similar piece of code in the `web` folder and we don't have `unreachable` in that context, but indeed I should be using `unreachable` here. Thank you for spotting this!

src/core/pdf_manager.js
timvandermeij: Yes, I'll take care of that.

src/core/pdf_manager.js
timvandermeij: D'oh, that's most likely what I missed... I'll give this another try when I'll address the other comments. I have been reading a bit about `async`/`await` and it does look really nice. Since we've not used it yet, I'll stick with regular promises for now, but it's something we may consider for the future (I think Brendan mentioned it some time ago as well).

Snuffleupagus: Regarding using `async`/`await`: Currently Babel doesn't have great support for this, if you want the output to be IE11 compatible (which we unfortunately do). There's the upstream issue https://github.com/babel/babel/issues/8121 and in-progress PR https://github.com/babel/babel/pull/7076 which looks relevant here, and are probably worth following for updates.

Rob--W (approved):

PR #9848, opened 2018-06-27
Annotation elements can be rendered with custom overrides

bion:
This allows `DefaultAnnotationLayerFactory` to accept an option at runtime for custom rendering of annotation elements. E.g.

```javascript
var pdfPageView = new pdfjsViewer.PDFPageView({
  container: container,
  id: pageNum,
  scale: scale,
  defaultViewport: viewport,
  annotationLayerFactory: new pdfjsViewer.DefaultAnnotationLayerFactory({
    customAnnotationElements: {
      TextWidgetAnnotationElement: CustomTextWidgetAnnotationElement
    }
  }),
  renderInteractiveForms: true,
});
```

This is useful for rendering custom UI over an interactive PDF without having to make a custom build of pdf.js.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -33,79 +33,184 @@ import {
  *   mainly for annotation icons. Include trailing slash.
  * @property {boolean} renderInteractiveForms
  * @property {Object} svgFactory
+ * @property {Object} customElements
  */
 
 class AnnotationElementFactory {
   /**
+   * @param {customAnnotationElements} options
+   */
+  constructor(options) {
+    this.customElements = options.customAnnotationElements || {};
+  }
+
+  /**
    * @param {AnnotationElementParameters} parameters
    * @returns {AnnotationElement}
    */
-  static create(parameters) {
+  create(parameters) {
     let subtype = parameters.data.annotationType;
 
     switch (subtype) {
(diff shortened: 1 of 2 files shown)


timvandermeij:
I understand your particular use case, but I don't think it's a good idea to have this in the library itself. It adds quite a bit of complexity for a feature that is only used by a very small number of users. Moreover, if it's about custom UI you should use the `getAnnotations` API instead to get the annotation data from the core layer and simply not use our default display layer to render them. Refer to:

https://github.com/mozilla/pdf.js/blob/master/web/annotation_layer_builder.js#L56

Note that https://github.com/mozilla/pdf.js/blob/master/web/annotation_layer_builder.js#L87 is what triggers our display layer to render the annotations, but if you want custom UI it's best to replace that instead with your own implementation. That file is located in the `web` folder and you can put your own display layer code there too, so you shouldn't need to change the build of PDF.js at all for your use case as far as I can tell.

There are quite a bit of customization requirements from different users and unfortunately we can't take them all into account in the library for complexity and maintenance reasons. In those situations, it's recommended to make a custom build. Fortunately, as stated above I think that's not even necessary for your use case, so hopefully that will work out for you. Thanks.

bion:
@timvandermeij thank you for the explanation, appreciated.

PR #9849, opened 2018-06-27
Replace the `cloneObj` helper function, in the viewer, with native `Object.assign` (PR 9795 follow-up)

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
I totally forgot we had this, so thank you for cleaning this up!

PR #9853, opened 2018-06-28
Fix re-rendering, using the same canvas, when rendering was previously cancelled (PR 8519 follow-up)

### LINKED ISSUE #9456: immediate render() after cancel()

Configuration:
- Web browser and its version: Google Chrome 64, FireFox 58
- Operating system and its version: Win7
- PDF.js version: 2.0.334
- Is a browser extension: No

```html
<!DOCTYPE html>
<html>
<head>
	<script type="text/javascript" src="//mozilla.github.io/pdf.js/build/pdf.js"></script>
</head>
<body>
	<canvas id="canvas" style="width:50em"></canvas>
	<script type="text/javascript">

var canvasContext = document.getElementById('canvas').getContext('2d');

PDFJS.getDocument('//cdn.mozilla.net/pdfjs/tracemonkey.pdf')
.then(function(pdf) {

	return pdf.getPage(1);
})
.then(function(page) {

	var renderTask = null;

	function renderPage() {

		if ( renderTask !== null ) {

			renderTask.cancel();
			return;
		}

		var viewport = page.getViewport(1, 0);

		renderTask = page.render({
			canvasContext: canvasContext,
			viewport: viewport,
		});

		renderTask.then(function () {
			
			renderTask = null;
		})
		.catch(function(err) {

			renderTask = null;
			if ( err.name === 'RenderingCancelledException' ) {
				
				renderPage();
				return;
			}
			
			console.log('Page render error', err);
		});
	}

	renderPage();
	renderPage();

})

	</script>
</body>
</html>

```

What went wrong?
The following error appear when the second `renderPage()` is called:
`Cannot use the same canvas during multiple render() operations. Use different canvas or ensure previous operations were cancelled or completed.`

In `cancel()` function, `canvasInRendering.delete(this._canvas)` is called before `RenderingCancelledException` is send, then in `initializeGraphics()`,  `canvasInRendering.has(this._canvas)` should be false ??

Snuffleupagus:
Currently if `RenderTask.cancel` is called *immediately* after rendering was started, then by the time that `InternalRenderTask.initializeGraphics` is called rendering will already have been cancelled.
However, we're still inserting the canvas into the `canvasInRendering` map, thus breaking any future attempts at re-rendering using the same canvas. Considering that `InternalRenderTask.cancel` always removes the canvas from the map, I cannot imagine that we'd ever want to re-add it *after* rendering was cancelled (it was likely just a simple oversight in PR #8519).

Fixes #9456.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2380,9 +2380,10 @@ var InternalRenderTask = (function InternalRenderTaskClosure() {
 
   InternalRenderTask.prototype = {
 
-    initializeGraphics:
-        function InternalRenderTask_initializeGraphics(transparency) {
-
+    initializeGraphics(transparency) {
+      if (this.cancelled) {
+        return;
+      }
       if (this._canvas) {
         if (canvasInRendering.has(this._canvas)) {
           throw new Error(
@@ -2393,9 +2394,6 @@ var InternalRenderTask = (function InternalRenderTaskClosure() {
         canvasInRendering.set(this._canvas, this);
       }
 
-      if (this.cancelled) {
-        return;
-      }
       if (this._pdfBug && globalScope.StepperManager &&
           globalScope.StepperManager.enabled) {
         this.stepper = globalScope.StepperManager.create(this.pageNumber - 1);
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
The Windows "failures" were, most likely, caused by a Firefox update (version `61` to `62`).

timvandermeij (approved):

PR #9855, opened 2018-06-29
Auto print feature with auth-built in

137zaira:
https://rocket-lawyer.atlassian.net/browse/DOC-1089


kitsh:
Sorry for the wrong P.R.s, this is the last one.

PR #9860, opened 2018-06-30
Update the localization README to indicate that the translations are imported from the Nightly channel

timvandermeij:


timvandermeij:
Trivial.

PR #9861, opened 2018-07-01
Refactor `PDFFindController` to use the 'pagesinit' event, dispatched on the `eventBus`, to resolve the `_firstPagePromise`

Snuffleupagus:
Rather than having to manually call a method on `PDFFindController` instances from `BaseViewer.setDocument`, thus essentially having to resolve the private `_firstPagePromise` from the "outside", this can be done easily with the 'pagesinit' event dispatched on the `eventBus` instead.
Please note this particular `PDFFindController` code pre-dates the `eventBus` by almost three years, which should explain why the code looks the way it does.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -409,6 +409,7 @@ let PDFViewerApplication = {
 
       this.findController = new PDFFindController({
         pdfViewer: this.pdfViewer,
+        eventBus,
       });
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Good idea to use the event bus here!

Issue #9358, opened 2018-01-12
Broken URL polyfill


rob-murray:
The **URL** polyfill has a couple of issues:

1) The check for native **URL** API is incorrect - it fails on Safari >9 when this has a perfectly good native implementation
2) The polyfill is incomplete and missing (afaict) **searchParams**

I use **pdf.js** in an application and the **URL** web api in various places, on Safari this is broken because the polyfill from this library is used instead of native code (Safari >9 has full implementation i believe) or the other polyfill I include (this fails the check used in this library incorrectly).

Configuration:
- Web browser and its version: Safari >9
- Operating system and its version: Anything
- PDF.js version: latest (v1.8.188)
- Is a browser extension: No

Steps to reproduce the problem:
1. 
2.

What is the expected behavior? 

If a browser does not have native URL api support then an up to date polyfill is used. If a browser does have native URL api support then the polyfill is not used.

What went wrong? 

A browser with native URL api support was forced to use the broken URL polyfill from pdf.js

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

n/a

Snuffleupagus:
> The polyfill is incomplete and missing (afaict) searchParams

For reference, this part is already tracked in issue #8726.

rob-murray:
Sorry, I missed that one when searching for existing issues... feel free to close this then if its already got an open issue.

Snuffleupagus:
> Sorry, I missed that one when searching for existing issues... feel free to close this then if its already got an open issue.

No worries; #8726 only seem to concern the `searchParams` issue, but the *first* part of this issue would appear separate (and thus valid) on its own:

> The check for native URL API is incorrect - it fails on Safari >9 when this has a perfectly good native implementation

The following checks are used to determine if a browser has native `URL` support. 
https://github.com/mozilla/pdf.js/blob/5a52ee0a795ca41f27ede587dc494cd304dc81e1/src/shared/compatibility.js#L189-L207

rob-murray:
Safari seems to encode the **%** so `u.href` above is `http://a/c%2520d`, as a result that check is false - I don't know why Safari does this and if it is actually incorrect or even if it matters. The implementation of the polyfill in this library evaluates this check to be true, as does chrome, firefox, etc.

```
// safari
u = new URL('b', 'http://a'); 
u.pathname = 'c%20d'; 
u.href // "http://a/c%2520d"
```

[...]

PR #9868, opened 2018-07-03
Stop exposing the `URL` polyfill in the global scope

Snuffleupagus:
This moves/exposes the `URL` polyfill similarily to the existing `ReadableStream` polyfill, rather than exposing it globally, to avoid interfering with any "outside" code.
Both the `URL` and `ReadableStream` polyfills are now exposed on the `pdfjsLib` object, such that they are accessible to the viewer components.
Furthermore, the `no-restricted-globals` ESLint rule is also enabled to prevent accidental usage of the native `URL`/`ReadableStream` implementations directly in the `src/` and `web/` folders; see also https://eslint.org/docs/rules/no-restricted-globals

Addresses the remaining TODO in https://github.com/mozilla/pdf.js/projects/6

---

By not exposing the polyfilled `URL` globally, the PR might also "fix" issues #8726 and #9358 for free.

/cc @yurydelendik


timvandermeij (approved):

timvandermeij:
Nice work! It's good to not change the global scope anymore with the `URL` polyfill.

PR #9874, opened 2018-07-07
[api-minor] Include export value for checkboxes

### LINKED ISSUE #8095: Missing property: exportValue of checkboxes AcroForms
[CheckBox.pdf](https://github.com/mozilla/pdf.js/files/794534/CheckBox.pdf)

Configuration:
- PDF.js version: 1.7.225

Acrobat allows multiple checkboxes to share the same name, making them act as radio buttons. While designing the document, every checkbox has an «Export Value». The attached sample pdf has two checkboxes named "Choice", one having "Yes" as export value and the other having "No".

Unfortunately, the «Export Value» of a checkbox is not available among all properties of a «Btn» control. Still, I am able to get it by adding the following line to the function «WidgetAnnotation» of «pdf.worker.js»: 

data.customAppearance = dict.get('AP');

and then, having the appearance of the checkbox, the following code displays then name and export value of every checkbox of a pdf document:

[script.txt](https://github.com/mozilla/pdf.js/files/794535/script.txt)

I simply wish that checkboxes have an exportValue property set.

bion:
Fixes #8095 

U.S. Government forms often use checkboxes as radios by specifying different export values of checkboxes with the same name. See the USCIS [I-130](https://www.uscis.gov/sites/default/files/files/form/i-130.pdf) as an example.

Adding the `exportValue` property to checkbox annotation widgets enables correct rendering of checkbox state for forms that do this.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -742,7 +742,7 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     this.data.pushButton = this.hasFieldFlag(AnnotationFieldFlag.PUSHBUTTON);
 
     if (this.data.checkBox) {
-      this._processCheckBox();
+      this._processCheckBox(params);
     } else if (this.data.radioButton) {
       this._processRadioButton(params);
     } else if (this.data.pushButton) {
@@ -752,11 +752,29 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     }
   }
 
-  _processCheckBox() {
-    if (!isName(this.data.fieldValue)) {
+  _processCheckBox(params) {
+    if (isName(this.data.fieldValue)) {
+      this.data.fieldValue = this.data.fieldValue.name;
+    }
+
+    const customAppearance = params.dict.get('AP');
+    if (!isDict(customAppearance)) {
       return;
     }
-    this.data.fieldValue = this.data.fieldValue.name;
+
+    const exportValueOptionsDict = customAppearance.get('D');
+    if (!isDict(exportValueOptionsDict)) {
+      return;
+    }
+
+    const exportValues = exportValueOptionsDict.getKeys();
+    const hasCorrectOptionCount = exportValues.length === 2;
+    if (!hasCorrectOptionCount) {
+      return;
+    }
+
+    this.data.exportValue = exportValues[0] === 'Off' ?
+      exportValues[1] : exportValues[0];
   }
 
   _processRadioButton(params) {
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -979,7 +979,34 @@ describe('annotation', function() {
       buttonWidgetDict = null;
     });
 
-    it('should handle checkboxes', function() {
+    it('should handle checkboxes with export value', function() {
+      buttonWidgetDict.set('V', Name.get('1'));
+
+      var appearanceStatesDict = new Dict();
+      var exportValueOptionsDict = new Dict();
+
+      exportValueOptionsDict.set('Off', 0);
+      exportValueOptionsDict.set('Checked', 1);
+      appearanceStatesDict.set('D', exportValueOptionsDict);
+      buttonWidgetDict.set('AP', appearanceStatesDict);
+
+      var buttonWidgetRef = new Ref(124, 0);
+      var xref = new XRefMock([
+        { ref: buttonWidgetRef, data: buttonWidgetDict, }
+      ]);
+
+      var annotation = AnnotationFactory.create(xref, buttonWidgetRef,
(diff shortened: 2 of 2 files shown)


dbalatero:
@timvandermeij can this be merged?

src/core/annotation.js
```diff
@@ -752,10 +752,20 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     }
   }
 
-  _processCheckBox() {
+  _processCheckBox(params) {
     if (!isName(this.data.fieldValue)) {
       return;
     }
+
+    var customAppearance = params.dict.get('AP');
```
timvandermeij: Change `var` to `const`.

src/core/annotation.js
```diff
@@ -752,10 +752,20 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
+
+    if (customAppearance) {
+      let exportValues = customAppearance.get('D').getKeys();
```
timvandermeij: Change `let` to `const`.

src/core/annotation.js
```diff
@@ -752,10 +752,20 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
+
+      this.data.exportValue = exportValues[0] === 'Off' ?
```
timvandermeij: I would like to have the unit tests extended so that we can be sure that this functionality does not break in the future. They can be found in `test/unit/annotation_spec.js`.

Snuffleupagus: Furthermore, this patch doesn't appear to do *anywhere near* enough validation done of the structure of this data itself; doing similar checks as in the `_processRadioButton` method just below would seem like a requirement to prevent unnecessary future bugs.

bion:
Hi @timvandermeij and @Snuffleupagus, I added an additional test case and validated the structure of the data as per @Snuffleupagus's recommendation to follow the pattern in `_processRadioButton`.

timvandermeij:
Code-wise this looks good, but I'm still a bit confused about the functionality. I thought the `Opt` field as described in the specification at https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/pdf_reference_1-7.pdf#page=688&zoom=auto,-247,644 would handle this, but since it's optional I can imagine that the state name in the appearance stream is used instead.

However, how should this work when the PDF is rendered? I tried making multiple checkboxes with the same name, but that doesn't make them act as radio buttons in HTML (not without some JavaScript at least). I understand that this patch makes the export value available in the API, but I think we should also have a consistent UI for this when we render them. They are rendered at https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js#L491-L517; note how they have no name.

In short, this patch makes it so that the export value of a checkbox is the state name other than the `Off` state. This sounds fine to me. My main remaining question is: how do we deal with this when rendering the PDF file?

bion:
@timvandermeij This will not result in a change for the default annotation rendering.

This change is useful in cases when the API consumer is interested in which checkbox has been interacted with by the user. Exposing the export value answers the question of "what is the semantic meaning of this checkbox".

Our use-case at [Boundless](https://www.boundless.com/) is an example of how this is useful. In our UI the user's interactions with the annotation layer are used to generate PDFs with the AcroForm fields filled out. If you look at the USCIS [I-130](https://www.uscis.gov/sites/default/files/files/form/i-130.pdf) you'll see that question 1 on page 1 is a set of four checkboxes, all with the same field name, but each with a different export value. Exposing the export value to the annotation layer allows us to provide a UI that can interpret the semantics of the user's interaction with those checkboxes.

timvandermeij:
Thank you for the clarification. In the USCIS I-130 PDF file, it mentions "Select only one box" for question 1 on page 1. To me this implies that the user could in fact select multiple, but is asked to only select one. Is it therefore OK to not change the default rendering and just render multiple checkboxes that may be checked, and leave it up to the API consumer to be able to work with multiple selections if the user does so? I.e., we don't have to enforce the "act as radio buttons" part in the UI?

bion:
@timvandermeij Updated the commit message per your instructions. Re: the checked button behavior, the USCIS I-130 PDF is often printed which is likely the reasoning behind their explicit instructions.

> I.e., we don't have to enforce the "act as radio buttons" part in the UI; is that what other viewers do?

Adobe Acrobat and Adobe Reader enforce the radio-like behavior but macOS Preview and Google Chrome do not. Given the discrepancy there I think leaving it up to the API consumer makes good sense.

timvandermeij (approved):

timvandermeij:
Looks good. Thank you for your contribution!

bion:
@timvandermeij thanks for supporting the change and getting it in!

PR #9880, opened 2018-07-08
Remove the `jp-JP-mac` locale

### LINKED ISSUE #9878: [Regression] The `gulp locale` command prints "Skipping invalid locale: ja-JP-mac" warnings on every invocation
After PR #9859, the `gulp locale` command is now printing `Skipping invalid locale: ja-JP-mac` on *every* invocation. The message is originating in https://github.com/mozilla/pdf.js/blob/b39a2ba57a11ed08f66fbf3844d6e51d488f3a83/gulpfile.js#L502-L505 and it should be easy to enough to "fix" this. However, please note that https://github.com/mozilla/pdf.js/pull/3766#issuecomment-37820267 *explicitly* asked for the `ja-JP-mac` locale to be excluded[1], so the preferred solution might instead be to remove the offending language code (and the [locale files themselves](https://github.com/mozilla/pdf.js/tree/master/l10n/ja-JP-mac)) from https://github.com/mozilla/pdf.js/blob/b39a2ba57a11ed08f66fbf3844d6e51d488f3a83/external/importL10n/locales.js#L30

/cc @timvandermeij 

---

[1] I don't recall the motivation for excluding that particular locale, maybe @yurydelendik can chime in.

timvandermeij:
It gives warnings when running `gulp locale`, in a previous patch it was explicitly asked to remove it and there is quite a bit of discussion about the validity of the locale at https://bugzilla.mozilla.org/show_bug.cgi?id=418485. In short, let's remove it to prevent these issues.

Fixes #9878.

*Moreover, at https://bugzilla.mozilla.org/show_bug.cgi?id=418485 there is a comment stating that "[...] ja-JP-mac is no longer exposed to web content [...]", so this most likely explains why it was asked to remove it in a previous patch since our language detection would never pick it up anyway.*


timvandermeij:
Trivial.

PR #8864, opened 2017-09-01
Replace the `isArray` helper function with the native `Array.isArray` function

Snuffleupagus:
*Follow-up to PR #8813.*

PR #9881, opened 2018-07-09
Replace the remaining occurences of `instanceof Array` with `Array.isArray()`

Snuffleupagus:
*Follow-up to PRs #8864 and #8813.*

As explained in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray, `instanceof Array` can have inconsistent behavior. To ensure that only `Array.isArray` is used, an ESLint plugin/rule is added to enforce this.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #9887, opened 2018-07-10
Convert more code in the `/src` folder to use ES6 classes, such that `Util.inherit` can be removed

Snuffleupagus:
Compared to the time when ES6 classes were originally introduced in the `/web` folder, things have now improved/changed:
 - A fair number of performance improvements for ES6 classes have landed in Firefox; please see https://bugzilla.mozilla.org/show_bug.cgi?id=1167472 for details.
 - ES6 classes are now used very frequently in the Firefox front-end code.
 - There's already (some) ES6 class usage in the `/src` folder, with (seemingly) no ill effects.
 - By converting more code in the `/src` folder to use proper ES6 classes, the `Util.inherit` function may be replaced by native code.

*Much more manageable diff with: https://github.com/mozilla/pdf.js/pull/9887/files?w=1*

### DIFF
--- a/src/core/cmap.js
+++ b/src/core/cmap.js
@@ -14,8 +14,8 @@
  */
 
 import {
-  CMapCompressionType, FormatError, isString, MissingDataException,
-  unreachable, Util, warn
+  CMapCompressionType, FormatError, isString, MissingDataException, unreachable,
+  warn
 } from '../shared/util';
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Nice clean-up! It's really good to reduce the dependecy on `src/shared/util.js` and use native code where possible.

PR #9895, opened 2018-07-15
fix cannot read div of undefined

hamzahmurrar:
fix cannot read div of undefined

### DIFF
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -354,7 +354,11 @@ function backtrackBeforeAllVisibleElements(index, views, top) {
   // (case 1, 2, or 4), which means finding a page that is above the current
   // page's top. If the found page is partially visible, we're definitely not in
   // case 3, and this assumption is correct.
-  let elt = views[index].div;
+  let elt = views[index] && views[index].div;
+
+  if (!elt) {
+    return 0;
+  }
   let pageTop = elt.offsetTop + elt.clientTop;
 
   if (pageTop >= top) {


timvandermeij:
I'm going to close this since no steps to reproduce to verify the patch were given and this scenario should not occur according to the comment in that function (and I've not seen it happening before). If there is in fact a bug here, please open an issue with steps to reproduce first so we can verify that it's indeed a bug in the code or something else (for example an issue in a custom deployment of PDF.js). Thank you.

hamzahmurrar:
try this please and zoom in for multiple times.

https://vadimdez.github.io/ng2-pdf-viewer/

timvandermeij:
I only get `ERROR Error: "Uncaught (in promise): TypeError: t[e] is undefined`, but the zooming works just fine. This looks more like an integration error in the `ng2-pdf-viewer` than a problem in PDF.js. Moreover, that viewer doesn't even seem to use the logic touched in this patch.

Snuffleupagus:
Unfortunately you still haven't provided *any* information to indicate that there's a bug in the PDF.js library/viewer *itself*.
With the limited information provided thus far, as mentioned in https://github.com/mozilla/pdf.js/pull/9895#issuecomment-405375358, this appears to be an issue with https://github.com/vadimdez/ng2-pdf-viewer/ which, while it's using the PDF.js library, is not connected to the PDF.js project as such.

Are you able to reproduce this problem with the default PDF.js viewer, which can be found at https://mozilla.github.io/pdf.js/web/viewer.html?
I cannot do that, and it would be (somewhat) surprising if you could since this sort of breakage would most likely have been caught already.

---

However, looking (quickly) at the implementation of https://github.com/vadimdez/ng2-pdf-viewer/, in particular the following file https://github.com/VadimDez/ng2-pdf-viewer/blob/master/src/app/pdf-viewer/pdf-viewer.component.ts, does highlight a number of questionable practices.

Note that it creates a `PDFViewer` instance, which is totally fine (and expected behaviour); see https://github.com/VadimDez/ng2-pdf-viewer/blob/c14aca86148d20656765531cc01a955f55915458/src/app/pdf-viewer/pdf-viewer.component.ts#L210

What's not, at all, expected however is that the `renderPage` method is then *manually* clearing out the DOM; see https://github.com/VadimDez/ng2-pdf-viewer/blob/c14aca86148d20656765531cc01a955f55915458/src/app/pdf-viewer/pdf-viewer.component.ts#L389 before *manually* creating `PDFPageView` instances at https://github.com/VadimDez/ng2-pdf-viewer/blob/c14aca86148d20656765531cc01a955f55915458/src/app/pdf-viewer/pdf-viewer.component.ts#L409 for rendering.
The whole point of using `PDFViewer`, rather than `PDFPageView` directly, is that the viewer provides abstractions for a lot of the operations (such as initializing the necessary `PDFPageView` instances, and controlling/triggering rendering). However, this mixing of `PDFViewer` and `PDFPageView` usage is *not* intended behaviour and is thus explicitly *unsupported* (for the PDF.js library point of view).

*To summarize:* The way that the https://github.com/VadimDez/ng2-pdf-viewer project is using the PDF.js library appears to be incorrect, which is a bug that should be reported to that project instead.

subatomicglue:
This may be true (and I *am* pushing on the `ng2-pdf-viewer` project to make the change):
```
To summarize: The way that the https://github.com/VadimDez/ng2-pdf-viewer project is using the PDF.js library appears to be incorrect, which is a bug that should be reported to that project instead.
```

But adding a bounds check for `views.length <= index` doesn't seem unreasonable either:
```
function backtrackBeforeAllVisibleElements(index, views, top) {
  if (index < 2 || views.length <= index) {
    return index;
  }
  var elt = views[index].div;
```

Though this totally makes sense as a bandwidth/annoyance issue for you:
```
However, this mixing of PDFViewer and PDFPageView usage is not intended behaviour and is thus explicitly unsupported (from the PDF.js library point of view).
```

Snuffleupagus:
> But adding a bounds check for views.length <= index doesn't seem unreasonable either:

Honestly, that would make even less sense than the change that *this* PR suggested. (Although this one wasn't perfect either, since `return 0;` probably ought to have been `return index;` instead.)

Anyway, as explained above, the problem (but not the only one) is that https://github.com/vadimdez/ng2-pdf-viewer/ is manually clearing out the DOM. Essentially it's "pulling the rug from under the feet" of the [`getVisibleElements`](https://github.com/mozilla/pdf.js/blob/20cd1b354b6cad6c040a507d962175d684721d5f/web/ui_utils.js#L416) function; thus outright failing, rather than swallowing the error, seem appropriate here.

subatomicglue:
You're right, and I didn't realize this issue was a PR, but now reviewing the code, something similar *would* be a better catch than the bounds check I proposed. Yes!
```
  let elt = views[index] && views[index].div;
   if (!elt) {
    return index;
  }
```

Cool.  Totally understood `ng2-pdf-viewer` is pulling the rug. :)

PR #9897, opened 2018-07-15
Prefer the Width/Height of the image data, rather than the image dictionary, for JPEG 2000 images (issue 9650)

### LINKED ISSUE #9650: Corrupted displaying of attached file
Link to PDF file here:
[Kred.Eingangsrechnungen.pdf](https://github.com/mozilla/pdf.js/files/1898753/Kred.Eingangsrechnungen.pdf)

Configuration:
- Web browser and its version: Google Chrome Version 65.0.3325.181 (Official Build) (64-bit)
- Operating system and its version: Ubuntu 17.10
- PDF.js version: 
- Is a browser extension: No

1. Open demo https://mozilla.github.io/pdf.js/web/viewer.html
2. Open attached file

Correct opened file
image

What went wrong?
Something wrong with displaying
image

Snuffleupagus:
- Add more validation of the /Filter entry, in image dictionaries, to the `PDFImage` constructor

   Given that the code is currently assuming that the /Filter entry is a `Name`, it probably cannot hurt to actually ensure that's the case.

 - Prefer the Width/Height of the image data, rather than the image dictionary, for JPEG 2000 images (issue 9650)

   According to the PDF specification, see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#page=45
   > When using the JPXDecode filter with image XObjects, the following changes to and constraints on some entries in the image dictionary shall apply (see 8.9.5, "Image Dictionaries" for details on these entries):
   >
   >  - Width and Height shall match the corresponding width and height values in the JPEG2000 data.
   >
   >  - . . .

   Hence it seems reasonable to use the Width/Height of the image data *itself*, rather than the image dictionary when there's a mismatch. Given that JPEG 2000 images are already being parsed, in order to obtain basic parameters, the actual Width/Height is readily available in the `PDFImage` constructor.

Fixes #9650.

### DIFF
--- a/src/core/image.js
+++ b/src/core/image.js
@@ -91,6 +91,9 @@ var PDFImage = (function PDFImageClosure() {
             var jpxImage = new JpxImage();
             jpxImage.parseImageProperties(image.stream);
             image.stream.reset();
+
+            image.width = jpxImage.width;
+            image.height = jpxImage.height;
             image.bitsPerComponent = jpxImage.bitsPerComponent;
             image.numComps = jpxImage.componentsCount;
             break;
@@ -105,13 +108,23 @@ var PDFImage = (function PDFImageClosure() {
     }
     // TODO cache rendered images?
 
-    this.width = dict.get('Width', 'W');
-    this.height = dict.get('Height', 'H');
+    let width = dict.get('Width', 'W');
+    let height = dict.get('Height', 'H');
 
-    if (this.width < 1 || this.height < 1) {
-      throw new FormatError(`Invalid image width: ${this.width} or ` +
-                            `height: ${this.height}`);
+    if ((Number.isInteger(image.width) && image.width > 0) &&
+        (Number.isInteger(image.height) && image.height > 0) &&
+        (image.width !== width || image.height !== height)) {
+      warn('PDFImage - using the Width/Height of the image data, ' +
+           'rather than the image dictionary.');
+      width = image.width;
+      height = image.height;
     }
+    if (width < 1 || height < 1) {
+      throw new FormatError(`Invalid image width: ${width} or ` +
+                            `height: ${height}`);
+    }
+    this.width = width;
+    this.height = height;
 
     this.interpolate = dict.get('Interpolate', 'I') || false;
     this.imageMask = dict.get('ImageMask', 'IM') || false;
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
@Rob--W Is this something that you'd be, time permitting of course, willing/able to review?

src/core/image.js
```diff
@@ -86,11 +86,9 @@ var PDFImage = (function PDFImageClosure() {
     if (dict.has('Filter')) {
       var filter = dict.get('Filter').name;
       if (filter === 'JPXDecode') {
-        var jpxImage = new JpxImage();
-        jpxImage.parseImageProperties(image.stream);
-        image.stream.reset();
-        image.bitsPerComponent = jpxImage.bitsPerComponent;
-        image.numComps = jpxImage.componentsCount;
+        const jpxProperties = JpxImage.parseImageProperties(image.stream);
```
Rob--W: It seems that `JpxImage` is always used as the temporary object to store output (your `jpxProperties` here).

https://github.com/mozilla/pdf.js/blob/ae07adf143a7680480fdb34df63910466448f79b/src/core/jpx_stream.js#L53-L54
https://github.com/mozilla/pdf.js/blob/620da6f4df6eb883ff89580b643ef0516035e0d0/src/core/image.js#L89-L90

Since `parseImageProperties` does not depend on pre-existing `JpxImage` instance methods or members, it seems to make sense to turn it into a static method. (in the past this was different: 894d9fe085b369b8857b87e277bbf67bc98fa2f6).

However, I'd still like to see this commit removed, for consistency with our other image decoding implementations (jpg.js, jbig2.js) also use the pattern of a instantiating a temporary class for parsing images.

Rob--W (commented):
Functionality-wise, looks good. Can you remove the refactor-commit, for the reason below?

PS. Reviewing the separate commits was a bit odd since it you had rebased the commits and ended up with the same CommitDate for every commit, and Github shows the commits in the wrong order because of that: https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/
In the future, if you are able to, try to not to have distinct times, e.g. by adding `exec sleep 2` in between the commits (if you use an interactive rebase).

src/core/image.js
Snuffleupagus: Fair point regarding the consistency, the re-factoring commit has been removed.

Rob--W (approved):

Rob--W:
Thanks for this PR!

timvandermeij:
I'm making reference images in another PR already, so there is no need to do that here anymore.

PR #9898, opened 2018-07-16
Change `waitOnEventOrTimeout`, in web/ui_utils.js, to return a regular `Promise` and remove the `createPromiseCapability` import

Snuffleupagus:
*Another small piece of clean-up of code I've previously written; follow-up to PR 8775.*

Importing `createPromiseCapability`, and then using it in just *one* spot, seems unnecessary since the `waitOnEventOrTimeout` function may just as well return a regular `Promise` directly.

### DIFF
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -13,8 +13,6 @@
  * limitations under the License.
  */
 
-import { createPromiseCapability } from 'pdfjs-lib';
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice work!

PR #9910, opened 2018-07-22
Useless concatenation of literals error by ESLint

PranjalAgni:
Fixed Error: Was concatenating two string literals line: 1524 app.js

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
 /* Copyright 2012 Mozilla Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
The [`no-useless-concat`](https://eslint.org/docs/rules/no-useless-concat) rule is already enabled, see https://github.com/mozilla/pdf.js/blob/7e139776692789fee4b2cedb443715665c5adebe/.eslintrc#L89 and linting is passing as expected. I really don't understand the purpose of this PR at all, especially since it touches virtually an entire file and given that basically *every* single change violates the prevailing (and agreed upon) coding style!

To save time, both for yourself and others, it's recommended to open an issue *first* if you have any questions about the coding style rather than submit a PR like this one

PranjalAgni:
Okay it was my first PR I didn't do it correctly!!
Will keep in mind before doing any PR now....

PR #9913, opened 2018-07-22
Provide custom messages for the `no-restricted-globals` ESLint rule, and refactor the `.eslintrc` files (PR 9868 follow-up)

Snuffleupagus:
Without providing useful (custom) error messages for the `no-restricted-globals` rule, see https://eslint.org/docs/rules/no-restricted-globals, it's quite likely that the rule will be incorrectly disabled rather than the required globals being imported as intended.

To reduced duplication of the `no-restricted-globals` rule in multiple `.eslintrc` files, it's instead moved to the top-level `.eslintrc` file and disabled as needed on a folder/file basis outside of `/src` and `/web`.


timvandermeij (approved):

timvandermeij:
Nice find! I didn't know that was possible, but it should help to make it clear why we have this.

PR #9917, opened 2018-07-23
Add a `BasePreferences.getAll` method and use it to fetch all Preferences at once in `PDFViewerApplication._readPreferences`; ensure that the `externalLinkTarget` option is correctly set when the viewer is embedded

Snuffleupagus:

@timvandermeij The first patch fixes a (small) regression from PR #9479, hence this one should probably be included in the updated `2.0.xyz` release too (as tracked in https://github.com/mozilla/pdf.js/projects/5).

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -142,6 +142,12 @@ let PDFViewerApplication = {
     }).then(() => {
       return this._initializeL10n();
     }).then(() => {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Much simpler; thanks!

PR #9920, opened 2018-07-24
[api-minor] Add an `IsLinearized` property to the `PDFDocument.documentInfo` getter, to allow accessing the linearization status through the API (via `PDFDocumentProxy.getMetadata`)

Snuffleupagus:


timvandermeij:
Should we label this as `[api-minor]` because it's an addition?

timvandermeij:
For completeness I'm also providing the Dutch translations. If you would like to include them in this patch, please feel free to; otherwise we can make a follow-up. Here they are:

```
document_properties_linearized=Snelle webweergave:
document_properties_linearized_yes=Ja
document_properties_linearized_no=Nee
```

Snuffleupagus:
> Should we label this as [api-minor] because it's an addition?

You're obviously correct (the change may be really small, but it's still adding "new" API functionality).

> For completeness I'm also providing the Dutch translations.

Nice, thank you! I'll add those as well soon (and probably some kind of localization comment as well).

timvandermeij (approved):

Issue #2813, opened 2013-02-26
Extrememly slow rendering on this pdf


qlum:
This particular pdf takes more then 10 minutes to load on pdf.js on firefox 20 while If I where to compare it to the internal pdf reader on chrome that one only takes 5 seconds to load the same file. While firefox is loading the file its also slow and not very responsive, if you where to zoom in after it loads it will be a big pain as well. 
https://www.prorail.nl/sites/default/files/spoorkaart_prorail_april_2013.pdf

timvandermeij:
Closing since this PDF file now renders in around six seconds after all performance optimizations since this issue was opened, which brings performance up to par with e.g., Okular and PDFium according to the original report. If there is anything more we can do, a separate issue should be opened with the exact details of what can be improved.

Piruzzolo:
@timvandermeij it loads very slowly on 77.0.1

PR #9925, opened 2018-07-26
Attempt to estimate the minimum required `buffer` length when initializing `StreamsSequenceStream` instances

Snuffleupagus:
For most other `DecodeStream` based streams, we'll attempt to estimate the minimum `buffer` length based on the raw stream data. The purpose of this is to avoid having to unnecessarily re-size the `buffer`, thus reducing the number of *intermediate* allocations necessary when decoding the stream data.
However, currently no such optimization is attempted for `StreamsSequenceStream`, and given that they can often be quite large that seems unfortunate. To improve this, at least somewhat, this patch utilizes the raw sizes of the `StreamsSequenceStream` sub-streams to estimate the minimum required `buffer` length.

Most likely this patch won't have a huge effect on memory consumption, however for pathological cases it should help reduce peak memory usage slightly.
One example is the PDF file in issue #2813, where currently the `StreamsSequenceStream` instances would grow their `buffer`s as `2 MiB -> 4 MiB -> 8 MiB -> 16 MiB -> 32 MiB`. With this patch, the same stream `buffers`s grow as `8 MiB -> 16 MiB -> 32 MiB`, thus avoiding a total of `12 MiB` of *intermediate* allocations (since there's two `StreamsSequenceStream` used, for rendering/text-extraction).

### DIFF
--- a/src/core/stream.js
+++ b/src/core/stream.js
@@ -126,6 +126,8 @@ var DecodeStream = (function DecodeStreamClosure() {
   var emptyBuffer = new Uint8Array(0);
 
   function DecodeStream(maybeMinBufferLength) {
+    this._rawMinBufferLength = maybeMinBufferLength || 0;
+
     this.pos = 0;
     this.bufferLength = 0;
     this.eof = false;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice find!

PR #9931, opened 2018-07-27
Refactor `getPage` (in the worker), and attempt to use the `Linearization` dictionary to lookup the first Page

### LINKED ISSUE #9716: Take advantage of linearization dictionary to get the first page
In the base viewer example `pdfDocument.getPage(1)` is used to size the viewport. 

When PDF.js gets the first page, it doesn't take advantage of the linearization dictionary, instead it relies on grabbing the full page tree (in a large PDF, this could be a performance issue due to multiple range requests being made to grab the table).

It seems, PDF.js should take advantage of the linearization dictionary when available to speed up getting the first page.

Snuffleupagus:
As expected, using the `Linearization` data doesn't appear to have had any (noticeable) performance impact. This can probably be attributed to [`Catalog.getPageDict`](https://github.com/mozilla/pdf.js/blob/51b0e60f9b344bf43b833295103505c58845ea19/src/core/obj.js#L468) only fetching the required nodes, and *not* the entire `Pages` tree, in combination with caching of already resolved nodes (in `pageKidsCountCache`).

*Edit:* To clarify the above, for Linearized files this patch *does* result in (at most) a handful of fewer invocations of [`this code`](https://github.com/mozilla/pdf.js/blob/d6f378fbafed2079310ed46a8efc519aa77a271a/src/core/obj.js#L486) since the first Page can be accessed directly.
So while the amount of data being loaded when fetching the *first* Page is reduced, the difference is really tiny in practice. Hence it seems that unless the server/connection is really slow, the difference would most likely not be seen/felt. (Also, keeping in mind that the default viewer will pre-render the next/previous page.)

Fixes #9716.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -13,12 +13,13 @@
  * limitations under the License.
  */
 
-import { Catalog, ObjectLoader, XRef } from './obj';
-import { Dict, isDict, isName, isStream } from './primitives';
 import {
-  getInheritableProperty, info, isArrayBuffer, isNum, isSpace, isString,
-  MissingDataException, OPS, shadow, stringToBytes, stringToPDFString, Util
+  assert, FormatError, getInheritableProperty, info, isArrayBuffer, isNum,
+  isSpace, isString, MissingDataException, OPS, shadow, stringToBytes,
+  stringToPDFString, Util
 } from '../shared/util';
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good to have this; thanks!

PR #9933, opened 2018-07-27
README.md: suggest usage of https instead of git protocol

perlun:
The `git` protocol is unencrypted which means other parties could potentially eavesdrop your traffic. `https` or `ssh` is often encouraged because of this. (For example, the Ruby package manager `bundler` prints a warning when `git` sources are being used.)


timvandermeij (approved):

timvandermeij:
Thank you!

PR #8064, opened 2017-02-12
[api-minor] Refactor fetching of built-in CMaps to utilize a factory on the `display` side instead, to allow users of the API to provide a custom CMap loading factory (e.g. for use with Node.js)

Snuffleupagus:
Currently the built-in CMap files are loaded in `src/core/cmap.js` using `XMLHttpRequest` directly. For some environments that might be a problem, hence this patch refactors that to instead use a factory to load built-in CMaps on the main thread and message the data to the worker thread.

This is inspired by other recent work, e.g. the addition of the `CanvasFactory`, and to a large extent on the IRC discussion starting at http://logs.glob.uno/?c=mozilla%23pdfjs&s=12+Oct+2016&e=12+Oct+2016#c53010.

*Please note:* While it certainly may be possible to improve the patch, it does work just fine locally as-is, i.e. the viewer still works and all unit/font/reference tests pass.

@yurydelendik Since this PR attempts to implement your idea from http://logs.glob.uno/?c=mozilla%23pdfjs&s=12+Oct+2016&e=12+Oct+2016#c53018, I'd appreciate if you could provide feedback on the implementation when you've got time.

***Edit:*** Based on this PR, we should e.g. be able to get the CMap unit-tests running on Travis, by providing a custom Node.js factory for the unit-test.

***Edit 2:*** Also fixes #4794, courtesy of the second commit.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/pdf.js/8064)

PR #9935, opened 2018-07-28
[Regression] Convert `Catalog.builtInCMapCache` into a `Map`, instead of an Object, to ensure that it's correctly reset (PR 8064 follow-up)

Snuffleupagus:
*Fixes a regression from PR #8064 (split off from PR #9931 since it probably deserves more visibility).* 

With the `builtInCMapCache` being a simple Object, it unfortunately means that the `Catalog.cleanup` method isn't resetting it as intended.
By just replacing the `builtInCMapCache` with an empty Object, existing references to it will not actually be updated. The result is that e.g. `Page` instances still keeps references to, what should have been removed, CMap data.

To fix these problems, the `builtInCMapCache` is converted into a `Map` instead (since it can be easily reset).

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -132,16 +132,15 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
     this.pdfFunctionFactory = pdfFunctionFactory;
 
     this.fetchBuiltInCMap = (name) => {
-      var cachedCMap = this.builtInCMapCache[name];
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good find; thanks!

Snuffleupagus:
@timvandermeij Thank you for landing this; and of course the other PRs yesterday :-)

PR #9936, opened 2018-07-29
Validate the Preferences when fetching them from storage

Snuffleupagus:
- Stop re-loading the Preferences in `PDFViewerApplication.open`, and remove the `BasePreferences.reload` method

   Given that the various Preferences are currently, and have been for quite some time, only used when initializing `PDFViewerApplication` re-loading them when a new PDF file is opened in the viewer is essentially a no-op.

   Furthermore, with the only usage of `BasePreferences.reload` now gone, the value of that method seems questionable at best. In the event that the functionality is actually needed again, similar to the `ViewHistory`, it'd probably make more sense to simply replace `PDFViewerApplication.preferences` with a new `BasePreferences` instance instead (using e.g. `DefaultExternalServices.createPreferences`).

 - Validate the Preferences when fetching them from storage

   When updating Preferences using the `set` method, the input is carefully validated. However, no validation is (currently) done when a `BasePreferences` instance is created, which probably isn't that great. Hence this patch that simply ignores, to not unnecessarily break loading of the viewer itself, any invalid Preferences.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -635,8 +635,6 @@ let PDFViewerApplication = {
     if (this.pdfLoadingTask) {
       // We need to destroy already opened document.
       return this.close().then(() => {
-        // Reload the preferences if a document was previously opened.
-        this.preferences.reload();
         // ... and repeat the open() call.
         return this.open(file, args);
       });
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good. Thank you for improving this!

Issue #9521, opened 2018-02-28
Offically release version 2.0


Snuffleupagus:
Before landing any larger (pending) PRs, it'd be a good idea to officially release version `2.0.x` of PDF.js *first* to reduce the regression risk in the initial release based on the `2.0` branch.

**Remaining TODOs:**
 - [x] Merge PR #9520.
 - [x] Merge PR #9493.

timvandermeij:
Closing since this is now complete.

Snuffleupagus:
@yurydelendik Since I'm not really familiar with all the details of https://github.com/mozilla/pdf.js/wiki/Release-Process, could you please help create a new PDF.js pre-release for version `2.0.419` i.e. at commit c33bf800cc87941cc681c3c54279ab1b9745650d?

Rob--W:
I edited the first post to include a reference to #9528. The documentation should be updated before the official 2.0 release.

timvandermeij:
I have also updated the project board at https://github.com/mozilla/pdf.js/projects/5 with all changes that we need to make.

prohtex:
Anything I can do to help with this?

Rob--W:
There are no blocking issues any more in the code (the JSFiddle examples are hosted elsewhere). Can we release 2.0 now? Anyone volunteering to write a migration guide for PDF.js 1.x code to 2.x?

wojtekmaj:
Regarding "Audit the existing CSS, and remove no longer needed (prefixed) rules", have you considered removing them all completely and use autoprefixer to target specific browsers as needed? Auditing that seems like a lot of work that could be automated.

timvandermeij:
I don't think I've seen that come by before, but it sounds like a good idea. However, I'm a bit worried about cases such as #6685. Will those be covered by such a module too?

wojtekmaj:
Actually, yes it will! :D 

![obraz](https://user-images.githubusercontent.com/5426427/38277361-6e168a00-3798-11e8-8525-96138dc06b12.png)

timvandermeij:
In that case, if anyone is willing to give Autoprefixer integration for PDF.js a shot, then please feel free to do so so we can test it. It might even fix some open issues given the comment above, which would be great (aside from the reduced maintenance work).

wojtekmaj:
Someone was willing :) https://github.com/mozilla/pdf.js/pull/9629
[...]

PR #9944, opened 2018-07-31
Add support for `async`/`await`, using Babel, and convert some (applicable) code to make use of it

Snuffleupagus:
*A good introduction to `async`/`await`, explaining the differences to `Promises`, can be found in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*

For proof-of-concept, this patch converts a couple of `Promise` returning methods to use `async` instead.
Please note that the `generic` build, based on this patch, has been successfully testing in IE11 (i.e. the viewer loads and nothing is obviously broken).

Being able to use modern JavaScript features like `async`/`await` is a huge plus, but there's one (obvious) side-effect: The size of the built files will increase slightly (unless `SKIP_BABEL == true`). That's unavoidable, but seems like a small price to pay in the grand scheme of things.

Finally, note that the `chromium` build target was changed to no longer skip Babel translation, since the Chrome extension still supports version `49` of the browser (where native `async` support isn't available).

*Much more manageable diff with https://github.com/mozilla/pdf.js/pull/9944/files?w=1*

### DIFF
--- a/web/chromecom.js
+++ b/web/chromecom.js
@@ -300,7 +300,7 @@ function setReferer(url, callback) {
 let storageArea = chrome.storage.sync || chrome.storage.local;
 
 class ChromePreferences extends BasePreferences {
-  _writeToStorage(prefObj) {
+  async _writeToStorage(prefObj) {
     return new Promise((resolve) => {
       if (prefObj === this.defaults) {
         let keysToRemove = Object.keys(this.defaults);
@@ -317,7 +317,7 @@ class ChromePreferences extends BasePreferences {
     });
   }
 
-  _readFromStorage(prefObj) {
+  async _readFromStorage(prefObj) {
     return new Promise((resolve) => {
       let getPreferences = (defaultPrefs) => {
         if (chrome.runtime.lastError) {
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,6 @@
 {
   "parserOptions": {
-    "ecmaVersion": 6,
+    "ecmaVersion": 8,
     "sourceType": "module",
   },
 
--- a/external/builder/preprocessor2.js
+++ b/external/builder/preprocessor2.js
@@ -288,6 +288,7 @@ function preprocessPDFJSCode(ctx, code) {
     },
   };
   var parseOptions = {
+    ecmaVersion: 8,
     locations: true,
     sourceFile: ctx.sourceFile,
     sourceType: 'module',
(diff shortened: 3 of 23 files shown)


web/chromecom.js
```diff
@@ -300,7 +300,7 @@ function setReferer(url, callback) {
 let storageArea = chrome.storage.sync || chrome.storage.local;
 
 class ChromePreferences extends BasePreferences {
-  _writeToStorage(prefObj) {
+  async _writeToStorage(prefObj) {
```
timvandermeij: Now that this is `async`, we should be able to remove the `return new Promise` code below. This also applies to the change just below and the changes in `web/firefoxcom.js`.

timvandermeij (commented):
I really like how this works and how it makes the code easier. Especially getting rid of the `Promise.resolve().then` construction makes it easier to reason about the code. I'm personally in favor of starting to use `async`/`await`, especially if we can make it work in the supported browsers. @brendandahl @yurydelendik Thoughts?

web/chromecom.js
Snuffleupagus: > Now that this is async, we should be able to remove the return new Promise code a line below.

Nope, that won't work :-)
Please note that, both here and in the other file you refer to, the `Promise`s are resolved from callback functions and that doesn't really work with `async`/`await`. (Maybe these shouldn't be marked as `async`, but that'd look slightly strange since the `BasePreferences` ones are.)

---

> I really like how this works and how it makes the code easier. 

It can (greatly) simplify certain code, as long as you keep in mind that not all `Promise` usage can/should be replaced with `async`/`await`; it really depends on the situation!
Keep in mind that `await` will *suspend* the following code, whereas a `Promise` allows later code to run; [these examples](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Examples) are quite good at highlighting the differences in behaviour.

Snuffleupagus:
Locally on Windows, I get the results as below (all sizes in bytes) for the `build/mozcentral` respectively `build/generic` directories for the corresponding builds:

|                    | `gulp mozcentral` | `gulp generic`
|-------------|-------------------|---------------
| **master**  | 3 442 404            | 10 831 041
| **patch**    | 3 440 333            | 11 039 049

The size of the `mozcentral` build is reduced *slightly*, which is great!
The size of the `generic` build does increase, but that's hopefully acceptable given the value that `async`/`await` support brings.

timvandermeij:
I'm fine with that. There is also https://github.com/mozilla/pdf.js/issues/9521#issuecomment-380981673. I'm glad that Babel seems to handle IE11 for us.

brendandahl:
I'm good with this, I would like to roll it out in chunks though. Can we open a PR with just the first commit? I'll review then we can land and start adding more pieces after a little time to make sure everything is going as expected.

PR #9948, opened 2018-08-02
Add (basic) unit-tests for the non-global `URL` constructor (PR 9868 follow-up)

Snuffleupagus:
This should really have been included in PR #9868, since it will help ensure that the `URL` constructor is correctly imported/exported by `src/shared/util.js`.


timvandermeij (approved):

timvandermeij:
Thank you for the follow-up!

PR #9954, opened 2018-08-03
Stop warning for non-Name /Filter entries in the `PDFImage` constructor (PR 9897 follow-up)

Snuffleupagus:
Fixes a stupid oversight on my part, since /Filter may (obviously) contain an Array, which resulted in unnecessary console warning spam in perfectly valid PDF files.
Note that it still makes sense to check that /Filter is actually a Name, before attempting to access its `name` property, but the warning should definitely be removed.

*Much smaller diff with https://github.com/mozilla/pdf.js/pull/9954/files?w=1*


timvandermeij (approved):

timvandermeij:
Thanks!

PR #9956, opened 2018-08-03
Allow loaded progress of 0 in unit tests.

### LINKED ISSUE #9947: Travis CI is suddenly failing on a unit test
From the logs:

```
Failures:
1) api getDocument creates pdf doc from URL
  Message:
    Expected false to equal true.
  Stack:
    Error: Expected false to equal true.
        at <Jasmine>
        at /home/travis/build/mozilla/pdf.js/build/lib/test/unit/api_spec.js:80:52
        at process._tickCallback (internal/process/next_tick.js:68:7)
```

This is strange because all checks were green when we merged the pull requests...

brendandahl:
Fixes #9947. The progress event has loaded = 0, total =105779.

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -80,7 +80,7 @@ describe('api', function() {
         loadingTask.promise
       ];
       Promise.all(promises).then(function (data) {
-        expect((data[0].loaded / data[0].total) > 0).toEqual(true);
+        expect((data[0].loaded / data[0].total) >= 0).toEqual(true);
         expect(data[1] instanceof PDFDocumentProxy).toEqual(true);
         expect(loadingTask).toEqual(data[1].loadingTask);
         loadingTask.destroy().then(done);


timvandermeij:
I'm wondering if it shouldn't even be `== 0`. If I read the code correctly, and please correct me if I'm wrong, then this only fetches the file (the `total` field), but does not actually read it (the `loaded` field), hence I think it should always be zero. Moreover, do you have an idea why it fails all of a sudden and not before?

brendandahl:
It appears with stream loading, onReceiveData is called with a chunk that is less than the chunk size which means no chunks are resolved and that screws up the calculation of how much of the file is loaded.  loaded = `this.stream.numChunksLoaded * this.chunkSize` 

I guess we could track how many bytes are loaded or just be fine with loaded being 0.

test/unit/api_spec.js
```diff
@@ -80,7 +80,7 @@ describe('api', function() {
         loadingTask.promise
       ];
       Promise.all(promises).then(function (data) {
-        expect((data[0].loaded / data[0].total) > 0).toEqual(true);
+        expect((data[0].loaded / data[0].total) >= 0).toEqual(true);
```
timvandermeij: Would the following work too? Doing it like this would assert that the file is loaded (i.e., the `total` attribute is nonzero) while also asserting and explaining why the `loaded` attribute is zero.

```javascript
// The `loaded` attribute represents the number of fetched full chunks. For stream
// loading, the received data is less than the size of a full chunk, so it is always zero.
expect(data[0].loaded).toEqual(0);
expect(data[0].total >= 0).toEqual(true);
```

Snuffleupagus:
> If I read the code correctly, and please correct me if I'm wrong, then this only fetches the file (the total field), but does not actually load a page of it (the loaded field). If that is not the correct way to interpret loaded and total, then does a value of zero make sense for this test?

This part of the unit-test is simply exercising the following callback function: https://github.com/mozilla/pdf.js/blob/f19ee127a31e794c558168025fc1cf027f1312e9/src/display/api.js#L437-L442, which forms the basis for the progress bar in the default viewer.

That unit-test is loading an *entire* file, which is want the `loadingTask.promise` waits for, so the values returned through the `onProgress` callback is simply the amount of the PDF data that was loaded thus far.
The purpose of the test was to ensure that at least *one* `onProgress` event is fired when a document is loaded; the actual values are probably less relevant here.

test/unit/api_spec.js
Snuffleupagus: It's *not* expected that `loaded` is ever zero normally, it's probably just a timing related artifact on e.g. newer Node.js versions.

timvandermeij: Thank you. I guess I'm just overthinking this.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #9958, opened 2018-08-03
Always fallback to system font on font failure.

### LINKED ISSUE #9949: Problem displaying PDF file with embedded fonts
I am working on the PDF library [printpdf](https://github.com/fschutt/printpdf) and noticed that Firefox doesn't display the fonts example correctly.

This could be a bug in pdf.js or a bug in my library. However, all other PDF viewers I've tested so far display the file correctly:

Evince:

image

Google Chrome:

image

Adobe Reader:

image

Foxit Reader:

image

xpdf:

image

Sumatra PDF:

image

... but Firefox:

image

Here is the corrupt PDF file: [test_fonts.pdf](https://github.com/mozilla/pdf.js/files/2254027/test_fonts.pdf)

The code that generated the PDF file is here: https://github.com/fschutt/printpdf/blob/master/examples/fonts.rs - if you want to reproduce this from source, clone the repository and run `cargo run --example fonts`.

I do get an error in the console, but it's very non-descriptive:

```
util.js:29 Warning: Error during font loading: Cannot read property 'length' of undefined
```

Configuration:
- Web browser and its version: Both Chrome 67.0.3396.99 and Firefox 61.0.1 (64-Bit)
- PDF.js version: master @ 716acf63d489bad06c9453f81f155546e4d77912
- Is a browser extension: No

1. Set up the build system as in the README
2. Open the `test_fonts.pdf` file

I've tried to debug this error. Again, it could be that something is wrong with my code, but I don't know Javascript too well to debug the problem. So far my notes were:

```
- message_handler.js:123 - "action[0].call(action[1], data.data);": action[1] is undefined for whatever reason, data.data is null
- worker.js:685 - "RenderPageRequest" - I think this is where the crash happens
- worker.js:686 - "page" variable seems to have an empty font cache - where are the fonts loaded?
- worker.js:686: It seems that page.getOperatorList is not executed correctly
```

The example itself embeds a font + a Unicode dictionary. The respective code for doing so is [here](https://github.com/fschutt/printpdf/blob/b362afa6caf735e566b52aba0d2b98dbdb0738c5/src/types/plugins/graphics/two_dimensional/font.rs#L157-L350), maybe a PDF expert can tell me what's wrong with this. Adobe Acrobat doesn't throw any errors aside from complaining about layers - the PDF validates for PDF/X3:2002 (except that it uses layers).

I could debug this further, but I find the architecture of pdf.js very complex and I am not very good in debugging JavaScript, so I wanted to ask if anyone could help on why this example doesn't work. Thanks in advance.

brendandahl:
The font in the PDF is marked as a CIDFontType0, but the font file is
actually a true type font. To fully address this issue we should really
peek into the font file and try to determine what it is. However, this
is the first case of this issue, so I think this solution is acceptable for
now.

Fixes #9949

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -621,9 +621,6 @@ var Font = (function FontClosure() {
           throw new FormatError(`Font ${type} is not supported`);
       }
     } catch (e) {
-      if (!(e instanceof FormatError)) {
-        throw e;
-      }
       warn(e);
       this.fallbackToSystemFont();
       return;
--- a/test/pdfs/.gitignore
+++ b/test/pdfs/.gitignore
@@ -88,6 +88,7 @@
 !bug1068432.pdf
 !bug1146106.pdf
 !bug1252420.pdf
+!issue9949.pdf
 !bug1308536.pdf
 !bug1337429.pdf
 !issue5564_reduced.pdf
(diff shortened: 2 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

PR #9960, opened 2018-08-04
Fail when MD5 of test files fails on bots.

### LINKED ISSUE #8042: Strict mode for testing on bots
It'd be nice if we had --strict option for testing that would fail the tests if an md5 fails. A couple times now while setting up the new bots I've had md5 checksums fail because of an issue and it really should kill the whole thing an not show success.

brendandahl:
It's easy to miss MD5 warnings on the bots, now it will fail.

Fixes #8042.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -356,7 +356,7 @@ function getTempFile(prefix, suffix) {
   return path;
 }
 
-function createTestSource(testsName) {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Nice addition!

PR #9961, opened 2018-08-04
Parse the font file to determine the correct type/subtype, rather than relying on the (often incorrect) data in the font dictionary

Snuffleupagus:
The current font type/subtype detection code is quite inconsistent/unwieldy. In some cases it will simply assume that the font dictionary is correct, in others it will somewhat "arbitrarily" check the actual font file (more of these cases have been added over the years to fix specific bugs).

As is evident from e.g. issue #9949, the font type/subtype detection code is continuing to cause issues. In an attempt to get rid of these hacks once and for all, this patch instead re-factors the type/subtype detection to *always* parse the font file.

Please note that, as far as I can tell, we still appear to need to rely on the composite font detection based on the font dictionary. However, even if the composite/non-composite detection would get it wrong, that shouldn't really matter too much given that there's basically only two different code-paths (for "TrueType-like" vs "Type1-like" fonts).

This PR provides a *complete* fix for issue #9949, in addition to the generally useful workaround implemented by PR #9958. Furthermore, this PR also ought to prevent any future font type/subtype detection bugs too.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -680,7 +680,8 @@ var Font = (function FontClosure() {
 
   function isTrueTypeFile(file) {
     var header = file.peekBytes(4);
-    return readUint32(header, 0) === 0x00010000;
+    return (readUint32(header, 0) === 0x00010000 ||
+            bytesToString(header) === 'true');
   }
 
   function isTrueTypeCollectionFile(file) {
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -550,36 +550,14 @@ var Font = (function FontClosure() {
       return;
     }
 
-    // Some fonts might use wrong font types for Type1C or CIDFontType0C
-    if (subtype === 'Type1C') {
-      if (type !== 'Type1' && type !== 'MMType1') {
-        // Some TrueType fonts by mistake claim Type1C
-        if (isTrueTypeFile(file)) {
-          subtype = 'TrueType';
-        } else {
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Since I've not run any tests locally, I suppose that this could all fail spectacularly :-)

/botio test

Snuffleupagus:
The failures should hopefully all be caused by a pre-existing existing, since the `isTrueTypeFile` helper function wasn't able to detect certain Mac fonts.

/botio test

timvandermeij (approved):

timvandermeij:
Thank you for improving this; much more readable this way and I verified that it fixes the issue too.

Snuffleupagus:
> Another possibility is doing something like tx and peek into the file:

But... that's what this PR does :-)

The newly added [`getFontFileType`](https://github.com/mozilla/pdf.js/blob/c8ee63319d0b0f1c04d7bfc23b846e5322ee34f7/src/core/fonts.js#L702) function will "peek" into the font files themselves to determine their *actual* types, by making use of these existing helper functions: https://github.com/mozilla/pdf.js/blob/c8ee63319d0b0f1c04d7bfc23b846e5322ee34f7/src/core/fonts.js#L656-L684

brendandahl:
Hah, okay, I was lazy and didn't look into what all the is<SomeType> method did. I assumed we were just looking at the font dictionary.

PR #9962, opened 2018-08-05
Throw MissingPDFException when file not found

### LINKED ISSUE #9791: Enable the missing PDF unit test for the API in NodeJS
The objective is to remove https://github.com/mozilla/pdf.js/blob/master/test/unit/api_spec.js#L166-L170 by fixing up `src/display/node_stream.js` to throw `MissingPDFException` when a PDF file is not found. Right now, this file is not handling errors in a consistent or even correct way compared with `src/display/network.js` and `src/display/fetch_stream.js`. Using the `createResponseStatusError` utility function might not make much sense in a Node.js environment. However, at the very least it seems that `MissingPDFException` or `UnknownErrorException` when we cannot tell that a PDF file is missing should be thrown manually. As is, the API (i.e., `getDocument`) is not returning the expected errors when loading fails in Node.js environments (as evident from this pending API unit test).

bhavya17037:
This does not work as expected, as fs.lstat() in node_steam_spec.js does not work correctly, as it does not give 404 response, even when file path is non existent, which may be part of a previous issue. 
Thus the unit test does not pass when a wrong URL is given.

output

Fixes #9791.

### DIFF
--- a/src/display/node_stream.js
+++ b/src/display/node_stream.js
@@ -20,7 +20,7 @@ let https = __non_webpack_require__('https');
 let url = __non_webpack_require__('url');
 
 import {
-  AbortException, assert, createPromiseCapability
+  AbortException, assert, createPromiseCapability, MissingPDFException
 } from '../shared/util';
 import {
   extractFilenameFromHeader, validateRangeRequestCapabilities
@@ -300,6 +300,10 @@ class PDFNodeStreamFullReader extends BaseFullReader {
     super(stream);
 
     let handleResponse = (response) => {
+      let status = response.status;
+      if (status === 404 || status === 0 && /^file:/.test(url)) {
+        throw new MissingPDFException('Missing PDF "' + url + '".');
+      }
       this._headersCapability.resolve();
       this._setReadableStream(response);
 
--- a/test/unit/node_stream_spec.js
+++ b/test/unit/node_stream_spec.js
@@ -30,7 +30,7 @@ describe('node_stream', function() {
   let server = null;
   let port = null;
   let pdf = url.parse(encodeURI('file://' + path.join(process.cwd(),
-                      './test/pdfs/tracemonkey.pdf'))).href;
+                      './test/pdfs/tracemonkeyWRONG_PATH.pdf'))).href;
   let pdfLength = 1016315;
 
   beforeAll((done) => {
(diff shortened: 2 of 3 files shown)


bhavya17037:
@timvandermeij 
Hi, I created this pull request as per your advice :)

src/display/node_stream.js
```diff
@@ -300,6 +300,10 @@ class PDFNodeStreamFullReader extends BaseFullReader {
     super(stream);
 
     let handleResponse = (response) => {
+      let status = response.status;
```
timvandermeij: Change this to:

```javascript
const status = response.status;
```

src/display/node_stream.js
```diff
@@ -300,6 +300,10 @@ class PDFNodeStreamFullReader extends BaseFullReader {
...
+      if (status === 404 || status === 0 && /^file:/.test(url)) {
+        throw new MissingPDFException('Missing PDF "' + url + '".');
```
timvandermeij: Change this to:

```javascript
throw new MissingPDFException(`Missing PDF "${url}".`);
```

test/unit/node_stream_spec.js
```diff
@@ -30,7 +30,7 @@ describe('node_stream', function() {
   let server = null;
   let port = null;
   let pdf = url.parse(encodeURI('file://' + path.join(process.cwd(),
-                      './test/pdfs/tracemonkey.pdf'))).href;
+                      './test/pdfs/tracemonkeyWRONG_PATH.pdf'))).href;
```
timvandermeij: You can revert this change. Looking at the original issue you're solving, you can just remove https://github.com/mozilla/pdf.js/blob/master/test/unit/api_spec.js#L167-L171 and the unit tests should still pass if your change works.

timvandermeij (commented):
You can revert the changes in `package-lock.json` since they're not relevant for this issue. If you're doing this in a new commit, together with the changes below, make sure to squash the commits into one commit afterwards (refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for how to do that easily). Thank you!

timvandermeij:
It looks like this is a bit harder than first anticipated because there are more edge cases to consider. Even though your patch was a good start, another contributor found these edge cases and resolved the issue in #9964. Therefore I'm closing this pull request, but thank you for helping out with this and feel free to take up another issue in the issue tracker!

PR #9964, opened 2018-08-05
Attempt to throw `MissingPDFException` when applicable in `node_stream.js` (issue 9791)

### LINKED ISSUE #9791: Enable the missing PDF unit test for the API in NodeJS

Snuffleupagus:
While I'm not very familiar with Node.js, this patch should (hopefully) cover the *multiple* ways in which files can be loaded in Node.js environments; while also taking into account e.g. the differences between the `XMLHttpRequest` and the `{http, https}.request` methods.
Furthermore, the `creates pdf doc from non-existent URL` API unit-test now passes in Node.js/Travis.

Fixes #9791.

### DIFF
--- a/src/display/node_stream.js
+++ b/src/display/node_stream.js
@@ -20,7 +20,7 @@ let https = __non_webpack_require__('https');
 let url = __non_webpack_require__('url');
 
 import {
-  AbortException, assert, createPromiseCapability
+  AbortException, assert, createPromiseCapability, MissingPDFException
 } from '../shared/util';
 import {
   extractFilenameFromHeader, validateRangeRequestCapabilities
@@ -300,6 +300,12 @@ class PDFNodeStreamFullReader extends BaseFullReader {
     super(stream);
 
     let handleResponse = (response) => {
+      if (response.statusCode === 404) {
+        const error = new MissingPDFException(`Missing PDF "${this._url}".`);
+        this._storedError = error;
+        this._headersCapability.reject(error);
+        return;
+      }
       this._headersCapability.resolve();
       this._setReadableStream(response);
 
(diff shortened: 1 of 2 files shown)


src/display/node_stream.js
```diff
@@ -359,17 +365,23 @@ class PDFNodeStreamRangeReader extends BaseRangeReader {
     }
     this._httpHeaders['Range'] = `bytes=${start}-${end - 1}`;
 
+    let handleResponse = (response) => {
+      if (response.statusCode === 404) {
+        const error = new MissingPDFException(`Missing PDF "${this._url}".`);
+        this._storedError = error;
```
timvandermeij: I think we should `return` after this too to avoid setting the readable stream to a failed response, just like above.

Snuffleupagus: Yes, that seems reasonable; fixed now.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #9975, opened 2018-08-11
Re-factor `destinations`/`getDestination` to reduce unnecessary duplication, and reject non-string inputs

Snuffleupagus:
- Re-factor `destinations`/`getDestination`, in the `Catalog`, to reduce unnecessary duplication

   Currently, these two methods contain the same boilerplate code for getting the /Dests data.

- Reject `getDestination`, in the API, for non-string inputs

  Note how e.g. the `getPage` method does basic validation of the input.


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #9977, opened 2018-08-13
Add support for `async`/`await` using Babel

Snuffleupagus:
For proof-of-concept, this patch converts a couple of `Promise` returning methods to use `async` instead.
Please note that the `generic` build, based on this patch, has been successfully testing in IE11 (i.e. the viewer loads and nothing is obviously broken).

Being able to use modern JavaScript features like `async`/`await` is a huge plus, but there's one (obvious) side-effect: The size of the built files will increase slightly (unless `SKIP_BABEL == true`). That's unavoidable, but seems like a small price to pay in the grand scheme of things.

Finally, note that the `chromium` build target was changed to no longer skip Babel translation, since the Chrome extension still supports version `49` of the browser (where native `async` support isn't available).

Split off from PR #9944 as requested by @brendandahl.

*Much more manageable diff with https://github.com/mozilla/pdf.js/pull/9977/files?w=1*

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,6 @@
 {
   "parserOptions": {
-    "ecmaVersion": 6,
+    "ecmaVersion": 8,
     "sourceType": "module",
   },
 
--- a/external/builder/preprocessor2.js
+++ b/external/builder/preprocessor2.js
@@ -288,6 +288,7 @@ function preprocessPDFJSCode(ctx, code) {
     },
   };
   var parseOptions = {
+    ecmaVersion: 8,
     locations: true,
     sourceFile: ctx.sourceFile,
     sourceType: 'module',
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -177,7 +177,14 @@ function createWebpackConfig(defines, output) {
           exclude: /src[\\\/]core[\\\/](glyphlist|unicode)/,
           options: {
             presets: skipBabel ? undefined : ['env'],
-            plugins: ['transform-es2015-modules-commonjs'],
+            plugins: [
+              'transform-es2015-modules-commonjs',
+              ['transform-runtime', {
+                'helpers': false,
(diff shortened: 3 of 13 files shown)


Snuffleupagus:
The one Windows failure is strange, since the commit included here is *exactly* the same as the first one in PR #9944 (where all tests ran successfully). One possible explanation is that Firefox was updated on the bot (which might not show up in the logs, since the bot uses a beta version); another explanation is that it's simply an intermittent.

/botio-windows test

Snuffleupagus:
I don't understand what to make of these intermittent failures (on Windows), since all tests seem to pass consistently in PR #9944 and the commit hashes are *identical* :confused:

brendandahl:
Doesn't look like it's just your changes, I got it to fail over in https://github.com/mozilla/pdf.js/pull/9567#issuecomment-412937390

src/display/api.js
```diff
@@ -2050,13 +2050,11 @@ var WorkerTransport = (function WorkerTransportClosure() {
...
 
-      messageHandler.on('FetchBuiltInCMap', function (data) {
+      messageHandler.on('FetchBuiltInCMap', function(data) {
         if (this.destroyed) {
           return Promise.reject(new Error('Worker was destroyed'));
         }
-        return this.CMapReaderFactory.fetch({
-          name: data.name,
-        });
+        return this.CMapReaderFactory.fetch(data);
```
brendandahl: Why this change?

brendandahl (changes requested):

src/display/api.js
Snuffleupagus: Sorry, this was unrelated the main objective of the patch; it's been reverted now.

Snuffleupagus: Sorry, this was unrelated to the main objective of the patch; it's been reverted now.

PR #9980, opened 2018-08-16
Ensure that `Uint8ClampedArray` is used for image data transfered by `getTransfers` (PR 9802 follow-up)

Snuffleupagus:
One of the `QueueOptimizer` cases wasn't updated to use `Uint8ClampedArray`s, which leads to inconsistent image data on the API side (but no actual rendering bugs, as far as I can tell).
To prevent future errors, a non-production/test-only `assert` was added to ensure that the relevant image data only uses `Uint8ClampedArray`s.

### DIFF
--- a/src/core/operator_list.js
+++ b/src/core/operator_list.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import { ImageKind, OPS } from '../shared/util';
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice find!

PR #9982, opened 2018-08-17
Use the Font Loading API in `MOZCENTRAL` builds, and `GENERIC` builds for Firefox version 63 and above (issue 9945, bug 1088663)

### LINKED ISSUE #9945: Use JS font loading API on Firefox
After [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1473742) lands we should be able to use the font loading API on Firefox. Unfortunately, there's not really a way to feature detect, so we'll have to check the Firefox version number.

Snuffleupagus:
*I was feeling bored, so this is an attempt at fixing issue #9945.*

This PR refactor/modernize/clean-up the font loading code, and then adds `MOZCENTRAL` support for the Font Loading API on top. /cc @brendandahl 

*Note:* For easier reviewing, I'd suggest looking at one commit at a time and also using the `?w=1` URL flag to reduce the size of the larger diffs.

Fixes #9945.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -589,8 +589,6 @@ var Font = (function FontClosure() {
     this.toUnicode = properties.toUnicode;
     this.encoding = properties.baseEncoding;
     this.seacMap = properties.seacMap;
-
-    this.loading = true;
   }
 
   Font.getFontID = (function () {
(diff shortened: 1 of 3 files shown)


brendandahl (changes requested):

mark-i-m:
Any progress on this? It would be nice to have the printing issues fixed. Thanks!

brendandahl:
I want to hold off on this until the end of the week so #9340 has more time to bake in Firefox Nightly. It was a pretty large change and it will be easier to see if any regression come from it or this change if they are spread out a bit.

PR #9987, opened 2018-08-19
[api-minor] Remove the obsolete `createBlob` helper function

Snuffleupagus:
At this point in time, all supported browsers have native support for `Blob`; please see https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob#Browser_compatibility.
Furthermore, note how the helper function was throwing an error if `Blob` isn't available anyway.


timvandermeij (approved):

timvandermeij:
Nice find!

PR #9988, opened 2018-08-19
Remove the unused `PDFDocument.mainXRefEntriesOffset` method

Snuffleupagus:
Not only is this method completely unused *now*, looking through the history of the code it never appears to have been used for anything either.
Years ago `mainXRefEntriesOffset` was included when creating `XRef` instances, however it wasn't actually used for anything (the parameter was never checked, nor assigned to a property on `XRef`).

If this method ever becomes useful (again) it's easy enough to restore it thanks to version control, but including dead code in the builds just seems wasteful.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #9990, opened 2018-08-19
Convert the `Catalog` class, in `src/core/obj.js`, to ES6 syntax

timvandermeij:
This patch is much easier to review with the `?w=1` flag: https://github.com/mozilla/pdf.js/pull/9990/files?w=1.

@Snuffleupagus Would you, time permitting of course, be willing to review this?

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -32,13 +32,14 @@ function fetchDestination(dest) {
   return isDict(dest) ? dest.get('D') : dest;
 }
 
-var Catalog = (function CatalogClosure() {
-  function Catalog(pdfManager, xref) {
+class Catalog {
+  constructor(pdfManager, xref) {
     this.pdfManager = pdfManager;
     this.xref = xref;
+
     this.catDict = xref.getCatalogObj();
     if (!isDict(this.catDict)) {
-      throw new FormatError('catalog object is not a dictionary');
+      throw new FormatError('Catalog object is not a dictionary.');
     }
 
     this.fontCache = new RefSetCache();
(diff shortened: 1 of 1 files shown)


src/core/obj.js
```diff
@@ -32,578 +32,600 @@ function fetchDestination(dest) {
...
-          // was reasonable enough to stick with the XML default charset,
-          // which is UTF-8.
-          try {
-            metadata = stringToUTF8String(bytesToString(stream.getBytes()));
-          } catch (e) {
-            if (e instanceof MissingDataException) {
-              throw e;
-            }
-            info('Skipping invalid metadata.');
+    const stream = this.xref.fetch(streamRef, !encryptMetadata);
```
Snuffleupagus: You're obviously just touching old code here, but this looks a bit convoluted. How about, in a separate commit, updating this to be a bit easier to read?

```javascript
const suppressEncryption = !(this.xref.encrypt &&
                             this.xref.encrypt.encryptMetadata);
const stream = this.xref.fetch(streamRef, suppressEncryption);
```

src/core/obj.js
```diff
@@ -32,578 +32,600 @@ function fetchDestination(dest) {
...
-      var obj = this.catDict.getRaw('PageLabels');
-      if (!obj) {
-        return null;
-      }
-      var pageLabels = new Array(this.numPages);
-      var style = null;
-      var prefix = '';
+    const pageLabels = new Array(this.numPages);
+    let style = null;
+    let prefix = '';
```
Snuffleupagus: Nit: These can probably just go on *one* line instead, i.e. `let style = null, prefix = '';` (since that's how e.g. currentLabel/currentIndex below is already formatted in the old code).

src/core/obj.js
```diff
@@ -32,578 +32,600 @@ function fetchDestination(dest) {
...
-        });
-      }
+    let total = 0;
+    function next(ref) {
+      return pagesBeforeRef(ref).then(function(args) {
+        if (!args) {
+          return total;
+        }
+        const count = args[0];
+        const parentRef = args[1];
```
Snuffleupagus: You should be able to use array destructuring instead, i.e. `const [count, parentRef] = args;`.

src/core/obj.js
```diff
@@ -712,41 +735,43 @@ var Catalog = (function CatalogClosure() {
...
             js = bytesToString(jsAction.getBytes());
           } else if (isString(jsAction)) {
             js = jsAction;
           }
 
           if (js) {
-            // Attempt to recover valid URLs from 'JS' entries with certain
-            // white-listed formats, e.g.
+            // Attempt to recover valid URLs from `JS` entries with certain
+            // white-listed formats, e.g.:
```
Snuffleupagus: Nit: It might be a personal preference, but I find that a colon after a period looks really ugly; can we please not do that :-)

src/core/obj.js
```diff
@@ -32,578 +32,600 @@ function fetchDestination(dest) {
...
-      return pageLabels;
-    },
+    }
+    return shadow(this, 'pageMode', pageMode);
+  }
+
+  get attachments() {
+    let attachments = null, nameTreeRef;
+    const obj = this.catDict.get('Names');
+    if (obj) {
```
Snuffleupagus: This is another potential for clean-up (again in a separate commit).
It looks like this code was basically copy-pasted from an older version of the `destinations` getter, but it could probably be changed to e.g.
```javascript
 get attachments() {
   const obj = this.catDict.get('Names');
   let attachments = null;

   if (obj && obj.has('EmbeddedFiles')) {
     const nameTree = new NameTree(obj.getRaw('EmbeddedFiles'), this.xref);
     const names = nameTree.getAll();
     for (let name in names) {
       const fs = new FileSpec(names[name], this.xref);
       if (!attachments) {
         attachments = Object.create(null);
       }
       attachments[stringToPDFString(name)] = fs.serializable;
     }
   }
   return shadow(this, 'attachments', attachments);
}
```

Snuffleupagus (approved):

timvandermeij:
Thank you for the review, @Snuffleupagus! The failure above is an intermittent one.

PR #9858, opened 2018-06-30
Additional Scroll/Spread mode clean-up (PR 9832 follow-up)

Snuffleupagus:
*Please refer to the individual commit messages for additional details.*

Since I had some free time yesterday, this is the additional clean-up mentioned in PR #9832.

PR #9991, opened 2018-08-20
Remove the deprecated ways, in `BaseViewer`, of setting the Scroll/Spread modes (PR 9858 follow-up)

Snuffleupagus:
Considering that a number of `[api-minor]` changes have landed since PR #9858, removing this code ought to be OK now (the less time these methods remain exposed, the better); implements https://github.com/mozilla/pdf.js/pull/9858#issuecomment-401730065.


timvandermeij (approved):

timvandermeij:
I agree, it's better to clean this up now. Thank you!

PR #10000, opened 2018-08-22
Remove the unused `defaultColor` property on `ColorSpace` instances

Snuffleupagus:
This property is not only completely unused now, it never actually appears to have been used. Even though the memory savings, from not initializing these extra typed arrays, won't be significant in the grand scheme of things it still seems completely unnecessary to keep allocating this data.

As far as I can tell, the main reason for the existence of `defaultColor` seem to be for documentation purposes. Hence the code is changed into comments instead, to keep the information around (but without the unnecessary allocations).

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -53,7 +53,7 @@ var ColorSpace = (function ColorSpaceClosure() {
     }
   }
 
-  // Constructor should define this.numComps, this.defaultColor, this.name
+  // Constructor should define this.numComps, this.name
   function ColorSpace() {
     unreachable('should not call ColorSpace constructor');
   }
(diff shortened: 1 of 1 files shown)


timvandermeij:
I think converting this to textual comments instead of code comments would be better, i.e., something like "// The default color is black.". It's not only more readable, but it also prevents people from throwing it away later since now it looks like unused code. Would you agree?

Snuffleupagus:
> I think converting this to textual comments instead of code comments would be better, i.e., something like `// The default color is black.`

That sounds really simple, until you start looking at either the [AlternateCS](https://github.com/mozilla/pdf.js/pull/10000/files#diff-b53da70ffcb20b79cde556d5176a903aR429) or [DeviceCmykCS](https://github.com/mozilla/pdf.js/pull/10000/files#diff-b53da70ffcb20b79cde556d5176a903aR730) cases :-(

> It's not only more readable,

In either of those cases, trying to explain with a "regular comment is going to be a lot less clear not to mention longer.
The easiest solution would of course be to simply throw out the comments completely, since they've never been used. Another, perhaps better, option could be to append something like `The default color is {insert typed array here}.` comments to existing JSDocs such as e.g. https://github.com/mozilla/pdf.js/blob/master/src/core/colorspace.js#L419-L424.

timvandermeij:
Looks better this way. Let's see how it goes.

/botio test

timvandermeij (approved):

timvandermeij:
Nice optimization!

PR #10007, opened 2018-08-24
Convert the code in `src/core/colorspace.js to use ES6 classes

Snuffleupagus:
- Reduces the amount of boilerplate code when defining the the sub-classes.

   Please note that a couple of the closures were kept, since it's not (yet) possible to include helper functions inside of `class`es.

 - Replace all occurences of `var` with `let`/`const` in `src/core/colorspace.js`

---

*Note:* For easier reviewing, I'd suggest looking at one commit at a time and also using the `?w=1` URL flag to reduce the size of particularly the first diff.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -18,194 +18,198 @@ import {
 } from '../shared/util';
 import { isDict, isName, isStream } from './primitives';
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you! Looks much easier this way.

PR #10012, opened 2018-08-26
Update translation ES-AR

VirginiaBalseiro:
Fixed an awkward translation and a typo.

### DIFF
--- a/l10n/es-AR/chrome.properties
+++ b/l10n/es-AR/chrome.properties
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Chrome notification bar messages and buttons
-unsupported_feature=Este documento PDF puede que no se muestre correctamente.
-unsupported_feature_forms=Este documento PDF cotiene formularios. Completar campos de un formulario no está soportado.
+unsupported_feature=Este documento PDF podría no mostrarse correctamente.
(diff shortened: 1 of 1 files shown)


timvandermeij:
While this looks good to me, unfortunately we cannot merge the patch. From https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md:

> Note that the translations for PDF.js in the `l10n` folder are synchronized with the Nightly channel of Mozilla Firefox. This means that we will only accept pull requests that add strings currently missing in the Nightly channel, but keep in mind that the changes will be overwritten when we synchronize again.
```

Therefore, any changes that we make here will be overwritten in the next (automatic) synchronization. Instead of fixing this here, please report the issue upstream to the `es-AR` localization team so they can fix it on their end. After that, the fix will automatically be pulled in here. You can do that either opening a ticket on Bugzilla with your suggested changes for the `es-AR` locale (see https://bugzilla.mozilla.org/buglist.cgi?field0-0-0=component;type0-0-0=regexp;value0-0-0=^es-AR%20/%20;field0-0-1=cf_locale;type0-0-1=regexp;value0-0-1=^es-AR%20/%20;resolution=---) or contact the maintainers via https://wiki.mozilla.org/L10n:Teams:es-AR so they can apply the fix. Thank you!

Issue #10002, opened 2018-08-22
Code splitting is not naming webpack chunk


cheryly279:
Is there a reason that you guys are not naming async loaded code? This relates to code here: https://github.com/mozilla/pdf.js/blob/master/src/display/api.js#L70, though there are probably other instances of this in the code.

When I bundle my app using webpack, there is an asset with a generic name like "0" where I would expect something like maybe "pdfjs_core". I traced it back to the line of code above. It can be fixed by adding a name as the 4th argument, based on this: https://webpack.js.org/api/module-methods/#require-ensure.

Is there some reason you did not include this?

timvandermeij:
Fixed by the pull request above.

PR #10015, opened 2018-08-27
Adding chunkname to async loaded code

cheryly279:
Improvement, see: https://github.com/mozilla/pdf.js/issues/10002

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -75,7 +75,7 @@ if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
         worker = __non_webpack_require__('./pdf.worker.js');
       }
       callback(worker.WorkerMessageHandler);
-    });
+    }, null, 'pdfjsWorker');
   }) : dynamicLoaderSupported ? (function (callback) {
     requirejs(['pdfjs-dist/build/pdf.worker'], function (worker) {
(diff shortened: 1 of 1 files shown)


src/display/api.js
```diff
@@ -75,7 +75,7 @@ if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
         worker = __non_webpack_require__('./pdf.worker.js');
       }
       callback(worker.WorkerMessageHandler);
-    });
+    }, null, 'pdfWorker');
```
Snuffleupagus: `pdfjsWorker` seems like a slightly better choice here, since it would be consistent with the name/alias [used in the bundles](https://github.com/mozilla/pdf.js/blob/436d2efa8a89edef3a05f70592e3a70fd0c7d11e/gulpfile.js#L288).

cheryly279: thanks, updated!

timvandermeij:
Looks good to me. Could you please squash the commits into one commit? Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for how to do that easily.

cheryly279:
Squashed.

timvandermeij (approved):

timvandermeij:
Thank you!

Issue #9953, opened 2018-08-03
PDF rendered incorrectly: white gap lines between neighbouring rectangles


ousia:
Attach (recommended) or Link to PDF file here:
https://gitlab.gnome.org/GNOME/evince/uploads/3593c890420275992e8bceb4bb5f3d88/frame-on-background.pdf

Configuration:
- Web browser and its version: Firefox 61.0.1
- Operating system and its version: Fedora 28
- PDF.js version: I don’t know.
- Is a browser extension: No

Steps to reproduce the problem:

The linked PDF file (created with OpenOffice) contains adjacent same-colored rectangles (see https://bz.apache.org/ooo/show_bug.cgi?id=127828 for further information).
When opened with PDF.js, thin white gap lines are visible between the rectangles.

PDF.js should display the [following output](https://bugs.freedesktop.org/attachment.cgi?id=140938), but it gives [this one](https://bugs.freedesktop.org/attachment.cgi?id=140939).

timvandermeij:
Most likely related to #3351.

ousia:
Subscribed to both  #3351 and  #3331.

Closing the issue myself.

Issue #9829, opened 2018-06-20
Pdf gets rendered with stripes


peirick:
Link to PDF file: [https://github.com/peirick/BugsInPdfJs/raw/master/BugWithStripes.pdf](https://github.com/peirick/BugsInPdfJs/raw/master/BugWithStripes.pdf)

Configuration:
- Web browser and its version: Firefox 60.0.2
- Operating system and its version: Windows 10
- PDF.js version: PDF.js: 2.0.402

Steps to reproduce the problem:
1. Open the pdf document

What is the expected behavior? (add screenshot)
In Chrome no stripes are shown:
![image](https://user-images.githubusercontent.com/15078132/41651336-9f03d864-7480-11e8-9aa4-b67b03bd0ffc.png)

What went wrong? (add screenshot)
In Firefox stripes are shown:
![image](https://user-images.githubusercontent.com/15078132/41651379-bd80e4a8-7480-11e8-875c-4c6df115c16e.png)

Snuffleupagus:
This is a duplicate of e.g. issue #3351.

Issue #9481, opened 2018-02-15
PDF appears with lines through several images


tonyjin:
Attach (recommended) or Link to PDF file here:
[Tabletop_report_digital.pdf](https://github.com/mozilla/pdf.js/files/1726090/Tabletop_report_digital.pdf)

Configuration:
- Web browser and its version: Chrome 64.0.3282.140, Firefox 59.0b7
- Operating system and its version: Mac OSX 10.12.6
- PDF.js version: Latest on pdfjs-dist (tested with v2.0.374 at time of ticket)
- Is a browser extension: No

Steps to reproduce the problem:
1. Visit http://mozilla.github.io/pdf.js/web/viewer.html?file=
2. Upload attached file

What is the expected behavior? (add screenshot)
Page 1:
![2018-02-14_16-48-13](https://user-images.githubusercontent.com/689351/36235667-dea57faa-11a6-11e8-9dfa-9e04bbf7f4f9.png)

Page 3:
![2018-02-14_16-48-51](https://user-images.githubusercontent.com/689351/36235685-f1824f68-11a6-11e8-8a60-ce5a5a3f4f00.png)

What went wrong? (add screenshot)
Page 1 (note horizontal lines):
![2018-02-14_16-49-17](https://user-images.githubusercontent.com/689351/36235695-013fccdc-11a7-11e8-8728-0fc642b91490.png)

Page 3:
![2018-02-14_16-49-45](https://user-images.githubusercontent.com/689351/36235711-11315e8a-11a7-11e8-8577-2f37b2d2de27.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): N/A, used viewer at http://mozilla.github.io/pdf.js/web/viewer.html?file=

Snuffleupagus:
Closing as duplicate of e.g. issue #3351.

Issue #9464, opened 2018-02-09
Horizontal stripes in images


fosple:
Attach (recommended) or Link to PDF file here:
[124368.pdf](https://github.com/mozilla/pdf.js/files/1710568/124368.pdf)
[109059.pdf](https://github.com/mozilla/pdf.js/files/1710569/109059.pdf)

Configuration:

Web browser and its version: Chrome 63.0.3239.132
Operating system and its version: MacOS 10.13.2
PDF.js version: pdfjs 1.9.426

Steps to reproduce the problem:
- Open PDF

What is the expected behavior? / What went wrong?

Stripes in image:

![124368_stripes](https://user-images.githubusercontent.com/4522947/36022960-24400f32-0d8b-11e8-93a9-6bf359877723.png)
![109059_stripes](https://user-images.githubusercontent.com/4522947/36023237-0ea2114c-0d8c-11e8-854e-023ecb5c3cd3.png)

Snuffleupagus:
Closing as duplicate of e.g. issue #3351.

PR #10016, opened 2018-08-27
Fix for white hairlines rendering in pdf

brianholleibm:
This fix will provide a solution to the following issues:

#3351
#4860
#3331
#9464 (Closed) (Duplicate)
#9481 (Closed) (Duplicate)
#9829 (Closed) (Duplicate)
#9953 (Closed) (Duplicate)

This solution takes into account the resolution of the viewer before painting images to the canvas and rounds the painting to the nearest full pixel. The problem in the issues above is caused by drawing to a canvas with a width/height that is more accurate than pixel width allows. This results in the drawing coloring part of a pixel, which causes a partial transparency. This transparency creates what appears to be a thin line between images. Rounding the images to the nearest pixel insures there will be no partially filled pixels, and therefore no thin lines.

Performance Tests
```
Run1:
browser | stat         | Count | Baseline(ms) | Current(ms) |  +/- |     %  | Result(P<.05)
------- | ------------ | ----- | ------------ | ----------- | ---- | ------ | -------------
Firefox | Overall      |    50 |        17427 |       16954 | -474 |  -2.72 |              
Firefox | Page Request |    50 |           11 |           9 |   -2 | -14.34 |              
Firefox | Rendering    |    50 |        17416 |       16944 | -472 |  -2.71 |   
```
```
Run2:
browser | stat         | Count | Baseline(ms) | Current(ms) |   +/- |     %  | Result(P<.05)
------- | ------------ | ----- | ------------ | ----------- | ----- | ------ | -------------
Firefox | Overall      |    50 |        18640 |       15158 | -3481 | -18.68 |        faster
Firefox | Page Request |    50 |           11 |          10 |    -1 |  -5.59 |              
Firefox | Rendering    |    50 |        18629 |       15148 | -3481 | -18.69 |        faster
```

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -1082,21 +1082,34 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       var ctx = this.ctx;
       var current = this.current;
       var x = current.x, y = current.y;
+      let xw, yh, origX, origY, pixelSize = this.getSinglePixelWidth();
       for (var i = 0, j = 0, ii = ops.length; i < ii; i++) {
         switch (ops[i] | 0) {
           case OPS.rectangle:
-            x = args[j++];
-            y = args[j++];
+            origX = args[j++];
+            origY = args[j++];
+            if (pixelSize) {
+              x = Math.floor(origX / pixelSize) * pixelSize;
+              y = Math.ceil(origY / pixelSize) * pixelSize;
+            } else {
+              x = origX;
+              y = origY;
+            }
             var width = args[j++];
             var height = args[j++];
             if (width === 0) {
-              width = this.getSinglePixelWidth();
+              width = pixelSize;
             }
             if (height === 0) {
-              height = this.getSinglePixelWidth();
+              height = pixelSize;
             }
-            var xw = x + width;
-            var yh = y + height;
+            if (pixelSize) {
+              xw = Math.ceil((origX + width) / pixelSize) * pixelSize;
+              yh = Math.ceil((origY + height) / pixelSize) * pixelSize;
+            } else {
+              xw = x + width;
+              yh = y + height;
+            }
             this.ctx.moveTo(x, y);
             this.ctx.lineTo(xw, y);
             this.ctx.lineTo(xw, yh);
@@ -1941,7 +1954,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       var ctx = this.ctx;
       // scale the image to the unit square
       ctx.scale(1 / w, -1 / h);
-
+      const pixelSize = this.getSinglePixelWidth();
+      h = Math.ceil(h / pixelSize) * pixelSize;
+      w = Math.ceil(w / pixelSize) * pixelSize;
       ctx.drawImage(domImage, 0, 0, domImage.width, domImage.height,
                     0, -h, w, h);
       if (this.imageLayer) {
@@ -2144,6 +2159,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
                                                   newWidth, newHeight);
         tmpCtx = tmpCanvas.context;
         tmpCtx.clearRect(0, 0, newWidth, newHeight);
+        const pixelSize = this.getSinglePixelWidth();
+        newHeight = Math.ceil(newHeight / pixelSize) * pixelSize;
+        newWidth = Math.ceil(newWidth / pixelSize) * pixelSize;
         tmpCtx.drawImage(imgToPaint, 0, 0, paintWidth, paintHeight,
                                      0, 0, newWidth, newHeight);
         imgToPaint = tmpCanvas.canvas;


src/display/canvas.js
```diff
@@ -1082,11 +1082,19 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
       var current = this.current;
       var x = current.x, y = current.y;
+      var xw, yh, origX, origY, pixelSize = this.getSinglePixelWidth();
       for (var i = 0, j = 0, ii = ops.length; i < ii; i++) {
         switch (ops[i] | 0) {
           case OPS.rectangle:
-            x = args[j++];
-            y = args[j++];
+            origX = args[0];
+            origY = args[1];
```
Snuffleupagus: Having these be defined as
```javascript
origX = args[j++];
origY = args[j++];
```
and then simply using `origX`/`origY` when defining `x`/`y` just below instead seems more readable (rather than having a bunch of extra `args[j++]` occurrences).

src/display/canvas.js
```diff
@@ -1082,11 +1082,19 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       var ctx = this.ctx;
       var current = this.current;
       var x = current.x, y = current.y;
+      var xw, yh, origX, origY, pixelSize = this.getSinglePixelWidth();
```
Snuffleupagus: Please remember to use `let`/`const` rather than `var` when adding *new* code.

src/display/canvas.js
```diff
@@ -1095,8 +1103,13 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
             if (height === 0) {
               height = this.getSinglePixelWidth();
```
Snuffleupagus: Nit: It might perhaps be a tiny bit more efficient to avoid extra function calls here, and in the `width === 0` case just above, by utilizing the now already available `pixelSize` value instead.

src/display/canvas.js
ThomasdenH: Because of the loop, `args[j++]` is not the same as `args[0]`.

src/display/canvas.js
```diff
@@ -1082,11 +1082,19 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
+      var xw, yh, origX, origY, pixelSize = this.getSinglePixelWidth();
       for (var i = 0, j = 0, ii = ops.length; i < ii; i++) {
         switch (ops[i] | 0) {
           case OPS.rectangle:
-            x = args[j++];
-            y = args[j++];
+            origX = args[0];
+            origY = args[1];
+            if (pixelSize) {
+              x = Math.floor(args[j++] / pixelSize) * pixelSize;
```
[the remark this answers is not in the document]
brianholle: I am new to html canvas's but here was my logic. Please verify I am looking at this correctly.
The goal is to extend all images outward to the nearest pixel. Compressing inward would generate gaps between images.

With each (x,y) point represented in the canvas with (0,0) being the top left most point.

Here, we are drawing a rectangle. So the goal is to extend each point as follows:

Top Left (x,y):  x rounds towards left (floor). Y rounds north towards y=0 (ceil)
Top Right (xw, y): xw rounds towards right (ceil). Y rounds north (ceil).
Bottom Right (xw, yh): xw rounds towards right (ceil). yh rounds south (while represented by positive value(?), so, to make it as large as possible, ceil.
Bottom Left (x, yh): x rounds towards left (floor). yh rounds south (ceil).

I have some confusion over how the Y variable is treated here. Is its value negative, considering its under the y=0 line (top of document)? Var height is set to negative in my example cases, while var Y is set to positive. Var YH is defined as height+y, where Y is normally always larger than height. This confusion could possibly lead to incorrect rounding.

src/display/canvas.js
```diff
@@ -1941,7 +1954,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       var ctx = this.ctx;
       // scale the image to the unit square
       ctx.scale(1 / w, -1 / h);
-
+      var pixelSize = this.getSinglePixelWidth();
```
timvandermeij: `var` should be `const`.

src/display/canvas.js
```diff
@@ -2144,6 +2159,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
                                                   newWidth, newHeight);
         tmpCtx = tmpCanvas.context;
         tmpCtx.clearRect(0, 0, newWidth, newHeight);
+        var pixelSize = this.getSinglePixelWidth();
```
timvandermeij: `var` should be `const`.

timvandermeij (commented):
After making these changes, could you squash the commits into one? Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for how to do that easily if you're not familiar with the process. It should get rid of the merge commit too.

src/display/canvas.js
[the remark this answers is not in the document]
brianholleibm: @timvandermeij  If you have a moment, please review this comment. I would like a second opinion before we merge, it may be possible that I have made a mistake in my decision to use .ceil / .floor in specific cases.

brendandahl (changes requested):
Looks like there a few real regressions where lines either disappear or are now too bold. Can you look into the failures?

src/display/canvas.js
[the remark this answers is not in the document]
brianholle: Actually, it seems I get better results with the following:
```
x = Math.floor(origX / pixelSize) * pixelSize;
y = Math.floor(origY / pixelSize) * pixelSize;

xw = Math.ceil((origX + width) / pixelSize) * pixelSize;
yh = Math.ceil((origY + height) / pixelSize) * pixelSize;
```
I believe that the canvas actually treats Y height as positive, with y=0 being the top of the page and y>0+n = bottom.

Where my confusion comes from, I think, is the example file that I am using. When drawing rectangles from the top of the page to the bottom, the following values are being passed into "constructPath": (571.41, 684.211, 5.303, -578.104)

These values are X, Y, Width, Height respectively. What is strange here is that, according to my believe of X/Y values on a canvas, (X,Y) actually represents the lower left hand corner of our rectangle instead of the expected upper left corner in this example. 

When YH is calculated, we are adding the negative height to our Y value, giving us the top side of our rectangle. In this case, we would want to round this value towards the top of the page (y=0) so we would want to floor this value.

However, if the usual case is that (X,Y) is the upper left corner of our image (like I originally expected), that would mean that height would be a positive value, and therefore YH would represent the bottom of our rectangle. In this case, we would want to make YH rounded upward towards the bottom of the document, aka Math.ceil(YH).

I am not too sure what determines the values of (X,Y, Height, Width) when passed into "constructPath." Can an assumption ever be made about which point (top left, top right, bottom left, bottom right) (X,Y) actually represents? If not, I need to add more code to determine the location of (X,Y) before making assumptions about the direction of rounding.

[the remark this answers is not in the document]
brianholle: @brendandahl

marktanner1331:
The issue is unfortunately more complex than believed:

A majority of PDF's will first set up a clipping path before drawing the image. The clipping path will be the exact dimensions of the image. 

This clipping path will suffer from the same sub-pixel rendering issues as described above, and cause the white lines to appear even though the image is technically the right size. 

To test this theory, try commenting out the clipping code in canvas.consumePath() leaving only the ctx.beginPath(); you'll notice that the white lines are gone.

ssaring:
We have the same rendering problem in our current project, it's reported in issue #11400.
It would be very helpful for us to know whether there's a chance for solving these problems.

Unfortunately I have no programming skills in the PDF and rendering context, is there any other way to support the progress?

jasonbolbar:
Is there any update for this? I'm interested in the solution for this issue. I tried to check the code cloning the repo from @brianholleibm and checking out to **3351_White_Bar_Fix** branch, later I ran **gulp generic** in order to generate the viewer and finally I tested in an iPhone using BrowserStack getting this result:
phone_test

Snuffleupagus:
Closing for now, since there's not been any updates for a couple of years and as-is this patch cannot land given that there's too many regressions in existing test-cases; a number of issues remain open to track this (please see https://github.com/mozilla/pdf.js/pull/10016#issue-211238996 for a list).

PR #10019, opened 2018-08-28
Add general support for re-dispatching events, on `EventBus` instances, to the DOM

Snuffleupagus:
This patch is the first step to be able to eventually get rid of the `attachDOMEventsToEventBus` function, by allowing `EventBus` instances to simply re-dispatch most[1] events to the DOM.
Note that the re-dispatching is purposely implemented to occur *after* all registered `EventBus` listeners have been serviced, to prevent the ordering issues that necessitated the duplicated page/scale-change events.

The DOM events are currently necessary for the `mozilla-central` tests, see https://hg.mozilla.org/mozilla-central/file/tip/browser/extensions/pdfjs/test, and perhaps also for custom deployments of the PDF.js default viewer.

Once this have landed, and been successfully uplifted to `mozilla-central`, I intent to submit [a patch to update the test-code](https://gist.github.com/Snuffleupagus/68bbd9650ad3144b7d311a7b2748ec98) to utilize the new preference. This will thus, eventually, make it possible to [remove the `attachDOMEventsToEventBus` functionality](https://gist.github.com/Snuffleupagus/19e3228a442f86921a9fde1cd61f34b6).

*Please note:* I've successfully ran all `mozilla-central` tests locally, with these patches applied.

---
[1] The exception being events that originated on the `window` or `document`, since those are already globally available anyway.

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -47,6 +47,10 @@
       "type": "boolean",
       "default": false
     },
+    "eventBusDispatchToDOM": {
+      "type": "boolean",
+      "default": false
+    },
     "pdfBugEnabled": {
       "title": "Enable debugging tools",
       "description": "Whether to enable debugging tools.",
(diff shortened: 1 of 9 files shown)


Snuffleupagus:
Now with unit-tests added; no other changes were made.

/botio unittest

timvandermeij (approved):

timvandermeij:
Looks good to me. Thanks!

PR #10028, opened 2018-09-01
Add initial support for "Whole words" searching in the viewer

Snuffleupagus:
As outlined in https://bugzilla.mozilla.org/show_bug.cgi?id=1282759 the internal Firefox name for the feature is `entireWord`, hence that name is used here as well for consistency (with "Whole words" being limited to the UI).

Given existing limitations of the PDF.js search functionality, e.g. the existing problems of searching across "new lines", there's some edge-cases where "Whole words" searching will ignore (valid) results.
However, considering that this is a pre-existing issue related to the way that the find controller joins text-content together, that shouldn't have to block this new feature in my opionion.

*Please note:* In order to enable this feature in the `MOZCENTRAL` version, a small follow-up patch for [PdfjsChromeUtils.jsm](https://hg.mozilla.org/mozilla-central/file/tip/browser/extensions/pdfjs/content/PdfjsChromeUtils.jsm) will be required once this has landed in `mozilla-central`.

**Edit:** See also https://bugzilla.mozilla.org/show_bug.cgi?id=1282759

### DIFF
--- a/test/unit/pdf_find_utils_spec.js
+++ b/test/unit/pdf_find_utils_spec.js
@@ -1,0 +1,56 @@
+/* Copyright 2018 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CharacterType, getCharacterType } from '../../web/pdf_find_utils';
+
+describe('pdf_find_utils', function() {
+  describe('getCharacterType', function() {
+    it('gets expected character types', function() {
+      const characters = {
+        'A': CharacterType.ALPHA_LETTER,
+        'a': CharacterType.ALPHA_LETTER,
+        '0': CharacterType.ALPHA_LETTER,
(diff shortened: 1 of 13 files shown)


test/unit/pdf_find_utils_spec.js
```diff
@@ -0,0 +1,49 @@
...
+        ' ': CharacterType.SPACE,
+        '\t': CharacterType.SPACE,
+        '\r': CharacterType.SPACE,
+        '\n': CharacterType.SPACE,
+        '\xA0': CharacterType.SPACE,
+        '-': CharacterType.PUNCT,
+        ',': CharacterType.PUNCT,
+        '.': CharacterType.PUNCT,
+        ';': CharacterType.PUNCT,
+        ':': CharacterType.PUNCT,
```
timvandermeij: Would it be possible to add a few more tests here for the missing character types, namely:

- HAN_LETTER
- KATAKANA_LETTER
- HIRAGANA_LETTER
- HALFWIDTH_KATAKANA_LETTER
- THAI_LETTER

I think we can simply find them with Wikipedia or perhaps the Mozilla test suite. It would be good to have test coverage for those too.

timvandermeij (changes requested):
Looks good with the comment addressed and a rebase. If you would like to add it, the Dutch translation for the new string is:

```
find_entire_word_label=Hele woorden
```

test/unit/pdf_find_utils_spec.js
Snuffleupagus: > Would it be possible to add a few more tests here for the missing character types [...]

Of course; this was just me being lazy and not wanting to spend to much time on it until there was feedback on the overall approach :-)

---

> If you would like to add it, the Dutch translation for the new string is:
> find_entire_word_label=Hele woorden

Thanks, it's been added to the patch.

timvandermeij (approved):

timvandermeij:
Looks good now. Thank you!

Issue #9513, opened 2018-02-27
Colors in JPGs with CMYK are not parsed correctly


photopea:
Hi, I have several CMYK files, where I get incorrect colors after decoding them with pdf.js

I discovered, that it could be fixed by inverting the input. At line 1051 (and four following lines) https://github.com/mozilla/pdf.js/blob/master/src/core/jpg.js#L1051 we should have Y = 255-data[i] and vice versa.

I would like to find the original author and ask him about this part of the code. I could not find any CMYK JPG, for which the current code would work. I only found 3 JPGs, for which my "inverted" code works.

![jpg11 1](https://user-images.githubusercontent.com/15955558/36703609-b469dbe0-1b5c-11e8-81b7-4e81e8f009e7.jpg)
![channel_digital_image_cmyk_color](https://user-images.githubusercontent.com/15955558/36703610-b4b605f6-1b5c-11e8-9e85-f09173887de5.jpg)
![insta](https://user-images.githubusercontent.com/15955558/36703611-b4f0bfac-1b5c-11e8-803a-a6e1b8835d85.jpg)

photopea:
Guys, I created a pull request. It solves the problem and does not break anything (but only for my set of testing JPGs). https://github.com/mozilla/pdf.js/pull/10029

PR #10029, opened 2018-09-01
Channels in four-channel images should be inverted.

photopea:
This solves these images: https://github.com/mozilla/pdf.js/issues/9513 and does not break any other 4-channel JPGs that I have. But it could be good to test it with corresponding PDFs too, if you have any.

### DIFF
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -1065,10 +1065,10 @@ var JpegImage = (function JpegImageClosure() {
       var Y, Cb, Cr, k;
       var offset = 0;
       for (var i = 0, length = data.length; i < length; i += 4) {
-        Y = data[i];
-        Cb = data[i + 1];
-        Cr = data[i + 2];
-        k = data[i + 3];
+        Y  = 255-data[i];
+        Cb = 255-data[i + 1];
+        Cr = 255-data[i + 2];
+        k  = 255-data[i + 3];
 
         data[offset++] = -122.67195406894 +
           Cb * (-6.60635669420364e-5 * Cb + 0.000437130475926232 * Cr -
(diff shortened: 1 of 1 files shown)


src/core/jpg.js
```diff
@@ -1065,10 +1065,10 @@ var JpegImage = (function JpegImageClosure() {
       var Y, Cb, Cr, k;
       var offset = 0;
       for (var i = 0, length = data.length; i < length; i += 4) {
-        Y = data[i];
-        Cb = data[i + 1];
-        Cr = data[i + 2];
-        k = data[i + 3];
+        Y  = 255-data[i];
```
timvandermeij: Here, and in other parts of the part, please leave one and only one space around the `-` and `=` operators, e.g., `Y = 255 - data[i]`.

timvandermeij (changes requested):
Thank you for looking into this. Aside from the comment below, please also add a reference test to this PR so we can avoid regressions later on. If you have a file that is free for us to use, please use the steps at https://github.com/mozilla/pdf.js/wiki/Contributing under `Generating reference images` to add it using the `test/add_test.js` tool.

Snuffleupagus:
> But it could be good to test it with corresponding PDFs too, if you have any.

Please see https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing for information about running test (and generating reference the images to test against). When modifying code residing in the `/src` folder, it's usually recommend/necessary to test locally when making changes.

Even without running any tests, I can pretty much guarantee that this patch would cause significant rendering regressions in existing test-cases.
*Ninja edit:* One broken example is https://github.com/mozilla/pdf.js/blob/master/test/pdfs/cmykjpeg.pdf

photopea:
@Snuffleupagus I extracted the "raw" JPG out of https://github.com/mozilla/pdf.js/blob/master/test/pdfs/cmykjpeg.pdf  and got this: 
pic

And it is displayed as "black" by JPG viewers. Could it be, that there are extra parameters in a PDF file, that make it look "bright"?

timvandermeij:
It may indeed be more difficult than this PR if I'm reading the thread in https://github.com/haraldk/TwelveMonkeys/issues/141#issuecomment-108570213 and https://github.com/haraldk/TwelveMonkeys/issues/141#issuecomment-109429682 correctly.

photopea:
@timvandermeij So I suggest, that we give one more Boolean parameter to getData: JpegImage.getData(w,h,forceRGBoutput, invertChannels) - false by default. getData() will distribute this parameter to sub-functions.

Then, you will call  getData with the last parameter depending on information outside the JPEG stream. Can you find that information and rewrite all getData() calls? Then, I will update the JPEG decoder.

timvandermeij:
I don't know enough about JPEG decoding to know where to fetch this information from outside the JPEG stream, so I can't really help with this. I would suggest to look up the JPEG stream part in the PDF specification to find out how this should work before adding any parameters.

PR #10031, opened 2018-09-02
Add a new parameter to `JpegImage.getData` to indicate the source of the image data (issue 9513)

Snuffleupagus:
The purpose of this patch is to provide a better default behaviour when `JpegImage` is used to parse standalone JPEG images with CMYK colour spaces.
Since the issue that the patch concerns is somewhat of a special-case, the implementation utilizes the already existing decode support in an attempt to minimize the impact w.r.t. code size.

*Please note:* It's always possible for the user of `JpegImage` to control image inversion, and thus override the new behaviour, by simply passing a custom `decodeTransform` array upon initialization.

***Edit:*** I should probably add that all of this feels hand-wavy, at least to me, as far as solutions go...

### DIFF
--- a/examples/image_decoders/jpeg_viewer.js
+++ b/examples/image_decoders/jpeg_viewer.js
@@ -58,7 +58,11 @@ var jpegImage = new pdfjsImageDecoders.JpegImage();
 jpegImage.parse(typedArrayImage);
 
 var width = jpegImage.width, height = jpegImage.height;
-var jpegData = jpegImage.getData(width, height, /* forceRGB = */ true);
+var jpegData = jpegImage.getData({
+  width,
+  height,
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
I have enough confidence in this solution because it doesn't modify the normal behavior (no test failures), is well documented in the comments with a reference to the original issue and is easy in terms of code. Thank you for helping out with this!

photopea:
I have a related question. I want to let users extract JPG out of a PDF file. Is it possible to build the "isSourcePDF" parameter into a binary JPG, without having to decompress it and compress it again? Maybe just change a few bytes in some metadata?

PR #10033, opened 2018-09-02
[api-minor] Implement a permissions API

### LINKED ISSUE #9972: Expose the document permissions via API
Does PDF.js have an API to retrieve the document permissions?

We would like to see if the document is printable or is editable.
This will be used for verification and be categorized.

If there are not, what is the plan for introducing the ability to retrieve this information.

timvandermeij:
Fixes #9972.

@Snuffleupagus Would you be willing to review this if you have time? I recommend to review this per commit and with the `?w=1` flag. For some reason the `?w=1` flag doesn't seem to work on GitHub for the second commit, but it does locally if you use `git show HEAD~1 --ignore-all-space` with this branch checked out.


src/core/obj.js
```diff
@@ -177,6 +177,48 @@ class Catalog {
...
+
+    let flags = encrypt.get('P');
+    if (!isNum(flags)) {
+      return null;
+    }
+
+    // PDF integer objects are represented internally in signed 2's complement
+    // form. Therefore, convert the signed decimal integer to a signed 2's
+    // complement binary integer so we can use regular bitwise operations on it.
+    flags += 2 ** 32;
```
Snuffleupagus: I assuming that you've already checked, but otherwise please make sure that Webpack/Babel is able to convert this into code compatible with older browsers. (Note that I'm expecting this to "just work" given [this option](https://github.com/mozilla/pdf.js/blob/master/external/builder/preprocessor2.js#L291)).

src/core/obj.js
```diff
@@ -177,6 +177,48 @@ class Catalog {
...
+
+    const permissions = [];
+    for (const key in PermissionFlag) {
+      const value = PermissionFlag[key];
```
Snuffleupagus: If you include something along these lines https://github.com/mozilla/pdf.js/pull/9982/commits/e59f952a59315a7346cc0fc7277c9067508c479f, then you ought to be able to use `for...of` instead here:
```javascript
for (const flag of PermissionFlag) {
  if (flags & flag) {
    ...
}
```

src/display/api.js
```diff
@@ -577,166 +577,179 @@ var PDFDataRangeTransport = (function pdfDataRangeTransportClosure() {
...
+  /**
+   * @return {Promise} A promise this is resolved with current statistics about
+   *   document structures (see {@link PDFDocumentStats}).
+   */
+  getStats() {
+    return this._transport.getStats();
+  }
+
+  /**
+   * Cleans up resources allocated by the document, e.g., created `@font-face`.
```
Snuffleupagus: Nit: Remove the comma placed *after* `e.g.`

src/display/api.js
```diff
@@ -691,6 +691,14 @@ class PDFDocumentProxy {
     return this._transport.getOutline();
   }
 
+  /**
+   * @return {Promise} A promise that is resolved with an {Array} that contains
+   *   the permission flags for the PDF document.
```
Snuffleupagus: Similar to the `getPageLabel` JSDocs, please mention the `null` case as well.

test/unit/api_spec.js
```diff
@@ -822,6 +823,63 @@ describe('api', function() {
         done.fail(reason);
       });
     });
+
+    it('gets non-existent permissions', function(done) {
+      doc.getPermissions().then(function(permissions) {
+        expect(permissions).toEqual(null);
+
+        done();
+      }).catch(function(reason) {
```
Snuffleupagus: Here, and elsewhere below, the slightly less verbose `}).catch(done.fail);` ought to work.

Snuffleupagus (approved):
> For some reason the ?w=1 flag doesn't seem to work on GitHub for the second commit,

Thanks for the heads-up, I was wondering why the diff looked so broken :-)

Overall this seems fine to me, modulo a couple of minor comments; nice work!

src/core/obj.js
timvandermeij: Good point. I found that `babel-present-env` fortunately includes the exponentiation operator plugin (see https://github.com/babel/babel/blob/master/packages/babel-preset-env/package.json#L36), so that should be fine.

src/core/obj.js
timvandermeij: I tried this and unfortunately it doesn't work. Objects, unlike other data structures such as `Map`s, are not iterable in JavaScript. According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/is_not_iterable there are some other ways to do this but I don't think it's much more readable.

Snuffleupagus: I totally overlooked that it's an Object, sorry about wasting your time here.

Issue #7799, opened 2016-11-11
Remove mozCurrentTransform usage


yurydelendik:
We use mozCurrentTransform track device pixel size and just to get combined transform. The mozCurrentTransform might not be efficient from platforms point of view (blocks painting pipeline?). It also complicates our code with creating and polyfilling scratch canvases.

Tracking transform internally might simplify canvas.js code and remove unneeded ctx.save/.restore operations. (See also https://github.com/mozilla/pdf.js/issues/7188)

Snuffleupagus:
> Tracking transform internally might simplify canvas.js code and remove unneeded ctx.save/.restore operations. (See also #7188)

Note that those `ctx.save/ctx.restore` commands *may* already be redundant, since [bug 1305963](https://bugzilla.mozilla.org/show_bug.cgi?id=1305963) could have addressed the source of the intermittent issues.

yurydelendik:
> Note that those ctx.save/ctx.restore commands may already be redundant, since bug 1305963 could have addressed the source of the intermittent issues.

We perform more of those, but just to find out the combined transform, which can be easy replaced by Util.transform operations.

calixteman:
This bug is fixed now: it can be closed.

PR #7330, opened 2016-05-16
Try to prevent `CanvasGraphics_getSinglePixelWidth` from intermittently returning incorrect values in Firefox (issue 7188)

Snuffleupagus:
_Note:_ Locally this patch causes a number of test "failures", all of which are completely imperceivable to the naked eye.

Fixes #7188.

Issue #7188, opened 2016-04-12
Thin line is rendered too thick


THausherr:
Link to PDF file (or attach file here):

Configuration:
- Web browser and its version: ff 45.0.1
- Operating system and its version: W7 64
- PDF.js version: newest
- Is an extension: ?

Steps to reproduce the problem:
1. https://mozilla.github.io/pdf.js/web/viewer.html
2. open the file from https://issues.apache.org/jira/secure/attachment/12652090/CIB-coons-vs-tensormesh.pdf

What is the expected behavior? (add screenshot)
![scr1](https://cloud.githubusercontent.com/assets/6665575/14466424/11fbc950-00d7-11e6-806a-675b84ffe901.jpg)

What went wrong? (add screenshot)
![scr2](https://cloud.githubusercontent.com/assets/6665575/14466432/1926376a-00d7-11e6-80f8-6f96e4f5a0c8.jpg)

I thought this is a regression from todays fix (thanks!) but it is also in the pre-fix version that it in my browser.

yurydelendik:
It is intermittent for me (on Mac OS X).

Snuffleupagus:
I can reproduce this at *some* zoom levels, but *not* at others, which seems very strange.
Interestingly enough, if I use the `Stepper` to render the file one operation at a time, it always works correctly *regardless* of zoom level. Furthermore, if I force the use of the `mozCurrentTransform` polyfill, https://github.com/mozilla/pdf.js/blob/master/src/display/canvas.js#L92-L93, it also works regardless of zoom level.

All this suggests, to me, that `this.ctx.{mozCurrentTransform,mozCurrentTransformInverse}` intermittently returns incorrect values (perhaps the previous ones?).
So one solution, that I've seen mentioned on IRC as a solution for the general transformation mess in `canvas.js`, might be that we track the transforms ourselves, to avoid these (apparent) platform issues.

However, in the meantime, I've found a possible workaround:
```diff
diff --git a/src/display/canvas.js b/src/display/canvas.js
index e9b3563..97a4c5c 100644
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -2298,7 +2298,9 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
     },
     getSinglePixelWidth: function CanvasGraphics_getSinglePixelWidth(scale) {
       if (this.cachedGetSinglePixelWidth === null) {
+        this.ctx.save();
[...]

PR #10034, opened 2018-09-02
Remove `getSinglePixelWidth` workaround

timvandermeij:
It's no longer necessary since https://bugzilla.mozilla.org/show_bug.cgi?id=1305963 is fixed quite some time ago.

While we're here, mark the `cachedGetSinglePixelWidth` member as being private and use ES6 syntax in the `getSinglePixelWidth` method.

Fixes a part of #7799.

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -435,7 +435,7 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       // the transformation must already be set in canvasCtx._transformMatrix.
       addContextCurrentTransform(canvasCtx);
     }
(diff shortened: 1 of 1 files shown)


timvandermeij:
The test failures are expected since they also occurred in the original patch from @Snuffleupagus in #7330. I verified them and the differences are invisible to the naked eye.

Snuffleupagus (approved):
Looks good to me, thanks for cleaning this up!

Note: I've tested this patch locally, with the PDF file from issue #7188, and everything appears to work correctly now.

PR #10041, opened 2018-09-05
Revert "Attempt to combine separate beginText/endText sequences in `getTextContent` (issue 9984)"

Snuffleupagus:
Reverts mozilla/pdf.js#9986 for causing Mochitest failures, please refer to https://bugzilla.mozilla.org/show_bug.cgi?id=1488420#c2 and the following discussion.

/cc @timvandermeij
/cc @rvandermeulen


timvandermeij:
I'll create an issue to look into why this is failing upstream, since I would like this functionality back, but I agree it's best to back it out for now so we can upstream.

/botio makeref

src/core/evaluator.js
```diff
@@ -1546,7 +1535,6 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
         task.ensureNotTerminated();
         timeSlotManager.reset();
         var stop, operation = {}, args = [];
-        let pendingBeginText = false;
```
Snuffleupagus: *Please note:* This is actually a bug in the original PR, since this line should be placed e.g. just after https://github.com/mozilla/pdf.js/pull/10041/files#diff-0b94c2e77a5259f7a728122fdbf9f46aL1533 instead to ensure that `pendingBeginText` may never be accidentally reset.

PR #10042, opened 2018-09-05
Set spread mode from query string

randomstuff:


Snuffleupagus:
Note that no such parameter exists in https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/pdf_open_parameters.pdf, and adding non-standard URL parameters is something that has previously always been avoided.

timvandermeij:
Closing since I agree with the comment above. We only aim to offer the parameters from the specification, and to make an exception there must be a good use case. I don't see why it would be beneficial to be able to change this via the URL parameters instead of just using the viewer toggles or the preferences (`spreadModeOnLoad`).

PR #10049, opened 2018-09-07
Fix font-string variable name typo

romainptit:
The font-string rebuild condition is always satisfied because the concerned variables are never set.


timvandermeij (approved):

timvandermeij:
Interesting find. Thank you for improving this!

PR #10052, opened 2018-09-08
Display the index of the currently active search result in the matches counter of the findbar (issue 6993, bug 1062025)

### LINKED ISSUE #6993: Display the index of the find results in find bar
Referencing #6580, is it feasible to add an index to the counter, e.g. "2 out of 5" instead of just showing the number "5" when (in this example) the second search result is highlighted?

And a tooltip that shows what the number means when hovering over the number may also be helpful.

Snuffleupagus:
For the `PDFFindBar` implementation, similar to the native Firefox findbar, the matches count displayed is now limited to a (hopefully) reasonable value; compare with https://searchfox.org/mozilla-central/rev/37663bb87004167184de6f2afa6b05875eb0528e/modules/libpref/init/all.js#1031

*Please note:* In order to enable this feature in the `MOZCENTRAL` version, a follow-up patch will be required once this has landed in `mozilla-central`.

Fixes #6993.

### DIFF
--- a/web/firefoxcom.js
+++ b/web/firefoxcom.js
@@ -209,6 +209,10 @@ PDFViewerApplication.externalServices = {
     FirefoxCom.request('updateFindControlState', data);
   },
 
+  updateFindMatchesCount(data) {
+    // FirefoxCom.request('updateFindMatchesCount', data);
+  },
+
   initPassiveLoading(callbacks) {
     let pdfDataRangeTransport;
 
--- a/l10n/en-US/viewer.properties
+++ b/l10n/en-US/viewer.properties
@@ -167,6 +167,13 @@ find_highlight=Highlight all
 find_match_case_label=Match case
 find_reached_top=Reached top of document, continued from bottom
 find_reached_bottom=Reached end of document, continued from top
+# LOCALIZATION NOTE (find_matches_count): "{{current}}" and "{{total}}" will be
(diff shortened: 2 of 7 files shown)


web/firefoxcom.js
```diff
@@ -209,6 +209,10 @@ PDFViewerApplication.externalServices = {
     FirefoxCom.request('updateFindControlState', data);
   },
 
+  updateFindMatchesCount(data) {
+    // FirefoxCom.request('updateFindMatchesCount', data);
```
Snuffleupagus: Note that this is commented, out for now, since https://searchfox.org/mozilla-central/rev/37663bb87004167184de6f2afa6b05875eb0528e/browser/extensions/pdfjs/content/PdfStreamConverter.jsm#719,740 will throw without a `mozilla-central` follow-up patch.

Snuffleupagus: Note that this is purposely commented out, for now, since https://searchfox.org/mozilla-central/rev/37663bb87004167184de6f2afa6b05875eb0528e/browser/extensions/pdfjs/content/PdfStreamConverter.jsm#719,740 will throw without the `mozilla-central` follow-up patch.

timvandermeij (approved):

timvandermeij:
Works really well. Thank you!

PR #10056, opened 2018-09-10
[Regression] Ensure that `PDFFindBar.updateResultsCount` doesn't throw when the viewer is closed, by providing proper default values

Snuffleupagus:
The error can be reproduced by opening any file in the viewer, and then running `PDFViewerApplication.close()` in the console.

/cc @timvandermeij 
/cc @rvandermeulen

*Edit:* Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1489996#c2

### DIFF
--- a/web/pdf_find_bar.js
+++ b/web/pdf_find_bar.js
@@ -153,7 +153,7 @@ class PDFFindBar {
     this.updateResultsCount(matchesCount);
   }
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
I'm not sure how I also missed this during review, but thank you for fixing it so quickly!

PR #10060, opened 2018-09-11
Ensure that `PDFFindController._requestMatchesCount` won't return broken data when searching starts (PR 10052 follow-up)

Snuffleupagus:
This is an unfortunate oversight on my part, which I stumbled upon when (locally) testing the `mozilla-central` follow-up patch necessary to enable the matches counter in the built-in PDF viewer.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #10061, opened 2018-09-11
Implement unit tests for the `isSameOrigin` and `createValidAbsoluteUrl` utility functions and use the `const` keyword for constants in `src/shared/util.js`

timvandermeij:
This helps to increase test coverage of the `src/shared/util.js` file.

@Snuffleupagus Would you be willing to review this?

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -332,7 +332,7 @@ function isSameOrigin(baseUrl, otherUrl) {
 }
 
 // Checks if URLs use one of the whitelisted protocols, e.g. to avoid XSS.
-function isValidProtocol(url) {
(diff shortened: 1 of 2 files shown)


Snuffleupagus (approved):
More unit-test coverage is always great to have, and collecting the constants in one place will make the file easier to deal with; thank you for doing this!

timvandermeij:
It turns out that the `URL` implementation in Firefox and Chrome differs slightly, so for the `mailto:` protocol Firefox gives origin `null` and Chrome gives origin `mailto://`. I switched to `blob:` instead which shows the same behavior in both browsers.

/botio unittest

PR #10063, opened 2018-09-12
Slightly improve the `isSourcePDF` parameter handling in `JpegImage` (PR 10031 follow-up)

Snuffleupagus:
Currently there's only a single spot in the code-base where `JpegImage.getData` is called, however it nonetheless seem like a good idea to ensure during tests that the `isSourcePDF` parameter is correctly set. (Especially considering that the PDF use-cases will break without it.)

Additionally, in `JpegImage._getLinearizedBlockData`, the code can be made a tiny bit more efficient by checking the value of `isSourcePDF` *first* to avoid useless checks (for the default PDF use-cases).

### DIFF
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -14,7 +14,7 @@
  */
 /* eslint-disable no-multi-spaces */
 
-import { warn } from '../shared/util';
+import { assert, warn } from '../shared/util';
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me. Thank you!

PR #10064, opened 2018-09-12
updated the year

vschs007:
year update


Snuffleupagus:
If, and that's a big if in my opinion (given that not all files are authored by Mozilla), there's a desire to update the year in the copyright notices of individual source files, then that should happen everywhere at once and not piecemeal like this.

For the built files, we're using special license headers anyway (see https://github.com/mozilla/pdf.js/blob/master/src/license_header.js and https://github.com/mozilla/pdf.js/blob/master/src/license_header_libre.js) so I'm not even sure what the motivation for this change is!?

Issue #7356, opened 2016-05-25
Add pdf_find_controller testing


yurydelendik:
We need to refactor and add PDFFindController unit tests. Currently it accepts PDFViewer however it uses only few methods to perform callbacks. It is recommended to remove dependency on PDFViewer, use only PDFDocument as source of the text content and generate events to update UI results.

yurydelendik:
(atm it's hard to add tests at #5579)

PR #10066, opened 2018-09-12
Refactor the find controller

timvandermeij:
I looked into #7356 and quickly found that the find controller code is a bit complex, mainly because there is quite a bit of code to actually compute the matches, but also because it performs other tasks like scrolling matches into view and updating matches on pages.

The objective of this pull request is to work towards making this issue easier to solve. We do so by removing unnecessary/dead code to simplify the code and to reduce the amount of state that the object has to keep track of. Moreover, we modernize the code to make it more readable. Finally, we move the code for scrolling matches into view to the text layer builder since that shouldn't be the responsibility of the find controller and the match rendering logic is already there.

If this is merged, the code will be easier to refactor further (as described in the issue mentioned before) and to start adding unit tests for it.

@Snuffleupagus Would you perhaps be willing to review this? I would suggest checking this per commit since I explained some of the changes in the commit message and each commit is relatively small.

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -57,8 +57,8 @@ class PDFFindController {
     this.reset();
 
     // Compile the regular expression for text normalization once.
-    let replace = Object.keys(CHARACTERS_TO_NORMALIZE).join('');
-    this._normalizationRegex = new RegExp('[' + replace + ']', 'g');
+    const replace = Object.keys(CHARACTERS_TO_NORMALIZE).join('');
+    this._normalizationRegex = new RegExp(`[${replace}]`, 'g');
   }
 
   get pageMatches() {
@@ -141,7 +141,7 @@ class PDFFindController {
   updateMatchPosition(pageIndex, matchIndex, elements, beginIdx) {
     if (this.selected.matchIdx === matchIndex &&
         this.selected.pageIdx === pageIndex) {
-      let spot = {
+      const spot = {
         top: FIND_SCROLL_OFFSET_TOP,
         left: FIND_SCROLL_OFFSET_LEFT,
       };
@@ -151,7 +151,7 @@ class PDFFindController {
   }
 
   _normalize(text) {
-    return text.replace(this._normalizationRegex, function (ch) {
+    return text.replace(this._normalizationRegex, function(ch) {
       return CHARACTERS_TO_NORMALIZE[ch];
     });
   }
@@ -164,8 +164,8 @@ class PDFFindController {
    */
   _prepareMatches(matchesWithLength, matches, matchesLength) {
     function isSubTerm(matchesWithLength, currentIndex) {
-      let currentElem = matchesWithLength[currentIndex];
-      let nextElem = matchesWithLength[currentIndex + 1];
+      const currentElem = matchesWithLength[currentIndex];
+      const nextElem = matchesWithLength[currentIndex + 1];
 
       // Check for cases like "TAMEd TAME".
       if (currentIndex < matchesWithLength.length - 1 &&
@@ -176,7 +176,7 @@ class PDFFindController {
 
       // Check for cases like "thIS IS".
       for (let i = currentIndex - 1; i >= 0; i--) {
-        let prevElem = matchesWithLength[i];
+        const prevElem = matchesWithLength[i];
         if (prevElem.skipped) {
           continue;
         }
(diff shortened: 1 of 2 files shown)


web/pdf_find_controller.js
```diff
@@ -359,23 +360,23 @@ class PDFFindController {
...
-      this.dirtyMatch = false;
+      this._dirtyMatch = false;
       this.selected.pageIdx = this.selected.matchIdx = -1;
-      this.offset.pageIdx = currentPageIndex;
-      this.offset.matchIdx = null;
-      this.resumePageIdx = null;
+      this._offset.pageIdx = currentPageIndex;
+      this._offset.matchIdx = null;
+      this._resumePageIdx = null;
       this.pageMatches = [];
```
Snuffleupagus: This is pre-existing, but it might be a tiny bit more efficient to do `this.pageMatches.length = 0;` here instead, since that would re-use the existing array rather than creating a new one.

web/pdf_find_controller.js
```diff
@@ -272,12 +275,11 @@ class PDFFindController {
   _calculateMatch(pageIndex) {
     let pageContent = this._normalize(this._pageContents[pageIndex]);
     let query = this._normalize(this.state.query);
-    let caseSensitive = this.state.caseSensitive;
-    let phraseSearch = this.state.phraseSearch;
+    const caseSensitive = this.state.caseSensitive;
+    const phraseSearch = this.state.phraseSearch;
     const entireWord = this.state.entireWord;
```
Snuffleupagus: Nit: This above three lines could be `const { caseSensitive, entireWord, phraseSearch, } = this.state;` instead, if you wanted to :-)

web/text_layer_builder.js
```diff
@@ -243,9 +246,17 @@ class TextLayerBuilder {
       let isSelected = (isSelectedPage && i === selectedMatchIdx);
       let highlightSuffix = (isSelected ? ' selected' : '');
 
+      // Scroll the selected match into view.
       if (this.findController) {
-        this.findController.updateMatchPosition(pageIdx, i, textDivs,
-                                                begin.divIdx);
+        if (this.findController.selected.matchIdx === i &&
+            this.findController.selected.pageIdx === pageIdx) {
```
Snuffleupagus: This might be (yet another) pre-existing issue, but `selected` feels like it should be internal FindController state not directly accessed from the outside in this way.

Can `selected` be made "private" in `PDFFindController`, and instead accessed e.g. through a `get selected() { return this._selected; };` method instead?

Snuffleupagus (changes requested):
Thanks for trying to make the (very old) find-related code slightly saner, one step at a time :-)

However, I'm setting `r-` for now, since one of the commits is causing pretty serious errors and needs to be removed; please refer to the inline comment for details.

web/pdf_find_controller.js
```diff
@@ -63,7 +63,6 @@ class PDFFindController {
 
   reset() {
     this.extractTextPromises = [];
-    this.pendingFindMatches = Object.create(null);
```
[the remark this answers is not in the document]
timvandermeij: Done in the new commit series. Thank you for finding this, since I did not encounter it in my local testing unfortunately!

web/pdf_find_controller.js
timvandermeij: Done in the new commit series.

web/pdf_find_controller.js
timvandermeij: Done in the new commit series.

web/text_layer_builder.js
timvandermeij: Done in the new commit series. The only ones that are still public are members that are being assigned to from outside. I do think that this pattern should be changed later on, but I'd rather not do that in this PR to keep its size limited.

web/pdf_find_controller.js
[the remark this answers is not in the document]
Snuffleupagus: No worries; I'm guessing that you've already figured out why that commit didn't work, but otherwise please let me know and I'll try my best to explain what the problem was.

web/text_layer_builder.js
Snuffleupagus: > I do think that this pattern should be changed later on, but I'd rather not do that in this PR to keep its size limited.

Good call, let's avoid too much scope creep here.

Snuffleupagus (approved):

PR #10070, opened 2018-09-13
Attempt to support plural forms in the matches counter of the findbar (issue 10067)

### LINKED ISSUE #10067: Give find_matches_count and find_matches_count_limit proper plural forms
Follow-up to #10028 (as noted by @flodolo in the [pdf.js 2.0.843 update bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1489996#c9)):
> Both find_matches_count and find_matches_count_limit need proper plural forms to be correctly localized.

@Snuffleupagus, is that something you can take?

Snuffleupagus:
Based on a quick look at https://github.com/fabi1cazenave/webL10n/#pluralization, it seems that supporting plural forms shouldn't be as difficult as I first thought it might be.

Fixes #10067

### DIFF
--- a/l10n/en-US/viewer.properties
+++ b/l10n/en-US/viewer.properties
@@ -168,13 +168,20 @@ find_match_case_label=Match case
 find_entire_word_label=Whole words
 find_reached_top=Reached top of document, continued from bottom
 find_reached_bottom=Reached end of document, continued from top
-# LOCALIZATION NOTE (find_matches_count): "{{current}}" and "{{total}}" will be
-# replaced by a number representing the index of the currently active find result,
-# respectively a number representing the total number of matches in the document.
-find_matches_count={{current}} of {{total}} matches
-# LOCALIZATION NOTE (find_matches_count_limit): "{{limit}}" will be replaced by
-# a numerical value.
-find_matches_count_limit=More than {{limit}} matches
+# LOCALIZATION NOTE (find_matches_count): The supported plural forms are
+# [zero|one|two|few|many|other], with [other] as the default value.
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good. Thank you!

flodolo:
CC @pike @zbraniecki

Unfortunately this is not a complete fix. In order to allow locales to translate, you need to provide all forms in your .properties (one, two, few, many, other), not just the ones needed by English.

webl10n is a relic from the past (Firefox OS), and had a very hacky plural support. 

It would be nice to take a look at [Fluent](https://projectfluent.org/), and see if PDF.js could be ported to that system (which is maintained, unlike webl10n).

timvandermeij:
I'll create follow-up issues for both points.

Snuffleupagus:
> Unfortunately this is not a complete fix. In order to allow locales to translate, you need to provide all forms in your .properties (one, two, few, many, other), not just the ones needed by English.

I don't understand this comment, since there shouldn't be anything preventing a non-English locale from adding whichever of the `one, two, few, many, other` plural forms needed for a particular locale (regardless of what forms are required for English)!?
@flodolo Since I don't claim to understand the details of the localization process, could you please clarify what you mean here?

Also, keep in mind that as outlined in https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md, the translations are imported from Firefox rather than maintained in this GitHub repo.

flodolo:
BTW, just realized that this should have landed with a new ID :-\

PR #10078, opened 2018-09-15
Improve plural support for the matches counter

### LINKED ISSUE #10076: Improve plural support for the matches counter
It turns out that only specifying `one` and `other` in #10070 is not enough, so we may need to add all other options too for other locales (note that there are no bugs because of this, it should only be more helpful for other languages this way).

timvandermeij:
Fixes #10076.

### DIFF
--- a/l10n/en-US/viewer.properties
+++ b/l10n/en-US/viewer.properties
@@ -168,20 +168,27 @@ find_match_case_label=Match case
 find_entire_word_label=Whole words
 find_reached_top=Reached top of document, continued from bottom
 find_reached_bottom=Reached end of document, continued from top
-# LOCALIZATION NOTE (find_matches_count): The supported plural forms are
-# [zero|one|two|few|many|other], with [other] as the default value.
+# LOCALIZATION NOTE (find_match_count): The supported plural forms are
+# [one|two|few|many|other], with [other] as the default value.
 # "{{current}}" and "{{total}}" will be replaced by a number representing the
 # index of the currently active find result, respectively a number representing
 # the total number of matches in the document.
-find_matches_count={[ plural(n) ]}
-find_matches_count[one]={{current}} of {{total}} match
-find_matches_count[other]={{current}} of {{total}} matches
-# LOCALIZATION NOTE (find_matches_count_limit): The supported plural forms are
+find_match_count={[ plural(total) ]}
+find_match_count[one]={{current}} of {{total}} match
+find_match_count[two]={{current}} of {{total}} matches
+find_match_count[few]={{current}} of {{total}} matches
+find_match_count[many]={{current}} of {{total}} matches
+find_match_count[other]={{current}} of {{total}} matches
+# LOCALIZATION NOTE (find_match_count_limit): The supported plural forms are
 # [zero|one|two|few|many|other], with [other] as the default value.
 # "{{limit}}" will be replaced by a numerical value.
-find_matches_count_limit={[ plural(n) ]}
-find_matches_count_limit[one]=More than {{limit}} match
-find_matches_count_limit[other]=More than {{limit}} matches
+find_match_count_limit={[ plural(limit) ]}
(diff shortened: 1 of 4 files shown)


flodolo (approved):
I have no way to test this, but I believe it should work.

l10n/en-US/viewer.properties
```diff
@@ -168,20 +168,28 @@ find_match_case_label=Match case
...
 # [zero|one|two|few|many|other], with [other] as the default value.
 # "{{current}}" and "{{total}}" will be replaced by a number representing the
 # index of the currently active find result, respectively a number representing
 # the total number of matches in the document.
-find_matches_count={[ plural(n) ]}
-find_matches_count[one]={{current}} of {{total}} match
-find_matches_count[other]={{current}} of {{total}} matches
-# LOCALIZATION NOTE (find_matches_count_limit): The supported plural forms are
+find_match_count={[ plural(n) ]}
+find_match_count[zero]={{current}} of {{total}} matches
```
Snuffleupagus: This particular case should never happen, since when there's no matches the counter is hidden.

l10n/en-US/viewer.properties
```diff
@@ -168,20 +168,28 @@ find_match_case_label=Match case
...
+# LOCALIZATION NOTE (find_match_count): The supported plural forms are
 # [zero|one|two|few|many|other], with [other] as the default value.
 # "{{current}}" and "{{total}}" will be replaced by a number representing the
 # index of the currently active find result, respectively a number representing
 # the total number of matches in the document.
-find_matches_count={[ plural(n) ]}
-find_matches_count[one]={{current}} of {{total}} match
-find_matches_count[other]={{current}} of {{total}} matches
-# LOCALIZATION NOTE (find_matches_count_limit): The supported plural forms are
+find_match_count={[ plural(n) ]}
```
Snuffleupagus: Having looked at the Firefox findbar implementation again, [see here](https://searchfox.org/mozilla-central/rev/7a91966c106a3dde924e040b4e5b72d33eb77bf7/toolkit/content/widgets/findbar.js#1133-1142), I'm no longer convinced that there's any attempting to format these numbers according to locale.
So it might be possible to simply this by remove the `toLocaleString()` calls from the JS code, and then change this to just `plural(total)` (and `plural(limit` for for the other string below).

Snuffleupagus: Having looked at the Firefox findbar implementation again, [see here](https://searchfox.org/mozilla-central/rev/7a91966c106a3dde924e040b4e5b72d33eb77bf7/toolkit/content/widgets/findbar.js#1133-1142), I'm no longer convinced that there's any attempt to format these numbers according to locale.
So it might be possible to simplify this by removing the `toLocaleString()` calls from the JS code, and change this to just `{[ plural(total) ]}` respectively `{[ plural(limit) ]}` for for the other string below.

l10n/en-US/viewer.properties
Snuffleupagus: Nope; please note that `total` must be non-zero for the message to be displayed in the first place, see https://github.com/mozilla/pdf.js/blob/master/web/pdf_find_bar.js#L162 and https://searchfox.org/mozilla-central/source/toolkit/content/widgets/findbar.js#1131.

I'll admit that this was perhaps not as clear as it could have been, given that the line in `pdf_find_bar.js` probably should have been e.g. `if (total > 0) {` instead.

timvandermeij: Done in the new commit.

l10n/en-US/viewer.properties
timvandermeij: Done in the new commit. The browser find bar indeed also doesn't do any formatting according to locale, so I've removed that here too.

Snuffleupagus (approved):

timvandermeij:
@rvandermeulen Could you do a new upstream merge that contains these changes? This is necessary for a correct localization of these strings.

PR #10086, opened 2018-09-19
Fix rendering multiple blanks in SVG

dmitryskey:
The given PR is a simple fix for the dealing with multiple blanks in the text blocks for the SVG renderer. This issue is specific to Safari on both MacOS and iOS and the best example is the I-9 form (see https://www.uscis.gov/system/files_force/files/form/i-9-paper-version.pdf) where bottom parts look like below: 
image

The root cause is that Safari joins all multiple blanks in the SVG tspan control and as result the x-component of it becomes out of sync with the text. Chrome and Firefox don't seem affected and keep text as it was rendered.

Here is the sample of the tspan:

`<svg:tspan font-family="Times" font-size="1px" y="-0.0913" x="54.733 54.983 55.2331 55.7891 56.2331 56.7331 57.1771 57.4271 57.9271 58.1771 58.6771 59.0101 59.2601" fill="rgb(35,31,32)">  
 Page 1 of 3</svg:tspan>`

Safari drops the second space before 'Page', as result we have 13 x coordinates and just 12 characters. So basically in this fix we ignore all blanks in the beginning of the text as well as multiple spaces between words. Search in the SVG text is affected a little bit but it works differently with Adobe Reader anyway where all multiple blanks in the search expression are ignored.

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -746,6 +746,8 @@ SVGGraphics = (function SVGGraphicsClosure() {
       var vertical = font.vertical;
       var widthAdvanceScale = fontSize * current.fontMatrix[0];
 
+      var prevCharacter = '';
+
       var x = 0, i;
       for (i = 0; i < glyphsLength; ++i) {
         var glyph = glyphs[i];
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
This appears to (essentially) be a duplicate of PR #8943, which was rejected since it adds browser specific compatibility hacks in `/core` code and thus negatively affects code readability/performance in all browsers.

dmitryskey:
Cool, thank you, I didn't know on this PR. It totally makes sense because this is a WebKit bug https://bugs.webkit.org/show_bug.cgi?id=112032. I'll keep the fix in my branch, but in general we shouldn't implement such a browser-specific workarounds. Closing the PR.

PR #10090, opened 2018-09-20
Ensure that all event properties are included, even if no (internal) listeners are registered, when re-dispatching events to the DOM (PR 10019 follow-up)

Snuffleupagus:


timvandermeij (commented):
The change looks good, but I would like to have an additional unit test that checks if the argument passing works if it's not too much work. Currently we only unit test dispatching of events without arguments and that may be why we missed this, so a unit test that verifies that both `const args = Array.prototype.slice.call(arguments, 1);` paths get used would be good to have.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this and extending the unit tests!

Snuffleupagus:
Thank you for the quick review!

PR #10100, opened 2018-09-22
Clear all find highlights when the findbar is closed (issue 7468)

### LINKED ISSUE #7468: Search highlighting not cleared after search is done
Link to PDF file (or attach file here): https://www.cca.edu/sites/default/files/pdf/08/word-to-pdf.pdf

Configuration:
- Web browser and its version: Chrome Version 54.0.2786.0 canary (64-bit)
- Operating system and its version: Win 10, 1511
- PDF.js version: 1.5.294
- Is an extension: Yes

1. Open pdf
2. Search for a string (e.g., pdf)
3. Close the search box

Text highlighting should be removed 

Text highlighting stays 

Snuffleupagus:
Please note that this will require a `mozilla-central` follow-up patch, e.g. looking [something like this](https://gist.github.com/Snuffleupagus/d5e9aa8fe824e2e34f85ccb1430ef3d7), in order for this to work in the built-in Firefox PDF viewer as well.

Fixes #7468.

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -53,9 +53,22 @@ class PDFFindController {
 
     this.reset();
 
+    eventBus.on('findbarclose', () => {
+      this._highlightMatches = false;
+
+      eventBus.dispatch('updatetextlayermatches', {
+        source: this,
+        pageIndex: -1,
+      });
+    });
+
     // Compile the regular expression for text normalization once.
     const replace = Object.keys(CHARACTERS_TO_NORMALIZE).join('');
     this._normalizationRegex = new RegExp(`[${replace}]`, 'g');
+  }
+
+  get highlightMatches() {
+    return this._highlightMatches;
   }
 
   get pageMatches() {
@@ -75,7 +88,7 @@ class PDFFindController {
   }
 
   reset() {
-    this.active = false; // If active, find results will be highlighted.
+    this._highlightMatches = false;
     this._pageMatches = [];
     this._pageMatchesLength = null;
     this._state = null;
@@ -353,7 +366,7 @@ class PDFFindController {
     const currentPageIndex = this._pdfViewer.currentPageNumber - 1;
     const numPages = this._pdfViewer.pagesCount;
 
-    this.active = true;
+    this._highlightMatches = true;
 
     if (this._dirtyMatch) {
       // Need to recalculate the matches, reset everything.
--- a/web/text_layer_builder.js
+++ b/web/text_layer_builder.js
@@ -56,6 +56,9 @@ class TextLayerBuilder {
     this.textLayerRenderTask = null;
     this.enhanceTextSelection = enhanceTextSelection;
 
+    this._boundEvents = Object.create(null);
+    this._bindEvents();
+
     this._bindMouse();
   }
 
@@ -329,6 +332,33 @@ class TextLayerBuilder {
 
     this.matches = this.convertMatches(pageMatches, pageMatchesLength);
     this.renderMatches(this.matches);
+  }
+
+  /**
+   * @private
+   */
+  _bindEvents() {
+    const { eventBus, _boundEvents, } = this;
+
+    _boundEvents.pageCancelled = (evt) => {
+      if (evt.pageNumber !== this.pageNumber) {
+        return;
+      }
+      if (this.textLayerRenderTask) {
+        console.error('TextLayerBuilder._bindEvents: `this.cancel()` should ' +
+          'have been called when the page was reset, or rendering cancelled.');
+        return;
+      }
+      // Ensure that all event listeners are cleaned up when the page is reset,
+      // since re-rendering will create new `TextLayerBuilder` instances and the
+      // number of (stale) event listeners would otherwise grow without bound.
+      for (const name in _boundEvents) {
+        eventBus.off(name.toLowerCase(), _boundEvents[name]);
+        delete _boundEvents[name];
+      }
+    };
+
+    eventBus.on('pagecancelled', _boundEvents.pageCancelled);
   }
 
   /**
--- a/web/text_layer_builder.js
+++ b/web/text_layer_builder.js
@@ -317,7 +317,7 @@ class TextLayerBuilder {
       clearedUntilDivIdx = match.end.divIdx + 1;
     }
 
-    if (this.findController === null || !this.findController.active) {
+    if (!this.findController || !this.findController.highlightMatches) {
       return;
     }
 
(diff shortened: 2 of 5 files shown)


web/pdf_find_controller.js
```diff
@@ -53,11 +53,19 @@ class PDFFindController {
 
     this.reset();
 
+    eventBus.on('findbarclose', () => {
```
timvandermeij: I think we can implement this feature without tracking any state in the find controller. We should be able to get rid of this `active` member variable if the text layer builder listens to the event on the event bus instead and tracks that member variable internally. After all it's a text layer state (whether or not to highlight the matches) and not find controller state (which only produces the matches and leaves the representation to the text layer builder).

In short, can we try removing all changes in this file and introducing `this._highlightMatches = true;` in the constructor of the text layer builder, which a listener for the `findbarclose` event that toggles this value, and then check this value at https://github.com/mozilla/pdf.js/pull/10100/files#diff-6a8d3c6754c77cfa0876314f0f54bed7R319?

Snuffleupagus: > After all it's a text layer state (whether or not to highlight the matches) and not find controller state (which only produces the matches and leaves the representation to the text layer builder).

While I cannot reasonably disagree with any of the above, given that `TextLayerBuilder` already have access to `PDFFindController` the current solution just like the simplest possible one :-)

> In short, can we try removing all changes in this file and introducing `this._highlightMatches = true;` in the constructor of the text layer builder, with a listener for the `findbarclose` event that toggles this value,

*There's a slight snag with this suggestion, which is why I went with the current solution:*
With the suggested changes there wouldn't be anything resetting `this._highlightMatches` for subsequent find operations, thus permanently breaking highlighting of matches once the findbar has been closed just *one* time.

In order for this to work, you'd basically need to start dispatching an event instead at line https://github.com/mozilla/pdf.js/blob/54d6c2436c17b6d0efc6fb8e36f36358eba1a9e6/web/pdf_find_controller.js#L356 and have the `TextLayerBuilder` instances listen for it in order to set `this._highlightMatches = true`. That feels slightly more involved, and there's also the question of what a good event name would be; `updatetextlayerhighlight` perhaps?

That should probably work too, although I've not tested it, but at the expense of a bit more event dispatching though. How do you think that we should proceed here?

Snuffleupagus: Having (quickly) tried using events instead, it unfortunately seems that the solution will need to become a bit more complex than current one for things to work out.
Most likely this can all be addressed in a good way, but it's not as easy as one would hope. Besides, there's also the issue described in https://github.com/mozilla/pdf.js/pull/10099#discussion_r219694481 which affects this PR as well.

web/text_layer_builder.js
```diff
@@ -331,6 +334,32 @@ class TextLayerBuilder {
...
+    _boundEvents.pageCancelled = (evt) => {
+      if (evt.pageNumber !== this.pageNumber) {
+        return;
+      }
+      if (this.textLayerRenderTask) {
+        console.error('TextLayerBuilder._bindEvents: `this.cancel()` should ' +
+                      'have been called when page rendering was cancelled.');
+        return;
+      }
+      for (const name in _boundEvents) {
```
timvandermeij: Could you add a comment here about why this is necessary? Basically a few lines of summary from the discussion in the other PR so we don't forget the reason why we added this and don't accidentally remove this later without a better alternative.

web/pdf_find_controller.js
timvandermeij: My idea was indeed that in the places where we manipulate the `this.active` member variable we'd use events instead, so where we set `this.active = true/false;` (doing a search sets it to `true` and closing the findbar sets it to `false`) we dispatch an event instead (`updatetextlayerhighlight` seems like a good name to me) and the text layer builder sets this value for the `this._highlightMatches` member variable that it tracks to display the highlighted matches or not. I didn't implement it, so it may indeed be a bit more complicated than I can foresee now, but I think that should work so we don't have to do anything in the find controller.

If this does turn out to be more complicated than what I described, can we at least rename `active` to `highlightMatches` or something similar? I find that a much more descriptive name and then we don't need the extra comment behind its declaration in the `reset` method.

Snuffleupagus: Having thought about all the various cases, the conclusion is unfortunately that we still need to track this state in `PDFFindController`.
***To summarize:*** There's simply no good way to relay the correct highlight state to *all* `TextLayerBuilder` instances, since they are created lazily; please read on for for additional details.

The important point here is that you need *one* source of truth, and not multiple, for whether or not find highlights should be displayed. There's only ever going to be one `PDFFindController` instance around, but an arbitrary number of `TextLayerBuilder` ones (i.e. one per page).

In particular, even attempting to track this in `TextLayerBuilder` is basically impossible when you consider all of the cases:
 - Assume that you set `this.highlightMatches = true` in the `TextLayerBuilder` constructor.
That means that for currently active pages, i.e. ones in the `PDFPageViewBuffer` cache, you'll receive the event to disable the highlights when closing the findbar. However when new pages are scrolled into view, and thus *new* `TextLayerBuilder` instances are created, those will incorrectly have highlights *enable* since they didn't exist at the point when the events (regardless if those are 'findbarclose' or 'updatetextlayerhighlight') were dispatched.

 - Assume instead that `this.highlightMatches = false` is set in the `TextLayerBuilder` constructor, and that an event is dispatched whenever searching starts.
Similar to the above, for *new* `TextLayerBuilder` instances created after searching started, no events are received and thus highlights will be permanently *disabled* in this case.

---

Furthermore please note that I choose to piggy-back on the 'updatetextlayermatches' event from your PR, for two reasons:
 - First of all, there's fewer events overall that way (and the name fits nicely here as well). Note that this PR alone means that each `TextLayerBuilder` instance registers two events, and with your PR that would have otherwise increase to a total of three events.

 - Second of all, by having only `PDFFindController` listen for 'findbarclose' events, there's no longer necessary to worry about the order in which the 'findbarclose' listeners are registered and run, thus guaranteeing that `highlightMatches` has been correctly set.

web/text_layer_builder.js
Snuffleupagus: Good point; please check that the added comment is clear enough.

web/pdf_find_controller.js
timvandermeij: Yes, the lazy loading is a really good point which indeed I didn't foresee in my idea. In that case the current code structure seems fine to me!

web/text_layer_builder.js
timvandermeij: Looks really good!

timvandermeij:
Overall this looks good to me, but I'll do another check in the coming days to make sure I didn't miss anything. We can probably land this first (and create a follow-up issue for `mozcentral`) and then I can rebase and remove the largest part of the first commit from my PR (only the findbar dependency part will probably remain).

Snuffleupagus:
> We can probably land this first (and create a follow-up issue for `mozcentral`) and then I can rebase and remove the largest part of the first commit from my PR (only the findbar dependency part will probably remain).

In all fairness, since your PR was submitted earlier and it's larger too, if you want we could probably land your PR first if you just include my 'pagecancelled' commit there too.

Anyway, I'm sorry for blocking your PR like this, but I really don't think it would have been good to (even temporarily) land code that would cause the number of `EventBus` listeners to grow arbitrarily large.

timvandermeij (approved):

timvandermeij:
Thank you for helping out with this!

PR #10111, opened 2018-09-28
Replace `String.prototype.substr()` occurrences with `String.prototype.substring()`

Snuffleupagus:
As outlined in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr, which refers to the ECMA-262 specification, using the `substr` function is advised against.

Hence this PR, which replaces all remaining `substr` occurrences with `substring` instead. Please refer to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr#Syntax respectively https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#Syntax for the differences between the two functions.

Note that in most cases in the code-base there's only one argument passed to `substr`, and those require no other changes except replacing "substr" with "substring". For the other cases, the `substr(start, length)` calls are changed to `substring(start, start + length)` instead.

### DIFF
--- a/docs/plugins/wintersmith-makerelative.coffee
+++ b/docs/plugins/wintersmith-makerelative.coffee
@@ -10,6 +10,6 @@ module.exports = (env, callback) ->
     depth = count(source, '/') # 1 being /
     ret = ""
     ret += "../" while depth = depth - 1
-    ret + dest.substr(1)
+    ret + dest.substring(1)
 
   callback()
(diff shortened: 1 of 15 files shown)


timvandermeij (approved):

timvandermeij:
Good find, thanks!

PR #6985, opened 2016-02-14
Remove XHTML leftovers from the viewer HTML files

timvandermeij:

PR #10113, opened 2018-09-29
the input label is not closed in the html file

forthealllight:
examples ——> mobile-viewer—————> viewer.html

in 68 line , the input label  is not closed


Snuffleupagus:
That's XHTML semantics, and the remaining occurrences were removed in PR #6985; hence this is invalid.

PR #10114, opened 2018-09-30
Update `{PDFLinkService, PDFDocumentProperties}.setDocument` to make the "url" parameter optional

Snuffleupagus:
This way the resetting of `PDFLinkService`/`PDFDocumentProperties` instances, as is done in `PDFViewerApplication.close`, only requires passing in *one* `null` argument instead of two.

---

*Also:* Prevent `getPDFFileNameFromURL` from breaking if the `url` parameter not a string


timvandermeij (approved):

timvandermeij:
Thank you!

PR #10117, opened 2018-09-30
Add support of Ink annotation

leblanc-simon:


timvandermeij (commented):
First of all, I would like to say that this is a really good patch, both in terms of code and functionality! Aside from the small comments below, please add a reference test so we can avoid regressions in the future. You can use the following test file I made earlier:

[Caret and ink.pdf](https://github.com/mozilla/pdf.js/files/2431762/Caret.and.ink.pdf)

You can do this in the same way as in https://github.com/mozilla/pdf.js/commit/ed8c0ebfa75bf39d9e8b350b7c0168c9e84d19f9. Download the file above and rename it to `annotation-caret-ink.pdf` and then place it in the `test/pdfs` folder. Add it to the `.gitignore` file and add an entry to the `test_manifest.json` file with the MD5 hash of the file.

Squash the commits into one when you're done with making these changes (refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for how to do that easily). I think it's good to merge after that!

leblanc-simon:
Thank you very much for your review and your PDF.
I've modified my code with your comments and squash my commits

timvandermeij (approved):
Good to go with the final comment addressed and squashed into the commit. I'll run the tests afterwards and if they all pass, we can merge this. Thank you!

leblanc-simon:
Sorry for forgetting :(
I also added the CSS rule as for other annotations
Thank you !

src/core/annotation.js
```diff
@@ -1013,6 +1016,32 @@ class PolygonAnnotation extends PolylineAnnotation {
...
 }
 
+class InkAnnotation extends Annotation {
+  constructor(parameters) {
+    super(parameters);
+
+    this.data.annotationType = AnnotationType.INK;
+
+    let dict = parameters.dict;
+    let rawInkLists = dict.getArray('InkList');
```
Snuffleupagus: Using "raw" in this name is somewhat misleading, since `getArray` will dereference the top level elements when fetching the data (which is very different to `getRaw`); please see the `Dict` class for additional details: https://github.com/mozilla/pdf.js/blob/master/src/core/primitives.js#L114-L134.

Could another name be used instead?
As always, naming things is one of the hard parts of computer science, but how about e.g. `originalInkLists` or `basicInkLists` or something instead!?

src/core/annotation.js
```diff
@@ -1013,6 +1016,32 @@ class PolygonAnnotation extends PolylineAnnotation {
...
+    for (let i = 0, ii = rawInkLists.length; i < ii; ++i) {
+      // The raw ink lists array contains arrays of numbers representing
+      // the alternating horizontal and vertical coordinates, respectively,
+      // of each vertex. Convert this to an array of objects with x and y
+      // coordinates.
+      this.data.inkLists.push([]);
+      for (let j = 0, jj = rawInkLists[i].length; j < jj; j += 2) {
+        this.data.inkLists[i].push({
+          x: rawInkLists[i][j],
+          y: rawInkLists[i][j + 1],
```
Snuffleupagus: Either of these could theoretically be indirect objects, and that needs to be handled to prevent future bugs (since `getArray` for good reason only dereferences top level elements).
So define `const xref = parameters.xref;` *above* the loops, and then change these lines to:
```javascript
  x: xref.fetchIfRef(rawInkLists[i][j]),
  y: xref.fetchIfRef(rawInkLists[i][j + 1]),
```

leblanc-simon:
Yes, naming is hard :)
Thanks for your reviews !

timvandermeij:

leblanc-simon:
A huge thanks for your help !

PR #10119, opened 2018-10-01
Attempt to simplify the `fileattachmentannotation` event dispatching

Snuffleupagus:
This attempts to reduced the level of indirection, and the amount of code, when dispatching `fileattachmentannotation` events, by removing the `PDFLinkService.onFileAttachmentAnnotation` method and just accessing `PDFLinkService.eventBus` directly in the `FileAttachmentAnnotationElement` constructor.
Given that other properties, such as `externalLinkTarget`/`externalLinkRel`, are already being accessed directly this pattern seems fine here as well.


Snuffleupagus:
Hopefully just a Windows intermittent...

/botio-windows test

timvandermeij (approved):

Issue #6810, opened 2015-12-27
Support annotations without appearance streams


timvandermeij:
Most annotations, even unsupported ones, render just fine in PDF.js because of their appearance stream. However, there exist PDF files with annotations that do not have an appearance stream (even though this is a deprecated practice). In the latter case, PDF.js displays nothing. We need to have a fallback to a default appearance stream in that case, or create an appearance stream ourselves using the data present for the annotation (such as quad points and color). Below is a list of known PDF files that contain annotations without appearance streams, for which the types are listed.
- <s>http://pdfkit.org/docs/guide.pdf (Underline, Highlight and StrikeOut)</s>
- <s>http://www.plpdf.com/examples/PLPDF-SDK/Simple_Examples/37_Markup_annotation/markup.pdf (Underline, Highlight and StrikeOut)</s>
- <s>https://drive.google.com/file/d/0B6YhVF9p-y2AU0RlYjRKakFQRjQ/view?usp=sharing (Highlight, from #6382)</s>
- https://drive.google.com/open?id=0B6YhVF9p-y2Ad3VsWDRHRkRxY2c (FreeText, from #6246)
- https://drive.google.com/file/d/0B-_5BgJGB33Ea2dQZnpUY2FvYkk/view (FreeText, from #6564)
- http://mirrors.rit.edu/CTAN/macros/latex/contrib/pdfcomment/doc/example.pdf (practically all annotation types, from #1973)

THausherr:
http://www.pdfill.com/example/pdf_commenting_new.pdf

fancycarp:
Waiting for this feature..^^

THausherr:
This one has diagonal annotations. A strategy that assumes them to be horizontal will result in weird effects (try showing the file in chrome in the built-in viewer or on an Android phone with the google viewer)
https://bugs.ghostscript.com/show_bug.cgi?id=693664

burtonator:
This is still a pretty serious issue.  Apparently poppler, one of the main PDF libraries used by Okular, etc, creates annotations without appearance streams.

When loading them into PDF.js its impossible to see the annotations.

timvandermeij:
@edsontrick The question is not related to this particular issue. Please post it on IRC instead of on an unrelated issue.

MzunguWaNjombe:
> Somebody should also create an issue in poppler. The /AP is required in PDF 2.0 except for Popup, Projection or Link.

The corresponding issue for Poppler seems to be this one:
[...]

Issue #2966, opened 2013-03-19
Support for Rich Text annotations


jan-tosovsky-cz:
In PDF annotations an HTML code can be used. Such annotations are not processed correctly in PDF.js (nothing is displayed).

qwer1304:
The situation is better in version 2.12.211 - some formatting is supported (e.g., bold, italics).
However, it seems that text size isn't respected as the following images show:
First from Acrobat Reader:
![image](https://user-images.githubusercontent.com/2311076/143679464-7d51ebaf-11a6-43a1-9d39-581f9b988653.png)

Then from pdf.js:
![image](https://user-images.githubusercontent.com/2311076/143679508-f7133fe1-76c7-439c-bc00-61edec86467c.png)

Notice how the superscript and subscript sizes is screwed up.

calixteman:
@qwer1304, please open a new issue with the pdf containing those annotations.

PR #10122, opened 2018-10-02
[WIP] Freetext annotation support

leblanc-simon:
Hi all,

it's a first implementation of FreeText annotation support. You can test with this PDF (found in #6810 ) : https://drive.google.com/open?id=0B6YhVF9p-y2Ad3VsWDRHRkRxY2c

* I moved setColor body to `src/shared/util.js` because I use it twice (and not only for this.color)
* I haven't found the right way to parse defaultApparence with `evaluator.getOperatorList`, so I code a creepy code with regexp (it's the main reason I set the merge request to WIP). Do you have any tips to handle this properly ?
* this implementation allows manage :
  * the case where the annotation is already rendered via stream appearances (only the popup is rended)
  * the case where the annotation contains richText
  * the last case : simple text annotation

Each advice is welcome :)

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -20,7 +20,6 @@ import {
 } from '../shared/util';
 import { Catalog, FileSpec, ObjectLoader } from './obj';
 import { Dict, isDict, isName, isRef, isStream } from './primitives';
-import { ColorSpace } from './colorspace';
 import { OperatorList } from './operator_list';
 import { Stream } from './stream';
 
@@ -71,6 +70,9 @@ class AnnotationFactory {
 
       case 'Text':
         return new TextAnnotation(parameters);
+
+      case 'FreeText':
+        return new FreeTextAnnotation(parameters);
 
       case 'Widget':
         let fieldType = getInheritableProperty({ dict, key: 'FT', });
@@ -285,36 +287,7 @@ class Annotation {
    *                        4 (CMYK) elements
    */
   setColor(color) {
-    let rgbColor = new Uint8ClampedArray(3);
-    if (!Array.isArray(color)) {
-      this.color = rgbColor;
-      return;
-    }
-
-    switch (color.length) {
-      case 0: // Transparent, which we indicate with a null value
-        this.color = null;
-        break;
-
-      case 1: // Convert grayscale to RGB
-        ColorSpace.singletons.gray.getRgbItem(color, 0, rgbColor, 0);
-        this.color = rgbColor;
-        break;
-
-      case 3: // Convert RGB percentages to RGB
-        ColorSpace.singletons.rgb.getRgbItem(color, 0, rgbColor, 0);
-        this.color = rgbColor;
-        break;
-
-      case 4: // Convert CMYK to RGB
-        ColorSpace.singletons.cmyk.getRgbItem(color, 0, rgbColor, 0);
-        this.color = rgbColor;
-        break;
-
-      default:
-        this.color = rgbColor;
-        break;
-    }
+    this.color = Util.createRgbColor(color);
   }
 
   /**
@@ -897,6 +870,24 @@ class TextAnnotation extends Annotation {
                        parameters.dict.get('Name').name : 'Note';
     }
     this._preparePopup(parameters.dict);
+  }
+}
+
+class FreeTextAnnotation extends Annotation {
+  constructor(parameters) {
+    super(parameters);
+
+    this.data.annotationType = AnnotationType.FREETEXT;
+
+    let dict = parameters.dict;
+
+    this.data.textColor = Util.createRgbColor(dict.getArray('TextColor'));
+    this.data.textStyle = dict.get('DS');
+    this.data.fontSize = dict.get('FontSize');
+    this.data.opacity = dict.get('CA') || 1;
+    this.data.richText = dict.get('RC');
+
+    this._preparePopup(dict);
   }
 }
 
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -945,7 +945,7 @@ class FreeTextAnnotation extends Annotation {
             data.fontRefName = a[i][0];
             data.fontSize = a[i][1];
             break;
-          case OPS.setGrayFill:
+          case OPS.setFillGray:
           case OPS.setFillRGBColor:
             data.textColor = a[i];
             break;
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
> I moved setColor body to `src/shared/util.js` because I use it twice (and not only for this.color)

Since you only appear to be using that function from `src/core/annotation.js`, please place it as a helper function *in that file* instead, since code in `src/shared/util.js` gets duplicated in both the main *and* worker threads which is totally unnecessary in this case (it just increases the total size of the PDF.js library).

src/core/annotation.js
```diff
@@ -897,6 +870,45 @@ class TextAnnotation extends Annotation {
...
+    super(parameters);
+
+    this.data.annotationType = AnnotationType.FREETEXT;
+    let dict = parameters.dict;
+
+    this.data.textColor = Util.createRgbColor(dict.getArray('TextColor'));
+    this.data.textStyle = dict.get('DS');
+    this.data.fontSize = dict.get('FontSize');
+    this.data.opacity = dict.get('CA');
+    this.data.richText = dict.get('RC');
```
timvandermeij: I think it's better to not include rich text processing in this patch. There is a separate issue for that at #2966 and we have to start simple here because fetching the appearance stream is already a bit difficult. Keeping the patch small will make it easier to catch any mistakes and to review, and rich text support is a nice-to-have that can be done in a follow-up patch.

timvandermeij (commented):
Good start!

In order to correctly implement this, we need to parse the appearance stream. To do this, we recently merged #9822 so the annotation code is asynchronous, which is a necessity for `evaluator.getOperatorList` to work correctly. In https://github.com/mozilla/pdf.js/pull/8625/files, an attempt was made to get the appearance data out of the stream, but it was never finished. It could be used as inspiration though for how to do this, because it's pretty close to how I think it should work. The most important logic is in https://github.com/mozilla/pdf.js/pull/8625/files#diff-18ca06e2bd5be4c9132b51f78388b8f0R739 and this should work just fine, but the only thing we haven't had time for yet to look into is if we can refactor `evaluator.getOperatorList` to work without a worker task, since that would make the logic must simpler and avoid dependencies.

I would therefore suggest to use the logic in the `parseAppearance` function from https://github.com/mozilla/pdf.js/pull/8625/files#diff-18ca06e2bd5be4c9132b51f78388b8f0R739 as a baseline for how it should work and find a way to use `evaluator.getOperatorList` nicely without passing in a bunch of new parameters like that patch did. I do think it's possible, it just needs a bit of research. However, once that is possible, it unlocks quite a bit of possibilities for other parts of the annotation layer too, so it's definitely useful.

leblanc-simon:
Hi ! Thanks for your comment. I move the createRgbColor function and remove for now the richText support.
I will read all your links to improve this PR

src/core/annotation.js
```diff
@@ -900,6 +902,61 @@ class TextAnnotation extends Annotation {
...
+      resources: resourcesFonts,
+      operatorList: opList,
+    }).then(() => {
+      let a = opList.argsArray;
+      let i;
+      for (i = 0; i < opList.fnArray.length; i++) {
+        let fn = opList.fnArray[i];
+        switch (fn | 0) {
+          case OPS.setFont:
+            data.fontRefName = a[i][0];
```
timvandermeij: You don't seem to be using this anywhere in the display layer, so we always fall back to the default font. I think you can implement this similar to https://github.com/mozilla/pdf.js/blob/ff2df9c5b64a700e06694ff3375cd54c10ccedf4/src/display/annotation_layer.js#L451-L490, which you might also need to extract in a separate function then.

src/core/annotation.js
```diff
@@ -900,6 +902,61 @@ class TextAnnotation extends Annotation {
...
+            data.fontSize = a[i][1];
+            break;
+          case OPS.setGrayFill:
```
timvandermeij: `setGrayFill` doesn't seem to exist, while `setFillGray` does (https://github.com/mozilla/pdf.js/blob/9c764da8af99a2e896e151108d74a5660f5e9332/src/shared/util.js#L231). Note that they have a different number of arguments (https://github.com/mozilla/pdf.js/blob/b40fb3814a05e7a9c489d995d46b3239e5936d29/src/core/evaluator.js#L2892-L2894) so please check if this works well. If you have not found a case of `setFillGray` in your test documents, let's just remove it from this patch.

src/display/annotation_layer.js
```diff
@@ -380,6 +383,82 @@ class TextAnnotationElement extends AnnotationElement {
...
+                                      data.textColor[2] | 0);
+      style += 'color: ' + textColor + ';';
+    }
+
+    if (data.fontSize) {
+      style += 'font-size: ' + data.fontSize + 'pt;';
+    }
+
+    div.setAttribute('style', style);
+
```
timvandermeij: You can remove this newline to make it a bit more compact.

src/display/annotation_layer.js
```diff
@@ -380,6 +383,82 @@ class TextAnnotationElement extends AnnotationElement {
...
+    div.innerText = data.contents;
+
```
timvandermeij: You can remove this newline to make it a bit more compact.

timvandermeij (commented):
You found a pretty clever solution to the appearance stream parsing problem; nice work!

Please add the test file that you linked as a reference test, just like you did for your ink annotation pull request, so we can avoid regressions.

src/core/annotation.js
```diff
@@ -900,6 +902,61 @@ class TextAnnotation extends Annotation {
...
+  getOperatorList(evaluator, task, renderForms) {
+    if (renderForms || this.appearance) {
+      return super.getOperatorList(evaluator, task, renderForms);
+    }
+
+    let data = this.data;
+
+    let opList = new OperatorList();
+    let appearanceStream = new Stream(stringToBytes(data.defaultAppearance));
+    let resourcesFonts = this.pdfManager.pdfDocument.catalog.fontCache;
```
Snuffleupagus: Sorry, but there's no way that this works correctly!
You're trying to use a `RefSetCache`, but what's expected here is a `Dict`; note how the `this.loadResources` method is used in other parts of this file to fetch resources in a correct way.

escapewindow:
@leblanc-simon Is this still wanted? If so, are you planning on working on it again? If not, do you mind if I work on it?

Assuming the answers are yes, not anytime soon, and no, I rebased and unbitrotted your work, then tried to address your review comments [here](https://github.com/mozilla/pdf.js/compare/master...escapewindow:freetext?expand=1). That helped me learn a bit about annotations; thank you!

timvandermeij:
Closing since this PR is going to be continued in #12093.

PR #10126, opened 2018-10-03
[Regression] Restore the ability to start searching before a document has loaded, and ignore searches for previously opened documents (PR 10099 follow-up)

Snuffleupagus:
*Please refer to the individual commit messages.*

/cc @timvandermeij Sorry to bother you, but do you have time to review this?

PR #10128, opened 2018-10-03
Make `PDFFindController` less confusing to use, by allowing searching to start when `setDocument` is called

Snuffleupagus:
*This patch is based on something that I noticed while working on PR #10126.*

The recent re-factoring of `PDFFindController` brought many improvements, among those the fact that access to `BaseViewer` is no longer required. However, with these changes there's one thing which now strikes me as not particularly user-friendly[1]: The fact that in order for searching to actually work, `PDFFindController.setDocument` must be called *and* a 'pagesinit' event must be dispatched (from somewhere).

For all other viewer components, calling the `setDocument` method[2] is enough in order for the component to actually be usable.
The `PDFFindController` thus stands out quite a bit, and it also becomes difficult to work with in any sort of custom implementation. For example: Imagine someone trying to use `PDFFindController` separately from the viewer[3], which *should* now be relatively simple given the re-factoring, and thus having to (somehow) figure out that they'll also need to manually dispatch a 'pagesinit' event for searching to work.

Note that the above even affects the unit-tests, where an out-of-place 'pagesinit' event is being used.
To attempt to address these problems, I'm thus suggesting that *only* `setDocument` should be used to indicate that searching may start. For the default viewer and/or the viewer components, `BaseViewer.setDocument` will now call `PDFFindController.setDocument` when the document is ready, thus requiring no outside configuration anymore[4]. For custom implementation, and the unit-tests, it's now as simple as just calling `PDFFindController.setDocument` to allow searching to start.

---
[1] I should have caught this during review of PR #10099, but unfortunately it's sometimes not until you actually work with the code in question that things like these become clear.

[2] Assuming, obviously, that the viewer component in question actually implements such a method :-)

[3] There's even a very recent issue, filed by someone trying to do just that.

[4] Short of providing a `PDFFindController` instance when creating a `BaseViewer` instance, of course.

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -99,6 +99,7 @@ class PDFFindController {
       return;
     }
     this._pdfDocument = pdfDocument;
+    this._firstPageCapability.resolve();
   }
 
   executeCommand(cmd, state) {
@@ -110,7 +111,7 @@ class PDFFindController {
     this._state = state;
     this._updateUIState(FindState.PENDING);
 
-    this._firstPagePromise.then(() => {
+    this._firstPageCapability.promise.then(() => {
       if (!this._pdfDocument ||
           (pdfDocument && this._pdfDocument !== pdfDocument)) {
         // If the document was closed before searching began, or if the search
@@ -160,17 +161,7 @@ class PDFFindController {
     clearTimeout(this._findTimeout);
     this._findTimeout = null;
 
-    this._firstPagePromise = new Promise((resolve) => {
-      const onPagesInit = () => {
-        if (!this._pdfDocument) {
-          throw new Error(
-            'PDFFindController: `setDocument()` should have been called.');
-        }
-        this._eventBus.off('pagesinit', onPagesInit);
-        resolve();
-      };
-      this._eventBus.on('pagesinit', onPagesInit);
-    });
+    this._firstPageCapability = createPromiseCapability();
   }
 
   _normalize(text) {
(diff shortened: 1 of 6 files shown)


web/pdf_find_controller.js
```diff
@@ -158,13 +159,7 @@ class PDFFindController {
...
     this._findTimeout = null;
 
-    this._firstPagePromise = new Promise((resolve) => {
-      const eventBus = this._eventBus;
-      eventBus.on('pagesinit', function onPagesInit() {
-        eventBus.off('pagesinit', onPagesInit);
-        resolve();
-      });
-    });
+    this._firstPageCapability = createPromiseCapability();
```
timvandermeij: I'm wondering if `_firstPageCapability` is a good name now that we're not actually waiting until the `pagesinit` event is emitted. Maybe `_initializationCapability` or something similar would be better; what do you think?

timvandermeij (commented):
Looks good with a rebase.

web/pdf_find_controller.js
Snuffleupagus: > I'm wondering if `_firstPageCapability` is a good name now that we're not actually waiting until the `pagesinit` event is emitted.

The answer is probably both yes and no at this point, depending on your perspective; naming thing is hard :-)

The name is still very much relevant for the default use-case, i.e. the viewer/components, since `PDFFindController.setDocument` is still called *after* the first page has been resolved in `BaseViewer.setDocument`.
However, for the general case where `PDFFindController` could now be used standalone from the viewer, the name is obviously somewhat strange.

Given that it's tracking internal state anyway, I don't know if the name matter too much though and keeping the current one is also more inline with the default viewer use-case.

Snuffleupagus: > I'm wondering if `_firstPageCapability` is a good name now that we're not actually waiting until the `pagesinit` event is emitted.

The answer is probably both yes and no at this point, depending on your perspective; naming things is hard :-)

The name is still very much relevant for the default use-case, i.e. the viewer/components, since `PDFFindController.setDocument` is still called *after* the first page has been resolved in `BaseViewer.setDocument`.
However, for the general case where `PDFFindController` could now be used standalone from the viewer, the name is obviously somewhat strange.

Given that it's tracking internal state anyway, I don't know if the name matter too much though and keeping the current one is also more inline with the default viewer use-case.

timvandermeij (approved):

timvandermeij:
Good improvement!

PR #10129, opened 2018-10-04
improve readability of CanvasGraphics's and SVGGraphics's constructPath

MeiKatz:
- extract sub-programs into own functions:
  - `rectangle(x, y, width, height)`
  - `moveTo(x, y)`
  - `lineTo(x, y)`
  - `curveTo(cp1x, cp1y, cp2x, cp2y, x, y)`
  - `curveTo2(cp2x, cp2y, x, y)`
  - `curveTo3(cp1x, cp1y, x, y)`
- align CanvasGraphics's and SVGGraphics's constructPath definition
  as much as possible

## Reason
`constructPath` has now improved readability and an user searching for information about the api and the way how `PDF.js` works can now easier see the parallels between the implementation for Canvas and for SVGs.

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -939,58 +939,62 @@ SVGGraphics = (function SVGGraphicsClosure() {
       var current = this.current;
       var x = current.x, y = current.y;
       current.path = this.svgFactory.createElement('svg:path');
-      var d = [];
       var opLength = ops.length;
 
       for (var i = 0, j = 0; i < opLength; i++) {
         switch (ops[i] | 0) {
           case OPS.rectangle:
-            x = args[j++];
-            y = args[j++];
-            var width = args[j++];
-            var height = args[j++];
-            var xw = x + width;
-            var yh = y + height;
-            d.push('M', pf(x), pf(y), 'L', pf(xw), pf(y), 'L', pf(xw), pf(yh),
-                   'L', pf(x), pf(yh), 'Z');
+            this.rectangle(
+              args[j++], // x
+              args[j++], // y
+              args[j++], // width
+              args[j++]  // height
+            );
             break;
           case OPS.moveTo:
-            x = args[j++];
-            y = args[j++];
-            d.push('M', pf(x), pf(y));
+            this.moveTo(
+              args[j++], // x
+              args[j++]  // y
+            );
             break;
           case OPS.lineTo:
-            x = args[j++];
-            y = args[j++];
-            d.push('L', pf(x), pf(y));
+            this.lineTo(
+              args[j++], // x
+              args[j++]  // y
+            );
             break;
           case OPS.curveTo:
-            x = args[j + 4];
-            y = args[j + 5];
-            d.push('C', pf(args[j]), pf(args[j + 1]), pf(args[j + 2]),
-                   pf(args[j + 3]), pf(x), pf(y));
-            j += 6;
+            this.curveTo(
+              args[j++], // cp1x
+              args[j++], // cp1y
+              args[j++], // cp2x
+              args[j++], // cp2y
+              args[j++], // x
+              args[j++]  // y
+            );
             break;
           case OPS.curveTo2:
-            x = args[j + 2];
-            y = args[j + 3];
-            d.push('C', pf(x), pf(y), pf(args[j]), pf(args[j + 1]),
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
Won't adding a whole bunch of (strictly unnecessary) methods add an unnecessary level of indirection and quite possibly even slow down rendering by requiring *many* more function calls then the current code?

src/display/svg.js
```diff
@@ -1001,6 +1005,90 @@ SVGGraphics = (function SVGGraphicsClosure() {
       current.setCurrentPoint(x, y);
     },
 
+    rectangle: function SVGGraphics_rectangle(x, y, width, height) {
+      var current = this.current;
+
+      if (current.path) {
```
MeiKatz: Personally not happy about the checks in every function. Maybe we could create `current.path` at the start and ensure that it exists.

MeiKatz:
@Snuffleupagus Yes, they're *strictly* unnecessary but they help getting what the code does. Also the methods are defined as single ops and maybe they may occur standing alone (not sure about this because I could not find any documentation about this behaviour).
I don't think that it will slow down rendering speed that much. But: my main point is that you can see with one look what the functions do. That is not the case in the current implementation IMHO.

Snuffleupagus:
*It can often, in general, be illuminating to look at the history of a particular piece of code.*

In this case, you're essentially proposing to revert part of PR #4683, whose purpose was to improve the rendering performance of in particular large/complex PDF files. In that PR, readability was thus purposely traded for *perhaps* slightly less easily readable code.

> ... maybe they may occur standing alone (not sure about this because I could not find any documentation about this behaviour).

I don't think that can happen, given the changes made in PR #4683.

> I don't think that it will slow down rendering speed that much.

Rather than guessing/assuming, please actually test the impact of your proposed changes!
Most likely, you'll have a slightly easier time convincing a reviewer to accept your patch if you can clearly demonstrate that it doesn't cause any performance regressions.

Please refer to https://github.com/mozilla/pdf.js/wiki/Benchmarking-your-changes for information about benchmarking, and perhaps https://github.com/mozilla/pdf.js/wiki/Contributing is useful too for general information about the test-suite.

The following PDF files contains *lots* of rendering operations, and would probably be good examples to use when benchmarking:
https://github.com/mozilla/pdf.js/files/1522715/wuppertal_2012.pdf
https://www.prorail.nl/sites/default/files/spoorkaart_prorail_april_2013.pdf

MeiKatz:
@Snuffleupagus Maybe I will turn this topic into another direction and write an example with improved readability while ignoring performance issues. So beginners would have a good point to start and can from there understand other implementation like the one for SVG or canvas.

timvandermeij:
I think there are two problems here.

The most important one is that the increased number of function calls will have a bad effect on performance, and eliminating them is one of the reasons why #4683 worked so well.

Moreover, I don't think that the current code is hard to understand and I don't feel that this patch makes it any clearer. For instance, all function names are the same as the operator, so you get `this.rectangle(...)` just below `case OPS.rectangle` and that doesn't give any more context. The same goes for the parameters and variable names: previously it would be `x = args[j++];` whereas now it's `args[j++]; // x`, which is equivalent and also doesn't provide more context. The remaining comments are of the form `// ensure height is not null` just above `if (height === 0)`, which is redundant.

In conclusion, I'm closing this since there does not appear to be much benefit in merging this unfortunately. We can definitely consider a patch though that adds inline documentation as comments, but then they do have to add context and not just explain the code.

MeiKatz:
I solved my readability issues by writing my own PDF.js core documentation that can be found [here](https://github.com/MeiKatz/pdfjs-docs/blob/master/README.md).

timvandermeij:
Thank you for your contribution! I think it can definitely help in understanding the core of PDF.js and I think it's also much better suited on a specialized wiki page than in inline comments. I have linked the documentation on the https://github.com/mozilla/pdf.js/wiki/Additional-Learning-Resources page.

MeiKatz:
> _Edit: Small feedback, it looks like the setTextRenderingIntent link in the overview does not work when clicked, while the others do._

I fixed it. The name `setTextRenderingIntent` was wrong because it is `setTextRenderingMode`.

I will try to improve my documentation when I have the time for it.

PR #10133, opened 2018-10-05
Set returnValues.suggestedLength to Content-Length if integer

### LINKED ISSUE #9103: total is undefined in LoadingTask.onProgress
https://jsfiddle.net/p3ybwp7d/1/

Configuration:
- Web browser and its version: *Google Chrome - Version 62.0.3202.75*
- Operating system and its version: Mac OS 10.13
- PDF.js version: *1.10.97*
- Is a browser extension: *No*

*Create loading task and add onProgress callback:*
```
let loadingTask: any = PDFJS.getDocument(this.src);

loadingTask.onProgress = (progressData) => {
  // progressData won't contain "total", only "loaded"
};
```

What is the expected behaviour? (add screenshot)
In previous versions, `onProgress` did return both `total` and `loaded`.

`total` field is undefined in `loadingTask.onProgress` callback.

https://jsfiddle.net/p3ybwp7d/1/

kevinleedrum:
Fixes #9103 by setting `returnValues.suggestedLength` to the value of `Content-Length` (provided it is an integer) even if returning early.

### DIFF
--- a/src/display/network_utils.js
+++ b/src/display/network_utils.js
@@ -27,6 +27,20 @@ function validateRangeRequestCapabilities({ getResponseHeader, isHttp,
     allowRangeRequests: false,
     suggestedLength: undefined,
   };
+
+  let length = parseInt(getResponseHeader('Content-Length'), 10);
+  if (!Number.isInteger(length)) {
+    return returnValues;
+  }
+
+  returnValues.suggestedLength = length;
+
+  if (length <= 2 * rangeChunkSize) {
+    // The file size is smaller than the size of two chunks, so it does not
+    // make any sense to abort the request and retry with a range request.
+    return returnValues;
+  }
+
   if (disableRange || !isHttp) {
     return returnValues;
   }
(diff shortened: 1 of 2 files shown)


timvandermeij:
Looks good. Could you squash the commits into one for review? Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for how to do that easily if you're not familiar with the process.

Snuffleupagus:
What about the final point from https://github.com/mozilla/pdf.js/issues/9103#issuecomment-363436612, is that not necessary after all?
> [...] so perhaps it makes sense to unconditionally use the value of the `Content-Length` header (unless `Transfer-Encoding` is specified but not starting with `identity`).

kevinleedrum:
> What about the final point from [#9103 (comment)](https://github.com/mozilla/pdf.js/issues/9103#issuecomment-363436612), is that not necessary after all?
> 
> > [...] so perhaps it makes sense to unconditionally use the value of the `Content-Length` header (unless `Transfer-Encoding` is specified but not starting with `identity`).

I believe the `Content-Length` should not be present anyway if `Transfer-Encoding` is specified.  https://tools.ietf.org/html/rfc7230#section-3.3.2

timvandermeij:
@Rob--W Time permitting, would you be willing to review this?

Rob--W (commented):
Looks good in general. The `defaultValues` variable in the test is now used only once, and it would make more sense to replace that only use with the value itself.

kevinleedrum:
>The `defaultValues` variable in the test is now used only once, and it would make more sense to replace that only use with the value itself.

Done. 👍

Rob--W (approved):
Thanks for the change. This is good to merge if the tests pass.

timvandermeij:
Thank you!

PR #10136, opened 2018-10-05
Remove the unused `mainContainer` parameter from the `Toolbar` constructor

Snuffleupagus:
Looking at the history of this code, this parameter has never been used.
I'm guessing that most likely the code in `web/toolbar.js` began life as a copy of `web/secondary_toolbar.js`, which would probably explain why that parameter exists.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -311,8 +311,8 @@ let PDFViewerApplication = {
     });
     this.findController = findController;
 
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good find; thank you for making this a bit easier!

PR #10138, opened 2018-10-05
Ensure that `Toolbar.setPageScale` always sets the `pageScaleValue` property to a valid value

Snuffleupagus:
Rather than having every invocation of `Toolbar._updateUIState` compute a valid `pageScaleValue`, it seems easier to simply ensure that it happens when the value is actually updated.


timvandermeij (approved):

timvandermeij:
Good improvement!

PR #10139, opened 2018-10-05
Ensure that the `PDFDocumentLoadingTask` is rejected when "setting up fake worker" failed (issue 10135)

### LINKED ISSUE #10135: Failing to load pdf.worker.js does not throw an error in pdfjs.getDocument

Configuration:
- PDF.js version: 2.0.550

1. Configure PDF.js badly so that pdf.worker.js is a 404 page
2. Run pdfjs.getDocument

1. getDocument Promise gets rejected

What went wrong?

1. Since creating worker failed, setupFakeWorker is called.
2. Inside it, `setupFakeWorkerGlobal()` never resolves, nor never throws an error, if something bad happens with fakeWorker. So adding `.catch()` to `setupFakeWorkerGlobal()` or even wrapping it with `try {} catch {}` does not make any difference.
3. getDocument Promise is never resolved, does not throw an error.

Snuffleupagus:
This should, hopefully, cover all the possible ways[1] in which "fake workers" are loaded. Given the different code-paths, adding unit-tests might not be that simple. ***Edit:*** Also given that "fake workers" are only intended as a fallback anyway, and aren't really tested currently, the patch is hopefully OK even without tests.

Note that in order to make this work, the various `fakeWorkerFilesLoader` functions were converted to return `Promises`. ***Edit:*** Smaller diff with https://github.com/mozilla/pdf.js/pull/10139/files?w=1

Fixes #10135.

---
[1] Unfortunately there's lots of them, for various build targets and configurations.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -45,9 +45,9 @@ const pdfjsFilePath =
   typeof document !== 'undefined' && document.currentScript ?
     document.currentScript.src : null;
 
-var fakeWorkerFilesLoader = null;
-var useRequireEnsure = false;
+let fakeWorkerFilesLoader = null;
 if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
+  let useRequireEnsure = false;
   // For GENERIC build we need to add support for different fake file loaders
   // for different frameworks.
   if (typeof window === 'undefined') {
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
I think that this should now account for all the different cases.

/botio unittest

timvandermeij (approved):

timvandermeij:
Thank you for improving the error handling here!

PR #10140, opened 2018-10-07
Combine the `pdfjsFilePath` and fallback `workerSrc` handling in `src/display/api.js`

Snuffleupagus:
With the way that the `getWorkerSrc()` helper function is implemented now, there's no longer a particularly strong reason for keeping the global `pdfjsFilePath` variable around.
With this patch the fallback `workerSrc` will thus, assuming is wasn't already set, be set to the "pdfjsFilePath" which simplifies the `getWorkerSrc()` function and reduces the amount of global state.

Finally, the global `workerSrc` variable was renamed to prevent shadowing.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -34,16 +34,10 @@ import { Metadata } from './metadata';
 import { PDFDataTransportStream } from './transport_stream';
 import { WebGLContext } from './webgl';
 
-var DEFAULT_RANGE_CHUNK_SIZE = 65536; // 2^16 = 65536
+const DEFAULT_RANGE_CHUNK_SIZE = 65536; // 2^16 = 65536
 
 let isWorkerDisabled = false;
-let workerSrc;
-
-const pdfjsFilePath =
-  typeof PDFJSDev !== 'undefined' &&
-  PDFJSDev.test('PRODUCTION && !(MOZCENTRAL || FIREFOX)') &&
-  typeof document !== 'undefined' && document.currentScript ?
-    document.currentScript.src : null;
+let fallbackWorkerSrc;
 
 let fakeWorkerFilesLoader = null;
 if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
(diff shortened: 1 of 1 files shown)


src/display/api.js
```diff
@@ -94,6 +88,15 @@ if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
     });
   }) : null;
 }
+if (typeof PDFJSDev !== 'undefined' &&
+    PDFJSDev.test('PRODUCTION && !(MOZCENTRAL || FIREFOX)') &&
+    !fallbackWorkerSrc && typeof document !== 'undefined') {
+  const pdfjsFilePath = document.currentScript && document.currentScript.src;
```
Snuffleupagus: I initially considered inlining the [`document.currentScript` polyfill](https://github.com/mozilla/pdf.js/blob/master/src/shared/compatibility.js#L57-L74) here, such that it could be removed since `document.currentScript` is used *once*, but that might not be a desired change!?

src/display/api.js
```diff
@@ -94,6 +88,15 @@ if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
     });
   }) : null;
 }
+if (typeof PDFJSDev !== 'undefined' &&
+    PDFJSDev.test('PRODUCTION && !(MOZCENTRAL || FIREFOX)') &&
```
Snuffleupagus: Also, I'm kind of wondering if this condition needs to be so complicated, since it seems that `PDFJSDev.test('GENERIC')` really ought to be sufficient?

src/display/api.js
timvandermeij: Indeed, let's keep it in `compatibility.js` even if it's only used once because I think it's better to consolidate all polyfills in one place as much as possible so we can easily go through them and remove them when they are no longer relevant.

src/display/api.js
timvandermeij: I'm not 100% sure I must say, but I also they those are equivalent.

Snuffleupagus: Given that this is only a fallback anyway, and that (as mentioned elsewhere) `GlobalWorkerOptions.workerSrc` should always be set, I cannot see the harm in changing this.
I've also quickly tested the `CHROMIUM` build, and given that the `workerSrc` is being specified through the viewer everything works fine.

timvandermeij (approved):

timvandermeij:

PR #10141, opened 2018-10-09
Add annotation without stream appearance support

leblanc-simon:
This PR display the annotation when there is no appearance stream (issue : #6810)
Annotation supported : 
* Line
* Square
* Circle
* Polyline
* Polygon
* Ink
* Highlight
* Underline
* Squiggly
* StrikeOut

Line, Polyline and Polygon have no support of end-line (EL in PDF). This support require a lot of code and I don't want to add it in this PR.

Many PDFs is not perfectly rendered because the border width of annotation (line, square, circle, polyline, polygone, ink) are set to 0 : https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L353

Here, the same PDF, with and without stream appearance : 
* https://www.leblanc-simon.fr/stream.pdf
* https://www.leblanc-simon.fr/no-stream.pdf

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -165,6 +165,41 @@ function getTransformMatrix(rect, bbox, matrix) {
   ];
 }
 
+/**
+ * Convert a PDF array color to a RGB array color.
+ * The default value is black, in RGB color space.
+ *
+ * @param {Array} color - The color array containing either 0
+ *                        (transparent), 1 (grayscale), 3 (RGB) or
+ *                        4 (CMYK) elements
+ */
+function createRgbColor(color) {
+  let rgbColor = new Uint8ClampedArray(3);
+  if (!Array.isArray(color)) {
+    return rgbColor;
+  }
+
+  switch (color.length) {
+    case 0: // Transparent, which we indicate with a null value
+      rgbColor = null;
+      break;
+
+    case 1: // Convert grayscale to RGB
+      ColorSpace.singletons.gray.getRgbItem(color, 0, rgbColor, 0);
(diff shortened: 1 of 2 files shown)


bidwej:
Hi @leblanc-simon and @timvandermeij, 

I am trying to use PDFJS to [show differences between documents](https://winmerge.org/). 

What would be the best way for me to programmatically create rectangle annotations? 

The [pdf-annotate](https://www.npmjs.com/package/@submitty/pdf-annotate.js) project appears to be outdated. I could probably add them using PDFLib as "[annotation_layer_builder](https://github.com/mozilla/pdf.js/blob/50bc4a18e8c564753365d927d5ec6a6d2cce3072/web/annotation_layer_builder.js)" reads annotations from the actual PDF file, but I'd like to keep dependencies to a minimum.

Do you think it would make sense for me to create an "annotation_controller" similar to "[find_controller](https://github.com/mozilla/pdf.js/blob/50bc4a18e8c564753365d927d5ec6a6d2cce3072/web/pdf_find_controller.js)"?

Thank you

timvandermeij:
@bidwej PDF.js is only for rendering PDF data, so programmatically creating annotations is not supported and will have to be built for a custom deployment. Moreover, this isn't the place to discuss such functionality because it does not have anything to do with what this PR is about, namely supporting annotations without appearance streams in the core.

bidwej:
Ok. Thank you

timvandermeij:
Closing since we're using the approach of generating appearance streams now and some of these annotation types are already done.

PR #10142, opened 2018-10-09
Added masks & gradients support, and group hierarchy management in SVG

fr33z00:
Hi all.
In order to have a better PDF to SVG conversion for one of our projects, I forked the repo and improved the SVG graphic rendering on some points (gradients support, path filling, masking, and groups hierarchy).
The tracemonkey file renders correctly (or at least better on the p.11 graph), as far as I checked.

The modifications do not respect the initial authors way of making groups, so a merge may not be acceptable, however, it could help or inspire people working on it.

I noticed that shadingFill ( = gradients) do not render exactly as on canvas (or desktop apps like evince) and couldn't figure out how to improve this, but the result looks pretty similar to what inkscape produces. See the example page below:

source:
[page1.pdf](https://github.com/mozilla/pdf.js/files/2460333/page1.pdf)

canvas rendering:
pdf2canvas

svg rendering:
pdf2svg

inkscape conversion  (do not pay attention to fonts):
pdf2svg_inkscape

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,13 @@ PDF.js is community-driven and supported by Mozilla Labs. Our goal is to
 create a general-purpose, web standards-based platform for parsing and
 rendering PDFs.
 
+## Difference with mozilla repo
+
+This fork improves SVG rendering of PDF docs, and more specifically:
++ adds masks support
++ adds radial/axial gradients (shadingFill) support
++ manages groups hierarchy
+
 ## Contributing
 
 PDF.js is an open source project and always looking for more contributors. To
--- a/README.md
+++ b/README.md
@@ -9,8 +9,7 @@ rendering PDFs.
 ## Difference with mozilla repo
 
 This fork improves SVG rendering of PDF docs, and more specifically:
-+ adds masks support
-+ adds radial/axial gradients (shadingFill) support
++ adds masks, pattern, blend mode, gradients (shading fill) support
 + manages groups hierarchy
 
 ## Contributing
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -830,7 +830,7 @@ class LineAnnotationElement extends AnnotationElement {
 
     // PDF coordinates are calculated from a bottom left origin, so transform
     // the line coordinates to a top left origin for the SVG element.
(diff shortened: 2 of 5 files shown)


timvandermeij:
Nice work; thank you for upstreaming your changes! I'm positive we can merge most parts of this if we split it up into multiple smaller pull requests, one per feature, and make it adhere to the current code structure.

fr33z00:
Tim,
The way the official code structure creates group doesn't stick to the PDF ops tree which explicitly requests groups creation before feeling them. I had no background on PDF documents internal stuff, no background on the pdf.js structure, and, of course, not much time to spend on this. I just couldn't, and still can't, figure out how to combine the operations needed to create a mask or a gradient with the postponed group creation and transform matrix application.
That's why I proposed my code as an inspiration.

fr33z00:
Tim (or anyone else), 
I improved my proposal on SVG, adding patterns & blend mode support, and fixing different things like clipping & transform matrix.
This leads to a perfect rendering of the tracemonkey file, as well as of my test file (see the screenshots below), and many other files I tried.
I now believe the way I used to build groups hierarchy is the only possible one. Thus, this pull request may now be considered as such.

SVG rendering (now comparable to my first post canvas rendering thanks to blend mode support)
screenshot_2018-11-07_pdf-example

tracemonkey page 11 (gradients fill)
screenshot_2018-11-07 tracemonkey

tracemonkey page 13 figure (patterns)
screenshot_2018-11-07 tracemonkey_patterns

timvandermeij:
Is there any way that this can be split into multiple smaller pull requests? It looks like this patch contains quite a bit of different functionality, which makes it hard to review. However, we are interested in merging this, so if you can make a pull request per functionality then this is much more likely to be merged.

fr33z00:
Ok, I'll try to split it in a series. However, the first one will be somehow "massive", because, how I already said, the change from post created groups to pre created cannot be splitted. And much of the work relies on that.
I then close this request and will start the series.

timvandermeij:
Sounds good! Even though the patch may be a bit larger, that's not a problem if it's one self-contained feature, so that should be fine.

PR #10149, opened 2018-10-10
add support to storage read process to pouchdb for sync between different devices

sherylynn:
start a pouchDB server 
add url to gulpfile.js
then u can enjoy syncing read process between your different devices

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,14 +56,22 @@ If everything worked out, install all dependencies for PDF.js:
 
     $ npm install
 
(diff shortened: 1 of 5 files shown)


timvandermeij:
Even though this is nice work, I don't think it should be in the library since it introduces a dependency that most people will not need/use and the functionality is also only interesting for a limited number of use cases. However, if you want to host a modified version of PDF.js with these changes, we'd be happy to mention it on https://github.com/mozilla/pdf.js/wiki/Gallery-of-user-projects-and-modifications. Thanks.

Issue #10151, opened 2018-10-11
SVG renderer misses footnotes


doublex:
Steps to reproduce the problem: 
Example: [r2l.pdf](https://github.com/mozilla/pdf.js/files/2470373/r2l.pdf)
````
npm install
gulp dist-install
wget https://github.com/mozilla/pdf.js/files/2470373/r2l.pdf
node examples/node/pdf2svg.js /tmp/r2l.pdf
````

What went wrong?
Footnotes are missing, from page 2 to page 16, e.g.:
![page-2](https://user-images.githubusercontent.com/274624/46828574-ce537100-cd9b-11e8-8643-fac8a2937f69.png)

timvandermeij:
Fixed by #10162. Open issues already exist for gradient support.

PR #10162, opened 2018-10-16
Normalize BBox of form XObjects in SVG back-end

janpe2:
In issue #10151, footnotes are missing because they are form XObjects that have unnormalized BBox entries, for example `/BBox [0.0 765.354 595.276 0.0]`. This makes the height of SVG `<rect>` element negative, so the clipping path fails and the form XObject becomes invisible.

This PR makes the footnote texts visible but some graphics are still missing because shadings haven't been implemented in the SVG back-end.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -299,6 +299,11 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       var dict = xobj.dict;
       var matrix = dict.getArray('Matrix');
       var bbox = dict.getArray('BBox');
+      if (Array.isArray(bbox) && bbox.length === 4) {
+        bbox = Util.normalizeRect(bbox);
+      } else {
+        bbox = null;
+      }
       var group = dict.get('Group');
       if (group) {
         var groupOptions = {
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
Nice find!
It might not work, but I wanted to ask nonetheless: Provided that it doesn't break the (default) `canvas` back-end, would it perhaps make sense to simply do the normalization [on the `/core` side instead](https://github.com/mozilla/pdf.js/blob/ecbdc508f7d872d2a734932ce12cb96153b004aa/src/core/evaluator.js#L301)?

janpe2:
I think that would be unnecessary because the canvas back-end has no problems when BBox is unnormalized.

Snuffleupagus:
> I think that would be unnecessary because the canvas back-end has no problems when BBox is unnormalized.

Perhaps, but there's some value in having the API return data in a format that works directly in *both* back-ends without having to remember to transform it one of them, hence my question.

timvandermeij:
I don't think there is a use case for needing an unnormalized bounding box, so I agree that it is probably better to handle this in core so any consumers don't need to worry about normalization anymore.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #10166, opened 2018-10-19
Bump version numbers for stable and beta releases

### LINKED ISSUE #10098: Bump versions in `pdfjs.config`
It looks like they did not get bumped after the last releases, so https://github.com/mozilla/pdf.js/blob/master/pdfjs.config is out-of-sync with https://github.com/mozilla/pdf.js/releases and https://mozilla.github.io/pdf.js/getting_started/#download never got updated too.

@brendandahl @yurydelendik Is it just a matter of bumping these versions or do we also need to change the `baseVersion` (not sure what that is)?

timvandermeij:
Fixes #10098.

### DIFF
--- a/pdfjs.config
+++ b/pdfjs.config
@@ -1,6 +1,6 @@
 {
-  "betaVersion": "1.10.88",
-  "stableVersion": "1.9.426",
+  "betaVersion": "2.0.550",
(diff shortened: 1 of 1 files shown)


timvandermeij:
Trivial.

PR #10167, opened 2018-10-20
Export `CMapCompressionType` and `PermissionFlag` on the `pdfjsLib` object (issue 10148, PR 10033 follow-up)

### LINKED ISSUE #10148: CMapCompressionType is not exported in the pdf.js distribution
As implemented in https://github.com/mozilla/pdf.js/pull/8064 and demonstrated in https://github.com/mozilla/pdf.js/issues/8881#issuecomment-327822546, a `CMapReaderFactory` must resolve a promise with an object containing a `compressionType` slot, the value of which should be one of the `CMapCompressionType` enumeration.

Unfortunately, `CMapCompressionType` is not exported by pdf.js as it is currently built, so implementing `CMapReaderFactory` requires copy/pasting the enumeration, or maybe hard-coding the value of `compressionType` (e.g. if all of the cmaps being loaded from disk are known to be compressed).

No PR at this time; I can see at least three reasonable places where `CMapCompressionType` could be exported, but I'd rather not guess at the maintainers' desired aesthetic. :smile:

Snuffleupagus:
`CMapCompressionType` makes a lot of sense to export, for anyone attempting to implement a custom `CMapReaderFactory`; fixes #10148.

`PermissionFlag` likewise needs to be exported, since otherwise the result of the `getPermissions` API methods becomes difficult to interpret; follow-up to #10033.

### DIFF
--- a/src/pdf.js
+++ b/src/pdf.js
@@ -92,6 +92,8 @@ exports.InvalidPDFException = pdfjsSharedUtil.InvalidPDFException;
 exports.MissingPDFException = pdfjsSharedUtil.MissingPDFException;
 exports.SVGGraphics = pdfjsDisplaySVG.SVGGraphics;
 exports.NativeImageDecoding = pdfjsSharedUtil.NativeImageDecoding;
+exports.CMapCompressionType = pdfjsSharedUtil.CMapCompressionType;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

PR #10170, opened 2018-10-21

timvandermeij:


timvandermeij:
What? I even removed the commit about updating packages and the preview still fails, but it works locally.

Snuffleupagus:
Most likely the problems are caused by the new Webpack version, since comparing the output of `gulp generic` with/without this patch highlights one important difference:
```diff
--- \master\build\pdf.js
+++ \patch\build\pdf.js
@@ -7794,7 +7794,7 @@
       require.ensure = require('node-ensure');
     }
     useRequireEnsure = true;
-  } else if (typeof require !== 'undefined' && typeof require.ensure === 'function') {
+  } else if (true && typeof require.ensure === 'function') {
     useRequireEnsure = true;
   }
   if (typeof requirejs !== 'undefined' && requirejs.toUrl) {
```
Either there's a regression in the new Webpack version, see the release info at https://github.com/webpack/webpack/releases, and/or there's something wrong with the bundle definitions in our `gulpfile.js`.

wojtekmaj:
How about updating to 4.21.0 first to narrow down which version has caused the issue?

timvandermeij:
In the end (in another pull request) I did that (see https://github.com/mozilla/pdf.js/blob/master/package.json#L40) and 4.21.0 was the last one that didn't have the issue. I created #10177 to follow up on this.

PR #10172, opened 2018-10-21
Improve formatting when copying text from the text layer

PalmerAL:
Currently, the textLayer uses `div` elements for every text item, which results in extra newlines being inserted when the text is copied into another application. For example, copying a few paragraphs from the tracemonkey file results in this:

screen shot 2018-10-21 at 11 46 45 am

This PR modifies the text layer to use `span`'s instead, which removes the unnecessary line breaks. In order to preserve line breaks between paragraphs, it detects when a text item appears to be the start of a new paragraph (based on its position relative to the last line and changes in font size). If the item is the start of a new paragraph, a `br` is inserted at the beginning of the text item. Additionally, spaces will be inserted at the end of each line if there is not already a space there, so that copying multiple lines results in the correct word spacing.

While this doesn't always work on every PDF, it seems to work pretty well on most common cases, and I haven't found any files where the new behavior is substantially worse than the old formatting. Some examples:

| File | Old | New |
|----- | ----- | ---- |
|compressed.tracemonkey-pldi-09.pdf | screen shot 2018-10-21 at 11 46 45 am | screen shot 2018-10-21 at 11 55 06 am |
| http://www.pdf995.com/samples/pdf.pdf | screen shot 2018-10-21 at 11 57 23 am | screen shot 2018-10-21 at 11 58 37 am |
| https://github.com/mozilla/pdf.js/blob/master/test/pdfs/TAMReview.pdf | screen shot 2018-10-21 at 12 02 01 pm | screen shot 2018-10-21 at 12 02 25 pm |
| https://eloquentjavascript.net/Eloquent_JavaScript.pdf | screen shot 2018-10-21 at 12 04 10 pm | screen shot 2018-10-21 at 12 04 53 pm |
| http://www.music.mcgill.ca/~cmckay/papers/musictech/ISMIR_2006_Codaich.pdf | screen shot 2018-10-21 at 12 06 09 pm | screen shot 2018-10-21 at 12 06 57 pm |

This only works with horizontal text; angled or vertical text will always use `span`'s, and will not have any line breaks. I think this is OK, because it's rare to have large paragraphs of text at an angle, but I can try to add support for that if you think it's important.

I would expect this to be slower than the previous implementation because it does more work to detect paragraphs and because it inserts additional DOM elements, although based on the measurements I've taken, memory and CPU use seem to be about the same as they were before. The Firefox profiler isn't recording any information for `appendText` for some reason, but since it doesn't have a large performance impact in Chrome I would expect it to be OK in Firefox as well.

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -105,41 +105,18 @@ var renderTextLayer = (function renderTextLayerClosure() {
 
     // line adjustments aren't supported for angled text
     if (angle === 0) {
-      /*
-      if this item:
-      * has approximately the same y-position as the last item
-      * is adjacent to the last item
-      it is likely on the same line of text, so it shouldn't be modified
-        */
       if (
         task._lastLineStart &&
-        Math.abs(task._lastLineBottom - (top + fontHeight)) <
-          (fontHeight * 0.4) &&
-        Math.abs(left - task._lastLineEnd) < 15
-        ) {
-
-        /* If the item:
-           * has approximately the same x-position as the start of the previous
-             line
-           * is immediately below the previous line
-           * is approximately the same font size
-          It is likely part of the same text block, which is being wrapped onto
-             a new line.
-          In this case, the new line should still be treated as an inline
-            element, but it might be necessary to insert a space between the
-            last word of the last line and the first word of this line.
-             */
-      } else if (
-        task._lastLineStart &&
-         Math.abs(task._lastLineBottom - top) < (fontHeight * 0.55) &&
-          /* outdent on 2nd line is OK (because the first line of a paragraph
-            is often indented), but indent is not (because it is almost always
-              an intentional line break)
-          */
-         ((task._lastLineStart - left) < (fontHeight * 2) &&
-          (task._lastLineStart - left) > (fontHeight * -0.5)) &&
+        top > task._lastLineBottom &&
+        Math.abs(task._lastLineBottom - top) < (fontHeight * 0.55) &&
+        /* outdent on 2nd line is OK (because the first line of a paragraph
+           is often indented), but indent is not (because it is almost always
+           an intentional line break)
+        */
+        ((task._lastLineStart - left) < (fontHeight * 2) &&
+        (task._lastLineStart - left) > (fontHeight * -0.5)) &&
         (fontHeight / task._lastFontSize > 0.9 &&
-          fontHeight / task._lastFontSize < 1.1)
+        fontHeight / task._lastFontSize < 1.1)
       ) {
         if (!isAllWhitespace(task._lastText[task._lastText.length - 1]) &&
             !isAllWhitespace(geom.str[0])) {
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -100,11 +100,8 @@ var renderTextLayer = (function renderTextLayerClosure() {
       top = tx[5] - (fontAscent * Math.cos(angle));
     }
 
-    /* determine where to insert line breaks in order to make copying text work
+    /* determine where to insert spaces in order to make copying text work
      as expected */
-
-    var topAdjustment = 0;
-    var hasPrecedingLineBreak = false;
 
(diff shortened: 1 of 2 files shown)


timvandermeij:
Could you test what the difference would be if we just switched `div` to `span` elements? The text layer logic is not easy because of the way the PDF format handles text, so less is more in terms of logic. If we can avoid a lot of line breaks with just that simple change, that would already be a big improvement and we can worry about the breaks between paragraphs later. This is all to make it easier to get the patch merged and to avoid regressions.

PalmerAL:
Sure. With just normal `span` elements, the lines between paragraphs are missing, and the last word of each line is combined with the first word of the next line:

screen shot 2018-10-22 at 5 12 49 pm

With the 2nd part of this patch (identifying lines that wrap around, and inserting spaces where necessary), the word spacing is correct (although there are some issues with hyphenated words, I'm not sure whether we should try to fix those or not):

screen shot 2018-10-22 at 5 15 00 pm

With the third part (insertion of `br` elements), the line breaks between paragraphs work correctly.

With just part 1, the results aren't really any better than the current behavior, so I'm not sure it makes sense to do that without also doing part 2, although I'm happy to take that out if you want me to (and part 3 can probably be removed without any side effects).

PalmerAL:
@timvandermeij I've removed the line break insertion, would you like me to remove the other part as well?

timvandermeij:
I'm thinking we should do this in steps because the text layer logic can be difficult to get right. The first step would be only converting `div` to `span`. Then we can merge that, evaluate if it doesn't break anything and then follow-up with the logic to improve the copy-paste behavior. I think there is already an advantage in using `span` over `div` if I look at the results from your screenshots before and after.

I'm also curious what others think of this. Any opinions about this @Snuffleupagus @brendandahl @yurydelendik?

Snuffleupagus:
> I'm also curious what others think of this. Any opinions about this

Having attempted to briefly look through the code in the PR, in its current form it unfortunately seems to be quite far from workable/acceptable.
Ignoring the code formatting problem, the logic is *really really* difficult to even attempt to follow and the approx. 10 different magical constants is definitely not helping matters either.[1]
Keep in mind that there's also a lot of different cases to consider when dealing with the `textLayer`, there's e.g. three different writing modes (left-to-right, right-to-left, top-to-bottom), and probably a lot more subtle edge-cases to consider which is why clear/readable code is important.

Quite frankly, in its current form this code would be extremely difficult (or even impossible?) for anyone to maintain effectively. With larger changes to something like the `textLayer`, it's almost guaranteed that there will be regressions, and attempting to fix those by randomly tweaking one or more of ten magical constants won't work :-(

---
[1] Even replacing those numbers with suitably named constants wouldn't really help here, since they still seem quite "randomly" chosen.

PalmerAL:
Thanks for the feedback! @timvandermeij I've opened another PR that changes the text layer items to `<span>`'s without any other changes here: #10197.

I'll see what I can do to try to simplify the code. In particular, with the `<br>` insertion removed, I think that most of the first if statement can be removed or merged into the second one, which should reduce the overall size of the code significantly and make it easier to understand.

Most of the constants relate to the amount of space that can occur between text items (in various dimensions) before they are considered as being part of a different text block, rather than a continuation of the same block. I'm not sure if there's any better non-random way to determine those, because text spacing is different in every PDF, and there's no standard amount of space between paragraphs - if you have any ideas for how to calculate this better, I'm happy to change it. The one constant that should be pretty easy to get rid of is the space width - it looks like that's already being calculated in evaluator.js, so it should just be a matter of sending that to appendText.

Regarding the edge cases you mentioned, my plan is to only enable space insertion for horizontal LTR text, and ignore everything else. For angled or non-LTR text, the text layer would just use `<span>`'s without any spaces inserted, which would result in something like the 1st screenshot of my comment above. The formatting won't be quite as good as it would be with space insertion enabled, but it should still be at least as good as the current behavior. I'm sure there are probably additional edge cases beyond that, but that should hopefully cut down on the number of things to consider and make this easier to test.

PR #10179, opened 2018-10-24
Change splice to pop in annotation tests

ThomasdenH:
This line in the annotation tests subtracts an array from a number. This is because `splice(-1, 1)` returns a one-element array, while `pop()` returns only the element itself.


timvandermeij (approved):

timvandermeij:
Nice find; thank you!

PR #10182, opened 2018-10-25
Small clean-up of the search related methods in `TextLayerBuilder`

Snuffleupagus:
This patch does four things:
 - Change the search related methods in `TextLayerBuilder` to be "private", since there're only called from within the class itself now.
 - Use `const` for local variables not intended to change in the search related methods in `TextLayerBuilder`.
 - Finally, removes most `this.findController` checks since they are redundant. Note how both `this._convertMatches` and `this._renderMatches` are *only* ever called, from `this._updateMatches`, when `this.findController` is actually defined. Hence there's really no need to repeat those checks all over the place, especially with all the relevant methods now being marked as "private".
 - Always initialize the `this._pageMatchesLength` property with an empty array, to simplify the code in `TextLayerBuilder`.

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -141,7 +141,7 @@ class PDFFindController {
     this._highlightMatches = false;
     this._pdfDocument = null;
     this._pageMatches = [];
-    this._pageMatchesLength = null;
+    this._pageMatchesLength = [];
     this._state = null;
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #10183, opened 2018-10-25
Attempt to improve the UX when searching, by changing the behaviour in a couple of edge-cases (bug 1237076, issue 4141 issue 5561, issue 6746)

### LINKED ISSUE #4141: Repositioning to active page for find works only if the match string changes.
Repositioning to active page for find works only if the match string changes.

Issue 3935 was closed in response to PR 3941. However, the find functionality now only starts at the current active page if the find string is altered. If a user navigates to another page, and then hits the "next" button (without altering the search string), then his search continues from wherever the last search began. This is confusing--apart from the fact that it differs from the behavior of Adobe acrobat. Adobe acrobat starts from the new active page when the next button is pressed—regardless of whether the search string is altered. 

Options for a fix:
1. Do what adobe does, i.e. start searching from the active page when it has been changed by the user. 
1. Since pdfjs runs in a browser (and is not adobe) we could do this more like FF does for un-editable HTML , which is:
   All find functionality starts from the current position (or last position) of the edit cursor. Even though it is not visible in FF in un-editable text (unless a selection has been made), the cursor position is nevertheless known and simply clicking on the page moves it.  This position is also moved by the find functionality. Also, after finding a match, the match is selected and can be copied directly from the page. 
2. Do both and give the integrator of pdfjs an option, perhaps including the runtime option of giving the user a look and feel choice.
3. Fix this as a bug in PR 3941 (i.e. temporarily treat as option 1), and then decide later what to do.

Comparison:
Option 1 is easier to implement as a simple bug fix, and it provides a similar interface to that of adobe.

Option 2 gives the user more fine-grained control (character instead of page level) over the start position of a search and gives users a more uniform presentation of different document types in applications such as FF. In addition, it perhaps fits better with the high weight given to stylistic simplicity and uniformity that has become increasingly popular to users of mobile platforms.

Option 3 is the hardest to implement, but it would allow different behaviors on different platforms. It has the disadvantage that it could be confusing to users when they see both behaviors in different places.

Option 4 gets a fix in right away and puts off a decision or waiting for the final behavior decision.

There are other options, such as some combination of adobe and other common behaviors. This issue could be a feature request or just a bug in what was done in PR 3941, depending on what is deemed best.

### LINKED ISSUE #5561: Toggling the "Highlight all" search feature should not change the view position
When the _Highlight all_ search feature (checkbox in the [demo](http://mozilla.github.io/pdf.js/web/viewer.html), button in the extension) is toggled, then the view position seems to focus the first hit on the current page.
In my opinion, the toggle shouldn't move the view at all, especially if the user turns _Highlight all_ off. Changing the focus is also contrary to Firefox's built-in highlight behavior (I haven't tested other browsers).

(FF34, Windows 7, current Demo version: 1.0.1011)

### LINKED ISSUE #6746: Highlighting all matches causes the page to jump while scrolling
STR:
1 - open http://arxiv.org/pdf/1509.03893v1.pdf (tested only in Nigthly with e10s enabled, but I've had users report this to me in other setups);
2 - make sure "Page Fit" is selected as view, it's easier to reproduce like this but I've managed to reproduce it with other zoom/view factors applied;
3 - search (Ctrl+F) for "alice" and click the "Highlight all" button;
4 - press F3 three times so that again the first of three instances of "alice" is active (it does not occur when the first instance of "alice" is active for the first time!);
5 - click in the document to focus it and press "home" key to jump back to the beginning;
6 - scroll down with "page down" button until you are at around page 7 or 8 and it should jump ahead to page 10 where the first instance of "alice" appears.

I've found this is because [TextLayerBuilder.renderMatches scrolls the match into view indiscriminately](https://github.com/mozilla/pdf.js/blob/master/web/text_layer_builder.js#L213-L216) every time a page is (re)rendered, even while just scrolling (close) to it.

Here's how I've fixed it in my add-on FindBar Tweak: https://github.com/Quicksaver/FindBar-Tweak/commit/c48dc2d5efa480e163064d15e5ca3e3e2bd08c82
- PDFFindController.nextMatch sets a flag to show the current match;
- PDFFindController.updateMatchPosition only actually scrolls to the match when this flag is set, and resets it immediately after scrolling.

Snuffleupagus:

This will obviously require careful manual testing, such that we don't accidentally make some aspect(s) of searching even worse, but it seems to work reasonably well in my initial tests.

*Edit:*
Fixes [bug 1237076](https://bugzilla.mozilla.org/show_bug.cgi?id=1237076)
Fixes #4141
Fixes #5561
Fixes #6746

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -13,9 +13,9 @@
  * limitations under the License.
  */
 
+import { getGlobalEventBus, scrollIntoView } from './ui_utils';
 import { createPromiseCapability } from 'pdfjs-lib';
 import { getCharacterType } from './pdf_find_utils';
-import { getGlobalEventBus } from './ui_utils';
 
 const FindState = {
   FOUND: 0,
@@ -25,6 +25,8 @@ const FindState = {
 };
 
 const FIND_TIMEOUT = 250; // ms
+const MATCH_SCROLL_OFFSET_TOP = -50; // px
+const MATCH_SCROLL_OFFSET_LEFT = -400; // px
 
 const CHARACTERS_TO_NORMALIZE = {
   '\u2018': '\'', // Left single quotation mark
@@ -159,8 +161,26 @@ class PDFFindController {
     });
   }
 
+  scrollMatchIntoView({ element = null, pageIndex = -1, matchIndex = -1, }) {
+    if (!this._scrollMatches || !element) {
+      return;
+    } else if (matchIndex === -1 || matchIndex !== this._selected.matchIdx) {
+      return;
+    } else if (pageIndex === -1 || pageIndex !== this._selected.pageIdx) {
+      return;
+    }
+    this._scrollMatches = false; // Ensure that scrolling only happens once.
+
+    const spot = {
+      top: MATCH_SCROLL_OFFSET_TOP,
+      left: MATCH_SCROLL_OFFSET_LEFT,
+    };
+    scrollIntoView(element, spot, /* skipOverflowHiddenElements = */ true);
+  }
+
   _reset() {
     this._highlightMatches = false;
+    this._scrollMatches = false;
     this._pdfDocument = null;
     this._pageMatches = [];
     this._pageMatchesLength = [];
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
> I think a PR with the first and third commit above would be good since they seem to go well together,

Unfortunately that won't work, since the third commit depends *directly* of the second one. Basically, unless there's a general way of preventing unnecessary scrolling, the third commit is pretty useless.

What I suppose could work, is to submit commits one and four together first, since those should pretty safe. Then each of commit two respectively three in separate PRs after that. How would you feel about that?

timvandermeij:
Yes, that sounds like a good plan to me.

PR #10184, opened 2018-10-26
Ensure that matches are not scrolled into after the findbar has been closed (PR 10100 follow-up)

Snuffleupagus:
This is something that I noticed while testing PR #10183, and it probably deserves a separate PR rather than being lumped in with those changes since it's a direct follow-up to PR #10100.

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -389,6 +389,13 @@ class PDFFindController {
     });
   }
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thank you! It's also good to see that the find code is becoming much more readable lately.

PR #10190, opened 2018-10-28
Add a helper method for `_getVisiblePages`, in `BaseViewer`, for the case where only a single page is displayed in the viewer

Snuffleupagus:
*This is another small piece of clean-up, split off from PR #10183.*

This is relevant for e.g. `PDFSinglePageViewer`, and `PDFViewer` with Presentation Mode active.
By moving this code to a helper method in `BaseViewer`, it's thus possible to reduce the amount of duplicate code that currently needed in `PDFViewer` and `PDFSinglePageViewer`.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -855,6 +855,30 @@ class BaseViewer {
       false : (this.container.scrollHeight > this.container.clientHeight));
   }
 
+  /**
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Good find!

PR #10197, opened 2018-10-30
Use `span` instead of `div` in the text layer

PalmerAL:
This improves copy/pasting text content since it reduces the amount of unnecessary newlines.

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -55,7 +55,7 @@ var renderTextLayer = (function renderTextLayerClosure() {
 
   function appendText(task, geom, styles) {
     // Initialize all used properties to keep the caches monomorphic.
-    var textDiv = document.createElement('div');
+    var textDiv = document.createElement('span');
     var textDivProperties = {
       style: null,
       angle: 0,
--- a/test/text_layer_test.css
+++ b/test/text_layer_test.css
@@ -22,7 +22,7 @@
   right: 0;
   bottom: 0;
 }
-.textLayer > div {
+.textLayer > span {
   position: absolute;
   white-space: pre;
   -webkit-transform-origin: 0% 0%;
--- a/web/text_layer_builder.css
+++ b/web/text_layer_builder.css
@@ -24,7 +24,7 @@
   line-height: 1.0;
 }
 
-.textLayer > div {
+.textLayer > span {
   color: transparent;
   position: absolute;
   white-space: pre;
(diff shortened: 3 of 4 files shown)


PalmerAL:
It looks like the CSS for the `<span>` elements isn't being applied for some reason, so they aren't positioned correctly. I tested a couple of the PDF's that are failing, and they all seem to work fine locally, so I can't think of any reason why they would be failing in the tests. Is there something I'm missing?

PalmerAL:
Thanks for pointing that out! It looks like there's some other CSS rules that depend on the items being divs as well, I think I've fixed them all in 45279d26137de1e4fbe30e79c979f7f9348161f7.

timvandermeij (approved):

timvandermeij:
Thank you for your contribution; I think this is  a good improvement! I copy/pasted the following paragraph of the Tracemonkey paper with and without your patch applied. Before applying your patch the result was:

```
A
typed trace
is a trace annotated with a type for every variable
(including temporaries) on the trace. A typed trace also has an entry
type map
giving the required types for variables used on the trace
before they are defined. For example, a trace could have a type map
(x: int, b: boolean)
, meaning that the trace may be entered
only if the value of the variable
x
is of type
int
and the value of
b
is of type
boolean
. The entry type map is much like the signature
of a function.
```

In contrast, after applying your patch the result was:

```
Atyped traceis a trace annotated with a type for every variable(including temporaries) on the trace. A typed trace also has an entrytype mapgiving the required types for variables used on the tracebefore they are defined. For example, a trace could have a type map(x: int, b: boolean), meaning that the trace may be enteredonly if the value of the variablexis of typeintand the value ofbis of typeboolean. The entry type map is much like the signatureof a function.
```

PalmerAL:
@timvandermeij Thanks for merging this! Any thoughs on what to do with #10172? If you think the complexity of that isn't worth it (which I completely understand), one alternative I've been thinking about is whether it would be possible to unconditionally insert a space before every text layer item, which would fix things like your example above; if that sounds OK to you, I can spend some time investigating that further.

timvandermeij:
@PalmerAL Yes, I think that one is still a bit too complex, at least in its current form. Perhaps we can do it in smaller steps, i.e., first try to get the spaces right and after that perhaps the newlines. If you can find a solution where the spaces are correct and no double ones are inserted, feel free to create a PR for that and we'll review. I'll close the current PR for you. Thank you for looking into this; if we take small steps here I'm sure we can make more improvements.

PR #10202, opened 2018-10-31
Attempt to clean-up/restore pending rendering operations on `RenderTask.cancel` (issue 10200)

### LINKED ISSUE #10200: Transformation matrix doesn't appear to be restored when renderTask.cancel() is called past a certain point
Configuration:
- Web browser and its version: Chrome 70.0.3538.77 Safari 12.0.1
- Operating system and its version: OSX 10.12.6
- PDF.js version: 2.1.19
- Is a browser extension: No

1. Call renderTask.cancel() after a certain period of time (I don't know pdf.js well enough to determine which step in the render function is causing the issue, but if the timeout below is set to 50ms the issue does not occur)

What is the expected behavior? Transformation matrix state should be identical after renderTask.cancel() has completed

What went wrong? The matrix is in a different state and has to be reset using setTransform.

The code I'm using...

pdfjsLib.GlobalWorkerOptions.workerSrc =
    '../build/pdf.worker.js';

console.log('Version:',pdfjsLib.version);

const canvas = document.getElementById('the-canvas');
const ctx = canvas.getContext('2d');
canvas.width = window.innerWidth * 2;
canvas.height = window.window.innerHeight * 2;

pdfjsLib.getDocument('./large_file.pdf').then(function getPdfHelloWorld(pdf) {

    pdf.getPage(1).then(function getPageHelloWorld(page) {

        let renderTask;

        renderTask = page.render({
            canvasContext: ctx,
            viewport: page.getViewport(1.5)
        });

        renderTask.promise.then(() => {

        }).catch(() => {

            //Have to reset the transform or fillRect produces a shape inconsistent with initial transform state
            ctx.setTransform(1, 0, 0, 1, 0, 0);

            ctx.fillStyle = 'red';
            ctx.fillRect(500,500,500,500);

        });

        setTimeout(() => {

            renderTask.cancel();

        }, 500);

    });

});

Snuffleupagus:
Please note that, given the lack of a runnable example, I'm not totally sure if this first of all is enough to *completely* address the issue as filed and second of all if we actually want this new behaviour.

Probably fixes #10200.

*Edit:* Given that every call to `PDFPageProxy.render` will create a new `InternalRenderTask`, closing the pending rendering operations on `cancel` in this way shouldn't be a problem.
Furthermore note that, on `cancel`, rendering isn't immediately aborted anyway, but will continue until [`CanvasGraphics.executeOperatorList`](https://github.com/mozilla/pdf.js/blob/f6bc93402ef35675ce61a278e3abe6b93d7fd393/src/display/canvas.js#L771) has finished processing the *current* chunk of the `operatorList`. Hence attempting to close any pending operations on `cancel` shouldn't be a problem, and would even be a good thing since it will ensure that any temporary canvases are [actually being removed](https://github.com/mozilla/pdf.js/blob/f6bc93402ef35675ce61a278e3abe6b93d7fd393/src/display/canvas.js#L859-L860) on `cancel`.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2439,6 +2439,9 @@ var InternalRenderTask = (function InternalRenderTaskClosure() {
     cancel: function InternalRenderTask_cancel() {
       this.running = false;
       this.cancelled = true;
+      if (this.gfx) {
+        this.gfx.endDrawing();
+      }
       if (this._canvas) {
         canvasInRendering.delete(this._canvas);
       }


timvandermeij (approved):

timvandermeij:
Looks good to me. Thanks!

PR #10203, opened 2018-10-31
Change `canvasInRendering` to a `WeakSet` instead of a `WeakMap`

Snuffleupagus:
Note how nowhere in the code `canvasInRendering.get()` is ever called, and that this structure is really only used to store references to `<canvas>` DOM elements.
The reason for this being a `WeakMap` is probably because at the time we weren't using `core-js` polyfills yet, and since there already existed a manually implemented `WeakMap` polyfill it was probably simpler to use that.


timvandermeij (approved):

timvandermeij:
Looks like a better choice to me too. Thanks!

PR #10215, opened 2018-11-03
attempts to fix scroll retain

WestonDease:
alters view_history.js

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -314,6 +314,15 @@
         "acorn": "^5.0.3"
       }
     },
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
Was the PR perhaps opened against this repository by mistake?

There's no explanation of what you're trying to do here, why the change should happen in the PDF.js library itself rather than in a custom fork, and finally what little code there is here is definitely not working.

timvandermeij:
Closing because of the reasons above.

PR #10217, opened 2018-11-03
For repeated 'findagain' operations, attempt to reset the search position if the user has e.g. scrolled in the document (issue 4141)

### LINKED ISSUE #4141: Repositioning to active page for find works only if the match string changes.

Snuffleupagus:
Currently we'll only attempt to start from the current page when a new search is done, however for 'findagain' operations we'll always continue from the last match position.
This could easily lead to confusing behaviour if the user has scrolled to a completely different part of the document. In an attempt to improve this somewhat, for repeated 'findagain' operations, we'll instead reset the position to the current page when it's *absolutely* certain that the user has scrolled.

Note that this required adding a new `BaseViewer` method, and exposing that through `PDFLinkService`, in order to check if a given page is visible.
In an attempt to avoid issues, in custom implementations of `PDFFindController`, the code checks for the existence of the `PDFLinkService.isPageVisible` method *before* using it.

Fixes #4141.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -878,6 +878,23 @@ class BaseViewer {
     throw new Error('Not implemented: _getVisiblePages');
   }
 
+  /**
+   * @param {number} pageNumber
+   */
+  isPageVisible(pageNumber) {
+    if (!this.pdfDocument) {
+      return false;
+    }
+    if (this.pageNumber < 1 || pageNumber > this.pagesCount) {
+      console.error(
+        `${this._name}.isPageVisible: "${pageNumber}" is out of bounds.`);
+      return false;
+    }
+    return this._getVisiblePages().views.some(function(view) {
+      return (view.id === pageNumber);
+    });
+  }
+
   cleanup() {
     for (let i = 0, ii = this._pages.length; i < ii; i++) {
       if (this._pages[i] &&
--- a/web/interfaces.js
+++ b/web/interfaces.js
@@ -77,6 +77,11 @@ class IPDFLinkService {
    * @param {Object} pageRef - reference to the page.
    */
   cachePageRef(pageNum, pageRef) {}
+
+  /**
+   * @param {number} pageNumber
+   */
+  isPageVisible(pageNumber) {}
 }
 
 /**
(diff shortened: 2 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Seems to work just fine. Thank you!

PR #10219, opened 2018-11-03
Convert `src/core/ps_parser.js` to use ES6 classes

Snuffleupagus:
Besides being a fairly small and self-contained file, this code also shows a possible way of defining static constants on classes.


timvandermeij (approved):

timvandermeij:
Thank you for converting this file!

PR #10223, opened 2018-11-05
When the search query changes, regardless of the search command, always re-calculate matches (bug 1030622)

Snuffleupagus:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1030622

*Edit:* When following the STR in https://bugzilla.mozilla.org/show_bug.cgi?id=1030622#c4, the problem occurs at **step 6** since the automatic population of the findInput apparently doesn't trigger a find event.
Obviously this could, and maybe even should, be seen as a bug in Firefox itself, however it's easy enough to fix in PDF.js and the patch makes sense in general since `PDFFindController.executeCommand` becomes less sensitive to user error when called (e.g. in a custom viewer implementation).

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -113,7 +113,7 @@ class PDFFindController {
   executeCommand(cmd, state) {
     const pdfDocument = this._pdfDocument;
 
-    if (this._state === null || this._shouldDirtyMatch(cmd)) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #10227, opened 2018-11-06
Update url-lib.js   code optimization

longFeiLi:
It is recommended to use es5
Or before adding get
```javascript
  get toString() {}
```

### DIFF
--- a/external/url/url-lib.js
+++ b/external/url/url-lib.js
@@ -483,7 +483,7 @@
   }
 
   JURL.prototype = {
-    toString() {
+    toString: function() {
       return this.href;
     },
     get href() {


Snuffleupagus:
This is a completely unnecessary change, since that file in bundled into the built `pdf.js`/`pdf.worker.js` files and will thus be automatically translated into ES5 compatible code by Babel; see e.g. the output in the `pdfjs-dist` repo: https://github.com/mozilla/pdfjs-dist/blob/ab06013da9a92dd94ff06a7fadba0345554dec9b/build/pdf.js#L7687-L7689

giridharkannan:
Am using pdfjs-dist in Angular project and what you said is not happening there. If possible, could you change that one like to es5 compactable line?

PR #10231, opened 2018-11-07
Stop scrolling the document when "Highlight All" is toggled in the findbar (issue 5561)

### LINKED ISSUE #5561: Toggling the "Highlight all" search feature should not change the view position

Snuffleupagus:
This is consistent with the general, e.g. HTML, search functionality of the Firefox browser.

Fixes #5561; and depends on PR #10220 landing *first* to suppress all scrolling when "Highlight All" changes.

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -117,7 +117,9 @@ class PDFFindController {
       this._dirtyMatch = true;
     }
     this._state = state;
-    this._updateUIState(FindState.PENDING);
+    if (cmd !== 'findhighlightallchange') {
+      this._updateUIState(FindState.PENDING);
+    }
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Much better; thanks!

PR #10234, opened 2018-11-07
Direct all logging to `console.warn` instead of `console.log`

cemerick:
This avoids polluting `stdout`, important for node.js programs that are intended to produce machine-readable output.

(Ticket/PR suggested in IRC.)

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -296,20 +296,20 @@ function getVerbosityLevel() {
 // end users.
 function info(msg) {
   if (verbosity >= VerbosityLevel.INFOS) {
-    console.log('Info: ' + msg);
+    console.warn('Info: ' + msg);
   }
 }
 
 // Non-fatal warnings.
 function warn(msg) {
   if (verbosity >= VerbosityLevel.WARNINGS) {
-    console.log('Warning: ' + msg);
+    console.warn('Warning: ' + msg);
   }
 }
 
 // Deprecated API function -- display regardless of the `verbosity` setting.
 function deprecated(details) {
-  console.log('Deprecated API usage: ' + details);
+  console.warn('Deprecated API usage: ' + details);
 }
 
 function unreachable(msg) {


src/shared/util.js
```diff
@@ -296,20 +296,20 @@ function getVerbosityLevel() {
 // end users.
 function info(msg) {
   if (verbosity >= VerbosityLevel.INFOS) {
-    console.log('Info: ' + msg);
+    console.warn('Info: ' + msg);
```
timvandermeij: I think this should be `console.info` since information messages are now shown incorrectly as warnings in the console.

cemerick: I actually did that first, too, which is when I learned that, for reasons passing understanding, `console.info` is [just an alias for `console.log`](https://github.com/nodejs/node/blob/af6d26281f734dbbf1bf497103d1cb4b85a52b2e/lib/console.js#L212), and thus _also_ pipes to `stdout`.

I think this means that there is truly no way for e.g. browser graphical consoles to usefully discriminate between simple log messages and warnings/errors, while also adhering to good `stdout`/`stderr` hygiene.

Snuffleupagus: *Please note:* It's easy to suppress all of these `info`/`warn` messages, if you simply include the [`verbosity` parameter](https://github.com/mozilla/pdf.js/blob/2194aef03e155c3ee0f2f1491132b5b4e6dd6cfd/src/display/api.js#L147-L148) when calling `getDocument`. Following e.g. the `pdf2svg` example, you'd simply add `verbosity: pdfjsLib.VerbosityLevel.ERRORS,` to the parameters listed in https://github.com/mozilla/pdf.js/blob/2194aef03e155c3ee0f2f1491132b5b4e6dd6cfd/examples/node/pdf2svg.js#L87-L90 for all of these messages to be completely suppressed.

Would that be sufficient to work around the problem being solved here, since it seems a tiny bit unfortunate to have Node.js implementation details control exactly how the `info`/`warn` helper functions are implemented?

cemerick: OK, I thought we were already setting verbosity (to warnings), but we actually weren't; it was being provided via the (maybe deprecated?) mechanism of setting options on a `PDFJS` global object prior to requiring/loading `pdf.js`. Moving that to the `.getDocument()` call does squelch the messages entirely.

Anyway, unstated in my initial problem description is that we _do_ want to capture logs (ours as well as pdf.js'), in parallel with consuming the machine-readable output of the program; so, just squelching logging isn't great.

FWIW, including this in our program does accomplish our objectives:

```js
console.log = console.info = console.warn;
```

We can simply carry on using that if you don't want to get wagged by the node.js tail here.

timvandermeij:
Closing since I also agree that it's better to leave this as-is as it's confusing to see information messages as warnings in the browser console and unfortunate to adapt the code specifically for a Node.js detail.

PR #10238, opened 2018-11-08
Move the `interface` definitions out of `src/core/worker.js` and into their own file

Snuffleupagus:
These interfaces are already used in different files, in both the `src/core/` and `src/display/` folders, and having them reside in their own file seems a lot clearer and is also similar to the existing viewer interfaces.

As part of moving the `interface` definitions, they're also converted to ES6 classes.


timvandermeij (approved):

timvandermeij:
I agree that it's better this way. Nice find!

PR #10245, opened 2018-11-09
Update `BaseViewer.scrollPageIntoView` to always validate the `pageNumber` parameter

Snuffleupagus:
Note that when e.g. presentation mode is active, we fail[1] to ensure that the `pageNumber` parameter is actually an integer before calling `_setCurrentPageNumber` (that method expects the argument be an integer).
Also changes the method signature, of `scrollPageIntoView`, to use object destructuring instead.

---
[1] Most likely, this is actually *my* oversight :-)

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -660,23 +660,21 @@ class BaseViewer {
    * Scrolls page into view.
    * @param {ScrollPageIntoViewParameters} params
    */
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #10246, opened 2018-11-10
Convert the remaining code in `src/display/api.js` to use ES6 classes

Snuffleupagus:
Apart from the `class` conversion, this also changes all `var` to `let`/`const` and (temporarily) enables the ESLint `no-var` rule in this file (until such a time that it can be enabled globally).

It's obviously possible that the `deprecated` calls aren't actually wanted here, and I won't insist on keeping them if there's any objections :-)
However, I do think that it would be good to at least do a slightly better job of "promoting" `PDFDocumentLoadingTask`/`RenderTask`, since they contain more functionality than just a promise (see e.g. the `destroy` respectively `cancel` methods).

*Please note:* It's highly recommended to look at each commit separately and ignoring whitespace changes, with `?w=1`, when reviewing this. For some reason GitHub completely chokes on the `PDFPageProxy` diff, so for convenience here's the `git diff -w` output: https://gist.github.com/Snuffleupagus/d44fd9aa1f63abe2e377e138de2d6fef

### DIFF
--- a/examples/learning/helloworld.html
+++ b/examples/learning/helloworld.html
@@ -28,11 +28,12 @@
   //
   // Asynchronous download PDF
   //
-  pdfjsLib.getDocument(url).then(function getPdfHelloWorld(pdf) {
+  var loadingTask = pdfjsLib.getDocument(url);
+  loadingTask.promise.then(function(pdf) {
     //
     // Fetch the first page
     //
-    pdf.getPage(1).then(function getPageHelloWorld(page) {
+    pdf.getPage(1).then(function(page) {
       var scale = 1.5;
       var viewport = page.getViewport(scale);
 
--- a/examples/learning/helloworld64.html
+++ b/examples/learning/helloworld64.html
@@ -39,9 +39,10 @@
 
   // Opening PDF by passing its binary data as a string. It is still preferable
   // to use Uint8Array, but string or array-like structure will work too.
-  pdfjsLib.getDocument({data: pdfData}).then(function getPdfHelloWorld(pdf) {
+  var loadingTask = pdfjsLib.getDocument({data: pdfData});
+  loadingTask.promise.then(function(pdf) {
     // Fetch the first page.
-    pdf.getPage(1).then(function getPageHelloWorld(page) {
+    pdf.getPage(1).then(function(page) {
       var scale = 1.5;
       var viewport = page.getViewport(scale);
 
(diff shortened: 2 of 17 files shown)


src/display/api.js
```diff
@@ -850,385 +850,395 @@ class PDFDocumentProxy {
...
-        let textContent = {
-          items: [],
-          styles: Object.create(null),
-        };
+  /**
+   * @param {getTextContentParameters} params - getTextContent parameters.
+   * @return {Promise} That is resolved a {@link TextContent}
+   * object that represent the page text content.
+   */
+  getTextContent(params) {
```
timvandermeij: Shouldn't this be `params = {}` to match the removed `params = params || {};` line?

examples/learning/helloworld.html
```diff
@@ -28,7 +28,8 @@ <h1>'Hello, world!' example</h1>
   //
   // Asynchronous download PDF
   //
-  pdfjsLib.getDocument(url).then(function getPdfHelloWorld(pdf) {
+  var loadingTask = pdfjsLib.getDocument(url);
+  loadingTask.promise.then(function getPdfHelloWorld(pdf) {
```
timvandermeij: Let's use an anonymous function here like in most other examples.

examples/learning/helloworld64.html
```diff
@@ -39,7 +39,8 @@ <h1>'Hello, world!' example</h1>
 
   // Opening PDF by passing its binary data as a string. It is still preferable
   // to use Uint8Array, but string or array-like structure will work too.
-  pdfjsLib.getDocument({data: pdfData}).then(function getPdfHelloWorld(pdf) {
+  var loadingTask = pdfjsLib.getDocument({data: pdfData});
+  loadingTask.promise.then(function getPdfHelloWorld(pdf) {
```
timvandermeij: Let's use an anonymous function here like in most other examples.

timvandermeij (commented):
Looks good; just a few minor comments. I agree with the deprecations.

src/display/api.js
Snuffleupagus: Technically it doesn't really matter, since `getTextContent` is basically just a convenient wrapper around `streamTextContent` now. However, given that the parameters are optional, it cannot hurt to indicate that with `params = {}` though :-)

timvandermeij (approved):

timvandermeij:
Awesome, thank you!

PR #10255, opened 2018-11-14
Avoid unnecessary `PDFFindController.scrollMatchIntoView` calls in `TextLayerBuilder._renderMatches` when `highlightAll` is set (PR 10201 follow-up)

Snuffleupagus:
When `highlightAll` is *not* set, there's only going to be a single match per page and unconditionally calling `PDFFindController.scrollMatchIntoView` doesn't really matter.
However, when `highlightAll` is set the current code may result in a large number of unnecessary `PDFFindController.scrollMatchIntoView` calls. Since `TextLayerBuilder._renderMatches` already checks if a particular match is the selected one, for highlighting purposes, it's simple enough to also skip scrolling completely for non-selected matches.

### DIFF
--- a/web/text_layer_builder.js
+++ b/web/text_layer_builder.js
@@ -238,15 +238,16 @@ class TextLayerBuilder {
       let match = matches[i];
       let begin = match.begin;
       let end = match.end;
-      let isSelected = (isSelectedPage && i === selectedMatchIdx);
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #10264, opened 2018-11-17
Update translations and packages

timvandermeij:
Moreover, resolve deprecation warnings for Jasmine. Jasmine recommends to use the `configure` method on the environment during boot. This commit makes the code correspond to how it's done in Jasmine's default boot file. The options dropdown in the HTML reporter now works again after these changes, because this broke in the upgrade to Jasmine 3, and the unit tests are executed in a random order by default, which is important to make sure the unit tests are self-contained and don't depend on the result of another unit test.


timvandermeij:
Trivial.

Issue #10177, opened 2018-10-23
Investigate upgrading Gulp, Webpack and Node.js dependencies


timvandermeij:
From #10176:

```
Webpack is pinned because versions higher than this cause the viewer not
to work (see https://github.com/mozilla/pdf.js/pull/10170#issuecomment-431697032).

Node.js is pinned for Travis CI because version 11 requires that we
update to Gulp 4.
```

Snuffleupagus:
> Webpack is pinned because versions higher than this cause the viewer not to work (see https://github.com/mozilla/pdf.js/pull/10170#issuecomment-431697032).

The problem is that Webpack, starting from version `4.22.0`, is now *parsing* the entire
```javascript
typeof __non_webpack_require__ !== 'undefined'
```
expression during the build process, causing the above line to be reduced to `true`.
The expected, and previous output, is that `__non_webpack_require__` is replaced with `require` as
```javascript
typeof require !== 'undefined'
```

According to the documentation, this *should* still work; see https://webpack.js.org/api/module-variables/#__non_webpack_require__-webpack-specific-

If someone has time (since a reduced test-case might be needed), this should be reported to https://github.com/webpack/webpack/issues

Snuffleupagus:
Was the Webpack regression described in https://github.com/mozilla/pdf.js/issues/10177#issuecomment-441279491 ever reported upstream, or has it perhaps already been fixed in a later release?

(You could probably hack around this on the PDF.js side, by using e.g. `PDFJSDev.eval` or similar, but that would IMHO be a last resort if the Webpack regression isn't fixed.)

timvandermeij:
I tried the latest release and it isn't fixed yet, nor do I really know how to report this in a way that it's actionable for upstream because I don't know that much about this part of the codebase to reduce it to a minimal example.

timvandermeij:
I have filed an upstream issue for this bug at https://github.com/webpack/webpack/issues/8829. If that is fixed, we can remove the workaround. Tracking this upstream now.

PR #10266, opened 2018-11-18
Upgrade to Gulp 4

timvandermeij:
This required the following changes in the Gulpfile:

- Defining a series of tasks is no longer done with arrays, but with the `gulp.series` function. The `web` target is refactored to use a smaller number of tasks to prevent tasks from running multiple times.
- Getting all tasks must now be done through the task registry.
- Tasks that don't return anything must call `done` upon completion.

Moreover, this upgrade allows us to use the latest Node.js on Travis CI again.

Fixes most parts of #10177.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "10"
+  - node
 cache:
   directories:
     - node_modules
--- a/examples/browserify/gulpfile.js
+++ b/examples/browserify/gulpfile.js
@@ -33,4 +33,4 @@ gulp.task('build-worker', function() {
     .pipe(gulp.dest(OUTPUT_PATH));
 });
 
-gulp.task('build', ['build-bundle', 'build-worker']);
+gulp.task('build', gulp.series('build-bundle', 'build-worker'));
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -439,13 +439,14 @@ function makeRef(done, bot) {
   });
 }
 
-gulp.task('default', function() {
+gulp.task('default', function(done) {
   console.log('Available tasks:');
-  var tasks = Object.keys(gulp.tasks);
+  var tasks = Object.keys(gulp.registry().tasks());
   tasks.sort();
   tasks.forEach(function (taskName) {
     console.log('  ' + taskName);
   });
+  done();
 });
 
 gulp.task('buildnumber', function (done) {
@@ -549,7 +550,7 @@ gulp.task('locale', function () {
   ]);
 });
 
-gulp.task('cmaps', function () {
+gulp.task('cmaps', function (done) {
   var CMAP_INPUT = 'external/cmaps';
   var VIEWER_CMAP_OUTPUT = 'external/bcmaps';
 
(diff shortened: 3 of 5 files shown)


timvandermeij:
It looks like we need to install `gulp-cli` globally on the bots for this to work; refer to https://github.com/gulpjs/gulp-cli/issues/84#issuecomment-280368039. I think `gulp` was already installed globally, but may need to be updated.

brendandahl:
What version of gulp-cli do we need on the bots?
On linux it looks like we have the latest version:
CLI version 2.0.1
Local version 3.9.1

timvandermeij:
Hm, it looks like `gulp-cli`is correct then, but it somehow still uses Gulp 3.9.1 instead of Gulp 4. Looking at https://github.com/pattern-lab/edition-node-gulp/wiki/Updating-to-Gulp-4, perhaps it's a difference between the globally installed Gulp and a locally installed Gulp (CLI)?

timvandermeij:
It looks like Gulp 4 got the `latest` tag just now, so until this is rebased the `master` builds will fail unfortunately.

timvandermeij:
@Snuffleupagus Could you perhaps review this so we can resolve the build failures?

Snuffleupagus:
Looks generally OK, based on a quick look :-)

However, there's two things which seem odd/wrong, but in the interest of un-breaking the builds I suppose that you could fix them in a follow-up PR (assuming it happens relatively soon) rather than here:
 - Looking at the `botio` output, e.g. in http://54.67.70.0:8877/bd8161ddbad6eaf/output.txt (from https://github.com/mozilla/pdf.js/pull/10266#issuecomment-447142005 and also on Travis, there's now a bunch of `Starting '<anonymous>'...` respectively `Finished '<anonymous>'` lines in the logs.
 - More problematic, looking again at http://54.67.70.0:8877/bd8161ddbad6eaf/output.txt, it seems that the various `gulp` tasks are run over-and-over (four times in total) which isn't great.

wojtekmaj:
FYI: This should fix tests on master

timvandermeij:
I'm not entirely sure what happened for the `web` target that it repeats the tasks. I'll have to look into that; most likely the targets just have superfluous dependencies or there are too many tasks for one purpose. The `<anonymous>` lines should be fixable by using named functions instead.

wojtekmaj:
Well, that's because you Gulp 4 don't have dependencies!

Here's more info and how to fix:

https://fettblog.eu/gulp-4-parallel-and-series/ <kbd>Ctrl</kbd><kbd>F</kbd> `Gulp 4 does not have any dependency check`

TL;DR

```js
gulp.task("after-this", afterThis));
gulp.task("do-this-1", ["after-this"], doThis1));
gulp.task("do-this-2", ["after-this"], doThis2));
```

-->

```js
gulp.task("after-this", afterThis));
gulp.task("do-this-1", doThis1));
gulp.task("do-this-2", doThis2));
gulp.series("after-this", gulp.parallel("do-this-1", "do-this-2");
```

timvandermeij:
Thank you for the feedback, @Snuffleupagus and @wojtekmaj! The builds should succeed again now, we got rid of quite a few old dependencies and Node.js doesn't report vulnerabilities anymore. I'll create a follow-up issue for the remaining improvements.

PR #10275, opened 2018-11-20
Fix `NameOrNumberTree.get` to actually perform a binary search to find the requested key

Snuffleupagus:
The intent of the code, based on existing comments, is to perform a binary search. However, because of what appears to be a typo in the code responsible for computing the current search index, this code is always checking *every* entry (albeit only at the "final" node) starting from the last one.


timvandermeij (approved):

timvandermeij:
Nice find!

Issue #10284, opened 2018-11-22
How to set cmap configs?


komiyamast:
Configuration:
- PDF.js version: master branch

Steps to reproduce the problem:
1. I want to set the params 'cmapUrl' and 'cMapPacked'.
2. However, it seems to be impossible to set these params from the outside.

https://github.com/mozilla/pdf.js/blob/7b4f3035d2373ff97c74bcdc8a8e387709649ec6/src/display/api.js#L1657-L1660

What is the expected behavior? (add screenshot)
Is there a way to do that as following code?

```
import pdfjsLib from 'pdfjs-dist/build/pdf'
pdfjsLib.cmapUrl = '...'
pdfjsLib.cMapPacked = true
```

I know this code worked in v1.10, but I cannot find similar code in latest master.

https://github.com/mozilla/pdf.js/blob/ea29ec83e1bb4ea44d71cccc451c6b9374e10fb9/src/display/global.js#L102-L107

komiyamast:
https://github.com/mozilla/pdf.js/pull/10285

PR #10285, opened 2018-11-22
add a agument to set params

komiyamast:
[IMO] Simple way to solve this issue
https://github.com/mozilla/pdf.js/issues/10284

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -221,7 +221,7 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *
  * @return {PDFDocumentLoadingTask}
  */
-function getDocument(src) {
+function getDocument(src, params) {
   const task = new PDFDocumentLoadingTask();
 
   let source;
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Sorry, but this is wrong!
If you actually read the JSDoc comments *just above* the `function getDocument(src) {` line, you'll see this is *already supported* since `getDocument` accepts an object: https://github.com/mozilla/pdf.js/blob/7b4f3035d2373ff97c74bcdc8a8e387709649ec6/src/display/api.js#L212-L224

The valid parameters in that object are specified by `DocumentInitParameters`, where all API options (such as `cMapUrl`/`cMapPacked`) are included; refer to: https://github.com/mozilla/pdf.js/blob/7b4f3035d2373ff97c74bcdc8a8e387709649ec6/src/display/api.js#L123-L202

komiyamast:
@Snuffleupagus

Thanks sir!!!
I'm sorry for forgetting to read the JSDoc.
All my trouble are solved.

`.getDocument({ data: uint8Array, cMapUrl, cMapPacked })`

PR #10291, opened 2018-11-23
Update natives secondary dependency for Node.js 11.0+ compatibility

wojtekmaj:
See: https://github.com/gulpjs/gulp/issues/2246

It looks like Node.js 11.x runs Gulp 3.x just fine if we update this secondary dependency. Issue #10177 may be partially resolved by this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5790,9 +5790,9 @@
       }
     },
     "natives": {
-      "version": "1.1.4",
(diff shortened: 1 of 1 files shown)


timvandermeij:
Closing since this is already covered by upgrading to Gulp 4, for which we have an open pull request too: #10266. The `natives` dependency will be dropped altogether when we merge that and the Travis configuration will also be fixed there. Thanks.

PR #10292, opened 2018-11-23
Replace uglify-es with Terser

wojtekmaj:
See discussion on why it matters:

https://github.com/webpack/webpack/issues/8350

TL;DR uglify-es is an abandoned package with a lof of bugs which are hard to debug, and Terser is a fork of uglify-es.

You may want to revisit #10177 as Webpack 4.26.0 also drops uglify-js in favor of Terser.


package.json
```diff
@@ -32,9 +32,9 @@
     "streamqueue": "^1.1.2",
     "systemjs": "^0.21.5",
     "systemjs-plugin-babel": "^0.0.25",
+    "terser": "^3.10.12",
     "ttest": "^1.1.0",
     "typogr": "^0.6.8",
-    "uglify-es": "^3.3.9",
```
timvandermeij: It looks like it's not removed from the `package.json` properly; refer to https://github.com/mozilla/pdf.js/pull/10292/files#diff-32607347f8126e6534ebc7ebaec4853dR13111.

wojtekmaj: It's not removed because outdated Webpack still uses it. Once you upgrade Webpack to `>=4.26.0`, it will disappear from `package-lock.json`.

timvandermeij (approved):

timvandermeij:
Looks good to me. Thanks!

PR #10293, opened 2018-11-23
Upgrade to Babel 7

wojtekmaj:
Updated using [babel-upgrade](https://github.com/babel/babel-upgrade) with custom fixes on top, mostly for custom Babel configuration.

I don't know how to verify it better, `gulp generic` passes without any error, I ran the viewer and it seems to be fine.
Happy to hear your comments :)


package.json
```diff
@@ -1,21 +1,24 @@
 {
   "name": "pdf.js",
   "version": "2.0.0",
+  "main": "build/lib/pdf.js",
```
Snuffleupagus: This must be reverted, since that's not at all desired/correct.

Please remember that the `lib` folder mostly exists for running the unit-tests on Travis, and the officially supported way of using the library is through the built `pdf.js`/`pdf.worker.js` files.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #10294, opened 2018-11-23
Add ESLint plugin import to be warned against invalid paths

wojtekmaj:
...also fixes some invalid syntax in `.eslintrc` files.


timvandermeij (approved):

timvandermeij:
Thank you for the improvement!

PR #10304, opened 2018-11-26
Add protection against directory traversal attacks

april:
This prevents people from making a request such as:

```bash
$ curl --path-as-is 'http://127.0.0.1:8888/../../../../../../../../../etc/passwd'
```

### DIFF
--- a/test/webserver.js
+++ b/test/webserver.js
@@ -80,7 +80,11 @@ WebServer.prototype = {
   _handler: function (req, res) {
     var url = req.url.replace(/\/\//g, '/');
     var urlParts = /([^?]*)((?:\?(.*))?)/.exec(url);
-    var pathPart = decodeURI(urlParts[1]), queryPart = urlParts[3];
+    // guard against directory traversal attacks,
+    // e.g. /../../../../../../../etc/passwd
+    // which let you make GET requests for files outside of this.root
+    var pathPart = path.normalize(decodeURI(urlParts[1]));
+    var queryPart = urlParts[3];
     var verbose = this.verbose;
 
     var methodHooks = this.hooks[req.method];


timvandermeij:
Looks good. Could you squash the commits into one? Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

april:
Sure, I'll take care of that tomorrow. Alternatively, you can merge and squash when you accept the commit.  :)

april:
Done! I'm so used to using merge+squash on GitHub that I don't usually think of doing that anymore.  :)

test/webserver.js
```diff
@@ -80,7 +80,11 @@ WebServer.prototype = {
   _handler: function (req, res) {
     var url = req.url.replace(/\/\//g, '/');
     var urlParts = /([^?]*)((?:\?(.*))?)/.exec(url);
-    var pathPart = decodeURI(urlParts[1]), queryPart = urlParts[3];
+    // guard against directory traversal attacks,
+    // e.g. /../../../../../../../etc/passwd
```
Snuffleupagus: This example doesn't seem to agree with the regular expression used below; did you intend this to be `../../../../../../../etc/passwd` instead?
Or is this comment actually correct, and the regular expression wrong?

april: `/../../../../../../../etc/passwd` is the path component of the URI. It will always start with `/`.

In theory, `path.normalize()` will turn `/../../../../../../../etc/passwd` into `/etc/passwd` as the path, which should fix pdf.js's current problem (as far as I can tell). The way the code parses URLs using regular expressions instead of using built-in libraries is pretty error-prone, so this was just a failsafe. If you would prefer, I can simply make it `path.normalize(decodeURI(urlParts[1]))`.

Snuffleupagus: If the comment is correct, doesn't that mean that the regular expression is wrong though?
Assuming that you actually want to transform `/../../../../../../../etc/passwd` into `/etc/passwd`, then the regexp needs to read `/^([\/\\]\.\.)+/` to work as expected.

brendandahl: Let's just go with the path.normalize. I lean towards avoiding regex when possible.

brendandahl (changes requested):

april:
Sounds good. I'm at the Mozilla All-Hands this week so I'll try to get this updated, but otherwise I will get it fixed first thing next week.

april:
Alright, the requested changes have been made.  :)

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #10309, opened 2018-11-27
Typos on strings #86 & 225

kant:


Snuffleupagus:
As is mentioned in https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md, the translations are *not* maintained in this repo:

timvandermeij:
Closing since the translations are not maintained here, only imported from upstream. Please report the issue to https://wiki.mozilla.org/L10n:Teams:es-AR instead since they can fix it and we can pull the changes automatically afterwards. Thanks.

PR #10318, opened 2018-11-29
In `GENERIC` builds, dispatch a "webviewerloaded" event (from the `webViewerLoad` function) before initializing the viewer

Snuffleupagus:
- Replace the `OVERRIDES` list in `PDFViewerApplication._readPreferences` with a `disablePreferences`, in `GENERIC` builds, `AppOption instead
   
   Rather than having a (somewhat) randomly choosen list of Preferences which `AppOptions` are allowed to override, it makes much more sense to simply add an AppOption to allow custom implementations to ignore Preferences altogether (it's also inline with the AppOption that allows the `ViewHistory` to be bypassed on load).

 - In `GENERIC` builds, dispatch a "webviewerloaded" event (from the `webViewerLoad` function) before initializing the viewer
   
   With the removal of the global `PDFJS` object, in PDF.js version `2.0`, the viewer options are no longer as easily accessible as they previously were (and issues have been filed about this).
In particular, since the viewer files aren't necessarily loaded *immediately*, this means that `PDFViewerApplication`/`PDFViewerApplicationOptions` aren't necessarily available directly. By dispatching an event once all viewer files are loaded but *before* the viewer initialization has run, setting `AppOptions` during load (in custom implementations of the default viewer) should hopefully become a little bit easier[1].

/cc @timvandermeij 

---
[1] In hindsight, this should probably have been implemented when the global `PDFJS` object was removed...

### DIFF
--- a/web/viewer.js
+++ b/web/viewer.js
@@ -203,6 +203,16 @@ function webViewerLoad() {
 
     window.PDFViewerApplication = pdfjsWebApp.PDFViewerApplication;
     window.PDFViewerApplicationOptions = pdfjsWebAppOptions.AppOptions;
+
+    if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
+      // Give custom implementations of the default viewer a simpler way to
+      // set various `AppOptions`, by dispatching an event once all viewer
+      // files are loaded but *before* the viewer initialization has run.
+      const event = document.createEvent('CustomEvent');
+      event.initCustomEvent('webviewerloaded', true, true, {});
+      document.dispatchEvent(event);
+    }
+
     pdfjsWebApp.PDFViewerApplication.run(config);
   }
 }
--- a/web/app.js
+++ b/web/app.js
@@ -169,21 +169,14 @@ let PDFViewerApplication = {
    * @private
    */
   async _readPreferences() {
-    // A subset of the Preferences that `AppOptions`, for compatibility reasons,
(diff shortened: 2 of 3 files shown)


web/viewer.js
```diff
@@ -203,6 +203,16 @@ function webViewerLoad() {
 
     window.PDFViewerApplication = pdfjsWebApp.PDFViewerApplication;
     window.PDFViewerApplicationOptions = pdfjsWebAppOptions.AppOptions;
+
+    if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
+      // Give custom implementations of the default viewer a simpler way to
+      // set various `AppOptions`, by dispatching an event once all viewer
+      // files are loaded but *before* the viewer initialization has run.
+      const event = document.createEvent('CustomEvent');
+      event.initCustomEvent('webviewerload', true, true, {});
```
timvandermeij: Would it be better to call this `webviewerloaded` to match https://github.com/mozilla/pdf.js/blob/master/web/app.js#L870?

Snuffleupagus: That's a really good point about consistency; however considering that the entire viewer isn't actually loaded at this point, that name seems ever so slightly "off" to me.
How about `webviewerinit` instead, since there's already a bunch of similarly named events (e.g. documentinit, baseviewerinit, pagesinit)?

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #10320, opened 2018-11-29
Add German translation for find_count*

jsonn:


timvandermeij:
From https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md:

> Note that the translations for PDF.js in the `l10n` folder are synchronized with the Nightly channel of Mozilla Firefox. This means that we will only accept pull requests that add strings currently missing in the Nightly channel, but keep in mind that the changes will be overwritten when we synchronize again.

Even though these are indeed additions, I see that they are already available in the Nightly channel at https://hg.mozilla.org/l10n-central/de/raw-file/default/browser/pdfviewer/viewer.properties, so they will automatically be imported with the next resync. Hence, there is no need to make these changes. Thank you.

PR #10323, opened 2018-12-01
Test the code as-is, in Node.js/Travis, rather than its Babel translated version

Snuffleupagus:
This patch does three things:

 - Updates the `gulp unittestcli` command, using `gulp lint` as a guide, such that it can be run locally on Windows without any modifications.

 - Updates the `gulp lib` command to support disabling of Babel through the `SKIP_BABEL` environment variable. Note that all other build targets support this mode, and there's no good reason for `lib` to be any different here.

 - Updates the `npm test` command, used in Node.js/Travis, to test the code as-is test. Since modern Node.js versions seem to have no problems with ES6 compatible code in general, we should just test the source code as-is instead (similar to the tests running on the regular bots).

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -898,11 +898,12 @@ gulp.task('lib', ['buildnumber'], function () {
     };
   }
   function preprocess(content) {
-    var noPreset = /\/\*\s*no-babel-preset\s*\*\//.test(content);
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #10334, opened 2018-12-05
[api-minor] Add support for OpenAction destinations (issue 10332)

Snuffleupagus:
Note that the OpenAction dictionary may contain other information besides just a destination array, e.g. instructions for auto-printing[1].
Given first of all that an arbitrary `Dict` cannot be sent from the Worker (since cloning would fail), and second of all that the data obviously needs to be validated, this patch purposely only adds support for fetching a destination from the OpenAction entry[2].

~~For the time being support for this isn't added to the default viewer, since it's really not clear to me if that's desired functionality. Furthermore, actually implementing this will require (some kind of) re-factoring of the `PDFHistory` initialization to prevent odd initial history state.~~

---
[1] This information is, currently in PDF.js, being included through the `getJavaScript` API method.

[2] This significantly reduces the complexity of the implementation, which seems fine for now. If there's ever need for other kinds of OpenAction to be fetched, additional API methods could/should be implemented as necessary (could e.g. follow the `getOpenActionWhatever` naming scheme).

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -117,6 +117,10 @@
       ],
       "default": 0
     },
+    "disableOpenActionDestination": {
+      "type": "boolean",
+      "default": true
+    },
     "disablePageLabels": {
       "type": "boolean",
       "default": false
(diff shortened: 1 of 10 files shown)


timvandermeij (approved):

timvandermeij:

PR #10337, opened 2018-12-06
Turn on ESLint in examples

wojtekmaj:
Inspired by #10322, I thought that more often than not, it would be handy to have ESLint in examples directory as well, checking if we don't have any preventable mistakes in the code.

* Enabled ESLint in examples directory
* Ran `eslint --fix`
* Manually fixed remaining 17 errors (mainly max-len and no-redeclare)

### DIFF
--- a/examples/.eslintrc
+++ b/examples/.eslintrc
@@ -1,0 +1,24 @@
+{
+  "extends": [
+    "../.eslintrc"
+  ],
+
+  "parserOptions": {
+    "ecmaVersion": 5,
+  },
+
+  "env": {
+    "es6": false,
+  },
+
+  "globals": {
+    "pdfjsImageDecoders": false,
+    "pdfjsLib": false,
+    "pdfjsViewer": false,
+    "Uint8Array": false,
+  },
+
+  "rules": {
+    "object-shorthand": ["error", "never"]
+  }
+}
--- a/examples/browserify/gulpfile.js
+++ b/examples/browserify/gulpfile.js
@@ -9,7 +9,7 @@ var OUTPUT_PATH = '../../build/browserify';
 var TMP_FILE_PREFIX = '../../build/browserify_';
 
 gulp.task('build-bundle', function() {
-  return browserify('main.js', {output: TMP_FILE_PREFIX + 'main.tmp'})
+  return browserify('main.js', { output: TMP_FILE_PREFIX + 'main.tmp', })
     .ignore(require.resolve('pdfjs-dist/build/pdf.worker')) // Reducing size
     .bundle()
     .pipe(source(TMP_FILE_PREFIX + 'main.tmp'))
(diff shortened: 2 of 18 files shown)


examples/.eslintrc
```diff
@@ -0,0 +1,11 @@
...
+  "extends": [
+    "../.eslintrc"
+  ],
+
+  "globals": {
+    "pdfjsImageDecoders": false,
+    "pdfjsLib": false,
+    "pdfjsViewer": false,
+  },
+}
```
Snuffleupagus: Since the (various) builds are still transpiled to ES5 compatible code by default, the examples ought to be ES5 compatible as well. Hence please set the `"object-shorthand": "off",` rule here and revert the relevant changes in the next commit.

examples/browserify/gulpfile.js
```diff
@@ -21,12 +21,12 @@ gulp.task('build-bundle', function() {
...
-  return browserify(workerSrc, {output: TMP_FILE_PREFIX + 'worker.tmp'})
+  return browserify(workerSrc, { output: TMP_FILE_PREFIX + 'worker.tmp', })
     .bundle()
     .pipe(source(TMP_FILE_PREFIX + 'worker.tmp'))
-    .pipe(streamify(uglify({compress:{
-      sequences: false // Chrome has issue with the generated code if true
-    }})))
+    .pipe(streamify(uglify({ compress: {
+      sequences: false, // Chrome has issue with the generated code if true
+    }, })))
```
Snuffleupagus: This isn't very readable, even as-is in `master`, let's reformat along the lines of:
```javascript
.pipe(streamify(uglify({
  compress: {
    sequences: false, // Chrome has issue with the generated code if true
  },
})))
```

examples/node/domstubs.js
```diff
@@ -135,6 +135,8 @@ function DOMElementSerializer(node) {
   this._childSerializer = null;
 }
 DOMElementSerializer.prototype = {
+  /* eslint-disable no-fallthrough */
```
Snuffleupagus: Please just add `/* falls through */` were applicable instead, rather than blanket disable like this.

examples/node/pdf2png/pdf2png.js
```diff
@@ -81,7 +82,9 @@ loadingTask.promise.then(function(pdfDocument) {
         if (error) {
           console.error('Error: ' + error);
         } else {
-          console.log('Finished converting first page of PDF file to a PNG image.');
+          console.log(
+            'Finished converting first page of PDF file to a PNG image.'
+          );
```
Snuffleupagus: For consistency: In the PDF.js code-base, this is formatted as:
```javascript
console.log(
  'Finished converting first page of PDF file to a PNG image.');
```

wojtekmaj:
@Snuffleupagus Fixed all the things I believe, and rebased to clean the mess :)

Snuffleupagus:
> Fixed all the things I believe, and rebased to clean the mess :)

That looks better, but it doesn't seem necessary for this to be *five* commits:
 - Please just fold the `Turn off ES6 syntax in examples folder, with the exception of Uint8Array` commit into the main `Turn on ESLint in examples directory, apply examples-specific exceptions` commit.
 - Also let's fold the `Format gulpfile.js for readibility` (nit: spelling) commit into `Manually fix remaining ESLint errors` commit.

wojtekmaj:
@Snuffleupagus Done!

timvandermeij (approved):

timvandermeij:
Looks good; thank you for improving this!

PR #10340, opened 2018-12-07
Tweak the `PDFHistory._tryPushCurrentPosition` method to work better with the different Scroll/Spread viewer modes

Snuffleupagus:
Given that a larger number of pages may now be visible at once, and importantly that their layout may be non-vertical, one of the conditions should be tweaked to not accidentally miss cases where a page is still visible.

Please note: This patch is based on code-inspection, and the only ill effect occurring without it would be a couple of (near) duplicate history entries in some *rare* edge-cases.

### DIFF
--- a/web/pdf_history.js
+++ b/web/pdf_history.js
@@ -329,8 +329,8 @@ class PDFHistory {
     }
 
     let forceReplace = false;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):
Thank you!

Issue #10344, opened 2018-12-10
Document Info Dict data are completely ignored in `getMetadata()`


Voileexperiments:
I'm trying [this pdf](https://arxiv.org/pdf/1703.01161.pdf) via #8180. There is an Document Info Dict entry in this document:

```
PTEX.Fullbanner: "This is pdfTeX, Version 3.14159265-2.6-1.40.17 (TeX Live 2016) kpathsea version 6.2.2"
```

However it is not shown inside the result of `getMetadata()`.

The problem lies within `get documentInfo()` in default `pdf.worker.js` (I don't know where these lines belong in the source, so I'm only referring to the dist version from the [online demo](https://mozilla.github.io/pdf.js/web/viewer.html)):

```javascript
get documentInfo() {
      var docInfo = {
        PDFFormatVersion: this.pdfFormatVersion,
        IsLinearized: !!this.linearization,
        IsAcroFormPresent: !!this.acroForm,
        IsXFAPresent: !!this.xfa
      };
      var infoDict;

      try {
        infoDict = this.xref.trailer.get('Info');
      } catch (err) {
        if (err instanceof _util.MissingDataException) {
          throw err;
        }

        (0, _util.info)('The document information dictionary is invalid.');
      }

      if ((0, _primitives.isDict)(infoDict)) {
        for (var key in DocumentInfoValidators) {
          if (infoDict.has(key)) {
            var value = infoDict.get(key);

            if (DocumentInfoValidators[key](value)) {
              docInfo[key] = typeof value !== 'string' ? value : (0, _util.stringToPDFString)(value);
            } else {
              (0, _util.info)('Bad value in document info for "' + key + '"');
            }
          }
        }
      }

      return (0, _util.shadow)(this, 'documentInfo', docInfo);
    },
```

And earlier in the code, `DocumentInfoValidators`:

```javascript
var DocumentInfoValidators = {
    Title: _util.isString,
    Author: _util.isString,
    Subject: _util.isString,
    Keywords: _util.isString,
    Creator: _util.isString,
    Producer: _util.isString,
    CreationDate: _util.isString,
    ModDate: _util.isString,
    Trapped: _primitives.isName
  };
```

`getMetadata` calls this:

```
handler.on('GetMetadata', function wphSetupGetMetadata(data) {
      return Promise.all([pdfManager.ensureDoc('documentInfo'), pdfManager.ensureCatalog('metadata')]);
    });
```

[...]

PR #10345, opened 2018-12-10
#10344 fix

Voileexperiments:
ref: https://github.com/mozilla/pdf.js/issues/10344

Iterate all keys in `infoDict` instead of all keys in `DocumentInfoValidators` which `infoDict` has the same key of, so as to not ignore the custom info dict entries.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -545,16 +545,14 @@ var PDFDocument = (function PDFDocumentClosure() {
       }
       if (isDict(infoDict)) {
         // Only fill the document info with valid entries from the spec.
-        for (let key in DocumentInfoValidators) {
-          if (infoDict.has(key)) {
-            const value = infoDict.get(key);
-            // Make sure the value conforms to the spec.
-            if (DocumentInfoValidators[key](value)) {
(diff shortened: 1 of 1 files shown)


src/core/document.js
```diff
@@ -545,16 +545,15 @@ var PDFDocument = (function PDFDocumentClosure() {
       }
       if (isDict(infoDict)) {
         // Only fill the document info with valid entries from the spec.
```
Snuffleupagus: This patch is very clearly breaking the assumption listed above, which exists to prevent errors since not all data structures (one example is `Dict`) can be sent from the worker to the main thread.

Assuming first of all that this is even desired functionality, the patch unfortunately cannot be accepted in its current form (furthermore there's also the issue of no tests being include here).

Voileexperiments: Okay, then *what* is the intended functionality? Info dict is clearly in the specs since every PDF accepts custom info dict entries. I do not see any discussions among the issues about this matter.

timvandermeij:
Closing in favor of the more complete solution in #10346.

PR #10350, opened 2018-12-12
Ignore the `fieldValue` for Signature annotations, since they're currently unsupported (issue 10374)

### LINKED ISSUE #10347: getAnnotations() promise not being resolved when page contains signature
When a pdf page contains 'Sig' type widget annotations and those annotations have a signature as their fieldValue, the promise returned by page.getAnnotations() does not resolve.

PDF with 'normal' annotations: http://www.pdfill.com/example/pdf_commenting_new.pdf
PDF with signature annotation: https://blogs.adobe.com/security/SampleSignedPDFDocument.pdf

Steps to reproduce:
```javascript
// PDF without signature annotations works properly, use {data: pdfData} for a demonstration
var loadingTask = pdfjsLib.getDocument({data: pdfDataSigned});
  console.log('PDF loaded');
  
  // Fetch the first page
  var pageNumber = 1;
  pdf.getPage(pageNumber).then(function(page) {
    console.log('Page loaded');
    page.getAnnotations().then(function(annotations) {
        // this is never executed if the page has signatures
        console.log('annotations loaded');
        for (let i = 0; i < annotations.length; i++) {
          console.log(annotations[i].subtype);
        }
      }, function (reason) {
      // this also does not happen if the page has signatures
      // Annotations loading error
      console.log(reason);
      });    
  });
}, function (reason) {
  // PDF loading error
  console.error(reason);
});
```
`pdfDataSigned` and `pdfData` are included in the fiddle as base64, which is too long to include here.
Fiddle demo (I recommend scrolling to the end of the javascript panel):
https://jsfiddle.net/luk_lo/mq9nyadh/1/

This may be related to #8429 but the provided fix is not helpful as it will prevent signature annotations from being returned by getAnnotations.
This bug is probably caused by the large fieldValue of the AnnotationsWidget as setting `this.fieldValue = true;` after the line https://github.com/mozilla/pdf.js/blob/2f4c7e01f5871f77e4e03f78525c330de43dbe76/src/core/annotation.js#L618 fixes the issue (of course the signature annotations returned by getAnnotations then do not have their proper fieldValue).

Snuffleupagus:
Given that Signature (Widget) annotations are currently not supported, since they cannot be validated, simply ignoring the `fieldValue` seems OK for now considering that attempting to blindly include unparsed/unvalidated data isn't very useful.

Fixes #10347.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -613,8 +613,11 @@ class WidgetAnnotation extends Annotation {
 
     data.readOnly = this.hasFieldFlag(AnnotationFieldFlag.READONLY);
 
-    // Hide signatures because we cannot validate them.
+    // Hide signatures because we cannot validate them, and unset the fieldValue
+    // since it's (most likely) a `Dict` which is non-serializable and will thus
+    // cause errors when sending annotations to the main-thread (issue 10347).
     if (data.fieldType === 'Sig') {
+      data.fieldValue = null;
       this.setFlags(AnnotationFlag.HIDDEN);
     }
   }


timvandermeij (approved):

timvandermeij:
I think this is fine too. Thank you for fixing this!

PR #10351, opened 2018-12-12
Attempt to ignore mouse wheel zooming during tab switches (bug 1503412)

Snuffleupagus:
This patch re-factors, and extends, the already existing `zoomDisabledTimeout` used during mouse wheel zooming.
Unfortunately I haven't got the required hardware to actually test this patch, but there's a decent chance that it will fix, or at least reduce, the problems reported in https://bugzilla.mozilla.org/show_bug.cgi?id=1503412.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #10354, opened 2018-12-13
Disable source-map generation when running tests

Snuffleupagus:
This should save, a little bit of, time/resources on the bots since source-maps aren't used for anything during testing.


timvandermeij (approved):

timvandermeij:
It's always a good thing to save resources. Thank you!

PR #10356, opened 2018-12-15
Use high DPI icons on semi-high DPI screens.

brendandahl:
The icons looks really fuzzy on a machine with a device pixel ratio of
1.5. Using the 2x icons looks much better.


timvandermeij (approved):

timvandermeij:
Thank you!

Issue #10315, opened 2018-11-28
Keyboard shortcut to show/hide sidebar?


GandalfSaxe:
I'd love to see a keyboard shortcut added to show/hide the outline sidebar :)

timvandermeij:
Fixed by the pull request above.

PR #10358, opened 2018-12-16
Add keyboard shortcut to toggle sidebar (F4, same as Adobe Reader).

tsilvap:
Add a keyboard shortcut to toggle the sidebar, addressing issue #10315

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2313,6 +2313,10 @@ function webViewerKeyDown(evt) {
       case 82: // 'r'
         PDFViewerApplication.rotatePages(90);
         break;
+
+      case 115: // F4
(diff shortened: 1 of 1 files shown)


tsilvap:
> Adobe Reader uses F4, and given that it's basically the PDF reference implementation it would make much more sense to use the same keyboard shortcut.

That makes sense. I suppose Adobe Reader has a much larger userbase as well, so <kbd>F4</kbd> will probably be more familiar to most users. I'll change the shortcut and amend the commit. (I'm assuming that's what I should do? I'm kind of new to this.)

timvandermeij (approved):

timvandermeij:
Thank you!

PR #10369, opened 2018-12-21
[api-minor] Change the `getViewport` method, on `PDFPageProxy`, to take a parameter object rather than a bunch of (randomly) ordered parameters

Snuffleupagus:
If, as PR #10368 suggests, more parameters should be added to `getViewport` I think that it would be a mistake to not change the signature *first* to avoid needlessly unwieldy call-sites.

To not break any existing code and third-party use-cases, this is obviously implemented with a deprecation warning *and* with a working fallback[1] for the old method signature.

/cc @timvandermeij 

---
[1] This is limited to `GENERIC` builds, which should be sufficient.

### DIFF
--- a/examples/acroforms/acroforms.js
+++ b/examples/acroforms/acroforms.js
@@ -37,7 +37,7 @@ loadingTask.promise.then(function(doc) {
           container: container,
           id: pageNum,
           scale: DEFAULT_SCALE,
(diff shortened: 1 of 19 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

PR #10376, opened 2018-12-23
Convert `src/core/chunked_stream.js` to ES6 syntax

timvandermeij:
While we're here, remove the unused `url` argument for the `ChunkedStreamManager` class that I found during the refactoring.

@Snuffleupagus Would you be willing to review this? I found that GitHub can't apply the `?w=1` flag for the first commit for some reason, so here is the output of `git show -w HEAD~1`: https://dpaste.de/vUQU

### DIFF
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -264,7 +264,6 @@ class ChunkedStreamManager {
     this.chunkSize = args.rangeChunkSize;
     this.stream = new ChunkedStream(this.length, this.chunkSize, this);
     this.pdfNetworkStream = pdfNetworkStream;
-    this.url = args.url;
     this.disableAutoFetch = args.disableAutoFetch;
     this.msgHandler = args.msgHandler;
 
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -18,8 +18,8 @@ import {
   MissingDataException
 } from '../shared/util';
 
-var ChunkedStream = (function ChunkedStreamClosure() {
-  function ChunkedStream(length, chunkSize, manager) {
+class ChunkedStream {
+  constructor(length, chunkSize, manager) {
     this.bytes = new Uint8Array(length);
     this.start = 0;
     this.pos = 0;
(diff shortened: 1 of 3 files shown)


src/core/chunked_stream.js
```diff
@@ -30,252 +30,238 @@ var ChunkedStream = (function ChunkedStreamClosure() {
...
+    const chunkSize = this.chunkSize;
+    if (begin % chunkSize !== 0) {
+      throw new Error(`Bad begin offset: ${begin}`);
+    }
+
+    // Using `this.length` is inaccurate here since `this.start` can be moved
+    // (see the `moveStart` method).
+    const end = begin + chunk.byteLength;
+    const length = this.bytes.length;
+    if (end % chunkSize !== 0 && end !== length) {
```
Snuffleupagus: Nit: Since `length` appears to be used only once, how do you feel about simply inlining `... && end !== this.bytes.length) {` here instead?

src/core/chunked_stream.js
```diff
@@ -30,252 +30,238 @@ var ChunkedStream = (function ChunkedStreamClosure() {
...
+    this.pos -= bytes.length;
+    return bytes;
+  }
+
+  getByteRange(begin, end) {
+    this.ensureRange(begin, end);
+    return this.bytes.subarray(begin, end);
+  }
+
+  skip(n = 1) {
```
Snuffleupagus: This changes the behaviour of the method for `n = 0`, please revert this!

The fact that various stream methods behaves "strangely" in the zero case is bad, but not something that's easily fixed now without auditing/re-writing most of the `/core` code :-(

src/core/chunked_stream.js
```diff
@@ -30,252 +30,238 @@ var ChunkedStream = (function ChunkedStreamClosure() {
...
+  }
+
+  moveStart() {
+    this.start = this.pos;
+  }
+
+  makeSubStream(start, length, dict) {
+    this.ensureRange(start, start + length);
+
+    function ChunkedStreamSubstream() {}
```
Snuffleupagus: Could this construction be changed to a class instead, or does that not work?

```javascript
class ChunkedStreamSubstream extends ChunkedStream {
  constructor() {}

  getMissingChunks() {
    ...
  }
}

File: src/core/chunked_stream.js [hunk 4]
```diff
@@ -290,283 +276,274 @@ var ChunkedStreamManager = (function ChunkedStreamManagerClosure() {
...
-      for (var chunk in chunksNeeded) {
-        chunk = chunk | 0;
-        if (!(chunk in this.requestsByChunk)) {
-          this.requestsByChunk[chunk] = [];
-          chunksToRequest.push(chunk);
-        }
-        this.requestsByChunk[chunk].push(requestId);
+    const chunksNeeded = Object.create(null);
+    this.chunksNeededByRequest[requestId] = chunksNeeded;
+    for (let i = 0, ii = chunks.length; i < ii; i++) {
```
Snuffleupagus: Could potentially use `for (const chunk of chunks) {` here instead.

src/core/chunked_stream.js
```diff
@@ -290,283 +276,274 @@ var ChunkedStreamManager = (function ChunkedStreamManagerClosure() {
...
+    if (!chunksToRequest.length) {
       return capability.promise;
-    },
+    }
 
-    getStream: function ChunkedStreamManager_getStream() {
-      return this.stream;
-    },
+    const groupedChunksToRequest = this.groupChunks(chunksToRequest);
+    for (let i = 0; i < groupedChunksToRequest.length; ++i) {
```
Snuffleupagus: Could potentially use `for (const groupedChunked of groupedChunksToRequest) {` here instead.

src/core/chunked_stream.js
```diff
@@ -290,283 +276,274 @@ var ChunkedStreamManager = (function ChunkedStreamManagerClosure() {
...
-      for (var i = 0; i < ranges.length; i++) {
-        var beginChunk = this.getBeginChunk(ranges[i].begin);
-        var endChunk = this.getEndChunk(ranges[i].end);
-        for (var chunk = beginChunk; chunk < endChunk; ++chunk) {
-          if (!chunksToRequest.includes(chunk)) {
-            chunksToRequest.push(chunk);
-          }
+  requestRanges(ranges = []) {
+    const chunksToRequest = [];
+    for (let i = 0; i < ranges.length; i++) {
```
Snuffleupagus: Could potentially use `for (const range of ranges) {` here instead.

src/core/chunked_stream.js
```diff
@@ -290,283 +276,274 @@ var ChunkedStreamManager = (function ChunkedStreamManagerClosure() {
...
-    onReceiveData: function ChunkedStreamManager_onReceiveData(args) {
-      var chunk = args.chunk;
-      var isProgressive = args.begin === undefined;
-      var begin = isProgressive ? this.progressiveDataLength : args.begin;
-      var end = begin + chunk.byteLength;
-
-      var beginChunk = Math.floor(begin / this.chunkSize);
-      var endChunk = end < this.length ? Math.floor(end / this.chunkSize) :
-                                         Math.ceil(end / this.chunkSize);
+    for (let i = 0; i < chunks.length; ++i) {
```
Snuffleupagus: Could potentially use `for (const chunk of chunks) {` here instead.

src/core/chunked_stream.js
```diff
@@ -290,283 +276,274 @@ var ChunkedStreamManager = (function ChunkedStreamManagerClosure() {
...
+      this._loadedStreamCapability.resolve(this.stream);
+    }
+
+    const loadedRequests = [];
+    for (let chunk = beginChunk; chunk < endChunk; ++chunk) {
+      // The server might return more chunks than requested.
+      const requestIds = this.requestsByChunk[chunk] || [];
+      delete this.requestsByChunk[chunk];
+
+      for (let i = 0; i < requestIds.length; ++i) {
```
Snuffleupagus: Could potentially use `for (const requestId of requestIds) {` here instead.

src/core/chunked_stream.js
```diff
@@ -290,283 +276,274 @@ var ChunkedStreamManager = (function ChunkedStreamManagerClosure() {
...
-      if (this.pdfNetworkStream) {
-        this.pdfNetworkStream.cancelAllRequests('abort');
-      }
-      for (var requestId in this.promisesByRequest) {
-        var capability = this.promisesByRequest[requestId];
-        capability.reject(new Error('Request was aborted'));
-      }
-    },
-  };
+    for (let i = 0, ii = loadedRequests.length; i < ii; ++i) {
```
Snuffleupagus: Could potentially use `for (const requestId of loadedRequests) {` here instead.

Snuffleupagus (changes requested):
Looks generally good, thank you; but setting r- for the `skip` method alone.

src/core/chunked_stream.js
timvandermeij: Unfortunately not in this case, because the variable `i` is used a few lines below.

src/core/chunked_stream.js
timvandermeij: I tried this, but I couldn't get it to work. I think it has something to do with extending the class that you're in.

Snuffleupagus (approved):

PR #10381, opened 2018-12-26
support urlencode contentDispositionFilename

wa8n:


web/app.js
```diff
@@ -1074,6 +1074,9 @@ let PDFViewerApplication = {
         ({ info, metadata, contentDispositionFilename, }) => {
       this.documentInfo = info;
       this.metadata = metadata;
+      if(contentDispositionFilename && contentDispositionFilename.includes('%')){
```
timvandermeij: This does not look like the correct place to fix it. Instead, it should be fixed in the function where the content disposition filename is fetched, most likely `extractFilenameFromHeader`. Moreover, a unit test needs to be added to https://github.com/mozilla/pdf.js/blob/f45e46d7adba9052a6ecd364e3630af858a74a90/test/unit/network_utils_spec.js#L161 to make sure it works. Finally, please elaborate a bit about what the exact problem is in the pull request description.

timvandermeij (changes requested):

timvandermeij:
Closing since this cannot be merged in the current state and no response was given. If you do believe something is wrong in PDF.js, please open an issue so we can discuss it and find out what the best fix would be.

PR #10383, opened 2018-12-26
automatically update URL hash, fixes #5753

### LINKED ISSUE #5753: feature request: page #anchoring
Linking someone to a specific section in a 191 page document is rather cumbersome, and it would be super useful if pdf.js could dynamically update the URL so that scrolling through a document appends a #page1, #page2, etc. for linking at least to a specific page.

KiaraGrouwstra:


timvandermeij:
Does this address the concerns from https://github.com/mozilla/pdf.js/issues/5753#issuecomment-75451859?

web/app.js
```diff
@@ -164,6 +165,9 @@ let PDFViewerApplication = {
     });
 
     this.initialized = true;
+
+    // update url on scroll
```
timvandermeij: You can remove this comment and add it to the `setupHashUpdating` function below so it's clear what the objective of that function is.

web/app.js
```diff
@@ -164,6 +165,9 @@ let PDFViewerApplication = {
...
+    setupHashUpdating();
```
timvandermeij: It feels like this is part of the initialization, so I think it should be moved above `this.initialized = true` above.

web/app.js
```diff
@@ -1769,6 +1773,27 @@ function webViewerSidebarViewChanged(evt) {
   }
 }
 
+function setupHashUpdating() {
+  let tid = false;
```
timvandermeij: Unclear name; what does this represent?

web/app.js
```diff
@@ -1769,6 +1773,27 @@ function webViewerSidebarViewChanged(evt) {
...
+
+  // set up the update function
+  let updatePos = () => {
+    return PDFViewerApplication.pdfHistory.pushCurrentPosition();
```
timvandermeij: I don't think `pushCurrentPosition` returns anything, so I think you can simplify by removing `updatePos` altogether and just calling this method in `setTimeout` below.

web/app.js
```diff
@@ -1769,6 +1773,27 @@ function webViewerSidebarViewChanged(evt) {
...
+  };
+
+  // schedule the update call for some time in the future.
+  let test = (evt) => {
```
timvandermeij: `test` is a bit unclear; maybe `updateHash` is better?

timvandermeij (changes requested):

Snuffleupagus:
Sorry, but this patch really isn't acceptable at all as far as I can tell, since it will be fighting with the (painstakingly implemented) history implementation already present in the default viewer.

First of all, the `PDFHistory` implementation is already *lazily* updating the history (using `updateviewarea` events), which renders this entire patch unnecessary. The only thing that this patch will do is to flood the browser history with lots of useless entries, which really isn't good and which the `PDFHistory` re-implementation sought to avoid.
Second of all, the `PDFHistory` method used here doesn't even touch the document URL so that's also strange (to say the least). However, that isn't at all something that should be "fixed", since hash changes doesn't really fit in great with the general history implementation (simply because the concept of *destinations* in the PDF format isn't a great fit for the web in general).

Quite frankly, this won't work and I don't understand why it's necessary to try and fix the referenced issue. Considering that even a working solution, whatever that would look like, will only add unnecessary complexity when all that's really needed is to press the `viewBookmark` button to obtain the current hash.

@timvandermeij This really ought to be closed as WONTFIX, as far as I'm concerned, and honestly that applies to the issue as well.

PR #10393, opened 2018-12-29
Convert `src/core/document.js` to ES6 syntax

timvandermeij:
The individual commits can be reviewed with the `?w=1` flag.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -12,6 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* eslint no-var: error */
 
 import {
   assert, FormatError, getInheritableProperty, info, isArrayBuffer, isBool,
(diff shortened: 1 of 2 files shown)


src/core/document.js
```diff
@@ -62,138 +61,211 @@ var Page = (function PageClosure() {
...
-        rotate = ((rotate % 360) + 360) % 360;
+  getContentStream() {
+    const content = this.content;
+    let stream;
+
+    if (Array.isArray(content)) {
+      // Fetching the individual streams from the array.
+      const xref = this.xref;
+      const streams = [];
+      for (let i = 0, ii = content.length; i < ii; ++i) {
```
Snuffleupagus: Optional nit: Could probably use `for ... of` instead here.

src/core/document.js
```diff
@@ -62,138 +61,211 @@ var Page = (function PageClosure() {
...
-    },
 
-    loadResources: function Page_loadResources(keys) {
-      if (!this.resourcesPromise) {
-        // TODO: add async `_getInheritableProperty` and remove this.
-        this.resourcesPromise = this.pdfManager.ensure(this, 'resources');
+      // Collect the operator list promises for the annotations. Each promise
+      // is resolved with the complete operator list for a single annotation.
+      const opListPromises = [];
+      for (let i = 0, ii = annotations.length; i < ii; i++) {
```
Snuffleupagus: Optional nit: Could probably use `for ... of` instead here.

src/core/document.js
```diff
@@ -62,138 +61,211 @@ var Page = (function PageClosure() {
...
+            partialEvaluator, task, renderInteractiveForms));
+        }
       }
-      return this.resourcesPromise.then(() => {
-        let objectLoader = new ObjectLoader(this.resources, keys, this.xref);
 
-        return objectLoader.load();
+      return Promise.all(opListPromises).then(function(opLists) {
+        pageOpList.addOp(OPS.beginAnnotations, []);
+        for (let i = 0, ii = opLists.length; i < ii; i++) {
```
Snuffleupagus: Optional nit: Could probably use `for ... of` instead here.

src/core/document.js
```diff
@@ -351,319 +372,302 @@ var PDFDocument = (function PDFDocumentClosure() {
...
-        var hex = hash[i].toString(16);
-        fileID += hex.length === 1 ? '0' + hex : hex;
+  // Find the header, get the PDF format version and setup the
+  // stream to start from the header.
+  checkHeader() {
+    const stream = this.stream;
+    stream.reset();
+
+    if (!find(stream, '%PDF-', 1024)) {
+      // May not be a PDF file.
```
Snuffleupagus: Given the old comment `May not be a PDF file, continue anyway.`, how about adjusting this one to something like `May not be a PDF file, but don't throw an error and let parsing continue.` instead?

src/core/document.js
```diff
@@ -351,319 +372,302 @@ var PDFDocument = (function PDFDocumentClosure() {
...
-          this._pagePromises.length = 0;
-          this.cleanup();
-
-          throw new XRefParseException();
+      hash = calculateMD5(this.stream.bytes.subarray(0,
+        FINGERPRINT_FIRST_BYTES), 0, FINGERPRINT_FIRST_BYTES);
+    }
+
+    let fingerprint = '';
+    for (let i = 0, ii = hash.length; i < ii; i++) {
```
Snuffleupagus: Optional nit: Could probably use `for ... of` instead here.

Snuffleupagus (approved):
Looks good to me, thank you!

I've left a couple of minor comments, and would also like to suggest adding `/* eslint no-var: error */
` to the top of files that you convert to ES6 to prevent future usage of `var`.

src/core/document.js
timvandermeij: Somehow I keep forgetting about this. Thank you!

PR #10394, opened 2018-12-29
Optimize the `Ref` class in `src/core/primitives.js`

timvandermeij:
The `toString` method always creates two string objects (for the 'R' character and for the `num` concatenation) and in the worst case creates three string objects (one more for the `gen` concatenation). For the Tracemonkey paper alone, this resulted in 12000 string objects when scrolling from the top to the bottom of the document. Since this is a hot function, it's worth minimizing the number of string objects, especially for large documents, to reduce peak memory usage.

This commit refactors the `toString` method to always create only one string object.


Snuffleupagus (approved):

timvandermeij:
Thank you for the review, @Snuffleupagus!

PR #10399, opened 2019-01-01
migrate to canvas 2.x api

MohammedEssehemy:
npm install canvas will download version 2.x in which the [api changed](https://github.com/Automattic/node-canvas/blob/master/CHANGELOG.md#200)


timvandermeij (approved):

timvandermeij:
Thank you! I verified locally that this makes the example work again.

PR #10408, opened 2019-01-03
Prevent errors, because of incorrect scope, in the `XMLParserBase._resolveEntities` method (issue 10407)

### LINKED ISSUE #10407: PDFjs.getMetadata() - XML parse bug,  _resolveEntities(s)
When calling getMetadata() on a pdf which I can not provide I get the following error:

```bash
TypeError: Cannot read property 'onResolveEntity' of undefined
```
The error is being traced back to line 65 of xml_parser.js:

https://github.com/mozilla/pdf.js/blob/1b84b2ed603a18974c701381f67ddca4b392f569/src/display/xml_parser.js#L48-L67

The problem is `this` of the string replace function will always undefined. The solution would be be to save a reference variable `var self = this;` in the outer scope and call `self.onResolveEntity()` on that reference variable inside the string replace function.

The PDF which causes this error has the string input to the `_resolveEntities(s)`:
```bash
s: '&#10;                                xmlns:xap=&apos;http://ns.adobe.c'
```
Therefore, two entities will be parsed: `#10` and `apos`. The error happens when the entity value is `apos`.

Snuffleupagus:
Now with a unit-test, courtesy of https://github.com/mozilla/pdf.js/pull/10408#issuecomment-451294285.

Fixes #10407.

### DIFF
--- a/src/display/xml_parser.js
+++ b/src/display/xml_parser.js
@@ -46,7 +46,7 @@ function isWhitespaceString(s) {
 
 class XMLParserBase {
   _resolveEntities(s) {
-    return s.replace(/&([^;]+);/g, function (all, entity) {
+    return s.replace(/&([^;]+);/g, (all, entity) => {
       if (entity.substring(0, 2) === '#x') {
         return String.fromCharCode(parseInt(entity.substring(2), 16));
       } else if (entity.substring(0, 1) === '#') {
--- a/test/unit/metadata_spec.js
+++ b/test/unit/metadata_spec.js
@@ -127,4 +127,23 @@ describe('metadata', function() {
 
     expect(isEmptyObj(metadata.getAll())).toEqual(true);
   });
+
+  it('should correctly handle metadata containing "&apos" (issue 10407)',
+      function() {
+    const data = '<x:xmpmeta xmlns:x=\'adobe:ns:meta/\'>' +
+      '<rdf:RDF xmlns:rdf=\'http://www.w3.org/1999/02/22-rdf-syntax-ns#\'>' +
+      '<rdf:Description xmlns:dc=\'http://purl.org/dc/elements/1.1/\'>' +
+      '<dc:title><rdf:Alt>' +
+      '<rdf:li xml:lang="x-default">&apos;Foo bar baz&apos;</rdf:li>' +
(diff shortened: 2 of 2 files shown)


timvandermeij:
I looked into this and managed to create a unit test:

```diff
+  it('should resolve entities correctly (issue 10407)', function() {
+    const data = '<x:xmpmeta xmlns:x=\'adobe:ns:meta/\'>' +
+      '<rdf:RDF xmlns:rdf=\'http://www.w3.org/1999/02/22-rdf-syntax-ns#\'>' +
+      '<rdf:Description xmlns:dc=\'http://purl.org/dc/elements/1.1/\'>' +
+      '<dc:title><rdf:Alt><rdf:li xml:lang="x-default">&apos;Foo bar baz&apos;</rdf:li>' +
+      '</rdf:Alt></dc:title></rdf:Description></rdf:RDF></x:xmpmeta>';
+    const metadata = new Metadata(data);
+
+    expect(metadata.has('dc:title')).toBeTruthy();
+    expect(metadata.has('dc:qux')).toBeFalsy();
+
+    expect(metadata.get('dc:title')).toEqual('\'Foo bar baz\'');
+    expect(metadata.get('dc:qux')).toEqual(null);
+
+    expect(metadata.getAll()).toEqual({ 'dc:title': '\'Foo bar baz\'', });
+  });
```

It's equal to the first unit test in the file with the title now containing `&apos;`, which needs to be resolved. I confirmed that this breaks without your fix and works with your fix, but please check as well and if it works, let's add it in this PR.

dhollenbeck:
Sorry about not providing a test pdf on the original issue. However, I would like to provide one now.
[issue-10407.pdf](https://github.com/mozilla/pdf.js/files/2725500/issue-10407.pdf)

I was able to reproduce a test case PDF file with all of the sensitive data striped out by using the following steps:
- pdftk ./source.pdf burst
- qpdf.exe pg_0001.pdf --pages qpdf-manual.pdf 1 -- out.pdf

timvandermeij:
Thank you for providing this, @dhollenbeck! It will definitely help us to verify the fix with the actual bad metadata.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #10412, opened 2019-01-04
Prevent errors, in `SimpleXMLParser.onEndElement`, when the stack has already been completely parsed (issue 10410)

### LINKED ISSUE #10410: PDFjs.getMetadata() - XML parse bug, onEndElement(name)
The attached test PDF throws an error in onEndElement.

https://github.com/mozilla/pdf.js/blob/1b84b2ed603a18974c701381f67ddca4b392f569/src/display/xml_parser.js#L366-L372

The problem is `this._stack.pop()` will return undefined when `this._stack` is an empty array. This will set `this._currentFragment` to undefined and line 368 will throw an error:

```bash
TypeError: Cannot read property 'length' of undefined
```

Test file: [toissue.pdf](https://github.com/mozilla/pdf.js/files/2725637/toissue.pdf)

For what it is worth I have recently looped 42,000 pdf files using PDFjs to extract meta data from them. PDFjs has worked very well with the exception of recently submitted issues regarding metadata. I did front run PDFjs with `qpdf --check` to filter out any corrupt PDF files. I used PDFjs because in addition to extracting metadata I also need to extract page content. In all, I extracted 6.3 million pages of content using PDFjs.

**Thank you very much for PDFjs!**

Snuffleupagus:
The error was triggered for a particular set of metadata, where an end tag was encountered without the corresponding begin tag being present in the data.
(The patch also fixes a minor oversight, from a recent PR, in the `SimpleDOMNode.nextSibling` method.)

Fixes #10410.

### DIFF
--- a/src/display/xml_parser.js
+++ b/src/display/xml_parser.js
@@ -287,6 +287,9 @@ class SimpleDOMNode {
       return undefined;
     }
     const index = childNodes.indexOf(this);
+    if (index === -1) {
+      return undefined;
+    }
     return childNodes[index + 1];
   }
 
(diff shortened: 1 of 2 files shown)


timvandermeij:
Looks good after a rebase.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #10419, opened 2019-01-05
[Proposal, partially done] Split UI utils for better tree shaking

wojtekmaj:
The idea behind this is motivated by two things:

* ui_utils is an enormous file, with many dependencies which are hard to follow. Splitting it into multiple components should make it more understandable.
* Using parts of lib, like we import PDFLinkService at React-PDF, imports the whole ui_utils file, with all its dependencies (like `@babel/runtime`). This is a more selfish motivation than the first one, but I think still  valid.

What do you think about this approach? Should I continue?

No code has been altered - only moved and added imports/exports if necessary.


timvandermeij:
I'm not against the idea of splitting up the file, but I'm a bit confused why it's necessary. I thought Webpack should already take care of only importing the classes/functions you actually use. I'm not a Webpack expert, so I may be wrong here though!

timvandermeij:
Having thought about this some more, I came to the conclusion that I mostly agree with https://github.com/mozilla/pdf.js/pull/10419#issuecomment-451651532, so I'm closing this. What we can consider is moving some utility functions closer to the code they're actually for (because some are only used in one or two places), but that wouldn't solve the more general import issue you appear to be having, so I don't think that's worth the effort. We can also consider patches that tweak tree shaking in Webpack. Other than that, I don't really see how importing the `ui_utils.js` file leads to Babel being required since it's not directly used in there, so that tells me that the usage of the files or something else is the actual problem here.

PR #10423, opened 2019-01-06
Add support for updating the document hash, off by default, when the browser history is updated (issue 5753)

Snuffleupagus:
This is *really* the best that we can do here, since other proposed solutions would interfere with (and break) the painstakingly implemented browsing history that's present in the default viewer.

I'm still not convinced that this is a good idea in general, but this patch implements it in a way where it is possible to toggle[1] for users that wish to have this feature. In particular, there's a couple of reasons why I'm not finding this feature necessary/great:
 - It's already possible to easily obtain the current hash, by simply clicking on the `viewBookmark` button at any time.
 - Hash changes requires a bit of special handling[2], i.e. extra code, to prevent issues when the browser history is traversed (see `PDFHistory._popState`). Currently this is only necessary when the user has manually changed the hash, with this patch it will always be the case (assuming the feature is active).
 - It's not always possible to change the URL when updating the browser history. For example: In the Firefox built-in viewer, the URL cannot be modified for local files (i.e. those using the `file://` protocol).
This leads to inconsistent behaviour, and may in some cases even result in errors being thrown and the history thus not updating, if the browser prevents changes to the URL during `pushState`/`replaceState` calls.

---
[1] Using the `historyUpdateUrl` viewer preference.

[2] This depends, to a great extent, on browsers always firing `popstate` events *before* `hashchange` events, which may or may not actually be guaranteed.

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -112,6 +112,24 @@ const hasDOM = typeof window === 'object' && typeof document === 'object';
   };
 })();
 
+// Provides support for String.prototype.startsWith in legacy browsers.
+// Support: IE, Chrome<41
+(function checkStringStartsWith() {
+  if (String.prototype.startsWith) {
+    return;
+  }
+  require('core-js/fn/string/starts-with');
+})();
+
+// Provides support for String.prototype.endsWith in legacy browsers.
+// Support: IE, Chrome<41
+(function checkStringEndsWith() {
+  if (String.prototype.endsWith) {
+    return;
+  }
+  require('core-js/fn/string/ends-with');
+})();
+
 // Provides support for String.prototype.includes in legacy browsers.
 // Support: IE, Chrome<41
 (function checkStringIncludes() {
@@ -222,6 +240,24 @@ const hasDOM = typeof window === 'object' && typeof document === 'object';
 })();
 
 } // End of !PDFJSDev.test('CHROME')
+
+// Provides support for String.prototype.padStart in legacy browsers.
+// Support: IE, Chrome<57
+(function checkStringPadStart() {
+  if (String.prototype.padStart) {
+    return;
+  }
+  require('core-js/fn/string/pad-start');
+})();
+
+// Provides support for String.prototype.padEnd in legacy browsers.
+// Support: IE, Chrome<57
+(function checkStringPadEnd() {
+  if (String.prototype.padEnd) {
+    return;
+  }
+  require('core-js/fn/string/pad-end');
+})();
 
 // Provides support for Object.values in legacy browsers.
 // Support: IE, Chrome<54
--- a/web/pdf_history.js
+++ b/web/pdf_history.js
@@ -31,6 +31,14 @@ const UPDATE_VIEWAREA_TIMEOUT = 1000; // milliseconds
  */
 
 /**
+ * @typedef {Object} InitializeParameters
+ * @property {string} fingerprint - The PDF document's unique fingerprint.
+ * @property {boolean} resetHistory - (optional) Reset the browsing history.
+ * @property {boolean} updateUrl - (optional) Attempt to update the document
+ *   URL, with the current hash, when pushing/replacing browser history entries.
+ */
+
+/**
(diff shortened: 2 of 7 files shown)


web/pdf_history.js
```diff
@@ -30,6 +30,14 @@ const UPDATE_VIEWAREA_TIMEOUT = 1000; // milliseconds
  * @property {EventBus} eventBus - The application event bus.
  */
 
+/**
+ * @typedef {Object} InitializeParameters
+ * @property {string} fingerprint - The PDF document's unique fingerprint.
+ * @property {boolean} resetHistory - (optional) Reset the browsing history.
+ * @property {boolen} updateUrl - (optional) Attempt to update the document URL,
```
timvandermeij: Nit: boolean

src/shared/compatibility.js
```diff
@@ -223,6 +241,24 @@ const hasDOM = typeof window === 'object' && typeof document === 'object';
...
+// Provides support for String.prototype.padStart in legacy browsers.
+// Support: IE, Chrome<57
+(function checkStringPadStart() {
+  if (String.prototype.padStart) {
+    return;
+  }
+  require('core-js/fn/string/pad-start');
+})();
+
+// Provides support for String.prototype.padStart in legacy browsers.
```
timvandermeij: Nit: padEnd. Also, we don't use padding in this patch, so why add these polyfills? I understand that since we use `startsWith`, `endsWith` is useful to have as well, but we don't use any of the padding functions as far as I can tell.

web/pdf_history.js
Snuffleupagus: D'oh, good catch!

src/shared/compatibility.js
Snuffleupagus: Too much copy-paste, thanks!

> but we don't use any of the padding functions as far as I can tell.

Actually we do, please see https://github.com/mozilla/pdf.js/blob/af31b980b0db7b110965cdb468ee608769835d1e/src/display/dom_utils.js#L403

timvandermeij (approved):

timvandermeij:
I'm fine with this since it's off by default. Thank you!

tycho01:
That looks great, thank you @Snuffleupagus! :tada:

Rob--W:
This feature has no "title" key [in `preferences_schema.json`](https://github.com/mozilla/pdf.js/blob/d0892e46e2b0d16ecd7f720f36f36202ae1c12cd/extensions/chromium/preferences_schema.json#L158-L161). This prevents the option from being rendered in the settings page of the Chrome extension.

Preferences should only be hidden if they are internal, non-user facing or instable work-in-progress features. It seems that this feature does not match that description.

@Snuffleupagus If this feature is ready for use by users, could you open a PR and add a meaningful title to the `preferences_schema.json`, to expose the pref to users?

Snuffleupagus:
> This feature has no "title" key [in `preferences_schema.json`](https://github.com/mozilla/pdf.js/blob/d0892e46e2b0d16ecd7f720f36f36202ae1c12cd/extensions/chromium/preferences_schema.json#L158-L161). This prevents the option from being rendered in the settings page of the Chrome extension.

That was completely intentional on my part. (It should still be possible to set the preference *manually* from the console, as far as I can tell.)

> Preferences should only be hidden if they are internal, non-user facing or instable work-in-progress features. It seems that this feature does not match that description.

https://github.com/mozilla/pdf.js/pull/10423#issue-242499995 outlines some of the reasons why enabling feature this is a bad idea in my opinion, and thus I really don't think it's a good idea to give it a prominent position in any options UI.

> If this feature is ready for use by users, could you open a PR and add a meaningful title to the `preferences_schema.json`, to expose the pref to users?

Honestly, I'd classify this as only "sort of" supported and thus use *only* at you're own risk.
Given everything said above, personally I have no intention to put additional work into this feature (and I only submitted this patch to prevent a solution causing outright breakage from landing).

Rob--W:
Thanks for the quick reply. The reasoning sounds good to me.

Issue #6847, opened 2016-01-08
Remember view position after refreshing the page


Rob--W:
Currently the view position is saved by a hash based on the file contents. When the page reloads, we should also take the last position into account, because it matches the normal behavior of browsers. Usually, when you reload a web page, the scroll offset is restored (even if the content of the page changed).

The motivation for this change comes from experiencing the following broken workflow:
1. Generate a local PDF file (`file://....pdf`).
2. Open PDF with PDF.js and scroll to some chapter in the PDF file.
3. Edit PDF file.
4. Refresh the PDF.js viewer (e.g. with F5).
5. Expected result: Retain the scroll position.  
   Actual result: Page 1 is shown in the viewport.

Technical notes:
- `performance.navigation.type` can be used to detect page reload versus navigations.
- `history.state` is preserved when a page reloads.

yashamon:
That would be awesome.

Fatehsandhu:
I'm a student at Seneca college learning open source, and I was hoping to work on this bug for my course.  If no one else is currently working on it, I'd like to give it a try.

timvandermeij:
Nobody indicated that they're working on it, so it's all yours! Feel free to contact us on IRC or leave a message here in case you have questions.

Fatehsandhu:
Hey thanks a lot for the quick reply. I really want to contribute to an open source project. I will start working on it right away. Since this my first time doing something like this, is there anything that I should know about?
Thanks a a lot!

timvandermeij:
I think all necessary information for this patch is listed on https://github.com/mozilla/pdf.js/wiki/Contributing. Unless you're touching files in the `src/` folder (which I do not expect; I expect you'll only need to touch files in the `web/` folder), you only need to run `gulp lint` and `gulp unittest` to verify your changes. You can run `gulp server` to start a local server to test your changes in the browser. If you have more questions, check the wiki, contact us on IRC or ask here.

Fatehsandhu:
Thanks, I will start reading through the files.

andreip:
I'm looking into this but I don't know if I understood the problem very well.

> 1 - Generate a local PDF file (file://....pdf).
> 3 - Edit PDF file.

So the issue is only related to building/generating my own PDF? E.g. building it with some pdf generator like latex/jspdf?

I did the following and couldn't reproduce:
1. Built myself a PDF and opened it with `http://localhost:8888/web/viewer.html?file=/andrei_test/a4.pdf`
[...]

PR #10424, opened 2019-01-06
Accept non-matching document fingerprints, in `PDFHistory`, when the viewer is reloaded (issue 6847)

Snuffleupagus:
This should hopefully be sufficient to address issue #6847, and given the limited impact of the code changes I'm not completely sure if this would need to be controlled by a preference!?

Initially my intention was to try and provide some (slightly more detailed) implementation suggestions in the issue, but having looked briefly at doing that it would essentially have amounted to actually writing the code anyway. (Especially considering that the recent questions seemed to more-or-less ignore the information already provided in the first post.)

Finally, note that since `performance.navigation.type` is marked as deprecated, a slightly different approach was choosen instead.

### DIFF
--- a/web/pdf_history.js
+++ b/web/pdf_history.js
@@ -112,7 +112,7 @@ class PDFHistory {
     this._destination = null;
     this._position = null;
 
-    if (!this._isValidState(state) || resetHistory) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me. I don't think there is a need for a preference here.

PR #10425, opened 2019-01-06
Write more unit tests for the find controller

### LINKED ISSUE #7356: Add pdf_find_controller testing

timvandermeij:
Fixes #7356.

### DIFF
--- a/test/unit/pdf_find_controller_spec.js
+++ b/test/unit/pdf_find_controller_spec.js
@@ -71,28 +71,145 @@ describe('pdf_find_controller', function() {
     pdfFindController = null;
   });
 
-  it('performs a basic search', function(done) {
-    pdfFindController.executeCommand('find', { query: 'Dynamic', });
+  function testSearch({ parameters, matchesPerPage, selectedMatch, }) {
+    return new Promise(function(resolve) {
+      pdfFindController.executeCommand('find', parameters);
 
-    const matchesPerPage = [11, 5, 0, 3, 0, 0, 0, 1, 1, 1, 0, 3, 4, 4];
-    const totalPages = matchesPerPage.length;
-    const totalMatches = matchesPerPage.reduce((a, b) => {
-      return a + b;
-    });
-
-    eventBus.on('updatefindmatchescount',
-        function onUpdateFindMatchesCount(evt) {
-      if (pdfFindController.pageMatches.length !== totalPages) {
-        return;
+      // The `updatefindmatchescount` event is only emitted if the page contains
+      // at least one match for the query, so the last non-zero item in the
+      // matches per page array corresponds to the page for which the final
+      // `updatefindmatchescount` event is emitted. If this happens, we know
+      // that any subsequent pages won't trigger the event anymore and we
+      // can start comparing the matches per page. This logic is necessary
+      // because we call the `pdfFindController.pageMatches` getter directly
+      // after receiving the event and the underlying `_pageMatches` array
+      // is only extended when a page is processed, so it will only contain
+      // entries for the pages processed until the time when the final event
+      // was emitted.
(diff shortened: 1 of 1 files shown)


test/unit/pdf_find_controller_spec.js
```diff
@@ -71,11 +71,23 @@ describe('pdf_find_controller', function() {
     pdfFindController = null;
   });
 
-  it('performs a basic search', function(done) {
-    pdfFindController.executeCommand('find', { query: 'Dynamic', });
+  function testSearch({ parameters, matchesPerPage, callback, }) {
```
Snuffleupagus: How about having this function return a `Promise` instead, and just remove the callback?

```javascript
function testSearch({ parameters, matchesPerPage, }) {
  return new Promise(function(resolve) {
    pdfFindController.executeCommand('find', parameters);

    // More code here...

    eventBus.on('updatefindmatchescount',
        function onUpdateFindMatchesCount(evt) {

      // Even more code here...

      resolve(); // Instead of callback();
  });
}
```

Then a callsite would just look like e.g.
```javascript
testSearch({
  parameters: {
    query: 'Dynamic',
    caseSensitive: false,
    entireWord: false,
    phraseSearch: true,
  },
  matchesPerPage: [11, 5, 0, 3, 0, 0, 0, 1, 1, 1, 0, 3, 4, 4],
}).then(done);
```

test/unit/pdf_find_controller_spec.js
```diff
@@ -92,7 +104,63 @@ describe('pdf_find_controller', function() {
         expect(pdfFindController.pageMatches[i].length)
           .toEqual(matchesPerPage[i]);
       }
-      done();
+      callback();
+    });
+  }
+
+  it('performs a normal search', function(done) {
```
Snuffleupagus: Could you please also add *another* test, basically identical to this one, but with `findPrevious: true` set such that that code-path is covered by tests as well.

Snuffleupagus (approved):
Looks really good, modulo a couple of small comments/questions.
Thank you for doing this :-)

test/unit/pdf_find_controller_spec.js
```diff
@@ -71,11 +71,23 @@ describe('pdf_find_controller', function() {
...
-    pdfFindController.executeCommand('find', { query: 'Dynamic', });
+  function testSearch({ parameters, matchesPerPage, callback, }) {
+    pdfFindController.executeCommand('find', parameters);
+
+    // The `updatefindmatchescount` event is only emitted if the page contains
+    // at least one result for the query, so the last non-zero item in the
+    // matches per page array corresponds to the page for which the final
+    // `updatefindmatchescount` event is emitted. If this happens, we know
+    // that any subsequent pages won't trigger the event anymore and we
+    // can start comparing the matches per page.
```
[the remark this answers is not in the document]
timvandermeij: Yes, indeed. I have extended the comment to explain this, but in short the `pdfFindController.pageMatches` getter built lazily, i.e., it only gets an entry for a page when it's actually processed. If the final event is emitted on page 9 out of 14, then the `_pageMatches` array will only contain 9 entries if we call the getter.

test/unit/pdf_find_controller_spec.js
timvandermeij: Done.

test/unit/pdf_find_controller_spec.js
timvandermeij: Done. I have also extended the other unit tests to actually verify that the selected match is correct.

Issue #10212, opened 2018-11-02
Adding new tab in the sidebar gives out errors


chitgoks:
Hi, i am adding a custom tab together with thumbnail, attachment and outlines. but the thing is, if i click on the 4th custom button that i added and show the contents , it gives out an error because it looks like either of the first 3 buttons need to have its contents visible.

```
ui_utils.js:313 Uncaught (in promise) TypeError: Cannot read property 'div' of undefined
    at backtrackBeforeAllVisibleElements (ui_utils.js:313)
    at getVisibleElements (ui_utils.js:457)
    at PDFThumbnailViewer._getVisibleThumbs (pdf_thumbnail_viewer.js:67)
    at PDFThumbnailViewer.forceRendering (pdf_thumbnail_viewer.js:251)
    at PDFRenderingQueue.renderHighestPriority (pdf_rendering_queue.js:76)
    at continueRendering (pdf_rendering_queue.js:165)
backtrackBeforeAllVisibleElements @ ui_utils.js:313
getVisibleElements @ ui_utils.js:457
_getVisibleThumbs @ pdf_thumbnail_viewer.js:67
```
I add  the hidden class to for example the thumbnail list if i wish to have the 4th custom icon content shown. But then that error message happens and it breaks

Any Ideas?

Snuffleupagus:
Quoting from https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md (emphasis mine):
> If you are developing a custom solution, first check the examples at https://github.com/mozilla/pdf.js#learning and search existing issues. If this does not help, **please prepare a short well-documented example that demonstrates the problem and make it accessible online** on your website, JS Bin, GitHub, etc. before opening a new issue or contacting us on the IRC channel -- **keep in mind that just code snippets won't help us troubleshoot the problem**.

timvandermeij:
I think you need to make a custom view at https://github.com/mozilla/pdf.js/blob/master/web/pdf_sidebar.js#L164, but as indicated it's impossible to tell without a runnable example. Closing until this is provided.

chitgoks:
I did a workaround via css and just changed the location of the first 3 views to negative value with no pointer events and a negative z-index to be sure.

PR #10432, opened 2019-01-08
Fix of out of bounds index occuring in backtrackBeforeAllVisibleElements

vlastimilmaca:
This was recurring issue known as "cannot read div of undefined". ( #10212, #9895)

Problem was not located in backtrackBeforeAllVisibleElements function, where others tried to fix it, but it was caused by insufficient check of result from binarySearchFirstItem function in getVisibleElements function. 

As BinarySearchFirstItem function says in comment - @returns {Number} Index of the first array element to pass the test, **or |items.length|** if no such element exists. 

But backtrackBeforeAllVisibleElements requires index to be index into the given array, so this must be checked before calling backtrack function.

If binarySearch function actually returns views.length, after this fix, getVisibleElements should return {first: undefined, second:undefined, views: []}, which I think is ok, because there are actually no visible pages.

### DIFF
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -447,16 +447,17 @@ function getVisibleElements(scrollEl, views, sortByVisibility = false,
     binarySearchFirstItem(views, horizontal ? isElementRightAfterViewLeft :
                                               isElementBottomAfterViewTop);
 
-  if (views.length > 0 && firstVisibleElementInd !== views.length && !horizontal) {
-    // If there are no pages or no visible page was found (binarySearch returned
-    // length of input array), this cannot be run - index would be out of bounds.
+  if (views.length > 0 && firstVisibleElementInd !== views.length &&
+    !horizontal) {
+    // If there are no pages or no visible page was found (binarySearch
+    // returned length of input array), this cannot be run,
+    // index would be out of bounds.
 
     // In wrapped scrolling (or vertical scrolling with spreads), with some page
     // sizes, isElementBottomAfterViewTop doesn't satisfy the binary search
     // condition: there can be pages with bottoms above the view top between
     // pages with bottoms below. This function detects and corrects that error;
     // see it for more comments.
-
 
     firstVisibleElementInd =
       backtrackBeforeAllVisibleElements(firstVisibleElementInd, views, top);
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -447,12 +447,17 @@ function getVisibleElements(scrollEl, views, sortByVisibility = false,
     binarySearchFirstItem(views, horizontal ? isElementRightAfterViewLeft :
                                               isElementBottomAfterViewTop);
 
-  if (views.length > 0 && !horizontal) {
+  if (views.length > 0 && firstVisibleElementInd !== views.length && !horizontal) {
+    // If there are no pages or no visible page was found (binarySearch returned
+    // length of input array), this cannot be run - index would be out of bounds.
+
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
> This was recurring issue known as "cannot read div of undefined". ( #10212, #9895)

Given that no one has yet to provide *any* evidence that this is a problem in the default viewer, but only seems to be an issue in custom implementations, and given that this code is *very* important[1] it's really not clear if you'd want to make any changes here.
Please keep in mind that previous attempts at patching this code have been rejected for those reasons!

---
[1] Any implementation errors here can completely break rendering in the PDF Viewer in Firefox.

Snuffleupagus:
> If it is very important code, it should be as defensive as possible in my opinion.

Maybe, but any changes to this code needs to be scrutinized *carefully* since no one wants to find out later that it broke rendering in Firefox (causing extra work to fix, and create uplift patches for prior Firefox versions).

And as already stated in https://github.com/mozilla/pdf.js/pull/10432#issuecomment-452252977, the effect of this patch will be to move `Error`s from one part of the code-base to another location, which really isn't helping in any way.

Snuffleupagus:
> If it is very important code, it should be as defensive as possible in my opinion.

Maybe, but any changes to this code needs to be scrutinized *carefully* since no one wants to find out later that it broke rendering in Firefox (causing extra work to fix, and requiring creation of uplift patches for prior Firefox versions).

And as already stated in https://github.com/mozilla/pdf.js/pull/10432#issuecomment-452252977, the effect of this patch will be to move `Error`s from one part of the code-base to another location, which really isn't helping in any way.

timvandermeij:
Like for the other pull request, here I'm also not convinced that this actually fixes the problem. To me it also looks like the problem will resurface in other place. Moreover, I have not seen any breakage like this in the default viewer, so I fully agree with https://github.com/mozilla/pdf.js/pull/10432#issuecomment-452252977. Closing until there is proof that the default viewer is also affected by this, otherwise it's not trivial to adjust this logic.

vlastimilmaca:
Ok. If you try to set display: none to any parent of #viewer element in default viewer and then you resize window or scroll, this error occurs. 

It is often case in custom applications, because pdf viewer is just small part of that applications and usually is needed to be hidden sometimes (ie. pdf view is located in accordion tabs or something like that ... or you need to hide it fast and handle destroy later). Display: none simulates that. 

Previous fixes actually introduced new value returned from backtrackBeforeAllVisibleElements, so I understanded your concerns, that it might produce errors somewhere else in code-base, because it would be a new state, which wasn't there before.

This fix however does not do that. 
Problem is, it is not clearly stated, that getVisibleElements can (and already does) return value, that there are "no visible elements" at all.  And it is not clear, what that return value is (visibles is an empty array, first,last are undefined) ... but it is there and it occurs, when you send in an empty array of views. But if you send filled array and function inside correctly detects, that there are no visible elements (returns items.lenght) it should return same result for "no visible elements". Which is what this fix does (there is no point of looking for other visible pages with backtrack function, if we already know, there are none). 
There are no changes in possible return values of getVisibleElements function, so it reeeally doesn't move any errors somewhere else. It fixes problem inside getVisibleElements.

Those, who are calling getVisibleElements, should already be aware, that "no visible elements" might occur (and in pdf.js repo they are aware of that ... as far as I checked, but I will check it again).

I understand your concerns and it is possible, that it might not occur naturally in default viewer, but it essentialy prevents usage anywhere else.

Issue #10395, opened 2018-12-29
PDFjs.getMetadata() - XML parse bug


dhollenbeck:
When getMetadata() is invoked on a specific PDF file, which I can not provide:
```js
var loader = pdfjs.getDocument(source);
loader.promise.then(function getDocumentSuccess(pdf) {
  pdf.getMetadata().then(function () {}).catch(...);
});
```
I get the following error:
```bash
(node:7452) TypeError: Cannot read property '0' of undefined
    at SimpleDOMNode.get (...\node_modules\pdfjs-dist\build\pdf.js:14097:29)
    at Metadata._parse (...\node_modules\pdfjs-dist\build\pdf.js:13740:19)
    at new Metadata (...\node_modules\pdfjs-dist\build\pdf.js:13698:12)
    at ...\pdfjs-dist\build\pdf.js:9421:34
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```
I believe the following line of code is the line with the bug:
https://github.com/mozilla/pdf.js/blob/2ce489c16f0f642df1874c0da46aff5a4221de0a/src/display/xml_parser.js#L280-L282

Obviously, `this.childNodes` could be undefined as is the case with my pdf file. It is being invoked via an automatic property **get** function above:
https://github.com/mozilla/pdf.js/blob/66ffdc4c5b63a135f201508264761fce17c08059/src/display/metadata.js#L77

The offending XML document has the structure of:
```js
{ documentElement: SimpleDOMNode { nodeName: '#text', nodeValue: '﻿' } }
```

timvandermeij:
@dhollenbeck Would it be possible to provide a test case without sensitive data here as well like you did for the other issue? Either a file or the XML itself is fine. This would allow us to add a unit test for this so we can avoid regressions in the future, and allow us to verify the fix.

dhollenbeck:
Absolutely, @timvandermeij 
[toissue.pdf](https://github.com/mozilla/pdf.js/files/2727418/toissue.pdf)

PR #10457, opened 2019-01-16
When parsing Metadata, attempt to remove "junk" before the first tag (PR 10398 follow-up)

Snuffleupagus:
This will allow the Metadata to be successfully extracted from the PDF file in issue #10395.
Furthermore, this patch also fixes a bug in `Metadata.get` which causes the method to return `null` rather than an empty string or zero (since either ought to be allowed).

### DIFF
--- a/src/display/metadata.js
+++ b/src/display/metadata.js
@@ -35,7 +35,9 @@ class Metadata {
   }
 
   _repair(data) {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Nice work!

PR #10462, opened 2019-01-16
Tweak the `DOMTokenList.toggle` polyfill (issue 10460)

### LINKED ISSUE #10460: Element.classList.toggle does not work in IE

Configuration:
- Web browser and its version: IE (any version)
- Operating system and its version: Windows 8/10
- PDF.js version: 2.0.943
- Is a browser extension: No

1. Execute document.getElementsByTagName('body')[0].classList.toggle('test') on the Internet Explorer console

Should add or remove a class to an element.

Throws an "invalid calling object" error

image

Snuffleupagus:
Fixes #10460.

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -101,14 +101,17 @@ const hasDOM = typeof window === 'object' && typeof document === 'object';
   if (div.classList.toggle('test', 0) === false) {
     return;
   }
-  const originalDOMTokenListToggle = DOMTokenList.prototype.toggle;
 
   DOMTokenList.prototype.toggle = function(token) {
     if (arguments.length > 1) {
       const force = !!arguments[1];
       return (this[force ? 'add' : 'remove'](token), force);
     }
-    return originalDOMTokenListToggle(token);
+
+    if (this.contains(token)) {
+      return (this.remove(token), false);
+    }
+    return (this.add(token), true);
   };
 })();
 
--- a/web/pdf_find_bar.js
+++ b/web/pdf_find_bar.js
@@ -180,7 +180,7 @@ class PDFFindBar {
     }
     Promise.resolve(matchesCountMsg).then((msg) => {
       this.findResultsCount.textContent = msg;
-      this.findResultsCount.classList[!total ? 'add' : 'remove']('hidden');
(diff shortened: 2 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

src/shared/compatibility.js
```diff
@@ -101,14 +101,17 @@ const hasDOM = typeof window === 'object' && typeof document === 'object';
...
     if (arguments.length > 1) {
       const force = !!arguments[1];
       return (this[force ? 'add' : 'remove'](token), force);
     }
-    return originalDOMTokenListToggle(token);
+
+    if (this.contains(token)) {
+      return (this.remove(token), false);
+    }
+    return (this.add(token), true);
```
mhuggins: Arguable, but would it have been work calling the original function?  The main benefit would be that if the underlying implementation to `toggle` changes, then this never needs to change.  i.e.:

```js
return originalDOMTokenListToggle.call(this, token);
```

mhuggins: Regardless, thanks for this fix!

timvandermeij: I think that's essentially what we had before this fix and it didn't work in IE. Since this is only a polyfill for IE and toggle functionality is trivial, I don't think we need to change this.

mhuggins: The original call was this:

```js
return originalDOMTokenListToggle(token);
```

My suggestion, which also fixes the problem in IE, is this:

```js
return originalDOMTokenListToggle.call(this, token);
```

It's ultimately not a big deal since the issue is now fixed.

timvandermeij: Ah, I see. The diff in this patch is actually old; in the current code it's minimized to https://github.com/mozilla/pdf.js/blob/master/src/shared/compatibility.js#L110-L113. In the current code I don't think calling the original function is really possible anymore since that one doesn't support the `force` parameter, so then we would need to special-case it and get back to the code in this diff. Indeed it doesn't really matter too much since it's a working polyfill and it will probably be removed once we drop IE support in a later major release anyway.

mhuggins: Seems like a moot point then, thanks for calling that out!

PR #10466, opened 2019-01-18
Move/refactor the code in the `BaseViewer.update` method to reduce duplication in the extending classes

Snuffleupagus:
Since most of the important rendering code is already (almost) identical between `PDFViewer.update` and `PDFSinglePageViewer.update`, it's possible to further reduce duplication by moving the code into `BaseViewer.update` instead.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -824,8 +824,28 @@ class BaseViewer {
     };
   }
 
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Good find!

Issue #10409, opened 2019-01-03
Improve unit testing for metadata parsing


timvandermeij:
Some cases were found that we did not cover, so go over these issues and complete test coverage where necessary:

- #10412
- #10407
- #10395
- #8903

timvandermeij:
Done!

PR #10471, opened 2019-01-19
Implement a unit test for metadata parsing to ensure that it's not vulnerable to the billion laughs attack

timvandermeij:


Snuffleupagus (approved):
Looks good to me, and nice with the additional clean-up!

This should also fix #10409, since it looks like all the other issues are already fixed.

PR #10473, opened 2019-01-20
Add an `Array.from` polyfill, using core-js, and remove some compatibility hacks from the `src/display/content_disposition.js` file

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
More standardization is always good. Thanks!

PR #10475, opened 2019-01-21
SVG Improvements Part I : group strategy

fr33z00:
Hi all.
As Discussed with Tim, here is the first PR of a series of 3 ment to improve SVG rendering.
It changes the group creation strategy, by creating groups according to the PDF requests.
This one is a bit massive, but there was no splitting possibility. It is mandatory for the next PR that will introduce patterns & gradients.

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -315,7 +315,10 @@ var SVGExtraState = (function SVGExtraStateClosure() {
 
     // Clipping
     this.activeClipUrl = null;
-    this.clipGroup = null;
+
+    //hierarchy
+    this.group = null;
+    this.parentGroup = null;
 
     this.maskId = '';
   }
@@ -409,7 +412,6 @@ SVGGraphics = (function SVGGraphicsClosure() {
 
     this.current = new SVGExtraState();
     this.transformMatrix = IDENTITY_MATRIX; // Graphics state matrix
-    this.transformStack = [];
     this.extraStack = [];
     this.commonObjs = commonObjs;
     this.objs = objs;
@@ -419,6 +421,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
     this.embedFonts = false;
     this.embeddedFonts = Object.create(null);
     this.cssStyle = null;
+    this.group = null;
     this.forceDataSchema = !!forceDataSchema;
   }
 
@@ -431,23 +434,34 @@ SVGGraphics = (function SVGGraphicsClosure() {
 
   SVGGraphics.prototype = {
     save: function SVGGraphics_save() {
-      this.transformStack.push(this.transformMatrix);
       var old = this.current;
+      old.parentGroup = this.parentGroup;
+      old.group = this.group;
       this.extraStack.push(old);
       this.current = old.clone();
+      this.current.activeClipUrl = null;
     },
 
     restore: function SVGGraphics_restore() {
-      this.transformMatrix = this.transformStack.pop();
       this.current = this.extraStack.pop();
 
       this.pendingClip = null;
-      this.tgrp = null;
+      this.parentGroup = this.current.parentGroup;
+      this.group = this.current.group;
     },
 
-    group: function SVGGraphics_group(items) {
+    createGroup: function SVGGraphics_group(items) {
       this.save();
+      this.parentGroup = this.group;
+      this.group = this.svgFactory.createElement('svg:g');
+      if (this.current.activeClipUrl)
+        this.group.setAttributeNS(null, 'clip-path', 
+                                  this.current.activeClipUrl);
+      this.parentGroup.appendChild(this.group);
       this.executeOpTree(items);
+      if (!this.group.children.length) {
+        this.parentGroup.removeChild(this.group);
+      }
       this.restore();
     },
 
@@ -480,10 +494,11 @@ SVGGraphics = (function SVGGraphicsClosure() {
     },
 
     transform: function SVGGraphics_transform(a, b, c, d, e, f) {
-      var transformMatrix = [a, b, c, d, e, f];
-      this.transformMatrix = Util.transform(this.transformMatrix,
-                                            transformMatrix);
-      this.tgrp = null;
+      this.transformMatrix = [a, b, c, d, e, f];
+      this.parentGroup = this.group;
+      this.group = this.svgFactory.createElement('svg:g');
+      this.group.setAttributeNS(null, 'transform', pm(this.transformMatrix));
+      this.parentGroup.appendChild(this.group);
     },
 
     getSVG: function SVGGraphics_getSVG(operatorList, viewport) {
@@ -665,8 +680,8 @@ SVGGraphics = (function SVGGraphicsClosure() {
           case OPS.endPath:
             this.endPath();
             break;
-          case 92:
-            this.group(opTree[x].items);
+          case OPS.group:
+            this.createGroup(opTree[x].items);
             break;
           default:
             warn('Unimplemented operator ' + fn);
@@ -713,6 +728,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
       this.current.tspan = this.svgFactory.createElement('svg:tspan');
       this.current.txtElement = this.svgFactory.createElement('svg:text');
       this.current.txtgrp = this.svgFactory.createElement('svg:g');
+      this.current.element = this.current.txtgrp;
       this.current.xcoords = [];
     },
 
@@ -832,8 +848,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
       current.txtElement.setAttributeNS(XML_NS, 'xml:space', 'preserve');
       current.txtElement.appendChild(current.tspan);
       current.txtgrp.appendChild(current.txtElement);
-
-      this._ensureTransformGroup().appendChild(current.txtElement);
+      this._clipnAppend(current.txtgrp);
     },
 
     setLeadingMoveText: function SVGGraphics_setLeadingMoveText(x, y) {
@@ -993,7 +1008,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
       current.path.setAttributeNS(null, 'd', d.join(' '));
       current.path.setAttributeNS(null, 'fill', 'none');
 
-      this._ensureTransformGroup().appendChild(current.path);
+      this._clipnAppend(this.current.path);
 
       // Saving a reference in current.element so that it can be addressed
       // in 'fill' and 'stroke'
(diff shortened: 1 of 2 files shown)


src/display/svg.js
```diff
@@ -1316,39 +1315,18 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
-        this.tgrp.setAttributeNS(null, 'transform', pm(this.transformMatrix));
-        if (this.current.activeClipUrl) {
-          this._ensureClipGroup().appendChild(this.tgrp);
-        } else {
-          this.svg.appendChild(this.tgrp);
-        }
-      }
-      return this.tgrp;
+    _clipnAppend: function SVGGraphics_clipnAppend(element) {
+      if (this.current.activeClipUrl)
```
timvandermeij: Always add curly brackets around `if` statements:

```javascript
if (this.current.activeClipUrl) {
  element.setAttributeNS(null, 'clip-path', this.current.activeClipUrl);
}
this.group.appendChild(element);
```

src/display/svg.js
```diff
@@ -1316,39 +1315,18 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
-        this.tgrp = this.svgFactory.createElement('svg:g');
-        this.tgrp.setAttributeNS(null, 'transform', pm(this.transformMatrix));
-        if (this.current.activeClipUrl) {
-          this._ensureClipGroup().appendChild(this.tgrp);
-        } else {
-          this.svg.appendChild(this.tgrp);
-        }
-      }
-      return this.tgrp;
+    _clipnAppend: function SVGGraphics_clipnAppend(element) {
```
timvandermeij: This name is a bit unclear. Perhaps rename this to `_appendToGroup`? The clipping seems more like a detail here.

src/display/svg.js
```diff
@@ -431,23 +434,34 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
       this.current = this.extraStack.pop();
 
       this.pendingClip = null;
-      this.tgrp = null;
+      this.parentGroup = this.current.parentGroup;
+      this.group = this.current.group;
     },
 
-    group: function SVGGraphics_group(items) {
+    createGroup: function SVGGraphics_group(items) {
```
timvandermeij: `function SVGGraphics_group` should be `function SVGGraphics_createGroup`

src/display/svg.js
```diff
@@ -431,23 +434,34 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
       this.save();
+      this.parentGroup = this.group;
+      this.group = this.svgFactory.createElement('svg:g');
+      if (this.current.activeClipUrl)
```
timvandermeij: Also here the curly brackets need to be added.

src/display/svg.js
```diff
@@ -315,7 +315,10 @@ var SVGExtraState = (function SVGExtraStateClosure() {
 
     // Clipping
     this.activeClipUrl = null;
-    this.clipGroup = null;
+
+    //hierarchy
```
timvandermeij: Perhaps `// Group hierarchy` is a bit more descriptive.

timvandermeij (commented):
First round of comments. It looks a bit simpler than before, so that's nice, but we'll have to do more testing to ensure nothing breaks. It would in the meantime help if you could explain the changes a bit more in text, i.e., by highlighting what is different/better compared to the previous situation. I don't really understand what "by creating groups according to the PDF requests" means exactly. Finally, please check the Travis CI build at https://travis-ci.org/mozilla/pdf.js/builds/482347027?utm_source=github_status&utm_medium=notification to see what else needs to be adjusted for linting to pass. Thank you!

fr33z00:
Tim,
I managed to match the coding rules and modified the comment and function name as requested.
I also fixed the problem with the 2 other pull requests.
Hope this will be good for a merge, this time.

To answer your questions, the PDF 'program' requests groups creation by calling OPS.group or OPS.beginGroup. By respecting this, not only is it easier to understand the svg hierarchy with a side by side comparision with opTree, but clipping, transform, shadingFill, etc... apply to the correct set of elements, which was not always the case, based on my bundle of test docs.

amit777:
Is there a public build containing your SVG fixes?  I'm happy to test them out across some of my PDFs and report back any errors.

fr33z00:
@amit777
Yes,you may use the tool I made this work for here : www.qwi-tech.com/QDB
Simply drag & drop your PDF file on the window

src/display/svg.js
```diff
@@ -419,6 +421,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
     this.embedFonts = false;
     this.embeddedFonts = Object.create(null);
     this.cssStyle = null;
+    this.group = null;
```
timvandermeij: Shouldn't we also do `this.parentGroup = null` here?

Also, let's rename `this.group` to `this._group` and `this.parentGroup` to `this._parentGroup` to indicate that they are private. This also avoids name clashes with the `group` method.

src/display/svg.js
```diff
@@ -431,23 +434,35 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
       this.current = this.extraStack.pop();
 
       this.pendingClip = null;
-      this.tgrp = null;
+      this.parentGroup = this.current.parentGroup;
+      this.group = this.current.group;
     },
 
-    group: function SVGGraphics_group(items) {
+    createGroup: function SVGGraphics_group(items) {
```
timvandermeij: This can still be `group` if you rename the member variables as indicated in the comment above.

src/display/svg.js
```diff
@@ -431,23 +434,35 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
       this.save();
+      this.parentGroup = this.group;
+      this.group = this.svgFactory.createElement('svg:g');
+      if (this.current.activeClipUrl) {
+        this.group.setAttributeNS(null, 'clip-path',
+                                  this.current.activeClipUrl);
```
timvandermeij: This should still fit on the previous line.

src/display/svg.js
```diff
@@ -431,23 +434,35 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
+      }
+      this.parentGroup.appendChild(this.group);
       this.executeOpTree(items);
+      if (!this.group.children.length) {
```
timvandermeij: Put a comment above this: `// Remove the group if it is empty after executing the operator tree.`

src/display/svg.js
```diff
@@ -665,8 +681,8 @@ SVGGraphics = (function SVGGraphicsClosure() {
           case OPS.endPath:
             this.endPath();
             break;
-          case 92:
-            this.group(opTree[x].items);
+          case OPS.group:
+            this.createGroup(opTree[x].items);
```
timvandermeij: This can still be `group` if you rename the member variables as indicated in the comment above.

src/display/svg.js
```diff
@@ -713,6 +729,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
       this.current.tspan = this.svgFactory.createElement('svg:tspan');
       this.current.txtElement = this.svgFactory.createElement('svg:text');
       this.current.txtgrp = this.svgFactory.createElement('svg:g');
+      this.current.element = this.current.txtgrp;
```
timvandermeij: Why is this necessary? I found no change when I removed it.

src/display/svg.js
```diff
@@ -832,8 +849,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
       current.txtElement.setAttributeNS(XML_NS, 'xml:space', 'preserve');
       current.txtElement.appendChild(current.tspan);
       current.txtgrp.appendChild(current.txtElement);
-
-      this._ensureTransformGroup().appendChild(current.txtElement);
+      this._clipAndAppendToGroup(current.txtgrp);
```
timvandermeij: `txtgrp` should be `txtElement` I think, since that was in the old code, or is that wrong?

src/display/svg.js
```diff
@@ -1011,31 +1027,15 @@ SVGGraphics = (function SVGGraphicsClosure() {
       clipCount++;
       var clipPath = this.svgFactory.createElement('svg:clipPath');
       clipPath.setAttributeNS(null, 'id', clipId);
-      clipPath.setAttributeNS(null, 'transform', pm(this.transformMatrix));
-      // A deep clone is needed when text is used as a clipping path.
```
timvandermeij: Why is the deep cloning removed? Is that not applicable anymore?

src/display/svg.js
```diff
@@ -1011,31 +1027,15 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
-
-      if (current.activeClipUrl) {
-        // The previous clipping group content can go out of order -- resetting
-        // cached clipGroups.
-        current.clipGroup = null;
-        this.extraStack.forEach(function (prev) {
-          prev.clipGroup = null;
-        });
-        // Intersect with the previous clipping path.
-        clipPath.setAttributeNS(null, 'clip-path', current.activeClipUrl);
```
timvandermeij: Why are we not setting the `clip-path` attribute in this method anymore?

src/display/svg.js
```diff
@@ -1107,11 +1107,10 @@ SVGGraphics = (function SVGGraphicsClosure() {
 
     fill: function SVGGraphics_fill() {
       var current = this.current;
-      if (current.element) {
```
timvandermeij: This `if` statement should not be removed; if was added in https://github.com/mozilla/pdf.js/commit/4e8f7070da236c6a3a8572ff141d6a4ce231bb94 to fix a bug.

timvandermeij (changes requested):
First of all, sorry for not getting back to this PR sooner. My time was limited lately and there were some other priorities, but I finally managed to get the other SVG back-end patches merged so it's easier to work with.

I think in general this patch simplifies the logic quite a bit, thereby also reducing the amount of state we have to keep track of, so that's great! I applied your patch on top of the current master and, aside from the review comments, found that the default Tracemonkey has a problem on page 6 where there's a large white area in the top-right corner of the page. There are also quite a few empty transform groups in the DOM, which should be cleaned up. I didn't notice these issues in the link you posted above, so perhaps you forgot some changes in this PR?

In short, could you rebase this patch onto the current master and address these reviews comments? Thank you!

src/display/svg.js
```diff
@@ -665,8 +681,8 @@ SVGGraphics = (function SVGGraphicsClosure() {
           case OPS.endPath:
             this.endPath();
             break;
-          case 92:
-            this.group(opTree[x].items);
+          case OPS.group:
```
timvandermeij: The `92` in the `opListToTree` function should also be replaced.

src/display/svg.js
```diff
@@ -1107,11 +1107,10 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
       var current = this.current;
-      if (current.element) {
-        current.element.setAttributeNS(null, 'fill', current.fillColor);
-        current.element.setAttributeNS(null, 'fill-opacity', current.fillAlpha);
-        this.endPath();
-      }
+      current.element.setAttributeNS(null, 'fill', current.fillColor);
+      current.element.setAttributeNS(null, 'fill-opacity', current.fillAlpha);
+      this.endPath();
+      this._clipAndAppendToGroup(current.element);
```
timvandermeij: Why is this necessary? I found no change when I removed it.

src/display/svg.js
```diff
@@ -1123,6 +1122,7 @@ SVGGraphics = (function SVGGraphicsClosure() {
         current.element.setAttributeNS(null, 'fill', 'none');
 
         this.endPath();
+        this._clipAndAppendToGroup(current.element);
```
timvandermeij: Why is this necessary? I found no change when I removed it.

fr33z00:
Hello, Tim.
First, let me apologize about the several mistakes introduced when I rebased. I should have used a side by side compare tool when I resolved conflicts. I'm not experienced in PRing, and, as everybody, am always in a hurry.
 
Secondly, since many changes were made on the root, and even patterns and shadings introduced, I would say continuing on the PR series seems not to be the best strategy. I even wonder if it has any interest, since all this stuff was made around patterns & shadings... However, if you confirm your interest in this way to manage groups, I'm ok to do it on a new PR.
What is your position?

timvandermeij:
I thought about this a bit more and perhaps it's indeed better to leave this as-is for now since we don't really experience issues with the current code.

Even though another patch indeed introduced shading pattern support in the SVG back-end, we're still missing tiling pattern support which you already made. It would be great it we could extract that from your existing PR into its own self-contained PR so that it's easy to review and merge. It should hopefully be easy to move that code into the existing shading pattern code in a similar way: there is currently https://github.com/mozilla/pdf.js/blob/master/src/display/svg.js#L985 which could be replaced with a call like `this._makeTilingPattern(args);` that creates the tiling pattern. I think it should be a matter of introducing a new method that does that based on your existing work.

Finally, you also have an open pull request for blend modes, which we could integrate after the tiling patterns part. That's also still unsupported and it looks like you have also done practically all of the work there already, so I think that's also mainly about integration into the current master branch.

This way we could complete your work in two specific, self-contained PRs and close the current three PRs. If you need help with this, feel free to let us know. We're always happy to help out, especially since this is great work and I really feel that it should be merged into the codebase. If you don't have time to do that, we can just leave the PRs open and another contributor can take it over. Aside from those two feature PRs, you can always open issues for any problems that remain since I recall you mentioning that some of your PDF documents were not rendered correctly with the current code. If you do, please attach a sample PDF file so we can look into the issues one at a time. Thanks again.

timvandermeij:
Closing because of the comment above.

PR #10477, opened 2019-01-21
SVG Improvements Part III :  fixes & improvements

fr33z00:
This is the third and last PR of a series of 3 ment to improve SVG rendering.
This one brings little fixes, blend modes support and removes the unnecessary 'svg:' prefix to elements.


timvandermeij:
I think something went wrong here because I see the same commit and in the other PR here.

timvandermeij:
The SVG back-end has recently been improved with shading pattern and tiling pattern support. However, it's still missing blend modes support, which this pull request (among other changes) seems to implement. Would it perhaps be possible for you to update this pull request to only implement blend modes support with the current (master) SVG back-end code? That way the scope is much smaller, as well as the number of changes.

PR #10482, opened 2019-01-22
Implement Decode entry in Indexed images

### LINKED ISSUE #10339: PDF not rendered correctly in canvas
https://drive.google.com/drive/folders/1Gw6flKyN8BHWyDuW2l3ofHgMZ2abEanC

## Configuration:
- Web browser and its version: Chrome 70.0.3538.110
- Operating system and its version: Ubuntu 18.04
- PDF.js version: latest source code
- Is a browser extension: no

1. replace defaultUrl with pdf file list in google drive.
2. pdf not rendered correctly in canvas.

#### *issue1*(Background Cover Text)
Background hide text

#### *issue2*(Background Color Not Correct, left is pdfjs, right is pdf app)
Background hide text

#### *issue1*(Background Cover Text)
Background hide text

#### *issue2* (Background Color Not Correct, left is pdfjs, right is pdf app)
Background Color Not Correct

janpe2:
Fixes #10339 in which page background images have an Indexed color space and an inverting entry `/Decode [255.0 0.0]`. `Decode` hasn't been implemented for Indexed, so wrong colors appeared.

In core/image.js I had to modify also `decodeCoefficients` and `decodeAddends` values if the color space is Indexed. Function `decodeAndClamp()` uses those values and it returns color components multiplied by `max`. (`max` is 255 when BitsPerComponent is 8.) `decodeAndClamp()` works fine if the components are in range [0, 1]. However, Indexed components are in range [0, `max`]. If `decodeAndClamp()` multiplied them again by `max`, the values would go out of range.


src/core/colorspace.js
```diff
@@ -565,8 +565,15 @@ class IndexedCS extends ColorSpace {
...
-  isDefaultDecode(decodeMap) {
-    return true; // Indexed color maps shouldn't be changed.
+  isDefaultDecode(decodeMap, bpc) {
+    if (!Array.isArray(decodeMap)) {
+      return true;
+    }
+    if (decodeMap.length !== 2) {
+      warn('Decode map length is not correct');
+      return true;
+    }
```
Snuffleupagus: Perhaps it would also be a good idea to check that the `bpc` parameter is an integer, larger than zero, in *addition* to the checks done above?

src/core/image.js
```diff
@@ -169,18 +169,21 @@ var PDFImage = (function PDFImageClosure() {
...
       this.decodeCoefficients = [];
       this.decodeAddends = [];
+      const isIndexed = this.colorSpace && this.colorSpace.name === 'Indexed';
       for (var i = 0, j = 0; i < this.decode.length; i += 2, ++j) {
         var dmin = this.decode[i];
         var dmax = this.decode[i + 1];
-        this.decodeCoefficients[j] = dmax - dmin;
-        this.decodeAddends[j] = max * dmin;
+        this.decodeCoefficients[j] = isIndexed ? ((dmax - dmin) / max) :
+                                     (dmax - dmin);
```
Snuffleupagus: Nit: Maybe a tiny bit easier to read with slightly adjusted indentation, e.g.
```javascript
this.decodeCoefficients[j] = isIndexed ? ((dmax - dmin) / max) :
                                         (dmax - dmin);
```

Snuffleupagus (approved):

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

PR #10494, opened 2019-01-25
Reduce unnecessary duplication of the `isDefaultDecode` methods on `ColorSpace` instances

Snuffleupagus:
The recent PR 10482 made me realize that I missed an opportunity for simplification when doing the class conversion of this code in PR #10007.


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #10499, opened 2019-01-26
Attempt to clean-up/restore pending rendering operations when errors occurs while a `RenderTask` runs (PR 10202 follow-up)

Snuffleupagus:
This piggybacks of the existing `cancel` functionality, to ensure that any pending operations are closed *and* that any temporary canvases are actually being removed.

Also simplifies `finishPaintTask` in `PDFPageView.draw` slightly, by converting it to an async function.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2434,7 +2434,7 @@ const InternalRenderTask = (function InternalRenderTaskClosure() {
       }
     }
 
-    cancel() {
+    cancel(error = null) {
       this.running = false;
       this.cancelled = true;
       if (this.gfx) {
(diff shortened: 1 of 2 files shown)


src/display/api.js
```diff
@@ -2480,10 +2480,10 @@ const InternalRenderTask = (function InternalRenderTaskClosure() {
     _scheduleNext() {
       if (this._useRequestAnimationFrame) {
         window.requestAnimationFrame(() => {
-          this._nextBound().catch(this.callback);
+          this._nextBound().catch(this.cancel.bind(this));
         });
       } else {
-        Promise.resolve().then(this._nextBound).catch(this.callback);
+        Promise.resolve().then(this._nextBound).catch(this.cancel(this));
```
Snuffleupagus: D'oh, really stupid typo, fix incoming.

timvandermeij (approved):

timvandermeij:
Nice find!

PR #10501, opened 2019-01-27
In `getVisibleElements`, check `firstVisibleElementInd` rather than `numViews` before backtracking (PR 10443 follow-up)

Snuffleupagus:
When `firstVisibleElementInd === 0`, regardless of the number of views, there's no reason to attempt to backtrack at all since it's never possible to find an element before the *first* one anyway.


timvandermeij (approved):

timvandermeij:
Makes sense. Thanks!

PR #10504, opened 2019-01-28
Enable the `no-useless-catch` ESLint rule

Snuffleupagus:
See https://eslint.org/docs/rules/no-useless-catch


timvandermeij (approved):

timvandermeij:
Good to have yet another ESLint rule enabled. Thanks!

PR #10510, opened 2019-01-31
Bugfix for Angular CLI (Node)

sboerjan:
Experiencing problems with Angular Universal because document/window is not available in CLI, this fixes it.

### DIFF
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -665,7 +665,8 @@ function waitOnEventOrTimeout({ target, name, delay = 0, }) {
  */
 let animationStarted = new Promise(function (resolve) {
   if ((typeof PDFJSDev !== 'undefined' && PDFJSDev.test('LIB')) &&
-      typeof window === 'undefined') {
+      typeof window === 'undefined' &&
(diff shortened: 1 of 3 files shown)


web/ui_utils.js
```diff
@@ -144,7 +144,7 @@ function scrollIntoView(element, spot, skipOverflowHiddenElements = false) {
  */
 function watchScroll(viewAreaElement, callback) {
   let debounceScroll = function(evt) {
-    if (rAF) {
+    if (rAF && typeof window === 'undefined') {
```
Snuffleupagus: I really don't think it's acceptable to run this check *every* time that scrolling happens in the default viewer, just so that you can support a special environment (Angular) which is apparently providing incomplete/incorrect `window` polyfills.

timvandermeij:
I agree with the comments above. This is basically working around a limitation in the custom implementation/in Angular. Not only are the changes not really clear as to why they are even necessary, they also introduce more logic for well-behaving environments. In short, I don't think PDF.js should be changed for this particular environment, at least not in the current way, unless there is a very simple polyfill that can be added to e.g., `compatibility.js` without touching other code.

PR #10517, opened 2019-02-03
[api-minor] Remove the `document.currentScript` polyfill

Snuffleupagus:
This polyfill is currently used in only *one* file, i.e. `src/display/api.js`, and only when trying to build a *fallback* `workerSrc` path.

Given that the global `workerSrc` should *always* be set[1] when using the PDF.js library[2], and that the fallback `workerSrc` should only be regarded as a best-effort solution anyway, there isn't a particularily strong reason to keep the compatibility code in my opinion.

---
[1] Other supported options include setting the global `workerPort`, or passing in a `PDFWorker` instance as part of the `getDocument` call.

[2] Which is clearly mentioned in the JSDocs in `src/display/worker_options.js`.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -90,7 +90,8 @@ if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) {
     });
   }) : null;
 
-  if (!fallbackWorkerSrc && typeof document !== 'undefined') {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
More standardization is always welcome. Thank you!

Issue #10232, opened 2018-11-07
Scrambled fonts with World Chlorine Council PDF


uswoods:
https://worldchlorine.org/wp-content/themes/brickthemewp/pdfs/WCC_Policy_Paper_Water_Chlorination.pdf

Configuration:
- Web browser and its version: Firefox 63.0.1
- Operating system and its version: Linux 4.19.0
- PDF.js version: included in Firefox

Steps to reproduce the problem:
1. Open page

Result:
![screenshot_2018-11-07 wcc_policy_paper_water_chlorination pdf](https://user-images.githubusercontent.com/42948244/48131786-9142a780-e291-11e8-93d3-3ce665079bbd.png)

timvandermeij:
I can reproduce this on Arch Linux with Firefox 63.0.1 (64-bit). I think the file is somewhat corrupt given the console logs:

```
Warning: Indexing all PDF objects pdf.worker.js:961:5
PDF 6c655ef4c750453994376a7cffe75877 [1.4 Adobe PDF Library 8.0 / Adobe InDesign CS3 (5.0.1)] (PDF.js: 2.0.775) viewer.js:1044:7
Warning: decodeScan - unexpected MCU data, current marker is: 4ab pdf.worker.js:961:5
Warning: Not enough parameters for vhcurveto; actual: 0, expected: 4 pdf.worker.js:961:5
Warning: Not enough parameters for hmoveto; actual: 0, expected: 1 pdf.worker.js:961:5
Warning: Not enough parameters for hvcurveto; actual: 0, expected: 4 pdf.worker.js:961:5
Warning: Not enough parameters for rcurveline; actual: 1, expected: 8 pdf.worker.js:961:5
Warning: Not enough parameters for hvcurveto; actual: 3, expected: 4 pdf.worker.js:961:5
Warning: Not enough parameters for rrcurveto; actual: 5, expected: 6 pdf.worker.js:961:5
Warning: Not enough parameters for hstemhm; actual: 1, expected: 2 pdf.worker.js:961:5
Warning: Not enough parameters for hvcurveto; actual: 0, expected: 4 pdf.worker.js:961:5
Warning: Out of bounds subrIndex for callgsubr pdf.worker.js:961:5
Warning: Not enough parameters for rlineto; actual: 0, expected: 2 pdf.worker.js:961:5
Warning: Not enough parameters for rmoveto; actual: 0, expected: 2 pdf.worker.js:961:5
[...]

Issue #10175, opened 2018-10-23
Font rendering issue in viewer demo


DanDeMicco:
Attach (recommended) or Link to PDF file here:
https://farmersinsurance.ent.box.com/s/wpb9wgr99s3ud0au5p882nv3l06s7juf

Configuration:
- Web browser and its version: chrome 69.0.3497.100, firefox 63.0
- Operating system and its version: OSX 10.12.6
- PDF.js version: latest available on https://mozilla.github.io/pdf.js/web/viewer.html
- Is a browser extension: no 

Steps to reproduce the problem:
1. download the file from the link above
2. open with https://mozilla.github.io/pdf.js/web/viewer.html
3. Notice font rendering is messed up

What is the expected behavior? (add screenshot)
PDF renders text correctly

What went wrong? (add screenshot)
Fonts render incorrectly
<img width="1002" alt="screen shot 2018-10-23 at 11 33 29 am" src="https://user-images.githubusercontent.com/661216/47382608-885dad80-d6b7-11e8-95a5-13320e86fb17.png">

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

DanDeMicco:
I downloaded a few releases (including latest and a few older ones) from https://github.com/mozilla/pdf.js/releases, ran the viewer locally and opened the same PDF and found the fonts rendered fine for whatever reason

DanDeMicco:
Following the font debugging steps, I extracted one of the fonts to an .otf, and tried to install it on my machine and got some warnings.


[XeQCUKSY.otf.zip](https://github.com/mozilla/pdf.js/files/2507599/XeQCUKSY.otf.zip)
<img width="627" alt="screen shot 2018-10-23 at 11 53 05 am" src="https://user-images.githubusercontent.com/661216/47383978-c01a2480-d6ba-11e8-91e6-946d31fbca75.png">

Any help or recommendations on how to debug this further? Is the font in the PDF simply corrupt?

janpe2:
The link to the PDF file doesn't work or it requires a login.

DanDeMicco:
Hey @janpe2 sorry, looks like the permissions got edited. Here is a new link to the PDF:  https://farmersinsurance.ent.box.com/s/46dz61y6tdjpv3z1l6zysbt59o29hdom

timvandermeij:
[...]

PR #10520, opened 2019-02-04
Fallback to the built-in font renderer when font loading fails

Snuffleupagus:
After PR #9340 all glyphs are now re-mapped to a Private Use Area (PUA) which means that if a font fails to load, for whatever reason[1], all glyphs in the font will now render as Unicode glyph outlines.
This obviously doesn't look good, to say the least, and might be seen as a "regression" since previously many glyphs were left in their original positions which provided a slightly better fallback[2].

Hence this patch, which implements a *general* fallback to the PDF.js built-in font renderer for fonts that fail to load (i.e. are rejected by the sanitizer). One caveat here is that this only works for the Font Loading API, since it's easy to handle errors in that case[3].

*Please note:* This patch doesn't fix any of the underlying PDF.js font conversion bugs that's responsible for creating corrupt font files, however it does *improve* rendering in a number of cases; refer to this possibly incomplete list:

https://bugzilla.mozilla.org/show_bug.cgi?id=1524888
https://github.com/mozilla/pdf.js/issues/10175
https://github.com/mozilla/pdf.js/issues/10232

---
[1] Usually because the PDF.js font conversion code wasn't able to parse the font file correctly.

[2] Glyphs fell back to some default font, which while not accurate was more useful than the current state.

[3] Furthermore I'm not sure how to implement this generally, assuming that's even possible, and don't really have time/interest to look into it either.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -579,9 +579,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       });
     },
 
-    handleSetFont:
-        function PartialEvaluator_handleSetFont(resources, fontArgs, fontRef,
-                                                operatorList, task, state) {
+    handleSetFont(resources, fontArgs, fontRef, operatorList, task, state) {
       // TODO(mack): Not needed?
       var fontName;
       if (fontArgs) {
@@ -603,9 +601,9 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           return new TranslatedFont('g_font_error',
             new ErrorFont('Type3 font load error: ' + reason), translated.font);
         });
-      }).then((translated) => {
+      }).then(async (translated) => {
         state.font = translated.font;
-        translated.send(this.handler);
+        await translated.send(this.handler);
         return translated.loadedName;
       });
     },
@@ -615,7 +613,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       var glyphs = font.charsToGlyphs(chars);
       var isAddToPathSet = !!(state.textRenderingMode &
                               TextRenderingMode.ADD_TO_PATH_FLAG);
-      if (font.data && (isAddToPathSet || this.options.disableFontFace ||
+      if (font.data && (isAddToPathSet || font.disableFontFace ||
+                        this.options.disableFontFace ||
                         state.fillColorSpace.name === 'Pattern')) {
         var buildPath = (fontChar) => {
           if (!font.renderer.hasBuiltPath(fontChar)) {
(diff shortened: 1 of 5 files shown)


brendandahl:
Nice idea! Hadn't thought of using our own renderer for fallback. I'll review a bit later today.

src/core/evaluator.js
```diff
@@ -603,9 +601,9 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           return new TranslatedFont('g_font_error',
             new ErrorFont('Type3 font load error: ' + reason), translated.font);
         });
-      }).then((translated) => {
+      }).then(async (translated) => {
         state.font = translated.font;
-        translated.send(this.handler);
+        await translated.send(this.handler);
```
brendandahl: Does this mean the promise won't be resolved now until the font has been loaded on the main thread?

Snuffleupagus: Yes; but unfortunately I cannot see how we would avoid this, even if it makes font loading a *tiny* bit slower in general (given the need to wait here, rather than sending the font and hoping it works).

In order for the fallback to work the `disableFontFace` flag must have been set *before* any other operators are parsed, since otherwise [`PartialEvaluator.handleText`](https://github.com/mozilla/pdf.js/blob/c0d6e46e392b327996eb0964b7932cb5bdde1727/src/core/evaluator.js#L613) could easily be called before we've established that falling back is necessary. That would then, easily, lead to intermittent bugs where not all `FontPath`s are available to [`FontFaceObject.getPathGenerator`](https://github.com/mozilla/pdf.js/blob/c0d6e46e392b327996eb0964b7932cb5bdde1727/src/display/font_loader.js#L389) on the main-thread.

*In closing:* Given that all glyphs are now re-mapped to the PUAs, and that glyph rendering is thus *guaranteed* to break completely when a font is rejected by the sanitizer, in my opinion we don't have much choice here unless we'd somehow be able to generate font data that can never fail to load.

Snuffleupagus: An addendum: Please note that waiting for the main-thread to load certain data is already being done elsewhere on the worker-side. See for example the [`JpegStream` case](https://github.com/mozilla/pdf.js/blob/c0d6e46e392b327996eb0964b7932cb5bdde1727/src/core/evaluator.js#L438-L468), where we're checking that "simple" JPEG images can actually be rendered natively by the browser and otherwise fallback to the built-in https://github.com/mozilla/pdf.js/blob/master/src/core/jpg.js implementation.

brendandahl: Just testing locally with tracemonkey this is adding 100-250ms to overall load time of page 1. I'd really like to avoid any new two way dependencies too.

How about we load the normal way as usual, but if on the main thread we find a font that error's during loading we re-request the page with a flag enabling your new async code? Or we could disable native fonts entirely for the entire page, if we want to avoid the back and forth.

Snuffleupagus: > Just testing locally with tracemonkey this is adding 100-250ms to overall load time of page 1. 

I noticed that too and have started to realize that the current solution, although simple, probably isn't going to cut it unfortunately.  I've been thinking of a slightly modified approach, to get rid of the waiting, and will see if that works out better.

> I'd really like to avoid any new two way dependencies too.

I understand what you mean, but I'm not sure if that can be avoided. In any case, the solution will not be as simply as I initially hoped.

> How about we load the normal way as usual, but if on the main thread we find a font that error's during loading we re-request the page with a flag enabling your new async code?

While this is a nice idea it seems like a can of worms I'd rather not open, nor have time to hack on, since there's *a lot* of parameters/state associated with parsing/rendering. Hence trying to cancel/restart everything correctly seems like it could become very "messy" to deal with.

>  Or we could disable native fonts entirely for the entire page, if we want to avoid the back and forth.

I'm not sure if that's what you meant here, but please note that you *really* don't want to unconditionally set `disableFontFace` for an entire page, since that will guarantee that non-embedded fonts cannot be rendered.

---

Anyway, I've got an idea for a modified approach which does away with the waiting and only will require *temporary* storing a bit more state on the worker-side. I'll see what I can come up with :-)

Snuffleupagus: I've closed this one, and opened PR #10539 instead, after having an out of the blue idea for a (hopefully) better solution.

PR #10525, opened 2019-02-06
Convert more code to use `classList.toggle` with the `force` parameter

Snuffleupagus:
There's a bunch of code, in the viewer, which for historical reasons use `switch` statements to add and remove CSS classes.
This code can be simplified, and unnecessary duplication avoided, by using `classList.toggle` instead.


timvandermeij (approved):

timvandermeij:
Nice clean-up!

Issue #10529, opened 2019-02-09
Copying characters on supplementary planes is broken


a4lg:
Attach (recommended) or Link to PDF file here: [sample PDF file](https://github.com/mozilla/pdf.js/files/2847592/sample.pdf)

Configuration:
- Web browser and its version: Firefox 65.0 (64-bit)
- Operating system and its version: Windows 10 1810 (64-bit)
- PDF.js version: 2.0.943 (pdf.js prebuilt) and 2.1.97 (browser extension in Firefox 65.0)
- Is a browser extension: Yes and No (see above)

Steps to reproduce the problem:
1. Open [sample PDF file](https://github.com/mozilla/pdf.js/files/2847592/sample.pdf)
2. Select all (Ctrl+A) and copy all the text (Ctrl+C)
3. Paste to some text editor (with proper Unicode and emoji support)

What is the expected behavior? (add screenshot)

Following text (without notes) is copied:

    OK: ☺ (Note: U+263A)
    NG: 😃 (Note: U+1F603)
    In the following example, hidden text (not supposed to be copied) is encoded:
    𠁔𠁨𠁩𠁳𠀠𠁩𠁳𠀊𠁮𠁯𠁴𠀉𠁷𠁨𠁡𠁴𠀠𠁉𠀠𠁷𠁡𠁮𠁴𠁥𠁤𠀮
    (Note: garbled Chinese text which I encoded a hidden string as U+00xx -> U+200xx)

What went wrong? (add screenshot)

Following text (without notes) is copied (Note: formatted for readability):

    OK: ☺ (Note: U+263A)
    NG:  (Note: U+0F603)
    In the following example, hidden text (not supposed to be copied) is encoded:
    This is not what I wanted.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

I have a fix and will pull-request later.

PR #10530, opened 2019-02-09
Fix copying on supplementary plane characters

a4lg:
When I copy text from PDF displayed on pdf.js, only lower 16-bit of code points are copied (upper bits [that require surrogate pair] are ignored). See an example on issue #10529.

This is because PartialEvaluator.readToUnicode method used classic [String.fromCharCode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode) (which does not support out-of-UCS2 code points) instead of ES6's [String.fromCodePoint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint) (which does).

Merging this pull request replaces `String.fromCharCode` to `String.fromCodePoint` where necessary (two of four occurrences in `evaluator.js`).


timvandermeij:
The Windows failures do not look related to this patch.

@Snuffleupagus Could you review this? Also, would a "text" reference test catch this? If so, it should be added.

Snuffleupagus:
> The Windows failures do not look related to this patch.

Firefox was updated on the bot, this is the "usual" sort of fallout from that.

> Also, would a "text" reference test catch this? If so, it should be added.

Yes it will[1], and a test-case is a requirement for accepting the patch.

---
[1] Try adding `#textLayer=visible` to a viewer with/without the patch, with the `pdfBugEnabled` pref set, and compare the output.

a4lg:
OK, I'm working on the test.

a4lg:
Oh, making the test was harder than I expected (I could not figure out what "text" tests do).
I can ... at least regenerate test PDF file only with **free/open source** fonts ([Source Sans Pro](https://github.com/adobe-fonts/source-sans-pro) and [Hanazono Mincho](http://fonts.jp/hanazono/)): [issue10529.pdf](https://github.com/mozilla/pdf.js/files/2848441/issue10529.pdf)

a4lg:
I read the "Contributing" page again and added a testcase. I'm not sure this is okay though.

a4lg:
I ran `gulp makeref` and noticed that some glyphs are missing on Linux (because of insufficient out-of-UCS2 coverage). I think this is still OK since it (barely) shows missing codepoint value in hex and the image after the fix looks very different.

Here is the reference image before/after the fix on Ubuntu 18.10 + additional fonts (as described in "[Bots](https://github.com/mozilla/pdf.js/wiki/Bots)") + Firefox 65.0.

Before the fix:
issue10529-bug-ref-1

After the fix:
issue10529-fixed-ref-1

I haven't tested it on Windows but I think it will not show missing glyphs because of preinstalled PMingLiU-ExtB font.

test/test_manifest.json
```diff
@@ -1451,6 +1451,13 @@
        "type": "eq",
        "nativeImageDecoderSupport": "none"
     },
+    {  "id": "issue10529",
+       "file": "pdfs/issue10529.pdf",
+       "md5": "1a4d404a137c610ff0c747cbea3b8666",
+       "rounds": 1,
+	   "link": false,
```
Snuffleupagus: Please, no tabs anywhere in the code-base.

Also, please follow https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

a4lg: I didn't notice a tab, thanks.

a4lg:
Fixed code style and squashed.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

fgilio:
Hi! I have a little question: Does this execute when copying text in the viewer or when building the text layer?

timvandermeij:
This runs when the text layer is built and therefore it impacts the way the text can be copied.

fgilio:
Awesome, thanks for clarifying!

Issue #10416, opened 2019-01-04
Release version 2.1


timvandermeij:
There some some `api-minor` patches and the example code on the website changed to the new syntax for `getViewport`, which caused some confusion for people that use the NPM package. We should make a release of version 2.1 to solve this and other issues we found and fixed after version 2.0.

wojtekmaj:
Hi Tim, is there any schedule? I'm really looking forward to this release!

wojtekmaj:
BTW this getViewport change is a breaking change in documented area of the app, so that's actually api-major unless you support both old and new way of passing the arguments.

timvandermeij:
I don't think we have anything in the queue holding back a release. I don't mind taking core of this, so perhaps we can already do it next week if everyone agrees (/cc @brendandahl @yurydelendik).

The `getViewport` change is backwards compatible; refer to https://github.com/mozilla/pdf.js/pull/10369/files#diff-0ecad279ed2252e3eb47a4d96ec1463cR921. The old way of calling will still work and show a deprecation notice that the user should change it. It will only be removed in a major version (3.x).

Snuffleupagus:
> make sure I'm not forgetting about something that really needs to be in version 2.1

Ideally, issue #10391 really ought to be fixed before the next release.

timvandermeij:
I'm able to make the release soon, but I'm a bit short on time to also address the issue mentioned above. @brendandahl Last time we released version 2.0, did you look at why the master branch for `pdfjs-dist` was not updated (see #10391). It would be nice if we can get that fixed before making a new release.

timvandermeij:
The issue above is now done, so I'll try to make the new release this weekend unless there are major objections (please comment here in that case). Doing that will also resolve the `getViewport` confusion.

timvandermeij:
The release of version 2.1.245 is done. However, because of misunderstood instructions on https://github.com/mozilla/pdf.js/wiki/Release-Process the version is called 2.2.2 on `pdfjs-dist` and NPM because I updated the `pdfjs.config` file before making the release. To prevent this from happening again, I updated to page to remove the section from https://github.com/mozilla/pdf.js/wiki/Release-Process/055c021ab425b3e88877a40d8cb18d3956d4832e about updating the version prefix since I read that as that it should happen first because `api-minor` changes were merged.

[...]

PR #10533, opened 2019-02-09
Bump the version number in `pdfjs.config`

timvandermeij:


pdfjs.config
```diff
@@ -1,6 +1,6 @@
 {
   "betaVersion": "",
```
Snuffleupagus: Just asking, but shouldn't there have been a beta version?

Also, what's up with the version number difference at https://github.com/mozilla/pdf.js/releases (i.e. `v2.1.245`) vs https://github.com/mozilla/pdfjs-dist/releases (i.e. `v2.2.2`).

timvandermeij: I commented about this on #10416. Basically, the instructions page wasn't really clear and I missed a step. I updated the page so that the next release will happen correctly again; sorry for the confusion!

PR #10534, opened 2019-02-09
Try to hack-around the broken `pdfjs-dist` links, such that they will point to the main library releases (issue 10391)

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
I verified manually that this indeed fixes the problem. Thank you!

PR #10539, opened 2019-02-11
[api-minor] Fallback to the built-in font renderer when font loading fails

Snuffleupagus:
After PR #9340 all glyphs are now re-mapped to a Private Use Area (PUA) which means that if a font fails to load, for whatever reason[1], all glyphs in the font will now render as Unicode glyph outlines.
This obviously doesn't look good, to say the least, and might be seen as a "regression" since previously many glyphs were left in their original positions which provided a slightly better fallback[2].

Hence this patch, which implements a *general* fallback to the PDF.js built-in font renderer for fonts that fail to load (i.e. are rejected by the sanitizer). One caveat here is that this only works for the Font Loading API, since it's easy to handle errors in that case[3].

The solution implemented in this patch does *not* in any way delay the loading of valid fonts, which was the problem with my previous attempt at a solution, and will only require a bit of extra work/waiting for those fonts that actually fail to load.

*Please note:* This patch doesn't fix any of the underlying PDF.js font conversion bugs that's responsible for creating corrupt font files, however it does *improve* rendering in a number of cases; refer to this possibly incomplete list:

 - [Bug 1524888](https://bugzilla.mozilla.org/show_bug.cgi?id=1524888)
 - Issue #10175
 - Issue #10232
 - Issue #10544

---
[1] Usually because the PDF.js font conversion code wasn't able to parse the font file correctly.

[2] Glyphs fell back to some default font, which while not accurate was more useful than the current state.

[3] Furthermore I'm not sure how to implement this generally, assuming that's even possible, and don't really have time/interest to look into it either.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -666,6 +666,10 @@ class PDFDocument {
     });
   }
 
+  fontFallback(id, handler) {
+    return this.catalog.fontFallback(id, handler);
+  }
+
   cleanup() {
     return this.catalog.cleanup();
   }
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -610,37 +610,18 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       });
     },
 
-    handleText: function PartialEvaluator_handleText(chars, state) {
-      var font = state.font;
-      var glyphs = font.charsToGlyphs(chars);
-      var isAddToPathSet = !!(state.textRenderingMode &
-                              TextRenderingMode.ADD_TO_PATH_FLAG);
-      if (font.data && (isAddToPathSet || this.options.disableFontFace ||
-                        state.fillColorSpace.name === 'Pattern')) {
-        var buildPath = (fontChar) => {
-          if (!font.renderer.hasBuiltPath(fontChar)) {
-            var path = font.renderer.getPathJs(fontChar);
-            this.handler.send('commonobj', [
-              font.loadedName + '_path_' + fontChar,
-              'FontPath',
-              path
-            ]);
-          }
-        };
+    handleText(chars, state) {
+      const font = state.font;
+      const glyphs = font.charsToGlyphs(chars);
 
-        for (var i = 0, ii = glyphs.length; i < ii; i++) {
-          var glyph = glyphs[i];
-          buildPath(glyph.fontChar);
-
-          // If the glyph has an accent we need to build a path for its
-          // fontChar too, otherwise CanvasGraphics_paintChar will fail.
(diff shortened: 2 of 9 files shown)


kalda341:
This is something that we would really benefit from. We currently have a hacky workaround in our application involving getting the document with `{stopAtErrors: true}` and attempting to render, catching any errors that occur and re-rendering with different options if we failed.

This is a much nicer solution.

brendandahl (approved):
Looks good! I forgot we had a cache of all the glyphs, that makes things much easier. It would be great if we could get a follow up bug to add a test case. I am also thinking, for testing, we may want a way to ensure that this path isn't allowed, unless the test specifically allows it (maybe a pref that defaults to enabled?).

Snuffleupagus:
> Looks good! I forgot we had a cache of all the glyphs, that makes things much easier.

Thank you; I'd also forgotten all about the `glyphCache`, and somehow it popped into my head when I was doing the dishes the other day and everything just fell into place :-)

> It would be great if we could get a follow up bug to add a test case.

The reason that I didn't add a test-case here, is that this is nothing more than a work-around. The end goal should still, as far as I'm concerned, be to fix the font bugs that cause us to generate a font file that the OpenType Sanitizer ends up rejecting.

>  I am also thinking, for testing, we may want a way to ensure that this path isn't allowed, unless the test specifically allows it (maybe a pref that defaults to enabled?).

In an ideal world, where the underlying PDF.js bugs have been fixed, I'm not sure what you'd actually test in that case!? (Unless you somehow corrupted the font file on the main-thread, prior to loading it.)

brendandahl:
> The reason that I didn't add a test-case here, is that this is nothing more than a work-around. The end goal should still, as far as I'm concerned, be to fix the font bugs that cause us to generate a font file that the OpenType Sanitizer ends up rejecting.

I agree, I'd still like to fix the underlying issue and for that specific PDF I have a fix in progress. Though it would still be nice to test this new code path and make sure it always works.

> In an ideal world, where the underlying PDF.js bugs have been fixed, I'm not sure what you'd actually test in that case!? (Unless you somehow corrupted the font file on the main-thread, prior to loading it.)

I think we'd need a font that we intentionally never fix. Or an idea I like better is: during the test, the test framework could override the font loading api and intentionally make it fail. We could add an option to test_manifest.json for this.

Snuffleupagus:
> Or an idea I like better is: during the test, the test framework could override the font loading api and intentionally make it fail. We could add an option to test_manifest.json for this.

Totally agree, that seems like the nicest/cleanest solution.

PR #10548, opened 2019-02-13
Generate the `default_preferences.json` file from `AppOptions`

Snuffleupagus:
Currently any editing of the preferences require updates in *three* separate files, which isn't a great developer experience to say the least.

This has annoyed me sufficiently to write this patch, which moves the definition of all preferences into `AppOptions` and adds a new `gulp` task to generate the `default_preferences.json` file for the builds where it's needed.


timvandermeij (approved):

timvandermeij:
Nice work! I'm glad that this has become easier to manage now.

PR #10554, opened 2019-02-15
Simplify the `updatetextlayermatches` event handling in `TextLayerBuilder`

Snuffleupagus:
This implements the nice suggestion from https://github.com/mozilla/pdf.js/pull/10099#discussion_r219698000, which I at the time didn't think would work. @timvandermeij I'm sorry about that!
I'll probably have to plead temporary insanity here, since it *should* have been totally obvious to me how this could be implemented. By simply not registering the event until the `textLayer` is actually rendered, removing the event on `cancel` works just fine.

This patch also removes the `pagecancelled` event, given that it's no longer used anywhere in the code-base and that its implemention was flawed since it wasn't guaranteed to be called in *every* situation where rendering was actually cancelled.

### DIFF
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -154,6 +154,7 @@ class PDFPageView {
 
   reset(keepZoomLayer = false, keepAnnotations = false) {
     this.cancelRendering(keepAnnotations);
+    this.renderingState = RenderingStates.INITIAL;
 
     let div = this.div;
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
> I'll probably have to plead temporary insanity here

No worries! Thank you for simplifying the implementation.

PR #10556, opened 2019-02-15
[api-minor] Expose the existence of a `Collection` dictionary via the  `getMetadata` API method (issue 10555)

### LINKED ISSUE #10555: Can I check whether pdf is portfolio?
[example of pdf portfolio](https://drive.google.com/file/d/10kBOZzg-1wK4BGyHDbO15_JUs4ynl_rU/view?usp=sharing) 

Configuration:
- Web browser and its version: Chrome
- PDF.js version: 2.0.487

About pdf portfolios: https://helpx.adobe.com/acrobat/using/overview-pdf-portfolios.html

Hello, as I know currently pdf portfolios are not previewable by pdf.js and only fallback is shown:
image

I am trying to make a validation of pdf files to block ones that can not be previewed. Portfolio pdf succesffuly rendered in a fallback with no exceptions or warnings about non supported features. Also, portfolios always have attachments but I can't rely on that cause simple pdf files can have them too and I do not want to block them.
The only way I found out is to check content of fallback, for example check - if pdf has only one page with this fallback then it is a portfolio and can't be previewed. But I am not sure it is a very reliable method.

Is there any convinient way to detect if pdf is a portfolio, like some `isPortfolio` flag?

Snuffleupagus:
Given the complexity of this functionality, and the fact that it doesn't seem widely used, I highly doubt that it'd ever make sense to support Collections; see also https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#M11.9.39646.2Heading.824.Collections

Fixes #10555.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -401,8 +401,7 @@ class PDFDocument {
       if (this.acroForm) {
         this.xfa = this.acroForm.get('XFA');
         const fields = this.acroForm.get('Fields');
-        if ((!fields || !Array.isArray(fields) || fields.length === 0) &&
-            !this.xfa) {
+        if ((!Array.isArray(fields) || fields.length === 0) && !this.xfa) {
           this.acroForm = null; // No fields and no XFA, so it's not a form.
         }
       }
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good. Thanks!

PR #10558, opened 2019-02-17
Re-factor the `PDFDataRangeTransport` unit-tests and enable them in Node.js/Travis

Snuffleupagus:
There doesn't appear to be any particular reason for only running these unit-tests in browsers, since the `PDFDataRangeTransport` functionality itself should be back-end agnostic.

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -14,13 +14,12 @@
  */
 
 import {
-  buildGetDocumentParams, NodeCanvasFactory, NodeFileReaderFactory,
-  TEST_PDFS_PATH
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for doing this!

Snuffleupagus:
Thanks for the very quick review!

I believe that we're now running all the possible unit-tests in Node.js/Travis, with the *sole exception* of test https://github.com/mozilla/pdf.js/blob/ff74a66672c2980aa3dc2dd1b6f1407026b831de/test/unit/api_spec.js#L656-L663

PR #10560, opened 2019-02-17
Update translations and packages

### LINKED ISSUE #10177: Investigate upgrading Gulp, Webpack and Node.js dependencies

timvandermeij:


timvandermeij:
Trivial.

PR #10573, opened 2019-02-21
Avoid truncating/breaking some Type3 glyphs in `compileType3Glyph` (bug 1245391, issue 10568)

### LINKED ISSUE #10568: "ほ" is displayed like "は"

[sample01.pdf](https://github.com/mozilla/pdf.js/files/2884257/sample01.pdf)

[sample02.pdf](https://github.com/mozilla/pdf.js/files/2884258/sample02.pdf)

Configuration:
FireFox 65.0.1 64bit
Windows 10 ver 1803
- PDF.js version:
2.1.97
- Is a browser extension:
No

1. Open sample1.pdf or sample2.pdf by FireFox.

"ほ" want to be displayed as the ruby of "保" in sample1.pdf.
"ほう" want to be displayed as the ruby of "報" in sample2.pdf.

"は" is displayed as the ruby of "保" in sample1.pdf.
"はう" is displayed as the ruby of "報" in sample2.pdf.

sample1

sample2

The images displayed by Edge (17.17134) on the left and FireFox (65.0.1) on the right.

Snuffleupagus:
*Hopefully this patch makes sense, since I cannot claim to fully understand this function.*

With the changes made in PR #3354 *some* Type3 glyph outlines are no longer rendering correctly, since the final paths were being accidentally ignored.
The fact that Type3 fonts are not very common in PDF documents, and that most Type3 glyphs are unaffected by this regression, probably explains why this has gone unnoticed since 2013.

Fixes [bug 1245391](https://bugzilla.mozilla.org/show_bug.cgi?id=1245391)
Fixes #10568


timvandermeij:
@brendandahl Could you perhaps check this?

src/display/canvas.js
```diff
@@ -325,7 +325,6 @@ function compileType3Glyph(imgData) {
 
       coords.push(p % width1);
       coords.push((p / width1) | 0);
-      --count;
```
yurydelendik: removing this is not preferable: could you try to move `--count` after `points[p]` is set to 0. you can see that at line 323 it may or may not do that, so in the else branch let's add the check `if (!points[p]) --count;`

Snuffleupagus: Sure, let's see how this works out on the bots :-)

(If it turns out to be more complicated, someone who actually understands this code will have to take over fixing the bug though.)

Snuffleupagus: It seems that this worked out just fine; does this mean that the patch is good to go now?

Snuffleupagus:
@yurydelendik Gentle review ping; is the updated patch OK?

yurydelendik (approved):
Thank you for the patch.

PR #10575, opened 2019-02-22
Generate .d.ts files from jsdoc information

oBusk:
I was looking to update https://www.npmjs.com/package/@types/pdfjs-dist to 2.0.0 but figured it would be better if the package contained the typings and tried to use https://www.npmjs.com/package/tsd-jsdoc to generate them. 

This is a start for generating `.d.ts` files from JSDoc to provide typeinformation for developers.

I have no idea if the generated `types.d.ts` is usable right now. 

Will also need some additions to gulpfile to add `types` to package.json and ship the typefile. Not sure if I can get that together.

Related: https://github.com/mozilla/pdf.js/pull/9369, https://github.com/mozilla/pdf.js/issues/7909


gulpfile.js
```diff
@@ -948,7 +949,10 @@ gulp.task('jsdoc', function (done) {
 
   var JSDOC_FILES = [
     'src/doc_helper.js',
+    'src/interfaces.js',
+    'src/core/primitives.js',
```
Snuffleupagus: This file makes absolutely no sense to include in API documentation, since it contains functionality which *only* runs in a web worker and isn't even accessible for an API consumer anyway (the same goes for most/all code in the `src/core/` folder); please see https://github.com/mozilla/pdf.js/pull/9369#issuecomment-358589858 for additional information.

---

Also, what's the point of *this* PR when there's already the #9369 in progress, which seems much more complete overall?

timvandermeij:
Even though this doesn't seem complete yet, I do really like the idea of generating the `ts` files from the JSDoc comments. For the other open PR, I'm still very unsure about maintenance since keeping multiple files in sync can be error-prone. This would allow us to still provide `ts` files and don't have the problem of added maintenance.

Perhaps you can turn this PR into a working version that contains the code to generate those files during a build and to fix up one or two classes with the required type annotations in the JSDocs? That way we can check how useful the output is without putting in a lot of work in this PR to convert everything at once. This also allows us to merge the ground work and extend it in chunks later on.

gulpfile.js
oBusk: Ok I see. Would it then make more sense to declare the interfaces that are used in the API with JSDoc so that the documentation can refer to them without risking confusing them for the datatypes used internally in the worker? 

I suppose I'd try to mock them up in `doc_helper` so that they may be corrected later.

---

Generating `.d.ts` files from JSDoc means less maintance because you don't have to remember to edit the external `.d.ts` file. Another bonus is that maintaining the JSDoc comments for dts typings would improve/structure the API documentation and vice versa.  I'm not guaranteeing that it'd be perfect, but to me it seems like it would be useful.

oBusk:
I've [updated](https://github.com/mozilla/pdf.js/pull/10575/commits/15f22c3e050c70bbaa3a6684f73c76f253e6eb75) `gulpfile.js` to include the types when building, and with the [addition](https://github.com/mozilla/pdf.js/pull/10575/commits/c33e65d760bc1a69edee7e790f7827852a5af285) to declare `api.js` as the module `pdfjs-dist`, I managed to import and get basic typing. So I guess the following steps would be to remove the internal files that I incorrectly included to get the types like `Name` etc., from the generating and declaring the necessary data types in either `api.js` or `doc_helper.js`. 

Was that a correct interpretation of your previous comment, @timvandermeij?

timvandermeij:
Sounds good to me. It's mainly about making a minimal working version so it's easier to review and we can iterate on it. For example, are the changes in https://github.com/mozilla/pdf.js/pull/10575/files#diff-18ca06e2bd5be4c9132b51f78388b8f0 already necessary given that you are mainly touching `api.js` for now?

oBusk:
@timvandermeij I took another swing, focusng on the files that was inputed into the JSDoc API documentation script originally. It's outputting a error free typingsfile and the basic typings seems to be correct.

image

src/core/annotation.js
```diff
@@ -494,6 +496,11 @@ class AnnotationBorderStyle {
     }
   }
 
+  /**
```
timvandermeij: Since this is only for documentation/typing purposes, could we perhaps find a way to move these kind of generic typings to a central place, for example https://github.com/mozilla/pdf.js/blob/master/src/doc_helper.js? The same applies to other of these blocks below.

src/display/api.js
```diff
@@ -883,42 +919,49 @@ class PDFPageProxy {
   }
 
   /**
-   * @return {number} Page number of the page. First page is 1.
+   * @type {number} Page number of the page. First page is 1.
```
timvandermeij: In the other files we still have `return`, so why is that changed to `type` here and below? It actually represents the return value of the function.

timvandermeij (commented):
Looks much better this way! I'm confident in the approach with regards to generating the typing files from the JSDoc comments since it makes maintenance and correctness much easier in my opinion. I left some comments after looking at this. I'm also curious what other contributors/maintainers think about this approach.

src/display/api.js
oBusk: The `.d.ts` was generated with no types when using `@return`, and as far as I could figure out, `get` properties should be documented like members using `@type`. In the current html API (draft) documentation you can see that the type is not shown https://mozilla.github.io/pdf.js/api/draft/PDFPageProxy.html#pageNumber

Here is a comparison
* Using `@return`
  image

* Using `@type`
  image

The reason why I did not change in all files was that i was focusing on `api.js`.

timvandermeij: Good point, I had no idea that there was actually a difference here. It's probably best to keep it as-is for this patch and create a follow-up issue to adjust the other places later.

acchou:
TypeScript 3.7 has native support for generating declaration files from jsdoc annotated source: https://www.typescriptlang.org/docs/handbook/release-notes/overview.html#--declaration-and---allowjs

Seems worth a try.

timvandermeij:
Most of the documentation improvements from this PR have already been merged in other PRs, so I would suggest to rebase this first to see which changes are left, since we concluded that the approach of generating the TypeScript bindings from JSDoc comments is the preferred approach. We can then choose to keep this way of generating the bindings, or look into the way from https://github.com/mozilla/pdf.js/pull/10575#issuecomment-569175253 which also looks promising.

oBusk:
@timvandermeij That sounds like a sensible way to approach the matter. I'll rebase and see how that turns out. Then I'll try typescript 3.7 like @acchou suggested.

oBusk:
@timvandermeij I managed to get it to work again. 

Types look something like this right now;
https://gist.github.com/oBusk/8bdbbdae7d5b23fd419c4af3fff8f57c

src/display/api.js
```diff
@@ -407,6 +420,7 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
     });
 }
 
+/** @lends <global> */
```
Snuffleupagus: We're already exposing a bunch interfaces for the unit-tests, hence please try to see if simply listing `PDFDocumentLoadingTask` with the exports at the bottom of the file works instead.

src/display/api.js
```diff
@@ -885,14 +922,14 @@ class PDFDocumentProxy {
  * PDF page operator list.
  *
  * @typedef {Object} PDFOperatorList
- * @property {Array} fnArray - Array containing the operator functions.
- * @property {Array} argsArray - Array containing the arguments of the
- *                               functions.
+ * @property {Array<Function>} fnArray - Array containing the operator
```
Snuffleupagus: This looks strange, since there's nothing which is `typeof "function"` in the array.

Snuffleupagus:
As mentioned previously, please remove any `src/core/` changes (or references to such files) since that code runs in a worker-thread and is thus not actually accessible directly in the API.

Furthermore, as also mentioned previously, please place any typescript *specific* definitions in a helper file (e.g. `src/doc_helper.js`) rather than directly in the API and let's avoid a bunch of TODO comments in this code.

Finally, please note https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

gulpfile.js
```diff
@@ -1102,6 +1103,28 @@ gulp.task("jsdoc", function(done) {
   });
 });
 
+gulp.task("dts", function(done) {
+  console.log();
+  console.log("### Generating typescript definitions");
+
+  var JSDOC_FILES = [
```
timvandermeij: This changed in the `jsdoc` target since this patch was made I think, so making this equal to https://github.com/mozilla/pdf.js/pull/10575/files#diff-b9e12334e9eafd8341a6107dd98510c9R1092 allows you to remove all `src/core` and `src/shared` changes from this patch and include all TypeScript-specific constructs in `src/doc_helper.js` as indicated by @Snuffleupagus.

src/display/api.js
```diff
@@ -82,13 +87,19 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
   createPDFNetworkStream = pdfNetworkStreamFactory;
 }
 
+/* eslint-disable max-len */
+/**
+ * @typedef {Int8Array | Uint8Array | Int16Array | Uint16Array | Int32Array | Uint32Array | Uint8ClampedArray | Float32Array | Float64Array} TypedArray //
```
timvandermeij: We should be able to remove the trailing `//` here.

timvandermeij:
The code looks simpler already and the generated types also look good, so it's good to see that this approach is working well. We have left a few comments above to simplify the code even further.

oBusk:
I've the comments and I'll try to look at them when I have time

I just felt like I should share what I found; I feel like using typescript 3.7 is probably a more "correct" way forward, as tsd-jsdoc is far less used than typescript (obviously). I just quickly [tried](https://github.com/oBusk/pdf.js/tree/05de182c094484f61c24216307dff1803c3b9888) but ran into similiar issues of refering to classes that are not declared globally and thus not able to export them, meaning `tsc` will fail with 
```
src/display/api.js:21:1 - error TS9005: Declaration emit for this file requires using private name 'PDFDocumentLoadingTask'. An explicit type annotation may unblock declaration emit.
```

I can't figure out a workaround.

src/display/api.js
oBusk: Exporting a `const` that happens to be assigned to a `class` does not mean that class get's exported. And I don't know if there's any way to export the `class` itself from within the closure.

src/display/api.js
oBusk: You're right. I think I just guessed based on the name. I ran an example and set the type to be `Array<number>` instead.

P0oOOOo0YA:
OK so what's the best approach here for Typescript developers? Install typescript 3.7 and let the typescript parser go through the source, parse `JSDoc` and build the types dynamically. At least there should be some hints for TS developers on the README.

oBusk:
@P0oOOOo0YA if you want to use pdfjs in your typescript project, currently your option is to install and use https://www.npmjs.com/package/@types/pdfjs-dist 

I've tried somewhat to create typings from the source, but using typescript I ran into issues due to the structure of the JS code that I don't grasp enough to really figure out what to do with typescript.

tamuratak:
I have made a PR, oBusk/pdf.js/pull/1,  to make `tsc` generates declaration files of the PDF.js repository. We have to add type annotations to some exported constants explicitly.

buzinas:
@oBusk Can you look into @tamuratak's PR above?

ineiti:
I tried to merge and fix the conflicts here: #12102

Unfortunately I cannot get it to run: https://github.com/ineiti/pdf_example doesn't seem to find the worker. It fails with:
```
core.js:4081 ERROR Error: Uncaught (in promise): Error: No "GlobalWorkerOptions.workerSrc" specified.
Error: No "GlobalWorkerOptions.workerSrc" specified.
```

And there is some work still to be done to have cleaner typescript definitions, as described in #12102

timvandermeij:
Let's continue this discussion in the new PR since it's more up-to-date than this one.

PR #10577, opened 2019-02-23
Move worker-thread only functions from `src/shared/util.js` and into a new `src/core/core_utils.js` file

Snuffleupagus:
The `src/shared/util.js` file is being bundled into both the `pdf.js` and `pdf.worker.js` files, meaning that its code is by definition duplicated.
Some main-thread only utility functions have already been moved to a separate `src/display_utils.js` file, and this patch simply extends that concept to utility functions which are used *only* on the worker-thread.

Note in particular the `getInheritableProperty` function, which expects a `Dict` as input and thus *cannot* possibly ever be used on the main-thread.


timvandermeij (approved):

timvandermeij:
Splitting the utilities seems like a good idea to me. Thanks!

PR #10579, opened 2019-02-24
Re-factor the `PDFSidebar` constructor to simplify its call-site (PR 10123 follow-up)

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #10582, opened 2019-02-25
Use `Promise.prototype.finally` in the `PDFRenderingQueue.renderView` method

Snuffleupagus:
Since [`finally`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally) is now supported through `src/shared/compatibility.js`, the temporary variable is no longer necessary.

Also, this patch fixes the inconsistent return type of `PDFRenderingQueue.getHighestPriority` when no pages/thumbnails are visible.


timvandermeij (approved):

timvandermeij:
That's nice, thanks!

PR #10585, opened 2019-02-25
Load built-in CMap files using the Fetch API when possible

### LINKED ISSUE #10587: Viewer does not support local files on Chromium
Configuration:
- Web browser and its version: QtWebEngine 5.12 (Chromium 69), Chrome 72
- PDF.js version: 2.0.943, 2.1.266
- Is a browser extension: No

1. Open viewer locally (file:///C:/pdftest/web/viewer.html)

Embedded PDF shown (it works in Firefox ESR 60).

What went wrong? 

Error in console: Fetch API cannot load file:///C:/pdftest/web/compressed.tracemonkey-pldi-09.pdf. URL scheme must be "http" or "https" for CORS request.

I have seen #9361 and #9020 but I don't think that's the solution in this case.

Snuffleupagus:
Fixes #10587.

### DIFF
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -490,5 +490,6 @@ export {
   StatTimer,
   DummyStatTimer,
   isFetchSupported,
+  isValidFetchUrl,
   loadScript,
 };
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -15,7 +15,7 @@
 
 import {
   assert, CMapCompressionType, removeNullCharacters, stringToBytes,
-  unreachable, Util, warn
+  unreachable, URL, Util, warn
 } from '../shared/util';
 
 const DEFAULT_LINK_REL = 'noopener noreferrer nofollow';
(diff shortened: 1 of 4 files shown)


src/display/display_utils.js
```diff
@@ -428,6 +448,19 @@ class DummyStatTimer {
   }
 }
 
+function isValidFetchUrl(url, baseUrl) {
```
timvandermeij: Could you add a unit test for this new utility function?

src/display/display_utils.js
```diff
@@ -66,19 +67,41 @@ class DOMCMapReaderFactory {
...
+      }).catch((reason) => {
+        throw new Error(`Unable to load ${this.isCompressed ? 'binary ' : ''}` +
+                        `CMap at: ${url}`);
+      });
     }
-    return new Promise((resolve, reject) => {
-      let url = this.baseUrl + name + (this.isCompressed ? '.bcmap' : '');
 
-      let request = new XMLHttpRequest();
+    // The Fetch API is not supported.
```
timvandermeij: Is there a reason why we don't use a polyfill for `fetch` instead of having manual fallback logic? Since it's not too much code I don't really have a strong preference, but it looks like `babel-preset-env` can do this (https://stackoverflow.com/questions/44908044/babel-not-polyfilling-fetch-when-using-babel-preset-env) and there is also https://github.com/github/fetch.

src/pdf.js
```diff
@@ -32,24 +32,28 @@ let pdfjsDisplayAPICompatibility = require('./display/api_compatibility.js');
 
 if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
   const isNodeJS = require('./shared/is_node.js');
+  const isFetchSupported = require('./shared/is_fetch_supported.js');
```
timvandermeij: If we move this up, we probably don't have to repeat this below.

timvandermeij (commented):
Looks good with the comments addressed.

src/display/display_utils.js
Snuffleupagus: That's a fair point, will fix this.

src/display/display_utils.js
Snuffleupagus: Given the (somewhat) advanced usage of `fetch` found in https://github.com/mozilla/pdf.js/blob/master/src/display/fetch_stream.js it's not clear to me if a polyfill would generally work well.

The main reason for *not* attempting to polyfill though, is that as you can see in the added checks `fetch` only works for the `http`/`https` protocols, whereas `XMLHttpRequest` can work for e.g. the `file`/`ftp`/... protocols as well.

src/pdf.js
Snuffleupagus: This was done purposely, to avoid loading this in `MOZCENTRAL` builds.

timvandermeij (approved):

timvandermeij:
Looks really good. Thanks!

PR #10589, opened 2019-02-26
Run the `custom_spec` unit-tests in Node.js/Travis (PR 10537 follow-up)

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Good find! With the necessary canvas factories in place, this does indeed work on Travis now.

Issue #10282, opened 2018-11-21
Paths are not rendered for svg-backend for some documents


valerii-cognite:
Attach (recommended) or Link to PDF file here: [1.PDF](https://github.com/mozilla/pdf.js/files/2604367/1.PDF)

Configuration:
- Web browser and its version: Firefox 62, reproduces on others as well
- Operating system and its version: Os X 10.13, reproduces on others as well
- PDF.js version: pdfjs-dist: 2.0.943
- Is a browser extension: no

Steps to reproduce the problem:
Load attached document in svg mode

What is the expected behavior? (add screenshot)
To show paths

What went wrong? (add screenshot)
All the paths for the document are not exist in DOM. Everything fine if to try to render in a canvas mode.
SVG:
![screen shot 2018-11-21 at 14 34 00](https://user-images.githubusercontent.com/38251534/48844658-caebd600-ed9a-11e8-82f9-0f09d5d6671c.png)
Canvas:
![screen shot 2018-11-21 at 14 35 24](https://user-images.githubusercontent.com/38251534/48844659-caebd600-ed9a-11e8-9a34-0a71bfd29e59.png)

PR #10590, opened 2019-02-26
Fix missing moveTos in SVG paths

### LINKED ISSUE #9167: When rendering pdfs using SVGs, SVG:Path error is shown
- PDF.JS version = 2.0.120

1. Render PDF from base64 encoded string to svg

What went wrong? 
I receive a JavaScript error when rendering as svg.
image

### LINKED ISSUE #10522: SVG path is incorrectly broken into 2 nodes
[test4.pdf](https://github.com/mozilla/pdf.js/files/2834751/test4.pdf)

Configuration:
- Web browser and its version: Chrome 56.0.2924.76
- Operating system and its version: Linux Mint 18.3
- PDF.js version: commit c0d6e46e392b327996eb0964b7932cb5bdde1727
- Is a browser extension: no

1. run "node pdf2svg.js test4.pdf"
2. load the output test4.1.svg into Chrome
3. Check the console error log.  You will see this: 
chrome_console_error

No errors.

One path in the output got split into 2 parts.
The initial "M" is in one node and the following "L" is in the next node.
These should be in the same node.
bad_svg_node

janpe2:
When `OperatorList` splits page contents into chunks of 1000 operators, some paths may be divided into two `OPS.contructPath` operators. The canvas back-end has no problems but the SVG back-end produces two `<svg:path>` elements and the latter one doesn't start with a moveTo command. Firefox doesn't draw this kind of SVG paths and Chrome logs an error in the console. 

My idea is to append path commands to the same `<svg:path>` element until the path is terminated by a painting operator, like `f` or `S`. It's the standard PDF behavior that painting operators finish and end the current path.

Fixes #9167
Fixes #10522

Also issue #10282 seems to have this problem. Now a missing line segment is easily visible in the top-right corner of the page
kuva

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -940,7 +940,6 @@ SVGGraphics = (function SVGGraphicsClosure() {
     constructPath: function SVGGraphics_constructPath(ops, args) {
       var current = this.current;
       var x = current.x, y = current.y;
-      current.path = this.svgFactory.createElement('svg:path');
       var d = [];
       var opLength = ops.length;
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #10591, opened 2019-02-27
Add unique glyph names for CFF fonts.

brendandahl:
Printing CFF fonts on MacOS was broken with the previous approach of just mapping
all the glyphs to notdef.

Fixes:
https://bugzilla.mozilla.org/show_bug.cgi?id=880181
https://bugzilla.mozilla.org/show_bug.cgi?id=1523991
https://bugzilla.mozilla.org/show_bug.cgi?id=1523667
https://bugzilla.mozilla.org/show_bug.cgi?id=1524367

### DIFF
--- a/src/core/cff_parser.js
+++ b/src/core/cff_parser.js
@@ -96,6 +96,8 @@ var CFFStandardStrings = [
   'Black', 'Bold', 'Book', 'Light', 'Medium', 'Regular', 'Roman', 'Semibold'
 ];
 
+const NUM_STANDARD_CFF_STRINGS = 391;
+
 var CFFParser = (function CFFParserClosure() {
   var CharstringValidationData = [
     null,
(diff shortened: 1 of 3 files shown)


src/core/cff_parser.js
```diff
@@ -1295,6 +1298,17 @@ var CFFCompiler = (function CFFCompilerClosure() {
       output.add(compiled.output);
       var topDictTracker = compiled.trackers[0];
 
+      // Create names for all of the glyphs so they can be put in the charset
+      // below.
+      let firstGlyphSid = cff.strings.strings.length + NUM_STANDARD_CFF_STRINGS;
```
Snuffleupagus: Nit: Assuming that `cff.strings` is an instance of `CFFStrings`, you should be able to use https://github.com/mozilla/pdf.js/blob/df7756d2a4a03831d836977ed47fa9e02f5a19c1/src/core/cff_parser.js#L945-L947 to simply write this as `let firstGlyphSid = cff.strings.count + NUM_STANDARD_CFF_STRINGS;`

Snuffleupagus (approved):
I'm assuming that testing this would be difficult in general, given that it's MacOS specific, but could e.g. a unit-test be added to avoid future regressions?

brendandahl:
> I'm assuming that testing this would be difficult in general, given that it's MacOS specific, but could e.g. a unit-test be added to avoid future regressions?

Yeah, even more difficult since this issue only seems to happen with printing on MacOS. Added a unit test.

timvandermeij:
@Snuffleupagus Could you perhaps re-check this now that changes have been made?

Snuffleupagus (approved):

src/core/cff_parser.js
```diff
@@ -1580,11 +1594,42 @@ var CFFCompiler = (function CFFCompilerClosure() {
...
+        let charsetIndex = 0;
+        let numCharsets = charset.charset.length;
+        for (let i = 1; i < out.length; i += 2) {
+          let sid = 0;
+          if (charsetIndex < numCharsets) {
+            let name = charset.charset[charsetIndex++];
+            sid = strings.getSID(name);
+            if (sid === -1) {
+              sid = 0;
+              warn(`Couldn't find ${name} in CFF strings`);
```
Snuffleupagus: @brendandahl This seem to have caused *a lot* of warning "spam" in the console even for the default `tracemonkey.pdf` file.
From a cursory look it appears that `charset.charset` only contains numbers, and not strings, for most PDF files.

PR #10597, opened 2019-02-28
Ensure that the temporary canvas created in `CanvasGraphics.isFontSubpixelAAEnabled` will be cleared

Snuffleupagus:
While this particular canvas may be small, there can still be an arbitrarily large number of them (one per page rendered), which can/will eventually add up memory wise. This can be easily avoided by using the `cachedCanvases` abstraction instead, which will ensure that the `isFontSubpixelAAEnabled` canvas is removed together with other temporary canvases in `CanvasGraphics.endDrawing`.


timvandermeij (approved):

timvandermeij:
Even though it might be small, every decrease in memory usage is good. Thanks!

PR #10601, opened 2019-03-01
Zero the width/height of the temporary canvas used during `TextLayer` rendering

Snuffleupagus:
The default size of these canvases seem to be `300 x 150` (two orders of magnitude larger than the ones in PR #10597), which probably is sufficient enough to matter since there's one such canvas for each textLayer that's rendered in the viewer.

Also fixes the incorrect rejection reason, i.e. one using a string rather than an `Error`, in the `TextLayerRenderTask.cancel` method.


timvandermeij (approved):

timvandermeij:
Nice find!

PR #10603, opened 2019-03-01
Change the `warn` to `info` in `CFFCompiler.compileCharset` to avoid loads of console "spam"

Snuffleupagus:
Note that even for something as simple as the default `tracemonkey.pdf` file, there's now *a lot* of warning messages printed.
Most likely the code actually needs to be adjusted to prevent a majority of these messages, but in the meantime it seems reasonable to use `info` rather than `warn` here.

/cc @brendandahl, see also https://github.com/mozilla/pdf.js/pull/10591#discussion_r261467368


brendandahl:
Interesting, I thought i tested that. I'll look at it again here in a bit.

PR #10604, opened 2019-03-01
Put the string name of the glyph in the charset array.

brendandahl:
Also, only warn once per font when missing a glyph name.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -3335,14 +3335,15 @@ var Type1Font = (function Type1FontClosure() {
       cff.globalSubrIndex = new CFFIndex();
 
       var count = glyphs.length;
-      var charsetArray = [0];
+      var charsetArray = ['.notdef'];
       var i, ii;
       for (i = 0; i < count; i++) {
-        var index = CFFStandardStrings.indexOf(charstrings[i].glyphName);
+        let glyphName = charstrings[i].glyphName;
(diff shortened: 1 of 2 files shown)


brendandahl:
@Snuffleupagus Can you review?
/botio test

src/core/fonts.js
```diff
@@ -3335,14 +3335,14 @@ var Type1Font = (function Type1FontClosure() {
       cff.globalSubrIndex = new CFFIndex();
 
       var count = glyphs.length;
-      var charsetArray = [0];
+      var charsetArray = ['.notdef'];
       var i, ii;
       for (i = 0; i < count; i++) {
         var index = CFFStandardStrings.indexOf(charstrings[i].glyphName);
```
Snuffleupagus: Nit: Would it (any) sense to define `let glyphName = charstrings[i].glyphName;` once at the top of the loop, rather than re-getting it potentially three times per iteration?

Snuffleupagus (approved):
Would it make sense to revert https://github.com/mozilla/pdf.js/blob/34022d2fd19caa98a56bc60c4e8ebde9eb7ea196/src/core/cff_parser.js#L956 to just `this.strings.push(value)` again, since nothing seems to be using that any more?

Thanks for the quick fix, looks good to me!

Snuffleupagus (approved):

PR #10605, opened 2019-03-01
Convert `let` to `const` if possible in, and improve unit test coverage for, `src/display/display_utils.js`

timvandermeij:
The commit messages include more details.

### DIFF
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -12,6 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* eslint no-var: error */
 
 import {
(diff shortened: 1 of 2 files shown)


src/display/display_utils.js
```diff
@@ -407,19 +407,19 @@ class StatTimer {
   }
 
   toString() {
-    let times = this.times;
+    const times = this.times;
```
Snuffleupagus: Completely optional nit: With the `for...of` loops below I'm not sure if this aliasing is necessary any more, also considering that the method isn't even called by default, so you could probably just remove this and simply format the two loops below as `for (const time of this.times) {` instead.

Snuffleupagus (approved):

PR #10608, opened 2019-03-02
Reduce usage of `Date.now()` in `src/core/worker.js`

Snuffleupagus:
Currently for every single parsed/rendered page there's no less than *four* `Date.now()` calls being made on the worker-side. This seems totally unnecessary, since the result of these calls are, by default, not used for anything *unless* the verbosity level is set to `INFO`.

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -14,10 +14,10 @@
  */
 
 import {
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #10610, opened 2019-03-03
Add type validation to the `default_preferences` generation (PR 10548 follow-up)

Snuffleupagus:
The generated `default_preferences.json` file is necessary when initializing the Firefox preferences, which only supports certain types, hence this patch adds additional validation to help prevent run-time errors in Firefox.

Given that these changes add a code-path to `AppOptions.getAll` which could throw, the `OptionKind.PREFERENCE` branch is also modified to require *exact* matching to prevent (future) errors in the viewer.

Finally the conditionally defined `defaultOptions` will no longer (potentially) be considered during the `gulp default_preferences` task, to make it more difficult for them to be accidentally included.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -535,7 +535,7 @@ gulp.task('default_preferences-pre', function() {
     ], { base: 'src/', }),
     gulp.src([
       'web/*.js',
-      '!web/{pdfjs,preferences,viewer}.js',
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #10613, opened 2019-03-04
Re-factor the `imageCache` used in `PartialEvaluator.getOperatorList`

Snuffleupagus:
This changes the `imageCache` from a simple object to a class, which will allow (future) support for re-sizing of large images.
Compared to my initial attempt(s) at implementing this, this patch purposely avoids trying to fix "everything" at once, since that required too large/invasive changes to the relevant code.

*Please note:* This patch, on its own, won't completely unblock PR #9255. However once this patch has landed, that PR can be re-factored on top of it and the new `ImageCache` class extended to properly support re-sized images.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -37,6 +37,7 @@ import {
   getSerifFonts, getStdFontMap, getSymbolsFonts
 } from './standard_fonts';
 import { getTilingPatternIR, Pattern } from './pattern';
+import { ImageCache, ImageCacheKind, NativeImageDecoder } from './image_utils';
 import { Lexer, Parser } from './parser';
 import { bidi } from './bidi';
 import { ColorSpace } from './colorspace';
@@ -47,7 +48,6 @@ import { getMetrics } from './metrics';
 import { isPDFFunction } from './function';
 import { JpegStream } from './jpeg_stream';
 import { MurmurHash3_64 } from './murmurhash3';
-import { NativeImageDecoder } from './image_utils';
 import { OperatorList } from './operator_list';
 import { PDFImage } from './image';
 
@@ -335,12 +335,16 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
         });
         imgData.cached = true;
         args = [imgData];
+
         operatorList.addOp(OPS.paintImageMaskXObject, args);
         if (cacheKey) {
-          imageCache[cacheKey] = {
+          imageCache.set({
+            key: cacheKey,
             fn: OPS.paintImageMaskXObject,
             args,
-          };
+            dimensions: { width: imgData.width, height: imgData.height,
+                          downsized: imgData.downsized, },
+          });
         }
         return Promise.resolve();
       }
@@ -388,10 +392,12 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
           operatorList.addOp(OPS.paintJpegXObject, args);
           if (cacheKey) {
-            imageCache[cacheKey] = {
+            imageCache.set({
+              key: cacheKey,
               fn: OPS.paintJpegXObject,
               args,
-            };
+              dimensions: { width: w, height: h, downsized: false, },
+            });
           }
         }, (reason) => {
           warn('Native JPEG decoding failed -- trying to recover: ' +
@@ -446,10 +452,12 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
       operatorList.addOp(OPS.paintImageXObject, args);
       if (cacheKey) {
-        imageCache[cacheKey] = {
+        imageCache.set({
+          key: cacheKey,
           fn: OPS.paintImageXObject,
           args,
-        };
+          dimensions: { width: w, height: h, downsized: false, },
+        });
       }
       return Promise.resolve();
     },
@@ -859,7 +867,10 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
       var self = this;
       var xref = this.xref;
-      var imageCache = Object.create(null);
+      const imageCache = new ImageCache({
+        handler: this.handler,
+        pageIndex: this.pageIndex,
+      });
 
       var xobjs = (resources.get('XObject') || Dict.empty);
       var patterns = (resources.get('Pattern') || Dict.empty);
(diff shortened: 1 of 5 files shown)


src/core/image_utils.js
```diff
@@ -0,0 +1,184 @@
...
+
+class ImageCache {
+  constructor({ handler, pageIndex, transform = null, }) {
+    this._handler = handler;
+    this._pageIndex = pageIndex;
+    this._transform = transform;
+    this._cache = Object.create(null);
+  }
+
+  get({ key, }) {
```
timvandermeij: If we only have a key, we don't need a parameter object here I think, unless you want to extend this later on?

src/core/image_utils.js
```diff
@@ -0,0 +1,184 @@
...
+  }
+}
+
+const ImageCacheKind = {
+  NON_EXISTING: 0,
+  EXISTING: 1,
+  NEEDS_RESIZING: 2,
+};
+
+class ImageCache {
```
timvandermeij: If would be good to have some unit tests for this new class. I think only the handler needs to be mocked for that.

timvandermeij (commented):
I'm seeing a bit of unused features in this code. I assume you'll extend this later, but I think it would help to add this to e.g., the commit message so we know what is to be updated later on, or preferably add it later when it actually becomes relevant. For example, `dimnesions`, `ctm` and `transform` are in `ImageCache`, but don't seem to be passed in/used inside the class anywhere. Perhaps I'm missing a bit of context here though; I think preparation work is fine (such as using parameter objects), but I'm not entirely sure about adding unused properties/features already when they will only be used later.

src/core/image_utils.js
Snuffleupagus: This may indeed be needed later, and I don't want to have to (unnecessarily) update the call-sites.

src/core/image_utils.js
Snuffleupagus: Sorry, but adding unit-tests would take time that I really don't have (and quite frankly don't want to spend *now*, since things may change subtly in the follow-up), especially since I still need to do additional work to actually wire-up everything in PR #9255.

Snuffleupagus: Also, I purposely adding lots of non-production/test-only `assert`s such that there *should* be a fair amount of implicit test-coverage already through the unit and reference tests.

Snuffleupagus:
> [...] but I think it would help to add this to e.g., the commit message so we know what is to be updated later on,

While I've got a basic idea of how it all will fit together in the end, I couldn't write this down without completely implementing the follow-up *first*. (Initially I tried to do "everything" at once, which just turned into a giant mess.)

> or preferably add it later when it actually becomes relevant.

The "later" here would be more-or-less immediately, since I should only need a free weekend to wire everything up for PR #9255 once the foundation is in place.

My apologies if this comes off as grumpy, but I've already spent a fair amount of time (and with more to come) essentially fixing someone else's PR here. Honestly, tagging the original issue as `5-good-beginner-bug` probably wasn't such a great idea in hindsight :-(

timvandermeij:
> My apologies if this comes off as grumpy, but I've already spent a fair amount of time (and with more to come) essentially fixing someone else's PR here.

I understand, and I think it's really nice of you to help out with getting that PR ready to merge! I'm fine with doing the unit tests in a follow-up since I can see it's easier to do that once the final code is in place. The assertions will do for now.

> I suppose the only thing I could possibly do here is to put this PR on hold until I've had time to actually implement the follow-up patch (on top of this PR and #9255), such that it's possible to provide a more complete picture.

I think that's a good idea. I also think it's easier to review once the code is complete too. Obviously we can still land this in these chunks.

Snuffleupagus:
> I suppose the only thing I could possibly do here is to put this PR on hold until I've had time to actually implement the follow-up patch (on top of this PR and #9255), such that it's possible to provide a more complete picture. (Everything should still land separately, of course.)

I've not yet gotten around to do this, on account of having a cold with a slight fever for the last couple of days, but hopefully will during the next weekend.

Also, a more general reason for all of this taking time is not only because of it being quite complex and involving both the main/worker-thread, but also me wanting to *try* to implement something that with a minimum of effort can be used if/when other kinds of image data will be downsized in the future.

timvandermeij:
> I've not yet gotten around to do this, on account of having a cold with a slight fever for the last couple of days, but hopefully will during the next weekend.

No problem; please take care of getting better first, that's much more important!

>  [...] to try to implement something that with a minimum of effort can be used if/when other kinds of image data will be downsized in the future.

That's actually a good idea as we have seen multiple attempts for other image kinds as well.

src/core/evaluator.js
```diff
@@ -335,12 +335,16 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
         });
         imgData.cached = true;
```
Snuffleupagus: Note for possible follow-up: This really doesn't make sense, since whether or not caching will actually happen depends on the existence of a `cacheKey`; should this be `imgData.cached = !!cacheKey; instead?

Snuffleupagus:
While continuing to work on this, I'm come across a few different pre-existing inefficiencies that have led to PRs such as #10644 and #10647, both of which are relevant to memory usage.

Currently I'm trying to come of with the best way of ensuring that *any* kind of downsized image data, no matter its type or how it's sent from the worker, will correctly cause the main-thread caches to be cleared such that re-rendering at a different scale won't break. At the moment, I'm still not sure exactly how/where to implement this.
Somewhat related to the above, but still a separate issue, is adding non-production/test-only `assert`s to ensure that certain image types won't ever be accidentally downsized in the future. At the moment, it's also not clear to me how best to implement this.

Snuffleupagus:
Sigh; after now having spent more time than I'd like to think about getting PR #9255 into shape for landing, I've now found that the approach in that PR is truly broken for many existing test-cases :-(

Unfortunately I made the mistake of not running all tests locally with *only* the patch from PR #9255 applied, before starting to hack on this. However, I'd assumed that PR #9255 was actually tested locally during development, but it seems pretty clear to me at this point that that was *not* actually the case :-(

The problem, which is *very* evident from the existing reference tests, is that many `ImageMask`s are simply downsized way too much and thus become blurry; for images containing "text" this is especially noticeable.

This is as far as I got: https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:ImageCache, which should work if it weren't for the fact that PR #9255 is broken; and honestly it makes no sense for me to sink even more time into this unless PR #9255 actually works and passes all tests *first*.

*This rant is probably in clear violation of the newly added code of conduct, but it's just so frustrating spending trying to fix something that was broken by design; sorry about this.*

timvandermeij:
I agree. To move forward, let's close this and first fix up the original PR in terms of functionality before spending more time on figuring out caching. Thank you for your efforts though; offering to help out with a "stuck" PR is really nice and in the process you did manage to find and fix some other issues related to this, which is really useful in general.

PR #10615, opened 2019-03-04
Handle corrupt ASCII85Decode inline images with whitespace "inside" of the EOD marker (issue 10614)

### LINKED ISSUE #10614: First Page of PDF Fails to Render - Blank Page
Configuration:
- Web browser and its version: Chrome 72.0.3626.119
- PDF.js version: https://mozilla.github.io/pdf.js/web/viewer.html as of 3/3/2019

1. [Download the PDF](https://github.com/mozilla/pdf.js/files/2927954/FOX.ALEXANDER.F.VS.FOX.BARBARA.E.2015-004684-FC-04.Doc-46-Memorandum-of-Law.Fla.11th.Cir.Ct.May.31.2016.pdf)
2. Go to web viewer demo: https://mozilla.github.io/pdf.js/web/viewer.html
3. Load PDF in 1

The first page renders fine in FoxIt (see screenshot 1), but fails on the webviewer (screenshot 2)

image

image

Snuffleupagus:
There's a number of things wrong with the PDF document, since its inline images are first all *a lot* larger than the 4 KB limit (as mandated by the specification, see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G7.1852045).

Furthermore the actual ASCII85Decode data is interspersed with *a lot* of needless whitespace, in particular also "inside" of the EOD (end-of-data) marker which thus completely breaks the detection.
Note that according to the specification, see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G6.1940130, this patch should be safe since it explicitly mentions that *all* whitespace should be ignored.

Fixes #10614.

### DIFF
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -336,14 +336,22 @@ var Parser = (function ParserClosure() {
      * Find the EOD (end-of-data) marker '~>' (i.e. TILDE + GT) of the stream.
      * @returns {number} The inline stream length.
      */
-    findASCII85DecodeInlineStreamEnd:
-        function Parser_findASCII85DecodeInlineStreamEnd(stream) {
+    findASCII85DecodeInlineStreamEnd(stream) {
       var TILDE = 0x7E, GT = 0x3E;
       var startPos = stream.pos, ch, length;
       while ((ch = stream.getByte()) !== -1) {
-        if (ch === TILDE && stream.peekByte() === GT) {
-          stream.skip();
-          break;
+        if (ch === TILDE) {
+          ch = stream.peekByte();
+          // Handle corrupt PDF documents which contains whitespace "inside" of
+          // the EOD marker (fixes issue10614.pdf).
+          while (isSpace(ch)) {
+            stream.skip();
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
> Note that according to the specification, see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G6.1940130, this patch should be safe since it explicitly mentions that _all_ whitespace should be ignored.

Note also the `Ascii85Stream` decoding in https://github.com/mozilla/pdf.js/blob/e1b01a601c7ff4c858d36f1b9277a287badacc1e/src/core/stream.js#L963-L965 and https://github.com/mozilla/pdf.js/blob/e1b01a601c7ff4c858d36f1b9277a287badacc1e/src/core/stream.js#L987-L989

timvandermeij (approved):

timvandermeij:
Thank you for finding and fixing this bug!

PR #10628, opened 2019-03-08
Try to improve text-selection for Type3 fonts that utilize a non-default /FontMatrix (bug 1513120)

Snuffleupagus:
For Type3 fonts text-selection is often not that great, and there's a couple of heuristics used to try and improve things. This patch simple extends those heuristics a bit, and fixes a pre-existing "naive" array comparison, but this all feels a bit brittle to say the least.

The existing Type3 test-coverage isn't that great in general, and in particular Type3 `text` tests are few and far between, hence why this patch adds *two* different new `text` tests.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1513120

/cc @brendandahl


src/shared/util.js
```diff
@@ -824,6 +824,15 @@ function isArrayBuffer(v) {
   return typeof v === 'object' && v !== null && v.byteLength !== undefined;
 }
 
+function isArraysEqual(arr1, arr2) {
```
brendandahl: Typo with the extra 's'?

brendandahl (approved):
r+ pending the removal of the 's'

PR #10633, opened 2019-03-09
Convert `MurmurHash3_64` to an ES6 class

Snuffleupagus:
Notable changes:
 - Remove the `return this;` from the `MurmurHash3_64.update` method, since it's completely unused and doesn't make a lot of sense.
 - Remove the loop(s) from the `MurmurHash3_64.hexdigest` method, since creating a temporary array and then looping over it is wasteful given how simple this can be written with modern JavaScript.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #10634, opened 2019-03-09
Move `NativeImageDecoder` into a separate file, and convert it to a `class`

Snuffleupagus:
Given the size of the `src/core/evaluator.js` file, it cannot hurt to move some of its (image related) helper functionality into a separate file.

Split from PR #10613, since I think this makes sense in general.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #10635, opened 2019-03-10
Convert `src/core/parser.js` to ES6 syntax and write more unit tests for the lexer and the parser

timvandermeij:
I recommend to review each commit with the `?w=1` flag to reduce the size of the diff.

### DIFF
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -34,7 +34,7 @@ const MAX_LENGTH_TO_CACHE = 1000;
 const MAX_ADLER32_LENGTH = 5552;
 
 function computeAdler32(bytes) {
-  let bytesLength = bytes.length;
+  const bytesLength = bytes.length;
   if (typeof PDFJSDev === 'undefined' ||
       PDFJSDev.test('!PRODUCTION || TESTING')) {
     assert(bytesLength < MAX_ADLER32_LENGTH,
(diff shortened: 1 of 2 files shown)


src/core/parser.js
```diff
@@ -1172,55 +1173,58 @@ class Lexer {
...
-      var hints = linDict.get('H'), hintsLength, item;
-      if (Array.isArray(hints) &&
-          ((hintsLength = hints.length) === 2 || hintsLength === 4)) {
-        for (var index = 0; index < hintsLength; index++) {
-          if (!(Number.isInteger(item = hints[index]) && item > 0)) {
-            throw new Error('Hint (' + index +
-                            ') in the linearization dictionary is invalid.');
+
+    function getHints(linDict) {
+      const hints = linDict.get('H'), hintsLength = hints.length;
```
Snuffleupagus: Drive-by-comment: Please don't move `hintsLength = hints.length;` here, since that will result in errors being thrown if `hints` isn't actually an Array (which is why the old code checked that *first*).

test/unit/parser_spec.js
```diff
@@ -14,110 +14,182 @@
...
-  describe('Lexer', function() {
-    it('should stop parsing numbers at the end of stream', function() {
-      const input = new StringStream('11.234');
-      const lexer = new Lexer(input);
-      expect(lexer.getNumber()).toEqual(11.234);
-    });
+  describe('Parser', function() {
+    describe('inlineStreamSkipEI', function() {
+      it('should skip over the EI marker if it is found', function() {
+        const string = 'BI /foo /bar EI<<';
```
Snuffleupagus: While it may not matter for these particular tests, please note that these commands do *not* represent a valid content stream :-P

Unless you want to test for a *specific* mode of corruption, as found in real-world PDFs, it's always best if the unit-test is actually correct in my experience; please find an actual example below (from 10388_reduced.pdf). 
```
BI 
/W 16
/H 17
/BPC 1
/IM true 
ID ÿÿÿÿþþþþþþÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿÿ
EI
```

src/core/parser.js
```diff
@@ -49,662 +50,705 @@ function computeAdler32(bytes) {
   return ((b % 65521) << 16) | (a % 65521);
 }
 
-var Parser = (function ParserClosure() {
-  function Parser(lexer, allowStreams, xref, recoveryMode) {
+class Parser {
+  constructor(lexer, allowStreams, xref, recoveryMode = false) {
```
Snuffleupagus: Another possible improvement here, in a separate commit or PR, would be to change the constructor to take an object instead to get rid of the often "ugly" call-sites (and document them for free):
```javascript
constructor({ lexer, xref, allowStreams = false, recoveryMode = false, }) {
```

src/core/parser.js
timvandermeij: Whoops, good point! I'll adjust that.

test/unit/parser_spec.js
timvandermeij: That's a fair point. In this case it didn't really matter, but I agree that it's probably better to use real data instead of fake data here.

src/core/parser.js
timvandermeij: I agree, I'll see what I can do.

test/unit/parser_spec.js
Snuffleupagus: *Apparently I wrote that code, so here a bit of unsolicited advice :-)*

As the comment mentions this method is only used for inline image streams with an EOD (end-of-data) marker, i.e. `DCTDecode` (== JPEG), `ASCII85Decode`, and `ASCIIHexDecode`; see https://github.com/mozilla/pdf.js/blob/d587abbceb97008ed744261c219c90d3ce981f2b/src/core/parser.js#L390-L393
When this method is called, the EOD marker in question has been successfully found and we simply want to advance the stream position just after the `EI` command (ignoring any white-space in between).

The example above wasn't that great in hindsight, since in practice it wouldn't apply here (no /Filter entry, and thus no EOD marker present).
The following is unlikely to be actually valid image data, so please keep that in mind, but let's consider the following example instead:
```
q
1 0 0 1 0 0 cm 
BI 
/W 10
/H 10
/F /A85 
ID abc123~>
EI
Q
```
When the main part of `findASCII85DecodeInlineStreamEnd` has run, the stream position is now just after `~>`. With the help of `inlineStreamSkipEI` the stream position will be advanced to just after `EI`, which is what you need here.

src/core/parser.js
timvandermeij: Fixed in the new commit series.

test/unit/parser_spec.js
timvandermeij: Fixed in the new commit series. The unit test cases are now more like what we encounter in real streams. I couldn't find a nice way to put binary image data in there and I think the mock value `abc123` is fine since our only purpose is to show that we can skip over the `EI` marker, so the actual binary data seems irrelevant here.

src/core/parser.js
timvandermeij: Since this patch is already quite big, I'd like to leave this for a possible follow-up.

Snuffleupagus: > Since this patch is already quite big, I'd like to leave this for a possible follow-up.

That's completely fine; just as long as it eventually happens, given how unwieldy the call-sites currently look :-)

Snuffleupagus (approved):

PR #10644, opened 2019-03-14
Ensure that `blob:` URLs will be revoked when pages are cleaned-up/destroyed (JPEG memory usage)

Snuffleupagus:
- Ensure that `blob:` URLs will be revoked when pages are cleaned-up/destroyed

   Natively supported JPEG images are sent as-is, using a `blob:` or possibly a `data` URL, to the main-thread for loading/decoding.
However there's currently no attempt at releasing these resources, which are held alive by `blob:` URLs, which seems unfortunately given that images can be arbitrarily large.

   As mentioned in https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL the lifetime of these URLs are tied to the document, hence they are not being removed when a page is cleaned-up/destroyed (e.g. when being removed from the `PDFPageViewBuffer` in the viewer).
This is easy to test with the help of `about:memory` (in Firefox), which clearly shows the number of `blob:` URLs becomming arbitrarily large *without* this patch. With this patch however the `blob:` URLs are immediately release upon clean-up as expected, and the memory consumption should thus be considerably reduced for long documents with (simple) JPEG images.

 - Remove the `src` attribute from `Image` objects used with natively supported JPEG images, when pages are cleaned-up/destroyed

   This will further help reduce the amount of image data that's currently being held alive, by explicitly removing the `src` attribute.

   Please note that this is mostly relevant for browsers which do not support `URL.createObjectURL`, or where `disableCreateObjectURL` was manually set by the user, since `blob:` URLs will be revoked (see the previous patch).
However, using `about:memory` (in Firefox) it does seem that this may also be generally helpful, given that calling `URL.revokeObjectURL` won't invalidate the image data itself (as far as I can tell).

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -23,7 +23,8 @@ import {
 } from '../shared/util';
 import {
   deprecated, DOMCanvasFactory, DOMCMapReaderFactory, DummyStatTimer,
-  loadScript, PageViewport, RenderingCancelledException, StatTimer
+  loadScript, PageViewport, releaseImageResources, RenderingCancelledException,
+  StatTimer
 } from './display_utils';
 import { FontFaceObject, FontLoader } from './font_loader';
 import { apiCompatibilityParams } from './api_compatibility';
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
This is a really nice find; thanks!

timvandermeij:
/botio makeref

(to fix reference test failures)

PR #10646, opened 2019-03-16
Implement linear-gradient, radial-gradient and dummy-pattern in SVGGraphics.

terurou:


src/display/svg.js
```diff
@@ -936,6 +946,91 @@ SVGGraphics = (function SVGGraphicsClosure() {
       this.current.tspan = this.svgFactory.createElement('svg:tspan');
       this.current.xcoords = [];
     },
+    setStrokeColorN: function SVGGraphics_setStrokeColorN(args) {
+      this.endPathcurrent.strokeColor = this._makeColorN_Pattern(args);
```
timvandermeij: Shouldn't this be `this.current.strokeColor`?

terurou: Thank you! I fixed this mistake.

src/display/svg.js
```diff
@@ -936,6 +946,91 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
+        return null;
+      }
+      return this._makeShadingPattern(args);
+    },
+    _makeShadingPattern: function SVGGraphics_makeShadingPattern(args) {
+      switch (args[0]) {
+        case 'RadialAxial':
+          var shadingId = 'shading' + shadingCount++;
+          var colorStops = args[2];
+          var gradient = this.svgFactory.createElement('svg:linearGradient');
```
timvandermeij: In the case of a radial gradient, I think we should use `svg:radialGradient` instead. According to https://developer.mozilla.org/en-US/docs/Web/SVG/Element/linearGradient, there are no `cx` and `cy` attributes, whereas `radialGradient` has those.

terurou: Thanks very much. I fixed it.
Maybe I mistaked when I refactored for the eslint warnings...

timvandermeij (approved):

timvandermeij:

Snuffleupagus:
@terurou Please remember to actually squash fix-up commits in the future, as was asked for above, since the result here was that this PR landed with broken intermediate state meaning that e.g. future bisecting will now be *a lot* more difficult.

timvandermeij:
You're right. I also forgot to choose the "squash and commit" option on GitHub.

terurou:
@Snuffleupagus Thank you for your advice. I will be careful from the next time.

PR #10653, opened 2019-03-19
Feature(Search): Adding Support for Searching w/ Index

RAMcInnes:
Currently, PDF.js does not support searching a document with an index. So if I wanted to search for the 3rd instance of the word "happy", I can't. This code change adds the parameter "doc_index" to allow for that. Now if a search term and an index is applied, the correct instance of the term will be highlighted.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1354,6 +1354,7 @@ let PDFViewerApplication = {
     eventBus.on('documentproperties', webViewerDocumentProperties);
     eventBus.on('find', webViewerFind);
     eventBus.on('findfromurlhash', webViewerFindFromUrlHash);
+    eventBus.on('findsearchindexhash', webViewerFindSearchIndexHash);
     eventBus.on('updatefindmatchescount', webViewerUpdateFindMatchesCount);
     eventBus.on('updatefindcontrolstate', webViewerUpdateFindControlState);
     if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
Personally I'm not convinced that this belongs in the PDF.js repository, as opposed to a custom implementation, for a number of reasons:
 - This functionality seems like enough of a special case to not warrant the added complexity in the `PDFFindController` (and related code); see also below.
 - This adds a non-standard hash parameter, which is something that we've always tried to avoid (which includes removing old non-standard ones from the PDF.js viewer).
 - This implementation will perform *really* poorly in long document, e.g. those with thousands of pages, since it requires waiting for all pages before this "special" search operation completes.
*Generally speaking:* There's not really any good way of implementing something like this given the asynchronous nature of the PDF.js API/viewer.

   Furthermore, as-is the patch will also cause unexpected behaviour/glitches in the viewer, since there's no attempt at "stopping" these potentially long-running search operation when e.g. a new search occurs.

 - This patch also includes a fair amount of unrelated code, which I can only assume is used in a custom implementation, but which unfortunately cannot be accepted in this repository.

timvandermeij:
I agree with the remarks mentioned above. I also don't really see this as a general use case, nor do we want to add non-standard parameters to the viewer.

RAMcInnes:
I am currently using this code in a custom implementation. Wasn't sure if it could be helpful for general use. Appreciate the feedback!

RAMcInnes:
@Snuffleupagus I have one or two other ideas I would like to propose for PDF.js. Is there a "list of standards" or a "proper way" to propose ideas? I don't mind creating other Pull Requests for you to look at, but if there is a better way, I would happy to follow it. Thanks for your help.

PR #10657, opened 2019-03-20
Update test case description

ThomasdenH:
In b1472cddbb07692089371dd041ad8b31717f6826, an evaluator test case changed, but its description was not updated accordingly.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #10668, opened 2019-03-22
Remove the Firefox-specific 'read with streaming' unit-test

### LINKED ISSUE #8851: Enable "read with streaming" test in test/unit/network_spec.js

Snuffleupagus:
Support for the non-standard `moz-chunked-arraybuffer` response type is in the process of being removed from Firefox; see e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1411865

For the time being, you probably want to keep support for this in the general PDF.js library given that feature detection is used. However, removing the unit-test immediately seems reasonable, since it will otherwise start failing once the platform support for `moz-chunked-arraybuffer` is gone.

Fixes #8851; please note that if unit-tests for the code in `fetch_stream.js` are wanted, which I'm assuming they are, those should live in their own file rather than being lumped into `network_spec.js` anyway.

### DIFF
--- a/test/unit/network_spec.js
+++ b/test/unit/network_spec.js
@@ -18,8 +18,6 @@ import { PDFNetworkStream } from '../../src/display/network';
 describe('network', function() {
   var pdf1 = new URL('../pdfs/tracemonkey.pdf', window.location).href;
   var pdf1Length = 1016315;
-  var pdf2 = new URL('../pdfs/pdf.pdf', window.location).href;
-  var pdf2Length = 32472771;
 
   it('read without stream and range', function(done) {
     var stream = new PDFNetworkStream({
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):
I'll make a follow-up ticket for introducing proper unit tests for streaming again. Thanks!

Issue #9702, opened 2018-05-05
Photon design


nt1m:
It would be nice if PDF.js's design would be in line with Firefox's.

To do that, here are some quick changes that would be possible:
* Remove background gradients/textures
* Use photon colors: https://design.firefox.com/photon/visuals/color.html#grey (use visual layering to pick shades of gray: https://mozilla.invisionapp.com/share/J6GQCZR2PZY#/screens/289803032_Desktop_Common)
* Use photon icons: https://design.firefox.com/icons

Huluti:
It's true that the reader looks very outdated... compared to the browser

timvandermeij:
Fixed by #11077.

PR #10673, opened 2019-03-23
Improve UI design (Change some colors)

shge:
Since the colors and gradations were a little outdated, I made some changes to the UI design. 😃

### DIFF
--- a/web/text_layer_builder.css
+++ b/web/text_layer_builder.css
@@ -20,7 +20,7 @@
   right: 0;
   bottom: 0;
   overflow: hidden;
-  opacity: 0.2;
+  opacity: 0.5;
   line-height: 1.0;
 }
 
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
Just making the entire default viewer, which is shipping in Firefox, essentially black really isn't an improvement at all. There's a number of issues such as e.g. the main toolbar being indistinguishable from the rest of the viewer, and generally a fairly incomplete attempt at changing the styling in any coherent way.

For a custom implementation you're obviously free to do whatever you want, but I don't think this is something that belongs in the main PDF.js library.

shge:
@Snuffleupagus 
Okay, I will change only selection colors.

Snuffleupagus:
> Okay, I will change only selection colors.

That's already tracked in issue #9184, and it's not really clear if/how any changes should actually be made here (since there's a degree of subjectivity involved).
As for these particular changes, they are not an improvement since for one they make the actual text *less* clear when it's selected (probably caused by the change in opacity).

timvandermeij:
I agree with the above. The only thing we may consider is #9702, but the designers from Firefox should come up with a design first.

PR #10674, opened 2019-03-23
Convert `src/display/svg.js` to ES6 syntax and implement `setRenderingIntent` and `setFlatness` for the SVG backend 

timvandermeij:

The objective is to modernize the SVG backend and bring it more in line with the canvas back-end. Once this is done, it becomes easier to include the open pull requests for the SVG backend as well as working towards implementing reference testing for the SVG backend. The latter is now difficult because the canvas and SVG backend don't share a common interface. After this patch, we can start to extract a common interface and have both backends implement that so they can be interchanged more easily. After that, the remaining drawing code and the test driver need to be made backend-agnostic.

In short, this pull request is the first step towards this goal (part of the open "SVG backend" project).

*I recommend to use the `?w=1` flag to considerably reduce the size of the diff.*

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -432,6 +432,14 @@ SVGGraphics = class SVGGraphics {
     this.embeddedFonts = Object.create(null);
     this.cssStyle = null;
     this.forceDataSchema = !!forceDataSchema;
+
+    // In `src/shared/util.js` the operator names are mapped to IDs.
+    // The list below represents the reverse of that, i.e., it maps IDs
+    // to operator names.
+    this._operatorIdMapping = [];
+    for (const op in OPS) {
+      this._operatorIdMapping[OPS[op]] = op;
+    }
   }
 
   save() {
@@ -493,11 +501,7 @@ SVGGraphics = class SVGGraphics {
   }
 
   convertOpList(operatorList) {
-    const REVOPS = [];
-    for (const op in OPS) {
-      REVOPS[OPS[op]] = op;
-    }
-
+    const operatorIdMapping = this._operatorIdMapping;
     const argsArray = operatorList.argsArray;
     const fnArray = operatorList.fnArray;
     const opList = [];
@@ -505,7 +509,7 @@ SVGGraphics = class SVGGraphics {
       const fnId = fnArray[i];
       opList.push({
         'fnId': fnId,
-        'fn': REVOPS[fnId],
+        'fn': operatorIdMapping[fnId],
         'args': argsArray[i],
       });
     }
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -595,6 +595,12 @@ SVGGraphics = class SVGGraphics {
         case OPS.setDash:
           this.setDash(args[0], args[1]);
           break;
+        case OPS.setRenderingIntent:
+          this.setRenderingIntent(args[0]);
+          break;
+        case OPS.setFlatness:
+          this.setFlatness(args[0]);
+          break;
         case OPS.setGState:
           this.setGState(args[0]);
           break;
@@ -1188,6 +1194,14 @@ SVGGraphics = class SVGGraphics {
     this.current.textHScale = scale / 100;
   }
 
+  setRenderingIntent(intent) {
+    // This operation is ignored since we haven't found a use case for it yet.
+  }
+
+  setFlatness(flatness) {
+    // This operation is ignored since we haven't found a use case for it yet.
+  }
+
   setGState(states) {
     for (const [key, value] of states) {
       switch (key) {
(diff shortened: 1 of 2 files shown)


src/display/svg.js
```diff
@@ -321,91 +322,101 @@ var SVGExtraState = (function SVGExtraStateClosure() {
...
 
-  /**
-   * Formats float number.
-   * @param value {number} number to format.
-   * @returns {string}
-   */
-  function pf(value) {
-    if (Number.isInteger(value)) {
-      return value.toString();
+// eslint-disable-next-line no-inner-declarations
```
timvandermeij: This is required (here and below) because these are now declared inside an `if` statement inside of in a closure like before (the `if` statement being the one that excludes all SVG code from non-generic builds).

timvandermeij:
@Snuffleupagus Would you perhaps, time permitting, be willing to review this?

Snuffleupagus:
> Would you perhaps, time permitting, be willing to review this?

I don't really have any experience working with the SVG code, so unfortunately I don't think I'll have time until the next weekend; but if you don't mind waiting I'll see what I can do :-)

timvandermeij:
No problem, and thanks!

src/display/svg.js
```diff
@@ -423,1061 +434,1045 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
-            break;
-          case 92:
-            this.group(opTree[x].items);
-            break;
-          default:
-            warn('Unimplemented operator ' + fn);
-            break;
+
+      for (const obj of argsArray[i]) {
+        const common = obj.substring(0, 2) === 'g_';
```
Snuffleupagus: Why not just use `obj.startsWith('g_')` instead here?
Also, the existing code below could probably be simplified a bit more and duplication removed; e.g. something like this:
```javascript
for (const obj of argsArray[i]) {
  const objsPool = obj.startsWith('g_') ? this.commonObjs : this.objs;

  const promise = new Promise((resolve) => {
    objsPool.get(obj, resolve);
  }
  this.current.dependencies.push(promise);
}
```

src/display/svg.js
```diff
@@ -423,1061 +434,1045 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
+    return this.loadDependencies(operatorList).then(() => {
+      this.transformMatrix = IDENTITY_MATRIX;
+      this.executeOpTree(this.convertOpList(operatorList));
+      return svgElement;
+    });
+  }
+
+  convertOpList(operatorList) {
+    const REVOPS = [];
+    for (const op in OPS) {
```
Snuffleupagus: Would it make sense to compute this *once* globally instead, since I'm assuming that this code is being run for every single page as-is?

src/display/svg.js
```diff
@@ -423,1061 +434,1045 @@ SVGGraphics = (function SVGGraphicsClosure() {
...
+  setTextRenderingMode(textRenderingMode) {
+    this.current.textRenderingMode = textRenderingMode;
+  }
+
+  setHScale(scale) {
+    this.current.textHScale = scale / 100;
+  }
+
+  setGState(states) {
+    for (const state of states) {
```
Snuffleupagus: Unless you actually need the `state` variable below, you could probably just do `for (const [key, value] of states) {` instead.

Snuffleupagus (approved):

src/display/svg.js
```diff
@@ -321,91 +322,101 @@ var SVGExtraState = (function SVGExtraStateClosure() {
...
-    if (Number.isInteger(value)) {
-      return value.toString();
+// eslint-disable-next-line no-inner-declarations
+function opListToTree(opList) {
+  let opTree = [];
+  const tmp = [];
+
+  for (const opListElement of opList) {
+    if (opListElement.fn === 'save') {
+      opTree.push({ 'fnId': 92, 'fn': 'group', 'items': [], });
```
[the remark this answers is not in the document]
timvandermeij: Yes, in fact that's part of the follow-up after this patch since it's already included in https://github.com/mozilla/pdf.js/pull/10475/files and the objective is to rebase that after this lands.

src/display/svg.js
timvandermeij: The objective is to remove the `if` statement around all SVG code later on if the SVG back-end is more stable. The only reason it's excluded from non-generic builds is because it was experimental and not yet used.

src/display/svg.js
timvandermeij: Good point. Fixed in the new commit.

src/display/svg.js
timvandermeij: Yes, that does sound like a good idea. Fixed in the new commit.

src/display/svg.js
timvandermeij: Fixed in the new commit.

PR #10675, opened 2019-03-24
[Firefox regression] Fix `disableRange=true` bug in `PDFDataTransportStream`

Snuffleupagus:
Currently if trying to set `disableRange=true` in the built-in PDF Viewer in Firefox, either through `about:config` or via the URL hash, the PDF document will never load. It appears that this has been broken for a couple of years, without anyone noticing.

Obviously it's not a good idea to set `disableRange=true`, however it seems that this bug affects the PDF Viewer in Firefox even with default settings:
 - In the case where `initialData` already contains the *entire* file, we're forced to dispatch a range request to re-fetch already available data just so that file loading may complete.
 - (In the case where the data arrives, via streaming, before being specifically requested through `requestDataRange`, we're also forced to re-fetch data unnecessarily.) *This part was removed, to reduce the scope/risk of the patch somewhat.*

In the cases outlined above, we're having to re-fetch already available data thus potentially delaying loading/rendering of PDF files in Firefox (and wasting resources in the process).

/cc @yurydelendik Could you please, time permitting of course, check that this patch makes sense?

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -344,6 +344,7 @@ function getDocument(src) {
         networkStream = new PDFDataTransportStream({
           length: params.length,
           initialData: params.initialData,
+          progressiveDone: params.progressiveDone,
           disableRange: params.disableRange,
           disableStream: params.disableStream,
         }, rangeTransport);
@@ -389,6 +390,7 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
   if (pdfDataRangeTransport) {
     source.length = pdfDataRangeTransport.length;
     source.initialData = pdfDataRangeTransport.initialData;
+    source.progressiveDone = pdfDataRangeTransport.progressiveDone;
   }
   return worker.messageHandler.sendWithPromise('GetDocRequest', {
     docId,
@@ -515,13 +517,15 @@ const PDFDocumentLoadingTask = (function PDFDocumentLoadingTaskClosure() {
  * @param {Uint8Array} initialData
  */
 class PDFDataRangeTransport {
-  constructor(length, initialData) {
+  constructor(length, initialData, progressiveDone = false) {
     this.length = length;
     this.initialData = initialData;
+    this.progressiveDone = progressiveDone;
 
     this._rangeListeners = [];
     this._progressListeners = [];
     this._progressiveReadListeners = [];
+    this._progressiveDoneListeners = [];
     this._readyCapability = createPromiseCapability();
   }
 
@@ -535,6 +539,10 @@ class PDFDataRangeTransport {
 
   addProgressiveReadListener(listener) {
     this._progressiveReadListeners.push(listener);
+  }
+
+  addProgressiveDoneListener(listener) {
+    this._progressiveDoneListeners.push(listener);
   }
 
   onDataRange(begin, chunk) {
@@ -555,6 +563,14 @@ class PDFDataRangeTransport {
     this._readyCapability.promise.then(() => {
       for (const listener of this._progressiveReadListeners) {
         listener(chunk);
+      }
+    });
+  }
+
+  onDataProgressiveDone() {
+    this._readyCapability.promise.then(() => {
+      for (const listener of this._progressiveDoneListeners) {
+        listener();
       }
     });
   }
--- a/src/display/transport_stream.js
+++ b/src/display/transport_stream.js
@@ -21,7 +21,9 @@ var PDFDataTransportStream = (function PDFDataTransportStreamClosure() {
     assert(pdfDataRangeTransport);
 
     this._queuedChunks = [];
-    var initialData = params.initialData;
+    this._progressiveDone = params.progressiveDone || false;
+
+    const initialData = params.initialData;
     if (initialData && initialData.length > 0) {
       let buffer = new Uint8Array(initialData).buffer;
       this._queuedChunks.push(buffer);
@@ -45,6 +47,10 @@ var PDFDataTransportStream = (function PDFDataTransportStreamClosure() {
 
     this._pdfDataRangeTransport.addProgressiveReadListener((chunk) => {
       this._onReceiveData({ chunk, });
+    });
+
+    this._pdfDataRangeTransport.addProgressiveDoneListener(() => {
+      this._onProgressiveDone();
     });
 
     this._pdfDataRangeTransport.transportReady();
@@ -80,6 +86,13 @@ var PDFDataTransportStream = (function PDFDataTransportStreamClosure() {
        }
     },
 
+    _onProgressiveDone() {
+      if (this._fullRequestReader) {
+        this._fullRequestReader.progressiveDone();
+      }
+      this._progressiveDone = true;
+    },
+
     _removeRangeReader:
         function PDFDataTransportStream_removeRangeReader(reader) {
       var i = this._rangeReaders.indexOf(reader);
@@ -92,7 +105,8 @@ var PDFDataTransportStream = (function PDFDataTransportStreamClosure() {
       assert(!this._fullRequestReader);
       var queuedChunks = this._queuedChunks;
       this._queuedChunks = null;
-      return new PDFDataTransportStreamReader(this, queuedChunks);
+      return new PDFDataTransportStreamReader(this, queuedChunks,
+                                              this._progressiveDone);
     },
 
     getRangeReader: function PDFDataTransportStream_getRangeReader(begin, end) {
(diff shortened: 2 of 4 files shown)


Snuffleupagus:
Apparently I'm not really understanding this code well enough to fix this, since one of the existing unit-tests started failing now :-(

https://github.com/mozilla/pdf.js/blob/9b5a937f78418ad6b114ca66d062cf2ae416d05f/test/unit/api_spec.js#L1496-L1527

yurydelendik (approved):
Really good. Thank you for the patch.

yurydelendik:
Actually let me think more about this PR. As I recall, we cannot trust _contentLength.

yurydelendik:
I spy https://github.com/mozilla/pdf.js/blob/89da03f4ee7087b8f8799b52f41a682275e34610/src/core/chunked_stream.js#L116 . It means if file is partially loaded, the range request is ignored. I wonder if we can do the same at the display layer, in case if it is just a async timing issue between data arriving to core layer and display one.

Snuffleupagus:
> Actually let me think more about this PR. As I recall, we cannot trust _contentLength.

Excellent point about `contentLength` not being trusted, I should have thought about that myself!

So the current patch obviously won't cut it, as far as solutions go, but I'm wondering if a possible fix would be to dispatch a message (e.g. "progressiveComplete") from https://searchfox.org/mozilla-central/rev/56705678f5fc363be5e0237e1686f619b0d23009/browser/extensions/pdfjs/content/PdfStreamConverter.jsm#626-628 which is then passed to `PDFDataTransportStream` such that the `PDFDataTransportStreamReader._done` property can be set?

> It means if file is partially loaded, the range request is ignored. I wonder if we can do the same at the display layer, in case if it is just a async timing issue between data arriving to core layer and display one.

Good find, that looks very much relevant here as well. Implementing something similar for `PDFDataTransportStream` doesn't look too difficult (famous last words) so I can try that later.

Snuffleupagus:
I've tried to quickly hack something together as outlined in https://github.com/mozilla/pdf.js/pull/10675#issuecomment-476223116, and I *think* it might work, however what I've got currently is too messy to submit a patch for consideration just yet.

---

A related problem appears to be that the `ChunkedStreamManager` is triggering range requests *irrespective* of them actually being supported or not, see https://github.com/mozilla/pdf.js/blob/9b5a937f78418ad6b114ca66d062cf2ae416d05f/src/core/chunked_stream.js#L286-L291
@yurydelendik Why aren't we checking `fullRequestReader.isRangeSupported`, or something, *before* initializing a new `PDFWorkerStreamRangeReader` instance there?
When range requests aren't supported it really doesn't make much sense, at least to me, to even attempt to dispatch them. Instead, shouldn't you simply wait for the streamed data to arrive "naturally" instead!?

It's starting to feel that I'm overlooking something simple here, but there's a lot of code to wrap your head around; any pointers will be most helpful.

Snuffleupagus:
I've tried to quickly hack something together as outlined in https://github.com/mozilla/pdf.js/pull/10675#issuecomment-476223116, and I *think* it might work, however what I've got currently is too messy to submit a patch for consideration just yet.

Snuffleupagus:
The updated patch seems to work both with `disableRange=true` in Firefox *and* with the unit-tests, and doesn't require outright trusting the `contentLength`. Note that it will also require a few minor, in my opinion at least, changes to the code in `mozilla-central`; please see the diff below.

@yurydelendik Thanks for helping out here; does this look like a more viable approach to you?
Also, the ideas from https://github.com/mozilla/pdf.js/pull/10675#issuecomment-476184811 are now deferred for a later patch, since I didn't want to touch too much code and thus risk regressing file loading in general.

```diff
diff --git a/browser/extensions/pdfjs/content/PdfStreamConverter.jsm b/browser/extensions/pdfjs/content/PdfStreamConverter.jsm
--- a/browser/extensions/pdfjs/content/PdfStreamConverter.jsm
+++ b/browser/extensions/pdfjs/content/PdfStreamConverter.jsm
@@ -168,16 +168,19 @@ PdfDataListener.prototype = {
     let writeOffset = 0;
     while (this.buffers.length) {
       let buffer = this.buffers.shift();
       combinedArray.set(buffer, writeOffset);
       writeOffset += buffer.length;
     }
     return combinedArray;
   },
+  get isDone() {
+    return !!this.isDataReady;
+  },
   finish: function PdfDataListener_finish() {
     this.isDataReady = true;
     if (this.oncompleteCallback) {
       this.oncompleteCallback(this.readData());
     }
   },
   error: function PdfDataListener_error(errorCode) {
     this.errorCode = errorCode;
@@ -601,45 +604,53 @@ class RangedChromeActions extends Chrome
     // requests, which we need to abort when we leave the page
     domWindow.addEventListener("unload", function unload(e) {
       domWindow.removeEventListener(e.type, unload);
       self.abortLoading();
     });
   }
 
   initPassiveLoading() {
-    var data;
+    let data, done;
     if (!this.streamingEnabled) {
       this.originalRequest.cancel(Cr.NS_BINDING_ABORTED);
       this.originalRequest = null;
       data = this.dataListener.readData();
+      done = this.dataListener.isDone;
       this.dataListener = null;
     } else {
       data = this.dataListener.readData();
+      done = this.dataListener.isDone;
 
       this.dataListener.onprogress = (loaded, total) => {
         this.domWindow.postMessage({
           pdfjsLoadAction: "progressiveRead",
           loaded,
           total,
           chunk: this.dataListener.readData(),
         }, PDF_VIEWER_ORIGIN);
       };
       this.dataListener.oncomplete = () => {
+        if (!done && this.dataListener.isDone) {
+          this.domWindow.postMessage({
+            pdfjsLoadAction: "progressiveDone",
+          }, PDF_VIEWER_ORIGIN);
+        }
         this.dataListener = null;
       };
     }
 
     this.domWindow.postMessage({
       pdfjsLoadAction: "supportsRangedLoading",
       rangeEnabled: this.rangeEnabled,
       streamingEnabled: this.streamingEnabled,
       pdfUrl: this.pdfUrl,
       length: this.contentLength,
       data,
+      done,
     }, PDF_VIEWER_ORIGIN);
 
     return true;
   }
 
   requestDataRange(args) {
     if (!this.rangeEnabled) {
       return;

```

yurydelendik (approved):
Approach is right. (The end goal was to have FirefoxCom match Streams API). Let's do it as you implemented for now. Thanks.

timvandermeij:
Have the changes in `mozilla-central` already been made, since I think that that's a requirement for merging this, or is that not strictly necessary to land this here?

Snuffleupagus:
> Have the changes in `mozilla-central` already been made, since I think that that's a requirement for merging this, or is that not strictly necessary to land this here?

Ideally, they should probably land *simultaneously* with the changes being made here.

timvandermeij:
> Also, I'm no longer able to submit patches for mozilla-central, since you now need a 2FA device to do so.

It's also possible for 2FA that you're presented with a secret key (I believe it's 16 characters) so that you can use, for example, Google Authenticator from a desktop. Refer to https://www.labnol.org/internet/google-authenticator-for-desktop/25341/ and https://github.com/grahammitchell/google-authenticator. Perhaps you already thought of such a solution though, but I thought I'd just mention it since it might help out with not needing an actual device.

Snuffleupagus:
> Perhaps you already thought of such a solution though, but I thought I'd just mention it since it might help out with not needing an actual device.

While that ought to work, and there's even e.g. a [Firefox addon](https://addons.mozilla.org/en-US/firefox/addon/auth-helper/) that could probably be used doesn't it sort of negate the whole point of 2FA though?
By using the one and the same device, doesn't that essentially negate the security benefits of using 2FA in the first place? Anyway, I really appreciate your suggestions!

timvandermeij:
> doesn't it sort of negate the whole point of 2FA by using one and the same device though

Not necessarily. Two-factor authentication is nothing more than adding an additional step to the login process. It looks like it doesn't actually has to be another device, it should just be another type of data that someone else cannot obtain (easily). From https://en.wikipedia.org/wiki/Wikipedia:Simple_2FA: "You can use apps like WinAuth and KeeWeb to handle 2FA tokens on a desktop computer. This is the recommended way to use 2FA if you don't have a smartphone or tablet computer". If your username/password would become public after a database breach, the 2FA token would still prevent people from logging in to your account.

Snuffleupagus:
> Have the changes in `mozilla-central` already been made, since I think that that's a requirement for merging this, or is that not strictly necessary to land this here?

Looking at this again I don't believe that needs to happen first, since the only thing that will happen without the `PdfStreamConverter.jsm` changes is that the `disableRange` regression simply won't be fixed yet.

@timvandermeij Mind landing this to get the ball rolling; thanks!
(Also, please feel free to mark a couple of my comments above as off-topic if you like.)

---

@rvandermeulen Could I please trouble you to include the following diff, https://gist.github.com/Snuffleupagus/960c70fe1775af04f8711f1699580a94, when the changes from this PR are uplifted to `mozilla-central` next?
(I'm still struggling with the whole Phabricator situation, and the Windows setup is "interesting" to say the least, so I'm unsure when I'll be able to submit patches myself.)

rvandermeulen:
Sure thing! FWIW, I've been using https://bitbucket.org/kmaglione/hgext/src/default/phabricator.py on Windows rather than messing with arc and it works well. Feel free to drop me a line on email or IRC if you need help setting it up.

timvandermeij:
Thank you!

PR #10678, opened 2019-03-25
Remove `moz-chunked-arraybuffer` support, and related code, from `src/display/network.js`

Snuffleupagus:
The `moz-chunked-arraybuffer` responseType is a non-standard property, which has been subsumed by the Fetch API, and it's in the process of being removed from Firefox; please see https://bugzilla.mozilla.org/show_bug.cgi?id=1120171 and https://bugzilla.mozilla.org/show_bug.cgi?id=1411865

*Please note:* Rather than waiting for both `Fetch` *and* `ReadableStream` to be available in e.g. a Firefox ESR version (which is probably going to be 68 at the earliest), let's just decide that PDF.js release `2.1.266` will be the last one with `moz-chunked-arraybuffer` support and land this patch (since nothing should outright break without it anyway).

### DIFF
--- a/src/display/network.js
+++ b/src/display/network.js
@@ -18,7 +18,6 @@ import {
   createResponseStatusError, extractFilenameFromHeader,
   validateRangeRequestCapabilities
 } from './network_utils';
-import globalScope from '../shared/global_scope';
 
 if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('FIREFOX || MOZCENTRAL')) {
   throw new Error('Module "./network" shall not ' +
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
> [...] please see https://bugzilla.mozilla.org/show_bug.cgi?id=1120171 and https://bugzilla.mozilla.org/show_bug.cgi?id=1411865

Both of these have now landed, hence I believe that we should just move forward with this PR.

timvandermeij (approved):

timvandermeij:
Very nice cleanup!

Issue #10093, opened 2018-09-21
NodeJS detection breaking pdf.js when used with Angular CLI and Electron


wartab:
Attach (recommended) or Link to PDF file here: Any PDF file

Configuration:
- Web browser and its version: Electron 2/3
- Operating system and its version: All
- PDF.js version: 2.0.550
- Is a browser extension:

Steps to reproduce the problem:
1. Use a working example of pdf.js within an Angular 6 App
2. It works in the browser
3. It doesn't work in Electron

What is the expected behavior? (add screenshot)
It should download the PDF files to render them, but it doesn't download them.

What went wrong? (add screenshot)
Since Electron contains the `process` variable (https://github.com/mozilla/pdf.js/blob/master/src/shared/is_node.js), pdf.js assumes that it needs to use the default `http` package from NodeJS to download the PDF file. But Angular will silently fail the import of `http` by returning an empty object (causing it to fail loading PDF files: See issue here https://github.com/VadimDez/ng2-pdf-viewer/issues/358).

I'm not certain why NodeJS is even considered here, but I think that a deeper feature detection would be useful to detect a clean NodeJS environment. My suggestion would be to check the existence of require("http").request and require("https").request before assuming they can be used.

I know this is very specific, but it would help us out to avoid dirty quick fixes such as destroying the `process` variable that might be otherwise useful to keep in Electron.

Snuffleupagus:
> I'm not certain why NodeJS is even considered here, but I think that a deeper feature detection would be useful to detect a clean NodeJS environment. My suggestion would be to check the existence of require("http").request and require("https").request before assuming they can be used.
> 
> I know this is very specific, but it would help us out to avoid dirty quick fixes such as destroying the `process` variable that might be otherwise useful to keep in Electron.

This all suggests that the actual problem here is that some libraries/frameworks are pretending to be Node.js, despite not actually containing all the methods/functionality of a *proper* Node.js environment.

It might be possible to improve the Node.js detection in https://github.com/mozilla/pdf.js/blob/master/src/shared/is_node.js, but if/how that should be done someone else will have to chime in on. @yurydelendik Any ideas/opinions here?

wartab:
I reality, Electron is a proper Node.js environment, it's just that (I'm speculating) Angular's Webpack configuration will silently fail. If I were to do the same with `window["require"]("http")`, it would work perfectly fine.

wartab:
This has still not been addressed.
[...]

PR #10228, opened 2018-11-06
Don't detect nw.js as node.js

romainptit:
nw.js is chrome plus nodejs.
It will succeed everywhere chrome succeeds, but fail in many cases where nodejs succeeds (see issue 9071).
So it's safer to consider it as a browser context rather than a nodejs context.

Fixes #9071.

PR #10681, opened 2019-03-26
fix: electron enviroment detection

CodeDaraW:
Like #10228 , just treat Electron as a browser, but not Node.js.  

https://electronjs.org/docs/api/process#processversionselectron


timvandermeij:
This should fix #10093.

timvandermeij (approved):

timvandermeij:
Thank you!

PR #10683, opened 2019-03-26
Use CMap in Type0 fonts when CFF is not a CID font

### LINKED ISSUE #9534: PDF.js does not render correctly
[tmp1520327772592644.pdf](https://github.com/mozilla/pdf.js/files/1784252/tmp1520327772592644.pdf)

Configuration:
- Web browser and its version: Firefox Quantum 58.0.2 (64 bits)
- Operating system and its version: Windows 10 Home
- PDF.js version: 119
- Is a browser extension: no

1. viewer.html?file=tmp1520327127416893.pdf
2.

ok

ko

http://81.184.7.209/v_119/web/viewer.html?file=tmp1520327127416893.pdf

janpe2:
Fixes #9534
in which the PDF uses composite Type0 fonts and the embedded CFF data is not in CID font format. The problems are caused by the `Encoding` entry whose value is an embedded CMap stream. The existing implementation ignores the CMap, so wrong glyphs are rendered. 

Most PDFs like this use `/Encoding /Identity-H` instead. They have no problems and their rendering is not affected by this PR.


janpe2:
Actually the PDF isn't broken. The spec mentions that CFF data can be used like this. Because the CFF doesn't support CIDs, glyph indexes are used as CID values.

Snuffleupagus (approved):

timvandermeij (approved):

timvandermeij:
Nice find, thanks!

PR #10685, opened 2019-03-26
Take the `FirstChar`/`LastChar` properties into account when computing the hash in `PartialEvaluator.preEvaluateFont` (issue 10665)

### LINKED ISSUE #10665: Badly rendered Times New Roman PS
[EMG8 -Cambridge Essentail Gold Maths 8-B 117.pdf](https://github.com/mozilla/pdf.js/files/2994541/EMG8.-Cambridge.Essentail.Gold.Maths.8-B.117.pdf)

Configuration:
- Web browser and its version: **Chrome 73**
- Operating system and its version: ** Macos  10.14.1 **
- PDF.js version: **2.2.91**
- Is a browser extension: No

1. 
Open the PDF (its a single page extracted from a book) and see the badly rendered fonts

image

On any other PDF reader it renders fine:
image

Current github pages viewer: https://mozilla.github.io/pdf.js/web/viewer.html

Notes:
We've managed make it work by editing the fonts within the PDF. It is currently "Times New Roman PS", re-rendering it with just "Times New Roman" seems to fix it.

There are no console errors, or any other visible signs of solutions such as missing CMaps.

Unfortunately we are not allowed to alter PDFs, so re-rendering each one is not a viable solution. 

If anyone can give any insight into this and a possible solution, that would be massively appreciated 🤠

Snuffleupagus:
Without this some fonts may incorrectly end up with matching `hash`es, thus breaking rendering since we'll not actually try to load/parse some of the fonts.

Fixes #10665.


src/core/evaluator.js
```diff
@@ -2352,6 +2350,10 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           }
         }
 
+        const firstChar = (dict.get('FirstChar') || 0);
+        const lastChar = (dict.get('LastChar') || (composite ? 0xFFFF : 0xFF));
+        hash.update(`${firstChar}-${lastChar}`);
```
Snuffleupagus: The updated patch only changed the separator here, since using the same one (i.e. comma) as in the `'Differences' array` case above may perhaps not be the best idea.

timvandermeij (approved):

timvandermeij:
Thank you! I'll make reference images in the other PR to save resources.

PR #10696, opened 2019-03-29
Update `ChunkedStream.makeSubStream` to actually check if (some) data exists when the `length` parameter is undefined

Snuffleupagus:
Note how `XRef.fetchUncompressed`, which is used *a lot* for most PDF documents, is calling the `makeSubStream` method without providing a `length` argument.
In practice this results in the `makeSubStream` method, on the `ChunkedStream` instance, calling the `ensureRange` method with `NaN` as the end position, thus resulting in no data being requested despite it possibly being necessary.

This may be quite bad, since in this particular case it will lead to a new `ChunkedStream` being created *and* also a new `Parser`/`Lexer` instance. Given that it's quite possible that even the very first `Parser.getObj` call could throw `MissingDataException`, this could thus lead to wasted time/resources (since re-parsing is necessary once the data finally arrives).

You obviously need to be very careful to not have `ChunkedStream.makeSubStream` accidentally requesting the *entire* file, hence its `this.end` property is of no use here, but it should be possible to at least check that the `start` of the data is present before any potentially expensive parsing occurs.

### DIFF
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -234,7 +234,20 @@ class ChunkedStream {
   }
 
   makeSubStream(start, length, dict) {
-    this.ensureRange(start, start + length);
+    if (length) {
+      this.ensureRange(start, start + length);
+    } else {
+      // When the `length` is undefined you do *not*, under any circumstances,
+      // want to fallback on calling `this.ensureRange(start, this.end)` since
+      // that would force the *entire* PDF file to be loaded, thus completely
+      // breaking the whole purpose of using streaming and/or range requests.
+      //
+      // However, not doing any checking here could very easily lead to wasted
+      // time/resources during e.g. parsing, since `MissingDataException`s will
+      // require data to be re-parsed, which we attempt to minimize by at least
+      // checking that the *beginning* of the data is available here.
+      this.ensureByte(start);
+    }
 
     function ChunkedStreamSubstream() {}
     ChunkedStreamSubstream.prototype = Object.create(this);
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -98,6 +98,10 @@ class ChunkedStream {
   }
 
   ensureByte(pos) {
+    if (pos < this.progressiveDataLength) {
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


kaoeeng:
ในวันที่ ส. 30 มี.ค. 2019 02:56 pdf.js test bot <notifications@github.com>
เขียนว่า:

Snuffleupagus:
Looking at the `browsertest` results it seems that the second patch may have reduced the runtime by approximately 1%, which is great considering just how simple it really is :-)

src/core/chunked_stream.js
```diff
@@ -98,6 +98,10 @@ class ChunkedStream {
   }
 
   ensureByte(pos) {
+    if (pos < this.progressiveDataLength) {
```
timvandermeij: I assume `pos` is 0-indexed, so the `<` check will work. However, why are we using `<=` then in `ensureRange` below? I think the `end` is also a 0-indexed position, so it that wrong below?

Snuffleupagus: The way that I've understood `MissingDataException`, and related code, is that `begin` is inclusive while `end` is *exclusive*, which should thus explain things; note the formatting used in https://github.com/mozilla/pdf.js/blob/47f208df360c750c40f06d4573377fd6e90b6e6d/src/core/core_utils.js#L32-L35 and compare with standard interval notation as outlined in https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints

To be absolutely sure though, I suppose that you might want @yurydelendik to weigh in here!

Snuffleupagus: @yurydelendik Sorry to bother you, but any chance that you have time to comment here?

Snuffleupagus: Also, given that a `ensureByte(pos)` call should essentially be equal to a `ensureRange(pos, pos + 1)` call, that would mean a `if (pos + 1 <= this.progressiveDataLength) {` check in the latter case which is how I arrived at the condition under discussion here.

timvandermeij (approved):

timvandermeij:
Thank you!

PR #10709, opened 2019-04-03
[api-minor] Add basic support for PageLayout in the API and the viewer

Snuffleupagus:
Please see the specification, https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G6.2393749, and refer to the inline comments for additional details.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -889,6 +889,8 @@ let PDFViewerApplication = {
 
     // Since the `setInitialView` call below depends on this being resolved,
     // fetch it early to avoid delaying initial rendering of the PDF document.
+    const pageLayoutPromise = pdfDocument.getPageLayout().catch(
+      function() { /* Avoid breaking initial rendering; ignoring errors. */ });
     const pageModePromise = pdfDocument.getPageMode().catch(
       function() { /* Avoid breaking initial rendering; ignoring errors. */ });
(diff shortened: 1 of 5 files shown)


web/app.js
```diff
@@ -2434,6 +2439,29 @@ function webViewerKeyDown(evt) {
   }
 }
 
+/**
+ * Converts API PageLayout values to the format used by `PDFViewer`.
+ * NOTE: This is supported to the extent that the viewer implements the
+ *       necessary Scroll/Spread modes (since SinglePage, TwoPageLeft,
+         and TwoPageRight all suggests using non-continuous scrolling).
```
timvandermeij: It looks like an asterisk is missing here at the beginning of the line.

timvandermeij (approved):
Looks good with the comment addressed and passing tests.

web/app.js
Snuffleupagus: D'oh, fixed now :-)

Thinking about this patch I suppose that it might surprise users when some documents now opens with Spread mode enabled, but given that this is already controllable via existing prefs[1] I really didn't want to add any new ones here.

---
In Firefox, setting `pdfjs.spreadModeOnLoad = 0` in `about:config` is now guaranteed to never enable any Spread mode by default.

timvandermeij:
Nice addition!

PR #10714, opened 2019-04-06
[Firefox] Ensure that loading progress is reported, and the loadingBar updated, when `disableRange=true` is set

Snuffleupagus:
With PR #10675 having fixed the completely broken `disableRange=true` setting in the Firefox version of PDF.js, I couldn't help but noticing that loading progress is never reported properly in that case.
Currently loading progress is only reported for the `rangeProgress` chrome-event, which obviously isn't dispatched with `disableRange=true` set. However, the `progressiveRead` chrome-event includes loading progress as well, but this information isn't being used in any way.
Furthermore, the `PDFDataRangeTransport.onDataProgress` method wasn't able to handle "complete" loading information, and neither was `PDFDataTransportStream._onProgress` since that method would only ever attempt to report it through a RangeReader (which won't exist when `disableRange=true` is set).

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -515,6 +515,7 @@ const PDFDocumentLoadingTask = (function PDFDocumentLoadingTaskClosure() {
  * Abstract class to support range requests file loading.
  * @param {number} length
  * @param {Uint8Array} initialData
+ * @param {boolean} progressiveDone
  */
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #10715, opened 2019-04-07
Add test for fetch_stream

### LINKED ISSUE #10671: Introduce unit tests for reading files with streaming
In #10668 we removed a Firefox-specific unit test for this, but this is useful functionality to unit test. Hence the idea to introduce new unit tests for this that work in all browsers and, if only code from `fetch_stream.js` is tested, that live in their own file.

mukulmishra18:
This PR adds test for `fetch_stream.js`. Not sure if we need to add some code for feature detection for native support of `fetch` and `ReadableStream` for unit test.

Fixes #10671

### DIFF
--- a/test/unit/fetch_stream_spec.js
+++ b/test/unit/fetch_stream_spec.js
@@ -1,0 +1,108 @@
+/* Copyright 2019 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


test/unit/fetch_stream_spec.js
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    const fullReader = stream.getFullReader();
+
+    let isStreamingSupported, isRangeSupported, fullReaderCancelled;
+    const promise = fullReader.headersReady.then(function() {
+      isStreamingSupported = fullReader.isStreamingSupported;
+      isRangeSupported = fullReader.isRangeSupported;
+      // We shall be able to close full reader without any issue.
+      fullReader.cancel('Don\'t need full reader');
```
Snuffleupagus: This really ought to be an `Error`, and not a plain string.

test/unit/fetch_stream_spec.js
```diff
@@ -0,0 +1,108 @@
+/* Copyright 2017 Mozilla Foundation
```
Snuffleupagus: This should probably read 2019 :-)

mukulmishra18: I think I still live in 2017 😱

timvandermeij (approved):

timvandermeij:
Nice work; thank you for writing more unit tests!

PR #10724, opened 2019-04-10
Add translation for central Kurdish(ckb) language

allekok:


Snuffleupagus:
As outlined in https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md, this unfortunately cannot be accepted since translations are imported from Firefox:

allekok:
I didn't know that. This part of documentation is not so clear.

allekok:
Should I close this PR?

timvandermeij:
The documentation is now updated to be more clear about this. I'll close this PR, but please feel free to send the contents of this PR to the upstream maintainers. You can find their contact information at https://wiki.mozilla.org/L10n:Teams:ku. If that is completed, we will import the changes here as well. Thanks!

allekok:
Ok, thanks.

PR #10727, opened 2019-04-11
Support (rare) Type3 fonts which contains image resources (issue 10717)

### LINKED ISSUE #10717: Error rendering multi-page PDF with emojis
[test.pdf](https://github.com/mozilla/pdf.js/files/3057353/test.pdf)
Document was generated using Google Docs and just copying the Balloon emoji into it from https://emojipedia.org/balloon/

Configuration:
- Web browser and its version: Chrome 73.0.3683.86 and Firefox 66.0.2
- Operating system and its version: Windows 10 and Mac OS X
- PDF.js version: 2.0.943 and 2.2.129
- Is a browser extension: No

1. Go to https://mozilla.github.io/pdf.js/web/viewer.html
2. Upload test.pdf

Render PDF fully

What went wrong?
It renders the first page, but fails to render the second, throwing the following error:
```
PDF.js v2.2.129 (build: 725a6959)
Message: Requesting object that isn't resolved yet img_p0_2.
Stack: Error: Requesting object that isn't resolved yet img_p0_2.
    at PDFObjects.get (https://mozilla.github.io/pdf.js/build/pdf.js:11908:15)
    at CanvasGraphics_executeOperatorList [as executeOperatorList] (https://mozilla.github.io/pdf.js/build/pdf.js:14801:24)
    at CanvasGraphics_showType3Text [as showType3Text] (https://mozilla.github.io/pdf.js/build/pdf.js:15556:14)
    at CanvasGraphics_showText [as showText] (https://mozilla.github.io/pdf.js/build/pdf.js:15372:21)
    at CanvasGraphics_executeOperatorList [as executeOperatorList] (https://mozilla.github.io/pdf.js/build/pdf.js:14791:22)
    at InternalRenderTask._callee$ (https://mozilla.github.io/pdf.js/build/pdf.js:12149:51)
    at tryCatch (https://mozilla.github.io/pdf.js/build/pdf.js:12235:17)
    at Generator.invoke [as _invoke] (https://mozilla.github.io/pdf.js/build/pdf.js:12410:22)
    at Generator.prototype.(anonymous function) [as next] (https://mozilla.github.io/pdf.js/build/pdf.js:12278:21)
    at asyncGeneratorStep (https://mozilla.github.io/pdf.js/build/pdf.js:9788:103)
```

Snuffleupagus:
The Type3 font type is not commonly used in PDF documents, as can be seen from telemetry data such as: https://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2019-04-09&include_spill=0&keys=__none__!__none__!__none__&max_channel_version=nightly%252F68&measure=PDF_VIEWER_FONT_TYPES&min_channel_version=nightly%252F57&processType=*&product=Firefox&sanitize=1&sort_by_value=0&sort_keys=submissions&start_date=2019-03-18&table=0&trim=1&use_submission_date=0 (see also https://github.com/mozilla/pdf.js/wiki/Enumeration-Assignments-for-the-Telemetry-Histograms#pdf_viewer_font_types).

Type3 fonts containing image resources are *very* rare in practice, usually they only contain path rendering operators, but as the issue shows they unfortunately do exist.
Currently these Type3-related image resources are not handled in any special way, and given that fonts are document rather than page specific rendering breaks since the image resources are thus not available to the *entire* document.
Fortunately fixing this isn't too difficult, but it does require adding a couple of Type3-specific code-paths to the `PartialEvaluator`. In order to keep the implementation simple, particularily on the main-thread, these Type3 image resources are completely decoded on the worker-thread to avoid adding too many special cases. This should not cause any issues, only marginally less efficient code, but given how rare this kind of Type3 font is adding premature optimizations didn't seem at all warranted at this point.

Fixes #10717.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2003,6 +2003,7 @@ class WorkerTransport {
           });
           break;
         case 'FontPath':
+        case 'FontType3Res':
           this.commonObjs.resolve(id, exportedData);
           break;
         default:
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -73,6 +73,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
     this.builtInCMapCache = builtInCMapCache;
     this.options = options || DefaultPartialEvaluatorOptions;
     this.pdfFunctionFactory = pdfFunctionFactory;
+    this.parsingType3Font = false;
 
     this.fetchBuiltInCMap = async (name) => {
       if (this.builtInCMapCache.has(name)) {
@@ -293,21 +294,21 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       });
     },
 
-    buildPaintImageXObject({ resources, image, isInline = false, operatorList,
-                             cacheKey, imageCache,
-                             forceDisableNativeImageDecoder = false, }) {
+    async buildPaintImageXObject({ resources, image, isInline = false,
+                                   operatorList, cacheKey, imageCache,
+                                   forceDisableNativeImageDecoder = false, }) {
       var dict = image.dict;
       var w = dict.get('Width', 'W');
       var h = dict.get('Height', 'H');
 
       if (!(w && isNum(w)) || !(h && isNum(h))) {
         warn('Image dimensions are missing, or not numbers.');
-        return Promise.resolve();
+        return;
       }
       var maxImageSize = this.options.maxImageSize;
       if (maxImageSize !== -1 && w * h > maxImageSize) {
         warn('Image exceeded maximum allowed size and was removed.');
-        return Promise.resolve();
+        return;
       }
 
       var imageMask = (dict.get('ImageMask', 'IM') || false);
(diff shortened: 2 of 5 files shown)


src/display/api.js
```diff
@@ -2003,6 +2003,7 @@ class WorkerTransport {
           });
           break;
         case 'FontPath':
+        case 'FontType3Res':
```
brendandahl: I may be missing something, but why can't we just stick these in the `objs` so we don't have to do all the typ3e checks in canvas?

brendandahl (changes requested):

src/display/api.js
Snuffleupagus: That's what's currently happening without this patch, and that breaks because fonts are common to the *entire* document (i.e. placed in `commonObjs`), whereas the `objs` are unique for *each* page.
Since the same font is used on multiple pages, but the image resources were only available to the first page, using `objs` cannot work for these kind of font resource.

Hence the only simple way, that I could find at least, was to make these resources available "globally" which thus necessitated small changes in `src/display/canvas.js`.

brendandahl (approved):

timvandermeij:
Thanks!

PR #10728, opened 2019-04-11
Annotations - _preparePopup method replaced with MarkupAnnotation

vlastimilmaca:
_preparePopup method replaced with MarkupAnnotation base class. 

This is just refactoring, so it shouldn't break anything. It should move annotation API closer to PDF spec and enable future expansion.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -397,24 +397,6 @@ class Annotation {
     this.appearance = normalAppearanceState.get(as.name);
   }
 
-  /**
-   * Prepare the annotation for working with a popup in the display layer.
-   *
-   * @private
-   * @memberof Annotation
-   * @param {Dict} dict - The annotation's data dictionary
-   */
-  _preparePopup(dict) {
-    if (!dict.has('C')) {
-      // Fall back to the default background color.
-      this.data.color = null;
-    }
-
-    this.data.hasPopup = dict.has('Popup');
-    this.data.title = stringToPDFString(dict.get('T') || '');
-    this.data.contents = stringToPDFString(dict.get('Contents') || '');
-  }
-
   loadResources(keys) {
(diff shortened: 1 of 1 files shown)


src/core/annotation.js
```diff
@@ -598,6 +580,26 @@ class AnnotationBorderStyle {
   }
 }
 
+/**
```
timvandermeij: This comment block can be removed because it only states where to find it in the spec, which is something we never really do elsewhere unless really required to understand the code, which isn't the case here.

src/core/annotation.js
```diff
@@ -598,6 +580,26 @@ class AnnotationBorderStyle {
...
+ * Class for Markup Annotations
+ * defined in spec: PDF 32000-1:2008 (page 391, chapter 12.5.6.2)
+ */
+class MarkupAnnotation extends Annotation {
+  constructor(parameters) {
+    super(parameters);
+    let dict = parameters.dict;
```
timvandermeij: `let` should be `const`.

timvandermeij (approved):
Good to merge with these comments addressed and passing tests.

vlastimilmaca:
Comments addressed, but I have no idea what to do with that failed test ... there aren't any annotations ... so no idea what is wrong with that ...

Snuffleupagus (commented):
Please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

vlastimilmaca:
Oh ... ok ... sorry ...

vlastimilmaca:
About the test … there is a color change of column separator in table … reference says #5c5c5c and test says #5b5b5b .. I really have no clue how or why this changed … did something like this happen before? Is it a bad idea to run that test again ?

timvandermeij:
Thank you!

(The reference test failure is intermittent and unfortunately happens from time to time. It's not related to your patch.)

Issue #10736, opened 2019-04-14
Books bound on the right-hand side spread with inverted order


mandel59:
Attach (recommended) or Link to PDF file here:

http://www.asahi-net.or.jp/~fj6k-hrks/aozora/utayomini_ataurusho.pdf
(via http://www.asahi-net.or.jp/~fj6k-hrks/aozora/)

Configuration:
- Web browser and its version: Firefox 66.0.3
- Operating system and its version: Windows 10
- PDF.js version: Firefox 66.0.3 built-in
- Is a browser extension: yes

Steps to reproduce the problem:
1. Select even spreads.

What is the expected behavior? (add screenshot)

![image](https://user-images.githubusercontent.com/1079715/56090088-1c2a6c80-5ed8-11e9-87d8-87c61b88e5d3.png)

What went wrong? (add screenshot)

![image](https://user-images.githubusercontent.com/1079715/56090080-fef59e00-5ed7-11e9-9e0e-d377a629e79b.png)

Because this book is [bound on the right-hand side](https://www.w3.org/TR/jlreq/#term.bound-on-the-right-hand-side), the page 4 should be a right page, not left.

PR #10738, opened 2019-04-14
[api-minor] Add support for ViewerPreferences in the API (issue 10736)

Snuffleupagus:
Please see the specification, https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#M11.9.12864.1Heading.71.Viewer.Preferences

Furthermore, note that this patch *only* adds API support and unit-tests but does not attempt to integrate e.g. the `ViewerPreferences -> Direction` property into the viewer (which would be necessary to address issue #10736).
The reason for this is that it's not entirely clear to me exactly if/how that could be implemented; e.g. would it be as simple as setting the `dir` attribute on the `viewerContainer` DOM element, or will it be more complicated?
There's also the question of how the `ViewerPreferences -> Direction` value interacts with the `PageMode`, and this will generally require a fair bit of manual testing. Since the direction of the *entire* viewer depends on the browser locale, there's also a somewhat open question regarding what default value to use for different locales.
Finally, if the viewer supports `ViewerPreferences -> Direction` then I'm assuming that it will be necessary to allow users to override the default value, which will require (most likely) new `SecondaryToolbar` buttons and icons for those etc.

Hence this patch only lays the necessary foundation for eventually addressing issue #10736, but defers the actual implementation until later. (Time permitting, I'll try to look into the viewer part later.)

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -14,9 +14,10 @@
  */
 
 import {
-  bytesToString, createPromiseCapability, createValidAbsoluteUrl, FormatError,
-  info, InvalidPDFException, isBool, isNum, isString, PermissionFlag, shadow,
-  stringToPDFString, stringToUTF8String, unreachable, warn
+  assert, bytesToString, createPromiseCapability, createValidAbsoluteUrl,
+  FormatError, info, InvalidPDFException, isBool, isNum, isString,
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving the API!

Issue #1076, opened 2012-01-13
Digital signatures in pdf.js


soa-x:
Hi! We are interested about when or if you´re going to implement the Signature Data View (Xades, Pades & Cades) in the PDF viewer (PDF.js)

Kind Regards

Alejandro Pinedo,
SOA-X

corynewey:
@yurydelendik @wolvz 
I am extremely interested in the digital signature display feature. I cloned wolvz's repository and built the code but when I try to use it, I'm getting a promise rejection with this error: "require is not defined". Could anyone explain what would cause that error and how I could fix it? Also, I am very willing to continue the work on this feature. I've worked with pdf's in Java via the iText library but working with them in javascript is something I've never tried before. Could someone direct me to some documentation/tutorial that could get me pointed in the right direction so that I can figure out where to start as I try to move this feature further down the field?

wolvz:
@corynewey here's a page I created that does signature validation on signed PDFs: http://wolvz.github.io/
It doesn't check if the certificate is trusted by the browser... it only tests if the certificate chain is "well formed".
I don't know if it is compatible with signed pdfs created with other applications than Adobe Reader... I tested a few PDFs and it worked fine...
Might give you some ideas if you try to do the integration with pdf.js... :)

mrpandya007:
I am confused. What i have to do in my pdf.js library to show digital signature pdf in pdfJS

mrpandya007:
![untitled](https://cloud.githubusercontent.com/assets/11521993/6702971/d9e1986c-cd59-11e4-8860-a220415f55b5.png)

This signature and certificate both showing in pdf.. but when i open it in pdf.js it wont showing there.

wolvz:
@mrpandya007, the digital signature doesn't show because the developers chose to hide it until they have signature verification feature working.
If I recall, if you want them simply to show up you have to comment lines 389 to 392 here: https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L389
Not sure though, and I can't test it right now.

mrpandya007:
@wolvz, Thanx for the response but its already commented. PLease look into this,, Crystal report's generated pdf signature is not showing but DevExpress generated pdf is showing it properly

mrpandya007:
Please answer me .... what i have to do.. I can give you that pdf to test

timvandermeij:
There is no good support for digital signatures yet, so this will have to be implemented.

mrpandya007:
[...]

PR #10739, opened 2019-04-14
Expose signature data to the main thread via WidgetAnnotation

stefanmaric:
A Signature widget is not available since it makes no sense to display
a signature if it cannot be validated.

But exposing the signature allows the PDF to verified outside of
PDF.js, but taking advantage of PDF.js' parsing.

See: Issue #1076
See: PR #7702
See: PR #10350

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -624,11 +624,17 @@ class WidgetAnnotation extends Annotation {
 
     data.readOnly = this.hasFieldFlag(AnnotationFieldFlag.READONLY);
 
-    // Hide signatures because we cannot validate them, and unset the fieldValue
-    // since it's (most likely) a `Dict` which is non-serializable and will thus
-    // cause errors when sending annotations to the main-thread (issue 10347).
+    // Hide signatures because we cannot validate them, and replace fieldValue
+    // with the signature, the content range, and signer's name only, since it's
+    // (most likely) a `Dict` which is non-serializable and will thus cause
+    // errors when sending annotations to the main-thread (issue 10347).
     if (data.fieldType === 'Sig') {
-      data.fieldValue = null;
+      data.fieldValue = {
+        ByteRange: data.fieldValue.get('ByteRange'),
+        Contents: data.fieldValue.get('Contents'),
+        Name: data.fieldValue.get('Name'),
+      };
+
       this.setFlags(AnnotationFlag.HIDDEN);
     }
   }


stefanmaric:
Don't have much hope on this being merged. This is just to satisfy a personal use case.

I think I would be capable to port PR #7702 into the updated codebase, but without a clear path on how to design the API so PDF.js can be given the root certificates, it makes no sense to work on this feature.

src/core/annotation.js
```diff
@@ -624,11 +624,17 @@ class WidgetAnnotation extends Annotation {
...
+    // Hide signatures because we cannot validate them, and replace fieldValue
+    // with the signature, the content range, and signer's name only, since it's
+    // (most likely) a `Dict` which is non-serializable and will thus cause
+    // errors when sending annotations to the main-thread (issue 10347).
     if (data.fieldType === 'Sig') {
-      data.fieldValue = null;
+      data.fieldValue = {
+        ByteRange: data.fieldValue.get('ByteRange'),
+        Contents: data.fieldValue.get('Contents'),
+        Name: data.fieldValue.get('Name'),
```
Snuffleupagus: In a malformed PDF document, there's nothing stopping any of the properties above from being e.g. a `Dict` or a `Stream`, thus making them non-serializable which will throw errors during `postMessage`.

Assuming that we actually want to do this, which seems questionable since this is essentially adding dead code which normally isn't accepted (there's no call-site/unit-test using this information), there would need to be *a lot* more validation added for this to be even remotely safe.

stefanmaric: > In a malformed PDF document, there's nothing stopping any of the properties above from being e.g. a `Dict` or a `Stream`, thus making them non-serializable which will throw errors during `postMessage`.

Oh, good point.

> Assuming that we actually want to do this, which seems questionable since this is essentially adding dead code

Totally agree. As I said, don't think this is useful at all for the project, but instead something I need for a project.

I've been looking around, but haven't found any way to get the data I need out of the worker. Is there something I'm missing? — It seems the main thread is prohibited from touching the PDF outside of the AST the worker offers.

timvandermeij:
I would also say that this on its own is not really enough for adding this to the API. Aside from the validation and test coverage problems, it's only adding a few of the many available fields for digital signatures and their validation. I think a patch that would expose and validate all required fields could be accepted, with sufficient test coverage, but as it stands this PR is unfortunately not complete enough.

PR #10742, opened 2019-04-16
[Firefox] Avoid displaying the indeterminate loadingBar when `disableStream=true` is set (PR 10714 follow-up)

Snuffleupagus:
While PR #10714 did address the `disableRange=true` case, it also managed to "break" the `disableStream=true` case instead since the indeterminate loadingBar is now displayed when it shouldn't; sorry about that!
The solution is simple enough though, don't attempt to fallback to `_fullRequestReader.onProgress` when handling "incomplete" loading information. *Edit:* Smaller diff with https://github.com/mozilla/pdf.js/pull/10742/files?w=1

/cc @timvandermeij


timvandermeij (approved):

timvandermeij:
Thank you!

PR #10746, opened 2019-04-17
Add links to PDF.js homepage and API reference in README.md

mvasigh:
Just a tiny quality-of-life change. I frequently find myself landing on the repo from search engines with no quick and visible way to go to the project homepage or API docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # PDF.js [![Build Status](https://travis-ci.org/mozilla/pdf.js.svg?branch=master)](https://travis-ci.org/mozilla/pdf.js)
 
-PDF.js is a Portable Document Format (PDF) viewer that is built with HTML5.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,6 +113,10 @@ More learning resources can be found at:
 
 + https://github.com/mozilla/pdf.js/wiki/Additional-Learning-Resources
 
+### API reference
```
Snuffleupagus: Isn't this one level heading too deep, and also "reference" seems like a strange choice of word here...
Also, I'm not sure why the text from https://mozilla.github.io/pdf.js/api/ needs to be copied verbatim below either (since it adds even more links that may become stale).

Why not simply append the information to the list just above, e.g. something like this instead:
```
The API documentation can be found at:
+ https://mozilla.github.io/pdf.js/api/
```

mvasigh: Added a commit with your suggested change, it fits much more nicely with the surrounding text.

timvandermeij (approved):

timvandermeij:
Thank you!

Issue #7372, opened 2016-05-30
Text Extraction / How to get font information like italic or bold


TuningGuide:
Hi, I'm sorry to misuse this issue reporting for a question. But I guess I get the most usefull information from here.

If I use

    let content = yield page.getTextContent();
    content.items[0]...

How can I tell that item 0 is italic or bold?

timvandermeij:
As far as I know this information is not exposed through the API. Marking as a feature for now.

Snuffleupagus:
In general, I don't think that it's actually possible to obtain this information with any certainty, considering that there's a number of ways a font could be made to look bold/italic.

For example, in non-embedded *standard* fonts, you can often tell by the font name ([as is done here](https://github.com/mozilla/pdf.js/blob/7c7d2399956c25491aa7d2548c9c784cc2b3cea8/src/core/fonts.js#L535-L537)).
For embedded fonts, it's usually a lot more difficult to tell, since the font name may be (and often is) somewhat arbitrary. In this case, whether or not the text looks bold/italic depends on the embedded font data itself.
Then there are the [font flags](https://github.com/mozilla/pdf.js/blob/7c7d2399956c25491aa7d2548c9c784cc2b3cea8/src/core/fonts.js#L101-L111), but based on experience I wouldn't really trust them, since PDF files unfortunately often provides incomplete/inconsistent/broken font information.
Furthermore, even though (in my experience) a majority of PDF files uses different fonts to represent regular/bold/italic text, there are exceptions. One such example is #7297, where the `Text rendering mode` is modified in order to simulate bold text.

So, to summarize, there's unfortunately no way that you can consistently obtain the information you're asking for. And as https://github.com/mozilla/pdf.js/issues/7297#issuecomment-217631260 said, this seems out-of-scope for text-selection/text-extraction use case.

TuningGuide:
@Snuffleupagus thank you for the insights. I'l have a look at the font names used, and hardcode it

PR #10753, opened 2019-04-19
make the original font name available from getTextContent()

jonklein:
The `fontName` available from `getTextContent()` calls appears to be the name of the font loaded to be used in text layer rendering, which can be different from the actual font name loaded from the PDF.  For example, in a test PDF, I see `getTextContent()` always return `fontName` "Helvetica" for chunks which are "Arial", ""Arial,Bold" and "Arial,BoldItalic".  

The original font name can be useful to applications in inferring information about the original PDF text, like determining whether adjacent text chunks belong together or not.  This can help applications improve selection and extraction efforts.

I wasn't able to find any other way to receive this information from the existing APIs, but many discussions/requests about how to retrieve similar information (such as https://github.com/mozilla/pdf.js/issues/7372).

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1259,6 +1259,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           };
         }
         textContentItem.fontName = font.loadedName;
+        textContentItem.originalFontName = font.name;
 
         // 9.4.4 Text Space Details
         var tsm = [textState.fontSize * textState.textHScale, 0,
@@ -1339,6 +1340,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           height: textChunk.height,
           transform: textChunk.transform,
           fontName: textChunk.fontName,
+          originalFontName: textChunk.originalFontName,
         };
       }
 
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -1169,6 +1169,7 @@ describe('api', function() {
             expect(items[0]).toEqual({
               dir: 'ltr',
               fontName: 'Times',
+              originalFontName: 'Times-Roman',
               height: 18,
               str: 'Issue 8276',
               transform: [18, 0, 0, 18, 441.81, 708.4499999999999],


timvandermeij:
Just so I understand, is the difference between `fontName` and `originalFontName` that the former may be different because of font substituations, i.e., if `Arial` is not available that it falls back to `Helvetica`? In that case, I'm fine with adding this property if it makes sense.

@Snuffleupagus Do you perhaps have an opinion/more context for this?

Snuffleupagus:
> @Snuffleupagus Do you perhaps have an opinion/more context for this?

Yes, and I'm not convinced that the additional memory usage that this feature will incur is going to be acceptable given that this is essentially unused code (for e.g. the default viewer).

The problem, as I see it, is that this is adding a new property to *every* single `textItem`, and there can obviously be an arbitrary number of those per page (e.g. easily in the thousands).
The key itself isn't exactly short, but the main issue is that the "original" font name can be up to 127 bytes long (given that it's of type `Name`). All of this will add up, and given the specialized use-case it's not really clear that you want to add this type of overhead to all `getTextContent`/`streamTextContent` calls by default.

timvandermeij:
Good point, that sounds very reasonable to me. It can always be done in a custom deployment of PDF.js, so in that case I'm closing this as won't-fix.

bmaupin:
Is there an alternative way to get the original font name?

If not, it does seem inefficient to add the original font name to `TextItem`, but what about adding it to `TextStyle`? Would a PR with that functionality be accepted? I'm not sure how feasible it would be to accomplish, but I wouldn't want to do the work if it wouldn't be accepted.

Based on this issue and the couple of thumbs up it's gotten (excluding my own :smile:), it seems like I'm not the only one that would be interested.

Thanks!

bmaupin:
... as for whether my idea will be possible, it seems pretty trivial, e.g.

```javascript
        textContent.styles[loadedName] = {
          fontFamily: font.fallbackName,
          originalFontName: font.name, // <--- here
          ascent: font.ascent,
          descent: font.descent,
          vertical: font.vertical,
        };
```

https://github.com/mozilla/pdf.js/blob/fa47d4b9b114579c8dd8971921c6ec2f23555105/src/core/evaluator.js#L2434-L2439

But would such a PR be accepted? Or is there another way to do this?

Thanks!

bmaupin:
In case anyone ends up here trying to get the original font name, it's now possible thanks to #15659. Here's some sample code:

```javascript
const { items, styles } = await pdfPage.getTextContent();

await pdfPage.getOperatorList();

for (const item of items) {
  if ('fontName' in item) {
    console.log('fontName=', item.fontName);
    console.log(
      'originalFontName=',
      page.commonObjs.get(item.fontName).name
    );
    console.log();
  }
}
```

And the output:

```
fontName= g_d0_f1
originalFontName= ArialMT

fontName= g_d0_f3
originalFontName= Arial-BoldMT

fontName= g_d0_f3
originalFontName= Arial-BoldMT

fontName= g_d0_f4
originalFontName= Arial-ItalicMT

fontName= g_d0_f4
originalFontName= Arial-ItalicMT

fontName= g_d0_f5
originalFontName= Arial-BoldItalicMT

fontName= g_d0_f5
originalFontName= Arial-BoldItalicMT
```

PR #10754, opened 2019-04-20
Add a `getDocId` method to the `idFactory`, in `Page` instances, to avoid passing around `PDFManager` instances unnecessarily (PR 7941 follow-up)

Snuffleupagus:
This way we can avoid manually building a "document id" in multiple places in `evaluator.js`, and it also let's us avoid passing in an otherwise unnecessary `PDFManager` instance when creating a `PartialEvaluator`.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -50,7 +50,7 @@ class AnnotationFactory {
     if (!isDict(dict)) {
       return;
     }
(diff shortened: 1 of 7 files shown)


timvandermeij (approved):

timvandermeij:
Nice find!

PR #10756, opened 2019-04-21
Attempt to handle corrupt PDF documents that contains path operators inside of text object (issue 10542)

### LINKED ISSUE #10542: Some hyperlinks and rest of line with the hyperlink is not shown
Some hyperlinks and rest of line with the hyperlink is not shown

PDF file is here:
[dc_iprcvisa.pdf](https://github.com/mozilla/pdf.js/files/2851693/dc_iprcvisa.pdf)

Configuration:
- Web browser and its version: Chrome 71.0.3578.98
- Operating system and its version: Windows 10 Pro
- PDF.js version: 2.2.2
- Is a browser extension: no

1. open [dc_iprcvisa.pdf](https://github.com/mozilla/pdf.js/files/2851693/dc_iprcvisa.pdf) in pdf.js
2. First hyperlink and rest of the line is not shown, second one is shown correctly

What is the expected behavior? (add screenshot) Both hyperlinks will be shown

Snuffleupagus:
First of all, while this simple approach appears to work OK in practice I'm not sure if it's the best way of addressing the problem (assuming that you even want to).
Second of all, while the solution implemented here only requires tracking/checking one new boolean in order for this to work, I'm nonetheless not entirely happy about this since it will add additional overhead (albeit *very* small) to the parsing of all PDF documents just for a handful of *corrupt* ones.

Fixes #10542.

---

@janpe2 As always; thank you for the excellent debugging done in https://github.com/mozilla/pdf.js/issues/10542#issuecomment-462912415 :-)

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -822,14 +822,30 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       return fontCapability.promise;
     },
 
-    buildPath: function PartialEvaluator_buildPath(operatorList, fn, args) {
+    buildPath(operatorList, fn, args, parsingText = false) {
       var lastIndex = operatorList.length - 1;
       if (!args) {
         args = [];
       }
       if (lastIndex < 0 ||
           operatorList.fnArray[lastIndex] !== OPS.constructPath) {
+        // Handle corrupt PDF documents that contains path operators inside of
+        // text objects, which may shift subsequent text, by enclosing the path
+        // operator in save/restore operators (fixes issue10542_reduced.pdf).
+        //
+        // Note that this will effectively disable the optimization in the
+        // `else` branch below, but given that this type of corruption is
+        // *extremely* rare that shouldn't really matter much in practice.
+        if (parsingText) {
+          warn(`Encountered path operator "${fn}" inside of a text object.`);
+          operatorList.addOp(OPS.save, null);
+        }
+
         operatorList.addOp(OPS.constructPath, [[fn], args]);
+
+        if (parsingText) {
+          operatorList.addOp(OPS.restore, null);
+        }
       } else {
         var opArgs = operatorList.argsArray[lastIndex];
         opArgs[0].push(fn);
@@ -881,6 +897,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
       var self = this;
       var xref = this.xref;
+      let parsingText = false;
       var imageCache = Object.create(null);
 
       var xobjs = (resources.get('XObject') || Dict.empty);
@@ -999,6 +1016,12 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                   operatorList.addOp(OPS.setFont, [loadedName, fontSize]);
                 }));
               return;
+            case OPS.beginText:
+              parsingText = true;
+              break;
+            case OPS.endText:
+              parsingText = false;
+              break;
             case OPS.endInlineImage:
               var cacheKey = args[0].cacheKey;
               if (cacheKey) {
@@ -1158,10 +1181,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
             case OPS.curveTo2:
             case OPS.curveTo3:
             case OPS.closePath:
-              self.buildPath(operatorList, fn, args);
-              continue;
             case OPS.rectangle:
-              self.buildPath(operatorList, fn, args);
+              self.buildPath(operatorList, fn, args, parsingText);
               continue;
             case OPS.markPoint:
             case OPS.markPointProps:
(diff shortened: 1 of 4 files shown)


src/core/evaluator.js
```diff
@@ -822,14 +822,30 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
       if (lastIndex < 0 ||
           operatorList.fnArray[lastIndex] !== OPS.constructPath) {
+        // Handle corrupt PDF documents that contains path operators inside of
+        // text objects, which may shift subsequent text, by enclosing the path
+        // operator in save/restore operators (fixes issue10542_reduced.pdf).
+        //
+        // Note that this will effectively disable the optimization in the
+        // `else` branch below, but given that this type of corruption is
+        // *extremely* rare that shouldn't really matter much in practice.
+        if (parsingText) {
```
Snuffleupagus: Before adding the test-case, I did a `SKIP_BABEL=true gulp test` run with `assert(!parsingText);` added just above this line. Since there wasn't any test failures, that made me a bit less uneasy about the approach of this patch.

Snuffleupagus:
>  I'm nonetheless not entirely happy about this since it will add additional overhead (albeit _very_ small) to the parsing of all PDF documents just for a handful of _corrupt_ ones.

On the other hand: By adding explicit `beginText`/`endText` cases to `getOperatorList` you avoid having to handle them in the `default` branch, which probably doesn't hurt given how extremely common those operators are.

src/core/evaluator.js
```diff
@@ -999,6 +1016,14 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                   operatorList.addOp(OPS.setFont, [loadedName, fontSize]);
                 }));
               return;
+            case OPS.beginText:
+              parsingText = true;
+              args = null;
+              break;
+            case OPS.endText:
+              parsingText = false;
+              args = null;
```
timvandermeij: Why does `args` need to be set to `null` here and above?

timvandermeij (commented):
> which probably doesn't hurt given how extremely common those operators are

I would agree, and I'm fine with the approach, but do you perhaps have a small benchmark of the performance before/after for e.g., the Tracemonkey paper or another document that has many `beginText`/`endText` operators for verification?

src/core/evaluator.js
Snuffleupagus: It's a "better safe than sorry" approach really, since neither of the `BT`/`ET` operators *should* ever have parameters. However, considering the kind of checks that happen in the `default` branch, just always resetting the `args` seemed easiest; see https://github.com/mozilla/pdf.js/blob/f87dc42780d9aaefe3b3c01c1127d88d05ac5763/src/core/evaluator.js#L1180-L1194

Snuffleupagus:
> [...] but do you perhaps have a small benchmark of the performance before/after for e.g., the Tracemonkey paper or another document that has many `beginText`/`endText` operators for verification?

I've not bothered to do that, since finding a test-case with a *relatively* large number of `BT`/`ET` operators isn't that easy; if benchmarking is a merge blocker I suppose I'll have to try and find the time to do that.

My point above though is that handling the `BT`/`ET` cases explicitly cannot possibly be less efficient than letting them fallback to the `default` case, and based on the runtime of the `reftests` on the bots there doesn't appear to be any general upwards trend in runtimes (compared to other recent runs on the bots).

src/core/evaluator.js
timvandermeij: Right, but now that we're actually handling these cases explicitly we don't enter the `default` branch for them anymore, so unless I'm missing something here the `args` are not actually used anywhere.

timvandermeij:
> My point above though is that handling the BT/ET cases explicitly cannot possibly be less efficient than letting them fallback to the default case, and based on the runtime of the reftests on the bots there doesn't appear to be any general upwards trend in runtimes (compared to other recent runs on the bots).

Yes, that's a valid point. Let's finish the comment above and then it should be fine by me.

src/core/evaluator.js
Snuffleupagus: > [...] so unless I'm missing something here the `args` are not actually used anywhere.

That's absolutely correct, but I was a bit worried that https://github.com/mozilla/pdf.js/blob/f87dc42780d9aaefe3b3c01c1127d88d05ac5763/src/core/evaluator.js#L2991-L3004 could *somehow* cause a `BT`/`ET` operator to end up with an `args` array.
Given that `getOperatorList` always calls `EvaluatorPreprocessor.read` in as outlined in https://github.com/mozilla/pdf.js/blob/f87dc42780d9aaefe3b3c01c1127d88d05ac5763/src/core/evaluator.js#L2961-L2967 that worry seems to be unfounded though, so I'll simply remove those lines again.

Really good point!

timvandermeij (approved):

timvandermeij:
Thank you!

PR #10765, opened 2019-04-24
Add `passive: false` to the `wheel` event listener, to work-around broken default behaviour in Chrome 73 and above (issue 10761)

### LINKED ISSUE #10761: Ctrl-wheel zooming also scales UI in Chrome
Attach (recommended) or Link to PDF file here: Any PDF will work

Configuration:
- Web browser and its version: Chrome 73.0.3683.103 x64
- Operating system and its version: Windows 10 Pro x64
- PDF.js version: 2.0.673
- Is a browser extension: Yes

1. Open any PDF
2. Hold CTRL key and use mouse wheel to zoom in/out

Document is zoomed in/out

The PDF reader UI is also scaled. It appears that Chrome is also zooming the entire page, including the PDF reader UI, as well as the PDF reader also zooming the document.

In previous versions only the document would be zoomed.

image

Note how the PDF viewer UI is scaled in this image. It can also be made absurdly small by zooming out:

image

https://chrome.google.com/webstore/detail/pdf-viewer/oemmndcbldboiebfnladdacbdfmadadm

Snuffleupagus:
Let's try this, since it doesn't appear to break scrolling/zooming in IE11.

Fixes #10761

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1399,7 +1399,7 @@ let PDFViewerApplication = {
     };
 
     window.addEventListener('visibilitychange', webViewerVisibilityChange);
-    window.addEventListener('wheel', webViewerWheel);
+    window.addEventListener('wheel', webViewerWheel, { passive: false, });
     window.addEventListener('click', webViewerClick);
     window.addEventListener('keydown', webViewerKeyDown);
     window.addEventListener('resize', _boundEvents.windowResize);


timvandermeij (approved):

timvandermeij:
Thanks!

diegocr:
fwiw, in Chrome 84 (latest version at time of this write), this is causing a choppy scrolling with some PDFs, and this printed in the console:

```
[Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. Consider marking event handler as 'passive' to make the page more responsive.
```

Have changed this to `{ passive: true }` and it does resolves the issue for me.

PR #10771, opened 2019-04-27
[api-minor] Implement creation/modification date for annotations

timvandermeij:
This includes the information in the core and display layers. The date parsing logic from the document properties is rewritten according to the specification and now includes unit tests.

Moreover, missing unit tests for the color of a popup annotation have been added.

Finally the styling of the popup is changed slightly to make the text a bit smaller (it's currently quite large in comparison to other viewers) and to make the drop shadow a bit more subtle. The former is done to be able to easily include the modification date in the popup similar to how other viewers do this.

### DIFF
--- a/l10n/en-US/viewer.properties
+++ b/l10n/en-US/viewer.properties
@@ -226,6 +226,10 @@ invalid_file_error=Invalid or corrupted PDF file.
 missing_file_error=Missing PDF file.
 unexpected_response_error=Unexpected server response.
 
+# LOCALIZATION NOTE (annotation_date_string): "{{date}}" and "{{time}}" will be
+# replaced by the modification date, and time, of the annotation.
+annotation_date_string={{date}}, {{time}}
+
 # LOCALIZATION NOTE (text_annotation_type.alt): This is used as a tooltip.
 # "{{type}}" will be replaced with an annotation type from a list defined in
 # the PDF spec (32000-1:2008 Table 169 – Annotation types).
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -16,7 +16,7 @@
 
 import {
   AnnotationBorderStyleType, AnnotationFieldFlag, AnnotationFlag,
-  AnnotationType, OPS, stringToBytes, stringToPDFString, Util, warn
+  AnnotationType, isString, OPS, stringToBytes, stringToPDFString, Util, warn
 } from '../shared/util';
 import { Catalog, FileSpec, ObjectLoader } from './obj';
 import { Dict, isDict, isName, isRef, isStream } from './primitives';
@@ -176,6 +176,8 @@ class Annotation {
   constructor(params) {
     let dict = params.dict;
 
+    this.setCreationDate(dict.get('CreationDate'));
+    this.setModificationDate(dict.get('M'));
     this.setFlags(dict.get('F'));
     this.setRectangle(dict.getArray('Rect'));
     this.setColor(dict.getArray('C'));
@@ -187,8 +189,10 @@ class Annotation {
       annotationFlags: this.flags,
       borderStyle: this.borderStyle,
       color: this.color,
+      creationDate: this.creationDate,
       hasAppearance: !!this.appearance,
       id: params.id,
+      modificationDate: this.modificationDate,
       rect: this.rectangle,
       subtype: params.subtype,
     };
@@ -237,6 +241,31 @@ class Annotation {
       return false;
     }
     return this._isPrintable(this.flags);
+  }
+
+  /**
+   * Set the creation date.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {string} creationDate - PDF date string that indicates when the
+   *                                annotation was originally created
+   */
+  setCreationDate(creationDate) {
+    this.creationDate = isString(creationDate) ? creationDate : null;
+  }
+
+  /**
+   * Set the modification date.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {string} modificationDate - PDF date string that indicates when the
+   *                                    annotation was last modified
+   */
+  setModificationDate(modificationDate) {
+    this.modificationDate = isString(modificationDate) ?
+                            modificationDate : null;
   }
 
   /**
@@ -946,6 +975,20 @@ class PopupAnnotation extends Annotation {
     this.data.parentId = dict.getRaw('Parent').toString();
     this.data.title = stringToPDFString(parentItem.get('T') || '');
     this.data.contents = stringToPDFString(parentItem.get('Contents') || '');
+
+    if (!parentItem.has('CreationDate')) {
+      this.data.creationDate = null;
+    } else {
+      this.setCreationDate(parentItem.get('CreationDate'));
+      this.data.creationDate = this.creationDate;
+    }
+
+    if (!parentItem.has('M')) {
+      this.data.modificationDate = null;
+    } else {
+      this.setModificationDate(parentItem.get('M'));
+      this.data.modificationDate = this.modificationDate;
+    }
 
     if (!parentItem.has('C')) {
       // Fall back to the default background color.
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -14,7 +14,8 @@
  */
 
 import {
-  addLinkAttributes, DOMSVGFactory, getFilenameFromUrl, LinkTarget
+  addLinkAttributes, DOMSVGFactory, getFilenameFromUrl, LinkTarget,
+  PDFDateString
 } from './display_utils';
 import {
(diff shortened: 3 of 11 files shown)


src/display/display_utils.js
```diff
@@ -491,6 +491,82 @@ function releaseImageResources(img) {
   img.removeAttribute('src');
 }
 
+const pdfDateStringRegex = new RegExp(
```
Snuffleupagus: Drive-by comment: Perhaps create this lazily instead?

src/display/display_utils.js
```diff
@@ -491,6 +491,82 @@ function releaseImageResources(img) {
...
+  * @param {string} input
+  * @return {Date}
+  */
+  static toDateObject(input) {
+    // Optional fields that don't satisfy the requirements from the regular
+    // expression (such as incorrect digit counts or numbers that are out of
+    // range) will fall back the defaults from the specification.
+    const matches = pdfDateStringRegex.exec(input);
+    if (!matches) {
+      throw new Error('Prefix and year are required in the date input');
```
Snuffleupagus: Drive-by comment: Perhaps use `FormatError` instead here?

web/pdf_document_properties.js
```diff
@@ -366,44 +366,10 @@ class PDFDocumentProperties {
...
-    } else if (utRel === '+') {
-      hours -= offsetHours;
-      minutes -= offsetMinutes;
-    }
 
-    // Return the new date format from the user's locale.
-    let date = new Date(Date.UTC(year, month, day, hours, minutes, seconds));
-    let dateString = date.toLocaleDateString();
-    let timeString = date.toLocaleTimeString();
+    const dateObject = PDFDateString.toDateObject(inputDate);
```
Snuffleupagus: Drive-by comment: Probably need to catch errors here, since this can throw?

Snuffleupagus:
Looking briefly at the test results, a couple of things stand out:
 - The "title" now appears to be slightly *smaller* than the "contents", was that actually intentional?
 - There's a lot of, in my opinion too much, *bottom* padding/margin/etc. on both the "title" respectively the "contents" lines of the popups.
 - I'm assuming that `{{date}}, {{time}}` strings wasn't intentional here :-)

timvandermeij:
The review comments have been addressed, but I haven't been able to find a way to make localization work in the test driver, which is why we see `{{date}}, {{time}}` in the reference images. It's supposed to be a fallback string in case the localization doesn't provide one, and the `l10n` provider then inserts the actual arguments there. It looks like we simply don't have this in the test driver since we only call `render` on a canvas and don't have a PDF viewer active.

I don't really see this as a blocker for this PR because we already cover the date parsing in the unit tests. However, a follow-up ticket to find a solution to this seems like a good idea because we never had this use-case before.

We also need to fix the CI since after a recent Node.js update the `canvas` library isn't installable anymore. The prebuilt file is not available due to an ABI change and some packages are missing on the CI to be able to compile it ourselves. According to https://github.com/avxsynth/avxsynth/issues/89#issuecomment-11977268 we should be able to fix this by installing some extra packages, which sounds like a good idea to have a fallback in case the prebuilt files don't work.

src/display/annotation_layer.js
```diff
@@ -724,18 +728,36 @@ class PopupElement {
...
+    // The modification date is shown in the popup instead of the creation
+    // date if it is available and can be parsed correctly, which is
+    // consistent with other viewers such as Adobe Acrobat. The localization
+    // string is equal to the one used for the document properties dialog
+    // for consistency.
+    try {
+      const dateObject = PDFDateString.toDateObject(this.modificationDate);
+      const modificationDate = document.createElement('span');
+      modificationDate.textContent = '{{date}}, {{time}}';
+      modificationDate.dataset.l10nId = 'document_properties_date_string';
```
Snuffleupagus: Something I forgot to mention previously: It's probably best to add a *new* string, rather than overloading an existing one in way.

timvandermeij: Yes, that's probably best. The commit is updated.

Snuffleupagus: Sorry, I didn't mean remove the existing one, but rather adding an *additional* new one for this use-case since locales may want different semantics depending on context.

src/core/annotation.js
```diff
@@ -945,6 +974,10 @@ class PopupAnnotation extends Annotation {
     this.data.parentType = isName(parentSubtype) ? parentSubtype.name : null;
     this.data.parentId = dict.getRaw('Parent').toString();
     this.data.title = stringToPDFString(parentItem.get('T') || '');
+    this.data.creationDate = isString(parentItem.get('CreationDate')) ?
+                             parentItem.get('CreationDate') : null;
+    this.data.modificationDate = isString(parentItem.get('M')) ?
+                                 parentItem.get('M') : null;
```
Snuffleupagus: Wouldn't it be much better to handle these properties in the same way as the `C` one below instead? (That would also get rid of the double `Dict` lookups here.)

src/display/display_utils.js
```diff
@@ -491,6 +491,87 @@ function releaseImageResources(img) {
...
+        '\'?' // Trailing apostrophe (optional)
+      );
+    }
+
+    // Optional fields that don't satisfy the requirements from the regular
+    // expression (such as incorrect digit counts or numbers that are out of
+    // range) will fall back the defaults from the specification.
+    const matches = pdfDateStringRegex.exec(input);
+    if (!matches) {
+      throw new FormatError('Prefix and year are required in the date input');
```
Snuffleupagus: I'm starting to wonder if having this return e.g. `null` instead of throwing an `Error` wouldn't simplify the call-sites quite a bit!?

Snuffleupagus: Similarly, how about checking that the `input` actually is a (non-empty) string at the *top* of the method to avoid unnecessary parsing here?

src/display/annotation_layer.js
```diff
@@ -724,18 +728,34 @@ class PopupElement {
...
-    let contents = this._formatContents(this.contents);
     let title = document.createElement('h1');
     title.textContent = this.title;
+    popup.appendChild(title);
+
+    // The modification date is shown in the popup instead of the creation
+    // date if it is available and can be parsed correctly, which is
+    // consistent with other viewers such as Adobe Acrobat.
+    try {
+      const dateObject = PDFDateString.toDateObject(this.modificationDate);
```
Snuffleupagus: Please refer to the comment in the `toDateObject` method.

web/pdf_document_properties.js
```diff
@@ -363,50 +363,14 @@ class PDFDocumentProperties {
...
-
-    // Return the new date format from the user's locale.
-    let date = new Date(Date.UTC(year, month, day, hours, minutes, seconds));
-    let dateString = date.toLocaleDateString();
-    let timeString = date.toLocaleTimeString();
-    return this.l10n.get('document_properties_date_string',
-                         { date: dateString, time: timeString, },
-                         '{{date}}, {{time}}');
+    try {
+      const dateObject = PDFDateString.toDateObject(inputDate);
```
Snuffleupagus: Please refer to the comment in the `toDateObject` method.

l10n/en-US/viewer.properties
```diff
@@ -96,9 +96,6 @@ document_properties_subject=Subject:
 document_properties_keywords=Keywords:
 document_properties_creation_date=Creation Date:
 document_properties_modification_date=Modification Date:
-# LOCALIZATION NOTE (document_properties_date_string): "{{date}}" and "{{time}}"
```
Snuffleupagus: Please don't remove this, refer to https://github.com/mozilla/pdf.js/pull/10771#discussion_r280981723 for additional details.

Snuffleupagus (changes requested):
This looks generally good; however I do have a couple of comments.

Really sorry about not getting to this PR sooner, since that may have caused you some unnecessary work!

l10n/en-US/viewer.properties
timvandermeij: Fixed in the new commit.

web/pdf_document_properties.js
timvandermeij: Fixed in the new commit.

src/display/annotation_layer.js
timvandermeij: Fixed in the new commit.

src/display/display_utils.js
timvandermeij: Fixed in the new commit. This is indeed a bit cleaner.

src/core/annotation.js
timvandermeij: Fixed in the new commit. I also added a unit test for the `C` one below since that was missing.

Snuffleupagus (approved):
r=me with passing tests; thank you!

PR #10784, opened 2019-05-02
Ensure that the `OperatorList` constructor actually initializes a `NullOptimizer` when intended (PR 9089 follow-up)

Snuffleupagus:
It appears that this has been broken ever since PR #9089, which also introduced this code, since the `QueueOptimizer`/`NullOptimizer` choice was made based on the still undefined `this.intent` property.

Furthermore, fixing this also uncovered the fact that the `NullOptimizer.reset` method was missing.

### DIFF
--- a/src/core/operator_list.js
+++ b/src/core/operator_list.js
@@ -525,6 +525,8 @@ var NullOptimizer = (function NullOptimizerClosure() {
     },
 
     flush() { },
+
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
That's a really good find; thanks!

PR #10805, opened 2019-05-10
Enable the `consistent-return` ESLint rule

Snuffleupagus:
This rule is already enabled in mozilla-central, and helps ensure more consistent functions/methods, see https://searchfox.org/mozilla-central/rev/b9da45f63cb567244933c77b2c7e827a057d3f9b/tools/lint/eslint/eslint-plugin-mozilla/lib/configs/recommended.js#119-120

Please see https://eslint.org/docs/rules/consistent-return for additional information.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -365,7 +365,7 @@ function getTempFile(prefix, suffix) {
 
 function createTestSource(testsName, bot) {
   var source = stream.Readable({ objectMode: true, });
-  source._read = function () {
+  source._read = function() {
     console.log();
     console.log('### Running ' + testsName + ' tests');
 
@@ -409,6 +409,7 @@ function createTestSource(testsName, bot) {
     testProcess.on('close', function (code) {
       source.push(null);
     });
+    return undefined;
   };
   return source;
 }
@@ -1245,9 +1246,10 @@ gulp.task('gh-pages-prepare', function () {
 gulp.task('wintersmith', function (done) {
   var wintersmith = require('wintersmith');
   var env = wintersmith('docs/config.json');
-  env.build(GH_PAGES_DIR, function (error) {
+  env.build(GH_PAGES_DIR, function(error) {
     if (error) {
-      return done(error);
+      done(error);
+      return;
     }
     replaceInFile(GH_PAGES_DIR + '/getting_started/index.html',
                   /STABLE_VERSION/g, config.stableVersion);
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -302,12 +302,12 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
       if (!(w && isNum(w)) || !(h && isNum(h))) {
         warn('Image dimensions are missing, or not numbers.');
-        return;
+        return undefined;
       }
       var maxImageSize = this.options.maxImageSize;
       if (maxImageSize !== -1 && w * h > maxImageSize) {
         warn('Image exceeded maximum allowed size and was removed.');
-        return;
+        return undefined;
       }
 
       var imageMask = (dict.get('ImageMask', 'IM') || false);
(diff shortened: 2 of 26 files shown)


gulpfile.js
```diff
@@ -1245,9 +1246,10 @@ gulp.task('gh-pages-prepare', function () {
 gulp.task('wintersmith', function (done) {
   var wintersmith = require('wintersmith');
   var env = wintersmith('docs/config.json');
-  env.build(GH_PAGES_DIR, function (error) {
+  env.build(GH_PAGES_DIR, function(error) {
     if (error) {
-      return done(error);
+      done(error);
+      return;
```
timvandermeij: `return undefined`?

src/core/evaluator.js
```diff
@@ -460,6 +461,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                                               [objId, 'FontType3Res', null]);
         }
         this.handler.send('obj', [objId, this.pageIndex, 'Image', null]);
+        return undefined;
```
timvandermeij: If we return in both the `then` and `catch` clauses, isn't all code below dead code now then?

web/download_manager.js
```diff
@@ -58,8 +58,8 @@ class DownloadManager {
 
   downloadData(data, filename, contentType) {
     if (navigator.msSaveBlob) { // IE10 and above
-      return navigator.msSaveBlob(new Blob([data], { type: contentType, }),
-                                  filename);
+      navigator.msSaveBlob(new Blob([data], { type: contentType, }), filename);
+      return;
```
timvandermeij: `return undefined`?

web/pdf_print_service.js
```diff
@@ -298,6 +298,7 @@ if (hasAttachEvent) {
       event.keyCode = 0;
       return false;
     }
+    return undefined;
```
timvandermeij: There are more `return` statements without `undefined` above. Why are they not picked up?

Also, shouldn't this return a boolean like we do in the `if` statement?

gulpfile.js
Snuffleupagus: Nope, because of "This rule requires `return` statements to either always or never specify values." as explained in https://eslint.org/docs/rules/consistent-return#rule-details.

Keep in mind that functions always, implicitly, returns `undefined` unless something else is specified.

src/core/evaluator.js
Snuffleupagus: Sorry, but I really don't follow. This is inside of a `Promise`, hence this only explicitly spells out the return value (as opposed to implicitly previously).

web/download_manager.js
Snuffleupagus: Please sse https://github.com/mozilla/pdf.js/pull/10805#discussion_r283094312

web/pdf_print_service.js
Snuffleupagus: Again, the key is consistent `return`s here; so https://github.com/mozilla/pdf.js/pull/10805#discussion_r283094312 here too :-)

I'm not sure that I really understand this code, but I suppose that there *could* be a difference between returning `false` and `undefined` in an `attachEvent`.
However, I think that I'll just disable that particular one since I don't want to unknowingly cause any regressions here.

gulpfile.js
timvandermeij: D'oh, I misread how the rule works. Thanks.

src/core/evaluator.js
timvandermeij: Yes, it's fine. I didn't see that this was inside a `Promise`.

timvandermeij:
Looks good. Thank you!

termux-io:
`173fbef05b3c88cc5a6e3448ca6e19c5b8b86713`

PR #10820, opened 2019-05-14
Annotations - Added parsing of IRT, RT, State and StateModel

### LINKED ISSUE #10677: Contents of annotations not (completely) visible

vlastimilmaca:
Hi, this is a last piece of #10716 .

This PR adds parsing of IRT, RT and IT into MarkupAnnotation. 
If RT is Group, it takes all properties from IRT annotation and sets them into current annotation, which is what should happen according to spec (see page 393, in the middle). This also happens for popups, which should fix #10677, where popup on strikeout annotation should have correct 'Contents' text. 

Parsing of contents has been moved from Markup annotation into Annotation, because according to spec, 'Contents' is common to all Annotations (see page 383, table 184).

IT is also parsed, but I'm not sure about it's meaning or usability in UI, but its easy to parse, so it is there …

There is also added 'isMarkup' property into Markup annotation data set, which should help in building new UI features using RT/IRT (discussion threads, states etc). 
ie. if isMarkup == true, it should be displayed in comments sidebar (if thats gonna be a thing), otherwise it should be skipped for that part of UI.

Finally there is parsing of State and StateModel in TextAnnotation, which is also preparation for new UI features, but it's not gonna impact anything for now.

Future annotation UI update should probably solve these two main issues:
- displaying replies inside of popup of IRT annotation or inside of other UI element (maybe new sidebar, or new "extended" popup for IRT annotation)
- not rendering annotations with RT=R or StateModel into canvas as they should be represented inside their IRT annotation's UI element.

Btw. As I was going through spec again, I noticed, that 'CreationDate' is a thing only for Markup Annotation, so maybe it should be moved too…

Thank you for reviews :).

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -16,8 +16,8 @@
 
 import {
   AnnotationBorderStyleType, AnnotationFieldFlag, AnnotationFlag,
-  AnnotationType, assert, isString, OPS, stringToBytes, stringToPDFString, Util,
-  warn
+  AnnotationReplyType, AnnotationType, assert, isString, OPS, stringToBytes,
+  stringToPDFString, Util, warn
 } from '../shared/util';
 import { Catalog, FileSpec, ObjectLoader } from './obj';
 import { Dict, isDict, isName, isRef, isStream } from './primitives';
@@ -643,16 +643,61 @@ class MarkupAnnotation extends Annotation {
     super(parameters);
 
     const dict = parameters.dict;
-    if (!dict.has('C')) {
-      // Fall back to the default background color.
-      this.data.color = null;
+
+    if (dict.has('IRT')) {
+      const rawIRT = dict.getRaw('IRT');
+      this.data.inReplyTo = isRef(rawIRT) ? rawIRT.toString() : null;
+
+      const rt = dict.get('RT');
+      this.data.replyType = isName(rt) ? rt.name : AnnotationReplyType.REPLY;
     }
 
-    this.setCreationDate(dict.get('CreationDate'));
-    this.data.creationDate = this.creationDate;
+    if (this.data.replyType === AnnotationReplyType.GROUP) {
+      // Subordinate annotations in a group should inherit
+      // the group attributes from the primary annotation.
+      const parent = dict.get('IRT');
 
-    this.data.hasPopup = dict.has('Popup');
-    this.data.title = stringToPDFString(dict.get('T') || '');
+      this.data.title = stringToPDFString(parent.get('T') || '');
+
+      this.setContents(parent.get('Contents'));
+      this.data.contents = this.contents;
+
+      if (!parent.has('CreationDate')) {
+        this.data.creationDate = null;
+      } else {
+        this.setCreationDate(parent.get('CreationDate'));
+        this.data.creationDate = this.creationDate;
+      }
+
+      if (!parent.has('M')) {
+        this.data.modificationDate = null;
+      } else {
+        this.setModificationDate(parent.get('M'));
+        this.data.modificationDate = this.modificationDate;
+      }
+
+      this.data.hasPopup = parent.has('Popup');
+
+      if (!parent.has('C')) {
+        // Fall back to the default background color.
+        this.data.color = null;
+      } else {
+        this.setColor(parent.getArray('C'));
+        this.data.color = this.color;
+      }
+    } else {
+      this.data.title = stringToPDFString(dict.get('T') || '');
+
+      this.setCreationDate(dict.get('CreationDate'));
+      this.data.creationDate = this.creationDate;
+
+      this.data.hasPopup = dict.has('Popup');
+
+      if (!dict.has('C')) {
+        // Fall back to the default background color.
+        this.data.color = null;
+      }
+    }
   }
 
   /**
@@ -969,6 +1014,7 @@ class TextAnnotation extends MarkupAnnotation {
 
     super(parameters);
 
+    const dict = parameters.dict;
     this.data.annotationType = AnnotationType.TEXT;
 
     if (this.data.hasAppearance) {
@@ -976,10 +1022,17 @@ class TextAnnotation extends MarkupAnnotation {
     } else {
       this.data.rect[1] = this.data.rect[3] - DEFAULT_ICON_SIZE;
       this.data.rect[2] = this.data.rect[0] + DEFAULT_ICON_SIZE;
-      this.data.name = parameters.dict.has('Name') ?
-                       parameters.dict.get('Name').name : 'Note';
+      this.data.name = dict.has('Name') ?
+                       dict.get('Name').name : 'Note';
     }
 
+    if (dict.has('State')) {
+      this.data.state = dict.get('State') || null;
+      this.data.stateModel = dict.get('StateModel') || null;
+    } else {
+      this.data.state = null;
+      this.data.stateModel = null;
+    }
   }
 }
 
@@ -1012,9 +1065,15 @@ class PopupAnnotation extends Annotation {
 
     let parentSubtype = parentItem.get('Subtype');
     this.data.parentType = isName(parentSubtype) ? parentSubtype.name : null;
-    this.data.parentId = dict.getRaw('Parent').toString();
-    this.data.title = stringToPDFString(parentItem.get('T') || '');
-    this.data.contents = stringToPDFString(parentItem.get('Contents') || '');
+    const rawParent = dict.getRaw('Parent');
+    this.data.parentId = isRef(rawParent) ? rawParent.toString() : null;
+
+    const rt = parentItem.get('RT');
+    if (isName(rt, AnnotationReplyType.GROUP)) {
+      // Subordinate annotations in a group should inherit
+      // the group attributes from the primary annotation.
+      parentItem = parentItem.get('IRT');
+    }
 
     if (!parentItem.has('M')) {
       this.data.modificationDate = null;
@@ -1040,6 +1099,9 @@ class PopupAnnotation extends Annotation {
         this.setFlags(parentFlags);
       }
     }
+
+    this.data.title = stringToPDFString(parentItem.get('T') || '');
+    this.data.contents = stringToPDFString(parentItem.get('Contents') || '');
   }
 }
 
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -86,6 +86,29 @@ const AnnotationType = {
   REDACT: 26,
 };
 
+const AnnotationStateModelType = {
+  MARKED: 'Marked',
+  REVIEW: 'Review',
+};
+
+const AnnotationMarkedState = {
+  MARKED: 'Marked',
+  UNMARKED: 'Unmarked',
+};
+
+const AnnotationReviewState = {
+  ACCEPTED: 'Accepted',
+  REJECTED: 'Rejected',
+  CANCELLED: 'Cancelled',
+  COMPLETED: 'Completed',
+  NONE: 'None',
+};
+
+const AnnotationReplyType = {
+  GROUP: 'Group',
+  REPLY: 'R',
+};
+
 const AnnotationFlag = {
   INVISIBLE: 0x01,
   HIDDEN: 0x02,
@@ -910,6 +933,10 @@ export {
   AnnotationBorderStyleType,
   AnnotationFieldFlag,
   AnnotationFlag,
+  AnnotationMarkedState,
+  AnnotationReplyType,
+  AnnotationReviewState,
+  AnnotationStateModelType,
   AnnotationType,
   FontType,
   ImageKind,
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -343,6 +343,220 @@ describe('annotation', function() {
 
       expect(markupAnnotation.creationDate).toEqual(null);
     });
+
+    it('should not parse IRT/RT when not defined', function (done) {
+      dict.set('Type', Name.get('Annot'));
+      dict.set('Subtype', Name.get('Text'));
+
+      const xref = new XRefMock([
+        { ref, data: dict, },
+      ]);
+
+      AnnotationFactory.create(xref, ref, pdfManagerMock,
+          idFactoryMock).then(({ data, }) => {
+        expect(data.inReplyTo).toBeUndefined();
+        expect(data.replyType).toBeUndefined();
+        done();
+      }, done.fail);
+    });
+
+    it('should parse IRT and set default RT when not defined.',
+        function (done) {
+      const annotationRef = new Ref(819, 0);
+      const annotationDict = new Dict();
+      annotationDict.set('Type', Name.get('Annot'));
+      annotationDict.set('Subtype', Name.get('Text'));
+
+      const replyRef = new Ref(820, 0);
+      const replyDict = new Dict();
+      replyDict.set('Type', Name.get('Annot'));
+      replyDict.set('Subtype', Name.get('Text'));
+      replyDict.set('IRT', annotationRef);
+
+      const xref = new XRefMock([
+        { ref: annotationRef, data: annotationDict, },
+        { ref: replyRef, data: replyDict, }
+      ]);
+      annotationDict.assignXref(xref);
+      replyDict.assignXref(xref);
+
+      AnnotationFactory.create(xref, replyRef, pdfManagerMock,
+          idFactoryMock).then(({ data, }) => {
+        expect(data.inReplyTo).toEqual(annotationRef.toString());
+        expect(data.replyType).toEqual('R');
+        done();
+      }, done.fail);
+    });
+
+    it('should parse IRT/RT for a group type', function (done) {
+      const annotationRef = new Ref(819, 0);
+      const annotationDict = new Dict();
+      annotationDict.set('Type', Name.get('Annot'));
+      annotationDict.set('Subtype', Name.get('Text'));
+      annotationDict.set('T', 'ParentTitle');
+      annotationDict.set('Contents', 'ParentText');
+      annotationDict.set('CreationDate', 'D:20180423');
+      annotationDict.set('M', 'D:20190423');
+      annotationDict.set('C', [0, 0, 1]);
+
+      const popupRef = new Ref(820, 0);
+      const popupDict = new Dict();
+      popupDict.set('Type', Name.get('Annot'));
+      popupDict.set('Subtype', Name.get('Popup'));
+      popupDict.set('Parent', annotationRef);
+      annotationDict.set('Popup', popupRef);
+
+      const replyRef = new Ref(821, 0);
+      const replyDict = new Dict();
+      replyDict.set('Type', Name.get('Annot'));
+      replyDict.set('Subtype', Name.get('Text'));
+      replyDict.set('IRT', annotationRef);
+      replyDict.set('RT', Name.get('Group'));
+      replyDict.set('T', 'ReplyTitle');
+      replyDict.set('Contents', 'ReplyText');
+      replyDict.set('CreationDate', 'D:20180523');
+      replyDict.set('M', 'D:20190523');
+      replyDict.set('C', [0.4]);
+
+      const xref = new XRefMock([
+        { ref: annotationRef, data: annotationDict, },
+        { ref: popupRef, data: popupDict, },
+        { ref: replyRef, data: replyDict, }
+      ]);
+      annotationDict.assignXref(xref);
+      popupDict.assignXref(xref);
+      replyDict.assignXref(xref);
+
+      AnnotationFactory.create(xref, replyRef, pdfManagerMock,
+          idFactoryMock).then(({ data, }) => {
+        expect(data.inReplyTo).toEqual(annotationRef.toString());
+        expect(data.replyType).toEqual('Group');
+        expect(data.title).toEqual('ParentTitle');
+        expect(data.contents).toEqual('ParentText');
+        expect(data.creationDate).toEqual('D:20180423');
+        expect(data.modificationDate).toEqual('D:20190423');
+        expect(data.color).toEqual(new Uint8ClampedArray([0, 0, 255]));
+        expect(data.hasPopup).toEqual(true);
+        done();
+      }, done.fail);
+    });
+
+    it('should parse IRT/RT for a reply type', function (done) {
+      const annotationRef = new Ref(819, 0);
+      const annotationDict = new Dict();
+      annotationDict.set('Type', Name.get('Annot'));
+      annotationDict.set('Subtype', Name.get('Text'));
+      annotationDict.set('T', 'ParentTitle');
+      annotationDict.set('Contents', 'ParentText');
+      annotationDict.set('CreationDate', 'D:20180423');
+      annotationDict.set('M', 'D:20190423');
+      annotationDict.set('C', [0, 0, 1]);
+
+      const popupRef = new Ref(820, 0);
+      const popupDict = new Dict();
+      popupDict.set('Type', Name.get('Annot'));
+      popupDict.set('Subtype', Name.get('Popup'));
+      popupDict.set('Parent', annotationRef);
+      annotationDict.set('Popup', popupRef);
+
+      const replyRef = new Ref(821, 0);
+      const replyDict = new Dict();
+      replyDict.set('Type', Name.get('Annot'));
+      replyDict.set('Subtype', Name.get('Text'));
+      replyDict.set('IRT', annotationRef);
+      replyDict.set('RT', Name.get('R'));
+      replyDict.set('T', 'ReplyTitle');
+      replyDict.set('Contents', 'ReplyText');
+      replyDict.set('CreationDate', 'D:20180523');
+      replyDict.set('M', 'D:20190523');
+      replyDict.set('C', [0.4]);
+
+      const xref = new XRefMock([
+        { ref: annotationRef, data: annotationDict, },
+        { ref: popupRef, data: popupDict, },
+        { ref: replyRef, data: replyDict, }
+      ]);
+      annotationDict.assignXref(xref);
+      popupDict.assignXref(xref);
+      replyDict.assignXref(xref);
+
+      AnnotationFactory.create(xref, replyRef, pdfManagerMock,
+          idFactoryMock).then(({ data, }) => {
+        expect(data.inReplyTo).toEqual(annotationRef.toString());
+        expect(data.replyType).toEqual('R');
+        expect(data.title).toEqual('ReplyTitle');
+        expect(data.contents).toEqual('ReplyText');
+        expect(data.creationDate).toEqual('D:20180523');
+        expect(data.modificationDate).toEqual('D:20190523');
+        expect(data.color).toEqual(new Uint8ClampedArray([102, 102, 102]));
+        expect(data.hasPopup).toEqual(false);
+        done();
+      }, done.fail);
+    });
+  });
+
+  describe('TextAnnotation', function() {
+    it('should not parse state model and state when not defined',
+        function (done) {
+      const annotationRef = new Ref(819, 0);
+      const annotationDict = new Dict();
+      annotationDict.set('Type', Name.get('Annot'));
+      annotationDict.set('Subtype', Name.get('Text'));
+      annotationDict.set('Contents', 'TestText');
+
+      const replyRef = new Ref(820, 0);
+      const replyDict = new Dict();
+      replyDict.set('Type', Name.get('Annot'));
+      replyDict.set('Subtype', Name.get('Text'));
+      replyDict.set('IRT', annotationRef);
+      replyDict.set('RT', Name.get('R'));
+      replyDict.set('Contents', 'ReplyText');
+
+      const xref = new XRefMock([
+        { ref: annotationRef, data: annotationDict, },
+        { ref: replyRef, data: replyDict, }
+      ]);
+      annotationDict.assignXref(xref);
+      replyDict.assignXref(xref);
+
+      AnnotationFactory.create(xref, replyRef, pdfManagerMock,
+          idFactoryMock).then(({ data, }) => {
+        expect(data.stateModel).toBeNull();
+        expect(data.state).toBeNull();
+        done();
+      }, done.fail);
+    });
+
+    it('should correctly parse state model and state when defined',
+        function (done) {
+      const annotationRef = new Ref(819, 0);
+      const annotationDict = new Dict();
+      annotationDict.set('Type', Name.get('Annot'));
+      annotationDict.set('Subtype', Name.get('Text'));
+
+      const replyRef = new Ref(820, 0);
+      const replyDict = new Dict();
+      replyDict.set('Type', Name.get('Annot'));
+      replyDict.set('Subtype', Name.get('Text'));
+      replyDict.set('IRT', annotationRef);
+      replyDict.set('RT', Name.get('R'));
+      replyDict.set('StateModel', 'Review');
+      replyDict.set('State', 'Rejected');
+
+      const xref = new XRefMock([
+        { ref: annotationRef, data: annotationDict, },
+        { ref: replyRef, data: replyDict, }
+      ]);
+      annotationDict.assignXref(xref);
+      replyDict.assignXref(xref);
+
+      AnnotationFactory.create(xref, replyRef, pdfManagerMock,
+          idFactoryMock).then(({ data, }) => {
+        expect(data.stateModel).toEqual('Review');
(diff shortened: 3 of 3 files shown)


vlastimilmaca:
I'll fix this after #10833 is merged, so it is clean. Please try to look in to this if there is anything else I should fix. I'm quite short on free time these days ...

src/core/annotation.js
```diff
@@ -617,14 +623,63 @@ class MarkupAnnotation extends Annotation {
     super(parameters);
     const dict = parameters.dict;
 
-    if (!dict.has('C')) {
-      // Fall back to the default background color.
-      this.data.color = null;
+    this.data.isMarkup = true;
+
+    if (dict.has('IT')) {
```
timvandermeij: If I read the specification correctly, this shouldn't be necessary for the IRT/RT logic, so if that's the case please remove it from this patch to keep the functionality limited to IRT/RT. If need be, it can later be added in a follow-up patch.

src/core/annotation.js
```diff
@@ -617,14 +623,63 @@ class MarkupAnnotation extends Annotation {
...
+      this.data.intent = dict.get('IT').name;
+    }
+
+    if (dict.has('IRT')) {
+      this.data.inReplyTo = dict.getRaw('IRT').toString();
+      this.data.replyType =
+      dict.has('RT') ? dict.get('RT').name : AnnotationReplyType.REPLY;
```
timvandermeij: This needs to be indented with two extra spaces.

src/core/annotation.js
```diff
@@ -617,14 +623,63 @@ class MarkupAnnotation extends Annotation {
...
     }
 
-    this.data.hasPopup = dict.has('Popup');
-    this.data.title = stringToPDFString(dict.get('T') || '');
-    this.data.contents = stringToPDFString(dict.get('Contents') || '');
+    if (this.data.replyType === AnnotationReplyType.GROUP) {
+      const parent = dict.get('IRT');
+
+      this.data.title = stringToPDFString(parent.get('T') || '');
+      this.data.subject = stringToPDFString(parent.get('Subj') || '');
```
timvandermeij: Also here, please remove any properties that weren't in the original code and aren't used anywhere yet to keep the patch small. The same applies to the `else` branch below for `subject`.

src/core/annotation.js
```diff
@@ -617,14 +623,63 @@ class MarkupAnnotation extends Annotation {
...
+      }
+
+      if (!parent.has('M')) {
+        this.data.modificationDate = null;
+      } else {
+        this.setModificationDate(parent.get('M'));
+        this.data.modificationDate = this.modificationDate;
+      }
+
+      // parent's popup should be used
```
timvandermeij: You can remove this comment since that's already clear from the code itself.

src/core/annotation.js
```diff
@@ -617,14 +623,63 @@ class MarkupAnnotation extends Annotation {
...
+      this.data.hasPopup = parent.has('Popup');
+
+      if (!parent.has('C')) {
+        // Fall back to the default background color.
+        this.data.color = null;
+      } else {
+        this.setColor(parent.getArray('C'));
+        this.data.color = this.color;
+      }
+
```
timvandermeij: Please remove this newline.

src/core/annotation.js
```diff
@@ -929,17 +984,22 @@ class TextAnnotation extends MarkupAnnotation {
...
     } else {
       this.data.rect[1] = this.data.rect[3] - DEFAULT_ICON_SIZE;
       this.data.rect[2] = this.data.rect[0] + DEFAULT_ICON_SIZE;
-      this.data.name = parameters.dict.has('Name') ?
-                       parameters.dict.get('Name').name : 'Note';
+      this.data.name = dict.has('Name') ?
+                       dict.get('Name').name : 'Note';
     }
 
+    if (dict.has('State')) {
```
timvandermeij: The spec states: "The state is not specified in the annotation itself but in a separate text annotation that refers to the original annotation by means of its IRT (“in reply to”) entry."

Should we check if an `IRT` entry is available here? Moreover, should this text annotation be rendered (which I think is the case at the moment) or should those annotations not be rendered because they only contain the state of another annotation? I think it's rather odd that the spec uses a different annotation for keeping state of another annotation, let alone it being a `Text` annotation, but that seems to be how it works...

Whether or not an annotation should be rendered is controlled in https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js#L358-L359, so you could add a check for `parameters.data.state !== null` there.

Finally, I think the data attributes for an annotation type should always be provided and set to `null` if they are not used. This way the API is consistent: you always get all attributes on which you can do a check. In this case, the `state` and `stateModel` properties would alway be provided, but set to `null` if they don't have a value.

src/core/annotation.js
```diff
@@ -973,8 +1033,15 @@ class PopupAnnotation extends Annotation {
     let parentSubtype = parentItem.get('Subtype');
     this.data.parentType = isName(parentSubtype) ? parentSubtype.name : null;
     this.data.parentId = dict.getRaw('Parent').toString();
-    this.data.title = stringToPDFString(parentItem.get('T') || '');
-    this.data.contents = stringToPDFString(parentItem.get('Contents') || '');
+
+    if (parentItem.has('RT')) {
+      if ((parentItem.get('RT').name ||
```
timvandermeij: Let's simplify this to just `if (parentItem.get('RT').name === AnnotationReplyType.GROUP) {` because `RT` must always exist because of the `has` check above.

src/core/annotation.js
```diff
@@ -617,14 +623,63 @@ class MarkupAnnotation extends Annotation {
...
+      this.data.inReplyTo = dict.getRaw('IRT').toString();
+      this.data.replyType =
+      dict.has('RT') ? dict.get('RT').name : AnnotationReplyType.REPLY;
     }
 
-    this.data.hasPopup = dict.has('Popup');
-    this.data.title = stringToPDFString(dict.get('T') || '');
-    this.data.contents = stringToPDFString(dict.get('Contents') || '');
+    if (this.data.replyType === AnnotationReplyType.GROUP) {
+      const parent = dict.get('IRT');
```
timvandermeij: Let's add a comment after the `if` statement: "Subordinate annotations in a group should inherit the group attributes from the primary annotation." (taken from the spec)

src/core/annotation.js
```diff
@@ -973,8 +1033,15 @@ class PopupAnnotation extends Annotation {
     let parentSubtype = parentItem.get('Subtype');
     this.data.parentType = isName(parentSubtype) ? parentSubtype.name : null;
     this.data.parentId = dict.getRaw('Parent').toString();
-    this.data.title = stringToPDFString(parentItem.get('T') || '');
-    this.data.contents = stringToPDFString(parentItem.get('Contents') || '');
+
+    if (parentItem.has('RT')) {
+      if ((parentItem.get('RT').name ||
+        AnnotationReplyType.REPLY) === AnnotationReplyType.GROUP) {
+        // If parent item is group reply type,
```
timvandermeij: Let's replace this comment with: "Subordinate annotations in a group should inherit the group attributes from the primary annotation." (taken from the spec)

src/shared/util.js
```diff
@@ -910,7 +933,11 @@ export {
   AnnotationBorderStyleType,
   AnnotationFieldFlag,
   AnnotationFlag,
+  AnnotationMarkedState,
+  AnnotationStateModelType,
   AnnotationType,
+  AnnotationReplyType,
```
timvandermeij: Just for readability, could you put these new entries in the correct (alphabetical) order?

timvandermeij (changes requested):
Looks good! If you rebase onto the current master and incorporate these comments, you can simplify the patch even further. I only haven't looked at the tests yet because I think they will also change a bit after the rebase and comments, but I will look at those when that is done. Thank you!

src/core/annotation.js
Snuffleupagus: *I'm going to comment on this __once__ here, but this problem exists in many places.*

Just because a particular object exists in a `Dict`, that does *not* imply that it's actually of the desired type (i.e. `Name` here). Hence you need to check that as well, before attempting to access a potentially non-existent `name` property, since there will potentially be errors thrown otherwise.

src/core/annotation.js
```diff
@@ -617,14 +623,63 @@ class MarkupAnnotation extends Annotation {
...
-      // Fall back to the default background color.
-      this.data.color = null;
+    this.data.isMarkup = true;
+
+    if (dict.has('IT')) {
+      this.data.intent = dict.get('IT').name;
+    }
+
+    if (dict.has('IRT')) {
+      this.data.inReplyTo = dict.getRaw('IRT').toString();
```
Snuffleupagus: Similar to my other comment, you need to actually validate this data before calling `toString()` on it to avoid errors.

src/core/annotation.js
vlastimilmaca: Yeah this is tricky … Maybe, as there is viewable and printable in Annotation base class, there should also exist renderable, which will tell, wheter annotation should be rendered into canvas ([added into opList (see code)](https://github.com/mozilla/pdf.js/blob/master/src/core/document.js#L243) ). Viewable will then tell, wheter it can be passed out of worker at all ([added into getAnnotationsData(see code)](https://github.com/mozilla/pdf.js/blob/master/src/core/document.js#L299) response). 

annotation.viewable is now used in both of those cases, but IMO it should be splitted. So in core/document.js there should exist isAnnotationRenderable as is now (which will check for annotation.renderable - for opList) and isAnnotationViewable (which will check for annotation.viewable - for getAnnotationsData). 

RT=R and State annotations would be renderable=false, but viewable=true.  Standard annotations and Group annotations (probably?) would be true,true. And those, which are now viewable=false would be false,false. But I would rather keep that into some future PRs as it might break a thing or two ...

src/core/annotation.js
```diff
@@ -1048,13 +1048,14 @@ class PopupAnnotation extends Annotation {
 
     let parentSubtype = parentItem.get('Subtype');
     this.data.parentType = isName(parentSubtype) ? parentSubtype.name : null;
-    this.data.parentId = dict.getRaw('Parent').toString();
+    const rawParent = dict.getRaw('Parent');
+    this.data.parentId = rawParent ? rawParent.toString() : null;
```
vlastimilmaca: I'm not sure, if this check is OK.

vlastimilmaca:
Thank you for checking this! I'm sorry that I didn't find time earlier as you probably already forgot what this PR was about :( (I almost did). I hope I addressed all requests, but I'm not sure if I correctly handled @Snuffleupagus request for checking getRaw() return values. If you have an idea on how to make it better, please let me know. Thanks !

PS. I'll squash it when it's ok.

src/core/annotation.js
```diff
@@ -617,14 +623,63 @@ class MarkupAnnotation extends Annotation {
     super(parameters);
     const dict = parameters.dict;
 
-    if (!dict.has('C')) {
-      // Fall back to the default background color.
-      this.data.color = null;
+    this.data.isMarkup = true;
```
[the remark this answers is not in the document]
vlastimilmaca: Removed.

src/core/annotation.js
```diff
@@ -633,16 +626,24 @@ class MarkupAnnotation extends Annotation {
     this.data.title = stringToPDFString(dict.get('T') || '');
 
     if (dict.has('IRT')) {
-      this.data.inReplyTo = dict.getRaw('IRT').toString();
-      this.data.replyType =
-      dict.has('RT') ? dict.get('RT').name : AnnotationReplyType.REPLY;
+      const rawIRT = dict.getRaw('IRT');
```
Snuffleupagus: What type is this supposed to be?

If `Ref`, then you need to check `isRef(...)` below; similar for other types of course.

src/core/annotation.js
```diff
@@ -633,16 +626,24 @@ class MarkupAnnotation extends Annotation {
...
+      this.data.inReplyTo = rawIRT ? rawIRT.toString() : null;
+
+      if (dict.has('RT')) {
```
Snuffleupagus: This looks way too complicated, why can't you simply replace all of this block with *just* the following instead?
```javascript
const rt = dict.get('RT');
this.data.replyType = isName(rt) ? rt.name : AnnotationReplyType.REPLY;
```

src/core/annotation.js
Snuffleupagus: Same as above.

Snuffleupagus: See comment above.

src/core/annotation.js
```diff
@@ -1048,13 +1048,14 @@ class PopupAnnotation extends Annotation {
...
+    const rawParent = dict.getRaw('Parent');
+    this.data.parentId = rawParent ? rawParent.toString() : null;
 
     if (parentItem.has('RT')) {
-      if ((parentItem.get('RT').name ||
-        AnnotationReplyType.REPLY) === AnnotationReplyType.GROUP) {
-        // If parent item is group reply type,
-        // we have to take data of it's master(IRT)
+      const rt = parentItem.get('RT');
+      if ((isName(rt) ? rt.name : null) === AnnotationReplyType.GROUP) {
```
Snuffleupagus: This can be made a lot simpler, if you look at the [definition of the `isName` helper function](https://github.com/mozilla/pdf.js/blob/d0892e46e2b0d16ecd7f720f36f36202ae1c12cd/src/core/primitives.js#L284-L286):
```javascript
if (isName(parentItem.get('RT'), AnnotationReplyType.GROUP)) {
```

src/core/annotation.js
vlastimilmaca: Honestly, I'm not sure … it's the same thing as parent in popup, so I used same approach as the one that is already in master ([https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L989](https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L989)). But it is probably Ref … so I do it like this? 
isRef(dict.get("IRT")) ? dict.getRaw("IRT").toString() : null

src/core/annotation.js
```diff
@@ -1048,13 +1048,14 @@ class PopupAnnotation extends Annotation {
 
     let parentSubtype = parentItem.get('Subtype');
     this.data.parentType = isName(parentSubtype) ? parentSubtype.name : null;
-    this.data.parentId = dict.getRaw('Parent').toString();
+    const rawParent = dict.getRaw('Parent');
+    this.data.parentId = rawParent ? rawParent.toString() : null;
 
     if (parentItem.has('RT')) {
```
Snuffleupagus: This can be made a lot simpler, and the `.has(...)` check is unnecessary here as well, if you look at the [definition of the `isName` helper function](https://github.com/mozilla/pdf.js/blob/d0892e46e2b0d16ecd7f720f36f36202ae1c12cd/src/core/primitives.js#L284-L286):
```javascript
const rt = parentItem.get('RT');
if (isName(rt, AnnotationReplyType.GROUP)) {
  ...
}
```

src/core/annotation.js
vlastimilmaca: I was not sure what get('RT') would do, if 'RT' is not there, but if that's ok, than it can be as you suggested :). Will do.

src/core/annotation.js
vlastimilmaca: That's cool, thank you :) ... didn't cross my mind that it can have two arguments.

src/core/annotation.js
Snuffleupagus: Please check the code, and you'll see what it returns in that case: https://github.com/mozilla/pdf.js/blob/d0892e46e2b0d16ecd7f720f36f36202ae1c12cd/src/core/primitives.js#L83-L97

src/core/annotation.js
Snuffleupagus: > isRef(dict.get("IRT")) ? dict.getRaw("IRT").toString() : null

That won't work at all since `dict.get` dereferences `Ref` objects, whereas `dict.getRaw` doesn't as its name suggests.
```javascript
const rawIRT = dict.getRaw('IRT');
this.data.inReplyTo = isRef(rawIRT) ? rawIRT.toString() : null;
```

src/shared/util.js
```diff
@@ -86,6 +86,29 @@ const AnnotationType = {
   REDACT: 26,
 };
 
+const AnnotationStateModelType = {
```
Snuffleupagus: Any particular reason for not making all of these added structures enums?

vlastimilmaca: They are stored as strings in annotations, which is what I need to compare it against. Javascript can do string enums?

src/core/annotation.js
vlastimilmaca: Yeah … thats gonna be undefined … so !isName … so it is ok :).

src/core/annotation.js
vlastimilmaca: I'm sorry ... I'm really tired ... after I looked into get method you showed me, it hit me :). Thanks.

src/shared/util.js
vlastimilmaca: Should I also order it alphabetically, or keep it together in 'logic' order like this? Move it somewhere ?

src/core/annotation.js
vlastimilmaca: You are right. Removed.

src/core/annotation.js
vlastimilmaca: Removed.

src/core/annotation.js
timvandermeij: That sounds like a good approach, but I agree it's best to keep that for a follow-up PR given its complexity.

src/shared/util.js
timvandermeij: I don't think the order matters much here, but it's usally nice to keep the `export` list at the bottom ordered so we can quickly spot if something is exported or not.

test/unit/annotation_spec.js
```diff
@@ -231,6 +231,243 @@ describe('annotation', function() {
 
       expect(annotation.color).toEqual(new Uint8ClampedArray([0, 0, 0]));
     });
+
+    it('should parse Contents', function () {
```
timvandermeij: This unit test can be removed now as it was included in the PR you rebased upon.

test/unit/annotation_spec.js
```diff
@@ -231,6 +231,243 @@ describe('annotation', function() {
...
+          idFactoryMock).then(({ data, }) => {
+            expect(data.title).toEqual('Author Name');
+
+            expect(data.inReplyTo).toBeUndefined();
+            expect(data.replyType).toBeUndefined();
+            done();
+          }, done.fail);
+      });
+
+  it('should correctly parse IRT and set default RT.',
```
timvandermeij: The indentation is a bit off here and in other parts of the added code in this file. This and the code below is part of the `describe` block and should therefore be indented with two extra spaces.

vlastimilmaca:
Round #2 of fixes. Thank you very much for reviews, it is very educational for me :) … I'm sorry you have to go through it though :(.

vlastimilmaca:
Just rebase...

timvandermeij:
I noticed that you also fixed the indentation in the unit tests; great! I'm hoping to play around with this and take another look very soon, but the changes that have been made so far look really good.

test/unit/annotation_spec.js
```diff
@@ -233,6 +233,226 @@ describe('annotation', function() {
     });
   });
 
+  describe('MarkupAnnotation', function () {
```
timvandermeij: In my previous PR I already added a `describe` block for this (see https://github.com/mozilla/pdf.js/blob/master/test/unit/annotation_spec.js#L320), so you can simply place your new tests in there so we don't have to create the block here again.

vlastimilmaca: Didn't see that :). Thank you. Moved.

vlastimilmaca:
Should be fine, don't know why it won't run lint test :/

timvandermeij:
Hm, me neither, but it isn't caused by your patch. Most likely a new intermittent failure that we haven't seen before. Let's try again...

/botio-windows fonttest

vlastimilmaca:
well I did a force-push and then after few minutes, probably before it get to finish lint test, I did another force push … that may broke it? Should I try another force push?  But I don't know what to fix … it's perfect now :D.

vlastimilmaca:
So …  what about this ? :)

timvandermeij:
I have finally worked my way through the backlog of pull requests, releases and other tasks, so I'll soon have time for this pull request again. Sorry for the delay!

vlastimilmaca:
Nice :), no problem. We should probably start by waking up Travis somehow … can you run some force command ?

timvandermeij:
I triggered Travis by closing and reopening the pull request.

src/core/annotation.js
```diff
@@ -1005,16 +1058,23 @@ class PopupAnnotation extends Annotation {
 
     let dict = parameters.dict;
     let parentItem = dict.get('Parent');
+
```
timvandermeij: Please revert this change to keep the diff small.

test/unit/annotation_spec.js
```diff
@@ -233,6 +233,74 @@ describe('annotation', function() {
     });
   });
 
+  describe('TextAnnotation', function() {
+    it('should not parse state model when not defined',
```
timvandermeij: Change this to `should not parse state model and state when not defined`.

test/unit/annotation_spec.js
```diff
@@ -233,6 +233,74 @@ describe('annotation', function() {
     });
   });
 
+  describe('TextAnnotation', function() {
```
timvandermeij: Could you place this `describe` block below the `AnnotationBorderStyle` `describe` block for a more logical ordering?

test/unit/annotation_spec.js
```diff
@@ -233,6 +233,74 @@ describe('annotation', function() {
...
+      annotationDict.set('Contents', 'TestText');
+      const annotationRef = new Ref(819, 0);
+
+      const replyDict = new Dict();
+      replyDict.set('Type', Name.get('Annot'));
+      replyDict.set('Subtype', Name.get('Text'));
+      replyDict.set('IRT', annotationDict);
+      replyDict.set('RT', Name.get('R'));
+      replyDict.set('Contents', 'ReplyText');
+
```
timvandermeij: Remove the newline here, just like above, for consistency (applies to other blocks in tests below too since the newlines are a bit inconsistent throughout these changes).

test/unit/annotation_spec.js
```diff
@@ -233,6 +233,74 @@ describe('annotation', function() {
...
+      AnnotationFactory.create(xref, replyRef, pdfManagerMock,
+        idFactoryMock).then(({ data, }) => {
+          expect(data.state).toBeNull();
+          expect(data.stateModel).toBeNull();
+          done();
+        }, done.fail);
+    });
+
+    it('should correctly parse state model and state',
```
timvandermeij: Change this to `should parse state model and state when defined`.

test/unit/annotation_spec.js
```diff
@@ -233,6 +233,74 @@ describe('annotation', function() {
...
+
+      const replyRef = new Ref(820, 0);
+
+      const xref = new XRefMock([
+        { ref: annotationRef, data: annotationDict, },
+        { ref: replyRef, data: replyDict, }
+      ]);
+
+      AnnotationFactory.create(xref, replyRef, pdfManagerMock,
+        idFactoryMock).then(({ data, }) => {
```
timvandermeij: Formatting of these blocks should be like this in terms of indentation:
https://github.com/mozilla/pdf.js/blob/6e0e725af401cb3a7a96aba15624b1c625967cc6/test/unit/annotation_spec.js#L113-L117
(applies to other blocks in tests below too)

test/unit/annotation_spec.js
```diff
@@ -233,6 +233,74 @@ describe('annotation', function() {
...
+          expect(data.inReplyTo).toBeDefined();
```
timvandermeij: I would suggest to leave out this line and the one below it because this test only tests `stateModel` and `state` being set correctly. The other fields should be covered in other tests, which you already have below, to keep the test's scope limited.

test/unit/annotation_spec.js
```diff
@@ -343,6 +411,154 @@ describe('annotation', function() {
...
+      ]);
+
+      AnnotationFactory.create(xref, popupRef, pdfManagerMock,
+        idFactoryMock).then(({ data, }) => {
+
+          expect(data.contents).toEqual('Correct Text');
+
+          done();
+        }, done.fail);
+    });
```
timvandermeij: Could you move this last test to the `PopupAnnotation`'s `describe` block at:
https://github.com/mozilla/pdf.js/blob/6e0e725af401cb3a7a96aba15624b1c625967cc6/test/unit/annotation_spec.js#L1674

That's a better place for this test since it's relevant in that scope.

test/unit/annotation_spec.js
```diff
@@ -343,6 +411,154 @@ describe('annotation', function() {
 
       expect(markupAnnotation.creationDate).toEqual(null);
     });
+
+    it('should correctly parse T and not ' +
```
timvandermeij: Let's make this test only about not parsing IRT/RT, so change this to `should not parse IRT/RT when not defined` and remove the `T` test since that is not really related to IRT/RT not being parsed.

test/unit/annotation_spec.js
```diff
@@ -343,6 +411,154 @@ describe('annotation', function() {
...
+        idFactoryMock).then(({ data, }) => {
+          expect(data.title).toEqual('Author Name');
+
+          expect(data.inReplyTo).toBeUndefined();
+          expect(data.replyType).toBeUndefined();
+          done();
+        }, done.fail);
+    });
+
+    it('should correctly parse IRT and set default RT.', function (done) {
```
timvandermeij: Change this to `should parse IRT and set default RT when not defined`.

test/unit/annotation_spec.js
```diff
@@ -343,6 +411,154 @@ describe('annotation', function() {
...
+      AnnotationFactory.create(xref, replyRef, pdfManagerMock,
+        idFactoryMock).then(({ data, }) => {
+
+          expect(data.inReplyTo).toEqual(annotationRef.toString());
+          expect(data.replyType).toEqual('R');
+          done();
+        }, done.fail);
+    });
+
+    it('should correctly parse IRT, RT and take Contents ' +
```
timvandermeij: There are basically two situations when IRT/RT is parsed: either it's a group annotation (and its contents should be inherited) or it's a reply annotation. How about updating this test and the one below so that this is clear from the name and so all fields that can be changed, namely title, contents, creation date, modification date, `hasPopup` and color, are checked. That way it's clear which two scenarios we expect and we make sure that all fields that can be changed are in fact tested instead of only the `Contents` field. Both tests are already pretty close to this, but needs some tweaks to get there, mainly by changing the name and testing all fields.

This one can then be called `should parse IRT/RT for a group type` and the other one can then be called `should parse IRT/RT for a reply type`.

timvandermeij (commented):
You'll see that the remaining comments are mostly about the tests because I hadn't looked into those in detail before. The non-test code looks really good this way. I think that we can merge this pull request once these comments are taken care of. I was quite busy last time and finally found the time to properly review this, so sorry for the delay and thank you for your patience!

vlastimilmaca (commented):
I agree with everything suggested. Just had one question about repositioning `describe` block.

test/unit/annotation_spec.js
vlastimilmaca: You're right, it is weird this way … so I tried to follow this pattern (I found it in other tests in this file and I think it is more readable)

```
Dict1

Dict2

Ref1
Ref2
xRef

ResultCheck
```

In one occasion, where I needed Ref1 before Dict2 I went with this pattern
```
Ref1
Dict1

Ref2
Dict2

xRef

ResultCheck
```
If you'd like to see it formatted some other way, please let me know :).

vlastimilmaca:
Thank you very much for your time, I hope I fixed it all. Sorry my code is such a pain to get it right :).

vlastimilmaca:
Just rebased to master (Travis didn't like how it was before, rebase seems to fixed it).

test/unit/annotation_spec.js
timvandermeij: Yes, I think that's even better. Let's go for that.

test/unit/annotation_spec.js
timvandermeij: Looks really good this way!

test/unit/annotation_spec.js
```diff
@@ -343,6 +403,148 @@ describe('annotation', function() {
...
+      popupDict.set('Parent', annotationDict);
+      annotationDict.set('Popup', popupDict);
+
+      const replyDict = new Dict();
+      replyDict.set('Type', Name.get('Annot'));
+      replyDict.set('Subtype', Name.get('Text'));
+      replyDict.set('IRT', annotationDict);
+      replyDict.set('RT', Name.get('Group'));
+      replyDict.set('T', 'ReplyTitle');
+      replyDict.set('Contents', 'ReplyTitle');
```
timvandermeij: Shouldn't this be `ReplyText`, just like in the test below?

test/unit/annotation_spec.js
```diff
@@ -343,6 +403,148 @@ describe('annotation', function() {
...
+      const replyRef = new Ref(821, 0);
+      const xref = new XRefMock([
+        { ref: annotationRef, data: annotationDict, },
+        { ref: popupRef, data: popupDict, },
+        { ref: replyRef, data: replyDict, }
+      ]);
+
+      AnnotationFactory.create(xref, replyRef, pdfManagerMock,
+          idFactoryMock).then(({ data, }) => {
+        expect(data.inReplyTo).toBeDefined();
```
timvandermeij: I think you can even do `expect(data.inReplyTo).toEqual(annotationRef.toString());`.

test/unit/annotation_spec.js
```diff
@@ -343,6 +403,148 @@ describe('annotation', function() {
...
```
timvandermeij: Same here regarding testing equality like in the test above.

test/unit/annotation_spec.js
```diff
@@ -1540,6 +1742,43 @@ describe('annotation', function() {
         done();
       }, done.fail);
     });
+
+    it('should correctly inherit Contents from group-master annotation ' +
+       'if parent has ReplyType == Group', function (done) {
+      const annotationDict = new Dict();
+      annotationDict.set('Type', Name.get('Annot'));
+      annotationDict.set('Subtype', Name.get('Text'));
+      annotationDict.set('Contents', 'Correct Text');
```
timvandermeij: Just like in the tests above, we should check all fields instead of only the contents, i.e., contents, title, modification date and color (no need for the flags).

timvandermeij (approved):
I'm approving this for merging for when these final comments (and the one comment you indicated about moving the `TextAnnotation` unit tests below the `MarkupAnnotation` ones) have been addressed. Nice work!

test/unit/annotation_spec.js
vlastimilmaca: You are right, but I have to change few things to get it running. 

Test for `.toBeDefined` is nonsense in how it is now, because I set `annotationDict` into `IRT` property. In actual `annotations.js` code, there is test for `isRef`, and that ends up being null (Dict is not a Ref and null passes `toBeDefined `test … hence the "nonsense").

So instead of setting annotationDict I have to set annotationRef and add this section to make it pass:
```
 annotationDict.assignXref(xref);
 popupDict.assignXref(xref);
 replyDict.assignXref(xref);
```
Setting of `Dict `into properties where should be `Ref `seems to be used in other tests too and I think it is probably not correct, because it is not what is in real PDFs? I'm not sure … I don't understand this part of PDF world enough :).

vlastimilmaca:
Should be all fixed. Thank you :).

timvandermeij (approved):

timvandermeij:
Thank you for your contribution! I think the end result is a patch that is very well tested and maintainable.

vlastimilmaca:
Nice. Thank you :).

PR #7775, opened 2016-11-01
Widget annotation: implement field name according to the specification

timvandermeij:
The original code is difficult to read and, more importantly, performs actions that are not described in the specification. It replaces empty names with a backtick and an index, but this behavior is not described in the specification. While the specification is not entirely clear about what should happen in this case, it does specify that the `T` field is optional and that multiple field dictionaries may have the same fully qualified name, so to achieve this it makes the most sense to ignore missing `T` fields during construction of the field name. This is the most specification-compliant solution and, judging by opened issue #6623, also the required and expected behavior.

Fixes #6623.
Part of #7613.

PR #7570, opened 2016-08-26
Create a fallback annotation `id` for entries in `Annots` dictionaries that are not indirect objects (issue 7569)

Snuffleupagus:
According to the PDF specification, see http://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#page=86, entries in `Annots` dictionaries should be indirect objects, but obviously there're PDF generators that ignore this.

Fixes #7569.

/cc @timvandermeij If you agree with the general idea of this patch, I'll add a unit-test as well.

PR #10826, opened 2019-05-16
Remove unused `ref` property from the `parameters` object used when creating annotations in `AnnotationFactory._create`

Snuffleupagus:
The use-cases for this property was removed in PRs #7570 and #7775, and it's been completely unused ever since the latter one.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #10833, opened 2019-05-18
[api-minor] Implement contents and creation date for the correct annotation types

timvandermeij:

Thanks to @vlastimilmaca for finding this! This patch cleans this up so it doesn't have to be done in the scope of the IRT/RT patch.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -176,7 +176,6 @@ class Annotation {
     const dict = params.dict;
 
     this.setContents(dict.get('Contents'));
-    this.setCreationDate(dict.get('CreationDate'));
     this.setModificationDate(dict.get('M'));
     this.setFlags(dict.get('F'));
     this.setRectangle(dict.getArray('Rect'));
@@ -190,7 +189,6 @@ class Annotation {
       borderStyle: this.borderStyle,
       color: this.color,
       contents: this.contents,
-      creationDate: this.creationDate,
       hasAppearance: !!this.appearance,
       id: params.id,
       modificationDate: this.modificationDate,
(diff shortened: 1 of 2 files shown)


timvandermeij:
@Snuffleupagus Do you perhaps have time to review this?

Snuffleupagus (approved):
The code changes look OK to me (however I've not checked the specification myself, so basically rubber-stamping that part :-)

src/core/annotation.js
```diff
@@ -243,15 +243,16 @@ class Annotation {
...
    *
    * @public
    * @memberof Annotation
-   * @param {string} creationDate - PDF date string that indicates when the
-   *                                annotation was originally created
+   * @param {string} contents - Text to display for the annotation or, if the
+   *                            type of annotation does not display text, a
+   *                            description of the annotation's contents
    */
-  setCreationDate(creationDate) {
```
Snuffleupagus: I suppose that you could've moved this method into the `MarkupAnnotation` class instead, rather than duplicating the code below, but it probably doesn't matter too much either way so...

vlastimilmaca: +1 to that, I will probably need that piece of code too … so more repetitions coming :)

timvandermeij: I agree that that's actually better. Thanks, I'm going to fix that.

Snuffleupagus: Actually, looking at this once more you probably want to just leave it on the base `Annotation` class, since you're calling it from both the `MarkupAnnotation` and `PopupAnnotation` class!?

vlastimilmaca: Nice catch :) … since only modification date is now shown in UI popup (I think), I can probably drop it from PopupAnnotation, because it should not be there anyway (as you correctly pointed out - popup is not markup). I added it there in IRT PR, because it was added in base Annotation class and it needed to be changed to correct value in Parent->RT=Group situation. So I think, it is ok to move it to MarkupAnnotation and remove it from PopupAnnotation, because it should not be needed there.

timvandermeij: Done in the new commit. Thank you both!

Issue #9248, opened 2017-12-07
[pdfjs-dist] 'worker-loader' requires 'webpack' as peerDependency


MichaelDeBoey:
When I install `pdfjs-dist` in a clean repo, I get the following warning:

```sh
npm WARN worker-loader@1.1.0 requires a peer of webpack@^2.0.0 || ^3.0.0 but none is installed. You must install peer dependencies yourself.
```

Please add `webpack` (at least) to the `peerDependencies` (`dependencies` can be as well, but will make this one bloated), since this is a [peerDependency](https://github.com/webpack-contrib/worker-loader/blob/master/package.json#L34) of `worker-loader`.

PR #10844, opened 2019-05-22
Remove webpack from peerDependencies

### LINKED ISSUE #9580: Consider moving Webpack support to a separate package
I propose moving webpack.js to a separate package.
Why?

If someone is instaling pdf.js, while NOT using Webpack, they will see the following warnings:

```
warning "yourpackage > pdfjs-dist@2.0.305" has unmet peer dependency "webpack@^2.0.0 || ^3.0.0".
warning "yourpackage > pdfjs-dist > worker-loader@1.1.1" has unmet peer dependency "webpack@^2.0.0 || ^3.0.0 || ^4.0.0".
warning "yourpackage > pdfjs-dist > worker-loader > schema-utils@0.4.5" has unmet peer dependency "webpack@^2.0.0 || ^3.0.0 || ^4.0.0".
```

By doing so, we could get rid of `worker-loader` and `webpack` dependencies, making required download/build time/test time significantly smaller.

### LINKED ISSUE #9733: Drop webpack peerDependency from pdfjs-dist project
The pdfjs-dist project should not include webpack as a peerDependency for it can really work without it when loading the scripts from build directory.

What went wrong?
I am using PDFJS in a component to visualize a PDF using Angular 6. No including webpack in my project (I do not need it at all), make a peerDependency warning popup every time I run any npm command or every time anyone install my component.

denkristoffer:
Closes #9733
Closes #9580

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1326,9 +1326,6 @@ gulp.task('dist-pre', gulp.series('generic', 'components', 'image_decoders',
       'node-ensure': '^0.0.0', // shim for node for require.ensure
       'worker-loader': '^2.0.0', // used in external/dist/webpack.json
     },
-    peerDependencies: {
-      'webpack': '^3.0.0 || ^4.0.0-alpha.0 || ^4.0.0', // from 'worker-loader'
-    },
     browser: {
       'fs': false,
       'http': false,


timvandermeij:
Closing since merging this isn't really a good option given that #9248 will return if we do that. Instead, what we should do is #9580 which is discussed in IRC at https://mozilla.logbot.info/pdfjs/20180606#c14862530-c14862634.

PR #10850, opened 2019-05-23
Scale stroking line width when using a tiling pattern.

### LINKED ISSUE #10566: table borders
[ffpdfjs_bug_tablelines.pdf](https://github.com/mozilla/pdf.js/files/2875821/ffpdfjs_bug_tablelines.pdf)

Configuration:
- Web browser and its version: Firefox Quantum 65.0 (64 bits)
- Operating system and its version: macOS Mojave 10.14.3
- PDF.js version: 
- Is a browser extension:

1. open attached pdf in firefox (drag and drop in new tab)
2. the rendering of the borders in the table changes with the level of zoom

expected behavior (from preview/or google chrome)
capture d ecran 2019-02-18 a 14 47 55
 

screenshot_2019-02-18 ffpdfjs_bug_tablelines pdf

This pdf was modified for anonymisation reasons. 
But you can find this in the metadata of the original file : 
```
File Permissions                : rw-r--r--
File Type                       : PDF
File Type Extension             : pdf
MIME Type                       : application/pdf
PDF Version                     : 1.6
Linearized                      : No
Encryption                      : Standard V4.4 (128-bit)
User Access                     : (none)
Creator                         : Excel
Producer                        : Mac OS X 10.12.6 Quartz PDFContext
XMP Toolkit                     : Adobe XMP Core 5.4-c006 80.159825, 2016/09/16-03:31:08
Format                          : application/pdf
Creator Tool                    : Excel```

### PR DESCRIPTION
**Pull request by brendandahl**
Fixes #10566.

### CODE REVIEW

**Comment by brendandahl** [comment]
Don't merge yet...I see another potential issue that I can fix that I want to bundle with this.

**Comment by janpe2** [comment]
This fixes #10566, which is probably the issue you were looking for.

File: src/display/canvas.js [hunk 1]
```diff
@@ -1156,10 +1153,22 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
         // for patterns, we transform to pattern space, calculate
         // the pattern, call stroke, and restore to user space
         ctx.save();
+        // The current transform will be replaced while building the pattern,
+        // but the line width needs to be adjusted by the current transform, so
+        // we must scale it. Technically, the line width could be scaled
+        // differently in the X and Y direction, but we can't control that,
+        // so just scale in one direction.
+        let transform = ctx.mozCurrentTransform;
```
timvandermeij: This variable can be `const` instead of `let`.

src/display/canvas.js
```diff
@@ -1156,10 +1153,22 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
...
+        var scale = Util.singularValueDecompose2dScale(transform)[0];
```
timvandermeij: This variable can be `const` instead of `var`.

timvandermeij:
I think this is good to merge once the two comments above are resolved.

brendandahl:
I won't have time to do the full fix yet, so for now we can live with this workaround. I've filed #10955 for the full fix.

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

> I won't have time to do the full fix yet, so for now we can live with this workaround. I've filed #10955 for the full fix.

Fixing this incrementally is perfectly fine because at least now it works and we have reference test coverage, making a future improvement easier to verify.

(Making new reference images in the other PR.)

PR #8168, opened 2017-03-16
Normalize the `BBox` entry in Tiling Pattern dictionaries (issue 8117)

Snuffleupagus:
According to the PDF specification, see http://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G7.3982967, the `BBox` entry should have the form `[left, bottom, right, top]`.
Since some PDF generators apparently violates the specification, we normalize the `BBox` to ensure that the pattern is (correctly) rendered.

Fixes #8117.

PR #10851, opened 2019-05-24
Apply bounding box before using shading patterns.

### LINKED ISSUE #8092: shading pattern bbox ignored
https://issues.apache.org/jira/secure/attachment/12658953/axsh02.pdf

Configuration:
- Web browser and its version: FF 51.0.1  32bit
- Operating system and its version: W7
- PDF.js version: latest

1. display file

axsh02-screenshot-adobe

axsh02-screenshot-pdfjs

The shading pattern fills the whole page, but it should leave some white margins on all sides. I suspect that the BBox is ignored.

axsh02-screenshot-pdfdebugger

brendandahl:
Fixes #8092

### DIFF
--- a/src/core/pattern.js
+++ b/src/core/pattern.js
@@ -96,6 +96,12 @@ Shadings.RadialAxial = (function RadialAxialClosure() {
     var cs = dict.get('ColorSpace', 'CS');
     cs = ColorSpace.parse(cs, xref, res, pdfFunctionFactory);
     this.cs = cs;
+    const bbox = dict.getArray('BBox');
+    if (Array.isArray(bbox) && bbox.length === 4) {
+      this.bbox = Util.normalizeRect(bbox);
+    } else {
+      this.bbox = null;
+    }
 
     var t0 = 0.0, t1 = 1.0;
     if (dict.has('Domain')) {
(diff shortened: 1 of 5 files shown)


src/core/pattern.js
```diff
@@ -96,6 +96,7 @@ Shadings.RadialAxial = (function RadialAxialClosure() {
     var cs = dict.get('ColorSpace', 'CS');
     cs = ColorSpace.parse(cs, xref, res, pdfFunctionFactory);
     this.cs = cs;
+    this.bbox = dict.getArray('BBox');
```
Snuffleupagus: Would it make sense to normalize the bounding box here (as is done elsewhere in this file), i.e. `Util.normalizeRect(dict.getArray('BBox'));`, to prevent future issues?

timvandermeij: I agree since we have seen breakage because of that before (see #8168 and #10162).

src/display/pattern_helper.js
```diff
@@ -17,17 +17,30 @@ import { FormatError, info, Util } from '../shared/util';
 
 var ShadingIRs = {};
 
+function applyBoundingBox(ctx, bbox) {
+  if (!bbox || typeof Path2D === 'undefined') {
+    return;
+  }
+  let width = bbox[2] - bbox[0];
```
timvandermeij: This variable (and the two below) can be `const` instead of `let`.

timvandermeij:
I think this is good to merge once the two comments above are resolved.

brendandahl:
Sorry for the delay on this and the other patch. I ended up finding quite a few problems with how we're doing patterns with respect to transforms and the patch is turning into much more work. I'll hopefully have some more time this week to give it another stab.

timvandermeij (approved):

timvandermeij:
Looks good; thank you!

/botio makeref

PR #10854, opened 2019-05-24
make print resolution configurable via AppOptions

kleins05:
There are many posts online regarding pdf.js print quality. From a [post](https://www.gyrocode.com/articles/how-to-increase-print-quality-of-pdf-file-with-pdf-js-viewer/) by Michael Ryvkin I learned that this can easily be fixed by increasing print resolution. To make this even easier without the need to change the pre-built viewer.js, I propose making print resolution configurable via AppOptions.

### DIFF
--- a/web/app_options.js
+++ b/web/app_options.js
@@ -233,6 +233,11 @@ if (typeof PDFJSDev === 'undefined' ||
     value: (typeof navigator !== 'undefined' ? navigator.language : 'en-US'),
     kind: OptionKind.VIEWER,
   };
+  defaultOptions.printResolution = {
+    /** @type {number} */
+    value: 150,
+    kind: OptionKind.VIEWER,
+  };
 }
 
 const userOptions = Object.create(null);
--- a/web/pdf_print_service.js
+++ b/web/pdf_print_service.js
@@ -15,6 +15,7 @@
 
 import { CSS_UNITS, NullL10n } from './ui_utils';
 import { PDFPrintServiceFactory, PDFViewerApplication } from './app';
+import { AppOptions } from './app_options';
 import { URL } from 'pdfjs-lib';
 
 let activeService = null;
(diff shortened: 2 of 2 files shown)


web/app_options.js
```diff
@@ -105,6 +105,11 @@ const defaultOptions = {
     value: false,
     kind: OptionKind.VIEWER + OptionKind.PREFERENCE,
   },
+  printResolution: {
```
Snuffleupagus: Please move this below to https://github.com/mozilla/pdf.js/blob/1421b2f2056ae6dac564469374ff43342a3fb891/web/app_options.js#L224-L236 instead, since the `printResolution` doesn't make sense for e.g. `MOZCENTRAL` builds.

web/pdf_print_service.js
```diff
@@ -26,10 +27,9 @@ function renderPage(activeServiceOnEntry, pdfDocument, pageNumber, size) {
   let scratchCanvas = activeService.scratchCanvas;
 
   // The size of the canvas in pixels for printing.
-  const PRINT_RESOLUTION = 150;
-  const PRINT_UNITS = PRINT_RESOLUTION / 72.0;
-  scratchCanvas.width = Math.floor(size.width * PRINT_UNITS);
-  scratchCanvas.height = Math.floor(size.height * PRINT_UNITS);
+  let printUnits = AppOptions.get('printResolution') / 72.0;
```
Snuffleupagus: Please don't change `const` to `let` unless actually necessary, and don't unnecessarily rename variables (it only increases the size of the diff).

Just change this back, and format it like this instead:
```javascript
const PRINT_RESOLUTION = AppOptions.get('printResolution') || 150;
const PRINT_UNITS = PRINT_RESOLUTION / 72.0;
```

Snuffleupagus (changes requested):

web/app_options.js
kleins05: Did that. Thanks.

web/pdf_print_service.js
kleins05: Fixed that, too. I had misconceptions about the use of `const` in Javascript, so I learned something, too.

timvandermeij (approved):

timvandermeij:
Really good. Thank you for contibuting this!

suguoyao:
I have same problem and it has been solved. Thanks!!!!! 👍👍

beijingalex:
> 我有同样的问题，它已经解决了。谢谢！！！！！👍👍

@suguoyao 你是怎么解决的？打印分辨率低的问题。

suguoyao:
@beijingalex  
Change the value of variable `PRINT_RESOLUTION` to 300. More about code please view this [post](https://www.gyrocode.com/articles/how-to-increase-print-quality-of-pdf-file-with-pdf-js-viewer/).

PR #10862, opened 2019-05-26
Add caching to reduce the number of `Ref` objects

Snuffleupagus:
This is similar to the existing caching used to reduced the number of `Cmd` and `Name` objects.
With the `tracemonkey.pdf` file, this patch changes the number of `Ref` objects as follows (in the default viewer):

|          | Loading the first page | Loading *all* the pages |
|----------|------------------------|-------------------------|
| `master` | 332                    | 3265                    |
| `patch`  | 163                    | 996                     |

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -627,7 +627,7 @@ class PDFDocument {
     const { catalog, linearization, } = this;
     assert(linearization && linearization.pageFirst === pageIndex);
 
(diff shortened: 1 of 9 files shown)


timvandermeij (approved):

timvandermeij:
I really like this idea. Thank you for optimizing this!

PR #10863, opened 2019-05-26
Ensure that the `Cmd`/`Name`/`Ref` caches are cleared when running other `cleanup` code

Snuffleupagus:
The purpose of these caches is to reduce peak memory usage, by only ever having *a single* instance of a particular object.
However, as-is these caches are never cleared and they will thus remain until the worker is destroyed. This could very well have a negative effect on total memory usage, particularly for large/long documents, hence it seems to make sense to clear out these caches together with various other ones.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -20,8 +20,8 @@ import {
   warn
 } from '../shared/util';
 import {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

Issue #10639, opened 2019-03-12
"Invalid or corrupted PDF files" is displayed


simonhong:
Attach (recommended) or Link to PDF file here:
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.144.7135&rep=rep1&type=pdf

Configuration:
- Web browser and its version: Chrome 72.0.3621.121 (Official Build) (64-bit)
- Operating system and its version: MacOS
- PDF.js version: PDF.js v2.0.673 (build: 31012570)
- Is a browser extension: Yes

Steps to reproduce the problem:
1. Click the above link
2.

What is the expected behavior? (add screenshot)
This is what I can see when I pasted chrome-extension prefixed url or reloading the error pdf page.
(`chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.144.7135&rep=rep1&type=pdf`)
![Screen Shot 2019-03-12 at 17 20 26](https://user-images.githubusercontent.com/6786187/54184930-26c5a080-44eb-11e9-9780-3b76b52c1eef.png)

What went wrong? (add screenshot)
This is what I can see when click the above link
![Screen Shot 2019-03-12 at 17 19 38](https://user-images.githubusercontent.com/6786187/54184903-17465780-44eb-11e9-8e72-fcb34372a610.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):
https://chrome.google.com/webstore/detail/pdf-viewer/oemmndcbldboiebfnladdacbdfmadadm

Snuffleupagus:
Possibly a duplicate of #10562.

simonhong:
@Snuffleupagus I think this is a different issue with #10562.
This is for some pdf isn't opened properly, whereas #10562 is the issue that pdf is opened by chrome's internal pdf viewer(pdfium?) instead of pdfjs extension.

simonhong:
I found the cause of this issue.
The reason is requesting twice to server in a very short interval.
To the second request, server redirects to `downloadsexceeded.html` instead of pdf content.
So, pdf.js complains it's invalid/corrupted pdf file.
It's maybe server's DDoS protection I think.

Why two requests are issued when user clicks that link?
First one is issued by browser for user click.
Then, pdfjs extension intercepts header response and redirects to extension url.
Then, one more requesting is issued by `pdf.js`.

[...]

PR #10869, opened 2019-05-29
Restore referrer preserving functionality in Chrome 72+

### LINKED ISSUE #10645: Referrer is not preserved in Chrome 72+
The functionality from #5823 (https://github.com/mozilla/pdf.js/blob/80135378cadd98b55a835446f0857e4bc30524e0/extensions/chromium/preserve-referer.js) is broken in Chrome 72 because the Referer request header is not included by default.

To add it, the "extraHeaders" option need to be added at [line 51](https://github.com/mozilla/pdf.js/blob/80135378cadd98b55a835446f0857e4bc30524e0/extensions/chromium/preserve-referer.js#L51) and [line 108](https://github.com/mozilla/pdf.js/blob/80135378cadd98b55a835446f0857e4bc30524e0/extensions/chromium/preserve-referer.js#L108).

Before publishing an update with the fix, we need to consider the impact of https://bugs.chromium.org/p/chromium/issues/detail?id=938560 , and whether we really need to continue the Referer preserving that was introduced to resolve #5676 .

Rob--W:
Tested by following the test case at https://github.com/mozilla/pdf.js/issues/10639#issuecomment-494474824

In Chrome 71 and earlier: No difference (PDF loads successfully).
In Chome 72 and later: Before this patch the PDF could not be loaded; after the patch the PDF loads successfully.

Fixes #10645

### DIFF
--- a/extensions/chromium/preserve-referer.js
+++ b/extensions/chromium/preserve-referer.js
@@ -41,14 +41,26 @@ var g_requestHeaders = {};
 // g_referrers[tabId][frameId] = referrer of PDF frame.
 var g_referrers = {};
 
+var extraInfoSpecWithHeaders; // = ['requestHeaders', 'extraHeaders']
+
 (function() {
   var requestFilter = {
     urls: ['*://*/*'],
     types: ['main_frame', 'sub_frame'],
   };
-  chrome.webRequest.onSendHeaders.addListener(function(details) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #10872, opened 2019-05-30
Re-use, rather than re-creating, some `Array`s when resetting them in `src/display/api.js`

Snuffleupagus:
Calling `someArray = []` will create a new Array, which seems completely unnecessary when it's sufficient to just call `someArray.length = 0` to achieve the same effect.

Even though I cannot imagine these particular cases having any noticeable performance impact, similar changes were made in `core/` code years ago since it's apparently more efficient memory wise.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1377,7 +1377,7 @@ class LoopbackPort {
   }
 
   terminate() {
-    this._listeners = [];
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #10877, opened 2019-05-31
Ignore Annotations with too large border `width`s, to prevent the `annotationLayer` from rendering it over the surrounding document (bug 1552113)

Snuffleupagus:
The border `width` will instead fallback to the default value of `1`, rather than ignoring it altoghether, to also ensure that e.g. `LinkAnnotation`s become clickable as intended.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1552113

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -16,7 +16,8 @@
 
 import {
   AnnotationBorderStyleType, AnnotationFieldFlag, AnnotationFlag,
-  AnnotationType, isString, OPS, stringToBytes, stringToPDFString, Util, warn
+  AnnotationType, assert, isString, OPS, stringToBytes, stringToPDFString, Util,
+  warn
 } from '../shared/util';
 import { Catalog, FileSpec, ObjectLoader } from './obj';
 import { Dict, isDict, isName, isRef, isStream } from './primitives';
@@ -361,6 +362,11 @@ class Annotation {
    * @param {Dict} borderStyle - The border style dictionary
    */
   setBorderStyle(borderStyle) {
+    if (typeof PDFJSDev === 'undefined' ||
+        PDFJSDev.test('!PRODUCTION || TESTING')) {
+      assert(this.rectangle, 'setRectangle must have been called previously.');
+    }
+
     this.borderStyle = new AnnotationBorderStyle();
     if (!isDict(borderStyle)) {
       return;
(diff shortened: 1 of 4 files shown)


src/core/annotation.js
```diff
@@ -497,16 +503,39 @@ class AnnotationBorderStyle {
...
+        // Truncate ridiculously large `width`s to the `Rect` entry of the
+        // annotation, to prevent the `annotationLayer` from rendering it
+        // over the surrounding document (fixes bug1552113.pdf).
+        //
+        // This is consistent with the behaviour in Adobe Reader.
+        if ((maxWidth > 0 && maxHeight > 0) &&
+            (width > maxWidth || width > maxHeight)) {
+          warn('AnnotationBorderStyle.setWidth - truncating large width: ' +
+               width);
+          width = Math.min(maxWidth, maxHeight);
```
Snuffleupagus: A somewhat related problem, which this patch probably ought to fix as well, is that these kind of Annotations aren't currently clickable.
It appears that only the "contents", and not the border itself, is part of the clickable area and in cases such as these that may becomes zero because of https://github.com/mozilla/pdf.js/blob/209e42043af86f54bae02c37d320350bf4f96a0c/src/display/annotation_layer.js#L175-L176

The simplest solution may actually be to outright just ignore the border `width`, i.e. setting it to zero, for the cases where it's clearly wrong; suggestions on all of this is much appreciated here!

Snuffleupagus: Having thought about this some more, I'm starting to favour the "set `width = 1` (i.e. the default value)"-approach in these cases.
Besides obviously fixing the issues with non-clickable links, it still displays at least *some* visible border which seems appropriate since the PDF file did ask for a border to be displayed (albeit a ridiculously large one).

timvandermeij: I noticed that Okular displays a similar large blue box and it looks like the Chrome web PDF viewer doesn't display the border at all, so the behavior differs. In the specification I could only find the sentence `If present, the border shall be drawn completely inside the annotation  rectangle` and I think that's also what you want to achieve in this version of the patch. However, in my opinion the border is still way too large and causes not only the clickable region to be invisible, but also the underlying superscript text as well. Therefore, I agree that if the width is too large, it's best to set it to the default width of 1 because, as you indicated as well, there should be some border. According to the spec, only if the width is 0 no border should be drawn, so I think that's another reason to not set it to 0 in this case, but to 1.

Snuffleupagus: > How does Adobe Reader render this?

The same way as this patch does, i.e. a blue rectangle limited in size to the Annotation `Rect` (thus obscuring *only* the superscript text). However, it's at least clickable in Adobe Reader.

I'll update the patch to use `width = 1` instead in this case, and this sort of problem ought to very rare in practice; thank you!

timvandermeij (approved):

timvandermeij:
Thank you for fixing this special case!

PR #10879, opened 2019-06-02
[api-minor] Attempt to reduce memory usage during printing, by always running `cleanup` once rendering has finished

Snuffleupagus:
Given that `cleanupAfterRender` is already set for large images, when handling 'obj' messages, this patch *should* thus be safe in general (since otherwise there ought be existing bugs related to cleanup and printing).


timvandermeij (approved):

timvandermeij:
Seems to work just fine; thanks!

PR #10888, opened 2019-06-05
My chages

ghost:


Snuffleupagus:
Duplicate of #9645 (and any sort of resizing really needs to happen on the worker-thread for performance reasons).

CIE81:
I opened this accidentaly, but I am putting together a PR for this. @Snuffleupagus resize will be harder to be done, in this case on the worker-thread, since we need access to the canvas to see if we need to resize and at what factor.

Snuffleupagus:
> resize will be harder to be done, in this case on the worker-thread, since we need access to the canvas to see if we need to resize and at what factor.

See e.g. https://github.com/mozilla/pdf.js/pull/9645#discussion_r182134534 and https://github.com/mozilla/pdf.js/pull/9645#issuecomment-455871097, and any sort of expensive computation *must* happen in the worker.

PR #10890, opened 2019-06-07
Add support for outline items, in the default viewer, which default to collapsed when the outline is built

Snuffleupagus:
The PDF specification supports this feature, which is commonly used in large/long documents (such as the spec itself), and it seems reasonably straightforward to implement; see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G11.2095911


timvandermeij (approved):

timvandermeij:
Thanks!

PR #10897, opened 2019-06-10
Remove a superfluous `linkService.isPageVisible` check from `PDFFindController` (PR 10217 follow-up)

Snuffleupagus:
Unless the `PDFLinkService` instance contains all of the expected methods, a lot of things will break in various places in the default viewer. Hence there's not much value in having this check, and outright falling seems more appropriate.

Finally, this also makes the return value explicit in this case, since that's consistent with the rest of the `PDFFindController._shouldDirtyMatch` method.


timvandermeij (approved):

timvandermeij:
Looks good!

Issue #4985, opened 2014-06-20
Tracemonkey is not rendering perfectly


timvandermeij:
- [x] Image on page 2 is not rendering properly.
- [x] Image on page 11 misses color and gradients.
- [ ] Image on page 13 misses some lines and fills.
- [x] The console also complains about some other missing operations.

terurou:
Fixed page 11 issue by #10646.

timvandermeij:
The rendering of page 13 is improved by the pull request above, but it's not entirely fixed yet.

Issue #10487, opened 2019-01-24
PDF renders incorrectly using SVG 


akankshaBiz:
Attach (recommended) or Link to PDF file here:
[Artwork-management-Process-Report.pdf](https://github.com/mozilla/pdf.js/files/2791207/Artwork-management-Process-Report.pdf)

Configuration:

Web browser and its version: Chrome
Operating system and its version: OSX High Sierra
PDF.js version: latest
Is a browser extension: no

I am trying to render the pdf as SVG but it is not correctly rendered.

actual file:
<img width="1186" alt="screenshot 2019-01-24 at 12 48 33 pm" src="https://user-images.githubusercontent.com/40818745/51661414-d647c180-1fd6-11e9-970c-de8a7a16d1b5.png">

rendered file:
<img width="967" alt="screenshot 2019-01-24 at 12 48 49 pm" src="https://user-images.githubusercontent.com/40818745/51661436-e5c70a80-1fd6-11e9-8253-8feec28e4cc6.png">

Also, I want to use hand tool cursor so I tried using the secondary tool as the hand tool is there but it seems we can't use specific tools. Since I want hand tool only I am passing that button in the options but in the code, we need to pass buttons for all the tools, why is that?

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

akankshaBiz:
Below is the screenshot of the error I am getting when this PDF renders.

<img width="352" alt="screenshot 2019-01-24 at 1 07 55 pm" src="https://user-images.githubusercontent.com/40818745/51662239-2e7fc300-1fd9-11e9-9a95-cefdae195575.png">

Snuffleupagus:
You need to provide *all* of the information requested in [ISSUE_TEMPLATE.md](https://github.com/mozilla/pdf.js/blob/master/.github/ISSUE_TEMPLATE.md) when opening an issue. If you ignore the `Attach (recommended) or Link to PDF file here:` point in particular, the issue is incomplete and will be closed as such (since it's impossible for anyone to help without access to the PDF file).

akankshaBiz:
@Snuffleupagus, I have attached the PDF in the issue. Please check this issue.

timvandermeij:
This will most likely be fixed by the patterns/gradients patch in the queue.

terurou:
Fixed partially by #10646.
Must implement these operators:

> Warning: Unimplemented operator beginGroup
[...]

PR #10902, opened 2019-06-14
Implement tiling patterns for the SVG back-end

### LINKED ISSUE #10796: When converting to SVG the "fill-opacity: 1" attribute should not be added if the "fill" attribute is missing
 [http://www.mediafire.com/file/jz26jy0zz326pw3/hypothermia-prevention-and-management-in-adults-having-surgery-update+Dec2016.pdf](http://www.mediafire.com/file/jz26jy0zz326pw3/hypothermia-prevention-and-management-in-adults-having-surgery-update+Dec2016.pdf)

Configuration:
- Web browser and its version: Google Chrome 73.0.3683.103 (Official Build) (64-bit)
- Operating system and its version: Ubuntu 18.04
- PDF.js version: "pdfjs-dist": "~2.0.943"
- Is a browser extension: No

1. Upload and convert the pdf file to SVG.

Create the svg file without the "fill-opacity: 1" attribute since the "fill" attribute is missing.

The pages from 2 to the last have the "fill-opacity: 1" attribute, but do not have the "fill" attribute, so the browser paints all these pages into black color by default: [https://www.dropbox.com/s/fgrcdr5573cwf81/Screen%20Shot%202019-05-02%20at%2015.37.21.png?dl=0](https://www.dropbox.com/s/fgrcdr5573cwf81/Screen%20Shot%202019-05-02%20at%2015.37.21.png?dl=0)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): None

ahuglajbclajep:
It may not be perfect, but some PDFs, such as #10796, can be rendered.

Fixes #10796.
Improves #4985.

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -982,10 +982,63 @@ SVGGraphics = class SVGGraphics {
    */
   _makeColorN_Pattern(args) {
     if (args[0] === 'TilingPattern') {
-      warn('Unimplemented pattern TilingPattern');
-      return null;
+      return this._makeTilingPattern(args);
     }
     return this._makeShadingPattern(args);
+  }
+
+  /**
+   * @private
+   */
+  _makeTilingPattern(args) {
+    const color = args[1];
+    const operatorList = args[2];
+    const matrix = args[3] || IDENTITY_MATRIX;
+    const [x0, y0, x1, y1] = args[4];
+    const xstep = args[5];
+    const ystep = args[6];
+    const paintType = args[7];
+
+    const tilingId = `shading${shadingCount++}`;
+    const [tx0, ty0] = Util.applyTransform([x0, y0], matrix);
+    const [tx1, ty1] = Util.applyTransform([x1, y1], matrix);
+    const [xscale, yscale] = Util.singularValueDecompose2dScale(matrix);
+    const txstep = xstep * xscale;
+    const tystep = ystep * yscale;
+
+    const tiling = this.svgFactory.createElement('svg:pattern');
+    tiling.setAttributeNS(null, 'id', tilingId);
+    tiling.setAttributeNS(null, 'patternUnits', 'userSpaceOnUse');
+    tiling.setAttributeNS(null, 'width', txstep);
+    tiling.setAttributeNS(null, 'height', tystep);
+    tiling.setAttributeNS(null, 'x', `${tx0}`);
+    tiling.setAttributeNS(null, 'y', `${ty0}`);
+
+    // Save current state.
+    const svg = this.svg;
+    const transformMatrix = this.transformMatrix;
+    const fillColor = this.current.fillColor;
+    const strokeColor = this.current.strokeColor;
+
+    const bbox = this.svgFactory.create(tx1 - tx0, ty1 - ty0);
+    this.svg = bbox;
+    this.transformMatrix = matrix;
+    if (paintType === 2) {
+      const cssColor = Util.makeCssRgb(...color);
+      this.current.fillColor = cssColor;
+      this.current.strokeColor = cssColor;
+    }
+    this.executeOpTree(this.convertOpList(operatorList));
+
+    // Restore saved state.
+    this.svg = svg;
+    this.transformMatrix = transformMatrix;
+    this.current.fillColor = fillColor;
+    this.current.strokeColor = strokeColor;
+
+    tiling.appendChild(bbox.childNodes[0]);
+    this.defs.appendChild(tiling);
+    return `url(#${tilingId})`;
   }
 
   /**


src/display/svg.js
```diff
@@ -982,8 +982,54 @@ SVGGraphics = class SVGGraphics {
...
+      const strokeColor = this.current.strokeColor;
+
+      const bbox = this.svgFactory.create(tx1 - tx0, ty1 - ty0);
+      this.svg = bbox;
+      this.transformMatrix = matrix;
+      if (paintType === 2) {
+        const cssColor = Util.makeCssRgb(...color);
+        this.current.fillColor = cssColor;
+        this.current.strokeColor = cssColor;
+      }
```
timvandermeij: Let's change this to:

```javascript
} else {
  warn(`Unimplemented tiling type ${paintType}`);
}
```

This way we still get notified in the console if an unsupported tiling type is used, which helps for debugging issues.

src/display/svg.js
```diff
@@ -982,8 +982,54 @@ SVGGraphics = class SVGGraphics {
    */
   _makeColorN_Pattern(args) {
     if (args[0] === 'TilingPattern') {
-      warn('Unimplemented pattern TilingPattern');
-      return null;
+      const color = args[1];
```
timvandermeij: Perhaps it's better to split the logic off to a new method named `_makeTilingPattern`, just like `_makeShadingPattern` below, for consistency.

timvandermeij (commented):
I haven't tested this yet, but I really like that the implementation is short and easy to follow!

> It may not be perfect, [...]

That's perfectly okay, in fact, it's much easier to review/merge if the patch is small. It's always preferred to have multiple smaller iterations than one big patch.

src/display/svg.js
ahuglajbclajep: paintType can be only `1` or `2` in the [PDF specification](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdf_reference_archive/pdf_reference_1-7.pdf#page=292).
This implementation also includes processing when `paintType === 1`, so its `else` clause seems unnecessary.

timvandermeij: You're right. Thank you for clarifying this.

src/display/svg.js
```diff
@@ -982,12 +982,65 @@ SVGGraphics = class SVGGraphics {
...
+    }
+    this.executeOpTree(this.convertOpList(operatorList));
+
+    // Restore saved state.
+    this.svg = svg;
+    this.transformMatrix = transformMatrix;
+    this.fillColor = fillColor;
+    this.strokeColor = strokeColor;
+
+    tiling.appendChild(bbox.childNodes[0]);
```
timvandermeij: This line appears to break the PDF file from #10487, which also uses tiling patterns. Could you check why that happens?

timvandermeij (commented):
I think this is good to merge when the final comment is addressed. I have updated the title and the description of the pull request with additional details and which issues are fixed/improved by this patch.

After you've made the final changes, please squash the commits into one commit. You can check https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for how to do that easily. Thank you!

src/display/svg.js
ahuglajbclajep: The #10487 PDF uses the TilingPattern on the first page only.
Before implementing TilingPattern, paths that should be drawn in red were black with `fill=""`.
Looking at the output of this PR implementation, the rendering of `<svg:pattern>` using `<svg:g clip-path="...">` is broken.
I can infer from this that there might be a mistake in the implementation associated with clip, but I didn't know exactly what was wrong.

src/display/svg.js
```diff
@@ -982,12 +982,65 @@ SVGGraphics = class SVGGraphics {
...
+      this.current.fillColor = cssColor;
+      this.current.strokeColor = cssColor;
+    }
+    this.executeOpTree(this.convertOpList(operatorList));
+
+    // Restore saved state.
+    this.svg = svg;
+    this.transformMatrix = transformMatrix;
+    this.fillColor = fillColor;
+    this.strokeColor = strokeColor;
```
janpe2: When you restore the state, the lines should be
```
this.current.fillColor = fillColor;
this.current.strokeColor = strokeColor;
```

ahuglajbclajep:
@timvandermeij @janpe2 Please review it.

src/display/svg.js
timvandermeij: I have been thinking about this some more and I think it's best to leave fixing this for a follow-up. The PDF file already wasn't rendering correctly and there is an open issue for it. Given that this patch improves the other PDF files, I don't see this as a blocker for landing this.

timvandermeij (approved):

timvandermeij:
Nice work; thank you for improving the SVG back-end!

PR #10903, opened 2019-06-15
Move `PDFWorkerStream` and related code to its own file

Snuffleupagus:
Since all other `IPDFStream` implementations live in their own files, it seems reasonable for these to do so as well.

Furthermore, converts all of the relevant code to ES6 classes and updates the interface definitions to mark a couple of methods `async`.

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -14,15 +14,16 @@
  */
 
 import {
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #10909, opened 2019-06-16
fix a small typo in the german localization

fugidev:


Snuffleupagus:
Unfortunately this cannot be accepted, please see https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md for additional details:
> Note that the translations for PDF.js in the `l10n` folder are imported from the Nightly channel of Mozilla Firefox, such that we don't have to maintain them ourselves. This means that we will not accept pull requests that add new languages and/or modify existing translations, unless the corresponding changes have been made in Mozilla Firefox first.

PR #10925, opened 2019-06-23
Enable the `eslint-plugin-no-unsanitized` ESLint plugin to disallow unsafe usage of e.g. `innerHTML`

Snuffleupagus:
See https://github.com/mozilla/eslint-plugin-no-unsanitized

Since we've generally never allowed e.g. `innerHTML`, which is enforced during review, there's only one linting failure with this patch. (Which is white-listed, according to the existing comment and the fact that it's test-only code.)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,6 +7,7 @@
   "plugins": [
     "import",
     "mozilla",
+    "no-unsanitized",
     "unicorn",
   ],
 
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
Hmm, https://travis-ci.org/mozilla/pdf.js/builds/549334374#L568-L572

timvandermeij:
The four lines that failed the Travis CI build before are of the form `(0, function)(parameter1, parameter2);`, and I don't really know what this is trying to achieve anyway. Is this some syntax for either calling a real function if it exists and otherwise calling `0` as a dummy or something?

It looks like something we should report upstream, but it would be even better if we could find a way to write those four lines differently. They are certainly confusing to me and apparently also to the plugin...

timvandermeij (approved):
I have been thinking about this some more and I don't yet see an easy fix for the operator list file. Let's merge this because I do see advantages from the additional linting for all files except for one; removing the ignore in the operator list file can always be done at a later time. Thanks!

PR #10926, opened 2019-06-23
Change the signature of the `Parser` constructor to take a parameter object

Snuffleupagus:
A lot of the `new Parser()` call-sites look quite unwieldy/ugly as-is, with a bunch of somewhat randomly ordered arguments, which we can avoid by changing the constructor to accept an object instead. As an added bonus, this provides better documentation without having to add inline argument comments in the code.

Fixes https://github.com/mozilla/pdf.js/pull/10635#discussion_r264059493


timvandermeij (approved):

timvandermeij:
Looks much better. Thank you!

PR #10938, opened 2019-06-29
Reduce the number of `isCmd` calls slightly in the `XRef` class

Snuffleupagus:
This reduces the total number of function calls, when reading the XRef table respectively when fetching uncompressed XRef entries.
Note in particular the `XRef.readXRefTable` method, where there're *two* back-to-back `isCmd` checks rather than just one.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -20,7 +20,7 @@ import {
   warn
 } from '../shared/util';
 import {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #10947, opened 2019-07-06
Make the `find` helper function, in `src/core/document.js`, more efficient by using `peekBytes` rather reading the stream one byte at a time

Snuffleupagus:
*Please note:* A a similar change was attempted in PR #5005, but it was subsequently backed out in PR #5069.

Unfortunately I don't think anyone ever tried to debug *exactly* why it didn't work, since it ought to have worked, and having re-tested this now I'm not able to reproduce the problem any more. However, given just how inefficient the current code is, with thousands of strictly unnecessary function calls for each `find` invocation, I'd really like to try fixing this again.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -15,8 +15,8 @@
 /* eslint no-var: error */
 
 import {
-  assert, FormatError, info, isArrayBuffer, isBool, isNum, isSpace, isString,
-  OPS, shadow, stringToBytes, stringToPDFString, Util, warn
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you! I agree that this should really work, and looking at how `peekBytes` and `bytesToString` work, this should indeed be equivalent, but more efficient.

PR #10949, opened 2019-07-06
Delay initialization of searching, in the viewer, until the first page has rendered

Snuffleupagus:
When searching occurs for the first time in a document, the `textContent` of all pages will be fetched from the API. If there's a pending search operation when the document loads that will thus lead to a lot of `getTextContent` calls very early on, which may unnecessarily delay rendering of the first page. Generally, in the viewer, a number of non-essential API calls[1] will be deferred until the first page has been rendered, and there's no good reason as far as I can tell to handle searching differently.

---
[1] Such as e.g. `getOutline` and `getAttachments`.


timvandermeij (approved):

timvandermeij:
Looks like a good idea. Thank you!

PR #10953, opened 2019-07-08
Actually subclass Error

SamB:
Convert all the places where we do `klass.prototype = new Error()`
to ES6 classes, so stack traces work in (e.g.) Chrome.

Sadly, it seems we still need the `name` property, but we can at least
put it on the prototype in every case, rather than having a few classes
arbitrarily put it on instances instead.

Everything else that used to be added to constructor/prototype appears
to be either superfluous or automatic.

P.S. The "Classes" section of the Style Guide is woefully out of date.

### DIFF
--- a/src/core/core_utils.js
+++ b/src/core/core_utils.js
@@ -28,43 +28,20 @@ function getLookupTableFactory(initializer) {
   };
 }
 
-const MissingDataException = (function MissingDataExceptionClosure() {
-  function MissingDataException(begin, end) {
+class MissingDataException extends Error {
+  constructor(begin, end) {
+    super(`Missing data [${begin}, ${end})`);
     this.begin = begin;
     this.end = end;
-    this.message = `Missing data [${begin}, ${end})`;
   }
+}
+MissingDataException.prototype.name = 'MissingDataException';
 
-  MissingDataException.prototype = new Error();
-  MissingDataException.prototype.name = 'MissingDataException';
-  MissingDataException.constructor = MissingDataException;
+class XRefEntryException extends Error {}
+XRefEntryException.prototype.name = 'XRefEntryException';
 
-  return MissingDataException;
-})();
-
-const XRefEntryException = (function XRefEntryExceptionClosure() {
-  function XRefEntryException(msg) {
-    this.message = msg;
-  }
-
-  XRefEntryException.prototype = new Error();
-  XRefEntryException.prototype.name = 'XRefEntryException';
-  XRefEntryException.constructor = XRefEntryException;
-
-  return XRefEntryException;
-})();
-
-const XRefParseException = (function XRefParseExceptionClosure() {
-  function XRefParseException(msg) {
-    this.message = msg;
-  }
-
-  XRefParseException.prototype = new Error();
-  XRefParseException.prototype.name = 'XRefParseException';
-  XRefParseException.constructor = XRefParseException;
-
-  return XRefParseException;
-})();
+class XRefParseException extends Error {}
+XRefParseException.prototype.name = 'XRefParseException';
 
 /**
  * Get the value of an inheritable property.
--- a/src/core/jbig2.js
+++ b/src/core/jbig2.js
@@ -17,17 +17,12 @@ import { log2, readInt8, readUint16, readUint32, shadow } from '../shared/util';
 import { ArithmeticDecoder } from './arithmetic_decoder';
 import { CCITTFaxDecoder } from './ccitt';
 
-let Jbig2Error = (function Jbig2ErrorClosure() {
-  function Jbig2Error(msg) {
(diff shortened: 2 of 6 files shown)


src/core/core_utils.js
```diff
@@ -28,43 +28,20 @@ function getLookupTableFactory(initializer) {
...
-
-  return XRefEntryException;
-})();
-
-const XRefParseException = (function XRefParseExceptionClosure() {
-  function XRefParseException(msg) {
-    this.message = msg;
   }
+}
+MissingDataException.prototype.name = 'MissingDataException';
```
timvandermeij: Why can't we just set the `name` property in the constructor everywhere?

timvandermeij (commented):
I'm in favor of doing this. I tried it before and it didn't work, but I don't recall why exactly, so we'll see how this goes when we run the tests. I have also updated the style guide.

src/core/core_utils.js
SamB: Well, that would *work* ... if we wanted to actually write the constructor in every single subclass.

Compare:
```javascript
class XRefEntryException extends Error {}
XRefEntryException.prototype.name = 'XRefEntryException';
```
with:
```javascript
class XRefEntryException extends Error {
  constructor(message) {
    super(message);
    this.name = 'XRefEntryException';
  }
}
```
I think it takes much less time to see that the two-line version is a trivial Error subclass than it does with the ... however-many-line version. And it looks like there are 7 such trivial subclasses.

My original reason for doing it that way, though, is because that is where it was in the old code **and** on the standard Error classes (see [ES6](https://www.ecma-international.org/ecma-262/6.0/#sec-nativeerror.prototype.name) and [ESNext](https://tc39.es/ecma262/#sec-nativeerror.prototype.name)). And I probably also had some silly notion that this "saves space on instances" ;-).

Snuffleupagus (changes requested):
There's actually a *very* good reason that this sort of change hasn't already been done, and that's because it *completely* breaks the development viewer (i.e. the `gulp server` one) which is immediately obvious when testing locally.

Hence this sort of change is dependent on updating the `systemJS` dependency first, and as such this PR unfortunately cannot be accepted.

timvandermeij:
I have made the issue above to track the `systemjs` upgrade. If that is done, we can retry this, but for now I'm closing this since we cannot land it.

SamB:
Hmm.  I rather suspect that the problematic transformation is technically found in [systemjs-plugin-babel](https://github.com/systemjs/plugin-babel), which apparently bundles a custom build of an unspecified version of babel.

No, wait, it's specified in [build-babel/package.json](https://github.com/systemjs/plugin-babel/blob/master/build-babel/package.json). But what is that, *two* versions of babel-core?

... Anyway, I figured out how to get the viewer to start w/ "gulp server" and my changes; I'll send you the needed systemjs config changes later ...

timvandermeij:
Good to hear! If you get it to work, you can choose to open a new pull request or update this one and we'll reopen it.

Snuffleupagus:
> If you get it to work, you can choose to open a new pull request or update this one and we'll reopen it.

Before re-submitting this, please make sure to test the patch *carefully* since it currently throws `DataCloneError: The object could not be cloned.` when the various `...Exception`s are encountered within the `MessageHandler` class; this was seen when testing the built viewer locally (i.e. the `gulp generic` one).

SamB:
@Snuffleupagus Yeah, sorry I failed at testing before; my eyes kind of glazed over trying to scan https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing for stuff that seemed relevant to "did I mess up these Error classes?"; at some point I noticed that `package.json` defined a `test` script and just ran that, hoping that CI would do more.

Anyway, um, is `systemjs.config.js` used in production at all?

Snuffleupagus:
> my eyes kind of glazed over trying to scan https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing for stuff that seemed relevant to "did I mess up these Error classes?"

Generally speaking, whenever you're touching code in the `src/` folder running `gulp test`[1] locally is basically necessary to catch mistakes/errors. Just make to do the follow the "Generating reference images" step *first* though.

> Anyway, um, is `systemjs.config.js` used in production at all?

It shouldn't be, but please check https://github.com/mozilla/pdf.js/blob/master/gulpfile.js to make sure.

When it was added, it was only used for the development viewer (i.e. `gulp server`) and for various test suites; this list is *hopefully* complete https://github.com/mozilla/pdf.js/search?q=%22systemjs.config.js%22&unscoped_q=%22systemjs.config.js%22&type=Code

---
[1] Sometimes `gulp unittest` may suffice, but this isn't one of those times :-)

PR #10962, opened 2019-07-11
Prevent "Uncaught promise" messages in the console when cancelling `TextLayer` tasks (PR 10601 follow-up)

Snuffleupagus:
Since `finally` won't stop error propagation, this causes unnecessary messages to be printed in the console whenever a `TextLayer` task is cancelled.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #10963, opened 2019-07-11
Ensure that `PDFViewerApplication.{zoomIn, zoomOut}` won't run when PesentationMode is active (PR 10652 follow-up)

Snuffleupagus:
Similar to the `zoomReset` method we need to ensure that this code won't run for zoom events originating within the browser UI itself, since checks in e.g. the `keydown` event handler won't help in that case.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -412,6 +412,9 @@ let PDFViewerApplication = {
   },
 
   zoomIn(ticks) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #10966, opened 2019-07-12
Add timestamp to the page rendered event.

brendandahl:
This is needed to track rendering time in Firefox's talos performance
framework. See https://bugzilla.mozilla.org/show_bug.cgi?id=1565680


timvandermeij (approved):

timvandermeij:
Good idea to start tracking this!

PR #10968, opened 2019-07-13
Remove useless `wrapReason` calls in the `MessageHandler` class

Snuffleupagus:
Currently `wrapReason` is manually called at *every* `resolveOrReject` call-site, despite it being completely unnecessary unless there's an actual error being handled. This is obviously inefficient, and it's easy enough to avoid by having `resolveOrReject` handle this only when actually needed.


timvandermeij (approved):

timvandermeij:
Nice find!

PR #10969, opened 2019-07-13
Add an API unit-test for the `stopAtErrors` option (PRs 8240 and 8922 follow-up)

Snuffleupagus:
Also fixes an inconsistency in the 'PageError' handler, for `getOperatorList`, in the API.


timvandermeij (approved):

timvandermeij:
Good to have more unit test coverage for this. Thanks!

PR #10970, opened 2019-07-13
Simplify, and inline, the `finalize` function in the `MessageHandler` class

Snuffleupagus:
The `finalize` helper function has only a *single* call-site, and furthermore it's just a one-liner too. Furthermore it's only ever called with a `Promise` as its argument, meaning that it's unnecessarily convoluted as well (i.e. the `Promise.resolve()` part shouldn't be necessary).
Hence this code can be both simplified *and* inlined at its only call-site instead.

### DIFF
--- a/src/shared/message_handler.js
+++ b/src/shared/message_handler.js
@@ -60,10 +60,6 @@ function resolveOrReject(capability, data) {
   }
 }
 
-function finalize(promise) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #10974, opened 2019-07-15
Simplify the `PDFDocument.fingerprint` method slightly

Snuffleupagus:
The way that this method handles documents without an `ID` entry in the Trailer dictionary feels overly complicated to me. Hence this patch adds `getByteRange` methods to the various Stream implementations[1], and utilize that rather than manually calling `ensureRange` when computing a fallback `fingerprint`.

---
[1] Note that `PDFDocument` is only ever initialized with either a `Stream` or a `ChunkedStream`, hence why the `DecodeStream.getByteRange` method isn't implemented.

### DIFF
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -218,6 +218,12 @@ class ChunkedStream {
   }
 
   getByteRange(begin, end) {
+    if (begin < 0) {
+      begin = 0;
+    }
+    if (end > this.end) {
+      end = this.end;
+    }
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
*Updated to fix a stupid typo in the commit message, no actual code changes were made.*

timvandermeij (approved):

timvandermeij:
Looks better this way, also with the clamping in those methods. Thanks!

PR #10978, opened 2019-07-17
Scripting

mikedeboer:
The scripting API is be combined into a single JS file (generated by gulp), so that the JS runtime may inject it in one go.
I believe I also fixed the styling of form fields to be properly inherited, but that's faded from my memory a bit now. Changes there look good to cherry-pick in anyways.

In the end this look like a proposal more than a complete implementation.

Please also checkout the accompanying patch for m-c: https://gist.github.com/mikedeboer/c517509c52a474394840b9c295f3cf92

My apologies if I ended up wasting your time with this, but I think it might be a good starting point regardless. Please feel free to throw it away without feeling bad 😜

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -132,6 +132,10 @@
       "type": "boolean",
       "default": false
     },
+    "disableScripting": {
+      "type": "boolean",
+      "default": false
+    },
     "disablePageMode": {
       "description": "DEPRECATED.",
       "type": "boolean",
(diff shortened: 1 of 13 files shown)


timvandermeij:
I notice that I miss quite a bit of context here. Could you provide more information about the intent of the patch? My main question is: what kind of scripting are we talking about? Is there a link to the relevant part of the PDF specification?

timvandermeij:
Closing since there was no response on this for months. I think it's better to open an issue first to discuss the context for this patch, because as of right now we cannot merge this unfortunately.

Issue #4042, opened 2013-12-18
Zooming of pdf.js viewer is affected by zooming of other tabs from same origin


blerner:
Apologies if this is a known issue...  I'm using Firefox 26.0 on Windows, with whatever version of pdf.js is bundled there.

Steps to reproduce:

1) Have two Firefox windows open side by side, to best see the effect.  In one, go to http://A.org/somefile.pdf, and in the other, go to http://A.org/someotherpage.html.
2) Select the window with the PDF open.  Use Ctrl + -/+ to zoom in and out of the PDF.  Note that the PDF.js chrome (the navigation bar with the page counts and icons) stays a fixed size, while the content zooms.
3) Select the other window, and again use Ctrl + -/+ to zoom in and out of the page.  Note the effect on the PDF window -- the chrome and the content all zoom in and out too.  (Actually, it's even subtler: with Ctrl+ +, the navigation bar gets larger, and the content does too, until it reaches the width of the window minus some padding.  Then further Ctrl+ + causes the navigation bar to grow more, and the padding to grow a bit, and hence the content to shrink slightly.  With Ctrl+-, the same thing happens in reverse, where the navigation bar steadily shrinks, and the content grows slightly as the padding shrinks, then the content continues to shrink along with the navigation bar.)

Snuffleupagus:
The latest patch in https://bugzilla.mozilla.org/show_bug.cgi?id=786674 will, among other things, fix this issue as well.

PR #10979, opened 2019-07-17
[Firefox] Re-factor the 'zoomreset' message handling in the viewer (PR 10652 follow-up)

Snuffleupagus:
Given that this special-case only matters for the Firefox PDF viewer, it's probably better to just move it into `firefoxcom.js` instead to reduce unnecessary confusion.

---

Also, unless [bug 786674](https://bugzilla.mozilla.org/show_bug.cgi?id=786674) gets backed out for causing new intermittent failures in [bug 1224848](https://bugzilla.mozilla.org/show_bug.cgi?id=1224848) (which has a potential patch), issue #4042 should now be fixed.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -437,13 +437,8 @@ let PDFViewerApplication = {
     this.pdfViewer.currentScaleValue = newScale;
   },
 
-  zoomReset(ignoreDuplicate = false) {
+  zoomReset() {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good idea, thanks.

PR #10990, opened 2019-07-19
Refactor the `onBeforeDraw`/`onAfterDraw` functionality used in `BaseViewer` and `PDFPageView`

Snuffleupagus:
This functionality is very old, and pre-dates e.g. the introduction of the `EventBus` by a number of years. Rather than attaching two callback functions to every single `PDFPageView` instance, it's thus now possible to utilize the `EventBus` such that you only need a grand total of two listeners to achieve the same result.

For the `onAfterDraw` callback the replacement is particularly simple, given that a 'pagerendered' event is already being dispatched in the appropriate spot. An added benefit here is the ability to remove the event listener, since we only ever care about *one* (arbitrary) page being rendered for the `BaseViewer.onePageRendered` promise.

For the `onBeforeDraw` callback, a new 'pagerender' event was thus added to replace the callback.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -157,6 +157,7 @@ class BaseViewer {
 
     this.scroll = watchScroll(this.container, this._scrollUpdate.bind(this));
     this.presentationModeState = PresentationModeState.UNKNOWN;
+    this._onBeforeDraw = this._onAfterDraw = null;
     this._resetView();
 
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good! I really like the decoupling of components using the event bus.

PR #10993, opened 2019-07-20
Add the `docBaseUrl` API parameter to `AppOptions` in the viewer

Snuffleupagus:
This unfortunately required a bit of special handling, to correctly deal with the various extension builds.


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #10995, opened 2019-07-21
Remove an unnecessary `PDFDocumentProperties.setFileSize` call, relevant for the Firefox built-in viewer, and use the "normal" code-path in `PDFViewerApplication.open` instead

Snuffleupagus:
Since calling `getDocument` with a `PDFDataRangeTransport` argument will always unconditionally override a manually provided `length` argument, see https://github.com/mozilla/pdf.js/blob/a1a667809f286b761644bf34a4fb21dd524da8c8/src/display/api.js#L390-L394, this patch should thus be safe.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -512,11 +512,7 @@ let PDFViewerApplication = {
     }
     this.externalServices.initPassiveLoading({
       onOpenWithTransport(url, length, transport) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks a bit easier; thanks!

PR #10996, opened 2019-07-21
Avoid creating a `PDFFindBar` instance, in the Firefox built-in viewer, when not actually necessary

Snuffleupagus:
This is similar to how `PDFPresentationMode` isn't used when the Fullscreen API isn't supported.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -352,7 +352,9 @@ let PDFViewerApplication = {
     });
     pdfLinkService.setHistory(this.pdfHistory);
 
-    this.findBar = new PDFFindBar(appConfig.findBar, eventBus, this.l10n);
+    if (!this.supportsIntegratedFind) {
+      this.findBar = new PDFFindBar(appConfig.findBar, eventBus, this.l10n);
+    }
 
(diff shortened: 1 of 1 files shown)


web/viewer.js
```diff
@@ -72,7 +72,6 @@ function getViewerConfiguration() {
       next: document.getElementById('next'),
       zoomIn: document.getElementById('zoomIn'),
       zoomOut: document.getElementById('zoomOut'),
-      viewFind: document.getElementById('viewFind'),
```
timvandermeij: This is still used in https://github.com/mozilla/pdf.js/blob/71d9f5f860a7ba3e7971c202fad78f8efe4f5b89/web/app.js#L1642 as far as I can tell, so won't this break?

Snuffleupagus: D'oh, that's obviously correct; thanks for spotting this obvious error on my part!

I'm removing the second commit, since it wasn't all that important anyway.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #11003, opened 2019-07-23
Ensure that setting the `zoomDisabledTimeout` isn't skipped, regardless of the supported zoom keys, when handling mouse wheel events (PR 7097 follow-up)

Snuffleupagus:
*Possible follow-up:* It probably wouldn't hurt to try and shorten the `supportedMouseWheelZoomModifierKeys` name a bit, but I'm not attempting that here since it'd also require updating `PdfStreamConverter.jsm` in mozilla-central in order to be consistent.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2093,17 +2093,15 @@ function setZoomDisabledTimeout() {
 }
 
 function webViewerWheel(evt) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #11012, opened 2019-07-28
Reduce the number of function calls in `EvaluatorPreprocessor.read`

Snuffleupagus:
For very large and complex PDF files this will help performance slightly, since `EvaluatorPreprocessor.read` is called a lot during parsing in the worker.

This patch was tested using the PDF file from issue #2618, i.e. http://bugzilla-attachments.gnome.org/attachment.cgi?id=226471, using the following manifest file:
```
[
    {  "id": "issue2618",
       "md5": "",
       "rounds": 200,
       "type": "eq"
    }
]
```

```
Firefox | Overall      |   200 |         3402 |        3358 | -43 | -1.28 |        faster
Firefox | Page Request |   200 |            1 |           1 |   0 | 26.71 |
Firefox | Rendering    |   200 |         3401 |        3357 | -44 | -1.28 |        faster
```

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -21,7 +21,7 @@ import {
 } from '../shared/util';
 import { CMapFactory, IdentityCMap } from './cmap';
 import {
-  Dict, isCmd, isDict, isEOF, isName, isRef, isStream, Name
+  Cmd, Dict, EOF, isDict, isName, isRef, isStream, Name
 } from './primitives';
 import {
   ErrorFont, Font, FontFlags, getFontType, IdentityToUnicodeMap, ToUnicodeMap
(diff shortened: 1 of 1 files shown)


src/core/evaluator.js
```diff
@@ -3061,7 +3061,7 @@ var EvaluatorPreprocessor = (function EvaluatorPreprocessorClosure() {
           operation.args = args;
           return true;
         }
-        if (isEOF(obj)) {
+        if (obj === EOF) { // isEOF
```
timvandermeij: This comment looks a bit superfluous because the line itself is already very clear.

timvandermeij (approved):

timvandermeij:
Thanks!

Issue #6811, opened 2015-12-27
Support quad points for annotations


timvandermeij:
The PDF reference states that Link and text markup annotations can specify quad points. https://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/pdf_reference_1-7.pdf#page=634&zoom=auto,-246,530 explains this principle. Right now we create one container per annotation, but when quad points are defined, we need to create multiple containers with rectangles defined by a group of eight quad points. Below are known files that will benefit from quad points support, which are files where an annotation takes up multiple lines (which is not too common, fortunately):
- https://blogs.adobe.com/acrolaw/files/acrolaw/comment_print_button.pdf
- https://github.com/mozilla/pdf.js/files/37756/compressed.tracemonkey-pldi-09.pdf (from #6655)

It is also possible that the annotation's rectangle is not valid at all when quad points are specified, in which case we need to use the (at least first eight) quad points to determine the rectangle for the container. Known files are:
- http://linorg.usp.br/CTAN/macros/latex/contrib/pdfcomment/doc/pdfcomment.pdf (page 11)
- http://www.plpdf.com/examples/PLPDF-SDK/Simple_Examples/37_Markup_annotation/markup.pdf

THausherr:
https://issues.apache.org/jira/secure/attachment/12906841/PDFBOX-4074-011922-p5-Annotations.pdf

timvandermeij:
Notes to self for implementing this:

- https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=403&zoom=auto,-215,540
- https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=411&zoom=auto,-215,71
- https://jsfiddle.net/6eLqahpn

timvandermeij:
Quad points are now supported in the core and the API, so this issue is only about the viewer support now.

PR #11013, opened 2019-07-28
[api-minor] Implement quadpoints for annotations in the core layer

timvandermeij:
This commit introduces the required core layer support for #6811. In a next commit, we can introduce the display layer support for this feature.

Note that four annotation types that didn't have unit tests before (because they practically didn't have any logic) now have unit tests, which is also an improvement for our unit test coverage.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -146,6 +146,39 @@ class AnnotationFactory {
   }
 }
 
+function getQuadPoints(dict, rect) {
+  if (!dict.has('QuadPoints')) {
+    return null;
+  }
+
+  // The region is described as a number of quadrilaterals.
+  // Each quadrilateral must consist of eight coordinates.
+  const quadPoints = dict.getArray('QuadPoints');
+  if (!Array.isArray(quadPoints) || quadPoints.length % 8 > 0) {
+    return null;
+  }
+
+  const quadPointsLists = [];
+  for (let i = 0, ii = quadPoints.length / 8; i < ii; i++) {
+    // Each series of eight numbers represents the coordinates for one
+    // quadrilateral in the order [x1, y1, x2, y2, x3, y3, x4, y4].
+    // Convert this to an array of objects with x and y coordinates.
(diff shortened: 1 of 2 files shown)


timvandermeij:
@Snuffleupagus Do you perhaps have time to review this?

Snuffleupagus:
> @Snuffleupagus Do you perhaps have time to review this?

Sure, I'll try to take a look during the weekend.

src/core/annotation.js
```diff
@@ -146,6 +146,39 @@ class AnnotationFactory {
...
+  }
+
+  // The region is described as a number of quadrilaterals.
+  // Each quadrilateral must consist of eight coordinates.
+  const quadPoints = dict.getArray('QuadPoints');
+  if (!Array.isArray(quadPoints) || quadPoints.length % 8 > 0) {
+    return null;
+  }
+
+  const quadPointsLists = [];
```
Snuffleupagus: Nit: You could probably initialize this Array with the desired format right away, but I don't know how much it matters; e.g. something like this:
```javascript
const numQuadPoints = quadPoints.length / 8;
const quadPointsLists = new Array(numQuadPoints).fill([]);
for (let i = 0; i < numQuadPoints; i++) {
  ...
```

Snuffleupagus (approved):
Looks good to me (although I've not looked very closely at the spec).

src/core/annotation.js
timvandermeij: I tried this, but it doesn't work because `fill` does not really work well with objects. From the MDN docs: `When fill gets passed an object, it will copy the reference and fill the array with references to that object.` This has the undesirable effect that pushing to one of the arrays pushes the value to all of them, which made the unit tests fail. According to https://medium.com/@wisecobbler/4-ways-to-populate-an-array-in-javascript-836952aea79f we could use the `fill` approach with `map`, but that looks a bit strange and also copies the whole array again, so I don't think there is an advantage in doing that.

PR #11019, opened 2019-07-30
Decode URL encoded filenames from content disposition headers

wa8n:
pdf.js cannot correctly display urlencode contentDispositionFilename, so I add the decode logic of contentDispositionFilename in extractFilenameFromHeader.

### DIFF
--- a/src/display/network_utils.js
+++ b/src/display/network_utils.js
@@ -61,6 +61,11 @@ function extractFilenameFromHeader(getResponseHeader) {
   const contentDisposition = getResponseHeader('Content-Disposition');
   if (contentDisposition) {
     let filename = getFilenameFromContentDispositionHeader(contentDisposition);
+    if (filename.includes('%')) {
+      try {
+        filename = decodeURIComponent(filename);
+      } catch (ex) {}
+    }
     if (/\.pdf$/i.test(filename)) {
       return filename;
     }
--- a/test/unit/network_utils_spec.js
+++ b/test/unit/network_utils_spec.js
@@ -238,6 +238,20 @@ describe('network_utils', function() {
         }
         throw new Error(`Unexpected headerName: ${headerName}`);
       })).toEqual('filename.pdf');
+
+      expect(extractFilenameFromHeader((headerName) => {
+        if (headerName === 'Content-Disposition') {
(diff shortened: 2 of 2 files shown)


Snuffleupagus:
First of all, this needs a better and more readable *commit message* then just "support urlencode contentDispositionFilename", since it's a bit difficult to decipher when looking at only the commit message (e.g. at the `git` command line) and not the code itself. 
Second of all, please make sure to follow https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

src/display/network_utils.js
```diff
@@ -61,6 +61,13 @@ function extractFilenameFromHeader(getResponseHeader) {
   const contentDisposition = getResponseHeader('Content-Disposition');
   if (contentDisposition) {
     let filename = getFilenameFromContentDispositionHeader(contentDisposition);
+    if (filename.includes('%')) {
+      try {
+        filename = decodeURIComponent(filename);
+      } catch (ex) {
+        // URIError "Malformed URI", e.g. for "%AA.pdf"
```
Snuffleupagus: This looks like an incomplete copy of https://github.com/mozilla/pdf.js/blob/323b2eabcfc24e883e1a435828d94864bb46349f/web/ui_utils.js#L584-L587, but I don't think it matters too much here so just shortening the code to `} catch (ex) { }` would suffice here.

wangsongyan: thank you! I have changed my code.

test/unit/network_utils_spec.js
```diff
@@ -238,6 +238,13 @@ describe('network_utils', function() {
         }
         throw new Error(`Unexpected headerName: ${headerName}`);
       })).toEqual('filename.pdf');
+
+      expect(extractFilenameFromHeader((headerName) => {
```
timvandermeij: I think it would also be good to add another test case for a filename with a % character in it, but that's not actually URL encoded, for example `100%.pdf`. In that case, no transformation should be done.

wangsongyan: test cases have been added.

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #11020, opened 2019-07-30
Add a work-around, in `glyphlist.js`, for bad PDF generators which use a non-standard `/f_f` string in the `Encoding` dictionary when referring to the ff ligature (issue 11016)

### LINKED ISSUE #11016: characters "ff" are not captured from return value of getTextContent()
[Sample3.pdf](https://github.com/mozilla/pdf.js/files/3440869/Sample3.pdf)

Configuration:
- Web browser and its version: not using browser
- Operating system and its version: MacOS 10.14.6 hosting a local nodejs server
- PDF.js version: pdfjs-dist@2.1.266
- Is a browser extension: no

1. 
Run the following code snippet to print out the text to console.
The words such as "stuff", "offer" have the same issue that the "ff" is missing from the return value of the getTextContent() function, and become a weird character which cannot be compared or seen, which you can see the result from the screenshot below.
The code snippet is the following:
function getPDFText(data, callback, errorCallBack) {
    
    var loadingTask = pdfjsLib.getDocument(data);

        var maxPages = pdf._pdfInfo.numPages;
        var countPromises = []; // collecting all page promises

        for (var j = 1; j <= maxPages; j++) {

            var page = pdf.getPage(j);
    
            countPromises.push(page.then(function(page) { // add page promise
                var textContent = page.getTextContent();
                return textContent.then(function(text){ // return content promise
                    return text.items.map(function (s) { return s.str; }).join(''); // value page text 
                });
            }));
        }
        // Wait for all pages and join text
        Promise.all(countPromises).then(callback);
        
    }).catch(errorCallBack);
}

2.
Expect to see the word "stuff" in the console.log, but instead only see "stu"

textExtractedScreenShot

Snuffleupagus:
This patch will not incur any (measurable) overhead, since the glyphlist is already quite long and one more entry won't really matter, which is important given that this sort of PDF corruption ought to be very rare.

Furthermore, this patch purposely does *not* add a bunch of similarly modified ligature names on pure speculation. Any similar additions, for other ligatures, should only be made if there's real-world examples of PDF files where it's actually necessary.

Fixes #11016.


timvandermeij (approved):

timvandermeij:
Thanks!

/botio makeref

PR #11029, opened 2019-07-31
[api-minor] Update telemetry to use 'categorical' histograms.

brendandahl:
Firefox telemetry supports using string labels now. Convert our integers
that we used for categories to just use strings.

The upstream work will happen in:
https://bugzilla.mozilla.org/show_bug.cgi?id=1566882

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1137,8 +1137,10 @@ let PDFViewerApplication = {
 
       if (typeof PDFJSDev !== 'undefined' &&
           PDFJSDev.test('FIREFOX || MOZCENTRAL')) {
-        let versionId = String(info.PDFFormatVersion).slice(-1) | 0;
-        let generatorId = 0;
+        // Telemetry labels must be C++ variable friendly.
+        const versionId = `v${info.PDFFormatVersion.replace('.', '_')}`;
+        let generatorId = 'other';
+        // Keep these in sync with mozilla central's Histograms.json.
         const KNOWN_GENERATORS = [
           'acrobat distiller', 'acrobat pdfwriter', 'adobe livecycle',
           'adobe pdf library', 'adobe photoshop', 'ghostscript', 'tcpdf',
@@ -1151,7 +1153,7 @@ let PDFViewerApplication = {
             if (!generator.includes(s)) {
               return false;
             }
-            generatorId = i + 1;
+            generatorId = s.replace(/[ .\-]/g, '_');
             return true;
           }.bind(null, info.Producer.toLowerCase()));
         }
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -696,7 +696,7 @@ class Parser {
         return new CCITTFaxStream(stream, maybeLength, params);
       }
       if (name === 'RunLengthDecode' || name === 'RL') {
-        xrefStreamStats[StreamType.RL] = true;
+        xrefStreamStats[StreamType.RLX] = true;
         return new RunLengthStream(stream, maybeLength);
       }
       if (name === 'JBIG2Decode') {
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -153,30 +153,30 @@ const AnnotationBorderStyleType = {
 };
 
 const StreamType = {
-  UNKNOWN: 0,
-  FLATE: 1,
-  LZW: 2,
-  DCT: 3,
-  JPX: 4,
-  JBIG: 5,
-  A85: 6,
-  AHX: 7,
-  CCF: 8,
-  RL: 9,
+  UNKNOWN: 'UNKNOWN',
+  FLATE: 'FLATE',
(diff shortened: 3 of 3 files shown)


web/app.js
```diff
@@ -1739,14 +1741,41 @@ function webViewerPageRendered(evt) {
...
       PDFViewerApplication.externalServices.reportTelemetry({
         type: 'documentStats',
-        stats,
+        stats: newStats,
       });
     });
   }
 }
 
+function toTelemtryNames(lookup, stats) {
```
timvandermeij: `toTelemetryNames` (typo)

web/app.js
```diff
@@ -1739,14 +1741,41 @@ function webViewerPageRendered(evt) {
...
+  // Replace the integer types with the string name from the constant.
+  let ret = [];
```
timvandermeij: This can be `const`.

web/app.js
```diff
@@ -1739,14 +1741,41 @@ function webViewerPageRendered(evt) {
...
+  for (let key in stats) {
+    if (!stats[key]) {
+      continue;
+    }
+    ret.push(lookup[key].padEnd(3, 'X'));
```
timvandermeij: Why the padding? Is there a specification for how those names should look? I'm only aware of https://github.com/mozilla/pdf.js/wiki/Enumeration-Assignments-for-the-Telemetry-Histograms, is that also still relevant?

web/app.js
```diff
@@ -1739,14 +1741,41 @@ function webViewerPageRendered(evt) {
...
+  }
+  return ret;
+}
+
+function swapKeyValue(obj) {
+  let ret = {};
```
timvandermeij: This can be `const`.

web/app.js
```diff
@@ -1739,14 +1741,41 @@ function webViewerPageRendered(evt) {
     });
     // It is a good time to report stream and font types.
     PDFViewerApplication.pdfDocument.getStats().then(function (stats) {
+      // Convert the stats to telemetry friendly names.
+      let newStats = {
```
timvandermeij: This can be `const`.

web/app.js
```diff
@@ -1137,8 +1137,10 @@ let PDFViewerApplication = {
 
       if (typeof PDFJSDev !== 'undefined' &&
           PDFJSDev.test('FIREFOX || MOZCENTRAL')) {
-        let versionId = String(info.PDFFormatVersion).slice(-1) | 0;
-        let generatorId = 0;
+        // Telemetry labels must be C++ variable friendly.
+        let versionId = `v${info.PDFFormatVersion.replace('.', '_')}`;
```
timvandermeij: This can be `const`.

web/app.js
Snuffleupagus: Also, isn't the sort-of default in PDF.js to use `Object.create(null);` rather than `{}`, to avoid possible issues if somehow someone would mess with the Object prototype?

web/app.js
```diff
@@ -1739,14 +1741,41 @@ function webViewerPageRendered(evt) {
     });
     // It is a good time to report stream and font types.
     PDFViewerApplication.pdfDocument.getStats().then(function (stats) {
+      // Convert the stats to telemetry friendly names.
+      let newStats = {
+        streamTypes: toTelemtryNames(swapKeyValue(StreamType),
+                                     stats.streamTypes),
+        fontTypes: toTelemtryNames(swapKeyValue(FontType), stats.fontTypes),
```
Snuffleupagus: Would it make sense to do the swap *once* for each type, and cache that, rather than repeating it for every single page that's rendered?

web/app.js
brendandahl: [Labels](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/collection/histograms.html#labels) have to pass the regex `^[a-z][a-z0-9_]+[a-z0-9]$` (3 char minimum).

web/app.js
brendandahl: Yeah, I could do that. I've also considered just changing the constant int values to be strings, which would simplify all of this. This would technically be a major api change since we exposed these stats, but seeing that we didn't expose the StreamType and FontType constants, I doubt anyone is using them.

Snuffleupagus: > I've also considered just changing the constant int values to be strings, which would simplify all of this.

That does sound like the easiest solution here.

> This would technically be a major api change since we exposed these stats, but seeing that we didn't expose the StreamType and FontType constants, I doubt anyone is using them.

I wouldn't worry about that either, since I also don't believe that `getStats` is used much (or even at all) outside of mozilla-central.

Snuffleupagus (approved):
I really like the simplicity of the new patch; looks good to me :-)

timvandermeij (approved):

timvandermeij:
Looks good!

PR #11030, opened 2019-07-31
Finds matches when the answer is partially on the current page and

freeone3000:
partially on the next page.

### DIFF
--- a/test/unit/pdf_find_controller_spec.js
+++ b/test/unit/pdf_find_controller_spec.js
@@ -100,7 +100,9 @@ describe('pdf_find_controller', function() {
 
       eventBus.on('updatefindmatchescount',
           function onUpdateFindMatchesCount(evt) {
-        if (pdfFindController.pageMatches.length !== totalPages) {
+        if (pdfFindController.pageMatches.length !== totalPages + 1) {
+          console.error('not returning for ' +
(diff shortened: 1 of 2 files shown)


Snuffleupagus (commented):
I'm very sorry to have to say this, but in its current form it's most likely going to be *very* difficult for anyone to successfully review this (and be confident that the code works as intended).

Quite frankly, this is all somewhat difficult to read/follow, and the added complexity would hurt maintainability of this code quite a bit. E.g. dealing with any regressions could quickly become hard and time consuming.

I did spot one outright error when skimming over the code, please see the inline comment, which I suppose may contribute to some of the complexity added in later "WIP" patches!?

(Also, the comments are unfortunately not always that easy to parse either, since none of them contain "complete" sentences i.e. ones starting with an upper-case letter and ending with a period.)

It's also not really clear, just from the code, how this will affect the highlighting of search results in the TextLayer?

Given how the find functionality is currently implemented, i.e. only considering pages in isolation, it's even possible that some re-writing/re-factoring would be necessary (or at least a good idea) in order for something like this to be easier to implement (rather than trying to "shoe-horn" this new functionality into the existing code).

freeone3000:
Understood. I'll reconsider with that approach. Thanks for your time.

PR #11033, opened 2019-08-01
Ensure that the loading indicator, in the pageNumber input, is hidden when the viewer is closed

Snuffleupagus:
Currently the indicator may remain visible even after the document has been closed, which seems weird given that no page is either visible nor rendering :-)


timvandermeij (approved):

timvandermeij:
Good find!

PR #11034, opened 2019-08-01
Ensure that `ReadableStream`s are cancelled with actual Errors

Snuffleupagus:
There's a number of spots in the current code, and tests, where `cancel` methods are not called with appropriate arguments (leading to Promises not being rejected with Errors as intended).
In some cases the cancel `reason` is implicitly set to `undefined`, and in others the cancel `reason` is just a plain String. To address this inconsistency, the patch changes things such that cancelling is done with `AbortException`s everywhere instead.

### DIFF
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -567,14 +567,13 @@ class ChunkedStreamManager {
     return Math.floor((end - 1) / this.chunkSize) + 1;
   }
 
-  abort() {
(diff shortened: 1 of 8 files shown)


Snuffleupagus:
Check that the Linux failure is still intermittent...

/botio-linux unittest

timvandermeij (approved):

timvandermeij:
Thanks!

PR #11038, opened 2019-08-02
[api-minor] Fix completely broken `getStats` method by returning stats in Objects, rather than in Arrays (PR 11029 follow-up)

Snuffleupagus:
With the changes to the `StreamType`/`FontType` "enums" in PR #11029, one unfortunate result is that `getStats` now *always* returns empty Arrays. Something that everyone, myself included, apparently missed is that you obviously cannot index an Array with Strings :-)

I wrongly assumed that the unit-tests would catch any bugs, but they apparently suffered from the same issue as the code in `src/core/`.

Another possible option could perhaps be to use `Set`s, rather than objects, but that will require larger changes since `LoopbackPort` (in `src/display/api.js`) doesn't support them.

/cc @brendandahl 
/cc @timvandermeij

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -1046,8 +1046,8 @@ var XRef = (function XRefClosure() {
     // prepare the XRef cache
     this.cache = [];
     this.stats = {
-      streamTypes: [],
-      fontTypes: [],
+      streamTypes: Object.create(null),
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

PR #11047, opened 2019-08-05
Support corrupt PDF files with invalid/non-existent Group /CS entries (issue 11045)

### LINKED ISSUE #11045: Cannot display figure in a pdf file generated with pdflatex

[from_pdftex.pdf](https://github.com/mozilla/pdf.js/files/3467425/from_pdftex.pdf)

[slice2.pdf](https://github.com/mozilla/pdf.js/files/3467447/slice2.pdf)

Configuration:
- Web browser and its version: Firefox 68.0.1 and Google Chrome 76.0.3809.87
- Operating system and its version: macOS High Sierra 10.13.6（17G8030）
- PDF.js version: 2.2.228 beta
- Is a browser extension: No

1. Load the attached pdf file, [from_pdftex.pdf](https://github.com/mozilla/pdf.js/files/3467425/from_pdftex.pdf), with PDF.js.
2. Figure 1 is blank.

figure 1 blank

Figure 1 should be displayed as Adobe Reader displays.

- The pdf file, from_pdftex.pdf, is generated with pdflatex.
- The figure, [slice2.pdf](https://github.com/mozilla/pdf.js/files/3467447/slice2.pdf),  included in the pdf file is a pdf file created with Affinity Designer.
- First reported by @rubendsa  at James-Yu/LaTeX-Workshop/issues/1577. 
- Notice that PDF.js can display slice2.pdf itself.

A source TeX file:
```tex
\documentclass[12pt]{article}
\usepackage{graphicx}

\begin{document}

\begin{figure}[h]
\begin{center}
\includegraphics[height=1in]{slice2.pdf}
\caption{\textit{Locations of the neutral point ($X_{NP}$) and center of gravity ($X_{CG})$ of an individual wing segment. }}
\label{fig:test2}
\end{center}
\end{figure}

\end{document}
```

Snuffleupagus:
The PDF file in question tries to reference a non-existent ColorSpace, which should be quite rare in practice.

Fixes #11045


src/core/evaluator.js
```diff
@@ -263,8 +263,14 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           groupOptions.isolated = (group.get('I') || false);
           groupOptions.knockout = (group.get('K') || false);
           if (group.has('CS')) {
-            colorSpace = ColorSpace.parse(group.get('CS'), this.xref, resources,
-                                          this.pdfFunctionFactory);
+            colorSpace = group.get('CS');
+            if (colorSpace) {
+              colorSpace = ColorSpace.parse(colorSpace, this.xref, resources,
```
Snuffleupagus: Another possible solution would be use `try ... catch` instead, but given that similar bugs doesn't appear to be common that seems a bit unnecessary at the moment (also since it would probably be less efficient). In that case though, you'd probably only want to ignore errors if `this.options.ignoreErrors === true`.

test/pdfs/issue11045.pdf.link
```diff
@@ -0,0 +1 @@
+https://github.com/mozilla/pdf.js/files/3467425/from_pdftex.pdf
```
timvandermeij: Since this file is almost trivial, I'd prefer to have it in the repository instead of a linked testcase. I'll ask the author if we may include it.

timvandermeij (approved):

timvandermeij:
Thank you for the fix!

/botio makeref

PR #11048, opened 2019-08-05
Use more compact keys in `PDFLinkService._pagesRefCache`

Snuffleupagus:
By using the same internal formatting here as in the `Ref.toString` method, in `src/core/primitives.js`, all cache-keys will become at least two bytes shorter (and most three bytes shorter).
Obviously this won't have a huge effect on memory since there's only one cache entry per page, but it nonetheless seems wasteful to use longer keys than strictly required.

### DIFF
--- a/web/pdf_link_service.js
+++ b/web/pdf_link_service.js
@@ -344,12 +344,14 @@ class PDFLinkService {
     if (!pageRef) {
       return;
     }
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good find!

PR #11051, opened 2019-08-07
Actually compare the `cropBox` and `mediaBox` correctly in the `Page.view` getter

Snuffleupagus:
The current code will only consider the `cropBox` and `mediaBox` as equal when they both point to the *same* underlying Array. In the case where a PDF file actually specifies both boxes independently, with the exact same values in each, the comparison will currently fail and lead to an unneeded intersection computation.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -15,8 +15,9 @@
 /* eslint no-var: error */
 
 import {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
That's an interesting find. Nice work!

PR #11059, opened 2019-08-08
Fallback gracefully when encountering corrupt PDF files with empty /MediaBox and /CropBox entries

Snuffleupagus:
This is based on a real-world PDF file I encountered very recently[1], although I'm currently unable to recall where I saw it.
Note that different PDF viewers handle these sort of errors differently, with Adobe Reader outright failing to render the attached PDF file whereas PDFium mostly handles it "correctly".

The patch makes the following notable changes:
 - Refactor the `cropBox` and `mediaBox` getters, on the `Page`, to reduce unnecessary duplication. (This will also help in the future, if support for extracting additional page bounding boxes are added to the API.)
 - Ensure that the page bounding boxes, i.e. `cropBox` and `mediaBox`, are never empty to prevent issues/weirdness in the viewer.
 - Ensure that the `view` getter on the `Page` will never return an empty intersection of the `cropBox` and `mediaBox`.
 - Add an *optional* parameter to `Util.intersect`, to allow checking that the computed intersection isn't actually empty.
 - Change `Util.intersect` to have consistent return types, since Arrays are of type `Object` and falling back to returning a `Boolean` thus seem strange.

---

[1] In that case I believe that only the `cropBox` was empty, but it seemed like a good idea to attempt to fix a bunch of related cases all at once.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -95,24 +95,28 @@ class Page {
                   this._getInheritableProperty('Resources') || Dict.empty);
   }
 
-  get mediaBox() {
-    const mediaBox = this._getInheritableProperty('MediaBox',
-                                                  /* getArray = */ true);
-    // Reset invalid media box to letter size.
-    if (!Array.isArray(mediaBox) || mediaBox.length !== 4) {
-      return shadow(this, 'mediaBox', LETTER_SIZE_MEDIABOX);
+  _getBoundingBox(name) {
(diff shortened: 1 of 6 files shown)


timvandermeij (approved):

timvandermeij:
Nice find! The third page doesn't render at all in Okular, so we're doing better now :-)

/botio makeref

Issue #11066, opened 2019-08-11
copy/paste, additional characters appear


CrossMountain:
# Attach (recommended) or Link to PDF file here:
[copy-paste-problem.pdf](https://github.com/mozilla/pdf.js/files/3489985/copy-paste-problem.pdf)

# Configuration:
- Web browser and its version:
Chrome 75
- Operating system and its version:
Mac OS 10.14
- PDF.js version:
2.0.0
- Is a browser extension:
No

# Steps to reproduce the problem:
1. Copy the first line "电子科技大学", then paste it to somewhere
2. It will become "电⼦子科技⼤大学".

# What is the expected behavior? (add screenshot)
It should be "电子科技大学"。

# What went wrong? (add screenshot)
It is because of the font's ToUnicode map's problem. You can check the pdf file's structure, you will see the ToUnicode in the font, it maps <09fc> to <2f26 5b50> which are "⼦子⼦子". Also, it maps <095f> to <2f24 5927> which are "⼤大⼤大".
![ToUnicode](https://user-images.githubusercontent.com/32870413/62834287-97304380-bc7c-11e9-9264-16fe9b430b6b.png)

# How to generate this pdf file?
The pdf is generated in Mac OS.
First, write "电子科技大学" in an txt file. 
Second, click the "File".
Third, click the "Export As PDF"
Then, it becomes what you see.

# Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):
I don't know how to upload it to a site.

# The issue will be solved in my pull request.
https://github.com/mozilla/pdf.js/pull/11067

timvandermeij:
The pull request above contains a discussion about a possible solution that was tried but wasn't a full solution. We should check if/how other PDF viewers handle this and come up with a fix that doesn't negatively impact existing (valid) PDF files.

Snuffleupagus:
This is a bug in the PDF document *itself*, rather than in the PDF.js library, and please note that you get *identical* copy-and-paste behaviour as Adobe Reader as well (i.e. the PDF reference implementation).
Furthermore, as shown by the previously attempted PR, fixing this edge-case without regressing *correct* PDF documents in the process appear to be quite difficult. This problem should also not be a common one, since this issue is only such report we've received in either GitHub or Bugzilla (to the best of my knowledge).

All-in-all, closing as WONTFIX for now given the points above.

PR #11067, opened 2019-08-11
fix: a ToUnicode map may maps a character to two unicode

CrossMountain:
some pdf generators will generate a ToUnicode map, which maps a character to more than one unicode in CJK (China, Japan, Korea) character set.

You can see the pdf as a example.
[copy-paste-problem.pdf](https://github.com/mozilla/pdf.js/files/3490033/copy-paste-problem.pdf)

The issue will explain more:https://github.com/mozilla/pdf.js/issues/11066


src/core/evaluator.js
```diff
@@ -1415,9 +1415,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
           var glyphUnicode = glyph.unicode;
           var NormalizedUnicodes = getNormalizedUnicodes();
-          if (NormalizedUnicodes[glyphUnicode] !== undefined) {
-            glyphUnicode = NormalizedUnicodes[glyphUnicode];
-          }
+          glyphUnicode = keepUnifiedIdeograph(glyphUnicode, NormalizedUnicodes);
```
timvandermeij: It doesn't look like `NormalizedUnicodes` is used elsewhere in the file, so how about, instead of passing it as a parameter, removing `NormalizedUnicodes` and its import here and just call it in the other file where it's also defined?

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
   return s;
 }
 
+function keepUnifiedIdeograph(str, normalizedUnicodesMap) {
+  // a character id map to a unicode. replace it if it is not unified
+  if (str.length === 1) {
+    if (normalizedUnicodesMap[str]) {
```
timvandermeij: I would keep the `!== undefined` check here so we know the behavior is equal to the situation before.

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
...
+      return normalizedUnicodesMap[str];
+    }
+    return str;
+  }
+  // a character id map to multiple unicode. remove not unified unicodes
+  var outStr = '';
```
timvandermeij: `var` should be `let`.

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
...
+  for (var i = 0; i < str.length; i++) {
```
timvandermeij: `var` should be `let` and the string length can be cached, so this can become `for (let i = 0, ii = str.length; i < ii; i++) {`

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
...
+    var char = str[i];
```
timvandermeij: `var` can be `const` here since we're not altering the variable.

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
...
+    // if a char is not exist in the map, then it is a normal unicode
```
timvandermeij: `if the character does not exist in the map...`

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
...
+    if (!normalizedUnicodesMap[char]) {
```
timvandermeij: Like above, using `=== undefined` here is perhaps better since that is what the data structure uses.

timvandermeij (commented):
In general this looks good; some minor comments to make the code more similar to the other code we have and easier to read.

We should also add a reference test for this to prevent regressions. Please add a `text` test for this. For example https://github.com/mozilla/pdf.js/commit/be1e12dbcb2cdb91f01090e35e050dc316300038 shows how to do this: add the file in `test/pdfs`, add a Git ignore entry and add the file to the test manifest.

Thank you!

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
...
 
+function keepUnifiedIdeograph(str, normalizedUnicodesMap) {
+  // a character id map to a unicode. replace it if it is not unified
+  if (str.length === 1) {
+    if (normalizedUnicodesMap[str]) {
+      return normalizedUnicodesMap[str];
+    }
+    return str;
+  }
+  // a character id map to multiple unicode. remove not unified unicodes
```
timvandermeij: `Character maps to multiple Unicodes; remove ununified Unicodes.`

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
   return s;
 }
 
+function keepUnifiedIdeograph(str, normalizedUnicodesMap) {
+  // a character id map to a unicode. replace it if it is not unified
```
timvandermeij: `Character maps to a single Unicode; replace it if it's not unified.`

Snuffleupagus (changes requested):
I'm slightly worried about the general approach of this patch, since just outright ignoring parts of a multiple Unicode value seems wrong and may cause regressions elsewhere!
Wouldn't it make much more sense to actually normalize all of the parts of the Unicode value, and then eliminate duplicates? E.g. something like this:
```javascript
const unicodeBuf = [];
for (let i = 0, ii = str.length; i < ii; i++) {
  const unicode = str[i];
  const normalized = normalizedUnicodesMap[unicode];
  if (normalized === undefined) {
    unicodeBuf.push(unicode);
    continue;
  }
  // Check that a simple *normalized* Unicode value isn't equal to the next character.
  if (normalized.length === 1 && i < (ii - 1) && normalized === str[i + 1]) {
    warn(`normalizeUnicode - eliminating duplicate Unicode value: ${unicode}`);
    continue;
  }
  unicodeBuf.push(normalized);
}
return unicodeBuf.join('') || str; // Always fallback to returning the original value, since an empty string doesn't make sense.
```

src/core/evaluator.js
CrossMountain: 1. `NormalizedUnicodes` is generated by function `getNormalizedUnicodes` in the file `core/unicode.js`.  
2. The function `getNormalizedUnicodes` is also used in the file `test/unit/unicode_spec.js`
3. If I remove the `NormalizedUnicodes `, I must change the `test/unit/unicode_spec.js`. So I decide to leave it to the lib designer
4. Also, current way is good for function separation

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
   return s;
 }
 
+function keepUnifiedIdeograph(str, normalizedUnicodesMap) {
```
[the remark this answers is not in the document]
CrossMountain: The function's name is form https://en.wikipedia.org/wiki/CJK_Unified_Ideographs. 
You can see the https://unicode.org/charts/PDF/UF900.pdf.
I think it is a official name.

src/core/unicode.js
```diff
@@ -1630,8 +1630,29 @@ function reverseIfRtl(chars) {
...
+  // a character id map to multiple unicode. remove not unified unicodes
+  var outStr = '';
+  for (var i = 0; i < str.length; i++) {
+    var char = str[i];
+    // if a char is not exist in the map, then it is a normal unicode
+    if (!normalizedUnicodesMap[char]) {
+      outStr = outStr + char;
+    }
+  }
+  return outStr;
```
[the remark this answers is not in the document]
CrossMountain: If this returns an empty string, It indicates that there is no unified ideographs in the original text. I think removing it at all will be right.

src/core/unicode.js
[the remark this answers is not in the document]
Snuffleupagus: I think that the comment still applies, since the new function is doing much more than *just* handling the ideographs (i.e. the str.length === 1 case).

src/core/unicode.js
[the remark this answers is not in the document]
Snuffleupagus: There should be no case, in the current `master` branch, where `glyph.unicode` is an empty string; hence please don't change that here!

src/core/unicode.js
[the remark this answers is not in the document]
CrossMountain: 1. `str.length===1`, replace it  with unified unicode or return original text
2. `str.length>2`, keep the unified unicode
What goes wrong here? Can you explain that in detail?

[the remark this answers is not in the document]
Snuffleupagus: The name feels weird, given that *all* toUnicode values regardless if they're Unified Ideographs or not are passed through this function.

src/core/unicode.js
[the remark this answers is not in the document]
Snuffleupagus: Please note that Adobe Reader, i.e. the PDF reference implementation, is handling copying the same way as the PDF.js library in this case. That suggests that the PDF generator is at fault here, and in cases where you're trying to work-around errors in particular PDF file it's usually not a good idea if a patch changes behaviour in an edge-case (given the regression risk).

src/core/unicode.js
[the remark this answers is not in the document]
CrossMountain: Apart from the function name temporarily.
If don't pass the toUnicode values to a function, how to judge whether it is Unified Ideograph?
Are you talking about the performance problem? The performance is almost the same as original code. Both are O(N). N is the number of the toUnicode value.

src/core/unicode.js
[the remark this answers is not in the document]
CrossMountain: I understand what's your meaning. I will change it to not to return an empty string

src/core/unicode.js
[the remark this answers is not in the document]
Snuffleupagus: This isn't about performance, it's about a confusing *name* of new helper function.

src/core/unicode.js
```diff
@@ -1630,8 +1630,30 @@ function reverseIfRtl(chars) {
...
+  // Character maps to multiple Unicodes; remove ununified Unicodes.
+  let outStr = '';
+  for (let i = 0, ii = str.length; i < ii; i++) {
+    const char = str[i];
+    // If the character does not exist in the map, it's a normal Unicode.
+    if (normalizedUnicodesMap[char] === undefined) {
+      outStr = outStr + char;
+    }
+  }
+  // If outStr==='', return the original str
```
Snuffleupagus: This comment isn't necessary, since the code below is self-explanatory.

Also, https://github.com/mozilla/pdf.js/pull/11067#pullrequestreview-276234576 isn't really addressed in any way in the updated patch.

CrossMountain: What's the difference between my code and your code?
It looks difference. But works the same.

src/core/unicode.js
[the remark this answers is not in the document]
CrossMountain: I know what's your meaning. I will change it to normalizeUnicode

src/core/unicode.js
Snuffleupagus: The difference is that what I suggested will *only* eliminate outright duplicates (which is the case in your PDF file), whereas your code may remove *arbitrary* unicodes.

These particular examples are probably not realistic/correct, but just to illustrate the principle at hand:
What happens if a `toUnicode` entry contains a character mapped to e.g. `<2f26 2f24>`? Wouldn't you actually want that to be transformed to `<5b50 5927>` (which doesn't happen with this patch)?
Or if a `toUnicode` entry contains e.g. `<2f26 5927>`, would you expect that to be reduced to only `<5927>`?

Again, since the purpose here should be to fix the issue at hand without accidentally regressing something else, it seems that taking some care not the do too much may be helpful.

test/pdfs/.gitignore
```diff
@@ -353,3 +353,4 @@
 !issue9972-3.pdf
 !tiling-pattern-box.pdf
 !tiling-pattern-large-steps.pdf
+!issue11066.pdf
```
Snuffleupagus: Missing new-line at the end of file here.

test/test_manifest.json
```diff
@@ -4232,5 +4232,12 @@
       "rounds": 1,
       "link": false,
       "type": "eq"
+    },
+    { "id": "issue11066",
+      "file": "pdfs/issue11066.pdf",
+      "md5": "ee5b3dc2e80da466f1268f95b6e002a6",
+      "rounds": 1,
+      "link": false,
+      "type": "eq"
```
Snuffleupagus: This needs to be `"type": "text"` instead, since otherwise it's not actually testing the intended code.

src/core/unicode.js
CrossMountain: You really think further than me. I will change it later

timvandermeij:
@Snuffleupagus Could you perhaps check this again? I'm not too familiar with the Unicode handling code unfortunately to properly review this.

src/core/unicode.js
```diff
@@ -1630,8 +1630,34 @@ function reverseIfRtl(chars) {
   return s;
 }
 
+function normalizeUnicode(str, normalizedUnicodesMap) {
```
Snuffleupagus: For performance reasons I still think we need a fast-path for the `str.length === 1` case here, i.e. what you had in a prior version of this patch:
```javascript
if (str.length === 1) {
  if (normalizedUnicodesMap[str] !== undefined) {
    return normalizedUnicodesMap[str];
  }
  return str;
}
```

src/core/unicode.js
```diff
@@ -1630,8 +1630,34 @@ function reverseIfRtl(chars) {
...
+  const unicodeBuf = [];
+  for (let i = 0, ii = str.length; i < ii; i++) {
+    const unicode = str[i];
+    const normalized = normalizedUnicodesMap[unicode];
+    if (normalized === undefined) {
+      // If a toUnicode entry contains two same unified Unicode
```
Snuffleupagus: Let's replace `same` with `identical`, and please remember to end all sentences with a period.

Snuffleupagus (changes requested):
Sorry about the delay in getting back to this (I was busy with other things the last few days).

Apparently this causes regressions in existing tests, which needs to be addressed; please see https://github.com/mozilla/pdf.js/wiki/Contributing for information about running tests locally.

I'd suggest reverting this to something similar to https://github.com/mozilla/pdf.js/pull/11067#pullrequestreview-276234576 in order to limit the regression risk here, since what you're attempting to fix is very much an edge case.

CrossMountain:
Sorry that I am getting back to this after days. I fix it a little again

CrossMountain:
Can somebody check this?

src/core/unicode.js
```diff
@@ -1630,8 +1630,41 @@ function reverseIfRtl(chars) {
...
+    return str;
+  }
+
+  const unicodeBuf = [];
+  for (let i = 0, ii = str.length; i < ii; i++) {
+    const unicode = str[i];
+    const normalized = normalizedUnicodesMap[unicode];
+    if (normalized === undefined) {
+      // If a toUnicode entry contains two same unified Unicode
+      if (unicodeBuf.includes(unicode) === false) {
```
timvandermeij: Instead of `=== false`, use `!unicodeBuf.includes(unicode)` instead.

src/core/unicode.js
```diff
@@ -1630,8 +1630,41 @@ function reverseIfRtl(chars) {
...
+        unicodeBuf.push(unicode);
+      }
+      continue;
+    }
+    // If a not unified and *simple* Unicode
+    if (normalized.length === 1) {
+      // If a toUnicode entry contains two same not unified Unicode
+      if (str.includes(normalized) === true ||
```
timvandermeij: You can simplify this to `if (str.includes(normalized) || unicodeBuf.includes(normalized)) {` so it probably even fits on one line. There is no need to check for `=== true` since it's already a boolean condition.

src/core/evaluator.js
```diff
@@ -1415,9 +1415,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
           var glyphUnicode = glyph.unicode;
           var NormalizedUnicodes = getNormalizedUnicodes();
-          if (NormalizedUnicodes[glyphUnicode] !== undefined) {
-            glyphUnicode = NormalizedUnicodes[glyphUnicode];
-          }
+          glyphUnicode = normalizeUnicode(glyphUnicode, NormalizedUnicodes);
```
timvandermeij: I don't think `NormalizedUnicodes` is used elsewhere in this file, so it looks a bit strange to pass it in as a parameter since it's not really a variable. Therefore I think we should just move the call and the import to the other file altogether and remove this parameter. Looking at https://github.com/mozilla/pdf.js/pull/11067#discussion_r315158836, what I meant was not to move it out of `unicode.js`, but simply to move the variable above to the other file.

timvandermeij (commented):
These are the final review comments as far as I can review this. I don't know enough about the unicode code to review that; perhaps @Snuffleupagus has some time to review that part?

src/core/unicode.js
Snuffleupagus: This entire check is definitely not correct, and it should thus explain the test failures seen with this patch.

Generally speaking, the regression clearly outlines that there's no reason why a ToUnicode entry cannot contain the same glyph more than once.

src/core/unicode.js
Snuffleupagus: If the condition above isn't correct, and causes regressions, then I really cannot see why this check is generally OK either.

Snuffleupagus (changes requested):
First of all, it seems that the latest version of this patch doesn't address the test failures as seen above. (Furthermore it seems that my last round of review comments were only partially addressed.)

However, looking at the test failures in more detail I've unfortunately come to the conclusion that the approach in this patch isn't going to be generally acceptable.
As can be seen from the test-failures, the general assumption that a ToUnicode entry cannot contain the same glyph more than once simply isn't correct unfortunately.

Based on these findings, I'm now not even convinced that a much more limited patch (along the lines of https://github.com/mozilla/pdf.js/pull/11067#pullrequestreview-276234576) would be acceptable either.

Please keep in mind that the error is with the PDF generator itself, since even Adobe Reader (i.e. the PDF reference implementation) handles text-selection and copying the same way as the PDF.js library. 
For cases such as these, when you're dealing with a bad PDF generator, it's imperative that a work-around cannot accidentally impact correct PDF files in any way. (And trying to tailor a work-around to *one* particular bad PDF generator can be difficult.)

@timvandermeij We should obviously keep the issue open, but I'd suggest closing the PR since it's not at all clear to me just how to adjust it such that errors/regressions cannot occur.

timvandermeij:
Thank you both for looking into this. It's indeed best to go back to the drawing board in that case since we don't want to regress existing PDF files, especially for broken PDF files that even Adobe Reader can't handle properly. Let's keep the issue open to track this problem, but close this PR.

Issue #2504, opened 2012-12-24
Slow performance


okbtsnr:
http://dl-ctlg.panasonic.com/jp/manual/sd/sd_rbm1000_0.pdf

PR #11069, opened 2019-08-14
Use streams for OperatorList chunking (issue 10023)

Snuffleupagus:
*Please note:* The majority of this patch was written by Yury, and it's simply been rebased and slightly extended to prevent issues when dealing with `RenderingCancelledException`.

By leveraging streams this (finally) provides a simple way in which parsing can be aborted on the worker-thread, which will ultimately help save resources.
With this patch worker-thread parsing will *only* be aborted when the document is destroyed, and not when rendering is cancelled. There's a couple of reasons for this:

 - The API currently expects the *entire* OperatorList to be extracted, or an Error to occur, once it's been started. Hence additional re-factoring/re-writing of the API code will be necessary to properly support cancelling and re-starting of OperatorList parsing in cases where the `lastChunk` hasn't yet been seen.
 - Even with the above addressed, immediately cancelling when encountering a `RenderingCancelledException` will lead to worse performance in e.g. the default viewer. When zooming and/or rotation of the document occurs it's very likely that `cancel` will be (almost) immediately followed by a new `render` call. In that case you'd obviously *not* want to abort parsing on the worker-thread, since then you'd risk throwing away a partially parsed Page and thus be forced to re-parse it again which will regress perceived performance.
 - This patch is already *somewhat* risky, given that it touches fundamentally important/critical code, and trying to keep it somewhat small should hopefully reduce the risk of regressions (and simplify reviewing as well).

Time permitting, once this has landed and been in Nightly for awhile, I'll try to work on the remaining points outlined above.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -195,7 +195,7 @@ class Page {
     });
   }
 
-  getOperatorList({ handler, task, intent, renderInteractiveForms, }) {
+  getOperatorList({ handler, sink, task, intent, renderInteractiveForms, }) {
     const contentStreamPromise = this.pdfManager.ensure(this,
                                                         'getContentStream');
     const resourcesPromise = this.loadResources([
@@ -220,7 +220,7 @@ class Page {
 
     const dataPromises = Promise.all([contentStreamPromise, resourcesPromise]);
     const pageListPromise = dataPromises.then(([contentStream]) => {
-      const opList = new OperatorList(intent, handler, this.pageIndex);
+      const opList = new OperatorList(intent, sink, this.pageIndex);
 
       handler.send('StartRenderPage', {
         transparency: partialEvaluator.hasBlendModes(this.resources),
@@ -244,7 +244,7 @@ class Page {
         function([pageOpList, annotations]) {
       if (annotations.length === 0) {
         pageOpList.flush(true);
-        return pageOpList;
+        return { length: pageOpList.totalLength, };
       }
 
       // Collect the operator list promises for the annotations. Each promise
@@ -264,7 +264,7 @@ class Page {
         }
         pageOpList.addOp(OPS.endAnnotations, []);
         pageOpList.flush(true);
-        return pageOpList;
+        return { length: pageOpList.totalLength, };
       });
     });
   }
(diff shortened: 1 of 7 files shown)


timvandermeij:
> [...] which will ultimately help save resources.

I don't entirely understand how that's different from what we currently have. Now we have `RenderPageChunk`, so only the chunks that are ready are actually send from the worker thread to the main thread. In this patch that's replaced by enqueueing the chunk on the sink, but at the point the chunk must also be ready, so the same resources go into preparing the chunk. Where does the resource saving come from then?

Snuffleupagus:
> Where does the resource saving come from then? I'm most likely missing something here, so it would be great if you have some more details about how this works.

By using `ReadableStream`s it's possible to *abort* the worker-thread parsing of an OperatorList, and not just stop the main-thread rendering. In this patch that only happens when `destroy`ing the document, but it will be possible to extend it to happen when rendering is cancelled as well.

Please keep in mind that with the current code, once `PartialEvaluator.getOperatorList` is called it will *always* run to completion (unless there's an error). This means that even pages which have e.g. been evicted from the `PDFPageViewBuffer` in the viewer, and thus have be cancelled, will still compete for resources on the worker-thread thus slowing down parsing/rendering of the visible pages.

Snuffleupagus:
> [...] I'll try to work on the remaining points outlined above.

I've been playing locally with a follow-up patch which also cancels the stream on `RenderingCancelledException`, using a short timeout, and the results look quite promising.

For example, testing with the PDF file from issue #2504 (where PDF.js performance is quite bad) locally and *immediately* upon file load scroll quickly to the end (such that rendering is triggered for all pages):
With the current `master` (or this patch), the time it takes to render the *last* page is usually between 30 and 40 seconds. With the stream cancelling on render cancel added, that time is cut in *half* which demonstrates just how much cancelling parsing on the worker-thread can help :-)

Snuffleupagus:
> I've been playing locally with a follow-up patch which also cancels the stream on `RenderingCancelledException`, using a short timeout, and the results look quite promising.

See https://github.com/Snuffleupagus/pdf.js/compare/getoplist-stream...Snuffleupagus:getOperatorList-cancel

timvandermeij (approved):

timvandermeij:
It took some time to look into how this works exactly, but it really looks good and it's also a bit easier to follow without the handler. Nice work!

PR #11070, opened 2019-08-16
Inline the `isString` check in the `Parser.getObj` method

Snuffleupagus:
For very large and complex PDF files this will help performance *slightly*, since `Parser.getObj` is called *a lot* during parsing in the worker.

This patch was tested using the PDF file from issue #2618, i.e. http://bugzilla-attachments.gnome.org/attachment.cgi?id=226471, with the following manifest file:
```
[
    {  "id": "issue2618",
       "md5": "",
       "rounds": 200,
       "type": "eq"
    }
]
```

```
Firefox | Overall      |   200 |         2847 |        2830 | -17 | -0.60 |        faster
Firefox | Page Request |   200 |            2 |           2 |   0 | -7.14 |
Firefox | Rendering    |   200 |         2844 |        2827 | -17 | -0.60 |        faster
```

### DIFF
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -89,7 +89,7 @@ class Parser {
     }
   }
 
-  getObj(cipherTransform) {
+  getObj(cipherTransform = null) {
     const buf1 = this.buf1;
     this.shift();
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #11073, opened 2019-08-16
Move polyfill for codePointAt to String prototype.

brendandahl:
This method belongs on the prototype not the String object.


timvandermeij (approved):

timvandermeij:
Nice find!

Issue #3528, opened 2013-07-31
Themes & Skins for pdf.js


milkovsky:
Are there any developed themes or skins for pdf.js exept <a href="http://mozilla.github.io/pdf.js/web/viewer.html">Dark theme</a>? I tried to find any opened issue for it. Then I googled but I haven't find any developed theme for pdf.js.

ibuilder:
Should just style with bootstrap and fontawesome (replace bulky images).  Any bootstrap skin would work then and there would be no need to build theming support.

speedplane:
For those looking for skinning inspiration, here is one I know of based on PDF.js:
https://peerlibrary.org/p/2uX4z4P2nGGwLQTNt/structural-adaptation-via-l_p-norm-oracle-inequalities

declanqian:
It would be awesome if there is other theme. I am currently using the pdf viewer Chrome extension to view pdfs and found it hard to change the theme by change the code which is installed directly from chrome web store.

oytuntez:
Or we could at least provide a dynamic CSS file, such as SCSS, Sass etc; so everybody can quickly implement their own small customizations such as colors and fonts.

medikoo:
I'm fine with official skin, however documentation says that it should not be used in unmodified form.

Therefore it'll be great to be provided with at least single skin that can be openly used without license issues. Does anyone know one?

timvandermeij:
I'm going to link #6706 here as it is also a bit relevant for this issue.

If anyone would like to work on making the CSS more generic and easier to adjust, we can review PRs that do so.

LordTaifleh:
I didn't find this issue at first, so i had opened another already. What is your status @milkovsky ?

milkovsky:
@LordTaifleh My status: I am not doing any development here. In the end I used the default theme because of lack of project budget.

LordTaifleh:
Alright,
I went with a solution via Firefox-plugin Stylish, making a template for overriding default style of pdf-js like this:
```
@namespace url(http://www.w3.org/1999/xhtml);

@-moz-document regexp(".*\.(p|P)(d|D)(f|F)") {
  
#toolbarContainer, .findbar, .secondaryToolbar {
	 background-color: #2f343f !important; 
	 background-image: url(images/texture.png);
}
body {
	background-color: #2f343f !important;;
	background-image: url(images/texture.png);
}
	#toolbarSidebar {

  background-color: #2f343f !important; 
	background-image: url(images/texture.png);
[...]

PR #11077, opened 2019-08-18
Implement Photon design for the viewer

sawanm9000:
- Flat icons
- Larger buttons make touch input easier
- No textures, unnecessary shadows and gradients - The overall UI is less taxing on the brain.
- No useless animations - Actions like hovering is registered fast - The overall UI now feels snappy.
- Lighter (as in kilobytes to load)
- New UI match with Firefox's UI.

Edit: Latest screenshot:
PDF js Dark

PDF js Light

Old look for comparison:
OLD PDF js


Snuffleupagus:
Please note that this *may* apply here as well: https://github.com/mozilla/pdf.js/pull/10673#issuecomment-475874671, since the current UI was based on the only existing official design specification.

(In any case, please keep in mind that none of the `package.json` related changes belong in a patch like this.)

Snuffleupagus:
Referencing issue #9702 (unfortunately there's not, as far as I know, an official Photon-compatible design specification for the PDF.js viewer).
 
*Please keep in mind that none of the `package.json` related changes belong in a patch like this.*

Snuffleupagus:
Both of the files `package.json` and ` package-lock.json` must be completely removed from the patch, and please make sure to follow also see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

Snuffleupagus:
Both of the files `package.json` and ` package-lock.json` must be completely removed from the patch, and please make sure to follow https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

web/viewer.css
```diff
@@ -643,14 +641,20 @@ html[dir='rtl'] .findbar .splitToolbarButton .toolbarButton:last-child {
 }
 
 .dropdownToolbarButton {
-  width: 120px;
-  max-width: 120px;
+  min-width: 158px;
+  max-width: 158px;
```
Snuffleupagus: Increasing these values will lead to *a lot* of wasted space for locales where the relevant strings are short, please revert this change!

web/viewer.css
```diff
@@ -682,19 +686,19 @@ html[dir='rtl'] .dropdownToolbarButton:after {
 }
 
 .dropdownToolbarButton > select {
-  min-width: 140px;
-  font-size: 12px;
+  width: 100%;
```
Snuffleupagus: Similar to the comment above, please revert this since it leads to an incorrect/inconsistent width until the `select` has been localized.

Snuffleupagus (changes requested):
Both of the files `package.json` and `package-lock.json` must be completely removed from the patch, and please make sure to follow https://github.com/mozilla/pdf.js/wiki/Squashing-Commits.

---

> On the right-hand side of the Zoom dropdown, the border looks "weird".

This seems to be a more general problem, which actually applies to *every* single button on the main Toolbar.

Also, why does both the Zoom dropdown button, the Find next/previous buttons, and the Overlay buttons have a lighter background than every other button? This honestly looks quite out of place.

*Edit:* Furthermore, in some situations the pageNumber focus ring is now incorrectly red, see e.g. 
pageNumber_focus

*Edit2:* Also, assuming that the general design is deemed OK: I'm guessing that this is probably something which will benefit from being broken up into a couple of *separate* patches, to simplify testing *and* to reduce the risk of regressions!?

utopianknight:
modern-look

The findbar now looks a lot like that of Firefox:
Firefox findbar

timvandermeij:
I'm noticing the following things:

- When the viewer is loading, the page number indicator has a red border.
- Almost all icons appear to be off-center on hover (hover over the + in the toolbar for example and notice that the space on the right/bottom is larger than on the left/top).
- If I compare the design with the existing one, the background color of the toolbar stands out for me. The one from the existing design has a color that is almost equal to the background color of the viewer itself, whereas in this design the toolbar appears to be a lot lighter, which looks a bit strange to me.

Aside from the above, I do see improvements in this design over the existing one, but I'm definitely not a UI/UX designer. @shorlander Is this something you can chime in on? Is this something that would fit in with the current Firefox style or do you/others have different plans altogether for #9702 for example?

utopianknight:
> If I compare the design with the existing one, the background color of the toolbar stands out for me. The one from the existing design has a color that is almost equal to the background color of the viewer itself, whereas in this design the toolbar appears to be a lot lighter, which looks a bit strange to me.

The colors have been updated. The scrollbars are now dark.

At this point, since there are no [guidelines ](https://design.firefox.com/photon/components/buttons.html)on some components, suggestions would be more helpful than criticisms.

designakt:
@bwinton asked us to review the new UI before integrating into Firefox.

@aaronrbenson and I agree that this is a really nice match to other Firefox UI.
We only noticed a few things:

- **a few minor tweaks**
  - in Firefox we do not need the “open with” icon _(currently don't have that either)_
  - when clicking download, default should be safe file _(as is now)_
  - the hover area seems to be bigger then on other Firefox buttons - would be great to align
  - on mac the dropdown seems to have a glitch on top and bottom 
    CleanShot 2020-07-15 at 09 23 26@2x

- **a really nice to have**
  Ideally we would also have the toolbar align with Firefox toolbar style - **dark in dark mode, and light in light mode**. - With a line separating the pdf toolbar from what is above.

- **and a question**
  We like the new presentation mode icon, but were wondering what got us to use this over the previous symbol. Looking at it we noticed that we seem to mix "presentation mode" and "fullscreen" descriptions in this feature. **Do we know that presentation mode is better understood than fullscreen?**

For any follow-up questions please ping @aaronrbenson

brendandahl:
@utopianknight I know this is very delayed, but do you think you'd have time to address the above comments and we'll get this merged in?

utopianknight:
> on mac the dropdown seems to have a glitch on top and bottom

I don't have a Mac so I'm going to need help on this.

> Do we know that presentation mode is better understood than fullscreen?

Fullscreen button is for going full screen (it has its own icon). Presentation button is to switch to presentation mode. The reason why I changed the icon is so that it best communicate what the button actually **does**, which is to switch to presentation mode.

Question:
Pinging @aaronrbenson
This is how the dropdown button and a field look like respectively:
DropDownFieldButton

The reason why I changed the 'Zoom' dropdown button to make it look like a field is so that the user can type in the exact zoom level and fine-tune as they wish, while also be able to select zoom preset from a dropdown menu. This might be a solution to issue #1260. So which would be best? Should I make look like a dropdown button or leave it as it is?

@brendandahl
I'll work on it during the weekends.

nt1m:
If this is blocked on the macOS select issue, I would recommend ignoring it for now, since this affects all `<select>` in general. I'm happy to look into fixing the root issue, though I'm first looking if there's an existing bug filed.

nt1m:
I'm looking into it in: https://bugzilla.mozilla.org/show_bug.cgi?id=1653382

utopianknight:
New update: Photon Design with dark and light theme and SVG icons.

[Compare](https://github.com/mozilla/pdf.js/compare/master...utopianknight:modern-look)

PDF js Dark

PDF js Light

High DPI preview:

PDF js Dark HiDPI

PDF js Light HiDPI

bwinton:
@utopianknight It's pretty exciting to see the progress on this! Do you know when it will be ready for another UX review (or perhaps it's ready now)? 😃

utopianknight:
@bwinton I don't know what the process is. I believe I'm finished with it but I'm also expecting your feedback so I can make further changes if necessary.

aaronrbenson:
This is looking really good, thanks for your help @utopianknight! I can put some more time and thought into this next Monday and provide more feedback (if any) but for now I think these improvements look great. :shipit:

web/viewer.css
```diff
@@ -15,11 +15,79 @@
...
 @import url(pdf_viewer.css);
 
+/** ############### VARIABLES ############### **/
+
+/* ------------ Initial ------------ */
+
 :root {
   --sidebar-width: 200px;
-  --sidebar-transition-duration: 200ms;
-  --sidebar-transition-timing-function: ease;
```
Snuffleupagus, line 21: These CSS variables should *not* be removed, please revert those changes.

web/viewer.css
```diff
@@ -15,11 +15,79 @@
 
 @import url(pdf_viewer.css);
 
+/** ############### VARIABLES ############### **/
+
+/* ------------ Initial ------------ */
+
```
Snuffleupagus, line 21: Unnecessary comment, please remove.

web/viewer.css
```diff
@@ -15,11 +15,79 @@
...
+  --outline-active-bg-color: rgba(0, 0, 0, 1);
+  --sidebaritem-bg-color: rgba(0, 0, 0, 0.15);
+  --doorhanger-bg-color: rgb(255, 255, 255);
+  --doorhanger-border-color: rgba(12, 12, 13, 0.2);
+  --doorhanger-hover-color: #ededed;
+  --doorhanger-separator-color: #dedede;
+  --doorhanger-icon-color: rgba(12, 12, 13, .9);
+}
+
+/* ------------ Dark Theme ------------ */
```
Snuffleupagus, line 54: The code below is pretty self-explanatory, hence this comment can be removed.

web/viewer.css
```diff
@@ -15,11 +15,79 @@
...
+    --sidebaritem-bg-color: rgba(255, 255, 255, 0.15);
+    --doorhanger-bg-color: rgb(74, 74, 79);
+    --doorhanger-border-color: #27272b;
+    --doorhanger-hover-color: #5C5C61;
+    --doorhanger-separator-color: #5c5c61;
+    --doorhanger-icon-color: rgba(249, 249, 250, .9);
+  }
+}
+
+/** ######################################### **/
```
Snuffleupagus, line 90: Unnecessary comment, please remove.

web/viewer.css
```diff
@@ -1362,12 +1474,10 @@ html[dir='rtl'] .outlineItemToggler::before {
...
   padding: 15px;
   border-spacing: 4px;
-  color: rgba(217, 217, 217, 1);
+  color: var(--main-color);
   font-size: 12px;
   line-height: 14px;
-  background-color: rgba(71, 71, 71, 1); /* fallback */
-  background-image: url(images/texture.png),
-                    linear-gradient(rgba(82, 82, 82,0.99), rgba(69, 69, 69, 0.95));
+  background-color: var(--doorhanger-bg-color); /* fallback */
```
Snuffleupagus, line 1481: Leftover comment at the end?

timvandermeij:
The only thing I found is that the bookmark button has much more padding on hover than the other buttons, which looks a bit strange. This is pending UX/UI approval from Mozilla, but to me it looks very good!

web/viewer.css
```diff
@@ -1196,12 +1312,14 @@ html[dir='rtl'] .outlineItem > .outlineItems {
   text-decoration: none;
   display: inline-block;
   min-width: 95%;
-  min-width: calc(100% - 4px); /* Subtract the right padding (left, in RTL mode)
+  min-width: calc(
+    100% - 4px
+  ); /* Subtract the right padding (left, in RTL mode)
```
Snuffleupagus, line 1318: The comment here is no longer aligned properly; and why even reformat this in the first place?

---

Given the overall size of this patch, please avoiding making unrelated changes.

web/viewer.css
```diff
@@ -1134,33 +1258,25 @@ html[dir='rtl'] .thumbnail {
...
   background-clip: padding-box;
-  box-shadow: 0 1px 0 rgba(255, 255, 255, 0.05) inset,
-              0 0 1px rgba(255, 255, 255, 0.2) inset,
-              0 0 1px rgba(0, 0, 0, 0.2);
   color: rgba(255, 255, 255, 0.9);
 }
 
 .thumbnail.selected > .thumbnailSelectionRing > .thumbnailImage {
-  box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.5);
+  /* box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.5); */
```
Snuffleupagus, line 1273: Assuming that this no longer necessary, then it should be removed rather than be commented out.

web/viewer.css
```diff
@@ -1238,29 +1356,31 @@ html[dir='rtl'] .attachmentsItem > button {
   color: rgba(255, 255, 255, 0.5);
 }
 .outlineItemToggler::before {
-  content: url(images/treeitem-expanded.png);
+  mask: url(images/treeitem-expanded.svg) no-repeat;
+  -webkit-mask: url(images/treeitem-expanded.svg) no-repeat;
```
Snuffleupagus, line 1361: Given that we (unfortunately) still need to support IE 11 in the default viewer, using `mask` is sadly not going to work (applies throughout the patch); note https://developer.mozilla.org/en-US/docs/Web/CSS/mask#Browser_compatibility

utopianknight:
> the bookmark button has much more padding on hover than the other buttons

Fixed.

web/viewer.css
```diff
@@ -15,11 +15,79 @@
...
+
+/* ------------ Initial ------------ */
+
 :root {
   --sidebar-width: 200px;
-  --sidebar-transition-duration: 200ms;
-  --sidebar-transition-timing-function: ease;
-}
+
+  --main-color: #0c0c0d;
```
Snuffleupagus, line 25: Can you please use a *consistent* colour format throughout the patch, rather than mixing formats? 

There's at least four different formats used below, e.g. `#0c0c0d;`, `#5C5C61;`, `rgb(255, 74, 74);`, and `rgba(245, 246, 247, 1);`; please note PR #11447.

web/viewer.css
```diff
@@ -288,18 +348,22 @@ html[dir='rtl'] #toolbarContainer, .findbar, .secondaryToolbar {
...
 
 @keyframes progressIndeterminate {
-  0% { left: -142px; }
-  100% { left: 0; }
+  0% {
+    left: -142px;
+  }
+  100% {
+    left: 0;
+  }
```
Snuffleupagus, line 362: These changes also appear unrelated to the patch, please revert.

web/viewer.css
```diff
@@ -218,21 +285,15 @@ html[dir='rtl'] #outerContainer.sidebarOpen #viewerContainer:not(.pdfPresentatio
 #toolbarSidebar {
   width: 100%;
   height: 32px;
-  background-color: rgba(66, 66, 66, 1); /* fallback */
-  background-image: url(images/texture.png),
-                    linear-gradient(rgba(77, 77, 77, 0.99), rgba(64, 64, 64, 0.95));
+  background-color: var(--sidebar-bg-color); /* fallback */
```
Snuffleupagus, line 288: Leftover comment?

web/viewer.css
```diff
@@ -218,21 +285,15 @@ html[dir='rtl'] #outerContainer.sidebarOpen #viewerContainer:not(.pdfPresentatio
...
-              0 0 1px rgba(0, 0, 0, 0.1);
+html[dir="ltr"] #toolbarSidebar {
+  box-shadow: inset -1px 0 0 rgba(0, 0, 0, 0.25), 0 1px 0 rgba(0, 0, 0, 0.15),
+    0 0 1px rgba(0, 0, 0, 0.1);
 }
-html[dir='rtl'] #toolbarSidebar {
-  box-shadow: inset 1px 0 0 rgba(0, 0, 0, 0.25),
-              inset 0 1px 0 rgba(255, 255, 255, 0.05),
-              0 1px 0 rgba(0, 0, 0, 0.15),
-              0 0 1px rgba(0, 0, 0, 0.1);
```
Snuffleupagus, line 235: This *appears* to be more unrelated re-formatting; please revert if there's no functional changes here.

web/viewer.css
```diff
@@ -179,26 +247,24 @@ html[dir='rtl'] #sidebarContent {
...
 
 #outerContainer.sidebarResizing #viewerContainer {
   /* Improve responsiveness and avoid visual glitches when the sidebar is resized. */
   transition-duration: 0s;
 }
 
-html[dir='ltr'] #outerContainer.sidebarOpen #viewerContainer:not(.pdfPresentationMode) {
+html[dir="ltr"]
+  #outerContainer.sidebarOpen
+  #viewerContainer:not(.pdfPresentationMode) {
```
Snuffleupagus, line 261: These changes look unrelated to the patch, and the added indentation makes it more difficult to read the rules below; please revert this.

web/viewer.css
```diff
@@ -179,26 +247,24 @@ html[dir='rtl'] #sidebarContent {
...
   transition-property: left;
   left: var(--sidebar-width);
 }
-html[dir='rtl'] #outerContainer.sidebarOpen #viewerContainer:not(.pdfPresentationMode) {
+html[dir="rtl"]
+  #outerContainer.sidebarOpen
+  #viewerContainer:not(.pdfPresentationMode) {
```
Snuffleupagus, line 267: Same as the above.

web/viewer.css
```diff
@@ -309,20 +373,27 @@ html[dir='rtl'] #toolbarContainer, .findbar, .secondaryToolbar {
...
-                                        rgba(221, 221, 221, 1) 95px, rgba(187, 187, 187, 1) 100px);
+  background: repeating-linear-gradient(
+    135deg,
+    rgba(187, 187, 187, 1) 0,
+    rgba(153, 153, 153, 1) 5px,
+    rgba(153, 153, 153, 1) 45px,
+    rgba(221, 221, 221, 1) 55px,
+    rgba(221, 221, 221, 1) 95px,
+    rgba(187, 187, 187, 1) 100px
+  );
```
Snuffleupagus, line 384: Looks like unnecessary reformatting.

Snuffleupagus (changes requested):

fgilio:
This looks really good 😍
Thanks @utopianknight!

web/images/loading-dark.svg
```diff
@@ -0,0 +1,2 @@
...
...fill-opacity=".93"/><path d="M3.1 6.37l-1.73-1S.5 4.87 1 4c.5-.87 1.37-.37 1.37-.37l1.73 1s.87.5.37 1.37c-.5.87-1.37.37-1.37.37z" fill-opacity=".86"/><path d="M3 9H1S0 9 0 8s1-1 1-1h2s1 0 1 1-1 1-1 1z" fill-opacity=".79"/><path d="M4.1 11.37l-1.73 1S1.5 12.87 1 12c-.5-.87.37-1.37.37-1.37l1.73-1s.87-.5 1.37.37c.5.87-.37 1.37-.37 1.37z" fill-opacity=".72"/><path d="M3.63 13.56l1-1.73s.5-.87 1.37-.37c.87.5.37 1.37.37 1.37l-1 1.73s-.5.87-1.37.37c-.87-.5-.37-1.37-.37-1.37z" fill-opacity=".65"/><path d="M7 15v-2s0-1 1-1 1 1 1 1v2s0 1-1 1-1-1-1-1z" fill-opacity=".58"/><path d="M10.63 14.56l-1-1.73s-.5-.87.37-1.37c.87-.5 1.37.37 1.37.37l1 1.73s.5.87-.37 1.37c-.87.5-1.37-.37-1.37-.37z" fill-opacity=".51"/><path d="M13.56 12.37l-1.73-1s-.87-.5-.37-1.37c.5-.87 1.37-.37 1.37-.37l1.73 1s.87.5.37 1.37c-.5.87-1.37.37-1.37.37z" fill-opacity=".44"/><path d="M15 9h-2s-1 0-1-1 1-1 1-1h2s1 0 1 1-1 1-1 1z" fill-opacity=".37"/><path d="M14.56 5.37l-1.73 1s-.87.5-1.37-.37c-.5-.87.37-1.37.37-1.37l1.73-1s.87-.5 1.37.37c.5.87-.37 1.37-.37 1.37z" fill-opacity=".30"/><path d="M9.64 3.1l.98-1.66s.5-.874 1.37-.37c.87.5.37 1.37.37 1.37l-1 1.73s-.5.87-1.37.37c-.87-.5-.37-1.37-.37-1.37z" fill-opacity=".23"/></svg>
```
Snuffleupagus, line 2: Please note that the old `loading-small.png` files where *animated* PNG images, whereas this new SVG (and `loading.svg` as well) appear to be just stationary images.

web/viewer.css
```diff
@@ -19,6 +19,140 @@
...
+  --secondaryToolbarButton-spreadNone-icon: url(images/secondaryToolbarButton-spreadNone.svg);
+  --secondaryToolbarButton-spreadOdd-icon: url(images/secondaryToolbarButton-spreadOdd.svg);
+  --secondaryToolbarButton-spreadEven-icon: url(images/secondaryToolbarButton-spreadEven.svg);
+  --secondaryToolbarButton-documentProperties-icon: url(images/secondaryToolbarButton-documentProperties.svg);
+}
+
+@media (prefers-color-scheme: dark) {
+  :root {
+    --toolbar-icon-opacity: 0.7;
+    --doorhanger-icon-opacity: 0.9;
```
Snuffleupagus, line 91: Nit: These values look identical to the ones above, is it necessary to redeclare them here?

brendandahl:
@Snuffleupagus is this good to go?

Snuffleupagus:
Unfortunately IE 11 is still causing trouble here, since the toolbar icons are not visible, please see

ie

utopianknight:
It's seems to be working fine on my version of internet explorer.

Internetexplorer11

Snuffleupagus:
The screen-shot in https://github.com/mozilla/pdf.js/pull/11077#issuecomment-673010632 was taken on Windows 7.

Furthermore the build logs contains the following output, which also needs fixing:
```
### Creating generic viewer
[17:51:14] gulp-postcss: viewer.css
postcss-css-variables: /home/ubuntu/pdfjs/botio-files-pdfjs/private/89bc36956c320d7/viewer.css:546:3: variable --scrollbar-color is undefined and used without a fallback
postcss-css-variables: /home/ubuntu/pdfjs/botio-files-pdfjs/private/89bc36956c320d7/viewer.css:546:3: variable --scrollbar-bg-color is undefined and used without a fallback
postcss-css-variables: /home/ubuntu/pdfjs/botio-files-pdfjs/private/89bc36956c320d7/viewer.css:546:3: variable --scrollbar-color is undefined and used without a fallback
postcss-css-variables: /home/ubuntu/pdfjs/botio-files-pdfjs/private/89bc36956c320d7/viewer.css:546:3: variable --scrollbar-bg-color is undefined and used without a fallback
postcss-css-variables: /home/ubuntu/pdfjs/botio-files-pdfjs/private/89bc36956c320d7/viewer.css:546:3: variable --scrollbar-color is undefined and used without a fallback
postcss-css-variables: /home/ubuntu/pdfjs/botio-files-pdfjs/private/89bc36956c320d7/viewer.css:546:3: variable --scrollbar-bg-color is undefined and used without a fallback
postcss-css-variables: /home/ubuntu/pdfjs/botio-files-pdfjs/private/89bc36956c320d7/viewer.css:546:3: variable --scrollbar-color is undefined and used without a fallback
postcss-css-variables: /home/ubuntu/pdfjs/botio-files-pdfjs/private/89bc36956c320d7/viewer.css:546:3: variable --scrollbar-bg-color is undefined and used without a fallback
```

brendandahl:
I tried it out here on win10 IE11 and looks fine to me too.

Snuffleupagus (commented):
Please note the build warnings mentioned in https://github.com/mozilla/pdf.js/pull/11077#issuecomment-674192940

Furthermore, there appears to be a merge-commit present in the PR, please make sure that you remove it; https://github.com/mozilla/pdf.js/wiki/Squashing-Commits may be helpful.

Snuffleupagus:
From a support perspective, it's going to be *extremely* problematic if the viewer is broken in *some* IE 11 versions (maybe even related to the OS version).
Hence I think that there's basically two possible options here, either:
 - Attempt to fix this the `toolbar` icons, such that all the viewer works in all IE 11 versions.
 - Create a *new* PDF.js release first, and officially announce that it's the last one with support for IE 11 (after which this PR can land).

@brendandahl How should we proceed here?

utopianknight:
I believe the following issue was fixed:

>       * on mac the dropdown seems to have a glitch on top and bottom
>         CleanShot 2020-07-15 at 09 23 26@2x

Testers needed.

brendandahl:
Ideally, I'd like to not drop support just because of a UI refresh.
@Snuffleupagus what version of IE11 is on Win7? 

image

utopianknight:
@Snuffleupagus Can you please use your developer tools and tell me what this is?

90047405-262dcc80-dcd2-11ea-81d4-952cdb211167

The only way for me to fix this is to use the same IE 11 version that you have. So I'd like to know this as well.

utopianknight:
@brendandahl According to [this](https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge) and [this](https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet), Microsoft dropped support for Windows 7's Internet Explorer a while ago. Why are we still supporting it?

Snuffleupagus:
> Ideally, I'd like to not drop support just because of a UI refresh.

Honestly though, it wouldn't *only* be because of the UI changes. The FAQ has listed IE 11 support as *limited* for well over a year now, see https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support and issue #11211, and generally speaking supporting IE 11 is causing (and has historically caused) a fair amount of grief/problems/limitations over the years.

> @Snuffleupagus what version of IE11 is on Win7?

I cannot be sure that it's the *latest* available one, since I'm having trouble updating IE 11, however my currently installed version is:
ie11-version

---

> Can you please use your developer tools and tell me what this is?

Nothing special, as far as I can tell, it appears to simply be *part* of an SVG image.

Interestingly though it appears that all of the toolbar icons do load, however they're not scaled or positioned correctly; note what happens when I try disabling a randomly chosen `position` rule:
position

Snuffleupagus:
Testing in Firefox, I noticed to additional things:
 - It appears that the updated `web/images/shadow.png` file becomes effectively *invisible* in `@media (prefers-color-scheme: dark)` mode; is it perhaps necessary to have two different versions of this image resource as well?
 - Also, if you're removing the `web/images/texture.png` file then please remove this line as well https://github.com/mozilla/pdf.js/blob/192afb87e34e1f7c30ad97e98599af643717eb9d/gulpfile.js#L772

utopianknight:
Please note that prefers-color-scheme does not seem to work on the Developer Edition of Firefox without an extension. Until this is fixed, should the light theme be the default or the dark theme?

>It appears that the updated web/images/shadow.png file becomes effectively invisible

It is still visible to me. If you quickly toggle its visibility, you'll notice that something looks off when it's absent. The reason why I don't want to make it too dark is so that PDF files with dark pages don't appear too fuzzy on the edges. If you're asking me to use CSS shadow instead of border-image, please know that this could introduce performance issues.

Snuffleupagus:
> It is still visible to me.

Really, even with the dark-mode enabled?
It obviously seems fine in the (default) light-mode; even looking at your screen-shots in https://github.com/mozilla/pdf.js/pull/11077#issue-308369353, `web/images/shadow.png` is not really discernible in the dark-mode (at least to my eye).

> If you're asking me to use CSS shadow instead of border-image, please know that this could introduce performance issues.

Of course not, I was around when the image was added (and the old CSS-only solution removed) for performance reasons :-)

utopianknight:
This is an optical illusion. The shadow will always look invisible when the page and background are contrasting. Here's how it looks like with dark page:

Annotation 2020-08-17 163206

The only to fix it is to get the page's color and generate a shadow for each one of them. It's possible. Here's an example: https://github.com/robinloeffel/cosha

timvandermeij:
For IE11, another scenario would be to do a major version bump for the next release and drop support completely. Not only do I agree with https://github.com/mozilla/pdf.js/pull/11077#issuecomment-674736476 but it would also simplify landing this patch so we don't have to waste time on IE11 anymore.

I think a major version bump would also be good given that many big features landed that received the most upvotes compared to other issues: AcroForms support, layers support, zoom fixes for OSX and TypeScript definitions. That's quite a bit and in my opinion deserves a major version bump.

What do you think @brendandahl?

utopianknight:
@timvandermeij It's fixed. Please call /botio-linux

brendandahl:
> For IE11, another scenario would be to do a major version bump for the next release (3.x) and drop support completely. Not only do I agree with [#11077 (comment)](https://github.com/mozilla/pdf.js/pull/11077#issuecomment-674736476) but it would also simplify landing this patch so we don't have to waste time on IE11 anymore. If someone _really_ needs IE11 support, the 2.x versions can still be used.
> 
> I think a major version bump would also be good given that many big features landed that received the most upvotes compared to other issues: AcroForms support, layers support, zoom fixes for OSX and TypeScript definitions. That's quite a bit and in my opinion deserves a major version bump.
> 
> What do you think @brendandahl?

I'm fine with the approach of dropping IE11 support. I'd like to stick to semantic versioning though, so I don't think we should bump the major version number since we haven't broken/deprecated any core API's.

@timvandermeij if we're fine with breaking IE11, is this good to go? I'd like to get it in for Firefox 81.

timvandermeij:
If @utopianknight can fix this with the version information provided by @Snuffleupagus I'd prefer that, but looking at https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666 I'm fine with dropping IE11 support for the next version without a major version bump, as long as we make this clear in the commit message and during the release.

I haven't done an in-depth review of this patch, so I'll leave it to you and/or @Snuffleupagus to merge this.

/botio-linux preview

web/viewer.css
```diff
@@ -539,164 +740,121 @@ html[dir='rtl'] .splitToolbarButton > .toolbarButton {
...
-.overlayButton:hover:active,
-.dropdownToolbarButton:hover:active,
-.secondaryToolbarButton:hover:active {
-  background-color: rgba(0, 0, 0, 0.2);
-  background-image: linear-gradient(rgba(255, 255, 255, 0.05), rgba(255, 255, 255, 0));
-  border-color: rgba(0, 0, 0, 0.35) rgba(0, 0, 0, 0.4) rgba(0, 0, 0, 0.45);
-  box-shadow: 0 1px 1px rgba(0, 0, 0, 0.1) inset,
-              0 0 1px rgba(0, 0, 0, 0.2) inset,
-              0 1px 0 rgba(255, 255, 255, 0.05);
+  background-color: var(--doorhanger-hover-color);
```
Snuffleupagus, line 830: Did you actually mean `--button-hover-color` here!?

Given that e.g. `.splitToolbarButton > .toolbarButton:hover` uses that css variable, the effect of *this* line is that different toolbar buttons end up looking slightly different when hovered.
Is this really intended, since it feels a bit inconsistent?

web/viewer.css
```diff
@@ -539,164 +740,121 @@ html[dir='rtl'] .splitToolbarButton > .toolbarButton {
...
+  margin-top: 2px !important;
 }
 .dropdownToolbarButton::after {
   position: absolute;
   display: inline-block;
-  top: 4px;
-  content: url(images/toolbarButton-menuArrows.png);
+  top: 6px;
+  content: var(--toolbarButton-menuArrow-icon);
+  pointer-events: none;
```
Snuffleupagus, line 857: I think that you'll need to add `max-width: 16px;` here as well, since otherwise the icon isn't visible on all IE 11 versions.

Snuffleupagus (changes requested):
The latest preview now *almost* works in IE 11, on Windows 7, note the screen-shot below; I've added an inline suggestion that should address the last missing icon.

Furthermore (and unrelated to IE), I've also noticed a slight inconsistency regarding the `.toolbarButton` hover effects, see the inline comment.

After the above have been fixed, I *believe* that this should now be good to land as-is even with regards to IE 11 on Windows 7 :-)

---
image

utopianknight:
Alright. I'm done. I believe patch is ready to merge. Green light from me.

Snuffleupagus:
The following is a, possibly incomplete, list of issues/bugs that's fixed by this PR:
 - #9702
 - #11572
 - #3528 (to the extent that we're going to "fix" it, since this patch should also *significantly* simplify UI customization)
 - [bug 1410598](https://bugzilla.mozilla.org/show_bug.cgi?id=1410598)
 - [bug 738947](https://bugzilla.mozilla.org/show_bug.cgi?id=738947) (that's essentially a WONTFIX at this point, since we're now aligning with the Photon design)

timvandermeij (approved):

timvandermeij:
Looks good to me now. Thank you for all your work!

szaimen:
🎉

utopianknight:
Awesome! Great job everyone.

natewind:
Wow, what an exceptionally ugly look! It’s like a textbook example of an unnecessary redesign, considering how perfect the old look was. Disgusting. @sawanm9000, you are the problem with modern design.

PR #11079, opened 2019-08-19
[TextLayer] Only cache the current `textDiv` style when `enhanceTextSelection` is enabled and use template strings in `expandTextDivs`` 

Snuffleupagus:
- [TextLayer] Only cache the current `textDiv` style when `enhanceTextSelection` is enabled

   This will help save a little bit of memory, by not storing one unused string for each `textDiv` in regular text-selection mode.

- [TextLayer] Reduce the amount of intermediary strings in `expandTextDivs`

  By using template strings, we can avoid some unnecessary string allocations (which is also helped by shortening a variable name).

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -628,49 +628,48 @@ var renderTextLayer = (function renderTextLayerClosure() {
       }
 
       for (var i = 0, ii = this._textDivs.length; i < ii; i++) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks like a good optimization. Thanks!

PR #11086, opened 2019-08-22
[TextLayer] Only cache the `originalTransform` when `enhanceTextSelection` is enabled

Snuffleupagus:
Given that this is completely unused in "regular" text-selection mode, there's no reason to unconditionally store one string for every `textDiv`.

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -537,14 +537,12 @@ var renderTextLayer = (function renderTextLayerClosure() {
       if (textDivProperties.isWhitespace) {
         return;
       }
-
-      let fontSize = textDiv.style.fontSize;
-      let fontFamily = textDiv.style.fontFamily;
+      const { fontSize, fontFamily, } = textDiv.style;
 
       // Only build font string and set to context if different from last.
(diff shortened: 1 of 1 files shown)


src/display/text_layer.js
```diff
@@ -668,8 +668,8 @@ var renderTextLayer = (function renderTextLayerClosure() {
             div.style.transform = transform;
           }
         } else {
-          div.style.padding = 0;
-          div.style.transform = divProps.originalTransform || '';
+          div.style.padding = null;
+          div.style.transform = divProps.originalTransform;
```
Snuffleupagus: According to https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style#Setting_styles:
> A style declaration is reset by setting it to `null` or an empty string, e.g., `elt.style.color = null`.

Hence the empty string isn't necessary here, since `originalTransform` defaults to `null`. Furthermore by not setting the `padding` to zero we can avoid `style =" ... padding: 0px; ..."` everywhere after expanding is reverted.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #11087, opened 2019-08-22
Add a way to disable external links.

brendandahl:


brendandahl:
@Snuffleupagus has reviewed already

PR #11089, opened 2019-08-23
[TextLayer] Use Arrays to build the total `transform` and `padding`, rather than concatenating Strings, in `expandTextDivs`

Snuffleupagus:

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -626,6 +626,7 @@ var renderTextLayer = (function renderTextLayerClosure() {
         expand(this);
         this._bounds = null;
       }
+      const transformBuf = [];
 
       for (var i = 0, ii = this._textDivs.length; i < ii; i++) {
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Interestingly the only `text` failure was in Linux and Chrome, with no other configuration affected. The likely culprit should be the first patch, and I suppose that we could change things to *only* skip negative padding but still set the transform property in that case!?

Snuffleupagus:
I'll try to submit these patches separately, to see if it's possible to pin-point where the movement in Chrome on Linux is coming from.

Snuffleupagus:
> I'll try to submit these patches separately, to see if it's possible to pin-point where the movement in Chrome on Linux is coming from.

This was split into PRs #11090, #11091, and #11092; please see https://github.com/mozilla/pdf.js/pull/11092#discussion_r317332374 which outlines the culprit.

PR #11090, opened 2019-08-23
[TextLayer] Use an Array to build the total `transform`, rather than concatenating Strings, in `expandTextDivs`

Snuffleupagus:
Furthermore, it's possible to re-use the same Array for all `textDiv`s on the page.

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -626,6 +626,7 @@ var renderTextLayer = (function renderTextLayerClosure() {
         expand(this);
         this._bounds = null;
       }
+      const transformBuf = [];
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice patches on the performance of the text layer recently! I assume we can do the same for the `padding` strings here?

Snuffleupagus:
> Nice patches on the performance of the text layer recently!

Thank you!

> I assume we can do the same for the `padding` strings here?

That's indeed the idea here, however I ran into some Linux/Chrome only test failures over in PR #11089 which is why I'm trying to split this into multiple PRs to hopefully get to the bottom of it.

PR #11091, opened 2019-08-23
[TextLayer] Only handle positive padding values in `expandTextDivs`

Snuffleupagus:
Given that browsers will reject padding values smaller than zero (which may be caused by limited numerical precision during calculations in the `expand` code), it makes no sense to include those when expanding the `textDiv`s.


Snuffleupagus:
Check if also ignoring the `transform` when padding is negative works...

/botio test

timvandermeij (approved):

timvandermeij:
Nice improvement!

PR #11092, opened 2019-08-23
[TextLayer] Use an Array to build the total `padding`, rather than concatenating Strings, in `expandTextDivs`

Snuffleupagus:
Furthermore, it's possible to re-use the same Array for all `textDiv`s on the page and the resulting padding string also becomes a lot more compact.
Please note that the `paddingLeft` branch was moved, since the padding values need to be ordered as `top, right, bottom, left`.

Finally, with this re-factoring it's no longer necessary to cache the original `style` string for every `textDiv` when `enhanceTextSelection` is enabled.

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -57,7 +57,6 @@ var renderTextLayer = (function renderTextLayerClosure() {
     // Initialize all used properties to keep the caches monomorphic.
     var textDiv = document.createElement('span');
     var textDivProperties = {
-      style: null,
       angle: 0,
       canvasWidth: 0,
       isWhitespace: false,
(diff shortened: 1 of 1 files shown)


src/display/text_layer.js
```diff
@@ -637,34 +632,41 @@ var renderTextLayer = (function renderTextLayerClosure() {
...
+              `translateX(${-divProps.paddingLeft / divProps.scale}px)`);
+          } else {
+            paddingBuf.push(0);
           }
 
-          if (padding !== '') {
-            div.setAttribute('style', divProps.style + padding);
+          const padding = paddingBuf.join(' ');
+          if (padding !== NO_PADDING) {
+            div.style.padding = padding;
```
Snuffleupagus: Having tested this incrementally on the bot, this line is actually the cause for the Linux/Chrome-only test failure.

It seems that setting `style.padding` directly will truncate some of the last decimals of the padding values, whereas `setAttribute` will keep all the decimals intact (based on checking with the Devtools Inspector).
The difference between keeping all decimals should be *really* small, or even imperceivable, which would explain why most platforms aren't affected by this change. However, it's certainly easy to imagine that the difference (even though it's very small) could trigger ever so slightly different rounding behaviour on some platforms.

All in all, I thus don't think we need to worry about the tiny movement in one of the tests :-)

timvandermeij (approved):

timvandermeij:
Nice work!

/botio makeref

PR #11093, opened 2019-08-24
Shorthand afterPrint signature in app.js

Priestch:


.gitignore
```diff
@@ -7,3 +7,4 @@ Makefile
 node_modules/
 examples/node/svgdump/
 examples/node/pdf2png/*.png
+.idea/
```
Snuffleupagus: Please remove this entire commit from the PR, since there's no such folder in the PDF.js repository.

Priestch: @Snuffleupagus This is the auto-generated IDE directory. If not ignored here, I have to check every time before add changes to commit. Is this not the right way to handle such situation?

Snuffleupagus: See e.g. https://github.com/mozilla/pdf.js/pull/8688#discussion_r128925079, since it doesn't seem reasonable to add ignores for potentially any number of different IDEs in this repository.

Priestch: @Snuffleupagus It makes sense. So should i just revert that commit or completely remove that commit?

timvandermeij (approved):

timvandermeij:
Thank you!

PR #11095, opened 2019-08-24
Include a reduced, non-linked PDF file for the attachments API unit test

### LINKED ISSUE #11063: Use a non-linked PDF file for the 'gets attachments' unit-test
To avoid skipping this unit-test in Node.js, since Travis cannot access the linked test-cases, we should add a *reduced* PDF file (containing attachments) directly to the repository and use that instead: https://github.com/mozilla/pdf.js/blob/fbe8c6127c6299d72faada32dd89f7bf5933497e/test/unit/api_spec.js#L694-L710

timvandermeij:
The file is created using Poppler's `pdfattach` utility.

Fixes #11063.


test/unit/api_spec.js
```diff
@@ -692,18 +692,15 @@ describe('api', function() {
...
         return pdfDoc.getAttachments();
       });
       promise.then(function (data) {
-        var attachment = data['Press Quality.joboptions'];
-        expect(attachment.filename).toEqual('Press Quality.joboptions');
+        var attachment = data['foo.txt'];
+        expect(attachment.filename).toEqual('foo.txt');
         expect(attachment.content instanceof Uint8Array).toBeTruthy();
-        expect(attachment.content.length).toEqual(30098);
+        expect(attachment.content.length).toEqual(9);
```
Snuffleupagus: Now that you've added a nice reduced test-case, can you please replace the two lines above with an *explicit* check of the file contents instead? I.e. just `expect(attachment.content).toEqual(new Uint8Array([98, 97, 114, 32, 98, 97, 122, 32, 10]);` should suffice.

(The only reason that didn't do that when adding the test was because of the large attachment size, which is no longer a problem with the new PDF file :-)

Snuffleupagus (approved):
This looks great already, but I've added one small comment; thanks for fixing this!

test/unit/api_spec.js
timvandermeij: Good point! Done in the new commit.

PR #11097, opened 2019-08-25
[TextLayer] Only measure the width of the text, in `_layoutText`, for multi-char text divs

Snuffleupagus:
For performance reasons single-char text divs aren't being scaled, as outlined in a comment in `appendText`. Hence it doesn't seem necessary, or even a good idea, to unconditionally measuring the width of the text in `_layoutText`.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #11104, opened 2019-08-28
[TextLayer] Avoid unnecessary font updates in `_layoutText` and remove `setAttribute` usage in `appendText`

### LINKED ISSUE #8066: text layer style micro-optimization resulted in requiring style-src 'unsafe-inline' in Content Security Policy
https://github.com/mozilla/pdf.js/commit/cb5f9df0c8932fe0db9f783ede7893b7efcadcdb started generating style strings automatically, which is forbidden by a standard Content Security Policy. It would be nice to have a fallback or perhaps reconsider the approach.

To reproduce, use `style-src 'self'` in CSP:

        <meta http-equiv="Content-Security-Policy" content="style-src 'self'" />

Snuffleupagus:
*This should conclude my recent series of patches for the `TextLayer`; smaller diff with https://github.com/mozilla/pdf.js/pull/11104/files?w=1*

 - [TextLayer] Avoid unnecessary font updates in `_layoutText` (PR 11097 follow-up)

   *This should obviously have been done in PR 11097, but for some reason I completely overlooked it; sorry about that.*

   There's no good reason to update the font unless you're actually going to measure the width of the textContent. This can reduce unnecessary font switching a fair bit, even for documents which are somewhat simple/short (in e.g. the `tracemonkey.pdf` file this cuts the amount of font switches almost in half).

 - [TextLayer] Remove `setAttribute` usage in `appendText` (issue 8066)

   One of the motivations for using `setAttribute` in the first place was to support more efficient DOM updates in the `expandTextDivs` method, since performance of the `enhanceTextSelection` mode can be somewhat bad when there's a lot of `textDivs` on the page.

   With recent `TextLayer` changes/optimizations it's no longer necessary to store a complete `style`-string for every `textDiv`, and we can thus re-visit the `setAttribute` usage.
Note that with the current code, in `appendText`, there's only *one* string per `textDiv` which avoids a bunch of temporary strings. While the changes in this patch means that there's now *three* strings per `textDiv` instead, the total length of these strings are now quite a bit shorter (42 characters to be exact).

Fixes #8066

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -48,11 +48,6 @@ var renderTextLayer = (function renderTextLayerClosure() {
     return !NonWhitespaceRegexp.test(str);
   }
 
-  // Text layers may contain many thousands of divs, and using `styleBuf` avoids
-  // creating many intermediate strings when building their 'style' properties.
-  var styleBuf = ['left: ', 0, 'px; top: ', 0, 'px; font-size: ', 0,
-                  'px; font-family: ', '', ';'];
-
   function appendText(task, geom, styles) {
     // Initialize all used properties to keep the caches monomorphic.
     var textDiv = document.createElement('span');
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
The test "failures" should have the same explanation as in https://github.com/mozilla/pdf.js/pull/11092#discussion_r317332374.

timvandermeij (approved):

timvandermeij:
Looks good; thank you for all your efforts regarding the text layer!

/botio makeref

Issue #1887, opened 2012-07-10
Prioritize which operator list is built on the worker.


bf:
Hello, 

I have experienced huge lags in PDF rendering when "skimming" (scrolling fast) over a 300+ page PDF file. According to my console, calls to `getOperatorList` take more than 1000ms each which I assume is the core of this performance issue.

It seems that for each page you scroll over, `getOperatorList` is called. The problem is, that this call finishes when the user has scrolled further for about 50 pages, so it is no longer relevant as it returns a result for a page that is not visible to the user any more. The end result is, that when I have reached page 350, I have to way for 10+ seconds until the currently visible document is rendered.

In my eyes, this performance issue can probably be saved by using a LIFO queue for rendering pages, so the last page is rendered as soon as possible, and then all the pages I have scrolled over.

Enclosed you find parts of the console output. Thank you!

```
/pdf.js/build/pdf.js:30363      page=   426 - getOperatorList: time=1189ms, len=1995
l10n.js:194                         [l10n] #thumb_page_canvas missing for [de]
/pdf.js/build/pdf.js:30363      page=   427 - getOperatorList: time=1022ms, len=1553
l10n.js:194                         [l10n] #thumb_page_canvas missing for [de]
/pdf.js/build/pdf.js:30363      page=   428 - getOperatorList: time=1294ms, len=2755
l10n.js:194                         [l10n] #thumb_page_canvas missing for [de]
/pdf.js/build/pdf.js:30363      page=   448 - getOperatorList: time=1324ms, len=87
/pdf.js/build/pdf.js:30363      page=   474 - getOperatorList: time=1027ms, len=1305
/pdf.js/build/pdf.js:30363      page=   484 - getOperatorList: time=1366ms, len=2564
/pdf.js/build/pdf.js:30363      page=   496 - getOperatorList: time=1089ms, len=1110
/pdf.js/build/pdf.js:30363      page=   498 - getOperatorList: time=1171ms, len=1010
/pdf.js/build/pdf.js:30363      page=   499 - getOperatorList: time=1064ms, len=612
/pdf.js/build/pdf.js:30363      page=   517 - getOperatorList: time=1412ms, len=2277
/pdf.js/build/pdf.js:30363      page=   520 - getOperatorList: time=1298ms, len=1958
/pdf.js/build/pdf.js:30363      page=   521 - getOperatorList: time=1264ms, len=2019
[...]

PR #11106, opened 2019-08-29
Abort, with a small delay, `getOperatorList` on the worker-thread when rendering is cancelled (PR 11069 follow-up)

Snuffleupagus:
With this patch we're finally able to abort worker-thread parsing of the `OperatorList`, rather than *only* aborting the main-thread rendering itself, when the `RenderTask.cancel` method is being called.
This will help improve perceived performance in the default viewer, especially when reading longer and more complex documents, since pages that've been scrolled out-of-view (and thus evicted from the cache) will no longer compete for parsing resources on the worker-thread.

*Please note:* With the implementation in this patch we're *not* aborting worker-thread parsing immediately on `RenderTask.cancel`, since that would lead to *worse* performance in many cases. For example: When zoom/rotation occurs in the viewer, while parsing/rendering is still ongoing, a `cancel` call will usually be (almost) immediately folled by a new `PDFPageProxy.render` call. In that case you obviously don't want to abort parsing on the worker-thread, since that would risk throwing away a partially parsed `OperatorList` and thus force unnecessary re-parsing which will regress perceived performance (especially for more complex documents).

When choosing a reasonable delay, before cancelling `getOperatorList` on the worker-thread when `RenderTask.cancel` is called, two different positions need to be considered:
 1. The delay needs to be short enough, since a timeout in the multiple seconds range would essentially make this entire functionality meaningless (by always allowing most/all pages enough time to finish parsing).

 2. The delay cannot be *too* short, since that would actually *reduce* performance in the zoom/rotation case outlined above. Furthermore, the time between `RenderTask.cancel` and `PDFPageProxy.render` calls will obviously be affected by both general computer performance and current CPU load.

It's certainly possible that the timeout may require some further tweaks, however the value settled on in this patch was easily *one order* of magnitude larger than the delta between cancel/render in my tests.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -38,6 +38,7 @@ import { PDFDataTransportStream } from './transport_stream';
 import { WebGLContext } from './webgl';
 
 const DEFAULT_RANGE_CHUNK_SIZE = 65536; // 2^16 = 65536
+const RENDERING_CANCELLED_TIMEOUT = 100; // ms
 
 let isWorkerDisabled = false;
 let fallbackWorkerSrc;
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
PR #11069 has now landed in `mozilla-central`, courtesy of [bug 1576613](https://bugzilla.mozilla.org/show_bug.cgi?id=1576613), but I'd like to delay landing of this PR a little while to make it easier to pin-point any issues (hopefully there won't be any :-).
Additionally, since there's been a couple of recent `MessageHandler` patches it probably won't hurt to wait for those to land in `mozilla-central` first as well.

Snuffleupagus:
This will also address (a significant) part of issue #1887, since we're now aborting parsing of both `OperatorList` and `TextContent` on the worker-thread when rendering is cancelled.

timvandermeij (approved):

timvandermeij:
Looks like a really good approach to me!

Snuffleupagus:
> Looks like a really good approach to me!

@timvandermeij Thanks a lot for reviewing and landing this; this should hopefully help performance a fair bit with long documents.

I tried to keep the implementation as simple I possibly could and the key, as far as I'm concerned, was to remove the entire `intentState` on abort rather than trying to resetting only some of its parameters.

PR #11108, opened 2019-08-31
Use more ES6 syntax in the annotation code

timvandermeij:
Given that there are some open issues regarding annotations, I think it's a good idea to, before we start working on them, make sure that the annotation code is cleaned up and completely uses ES6 syntax. This commit series does that.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -47,18 +47,18 @@ class AnnotationFactory {
    * @private
    */
   static _create(xref, ref, pdfManager, idFactory) {
-    let dict = xref.fetchIfRef(ref);
+    const dict = xref.fetchIfRef(ref);
     if (!isDict(dict)) {
       return undefined;
     }
-    let id = isRef(ref) ? ref.toString() : `annot_${idFactory.createObjId()}`;
+    const id = isRef(ref) ? ref.toString() : `annot_${idFactory.createObjId()}`;
 
     // Determine the annotation's subtype.
     let subtype = dict.get('Subtype');
     subtype = isName(subtype) ? subtype.name : null;
 
     // Return the right annotation object based on the subtype and field type.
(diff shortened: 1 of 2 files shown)


src/core/annotation.js
```diff
@@ -181,20 +181,20 @@ function getQuadPoints(dict, rect) {
...
-  let bounds = Util.getAxialAlignedBoundingBox(bbox, matrix);
-  let minX = bounds[0];
-  let minY = bounds[1];
-  let maxX = bounds[2];
-  let maxY = bounds[3];
+  const bounds = Util.getAxialAlignedBoundingBox(bbox, matrix);
+  const minX = bounds[0];
+  const minY = bounds[1];
+  const maxX = bounds[2];
+  const maxY = bounds[3];
```
Snuffleupagus: Nit: You could use Array destructuring here, i.e.
```javascript
const [minX, minY, maxX, maxY] =
  Util.getAxialAlignedBoundingBox(bbox, matrix);
```

src/core/annotation.js
```diff
@@ -958,27 +955,27 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
+    const normalAppearanceState = appearanceStates.get('N');
     if (!isDict(normalAppearanceState)) {
       return;
     }
-    let keys = normalAppearanceState.getKeys();
-    for (let i = 0, ii = keys.length; i < ii; i++) {
-      if (keys[i] !== 'Off') {
-        this.data.buttonValue = keys[i];
+    const keys = normalAppearanceState.getKeys();
+    for (const key of keys) {
```
Snuffleupagus: Nit: I think `for (const key of normalAppearanceState.getKeys()) {` ought to work as well.

Snuffleupagus (approved):

PR #11110, opened 2019-09-01
Remove support for the `scope` parameter in the `MessageHandler.on` method

Snuffleupagus:
At this point in time it's easy to convert the `MessageHandler.on` call-sites to use arrow functions, and thus let the JavaScript engine handle scopes for us, rather than having to manually keep references to the relevant scopes in `MessageHandler`.[1]
An additional benefit of this is that a couple of `Function.prototype.call()` instances can now be converted into "normal" function calls, which should be a tiny bit more efficient.

All in all, I don't see any compelling reason why it'd be necessary to keep supporting custom `scope`s in the `MessageHandler` implementation.

---
[1] In the event that a custom scope is ever needed, simply using `bind` on the handler function when calling `MessageHandler.on` ought to work as well.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1875,7 +1875,7 @@ class WorkerTransport {
   setupMessageHandler() {
     const { messageHandler, loadingTask, } = this;
 
-    messageHandler.on('GetReader', function(data, sink) {
+    messageHandler.on('GetReader', (data, sink) => {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Nice find! Thank you for simplifying this.

PR #11111, opened 2019-09-01
Inline the `resolveCall` helper function at its call-sites in `MessageHandler`

Snuffleupagus:
There's only three call-sites and one of them doesn't even need the complete functionality of `resolveCall`, hence it seems reasonable to just inline this code.
An additional benefit of this is that the `Function.prototype.apply()` instance can also be converted into "normal" function calls, which should be a tiny bit more efficient.

The patch also replaces a number of unnecessary arrow functions, in relevant parts of the `MessageHandler` code, with "normal" functions instead.
Finally, all `Promise.resolve().then(...)` calls are replaced with `new Promise(...)` instead since the latter is a tiny bit more efficient. This also explains the test failures on the Linux bot, with a prior version of the patch, since the `Promise.resolve().then(...)` format essentially creates two Promises thus causing additional delay.

### DIFF
--- a/src/shared/message_handler.js
+++ b/src/shared/message_handler.js
@@ -30,13 +30,6 @@ const StreamKind = {
   START_COMPLETE: 8,
 };
 
-async function resolveCall(fn, args) {
-  if (!fn) {
-    return undefined;
-  }
-  return fn.apply(null, args);
-}
-
 function wrapReason(reason) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Indeed, no need for that to be a helper function. Thanks!

PR #11112, opened 2019-09-01
Remove the API/Worker version warning message in `TESTING` mode

Snuffleupagus:
The warning messages turn out to be more annoying than helpful when looking at the `console` during tests, so let's just remove them.


timvandermeij (approved):

timvandermeij:
I agree, looks good!

PR #11113, opened 2019-09-01
fix(): Parsing problem: Query params provided in the file's URL are n…

lamisChebbi:
This problem occurs when your provide a file's URL with query params. The parsing stops at the first key and consequently the query params will not be sent to the server. This used to work in the previous release.


Snuffleupagus:
Please note that patches cannot be accepted if submitted against the `gh-pages` branch.

Furthermore, it's quite difficult to understand *exactly* what this patch attempts to fix without either an example or clear steps to reproduce the problem. Finally, please note that [as mentioned in the FAQ](https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#file) the file URL should be encoded with `encodeURIComponent()`.

timvandermeij:
I... have no idea what the bot did here, but it doesn't seem like it actually got pushed anywhere. In any case, I also didn't understand what this actually fixes, so please open an issue first with a clear problem description and steps to reproduce so we can discuss it.

lamisChebbi:
Hello @timvandermeij @Snuffleupagus 
I created the issue [#11377](https://github.com/mozilla/pdf.js/issues/11377) relative to this problem.
Thanks

Issue #10935, opened 2019-06-28
Implement blend modes for the SVG back-end


timvandermeij:
The initial implementation for this is in #10477, but due to many other code changes and renewed insights it was not completed. It can, however, be used as inspiration for a new pull request on top of the current SVG back-end code.

PR #11116, opened 2019-09-03
Implement BM for the SVG back-end

ahuglajbclajep:
I implemented blend modes referring to #10477.
Since beginGroup and endGroup are not implemented, it seems to be effective only for the background color, but for example, the display of the PDF file from #10487 is close to the original.

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -320,6 +320,9 @@ class SVGExtraState {
     this.clipGroup = null;
 
     this.maskId = '';
+
+    this.parentGroup = null;
+    this.currentGroup = null;
   }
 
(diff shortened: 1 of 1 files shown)


janpe2:
If I open any PDF, I get `TypeError: this.parentGroup is null`. Are you sure the SVG back-end was enabled when you tested the PR?

timvandermeij:
> If I open any PDF, I get `TypeError: this.parentGroup is null`.

I can reproduce this with e.g., the Tracemonkey PDF file. Closing since this PR cannot be merged in its current form. This is still useful to have, so the issue is tracked in #10935.

PR #11115, opened 2019-09-02
Ensure that `Error`s are handled correctly when using `postMessage` with Streams in `MessageHandler`

Snuffleupagus:
*Please refer to the individual commit messages for additional details.*

PR #11118, opened 2019-09-04
Transfer, rather than copy, CMap data to the worker-thread

Snuffleupagus:
It recently occurred to me that the CMap data should be an excellent candidate for transfering.
This will help reduce peak memory usage for PDF documents using CMaps, since transfering of data avoids duplicating it on both the main- and worker-threads.

Unfortunately it's not possible to actually transfer data when *returning* data through `sendWithPromise`, and another solution had to be used.
Initially I looked at using one message for requesting the data, and another message for returning the actual CMap data. While that should have worked, it would have meant adding a lot more complexity particularly on the worker-thread.
Hence the simplest solution, at least in my opinion, is to utilize `sendWithStream` since that makes it *really* easy to transfer the CMap data. (This required PR #11115 to land first, since otherwise CMap fetch errors won't propagate correctly to the worker-thread.)

Please note that the patch *purposely* only changes the API to Worker communication, and not the API *itself* since changing the interface of `CMapReaderFactory` would be a breaking change.
Furthermore, given the relatively small size of the `.bcmap` files (the largest one is smaller than the default range-request size) streaming doesn't really seem necessary either.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -78,8 +78,24 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       if (this.builtInCMapCache.has(name)) {
         return this.builtInCMapCache.get(name);
       }
-      const data = await this.handler.sendWithPromise('FetchBuiltInCMap',
-                                                      { name, });
+      const readableStream = this.handler.sendWithStream('FetchBuiltInCMap', {
+        name,
+      });
+      const reader = readableStream.getReader();
+
+      const data = await new Promise(function(resolve, reject) {
+        function pump() {
+          reader.read().then(function({ value, done, }) {
+            if (done) {
+              return;
+            }
+            resolve(value);
+            pump();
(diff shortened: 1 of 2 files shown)


timvandermeij:
I think this is a really good idea. Do you perhaps have measurements of the peak memory usage before/after this patch, or is that difficult to do?

Snuffleupagus:
> Do you perhaps have measurements of the peak memory usage before/after this patch, or is that difficult to do?

I'd expect that the amount of allocated memory is reduced by (approximately) the total size of the loaded CMap files. When looking at `about:memory` it *appears* that the size of the `ArrayBuffer` allocations are reduced by the expected amount, but please keep in mind that I'm really not very knowledgeable about these matters (i.e. measuring detailed memory usage).

timvandermeij (approved):

timvandermeij:
Nice work; thanks!

PR #11123, opened 2019-09-05
[api-minor] Remove the `postMessageTransfers` parameter, and thus the ability to manually disable transferring of data, from the API

Snuffleupagus:
By transferring, rather than copying, `ArrayBuffer`s between the main- and worker-threads, you can avoid unnecessary allocations by only having *one* copy of the same data.
Hence manually setting `postMessageTransfers: false`, when calling `getDocument`, is a performance footgun[1] which will do nothing but waste memory.

Given that every reasonably modern browser supports `postMessage` transfers[2], I really don't see why it should be possible to force-disable this functionality.
Looking at the browser support, for `postMessage` transfers[2], it's highly unlikely that PDF.js is even usable in browsers without it. However, the feature testing of `postMessage` transfers is kept for the time being just to err on the safe side.

---
[1] This is somewhat similar to the, now removed, `disableWorker` parameter which also provided API users a much too simple way of reducing performance.

[2] See e.g. https://developer.mozilla.org/en-US/docs/Web/API/MessagePort/postMessage#Browser_compatibility and https://developer.mozilla.org/en-US/docs/Web/API/Transferable#Browser_compatibility

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -73,19 +73,7 @@ var WorkerMessageHandler = {
       // making sure postMessage transfers are working
       var supportTransfers = data[0] === 255;
       handler.postMessageTransfers = supportTransfers;
-      // check if the response property is supported by xhr
-      var xhr = new XMLHttpRequest();
-      var responseExists = 'response' in xhr;
-      // check if the property is actually implemented
-      try {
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good! I also don't see any reason to keep this around.

PR #11129, opened 2019-09-08
Prevent "offsetParent is not set -- cannot scroll" errors when the viewer loads in e.g. a hidden <iframe>

Snuffleupagus:
Besides avoiding errors during loading, this also ensures that the document will be correctly scrolled/zoomed into view once the viewer becomes visible.
This "new" behaviour was always intended, see PR #2613, however various re-factoring over the years seem to have broken this (and I'm probably at least somewhat responsible for that).


timvandermeij (approved):

timvandermeij:
Thanks!

PR #11130, opened 2019-09-09
Don't `enqueue` unused properties when sending 'GetOperatorList' data from the worker-thread (PR 11069 follow-up)

Snuffleupagus:
With the changes made in PR #11069, it's no longer necessary to include the `pageIndex`/`intent` parameters when sending 'GetOperatorList' data. In the previous implementation these properties were used to associate the `OperatorList` with the correct `RenderTask`, however now that `ReadableStream`s are used that's handled automatically and it's thus dead code at this point.

### DIFF
--- a/src/core/operator_list.js
+++ b/src/core/operator_list.js
@@ -648,14 +648,10 @@ var OperatorList = (function OperatorListClosure() {
       this._totalLength += length;
 
       this._streamSink.enqueue({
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Nice simplification; thank you!

PR #11133, opened 2019-09-10
Enable the `no-async-promise-executor` ESLint rule

Snuffleupagus:
This rule is in the process of being rolled out in mozilla-central, and it helps avoid unnecessary `async` functions together with `new Promise(...)`.

Please see https://eslint.org/docs/rules/no-async-promise-executor for additional information.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -35,6 +35,7 @@
 
     // Possible errors
     "for-direction": "error",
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
More linting is always good. Thanks!

PR #11134, opened 2019-09-11
Ensure that `addLinkAttributes` is always called with a valid `url` parameter

Snuffleupagus:
There's no good reason for calling this helper function without a `url` parameter, and this way we can prevent that from happening.
Note how the `PDFOutlineViewer` call-site was already doing the right thing here, and only the `LinkAnnotationElement` call-site needed a small adjustment to make it work.

*Smaller diff with https://github.com/mozilla/pdf.js/pull/11134/files?w=1*


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #11138, opened 2019-09-11
[CRX] Ignore auto-generated <embed> from Chrome

### LINKED ISSUE #11137: Cannot print from Google Sheets with the PDF Viewer Chrome extension enabled in Chrome 76+
A user reported to me by mail that they were unable to print from Google sheets.
This used to work, but broke in Chrome 76 due to https://chromium.googlesource.com/chromium/src/+/f38273f2644b29a065de0d2585a3abab3214ecd6 and https://chromium.googlesource.com/chromium/src/+/7703c8f99911d949f162021a3c70b00314ca718c

Configuration:
- Web browser and its version: Chromium 77.0.3865.75 (fine in 75; this is a regression in Chrome 76).
- PDF.js version: 2.2.191
- Is a browser extension: yes

1. Open any spreadsheet in Google Docs, e.g. open a result from https://duckduckgo.com/?q=site%3Adocs.google.com%2Fspreadsheets&ia=web
2. Press Ctrl-P to open the print preview and click on "Next" to start printing.

The spreadsheet should be printed.

What went wrong?
The spreadsheet is not printed. The console shows:

```
Resource interpreted as Document but transferred with MIME type application/pdf: "https://docs.google.com/spreadsheets/d/1-3kyC2SYDvJwULt9FWL3A6Cy3jq0I5ox-DhQr64qydc/pdf?id=1-3kyC2SYDvJwULt9FWL3A6Cy3jq0I5ox-DhQr64qydc".
Access to XMLHttpRequest at 'about:blank' from origin 'chrome-extension://oemmndcbldboiebfnladdacbdfmadadm' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https.
GET about:blank net::ERR_FAILED
Uncaught (in promise) Error: Unexpected server response.
    at viewer.js:1103
```

Google Sheets sends a POST request to generate a document. Because POST requests are not intercepted by the extension, the extension does not attempt to intercept the load at first (even though the MIME type is "application/pdf"). Then, the browser (Chrome) renders a document with the following content:
`<!doctype html><html><body style='height: 100%; width: 100%; overflow: hidden; margin:0px; background-color: rgb(82, 86, 89);'><embed style='position:absolute; left: 0; top: 0;'width='100%' height='100%' src='about:blank' type='application/pdf' internalid='84F31E2198CABEAFE5E7308ACDFD2713'></embed></body></html>`

Because it's an `<embed>` with the PDF MIME type, the extension attempts to render it anyway. Since `about:blank` can obviously not contain a PDF, this breaks.

We should fix this, by ignoring URLs from `<embed>`s that cannot be loaded by the extension.

Rob--W:
Fixes #11137

### DIFF
--- a/extensions/chromium/contentscript.js
+++ b/extensions/chromium/contentscript.js
@@ -62,6 +62,12 @@ function watchObjectOrEmbed(elem) {
     // Until #4483 is fixed, POST requests should be ignored.
     return;
   }
+  if (elem.tagName === 'EMBED' && elem.src === 'about:blank') {
+    // Starting from Chrome 76, internal embeds do not have the original URL,
+    // but "about:blank" instead.
+    // See https://github.com/mozilla/pdf.js/issues/11137
+    return;
+  }
 
   if (elem.__I_saw_this_element) {
     return;


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #11142, opened 2019-09-13
Stop dispatching a "localized" event from the webL10n library, since it's unused and may clash with the default viewer

Snuffleupagus:
Currently this leads to confusing behaviour with `eventBusDispatchToDOM = true` set, since then there's *two* different "localized" events being fired.

### DIFF
--- a/external/webL10n/l10n.js
+++ b/external/webL10n/l10n.js
@@ -26,6 +26,9 @@
     - Removes consoleWarn and consoleLog and use console.log/warn directly.
     - Removes window._ assignment.
     - Remove compatibility code for OldIE.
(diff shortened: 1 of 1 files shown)


external/webL10n/l10n.js
```diff
@@ -26,6 +26,9 @@
     - Removes consoleWarn and consoleLog and use console.log/warn directly.
     - Removes window._ assignment.
     - Remove compatibility code for OldIE.
+    - Replaces `String.prototype.substr()` with `String.prototype.substring()`.
```
Snuffleupagus: This line *should* have been included with PR #10111.

timvandermeij (approved):

timvandermeij:
Looks good!

PR #11158, opened 2019-09-17
Avoid floating point inaccuracy in gradient color stops

### LINKED ISSUE #7339: Color bar produced by PGFPlots does not render correctly
- [output.pdf](https://github.com/mozilla/pdf.js/files/271335/output.pdf)
- [PGFPlots manual](http://pgfplots.sourceforge.net/pgfplots.pdf) (See e.g. section 4.7.6 Color Maps)

Configuration:
- Web browser and its version: Firefox 46.0, Nightly 49.0a1, Chrome 50.0
- Operating system and its version: Windows 7 64-bit
- PDF.js version: 
  - Version built into Firefox 46.0, Nightly 49.0a1
  - Version at https://mozilla.github.io/pdf.js/web/viewer.html

Steps to reproduce the problem: Open attached PDF.

What went wrong/expected behavior: 
screenshot

janpe2:
Fixes #7339 in which the `Domain` entry of an axial shading is `[0.0 161.39935]` instead of the default value `[0 1]`. The existing implementation has a problematic `for` loop: `i` and `step` are floating point values that are always inaccurate, so the end value `i === t1` never occurs. As a result, the last color stop is not created and an unpainted region appears at the end of the gradient.

I have changed the loop so that `i` is always an integer. Color stop offsets are 0.0, 0.1, 0.2, ..., 1.0 like they were previously.


janpe2:
I don't know why the regression test `firefox-bug878194-page4` failed. The page doesn't contain any shadings/gradients.

timvandermeij:
It looks like an intermittent failure since it doesn't happen on Linux. Let's try again...

/botio-windows test

timvandermeij (approved):

timvandermeij:
Nice find! I also find the new code easier to read.

/botio makeref

PR #11168, opened 2019-09-21
Remove unmaintained localizations and update translations and packages

### LINKED ISSUE #11167: Remove unmaintained localizations
A number of localizations were recently removed from Firefox Nightly[1], and it thus seem that we probably want to remove them here as well if they're not actively maintained any more.

This affects the following localizations (most of which haven't been updated for years):
`as`, `crh`, `en-ZA`, `mai`, `ml`, `or`

Removal would entail deleting the relevant directories from https://github.com/mozilla/pdf.js/tree/master/l10n, as well as updating the list of available locales at https://github.com/mozilla/pdf.js/blob/e91836f8324e368743d05d5c305f999f37332f70/external/importL10n/locales.js#L22-L36

---
[1] Please find additional details in https://bugzilla.mozilla.org/show_bug.cgi?id=1535299 and https://bugzilla.mozilla.org/show_bug.cgi?id=1573386

timvandermeij:
Fixes #11167.


timvandermeij:
Trivial.

PR #11169, opened 2019-09-22
Reduce the number of function calls in the `Dict` class

Snuffleupagus:
The following changes were made:
 - Remove unnecessary `typeof` checks in the `get`/`getAsync` methods.
 - Reduce unnecessary code duplication in the `get`/`getAsync` methods.
 - Inline the `Ref` checks in the `get`/`getAsync`/`getArray` methods, since it helps avoid many unnecessary functions calls. I.e. this way it's possible to directly call `XRef.{fetch, fetchAsync)` only when necessary, rather than always having to call `XRef.{fetchIfRef, fetchIfRefAsync)`.

This patch was tested using the PDF file from issue 2618, i.e. http://bugzilla-attachments.gnome.org/attachment.cgi?id=226471, using the following manifest file:
```
[
    {  "id": "issue2618",
       "md5": "",
       "rounds": 250,
       "type": "eq"
    }
]
```
```
Firefox | Overall      |   250 |         2821 |        2790 | -32 | -1.12 |        faster
Firefox | Page Request |   250 |            2 |           2 |   0 |  6.68 |
Firefox | Rendering    |   250 |         2820 |        2788 | -32 | -1.13 |        faster
```

### DIFF
--- a/src/core/primitives.js
+++ b/src/core/primitives.js
@@ -81,59 +81,47 @@ var Dict = (function DictClosure() {
     },
 
     // automatically dereferences Ref objects
-    get: function Dict_get(key1, key2, key3) {
-      var value;
-      var xref = this.xref, suppressEncryption = this.suppressEncryption;
-      if (typeof (value = this._map[key1]) !== 'undefined' ||
-          key1 in this._map || typeof key2 === 'undefined') {
-        return xref ? xref.fetchIfRef(value, suppressEncryption) : value;
+    get(key1, key2, key3) {
+      let value = this._map[key1];
+      if (value === undefined && !(key1 in this._map) && key2 !== undefined) {
+        value = this._map[key2];
+        if (value === undefined && !(key2 in this._map) && key3 !== undefined) {
+          value = this._map[key3];
+        }
       }
-      if (typeof (value = this._map[key2]) !== 'undefined' ||
-          key2 in this._map || typeof key3 === 'undefined') {
-        return xref ? xref.fetchIfRef(value, suppressEncryption) : value;
+      if (value instanceof Ref && this.xref) {
+        return this.xref.fetch(value, this.suppressEncryption);
       }
-      value = this._map[key3];
-      return xref ? xref.fetchIfRef(value, suppressEncryption) : value;
+      return value;
     },
 
     // Same as get(), but returns a promise and uses fetchIfRefAsync().
-    getAsync: function Dict_getAsync(key1, key2, key3) {
-      var value;
-      var xref = this.xref, suppressEncryption = this.suppressEncryption;
-      if (typeof (value = this._map[key1]) !== 'undefined' ||
-          key1 in this._map || typeof key2 === 'undefined') {
-        if (xref) {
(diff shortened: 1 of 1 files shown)


src/core/primitives.js
```diff
@@ -81,59 +81,47 @@ var Dict = (function DictClosure() {
...
+          value = this._map[key3] || null;
+        }
       }
-      if (typeof (value = this._map[key2]) !== 'undefined' ||
-          key2 in this._map || typeof key3 === 'undefined') {
-        return xref ? xref.fetchIfRef(value, suppressEncryption) : value;
+      if (value instanceof Ref && this.xref) {
+        return this.xref.fetch(value, this.suppressEncryption);
       }
-      value = this._map[key3] || null;
```
Snuffleupagus: This, and the identical line in the method below, looks strange to me. Why would you fallback to `null` in only *one* particular case, especially since it's outright wrong if `this._map[key3]` would ever be "falsy" (i.e. `zero` or `false`)!?

There's even a comment in one of the [unit-tests](https://github.com/mozilla/pdf.js/blob/9596d702a993bf3fc8a9a3694bf960ee0b83cbda/test/unit/primitives_spec.js#L71-L72) which mentions this inconsistency, and the only reason that this isn't breaking anything is that there's only *one* [call-site passing in three keys](https://github.com/mozilla/pdf.js/blob/9596d702a993bf3fc8a9a3694bf960ee0b83cbda/src/core/evaluator.js#L2557).

timvandermeij: Fixed in #11171, so with a rebase this is done.

src/core/primitives.js
```diff
@@ -81,59 +81,47 @@ var Dict = (function DictClosure() {
...
+      if (value === undefined && !(key1 in this._map) && key2 !== undefined) {
+        value = this._map[key2];
+        if (value === undefined && !(key2 in this._map) && key3 !== undefined) {
+          value = this._map[key3] || null;
+        }
       }
-      if (typeof (value = this._map[key2]) !== 'undefined' ||
-          key2 in this._map || typeof key3 === 'undefined') {
-        return xref ? xref.fetchIfRef(value, suppressEncryption) : value;
+      if (value instanceof Ref && this.xref) {
```
timvandermeij: Would it be better to (here and below) do the `this.xref` check first, given that it's probably cheaper than the `instanceof` check? These methods are usually called a lot, so maybe that's worth it.

Snuffleupagus: *It's possible that this is going too far into the territory of micro-optimizations, but:*
Statistically speaking most `Dict`s will have `this.xref` defined, whereas only some entries will contain references. By using this ordering, we can skip what most of the time amounts to redundant `this.xref` checks, but the difference is probably unmeasurable though...

timvandermeij: Right, I thought so, but thought it couldn't hurt to ask. Looks good in that case.

timvandermeij (approved):

timvandermeij:
Looks good, and also a lot more readable, to me. Thanks!

PR #11171, opened 2019-09-23
Fix the inconsistent return types for `Dict.{get, getAsync}`

Snuffleupagus:
Having these methods fallback to returning `null` in only *one* particular case seems outright wrong, since a "falsy" value will thus be handled incorrectly.
The only reason that this hasn't caused issues in practice is that there's only one call-site passing in three keys, and in that case we're trying to read a font file where falling back to `null` isn't a problem.


timvandermeij (approved):

timvandermeij:
The new situation looks much better. For reference, the only call-site was https://github.com/mozilla/pdf.js/blob/9596d702a993bf3fc8a9a3694bf960ee0b83cbda/src/core/evaluator.js#L2557.

Thanks!

PR #11173, opened 2019-09-23
Replace the bundled `ReadableStream` polyfill with the `web-streams-polyfill` npm package (issue 11157)

### LINKED ISSUE #11157: Investigate replacing the `ReadableStream` polyfill
Rather than having to maintain a `ReadableStream` polyfill in the PDF.js library, it would be much better to use an existing library/package.
Especially since our `ReadableStream` polyfill has never, as far as I can tell, been updated and given that it's taken directly from the reference implementation (i.e. it may not be as efficient as it could be).

The following one could be a good candidate (the author seem to be involved in the Streams Standard itself): https://github.com/MattiasBuelens/web-streams-polyfill (available [through npm as well](https://www.npmjs.com/package/web-streams-polyfill)), since it seems to be regularly updated/maintained.

Snuffleupagus:
Compared to the recently replaced `URL` polyfill, the new `ReadableStream` polyfill isn't being exported globally for two reasons:
 - We're currently checking for the existence of a global `ReadableStream` implementation when determining if the Fetch API will be used; please see `isFetchSupported` in the src/display/display_utils.js file.
 - Given that it's much newer functionality (compared to `URL`) and that not all browsers may implement all parts of the specification yet, not exposing the `ReadableStream` globally seems safer for now.

Fixes #11157


timvandermeij (approved):

timvandermeij:
This is a very nice clean-up!

Snuffleupagus:
Thanks for the reviews!

Would now perhaps be a good idea to increase the minor version number, since there's been a couple of `[api-minor]` PRs since the last release?

Furthermore, would now also be a good idea to create a new (pre-) release, since there's been a couple of recent PRs which are significant (e.g. Streams for `getOperatorList`, changed `getStats`, and the replacement of bundled polyfills).

Snuffleupagus:
Thanks for the reviews!

Would now perhaps be a good idea to increase the minor version number, since there's been a couple of `[api-minor]` PRs since the last release?

Furthermore, would now also be a good idea to create a new (pre-) release, since there's been a couple of recent PRs which may be significant (e.g. Streams for `getOperatorList` and the replacement of bundled polyfills)?

Snuffleupagus:
Thanks for the reviews!

Would now perhaps be a good idea to increase the minor version number, since there's been a couple of `[api-minor]` PRs since the last release?

Furthermore, would now also be a good idea to create a new (pre-) release, since there's been a couple of recent PRs which may be significant (e.g. Streams for `getOperatorList` and the replacement of bundled polyfills)?

Snuffleupagus:
Thanks for the reviews!

Would now perhaps be a good idea to increase the minor version number, since there's been a couple of `[api-minor]` PRs since the last release?

Furthermore, would now also be a good time to create a new (pre-) release, since there's been a couple of recent PRs which may be significant (e.g. Streams for `getOperatorList` and the replacement of bundled polyfills)?

Snuffleupagus:
Thanks for the reviews!

Would now perhaps be a good time increase the minor version number, since there's been a couple of `[api-minor]` PRs since the last release?

Furthermore, would now also be a good time to create a new (pre-) release, since there's been a couple of recent PRs which may be significant (e.g. Streams for `getOperatorList` and the replacement of bundled polyfills)?

Snuffleupagus:
Thanks for the reviews!

Would now perhaps be a good time to increase the minor version number, since there's been a couple of `[api-minor]` PRs since the last release?

Furthermore, would now also be a good idea to create a new (pre-) release, since there's been a couple of recent PRs which may be significant (e.g. Streams for `getOperatorList` and the replacement of bundled polyfills)?

Snuffleupagus:
Thanks for the reviews!

Would now perhaps be a good time to increase the minor version number, since there's been a couple of `[api-minor]` PRs fairly recently?

Furthermore, how about creating a new (pre-) release, since there's been a couple of recent PRs which may be significant (e.g. Streams for `getOperatorList` and the replacement of bundled polyfills)?

timvandermeij:
Yes, I'll make an issue to do the release soon.

PR #11178, opened 2019-09-26
made a dark mode

BraydenTW:
I thought that you should make your website have a dark mode

### DIFF
--- a/docs/contents/css/main.css
+++ b/docs/contents/css/main.css
@@ -1,4 +1,5 @@
 body {
+  background: #111;
 }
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
> I thought that you should make your website have a dark mode

There's no explanation of why, which seems especially relevant considering that the proposed changes results in the text becoming almost unreadable; see an excerpt below:

unreadable

I'd suggest closing this as WONTFIX, since it doesn't seem necessary to tweak all of the CSS just to be able to change the background colour.

PR #11185, opened 2019-09-28
Convert the various `...Exception`s to proper classes, to reduce code duplication

Snuffleupagus:
By utilizing a base "class", things become significantly simpler. Unfortunately the new `BaseException` cannot be a proper ES6 class and just extend `Error`, since the SystemJS dependency doesn't seem to play well with that.
Note also that we (generally) need to keep the `name` property on the actual `...Exception` object, rather than on its prototype, since the property will otherwise be dropped during the structured cloning used with `postMessage`.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -401,99 +401,54 @@ function shadow(obj, prop, value) {
   return value;
 }
 
-const PasswordException = (function PasswordExceptionClosure() {
-  function PasswordException(msg, code) {
-    this.name = 'PasswordException';
-    this.message = msg;
+const BaseException = (function BaseExceptionClosure() {
+  function BaseException(message) {
+    if (this.constructor === BaseException) {
+      unreachable('Cannot initialize BaseException.');
+    }
+    this.message = message;
+    this.name = this.constructor.name;
+  }
+  BaseException.prototype = new Error();
+  BaseException.constructor = BaseException;
+
+  return BaseException;
+})();
+
+class PasswordException extends BaseException {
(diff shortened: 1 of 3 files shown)


src/shared/util.js
```diff
@@ -401,99 +401,54 @@ function shadow(obj, prop, value) {
   return value;
 }
 
-const PasswordException = (function PasswordExceptionClosure() {
-  function PasswordException(msg, code) {
-    this.name = 'PasswordException';
-    this.message = msg;
-    this.code = code;
+const BaseException = (function BaseExceptionClosure() {
```
Snuffleupagus: Once extending `Error` isn't a problem any more, when SystemJS has been updated, it should possible to simplify even further to e.g.
```javascript
class BaseException extends Error {
  constructor(message) {
    if (this.constructor === BaseException) {
      unreachable('Cannot initialize BaseException.');
    }
    super(message);
    this.name = this.constructor.name;
  }
}
```

Snuffleupagus: Once extending `Error` isn't a problem any more, when SystemJS has been updated/replaced, it should possible to simplify this even further to e.g.
```javascript
class BaseException extends Error {
  constructor(message) {
    if (this.constructor === BaseException) {
      unreachable('Cannot initialize BaseException.');
    }
    super(message);
    this.name = this.constructor.name;
  }
}
```

timvandermeij (approved):

timvandermeij:
Looks much cleaner. Thanks!

PR #11186, opened 2019-09-29
Improve the heuristics, in `PartialEvaluator._buildSimpleFontToUnicode`, for glyphNames of the Cdd{d}/cdd{d} format (issue 9655)

### LINKED ISSUE #9655: Content messed up in the textLayer div
[problem.pdf](https://github.com/mozilla/pdf.js/files/1901514/problem.pdf)

Configuration:
- Web browser and its version: Firefox Developer Edition 60.0b11 (64-bit)
- Operating system and its version: Windows 10 Pro 1709
- PDF.js version: Built-in version inside the Firefox browser.
- Is a browser extension: Built-in PDF.js inside the firefox browser.

1. Right click on the PDF file, and select Open With
2. Select Firefox Developer Edition.
3. Try to select text and find it is not possible.
4. Press F12 to open the Inspector, and locate the textLayer div.
5. Found that the font-size for each div is abnormal, and the text content for each div is somehow messed up and has nothing to do with the text content which we see on the PDF.

We should be able to select the text displayed, and the text content inside the textLayer div should match with what we see.
correct

incorrect

None

Snuffleupagus:
*Please note:* I've been thinking about possible ways of addressing this issue for a while now, but all of the solutions I came up with became too complicated and thus hurt readability of the code.
However, it occured to me that we're essentially trying to add a heuristic *on top* of another heuristic, and that it shouldn't matter how efficient the code is as long as it works.

In the PDF file in the issue the Encoding contains glyphNames of the `Cdd` format, which our existing heuristics will treat as base 10 values. However, in this particular file they actually contain base 16 values, which we thus attempt to detect and fix such that text-selection works.

Fixes #9655.


timvandermeij:
Thank you; it keep amazing me how many non-compliant PDF files exist and how different their breakage is...

kuyeduwu:
thank you for fixing this

PR #11187, opened 2019-10-01
Convert the various image decoder `...Error`s to classes extending `BaseException` (PR 11185 follow-up)

Snuffleupagus:
Somehow I missed these in PR #11185, but there's no good reason not to convert them as well.


timvandermeij (approved):

timvandermeij:
Thanks!

Issue #11175, opened 2019-09-24
How to fallback to some system font when the font is not embedded?


huzjakd:
Configuration:
Custom application. Any browser, Windows, version 2.1.266.

Steps to reproduce the problem:
1. Open any PDF that doesn't have some fonts embedded.
2. You get a console message about missing fonts. It is logged by evaluator.js in "loadFont" function ("fontRef not available").

What is the expected behavior?
It is all as expected.

What went wrong?
When it gets to this point, instead of the console message, I would like the pdf.js to load some default system font. I see there is a function in fonts.js: "fallbackToSystemFont" but I can't quite get it to work.

timvandermeij:
Fixed by the pull request above.

Issue #11162, opened 2019-09-19
PDF with data on top of it - all data is lost when opening via pdf.js


danielstachowiak:
Here is the example file that fails to open correctly in pdf.js on your demo page (and in our instance):
[dummkj19.pdf](https://github.com/mozilla/pdf.js/files/3631530/dummkj19.pdf)
It is a layered document (a pdf with data on top of it). Only the pdf opens, the data is lost.

Configuration:
- Web browser and its version: any browser
- Operating system and its version: any operating system
- PDF.js version: Your own DEMO version
- Is a browser extension: no

Steps to reproduce the problem:
1. Upload the above file via the Demo button on pdf.js website

2. Expected result: a pdf with data
![2](https://user-images.githubusercontent.com/454432/65250302-46b6cc00-daed-11e9-9f46-f77b690d6bcc.jpg)

3. Actual result: empty pdf (all data is lost)
![1](https://user-images.githubusercontent.com/454432/65250226-27b83a00-daed-11e9-9990-ee6c10467779.jpg)

We are happy to pay a modest fee for a quick fix.
Please contact support@mydocsafehq.com if you would like to discuss this further.

devstore49:
I`m sure to resolve this issue.
Thank you.

LuckyStar121:
I need to check your code right now.

Snuffleupagus:
There's something wrong with one of the `ExtGState` entries for the first page, either in the PDF file itself and/or in the way that PDF.js handles GState.
However, even if that would be fixed the PDF file is still corrupt since it doesn't embed *all* of its fonts (i.e. the same problem as outlined in https://github.com/mozilla/pdf.js/issues/11068#issuecomment-521589843).

*Finally, given the two comments just above:* Unless you have something add to an issue, please refrain from posting essentially meaningless/empty comments since that just clutters the discussion.

THausherr:
Amazingly, the font that is missing in the resources has the same name! (Ffcs1)

Snuffleupagus:
Should this be closed as fixed now, given the PR referenced above?

PR #11191, opened 2019-10-02
Embedded Arimo font widths and used that when fonts are missing

huzjakd:
This fixes this bug: https://github.com/mozilla/pdf.js/issues/11162
And answers my question: https://github.com/mozilla/pdf.js/issues/11175

I embedded Arimo font widths and used them to fall back to a default font when the fonts are missing.

It passes lint tests but doesn't pass 8-10 others.
When comparing screenshots of failed tests the only differences are a few pixels that are not visible by the naked eye.
Also in the attached file, you don't normally see the number 3 in "Page 13" but with this fix, you see it which causes the test to fail.
[pr4922.pdf](https://github.com/mozilla/pdf.js/files/3681426/pr4922.pdf)


timvandermeij:
I can't really review this implementation because I'm not too familiar with the font conversion code so I'll leave that for others, but I think in general the idea of falling back to a default font is better than not having any font at all.

src/core/evaluator.js
```diff
@@ -723,29 +723,70 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           return errorFont();
         }
       }
+      let fallbackFont = false;
+      let fallbackFontProperties = false;
       if (!fontRef) {
-        warn('fontRef not available');
-        return errorFont();
+        // warn('fontRef not available');
+        // return errorFont();
```
Snuffleupagus: Please note that commented-out code isn't allowed in general, and in this case you very much want to print the warning since there's simply no guarantee that things will actually work particularly well in general with this sort work-around.

Also, an sending an `UnsupportedFeature` notification would seem appropriate as well.

src/core/evaluator.js
```diff
@@ -723,29 +723,70 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+
+        // fallback to some default font
```
Snuffleupagus: Nit: Please ensure that added comments always are complete sentences, i.e. starting with an upper-case letter and end with a dot.

src/core/evaluator.js
```diff
@@ -723,29 +723,70 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+        let widthsWithNulls = [null, null, null, null, null, null, null,
```
Snuffleupagus: For the general PDF.js library, you should use widths corresponding a *standard font* and we also want to avoid defining this data inline.

Please find already embedded *standard font* widths in https://github.com/mozilla/pdf.js/blob/master/src/core/metrics.js, where the `Helvetica` ones seem like a good choice for the fallback font being constructed here.

src/core/evaluator.js
```diff
@@ -723,29 +723,70 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+        let widthsWithoutNulls = [];
+        for (let i = 0; i < widthsWithNulls.length; i++) {
+          if (widthsWithNulls[i] !== null) {
+            widthsWithoutNulls[i] = widthsWithNulls[i];
+          }
+        }
+        fallbackFontProperties = {
+          fallbackFont: true,
+          firstChar: 0,
+          name: 'BCDEEE+Arimo',
```
Snuffleupagus: This name seems almost like it's taken from an *actual* PDF file, which doesn't seem entirely appropriate for a *general* fallback font; how about simply 'PDFJS-FallbackFont' or similar instead.

src/core/evaluator.js
```diff
@@ -723,29 +723,70 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+        278, 469, 556, 333, 556, 556, 500, 556, 556, 278, 556, 556, 222,
+        222, 500, 222, 833, 556, 556, 556, 556, 333, 500, 278, 556, 500,
+        722, 500, 500, 500, 334, 260, 334, 584];
+        let widthsWithoutNulls = [];
+        for (let i = 0; i < widthsWithNulls.length; i++) {
+          if (widthsWithNulls[i] !== null) {
+            widthsWithoutNulls[i] = widthsWithNulls[i];
+          }
+        }
+        fallbackFontProperties = {
```
Snuffleupagus: Please note that a plain `Object` isn't an appropriate data-structure here, since it for one seem to cause *a lot* of unwanted special-casing throughout the font parsing code below.

Instead you need to create a *valid* font dictionary here, with dummy entries as needed, to ensure that the fallback font dictionary can be parsed as-is by the relevant code below.

Snuffleupagus (changes requested):
I've tried to do an initial review pass, please see the inline comments.

Please note that the single most important point here is that, for reasons of both readability and maintainable of the code, you need to create a *valid* font dictionary such that all of the special cases (in the various font parsing functions) can be removed.

huzjakd:
The way I see it when some font is not embedded in the pdf, the pdf.js always logs a warning and returns the ErrorFont and doesn't display anything in the pdf. At this point, I just want to embed/construct my own font and use it to display something anyway.

timvandermeij:
Closing since the pull request above built upon the changes in this pull request and was merged. Thank you for contributing this!

PR #11206, opened 2019-10-04
Preserve Color information (CMYK) to display stage

kriffe:
To make pdf.js more usable together with color critical applications the original color information (like CMYK) should be kept all the way to the final stages of display before converted to rgb. 

This would open up for more possibilities of using the correct color information from a pdf document. Examples of this could be helping initiatives like the dropped svg 1.2 draft https://www.w3.org/TR/SVGColor12/#device, or allowing usage of alternative CMYK conversion methids, or opening up for more possibilities to export pdf other formats with preserved color info)

**Example**: Using the page.getOperatorList() call would still have the correct color information if the CMYK wasnt converted in the RGB (can be used in pdf to svg conversion of a page) 
```js
page.getOperatorList().then(function (opList) {
    const svgGfx = new SVGGraphics(page.commonObjs, page.objs)
    return svgGfx.getSVG(opList, viewport)
})
```

**EDITED: Proposed changes**
* Keep CMYK info when doing rgb conversion in evaluator.js (https://github.com/mozilla/pdf.js/blob/master/src/core/evaluator.js#L1137)
* Add the rgb args to svg.js and canvas.js and leave the cmyk for now (https://github.com/mozilla/pdf.js/blob/master/src/display/svg.js#L584 & 
 https://github.com/mozilla/pdf.js/blob/master/src/display/canvas.js#L1695)
* ~~Move ColorSpace.js and Primitives.js from core to shared (needed for the move of conversion). Causes all the file changes~~

Please see the canvas.js and svg.js files for the changes. ~~The rest of the changes should only be due to the needed move of ColorSpace.js and Primitives.js to shared.~~

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -14,16 +14,15 @@
  */
 
 import {
-  AbortException, assert, CMapCompressionType, ColorSpace,
-  createPromiseCapability, FONT_IDENTITY_MATRIX,
-  FormatError, IDENTITY_MATRIX, info, isArrayEqual, isNum,
-  isString, NativeImageDecoding, OPS, stringToPDFString,
-  TextRenderingMode, UNSUPPORTED_FEATURES, Util, warn
+  AbortException, assert, CMapCompressionType, createPromiseCapability,
+  FONT_IDENTITY_MATRIX, FormatError, IDENTITY_MATRIX, info, isArrayEqual, isNum,
+  isString, NativeImageDecoding, OPS, stringToPDFString, TextRenderingMode,
+  UNSUPPORTED_FEATURES, Util, warn
 } from '../shared/util';
 import { CMapFactory, IdentityCMap } from './cmap';
 import {
   Cmd, Dict, EOF, isDict, isName, isRef, isStream, Name
-} from '../shared/primitives';
+} from './primitives';
 import {
   ErrorFont, Font, FontFlags, getFontType, IdentityToUnicodeMap, ToUnicodeMap
 } from './fonts';
@@ -40,6 +39,7 @@ import {
 import { getTilingPatternIR, Pattern } from './pattern';
 import { Lexer, Parser } from './parser';
 import { bidi } from './bidi';
+import { ColorSpace } from './colorspace';
 import { DecodeStream } from './stream';
 import { getGlyphsUnicode } from './glyphlist';
 import { getLookupTableFactory } from './core_utils';
@@ -1134,6 +1134,16 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               args = ColorSpace.singletons.gray.getRgb(args, 0);
               fn = OPS.setStrokeRGBColor;
               break;
+            // case OPS.setFillCMYKColor:
+            //   stateManager.state.fillColorSpace = ColorSpace.singletons.cmyk;
+            //   args = ColorSpace.singletons.cmyk.getRgb(args, 0);
+            //   fn = OPS.setFillRGBColor;
+            //   break;
+            // case OPS.setStrokeCMYKColor:
+            //   stateManager.state.strokeColorSpace = ColorSpace.singletons.cmyk;
+            //   args = ColorSpace.singletons.cmyk.getRgb(args, 0);
+            //   fn = OPS.setStrokeRGBColor;
+            //   break;
             case OPS.setFillRGBColor:
               stateManager.state.fillColorSpace = ColorSpace.singletons.rgb;
               args = ColorSpace.singletons.rgb.getRgb(args, 0);
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1136,11 +1136,13 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               break;
             case OPS.setFillCMYKColor:
               stateManager.state.fillColorSpace = ColorSpace.singletons.cmyk;
-              args.push(...ColorSpace.singletons.cmyk.getRgb(args, 0));
+              args = ColorSpace.singletons.cmyk.getRgb(args, 0);
+              fn = OPS.setFillCMYKColor;
               break;
             case OPS.setStrokeCMYKColor:
               stateManager.state.strokeColorSpace = ColorSpace.singletons.cmyk;
-              args.push(ColorSpace.singletons.cmyk.getRgb(args, 0));
+              args = ColorSpace.singletons.cmyk.getRgb(args, 0);
+              fn = OPS.setStrokeCMYKColor;
               break;
             case OPS.setFillRGBColor:
               stateManager.state.fillColorSpace = ColorSpace.singletons.rgb;
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1134,16 +1134,16 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               args = ColorSpace.singletons.gray.getRgb(args, 0);
               fn = OPS.setStrokeRGBColor;
               break;
-            case OPS.setFillCMYKColor:
-              stateManager.state.fillColorSpace = ColorSpace.singletons.cmyk;
-              args = ColorSpace.singletons.cmyk.getRgb(args, 0);
-              fn = OPS.setFillRGBColor;
-              break;
-            case OPS.setStrokeCMYKColor:
-              stateManager.state.strokeColorSpace = ColorSpace.singletons.cmyk;
-              args = ColorSpace.singletons.cmyk.getRgb(args, 0);
-              fn = OPS.setStrokeRGBColor;
-              break;
+            // case OPS.setFillCMYKColor:
+            //   stateManager.state.fillColorSpace = ColorSpace.singletons.cmyk;
+            //   args = ColorSpace.singletons.cmyk.getRgb(args, 0);
+            //   fn = OPS.setFillRGBColor;
+            //   break;
+            // case OPS.setStrokeCMYKColor:
+            //   stateManager.state.strokeColorSpace = ColorSpace.singletons.cmyk;
+            //   args = ColorSpace.singletons.cmyk.getRgb(args, 0);
+            //   fn = OPS.setStrokeRGBColor;
+            //   break;
             case OPS.setFillRGBColor:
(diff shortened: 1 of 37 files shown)


Snuffleupagus:
Please note that all parsing, which includes ColourSpaces, is *purposely* done on the worker-thread for performance reasons.[1] Although we *used* to handled ColourSpaces on the main-thread, that was changed *deliberately* in PR #4824 as part of addressing issue #3474.

Hence I personally don't believe that we'd want to revert this, since it would first of all risk regressing performance. Furthermore, it would also create a situation where different ColourSpaces are handled inconsistently, which does not seem great from an API perspective, and it would also add additional code complexity despite no code/tests in the PDF.js library actually utilizing this.

---
[1] And as far as I'm aware, browsers also aren't generally able to use anything other than RGB anyway.

kriffe:
The entire printing industry is moving towards using more web applications in their business models, and pdf.js could be a very useful library in that context. This is the main reason why discarding the color information is bad, even if its done for performance and that the browser display wont handle anything other than RGB in the near future.

Scenario where valuable CMYK information is lost (its not the same pdf any more)
```
Open with pdf.js -> parse in webbapp -> Export to pdf -> Print document
```

Primitives was not intended to be moved. Perhaps there are other ways to solve this than my first attempt?

**Alternative ways to preserve color info**
* Convert in core and keep both cmyk and rgb values in the args list
* Isolate only the primitive used in ColorSpace
* Optional args somewhere how the document should be parsed (keepOriginalColorInfomration: true)


What do you think?

kriffe:
Went forward with implementing the alternative with using args for cmyk information, instead of moving the ColorSpace and primitives functions from core

* Performance should now be unaffected
* ColorSpace and Primitives stays in core
* Consistent color conversion internally but possible for accessing cmyk through page.getOperatorList()

src/core/evaluator.js
```diff
@@ -1136,13 +1136,11 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               break;
             case OPS.setFillCMYKColor:
               stateManager.state.fillColorSpace = ColorSpace.singletons.cmyk;
-              args = ColorSpace.singletons.cmyk.getRgb(args, 0);
-              fn = OPS.setFillRGBColor;
+              args.push(...ColorSpace.singletons.cmyk.getRgb(args, 0));
```
Snuffleupagus: One problem with these changes is that you're now passing completely unparsed/non-validated data to the main-thread, which you do *not* want to do since it will throw Errors on `postMessage` if the `args` isn't of the expected format (which could happen for e.g. corrupt documents).

Trying to address this would (again) add runtime overhead, despite it being completely unused in the default PDF.js library.

src/display/canvas.js
```diff
@@ -1698,6 +1699,14 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       this.current.fillColor = color;
       this.current.patternFill = false;
     },
+    setStrokeCMYKColor: function CanvasGraphics_setStrokeCMYKColor(c,
+                                                    m, y, k, r, g, b) {
+      this.setStrokeRGBColor(r, g, b);
```
Snuffleupagus: This, and the similar changes below, will add run-time overhead (two function calls rather than just one) to the default PDF.js use-case despite this not really being used for anything, which overall doesn't seem great.

Snuffleupagus (changes requested):
There's unfortunately still potential issues with this approach, see inline comments, and personally I'm thus still not convinced that it's a good idea.
Also, the updated patch still results in an "inconsistent" API-surface for different ColourSpaces.
Finally, this adds unused/untested code to the default PDF.js library which isn't a good idea since that makes maintaining the code much harder in the future.

---

Since this PR mentions the SVG-backend, please note that that [as mention in the FAQ](https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#backends) isn't officially supported/recommended at this point.

kriffe:
@Snuffleupagus My hopes would be that the steps like this would take the library closer towards usage of the svg output, since it opens up more possibilities to use the library (not just a viewer)

src/core/evaluator.js
kriffe: I'm not sure I follow the logic here. The getRgb(..) call is the same as the original code and the only difference is how its stacked in the args array. Is there logic in the original getRgb() that asserts that it always is a 3 argument array coming out?

Perhaps a solution with rgb first and cmyk as the following args solves this
```js
var rgb = ColorSpace.singletons.cmyk.getRgb(args, 0);
args = [rgb[0], rgb[1], rgb[2], args[0], args[1], args[2], args[3]]
```

Extension for handling the undefined case (if not handled anywhere else could be)
```js
args = [ 
             rgb[0] || 0,
             rgb[1] || 0, 
             rgb[2] || 0, 
             args[0] || 0, 
             args[1] || 0, 
             args[2] || 0, 
             args[3] || 0
]
```

src/display/canvas.js
kriffe: Is code duplication prefered to duplicate calls?

```js
setStrokeCMYKColor: function CanvasGraphics_setStrokeCMYKColor(c,
                                                    m, y, k, r, g, b) {
      var color = Util.makeCssRgb(r, g, b);
      this.ctx.fillStyle = color;
      this.current.fillColor = color;
      this.current.patternFill = false;
}
```

src/core/evaluator.js
Snuffleupagus: The "original" `args` array could contain *anything*, i.e. even un-clonable data, and you're now passing it as-is without any validation. With the current code, any unexpected data-structure would lead to an Error that we're able to handle gracefully within `PartialEvaluator.getOperatorList`, however with these changes it may instead cause a DataCloneError during `postMessage`.

src/display/canvas.js
Snuffleupagus: Generally you'd want to avoid code-duplication as well, especially when it's essentially unused by default.

Snuffleupagus:
> My hopes would be that the steps like this would take the library closer towards usage of the svg output,

Unfortunately, the SVG-backend is probably quite far from being generally usable.

And as mentioned previously, it would be unfortunate to have inconsistent API output for different ColourSpaces (i.e. what about all the more "exotic" ones).

> [...] (not just a viewer)

Well, the viewer is *the* main purpose for this library existing in the first place :-)

src/core/evaluator.js
kriffe: Will it not throw the error when parsing the invalid args in the getRgb(..)?

src/display/canvas.js
kriffe: I would reason the same. But I didnt think that the overhead for two function calls in a row had any significant impact on the overall performance in the application (as the solution is implemented right now).

kriffe:
> And as mentioned previously, it would be unfortunate to have inconsistent API output for different ColourSpaces (i.e. what about all the more "exotic" ones).

@Snuffleupagus Regarding the inconsisten api output I agree that it is nice with a single way of describing color. But its unfortunate that the rgb representation drops information and thus will always loose color information for all types of color representations. Using a standard that describes the full color space would be the ideal. Second to that I would rank preserving the original information, even if it leads to different number of args depending on color type.

> Well, the viewer is the main purpose for this library existing in the first place :-)

Even if its just used as a viewer it would still be nice that you could be used for printing direct from what you see with correct color representation (Ctrl + P in viewer today), even if the browsers arent quite ready for that as of today. But I personally think the pdf.js has grown into such a wide spread library for handling the pdf in a web/node context that it would be a shame if it wasnt used for more than just human viewing of the pdfs. :-)

src/core/evaluator.js
Snuffleupagus: Maybe, but given just how badly some PDF generators can corrupt a document you quite simply cannot depend on that. Sending non-validated data, in any form, is really a non-starter here (and elsewhere).

src/display/canvas.js
Snuffleupagus: Reducing the total number of function calls can add-up in certain cases, and even if this particular instance perhaps isn't too hot, it still doesn't seem necessary when it's (again) essentially unused.

Snuffleupagus:
> Using a standard that describes the full color space would be the ideal. Second to that I would rank preserving the original information, even if it leads to different number of args depending on color type.

For me personally, I'm still of the opinion that inconsistent handling of data in the API is never a good idea.
Also, note that in general you don't want to add what amounts to unused code, since that may cause unnecessary overhead later on. It's always easy to add code, but then it needs to be maintained :-)

> Even if its just used as a viewer it would still be nice that you could be used for printing direct from what you see with correct color representation (Ctrl + P in viewer today),

In general printing support is *very* limited in most browsers (marginally better with `mozPrintCallback` in Firefox), and most likely there's many other roadblocks to clear before printing becomes even halfway decent.

timvandermeij:
I'm also not really in favor of the inconsistency in the API and the fact that additional data is always passed that the viewer doesn't use. Would it be an idea to do it differently by giving `getOperatorList` an additional parameter, say `convertColorSpace`, defaulting to `true`, so that other consumers of the API can set it to `false` to get the original color space data? I hope doing so would keep the number of code changes small and we don't affect the current library use by default.

kriffe:
@timvandermeij : I think the usage of an optional argument on the getOperatorList would be a very elegant solution.

From what I see this seems to need the following changes
* Passing of an optional argument in the api.js getOperatorList to the worker thread https://github.com/mozilla/pdf.js/blob/master/src/display/api.js#L1115
* Passing the argument into the PartialEvaluator https://github.com/mozilla/pdf.js/blob/master/src/core/document.js#L210 (could the existing options arg object be used?)
* Make fill/stroke color operators return the original color information to arg instead of a rgb Uint8ClampedArray (perhaps the .base object present in at least some of the colors)?

Could this be correct or am I missing some other side effects?

timvandermeij:
I think that's correct. Given that we have used optional (feature) arguments before, see https://github.com/mozilla/pdf.js/commit/cfc2f36f5c00f17317fe68e1fc74be4566d1767a amongst others where an optional `background` parameter is added, it seems like a good solution to me because the normal behavior is not changed and unit tests can be added that provide the optional parameter so we have test coverage.

timvandermeij:
Closing since no changes were made to this in over a month. If the changes here are still relevant, please open a new pull request with the approach above and a unit test.

PR #11209, opened 2019-10-06
Use square brackets for optional properties in the JSDoc comments of `src/display/api.js`

tamuratak:
use square brackets to indicate properties are [optional](https://devdocs.io/jsdoc/tags-property) in JSDoc. Related to #10575. Without the square brackets, that PR would generate an incorrect `d.ts` file.

The diff becomes lengthy due to the restriction of maximal line length. If I can remove the word `(optional)`, it becomes much simpler.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -127,32 +127,31 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  * Document initialization / loading parameters object.
  *
  * @typedef {Object} DocumentInitParameters
- * @property {string}     url   - The URL of the PDF.
- * @property {TypedArray|Array|string} data - Binary PDF data. Use typed arrays
- *   (Uint8Array) to improve the memory usage. If PDF data is BASE64-encoded,
- *   use atob() to convert it to a binary string first.
- * @property {Object}     httpHeaders - Basic authentication headers.
- * @property {boolean}    withCredentials - Indicates whether or not cross-site
- *   Access-Control requests should be made using credentials such as cookies
- *   or authorization headers. The default is false.
- * @property {string}     password - For decrypting password-protected PDFs.
- * @property {TypedArray} initialData - A typed array with the first portion or
- *   all of the pdf data. Used by the extension since some data is already
+ * @property {string}     [url] - The URL of the PDF.
+ * @property {TypedArray|Array|string} [data] - Binary PDF data. Use typed
+ *    arrays (Uint8Array) to improve the memory usage. If PDF data is
+ *    BASE64-encoded, use atob() to convert it to a binary string first.
+ * @property {Object}     [httpHeaders] - Basic authentication headers.
+ * @property {boolean}    [withCredentials] - Indicates whether or not
+ *   cross-site Access-Control requests should be made using credentials such
+ *   as cookies or authorization headers. The default is false.
+ * @property {string}     [password] - For decrypting password-protected PDFs.
+ * @property {TypedArray} [initialData] - A typed array with the first portion
+ *   or all of the pdf data. Used by the extension since some data is already
  *   loaded before the switch to range requests.
- * @property {number}     length - The PDF file length. It's used for progress
- *   reports and range requests operations.
- * @property {PDFDataRangeTransport} range
- * @property {number}     rangeChunkSize - Optional parameter to specify
- *   maximum number of bytes fetched per range request. The default value is
- *   2^16 = 65536.
- * @property {PDFWorker}  worker - (optional) The worker that will be used for
+ * @property {number}     [length] - The PDF file length. It's used for
+ *   progress reports and range requests operations.
+ * @property {PDFDataRangeTransport} [range]
+ * @property {number}     [rangeChunkSize] - Specify maximum number of bytes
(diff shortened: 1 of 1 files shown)


src/display/api.js
```diff
@@ -142,63 +142,63 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
...
  *   and 'none' where JPEG images will be decoded entirely by PDF.js.
  *   The default value is 'decode'.
- * @property {string} cMapUrl - (optional) The URL where the predefined
+ * @property {string} [cMapUrl] - (optional) The URL where the predefined
  *   Adobe CMaps are located. Include trailing slash.
- * @property {boolean} cMapPacked - (optional) Specifies if the Adobe CMaps are
- *   binary packed.
- * @property {Object} CMapReaderFactory - (optional) The factory that will be
+ * @property {boolean} [cMapPacked] - (optional) Specifies if the Adobe CMaps
+ * are binary packed.
```
Snuffleupagus: Please fix the indentation here.

src/display/api.js
```diff
@@ -142,63 +142,63 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
...
  *   requests, then the PDF will be fetched in chunks.
  *   The default value is `false`.
- * @property {boolean} disableStream - (optional) Disable streaming of PDF file
- *   data. By default PDF.js attempts to load PDFs in chunks.
- *   The default value is `false`.
- * @property {boolean} disableAutoFetch - (optional) Disable pre-fetching of PDF
- *   file data. When range requests are enabled PDF.js will automatically keep
- *   fetching more data even if it isn't needed to display the current page.
+ * @property {boolean} [disableStream] - (optional) Disable streaming of PDF
+ *   file data. By default PDF.js attempts to load PDFs in chunks.
```
Snuffleupagus: Please don't remove the ```The default value is `false`.``` line here.

src/display/api.js
```diff
@@ -807,9 +807,10 @@ class PDFDocumentProxy {
  *
  * @typedef {Object} GetViewportParameters
  * @property {number} scale - The desired scale of the viewport.
- * @property {number} rotation - (optional) The desired rotation, in degrees, of
- *   the viewport. If omitted it defaults to the page rotation.
- * @property {boolean} dontFlip - (optional) If true, the y-axis will not be
+ * @property {number} [rotation] - (optional) The desired rotation,
+ *   in degrees, of *   the viewport. If omitted it defaults to the page
```
Snuffleupagus: Please remove the `*   ` inserted above.

Snuffleupagus (changes requested):
Please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits, and it would be a good idea to update the commit message to actually include JSDoc (since otherwise it's really impossible to understand what the patch does without looking at the code).

src/display/api.js
tamuratak: Thank you for your quick reply. Fixed.

src/display/api.js
tamuratak: Fixed

src/display/api.js
tamuratak: I think the line is not removed.

tamuratak:
The commits are squashed, and the word `JSDoc` is included in the commit message, as requested.

src/display/api.js
```diff
@@ -807,9 +807,10 @@ class PDFDocumentProxy {
  *
  * @typedef {Object} GetViewportParameters
  * @property {number} scale - The desired scale of the viewport.
- * @property {number} rotation - (optional) The desired rotation, in degrees, of
- *   the viewport. If omitted it defaults to the page rotation.
- * @property {boolean} dontFlip - (optional) If true, the y-axis will not be
+ * @property {number} [rotation] - (optional) The desired rotation,
+ *   in degrees, of the viewport. If omitted it defaults to the page
+ *   rotation.
```
Snuffleupagus: Doesn't this fit on the previous line now?

tamuratak: Fixed.

timvandermeij:
Here is how this renders now: http://54.67.70.0:8877/c86997eac19cc98/api/draft/global.html#DocumentInitParameters

I agree that the double optional bit is not necessary anymore, given that `(optional)` or other variants were used before because we didn't use the brackets, which are a better solution.

Snuffleupagus:
> Here is how this renders now: http://54.67.70.0:8877/c86997eac19cc98/api/draft/global.html#DocumentInitParameters

*This is obviously a pre-existing issue, however:*
Looking at this list, shouldn't *all* properties actually be marked as optional though?

timvandermeij:
Yes, all properties should be optional as far as I can tell. If that and https://github.com/mozilla/pdf.js/pull/11209#issuecomment-538740830 are done, this patch looks good to me.

tamuratak:
All the properties of DocumentInitParameters are marked as optional.

`intent` of RenderParameters is also marked as optional.

timvandermeij (approved):

timvandermeij:
Nice work; thank you for improving the documentation!

PR #11210, opened 2019-10-06
Delete the `@ignore` tag in the JSDoc comment for the `PDFObjects` class

tamuratak:
delete `@ignore` for `class PDFObjects` in JSDoc. Related to #10575. Due to the `@ignore`, `PDFObjects` is not included in the `d.ts` file generated in that PR.

Since `PDFObjects` can be arguments of the constructor of `SVGGraphics`, it is not for internal use only, and `@ignore` is not appropriate.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2433,7 +2433,6 @@ class WorkerTransport {
  * A PDF document and page is built of many objects. E.g. there are objects for
  * fonts, images, rendering code, etc. These objects may get processed inside of
  * a worker. This class implements some basic methods to manage these objects.
- * @ignore
  */
 class PDFObjects {
   constructor() {


Snuffleupagus (changes requested):
Similar to your other PR, the commit message ought to include JSDoc as well.

tamuratak:
I have edited the commit message as requested.

timvandermeij:
Is `SVGGraphics` part of the public API, since it's an experiment feature? I don't think we should expose the `PDFObjects` class given that it's internal to the library.

timvandermeij:
Thinking about this some more, the `PDFObjects` and `SVGGraphics` classes should not be part of the public API, so this change should not be necessary.

PR #11213, opened 2019-10-06
Update the `gulp importl10n` command to fetch the active language codes

Snuffleupagus:
- Update the `gulp importl10n` command to fetch the active language codes

   Rather than having to manually maintain a static list of language codes, it's much easier to simply fetch the active ones from `mozilla-central` instead.

   As part of this the code in `external/importL10n/locales.js` was modernized slightly, by using Promises/async functions to get rid of a bunch of annoying callbacks (which shouldn't be a problem for reasonably modern Node.js versions).

 - Update l10n files

 - Have `gulp importl10n` list locales which may be candidates for removal

   Rather than having the script remove locales automatically, which seems like a heavy-handed approach at least initially, listing these for manual checking seems nice though.

*Smaller diff with https://github.com/mozilla/pdf.js/pull/11213/files?w=1*

### DIFF
--- a/external/importL10n/locales.js
+++ b/external/importL10n/locales.js
@@ -19,27 +19,41 @@ var fs = require('fs');
 var https = require('https');
 var path = require('path');
 
-// Defines all languages that have a translation at mozilla-central.
+// Fetches all languages that have an *active* translation in mozilla-central.
 // This is used in gulpfile.js for the `importl10n` command.
-var langCodes = [
-  'ach', 'af', 'ak', 'an', 'ar', 'ast', 'az', 'be', 'bg', 'bn-BD', 'bn-IN',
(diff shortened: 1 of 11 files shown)


timvandermeij (approved):

timvandermeij:
Good to see this script getting some improvements!

Snuffleupagus:
As always, thanks for reviewing/merging this (and other PRs)!

With these changes, the following locales are listed as candidates for removal:
```
Consider removing the following unmaintained locales:
ak, bn-BD, bn-IN, csb, hto, kok, ks, ku, lg, meh, mn, nso, rw, sah, sat, sw, ta-LK, tn, tsz, zam, zu
```
It looks like most of the haven't been updated for years, but I didn't look too closely and thus wasn't sure if outright removal was the right hing to do in *this* PR.

timvandermeij:
I like that it notifies us, but allows us to take action if required. I do think removing those locales would be fine for a follow-up patch.

PR #11218, opened 2019-10-08
Attempt to fallback to a default font, for non-available ones, in `PartialEvaluator.loadFont`

### LINKED ISSUE #11068: Viewer does not render data from a filled pdf (all data blank)
Try rendering the attached document in the demo viewer: all fields appear blank.

[1 pager malfunctioning file.pdf](https://github.com/mozilla/pdf.js/files/3502038/1.pager.malfunctioning.file.pdf)

This issue is critical for all users who need to use fillable pdf documents.

### LINKED ISSUE #11162: PDF with data on top of it - all data is lost when opening via pdf.js

Snuffleupagus:
Fixes #11068 
Fixes #11162

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -711,21 +711,35 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       var fontRef, xref = this.xref;
       if (font) { // Loading by ref.
         if (!isRef(font)) {
-          throw new Error('The "font" object should be a reference.');
+          throw new FormatError('The "font" object should be a reference.');
         }
         fontRef = font;
       } else { // Loading by name.
         var fontRes = resources.get('Font');
         if (fontRes) {
           fontRef = fontRes.getRaw(fontName);
-        } else {
-          warn('fontRes not available');
-          return errorFont();
         }
       }
       if (!fontRef) {
-        warn('fontRef not available');
-        return errorFont();
+        const partialMsg =
+          `Font "${fontName || (font && font.toString())}" is not available`;
+
+        if (!this.options.ignoreErrors && !this.parsingType3Font) {
+          warn(`${partialMsg}.`);
+          return errorFont();
+        }
+        // Font not found -- sending unsupported feature notification.
+        this.handler.send('UnsupportedFeature',
+                          { featureId: UNSUPPORTED_FEATURES.font, });
+        warn(`${partialMsg} -- attempting to fallback to a default font.`);
+
+        // Falling back to a default font to avoid completely broken rendering,
+        // but note that there're no guarantees that things will look "correct".
+        fontRef = new Dict();
+        fontRef.set('BaseFont', Name.get('PDFJS-FallbackFont'));
+        fontRef.set('Type', Name.get('FallbackType'));
+        fontRef.set('Subtype', Name.get('FallbackType'));
+        fontRef.set('Encoding', Name.get('WinAnsiEncoding'));
       }
 
       if (this.fontCache.has(fontRef)) {


src/core/evaluator.js
```diff
@@ -708,24 +708,39 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
+        // Font not found -- sending unsupported feature notification.
+        this.handler.send('UnsupportedFeature',
+                          { featureId: UNSUPPORTED_FEATURES.font, });
+
+        // Falling back to a default font to avoid completely broken rendering,
+        // but note that there're no guarantees that things will look "correct".
+        fontRef = new Dict();
+        fontRef.set('BaseFont', Name.get('PDFJS-FallbackFont'));
+        fontRef.set('Type', Name.get('Type1'));
+        fontRef.set('Subtype', Name.get('Type1'));
```
Snuffleupagus: Assuming that it wouldn't cause problems elsewhere, using some kind of dummy Type/Subtype would probably be appropriate since you'd otherwise risk skewing font telemetry results.

Snuffleupagus:
@timvandermeij Please note that I'm not entirely sure about the way forward here!
I wasn't really able to make sense of https://github.com/mozilla/pdf.js/pull/11191#issuecomment-539368927, and there wasn't any requests for clarification on any specific comment, and thus I decided to (quickly) try and implement my idea just to see if it'd actually work as expected.

src/core/evaluator.js
```diff
@@ -708,24 +708,39 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
...
       } else { // Loading by name.
-        var fontRes = resources.get('Font');
+        fontRes = resources.get('Font');
         if (fontRes) {
           fontRef = fontRes.getRaw(fontName);
         } else {
-          warn('fontRes not available');
-          return errorFont();
+          warn(`fontRes "${fontName}" not available.`);
+          // Attempting to fallback, see the code-path just below...
```
timvandermeij: I think we should put this in the warning here/below instead (or both) so it's clear from the console that this is a recovery path. We did this before for some other recovery messages.

timvandermeij:
The patch itself looks clear to me and seeing the reference images this does seem to address the issue. I have one comment above. Looking at the other patch after the updates, I don't see too many differences anymore.

Snuffleupagus:
> Looking at the other patch after the updates, I don't see too many differences anymore.

Since my intention wasn't to step all over the other patch, and I threw this one together *before* seeing the updated version, I'm just not sure how we should proceed here!?

However, the other one still has a couple of issues with the implementation (it's not as complete as it ought to be, and it also abuses `Dict`/`Name`-structures a bit).

timvandermeij:
I would say we can continue with this more complete patch, but use https://help.github.com/en/articles/creating-a-commit-with-multiple-authors to credit both authors for the work.

src/core/evaluator.js
Snuffleupagus: Good point; I've removed the (frankly) unnecessary warning duplication, and also added different warning messages depending on whether the fallback/Error paths are taken.

timvandermeij (approved):

timvandermeij:
Nice work!

/botio makeref

PR #11219, opened 2019-10-08
[api-minor] Replace all `deprecated` calls with throwing of actual `Error`s

Snuffleupagus:
All of these methods have been marked as `deprecated` in *three* releases now, and I'd thus like to (slowly) move towards complete removal.

However rather than just removing the methods right away, which would cause somewhat cryptic failures, this patch tries to implement a hopefully reasonable middle ground by throwing `Error`s with (essentially) the same information as the previous warnings.

While the previous `deprecated` messages could perhaps be seen as optional, with these changes API consumers will now be forced to actually migrate their code.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -23,9 +23,8 @@ import {
   unreachable, warn
 } from '../shared/util';
 import {
-  deprecated, DOMCanvasFactory, DOMCMapReaderFactory, DummyStatTimer,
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Indeed time to get rid of this. Thanks!

Issue #8305, opened 2017-04-17
What should be the value for `workerSrc`?


yinso:
Link to PDF file (or attach file here):

Configuration:
- Web browser and its version: Chrome 57.0.2987.133
- Operating system and its version: OSX 10.11.6
- PDF.js version: 1.7.365
- Is an extension: bundled via webpack 2.2.1

I tried to follow [the documentation for webpack](https://github.com/mozilla/pdf.js/wiki/Setup-pdf.js-in-a-website#with-webpack) but it's not working (keep getting setting up fake worker), probably due to me not understanding what exactly should go into `workerSrc`.

Is `workerSrc`:

* a URL that'll get resolved by the browser? e.g. `PDFJS.workerSrc = '/pdf.worker.js';`
* a Node require spec that'll get resolved and bundled into the same file via webpack & node?
   e.g. `PDFJS.workerSrc = 'pdfjs-dist/build/pdf.worker.js';`
* a relative path that'll get resolved and bundled into the same file via webpack & node?
  e.g. `PDFJS.workerSrc = '../../node_modules/pdfjs-dist/build/pdf.worker.js';`
* something else?

Thanks for any pointers.

evadne:
@yinso should be something the browser can hit. You may not want to use a relative path since it may break if anything else in your stack makes asset digests. It is best if you render that from your web framework.

timvandermeij:
Moreover, you can refer to the Webpack example at https://github.com/mozilla/pdf.js/tree/master/examples/webpack. Closing as answered.

yinso:
Thanks guys for the help - appreciated.

luistak:
Try this:



      const pdfjs = await import('pdfjs-dist/build/pdf');
      const pdfjsWorker = await import('pdfjs-dist/build/pdf.worker.entry');

      pdfjs.GlobalWorkerOptions.workerSrc = pdfjsWorker;

      ...

llamamoray:
I think @luistak solution should be added to the documentation. It's the cleanest way to load the worker in using webpack/create-react-app

Mageenz:
It's the cleanest way to load the worker in using webpack/create-react-app

gremo:
@luistak solution is good but still getting the error in console: **Warning: Setting up fake worker.**

Any help?

luistak:
> @luistak solution is good but still getting the error in console: **Warning: Setting up fake worker.**
> 
> Any help?

[...]

Issue #8204, opened 2017-03-28
No PDFJS.workerSrc specified error on using pdfjs-dist package from npm 


the-fallen:
![test](https://cloud.githubusercontent.com/assets/11062448/24399416/4f1b9b38-13ca-11e7-8802-4e7316c69cb2.jpg)

```
function usage(blob) {
    console.log('usage called!');
    var fileReader = new FileReader();
    fileReader.onload = function (blob) {
        require('pdfjs-dist');
        var fs = require('fs');
        PDFJS.getDocument(blob.target.result).then(function (pdfDocument) {
        console.log('Number of pages: ' + pdfDocument.numPages);
        });
    }
    fileReader.readAsArrayBuffer(blob);
}
```

implemented as shown in the example for node, but throws error for not specifying workingSrc. 
I'm using content scripts for a chrome extension which disallow document.currentscript as fcfort commented in [this](https://github.com/mozilla/pdf.js/issues/5344) issue

the-fallen:
As fcfort commented, it works when i use

`PDFJS.workerSrc = chrome.extension.getURL("libs/pdf.worker.js");
`
before calling PDFJS.document and put my pdf.worker.js in root folder(root-directory/libs/pdf.worker.js).
But this isn't a clean solution. I hope there's a fix for this.

Arsh-kalsi:
Go to pdf.js 

search function getWorkerSrc()

replace this lines

pdfjsFilePath = "YOUR_PATH_TO_JS_FILE/pdf.worker.js";
    if (pdfjsFilePath) {
      return pdfjsFilePath;
    }

Snuffleupagus:
> But this isn't a clean solution. I hope there's a fix for this.

That's not a work-around, but actually the *correct* solution!
You should *always* set the `workerSrc` option before calling `getDocument`, which is even [mentioned in the docs](https://github.com/mozilla/pdf.js/blob/c33bf800cc87941cc681c3c54279ab1b9745650d/src/display/worker_options.js#L30-L31), to ensure that the worker file will be loaded correctly.

Please keep in mind that even though we attempt to [obtain a fallback path](https://github.com/mozilla/pdf.js/blob/c33bf800cc87941cc681c3c54279ab1b9745650d/src/display/api.js#L1304-L1311), there's no simply no guarantee that it will be successful if the `workerSrc` option wasn't explicitly set by the user.

Closing as answered.

luistak:
You should try this:



[...]

PR #11220, opened 2019-10-09
Added create react app example with typescript and basic usage

luistak:
Following this issues:

https://github.com/mozilla/pdf.js/issues/8204#issuecomment-535233292

and 

https://github.com/mozilla/pdf.js/issues/8305#issuecomment-536490041

i have added a simple create react app using pdfjs


Snuffleupagus:
> i have added a simple create react app using pdfjs

Adding close to 20000 lines of code doesn't really seem like a "simple" example to me, is all of this actually necessary!?

Snuffleupagus:
> i have added a simple create react app using pdfjs

Adding close to 20000 lines of code doesn't really seem like a "simple" example to me, is all of this actually necessary!?

One thing that needs to be seriously considered here is to *not* commit the `yarn.lock` files (especially the top-level one), since this is only an example and not shipping code.

Snuffleupagus:
> i have added a simple create react app using pdfjs

Adding close to 20000 lines of code doesn't really seem like a "simple" example to me, is all of this actually necessary!?

One thing that needs to be seriously considered here is to *not* commit the `yarn.lock` files (especially the top-level one since we're using `package-lock.json` in this library), since this is only an example and not shipping code.

luistak:
> n



> > i have added a simple create react app using pdfjs
> 
> Adding close to 20000 lines of code doesn't really seem like a "simple" example to me, is all of this actually necessary!?
> 
> One thing that needs to be seriously considered here is to _not_ commit the `yarn.lock` files (especially the top-level one since we're using `package-lock.json` in this library), since this is only an example and not shipping code.

I apologize for committing the yarn lock files in this issue, already fixed it.

I referenced the PR as a simple react app, because i just runned the `npx create-react-app my-app` and added a simple usage of PDFjs with an app created with `[create-react-app](https://github.com/facebook/create-react-app)`

package.json
```diff
@@ -15,6 +15,7 @@
     "escodegen": "^1.12.0",
     "eslint": "^6.4.0",
     "eslint-config-prettier": "^6.3.0",
+    "eslint-config-react-app": "^5.0.2",
```
Snuffleupagus: Is it necessary to add this "globally", or would it suffice to add it to the example-specific page `package.json` file?
(If so, the `package-lock.json` file can be reverted as well.)

Snuffleupagus (changes requested):
Thanks for removing the `yarn.lock` files!
Please also see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

---

Are the `*.png` files completely necessary here?

package.json
luistak: It was needed before i add typescript to the project, after that it's really not necessary, already removed it

luistak: I remembred the reason of the config, it's because of the Ci failure, should i add it again?

image

Snuffleupagus: Would it work to add it to `examples/create-react-app/package.json` file, rather than the top-level `package.json` file?

luistak: Didn't work, i tried it, only in the top-level :(

examples/create-react-app/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
timvandermeij: Should we give this a better name, like `PDF.js rendering example for React`, and a short name?

examples/create-react-app/src/app.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+import Pdf from './pdf';
+
+export default function App () {
+  return (
+    <main>
+      <h1>Simple pdf render example</h1>
```
timvandermeij: `PDF.js rendering example for React` also here maybe?

examples/create-react-app/src/index.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import App from './app';
+
+import './index.css';
+
+ReactDOM.render(<App />, document.getElementById('root'));
+
```
timvandermeij: Let's remove these two blank lines.

examples/create-react-app/src/pdf.tsx
```diff
@@ -0,0 +1,62 @@
...
+    />
+  );
+}
+
+PdfComponent.propTypes = {
+  url: PropTypes.string
+};
+
+PdfComponent.defaultProps = {
+  url: 'https://raw.githubusercontent.com/mozilla/pdf.js/ba2edeae/examples/learning/helloworld.pdf'
```
timvandermeij: Can we use `https://github.com/mozilla/pdf.js/blob/master/examples/learning/helloworld.pdf` here. Alternatively, perhaps better, would be to copy the file into the example directory and use the local file instead, to make the example not depend on GitHub at all.

package.json
timvandermeij: I also don't think we should have this here because the example should be self-contained. If need be, we can exclude the directory from ESLint, but I would suggest trying to remove the following lines from your `package.json` first:

```
  "eslintConfig": {
    "extends": "react-app"
  },
```

Perhaps that works?

timvandermeij (commented):
In general this looks good and I'm glad that useful examples are being contributed!

examples/create-react-app/src/index.tsx
luistak: Done :)

examples/create-react-app/src/pdf.tsx
luistak: Done :)

examples/create-react-app/src/app.tsx
luistak: For sure! Already changed it

examples/create-react-app/public/manifest.json
luistak: Done :)

package.json
luistak: i'll try it

luistak: Worked! Thanks @timvandermeij!

timvandermeij:
Thank you for providing this example and making it self-contained!

PR #11222, opened 2019-10-09
Remove locales which are unmaintained and/or not shipping in Nightly (PR 11213 follow-up)

Snuffleupagus:
With the sole exception of `meh`[1], none of the locales removed here have even been updated since the last change was made to the default `en-US` locale.

Please note that if/when these locales would start shipping in Nightly again, they will automatically be re-added in PDF.js as well with the `gulp importl10n` command.

---
[1] The current translation is also somewhat incomplete, to put it mildly.


timvandermeij (approved):

timvandermeij:

Issue #6243, opened 2015-07-21
PDF renders incorrectly with error


sethnk:
Tested in the latest build (1.1.326).

PDF file:  http://seth.redirectme.net/med/pdf/55ae64d8a0db5.pdf

The following PDF renders incorrectly and throws the following error:

Message: Invalid number: ¬
Stack: Error: Invalid number: ¬
    at error (chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:247:9)
    at Object.Lexer_getNumber [as getNumber](chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:30760:9)
    at Object.Lexer_getObj [as getObj](chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:31001:23)
    at Object.Parser_shift [as shift](chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:30142:32)
    at Object.Parser_makeStream [as makeStream](chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:30519:12)
    at Object.Parser_getObj [as getObj](chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:30187:28)
    at Object.XRef_fetchUncompressed [as fetchUncompressed](chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:4289:28)
    at Object.XRef_fetch [as fetch](chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:4246:26)
    at Object.XRef_fetchIfRef [as fetchIfRef](chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:4227:19)
    at Object.Dict_get [as get](chrome-extension://oemmndcbldboiebfnladdacbdfmadadm/content/build/pdf.worker.js:3120:28)

timvandermeij:
I get `bad Xref entry` and `"Error: Illegal character: 41`, but the cause is the same: the parser is probably ending up in the middle of stream because the Xref stream header cannot be read.

Snuffleupagus:
[...]

PR #5910, opened 2015-04-04
Fix error reading concatenated pdfs

jordan-thoms:
This fixes issue #5909

Issue #5909, opened 2015-04-04
Handle some cases where two PDFs were concatenated together.


jordan-thoms:
Hi!

We've had a report with a PDF file which PDF.js could not open, but other PDF readers (e.g. Preview, Adobe Reader), could open without issues.

In investigating this, it turns out the PDF file is actually two different PDFs which were concatenated one after the other into a single file. Obviously not a valid PDF, but given that other PDF readers will show one of the PDFs in this case, we probably should too (*sigh*).

There was some previous discussion of this on issue #5255.

I've found the issue preventing us from showing one of the PDFs - it's because in the XRef indexObjects method in obj.js, we don't stop reading when we hit a %%EOF marker, and we will happily overwrite things in the entries array. 

Stopping at %%EOF has the side effect of breaking reading linearized PDFs that have a broken startxref value (because they contain one at the beginning too), so as an alternative fix I have implemented a check where we don't write into the entries array if a value is already present. I'll be pushing up a PR with that shortly.

I've uploaded the file here: https://www.dropbox.com/s/pvcuea8fmg1nc3c/21063565.pdf?dl=0

Thanks
Jordan

timvandermeij:
Fixed by #5910.

PR #11231, opened 2019-10-11
Allow over-writing entries, in `XRef.indexObjects`, only when the generation number matches (issues 11230, 11139, 9552, 9129, 7303)

### LINKED ISSUE #7303: dict.get is not a function
There is a problem rendering this PDF: 
[ws_protectyourwork_e.pdf](https://github.com/mozilla/pdf.js/files/254710/ws_protectyourwork_e.pdf)

Using the standard web viewer example (build/generic/web/viewer.html) it runs into an exception, showing the error:

PDF.js v1.5.232 (build: a682cce)
Message: dict.get is not a function

screen shot 2016-05-09 at 11 06 41

This happens in Chrome, Firefox, Safari on the Mac - haven't tried any others. Message on Safari is: dict.get is not a function. (In 'dict.get('Filter')', 'dict.get' is undefined)

I hope this helps. Not sure if there is a problem with the PDF but I guess it would ideally catch the error anyway. PDF seems to display OK in other software.

Thanks,

Dan

### LINKED ISSUE #9129: Edited pages rendered as if not changed
https://zenodo.org/record/1048152#.WgrUmd9fjeQ
or
https://waset.org/publications/10006645/red

Configuration:
mozilla ( under linux/windows), 
- PDF.js version:
- Is a browser extension:

0. Our editors format papers, and our system using zend php, adding necessary information to header, footer, side notes.
1. Open the document, there is no header, footer, side notes 

screencapture-waset-org-publications-10006645-r-1510672765952

screenshot-2017-11-14 microsoft word - 17ae030135 3 - red

expected:
https://waset.org/publication/Red/10006645
not working  :
https://zenodo.org/record/1048152#.WgrUmd9fjeQ

### LINKED ISSUE #9552: pdf.js seems to be ignoring annotations added using Preview.app
https://github.com/Napoleon-BlownApart/Firefox_pdf.js_Problem

Configuration:
Nightly 49.0.1  (Last version that does not require signed addons)
MacOS 10.13.3
- PDF.js version:
*Standard version shipped with the build mentioned above.*
- Is a browser extension:

1. Open PDF using Firefox (Nightly in my case) and two yellow annotations will be visible in the 2nd column of text on page 1.
2. Open PDF using Safari, or preview, (possibly other readers including Adobe), and four (4) other additional annotations will now be visible, also on the 1st page;
2a)  A "Note" in the top left,
2b) A comment "This is from Preview" near the top left,
2c) The first five (5) words of the Abstract will be highlighted in red, and
2d) The first five (5) words of the Introduction will be highlighed in orange.

All annotations are rendered.
screen shot 2018-03-13 at 1 00 05 am

Some annotations are not visible.
screen shot 2018-03-13 at 12 59 55 am

### LINKED ISSUE #11139: Missing first page
[FR Cekada SedesUnCum.pdf](https://github.com/mozilla/pdf.js/files/3602024/FR.Cekada.SedesUnCum.pdf)

Configuration:
- Web browser and its version: Firefox 69 & Chrome 77
- Operating system and its version: Ubuntu 16.04.6 LTS
- PDF.js version: 2.3.160
- Is a browser extension: No

1. Open the attached PDF.
2. The first page is missing. The page displayed as first is actually the second one.

### LINKED ISSUE #11230: Majority of Document is Cut Off
[Y._Tsividis_A_First_Lab_in_Circuits_and_Electronics.pdf](https://github.com/mozilla/pdf.js/files/3717821/Y._Tsividis_A_First_Lab_in_Circuits_and_Electronics.pdf)

Configuration:
- Web browser and its version: Google Chrome 77.0.3865.120
- Operating system and its version: Windows 10 1903 (Build 18362.418)
- PDF.js version: 2.3.164
- Is a browser extension: Yes

1. Open attached document
2. Scroll to bottom

In all other PDF readers, such as the one in Microsoft Edge, 139 pages are rendered
image

Only 15 pages are rendered
image

JS Console Messages:
```
pdf.worker.js:1710 Warning: Indexing all PDF objects
viewer.js:1502 PDF d216c1e858407c419356a28ea6096b93 [1.6 Acrobat Distiller 7.0 (Windows) / PScript5.dll Version 5.2] (PDF.js: 2.3.164 [WebGL])
pdf.worker.js:1710 Warning: Invalid absolute docBaseUrl: "file:///C:/Users/nkizz/Desktop/Textbooks/Y._Tsividis_A_First_Lab_in_Circuits_and_Electronics.pdf".
```

Snuffleupagus:
**TODO:**
 - [x] Manually check open/closed issues tagged with `4-corrupted-pdf` such that there's no obvious regressions.
*I've done a quick pass already, but additional help would be very much appreciated here.*
 - [x] Convert (relevant) pre-existing `load` tests to `eq` tests instead, to help avoid unnecessary regressions. Note that some files will render, but not as intended when compared with e.g. Adobe Reader, depending on which objects are chosen (see e.g. issues with incorrect total number of pages).
 - [x] Add as many new `eq` tests as possible, based on the list above.

---

This patch is making me somewhat worried about future regressions, since it's certainly easy to imagine this completely breaking certain kinds of corrupt/edited PDF documents while fixing others.[1]

Obviously it passes all existing reference tests (and even improves one), however compared to many other patches there's no telling how much it could break. *Edit:* Obviously any breakage would be limited to *corrupt* PDF documents, since valid ones does not enter this method at all. *Edit2:* Also, this patch should mostly affect the "multiple documents in one PDF file" case, but not necessarily PDF documents with just a broken XRef table.
The only reason that I'm even submitting this patch, is because of the number of open issues that it would address.

Generally speaking though, the best course of action would probably be if `XRef.indexObjects` was re-written to be much more robust (since it currently feels somewhat hand-wavy in parts). E.g. by actually checking/validating more of the objects before committing to them.

Fixes #11230
Fixes #11139
Fixes #9552
Fixes #9129
Fixes #7303

Also partially improves #6243, however parts of it regresses as well. (A couple of pages now renders "better", but one page no longer renders and an image looks worse.)
Generally though it seems that none of the PDF viewers I've tried agree completely on how the document should look, and this might be a case where the document is just too damaged to "fix".

---
[1] Especially given that it's reverting part of PR #5910, however in the case of issue #5909 it seems that other (more recent) changes have actually made that PR redundant.


Snuffleupagus:
*Temporarily dumping some PDF files here, to have a (slightly) better place to point linked test-cases to.*
[issue5909_original.pdf](https://github.com/mozilla/pdf.js/files/3720894/issue5909_original.pdf)

timvandermeij (approved):

timvandermeij:
I think this is a very nice improvement. Even though you're absolutely right about other breakage that may still occur, given that those PDF files are usually quite broken, we have a lot more test coverage now than we had before, so future changes will be less risky. Thanks!

PR #11234, opened 2019-10-12
Convert a number of reference tests, for documents with corrupt XRef tables, from `load` to `eq`

Snuffleupagus:
As part of attempting to fix a number issues containing PDF documents with corrupt XRef tables, I'd like to improve the reference test-coverage slightly *first*.
Obviously this will increase the runtime of the tests a bit, however I'd rather "waste" resources on the bots instead of developer time fixing regressions which could have been avoided.


test/pdfs/.gitignore
```diff
@@ -311,7 +311,6 @@
 !issue5567.pdf
 !issue5701.pdf
 !issue5896.pdf
-!issue5909.pdf
```
timvandermeij: Why is this one changed to a linked test case?

Snuffleupagus: *I should obviously have included this in the commit message, sorry about that!*

Because the *reduced* test-case is just two completely blank pages, and it's thus impossible to know if any future code changes breaks rendering (even though the file manages to load).

Snuffleupagus: Also, note that the PR which added that test was responsible for [the line that I'm trying to partially revert](https://github.com/mozilla/pdf.js/blob/c701ae3f4cd5f1713e09a1b0da0096664ca37f8f/src/core/obj.js#L1418) over in #11231 and I thus wanted to ensure proper test-coverage before making changes.

timvandermeij: Clear, thanks!

timvandermeij (approved):

timvandermeij:
Test coverage is indeed very important for such changes. Looks good!

/botio makeref

PR #11235, opened 2019-10-12
Improve the JSDoc comments

timvandermeij:
This commit series makes our JSDoc comments more consistent and allows us to generate better API documentation and TypeScript definition files.

I have been watching the discussions about both for some time now and figured we should get started with making improvements because I think it would be ideal from a maintenance perspective if we could generate them both instead of maintaining them manually. Therefore, I collected  improvements both from the various discussions and the JSDoc documentation so the resulting output becomes more precise.

Note that this is groundwork for more documentation improvements, so more work definitely needs to be done. Thanks to the various people that discussed this and did proposals for this.

Before:
https://mozilla.github.io/pdf.js/api/draft/PDFDocumentProxy.html

After:
http://54.67.70.0:8877/15dd7d913e2aab9/api/draft/module-pdfjsLib-PDFDocumentProxy.html

(notice that all getters now have types and descriptions)

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1193,7 +1193,8 @@ class PDFPageProxy {
   }
 
   /**
-   * Destroys page object.
+   * Destroys the page object.
+   * @private
    */
   _destroy() {
     this.destroyed = true;
@@ -1236,9 +1237,8 @@ class PDFPageProxy {
   }
 
   /**
-   * For internal use only. Attempts to clean up if rendering is in a state
-   * where that's possible.
-   * @ignore
+   * Attempts to clean up if rendering is in a state where that's possible.
+   * @private
    */
   _tryCleanup(resetStats = false) {
     if (!this.pendingCleanup ||
@@ -1262,8 +1262,7 @@ class PDFPageProxy {
   }
 
   /**
-   * For internal use only.
-   * @ignore
+   * @private
    */
   _startRenderPage(transparency, intent) {
     const intentState = this.intentStates[intent];
@@ -1279,8 +1278,7 @@ class PDFPageProxy {
   }
 
   /**
-   * For internal use only.
-   * @ignore
+   * @private
    */
   _renderPageChunk(operatorListChunk, intentState) {
     // Add the new chunk to the current operator list.
@@ -1302,8 +1300,7 @@ class PDFPageProxy {
   }
 
   /**
-   * For internal use only.
-   * @ignore
+   * @private
    */
   _pumpOperatorList(args) {
     assert(args.intent,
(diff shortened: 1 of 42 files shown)


src/display/api.js
```diff
@@ -15,6 +15,10 @@
 /* globals requirejs, __non_webpack_require__ */
 /* eslint no-var: error */
 
+/**
+ * @module pdfjs
```
Snuffleupagus: Given that the `PDFJS` global was removed, and that we're now using `pdfjsLib` for code running on the main-thread, is that name really appropriate at this point!?

I readily admit not knowing much about JSDoc `@module`, but this was something that stood out to me.

timvandermeij: I checked how we import `pdfjs-dist` in the examples and we indeed use `pdfjsLib` there, so that sounds like a better name for the documentation too. I have changed this.

Snuffleupagus:
Another thing I just noticed is that `PDFWorker` is now listed *twice* under the "Classes" heading.

timvandermeij:
Yes, this is because the closure is called `PDFWorker` and the inner class is called `PDFWorker`. If you click the links, they are in fact different. Looking at this again, maybe we should leave out `PDFWorker` entirely, since normally API users shouldn't need it?

Snuffleupagus:
> Looking at this again, maybe we should leave out `PDFWorker` entirely since I don't think API consumers should need it, or do you think there is a use case for including it?

Given https://github.com/mozilla/pdf.js/blob/4af2755f714f562787ed4302dbc5c599a4056f39/src/display/api.js#L146-L147 you probably need to keep it. However, couldn't you fix things simply by moving the docs to https://github.com/mozilla/pdf.js/blob/4af2755f714f562787ed4302dbc5c599a4056f39/src/display/api.js#L1593-L1596

src/display/api.js
```diff
@@ -1590,9 +1583,15 @@ const PDFWorker = (function PDFWorkerClosure() {
   }
 
   /**
-   * @param {PDFWorkerParameters} params - The worker initialization parameters.
+   * The web worker abstraction controls the instantiation of PDF documents.
```
Snuffleupagus: Please keep the `PDF.js`-part of this line, perhaps as `PDF.js web worker abstraction, which controls the instantiation of PDF documents.`, since this is different from the general concept of the [Web Workers API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API).

Snuffleupagus (approved):
Looks fine by me, modulo one small last comment; thanks!

src/display/api.js
timvandermeij: Done. Thanks!

PR #11237, opened 2019-10-13
Update `src/shared/compatibility.js` to only run with `SKIP_BABEL = false` set

Snuffleupagus:
Rather than specifying certain build targets manually, it seems much more appropriate (and future-proof) to use the `SKIP_BABEL` build target instead.

Also, the patch adds a missing `/* eslint no-var: error */` line since I'm touch the file anyway and no code-changes were necessary for it.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #11241, opened 2019-10-14
[api-minor] Produce non-translated/non-polyfilled builds by default

Snuffleupagus:


gulpfile.js
```diff
@@ -725,6 +726,40 @@ gulp.task('generic', gulp.series('buildnumber', 'default_preferences', 'locale',
   ]);
 }));
 
+// Builds the generic production viewer that should be compatible with most
+// older HTML5 browsers.
+gulp.task('generic-es5', gulp.series('buildnumber', 'default_preferences',
+                                     'locale', function() {
```
timvandermeij: Here, and similar below, we could extract the `function` and give it `skipBabel` parameter so we don't have to duplicate it (unless I'm overlooking something, they seem equal aside from the `SKIP_BABEL` parameter). Something like (untested):

```javascript
function createGenericViewer(skipBabel) {
   ...
   return merge([...]);
}

gulp.task('generic', gulp.series('buildnumber', 'default_preferences', 'locale', function() {
  createGenericViewer(/* skipBabel */ true);
}));

gulp.task('generic-es5', gulp.series('buildnumber', 'default_preferences', 'locale', function() {
  createGenericViewer(/* skipBabel */ false);
}));

**Comment by Snuffleupagus** [comment]
> Viewer: http://54.67.70.0:8877/ee02a2e381b03ca/web/viewer.html

This works in e.g. Firefox Nightly 71, but not in e.g. IE 11.

However, the modified link http://54.67.70.0:8877/ee02a2e381b03ca/es5/web/viewer.html works in both, as intended.

File: gulpfile.js (hunk 1 above)
Comment (Snuffleupagus) [reply gulpfile.js]: > [...] they seem equal aside from the `SKIP_BABEL` parameter

For what it's worth, there's also a difference in output directories...

While you're obviously correct that it *should* be possible to reduce the size of the `gulpfile.js` changes, I'd very much like to get an OK on the general idea of this PR before spending too much time on it.

In no particular order, there's a couple of things that we'd need to consider:
 - Are we OK with changing the default to producing non-translated/non-polyfilled builds by default, and instead having to *explicitly* specify when ES5-compatible builds are wanted?
   Personally I'm hoping this will be seen as a step in the right direction, since it might also allow further modernizing of the build system.
 - ES5-compatible `generic` builds are obviously still provided, but is `api-minor` enough since these changes may require that implementers manually change what files they depend on?
   Also, are we OK with only providing pre-built ES5-compatiable files for the `generic` build target, since this simplifies things considerably?
 - What about all of the `examples`?
   Personally I'd like to avoid changing them wholesale to use the ES5-compatible files, but how should we document this (inline comments, Wiki entry, and/or something else; and will people even bother to read it)?

Comment (timvandermeij) [reply gulpfile.js]: I'm personally also in favor of preferring ES6 builds over ES5 builds. However, I'm not sure if `[api-minor]` is the right classification for this change since it's quite a behavior change; it feels more like `[api-major]` to me.

I'm thinking that to make this work for `[api-minor]` we can do the reverse, namely keep the default at ES5 but also provide ES6 builds. This is already a step in the right direction because now we don't really provide ES6 builds, so it will allow people to start using it and provide us with feedback. Moreover, we and users don't need to adjust anything immediately. For the next major version, we can then swap the builds, making ES6 the default and ES5 the alternative option. In the major version after that, we can then hopefully drop the ES5-only build altogether.

What do you think? Also /cc @yurydelendik @brendandahl for opinions on this.

Comment (Snuffleupagus) [reply gulpfile.js]: > I'm thinking that to make this work for `[api-minor]` we can do the reverse, namely keep the default at ES5 but also provide ES6 builds.

Honestly, this feels like *way* more work/complication compared to the current approach (and certainly not something I'd bother doing, unless the users complaining about there not being any non-translated builds actually compensating me for my time :-).

Even trying to do something like this would also make e.g. the `pdfjs-dist` repository really big and complicated, not to mention *even slower* to generate, since you'd probably need to essentially duplicate all of it's code which is precisely what I actively tried to avoid here.
In that case I'd personally much rather just immediately increment the `major` version, and keep the current approach of modern builds with a few exceptions, and then be done with it since this simplifies things considerably!

Furthermore, note that the non-translated builds aren't actually [ECMAScript 2015](https://en.wikipedia.org/wiki/ECMAScript#6th_Edition_-_ECMAScript_2015) (i.e. older classification ES6) but rather at least [ECMAScript 2018](https://en.wikipedia.org/wiki/ECMAScript#9th_Edition_-_ECMAScript_2018) (i.e. older classification ES9). Hence labelling something as ES6 isn't even remotely correct

Comment (Snuffleupagus) [reply gulpfile.js]: *Sorry if the above seemed a bit grumpy!*

Some additional thoughts:

> [...] so it will allow people to start using it and provide us with feedback.

Assuming that a modern, and up-to-date, standards-compliant browser is being used I really cannot imagine there being any problems with using a non-translated/non-polyfilled build (based on the fact that the PDF Viewer in Firefox has been shipping this way for years).

Also, simply incrementing the `major` version should definitely allow to move forward and "fix" things like e.g. issue #11225 
Finally, considering that the *first* official release from the `2.0` branch was released a year ago, it doesn't seem so bad to me to increase that now (especially for these kind of build improvements).

Comment (timvandermeij) [reply gulpfile.js]: Incrementing the major version number for the reasons you mentioned sounds like a good idea, so perhaps that's indeed an easier approach. I'll think about this plan a bit more.

**Comment by Snuffleupagus** [comment]
Part of this work probably ought to *also* include some sort of clean-up/re-factoring/simplification of the fake worker loading code, since the current state feels like a mess (and given the complaints about it):
https://github.com/mozilla/pdf.js/blob/bbd2386bd9b025d2ea70cd43eec842e674f17437/src/display/api.js#L49-L106

**Comment by Snuffleupagus** [comment]
> [...] clean-up/re-factoring/simplification of the fake worker loading code,

I've tried to strip this down to its bare essentials, see https://github.com/Snuffleupagus/pdf.js/tree/fakeWorkerLoader, with the intention of only supporting Node.js *and* the `LIB` build target out-of-the-box.
However, doing so causes some issues with the [webpack example](https://github.com/mozilla/pdf.js/tree/master/examples/webpack); while it still runs successfully there's some `Critical dependency: the request of a dependency is an expression` errors during the build step. (You'd expect that most users will rely on pre-built files from e.g. `pdfjs-dist`, so I'm not sure how much this would actually matter...)

All in all, I'm nowhere near knowledgeable enough about various module loaders to be able to sort out issues like these unfortunately. Hence I'm not really sure about the way forward here (and given that these sort of module loader changes brings no benefit to me *personally*, I'm finding it really hard to justify spending even more of my spare time on it).

**Comment by Snuffleupagus** [comment]
> [...] with the intention of only supporting Node.js _and_ the `LIB` build target out-of-the-box.

Although I suppose that you *could* just remove all non-browser "fakeWorkerLoader" fallback code and simply update the Node.js examples and `gulp lib` output to *manually* load the `pdf.worker.js` file, but I'm not sure if that would be considered an overall bad move!?

**Comment by timvandermeij** [comment]
I think the problem is that we're all not really knowledgeable about module loaders and we also don't have contributors maintaining this. Looking at #9580 perhaps it's not even desired to have support for Webpack in this repository? I think we should go back to basics a bit: support only the use cases that are relevant and most often used, and leave the rest to third-party implementations or contributors if they really wish to have this support in the repository.

@yurydelendik @brendandahl What do you think?

**Comment by Snuffleupagus** [comment]
> However, doing so causes some issues with the [webpack example](https://github.com/mozilla/pdf.js/tree/master/examples/webpack); while it still runs successfully there's some `Critical dependency: the request of a dependency is an expression` errors during the build step.

This is, as far some people are concerned (me included), a bug in Webpack since it's currently *impossible to force* Webpack to leave a `require` statement alone in a general way; please see https://github.com/webpack/webpack/issues/8826.

The only "solution" that currently seem to exist, as gross at it would be, is probably to use `eval('require')` since Webpack cannot parse that...

**Comment by timvandermeij** [comment]
Ugh. If it's required to be able to move on, I guess it's better than doing nothing, but it's also not great either. Don't we have the `__non_webpack_require__` thing to tell Webpack to not touch the `require` statement though? I think we use it in a few places just for this purposes as far as I'm aware, although I must admit not to know much about this, but maybe it helps: https://github.com/mozilla/pdf.js/search?q=__non_webpack_require__&unscoped_q=__non_webpack_require__

**Comment by Snuffleupagus** [comment]
> Don't we have the `__non_webpack_require__` thing to tell Webpack to not touch the `require` statement though?

Yes, but in the *built* files that's replaced with a standard `require` (as expected) and that causes problems e.g. with https://github.com/mozilla/pdf.js/tree/master/examples/webpack in the form of `Critical dependency: ...` warnings and in third-party bundling such as e.g. issue #10253.

**Comment by Snuffleupagus** [comment]
I do think that we should consider moving forward with this PR, in one way or another, since the only realistic way to address repeated issues/complaints such as e.g. issue #11036 would be to provide builds without *any* polyfills and simply let users handle that themselves.
(There's also been a fair number of other clean-up/simplification, related to fake worker loaders and Webpack recently.)

As long as we make a *big note*, on the relevant release page, about the need to pick the right kind of build I'm not sure if making something like this an `api-minor` change is completely out of the question!?
Note also the updated page "Getting Started" page, see http://54.67.70.0:8877/62b81ab5067f4ce/getting_started/#download

**Comment by timvandermeij** [comment]
Having thought a bit more about this I'm also not opposed to making this an `[api-minor]` change anymore. I do wonder if it's possible to tweak the download page a bit by not showing the buttons in the ES5-compatible section for older releases since those links are now broken on the page you linked.

**Comment by Snuffleupagus** [comment]
> I do wonder if it's possible to tweak the download page a bit by not showing the buttons in the ES5-compatible section for older releases since those links are now broken on the page you linked.

Excellent point; yes that'll be somewhat of a problem, until the releases "catch" up to the new format.

I'm not sure how to fix this *generally*, short of manually hard-coding the links after the relevant releases are done (perhaps that's not so bad though, since it'd be temporary)?

**Comment by Snuffleupagus** [comment]
> I'm not sure how to fix this _generally_, short of manually hard-coding the links after the relevant releases are done (perhaps that's not so bad though, since it'd be temporary)?

The latest version of the patch does something along those line, and will thus require manual fix-ups after the next *two* releases but then "things should just work"<sup>TM</sup>.

File: docs/contents/getting_started/index.md [hunk 2]
```diff
@@ -38,20 +38,33 @@ Before downloading PDF.js please take a moment to understand the different layer
...
+  <div class="col-md-4">
+    <h3>Prebuilt (ES5-compatible)</h3>
+    <p>
+      Includes the generic build of PDF.js and the viewer.
+    </p>
+    <span class="GROUP_CLASS">
+      <a type="button" class="btn btn-primary" href="https://github.com/mozilla/pdf.js/releases/download/vSTABLE_VERSION/pdfjs-STABLE_VERSION-dist.zip">Stable (vSTABLE_VERSION)</a>
+      <a type="button" class="btn btn-warning HIDDEN_CLASS" href="https://github.com/mozilla/pdf.js/releases/download/vBETA_VERSION/pdfjs-BETA_VERSION-dist.zip">Beta (vBETA_VERSION)</a>
+    </span>
+    </span>
```
timvandermeij: Duplicate closing tag?

examples/node/getinfo.js
```diff
@@ -8,7 +8,7 @@
 //
 
 // Run `gulp dist-install` to generate 'pdfjs-dist' npm package files.
-var pdfjsLib = require("pdfjs-dist");
+var pdfjsLib = require("pdfjs-dist/es5/build/pdf.js");
```
timvandermeij: What is the reason for using ES5-compatible builds for the examples? Wouldn't it be better to switch them to the non-ES5-compatible builds so we promote that instead?

gulpfile.js
```diff
@@ -737,8 +738,8 @@ function preprocessHTML(source, defines) {
   return createStringSource(source.substr(i + 1), out);
 }
 
-// Builds the generic production viewer that should be compatible with most
-// modern HTML5 browsers.
+// Builds the generic production viewer that is only compatible with up-to-date
+// HTML5 browsers, which implements modern ECMAScript features.
```
timvandermeij: `which implement`

gulpfile.js
```diff
@@ -786,6 +787,49 @@ gulp.task(
   })
 );
 
+// Builds the generic production viewer that should be compatible with most
```
timvandermeij: I'm wondering if there is a way to merge these two very similar targets in some way. If that's difficult, it's not a must-have, but it would be nice to reduce the size of the file. Same for the target below.

gulpfile.js
```diff
@@ -786,6 +787,49 @@ gulp.task(
...
+      gulp
+        .src(["external/bcmaps/*.bcmap", "external/bcmaps/LICENSE"], {
+          base: "external/bcmaps",
+        })
+        .pipe(gulp.dest(GENERIC_ES5_DIR + "web/cmaps")),
+      preprocessHTML("web/viewer.html", defines).pipe(
+        gulp.dest(GENERIC_ES5_DIR + "web")
+      ),
+      preprocessCSS("web/viewer.css", "generic", defines, true)
+        .pipe(postcss([autoprefixer(AUTOPREFIXER_CONFIG)]))
```
timvandermeij: This appears to be missing the CSS variables/`calc` plugins.

timvandermeij:
I'm fine with the approach taken here. If this is merged, I'll extend the release wiki page so we don't forget to manually update the website.

docs/contents/getting_started/index.md
Snuffleupagus: Yes, copy-paste failure.

examples/node/getinfo.js
Snuffleupagus: Unfortunately Node.js still requires *some* polyfills (e.g. `ReadableStream`), which is why this is done.
However, the other (browser) examples are using the "correct" file, although I probably ought to update all of the existing ones to add a *commented out* line with the es5 version for documentation purposes.

gulpfile.js
Snuffleupagus: I'll try to fix this, since the duplication is mildly annoying.

gulpfile.js
Snuffleupagus: Yes, more rebase errors.
(We may want to look into *not* providing polyfilled CSS for the modern builds, but since those things aren't as easy to polyfill as missing JS features I decided to leave that alone for now.)

examples/node/getinfo.js
Snuffleupagus: Unfortunately Node.js still requires *some* polyfills, e.g. `ReadableStream`, which is why this is done.

Snuffleupagus:
> I'm fine with the approach taken here. If this is merged, I'll extend the release wiki page so we don't forget to manually update the website.

Thanks for the review, hopefully I've managed to address all your comments.

Snuffleupagus:
> I'm fine with the approach taken here. If this is merged, I'll extend the release wiki page so we don't forget to manually update the website.

Thanks for the review, hopefully I've managed to address all your comments.

/botio-linux preview

timvandermeij (approved):

timvandermeij:
Let's do this. I'm seeing this as a good step forward. Thanks!

Snuffleupagus:
@timvandermeij Thanks for good review comments!

I've opened a follow-up PR to clean-up the state on the bots a bit, see https://github.com/mozilla/botio-files-pdfjs/pull/26

PR #11243, opened 2019-10-15
Add a fallback for non-embedded *composite* Verdana fonts (issue 11242)

### LINKED ISSUE #11242: PDF fonts not displaying correctly
[Font Issues - Example PDF.pdf](https://github.com/mozilla/pdf.js/files/3728976/Font.Issues.-.Example.PDF.pdf)

Configuration:
- Web browser and its version: Firefox Quantum (69.0.3 (64-bit))
- Operating system and its version: Windows 8.1 Pro
- PDF.js version: Not sure (latest Firefox browser build as above)
- Is a browser extension: Not sure

1. Open the attached file (Font Issues - Example PDF.pdf) in Chrome or Adobe or any other PDF viewer and the fonts display correctly.
2. Open the same PDF in Firefox and all the fonts display incorrectly.

PDF displayed correctly in Acrobat

All fonts being rendered and displayed incorrectly.
Screenshot when viewed in Firefox

Snuffleupagus:
Obviously this won't look exactly right, but considering that the PDF file doesn't bother embedding non-standard fonts this is the best that we can do here.

Fixes #11242


Snuffleupagus:
Hmm, this seems to cause unwanted movement in existing reference tests; closing for now...

src/core/fonts.js
```diff
@@ -1257,8 +1258,20 @@ var Font = (function FontClosure() {
...
+          map[+charCode] = unicodeCharCode;
         });
+
+        // Attempt to improve the glyph mapping for (some) composite fonts that
+        // appear to lack meaningful ToUnicode data.
+        if (this.composite && this.toUnicode instanceof IdentityToUnicodeMap) {
+          if (/Verdana/i.test(name)) { // Fixes issue11242_reduced.pdf
+            const GlyphMapForStandardFonts = getGlyphMapForStandardFonts();
+            for (charCode in GlyphMapForStandardFonts) {
+              map[+charCode] = GlyphMapForStandardFonts[charCode];
```
timvandermeij: I can't help but wonder why we need the plus character here and in other parts of this patch. As far as I can tell from https://github.com/mozilla/pdf.js/blob/de56518799b869d1cb4d313c7628a8aa956966f8/src/core/standard_fonts.js#L269 and other code, character codes should already be numeric. Is that true or am I missing something here?

Snuffleupagus: When using the `for (key in object)` format the `key` ends up being converted to a string, which is something you don't want when indexing an Array.

timvandermeij: Ah, that's it. Makes sense; thanks!

timvandermeij (approved):

timvandermeij:
Looks good!

/botio makeref

PR #11246, opened 2019-10-15
Tweak the `isIOS` check, in `web/viewer_compatibility`, to attempt to support newer Safari versions (issue 11245)

### LINKED ISSUE #11245: Zoom limit on iOS
[zoom test.pdf](https://github.com/mozilla/pdf.js/files/3729738/zoom.test.pdf)

Configuration:
- Web browser and its version: Safari 12
- Operating system and its version: ipadOS 13.1.2
- PDF.js version: 2.1.266
- Is a browser extension: no

1.  Open attached file in Safari on an iPad running ipadOS 13 in the pdfjs demo viewer https://mozilla.github.io/pdf.js/web/viewer.html
2. click zoom + button to zoom into 220%.
3. The file is no longer displayed instead all white is shown. If a Safari inspector is attached it will write "Canvas area exceeds the maximum limit (width * height > 16777216)." in the console.

What is the expected behavior? (add screenshot) : to be able to zoom more than 200%

What went wrong? (add screenshot): white is shown instead of the content of the file.

It seems that the zoom limit is different for different files. Is there a way to calculate what it is at runtime each file so I avoid showing white instead of the file without having to set a very low zoom limit on iOS devices?

Snuffleupagus:
Obviously userAgent checks aren't that great, since it's very easy to spoof, but it probably doesn't hurt to attempt to extend this (since it's limited to `GENERIC` builds).
Besides, anyone using the default viewer can always set the `maxCanvasPixels` option to a value of their choosing.

Fixes #11245

### DIFF
--- a/web/viewer_compatibility.js
+++ b/web/viewer_compatibility.js
@@ -17,8 +17,14 @@ let compatibilityParams = Object.create(null);
 if (typeof PDFJSDev === 'undefined' || PDFJSDev.test('GENERIC')) {
   const userAgent =
     (typeof navigator !== 'undefined' && navigator.userAgent) || '';
+  const platform =
+    (typeof navigator !== 'undefined' && navigator.platform) || '';
+  const maxTouchPoints =
+    (typeof navigator !== 'undefined' && navigator.maxTouchPoints) || 1;
+
   const isAndroid = /Android/.test(userAgent);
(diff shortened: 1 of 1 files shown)


timvandermeij:
Thank you for improving this!

PR #11247, opened 2019-10-15
Change variable assignment

pedroprado010:
variable locaData first value was never used.


timvandermeij (approved):

timvandermeij:
Thank you! (The reference test failures are known intermittents and not caused by this patch.)

PR #11250, opened 2019-10-16
Silence the Autoprefixer message being printed with all build logs

Snuffleupagus:
The following is printed with every build, which gets kind of annoying when looking at the logs:
```
Replace Autoprefixer browsers option to Browserslist config.
Use browserslist key in package.json or .browserslistrc file.

Using browsers option cause some error. Browserslist config
can be used for Babel, Autoprefixer, postcss-normalize and other tools.

If you really need to use option, rename it to overrideBrowserslist.

Learn more at:
https://github.com/browserslist/browserslist#readme
https://twitter.com/browserslist
```

Given how we're using Autoprefixer, with a number of build-specific configs, simply changing the option name seems like the easiest solution here.

(I'm also adding a couple of newlines at the `autoprefixer` call-sites, to aid readability.)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -74,7 +74,7 @@ var config = JSON.parse(fs.readFileSync(CONFIG_FILE).toString());
 
 // Default Autoprefixer config used for generic, components, minified-pre
 var AUTOPREFIXER_CONFIG = {
-  browsers: [
+  overrideBrowserslist: [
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for cleaning up the logs!

PR #11253, opened 2019-10-16
Avoid handling keyboard shortcuts for `contentEditable` DOM elements (issue 7156)

### LINKED ISSUE #7156: Shortcuts also work in rich text editors
Configuration:
- Web browser and its version: Chrome latest
- Operating system and its version: Win10
- PDF.js version: there is no version number in viewer.js :(
- Is an extension: Yes. I've added an rich text editor to make annotations to the PDF

1. Adding RTE to viewer.html
2. Pressing 'r', 'k', 'p', 'j', 'n' or any other shortcut without Shift or Control in RTE.

The pressed key should be only inserted in to the RTE.

What went wrong?
The shortcut will be also performed in pdf.js (rotating page while pressing 'r' ...).
So far this is the correct behaviour. But there is already code for exeptions starting in line 7823 in viewer.js. I.e. for input, textarea and select fields. But this won't work with div based rich text editors.
Would it make sense to check also "Element.attributes.contenteditable" so that the shortcuts won't get handled if the focus is in an editable element?

Snuffleupagus:
The issue has been open for years now, and has even been marked with `5-good-beginner-bug` for *months*, without any movement.
Considering just how simple the suggested solution is, I'm submitting this patch just to close out a long-standing issue.

Fixes #7156

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2283,7 +2283,8 @@ function webViewerKeyDown(evt) {
   let curElementTagName = curElement && curElement.tagName.toUpperCase();
   if (curElementTagName === 'INPUT' ||
       curElementTagName === 'TEXTAREA' ||
-      curElementTagName === 'SELECT') {
+      curElementTagName === 'SELECT' ||
+      (curElement && curElement.isContentEditable)) {
     // Make sure that the secondary toolbar is closed when Escape is pressed.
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #11256, opened 2019-10-17
[Firefox] Remove unused addon `l10n` files (PR 9566 follow-up)

Snuffleupagus:
- [Firefox] Stop building the `metadata.inc`/`chrome.manifest.inc` files during `gulp locale` (PR 9566 follow-up)

   With the removal of the (standalone) Firefox building code in PR #9566 (a year and a half ago), these files are now completely unused.
Hence it doesn't really make sense to keep building them as part of `gulp locale`, and the existing files in the `l10n` folder can also be removed (thanks to version control, they're easy enough to restore should the need ever arise).

 - [Firefox] Stop fetching the `chrome.properties` files during `gulp importl10n` (PR 9566 follow-up)

   With the removal of the (standalone) Firefox building code in PR #9566 (a year and a half ago), these files are now completely unused in the GitHub repository[1].
Hence it doesn't really seem necessary to keep fetching them with `gulp importl10n`, and the existing files in the `l10n` folder can also be removed (thanks to version control, they're easy enough to restore should the need ever arise).

   The patch also allows an additional simplification, for the `gulp locale` and `gulp mozcentral` commands, since it's now possible to stop writing `l10n` files to the `extensions/firefox/` folder and instead just copy them similar to other build targets.

   ---
   [1] They're obviously still used in `mozilla-central`, for fallback messages displayed through `PdfStreamConverter.jsm`, but that doesn't make it necessary to keep them *here* as far as I'm concerned.

 - Update l10n files


timvandermeij (approved):

timvandermeij:
Looks like a good change to me!

PR #11261, opened 2019-10-19
Re-factor sending of various Exceptions from the worker to the API

Snuffleupagus:
As can be seen in the API, there's a number of document loading Exception handlers which are both really simple and highly similar. Hence these are changed such that all the relevant Exceptions are sent via *one* message instead.

Furthermore, the patch also avoids unnecessarily re-creating `UnknownErrorException`s at the worker side and removes an unnecessary `bind` call.


timvandermeij (approved):

timvandermeij:
Looks like a good simplification. Thanks!

PR #11262, opened 2019-10-19
Re-add the en-US `chrome.properties` l10n file, to avoid it being removed at mozilla-central (PR 11256 follow-up)

Snuffleupagus:
Unfortunately I forgot to test `gulp mozcentraldiff` with PR #11256, which is really bad since it will cause the en-US `chrome.properties` l10n file to be deleted at mozilla-central; sorry about breaking this!

In order to address this we'll have to re-add (only) the en-US `chrome.properties` l10n file, which is simple enough. Furthermore, since we're not doing any sort of build-specific parsing of the l10n files, we can just copy the en-US files as-is rather than having to run `gulp locale` during `gulp mozcentral`.

### DIFF
--- a/external/importL10n/locales.js
+++ b/external/importL10n/locales.js
@@ -22,6 +22,8 @@ var path = require('path');
 // Fetches all languages that have an *active* translation in mozilla-central.
 // This is used in gulpfile.js for the `importl10n` command.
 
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this so quickly!

Issue #11264, opened 2019-10-20
Duplicate and contradictory CSS rule in viewer.css


stephanrauh:
Hi team,

I've just noticed a duplicate definition of a CSS rule:

```CSS
.toolbarField {
  ...
  border: 1px solid transparent;
   ...
  border: 1px solid hsla(0,0%,0%,.35);
  border-color: hsla(0,0%,0%,.32) hsla(0,0%,0%,.38) hsla(0,0%,0%,.42);
```

IMHO there's one or two rules too many.

For the sake of convenience, here's the [link to the line in the repo](https://github.com/mozilla/pdf.js/blob/d7f651aa9885c1848d1c8694c038bf4e0e09d6fe/web/viewer.css#L1030).

Best regards,
Stephan

Virtual1:
I'd like to claim this, if possible? See if I can figure out & fix it. 
Thanks 
Jessica

timvandermeij:
It's yours! I would check which rules are actually applied by the browser for `.toolbarField` using the developer console. I'm pretty sure the `border: 1px solud transparent` is overwritten by the other `border` entry, but I don't know which border color is used. I also think there are a few more CSS entries where the border has a color, but there is also a separate `border-color` property; I think the latter overrules the former.

Zdobnov:
As this issue was still opened, I had an initiative: https://github.com/mozilla/pdf.js/pull/11268. This PR refactors the `viewer.css` a bit + removes duplications mentioned above.
I hope it will be useful.
Thanks.

Virtual1:
Hi, I guess @Zdobnov fixed it, let me know if there's still something to check here?
Thanks,
Jessica

Snuffleupagus:
Given that someone had already expressed interest in working on this *less than a day ago*, in https://github.com/mozilla/pdf.js/issues/11264#issuecomment-544543001 above, generally speaking it seems like somewhat bad form to just "steal" the issue.

stephanrauh:
@Virtual1 If you want to help the project, you could have a quick look at #11264 and confirm this bug has been solved. You could also have a look at a second code duplication. Originally, I wanted to open a second ticket, but chances are it has been solved, so we can probably simplify the process:

![image](https://user-images.githubusercontent.com/3045767/67318608-00012b00-f50c-11e9-8f35-798b6a82ca83.png)
[...]

PR #11268, opened 2019-10-22
refactor: viewer.css refactored, duplications removed

Zdobnov:
## Context
This ticket requires refactoring of the `viewer.css file`: https://github.com/mozilla/pdf.js/issues/11264

## What was done
- border- & box-shadow duplications removed;

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -1027,12 +1027,12 @@ html[dir='rtl'] .verticalToolbarSeparator {
 .toolbarField {
   padding: 3px 6px;
   margin: 4px 0 4px 0;
-  border: 1px solid transparent;
   border-radius: 2px;
   background-color: hsla(0,0%,100%,.09);
   background-image: linear-gradient(hsla(0,0%,100%,.05), hsla(0,0%,100%,0));
   background-clip: padding-box;
-  border: 1px solid hsla(0,0%,0%,.35);
+  border-width: 1px;
+  border-style: solid;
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
> refactored all `z-index` values, as there is no necessity to have such a huge abyss between 10k & 30k (nobody will use 13,876 for example).

That doesn't seem like it belong in this PR, since all of the other changes are essentially "mechanical" and probably won't require that much testing.
There's also the question if we even want to change this, and if so *how*, since I believe that the `z-index`s were purposely implemented such that there's some "space" between them to more easily avoid errors.

Zdobnov:
@Snuffleupagus if it's necessary to implement something new with `z-index`, between 5 and 10 values there are still 4 steps. In the previous solution values where 199 and 200, and 10k, 20k, etc. at the same moment. It was a bit confused.

Snuffleupagus:
> Anyway, if that's a problem, I can remove all changes excluding duplications avoiding, just let me know.

As is hopefully clear from https://github.com/mozilla/pdf.js/pull/11268#issuecomment-544907443, I obviously think that part should be reverted.
(Also, it's always best if the *commit message* itself contains all of the information from the PR description.)

Zdobnov:
@Snuffleupagus thank you for advice, did some changes, now the PR contains only necessary changes, may I ask you to check it?
Thank you.

timvandermeij (approved):

timvandermeij:

PR #11271, opened 2019-10-23
Re-factor `StatTimer` usage, in `src/display/api.js`, and remove `DummyStatTimer`

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Good simplification!

PR #11281, opened 2019-10-26
Support Blend Modes which are specified in an Array of Names (issue 11279)

Snuffleupagus:
According to the specification, the first *supported* Blend Mode should be choosen in this case; please see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G10.4848607

Fixes #11279

PR #11283, opened 2019-10-28
Ensure that `PartialEvaluator.hasBlendModes` handles Blend Modes in Arrays (PR 11281 follow-up)

Snuffleupagus:
I completely overlooked this in PR #11281, but you obviously need to make similar changes in `PartialEvaluator.hasBlendModes` since it will otherwise ignore valid Blend Modes.


timvandermeij (approved):

timvandermeij:
Thank you for the follow-up!

PR #11284, opened 2019-10-28
Make the `ObjectLoader` use more efficient methods when determining if data needs to be loaded

Snuffleupagus:
Currently, for data in `ChunkedStream` instances, the `getMissingChunks` method is used in a couple of places to determine if data is already available or if it needs to be loaded.

When looking at how `ChunkedStream.getMissingChunks` is being used in the `ObjectLoader` you'll notice that we don't actually care about which *specific* chunks are missing, but rather only want essentially a yes/no answer to the "Is the data available?" question.
Furthermore, when looking at how `ChunkedStream.getMissingChunks` itself is implemented you'll notice that it (somewhat expectedly) always iterates over *all* chunks.

All in all, using `ChunkedStream.getMissingChunks` in the `ObjectLoader` seems like an unnecessary "heavy" and roundabout way to obtain a boolean value. However, it turns out there already exists a `ChunkedStream.allChunksLoaded` method, consisting of a *single* simple check, which seems like a perfect fit for the `ObjectLoader` use cases.
In particular, once the *entire* PDF document has been loaded (which is usually fairly quick with streaming enabled), you'd really want the `ObjectLoader` to be as simple/quick as possible (similar to e.g. loading a local files) which this patch should help with.

Note that I wouldn't expect this patch to have a huge effect on performance, but it will nonetheless save some CPU/memory resources when the `ObjectLoader` is used. (As usual this should help larger PDF documents, w.r.t. both file size and number of pages, the most.)

### DIFF
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -285,6 +285,12 @@ class ChunkedStream {
       }
       return missingChunks;
     };
+    ChunkedStreamSubstream.prototype.allChunksLoaded = function() {
+      if (this.numChunksLoaded === this.numChunks) {
+        return true;
+      }
+      return this.getMissingChunks().length === 0;
+    };
 
     const subStream = new ChunkedStreamSubstream();
     subStream.pos = subStream.start = start;
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -2038,13 +2038,13 @@ var FileSpec = (function FileSpecClosure() {
  */
 let ObjectLoader = (function() {
   function mayHaveChildren(value) {
-    return isRef(value) || isDict(value) || Array.isArray(value) ||
-           isStream(value);
+    return (value instanceof Ref) || (value instanceof Dict) ||
+           Array.isArray(value) || isStream(value);
   }
 
   function addChildren(node, nodesToVisit) {
-    if (isDict(node) || isStream(node)) {
-      let dict = isDict(node) ? node : node.dict;
+    if ((node instanceof Dict) || isStream(node)) {
+      let dict = (node instanceof Dict) ? node : node.dict;
       let dictKeys = dict.getKeys();
       for (let i = 0, ii = dictKeys.length; i < ii; i++) {
         let rawValue = dict.getRaw(dictKeys[i]);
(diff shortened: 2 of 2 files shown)


src/core/obj.js
```diff
@@ -2072,14 +2071,14 @@ let ObjectLoader = (function() {
...
 
   ObjectLoader.prototype = {
-    load() {
-      this.capability = createPromiseCapability();
-      // Don't walk the graph if all the data is already loaded.
-      if (!(this.xref.stream instanceof ChunkedStream) ||
-          this.xref.stream.getMissingChunks().length === 0) {
-        this.capability.resolve();
-        return this.capability.promise;
+   async load() {
```
timvandermeij: Missing one space of indentation here.

src/core/chunked_stream.js
```diff
@@ -285,6 +285,12 @@ class ChunkedStream {
       }
       return missingChunks;
     };
+    ChunkedStreamSubstream.prototype.allChunksLoaded = function() {
+      if (this.numChunksLoaded === this.numChunks) {
+        return true;
+      }
+      return this.getMissingChunks().length === 0;
```
timvandermeij: Why is this check still needed here, i.e., why isn't the check above sufficient?

src/core/obj.js
Snuffleupagus: Whoops, looks like just pressing <tab> in the editor didn't work there :-)

src/core/chunked_stream.js
Snuffleupagus: The first check is essentially just a fast-path, but not *necessarily* a required condition for considering a `ChunkedStreamSubstream` instance as completely loaded.

Please keep in mind that we're inheriting from a `ChunkedStream` instance above, note in particular the special `getMissingChunks` method used here, hence why only a partially loaded `ChunkedStream` can still be a fully loaded `ChunkedStreamSubstream`.

Hopefully this all makes sense, since it does look a bit weird at first!

timvandermeij: Ah, the special `getMissingChunks` method is what I overlooked here. Thanks!

timvandermeij (approved):

timvandermeij:
Good improvement!

PR #11290, opened 2019-10-30
[MessageHandler] Re-factor and convert the code to a proper `class`

Snuffleupagus:
This should, most likely, finish the somewhat recent round of clean-up/re-factoring of the `MessageHandler` code. For easier reviewing, I'd suggest looking at each commit separately and also ignoring white-space changes where appropriate.


timvandermeij (approved):

timvandermeij:
I like it. Thanks!

PR #11294, opened 2019-10-31
Remove some unused `require` statements, used when loading fake workers, in non-`PRODUCTION` mode

Snuffleupagus:
The code in question is *only* relevant in non-`PRODUCTION` mode, i.e. the *development* version of the viewer run with `gulp server`, and has been completely unused at least since SystemJS was added.
I really cannot see any reason to keep this, since it's code which first of all isn't shipping and secondly isn't even being used in the development viewer.


timvandermeij (approved):

timvandermeij:

PR #11295, opened 2019-10-31
function -> ES6 arrow function

archsyscall:
I changed the callback function from function to arrow function.

### DIFF
--- a/src/display/network.js
+++ b/src/display/network.js
@@ -14,15 +14,18 @@
  */
 /* eslint no-var: error */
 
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
First of all, this PR contains *a lot* of completely unrelated changes which are not relevant to the thing you're trying to "fix", breaking existing code-formatting, and failing the lint job.
Secondly, and more importantly, we only want to use arrow functions when *actually* necessary. In either of these cases, there's no need to use them since we don't need to access `this` of the surrounding scope.

All-in-all, this unfortunately cannot be accepted.

PR #11298, opened 2019-11-01
Publish event when find matches have finished rendering

marginalhours:
This might be a bit niche but would be very useful for an application I'm working on! 

When using the PDF search capability, it would be nice to have a way of knowing for definite when the viewer has finished jumping to the next match. Since the next match might be on a page which hasn't loaded yet, the creation of the element indicating the match is asynchronous.

My understanding is that `TextLayerBuilder` subscribes to the `updatetextlayermatches` event, which causes them to call `_renderMatches`, which in turn mutates the DOM in order to set the correct class on the span indicating the current match. 

This PR adds a `findmatchesrendered` event to the end of the `_renderMatches` function, which will be fired after all the element changes have happened. 

This allows for knowing exactly when the new current match element is present in the DOM.

### DIFF
--- a/web/text_layer_builder.js
+++ b/web/text_layer_builder.js
@@ -291,6 +291,11 @@ class TextLayerBuilder {
     if (prevEnd) {
       appendTextToDiv(prevEnd.divIdx, prevEnd.offset, infinity.offset);
     }
+
+    this.eventBus.dispatch('findmatchesrendered', {
+      source: this,
+      pageNumber: this.pageNumber,
+    });
   }
 
   _updateMatches() {


Snuffleupagus:
A few observations based on a quick look:

 - This is basically adding, from the perspective of the default viewer, completely unused code. The issue with adding unused code is that it increases the maintenance overhead (for anyone needing to touch this code), and it would thus be at risk of being removed and/or breaking at a future time.

   If there's a way to easily add e.g. a unit-test for this, the likelihood of it being accepted most likely increases a whole lot :-)

 - > This allows for knowing exactly when the new current match element is present in the DOM.

   Please note that in the `highlightAll` case, *none* of the rendered matches need to be the *current* match. Whether or not that matters for your use-case I've got no idea, but it seemed like a good idea to at least make a note about it :-)

 - The implementation appears to be somewhat inconsistent, given that there's a couple of early returns in `TextLayerBuilder._renderMatches`, leading to the new event not firing consistently.

    Firing the event [after this line instead](https://github.com/mozilla/pdf.js/blob/30ef05c16186e9802ed91ae82808149ed5b20832/web/text_layer_builder.js#L327) would probably address that, perhaps with a `numMatches: matches.length,` parameter added to the event to allow differentiating the cases (similar to `textlayerrendered`, which is always firing regardless of the number of textDivs). 

 - Assuming that this PR is deemed acceptable, then the *commit message* itself should contain all of the information you've added to the PR description (since the latter isn't available e.g. from the `Git` command line).

MisterKeefe:
Thanks for the quick response @Snuffleupagus -- I probably should have raised an issue for this instead! From reading around a bit more, it looks like the right thing to do is probably bake my own viewer rather than trying to make changes to the default viewer code. I'll close this :+1:

Snuffleupagus:
> [...] it would be nice to have a way of knowing for definite when the viewer has finished jumping to the next match.

Again, I don't know exactly what your requirement are, however it just occurred to me you may want to dispatch the event at the end this method instead https://github.com/mozilla/pdf.js/blob/30ef05c16186e9802ed91ae82808149ed5b20832/web/pdf_find_controller.js#L179-L194
That way, it only fires *after* "jumping to the next match" as asked for above, and won't trigger for e.g. `highlightAll` changes.

Honestly, at least to me, it seems that the chance of that being accepted here seems much larger since for one it probably should be a lot easier to add a unit-test for it :-)

PR #11300, opened 2019-11-01
Simplifies code contributions by automating the dev setup

nisarhassan12:
Hi! there, :slightly_smiling_face:

this PR adds support for Gitpod, a free online tool that lets you automate your dev environment, eases the hurdles for contributions and generally working on GitHub projects. It allows anyone to get a ready-to-code dev environment for any branch, issue and pull request with a single click.
In the case of this project, it will launch a ready to code workspace with the dependencies installed and the web server running so that contributors can start hacking around with the project without wasting any precious time on the development setup.

You can give it a try on my fork of the repo: https://gitpod.io/#https://github.com/nisarhassan12/pdf.js

image

It seems to work well! Please let me know if there is anything that can be improved. I would love to help :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,12 @@ It is also possible to view all test PDF files on the right side by opening:
 
 + http://localhost:8888/test/pdfs/?frame
 
+## Online setup with a single click
+
+It's recommended to follow [Getting the Code] above, however you can also use [Gitpod](https://gitpod.io) (a free online VS Code like IDE). With a single click it will launch a ready to use workspace, with all the dependencies pre-installed and the web server running, so that you can start coding straight away.
+
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/mozilla/pdf.js)
+
 ## Building PDF.js
 
 In order to bundle all `src/` files into two production scripts and build the generic
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,7 @@
+FROM gitpod/workspace-full-vnc
+
+USER gitpod
+
+RUN sudo apt-get update && \
+    sudo apt-get install -yq firefox && \
+    sudo rm -rf /var/lib/apt/lists/*
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,13 @@
+image:
+  file: .gitpod.Dockerfile
+tasks:
+  - command: |
+      gp await-port 8888 && gp preview $(gp url 8888)/web/viewer.html && echo '[{"name": "Firefox","path": "/usr/bin/firefox"}]' | jq '.' > test/resources/browser_manifests/browser_manifest.json
+
+  - init: npm install -g gulp-cli && npm install
+    command: gulp server
+ports:
+  - port: 8888
+    onOpen: ignore
+  - port: 6080
+    onOpen: ignore


Snuffleupagus:
> It allows anyone to get a ready-to-code dev environment for any branch, issue and pull request with a single click.

Would this work-flow actually support running any of the [PDF.js test-suits](https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing), or even running lint, since making it easier for people to submit completely untested PRs probably isn't such a great idea!?

timvandermeij:
Yes, running lint and unit tests would be a very good idea. Ideally you can also run the regression tests though, but I don't know if that is possible in the workflow; running lint and unit tests should already catch most issues.

Snuffleupagus:
> Ideally you can also run the regression tests though, but I don't know if that is possible in the workflow

Unless it's possible to follow the entire work-flow *and* run all test-suite (including reference tests) mentioned in https://github.com/mozilla/pdf.js/wiki/Contributing, then advertising this in the README seems like it might cause some unnecessary friction for a would be contributor.

For example: Let's assume that we add this and someone uses it to submit a patch touching e.g. `core/` code, and the patch subsequently needs revising because it causes test failures.
At this point, it's quite likely that a reviewer will suggest running tests locally while developing the patch, pointing to https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing, to which the contributor may then reply:
`But, I just followed the "Online setup with a single click" section from the README and {insert test-suite} here won't run there!`

That sort of situation doesn't seem great at all, at least as far as I'm concerned.

nisarhassan12:
Thanks! Both linting and the test suite seem to work well :slightly_smiling_face: . 

image
image


With this workflow, contributors don't have to generate references images Gitpod automatically does that.

Both running the linting and the test suite is same as it is in local setup i.e `gulp lint` for linting and `gulp test` for running the test suite both can be run via the default built-in terminal.


For the server port once the test suite is started contributors can view it from accessing port 6080 from the ports tab in the bottom right corner:   
image

README.md
```diff
@@ -71,6 +71,12 @@ It is also possible to view all test PDF files on the right side by opening:
 
 + http://localhost:8888/test/pdfs/?frame
 
+## Online Setup with a single click
+
+You can use [gitpod](https://gitpod.io) (A free online VS Code like IDE). With a single click it will launch a ready to code workspace with all the dependencies pre-installed and the web server running so that you can start hacking around straight away without wasting any precious time.
```
Snuffleupagus: I'd suggest prefacing this section by suggesting that people follow the `Getting the Code` section above, since that should still be considered the primary/recommended way to contribute.
Furthermore, let's remove "wasting time" bit, since that seems unnecessary and out of place.

All in all, I'd suggest something like the below (which also contains some fixes for weird casing):
```
It's recommended to follow [Getting the Code] above, however you can also use [Gitpod](https://gitpod.io) (a free online VS Code like IDE).
With a single click it will launch a ready to use workspace, with all the dependencies pre-installed and the web server running, so that you can start coding straight away.
```

Snuffleupagus (changes requested):
Please note all patches much contain a good *commit message* that describes exactly what's being changed and why the change is made; hence something like e.g. "Configures gitpod" is not at all useful (even more so when using the Git command line).

README.md
nisarhassan12: Thanks! Your recommendation here sounds more legit and simple to me.

nisarhassan12:
Thanks! @Snuffleupagus I have made the changes and updated the branch.

Snuffleupagus:
> With this workflow, contributors don't have to generate references images Gitpod automatically does that.

At what point is this actually done?
Because as is clearly outlined in https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing, this steps needs to happen *before* any changes have been made to be of any use.

nisarhassan12:
> > With this workflow, contributors don't have to generate references images Gitpod automatically does that.
> 
> At what point is this actually done?
> Because as is clearly outlined in https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing, this step needs to happen _before_ any changes have been made in order for the reference images to be correct.

Oops! Sorry! I was absolutely wrong :disappointed:  I meant to say that contributors don't have to run this command `cp test/resources/browser_manifests/browser_manifest.json.example test/resources/browser_manifests/browser_manifest.json` which is not actually generating the reference images. The contributors can generate references images as it is in the local setup by running `gulp makeref` before they make any changes.

timvandermeij (approved):

timvandermeij:
Let's give this a try. Thank you!

PR #11301, opened 2019-11-01
Ensure that the `peekByte` methods, on the various Streams, handles end of data correctly (PR 5286 follow-up)

Snuffleupagus:
When the end of data has already been reached for the various Streams, the `getByte` methods will return `-1` to signal that to the caller. Note however that the current position obviously won't be incremented in this case, meaning that the `peekByte` methods will in this case *incorrectly* decrement the position.

Thankfully the corresponding `peekBytes` shouldn't be affected by this bug, since they decrement the current position with the *actually* returned number of bytes.

I'm not aware of any bugs caused by this blatant oversight, but that doesn't mean this shouldn't be fixed :-)

### DIFF
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -213,7 +213,9 @@ class ChunkedStream {
 
   peekByte() {
     const peekedByte = this.getByte();
-    this.pos--;
+    if (peekedByte !== -1) {
+      this.pos--;
+    }
     return peekedByte;
   }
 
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

Snuffleupagus:
Well, I did *cause* it so fixing it was probably the least I could do here :-)

As always, thanks for reviewing and merging this!

PR #11302, opened 2019-11-02
Slightly re-factor setting of the link `target` in `addLinkAttributes`

Snuffleupagus:
I happened to look at this code and the way that the link target is set seems unecessarily convoluted, since we're using `Object.values` and `Array.prototype.includes` for *every* link being parsed.
Given that the number of link targets are so few, the easist solution honestly seem to be to just use a `switch` statement to do the link target mapping.

### DIFF
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -326,14 +326,6 @@ const LinkTarget = {
   TOP: 4,
 };
 
-const LinkTargetStringMap = [
-  '',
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
I definitely agree with this; I also had to re-read the code to see how it worked. Thank you for simplifying this!

PR #11303, opened 2019-11-03
maintain consistency

archsyscall:
Like other callback functions, i have omitted space.

### DIFF
--- a/src/display/network.js
+++ b/src/display/network.js
@@ -409,7 +409,7 @@ class PDFNetworkStreamFullRequestReader {
   cancel(reason) {
     this._done = true;
     this._headersReceivedCapability.reject(reason);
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Honestly, it doesn't seem at all necessary to fix "issues" like this piecewise since it will only cause `git blame` to be needlessly shifted for these lines. *If* these sort of changes should be done, they ought to be enforced by ESLint such that the *entire* code-base is fixed at once.

Also, please note all patches much contain a good *commit message* that describes exactly what's being changed *and* why the change is made; please note that "maintain consistency" isn't useful.

timvandermeij:
I agree. There are definitely more places where this is the case, so using an ESLint rule like https://eslint.org/docs/rules/space-before-function-paren would be a much better solution.

PR #11304, opened 2019-11-03
Re-factor the `ObjectLoader._walk` method to be properly asynchronous

Snuffleupagus:
Rather than having to store a `PromiseCapability` on the `ObjectLoader` instances, we can simply convert `_walk` to be `async` and thus have the same functionality with native JavaScript instead.


timvandermeij (approved):

timvandermeij:
Nice improvement!

PR #11305, opened 2019-11-04
Fix nodejs core module :  ignore url module for web browser

mgrigis:


timvandermeij:
Could you elaborate about what this fixes exactly? There is no issue associated with this PR, so it's a bit unclear why this change is necessary. Is something broken now?

mgrigis:
we try to integrate the library in an ember application, and we cannot to this due to the url module dependency (part of core nodejs). The solution was to exclude it  like other nodejs core module.

Maybe it can be done in a different way.

mgrigis:
> Could you elaborate about what this fixes exactly? There is no issue associated with this PR, so it's a bit unclear why this change is necessary. Is something broken now?

we try to integrate the library in an ember application, and we cannot to this due to the url module dependency (part of core nodejs). The solution was to exclude it like other nodejs core module.

Maybe it can be done in a different way.

timvandermeij:
@Snuffleupagus I think this is safe, but I can't fully oversee this. What do you think?

Snuffleupagus:
> @Snuffleupagus I think this is safe, but I can't fully oversee this. What do you think?

Based on https://github.com/mozilla/pdf.js/blob/6763e1680482b4e91d853445dfd3d9c0f82936e9/src/display/node_stream.js#L17-L20 and https://github.com/mozilla/pdf.js/blob/6763e1680482b4e91d853445dfd3d9c0f82936e9/src/display/svg.js#L136 it seems to me that `url` was simply left out of that list; so it looks OK as far as I can tell :-)

timvandermeij (approved):

timvandermeij:
Thanks!

PR #11307, opened 2019-11-05
Support UTF-16 little-endian strings in the `stringToPDFString` helper function (bug 1593902)

Snuffleupagus:
The bug report seem to suggest that we don't support UTF-16 strings with a BOM (byte order mark), which we *actually* do as evident by both the code and a unit-test.
The issue at play here is rather that we previously only supported big-endian UTF-16 BOM, and the `Title` string in the PDF document is using a *little-endian* UTF-16 BOM instead.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1593902

*Edit:* The PDF spec only mentions *big* endian as supported (in addition to `PDFDocEncoding` of course), see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G6.1957385, hence this is actually a case where a PDF generator is (yet again) creating corrupt/invalid data.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -752,6 +752,12 @@ function stringToPDFString(str) {
       strBuf.push(String.fromCharCode(
         (str.charCodeAt(i) << 8) | str.charCodeAt(i + 1)));
     }
+  } else if (str[0] === '\xFF' && str[1] === '\xFE') {
+    // UTF16LE BOM
+    for (let i = 2; i < length; i += 2) {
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
Forgot to extend one of the existing unit-tests...

/botio unit-test

timvandermeij (approved):

timvandermeij:
Nice find!

PR #11312, opened 2019-11-08
Convert `Catalog.getPageDict` to an `async` method

Snuffleupagus:
This makes it possible to remove the internal `next` helper function, and also gets rid of the need to manually resolve/reject a `PromiseCapability`.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -14,10 +14,9 @@
  */
 
 import {
-  assert, bytesToString, createPromiseCapability, createValidAbsoluteUrl,
-  FormatError, info, InvalidPDFException, isBool, isNum, isString,
-  PermissionFlag, shadow, stringToPDFString, stringToUTF8String, unreachable,
-  warn
+  assert, bytesToString, createValidAbsoluteUrl, FormatError, info,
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for refactoring this! I found the `next` mechanism a bit hard to read.

Snuffleupagus:
> I found the `next` mechanism a bit hard to read.

Yeah, the old setup was essentially a recursive function which was called asynchronously. This was also a case where the use of a `PromiseCapability` probably didn't help readability either.

I suppose the one criticism you *could* have of this patch, is that it's adding asynchronous behaviour inside of a loop. However I think all the skipped function calls and the general clean-up outweighs that.

Snuffleupagus:
Sigh, it seems that I forgot to test this with really large files :-(

With a large file taken from an old GitHub issue (warning, the file is 16 MB): https://github.com/mozilla/pdf.js/files/876321/kjv.pdf there's now a considerable "Page Request" regression for the *second* page upon document load; from 2000 -> 3000 ms which obviously doesn't look good.

In light of this I don't believe that it's a good idea keep these changes unfortunately, since the new way in which this code is asynchronous obviously performs worse in some cases; sorry about the mess here!

Issue #6179, opened 2015-07-06
Strange annotation borders after PR 5585


Snuffleupagus:
In http://tug.ctan.org/tex-archive/macros/latex/contrib/pdfcomment/doc/pdfcomment.pdf#page=6 (note that the issue is present in the entire file), most of the annotations now have a coloured border, which they didn't have prior to PR #5585. *Please refer to the screen-shot below.*

Interestingly enough, when comparing the border colours to the rendering in Adobe Reader, it appears that the borders are rendered with the colours that the actual annotations *should* have.

I'm tentatively marking this as a regression from #5585, since there shouldn't be any borders visible.
/cc @timvandermeij 

---
![borders](https://cloud.githubusercontent.com/assets/2692120/8520761/b0aa87ee-23dd-11e5-94f0-fbd583112a3d.png)

---

![borders_reader](https://cloud.githubusercontent.com/assets/2692120/8520933/5a963626-23df-11e5-926d-adb4f7b1272a.png)

timvandermeij:
From the specification:

"If neither the Border nor the BS entry is present, the border is drawn as a solid line with a width of 1 point."

In this case, all annotations lack a Border and a BS entry, so I think the functionality in my PR is actually the right one according to the spec. Since it is not the first time that Adobe Reader violates its own spec, I will create a patch that does not draw the border when no Border or BS dictionaries are available.

CodingFabian:
@timvandermeij maybe in that case they did implement the border but had set the color to transparent? I think we came to the conclusion that transparent would be the logical case but adobe in the other case decided to make them black....

timvandermeij:
That is also a possibility, but the specification unfortunately says nothing about this, so there is no way to tell for sure. With the merged patch above we now have the same display as Adobe Reader, so we should be good now, even though the specification should have been more clear about this.

PR #11313, opened 2019-11-09
Ensure that Popup annotations, where the parent annotation is a polyline, will always be possible to open/close (issue 11122)

### LINKED ISSUE #11122: Ink annotation popup not shown

[ink_annotation.pdf](https://github.com/mozilla/pdf.js/files/3579759/ink_annotation.pdf)

Configuration:
- Web browser and its version: Firefox Developer 70.0b3 (64-bit)
- Operating system and its version: Windows 10 1903 18362.329
- PDF.js version: [Demo page](https://mozilla.github.io/pdf.js/web/viewer.html)
- Is a browser extension: No

1. Open the file
2. Popup for the ink annotation which has line width 1 is not shown

### _What is the expected behavior?_

When the cursor is over the ink annotation, the popup annotation should show up as a hover element.

### _What went wrong?_

The problem seems to be how PDF.js handles default missing 'BS' entries in the annotation dictionary. According to the standard a BorderStyle entry with width 1 and style solid is a default entry and doesn't need to be serialized. PDF.js seems to rely on this entry for the popup detection/display and when it's missing the popup does not show.

Snuffleupagus:
For Popup annotation trigger elements consisting of an arbitrary polyline, you need to ensure that the 'stroke-width' is always non-zero since otherwise it's impossible to actually open/close the popup.

Unfortunately I don't believe that any of the test-suites can be used to test this, hence why no tests are included in the patch.

Fixes #11122

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -879,7 +879,9 @@ class LineAnnotationElement extends AnnotationElement {
     line.setAttribute('y1', data.rect[3] - data.lineCoordinates[1]);
     line.setAttribute('x2', data.rect[2] - data.lineCoordinates[2]);
     line.setAttribute('y2', data.rect[3] - data.lineCoordinates[3]);
-    line.setAttribute('stroke-width', data.borderStyle.width);
+    // Ensure that the 'stroke-width' is always non-zero, since otherwise it
+    // won't be possible to open/close the popup (note e.g. issue 11122).
+    line.setAttribute('stroke-width', data.borderStyle.width || 1);
     line.setAttribute('stroke', 'transparent');
 
     svg.appendChild(line);
@@ -927,7 +929,9 @@ class SquareAnnotationElement extends AnnotationElement {
     square.setAttribute('y', borderWidth / 2);
     square.setAttribute('width', width - borderWidth);
     square.setAttribute('height', height - borderWidth);
-    square.setAttribute('stroke-width', borderWidth);
+    // Ensure that the 'stroke-width' is always non-zero, since otherwise it
+    // won't be possible to open/close the popup (note e.g. issue 11122).
+    square.setAttribute('stroke-width', borderWidth || 1);
     square.setAttribute('stroke', 'transparent');
     square.setAttribute('fill', 'none');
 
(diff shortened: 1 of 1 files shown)


timvandermeij:
Initially I was a bit confused because I thought the default border width was 1 if it is not found. It turns out that I added https://github.com/mozilla/pdf.js/commit/3a6eed6248d5d4c985b3315e920401c1ccca90cb later on to fix another issue, which might have had this issue as an effect. Looking at #6179, the original issue that required that change, there is also https://github.com/mozilla/pdf.js/issues/6179#issuecomment-118978210. Perhaps that's a better fix, i.e., if no border is present consider it to be width 1, but set its color to transparent? This might actually be more in line with the specification and would address both issues. We could even add a test case for the file in #6179 to verify this; I don't know why I didn't do that for that commit. What do you think?

Snuffleupagus:
> What do you think?

Hmm, interesting idea!
However would that actually be guaranteed to always work correctly, since I suppose that an annotation could explicitly set the border to `0` in which case wouldn't this problem just re-surface!?

timvandermeij:
Yes, but if the annotation explicitly sets a border width of 0, I think that should be respected and I don't think any other PDF viewer will show the popup in that case. This patch only applies to the case where no `Border`/`BS` entries are set at all.

Snuffleupagus:
> [...] and I don't think any other PDF viewer will show the popup in that case.

Assuming that this exact behaviour can be reproduced in e.g. Adobe Reader as well, then your solution definitely looks better :-)

Snuffleupagus:
> > [...] and I don't think any other PDF viewer will show the popup in that case.
> 
> Assuming that this exact behaviour can be reproduced in e.g. Adobe Reader as well, then your solution definitely looks better :-)

I've just tried this, by modifying the PDF document from #11122, and it seems that even when setting the `BS` entry explicitly to `/W 0` the popup can still be toggled just fine in Adobe Reader; hence I'm unfortunately not sure that this idea is viable (despite the fact that I *really* like it a lot).

Snuffleupagus:
> > [...] and I don't think any other PDF viewer will show the popup in that case.
> 
> Assuming that this exact behaviour can be reproduced in e.g. Adobe Reader as well, then your solution definitely looks better :-)

I've just tried this, by modifying the PDF document from #11122, and it seems that even when setting the `BS` entry explicitly to `/W 0` the popup can still be toggled just fine in Adobe Reader; hence I'm unfortunately not sure that this idea is viable (despite the fact that I *really* like it a lot).

timvandermeij (approved):

timvandermeij:
I also thought about this some more, but found that this patch is probably closest to what Adobe Reader does. Even if we chose the one pixel border without color, the explicit `/W 0` case wouldn't be covered whereas this patch does. Thank you for fixing this!

Snuffleupagus:
Thanks for landing this, despite the implementation not being that great.

*Following up on https://github.com/mozilla/pdf.js/pull/11313#issuecomment-552099005:* Do you have time to add a test-case for issue #6179 now, or should I just submit a PR with a "linked" annotation reference test for now?

timvandermeij:
I'm hoping it's not too difficult to create a reduced test case for this, so I'll try to make that today. If it doesn't work out, we can always use a linked test case.

PR #11315, opened 2019-11-10
Include a reduced test case for annotations without a `Border`/`BS` entry (PR 6180 follow-up)

timvandermeij:
This should have been included in the original PR. I verified that the border does appear without PR 6180 applied and disappears with PR 6180 applied.

/cc @Snuffleupagus


Snuffleupagus (approved):
Looks good to me; thanks for adding more test-coverage!

PR #11316, opened 2019-11-10
Convert `globalScope` and `isNodeJS` to proper modules

Snuffleupagus:
Slightly unrelated to the rest of the patch, but this also removes an out-of-place `globals` definition from the `web/viewer.js` file.

### DIFF
--- a/src/shared/is_node.js
+++ b/src/shared/is_node.js
@@ -12,13 +12,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* globals module, process */
+/* globals process */
 
-module.exports = function isNodeJS() {
+function isNodeJS() {
   // NW.js / Electron is a browser context, but copies some Node.js objects; see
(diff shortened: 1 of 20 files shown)


src/shared/is_node.js
```diff
@@ -12,13 +12,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* globals module, process */
+/* globals process */
 
-module.exports = function isNodeJS() {
+function isNodeJS() {
```
Snuffleupagus: I'm not entirely sure why this *needs* to be a function, since it should be marginally more efficient/simpler to have this be a constant!?

timvandermeij: I'm fine with making this a constant. That should indeed be simpler given that computing this only once suffices.

timvandermeij (approved):

timvandermeij:
Looks better like this. Thanks!

PR #11318, opened 2019-11-11
[PDFHistory] Move the IE11 `pushState`/`replaceState` work-around to `src/shared/compatibility.js` (PR 10461 follow-up)

Snuffleupagus:
I've always disliked the solution in PR #10461, since it required changes to the `PDFHistory` code itself to deal with a bug in IE11.
Now that IE11 support is limited, it seems reasonable to remove these `pushState`/`replaceState` hacks from the main code-base and simply use polyfills instead.


timvandermeij (approved):

timvandermeij:
Looks much better this way. The more we can isolate browser-specific hacks, the better.

PR #11325, opened 2019-11-13
Added custom file transform function

### LINKED ISSUE #11125: Remove gulp-transform dependency
It appears gulp-transform has been abandoned by the author and it contains a dependency on an old version of gulp-util which has a critical security vulnerability. I don't think the vulnerability is actually an issue for our use, but it would be nice to have a clean audit of the pdf.js deps. 

I think we could replace it with something like https://gist.github.com/CITguy/079c631ac5f181939e08

smohtadi:
Fixes #11125


gulpfile.js
```diff
@@ -40,6 +39,7 @@ var webpack2 = require('webpack');
 var webpackStream = require('webpack-stream');
 var Vinyl = require('vinyl');
 var vfs = require('vinyl-fs');
+var through = require('through2');
```
Snuffleupagus: If this patch works it's by luck alone, since you must add the [through2](https://www.npmjs.com/package/through2) package to https://github.com/mozilla/pdf.js/blob/master/package.json for this to be correct.

timvandermeij:
Could you squash the commits into one? Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits for how to do this easily if you're not familiar with the process.

Snuffleupagus:
I've diffed the output of `gulp mozcentral` and `gulp generic` locally, and there were no significant changes between `master` and this patch (except changed version numbers, as you'd expected).

/botio test

timvandermeij (approved):

timvandermeij:

PR #11334, opened 2019-11-15
Use `getBytes`, rather than looping over `getByte`, in `FlateStream.prototype.readBlock`

Snuffleupagus:
*Please note:* A a similar change was attempted in PR #5005, but it was subsequently backed out (in PR #5069) since other parts of the patch caused issues.

With these changes, it's possible to replace repeated function calls within a loop with just a single function call and subsequent assignment instead.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #11338, opened 2019-11-17
Reduce duplication when registering event listeners for the `Toolbar` buttons, and other small clean-up

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Makes sense. Thanks!

PR #11350, opened 2019-11-21
Convert `{BaseViewer, PDFThumbnailViewer}._pagesRequests` from an Array to a WeakMap

Snuffleupagus:
Obviously the `_pagesRequests` functionality is *mainly* used when `disableAutoFetch` is set, but it will also be used during ranged/streamed loading of documents.
However, the `_pagesRequests` property is currently an Array which seems a bit strange:

 - Arrays are zero-indexed, but the first element will never actually be set in the code.
 - The `_pagesRequests` Array is never cleared, unless a new document is loaded, and once the `PDFDocumentProxy.getPage` call has resolved/rejected the element is just replaced by `null`.
 - Unless the document is browsed *in order* the resulting `_pagesRequests` Array can also be arbitrarily sparse.

All in all, I don't believe that an Array is an appropriate data structure to use for this purpose.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -533,7 +533,7 @@ class BaseViewer {
     this._buffer = new PDFPageViewBuffer(DEFAULT_CACHE_SIZE);
     this._location = null;
     this._pagesRotation = 0;
-    this._pagesRequests = [];
+    this._pagesRequests = new WeakMap();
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good improvement, thanks!

PR #11352, opened 2019-11-22
Look for font locally first before fetching from url

msschambach:


Snuffleupagus (changes requested):
Unfortunately this patch cannot be considered in its current form, since there's no explanation as to *why* these changes should be made and/or what bugs this would fix. Furthermore, there's also no test-case included nor any steps to reproduce the problem (whatever it might be).

Was this patch perhaps submitted against the main PDF.js repository by mistake?

msschambach:
> Unfortunately this patch cannot be considered in its current form, since there's no explanation as to _why_ these changes should be made and/or what bugs this would fix. Furthermore, there's also no test-case included nor any steps to reproduce the problem (whatever it might be).
> 
> Was this patch perhaps submitted against the main PDF.js repository by mistake?

It wasn't by mistake but as this is a draft PR, I assume I can prepare it and publish when ready, right?

Snuffleupagus:
> It wasn't by mistake but as this is a draft PR, I assume I can prepare it and publish when ready, right?

Sure; but it might be possible for you to save yourself some time by outlining (at least in broad strokes) *upfront* why these changes would be necessary, since it's somewhat difficult to understand how this makes sense given how that part of the PDF.js font loading actually works.
Otherwise there *could* be a risk that you're spending a lot of time/effort on a patch which may not be accepted in the end.

msschambach:
Closing because I cannot continue to work on this at the moment.

PR #11353, opened 2019-11-22
Remove `document.attachEvent` code from `web/pdf_print_service.js` since it's only necessary for old IE versions

Snuffleupagus:
This code was originally added to support IE10 (and below), however with those browsers *explicitly* unsupported since PDF.js version `2.0` this code is now dead.


timvandermeij (approved):

timvandermeij:
Nice find, removing old IE-only code!

PR #11354, opened 2019-11-23
Enable the `getter-return`, `no-dupe-else-if`, and `no-setter-return` ESLint rules

Snuffleupagus:
All of these rules can help catch errors during development. Please note that only `getter-return` required a few changes, which was limited to disabling the rule in a couple of spots; please find additional details about these rules at:
 - https://eslint.org/docs/rules/getter-return
 - https://eslint.org/docs/rules/no-dupe-else-if
 - https://eslint.org/docs/rules/no-setter-return


timvandermeij (approved):

timvandermeij:
Thank you for making our linting coverage better!

PR #11356, opened 2019-11-25
Remove outdated, and misleading, JSDoc comment from the `PDFDocument` class

Snuffleupagus:
The contents of this comment hasn't been correct for *years*, ever since the library was properly split into main/worker-threads, so it's probably high time for this to be updated.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #11363, opened 2019-11-27
Remove the `Number.isInteger` checks from `XRef.fetchUncompressed` (PR 8857 follow-up)

Snuffleupagus:
Having ran the entire test-suite locally with these `Number.isInteger` checks removed, there wasn't a single test failure anywhere; see also PR #8857.
Hence everything points to this being completely unnecessary now, however I'm still slightly worried about inadvertently breaking something with a simple code removal. Hence these checks are moved, and a warning is printed, such that they only occur when the there was a type mismatch; hence there's fewer function calls in the normal case.

*Edit:* Smaller and simpler diff with https://github.com/mozilla/pdf.js/pull/11363/files?w=1

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -1682,12 +1682,6 @@ var XRef = (function XRefClosure() {
       var obj2 = parser.getObj();
       var obj3 = parser.getObj();
 
-      if (!Number.isInteger(obj1)) {
-        obj1 = parseInt(obj1, 10);
-      }
-      if (!Number.isInteger(obj2)) {
-        obj2 = parseInt(obj2, 10);
-      }
       if (obj1 !== num || obj2 !== gen || !(obj3 instanceof Cmd)) {
         throw new XRefEntryException(`Bad (uncompressed) XRef entry: ${ref}`);
       }


Snuffleupagus:
I suppose that another, riskier, option would be to just remove this code and wait and see if there's any regressions (at least that way we'd get test-cases)!?

timvandermeij:
> I suppose that another, riskier, option would be to just remove this code and wait and see if there's any regressions (at least that way we'd get test-cases)!?

I'm also on the fence here. I do think that the current patch is better because it limits how often it's run, but on the other hand I also don't like code for which we have no indication that it's actually necessary. Is there perhaps a way to tell with e.g., `git blame` why this was introduced in the first place?

Snuffleupagus:
> Is there perhaps a way to tell with e.g., `git blame` why this was introduced in the first place?

Not really in any way that helps unfortunately, since I couldn't replicate the test-failure that apparently caused that to be added; please see https://github.com/mozilla/pdf.js/pull/8857#issue-138687343 from which I'm extracting the relevant sections: 
> At first I really couldn't understand why `parseInt` is being used like it is in `XRef.fetchUncompressed`, since the `num` and `gen` properties of an object reference should _always_ be integers.
> However, doing a bit of code archaeology pointed to PR [#4348 (comment)](https://github.com/mozilla/pdf.js/pull/4348#issuecomment-36367149), and it thus seem that this was a very deliberate change. Since I didn't want to inadvertently introduce any regressions, I've kept the `parseInt` calls intact but moved them to occur _only_ when actually necessary.[1]
>
> [1] I even ran all tests locally, with an added `assert(Number.isInteger(obj1) && Number.isInteger(obj2));` check, and everything passed with flying colours.
> However, since it appears that this was in fact necessary at one point, one possible explanation is that the failing test-case(s) has now been replaced by reduced ones.

timvandermeij:
The original inclusion of this code seemed to be necessary for the `ichiji` reference test as far as I can tell from https://github.com/mozilla/pdf.js/pull/4348#issuecomment-36367149. In https://github.com/mozilla/pdf.js/commits/66ffdc4c5b63a135f201508264761fce17c08059/test/pdfs/ichiji.pdf.link I don't see anything changed about this file, so given that the `ichiji` reference test still passed now when the code is removed and that the original file doesn't appear to be changed, I would say it's safe to just get rid of the code altogether.

Snuffleupagus:
> The original inclusion of this code seemed to be necessary for the `ichiji` reference test as far as I can tell from [#4348 (comment)](https://github.com/mozilla/pdf.js/pull/4348#issuecomment-36367149).

You may be right; I wasn't really able to tell if/how that particular test-case was the culprit, based on my reading of that comment :-)

> I would say it's safe to just get rid of the code altogether.

Let's just try it, the worst thing that will happen is that there'll be a test-case available in a couple of weeks/months :P

timvandermeij (approved):

timvandermeij:
Looks good!

Issue #8802, opened 2017-08-21
Zoom is too sensitive on OS X


staktrace:
(Filing this as an upstream bug for https://bugzilla.mozilla.org/show_bug.cgi?id=1392361)

Link to PDF file (or attach file here): https://www.library.yorku.ca/spark/books_journals_more/PARCA%20Test_Final.pdf (or any PDF)

Configuration:
- Web browser and its version: Firefox 55+
- Operating system and its version: OS X 10.11.6
- PDF.js version: Whatever's in Firefox now
- Is an extension: no?

Steps to reproduce the problem:
1. Load PDF
2. Pinch zoom on macbook trackpad, or scroll using ctrl+two finger pan

What is the expected behavior? (add screenshot)
Page zooms gradually

What went wrong? (add screenshot)
Page zooms far too much

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

lizzard:
I'm also seeing this problem on MacOS 10.12.6 with default zoom/pinch system settings.

schneeschmelze:
For the record: No change in Firefox 60. :(

davidanderle:
I'm having the same issue on Windows 10. Zooming with pinch is too sensitive.

dimonoid:
Windows 10 1703, still a the same. Carbon X1 4th gen

lxieyang:
Same on macOS 10.14. Too sensitive.

Nariety:
Firefox 68.0b3, the issue still persists. The pdf touchpad zooming is too sensitive to the point where it's unusable.

bubblefoil:
I may have a related issue. I use a Logitech mouse with smooth scroll feature. With Logitech SetPoint extension the browser correctly interprets mouse wheel events with decimal deltaY, e.g. 0.025, and scrolls perfectly smoothly. It also handles ctrl-wheel zoom quite well. But PDF viewer does not. It does either nothing or zooms way too much.
I've tried logging wheel events. Indeed, there is many of them with small delta.

dimonoid:
Can confirm. It is still an issue on ThinkPad x1 carbon 4th gen. Whel using either track point or trackpad. When I use mouse wheel instead, it works just fine.

namasaki:
still not fixed

locness3:
Yeah, still present, firefox 75

ViacheslavLabs:
Still not fixed, firefox 78

[...]

PR #11365, opened 2019-11-27
Add a heuristic, for mouse wheel events, to avoid the document toggling between the min/max zoom levels (issue 8802)

Snuffleupagus:
This is obviously not a *proper* solution, since I've got no idea why this is even a problem (and don't have the required hardware to debug it either).
However, it seems that it should be reasonably easy to at least add a workaround for now, which surely is better than the current state.

*Submitting as a "Draft PR" for now, since we'd need people affected by this to test if this even helps before moving forward with it.*

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -51,6 +51,7 @@ const DEFAULT_SCALE_DELTA = 1.1;
 const DISABLE_AUTO_FETCH_LOADING_BAR_TIMEOUT = 5000; // ms
 const FORCE_PAGES_LOADED_TIMEOUT = 10000; // ms
 const WHEEL_ZOOM_DISABLED_TIMEOUT = 1000; // ms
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
@staktrace Would you be able to test the preview build, see http://54.67.70.0:8877/48611426867ac14/web/viewer.html, to see if it helps with issue #8802?

staktrace:
@Snuffleupagus I tried it, and it doesn't really help - it might be marginally better but from a user's point of view controlling the zoom via trackpad is still effectively impossible.

Snuffleupagus:
> [...] it might be marginally better but from a user's point of view controlling the zoom via trackpad is still effectively impossible.

That's unfortunate, but not really all that surprising; thanks for taking the time to test this!

PR #11373, opened 2019-11-30
Slightly simplify the XRef cache lookup in `XRef.fetch`

Snuffleupagus:
Note that the XRef cache will only hold objects returned through `Parser.getObj`, and indirectly via `Lexer.getObj`. Since neither of those methods will ever return `undefined`, we can simply `assert` that when inserting objects into the cache and thus get rid of one function call when doing cache lookups.

Obviously this won't have a huge effect on performance, however `XRef.fetch` is usually called *a lot* in larger documents and this patch thus cannot hurt.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -1636,8 +1636,11 @@ var XRef = (function XRefClosure() {
       }
       const num = ref.num;
 
-      if (this._cacheMap.has(num)) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks like a useful change to me!

PR #11374, opened 2019-12-01
Set the first `pdfPage` immediately in `{BaseViewer, PDFThumbnailViewer}.setDocument`

Snuffleupagus:
*This patch is simple enough that I almost feel like I'm overlooking some trivial reason why this would be a bad idea.*

Note how in `{BaseViewer, PDFThumbnailViewer}.setDocument` we're always getting the *first* `pdfPage` in order to initialize all pages/thumbnails.
However, once that's done the first `pdfPage` is simply ignored and rendering of the first page thus requires calling `PDFDocumentProxy.getPage` yet again. (And in the `BaseViewer` case, it's even done once more after `onePageRenderedCapability` is resolved.)

All in all, I cannot see why we cannot just immediately set the first `pdfPage` and thus avoid an early round-trip to the API in the `_ensurePdfPageLoaded` method before rendering can begin.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -421,9 +421,9 @@ class BaseViewer {
 
     // Fetch a single page so we can get a viewport that will be the default
     // viewport for all pages
-    firstPagePromise.then((pdfPage) => {
+    firstPagePromise.then((firstPdfPage) => {
       let scale = this.currentScale;
-      let viewport = pdfPage.getViewport({ scale: scale * CSS_UNITS, });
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
> However, once that's done the first pdfPage is simply ignored

I'm also a bit puzzled by this and I can't imagine why we wouldn't use it if it's already loaded. The patch looks good to me; thanks!

PR #11382, opened 2019-12-04
Fix an incorrect condition in `BaseViewer.isPageVisible` (PR 10217 follow-up)

Snuffleupagus:
This was a blatant oversight in PR #10217, since there's obviously no `this.pageNumber` property anywhere in the `BaseViewer`. Luckily this shouldn't have caused any bugs, since the only call-site is also validating the `pageNumber` (but correctly that time).

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -935,7 +935,7 @@ class BaseViewer {
     if (!this.pdfDocument) {
       return false;
     }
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Funny how we both overlooked this. Thank you for the follow-up fix!

PR #11387, opened 2019-12-05
Handle corrupt ASCII85Decode inline images with truncated EOD markers (issue 11385)

### LINKED ISSUE #11385: Specific PDF file does not render correctly
Configuration:
- Web browser and its version: Chrome 78
- PDF.js version: your demo (https://mozilla.github.io/pdf.js/web/viewer.html) and also our .net implemented library (latest)

We have a PDF file and it does not render correctly in pdfJs.
It looks like it has rendered just a part of the page.  When I open the same PDF file with PDF reader, it looks ok.
I have also checked the PDF file with PDF validator and it says that the file is 100% valid. 
There are no errors in JS console.

Can you please let me know, what causes the problem and how to solve it?

PDF will be sent through secure exchange. Please, let me know, how that works.

Snuffleupagus:
In the PDF document in question, there's an ASCII85Decode inline image where the '>' part of EOD (end-of-data) marker is missing; hence the PDF document is corrupt.

Fixes #11385

### DIFF
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -344,6 +344,8 @@ class Parser {
     let startPos = stream.pos, ch, length;
     while ((ch = stream.getByte()) !== -1) {
       if (ch === TILDE) {
+        const tildePos = stream.pos;
+
         ch = stream.peekByte();
         // Handle corrupt PDF documents which contains whitespace "inside" of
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Nice work! Let's make reference images.

/botio makeref

PR #11388, opened 2019-12-05
Remove the `viewer` option from the `PDFPresentationMode` constructor

Snuffleupagus:
The `viewer` option was *only* used for checking that a document is loaded in `PDFPresentationMode.request`, however that's just as easy to do by simply utilizing `BaseViewer.pagesCount` instead and this way we can also avoid the DOM lookup.


timvandermeij (approved):

timvandermeij:
Nice decoupling!

PR #11390, opened 2019-12-07
Ensure that `PDFDocument.checkFirstPage` waits for cleanup to complete (PR 10392 follow-up)

Snuffleupagus:
Given how this method is currently used there shouldn't be any fonts loaded at the point in time where it's called, but it does seem like a bad idea to assume that that's always going to be the case. Since `PDFDocument.checkFirstPage` is already asynchronous, it's easy enough to simply await `Catalog.cleanup` here.

(The patch also makes a tiny simplification in a loop in `Catalog.cleanup`.)

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -664,13 +664,13 @@ class PDFDocument {
   }
 
   checkFirstPage() {
-    return this.getPage(0).catch((reason) => {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #11391, opened 2019-12-08
[api-minor] Replace `globalScope` with the standard `globalThis` property instead

Snuffleupagus:
Please see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis and note that most (reasonably) modern browsers have supported this for a while now, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#Browser_compatibility

Since ESLint doesn't support this new global yet, it was added to the `globals` list in the top-level configuration file to prevent issues.

Finally, for older browsers a polyfill was added in `ssrc/shared/compatibility.js`.

---

*Edit:* Adding `[api-minor]` here just in case, since I suppose that these changes may be significant (especially the second commit).


Snuffleupagus:
The test "failures" are, most likely, caused by a Firefox update on the bots.

src/shared/compatibility.js
```diff
@@ -14,13 +14,18 @@
...
-
 // Skip compatibility checks for modern builds and if we already ran the module.
 if ((typeof PDFJSDev === 'undefined' || !PDFJSDev.test('SKIP_BABEL')) &&
-    !globalScope._pdfjsCompatibilityChecked) {
+    (typeof globalThis === 'undefined' ||
+     !globalThis._pdfjsCompatibilityChecked)) {
 
-globalScope._pdfjsCompatibilityChecked = true;
+// Provides support for globalThis in legacy browsers.
+// Support: IE.
```
timvandermeij: This should also include Edge and Opera since both don't support this feature and both browsers are supported according to our FAQ.

src/shared/compatibility.js
```diff
@@ -223,11 +228,11 @@ const isIE = /Trident/.test(userAgent);
     // need to be polyfilled for the IMAGE_DECODERS build target.
     return;
   }
-  if (globalScope.Promise && (globalScope.Promise.prototype &&
-                              globalScope.Promise.prototype.finally)) {
+  if (globalThis.Promise && (globalThis.Promise.prototype &&
+                              globalThis.Promise.prototype.finally)) {
```
timvandermeij: One space of indentation too much here.

timvandermeij:
> The test "failures" are, most likely, caused by a Firefox update on the bots.

Yes, I'll make new images once this patch is merged.

src/shared/compatibility.js
Snuffleupagus: Sure, fixed now :-)

src/shared/compatibility.js
Snuffleupagus: Whoops, nice catch; although I'm going to blame the search-and-replace feature of my editor for nothing handling that case correctly :-P

timvandermeij:
Thank you!

PR #11398, opened 2019-12-09
Attempt to improve the `PDFDocument` error message for empty files (issue 5887)

### LINKED ISSUE #5887: Better error message for zero-sized file
EDIT: It turns out that the file is really broken: It is a size-zero file! I think there should be a better message if the file is zero or doesn't start with the "%PDF" marker. The current mesage implies that PDF.js has an issue - while the file has the issue.

That's with PDF.js as shipping with Firefox 36 under Linux.

The PDF file under http://www.bundesnetzagentur.de/cae/servlet/contentblob/38252/publicationFile/2591/Mitteilung902005VerzeichniId1323pdf.pdf doesn't render at all - only the dark background and is shown. The web console shows:

"Error: stream must have data" pdf.worker.js:250:4
"assert@resource://pdf.js/build/pdf.worker.js:267:5
init@resource://pdf.js/build/pdf.worker.js:2312:5
PDFDocument@resource://pdf.js/build/pdf.worker.js:2303:7
NetworkPdfManager@resource://pdf.js/build/pdf.worker.js:1975:24
getPdfManager@resource://pdf.js/build/pdf.worker.js:32962:24
wphSetupDoc@resource://pdf.js/build/pdf.worker.js:33157:7
messageHandlerComObjOnMessage@resource://pdf.js/build/pdf.worker.js:1219:9
" pdf.worker.js:252:0
"Warning: Unsupported feature "unknown"" pdf.worker.js:235:4
"Warning: Unsupported feature "unknown"" pdf.js:235:4

PDF.js Version 1.0.937 (build: 308646d)

Snuffleupagus:
Given that the error in question is surfaced on the API-side, this patch makes the following changes:
 - Updates the wording such that it'll hopefully be slightly easier for users to understand.
 - Changes the plain `Error` to an `InvalidPDFException` instead, since that should work better with the existing Error handling.
 - Adds a unit-test which loads an empty PDF document (and also improves a pre-existing `InvalidPDFException` message and its test-case).

Fixes #5887

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -15,9 +15,9 @@
 /* eslint no-var: error */
 
 import {
-  assert, bytesToString, FormatError, info, isArrayBuffer, isArrayEqual, isBool,
-  isNum, isSpace, isString, OPS, shadow, stringToBytes, stringToPDFString, Util,
-  warn
+  assert, bytesToString, FormatError, info, InvalidPDFException, isArrayBuffer,
+  isArrayEqual, isBool, isNum, isSpace, isString, OPS, shadow, stringToBytes,
+  stringToPDFString, Util, warn
 } from '../shared/util';
 import { Catalog, ObjectLoader, XRef } from './obj';
 import { Dict, isDict, isName, isStream, Ref } from './primitives';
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
That should indeed be a more clear message. Thanks, also for adding a unit test!

PR #11406, opened 2019-12-13
Re-factor the `find` helper function, in `src/core/document.js`, to search through the raw bytes rather than a string

Snuffleupagus:
During initial parsing of every PDF document we're currently creating a few `1 kB` strings, in order to find certain commands needed for initialization.
This seems inefficient, not to mention completely unnecessary, since we can just as well search through the raw bytes directly instead (similar to other parts of the code-base). One small complication here is the need to support backwards search, which does add some amount of "duplication" to this function.

The main benefits here are:
 - No longer necessary to allocate *temporary* `1 kB` strings during initial parsing, thus saving some memory.
 - In practice, for well-formed PDF documents, the number of iterations required to find the commands are usually very low. (For the `tracemonkey.pdf` file, there's a *total* of only 30 loop iterations.)

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -15,8 +15,8 @@
 /* eslint no-var: error */
 
 import {
-  assert, bytesToString, FormatError, info, InvalidPDFException, isArrayBuffer,
-  isArrayEqual, isBool, isNum, isSpace, isString, OPS, shadow, stringToBytes,
+  assert, FormatError, info, InvalidPDFException, isArrayBuffer, isArrayEqual,
+  isBool, isNum, isSpace, isString, OPS, shadow, stringToBytes,
   stringToPDFString, Util, warn
 } from '../shared/util';
 import { Catalog, ObjectLoader, XRef } from './obj';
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
I ran `SKIP_BABEL=true gulp test` locally with the diff below applied on top of the patch, to check that no edge-cases inadvertently broke, and all unit/reference tests passed as expected.
```diff
diff --git a/src/core/document.js b/src/core/document.js
index b58ea7b4..4e8118a5 100644
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -15,8 +15,8 @@
 /* eslint no-var: error */
 
 import {
-  assert, FormatError, info, InvalidPDFException, isArrayBuffer, isArrayEqual,
-  isBool, isNum, isSpace, isString, OPS, shadow, stringToBytes,
+  assert, bytesToString, FormatError, info, InvalidPDFException, isArrayBuffer,
+  isArrayEqual, isBool, isNum, isSpace, isString, OPS, shadow, stringToBytes,
   stringToPDFString, Util, warn
 } from '../shared/util';
 import { Catalog, ObjectLoader, XRef } from './obj';
@@ -363,6 +363,10 @@ function find(stream, signature, limit = 1024, backwards = false) {
   const scanBytes = stream.peekBytes(limit);
   const scanLength = scanBytes.length - signatureLength;
 
+  const str = bytesToString(stream.peekBytes(limit));
+  const index = backwards ? str.lastIndexOf(bytesToString(signature)) :
+                            str.indexOf(bytesToString(signature));
+
   if (scanLength <= 0) {
     return false;
   }
@@ -377,6 +381,8 @@ function find(stream, signature, limit = 1024, backwards = false) {
         j++;
       }
       if (j >= signatureLength) { // `signature` found.
+        assert(index === (pos - signatureEnd), 'backwards');
+
         stream.pos += (pos - signatureEnd);
         return true;
       }
@@ -390,6 +396,8 @@ function find(stream, signature, limit = 1024, backwards = false) {
         j++;
       }
       if (j >= signatureLength) { // `signature` found.
+        assert(index === pos, 'forwards');
+
         stream.pos += pos;
         return true;
       }
```

Snuffleupagus:
I ran `SKIP_BABEL=true gulp test` locally with the diff below applied on top of the patch, to check that no edge-case inadvertently broke, and all unit/reference tests passed as expected.

```diff
diff --git a/src/core/document.js b/src/core/document.js
index b58ea7b4..07136d36 100644
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -15,8 +15,8 @@
 /* eslint no-var: error */
 
 import {
-  assert, FormatError, info, InvalidPDFException, isArrayBuffer, isArrayEqual,
-  isBool, isNum, isSpace, isString, OPS, shadow, stringToBytes,
+  assert, bytesToString, FormatError, info, InvalidPDFException, isArrayBuffer,
+  isArrayEqual, isBool, isNum, isSpace, isString, OPS, shadow, stringToBytes,
   stringToPDFString, Util, warn
 } from '../shared/util';
 import { Catalog, ObjectLoader, XRef } from './obj';
@@ -363,7 +363,13 @@ function find(stream, signature, limit = 1024, backwards = false) {
   const scanBytes = stream.peekBytes(limit);
   const scanLength = scanBytes.length - signatureLength;
 
+  const str = bytesToString(stream.peekBytes(limit));
+  const index = backwards ? str.lastIndexOf(bytesToString(signature)) :
+                            str.indexOf(bytesToString(signature));
+
   if (scanLength <= 0) {
+    assert(index === -1, 'empty');
+
     return false;
   }
   if (backwards) {
@@ -377,6 +383,8 @@ function find(stream, signature, limit = 1024, backwards = false) {
         j++;
       }
       if (j >= signatureLength) { // `signature` found.
+        assert(index === (pos - signatureEnd), 'backwards');
+
         stream.pos += (pos - signatureEnd);
         return true;
       }
@@ -390,12 +398,16 @@ function find(stream, signature, limit = 1024, backwards = false) {
         j++;
       }
       if (j >= signatureLength) { // `signature` found.
+        assert(index === pos, 'forwards');
+
         stream.pos += pos;
         return true;
       }
       pos++;
     }
   }
+  assert(index === -1, 'not found');
+
   return false;
 }
```

timvandermeij (approved):

timvandermeij:
Looks like a good change to me. It's indeed a bit wasteful to convert to strings first. Thanks!

PR #11411, opened 2019-12-16
created a variable to hold the completed default preferences

dhuang612:
created a new variable:
```
defaultPreferences
```

set it underneath the task create default_preferences to be the completed run of this function.
setup under task `lint` & `lib` the new variable

ran npm test and confirmed that it completed without issues.
Please let me know if there are any other changes that need to be made!

Related to #11414.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -567,18 +567,21 @@ gulp.task('default_preferences-pre', function() {
     .pipe(gulp.dest(DEFAULT_PREFERENCES_DIR + 'lib/'));
 });
 
-gulp.task('default_preferences', gulp.series('default_preferences-pre',
-    function(done) {
+function default_preferences(done) {
   var AppOptionsLib =
     require('./' + DEFAULT_PREFERENCES_DIR + 'lib/web/app_options.js');
   var AppOptions = AppOptionsLib.AppOptions;
   var OptionKind = AppOptionsLib.OptionKind;
-
   createStringSource('default_preferences.json', JSON.stringify(
-      AppOptions.getAll(OptionKind.PREFERENCE), null, 2))
+    AppOptions.getAll(OptionKind.PREFERENCE), null, 2))
     .pipe(gulp.dest(BUILD_DIR))
     .on('end', done);
-}));
+}
+
+gulp.task('generatePreferences', gulp.series('default_preferences-pre',
+default_preferences), function(done) {
+  done();
+});
 
 gulp.task('locale', function () {
   var VIEWER_LOCALE_OUTPUT = 'web/locale/';
@@ -681,7 +684,7 @@ function preprocessHTML(source, defines) {
 
 // Builds the generic production viewer that should be compatible with most
 // modern HTML5 browsers.
-gulp.task('generic', gulp.series('buildnumber', 'default_preferences', 'locale',
+gulp.task('generic', gulp.series('buildnumber', 'generatePreferences', 'locale',
                                  function() {
   console.log();
   console.log('### Creating generic viewer');
@@ -749,7 +752,7 @@ gulp.task('image_decoders', gulp.series('buildnumber', function() {
   return createImageDecodersBundle(defines).pipe(gulp.dest(IMAGE_DECODERS_DIR));
 }));
 
-gulp.task('minified-pre', gulp.series('buildnumber', 'default_preferences',
+gulp.task('minified-pre', gulp.series('buildnumber', 'generatePreferences',
                                       'locale', function() {
   console.log();
   console.log('### Creating minified viewer');
@@ -847,7 +850,7 @@ function preprocessDefaultPreferences(content) {
           content + '\n');
 }
 
-gulp.task('mozcentral-pre', gulp.series('buildnumber', 'default_preferences',
+gulp.task('mozcentral-pre', gulp.series('buildnumber', 'generatePreferences',
                                         function() {
   console.log();
   console.log('### Building mozilla-central extension');
@@ -898,7 +901,7 @@ gulp.task('mozcentral-pre', gulp.series('buildnumber', 'default_preferences',
 
 gulp.task('mozcentral', gulp.series('mozcentral-pre'));
 
-gulp.task('chromium-pre', gulp.series('buildnumber', 'default_preferences',
+gulp.task('chromium-pre', gulp.series('buildnumber', 'generatePreferences',
                                       'locale', function() {
   console.log();
   console.log('### Building Chromium extension');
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
This honestly doesn't seem like an appropriate solution, since running any particular `gulp` task by default *every* time that the gulpfile is loaded quite frankly seems incorrect.

The correct approach would probably be something along these lines instead: https://github.com/gulpjs/gulp/issues/1392

dhuang612:
So,

I've reviewed the link you shared.

The way they  got rid of the second call was to setup a new task that called all the other functions at once together.

I tried to setup something similar however, the newly created task never runs?
```
gulp.task('createPreferences', gulp.series('default_preferences', gulp.series('default_preferences-pre'), gulp.parallel('lib', 'lint') ));
```
I have set this up underneath the tasks lint and lib but it's giving me an error related to the settings.

timvandermeij:
I think the problem is outlined well in https://github.com/gulpjs/gulp/issues/1392#issuecomment-157161298. Our code is converted from Gulp 3, where we had the concept of dependencies in the API, to Gulp 4 which doesn't have this concept in the API anymore, without taking this into account, so everything we now treat as a dependency is actually executed multiple times. Normally that's not really a problem aside from increased build times, but in the case of the default preferences it is.

Hence we should look at this in a broader sense: how do we map Gulp 3's dependency model onto Gulp 4? The linked issue should give some ideas about how to do this, but there is also https://fettblog.eu/gulp-4-parallel-and-series/ which seems to explain it well.

dhuang612:
created a task to generate the default_preferences

then a second task under **lib** and **lint** that calls the series on first task and parallel on **lib** and **lint**
```
[23:21:54] Starting 'createPreferences'...


[23:22:00] Finished 'createPreferences' after 7.36 s

[23:22:00] Finished 'createPreferences' after 5.63 s
```

createPreference is started once and finishes twice once for lib and one for lint

dhuang612:
so the tests are completely locally on my side when I use nvm with node v8.16.0
Is there a reason that travis ci file is setup to use node v13.5?

Snuffleupagus:
It's very difficult to understand/follow these changes, and looking at the long list of commits I'm sorry to say that it does seem a bit like you've just tried a bunch of things without perhaps a clear plan as to the end result. (There's at least *one* newly added `gulp` task that doesn't even appear to be used.)

Furthermore, I'm not sure that these changes are that close to addressing the underlying issue, since the whole point here is that the 'default_preferences' task should only run *once* for each `npm test` invocation (which will most likely require more invasive changes, as alluded to in https://github.com/mozilla/pdf.js/pull/11411#issuecomment-566785830). All in all, it might be that some prior familiarity with `gulp` is necessary to more easily make progress on the issue.

dhuang612:
I was pushing a number of commits because things were passing locally on my computer but when they hit you're side it was failing. I'm following the guide you provided in which they separate out the individual tasks into functions and then create a task to run the function as needed.

dhuang612:
to confirm that default_preferences function is only running once, included a console.log at the top
Screen Shot 2019-12-18 at 4 00 46 PM

package-lock.json
```diff
@@ -1977,9 +1977,9 @@
       },
```
timvandermeij: All changes in this file should be reverted.

gulpfile.js
```diff
@@ -1289,7 +1305,7 @@ gulp.task('web', gulp.series('generic', 'jsdoc', 'gh-pages-prepare',
                              'wintersmith', 'gh-pages-git'));
 
 gulp.task('dist-pre', gulp.series('generic', 'components', 'image_decoders',
-                                  'lib', 'minified', function() {
+                                      'lib', 'minified', function() {
```
timvandermeij: Unrelated change that should be reverted.

gulpfile.js
```diff
@@ -1153,10 +1160,9 @@ gulp.task('unittestcli', gulp.series('testing-pre', 'lib', function(done) {
   });
 }));
 
-gulp.task('lint', gulp.series('default_preferences', function(done) {
+function genLint(done) {
   console.log();
   console.log('### Linting JS files');
-
```
timvandermeij: Unrelated change that should be reverted.

gulpfile.js
```diff
@@ -567,18 +567,21 @@ gulp.task('default_preferences-pre', function() {
...
-    function(done) {
+function default_preferences(done) {
   var AppOptionsLib =
     require('./' + DEFAULT_PREFERENCES_DIR + 'lib/web/app_options.js');
   var AppOptions = AppOptionsLib.AppOptions;
   var OptionKind = AppOptionsLib.OptionKind;
-
   createStringSource('default_preferences.json', JSON.stringify(
-      AppOptions.getAll(OptionKind.PREFERENCE), null, 2))
+    AppOptions.getAll(OptionKind.PREFERENCE), null, 2))
```
timvandermeij: Unrelated change that should be reverted.

timvandermeij (changes requested):
I'm also having a hard time understanding why this actually fixes things, because as far as I can tell we still run the task multiple times, but with more indirection. Why does this actually make it run once?

In general, try to keep the patch limited to only required changes because that makes it much easier to review. Changes like whitespace and `package-lock.json` can be removed. Thanks.

Snuffleupagus:
> I'm also having a hard time understanding why this actually fixes things, because as far as I can tell we still run the task multiple times, but with more indirection. Why does this actually make it run once?

As I've said elsewhere, this probably isn't really a good beginner bug unfortunately.

dhuang612:
Isn't this issue already closed by you're commit?
Fixed by #11413

so I can just close this right?

timvandermeij:
The intermittent failure is fixed by that, but the underlying issue is not. We have made a separate ticket to fix that, but as indicated this is a bit harder than we expected, so let's indeed close this and come up with a good plan for that. Thanks.

dhuang612:
> > I'm also having a hard time understanding why this actually fixes things, because as far as I can tell we still run the task multiple times, but with more indirection. Why does this actually make it run once?
> 
> As I've said elsewhere, this probably isn't really a good beginner bug unfortunately.

I mean I wish it could have been properly labelled then since I spent almost two days trying?

timvandermeij:
Yes, unfortunately sometimes it's not entirely clear how hard it will be until it's actually being implemented, which was the case here. Sorry for that.

dhuang612:
Also, 
@Snuffleupagus 
I just wanted to leave a comment that if someone on you're side is going to take over an issue, it seems like common courtesy to notify the person who currently trying to fix the problem?

When I asked to work on this issue, it was assigned as a beginner issue with no priority noted for how quickly this needed to be completed. I feel that at 2 days it was still acceptable for me to be trying to resolve this and it shouldn't have been ripped away from me. 

Thanks

PR #11413, opened 2019-12-18
Re-factor the `npm test` command, used by Travis, to avoid running the 'default_preferences' tasks concurrently (issue 10732)

Snuffleupagus:
*Please note:* This patch does *not* prevent the 'default_preferences' task from running more than once during `npm test`, but it does ensure that the tasks won't run *concurrently* by running the relevant tests in *series*.

While it would obviously still make sense to re-factor the gulpfile to account for changes in `gulp` version 4, by at least tweaking the `npm test` command the intermittent failures on Travis should at least go away.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1502,3 +1502,6 @@ gulp.task('externaltest', function (done) {
                 { stdio: 'inherit', });
   done();
 });
+
+gulp.task('npm-test', gulp.series(gulp.parallel('lint', 'externaltest'),
+                                  'unittestcli'));
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "yargs": "^11.1.1"
   },
   "scripts": {
-    "test": "env SKIP_BABEL=true gulp lint unittestcli externaltest"
+    "test": "env SKIP_BABEL=true gulp npm-test"
   },
   "repository": {
     "type": "git",


timvandermeij:
I agree. It would be helpful to at least get rid of the intermittent failures and fix the issues regarding Gulp 4 changes in a separate scope.

Snuffleupagus:
> I agree. It would be helpful to at least get rid of the intermittent failures and fix the issues regarding Gulp 4 changes in a separate scope.

Yeah, that was my thought (since a proper Gulp 4 migration is a larger undertaking); and I had trouble following all the changes in the other PR.

Snuffleupagus:
> I agree. It would be helpful to at least get rid of the intermittent failures and fix the issues regarding Gulp 4 changes in a separate scope.

Yeah, that was my thought (since a proper Gulp 4 migration is a larger undertaking); and I had trouble following all the changes in the other PR.

timvandermeij (approved):

Snuffleupagus:
> I agree. It would be helpful to at least get rid of the intermittent failures and fix the issues regarding Gulp 4 changes in a separate scope.

Yeah, that was my thought as well (since a proper Gulp 4 migration is a larger undertaking); and I had trouble following all the changes in the other PR (and it doesn't seem to prevent the concurrency issue either).

timvandermeij:
Thank you for improving the situation here!

PR #9904, opened 2018-07-20
Modified Examples to work without systemjs

RonLek:
Closes #9784 

@timvandermeij  @Snuffleupagus  Please review for changes, if any. Thanks!

PR #11418, opened 2019-12-19
[api-minor] Simplify the *fallback* fake worker loader code in `src/display/api.js`

Snuffleupagus:

### DIFF
--- a/examples/browserify/main.js
+++ b/examples/browserify/main.js
@@ -5,7 +5,7 @@
 
 var pdfjsLib = require('pdfjs-dist');
 
-var pdfPath = '../helloworld/helloworld.pdf';
+var pdfPath = '../learning/helloworld.pdf';
 
 // Setting worker path to worker bundle.
 pdfjsLib.GlobalWorkerOptions.workerSrc =
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -12,7 +12,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* globals __non_webpack_require__ */
 /* eslint no-var: error */
 
 /**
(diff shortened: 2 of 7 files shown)


src/display/api.js
```diff
@@ -1534,8 +1534,21 @@ const PDFWorker = (function PDFWorkerClosure() {
         return worker.WorkerMessageHandler;
       }
       if ((typeof PDFJSDev !== 'undefined' && PDFJSDev.test('GENERIC')) &&
+          // eslint-disable-next-line no-undef
           (isNodeJS && typeof __non_webpack_require__ === 'function')) {
-        const worker = __non_webpack_require__(getWorkerSrc());
```
timvandermeij: Now that we don't use `__non_webpack_require__` in this `if` body anymore, I think we can also remove it from the `if` statement, together with its ESLint ignore line. If I'm not mistaken the same applies a few lines above.

timvandermeij: Considering this I also don't think it can hurt to do an extra check here. It's also not performance-heavy code.

examples/browserify/main.js
```diff
@@ -5,7 +5,7 @@
 
 var pdfjsLib = require('pdfjs-dist');
 
-var pdfPath = '../helloworld/helloworld.pdf';
```
Snuffleupagus: It also interesting that this example, and the one below, has been "broken" ever since PR #9904 without anyone noticing (more than a year ago). This would suggest, at least to me, that these examples see very limited use and that cleaning up the fake worker loader code thus isn't a bad idea.

timvandermeij (approved):

timvandermeij:
Looks much simpler this way. I always found this part of the code base to be very hard to read, so thank you for simplifying this!

Snuffleupagus:
> I always found this part of the code base to be hard to read, so thank you for simplifying this!

Completely agree regarding the "hard to read" part; and *hopefully* people aren't using fake workers anyway :-)
I also really hope that the `eval` usage won't start causing trouble in Node.js, but I purposely put that in its own commit to make finding/reverting it easier later on.

Snuffleupagus:
@timvandermeij Also, I forgot to say thank you for landing this :-D

Given that there's been a couple of `api-minor` changes lately, would now perhaps be a good time to bump the minor version number and create a new release?

timvandermeij:
Yes, I'll plan a release for next week. It has been a while since the last release, so it sounds like a good plan.

PR #11422, opened 2019-12-21
Use the `strict` mode `assert` in the pdf2png Node.js example (issue 10768)

### LINKED ISSUE #10768: assert: when using the strict mode, any assert function will use the equality used in the strict function mode
Why you stiil use assert without strict mode?
[https://github.com/mozilla/pdf.js/blob/master/examples/node/pdf2png/pdf2png.js](url)

**NodeJs documentation:**

> Legacy mode#
> Stability: 0 - Deprecated: Use strict mode instead.
> When accessing assert directly instead of using the strict property, the Abstract Equality Comparison will be used for any function without "strict" in its name, such as assert.deepEqual().

Snuffleupagus:
See https://nodejs.org/api/assert.html#assert_strict_mode

Should, hopefully, fix #10768

### DIFF
--- a/examples/node/pdf2png/pdf2png.js
+++ b/examples/node/pdf2png/pdf2png.js
@@ -14,7 +14,7 @@
  */
 
 var Canvas = require('canvas');
-var assert = require('assert');
+var assert = require('assert').strict;
 var fs = require('fs');
 
 function NodeCanvasFactory() {}


timvandermeij (approved):

timvandermeij:
Thanks!

PR #11423, opened 2019-12-21
Ignore Metadata entries with incorrectly encoded characters, when setting the viewer title (bug 1605526)

Snuffleupagus:
Apparently Ghostscript can, in some cases, generate/include `Metadata` with incorrectly encoded characters.[1] This results in the viewer title looking wrong, which we thus attempt to avoid by falling back to the `Info` entry instead.

*Please note:* Obviously it would be better if this was fixed in the `Metadata` parser instead, rather than using a viewer work-around, but I'm just not sure how or even *if* that could actually be done given that the `Metadata` stream contains no trace of the *original* character.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1605526

---
[1] The problematic characters are from the Specials Unicode block, see https://en.wikipedia.org/wiki/Specials_(Unicode_block)

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1126,7 +1126,11 @@ let PDFViewerApplication = {
       if (metadataTitle) {
         // Ghostscript can produce invalid 'dc:title' Metadata entries:
         //  - The title may be "Untitled" (fixes bug 1031612).
-        if (metadataTitle !== 'Untitled') {
+        //  - The title may contain incorrectly encoded characters, which thus
+        //    looks broken, hence we ignore the Metadata entry when it contains
+        //    characters from the Specials Unicode block (fixes bug 1605526).
+        if (metadataTitle !== 'Untitled' &&
+            !/[\uFFF0-\uFFFF]/g.test(metadataTitle)) {
           pdfTitle = metadataTitle;
         }
       }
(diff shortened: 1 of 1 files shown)


web/app.js
```diff
@@ -1117,16 +1117,25 @@ let PDFViewerApplication = {
...
+      const metadataTitle = metadata && metadata.get('dc:title');
+      if (metadataTitle) {
+        // Ghostscript can produce invalid 'dc:title' Metadata entries:
+        //  - The title may be "Untitled" (fixes bug 1031612).
+        //  - The title may contain incorrectly encoded characters, which thus
+        //    looks broken, hence we ignore the Metadata entry when it contains
+        //    characters from the Specials Unicode block (fixes bug 1605526).
+        //
+        //    (To avoid displaying no viewer title at all, always fallback to
+        //     the Metadata entry regardless when no valid Info entry exists.)
```
Snuffleupagus: I could easily be convinced to remove this special-case, which was *only* added in an attempt to not change the current behaviour too much!

Snuffleupagus: I could easily be convinced to remove this special-case, which was *only* added in an attempt to not change the current behaviour too much!

Snuffleupagus: Having thought about this some more I've decided to just skip the special handing for the `!pdfTitle` case, since the viewer title really seems like a best effort solution anyway with no guarantees that it'll be perfect.

timvandermeij (approved):

timvandermeij:
I also looked at the XMP data and couldn't find the original character too, so I think this is the only possible solution here. Thank you for fixing this, and for mentioning the bug IDs for future reference!

PR #11424, opened 2019-12-22
Rewrite fonts_utils.js to use const

wojtekmaj:


src/shared/fonts_utils.js
```diff
@@ -413,16 +413,16 @@ function writeToFile(aBytes, aFilePath) {
   }
 
   netscape.security.PrivilegeManager.enablePrivilege('UniversalXPConnect');
-  var Cc = Components.classes,
+  const Cc = Components.classes,
       Ci = Components.interfaces;
```
Snuffleupagus: Here, and in a lot of the other PRs, you'll need to fix the indentation of lines such as these (since the alignment is now wrong).

Snuffleupagus (changes requested):

src/shared/fonts_utils.js
```diff
@@ -220,8 +220,8 @@ function readFontIndexData(aStream, aIsByte) {
     throw new FormatError(offsize + ' is not a valid offset size');
   }
 
-  var offsets = [];
-  var i;
+  const offsets = [];
+  let i;
```
timvandermeij: We should inline this variable in the `for` loops as done above and in other files so we can remove it since it's very local for the loops.

src/shared/fonts_utils.js
```diff
@@ -413,16 +413,16 @@ function writeToFile(aBytes, aFilePath) {
   }
 
   netscape.security.PrivilegeManager.enablePrivilege('UniversalXPConnect');
-  var Cc = Components.classes,
-      Ci = Components.interfaces;
-  var file = Cc['@mozilla.org/file/local;1'].createInstance(Ci.nsILocalFile);
+  const Cc = Components.classes,
+        Ci = Components.interfaces;
```
timvandermeij: I think we should simply make a second statement for this since we aim to avoid multiple declarations in one statement as much as possible for readability (we barely have it in the codebase, and this part is fairly old code):

```javascript
const Cc = Components.classes;
const Ci = Components.interfaces;
```

wojtekmaj: I'd be happy to do it in another PR for clarity of our actions, what do you think?

Snuffleupagus: > I'd be happy to do it in another PR for clarity of our actions,

In this *particular* case, the changes are too small to warrant another PR in my opinion; hence simply addressing all comments and squashing everything together should suffice.

timvandermeij: Yes, please keep it within this PR and squash all the changes together.

wojtekmaj: Done!

Snuffleupagus:
Even ignoring the fact that we now want to avoid making changes like these on an *individual* file basis (to minimize churn), there's also the fact that this file is explicitly *excluded* from linting courtesy of https://github.com/mozilla/pdf.js/blob/dced0a38212b7544200f395abb477d1612082193/.eslintignore#L10-L11

/cc @timvandermeij All-in-all, should we just close this as WONTFIX?

---

There's also the slightly off-topic question, probably best discussed separately, if it even makes sense to keep those two files in the repository!?
Looking at the code, e.g. the global imports and the Mozilla browser-chrome `Components` usage, all of this looks *really* old and it simply won't be possible (nor have been for years) to actually run it now.

timvandermeij:
Yes, it's probably best to close this and revisit it later. Thanks!

PR #11434, opened 2019-12-22
Use const in pdf_thumbnail_view.js

wojtekmaj:


web/pdf_thumbnail_view.js
```diff
@@ -371,9 +372,9 @@ class PDFThumbnailView {
     // drawImage does an awful job of rescaling the image, doing it gradually.
     let reducedWidth = canvas.width << MAX_NUM_SCALING_STEPS;
     let reducedHeight = canvas.height << MAX_NUM_SCALING_STEPS;
-    let reducedImage = TempImageFactory.getCanvas(reducedWidth,
+    const reducedImage = TempImageFactory.getCanvas(reducedWidth,
```
timvandermeij: The line below needs to be aligned.

wojtekmaj: Done!

timvandermeij:
Closing since we decided to convert the entire `web` folder automatically using ESLint's `--fix` parameter in ##11450 to both reduce code churn and to be done with the conversion quickly. Thank you for helping out with this though; your patches prompted us to look into ways to automate tasks more, like the linting and `const` conversion, so that's really positive!

PR #11435, opened 2019-12-22
Add lint command for easy linting

wojtekmaj:
Adds pretty standard command to run ESLint easily

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "yargs": "^11.1.1"
   },
   "scripts": {
-    "lint": "eslint .",
+    "lint": "gulp lint",
     "test": "env SKIP_BABEL=true gulp npm-test"
   },
   "repository": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -53,6 +53,7 @@
     "yargs": "^11.1.1"
   },
   "scripts": {
+    "lint": "eslint .",
```
Snuffleupagus: This looks similar to e.g. PR #8415, which was rejected. Furthermore, *if* this PR is deemed generally acceptable then this should be `gulp lint` and nothing else.

Snuffleupagus: I think that https://github.com/mozilla/pdf.js/pull/8415#issuecomment-302249805 sums it up quite nicely, and besides `gulp lint` shouldn't be more difficult to type than `npm lint` :-), hence WONTFIX seems like the appropriate resolution for this PR too.

PR #11436, opened 2019-12-22
Update webpack-stream to fix vulnerabiliy reported by npm and dedupe Webpack

wojtekmaj:
https://npmjs.com/advisories/1084

webpack-stream between 4.0.3 and 5.0.0 added official support for Webpack 4. Which is nice since we ARE using Webpack 4... Also, dropped support for Node.js 4 which shouldn't be a big deal for us since we are already using packages that are incompatible with Node.js 4 (Webpack 4.x supports Node.js 6 and up).

A nice side effect is that this update removes a ton of duplicate dependencies - webpack-stream 4.0.3 included a _second copy_ of Webpack, version 3.x, in our node_modules!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2623,17 +2623,6 @@
         "sha.js": "^2.4.8"
       }
     },
-    "cross-spawn": {
-      "version": "5.1.0",
-      "resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-5.1.0.tgz",
-      "integrity": "sha1-6L0O/uWPz/b4+UUQoKVUu/ojVEk=",
-      "dev": true,
-      "requires": {
-        "lru-cache": "^4.0.1",
-        "shebang-command": "^1.2.0",
-        "which": "^1.2.9"
-      }
-    },
     "crypto-browserify": {
       "version": "3.12.0",
       "resolved": "https://registry.npmjs.org/crypto-browserify/-/crypto-browserify-3.12.0.tgz",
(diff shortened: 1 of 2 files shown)


timvandermeij:
I wonder why it showed success, because the build failed. From the logs:

> You ran Webpack twice. Each instance only supports a single concurrent compilation at a time.

The preview commands runs `gulp web`, so I think this update changed something in the build process. Does this also happen locally?

wojtekmaj:
Looks like there has been a regression in webpack-stream 5.1.0 and up; current solution (perfectly satisfactory for this PR) is to upgrade to 5.0.x, not all the way.

Snuffleupagus:
> Does this also happen locally?

*For future reference:* When making changes that could, in one way or another, potentially break the builds you should run all of the appropriate tests locally while developing the patch.
In this case that would entail running `gulp generic` and checking that the build works, and obviously also running the *complete* test-suite with `gulp test` to ensure that all tests pass.

> Looks like there has been a regression in webpack-stream 5.1.0 and up;

That seems to be tracked in issue https://github.com/shama/webpack-stream/issues/201, and the regression is apparently PR https://github.com/shama/webpack-stream/pull/109, which has been open for over a year :-(

If that regression isn't addressed, we may need to start looking for alternatives to `webpack-stream`...

wojtekmaj:
That failure looks invalid to me!

timvandermeij (approved):

timvandermeij:
Yes, it's completely unrelated to this patch. Nice clean-up!

PR #11437, opened 2019-12-22
Extract & use createHeaders helper

wojtekmaj:
Prevents code duplication by implementing createHeaders enclosing duplicated pieces of code

### DIFF
--- a/src/display/fetch_stream.js
+++ b/src/display/fetch_stream.js
@@ -33,6 +33,18 @@ function createFetchOptions(headers, withCredentials, abortController) {
   };
 }
 
+function createHeaders(httpHeaders) {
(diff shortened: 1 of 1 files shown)


Snuffleupagus (changes requested):
Nit: It would be nice if the *commit message* actually mentioned where these changes are being made, e.g. something like `Extract & use createHeaders helper in src/display/fetch_stream.js`, since otherwise it's not really possible to understand the changes without looking at the code.

wojtekmaj:
@Snuffleupagus Sure thing, done!

timvandermeij (approved):

timvandermeij:
Good clean-up!

PR #11438, opened 2019-12-22
Add ESLint rule for indentation

wojtekmaj:
I find formatting a tedious and unproductive task. I often see in code reviews that you want the code indented properly.

I'm trying to understand all the rules you have in your mind and so far, I was able to form the following rules:

* 2 spaces are used for indentation
* If array elements won't fit in one line, align the elements in the next line with the first element:
  ```js
  const arr = ['baa', 'bab', bac',
               'bad', 'bae'];
  ```
* If arguments in call expressions won't fit one line, align the arguments in the next line with the first argument:
  ```js
   function foo(baa, bab, bac
                bad, bae) { ...
  ```
  There are many cases in which this rule is being broken, for example
  ```js
    this.thumbnailButton.classList.toggle('toggled',
      view === SidebarView.THUMBS);
  ```
* If parameters in function expressions won't fit one line, align the parameters in the next line with the first parameter:
  ```js
   const result = foo(baa, bab, bac
                      bad, bae)
  ```
* If object properties won't fit in one line, align the properties in the next line with the first property:
  ```js
  const obj = { baa: 1, bab: 2, bac: 3,
                bad: 4, bae: 4 };
* Switch case should be formatted like this:
  ```js
  switch (case) {
    case 1: ...
    case 2: ...
  }
  ``` 
* Multiple variable declarations should align with the first one:
  ```js
  const a = 0,
        b = 1;
  ```

This ESLint configuration enforces these rules.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -37,6 +37,13 @@
     // Possible errors
     "for-direction": "error",
     "getter-return": "error",
+    "indent": ["error", 2, {
+      "CallExpression": { "arguments": "first" },
+      "FunctionExpression": { "parameters": "first" },
+      "ObjectExpression": "first",
+      "SwitchCase": 1,
+      "VariableDeclarator": "first"
+    }],
     "no-async-promise-executor": "error",
     "no-cond-assign": ["error", "except-parens"],
     "no-constant-condition": ["error", { "checkLoops": false, }],


timvandermeij:
I think it's really good to check what we can achieve with linting for formatting code, especially since automating this task will make code consistency and reviews easier. It looks like you're correct about most points. I'm not sure if we have a convention about the third point, since you also mentioned lots of variations here. I think it depends on the length of the line which one we choose, so we should either make a fixed choice or exclude that one from linting. Finally, we usually don't prefer multiple declarations on one line, and if we must then they should stay on one line. If it doesn't fit anymore, just make a new statement.

timvandermeij:
If we don't add the third rule for now, how many linting errors do we have left? If that is fixable, I would say that we can go for that.

Snuffleupagus:
If we're going to touch basically the entire code-base, then I think that we should do it the "right" way (to minimize unnecessary churn) and try to align a bit better with mozilla-central at the same time.

I've not had time yet, but will during Christmas, to look into using Prettier instead to end all formatting discussions once and for all. Hence I'd prefer that we WONTFIX this, since a better solution is available.

wojtekmaj:
Ah! Prettier would be perfect. Looks like we're not gonna avoid a massive codebase modification anyway - after a lot of experimentation with these rules I ended up with ~5500 lines to be modified anyway. Which isn't too bad, but ain't minor as well.

timvandermeij:
@wojtekmaj Just a heads-up: we switched to Prettier in #11446, so code formatting should be less tedious now :+1:

PR #11440, opened 2019-12-23
Tweak the "gets page stats after rendering page, with `pdfBug` set" unit-test to remove an intermittent failure on Travis

Snuffleupagus:
I recently noticed a couple of intermittent failures on Travis, hence this patch which changes the expectation to be identical to the 'Page Request' check in the preceding test-case.


timvandermeij (approved):

timvandermeij:
Thank you for fixing this so quickly! I was just about to file an issue for this, but you were faster ;-)

PR #11445, opened 2019-12-24
Include a unit test for OpenAction dictionaries without `Type` entries (PR 11443 follow-up)

timvandermeij:
The original issue did not contain a (reduced) test case that we could include and linked test cases are not ideal for unit tests, so the original PR could only be verified manually.

I found this a bit unfortunate considering that the print data is exposed through the API, so I thought about how we could have an automated test and managed to create a reduced test case with the OpenAction dictionary from the file in the original issue.

Therefore, this commit includes a unit test for parsing OpenAction dictionaries without `Type` entries. I verified that this PDF file behaves the same as the original one, i.e., no print dialog is shown for older viewers and the print dialog is shown for the most recent viewer.

@Snuffleupagus Would you perhaps be willing to review this?

### DIFF
--- a/test/pdfs/.gitignore
+++ b/test/pdfs/.gitignore
@@ -350,6 +350,7 @@
 !operator-in-TJ-array.pdf
 !issue7878.pdf
 !font_ascent_descent.pdf
+!issue11442_reduced.pdf
 !issue8097_reduced.pdf
 !transparent.pdf
 !xobject-image.pdf
--- a/test/pdfs/issue11442_reduced.pdf
+++ b/test/pdfs/issue11442_reduced.pdf
@@ -1,0 +1,72 @@
+%PDF-1.7
+%‚„œ”
+1 0 obj 
+<<
+/OpenAction 
+<<
+/N /Print
+/S /Named
+>>
+/Pages 2 0 R
+/Type /Catalog
+>>
+endobj 
+2 0 obj 
+<<
+/Kids [3 0 R]
+/Count 1
+/Type /Pages
+>>
+endobj 
+3 0 obj 
+<<
+/Parent 2 0 R
+/MediaBox [0 0 200 50]
+/Resources 
+<<
+/Font 
+<<
+/F1 4 0 R
+>>
+>>
+/Contents 5 0 R
+/Type /Page
+>>
+endobj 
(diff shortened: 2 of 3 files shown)


Snuffleupagus (commented):
My apologies for not adding a test-case! I was *really* short on time yesterday, but given that the fix was so simple I figured I'd just throw together a quick patch.

That's for adding a new unit-test!
However, I couldn't help noticing that the reduced PDF file is corrupt (there's a `Warning: Indexing all PDF objects` message) and I'd generally suggest that we don't add such files *unless* we actually want/need to test the code-paths specifically intended for corrupt files.

In the hopes that I'm not overstepping, and feel free to just ignore this, but here's a minimal PDF file which should reproduce the issue and that you may use if you want to: [issue11442_reduced.pdf](https://github.com/mozilla/pdf.js/files/4000480/issue11442_reduced.pdf)

timvandermeij:
That's really helpful, thanks! I'll update the patch. Would you mind explaining how you constructed it? I'm trying to learn how to create such test cases. I usually edit the PDF contents in an editor (decompressing the streams with `qpdf` if necessary) and fix the XRef table afterwards with Ghostscript's `gs` command, but for this PDF it didn't work because it dropped the `/OpenAction` entry completely. I figured there must be a better/simpler way to edit PDFs and fix the XRef table afterwards.

Snuffleupagus:
> Would you mind explaining how you constructed it?

By hand, using a *very* basic file I created years ago as the starting point: [reduced_test-case.pdf](https://github.com/mozilla/pdf.js/files/4000845/reduced_test-case.pdf).
After the `/OpenAction` dictionary, as taken from the original file, was manually added to `/Root` dictionary I simply used the [PDFtk command-line](https://www.pdflabs.com/tools/pdftk-server/) to generate a valid PDF document.

I've been using that tool for *years* now, basically running every reduced test-case I've ever created through it, and I'm very happy with it :-)
I've also used that tool a lot when debugging, since it's also able to e.g. decompress streams (to get at the underlying data easier) and split out individual pages from a document.
Some *really* broken files can be difficult for the tools to handle, but generally it works very well!

Snuffleupagus (approved):
Thanks for doing this, r=me

PR #11446, opened 2019-12-25
Enable auto-formatting of the entire code-base using Prettier (issue 11444)

### LINKED ISSUE #11444: Test-drive Prettier
This is a tracking issue since `mozilla-central` uses Prettier upstream, so to align with them it's probably best if we adopt it as well if it works for us. This should help with code formatting.

Being worked on by @Snuffleupagus.

Snuffleupagus:
Note that Prettier, purposely, has only limited [configuration options](https://prettier.io/docs/en/options.html). The configuration file is based on [the one in `mozilla central`](https://searchfox.org/mozilla-central/source/.prettierrc) with just a few additions (to avoid future breakage if the defaults ever changes).

Prettier is being used for a couple of reasons:

 - To be consistent with `mozilla-central`, where Prettier is already in use across the tree.

 - To ensure a *consistent* coding style everywhere, which is automatically enforced during linting (since Prettier is used as an ESLint plugin). This thus ends "all" formatting disussions once and for all, removing the need for review comments on most stylistic matters.

Many ESLint options are now redundant, and I've tried my best to remove all the now unnecessary options (but I may have missed some).
Note also that since Prettier considers the `printWidth` option as a guide, rather than a hard rule, this patch resorts to a small hack in the ESLint config to ensure that *comments* won't become too long.

*Please note:* This patch is generated automatically, by appending the `--fix` argument to the ESLint call used in the `gulp lint` task. It will thus require some additional clean-up, which will be done in a *separate* commit.

(On a more personal note, I'll readily admit that some of the changes Prettier makes are *extremely* ugly. However, in the name of consistency we'll probably have to live with that.)

Fixes #11444.


Snuffleupagus:
While the first and last commits should be possible to review manually, the main commit is unfortunately be feasible to review (and enabling Prettier everywhere at once seemed like a good idea).
Hence it's probably going to be necessary to trust that ESLint `--fix` did the "right" thing here, and also that I ran the command correctly :-)

Snuffleupagus:
*Once Prettier auto-formatting has landed:* I think that we should avoid doing things like e.g. the recent `prefer-const` conversion on a file-by-file basis. and rather do those things globally and/or on a directory basis instead (since I'm not crazy about the code churn the previous method results in).

timvandermeij (approved):

timvandermeij:
Thank you for improving this! I think it's a huge step forward in terms of consistency with `mozilla-central` and in the code base itself. It will also help during development work and reviews.

I checked everything except for the main commit, but I did look over it to see if all changes look correct. I'm indeed going to trust that ESLint did the work correctly, and the passing tests and linting prove this.

I have reviewed this pull request first because I figured rebasing this would be horrible, so now we can rebase the other patches easily.

Snuffleupagus:
Thanks a lot for taking the time to review this; and also for prioritizing getting this landed quickly to avoid painful rebasing :-)

wojtekmaj:
Holy crap, that's how to enter 2020 in style! 😎 Congratulations!

PR #11447, opened 2019-12-25
Convert all non-RGBA colors to RGBA colors

### LINKED ISSUE #11372: too many colors in viewer.css
I'm trying to re-skin the viewer, in order to comply with

> we do ask if you plan to embed the viewer in your own site, that it not just be an unmodified version. Please re-skin it or build upon it.

and to make it themeable (within Atom).

However, that proves to be made extremely hard by the way the viewer is styled. I was able to replace the icons (images) by alternatives from Awesome Font. Now I'm trying to change the colors and I found that (according to [CSS Color Extractor](http://www.css-color-extractor.com/))

there are **83 unique color specifications** used in `viewer.css`,
and they result in **70 different RGBA colors**.

And that doesn't even count things linke
```
opacity: 0.2;
background: #ff0;
```

That's not just a very large number, it also means that some of the specifications are unnecessarily different. Which is consistent with the fact that the color specifications have all possible different formats:
**rgb(), rgba(), hsl(), hsla(), 3-digit hex, 6-digit hex, named (black, white, blue, yellow).**

Black is specified as **black and #000**,
transparency as **transparent, rgba(0,0,0,0), and hsla(0,0%,100%,0)**.

There are tiny variations like: **hsla(0,0%,100%,0), hsla(0,0%,100%,.02), hsla(0,0%,100%,.03), hsla(0,0%,100%,.05), hsla(0,0%,100%,.08), hsla(0,0%,100%,.09), and hsla(0,0%,100%,.15).**
Here's what that looks like on a black background:
image
In case you can't tell: there is a white "A" in each box, in different shades of basically invisible.

That doesn't just make it nearly impossible to comply with the ask quoted above. It is also bad design practice, and makes it extremely hard for you if at some point you're planning to actually give the viewer a modern or themeable design.

timvandermeij:
In our usage instructions for the viewer we state the following:

> We do ask if you plan to embed the viewer in your own site, that it not just be an unmodified version. Please re-skin it or build upon it.

Users have indicated that this is currently quite hard because we use no less than seven different ways to specify colors in CSS, namely RGB, RGBA, HSL, HSLA, three-digit HEX, six-digit HEX and named, leading to 83 unique color specifications.

This commit series simplifies our color specifications across the entire code base, but in particular in `web/viewer.css`, to specify all colors in RGBA instead. The choice for RGBA is because RGB is well-known for web development and because we need to be able to express opacity (alpha channel) in the color specification. Doing so brings down the number of unique color specifications to 70. This should be acceptable now because quite a few of the unique color specifications are actually opacity variations on the same color, and given that all colors are now in the same format, using find/replace is now possible, making the re-skinning task a lot easier.

Fixes #11372.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -466,7 +466,7 @@ html[dir='ltr'] .doorHangerRight:before {
 }
 
 #findInput.notFound {
-  background-color: rgb(255, 102, 102);
+  background-color: rgba(255, 102, 102, 1);
 }
 
 #toolbarViewerMiddle {
@@ -1511,7 +1511,7 @@ html[dir='rtl'] #documentPropertiesOverlay .row > * {
 }
 
 #viewer.textLayer-visible .canvasWrapper {
-  background-color: rgb(128,255,128);
+  background-color: rgba(128, 255, 128, 1);
 }
 
 #viewer.textLayer-visible .canvasWrapper canvas {
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -326,7 +326,7 @@ html[dir='rtl'] #toolbarContainer, .findbar, .secondaryToolbar {
   min-width: 16px;
   padding: 0px 6px 0px 6px;
   margin: 4px 2px 4px 2px;
-  color: hsl(0,0%,85%);
+  color: rgba(217, 217, 217, 1);
   font-size: 12px;
   line-height: 14px;
   text-align: left;
@@ -360,7 +360,7 @@ html[dir='rtl'] .findbar {
   width: 200px;
 }
 #findInput::-webkit-input-placeholder {
-  color: hsl(0, 0%, 75%);
+  color: rgba(191, 191, 191, 1);
 }
 #findInput::placeholder {
   font-style: italic;
@@ -451,8 +451,8 @@ html[dir='ltr'] .doorHangerRight:before {
 }
 
 #findResultsCount {
-  background-color: hsl(0, 0%, 85%);
-  color: hsl(0, 0%, 32%);
+  background-color: rgba(217, 217, 217, 1);
+  color: rgba(82, 82, 82, 1);
   text-align: center;
   padding: 3px 4px;
 }
@@ -719,7 +719,7 @@ html[dir='rtl'] .dropdownToolbarButton {
 .dropdownToolbarButton > select {
   min-width: 140px;
   font-size: 12px;
-  color: hsl(0,0%,95%);
+  color: rgba(242, 242, 242, 1);
   margin: 0;
   padding: 3px 2px 2px;
   border: none;
@@ -727,7 +727,7 @@ html[dir='rtl'] .dropdownToolbarButton {
 }
 
 .dropdownToolbarButton > select > option {
-  background: hsl(0,0%,24%);
+  background: rgba(61, 61, 61, 1);
 }
 
 #customScaleOption {
@@ -1036,7 +1036,7 @@ html[dir='rtl'] .verticalToolbarSeparator {
   border-color: hsla(0,0%,0%,.32) hsla(0,0%,0%,.38) hsla(0,0%,0%,.42);
   box-shadow: 0 1px 0 hsla(0,0%,0%,.05) inset,
               0 1px 0 hsla(0,0%,100%,.05);
-  color: hsl(0,0%,95%);
+  color: rgba(242, 242, 242, 1);
   font-size: 12px;
   line-height: 14px;
   outline-style: none;
@@ -1085,7 +1085,7 @@ html[dir='rtl'] .verticalToolbarSeparator {
   margin: 4px 2px 4px 0;
   border: 1px solid transparent;
   border-radius: 2px;
-  color: hsl(0,0%,85%);
+  color: rgba(217, 217, 217, 1);
   font-size: 12px;
   line-height: 14px;
   text-align: left;
@@ -1373,7 +1373,7 @@ html[dir='rtl'] .outlineItemToggler::before {
   display: inline-block;
   padding: 15px;
   border-spacing: 4px;
-  color: hsl(0,0%,85%);
+  color: rgba(217, 217, 217, 1);
   font-size: 12px;
   line-height: 14px;
   background-color: #474747; /* fallback */
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -34,7 +34,7 @@ html {
 body {
   height: 100%;
   width: 100%;
-  background-color: #404040;
+  background-color: rgba(64, 64, 64, 1);
   background-image: url(images/texture.png);
 }
 
@@ -217,7 +217,7 @@ html[dir='rtl'] #outerContainer.sidebarOpen #viewerContainer:not(.pdfPresentatio
 #toolbarSidebar {
   width: 100%;
   height: 32px;
-  background-color: #424242; /* fallback */
+  background-color: rgba(66, 66, 66, 1); /* fallback */
   background-image: url(images/texture.png),
                     linear-gradient(rgba(77, 77, 77, 0.99), rgba(64, 64, 64, 0.95));
 }
@@ -252,7 +252,7 @@ html[dir='rtl'] #sidebarResizer {
 #toolbarContainer, .findbar, .secondaryToolbar {
   position: relative;
   height: 32px;
-  background-color: #474747; /* fallback */
+  background-color: rgba(71, 71, 71, 1); /* fallback */
   background-image: url(images/texture.png),
                     linear-gradient(rgba(82, 82, 82, 0.99), rgba(69, 69, 69, 0.95));
 }
@@ -456,7 +456,7 @@ html[dir='ltr'] .doorHangerRight:before {
 
 #findMsg {
   font-style: italic;
-  color: #A6B7D0;
+  color: rgba(166, 183, 208, 1);
 }
 #findMsg:empty {
   display: none;
@@ -901,7 +901,7 @@ html[dir="rtl"] #viewOutline.toolbarButton::before {
   top: 1px;
   /* Create a filled circle, with a diameter of 9 pixels, using only CSS: */
   content: '';
-  background-color: #70DB55;
+  background-color: rgba(112, 219, 85, 1);
   height: 9px;
   width: 9px;
   border-radius: 50%;
@@ -1314,7 +1314,7 @@ html[dir='rtl'] .outlineItemToggler::before {
 }
 
 #errorWrapper {
-  background: none repeat scroll 0 0 #FF5555;
+  background: none repeat scroll 0 0 rgba(255, 85, 85, 1);
   color: rgba(255, 255, 255, 1);
   left: 0;
   position: absolute;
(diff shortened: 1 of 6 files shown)


web/viewer.css
```diff
@@ -309,12 +309,10 @@ html[dir='rtl'] #toolbarContainer, .findbar, .secondaryToolbar {
   left: 0;
   height: 100%;
   width: calc(100% + 150px);
-
   background: repeating-linear-gradient(135deg,
-                                        #bbb 0, #999 5px,
-                                        #999 45px, #ddd 55px,
-                                        #ddd 95px, #bbb 100px);
-
+                                        rgba(187,187,187,1) 0, rgba(153,153,153,1) 5px,
```
Snuffleupagus: *Nice idea, trying to consolidate the colours.*
 
The above line is just one example, but as far as I'm concerned it does highlight a slight problem that comes with disallowing hex values:
Looking at the hex values on the left, it's immediately obvious that the values are evenly spaced and symmetrical. With the RGBA format, not so much...

*Personally*, I find that the hexadecimal format can often be nicer to work with than the rgba one.

Hence my general question here, is if we could perhaps use the `#RRGGBB[AA]` or `#RGB[A]` formats instead?
I know that those formats have somewhat limited browser support (hi IE11/Edge), but there appears to be PostCSS plugins that could maybe handle the conversion for us: https://github.com/postcss/postcss/blob/master/docs/plugins.md

timvandermeij: I should mention that the idea for this commit series is to unify all color specifications into a single format while avoiding any CSS pre-/post-processing to keep the code and the build process simple.

I think the problem with the `#RRGGBB[AA]` format is not so much the limited browser support since that's fixable with post-processing, but I also find it harder to read. In the RGBA format, which seems most common for other projects too, the alpha channel is simply a value from 0 to 1 whereas in the `#RRGGBB[AA]` format that's less obvious because the alpha value is also in hexadecimal. Moreover, I don't see much difference in terms of this particular example, mainly because the values are aligned in a way that still allows you to see that there is symmetry. In short, I went for the RGBA format because of readability and because it's natively supported and often used, mainly with the idea of keeping it simple.

Initially I thought that switching to SCSS would be a good idea so that we could define the unique colors as variables. This would allow us to reduce the number of color specifications by expressing colors with opacities in terms of their root color, like done at https://stackoverflow.com/questions/21189495/setting-opacity-on-primary-color-in-sass. However, I quickly found that this would require conversions to the SCSS format and extra build steps for pre-processing, so I dismissed that idea. Then I thought that we could simply use native CSS variables, which only IE doesn't support, and that is fixable if we use PostCSS like you suggested. It would look something like https://stackoverflow.com/questions/40010597/how-do-i-apply-opacity-to-a-css-color-variable. Now that everything is in one format, the idea is that it's easier to extract the root colors into variables in a follow-up.

In conclusion, I'm now wondering what we best way forward is. Either we choose a simple and native solution, which is what this patch implements, or accept that we need to do pre-/post-processing in our build steps and use other techniques like `#RRGGBB[AA]` and CSS variables. Extracting colors into variables may be useful for us, but may also be overkill because we don't have *that* many colors, so I'm just not sure if we need all that complexity?!

What do you think about this? Should we prepare for CSS pre-/post-processing or should we try to avoid it for now?

Snuffleupagus: > [...] while avoiding any CSS pre-/post-processing to keep the code and the build process simple.

The reason that I even mentioned PostCSS plugins, is that we're already using Autoprefixer and PostCSS in the builds; see https://github.com/mozilla/pdf.js/blob/master/gulpfile.js

> [...] the alpha channel is simply a value from 0 to 1 whereas in the `#RRGGBB[AA]` format that's less obvious because the alpha value is also in hexadecimal.

I suppose that the part about the alpha channel behaviour indeed may look slightly confusing to users.

> Moreover, I don't see much difference in terms of this particular example, mainly because the values are aligned in a way that still allows you to see that there is symmetry.

We'll probably have to agree to disagree about the general readability/usability of hex vs rgb colour values, and I'll concede that this has a high degree of subjectivity attached :-)

> Then I thought that we could simply use native CSS variables, which only IE doesn't support, and that is fixable if we use PostCSS like you suggested.

Yes, there appears to be plugins which are able to inline CSS variables such that everything works in older browsers.
However, the idea of using CSS variables a lot more to improve both maintenance and customizability would seem like a nice feature. If such a re-factoring ever happens, having a build step that removes those variables thus seem like a bad idea. A much better option, assuming it would work of course, may be to try using something like https://github.com/jhildenbiddle/css-vars-ponyfill in `GENERIC` builds (to simply re-write CSS variables at runtime instead).

---

 - Generally though, if we're going to use the rgba format I think that the values ought to be spaced properly for improved readability (since we wouldn't allow "glued" values in e.g. JavaScript). Hence these values need to be formatted at `rgba(255, 255, 255, 1);` as far as I'm concerned.

 - Comparing the preview with `master`, there does appear to be colours which are now off-by-one. Using the Firefox devtools "Inspector", and looking at the "Computed" tab for e.g. the `#toolbarContainer` element there's a lot of small movement. Can this be fixed, please?

All in all, if the two points above are addressed I'd be happy to sign off on these changes :-)

timvandermeij: > Generally though, if we're going to use the rgba format I think that the values ought to be spaced properly for improved readability [...]

I agree with that; I'll change that.

> Comparing the preview with master, there does appear to be colours which are now off-by-one in all components [...]

Yes, it looks like my color converter tool got confused about the opacity part. I'll have to go over all occurrences again anyway to space them properly, so I'll also fix those occurrences.

timvandermeij: Both changes should be done now.

web/viewer.css
```diff
@@ -1020,23 +1017,23 @@ html[dir='rtl'] .verticalToolbarSeparator {
...
 }
 
 .toolbarField {
   padding: 3px 6px;
   margin: 4px 0 4px 0;
   border-radius: 2px;
-  background-color: hsla(0,0%,100%,.09);
-  background-image: linear-gradient(hsla(0,0%,100%,.05), hsla(0,0%,100%,0));
+  background-color: rgba(255, 255, 255, 0.09);
+  background-image: linear-gradient(rgba(255, 255, 255, 0.05), rgba(0, 0, 0, 0));
```
Snuffleupagus: The *second* colour value doesn't look correct here; it should be `rgba(255, 255, 255, 0)` as far as I can tell.

Snuffleupagus (changes requested):
Sorry, but I've spotted at least one more conversion error here :-P

Please note that I've *not* checked all values, but I rather just picked a couple at random with the devtools until I stumbled onto a broken one.

Just of of curiosity, how did you convert these values (since unless it's fully scripted, the risk of mistakes makes me a bit wary of these types of changes)?

web/viewer.css
timvandermeij: The converter that I used for these changes (https://bl.ocks.org/njvack/02ad8efcb0d552b0230d) simplified this to `rgba(0, 0, 0, 0)` because the opacity is 0. I chose that converter because it accepts any valid CSS color specification and always outputs RGBA based on the actual value of a rendered pixel. I could obviously change this to `rgba(255, 255, 255, 0)`, but the effect would be the same because of the 0 opacity.

I do agree with you that it's perhaps a bit confusion, so I'll update the patch to use the original values instead. Note that this only applies to values where the opacity component is 0; the other values were already converted as-is, so those have not been changed.

timvandermeij: Done. This only required changes in the second commit (HSLA to RGBA conversion).

web/viewer.css
```diff
@@ -1070,13 +1070,13 @@ html[dir='rtl'] .verticalToolbarSeparator {
...
-  border-color: hsla(0,0%,0%,.4) hsla(0,0%,0%,.43) hsla(0,0%,0%,.45);
+  background-color: rgba(255, 255, 255, 0.11);
+  border-color: rgba(0, 0, 0, 0.4) rgba(0, 0, 0, 0.43) rgba(0, 0, 0, 0.45);
 }
 
 .toolbarField:focus {
-  background-color: hsla(0,0%,100%,.15);
-  border-color: hsla(204,100%,65%,.8) hsla(204,100%,65%,.85) hsla(204,100%,65%,.9);
+  background-color: rgba(255, 255, 255, 0.15);
+  border-color: rgba(76, 184, 255, 0.8) rgba(76, 184, 255, 0.85) rgba(76,184, 255, 0.9);
```
Snuffleupagus: Nit: `77` in the *first* component for every value here?

web/viewer.css
```diff
@@ -1452,8 +1452,8 @@ html[dir='rtl'] #documentPropertiesOverlay .row > * {
 }
 
 .fileInput {
-  background: white;
-  color: black;
```
Snuffleupagus: Huh, why are these colours even set when the element has `visibility: hidden;`!?

Obviously *not* something that should to be fixed here, but I cannot help thinking that the CSS files might generally benefit from some clean-up since they probably contain a fair amount of rules only relevant for now unsupported browsers.

web/viewer.css
```diff
@@ -1490,11 +1490,11 @@ html[dir='rtl'] #documentPropertiesOverlay .row > * {
   font-weight: bold;
 }
 .debuggerShowText {
-  background: none repeat scroll 0 0 yellow;
-  color: blue;
+  background: none repeat scroll 0 0 rgba(255, 255, 0, 1);
+  color: rgba(0,0, 255, 1);
```
Snuffleupagus: Nit: Missing a space here.

Snuffleupagus (changes requested):
Sorry, but I have just a few small comments after going through these patches line-by-line.

Furthermore, these patches seem to be missing https://github.com/mozilla/pdf.js/blob/master/test/annotation_layer_builder_overrides.css and please also check https://github.com/mozilla/pdf.js/blob/master/test/text_layer_test.css (although that one seems OK).

web/viewer.css
Snuffleupagus: > [...] so I'll update the patch to use the original values instead.

Thanks; that did improve readability a tiny bit in the affected places, since it's now immediately obvious that only the alpha channel changes rather than the base colour itself.

web/viewer.css
timvandermeij: Fixed.

web/viewer.css
timvandermeij: Fixed.

web/viewer.css
timvandermeij: That's... a very good question. I can only imagine that the visibility is toggled/used to be toggled using JavaScript. It probably cannot hurt to go over the CSS in a follow-up as an incremental improvement.

timvandermeij:
I have also checked the other CSS files. The first one required some changes which are combined in the "three-digit HEX" commit and the last one didn't require changes.

Snuffleupagus (approved):
Looks good to me now; thanks for bringing some much needed consistency to this part of the code-base.

Also, sorry about the back-and-forth here during review!

timvandermeij:
Thank you for the thorough review!

allefeld:
Great, thank you! @timvandermeij

PR #11449, opened 2019-12-26
Move the regular expression, used with auto printing in the viewer, to `web/ui_utils.js` and also use it in the API unit-tests

Snuffleupagus:
Rather than having a copy of this regular expression in the `test/unit/api_spec.js` file, with a comment about keeping it up-to-date with the code in the viewer (note the incorrect file reference as well), we can just import it instead to simplify all of this.

*This is something I noticed while looking at PR #11445, I'll rebase this once that one has landed.*

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -44,6 +44,7 @@ import {
   PDFPageProxy,
   PDFWorker,
 } from "../../src/display/api";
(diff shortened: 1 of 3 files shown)


timvandermeij:
Looks good if the newly added unit test is also adjusted.

timvandermeij (approved):

timvandermeij:
Nice improvement!

PR #11450, opened 2019-12-26
Enable the ESLint `prefer-const` rule in the `web/` directory

Snuffleupagus:
- Enable the ESLint `prefer-const` rule in the `web/` directory

   Please find additional details about the ESLint rule at https://eslint.org/docs/rules/prefer-const

   Note that this patch is generated automatically, by using the ESLint `--fix` argument, and will thus require some additional clean-up (which is done separately).

 - Fix remaining linting errors, from enabling the `prefer-const` ESLint rule in the `web/` directory

   This covers the handful of cases that the `--fix` command couldn't deal with, and the changes aren't just fixing the linting errors but attempt to slightly improve the relevant code.

### DIFF
--- a/web/.eslintrc
+++ b/web/.eslintrc
@@ -9,5 +9,6 @@
 
     // ECMAScript 6
     "no-var": "error",
+    "prefer-const": "error",
   },
 }
--- a/web/annotation_layer_builder.js
+++ b/web/annotation_layer_builder.js
@@ -64,7 +64,7 @@ class AnnotationLayerBuilder {
         return;
       }
 
-      let parameters = {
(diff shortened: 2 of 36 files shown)


timvandermeij:
I like the idea of doing this all at once, both to reduce code churn and to simply be done with it quickly :-)

I'm afraid the previous merge required a rebase here, but hopefully it's not too difficult given that ESLint's `--fix` flag was used to generate this.

timvandermeij (approved):

timvandermeij:
Nice work; glad that this conversion is done for the `web` folder now!

PR #11454, opened 2019-12-28
Move the SegoeUISymbol font to the `getNonStdFontMap` (PR 8698 follow-up)

Snuffleupagus:
For reasons that I now cannot even begin to understand, the non-standard SegoeUISymbol font was placed in the `getStdFontMap`. That honestly makes no sense, hence this patch which does what I *should* have done from the start.


timvandermeij (approved):

timvandermeij:
Thanks!

Issue #10965, opened 2019-07-11
Upgrade `systemjs` to the most recent version


timvandermeij:
Currently we use the `systemjs` 0.x version, but the most recent version as of writing is the 4.x version. Updating is a bit difficult because of API changes, but should in return simplify the logic and also allow us to e.g., retry #10953 afterwards since that it blocked on this update.

More information can be found on https://guybedford.com/systemjs-2.0.

Samrose-Ahmed:
I'd like to take this.

timvandermeij:
Sure, please do!

arvind0598:
If this is still inactive, can I try my hand at this?

timvandermeij:
We're one month further and we haven't seen any activity on this, so feel free to work on this!

Snuffleupagus:
For anyone working on this, please make sure to *carefully* test your patch prior to submitting a PR. This entails, at the very least:
 - Ensure that the development viewer works correctly, by running `gulp server` and:
   - Open the development viewer, with http://localhost:8888/web/viewer.html, in a browser and make sure that the viewer both loads *and* works correctly without any Errors in the console.
   - Run the unit-tests directly in a browser, with http://localhost:8888/test/unit/unit_test.html, and ensure that all same tests pass with the patch as with the `master` branch.
 - Run `gulp unittest` locally, and ensure that all tests pass.
 - Run `gulp fonttest` locally, and ensure that all tests pass.
 - Run `gulp browsertest` locally, and ensure that all tests pass.
 - Finally run `gulp test` locally, and ensure that all tests pass.

Useful resources:
 - https://github.com/mozilla/pdf.js#getting-the-code
 - https://github.com/mozilla/pdf.js/wiki/Contributing

arvind0598:
I need some help with `gulp browsertest`. When I run it, there are a lot of outputs that say `WARNING: no reference snapshot ref/linux/Firefox/issue8229/1.png`, or something similar. 

Then the following happens, and the test fails with 546 errors. I'm not sure how to get this working so that I can start working myself. `unittest` and `fonttest` pass completely.

`TEST-UNEXPECTED-FAIL | test failed Chrome has not responded in 120s`
`TEST-UNEXPECTED-FAIL | test failed Firefox has not responded in 120s`

I'm running Firefox 68.0.2 and Chromium 76.0.3809.100-0ubuntu0.18.04.1 on Ubuntu 18.04.3

Snuffleupagus:
[...]

PR #11459, opened 2019-12-31
Convert the `src/pdf.js` and `src/pdf.worker.js` files to use standard `import`/`export` statements; Re-factor `setPDFNetworkStreamFactory`, in src/display/api.js, to take an asynchronous function 

Snuffleupagus:

I've looked briefly at issue #10965, and have concluded that given the changes in SystemJS the upgrade may a bit tricky (since for one I wasn't able to *easily* find a Babel plugin for the new version).

The ideal solution, as outlined briefly in https://github.com/mozilla/pdf.js/issues/10965#issuecomment-568252833, would probably be to try and avoid locking us into these sort of dev-dependencies in the future if at all possible. To that end, I'm thus proposing that we start migrating the code away from `require` statements (unless protected by pre-processor and/or run-time checks) and towards standardized `import`s instead wherever possible.

*Note:* I've not tested it yet, but it *may* even be possible to get rid of SystemJS everywhere *except* in the worker (in development mode) right now. Doing so would also allow us to slowly start using modern ECMAScript features, which the SystemJS Babel plugin doesn't support.

### DIFF
--- a/src/display/api_compatibility.js
+++ b/src/display/api_compatibility.js
@@ -13,10 +13,10 @@
  * limitations under the License.
  */
 
+import { isNodeJS } from "../shared/is_node.js";
+
 const compatibilityParams = Object.create(null);
 if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) {
-  const { isNodeJS } = require("../shared/is_node.js");
-
   const userAgent =
     (typeof navigator !== "undefined" && navigator.userAgent) || "";
   const isIE = /Trident/.test(userAgent);
@@ -42,4 +42,4 @@ if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) {
   })();
 }
 
-exports.apiCompatibilityParams = Object.freeze(compatibilityParams);
+export const apiCompatibilityParams = Object.freeze(compatibilityParams);
--- a/src/pdf.js
+++ b/src/pdf.js
@@ -30,7 +30,24 @@ var pdfjsDisplaySVG = require("./display/svg.js");
 const pdfjsDisplayWorkerOptions = require("./display/worker_options.js");
 const pdfjsDisplayAPICompatibility = require("./display/api_compatibility.js");
 
-if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) {
+if (typeof PDFJSDev === "undefined" || !PDFJSDev.test("PRODUCTION")) {
+  const streamsPromise = Promise.all([
+    SystemJS.import("pdfjs/display/network.js"),
+    SystemJS.import("pdfjs/display/fetch_stream.js"),
+  ]);
+  pdfjsDisplayAPI.setPDFNetworkStreamFactory(params => {
+    return streamsPromise.then(streams => {
+      const [{ PDFNetworkStream }, { PDFFetchStream }] = streams;
+      if (
+        pdfjsDisplayDisplayUtils.isFetchSupported() &&
+        pdfjsDisplayDisplayUtils.isValidFetchUrl(params.url)
+      ) {
+        return new PDFFetchStream(params);
+      }
+      return new PDFNetworkStream(params);
+    });
+  });
+} else if (PDFJSDev.test("GENERIC")) {
   const { isNodeJS } = require("./shared/is_node.js");
   if (isNodeJS) {
     const PDFNodeStream = require("./display/node_stream.js").PDFNodeStream;
--- a/src/pdf.js
+++ b/src/pdf.js
@@ -12,36 +12,66 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint-disable no-unused-vars */
 
-"use strict";
+import {
+  addLinkAttributes,
+  getFilenameFromUrl,
+  isFetchSupported,
+  isValidFetchUrl,
+  LinkTarget,
+  loadScript,
+  PDFDateString,
+  RenderingCancelledException,
+} from "./display/display_utils.js";
+import {
+  build,
+  getDocument,
+  LoopbackPort,
+  PDFDataRangeTransport,
+  PDFWorker,
+  setPDFNetworkStreamFactory,
+  version,
+} from "./display/api.js";
+import {
+  CMapCompressionType,
+  createObjectURL,
+  createPromiseCapability,
+  createValidAbsoluteUrl,
+  InvalidPDFException,
+  MissingPDFException,
+  NativeImageDecoding,
+  OPS,
+  PasswordResponses,
+  PermissionFlag,
+  removeNullCharacters,
+  shadow,
+  UnexpectedResponseException,
+  UNSUPPORTED_FEATURES,
+  Util,
+  VerbosityLevel,
+} from "./shared/util.js";
+import { AnnotationLayer } from "./display/annotation_layer.js";
+import { apiCompatibilityParams } from "./display/api_compatibility.js";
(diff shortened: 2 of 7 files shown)


src/display/api_compatibility.js
```diff
@@ -13,10 +13,10 @@
  * limitations under the License.
  */
 
-let compatibilityParams = Object.create(null);
-if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) {
-  const { isNodeJS } = require("../shared/is_node");
```
timvandermeij: The commit message says "[...] `require` statements in code-paths not protected by pre-processor and/or run-time checks", but this one is inside a pre-processor check. Is this safe to change?

src/display/api_compatibility.js
```diff
@@ -28,18 +28,18 @@ if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) {
...
     if (isNodeJS) {
-      compatibilityParams.disableFontFace = true;
-      compatibilityParams.nativeImageDecoderSupport = "none";
+      apiCompatibilityParams.disableFontFace = true;
+      apiCompatibilityParams.nativeImageDecoderSupport = "none";
     }
   })();
 }
 
-exports.apiCompatibilityParams = Object.freeze(compatibilityParams);
```
timvandermeij: Do you happen to know why we had `Object.freeze` here and in the viewer file in the first place? I can only imagine that it should not be mutated later on, but AFAICT we never do that anyway.

src/pdf.worker.js
```diff
@@ -12,13 +12,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint-disable no-unused-vars */
 
-"use strict";
```
timvandermeij: Why exactly is this removed (here and in the `pdf.js` file)?

src/pdf.js
```diff
@@ -30,11 +30,26 @@ var pdfjsDisplaySVG = require("./display/svg.js");
...
+
+    if (
+      pdfjsDisplayDisplayUtils.isFetchSupported() &&
+      pdfjsDisplayDisplayUtils.isValidFetchUrl(params.url)
+    ) {
+      return new PDFFetchStream(params);
+    }
+    return new PDFNetworkStream(params);
+  });
+} else if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("GENERIC")) {
```
timvandermeij: The `typeof PDFJSDev !== "undefined"` condition can be removed here and in the case below since the case above already handles that.

src/pdf.js
```diff
@@ -30,11 +30,26 @@ var pdfjsDisplaySVG = require("./display/svg.js");
 let pdfjsDisplayWorkerOptions = require("./display/worker_options.js");
 let pdfjsDisplayAPICompatibility = require("./display/api_compatibility.js");
 
-if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) {
+if (typeof PDFJSDev === "undefined" || !PDFJSDev.test("PRODUCTION")) {
+  pdfjsDisplayAPI.setPDFNetworkStreamFactory(async params => {
+    const [{ PDFNetworkStream }, { PDFFetchStream }] = await Promise.all([
+      SystemJS.import("../src/display/network.js"),
```
timvandermeij: It's not entirely clear to me why this new code block is necessary since it adds new `SystemJS` usage. I can see that we're now not running the `require` statement for the development viewer case anymore, but since it's being exchanged by direct `SystemJS` usage I'm not sure what the advantage is here?

timvandermeij (commented):
I like the idea of avoiding `SystemJS` as much as possible by using standard imports/exports instead. I do have a few questions about the implementation, but in general it looks good!

src/display/api_compatibility.js
Snuffleupagus: This particular check passes in developer mode, note the `typeof PDFJSDev === "undefined" || ` part, and generally I do think that it's a good idea to run as much of the code as possible during development to help catch issues earlier (which is why I didn't change the condition here).

So yes, this change should be OK and the viewer still works locally with it :-)

src/display/api_compatibility.js
Snuffleupagus: > Do you happen to know why we had `Object.freeze` here and in the viewer file in the first place?

Yes, to avoid the object being mutated by accident since you can still add/remove/change properties even in a `const` object.
The better question is why I changed this here, to which there's no good answer; I'll revert these changes, well spotted!

src/pdf.worker.js
Snuffleupagus: Because using native `import` statements will automatically put the code into strict mode, according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import, and looking at the built files (the `mozcentral` build target) there's also still `"use strict";` lines everywhere as expected.

src/pdf.js
Snuffleupagus: Yes, that seems reasonable.

src/pdf.js
Snuffleupagus: The main idea in this PR was to reduce usage of `require` in code which runs in development mode, and the idea here is that we'll be able to easily converted these ones to native [dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Dynamic_Imports) later.

Another "fun" follow-up that will be necessary is to ensure that *all* import statements in the entire code-base have a `.js` extension, since otherwise the native `import` will fail with MIMI type errors...

timvandermeij:
This needs a rebase after the other patches. I'll make a separate issue for the bot issues.

Snuffleupagus:
> I'll make a separate issue for the bot issues.

Thanks; we obviously need to hold off on landing this until there's successful test runs in all configurations.

Since the tests passed with an earlier version of the PR, I *really* hope that there wasn't some kind of problem introduced after that; however all tests did pass successfully locally just now.

timvandermeij:
Could you rebase this patch so we can run the tests again?

Snuffleupagus:
> Could you rebase this patch so we can run the tests again?

We can probably do the new release first, before moving forward with this (there's probably a couple of other PRs that should be prioritized instead).

timvandermeij:
I forgot about this PR, but given that the release has been made and the bots should be up and running again, would it be possible to get this rolling again? I think reducing the dependency on SystemJS would be a good thing.

Snuffleupagus:
> [...] would it be possible to get this rolling again?

Sure I can try, but there's one thing that I really do *not* like with it though:
By making the `setPDFNetworkStreamFactory` (and related) functionality asynchronous there was an easily spotted regression in test the run-times, which I've not yet figured out how to avoid (or at least minimize).

Snuffleupagus:
> Sure I can try, but there's one thing that I really do _not_ like about it though:
> By making the `setPDFNetworkStreamFactory` (and related) functionality asynchronous there was an easily spotted regression in test run-times, which I've not yet figured out how to avoid (or at least minimize).

Having re-factored the first commit slightly (reducing async-ness), it turns out that this was actually a red herring because on its own there's no particularly noticeable slowdown with *only* the first commit applied.

Rather it seems that the changes in the second commit, i.e. the use of standard `import`/`export`-statements, is the cause of the problems. Essentially, with those changes the top-level exports in the *built* `pdf.js` and `pdf.worker.js` files now uses a slightly different (and by the looks of it slower) format.
It's possible that this could be fixed by some Webpack option, or you could maybe even hack-around it by using a regular expression on the built files to convert the top-level exports; however I'm slightly out of my depth here w.r.t. the Webpack bundling unfortunately.

PR #11463, opened 2020-01-01
[Firefox] Ensure that telemetry will be correctly recorded even when the PDF header is missing (bug 1606566)

Snuffleupagus:
As suggested in the bug we'll now always fallback to "other", which should be fine considering that that value is listed in the histogram definition; see https://searchfox.org/mozilla-central/rev/331f0c3b25089c9a16be65f4dc8c601aeaac8cc4/toolkit/components/telemetry/Histograms.json#9097-9106

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1606566

/cc @Rob--W

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1314,7 +1314,28 @@ const PDFViewerApplication = {
           PDFJSDev.test("FIREFOX || MOZCENTRAL")
         ) {
           // Telemetry labels must be C++ variable friendly.
-          const versionId = `v${info.PDFFormatVersion.replace(".", "_")}`;
+          let versionId = "other";
+          // Keep these in sync with mozilla central's Histograms.json.
+          const KNOWN_VERSIONS = [
+            "1.0",
+            "1.1",
+            "1.2",
+            "1.3",
+            "1.4",
+            "1.5",
+            "1.6",
+            "1.7",
+            "1.8",
+            "1.9",
+            "2.0",
+            "2.1",
+            "2.2",
+            "2.3",
+          ];
+          if (KNOWN_VERSIONS.includes(info.PDFFormatVersion)) {
+            versionId = `v${info.PDFFormatVersion.replace(".", "_")}`;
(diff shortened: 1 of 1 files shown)


web/app.js
```diff
@@ -1314,7 +1314,10 @@ const PDFViewerApplication = {
           PDFJSDev.test("FIREFOX || MOZCENTRAL")
         ) {
           // Telemetry labels must be C++ variable friendly.
-          const versionId = `v${info.PDFFormatVersion.replace(".", "_")}`;
+          let versionId = "other";
+          if (info.PDFFormatVersion) {
```
Rob--W: Strictly speaking, we should probably verify that the version is one of the accepted values before assigning to the variable. The PDFFormatVersion field can have any value, as seen at https://github.com/mozilla/pdf.js/blob/81af207428b2fafdf4373ad6d2689928f72ee2d6/src/core/document.js#L513
and https://github.com/mozilla/pdf.js/blob/81af207428b2fafdf4373ad6d2689928f72ee2d6/src/core/document.js#L632

Rob--W (approved):
r+ because it fixes the reported bug. It would be nice if you can fix the following issue as well, though.

web/app.js
Snuffleupagus: Fair point, I've added validation of the same format as for the generator just below; thanks for the review!

Rob--W: The existing format looks needlessly complicated. Let's change it to:

```
if (KNOWN_VERSIONS.includes(info.PDFFormatVersion)) {
  versionId = `...`;
}
```

Snuffleupagus: Fixed; now there's less copy-and-paste in the patch :-)

Rob--W (approved):

Rob--W:
Thanks for the quick patch for my bug report.

Happy New Year!

PR #11465, opened 2020-01-02
Ensure that all `import` and `require` statements, in the entire code-base, have a `.js` file extension

Snuffleupagus:
In order to eventually get rid of SystemJS and start using native `import`s instead, we'll need to provide "complete" file identifiers since otherwise there'll be MIME type errors when attempting to use `import`.


timvandermeij (approved):

timvandermeij:
Looks good! Let's hope the Windows bot starts working again, otherwise we'll have to kick it.

PR #11466, opened 2020-01-02
[Firefox] Slightly simplify how the generator is determined, when sending telemetry data from the viewer

Snuffleupagus:
When working on PR #11463 I couldn't help thinking that the `Array.prototype.some` callback function, used when determining the generator, was somewhat difficult to read with its partly unused and strangely named parameters.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1365,16 +1365,16 @@ const PDFViewerApplication = {
             "fpdf",
           ];
           if (info.Producer) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you! I also had to re-read this a few times before I understood the old code; the new one is indeed easier to read.

PR #11482, opened 2020-01-07
Convert `src/core/jpg.js` to use the `readUint16` helper function in `src/core/core_utils.js`, rather than re-implementing it twice

Snuffleupagus:

### DIFF
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -14,6 +14,7 @@
  */
 
 import { assert, BaseException, warn } from "../shared/util.js";
+import { readUint16 } from "./core_utils.js";
 
 class JpegError extends BaseException {
   constructor(msg) {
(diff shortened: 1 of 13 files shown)


Snuffleupagus:
There's no images, of any kind, in the failing reference test above.
However, Firefox was just updated to version 73 on the bot and we've seen before that updates can change how non-embedded CJK fonts render.

src/core/jpg.js
```diff
@@ -148,8 +149,9 @@ var JpegImage = (function JpegImageClosure() {
         var nextByte = data[offset++];
         if (nextByte) {
           if (nextByte === 0xdc && parseDNLMarker) {
-            offset += 2; // Skip data length.
-            const scanLines = (data[offset++] << 8) | data[offset++];
+            offset += 2; // Skip marker length.
+
+            const scanLines = readUint16(data, offset);
```
timvandermeij: Is this equivalent? `readUint16` does `(data[offset] << 8) | data[offset + 1]` whereas the previous code did `(data[offset++] << 8) | data[offset++]`, so the offset is incremented for the bitwise shift too.

Snuffleupagus: You're correct, in order for the new code to be completely identical to the old one there should be a `offset += 2;` line afterwards.
In this *particular* case it really doesn't matter since there's an Error thrown below, but generally this probably isn't great if this code is ever re-factored; I'll just fix this to avoid any future issues.

timvandermeij (approved):

timvandermeij:
Nice refactoring!

PR #11488, opened 2020-01-08
Enable the `no-nested-ternary` ESLint rule in the `web/` directory

Snuffleupagus:
This rule is already enabled in mozilla-central, and helps avoid some confusing formatting[1], see https://searchfox.org/mozilla-central/rev/a92ed79b0bc746159fc31af1586adbfa9e45e264/tools/lint/eslint/eslint-plugin-mozilla/lib/configs/recommended.js#209-210

Since some cases may not be aren't entirely straightforward to convert, and some we probably don't want to change either[2], this patch is limited to the `web/` directory as a proof-of-concept.

Please find additional details about the ESLint rule at https://eslint.org/docs/rules/no-nested-ternary

---
[1] One example with *three* ternary statements: https://github.com/mozilla/pdf.js/blob/93aa613db7a874e6ed964f394241605d5586c142/src/core/fonts.js#L2623-L2630

[2] One example that should be whitelisted: https://github.com/mozilla/pdf.js/blob/93aa613db7a874e6ed964f394241605d5586c142/src/core/jbig2.js#L82-L92

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -136,6 +136,7 @@
     "new-cap": ["error", { "newIsCap": true, "capIsNew": false, }],
     "no-array-constructor": "error",
     "no-multiple-empty-lines": ["error", { "max": 1, "maxEOF": 0, "maxBOF": 1, }],
+    "no-nested-ternary": "off",
     "no-new-object": "error",
     "spaced-comment": ["error", "always", {
       "block": {
--- a/web/.eslintrc
+++ b/web/.eslintrc
@@ -7,6 +7,9 @@
     // Plugins
     "import/no-unresolved": ["error", { "ignore": ["pdfjs-lib"]}],
 
+    // Stylistic Issues
+    "no-nested-ternary": "error",
+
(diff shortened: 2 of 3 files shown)


timvandermeij (approved):

timvandermeij:
I agree with the decision at `mozilla-central` for enabling this since I also found the nested ternary statements harder to read than the new structure. Thanks!

Snuffleupagus:
> [...] since I also found the nested ternary statements harder to read than the new structure.

At least to my eye, the recent introduction of Prettier actually made some of these cases even *less* readable than before since any helpful indentation is now forbidden. That's actually what prompted me to start looking into this now, however a full conversion may take some time since it's not always clear immediately what the best solution is (and some cases we'll want to whitelist).

Also, we used to have *loads* of these nested ternary statements in the Colourspace/Image code, but converting that code to use `Uint8ClampedArray` got rid of most of those; thanks for landing this!

PR #11489, opened 2020-01-08
Remove the `FIREFOX` build flag, since it's completely unused and simplify a couple of `PDFJSDev` checks

Snuffleupagus:
- Remove the `FIREFOX` build flag, since it's completely unused

    After PR 9566, which removed all of the old Firefox extension code, the `FIREFOX` build flag is no longer used for anything.
    It thus seems to me that it should be removed, for a couple of reasons:
     - It's simply dead code now, which only serves to add confusion when looking at the `PDFJSDev` calls.
     - It used to be that `MOZCENTRAL` and `FIREFOX` was *almost* always used together. However, ever since PR 9566 there's obviously been no effort put into keeping the `FIREFOX` build flags up to date.
     - In the event that a new, Webextension based, Firefox addon is created in the future you'd still need to audit all `MOZCENTRAL` (and possibly `CHROME`) build flags to see what'd make sense for the addon.

- Simplify, and tweak, a couple of `PDFJSDev` checks

   This removes a couple of, thanks to preceeding code, unnecessary `typeof PDFJSDev` checks, and also fixes a couple of incorrectly implemented (my fault) checks intended for `TESTING` builds.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -88,7 +88,6 @@ var DEFINES = {
   TESTING: false,
   // The main build targets:
   GENERIC: false,
-  FIREFOX: false,
   MOZCENTRAL: false,
   CHROME: false,
   MINIFIED: false,
@@ -173,7 +172,6 @@ function createWebpackConfig(defines, output) {
     .readFileSync("./src/license_header_libre.js")
     .toString();
   var enableSourceMaps =
-    !bundleDefines.FIREFOX &&
     !bundleDefines.MOZCENTRAL &&
     !bundleDefines.CHROME &&
     !bundleDefines.TESTING;
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -1291,7 +1291,10 @@ var JpegImage = (function JpegImageClosure() {
     },
 
     getData({ width, height, forceRGB = false, isSourcePDF = false }) {
-      if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("TESTING && !LIB")) {
+      if (
+        typeof PDFJSDev === "undefined" ||
+        PDFJSDev.test("!PRODUCTION || TESTING")
+      ) {
         assert(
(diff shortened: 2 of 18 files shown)


Snuffleupagus:
I should probably split out the *third* commit into a separate PR, since it's not directly relevant to the other changes.

> While I briefly considered simply removing the pre-processor checks altogether, since doing so wouldn't break anything, that might look too out-of-place in non-`MOZCENTRAL` builds.

However, I'm still on the fence about the above since running the telemetry code in other builds shouldn't break anything now. Obviously it's not necessary in e.g. `GENERIC` builds, but it'd help ensure that this important code is better tested. @timvandermeij Opinions?

timvandermeij:
If the pre-processor checks don't do anything because the telemetry reporting in `GENERIC` build is replaced by a dummy method, I'm fine with removing them to simplify the code. I do wonder what we actually gain in terms of testing coverage because we would essentially be checking if calls to a dummy method can be made locally, correct? Actually testing the telemetry code locally is not really possible without a proper mock I think...

Snuffleupagus:
> I do wonder what we actually gain in terms of testing coverage because we would essentially be checking if calls to a dummy method can be made locally, correct?

That's correct, but my (perhaps unnecessary) idea was that running e.g. [this](https://github.com/mozilla/pdf.js/blob/40f531ee873f523e2991fd6c2ca00a336fdbfbdd/web/app.js#L1316-L1381) and [this](https://github.com/mozilla/pdf.js/blob/40f531ee873f523e2991fd6c2ca00a336fdbfbdd/web/app.js#L1995-L2001) code in more builds wouldn't be a bad idea!?

timvandermeij:
I see. That sounds reasonable so we at least cover that code.

timvandermeij (approved):

timvandermeij:
Good clean-up!

PR #11491, opened 2020-01-08
[DO NOT MERGE] update chrome.

brendandahl:


brendandahl:
Looks like most of the changes were minor text movement or slight offsets.

/botio makeref

timvandermeij:
Hm, that's the first time I've seen makeref ever fail. On the positive side, both bots are at least running the tests again. Now we only have to make sure that both have the most recent browser versions and the makeref works.

Snuffleupagus:
> Hm, that's the first time I've seen makeref ever fail.

Apparently the actual generation of reference images worked fine, but some of the forward-back-forward tests now seem to be intermittent in Google Chrome.

> Now we only have to make sure that both have the most recent browser versions

It seems fine to me, according to the excerpts below.

Linux bot:
```
User agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.42 Safari/537.36
User agent: Mozilla/5.0 (X11; Linux x86_64; rv:73.0) Gecko/20100101 Firefox/73.0
```

Windows bot:
```
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.42 Safari/537.36
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:73.0) Gecko/20100101 Firefox/73.0
```

Snuffleupagus:
> Hm, that's the first time I've seen makeref ever fail.

Yeah, it unfortunately seems that a large number of the reference tests are now *intermittent* in up-to-date versions of Google Chrome :-(

timvandermeij:
It only seems to happen in the `fbf` tests as far as I can tell. Let's retry.

/botio makeref

brendandahl:
We may need a way to disable FBF in chrome.

Snuffleupagus:
> We may need a way to disable FBF in chrome.

Actually, the problem seems to be worse than that :-(
Looking at the test results just above, after two back-to-back `makeref` calls in this PR, it almost looks as if (almost) the entire test-suite is now intermittently failing in Google Chrome!?

timvandermeij:
I know it's not ideal since we always want to test with the latest version, but is it possible to install an older version of Chrome, or will the auto-updater always update to the latest version? I remember that we disabled Chrome reference tests before, so if nothing else works, that may also be a (far from ideal) solution for now.

Snuffleupagus:
> but is it possible to install an older version of Chrome,
> ...
> I remember that we disabled Chrome reference tests before,

The whole point of updating the version was IMO to get support for recent e.g. JavaScript features, to not block future development; so in that case the latter suggestion *may* need to be considered.

brendandahl:
> > We may need a way to disable FBF in chrome.
> 
> Actually, the problem seems to be worse than that :-(
> Looking at the test results just above, after two back-to-back `makeref` calls in this PR, it looks as if (almost) the entire test-suite is now intermittently failing in Google Chrome!?

I may be missing something, but the failures look the same for the two `botio tests`. If makeref fails, it won't update the references, so I don't think we can say the whole thing is intermittent.

Snuffleupagus:
Sorry, I'm dumb; please disregard my earlier comment since I completely overlooked that the reference images failed to update...

Snuffleupagus:
> We may need a way to disable FBF in chrome.

Perhaps something like the following, which tries to disable things as little as possible?
https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:skip-FBF-tests-Chrome-makeref

timvandermeij:
> Perhaps something like the following, which tries to disable things as little as possible?

Yes, that looks reasonable. Since the other tests work, disabling only the FBF tests for Google Chrome sounds like the change with the least impact so we can start using the bots again. If we do this, we should make a follow-up issue to enable it later on again.

PR #11502, opened 2020-01-11
Fix a couple of cases where Prettier broke existing formatting (PR 11446 follow-up)

Snuffleupagus:
These two cases should have been whitelisted prior to re-formatting respectively had the comments fixed afterwards, however I unfortunately missed them because of the massive size of the diff.


timvandermeij (approved):

timvandermeij:
Yes, the size of the diff made this pretty much impossible to catch. Thanks!

PR #11503, opened 2020-01-11
Remove the unused `id` properties from page and thumbnail canvas/image DOM elements (issue 11499)

### LINKED ISSUE #11499: #pageN anchor not working on first page load
Attach (recommended) or Link to PDF file here: https://epa.oszk.hu/00800/00861/00032/pdf/02_kristo.pdf#page2

Configuration:
- Web browser and its version: Firefox ESR 68.4.1 & Developer Edition 73.0b3
- Operating system and its version: Debian 9 “stretch”
- PDF.js version: 2.2.178, 2.4.254 (the ones built in these versions of Firefox)
- Is a browser extension: no, core Firefox

1. Open the above link

The PDF opens with page 2.
PDF showing second page

What went wrong?
The PDF opens with page 1.
PDF showing first page

Additional details:
The function works after page load (when I enter a different anchor name in the URL bar, Firefox goes straight there, without reloading the page; that’s OK). When the anchor is present in the URL on page load, the following error pops up on the browser console:
```
PDFLinkService.navigateTo: "null" is not a valid destination array, for dest="page2". viewer.js:6570:17
```
(the line number is 6507 in v2.2.178, otherwise the same error message). No further stack trace is provided.

Snuffleupagus:
- Remove the unused `id` properties from page and thumbnail canvas/image DOM elements (issue 11499)

   As described in the issue, having a DOM element with `id=page2` (or any other number) will automatically cause that element to become linkable through the URL hash. That's currently leading to some confusing and outright wrong behaviour, since it obviously only works for pages that have been loaded and rendered.

   For PDF documents the only officially supported way to reference a particular page through the URL hash is using the `#page=2` format, which also works for all pages regardless if they're loaded or not.

   As far as I can tell there's nothing in the PDF.js default viewer that actually depends on the page/thumbnail `id` at this point in time, hence why I believe that this removal ought to be safe.
Just as a pre-caution this patch adds an `aria-label` to the page canvas, similar to the thumbnail canvas/image, to at least keep this information in the DOM.

 - Add helper functions to reduce unnecessary duplication when fetching l10n messages in `PDFThumbnailView`

Fixes #11499

### DIFF
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -559,7 +559,11 @@ class PDFPageView {
 
     const viewport = this.viewport;
     const canvas = document.createElement("canvas");
-    canvas.id = this.renderingId;
+    this.l10n
+      .get("page_canvas", { page: this.id }, "Page {{page}}")
+      .then(msg => {
+        canvas.setAttribute("aria-label", msg);
+      });
 
     // Keep the canvas hidden until the first draw callback, or until drawing
     // is complete when `!this.renderingQueue`, to prevent black flickering.
--- a/web/pdf_thumbnail_view.js
+++ b/web/pdf_thumbnail_view.js
@@ -258,11 +258,9 @@ class PDFThumbnailView {
     if (this.renderingState !== RenderingStates.FINISHED) {
       return;
     }
-    const id = this.renderingId;
     const className = "thumbnailImage";
 
     if (this.disableCanvasToImageConversion) {
-      this.canvas.id = id;
       this.canvas.className = className;
       this.l10n
         .get(
@@ -279,7 +277,6 @@ class PDFThumbnailView {
       return;
     }
     const image = document.createElement("img");
-    image.id = id;
     image.className = className;
     this.l10n
       .get(
--- a/web/pdf_thumbnail_view.js
+++ b/web/pdf_thumbnail_view.js
@@ -125,11 +125,9 @@ class PDFThumbnailView {
 
     const anchor = document.createElement("a");
     anchor.href = linkService.getAnchorUrl("#page=" + id);
-    this.l10n
-      .get("thumb_page_title", { page: id }, "Page {{page}}")
-      .then(msg => {
-        anchor.title = msg;
-      });
+    this._thumbPageTitle.then(msg => {
+      anchor.title = msg;
+    });
     anchor.onclick = function() {
       linkService.page = id;
       return false;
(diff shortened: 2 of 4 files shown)


web/pdf_thumbnail_view.js
```diff
@@ -452,8 +438,20 @@ class PDFThumbnailView {
     this._convertCanvasToImage();
   }
 
-  get pageId() {
-    return this.pageLabel !== null ? this.pageLabel : this.id;
+  get _thumbPageTitle() {
+    return this.l10n.get(
```
timvandermeij: Can we use the shadow utility so we only fetch this once? We have multiple call sites and the label should always be the same for the same thumbnail view, so this might make it a bit faster. Same below.

web/pdf_page_view.js
```diff
@@ -559,7 +559,11 @@ class PDFPageView {
 
     const viewport = this.viewport;
     const canvas = document.createElement("canvas");
-    canvas.id = this.renderingId;
```
timvandermeij: I vaguely remember that this is/was used for cleaning up canvases if they are out of view, but most likely this has changed over the years. Could you check if that still works as expected with this patch? If so, I also think this is now unused; if it's easy to find, perhaps the Git logs will tell us why this was ever introduced?

web/pdf_thumbnail_view.js
Snuffleupagus: Nope, since these will generally be called more than once for documents with page labels :-)

web/pdf_page_view.js
Snuffleupagus: Are you perhaps thinking of https://github.com/mozilla/pdf.js/blob/4729fdc0e930154c9e4c4688b6239ed4708f0e5e/web/base_viewer.js#L84-L118
In that case that shouldn't be a problem since it's dealing with `PDFPageView` instances.
I'll try to check the blame as well, but given how the code has been refactored over the years that may be a bit tricky...

Snuffleupagus: OK, I've tried going back approx. 10 years in the history and I'm not able to easily spot why these DOM elements need an `id`. Obviously I could be missing something, but it really seems unused.

timvandermeij: Alright, then it seems like it's indeed unused. Thank you for checking!

web/pdf_thumbnail_view.js
timvandermeij: Good point!

timvandermeij (approved):

timvandermeij:

PR #11508, opened 2020-01-13
Simplify the handling of unsupported/incorrect markers in `src/core/jpg.js`

Snuffleupagus:
- Re-factor the "incorrect encoding" check, since this can be easily achieved using the general `findNextFileMarker` helper function (with a suitable `startPos` argument).

 - Tweak a condition, to make it easier to see that the end of the data has been reached.

 - Add a reference test for issue 1877, since it's what prompted the "incorrect encoding" check.

### DIFF
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -1031,17 +1031,14 @@ var JpegImage = (function JpegImageClosure() {
             break;
 
           default:
-            if (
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

/botio makeref
(also for the other patch)

PR #11510, opened 2020-01-13
Skip failing FBF tests in Google Chrome when running `makeref` on the bots

Snuffleupagus:
Note that this will still allow the FBF tests to run locally, and also on the bots when invoked with `test`/`browsertest` (to not lose all the FBF test-coverage), but will no longer prevent `makeref` from running successfully on the bots.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -472,7 +472,7 @@ function checkEq(task, results, browser, masterMode) {
   }
 }
 
-function checkFBF(task, results, browser) {
+function checkFBF(task, results, browser, masterMode) {
   var numFBFFailures = 0;
   var round0 = results[0],
     round1 = results[1];
@@ -487,6 +487,20 @@ function checkFBF(task, results, browser) {
       continue;
     }
     if (r0Page.snapshot !== r1Page.snapshot) {
+      // The FBF tests fail intermittently in Google Chrome when run on the
+      // bots, ignoring `makeref` failures for now; see https://github.com/mozilla/pdf.js/pull/11491
+      if (masterMode && /chrom(e|ium)/i.test(browser)) {
+        console.log(
+          "TEST-SKIPPED | forward-back-forward test " +
+            task.id +
+            " | in " +
+            browser +
+            " | page" +
+            (page + 1)
+        );
+        continue;
+      }
+
       console.log(
         "TEST-UNEXPECTED-FAIL | forward-back-forward test " +
           task.id +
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Let's try this...
/botio makeref

Snuffleupagus:
The Windows bot seems to need a restart, or something...

The Linux bot *mostly* works now, but unfortunately there's a number of Google Chrome test failures that seem intermittent. However, we might have to live with that!?

Snuffleupagus:
The Windows bot seem to need a restart, or something...

The Linux bot *mostly* works now, but unfortunately there's a number of Google Chrome test failures that seem intermittent. However, we might just have to live with this and simply disregard those failures when looking at the test logs!?

brendandahl:
It appears chrome on windows and linux is no longer reproducible. Last time we were able to extract the rendering that had an issue and file a bug with chromium (which I think they fixed).

brendandahl:
/botio-linux makeref

Trying out headless mode to see if it's stable.

timvandermeij:
Since Google Chrome is so inconsistent now, should we just disable it for now so we can at least continue development, and try to enable it later on again using a follow-up issue?

Snuffleupagus:
> Trying out headless mode to see if it's stable.

Unfortunately that didn't seem to help :-(
It seems that headless also made the tests slower, unless I'm missing something.

> Since Google Chrome is so inconsistent now, should we just disable it entirely for now so we can at least continue development, and try to enable it later on again using a follow-up issue?

Either that, or we continue to run tests in Chrome and simply disregard these intermittent failures when looking at the results (although I suppose that this *could* increase the risk of actual regressions sneaking past a reviewer).

timvandermeij:
I noticed that the Linux bot seems down right now since I cannot reach previews or reference image comparisons.

How about we do the the following?

1. Switch both both from headless mode back to non-headless mode since that didn't make a difference in stability and seemed to have make the bots a bit slower.
2. Disable Chrome on both bots for now and make a follow-up issue to determine the case.
3. Restart both bots to start fresh.

We could replace step 2 with only disabling the FBF tests as done in this patch, but then we'll have to ignore the intermittent failures which might make it harder to see what actually failed. I'm on the fence about this step, so what do you think @brendandahl?

brendandahl:
I think we try to leave chrome chrome enabled for now and see how bad the intermittents are. If we're spending too much time validating tests then we can disable.

timvandermeij:
I agree that this approach is probably best. I'll make a follow-up ticket for the remaining intermittents and for re-enabling the FBF tests once Chrome stabilizes again.

PR #11515, opened 2020-01-16
Cache the fallback font dictionary on the `PartialEvaluator` (PR 11218 follow-up)

Snuffleupagus:
- Ensure that full clean-up is always run when handling the "Terminate" message in `src/core/worker.js`

   This is beneficial in situations where the Worker is being re-used, for example with fake workers, since it ensures that things like font resources are actually released.

 - Cache the fallback font dictionary on the `PartialEvaluator` (PR 11218 follow-up)

   This way we'll benefit from the existing font caching, and can thus avoid re-creating a fallback font over and over again during parsing.
(Thece changes necessitated the previous patch, since otherwise breakage could occur e.g. with fake workers.)

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -33,7 +33,14 @@ import {
   warn,
 } from "../shared/util.js";
 import { Catalog, ObjectLoader, XRef } from "./obj.js";
-import { Dict, isDict, isName, isStream, Ref } from "./primitives.js";
+import {
+  clearPrimitiveCaches,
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
Let's see if the bots are OK enough to actually run tests now...

/botio test

timvandermeij (approved):

timvandermeij:
Thanks!

PR #11517, opened 2020-01-17
Remove gloss on zoom dropdown in Safari

MichielDG:
Screenshot-pdf js-Safari

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -721,6 +721,7 @@ html[dir='rtl'] .dropdownToolbarButton {
   padding: 3px 2px 2px;
   border: none;
   background: rgba(0,0,0,0); /* Opera does not support 'transparent' <select> background */
+  -webkit-appearance: none; /* remove gloss in Safari */
 }
 
 .dropdownToolbarButton > select > option {


timvandermeij:
Looks good, but could you amend the commit message to `Remove gloss on zoom dropdown in Safari` to make it more specific and a bit more neutral?

Snuffleupagus:
Besides https://github.com/mozilla/pdf.js/pull/11517#issuecomment-575801291, there's also the question if this is actually a desirable change given that it looks worse in Google Chrome now (given that there's a Chrome extension built from the PDF.js project):

zoom

timvandermeij:
I didn't check in Google Chrome, but that is indeed a regression. Chrome has a lot more users and we also build an extension for Chrome, so a fix for the gloss in Safari should also only apply to Safari. If you can make this patch only apply to Safari, we can accept it, but otherwise we need to close it unfortunately.

timvandermeij:
I looked at this again and found that a solution for this can only be accepted if it only applies to Safari, hence it's unfortunately not mergeable in its current state. Closing because of that, but we'll be happy to look at a new PR if there is a way to only make this work for Safari. Thanks.

PR #11520, opened 2020-01-19
Update the `GlobalWorkerOptions.workerSrc` JSDoc comment

Snuffleupagus:
This particular JSDoc comment is fairly old and it also contains some now unrelated/confusing information.
The only way to *guarantee* that the PDF.js library works as expected is to correctly set the global `workerSrc`[1], hence giving the impression that the option isn't strictly necessary is thus incorrect.

---
[1] Since advertising the fallbackWorkerSrc functionality definitely seems like the *wrong* thing to do.

### DIFF
--- a/src/display/worker_options.js
+++ b/src/display/worker_options.js
@@ -25,12 +25,10 @@ GlobalWorkerOptions.workerPort =
     : GlobalWorkerOptions.workerPort;
 
 /**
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #11521, opened 2020-01-19
Remove the `supportsDocumentColors` warning in MOZCENTRAL builds (bug 844349 follow-up)

Snuffleupagus:
With https://bugzilla.mozilla.org/show_bug.cgi?id=844349 now being fixed in Firefox, the textLayer will now actually stay hidden as intended regardless of the browser settings.
Hence it should no longer be necessary to display the fallback bar, nor print a warning in the console, for documents which contains a textLayer.

Besides removing the `supportsDocumentColors` methods in the default viewer, we can also remove a now unused l10n string.


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #11523, opened 2020-01-20
Add a heuristic, in `src/core/jpg.js`, to handle JPEG images with a wildly incorrect SOF (Start of Frame) `scanLines` parameter (issue 10880)

### LINKED ISSUE #10880: Attached PDF not able View (7th page)
Configuration:
- Web browser and its version:   Chrome latest version
- Operating system and its version:   Windows 10
- PDF.js version:     tried in all versions
- Is a browser extension:  No

1.  Please try to open in Browser and 7th page 
2.

Snuffleupagus:
- Add basic validation of the `scanLines` parameter in JPEG images, before delegating decoding to the browser

   In some cases PDF documents can contain JPEG images that the native browser decoder cannot handle, e.g. images with DNL (Define Number of Lines) markers or images where the SOF (Start of Frame) marker contains a wildly incorrect `scanLines` parameter.
   Currently, for "simple" JPEG images, we're relying on native image decoding to *fail* before falling back to the implementation in `src/core/jpg.js`. In some cases, note e.g. issue 10880, the native image decoder doesn't outright fail and thus some images may not render.

   In an attempt to improve the current situation, this patch adds additional validation of the JPEG image SOF data to force the use of `src/core/jpg.js` directly in cases where the native JPEG decoder cannot be trusted to do the right thing.
   The only way to implement this is unfortunately to parse the *beginning* of the JPEG image data, looking for a SOF marker. To limit the impact of this extra parsing, the result is cached on the `JpegStream` instance and this code is only run for images which passed all of the pre-existing "can the JPEG image be natively rendered and/or decoded" checks.

   ---

   *Slightly off-topic:* Working on this *really* makes me start questioning if native rendering/decoding of JPEG images is actually a good idea.
   There's certain kinds of JPEG images not supported natively, and all of the validation which is now necessary isn't "free". At this point, in the `NativeImageDecoder`, we're having to check for certain properties in the image dictionary, parse the `ColorSpace`, and finally read the actual image data to find the SOF marker.
   Furthermore, we cannot just send the image to the main-thread and be done in the "JpegStream" case, but we also need to wait for rendering to complete (or fail) before continuing with other parsing.
   In the "JpegDecode" case we're even having to parse part of the image on the main-thread, which seems completely at odds with the principle of doing all heavy parsing in the Worker, and there's also a couple of potentially large (temporary) allocations/copies of TypedArray data involved as well.

- Add a heuristic, in `src/core/jpg.js`, to handle JPEG images with a wildly incorrect SOF (Start of Frame) `scanLines` parameter (issue 10880)

   *This whole patch feels somewhat arbitrary, and I'd be slightly worried about possibly breaking something else.*

   To limit the impact of these changes, we only re-parse JPEG images using a reduced `scanLines` value if and only if: An unexpected EOI (End of Image) marker was encountered during decoding of Scan data *and* the "actual" `scanLines` value is at least one order of magnitude smaller than expected.

Fixes #10880

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -512,7 +512,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           this.xref,
           resources,
           this.pdfFunctionFactory
-        )
+        ) &&
+        image.maybeValidDimensions
       ) {
         // These JPEGs don't need any more processing so we can just send it.
         return this.handler
--- a/src/core/image_utils.js
+++ b/src/core/image_utils.js
@@ -41,7 +41,8 @@ class NativeImageDecoder {
         this.xref,
         this.resources,
         this.pdfFunctionFactory
-      )
+      ) &&
+      image.maybeValidDimensions
     );
   }
 
--- a/src/core/jpeg_stream.js
+++ b/src/core/jpeg_stream.js
@@ -109,6 +109,135 @@ const JpegStream = (function JpegStreamClosure() {
     this.eof = true;
   };
 
+  Object.defineProperty(JpegStream.prototype, "maybeValidDimensions", {
+    get: function JpegStream_maybeValidDimensions() {
+      const { dict, stream } = this;
+      const dictHeight = dict.get("Height", "H");
+      const startPos = stream.pos;
+
+      let validDimensions = true,
+        foundSOF = false,
+        b;
+      while ((b = stream.getByte()) !== -1) {
+        if (b !== 0xff) {
+          // Not a valid marker.
+          continue;
+        }
+        switch (stream.getByte()) {
+          case 0xc0: // SOF0
+          case 0xc1: // SOF1
+          case 0xc2: // SOF2
+            // These three SOF{n} markers are the only ones that the built-in
+            // PDF.js JPEG decoder currently supports.
+            foundSOF = true;
+
+            stream.pos += 2; // Skip marker length.
+            stream.pos += 1; // Skip precision.
+            const scanLines = stream.getUint16();
+
+            // The "normal" case, where the image data and dictionary agrees.
+            if (scanLines === dictHeight) {
+              break;
+            }
+            // A DNL (Define Number of Lines) marker is expected,
+            // which browsers (usually) cannot decode natively.
+            if (scanLines === 0) {
+              validDimensions = false;
+              break;
+            }
+            // The dimensions of the image, among other properties, should
+            // always be taken from the image data *itself* rather than the
+            // XObject dictionary. However there's cases of corrupt images that
+            // browsers cannot decode natively, for example:
+            //  - JPEG images with DNL markers, where the SOF `scanLines`
+            //    parameter has an unexpected value (see issue 8614).
+            //  - JPEG images with too large SOF `scanLines` parameter, where
+            //    the EOI marker is encountered prematurely (see issue 10880).
+            // In an attempt to handle these kinds of corrupt images, compare
+            // the dimensions in the image data with the dictionary and *always*
+            // let the PDF.js JPEG decoder (rather than the browser) handle the
+            // image if the difference is larger than one order of magnitude
+            // (since that would generally suggest that something is off).
+            if (scanLines > dictHeight * 10) {
+              validDimensions = false;
+              break;
+            }
+            break;
+
+          case 0xc3: // SOF3
+          /* falls through */
+          case 0xc5: // SOF5
+          case 0xc6: // SOF6
+          case 0xc7: // SOF7
+          /* falls through */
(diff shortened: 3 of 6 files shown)


timvandermeij:
If I recall correctly, there was a discussion a while ago about only using our own JPEG decoder instead of native decoding because native decoding is not always possible. Looking at this patch, it requires quite a lot of additional code to be able to feature-detect if we can use the native decoder or not, so I agree with your analysis that we should perhaps wonder if we shouldn't just always use own own JPEG decoder. I don't know exactly what the pros/cons were of both solutions, but I can imagine that native decoding is faster. Do you happen to know what advantages/disadvantages may be? If so, we could perhaps reconsider this after having worked with native decoding for quite a while now.

Snuffleupagus:
> If I recall correctly, there was a discussion a while ago about only using our own JPEG decoder instead of native decoding because native decoding is not always possible.

Correct, there was a brief discussion about this in https://github.com/mozilla/pdf.js/pull/6984#issuecomment-184778220 and I also tried it in https://github.com/mozilla/pdf.js/pull/8637#issue-129873391.
At that time there were some failures, but I've since fixed those such that `src/core/jpg.js` can now successfully render *all* JPEG images in our test-suite (and I also [wrote a patch](https://github.com/mozilla/pdf.js/pull/9325) that improved the quality of some JPEG images).

At this point in time, there's two kinds of failure when running the reference tests with https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:rm-nativeImageDecoderSupport:
 - Changes which are basically imperceivable to the naked eye, where some pixels in the images are essentially off-by-one (in all components), which could probably be attributed to things such as different rounding behaviour in the browser/PDF.js JPEG decoder.
   This type of "failure" accounts for the vast majority of the total number (227 if I recall correctly) of changes in the reference tests.
 - Changes where the JPEG images now looks slightly blurrier than with the native browser decoder. For quite some time I've just assumed that this pointed to some sort of *general* deficiency in the `src/core/jpg.js` implementation, however I very recently discovered when comparing two viewers side-by-side that the differences *vanish at higher zoom levels* (usually closer to 200% is enough).
   Basically, if you disable [this downscaling in canvas.js](https://github.com/mozilla/pdf.js/blob/474fe1757e654e67d20a53774edb293aa073ce0b/src/display/canvas.js#L2354-L2393) (which is what happens when zooming in) the differences simply vanish!
   Hence I'm pretty satisfied that there's no *significant* problems with the `src/core/jpg.js` implementation, and the problems are rather tied to general image quality (for which there's already open bugs) when downscaling occurs; all-in-all I don't think that this should block a decision to use `src/core/jpg.js` exclusively!

> I don't know exactly what the pros/cons were of both solutions, but I can imagine that native decoding is faster.

That seems like a safe assumption, and one that I'd fully support. However, even if the native browser JPEG decoder is generally faster, actually using it involves sending data between the worker- and main-threads (and waiting for responses). Furthermore, as we both noted above, there's also a fair bit of overhead involved just in deciding what JPEG decoder to use (which probably isn't free).
There's also the question of memory usage with ["JpegDecode"](https://github.com/mozilla/pdf.js/blob/474fe1757e654e67d20a53774edb293aa073ce0b/src/display/api.js#L2330-L2392) (and the issue of main-thread image parsing), but I'm probably not the right person to evaluate this properly though...
 
In my *quick* testing though, basically running `SKIP_BABEL=true gulp browsertest` with `master` respectively [`rm-nativeImageDecoderSupport`](https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:rm-nativeImageDecoderSupport), I'm not seeing any *huge* and thus easily spotted regressions with `src/core/jpg.js`.
However, in order for the comparison to be fair we probably need land *this* PR first (obviously assuming that its `src/core/jpg.js` parts seem OK) since otherwise the comparison doesn't seem correct.

timvandermeij:
I'm still unsure about if we should take this approach. As you already noted there are downsides that we should consider, and in general the heuristic feels a bit fragile. Wouldn't it be better to check if #11601 works better instead, since it avoids quite a bit of overhead and ensures that we can always decode these kinds of images properly? I would be convinced with a benchmark of this PR versus #11601; if there are no visible changes in rendering and no significant regressions in runtime, I'd much rather use that approach instead. Would that be possible to check?

Snuffleupagus:
> Wouldn't it be better to check if #11601 works better instead

That'd only replace the *first* commit in this PR, but even with that done you'd *still* need the `src/core/jpg.js` changes to actually address the referenced issue.

timvandermeij (approved):

timvandermeij:
I have thought about this some more and I think it's indeed best to merge this now to get the issue resolved. We can look at the native decoding later. Thanks!

PR #11528, opened 2020-01-21
Hide .notdef glyphs in non-embedded Type1 fonts and don't ignore Widths

### LINKED ISSUE #11403: Invalid characters are printing in PDF when we have &nbsp - NBSP (non-breaking space)
[1.pdf](https://github.com/mozilla/pdf.js/files/3951841/1.pdf)

Configuration:
- PDF.js version: latest
- Is a browser extension:

1. Load the attached document
2. When it is rendered can notice junk characters in document bullet points 

should display without any junk characters
This works fine when we open same file in Adobe reader and in chrome pdf viewer
image

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):  tested with uploading document  this to demo viewer there also we see this problem

janpe2:
Fixes #11403
The PDF uses the non-embedded Type1 font Helvetica. Character codes 194 and 160 (`Â` and `NBSP`) are encoded as `.notdef`. We shouldn't show those glyphs because it seems that Acrobat Reader doesn't draw glyphs that are named `.notdef` in fonts like this.

In addition to testing `glyphName === ".notdef"`, we must test also `glyphName === ""` because the name `""` is used in `core/encodings.js` for undefined glyphs in encodings like `WinAnsiEncoding`.

The solution above hides the `Â` characters but now the replacement character (space) appears to be too wide. I found out that PDF.js ignores font's `Widths` array if the font has no `FontDescriptor` entry. That happens in #11403, so the default widths of Helvetica were used as specified in `core/metrics.js` and `.nodef` got a width of 333. The correct width is 0 as specified by the `Widths` array in the PDF. Thus we must never ignore `Widths`.

In local testing the rendering of test `bug1443140` has now changed. The PDF uses non-embedded fonts like Times-Roman and they have `Widths` but no `FontDescriptor`.

Even though the issue is labeled `4-printing`, the problems appear also when rendered on-screen.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -2945,6 +2945,8 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       var type = preEvaluatedFont.type;
       var maxCharIndex = composite ? 0xffff : 0xff;
       var properties;
+      const firstChar = dict.get("FirstChar") || 0;
+      const lastChar = dict.get("LastChar") || maxCharIndex;
 
       if (!descriptor) {
         if (type === "Type3") {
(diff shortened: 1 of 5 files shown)


Snuffleupagus:
I'll happily defer the actual review to @brendandahl, however one small suggestion:
Since you included *a lot* of really good information in https://github.com/mozilla/pdf.js/pull/11528#issue-365465123, mind adding that to the *commit message* as well such that it's available e.g. on the Git command line too (and not just at GitHub).

janpe2:
To justify my solution, see the source code of Apache PDFBox:
https://github.com/apache/pdfbox/blob/da6abfdd2f518fd9cbc1dfce330a7896fd7d2ed9/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/font/PDType1Font.java#L626-L633
(The comment for *embedded* fonts is not true.)

brendandahl:
/botio test
FYI: there will likely be some intermittent failures on chrome.

brendandahl (approved):

PR #11529, opened 2020-01-21
Display the CSS tab in the interactive examples (issue 11506 follow-up)

Snuffleupagus:
I just happened to notice that we've apparently never displayed the "CSS" tab [in the interactive examples](https://mozilla.github.io/pdf.js/examples/index.html#interactive-examples), which probably limits the number of users who'll notice the `direction: ltr;` CSS rule on the canvas :-)


timvandermeij (approved):

timvandermeij:
Good point!

PR #11533, opened 2020-01-23
Remove the "useOnlyCssZoom" (debugging) hash parameter

Snuffleupagus:
The debugging hash parameters[1] are intended to facilitate access to various tools/settings in PRODUCTION builds, protected by the `pdfBugEnabled` preference.
At this point, the remaining debugging hash parameters are mainly intended to allow access to the `PDFBug` tools and/or to quickly toggle certain larger features.

The "useOnlyCssZoom" functionality doesn't really seem to fit in with the rest of these hash parameters, since:
 - This is, comparatively speaking, a minor viewer-specific feature.
 - The zooming implementation will (almost) always fallback to CSS-only zooming, for any document, once the canvases becomes large enough. Hence, the majority of the CSS zooming feature can still be tested *directly* in any build of the viewer.
 - After the initial implementation, years ago, the CSS-only zooming code in question hasn't changed much (or even at all), i.e. it doesn't seem like an active development target.[2]
 - If the "useOnlyCssZoom" functionality was added today, it's unlikely that a hash parameter would've been added.
 - Last, but not least, there's also a `useOnlyCssZoom` preference hence toggling this functionality shouldn't be too difficult (e.g. if someone needs to hack on it).

All in all, I'm thus suggesting that we remove the "useOnlyCssZoom" hash parameter.

---
[1] Originally these hash parameters could be used directly in any build, which was bad since it would allow any link to potentially disable functionality and/or reduce performance.

[2] If it had seen active development over the years, I'd be *much* more inclined to keep the hash parameter.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -239,6 +239,7 @@ const PDFViewerApplication = {
   },
 
   /**
+   * Potentially parse special debugging flags in the hash section of the URL.
    * @private
    */
   async _parseHashParameters() {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good idea to clean up the hash parameters a bit. Thanks!

PR #11534, opened 2020-01-24
Enable the ESLint `prefer-const` rule globally (PR 11450 follow-up)

Snuffleupagus:


timvandermeij (approved):
Looks good when the merge conflict is resolved. Thank you for doing this!

Snuffleupagus:
> Looks good when the merge conflict is resolved. Thank you for doing this!

As always, thank you for reviewing these bigger patches!
(The fact that it mostly done automatically hopefully helps to simplify the review process a bit though...)

We *may* also want to look into enabling the `no-var` rule globally at some point, but I don't yet know if `--fix` it'll handle *enough* cases to make a mostly automatic conversion feasible.

PR #11535, opened 2020-01-25
Render Popup annotations last, once all other annotations have been rendered (issue 11362)

### LINKED ISSUE #11362: Annotations not shown in pdf.js web viewer
Attached PDF doesn't show popup annotations when view in https://mozilla.github.io/pdf.js/web/viewer.html

[Risk Advice Guide-Copy2.pdf](https://github.com/mozilla/pdf.js/files/3895671/Risk.Advice.Guide-Copy2.pdf)

Configuration:
- Web browser and its version:  all browsers, tested with Chrome and Firefox latest versions.
- Operating system and its version: all OS< tested with Windows and iOS
- PDF.js version: current (online version)

1. Visit https://mozilla.github.io/pdf.js/web/viewer.html
2. Upload the above file
3. Go to page 4,hover mouse over one of the checkbox

Annotation 2019-11-27 154335

The annotation did not pop up

Snuffleupagus:
In the current `AnnotationLayer` implementation, Popup annotations require that the parent annotation have already been rendered (otherwise they're simply ignored).
Usually the annotations are ordered, in the `/Annots` array, in such a way that this isn't a problem, however there's obviously no guarantee that all PDF generators actually do so. Hence we simply ensure, when rendering the `AnnotationLayer`, that the Popup annotations are handled last.

Fixes #11362

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -1417,10 +1417,26 @@ class AnnotationLayer {
    * @memberof AnnotationLayer
    */
   static render(parameters) {
+    const sortedAnnotations = [],
+      popupAnnotations = [];
+    // Ensure that Popup annotations are handled last, since they're dependant
+    // upon the parent annotation having already been rendered (please refer to
+    // the `PopupAnnotationElement.render` method); fixes issue 11362.
     for (const data of parameters.annotations) {
       if (!data) {
         continue;
       }
+      if (data.annotationType === AnnotationType.POPUP) {
+        popupAnnotations.push(data);
+        continue;
+      }
+      sortedAnnotations.push(data);
+    }
+    if (popupAnnotations.length) {
+      sortedAnnotations.push(...popupAnnotations);
+    }
+
+    for (const data of sortedAnnotations) {
       const element = AnnotationElementFactory.create({
         data,
         layer: parameters.div,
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
Submitting as a "Draft" PR for now, since I've not had time to create a reduced reference test yet.

/botio-linux preview

src/display/annotation_layer.js
```diff
@@ -1417,7 +1417,26 @@ class AnnotationLayer {
...
    */
   static render(parameters) {
+    const sortedAnnotations = [];
+    let popupAnnotations = null;
+    // Ensure that Popup annotations are handled last, since they're dependant
+    // upon the parent annotation having already been rendered (please refer to
+    // the `PopupAnnotationElement.render` method); fixes issue 11362.
     for (const data of parameters.annotations) {
+      if (data.annotationType === AnnotationType.POPUP) {
+        if (!popupAnnotations) {
```
timvandermeij: I would just initialize this list above, just like `sortedAnnotations`, so we don't need this `if` statement. It's only one list and this method is not called often.

Snuffleupagus: Sure, that works too!

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #11537, opened 2020-01-26
Send the `verbosity` level when setting up fake workers (issue 11536)

### LINKED ISSUE #11536: getDocument in pdfjs-dist package does not respect verbosity parameter
This is in Node.js using the [pdfjs-dist](https://www.npmjs.com/package/pdfjs-dist) npm package. That repo doesn't seem to accept issues so posting it here.

pdfjs-dist version: 2.2.228

```typescript
import {VerbosityLevel, getDocument} from 'pdfjs-dist';
(async () => {
    await getDocument({url: 'path/to/any/pdf.pdf', verbosity: VerbosityLevel.ERRORS}).promise;
})();
```

`'path/to/any/pdf.pdf'` must be a pdf that generates warnings. Warnings are still printed even with the VerbosityLevel.ERRORS in the code above.

NOTE: pdfjs-dist might just be out of date? I noticed it was last published 4 months ago.

I did some digging and discovered the following:

By adding the following console.log in `pdf.js`:
```javascript
  var PDFWorker =
  /*#__PURE__*/
  function () {
    function PDFWorker() {
        console.log('constructing PDFWorker', (0, _util.getVerbosityLevel)());
```
I can see `0` (VerbosityLevel.ERRORS) being printed out as you would hope.

However, in `pdf.worker.js`, where the warning function is actually defined, it's not reading `verbosity` from util at all, it's simply defined within that file as `var verbosity = VerbosityLevel.WARNINGS;`. So I'm guessing simply fixing that will fix this.

FWIW I'm getting `Warning: getFontFileType: Unable to detect correct font file Type/Subtype.` from the `function warn(msg) {` definition in `pdf.worker.js` and I simply want to ignore it.

Snuffleupagus:
Interestingly the viewer already seem to work correctly as-is, with workers disabled and a non-standard `verbosity` level.
Hence this is possibly Node.js specific, but given that the issue is lacking *both* the PDF file in question and a runnable test-case, so this patch is essentially a best-effort guess at what the problem could be.

Fixes #11536

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1883,6 +1883,10 @@ const PDFWorker = (function PDFWorkerClosure() {
           const messageHandler = new MessageHandler(id, id + "_worker", port);
           this._messageHandler = messageHandler;
           this._readyCapability.resolve();
+          // Send global setting, e.g. verbosity level.
+          messageHandler.send("configure", {
+            verbosity: this.verbosity,
+          });
         })
         .catch(reason => {
           this._readyCapability.reject(


timvandermeij (approved):

timvandermeij:
Thanks!

Issue #11526, opened 2020-01-21
Incorrect spacing of vertial japanese text


e-taka:
Attach (recommended) or Link to PDF file here:
https://www.nta.go.jp/taxes/tetsuzuki/shinsei/shinkoku/zoyo/yoshiki2019/pdf/006.pdf

Configuration:
- Web browser and its version: Firefox 72.0.1
- Operating system and its version: macOS 10.14
- PDF.js version: 2.2.228
- Is a browser extension: No

Steps to reproduce the problem:
1.  Open this link https://www.nta.go.jp/taxes/tetsuzuki/shinsei/shinkoku/zoyo/yoshiki2019/pdf/006.pdf

What is the expected behavior? (add screenshot)
By Acrobat Reader, open above link:
<img width="70" alt="スクリーンショット 2020-01-21 10 17 53" src="https://user-images.githubusercontent.com/1781921/72767625-f5078500-3c37-11ea-801f-817ae7e9350a.png">

What went wrong? (add screenshot)
![スクリーンショット 2020-01-21 10 18 12](https://user-images.githubusercontent.com/1781921/72767634-fa64cf80-3c37-11ea-854a-33e0fac2d63f.png)

Snuffleupagus:
Possibly a duplicate of #7687.

tamuratak:
The cause is that `fontDirection` in `src/display/canvas.js` is always `1` even when the writing mode is vertical. I have confirmed that the following patch solves this problem and #7687. However, this patch is not an appropriate solution since it leaves `fontDirection` as it is. I am not sure what the purpose of `fontDirection` is.  Is it for right-to-left texts?

```diff
diff --git a/src/display/canvas.js b/src/display/canvas.js
index eaec1a1d..60af1309 100644
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -1719,7 +1719,7 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
           }
         }
 
-        var charWidth = width * widthAdvanceScale + spacing * fontDirection;
+        var charWidth = width * widthAdvanceScale - spacing * spacingDir;
         x += charWidth;
 
         if (restoreNeeded) {
```

I hope this would help your development.

Snuffleupagus:
> I have confirmed that the following patch using `spacingDir` instead of `fontDirection` solved #11526 and #7687.

Did it pass all existing PDF.js tests too?
[...]

Issue #7687, opened 2016-10-04
Some Japanese characters overlap


futot1:
Link to PDF file (or attach file here):
[610510_p37.pdf](https://github.com/mozilla/pdf.js/files/507848/610510_p37.pdf)

Configuration:
- Web browser and its version: Firefox
- Operating system and its version: Windows 7
- PDF.js version: Current
- Is an extension:

Steps to reproduce the problem:
1. I accessed http://mozilla.github.io/pdf.js/web/viewer.html
2. Opened attached PDF

What is the expected behavior? (add screenshot)

What went wrong? (add screenshot)
![pdf_error](https://cloud.githubusercontent.com/assets/7111820/19065750/3f8cb246-8a50-11e6-879d-dba35461b6ea.png)
Some characters overlapping as indicated.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

timvandermeij:
Fixed by #11540.

PR #11540, opened 2020-01-28
Fix text spacing with vertical fonts. #7687 and #11526.

tamuratak:
Fix text spacing with vertical fonts. #7687 and #11526.

Since we subtract here,
https://github.com/mozilla/pdf.js/blob/474fe1757e654e67d20a53774edb293aa073ce0b/src/display/canvas.js#L1729-L1731
when the writing mode is vertical, we have to reverse the sign of `spacing`.

I will reedit the commit message later.

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -1719,7 +1719,12 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
           }
         }
 
-        var charWidth = width * widthAdvanceScale + spacing * fontDirection;
+        var charWidth;
+        if (vertical) {
+          charWidth = width * widthAdvanceScale - spacing * fontDirection;
+        } else {
+          charWidth = width * widthAdvanceScale + spacing * fontDirection;
+        }
         x += charWidth;
 
         if (restoreNeeded) {
--- a/test/pdfs/issue11526.pdf.link
+++ b/test/pdfs/issue11526.pdf.link
@@ -1,0 +1,1 @@
+https://web.archive.org/web/20200210200830/https://www.nta.go.jp/taxes/tetsuzuki/shinsei/shinkoku/zoyo/yoshiki2019/pdf/006.pdf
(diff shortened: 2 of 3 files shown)


timvandermeij:
FYI: we expect test failures because of recent bot problems, so they might not all be related to your PR.

timvandermeij:
The patch looks good, but it should include a (linked) reference test to prevent regressions from happening later on. You can use https://github.com/mozilla/pdf.js/commit/ae2cc9119b02bb6d81d677f76af9ce9441394262 as inspiration for how to do that. The idea is to add a file at `test/pdfs/issue11526.pdf.link` containing an always-reachable link of this PDF file to the Web Archive. You can use https://web.archive.org/web/20200210200830/https://www.nta.go.jp/taxes/tetsuzuki/shinsei/shinkoku/zoyo/yoshiki2019/pdf/006.pdf for that. Finally, you add an entry like https://web.archive.org/web/20200210200830/https://www.nta.go.jp/taxes/tetsuzuki/shinsei/shinkoku/zoyo/yoshiki2019/pdf/006.pdf to the test manifest so the test is run.

Snuffleupagus:
>  Finally, you add an entry like [ae2cc91#diff-b19c886c212474d3b37d6026a795a246R2686-R2693](https://github.com/mozilla/pdf.js/commit/ae2cc9119b02bb6d81d677f76af9ce9441394262#diff-b19c886c212474d3b37d6026a795a246R2686-R2693) to the test manifest so the test is run.

Well, the test in this case needs to `"type": "eq"` to actually test what you want here :-P

---

Also, the commit message seems to link the wrong part of the specification since section "8.9.7 Inline Images" probably isn't correct :-)

tamuratak:
Thank you. I have added a test as requested.

timvandermeij (approved):

timvandermeij:

/botio makeref

PR #11545, opened 2020-01-28
Use the native `URL.createObjectURL` method in `web/debugger.js`

Snuffleupagus:
There's no particular reason for using the PDF.js helper function `createObjectURL`[1] in the debugger, instead of the native `URL.createObjectURL` directly, for a couple of reasons:
 - The relevant code-path only applies to fonts loaded with the Font Loading API, and this isn't supported in IE anyway.
 - The debugger can, since quite some time, not even be loaded in IE any more.
 - General support for IE is now limited, and there's no guaratee that everything actually works.

---
[1] It provides a fallback for browsers with broken `Blob` support, which as usual means Internet Explorer :-P

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -38,7 +38,6 @@ import {
 import { AppOptions, OptionKind } from "./app_options.js";
 import {
   build,
-  createObjectURL,
   getDocument,
   getFilenameFromUrl,
   GlobalWorkerOptions,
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Always good to start using native methods; thanks!

PR #11547, opened 2020-01-29
Use fewer multiplications in `JpegImage._convertCmykToRgb`

Snuffleupagus:
*Note:* This is inspired by PR #5473, which made similar changes for another kind of JPEG data.

Since the implementation in `src/core/jpg.js` only supports 8-bit data, as opposed to similar code in `src/core/colorspace.js`, the computations can be further simplified since the `scale` is always constant.
By updating the coefficients, effectively inlining the `scale`, we'll thus avoid *four* multiplications for each loop iteration.

Unfortunately I wasn't able, based on a quick look through the test-files, to find a sufficiently *large* CMYK JPEG image in order for these changes to really show up in benchmark results. However, when testing the `cmykjpeg.pdf` manually there's a total of `120 000` fewer multiplication with this patch.


timvandermeij:
@brendandahl Could you perhaps fix the Windows bot failure?

timvandermeij (approved):

timvandermeij:
Good idea!

Issue #10988, opened 2019-07-19
SVG font kerning issue


johnkingsley:
Attach (recommended) or Link to PDF file here:
[font_issue.pdf](https://github.com/mozilla/pdf.js/files/3408967/font_issue.pdf)
[font_issue.svg.zip](https://github.com/mozilla/pdf.js/files/3408968/font_issue.svg.zip)

Configuration:
- Web browser and its version: Version 56.0.2924.76 (64-bit)
- Operating system and its version: Linux Mint 18.3
- PDF.js version: v2.2.228
- Is a browser extension: no

Steps to reproduce the problem:
1. Using pdf.js create the SVG file from the attached PDF file
2. Look at the SVG file in Chrome. Compare with PDF result.

What is the expected behavior? (add screenshot)
When viewing the PDF file, it looks good:
![good_output](https://user-images.githubusercontent.com/6750405/61502440-ef3b9900-a9a1-11e9-9f72-9e319bd3a75a.png)

What went wrong? (add screenshot)
When viewing the SVG file, the characters are too closely spaced:
![bad_output](https://user-images.githubusercontent.com/6750405/61502446-f498e380-a9a1-11e9-82d5-53b01ad9fbbd.png)

The problem seems to be related to when TrueType fonts are used.
Same issue occurs on both Linux and Windows.
The SVG looks bad in Chrome, Firefox and MS Edge.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

timvandermeij:
Fixed by #11553.

PR #11553, opened 2020-01-30
Fix the horizontal scaling of texts with SVG backend. #10988

tamuratak:
Fix the horizontal scaling of texts with SVG backend. #10988.

In the same way as `src/display/canvas.js`.

https://github.com/mozilla/pdf.js/blob/517ccb7a394148942616433d49782cc8126a49fe/src/display/canvas.js#L1624

スクリーンショット 2020-01-31 7 08 51


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

Issue #11555, opened 2020-01-31
The vertical writing mode with horizontal scaling is not rendered correctly


tamuratak:
Attach (recommended) or Link to PDF file here:

[hello-abc.pdf](https://github.com/mozilla/pdf.js/files/4139771/hello-abc.pdf)

Configuration:
- Web browser and its version: Firefox 72.0.2
- Operating system and its version: macOS High Sierra 10.13.6
- PDF.js version: 2.4.321
- Is a browser extension: No

Steps to reproduce the problem:
1. Load the attached pdf file.
2. Characters overlap.

What is the expected behavior? (add screenshot)

![スクリーンショット 2020-01-31 23 16 43](https://user-images.githubusercontent.com/10665499/73546004-cc755b80-447f-11ea-8c7b-941629b72caf.png)

What went wrong? (add screenshot)

![スクリーンショット 2020-01-31 23 17 22](https://user-images.githubusercontent.com/10665499/73546056-e1ea8580-447f-11ea-98be-ce6835247301.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): https://mozilla.github.io/pdf.js/web/viewer.html

timvandermeij:
Fixed by #11556.

PR #11556, opened 2020-01-31
Fix the vertical writing mode with horizontal scaling. #11555.

tamuratak:
Fix the vertical writing mode with horizontal scaling. #11555.
It is not valid to multiply `textHScale` when the writing mode is vertical.

See 9.4.4 Text Space Details, https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G8.1694762

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -1732,7 +1732,7 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
         }
       }
       if (vertical) {
-        current.y -= x * textHScale;
+        current.y -= x;
       } else {
         current.x += x * textHScale;
       }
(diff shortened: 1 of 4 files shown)


timvandermeij:
Looks good, but like the other PR, this one also needs a regression test. Fortunately you have a reduced test case in the original issue which you can use for this. You can insert it like in the other PR, except that it's not a link file but actually the PDF file itself; see https://github.com/mozilla/pdf.js/pull/11528/files for an example. In comparison to your other PR, you remove the `"link"` part and add the PDF file to `.gitignore`.

tamuratak:
I have added a test as requested.

timvandermeij (approved):

timvandermeij:
Nice work; thank you for fixing this!

/botio makeref

PR #11557, opened 2020-02-01
Avoid re-calculating the `xScaleBlockOffset` when not necessary in `JpegImage._getLinearizedBlockData`

Snuffleupagus:
As can be seen in the code, the `xScaleBlockOffset` typed array doesn't depend on the actual image data but only on the width and x-scale. The width is obviously consistent for an image, and it turns out that in practice the `componentScaleX` is quite often identical between two (or more) adjacent image components.
All-in-all it's thus not necessary to *unconditionally* re-compute the `xScaleBlockOffset` when getting the JPEG image data.

While avoiding, in many cases, one or more loops can never be a bad thing these changes are unfortunately completely dominated by the rest of the JpegImage code and consequently doesn't really show up in benchmark results. *Hence I'd understand if this patch is ultimately deemed not necessary.*

### DIFF
--- a/src/core/jpg.js
+++ b/src/core/jpg.js
@@ -1105,6 +1105,7 @@ var JpegImage = (function JpegImageClosure() {
       var data = new Uint8ClampedArray(dataLength);
       var xScaleBlockOffset = new Uint32Array(width);
       var mask3LSB = 0xfffffff8; // used to clear the 3 LSBs
+      let lastComponentScaleX;
 
       for (i = 0; i < numComponents; i++) {
         component = this.components[i];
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
This, and the other recent patches for `src/core/jpg.js`, is an attempt at finding any simple optimization opportunities in the code. If we want to look into using `src/core/jpg.js` exclusively for all JPEG images, it seems to me that trying to squeeze every bit of "free" performance out of the code cannot hurt :-)

timvandermeij (approved):

timvandermeij:
Let's do this. Thanks!

PR #11558, opened 2020-02-02
Remove the `eslint-disable no-var` rule from the `web/grab_to_pan.js` file

Snuffleupagus:
Since this rule is now enforced in the entire `/web` folder, there's no good reason for it to remain disabled in this file. Most likely, it's added to reduce the number of linting errors when we started enforcing block-scoped variables.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #11559, opened 2020-02-02
Fix how curveTo2 (v operator) is translated to SVG

bhasto:
Based on [PDF spec](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf), page 133, with `v` operator, current point should be used as the first control point of the curve.

Do not overwrite current point before an SVG curve is built, so it can be actually used as first control point.

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -1231,8 +1231,6 @@ if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) {
             j += 6;
             break;
           case OPS.curveTo2:
-            x = args[j + 2];
-            y = args[j + 3];
             d.push(
               "C",
               pf(x),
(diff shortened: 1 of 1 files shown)


timvandermeij:
Interesting. Do you have an actual PDF file that is broken because of this? I think the fix is fine because we do that already in the canvas back-end, see https://github.com/mozilla/pdf.js/blob/794744c3fad3411a886b697cd3c4c6cb96725880/src/display/canvas.js#L1229-L1241.

bhasto:
Unfortunately I don’t have an image readily available that I could share in a public forum.

I can try to come up with one, but I don’t have experience creating PDFs, so it’ll take some time.

timvandermeij (approved):

timvandermeij:
Thank you!

PR #11566, opened 2020-02-04
Enable the ESLint `no-shadow` rule

Snuffleupagus:
This rule is *not* currently enabled in mozilla-central, but it appears commented out[1] in the ESLint definition file; see https://searchfox.org/mozilla-central/rev/2e355fa82aaa87e8424a9927c8136be184eeb6c7/tools/lint/eslint/eslint-plugin-mozilla/lib/configs/recommended.js#238-239

Unfortunately this rule is, for fairly obvious reasons, impossible to `--fix` automatically (even partially) and each case thus require careful manual tweaks.
Hence this ESLint rule is, by some margin, probably the most difficult one that we've enabled thus far. However, using this rule does seem like a good idea in general since allowing variable shadowing could lead to subtle (and difficult to find) bugs or at the very least confusing code.

*Please note:* Possibly these changes ought to have been split into *multiple* commits, e.g. one per directory, so if this is considered "un-reviewable" as-is I can try to do that.

Please find additional details about the ESLint rule at https://eslint.org/docs/rules/no-shadow

---
[1] This suggests, at least to me, a desire to possibly enable it at some future point. Most likely, a *very* large number of failures currently prevent doing so.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -119,8 +119,8 @@
     "no-catch-shadow": "error",
     "no-delete-var": "error",
     "no-label-var": "error",
+    "no-shadow": "error",
     "no-shadow-restricted-names": "error",
-    "no-shadow": "off",
     "no-undef-init": "error",
     "no-undef": ["error", { "typeof": true, }],
     "no-unused-vars": ["error", {
--- a/external/builder/preprocessor2.js
+++ b/external/builder/preprocessor2.js
@@ -289,7 +289,7 @@ function traverseTree(ctx, node) {
   for (var i in node) {
     var child = node[i];
     if (typeof child === "object" && child !== null && child.type) {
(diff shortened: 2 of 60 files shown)


timvandermeij:
@brendandahl Could you kick the Windows bot again? It seems to stop relatively often lately due to not being able to remove a file.

timvandermeij:
In order to get this merged, I'd like to say that the smaller chunks really helped for me to review it. I'm hoping this can be split up even more so we can enable this rule, because shadowing can be an annoying source of bugs.

Snuffleupagus:
> I'd like to say that the smaller chunks really helped for me to review it. I'm hoping this can be split up even more so we can enable this rule,

My intention is to split this into one PR per directory, but I've not yet had time to do that; hopefully soon though :-)

PR #11567, opened 2020-02-04
Add support for CSS variables using the `PostCSS CSS Variables` package (issue 11462)

### LINKED ISSUE #11462: Use CSS variables in `viewer.css`
In order to make re-skinning the viewer easier, we should look into starting to use CSS variables in `viewer.css`. Not only does this reduce duplication, it also allows us and other users to more easily change the appearance of the viewer.

There is the option of using something like https://github.com/jhildenbiddle/css-vars-ponyfill in `GENERIC` builds to rewrite CSS variables at runtime so we don't need to preprocess the CSS. If that doesn't work, using `PostCSS` is a possibility since we already use it. Either way, we need to make sure that CSS variables also work for older browsers that we still support.

Snuffleupagus:
Having thought *briefly* about using `css-vars-ponyfill`, I'm no longer convinced that it'd be a good idea. The reason is that if we actually want to properly support CSS variables, then that functionality should be available in *all* of our CSS files.
Note in particular the `pdf_viewer.css` file that's built as part of the `COMPONENTS` target, in which case I really cannot see how a rewrite-at-the-client solution would ever be guaranteed to always work correctly and without accidentally touching other CSS in the surrounding application.

All-in-all, simply re-writing the CSS variables at build-time seems much easier and is thus the approach taken in this patch; courtesy of https://github.com/MadLittleMods/postcss-css-variables
By using its `preserve` option, the built files will thus include *both* a fallback and a modern `var(...)` format[1]. As a proof-of-concept this patch removes a couple of manually added fallback values, and converts an additional sidebar related property to use a CSS variable.

Fixes #11462

---
[1] Comparing the `master` branch with this patch, when using `gulp generic`, produces the following diff for the built `web/viewer.css` file:
```diff
@@ -408,6 +408,7 @@

 :root {
   --sidebar-width: 200px;
+  --sidebar-transition-duration: 200ms;
 }

 * {
@@ -550,27 +551,28 @@
   position: absolute;
   top: 32px;
   bottom: 0;
-  width: 200px; /* Here, and elsewhere below, keep the constant value for compatibility
-                   with older browsers that lack support for CSS variables. */
+  width: 200px;
   width: var(--sidebar-width);
   visibility: hidden;
   z-index: 100;
   border-top: 1px solid rgba(51, 51, 51, 1);
   -webkit-transition-duration: 200ms;
           transition-duration: 200ms;
+  -webkit-transition-duration: var(--sidebar-transition-duration);
+          transition-duration: var(--sidebar-transition-duration);
   -webkit-transition-timing-function: ease;
           transition-timing-function: ease;
 }
 html[dir='ltr'] #sidebarContainer {
   -webkit-transition-property: left;
   transition-property: left;
-  left: -200px;
+  left: calc(-1 * 200px);
   left: calc(-1 * var(--sidebar-width));
 }
 html[dir='rtl'] #sidebarContainer {
   -webkit-transition-property: right;
   transition-property: right;
-  right: -200px;
+  right: calc(-1 * 200px);
   right: calc(-1 * var(--sidebar-width));
 }

@@ -640,6 +642,8 @@
 #viewerContainer:not(.pdfPresentationMode) {
   -webkit-transition-duration: 200ms;
           transition-duration: 200ms;
+  -webkit-transition-duration: var(--sidebar-transition-duration);
+          transition-duration: var(--sidebar-transition-duration);
   -webkit-transition-timing-function: ease;
           transition-timing-function: ease;
 }
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -19,6 +19,7 @@
 "use strict";
 
 var autoprefixer = require("autoprefixer");
+var calc = require("postcss-calc");
 var cssvariables = require("postcss-css-variables");
 var fancylog = require("fancy-log");
 var fs = require("fs");
@@ -772,6 +773,7 @@ gulp.task(
         .pipe(
           postcss([
             cssvariables(CSS_VARIABLES_CONFIG),
+            calc(),
             autoprefixer(AUTOPREFIXER_CONFIG),
           ])
         )
@@ -807,6 +809,7 @@ gulp.task(
         .pipe(
           postcss([
             cssvariables(CSS_VARIABLES_CONFIG),
+            calc(),
             autoprefixer(AUTOPREFIXER_CONFIG),
           ])
         )
@@ -868,6 +871,7 @@ gulp.task(
         .pipe(
           postcss([
             cssvariables(CSS_VARIABLES_CONFIG),
+            calc(),
             autoprefixer(AUTOPREFIXER_CONFIG),
           ])
         )
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -19,6 +19,7 @@
 "use strict";
 
 var autoprefixer = require("autoprefixer");
+var cssvariables = require("postcss-css-variables");
 var fancylog = require("fancy-log");
 var fs = require("fs");
 var gulp = require("gulp");
@@ -81,6 +82,9 @@ var AUTOPREFIXER_CONFIG = {
     "> 0.5%",
     "not dead",
   ],
+};
+var CSS_VARIABLES_CONFIG = {
+  preserve: true,
 };
 
 var DEFINES = {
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
Sign, it seems that IE11 cannot handle `calc(-1 * 200px);` correctly or specifically the `-1` part here...

We could possibly use [another PostCSS package](https://github.com/postcss/postcss-calc) to solve that, but then again maybe not since I just found https://github.com/postcss/postcss-calc/issues/87 :-(

Snuffleupagus:
This seems to work now, even in IE11, without anything being obviously broken :-)

Once this's landed we should thus be free to use CSS variables to simplify/re-factor the existing rules (and obviously when adding new rules), however I don't intend to convert more rules *in this PR* since the changes are mostly for proof-of-concept reasons and to allow validation of the overall approach.

timvandermeij (approved):

timvandermeij:
Looks good! I'm glad that we can start using CSS variables properly now.

PR #11568, opened 2020-02-05
Ensure that the PDF header contains an actual number (PR 11463 follow-up)

Snuffleupagus:
While it would be nice to change the `PDFFormatVersion` property, as returned through `PDFDocumentProxy.getMetadata`, to a number (rather than a string) that would unfortunately be a breaking API change.
However, it does seem like a good idea to at least *validate* the PDF header version on the worker-thread, rather than potentially returning an arbitrary string.


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #11569, opened 2020-02-05
Replace most remaining `Element.setAttribute("style", ...)` usage with `Element.style = ...` instead

### LINKED ISSUE #11576: CSP unsafe-inline CSS violations
Separating this from issue #11036 .

I am getting unsafe-inline errors for the CSS, for example on this line of code:
https://github.com/mozilla/pdf.js/blob/master/web/ui_utils.js#L893

Looks like there are several places using code like that, which would need to be changed to fix these errors.

`Refused to apply inline style because it violates the following Content Security Policy directive: "style-src ...". Either the 'unsafe-inline' keyword, a hash ('sha256-MW4Iy/yu3WipUpTMM/T6OjvUu9R6vUwutodlmYNo6qM='), or a nonce ('nonce-...') is required to enable inline execution.`

Snuffleupagus:
This should hopefully be useful in environments where restrictive CSPs are in effect.
In most cases the replacement is entirely straighforward, and there's only a couple of special cases:
 - For the `src/display/font_loader.js` and `web/pdf_outline_viewer.js `cases, since the elements aren't appended to the document yet, it shouldn't matter if the style properties are set one-by-one rather than all at once.
 - For the `web/debugger.js` case, there's really no need to set the `padding` inline at all and the definition was simply moved to `web/viewer.css` instead.

*Please note:* There's still *a single* case left, in `web/toolbar.js` for setting the width of the zoom dropdown, which is left intact for now.
The reasons are that this particular case shouldn't matter for users of the general PDF.js library, and that it'd make a lot more sense to just try and re-factor that very old code anyway (thus fixing the `setAttribute` usage in the process).

Fixes #11576

### DIFF
--- a/src/display/font_loader.js
+++ b/src/display/font_loader.js
@@ -341,12 +341,11 @@ if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")) {
       names.push(loadTestFontId);
 
       const div = document.createElement("div");
-      div.setAttribute(
-        "style",
-        "visibility: hidden;" +
-          "width: 10px; height: 10px;" +
-          "position: absolute; top: 0px; left: 0px;"
-      );
+      div.style.visibility = "hidden";
+      div.style.width = div.style.height = "10px";
+      div.style.position = "absolute";
+      div.style.top = div.style.left = "0px";
+
       for (i = 0, ii = names.length; i < ii; ++i) {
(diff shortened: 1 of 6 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this! It's also more readable in my opinion.

PR #11570, opened 2020-02-06
Re-factor `Toolbar._adjustScaleWidth` to improve/simplify how the zoom dropdown width is calculated

### LINKED ISSUE #11576: CSP unsafe-inline CSS violations

Snuffleupagus:
This patch contains some *much* needed clean-up of, and improvements to, this old code thus addressing a number of issues:

 - Set more reasonable *default* values for the widths, in `web/viewer.css`, since the current ones are actually too small even for the (default) `en-US` locale.
   This obviously result in a slightly larger zoom dropdown width for many locales, but the more consistent UI does look good to me.

 - Stop setting the `min-width`/`max-width` and just use `width` instead.

 - Set an explicit `height` of the zoom dropdown, in an attempt to get Google Chrome to display it with the same height as the toolbar buttons.

 - Remove additional `Element.setAttribute("style", ...)` usage.

 - Actually check *all* of the predefined l10n strings, since the old implementation (implicitly) assumed that the currently selected one was the longest (note e.g. the `ja-JP` locale where one string is considerably longer than the rest).

 - Stop invalidating the DOM multiple times when doing the measurements. This was achieved by using a temporary in-memory `canvas`, and we now only need to query the DOM once in order to get the current font properties of the zoom dropdown.

Fixes #11576

### DIFF
--- a/web/toolbar.js
+++ b/web/toolbar.js
@@ -24,8 +24,9 @@ import {
 } from "./ui_utils.js";
 
 const PAGE_NUMBER_LOADING_INDICATOR = "visiblePageIsLoading";
-const SCALE_SELECT_CONTAINER_PADDING = 8;
-const SCALE_SELECT_PADDING = 22;
+// Keep the two values below up-to-date with the values in `web/viewer.css`:
+const SCALE_SELECT_CONTAINER_WIDTH = 140; // px
+const SCALE_SELECT_WIDTH = 162; // px
 
 /**
  * @typedef {Object} ToolbarOptions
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #11577, opened 2020-02-08
Prevent circular references in the /Pages tree

Snuffleupagus:


Rob--W:
If I'm not mistaken, this looks like something that could reasonably easily be covered by a unit test. Could you update the PR with a unit test?

Snuffleupagus:
> Could you update the PR with a unit test?

While I already have a test-case locally, please note that I'd prefer to wait for @brendandahl to sign-off on that though.

web/pdf_rendering_queue.js
```diff
@@ -164,9 +164,14 @@ class PDFRenderingQueue {
...
-        view.draw().finally(() => {
-          this.renderHighestPriority();
-        });
+        view
+          .draw()
+          .finally(() => {
+            this.renderHighestPriority();
+          })
+          .catch(() => {
+            /* Avoid "Uncaught promise" messages in the console. */
```
brendandahl: Can we either log the error or remove the catch? Empty catch statements can make debugging very painful.

brendandahl (changes requested):
Almost there, I'd also like to see a test case and the one comment addressed.

web/pdf_rendering_queue.js
Snuffleupagus: Fair points, any errors here are now logged (such that there's no "Uncaught promise" warnings) and a unit-test were added.

brendandahl (approved):

timvandermeij (approved):

timvandermeij:
Thank you!

PR #11581, opened 2020-02-10
Enable the ESLint `no-var` rule in the `src/` folder

Snuffleupagus:
Previously we've been selectively enabling this rule on a file-by-file basis (and the `web/` folder is already done), but it seems like a good idea to enable this in the entire `src/` folder now.
Most likely we don't want to enable this rule in *all* files/folders, e.g. the `examples/` folder should probably be left alone for now, hence why this rule isn't enabled globally.

With the recent introduction of Prettier this sort of mass enabling of ESLint rules also becomes a lot easier, since the code will be automatically reformatted as necessary to account for e.g. changed line lengths.

*Please note:* This patch is generated automatically, by appending the `--fix` argument to the ESLint call used in the `gulp lint` task. It will thus require some additional manual fixes, which will be done in a *separate* commit.

### DIFF
--- a/src/.eslintrc
+++ b/src/.eslintrc
@@ -1,0 +1,10 @@
+{
+  "extends": [
+    "../.eslintrc"
+  ],
+
+  "rules": {
+    // ECMAScript 6
+    "no-var": "error",
+  },
+}
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -12,7 +12,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint no-var: error */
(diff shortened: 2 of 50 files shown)


Snuffleupagus:
The manual changes required here will take some time/care to get right, since the ones involved with `switch/case` cannot just be converted as-is; see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#Block-scope_variables_within_switch_statements

I'll wait for PR #11566 to land *first*, before continuing to work on this, since there'd probably be a fair bit of merge conflicts to deal with otherwise...

PR #11586, opened 2020-02-11
[TextLayer] Immediately set the padding, rather than checking if it's empty, in `expandTextDivs`

Snuffleupagus:
In practice it's extremely rare[1] for the padding to be zero in *all* components, hence it seems better to just set it directly rather than creating a temporary variable and checking for the "no padding"-case.

---
[1] In the `tracemonkey.pdf` file that only happens with `0.08%` of all text elements.

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -658,7 +658,6 @@ var renderTextLayer = (function renderTextLayerClosure() {
         expand(this);
         this._bounds = null;
       }
-      const NO_PADDING = "0 0 0 0";
       const transformBuf = [],
         paddingBuf = [];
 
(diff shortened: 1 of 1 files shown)


brendandahl:
Was there a specific PDF that this was added for?

Snuffleupagus:
> Was there a specific PDF that this was added for?

Don't think so, it's probably just a case of a premature optimization.
(Also, this is only relevant for the "enhanced" text-selection mode which is still disabled by default.)

brendandahl (approved):
r+ if the windows bot runs okay

timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #11590, opened 2020-02-12
Update links from IRC to Riot.

brendandahl:
Mozilla's IRC is going away and we're migrating to Matrix/Riot.


Snuffleupagus (approved):
Please update https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md as well.

Also, please update the following Wiki pages:
https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions
https://github.com/mozilla/pdf.js/wiki/Contributing

Snuffleupagus:
In addition to https://github.com/mozilla/pdf.js/pull/11590#pullrequestreview-357813308, one question: Should this perhaps be called "Matrix" in the docs (rather than "Riot"), since the former is the protocol and the latter simply a client (as far as I understand, based on reading https://wiki.mozilla.org/Matrix)?

(If I'd still have push access to this repo, I'd happily fix-up the PR such that it could land.)

brendandahl:
Yeah, matrix is more appropriate. I'll update.

Snuffleupagus (approved):

PR #11593, opened 2020-02-13
Remove `LoopbackPort.postMessage` special-case for polyfilled `TypedArray`s

Snuffleupagus:
Given that all `TypedArray` polyfills were removed in PDF.js version `2.0`, since native support is now required, this branch has been dead code for awhile.


Snuffleupagus:
The `Loopback` port is only used with fake workers, hence the relevant block of code will only run in Node.js/Travis unit-tests. However, I suppose that it cannot hurt to run the *standard browser* unit-tests just in case.

/botio unittest

timvandermeij (approved):

timvandermeij:
Good catch!

PR #11596, opened 2020-02-13
Re-factor how `Metadata` class instances store its data internally

Snuffleupagus:
Please note that these changes do *not* affect the *public* interface of the `Metadata` class, but only touches internal structures.[1]

These changes were prompted by looking at the `getAll` method, which simply returns the "private" metadata object to the consumer. This seems wrong conceptually, since it allows way too easy/accidental changes to the internal parsed metadata.
As part of fixing this, the internal metadata was changed to use a `Map` rather than a plain Object.

---
[1] Basically, we shouldn't need to worry about someone depending on internal implementation details.

### DIFF
--- a/src/display/metadata.js
+++ b/src/display/metadata.js
@@ -37,7 +37,7 @@ class Metadata {
   _repair(data) {
     // Start by removing any "junk" before the first tag (see issue 10395).
     return data
-      .replace(/^([^<]+)/, "")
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice improvements; thank you!

PR #11598, opened 2020-02-14
Add polyfills to support iteration of `Map` and `Set`

Snuffleupagus:
Without this, things such as e.g. `Metadata.getAll` is broken in IE11 (see PR #11596).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Browser_compatibility

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#Browser_compatibility


timvandermeij (approved):

timvandermeij:
Thanks for the follow-up!

PR #11601, opened 2020-02-15
[api-minor] Decode all JPEG images with the built-in PDF.js decoder in `src/core/jpg.js`

### LINKED ISSUE #7041: Poorly rendered image quality

Snuffleupagus:

Fixes https://github.com/mozilla/pdf.js/pull/6984#issuecomment-184778220
Fixes #7041


Snuffleupagus:
Sign; oh great, since Firefox was update to the Nightly branch (during the switch to Puppeteer) the Windows bot is now seems to be affected by https://bugzilla.mozilla.org/show_bug.cgi?id=1607630

Snuffleupagus:
Sigh; oh great, since Firefox was update to the Nightly branch (during the switch to Puppeteer) the Windows bot is now seems to be affected by https://bugzilla.mozilla.org/show_bug.cgi?id=1607630

Snuffleupagus:
Sigh; oh great, since Firefox was update to the Nightly branch (during the switch to Puppeteer) the Windows bot now seem to be affected by https://bugzilla.mozilla.org/show_bug.cgi?id=1607630

timvandermeij:
Oof, that's a nasty bug. I can confirm I have not seen such corruption on Linux so far. On the one hand I'm happy we're finding it now so we can bump the upstream issue so it hopefully won't reach the stable release, but on the other hand it does mess with our tests now. We should see how it goes; otherwise there are ways to tell Puppeteer to use another version than the built-in versions.

Snuffleupagus:
> [,..] otherwise there are ways to tell Puppeteer to use another version than the built-in versions.

If all the intermittent font rendering failures, which are not related to this patch, is blocking reviewing/landing of this patch then going back to Firefox 76 for the time being would be nice!

timvandermeij:
The bug is fixed upstream, but I'm not sure if it made it to today's Nightly yet. Let's see...

/botio test

timvandermeij (approved):

timvandermeij:
Thank you! This greatly simplifies the code and should improve performance.

/botio makeref

PR #11602, opened 2020-02-16
Move validation of `chromium/preferences_schema.json` to its own gulp task

Snuffleupagus:
With the way that the `default_preferences.json` file is now generated at build time, the `gulp lint` task is now noticeably slower than before. This slowdown has been, and still is, somewhat annoying during the deployment of new ESLint rules.

Hence this patch, which moves the `chromium/preferences_schema.json` validation from `gulp lint` and into a new `gulp lint-chromium` task instead. *Obviously* this new task is run as part of the `gulp npm-test` task, and thus through `npm test` on Node.js/Travis, such that it's still being tested as before.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1389,43 +1389,45 @@ gulp.task(
   })
 );
 
+gulp.task("lint", function(done) {
+  console.log();
+  console.log("### Linting JS files");
+
+  // Ensure that we lint the Firefox specific *.jsm files too.
+  var options = [
+    "node_modules/eslint/bin/eslint",
+    "--ext",
+    ".js,.jsm",
+    ".",
+    "--report-unused-disable-directives",
+  ];
+  var esLintProcess = startNode(options, { stdio: "inherit" });
+  esLintProcess.on("close", function(code) {
+    if (code !== 0) {
+      done(new Error("ESLint failed."));
+      return;
+    }
+    console.log("files checked, no errors found");
+    done();
+  });
+});
+
 gulp.task(
-  "lint",
+  "lint-chromium",
   gulp.series("default_preferences", function(done) {
     console.log();
-    console.log("### Linting JS files");
+    console.log("### Checking supplemental Chromium files");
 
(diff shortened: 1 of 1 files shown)


timvandermeij:
Hm, it looks like "files checked, no errors found" is no longer printed and that's what https://github.com/mozilla/botio-files-pdfjs/blob/master/on_cmd_lint.js checks for. Either that can be fixed, or we can change the bot command to look at the exit code instead of string matching.

Snuffleupagus:
> Hm, it looks like "files checked, no errors found" is no longer printed and that's what https://github.com/mozilla/botio-files-pdfjs/blob/master/on_cmd_lint.js checks for.

Sorry, I didn't think to look for that!
While I do think that that's a somewhat strange way for the bots to handle this, it seems easier to just reinstate the log line (also since *all* botio scripts seem to use that pattern).

gulpfile.js
```diff
@@ -1389,43 +1389,45 @@ gulp.task(
...
+    ".",
+    "--report-unused-disable-directives",
+  ];
+  var esLintProcess = startNode(options, { stdio: "inherit" });
+  esLintProcess.on("close", function(code) {
+    if (code !== 0) {
+      done(new Error("ESLint failed."));
+      return;
+    }
+    console.log("files checked, no errors found");
```
timvandermeij: I'm only wondering now what will happen if the `lint` task prints this, but the `lint-chromium` task fails. Since this is already printed, won't be bots just see the success and miss the failed other target?

Snuffleupagus: Actually the bots won't run `gulp lint-chromium` with this patch, but I'm not sure how much that matters given that it *is* run on Travis and `botio lint` isn't too commonly used these days.

timvandermeij: Yes, you're right. No worries then.

timvandermeij (approved):

timvandermeij:
Nice that we can run linting faster now. Thanks!

PR #11603, opened 2020-02-16
Enable the `no-buffer-constructor` ESLint rule

Snuffleupagus:
According to https://nodejs.org/api/buffer.html#buffer_class_buffer: `new Buffer(...)` is deprecated in up-to-date versions of Node.js, hence you want to prevent it from being accidentally used.

Please see https://eslint.org/docs/rules/no-buffer-constructor for additional information.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -127,6 +127,9 @@
       "args": "none",
     }],
 
+    // Node.js and CommonJS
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

PR #11604, opened 2020-02-16
Enable the `unicorn/prefer-starts-ends-with` ESLint plugin rule

Snuffleupagus:
This complements the existing `mozilla/use-includes-instead-of-indexOf` plugin rule, by also disallowing unnecessary regular expressions when comparing strings.

Please see https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-starts-ends-with.md for additional information.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #11605, opened 2020-02-16
Update `gulp lint` to support passing of the `--fix` argument on the command line

Snuffleupagus:
*I've had this patch locally for awhile, but have apparently missed to upstream it.*

This simplifies enabling of new ESLint rules, since most of them support automatic fixing of errors, without having to edit `gulpfile.js` or manually invoke ESLint directly.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #11607, opened 2020-02-17
Add a `PDFViewerApplication.initializedPromise` property to allow (easier) tracking of when the default viewer has been initialized

Snuffleupagus:
This complements the existing `PDFViewerApplication.initialized` boolean property, and may be helpful for custom implementations of the default viewer. This will thus provide users of the default viewer an alternative to setting the preference to dispatch events to the DOM (and listen for the "localized" event), since they can instead use:
```javascript
    // The viewer has now been initialized.
  })
});
```

Note that in order to avoid manually tracking the initialization state *twice*, this implementation purposely uses the `PromiseCapability` functionality to handle both `PDFViewerApplication.initialized` and `PDFViewerApplication.initializedPromise` internally.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -38,6 +38,7 @@ import {
 import { AppOptions, OptionKind } from "./app_options.js";
 import {
   build,
+  createPromiseCapability,
   getDocument,
   getFilenameFromUrl,
   GlobalWorkerOptions,
@@ -128,7 +129,7 @@ class DefaultExternalServices {
 
 const PDFViewerApplication = {
   initialBookmark: document.location.hash.substring(1),
-  initialized: false,
+  _initializedCapability: createPromiseCapability(),
   fellback: false,
   appConfig: null,
   pdfDocument: null,
@@ -215,7 +216,7 @@ const PDFViewerApplication = {
       this.eventBus.dispatch("localized", { source: this });
     });
 
-    this.initialized = true;
+    this._initializedCapability.resolve();
   },
 
   /**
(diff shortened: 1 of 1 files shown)


timvandermeij:
Perhaps a bit off-topic for this particular patch, but I'm wondering how we should document the availability of this (and other) features for third-party users. There are quite a few tips and tricks third-party implementers can use, but we don't really document them nor do we have examples for all of them. This is one of those things that can be really useful, but there should be a way to discover that it exists.

Do you have ideas for this? I'm thinking a wiki page could suffice and we can update it incrementally.

Snuffleupagus:
> [...] but I'm wondering how we should document the availability of this (and other) features for third-party users. There are quite a few tips and tricks third-party implementers can use, but we don't really document them nor do we have examples for all of them.

That seems like a fair point!

As long as any documentation *prominently* features the standard "please don't use the viewer as-is"-disclaimer that might be helpful, then I suppose that more documentation cannot hurt.

Furthermore, you'd probably also need to make it very clear that while the *general* PDF.js library is versioned and shouldn't simply break in backwards incompatible ways (the latter probably applies to some extent to the "viewer components" too), the *default viewer* itself shouldn't be considered stable and things may change/break as needed for e.g. Firefox's needs. 

> I'm thinking a wiki page could suffice and we can update it incrementally.

That sounds like the easiest solution, and shouldn't be too much effort to keep up-to-date.

timvandermeij (approved):

timvandermeij:
Nice improvement! I have made https://github.com/mozilla/pdf.js/wiki/Third-party-viewer-usage to track tips and tricks for custom viewer deployments. If you have additions, feel free to let me know and we can update the page.

PR #11608, opened 2020-02-17
Add a `ignoreDestinationZoom` option/preference to allow users to preserve the current zoom level when navigating to internal destinations (issue 5064, 11606)

### LINKED ISSUE #5064: Increased zoom when clicking internal URL
In the document at https://pdf.yt/d/1dKWAxs03AvnYqkt, when clicking a reference link (eg. "Lau et al" on page 17), the viewer not only jumps to the footer, but also sets it to 307.13% zoom. I can't quite understand the URL format of those reference links, but perhaps the nullbytes(?) in there are messing something up?

---

_PDF 70ce9de8858387c1ada82e28f458dd7 [1.7 Acrobat Distiller 8.1.0 (Windows) / Elsevier](PDF.js: 1.0.21), Google Chrome 35.0.1916.153, openSUSE 13.1 64-bits_

### LINKED ISSUE #11606: Add option to allow users to override the PDF's zoom/scaling mode
Attach (recommended) or Link to PDF file here: https://www.ssa.gov/pubs/EN-05-10004.pdf
(simplified example, the user who reported the problem experienced this issue with a PDF file of 500+ pages)

1. Open the above PDF file.
2. Show the document outline (via the sidebar)
3. Click on "Contacting Social Security" and observe that the zoom status changed from Automatic to "Page Fit"
4. Change Zoom to something else, e.g. "Page Width"
5. Click on "Contacting Social Security" in the document outline again.

After step 5, the zoom level should have stayed at "Page Width".

What went wrong?
After step 5, the zoom level changed to "Page Fit", because the PDF document defines a `/GoTo` to a named destination with `/Fit`. Consequently, when PDF.js navigates to this, the zoom level is changed. To recover from this, the user is forced to click on the Zoom dropdown menu and change the zoom level again.

To improve the user experience of this situation, we should offer a way to allow the user to ignore the zoom. At the very least, we could add a preference to allow users to disable this behavior.
We could also consider adding some form of auto detection: If a user reverts the zoom a few times in a row, then we would ignore the forced requests to change the zoom level.

In terms of implementation, the check could be added before
https://github.com/mozilla/pdf.js/blob/ba1d8d8728050d7a83a4a0255e030b19e55ffb22/web/base_viewer.js#L837-L838
or
https://github.com/mozilla/pdf.js/blob/93aa613db7a874e6ed964f394241605d5586c142/web/pdf_link_service.js#L158

Snuffleupagus:
Fixes #5064
Fixes #11606

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -159,6 +159,12 @@
       "type": "boolean",
       "default": false
     },
+    "ignoreDestinationZoom": {
+      "title": "Ignore the zoom argument in destinations",
+      "description": "When enabled it will maintain the currently active zoom level, rather than letting the PDF document modify it, when navigating to internal destinations.",
+      "type": "boolean",
+      "default": false
+    },
     "enablePrintAutoRotate": {
       "title": "Automatically rotate printed pages",
       "description": "When enabled, pages whose orientation differ from the first page are rotated when printed.",
--- a/web/app.js
+++ b/web/app.js
@@ -354,6 +354,7 @@ const PDFViewerApplication = {
       eventBus,
       externalLinkTarget: AppOptions.get("externalLinkTarget"),
       externalLinkRel: AppOptions.get("externalLinkRel"),
+      ignoreDestinationZoom: AppOptions.get("ignoreDestinationZoom"),
     });
     this.pdfLinkService = pdfLinkService;
 
--- a/web/app_options.js
+++ b/web/app_options.js
@@ -86,6 +86,11 @@ const defaultOptions = {
     value: false,
     kind: OptionKind.VIEWER + OptionKind.PREFERENCE,
   },
+  ignoreDestinationZoom: {
+    /** @type {boolean} */
+    value: false,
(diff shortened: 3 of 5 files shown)


extensions/chromium/preferences_schema.json
```diff
@@ -159,6 +159,10 @@
       "type": "boolean",
       "default": false
     },
+    "ignoreDestinationZoom": {
```
Snuffleupagus: Should this be exposed in the UI (in that case there should probably be a title/description added)?

timvandermeij: We have seen quite a few questions about this particular behavior, and personally I also always found it more confusing than helpful, so I think it's good to expose this in the UI with a description saying that the option overrides the behavior that PDF files can change the zoom level.

Snuffleupagus: I'm always finding these sort of descriptions/explanations somewhat difficult to write (in such a way that they really benefit users), but at least the patch is now updated to address this point (and it's always possible to improve the wording later on).

timvandermeij (approved):

timvandermeij:
Looks good!

PR #11616, opened 2020-02-18
Update the PDF.js web page to link to both versions of the demo viewer (PR 11241 follow-up)

Snuffleupagus:
Rather than adding two buttons, it seems easier to simply link to the relevant section of the README instead (since it also means fewer things to keep up-to-date).


timvandermeij (approved):

timvandermeij:
Thank you for the follow-up!

PR #11620, opened 2020-02-21
Remove the unused `thisArg` from `RefSetCache.forEach`

Snuffleupagus:
Given that this is completely unused, and that a "normal" function call may be a *tiny* bit more efficient, there's no good reason as far as I can tell to keep it.


timvandermeij (approved):

timvandermeij:
That's also a bit more readable. Thanks!

PR #11623, opened 2020-02-22
Use the ESLint `no-restricted-syntax` rule to prevent direct usage of `new Cmd()`/`new Name()`/`new Ref()`

Snuffleupagus:
Given that all of these primitives implement caching, to avoid unnecessarily duplicating those objects *a lot* during parsing, it would thus be good to actually enforce usage of `Cmd.get()`/`Name.get()`/`Ref.get()` in the code-base.
Luckily it turns out that there's an ESLint rule, which is fairly easy to use, that can be used to disallow arbitrary JavaScript syntax.

Please find additional details about the ESLint rule at https://eslint.org/docs/rules/no-restricted-syntax

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -143,6 +143,20 @@
     "no-multiple-empty-lines": ["error", { "max": 1, "maxEOF": 0, "maxBOF": 1, }],
     "no-nested-ternary": "error",
     "no-new-object": "error",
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Sounds like a good thing to do. Thanks!

PR #11625, opened 2020-02-24
Use the same non-embedded Wingdings fallback for fonts named "Wingdings-Regular" too (PR 5463 follow-up, issue 11451)

### LINKED ISSUE #11451: wrong character rendered
This file has a character (symbol) that pdf.js renders wrongly, while most pdf readers get it correctly:
[error2.pdf](https://github.com/mozilla/pdf.js/files/4003700/error2.pdf)

Configuration:

Web browser and its version: chrome browser latest version
Operating system and its version: windows 10 latest version
PDF.js version: online
Is a browser extension: no
1.  open the above file at https://mozilla.github.io/pdf.js/web/viewer.html

correct character: **ò**
pdf.js renders as: **Ú** 
(these are wingding symbols shown in below screenshot, but the actual character is as copied and pasted into a text editor)

image

image

Snuffleupagus:
This patch extends the existing heuristics, which are really the best that we can do in general for these kinds of non-embedded *and* non-standard fonts.

Furthermore, this patch also tries to improve the copy-and-paste behaviour for non-embedded Wingdings fonts by also using the `ZapfDingbatsEncoding` in this case.

*Note:* I'm not sure that adding additional tests for Wingdings fonts matters that much, given how limited our "support" for them really is.

Fixes #11451

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -2373,7 +2373,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           if (!properties.file) {
             if (/Symbol/i.test(properties.name)) {
               encoding = SymbolSetEncoding;
-            } else if (/Dingbats/i.test(properties.name)) {
+            } else if (/Dingbats|Wingdings/i.test(properties.name)) {
               encoding = ZapfDingbatsEncoding;
             }
           }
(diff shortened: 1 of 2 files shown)


timvandermeij:
@brendandahl Could you kick the Windows bot again? Some PRs are currently blocked on this because we can't make reference images or run the tests on Windows.

timvandermeij (approved):

timvandermeij:
Thank you!

/botio makeref

PR #11631, opened 2020-02-26
[api-minor] Deprecate `getGlobalEventBus` and update the "viewer components" examples accordingly

Snuffleupagus:
- [api-minor] Deprecate `getGlobalEventBus` and update the "viewer components" examples accordingly

   To avoid outright breaking third-party usages of the "viewer components" the `getGlobalEventBus` functionality is left intact, but a deprecation message is printed if the function is invoked.

   The various examples are updated to *explicitly* initialize an `EventBus` instance, and provide that when initializing the relevant viewer components.

 - Re-factor the `EventBus` to allow servicing of "external" event listeners *after* the viewer components have updated

   Since the goal has always been, essentially since the `EventBus` abstraction was added, to remove all dispatching of DOM events[1] from the viewer components this patch tries to address one thing that came up when updating the examples:
   The DOM events are always dispatched last, and it's thus guaranteed that all internal event listeners have been invoked first.
   However, there's no such guarantees with the general `EventBus` functionality and the order in which event listeners are invoked is *not* specified. With the promotion of the `EventBus` in the examples, over DOM events, it seems like a good idea to at least *try* to keep this ordering invariant[2] intact.

   Obviously this won't prevent anyone from manually calling the new *internal* viewer component methods on the `EventBus`, but hopefully that won't be too common since any existing third-party code would obviously use the `on`/`off` methods and that all of the examples shows the *correct* usage (which should be similarily documented on the "Third party viewer usage" Wiki-page).

   ---
   [1] Looking at the various Firefox-tests, I'm not sure that it'll be possible to (easily) re-write all of them to not rely on DOM events (since getting access to `PDFViewerApplication` might be generally difficult/messy depending on scopes).
   In any case, even if technically feasible, it would most likely add *a lot* of complication that may not be desireable in the various Firefox-tests. All-in-all, I'd be fine with keeping the DOM events only for the `MOZCENTRAL` target and gated on `Cu.isInAutomation` (or similar) rather than a preference.

   [2] I wouldn't expect any *real* bugs in a custom implementation, simply based on event ordering, but it nonetheless seem like a good idea if any "external" events are still handled last.


timvandermeij (approved):

timvandermeij:
Nice step towards less DOM event usage. Thanks!

timvandermeij:
I have also updated https://github.com/mozilla/pdf.js/wiki/Third-party-viewer-usage now.

Snuffleupagus:
> I have also updated https://github.com/mozilla/pdf.js/wiki/Third-party-viewer-usage now.

Thanks; I've also got a suggested update to the pre-existing example in that Wiki-page in a commit message in a future patch series (which should obviously no land until *after* the upcoming release): https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:rm-getGlobalEventBus

PR #11644, opened 2020-02-28
[api-minor] Add more general OpenAction support (PR 10334 follow-up, issue 11642)

### LINKED ISSUE #11642: PDF With OpenAction set to Print shows yellow warning about incorrect PDF
[openaction.pdf](https://github.com/mozilla/pdf.js/files/4265422/openaction.pdf)
(This PDF had been borrowed from the Firefox case where OpenAction support had been added originally: https://bugzilla.mozilla.org/show_bug.cgi?id=1001080)

Configuration:
- Web browser and its version: 73.0.1
- Operating system and its version: Win10x64
- PDF.js version: 2.4.254
- Is a browser extension: No

1. Open attached PDF in the Firefox 
2. Observe yellow warning and console message: "Warning: JavaScript is not supported"

What is the expected behavior? There shouldn't be any warnings.

image

Snuffleupagus:
This patch deprecates the existing `getOpenActionDestination` API method, in favor of a better and more general `getOpenAction` method instead. (For now JavaScript actions, related to printing, are still handled as before.)

By clearly separating "regular" Print actions from the JavaScript handling, it's thus possible to get rid of the somewhat annoying and strictly incorrect warning when the viewer loads.

Fixes #11642

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -592,9 +592,12 @@ class Catalog {
     return shadow(this, "viewerPreferences", prefs);
   }
 
-  get openActionDestination() {
+  /**
+   * NOTE: "JavaScript" actions are, for now, handled by `get javaScript` below.
+   */
+  get openAction() {
     const obj = this.catDict.get("OpenAction");
-    let openActionDest = null;
+    let openAction = null;
 
     if (isDict(obj)) {
       // Convert the OpenAction dictionary into a format that works with
@@ -602,16 +605,27 @@ class Catalog {
       const destDict = new Dict(this.xref);
       destDict.set("A", obj);
 
-      const resultObj = { url: null, dest: null };
+      const resultObj = { url: null, dest: null, action: null };
       Catalog.parseDestDictionary({ destDict, resultObj });
 
       if (Array.isArray(resultObj.dest)) {
-        openActionDest = resultObj.dest;
+        if (!openAction) {
+          openAction = Object.create(null);
+        }
+        openAction.dest = resultObj.dest;
+      } else if (resultObj.action) {
+        if (!openAction) {
+          openAction = Object.create(null);
+        }
+        openAction.action = resultObj.action;
       }
     } else if (Array.isArray(obj)) {
-      openActionDest = obj;
+      if (!openAction) {
+        openAction = Object.create(null);
+      }
+      openAction.dest = obj;
     }
-    return shadow(this, "openActionDestination", openActionDest);
+    return shadow(this, "openAction", openAction);
   }
 
   get attachments() {
@@ -668,27 +682,10 @@ class Catalog {
       }
     }
 
-    // Append OpenAction actions to the JavaScript array.
-    const openActionDict = this.catDict.get("OpenAction");
-    if (
-      isDict(openActionDict) &&
-      (isName(openActionDict.get("Type"), "Action") ||
-        !openActionDict.has("Type"))
-    ) {
-      const actionType = openActionDict.get("S");
-      if (isName(actionType, "Named")) {
-        // The named Print action is not a part of the PDF 1.7 specification,
-        // but is supported by many PDF readers/writers (including Adobe's).
-        const action = openActionDict.get("N");
-        if (isName(action, "Print")) {
-          if (!javaScript) {
-            javaScript = [];
-          }
-          javaScript.push("print({});");
-        }
-      } else {
-        appendIfJavaScriptDict(openActionDict);
-      }
+    // Append OpenAction "JavaScript" actions to the JavaScript array.
+    const openAction = this.catDict.get("OpenAction");
+    if (isDict(openAction) && isName(openAction.get("S"), "JavaScript")) {
+      appendIfJavaScriptDict(openAction);
     }
 
     return shadow(this, "javaScript", javaScript);
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -458,8 +458,8 @@ var WorkerMessageHandler = {
       return pdfManager.ensureCatalog("viewerPreferences");
     });
 
-    handler.on("GetOpenActionDestination", function(data) {
-      return pdfManager.ensureCatalog("openActionDestination");
+    handler.on("GetOpenAction", function(data) {
+      return pdfManager.ensureCatalog("openAction");
     });
 
     handler.on("GetAttachments", function wphSetupGetAttachments(data) {
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -668,11 +668,18 @@ class PDFDocumentProxy {
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Array} containing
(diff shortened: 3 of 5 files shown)


Snuffleupagus:
botio-windows fonttest

timvandermeij:
Could you elaborate a bit on how this solves the issue, i.e., why was the fallback bar shown in the first place? AFAICT we handled the print actions before too.

Snuffleupagus:
> Could you elaborate a bit on how this solves the issue, i.e., why was the fallback bar shown in the first place? AFAICT we handled the print actions before too.

In this case the `OpenAction` dictionary contains a simple entry (see below), similar to other more standard Named Actions (as handled e.g. [by `PDFLinkService`](https://github.com/mozilla/pdf.js/blob/af8d0b9597ccd0e020910eafd74dd6ad140db520/web/pdf_link_service.js#L335-L381)), but no *actual* JavaScript.
The old implementation, for history reasons probably, converted that to "fake" JavaScript; see [this code](https://github.com/mozilla/pdf.js/pull/11644/files#diff-3fd0fc66c9db990d996e595419b3f1e1L687).
All-in-all, it seems completely wrong to warn about a non-existent script in that case.

```
OpenAction (dict) [id: 34, gen: 0]
  S = /Named
  Type = /Action
  N = /Print
```

Snuffleupagus:
@timvandermeij Since I personally find that explaining things in a clear and concise way (in writing) can be difficult; did https://github.com/mozilla/pdf.js/pull/11644#issuecomment-596806448 address your question is a satisfactory way? If not, please ask again :-)

timvandermeij:
Yes, I see that the old code did not trigger on the named action itself but instead converted that named action to JS that could be executed by the browser. I'm also not entirely sure why we do that, but it does look better to not generate the "fake" JS.

timvandermeij (approved):

timvandermeij:
Looks really good! I have confirmed that the warning no longer appears, and separating the named action handling from the JS handling for printing also looks cleaner. Thanks!

automatedbugreportingfacility:
@Snuffleupagus Can you check if the branches in `openAction()` are correct? `if (!openAction)` is checked in a number of places, but that is always true AFAICT, so I was wondering if some `else` in the containing branches is wrong here.

Snuffleupagus:
> Can you check if the branches in `openAction()` are correct?

Sorry, but I don't really understand the question; actual links *directly* to the relevant code would definitely help here.

> so I was wondering if some `else` in the containing branches is wrong here.

If you think that the current implementation is incorrect, then references to the PDF specification and/or PDF documents where the code as written isn't working would be most helpful; as-is I'm unfortunately a bit lost here...

automatedbugreportingfacility:
Sorry if I was unclear. At https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L609 `openAction` is initialized to `null`. At https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L621 it is checked whether `openAction` is null (although it can't be any other value). At https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L626 `openAction` is checked again, but the branch at line 625 is an `else` branch, so `openAction` couldn't be initialized at line 622, so `if (!openAction)` always evaluates to true again. I don't know if the current implementation is incorrect (i.e. if `openAction` can specify both `dest` and `action` at the same time), but I noticed the tautologies, so I thought I'd bring it to your attention.

Snuffleupagus:
> I don't know if the current implementation is incorrect (i.e. if openAction can specify both dest and action at the same time)

The `OpenAction` entry in the catalog dictionary is specified here (with additional references to the format used): https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#page=82

Ignoring the specification for a moment; basically I don't think the situation you describe will ever happen in practice, at least I've never seen it, and anyway `Catalog.parseDestDictionary` will only extract *one* entry from the dictionary.

> [...] so `if (!openAction)` always evaluates to true again.

We can't very well initialize the Object unconditionally, since we want the default case (i.e. no `OpenAction` dictionary present) to be easy to detect, hence the reason for initializing the Object only where necessary.

> At https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L621 it is checked whether `openAction` is null (although it can't be any other value). At https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L626 `openAction` is checked again,

Well, generally speaking it was much easier to simply treat each individual case *separately* such that e.g. any future changes to this code won't end up accidentally overwriting data. 

---

Given that the code, as written, works fine as far as I'm reading the specification it should hopefully not be necessary to make changes here (unless there's PDF examples to the contrary of course).

Snuffleupagus:
> I don't know if the current implementation is incorrect (i.e. if openAction can specify both dest and action at the same time)

The `OpenAction` entry in the catalog dictionary is specified here (with additional references to the format used): https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#page=82

Ignoring the specification for a moment; basically I don't think the situation you describe will ever happen in practice, at least I've never seen it, and anyway `Catalog.parseDestDictionary` will only extract *one* entry from the dictionary.

> [...] so `if (!openAction)` always evaluates to true again.

We can't initialize the Object unconditionally, since we want the default case (i.e. no `OpenAction` entry present) to be easy to detect, hence the reason for initializing the Object only where necessary.

> At https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L621 it is checked whether `openAction` is null (although it can't be any other value). At https://github.com/mozilla/pdf.js/blob/master/src/core/obj.js#L626 `openAction` is checked again,

Well, generally speaking it was much easier to simply treat each individual case *separately* (such that e.g. any future changes to this code won't end up accidentally overwriting data).

*Ninja-edit:* Basically you're saying, assuming I understand it correctly, that the it would be possible to assign `openAction = Object.create(null);` *unconditionally*. While that should work, given the current implementation, that however shouldn't mean that the existing code is thus wrong here :-)

---

Given that the code, as written, works fine as far as I'm reading the specification it should hopefully not be necessary to make changes here (unless there's PDF examples to the contrary of course).

automatedbugreportingfacility:
> Edit: Basically you're saying, assuming I understand it correctly, that it would be possible to assign openAction = Object.create(null); unconditionally in each branch. While that should work, given the current implementation, that however doesn't necessarily mean that the existing code is thus wrong here :-)

Yes, that'd be possible, `if (true)` is picked up by some static analyzers, btw: https://lgtm.com/projects/g/mozilla/pdf.js/?mode=list 

If all is good according to the spec (and there's an OR between destination and action, so it seems so), then it's good, thanks for checking.

timvandermeij:
That's an interesting website. I'll open a new ticket to see if we can resolve some of the warnings there. Most code is probably fine, but it can't hurt to check and maybe to rewrite it a bit to not trip up static analyzers. Thanks!

PR #11646, opened 2020-03-01
Ensure that automatic printing still works when the viewer and/or its pages are hidden (bug 1618621, bug 1618955)

### LINKED ISSUE #8208: FireFox - "offsetParent is not set -- cannot scroll  viewer.js:125:3"
Link to website with sample PDF file:
https://codecentral.devexpress.com/T227361/?Print=true

Configuration:
FireFox v.54.0a2
Win10 x64 [Version 10.0.14393]
- PDF.js version:
(PDF.js: 1.7.348)
- Is an extension:
No (default pdf.js viewer in the firefox)

1. open site https://codecentral.devexpress.com/T227361/
2. press the "Print" button
direct preview of pdf file is working normally.
but if pdf file opened from the web page - there will be js error and no pfd.js viewer opened.

The PDF.js viewer is not opened.

console output:
> offsetParent is not set -- cannot scroll  viewer.js:125:3
> PDF 0a60e3235e22da40aa53a4e284cde0a4 [1.4 Developer Express Inc. DXperience (tm) v16.2.3 / -] (PDF.js: 1.7.348)

Snuffleupagus:
Please note that this patch, on its own, won't magically fix all of these printing bugs without [bug 1618553](https://bugzilla.mozilla.org/show_bug.cgi?id=1618553) also being fixed.
(However I don't foresee that being too difficult, famous last words :-), but it will as suggested require a platform API that we can notify when the viewer is ready.)

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1618621
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1618955
Fixes #8208

/cc @brendandahl

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -397,6 +397,27 @@ class BaseViewer {
   }
 
   /**
+   * @private
+   */
+  _onePageRenderedOrForceFetch() {
+    // Unless the viewer *and* its pages are visible, rendering won't start and
+    // `this._onePageRenderedCapability` thus won't be resolved.
+    // To ensure that automatic printing, on document load, still works even in
+    // those cases we force-allow fetching of all pages when:
+    //  - The viewer is hidden in the DOM, e.g. in a `display: none` <iframe>
+    //    element; fixes bug 1618621.
+    //  - The viewer is visible, but none of the pages are (e.g. if the
+    //    viewer is very small); fixes bug 1618955.
+    if (
+      !this.container.offsetParent ||
+      this._getVisiblePages().views.length === 0
+    ) {
+      return Promise.resolve();
+    }
+    return this._onePageRenderedCapability.promise;
+  }
+
+  /**
    * @param pdfDocument {PDFDocument}
    */
   setDocument(pdfDocument) {
@@ -492,7 +513,7 @@ class BaseViewer {
         // Fetch all the pages since the viewport is needed before printing
         // starts to create the correct size canvas. Wait until one page is
         // rendered so we don't tie up too many resources early on.
-        this._onePageRenderedCapability.promise.then(() => {
+        this._onePageRenderedOrForceFetch().then(() => {
           if (this.findController) {
             this.findController.setDocument(pdfDocument); // Enable searching.
           }
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -391,9 +391,9 @@ class BaseViewer {
   /**
    * @private
    */
-  get _setDocumentViewerElement() {
+  get _viewerElement() {
     // In most viewers, e.g. `PDFViewer`, this should return `this.viewer`.
-    throw new Error("Not implemented: _setDocumentViewerElement");
+    throw new Error("Not implemented: _viewerElement");
   }
 
   /**
@@ -479,7 +479,7 @@ class BaseViewer {
 
         for (let pageNum = 1; pageNum <= pagesCount; ++pageNum) {
           const pageView = new PDFPageView({
-            container: this._setDocumentViewerElement,
+            container: this._viewerElement,
             eventBus: this.eventBus,
             id: pageNum,
             scale,
--- a/web/pdf_single_page_viewer.js
+++ b/web/pdf_single_page_viewer.js
@@ -27,12 +27,12 @@ class PDFSinglePageViewer extends BaseViewer {
     });
   }
 
-  get _setDocumentViewerElement() {
+  get _viewerElement() {
     // Since we only want to display *one* page at a time when using the
     // `PDFSinglePageViewer`, we cannot append them to the `viewer` DOM element.
     // Instead, they are placed in a `DocumentFragment`, and only the current
     // page is displayed in the viewer (refer to `this._ensurePageViewVisible`).
-    return shadow(this, "_setDocumentViewerElement", this._shadowViewer);
+    return shadow(this, "_viewerElement", this._shadowViewer);
   }
 
   _resetView() {
--- a/web/pdf_viewer.js
+++ b/web/pdf_viewer.js
@@ -17,8 +17,8 @@ import { BaseViewer } from "./base_viewer.js";
 import { shadow } from "pdfjs-lib";
 
 class PDFViewer extends BaseViewer {
-  get _setDocumentViewerElement() {
-    return shadow(this, "_setDocumentViewerElement", this.viewer);
+  get _viewerElement() {
+    return shadow(this, "_viewerElement", this.viewer);
   }
 
   _scrollIntoView({ pageDiv, pageSpot = null, pageNumber = null }) {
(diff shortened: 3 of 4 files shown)


Snuffleupagus:
While we still need to sort out the "load" event, in order to fix all of these issues completely, even as-is this PR is a *considerable* improvement in my testing.

Having tried all of the test-cases from the bugs/issues linked in https://github.com/mozilla/pdf.js/pull/11646#issue-382047534 in a Firefox build with this patch applied, they all work now *if* the user triggers printing *manually* since that gives the viewer/document enough time to load.
(Only the *immediate* automatic printing on load in an `<iframe>` is still broken, pending the "load" event fix.)

@brendandahl Can we land this PR now, since it's ready, rather than waiting for the "load" event fix?

Snuffleupagus:
@brendandahl Gentle review ping; as outlined in https://github.com/mozilla/pdf.js/pull/11646#issuecomment-594461119 this patch improves the current situation enough to warrant landing as soon as possible in my opinion :-)

Snuffleupagus:
*Rebasing onto `master` to ensure that the patch still applies cleanly; no code changes were made.*

web/base_viewer.js
```diff
@@ -396,6 +396,51 @@ class BaseViewer {
     throw new Error("Not implemented: _setDocumentViewerElement");
   }
 
+  /**
+   * @private
+   */
+  get _setDocumentAllowFetchPages() {
```
brendandahl: Having a getter that is called `set` is a bit confusing. How about a regular method that is something like `autoFetchDelay()`?

brendandahl (changes requested):

web/base_viewer.js
Snuffleupagus: > Having a getter that is called set is a bit confusing.

That was simply to sort-of "tie" the functionality to the `setDocument` method, similar to the exisiting `_setDocumentViewerElement` getter.

> How about a regular method that is something like `autoFetchDelay()`?

Out of curiosity: Is there a reason that a regular method would be preferred over a getter, e.g. related to performance?
Also, how about simply `allowFetchPages` for the name instead? At least to me a name such as "autoFetchDelay" suggests that something will always load, however that's not actually guaranteed here (e.g. in the `disableAutoFetch=true` case).

brendandahl: I also find `_setDocumentViewerElement` pretty confusing. 

There's not really a perf difference or hard fast rule, it's more I prefer that getters be relatively simple property accessors vs a method signals that there's going to be more "work" going on when I call it.

Good point on the autoFetch term. `_allowFetchPages` implies i could do something like `if (_allowFetchPages())`, which also feels a bit off too. 

Before we brainstorm some more names, what about if we changed the viewer to not only wait for an request animation frame, but race it with a requestIdleCallback with the timeout argument?

brendandahl: e.g. in https://github.com/mozilla/pdf.js/blob/c3f4690bde8137d80c74203b1ad91476fc2ca160/src/display/api.js#L2870

Snuffleupagus: > [...] it's more I prefer that getters be relatively simple property accessors vs a method signals that there's going to be more "work" going on when I call it.

Well, I can't really argue with that :-)

> what about if we changed the viewer to not only wait for an request animation frame, but race it with a requestIdleCallback with the timeout argument?

Sorry, but I'm not entirely sure that I follow your train of thought here; mind elaborating?

Also, cross-browser support for `requestIdleCallback` doesn't seem to be that great according to https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback#Browser_compatibility (since we still have to support both Safari and IE11), so you'd have to rely on polyfills.

> https://github.com/mozilla/pdf.js/blob/c3f4690bde8137d80c74203b1ad91476fc2ca160/src/display/api.js#L2870

Again, I'm sure that I understand how that relates to the viewer...

brendandahl: Part of the reason we don't ever resolve the onePageRendered promise is because we wait for a request animation frame, which we never get when the pdf viewer is hidden (we'd still also have to handle no pages being visible issue). After thinking about this some more though, we'd be rendering a page that isn't visible and doing work we don't really need to do.

Looking at telemetry, a lot of PDFs can take more than a 1 second to render, so I'd pref we handle the above bugs with without racing promises and a timeout.

Can we instead do something like:
```
onePageRenderedOrForceFetch() {
  if (!this.container.offsetParent || 
        !this._location || /* not sure why this is here below */
        this._getVisiblePages().views.length === 0) {
    return Promise.resolve();
  }
  return this._onePageRenderedCapability.promise;
}
```

Snuffleupagus: > [...] a lot of PDFs can take more than a 1 second to render, so I'd pref we handle the above bugs with without racing promises and a timeout.

The "viewer hidden"-case is easy to handle (with just `this.container.offsetParent`), but the "viewer visible, but all pages hidden"-case is more complicated.
I simply wanted to avoid doing more work *every single time* that the viewer loads normally, with pages being visible, since [`getVisibleElements`](https://github.com/mozilla/pdf.js/blob/master/web/ui_utils.js#L441) may be a bit expensive in long document (thousands of pages).

> Can we instead do something like:
> 
> ```
> onePageRenderedOrForceFetch() {
>   if (!this.container.offsetParent || 
>         !this._location || /* not sure why this is here below */
>         this._getVisiblePages().views.length === 0) {
>     return Promise.resolve();
>   }
>   return this._onePageRenderedCapability.promise;
> }
> ```

If we do this we're *forced* to always call `this._getVisiblePages()` every time that the viewer loads, just to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1618955 which should hopefully be a *very* rare use-case.

brendandahl: Could we do a cheaper check to see if just the page we want to render is visible?

Snuffleupagus: > Could we do a cheaper check to see if just the page we want to render is visible?

Sure we can, it just took me a little while to figure out a good/safe way of doing so!
I've implemented this as a special case in `getVisibleElements`, such that we'll check the first page immediately but still fallback to the normal binary search to hopefully prevent any future errors.

The latest version of the patch thus implements your suggestion from https://github.com/mozilla/pdf.js/pull/11646#discussion_r396621823; thank you!

Snuffleupagus: Sorry, I managed to made a *really* stupid mistake when reading the existing code!

Apparently I completely missed this part of the `binarySearchFirstItem` helper function: https://github.com/mozilla/pdf.js/blob/85838fc50507f4a98d97930b4ba8b40f388d39a4/web/ui_utils.js#L228-L237
Obviously that code already contains a fast-path for the *first* element, which means that just calling `this._getVisiblePages()` immediately after the pages have been initialized should be completely fine (since the first page is always visible initially).

Again, sorry about the churn here!
The latest version of the patch implement your suggestion from https://github.com/mozilla/pdf.js/pull/11646#discussion_r396621823; thank you!

brendandahl (approved):

timvandermeij:
Thanks!

PR #11653, opened 2020-03-02
Ensure that there's always a setFont (Tf) operator before text rendering operators (issue 11651)

### LINKED ISSUE #11651: TJ operation without text but only offsets breaks rending

Configuration:
- Creator/Producer: typeset.sh
- Firefox 73.0.1 (64-bit)
- macOS Catalina (10.15.3)
- PDF.js: 2.4.254
- Is a browser extension: YES

A PDF operation with a not quite valid TJ operation that only contains offsets breaks the rest of the page.

Open PDF
[inline-block2.actual.pdf](https://github.com/mozilla/pdf.js/files/4275710/inline-block2.actual.pdf)
```
stream
q
1 0 0 1 0 300 cm
BT
1 0 0 1 0 -20.82 Tm
[-3092 -3343] TJ #invalid - but should not break
ET
Q
```

Skip the operation and continue rendering as normal.
expected

actual

Snuffleupagus:
The PDF document in question is *corrupt*, since it contains multiple instances of incorrect operators.
We obviously don't want to slow down parsing of *all* documents (since most are valid), just to accommodate a particular bad PDF generator, hence the reason for the inline check before calling the `ensureStateFont` method.

Fixes #11651 (assuming we care to do so...)


timvandermeij (approved):

timvandermeij:
Looks good!

timvandermeij:
/botio makeref (also for the other PR)

PR #11654, opened 2020-03-02
Simplify the `BaseFontLoader.isFontLoadingAPISupported` getter

Snuffleupagus:
It's no longer necessary to special-case this getter in the `GenericFontLoader` case, since the GENERIC build hasn't been using `mozPrintCallback` for years now (furthermore Firefox 63 is really old as well).

### DIFF
--- a/src/display/font_loader.js
+++ b/src/display/font_loader.js
@@ -113,9 +113,9 @@ class BaseFontLoader {
     unreachable("Abstract method `_queueLoadingCallback`.");
   }
 
-  // eslint-disable-next-line getter-return
   get isFontLoadingAPISupported() {
-    unreachable("Abstract method `isFontLoadingAPISupported`.");
+    const supported = typeof document !== "undefined" && !!document.fonts;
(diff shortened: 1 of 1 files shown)


src/display/font_loader.js
```diff
@@ -113,9 +113,9 @@ class BaseFontLoader {
     unreachable("Abstract method `_queueLoadingCallback`.");
   }
 
-  // eslint-disable-next-line getter-return
   get isFontLoadingAPISupported() {
-    unreachable("Abstract method `isFontLoadingAPISupported`.");
+    const supported = document !== "undefined" && !!document.fonts;
```
timvandermeij: I think this should be `typeof document`, just like in the old code, because that gives a string return value whereas `document` is most likely the `undefined` type.

Snuffleupagus: Sigh, this is a *really* stupid copy-paste error on my part; thanks for spotting it :-)

timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #11655, opened 2020-03-03
[api-minor] Remove the `getGlobalEventBus` viewer functionality, and the `eventBusDispatchToDOM` option/preference (PR 11631 follow-up)

Snuffleupagus:
- **[api-minor] Remove the `getGlobalEventBus` viewer functionality (PR 11631 follow-up)**

   The correct/intended way of working with the "viewer components" is by providing an `EventBus` instance upon initialization, and the `getGlobalEventBus` was only added for backwards compatibility.
Note, for example, that using `getGlobalEventBus` doesn't really work at all well with a use-case where there's *multiple* `PDFViewer` instances on a one page, since it may then be difficult/impossible to tell which viewer a particular event originated from.

   All of the "viewer components" examples have been previously updated, such that there's no longer any code/examples which relies on the now removed `getGlobalEventBus` functionality.

 - **[api-minor] Remove the `eventBusDispatchToDOM` option/preference, and thus the general ability to dispatch "viewer components" events to the DOM**

   This functionality was only added to the default viewer for backwards compatibility and to support the various PDF viewer tests in mozilla-central, with the intention to eventually remove it completely.
While the different mozilla-central tests cannot be *easily* converted from DOM events, it's however possible to limit that functionality to only MOZCENTRAL builds *and* when tests are running.

   Rather than depending of the re-dispatching of internal events to the DOM, the default viewer can instead be used in e.g. the following way:
   ```javascript
       // The viewer has now been initialized, and its properties can be accessed.

       PDFViewerApplication.eventBus.on("pagerendered", function(event) {
         console.log("Has rendered page number: " + event.pageNumber);
       });
     });
   });
   ```

### DIFF
--- a/web/firefoxcom.js
+++ b/web/firefoxcom.js
@@ -345,6 +345,13 @@ class FirefoxExternalServices extends DefaultExternalServices {
     );
     return shadow(this, "supportedMouseWheelZoomModifierKeys", support);
   }
+
+  static get isInAutomation() {
+    // Returns the value of `Cu.isInAutomation`, which is only `true` when e.g.
+    // various test-suites are running in mozilla-central.
+    const isInAutomation = FirefoxCom.requestSync("isInAutomation");
+    return shadow(this, "isInAutomation", isInAutomation);
+  }
 }
 PDFViewerApplication.externalServices = FirefoxExternalServices;
 
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -58,10 +58,6 @@
       "type": "boolean",
       "default": false
     },
-    "eventBusDispatchToDOM": {
-      "type": "boolean",
-      "default": false
-    },
     "pdfBugEnabled": {
       "title": "Enable debugging tools",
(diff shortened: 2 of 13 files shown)


Snuffleupagus:
With the necessary `PdfStreamConverter.jsm` method landed in [bug 1625430](https://bugzilla.mozilla.org/show_bug.cgi?id=1625430), this PR should be ready for review now.

web/firefoxcom.js
```diff
@@ -345,6 +345,11 @@ class FirefoxExternalServices extends DefaultExternalServices {
     );
     return shadow(this, "supportedMouseWheelZoomModifierKeys", support);
   }
+
+  static get isInAutomation() {
+    const isInAutomation = FirefoxCom.requestSync("isInAutomation");
```
timvandermeij: Since I had to look up what `isInAutomation` did exactly, would it be good to add a one-line comment here along the lines of "`isInAutomation` checks if the code is run in the context of the Firefox test suite"?

timvandermeij (approved):

timvandermeij:
Nice work!

Issue #7442, opened 2016-06-25
Search PDF for phrases and terms simultaneously


mainseq:
https://mozilla.github.io/pdf.js/web/viewer.html

Configuration:
- Web browser and its version: Chrome
- Operating system and its version: Windows 10
- PDF.js version: Current
- Is an extension: No

Steps to reproduce the problem:

I am trying to search tracemonkey for "Dynamic languages", "Google Docs", and JavaScript at the same time. How would I construct my URL? This used to work in #5496. Could someone post an example? I tried everything- must not work. Thanks.

What is the expected behavior? Words should be highlighted

What went wrong? Words not highlighted

I contributed 100+ hours to the original PR.

mainseq:
Instead of honoring the double-quotes in the #search it just strips them out. By parsing the terms correctly in the match array (viewer.js) it works perfectly. I made it work correctly in my tests by adding one line of code.

mainseq:
calcFindWordMatch: function PDFFindController_calcFindWordMatch(
      query, pageIndex, pageContent) {
      var matchesWithLength = [];
      // Divide the query into pieces and search for text on each piece.
      var queryArray = query.match(/\S+/g);
      var subquery, subqueryLen, matchIdx;

      for (var i = 0, len = queryArray.length; i < len; i++) {
        subquery = queryArray[i];
        subqueryLen = subquery.length;
        matchIdx = -subqueryLen;
        
        **//check for equal-sign delimited phrase (supported in #5496)
        //parsing dbl quote phrases will require an additional loop
        subquery = subquery.replace(/\=/ig, ' ');**

        while (true) {
          matchIdx = pageContent.indexOf(subquery, matchIdx + subqueryLen);
          if (matchIdx === -1) {
            break;

yurydelendik:
There is no intent to extend functionality of the `#search=` for compatibility with other readers. (However internal search engine PDFFindController API can be modified to to perform better search queries, but it shall not regress existing functionality)

mainseq:
You would certainly want to highlight a set of phrases just as you would want to highlight a set of terms. I believe that functionality is currently broken. Only one phrase at time can be highlighted. The fix is rather trivial.

kellyrmilligan:
[...]

PR #11659, opened 2020-03-03
[feat] add multiphrase search

korial29:
This feature allows to make a multi-phrase search in the pdf.

Combined with the highlightAll option, we are able to highlight multiple phrases.

Each phrases must be separated by "$$" and passed in the search url parameter.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2272,6 +2272,7 @@ function webViewerFind(evt) {
   PDFViewerApplication.findController.executeCommand("find" + evt.type, {
     query: evt.query,
     phraseSearch: evt.phraseSearch,
+    multiplePhraseSearch: evt.multiplePhraseSearch,
     caseSensitive: evt.caseSensitive,
     entireWord: evt.entireWord,
     highlightAll: evt.highlightAll,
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
Was this PR perhaps submitted to the PDF.js repository by mistake, since it contains changes that only seem relevant for a custom implementation (e.g. in `web/pdf_print_service.js`) of the default viewer!?

Furthermore, please note that we don't want to add any new *non-standard* hash parameters; as was explained in https://github.com/mozilla/pdf.js/issues/7442#issuecomment-229696596. (Also, the separator used here doesn't really make sense IMHO...)

Finally, note that we purposely don't want to add features to the PDF.js default viewer findbar, unless said feature *also* exists in the native Firefox findbar (since that's being used in the built-in Firefox PDF viewer).

All-in-all, I'm sorry to say that I don't think this PR is acceptable here and would suggest closing it as a duplicate of issue #7442.

korial29:
Thank you for your fast answer :)
I removed the web/pdf_print_service.js, it was not supposed to be pushed, sorry.

PR #11660, opened 2020-03-03
Fix Type1 font parsing when .notdef is not at index zero

### LINKED ISSUE #11477: Incorrect text rendering (letter "A" sprinkled everywhere)
### Attach (recommended) or Link to PDF file here:
URL: https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.693.7686&rep=rep1&type=pdf  
Attached (same file): [10.1.1.693.7686.pdf](https://github.com/mozilla/pdf.js/files/4025798/10.1.1.693.7686.pdf)

Configuration: | .
-- | --
Web browser and its version | Firefox 71.0 (64-bit)
Operating system and its version | Arch Linux
PDF.js version | Firefox 71.0 (64-bit)
Is a browser extension | Yes, Firefox's built-in PDF.js

1. Open URL / Attachment
2. Observe

### What is the expected behavior? (add screenshot)
(As viewer by atril 1.22.2-3 (MATE document viewer))
image

### What went wrong? (add screenshot)
image

janpe2:
Fixes #11477
The PDF draws many space characters but the embedded fonts don't have a glyph named `space`, so `.notdef` should be drawn instead. PDF.js assumed that Type1 fonts define `.notdef` as the first glyph (index 0). However, now the fonts have the glyph `A` at index 0 and `.notdef` is the last one, so `A` appears where spaces are expected.

Because the rest of the font machinery in `core/fonts.js` assumes `.notdef` is at index zero, it's easiest to modify `core/type1_parser.js` so that it "repairs" fonts and makes sure `.notdef` is at index 0.

We are allowed to change the order of glyphs because Type1 fonts identify glyphs using glyph *names* instead of indices. PDF.js assigns indices to the glyphs when the Type1 parser has finished its job and returns an array of charstrings. 

The Type1 Spec doesn't require that `.notdef` must be at index 0, so the fonts are not broken.


janpe2:
To prove that changing the order of glyphs causes no problems, I temporarily added this "shuffling" in the end of Type1Parser.prototype.extractFontProgram
```
if (program.charstrings.length >= 8) {
  console.log("Shuffling Type1 glyphs");
  const array = program.charstrings;
  let tmp;
  tmp = array[1]; array[1] = array[3]; array[3] = tmp;
  tmp = array[2]; array[2] = array[4]; array[4] = tmp;
  tmp = array[5]; array[5] = array[7]; array[7] = tmp;
}
```
The PDF from the issue still renders fine.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #11664, opened 2020-03-04
Prevent the zoom dropdown icon from being overridden when the element is `:active` (bug 1619595)

Snuffleupagus:
This changes the dropdown icon from being set using the `background` CSS property, to being set with `::after` which is *similar* to all the other toolbar button icons (which use `::before`).
Also tweaks the dropdown `background-color` on `:hover` slightly, since the other changes made it too light.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1619595


timvandermeij (approved):

timvandermeij:
This does look better. Thanks!

PR #11667, opened 2020-03-05
Add a deprecation warning for the `eventBusDispatchToDOM` option/preference (PR 11631 follow-up)

Snuffleupagus:
- **Update the `waitOnEventOrTimeout` helper function to handle internal events consistently with the rest of the viewer components (PR 11631 follow-up)**

   I overlooked this in PR #11631; sorry about that!

   Also, ensure that `EventBus` instances *always* track "external" events using a boolean regardless of the actual option value.

 - **Move the `dispatchDOMEvent` functionality out from the `EventBus` and add a deprecation warning for the `eventBusDispatchToDOM` option/preference (PR 11631 follow-up)**

   It occured to me that similar to the `getGlobalEventBus` function, it's probably a good idea to *also* notify users of the fact that `eventBusDispatchToDOM` is now deprecated.

   Rather than depending of the re-dispatching of internal events to the DOM, the default viewer can instead be used in e.g. the following way:
   ```javascript
       // The viewer has now been initialized, and its properties can be accessed.

       PDFViewerApplication.eventBus.on("pagerendered", function(event) {
         console.log("Has rendered page number: " + event.pageNumber);
       });
     });
   });
   ```

### DIFF
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -716,7 +716,7 @@ function waitOnEventOrTimeout({ target, name, delay = 0 }) {
 
     function handler(type) {
       if (target instanceof EventBus) {
-        target.off(name, eventHandler);
+        target._off(name, eventHandler);
       } else {
         target.removeEventListener(name, eventHandler);
       }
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Yes, that's indeed a good idea. Thanks!

PR #11673, opened 2020-03-06
Update the `FontLoader.bind` method to avoid explicitly returning `undefined`

Snuffleupagus:
The only reason for the `return undefined;` lines was to appease the ESLint `consistent-return` rule, but that's not actually necessary if you make use of the fact that the method is `async` and that we can thus await the Promise rather than returning it.


timvandermeij (approved):

timvandermeij:
Nice find!

PR #11680, opened 2020-03-09
Prevent lookup errors in `PartialEvaluator.hasBlendModes` from breaking all parsing/rendering of a page (issue 11678)

### LINKED ISSUE #11678: bad XRef entry
How to discard the problematic data so that no errors can affect the rendering of the entire page

Snuffleupagus:
The PDF document in question is *corrupt*, since it contains an XObject with a truncated dictionary and where the stream contents start without a "stream" operator.

Fixes #11678

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -61,6 +61,7 @@ import {
   WinAnsiEncoding,
   ZapfDingbatsEncoding,
 } from "./encodings.js";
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

/botio makeref

PR #11686, opened 2020-03-10
Magnifier positioning in reftest analyzer

janpe2:
When reftest analyzer shows magnified pixels, there is a seemingly random offset between the mouse position and the magnified position. The reason for this is that reftest analyzer assumes all images have 800 * 1000 pixels but actually the test images have varying sizes.

I copied the needed lines from
https://github.com/mozilla/gecko-dev/blob/master/layout/tools/reftest/reftest-analyzer.xhtml
(I assume that originally reftest analyzer has been copied from similar Mozilla repositories.)

### DIFF
--- a/test/resources/reftest-analyzer.html
+++ b/test/resources/reftest-analyzer.html
@@ -104,7 +104,7 @@ Original author: L. David Baron <dbaron@dbaron.org>
         <input type="checkbox" id="differences"> Circle differences
       </label>
     </form>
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
> (I assume that originally reftest analyzer has been copied from similar Mozilla repositories.)

That would be correct, as far as I can recall.

/botio test

timvandermeij (approved):

timvandermeij:
Nice improvement!

PR #11687, opened 2020-03-10
Document properties

derek-nguyen123:
Added focus to close dialog when opening document properties.

### DIFF
--- a/web/pdf_document_properties.js
+++ b/web/pdf_document_properties.js
@@ -84,10 +84,10 @@ class PDFDocumentProperties {
     );
 
     if (eventBus) {
-      eventBus._on("pagechanging", evt => {
+      eventBus.on("pagechanging", evt => {
         this._currentPageNumber = evt.pageNumber;
       });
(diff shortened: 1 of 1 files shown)


Snuffleupagus (changes requested):
Assuming that we want to make this sort of change, I'm sorry to say that in its current form the patch isn't correct.
First of all, you're not allowed to lookup a DOM element inline like that, and you need to use the `closebutton` as defined in https://github.com/mozilla/pdf.js/blob/af8d0b9597ccd0e020910eafd74dd6ad140db520/web/pdf_document_properties.js#L63
Secondly, this patch sets the focus even if the overlay doesn't open which can obviously cause issues with incorrectly stealing of focus (you'd need to move the line down to the Promise handler to address that).

Also, please make sure that you write good commit messages, outlining (briefly) both what's being changed and why.

timvandermeij:
I'm not really convinced this should be fixed, and if we do it should be a simple change. Since there has been no response to the comment above, let's close this for now.

PR #11689, opened 2020-03-11
11658 return preserve error types on web

### LINKED ISSUE #11658: Preserve error types (e.g. InvalidPDFException, MissingPDFException)
Configuration:
- Web browser and its version: Chrome 80.0.3987.122 64-bit
- PDF.js version: 2.4.384 (build: 30e0f028)
- Is a browser extension: no

1. Set a non-existing (404) PDF URL for the viewer
2. Open pdf.js
3. Open DevTools Inspector
4. Find uncaught error of generic error/exception type `Uncaught (in promise) Error: Missing PDF file.`

4. Find uncaught error of specific error/exception type `Uncaught (in promise) MissingPDFException: Missing PDF file.`

By preserving the exception type, more fine-grained error handling could be performed via client-side logic (e.g. redirect to a search page if a PDF is not found, or to a ticket system in case of invalid PDF files).

Apparently, the error message translation in [app.js](https://github.com/mozilla/pdf.js/blob/30e0f028b54c50b1bfe5d90e878b62003c8e5173/web/app.js#L816) always throws a generic Error instead of re-throwing with the same exception class of the previous exception.

IMHO, changing the exception class should be BC compatible, as the custom exceptions such as [MissingPDFException extend BaseException](https://github.com/mozilla/pdf.js/blob/30e0f028b54c50b1bfe5d90e878b62003c8e5173/src/shared/util.js#L437), which in turn [uses the generic Error as prototype](https://github.com/mozilla/pdf.js/blob/30e0f028b54c50b1bfe5d90e878b62003c8e5173/src/shared/util.js#L407).

Thank you very much!

huang325:
Fixes #11658  Update web/app.js to throw preserved error type such when pdf file missing or invalid.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -784,6 +784,10 @@ const PDFViewerApplication = {
 
         const message = exception && exception.message;
         let loadingErrorMessage;
+        // #11658 return preserved error types
+        let _errorInstance = function(msg) {
+          return new Error(msg);
+        };
         if (exception instanceof InvalidPDFException) {
           // change error message also for other builds
           loadingErrorMessage = this.l10n.get(
@@ -791,6 +795,10 @@ const PDFViewerApplication = {
             null,
             "Invalid or corrupted PDF file."
           );
+          // Update genaric error type with preserved error type
+          _errorInstance = function(msg) {
+            return new InvalidPDFException(msg);
+          };
         } else if (exception instanceof MissingPDFException) {
           // special message for missing PDF's
           loadingErrorMessage = this.l10n.get(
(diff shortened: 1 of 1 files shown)


Snuffleupagus (changes requested):
Assuming that we even want to fix the referenced issue, which doesn't seem entirely clear at this point (since no explicit decision was made in the issue), I'm sorry to say that this patch really does *not* seem like the correct solution.

For example: why are you re-wrapping the errors rather than simply throwing the existing errors?
As-is this is adding a completely unnecessary/unwanted complexity to the code, and is also not preserving the *actual* Errors as they arrive from the API, for something that (again assuming that we want to change it) should be a one-liner patch.

Note that it you add code comments (the sort of comments added here doesn't seem necessary), they should be formatted/spelled correctly (e.g. end with a period).

Also, please make sure that you write good commit messages, outlining (briefly) both what's being changed *and* why.

timvandermeij:
I also find that this adds code complexity for something that could be solved easier by re-throwing the original exception. I do think the original issue has a good point about being able to have more fine-grained error handling, but it can most likely be implemented in a simpler way. Thanks.

Issue #11154, opened 2019-09-17
Function suggestion: Horizontal Scrolling from right to left


heangfat:
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version: Any
- Operating system and its version: Any
- PDF.js version: Any (master)
- Is a browser extension:

Steps to reproduce the problem:
1. Load any PDF file.
2. Choose Horizontal Scrolling from Tools menu.

What is the expected behavior? (add screenshot)
In Horizontal Scrolling mode, there should be 2 order options for pages: left to right (existing) and right to left (desired). The desired RTL page order is for reading CJK vertical texts. Below shows a correct screenshot:
![correctOrderRTL](https://user-images.githubusercontent.com/20491276/65020141-2527da00-d95f-11e9-8845-758f53827db7.jpg)

What went wrong? (add screenshot)
Currently there is only one order in Horizontal Scrolling: left to right. But for CJK vertical texts, lines are ordered from right to left, i.e., Line _k_+1 is on the left of Line _k_, thus Page _N_+1 should be placed on the left of Page _N_.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): N.A.

thesutphin:
Do you have anymore information on this? I'm thinking of looking more into this. Any info you got would be helpful!

andriasw:
So the following is the issue I would have come up with, but this issue #11154 is similar so I'll just try to have my comment here and make things clear.

=============

### Title: Add right-to-left rearragement support?

Some books published in Taiwan use right-to-left layout for pages. The feature request is: add a feature where, especially in horizontal scrolling mode, one can read PDFs in a right-to-left order by rearraging the pages.

I do not think that the request have to do with CJK specifically, and did not find the original explanation by @heangfat  clear enough. All that is needed is that, **when in horizontal mode with the desire RTL option enabled, the pages begin rightmost, and the last page should be on the leftmost side.**

All following screenshots of PDF.js version 2.3.164, based on the [Chrome extension](https://chrome.google.com/webstore/detail/pdf-viewer/oemmndcbldboiebfnladdacbdfmadadm) by @Rob--W , last updated September 13, 2019 (I hope it keeps in pace with the PDF.js releases btw if necessary).

To help with comprehension for those who are not familiar with CJK layout or any East Asian languages, I am going to demostrate using a 4-page PDF with numbers in the pages respectively.

[...]

PR #11691, opened 2020-03-11
11154

huang325:
#11154 Create a button to reverse the page traversal order on web.


Snuffleupagus (changes requested):
Unfortunately this patch would actually break way more than it fixes, since it *completely* breaks the existing spreadModes given the way that it's implemented by simply overwriting *any* existing page structure. (And adding a bunch of hacks to try and work-around that does *not* seem like a good idea for reasons of readability/maintainability/testability/review-ability of the code.)

While the issue only mentions horizontal scrolling, if we actually want to implement some kind of page-reversal, it should be done in a much more general way such that the viewer *also* supports API `PageLayout` values such as e.g. TwoColumnLeft/TwoColumnRight (i.e. handling spreadModes correctly).

Besides the issues mentioned above, which suggests that this wasn't properly tested (since it breaks existing functionality), there's also the question if the functionality of the patch *itself* is even working correctly on its own!?
Unfortunately it turns out that when you enable horizontal scrolling, and then reverse the pages, rendering stops working *completely* when you actually scroll around in the viewer (pages remain blank). While the problem *could* perhaps point to a pre-existing issue with [`getVisibleElements`](https://github.com/mozilla/pdf.js/blob/6db8e085eed4a4609b24716305a9633cb0b78a93/web/ui_utils.js#L414-L561) (which may or may not be simple to address), it nonetheless doesn't change the fact that the functionality in this patch isn't working even "on its own" and also doesn't really seem well tested.

(Also, you need to make sure that to write good commit messages, outlining (briefly) both what's being changed and why.)

---

Unfortunately this patch is fairly broken, since it doesn't even correctly address the referenced issue, and it also causes significant problems elsewhere; hence I'd probably recommend closing this PR since it seems too far from a correct and workable solution.

All-in-all, it seems that the referenced issue is not really a good beginner bug since fixing all of this could be somewhat complex and time-consuming (especially for someone new to the code).

PR #11698, opened 2020-03-13
Don't accidentally accept invalid glyphNames which *appear* to follow the Cdd{d}/cdd{d} format in `PartialEvaluator._buildSimpleFontToUnicode` (issue 11697)

### LINKED ISSUE #11697: Large blocks of text isn't being rendered
I'm using PDF.js to render many different PDFs from many different sources, but this particular source PDFs has issues for some reason. It also fails to fully render in Firefox directly. Chrome handles this PDF fine somehow. I've been struggling to find the root cause and solution over the past few days. Any help our guidance? I presume it is a faulty PDF but doubt I can ask the source to change their workflow.

[051325202001172HfYHE46U6xM2Ywwht0S8C4A000401010900.pdf](https://github.com/mozilla/pdf.js/files/4331272/051325202001172HfYHE46U6xM2Ywwht0S8C4A000401010900.pdf)

Configuration:
- Web browser and its version: Firefox 73.0.1
- Operating system and its version: macOS 10.15.3 (although reproducible in other OS)
- PDF.js version: latest
- Is a browser extension: latest

1. Open attached PDF file in Firefox or PDF.js viewer or PDF.js extension
2. Missing text (which can be seen in Chrome)

Screen Shot 2020-03-13 at 20 13 03

Screen Shot 2020-03-13 at 20 12 26

Snuffleupagus:
The /Differences array of the problematic font contains a `/c.1` entry, which is consequently detected as a *possible* Cdd{d}/cdd{d} glyphName by the existing heuristics.
Because of how the base 10 conversion is implemented, which is necessary for the base 16 special case, the parsed charCode becomes `0.1` thus causing `String.fromCodePoint` to throw since that obviously isn't a valid code point.

To fix the referenced issue, and to hopefully prevent similar ones in the future, the patch adds *additional* validation of the charCode found by the heuristics.

Fixes #11697


timvandermeij (approved):

timvandermeij:
Thank you for debugging and fixing this!

/botio makeref

PR #11706, opened 2020-03-17
Move the `webViewerOpenFile` function, and the "openfile" eventBus listener, since they only matter in GENERIC builds of the default viewer

Snuffleupagus:
This is similar to the existing implementation of the `webViewerFileInputChange` function and its associated "fileinputchange" event.


timvandermeij (approved):

timvandermeij:
Nice that this is now more grouped together. Thanks!

PR #11707, opened 2020-03-18
Always prefer the PDF.js JPEG decoder for very large images, in order to reduce peak memory usage (issue 11694)

### LINKED ISSUE #11694: Out of memory when encountering complex pdf
[complex.pdf](https://github.com/mozilla/pdf.js/files/4328503/default.pdf)

Configuration:
- Web browser and its version: nwjs 0.34.5 win32
- Operating system and its version:Windows10 64x RAM 8.00GB
- PDF.js version: 2.4.435
- Is a browser extension: yes

1. Load pdf.js file
2. Load pdf file

Just want to load and display normally！！！o(╥﹏╥)o
The client crashed...  (；´д｀)ゞ,   I think it may be out of memory
ヾ(ｏ･ω･)ﾉ no link...

Snuffleupagus:
When JPEG images are decoded by the browser, on the main-thread, there's a handful of short-lived copies of the image data; see https://github.com/mozilla/pdf.js/blob/c3f4690bde8137d80c74203b1ad91476fc2ca160/src/display/api.js#L2364-L2408
That code thus becomes quite problematic for very big JPEG images, since it increases peak memory usage a lot during decoding. In the referenced issue there's a couple of JPEG images whose dimensions are `10006 x 7088` (i.e. ~68 mega-pixels), which causes the *peak* memory usage to increase by close to `1 GB` (i.e. one giga-byte) in my testing.

By letting the PDF.js JPEG decoder, rather than the browser, handle very large images the *peak* memory usage is considerably reduced and the allocated memory also seem to be reclaimed faster.

*Please note:* This will lead to movement in some existing `eq` tests. Refer to https://github.com/mozilla/pdf.js/pull/11523#issuecomment-578740441 for an explanation of the different test "failures".

Fixes #11694 (to the extent that doing so is possible, given the size of the JPEG images).

### DIFF
--- a/src/core/jpeg_stream.js
+++ b/src/core/jpeg_stream.js
@@ -134,6 +134,17 @@ const JpegStream = (function JpegStreamClosure() {
             stream.pos += 2; // Skip marker length.
             stream.pos += 1; // Skip precision.
             const scanLines = stream.getUint16();
+            const samplesPerLine = stream.getUint16();
+
+            // Letting the browser handle the JPEG decoding, on the main-thread,
+            // will cause a *large* increase in peak memory usage since there's
+            // a handful of short-lived copies of the image data. For very big
+            // JPEG images, always let the PDF.js image decoder handle them to
+            // reduce overall memory usage during decoding (see issue 11694).
+            if (scanLines * samplesPerLine > 1e6) {
+              validDimensions = false;
+              break;
+            }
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
I can confirm that the tab doesn't crash anymore. In general I also think that this is a better approach for large images. Thanks!

PR #11715, opened 2020-03-19
make the description for gitpod setup precise.

nisarhassan12:


README.md
```diff
@@ -81,8 +81,13 @@ It is also possible to view all test PDF files on the right side by opening:
...
 However, you may also use [Gitpod](https://gitpod.io) (a free online IDE). With
-a single click it will launch a ready to use development environment, with all
-the necessary dependencies pre-installed and the web server running.
+a single click it will launch a workspace and automatically:
+
+- clone the pdf.js repo.
+- install the dependencies.
+- start the webserver via `gulp server`.
+
+So that you can start straight away.
```
Snuffleupagus: This sentence just looks completely out-of-context and doesn't really fit with the preceding text.

Snuffleupagus (changes requested):
If the point of Gitpod is that it handles all setup for you, why does it matter to the user *exactly* what happens "behind the scenes"?

Since there's no motivation of *why* this is necessary, I'd be inclined to suggest closing this PR as wontfix.

timvandermeij:
I'm also missing the point a bit here. If people are interested in how it works exactly, they would just set it up themselves. The point of the Gitpod setup is to keep things as simple as possible.

PR #11716, opened 2020-03-19
[api-minor] Change the pageIndex, on `PDFPageProxy` instances, to a private property

Snuffleupagus:
This property has never been documented and/or *intentionally* exposed through the API, instead the `PDFPageProxy.pageNumber` property is the documented/intended API to use here.
Hence pageIndex is changed to a "private" property on `PDFPageProxy` instances, and internal API functionality is also updated to *consistently* use `this._pageIndex` rather than a mix of formats.


timvandermeij (approved):

timvandermeij:
Yes, this makes more sense and indeed, that property should have been private. Thanks for fixing this!

PR #11719, opened 2020-03-20
Remove the unused `wideChars` property on `Font` instances

Snuffleupagus:
This property was added in PR #1599 (almost eight years ago), but has been unused ever since PR #3674 (six and a half years ago).

PR #11721, opened 2020-03-20
Don't let Travis run `npm update` during setup

Snuffleupagus:
Letting Travis update npm packages can lead to unexpected and completely unrelated failures for any PR and/or merge, see e.g. #11719, if there's ever backwards *incompatible* changes when a package is updated.

The *exact* package versions are specified in `package-lock.json`, and they should thus be used when running tests. Note that the bots won't update npm packages, and neither should Travis as far as I'm concerned.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,3 @@ install:
   - npm install -g npm@latest
   - npm install -g gulp-cli
   - npm install
-  - npm update


timvandermeij (approved):

timvandermeij:
It looks like I added that four years ago and I must say I can't recall why, but it shouldn't be necessary to have. Thanks!

Snuffleupagus:
> It looks like I added that four years ago and I must say I can't recall why, but it shouldn't be necessary to have. Thanks!

The only thing I can possibly think of is that we didn't use a `package-lock.json` file until much later, meaning that the installed package versions were more arbitrary. Perhaps the idea was that it's better to be guaranteed to use the latest versions, rather than a more arbitrary state.

PR #11724, opened 2020-03-21
Only build the necessary `web/` files during the `gulp default_preferences` task

Snuffleupagus:
By explicitly specifying only the required `web/` files, the runtime of the gulp task is reduced by approximately 30 percent.


timvandermeij (approved):

timvandermeij:
Thank you for optimizing this!

PR #11725, opened 2020-03-21
Re-factor `PDFViewerApplication.load` such that `{PDFViewer, PDFThumbnailViewer}.setDocument` happens slightly earlier

Snuffleupagus:
*Please refer the the individual commit messages.*

Much smaller diff with https://github.com/mozilla/pdf.js/pull/11725/files?w=1

---
*Edit:* I also, as a follow-up, want to look into delaying the `getPageLabels`/`getMetadata` calls until e.g. `firstPagePromise` is resolved since neither of those API calls should be *necessary* early on.


timvandermeij (approved):

timvandermeij:
Looks good to me. Thanks!

PR #11727, opened 2020-03-22
Add a heuristic to scale even single-char text, when the horizontal/vertical scaling differs significantly (issue 11713)

### LINKED ISSUE #11713: Incorrect selection with enlarged text
**Attach PDF file here:**
[selection.pdf](https://github.com/mozilla/pdf.js/files/4354103/selection.pdf)

**Configuration:**
- Web browser and its version: Firefox 74.0 (64 bits)
- Operating system and its version: Windows 10 version 18362.720
- PDF.js version: the one provided by firefox
- Is a browser extension: no

1. Open the file
2. Select all the content

**What is the expected behavior:**
selection expected

This is a screenshot of adobe acrobat, the selection is no exactly what is expected but a illustration of how the selection should scale with the characters.

**What went wrong:**
selection wrong

The selection is wrong when the characters are scaled.

This example is very simplified, but it create selection confusions with mathematical expressions that need multiples characters to be scaled.

Snuffleupagus:
At this point in time, compared to when the "ignore single-char" code was added, we *should* generally be doing a much better job of combining text into as few chunks as possible.
However, there's still bad cases where we're not able to combine text as much as one would like, which is why I'm *not* proposing to simply measure/scale all text. Instead this patch will to only measure/scale single-char text in cases where the horizontal/vertical scale is off significantly, since that's were you'd expect bad text-selection behaviour otherwise.

Note that most of the movement caused by this patch is with Type3 fonts, which is a somewhat special font type and one where our current text-selection behaviour is probably the least good.

Fixes #11713


timvandermeij (approved):

timvandermeij:
Seems to work nicely. Thanks!

/botio makeref

PR #11728, opened 2020-03-22
Remove unnecessary checks from the `PDFDocumentProperties` constructor

Snuffleupagus:
Given that none of the relevant options are marked as optional in the code/JSDocs, and that the `PDFDocumentProperties` class is specific to the default viewer (and not exposed as part of the viewer components), there's no good reason as far as I can tell for these checks.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #11734, opened 2020-03-23
Remove old API methods which were previously converted to throwing (PR 11219 follow-up)

Snuffleupagus:
- **Remove old API methods which were previously converted to throwing (PR 11219 follow-up)**

   These methods were deprecated already in PDF.js version `2.1.266`, see PRs #10246 and #10369, and were converted to throw `Error`s upon invocation in PDF.js version `2.4.456`, see PR #11219.
Hence it ought to be possible to remove these methods now.

 - **Exclude the `setPDFNetworkStreamFactory` function from the built API docs**

   Please note that the `setPDFNetworkStreamFactory` functionality isn't exposed in the public API, i.e. not listed among the exports in the `src/pdf.js` file, and that even if it were it wouldn't really be useful considering that none of the `PDFNetworkStream`/`PDFFetchStream`/`PDFNodeStream` classes are exported either.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -481,17 +481,6 @@ const PDFDocumentLoadingTask = (function PDFDocumentLoadingTaskClosure() {
         }
       });
     }
-
-    /**
-     * Registers callbacks to indicate the document loading completion.
-     * @ignore
-     */
-    then(onFulfilled, onRejected) {
-      throw new Error(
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Agreed. Thanks!

PR #11736, opened 2020-03-23
Add `passive: false` to the `wheel` event listener used in `PDFPresentationMode` (issue 11735, PR 10765 follow-up)

### LINKED ISSUE #11735: Error on chrome with presentation mode scroll
On chrome with presentation mode active, when the user scroll we have an error in the console 

'Unable to preventDefault inside passive event listener'

It bring some random behavior (in my in integration of the viewer when the user scroll we reache the next page and a little more, and the user cannot see the full content of the pdf because it never display).

I find a patch, I don't know if it is the good way to do it.
I add this in pdf_presentation_mode

    'window.addEventListener("wheel", this.mouseWheelBind, {passive: false});'

And all work fine in chrome

Snuffleupagus:
Fixes #11735

### DIFF
--- a/web/pdf_presentation_mode.js
+++ b/web/pdf_presentation_mode.js
@@ -423,7 +423,7 @@ class PDFPresentationMode {
 
     window.addEventListener("mousemove", this.showControlsBind);
     window.addEventListener("mousedown", this.mouseDownBind);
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks! This should indeed not be a passive event listener.

PR #11738, opened 2020-03-23
Remove variable shadowing from the JavaScript files in the `src/core/` folder

Snuffleupagus:
*This is part of a series of patches that will try to split PR #11566 into smaller chunks, to make reviewing more feasible.*

Once all the code has been fixed, we'll be able to eventually enable the ESLint no-shadow rule; see https://eslint.org/docs/rules/no-shadow

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -1269,7 +1269,6 @@ var Font = (function FontClosure() {
 
     fallbackToSystemFont: function Font_fallbackToSystemFont() {
       this.missingFile = true;
-      var charCode, unicode;
       // The file data is not specified. Trying to fix the font name
       // to be used with the canvas.font.
       var name = this.name;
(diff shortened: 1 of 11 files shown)


src/core/fonts.js
```diff
@@ -1409,7 +1408,7 @@ var Font = (function FontClosure() {
         tables["post"] = null;
 
         for (let i = 0; i < numTables; i++) {
-          const table = readTableEntry(font);
+          const table = readTableEntry(file);
```
timvandermeij: This one actually looks like it caused a bug since we always read the entry for the `font` variable of the upper scope here and not for the given `file` parameter. If that's indeed the case, I'm surprised we haven't seen any problems because of it, but it would show that deshadowing is a good thing!

timvandermeij (approved):

timvandermeij:
Thanks a lot!

src/core/fonts.js
Snuffleupagus: Yes, this case is absolutely confusing!

The only reason that this hasn't caused any bugs is that these properties are Objects (in this case `instanceof Stream`), and those are essentially passed by reference rather than value in JavaScript (the names `font`/`file` thus point to the same data).

PR #11741, opened 2020-03-24
Remove Gitpod support (issue 11732)

### LINKED ISSUE #11732: Remove Gitpod support
Personally I was somewhat sceptical of the idea of adding support for Gitpod when it originally landed; however I believe that we should re-evaluate if including it was really a good idea.

Should we have evaluated this much more thoroughly *before* accepting PR #11300? Of course, but everything is easier in hindsight :-)
I've noticed the following issues, with Gitpod support, thus far (there may be more):

- None of the PDF.js contributors know/use Gitpod, which means that we cannot provide help/assistance if a new contributor run into trouble trying to use it. Basically, if there's any problems we'd point to the "regular" instructions in https://github.com/mozilla/pdf.js#getting-the-code and suggest to follow that work-flow instead.
Suggesting a work-flow, even an optional one, and then not being able to support it seems wrong (and forcing the core contributors to learn it seems even more wrong).

 - Support for Gitpod may break at any time, since it's completely untested (it may even be broken already, who knows...). Actually supporting it would force the core contributors to manually test it, somewhat regularly, and then spend time/effort to keep it working. Given that all core contributors are following the "regular" work-flow, this seems like something that would unnecessarily take away from getting actual useful work done.

 - As part of the motivation for supporting Gitpod, in https://github.com/mozilla/pdf.js/pull/11300#issue-335434981, the following was stated:
   > [...] Gitpod, a free online tool that lets you automate your dev environment

   Unfortunately that, as far as I'm concerned, significantly misrepresents what [Gitpod](https://www.gitpod.io/) really is with respect to the use of the word "free" here!
   Looking at https://www.gitpod.io/pricing/, it's pretty clear that it's first and foremost a *commercial* service and secondly that its "free" option is *severely* limited (with no special provision for e.g. open-source projects).

 - Since Gitpod is clearly a commercial service, where the "free" option is very limited, it could give contributors the wrong impression of the PDF.js library and thus by extension Mozilla. Let's say that someone uses it enough to hit the maximum hours per month, they could thus *possibly* even interpret things as Mozilla (indirectly) suggesting that they *pay* to continue to contribute.
  For an organization, such as Mozilla, where everything is open/free this doesn't exactly look good.

 - With the PDF.js library supporting Gitpod, this could *possibly* be seen by some as an advertisement and/or even endorsement of a *commercial* product. That *may* even be a violation of Mozilla guidelines/policies (I'm not a lawyer), but at the very least it could very well end up reflecting poorly on Mozilla in the event that there's e.g. any security issues with Gitpod.

   *Ninja-edit:* With the caveat that I don't know if https://twitter.com/gitpod is the offical Twitter account of Gitpod, note https://twitter.com/gitpod/status/1196374786198396928

*In summary:* Given all of the points above, I'd *very strongly* suggest removing Gitpod support from the PDF.js library.

Snuffleupagus:
Fixes #11732


jankeromnes:
See also https://github.com/mozilla/pdf.js/issues/11732#issuecomment-603090995

I don't mind Gitpod support being removed, but please first consider the 800+ contributors who rely on Gitpod to work on PDF.js.

PR #11744, opened 2020-03-24
The first glyph in CFF CIDFonts must be named 0 instead of ".notdef"

### LINKED ISSUE #11718: Incorrect rendering on macOS
Attach (recommended) or Link to PDF file here: https://pdos.csail.mit.edu/6.824/papers/vm-ft.pdf

Configuration:
- Web browser and its version: Firefox 74.0
- Operating system and its version: macOS 10.15.3
- PDF.js version:  PDF 1d2d2343d8fc3e49ada313170169d1a2 [1.5 Adobe PDF Library 8.0 / Adobe InDesign CS3 (5.0.2)] (PDF.js: 2.4.326)
- Is a browser extension: No, I guess.

1. Open the PDF file in Firefox and check the second paragraph of "Introduction"

图片

图片

janpe2:
Fixes #11718 in which the `ff` ligature glyph is at index zero in a CFF font. Beacuse this is a CIDFont, glyph names are CIDs, which are integers. Thus the string `".notdef"` is not correct. The rest of the charset data is already parsed correctly as integers when the boolean argument `cid` is true.


timvandermeij (approved):

timvandermeij:
Nice find! Thank you for fixing this bug.

/botio makeref (also for another patch that landed)

PR #11746, opened 2020-03-25
Create the glyph mapping correctly for composite Type1, i.e. CIDFontType0, fonts (issue 11740)

### LINKED ISSUE #11740: PDF not rendering
[book.pdf](https://github.com/mozilla/pdf.js/files/4373205/book.pdf)

Configuration:
- Web browser and its version: Chromium 80.0.3987.87, Firefox 74.0
- Operating system and its version: Xubuntu 18.04
- PDF.js version: I was checking the latest version available at "https://mozilla.github.io/pdf.js/web/viewer.html"
- Is a browser extension: No

Upload attached pdf file to "https://mozilla.github.io/pdf.js/web/viewer.html" and scroll to the second page.

Chromium built-in reader renders it correctly.
built-in-reader-screenshot

But in PDF.js all the pages except the first one are blank.
pdf-js-screenshot

Snuffleupagus:

Fixes #11740


src/core/fonts.js
```diff
@@ -3353,8 +3353,10 @@ function type1FontGlyphMapping(properties, builtInEncoding, glyphNames) {
         charCodeToGlyphId[charCode] = 0; // notdef
       }
     }
-  } else if (isSymbolicFont) {
+  } else if (isSymbolicFont || properties.hasCustomBuiltInEncoding) {
```
janpe2: Did you notice that the fonts are actually CIDFonts (but their embedded data streams are basic Type1 fonts (non-CIDFonts)). CIDFonts should not be affected by Encodings, so I think your solution based on  `hasCustomBuiltInEncoding` is not the correct one for CIDFonts. Instead I suggest the following fix

```diff
diff --git a/src/core/fonts.js b/src/core/fonts.js
index 614048c9..761bd39c 100644
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -3599,6 +3599,19 @@ var Type1Font = (function Type1FontClosure() {

     getGlyphMapping: function Type1Font_getGlyphMapping(properties) {
       var charstrings = this.charstrings;
+
+      if (properties.composite) {
+        const charCodeToGlyphId = Object.create(null);
+        let glyphId, charCode;
+        // Map CIDs directly to GIDs.
+        for (glyphId = 0; glyphId < charstrings.length; glyphId++) {
+          charCode = properties.cMap.charCodeOf(glyphId);
+          // Add 1 because glyph 0 is duplicated.
+          charCodeToGlyphId[charCode] = glyphId + 1;
+        }
+        return charCodeToGlyphId;
+      }
+
       var glyphNames = [".notdef"],
         glyphId;
       for (glyphId = 0; glyphId < charstrings.length; glyphId++) {
```

Thus each CID is mapped to an equal GID value. A similar CID mapping can be found in `CFFFont.prototype.getGlyphMapping`.

Snuffleupagus: > Did you notice that the fonts are actually CIDFonts (but their embedded data streams are basic Type1 fonts (non-CIDFonts)).

I'm aware of that :-)
The reason for the current approach was that it, at the time, seemed better to trust the *actual* font program rather than the font dictionary (since we've seen time and time again how the latter can be wrong in all kind of "interesting" ways, which thus breaks font rendering).

However your suggestion does seem more in-line with the specification, so assuming that it fixes the issue and doesn't cause regressions elsewhere, it might indeed be better.
I'll update the PR once I've had time to run all tests; thank you!

Snuffleupagus: That seems to have worked; thanks again!

src/core/type1_parser.js
```diff
@@ -457,10 +457,14 @@ var Type1Parser = (function Type1ParserClosure() {
     if (encrypted) {
       var data = stream.getBytes();
       var isBinary = !(
-        isHexDigit(data[0]) &&
+        (isHexDigit(data[0]) || isWhiteSpace(data[0])) &&
         isHexDigit(data[1]) &&
         isHexDigit(data[2]) &&
-        isHexDigit(data[3])
+        isHexDigit(data[3]) &&
+        isHexDigit(data[4]) &&
```
timvandermeij, line 464: In the specification I find the case that the first ciphertext byte may not be whitespace, but I cannot find directly why we would need to check the first eight ciphertext bytes for ASCII hexadecimal character codes. It states that "the  first 4 ciphertext bytes must obey certain restrictions", so no guarantees are given about the remaining bytes as far as I can tell, but I'm probably missing something here...

Snuffleupagus: Note (emphasis mine):
> To distinguish between binary and ASCII hexadecimal input,
the first 4 ciphertext bytes must obey certain restrictions.
**Remember that two ASCII hexadecimal characters represent 1
ciphertext byte, while 1 binary byte represents 1 ciphertext
byte.**

and furthermore:
> **At least one of the first 4 ciphertext bytes must not be one of
the ASCII hexadecimal character codes (a code for 0-9, A-F, or
a-f).** These restrictions can be satisfied by adjusting the initial
random plaintext bytes as necessary.

and finally:
> If the eexec-encrypted text is supplied in binary form, then every
byte is considered part of the ciphertext. If the eexec-encrypted
text is supplied in ASCII hexadecimal form, then ASCII white
space characters (blank, tab, carriage return and line feed) may be
freely interspersed within the encrypted text, **except in the first
eight characters.**

Hence the first 8 bytes of raw data should be the first 4 ciphertext bytes, for ASCII encoded data, as far as I'm reading it at least.

timvandermeij: Your last quote mentions eight ASCII characters and your first quote states that two ASCII characters are represented by one ciphertext byte, so eight ASCII characters correspond to four ciphertext bytes, which would match the remaining text.

Out of curiosity I looked at how for example PDFBox does it and they also seem to check only the first four bytes; see https://github.com/apache/pdfbox/blob/8b053ccddeb3a2722716085cf80ec7fdec43943f/fontbox/src/main/java/org/apache/fontbox/type1/Type1Parser.java#L881-L901. Interestingly, they check for whitespace in all four bytes, and that is also something I can read from your last quote: in the first eight characters (two characters per byte, so four bytes) no whitespace characters may exist.

Therefore, I *think* that the PDFBox implementation matches the specification most closely, but I think the specification is ambiguous here, so I do see how this can be interpreted in multiple ways :(

Snuffleupagus: > Out of curiosity I looked at how for example PDFBox does it and they also seem to check only the first four bytes

That does seem a bit odd to me, given the actual specification text...

> Interestingly, they check for whitespace in all four bytes,

That doesn't really seem to follow the specification as far as I can tell, given that it explicitly states:
> [...] then ASCII white space characters (...) may be freely interspersed within the encrypted text, **except** in the first eight characters.

All-in-all, given that the PR fixes an existing bug without any regressions elsewhere, is this good to go as is?

timvandermeij: > All-in-all, given that the PR fixes an existing bug without any regressions elsewhere

Well, I can't argue with that, and given the ambiguity I can't prove my view is correct either, so let's just go for it :-)

timvandermeij (approved):

timvandermeij:

/botio makeref

PR #11747, opened 2020-03-25
Add `passive: false` when removing wheel listeners

gdh1995:
Code of listening `wheel` event uses `{passive: false}`, while this argument will be treated as `true` before Firefox 49, according to https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Browser_compatibility .

This commit adds it when removing wheel listeners, so that such listeners can be really removed.

This is a simple complement to #11736.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1719,7 +1719,7 @@ const PDFViewerApplication = {
     const { _boundEvents } = this;
 
     window.removeEventListener("visibilitychange", webViewerVisibilityChange);
-    window.removeEventListener("wheel", webViewerWheel);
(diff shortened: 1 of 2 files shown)


timvandermeij:
From https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener: "It's worth noting that some browser releases have been inconsistent on this, and unless you have specific reasons otherwise, it's probably wise to use the same values used for the call to addEventListener() when calling removeEventListener()."

So this should be fine. Thank you!

timvandermeij (approved):

PR #11749, opened 2020-03-25
Support for array queries (allows to highlight "hello world" and "my name is Allice" in the same search query)

### LINKED ISSUE #7442: Search PDF for phrases and terms simultaneously

### LINKED ISSUE #11128: Highlighting multiple phrases at a time not working
#search=whale%20that%3dIshmael&phrase=true
I am trying to highlight multiple phrases at once. This should specifically highlight the phrase "whale that" and "Ishmael". However, it does not highlight these. 
[Etash.pdf](https://github.com/mozilla/pdf.js/files/3587087/Etash.pdf)
image
If I set phrase to false, I get this:
image

Configuration:
- chromium
- Any OS

### LINKED ISSUE #11748: [Question] Do not remove previous matches highlights
I'm trying to modify pdf.js to accept an array of terms in the "find" argument (i added a different one to maintain compatibility with original software). Those terms may be single or multi word, and if there's a term like "hello world" a hello not followed by world must not match.

I've done everything by now, and it works fine: the system recieves the json stringified list, creates an array and iterates over it executing the search function.

	for(i=0; i<complexSearch.length;i++){
            var term = complexSearch[i].trim();
        	this.eventBus.dispatch('findfromurlhash', {
        	      source: this,
        	      query: term,
        	      phraseSearch: true
        	  });
        }

The problem is that when it searches for the second term, it deletes all previous matches, third term deletes second, and o on, so at the end only the last term is highlishted.

**My question is: Where should I look to avoid this behaviour? I don't mind if it is a dirty fix like "it will never be able to remove highlights again", I wouldn't mind.**

I don't need a ready to implement answer either, if someone points me in the right direction It'd be enough to get started, I'm becoming crazy. I'm not a JS developer, so this kind of complexity is far above my head.

**Edit**:
As stated in the answers, I solved this by adding a _calculateArrayMatch.
I can update this to be compilant with the master and make a clean pull request with it, if someone is interested.

dgoiko:
Added support for array queries in order to allow complex searches like ["Hello world", "I'm Alice", "bye"] to combine phrase search with standard term search. Multi-word terms are searched just the way phrase search would work.

Included searchArray param (analogue to search) to not interfere with existing functionalities.

Arrays are provided as JSON stringified text via the new searchArray param (just as you would do with the existing search param, but it will use JSON.parse(params['searchArray ']) instead of params['searchArray '].replace(/"/g,'');

closes #11748 
closes #7442
closes  #11128

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -241,22 +241,26 @@ class PDFFindController {
    * @type {array} The (current) normalized search query array.
    */
   get _arrayQuery() {
-	if(!this._isArrayQuery()) return [this._query];
-	this._rawQuery = this._state.query;
-	this._normalizedQuery = [];
-	for(let i=0; i<this._state.query.length; i++) {
-		this._normalizedQuery.push(normalize(this._state.query[i]))
-	}
+    if(!this._isArrayQuery()) {
+      return [this._query];
+    }
+    this._rawQuery = this._state.query;
+    this._normalizedQuery = [];
+    for (let i = 0; i < this._state.query.length; i++) {
+      this._normalizedQuery.push(normalize(this._state.query[i]))
+    }
     return this._normalizedQuery;
   }
 
   _isArrayQuery() {
-	return Array.isArray(this._state.query);
+    return Array.isArray(this._state.query);
   }
 
   _isEmptyQuery() {
-	if(this._isArrayQuery()) return this._state.query.length == 0;
-	return this._query === "";
+    if (this._isArrayQuery()) {
+      return this._state.query.length === 0;
+    }
+    return this._query === "";
   }
 
   _shouldDirtyMatch(cmd, state) {
@@ -435,6 +439,7 @@ class PDFFindController {
       this._pageMatchesLength[pageIndex]
     );
   }
+
   _calculateMatch(pageIndex) {
     let pageContent = this._pageContents[pageIndex];
     let query;
@@ -448,7 +453,7 @@ class PDFFindController {
       query = this._arrayQuery;
       if (!caseSensitive) {
         pageContent = pageContent.toLowerCase();
-        for(i=0; i<query.length; i++) {
+        for (let i = 0; i < query.length; i++) {
           query[i] = query[i].toLowerCase();
         }
       }
@@ -466,8 +471,6 @@ class PDFFindController {
         this._calculateWordMatch(query, pageIndex, pageContent, entireWord);
       }
     }
-
-    
 
     // When `highlightAll` is set, ensure that the matches on previously
     // rendered (and still active) pages are correctly highlighted.
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -237,6 +237,28 @@ class PDFFindController {
     return this._normalizedQuery;
   }
 
+  /**
+   * @type {array} The (current) normalized search query array.
+   */
+  get _arrayQuery() {
+	if(!this._isArrayQuery()) return [this._query];
+	this._rawQuery = this._state.query;
+	this._normalizedQuery = [];
+	for(let i=0; i<this._state.query.length; i++) {
+		this._normalizedQuery.push(normalize(this._state.query[i]))
+	}
+    return this._normalizedQuery;
+  }
+
+  _isArrayQuery() {
+	return Array.isArray(this._state.query);
+  }
+
+  _isEmptyQuery() {
+	if(this._isArrayQuery()) return this._state.query.length == 0;
+	return this._query === "";
+  }
+
   _shouldDirtyMatch(cmd, state) {
     // When the search query changes, regardless of the actual search command
     // used, always re-calculate matches to avoid errors (fixes bug 1030622).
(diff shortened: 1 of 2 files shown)


dgoiko:
This could also be improved with a "delimiter" parameter, but I think arrays are more versatile (although urls are uglier with arrays)

Snuffleupagus:
Generally speaking we do *not* want to add any non-standard hash parameter in the PDF.js default viewer, see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/pdf_open_parameters.pdf; please also refer to https://github.com/mozilla/pdf.js/issues/7442#issuecomment-229696596.

dgoiko:
@Snuffleupagus first of all, thanks for such a fast reply and for contributing on this nice reader, it has saved my life a couple times by now :). Now to the issue:

After seeing the comment, I guess that it doesn't matter if I change the parameters, as maintainers don't pretend to extend search functionality. If that's the case,, no problem, I just hope this MR helps someone who needs this functionality, however, if you want this feature introduced, I can work a way to make it work with current parameters.

First idea:

- Using existing search parameter, if it starts with [ and ends with ] (note the absence of ") it is considered to be an array and if it starts with " it will be threated as a string.

This way, the viewer would not introduce any new parameter, will be totally backwards compatible with any previous implementations AND will provide the functionality without creating any new pdf parameter.

Tell me if this sounds OK and I'll work on it.

Snuffleupagus:
> Using existing search parameter, if it starts with [ and ends with ] (note the absence of ") it is considered to be an array and if it starts with " it will be threated as a string.

That would be a breaking change, since in the current implementation the use of quotation marks just so happens to be optional; note how the following links produce *identical* search results:
[http://mozilla.github.io/pdf.js/web/viewer.html#search="mozilla javascript"](http://mozilla.github.io/pdf.js/web/viewer.html#search="mozilla%20javascript")
[http://mozilla.github.io/pdf.js/web/viewer.html#search=mozilla javascript](http://mozilla.github.io/pdf.js/web/viewer.html#search=mozilla%20javascript)

> Removing everything from the parameters, but leaving the methods.

Well, that's exactly what's already mentioned in https://github.com/mozilla/pdf.js/issues/7442#issuecomment-229696596 as a possible approach :-)

For this PR to be accepted, there's likely a few things that need to be done:
 - Remove the new hash parameter.
 - Decide if the ability to simply pass an Array, rather than a String, as the `query` parameter is a good API, or if some other approach would be preferred. @timvandermeij Any opinions on this (and the patch in general)?
 - Add unit-tests for the new find functionality.
 - Properly review/test the new code (based on a cursory look, there may be some room for improvement).

dgoiko:
Okey, i'll work on it.

I'll start by removing the hash parameter and look into your existing unit tests to do it with a similar style, until someone gives a hint about the API approach.

There'll be for sure room for improvements in my code, as I'm not a frontend dev, so probably my JS code is far from perfection. I was thinking about including a _isSameArrayQuery param to avoid normalizing the array query over and over, just as the original code does with strings, however, String comparision is really cheap compared with iterating an array comparing (although it still looks cheaper than running the normalization regexps more than once)

timvandermeij:
I agree with https://github.com/mozilla/pdf.js/pull/11749#issuecomment-604469876 here. The hash parameter is definitely a no-go as far as I'm concerned, but extending the API would be okay given that unit tests are added.

I have a few questions about the use case here:

- Do other viewers allow the user to do such searches, and if so, how is the input given there? That may be help to decide upon a consistent API for this.
- Is the use case here to be able to quickly highlight multiple search terms? So if I search for `foo` and `bar baz`, would it highlight all occurences of `foo` (separately) and `bar baz` (combined)?

m1stermanager:
If anybody needs some assistance on this I wouldn't mind helping out. @timvandermeij to your two questions:
1. i would anticipate something like phraseSearch: false and query: `testing "test 123" query` to search on "testing", "test 123" and "query" separately.... grouping based on the presence of the double quotes obviously. 
2. yes that is the use case I have in mind (i have no particular ties to this PR other than just needing/wanting this functionality)

timvandermeij:
Closing since this PR is quite old/stale by now and there were concerns w.r.t. the implementation. We'll keep the issues open for further discussion.

valh1996:
Hi @timvandermeij,

The code being complex to assimilate, I would like to know if you could redirect me to the portion of code that causes the problem described below please?

I tried to implement this feature on my side on version `2.14.305` by revising it very slightly. The code is therefore very close to this PR with some adaptations since things have changed since this last version.

Now I can pass pass this kind of query:

```js
eventBus.dispatch('find', {
  type: '',
  highlightAll: true,
  query: [
    {
      query: 'nucleaire',
      matchDiacritics: false,
      entireWord: false,
      caseSensitive: false,
    },
    {
      query: 'valentin_TEST',
      matchDiacritics: true,
      entireWord: true,
      caseSensitive: true,
    }
  ]
});
```

The result is rather positive since the order of the queries appears in the same order in the PDF:
CleanShot 2022-07-27 at 21 45 37

My problem is that now, if I reverse the order of my `query` array that I pass to the `eventBus`, then only the last query is highlighted as if it overwrites the others:

```js
[
  {
    query: 'valentin_TEST', // now FIRST, but appears after `nucleaire` in PDF
    matchDiacritics: true,
    entireWord: true,
    caseSensitive: true,
  }, {
    query: 'nucleaire',
    matchDiacritics: false,
    entireWord: false,
    caseSensitive: false,
  },
]
```

Result:
CleanShot 2022-07-27 at 21 48 03

For information, I mainly modified the [#calculateRegExpMatch](https://github.com/mozilla/pdf.js/blob/master/web/pdf_find_controller.js#L559-L586) method to accept a query array.

NOTE: I tried to display the received matches in the `#renderMatches` method and I get my 6 corresponding matches. So I don't understand why my previous matches are deleted?

PR #11755, opened 2020-03-26
Remove the unused `sizes` and `encoding` properties on `Font` instances

Snuffleupagus:
The `sizes` property doesn't appear to have been used ever since the code was first split into main/worker-threads, which is so many years ago that I wasn't able to easily find exactly in which PR/commit it became unused.

The `encoding` property is always assigned the `properties.baseEncoding` value, however the `PartialEvaluator` doesn't actually compute/set that value any more. Again it was difficult to determine when it became unused, but it's been that way for years.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -523,7 +523,6 @@ var Font = (function FontClosure() {
     this.name = name;
     this.loadedName = properties.loadedName;
     this.isType3Font = properties.isType3Font;
-    this.sizes = [];
     this.missingFile = false;
 
     this.glyphCache = Object.create(null);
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
I've started looking at https://github.com/mozilla/pdf.js/blob/fa4b431091c51a82315dce7da7b848cc6498bcea/src/core/fonts.js#L1264-L1265 and have realized that there's unfortunately *a large number* of apparently unused properties being sent to the main-thread. However, before diving into all of this I wanted to at least remove a few properties that are guaranteed to be unused; hence this PR.

timvandermeij (approved):

timvandermeij:

PR #11757, opened 2020-03-29
Fix most failures, in the `src/` folder, reported by the ESLint `no-var` rule

Snuffleupagus:
Previously we've been selectively enabling this rule on a file-by-file basis, with the `web/` folder already done, but it seems like a good idea to also enable it in `src/` folder.[1]

This patch contains all possible automatic fixes, as done by the ESLint `--fix` argument, but unfortunately there's a fair number of cases that can only be fixed manually.[2]
Since there's still ESLint failures, with this patch, the `no-var` rule is thus not enabled for the `src/` folder just yet.

With the introduction of Prettier this sort of mass enabling of ESLint rules also becomes a lot easier, since the code will be automatically reformatted as necessary to account for e.g. changed line lengths.

---
[1] Most likely we don't want to enable this rule in all files/folders, e.g. the `examples/` folder probably needs to be left alone for the time being.

[2] Particularly the failures involving `switch`/`case` will require some care, note https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#Block-scope_variables_within_switch_statements

### DIFF
--- a/src/.eslintrc
+++ b/src/.eslintrc
@@ -1,0 +1,10 @@
+{
+  "extends": [
+    "../.eslintrc"
+  ],
+
+  "rules": {
+    // ECMAScript 6
+    "no-var": "off",
+  },
+}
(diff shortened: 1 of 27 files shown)


Snuffleupagus:
Based on a couple of `gulp test`, and in particular `gulp browsertest`, runs locally it seems that this unfortunately causes *completely unacceptable* regressions; the run-time increases by approx. 5% compared to `master`.

Hence it's pretty clear to me that we should *definitely not* make these types of changes, at least not until performance of `let`/`const` improves in Firefox.

PR #11765, opened 2020-03-30
[api-minor] Change `Font.exportData` to use an explicit white-list of exportable properties

Snuffleupagus:
This patch addresses an existing, and very long standing, TODO in the code such that it's no longer possible to send arbitrary/unnecessary font properties to the main-thread.
Furthermore, by having a white-list it's also very easy to see *exactly* which font properties are being exported.

Please note that in its current form, the list of exported properties contains *every* possible enumerable property that may exist in a `Font` instance.
In practice no single font will contain *all* of these properties, and e.g. embedded/non-embedded/Type3 fonts will all differ slightly with respect to what properties are being defined. Hence why only explicitly set properties are included in the exported data, to avoid half of them being `undefined`, which however should not be a problem for any existing consumer.

Since a fair number of these font properties are completely *internal* functionality, and doesn't make any sense on the main-thread, follow-up patch(es) will be required to trim down the list. (I purposely included all properties here for brevity and future documentation purposes.)

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1761,7 +1761,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
             fontFamily: font.fallbackName,
             ascent: font.ascent,
             descent: font.descent,
-            vertical: !!font.vertical,
+            vertical: font.vertical,
           };
         }
         textContentItem.fontName = font.loadedName;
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
I'm going to split this up a bit, since some of the planned changes will be slightly more involved...

PR #11767, opened 2020-03-30
Replace the RTL images with CSS transforms of the standard images (issue 11766)

### LINKED ISSUE #11766: Use CSS transform instead of including RTL equivalents of images
https://github.com/mozilla/pdf.js/blob/4ff67a518e676a2520c34a11e2c635aac6195681/web/viewer.css#L776-L816

We should just use `transform: scaleX(-1)` instead of RTL equivalents of images.

Those RTL images also cause duplicate files when packaging Firefox: https://searchfox.org/mozilla-central/rev/fa2df28a49883612bd7af4dacd80cdfedcccd2f6/browser/installer/allowed-dupes.mn#66-73

Snuffleupagus:
This avoids unnecessary duplication of many images, thus reducing the size of PDF.js image resources slightly.

Note that since the images should only be flipped horizontally, this required specifying the horizontal/vertical scaling separately for the hiDPI-images.

Fixes #11766


timvandermeij (approved):

timvandermeij:
Nice work! Aside from a reduction in image resources, this also makes the viewer easier to restyle, both for us and for third-party usage.

Issue #11768, opened 2020-03-31
PDF Render Error


coldbess:
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version: FireFox 74.0
- Operating system and its version: MacOS 10.14.6
- PDF.js version: 2.3.200
- Is a browser extension: No

Steps to reproduce the problem:
1. Use [sample.pdf](https://github.com/mozilla/pdf.js/files/4406710/sample.pdf)

2. upload to https://mozilla.github.io/pdf.js/web/viewer.html

What is the expected behavior? (add screenshot)
<img width="259" alt="sample" src="https://user-images.githubusercontent.com/5848499/77985935-5fc7f180-7348-11ea-8760-850eb5d1c84b.png">
What went wrong? (add screenshot)
<img width="239" alt="error" src="https://user-images.githubusercontent.com/5848499/77986052-b9c8b700-7348-11ea-8db7-cd0b3d315ad6.png">

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): No

timvandermeij:
I can reproduce this using Arch Linux and Firefox. The following messages show up in the console:

```
Warning: getOperatorList - ignoring errors during "GetOperatorList: page 0" task: "RangeError: 14737632 is not a valid code point".
Warning: getTextContent - ignoring errors during "GetTextContent: page 0" task: "RangeError: 14737632 is not a valid code point".
```

coldbess:
This link is OK
http://54.67.70.0:8877/4d1091cde93c899/web/viewer.html
Thanks a lot

PR #11769, opened 2020-03-31
Ensure that `Font.charToGlyph` won't fail because `String.fromCodePoint` is given an invalid code point (issue 11768)

Snuffleupagus:
*Please note:* This patch on its own is *not* sufficient to address the underlying problem in the referenced issue, hence why no test-case is included since the *actual* bug still needs to be fixed.

As can be seen in the specification, https://tc39.es/ecma262/#sec-string.fromcodepoint, `String.fromCodePoint` will throw a RangeError for invalid code points.

In the event that a CMap, in a composite font, contains invalid data and/or we fail to parse it correctly, it's thus possible that the glyph mapping that we build end up with entires that cause `String.fromCodePoint` to throw and thus `Font.charToGlyph` to break.
If that happens, as is the case in issue #11768, significant portions of a page/document may fail to render which seems very unfortunate.

While this patch doesn't fix the underlying problem, it's hopefully deemed useful not only for the referenced issue but also to prevent similar bugs in the future.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -3216,10 +3216,14 @@ var Font = (function FontClosure() {
         };
       }
 
-      var fontChar =
-        typeof fontCharCode === "number"
-          ? String.fromCodePoint(fontCharCode)
-          : "";
+      let fontChar = "";
+      if (typeof fontCharCode === "number") {
+        if (fontCharCode <= 0x10ffff) {
+          fontChar = String.fromCodePoint(fontCharCode);
+        } else {
+          warn(`charToGlyph - invalid fontCharCode: ${fontCharCode}`);
+        }
+      }
 
       var glyph = this.glyphCache[charcode];
       if (


src/core/fonts.js
```diff
@@ -3216,10 +3216,14 @@ var Font = (function FontClosure() {
         };
       }
 
-      var fontChar =
-        typeof fontCharCode === "number"
-          ? String.fromCodePoint(fontCharCode)
-          : "";
+      let fontChar = "";
```
timvandermeij, line 3219: Is this method called often? If it's often, it may be beneficial to leave it undefined here and only set it to the empty string in the exception cases to avoid unnecessary string allocation in the common case when the code point is valid.

src/core/fonts.js
```diff
@@ -3216,10 +3216,14 @@ var Font = (function FontClosure() {
...
+      if (typeof fontCharCode === "number") {
+        if (fontCharCode < 0x110000) {
```
timvandermeij, line 3221: Would it perhaps be better to use `< 0 or > 0x10FFFF` just like in the specification, so also negative numbers are handled here?

src/core/fonts.js
Snuffleupagus: > Is this method called often?

Yes, but comparing the run-time of the tests here to other PRs there really doesn't seem to be any noticeable slowdowns overall (I *think* browsers may be able to optimize/re-use strings internally).

> and only set it to the empty string in the exception cases

While I did consider that, it's not clear to me that it's entirely necessary here (also the code would become a bit "messy" in that case IMHO).
Also, you're still having to allocate the variable (albeit implicitly to `undefined`) which is probably why it doesn't seem to matter in practice.

src/core/fonts.js
Snuffleupagus: > Would it perhaps be better to check for `< 0 or > 0x10FFFF` just like in the specification,

I thought about the `< 0` case, but honestly it seems to me that that would indicate *much larger* underlying problems in the code to the point that I'd not mind a breaking error (also considering that the `String.fromCodePoint` changes landed ~1.5 years ago and this is the first time they've caused a problem).

> can we at least change this to `<= 0x10FFFF` so it's easier to match with the spec?

If you think the current formatting is too confusing, maybe because of the amount of digits involved, I suppose that that can be changed :-)

timvandermeij (approved):

timvandermeij:
Thanks!

PR #11771, opened 2020-04-01
Fail early, in modern `GENERIC` builds, if certain required browser functionality is missing (issue 11762)

### LINKED ISSUE #11762: Node support broken in 2.4.456 due to missing ReadableStream
Node.js 13.10.1

Consider a simple test program:
```
const pdf = require('pdfjs-dist/build/pdf.js');
const pdfPath = "test.pdf";

(async function main() {
    const doc = await pdf.getDocument(pdfPath).promise;
    const nPages = doc.numPages;
    for (let i = 1; i <= nPages; i++) {
        const page = await doc.getPage(i);
        let annots = await page.getAnnotations();
        console.log(annots);
    }
})();
```
I doesn't matter what `test.pdf` is, this should repro with any valid pdf.

With 2.3.200 (i.e. current npm release), it works:
```
PS C:\Users\kjk\src\tmppdf> yarn add pdfjs-dist
success Saved 1 new dependency.
info Direct dependencies
└─ pdfjs-dist@2.3.200

PS C:\Users\kjk\src\tmppdf> node test.js
[]
```
You can see the program works and prints an empty list of annotations.

With 2.4.456 (which is current @ next):
```
PS C:\Users\kjk\src\tmppdf> yarn add pdfjs-dist@next
success Saved 1 new dependency.
info Direct dependencies
└─ pdfjs-dist@2.4.456

PS C:\Users\kjk\src\tmppdf> node test.js
(node:28516) UnhandledPromiseRejectionWarning: ReferenceError: ReadableStream is not defined
    at MessageHandler.sendWithStream (C:\Users\kjk\src\tmppdf\node_modules\pdfjs-dist\build\pdf.worker.js:45460:5)
    at new PDFWorkerStreamReader (C:\Users\kjk\src\tmppdf\node_modules\pdfjs-dist\build\pdf.worker.js:45833:45)
    at PDFWorkerStream.getFullReader (C:\Users\kjk\src\tmppdf\node_modules\pdfjs-dist\build\pdf.worker.js:45797:31)
    at getPdfManager (C:\Users\kjk\src\tmppdf\node_modules\pdfjs-dist\build\pdf.worker.js:306:35)
    at setupDoc (C:\Users\kjk\src\tmppdf\node_modules\pdfjs-dist\build\pdf.worker.js:457:7)
    at wphReady (C:\Users\kjk\src\tmppdf\node_modules\pdfjs-dist\build\pdf.worker.js:646:7)
    at LoopbackPort.MessageHandler._onComObjOnMessage (C:\Users\kjk\src\tmppdf\node_modules\pdfjs-dist\build\pdf.worker.js:45410:7)
    at C:\Users\kjk\src\tmppdf\node_modules\pdfjs-dist\build\pdf.js:2062:18
    at Array.forEach (<anonymous>)
    at C:\Users\kjk\src\tmppdf\node_modules\pdfjs-dist\build\pdf.js:2061:23
(node:28516) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:28516) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

You can see this fails because it tries to create a `ReadableStream` object, which is a browser-only API.

There is a `ReadableStream` polyfill created by `checkReadableStream` in `compatibility.js` but it appears it isn't installed.

This could be packaging problem.

Or it could be that this assumption is invalid:
```
(function checkReadableStream() {
  if (typeof PDFJSDev !== 'undefined' && PDFJSDev.test('IMAGE_DECODERS')) {
    // The current image decoders are synchronous, hence `ReadableStream`
    // shouldn't need to be polyfilled for the IMAGE_DECODERS build target.
    return;
  }
```
i.e. `ReadableStream` is always needed because it's used by `sendWithStream`.

Most likely the checkin that introduced this is https://github.com/mozilla/pdf.js/commit/e24050fa137b6696dc8b5b38ef820fdca7dcad05

It changed `message_handler.js` to rely on (apparently missing) global polyfill of `ReadableStream` instead of importing it directly.

As a side note: if this gets fixed, it would be good to add an automated test that uses the pdfjs-dist to catch this.

Also, when `ReadableStream` became a global object, this check might need to be updated:
```
function isFetchSupported() {
  return (
    typeof fetch !== "undefined" &&
    typeof Response !== "undefined" &&
    "body" in Response.prototype &&
    typeof ReadableStream !== "undefined"
  );
}
```
(because now `typeof ReadableStream !== "undefined"` should always be `true`).

Snuffleupagus:
With two kind of builds now being produced, with/without translation/polyfills, it's unfortunately somewhat easy for users to accidentally pick the wrong one.

In the case where a user would attempt to use a modern build of PDF.js in an older browser, such as e.g. IE11, the failure would be immediate when the code is loaded (given the use of unsupported ECMAScript features).
However in some browsers/environments, in particular Node.js, a modern PDF.js build may load correctly and thus *appear* to function, only to fail for e.g. certain API calls. To hopefully lessen the support burden, and to try and improve things overall, this patch adds checks to ensure that a modern build of PDF.js cannot be used in browsers/environments which lack native support for critical functionality (such as e.g. `ReadableStream`). Hence we'll fail early, with an error message telling users to pick an ES5-compatible build instead.

To ensure that we actually test things better especially w.r.t. usage of the PDF.js library in Node.js environments, the `gulp npm-test` task as used by Node.js/Travis was changed (back) to test an ES5-compatible build.
(Since the bots still test the code as-is, without transpilation/polyfills, this shouldn't really be a problem as far as I can tell.)
As part of these changes there's now both `gulp lib` and `gulp lib-es5` build targets, similar to e.g. the generic builds, which thanks to some re-factoring only required adding a small amount of code.

*Please note:* While it's probably too early to tell if this will be a widespread issue, it's possible that this is the sort of patch that *may* warrant being `git cherry-pick`ed onto the current beta version (v2.4.456).

Fixes #11762

*Much smaller/easier diff with https://github.com/mozilla/pdf.js/pull/11771/files?w=1*

### DIFF
--- a/external/dist/lib/README.md
+++ b/external/dist/lib/README.md
@@ -3,5 +3,5 @@ pre-built library as found in e.g. the `/build`, `/web`, and `/image_decoders`
 folders in the root of this repository.
 
 Please note that the "lib" build target exists mostly to enable unit-testing in
-Node.js/Travis, and that you'll need to handle e.g. any Node.js dependencies
-yourself if using the files in this folder.
+Node.js/Travis, and that you'll need to handle e.g. any necessary polyfills
+and/or Node.js dependencies yourself if using the files in this folder.
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1150,92 +1150,105 @@ gulp.task("jsdoc", function(done) {
   });
 });
 
+function buildLib(defines, dir) {
+  // When we create a bundle, webpack is run on the source and it will replace
+  // require with __webpack_require__. When we want to use the real require,
+  // __non_webpack_require__ has to be used.
+  // In this target, we don't create a bundle, so we have to replace the
+  // occurences of __non_webpack_require__ ourselves.
+  function babelPluginReplaceNonWebPackRequire(babel) {
+    return {
+      visitor: {
+        Identifier(curPath, state) {
+          if (curPath.node.name === "__non_webpack_require__") {
+            curPath.replaceWith(babel.types.identifier("require"));
+          }
+        },
+      },
+    };
+  }
+  function preprocess(content) {
+    var skipBabel =
+      bundleDefines.SKIP_BABEL || /\/\*\s*no-babel-preset\s*\*\//.test(content);
+    content = preprocessor2.preprocessPDFJSCode(ctx, content);
+    content = babel.transform(content, {
+      sourceType: "module",
+      presets: skipBabel ? undefined : ["@babel/preset-env"],
+      plugins: [
+        "@babel/plugin-transform-modules-commonjs",
+        [
+          "@babel/plugin-transform-runtime",
+          {
+            helpers: false,
+            regenerator: true,
+          },
+        ],
+        babelPluginReplaceNonWebPackRequire,
+      ],
+    }).code;
+    var removeCjsSrc = /^(var\s+\w+\s*=\s*(_interopRequireDefault\()?require\(".*?)(?:\/src)(\/[^"]*"\)\)?;)$/gm;
+    content = content.replace(removeCjsSrc, (all, prefix, interop, suffix) => {
+      return prefix + suffix;
+    });
+    return licenseHeaderLibre + content;
+  }
+  var babel = require("@babel/core");
+  var versionInfo = getVersionJSON();
+  var bundleDefines = builder.merge(defines, {
+    BUNDLE_VERSION: versionInfo.version,
+    BUNDLE_BUILD: versionInfo.commit,
+    TESTING: process.env["TESTING"] === "true",
+  });
+  var ctx = {
+    rootPath: __dirname,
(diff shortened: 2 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Looks like a good change to me; thanks! Note that the Travis CI build did run and pass, but for some reason it's not visible on GitHub.

PR #11773, opened 2020-04-02
[api-minor] Change `Font.exportData` to use an explicit white-list of exportable properties, and stop exporting internal/unused properties

Snuffleupagus:
*Please refer to the individual commit messages for addition details.*

---

Note that there's *still* a number of potentially very large data-structures, containing e.g. Arrays and Objects, being exported despite being completely unused on the main-thread and/or in the API.
However, since we cannot be sure that they're  completely unused in all custom implementations of the PDF.js library, we probably cannot just remove them without also providing users with a way to still get this information (that we've been *accidentally* exposing for years).

To not unnecessarily delay the changes in this PR, which on their own should already be beneficial in my opinion, I thus decided to post-phone any further clean-up to a separate PR.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -87,6 +87,49 @@ var PDF_GLYPH_SPACE_UNITS = 1000;
 // custom one. Windows just refuses to draw glyphs with seac operators.
 var SEAC_ANALYSIS_ENABLED = true;
 
+const EXPORT_DATA_PROPERTIES = [
+  "_shadowWidth",
+  "ascent",
+  "bbox",
+  "black",
+  "bold",
+  "cMap",
+  "charProcOperatorList",
+  "charsCache",
+  "cidEncoding",
+  "composite",
+  "data",
+  "defaultEncoding",
+  "defaultVMetrics",
+  "defaultWidth",
+  "descent",
+  "differences",
+  "fallbackName",
+  "fallbackToUnicode",
+  "fontMatrix",
+  "fontType",
+  "glyphCache",
+  "isMonospace",
+  "isOpenType",
+  "isSerifFont",
+  "isSymbolicFont",
+  "isType3Font",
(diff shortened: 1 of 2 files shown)


timvandermeij:
The bot should be fixed again.

/botio-windows test

Snuffleupagus:
It seems that the Windows bot have decided to take an extended vacation :-)

Anyway, given that the tests pass on the Linux bot (modulo the Chrome-intermittent ones) and that all tests *obviously* also pass locally on Windows for me; do we really need to wait for the Windows bot to be re-started again?
If this patch was fundamentally broken in some way, I'd have expected lots of errors everywhere...

src/core/fonts.js
```diff
@@ -1258,12 +1301,14 @@ var Font = (function FontClosure() {
...
 
-    exportData: function Font_exportData() {
-      // TODO remove enumerating of the properties, e.g. hardcode exact names.
-      var data = {};
-      for (var i in this) {
-        if (this.hasOwnProperty(i)) {
-          data[i] = this[i];
+    exportData() {
+      const data = Object.create(null);
+      let property, value;
```
timvandermeij, line 1300: I think we can remove this line if we use `for (const property ...` and `const value = this...` below.

timvandermeij:
Looks good after the comment above is addressed.

src/core/fonts.js
Snuffleupagus: Actually, that seems *slightly* slower based on bot run-time hence I'm reverting this change (besides those variables will get reused with the follow-up PR anyway).

timvandermeij (approved):

timvandermeij:
Nice improvements!

Snuffleupagus:
> Nice improvements!

Thanks for the review; the real improvements are PR #11777 though :-)

PR #11777, opened 2020-04-03
[api-minor] Change `Font.exportData` to, by default, stop exporting properties which are completely unused on the main-thread and/or in the API (PR 11773 follow-up)

Snuffleupagus:
*Smaller diff with https://github.com/mozilla/pdf.js/pull/11777/files?w=1*

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -92,21 +92,17 @@ const EXPORT_DATA_PROPERTIES = [
   "bbox",
   "black",
   "bold",
-  "cMap",
   "charProcOperatorList",
   "composite",
   "data",
-  "defaultEncoding",
   "defaultVMetrics",
   "defaultWidth",
   "descent",
-  "differences",
   "fallbackName",
   "fontMatrix",
   "fontType",
   "isMonospace",
   "isSerifFont",
-  "isSymbolicFont",
   "isType3Font",
   "italic",
   "loadedName",
@@ -114,12 +110,19 @@ const EXPORT_DATA_PROPERTIES = [
   "missingFile",
   "name",
   "remeasure",
-  "seacMap",
   "subtype",
-  "toFontChar",
-  "toUnicode",
   "type",
   "vertical",
+];
+
+const EXPORT_DATA_EXTRA_PROPERTIES = [
+  "cMap",
+  "defaultEncoding",
+  "differences",
+  "isSymbolicFont",
+  "seacMap",
+  "toFontChar",
+  "toUnicode",
   "vmetrics",
   "widths",
 ];
@@ -1295,10 +1298,14 @@ var Font = (function FontClosure() {
       return shadow(this, "renderer", renderer);
     },
 
-    exportData() {
+    exportData(extraProperties = false) {
+      const exportDataProperties = extraProperties
+        ? [...EXPORT_DATA_PROPERTIES, ...EXPORT_DATA_EXTRA_PROPERTIES]
+        : EXPORT_DATA_PROPERTIES;
+
       const data = Object.create(null);
       let property, value;
-      for (property of EXPORT_DATA_PROPERTIES) {
+      for (property of exportDataProperties) {
         value = this[property];
         // Ignore properties that haven't been explicitly set.
         if (value !== undefined) {
@@ -3352,7 +3359,7 @@ var ErrorFont = (function ErrorFontClosure() {
     charsToGlyphs: function ErrorFont_charsToGlyphs() {
       return [];
     },
-    exportData: function ErrorFont_exportData() {
+    exportData(extraProperties = false) {
       return { error: this.error };
     },
   };
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -145,6 +145,11 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *   converted to OpenType fonts and loaded via font face rules. If disabled,
  *   fonts will be rendered using a built-in font renderer that constructs the
  *   glyphs with primitive path commands. The default value is `false`.
+ * @property {boolean} [fontExtraProperties] - Include additional properties,
+ *   which are unused during rendering of PDF documents, when exporting the
+ *   parsed font data from the worker-thread. This may be useful for debugging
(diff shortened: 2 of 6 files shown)


src/core/fonts.js
```diff
@@ -1305,6 +1308,15 @@ var Font = (function FontClosure() {
           data[property] = value;
         }
       }
+      if (extraProperties) {
```
timvandermeij, line 1311: Instead of repeating this loop, would it be possible to concatenate `EXPORT_DATA_PROPERTIES` with `EXPORT_DATA_EXTRA_PROPERTIES` above in case `extraProperties = true` instead?

src/display/api.js
```diff
@@ -145,6 +145,11 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *   converted to OpenType fonts and loaded via font face rules. If disabled,
  *   fonts will be rendered using a built-in font renderer that constructs the
  *   glyphs with primitive path commands. The default value is `false`.
+ * @property {boolean} [fontExtraProperties] - Include additional properties,
+ *   which is unused during rendering of PDF documents, when exporting the
```
timvandermeij, line 149: `which are unused`

src/display/api.js
```diff
@@ -145,6 +145,11 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
...
+ *   parsed font data from the worker-thread. This may be useful for custom
```
timvandermeij, line 150: I'm wondering: is the objective to keep support this or is it only meant as a temporary solution so we can check if someone is actually using this and providing a way for them to get the data?

If we want to get rid of this eventually, I'm not sure about the `may be useful for custom implementations` part. Given that these properties were never meant to be exposed, and are of very limited use anyway if someone is actually using them, it feels like we should emphasize here that this option is only provided to for temporary backwards compatibility and that it may be removed in new releases. Otherwise it may read as if we're advertising the usage for third-party implementations.

src/display/api.js
```diff
@@ -251,7 +256,20 @@ function getDocument(src) {
   params.rangeChunkSize = params.rangeChunkSize || DEFAULT_RANGE_CHUNK_SIZE;
   params.CMapReaderFactory = params.CMapReaderFactory || DOMCMapReaderFactory;
   params.ignoreErrors = params.stopAtErrors !== true;
-  params.pdfBug = params.pdfBug === true;
+
+  const pdfBugEnabled = params.pdfBug === true;
+  params.pdfBug = pdfBugEnabled;
+  // To simplify debugging, always force-enable `fontExtraProperties` when
```
timvandermeij, line 262: Given the limited usefulness outlined in the commit messages, does the data actually provide helpful insights when debugging font problems? I'm wondering why we would want this in debug mode.

src/core/fonts.js
Snuffleupagus: It's obviously possible to do that, but I wanted to avoid doing extra work/allocations in the "normal" case. The only solution satisfying all these criteria, which I'm able to come up with, is perhaps slightly ugly; please see following the diff:
```diff
diff --git a/src/core/fonts.js b/src/core/fonts.js
index f5b6e09c..241e50da 100644
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -1299,24 +1299,19 @@ var Font = (function FontClosure() {
     },
 
     exportData(extraProperties = false) {
+      const exportDataProperties = extraProperties
+        ? [...EXPORT_DATA_PROPERTIES, ...EXPORT_DATA_EXTRA_PROPERTIES]
+        : EXPORT_DATA_PROPERTIES;
+
       const data = Object.create(null);
       let property, value;
-      for (property of EXPORT_DATA_PROPERTIES) {
+      for (property of exportDataProperties) {
         value = this[property];
         // Ignore properties that haven't been explicitly set.
         if (value !== undefined) {
           data[property] = value;
         }
       }
-      if (extraProperties) {
-        for (property of EXPORT_DATA_EXTRA_PROPERTIES) {
-          value = this[property];
-          // Ignore properties that haven't been explicitly set.
-          if (value !== undefined) {
-            data[property] = value;
-          }
-        }
-      }
       return data;
     },
```
Do you think that would be a better approach overall?

src/display/api.js
Snuffleupagus: Fixed, thanks!

src/display/api.js
Snuffleupagus: *Please refer to the comment above.*

src/display/api.js
Snuffleupagus: > I'm wondering: is the objective to keep support this or is it only meant as a temporary solution so we can check if someone is actually using this and providing a way for them to get the data?

Good question; I'm not really sure if/when we even can/want to eventually get rid of this.

Note that with debugging enabled in the viewer, e.g. with `#pdfBug=all`, it's currently possible to log all of these properties on the main-thread which occasionally does come in handy while debugging font issues.

> If we want to get rid of this eventually, I'm not sure about the `may be useful for custom implementations` part.
> [...]
> Otherwise it may read as if we're advertising the usage for third-party implementations.

You're right, let's at least remove that part to not unnecessarily promoting it.

src/display/api.js
Snuffleupagus: Although I suppose that we could remove the special-casing here, if you think that this code is unnecessarily confusing, and simply set `fontExtraProperties` in https://github.com/mozilla/pdf.js/blob/0400109b8763e53cf657d72fbf1bdab2df9927b9/web/app.js#L310-L314 instead?

Snuffleupagus:
There's one additional thing that could reduce memory usage of fonts *even further*, by removing the `font.data` property after the font has been attached to the document by `FontLoader.bind`; i.e. before inserting the font into `this.commonObjs` in https://github.com/mozilla/pdf.js/blob/0400109b8763e53cf657d72fbf1bdab2df9927b9/src/display/api.js#L2239-L2252

Unfortunately that'll probably not work with the SVG back-end though, which is why I'm not attempting this here (maybe a good follow-up PR assuming it can be made to work generally).

timvandermeij (approved):

timvandermeij:
Let's do this; thanks!

PR #11781, opened 2020-04-05
Update the "gulp jsdoc" task to account for API changes in the `mkdirp` package (PR 11772 follow-up)

Snuffleupagus:
I completely overlooked the fact that we had *one* occurrence of an asynchronous `mkdirp` call in the gulpfile, which thus breaks since the package now uses Promises rather than a callback function; sorry about that!


timvandermeij (approved):

timvandermeij:
Thank you for the follow-up!

Issue #9474, opened 2018-02-12
Pdf scanned at 600dpi does not load


kavisherlock:
Attach (recommended) or Link to PDF file here: 
[pdf600dpi.pdf](https://github.com/mozilla/pdf.js/files/1717166/pdf600dpi.pdf)

Configuration:
- Web browser and its version: Safari, Chrome (latest)
- Operating system and its version: iOS 11.2 (on iPad and iPhone)
- PDF.js version: 1.10.97 (or whatever's on https://mozilla.github.io/pdf.js/web/viewer.html)
- Is a browser extension: No

Steps to reproduce the problem:
1. Load the pdf.js demo page on an iPad/iPhone (https://mozilla.github.io/pdf.js/web/viewer.html)
2. Upload the above pdf (or any other pdf scanned at 600 dpi)

What is the expected behaviour? (add screenshot)
Pdf loads (the screenshot is from desktop browser)
<img width="1679" alt="screen shot 2018-02-12 at 11 25 55 am" src="https://user-images.githubusercontent.com/23285466/36110242-8e85f0ec-0fe7-11e8-931e-37c9ff5630ff.png">

What went wrong? (add screenshot)
Pdf does not load and gives the following: warning in the console "Canvas area exceeds the maximum limit (width * height > 16777216)."
The issue is because the pdf was scanned at 600 dpi from an A4-size page, making the canvas area 600 * 8.5 * 600 * 11, which is indeed greater than 16777216. This works properly on my desktop browsers, but not on iPad or iPhone.
iPad:
![img_0252](https://user-images.githubusercontent.com/23285466/36110019-e0a74bce-0fe6-11e8-84ee-86cbc8611032.PNG)

iPhone:
![img_0641](https://user-images.githubusercontent.com/23285466/36110020-e0b9e4be-0fe6-11e8-84fa-a47c7ebf4512.JPG)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): https://mozilla.github.io/pdf.js/web/viewer.html

uncleramsay:
Hi, is there any update on this issue? I'm facing the same problem.

I see that a PR was opened, but was never merged - is there any guidance on how we can deal with this issue?

Thanks

lorkel:
Also facing this issue - is there any way to set the maximum canvas size?

KeiBontakun:
I got the same issue with current pdf.js demo page.
[...]

Issue #6741, opened 2015-12-09
PDF with huge embedded image is displayed as a blank page


Rob--W:
PDF: [DC000933-uncompressed.pdf](https://github.com/mozilla/pdf.js/files/57455/DC000933-uncompressed.pdf)

If I open the above PDF in Chrome or Firefox, it is shown as blank at best. Firefox sometimes crashes if I (re)load a couple of times (even at a fresh start-up). I haven't built Firefox with symbols, but GDB shows that the stack trace ends in `mozalloc_abort`, which suggets that this segfault is caused by a memory allocation error.

This PDF is quite unusual, it has a width and height with excessive values:

```
<< /Contents 10 0 R /MediaBox [ 0 0 4444 2592 ] /Parent 6 0 R /Resources << /ExtGState << /G0 11 0 R /G1 12 0 R /G2 13 0 R >> /ProcSet [ /ImageB /ImageC /ImageI ] /XObject << /I0 14 0 R >> >> /Type /Page >>

...

<< /BitsPerComponent 1 /ColorSpace /DeviceGray /DecodeParms << /Columns 24688 /K -1 /Rows 14400 >> /Filter /CCITTFaxDecode /Height 14400 /Name /I0 /Subtype /Image /Type /XObject /Width 24688 /Length 972551 >>
```

It was allegedly created by MetaPrint.

timvandermeij:
Yes, we have seen such PDF files before with images with a huge width and height. We might need to implement something to recognize that and downscale the image.

AnasAl-tamemy:
i think it should return an error so that developers can handle the open pdf failure.

gitarra:
These files render fine at least with Evince and PDFium.

redlum94:
Is there already any news on this issue? We run into the same issue with half of our PDFs. As they consist of scanned building plans. Or is there a known cause of this problem?

Rob--W:
@redlum94 Scanned building plans sound like large images (especially with a high resolution). In the first comment, @timvandermeij already suggested that we should downscale the image when needed.

I have worked on image downscaling before in PR #6606 , but that patch has bitrotten.

dylanwoo:
Does anyone know if there is a global variable I can look at to see if the viewer is empty or not?

peterkmurphy:
I am another user affected by the bug. My tool also works with architectural diagrams, where pages sizes can be A1 or greater in size.

[...]

PR #11783, opened 2020-04-06
IOS 600 DPI Fix - #9474

ghost:
This fix is linked to the following issue: #9474.

Though there is another PR for this, that has been closed, https://github.com/mozilla/pdf.js/pull/9645, the current PR has been revisited and has the current pdfjs merged in.

For the RGB images, the idea is to resize the image by reducing each scale * scale box of pixels to one pixel. We are parsing the image with a square box of scale * scale and for each iteration we are calculating an average of the pixels color inside the box and create one pixel with the average value. The algorithm works perfect for images that have the height and width multiple of scale. To address images that do not fit in that category we have applied the algorithm separate to the main area of the image (the area that is divisible by scale) and to the side areas, the right side, the bottom side and the bottom right.
For the B&W images the idea is to resize the image by increasing the image depth. From a B& W image we are going to create a Gray-scale image.
In a B&W image each bit represent a pixel. We have created an algorithm to scale by 2 and another one to scale by 4.
To scale the image by 2 we've grouped each 2 consecutive bits from the current line and the next line (line + 1) in a square of 4 bits. Then we've created a new byte (pixel) by multiplying the number of bits set to 1 (white) with 256 (white value on a gray-scale image) and divided the result by 4 (because we have considered 4 pixels).
To scale the image by 4 we've grouped each 4 consecutive bits from the current line with 4 consecutive bits from the next 3 lines below in a square of 16 bits. Then we've created a new byte (pixel) by multiplying the number of bits set to 1 with 256 and divided the result by 16 (because we've considered 16 pixels).
The problem for the B&W scaling algorithms is when the image width is not divisible by 8 and/or image height is not divisible by scale. For those cases we've created a separate scaling algorithm which we will call for each side areas (right side, bottom side and bottom right).
While an RGB image can be scaled by any factor, the B&W images can only be scaled by 2 or by 4. However a scale by 2 means the new image size will be 4 times smaller, and a scale by 4 will create a new image with a size of 16 times smaller. We chose to do the B&W scaling only by 2 and 4 because of performance.
Below are the PDFJS files that I've changed:
• src/core/image.js
• src/display/canvas.js
• src/shared/util.js

### DIFF
--- a/src/display/scale.js
+++ b/src/display/scale.js
@@ -3,7 +3,11 @@ import { ImageKind } from "../shared/util.js";
 class ScaleJS {
   isMobile() {
     if (typeof navigator !== "undefined") {
-      return /Mobi/.test(navigator.userAgent);
+      return (
+        /iPad|iPhone|iPod/.test(navigator.userAgent) &&
+        window &&
+        !window.MSStream
+      );
     }
     return false;
   }
--- a/src/display/scale.js
+++ b/src/display/scale.js
@@ -1,0 +1,513 @@
+import { ImageKind } from "../shared/util.js";
+
+class ScaleJS {
+  isMobile() {
+    if (typeof navigator !== "undefined") {
+      return /Mobi/.test(navigator.userAgent);
+    }
+    return false;
+  }
+
+  determineScale(width, height) {
+    var testColor = "#ffffff";
+    var colorData = new Uint8ClampedArray(4);
+    var scale = 0;
+    if (typeof window !== "undefined" && this.isMobile()) {
+      var canvas = document.createElement("canvas");
+      var context = canvas.getContext("2d");
+      while (colorData[0] !== 255) {
+        scale += 1;
+        canvas.width = width / scale;
+        canvas.height = height / scale;
+        context.fillStyle = testColor;
+        context.fillRect(0, 0, 1, 1);
+        colorData = context.getImageData(0, 0, 1, 1).data;
+      }
+    }
+    return scale;
+  }
+
+  // Determine if we need to scale the PDF image
+  needsScale(width, height, colorSpace, bpc) {
+    var scaleObj = { width, height, scale: 1, transformImage: false };
+
+    var scale = this.determineScale(width, height);
+
+    if (scale > 1) {
+      scaleObj.transformImage = true;
+      scaleObj.scale = scale;
+      if (colorSpace === "DeviceGray" && bpc === 1 && scaleObj.scale > 2) {
+        scaleObj.scale = 4;
+      }
+      scaleObj.width = Math.ceil(width / scaleObj.scale);
+      scaleObj.height = Math.ceil(height / scaleObj.scale);
+    }
+    return scaleObj;
+  }
+
+  scaleImg(imgData) {
+    if (!imgData.scaleProcessed) {
+      var imgScaleObj;
+      if (imgData.kind === ImageKind.GRAYSCALE_1BPP) {
+        imgScaleObj = this.needsScale(
+          imgData.width,
+          imgData.height,
+          "DeviceGray",
+          1
+        );
+        if (imgScaleObj.transformImage) {
+          imgData.kind = ImageKind.GRAYSCALE_8BPP;
+          imgData.data = this.scaleBWImage(
+            imgData.data,
+            imgData.width,
+            imgData.height,
+            (imgData.width + 7) >> 3,
+            imgScaleObj.scale
+          );
+          imgData.width = imgScaleObj.width;
+          imgData.height = imgScaleObj.height;
+        }
+      } else if (imgData.kind === ImageKind.RGB_24BPP) {
+        imgScaleObj = this.needsScale(imgData.width, imgData.height, "RGB", 1);
+        if (imgScaleObj.transformImage) {
+          imgData.data = this.scaleRGBImage(
+            imgData.data,
+            imgData.width,
+            imgData.height,
+            imgScaleObj.scale
+          );
+          imgData.width = imgScaleObj.width;
+          imgData.height = imgScaleObj.height;
+        }
+      }
+      imgData.scaleProcessed = true;
+    }
+    return imgData;
+  }
+
+  scaleMask(imgData) {
+    if (!imgData.scaleProcessed) {
+      imgData.isMaskScaled = false;
+      var imgScaleObj = this.needsScale(
+        imgData.width,
+        imgData.height,
+        "DeviceGray",
+        1
+      );
+      if (imgScaleObj.transformImage) {
+        imgData.data = this.scaleBWImage(
+          imgData.data,
+          imgData.width,
+          imgData.height,
+          (imgData.width + 7) >> 3,
+          imgScaleObj.scale
+        );
+        imgData.width = imgScaleObj.width;
+        imgData.height = imgScaleObj.height;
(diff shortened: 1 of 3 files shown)


src/display/scale.js
```diff
@@ -0,0 +1,513 @@
+import { ImageKind } from "../shared/util.js";
+
+class ScaleJS {
+  isMobile() {
+    if (typeof navigator !== "undefined") {
+      return /Mobi/.test(navigator.userAgent);
```
automatedbugreportingfacility, line 6: I'm not sure if this check is correct. For example, the document linked in issue #9474 loads just fine in Firefox Preview Nightly and Google Chrome on Android 10 (Samsung Galaxy Note 10+). Maybe a more specific user agent check could be introduced to avoid unnecessary scaling, which has a side effect of degrading image quality.

Snuffleupagus:
> Though there is another PR for this, that has been closed, #9645, the current PR has been revisited and has the current pdfjs merged in.

Given that that PR wasn't accepted back then, it's not really clear why it should be accepted now unfortunately. Especially since this patch still seem to suffer from the same problems as the old one, note e.g. https://github.com/mozilla/pdf.js/pull/9645#issuecomment-381969495, https://github.com/mozilla/pdf.js/pull/9645#pullrequestreview-112784167, and that this (possible) heavy work happens on the main-thread.

src/display/scale.js
CIE81: We can use something like:  /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream; to narrow it down to IOS mobile devices.

CIE81:
> > Though there is another PR for this, that has been closed, #9645, the current PR has been revisited and has the current pdfjs merged in.
> 
> Given that that PR wasn't accepted back then, it's not really clear why it should be accepted now unfortunately. Especially since this patch still seem to suffer from the same problems as the old one, note e.g. [#9645 (comment)](https://github.com/mozilla/pdf.js/pull/9645#issuecomment-381969495), [#9645 (review)](https://github.com/mozilla/pdf.js/pull/9645#pullrequestreview-112784167), that the patch doesn't seem to take into account zooming in the viewer (which could invalidate a downsized image), and that this (possibly) heavy work shouldn't happen on the main-thread.

PDF viewer zooming works with this fix for the problematic pdfs. Regarding the possible heavy work on the main thread, when we load a pdf, even if we are doing all the computations in a separate worker, the main thread still has to wait for the scaling to complete before doing anything, since it has to load the document, so I don't know what would be the use of having another thread for this.

Benguenette:
Hi, any chance this issue could be fixed for web browser (PC with Chrome).

We built pdf.js and pdf.worker.js from this fork successfully -> https://github.com/CIE81/pdf.js/tree/IOS600DPI. It does not seem to fix our problem:

We are using pdfjs on our webapp on Chrome (not IOS).
pdf file of 150 dpi always worked good.
pdf file of 300 dpi and up didn't work before (blank page) and still doesn't work with the new build.
Any suggestion? Is it normal?

Thank you in advance

CIE81:
> Hi, any chance this issue could be fixed for web browser (PC with Chrome).
> 
> We built pdf.js and pdf.worker.js from this fork successfully -> https://github.com/CIE81/pdf.js/tree/IOS600DPI. It does not seem to fix our problem:
> 
> We are using pdfjs on our webapp on Chrome (not IOS).
> pdf file of 150 dpi always worked good.
> pdf file of 300 dpi and up didn't work before (blank page) and still doesn't work with the new build.
> Any suggestion? Is it normal?
> 
> Thank you in advance

The issue that is fixed addressed in this fork was happening due to a known webkit problem on IOS mobile devices. We didn't have any issue with 600dpi scanned PDF files in chrome on a windows PC before.

Does your issue only happens with this branch, or with the pdfjs master too ? Can you send a file you have problems viewing it in Chrome ?

Ioan

CIE81:
> Hi loan,
> 
> Thank for the reply. We had the problem with pdfjs master and this branch too (that we tried only for this week).
> I attached the files with different dpi. You will notice, 300dpi and over doesn't work unfortunately.
> 
> We looked at several issue about this but didn't find any solution.
> You can have similar problem here: #6741
> 
> It would mean the world if we could have a fix or a workaround. Thx.
> [293-resized 72dpi.pdf](https://github.com/mozilla/pdf.js/files/4704757/293-resized.72dpi.pdf)
> [293-resized 150dpi.pdf](https://github.com/mozilla/pdf.js/files/4704758/293-resized.150dpi.pdf)
> [293-resized 300dpi.pdf](https://github.com/mozilla/pdf.js/files/4704759/293-resized.300dpi.pdf)
> [293-Original 1200 dpi.pdf](https://github.com/mozilla/pdf.js/files/4704760/293-Original.1200.dpi.pdf)

Hi,

The width x height of the  images is 26,484 x 20,378, that is the issue. 

_"All web browsers limit the canvas element's width, height, and area. For Google Chrome, the maximum allowable width and height are 32,767 pixels and the maximum allowable area is 268,435,456 pixels."_

The area of your images is 539,690,952.

The good news, though, is that the algorithm will work with a very small change. 

If you remove the isMobile() check from src/display.scale.js determineScale method(I think it is line 15), it will also consider non mobile devices too, and if the canvas is too big to be rendered, it will scale the image.

@Benguenette Please try it and let me know. It worked for me.

@Snuffleupagus This fix can solve other canvas sizing issues too. The documents will not be scaled if they can be render (determineScale method checks that), so I don't think there is any risk in removing the mobile IOS devices check.

Ioan

Benguenette:
Hi loan.

It is working. Thank you! Finally, no longer headaches.

One last thing - it was easier for me to give you the original file (which I deleted on my end), but since your replied my message, the link to the files are still there. Can I ask you to delete the files please.

Thank you again. 👍

CIE81:
Glad it is working :).

Hope to get this branch in PDFJS too.

I have deleted the files, please check.

Thank you,
Ioan

Benguenette:
> Glad it is working :).
> 
> Hope to get this branch in PDFJS too.
> 
> I have deleted the files, please check.
> 
> Thank you,
> Ioan

Sorry for getting back at you about that again - Any chance you could delete your Edited version? 
image

Thanks :)

CIE81:
Let me know if it is ok now :)

Ioan

CIE81:
To sum it up, this fix can address large canvas sizing issue son any browser/device, so I might remove the mobile check.

We will not resize all the images, just the ones that are rendered on a canvas too latge to be suported by the browser.

Ioan

Benguenette:
> Let me know if it is ok now :)
> 
> Ioan

It is! Thank you.

stsje:
Any plans on merging this pull request? 600 DPI when printing is still an issue in iOS

Snuffleupagus:
> Any plans on merging this pull request? 600 DPI when printing is still an issue in iOS

Please note that iOS isn't the primary development target for the PDF.js library (rather it's the Firefox built-in PDF Viewer).

As-is this PR would add *a lot* of additional complexity to the rendering code (and reviewing this looks anything but trivial), thus increasing the maintenance/testing burden for the `src/display/canvas.js`-code considerably, and there's also the question of the performance impact by running this kind of code on the main-thread.

All-in-all, let's close this as WONTFIX for now given its age and the points mentioned previously (here and in the other PR).

Issue #10773, opened 2019-04-28
missing support for "view" parameter in URL fragment identifiers


markmatney:
According to [Parameters for Opening PDF Files (version 9.0)](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/pdf_open_parameters_v9.pdf#page=6) (and referenced in [RFC 8118](https://tools.ietf.org/html/rfc8118#section-3)):

> view=Fit
> view=FitH
> view=FitH,top
> view=FitV
> view=FitV,left
> view=FitB
> view=FitBH
> view=FitBH,top
> view=FitBV
> view=FitBV,left
>
> Set the view of the displayed page, using the keyword values defined in the PDF language specification. For more information, see the PDF Reference. Scroll values left and top are floats or integers in a coordinate system where 0,0 represents the top left corner of the visible page, regardless of document rotation.

It appears that the `view` parameter in PDF URL fragment identifiers is currently unsupported. To verify, I tested as follows:

---

Configuration:
- Web browser and its version: Firefox Quantum 66.0.3 (64-bit)
- Operating system and its version: GNU/Linux x86_64
- PDF.js version: https://github.com/mozilla/pdf.js/tree/762c58e0fcfdc8f3fd23ce22a15ee044d8d372ec
- Is a browser extension: no

Steps to reproduce the problem:
1. Check out the repository, run `npm install` and `gulp server`, then navigate the viewer to a PDF that _isn't_ automatically zoomed in to fill your viewport width on viewer initialization (http://localhost:8888/web/viewer.html?file=%2Ftest%2Fpdfs%2Ftracemonkey.pdf works for me).
2. Append `#page=1&view=FitH` to the URL in the address bar, and reload the page.

What is the expected behavior?

The viewer should set the zoom level so that the PDF fills the entire width of the viewport.

What went wrong?

The viewport was not transformed as expected.

---

It looks like the problem is mainly with `PDFLinkService.setHash`. There is no check for the `view` parameter.

Interestingly enough, `Fit*` values _are_ being checked for in the `zoom`parameter, but I'm having trouble understanding why (since none of [RFC 3778](https://tools.ietf.org/html/rfc3778#section-3), RFC 8118, or the PDF Open Parameters document say that those are vallid values for `zoom`). The deepest I was able to dig was [this PR](https://github.com/mozilla/pdf.js/pull/3910), but it looks like `Fit*` values have been handled this way even before that.

residentevil1256:
[...]

PR #11786, opened 2020-04-07
Implemented #10773 with potential fix on #2843

huang325:
Add support for `view` parameter of pdf file provided in web url.
Attempt to solve "upside-down" issue.


Snuffleupagus:
> Attempt to solve "upside-down" issue.

Again, that part would be a case of a patch breaking way more than it fixes.
The changes you've made in `web/base_viewer.js` are unfortunately wrong, since they're breaking all handling of destinations in PDF documents. Issue 2843 concerns *only* the hash parameters, and is somewhat difficult to address in a good/efficient way (since it seems that the "zoom" hash parameter may even need to be rotational invariant).
Please just remove that part from the PR, since it's not correct and again please make sure to carefully test your code before submitting a PR.

> Add support for `view` parameter of pdf file provided in web url.

If that is done, assuming the rest of the patch is OK, that possibly means that the "Fit" handling ought to be removed from the "zoom"-parameter.

huang325:
Hi @Snuffleupagus ,
Thanks for your review, I've reverted my changes in `base_viewer.js` and removed the Fit related part in zoom.

I'm still thinking what would be the best choice to fix 2843. I considered to make another wrapper outside `scrollPageIntoView` which only used by Fit and zoom, but seems that's not elegent. As you were previously working on this issue, I'd like to discuss and hear any advice from you.

test/unit/test_utils.js
```diff
@@ -175,6 +175,52 @@ function createIdFactory(pageIndex) {
   return page.idFactory;
 }
 
+class BaseViewerMock {
```
Snuffleupagus, line 178: Sorry, but how is adding this mock class with a *completely different* (not to mention incomplete) interface than the actual `BaseViewer` class in any way helpful/useful here!?

This is actually a case where I don't think that the added unit-tests add much, if any, value, since they don't really exercise *any* of the normal code-paths in e.g. `PDFViewer`. Consequently I'd suggest removing these unit-tests from the patch.

web/pdf_link_service.js
```diff
@@ -257,20 +257,7 @@ class PDFLinkService {
...
-            zoomArg === "FitV" ||
-            zoomArg === "FitBV"
-          ) {
-            dest = [
-              null,
-              { name: zoomArg },
-              zoomArgs.length > 1 ? zoomArgs[1] | 0 : null,
-            ];
-          } else if (zoomArg === "FitR") {
+          if (zoomArg === "FitR") {
```
Snuffleupagus, line 260: Why wasn't this case simply moved to the "view" branch below? It might not be in the open-parameters specification, but since we supported it previously it should remain available.

Snuffleupagus (changes requested):
> I'm still thinking what would be the best choice to fix 2843.

That *really* isn't a good beginner bug, otherwise it'd have been fixed a long time ago.

---

This PR contains *a lot* of changes that shouldn't be part of it, which thus needs to be removed; please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

Furthermore, despite being asked on numerous occasions now the patches still lack good and meaningful commit messages.

Finally, I'd suggest simplifying this PR to *only* the following diff (which is the minimal solution to issue 10773):
```diff
diff --git a/web/pdf_link_service.js b/web/pdf_link_service.js
index 97e0eb53..34f86efa 100644
--- a/web/pdf_link_service.js
+++ b/web/pdf_link_service.js
@@ -257,40 +257,49 @@ class PDFLinkService {
             zoomArgNumber ? zoomArgNumber / 100 : zoomArg,
           ];
         } else {
-          if (zoomArg === "Fit" || zoomArg === "FitB") {
-            dest = [null, { name: zoomArg }];
-          } else if (
-            zoomArg === "FitH" ||
-            zoomArg === "FitBH" ||
-            zoomArg === "FitV" ||
-            zoomArg === "FitBV"
-          ) {
+          console.error(
+            `PDFLinkService.setHash: "${zoomArg}" is not ` +
+              "a valid zoom value."
+          );
+        }
+      }
+      if ("view" in params) {
+        const zoomArgs = params.zoom.split(","); // scale,left,top
+        const zoomArg = zoomArgs[0];
+
+        if (zoomArg === "Fit" || zoomArg === "FitB") {
+          dest = [null, { name: zoomArg }];
+        } else if (
+          zoomArg === "FitH" ||
+          zoomArg === "FitBH" ||
+          zoomArg === "FitV" ||
+          zoomArg === "FitBV"
+        ) {
+          dest = [
+            null,
+            { name: zoomArg },
+            zoomArgs.length > 1 ? zoomArgs[1] | 0 : null,
+          ];
+        } else if (zoomArg === "FitR") {
+          if (zoomArgs.length !== 5) {
+            console.error(
+              'PDFLinkService.setHash: Not enough parameters for "FitR".'
+            );
+          } else {
             dest = [
               null,
               { name: zoomArg },
-              zoomArgs.length > 1 ? zoomArgs[1] | 0 : null,
+              zoomArgs[1] | 0,
+              zoomArgs[2] | 0,
+              zoomArgs[3] | 0,
+              zoomArgs[4] | 0,
             ];
-          } else if (zoomArg === "FitR") {
-            if (zoomArgs.length !== 5) {
-              console.error(
-                'PDFLinkService.setHash: Not enough parameters for "FitR".'
-              );
-            } else {
-              dest = [
-                null,
-                { name: zoomArg },
-                zoomArgs[1] | 0,
-                zoomArgs[2] | 0,
-                zoomArgs[3] | 0,
-                zoomArgs[4] | 0,
-              ];
-            }
-          } else {
-            console.error(
-              `PDFLinkService.setHash: "${zoomArg}" is not ` +
-                "a valid zoom value."
-            );
           }
+        } else {
+          console.error(
+            `PDFLinkService.setHash: "${zoomArg}" is not ` +
+              "a valid view value."
+          );
         }
       }
       if (dest) {
```

markmatney:
@huang325 I really appreciate you taking a look at this! I'm not a maintainer so I can't give much implementation advice, but since I filed #10773 I thought I would help translate the response you've gotten, in case you're stuck:
- It looks like the maintainer has [suggested a fix](https://github.com/mozilla/pdf.js/pull/11786#pullrequestreview-389803607) -- sounds like they want you to update your branch to have a single commit with those changes only, with a good commit message.
- You might take a look at https://chris.beams.io/posts/git-commit/ for guidance on how to write a good commit message. Try to write yours in a style that's consistent with [the rest of the commits in this project](https://github.com/mozilla/pdf.js/commits/master).

Although, thinking about this a little more, **I'm not so sure it's even a good idea to add this feature while #2843 is still open.** It won't work properly until that issue is fixed, and it seems like not a good practice in general to add a feature _that we already know will be buggy_ to the public API, especially to the PDF rendering engine of a web browser with (perhaps hundreds of) millions of users. @Snuffleupagus thoughts?

huang325:
Hi,
Thanks for the updates. I will update my PR according to the feedback.

web/pdf_link_service.js
huang325: Hi,
That is because FitR is not included in [the document](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/pdf_open_parameters_v9.pdf#page=6). So I wonder this only belongs to zoom, not view.

timvandermeij:
Closing since there has been no activity for a month. If you continue working on this, please open a new PR with all comments addressed. Thanks.

PR #11789, opened 2020-04-08
Add a new `pdfjs.enablePermissions` preference, off by default, to allow the PDF documents to disable copying in the viewer (bug 792816)

### LINKED ISSUE #3471: Security: Ignores protection from Copy
Documents generated with FPDF with security (namely protection against copy - I haven't check print, password or other protections) not working.

Here is a script from the FPDF website which containts a link to its output.
http://www.fpdf.org/en/script/script37.php

The protection works on Acrobat Reader 8.\* in Win Vista

I am using FF 22.

Thanks

Snuffleupagus:
*Recently I've been looking through some older issues/bugs, to see if there's any ones that are now easy to address; this is a simple one that I found.* 

---

*Please note:* Most of the necessary API work was done in PR #10033, and the only remaining thing to do here was to implement it in the viewer.

The new preference should thus allow e.g. enterprise users to disable copying in the viewer, for PDF documents whose permissions specify that.

In order to simplify things the "copy"-permission was implemented using CSS, as suggested in https://bugzilla.mozilla.org/show_bug.cgi?id=792816#c55, which should hopefully suffice.[1]
The advantage of this approach, as opposed to e.g. disabling the `textLayer` completely, is first of all that it ensures that searching still works correctly even in copy-protected documents. Secondly this also greatly simplifies the overall implementation, since it doesn't require a lot of code for something that's disabled by default.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=792816 (assuming we care to do so).
Fixes #3471
/cc @brendandahl 

---
[1] As the discussion in the bug shows, this kind of copy-protection is not very strong and is also generally easy to remove/circumvent in various ways. Hence a simple solution, targeting "regular"-users rather than "power"-users is hopefully deemed acceptable here.

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -151,6 +151,10 @@
       "type": "boolean",
       "default": false
     },
+    "enablePermissions": {
+      "type": "boolean",
+      "default": false
+    },
     "historyUpdateUrl": {
       "type": "boolean",
       "default": false
--- a/web/app.js
+++ b/web/app.js
@@ -48,6 +48,7 @@ import {
   MissingPDFException,
   OPS,
   PDFWorker,
+  PermissionFlag,
   shadow,
   UnexpectedResponseException,
   UNSUPPORTED_FEATURES,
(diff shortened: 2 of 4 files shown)


mkaply:
Was 10003 already merged into Firefox?

Snuffleupagus:
> Was 10003 already merged into Firefox?

Yes, it landed all the way back in Firefox 64 (see https://bugzilla.mozilla.org/show_bug.cgi?id=1488420); assuming that you meant #10033 of course :-)

mkaply:
Great. As soon as this lands in Firefox, I'll create an enterprise policy to enable.

PR #11797, opened 2020-04-10
Don't bundle the fallback `grab`/`grabbing` cursor images when running `gulp mozcentral`

Snuffleupagus:
These cursor images are only necessary as a fallback for older browsers, hence there's no reason to keep shipping them in Firefox as far as I can tell; see https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#Browser_compatibility

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1027,11 +1027,17 @@ gulp.task(
     var version = versionJSON.version,
       commit = versionJSON.commit;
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
I would agree that Firefox does not need this. Thanks!

PR #11806, opened 2020-04-12
[api-minor] Fix the return value of `PDFDocumentProxy.getViewerPreferences` when no viewer preferences are present (PR 10738 follow-up)

Snuffleupagus:
This patch fixes yet another instalment in the never-ending series of "what the \*bleep\* was I thinking", by changing the `PDFDocumentProxy.getViewerPreferences` method to return `null` by default.
Not only is this method now consistent with many other API methods, for the data not present case, but it also avoids having to e.g. loop through an object to check if it's actually empty (note the old unit-test).

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -479,7 +479,7 @@ class Catalog {
     };
 
     const obj = this.catDict.get("ViewerPreferences");
-    const prefs = Object.create(null);
+    let prefs = null;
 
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me pending a rebase.

timvandermeij:
Thanks!

PR #11809, opened 2020-04-14
Update Prettier to version 2.0

Snuffleupagus:
Please note that these changes were done automatically, using `gulp lint --fix`.

Given that the major version number was increased, there's a fair number of (primarily whitespace) changes; please see https://prettier.io/blog/2020/03/21/2.0.0.html
In order to reduce the size of these changes somewhat, this patch maintains the old "arrowParens" style for now (once mozilla-central updates Prettier we can simply choose the same formatting, assuming it will differ here).

*Much smaller/easier diff with https://github.com/mozilla/pdf.js/pull/11809/files?w=1*

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,5 @@
 {
+  "arrowParens": "avoid",
   "endOfLine": "lf",
   "printWidth": 80,
   "semi": true,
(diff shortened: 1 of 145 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for updating this!

PR #11810, opened 2020-04-15
A couple of small `String.fromCodePoint` improvements (PR 11698 and 11769 follow-up)

Snuffleupagus:
- Add a reduced test-case for issue 11768, to prevent future regressions.
   (Given that PR 11769 is only a work-around, rather than a proper solution, it may not be entirely accurate for the issue to be closed as fixed.)

 - Add more validation of the charCode, as found by the heuristics, in `PartialEvaluator._buildSimpleFontToUnicode` to prevent future issues.


timvandermeij (approved):

timvandermeij:
Thank you for the follow-up!

PR #11813, opened 2020-04-16
Suppress browser autofill on page number

AndyTeas:
What the user did:
    Open the PDF Viewer in Chrome;
    Mouse click into the “Page number” input field;

What they saw:
    A pop-up list with seemingly random numbers

What you were expecting to see:
    Nothing

What they see is the Chrome “Autofill” feature at work – that is suggesting values that you have previously entered into number fields in forms, as possible values you may want to enter into this field.  The list has nothing to do with the PDF currently open but the user does not know this.


timvandermeij (approved):

timvandermeij:

PR #6915, opened 2016-01-22
Refactor lookup hash tables/objects

yurydelendik:
Fixes #6774 (by adding lazy initialization to the large lookup tables).

Mostly refactors fonts.js file by moving large tables to separate files. Also replaces literal `{}` usages for cache and lookup tables.

PR #11818, opened 2020-04-17
Enable the `dot-notation` ESLint rule

Snuffleupagus:
*Please note:* These changes were done automatically, using the `gulp lint --fix` command.

This rule is already enabled in mozilla-central, see https://searchfox.org/mozilla-central/rev/567b68b8ff4b6d607ba34a6f1926873d21a7b4d7/tools/lint/eslint/eslint-plugin-mozilla/lib/configs/recommended.js#103-104

The main advantage, besides improved consistency, of this rule is that it reduces the size of the code (by 3 bytes for each case). In the PDF.js code-base there's close to 8000 instances being fixed by the `dot-notation` ESLint rule, which end up reducing the size of even the *built* files significantly; the total size of the `gulp mozcentral` build target changes from `3 247 456` to `3 224 278` bytes, which is a *reduction* of `23 178` bytes (or ~0.7%) for a completely mechanical change.

A large number of these changes affect the (large) lookup tables used on the worker-thread, but given that they are still initialized lazily I don't *think* that the new formatting this patch introduces should undo any of the improvements from PR #6915.

Please find additional details about the ESLint rule at https://eslint.org/docs/rules/dot-notation


timvandermeij (approved):

timvandermeij:
I think this is also nicer visually, but the consistency is primarily nice. Thank you!

PR #11805, opened 2020-04-12
Always skip over any additional, unexpected, RSTx (restart) markers in corrupt JPEG images (issue 11794)

Snuffleupagus:
Fixes #11794 

*Much smaller/simpler diff with https://github.com/mozilla/pdf.js/pull/11805/files?w=1*

---

With the changes in this PR, I believe that it *should* also be possible to simply remove the code-block shown below. However, I wanted to avoid increasing the size/scope of this PR even more, given that it already contains a clean-up commit.
https://github.com/mozilla/pdf.js/blob/a4dd081d7be075c102abe164c74d5900c73066a8/src/core/jpg.js#L456-L465

PR #11819, opened 2020-04-17
[src/core/jpg.js] Remove redundant marker validation at the end of the `decodeScan` function (PR 11805 follow-up)

Snuffleupagus:
With the MCU parsing changes made in PR #11805, the final marker validation is no longer necessary before the `decodeScan` function returns.


timvandermeij (approved):

timvandermeij:
Let's do this; thanks!

PR #11826, opened 2020-04-20
Change the "download" keyboard shortcut (Ctrl+S) handling, in GENERIC/CHROME builds, to utilize the `EventBus` (issue 11657); add a new "openfile" keyboard shortcut (Ctrl+O), in GENERIC builds 

### LINKED ISSUE #11657: No event bus used for Ctrl+S
This is a minor niggle, but it seems like there is a good way to run code whenever the user downloads or prints, by using `PDFViewerApplication.eventBus.on("download", ..)` and `PDFViewerApplication.eventBus.on("beforeprint", ..)`. This works if the user clicks the download or print button, and works with Ctrl+P for print.

However, Ctrl+S to trigger the download does not use the eventBus. It would be nice to also use the eventBus here, so there is a single point to hook onto.

Snuffleupagus:
- **Change the "download" keyboard shortcut (Ctrl+S) handling, in GENERIC/CHROME builds, to utilize the `EventBus` (issue 11657)**

   This improves the consistency of the "download" handling, in the default viewer, such that the `Toolbar`/`SecondaryToolbar` buttons *and* the keyboard shortcut are now handled in the same way (using the `EventBus`).

   Given that the "download" keyboard shortcut handling is limited to GENERIC/CHROME builds and that the issue does raise a valid point about only being able to observe *some* downloads, these changes seem acceptable in this particular case.

   Finally the pre-processor condition is adjusted to *explicitly*, rather than implicitly, list the affected build targets.

   Fixes #11657

 - **Add a new "openfile" keyboard shortcut (Ctrl+O), in GENERIC builds**

   Somewhat surprisingly, despite the GENERIC viewer implementing "openfile" support, there's never been a keyboard shortcut available. Similar to the previous patch, this utilizes the `EventBus` for consistency with the `Toolbar`/`SecondaryToolbar` buttons.

---

*Please note:* This patch should NOT be construed as carte blanche to simply convert all of the code in `webViewerKeyDown`, or elsewhere, to make use of the `EventBus` instead of direct function calls.
Any further changes, along the lines in this patch, would need to be evaluated on a case-by-case basis to determine if they are actually wanted, given that many/most existing cases in `webViewerKeyDown` should already be *indirectly* observable through the `EventBus` instance.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2571,12 +2571,14 @@ function webViewerKeyDown(evt) {
     }
   }
 
-  if (typeof PDFJSDev === "undefined" || !PDFJSDev.test("MOZCENTRAL")) {
+  if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC || CHROME")) {
+    const { eventBus } = PDFViewerApplication;
+
     // CTRL or META without shift
     if (cmd === 1 || cmd === 8) {
       switch (evt.keyCode) {
         case 83: // s
-          PDFViewerApplication.download();
+          eventBus.dispatch("download", { source: window });
           handled = true;
           break;
       }
--- a/web/app.js
+++ b/web/app.js
@@ -2581,6 +2581,13 @@ function webViewerKeyDown(evt) {
           eventBus.dispatch("download", { source: window });
           handled = true;
           break;
+
+        case 79: // o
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks like a good improvement; thanks!

PR #11834, opened 2020-04-22
Preserve error types during translation

### LINKED ISSUE #11658: Preserve error types (e.g. InvalidPDFException, MissingPDFException)

xelan:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | yes
| New feature?  | yes
| Deprecations? | no
| Tickets       | Fix #11658 
| License       | Apache License 2.0

By preserving the exception type, more fine-grained error handling can be performed via client-side logic (e.g. redirect to a search page if a PDF is not found, or to a ticket system in case of invalid PDF files).

The original exception type and its properties (e.g. status code) are preserved by cloning the error object and assigning the translated message. Note that only a shallow clone is performed for simplicity.

Fixes #11658

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -817,7 +817,7 @@ const PDFViewerApplication = {
 
         return loadingErrorMessage.then(msg => {
           this.error(msg, { message });
-          throw new Error(msg);
+          throw exception;
         });
       }
     );


Snuffleupagus (changes requested):
This seem like way too much code/complexity for an edge-case, and in a previous PR (see #11689) the consensus was that *if* this should be fixed then simply re-throwing the original exception would be preferable.

xelan:
Ok, if the possible change of the original exception outside of pdf.js is acceptable, I can just rethrow the exception with the translated message, without any cloning :smile:

xelan:
I've pushed the updated, simplified version which rethrows the original exception.

web/app.js
```diff
@@ -817,7 +817,10 @@ const PDFViewerApplication = {
 
         return loadingErrorMessage.then(msg => {
           this.error(msg, { message });
-          throw new Error(msg);
+
+          exception.message = msg;
+
+          throw exception;
```
Snuffleupagus, line 823: This is still modifying, albeit less than the previously, the original exception though.

Why not simply `throw exception;` here, since the translation is mostly intended for the default viewer `errorWrapper`?

xelan: Thanks for your feedback, @Snuffleupagus!

That would of course also work. However it would be a BC break if an application using pdf.js relies on the translated exception message (e.g. to display user-facing error messages).

Snuffleupagus: The default viewer was created for usage in Firefox, and is *not* part of the versioned [public API](see https://github.com/mozilla/pdf.js/blob/master/src/display/api.js).
There's also the "viewer components", see https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.component.js and the examples in https://github.com/mozilla/pdf.js/tree/master/examples/components, which we also attempt to not break unnecessarily.

For the [default viewer](https://github.com/mozilla/pdf.js/blob/master/web/app.js), while we don't go out of our way to break backwards compatibility, there's absolutely *no guarantees* that there won't be changes; note also that we even ask people not to use it as -is e.g. in http://mozilla.github.io/pdf.js/getting_started/#introduction (emphasis mine):

---

> [...] if an application using pdf.js relies on the translated exception message (e.g. to display user-facing error messages).

Apart from the above, this particular situation also seem to be enough of an edge-case that I wouldn't really consider the change I suggested above to be a problem.
It'd even be more consistent, since it means that the viewer is now returning *exactly* the same Errors as if the "raw" API was used directly.

timvandermeij: I agree with the above. This looks good if the original error is re-thrown, i.e., unchanged.

xelan: That sounds good, thank you very much for the feedback. I've adapted the code accordingly.

timvandermeij (approved):

timvandermeij:
Let's do this. Thank you for your contribution!

xelan:
> Let's do this. Thank you for your contribution!

Great :+1: Your welcome. Thanks for the feedback and the quick and professional handling.

PR #11835, opened 2020-04-22
[api-minor] Change `PageViewport` to throw when the `rotation` is not a multiple of 90 degrees

Snuffleupagus:
As evident from the code, `PageViewport` only supports[1] `rotation` values which are a multiple of 90 degrees. Besides it being somewhat difficult to imagine meaningful use-cases for a non-multiple of 90 degrees `rotation`, the code also becomes both simpler and more efficient by not having to consider arbitrary `rotation` values.

However, any invalid rotation will *silently* fallback to assume zero `rotation` which probably isn't great for e.g. `PDFPageProxy.getViewport` in the API. Hence this patch, which will now enforce that only valid `rotation` values are accepted.

---
[1] As far as I can tell, from looking through the history, nothing else has ever been supported either.

### DIFF
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -245,13 +245,16 @@ class PageViewport {
         rotateC = -1;
         rotateD = 0;
         break;
-      // case 0:
-      default:
+      case 0:
         rotateA = 1;
         rotateB = 0;
         rotateC = 0;
         rotateD = -1;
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Makes sense; thank you!

PR #11837, opened 2020-04-22
Always attempt to dispatch the "webviewerloaded" event at the embedding `document` (PR 10318 follow-up, issue 11829)

### LINKED ISSUE #11829: Set locale don't work.
Hi

I try to set locale with PDFViewerApplicationOptions.set('locale', 'es-ES'), and it change the locale I tested by .get('locale') , but this action don't change the viewer to new language.

Configuration:
- Web browser and its version: Chrome 81.0.4044.113 , Safari   13.1 (15609.1.20.111.8)
- Operating system and its version: OSX 10.13
- PDF.js version: 2.3.200

1. console.log(PDFViewerApplicationOptions.get('locale')); // en-US
2. PDFViewerApplicationOptions.set('locale', 'es-ES');
3. console.log(PDFViewerApplicationOptions.get('locale')); // es-ES
4. but the viewer don't change the language:

a

Is this way the best way to change the locale? or is possible change the locale with other property?

And thanks for this amazing tool!
Best

Snuffleupagus:
This is necessary in order to support cases where the default viewer is embedded in a *dynamically* created <iframe> element.

In order to also support a use-case where there's *multiple* <iframe> elements (containing default viewers) on the same page, the "webviewerloaded" event now includes a `source` detail parameter such that it's possible to associate an event with the relevant DOM element.

Fixes #11829

### DIFF
--- a/web/viewer.js
+++ b/web/viewer.js
@@ -211,8 +211,20 @@ function webViewerLoad() {
       // set various `AppOptions`, by dispatching an event once all viewer
       // files are loaded but *before* the viewer initialization has run.
       const event = document.createEvent("CustomEvent");
-      event.initCustomEvent("webviewerloaded", true, true, {});
-      document.dispatchEvent(event);
+      event.initCustomEvent("webviewerloaded", true, true, {
+        source: window,
+      });
+      try {
+        // Attempt to dispatch the event at the embedding `document`,
+        // in order to support cases where the viewer is embedded in
+        // a *dynamically* created <iframe> element.
+        parent.document.dispatchEvent(event);
+      } catch (ex) {
+        // The viewer could be in e.g. a cross-origin <iframe> element,
+        // fallback to dispatching the event at the current `document`.
+        console.error(`webviewerloaded: ${ex}`);
+        document.dispatchEvent(event);
+      }
     }
 
     pdfjsWebApp.PDFViewerApplication.run(config);


web/viewer.js
```diff
@@ -211,8 +211,12 @@ function webViewerLoad() {
...
       // files are loaded but *before* the viewer initialization has run.
       const event = document.createEvent("CustomEvent");
-      event.initCustomEvent("webviewerloaded", true, true, {});
-      document.dispatchEvent(event);
+      event.initCustomEvent("webviewerloaded", true, true, {
+        source: window,
+      });
+      // Dispatch the event at the top-level `document`, to support cases where
+      // the viewer is embedded in a *dynamically* created <iframe> element.
+      top.document.dispatchEvent(event);
```
automatedbugreportingfacility, line 219: `top.document` will throw if top is a cross-origin window. Not sure if that's relevant here and/or should be considered.

web/viewer.js
```diff
@@ -211,8 +211,20 @@ function webViewerLoad() {
...
-      event.initCustomEvent("webviewerloaded", true, true, {});
-      document.dispatchEvent(event);
+      event.initCustomEvent("webviewerloaded", true, true, {
+        source: window,
+      });
+      try {
+        // Attempt to dispatch the event at the top-level `document`,
+        // in order to support cases where the viewer is embedded in
+        // a *dynamically* created <iframe> element.
+        top.document.dispatchEvent(event);
```
automatedbugreportingfacility, line 221: I was wondering if dispatching this event to the embedding document would be more correct, i.e. `parent.document`. I'm pretty sure in the vast majority of cases the embedder will be `top`, so maybe it doesn't matter that much :-)

web/viewer.js
```diff
@@ -211,8 +211,20 @@ function webViewerLoad() {
...
+      });
+      try {
+        // Attempt to dispatch the event at the embedding `document`,
+        // in order to support cases where the viewer is embedded in
+        // a *dynamically* created <iframe> element.
+        parent.document.dispatchEvent(event);
+      } catch (ex) {
+        // The viewer could be in e.g. a cross-origin <iframe> element,
+        // fallback to dispatching the event at the current `document`.
+        console.error(`webviewerloaded: ${ex}`);
```
timvandermeij, line 225: Since there is a valid case in which this can happen and we have a fallback to handle it, do we need the error log here?

Snuffleupagus: In this special case you won't be able to (easily) listen for the "webviewerloaded" event if using a *dynamically* created <iframe> element, and there's basically no reasonable way to fix the referenced issue in that case unfortunately.

By at least logging the error here, we should hopefully save time debugging if someone submits a future issue specifically for a cross-origin and dynamically created <iframe> element for the viewer.

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #11847, opened 2020-04-26
Move the `maybeValidDimensions` check, used with JPEG images, to occur earlier (PR 11523 follow-up)

Snuffleupagus:
Given that the `NativeImageDecoder.{isSupported, isDecodable}` methods require both dictionary lookups *and* ColorSpace parsing, in hindsight it actually seems more reasonable to the `JpegStream.maybeValidDimensions` checks *first*.


timvandermeij (approved):

timvandermeij:
Good idea!

PR #11863, opened 2020-04-30
Add more categories of unsupported features.

### LINKED ISSUE #11815: Collect telemetry on what causes the fallback bar
To get a better idea of what is causing the fallback bar to show the most, we need to collect more detailed information on what is triggering it. We already somewhat do this with the [UNSUPPORTED_FEATURES](https://github.com/mozilla/pdf.js/blob/c218e94f664c97d3c95994d57b11d11488dae54b/src/shared/util.js#L290-L297) enum, but we need to audit each place that it is used and create more finer grained categories.

brendandahl:
Fixes #11815

For non-generic missing features that are errors I've prefixed them with "error."

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -282,7 +282,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                   // Error(s) in the ExtGState -- sending unsupported feature
                   // notification and allow parsing/rendering to continue.
                   this.handler.send("UnsupportedFeature", {
-                    featureId: UNSUPPORTED_FEATURES.unknown,
+                    featureId: UNSUPPORTED_FEATURES.errorExtGState,
                   });
                   warn(`hasBlendModes - ignoring ExtGState: "${ex}".`);
                   continue;
@@ -344,7 +344,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
                 // Error(s) in the XObject -- sending unsupported feature
                 // notification and allow parsing/rendering to continue.
                 this.handler.send("UnsupportedFeature", {
-                  featureId: UNSUPPORTED_FEATURES.unknown,
+                  featureId: UNSUPPORTED_FEATURES.errorXObject,
                 });
                 warn(`hasBlendModes - ignoring XObject: "${ex}".`);
                 continue;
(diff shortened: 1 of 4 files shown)


src/core/evaluator.js
```diff
@@ -802,7 +802,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               // Error in the font data -- sending unsupported feature
               // notification.
               this.handler.send("UnsupportedFeature", {
-                featureId: UNSUPPORTED_FEATURES.font,
+                featureId: UNSUPPORTED_FEATURES.errorType3FontLoad,
```
Snuffleupagus, line 805: Since all the other font-related ones follow a `errorFont...` pattern, how about naming this one `errorFontLoadType3` instead since that's also consistent with the actual method name used above?

src/display/font_loader.js
```diff
@@ -400,7 +402,9 @@ class FontFaceObject {
         throw ex;
       }
       if (this._onUnsupportedFeature) {
-        this._onUnsupportedFeature({ featureId: UNSUPPORTED_FEATURES.font });
+        this._onUnsupportedFeature({
+          featureId: UNSUPPORTED_FEATURES.errorFontGenPath,
```
Snuffleupagus, line 406: `errorFontGetPath`?

src/shared/util.js
```diff
@@ -294,6 +294,18 @@ const UNSUPPORTED_FEATURES = {
   smask: "smask",
   shadingPattern: "shadingPattern",
   font: "font",
```
Snuffleupagus, line 296: Assuming that the `font` key is actually unused now, should it be removed?

The same question also applies for `unknown` key above.

brendandahl: These are exposed on the API, so I'll just mark them as `@deprecated` for now,

Snuffleupagus: > These are exposed on the API,

That's certainly true, but considering that nothing (in the PDF.js code) actually uses them now I just wondered what good they do for an API user at this point!? 

Anyway, it probably doesn't matter much either way here and the patch looks good as-is :-)

Snuffleupagus (approved):

src/shared/util.js
```diff
@@ -288,12 +288,26 @@ const OPS = {
...
+  errorExtGState: "errorExtGState",
+  errorXObject: "errorXObject",
+  errorFontLoadType3: "errorFontLoadType3",
+  errorFontState: "errorFontState",
+  errorFontMissing: "errorFontMissing",
+  errorFontTranslate: "errorFontTranslate",
+  errorColorSpace: "errorColorSpace",
+  errorOperatorList: "errorOperatorList",
+  errorFontToUnicode: "errorFontToUnicode",
+  errorFontLoadNativeFont: "errorFontLoadNative",
```
timvandermeij, line 309: I think this should be `errorFontLoadNative` to match the value here and the usage above.

timvandermeij (approved):

timvandermeij:
Thank you!

PR #11864, opened 2020-05-01
Remove the `create-react-app` example (issue 11729)

### LINKED ISSUE #11729: create-react-app example uses a fake worker
Configuration:
- PDF.js version: 2.4.456

1. `cd examples/create-react-app`
2. `npm install && npm start`
3. Open the js console
4. Observe `Warning: Setting up fake worker.`

Example loads pdfjs.worker as an actual web worker

What went wrong?
A fake worker was used

Currently, `examples/create-react-app/package.json` depends on pdfjs-dist version 2.2.228, but I upgraded it to 2.4.456 and observed the same behavior.

A workaround is to serve the js worker statically in the `public/`, but it would be nice if the example included in the pdf.js repository were working out of the box.

Snuffleupagus:
Given that none of the PDF.js contributors know React, maintaining and/or providing supporting for the example isn't really feasible unfortunately.
Even something as simple as running/testing the example becomes difficult for anyone completely unfamiliar with React, and furthermore:

 - It's very difficult to tell if the example demonstrates React best-practices, since the PDF.js contributors don't know React.

 - We also have no reasonable way of keeping the example up-to-date with changes in React.

 - The React example, in its current form, is even *hard-coding* the PDF.js version to a now unsupported version.

 - The example is currently triggering "fake worker" usage, see issue 11729, which is really *really* bad. Note that the "fake worker" functionality is *only* intended as a fallback, and it should absolutely *not* under any circumstances be advertised and certainly shouldn't be triggered in official PDF.js examples.

Fixes #11729


timvandermeij (approved):

timvandermeij:
I can only agree with this, and will make sure to be more critical about adding such examples in the future. Thanks!

Snuffleupagus:
> [...] and will make sure to be more critical about adding such examples in the future.

Don't worry about that, it was worth a try adding a new example :-)

(Given that the relevant issue had been open for over a month, with no-one attempting to fix the example, I thus figured that submitting this patch made sense.)

PR #11872, opened 2020-05-04
Gracefully handle annotation parsing errors in `Page.getOperatorList` (issue 11871)

### LINKED ISSUE #11871: Only page one shown of two sided PDF
[bugdoc.pdf](https://github.com/mozilla/pdf.js/files/4575091/bugdoc.pdf)

Configuration:
- Current Chrome, Edge and FF

- Occurs on Win, Mac and Linux:
- PDF.js version: verified on 2.3.200 and 2.5.145
- Is a browser extension: no

1. Open attached document in the viewer online demo.
2. Only page 1 is shown, page 2 stays empty in preview. 

What is the expected behavior? both pages displayed with content.

What went wrong? Looks like the dropdown causes the problems.

Snuffleupagus:
This should ensure that a page will always render successfully, even if there's errors during the Annotation fetching/parsing.
Additionally the `OperatorList.addOpList` method is also adjusted to ignore invalid data, to make it slightly more robust.

Fixes #11871

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -302,11 +302,15 @@ class Page {
         for (const annotation of annotations) {
           if (isAnnotationRenderable(annotation, intent)) {
             opListPromises.push(
-              annotation.getOperatorList(
-                partialEvaluator,
-                task,
-                renderInteractiveForms
-              )
+              annotation
+                .getOperatorList(partialEvaluator, task, renderInteractiveForms)
+                .catch(function (reason) {
+                  warn(
+                    "getOperatorList - ignoring annotation data during " +
+                      `"${task.name}" task: "${reason}".`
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

timvandermeij:
I have not made reference images on purpose because of the broken fonts in Firefox Nightly. I considered changing Puppeteer to use another Firefox version, but I noticed that the problem in https://bugzilla.mozilla.org/show_bug.cgi?id=1607630 has been tracked down, so if a solution is merged in the coming days we'll get it automatically and we can makeref in another patch. If the upstream solution takes too long, we can still switch.

PR #11873, opened 2020-05-05
Use the ESLint `no-restricted-syntax` rule to ensure that `assert` is always called with two arguments

Snuffleupagus:
Having `assert` calls without a message string isn't very helpful when debugging, and it turns out that it's easy enough to make use of ESLint to enforce better `assert` call-sites.
In a couple of cases the `assert` calls were changed to "regular" throwing of errors instead, since that seemed more appropriate.

Please find additional details about the ESLint rule at https://eslint.org/docs/rules/no-restricted-syntax

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -146,6 +146,10 @@
     "no-new-object": "error",
     "no-restricted-syntax": ["error",
       {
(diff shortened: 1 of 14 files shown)


timvandermeij (approved):

timvandermeij:
Looks like a useful addition to me. Thanks!

PR #11879, opened 2020-05-07
Enable the ESLint `grouped-accessor-pairs` rule

Snuffleupagus:
This rule complements the existing `accessor-pairs` nicely, and ensures that a getter/setter pair is always consistently ordered.

Please find additional details about this rule at https://eslint.org/docs/rules/grouped-accessor-pairs


timvandermeij (approved):

timvandermeij:
Nice improvement!

PR #11880, opened 2020-05-07
Attempt to respect the "zoom" hash parameter, even when the "nameddest" parameter is present (issue 11875)

### LINKED ISSUE #11875: Option zoom is ignored if nameddest is set
Configuration:
- Web browser and its version: Chrome (81.0.4044.129) but happens in other browsers too
- PDF.js version: 2.4.456

1. Open viewer with nameddest and zoom param defined in url.
   https://mozilla.github.io/pdf.js/web/viewer.html#zoom=200&nameddest=mydummydest

Viewer should use zoom from url parameter. It would be okay if destination zoom overrides it if ignoreDestinationZoom is set to false but even if no destination zoom is defined or if destination does not exist zoom is not working. 

What went wrong? 

Function setHash from viewer.js returns early if nameddest is defined and zoom is not processed in this case.

```
setHash(hash) {
    let pageNumber, dest;

    if (hash.includes("=")) {
      const params = (0, _ui_utils.parseQueryString)(hash);

      if ("search" in params) {
        this.eventBus.dispatch("findfromurlhash", {
          source: this,
          query: params["search"].replace(/"/g, ""),
          phraseSearch: params["phrase"] === "true"
        });
      }
      if ("nameddest" in params) {
        this.navigateTo(params.nameddest);
        return;
      }

      if ("page" in params) {
        pageNumber = params.page | 0 || 1;
      }

      if ("zoom" in params) {
        const zoomArgs = params.zoom.split(",");
        const zoomArg = zoomArgs[0];
        const zoomArgNumber = parseFloat(zoomArg);

        if (!zoomArg.includes("Fit")) {
          dest = [null, {
            name: "XYZ"
          }, zoomArgs.length > 1 ? zoomArgs[1] | 0 : null, zoomArgs.length > 2 ? zoomArgs[2] | 0 : null, zoomArgNumber ? zoomArgNumber / 100 : zoomArg];
        } else {
          if (zoomArg === "Fit" || zoomArg === "FitB") {
            dest = [null, {
              name: zoomArg
            }];
          } else if (zoomArg === "FitH" || zoomArg === "FitBH" || zoomArg === "FitV" || zoomArg === "FitBV") {
            dest = [null, {
              name: zoomArg
            }, zoomArgs.length > 1 ? zoomArgs[1] | 0 : null];
          } else if (zoomArg === "FitR") {
            if (zoomArgs.length !== 5) {
              console.error('PDFLinkService.setHash: Not enough parameters for "FitR".');
            } else {
              dest = [null, {
                name: zoomArg
              }, zoomArgs[1] | 0, zoomArgs[2] | 0, zoomArgs[3] | 0, zoomArgs[4] | 0];
            }
          } else {
            console.error(`PDFLinkService.setHash: "${zoomArg}" is not ` + "a valid zoom value.");
          }
        }
      }

      if (dest) {
        this.pdfViewer.scrollPageIntoView({
          pageNumber: pageNumber || this.page,
          destArray: dest,
          allowNegativeOffset: true
        });
      } else if (pageNumber) {
        this.page = pageNumber;
      }

      if ("pagemode" in params) {
        this.eventBus.dispatch("pagemode", {
          source: this,
          mode: params.pagemode
        });
      }
    } else {
      dest = unescape(hash);

      try {
        dest = JSON.parse(dest);

        if (!Array.isArray(dest)) {
          dest = dest.toString();
        }
      } catch (ex) {}

      if (typeof dest === "string" || isValidExplicitDestination(dest)) {
        this.navigateTo(dest);
        return;
      }

      console.error(`PDFLinkService.setHash: "${unescape(hash)}" is not ` + "a valid destination.");
    }
  }
```

Not working on initial load
https://mozilla.github.io/pdf.js/web/viewer.html#zoom=200&nameddest=mydummydest
Working in initial load
https://mozilla.github.io/pdf.js/web/viewer.html#zoom=200

If i load the viewer with zoom only and add the nameddest param after this zoom will be processed and document will jumpt to nameddest if it does exist.

Snuffleupagus:
Given that the `PDFLinkService.setHash` method itself if completely synchronous, moving the handling of "nameddest" to occur last *shouldn't* cause any problems (famous last words).
This way the destination will still override any previous parameter, such as e.g. the "page", as expected. Furthermore, given that the `PDFLinkService.navigateTo` method is asynchronous that should provide additional guarantees that the "nameddest" parameter is always respected.

As sort-of expected, this fairly innocent looking change also required some tweaks in the `PDFHistory` to prevent dummy history entires upon document load (only an issue when both "page" *and* "nameddest" parameters are provided in the hash).

Fixes #11875

### DIFF
--- a/web/pdf_history.js
+++ b/web/pdf_history.js
@@ -117,7 +117,9 @@ class PDFHistory {
     this._position = null;
 
     if (!this._isValidState(state, /* checkReload = */ true) || resetHistory) {
-      const { hash, page, rotation } = this._parseCurrentHash();
+      const { hash, page, rotation } = this._parseCurrentHash(
+        /* checkNameddest = */ true
+      );
 
       if (!hash || reInitialized || resetHistory) {
         // Ensure that the browser history is reset on PDF document load.
@@ -490,16 +492,20 @@ class PDFHistory {
   /**
    * @private
    */
-  _parseCurrentHash() {
+  _parseCurrentHash(checkNameddest = false) {
     const hash = unescape(getCurrentHash()).substring(1);
-    let page = parseQueryString(hash).page | 0;
+    const params = parseQueryString(hash);
+
+    const nameddest = params.nameddest || "";
+    let page = params.page | 0;
 
     if (
       !(
         Number.isInteger(page) &&
         page > 0 &&
         page <= this.linkService.pagesCount
-      )
+      ) ||
+      (checkNameddest && nameddest.length > 0)
     ) {
       page = null;
     }
--- a/web/pdf_link_service.js
+++ b/web/pdf_link_service.js
@@ -233,10 +233,6 @@ class PDFLinkService {
         });
       }
       // borrowing syntax from "Parameters for Opening PDF Files"
-      if ("nameddest" in params) {
-        this.navigateTo(params.nameddest);
-        return;
-      }
       if ("page" in params) {
         pageNumber = params.page | 0 || 1;
       }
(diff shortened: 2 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #11889, opened 2020-05-09
Handle errors individually for each annotation in the `_parsedAnnotations` getter

Snuffleupagus:
While working on PR #11872, it occurred to me that it probably wouldn't be a bad idea to change the `_parsedAnnotations` getter to handle errors individually for each annotation. This way, one broken/corrupt annotation won't prevent the rest of them from being e.g. fetched through the API.


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

Issue #11858, opened 2020-04-29
No minified build for older browsers from 2.4.456


kleins05:
There does not seem to be a way to build a minified version for older browsers. 

I used to use `gulp minified` as stated in the [FAQ](https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#the-pdfjs-files-are-too-big-is-it-possible-to-obtain-minified-versions-of-the-js-files)

Since 2.4.456 produces a non-polyfilled build by default (SKIP_BABEL=true in gulpfile.js) and since we need to support IE11, this does not work anymore.

Whilst there is a gulp task _generic-es5_ to complement the _generic_ task, the same is missing for the _minified_ task. Is there another easy way to build a minified version for older browsers without modifying gulpfile.js? If not, would you accept a pull request, if I added it?

The FAQ should be updated, too, to mention that the minified task builds a modern browser version.

Snuffleupagus:
> I used to use `gulp minified` as stated in the [FAQ](https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#the-pdfjs-files-are-too-big-is-it-possible-to-obtain-minified-versions-of-the-js-files)

When the `-es5` build tasks were added, the main focus was on things needed for the builds that are published at https://github.com/mozilla/pdf.js/releases respectively https://github.com/mozilla/pdfjs-dist

> Is there another easy way to build a minified version for older browsers without modifying gulpfile.js?

No easy way currently, as far as I know.

> If not, would you accept a pull request, if I added it?

Well, that probably depends rather heavily on *exactly* what sort of solution you have in mind here.

Adding an *almost* verbatim copy of the existing `minified` task would be a non-started, and the only acceptable solution would be one that re-factors the existing build tasks such that the code is re-used between the `minified` and (new) `minified-es5` tasks. 

> The FAQ should be updated, too, to mention that the minified task builds a modern browser version.

Yes, that probably makes sense.

timvandermeij:
I agree with the above. Having a new task `minified-es5` would indeed be acceptable if the existing code is re-used. You can take a look at the existing `-es5` tasks in the Gulpfile for inspiration. If such a solution can be made, we'll definitely review it.

Snuffleupagus:
Having looked *briefly* at the relevant code, this is almost trivial to implement and done right the net increase is only ~10 lines of code; I might submit a patch at some point (time permitting).

stephanrauh:
[...]

PR #11891, opened 2020-05-10
#11858 Minified build for older browsers called minified-es5

nkimadusanka:
**Developer Comment**

- Add new gulp task to support older browsers.
- New gulp taks name as `minified-es5`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -943,12 +943,12 @@ gulp.task(
 
 gulp.task("minified", gulp.series("minified-post"));
 
-gulp.task("minifined-es5-config", function(done) {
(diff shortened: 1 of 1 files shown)


Snuffleupagus (changes requested):
Sorry, but this is unfortunately *not* at all the correct solution here.

First of all, you are not allowed to change the *default* values of the build targets like this (note how the rest of this file uses `DEFINES`).
Second of all, this outputs the `minified`/`minified-es5` files in the *same* folder thus overwriting any previous one and without providing *any* indication what sort of build the folder actually contains.
Finally, this doesn't follow the general re-use pattern of tasks for modern/es5-compatible builds.

PR #11892, opened 2020-05-10
Add a `minified-es5` gulp task (issue 11858)

### LINKED ISSUE #11858: No minified build for older browsers from 2.4.456

Snuffleupagus:
-  Ensure that the `DEFINES` build target constants, in `gulpfile.js`, cannot be changed

 - Add a `minified-es5` gulp task (issue 11858)

   By re-factoring the existing gulp tasks, most of the code can be re-used for both the existing `gulp minified` as well as the new `gulp minified-es5` task.

Fixes #11858

*Smaller diff with https://github.com/mozilla/pdf.js/pull/11892/files?w=1*

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -57,6 +57,7 @@ var COMPONENTS_ES5_DIR = BUILD_DIR + "components-es5/";
 var IMAGE_DECODERS_DIR = BUILD_DIR + "image_decoders";
 var DEFAULT_PREFERENCES_DIR = BUILD_DIR + "default_preferences/";
 var MINIFIED_DIR = BUILD_DIR + "minified/";
+var MINIFIED_ES5_DIR = BUILD_DIR + "minified-es5/";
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice addition!

PR #11905, opened 2020-05-14
Reduce the usage of `require` statements in code-paths not protected by pre-processor and/or run-time checks

Snuffleupagus:
This replaces some additional `require`/`exports` usage with standard `import`/`export` statements instead.
Hence another, small, part in the effort to reduce the reliance on SystemJS-specific functionality in the development viewer.

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -678,7 +678,7 @@ function isMessagePort(maybePort) {
   );
 }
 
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thank you!

PR #11907, opened 2020-05-15
Minimal attempt to fix #8408

### LINKED ISSUE #8408: Text-selection error in double-column PDF
[Nat Struct Mol Biol 2008 Zenklusen.pdf](https://github.com/mozilla/pdf.js/files/975293/Nat.Struct.Mol.Biol.2008.Zenklusen.pdf)

Configuration:
- PDF.js version: 1.7.225
- Is an extension: No

1. double click any word at the end of a line

image

image

image

image

JZ-wenxuan:
Added a line break after every text element to prevent Chrome from thinking that the lines are connected.
I'm not familiar with this project so I'm not sure if this is a good implementation, but #8408 should be easy to solve with any similar idea.

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -607,6 +607,9 @@ var renderTextLayer = (function renderTextLayerClosure() {
       }
       this._textDivProperties.set(textDiv, textDivProperties);
       this._container.appendChild(textDiv);
+      if (window.chrome) {
+        this._container.appendChild(document.createElement("br"));
+      }
     },
 
     _render: function TextLayer_render(timeout) {


src/display/text_layer.js
```diff
@@ -607,6 +607,9 @@ var renderTextLayer = (function renderTextLayerClosure() {
       }
       this._textDivProperties.set(textDiv, textDivProperties);
       this._container.appendChild(textDiv);
+      if (window.chrome) {
```
Snuffleupagus, line 610: Generally speaking, we don't want these sort of browser-specific compatibility hacks in the main code-base for reasons of e.g. readability/maintainability/performance.

Snuffleupagus (changes requested):
Sorry, but this is unfortunately not an acceptable solution!

Besides the inline comment, this patch will actually cause *much* worse text-selection behaviour in many documents (e.g. even in the `tracemonkey.pdf` file) by introducing *a lot* of additional line-breaks where they don't belong (i.e. in the middle of lines, whenever the font changes).

One of the reasons that we changed from `div`s to `span`s was to reduce unnecessary line-breaks in the middle of text when copying, and this patch would essentially regress that in Chromium-based browsers.

All-in-all, text-selection is a very complex topic and a solution like this one unfortunately isn't close to viable.

Issue #11612, opened 2020-02-17
PDFs containing pages with high resolution TIFFs render slowly


zxqx:
**Attach (recommended) or Link to PDF file here**: [TEST.pdf](https://github.com/mozilla/pdf.js/files/4216176/TEST.pdf) - a contrived document containing pages with high resolution TIFFs (6592x5100)

**Configuration**:

- Web browser and its version: Chrome 80.0.3987.87
- Operating system and its version: macOS Mojave (10.14.6)
- PDF.js version: v2.3.200 _and_ latest unstable
- Is a browser extension: No

**Steps to reproduce the problem**:
1. Upload [TEST.pdf](https://github.com/mozilla/pdf.js/files/4216176/TEST.pdf) using the [Mozilla PDF.js demo](https://mozilla.github.io/pdf.js/web/viewer.html)
2. Scroll down to the last page

**What is the expected behavior?** The last page renders in a reasonable amount of time.

**What went wrong?** The last page takes ~15-30 seconds to render.

![Screen Shot 2020-02-17 at 3 12 20 PM](https://user-images.githubusercontent.com/1393139/74686728-02148700-5198-11ea-9dad-35eb6ea3b212.png)

Snuffleupagus:
Rendering times improved significantly with PR #17946 (almost 50% in my quick tests), which combined with the pre-existing caching of global image-data means that we've done everything that we realistically can to fix this.

Issue #11518, opened 2020-01-17
PDF loading is very slow and hangs


learnyst:
Attach (recommended) or Link to PDF file here:
[set-2_lyst7242.pdf](https://github.com/mozilla/pdf.js/files/4076920/set-2_lyst7242.pdf)

Configuration:
- Web browser and its version:
Chrome browser, version 79

- Operating system and its version:
Mac OS, Mojave 10.14
Also same behavior in windows 10

- PDF.js version: Using the hosted demo

- Is a browser extension: No

Steps to reproduce the problem:
1. Just open the PDF in hosted demo

What is the expected behavior? (add screenshot)
PDF should load without hanging

What went wrong? (add screenshot)
PDF loading is very very slow and hangs

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

learnyst:
@timvandermeij is there any workaround to solve this performance issue?

claudiawisniewski:
I experience the same Issue on Ubuntu 18.04.1
![2020-01-24 14-44-32](https://user-images.githubusercontent.com/56873264/73080613-66f4fe00-3ec6-11ea-9da4-96f23dfb9052.png)

claudiawisniewski:
I'm experiencing the same Issue on Ubuntu 18.04.1 with pdfjs-dist version: "2.0.943"
![2020-01-24 14-44-32](https://user-images.githubusercontent.com/56873264/73080958-ff8b7e00-3ec6-11ea-959a-4109f3d13615.png)

amit777:
This PDF loaded pretty slow on my macbook air circa 2015 using Preview.  I loaded it in the latest PDFJS, and it seems like similar performance (if not even slightly better).
```
const pdfjsVersion = '2.7.90';
const pdfjsBuild = '6728c8fa6';
```

PR #11912, opened 2020-05-18
Attempt to cache repeated images at the document, rather than the page, level (issue 11878)

### LINKED ISSUE #11878: Render issue when scrolling PDFs with large JPEGs
**Attach (recommended) or Link to PDF file here:**
[jpeg-scroll-issue.pdf](https://github.com/mozilla/pdf.js/files/4587118/jpeg-scroll-issue.pdf)

**Configuration:**
- Web browser and its version: Chrome 81.0.4044.138 (Official Build) (64-bit)
- Operating system and its version: macOS 10.15.4 (19E287) 
- PDF.js version: commit d822578450262db7290e2f33f14c6823b7140ad6 of this repo

1. Open attached PDF
2. Scroll down to the middle of document
3. Observe page rendering

**What is the expected behavior?** 
When scrolling, the subsequent pages should render relatively quickly.

**What went wrong?**
The page is blank for awhile before the pages render.

## Additional Details

I believe that after this commit:
https://github.com/mozilla/pdf.js/commit/62a9c26cdae9118dea89dc18b5cad844bbc7faf5

The PDF attached starts to exhibit the described behaviour. When I revert that specific commit and test again, the render of the image is almost instantaneous.

I noticed that with the changes in the above commit, the chrome dev tools profiler shows an uptick of this routine that seems to be consuming the worker when scrolling in the document. 

Screen Shot 2020-05-06 at 10 20 33 AM

Snuffleupagus:
Currently image resources, as opposed to e.g. font resources, are handled exclusively on a page-specific basis. Generally speaking this makes sense, since pages are separate from each other, however there's PDF documents where many (or even all) pages actually references exactly the image resources (through the XRef table). Hence, in some cases, we're decoding the *same* images over and over for every page which is obviously slow and wasting both CPU and memory resources better used elsewhere.[1]

Obviously we cannot simply treat all image resources as-if they're used throughout the entire PDF document, since that would end up increasing memory usage too much.[2]
However, by introducing a `GlobalImageCache` in the worker we can track image resources that appear on more than one page. Hence we can switch image resources from being page-specific to being document-specific, once the image resource has been seen on more than a certain number of pages.

In many cases, such as e.g. the referenced issue, this patch will thus lead to reduced memory usage for image resources. Scrolling through all pages of the document, there's now only a few main-thread copies of the same image data, as opposed to one for each rendered page (i.e. there could theoretically be *twenty* copies of the image data).
While this obviously benefit both CPU and memory usage in this case, for *very* large image data this patch *may* possibly increase persistent main-thread memory usage a tiny bit. Thus to avoid negatively affecting memory usage too much in general, particularly on the main-thread, the `GlobalImageCache` will *only* cache a certain number of image resources at the document level and simply fallback to the default behaviour.

Unfortunately the asynchronous nature of the code, with ranged/streamed loading of data, actually makes all of this much more complicated than if all data could be assumed to be immediately available.[3]

*Please note:* The patch will lead to *small* movement in some existing test-cases, since we're now using the built-in PDF.js JPEG decoder more. This was done in order to simplify the overall implementation, especially on the main-thread, by limiting it to only the `OPS.paintImageXObject` operator.

Fixes #11878 (and probably a few more issues/bugs).

Also *slightly* improves cases such e.g. issue #11518, issue #11612, and [bug 1536420](https://bugzilla.mozilla.org/show_bug.cgi?id=1536420)

---
[1] There's e.g. PDF documents that use the same image as background on all pages.

[2] Given that data stored in the `commonObjs`, on the main-thread, are only cleared manually through `PDFDocumentProxy.cleanup`. This as opposed to data stored in the `objs` of each page, which is automatically removed when the page is cleaned-up e.g. by being evicted from the cache in the default viewer.

[3] If the latter case were true, we could simply check for repeat images *before* parsing started and thus avoid handling *any* duplicate image resources.

### DIFF
--- a/src/core/image_utils.js
+++ b/src/core/image_utils.js
@@ -14,8 +14,10 @@
  */
 /* eslint no-var: error */
 
+import { assert, info, shadow } from "../shared/util.js";
 import { ColorSpace } from "./colorspace.js";
 import { JpegStream } from "./jpeg_stream.js";
+import { RefSetCache } from "./primitives.js";
 import { Stream } from "./stream.js";
 
 class NativeImageDecoder {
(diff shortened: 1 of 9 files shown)


Snuffleupagus:
Landing *this* PR before considering moving forward with PR #11601 would probably simplify things considerably, since it would reduce the number of test-failures in that one quite a lot.

src/core/image_utils.js
```diff
@@ -111,4 +113,99 @@ class NativeImageDecoder {
...
+const NUM_PAGES_THRESHOLD = 2;
+const MAX_IMAGES_TO_CACHE = 10;
+
+const GlobalImageCacheKind = {
+  NOT_CACHED: 1,
+  SHOULD_CACHE: 2,
+  IS_CACHED: 3,
+};
+
+class GlobalImageCache {
```
timvandermeij, line 125: Given that this class is important for correct image rendering, I would like to see unit tests being added for it. I think it's also the only way to get some test coverage for this patch and it should be relatively easy to test this class since we can get/set entries and assert that they are handled correctly.

Snuffleupagus: > I would like to see unit tests being added for it.

Well, I suppose that's a fair point; I'll see what I can come up with here :-)

> I think it's also the only way to get some test coverage for this patch

As mentioned in the commit message, since we're now handling repeated JPEG images differently there's *some* amount of test coverage already.
Although I suppose that complements, rather than replaces, dedicated unit-tests though...

Snuffleupagus: I've added some (high-level) unit-tests to ensure that the API receives the expected data, which I hope is useful as a complement to the reference tests. (I'm currently a bit short on time unfortunately.)

Snuffleupagus: I've added some (high-level) unit-tests to ensure that the API receives the expected image data, which I hope is useful as a complement to the reference tests. (I'm currently a bit short on time unfortunately.)

timvandermeij (approved):

timvandermeij:
Really nice work! The unit test looks good.

/botio makeref

Snuffleupagus:
> Really nice work! The unit test looks good.

Thanks for landing this; I'm really hoping that the changes will result in user-perceived improvements in these kind of PDF documents.

After this landed (obviously), I've also found a possible improvement related to cleanup; please see PR #11926 for additional details.

Finally, just yesterday, I've also realized that the old pre-existing *page*-level [`imageCache`](https://github.com/mozilla/pdf.js/blob/973f39b55816a4f2d50cd8acb09dc9c9ba3b7ad4/src/core/evaluator.js#L1307) is possibly missing some repeated *page*-level images, since the caching is done only by name (e.g. of the format "Im0", "Im1", ...). In some cases there's PDF documents where one page contains *hundreds* of distinctly named images, despite them referring (via references) to only a handful of actual image objects. I'll look into fixing that as well, possibly with a new `LocalImageCache` class, since it seems somewhat ridiculous to re-parse *any* image over and over.

PR #11914, opened 2020-05-19
Convert the `src/pdf.js` and `src/pdf.worker.js` files to use standard `import`/`export` statements

Snuffleupagus:

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -65,7 +65,8 @@ const RENDERING_CANCELLED_TIMEOUT = 100; // ms
  * @typedef {function} IPDFStreamFactory
  * @param {DocumentInitParameters} params - The document initialization
  *   parameters. The "url" key is always present.
- * @returns {IPDFStream}
+ * @returns {Promise} A promise, which is resolved with an instance of
+ *   {IPDFStream}.
  * @ignore
  */
 
@@ -80,7 +81,7 @@ let createPDFNetworkStream;
  * data transport.
  * @param {IPDFStreamFactory} pdfNetworkStreamFactory - The factory function
  *   that takes document initialization parameters (including a "url") and
- *   returns an instance of {IPDFStream}.
+ *   returns a promise which is resolved with an instance of {IPDFStream}.
  * @ignore
  */
 function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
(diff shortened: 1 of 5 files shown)


Snuffleupagus:
After this PR, the only thing preventing us from completely removing SystemJS is still https://bugzilla.mozilla.org/show_bug.cgi?id=1247687

However, we could actually limit SystemJS usage to only the worker and unit-tests immediately, thanks to an `import-maps` polyfill; please see https://github.com/mozilla/pdf.js/commit/714ad6a717639e61ca4b632824a344f346e4ad3b

src/display/api.js
```diff
@@ -313,52 +314,60 @@ function getDocument(src) {
...
+      });
+
+      return Promise.all([workerIdPromise, networkStreamPromise]).then(data => {
+        if (task.destroyed) {
+          throw new Error("Loading aborted");
+        }
+
+        const messageHandler = new MessageHandler(
+          docId,
+          /* workerId = */ data[0],
```
timvandermeij, line 358: Can we use array destructuring for `data` above so we don't need the inline comments here and below to make it self-documenting? Something like `.then(([workerId, networkStream]) => {` should work I think.

src/display/api.js
```diff
@@ -80,7 +81,7 @@ let createPDFNetworkStream;
  * data transport.
  * @param {IPDFStreamFactory} pdfNetworkStreamFactory - The factory function
  *   that takes document initialization parameters (including a "url") and
- *   returns an instance of {IPDFStream}.
+ *   returns a promise which is resolved with an instance of {IPDFStream}.
```
timvandermeij, line 84: Isn't this patch `[api-minor]` since the `setPDFNetworkStreamFactory` is exposed in the public API, or will old-style `pdfNetworkStreamFactory` parameters still work?

Snuffleupagus: > Isn't this patch `[api-minor]` since the `setPDFNetworkStreamFactory` is exposed in the public API,

It's not actually part of the public API though, given that it's not being exported by the https://github.com/mozilla/pdf.js/blob/master/src/pdf.js file.

> or will old-style `pdfNetworkStreamFactory` parameters still work?

Any existing code should "just" work, note also that I didn't re-write any of the existing call-sites in `src/pdf.js` either.

src/display/api.js
Snuffleupagus: Fair point, consider it done :-)

timvandermeij:
Looks good after the comment above are addressed.

src/display/api.js
timvandermeij: Whoops, I accidentally looked at the exports of `api.js`... Yes, this can be ignored then.

timvandermeij (approved):

timvandermeij:
Yet another step closer to less SystemJS usage. Thank you for working on this!

PR #11919, opened 2020-05-20
Reduce usage of SystemJS, in the development viewer, even further

### LINKED ISSUE #10965: Upgrade `systemjs` to the most recent version

Snuffleupagus:
With these changes SystemJS is now only used, during development, on the worker-thread and in the unit/font-tests, since Firefox is currently missing support for worker modules; please see https://bugzilla.mozilla.org/show_bug.cgi?id=1247687

Hence all the JavaScript files in the `web/` and `src/display/` folders are now loaded *natively* by the browser (during development) using standard `import` statements/calls, thanks to a nice `import-maps` polyfill.

*Please note:* As soon as https://bugzilla.mozilla.org/show_bug.cgi?id=1247687 is fixed in Firefox, we should be able to remove all traces of SystemJS and thus finally be able to use every possible modern JavaScript feature.

---

Depends upon, and includes, PR #11914 

Fixes #10965 (given that it doesn't make much sense to spend time and effort on that now, since SystemJS is now used a lot less and that it'll hopefully be gone completely somewhat soon).

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1650,7 +1650,9 @@ const PDFWorker = (function PDFWorkerClosure() {
       }
       if (typeof PDFJSDev === "undefined" || !PDFJSDev.test("PRODUCTION")) {
         if (typeof SystemJS !== "object") {
-          throw new Error("SystemJS must be used to load fake worker.");
+          // Manually load SystemJS, since it's only necessary for fake workers.
+          await loadScript("../node_modules/systemjs/dist/system.js");
+          await loadScript("../systemjs.config.js");
         }
         const worker = await SystemJS.import("pdfjs/core/worker.js");
         return worker.WorkerMessageHandler;
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,6 @@
 {
   "parserOptions": {
-    "ecmaVersion": 8,
+    "ecmaVersion": 2020,
     "sourceType": "module",
   },
 
@@ -17,13 +17,12 @@
 
   "env": {
     "browser": true,
-    "es6": true,
+    "es2020": true,
     "worker": true,
     "amd": true,
   },
 
   "globals": {
-    "globalThis": false,
     "PDFJSDev": false,
     "exports": false,
     "SystemJS": false,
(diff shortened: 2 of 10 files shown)


Snuffleupagus:
Most of the testing of this patch will unfortunately need to be done locally, and manually, since the changes shouldn't affect the built files in any way...

/botio test

src/display/api.js
```diff
@@ -1650,7 +1650,9 @@ const PDFWorker = (function PDFWorkerClosure() {
       }
       if (typeof PDFJSDev === "undefined" || !PDFJSDev.test("PRODUCTION")) {
         if (typeof SystemJS !== "object") {
-          throw new Error("SystemJS must be used to load fake worker.");
+          // Manually load SystemJS, since it's only necessary for fake workers.
+          await loadScript("../node_modules/systemjs/dist/system.js");
```
timvandermeij, line 1654: Shouldn't this be `../../` instead of `../` since, contrary to the same change in the `web` folder, this file is inside another subfolder?

Snuffleupagus: I'd be inclined to agree with you here, but interestingly the patch as written actually works (when this code-path is forced, since the development viewer won't generally hit it).

The explanation would rather be that when this file is loaded, it's done through the `web/viewer.html` file which does seem to end up being the effective "scope" used here.

Finally, this is fallback code which should *never* be hit in development mode and this is only necessary until https://bugzilla.mozilla.org/show_bug.cgi?id=1247687 is fixed.

Snuffleupagus:
It seems that https://bugzilla.mozilla.org/show_bug.cgi?id=1607630 has landed in Nightly now, let's see if this works...

/botio makeref

timvandermeij (approved):

timvandermeij:
Seems to work just fine locally. Thanks!

PR #11927, opened 2020-05-24
Implement fill opacity for shading patterns in the SVG back-end

### LINKED ISSUE #11812: Rendering PDF which include shapes who have opacity (SVG renderer)
I attached the svg as txt because of the github upload restrictions.
[page3.txt](https://github.com/mozilla/pdf.js/files/4481235/page3.txt)
[ton_page3.pdf](https://github.com/mozilla/pdf.js/files/4481237/ton_page3.pdf)

Configuration:
Web browser and its version: Latest Chrome (80.0.3987.149)
PDF.js version: pdfjs-dist (2.4.456)

Use the PDF above in the SVG renderer and you will see that image in the middle has two cirlces. the right circle should have 3 circles around them that each has less opacity. In the SVG renderer these circles are rendered with full opacity.

timvandermeij:
In the PDF file from the issue below, the fill alpha (`ca`) is set before drawing the circles using the `setGState` operator. Doing so causes the global alpha to be set on the canvas' context for the canvas back-end, but this was not handled in the SVG back-end. This patch fixes that by taking the fill opacity into account when drawing shading patterns in the same way as done elsewhere so it is only included if the value is non-default.

Fixes #11812.

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -1090,6 +1090,9 @@ if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) {
       rect.setAttributeNS(null, "width", x1 - x0);
       rect.setAttributeNS(null, "height", y1 - y0);
       rect.setAttributeNS(null, "fill", this._makeShadingPattern(args));
+      if (this.current.fillAlpha < 1) {
+        rect.setAttributeNS(null, "fill-opacity", this.current.fillAlpha);
+      }
       this._ensureTransformGroup().appendChild(rect);
     }
 


timvandermeij:
You can test this by opening the preview viewer below and running `PDFViewerApplication.preferences.set('renderer', 'svg');` in the console. If you refresh the page and open the file from the issue, you'll see three circles with different opacities now.

Snuffleupagus (approved):
Looks good to me, thanks!

*Super tiny nit:* Shouldn't it be `ca`, i.e. lower-case, in the commit message since the upper-case variant is  stroke-related based on e.g. https://github.com/mozilla/pdf.js/blob/3b615e4ca369ff31e6399dea80383ffeffe96240/src/display/canvas.js#L1005-L1011
This confused me for a second when reading the commit message, before looking at the actual code.

timvandermeij:
Good point; I have amended the commit and PR message. Thanks!

Snuffleupagus:
> Good point; I have amended the commit and PR message.

Thank you; as an aside you do have to wonder why anyone thought it'd be a good idea to specify two separate operators with the same name and differing *only* in their casing...

PR #11929, opened 2020-05-24
Transfer partial page-fit functions from zoom to view parameter in URL

huang325:
This is the follow up fix for #10773  as the [previous pull request](https://github.com/mozilla/pdf.js/pull/11786) is closed.
Changes are eliminated based on former feedback.


huang325:
I still keep "FitR" in `zoom` as it's not included in the documentation of `view` and I'm not sure how it is described in other standards.

Issue #4958, opened 2014-06-17
Extremely slow loading steam machine brochure


samlh:
http://media.steampowered.com/store/steammachines/SteamMachinesBroc_WEB.PDF

I ran out of patience a couple minutes in.

Firefox 31.0a2 (2014-06-09)
"PDF f9d89da5f9d426088edf86f954afe29 [1.3 Adobe PDF Library 10.0.1 / Adobe InDesign CS6 (Macintosh)](PDF.js: 1.0.374)"

samlh:
Happily, the pdf eventually loads now. Firefox tab peak memory usage ~3GB. The speed is also much improved. Closing.

PR #11930, opened 2020-05-25
Improve the *local* image caching in `PartialEvaluator.getOperatorList`

Snuffleupagus:
Currently the local `imageCache`, as used in `PartialEvaluator.getOperatorList`, will miss certain cases of repeated images because the caching is *only* done by name (usually using a format such as e.g. "Im0", "Im1", ...).
However, in some PDF documents the `/XObject` dictionaries many contain hundreds (or even thousands) of distinctly named images, despite them referring to only a handful of actual image objects (via the XRef table).

With these changes we'll now cache *local* images using both name and (where applicable) reference, thus improving re-usage of images resources even further.

This patch was tested using the PDF file from [bug 857031](https://bugzilla.mozilla.org/show_bug.cgi?id=857031), i.e. https://bug857031.bmoattachments.org/attachment.cgi?id=732270, with the following manifest file:
```
[
    {  "id": "bug857031",
       "file": "../web/pdfs/bug857031.pdf",
       "md5": "",
       "rounds": 250,
       "lastPage": 1,
       "type": "eq"
    }
]
```

```
-- Grouped By browser, page, stat --
browser | page | stat         | Count | Baseline(ms) | Current(ms) | +/- |    %  | Result(P<.05)
------- | ---- | ------------ | ----- | ------------ | ----------- | --- | ----- | -------------
firefox | 0    | Overall      |   250 |         2749 |        2656 | -93 | -3.38 |        faster
firefox | 0    | Page Request |   250 |            3 |           4 |   1 | 50.14 |        slower
firefox | 0    | Rendering    |   250 |         2746 |        2652 | -94 | -3.44 |        faster
```

While this is certainly an improvement, since we now avoid re-parsing ~1000 images on the first page, all of the image resources are small enough that the total rendering time doesn't improve that much in this particular case.

In pathological cases, such as e.g. the PDF document in issue #4958, the improvements with this patch can be very significant. Looking for example at page 2, from issue #4958, the rendering time drops from ~60 seconds with `master` to ~30 seconds with this patch (obviously still slow, but it really showcases the potential of this patch nicely).

Finally, note that there's also potential for additional improvements by re-using `LocalImageCache` instances for e.g. /XObject data of the `Form`-type. However, given that recent changes in this area I purposely didn't want to complicate *this* patch more than necessary.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1724,7 +1724,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
 
       // The xobj is parsed iff it's needed, e.g. if there is a `DO` cmd.
       var xobjs = null;
-      var skipEmptyXObjs = Object.create(null);
+      const emptyXObjectCache = new LocalImageCache();
 
       var preprocessor = new EvaluatorPreprocessor(stream, xref, stateManager);
 
@@ -2200,7 +2200,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
               }
 
               var name = args[0].name;
-              if (name && skipEmptyXObjs[name] !== undefined) {
+              if (name && emptyXObjectCache.getByName(name)) {
                 break;
               }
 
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks a lot!

Snuffleupagus:
> Looks good; thanks a lot!

Thanks for landing this; hopefully the recent image-caching improvements will help reduce both rendering times *and* memory usage in lots of PDF files :-)

timvandermeij:
I have planned to do that on Monday since, if for example the NPM sync doesn't work, @brendandahl might be able to do that manually on Tuesday for example.

PR #11940, opened 2020-05-27
Add comments to the `export` list in the `src/pdf.js` file (PR 11914 follow-up)

Snuffleupagus:
When converting this file to use standard `import`/`export` statements, I sorted the exports in the same order as the imports to simplify things.

However, looking at the list of `export`ed properties it probably doesn't hurt to add a couple of comments to clarify from where specifically the `export`s originated.


timvandermeij (approved):

timvandermeij:
Good idea! It's indeed quite a big list.

PR #11947, opened 2020-05-31
Ensure that that we don't attempt to cache *inline* images in the `GlobalImageCache` (PR 11912 follow-up)

Snuffleupagus:
Since *inline* images, i.e. those defined inside of `/Contents` streams, are by their very definition page-specific it thus seem like a good idea to actually enforce that they won't accidentally end up in the `GlobalImageCache`.


timvandermeij (approved):

timvandermeij:
Thanks!

Issue #11838, opened 2020-04-22
Only trigger fallback bar for forms when user interacts with page


brendandahl:
We're investigating ways to show the fallback bar less. One option is only show the bar when someone clicks within a document.

timvandermeij:
I wondered how many times the fallback bar is still shown (since I barely see it anymore myself) and what the reasons for that are, and fortunately the new telemetry helps a lot here. The results [over here](https://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2020-05-30&include_spill=0&keys=__none__!__none__!__none__&max_channel_version=nightly%252F78&measure=PDF_VIEWER_FALLBACK_REASON&min_channel_version=null&processType=*&product=Firefox&sanitize=1&sort_by_value=0&sort_keys=submissions&start_date=2020-05-20&table=0&trim=1&use_submission_date=0) are quite interesting and show that it's probably beneficial to make this change. It also gives some insights into other changes that might help here for other bars in the chart.

bwinton:
Two questions:
1) How hard would it be to only show the fallback bar if the user tries to interact with the form (instead of clicking anywhere in the PDF)?
2) Will this also show the fallback bar for people who use accessibility technologies, and try to tab to the form fields?

timvandermeij:
Fixed in the pull request above.

PR #11953, opened 2020-06-01
For #11838: trigger fallback bar after user clicks in pdf

emalysz:


web/app.js
```diff
@@ -868,14 +869,17 @@ const PDFViewerApplication = {
       PDFJSDev.test("MOZCENTRAL || GENERIC")
     ) {
       // Only trigger the fallback once so we don't spam the user with messages
-      // for one PDF.
-      if (this.fellback) {
+      // for one PDF. It should only be triggered once the user has clicked
+      // within the PDF.
+      if (!PDFViewerApplication._hasInteracted) {
```
Snuffleupagus, line 874: Unnecessary `PDFViewerApplication` here, could just be `this` instead.

web/app.js
```diff
@@ -868,14 +869,17 @@ const PDFViewerApplication = {
...
+        if (!this._featureId) {
+          this._featureId = featureId;
+        }
         return;
       }
-      this.fellback = true;
```
Snuffleupagus, line 875: Unless I'm mistaken, with these changes you no longer prevent the fallback bar from being triggered *multiple* times (so the comment above no longer holds).

It's probably possible to come up with something better implementation wise than what I'm suggesting below, but it seems that you may still need the `this.fellback` boolean (or similar) and that something like the following is closer to what you probably want here (at least from a functionality point-of-view):
```javascript
if (!this._featureId) { <-- I'm assuming that you want to still capture the featureId value from the *first* invocation of this method.
  this._featureId = featureId;
}
if (!this._hasInteracted) {
  return;
}
if (this.fellback) {
  return;
}
this.fellback = true;
this.externalServices.fallback(
  // The rest of the method here...
```

web/app.js
```diff
@@ -2473,6 +2477,8 @@ function webViewerWheel(evt) {
 }
 
 function webViewerClick(evt) {
+  PDFViewerApplication._hasInteracted = true;
+  PDFViewerApplication.fallback();
```
Snuffleupagus, line 2481: Nit: Maybe a blank line here, since the above is unrelated to the code below?

timvandermeij:
Moreover, it looks like this happens for every type of error. Shouldn't we limit this to just form errors for now? I think there are also types of errors that for example prevent pages from being shown or being rendered correctly, and I think that's something you want to be notified about regardless of whether the user clicks or not.

emalysz:
@timvandermeij @Snuffleupagus Thanks for the review! 

So looking at https://www.irs.gov/pub/irs-pdf/f1040.pdf as an example, it would now display the fallback bar with the new error message "The PDF document may not be displayed correctly" due to a font error. It would no longer display the "This PDF contains forms..." message, but the fallback bar is still present.

Is this desired behavior?

Snuffleupagus:
> [...]
> Is this desired behavior?

Good point, probably not. Taking https://github.com/mozilla/pdf.js/pull/11953#issuecomment-637131180 into account while also fixing the other issues (with the fallback bar showing multiple times) seems a bit trickier than you'd expect.

Hopefully I'm not overstepping here, but would something like the following work?

```diff
diff --git a/web/app.js b/web/app.js
index 1d442f0a..550cce69 100644
--- a/web/app.js
+++ b/web/app.js
@@ -189,6 +189,7 @@ const PDFViewerApplication = {
   externalServices: DefaultExternalServices,
   _boundEvents: {},
   contentDispositionFilename: null,
+  _delayedFallbackFeatureId: null,
 
   // Called once when the document is loaded.
   async initialize(appConfig) {
@@ -867,6 +868,11 @@ const PDFViewerApplication = {
       typeof PDFJSDev === "undefined" ||
       PDFJSDev.test("MOZCENTRAL || GENERIC")
     ) {
+      // Comment about forms here...
+      if (this._delayedFallbackFeatureId) {
+        featureId = this._delayedFallbackFeatureId;
+        this._delayedFallbackFeatureId = null; <-- Reset this to prevent (possibly) every single click event from trying to show the fallback bar.
+      }
       // Only trigger the fallback once so we don't spam the user with messages
       // for one PDF.
       if (this.fellback) {
@@ -1317,7 +1323,7 @@ const PDFViewerApplication = {
 
     if (info.IsAcroFormPresent) {
       console.warn("Warning: AcroForm/XFA is not supported");
-      this.fallback(UNSUPPORTED_FEATURES.forms);
+      this._delayedFallbackFeatureId = UNSUPPORTED_FEATURES.forms;
     }
 
     if (
@@ -2473,6 +2479,13 @@ function webViewerWheel(evt) {
 }
 
 function webViewerClick(evt) {
+  if (
+    PDFViewerApplication._delayedFallbackFeatureId &&
+    PDFViewerApplication.pdfViewer.containsElement(evt.target) <-- Avoid triggering the fallback bar when the user clicks on e.g. the toolbar or sidebar.
+  ) {
+    PDFViewerApplication.fallback();
+  }
+
   if (!PDFViewerApplication.secondaryToolbar.isOpen) {
     return;
   }
```

emalysz:
@Snuffleupagus I combined our approaches to make sure that we indicate when the click has happened with '_hasInteracted'. Without that, I don't believe it would have returned early. After a discussion with the pdf team, it was decided thatif it's a form and script error message, then we delay the fallback bar until after the click. For other errors, we should immediately display.

Snuffleupagus:
> it was decided thatif it's a form and script error message, then we delay the fallback bar until after the click.

But doesn't this mean that unless *both* forms/javascript entries exists (which isn't generally guaranteed), the fallback bar won't actually display with this patch; is that actually intentional?

For some PDF documents, this thus means that we no longer show the fallback bar at all, right?

---

Basically, my question is if it shouldn't have been forms OR javascript, and the conditions thus `>= 1` instead!?

Finally, note that with this patch the fallback bar message will also change, in the way described in https://github.com/mozilla/pdf.js/pull/11953#issuecomment-637182713, given the order of which the `PDFViewerApplication.fallback` method is called (since all call-sites are in asynchronous code).

emalysz:
@Snuffleupagus Thanks!

Yeah unfortunately we'd still be displaying a fallback bar for the font error. But in the event that a font error drastically changes the appearance of a form pdf, it seems reasonable to show this bar regardless of a click.

emalysz:
https://github.com/mozilla/pdf.js/issues/11838#issuecomment-638206193 mentions accessibility risks. I consulted with the accessibility team, and they thought that we should trigger the fallback bar if the user clicks or tabs in. I have added a new event handler for Tab key presses.

brendandahl:
@Snuffleupagus is this good to go now?

Snuffleupagus:
> @Snuffleupagus is this good to go now?

Yes, I think so. (There's probably some potential for simplification, especially once #11969 has landed as well, but that can probably be handled in a follow-up though.)

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

Issue #9581, opened 2018-03-18
PDF Taking a Long Time To Render


Jon-Murray:
[architectural_-_annotation_scaling_and_multileaders.SECTIONS AND DETAILSs2 (2).png.pdf](https://github.com/mozilla/pdf.js/files/1822749/architectural_-_annotation_scaling_and_multileaders.SECTIONS.AND.DETAILSs2.2.png.pdf)
![architectural_-_annotation_scaling_and_multileaders sections and detailss2](https://user-images.githubusercontent.com/2168795/37567041-ff19c8f6-2ab8-11e8-8482-c35a7d9ed6cc.png)

Hi guys,

Apologies, as I hate these types of bug filings, but I have a relatively simple PDF which takes a long time to load on FF 59. This loads pretty much instantly in other PDF viewers. The PDF is just an embedded PNG inside a PDF document (though the attached document has a simple annotation added as well, this happens without it). It is causing me some grief as it takes ~40 seconds to load in Firefox/embedded PDFJS (both the most up-to-date versions as of March 2018) but has no issues in other browsers. Both the PDF and the original PNG should be attached here.

Configuration:
- Web browser and its version: FF 59
- Operating system and its version: Windows, Ryzen 1700, 16GB RAM
- PDF.js version: (PDF.js: 2.0.288) and older
- Is a browser extension: Both native FF and using PDFJS in other browsers

Steps to reproduce the problem:
1.  Open the attached PDF

What is the expected behavior? (add screenshot) Renders ~5s

What went wrong? (add screenshot): Consistently renders ~40s

I really have no further information nor the skills to say why the resultant PDF is complex, sorry. 

Thanks!

Jon-Murray:
![architectural_-_annotation_scaling_and_multileaders sections and detailss2](https://user-images.githubusercontent.com/2168795/37567046-102fd392-2ab9-11e8-80bf-d5603d480dce.png)

timvandermeij:
Evince takes about six seconds to render this PDF file. I think the problem is that the embedded image is 8000 x 5333 pixels, which is really large. Vector images would be much faster than large embedded PNG images. To fix this, we most likely need canvas tiling.

Jon-Murray:
[...]

PR #11963, opened 2020-06-02
Avoid calling Math.pow if possible.

tamuratak:
Related to #9581.

When loading a high-resolution image, I have found that the most time-consuming part is converting sRGB to RGB. Especially calling `Math.pow` is expensive. The following profile of Chrome is the one when loading the PDF file of #9581.

スクリーンショット 2020-06-03 8 10 58

スクリーンショット 2020-06-03 8 11 41

In this PR, we avoid calling `Math.pow` if the value of `color` is close enough to `1`. The threshold is determined to assure that the result is larger than `254.5`.

```
> ((1 + 0.055) * 0.99554525**(1/2.4) - 0.055)*255
 254.50000003134699
```

The benchmark result for the PDF file of #9581 is significant.

```
$  node stats/statcmp.js stats/results/baseline.json stats/results/current.json
browser | stat         | Count | Baseline(ms) | Current(ms) |    +/- |     %  | Result(P<.05)
------- | ------------ | ----- | ------------ | ----------- | ------ | ------ | -------------
chrome  | Overall      |     3 |        26821 |       11648 | -15172 | -56.57 |        faster
chrome  | Page Request |     3 |            7 |           7 |      0 |   0.00 |              
chrome  | Rendering    |     3 |        26814 |       11641 | -15173 | -56.59 |        faster
firefox | Overall      |     3 |        14270 |        9383 |  -4887 | -34.25 |        faster
firefox | Page Request |     3 |            6 |           5 |     -1 | -16.67 |              
firefox | Rendering    |     3 |        14264 |        9378 |  -4886 | -34.26 |        faster
```

One of other possible solutions is to avoid converting sRGB to RGB if not necessary. I think it is not necessary when we just display images without transparency operations.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -1057,6 +1057,15 @@ const CalRGBCS = (function CalRGBCSClosure() {
     if (color <= 0.0031308) {
       return adjustToRange(0, 1, 12.92 * color);
     }
+    // Optimization:
+    // If color is close enough to 1, skip calling the following transform
+    // since calling Math.pow is expensive. If color is larger than
+    // the threshold, the final result is larger than 254.5 since
+    // ((1 + 0.055) * 0.99554525 ** (1 / 2.4) - 0.055) * 255 ===
+    // 254.50000003134699
+    if (color >= 0.99554525) {
+      return 1;
+    }
     return adjustToRange(0, 1, (1 + 0.055) * color ** (1 / 2.4) - 0.055);
   }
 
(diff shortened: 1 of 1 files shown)


src/core/colorspace.js
```diff
@@ -1057,6 +1057,9 @@ const CalRGBCS = (function CalRGBCSClosure() {
     if (color <= 0.0031308) {
       return adjustToRange(0, 1, 12.92 * color);
     }
+    if (color >= 0.99554525) {
```
Snuffleupagus, line 1060: Please add a comment above this line, explaining exactly how this value was determined!
As-is, this line looks quite wrong until you've actually read https://github.com/mozilla/pdf.js/pull/11963#issue-426882562

Snuffleupagus (commented):
While https://github.com/mozilla/pdf.js/pull/11963#issue-426882562 contains *a lot* of nice information/context, please make sure that that information is also present in the actual *commit message* as well.

That way the information is generally available on e.g. the Git command line too (which is often useful when working with the code locally or when bisecting something). Also, explicitly mentioning that this applies to the CalRGB colorspace in the title/commit message probably doesn't hurt either :-)

tamuratak:
Thank you for your responses. I have edited as requested.

timvandermeij (approved):

timvandermeij:
Nice work; thank you for figuring this out!

Issue #11931, opened 2020-05-25
PDF rendered strangely


photopea:
This PDF (the JPG image inside it) is rendered strangely. 

The three channels inside are RGB, while your parser thinks they are Y Cb Cr, and performs _convertYccToRgb()

Actually, the PDF in pdf.js looks the same as in Adobe Reader, but differently than in Chrome PDF viewer. But if you "cut out" a raw JPG file from the PDF, image viewers render it as the Chrome PDF viewer.

[head.pdf](https://github.com/mozilla/pdf.js/files/4679027/head.pdf)
[head.jpg](https://user-images.githubusercontent.com/15955558/82846232-1929ea00-9ee8-11ea-8f25-53a72b14160c.jpg)

havocbcn:
In #11967 it's the explanation and the solution, I'm waiting further instructions for merging

timvandermeij:
Fixed by the pull request above.

PR #11966, opened 2020-06-04
Do not transform jpeg RGB components

havocbcn:
Hello,

Fix issue #11931

The image is encoded using RGB, not YCbCr

here : https://stackoverflow.com/questions/50798014/determining-color-space-for-jpeg/50861048

_On ISO/IEC 10918-6:2013 (E), section 6.1:_  (http://www.itu.int/rec/T-REC-T.872-201206-I/en)

_Images encoded with three components are assumed to be RGB data encoded as YCbCr unless the image contains an APP14 marker segment as specified in 6.5.3, in which case the colour encoding is considered either RGB or YCbCr according to the application data of the APP14 marker segment_ 

The head.jpg doesn't have a app14, so image seems to be using YCbCr not RGB

But, the second best answer points to look up the implementation of a Java jpeg library, seems that if components are called R, G and B, the space color is RGB, and this is the case, the components are called in this way.

This pull request maintain the index in the components struct to allow _isColorConversionNeeded to do not convert if the components are called R, G and B.

The Pdf is now showed correctly


havocbcn:
Sorry, trying to squash

havocbcn:
Sorry, messed up, another PR in #11967

Issue #11961, opened 2020-06-02
Collect telemetry on all unsupported features


emalysz:
In #11815, we collected telemetry on what error triggered the fallback bar. The first error message is what is shown in the fallback bar. We need expand this and collect telemetry on all error messages to give us a better understanding of the frequency of each type of error. 

Example with more than one error: 
https://www.irs.gov/pub/irs-pdf/f1040.pdf

Snuffleupagus:
One possible way of doing this could be to decouple the showing of the fallback bar from the reporting of telemetry data, to not complicate e.g. the `PDFViewerApplication.fallback` method unnecessarily; perhaps something along these lines:

 - Remove the `PdfJsTelemetry.onFallback(featureId);` line from `ChromeActions.fallback`; see https://searchfox.org/mozilla-central/rev/559b25eb41c1cbffcb90a34e008b8288312fcd25/browser/extensions/pdfjs/content/PdfStreamConverter.jsm#499
 - Add a new case, e.g. "fallback" (or perhaps something better named) to the `switch`-block in `ChromeActions.reportTelemetry` (perhaps similar to the "documentStats" case); see https://searchfox.org/mozilla-central/rev/559b25eb41c1cbffcb90a34e008b8288312fcd25/browser/extensions/pdfjs/content/PdfStreamConverter.jsm#424-481
 - Always report telemetry in `PDFViewerApplication.fallback`, by adding something like
    ```javascript
   this.externalServices.reportTelemetry({
      type: "fallback",
      featureId,
   });
   ```
   immediately after line https://github.com/mozilla/pdf.js/blob/8fc1126b5a2b6d2fa0cfa98a061882ef21c1abc9/web/app.js#L865-L869

emalysz:
Upstream bug to add telemetry in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1643402

PR #11969, opened 2020-06-04
For #11961: collect telemetry on all unique unsupported features that…

emalysz:
… trigger fallback error.

This expands upon the telemetry we are collecting that shows the fallback error.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -864,11 +864,24 @@ const PDFViewerApplication = {
       .catch(downloadByUrl); // Error occurred, try downloading with the URL.
   },
 
+  _recordFallbackErrorTelemetry(featureId) {
+    if (typeof PDFJSDev === "undefined" || PDFJSDev.test("MOZCENTRAL")) {
+      this.externalServices.reportTelemetry({
+        type: "unsupportedFeature",
+        featureId,
+      });
+    }
+  },
+
   fallback(featureId) {
     if (
       typeof PDFJSDev === "undefined" ||
       PDFJSDev.test("MOZCENTRAL || GENERIC")
     ) {
+      if (featureId) {
+        this._recordFallbackErrorTelemetry(featureId);
+      }
+
       // For PDFs that contain script and form errors, we should only trigger
       // the fallback once the user has interacted with the page.
       if (this._delayedFallbackFeatureIds.length >= 1 && this._hasInteracted) {
@@ -1247,6 +1260,7 @@ const PDFViewerApplication = {
         }
         console.warn("Warning: JavaScript is not supported");
         this._delayedFallbackFeatureIds.push(UNSUPPORTED_FEATURES.javaScript);
+        this._recordFallbackErrorTelemetry(UNSUPPORTED_FEATURES.javaScript);
         return true;
       });
 
@@ -1329,6 +1343,7 @@ const PDFViewerApplication = {
     if (info.IsAcroFormPresent) {
       console.warn("Warning: AcroForm/XFA is not supported");
       this._delayedFallbackFeatureIds.push(UNSUPPORTED_FEATURES.forms);
+      this._recordFallbackErrorTelemetry(UNSUPPORTED_FEATURES.forms);
     }
 
     if (


web/app.js
```diff
@@ -867,6 +868,15 @@ const PDFViewerApplication = {
       typeof PDFJSDev === "undefined" ||
       PDFJSDev.test("MOZCENTRAL || GENERIC")
     ) {
+      if (featureId && !this._unsupportedFeatureErrors.includes(featureId)) {
```
emalysz, line 871: null check for `featureId` because of the changes made in https://github.com/mozilla/pdf.js/pull/11953. That would trigger fallback() with a null featureId

timvandermeij:
I think I'm missing something here, but we already have the `PDF_VIEWER_FALLBACK_REASON` telemetry probe at https://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2020-05-30&include_spill=0&keys=__none__!__none__!__none__&max_channel_version=nightly%252F78&measure=PDF_VIEWER_FALLBACK_REASON&min_channel_version=null&processType=*&product=Firefox&sanitize=1&sort_by_value=0&sort_keys=submissions&start_date=2020-05-20&table=0&trim=1&use_submission_date=0, so we should already collect the unique unsupported features?

emalysz:
Hey @timvandermeij, sorry if I wasn't clear about this. The `PDF_VIEWER_FALLBACK_REASON` tells us the reason the fallback bar is shown. However, forms can have multiple errors (https://www.irs.gov/pub/irs-pdf/f1040.pdf will have form, javascript, and font errors). Since only the first error is shown, we wanted to add telemetry to see the total number of errors that trigger the fallback (meaning most are not reflected in the fallback bar). This probe would allow us to record all three of those above errors, not just form.

web/app.js
```diff
@@ -1235,6 +1244,11 @@ const PDFViewerApplication = {
           return false;
         }
         console.warn("Warning: JavaScript is not supported");
+        this.externalServices.reportTelemetry({
```
timvandermeij, line 1247: Is there a reason why this cannot be done (unconditionally) in the `fallback` method? There are more kinds of unsupported features, so why are only these special-cased here?

timvandermeij:
> Since only the first error is shown, we wanted to add telemetry to see the total number of errors that trigger the fallback (meaning most are not reflected in the fallback bar).

Ah, that was the missing piece for me. Thank you for clarifying that! I was always under the impression that we reported all errors instead of only the first one.

web/app.js
emalysz: Yeah I originally thought the same thing, but was hesitant given my work in #11953 (which I think will land before this patch). There, we would trigger the fallback for form and script errors only when the user clicks, and `featureId` would be null. 

So once that lands, I could implement a loop to report the errors within `this._delayedFallbackFeatureIds` in the `fallback` method, or the errors could be reported in these two places when we trigger a console.warn. Which would you prefer?

timvandermeij: Good point. I'm thinking it's easiest if we can consolidate the telemetry reporting in one place, but we can see how it looks once the other PR is merged.

web/app.js
Snuffleupagus: Unless I'm mistaken, wouldn't it be a lot simpler/safer to do this kind of validation/tracking of seen `featureId`s in the new case that you're adding to [`ChromeActions.reportTelemetry`](https://searchfox.org/mozilla-central/rev/598e50d2c3cd81cd616654f16af811adceb08f9f/browser/extensions/pdfjs/content/PdfStreamConverter.jsm#424) and simply report unconditionally here instead (note how e.g. `documentStats` is implemented)?

Since that's the central spot where all telemetry results pass through, you'd be thus be guaranteed to only report a specific `featureId` value *once*.
By tracking seen `featureId`s in manually in `web/app.js`, as is done in this patch, it seems much easier for any future changes to accidentally e.g. double-report telemetry data if one forgets to add a `this._unsupportedFeatureErrors.push(featureId);` line.

web/app.js
```diff
@@ -869,6 +869,20 @@ const PDFViewerApplication = {
...
+          type: "unsupportedFeature",
+          featureId,
+        });
+      } else {
+        for (const unsupportedFeature of this._delayedFallbackFeatureIds) {
+          this.externalServices.reportTelemetry({
+            type: "unsupportedFeature",
+            featureId: unsupportedFeature,
+          });
```
timvandermeij, line 882: Travis CI complains about extra whitespace at the end of the line.

timvandermeij:
Looks good to me. Can we already land this or should we wait for the upstream change to be done first?

web/app.js
emalysz: Thanks for pointing that out @Snuffleupagus. I opted to create a helper method to avoid `reportTelemetry` code duplication. 

Another option I considered was calling fallback here but passing in an optional bool that would tell us to immediately exit the function following the telemetry collection.

Which solution do you each prefer? @timvandermeij

Snuffleupagus: > Another option I considered was calling fallback here but passing in an optional bool that would tell us to immediately exit the function following the telemetry collection.

That honestly sounds worse than the current solution, as far as I'm concerned :-)

(There may be potential for some additional clean-up, since I do think that we're now tracking a bit more state than is probably desirable, but that's perhaps better for a follow-up patch.)

web/app.js
```diff
@@ -864,11 +864,27 @@ const PDFViewerApplication = {
       .catch(downloadByUrl); // Error occurred, try downloading with the URL.
   },
 
+  _recordFallbackErrorTelemetry(featureId) {
+    if (
+      typeof PDFJSDev === "undefined" ||
+      PDFJSDev.test("MOZCENTRAL || GENERIC")
```
timvandermeij, line 870: Why `GENERIC` here as well? As far as I can see `reportTelemetry` is only implemented and useful in `MOZCENTRAL` builds.

timvandermeij (approved):

timvandermeij:
Thank you!

PR #11972, opened 2020-06-05
Change the `loadedChunks` property, on `ChunkedStream` instances, from an Array to a Set

Snuffleupagus:
In the old code the use of an Array meant that we had to *manually* track the `numChunksLoaded` property, given that simply using the Array `length` wouldn't have worked since there's no guarantee that the data is loaded in order when e.g. range requests are in use.

Tracking closely related state *separately* in this manner never seem like a good idea, and we can now instead utilize a Set to avoid that.

### DIFF
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -29,8 +29,7 @@ class ChunkedStream {
     this.pos = 0;
     this.end = length;
     this.chunkSize = chunkSize;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Less is more when it comes to tracking related state. Thanks!

PR #11973, opened 2020-06-05
Added missing `page_canvas` and removed `web_fonts_disabled`.

AlShaikh:
This removes the console warning `#page_canvas is undefined.`.

### DIFF
--- a/l10n/ar/viewer.properties
+++ b/l10n/ar/viewer.properties
@@ -148,6 +148,8 @@ thumbs_label=مُصغّرات
 findbar.title=ابحث في المستند
 findbar_label=ابحث
 
+# LOCALIZATION NOTE (page_canvas): "{{page}}" will be replaced by the page number.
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Please see https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md:
> Note that the translations for PDF.js in the l10n folder are imported from the Nightly channel of Mozilla Firefox, such that we don't have to maintain them ourselves. This means that we will not accept pull requests that add new languages and/or modify existing translations, unless the corresponding changes have been made in Mozilla Firefox first.

timvandermeij:
Closing since we cannot merge this as the changes will be overwritten on the next sync. Instead, please send this change to one of the Arabic translation maintainers; you can find their contact information here: https://wiki.mozilla.org/L10n:Teams:ar. Thank you.

PR #11974, opened 2020-06-07
A couple of small image caching/sending improvements

Snuffleupagus:
- Remove an unnecessary `RefSetCache.prototype.has()` call from `GlobalImageCache.getData`

   We can simply attempt to get the data *directly*, and instead check the result, rather than first checking if it exists.

 - Extract the actual sending of image data from the `PartialEvaluator.buildPaintImageXObject` method

   After PRs #10727 and #11912, the code responsible for sending the decoded image data to the main-thread has now become a fair bit more involved the previously.
   To reduce the amount of duplication here, the actual code responsible for sending the data is thus extracted into a new helper method instead.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -439,6 +439,30 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
       });
     },
 
+    _sendImgData(objId, imgData, cacheGlobally = false) {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

/botio makeref (in the hope to remove some failures)

PR #11977, opened 2020-06-07
Convert the `RefSet` primitive to a proper class and use a `Set` internally

timvandermeij:
The `RefSet` primitive predates ES6, so that most likely explains why an object is used internally to track the entries. However, nowadays we can use built-in JavaScript sets for this purpose. Built-in types are often more efficient/optimized and using it makes the code a bit more clear since we don't have to assign `true` to keys anymore just to indicate their presence.

### DIFF
--- a/src/core/primitives.js
+++ b/src/core/primitives.js
@@ -221,28 +221,23 @@ var Ref = (function RefClosure() {
 
 // The reference is identified by number and generation.
 // This structure stores only one instance of the reference.
-var RefSet = (function RefSetClosure() {
-  // eslint-disable-next-line no-shadow
-  function RefSet() {
-    this.dict = Object.create(null);
+class RefSet {
+  constructor() {
+    this._set = new Set();
   }
 
-  RefSet.prototype = {
-    has: function RefSet_has(ref) {
-      return ref.toString() in this.dict;
-    },
+  has(ref) {
+    return this._set.has(ref.toString());
+  }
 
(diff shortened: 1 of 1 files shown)


Snuffleupagus (changes requested):
Really sorry, but I'm not at all keen on these changes!

The `RefSet` abstraction is, as its name suggests, written specifically to operate only on `Ref` objects and I worry that simply using Sets directly in this manner makes the code more difficult to read/modify and that bugs are thus much easier to introduce.
Essentially, in the worker, it's generally preferable to avoid depending directly on the stringified version of a `Ref` object if possible.

Basically, unless there's test-cases where benchmarking shows *clear* improvements, then I'd ask that we please don't make these changes.

timvandermeij:
I understand your concern here. I have updated the patch to keep `RefSet`, but instead refactor it to use a set internally instead of an object. Would that be better?

Snuffleupagus:
> I have updated the patch to keep `RefSet`, but instead refactor it to use a set internally instead of an object. Would that be better?

Yes, that seems completely fine by me; thank you for doing this!

(Also, possibly this could make `RefSet.has()` more efficient since I recall some older issues where avoiding `in` usage helped performance.)

---

Just for good measure, let's run all tests :-)

/botio test

Snuffleupagus:
An idea could be to try and do something similar, but obviously with a `Map` instead, for the `RefSetCache`; I'd happily review such a patch :-)

PR #11992, opened 2020-06-11
Prevent the (old) preprocessor from appending trailing whitespace when removing closing HTML comments

Snuffleupagus:
This can currently be seen in the *built* `web/viewer.html` file, at the line containing "  <script src="viewer.js"></script>  ".


timvandermeij (approved):

timvandermeij:
Looks good!

PR #11993, opened 2020-06-11
Change the `dependencies` property, on `OperatorList` instances, from an Object to a Set

Snuffleupagus:
Since this is completely internal functionality, and furthermore limited to the worker-thread, this change should thus not have any observable effect for e.g. an API-user.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #11997, opened 2020-06-12
Add basic support for the nullish coalescing operator `??`

Snuffleupagus:
For now we need to use a Babel-plugin, since Webpack 4.x doesn't seem to support it yet. (Most likely we'll have to update to Webpack 5, once that becomes available, in order for this to be directly supported. This is thus also blocked on removing the `webpack-stream` package.)

While the `??` operator will thus always be transpiled by Babel, even in modern builds, simply supporting it for development purposes seems like a step in the right direction.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -215,6 +215,12 @@ function createWebpackConfig(defines, output) {
           options: {
             presets: skipBabel ? undefined : ["@babel/preset-env"],
             plugins: [
+              [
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks like a good step forward. Thanks!

PR #12000, opened 2020-06-13
Ensure that `PDFImage.buildImage` won't accidentally swallow errors, e.g. from ColorSpace parsing (issue 6707, PR 11601 follow-up)

Snuffleupagus:
Because of a really stupid `Promise`-related mistake on my part, when re-factoring `PDFImage.buildImage` during the `NativeImageDecoder` removal, we're no longer re-throwing errors occuring during image parsing/decoding as intended.
The result is that some (fairly) corrupt documents will never finish loading, and unfortunately there were apparently no sufficiently corrupt images in the test-suite to catch this.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -579,7 +579,7 @@ var PartialEvaluator = (function PartialEvaluatorClosure() {
           return this._sendImgData(objId, imgData, cacheGlobally);
         })
         .catch(reason => {
-          warn("Unable to decode image: " + reason);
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
This is a regression that I stumbled upon while working on https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:cache-ColorSpace-locally :-)

timvandermeij (approved):

timvandermeij:
Thank you for finding and fixing this, and good that the test coverage is extended to avoid this in the future!

/botio makeref

PR #12002, opened 2020-06-15
Small improvements in `CFFCompiler.encodeNumber` and `CFFCompiler.encodeFloat`

Snuffleupagus:
- Simplify the "is integer" checks in `CFFCompiler.encodeNumber`

   The `isNaN` check is obviously redundant, since `NaN` is the only value that isn't equal to itself; see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN#Examples

   The `parseFloat`/`parseInt` comparison would make sense if the `value` ever contains a String, which however is never actually the case. Besides looking through the code, I've also run the entire test-suite locally with `assert(typeof value === "number", "encodeNumber");` added at the top of the method and there were no failures.

   Hence we can simplify the "is integer" check a bit in the `CFFCompiler.encodeNumber` method.

 - Lazily initialize, and cache, the regular expression used in `CFFCompiler.encodeFloat`

   There's no particular reason for re-creating the regular expression over and over for every `encodeFloat` invocation, as far as I can tell.

### DIFF
--- a/src/core/cff_parser.js
+++ b/src/core/cff_parser.js
@@ -1366,6 +1366,8 @@ var CFFOffsetTracker = (function CFFOffsetTrackerClosure() {
 
 // Takes a CFF and converts it to the binary representation.
 var CFFCompiler = (function CFFCompilerClosure() {
+  let EncodeFloatRegExp = null; // Lazily initialized by `encodeFloat`.
+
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good find!

PR #12004, opened 2020-06-16
Convert the code in `src/core/worker.js` to use ES6 classes

Snuffleupagus:
This removes one instance of `// eslint-disable-next-line no-shadow`, which our old pseudo-classes necessitated.

*Please note:* I'm purposely not doing any `var` to `let`/`const` conversion here, since it's generally better to (if possible) do that automatically on e.g. a directory basis instead.

*Smaller diff with https://github.com/mozilla/pdf.js/pull/12004/files?w=1*


timvandermeij (approved):

timvandermeij:
Thank you!

PR #12012, opened 2020-06-18
Improve (local) caching of parsed `ColorSpace`s (PR 12001 follow-up)

Snuffleupagus:
This patch contains the following *notable* improvements:
 - Changes the `ColorSpace.parse` call-sites to, where possible, pass in a reference rather than actual ColorSpace data (necessary for the next point).
 - Adds (local) caching of `ColorSpace`s by `Ref`, when applicable, in addition the caching by name. This (generally) improves `ColorSpace` caching for e.g. the SMask code-paths.
 - Extends the (local) `ColorSpace` caching to also apply when handling Images and Patterns, thus further reducing unneeded re-parsing.
 - Adds a new `ColorSpace.parseAsync` method, almost identical to the existing `ColorSpace.parse` one, but returning a Promise instead (this simplifies some code in the `PartialEvaluator`).

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -259,8 +259,8 @@ class ColorSpace {
     return shadow(this, "usesZeroToOneRange", true);
   }
 
-  static parse(cs, xref, res, pdfFunctionFactory) {
-    const IR = this.parseToIR(cs, xref, res, pdfFunctionFactory);
+  static parse({ cs, xref, resources = null, pdfFunctionFactory }) {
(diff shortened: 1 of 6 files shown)


timvandermeij (approved):

timvandermeij:

/botio makeref

Snuffleupagus:
> Thank you for working on this!

As always, thanks for landing the PR!

(I'm currently looking into caching of parsed Functions, which may further improve some documents.)

PR #12017, opened 2020-06-21
Convert the `PDFPageProxy.intentStates` property from an `Object` to a `Map`

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Looks better to me too. Thanks!

PR #12021, opened 2020-06-24
Move the `fetchBuiltInCMap` method to the `PartialEvaluator.prototype`

Snuffleupagus:
Defining this *inline* in the "constructor" looks slightly weird (I really don't know why I wrote it like that originally), and it can simply be changed to a regular method instead.


Snuffleupagus:
The Firefox test-failures, on the Linux bot, look identical to those in the latest test-runs in PR #12012, hence those should be caused by a Firefox update rather than this patch.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #8043, opened 2017-02-07
Allow automatic print rotation via the enablePrintAutoRotate preference

Rob--W:
Tests:

1. `gulp server`
2. Download [test1.pdf](https://github.com/mozilla/pdf.js/files/759312/test1.pdf) and [test2.pdf](https://github.com/mozilla/pdf.js/files/759311/test2.pdf) (courtesy of #6103) and save them to the pdf.js repo.
3. Open the viewer, run `PDFViewerApplication.preferences.set('enablePrintAutoRotate', true);` and refresh.
4. Press Ctrl + P within the page (or click on the print button).
5. Verify that the pages have the same orientation.
6. Repeat step 3 - 5 for test1.pdf
7. Repeat step 3 with `false` instead of `true`, then perform step 4 and confirm that the page rotation did not change.

Fixes #6696, supersedes #6103

Issue #2851, opened 2013-03-01
printing not correct for landscape


ce03:
This pdf file is landscape, and is displayed as such in the browser.  When I print it, it appears as half of a portrait page. If I use Print Preview I can change to landscape but I don't want to have to remember to do this.

Sample file - http://services.land.vic.goc.au/pdfjs/map1.pdf
Screen grab - http://services.land.vic.goc.au/pdfjs/map1.jpg

silentsakky:
Does anyone have any fix for this issue?

ousia:
_Acrobat_ has an orientation parameter in the printing  dialog that can be set to auto, such as in:

![not this](https://helpx.adobe.com/acrobat/kb/print-mixed-page-sizes-acrobat/_jcr_content/main-pars/procedure_0/proc_par/step_2/step_par/img_0.img.png/mixed_win_step3.png)

Image linked from _Adobe_ and it has another property selected.

Maybe it would be helpful that PDF.js implements page auto-orientation as default.

jscher2000:
Could something like this help in adapting to the unknown orientation the browser will use:

````
@media print and (orientation: portrait) {
  /* Rotate page if it is landscape */
}
@media print and (orientation: landscape) {
  /* Rotate page if it is portrait */
}
````

stbear:
I've changed the following lines of code:
`return pagesOverview.map(function (size) {
        if (isFirstPagePortrait === (0, _ui_utils.isPortraitOrientation)(size)) {
          return size;
        }`
to
`return pagesOverview.map(function (size, index) {
        if ((index == 0 && isFirstPagePortrait || index > 0 && (0, _ui_utils.isPortraitOrientation)(size))) {
          return size;
        }`
and it works fine for me. Didn't test it much, since our app with embedded PDF.js will be deployed to end users in fall.

timvandermeij:
Closing since this is fixed by the PR above.

PR #12023, opened 2020-06-24
Fix auto-rotate for printing landscape PDFs where the first page is also landscape

### LINKED ISSUE #9297: Auto-rotate for printing does not work
Attach (recommended) or Link to PDF file here: http://www.co.dutchess.ny.us/CountyGov/Departments/StopDWI/DontBlowIt.pdf

Configuration:
- Web browser and its version: Firefox 57.0.2 (64 bits)
- Operating system and its version: Arch Linux
- PDF.js version: 1.9.583
- Is a browser extension: no (embedded in Firefox)

1. Enable `pdfjs.enablePrintAutoRotate` in `about:config`.
2. Open the file.
3. Open the print preview.
4. Notice that the landscape pages are not rotated.

They should be rotated to fill the entire landscape page.

They are sized onto a portrait page.

-

escapewindow:
The current behavior for `getPagesOverview` assumes we want to only
auto-rotate if:

- `enablePrintAutoRotate` is `true`
- `isFirstPagePortrait !== isPortraitOrientation(size)`

This second check is what is breaking #9297. The two PDFs linked have a
landscape orientation first page, as well as subsequent pages. Since
`false === false`, we print portrait.

Let's drop the comparison with `isFirstPagePortrait`, and print
landscape if `!isPortraitOrientation(size)`.

Fixes #9297.

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -167,7 +167,7 @@
     },
     "enablePrintAutoRotate": {
       "title": "Automatically rotate printed pages",
-      "description": "When enabled, pages whose orientation differ from the first page are rotated when printed.",
+      "description": "When enabled, landscape pages are rotated when printed.",
       "type": "boolean",
       "default": false
     },
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -67,8 +67,7 @@ const DEFAULT_CACHE_SIZE = 10;
  * @property {boolean} [renderInteractiveForms] - Enables rendering of
  *   interactive form elements. The default is `false`.
  * @property {boolean} [enablePrintAutoRotate] - Enables automatic rotation of
- *   pages whose orientation differ from the first page upon printing. The
- *   default is `false`.
+ *   landscape pages upon printing. The default is `false`.
  * @property {string} renderer - 'canvas' or 'svg'. The default is 'canvas'.
  * @property {boolean} [enableWebGL] - Enables WebGL accelerated rendering for
  *   some operations. The default value is `false`.
@@ -1203,9 +1202,8 @@ class BaseViewer {
     if (!this.enablePrintAutoRotate) {
       return pagesOverview;
     }
-    const isFirstPagePortrait = isPortraitOrientation(pagesOverview[0]);
     return pagesOverview.map(function (size) {
-      if (isFirstPagePortrait === isPortraitOrientation(size)) {
+      if (isPortraitOrientation(size)) {
         return size;
       }
       return {


escapewindow:
Per https://github.com/mozilla/pdf.js/pull/8043#issuecomment-349817427 , this seems to be a fix for #2851 as well. I don't know why we only rotate pages if the first page is portrait, introduced in #8043 ...

escapewindow:
@timvandermeij do you have any guidance here?

I'm guessing that either we want this fix, or we want to keep the old behavior, and fix landscape printing some other way. If the latter is true, I'm curious if we have any preferred workflows (another pref? change `enablePrintAutoRotate` to an enum?)

Even-x (approved):

Yay295:
Looking at #8043 I feel like there could be some more changes made here.

1. `isPortraitOrientation()` was previously only called in two places. Now that you've removed one of them, it could be inlined.
2. The description for `enablePrintAutoRotate` in `extensions/chromium/preferences_schema.json` and `web/pdf_viewer.js` should be updated since it's not "automatic rotation of pages **whose orientation differ from the first page**" anymore.

I also think `enablePrintAutoRotate` should be true by default, but that's a different discussion.

escapewindow:
> The current behavior for `getPagesOverview` assumes we want to only
> auto-rotate if:
> 
>     * `enablePrintAutoRotate` is `true`
> 
>     * `isFirstPagePortrait !== isPortraitOrientation(size)`
> 
> This second check is what is breaking #9297. The two PDFs linked have a
> landscape orientation first page, as well as subsequent pages. Since
> `false === false`, we print portrait.
> 
> Let's drop the comparison with `isFirstPagePortrait`, and print
> landscape if `!isPortraitOrientation(size)`.

Ideally, we'd replace the second check with a comparison against the printer page orientation. We may be able to get that information via print preview support.

escapewindow:
> Looking at #8043 I feel like there could be some more changes made here.
> 
>     1. `isPortraitOrientation()` was previously only called in two places. Now that you've removed one of them, it could be inlined.

Hm, is that true?

```
(3.8.0:3.7.4:2.7.17) akimoz: ~/src/pdf/pdf.js (issue9297) [09:38:58]
10234$ rg -l isPortraitOrientation
web/pdf_document_properties.js
web/base_viewer.js
web/ui_utils.js
test/unit/ui_utils_spec.js
```

>     2. The description for `enablePrintAutoRotate` in `extensions/chromium/preferences_schema.json` and `web/pdf_viewer.js` should be updated since it's not "automatic rotation of pages **whose orientation differ from the first page**" anymore.

Makes sense. Done.

> I also think `enablePrintAutoRotate` should be true by default, but that's a different discussion.

I'm leaning that way, but I'm not sure if there are reasons for the current behavior that I'm not aware of.

Yay295:
Oh, you're right about `isPortraitOrientation()`. I was only looking at the changes in #8043 and didn't check that it had been used elsewhere since then.

test/pdfs/.gitignore
```diff
@@ -168,6 +168,7 @@
 !issue6612.pdf
 !alphatrans.pdf
 !pattern_text_embedded_font.pdf
+!issue9297.pdf
```
timvandermeij, line 171: The test changes in this PR can be removed because we don't have support for testing printing, so this test does not actually test printing at all now.

escapewindow: Removed and rebased.

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

escapewindow:
Thank you for the review!

ilium007:
How can I patch my Firefox installs on OSX?

timvandermeij:
This will be merged soon into Firefox Nightly and after that follow the normal Firefox release schedule.

PR #12024, opened 2020-06-25
Update Needle to 2.5.0 or greater.

DylanLacey:
Versions of `needle` prior to `2.5.0` cannot cope with redirects ([as documented](https://github.com/tomas/needle/issues/312)).

This prevents prebuilt `canvas` binaries from being downloaded on MacOS,
requiring the global install of its dependencies.

Updating `needle` restores it to functionality, addressing this.  This change avoids
the need to add `request` to `package.json` or bundle pre-built binaries,  it also obsoletes https://github.com/mozilla/pdf.js/pull/12018.

(NB: needle was already in use for this package.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8919,9 +8919,9 @@
       "dev": true
     },
     "needle": {
-      "version": "2.4.0",
-      "resolved": "https://registry.npmjs.org/needle/-/needle-2.4.0.tgz",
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Let's do this. Thank you!

PR #12028, opened 2020-06-26
Attempt to detect inline images which contain "EI" sequence in the actual image data (issue 11124)

### LINKED ISSUE #11124: Reading inline image data is incorrectly implemented [BUG]

[bi.pdf](https://github.com/mozilla/pdf.js/files/3584861/bi.pdf)

Along these lines in https://github.com/mozilla/pdf.js/blob/master/src/core/parser.js:
``` node js
  makeInlineImage(cipherTransform) {
...
    this.buf2 = Cmd.get('EI'); 
```

I suppose it simply searches for "EI" sequence separated by whitespaces. But that is not enough. It is needed to calculate size of decoded image data from parameters like width, height, bit per component and color space. Then decode data and see for EOD or look for exactly right number of decoded bytes and then stop parsing image data as not every filter has EOD.

This is content stream from attached pdf:
```
100 0 0 100 0 0 cm
BI /W 4 /H 4 /CS /RGB /BPC 8
ID
00000z0z00zzz00z0zzz0zzzEI aazazaazzzaazazzzazzz
EI
```

1. Open attached bi.pdf

Screenshot 2019-09-06 at 19 06 30

This is how it looks in any other pdf viewer. All pixels are present and interpolation is off.

Screenshot 2019-09-06 at 18 54 20

Seems it stoped decoding image in middle where is byte sequence "EI " but in this context meant as color of pixel instead of "End of Image" keyword.

Another thing is that it uses image interpolation even that it should be disabled by default in pdf.

To enable interpolation it has to be:
```
BI /W 4 /H 4 /CS /RGB /BPC 8 /I true
```

Then it would look like this in correct pdf viewer:
Screenshot 2019-09-06 at 19 11 47

Snuffleupagus:
This should reduce the possibility of accidentally truncating some inline images, while *not* causing the "EI" detection to become significantly slower.[1]
There's obviously a possibility that these added checks are not sufficient to catch *every* single case of "EI" sequences within the actual inline image data, but without specific test-cases I decided against over-engineering the solution here.

*Please note:* The interpolation issues are somewhat orthogonal to the main issue here, which is the truncated image, and it's already tracked elsewhere.

Fixes #11124

---
[1] I've looked at the issue a few times, and this is the first approach that I was able to come up with that didn't cause *unacceptable* performance regressions in e.g. issue #2618.

### DIFF
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -203,10 +203,11 @@ class Parser {
       I = 0x49,
       SPACE = 0x20,
       LF = 0xa,
-      CR = 0xd;
-    const n = 10,
+      CR = 0xd,
       NUL = 0x0;
-    const startPos = stream.pos;
+    const lexer = this.lexer,
+      startPos = stream.pos,
+      n = 10;
     let state = 0,
       ch,
       maybeEIPos;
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

/botio makeref

misos1:
Should not be the end of image data found by looking at /W, /H, /BPC and /F entries and at image data itself? For example DCT image data ends at EOI marker. For filters where size of data is not known from data itself it should be clear from width, height and bpc how many bytes are needed. And only after that search for EI and only use such "heuristic" as fallback in case pdf.js does not support given filter.

PR #12039, opened 2020-06-29
[api-minor] Use the `NodeCanvasFactory`/`NodeCMapReaderFactory` classes as defaults in Node.js environments (issue 11900)

### LINKED ISSUE #11900: pdf2png example cannot render Japanese font
Configuration:
N/A? Just using pdf.js (8b9492a5c41d0776488a27021981b60fbf23a3bf) by node

macOS 10.15.3

- PDF.js version:
8b9492a5c41d0776488a27021981b60fbf23a3bf

- Is a browser extension:
N/A

1. `curl -o issue.pdf https://www.soumu.go.jp/johotsusintokei/whitepaper/ja/h25/pdf/n4100000.pdf`
2. apply the following diff:

```diff
--- a/examples/node/pdf2png/pdf2png.js
+++ b/examples/node/pdf2png/pdf2png.js
@@ -51,7 +51,7 @@ NodeCanvasFactory.prototype = {
 var pdfjsLib = require("pdfjs-dist/es5/build/pdf.js");

 // Relative path of the PDF file.
-var pdfURL = "../../../web/compressed.tracemonkey-pldi-09.pdf";
+var pdfURL = "issue.pdf";

 // Read the PDF file into a typed array so PDF.js can load it.
 var rawData = new Uint8Array(fs.readFileSync(pdfURL));
```

3. run `node pdf2png.js`

image

image

This document is published by the Government of Japan: https://www.soumu.go.jp/johotsusintokei/whitepaper/ja/h25/pdf/n4100000.pdf

Snuffleupagus:
This moves, and slightly simplifies, code that's currently residing in the unit-test utils into the actual library, such that it's bundled with `GENERIC`-builds and used in e.g. the API-code.

As an added bonus, this also brings out-of-the-box support for CMaps in e.g. the Node.js examples.

Fixes #11900

*Edit:* Slightly smaller diff with https://github.com/mozilla/pdf.js/pull/12039/files?w=1

### DIFF
--- a/examples/node/pdf2png/pdf2png.js
+++ b/examples/node/pdf2png/pdf2png.js
@@ -50,14 +50,21 @@ NodeCanvasFactory.prototype = {
 
 var pdfjsLib = require("pdfjs-dist/es5/build/pdf.js");
 
-// Relative path of the PDF file.
-var pdfURL = "../../../web/compressed.tracemonkey-pldi-09.pdf";
+// Some PDFs need external cmaps.
+var CMAP_URL = "../../../node_modules/pdfjs-dist/cmaps/";
+var CMAP_PACKED = true;
 
-// Read the PDF file into a typed array so PDF.js can load it.
-var rawData = new Uint8Array(fs.readFileSync(pdfURL));
+// Loading file from file system into typed array.
+var pdfPath =
+  process.argv[2] || "../../../web/compressed.tracemonkey-pldi-09.pdf";
+var data = new Uint8Array(fs.readFileSync(pdfPath));
 
 // Load the PDF file.
-var loadingTask = pdfjsLib.getDocument(rawData);
+var loadingTask = pdfjsLib.getDocument({
+  data: data,
+  cMapUrl: CMAP_URL,
+  cMapPacked: CMAP_PACKED,
+});
 loadingTask.promise
   .then(function (pdfDocument) {
     console.log("# PDF document loaded.");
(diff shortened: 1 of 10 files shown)


src/display/api.js
```diff
@@ -863,9 +874,9 @@ class PDFDocumentProxy {
  *                    just before viewport transform.
  * @property {Object} [imageLayer] - An object that has beginLayout,
  *                    endLayout and appendImage functions.
- * @property {Object} [canvasFactory] - The factory that will be used
+ * @property {Object} [canvasFactory] - The factory instance that will be used
```
Snuffleupagus, line 877: I'm *really* not happy about the discrepancy between `CMapReaderFactory` (on `getDocument`) and `canvasFactory`, since in the first case the user needs to pass a class and in the second one a class *instance*.
I've tried to make this at least *slightly* clearer by updating the JSDoc comment here, but it still doesn't seem great. Unfortunately it might simply be (years) too late to try and "fix" this, and furthermore I'm also not entirely sure how to improve it (given how `canvasFactory` is being used). Oh, well...

examples/node/pdf2png/pdf2png.js
```diff
@@ -50,14 +50,20 @@ NodeCanvasFactory.prototype = {
```
Snuffleupagus, line 50: Maybe we should also export the various canvas/CMap-factories in the API as well, such that we could remove things like e.g. https://github.com/mozilla/pdf.js/blob/9993397e30bdd6c3e63800753da3e4c971fa2f35/examples/node/pdf2png/pdf2png.js#L20-L49
However, I'm not sure if that's (generally) desirable and it also felt somewhat orthogonal to the rest of the patch...

timvandermeij (approved):

timvandermeij:
Looks like a good step forwards to me. Thank you!

PR #12044, opened 2020-06-30
Adjust the heuristics used when dealing with rectangles, i.e. `re` operators, with zero width/height (issue 12010)

### LINKED ISSUE #12010: Corner rendered incorrectly in figure
Link to PDF file: http://mirrors.ctan.org/graphics/pgf/contrib/circuitikz/doc/circuitikzmanual.pdf (version 1.1.2)

Configuration:
- Web browser and its version: Firefox 77.0.1
- PDF.js version: 2.5.95 (included in Firefox)

Consider the bottom left corner in the figure at the bottom of page 148.

correct

What went wrong?
wrong

Snuffleupagus:
Fixes #12010


src/display/canvas.js
```diff
@@ -1185,20 +1185,20 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
             y = args[j++];
             var width = args[j++];
             var height = args[j++];
-            if (width === 0) {
+            if (width === 0 && ctx.lineWidth < this.getSinglePixelWidth()) {
```
timvandermeij, line 1188: Can't we do this `getSinglePixelWidth` function call only once? We're doing it four times in the worst case now, and the path construction code is usually quite hot.

timvandermeij (commented):
Looks good with the comment addressed and a rebase.

timvandermeij:
Note to self: makeref after this will also apply to the already merged PRs while the Linux bot was down.

src/display/canvas.js
Snuffleupagus: > We're doing it four times in the worst case now,

Well, that case *almost never* happens in practice since in most cases both the width *and* height is non-zero.
Hence I did it this way on purpose, since otherwise *every* single `re` operator would unnecessarily require a `this.getSinglePixelWidth()` call.

timvandermeij: That's seems fair; thanks!

timvandermeij (approved):

PR #12048, opened 2020-06-30
Checked checkboxes weren't printed correctly in acroforms

calixteman:
Is there a way to create a test to test a print issue ?

### DIFF
--- a/web/annotation_layer_builder.css
+++ b/web/annotation_layer_builder.css
@@ -106,8 +106,9 @@
 .annotationLayer .buttonWidgetAnnotation.checkBox input:checked:before,
 .annotationLayer .buttonWidgetAnnotation.checkBox input:checked:after {
   height: 80%;
-  left: 45%;
-  width: 1px;
+  left: 35%;
+  width: 0;
+  outline: 1px solid;
 }
 
 .annotationLayer .buttonWidgetAnnotation.checkBox input:checked:before {


Snuffleupagus:
Isn't it enough to simply test this "normally", and not worry about the printing part, using an `eq` test with the `"forms": true` parameter set (see the `test/test_manifest.json` file)?
Related, is this patch actually passing all existing tests locally (the `"forms": true` test-coverage is quite limited, but it does exists)?

See https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing

calixteman:
Yep tests are ok locally except that I get:
```
### Running browser tests
tmp/ exists -- unable to proceed with testing
```
The test would require at least a "user" interaction to click on a checkbox, is it possible ?

Snuffleupagus:
> Yep tests are ok locally except that I get:
> 
> ```
> ### Running browser tests
> tmp/ exists -- unable to proceed with testing
> ```

That does sound like none of the ref-tests actually ran, though!?
Did you follow the "Generating reference images"-section, from the link above, *before* running the tests?
(And if necessary, manually removing the `test/tmp/` directory between test-runs may be necessary.)

> The test would require at least a "user" interaction to click on a checkbox, is it possible?

No, unfortunately nothing like that is supported.

calixteman:
Yes, I made a `gulp makeref` before... but I've this error.
Anyway I removed `tmp` dir and then test ran fine.
The goal of this patch is to fix a printing issue after a user interaction... so I guess that currently there is no way to correctly test that.

timvandermeij:
I'm confused here because we explicitly don't support printing AcroForms yet. We don't even render the annotation layer onto the final canvas that is printed, so this must rely on browser printing then. Could you share a before/after patch screenshot for comparison?

timvandermeij:
We merged checkbox printing and I couldn't find anything wrong with the print back then. Do we still need to do anything here? If so, please indicate what this fixes with a screenshot.

PR #12049, opened 2020-07-01
Enable the `no-promise-executor-return` ESLint rule

Snuffleupagus:
Please see https://eslint.org/docs/rules/no-promise-executor-return for additional information.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #12056, opened 2020-07-04
Refactor/simplify the "delayedFallback" handling in the default viewer

Snuffleupagus:
There's a few things that could be improved in the current implementation, such as:
 - It's currently necessary to *both* manually track the `featureId`s which should trigger delayed fallback, as well as manually report telemetry in affected cases.
Obviously there's only two call-sites as of now (forms and javaScript), but it still feels somewhat error-prone especially if more cases were to be added in the future. To address this, this patch adds a new (private) method which abstracts away these details from the call-sites.

 - Generally, it also seems nice to reduce *and* simplify the amount of state we need to store on `PDFViewerApplication` in order to support the "delayedFallback" functionality.
Also, having to *manually* work with the "delayedFallback"-array in multiple places feels cumbersome and makes e.g. the `PDFViewerApplication.fallback` method less clear as to its behaviour.

 - Having code *outside* of `PDFViewerApplication`, i.e. in the event handlers, directly access properties which are marked as "private" via a leading underscore doesn't seem that great in general.
Furthermore, having the event handlers directly deal with that should be internal state also seem unfortunate. To address this, the patch will instead make use of a new `PDFViewerApplication.triggerDelayedFallback` callback.

 - There's at least one code-path in the viewer, see `PDFViewerApplication.error`, where `fallback` can be called without an argument.
It's currently possible (although maybe somewhat unlikely) that such a call *could* be overridden by the `featureId` of a pending "delayedFallback" call, thus not reporting the *correct* fallback reason.

 - The "delayedFallback"-state weren't being reset on document close (which shouldn't affect Firefox, but nonetheless it ought to be fixed).

*Smaller diff with https://github.com/mozilla/pdf.js/pull/12056/files?w=1*

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -189,8 +189,7 @@ const PDFViewerApplication = {
   externalServices: DefaultExternalServices,
   _boundEvents: {},
   contentDispositionFilename: null,
-  _hasInteracted: false,
-  _delayedFallbackFeatureIds: [],
+  triggerDelayedFallback: null,
 
   // Called once when the document is loaded.
   async initialize(appConfig) {
@@ -690,6 +689,7 @@ const PDFViewerApplication = {
     this.url = "";
     this.baseUrl = "";
     this.contentDispositionFilename = null;
+    this.triggerDelayedFallback = null;
 
     this.pdfSidebar.reset();
     this.pdfOutlineViewer.reset();
@@ -864,12 +864,29 @@ const PDFViewerApplication = {
       .catch(downloadByUrl); // Error occurred, try downloading with the URL.
   },
 
-  _recordFallbackErrorTelemetry(featureId) {
-    if (typeof PDFJSDev === "undefined" || PDFJSDev.test("MOZCENTRAL")) {
+  /**
+   * For PDF documents that contain e.g. forms and javaScript, we should only
+   * trigger the fallback bar once the user has interacted with the page.
+   * @private
+   */
+  _delayedFallback(featureId) {
+    if (
+      typeof PDFJSDev === "undefined" ||
+      PDFJSDev.test("MOZCENTRAL || GENERIC")
+    ) {
+      // Ensure that telemetry is always reported, since it's not guaranteed
+      // that the fallback bar will be shown (depends on user interaction).
       this.externalServices.reportTelemetry({
         type: "unsupportedFeature",
         featureId,
       });
+
+      if (!this.triggerDelayedFallback) {
+        this.triggerDelayedFallback = () => {
+          this.fallback(featureId);
+          this.triggerDelayedFallback = null;
+        };
+      }
     }
   },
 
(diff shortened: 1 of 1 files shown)


web/app.js
```diff
@@ -864,12 +864,29 @@ const PDFViewerApplication = {
...
-    if (typeof PDFJSDev === "undefined" || PDFJSDev.test("MOZCENTRAL")) {
+  /**
+   * For PDF documents that contain e.g. forms and javaScript, we should only
+   * trigger the fallback bar once the user has interacted with the page.
+   * @private
+   */
+  _delayedFallback(featureId) {
+    if (
+      typeof PDFJSDev === "undefined" ||
+      PDFJSDev.test("MOZCENTRAL || GENERIC")
```
timvandermeij, line 875: Telemetry is only implemented (and useful?) for `MOZCENTRAL` as far as I can tell, so why is `GENERIC` added as well?

web/app.js
```diff
@@ -878,24 +895,16 @@ const PDFViewerApplication = {
...
-      }
-
-      // For PDFs that contain script and form errors, we should only trigger
-      // the fallback once the user has interacted with the page.
-      if (this._delayedFallbackFeatureIds.length >= 1 && this._hasInteracted) {
-        featureId = this._delayedFallbackFeatureIds[0];
-        // Reset to prevent all click events from showing fallback bar.
-        this._delayedFallbackFeatureIds = [];
-      }
+      this.externalServices.reportTelemetry({
```
timvandermeij, line 898: Aren't forms and JavaScript now reported twice? They trigger `_delayedFallback` which unconditionally reports the telemetry and creates a callback that calls `fallback` which also reports the telemetry here.

web/app.js
Snuffleupagus: To actually get this code to run in more builds, and thus make it slightly better tested, since having functionality which is important in mozilla-central and which is currently only run in the development viewer seems suboptimal to me.

Originally the `FirefoxCom`-calls were placed inline throughout the viewer code, using pre-processor statements, but at this point there's really nothing stopping us from simply running this kind of code everywhere thanks to dummy methods on `DefaultExternalServices`. Should we perhaps just attempt to report telemetry *unconditionally* in all builds instead?

web/app.js
Snuffleupagus: Yes, but this isn't a problem neither here nor generally since the mozilla-central side already deals with duplicates; please see https://searchfox.org/mozilla-central/rev/e11d919cdc8a909354eb2c3e19904d9229c55d88/toolkit/components/pdfjs/content/PdfStreamConverter.jsm#433-499

web/app.js
timvandermeij: Yes, given that the dummies are in place, I have no problem running this unconditionally.

timvandermeij (approved):

timvandermeij:

PR #12057, opened 2020-07-04
Allow `BaseLocalCache` to, optionally, only allocate storage for caching of references (PR 12034 follow-up)

Snuffleupagus:
*Yet another instalment in the never-ending series of things that you think of __after__ a patch has landed.*

Since `Function`s are only cached by reference, we thus don't need to allocate storage for names in `LocalFunctionCache` instances. Obviously the effect of these changes are *really tiny*, but it seems reasonable in principle to avoid allocating data structures that are guaranteed to be unused.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12060, opened 2020-07-05
Convert the code in `src/core/evaluator.js` to use standard classes

Snuffleupagus:
- Move some constants and helper functions out of the `PartialEvaluator` closure

   This will simplify the `class` conversion in the next patch, and with modern JavaScript the moved code is still limited to the current module scope.

   *Please note:* For improved consistency with our usual formatting, the `TILING_PATTERN`/`SHADING_PATTERN` constants where re-factored slightly.

 - Convert the code in `src/core/evaluator.js` to use standard classes

   This removes additional `// eslint-disable-next-line no-shadow` usage, which our old pseudo-classes necessitated.

   Most of the re-formatting changes, after the `class` definitions and methods were fixed, were done automatically by Prettier.

   *Please note:* I'm purposely not doing any `var` to `let`/`const` conversion here, since it's generally better to (if possible) do that automatically on e.g. a directory basis instead.

---

*Much smaller and more manageable diff with https://github.com/mozilla/pdf.js/pull/12060/files?w=1*

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -163,28 +163,35 @@ function normalizeBlendMode(value, parsingArray = false) {
 }
 
 // Trying to minimize Date.now() usage and check every 100 time.
-var TIME_SLOT_DURATION_MS = 20;
-var CHECK_TIME_EVERY = 100;
-function TimeSlotManager() {
-  this.reset();
-}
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

Issue #11610, opened 2020-02-17
Inconsistent fontName value


umutesen:
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version: Chrome Version 80.0.3987.100 (Official Build) (64-bit)
- Operating system and its version: Windows 7
- PDF.js version: pdfjs-dist ^2.2.228 with Angular 6
- Is a browser extension: No

Steps to reproduce the problem:
1. Read pdf lines of the same document multiple times

```
public async readPdfLines(pdfUrl: string): Promise<any[]> {
    const pdf = await pdfjsLib.getDocument(pdfUrl).promise;
    const lines: any[] = [];

    for (let i = 1; i <= pdf._pdfInfo.numPages; i++) {
      const page = await pdf.getPage(i);
      const textContent = await page.getTextContent();
      textContent.items.forEach(line => lines.push(line));
    }

    return lines;
  }

this.downloadedPdfLines = await this.pdfReader.readPdfLines(url);

```

2. Font name associated with each line is inconsistent, the first read produce  `g_d0_f8`, second `g_d2_f26` etc. I am having to add a new or case every time I increase the runtime on the document.

        // First line of each cover in PDF
        const firstLines = this.downloadedPdfLines
            .filter(x => (x.fontName === "g_d0_f8" ||
                x.fontName === "g_d1_f17" ||
                x.fontName === "g_d2_f26" ||
                x.fontName === "g_d3_f35" ||
                x.fontName === "g_d4_f44" ||
                x.fontName === "g_d5_f53" ||
                x.fontName === "g_d6_f62") && x.height === 11);

What is the expected behavior? (add screenshot)
font name should always be the same for the same line of text

Snuffleupagus:
> font name should always be the same for the same line of text

That unfortunately cannot be done, since font names must be *unique* for every `getDocument` call to prevent errors if there's ever multiple documents opened in parallel. I.e. the `n` part in `g_d{n}_f1` will always be unique for each `getDocument` call, and this is thus working as expected.

umutesen:
Thank you for clarification, is there a way to disable this feature? In my case there is no requirement for opening documents in parallel.

Snuffleupagus:
> Thank you for clarification, is there a way to disable this feature?

No; and such an option isn't something that would be good to add either[1]. Please note that the current *unique* font names were added to address specific (and somewhat) reoccurring bugs with broken fonts.

[...]

PR #12070, opened 2020-07-07
Re-factor the `idFactory` functionality, used in the `core/`-code, and move the `fontID` generation into it

Snuffleupagus:
Note how the `getFontID`-method in `src/core/fonts.js` is *completely* global, rather than properly tied to the current document. This means that if you repeatedly open and parse/render, and then close, even the *same* PDF document the `fontID`s will still be incremented continuously.

For comparison the `createObjId` method, on `idFactory`, will always create a *consistent* id, assuming of course that the document and its pages are parsed/rendered in the same order.

In order to address this inconsistency, it thus seems reasonable to add a new `createFontId` method on the `idFactory` and use that when obtaining `fontID`s. (When the current `getFontID` method was added the `idFactory` didn't actually exist yet, which explains why the code looks the way it does.)
*Please note:* Since the document id is (still) part of the `loadedName`, it's thus not possible for different documents to have identical font names.

Implements the idea from https://github.com/mozilla/pdf.js/issues/11610#issuecomment-587099759

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -28,6 +28,7 @@ import {
   shadow,
   stringToBytes,
   stringToPDFString,
+  unreachable,
   Util,
   warn,
 } from "../shared/util.js";
@@ -71,6 +72,7 @@ class Page {
     pageIndex,
     pageDict,
     ref,
+    globalIdFactory,
     fontCache,
     builtInCMapCache,
     globalImageCache,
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #12076, opened 2020-07-08
Add form printing support

calixteman:
This PR adds support to print forms:
 - textfields;
 - checkbox;
 - radio.
I actively tested with the form f1040.pdf (https://www.irs.gov/pub/irs-pdf/f1040.pdf) and it works pretty well.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -30,7 +30,6 @@ import {
 import { Catalog, FileSpec, ObjectLoader } from "./obj.js";
 import { Dict, isDict, isName, isRef, isStream } from "./primitives.js";
 import { ColorSpace } from "./colorspace.js";
-import { EvalState } from "./evaluator.js";
 import { getInheritableProperty } from "./core_utils.js";
 import { OperatorList } from "./operator_list.js";
 import { StringStream } from "./stream.js";
@@ -1023,15 +1022,63 @@ class TextWidgetAnnotation extends WidgetAnnotation {
       return buf;
     }
 
+    if (this.data.multiLine) {
+      const renderedText = this.handleMultiline(
+        value,
+        font,
+        fontSize,
+        totalWidth,
+        alignment,
+        hPadding,
+        vPadding
+      );
+      const totalHeight = this.data.rect[3] - this.data.rect[1];
+      return `/Tx BMC q BT ${defaultAppearance} 1 0 0 1 0 ${totalHeight} Tm ${renderedText} ET Q EMC`;
+    }
+
     const alignment = this.data.textAlignment;
     if (alignment === 0 || alignment > 2) {
       // Left alignment: nothing to do
       return `/Tx BMC q BT ${defaultAppearance} 1 0 0 1 ${hPadding} ${vPadding} Tm (${value}) Tj ET Q EMC`;
     }
 
-    // We need to get the width of the text in order to align it correctly
-    const glyphs = font.charsToGlyphs(value);
+    const renderedText = this.renderPDFText(
+      value,
+      font,
+      fontSize,
+      totalWidth,
+      alignment,
+      hPadding,
+      vPadding
+    );
+    return `/Tx BMC q BT ${defaultAppearance} 1 0 0 1 0 0 Tm ${renderedText} ET Q EMC`;
+  }
 
+  async getFontData(evaluator, task) {
+    const operatorList = new OperatorList();
+    const initialState = {};
+    await evaluator.getOperatorList({
+      stream: new StringStream(this.data.defaultAppearance),
+      task,
+      resources: this.fieldResources,
+      operatorList,
+      initialState,
+    });
+
+    return [initialState.font, initialState.fontSize];
+  }
+
+  renderPDFText(
+    text,
+    font,
+    fontSize,
+    totalWidth,
+    alignment,
+    hPadding,
+    vPadding
+  ) {
+    // We need to get the width of the text in order to align it correctly
+    const glyphs = font.charsToGlyphs(text);
     const scale = fontSize / 1000;
     let width = 0;
     for (const glyph of glyphs) {
@@ -1042,26 +1089,101 @@ class TextWidgetAnnotation extends WidgetAnnotation {
     if (alignment === 1) {
       // Center
       shift = (totalWidth - width) / 2;
-    } else {
+    } else if (alignment === 2) {
       // Right
       shift = totalWidth - width - hPadding;
+    } else {
+      shift = hPadding;
     }
+    shift = shift.toFixed(2);
+    vPadding = vPadding.toFixed(2);
 
-    return `/Tx BMC q BT ${defaultAppearance} 1 0 0 1 ${shift} ${vPadding} Tm (${value}) Tj ET Q EMC`;
+    return `${shift} ${vPadding} Td (${text}) Tj`;
   }
 
-  async getFontData(evaluator, task) {
-    const operatorList = new OperatorList();
-    const initialState = new EvalState();
-    await evaluator.getOperatorList({
-      stream: new StringStream(this.data.defaultAppearance),
-      task,
-      resources: this.fieldResources,
-      operatorList,
-      initialState,
-    });
+  handleMultiline(text, font, fontSize, width, alignment, hPadding, vPadding) {
+    const lines = text.replace("\r\n", "\n").split("\n");
+    let buf = "";
+    const totalWidth = alignment === 1 ? width : width - hPadding;
+    for (const line of lines) {
+      const chunks = this.splitLine(line, font, fontSize, totalWidth);
+      for (const chunk of chunks) {
+        if (buf === "") {
+          buf = this.renderPDFText(
+            chunk,
+            font,
+            fontSize,
+            width,
+            alignment,
+            hPadding,
+            -fontSize
+          );
+        } else {
+          buf +=
+            "\n" +
+            this.renderPDFText(
+              chunk,
+              font,
+              fontSize,
+              width,
+              alignment,
+              0,
+              -fontSize
+            );
+        }
+      }
+    }
 
-    return [initialState.font, initialState.fontSize];
+    return buf;
+  }
+
+  splitLine(line, font, fontSize, width) {
+    const scale = fontSize / 1000;
+    const white = font.charsToGlyphs(" ", true)[0].width * scale;
+    const chunks = [];
+
+    let lastSpacePos = -1,
+      startChunk = 0,
+      currentWidth = 0;
+
+    for (let i = 0; i < line.length; i++) {
+      const character = line.charAt(i);
+      if (character === " ") {
+        if (currentWidth + white > width) {
+          // We can break here
+          chunks.push(line.substring(startChunk, i));
+          startChunk = i;
+          currentWidth = white;
+          lastSpacePos = -1;
+        } else {
+          currentWidth += white;
+          lastSpacePos = i;
+        }
+      } else {
+        const charWidth = font.charsToGlyphs(character, false)[0].width * scale;
+        if (currentWidth + charWidth > width) {
+          // We must break to the last white position (if one)
+          if (lastSpacePos !== -1) {
+            chunks.push(line.substring(startChunk, lastSpacePos + 1));
+            startChunk = i = lastSpacePos + 1;
+            lastSpacePos = -1;
+            currentWidth = 0;
+          } else {
+            // Just break in the middle of the word
+            chunks.push(line.substring(startChunk, i));
+            currentWidth = charWidth;
+          }
+        } else {
+          currentWidth += charWidth;
+        }
+      }
+    }
+
+    if (startChunk < line.length) {
+      chunks.push(line.substring(startChunk, line.length));
+    }
+
+    return chunks;
   }
 }
 
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -923,6 +923,7 @@ class TextWidgetAnnotation extends WidgetAnnotation {
       maximumLength = null;
     }
     this.data.maxLen = maximumLength;
+    this.isPassword = this.hasFieldFlag(AnnotationFieldFlag.PASSWORD);
 
     // Process field flags for the display layer.
     this.data.multiLine = this.hasFieldFlag(AnnotationFieldFlag.MULTILINE);
@@ -986,7 +987,9 @@ class TextWidgetAnnotation extends WidgetAnnotation {
   }
 
   async getAppearance(evaluator, task, annotationStorage) {
-    if (!annotationStorage) {
+    // If it's a password textfield then no rendering to avoid to leak it.
+    // see 12.7.4.3, table 228
+    if (!annotationStorage || this.data.isPassword) {
       return null;
     }
     const value = annotationStorage[this.data.id] || "";
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -23,7 +23,6 @@ import {
   assert,
   isString,
   OPS,
-  stringToBytes,
   stringToPDFString,
   Util,
   warn,
@@ -33,7 +32,7 @@ import { Dict, isDict, isName, isRef, isStream } from "./primitives.js";
 import { ColorSpace } from "./colorspace.js";
 import { getInheritableProperty } from "./core_utils.js";
 import { OperatorList } from "./operator_list.js";
-import { Stream } from "./stream.js";
+import { StringStream } from "./stream.js";
 
 class AnnotationFactory {
   /**
@@ -81,7 +80,7 @@ class AnnotationFactory {
       subtype,
       id,
       pdfManager,
-      acroForm,
+      acroForm: acroForm instanceof Dict ? acroForm : Dict.empty,
     };
 
     switch (subtype) {
@@ -219,7 +218,6 @@ function getTransformMatrix(rect, bbox, matrix) {
 
   const xRatio = (rect[2] - rect[0]) / (maxX - minX);
   const yRatio = (rect[3] - rect[1]) / (maxY - minY);
-
   return [
     xRatio,
     0,
@@ -476,7 +474,7 @@ class Annotation {
    */
   setAppearance(dict) {
     this.appearance = null;
-    this.checkAppearance = null;
+    this.checkedAppearance = null;
 
     const appearanceStates = dict.get("AP");
     if (!isDict(appearanceStates)) {
@@ -499,7 +497,6 @@ class Annotation {
     if (!isName(as) || !normalAppearanceState.has(as.name)) {
       return;
     }
-
     this.appearance = normalAppearanceState.get(as.name);
   }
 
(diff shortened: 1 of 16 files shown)


web/app_options.js
```diff
@@ -127,7 +127,7 @@ const defaultOptions = {
   },
   renderInteractiveForms: {
     /** @type {boolean} */
-    value: false,
+    value: true,
```
timvandermeij, line 130: This change should be reverted because we should first test for a while and make sure that all form elements can be printed properly before enabling this for all users.

src/display/canvas.js
```diff
@@ -1598,6 +1599,8 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       var defaultVMetrics = font.defaultVMetrics;
       var widthAdvanceScale = fontSize * current.fontMatrix[0];
 
+      // console.log(glyphs, fontSize, current.fontMatrix[0], fontSizeScale);
```
timvandermeij, line 1602: The changes in this file should be removed (debug print and irrelevant whitespace change).

web/annotation_storage.js
```diff
@@ -0,0 +1,69 @@
...
+  constructor(storage = {}) {
+    this._storage = storage;
+  }
+
+  /**
+   * Get the value for a given key
+   *
+   * @public
+   * @memberof AnnotationStorage
+   * @param {Object} key
```
timvandermeij, line 26: Aren't the keys always `string` types (here and below). If possible, these types should be as specific as possible.

src/core/annotation.js
```diff
@@ -970,6 +1082,31 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     }
   }
 
+  getOperatorList(evaluator, task, renderForms, annotationStorage) {
+    if (annotationStorage) {
+      const value = annotationStorage[this.data.id] || false;
+      if (value && this.checkedAppearance) {
+        const savedAppearance = this.appearance;
+        this.appearance = this.checkedAppearance;
+        const opL = super.getOperatorList(
```
timvandermeij, line 1091: Let's rename `opL` to `operatorList` for readability.

src/core/annotation.js
```diff
@@ -471,6 +476,7 @@ class Annotation {
    */
   setAppearance(dict) {
     this.appearance = null;
+    this.checkAppearance = null;
```
timvandermeij, line 479: I think this should be `this.checkedAppearance = null;` given the other occurrences.

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
       this.data.maxLen !== null;
   }
 
-  getOperatorList(evaluator, task, renderForms) {
-    if (renderForms || this.appearance) {
-      return super.getOperatorList(evaluator, task, renderForms);
-    }
+  getOperatorList(evaluator, task, renderForms, annotationStorage) {
+    const self = this;
```
timvandermeij, line 939: If arrow functions are used, we shouldn't need variables like `self` because the `this` scope is propagated then.

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+            task,
+            renderForms,
+            annotationStorage
+          );
+        }
 
-    const operatorList = new OperatorList();
+        const operatorList = new OperatorList();
+
+        // Even if there is an appearance stream, ignore it. Self is the
```
timvandermeij, line 953: `Self` -> `This`; looks like this was automatically renamed ;-)

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+          .then(function () {
+            operatorList.addOp(OPS.endAnnotation, []);
+            return operatorList;
+          });
+      }
+    );
+  }
+
+  getAppearance(evaluator, task, annotationStorage) {
+    // TODO: handle password: we mustn't write it in clear
```
timvandermeij, line 991: How exactly are passwords involved here? I only know of a password for password-protected PDF files, but once that's entered the forms themselves should not contain password fields, unless I missed this in the spec?

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+    if (!annotationStorage) {
+      return Promise.resolve("");
```
timvandermeij, line 993: Instead of these empty strings, I would it would be nicer to return a string when the appearance is valid and `null` otherwise. This is what we do in other parts of the codebase too to make it more clear to the caller that the value is not to be used.

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+    }
+    const value = annotationStorage[this.data.id] || "";
+    if (value === "") {
+      return Promise.resolve("");
+    }
+    const x = 2;
```
timvandermeij, line 999: 2 is a bit of a magic number here. Can we add a comment to explain why 2 is chosen here? Is it some sort of padding (it's also used below)?

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+    const y = 2;
+    if (this.data.comb) {
+      const width = this.data.rect[2] - this.data.rect[0];
+      const combWidth = (width / this.data.maxLen).toFixed(2);
+      let buf = `/Tx BMC q BT ${this.data.defaultAppearance} 1 0 0 1 ${x} ${y} Tm`;
+      let first = true;
+      for (const c of value) {
```
timvandermeij, line 1006: I assume `c` is for `character`? Can we rename it to that for readability (or at least to `char`)?

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+      buf += " ET Q EMC";
+      return Promise.resolve(buf);
+    }
 
-    // Even if there is an appearance stream, ignore it. This is the
-    // behaviour used by Adobe Reader.
-    if (!this.data.defaultAppearance) {
-      return Promise.resolve(operatorList);
+    const defaultAppearance = this.data.defaultAppearance;
+    const app = `/Tx BMC q BT ${defaultAppearance} 1 0 0 1 ${x} ${y} Tm (${value}) Tj ET Q EMC`;
```
timvandermeij, line 1019: Let's call this `appearance` because `app` may indicate a different meaning. If the line becomes too long, it can be split over multiple lines with the template literals.

src/core/annotation.js
```diff
@@ -942,15 +1035,34 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
       .then(function () {
-        return operatorList;
+        const fns = operatorList.fnArray;
+        const args = operatorList.argsArray;
+        const fontSize = args[fns.findIndex(f => f === OPS.setFont)][1];
+        const glyphs = args[fns.findIndex(f => f === OPS.showText)][0];
+
+        const scale = fontSize / 1000;
+        let width = 0;
+        for (const g of glyphs) {
```
timvandermeij, line 1045: `g` -> `glyph` (we try to avoid one-letter variables for readability)

timvandermeij (commented):
These comments are from an initial pass over the patch. Obviously it needs a more in depth review and more testing, but I must say that it looks a lot like what I had in mind for the printing support. Thank you for working on this!

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+          })
+          .then(function () {
+            operatorList.addOp(OPS.endAnnotation, []);
+            return operatorList;
+          });
+      }
+    );
+  }
+
+  getAppearance(evaluator, task, annotationStorage) {
```
Snuffleupagus, line 990: Nit: If this method is made `async`, you can return directly without having to wrap every return value with `Promise.resolve()`.

src/display/api.js
```diff
@@ -381,6 +381,7 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
     source.initialData = pdfDataRangeTransport.initialData;
     source.progressiveDone = pdfDataRangeTransport.progressiveDone;
   }
+
```
Snuffleupagus, line 384: Unrelated white-space change, here and below, please revert.

test/unit/annotation_spec.js
```diff
@@ -37,6 +37,11 @@ describe("annotation", function () {
   class PDFManagerMock {
     constructor(params) {
       this.docBaseUrl = params.docBaseUrl || null;
+      this.pdfDocument = {
+        acroForm: () => {
+          return { get: x => undefined };
+        },
```
Snuffleupagus, line 43: The `acroForm` should contain an actual `Dict`, rather than a regular object. Hence please change this to: 
```suggestion
        acroForm: new Dict(),
```

web/annotation_layer_builder.js
```diff
@@ -124,6 +127,7 @@ class DefaultAnnotationLayerFactory {
   createAnnotationLayerBuilder(
     pageDiv,
     pdfPage,
+    annotationStorage,
```
Snuffleupagus, line 130: Please note that this requires changes here as well: https://github.com/mozilla/pdf.js/blob/72d71ba6a556093a1c88525e827e59590c9151f1/web/interfaces.js#L161-L181

(This change alone probably qualifies this patch for the `[api-minor]` tag, although the general functionality added probably does so as well.)

web/app.js
```diff
@@ -187,6 +188,7 @@ const PDFViewerApplication = {
   url: "",
   baseUrl: "",
   externalServices: DefaultExternalServices,
+  annotationStorage: new AnnotationStorage(),
```
Snuffleupagus, line 191: Assuming that the `annotationStorage` is document-specific, you'll have to initialize this for each file opened in the viewer rather than globally like this (since the default viewer supports opening more than the initial document).

Snuffleupagus (changes requested):

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+        ];
+
+        const transform = getTransformMatrix(self.data.rect, bbox, matrix);
+        operatorList.addOp(OPS.beginAnnotation, [
+          self.data.rect,
+          transform,
+          matrix,
+        ]);
+
+        const stream = new Stream(stringToBytes(content));
```
Snuffleupagus, line 974: If you import `StringStream` from https://github.com/mozilla/pdf.js/blob/master/src/core/stream.js, you can instead do
```suggestion
        const stream = new StringStream(content);
```

src/core/annotation.js
```diff
@@ -920,20 +935,98 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+    const alignment = this.data.textAlignment;
+    if (alignment === 0 || alignment > 2) {
+      // Left alignment: nothing to do
+      return Promise.resolve(app);
     }
 
-    const stream = new Stream(stringToBytes(this.data.defaultAppearance));
+    // We need to get the width of the text in order to align it correctly
+    const operatorList = new OperatorList();
+    const stream = new Stream(stringToBytes(app));
```
Snuffleupagus, line 1028: Similar to https://github.com/mozilla/pdf.js/pull/12076/files#r452125591

src/core/annotation.js
calixteman: Agreed.
poppler is using 2:
https://gitlab.freedesktop.org/poppler/poppler/-/blob/master/poppler/Annot.cc#L4411
And I'll fix the vPadding in a next patch based on fontSize

calixteman:
Thanks a lot for your review: I really appreciate.
I need to fix the default vertical padding and handle multilines text.

web/app.js
calixteman: Good catch: I didn't know that... I'm a kind of newbie.

src/core/annotation.js
```diff
@@ -920,37 +933,141 @@ class TextWidgetAnnotation extends WidgetAnnotation {
...
+
+  async getFontData(evaluator, task) {
     const operatorList = new OperatorList();
+    const initialState = new EvalState();
+    await evaluator.getOperatorList({
+      stream: new StringStream(this.data.defaultAppearance),
+      task,
+      resources: this.fieldResources,
+      operatorList,
+      initialState,
```
Snuffleupagus, line 1061: Please note the default value in https://github.com/mozilla/pdf.js/blob/72d71ba6a556093a1c88525e827e59590c9151f1/src/core/evaluator.js#L1224-L1235 hence you shouldn't need to pass the `initialState` here (nor export `EvalState` in the other file).

src/core/evaluator.js
```diff
@@ -783,6 +783,7 @@ class PartialEvaluator {
       })
       .then(translated => {
         state.font = translated.font;
+        state.fontSize = fontArgs[1];
```
Snuffleupagus, line 786: The `fontArgs` isn't guaranteed to be defined, thus causing this to throw in some cases, please define `fontSize` similar to the existing `fontName` handling further up in this method.

src/core/annotation.js
calixteman: I removed EvalState export but I want to pass an initialState to be able to get the font and the font size from it in order to compute text width (and split it for multiline textfields).

calixteman: If you want to chat about that or anything else, then I'm on riot in the pdf.js channel, my nick is "calixte" (in CEST timezone).

Snuffleupagus: Sorry, I completely overlooked that you actually accessed the state afterwards; in that case exporting/using it like you previously did is probably OK.

(Although it does feel slightly "hacky" to do things in that way, given how the `getOperatorList` method in currently implemented/used, however at the top of my head I don't really have a much better suggestion.)

src/core/evaluator.js
```diff
@@ -3519,7 +3521,6 @@ class EvalState {
   constructor() {
     this.ctm = new Float32Array(IDENTITY_MATRIX);
     this.font = null;
-    this.fontSize = 0;
```
Snuffleupagus: You definitely want to keep this line though, to ensure that there's always a meaningful default value.

(The changes in the `handleSetFont` method looks great, by the way.)

calixteman:
@timvandermeij @Snuffleupagus, could you review this PR please ?

src/core/annotation.js
```diff
@@ -877,17 +888,315 @@ class WidgetAnnotation extends Annotation {
...
+      alignment,
+      hPadding,
+      vPadding
+    );
+    return `/Tx BMC q BT ${defaultAppearance} 1 0 0 1 0 0 Tm ${renderedText} ET Q EMC`;
+  }
+
+  async getFontData(evaluator, task) {
+    const operatorList = new OperatorList();
+    const initialState = new EvalState();
```
Snuffleupagus, line 1045: Really sorry, but I'm going to have to come back to my previous comments about how using an `initialState` in this way is (for a lack of a better word) "abusing" the `getOperatorList` method in ways that were never intended and why it unfortunately isn't *guaranteed* to always work.

It would actually be fairly easy to create a *theoretically valid* PDF document, where the `intentState` would still be *empty* even after parsing of the `defaultAppearance` has completed. Basically, simply wrapping the `defaultAppearance` in save/restore operators would break the code below which relies on `initialState` to contain the desired properties (note how the `StateManager`, by necessity, deals with save/restore ops).

calixteman: So, what could be a good way to get font and font size ?
Or maybe I can use a special state which clones to itself and then get the last used font from it, wdyt ?

Snuffleupagus: > So, what could be a good way to get font and font size ?

I honestly don't know, off the top of my head, since I don't believe that the `getOperatorList` functionality was really created for the particular use-cases you have mind here.

> Or maybe I can use a special state which clones to itself and then get the last used font from it, wdyt ?

Assuming it won't negatively impact performance of the "normal" PDF document parsing, that could possibly work. (Although it still, to me, feels somewhat unfortunate that we're having to resort to these sort of tricks :-)  

---

Of course, you might also want to see check if @brendandahl has any better ideas (than me).

web/firefox_print_service.js
```diff
@@ -109,7 +110,7 @@ PDFPrintServiceFactory.instance = {
     return shadow(this, "supportsPrinting", value);
   },
 
-  createPrintService(pdfDocument, pagesOverview, printContainer) {
+  createPrintService(pdfDocument, pagesOverview, printContainer, l10n) {
```
timvandermeij, line 113: I don't really see why this change is necessary?

src/core/annotation.js
```diff
@@ -877,17 +888,315 @@ class WidgetAnnotation extends Annotation {
...
+    return super.getOperatorList(
+      evaluator,
+      task,
+      renderForms,
+      annotationStorage
+    );
+  }
+}
+
+class WidgetWithTextAnnotation extends WidgetAnnotation {
```
timvandermeij, line 906: The classes in this file directly match the specification which makes it useful to map it one-to-one. I would therefore not introduce a new base class here, mainly because it adds yet another level of inheritance which makes it harder to follow, but also because this is not a class that the specification contains. Instead, we can simply put these methods on the `WidgetAnnotation` class itself and overwrite them in subclasses if necessary. This way we have one less level of inheritance while still keeping the functionality.

timvandermeij (commented):
Nice work so far!

This has become quite a big patch now compared to how it started, which makes sense because it developed along the way, but it makes it hard to make sure everything is alright in review. If this works correctly, we should split it up into multiple smaller PRs that we can more easily review and test/merge. In particular because we have no good way of testing printing other than manually, it would be nice to at least have unit tests for the parts that we can test. Since that would make the patch even bigger, I'm thinking splitting it is a good idea. For example, we could split this up into three PRs:

- One PR that only introduces the storage, with unit tests, and passes it to the necessary components, but don't actually do anything with it yet.
- One PR that basically contains the changes in `src/core/annotation.js` since they should be self-contained, with unit tests because that should be relatively easy to unit tests (see the existing unit tests for this).
- Finally, one PR that basically contains the changes in `src/display/annotation_layer.js` and the remaining changes.

However, it might make more sense to make the split per field type:

- One PR for the storage with unit tests again.
- One PR for the core/display layer changes for radio buttons.
- One PR like the above but for checkboxes.
- Finally, one PR for the above but for text fields.

That way it's much more self-contained and makes merging much quicker since it allows us to better oversee the changes, especially with good unit test coverage in place.

calixteman:
I splited this PR in multiple ones:
https://github.com/mozilla/pdf.js/pull/12106
https://github.com/mozilla/pdf.js/pull/12107
https://github.com/mozilla/pdf.js/pull/12108
https://github.com/mozilla/pdf.js/pull/12112

timvandermeij:
Closing this one in favor of the split up PRs instead.

PR #12083, opened 2020-07-10
Remove the IR (internal representation) part of the ColorSpace parsing

Snuffleupagus:
Originally ColorSpaces were only *partially* parsed on the worker-thread, to obtain an IR-format which was sent to the main-thread. This had the somewhat unfortunate side-effect of causing the majority of the (potentially heavy) ColorSpace parsing to happen on the main-thread.
Hence PR #4824 which, among other things, changed ColorSpaces to be *fully* parsed on the worker-thread with only RGB-data being sent to the main-thread.

While it thus originally was necessary to have `ColorSpace.{parseToIR, fromIR}` methods, to handle the worker/main-thread split, that's no longer the case and we can thus reduce all of the ColorSpace parsing to one method instead.

Currently, when parsing a ColorSpace, we call `ColorSpace.parseToIR` which parses the ColorSpace-data from the document and then creates the IR-format. We then, immediately, call `ColorSpace.fromIR` which parses the IR-format and then finally creates the actual ColorSpace.[1]
All-in-all, this leads to a fair amount of unnecessary indirection which also (in my opinion) makes the code less clear.

Obviously these changes are not really expected to have a significant effect on performance, especially with the recently added caching of ColorSpaces, however there'll now be strictly fewer function calls, less memory allocated, and overall less parsing required during ColorSpace-handling.

---
[1] For ICCBased ColorSpaces, given the validation necessary, this currently even leads to parsing an /Alternate ColorSpace *twice*.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -336,8 +336,12 @@ class ColorSpace {
           "before calling `ColorSpace.parseAsync`."
       );
     }
-    const IR = this.parseToIR(cs, xref, resources, pdfFunctionFactory);
-    const parsedColorSpace = this.fromIR(IR);
+    const parsedColorSpace = this._parse(
+      cs,
+      xref,
+      resources,
+      pdfFunctionFactory
+    );
 
     // Attempt to cache the parsed ColorSpace, by name and/or reference.
     this._cache(cs, xref, localColorSpaceCache, parsedColorSpace);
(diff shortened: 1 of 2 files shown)


src/core/colorspace.js
```diff
@@ -696,22 +639,18 @@ class IndexedCS extends ColorSpace {
...
     if (isStream(lookup)) {
-      this.lookup = new Uint8Array(length);
       const bytes = lookup.getBytes(length);
       this.lookup.set(bytes);
-    } else if (isString(lookup)) {
-      this.lookup = new Uint8Array(length);
+    } else if (typeof lookup === "string") {
       for (let i = 0; i < length; ++i) {
-        this.lookup[i] = lookup.charCodeAt(i);
+        this.lookup[i] = lookup.charCodeAt(i) & 0xff;
```
Snuffleupagus, line 650: This is consistent with https://github.com/mozilla/pdf.js/blob/38cf49bb8181c4c8b8effef44fbd9d4068fb0d17/src/shared/util.js#L499

timvandermeij (approved):

timvandermeij:
Thanks for improving this!

PR #12085, opened 2020-07-11
Make the detection of Node.js environments on Electron strict.

tamuratak:
Make the detection of Node.js environments on Electron strict. The main process and its child processes should be detected as Node.js environments.

- https://www.electronjs.org/docs/api/process#processtype-readonly

Without this PR, `pdfjsLib.getDocument(...).promise.then` fails with an error message `Setting up fake worker failed: "No "GlobalWorkerOptions.workerSrc" specified.".` on the main process and its child processes of Electron.

### DIFF
--- a/src/shared/is_node.js
+++ b/src/shared/is_node.js
@@ -16,11 +16,12 @@
 
 // NW.js / Electron is a browser context, but copies some Node.js objects; see
 // http://docs.nwjs.io/en/latest/For%20Users/Advanced/JavaScript%20Contexts%20in%20NW.js/#access-nodejs-and-nwjs-api-in-browser-context
-// https://electronjs.org/docs/api/process#processversionselectron
+// https://www.electronjs.org/docs/api/process#processversionselectron-readonly
+// https://www.electronjs.org/docs/api/process#processtype-readonly
 const isNodeJS =
   typeof process === "object" &&
   process + "" === "[object process]" &&
   !process.versions.nw &&
-  !process.versions.electron;
+  !(process.versions.electron && process.type && process.type !== "browser");
 
 export { isNodeJS };


src/shared/is_node.js
```diff
@@ -16,11 +16,12 @@
 
 // NW.js / Electron is a browser context, but copies some Node.js objects; see
 // http://docs.nwjs.io/en/latest/For%20Users/Advanced/JavaScript%20Contexts%20in%20NW.js/#access-nodejs-and-nwjs-api-in-browser-context
-// https://electronjs.org/docs/api/process#processversionselectron
+// https://www.electronjs.org/docs/api/process#processversionselectron
```
Snuffleupagus, line 19: If you're touching this line anyway, please also fix the link to point directly to the relevant section: https://www.electronjs.org/docs/api/process#processversionselectron-readonly

src/shared/is_node.js
```diff
@@ -16,11 +16,12 @@
...
+// https://www.electronjs.org/docs/api/process#processtype-readonly
 const isNodeJS =
   typeof process === "object" &&
   process + "" === "[object process]" &&
   !process.versions.nw &&
-  !process.versions.electron;
+  !(process.versions.electron && process.type && process.type !== "browser");
```
Snuffleupagus, line 25: Quoting from https://www.electronjs.org/docs/api/process#processtype-readonly:
> A String representing the current process's type, can be "browser" (i.e. main process), "renderer", or "worker" (i.e. web worker).

Hence my question: Will `process.type !== "browser"` actually do the right thing when this code is run on the worker-thread in a browser?

tamuratak: Which is what you are saying with the sentence, 
> when this code is run on the worker-thread in a browser?":
- Running `build/pdf.wroker.js` on the worker-thread in a browser.
- Running `build/pdf.js` with OffscreenCanvas on the worker-thread in a browser.

In either case, it should work properly. I will check.

Snuffleupagus: > Running `build/pdf.worker.js` on the worker-thread in a browser.

The changes you're making in this patch only applies to environments that pretend to be Node.js, when in reality they are not.
Hence my question pertains to using the Electron-framework in *a browser*, with workers enabled, and ensuring that e.g. https://github.com/mozilla/pdf.js/blob/7dabc5ecc8347fbda8ce4a219c182790d9ec9187/src/core/worker.js#L676-L684 works as expected.

> Running `build/pdf.js` with OffscreenCanvas on the worker-thread in a browser.

We don't support OffscreeCanvas, so that doesn't really matter right now.

src/shared/is_node.js
tamuratak: I have edited as requested.

src/shared/is_node.js
tamuratak: I have created a repository to check this PR.
- https://github.com/tamuratak/electron-pdfjs-sample

```
git clone git@github.com:tamuratak/electron-pdfjs-sample.git
cd electron-pdfjs-sample 
npm i  (Install Electron locally)
npx electron app/
```

Then the PDF.js viewer launches. You can see through the dev tools that WebWoker works well. In both cases that `nodeIntegration` and `nodeIntegrationInWorker` are `true` or `false`, we can see that it woks well.

All the files at `app/src/viewer` are copied from `pdfjs-2.4.456-dist.zip`.  I have edited `app/src/viewer/pdf.js` and `app/src/viewer/pdf.worker.js` as this PR. I have also edited `app/src/viewer/viewer.js` and `app/src/viewer/viewer.html`.



```diff
$  diff -u ../tmp/pdfjs-2.4/build/pdf.js app/src/viewer/pdf.js 
--- ../tmp/pdfjs-2.4/build/pdf.js	2020-03-19 22:43:52.000000000 +0900
+++ app/src/viewer/pdf.js	2020-07-12 10:58:53.000000000 +0900
@@ -4166,7 +4166,7 @@
   value: true
 });
 exports.isNodeJS = void 0;
-const isNodeJS = typeof process === "object" && process + "" === "[object process]" && !process.versions["nw"] && !process.versions["electron"];
+const isNodeJS = typeof process === "object" && process + "" === "[object process]" && !process.versions["nw"] && !(process.versions.electron && process.type && process.type !== "browser");
 exports.isNodeJS = isNodeJS;
 
 /***/ }),
```

```diff
$  diff -u ../tmp/pdfjs-2.4/build/pdf.worker.js app/src/viewer/pdf.worker.js 
--- ../tmp/pdfjs-2.4/build/pdf.worker.js	2020-03-19 22:44:08.000000000 +0900
+++ app/src/viewer/pdf.worker.js	2020-07-12 10:58:53.000000000 +0900
@@ -45266,7 +45266,7 @@
   value: true
 });
 exports.isNodeJS = void 0;
-const isNodeJS = typeof process === "object" && process + "" === "[object process]" && !process.versions["nw"] && !process.versions["electron"];
+const isNodeJS = typeof process === "object" && process + "" === "[object process]" && !process.versions["nw"] && !(process.versions.electron && process.type && process.type !== "browser");
 exports.isNodeJS = isNodeJS;
 
 /***/ }),
```

```diff
$  diff -u ../tmp/pdfjs-2.4/web/viewer.js app/src/viewer/viewer.js
--- ../tmp/pdfjs-2.4/web/viewer.js	2020-03-19 22:43:56.000000000 +0900
+++ app/src/viewer/viewer.js	2020-07-12 10:58:53.000000000 +0900
@@ -3629,11 +3629,11 @@
     kind: OptionKind.API
   },
   workerPort: {
-    value: null,
+    value: new Worker("pdf.worker.js"),
     kind: OptionKind.WORKER
   },
   workerSrc: {
-    value: "../build/pdf.worker.js",
+    value: "pdf.worker.js",
     kind: OptionKind.WORKER
   }
 };
```

```diff
$  diff -u ../tmp/pdfjs-2.4/web/viewer.html app/src/viewer/viewer.html 
--- ../tmp/pdfjs-2.4/web/viewer.html	2020-03-19 22:43:44.000000000 +0900
+++ app/src/viewer/viewer.html	2020-07-12 10:58:53.000000000 +0900
@@ -34,7 +34,7 @@
 
 <!-- This snippet is used in production (included from viewer.html) -->
 <link rel="resource" type="application/l10n" href="locale/locale.properties">
-<script src="../build/pdf.js"></script>
+<script src="pdf.js"></script>
 
 
     <script src="viewer.js"></script>
```

timvandermeij (approved):

timvandermeij:

PR #12087, opened 2020-07-11
Add local caching of "simple" Graphics State (ExtGState) data in `PartialEvaluator.{getOperatorList, getTextContent}` (issue 2813)

Snuffleupagus:
- Add local caching of "simple" Graphics State (ExtGState) data in `PartialEvaluator.getOperatorList` (issue 2813)

   This patch will help pathological cases the most, with issue #2813 being a particularily problematic example. While there's only *four* `/ExtGState` resources, there's a total `29062` of `setGState` operators. Even though parsing of a single `/ExtGState` resource is quite fast, having to re-parse them thousands of times does add up quite significantly.

   For simplicity we'll only cache "simple" `/ExtGState` resource, since e.g. the general `SMask` case cannot be easily cached (without re-factoring other code, which may have undesirable effects on general parsing).

   By caching "simple" `/ExtGState` resource, we thus improve performance by:
    - Not having to fetch/validate/parse the same `/ExtGState` data over and over.
    - Handling of repeated `setGState` operators becomes *synchronous* during the `OperatorList` building, instead of having to defer to the event-loop/microtask-queue since the `/ExtGState` parsing is done asynchronously.

   ---

   Obviously I had intended to include (standard) benchmark results with this patch, but for reasons I don't understand the test run-time (even with `master`) of the document in issue #2813 is *a lot* slower than in the development viewer (making normal benchmarking infeasible).
However, testing this manually in the development viewer (using `pdfBug=Stats`) shows a *reduction* of `~10 %` in the rendering time of the PDF document in issue #2813.

 - Add local caching of non-font Graphics State (ExtGState) data in `PartialEvaluator.getTextContent`

   It turns out that `getTextContent` suffers from *similar* problems with repeated GStates as `getOperatorList`; please see the previous patch.

   While only `/ExtGState` resources containing Fonts will actually be *parsed* by `PartialEvaluator.getTextContent`, we're still forced to fetch/validate repeated `/ExtGState` resources even though *most* of them won't affect the textContent (since they mostly contain purely graphical state).

   With these changes we also no longer need to immediately reset the current text-state when encountering a `setGState` operator, which may thus improve text-selection in some cases.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -75,7 +75,11 @@ import {
 import { getTilingPatternIR, Pattern } from "./pattern.js";
 import { isPDFFunction, PDFFunctionFactory } from "./function.js";
 import { Lexer, Parser } from "./parser.js";
-import { LocalColorSpaceCache, LocalImageCache } from "./image_utils.js";
+import {
+  LocalColorSpaceCache,
+  LocalGStateCache,
+  LocalImageCache,
+} from "./image_utils.js";
 import { bidi } from "./bidi.js";
 import { ColorSpace } from "./colorspace.js";
 import { DecodeStream } from "./stream.js";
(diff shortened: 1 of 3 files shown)


timvandermeij:

PR #12089, opened 2020-07-12
Convert `RefSetCache` to a proper class and to use a `Map` internally

timvandermeij:
This is a follow-up on https://github.com/mozilla/pdf.js/pull/11977#issuecomment-653795328.

### DIFF
--- a/src/core/primitives.js
+++ b/src/core/primitives.js
@@ -239,46 +239,41 @@ class RefSet {
   }
 }
 
-var RefSetCache = (function RefSetCacheClosure() {
-  // eslint-disable-next-line no-shadow
-  function RefSetCache() {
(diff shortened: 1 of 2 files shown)


Snuffleupagus (approved):
Can you please *also* unit-test the `RefSetCache.has(...) === false` case, since unless I'm missing something that case isn't covered here.

Also, please run the *entire* test-suite just to ensure that there's no overall negative effects, performance wise, from these changes. I don't really expect there to be, given how/where the `RefSetCache` is used, but just in case. (Recently I looked into using a `Map`, rather than a regular Object, in `Dict` and *that* did seem to regress some documents.)

---

r=me, with the above addressed; thank you!

PR #12090, opened 2020-07-12
Stop special-casing the (very unlikely) "no `/XObject` found"-scenario, when parsing `OPS.paintXObject` operators, in `PartialEvaluator.{getOperatorList, getTextContent}`

Snuffleupagus:
Originally there weren't any (generally) good ways to handle errors gracefully, on the worker-side, however that's no longer the case and we can simply fallback to the existing `ignoreErrors` functionality instead.
Also, please note that the "no `/XObject` found"-scenario should be *extremely* unlikely in practice and would only occur in corrupt/broken documents.

Note that the `PartialEvaluator.getOperatorList` case is especially bad currently, since we'll simply (attempt to) send the data as-is to the main-thread. This is quite bad, since in a corrupt/broken document the data *could* contain anything and e.g. be unclonable (which would cause breaking errors).
Also, we're (obviously) not attempting to do anything with this "raw" `OPS.paintXObject` data on the main-thread and simply ensuring that we never send it definately seems like the correct approach.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1331,11 +1331,6 @@ class PartialEvaluator {
                   xobj = xref.fetch(xobj);
                 }
 
-                if (!xobj) {
-                  operatorList.addOp(fn, args);
-                  resolveXObject();
-                  return;
-                }
                 if (!isStream(xobj)) {
                   throw new FormatError("XObject should be a stream");
                 }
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Nice find!

Issue #11941, opened 2020-05-29
minified-es5 fails to "load" / correctly handle defective pdf-files


pfeldbaumer:
Configuration:
  latest master / beta 2.4.456
  minified-es5 build
  various (chromium) based browser-versions (from .49 to .81)

Steps to reproduce the problem:
  call pdfjs.getDocument(path_to_invalid_pdf_file) with minified-es5 pdf.js and pdf.worker.js
  use e.g. any text-document as "path_to_invalid_pdf_file" (so a valid file, but no pdf-document)

Expected behavior:
  PDFDocumentLoadingTask.promise should get rejected with an InvalidPDFException as parameter

Observed behaviour:
  code throws an (unhandled) exception "Unhandled promise rejection TypeError: Cannot read property 'message' of undefined"

In current minified-es5 builds calling .getDocument(path_to_invalid_pdf_file) does not correctly "work", it leads to:
Unhandled promise rejection TypeError: Cannot read property 'message' of undefined
    at main.js:...
    at lib/pdf.min.js:22:64582
    at MutationObserver.n (lib/pdf.min.js:22:69954)

instead of rejecting the promise on the PDFDocumentLoadingTask.

This happens with any "invalid" pdf-file (e.g. a text-document instead of a "real" pdf), loading a "valid" pdf however works absolutely correctly (with any version, also minified-es5).
Using a generic-es5 build produces the expected result (an InvalidPDFException is thrown).

The problem itself however seems to originitate in (minified-es5) pdf.worker.js, because using minified pdf.js with _non-minified_ pdf.worker.js behaves as expected?

ps: did not test with es6 (minified)..

Snuffleupagus:
(Forward) duplicate of #12209

PR #12091, opened 2020-07-13
fix: added fallback when no data passed to message handler

keroxp:
Sometimes `event.data` seems to get null.

### DIFF
--- a/src/shared/message_handler.js
+++ b/src/shared/message_handler.js
@@ -85,7 +85,7 @@ class MessageHandler {
 
     this._onComObjOnMessage = event => {
       const data = event.data;
-      if (data.targetName !== this.sourceName) {
+      if (!data || data.targetName !== this.sourceName) {
         return;
       }
       if (data.stream) {


Snuffleupagus (changes requested):
Please provide a test-case that triggers this problem, since this shouldn't actually happen!

Most likely the actual problem is elsewhere in the code-base, and as-is the patch seems like an attempt to "wallpaper over" an actual bug (rather than fixing it correctly).

keroxp:
@Snuffleupagus Year I'd like believe this never happen. But actually this occurred in our project few times (not many). I couldn't reproduce it and determine what is the case. It might begin to occur from  pdfjs-dist@2.4.456

Here is actual bug report from bugsnag (bug tracker).

[Screenshot from Gyazo](https://gyazo.com/f4cb09b5e6ef3c142d0072adb231020c)

Most errors occurred on Chrome on android. 

> Most likely the actual problem is elsewhere in the code-base

This code seems to not be available from public api and we don't use it directly. Do you have any idea about the case?
Thanks.

Snuffleupagus:
> Here is actual bug report from bugsnag (bug tracker).

It's *really* difficult to tell from just a screen-shot, but it looks like maybe you're using a `minified` build of the PDF.js library?
If so, I'd suggest using the "regular" builds instead and see if the issue persists. Generally though, without a test-case I'm sorry to say that I don't think it's a good idea to make the changes suggests here.

(There's been other issues, see e.g. #11941, which suggests that there may be some issues with `minified` builds.)

keroxp:
@Snuffleupagus Yes, we use minified pdf worker, and I don't think the case is minifying. I think the case is that message event without data is sent from worker by some reason. It is obvious that error occurred at comparison between `data.targetName` and `this.sourceName` (Source mapping is working correctly).

Snuffleupagus:
> [...] and I don't think the case is minifying.

Rather than *assuming* that it's unrelated, as mentioned in https://github.com/mozilla/pdf.js/pull/12091#issuecomment-657437119 please actually test with a "regular" build instead (i.e. using `pdf.js`/`pdf.worker.js` files).

> I think the case is that message event without data is sent from worker by some reason.

Again, that shouldn't actually be happening (and it's quite frankly difficult to see how/where that'd occur, without a test-case).

timvandermeij:
I agree with the analysis above that this should never happen and if it does, it should be fixed at the source instead, so this is not a fix we can consider without clear steps to reproduce the issue and a reason for why it can only be addressed here and not at the source.

Issue #11637, opened 2020-02-27
Adding Standard 14 Font Support - For Acroform use


bpetty-formfast:
The Issue:
Today Acroform text input elements are not being styled correctly in regard to font, font size, etc.
There are two issues there, one involving a half baked font reference issue where fontRefName is never being set.  Even when implemented, there may never be a font reference.  The second issue is that pdf.js does not have "Standard 14" font support that describes 14 fonts that acrobat reader instinctively knows about.

Questions:
1. Would the project be interested if I submit a pull request adding support for these fonts?  This would make acroform input look better until full font reference support is eventually added.
2. If so, I have a basic implementation that works but it uses my own function for parsing DA for text annotations.  I think the right way would be to use the EvaluatorPreprocessor, but being new to the codebase I am not sure exactly how to go about leveraging it, or if you would like any of this information exposed through your public api when someone uses pdfjs-dist to find all annotations, or if this should be used in a more general place.  So basically I am asking if someone would provide some guidance on what you would like to see, if you would even find this useful in the first place.

Thanks!
![image](https://user-images.githubusercontent.com/18601381/75489862-ace72b00-5978-11ea-803f-cc978b244b97.png)

timvandermeij:
There is some DA parsing in https://github.com/mozilla/pdf.js/pull/10122/files, but it was never finished.

I think we don't have the 14 standard fonts because of licensing issues, but perhaps @Snuffleupagus can chime in on that part?

bpetty-formfast:
@timvandermeij, thanks for the pull reference.

My thought was to have a small array of these fonts with their properties.  Something like this:
` HeBO: { name: 'Helvitica-BoldOblique', root: 'Helvitica', bold: true, italic: true },`

We could parse the DA to get the font-code and size, use the code to look up the actual font and font characteristics, and then do the equivalent of "_setTextStyle" in annotation_layer.js.

Because this is strictly working in the display within the browser, the fonts do not necessarily need to be shipped (I assume that was the licensing issue).  All we would be doing is simply providing browser hints to what we are looking for.  I can see this being useful until the fontRefName issue for acroforms is resolved, and for PDFs that do not actually embed these fonts at all.

brendandahl:
> I think we don't have the 14 standard fonts because of licensing issues, but perhaps @Snuffleupagus can chime in on that part?

I think we could use the fonts in pdfium since the project is also apache 2.0.

sayo96:
[...]

PR #12093, opened 2020-07-13
[wip] Freetext annotation DA parsing

escapewindow:
Rescuing changes from #10122 .

I got here from https://github.com/mozilla/pdf.js/issues/11637#issuecomment-592206053 , which I think I got to from https://github.com/mozilla/pdf.js/issues/7613 .


escapewindow:
Reftests are currently busted. Every test failure seems to be one-digit-color off (e.g., `(198, 198, 198)` instead of `(197, 197, 197)`; I'm not sure how big a deal that is?) except for the two freetext annotation tests, which makes sense.

Original `issue9552.pdf`:

image

Patched, broken (or at least different) in both Fx and Chromium:

image

Original `annotation-freetext.pdf`:

image

Patched, broken in Chromium:

image

The latter definitely looks worse. I'm going to try to figure out what broke.

src/core/annotation.js
```diff
@@ -1110,6 +1109,61 @@ class TextAnnotation extends MarkupAnnotation {
   }
 }
 
+class FreeTextAnnotation extends Annotation {
```
timvandermeij, line 1112: Could you move this back to the original place in the file? That will make the diff smaller and easier to review.

src/core/annotation.js
```diff
@@ -1110,6 +1109,61 @@ class TextAnnotation extends MarkupAnnotation {
...
+
+    return evaluator
+      .getOperatorList({
+        stream: appearanceStream,
+        task,
+        resources: resourcesFonts,
+        operatorList: opList,
+      })
+      .then(() => {
+        const a = opList.argsArray;
```
timvandermeij, line 1146: Let's rename this to `args` for a more descriptive name (we always try to avoid single-character names for readability).

src/core/annotation.js
```diff
@@ -1110,6 +1109,61 @@ class TextAnnotation extends MarkupAnnotation {
...
+        let i;
+        for (i = 0; i < opList.fnArray.length; i++) {
```
timvandermeij, line 1148: We normally do this so the `let i` can be inlined and the length is cached for the loop: `for (let i = 0, ii = opList.fnArray.length; i < ii; i++) {`

src/core/annotation.js
```diff
@@ -1110,6 +1109,61 @@ class TextAnnotation extends MarkupAnnotation {
...
+          const fn = opList.fnArray[i];
+          switch (fn | 0) {
+            case OPS.setFont:
+              data.fontRefName = a[i][0];
+              data.fontSize = a[i][1];
+              break;
+            case OPS.setFillGray:
+            case OPS.setFillRGBColor:
+              data.textColor = a[i];
```
timvandermeij, line 1157: Doesn't this also need a call to `createRgbColor` or it is already in the right format?

src/display/annotation_layer.js
```diff
@@ -407,6 +439,93 @@ class TextAnnotationElement extends AnnotationElement {
   }
 }
 
+class FreeTextAnnotationElement extends AnnotationElement {
```
timvandermeij, line 442: Same here for putting it back in its original place to reduce the diff size.

src/display/annotation_layer.js
```diff
@@ -407,6 +439,93 @@ class TextAnnotationElement extends AnnotationElement {
...
+   * @public
+   * @memberof FreeTextAnnotationElement
+   * @returns {HTMLSectionElement}
+   */
+  render() {
+    this.container.className = "freeTextAnnotation";
+    const data = this.data;
+
+    if (!this.data.hasPopup) {
+      this._createPopup(this.container, null, this.data);
```
timvandermeij, line 464: `this.data` can be `data` with the declaration above.

src/display/annotation_layer.js
```diff
@@ -407,6 +439,93 @@ class TextAnnotationElement extends AnnotationElement {
...
+    }
+
+    const div = document.createElement("div");
+    const divWidth = data.rect[2] - data.rect[0];
+    const divHeight = data.rect[3] - data.rect[1];
+    let style = "width: " + divWidth + "px;";
```
timvandermeij, line 470: Let's use template strings here and for the other style strings in this method:

```
let style = `width: ${divWidth}px;`
```

src/display/annotation_layer.js
```diff
@@ -407,6 +439,93 @@ class TextAnnotationElement extends AnnotationElement {
...
+    style += "height: " + divHeight + "px;";
+
+    if (data.hasAppearance) {
+      // The annotation is already rendering by stream appareance,
+      // we should only create popup if necessary
+      div.setAttribute("style", style);
+      this.container.append(div);
+
+      if (!this.data.hasPopup) {
+        this._createPopup(this.container, div, data);
```
timvandermeij, line 480: We should only create the popup once and it's already happening above. I think the question is also here: why does the separate `div` need to be created? Can't `this.container` just be used considering it's also already of the correct width/height so you don't have to calculate that yourself anymore?

src/display/annotation_layer.js
```diff
@@ -407,6 +439,93 @@ class TextAnnotationElement extends AnnotationElement {
...
+      }
+
+      return this.container;
+    }
+
+    // Color define the background color of the annotation
```
timvandermeij, line 486: The code already makes this clear enough so we don't need this comment.

src/display/annotation_layer.js
```diff
@@ -407,6 +439,93 @@ class TextAnnotationElement extends AnnotationElement {
...
+        data.color[2] | 0
+      );
+      style += "background-color: " + backgroundColor + ";";
+    }
+
+    if (data.opacity) {
+      style += "opacity: " + data.opacity + ";";
+    }
+
+    // Text style
```
timvandermeij, line 500: Same here, the code is clear enough so this comment can be removed.

src/display/annotation_layer.js
```diff
@@ -407,6 +439,93 @@ class TextAnnotationElement extends AnnotationElement {
...
+      this.page.commonObjs.has(this.data.fontRefName)
+    ) {
+      font = this.page.commonObjs.get(this.data.fontRefName);
+    }
+    setTextStyle(this.data, div, font);
+
+    div.innerText = data.contents;
+    this.container.append(div);
+
+    this._createPopup(this.container, div, data);
```
timvandermeij, line 523: Same here for the popup, it should be conditional like above and only happen once.

test/test_manifest.json
```diff
@@ -4189,6 +4189,12 @@
        "rounds": 1,
        "type": "eq"
     },
+    {  "id": "issue10122",
+       "file": "pdfs/issue10122.pdf",
+       "md5": "80a0a0e34433c5b3fd4a57778bb2b4b4",
+       "rounds": 1,
+       "type": "eq"
```
timvandermeij, line 4196: You also need https://github.com/mozilla/pdf.js/blob/3404e5ee1919fda1ce7ce5908db19f7e82f63177/test/test_manifest.json#L4450 here, otherwise the annotation is not rendered in the reference tests.

src/core/annotation.js
```diff
@@ -1110,6 +1109,61 @@ class TextAnnotation extends MarkupAnnotation {
...
+  getOperatorList(evaluator, task, renderForms) {
+    if (renderForms || this.appearance) {
+      return super.getOperatorList(evaluator, task, renderForms);
+    }
+
+    const data = this.data;
+
+    const opList = new OperatorList();
+    const appearanceStream = new Stream(stringToBytes(data.defaultAppearance));
+    const resourcesFonts = this.pdfManager.pdfDocument.catalog.fontCache.dict;
```
Snuffleupagus, line 1136: Unfortunately https://github.com/mozilla/pdf.js/pull/10122#discussion_r224351991 still applies here, there's simply no way that this works correctly.

src/display/annotation_layer.js
timvandermeij: I also think the fact that multiple popups are rendered is what's breaking the reference tests.

escapewindow: Once I commented out the first popup, the reftests were no longer broken. Thanks!
I do still see some minor differences between the reference and test reftests, but they seem very minor.

src/core/annotation.js
escapewindow: I tried resolving this in https://github.com/mozilla/pdf.js/pull/12093/commits/e9ed30cbe958a9aa430636443a628f769c96f79c . Does that look sane?

escapewindow:
Hm, but I lost the freetext annotations in issue10122.pdf in b20af4a .

src/display/annotation_layer.js
Snuffleupagus: > I do still see some minor differences between the reference and test reftests, but they seem very minor.

Even if changes in the reference tests are minor, please make sure that you can explain *why* they happen.
(Unexpected changes, without further investigation, *could* maybe point to something being wrong.)

src/core/annotation.js
```diff
@@ -1188,6 +1187,60 @@ class FreeTextAnnotation extends MarkupAnnotation {
...
+
+  getOperatorList(evaluator, task, renderForms) {
+    if (renderForms || this.appearance) {
+      return super.getOperatorList(evaluator, task, renderForms);
+    }
+
+    const data = this.data;
+
+    const appearanceDict = this.appearance.dict;
+    const appearanceStream = new Stream(stringToBytes(data.defaultAppearance));
```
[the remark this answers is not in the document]
escapewindow: It looks like we use the original construct one other place in the file. I updated both places.

src/core/annotation.js
```diff
@@ -1188,6 +1187,60 @@ class FreeTextAnnotation extends MarkupAnnotation {
...
+      return evaluator
+        .getOperatorList({
+          stream: appearanceStream,
+          task,
+          resources,
+          operatorList: opList,
+        })
+        .then(() => {
+          const args = opList.argsArray;
+          for (let i = 0, ii = opList.fnArray.length; i < ii; i++) {
```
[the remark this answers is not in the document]
escapewindow: `OperatorList.length` appears to [return `this.argsArray.length`](https://github.com/mozilla/pdf.js/blob/master/src/core/operator_list.js#L629-L632) instead of `this.fnArray.length`. Are they always the same length?

escapewindow:
> Even if changes in the reference tests are minor, please make sure that you can explain why they happen.
(Unexpected changes, without further investigation, could maybe point to something being wrong.)

This will likely take some time. For instance, in chrome-tracemonkey-text-page7, there's a single pixel that's now `b7ffb7` instead of `b8ffb8`... similar small differences changed 24 different reftests. Has this type of single-pixel difference occurred before? Is there something I should look for, or do you have a recommended way of debugging?

timvandermeij:
Oh, there are currently some regression test failures in Chrome due to intermittent rendering differences in the Chrome engine. If it's only in Chrome and not in Firefox, it's most likely not a concern (at least, not caused by your code).

escapewindow:
Ah, yes, the differences are all Chrome. Thanks! That saved me a lot of time.
I still need to debug why removing the `div` removed the annotations in https://github.com/mozilla/pdf.js/commit/b20af4a9067a08b74c5a9ab9dd59a38047b5429c , and address [this comment](https://github.com/mozilla/pdf.js/pull/12093#discussion_r457897637). Getting closer :)

PR #12095, opened 2020-07-14
Add support for optional marked content.

### LINKED ISSUE #269: Feature request for the future: PDF layers
That is all.

The code looks really promising!

brendandahl:
Add a new method to the API to get the optional content configuration. Add
a new render task param that accepts the above configuration.
For now, the optional content is not controllable by the user in
the viewer, but renders with the default configuration in the PDF.

All of the test files added exhibit different uses of optional content.

Fixes #269.


src/core/evaluator.js
```diff
@@ -1695,6 +1755,21 @@ class PartialEvaluator {
             // e.g. as done in https://github.com/mozilla/pdf.js/pull/6266,
             // but doing so is meaningless without knowing the semantics.
             continue;
+          case OPS.beginMarkedContentProps:
+            if (!isName(args[0])) {
+              warn(`Expected name for beginMarkedContentProps arg0=${args[0]}`);
+              continue;
+            }
+            fn = OPS.beginMarkedContentProps;
```
Snuffleupagus, line 1763: This line looks superfluous, since `fn === OPS.beginMarkedContentProps` must hold in order for this block to have been entered!?

src/core/obj.js
```diff
@@ -254,6 +254,62 @@ class Catalog {
...
+        intent: group.get("Intent"), // TODO validate this.
+      });
+    }
+    const defaultConfig = properties.get("D");
+    if (!defaultConfig) {
+      return null;
+    }
+    const config = this._readOptionalContentConfig(defaultConfig, groupRefs);
+    config.groups = groups;
+    return config;
```
Snuffleupagus, line 287: Similar to all other methods here, see e.g. `get permissions()` above, please wrap this in `try...catch` to prevent breaking errors if data fetching fails (since Dict.get(...) calls *could* throw).

src/core/obj.js
```diff
@@ -254,6 +254,62 @@ class Catalog {
...
+  }
+
+  _readOptionalContentConfig(config, contentGroupRefs) {
+    // console.log(contentGroupRefs);
```
Snuffleupagus, line 291: Left-over debugging statement?

src/core/obj.js
Snuffleupagus: Will `get optionalContentConfig()` always return the same data for a document with Optional Content groups?
If so, please `shadow` the return values to avoid having to repeatedly re-parsing this for *every* single page. (Or at least, please `shadow` the `return null;` cases.)

src/core/evaluator.js
```diff
@@ -1695,6 +1755,21 @@ class PartialEvaluator {
...
+            if (args[0].name === "OC") {
+              args = ["OC", self.parseMarkedContentProps(args[1], resources)];
```
Snuffleupagus, line 1765: With `parseMarkedContentProps` being asynchronous, as suggested above, this would need to become something like this (which thus handles any errors gracefully):
```javascript
next(
  self.parseMarkedContentProps(args[1], resources)
    .then(data => {
       operatorList.addOp(OPS.beginMarkedContentProps, ["OC", data]);
    })
    .catch(reason => {
      if (reason instanceof AbortException) {
        return;
      }
      if (self.options.ignoreErrors) {
        self.handler.send("UnsupportedFeature", {
          featureId: /* Some new UNSUPPORTED_FEATURES id */,
        });
        warn(`getOperatorList - ignoring beginMarkedContentProps: "${reason}".`);
        return;
      }
      throw reason;
    })
);
return;
```

src/core/evaluator.js
```diff
@@ -422,8 +422,16 @@ class PartialEvaluator {
         smask,
         isolated: false,
         knockout: false,
+        optionalContent: null,
       };
 
+      if (dict.has("OC")) {
+        groupOptions.optionalContent = this.parseMarkedContentProps(
```
Snuffleupagus, line 429: Nit: With parseMarkedContentProps being asynchronous, as suggested below, this would need to be `groupOptions.optionalContent = await this.parseMarkedContentProps(` instead.

src/display/canvas.js
```diff
@@ -1698,10 +1715,10 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
         // Only attempt to draw the glyph if it is actually in the embedded font
         // file or if there isn't a font file so the fallback font is shown.
         if (glyph.isInFont || font.missingFile) {
-          if (simpleFillText && !accent) {
+          if (simpleFillText && !accent && this.contentVisible) {
             // common case
             ctx.fillText(character, scaledX, scaledY);
-          } else {
+          } else if (this.contentVisible) {
```
Snuffleupagus, line 1721: It would in my opinion be, a little bit, clearer if you simply enclosed this entire code-block https://github.com/mozilla/pdf.js/blob/6c39aff3743564f3418ba17344e14819f923a427/src/display/canvas.js#L1698-L1717
in `if (this.contentVisible) { ... }`  rather than "inlining" these checks.

Snuffleupagus: Also, unless I'm mistaken there's no similar checks in the `showType3Text` below. Don't you need to account for `this.contentVisible` there as well?

src/display/api.js
```diff
@@ -2650,6 +2686,19 @@ const InternalRenderTask = (function InternalRenderTaskClosure() {
...
+            new Error(
+              "Cannot use the same canvas during multiple render() operations. " +
+                "Use different canvas or ensure previous operations were " +
+                "cancelled or completed."
+            )
+          );
+        } else {
+          canvasInRendering.add(this._canvas);
+        }
+      }
```
Snuffleupagus, line 2701: Which test-case required moving this code, and is there another way to deal with that?
(Since the commit message says: `- Reject the InternalRenderTask earlier now that InitializeGraphics is waits for another promise.`)

I've not actually tested it, but based on reading the code I'm not entirely convinced that moving this code will actually do the right thing in general.
Currently `initializeGraphics` will throw https://github.com/mozilla/pdf.js/blob/6c39aff3743564f3418ba17344e14819f923a427/src/display/api.js#L2668-L2672 which thus means that we're reaching this catch https://github.com/mozilla/pdf.js/blob/6c39aff3743564f3418ba17344e14819f923a427/src/display/api.js#L1118-L1121 which thus calls https://github.com/mozilla/pdf.js/blob/6c39aff3743564f3418ba17344e14819f923a427/src/display/api.js#L1054-L1081 as expected and all of the required clean-up will thus run.

With your changes here, it doesn't seem like you're actually *guaranteed* that `complete` is ever called correctly!?

Snuffleupagus (changes requested):

src/display/api.js
brendandahl: It was the test that calls render twice. It doesn't seem to be failing anymore. This code is quite the mix of callbacks and old style promises.

src/display/api.js
```diff
@@ -2447,6 +2473,16 @@ class WorkerTransport {
     return this.messageHandler.sendWithPromise("GetOutline", null);
   }
 
+  getOptionalContentConfig() {
+    return this.messageHandler
+      .sendWithPromise("GetOptionalContentConfig", null)
+      .then(results => {
+        const config = new OptionalContentConfig();
```
Snuffleupagus, line 2480: Just wondering: Any particular reason to not simply pass the `results` when creating the `OptionalContentConfig` instance, and thus essentially inline the `initialize` method in the constructor?

src/display/api.js
```diff
@@ -2447,6 +2473,16 @@ class WorkerTransport {
     return this.messageHandler.sendWithPromise("GetOutline", null);
   }
 
+  getOptionalContentConfig() {
+    return this.messageHandler
+      .sendWithPromise("GetOptionalContentConfig", null)
+      .then(results => {
```
Snuffleupagus, line 2479: Would it make (any) sense to cache this on the `WorkerTransport`-instance, to avoid having to wait for the worker-thread to respond for rendering to be able to start?

E.g. adding `this._optionalContentConfigPromise = null` in the `WorkerTransport` constructor and changing the code here to something like:
```javascript
if (!this._optionalContentConfigPromise) {
  this._optionalContentConfigPromise =
    this.messageHandler.sendWithPromise("GetOptionalContentConfig", null);
}
this._optionalContentConfigPromise.then(results => {
  ...
```

src/core/obj.js
```diff
@@ -254,6 +254,69 @@ class Catalog {
...
+      // Ensure all the optional content groups are valid.
+      for (const groupRef of groupsData) {
+        if (!isRef(groupRef)) {
+          continue;
+        }
+        groupRefs.push(groupRef);
+        const group = properties.xref.fetchIfRef(groupRef);
+        groups.push({
+          id: groupRef.toString(),
+          name: group.get("Name"),
```
Snuffleupagus, line 279: Since this is, I believe, a string; don't you need to wrap it in `stringToPDFString()` to avoid issues?

src/core/obj.js
```diff
@@ -254,6 +254,69 @@ class Catalog {
...
+            onParsed.push(value.toString());
+          }
+        }
+      }
+      return onParsed;
+    }
+
+    return {
+      name: config.get("Name"),
+      creator: config.get("Creator"),
```
Snuffleupagus, line 313: Since these are, I believe, strings; don't you need to wrap them in `stringToPDFString()` to avoid issues?

src/core/obj.js
```diff
@@ -254,6 +254,69 @@ class Catalog {
...
+      baseState: config.get("BaseState"),
```
Snuffleupagus, line 314: According to the specification, https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G7.3883884, this should be a `Name`. In the `OptionalContentConfig` class you're accessing them as if they were mere strings, which I don't think will work!?

I believe that you'll need to either validate that it's a name and (essentially) return `config.get("BaseState").name,` here, or possibly return the name as-is a and fix the lookup in `OptionalContentConfig` instead.

src/core/evaluator.js
```diff
@@ -1224,6 +1232,61 @@ class PartialEvaluator {
...
+        let expression = null;
+        if (optionalContent.get("VE")) {
+          // TODO support content epxressions.
+          expression = true;
+        }
+
+        return {
+          type: optionalContentType,
+          ids: groupIds,
+          policy: optionalContent.get("P"),
```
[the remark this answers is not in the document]
Snuffleupagus: According to the specification, https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G7.3883825, this should be a `Name`. In the `OptionalContentConfig` class you're accessing them as if they were mere strings, which I don't think will work!?

I believe that you'll need to either validate that it's a name and (essentially) return `optionalContent.get("P").name,` here, or possibly return the name as-is a and fix the lookup in `OptionalContentConfig` instead.

Snuffleupagus (changes requested):

src/core/evaluator.js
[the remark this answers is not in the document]
brendandahl: I thought there was a test PDF for this, but appears there aren't any with a policy attribute. If I have some time I'll try to make some more tests.

src/core/evaluator.js
```diff
@@ -1695,6 +1755,45 @@ class PartialEvaluator {
             // e.g. as done in https://github.com/mozilla/pdf.js/pull/6266,
             // but doing so is meaningless without knowing the semantics.
             continue;
+          case OPS.beginMarkedContentProps:
+            if (!isName(args[0])) {
+              warn(`Expected name for beginMarkedContentProps arg0=${args[0]}`);
```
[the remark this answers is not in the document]
brendandahl: I started to do do this, but sometimes the marked content isn't an "OC" which then causes us to wait async when we really don't need to.

src/display/api.js
brendandahl: I had plans of doing some unit testing this differently. Not going to do that now, so I'll merge it.

src/core/obj.js
```diff
@@ -254,6 +254,82 @@ class Catalog {
...
+            : null,
+          intent: isString(group.get("Intent"))
+            ? stringToPDFString(group.get("Intent"))
+            : null,
+        });
+      }
+      const defaultConfig = properties.get("D");
+      if (!defaultConfig) {
+        return shadow(this, "optionalContentConfig", null);
+      }
```
Snuffleupagus, line 290: Nit: Would it make sense to move this to occur earlier in the method, since the `groupsData` parsing above seems pointless in the `!defaultConfig` case?

Snuffleupagus (approved):
r=me, with the commits squashed and all tests passing.

src/core/evaluator.js
```diff
@@ -1224,6 +1232,63 @@ class PartialEvaluator {
...
+            groupIds.push(ocg.toString());
+          });
+        } else {
+          // Dictionary, just use the obj id.
+          groupIds.push(optionalContentGroups.objId);
+        }
+
+        let expression = null;
+        if (optionalContent.get("VE")) {
+          // TODO support content epxressions.
```
timvandermeij, line 1270: Nit: `expressions`

src/core/worker.js
```diff
@@ -485,6 +485,13 @@ class WorkerMessageHandler {
       return pdfManager.ensureCatalog("documentOutline");
     });
 
+    handler.on(
+      "GetOptionalContentConfig",
+      function wphSetupGetOptionalContentConfig(data) {
```
timvandermeij, line 490: This handler can be shortened a bit by using an anonymous function like the one below.

src/display/api.js
```diff
@@ -709,6 +710,15 @@ class PDFDocumentProxy {
     return this._transport.getOutline();
   }
 
+  /**
+   * @returns {Promise} A promise that is resolved with an
+   * {OptionalContentConfig} that will have all the optional content groups (if
+   * the document has any). @see {OptionalContentConfig}
```
timvandermeij, line 716: Shouldn't `@see` be on a new line for JSDoc?

src/display/api.js
```diff
@@ -882,6 +892,11 @@ class PDFDocumentProxy {
  *                    CSS <color> value, a CanvasGradient object (a linear or
  *                    radial gradient) or a CanvasPattern object (a repetitive
  *                    image). The default value is 'rgb(255,255,255)'.
+ * @property {Promise} [optionalContentConfigPromise] - A promise that should
+ *                     resolve with an {OptionalContentConfig} created from
+ *                     PDFDocumentProxy.getOptionalContentConfig. If null, the
+ *                     config will automatically fetched with the default
```
timvandermeij, line 898: Nit: `automatically be fetched`

src/display/optional_content_config.js
```diff
@@ -0,0 +1,107 @@
+/* Copyright 2012 Mozilla Foundation
```
timvandermeij, line 1: Nit: 2020

timvandermeij (commented):
Minor comments, also looks good to me after this.

Snuffleupagus (changes requested):
According to the test logs, there's *two* actual errors thrown during rendering:
```
TEST-UNEXPECTED-FAIL | test failed issue7769 | in firefox | page1 round 1 | render : Error: Optional content group must be found.
TEST-UNEXPECTED-FAIL | test failed issue7769 | in chrome | page1 round 1 | render : Error: Optional content group must be found.
```
Unfortunately, it appears that this particular error occurs in such a way that the failures aren't displayed in the "Reftest analyzer" (and I thus missed it previously).

src/display/optional_content_config.js
```diff
@@ -0,0 +1,107 @@
...
+    for (const on of data.on) {
+      this.groups.get(on).visible = true;
+    }
+
+    for (const off of data.off) {
+      this.groups.get(off).visible = false;
+    }
+  }
+
+  isVisible(group) {
```
Snuffleupagus, line 58: For PDF documents where the `OptionalContentConfig` is initialized with `data === null`, such as `issue7769` in the test-suite, where `this.groups` is thus *always empty* we still need to ensure that rendering works correctly. Hence I'd suggest, in addition to the inline checks mentioned elsewhere, that you also add the following checks here:
```suggestion
  isVisible(group) {
    if (!this.groups.size) {
      warn("No optional content groups found.");
      return true;
    }
```

Snuffleupagus (changes requested):

src/display/optional_content_config.js
```diff
@@ -0,0 +1,107 @@
...
+        for (const id of group.ids) {
+          if (this.groups.get(id).visible) {
+            return false;
+          }
+        }
+        return true;
+      }
+      throw new Error(`Unknown optional content policy ${group.policy}.`);
+    }
+    throw new Error(`Unknown group type ${group.type}.`);
```
Snuffleupagus, line 103: Given the issues seen in the `issue7769` document, I'm now also wondering if it'd make more sense to warn rather than throw in these two cases!?
That way rendering would still "work", rather than outright breaking.

src/core/evaluator.js
```diff
@@ -422,8 +422,16 @@ class PartialEvaluator {
         smask,
         isolated: false,
         knockout: false,
+        optionalContent: null,
       };
 
+      if (dict.has("OC")) {
```
janpe2, line 428: All form XObjects can have an `OC` entry, not just transparency groups that have a `Group` entry. Then you may need to pass the parsed `OC` value also to `paintFormXObjectBegin/End` and handle it in canvas.js.

janpe2:
You should push a `{ visible: true }` object also in `beginMarkedContent` ("BMC").
https://github.com/mozilla/pdf.js/blob/d69fb446bff53e4d8eddc520635b004f4ac986d3/src/display/canvas.js#L2444
"EMC" is the end of both "BMC" and "BDC" sections. "BMC" is not related to optional content but some PDFs may use it and the marked content stack might go wrong.

src/display/optional_content_config.js
janpe2: Section 8.11.4.2 in the standard says this about the `OCProperties` entry which is missing in issue7769:
>if it is missing, a conforming reader shall 
ignore any optional content structures in the document.
I think this means that all graphics should be made visible.

Snuffleupagus:
@brendandahl Given that this PR seems *really* close to being done, since there's just a few more comments to fix (such that all tests pass), do you perhaps have time to fix the remaining things such that this can land?

I've been playing around with this patch a fair bit locally[1], and it seems to work really well in practice :-)
Also, the overall implementation is a lot simpler than I thought it would be when reading the specification. 

---

[1] I've got, more-or-less, finished patches for the viewer integration.

src/core/evaluator.js
brendandahl: Good catch, I'll just insert a `beginMarkedContentProps` before the form object so I don't have to handle this for both.

src/display/api.js
```diff
@@ -965,6 +976,11 @@ class PDFDocumentProxy {
  *                    image). The default value is 'rgb(255,255,255)'.
  * @property {Object} [annotationStorage] - Storage for annotation data in
  *                    forms.
+ * @property {Promise} [optionalContentConfigPromise] - A promise that should
```
timvandermeij, line 979: Once #12156 lands, this should be changed to mention the promise type and to have each line start with two spaces as per the standard.

timvandermeij (commented):
I'm leaving the review for @Snuffleupagus, but I'm just checking the comments/types here. For context, we have been working a lot on that the last few days in #12102 and follow-up pull requests. The main priority was to get TypeScript definitions generated automatically from the source code and to get the comments in a consistent format, so this requires a small number of changes here to make sure that once #12156 lands everything is still consistent and the generated types are correct. Thanks!

timvandermeij:
By the way, the unit test failures (and most likely the reference test failures as well) are known intermittents and are tracked in #12123 (the Windows one is a new one that I now mentioned there).

src/display/canvas.js
```diff
@@ -2446,15 +2496,28 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
     },
     beginMarkedContent: function CanvasGraphics_beginMarkedContent(tag) {
       // TODO Marked content.
```
Snuffleupagus, line 2498: Nit: Remove this comment?

Snuffleupagus (approved):
Looks good to me, thank you!

PR #12099, opened 2020-07-15
Use a `RefSet`, rather than a plain Object, for tracking already processed nodes in `PartialEvaluator.hasBlendModes`

Snuffleupagus:


src/core/evaluator.js
```diff
@@ -322,20 +323,20 @@ class PartialEvaluator {
...
-          processed[xObject.dict.objId] = true;
+          processed.put(xObject.dict.objId);
         }
         var xResources = xObject.dict.get("Resources");
         if (!(xResources instanceof Dict)) {
           continue;
         }
         // Checking objId to detect an infinite loop.
-        if (xResources.objId && processed[xResources.objId]) {
+        if (xResources.objId && processed.has(xResources)) {
```
timvandermeij, line 333: I think this should be `processed.has(xResources.objId)`.

Snuffleupagus: Of course it should, I really don't know what happened there; thank you!

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #12102, opened 2020-07-18
Add types annotations

ineiti:
Follow-up of #10575:
- merged oBusk's and tamuratak's PRs
- fixed merge conflicts (hopefully correctly)
- added an Array type for a minimum example

I build it using:
```
npm ci
gulp generic
npx tsc -p .
```

And then trying to run the example from https://github.com/ineiti/pdf_example - I hope that's not too complex example to start from...

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -63,6 +63,7 @@ var GH_PAGES_DIR = BUILD_DIR + "gh-pages/";
 var SRC_DIR = "src/";
 var LIB_DIR = BUILD_DIR + "lib/";
 var DIST_DIR = BUILD_DIR + "dist/";
+var TYPES_BUILD_DIR = BUILD_DIR + "types/";
 var COMMON_WEB_FILES = ["web/images/*.{png,svg,gif,cur}", "web/debugger.js"];
 var MOZCENTRAL_DIFF_FILE = "mozcentral.diff";
 
@@ -1141,6 +1142,21 @@ gulp.task("jsdoc", function (done) {
   });
 });
 
+gulp.task("types", function (done) {
+  console.log("### Generating typescript definitions using tsc");
+  var args = [
+    "target ES2020",
+    "allowJS",
+    "declaration",
+    `outDir ${TYPES_BUILD_DIR}`,
+    "strict",
+    "esModuleInterop",
+    "forceConsistentCasingInFileNames",
+    "emitDeclarationOnly",
+  ].join(" --");
+  exec(`node_modules/.bin/tsc --${args} src/pdf.js`, done);
+});
+
 function buildLib(defines, dir) {
   // When we create a bundle, webpack is run on the source and it will replace
   // require with __webpack_require__. When we want to use the real require,
@@ -1334,6 +1350,42 @@ gulp.task(
   gulp.series("testing-pre", "generic", "components", function (done) {
     makeRef(done, true);
   })
+);
+
+gulp.task(
+  "typestest",
+  gulp.series(
+    "testing-pre",
+    "generic",
+    "types",
+
+    function () {
+      var packageJsonSrc = packageBowerJson()[0];
+      var TYPESTEST_DIR = BUILD_DIR + "typestest/";
+
+      return merge([
+        packageJsonSrc.pipe(gulp.dest(TYPESTEST_DIR)),
+        gulp
+          .src([
+            GENERIC_DIR + "build/pdf.js",
+            GENERIC_DIR + "build/pdf.worker.js",
+            SRC_DIR + "pdf.worker.entry.js",
+          ])
+          .pipe(gulp.dest(TYPESTEST_DIR + "build/")),
+        gulp
+          .src(TYPES_BUILD_DIR + "**/**")
+          .pipe(gulp.dest(TYPESTEST_DIR + "build/")),
+      ]);
+    },
+    function (done) {
+      exec(`node_modules/.bin/tsc -p test/types`, function (err, stdout) {
+        if (err !== null) {
+          console.log("couldn't compile typescript test: " + stdout);
+        }
+        done(err);
+      });
+    }
+  )
 );
 
 gulp.task("baseline", function (done) {
@@ -1576,6 +1628,52 @@ gulp.task(
   )
 );
 
+function packageBowerJson() {
+  var VERSION = getVersionJSON().version;
+  var DIST_NAME = "pdfjs-dist";
+  var DIST_DESCRIPTION = "Generic build of Mozilla's PDF.js library.";
+  var DIST_KEYWORDS = ["Mozilla", "pdf", "pdf.js"];
+  var DIST_HOMEPAGE = "http://mozilla.github.io/pdf.js/";
+  var DIST_BUGS_URL = "https://github.com/mozilla/pdf.js/issues";
+  var DIST_LICENSE = "Apache-2.0";
+  var npmManifest = {
+    name: DIST_NAME,
+    version: VERSION,
+    main: "build/pdf.js",
+    types: "build/pdf.d.ts",
+    description: DIST_DESCRIPTION,
+    keywords: DIST_KEYWORDS,
+    homepage: DIST_HOMEPAGE,
+    bugs: DIST_BUGS_URL,
+    license: DIST_LICENSE,
+    browser: {
+      canvas: false,
+      fs: false,
+      http: false,
+      https: false,
+      url: false,
+      zlib: false,
+    },
+    format: "amd", // to not allow system.js to choose 'cjs'
+    repository: {
+      type: "git",
+      url: DIST_REPO_URL,
+    },
+  };
+  var bowerManifest = {
+    name: DIST_NAME,
+    version: VERSION,
+    main: ["build/pdf.js", "build/pdf.worker.js"],
+    ignore: [],
+    keywords: DIST_KEYWORDS,
+  };
+
+  return [
+    createStringSource("package.json", JSON.stringify(npmManifest, null, 2)),
+    createStringSource("bower.json", JSON.stringify(bowerManifest, null, 2)),
+  ];
+}
+
 gulp.task(
   "dist-pre",
   gulp.series(
@@ -1586,9 +1684,8 @@ gulp.task(
     "image_decoders",
     "lib",
     "minified",
+    "types",
     function () {
-      var VERSION = getVersionJSON().version;
-
       console.log();
       console.log("### Cloning baseline distribution");
 
@@ -1601,50 +1698,7 @@ gulp.task(
       rimraf.sync(path.join(DIST_DIR, "*"));
 
       // Rebuilding manifests
-      var DIST_NAME = "pdfjs-dist";
-      var DIST_DESCRIPTION = "Generic build of Mozilla's PDF.js library.";
-      var DIST_KEYWORDS = ["Mozilla", "pdf", "pdf.js"];
-      var DIST_HOMEPAGE = "http://mozilla.github.io/pdf.js/";
-      var DIST_BUGS_URL = "https://github.com/mozilla/pdf.js/issues";
-      var DIST_LICENSE = "Apache-2.0";
-      var npmManifest = {
-        name: DIST_NAME,
-        version: VERSION,
-        main: "build/pdf.js",
-        description: DIST_DESCRIPTION,
-        keywords: DIST_KEYWORDS,
-        homepage: DIST_HOMEPAGE,
-        bugs: DIST_BUGS_URL,
-        license: DIST_LICENSE,
-        browser: {
-          canvas: false,
-          fs: false,
-          http: false,
-          https: false,
-          url: false,
-          zlib: false,
-        },
-        format: "amd", // to not allow system.js to choose 'cjs'
-        repository: {
-          type: "git",
-          url: DIST_REPO_URL,
-        },
-      };
-      var packageJsonSrc = createStringSource(
-        "package.json",
-        JSON.stringify(npmManifest, null, 2)
-      );
-      var bowerManifest = {
-        name: DIST_NAME,
-        version: VERSION,
-        main: ["build/pdf.js", "build/pdf.worker.js"],
-        ignore: [],
-        keywords: DIST_KEYWORDS,
-      };
-      var bowerJsonSrc = createStringSource(
-        "bower.json",
-        JSON.stringify(bowerManifest, null, 2)
-      );
+      var [packageJsonSrc, bowerJsonSrc] = packageBowerJson();
 
       return merge([
         packageJsonSrc.pipe(gulp.dest(DIST_DIR)),
@@ -1698,6 +1752,9 @@ gulp.task(
         gulp
           .src(LIB_DIR + "**/*", { base: LIB_DIR })
           .pipe(gulp.dest(DIST_DIR + "lib/")),
+        gulp
+          .src(TYPES_BUILD_DIR + "**/**")
+          .pipe(gulp.dest(DIST_DIR + "build/")),
       ]);
     }
   )
@@ -1847,7 +1904,7 @@ gulp.task("externaltest", function (done) {
 gulp.task(
   "npm-test",
   gulp.series(
-    gulp.parallel("lint", "externaltest", "unittestcli"),
+    gulp.parallel("lint", "externaltest", "unittestcli", "typestest"),
     "lint-chromium"
   )
 );
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -97,13 +97,19 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
   createPDFNetworkStream = pdfNetworkStreamFactory;
 }
 
+/* eslint-disable max-len */
+/**
+ * @typedef {Int8Array | Uint8Array | Int16Array | Uint16Array | Int32Array | Uint32Array | Uint8ClampedArray | Float32Array | Float64Array} TypedArray
+ */
+/* eslint-enable max-len */
+
 /**
  * Document initialization / loading parameters object.
  *
  * @typedef {Object} DocumentInitParameters
  * @property {string}     [url] - The URL of the PDF.
- * @property {TypedArray|Array|string} [data] - Binary PDF data. Use typed
- *    arrays (Uint8Array) to improve the memory usage. If PDF data is
+ * @property {TypedArray|Array<number>|string} [data] - Binary PDF data. Use
+ *    typed arrays (Uint8Array) to improve the memory usage. If PDF data is
  *    BASE64-encoded, use atob() to convert it to a binary string first.
  * @property {Object}     [httpHeaders] - Basic authentication headers.
  * @property {boolean}    [withCredentials] - Indicates whether or not
@@ -171,10 +177,12 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
 
 /**
  * @typedef {Object} PDFDocumentStats
- * @property {Object} streamTypes - Used stream types in the document (an item
- *   is set to true if specific stream ID was used in the document).
- * @property {Object} fontTypes - Used font types in the document (an item
- *   is set to true if specific font ID was used in the document).
+ * @property {Object<string, boolean>} streamTypes - Used stream types in the
+ *   document (an item is set to true if specific stream ID was used in the
+ *   document).
+ * @property {Object<string, boolean>} fontTypes - Used font types in the
+ *   document (an item is set to true if specific font ID was used in the
+ *   document).
  */
 
 /**
@@ -414,6 +422,51 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
     });
 }
 
+/**
+ * The loading task controls the operations required to load a PDF document
+ * (such as network requests) and provides a way to listen for completion,
+ * after which individual pages can be rendered.
+ *
+ * @typedef {Object} PDFDocumentLoadingTask
+ * @property {string} docId
+ * Unique document loading task id -- used in MessageHandlers.
+ * @property {boolean} destroyed
+ * Shows if loading task is destroyed.
+ * @property {cbOnPassword} [onPassword]
+ * Callback to request a password if wrong or no password was provided.
+ * The callback receives two parameters: function that needs to be called
+ * with new password and reason (see {PasswordResponses}).
+ * @property {cbOnProgress} onProgress
+ * Callback to be able to monitor the loading progress of the PDF file
+ * (necessary to implement e.g. a loading bar). The callback receives
+ * an {Object} with the properties: {number} loaded and {number} total.
+ * @property {cbOnUnsupportedFeature} onUnsupportedFeature
+ * Callback for when an unsupported feature is used in the PDF document.
+ * The callback receives an {UNSUPPORTED_FEATURES} argument.
+ * @property {Promise<PDFDocumentProxy>} promise
+ * Promise for document loading task completion.
+ * @property {cbDestroy} destroy
+ * Aborts all network requests and destroys worker.
+ * Returns a promise that is resolved after destruction activity is completed.
+ *
+ * @callback cbOnPassword
+ * @param {string} password
+ * @param {number} reason
+ *
+ * @callback cbOnProgress
+ * @param {number} loaded
+ * @param {number} total
+ *
+ * @callback cbOnUnsupportedFeature
+ * @param {string} featureId
+ *
+ * @callback cbDestroy
+ */
+
+/**
+ * @type {any}
+ * @ignore
+ */
 const PDFDocumentLoadingTask = (function PDFDocumentLoadingTaskClosure() {
   let nextDocumentId = 0;
 
@@ -464,7 +517,7 @@ const PDFDocumentLoadingTask = (function PDFDocumentLoadingTaskClosure() {
 
     /**
      * Promise for document loading task completion.
-     * @type {Promise}
+     * @type {Promise<PDFDocumentProxy>}
      */
     get promise() {
       return this._capability.promise;
@@ -472,8 +525,8 @@ const PDFDocumentLoadingTask = (function PDFDocumentLoadingTaskClosure() {
 
     /**
      * Aborts all network requests and destroys worker.
-     * @returns {Promise} A promise that is resolved after destruction activity
-     *                    is completed.
+     * @returns {Promise<void>} A promise that is resolved after destruction
+     *                          activity is completed.
      */
     destroy() {
       this.destroyed = true;
@@ -495,11 +548,13 @@ const PDFDocumentLoadingTask = (function PDFDocumentLoadingTaskClosure() {
 
 /**
  * Abstract class to support range requests file loading.
- * @param {number} length
- * @param {Uint8Array} initialData
- * @param {boolean} progressiveDone
  */
 class PDFDataRangeTransport {
+  /**
+   * @param {number} length
+   * @param {Uint8Array} initialData
+   * @param {boolean} progressiveDone
+   */
   constructor(length, initialData, progressiveDone = false) {
     this.length = length;
     this.initialData = initialData;
@@ -603,8 +658,8 @@ class PDFDocumentProxy {
 
   /**
    * @param {number} pageNumber - The page number to get. The first page is 1.
-   * @returns {Promise} A promise that is resolved with a {@link PDFPageProxy}
-   *   object.
+   * @returns {Promise<PDFPageProxy>} A promise that is resolved with
+   *    a {@link PDFPageProxy} object.
    */
   getPage(pageNumber) {
     return this._transport.getPage(pageNumber);
@@ -613,16 +668,17 @@ class PDFDocumentProxy {
   /**
    * @param {{num: number, gen: number}} ref - The page reference. Must have
    *   the `num` and `gen` properties.
-   * @returns {Promise} A promise that is resolved with the page index (starting
-   *   from zero) that is associated with the reference.
+   * @returns {Promise<{num: number, gen: number}>} A promise that is resolved
+   *   with the page index (starting from zero) that is associated with the
+   *   reference.
    */
   getPageIndex(ref) {
     return this._transport.getPageIndex(ref);
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with a lookup table for
-   *   mapping named destinations to reference numbers.
+   * @returns {Promise<Object<string,any>>} A promise that is resolved with
+   *   a lookup table for mapping named destinations to reference numbers.
    *
    * This can be slow for large documents. Use `getDestination` instead.
    */
@@ -632,74 +688,88 @@ class PDFDocumentProxy {
 
   /**
    * @param {string} id - The named destination to get.
-   * @returns {Promise} A promise that is resolved with all information
-   *   of the given named destination.
+   * @returns {Promise<Array<any>>} A promise that is resolved with all
+   *   information of the given named destination.
    */
   getDestination(id) {
     return this._transport.getDestination(id);
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Array} containing
-   *   the page labels that correspond to the page indexes, or `null` when
-   *   no page labels are present in the PDF file.
+   * @returns {Promise<Array<string> | null>} A promise that is
+   *   resolved with an {Array} containing the page labels that correspond to
+   *   the page indexes, or `null` when no page labels are present in the PDF
+   *   file.
    */
   getPageLabels() {
     return this._transport.getPageLabels();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with a {string} containing
-   *   the page layout name.
+   * @returns {Promise<string>} A promise that is resolved with a {string}
+   *   containing the page layout name.
    */
   getPageLayout() {
     return this._transport.getPageLayout();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with a {string} containing
-   *   the page mode name.
+   * @returns {Promise<string>} A promise that is resolved with a {string}
+   *   containing the page mode name.
    */
   getPageMode() {
     return this._transport.getPageMode();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Object} containing
-   *   the viewer preferences, or `null` when no viewer preferences are present
-   *   in the PDF file.
+   * @returns {Promise<Object>} A promise that is resolved with an {Object}
+   *   containing the viewer preferences.
    */
   getViewerPreferences() {
     return this._transport.getViewerPreferences();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Object} containing
-   *   the currently supported actions, or `null` when no OpenAction exists.
+   * @returns {Promise<any | null>} A promise that is resolved with an {Array}
+   *   containing the destination, or `null` when no open action is present
+   *   in the PDF.
    */
   getOpenAction() {
     return this._transport.getOpenAction();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with a lookup table for
-   *   mapping named attachments to their content.
+   * @returns {Promise<any>} A promise that is resolved with a lookup table
+   *   for mapping named attachments to their content.
    */
   getAttachments() {
     return this._transport.getAttachments();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Array} of all the
-   *   JavaScript strings in the name tree, or `null` if no JavaScript exists.
+   * @returns {Promise<Array<string> | null>} A promise that is
+   *   resolved with an {Array} of all the JavaScript strings in the name tree,
+   *   or `null` if no JavaScript exists.
    */
   getJavaScript() {
     return this._transport.getJavaScript();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Array} that is a
-   * tree outline (if it has one) of the PDF. The tree is in the format of:
+   * @typedef {Object} OutlineNode
+   * @property {string} title
+   * @property {boolean} bold
+   * @property {boolean} italic
+   * @property {Uint8ClampedArray} color
+   * @property {any} dest
+   * @property {string} url
+   * @property {Array<OutlineNode>} items
+   */
+
+  /**
+   * @returns {Promise<Array<OutlineNode>>} A promise that is resolved with
+   * an {Array} that is a tree outline (if it has one) of the PDF. The tree is
+   * in the format of:
    * [
    *   {
    *     title: string,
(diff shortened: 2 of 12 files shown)


tsconfig.json
```diff
@@ -0,0 +1,69 @@
...
+    // "importHelpers": true,                 /* Import emit helpers from 'tslib'. */
+    // "downlevelIteration": true,            /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
+    // "isolatedModules": true,               /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
+
+    /* Strict Type-Checking Options */
+    "strict": true,                           /* Enable all strict type-checking options. */
+    // "noImplicitAny": true,                 /* Raise error on expressions and declarations with an implied 'any' type. */
+    // "strictNullChecks": true,              /* Enable strict null checks. */
+    // "strictFunctionTypes": true,           /* Enable strict checking of function types. */
+    // "strictBindCallApply": true,           /* Enable strict 'bind', 'call', and 'apply' methods on functions. */
```
timvandermeij, line 30: Is there a reason to keep all commented code around? If not, we should get rid of it so only the options that are actually used in the file are visible.

timvandermeij (commented):
We do need a Gulp task here to generate the type annotations like in the other WIP PR because it should be automated and end up in each release. Another question I have is: given that we're not TypeScript users here, how can we be sure that the generated type annotations are valid (not broken by any changes we make)? Is there a validator for that that we can use?

tsconfig.json
ineiti: That's just the standard `tsc --init` boilerplate and can be cleaned up of course...

ineiti:
> How can we be sure that the generated type annotations are valid (not broken by any changes we make)? Is there a validator for that that we can use?

Hmm - the only idea I have is to create some smoke-screen tests that make sure all the types are still correct and can be at least compiled. Because once `tsc` compiles a code, it means that the types are OK.

ineiti:
> We do need a Gulp task here to generate the type annotations like in the other WIP PR because it should be automated and end up in each release.

I'm new to gulp, so I'd need some guidance here. Honestly, I was never able to understand where the different versions of the project come from.

Also, it's the first time I see the idea of converting jsdoc to ts-types ;) So any help from @oBusk and @tamuratak is very welcome! I'm currently not even sure how to merge all the `*.d.ts` files into one `pdf.d.ts` file - is it enough to concatenate all of them?

Is there some documentation that describes how the pdfjs-dist build is done? I stopped at trying to understand where the `PDFJSDev.eval(` comes from...

ineiti:
I'm not sure I'm doing the right thing here ;) But this _nearly_ works:

```
$ npx tsc -p .
src/display/text_layer.js:21:1 - error TS9005: Declaration emit for this file requires using private name 'TextLayerRenderTask'. An explicit type annotation may unblock declaration emit.

21 import {
   ~~~~~~
```

It complains about the `TextLayerRenderTask` definition, which is kept private in `text_layer.js`, which looks reasonable that it's a problem. But I don't know how to solve it... Any idea?

If I remove 
```
   * @returns {TextLayerRenderTask}
```
in `text_layer.js`, it compiles and creates a `pdf.d.ts` file which looks somewhat reasonable. Besides all the returns of `Object`, which are mostly useless, e.g., `getDocument` returns an `Object`.

So there's still a lot of work to do.

ineiti:
So this is starting to be nearly usable ;) After a
```
gulp dist
npx tsc -p .
```
and then importing the local pdf.js in my project using `tsconfig.app.json`-`paths`, it works! So I think I 'just' need to do the following (any help is welcome):

- [ ] add the typescript-definitions to the `dist` gulp-target - @timvandermeij is this the correct target?
- [ ] write a typescript spec-test - I'm looking forward to emulate webworkers in the test...
- [ ] clean up the JSDoc and export everything through `pdf.js`
- [ ] decide if `pdf.js` should be the entry-point, or if I should add an `index.js` that exports everything

timvandermeij:
I think the Gulp task was already made in https://github.com/mozilla/pdf.js/pull/10575/files and should already take care of bundling it in the distribution correctly.

ineiti:
OK - this latest commits add the `dts` gulp-target. Instead of using `jsdoc` from @oBusk , it uses `tsc` for creation of the types.

I chose to focus on `pdf.d.ts`, because that allows you to do the following:

```
import {getDocument, GlobalWorkerOptions, PDFDocumentProxy, PDFPageProxy} from "pdfjs-dist/build/pdf";
import pdfjsWorker from "pdfjs-dist/build/pdf.worker.entry";
GlobalWorkerOptions.workerSrc = pdfjsWorker;
```

But this means that all types will need to be exported by `pdf.js`. If somebody thinks this is a bad idea, please speak up and tell me why ;)

gulpfile.js
```diff
@@ -1141,6 +1142,13 @@ gulp.task("jsdoc", function (done) {
   });
 });
 
+gulp.task("dts", function (done) {
```
Snuffleupagus, line 1145: `dts` doesn't feel like a very descriptive task name, can we call it e.g. `types` instead?

Basically, let's replace all `dts`/`DTS` occurrences with `types`/`TYPES` instead.

tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
```
Snuffleupagus, line 1: The JSON format doesn't support comments, hence this file isn't actually valid.

src/core/annotation.js
```diff
@@ -45,8 +45,8 @@ class AnnotationFactory {
    * @param {Object} ref
    * @param {PDFManager} pdfManager
    * @param {Object} idFactory
-   * @returns {Promise} A promise that is resolved with an {Annotation}
-   *   instance.
+   * @returns {Promise<Annotation>} A promise that is resolved with an
```
Snuffleupagus, line 48: As mentioned *multiple* times in PR #10575, no files in the `src/core/` directory should be included in this PR since that code isn't actually accessible through the API (given that it runs in a worker-thread).

src/display/api.js
```diff
@@ -604,16 +618,16 @@ class PDFDocumentProxy {
   /**
    * @param {{num: number, gen: number}} ref - The page reference. Must have
    *   the `num` and `gen` properties.
-   * @returns {Promise} A promise that is resolved with the page index (starting
-   *   from zero) that is associated with the reference.
+   * @returns {Promise<{num: number, gen: number}>} A promise that is resolved
+   *   with the page index that is associated with the reference.
```
Snuffleupagus, line 622: This removed part of the existing documentation, i.e. the parenthesised part; please revert that.

src/display/api.js
```diff
@@ -69,6 +69,11 @@ const DefaultCMapReaderFactory =
     ? NodeCMapReaderFactory
     : DOMCMapReaderFactory;
 
+/**
+ * TODO https://github.com/mozilla/pdf.js/pull/10575
```
Snuffleupagus, line 73: *Here and elsewhere:* Please ensure that you remove any TODO-comments from the final PR.

src/display/api.js
```diff
@@ -69,6 +69,11 @@ const DefaultCMapReaderFactory =
...
+ * @typedef {any} IPDFStream
```
Snuffleupagus, line 74: Also, since the `setPDFNetworkStreamFactory` neither is *nor should* be exposed in the API, is this actually necessary?
(The definition can be found in https://github.com/mozilla/pdf.js/blob/master/src/interfaces.js)

gulpfile.js
```diff
@@ -1141,6 +1142,13 @@ gulp.task("jsdoc", function (done) {
...
+  console.log();
+  console.log("### Generating typescript definitions");
+
+  exec("tsc -p .", done);
```
Snuffleupagus, line 1149: Also, I do wonder if it'd make sense to define the configuration *inline* here rather than using a separate JSON file!?

For one thing, having the output directory of the `tsc` command controlled by an *external* file doesn't seem great when other gulp tasks (i.e. `dist-pre`) depend on being able to find the files is a particular spot.

src/display/api.js
```diff
@@ -2582,14 +2626,14 @@ class RenderTask {
      * Callback for incremental rendering -- a function that will be called
      * each time the rendering is paused.  To continue rendering call the
      * function that is the first argument to the callback.
-     * @type {function}
+     * @type {() => {}}
```
Snuffleupagus, line 2629: Without knowing anything about TypeScript, this does look a bit weird; is this actually correct!?

Snuffleupagus (changes requested):
Once you're getting this closer to done, please remember https://github.com/mozilla/pdf.js/wiki/Squashing-Commits
(Also, note that it's possible to list *multiple authors* of a commit, [see e.g. here](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line).)

src/display/api.js
```diff
@@ -185,7 +198,7 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  * @param {string|TypedArray|DocumentInitParameters|PDFDataRangeTransport} src
  * Can be a url to where a PDF is located, a typed array (Uint8Array)
  * already populated with data or parameter object.
- * @returns {PDFDocumentLoadingTask}
+ * @returns {PDFDocumentLoadingTasks}
```
Snuffleupagus, line 201: The class is named `PDFDocumentLoadingTask` (and that shouldn't be changed); why did you add a trailing `s` here?

src/display/api.js
```diff
@@ -775,16 +805,16 @@ class PDFDocumentProxy {
...
+   *   {DocumentInitParameters}, which are either needed in the viewer and/or
+   *   whose default values may be affected by the `apiCompatibilityParams`.
    */
   get loadingParams() {
     return this._transport.loadingParams;
   }
 
   /**
-   * @type {PDFDocumentLoadingTask} The loadingTask for the current document.
+   * @type {PDFDocumentLoadingTasks} The loadingTask for the current document.
```
Snuffleupagus, line 817: The class is named `PDFDocumentLoadingTask` (and that shouldn't be changed); why did you add a trailing `s` here?

gulpfile.js
ineiti: done

tsconfig.json
ineiti: removed and added arguments in `gulpfile.js`

src/display/api.js
ineiti: reverted

gulpfile.js
ineiti: done

src/display/api.js
ineiti: Hmm - it _is_ a valid typescript type (as far as I remember ;) And I put it there because it was complaining. Now I put back `function`, and it's OK. Go figure...

src/core/annotation.js
ineiti: reverted

src/display/api.js
ineiti: Looking back at the original PR, this was about moving those parts to another place:

https://github.com/mozilla/pdf.js/pull/10575#discussion_r260971909

Is this still a good idea?

src/display/api.js
ineiti: I have no idea. I want to use pdf.js mainly for showing pdfs on screen and get the text that is stored in the pdf. But there is so much more you can do with this library that honestly, I have no idea. You tell me to remove it, it's gone. Somebody tells me it's useful, it stays.

All I want is to show pages of a PDF document ;)

So, should it stay or should it go?

ineiti:
One commit a day keeps the doctor away... I included all comments from @Snuffleupagus , except those I'm not sure what to do. Currently you can change the

```
TYPES_INT = "tsc"; // or "jsdoc"
```
in `gulpfile.js`, and then do
```
gulp types
```
or
```
gulp dist
```

My TODO-list from above didn't get shorter, unfortunately. And while JSDoc seems to give better documentation, I cannot use it. TSC gives me better documentation, but currently some of it is missing :(

src/display/api.js
Snuffleupagus: > Somebody tells me it's useful, it stays.

Not quite; as mentioned above, this simply shouldn't be part of the public API.

gulpfile.js
```diff
@@ -1141,6 +1145,36 @@ gulp.task("jsdoc", function (done) {
...
 
+gulp.task("types", function (done) {
+  console.log();
+  console.log("### Generating typescript definitions using " + TYPES_INT);
+
+  if (TYPES_INT === "tsc") {
+    var args =
+      `--target ES2020 --allowJS --declaration --outDir ${TYPES_BUILD_DIR} ` +
+      `--strict --esModuleInterop --forceConsistentCasingInFileNames ` +
+      `--emitDeclarationOnly`;
```
Snuffleupagus, line 1156: This really ought to be an array instead, to improve readability. You can simply `join` the elements below instead.

gulpfile.js
```diff
@@ -1141,6 +1145,36 @@ gulp.task("jsdoc", function (done) {
...
+    exec(`tsc ${args} src/pdf.js`, done);
+  } else {
+    var JSDOC_FILES = [
+      "src/display/api.js",
+      "src/shared/util.js",
+      "src/core/annotation.js",
```
Snuffleupagus, line 1162: Again, please no `src/core/` files in this PR.

gulpfile.js
```diff
@@ -63,9 +63,13 @@ var GH_PAGES_DIR = BUILD_DIR + "gh-pages/";
 var SRC_DIR = "src/";
 var LIB_DIR = BUILD_DIR + "lib/";
 var DIST_DIR = BUILD_DIR + "dist/";
+var TYPES_BUILD_DIR = BUILD_DIR + "types/";
 var COMMON_WEB_FILES = ["web/images/*.{png,svg,gif,cur}", "web/debugger.js"];
 var MOZCENTRAL_DIFF_FILE = "mozcentral.diff";
 
+// This is either `tsc` or `jsdoc`
+var TYPES_INT = "tsc";
+
```
Snuffleupagus, line 72: Why do you need two *different* modes though, since I believe that it was mentioned (in the previous PRs) that the JSDoc solution wasn't "powerful"/"useful" enough.

Can the implementation be simplified to just use *one* mode, since having something that'd require *manual* editing of the gulpfile to toggle will basically render it unused any way?

ineiti: of course - this is still to test the difference. I did not see the reference to JSDoc not being "powerful" enough. There will only be one...

ineiti:
@Snuffleupagus - two questions from my TODO-list where I'd like to have your input:
- currently `tsc` chokes on `display/text_layer::TextLayerRenderTask` and `api::PDFDocumentLoadingTask` as they are defined in a somewhat 'private' way (I never saw this construct before...) - you can remove the trailing `_` to see how it fails. Do you have an easy idea how I can fix this?
- is it OK if I add an `index.js` which exports everything needed? Then I'll reference that one in the `package.json` as the types-file. Or should I call it `types.js`?

Snuffleupagus:
> two questions from my TODO-list where I'd like to have your input:

Unfortunately I don't really know anything about TypeScript, sorry!

---

My interest here is essentially limited to making sure that a solution, whatever it may end up looking like, is reasonable/maintainable from a general PDF.js-library perspective. And also, that we don't accidentally end up exposing *unintended* functionality through the public API (i.e. as defined in the `src/pdf.js` file).

tamuratak:
@ineiti  This PR, ineiti/pdf.js/pull/1, makes the return type of `getDocument` to be `PDFDocumentLoadingTask` with `tsc`.

oBusk:
I added https://github.com/ineiti/pdf.js/pull/2 to compile the definitions using typescript via gulp-typescript rather than `tsc`. `exec('tsc')` is flimsy, didn't run on my machine

ineiti:
> @ineiti This PR, [ineiti/pull/1](https://github.com/ineiti/pdf.js/pull/1), makes the return type of `getDocument` to be `PDFDocumentLoadingTask` with `tsc`.

thanks - merged.

ineiti:
> I added [ineiti#2](https://github.com/ineiti/pdf.js/pull/2) to compile the definitions using typescript via gulp-typescript rather than `tsc`. `exec('tsc')` is flimsy, didn't run on my machine

thanks - merged...

ineiti:
@oBusk - I think I merged too fast. The PR you gave only creates one file, and I cannot use it in my project afterwards... So now I'm playing with gulp-typescript, trying to create a valid `index.d.ts` file... If you want to look at it, I pushed the latest commits including your PR.

I also removed the `tsconfig.json`, as per @Snuffleupagus request, and inlined it in the gulpfile.js.

So this latest commit doesn't work :(

src/index.js
```diff
@@ -0,0 +1,183 @@
+/* Copyright 2012 Mozilla Foundation
```
Snuffleupagus, line 1: Please note that adding this file really isn't a good idea, since the *public API* is specified in `src/pdf.js` and having a file that you need to *manually* keep in sync isn't a desirable solution from a maintainability perspective.

Furthermore, this is also exporting *a lot* of things that the API doesn't/shouldn't expose.

ineiti: Thanks - that answers my question from before ;)

There is at least one dependency that is not in `pdf.js`, which is `PDFDocumentLoadingTask` - is it OK to add them in the `pdf.js` then?

src/display/api.js
ineiti: No reply - removed the TODOs.

src/display/api.js
ineiti: removed

gulpfile.js
ineiti: removed

src/index.js
ineiti: added there for the moment.

ineiti:
OK, merged the commits, added the contributors, went again through all comments on the code.

- I couldn't get @oBusk 's proposal of using `gulp-typescript` to work, due to https://github.com/ivogabe/gulp-typescript/issues/190 - if @oBusk sees a way to create all necessary files, I'm happy to use `gulp-typescript`.
- pointed `exec` to `node_modules/.bin/tsc` - @oBusk, can you please try again if this works?

The last task is how to test this thing - if I use typescript, I'll have to build it first before I can test it - is that a good idea?

timvandermeij:
> The last task is how to test this thing - if I use typescript, I'll have to build it first before I can test it - is that a good idea?

If the fact that it compiles is enough for the types to be valid, then such a test would be fine. For the unit tests for example we already build the required parts first before running the tests, so a test that would build the types and check if the build succeeds would probably be a good enough test.

ineiti:
> I really shouldn't be listed as an author on any of these patches, since I've not written a single line of code here :-)

Oh, managing PRs is a hassle, and you do a great job, so I don't mind having your name in there. But if you're afraid people will ask you about it when it breaks, I'll remove you ;)

ineiti:
OK - thanks for all the comments so far. Now I feel confident that all is in place - but of course I'll be happy to add/remove what you think is necessary.

I'm sure that the type-definitions are not ideal yet. I hope that other people can chime in here or once this is merged. Changes should be picked up automatically then.

tamuratak:
I want to confirm, is it acceptable to change the import and the export of `src/pdf.js`?

https://github.com/mozilla/pdf.js/pull/12102/files#diff-3841ca9d0e4702635b24c39b593d0610

```diff
diff --git a/src/pdf.js b/src/pdf.js
index c6a6d133..81e9c793 100644
--- a/src/pdf.js
+++ b/src/pdf.js
@@ -28,6 +28,9 @@ import {
   getDocument,
   LoopbackPort,
   PDFDataRangeTransport,
+  PDFDocumentLoadingTask,
+  PDFDocumentProxy,
+  PDFPageProxy,
   PDFWorker,
   setPDFNetworkStreamFactory,
   version,
@@ -136,6 +139,9 @@ export {
   getDocument,
   LoopbackPort,
   PDFDataRangeTransport,
+  PDFDocumentLoadingTask,
+  PDFDocumentProxy,
+  PDFPageProxy,
   PDFWorker,
   version,
   // From "./shared/util.js":
```

ineiti:
> I want to confirm, is it acceptable to change the import and the export of `src/pdf.js`?

And there might be more exports in the future!

If this is not acceptable, I see two alternatives:
1. use `index.js` to export everything, and reference that in the `package.json::types`, but @Snuffleupagus asked me to remove `index.js`
2. tell users to import non-`pdf.js` types from their sub-directory - but I consider this bad practice:
```
import {PDFDocumentLoadingTask} from "pdfjs-dist/display/api"
```

tamuratak:
Since we have type inference in TypeScript, I think we do not have to export these types. We do not have to write these types explicitly in codes generally.

ineiti:
> Since we have type inference in TypeScript, I think we do not have to export these types. We do not have to write these types explicitly in codes generally.

I could probably rewrite my code, but I know that I have stuff like:
```
class Something {
    pdfPage: PDFPageProxy;

    loadPage(){
        ....
        this.pdfPage = this.doc.getPage(1);
    }
}
```

And for this it's nice to have the types.

tamuratak:
Are you saying that because you do not want to rewrite your codebase, PDF.js should change to fit yours?

ineiti:
> Are you saying that because you do not want to rewrite your codebase, PDF.js should change to fit yours?

I would never dare proposing such a thing ;) But after 40 years of programming I know very well that I often write non-optimal code... So proposing a class that I'm writing makes me add: _I could probably rewrite my code_...

Not exporting these types makes it hard:
- for the IDE to know what methods to propose for auto-completion
- for the typescript compiler to know the method exists on the class and to pass when using `strict`
- to pass `PDFPageProxy` around: how can I put it in the arguments of the method if the type doesn't exist? (hint: I don't want to use `any`)

I'm OK to not export classes that are for internal use only. So I did delete the references @Snuffleupagus asked me to delete. But `PDFPageProxy`, `PDFDocumentLoadingTask`, and `PDFDocumentProxy` are actually visible to the user of PDF.js, even if it's only as returned values, so I think the types should be available to typescript. 

As I wrote above, there are probably also other classes that should be exported, too, but I didn't do much more than printing pages on the screen for the moment.

tamuratak:
Some of your points are wrong:

> for the IDE to know what methods to propose for auto-completion

For auto-completion, it is enough that types are properly written in `*.d.ts` files. It is not necessary that they are exported.

> for the typescript compiler to know the method exists on the class and to pass when using strict

For this purpose, again, it is enough that types are properly written in `*.d.ts` files. It is not necessary that they are exported.

ineiti:
> Some of your points are wrong:

Indeed - IntelliJ is smart enough to find the definitions itself. So the question is whether 

```
import {PDFDocumentProxy, PDFPageProxy} from "pdfjs-dist/display/api";
```
is nicer or
```
import {PDFDocumentProxy, PDFPageProxy} from "pdfjs-dist";
```

If I understand you, you prefer the former...

So I merged the pending PRs, applied linting, squashed, rebased, force-pushed, ...

ineiti:
@tamuratak - might the error in the travis-test https://travis-ci.org/github/mozilla/pdf.js/builds/711717631#L329
```
couldn't compile typescript test: build/dist/build/display/api.d.ts(888,134): error TS2339: Property 'annotationStorage' does not exist on type 'RenderParameters'.
```
be due to one of the annotations you added? Can you have a look at it please?

tamuratak:
> If I understand you, you prefer the former...

No. I am saying that `import {PDFDocumentProxy, PDFPageProxy}` is not necessary.

ineiti:
> No, you don't understand. I am saying that `import {PDFDocumentProxy, PDFPageProxy}` is not necessary.

Please have a look at my comment at ineiti/pdf.js#6

Maxchii:
Glad to see types are being worked on! thank you!
Is there a compiled types.d files which can be used for the time being while this is being worked on?

ineiti:
> Glad to see types are being worked on! thank you!
> Is there a compiled types.d files which can be used for the time being while this is being worked on?

I use the following - of course you'll have to adjust the paths ;)

https://github.com/ineiti/pdf_example/blob/master/src/tsconfig.app.json

Then in the checked-out pdf.js repo you can do `gulp dist`, which will create a distribution for you.

ineiti:
Here we go - all merged. The only difference with @tamuratak is the `main.ts` test-file, else it's all merged.

ineiti:
@tamuratak merged your PR, but now it fails again - I cannot find the `annotationStorage` mentioned in the failing test, and locally it passes.

https://travis-ci.org/github/mozilla/pdf.js/builds/712678774#L332

Just for the record: dfddadfsfsdkfkkdkjkj;kkkfssfsfssssslfjkjl;kjlkjlkjlksfsfsdkljdkljkjldfsakskfldjklj (types my 17-year old daughters)...

tamuratak:
@ineiti Run `git rebase master`

It's getting to be a pain in the ass to explain to you. I want to edit PR directly. Could you please close this PR so I can create a new one?

ineiti:
> It's getting to be a pain in the ass to explain to you.

That's quite some language there - I'm merely merging your PRs and they make the build fail. So, yes, I should better test your PRs before merging...

> Could you please close this PR so I can create a new one?

Also, you're always free to create your own PR, just like I did with this one.

tamuratak:
> Also, you're always free to create your own PR, just like I did with this one.

Creating another new PR for the exact same issue is not a good idea because it forces the development team of PDF.js to do more work.

ineiti:
> Creating another new PR for the exact same issue is not a good idea because it forces the development team of PDF.js to do more work.

Then just feel free to continue working on this PR. As a non-paid contributor, I'm happy to learn and to share the knowledge I have. This is my payment for contributing to open source software. Plus being able to scratch my itch, of course.

Let's see whether the PR in its current form is OK for the team to be merged. 

@timvandermeij @Snuffleupagus ready for merge?

ineiti:
Before you merge, there is an important thing you need to take into consideration: as has been shown with the addition of `annotationStorage`, some of the fields you'll be adding over time might make `ts-test` fail! 

So you'll have some additional care-taking to do after this PR. Of course I think it is worth it ;) But please be aware that this PR is not a `merge and forget` one.

src/display/text_layer.js
```diff
@@ -23,24 +23,39 @@ import {
...
  *   and shall be initially be set to empty array.
  * @property {number} [timeout] - Delay in milliseconds before rendering of the
  *   text runs occurs.
  * @property {boolean} [enhanceTextSelection] - Whether to turn on the text
  *   selection enhancement.
  */
+
+/**
+ * @typedef {Object} TextLayerRenderTask
+ *
```
timvandermeij, line 47: Let's remove all empty lines in this comment block to match the other blocks and make it more compact.

test/ts-types/main.ts
```diff
@@ -0,0 +1,27 @@
+import {getDocument} from "pdfjs-dist";
+
+class MainTest {
+    task: ReturnType<typeof getDocument> | undefined
```
timvandermeij, line 4: Also use two-space indentation in this file for consistency with the other code.

test/ts-types/main.ts
```diff
@@ -0,0 +1,27 @@
+import {getDocument} from "pdfjs-dist";
```
timvandermeij, line 1: `{ getDocument }` (spaces for consistency with all other imports)

test/ts-types/main.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    constructor(public file: string) {
+    }
+
+    loadPdf() {
+        this.task = getDocument("file://" + this.file);
+        return this.task.promise;
+    }
+
+    get pdf() {
```
timvandermeij, line 14: This method and the one below is unused in the test, so let's remove it to have a minimal test.

test/ts-types/main.ts
```diff
@@ -0,0 +1,27 @@
...
+        return this.task?.promise;
+    }
+
+    get page() {
+        return this.pdf?.then(doc => doc.getPage(1));
+    }
+
+}
+
+const mt = new MainTest("./test.pdf");
```
timvandermeij, line 24: There is no such file in this directory, so how can this work? I presume this `main.ts` file is used to ensure that the generated types can be used again, which is great, so can't this test be even more minimal by simply using some of the types and being done? No need to actually load a file I think.

timvandermeij (commented):
I think this looks good. It took a couple of iterations by multiple people, but I hope we can converge to one PR that we can merge. I'm aware that this will require us to keep the documentation and types up-to-date, but given that the types are now generated from our code makes me far less worried about that than before when the types were separate from our code. That's also why I think a test is important, because without a failing test we wouldn't know if anything broke, so thank you for including that!

It looks like we can merge this after the comments are addressed. I do have some questions, as stated in the comments, about the test since I'm not entirely seeing if the purpose is just to test if the compiled types can be used again, which is what I expected, or if it's supposed to do more than that given that it actually loads a file.

tamuratak:
@timvandermeij  Notice that `npx gulp jsdoc` fails too.

test/ts-types/main.ts
ineiti: In fact the test has only one goal: check if it can compile `main.ts` with `strict: true` (which is even more strictly checking everything compiles well). Once the compilation succeeds, the test is successful.

Personally I would've liked to also import some other types, but @tamuratak insisted that only one method is imported here. Seeing that it correctly failed for the missing `annotationStorage` parameter, this seems to be enough.

gulpfile.js
```diff
@@ -1336,6 +1352,41 @@ gulp.task(
...
+  gulp.series(
+    "testing-pre",
+    "generic",
+    "types",
+
+    function () {
+      var packageJsonSrc = packageBowerJson()[0];
+
+      return merge([
+        packageJsonSrc.pipe(gulp.dest(DIST_DIR)),
```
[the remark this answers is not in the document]
ineiti: OK - changed this to `BUILD_DIR + "typestest/"` - is that better?

src/display/api.js
```diff
@@ -495,11 +544,14 @@ const PDFDocumentLoadingTask = (function PDFDocumentLoadingTaskClosure() {
 
 /**
  * Abstract class to support range requests file loading.
- * @param {number} length
- * @param {Uint8Array} initialData
- * @param {boolean} progressiveDone
  */
 class PDFDataRangeTransport {
+  /**
+   *
```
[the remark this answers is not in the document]
ineiti: I hope I got em all now using a `grep` for comment-begin and comment-end.

src/display/api.js
```diff
@@ -613,16 +665,17 @@ class PDFDocumentProxy {
...
+   *   reference.
    */
   getPageIndex(ref) {
     return this._transport.getPageIndex(ref);
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with a lookup table for
-   *   mapping named destinations to reference numbers.
+   * @returns {Promise< Object<string,any> >} A promise that is resolved with
```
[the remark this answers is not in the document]
ineiti: Seems to be only this file, so I removed all `< ` with `<` and ` >` with `>` in the comments.

src/display/api.js
```diff
@@ -632,74 +685,88 @@ class PDFDocumentProxy {
...
    */
   getDestination(id) {
     return this._transport.getDestination(id);
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Array} containing
-   *   the page labels that correspond to the page indexes, or `null` when
-   *   no page labels are present in the PDF file.
+   * @returns {Promise< Array<string> > | Promise<null>} A promise that is
```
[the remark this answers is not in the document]
ineiti: done

src/display/api.js
```diff
@@ -632,74 +685,88 @@ class PDFDocumentProxy {
...
+   *   for mapping named attachments to their content.
    */
   getAttachments() {
     return this._transport.getAttachments();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Array} of all the
-   *   JavaScript strings in the name tree, or `null` if no JavaScript exists.
+   * @returns {Promise< Array<string> > | Promise<null>} A promise that is
```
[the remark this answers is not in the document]
ineiti: done

src/display/api.js
```diff
@@ -879,8 +948,8 @@ class PDFDocumentProxy {
  * @property {boolean} [renderInteractiveForms] - Whether or not
  *                     interactive form elements are rendered in the display
  *                     layer. If so, we do not render them on canvas as well.
- * @property {Array}  [transform] - Additional transform, applied
- *                    just before viewport transform.
+ * @property {Array<any>}  [transform] - Additional transform, applied
```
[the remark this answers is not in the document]
ineiti: Interesting - I tried out https://www.npmjs.com/package/eslint-plugin-jsdoc to format all jsdoc, and the above is in fact the default output of `eslint-plugin-jsdoc`...

Anyway - removing the space for now. Perhaps it might be worth adding the `eslint-plugin-jsdoc` to the tools?

src/display/api.js
```diff
@@ -719,45 +786,46 @@ class PDFDocumentProxy {
...
     return this._transport.getPermissions();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Object} that has
-   *   `info` and `metadata` properties. `info` is an {Object} filled with
-   *   anything available in the information dictionary and similarly
-   *   `metadata` is a {Metadata} object with information from the metadata
-   *   section of the PDF.
+   * @returns {Promise<{ info: any, metadata: any }>} A promise that is
```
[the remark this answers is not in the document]
ineiti: also added `metadata: Metadata`

ineiti:
> @timvandermeij Notice that `npx gulp jsdoc` fails too.

Hmm - that's strange. I was able to re-write the callbacks in api.js like this:
https://github.com/ineiti/pdf.js/blob/add_types_annotations/src/display/api.js#L435

and then it passes `gulp jsdoc`.

However, there are other places like that happily accept callback definitions:
https://github.com/ineiti/pdf.js/blob/add_types_annotations/src/display/text_layer.js#L48

But, if I copy the line from `text_layer.js` to `api.js`, `gulp jsdoc` complains again...

ineiti:
@timvandermeij - I closed the trivial comments that have been fixed, but left the other comments open where I think you need to make sure that my proposition is correct. I hope that matches your normal workflow.

gulpfile.js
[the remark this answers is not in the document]
timvandermeij: Definitely!

test/ts-types/main.ts
timvandermeij: Yes, that makes the intent more clear.

ineiti:
I see some regressions failed - but it looks like its in the rendering part, which I think we didn't touch. Is there anything we can do to fix this? Or is this expected?

timvandermeij:
No worries about those; those are known intermittent failures that have nothing to do with your patch and should be fixed once we upgrade Puppeteer in #12123.

dqisme:
So excited pdfjs is on the way to TypeScript! 🍺
It's been a long time...

timvandermeij (approved):
The most important acceptance criteria for TypeScript definitions were that the public API may not be changed (so no changes in `src/pdf.js`), that the solution is maintainable for PDF.js developers (by not having the types separated from the code but generated from the JSDoc comments) and that the solution is tested automatically (so that we know the types will work and get a failing test otherwise).

This pull request checks all those boxes. It improves the existing documentation quite a bit and gives an extra incentive to keep the documentation up-to-date because a test now fails if for example a new parameter is not added to the corresponding JSDoc comment.

All in all, I think this is nice work from everyone involved and we can now merge it! We are open for contributions from the community to improve the documentation even further, not only for better type annotations but also for the better API documentation.

Thank you for your work on this!

Snuffleupagus:
A couple of questions:
 - Was the `jsdoc` command tested, to ensure that the current https://mozilla.github.io/pdf.js/api/draft/index.html page isn't breaking because of all these JSDoc changes.
 - I'm not sure that running this new test on Travis *by default* was such a great idea, since besides increasing the runtime quite a bit it *also* means that any TypeScript errors will not (likely) be seen as a blocker to landing patches.
   Basically, the PDF.js library is still a *JavaScript project*, and accepting this PR does *not* absolve TypeScript-users from helping to maintain the types. It does seem, in my opinion, to be somewhat unfair to now essentially require TypeScript-knowledge from every PDF.js contributor (which is what having these tests run by default amounts to).
 - It would have been good if the newly added `typedef`s followed the existing formatting a lot better, please see the expected format in https://github.com/mozilla/pdf.js/blob/00a8b42e6703d824049eb317c05ada2eba3799a2/src/display/api.js#L106-L176

I'd really like to get the above points addressed, one way or another, since especially the second one may quickly become an issue.

timvandermeij:
> Was the jsdoc command tested

Yes, I looked at the bot output and also ran it locally after rebasing onto the current master before merging, also because this was commented as not working in an earlier version of the patch. I also ran all test commands locally after rebasing to make sure it works locally as well as after the most recently merged patches.

> I'm not sure that running this new test on Travis by default was such a great idea, since besides increasing the runtime quite a bit it also means that any TypeScript errors will now (likely) be seen as a blocker to landing patches.

The runtime impacts from generating the TypeScript definitions is minimal in my testing (at most few seconds). Moreover, it only seems to fail if there is a mismatch between the actual code and the JSDoc comments. This showed up in an earlier version of the patch where the `annotationStorage` parameter was missing, and playing locally with this patch I found the same. It's basically a safeguard that forces us to keep the JSDoc comments up-to-date, which I only see as a good thing. No TypeScript specific knowledge should be required for this. If it ever turns out to be the case, we can always disable the tests if need be, but since we're distributing the types I'd really like to have _some_ way of knowing that what we distribute is correct.

> It would have been good if the newly added typedefs followed the existing formatting a lot better

Agreed, but I must note that consistency was already an issue in that file even before this patch to begin with, and can be dealt with in a follow-up. Even before I had already seen JSDoc comments where the descriptions are separated with a dash versus without, are with/without two spaces on the next line, are aligned on the next line, et cetera.

Snuffleupagus:
> No TypeScript specific knowledge should be required for this.

Based on some of the changes seen in this patch, I do have some doubts about that; hence why this worries me!

As long as we're still going to, on principle, accept patches that fail the `typestest` (and opt for disabling the test instead), I suppose that somewhat alleviates my primary concern here.

src/display/api.js
```diff
@@ -632,74 +688,88 @@ class PDFDocumentProxy {
...
   getPageMode() {
     return this._transport.getPageMode();
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Object} containing
-   *   the viewer preferences, or `null` when no viewer preferences are present
-   *   in the PDF file.
+   * @returns {Promise<Object>} A promise that is resolved with an {Object}
+   *   containing the viewer preferences.
```
Snuffleupagus, line 726: This remove part of the existing comment, why?

src/display/api.js
```diff
@@ -632,74 +688,88 @@ class PDFDocumentProxy {
...
   }
 
   /**
-   * @returns {Promise} A promise that is resolved with an {Array} containing
-   *   the page labels that correspond to the page indexes, or `null` when
-   *   no page labels are present in the PDF file.
+   * @returns {Promise<Array<string> | null>} A promise that is
+   *   resolved with an {Array} containing the page labels that correspond to
+   *   the page indexes, or `null` when no page labels are present in the PDF
+   *   file.
```
Snuffleupagus, line 702: This first line is now weirdly short, compared to the rest of them, it looks like this was incorrectly re-formatted.

timvandermeij:
Definitely. My reasoning is that everything that we distribute should be tested, and preferably automatically and in the earliest stage possible so we quickly find out about potential problems. If this PR did not have such a test I would rather not accept it especially because this is not a TypeScript project, but a (vanilla) JavaScript one. That would mean that we would distribute TypeScript definitions without having a clue if they actually work, which is not great.

Having a test at least tells us that a patch is going to break the TypeScript definitions. allowing us to consider that and make a decision on what to do with that. Most if not all the time it's likely to turn out to be a missing JSDoc comment update, which is not only useful for the TypeScript definitions but more importantly also for the consistency of our own documentation (since we've seen that this tends to be forgotten sometimes). I therefore see `typestest` as one that kills two birds with one stone: it ensures internal documentation consistency to a certain degree and it ensures that externally exposed type definitions are validated. Not running a test like that would hide such problems.

I can't really imagine any failures from `typestest` to be difficult to solve. Of course, if I turn out to be wrong here and we get errors that we cannot work with, we can definitely disable the test and make a follow-up issue to get it working again, if necessary by asking the various TypeScript contributors for assistance.

Hopefully this helps a bit with the concerns here!

timvandermeij:
Moreover, I'll make some time today to address the comments above and work on overall comment consistency within the `src/display/api.js` file since those are valid points and I'd like to take that concern away.

src/display/api.js
timvandermeij: I missed this in the review and will put that back as part of the upcoming patch. Thank you for noticing!

src/display/api.js
timvandermeij: I think so. I'll address that in the upcoming patch too.

gulpfile.js
```diff
@@ -1336,6 +1352,42 @@ gulp.task(
...
+            GENERIC_DIR + "build/pdf.js",
+            GENERIC_DIR + "build/pdf.worker.js",
+            SRC_DIR + "pdf.worker.entry.js",
+          ])
+          .pipe(gulp.dest(TYPESTEST_DIR + "build/")),
+        gulp
+          .src(TYPES_BUILD_DIR + "**/**")
+          .pipe(gulp.dest(TYPESTEST_DIR + "build/")),
+      ]);
+    },
```
Snuffleupagus, line 1379: Also, defining this inline isn't really helping readability of the overall task, and it really ought to be moved into a helper function/task (e.g. `typestest-pre`) similar to what's done in other parts of the gulpfile.

gulpfile.js
```diff
@@ -1336,6 +1352,42 @@ gulp.task(
...
+    function (done) {
+      exec(`node_modules/.bin/tsc -p test/types`, function (err, stdout) {
+        if (err !== null) {
```
Snuffleupagus, line 1382: This should probably just be `if (err) {`, right?

Maxchii:
@timvandermeij Any ETA on when these changes will be reflected in the pdfjs-dist repository?

timvandermeij:
No, there is no release date planned yet.

SleeplessByte:
If someone can build it, and post the types here, we can manually list the file in our projects until it's released.

timvandermeij:
The types are available in the current pre-release (beta) version of PDF.js; see https://github.com/mozilla/pdf.js/releases and https://www.npmjs.com/package/pdfjs-dist/v/next.

PR #12106, opened 2020-07-22
[api-minor] Add an annotation storage in order to save annotation data in acroforms

calixteman:
The goal of this storage is to be able to save form data from the html layer and then be able to print and save them.

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -532,6 +532,7 @@ class WorkerMessageHandler {
               task,
               intent: data.intent,
               renderInteractiveForms: data.renderInteractiveForms,
+              annotationStorage: data.annotationStorage,
             })
             .then(
               function (operatorListInfo) {
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -140,6 +140,7 @@ class AnnotationElement {
     this.imageResourcesPath = parameters.imageResourcesPath;
     this.renderInteractiveForms = parameters.renderInteractiveForms;
     this.svgFactory = parameters.svgFactory;
+    this.annotationStorage = parameters.annotationStorage;
 
     if (isRenderable) {
       this.container = this._createContainer(ignoreBorder);
@@ -1450,6 +1451,7 @@ class AnnotationLayer {
         imageResourcesPath: parameters.imageResourcesPath || "",
         renderInteractiveForms: parameters.renderInteractiveForms || false,
         svgFactory: new DOMSVGFactory(),
+        annotationStorage: parameters.annotationStorage,
       });
       if (element.isRenderable) {
         parameters.div.appendChild(element.render());
--- a/src/display/annotation_storage.js
+++ b/src/display/annotation_storage.js
@@ -1,0 +1,57 @@
+/* Copyright 2020 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class AnnotationStorage {
+  constructor() {
+    this._storage = {};
+  }
+
+  /**
+   * Get the value for a given key if it exists
+   * or store and return the default value
+   *
+   * @public
+   * @memberof AnnotationStorage
+   * @param {string} key
+   * @param {Object} defaultValue
+   * @returns {Object}
+   */
+  getOrCreateValue(key, defaultValue) {
+    if (key in this._storage) {
+      return this._storage[key];
+    }
+
+    this._storage[key] = defaultValue;
+    return defaultValue;
+  }
+
+  /**
+   * Set the value for a given key
+   *
+   * @public
+   * @memberof AnnotationStorage
+   * @param {string} key
+   * @param {Object} value
+   */
+  setValue(key, value) {
+    this._storage[key] = value;
+  }
+
+  getAll() {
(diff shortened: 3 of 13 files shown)


src/core/worker.js
```diff
@@ -517,6 +517,7 @@ class WorkerMessageHandler {
       "GetOperatorList",
       function wphSetupRenderPage(data, sink) {
         var pageIndex = data.pageIndex;
+        const annotationStorage = data.annotationStorage;
```
timvandermeij, line 520: Since the annotation storage is only used below, we don't need a variable here and can just use `annotationStorage: data.annotationStorage,` below.

src/display/annotation_layer.js
```diff
@@ -542,7 +543,6 @@ class CheckboxWidgetAnnotationElement extends WidgetAnnotationElement {
    */
   render() {
     this.container.className = "buttonWidgetAnnotation checkBox";
-
```
timvandermeij, line 545: This change should be reverted since it's not related to the storage.

src/display/annotation_storage.js
```diff
@@ -0,0 +1,57 @@
+/* Copyright 2014 Mozilla Foundation
```
timvandermeij, line 1: 2020

src/display/annotation_storage.js
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class AnnotationStorage {
+  constructor(storage = {}) {
```
timvandermeij, line 17: I don't see this argument being provided anywhere, so can't we just remove it and use `this._storage = {};` below for simplicity?

src/display/annotation_storage.js
```diff
@@ -0,0 +1,57 @@
...
+    this._storage = storage;
+  }
+
+  /**
+   * Get the value for a given key if it doesn't exist
```
timvandermeij, line 22: `if it exists or store and return the default value`

src/display/annotation_storage.js
```diff
@@ -0,0 +1,57 @@
...
+   * or store and return the defaultValue
+   *
+   * @public
+   * @memberof AnnotationStorage
+   * @param {String} key
```
timvandermeij, line 27: `String` -> `string` since `typeof 'foo' === 'string'` (same for the method below)

src/display/annotation_storage.js
```diff
@@ -0,0 +1,57 @@
...
+   * @param {Object} defaultValue
+   * @returns {Object}
+   */
+  getOrCreate(key, defaultValue) {
```
timvandermeij, line 31: `getOrCreateValue` is maybe a more consistent name with `setValue` below?

src/display/annotation_storage.js
```diff
@@ -0,0 +1,57 @@
...
+   * @public
+   * @memberof AnnotationStorage
+   * @param {String} key
+   * @param {Object} value
+   */
+  setValue(key, value) {
+    this._storage[key] = value;
+  }
+
+  getDict() {
```
timvandermeij, line 52: `getAll` maybe since that avoids having the type in the name (that should be in a JSDoc comment above the method similar to the other ones).

src/display/api.js
```diff
@@ -576,6 +577,14 @@ class PDFDocumentProxy {
   constructor(pdfInfo, transport) {
     this._pdfInfo = pdfInfo;
     this._transport = transport;
+    this._annotationStorage = new AnnotationStorage();
+  }
+
+  /**
+   * @type {AnnotationStorage} storage for annotations.
```
timvandermeij, line 584: Perhaps `Storage for annotation data in forms` is more descriptive here?

test/unit/jasmine-boot.js
```diff
@@ -13,26 +13,26 @@
...
- included in all copies or substantial portions of the Software.
-
- THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
- EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
- MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
- NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
- LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
- OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
- WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+   Copyright (c) 2008-2016 Pivotal Labs
```
timvandermeij, line 16: These whitespace changes should be reverted to keep the diff small and related to the storage.

timvandermeij (commented):
Looks good to me with these comments addressed. Thank you for splitting this up so we can get this merged soon!

src/display/annotation_storage.js
calixteman: pffff... copy/paste....

test/unit/jasmine-boot.js
calixteman: `gulp lint --fix` side effect... or maybe I did something wrong: dunno.

src/display/annotation_storage.js
```diff
@@ -0,0 +1,57 @@
...
+    this._storage[key] = defaultValue;
+    return defaultValue;
+  }
+
+  /**
+   * Set the value for a given key
+   *
+   * @public
+   * @memberof AnnotationStorage
+   * @param {String} key
```
timvandermeij, line 45: `String` -> `string` was done above, but missed here.

timvandermeij (approved):
Good to merge with the final comment addressed and the commits squashed into one. Thank you!

timvandermeij (approved):

timvandermeij:
First part done; thank you!

PR #12107, opened 2020-07-22
Add support for checkboxes printing

calixteman:
This patch allows to print filled checkbox with the appearance they've in the PDF file.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -49,18 +49,21 @@ class AnnotationFactory {
    *   instance.
    */
   static create(xref, ref, pdfManager, idFactory) {
-    return pdfManager.ensure(this, "_create", [
-      xref,
-      ref,
-      pdfManager,
-      idFactory,
-    ]);
+    return pdfManager.ensureDoc("acroForm").then(acroForm => {
+      return pdfManager.ensure(this, "_create", [
+        xref,
+        ref,
+        pdfManager,
+        idFactory,
+        acroForm,
+      ]);
+    });
   }
 
   /**
    * @private
    */
-  static _create(xref, ref, pdfManager, idFactory) {
+  static _create(xref, ref, pdfManager, idFactory, acroForm) {
     const dict = xref.fetchIfRef(ref);
     if (!isDict(dict)) {
       return undefined;
@@ -78,6 +81,7 @@ class AnnotationFactory {
       subtype,
       id,
       pdfManager,
+      acroForm: acroForm instanceof Dict ? acroForm : Dict.empty,
     };
 
     switch (subtype) {
@@ -514,8 +518,9 @@ class Annotation {
       return Promise.resolve(new OperatorList());
     }
 
+    const appearance = this.appearance;
     const data = this.data;
-    const appearanceDict = this.appearance.dict;
+    const appearanceDict = appearance.dict;
     const resourcesPromise = this.loadResources([
       "ExtGState",
       "ColorSpace",
@@ -533,14 +538,14 @@ class Annotation {
       opList.addOp(OPS.beginAnnotation, [data.rect, transform, matrix]);
       return evaluator
         .getOperatorList({
-          stream: this.appearance,
+          stream: appearance,
           task,
           resources,
           operatorList: opList,
         })
         .then(() => {
           opList.addOp(OPS.endAnnotation, []);
-          this.appearance.reset();
+          appearance.reset();
           return opList;
         });
     });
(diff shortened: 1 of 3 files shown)


test/unit/annotation_spec.js
```diff
@@ -1659,6 +1684,60 @@ describe("annotation", function () {
...
+          annotationStorage[id] = true;
+          return annotation.getOperatorList(
+            partialEvaluator,
+            task,
+            false,
+            annotationStorage
+          );
+        }, done.fail)
+        .then(opList => {
+          expect(opList.argsArray.length).toEqual(3);
```
timvandermeij, line 1734: I think we should also check the `opList.fnArray` value here since aside from the arguments the functions (commands) are equally important.

timvandermeij (commented):
Comments from an intial look. Could you also rebase onto the current master so we only have the relevant checkbox changes in here? It will be a bit easier to review and test then. Thanks!

test/unit/annotation_spec.js
```diff
@@ -1659,6 +1684,60 @@ describe("annotation", function () {
...
+      AnnotationFactory.create(
+        xref,
+        buttonWidgetRef,
+        pdfManagerMock,
+        idFactoryMock
+      )
+        .then(annotation => {
+          const id = annotation.data.id;
+          const annotationStorage = {};
+          annotationStorage[id] = true;
```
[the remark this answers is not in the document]
calixteman: I tried that but unfortunately it leads to an error

timvandermeij:
I'm going to play around with it a bit more, but it works really well so far!

src/core/annotation.js
```diff
@@ -1003,6 +1039,13 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
 
     this.data.exportValue =
       exportValues[0] === "Off" ? exportValues[1] : exportValues[0];
+
+    const normalAppearanceState = customAppearance.get("N");
```
timvandermeij, line 1043: The spec seems to call this just `normalAppearance`, so let's use that name too.

src/core/annotation.js
```diff
@@ -980,6 +991,31 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
+        const operatorList = super.getOperatorList(
+          evaluator,
+          task,
+          renderForms,
+          annotationStorage
+        );
+        this.appearance = savedAppearance;
+        return operatorList;
+      }
+      return Promise.resolve(new OperatorList());
```
timvandermeij, line 1009: Shouldn't we treat this case just like the caser where we have no `annotationStorage`, i.e., let this fall through and let the parent class handle it? I would think we only want to special-case the case where we have a value and a checked appearance; any other case would need the default parent `getOperatorList` call.

test/unit/annotation_spec.js
```diff
@@ -1659,6 +1685,64 @@ describe("annotation", function () {
         done();
       }, done.fail);
     });
+
+    it("should get rendered annotation for printing", function (done) {
```
timvandermeij, line 1689: `"should render checkboxes for printing"` (since radio buttons will be added in a next patch in another test).

test/unit/annotation_spec.js
```diff
@@ -1659,6 +1685,64 @@ describe("annotation", function () {
         done();
       }, done.fail);
     });
+
```
timvandermeij, line 1688: Let's move this new unit test just above the `"should handle radio buttons with a field value"` test so the checkbox tests are nicely grouped.

test/unit/annotation_spec.js
```diff
@@ -1659,6 +1685,64 @@ describe("annotation", function () {
...
+    it("should get rendered annotation for printing", function (done) {
+      buttonWidgetDict.set("V", Name.get("1"));
```
timvandermeij, line 1690: Since we're not testing the field value in this test, we might not even need to set it.

timvandermeij (commented):
Looks good to me once these comments are addressed.

src/core/annotation.js
calixteman: If we've a checked box (in the base pdf) and we uncheck it then falling through implies that we'll print a checked box.
Anyway in reading the spec, we can have a "Off" appearance so I need to handle that case.

test/unit/annotation_spec.js
```diff
@@ -1630,6 +1656,62 @@ describe("annotation", function () {
       }, done.fail);
     });
 
+    it("should render annotation for printing", function (done) {
```
timvandermeij, line 1659: This unit test should be named `should render checkboxes for printing` to make the distinction with radio buttons for a later patch, and it should be moved up just above the radio button unit test above (line 1625) so that the checkbox tests are grouped.

timvandermeij (approved):
Good to merge once the final comment is addressed!

timvandermeij (approved):

timvandermeij:
Nice work!

PR #12108, opened 2020-07-22
Add support for radios printing

calixteman:


timvandermeij:
When the checkbox PR is merged, I think this only requires a rebase and a unit test similar to the checkbox one but for radio buttons.

src/core/annotation.js
```diff
@@ -1076,6 +1076,7 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
         break;
       }
     }
+    this.checkedAppearance = normalAppearanceState.get(this.data.buttonValue);
```
timvandermeij, line 1079: Once the checkbox PR is merged, this most likely also requires setting `uncheckedAppearance`.

src/display/annotation_layer.js
```diff
@@ -584,15 +584,35 @@ class RadioButtonWidgetAnnotationElement extends WidgetAnnotationElement {
...
-    element.name = this.data.fieldName;
-    if (this.data.fieldValue === this.data.buttonValue) {
+    element.name = data.fieldName;
+    if (value) {
       element.setAttribute("checked", true);
     }
 
+    element.addEventListener("change", function (event) {
+      const name = event.target.name;
+      for (const el of document.getElementsByName(name)) {
```
timvandermeij, line 605: `el` -> `element` for better readability (we try to avoid abbrevations/shortened names)

calixteman: element is already used so I changed it for `radio`

calixteman:
I added a test so this PR is ready to be reviewed.

src/core/annotation.js
```diff
@@ -1088,6 +1088,8 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
```
timvandermeij, line 1091: Let's add a newline above this line like we did for the checkboxes to separate the appearance from the button value.

src/core/annotation.js
```diff
@@ -1088,6 +1088,8 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
+    this.uncheckedAppearance = normalAppearanceState.get("Off") || null;
```
timvandermeij, line 1092: Hm, this was already there, but it might be better to rename `normalAppearanceState` to `normalAppearance` here too like we did for the checkboxes to match the specification more closely.

test/unit/annotation_spec.js
```diff
@@ -1741,6 +1741,93 @@ describe("annotation", function () {
         done();
       }, done.fail);
     });
+
+    it("should render radios for printing", function (done) {
```
timvandermeij, line 1745: `radios` -> `radio buttons` for consistency with the other tests.

test/unit/annotation_spec.js
```diff
@@ -1741,6 +1741,93 @@ describe("annotation", function () {
...
+            OPS.beginAnnotation,
+            OPS.setFillRGBColor,
+            OPS.endAnnotation,
+          ]);
+          expect(opList.argsArray[1]).toEqual(
+            new Uint8ClampedArray([26, 51, 76])
+          );
+          return annotation;
+        }, done.fail)
+        .then(annotation => {
```
timvandermeij, line 1808: Testing the unchecked flow is a pretty good addition that we didn't think of when the checkboxes were made. Could we perhaps add a second commit to the PR that does two things:

1. Move the checkbox unit test from line 1659 to to line 1624 so it's grouped with the other checkbox tests (now it's a checkbox test in between radio buttons tests which makes it a bit hard to find).
2. Introduce the same flow for testing the unchecked case there too.

Doing so would be a nice way to improve checkbox test coverage and for consistency.

timvandermeij (commented):
Looks good with the final comments addressed!

timvandermeij (approved):

timvandermeij:
Another piece of the puzzle done. Thank you!

PR #12112, opened 2020-07-23
Add support for textfied & choice widgets printing

calixteman:


calixteman:
@timvandermeij, ping

brendandahl:
Tim, if you don't have time, I free to review this as well.

timvandermeij:
Like the original forms PR this one is pretty big, making it difficult to oversee that all cases are handled and tested. Can we split off the multiline field handling to a follow-up PR to limit the scope here a bit so it's easier to review?

src/shared/util.js
```diff
@@ -789,6 +789,10 @@ function stringToPDFString(str) {
   return strBuf.join("");
 }
 
+function escapeString(str) {
```
timvandermeij, line 792: Why is `escapeString` necessary in this PR? It's not trivial to read, so if we really need it it would need a unit test like the other `util.js` functions.

src/core/annotation.js
```diff
@@ -888,18 +888,314 @@ class WidgetAnnotation extends Annotation {
...
     if (renderForms) {
       return Promise.resolve(new OperatorList());
     }
-    return super.getOperatorList(
-      evaluator,
-      task,
-      renderForms,
-      annotationStorage
+
+    if (!this.data.hasText) {
```
timvandermeij, line 895: Does `hasText` indicate that it's an editable widget? Maybe `editable` would be a better name then?

src/core/annotation.js
```diff
@@ -888,18 +888,314 @@ class WidgetAnnotation extends Annotation {
...
+    const defaultPadding = 2;
+
+    // Default horizontal padding: can we have an heuristic to guess it?
+    const hPadding = defaultPadding;
+    const totalHeight = this.data.rect[3] - this.data.rect[1];
+    const totalWidth = this.data.rect[2] - this.data.rect[0];
+
+    const fontInfo = await this.getFontData(evaluator, task);
+    const [font, fontName] = fontInfo;
+    let [, , fontSize] = fontInfo;
```
timvandermeij, line 978: This looks a bit odd. Can we at least use underscores for the unused parts? Otherwise `let fontSize = fontInfo[2]` would probably be more readable.

calixteman:
> Like the original forms PR this one is pretty big, making it difficult to oversee that all cases are handled and tested. Can we split off the multiline and comb field handling to a follow-up PR to limit the scope here so it's easier to review?

Ok I splited the PR into 3: #12175, #12176 and #12177.
It'd be nice to review all of them and so I can fix problems asap.
We'd like to have print & save support in next Firefox nightly.

timvandermeij:
Thank you! I'll take care of reviewing them today. There has been quite a lot of activity lately for PDF.js, which is _really_ good, but also takes up quite some time to get everything reviewed. The previous patches turned out really good though, so I have no doubt we can get these merged soon as well!

timvandermeij:
Closing this one in favor of the split up ones.

PR #12114, opened 2020-07-23
Remove a couple of unnecessary `PDFJSDev` checks from the viewer

Snuffleupagus:
- Given the `DefaultExternalServices` implementation, the `PDFViewerApplication.supportsDocumentFonts` getter is guaranteed to be defined and we can thus remove some (now) unnecessary `PDFJSDev` checks from the `webViewerInitialized` function.
 - By slightly tweaking the "pdfBugEnabled" definition in `web/app_options`, similar to the existing ones for "workerSrc" and "cMapUrl", we can remove some `PDFJSDev` checks from the `PDFViewerApplication._parseHashParameters` method.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -299,11 +299,7 @@ const PDFViewerApplication = {
    * @private
    */
   async _parseHashParameters() {
-    if (
-      typeof PDFJSDev !== "undefined" &&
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12124, opened 2020-07-26
Include the browser name when printing unit-test results

Snuffleupagus:
This uses a similar format to the reference-test logging, and will help determine in *exactly* which browser the failure occurred (since the tests run concurrently).

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -750,7 +750,8 @@ function unitTestPostHandler(req, res) {
     }
     var session = getSession(data.browser);
     session.numRuns++;
-    var message = data.status + " | " + data.description;
+    var message =
+      data.status + " | " + data.description + " | in " + session.name;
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Interestingly, the failing unit-tests seem to (as I suspected) be limited to Google Chrome on Linux.

timvandermeij:
I have made #12125 to hopefully improve the situation a bit. I can reproduce the intermittent unit tests failures for Chrome on Linux, but with that patch it _seems_ to be less tests failing and less frequent. We'll see...

Snuffleupagus:
> I have made #12125 to hopefully improve the situation a bit.

Thanks, fingers crossed that it helps :-)

---

Anyway, I do believe that this patch is useful regardless since it makes the logs easier to read.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12125, opened 2020-07-26
Improve test bot stability

### LINKED ISSUE #11998: Remove extraction log suppression code
I filed https://github.com/puppeteer/puppeteer/issues/5741 before and it just got fixed upstream. If version > 3.3.0 is released, we should upgrade and remove the extraction log suppression code since it's not necessary anymore.

timvandermeij:


timvandermeij:
Hm, not sure what's up with Windows here. Maybe an old process is still running?

test/test.js
```diff
@@ -828,8 +827,9 @@ function startBrowsers(rootUrl, initSessionCallback) {
       `&delay=${options.statsDelay}` +
       `&masterMode=${options.masterMode}`;
     const startUrl = rootUrl + queryParameters;
+    const headless = options.strictVerify;
```
Snuffleupagus, line 830: *Probably bike-shedding, but anyway:* Would it make *slightly* more sense to use e.g. the `noPrompts` option here since that's "feels" somewhat easier to logically related to a no-UI mode!?

timvandermeij:
The Windows bot needs a kick first before continuing here.

Snuffleupagus (approved):
Looks good to me; thanks for getting the tests working again!

PR #12126, opened 2020-07-26
Attempt to reduce intermittent failures in the "cleans up document resources during rendering of page" unit-test

Snuffleupagus:
This patch should *hopefully* remove the `Unhandled promise rejection: ...` errors, by returning the "final" promise. Also, by pausing/delaying of rendering slightly the likelihood of the test failing in the first place should thus be reduced.

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -1906,7 +1906,11 @@ describe("api", function () {
               viewport,
             });
 
-            pdfDoc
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice find! It hasn't failed anymore and it's also more consistent with the unit tests above it where the cleanup promise is also returned.

PR #12127, opened 2020-07-26
Improve how Type3-fonts with dependencies are handled

Snuffleupagus:
While the `CharProcs` streams of Type3-fonts *usually* don't rely on dependencies, such as e.g. images, it does happen in some cases.

Currently any dependencies are simply appended to the parent operatorList, which in practice means *only* the operatorList of the *first* page where the Type3-font is being used.
However, there's one thing that's slightly unfortunate with that approach: Since fonts are global to the PDF document, we really ought to ensure that any Type3 dependencies are appended to the operatorList of *all* pages where the Type3-font is being used. Otherwise there's a theoretical risk that, if one page has its rendering paused, another page may try to use a Type3-font whose dependencies are not yet fully resolved. In that case there would be errors, since Type3 operatorLists are executed synchronously.

Hence this patch, which ensures that all relevant pages will have Type3 dependencies appended to the main operatorList. (Note here that the `OperatorList.addDependencies` method, via `OperatorList.addDependency`, ensures that a dependency is only added *once* to any operatorList.)

Finally, these changes also remove the need for the "waiting for the main-thread"-hack that was added to `PartialEvaluator.buildPaintImageXObject` as part of fixing issue 10717.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -456,7 +456,7 @@ class PartialEvaluator {
     const transfers = imgData ? [imgData.data.buffer] : null;
 
     if (this.parsingType3Font) {
-      return this.handler.sendWithPromise(
+      return this.handler.send(
         "commonobj",
         [objId, "FontType3Res", imgData],
         transfers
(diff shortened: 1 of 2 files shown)


timvandermeij:
This time there was an intermittent unit test failure on Windows too. I'm going to compile a list of the failures on the intermittent unit test failures issue.

timvandermeij (approved):

timvandermeij:
Looks good to me; thanks!

PR #12131, opened 2020-07-27
[api-minor] Allow loading pdf fonts into another document.

### LINKED ISSUE #8271: Font lost when rendered in an iframe
I'm trying to use an iframe to print a PDF easily, everything is ok but the fonts.

Configuration:
- Web browser and its version: chrome 57
- Operating system and its version: MacOS Sierra 10.12.7
- PDF.js version: pdfjs-dist 1.8.177

1. Dynamically create an iframe with a canvas, render the pdf in the canvas

Here is my print function which takes the pdfDoc as parameter (from `PDFJS.getDocument(url)`)

```javascript
function print(pdfDoc) {
      if (!pdfDoc || !pdfDoc.numPages) {
        return ;
      }

      var scale = 5;
      var iframe = window.document.createElement('iframe');
      var doc = iframe.contentDocument || iframe.contentWindow.document;
    
      window.document.body.append(iframe);

      doc.open();
      doc.write(
        '<!DOCTYPE html>' +
          '<html>' +
            '<head>' +
              '<style>' +
                '* {margin: 0 !important; padding: 0 !important}' +
                'div{width:612pt; height:792pt;}' +
                'canvas{width:100%; height:100%;}' +
            '</style>' +
          '</head>' +
          '<body></body>' +
        '</html>'
      );
      doc.close();

      function renderPage(num) {
        console.log('Rendering page %d of %d', num, pdfDoc.numPages);

        return pdfDoc
          .getPage(num)
          .then(function (page) {
            var div = doc.createElement('div');
            var canvas = doc.createElement('canvas');
            div.appendChild(canvas);
            doc.body.appendChild(div);
            var viewport = page.getViewport(scale);
            var ctx = canvas.getContext('2d');
            canvas.width = viewport.width;
            canvas.height = viewport.height;
            return page.render({canvasContext: ctx, viewport: viewport}).promise;
          })
          .then(function () {
            if (num < pdfDoc.numPages) {
              return renderPage(num + 1);
            }
          });
      }

      renderPage(1)
        .then(function () {
          iframe.contentWindow.print();
          iframe.remove();
        });
    }
```

On the left, the main page canvas, on the right, the iframe canvas.

image

The rendering in the iframe seems to not embed the fonts.

jsg2021:
This adds support to specify a document context the pdf will render into when getting the pdf document. I wanted to automatically infer the document, based on the container element, but the fonts are loaded in a way that is detached from the viewer. So this makes the getDocument support a param `ownerDocumentContext` (better name suggestions are welcome)

Fixes #8271

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -152,6 +152,9 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *   parsed font data from the worker-thread. This may be useful for debugging
  *   purposes (and backwards compatibility), but note that it will lead to
  *   increased memory usage. The default value is `false`.
+ * @property {HTMLDocument} [ownerDocument] - Specify an explicit document
+ *   context to create elements with and to load resources, such as fonts,
+ *   into. Defaults to the current document.
  * @property {boolean} [disableRange] - Disable range request loading
  *   of PDF files. When enabled, and if the server supports partial content
  *   requests, then the PDF will be fetched in chunks.
@@ -267,6 +270,9 @@ function getDocument(src) {
   }
   if (typeof params.disableFontFace !== "boolean") {
     params.disableFontFace = apiCompatibilityParams.disableFontFace || false;
+  }
+  if (typeof params.ownerDocument === "undefined") {
+    params.ownerDocument = globalThis.document;
   }
 
   if (typeof params.disableRange !== "boolean") {
@@ -907,9 +913,10 @@ class PDFDocumentProxy {
  * @alias PDFPageProxy
  */
 class PDFPageProxy {
-  constructor(pageIndex, pageInfo, transport, pdfBug = false) {
+  constructor(pageIndex, pageInfo, transport, ownerDocument, pdfBug = false) {
     this._pageIndex = pageIndex;
     this._pageInfo = pageInfo;
+    this._ownerDocument = ownerDocument;
     this._transport = transport;
     this._stats = pdfBug ? new StatTimer() : null;
     this._pdfBug = pdfBug;
@@ -1036,7 +1043,9 @@ class PDFPageProxy {
       intentState.streamReaderCancelTimeout = null;
     }
 
-    const canvasFactoryInstance = canvasFactory || new DefaultCanvasFactory();
+    const canvasFactoryInstance =
+      canvasFactory ||
+      new DefaultCanvasFactory({ ownerDocument: this._ownerDocument });
     const webGLContext = new WebGLContext({
       enable: enableWebGL,
     });
@@ -1944,6 +1953,7 @@ class WorkerTransport {
     this.fontLoader = new FontLoader({
       docId: loadingTask.docId,
       onUnsupportedFeature: this._onUnsupportedFeature.bind(this),
+      ownerDocument: params.ownerDocument,
     });
     this._params = params;
     this.CMapReaderFactory = new params.CMapReaderFactory({
@@ -2387,6 +2397,7 @@ class WorkerTransport {
           pageIndex,
           pageInfo,
           this,
+          this._params.ownerDocument,
           this._params.pdfBug
         );
         this.pageCache[pageIndex] = page;
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -65,11 +65,16 @@ class BaseCanvasFactory {
 }
 
 class DOMCanvasFactory extends BaseCanvasFactory {
+  constructor({ ownerDocument = globalThis.document } = {}) {
+    super();
+    this._document = ownerDocument;
+  }
+
   create(width, height) {
     if (width <= 0 || height <= 0) {
       throw new Error("Invalid canvas size");
     }
-    const canvas = document.createElement("canvas");
+    const canvas = this._document.createElement("canvas");
     const context = canvas.getContext("2d");
     canvas.width = width;
     canvas.height = height;
--- a/src/display/font_loader.js
+++ b/src/display/font_loader.js
@@ -25,12 +25,17 @@ import {
 } from "../shared/util.js";
 
 class BaseFontLoader {
-  constructor({ docId, onUnsupportedFeature }) {
+  constructor({
+    docId,
+    onUnsupportedFeature,
+    ownerDocument = globalThis.document,
+  }) {
     if (this.constructor === BaseFontLoader) {
       unreachable("Cannot initialize BaseFontLoader.");
     }
     this.docId = docId;
     this._onUnsupportedFeature = onUnsupportedFeature;
+    this._document = ownerDocument;
 
     this.nativeFontFaces = [];
     this.styleElement = null;
@@ -38,15 +43,15 @@ class BaseFontLoader {
 
   addNativeFontFace(nativeFontFace) {
     this.nativeFontFaces.push(nativeFontFace);
-    document.fonts.add(nativeFontFace);
+    this._document.fonts.add(nativeFontFace);
   }
 
   insertRule(rule) {
     let styleElement = this.styleElement;
     if (!styleElement) {
-      styleElement = this.styleElement = document.createElement("style");
+      styleElement = this.styleElement = this._document.createElement("style");
       styleElement.id = `PDFJS_FONT_STYLE_TAG_${this.docId}`;
-      document.documentElement
+      this._document.documentElement
         .getElementsByTagName("head")[0]
         .appendChild(styleElement);
     }
@@ -56,8 +61,8 @@ class BaseFontLoader {
   }
 
   clear() {
-    this.nativeFontFaces.forEach(function (nativeFontFace) {
-      document.fonts.delete(nativeFontFace);
+    this.nativeFontFaces.forEach(nativeFontFace => {
+      this._document.fonts.delete(nativeFontFace);
     });
     this.nativeFontFaces.length = 0;
 
@@ -116,7 +121,8 @@ class BaseFontLoader {
   }
 
   get isFontLoadingAPISupported() {
-    const supported = typeof document !== "undefined" && !!document.fonts;
+    const supported =
+      typeof this._document !== "undefined" && !!this._document.fonts;
     return shadow(this, "isFontLoadingAPISupported", supported);
   }
 
@@ -146,8 +152,8 @@ if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")) {
   // PDFJSDev.test('CHROME || GENERIC')
 
   FontLoader = class GenericFontLoader extends BaseFontLoader {
-    constructor(docId) {
-      super(docId);
+    constructor(params) {
+      super(params);
       this.loadingContext = {
         requests: [],
         nextRequestId: 0,
@@ -254,7 +260,7 @@ if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")) {
       let i, ii;
 
       // The temporary canvas is used to determine if fonts are loaded.
-      const canvas = document.createElement("canvas");
+      const canvas = this._document.createElement("canvas");
       canvas.width = 1;
       canvas.height = 1;
       const ctx = canvas.getContext("2d");
@@ -316,22 +322,22 @@ if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")) {
       }
       names.push(loadTestFontId);
 
-      const div = document.createElement("div");
+      const div = this._document.createElement("div");
       div.style.visibility = "hidden";
       div.style.width = div.style.height = "10px";
       div.style.position = "absolute";
       div.style.top = div.style.left = "0px";
 
       for (i = 0, ii = names.length; i < ii; ++i) {
-        const span = document.createElement("span");
+        const span = this._document.createElement("span");
         span.textContent = "Hi";
         span.style.fontFamily = names[i];
         div.appendChild(span);
       }
-      document.body.appendChild(div);
+      this._document.body.appendChild(div);
 
-      isFontReady(loadTestFontId, function () {
-        document.body.removeChild(div);
+      isFontReady(loadTestFontId, () => {
+        this._document.body.removeChild(div);
         request.complete();
       });
       /** Hack end */
--- a/test/unit/custom_spec.js
+++ b/test/unit/custom_spec.js
@@ -107,3 +107,119 @@ describe("custom canvas rendering", function () {
       .catch(done.fail);
   });
 });
+
+describe("alternate document context", function () {
+  const FontFace = global.FontFace;
+
+  let altDocument;
+  let CanvasFactory;
+  let elements;
+
+  beforeEach(() => {
+    global.FontFace = function MockFontFace(name) {
+      this.family = name;
+    };
+
+    elements = [];
+    const createElement = name => {
+      const element = {
+        tagName: name,
+        remove() {
+          this.remove.called = true;
+        },
+      };
+      if (name === "style") {
+        element.sheet = {
+          cssRules: [],
+          insertRule(rule) {
+            this.cssRules.push(rule);
+          },
+        };
+      }
+      elements.push(element);
+      return element;
+    };
+    altDocument = {
+      fonts: new Set(),
+      createElement,
+      documentElement: {
+        getElementsByTagName: () => [{ appendChild: () => {} }],
+      },
+    };
+
+    CanvasFactory = isNodeJS
+      ? new NodeCanvasFactory()
+      : new DOMCanvasFactory({ ownerDocument: altDocument });
+  });
+
+  afterEach(() => {
+    global.FontFace = FontFace;
+    CanvasFactory = null;
+    elements = null;
+  });
+
+  it("should use given document for loading fonts (with Font Loading API)", async function () {
+    const getDocumentParams = buildGetDocumentParams(
+      "TrueType_without_cmap.pdf",
+      {
+        disableFontFace: false,
+        ownerDocument: altDocument,
+      }
+    );
+
+    const loadingTask = getDocument(getDocumentParams);
+    const doc = await loadingTask.promise;
+    const page = await doc.getPage(1);
+
+    const viewport = page.getViewport({ scale: 1 });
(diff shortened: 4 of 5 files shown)


src/display/api.js
```diff
@@ -1942,6 +1942,7 @@ class WorkerTransport {
     this.loadingTask = loadingTask;
     this.commonObjs = new PDFObjects();
     this.fontLoader = new FontLoader({
+      ownerDocumentContext: params.ownerDocumentContext,
```
Snuffleupagus, line 1945: Please move this to the bottom of the list instead, to be consistent with ordering used in ` src/display/font_loader.js`.

src/display/font_loader.js
```diff
@@ -25,28 +25,35 @@ import {
...
+    docId,
+    onUnsupportedFeature,
+    ownerDocumentContext = globalThis.document,
+  }) {
     if (this.constructor === BaseFontLoader) {
       unreachable("Cannot initialize BaseFontLoader.");
     }
     this.docId = docId;
     this._onUnsupportedFeature = onUnsupportedFeature;
+    this._ownerDocumentContext = ownerDocumentContext;
```
Snuffleupagus, line 38: Why not simply `this._document = ownerDocument;` instead, to reduce the verbosity in the rest of the code?

Snuffleupagus (changes requested):
> (better name suggestions are welcome)

How about simply `ownerDocument`, which seems like slightly less of a mouthful :-)

Also, please note that all `getDocument` parameters should be documented in https://github.com/mozilla/pdf.js/blob/403816040ef64ae17758439948771c2920f96ac4/src/display/api.js#L100-L170

Furthermore, would it be possible to add a unit-test to ensure that this works (and continues to work) correctly?
A suitable place for the test might be in https://github.com/mozilla/pdf.js/blob/master/test/unit/custom_spec.js

---

Finally, please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

jsg2021:
I'll make those changes 😊 

I was avoiding _document and ownerDocument to attempt to avoid confusion with the pdf document. though, that's not really referenced throughout the fontloader. I'll use those names. 😊

src/display/api.js
jsg2021: Will do

src/display/api.js
```diff
@@ -152,6 +152,9 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *   parsed font data from the worker-thread. This may be useful for debugging
  *   purposes (and backwards compatibility), but note that it will lead to
  *   increased memory usage. The default value is `false`.
+ * @property {HTMLDocument} [ownerDocument] - Specify an explicit document
+ *   context to create elements for and to load resouces into, such as
```
Snuffleupagus, line 156: Should `for` perhaps be replaced by `with` above?

Also, please fix the spelling of `resouces` -> `resources` here.

src/display/api.js
```diff
@@ -152,6 +152,9 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
...
+ *   fonts. Defaults to the current document.
```
Snuffleupagus, line 157: @timvandermeij Do you think this documentation is clear enough for users, or do you have any suggestions for e.g. improved wording here (or any other suggests about the overall implementation)?

test/unit/custom_spec.js
```diff
@@ -107,3 +107,91 @@ describe("custom canvas rendering", function () {
...
+
+  let CanvasFactory;
+
+  beforeAll(() => {
+    global.FontFace = class {};
+    CanvasFactory = isNodeJS ? new NodeCanvasFactory() : new DOMCanvasFactory();
+  });
+
+  afterAll(() => {
+    global.FontFace = FontFace;
```
Snuffleupagus, line 122: Please also add `CanvasFactory = null;` here.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,91 @@ describe("custom canvas rendering", function () {
...
+        this.cssRules.push(rule);
+      },
+    };
+    const getDocumentParams = buildGetDocumentParams(
+      "TrueType_without_cmap.pdf",
+      {
+        disableFontFace: false,
+        ownerDocument: {
+          fonts: null,
+          createElement(x) {
```
Snuffleupagus, line 167: As a rule we don't want to use one-letter variable/argument names, since those are generally too non-descriptive.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,91 @@ describe("custom canvas rendering", function () {
...
+    const viewport = page.getViewport({ scale: 1 });
+    const canvasAndCtx = CanvasFactory.create(viewport.width, viewport.height);
+
+    await page.render({
+      canvasContext: canvasAndCtx.context,
+      viewport,
+    }).promise;
+
+    expect(domDoc.fonts.size).toBe(1);
+
```
Snuffleupagus, line 148: You should also be destroying the `CanvasFactory` as well, both here and in the next test.

Snuffleupagus (changes requested):

src/display/api.js
jsg2021: derp. Sorry about that.

test/unit/custom_spec.js
jsg2021: Yeah, was going for a minimum viable mock for the interface. I've updated,

test/unit/custom_spec.js
```diff
@@ -107,3 +107,97 @@ describe("custom canvas rendering", function () {
...
+    const viewport = page.getViewport({ scale: 1 });
+    const canvasAndCtx = CanvasFactory.create(viewport.width, viewport.height);
+
+    await page.render({
+      canvasContext: canvasAndCtx.context,
+      viewport,
+    }).promise;
+
+    expect(domDoc.fonts.size).toBe(1);
```
timvandermeij, line 149: Is there not something we can assert about the font here as well, for example its name? Just to make sure that the expected font is loaded.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,97 @@ describe("custom canvas rendering", function () {
...
+    const domDoc = { fonts: new Set() };
+    const getDocumentParams = buildGetDocumentParams(
+      "TrueType_without_cmap.pdf",
+      {
+        disableFontFace: false,
+        ownerDocument: domDoc,
+      }
+    );
+
+    const task = getDocument(getDocumentParams);
```
timvandermeij, line 136: I think this is usually called `loadingTask` in other tests.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,97 @@ describe("custom canvas rendering", function () {
...
+
+    expect(domDoc.fonts.size).toBe(1);
+
+    await doc.destroy();
+    await task.destroy();
+    CanvasFactory.destroy(canvasAndCtx);
+    expect(domDoc.fonts.size).toBe(0);
+  });
+
+  it("should use given document for loading fonts (legacy path)", async function () {
```
timvandermeij, line 157: It could be that I'm missing a bit of knowledge about the font loader here, but what legacy path is meant here?

timvandermeij (commented):
Looks good to me aside from these comments.

test/unit/custom_spec.js
jsg2021: We can assert the font name was loaded.

test/unit/custom_spec.js
jsg2021: There are two paths. One path adds `FontFace` instances directly to `document.fonts` and the other creates a `<style>` element and appends `@font-face` rules to it.

jsg2021:
I just had an idea, and if it works, it would make this fix 100% transparent. :) 

now that the document is plumbed through, I might be able to set ownerDocument on the pdfDocument in viewer.setDocument.

src/display/display_utils.js
```diff
@@ -65,11 +65,16 @@ class BaseCanvasFactory {
...
+    super();
+    this._document = ownerDocument;
+  }
+
   create(width, height) {
     if (width <= 0 || height <= 0) {
       throw new Error("Invalid canvas size");
     }
-    const canvas = document.createElement("canvas");
+    const canvas = this._document.createElement("canvas");
```
Snuffleupagus, line 77: Is this actually a *necessary* change for this patch though, since I cannot imagine that the exact `document` which you use to just *create* the element matters!?

This would e.g. make the `DOMCanvasFactory` and `NodeCanvasFactory` interfaces slightly different (which you'd want to avoid).

test/unit/custom_spec.js
```diff
@@ -107,3 +107,104 @@ describe("custom canvas rendering", function () {
...
+
+    await page.render({
+      canvasContext: canvasAndCtx.context,
+      viewport,
+    }).promise;
+
+    expect(domDoc.fonts.size).toBe(1);
+    const [font] = Array.from(domDoc.fonts);
+    console.dir(font);
+    expect(font.family).toMatch(/g_d[0-9a-f]+_f1/);
```
Snuffleupagus, line 154: This regular expression, and the one below, isn't entirely correct since the `docId` is always a number and nothing else. Hence you should drop the `a-f` part, and this can be simplified to `/g_d\d+_f1/` (similarily below).

test/unit/custom_spec.js
Snuffleupagus: Probably clearer to call this test `should use given document for loading fonts (with CSS rules)`.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,104 @@ describe("custom canvas rendering", function () {
...
+    };
+    CanvasFactory = isNodeJS ? new NodeCanvasFactory() : new DOMCanvasFactory();
+  });
+
+  afterAll(() => {
+    global.FontFace = FontFace;
+    CanvasFactory = null;
+  });
+
+  it("should use given document for loading fonts", async function () {
```
Snuffleupagus, line 128: Probably clearer to call this test `should use given document for loading fonts (with Font Loading API)`.

Snuffleupagus (changes requested):

src/display/display_utils.js
jsg2021: Yes. The canvas only has access to the resources owned by the creating (owner) document. So, calls to measure/draw text will be inaccurate without this. 

I didn't just change this without cause, I still had missing fonts on some pages. Patching the creation of canvases to use the owner document fixes it.

The interface is the same. They have an optional argument for the constructor (Node version doesn't care, so it ignores) I could explicitly add that to the node constructor?

test/unit/custom_spec.js
jsg2021: I'll update the test names to use those labels.

test/unit/custom_spec.js
jsg2021: I didn't find the source of that string, so I went with a safe-ish pattern. I'm trying to tread lightly and respectfully with my changes. 😊 

I'll simplify the patterns.

src/display/display_utils.js
jsg2021: If you insert (adopt) a canvas created in a foreign document into a new document, it will gain access to fonts loaded in your document and lose access to fonts from its original document. This point is especially important for scratch/off-screen canvases.

jsg2021:
> I just had an idea, and if it works, it would make this fix 100% transparent. :)
> 
> now that the document is plumbed through, I might be able to set ownerDocument on the pdfDocument in viewer.setDocument.

After testing this idea, it turns out it won't work. The document is already in use by the time `setDocument` gets called. We will just have to set ownerDocument when calling `getDocument` from the API if we intend to render the viewer in another document.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,104 @@ describe("custom canvas rendering", function () {
...
+    const canvasAndCtx = CanvasFactory.create(viewport.width, viewport.height);
+
+    await page.render({
+      canvasContext: canvasAndCtx.context,
+      viewport,
+    }).promise;
+
+    expect(domDoc.fonts.size).toBe(1);
+    const [font] = Array.from(domDoc.fonts);
+    console.dir(font);
```
timvandermeij, line 153: Left-over debug print should be removed.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,104 @@ describe("custom canvas rendering", function () {
...
+    expect(font.family).toMatch(/g_d\d+_f1/);
+
+    await doc.destroy();
+    await loadingTask.destroy();
+    CanvasFactory.destroy(canvasAndCtx);
+    expect(domDoc.fonts.size).toBe(0);
```
timvandermeij, line 159: Can/should we do a similar check in the test below?

test/unit/custom_spec.js
```diff
@@ -107,3 +107,104 @@ describe("custom canvas rendering", function () {
...
+      "TrueType_without_cmap.pdf",
+      {
+        disableFontFace: false,
+        ownerDocument: domDoc,
+      }
+    );
+
+    const loadingTask = getDocument(getDocumentParams);
+    const doc = await loadingTask.promise;
+
```
timvandermeij, line 140: Let's remove this newline like in the test below.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,104 @@ describe("custom canvas rendering", function () {
...
+describe("alternate document context", function () {
+  const FontFace = global.FontFace;
+
+  let CanvasFactory;
+
+  beforeAll(() => {
+    global.FontFace = function MockFontFace(name) {
+      this.family = name;
+    };
+    CanvasFactory = isNodeJS ? new NodeCanvasFactory() : new DOMCanvasFactory();
```
timvandermeij, line 120: The `DOMCanvasFactory` gets no parameter here, even though that was added above. Is that expected?

timvandermeij (commented):
Looks good to me with these final comments addressed. If @Snuffleupagus is also happy with it, we can merge it.

test/unit/custom_spec.js
jsg2021: yep

test/unit/custom_spec.js
jsg2021: I've added it

src/display/api.js
```diff
@@ -152,6 +152,9 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *   parsed font data from the worker-thread. This may be useful for debugging
  *   purposes (and backwards compatibility), but note that it will lead to
  *   increased memory usage. The default value is `false`.
+ * @property {HTMLDocument} [ownerDocument] - Specify an explicit document
+ *   context to create elements with and to load resources into, such as
+ *   fonts. Defaults to the current document.
```
Snuffleupagus, line 157: This *might* flow a little better with:
```suggestion
 * @property {HTMLDocument} [ownerDocument] - Specify an explicit document
 *   context to create elements with and to load resources, such as fonts,
 *   into. Defaults to the current document.
```

src/display/display_utils.js
```diff
@@ -65,11 +65,16 @@ class BaseCanvasFactory {
 }
 
 class DOMCanvasFactory extends BaseCanvasFactory {
+  constructor(ownerDocument = globalThis.document) {
```
Snuffleupagus, line 68: This needs to be future-proof, to avoid issues if/when we need to add more parameters here; hence please change to
```suggestion
  constructor({ ownerDocument = globalThis.document, } = {}) {
```
and update the call-site in `src/display/api.js` accordingly.

src/display/api.js
Snuffleupagus: Furthermore, please add a default value [after this line](https://github.com/mozilla/pdf.js/blob/c5663f2f6b95680c9b7ce0c0f1e0021aa22deebe/src/display/api.js#L270), similar to the other `getDocument` options:
```javascript
if (typeof params.ownerDocument === "undefined") {
  params.ownerDocument = globalThis.document;
}
```

src/display/api.js
```diff
@@ -907,9 +910,10 @@ class PDFDocumentProxy {
  * @alias PDFPageProxy
  */
 class PDFPageProxy {
-  constructor(pageIndex, pageInfo, transport, pdfBug = false) {
+  constructor(pageIndex, pageInfo, transport, ownerDocument, pdfBug = false) {
     this._pageIndex = pageIndex;
     this._pageInfo = pageInfo;
+    this._document = ownerDocument;
```
Snuffleupagus, line 916: Please keep the `ownerDocument` name here, given the updated `DOMCanvasFactory` signature (see comment above).

Snuffleupagus (changes requested):
Looks OK to me, with the final comments addressed and all test passing.

timvandermeij (approved):

timvandermeij:
Thank you for your contribution! (The reference tests failures are unrelated intermittents.)

jsg2021:
awesome 😊

test/unit/custom_spec.js
```diff
@@ -107,3 +107,119 @@ describe("custom canvas rendering", function () {
       .catch(done.fail);
   });
 });
+
+describe("alternate document context", function () {
+  const FontFace = global.FontFace;
```
Snuffleupagus, line 112: This is completely wrong, since `global` isn't actually defined other than in Node.js and these unit-tests now fail in browsers with `ReferenceError: global is not defined ...`

Fixing that locally, however, *still* causes these tests to fail, now with `TypeError: canvas.getContext is not a function ...` in browsers.

When reviewing this, I simply assumed that the new unit-tests were actually *tested* and thus working, but looking at them again I'm no longer sure that they're doing the right thing in general unfortunately. (E.g. the `ownerDocument` probably *needs* to be a valid `HTMLDocument`-instance, rather than a regular object.)

It's strange that this didn't cause the bots to fail, but it's clear from the logs that these new tests didn't even run. Unfortunately I cannot see an *immediate* fix for the unit-tests, and just removing the unit-tests seem like the wrong approach.

@timvandermeij Can you please back out this patch, since this shouldn't have landed as-is.

timvandermeij: This patch is now reverted. It's indeed really not great that Jasmine doesn't even run the test and therefore the bots think all is fine. I'll try to find out why that happens since we want to get a failed test if that ever happens.

PR #12134, opened 2020-07-28
little design change of the pageNumber input in viewer.css

skillinator:
this is alittle design change of the pageNumber input

- by hovering over the pageNumber input, the border-color of the pageNumber input is now a dark blue


Snuffleupagus:
First of all, there's not really any motivation provided for *why* this kind of (very minor) UI change is even desirable/necessary.
Second of all, there's work under-way to update/modernize the entire UI of the default viewer (which would render this patch moot anyway).

All-in-all, this patch unfortunately doesn't seem like something that we'd want in the PDF.js default viewer.

skillinator:
Ok
thank you for your answer

PR #12135, opened 2020-07-28
[src/core/worker.js] Remove a useless Promise handler from the `pdfManagerReady` function

Snuffleupagus:
Looking carefully at this code, you'll notice that the `loadDocument` function has no less than *three* Promise handling functions. This obviously makes no sense, since a Promise can only have one resolve and one reject handler.

Hence the final `onFailure`-case is unreachable, which only serves to add confusion when reading the code. Note that this code has been re-factored more than once over the years, but it seems as if this may even have been incorrect already in PR #3310 (and no-one have noticed for seven years :-).

*Smaller diff with https://github.com/mozilla/pdf.js/pull/12135/files?diff=split&w=1*

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -371,25 +371,21 @@ class WorkerMessageHandler {
       function pdfManagerReady() {
         ensureNotTerminated();
 
-        loadDocument(false).then(
-          onSuccess,
-          function loadFailure(ex) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Ha, nice catch!

PR #12137, opened 2020-07-28
Add support for saving forms

calixteman:


calixteman:
@timvandermeij, @Snuffleupagus, @brendandahl, I'll add some tests tomorrow but it'd be nice to have a first feedback asap.

web/pdf_save_service.js
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { PDFSaveServiceFactory } from "./app.js";
+import { writeXRef } from "../src/core/writer.js";
```
Snuffleupagus, line 17: I'm *really* surprised that this even works, because it absolutely shouldn't. Most likely, you end up *indirectly* bundling a significant portion of the worker-thread code into the viewer, which you *absolutely* should not do!

Instead you'll need to split that file into main/worker-thread parts, with the code placed in the `src/display/` respectively the `src/core/` directories as needed. Note how functionality is exported in https://github.com/mozilla/pdf.js/blob/master/src/pdf.js and how the viewer accesses those through the `pdfjs-lib` alias.

src/core/writer.js
```diff
@@ -0,0 +1,164 @@
...
+ */
+/* eslint no-var: error */
+
+import { bytesToString, escapeString, stringToBytes } from "../shared/util.js";
+import { calculateMD5 } from "./crypto.js";
+import { Dict } from "./primitives.js";
+
+function writeDict(dict, buffer) {
+  buffer += "<<";
+  for (const [key, value] of Object.entries(dict._map)) {
```
Snuffleupagus, line 23: This may be moot with my other comment addressed (about a correct main/worker-thread split), but please note the `_map` property on a `Dict` instance should be considered *private* and consequently you *really* aren't supposed to be doing this!

src/core/worker.js
```diff
@@ -513,6 +515,45 @@ class WorkerMessageHandler {
...
     });
 
+    handler.on("ResetNewRef", function wphSetupResetNewRef() {
+      pdfManager.pdfDocument.xref.resetNewRef();
+    });
+
+    handler.on("GetInfoForNewXRef", function wphSetupGetInfoForNewXRef() {
+      const xref = pdfManager.pdfDocument.xref;
+      if (xref.trailer) {
+        const _info = {};
```
Snuffleupagus, line 525: *Nit (here and elsewhere):* In the PDF.js library we usually use `Object.create(null)`, rather than `{}`, to prevent any possible issues if a library user have e.g. extended the `Object.prototype` incorrectly.

src/core/writer.js
```diff
@@ -0,0 +1,164 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* eslint no-var: error */
+
+import { bytesToString, escapeString, stringToBytes } from "../shared/util.js";
+import { calculateMD5 } from "./crypto.js";
+import { Dict } from "./primitives.js";
```
Snuffleupagus, line 19: Note that these cannot be used on the main-thread, since that will end up pulling in *a lot* of the `src/core/` functionality into the *built* `pdf.js` file. (The referenced code should only appear in the *built* `pdf.worker.js` file.)

It may be helpful to look in the `gulpfile.js` or perhaps simply building the generic PDF.js library, using `gulp generic`, and look at the where various code ends up (with/without these patches applied).

Snuffleupagus (changes requested):
My general comment would be that you'll need to be *very careful* to not accidentally create a *corrupt* PDF document when editing it. (Given how much trouble corrupt PDF documents have caused us over the years, we really ought to avoid creating such ones ourselves.)

Furthermore, you'll probably want to test this not only with "simple" and/or well-formed documents, to make sure that things still work with e.g. PDF documents that use `XRefStm`s (i.e. cross-reference streams) and/or those where the XRef table is corrupt.

Also, I'm guessing that you may also want to hide the toolbar buttons (or at least have them be `disabled`), until editing has occurred!?

Finally, note that as I've remarked inline, you'll need to split the XRef writing part into proper main/worker-thread parts.

brendandahl:
Looking good overall, but as Jonas mentioned we need to split up the worker/main thread code more. Or we could move most everything to the worker. What about exposing a higher level API, something where the viewer would do:
1. Collect all the annotation storage data in a serializable format
2. Pass that data into a new function that lives in src/display/api.js on the `PDFDocumentProxy` (e.g. `save(annotationStorages)` which would in turn:
    1. Send it to the worker which would do all the xref updating
    2. Return updated PDF data, that would resolve the promise

calixteman:
> My general comment would be that you'll need to be _very careful_ to not accidentally create a _corrupt_ PDF document when editing it. (Given how much trouble corrupt PDF documents have caused us over the years, we really ought to avoid creating such ones ourselves.)
I don't want to write corrupted PDF!
I opened newly created PDF with evince successfully and I hope that you help me to find out errors if any.

> 
> Furthermore, you'll probably want to test this not only with "simple" and/or well-formed documents, to make sure that things still work with e.g. PDF documents that use `XRefStm`s (i.e. cross-reference streams) and/or those where the XRef table is corrupt.
> 
Do you know where I can find these kind of PDF docs with a form inside ?

> Also, I'm guessing that you may also want to hide the toolbar buttons (or at least have them be `disabled`), until editing has occurred!?
> 
Done in my last push.
For now I just added a smiley for the icon: I'm not the guy to draw a new icon.

> Finally, note that as I've remarked inline, you'll need to split the XRef writing part into proper main/worker-thread parts.
Addressed too.

Snuffleupagus (changes requested):
> Do you know where I can find these kind of PDF docs with a form inside ?

Besides scraping the test-suite for all PDF documents with forms (`grep`-ing for something like "/AcroForm" might work), I don't really know unfortunately.

calixteman:
This patch is ready for review.
It'd be nice to have a review for the beginning of next week, I'll address the issues you'll find and then we can merge it wednesday to have this patch in next firefox version.
@reviewers thx for your help.

web/pdf_save_service.js
```diff
@@ -0,0 +1,64 @@
...
+
+PDFSaveService.prototype = {
+  saveDocument() {
+    return this.pdfDocument.saveDocument(
+      this.pagesCount,
+      this.pdfDocument.annotationStorage.getAll()
+    );
+  },
+
+  get active() {
```
timvandermeij, line 33: I don't think this is ever called, so let's remove it in that case.

web/pdf_save_service.js
```diff
@@ -0,0 +1,64 @@
...
+  createSaveService(pdfDocument, pagesCount) {
+    if (activeSaveService) {
+      throw new Error("The save service is created and active.");
+    }
+    activeSaveService = new PDFSaveService(pdfDocument, pagesCount);
+    return activeSaveService;
+  },
+};
+
+export { PDFSaveService };
```
timvandermeij, line 64: I don't think this is/should be used elsewhere because the factory gives you an instance, so this should probably be removed.

timvandermeij (commented):
Phew, that was quite a large review. Despite the number of review comments, this looks really good code-wise (not tested yet)! I'm only not really sure about the XRef parts of the patch since I don't know too much about that to judge if the code is correct, and that must be correct because we don't want to make malformed documents. I would really appreciate a review from @Snuffleupagus and/or @brendandahl for that part of the patch.

src/core/annotation.js
```diff
@@ -953,6 +962,75 @@ class WidgetAnnotation extends Annotation {
...
+      const bbox = [
+        0,
+        0,
+        this.data.rect[2] - this.data.rect[0],
+        this.data.rect[3] - this.data.rect[1],
+      ];
+
+      const newRef = evaluator.xref.getNewRef();
+      const AP = new Dict(null);
+      AP.set("N", newRef);
```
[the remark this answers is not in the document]
calixteman: I agree with you... so I replaced the ref by a stream but the file then isn't correctly rendered in evince.
So I open a form on a mac with acrobat reader, modified it, saved it and then open it a text editor: the result is we've `<< /AP << /N 251 0 R >>...`.
I tried to find some info about annotation and incremental updates in the spec but I didn't manage to find something.
So I guess it's correct.

src/core/annotation.js
```diff
@@ -953,6 +962,75 @@ class WidgetAnnotation extends Annotation {
...
+      bufferOriginal.push("\nendobj\n");
+
+      const bufferNew = [`${newRef.num} ${newRef.gen} obj\n`];
+      writeDict(appearanceDict, bufferNew, transfNew);
+      bufferNew.push(" stream\n");
+      bufferNew.push(appearance);
+      bufferNew.push("\nendstream\nendobj\n");
+
+      return [
+        { ref: this.ref.num, data: bufferOriginal.join("") },
```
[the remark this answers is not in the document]
calixteman: We change the `V` field in the old one so we must rewrite it and the new one is just for the appearance.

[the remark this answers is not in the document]
timvandermeij: It would be great if that could be added as a comment just above the `return [` line so this becomes clear from the code.

src/core/annotation.js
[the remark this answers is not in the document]
timvandermeij: Hm, okay. I don't know too much about this, so let's ping @brendandahl and/or @Snuffleupagus to see if they perhaps know more.

src/core/annotation.js
```diff
@@ -1354,6 +1537,7 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     if (isDict(fieldParent) && fieldParent.has("V")) {
       const fieldParentValue = fieldParent.get("V");
       if (isName(fieldParentValue)) {
+        this.parentRef = params.dict.getRaw("Parent");
```
timvandermeij, line 1540: Given that `this.parentRef` might also be a `Dict` instead of a `Ref` that the name suggest, would `this.parent` perhaps be a better name?

src/core/writer.js
```diff
@@ -0,0 +1,221 @@
...
+}
+
+function writeStream(stream, buffer, transform) {
+  writeDict(stream.dict, buffer, transform);
+  buffer.push(" stream\n");
+  let string = bytesToString(stream.getBytes());
+  if (transform !== null) {
+    string = transform.encryptString(string);
+  }
+  buffer.push(string);
```
timvandermeij, line 37: I think we also need `escapeString` here. Better yet: why not call `writeValue` here since it will take care of that and the encryption as well so there is less code here.

timvandermeij (commented):
I only have two new comments left now that all changes have been made, and the following three comments from above are still open:

- https://github.com/mozilla/pdf.js/pull/12137#discussion_r468151546
- https://github.com/mozilla/pdf.js/pull/12137#discussion_r468151315
- https://github.com/mozilla/pdf.js/pull/12137#discussion_r467616648

Once these five comments are addressed, this looks good to me code-wise. It's only pending manual testing then, and any checks from @brendandahl and/or @Snuffleupagus in particular for the XRef part of the patch.

Really nice work so far!

brendandahl (approved):
r+ with the last comments from Tim. There's a few minor things I'd like to tweak/add, but they can be addressed in follow ups.

Snuffleupagus (changes requested):
Unfortunately I don't have nearly enough time to properly review the most complicated/important parts, which is the actual document/XRef writing.

However, I've left a couple of higher-level comments about API and/or default viewer issues that should be addressed; note e.g. that currently this is broken once you've actually *built* the viewer.

web/app.js
```diff
@@ -905,6 +905,41 @@ const PDFViewerApplication = {
...
+    const filename =
+      this.contentDispositionFilename || getPDFFileNameFromURL(this.url);
+    const downloadManager = this.downloadManager;
+    downloadManager.onerror = err => {
+      // This error won't really be helpful because it's likely the
+      // fallback won't work either (or is already open).
+      this.error(`PDF failed to be saved: ${err}`);
+    };
+
+    // If the pdf is not ready then save is equivalent to download
```
Snuffleupagus, line 921: Nit: Please copy the old comment *verbatim* from the `download` method above.

src/core/writer.js
calixteman: I think we don't need to escape the string when it's a stream.

src/core/annotation.js
```diff
@@ -953,6 +962,75 @@ class WidgetAnnotation extends Annotation {
...
+    bufferOriginal.push("\nendobj\n");
+
+    const bufferNew = [`${newRef.num} ${newRef.gen} obj\n`];
+    writeDict(appearanceDict, bufferNew, newTransform);
+    bufferNew.push(" stream\n");
+    bufferNew.push(appearance);
+    bufferNew.push("\nendstream\nendobj\n");
+
+    return [
+      { ref: this.ref.num, data: bufferOriginal.join("") },
```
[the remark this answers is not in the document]
calixteman: The ref here is just for the xref table so afaik we just the need to have the ref number but not the gen number.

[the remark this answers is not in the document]
Snuffleupagus: But then you *do* need to `gen` as well, since it's actually written into the XRef; please see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G6.1839814

src/core/worker.js
```diff
@@ -513,6 +516,67 @@ class WorkerMessageHandler {
...
 
+    handler.on("SaveDocument", function ({
+      numPages,
+      annotationStorage,
+      filename,
+    }) {
+      pdfManager.requestLoadedStream();
+      const promises = [pdfManager.onLoadedStream()];
+      const document = pdfManager.pdfDocument;
+      for (let pageIndex = 0, len = numPages; pageIndex < len; pageIndex++) {
```
Snuffleupagus, line 527: `for (let pageIndex = 0; pageIndex < numPages; pageIndex++) {`

web/app.js
```diff
@@ -240,6 +240,7 @@ const PDFViewerApplication = {
   async initialize(appConfig) {
     this.preferences = this.externalServices.createPreferences();
     this.appConfig = appConfig;
+    this._saveInProgress = false;
```
Snuffleupagus, line 243: Please put the definition after the following line instead: https://github.com/mozilla/pdf.js/blob/7fb01f9f2a4d3d6e2680cf9f6df3c83446d6e763/web/app.js#L237
and also remember to reset it after this line as well: https://github.com/mozilla/pdf.js/blob/7fb01f9f2a4d3d6e2680cf9f6df3c83446d6e763/web/app.js#L733

web/app.js
```diff
@@ -905,6 +906,41 @@ const PDFViewerApplication = {
...
+    }
+
+    this._saveInProgress = true;
+    this.pdfDocument
+      .saveDocument(this.pdfDocument.annotationStorage)
+      .then(data => {
+        const blob = new Blob([data], { type: "application/pdf" });
+        downloadManager.download(blob, url, filename);
+      })
+      .catch(this.download); // Error occurred, try downloading with the URL.
```
Snuffleupagus, line 940: Please check that this works, since the scopes probably won't be correct here. Most likely, you need to either do:
```javascript
.catch(() => {
  this.download();
});
```
or
```javascript
.catch(this.download.bind(this));
```

web/app.js
```diff
@@ -2883,8 +2923,18 @@ const PDFPrintServiceFactory = {
...
 
+/* Abstract factory for the save service. */
+const PDFSaveServiceFactory = {
+  instance: {
+    createSaveService() {
+      throw new Error("Not implemented: createSaveService");
+    },
+  },
+};
+
```
Snuffleupagus, line 2934: Please remove this, since it's now unused.

web/viewer.js
```diff
@@ -193,7 +193,13 @@ function webViewerLoad() {
...
       import("pdfjs-web/genericcom.js"),
       import("pdfjs-web/pdf_print_service.js"),
-    ]).then(function ([app, appOptions, genericCom, pdfPrintService]) {
+    ]).then(function ([
+      app,
+      appOptions,
+      genericCom,
+      pdfPrintService,
+      pdfSaveService,
+    ]) {
```
Snuffleupagus, line 202: Please revert this change, since it's now unused.

web/app.js
```diff
@@ -1674,7 +1709,12 @@ const PDFViewerApplication = {
     eventBus._on("presentationmodechanged", webViewerPresentationModeChanged);
     eventBus._on("presentationmode", webViewerPresentationMode);
     eventBus._on("print", webViewerPrint);
-    eventBus._on("download", webViewerDownload);
+    eventBus._on(
+      "download",
+      AppOptions.get("renderInteractiveForms")
+        ? webViewerSave
+        : webViewerDownload
+    );
```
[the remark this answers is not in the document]
timvandermeij: #12192 is now merged, so the size check can now be added.

timvandermeij (approved):
Code-wise this looks good to me when the final comments are addressed, namely:

- https://github.com/mozilla/pdf.js/pull/12137#discussion_r468363081 (the size check should be added now that the other PR is merged)
- https://github.com/mozilla/pdf.js/pull/12137#discussion_r468459391 (this line should be removed)

Functionality-wise Brendan will check the XRef part, do some manual tests and merge it since I'll be away for a few days. Indeed, any other points can be picked up in a follow-up.

You've clearly put a lot of work into this, and it shows. Thank you for your work!

PR #12139, opened 2020-07-29
WIP - Parse acroform dictionary

escapewindow:
https://github.com/mozilla/pdf.js/pull/8625#issuecomment-660481227 - picking up changes from #8625.

To address:

- [X] unbitrot - @dmitryskey largely fixed this.
- [ ] firefox-issue7115-page1 - missing green border
- [ ] firefox-bug1552113-page1 - missing black border
- [ ] firefox-f1040-forms-page2 - pin border grey not black
- [ ] firefox-link-annotation-border-page1 - missing green borders
- [ ] firefox-annotation-tx-forms-page1 - font size larger
- [ ] firefox-annotation-tx2-forms-page1 - has green border, darker blue background, larger font
- [ ] firefox-annotation-tx3-forms-page1 - has green border, darker blue background, larger font
- [ ] firefox-annotation-border-styles.pdf-page1 - missing borders
- [X] firefox-annotation-text-widget-forms-page1 - improved borders!
- [ ] address [performance concerns](https://github.com/mozilla/pdf.js/pull/8625#issuecomment-660481227) around using a new worker for additional parsing


escapewindow:
I took a stab at refactoring how/when we parse the acroform dictionary [here](https://github.com/mozilla/pdf.js/compare/master...escapewindow:acrodict-scratch?expand=1). Not 100% sure about it yet.

escapewindow:
Looking at this diff, I don't see us calling [`setDefaultAppearance`](https://github.com/mozilla/pdf.js/pull/12139/files#diff-18ca06e2bd5be4c9132b51f78388b8f0R959) anywhere.

dmitryskey:
I'm not sure for this PR, in my branch `_create` function has the `switch` like following. After initializing by default the `WidgetAnnotation` I set up its default properties.

```javascript
let widget;

switch (fieldType) {
  case 'Tx':
      widget = new TextWidgetAnnotation(parameters);
      break;
  case 'Btn':
      widget = new ButtonWidgetAnnotation(parameters);
      break;
  case 'Ch':
      widget = new ChoiceWidgetAnnotation(parameters);
      break;
  default:
      widget = new WidgetAnnotation(parameters);
      warn('Unimplemented widget field type "' + fieldType + '", ' +
          'falling back to base field type.');
      break;
  }

  return widget.setDefaultAppearance(parameters);
```

src/core/worker.js
```diff
@@ -512,11 +512,16 @@ class WorkerMessageHandler {
...
       "GetOperatorList",
       function wphSetupRenderPage(data, sink) {
-        var pageIndex = data.pageIndex;
+        const pageIndex = data.pageIndex;
         pdfManager.getPage(pageIndex).then(function (page) {
-          var task = new WorkerTask(`GetOperatorList: page ${pageIndex}`);
+          const task = new WorkerTask(`GetOperatorList: page ${pageIndex}`);
           startWorkerTask(task);
 
+          const annotationTask = new WorkerTask(
```
dmitryskey, line 520: I'd say this was the most critical change, as you can see I had to create another task for the parsing operator list under annotations. Obviously this brings performance concerns.

escapewindow:
> I'm not sure for this PR, in my branch `_create` function has the `switch` like following.

Which branch? In your `master` branch, you have [this block](https://github.com/dmitryskey/pdf.js/blob/master/src/core/annotation.js#L97-L115), if I'm looking at the right file.

Added this change; I'll check to see how that changes its behavior.

src/core/worker.js
escapewindow: On my list. I'm wondering if we can parse this in `Page` parsing currently.

escapewindow:
It looks like the parameters don't have an evaluator.

```
1) annotation ChoiceWidgetAnnotation should handle array field values

  Message:

    Failed: Cannot read property 'getOperatorList' of undefined

  Stack:

    TypeError: Cannot read property 'getOperatorList' of undefined

        at ChoiceWidgetAnnotation.setDefaultAppearance (/home/travis/build/mozilla/pdf.js/build/lib-es5/core/annotation.js:862:31)
        at Function._create (/home/travis/build/mozilla/pdf.js/build/lib-es5/core/annotation.js:154:25)
        at /home/travis/build/mozilla/pdf.js/build/lib-es5/test/unit/annotation_spec.js:76:27
        at <Jasmine>
        at PDFManagerMock.ensure (/home/travis/build/mozilla/pdf.js/build/lib-es5/test/unit/annotation_spec.js:72:16)
        at /home/travis/build/mozilla/pdf.js/build/lib-es5/core/annotation.js:95:27
```

escapewindow:
Ah, this is likely due to the test driver not containing `evaluator` in its parameters. Not sure if we should fix this, only to remove it when we change when we parse the dictionary, or if we should move the parsing first.

escapewindow:
Wondering if https://github.com/mozilla/pdf.js/blob/5a66c56eca31a924a755112a393ecd2621fb25d5/src/core/annotation.js#L809-L812 is sufficient, and we only need the switch.

escapewindow:
Attempted to use an [evaluator](https://github.com/mozilla/pdf.js/compare/master...escapewindow:evaluator-acrodict?expand=1). Not sure how to use this now, though; should I be saving an operator list? passing `data` down?.

escapewindow:
[This](https://github.com/mozilla/pdf.js/compare/master...escapewindow:defaultAppearance?expand=1) is a working patch that parses the DA as a string. Still stuck passing it down -- digging into the web/ rendering steps.

escapewindow:
Working patch that doesn't create a new task and doesn't break reftests is in https://github.com/mozilla/pdf.js/pull/12207 . It doesn't contain all the changes here, but hopefully has enough to support them.

escapewindow:
Going to close in favor of https://github.com/mozilla/pdf.js/pull/12207 . If/when that lands, we can adjust the remaining pieces.

Issue #9831, opened 2018-06-21
search parameter viewer.html doesn't open automatically findbar with search string in text field


van-nes:
A pdf.js viewer.html call with (e.g.) search parameter:
    www.dezijpe.nl/pdfjs/web/viewer.html?file=/zhb/pdf/2007030311.pdf#search=abc&phrase=true
works fine, BUT doesn't open automatically the findbar.

Only after using manually the magnifier icon (findbar.title), the findbar will be opened, however
with empty search field (missing the search string from: #search=abc).
Only the number of hits is correctly shown.

How exactly (implementation in viewer.js ?) to:
1. automatically open findbar (PDFViewerApplication.findBar.open(); ?)
2. automatically include the search text-string from param #serach =

Thanks for help.

erzaehlsalex:
@van-nes Did you create a PR for that? I think this is a really good feature!
@timvandermeij This is a really small feature that could be merged easily, right? :)

Snuffleupagus:
This is essentially a duplicate of issue #8185, and should be closed as such.

> This is a really small feature that could be merged easily, right? :)

Unfortunately not, since that completely ignores the Firefox integration; please refer to https://github.com/mozilla/pdf.js/issues/8185#issuecomment-288717592 and the discussion in #3565.

Snuffleupagus:
@timvandermeij Mind closing this one, since it's a duplicate of issue #8185?

Issue #8185, opened 2017-03-22
Search Widget occurence navigation not working when using "search" hash parameter.


devanonyme:
Link to PDF file (or attach file here):
Can be reproduced using the demo / sample : https://mozilla.github.io/pdf.js/web/viewer.html#search=language

Configuration:
- Web browser and its version:
any
- Operating system and its version:
any
- PDF.js version:
Found in latest stable (v1.6.210), reproducible in online demo (1.7.378)
- Is an extension:

Steps to reproduce the problem:
1. Use a url with search param : https://mozilla.github.io/pdf.js/web/viewer.html#search=language
2. Open the search widget (with the intention of using "next occurence" by exemple).

What is the expected behavior? (add screenshot)
- Clicking on "next occurence" should be working.
- The current active search term should be visible instead of place holder "Find in document...".

What went wrong? (add screenshot)
- Previous / next occurence are not working.
- Current search term is not visible.

![screen shot 03-22-17 at 02 14 pm](https://cloud.githubusercontent.com/assets/21067311/24214133/036bac3c-0f0b-11e7-82d2-5ab93c37cbba.PNG)

devanonyme:
It's probably not the clean solution (should be done within findfromurlhash handler?) but I have patched our local version with this:

```
    setHash: function PDFLinkService_setHash(hash) {
      var pageNumber, dest;
      if (hash.indexOf('=') >= 0) {
        var params = parseQueryString(hash);
        if ('search' in params) {

          **// If search param exist. Set it in Search bar find field.
          // So that the user can use next occurent as expected.
          var searchText = params['search'].replace(/"/g, '');
          if (searchText) {
            PDFViewerApplication.findBar.findField.value = searchText;
          }**

          this.eventBus.dispatch('findfromurlhash', {
            source: this,
            query: **searchText**,
            ...
```

yurydelendik:
See also https://github.com/mozilla/pdf.js/pull/5579#issuecomment-220661497

Issue #3565, opened 2013-08-12
Named action for search not working in addon version


timvandermeij:
http://www.informatik.uni-freiburg.de/~frank/ENG/beamer/example/Beamer-class-example1.pdf

Was caused during review at https://github.com/mozilla/pdf.js/issues/3057#discussion_r4502073. Now, in the addon, nothing happens when clicking the magnifying glass, only in the github.io version it's working. We should always use the Firefox find bar where possible, and otherwise the PDF.js find bar. We can use the method from https://github.com/mozilla/pdf.js/pull/2485/files#diff-0 for this purpose.

@yurydelendik Since you did the review for the named actions PR, what do you think about this?

timvandermeij:
Closing since in all these years this hasn't been brought up anymore and there are indeed security/usability requirements to consider for this.

PR #12141, opened 2020-07-30
Populate the find field with the search query when URL has #search hash

phillipj:
These changes improves the existing search functionality triggered when the URL contains a `#search` hash.

In addition to performing the actual search immediately like before, this will ensure the search text field inside the find bar gets populated with the text currently being searched for.

**Picture says more than a thousand words;** the goal is for `tesla` being populated in the search field below because pdf.js was loaded with `#search=tesla` in the URL:

Screenshot 2020-07-30 at 10 09 01

Does this look like a viable approach to you?

_Kudos to @marill for finding this workaround._

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2349,6 +2349,10 @@ function webViewerFindFromUrlHash(evt) {
     highlightAll: true,
     findPrevious: false,
   });
+
+  if (PDFViewerApplication.findBar) {
+    PDFViewerApplication.findBar.findField.value = evt.query;
+  }
 }
 
 function webViewerUpdateFindMatchesCount({ matchesCount }) {


timvandermeij (approved):

timvandermeij:
Looks good to me. Thank you!

phillipj:
Appreciate the quick response!

Just noticed what gives an impression `findBar.findField` could be `null` at times... 

https://github.com/mozilla/pdf.js/blob/c5663f2f6b95680c9b7ce0c0f1e0021aa22deebe/web/pdf_find_bar.js#L33

Would that ever happen in practise when `app.js` is executed? If so, I'll happily open a follow-up PR that ensures these changes does not cause `TypeError: Cannot set property 'value' of null` going forward.

Snuffleupagus:
Sorry, but why was this patch accepted when it introduces *inconsistent* behaviour depending on `supportsIntegratedFind` and thus indirectly the build target (e.g. `GENERIC` vs. `MOZCENTRAL`)?

Previously we've rejected these kind of changes, because of the inconsistency they would bring, especially considering that this (for security reason) cannot be easily implemented in the Firefox PDF viewer; please note prior discussion in e.g. issues #3565, #8185, and #9831.

phillipj:
@Snuffleupagus I wasn't aware of those concerns, thanks for bringing that up.

Sounds like these changes as is should be reverted from `master`.

Do you have any advice for me to work on improvements that addresses these inconsistencies? Alternatively your recommended approach for doing what was intended with these changes?

Snuffleupagus:
> Sounds like these changes as is should be reverted from `master`.

This obviously isn't a huge and/or breaking change, so I don't know what the consensus will regarding a back-out.
(It might simply have been good to have the discussion *beforehand*, but mistaken can happen since we're all human :-)

Hence I suppose that this *could* stay, as far I'm concerned, if and only if:
 - It's agreed upon that this small behaviour inconsistency is OK, in this case.
 - We don't treat this as a precedent for introducing additional/larger inconsistencies, for different build targets, for the find functionality.

> Do you have any advice for me to work on improvements that addresses these inconsistencies?

I don't think there's a good/secure way of handling this generally, in the Firefox *browser* findbar as well, since https://github.com/mozilla/pdf.js/issues/3565#issuecomment-22566903 still very much applies as far as I can tell.

phillipj:
As I obviously can't comment on if's your raising, I can at least share the use case we've got in the project I'm working on -- consider or ignore as you please.

Our end-users finds PDF documents either by navigating to them or as part of the product's search results.

The off the shelf Viewer is embedded via an `<iframe>` and we set `#search=` accordingly when we want PDF.js to go into "search mode" immediately and focus the first occurrence.

Without the changes in this PR, the user experience is not totally intuitive though. End-users obviously appreciate the convenience of text matches being highlighted right away. But it's confusing when they open the find bar and the text field is empty -- although it is has obviously searched for something seeing text is highlighted (like `tesla` in the PR description example).

<hr>

Inspired by the [Third party viewer usage wiki](https://github.com/mozilla/pdf.js/wiki/Third-party-viewer-usage) I've found another workaround which isn't too bad. Though from a PDF.js rookie and using-project's perspective, it is still awkward that code needs to be written fill in the find text field 🤷‍♂️ 

```js
document.addEventListener("webviewerloaded", function(event) {
  const PDFViewerApplication = event.detail.source.PDFViewerApplication;

    PDFViewerApplication.findBar.findField.value = "tesla"; // same value as provided in URL's #search hash
  });
});
```

Would you be more positive to these changes if they were located in [`./web/viewer.js`](https://github.com/mozilla/pdf.js/blob/master/web/viewer.js) instead of [./web/app.js](https://github.com/mozilla/pdf.js/blob/master/web/app.js)? Or would that still affect the Firefox browser as you mentioned?

Snuffleupagus:
> Would you be more positive to these changes if they were located in [`./web/viewer.js`](https://github.com/mozilla/pdf.js/blob/master/web/viewer.js) instead of [./web/app.js](https://github.com/mozilla/pdf.js/blob/master/web/app.js)?

That would be *a lot* worse, in every way, given that `web/viewer.js` is used only for configuration/loading purposes :-)

There's only one decision to be make here: Either we let these changes stay as-is (in this case), or we revert them wholesale (and reject any future attempt at changing this).

> Or would that still affect the Firefox browser as you mentioned?

Sorry, I really don't understand this question.
The point here is merely that the changes made in this PR unfortunately cannot be applied to the Firefox PDF viewer (which is the reason for this library existing in the first place), since it uses the native *browser* findbar, for the reasons mention at the bottom of https://github.com/mozilla/pdf.js/pull/12141#issuecomment-667025548.

timvandermeij:
I forgot about the integrated findbar here; sorry for that. I do think that the change makes sense because having a search query enabled yet not seeing it in the findbar is indeed inconsistent.

However you have a valid point about us rejecting such changes before, so making an exception in this case would also be inconsistent.

I'll make a PR to revert this change for now and open an issue where we can discuss how to move forward with the findbar since I do feel that the lack of customizability of the integrated one is holding us back sometimes as evidenced by the various PRs that were made before. I personally don't mind them being a bit inconsistent because they are never used together, but I think this is a discussion for outside of this PR.

Snuffleupagus:
> I do think that the change makes sense because having a search query enabled yet not seeing it in the findbar is indeed inconsistent.

Completely agreed; I've never questioned the utility of the feature itself, just how it's implemented (or not) across various build targets.

> I personally don't mind them being a bit inconsistent because they are never used together,

Agreed, and in *this* case the difference is small enough to probably not matter much...

---

It should be possible, without too much trouble, to support pre-populating the Firefox findbar only when `#search=term` is used on *initial* document loading. However, that'd likely be *even* worse than the current situation of never updating the Firefox findbar at all.

timvandermeij:
I have reverted the change in #12148 pending the discussion in #12149. Once a decision is made there, this can either simply be merged again or it won't be introduced anymore.

phillipj:
Sounds good to have a dedicated discussion settling the overall policy for find bar changes and behaviour going forward :+1:

I won't be participating in the discussion as my PDF.js incompetence is obvious. Feel free to ping me if you want input from an using project' point of view.

burotica49:
It is possible to do it this way

```
<script>
    document.addEventListener("webviewerloaded", function(event) {
      const queryString = window.location.hash;
      const search = queryString.replace('#search=', '');
      const PDFViewerApplication = event.detail.source.PDFViewerApplication;

        PDFViewerApplication.findBar.findField.value = search; // same value as provided in URL's #search hash
      });
    });
  </script>
```

PR #12143, opened 2020-07-30
Fix hanging reference test (PR 12107 follow-up)

escapewindow:
The f1040-annotations reftest started hanging after #12107. We traced
this to `TypeError: can't access property "getOrCreateValue", storage is
undefined`.

We essentially need to add `annotationStorage` to the parameters in
test/driver.js.


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

Issue #11996, opened 2020-06-12
Replace the `webpack-stream` dependency


Snuffleupagus:
With PR #11436 we became locked at a *particular* (older) version of the [`webpack-stream` dependency](https://www.npmjs.com/package/webpack-stream), because of bugs in the latest versions; see https://github.com/mozilla/pdf.js/pull/11436#issuecomment-568430643 for details.

Looking at the [GitHub-page for `webpack-stream`](https://github.com/shama/webpack-stream), it unfortunately seems that the project may be unmaintained/dead at this point given that the latest commit was in 2018.
Hence, rather than trying to update `webpack-stream` it's probably better to attempt to replace it completely instead (possibly with our own code if necessary) to ensure that we'll be able to e.g. update to Webpack 5 when it's released (which will likely be needed to *properly* support new ECMAScript features).

timvandermeij:
This dependency is now updated to the most recent version in #12397. However, given that's it's a very inactively maintained dependency I also think we should get rid of it. Moreover, we only seem to be using it in one place in the Gulpfile...

timvandermeij:
Now that we're using Webpack 5, the dependency got updated and there was more development on their GitHub page, do we still want to do this?

Snuffleupagus:
> Now that we're using Webpack 5

Given that the Webpack upgrade *itself* wasn't actually blocked by the `webpack-stream` dependency, as I incorrectly assumed previously, I suppose this reduces the severity of this issue considerably.

> the dependency got updated and there was more development on their GitHub page, do we still want to do this?

To me, this all depends on whether or not they'll try to re-introduce the broken caching mechanism (which originally forced us to pin the version number).

> I guess we could make https://github.com/mozilla/pdf.js/blob/master/package.json#L61 use `^` instead of `~` now.

I suppose that could work, as long as we remember to check `webpack-stream` for any possible breakage when updating the npm packages (which I've been doing the last few times).
If you want to submit a PR with this change, and then simply close this issue, then please be my guest :-)

PR #12144, opened 2020-07-31
Prevent `Uncaught (in promise) AbortException` when running the unit-tests

Snuffleupagus:
These errors can/will occur if data is still loading when the document is destroyed, which is the case in the API unit-tests that load the `tracemonkey.pdf` file.
While this patch prevents these kind of problems, and thus allows us to update Jasmine again, I cannot help but thinking that it's slightly "hacky". Basically, we'll simply catch and ignore (some) rejected promises once the document is destroyed and/or its data loading is aborted. However, I don't *think* that these changes should cause issues in general, since we don't really care about errors once document destruction has started (note e.g. the fair number of `catch` handlers ignoring `AbortException`s already).

### DIFF
--- a/src/core/chunked_stream.js
+++ b/src/core/chunked_stream.js
@@ -410,18 +410,24 @@ class ChunkedStreamManager {
       requestIds.push(requestId);
     }
 
-    if (!chunksToRequest.length) {
-      return capability.promise;
+    if (chunksToRequest.length > 0) {
+      const groupedChunksToRequest = this.groupChunks(chunksToRequest);
+      for (const groupedChunk of groupedChunksToRequest) {
+        const begin = groupedChunk.beginChunk * this.chunkSize;
+        const end = Math.min(
+          groupedChunk.endChunk * this.chunkSize,
+          this.length
+        );
+        this.sendRequest(begin, end);
+      }
     }
 
-    const groupedChunksToRequest = this.groupChunks(chunksToRequest);
-    for (const groupedChunk of groupedChunksToRequest) {
-      const begin = groupedChunk.beginChunk * this.chunkSize;
-      const end = Math.min(groupedChunk.endChunk * this.chunkSize, this.length);
-      this.sendRequest(begin, end);
-    }
-
-    return capability.promise;
+    return capability.promise.catch(reason => {
+      if (this.aborted) {
+        return; // Ignoring any pending requests after abort.
+      }
+      throw reason;
+    });
   }
 
   getStream() {
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2048,6 +2048,13 @@ class WorkerTransport {
 
       sink.onCancel = reason => {
         this._fullReader.cancel(reason);
+
+        sink.ready.catch(readyReason => {
+          if (this.destroyed) {
+            return; // Ignore any pending requests if the worker was terminated.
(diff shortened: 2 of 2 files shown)


src/display/api.js
```diff
@@ -2127,6 +2134,13 @@ class WorkerTransport {
 
       sink.onCancel = reason => {
         rangeReader.cancel(reason);
+
+        sink.ready.catch(readyReason => {
+          if (this.destroyed) {
+            return; // Ignore any pending requests if the worker was terminated.
+          }
+          throw readyReason;
+        });
```
Snuffleupagus, line 2143: Compared to the `fullReader` case above, I wasn't actually able to trigger this code-path in any of the unit-tests. However, given the (fairly) small amount of code, the fact that this is consistent with the above, and to prevent any future issues I figured that this cannot hurt.

Given the (relatively) small size of a range request, as compared to the full request, triggering this is probably a lot more difficult and timing-dependent. E.g. it *may* still happen for a slow connection and/or a large `rangeChunkSize` option.

src/core/chunked_stream.js
```diff
@@ -410,18 +410,24 @@ class ChunkedStreamManager {
...
+      }
     }
 
-    return capability.promise;
+    return capability.promise.catch(reason => {
+      if (this.aborted) {
+        return; // Ignoring any pending requests after abort.
+      }
+      throw reason;
+    });
```
Snuffleupagus, line 430: These changes weren't necessary to get the unit-tests to run *as-is* with an updated Jasmine version, however with `disableStream = true` set they still failed with `Uncaught (in promise) AbortException` which thus necessitated these changes.

Given that we're already ignoring read data, in `sendRequest`, this is to some extent thus "consistent" behaviour.

Snuffleupagus:
Unfortunately there's apparently *one* remaining unit-test failure, which is limited to Google Chrome on the Linux bot:
```
TEST-UNEXPECTED-FAIL | creates pdf doc from non-existent URL | in chrome | Unhandled promise rejection: MissingPDFException: Missing PDF "http://127.0.0.1:34548/test/pdfs/non-existent.pdf". 
```
Perhaps unsurprisingly, I cannot reproduce this locally in Chrome on Windows since `gulp unittest` passes without issues. Could we still *possibly* proceed with this patch as-is, given that it improves things overall[1], since I'm not sure when (or even if) I'll be able to fix the remaining failure?

---
[1] Unfortunately, even the patch as-is took me *hours* of debugging/testing, spread out over a couple of days, to arrive at.

timvandermeij:
> Could we still possibly proceed with this patch [...]

I would say that these changes make sense if the Jasmine update itself is left out of this PR until the last failure is fixed.

Snuffleupagus:
> I would say that these changes make sense if the Jasmine update itself is left out of this PR until the last failure is fixed.

Done; although I do find it to be quite unfortunate having to do so, since I worry that we'll "never" actually get the final test-failure fixed and thus be stuck at an old Jasmine-version indefinitely. (See e.g. issue #11996 which tracks another, fairly old, dependency problem that may sooner of later become a real problem.)

/botio unittest

timvandermeij (approved):

timvandermeij:
Looks good to me; another step closer to updating Jasmine in the end.

Issue #12152, opened 2020-08-01
Fix pull request #12131 so it can be merged again


timvandermeij:
It looks like the only problem that required us to do a revert of the patch is the fact that the unit tests in #12131 don't pass locally. However, they don't even run at all on the bots, hence we thought this was safe to merge given that all tests passed.

@timvandermeij Investigate why the bots did not run the test so this won't happen again.
@jsg2021 Could you improve the unit test so that we can try merging the patch again? Note the comment on the PR about what fails. In particular, we need to make sure that the tests work locally too; you can do that by running `gulp unittest` to test in the browser and `gulp unittestcli` to test in Node.js. Both should pass in order for the patch to be merged again. Thanks!

jsg2021:
yep, will do. 

shouldn't `npm test` run both those? I was using that to run and validate locally.

timvandermeij:
Ah, now I understand. `npm test` runs the unit tests in Node.js and it was made this way so that Travis CI has a direct entrypoint for running the tests. The `gulp unittest` command instead runs the unit tests in the browser instead. There is a difference because we cannot run the same tests in the browser and in Node.js due to platform differences.

timvandermeij:
In short, the `npm` commands should not be used for development. Only the `gulp` tasks are valid for that. See https://github.com/mozilla/pdf.js/wiki/Contributing for more information.

timvandermeij:
I did find the root cause for the fact that the test failure was not reported on the bots, so I'll make a patch for that so that such problems don't go unnoticed again.

jsg2021:
@timvandermeij I've fixed the tests and pushed the branch. Should I issue a new pull request?

timvandermeij:
Yes, indeed. We have just merged the fix for the test reporter so that such problems don't go unnoticed anymore in the future, so please rebase onto the current master branch and create a new PR. Thank you!

timvandermeij:
Closing since the test reporter issue is fixed and there is a new PR for this now.

Issue #12123, opened 2020-07-26
Fix intermittent unit/reference test failures


timvandermeij:
To debug this, pull request #12124 included the browser name in the logs for the unit tests too.

The following is a list of unit test failures we have seen:

- [x] `TEST-UNEXPECTED-FAIL | cleans up document resources during rendering of page | in firefox | Unhandled promise rejection: Error: shall fail cleanup in http://127.0.0.1:38165/node_modules/systemjs/dist/system.js line 4 > eval (line 1585)` (Linux; fixed in #12126)
- [ ] `TEST-UNEXPECTED-FAIL | creates pdf doc from non-existent URL | in chrome | Unhandled promise rejection: MissingPDFException: Missing PDF "http://127.0.0.1:38165/test/pdfs/non-existent.pdf".` (Linux)
- [x] `TEST-UNEXPECTED-FAIL | gets default page mode | in chrome | Unhandled promise rejection: AbortException: Worker was terminated.` (Linux; fixed in #12144/#12136)
- [x] `TEST-UNEXPECTED-FAIL | gets non-existent outline | in chrome | Unhandled promise rejection: AbortException: Worker was terminated.` (Linux; fixed in #12144/#12136)
- [x] `TEST-UNEXPECTED-FAIL | gets operatorList with JPEG image (issue 4888) | in chrome | Unhandled promise rejection: AbortError: The user aborted a request.` (Windows; fixed in #12144/#12136)
- [x] `TEST-UNEXPECTED-FAIL | should correctly render PDFs in parallel | in firefox | Unhandled promise rejection: AbortError: The operation was aborted.  (line 369)` (Windows; fixed in #12144/#12136)
- [x] `TEST-UNEXPECTED-FAIL | multiple render() on the same canvas | in firefox | Failed: shall fail rendering` (Windows; fixed in #12171)

It's clear from this overview that most problems are related to unexpected action abortions, both in Chrome/Firefox and on Windows/Linux. This wasn't happening a few days ago, so maybe something regressed? The second one is also interesting since we explicitly catch the exception in the unit test.

Pull request #12125 tries to improve the situation for the reference test failures.

timvandermeij:
Improvements made for this:

- #12124 (merged)
- #12126 (merged)
- #12125 (in progress)

timvandermeij:
[...]

PR #12153, opened 2020-08-01
Unit test improvements

timvandermeij:
This commit series fixes the part of #12152 that caused the failure to go unnoticed and fixes the part of #12123 about updating Jasmine.

For the latter, I have run the unit tests locally around 100 times and the one remaining intermittent failure never happened once for me. Having thought about this more, I now also believe that it's no longer a blocker for the Jasmine update now that the other intermittent failures have actually been resolved (they happened far more often). We'll keep the issue open to track the last intermittent failure, but it's no longer a blocker for us.

### DIFF
--- a/test/unit/testreporter.js
+++ b/test/unit/testreporter.js
@@ -51,7 +51,21 @@ var TestReporter = function (browser) {
     sendInfo("Started tests for " + browser + ".");
   };
 
-  this.suiteStarted = function (result) {};
+  this.suiteStarted = function (result) {
+    // Normally suite starts don't have to be reported because the individual
+    // specs inside them are reported, but it can happen that the suite cannot
+    // start, for instance due to an uncaught exception in `beforeEach`. This
+    // is problematic because the specs inside the suite will never be found
+    // and run, so if we don't report the suite start failure here it would be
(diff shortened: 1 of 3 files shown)


timvandermeij:
I'm going to do a unit test run that _includes_ the original commit with the failure that went unnoticed before to show that the unit tests now actually fail because of that. I'll remove that commit afterwards.

/botio unittest

test/unit/testreporter.js
```diff
@@ -48,21 +48,37 @@ var TestReporter = function (browser) {
 
   this.jasmineStarted = function (suiteInfo) {
     this.runnerStartTime = this.now();
-    sendInfo("Started tests for " + browser + ".");
+
+    const total = suiteInfo.totalSpecsDefined;
+    const seed = suiteInfo.order.seed;
+    sendInfo(`Started ${total} tests for ${browser} with seed ${seed}.`);
```
Snuffleupagus, line 54: This is *really* nice, and one of those things that you can't help but wonder why we haven't done years ago!

Snuffleupagus (approved):
Looks good to me, thanks for fixing this!

timvandermeij:
The alternate document commit is removed again.

/botio unittest

PR #12154, opened 2020-08-01
[api-minor] Allow loading pdf fonts into another document.

### LINKED ISSUE #12152: Fix pull request #12131 so it can be merged again

jsg2021:
Addresses PR #12131 browser test failures.
Fixes #12152

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -163,6 +163,9 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *   parsed font data from the worker-thread. This may be useful for debugging
  *   purposes (and backwards compatibility), but note that it will lead to
  *   increased memory usage. The default value is `false`.
+ * @property {HTMLDocument} [ownerDocument] - Specify an explicit document
+ *   context to create elements with and to load resources, such as fonts,
+ *   into. Defaults to the current document.
  * @property {boolean} [disableRange] - Disable range request loading of PDF
  *   files. When enabled, and if the server supports partial content requests,
  *   then the PDF will be fetched in chunks. The default value is `false`.
@@ -281,6 +284,9 @@ function getDocument(src) {
   }
   if (typeof params.disableFontFace !== "boolean") {
     params.disableFontFace = apiCompatibilityParams.disableFontFace || false;
+  }
+  if (typeof params.ownerDocument === "undefined") {
+    params.ownerDocument = globalThis.document;
   }
 
   if (typeof params.disableRange !== "boolean") {
@@ -976,9 +982,10 @@ class PDFDocumentProxy {
  * Proxy to a `PDFPage` in the worker thread.
  */
 class PDFPageProxy {
-  constructor(pageIndex, pageInfo, transport, pdfBug = false) {
+  constructor(pageIndex, pageInfo, transport, ownerDocument, pdfBug = false) {
     this._pageIndex = pageIndex;
     this._pageInfo = pageInfo;
+    this._ownerDocument = ownerDocument;
     this._transport = transport;
     this._stats = pdfBug ? new StatTimer() : null;
     this._pdfBug = pdfBug;
@@ -1111,7 +1118,9 @@ class PDFPageProxy {
       intentState.streamReaderCancelTimeout = null;
     }
 
-    const canvasFactoryInstance = canvasFactory || new DefaultCanvasFactory();
+    const canvasFactoryInstance =
+      canvasFactory ||
+      new DefaultCanvasFactory({ ownerDocument: this._ownerDocument });
     const webGLContext = new WebGLContext({
       enable: enableWebGL,
     });
@@ -2028,6 +2037,7 @@ class WorkerTransport {
     this.fontLoader = new FontLoader({
       docId: loadingTask.docId,
       onUnsupportedFeature: this._onUnsupportedFeature.bind(this),
+      ownerDocument: params.ownerDocument,
     });
     this._params = params;
     this.CMapReaderFactory = new params.CMapReaderFactory({
@@ -2484,6 +2494,7 @@ class WorkerTransport {
           pageIndex,
           pageInfo,
           this,
+          this._params.ownerDocument,
           this._params.pdfBug
         );
         this.pageCache[pageIndex] = page;
--- a/test/unit/custom_spec.js
+++ b/test/unit/custom_spec.js
@@ -107,3 +107,127 @@ describe("custom canvas rendering", function () {
       .catch(done.fail);
   });
 });
+
+describe("custom ownerDocument", function () {
+  const FontFace = globalThis.FontFace;
+
+  const checkFont = font => /g_d\d+_f1/.test(font.family);
+  const checkFontFaceRule = rule =>
+    /^@font-face {font-family:"g_d\d+_f1";src:/.test(rule);
+
+  beforeEach(() => {
+    globalThis.FontFace = function MockFontFace(name) {
+      this.family = name;
+    };
+  });
+
+  afterEach(() => {
+    globalThis.FontFace = FontFace;
+  });
+
+  function getMocks() {
+    const elements = [];
+    const createElement = name => {
+      let element =
+        typeof document !== "undefined" && document.createElement(name);
+      if (name === "style") {
+        element = {
+          tagName: name,
+          sheet: {
+            cssRules: [],
+            insertRule(rule) {
+              this.cssRules.push(rule);
+            },
+          },
+        };
+        Object.assign(element, {
+          remove() {
+            this.remove.called = true;
+          },
+        });
+      }
+      elements.push(element);
+      return element;
+    };
+    const ownerDocument = {
+      fonts: new Set(),
+      createElement,
+      documentElement: {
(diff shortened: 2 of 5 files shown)


test/unit/custom_spec.js
```diff
@@ -107,3 +107,129 @@ describe("custom canvas rendering", function () {
       .catch(done.fail);
   });
 });
+
+describe("alternate document context", function () {
+  const global = typeof globalThis !== "undefined" ? globalThis : window;
```
Snuffleupagus, line 112: Sorry, but this isn't correct. The whole point of `globalThis` is that it's always defined, and in a "normal" browser context `globalThis === window`.

Hence please remove this line, and change `global` to `globalThis` below.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,129 @@ describe("custom canvas rendering", function () {
       .catch(done.fail);
   });
 });
+
+describe("alternate document context", function () {
```
Snuffleupagus, line 111: Also, not that I look at this again a much better name would be `"custom ownerDocument"`, since that's actually consistent with the parameter name.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,129 @@ describe("custom canvas rendering", function () {
...
+        Object.assign(element, {
+          remove() {
+            this.remove.called = true;
+          },
+        });
+      }
+      elements.push(element);
+      return element;
+    };
+    const altDocument = {
```
Snuffleupagus, line 154: `ownerDocument`, here and elsewhere.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,129 @@ describe("custom canvas rendering", function () {
       .catch(done.fail);
   });
 });
+
+describe("alternate document context", function () {
+  const global = typeof globalThis !== "undefined" ? globalThis : window;
+  const FontFace = global.FontFace;
+
+  const ourFont = font => /g_d\d+_f1/.test(font.family);
```
Snuffleupagus, line 115: `checkFont` seems like a better name.

test/unit/custom_spec.js
```diff
@@ -107,3 +107,129 @@ describe("custom canvas rendering", function () {
...
+  const ourFontRule = rule =>
+    /^@font-face {font-family:"g_d\d+_f1";src:/.test(rule);
+  const has = (fn, iterable) => !![...iterable].find(fn);
```
Snuffleupagus, line 118: Both this function name, its implementation, and its usage all seem a slightly unclear; can you please re-factor the checking to be a bit more straightforward and easy to follow?

test/unit/custom_spec.js
```diff
@@ -107,3 +107,129 @@ describe("custom canvas rendering", function () {
       .catch(done.fail);
   });
 });
+
+describe("alternate document context", function () {
+  const global = typeof globalThis !== "undefined" ? globalThis : window;
+  const FontFace = global.FontFace;
+
+  const ourFont = font => /g_d\d+_f1/.test(font.family);
+  const ourFontRule = rule =>
```
Snuffleupagus, line 116: `checkFontFaceRule` seems like a better name.

Snuffleupagus (changes requested):

test/unit/custom_spec.js
jsg2021: This was my first inclination. I didn't know how far back you wanted to support. 

Sorry, for the irritation￼.

src/display/api.js
```diff
@@ -981,9 +987,10 @@ class PDFDocumentProxy {
  * Proxy to a PDFPage in the worker thread.
  */
 class PDFPageProxy {
-  constructor(pageIndex, pageInfo, transport, pdfBug = false) {
+  constructor(pageIndex, pageInfo, transport, ownerDocument, pdfBug = false) {
```
timvandermeij, line 990: Should we use the `ownerDocument = globalThis.document` fallback here too since the parameter is optional?

test/unit/custom_spec.js
```diff
@@ -107,3 +107,131 @@ describe("custom canvas rendering", function () {
...
+    expect(collectionMatches(ownerDocument.fonts, checkFont)).toBeTruthy();
+    await doc.destroy();
+    await loadingTask.destroy();
+    CanvasFactory.destroy(canvasAndCtx);
+    expect(ownerDocument.fonts.size).toBe(0);
+  });
+
+  it("should use given document for loading fonts (with CSS rules)", async function () {
+    const { ownerDocument, elements, CanvasFactory } = getMocks();
+    ownerDocument.fonts = null;
```
timvandermeij, line 206: Is this line necessary? You're creating new mocks in each test, so this will be an empty set by default, but since you're only checking the style tags in this test that shouldn't matter, right?

timvandermeij (approved):
Looks good to me with these final comments addressed and a rebase. r? @Snuffleupagus for a final check to make sure all is fine now. Thanks!

src/display/api.js
jsg2021: it probably shouldn't be optional internally, and only optional on public api

test/unit/custom_spec.js
jsg2021: that line is what triggers the font loader to fallback. If the document has a truthy value at that key it tries to use that api.

jsg2021:
rebased:)

test/unit/custom_spec.js
timvandermeij: Ah, I didn't know that. Thanks!

src/display/api.js
timvandermeij: That seems alright. Thanks!

timvandermeij:
Ping @Snuffleupagus for a final check to make sure I didn't miss anything here so we can merge this again.

Snuffleupagus:
> Gentle ping @Snuffleupagus if you have time for a final check to make sure I didn't miss anything.

Sorry about the delay here, I've been swamped with other things!

Unfortunately I still believe that https://github.com/mozilla/pdf.js/pull/12154#discussion_r464066181 applies here, even in the latest version of the patch, since https://github.com/mozilla/pdf.js/pull/12154/files#diff-348626e82475ec3536e620f4e5508f75R117-R118 looks quite cryptic and it's not (at least to me) *immediately* obvious what's actually happing there without *very careful* reading of the code.

@timvandermeij What's your opinion of this specific part of the unit-tests, since I worry that this admittedly clever but very terse checker function may hurt maintainability of these tests?

jsg2021:
@Snuffleupagus I'm struggling to understand. Is there a builtin assertion I could use? How is this cryptic? We just want to verify the set or list has the item matching our pattern?

jsg2021:
@Snuffleupagus I've updated to add a custom matcher which makes the callsite look like this:
https://github.com/mozilla/pdf.js/pull/12154/commits/a32ec02381bd466841c66eab8555d439b6827a73#diff-348626e82475ec3536e620f4e5508f75R230
```js
expect(ownerDocument.fonts).toContainElementMatching(checkFont);
```

The custom matcher is here: https://github.com/mozilla/pdf.js/pull/12154/commits/a32ec02381bd466841c66eab8555d439b6827a73#diff-348626e82475ec3536e620f4e5508f75R22-R54
and it replaces:
```js
const collectionMatches = (iterable, matcherFunction) =>
    Array.from(iterable).find(matcherFunction) !== undefined;
```

timvandermeij:
Oof, that's indeed _a lot_ more code for something that's not that difficult. Compared to what it was in the first iteration I think

```javascript
const collectionMatches = (iterable, matcherFunction) =>
    Array.from(iterable).find(matcherFunction) !== undefined;
```

was the most clear. Maybe it would help to not make it a variable like this but just inline the check? So:

```javascript
Array.from(ownerDocument.fonts).find(checkFont) !== undefined
```

I think `collectionMatches` is fine, at least I also don't see a shorter/more clear alternative here.

jsg2021:
Suggestion incorporated and rebased.

timvandermeij (approved):

timvandermeij:
Thank you!

jsg2021:
@timvandermeij What does the release schedule look like?

timvandermeij:
We don't have a fixed release schedule. I believe every week or so the code is merged into Firefox Nightly so the builds get tested by users, and NPM releases are usually made when a number of `[api-minor]` patches have landed or when lots of community members request a release. We're currently in the process of implementing support for AcroForms and layers, two big features, so I think we'll make a release after that's done.

jsg2021:
I look forward for the next npm release of pdfjs-dist 😊

PR #12156, opened 2020-08-02
Improve the types generation and API documentation (PR 12102 follow-up)

timvandermeij:
The commit messages contain more information about the individual changes. I recommend looking at the individual commits with the `?w=1` flag since that reduces the diff size for some of them.

/cc @Snuffleupagus since this should fix the consistency problems and your last comments on the types patch.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -666,8 +666,8 @@ class PDFDocumentProxy {
   }
 
   /**
-   * @returns {Promise<Object<string,any>>} A promise that is resolved with
-   *   a lookup table for mapping named destinations to reference numbers.
+   * @returns {Promise<Object<string, Array<any>>>} A promise that is resolved
+   *   with a mapping from named destinations to references.
    *
    * This can be slow for large documents. Use `getDestination` instead.
    */
@@ -685,10 +685,9 @@ class PDFDocumentProxy {
   }
 
   /**
-   * @returns {Promise<Array<string> | null>} A promise that is
-   *   resolved with an {Array} containing the page labels that correspond to
-   *   the page indexes, or `null` when no page labels are present in the PDF
-   *   file.
+   * @returns {Promise<Array<string> | null>} A promise that is resolved with
+   *   an {Array} containing the page labels that correspond to the page
+   *   indexes, or `null` when no page labels are present in the PDF file.
    */
   getPageLabels() {
     return this._transport.getPageLabels();
@@ -711,8 +710,9 @@ class PDFDocumentProxy {
   }
 
   /**
-   * @returns {Promise<Object>} A promise that is resolved with an {Object}
-   *   containing the viewer preferences.
+   * @returns {Promise<Object | null>} A promise that is resolved with an
+   *   {Object} containing the viewer preferences, or `null` when no viewer
+   *   preferences are present in the PDF file.
    */
   getViewerPreferences() {
     return this._transport.getViewerPreferences();
@@ -787,7 +787,7 @@ class PDFDocumentProxy {
   }
 
   /**
-   * @returns {Promise<Array<string | null>>} A promise that is resolved with
+   * @returns {Promise<Array<number> | null>} A promise that is resolved with
    *   an {Array} that contains the permission flags for the PDF document, or
    *   `null` when no permissions are present in the PDF file.
    */
@@ -2703,8 +2703,7 @@ class RenderTask {
      * Callback for incremental rendering -- a function that will be called
      * each time the rendering is paused.  To continue rendering call the
      * function that is the first argument to the callback.
-     * @callback
-     * @param {function}
+     * @type {function}
      */
     this.onContinue = null;
   }
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -89,6 +89,7 @@ let createPDFNetworkStream;
 /**
  * Sets the function that instantiates an {IPDFStream} as an alternative PDF
  * data transport.
+ *
  * @param {IPDFStreamFactory} pdfNetworkStreamFactory - The factory function
  *   that takes document initialization parameters (including a "url") and
  *   returns a promise which is resolved with an instance of {IPDFStream}.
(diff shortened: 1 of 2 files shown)


src/display/api.js
```diff
@@ -107,72 +108,72 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
...
+ * @property {string} [docBaseUrl] - The base URL of the document, used when
+ *   attempting to recover valid absolute URLs for annotations, and outline
+ *   items, that (incorrectly) only specify relative URLs.
+ * @property {string} [cMapUrl] - The URL where the predefined Adobe CMaps are
+ *   located. Include the trailing slash.
+ * @property {boolean} [cMapPacked] - Specifies if the Adobe CMaps are binary
+ *   packed or not.
+ * @property {Object} [CMapReaderFactory] - The factory that will be used when
+ *   reading built-in CMap files. Providing a custom factory is useful for
+ *   environments without `XMLHttpRequest` support, such as e.g. Node.js. The
```
Snuffleupagus, line 142: This comment is quite old, so you probably want to mention `Fetch API`/`XMLHttpRequest` here now.

src/display/api.js
```diff
@@ -107,72 +108,72 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
...
- *   fonts will be rendered using a built-in font renderer that constructs the
- *   glyphs with primitive path commands. The default value is `false`.
+ * @property {number} [maxImageSize] - The maximum allowed image size in total
+ *   pixels, i.e. width * height. Images above this value will not be rendered.
+ *   Use -1 for no limit, which is also the default value.
+ * @property {boolean} [isEvalSupported] - Determines if we can evaluate strings
+ *   as JavaScript. Primarily used to improve performance of font rendering, and
+ *   when parsing PDF functions. The default value is `true`.
+ * @property {boolean} [disableFontFace] - By default fonts are converted to
+ *   OpenType fonts and loaded via font face rules. If disabled, fonts will be
```
Snuffleupagus, line 155: Nit: `font face rules` -> `@font-face rules`.

src/display/api.js
```diff
@@ -187,13 +188,13 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
 
 /**
  * This is the main entry point for loading a PDF and interacting with it.
- * NOTE: If a URL is used to fetch the PDF data a standard XMLHttpRequest(XHR)
- * is used, which means it must follow the same origin rules that any XHR does
- * e.g. No cross domain requests without CORS.
+ * NOTE: If a URL is used to fetch the PDF data a standard XMLHttpRequest (XHR)
```
Snuffleupagus, line 191: You probably want to mention the Fetch API here, since that's now the default in modern browsers (with XMLHttpRequest used as fallback).

src/display/api.js
```diff
@@ -613,7 +615,7 @@ class PDFDataRangeTransport {
 }
 
 /**
- * Proxy to a PDFDocument in the worker thread. Also, contains commonly used
+ * Proxy to a `PDFDocument` in the worker thread that contains commonly used
  * properties that can be read synchronously.
```
Snuffleupagus, line 619: This comment is probably *very* old, and with less than a handful of exceptions almost all of it is now *a*synchronous. Hence I wonder if it the second sentence, or its contents, even makes sense here still!?

Also, compare with the `PDFPageProxy` class JSDocs below.

src/display/api.js
```diff
@@ -897,76 +885,73 @@ class PDFDocumentProxy {
...
- *                    CSS <color> value, a CanvasGradient object (a linear or
- *                    radial gradient) or a CanvasPattern object (a repetitive
- *                    image). The default value is 'rgb(255,255,255)'.
+ *   when creating canvases. The default value is {new DOMCanvasFactory()}.
+ * @property {Object} [background] - Background to use for the canvas. Can use
+ *   any valid `canvas.fillStyle`: a `DOMString` parsed as CSS <color> value,
+ *   a `CanvasGradient` object (a linear or radial gradient) or a
+ *   `CanvasPattern` object (a repetitive image). The default value is
+ *   'rgb(255,255,255)'.
  * @property {Object} [annotationStorage] - Storage for annotation data in
```
Snuffleupagus, line 940: Shouldn't this be `@property {AnnotationStorage} [annotationStorage] - ...`, since this shouldn't just a regular Object?

src/display/api.js
```diff
@@ -897,76 +885,73 @@ class PDFDocumentProxy {
...
  * @property {Object} [canvasFactory] - The factory instance that will be used
- *                    when creating canvases. The default value is
- *                    {new DOMCanvasFactory()}.
- * @property {Object} [background] - Background to use for the canvas.
- *                    Can use any valid canvas.fillStyle: A DOMString parsed as
- *                    CSS <color> value, a CanvasGradient object (a linear or
- *                    radial gradient) or a CanvasPattern object (a repetitive
- *                    image). The default value is 'rgb(255,255,255)'.
+ *   when creating canvases. The default value is {new DOMCanvasFactory()}.
+ * @property {Object} [background] - Background to use for the canvas. Can use
```
Snuffleupagus, line 935: Given the default value, shouldn't this ideally be `@property {<Object | string>} [background] - ...` or something similar?

src/display/api.js
```diff
@@ -755,22 +757,8 @@ class PDFDocumentProxy {
...
   /**
-   * @returns {Promise<Array<OutlineNode>>} A promise that is resolved with
-   * an {Array} that is a tree outline (if it has one) of the PDF. The tree is
-   * in the format of:
-   * [
-   *   {
-   *     title: string,
-   *     bold: boolean,
-   *     italic: boolean,
-   *     color: rgb Uint8ClampedArray,
```
Snuffleupagus, line 778: The fact that the `color` is in the RGB-format is missing from the `typedef`.

src/display/api.js
```diff
@@ -755,22 +757,8 @@ class PDFDocumentProxy {
...
-   *     count: integer or undefined,
```
Snuffleupagus, line 779: This property isn't even listed in the `typedef`, strangely enough.

src/display/api.js
```diff
@@ -755,22 +757,8 @@ class PDFDocumentProxy {
...
-   *     dest: dest obj,
```
Snuffleupagus, line 780: This is listed as `any` in the typedef, when it probably ought to be `<string | Array | null>`.

src/display/api.js
```diff
@@ -755,22 +757,8 @@ class PDFDocumentProxy {
...
-   *     url: string,
```
Snuffleupagus, line 781: This is listed as `any` in the typedef, when it probably ought to be `<string | null>`.

---

I know that you didn't write it, but please double-check the `typedef` to ensure that it makes sense (in addition to the things I spotted above).

Snuffleupagus (commented):
Looks *much* better already, however I couldn't help but spot a couple of (mostly pre-existing) issues.

Also, would it be possible to re-format this somehow, e.g. with line-breaks, to remove the `eslint-disable`/`eslint-enable` statements and improve overall readability (the ordering is also slightly weird, since `Uint8ClampedArray` should follow `Uint8Array`):
https://github.com/mozilla/pdf.js/blob/00a8b42e6703d824049eb317c05ada2eba3799a2/src/display/api.js#L100-L104

src/display/api.js
```diff
@@ -777,7 +777,7 @@ class PDFDocumentProxy {
   }
 
   /**
-   * @returns {Promise<Array<string | null>>} A promise that is resolved with
+   * @returns {Promise<Array<number | null>>} A promise that is resolved with
```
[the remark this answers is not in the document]
timvandermeij: You're completely right, this is indeed incorrect.

src/display/api.js
```diff
@@ -665,7 +665,7 @@ class PDFDocumentProxy {
   }
 
   /**
-   * @returns {Promise<Object<string,any>>} A promise that is resolved with
+   * @returns {Promise<Object<string, number>>} A promise that is resolved with
```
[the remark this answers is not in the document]
timvandermeij: Good find!

src/display/api.js
timvandermeij: I checked the code and the unit test and found two more properties that were missing. It should be complete and correct now.

timvandermeij:
For easier review, the interdiffs at https://github.com/mozilla/pdf.js/compare/3f38d3fd2b8a284b3d7cce2b4b97f03c4f958e94..d253e3923222f14a3cc5062cd83b7b018e3b4d67 (ignore the last four files since they are from a merge of the existing `master` branch) and https://github.com/mozilla/pdf.js/compare/d253e3923222f14a3cc5062cd83b7b018e3b4d67..bd55f3312d510f1a5ef2e34471b911797f9a0cc0 (for removing the ESLint ignores) display the changes I made.

Snuffleupagus (approved):
Much better, thank you for doing this!

PR #12157, opened 2020-08-03
Fix the type of SVGGraphics (PR 12102 follow-up)

tamuratak:
Fix the type of `SVGGraphics`.

The current generated type of `SVGGraphics` is:
```typescript
export function SVGGraphics(): never;
```
We can never call `new SVGGraphics(...)`  in TypeScript.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12158, opened 2020-08-03
Annotate some properties as private explicitly in src/display/api.js (PR 12102 follow-up)

tamuratak:
Annotate some properties as private explicitly in `src/display/api.js`. Otherwise, they become public in generated types of TypeScript.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -560,10 +560,15 @@ class PDFDataRangeTransport {
     this.initialData = initialData;
     this.progressiveDone = progressiveDone;
 
+    /** @private */
     this._rangeListeners = [];
+    /** @private */
     this._progressListeners = [];
+    /** @private */
     this._progressiveReadListeners = [];
+    /** @private */
     this._progressiveDoneListeners = [];
+    /** @private */
     this._readyCapability = createPromiseCapability();
   }
 
@@ -630,8 +635,11 @@ class PDFDataRangeTransport {
  */
 class PDFDocumentProxy {
   constructor(pdfInfo, transport) {
+    /** @private */
     this._pdfInfo = pdfInfo;
+    /** @private */
     this._transport = transport;
+    /** @private */
     this._annotationStorage = new AnnotationStorage();
   }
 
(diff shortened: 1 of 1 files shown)


Snuffleupagus (commented):
Using a leading underscore is pretty much the *standard* way of denoting "private" properties in JavaScript (things like private class fields are unfortunately too new for us to use), and having to add these sort of comments all over the code just to appease TypeScript doesn't seem that great.

timvandermeij:
@tamuratak Is there perhaps a way to tell `tsc` to treat variables with underscores as private so we don't have to add those annotations? I know that the underscores are nothing else but a convention and no strict rule, but it would be nice if `tsc` could be configured to adhere to that convention.

src/display/api.js
```diff
@@ -1062,10 +1076,12 @@ class PDFPageProxy {
    */
   getAnnotations({ intent = null } = {}) {
     if (!this.annotationsPromise || this.annotationsIntent !== intent) {
+      /** @private */
```
timvandermeij, line 1079: I would also think that these two are private, but they don't have an underscore. That's actually the case for more properties in this file I'm afraid, but that's a concern for another time.

tamuratak:
@timvandermeij No, there is not such a way as far as I know. 

I would not push this PR strongly because codes compile without errors even if these properties are public.

src/display/api.js
Snuffleupagus: > Until the time we do an API change, these are probably regarded as public now.

Not really, as far as I'm concerned, since those properties shouldn't be considered public and/or relied upon in any way; they are only intended to be *internal* functionality.
Generally speaking, I'd say that things which are not documented in `src/display/api.js` should *not* be considered public and can thus stop working at any time.

Snuffleupagus: Anyway, I don't think that we should accept this patch since it introduces an unfortunate precedent (as far as I'm concerned).
Given the established pattern of `leading underscore == private` in JavaScript, I think it's (somewhat) up to API users to be mindful of that.

---

If we want to make any changes, simply adding leading underscores to the two properties in question here seems straightforward enough; and I'd *not* even consider it an `api-minor` change.

timvandermeij:
I think we can indeed safely assume that everyone knows that underscores mean private, so even if they are exposed now that's not a big issue. Let's close this.

PR #12160, opened 2020-08-03
Use typedef to define the type of GlobalWorkerOptions (PR 12102 follow-up)

tamuratak:
Use `typedef` to define the type of `GlobalWorkerOptions`.

It would be much better to write explicitly in JSDoc that the type of `GlobalWorkerOptions.workerPort`  is `Worker`.

### DIFF
--- a/src/display/worker_options.js
+++ b/src/display/worker_options.js
@@ -13,24 +13,25 @@
  * limitations under the License.
  */
 
+/**
+ * @typedef {Object} GlobalWorkerOptionsType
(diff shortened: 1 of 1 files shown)


Snuffleupagus (commented):
Please actually format this consistently/correctly, see e.g. PR #12156.

tamuratak:
I have edited as requested.

timvandermeij (approved):

timvandermeij:
Looks good!

PR #12161, opened 2020-08-03
Add types to functions exported as API in src/pdf.js (PR 12102 follow-up)

tamuratak:
Add types to functions exported as API in `src/pdf.js`.


timvandermeij (approved):

timvandermeij:
Good improvement!

PR #12163, opened 2020-08-03
Re-factor the dispatching of "attachmentsloaded" events, when the PDF document contains no "regular" attachments

Snuffleupagus:
- Use a `DocumentFragment` when building the attachmentsView in `PDFAttachmentViewer.render`

   This approach is already used in other parts of the code-base, see e.g. `PDFOutlineViewer`, and has the advantage of only invalidating the DOM once rather than for every attachment item.

 - Re-factor the dispatching of "attachmentsloaded" events, when the PDF document contains no "regular" attachments

   Since the attachment fetching/parsing is already asynchronous, possibly delaying the dispatching of an "attachmentsloaded" event should thus not be a problem in general.
   Note that in some cases, i.e. PDF documents with no "regular" attachments and only FileAttachment annotations, we'll thus no longer dispatch an "attachmentsloaded" event when `attachmentsCount === 0` and instead wait for the FileAttachment parsing to finish. (The use of a timeout still guarantees that an "attachmentsloaded" event is *eventually* dispatched though.)

   This patch *considerably* simplifies the "attachmentsloaded" event handler, in `PDFSidebar`, since the details are now abstracted away from the consumer.

   Finally, add a check in `_appendAttachment` to ensure (indirectly) that the FileAttachment annotation is actually relevant for the active document.

### DIFF
--- a/web/pdf_attachment_viewer.js
+++ b/web/pdf_attachment_viewer.js
@@ -129,23 +129,22 @@ class PDFAttachmentViewer {
    * @param {PDFAttachmentViewerRenderParameters} params
    */
   render({ attachments, keepRenderedCapability = false }) {
-    let attachmentsCount = 0;
-
     if (this.attachments) {
       this.reset(keepRenderedCapability === true);
     }
     this.attachments = attachments || null;
 
     if (!attachments) {
-      this._dispatchEvent(attachmentsCount);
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
I agree that this is a simpler approach. Thanks!

PR #12165, opened 2020-08-04
Allow fuzzy matches of PDF ref tests.

brendandahl:


brendandahl:
First run to see what some sane defaults for the `MAX_PIXEL_DIFFERENCE` and `MAX_TOTAL_PIXELS` should be.
/botio test

brendandahl:
Hmm..now do we want to have settings per file or global?

package.json
```diff
@@ -67,5 +67,8 @@
     "type": "git",
     "url": "git://github.com/mozilla/pdf.js.git"
   },
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "dependencies": {
+    "pngjs": "^5.0.0"
```
Snuffleupagus, line 72: Please move this to the `devDependencies` section above.

Snuffleupagus (commented):
> Hmm..now do we want to have settings per file or global?

While global settings would obviously be the simplest, having it be local to each file does seem generally preferable to me since that should reduce the risk of accidentally overlooking new regressions.

Also, if it's done on a file basis: Would it perhaps be possible to only allow "failures" on particular operating systems and/or browsers, to further reduce the risk of missing new regressions?

brendandahl:
Yeah, per file/page/os/browser seems best. I haven't kept a super close eye on the intermittents, but it seems like they are usually the same files/pages?

timvandermeij:
In general this could indeed be useful since we don't want to end up overlooking regressions. I'm still hoping that a Puppeteer upgrade to the most recent version will also fix (some of) the intermittents, but I couldn't do the upgrade because the Windows bot kept failing with the new version for some reason. See for example http://54.215.176.217:8877/a1da348904f4db2/output.txt at the end. Not sure why, because the Linux bot works just fine with the most recent Puppeteer version. Are there perhaps processes that are still alive on the Windows bot? If those can be killed or we have an idea on how to fix that error, I'd be happy to redo the Puppeteer upgrade patch.

brendandahl:
Now that the failures seem more consistent on chrome we could revive this. I'm going to do a few runs to see how many pixel differences we're seeing.

/botio test

brendandahl:
Another round
/botio test

timvandermeij:
Closing since this is a very old PR and the number of intermittent failures in the reference tests is much more under control than it was when this patch was made. If we still want to do this for the remaining few failures we can always reopen/revisit this. Thanks.

PR #12168, opened 2020-08-04
Fix the `gulp types` task to run on Windows, and place the TypeScript definitions correctly in `pdfjs-dist`

Snuffleupagus:
- Fix the `gulp types` task to run on Windows. Currently this fails, and the solution was to "borrow" the same formatting as used in the `gulp jsdoc` task.

 - Place the TypeScript definitions in their own `types` directory, when building `pdfjs-dist`. These should *not* be cluttering the main `build` directory, especially since the generated TypeScript definitions consists of *multiple folders*. (Only if the TypeScript definitions would be concatenated into *a single file*, would placing them directly in `pdfjs-dist/build` be acceptable.)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1155,7 +1155,7 @@ gulp.task("types", function (done) {
     "forceConsistentCasingInFileNames",
     "emitDeclarationOnly",
   ].join(" --");
-  exec(`node_modules/.bin/tsc --${args} src/pdf.js`, done);
+  exec(`"node_modules/.bin/tsc" --${args} src/pdf.js`, done);
 });
 
 function buildLib(defines, dir) {
@@ -1366,7 +1366,9 @@ gulp.task(
           SRC_DIR + "pdf.worker.entry.js",
         ])
         .pipe(gulp.dest(TYPESTEST_DIR + "build/")),
-      gulp.src(TYPES_DIR + "**/**").pipe(gulp.dest(TYPESTEST_DIR + "build/")),
+      gulp
+        .src(TYPES_DIR + "**/*", { base: TYPES_DIR })
+        .pipe(gulp.dest(TYPESTEST_DIR + "types/")),
     ]);
   })
 );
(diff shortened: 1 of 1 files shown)


tamuratak:
You should change the `types` entry in the generated `package.json` in `pdfjs-dist` to point to the `pdf.d.ts` file.

```diff
diff --git a/gulpfile.js b/gulpfile.js
index 7da61ed1..6f3d480a 100644
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1640,7 +1640,7 @@ function packageBowerJson() {
     name: DIST_NAME,
     version: VERSION,
     main: "build/pdf.js",
-    types: "build/pdf.d.ts",
+    types: "types/pdf.d.ts",
     description: DIST_DESCRIPTION,
     keywords: DIST_KEYWORDS,
     homepage: DIST_HOMEPAGE,
```

See https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

tamuratak:
The following fixes `typestest`:
```diff
diff --git a/gulpfile.js b/gulpfile.js
index e44d6e04..51b9921f 100644
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1374,7 +1374,7 @@ gulp.task(
           .pipe(gulp.dest(TYPESTEST_DIR + "build/")),
         gulp
           .src(TYPES_BUILD_DIR + "**/**")
-          .pipe(gulp.dest(TYPESTEST_DIR + "build/")),
+          .pipe(gulp.dest(TYPESTEST_DIR + "types/")),
       ]);
     },
     function (done) {
```

I think `typestest` should directly run a test against files in`build/dist` or `node_modules/pdfjs-dist`, instead of copying files.

gulpfile.js
```diff
@@ -1154,7 +1154,7 @@ gulp.task("types", function (done) {
     "forceConsistentCasingInFileNames",
     "emitDeclarationOnly",
   ].join(" --");
-  exec(`node_modules/.bin/tsc --${args} src/pdf.js`, done);
+  exec(`"node_modules/.bin/tsc" --${args} src/pdf.js`, done);
```
timvandermeij, line 1157: If this fixes things on Windows, shouldn't https://github.com/mozilla/pdf.js/blob/71f8e1f538386fcbdb31d8cc098572ce32bf5c56/gulpfile.js#L1377 be changed too?

Snuffleupagus: Whoops, good catch :-)

timvandermeij (approved):

timvandermeij:
Tested locally as well and this seems to work just fine. Thanks!

PR #12169, opened 2020-08-04
Extract common methods from `PDFOutlineViewer`/`PDFAttachmentViewer` into a new abstract `BaseTreeViewer` class

Snuffleupagus:
These two classes are unsurprisingly quite similar, and with upcoming changes[1] the amount of (essentially) duplicated code will increase even further.

Notable changes:
 - Collect shared functionality in the `BaseTreeViewer` class, reducing both current and future code-duplication.
 - Reduce unnecessary duplication in the CSS rules, which will be particularly useful with upcoming changes.
 - Tweak the attachmentsView to use links, rather than buttons, to simplify (primarily) the CSS rules.

~~*Edit:* Currently blocked by PR #12163 landing first.~~

---
[1] Once API support for "Optional Content" lands, I've got more-or-less finished patches to add viewer support as well.

### DIFF
--- a/web/base_tree_viewer.js
+++ b/web/base_tree_viewer.js
@@ -1,0 +1,111 @@
+/* Copyright 2020 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


brendandahl (approved):
Looks good, thx!

timvandermeij (approved):

timvandermeij:

Issue #11954, opened 2020-06-01
Verify correctness of the required browser features pages


timvandermeij:
On https://github.com/mozilla/pdf.js/wiki/Required-Browser-Features we mention required browser features for PDF.js, but I have the feeling that not all of them are necessary nowadays since we moved to using more modern features over the years. Moreover, as indicated in https://github.com/mozilla/pdf.js/issues/11950#issuecomment-637149085, some features may be missing such as range requests or `async/await`.

In short, we should verify which are still necessary and update the wiki page and the automated feature detection if possible.

timvandermeij:
There is also the idea to replace the supported browsers overview with a verification based on which demo works; see https://github.com/mozilla/pdf.js/pull/12170#issuecomment-687562858. If we use that, we might also consider dropping the required browser features page since I think it's not relevant then anymore.

Snuffleupagus:
Given that this page hasn't received any *meaningful* updates since 2014[1], and that first of all fixing it and secondly keeping it up-to-date will take quite a bit of effort, I'd suggest that we move forward with https://github.com/mozilla/pdf.js/issues/11954#issuecomment-687587562 instead.

---
[1] There's been changes since then, but that's when the last feature-test was added.

PR #12170, opened 2020-08-04
[api-minor] Add support for toggling of Optional Content in the viewer (issue 12096) 

### LINKED ISSUE #12096: Add UI to control what optional content is visible
Once #12095 lands, we should add a UI view similar to the other side bar views that controls what optional content layers are visible.

To do this we'll need to:
- add a class similar to PDFOutlineViewer
- wire that class to set the visibility of the groups in the OptionalContentConfig 
- pass in the modified OptionalContentConfig to page.render()
- add a test that changes what groups are visible

Snuffleupagus:
~~*Please note:* Currently blocked by PRs #12163 and #12169 landing first.~~

Fixes #12096


timvandermeij:
I have been playing with this a bit and it seems to work nicely, but there is one thing I don't really understand. If you enable some layers that are not enabled by default and then double-click the layers icon, all checkboxes reset to their default state. I would expect the checkboxes to remain in the same configuration and only the subtrees to be collapses/expanded. Is this intentional behavior or is it a bug?

Snuffleupagus:
>  If you enable some layers that are not enabled by default and then double-click the layers icon, all checkboxes reset to their default state.

It doesn't *just* reset the sidebarView though, it actually re-renders the *entire* document with the *default* visibility state of all optional content groups.
Hence this is absolutely intended behaviour, since otherwise there'd be no way of resetting the rendering of a document to its initial state, short of reloading it which may not always be an option.

>  I would expect the checkboxes to remain in the same configuration and only the subtrees to be collapses/expanded.

That's the behaviour of the outlineView, but for the layersView the resetting (as described above) seems *much more* useful in general as far as I'm concerned. Note also that most PDF documents with optional content groups only have a single nesting level as well.

---

All-in-all, what we'd probably need to fix all of this in a better way is a small "menu" in the sidebar itself (maybe similar to the secondaryToolbar), where context-specific sidebarView functionality is placed (since the double-click on the toggle-button probably isn't great UX overall); however that seems like a separate issue :-)

timvandermeij:
I just now found out where my confusion comes from. If you open the preview and hover over the layers icon, the tooltip is about collapsing/expanding the tree, which explains my confusion from above when the visibility state suddenly got reset. It's because the tooltip text and localization files are out of sync; see:

https://github.com/mozilla/pdf.js/pull/12170/files#diff-a72ef320c3225c98b2163b89a3e8ff62R89

for the incorrect text and:

https://github.com/mozilla/pdf.js/pull/12170/files#diff-1be7f68f09319730f08fe605db4470aeR146

for the correct text. I think this is just a simple copy-paste mistake.

Snuffleupagus:
> I think this is just a simple copy-paste mistake.

Whoops, fixed now :-)
Having run `gulp locale` locally helped cover this, so thanks for spotting the error!

Snuffleupagus:
Somehow I completely forgot about the thumbnails in the initial patch; they will now *always* render with all Optional Content Groups at their default visibility state.

/botio-linux preview

Snuffleupagus:
*I believe that this should now, finally, be ready for full review :-)*

src/core/obj.js
```diff
@@ -315,6 +315,67 @@ class Catalog {
       return onParsed;
     }
 
+    function parseOrder(refs, nestedLevels = 0) {
```
janpe2, line 318: The first object in a subarray of the `Order` entry can be also an OCG instead of a string. The OCG becomes an expandable node. If it is set to invisible, all child nodes should get grayed out. See these files:
[PDFill_layers.pdf](https://github.com/mozilla/pdf.js/files/5052227/PDFill_layers.pdf)
[pdflib_hierarchical_layers.pdf](https://github.com/mozilla/pdf.js/files/5052228/pdflib_hierarchical_layers.pdf)
Also the mutually exclusive feature of radio buttons (`RBGroups`) and locked OCGs seems to be still unimplemented, if I'm right?

l10n/sv-SE/viewer.properties
```diff
@@ -137,17 +137,20 @@ print_progress_close=Avbryt
...
 attachments.title=Visa Bilagor
 attachments_label=Bilagor
+layers.title=Visa lager (dubbelklicka för att återställa alla lager till ursrungligt läge)
+layers_label=Lager
 thumbs.title=Visa miniatyrer
 thumbs_label=Miniatyrer
 findbar.title=Sök i dokument
 findbar_label=Sök
 
+additional_layers=Ytterliggare lager
```
janpe2, line 153: There are some tiny typos in this Swedish file. Or will these be overwritten when strings arrive from the translation process?

janpe2:
The edges of the icons look a bit jagged.
kuva

l10n/sv-SE/viewer.properties
Snuffleupagus: Yes, there's obviously an embarrassing typo on this line; fixed now :-)
(It seems that a bad spell-checker might have "helped" me here.)

Generally speaking, we're importing all translations from Firefox and all `l10n/` files are thus updated/overwritten regularly; any pre-existing bugs thus need to be filed against the Firefox localization in Bugzilla.
(When adding *new* strings, I simply find it helpful to add not just the default language for testing purposes.)

src/core/obj.js
Snuffleupagus: > The first object in a subarray of the `Order` entry can be also an OCG instead of a string. The OCG becomes an expandable node.

Very interesting, I hadn't seen those types of PDF documents when writing the patches.
I'll try to see how difficult implementing that would be; or if we could possibly defer that for a follow-up PR.

> Also the mutually exclusive feature of radio buttons (`RBGroups`) and locked OCGs seems to be still unimplemented, if I'm right?

Yes, those weren't implemented in PR #12095 and here I simply implemented the *bare minimum* of additional "core" functionality to get the viewer integration to work OK.
I'd really prefer if those things were implemented separately though, since this PR is already quite large (and those features also don't seem *that* common); there's also the question if we'd even want to respect the "locked OCGs" in the viewer.

Snuffleupagus:
> The edges of the icons look a bit jagged.

Yes, and that's even mentioned in the commit message :-)
Please note:
> Also, the current Layers-button icons are a little rough around the edges, quite literally, but given that the viewer will soon have its UI modernized anyway they hopefully suffice in the meantime.

src/core/obj.js
Snuffleupagus: > > The first object in a subarray of the `Order` entry can be also an OCG instead of a string. The OCG becomes an expandable node.
> 
> Very interesting, I hadn't seen those types of PDF documents when writing the patches.
> I'll try to see how difficult implementing that would be; or if we could possibly defer that for a follow-up PR.

Having looking (quickly) at this, I'd *really* prefer if we could defer that to a follow-up patch instead.

janpe2:
> > The edges of the icons look a bit jagged.

> Yes, and that's even mentioned in the commit message :-)

Sorry, I missed the commit message.

Snuffleupagus:
@brendandahl Do you have time to review this PR, since it adds (basic) support for toggling of optional content in the viewer?

timvandermeij:
This needs a rebase for the new Photon UI. @utopianknight Could you perhaps provide an SVG image for the layers icon that we can use here in the Photon style?

utopianknight:
It looks like this. Is this alright? Where do I upload this?

Annotation 2020-08-20 142531

utopianknight:
`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 4.233 4.233" height="16" width="16"><path d="M.15 2.992c-.198.1-.2.266-.002.365l1.604.802a.93.93 0 00.729-.001l1.602-.801c.198-.1.197-.264 0-.364l-.695-.348c-1.306.595-2.542 0-2.542 0m-.264.53l.658-.329c.6.252 1.238.244 1.754 0l.659.329-1.536.768zM.15 1.935c-.198.1-.198.265 0 .364l1.604.802a.926.926 0 00.727 0l1.603-.802c.198-.099.198-.264 0-.363l-.694-.35c-1.14.56-2.546.001-2.546.001m-.264.53l.664-.332c.52.266 1.261.235 1.75.002l.659.33-1.537.768zM.15.877c-.198.099-.198.264 0 .363l1.604.802a.926.926 0 00.727 0l1.603-.802c.198-.099.198-.264 0-.363L2.481.075a.926.926 0 00-.727 0zm.43.182L2.117.29l1.538.769-1.538.768z"/></svg>`

Snuffleupagus:
> It looks like this. Is this alright?

That looks *great*, thank you!

> Where do I upload this?

Simply copy-pasting the raw SVG code, that you provided, seemed to work fine.

brendandahl (changes requested):
I've looked through the src/* files and just have a few minor changes. @timvandermeij I think you're more familiar with the viewer changes, is there anything else you wanted address?

brendandahl (approved):

Snuffleupagus:
*Rebased to `master` to ensure that it still applies cleanly, no actual code changes were made here.*

timvandermeij (approved):

timvandermeij:
Let's do this. Awesome work!

Snuffleupagus:
Thanks for the reviews here!

---

Slightly unrelated, but should we perhaps create a new release now?
There's been three months since the last one, and there's been *a lot* of nice changes landing lately such as optional content, improved form functionality, long-standing font bugs fixed, and more.

I'd also very much suggest that we officially note the next release as the *last one with IE 11/Edge (Trident-based) support*, to moving forward with some much needed clean-up/improvements.

brendandahl:
> Slightly unrelated, but should we perhaps create a new release now?

SGTM

tbabej:
Thanks for the great work here guys!

timvandermeij:
Good idea. We're now at a point where the functionality has landed and the bugs from testing have been fixed. I'll make the release today.

Snuffleupagus:
>  I'll make the release today.

Thank you for doing this!

---

Just one question here: How about IE 11, couldn't we have explicitly marked the new release as the last one with IE 11 support!?

One example of an IE 11 issue that impacts the Firefox built-in PDF Viewer negatively is the need to ship *separate* toolbar images for the light/dark themes, thus effectively *doubling* the size of the image resources for no good reason except for IE 11 support.

timvandermeij:
I forgot to add that while making the release. I'm not sure what happens if we edit a pre-release on GitHub, if the hooks are re-triggered or not. I'll add this as a question to the NPM issue.

Snuffleupagus:
> I'm not sure what happens if we edit a pre-release on GitHub, if the hooks are re-triggered or not. I'll add this as a question to the NPM issue.

Good point; in the meantime though updating https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support to list IE 11/Edge (Trident-based) as *unsupported* after `v2.6.347` might not be a bad idea.

However, looking at that list it does seem pretty out-of-date in general and it could probably do with some clean-up. Keeping that list even remotely current is (and has been) difficult since we cannot really test all of the mentioned configurations, hence I wonder if there's perhaps a better way to handle things!?
My somewhat naive idea would be to use https://github.com/mozilla/pdf.js#online-demo in order to define supported browsers:
 - If the "Modern browsers"-version works in a particular browser, it's regarded as supported.
 - If the "Older browsers"-version works in a particular browser, it's regarded as *partially* supported.
 - If none of demo viewers work in a particular browser, it's simply *not* supported.

It might be more appropriate to discuss this further in a separate issue, but the above is one idea that attempt to limited the maintenance burden of the list.

timvandermeij:
I have updated the wiki and added the online demo idea to #11954 which I think is related since I think the required browser features page is in the same way out of date.

PR #12171, opened 2020-08-04
Attempt to reduce intermittent failures in the "multiple render() on the same canvas" unit-test

Snuffleupagus:
This patch should *hopefully* remove the intermittent unit-test failure, by using the *same* `optionalContentConfigPromise` for both `renderTask`s and thus get more predictable timing behaviour.

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -1826,6 +1826,8 @@ describe("api", function () {
     });
 
     it("multiple render() on the same canvas", function (done) {
+      const optionalContentConfigPromise = pdfDocument.getOptionalContentConfig();
+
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice find; thank you for getting rid of all those pesky intermittents the last few days!

Snuffleupagus:
> Nice find; thank you for getting rid of all those pesky intermittents the last few days!

No problem; in the event that this patch still isn't enough, there's additional things can be done to *really* fix the test. However that may render it somewhat "messy", hence I went with a simpler approach for now.

PR #12172, opened 2020-08-05
Fix the type definition of TypedArray. (PR 12156 follow-up)

tamuratak:
Fix the type definition of `TypedArray`. 

The type of `TypedArray` is `any` in the current generated files.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -100,9 +100,10 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
 }
 
 /**
- * @typedef {
- *   Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array |
- *   Int32Array | Uint32Array | Float32Array | Float64Array
+ * @typedef { Int8Array | Uint8Array | Uint8ClampedArray |
+ *            Int16Array | Uint16Array |
+ *            Int32Array | Uint32Array | Float32Array |
+ *            Float64Array
  * } TypedArray
  */
 


src/display/api.js
```diff
@@ -99,12 +99,11 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
...
 
+/* eslint-disable max-len */
 /**
- * @typedef {
- *   Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array |
- *   Int32Array | Uint32Array | Float32Array | Float64Array
- * } TypedArray
+ * @typedef {Int8Array | Uint8Array | Uint8ClampedArray | Int16Array | Uint16Array | Int32Array | Uint32Array | Float32Array | Float64Array} TypedArray
  */
+/* eslint-enable max-len */
```
Snuffleupagus, line 106: The reason that we enforce a maximum line-length, even for comments, is related to readability. Hence disabling of the ESLint `max-len` rule should basically never happen (even more so now that Prettier is being used).

(Furthermore, note that we always prefer using `// eslint-disable-line` or `// eslint-disable-next-line`, since this sort of blanket disabling increases the risk that one forgets to re-enable a rule.)

---

However, based on a very quick test there seems to be a *much* better solution to this problem:
```javascript
/**
 * @typedef { Int8Array | Uint8Array | Uint8ClampedArray |
 *            Int16Array | Uint16Array |
 *            Int32Array | Uint32Array | Float32Array |
 *            Float64Array
 * } TypedArray
 */
```

Snuffleupagus (changes requested):

src/display/api.js
tamuratak: I have edited as suggested. Thanks.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #12173, opened 2020-08-05
[api-minor] Fix the `annotationStorage` parameter in `PDFPageProxy.render`

Snuffleupagus:
While the parameter name (clearly) suggests that an `AnnotationStorage`-instance is expected, looking at the only call-sites that include the parameter (i.e. the `PDFPrintServiceFactory` instances) it actually contains just a normal Object.

Hence it seems much more reasonable to actually pass a valid `AnnotationStorage`-instance, as the name suggests, and simply have `PDFPageProxy.render` do the `annotationStorage.getAll()` call. (Since we cannot send an `AnnotationStorage`-instance as-is to the worker-thread, given the "structured clone algorithm".)

### DIFF
--- a/web/firefox_print_service.js
+++ b/web/firefox_print_service.js
@@ -57,7 +57,7 @@ function composePage(
           transform: [PRINT_UNITS, 0, 0, PRINT_UNITS, 0, 0],
           viewport: pdfPage.getViewport({ scale: 1, rotation: size.rotation }),
           intent: "print",
-          annotationStorage: pdfDocument.annotationStorage.getAll(),
+          annotationStorage: pdfDocument.annotationStorage,
         };
         return pdfPage.render(renderContext).promise;
       })
--- a/web/interfaces.js
+++ b/web/interfaces.js
@@ -165,7 +165,8 @@ class IPDFAnnotationLayerFactory {
   /**
    * @param {HTMLDivElement} pageDiv
    * @param {PDFPage} pdfPage
-   * @param {AnnotationStorage} [annotationStorage]
+   * @param {AnnotationStorage} [annotationStorage] - Storage for annotation
+   *   data in forms.
    * @param {string} [imageResourcesPath] - Path for image resources, mainly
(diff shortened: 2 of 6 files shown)


web/firefox_print_service.js
```diff
@@ -80,17 +86,29 @@ function FirefoxPrintService(pdfDocument, pagesOverview, printContainer) {
   this.pdfDocument = pdfDocument;
   this.pagesOverview = pagesOverview;
   this.printContainer = printContainer;
+  this._annotationStorage = pdfDocument.annotationStorage;
```
timvandermeij, line 89: In contrast to the print resolution from the other patch, the annotation storage is available on the `pdfDocument` variable that we already pass in everywhere, so why should we split this off? It's an extra variable we have pass everywhere and the number of function parameters is already quite high.

web/interfaces.js
```diff
@@ -165,7 +165,8 @@ class IPDFAnnotationLayerFactory {
   /**
    * @param {HTMLDivElement} pageDiv
    * @param {PDFPage} pdfPage
-   * @param {AnnotationStorage} [annotationStorage]
+   * @property {AnnotationStorage} [annotationStorage] - Storage for annotation
```
timvandermeij, line 168: This should be `@param` given that it's a function parameter and not a type property.

Snuffleupagus: Sorry, too much copy-and-paste; fixed now!

web/firefox_print_service.js
Snuffleupagus: I suppose that it's not *strictly* necessary, since the `AnnotationStorage`-instance is a bit of a special-case given that it's always unconditionally defined (whether that's a good idea is another question, but it's probably small enough to not really matter).

However, you obviously have a point about the unwieldy signatures (and that will get worse soon anyway...), so I'll just revert these changes; thank you!

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12174, opened 2020-08-05
Do the `AppOptions.get("printResolution")` lookup once in `web/app.js `, when initializing `PDFPrintServiceFactory`-instances, rather than for every printed page

Snuffleupagus:
There's really no point in repeating these lookups over and over, since the value should be constant for the duration of one print invocation anyway.


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #12175, opened 2020-08-05
Support textfield and choice widgets for printing

calixteman:
There will have an other patch to handle multiline textfields and comb textfields.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -21,9 +21,9 @@ import {
   AnnotationReplyType,
   AnnotationType,
   assert,
+  escapeString,
   isString,
   OPS,
-  stringToBytes,
   stringToPDFString,
   Util,
   warn,
@@ -33,7 +33,7 @@ import { Dict, isDict, isName, isRef, isStream } from "./primitives.js";
 import { ColorSpace } from "./colorspace.js";
 import { getInheritableProperty } from "./core_utils.js";
 import { OperatorList } from "./operator_list.js";
-import { Stream } from "./stream.js";
+import { StringStream } from "./stream.js";
 
 class AnnotationFactory {
   /**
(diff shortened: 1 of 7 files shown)


src/core/annotation.js
```diff
@@ -888,17 +888,198 @@ class WidgetAnnotation extends Annotation {
...
+      descent = 0;
+    }
+
+    const vPadding = defaultPadding + Math.abs(descent) * fontSize;
+    const defaultAppearance = this.data.defaultAppearance;
+
+    const alignment = this.data.textAlignment;
+    if (alignment === 0 || alignment > 2) {
+      // Left alignment: nothing to do
+      return `/Tx BMC q BT ${defaultAppearance} 1 0 0 1 ${hPadding} ${vPadding} Tm (${value}) Tj ET Q EMC`;
```
timvandermeij, line 993: Split this line up to not go over the 80 character limit.

src/core/annotation.js
```diff
@@ -888,17 +888,198 @@ class WidgetAnnotation extends Annotation {
...
+    const renderedText = this.renderPDFText(
+      value,
+      font,
+      fontSize,
+      totalWidth,
+      alignment,
+      hPadding,
+      vPadding
     );
+    return `/Tx BMC q BT ${defaultAppearance} 1 0 0 1 0 0 Tm ${renderedText} ET Q EMC`;
```
timvandermeij, line 1005: Same here for the line length.

src/core/annotation.js
```diff
@@ -888,17 +888,198 @@ class WidgetAnnotation extends Annotation {
...
+  }
+
+  async getFontData(evaluator, task) {
```
timvandermeij, line 1008: `_getFontData`

src/core/annotation.js
```diff
@@ -888,17 +888,198 @@ class WidgetAnnotation extends Annotation {
...
+      task,
+      resources: this.fieldResources,
+      operatorList,
+      initialState,
+    });
+
+    return [initialState.font, initialState.fontName, initialState.fontSize];
+  }
+
+  computeAutoSizedFont(font, fontName, fontSize, height) {
```
timvandermeij, line 1030: `_computeFontSize` (also a bit shorter)

src/core/annotation.js
```diff
@@ -888,17 +888,198 @@ class WidgetAnnotation extends Annotation {
...
+    if (fontSize === null || fontSize === 0) {
+      const em = font.charsToGlyphs("M", true)[0].width / 1000;
+      // According to https://en.wikipedia.org/wiki/Em_(typography)
+      // an average cap height should be 70% of 1em
+      const capHeight = 0.7 * em;
+      // 1.5 * capHeight * fontSize seems to be a good value for lineHeight
+      fontSize = Math.max(1, Math.floor(height / (1.5 * capHeight)));
+
+      let re = new RegExp(`/${fontName}\\s+[0-9\.]+\\s+Tf`);
```
timvandermeij, line 1039: `re` -> `fontRegex` (more descriptive)

src/core/annotation.js
```diff
@@ -888,17 +888,198 @@ class WidgetAnnotation extends Annotation {
...
+      }
+      this.data.defaultAppearance = this.data.defaultAppearance.replace(
+        re,
+        `/${fontName} ${fontSize} Tf`
+      );
+    }
+    return fontSize;
+  }
+
+  renderPDFText(
```
timvandermeij, line 1052: `_renderText` (also a bit shorter)

src/display/annotation_layer.js
```diff
@@ -678,6 +688,12 @@ class ChoiceWidgetAnnotationElement extends WidgetAnnotationElement {
       selectElement.appendChild(optionElement);
     }
 
+    selectElement.addEventListener("change", function (event) {
```
timvandermeij, line 691: Shouldn't we also use `storage.getOrCreateValue` in this function to select the value in the storage a few lines above?

src/shared/util.js
```diff
@@ -948,4 +953,5 @@ export {
   utf8StringToString,
   warn,
   unreachable,
+  escapeString,
```
timvandermeij, line 956: Please order this one alphabetically in the list for a better overview.

test/unit/annotation_spec.js
```diff
@@ -1552,6 +1570,74 @@ describe("annotation", function () {
...
+          annotationStorage[id] = "test (print)";
+          return annotation.getAppearance(
+            partialEvaluator,
+            task,
+            annotationStorage
+          );
+        }, done.fail)
+        .then(appearance => {
+          expect(appearance).toEqual(
+            "/Tx BMC q BT /Helv 11 Tf 1 0 0 1 0 0 Tm 2.00 2.00 Td (test \\(print\\)) Tj ET Q EMC"
```
timvandermeij, line 1603: Split this up to stay within the 80 character line length.

test/unit/annotation_spec.js
```diff
@@ -1552,6 +1570,74 @@ describe("annotation", function () {
       }
       promise.then(done, done.fail);
     });
+
+    it("should render auto-sized text for printing", function (done) {
```
timvandermeij, line 1574: Can you also add a test `should render regular text for printing`, i.e., where the auto-size logic is not called and the font size from the default appearance is used? I think all cases are covered then: regular text, auto-sized text and password fields.

test/unit/annotation_spec.js
```diff
@@ -2128,6 +2231,39 @@ describe("annotation", function () {
...
+          annotationStorage[id] = "a value";
+          return annotation.getAppearance(
+            partialEvaluator,
+            task,
+            annotationStorage
+          );
+        }, done.fail)
+        .then(appearance => {
+          expect(appearance).toEqual(
+            "/Tx BMC q BT /Helv 5 Tf 1 0 0 1 0 0 Tm 2.00 2.00 Td (a value) Tj ET Q EMC"
```
timvandermeij, line 2262: Same for the line length.

timvandermeij (commented):
Nice work! Should be fine after these comments.

src/shared/util.js
```diff
@@ -793,6 +793,11 @@ function stringToPDFString(str) {
   return strBuf.join("");
 }
 
+function escapeString(str) {
```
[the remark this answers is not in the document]
calixteman: > Also, if the string contains unicode we need to ensure the string starts with the unicode BOM.

It tryed to add an UTF-16BE BOM but the string is still considered as normal string, for example `<FEFF0041>` is equivalent to `"\xfe\xff\00\x41"` when used with command `Tj`.
I tryed this pdf in pdf.js:
```
%PDF-1.4
1 0 obj <</Type /Catalog /Pages 2 0 R>>
endobj
2 0 obj <</Type /Pages /Kids [3 0 R] /Count 1>>
endobj
3 0 obj<</Type /Page /Parent 2 0 R /Resources 4 0 R /MediaBox [0 0 500 800] /Contents 6 0 R>>
endobj
4 0 obj<</Font <</F1 5 0 R>>>>
endobj
5 0 obj<</Type /Font /Subtype /Type1 /BaseFont /Helvetica>>
endobj
6 0 obj
<</Length 44>>
stream
BT /F1 24 Tf 175 720 Td <FEFF004100420043> Tj ET
endstream
endobj
xref
0 7
0000000000 65535 f
0000000009 00000 n
0000000056 00000 n
0000000111 00000 n
0000000212 00000 n
0000000250 00000 n
0000000317 00000 n
trailer <</Size 7/Root 1 0 R>>
startxref
413
%%EOF
```
and the string is displayed as a 8 chars string.
So I guess that when we've a non-ascii string we must find a font which can display it and add the necessary resources and then re-encode the string for this font.
We could probably cheat when printing in trying to render the string on the canvas with an unicode font but for the saving part we need to handle that correctly.
So if I'm correct we could probably do that later in an other patch.
Wdyt ?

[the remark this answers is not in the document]
timvandermeij: I agree that that should be done in a follow-up, mainly to limit the scope of this patch but also because it's not a big problem for now (indeed, for saving this should be taken care of and can then also be done for rendering).

timvandermeij (approved):

timvandermeij:
Awesome work!

PR #12176, opened 2020-08-05
Support multiline textfields for printing

calixteman:


src/core/annotation.js
```diff
@@ -1081,6 +1094,92 @@ class WidgetAnnotation extends Annotation {
 
     return `${shift} ${vPadding} Td (${text}) Tj`;
   }
+
+  handleMultiline(text, font, fontSize, width, alignment, hPadding, vPadding) {
+    const lines = text.replace("\r\n", "\n").split("\n");
+    let buf = "";
+    const totalWidth = alignment === 1 ? width : width - 2 * hPadding;
```
timvandermeij, line 1101: This could use a comment. What is `alignment === 1`; center, left of right? Why is that special cased?

src/core/annotation.js
```diff
@@ -1081,6 +1094,92 @@ class WidgetAnnotation extends Annotation {
...
+      for (const chunk of chunks) {
+        if (buf === "") {
+          buf = this.renderPDFText(
+            chunk,
+            font,
+            fontSize,
+            width,
+            alignment,
+            hPadding,
+            -fontSize
```
timvandermeij, line 1113: Why take the negative here?

src/core/annotation.js
```diff
@@ -1081,6 +1094,92 @@ class WidgetAnnotation extends Annotation {
...
+              -fontSize
+            );
+        }
+      }
+    }
+
+    return buf;
+  }
+
+  splitLine(line, font, fontSize, width) {
```
timvandermeij, line 1134: Let's rename this to `_splitLine` to make it private.

src/core/annotation.js
```diff
@@ -1081,6 +1094,92 @@ class WidgetAnnotation extends Annotation {
...
+    const scale = fontSize / 1000;
+    const white = font.charsToGlyphs(" ", true)[0].width * scale;
```
timvandermeij, line 1136: I think `const whitespace` would be bit more descriptive here.

src/core/annotation.js
```diff
@@ -1081,6 +1094,92 @@ class WidgetAnnotation extends Annotation {
...
+    const chunks = [];
+
+    let lastSpacePos = -1,
+      startChunk = 0,
+      currentWidth = 0;
+
+    for (let i = 0; i < line.length; i++) {
```
timvandermeij, line 1143: We usually do `for (let i = 0, ii = line.length; i < ii; i++) {` to cache the line length once.

src/core/annotation.js
```diff
@@ -1081,6 +1094,92 @@ class WidgetAnnotation extends Annotation {
...
+          currentWidth = white;
+          lastSpacePos = -1;
+        } else {
+          currentWidth += white;
+          lastSpacePos = i;
+        }
+      } else {
+        const charWidth = font.charsToGlyphs(character, false)[0].width * scale;
+        if (currentWidth + charWidth > width) {
+          // We must break to the last white position (if one)
```
timvandermeij, line 1159: `last whitespace position (if available)`

test/unit/annotation_spec.js
```diff
@@ -1638,6 +1638,41 @@ describe("annotation", function () {
           done();
         }, done.fail);
     });
+
+    it("should render splited text for printing", function (done) {
```
timvandermeij, line 1642: `should render multiline text for printing`

test/unit/annotation_spec.js
```diff
@@ -1638,6 +1638,41 @@ describe("annotation", function () {
...
+      AnnotationFactory.create(
+        xref,
+        textWidgetRef,
+        pdfManagerMock,
+        idFactoryMock
+      )
+        .then(annotation => {
+          const id = annotation.data.id;
+          const annotationStorage = {};
+          annotationStorage[id] = "a aa aaa aaaa aaaaa aaaaaa";
```
timvandermeij, line 1662: Can we make this string so that it also covers the "break in the middle of string" (no whitespace found) codepath, since that seems to be uncovered here? Perhaps introducing a large word would already be enough. If this line becomes longer than 80 characters, then concatenate smaller substrings.

test/unit/annotation_spec.js
```diff
@@ -1638,6 +1638,41 @@ describe("annotation", function () {
...
+          return annotation.getAppearance(
+            partialEvaluator,
+            task,
+            annotationStorage
+          );
+        }, done.fail)
+        .then(appearance => {
+          expect(appearance).toEqual(
+            "/Tx BMC q BT /Helv 5 Tf 1 0 0 1 0 10 Tm 2.00 -5.00 Td (a aa aaa ) Tj\n0.00 -5.00 Td (aaaa aaaaa ) Tj\n0.00 -5.00 Td (aaaaaa) Tj ET Q EMC"
```
timvandermeij, line 1671: Same here about concatenating smaller substrings to stay within the 80 character limit.

timvandermeij (commented):
Looks good with the comments addressed (only the second commit was checked here since the rest is in another PR).

src/core/annotation.js
calixteman: I added a comment to explain.

timvandermeij:
This doesn't seem to be working correctly. I tried inserting text into the multiline textfield from https://github.com/mozilla/pdf.js/blob/master/test/pdfs/annotation-text-widget.pdf (and other PDFs), but the print result is an empty textfield.

calixteman:
I didn't manage to reproduce using print button but I managed to have an incorrect behaviour using keyboard shortcut (and in this case `change` event was not triggered so I replaced it with `input` event).
Anyway, could you try again ?
I fixed a bug with `\r` as line separator and with non-empty fields which have been changed into empty ones.
I'll add some tests in the weekend.

timvandermeij:
Unfortunately the problem remains. Steps to reproduce:

1. Visit http://54.67.70.0:8877/a931282262c09f9/web/viewer.html (preview link above)
2. Run `PDFViewerApplication.preferences.set('renderInteractiveForms', true)` in the console and refresh.
3. Download https://github.com/mozilla/pdf.js/blob/master/test/pdfs/annotation-text-widget.pdf and open the file with the viewer.
4. Enter multiline text in the textarea.
5. Click the print button in the toolbar or Ctrl-P and print the result to PDF.

Attached is the result I get from this: 
[out.pdf](https://github.com/mozilla/pdf.js/files/5045344/out.pdf)

src/display/annotation_layer.js
```diff
@@ -462,7 +462,7 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         element.setAttribute("value", textContent);
       }
 
-      element.addEventListener("change", function (event) {
+      element.addEventListener("input", function (event) {
```
timvandermeij, line 465: If this is a problem here, shouldn't it also be changed for checkbox and radio button widgets from previous PRs? If so, please change that in a separate commit.

calixteman:
It was because of the alignment var created after its use... I wonder why neither the linter nor firefox dev edition didn't catch it...
Anyway I fixed it and added a test for multiple possible EOL.
About the event: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event
So it seems that `change` is ok for checkbox and radios.

timvandermeij (approved):

timvandermeij:
This works fine now; thanks!

PR #12177, opened 2020-08-05
Support comb textfields for printing

calixteman:


src/core/annotation.js
```diff
@@ -987,6 +987,22 @@ class WidgetAnnotation extends Annotation {
     const vPadding = defaultPadding + Math.abs(descent) * fontSize;
     const defaultAppearance = this.data.defaultAppearance;
 
+    if (this.data.comb) {
```
timvandermeij, line 990: Since `comb` is a specific property of text widget annotations (and not other widget annotations), this code should live in the `TextWidgetAnnotation` class instead.

src/core/annotation.js
```diff
@@ -987,6 +987,22 @@ class WidgetAnnotation extends Annotation {
...
+      const combWidth = (totalWidth / this.data.maxLen).toFixed(2);
+      let buf = `/Tx BMC q BT ${defaultAppearance} 1 0 0 1 ${hPadding} ${vPadding} Tm`;
+      let first = true;
+      for (const character of value) {
+        if (first) {
+          buf += ` (${character}) Tj`;
```
timvandermeij, line 996: To reduce the number of intermediate strings being created, we normally don't append strings to strings. Instead, we push all strings to an array and join them at the end, which is a bit more efficient. Can we do that here as well? Look at https://github.com/mozilla/pdf.js/blob/50bc4a18e8c564753365d927d5ec6a6d2cce3072/src/core/core_utils.js#L113-L131 for the pattern we use.

Moreover, this will also fix the line `let buf = ...` above since that is longer than the 80 character limit. If we split that large string into two strings we can simply push those into the array so it fits with the character limit again (I'm surprised ESLint doesn't complain about this already...).

test/unit/annotation_spec.js
```diff
@@ -1673,6 +1673,42 @@ describe("annotation", function () {
...
+          annotationStorage[id] = "aaaaaaaa";
+          return annotation.getAppearance(
+            partialEvaluator,
+            task,
+            annotationStorage
+          );
+        }, done.fail)
+        .then(appearance => {
+          expect(appearance).toEqual(
+            "/Tx BMC q BT /Helv 5 Tf 1 0 0 1 2 2 Tm (a) Tj 8.00 0 Td (a) Tj 8.00 0 Td (a) Tj 8.00 0 Td (a) Tj 8.00 0 Td (a) Tj 8.00 0 Td (a) Tj 8.00 0 Td (a) Tj 8.00 0 Td (a) Tj ET Q EMC"
```
timvandermeij, line 1707: This line also exceeds the 80 character limit. Please split it up by concatenating smaller substrings.

timvandermeij (commented):
Looks good with these comments addressed. (I only checked the last commit for this PR since the rest is in other PRs.)

src/core/annotation.js
```diff
@@ -987,6 +987,22 @@ class WidgetAnnotation extends Annotation {
     const vPadding = defaultPadding + Math.abs(descent) * fontSize;
     const defaultAppearance = this.data.defaultAppearance;
 
+    if (this.data.comb) {
+      const combWidth = (totalWidth / this.data.maxLen).toFixed(2);
```
timvandermeij, line 991: To prevent this method from being too big, I would suggest to move the comb handling code to a private method named `_renderCombText` similar to what you did for the multiline strings.

timvandermeij:
The combs work fine, it's only the multiline text fields that don't work. For completeness I'm attaching the document where the comb is rendered, but the multiline is not:

[out.pdf](https://github.com/mozilla/pdf.js/files/5045363/out.pdf)

timvandermeij:
This requires a rebase after the multiline text PR.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #12179, opened 2020-08-06
Add support for `Object.fromEntries`

Snuffleupagus:
This provides a simpler way of creating an `Object` from e.g. a `Map`, without having to manually iterate over it.
Please see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries

### DIFF
--- a/src/display/metadata.js
+++ b/src/display/metadata.js
@@ -118,11 +118,7 @@ class Metadata {
   }
 
   getAll() {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
That's a nice feature; thanks!

PR #12182, opened 2020-08-07
Fix the type of PDFDocumentLoadingTask.destroy.

tamuratak:
Fix the type of `PDFDocumentLoadingTask.destroy`.


timvandermeij (approved):

timvandermeij:
Thank you for helping out with the types!

PR #12186, opened 2020-08-07
Fix bad truetype loca tables.

brendandahl:
Some fonts have loca tables that aren't sorted or use 0 as an offset to
signal a missing glyph. This fixes the bad loca tables by sorting them
and then rewriting the loca table and potentially re-ordering the glyf
table to match.

Fixes #11131, #11561
and
https://bugzilla.mozilla.org/show_bug.cgi?id=1650302
https://bugzilla.mozilla.org/show_bug.cgi?id=846014
https://bugzilla.mozilla.org/show_bug.cgi?id=844092

PR #12187, opened 2020-08-08
Add a test-case for issue 4398

Snuffleupagus:
Issue #4398 was fixed by PR #4437, however a test-case wasn't included as far as I can tell. Given that PR #12186 is now in the process of re-factoring that code, adding a test-case cannot hurt as far as I'm concerned.


timvandermeij (approved):

timvandermeij:
Good idea; thanks!

PR #8081, opened 2017-02-19
iOS Chrome: Fix broken download button

archive64:
Reproduction Steps
---

1. Open Chrome on iOS (I'm using an iPhone 6, iOS 10.2.1, Chrome 56.0.2924.79)
2. Open https://mozilla.github.io/pdf.js/web/viewer.html (I tested on commit cfaa621a05cb5360f6834cf96a16c44c1a9aed5b)
3. Open the dropdown menu and click "Download."
4. Nothing happens.

On iOS Safari, however, the download button works fine.

Cause
---

It appears that `URL.createObjectURL` is unreliable in iOS Chrome. Going to the blob URL doesn't do anything. Opening the blob URL in a new window shows a blank page. See https://groups.google.com/a/chromium.org/forum/#!topic/chromium-html5/RKQ0ZJIj7c4 for more details.

Solution
---

Add iOS Chrome to the list of browsers that set `disableCreateObjectURL` in `compatibility.js`. The `downloadUrl` fallback works fine.

I'm not sure if this is the appropriate place to make the change, since I had to rearrange some code in `download_manager.js`. Please let me know if I should make any different changes.

PR #12191, opened 2020-08-10
Remove the `disableCreateObjectURL` option from `web/app_options.js`

Snuffleupagus:
Prior to PR #11601, the `disableCreateObjectURL` option was present on `getDocument` in the API, since it was (potentially) used when decoding JPEG images natively in the browser. Hence setting this option, which was done automatically using compatibility-code, were in some browsers necessary in order for e.g. JPEG images to be correctly rendered.

The downside of the `disableCreateObjectURL` option is that memory usage increases significantly, since we're forced to build and use `data:` URIs (rather than `blob:` URLs).
However, at this point in time the `disableCreateObjectURL` option is only necessary for *some* (non-essential) functionality in the default viewer; in particular:
 - The openfile functionality, used only when manually opening a new file in the default viewer.
 - The download functionality, used when downloading either the PDF document itself or its attached files (if such exists).
 - The print functionality, in the generic `PDFPrintService` implementation.

Hence neither the general PDF.js library, nor the *basic* functionality of the default viewer, depends on the `disableCreateObjectURL` option any more; which is why I'm thus proposing that we remove the option since using it is a performance footgun.

*Please note:* To not outright break currently "supported" browsers, which lack proper `URL.createObjectURL` support, this patch purposely keeps the compatibility-code to explicitly disable `URL.createObjectURL` usage *only* for browsers which are known to not work correctly.[1]

While it's certainly possible that there's additional, likely older, browsers with broken `URL.createObjectURL` support, the last time that these types of problems were reported was over *three* years ago.[2]
Hence in the *very* unlikely event that additional problems occur, as a result of these changes, we can either add a new case in the compatibility-code or simply declare the affected browser as unsupported.

---
[1] Which are IE11 (see issue #3977), and Google Chrome on iOS (see PR #8081).

[2] Given that `URL.createObjectURL` is used by default, you'd really expect more reports if these problems were widespread.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -72,6 +72,7 @@ import { PDFThumbnailViewer } from "./pdf_thumbnail_viewer.js";
 import { PDFViewer } from "./pdf_viewer.js";
 import { SecondaryToolbar } from "./secondary_toolbar.js";
 import { Toolbar } from "./toolbar.js";
+import { viewerCompatibilityParams } from "./viewer_compatibility.js";
 import { ViewHistory } from "./view_history.js";
 
 const DEFAULT_SCALE_DELTA = 1.1;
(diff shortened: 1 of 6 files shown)


timvandermeij (approved):

timvandermeij:
I agree with the points that you mentioned. Thanks!

PR #12192, opened 2020-08-10
A couple of (small) tweaks of the `AnnotationStorage` (PR 12173 follow-up)

Snuffleupagus:
- Initialize the `AnnotationStorage`-instance, on `PDFDocumentProxy`, lazily.
 - Change the `AnnotationStorage` to use a `Map` internally, rather than a regular Object (simplifies the following points).
 - Let `AnnotationStorage.getAll` return `null` when there's no data stored, to avoid unnecessary parsing on the worker-thread. This ought to "just work", since the worker-thread code *should* already handle the `!annotationStorage` case everywhere.
 - Add a new `AnnotationStorage.size` getter, to be able to easily tell if there's any data stored.

### DIFF
--- a/src/display/annotation_storage.js
+++ b/src/display/annotation_storage.js
@@ -18,7 +18,7 @@
  */
 class AnnotationStorage {
   constructor() {
-    this._storage = Object.create(null);
+    this._storage = new Map();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #12193, opened 2020-08-10
Set `direction: ltr;` on the canvases used during printing (bug 1335712)

Snuffleupagus:
This essentially mirrors the CSS rules used for the viewer itself, see `web/pdf_viewer.css`, and according to my very quick tests it seems to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1335712

/cc @brendandahl


timvandermeij (approved):

timvandermeij:
Looks good to me; thanks!

PR #12199, opened 2020-08-11
Fix #12198 by setting a non-transparent background

### LINKED ISSUE #12198: The zoom popup in the viewer shows a weird border at the top and bottom on macOS
87592901-e8498280-c6ea-11ea-8f3e-ac8825149edd

Configuration:
- Seen on Firefox Nightly 81.0a1 (2020-08-01)

1. Use macOS
2. Click on the zoom select at the top of the viewer

msujaws:
The transparent background causes Firefox to show the rounded popup border on macOS. Because the popup is transparent and also has a background-color that is applied to the option children, the transparent background gets evaluated to the user-agent color, which is a shade of grey.

Setting an explicit color here is the simplest route to fix this. I used rgba(82,82,82,.95) as it was one of the stop colors of the toolbar gradient.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -713,7 +713,7 @@ html[dir='rtl'] .dropdownToolbarButton::after {
   margin: 0;
   padding: 3px 2px 2px;
   border: none;
-  background: rgba(0,0,0,0); /* Opera does not support 'transparent' <select> background */
+  background-color: rgba(82,82,82,.95);
 }
 
 .dropdownToolbarButton > select > option {


timvandermeij (commented):
Looks good to me, but I'd like to have someone check how this looks on Windows since I'm on Linux and Firefox lets the OS render the dropdowns there. @Snuffleupagus Could you perhaps check if the preview above renders fine on Windows?

Snuffleupagus:
My general question would be if it's even worth fixing/spending time on this *separately*, given that PR #11077 now looks fairly close to being done!?

Snuffleupagus:
My general question would be if it's even worth fixing/spending time on this *separately*, given that PR #11077 now looks fairly close to being done!?

---

Looking at the actual preview: It appears that this has removed any trace of a `:hover` effect, on Windows at least, which does look a bit weird.

brendandahl:
We want to land #11077 for the current nightly release of Firefox, so I don't think this is worth pursuing.

msujaws:
This is how it looks on Windows,
image

msujaws:
> 
> 
> We want to land #11077 for the current nightly release of Firefox, so I don't think this is worth pursuing.

Okay, I'll wait for #11077 to land and then see if this same fix is needed there. I was told that it might which is why my attention is here.

timvandermeij:
#11077 just landed. Could you check if any change is still necessary and if so, fixup this patch?

timvandermeij:
Closing since the new design should have fixed this (it's also using `background-color` with a non-transparent color now).

PR #12201, opened 2020-08-12
Enable renderInteractiveForms by default.

brendandahl:
The forms seem far enough along to enable by default now.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1415,8 +1415,14 @@ const PDFViewerApplication = {
       this.setTitle(contentDispositionFilename);
     }
 
-    if (info.IsAcroFormPresent) {
-      console.warn("Warning: AcroForm/XFA is not supported");
+    if (info.IsXFAPresent) {
+      console.warn("Warning: XFA is not supported");
+      this._delayedFallback(UNSUPPORTED_FEATURES.forms);
+    } else if (
+      info.IsAcroFormPresent &&
+      !this.pdfViewer.renderInteractiveForms
+    ) {
+      console.warn("Warning: AcroForm support is not enabled");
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
If this is enabled by default now, it does seem *slightly* strange to still warn/fallback in that case :-)

Hence, changing this line https://github.com/mozilla/pdf.js/blob/7edc5cb79fe829d45fed85e2b4b6d8594522cc10/web/app.js#L1418 to the following may be reasonable?
```javascript
if (info.IsAcroFormPresent && !this.pdfViewer.renderInteractiveForms) {
```
(Rather than checking the `AppOptions` here, it probably makes sense to check what the viewer is *actually* using.)

brendandahl:
I think we'll also need to check if the the form is XFA. Unfortunately, it seems we'll still show the warning bar in some cases where things work fine with just acroform support, but the PDF has XFA and acrofroms e.g. f1040.pdf.

web/app.js
```diff
@@ -1415,8 +1415,14 @@ const PDFViewerApplication = {
...
-    if (info.IsAcroFormPresent) {
-      console.warn("Warning: AcroForm/XFA is not supported");
+    if (info.IsXFAPresent) {
+      console.warn("Warning: XFA is not supported");
+      this._delayedFallback(UNSUPPORTED_FEATURES.forms);
+    } else if (
+      info.IsAcroFormPresent &&
+      !this.pdfViewer.renderInteractiveForms
+    ) {
+      console.warn("Warning: AcroForm support is not eabled");
```
Snuffleupagus, line 1425: Typo: `eabled` -> `enabled`

Snuffleupagus (approved):
r=me, with the typo fixed.

PR #12203, opened 2020-08-12
Bug 1392361 - Fix zooming sensitivity on macOS

### LINKED ISSUE #8802: Zoom is too sensitive on OS X

staktrace:
The original code would get a long sequence of miniscule "tick" values while
pinch-zooming, and each tick value would cause a 1.1x zoom. So even the smallest
pinch gesture on a trackpad would cause high amounts of zoom. This patch
accumulates the wheel deltas until they reach an integer threshold (with a
tweak of the scaling factor to make it feel more natural) at which point it
triggers the zoom based on the integer component of the accumulated delta. The
fractional part is retained in the accumulator.

Fixes #8802

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -26,7 +26,7 @@ import {
   MAX_SCALE,
   MIN_SCALE,
   noContextMenuHandler,
-  normalizeWheelEventDelta,
+  normalizeWheelEventDirection,
   parseQueryString,
   PresentationModeState,
   ProgressBar,
@@ -237,6 +237,7 @@ const PDFViewerApplication = {
   contentDispositionFilename: null,
   triggerDelayedFallback: null,
   _saveInProgress: false,
+  _wheelUnusedTicks: 0,
 
   // Called once when the document is loaded.
   async initialize(appConfig) {
@@ -1835,6 +1836,22 @@ const PDFViewerApplication = {
     _boundEvents.windowBeforePrint = null;
     _boundEvents.windowAfterPrint = null;
   },
+
+  accumulateWheelTicks(ticks) {
+    // If the scroll direction changed, reset the accumulated wheel ticks.
+    if (
+      (this._wheelUnusedTicks > 0 && ticks < 0) ||
+      (this._wheelUnusedTicks < 0 && ticks > 0)
+    ) {
+      this._wheelUnusedTicks = 0;
+    }
+    this._wheelUnusedTicks += ticks;
+    const wholeTicks =
+      Math.sign(this._wheelUnusedTicks) *
+      Math.floor(Math.abs(this._wheelUnusedTicks));
+    this._wheelUnusedTicks -= wholeTicks;
+    return wholeTicks;
+  },
 };
 
 let validateFileURL;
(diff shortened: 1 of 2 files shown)


staktrace:
I tested this on Firefox Nightly with the following inputs:
- macOS with trackpad pinch, with apz.allow_zooming=true,false
- macOS with ctrl+two-finger-scroll, with apz.allow_zooming=true,false
- Windows trackpad two finger pinch (uses direct manipulation), with apz.allow_zooming=true
- Windows trackpad two finger pinch with apz.allow_zooming=false (does nothing)
- Windows trackpad ctrl+scroll, with apz.allow_zooming=true,false

Also tested on Chrome on macOS with trackpad pinch and ctrl+scroll.

In all cases the behaviour seemed reasonable.

staktrace:
There's a deployed version at https://mozilla.staktrace.com/tmp/fixzoom/generic/web/viewer.html if you want to try yourself.

Snuffleupagus:
Unfortunately this performs significantly *worse* when using a mouse with a scrollwheel, in my case a Logitech G403, on Windows.

I've testing using <kbd>Ctrl</kbd>+<kbd>scrollwheel up</kbd>, one increment at a time, and get the following sequence of zoom values:
 - With the PDF.js `master` branch: `auto (= 125%)` -> `140%` -> `160%` -> `180%` -> `200%` -> ...
   This is basically equivalent to the values I get when zooming using only the keyboard, i.e. with <kbd>Ctrl</kbd>+<kbd>+</kbd>.
 - With your demo:  `auto (= 125%)` -> `180%` -> `250%` -> `350%` -> `480%` -> ...
   This unfortunately renders mousewheel zooming more-or-less unusable as-is.

staktrace:
Indeed, I can reproduce. I should have tried that before submitting the patch, sorry. Fix coming.

staktrace:
Hm so actually this is because Windows defaults to three lines per mousewheel roll on line-based devices. On macOS it seems to default to 1. I guess for line-based devices we should consider one tick per event or something, since line-based devices are generally wheels with clicky wheels and one click would nicely correspond to one zoom step.

staktrace:
Patch updated. New version is deployed at https://mozilla.staktrace.com/tmp/fixzoom2/generic/web/viewer.html - please give it a whirl and let me know how it feels.

Snuffleupagus:
> Patch updated. New version is deployed at https://mozilla.staktrace.com/tmp/fixzoom2/generic/web/viewer.html - please give it a whirl and let me know how it feels.

That's looks/feels *much* better, thank you!

/botio-linux preview

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #12207, opened 2020-08-13
wip - add support for defaultAppearance fonts in text areas 

escapewindow:
This works! If I override the `DA` with something like `/CourierNewPSMT 8 Tf 0 g` in `i9.pdf`, then the text I type in a textarea shows up as Courier instead of Helvetica.

I do have some questions.

- I went the evaluator route because I wanted to load the font in our font cache and get a `fontRefName` back. I wasn't sure if we also wanted to preserve the `OperatorList` ?
  - Because I'm using the evaluator, I need a `handler` and `task`. I added calls to both `Page.getOperatorList` and `Page.extractTextContent` to reuse those handlers and tasks. I'm not sure if there's a more elegant way.
- I'm currently loading the `defaultAppearanceData` once per page, which might be good if we need the `OperatorList` or populate the font cache per page. Otherwise we might be able to load it once per `PDFDocument`. (I'm trying the `PDFDocument` solution in https://github.com/escapewindow/pdf.js/commits/page-acrodict-wip and it breaks the entire form badly. It's quite possible I'm doing something wrong, however.)
- I'm unable to do anything with `fontColor`. I have an rgb color, but setting `style.color` doesn't seem to do anything. If I `console.log(style.color)` before and after, it's unchanged. Maybe it's read-only (or maybe I did something wrong again)?

Most of this is new code, but it's based on @dmitryskey's code from #8625. I commited the rebased patch with @dmitryskey as the author, so we both show up.

What do you think @brendandahl ?

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -50,7 +50,7 @@ class AnnotationFactory {
    * @returns {Promise} A promise that is resolved with an {Annotation}
    *   instance.
    */
-  static create(xref, ref, pdfManager, idFactory) {
+  static create(xref, ref, pdfManager, idFactory, defaultAppearanceData) {
     return pdfManager.ensureCatalog("acroForm").then(acroForm => {
       return pdfManager.ensure(this, "_create", [
         xref,
@@ -58,6 +58,7 @@ class AnnotationFactory {
         pdfManager,
         idFactory,
         acroForm,
+        defaultAppearanceData,
       ]);
     });
   }
@@ -65,7 +66,14 @@ class AnnotationFactory {
   /**
    * @private
    */
-  static _create(xref, ref, pdfManager, idFactory, acroForm) {
+  static _create(
+    xref,
+    ref,
+    pdfManager,
+    idFactory,
+    acroForm,
+    defaultAppearanceData
+  ) {
     const dict = xref.fetchIfRef(ref);
     if (!isDict(dict)) {
       return undefined;
@@ -86,6 +94,10 @@ class AnnotationFactory {
       id,
       pdfManager,
       acroForm: acroForm instanceof Dict ? acroForm : Dict.empty,
+      defaultAppearanceData:
+        defaultAppearanceData instanceof Dict
+          ? defaultAppearanceData
+          : Dict.empty,
     };
 
     switch (subtype) {
@@ -242,6 +254,7 @@ class Annotation {
   constructor(params) {
     const dict = params.dict;
 
+    this.defaultAppearanceData = params.defaultAppearanceData;
     this.setContents(dict.get("Contents"));
     this.setModificationDate(dict.get("M"));
     this.setFlags(dict.get("F"));
@@ -1285,6 +1298,14 @@ class WidgetAnnotation extends Annotation {
 class TextWidgetAnnotation extends WidgetAnnotation {
   constructor(params) {
     super(params);
+    if (params.defaultAppearanceData) {
+      const keys = ["fontRefName", "fontName", "fontSize", "fontColor"];
+      for (const key of keys) {
+        if (key in params.defaultAppearanceData && !(key in this.data)) {
+          this.data[key] = params.defaultAppearanceData[key];
+        }
+      }
+    }
 
     this._hasText = true;
 
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -48,7 +48,12 @@ import {
   XRefEntryException,
   XRefParseException,
 } from "./core_utils.js";
-import { NullStream, Stream, StreamsSequenceStream } from "./stream.js";
+import {
+  NullStream,
+  Stream,
+  StreamsSequenceStream,
+  StringStream,
+} from "./stream.js";
 import { AnnotationFactory } from "./annotation.js";
 import { calculateMD5 } from "./crypto.js";
 import { Linearization } from "./parser.js";
@@ -76,6 +81,8 @@ class Page {
     fontCache,
     builtInCMapCache,
     globalImageCache,
+    defaultAppearance,
+    defaultResources,
   }) {
     this.pdfManager = pdfManager;
     this.pageIndex = pageIndex;
@@ -87,6 +94,9 @@ class Page {
     this.globalImageCache = globalImageCache;
     this.evaluatorOptions = pdfManager.evaluatorOptions;
     this.resourcesPromise = null;
+    this.defaultAppearancePromise = null;
+    this.defaultAppearance = defaultAppearance;
+    this.defaultResources = defaultResources;
 
     const idCounters = {
       obj: 0,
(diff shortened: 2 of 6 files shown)


escapewindow:
I do have a text-only parsing branch [here](https://github.com/mozilla/pdf.js/compare/master...escapewindow:defaultAppearance), if the evaluator solution is too heavy handed.

escapewindow:
Trying to parse the DA of each widget [here](https://github.com/escapewindow/pdf.js/commits/page-acrodict2); that's not currently working.

escapewindow:
This is rebased and working again.

@timvandermeij @Snuffleupagus @brendandahl What are your thoughts here? We no longer have the additional `AnnotationWorkerTask` from #8625, but we're only parsing the main document DA and DR and not parsing each widget's DA. As mentioned above, I've attempted other approaches that try to combine the approaches. I haven't yet been successful. Other options might include:

- Drop the evaluator `getAcroformDefaultAppearanceData` in favor of a text parser like [this](https://github.com/mozilla/pdf.js/compare/master...escapewindow:defaultAppearance),
- add the evaluator call to each widget
- go with @dmitryskey's approach of adding an `AnnotationWorkerTask`.

src/core/annotation.js
```diff
@@ -242,6 +254,7 @@ class Annotation {
   constructor(params) {
     const dict = params.dict;
 
+    this.defaultAppearanceData = params.defaultAppearanceData;
```
timvandermeij, line 257: I think this can be removed since the only place that uses `defaultAppearanceData` is in the `TextWidgetAnnotation` class and it fetches it from the `params`, not from this member variable.

src/core/annotation.js
```diff
@@ -1285,6 +1298,14 @@ class WidgetAnnotation extends Annotation {
 class TextWidgetAnnotation extends WidgetAnnotation {
   constructor(params) {
     super(params);
+    if (params.defaultAppearanceData) {
```
timvandermeij, line 1301: I think you already indicated this, but individual annotations might have their own appearance, so the global one should only be used as a fallback in that case.

src/core/document.js
```diff
@@ -125,11 +135,13 @@ class Page {
...
-    return shadow(
-      this,
-      "resources",
-      this._getInheritableProperty("Resources") || Dict.empty
-    );
+
+    const resources = this._getInheritableProperty("Resources") || Dict.empty;
+    const acroForm = this.pdfManager.pdfDocument.catalog.acroForm;
+    if (acroForm && acroForm.get("DR")) {
+      resources.defaultResources = acroForm.get("DR");
```
timvandermeij, line 142: Hm, I wouldn't expect the normal resources dictionary to be altered with a non-standard attribute. Shouldn't we merge the dictionaries instead? Perhaps the just merged https://github.com/mozilla/pdf.js/pull/12361/files can serve as inspiration.

src/core/document.js
```diff
@@ -375,6 +397,24 @@ class Page {
...
 
+  handleDefaultAppearance(partialEvaluator, task) {
+    if (this.defaultAppearanceData) {
+      return new Promise();
+    }
+
+    const opList = new OperatorList();
+
+    const appearanceStream = new StringStream(this.defaultAppearance);
+    this.defaultAppearanceData = Dict.empty;
```
timvandermeij, line 408: We should use `new Dict` here since we should never mutate `Dict.empty` as it's always an empty dictionary (not sure if that is even allowed now, and if so we should probably make that impossible).

src/core/document.js
```diff
@@ -552,7 +602,7 @@ class PDFDocument {
     this.stream = stream;
     this.xref = new XRef(stream, pdfManager);
     this._pagePromises = [];
-    this._version = null;
```
timvandermeij, line 555: `this._version` should stay.

src/core/evaluator.js
```diff
@@ -2631,6 +2631,131 @@ class PartialEvaluator {
...
+    if (!operatorList) {
+      throw new Error(
+        'getAcroformDefaultAppearance: missing "operatorList" parameter'
+      );
+    }
+
+    var self = this;
+    var xref = this.xref;
+
+    var stateManager = new StateManager(initialState);
```
timvandermeij, line 2653: Change `var` to `const`/`let` in this method.

src/display/annotation_layer.js
```diff
@@ -480,22 +481,28 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         element.classList.add("comb");
         element.style.letterSpacing = `calc(${combWidth}px - 1ch)`;
       }
+      if (
```
timvandermeij, line 484: I don't think we actually want this in edit mode. Edit mode has other styles to clearly differentiate it from display mode. In display mode we do want this, and that's why it only lived in the `else` clause. The objective of this patch is also to make the final rendering on display/print/save in the right style.

src/display/annotation_layer.js
```diff
@@ -515,7 +522,9 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
   _setTextStyle(element, font) {
     // TODO: This duplicates some of the logic in CanvasGraphics.setFont().
     const style = element.style;
-    style.fontSize = `${this.data.fontSize}px`;
+    if (this.data.fontSize) {
```
timvandermeij, line 525: Shouldn't this always be set?

src/display/annotation_layer.js
```diff
@@ -535,6 +544,7 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
     const fontFamily = font.loadedName ? `"${font.loadedName}", ` : "";
     const fallbackName = font.fallbackName || "Helvetica, sans-serif";
     style.fontFamily = fontFamily + fallbackName;
+    // XXX revisit fontColor. Setting style.color doesn't stick.
```
timvandermeij, line 547: Hm, I think this worked in the original patch. Maybe the values need to be converted to RGB first?

test/unit/annotation_spec.js
```diff
@@ -1863,6 +1863,36 @@ describe("annotation", function () {
           done();
         }, done.fail);
     });
+
+    it("should expose document defaultAppearance", function (done) {
```
timvandermeij, line 1867: `"should use the default appearance of the document"`

test/unit/annotation_spec.js
```diff
@@ -1863,6 +1863,36 @@ describe("annotation", function () {
...
+      const defaultAppearanceData = Dict.empty;
```
timvandermeij, line 1868: Same here for `new Dict` instead of mutating `Dict.empty`.

timvandermeij (commented):
First of all, sorry for the review delay. It has been quite busy lately with all the AcroForm work, so I only just now had time to take a look at this. I have added some initial review comments. I also think/hope some things can be simplified; preferably I'd have the logic mostly in the annotation code files and not in the document file, but it will require a bit more time to check if that is possible.

timvandermeij:
Closing since this PR is superseded by #12831 (default appearance parser) and #12828 (using the default appearance data in the display layer).

PR #12210, opened 2020-08-13
Keep the original class/function names when minifying code (issue 12209)

### LINKED ISSUE #11941: minified-es5 fails to "load" / correctly handle defective pdf-files

### LINKED ISSUE #12209: Minified PDFJS worker unable to properly serialize exceptions
I primarily use [pdfjs-dist](https://github.com/mozilla/pdfjs-dist) and have up until recently loaded the [minified worker](https://github.com/mozilla/pdfjs-dist/blob/master/build/pdf.worker.min.js).

When upgrading from version 2.1.266 to version 2.4.456, I noticed that the errors thrown from `getDocument` were `undefined` in the event of a bad URL, where they used to come in as `MissingPDFException`, for example. I suspect the problem is also present on 2.5.207 and was introduced in 2.3.200.

I believe the source of the problem is due to the fact that reflection is now being used by `BaseException` and the code I'm using is minified, so the names of the classes are being changed. The problem was likely introduced in this commit with this line of code:

https://github.com/mozilla/pdf.js/commit/5d93fda4f2f6b4b337b4e334b799c227ce4aeedd#diff-8d2b578d58d4d8c8a2975020fbbc84d9R410

This causes problems when trying to deserialize the exception when passed from the worker as this switch statement results in `undefined`: https://github.com/mozilla/pdf.js/blob/1a6816ba98fd437024f6f0ddfa78e6d4dbf11371/src/display/api.js#L2255-L2281

In the case of the minified file one can see that `ex` will come through as an object that looks like:

```
{ 
  message: "Failed to fetch",
  name: "e",
  details: "TypeError: Failed to fetch"
}
```

Which does not match the string values in the switch, which will never be minified themselves because they are string literals.

Snuffleupagus:
While this will obviously increase the size of the output of `gulp minified`/`gulp minified-es5` *slightly*, the resulting files are still a lot smaller than the non-minified builds.

See https://github.com/terser/terser#minify-options for information about various Terser options.

Fixes #12209
Fixes #11941

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -909,18 +909,30 @@ function parseMinified(dir) {
   console.log("### Minifying js files");
 
   var Terser = require("terser");
-  // V8 chokes on very long sequences. Works around that.
-  var optsForHugeFile = { compress: { sequences: false } };
+  var options = {
+    compress: {
+      // V8 chokes on very long sequences, work around that.
+      sequences: false,
+    },
+    keep_classnames: true,
+    keep_fnames: true,
+  };
 
-  fs.writeFileSync(dir + "/web/pdf.viewer.js", Terser.minify(viewerFiles).code);
-  fs.writeFileSync(dir + "/build/pdf.min.js", Terser.minify(pdfFile).code);
+  fs.writeFileSync(
+    dir + "/web/pdf.viewer.js",
+    Terser.minify(viewerFiles, options).code
+  );
+  fs.writeFileSync(
+    dir + "/build/pdf.min.js",
+    Terser.minify(pdfFile, options).code
+  );
   fs.writeFileSync(
     dir + "/build/pdf.worker.min.js",
-    Terser.minify(pdfWorkerFile, optsForHugeFile).code
+    Terser.minify(pdfWorkerFile, options).code
   );
   fs.writeFileSync(
     dir + "image_decoders/pdf.image_decoders.min.js",
-    Terser.minify(pdfImageDecodersFile).code
+    Terser.minify(pdfImageDecodersFile, options).code
   );
 
   console.log();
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Nice work!

PR #12212, opened 2020-08-13
Bug 1643508 - Disable touch-based pinch zooming on pdf.js.

staktrace:
Currently using a touchscreen with pdf.js doesn't work so well. In Firefox,
with apz.allow_zooming = false (default on current release/beta), it does a
reflow zoom which makes the UI elements bigger. And with apz.allow_zooming = true
(default on current Firefox nightly), or in Chrome, it does a smooth pinch-zoom
but that also scales up the entire UI. Neither of these is a particularly good
experience, so this patch just disables any multi-touch gestures. Touch-based
panning (which involves a single touch point) is left unaffected.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1760,6 +1760,9 @@ const PDFViewerApplication = {
 
     window.addEventListener("visibilitychange", webViewerVisibilityChange);
     window.addEventListener("wheel", webViewerWheel, { passive: false });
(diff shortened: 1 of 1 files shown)


staktrace:
Currently deployed to https://mozilla.staktrace.com/tmp/fixtouch/generic/web/viewer.html

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12229, opened 2020-08-17
Dispatch event when annotations have finished rendering.

brendandahl:
This is needed for some smoke tests in mozilla central for testing forms
in pdf.js.

Note: AnnotationLayerBuilder.render() doesn't really need to be async, but
we're talking of making the annotation's render functions async, so this
will make that switch easier.

### DIFF
--- a/web/annotation_layer_builder.js
+++ b/web/annotation_layer_builder.js
@@ -60,9 +60,11 @@ class AnnotationLayerBuilder {
   /**
    * @param {PageViewport} viewport
    * @param {string} intent (default value is 'display')
+   * @returns {Promise<void>} A promise that is resolved when rendering of the
+   *   annotations is complete.
    */
   render(viewport, intent = "display") {
-    this.pdfPage.getAnnotations({ intent }).then(annotations => {
+    return this.pdfPage.getAnnotations({ intent }).then(annotations => {
       if (this._cancelled) {
         return;
       }
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -145,6 +145,24 @@ class PDFPageView {
   /**
    * @private
    */
+  async _renderAnnotationLayer() {
+    let error = null;
+    try {
+      await this.annotationLayer.render(this.viewport, "display");
+    } catch (ex) {
+      error = ex;
+    } finally {
+      this.eventBus.dispatch("annotationlayerrendered", {
+        source: this,
+        pageNumber: this.id,
+        error,
+      });
+    }
+  }
+
+  /**
+   * @private
+   */
   _resetZoomLayer(removeFromDOM = false) {
     if (!this.zoomLayer) {
       return;
@@ -384,7 +402,7 @@ class PDFPageView {
     }
 
     if (redrawAnnotations && this.annotationLayer) {
-      this.annotationLayer.render(this.viewport, "display");
+      this._renderAnnotationLayer();
     }
   }
 
@@ -542,7 +560,7 @@ class PDFPageView {
           this.l10n
         );
       }
-      this.annotationLayer.render(this.viewport, "display");
+      this._renderAnnotationLayer();
     }
     div.setAttribute("data-loaded", true);
 


web/annotation_layer_builder.js
```diff
@@ -60,9 +60,11 @@ class AnnotationLayerBuilder {
   /**
    * @param {PageViewport} viewport
    * @param {string} intent (default value is 'display')
+   * @returns {Promise} A promise resolved when rendering the annotations is
```
timvandermeij, line 63: `Promise` -> `Promise<void>`

web/annotation_layer_builder.js
```diff
@@ -60,9 +60,11 @@ class AnnotationLayerBuilder {
...
+   * complete.
```
timvandermeij, line 64: We standardized the JSDoc comments recently to have two-space indentation if the line belongs to a previous line, so `complete` should have two extra spaces before it.

web/annotation_layer_builder.js
timvandermeij, line 63: `Promise` -> `Promise<void>`

web/pdf_page_view.js
```diff
@@ -142,6 +142,25 @@ class PDFPageView {
...
 
+  /**
+   * @private
+   */
+  async _renderAnnotationLayer() {
+    let error = null;
+    try {
+      await this.annotationLayer.render(this.viewport, "display");
+      error = false;
+    } catch (e) {
```
Snuffleupagus, line 153: Nit: Let's avoid adding one-letter variable names; please use at least `ex` instead.

web/pdf_page_view.js
```diff
@@ -142,6 +142,25 @@ class PDFPageView {
...
+      error = e;
+    } finally {
+      this.eventBus.dispatch("annotationsrender", {
```
Snuffleupagus, line 156: Please call this `annotationlayerrendered` instead, since that's consistent with the event used in the text-layer.

web/pdf_page_view.js
```diff
@@ -142,6 +142,25 @@ class PDFPageView {
...
   }
 
+  /**
+   * @private
+   */
+  async _renderAnnotationLayer() {
+    let error = null;
+    try {
+      await this.annotationLayer.render(this.viewport, "display");
+      error = false;
```
timvandermeij, line 152: I think having `error` either be `null` or an actual value is more consistent with for example the API methods, so it looks better to me if this line is removed.

web/pdf_page_view.js
timvandermeij, line 156: This should probably be called `annotationsrendered` for consistency with existing events. (compare e.g., `textlayerrendered`).

web/pdf_page_view.js
Snuffleupagus, line 152: Do you actually need this?
Or in other words, why can't the default value (when no error occurred) simply be `null` instead?

web/annotation_layer_builder.js
Snuffleupagus, line 64: It feels like this sentence is missing a couple of words; how about the following instead?
```suggestion
   * @returns {Promise} A promise that is resolved when rendering of the
   *   annotations is complete.
```

Snuffleupagus (changes requested):

web/pdf_page_view.js
timvandermeij: Yes, `annotationlayerrendered` is an even better name than my `annotationsrendered` suggestion above. Let's use this instead.

timvandermeij:
Looks good with the two final comments above addressed.

web/annotation_layer_builder.js
```diff
@@ -60,9 +60,11 @@ class AnnotationLayerBuilder {
   /**
    * @param {PageViewport} viewport
    * @param {string} intent (default value is 'display')
+   * @returns {Promise<void>} A promise that is resolved when rendering of the
+   * annotations is complete.
```
timvandermeij, line 64: The two spaces before `annotations` seem to have disappeared.

timvandermeij (approved):
Looks good to me with the final comment addressed. Thanks!

timvandermeij:
Thanks!

PR #12230, opened 2020-08-17
Implement reference testing for printing

### LINKED ISSUE #12225: Implement reference testing for printing
The printing code is currently only tested manually, which introduces a regression risk. Normally the printing code more or less prints the displayed canvas directly, but now that we have form fill support, the printed canvas differs from the displayed canvas because the form data is rendered onto it. Unit tests are in place to ensure that the operator lists are correct, but it would be even better if we can have a reference test for the printed canvas too.

timvandermeij:
Fixes #12225.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -519,7 +519,8 @@ var Driver = (function DriverClosure() {
 
               // Initialize various `eq` test subtypes, see comment below.
               var renderAnnotations = false,
-                renderForms = false;
+                renderForms = false,
+                renderPrint = false;
 
               var textLayerCanvas, annotationLayerCanvas;
               var initPromise;
@@ -559,6 +560,7 @@ var Driver = (function DriverClosure() {
                 // accidentally changing the behaviour for other types of tests.
                 renderAnnotations = !!task.annotations;
                 renderForms = !!task.forms;
+                renderPrint = !!task.print;
 
                 // Render the annotation layer if necessary.
                 if (renderAnnotations || renderForms) {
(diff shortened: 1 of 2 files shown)


test/driver.js
```diff
@@ -604,6 +606,16 @@ var Driver = (function DriverClosure() {
...
               };
+              if (renderPrint) {
+                const annotationStorage = task.annotationStorage;
+                const docAnnotationStorage = task.pdfDoc.annotationStorage;
+                for (const [key, value] of Object.entries(annotationStorage)) {
+                  docAnnotationStorage.setValue(key, value);
+                }
+                renderContext.annotationStorage = docAnnotationStorage;
+                renderContext.intent = "print";
+              }
```
Snuffleupagus, line 617: Would there also, generally, be a good idea to be able to test the "print" intent code-paths without the need for an `annotationStorage` to be present?

If so, it may perhaps make sense to do something like the below!?
```suggestion
              if (renderPrint) {
                const annotationStorage = task.annotationStorage;
                if (annotationStorage) {
                  const docAnnotationStorage = task.pdfDoc.annotationStorage;
                  for (const [key, value] of Object.entries(annotationStorage)) {
                    docAnnotationStorage.setValue(key, value);
                  }
                  renderContext.annotationStorage = docAnnotationStorage;
                }
                renderContext.intent = "print";
              }
```

timvandermeij: I thought about this, but normally the print intent would be equal to the display intent, so it's not useful in that case. Only in the case of AcroForms the print rendering differs from the display rendering because the operator list is altered according to the form data present in the annotation storage. Therefore, for now I thought it would be best to let printing and annotation storage go hand-in-hand, and if we find a use case for leaving out the annotation storage that can simply be changed later on.

timvandermeij:
Note to self: update https://github.com/mozilla/pdf.js/wiki/Contributing after this patch lands and include any missing types (`forms`).

test/driver.js
Snuffleupagus: There's a couple of cases that come to mind, for example:
 - Any "regular" Annotation may be printable, but not viewable, hence the intent may even matter for simple documents.
 - Some optional content groups can be marked as printing-only, although we don't yet support that.

Generally speaking, there's probably a lot more things that can be printing-only and where being able to test with just `"print": true` makes sense.
Hence I'd still recommend making these changes here, since unless I'm missing something it ought to be almost trivial and avoids churn later on :-)

Snuffleupagus (approved):
Looks good to me, modulo one comment; thanks for doing this!

test/driver.js
timvandermeij: Ah, yes. I forgot about the printable-only annotations and didn't know about the printable-only optional content groups. In that case it's fair to not require the annotation storage, so I'll make the change you suggested. Thanks!

Issue #12233, opened 2020-08-18
Checkmarks are not shown in Print Preview and are not printed in pdf


bogdanmaris:
**Affected versions**
* 81.0a1 (2020-08-18)

**Affected platforms**
* Windows 10 64bit
* macOS 10.15
* Ubuntu 18.04 64bit

**Steps to reproduce**
pdfjs.renderInteractiveForms - true

1. Launch Firefox
2. Access a PDF form with fillable content (eg: http://foersom.com/net/HowTo/data/OoPdfFormExample.pdf)
3. Tick some checkboxes
4. Enter some text and some numbers
5. Enter Print Preview
6. Print to pdf

**Expected result**
* All the input is saved in Print Preview as well and printed in PDF.

**Actual result**
* Checkmarks are not saved in Print Preview and are not printed on file.

**Additional notes**
* S2 as suggested severity, due to data loss.

bogdanmaris:
Hmm, this looks like it has something to do with that particular PDF, using https://www.irs.gov/pub/irs-pdf/f1040.pdf the checkmarks are saved both in print preview and printed to pdf

timvandermeij:
I have found the cause and am working on a fix. The problem is that the value given to `_renderText` is `undefined`, while it _should_ always be a string. It works if all fields are explicitly changed, but not if some are not changed. This is a strange edge case, but thanks for finding it!

timvandermeij:
The problem is with the choice widgets where the field value never matches the display value because it's encoded differently (contains the UTF-16 BOM and other characters). This requires two fixes:

- if no field value is available (the first few choice widgets), put the first value (default) in the annotation storage so it's never `undefined`;
- if a field value is available, make sure it's decoded first before comparing.

Issue #12202, opened 2020-08-12
Warn the user if navigating away from form with un-saved data


brendandahl:
If a user has changed an acroform and they attempt to navigate away we should warn them that they may loose unsaved data.

PR #12241, opened 2020-08-18
confirm if leaving a modified form without saving

escapewindow:
There may be a nicer way to track form modification; not sure.

### DIFF
--- a/src/display/annotation_storage.js
+++ b/src/display/annotation_storage.js
@@ -19,6 +19,14 @@
 class AnnotationStorage {
   constructor() {
     this._storage = new Map();
+    this._modified = false;
+
+    // Callbacks to signal when the modification state is set or reset.
+    // This is used by the viewer to only bind on `beforeunload` if forms
+    // are actually edited to prevent doing so unconditionally since that
+    // can have undesirable efffects.
+    this.onSetModified = null;
+    this.onResetModified = null;
   }
 
   /**
@@ -49,6 +57,9 @@ class AnnotationStorage {
    * @param {Object} value
    */
   setValue(key, value) {
+    if (this._storage.get(key) !== value) {
+      this.setModified();
+    }
     this._storage.set(key, value);
   }
 
@@ -61,6 +72,24 @@ class AnnotationStorage {
 
   get size() {
     return this._storage.size;
+  }
+
+  setModified() {
+    if (!this._modified) {
+      this._modified = true;
+      if (typeof this.onSetModified === "function") {
+        this.onSetModified();
+      }
+    }
+  }
+
+  resetModified() {
+    if (this._modified) {
+      this._modified = false;
+      if (typeof this.onResetModified === "function") {
+        this.onResetModified();
+      }
+    }
   }
 }
 
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2536,12 +2536,16 @@ class WorkerTransport {
   }
 
   saveDocument(annotationStorage) {
-    return this.messageHandler.sendWithPromise("SaveDocument", {
-      numPages: this._numPages,
-      annotationStorage:
-        (annotationStorage && annotationStorage.getAll()) || null,
-      filename: this._fullReader.filename,
-    });
+    return this.messageHandler
+      .sendWithPromise("SaveDocument", {
+        numPages: this._numPages,
+        annotationStorage:
+          (annotationStorage && annotationStorage.getAll()) || null,
+        filename: this._fullReader.filename,
+      })
+      .finally(() => {
+        annotationStorage.resetModified();
+      });
   }
 
   getDestinations() {
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -127,6 +127,12 @@ function isSameScale(oldScale, newScale) {
   return false;
 }
 
+function beforeUnload(evt) {
+  evt.preventDefault();
+  evt.returnValue = "";
+  return false;
+}
+
 /**
  * Simple viewer control to display PDF content/pages.
  * @implements {IRenderableView}
@@ -436,6 +442,12 @@ class BaseViewer {
     const firstPagePromise = pdfDocument.getPage(1);
 
     const annotationStorage = pdfDocument.annotationStorage;
+    annotationStorage.onSetModified = function () {
+      window.addEventListener("beforeunload", beforeUnload);
+    };
+    annotationStorage.onResetModified = function () {
+      window.removeEventListener("beforeunload", beforeUnload);
+    };
 
     this._pagesCapability.promise.then(() => {
       this.eventBus.dispatch("pagesloaded", {
--- a/test/unit/annotation_storage_spec.js
+++ b/test/unit/annotation_storage_spec.js
@@ -38,5 +38,48 @@ describe("AnnotationStorage", function () {
       expect(value).toEqual("an other string");
       done();
     });
+
+    it("should call onSetModified() if value is changed", function (done) {
+      const annotationStorage = new AnnotationStorage();
+      let called = false;
+      const callback = function () {
+        called = true;
+      };
+      annotationStorage.onSetModified = callback;
+      annotationStorage.getOrCreateValue("asdf", "original");
+      expect(called).toBe(false);
+
+      // not changing value
+      annotationStorage.setValue("asdf", "original");
+      expect(called).toBe(false);
+
+      // changing value
+      annotationStorage.setValue("asdf", "modified");
+      expect(called).toBe(true);
+      done();
+    });
+  });
+
+  describe("ResetModified", function () {
+    it("should call onResetModified() if set", function (done) {
+      const annotationStorage = new AnnotationStorage();
+      let called = false;
+      const callback = function () {
+        called = true;
+      };
+      annotationStorage.onResetModified = callback;
+      annotationStorage.getOrCreateValue("asdf", "original");
+
+      // not changing value
+      annotationStorage.setValue("asdf", "original");
+      annotationStorage.resetModified();
+      expect(called).toBe(false);
+
+      // changing value
+      annotationStorage.setValue("asdf", "modified");
+      annotationStorage.resetModified();
+      expect(called).toBe(true);
+      done();
+    });
   });
 });


timvandermeij:
Relates to #12202.

src/display/annotation_layer.js
```diff
@@ -499,6 +499,7 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
     if (this.data.textAlignment !== null) {
       element.style.textAlign = TEXT_ALIGNMENT[this.data.textAlignment];
     }
+    storage.modified = false;
```
timvandermeij, line 502: Instead of tracking this at the call sites, which is error-prone since it may be forgotten, we can simply track this in the `AnnotationStorage` class itself internally, i.e., mark it as modified when `getOrCreate/setValue` is called with a non-existent value.

src/display/annotation_storage.js
```diff
@@ -19,6 +19,7 @@
 class AnnotationStorage {
   constructor() {
     this._storage = new Map();
+    this.modified = false;
```
timvandermeij, line 22: Make this private, i.e., `this._modified`, and add a getter `get modified` for this property similar to the `size` property for external usage.

web/app.js
```diff
@@ -2618,6 +2620,19 @@ function webViewerKeyUp(evt) {
...
 }
 
+function beforeUnload(evt) {
+  // Unsaved form: are you sure?
+  if (
+    PDFViewerApplication.pdfDocument &&
+    PDFViewerApplication.pdfDocument.annotationStorage.modified
+  ) {
+    evt.preventDefault();
+    evt.returnValue = "";
```
timvandermeij, line 2630: I'm not entirely sure what this does, but what I _think_ should happen is that a dialog is opened, similar to the document properties/password dialogs, that asks the user to confirm/cancel. The strings used should be localized by adding them to the `en-US` locale file (check how the other dialogs use localized strings).

src/display/annotation_layer.js
escapewindow: This means we end up `modified` when we set the default values in the acroform.
@brendandahl suggested adding something like an `isInitialValue` kwarg to the set methods. I'll need to be able to reset the `modified` flag on save, so I'll add a method to allow for that.

web/app.js
escapewindow: This appears to already have strings associated? I get `This page is asking you to confirm that you want to leave - data you have entered may not be saved.` after modifying a form and navigating away.

src/display/annotation_layer.js
escapewindow: Done and done, but resetting the modified flag on save doesn't appear to be working.

escapewindow: Fixed with https://github.com/mozilla/pdf.js/pull/12241/commits/8f17a49ba7848df5abd0ff4d6e8044c86d5b2bdc . I may be clearing `modified` slightly early. But we appear to be passing in the *values* of `annotationStorage` rather than the object; I could potentially change that.

escapewindow: Resolved in https://github.com/mozilla/pdf.js/pull/12241/commits/3173e7235e37795e7e1d0e4386797f65a19dda1f .

Snuffleupagus:
According to https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event, `beforeunload` has undesirable effects in general:
> [...] Attaching an event handler/listener to window or document's beforeunload event prevents browsers from using in-memory page navigation caches, like Firefox's Back-Forward cache or WebKit's Page Cache.

Furthermore, years ago, we used to have a `beforeunload` event listener in another part of the code-base, but removed it because  it caused problems. (Unfortunately I don't remember all the details right now, since it was well over *five* years ago now.)

All-in-all, it seems really unfortunate as far as I'm concerned to use a `beforeunload` event unconditionally like this. Can't we *at least* limit this to documents with forms, to lessen the impact of these changes (unless a solution without `beforeunload` can be had)?

src/display/annotation_layer.js
```diff
@@ -451,7 +451,11 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
       // NOTE: We cannot set the values using `element.value` below, since it
       //       prevents the AnnotationLayer rasterizer in `test/driver.js`
       //       from parsing the elements correctly for the reference tests.
-      const textContent = storage.getOrCreateValue(id, this.data.fieldValue);
+      const textContent = storage.getOrCreateValue(
+        id,
+        this.data.fieldValue,
+        true
```
Snuffleupagus, line 457: Calling a function/method with an explicit boolean parameter like this makes it more difficult to know immediately, when reading the code, what the parameter actually represents without looking up the definition.

Usually, to aid readability in these cases, adding an inline comment will help. Hence changing this line, and similar lines below, to the following would help: 
```suggestion
        /* isInitialValue = */ true
```

src/display/api.js
```diff
@@ -2536,12 +2536,17 @@ class WorkerTransport {
...
+      .sendWithPromise("SaveDocument", {
+        numPages: this._numPages,
+        annotationStorage:
+          (annotationStorage && annotationStorage.getAll()) || null,
+        filename: this._fullReader.filename,
+      })
+      .then(results => {
+        annotationStorage.setUnmodified();
+        return results;
+      });
```
Snuffleupagus, line 2549: If `this.messageHandler.sendWithPromise("SaveDocument", { ... })` is ever rejected, the modified-state of the `AnnotationStorage` thus won't be reset.

Obviously you could argue that it shouldn't be reset in that case, since saving did *not* in fact succeed. However if saving failed once, it seems unlikely a new `saveDocument` call won't also fail. Hence I wonder if simply resetting the modified-state *unconditionally* here won't be generally preferable, in order to avoid repeatedly prompting the user if saving ever fails.

All-in-all, would it thus make sense to change this block to something like the following instead?
```suggestion
      .finally(() => {
        annotationStorage.resetModified();
      });
```

Snuffleupagus (commented):
In addition to https://github.com/mozilla/pdf.js/pull/12241#issuecomment-675913238, I've also got a couple of questions/suggestions based on a cursory look at the code.

src/display/annotation_layer.js
timvandermeij: On top of that, I this such a parameter is not really necessary. I think we can avoid touching _any_ callsite here by tracking this internally in the `AnnotationStorage` class. The first time the default values are stored in the storage, so any second change means it's modified.

escapewindow:
> According to https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event, `beforeunload` has really undesirable effects in general:
> 
> > [...] Attaching an event handler/listener to window or document's beforeunload event prevents browsers from using in-memory page navigation caches, like Firefox's Back-Forward cache or WebKit's Page Cache.
> 
> Furthermore, years ago, we used to have a `beforeunload` event listener in another part of the code-base, but removed it because it caused problems. (Unfortunately I don't remember all the details right now, since it was well over _five_ years ago now.)
> 
> All-in-all, it seems really unfortunate as far as I'm concerned to use a `beforeunload` event unconditionally like this. Can't we _at least_ limit this to documents with forms, to lessen the impact of these changes (unless a solution without `beforeunload` can be had)?

I took a look at this. It looks like `this.bindWindowEvents()` happens before `_initializeMetadata`, and possibly even `load`, so I don't have a `this.documentInfo.IsAcroFormPresent` or a `this.pdfDocument` to refer to. Another solution could be to bind `beforeunload` elsewhere, outside of `bindWindowEvents`. Do you have any recommendations here? (Or am I mistaken about any of my assumptions?)

escapewindow:
> bdahl
> aki: what about binding/unbinding on modified?
> aki
> that makes sense. is there a destroy() or similar?
> i'll try that and test
> bdahl
> may need to add an event that is dispatched from annotation storage on modification and reset?

escapewindow:
> > bdahl
> > aki: what about binding/unbinding on modified?
> > aki
> > that makes sense. is there a destroy() or similar?
> > i'll try that and test
> > bdahl
> > may need to add an event that is dispatched from annotation storage on modification and reset?

[This patch](https://gist.github.com/escapewindow/ca1de62545de83e6f59d0fc4fbc1fadf) doesn't appear to work. I `git stash`ed that, then tried looking for how to send an event from annotation storage. `dispatchEvent` seems to only be in `web/`, and message handlers appear to only go between core and display?

escapewindow:
> bdahl
> aki: could add two properties to class AnnotationStorage e.g. onReset and onModified that the viewer could set as callbacks. then when those are called in the viewer you'd add/remove the unload listener

This worked! https://github.com/mozilla/pdf.js/pull/12241/commits/488f6748cd43369d26ee03726042a7309a081da9

src/display/annotation_storage.js
```diff
@@ -37,6 +40,7 @@ class AnnotationStorage {
     }
 
     this._storage.set(key, defaultValue);
+
```
timvandermeij, line 43: Please revert this change to keep the diff small (it's unrelated to this PR).

src/display/annotation_storage.js
```diff
@@ -62,6 +69,28 @@ class AnnotationStorage {
   get size() {
     return this._storage.size;
   }
+
+  get modified() {
```
timvandermeij, line 73: I don't think this getter is used anymore, so it can be removed.

src/display/annotation_storage.js
```diff
@@ -49,6 +53,9 @@ class AnnotationStorage {
    * @param {Object} value
    */
   setValue(key, value) {
+    if (this._storage.get(key) !== value) {
```
timvandermeij, line 56: The only situation where this won't work is https://github.com/mozilla/pdf.js/blob/b061c300b48531204bc136ce0bc4124a8ba43ef8/src/display/annotation_layer.js#L687 because that class is the only widget that doesn't use `getOrCreateValue` first. You can verify this by opening https://github.com/mozilla/pdf.js/blob/master/test/pdfs/annotation-choice-widget.pdf and without changing anything navigating away from the page, and you'll get the warning.

However, I'll most likely need to change that anyway for #12233, so this should be fine, but merging this patch is therefore blocked on fixing that issue, which I'm working on right now.

src/display/annotation_storage.js
```diff
@@ -19,6 +19,9 @@
 class AnnotationStorage {
   constructor() {
     this._storage = new Map();
+    this._modified = false;
+    this.onModified = null;
```
timvandermeij, line 23: Could you add a comment for why these callbacks are necessary? If I understood the discussion correctly it's only there to ensure that we bind on `beforeunload` when forms are edited and not in other cases for performance/usability reasons, right? In that case, perhaps something like this would work?

```javascript
this._modified = false;

// Callbacks to signal when the modification state is set or reset.
// This is used by the viewer to only bind on `beforeunload` if forms
// are actually edited to prevent doing so unconditionally since that
// can have undesirable efffects.
this.onModified = null;
this.onReset = null;
```

Moreover, I would rename the callbacks to `this.onSetMofied` and `this.onResetModified` to match the method names.

web/base_viewer.js
```diff
@@ -127,6 +127,20 @@ function isSameScale(oldScale, newScale) {
   return false;
 }
 
+function onModifiedAnnotationStorage() {
+  window.addEventListener("beforeunload", beforeUnload);
```
timvandermeij, line 131: This function is not called anywhere else and trivial, so wouldn't it be easier to just inline it below like this (same for the reset function)?

```javascript
annotationStorage.onModified = function() {
  window.addEventListener("beforeunload", beforeUnload);
};
```

That way the relevant code is more in one place which makes it easier to follow in my opinion.

timvandermeij (commented):
Looks good to me with these comments addressed.

src/display/annotation_storage.js
```diff
@@ -49,6 +53,9 @@ class AnnotationStorage {
    * @param {Object} value
    */
   setValue(key, value) {
+    if (this._storage.get(key) !== value) {
+      this.setModified();
```
timvandermeij, line 57: The modification logic should get a unit test. The unit tests for the annotation storage are in https://github.com/mozilla/pdf.js/blob/master/test/unit/annotation_storage_spec.js. I would suggest to add a new one that adds the `onSetModified` and `onResetModified` callbacks and asserts that they are called when `setModified` and `resetModified` are called, and when `setValue` is used, basically covering all lines added here.

src/display/annotation_storage.js
brendandahl: We're getting down to the wire on Firefox's soft freeze coming up. Since the failing behavior isn't too bad, I'm going to merge this in so we have as small as possible updates in mozilla central.

Snuffleupagus:
Sorry, but I'm finding the decision to add a global event listener in `web/base_viewer.js` slightly strange and inconsistent (since there's no others anywhere in the file). Note that when the viewer was originally split into components, we *purposely* tried to move e.g. any direct DOM accesses out of them.

Also, I understand that the following is more difficult to implement, but nonetheless: Note that this will not warn when switching documents, which means that if you call e.g. `PDFViewerApplication.close` (or directly call `BaseViewer.setDocument` when using the viewer components standalone) there won't be any warning about losing form data.

Snuffleupagus:
Sorry, but I'm finding the decision to add a global event listener in `web/base_viewer.js` slightly strange and inconsistent (since there's no others anywhere in the file). Note that when the viewer was originally split into components, we *purposely* tried to move e.g. any direct DOM accesses out of them.

Hence it does seem, at least to me, that putting that code in `PDFViewerApplication.load` (in `web/app.js`) would have seemed slightly more expected overall. 

---

Also, I understand that the following is more difficult to implement, but nonetheless: Note that this will not warn when switching documents, which means that if you call e.g. `PDFViewerApplication.close` (or directly call `BaseViewer.setDocument` when using the viewer components standalone) there won't be any warning about losing form data.

Snuffleupagus:
Sorry, but I'm finding the decision to add a global event listener in `web/base_viewer.js` slightly strange and inconsistent (since there's no others anywhere in the file). Note that when the viewer was originally split into components, we *purposely* tried to move e.g. any direct DOM accesses out of them.

Hence it does seem, at least to me, that placing that code in `PDFViewerApplication.load` (in `web/app.js`) instead would have been slightly more expected overall. 

---

Also, I understand that the following is more difficult to implement, but nonetheless: Note that this will not warn when switching documents, which means that if you call e.g. `PDFViewerApplication.close` (or directly call `BaseViewer.setDocument` when using the viewer components standalone) there won't be any warning about losing form data.

PR #12246, opened 2020-08-19
Return searchString with the match count

DesWurstes:
Firefox [D85140](https://phabricator.services.mozilla.com/D85140#2729587) needs this since on macOS the search string is saved to the find clipboard if there are search results or if this is the first find that fails.

An alternative considered was returning it here:

https://github.com/mozilla/pdf.js/blob/50bc4a18e8c564753365d927d5ec6a6d2cce3072/web/pdf_find_controller.js#L726-L731

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -720,7 +720,11 @@ class PDFFindController {
     if (current < 1 || current > total) {
       current = total = 0;
     }
-    return { current, total };
+    return {
+      current,
+      total,
+      searchString: this._query,
+    };
   }
 
   _updateUIResultsCount() {


Snuffleupagus (changes requested):
Sorry, but is it really intended/necessary to send this with *every* single `updatefindmatchescount` event, instead of just *once* with `updatefindcontrolstate` just below!?

The effect of these patched together will be to call `SetClipboardSearchString` (in mozilla-central) once for *every* match in a document, of which there could very well be *thousands*.

(There's also some other things that I think can be improved here, but let's figure of the above *first* to avoid unnecessary review comments.)

DesWurstes:
@Snuffleupagus Thanks for the pointer! I'll reopen it with the updated patch

PR #12247, opened 2020-08-19
 Improve the field value parsing for choice widgets to handle `null` values

timvandermeij:
The specification states that the field value is `null` if no item is selected and we didn't handle this case properly. Even though this did not break the rendering because we always convert the value to an array and the `includes` check in the display layer would simply not match, the field value would be `[null]` which is not expected and strange from an API perspective.

This commit fixes that by ensuring that we return an empty array in case the field value is `null`. The API therefore still always gives an array for the field value, but now the code is more specific so that the value is either an empty array or an array of strings.

Fixes something I noticed while debugging #12233 (see https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=454&zoom=auto,-215,765 for the specification).

_Slightly easier review with the `?w=1` parameter._

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -1610,11 +1610,14 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
       }
     }
 
-    // Determine the field value. In this case, it may be a string or an
-    // array of strings. For convenience in the display layer, convert the
-    // string to an array of one string as well.
-    if (!Array.isArray(this.data.fieldValue)) {
+    // The field value can be `null` if no item is selected, a string if one
+    // item is selected or an array of strings if multiple items are selected.
+    // For consistency in the API and convenience in the display layer, we
+    // always make the field value an array with zero, one or multiple items.
+    if (isString(this.data.fieldValue)) {
       this.data.fieldValue = [this.data.fieldValue];
+    } else if (!this.data.fieldValue) {
+      this.data.fieldValue = [];
     }
 
     // Process field flags for the display layer.
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -2478,48 +2478,32 @@ describe("annotation", function () {
       }, done.fail);
     });
 
-    it("should handle array field values", function (done) {
-      const fieldValue = ["Foo", "Bar"];
+    it("should convert the field value to an array", function (done) {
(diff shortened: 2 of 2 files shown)


src/core/annotation.js
```diff
@@ -1610,11 +1610,14 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
...
-    // Determine the field value. In this case, it may be a string or an
-    // array of strings. For convenience in the display layer, convert the
-    // string to an array of one string as well.
-    if (!Array.isArray(this.data.fieldValue)) {
+    // The field value can be `null` if no item is selected, a string if one
+    // item is selected or an array of strings if multiple items are selected.
+    // For consistency in the API and convenience in the display layer, we
+    // always make the field value an array with zero, one or multiple items.
+    if (isString(this.data.fieldValue)) {
       this.data.fieldValue = [this.data.fieldValue];
```
Snuffleupagus, line 1618: If this is a String, shouldn't you also wrap the value in `stringToPDFString` to prevent issues elsewhere?
```suggestion
    if (isString(this.data.fieldValue)) {
      this.data.fieldValue = [stringToPDFString(this.data.fieldValue)];
```

src/core/annotation.js
```diff
@@ -1610,11 +1610,14 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
...
+    } else if (!this.data.fieldValue) {
+      this.data.fieldValue = [];
     }
```
Snuffleupagus, line 1621: Furthermore, in the Array-case: Since the comment says that it should be an array of strings, don't you also need to iterate through the array and validate/use `stringToPDFString` on all elements?

timvandermeij: This patch is only about handling the possible `null` value that's in the spec. It doesn't fix #12233, but is only something I noticed while debugging that. Fixing #12233 would most likely require using `stringToPDFString`, but I've got a way to do that in a more centralized manner. That's however part of the patch that will address that issue.

(Same for the comment above.)

Snuffleupagus: Since I couldn't help but noticing the lack of `stringToPDFString` in this this code, I had to ask.
Assuming that this will all be fixed soon anyway, this seems fine to me as-is.

Snuffleupagus (approved):

PR #12248, opened 2020-08-19
Support file save triggered from the Firefox integrated version.

brendandahl:
Related to https://bugzilla.mozilla.org/show_bug.cgi?id=1659753

This allows Firefox trigger a "download" event from ctrl/cmd+s or the "Save
Page As" context menu, which in turn lets pdf.js generate a new PDF if
there is form data to save.

I also now set an `isFallback` flag on downloads triggered from the
fallback bar so Firefox can keep launching the "open with" dialog instead
of the "save as" dialog.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2541,7 +2541,7 @@ class WorkerTransport {
         numPages: this._numPages,
         annotationStorage:
           (annotationStorage && annotationStorage.getAll()) || null,
-        filename: this._fullReader.filename,
+        filename: this._fullReader ? this._fullReader.filename : null,
       })
       .finally(() => {
         annotationStorage.resetModified();
--- a/web/app.js
+++ b/web/app.js
@@ -874,7 +874,7 @@ const PDFViewerApplication = {
     );
   },
 
-  download() {
+  download({ sourceEventType = "download" }) {
     function downloadByUrl() {
       downloadManager.downloadUrl(url, filename);
     }
@@ -902,12 +902,12 @@ const PDFViewerApplication = {
       .getData()
       .then(function (data) {
         const blob = new Blob([data], { type: "application/pdf" });
-        downloadManager.download(blob, url, filename);
+        downloadManager.download(blob, url, filename, sourceEventType);
       })
       .catch(downloadByUrl); // Error occurred, try downloading with the URL.
   },
 
-  save() {
+  save({ sourceEventType = "download" }) {
     if (this._saveInProgress) {
       return;
     }
@@ -927,7 +927,7 @@ const PDFViewerApplication = {
     // When the PDF document isn't ready, or the PDF file is still downloading,
     // simply download using the URL.
     if (!this.pdfDocument || !this.downloadComplete) {
-      this.download();
+      this.download({ sourceEventType });
       return;
     }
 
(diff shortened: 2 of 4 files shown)


web/app.js
```diff
@@ -874,7 +874,7 @@ const PDFViewerApplication = {
     );
   },
 
-  download() {
+  download(isFallback) {
```
Snuffleupagus, line 877: Please change this to the following, to clearly mark the parameter as optional:
```suggestion
  download(isFallback = false) {
```

web/download_manager.js
```diff
@@ -64,7 +64,12 @@ class DownloadManager {
     download(blobUrl, filename);
   }
 
-  download(blob, url, filename) {
+  /**
+   * @param isFallback {boolean} Used to signal that the download was triggered
+   *   from the fallback bar. The version of PDF.js integrated with Firefox uses
+   *   this to to determine whether to show the "save as" or "open with" dialog.
+   */
+  download(blob, url, filename, isFallback) {
```
Snuffleupagus, line 72: Please change this to the following, to clearly mark the parameter as optional:
```suggestion
  download(blob, url, filename, isFallback = false) {
```

web/firefoxcom.js
```diff
@@ -115,7 +115,7 @@ class DownloadManager {
     );
   }
 
-  download(blob, url, filename) {
+  download(blob, url, filename, isFallback) {
```
Snuffleupagus, line 118: Please change this to the following, to clearly mark the parameter as optional:
```suggestion
  download(blob, url, filename, isFallback = false) {
```

web/app.js
```diff
@@ -988,7 +988,7 @@ const PDFViewerApplication = {
         if (!download) {
           return;
         }
-        PDFViewerApplication.download();
+        PDFViewerApplication.download(true /* isFallback */);
```
Snuffleupagus, line 991: Nit: In the code-base, this is *usually* formatted the following way.
```suggestion
        PDFViewerApplication.download(/* isFallback = */ true);
```

src/display/api.js
```diff
@@ -2540,7 +2540,7 @@ class WorkerTransport {
       numPages: this._numPages,
       annotationStorage:
         (annotationStorage && annotationStorage.getAll()) || null,
-      filename: this._fullReader.filename,
+      filename: this._fullReader ? this._fullReader.filename : null,
```
Snuffleupagus, line 2543: Good catch!

Snuffleupagus (approved):
Looks good to me, modulo a couple of *small* suggestions.

brendandahl:
This is still a bit WIP, will update in a bit here once we decide how things should work in the bugzilla bug.

web/app.js
```diff
@@ -1722,6 +1722,7 @@ const PDFViewerApplication = {
     eventBus._on("presentationmode", webViewerPresentationMode);
     eventBus._on("print", webViewerPrint);
     eventBus._on("download", webViewerDownload);
+    eventBus._on("save", webViewerSave);
```
Snuffleupagus, line 1725: This requires a corresponding `eventBus._off(...)` line in `PDFViewerApplication.unbindEvents` below.

PR #12250, opened 2020-08-20
Update the zoom dropdown width calculation to work better in locales with long zoom-strings (PR 11077 follow-up)

Snuffleupagus:
With the changes in PR #11077, the zoom dropdown now looks "squashed" in locales with longer than average zoom-strings[1]. The reason is that the zoom-value and the dropdown-icon are too close together, which doesn't look good in affected locales.

To fix this, the following changes are made:
 - Increase the calculated dropdown width, in `Toolbar._adjustScaleWidth`, to account for the much wider icon (7 px -> 16 px) and the increased padding.
 - Move the dropdown-icon *slightly* outwards, and also *slightly* reduce the left (right in RTL locales) padding of the dropdown-contents.
 - Finally, remove the right (left in RTL locales) padding to reduce the chance of the *default* browser dropdown-icon being visible.

---
[1] This affects e.g. the `de` and `nl` locales, but there's probably other examples as well.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -352,7 +352,7 @@ html[dir='rtl'] #outerContainer.sidebarOpen #viewerContainer:not(.pdfPresentatio
   background-color: var(--sidebar-bg-color);
 }
 html[dir='ltr'] #toolbarSidebar {
-  box-shadow: inset -1px 0 0 rgba(0, 0, 0, 0.25), 
+  box-shadow: inset -1px 0 0 rgba(0, 0, 0, 0.25),
     0 1px 0 rgba(0, 0, 0, 0.15),
     0 0 1px rgba(0, 0, 0, 0.1);
 }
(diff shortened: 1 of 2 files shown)


web/viewer.css
```diff
@@ -352,7 +352,7 @@ html[dir='rtl'] #outerContainer.sidebarOpen #viewerContainer:not(.pdfPresentatio
   background-color: var(--sidebar-bg-color);
 }
 html[dir='ltr'] #toolbarSidebar {
-  box-shadow: inset -1px 0 0 rgba(0, 0, 0, 0.25),
```
Snuffleupagus, line 355: I didn't manually change this line, but upon saving my editor removed this trailing space; given that we shouldn't have trailing spaces anywhere keeping this change thus seem OK here.

timvandermeij (approved):

timvandermeij:
Thank you for the follow-up!

PR #12251, opened 2020-08-20
Return the query with the findcontrolstate

DesWurstes:
of [D85140](https://phabricator.services.mozilla.com/D85140).

@Snuffleupagus any suggestions?

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2438,12 +2438,18 @@ function webViewerUpdateFindMatchesCount({ matchesCount }) {
   }
 }
 
-function webViewerUpdateFindControlState({ state, previous, matchesCount }) {
+function webViewerUpdateFindControlState({
+  state,
+  previous,
+  matchesCount,
+  rawQuery,
+}) {
   if (PDFViewerApplication.supportsIntegratedFind) {
     PDFViewerApplication.externalServices.updateFindControlState({
       result: state,
       findPrevious: previous,
       matchesCount,
+      rawQuery,
     });
   } else {
     PDFViewerApplication.findBar.updateUIState(state, previous, matchesCount);
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -736,6 +736,7 @@ class PDFFindController {
       state,
       previous,
       matchesCount: this._requestMatchesCount(),
+      rawQuery: this._state ? this._state.query : null,
     });
   }
 }


web/pdf_find_controller.js
```diff
@@ -736,6 +736,7 @@ class PDFFindController {
       state,
       previous,
       matchesCount: this._requestMatchesCount(),
+      query: this._query,
```
Snuffleupagus, line 739: Note that this won't actually return *exactly* what the findbar input contains, which may be surprising to the user.
What you actually want here is probably `rawQuery: this._rawQuery,` instead.

web/app.js
```diff
@@ -2438,12 +2438,18 @@ function webViewerUpdateFindMatchesCount({ matchesCount }) {
   }
 }
 
-function webViewerUpdateFindControlState({ state, previous, matchesCount }) {
+function webViewerUpdateFindControlState({
+  state,
+  previous,
+  matchesCount,
+  query,
```
Snuffleupagus, line 2445:
```suggestion
  rawQuery,
```

Snuffleupagus (changes requested):
Also, please note that in the upstream patch you'll need to extend the existing validation to ensure that the new parameter is a string, before including it in the "PDFJS:Parent:updateControlState" message, see https://searchfox.org/mozilla-central/rev/6cc48251bb97600fdf11a5b4c5f621bfc8606d55/toolkit/components/pdfjs/content/PdfStreamConverter.jsm#536-565

DesWurstes:
Thanks!

web/pdf_find_controller.js
```diff
@@ -736,6 +736,7 @@ class PDFFindController {
       state,
       previous,
       matchesCount: this._requestMatchesCount(),
+      rawQuery: this._rawQuery,
```
Snuffleupagus, line 739: Really sorry, I just looked at this more in-depth and realized that this property is not guaranteed to always be set when `_updateUIState` is called.

The following should be much safer in general:
```suggestion
      rawQuery: this._state ? this._state.query : null,
```
assuming that you also add validation similar to https://searchfox.org/mozilla-central/rev/6cc48251bb97600fdf11a5b4c5f621bfc8606d55/toolkit/components/pdfjs/content/PdfStreamConverter.jsm#554-557, i.e. adding the following in `PdfStreamConverter.jsm` (such that the parameter is optional):
```javascript
const rawQuery = typeof data.rawQuery === "string" ?  data.rawQuery : null;
```

Snuffleupagus (approved):
r=me, with the final comment addressed.

Sorry about the back-and-forth here, it's slightly more difficult to review these types of patches that need to be split between Bugzilla/Github.

web/pdf_find_controller.js
DesWurstes: I have added:

```js
    // Same for the `rawQuery` property.
    let rawQuery = null;
    if (typeof data.rawQuery === "string") {
      rawQuery = data.rawQuery;
    }
```

timvandermeij (approved):

timvandermeij:
Thanks!

Issue #1773, opened 2012-05-29
PDF doesn't display anything, complains about my pdf viewer


samth:
This PDF http://www.northeastern.edu/hrm/pdfs/resources/benefits/MBTA-pretax-form-July2012.pdf produces the following error message as the contents of the pdf:

```
Please wait... 

If this message is not eventually replaced by the proper contents of the document, your PDF 
viewer may not be able to display this type of document. 

You can upgrade to the latest version of Adobe Reader for Windows®, Mac, or Linux® by 
visiting  http://www.adobe.com/products/acrobat/readstep2.html. 

For more assistance with Adobe Reader visit  http://www.adobe.com/support/products/
acrreader.html.
```

Neither Evince nor OS X Preview works either.

DuncanGEMcMillan:
Solved!! I updated my Adobe Reader to the latest and greatest version using Chrome and getting the same message. 
I used Internet Explorer instead of Chrome and it opened up right away.

Andsbf:
Apparently Adobe found a way to screw the nice standard "PDF". :(

dbrisinda:
After upgrading to latest, which presumably would install the necessary plugins, these horrid XFA pdfs still don't load in Safari. My heartfelt wishes for a fast and painful death to Adobe.

SegersIan:
The reason is simple. You need to launch this in the original Acrobat Adobe Reader, not 'in-browser'. 'In-browser' pdf reading is often a plugin or is 3th party rendering and does not support interactive elements. So the simple fix is, download locally the pdf and open it with the actual Acrobat Adobe Reader and not in browser or with another reader than the Adobe's one.

emilianbold:
Sadly these XFA form PDFs can only be opened and printed from Adobe Reader.

My workaround is to use this printer driver on macOS https://github.com/emilianbold/PDFwriter and get Adobe to print to it. This produces a normal PDF that pdf.js / Preview can open.

marco-c:
The forms appear to be working well now. @calixteman do you want to add them to reftests?

PR #12254, opened 2020-08-20
only warn on xfa if an acroform is not also present

escapewindow:


timvandermeij:
Unfortunately I don't think this works. If XFA is present, by definition AcroForm is too since XFA is inside the AcroForm dictionary, so the condition in this statement can never happen. Refer to https://github.com/mozilla/pdf.js/blob/7edc5cb79fe829d45fed85e2b4b6d8594522cc10/src/core/document.js#L582-L598 for the XFA/AcroForm detection.

I'm not really sure what the exact relation is between XFA and AcroForms. I have always seen forms with either only AcroForm or with AcroForm and XFA, but in the combined cases I have yet to find a PDF that wasn't fillable when we ignore XFA (which we do now). That's why I _think_ XFA is only an extension that we can ignore, but I don't know if there are real-life examples of files where an AcroForm form cannot be filled in properly because XFA features are missing (for example to expand/show hidden form fields based on a selection of another widget).

@brendandahl Do you perhaps know more about how XFA relates to AcroForm? My feeling tells me that we can safely ignore XFA completely and only log a warning for it (so no fallback bar) because I haven't seen any cases where XFA was required for filling in the form, but I'm not entirely sure. On the other hand, if such a file were to surface, we can always reconsider.

TLDR; I would say we only log a warning if XFA is present and nothing else (no fallback bar).

escapewindow:
Hm, maybe I could check for `this.xfa && !this.acroForm.get("fields")` (or the check `(!Array.isArray(fields) || fields.length === 0)`, rather)? Removing the fallback bar is easy too.

timvandermeij:
If it can happen that XFA is present _and_ the AcroForm dictionary contains no fields, then that might indeed be better since the user then correctly gets notified that filling the work definitely won't work. The question is: does the specification even allow that?

escapewindow:
I'm not sure -- I think @brendandahl said most xfa forms have acroforms, but some don't. The spec seems to say xfa forms should have acroforms. I pushed the fallback bar removal only change; I can push the "fallback bar if there's an xfa without acroform fields" change as well if wanted.

timvandermeij:
I can't find it directly in the specification, but Apache PDFBox does handle the case where XFA is present and the AcroForm field are missing; see https://github.com/apache/pdfbox/blob/e3150a327c4291654451fce149f3665fa294e498/pdfbox/src/main/java/org/apache/pdfbox/pdmodel/interactive/form/PDAcroForm.java#L645-L653. They call it "dynamic XFA" and I found that Wikipedia also mentions this at https://en.wikipedia.org/wiki/XFA.

I think the conclusion is:
- Dynamic XFA: `AcroForm -> XFA` entry is present and `AcroForm -> Field` is an empty array. This is unsupported and should show the fallback bar. I haven't found any viewer other that Adobe's viewers that can handle this, and it doesn't seem common at all.
- Static XFA: `AcroForm -> XFA` entry is present and `AcroForm -> Fields` is a non-empty array. This is supported since we can use the AcroForm data and ignore the XFA data since it should not be important (if we ever find a file where it is, we can reconsider, but for now I don't have any reason to believe otherwise). From Wikipedia: "In a static form the form’s appearance and layout is fixed, regardless of the field content. Any unfilled fields are present in the form."

escapewindow:
I tried https://gist.github.com/escapewindow/2e13c19e1ba47e25c09570e234c7f509 , and got

```
Uncaught (in promise) TypeError: can't access property "get", this.pdfDocument.acroForm is undefined
    _initializeMetadata http://localhost:8888/web/app.js:1422
    load http://localhost:8888/web/app.js:1316
    open http://localhost:8888/web/app.js:832
57d4621c-0c4c-bf46-ac7a-86620ceec23c:1422:22
    _initializeMetadata http://localhost:8888/web/app.js:1455
    InterpretGeneratorResume self-hosted:1478
    AsyncFunctionNext self-hosted:684
    (Async: async)
    load http://localhost:8888/web/app.js:1316
    open http://localhost:8888/web/app.js:832
```

This is odd to me -- I'd guess if `info.IsAcroFormPresent` is set, that `pdfDocument.acroForm` would be set... either that's not true, or I have a bug I'm overlooking.

I may poke at changing `docInfo` to add an `isDynamicXFAPresent` ?

timvandermeij:
The AcroForm fields information is not exposed because it's internal functionality (only the `getAnnotations` API call gives out annotation information in a structured way). I'll try to give this some more thought over the weekend since the AcroForm/XFA logic isn't completely trivial now and it would be great if we can simplify it to also make the checks here much easier. I'll get back to you on this soon.

escapewindow:
Attempted `IsDynamicXFAPresent` in https://github.com/mozilla/pdf.js/pull/12254/commits/7cd98488155a5262e518f9292c49e74c32ff073f . Previous attempts in https://github.com/mozilla/pdf.js/compare/master...escapewindow:broken-xfa-acroform?expand=1 .

> The AcroForm fields information is not exposed because it's internal functionality (only the `getAnnotations` API call gives out annotation information in a structured way). I'll try to give this some more thought over the weekend since the AcroForm/XFA logic isn't completely trivial now and it would be great if we can simplify it to also make the checks here much easier. I'll get back to you on this soon.

Thanks!

escapewindow:
> bdahl
> aki: https://github.com/mozilla/pdf.js/issues/1773#issuecomment-46214942 is an example where the viewer shows something totally different if XFA isn't supported
> aki
> boo
> so we should have a warning but no fallback bar? or keep the current behavior?
> bdahl
> i guess for that type, show the fallback bar right away. i'm not actually sure how to detect the above type of PDF though
> aki
> yeah

timvandermeij:
I have spent the majority of today figuring out how this works exactly and it's a bit more difficult than I initially thought to cover all cases, but I did find a solution that covers all known cases including the one mentioned just above. I have made #12271 with the full solution.

escapewindow:
Closing in favor of #12271.

PR #12259, opened 2020-08-21
Fix handling of symbolic fonts and unicode cmaps.

### LINKED ISSUE #12120: Missing prominent character in pdf.
I found other tickets about this kind of issue but I have no idea if it the same problem.

Attach (recommended) or Link to PDF file here: [Planning_1988.pdf](https://github.com/mozilla/pdf.js/files/4973247/Planning_1988.pdf)

Configuration:

Firefox 78.0.2 (64 bits)

Windows version 2004 19041.388

- PDF.js version:

2.5.179 and on live : 2.6.138

- Is a browser extension: ??? It's the pdf.js from firefox.

From adobe reader :

adobe

From PDF.js

pdfjs

Output of the console : null

brendandahl:
In issue 12120, the font has a 1,0 cmap and is marked symbolic which according
to the spec means we should directly use the cmap instead of the extra steps
that are defined in 9.6.6.4. However, just fixing that caused bug 1057544 to
break. The font in bug 1057544 has a 0,1 cmap (Unicode 1.1) which we were not
using, but is easy to support. We're also easily able to support some of the
other unicode cmaps, so I added those as well.

Fixes #12120.


timvandermeij (commented):
Looks good to me, but I'd like @Snuffleupagus to also look at this since I'm not too familiar with the font logic.

Snuffleupagus (approved):
Looks good to me, thank you!

timvandermeij (approved):

timvandermeij:
Nice work!

PR #12268, opened 2020-08-23
Reduce the leading padding for the `numPages` span (PR 11077 follow-up)

Snuffleupagus:
Currently there's enough leading padding that the `numPages` span feels somewhat "disconnected" from the `pageNumber` input, which seems unfortunate when they contain related state.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #12269, opened 2020-08-23
Add support for missing appearances for hightlights, strikeout, squiggly and underline annotations.

calixteman:
This patch aims to fix:
https://bugzilla.mozilla.org/show_bug.cgi?id=1538111
and issue #6810.


src/core/annotation.js
```diff
@@ -791,6 +794,58 @@ class MarkupAnnotation extends Annotation {
   setCreationDate(creationDate) {
     this.creationDate = isString(creationDate) ? creationDate : null;
   }
+
+  _setDefaultAppearance(xref, drawColor, fillColor, blendMode, pointsCallback) {
```
Snuffleupagus, line 798: `drawColor` -> `strokeColor`, since the latter is consistent with both the specification and the PDF.js implementation.

Also, given the number of parameters the readability of the call-sites are suffering quite a bit in my opinion, hence changing this method to take a parameter object instead would really help:
```suggestion
  _setDefaultAppearance({
    xref,
    strokeColor,
    fillColor,
    blendMode,
    pointsCallback,
  }) {
```

calixteman:
This patch is ready for review, thanks.

timvandermeij (commented):
I have tested the viewer with the four files from my comment and they render exactly as expected. Nice work so far!

src/core/annotation.js
```diff
@@ -1854,9 +1919,28 @@ class HighlightAnnotation extends MarkupAnnotation {
 
     this.data.annotationType = AnnotationType.HIGHLIGHT;
 
-    const quadPoints = getQuadPoints(parameters.dict, this.rectangle);
+    const quadPoints = getQuadPoints(parameters.dict, null);
```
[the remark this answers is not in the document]
calixteman: The pdf in https://bugzilla.mozilla.org/show_bug.cgi?id=1538111 has a "wrong" rect for the second highlight.
In poppler they removed too the check with rectangle too:
https://github.com/freedesktop/poppler/commit/3d4985f14e54ddcc64ea654b23e931b7e6acfbdc
And I tested the pdf in the above bug in Acrobat and the second highlight is rendered correctly.
So I guess it's the correct way to do...

src/core/annotation.js
```diff
@@ -1854,9 +1919,28 @@ class HighlightAnnotation extends MarkupAnnotation {
...
     if (quadPoints) {
       this.data.quadPoints = quadPoints;
+      if (!this.appearance) {
+        // Default color is yellow
+        const fillColor = parameters.dict.has("C")
+          ? Array.from(this.color).map(c => c / 255)
```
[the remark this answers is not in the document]
calixteman: You're right but we need to get the 3 components as floats.

src/core/annotation.js
```diff
@@ -1854,9 +1919,28 @@ class HighlightAnnotation extends MarkupAnnotation {
 
     this.data.annotationType = AnnotationType.HIGHLIGHT;
 
-    const quadPoints = getQuadPoints(parameters.dict, this.rectangle);
+    const quadPoints = getQuadPoints(parameters.dict, null);
     if (quadPoints) {
       this.data.quadPoints = quadPoints;
+      if (!this.appearance) {
+        // Default color is yellow
```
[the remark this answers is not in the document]
calixteman: I found these values in checking in pdfium.

src/core/annotation.js
```diff
@@ -1893,9 +2020,34 @@ class StrikeOutAnnotation extends MarkupAnnotation {
...
+        const strokeColor = parameters.dict.has("C")
+          ? Array.from(this.color).map(c => c / 255)
+          : [0, 0, 0];
+        this._setDefaultAppearance({
+          xref: parameters.xref,
+          extra: "[] 0 d 1 w",
+          strokeColor,
+          pointsCallback: (buffer, points) => {
+            buffer.push(
+              `${(points[0].x + points[2].x) / 2} ${
```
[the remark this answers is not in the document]
calixteman: I guess it's a `gulp lint --fix` consequence

test/test_manifest.json
```diff
@@ -990,6 +990,12 @@
        "lastPage": 1,
        "type": "eq"
     },
+    {  "id": "bug1538111",
```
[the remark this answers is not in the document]
calixteman: Thx a lot for these files.

src/core/annotation.js
```diff
@@ -200,7 +200,10 @@ function getQuadPoints(dict, rect) {
 
       // The quadpoints should be ignored if any coordinate in the array
       // lies outside the region specified by the rectangle.
```
timvandermeij, line 202: Let's extend this comment to mention why `rect` can be `null`:

```
      // The quadpoints should be ignored if any coordinate in the array
      // lies outside the region specified by the rectangle. The rectangle
      // can be `null` for markup annotations since their rectangle may be
      // incorrect (fixes bug 1538111).
````

File: test/pdfs/.gitignore [hunk 8]
```diff
@@ -355,6 +359,7 @@
 !openoffice.pdf
 !issue7014.pdf
 !issue8187.pdf
+!StrikeOut.without.appearance.stream.pdf
```
timvandermeij, line 362: Let's call this file `annotation-strikeout-without-appearance.pdf` and put it below the existing `annotation-strikeout.pdf` to group them a bit more and to have a similar naming scheme. Similar for the other three files.

test/pdfs/.gitignore
```diff
@@ -114,6 +114,7 @@
 !bug1252420.pdf
 !bug1513120_reduced.pdf
 !bug1552113.pdf
+!annotation_markup_multiline_no_ap.pdf
```
timvandermeij, line 117: Let's call this file `bug1538111.pdf` to match the other bug files.

test/test_manifest.json
```diff
@@ -1191,6 +1191,12 @@
        "link": true,
        "type": "eq"
     },
+    {  "id": "Highlight.without.appearance.stream",
```
timvandermeij, line 1194: Same here about renaming to the same as the file name and putting it below the existing highlight test. Similar for the other three files.

test/test_manifest.json
```diff
@@ -1191,6 +1191,12 @@
...
+       "file": "pdfs/Highlight.without.appearance.stream.pdf",
+       "md5": "a1f2811324fa1ff0c9f1778697413dad",
+       "rounds": 1,
+       "type": "eq"
```
timvandermeij, line 1198: The new tests in this file _all_ need to have `"annotations": true` in order to render the annotations.

timvandermeij (approved):

timvandermeij:
Nice work; thanks!

/botio makeref

PR #12271, opened 2020-08-23
Improve AcroForm/XFA form type detection

### LINKED ISSUE #12217: Notification for PDF containing forms with pref enabled still displayed on some files
**Affected versions**
* 81.0a1 (2020-08-16)

**Affected platforms**
* Windows 10, macOS 11.0, Ubuntu 16;

**Steps to reproduce**
pdfjs.renderInteractiveForms - enabled

1. Launch Firefox;
2. Access a PDF form with fillable content such as [link](http://www.aloaha.com/wp-content/uploads/2016/07/SampleForm-1.pdf);
3. Click inside any field, start filling in data;

**Expected result**
* data is filled, no unexpected warning notifications appear;

**Actual result**
* warning banner for [Document containing forms & Filling of forms unsupported] is displayed;
 
Additional notes:
* S3 as suggested severity;
* attached screenshot with the issue;
Screenshot 2020-08-17 at 10 45 29

timvandermeij:

Fixes #12217.
Replaces #12254.

For completeness, here is the list of test PDF files with the form type detection results from before/after this patch (note that we only fallback if _only_ XFA is available):

| File  | Before | After | Notes |
| ------ | -------- | ------- | ------- |
| http://www.aloaha.com/wp-content/uploads/2016/07/SampleForm-1.pdf  | AcroForm, XFA, fallback | AcroForm, XFA, no fallback|
| http://www.cic.gc.ca/english/pdf/kits/forms/IMM5257E.PDF  | AcroForm, XFA, fallback | XFA, fallback | `SigFlags` bit set and nested signature in `Fields`.
| https://web.archive.org/web/20121105185256if_/http://www.northeastern.edu/hrm/pdfs/resources/benefits/MBTA-pretax-form-July2012.pdf | AcroForm, XFA, fallback | XFA, fallback | Empty `Fields`.
| https://web.archive.org/web/20110918100215/http://www.irs.gov/pub/irs-pdf/f1040.pdf | AcroForm, no fallback | AcroForm, no fallback | `SigFlags` available, but first bit not set.
| https://github.com/mozilla/pdf.js/files/762326/212241.6.pdf | AcroForm, no fallback | AcroForm, no fallback | `SigFlags` bit set and multiple signatures in `Fields`.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -600,19 +600,6 @@ class PDFDocument {
       info("Cannot fetch AcroForm entry; assuming no AcroForms are present");
       this.acroForm = null;
     }
-
-    // Check if a Collection dictionary is present in the document.
-    try {
-      const collection = this.catalog.catDict.get("Collection");
-      if (isDict(collection) && collection.size > 0) {
-        this.collection = collection;
-      }
-    } catch (ex) {
-      if (ex instanceof MissingDataException) {
-        throw ex;
-      }
-      info("Cannot fetch Collection dictionary.");
-    }
   }
 
   get linearization() {
@@ -745,7 +732,7 @@ class PDFDocument {
       IsLinearized: !!this.linearization,
       IsAcroFormPresent: !!this.acroForm,
       IsXFAPresent: !!this.xfa,
-      IsCollectionPresent: !!this.collection,
+      IsCollectionPresent: !!this.catalog.collection,
     };
 
     let infoDict;
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -573,7 +573,8 @@ class PDFDocument {
   }
 
   parse(recoveryMode) {
-    this.setup(recoveryMode);
+    this.xref.parse(recoveryMode);
+    this.catalog = new Catalog(this.pdfManager, this.xref);
 
     // The `checkHeader` method is called before this method and parses the
     // version from the header. The specification states in section 7.5.2
@@ -674,11 +675,6 @@ class PDFDocument {
 
   parseStartXRef() {
     this.xref.setStartXRef(this.startXRef);
-  }
-
-  setup(recoveryMode) {
-    this.xref.parse(recoveryMode);
-    this.catalog = new Catalog(this.pdfManager, this.xref);
   }
 
   get numPages() {
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -584,21 +584,13 @@ class PDFDocument {
     }
 
     // Check if AcroForms are present in the document.
-    try {
-      this.acroForm = this.catalog.catDict.get("AcroForm");
-      if (this.acroForm) {
-        this.xfa = this.acroForm.get("XFA");
-        const fields = this.acroForm.get("Fields");
-        if ((!Array.isArray(fields) || fields.length === 0) && !this.xfa) {
-          this.acroForm = null; // No fields and no XFA, so it's not a form.
-        }
+    this._hasAcroForm = !!this.catalog.acroForm;
+    if (this._hasAcroForm) {
+      this.xfa = this.catalog.acroForm.get("XFA");
+      const fields = this.catalog.acroForm.get("Fields");
+      if ((!Array.isArray(fields) || fields.length === 0) && !this.xfa) {
+        this._hasAcroForm = false; // No fields and no XFA, so it's not a form.
       }
-    } catch (ex) {
-      if (ex instanceof MissingDataException) {
-        throw ex;
-      }
-      info("Cannot fetch AcroForm entry; assuming no AcroForms are present");
-      this.acroForm = null;
     }
   }
 
@@ -730,7 +722,7 @@ class PDFDocument {
     const docInfo = {
       PDFFormatVersion: version,
       IsLinearized: !!this.linearization,
-      IsAcroFormPresent: !!this.acroForm,
+      IsAcroFormPresent: this._hasAcroForm,
       IsXFAPresent: !!this.xfa,
       IsCollectionPresent: !!this.catalog.collection,
     };
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -552,6 +552,7 @@ class PDFDocument {
     this.stream = stream;
     this.xref = new XRef(stream, pdfManager);
     this._pagePromises = [];
+    this._version = null;
 
     const idCounters = {
       font: 0,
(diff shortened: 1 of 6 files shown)


src/core/document.js
```diff
@@ -715,6 +697,61 @@ class PDFDocument {
...
   }
 
+  /**
+   * @private
+   */
+  _hasOnlyDocumentSignatures(fields) {
+    return fields.every(field => {
+      field = this.xref.fetchIfRef(field);
+      if (field.has("Kids")) {
+        return this._hasOnlyDocumentSignatures(field.get("Kids"));
```
Snuffleupagus, line 707: Given that this is a *recursive* function call, you really must protect against infinite loops in this method!

One option could be to track all seen `Kids` nodes and break if anyone is ever repeated, however that relies on the assumption that they're always `Ref`s which may not be guaranteed in the general case.
The easiest solution is probably to just keep track of the recursion-depth, and break when it reaches some reasonable limit (e.g. `10`?).

src/core/document.js
```diff
@@ -715,6 +697,61 @@ class PDFDocument {
...
+      }
+      const fieldType = field.get("FT");
+      const isSignature = isName(fieldType) && fieldType.name === "Sig";
```
Snuffleupagus, line 710: This can be actually be simplified to the e.g. following:
```suggestion
      const isSignature = isName(/* field type = */ field.get("FT"), "Sig");
```

src/core/document.js
```diff
@@ -744,7 +742,7 @@ class PDFDocument {
       IsLinearized: !!this.linearization,
       IsAcroFormPresent: !!this.acroForm,
       IsXFAPresent: !!this.xfa,
-      IsCollectionPresent: !!this.collection,
+      IsCollectionPresent: this._hasCollection,
```
Snuffleupagus, line 782: With the changes to the `Catalog.collection` getter, this can be simplified to *only* the following and `this._hasCollection` removed since it's unused (in later commits, you're also removing the temporarily added `this._hasAcroForm` boolean):
```suggestion
      IsCollectionPresent: !!this.catalog.collection,
```

src/core/document.js
```diff
@@ -715,6 +697,61 @@ class PDFDocument {
...
+        // the first bit of the `SigFlags` integer (see Table 219 in the
+        // specification).
+        const fields = acroForm.get("Fields");
+        const hasFields = Array.isArray(fields) && fields.length > 0;
+        const sigFlags = acroForm.get("SigFlags");
+        const hasDocumentSignatures = !!(sigFlags & 0x1);
+        let hasOnlyDocumentSignatures = false;
+        if (hasDocumentSignatures) {
+          hasOnlyDocumentSignatures = this._hasOnlyDocumentSignatures(fields);
+        }
```
Snuffleupagus, line 743: Nit: You can probably collapse this a little bit, since there's intermediate state that's not really being used here.
```suggestion
        const hasOnlyDocumentSignatures =
          !!(sigFlags & 0x1) && this._hasOnlyDocumentSignatures(fields);
          
```

Snuffleupagus (changes requested):
This definitely looks like a big step in the right direction, given how complicated the AcroForm/XFA situation apparently is; however there's a couple of things I'd suggest changing.

I've not completely reviewed all of this code yet, notably the new unit-tests, but figured I'd wait with that until the patches are updated.

src/core/document.js
```diff
@@ -714,6 +697,61 @@ class PDFDocument {
...
+        Array.isArray(rectangle) && rectangle.every(value => value === 0);
+      return isSignature && isInvisible;
+    });
+  }
+
+  get formInfo() {
+    const formInfo = { hasAcroForm: false, hasXfa: false };
+    try {
+      const acroForm = this.catalog.acroForm;
+      if (acroForm) {
```
Snuffleupagus, line 722: Also, with the error handling added to `Catalog.acroForm` you can swap these lines around a little bit:
```suggestion
    const acroForm = this.catalog.acroForm;
    if (acroForm) {
      try {
```

timvandermeij:
Thank you for the review! I have addressed all comments in the new commit series.

src/core/document.js
```diff
@@ -574,40 +575,12 @@ class PDFDocument {
   parse(recoveryMode) {
     this.setup(recoveryMode);
```
Snuffleupagus, line 576: Now that you have *significantly* simplified the `parse` method, and thank you for doing that since it always felt a bit unwieldy, I believe that you can simply inline `PDFDocument.setup` here given that it only contains two lines of code:
https://github.com/mozilla/pdf.js/blob/34cec29288212a6ca5bb9d703d6c5102c47e635d/src/core/document.js#L706-L709

Snuffleupagus (approved):
I've added three more comments, after which I believe that this is good to go :-) 

This code now looks much nicer, thank you!

src/core/document.js
```diff
@@ -714,6 +687,67 @@ class PDFDocument {
...
+    const acroForm = this.catalog.acroForm;
+    if (!acroForm) {
+      return shadow(this, "formInfo", formInfo);
+    }
+
+    try {
+      // The document contains XFA data if the `XFA` entry is a non-empty
+      // array or stream.
+      const xfa = acroForm.get("XFA");
+      const hasXfa = (Array.isArray(xfa) || isStream(xfa)) && xfa.length > 0;
```
Snuffleupagus, line 726: One final thing: When the `XFA`-entry is a Stream, can it ever be a something that's (indirectly) a `DecodeStream`?
In that case, there's no concept of exact Stream-length and consequently no `length` getter. Hence I believe that to prevent *any* future issues here, you want the following instead which is guaranteed to work for *all* Stream-implementations:
```suggestion
      const hasXfa =
        (Array.isArray(xfa) && xfa.length > 0) ||
        (isStream(xfa) && !xfa.isEmpty);
```

src/core/document.js
```diff
@@ -697,6 +687,67 @@ class PDFDocument {
...
+        }
+        return this._hasOnlyDocumentSignatures(
+          field.get("Kids"),
+          recursionDepth
+        );
+      }
+      const isSignature = isName(field.get("FT"), "Sig");
+      const rectangle = field.get("Rect");
+      const isInvisible =
+        Array.isArray(rectangle) && rectangle.every(value => value === 0);
```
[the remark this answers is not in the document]
timvandermeij: I found this while looking at the various files, but I looked up some more documentation on this and found https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/acrobat_digital_signature_appearances_v9.pdf#page=7&zoom=auto,-216,358 that mentions that "for invisible signatures, the annotation has a rectangle array with values [ 0 0 0 0 ]", so the current code should be correct as it matches the specification and what I found in the files.

src/core/document.js
timvandermeij: The specification is not really clear on that, but since it wouldn't be the first time we see unexpected things in PDF files, I'll indeed handle this case. Moreover, the `isEmpty` getter is nicer since it works in all cases; I didn't know we had that.

PR #12285, opened 2020-08-26
Ensure that the `sourceEventType` parameter is actually optional in `PDFViewerApplication.{download, save}` (PR 12248 follow-up)

Snuffleupagus:
Without these changes, clicking on the "Open With Different Viewer"-button on the Firefox fallback bar won't actually do anything and the following is printed in the web-console:
```
Uncaught TypeError: (destructured parameter) is undefined
    download resource://pdf.js/web/viewer.js:956
    response resource://pdf.js/web/viewer.js:1054
    listener resource://pdf.js/web/viewer.js:11891
viewer.js:956:1
```

Furthermore, this patch also fixes `PDFViewerApplication.fallback` to pass in an explicit `sourceEventType` when triggering downloading. While this, on its own, would obviously have been sufficient to fix the bug described above, it seems wrong to outright break backwards compatibility of any older `PDFViewerApplication.download` calls.

---

/cc @brendandahl 
/cc @timvandermeij

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -874,7 +874,7 @@ const PDFViewerApplication = {
     );
   },
 
-  download({ sourceEventType = "download" }) {
+  download({ sourceEventType = "download" } = {}) {
     function downloadByUrl() {
       downloadManager.downloadUrl(url, filename);
     }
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Given that PR #12248 was included in https://bugzilla.mozilla.org/show_bug.cgi?id=1660483, which landed in Firefox 81, we unfortunately *must uplift* this patch since the fallback bar will otherwise be broken once Firefox 81 is officially released.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

@brendandahl Can you take care of getting this uplifted?

PR #12286, opened 2020-08-26
Use a link, rather than `window.open`, when opening PDF attachments in Firefox (bug 1661259)

Snuffleupagus:
Unfortunately e.g. ad blockers can interfere with `window.open` calls, thus preventing PDF attachments from being opened/viewed. For the MOZCENTRAL-build, we can work-around this problem by using a (hidden) link instead.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1661259


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #12287, opened 2020-08-26
Bug 1661226 - Push button are not rendered with renderInteractiveForms enabled

calixteman:
Bugzilla bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1661226

### DIFF
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -4572,6 +4572,16 @@
        "type": "eq",
        "annotations": true
     },
+    {  "id": "issue4872-forms",
+       "file": "pdfs/issue4872.pdf",
+       "md5": "21c6cbc682140d6f6017bbeb45892053",
+       "rounds": 1,
+       "link": true,
+       "firstPage": 1,
+       "lastPage": 1,
+       "type": "eq",
+       "forms": true
+    },
     {  "id": "issue6108",
        "file": "pdfs/issue6108.pdf",
        "md5": "8961cb55149495989a80bf0487e0f076",
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -1387,6 +1387,15 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
   }
 
   getOperatorList(evaluator, task, renderForms, annotationStorage) {
+    if (this.data.pushButton) {
+      return super.getOperatorList(
+        evaluator,
+        task,
+        false, // we use normalAppearance to render the button
+        annotationStorage
+      );
+    }
+
     if (annotationStorage) {
       const value = annotationStorage[this.data.id] || false;
       let appearance;


test/test_manifest.json
```diff
@@ -1599,6 +1599,12 @@
        "lastPage": 6,
        "type": "eq"
     },
+    {  "id": "bug1661226",
+       "file": "pdfs/bug1661226.pdf",
+       "md5": "ff8328a52b48026342aedb5620f27e6d",
+       "rounds": 1,
+       "type": "eq"
```
Snuffleupagus, line 1606: Doesn't this need to have `"forms": true` added as well here, to actually trigger the code-path that causes the problems?
As-is I'm not sure that you're actually testing what you want here, since the bug only manifests with `pdfjs.renderInteractiveForms = true` set.

Snuffleupagus (changes requested):

test/test_manifest.json
brendandahl: I've filed #12288 in regards to this.

test/pdfs/.gitignore
```diff
@@ -139,6 +139,7 @@
 !issue3205r.pdf
 !issue3207r.pdf
 !issue3263r.pdf
+!bug1661226.pdf
```
timvandermeij, line 142: I already included a file with a pushbutton in https://github.com/mozilla/pdf.js/commit/0fe80df2a75a70162ab3bae56eb9fd943f4726ab#diff-b19c886c212474d3b37d6026a795a246. The reason it didn't cause rendering problems is because it's not also run with `forms: true`.

Instead of adding this new file, please use the existing one by reverting the changes in the `test` folder in this PR and including a new entry in the `test_manifest.json` file as follows:

```
    {  "id": "issue4872",
       "file": "pdfs/issue4872.pdf",
       "md5": "21c6cbc682140d6f6017bbeb45892053",
       "rounds": 1,
       "link": true,
       "firstPage": 1,
       "lastPage": 1,
       "type": "eq",
       "forms": true
    },
```

I have verified that the bug also happens for that file.

test/test_manifest.json
```diff
@@ -1599,6 +1599,16 @@
        "lastPage": 6,
        "type": "eq"
     },
+    {  "id": "issue4872",
```
Snuffleupagus, line 1602: But, there's already a test with this id. Please keep in mind that every test must have a *unique* id, since otherwise you'd overwrite the existing reference images.

Hence this needs to be adjusted, e.g. to `"id": "issue4872-forms"` to prevent errors.
Also, to simplify things when looking at the `test_manifest.json` file, please place this new entry *just after* https://github.com/mozilla/pdf.js/blob/2962f4a2c5724466901428a10f4021e5047634bb/test/test_manifest.json#L1611

calixteman: About id: that was my guess.
Thx

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

Issue #12280, opened 2020-08-25
Issue when printing or saving a form containing non latin characters


calixteman:
Attach (recommended) or Link to PDF file here:
https://www.nta.go.jp/taxes/tetsuzuki/shinsei/annai/joyaku/annai/pdf2/250.pdf

Configuration:
- Web browser and its version: Firefox 81.0a1
- Operating system and its version: Debian 5.6.7-1
- PDF.js version: latest in Firefox nightly
- Is a browser extension: No

Steps to reproduce the problem:
1. Open the pdf and paste the following text "本 店 又 は 主 た る 事 務 所 の 所 在 地" in "Full name" and in "Domicile and residence" textfields
2. Print or Save the pdf 

What is the expected behavior? (add screenshot)
![image](https://user-images.githubusercontent.com/5641725/91207547-30f05480-e709-11ea-9abd-2af5d970dc46.png)

What went wrong? (add screenshot)
![image](https://user-images.githubusercontent.com/5641725/91207663-69902e00-e709-11ea-8a1d-1ddd30cf0621.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

calixteman:
FYI, I'm fixing it and I should have a patch quickly, so @timvandermeij you can assign it to me, thanks.

Issue #12226, opened 2020-08-17
Parse the `AcroForm` dictionary


timvandermeij:
We don't really use this data now, but it could contain the default appearance for form elements. It seems like the current font code for text widgets is dead as `fontRefName` is never set, so we should parse this dictionary and extract the font information from it.

Snuffleupagus:
> [...] as `fontRefName` is never set,

Also, note that's a slightly misleading name and it really ought to be renamed to `loadedName` (or possibly `fontLoadedName`) as part of fixing this issue.

Snuffleupagus:
> [...] as `fontRefName` is never set,

Also, note that's a slightly misleading name and it really ought to be changed to `loadedName` (or possibly `fontLoadedName`) as part of fixing this issue.

calixteman:
I'm not sure to understand this issue...
Don't we already do that:
https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L815-L824 ?

timvandermeij:
You're right that we not at least use the data for printing in contrast to before where we didn't use it at all. This issue is about using it for non-form rendering, so just as normal annotations. For printing we render the annotation onto the canvas, but for displaying we render an overlay element so e.g., text selection is possible. That code still doesn't work properly; look at https://github.com/mozilla/pdf.js/blob/7df8aa34a5cd4dc41964de5f942376d26ec366b6/src/display/annotation_layer.js#L489-L497 for example and notice that there is no code that actually _sets_ `fontRefName`.

timvandermeij:
Closing since this code is now replaced by working code and we use the default appearance properly.

PR #12292, opened 2020-08-27
Fix encoding issues when printing/saving a form with non-ascii characters

calixteman:
This patch aims to fix issue #12280.
I played with this japanese form:
https://www.nta.go.jp/taxes/tetsuzuki/shinsei/annai/joyaku/annai/pdf2/250.pdf
I saved a modified version of this file and open it in Acrobat Reader (osx version) successfully.

@brendandahl, @timvandermeij, @Snuffleupagus I'm not completely sure that the fonction Font::encodeString is correct so don't hesitate to help me to fix it, your feedback is highly appreciated.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -23,10 +23,12 @@ import {
   assert,
   escapeString,
   getModificationDate,
+  isAscii,
   isString,
   OPS,
   shadow,
   stringToPDFString,
+  stringToUTF16BEString,
   unreachable,
   Util,
   warn,
@@ -1222,7 +1224,7 @@ class WidgetAnnotation extends Annotation {
       appearance = newTransform.encryptString(appearance);
     }
 
-    dict.set("V", value);
+    dict.set("V", isAscii(value) ? value : stringToUTF16BEString(value));
     dict.set("AP", AP);
     dict.set("M", `D:${getModificationDate()}`);
 
@@ -1298,16 +1300,6 @@ class WidgetAnnotation extends Annotation {
     const defaultAppearance = this.data.defaultAppearance;
     const alignment = this.data.textAlignment;
 
-    if (this.data.comb) {
-      return this._getCombAppearance(
-        defaultAppearance,
-        value,
-        totalWidth,
-        hPadding,
-        vPadding
-      );
-    }
-
     if (this.data.multiLine) {
       return this._getMultilineAppearance(
         defaultAppearance,
@@ -1322,18 +1314,34 @@ class WidgetAnnotation extends Annotation {
       );
     }
 
+    // TODO: need to handle chars which are not in the font.
+    const encodedString = font.encodeString(value).join("");
+
+    if (this.data.comb) {
+      return this._getCombAppearance(
+        defaultAppearance,
+        font,
+        encodedString,
+        totalWidth,
+        hPadding,
+        vPadding
+      );
+    }
+
     if (alignment === 0 || alignment > 2) {
       // Left alignment: nothing to do
       return (
         "/Tx BMC q BT " +
         defaultAppearance +
-        ` 1 0 0 1 ${hPadding} ${vPadding} Tm (${escapeString(value)}) Tj` +
+        ` 1 0 0 1 ${hPadding} ${vPadding} Tm (${escapeString(
+          encodedString
+        )}) Tj` +
         " ET Q EMC"
       );
     }
 
     const renderedText = this._renderText(
-      value,
+      encodedString,
       font,
       fontSize,
       totalWidth,
@@ -1373,10 +1381,21 @@ class WidgetAnnotation extends Annotation {
 
   _computeFontSize(font, fontName, fontSize, height) {
     if (fontSize === null || fontSize === 0) {
-      const em = font.charsToGlyphs("M")[0].width / 1000;
-      // According to https://en.wikipedia.org/wiki/Em_(typography)
-      // an average cap height should be 70% of 1em
-      const capHeight = 0.7 * em;
+      let capHeight;
+      if (font.capHeight) {
+        capHeight = font.capHeight;
+      } else {
+        const glyphs = font.charsToGlyphs(font.encodeString("M").join(""));
+        if (glyphs.length === 1 && glyphs[0].width) {
+          const em = glyphs[0].width / 1000;
+          // According to https://en.wikipedia.org/wiki/Em_(typography)
+          // an average cap height should be 70% of 1em
+          capHeight = 0.7 * em;
+        } else {
+          capHeight = 0.7;
+        }
+      }
+
       // 1.5 * capHeight * fontSize seems to be a good value for lineHeight
       fontSize = Math.max(1, Math.floor(height / (1.5 * capHeight)));
 
@@ -1510,11 +1529,12 @@ class TextWidgetAnnotation extends WidgetAnnotation {
       this.data.maxLen !== null;
   }
 
-  _getCombAppearance(defaultAppearance, text, width, hPadding, vPadding) {
+  _getCombAppearance(defaultAppearance, font, text, width, hPadding, vPadding) {
     const combWidth = (width / this.data.maxLen).toFixed(2);
     const buf = [];
-    for (const character of text) {
-      buf.push(`(${escapeString(character)}) Tj`);
+    const positions = font.getCharPositions(text);
+    for (const [start, end] of positions) {
+      buf.push(`(${escapeString(text.substring(start, end))}) Tj`);
     }
 
     const renderedComb = buf.join(` ${combWidth} 0 Td `);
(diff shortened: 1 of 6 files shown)


Snuffleupagus (changes requested):
Really sorry, but I don't think that the changes you're making in `src/core/primitives.js` (and its usages) are at all correct or desirable here (it's simply *too* hacky in my opinion); this is really not how a `Dict` is supposed to work!

The correct solution might be to fix issue #12226 *first*, see the in-progress patch in PR #12207 (which I believe will require a fair number of changes before it's ready to land), since that being fixed *correctly* should mean that the `AcroForm` font-resources are available in a more standardized way.

(Also, merging of `Dict` data needs to be done *very* carefully if that's actually necessary.)

---

Other than that, I've not really looked at the rest of the patch since I believe that the above is enough of a blocker.

calixteman:
Agreed, I forgot to move it in another patch.
Anyway the issue with this form is that the font we have in DA is /MSGothic and it isn't in DR but in Acroform::DR which is obviously an error.
I will fix this issue in another patch.

calixteman:
I removed the DR issue fix so there is non-buggy textfield to play with: "Domicile or residence" one on the first page.

calixteman:
I filed an issue for the DA/DR problem:
https://github.com/mozilla/pdf.js/issues/12294

Snuffleupagus:
> I filed an issue for the DA/DR problem:
> #12294

But isn't that just another version of issue #12226, as mention above!?

timvandermeij:
Overall this patch looks good to me, but I'm having trouble reviewing the `fonts.js` code since I'm not really familiar with it. Maybe @Snuffleupagus or @brendandahl can chime in for that?

src/core/annotation.js
```diff
@@ -1286,37 +1297,42 @@ class WidgetAnnotation extends Annotation {
...
     const [font, fontName] = fontInfo;
-    const fontSize = this._computeFontSize(...fontInfo, totalHeight);
     this._fontName = fontName;
 
-    let descent = font.descent;
-    if (isNaN(descent)) {
-      descent = 0;
-    }
+    const lineHeight = totalHeight;
+    const fontSize = this._computeFontSize(...fontInfo, lineHeight);
```
Snuffleupagus, line 1303: The `lineHeight` doesn't seem to be used for anything else here; can you please revert these changes to reduce the overall size of the diff?

src/core/annotation.js
```diff
@@ -1286,37 +1297,42 @@ class WidgetAnnotation extends Annotation {
...
+        totalHeight,
+        alignment,
         hPadding,
         vPadding
       );
     }
 
-    if (this.data.multiLine) {
-      return this._getMultilineAppearance(
+    // TODO: need to handle chars which are not in the font
```
Snuffleupagus, line 1327: Nit: Period at the end of sentences please!

src/core/annotation.js
```diff
@@ -1568,49 +1601,59 @@ class TextWidgetAnnotation extends WidgetAnnotation {
   }
 
   _splitLine(line, font, fontSize, width) {
-    if (line.length <= 1) {
+    // TODO: need to handle chars which are not in the font
```
Snuffleupagus, line 1604: Nit: Period at the end of sentences please!

src/core/annotation.js
```diff
@@ -1373,10 +1391,24 @@ class WidgetAnnotation extends Annotation {
...
-      // According to https://en.wikipedia.org/wiki/Em_(typography)
-      // an average cap height should be 70% of 1em
-      const capHeight = 0.7 * em;
+      let capHeight;
+      if (font.capHeight) {
+        capHeight = font.capHeight;
+      } else {
+        const glyphs = font.charsToGlyphs(
+          font.encodeString("M").join(""),
+          true
```
Snuffleupagus, line 1400: That method doesn't take more than one parameter, why are you passing `true` here!?

src/core/cmap.js
```diff
@@ -338,6 +338,22 @@ class CMap {
...
+      const codespaceRange = codespaceRanges[n];
+      for (let k = 0, kk = codespaceRange.length; k < kk; ) {
+        const low = codespaceRange[k++];
+        const high = codespaceRange[k++];
+        if (charCode >= low && charCode <= high) {
+          return n + 1;
+        }
+      }
+    }
+    return 0;
```
Snuffleupagus, line 354: Is this default value correct, since the `readCharCode` method above returns `length = 1` in the fallback case?

src/core/fonts.js
```diff
@@ -3351,9 +3352,119 @@ var Font = (function FontClosure() {
       return (charsCache[charsCacheKey] = glyphs);
     },
 
+    getGlyphsAndCharPosition: function Font_getGlyphsAndCharPosition(chars) {
```
Snuffleupagus, line 3355: Nit: Here and elsewhere below, when adding *new* code you can simply format methods as `getGlyphsAndCharPosition(chars) {` instead.

src/core/fonts.js
```diff
@@ -3351,9 +3352,119 @@ var Font = (function FontClosure() {
...
+        }
+      }
+
+      return glyphs;
+    },
+
+    getCharPosition: function Font_getCharPosition(chars) {
+      // Chars can have different sizes
+      // so get the char positions in the string
+      // it's useful to split a string at correct char positions.
```
Snuffleupagus, line 3391: This isn't all that easy to ready, since it feels like you're using line-breaks instead of proper punctuation here.
Also, should this information perhaps be a JSDoc comment instead?

src/core/fonts.js
```diff
@@ -3351,9 +3352,119 @@ var Font = (function FontClosure() {
...
+    },
+
     get glyphCacheValues() {
       return Object.values(this.glyphCache);
     },
+
+    encodeString: function Font_encodeString(str) {
+      // The function returns an array of strings
+      // at even positions there is a valid string (may be empty)
+      // and at odd positions string where chars are not in the font
```
Snuffleupagus, line 3419: Again this isn't all that easy to ready, since it feels like you're using line-breaks instead of proper Spunctuation here.
Also, should this information perhaps be a JSDoc comment instead?

src/core/fonts.js
```diff
@@ -3351,9 +3352,119 @@ var Font = (function FontClosure() {
...
+      const buffers = [];
+      let buf = [];
+      let isError = false;
+
+      for (let i = 0, ii = str.length; i < ii; i++) {
+        const unicode = str.codePointAt(i);
+        let hasError = false;
+        if (unicode > 0xd7ff && (unicode < 0xe000 || unicode > 0xfffd)) {
+          // unicode is represented by two u16
```
Snuffleupagus, line 3428: What is `u16`? Please improve/expand this comment, since it's not all that clear currently!

src/core/fonts.js
Snuffleupagus: Also, this is an *almost* verbatim copy of the existing `charsToGlyphs` method, which feels quite unfortunate!
Furthermore, this ignores the existing caching that's used in that method, which can lead to quite a bit of unnecessary re-parsing here.

Given how this is used, with only one call-site as far as I can tell, could you not simply call both the *existing* `charsToGlyphs` and then the *new* `getCharPosition` methods in place of this "combine" one!?
(Considering the current lack of caching in the new methods, I really don't see that being much slower overall.)

src/core/fonts.js
Snuffleupagus: Also, you probably want to add a *lazily* initialized cache for the data (something similar to `charsToGlyphs` above), to avoid unnecessary re-parsing here.

src/core/fonts.js
```diff
@@ -3351,9 +3352,119 @@ var Font = (function FontClosure() {
...
+        }
+        if (this.toUnicode) {
+          const char = String.fromCodePoint(unicode);
+          const charCode = this.toUnicode.charCodeOf(char);
+          if (charCode === -1) {
+            hasError = true;
+          } else {
+            if (isError) {
+              buffers.push(buf.join(""));
+              buf = [];
```
Snuffleupagus, line 3439: By doing `buf.length = 0`, you can re-use the *same* Array rather than creating a new one (which should be more efficient overall).

src/core/fonts.js
```diff
@@ -3351,9 +3352,119 @@ var Font = (function FontClosure() {
...
       return Object.values(this.glyphCache);
     },
+
+    encodeString: function Font_encodeString(str) {
+      // The function returns an array of strings
+      // at even positions there is a valid string (may be empty)
+      // and at odd positions string where chars are not in the font
+      const buffers = [];
+      let buf = [];
+      let isError = false;
```
Snuffleupagus, line 3422: The very similarly named `isError`/`hasError` variables makes this code, at first sight, a bit difficult to follow efficiently.

Snuffleupagus (changes requested):
I've added a couple of comments, based on an very cursory look at the patch.

src/core/fonts.js
calixteman: You're right: I removed that stuff.

src/core/fonts.js
```diff
@@ -3351,9 +3352,88 @@ var Font = (function FontClosure() {
...
+     * Encode a js string using font encoding.
+     * The resulting array contains an encoded string at even positions
+     * (can be empty) and a non-encoded one at odd positions.
+     * @param {String} a js string.
+     * @returns {Array<String>} an array of encoded strings or non-encoded ones.
+     */
+    encodeString(str) {
+      const buffers = [];
+      const currentBuf = [];
+      const hasCurrentBufErrors = () => buffers.length % 2 === 1;
```
Snuffleupagus, line 3397: The name of this helper function suggests (to me) that the intention is to check the `currentBuf` array, but that's not actually happening here; if this is indeed intended behaviour a comment might not hurt here!

Snuffleupagus (commented):
Also, looking at the previous test-results there's some *vertical* movement of text in existing `print`-tests; was this an intentional change here?

Snuffleupagus (approved):
I've not looked in detail at the unit-tests, but overall the `core` (and importantly `fonts.js`) parts seem reasonable now :-)

I don't know if either Brendan or Tim wants to give this a final once-over, before we land it?

timvandermeij (approved):

timvandermeij:
Looks good to me too. Thanks!

PR #12293, opened 2020-08-27
Download, rather than opening, PDF attachments in Firefox (bug 1661259, PR 12286 follow-up)

Snuffleupagus:
Unfortunately the work-around implemented in PR #12286 didn't actually work in all cases, please refer to the previous commit messages.
To prevent opening of PDF attachments from being completely broken for some users, we'll simply force-download them for now in MOZCENTRAL-builds to unbreak things. (Given that the "Open with" dialog now features a "Open with Firefox"-option, this is less bad than it previously would've been.)

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1661259


timvandermeij (approved):

timvandermeij:
Thank you; that does seem like the best approach for now.

Issue #12294, opened 2020-08-27
Invalid DA for some text annotations which use a font not in DR


calixteman:
Attach (recommended) or Link to PDF file here:
https://www.nta.go.jp/taxes/tetsuzuki/shinsei/annai/joyaku/annai/pdf2/250.pdf

This pdf contains some text annotations with (e.g. "Full name" field on the 1st page):
 - `DA = "/MSGothic 8.5 Tf 0 g"`
 - `DR = << /Font << /KozMinPr6N-Regular 2867 0 R >> >>`

When trying to print/save such a modified textfield then it isn't rendered correctly because the font is not found in DR.
According to the specs it seems that it's an error.
Since `/MSGothic` is present in Acroform::DR, then a way to fix this issue is to use a fallback dict  when a resource is not available.

Do we want to fix this issue with such a workaround (I mean fix the DR: it tryed and it works well) ?
Or do we want to wait for a more general fix to handle chars which are not in the font coming from DA ?
Or anything else ?

Snuffleupagus:
> Since `/MSGothic` is present in Acroform::DR, then a way to fix this issue is to use a fallback dict when a resource is not available.

The correct solution, as mentioned in https://github.com/mozilla/pdf.js/pull/12292#pullrequestreview-476774356, is to first fix issue #12226 and then *correctly* merge the two different `DR` dictionaries.

waluyo102:
Is good

timvandermeij:
Closing since this is fixed by #12361.

PR #12296, opened 2020-08-27
Add support, in `Dict.merge`, for merging of "sub"-dictionaries

Snuffleupagus:
This allows for merging of dictionaries one level deeper than previously. This could be useful e.g. for /Resources dictionaries, where you want to e.g. merge their respective /Font dictionaries (and other) together rather than picking just the first one.

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -390,4 +390,13 @@ if (
     }
     Object.values = require("core-js/es/object/values.js");
   })();
+
+  // Provides support for Object.entries in legacy browsers.
+  // Support: IE, Chrome<54
+  (function checkObjectEntries() {
+    if (Object.entries) {
+      return;
+    }
+    Object.entries = require("core-js/es/object/entries.js");
+  })();
 }
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -114,7 +114,7 @@ class Page {
     if (value.length === 1 || !isDict(value[0])) {
       return value[0];
     }
-    return Dict.merge(this.xref, value);
+    return Dict.merge({ xref: this.xref, dictArray: value });
   }
 
   get content() {
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -726,8 +726,10 @@ class PartialEvaluator {
     const tilingOpList = new OperatorList();
     // Merge the available resources, to prevent issues when the patternDict
     // is missing some /Resources entries (fixes issue6541.pdf).
-    const resourcesArray = [patternDict.get("Resources"), resources];
-    const patternResources = Dict.merge(this.xref, resourcesArray);
+    const patternResources = Dict.merge({
+      xref: this.xref,
+      dictArray: [patternDict.get("Resources"), resources],
+    });
 
     return this.getOperatorList({
       stream: pattern,
--- a/src/core/primitives.js
+++ b/src/core/primitives.js
@@ -173,22 +173,61 @@ var Dict = (function DictClosure() {
 
   Dict.empty = new Dict(null);
 
-  Dict.merge = function (xref, dictArray) {
+  Dict.merge = function ({ xref, dictArray, mergeSubDicts = false }) {
     const mergedDict = new Dict(xref);
 
-    for (let i = 0, ii = dictArray.length; i < ii; i++) {
-      const dict = dictArray[i];
-      if (!isDict(dict)) {
-        continue;
-      }
-      for (const keyName in dict._map) {
-        if (mergedDict._map[keyName] !== undefined) {
+    if (!mergeSubDicts) {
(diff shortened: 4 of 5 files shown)


src/core/annotation.js
```diff
@@ -818,10 +818,15 @@ class WidgetAnnotation extends Annotation {
       "";
     const fieldType = getInheritableProperty({ dict, key: "FT" });
     data.fieldType = isName(fieldType) ? fieldType.name : null;
-    this.fieldResources =
-      getInheritableProperty({ dict, key: "DR" }) ||
-      params.acroForm.get("DR") ||
-      Dict.empty;
+
+    this.fieldResources = Dict.merge({
```
calixteman, line 822: In general I'm fine with this change but when saving a form, we copy the resources dictionary:
https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L1040
So at the end it'll increase the file size uselessly, especially if the pdf is well formated.
That lets me think that we don't need to add fieldResources when this one is coming from Acroform...
To avoid that, my idea, when fixing https://github.com/mozilla/pdf.js/issues/12294, was to have a kind of `lazy` merge:
just put in the dict what we really need.

Snuffleupagus: > So at the end it'll increase the file size uselessly, especially if the pdf is well formated.

Given that the resources are usually `Ref`s, that should hopefully reduce the size-impact in general; however I do agree that avoiding this if at all possible would be preferable.

> That lets me think that we don't need to add fieldResources when this one is coming from Acroform...

Assuming that you can detect this case easily, and implement it without mutating any *existing* `Dict`s, that could perhaps work (although I'm not entire sure I understand exactly what you have in mind). Please explore this idea further!

(If you do need to merge `Dict`s, with sub-resources, this patch provides a correct way of doing so.)

> To avoid that, my idea, when fixing #12294, was to have a kind of `lazy` merge:
> just put in the dict what we really need.

The problem was with *how* it was implemented, since you unfortunately added special-cases (and dare I say hacks) to the *general* `Dict` abstraction used throughout the code-base to satisfy just *one* very specific use-case.
Hence my objections, since that wasn't really maintainable and also would have meant mutating existing `Dict`s in a way that you absolutely shouldn't do.

---

I'll remove the `src/core/annotation.js` changes from this patch, but I think that it's probably generally useful to have it regardless.

calixteman: > > So at the end it'll increase the file size uselessly, especially if the pdf is well formated.
> 
> Given that the resources are usually `Ref`s, that should hopefully reduce the size-impact in general; however I do agree that avoiding this if at all possible would be preferable.
> 
> > That lets me think that we don't need to add fieldResources when this one is coming from Acroform...
> 
> Assuming that you can detect this case easily, and implement it without mutating any _existing_ `Dict`s, that could perhaps work (although I'm not entire sure I understand exactly what you have in mind). Please explore this idea further!
I'll file a bug for that: my idea is to reduce the file size when saving.

> 
> (If you do need to merge `Dict`s, with sub-resources, this patch provides a _correct_ way of doing so.)
> 
> > To avoid that, my idea, when fixing #12294, was to have a kind of `lazy` merge:
> > just put in the dict what we really need.
> 
> The problem was with _how_ it was implemented, since you unfortunately added special-cases (and dare I say hacks) to the _general_ `Dict` abstraction used throughout the code-base to satisfy just _one_ very specific use-case.
> Hence my objections, since that wasn't really maintainable and also would have meant mutating existing `Dict`s in a way that you absolutely shouldn't do.

Agreed !
Maybe we could create an object LazyMergedDict (extends Dict) which will keep a list of dicts. When we make a `get`, we have a look in each dict until we find it and then we add the key+value in the dict.
Wdyt ? do you have other ideas ?

> 
> I'll remove the `src/core/annotation.js` changes from this patch, but I think that the better `Dict.merge` functionality is probably generally useful to have regardless.

src/shared/compatibility.js
```diff
@@ -390,4 +390,13 @@ if (
     }
     Object.values = require("core-js/es/object/values.js");
   })();
+
+  // Provides support for Object.entires in legacy browsers.
```
timvandermeij, line 394: `entires` -> `entries`

src/shared/compatibility.js
```diff
@@ -390,4 +390,13 @@ if (
...
+  // Support: IE, Chrome<54
+  (function checkObjectValues() {
```
timvandermeij, line 396: `checkObjectEntries`

dimas301:
Please stop

Pada tanggal Sen, 31 Agt 2020 02.28, Tim van der Meij <
notifications@github.com> menulis:

timvandermeij (approved):

timvandermeij:
Thank you for implementing support for this!

Issue #8402, opened 2017-05-11
Verify if `web/viewer.js` should be a module as well


yurydelendik:
See #7916.

(Converted and removed from ES6 modules project)

It's mostly blocked by our PDFViewerApplication type/global usage. We need to have application class (e.g. Application) that will accept some configuration which allow multiple instances of application to be hosted on the same page. DOM event bindings can be done via some manager (ApplicationManager).

PR #12297, opened 2020-08-28
[WIP] Port web/app.js to class style

Priestch:
Should we port global `PDFViewerApplication` to class style? 

I want to show multiple PDF documents in one web page, so I hacked the whole` web/app.js` to class style, and other related issues(DOM selectors, AppOptions etc.) to create multiple instance. Although suffers some global window event issues, but basically it can show multiple documents. I wonder if class style `PDFViewerApplication` is good for pdf.js, if so, I can do the port work.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -179,65 +179,72 @@ class DefaultExternalServices {
   }
 }
 
-const PDFViewerApplication = {
-  initialBookmark: document.location.hash.substring(1),
-  _initializedCapability: createPromiseCapability(),
-  fellback: false,
-  appConfig: null,
-  pdfDocument: null,
-  pdfLoadingTask: null,
-  printService: null,
-  /** @type {PDFViewer} */
-  pdfViewer: null,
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
> Should we port global `PDFViewerApplication` to class style?

Given that we've closed issue #8402 as WONTFIX, since this would be completely unnecessary for Firefox and also would add a lot of unneeded complexity/code-churn, this answer at this point in time is thus "no".

> I want to show multiple PDF documents in one web page,

You could already do that, simply by placing each viewer in an iframe-element.

Finally, please also note http://mozilla.github.io/pdf.js/getting_started/#introduction (emphasis mine):

Priestch:
@Snuffleupagus There are some more details I did not mention, I also have to add some annotation like functions to the PDF viewer, so use iframe is too complex for code share among multiple independent projects.

Anyway, the class style code is more simpler to me.

PR #12299, opened 2020-08-28
Update the indeterminate progressBar to the new design (PR 11077 follow-up)

Snuffleupagus:
This special progressBar is only used in the (fortunately) rare case when a server doesn't provide a valid `Content-Length` header. Since this progressBar isn't normally seen, when testing the default viewer, it's certainly very easy to see why these CSS rules were missed during review.

Furthermore, this patch also makes a couple of *small* progressBar CSS tweaks not related to the colours.


timvandermeij (approved):

timvandermeij:
Nice find!

PR #12301, opened 2020-08-29
Add (basic) support for Stylelint, to allow linting of CSS files

Snuffleupagus:
Add (basic) support for Stylelint, to allow linting of CSS files

This is *similar* to the existing linting for JavaScript files, but covers CSS files instead.
While there's a lot of rules that could potentially be used, the main advantage of using Stylelint is that it has Prettier integration which means that we can automatically enforce a *consistent* style for our CSS files as well.

As a proof of concept, this patch is purposely limited to:
 - Adding a simple rule, here `block-no-empty` is chosen; see https://stylelint.io/user-guide/rules/block-no-empty
 - Adding Prettier integration, to unify the style of our CSS files.

Please find additional information at https://stylelint.io/


timvandermeij (approved):

timvandermeij:
This is a nice idea. Thanks!

PR #12302, opened 2020-08-30
Fixed bug of "cannot read property 'div' of undefined"

Gorgchap:
image
There are two different single page pdf documents (this mistake is at the second one)

### DIFF
--- a/web/pdf_single_page_viewer.js
+++ b/web/pdf_single_page_viewer.js
@@ -49,7 +49,9 @@ class PDFSinglePageViewer extends BaseViewer {
     const viewerNodes = this.viewer.childNodes;
     switch (viewerNodes.length) {
       case 0: // Should *only* occur on initial loading.
-        this.viewer.appendChild(pageView.div);
+        if (pageView !== undefined) {
+          this.viewer.appendChild(pageView.div);
+        }
         break;
       case 1: // The normal page-switching case.
         if (viewerNodes[0] !== previousPageView.div) {


Snuffleupagus (changes requested):
That shouldn't actually happen though, and why is *this* particular `case` special here?
Sorry, but the patch just feels like an attempt to "wallpaper over" an actual problem that lies *elsewhere*.

Please provide clear and concise steps to reproduce the problem, or ideally a runnable test-case (see below), because as-is this unfortunately cannot be accepted.

---

Please see https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md (emphasis mine):
> If you are developing a custom solution, first check the examples at https://github.com/mozilla/pdf.js#learning and search existing issues. If this does not help, **please prepare a short well-documented example that demonstrates the problem and make it accessible online** on your website, JS Bin, GitHub, etc. before opening a new issue or contacting us in the Matrix room -- **keep in mind that just code snippets won't help us troubleshoot the problem**. cannot be accepted.

Gorgchap:
@Snuffleupagus 
This problem occurs at using of these libraries: https://github.com/viktorhajer/simple-pdf-viewer, https://github.com/VadimDez/ng2-pdf-viewer (I use the first one). Both of them use this lib. In my project there is an opportunity to create two different pdf-documents and add then to single HTML page (as it shown at screenshot).

Here is a link to issue: https://github.com/VadimDez/ng2-pdf-viewer/issues/224

Snuffleupagus:
> This problem occurs at using of these libraries: https://github.com/viktorhajer/simple-pdf-viewer, https://github.com/VadimDez/ng2-pdf-viewer (I use the first one).

Those are *different* libraries, that we know nothing about here and consequently cannot provide support for.

> Both of them use this lib.

Since those projects are unrelated to the PDF.js library, it's possible that they're doing something wrong and thus triggering a problem (has happen in some cases).
Can you reproduce the problem when using the PDF.js library *directly*, without another library on top of it?

> In my project there is an opportunity to create two different pdf-documents and add then to single HTML page (as it shown at screenshot).

As mentioned above, please provide a *runnable* example. Without this, it's impossible to determine what the *correct* solution is and as mentioned we unfortunately cannot accept this as-is.

> Here is a link to issue: [VadimDez/ng2-pdf-viewer#224](https://github.com/VadimDez/ng2-pdf-viewer/issues/224)

That issue is from a *different* project, and crucially it also mentions a couple of different errors (based on the differing stack-traces).

Snuffleupagus:
> This problem occurs at using of these libraries: https://github.com/viktorhajer/simple-pdf-viewer, https://github.com/VadimDez/ng2-pdf-viewer (I use the first one).

Looking at https://github.com/viktorhajer/simple-pdf-viewer/blob/e2d7b70f9eebbe6b08b4c8cdddb6e05bf885ed86/libs/simple-pdf-viewer/src/simplePdfViewer.component.ts#L417-L421, this is very clearly a bug in that project (as I suspected here).
When initializing a `PDFSinglePageViewer`-instance (or a `PDFViewer`-instance for that matter), you *must* provide an `EventBus`-instance; please refer to the *official* PDF.js example in https://github.com/mozilla/pdf.js/blob/master/examples/components/singlepageviewer.js

Note that the code in https://github.com/viktorhajer/simple-pdf-viewer is already broken if used with the *latest* version of the PDF.js library, see https://github.com/mozilla/pdf.js/releases, and keep in mind that every `PDFViewer`/`PDFSinglePageViewer` should be initialized with its own `EventBus`-instance.

All-in-all, there's nothing wrong with the PDF.js project and the other project should adapt such that things work correctly.

PR #12310, opened 2020-09-01
Add support for /Print and /SaveAs named actions in the viewer (issue 12308)

### LINKED ISSUE #12308: PDF - button(s) from in-page forms don't trigger actions
**Note**
* cross-checking with the interaction of the buttons revealed that, in fact the Print button would trigger the printing action in Chrome/Edge but not in Safari/Firefox;
* the save button doesn't do anything in any browser;

**Affected versions**
* 81.0b4, 82.0a1 (2020-08-31);

**Affected platforms**
* macOS 10.15, Windows 10;

**Steps to reproduce**
pdfjs.renderInteractiveForms - true

1. Launch Firefox;
2. Access https://royalegroupnyc.com/wp-content/uploads/seating_areas/sample_pdf.pdf
3. Scroll down;
4. Click on the Save/Print button(s);

**Expected result**
* Print/Save button should trigger the Print/Save action(s);

**Actual result**
* nothing happens;

**Regression range**
* not a regression;

**Additional notes**
* S2/S3 as suggested severity.

Snuffleupagus:
All of the core/display functionality needed to support this already exists, we simply need to handle these named actions in the viewer and the buttons will "just" work.

Unfortunately there's not really any good way of (automatically) testing this, but given the size and scope of the patch that's hopefully OK.

Fixes #12308

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2210,10 +2210,9 @@ function webViewerPageMode({ mode }) {
 }
 
 function webViewerNamedAction(evt) {
-  // Processing couple of named actions that might be useful.
-  // See also PDFLinkService.executeNamedAction
-  const action = evt.action;
-  switch (action) {
+  // Processing a couple of named actions that might be useful, see also
+  // `PDFLinkService.executeNamedAction`.
+  switch (evt.action) {
     case "GoToPage":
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Works nicely; thanks!

PR #12314, opened 2020-09-01
Add support for lazy merging dictionaries in annotations

calixteman:
It aims to fix for example issue #12294.
The merge have to be lazy to avoid to write too much potentially useless data when saving a form.

### DIFF
--- a/src/core/primitives.js
+++ b/src/core/primitives.js
@@ -91,11 +91,11 @@ var Dict = (function DictClosure() {
 
     // automatically dereferences Ref objects
     get(key1, key2, key3) {
-      let value = this._map[key1];
+      let value = this.getRaw(key1);
       if (value === undefined && key2 !== undefined) {
-        value = this._map[key2];
+        value = this.getRaw(key2);
         if (value === undefined && key3 !== undefined) {
-          value = this._map[key3];
+          value = this.getRaw(key3);
         }
       }
       if (value instanceof Ref && this.xref) {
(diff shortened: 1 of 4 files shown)


src/core/primitives.js
```diff
@@ -91,11 +91,11 @@ var Dict = (function DictClosure() {
 
     // automatically dereferences Ref objects
     get(key1, key2, key3) {
-      let value = this._map[key1];
```
timvandermeij, line 94: I believe these primitives are called _a lot_, so introducing extra function calls here might cause a performance regression. If we do this, it will have to be benchmarked carefully.

timvandermeij (commented):
@Snuffleupagus Do you perhaps have time to review this patch since you looked into the dictionary primitive before? I have the feeling that the functionality may be useful, but I'm unsure about the implementation.

src/core/primitives.js
calixteman: Agreed.
My first try was to use a Proxy object which should be fine to avoid any kind of perf regression but I didn't find a polyfill for it (according to MDN it isn't supported with IE) and so I gave up this solution.
But I guess that it should be possible now to use this solution since IE support has been removed.
Wdyt ?

Snuffleupagus: I think we should aim for a *much* simpler solution, specifically for the case at hand here, rather than trying to extend a primitive such as `Dict` since there's other problems with this patch as far as I can tell.
E.g. the `getLazyMergeDict` function is mutating the provided `dict` parameter, which could *very* easily lead to issues elsewhere and even outright bugs if used without the out-most care (since you'd not expect that kind of behaviour).

Since I hack on PDF.js in my spare time, I've not yet had the time to look into all of this, but hopefully will later this week.

timvandermeij:
Closing since this is replaced by #12361.

PR #12315, opened 2020-09-02
print: Make the firefox printing code able to fail and be re-invoked.

emilio:
This fixes a set of issues described in Mozilla bug 1662426[1].

In particular, once the print callback fails once (because the printing
operation has been canceled in Gecko / replaced by a newer one, for example) it
can't be re-invoked.

This patch fixes it by properly cancelling the render task if it throws, or if
the print callback is called again while ongoing.

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1662426

### DIFF
--- a/web/firefox_print_service.js
+++ b/web/firefox_print_service.js
@@ -41,6 +41,14 @@ function composePage(
   canvasWrapper.appendChild(canvas);
   printContainer.appendChild(canvasWrapper);
 
+  // A callback for a given page may be executed multiple times for different
+  // print operations (think of changing the print settings in the browser).
+  //
+  // Since we don't support queueing multiple render tasks for the same page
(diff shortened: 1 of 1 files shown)


web/firefox_print_service.js
```diff
@@ -41,6 +41,7 @@ function composePage(
   canvasWrapper.appendChild(canvas);
   printContainer.appendChild(canvasWrapper);
 
+  let currentRenderTask = null;
```
brendandahl, line 44: Can you add a comment here why we have `currentRenderTask` and  `thisRenderTask`?

brendandahl (approved):
r+ with comment

emilio:
I added a comment, let me know if you want me to tweak it :)

PR #12321, opened 2020-09-03
Ensure that the `viewer` property, on `BaseViewer`-instances, is a valid div-element (issue 12320)

Snuffleupagus:
This should help prevent future issues, caused by the user omitting the `viewer` option and/or providing an incorrect `container` option, when initializing a `BaseViewer`-instance.


timvandermeij (approved):

timvandermeij:
Good idea! Hopefully this will prevent some issues.

PR #12326, opened 2020-09-04
Revert "Download, rather than opening, PDF attachments in Firefox (bug 1661259, PR 12286 follow-up)"

Snuffleupagus:
This reverts commit 2a0de0b66b41c345cd357cbb828b3ca6f49fd395.

I can no longer reproduce these issues locally, and if ad blockers are still interfering with this functionality we really ought to pursue a mozilla-central solution to the problem instead. (Also, I'm no longer getting an "Open with Firefox"-option in the "Open with"-dialog making the PDF attachments experience worse for all users.)

### DIFF
--- a/web/pdf_attachment_viewer.js
+++ b/web/pdf_attachment_viewer.js
@@ -172,7 +172,6 @@ class PDFAttachmentViewer extends BaseTreeViewer {
 
       const element = document.createElement("a");
       if (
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
I agree; thanks!

PR #12328, opened 2020-09-05
Remove code-paths only relevant for IE 11/Edge (non-Chromium based) from the `web/` folder

Snuffleupagus:
This patch purposely starts small, by removing IE-specific code from various JS/CSS files in the `web/` folder.

There's obviously lots of potential for additional clean-up, especially the removal of no longer necessary polyfills in `src/shared/compatibility.js`, however that will require some care considering that certain polyfills may also be necessary for e.g. Node.js or the Chromium-extension as well.
Generally speaking, once we start removing polyfills it's probably a good idea to consult the compatibility information on https://developer.mozilla.org/ and also https://caniuse.com/ first. (Deciding on the lowest supported Chromium version, for the extension, would also seem like a good idea.)

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -611,15 +611,13 @@ const PDFViewerApplication = {
       support = !!(
         doc.requestFullscreen ||
         doc.mozRequestFullScreen ||
-        doc.webkitRequestFullScreen ||
-        doc.msRequestFullscreen
+        doc.webkitRequestFullScreen
       );
 
       if (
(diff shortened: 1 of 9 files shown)


timvandermeij (approved):

timvandermeij:
Nice clean-up!

mr1260356:
Cara membuat materi pdf gmn

PR #12329, opened 2020-09-05
Rename the `--outline-...` CSS variables to `--treeitem-...` instead (PR 11077 follow-up)

Snuffleupagus:
Given that the outlineView/attachmentsView/layersView all share a common base-class and CSS rules, see PRs 12169 and 12170, the names of the CSS variables in question feels slightly strange now.


timvandermeij (approved):

timvandermeij:
Looks better!

PR #12331, opened 2020-09-06
[api-minor] Only support browsers/environments that have *basic* support for `Promise` natively

Snuffleupagus:
Based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Browser_compatibility and https://caniuse.com/#feat=promises, all even remotely modern browsers already support *basic* `Promise` functionality natively.

The only reason for keeping the `Promise` polyfill (at all) is to be able to support recent additions to the specification, such as e.g. `finally` and `allSettled`.
Note that this patch will, on its own, remove support for IE 11/Edge (non-Chromium based) in both the general PDF.js library and the default viewer.


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #12333, opened 2020-09-06
Add tooltip if any in annotations layer

calixteman:
This patch aims to fix:
https://bugzilla.mozilla.org/show_bug.cgi?id=1661419

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -1533,6 +1533,7 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       this.hasFieldFlag(AnnotationFieldFlag.RADIO) &&
       !this.hasFieldFlag(AnnotationFieldFlag.PUSHBUTTON);
     this.data.pushButton = this.hasFieldFlag(AnnotationFieldFlag.PUSHBUTTON);
+    this.data.isTooltipOnly = false;
 
     if (this.data.checkBox) {
       this._processCheckBox(params);
@@ -1770,10 +1771,12 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
   }
 
   _processPushButton(params) {
-    if (!params.dict.has("A")) {
+    if (!params.dict.has("A") && !this.data.alternativeText) {
       warn("Push buttons without action dictionaries are not supported");
       return;
     }
+
+    this.data.isTooltipOnly = !params.dict.has("A");
 
     Catalog.parseDestDictionary({
       destDict: params.dict,
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -293,7 +293,8 @@ class LinkAnnotationElement extends AnnotationElement {
     const isRenderable = !!(
       parameters.data.url ||
       parameters.data.dest ||
-      parameters.data.action
+      parameters.data.action ||
+      parameters.data.isTooltipOnly
     );
     super(parameters, isRenderable);
   }
@@ -322,8 +323,10 @@ class LinkAnnotationElement extends AnnotationElement {
       });
     } else if (data.action) {
       this._bindNamedAction(link, data.action);
-    } else {
+    } else if (data.dest) {
       this._bindLink(link, data.dest);
+    } else {
+      this._bindLink(link, "");
     }
 
     this.container.appendChild(link);
@@ -420,6 +423,10 @@ class WidgetAnnotationElement extends AnnotationElement {
    */
   render() {
     // Show only the container for unsupported field types.
+    if (this.data.alternativeText) {
+      this.container.title = this.data.alternativeText;
+    }
+
     return this.container;
   }
 }
@@ -647,6 +654,11 @@ class PushButtonWidgetAnnotationElement extends LinkAnnotationElement {
     // as performing actions on form fields (resetting, submitting, et cetera).
     const container = super.render();
     container.className = "buttonWidgetAnnotation pushButton";
+
+    if (this.data.alternativeText) {
+      container.title = this.data.alternativeText;
+    }
+
     return container;
   }
 }
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -2436,6 +2436,49 @@ describe("annotation", function () {
         })
         .catch(done.fail);
     });
+
+    it("should handle push buttons", function (done) {
+      const buttonWidgetRef = Ref.get(124, 0);
+      buttonWidgetDict.set("Ff", AnnotationFieldFlag.PUSHBUTTON);
+      buttonWidgetDict.set("A", "whatever");
+
+      const xref = new XRefMock([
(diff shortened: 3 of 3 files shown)


src/display/annotation_layer.js
```diff
@@ -165,6 +165,10 @@ class AnnotationElement {
     let width = data.rect[2] - data.rect[0];
     let height = data.rect[3] - data.rect[1];
 
+    if (this.data.alternativeText) {
```
timvandermeij, line 168: Alternative text is only defined on widget annotations, so in this file this code should also be in the `WidgetAnnotationElement` class. I would say you can put it in https://github.com/mozilla/pdf.js/blob/2f3162167d664208716013486d6697689d1d08a4/src/display/annotation_layer.js#L429-L432 since you have `this.container` there.

src/display/annotation_layer.js
```diff
@@ -165,6 +165,10 @@ class AnnotationElement {
...
+      container.setAttribute("title", this.data.alternativeText);
```
timvandermeij, line 169: I think you can simplify this to `container.title = this.data.alternativeText;`. We should only use `setAttribute` for non-default attributes.

src/core/annotation.js
```diff
@@ -1366,6 +1366,7 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
 
     this.checkedAppearance = null;
     this.uncheckedAppearance = null;
+    this.data.isTooltipOnly = false;
```
timvandermeij, line 1369: Let's move this below the newline so all `this.data` setting is combined.

src/display/annotation_layer.js
```diff
@@ -293,8 +297,10 @@ class LinkAnnotationElement extends AnnotationElement {
     const isRenderable = !!(
       parameters.data.url ||
       parameters.data.dest ||
-      parameters.data.action
+      parameters.data.action ||
+      parameters.data.isTooltipOnly
```
timvandermeij, line 301: Does this actually work as expected? `isTooltipOnly` is only available in button widget annotations, but this is the link annotation class, which is not a button widget annotation?

src/core/annotation.js
```diff
@@ -1596,11 +1597,13 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
   }
 
   _processPushButton(params) {
-    if (!params.dict.has("A")) {
+    if (!params.dict.has("A") && !params.dict.has("TU")) {
       warn("Push buttons without action dictionaries are not supported");
       return;
     }
 
+    this.data.isTooltipOnly = !params.dict.has("A");
```
timvandermeij, line 1605: We cannot really test the display side of this patch, but we can unit test the core part, especially since we don't have unit tests for push buttons yet (we do have them for the other types of button widget annotations). Let's add two unit tests below https://github.com/mozilla/pdf.js/blob/master/test/unit/annotation_spec.js#L2319:

`it("should handle push buttons"` -> simple unit test to test rendering a regular push button, so without `TU` and with `A`
`it("should handle push buttons that act as a tooltip only"` -> same, but with `TU` and without `A`.

src/display/annotation_layer.js
calixteman: PushButton isn't a widget but a Link so it's why...
I slightly modify the code to put the isTooltipOnly in PushButton

src/display/annotation_layer.js
```diff
@@ -289,13 +289,15 @@ class AnnotationElement {
...
 class LinkAnnotationElement extends AnnotationElement {
-  constructor(parameters) {
-    const isRenderable = !!(
+  constructor(parameters, isRenderable = false) {
+    const _isRenderable = !!(
       parameters.data.url ||
       parameters.data.dest ||
-      parameters.data.action
+      parameters.data.action ||
+      isRenderable
```
Snuffleupagus, line 297: Why is this necessary though; couldn't you simply add `parameters.data.isTooltipOnly` instead here, since that will only (possibly) be defined for the "PushButton pretending to be LinkAnnotation"-case as far as I can tell?

This way, you'd reduce the amount of constructor-related changes in this file.

calixteman: I did that in the first version but it's quite confusing to have in LinkAnnotation (display) something defined in a WidgetAnnotation (core) so I moved the stuff in PushButton for this reason.
Anyway I'm fine to change it back... just tell me what you prefer.

test/unit/annotation_spec.js
```diff
@@ -2340,6 +2340,49 @@ describe("annotation", function () {
         })
         .catch(done.fail);
     });
+
+    it("should handle push buttons", function (done) {
+      const buttonWidgetRef = Ref.get(124, 0);
+      buttonWidgetDict.set("Ff", 0x10000);
```
timvandermeij, line 2346: (Here and below) Note that we have constant for this, namely `AnnotationFieldFlag.PUSHBUTTON`, that we can use here instead of the magic number.

src/display/annotation_layer.js
brendandahl: @Snuffleupagus thoughts?

Snuffleupagus: > @Snuffleupagus thoughts?

Sorry, this PR slipped my mind!

With the variable re-name reverted at https://github.com/mozilla/pdf.js/pull/12333/files#diff-1097e3b131ba22f8e8a6c2f20e3ab5ecR293, I honestly think that this is fine as-is :-)

Snuffleupagus: > @Snuffleupagus thoughts?

Sorry, this PR slipped my mind!

With the variable name changed back to `isRenderable` at https://github.com/mozilla/pdf.js/pull/12333/files#diff-1097e3b131ba22f8e8a6c2f20e3ab5ecR293, I honestly think that this is fine as-is :-)

src/display/annotation_layer.js
```diff
@@ -290,12 +290,14 @@ class AnnotationElement {
 
 class LinkAnnotationElement extends AnnotationElement {
   constructor(parameters) {
-    const isRenderable = !!(
+    const _isRenderable = !!(
```
timvandermeij, line 293: Why this change? It's a local variable, so it doesn't have to be marked with an underscore since it's private to this method anyway.

src/display/annotation_layer.js
```diff
@@ -290,12 +290,14 @@ class AnnotationElement {
...
       parameters.data.url ||
       parameters.data.dest ||
-      parameters.data.action
+      parameters.data.action ||
+      parameters.data.isTooltipOnly
     );
-    super(parameters, isRenderable);
+
```
timvandermeij, line 299: Same here, we can revert the changes in this line and the next one to get a smaller diff.

timvandermeij (commented):
Looks good with the two comments addressed.

timvandermeij (approved):

timvandermeij:
Thank you!

PR #12342, opened 2020-09-08
Move display/xml_parser.js in shared to use it in worker

calixteman:


timvandermeij:
Closing since this is done in #12344.

Issue #12319, opened 2020-09-03
Saving specific PDF that contains input data will not be displayed in Adobe Reader in other 3rd party apps


bogdanmaris:
**Affected versions**
* Firefox 81.0b5
* Latest Nightly 82.0a1

**Affected platforms**
* Ubuntu 18.04 64bit
* Windows 10 64
* macOS 10.15

**Steps to reproduce**
1. Open the following pdf https://www.irs.gov/pub/irs-pdf/f1040.pdf
2. Fill some data
3. Click the download button to save the file locally
4. Open the saved PDF in: 
Windows: Adobe Reader/Foxit
macOS: Adobe Reader
Ubuntu: Adobe Reader

**Expected result**
* Data saved in Firefox is displayed in Adobe Reader after saving locally.

**Actual result**
* Data saved in step 2 is not displayed once the PDF is opened in Adobe Reader.

**Regression range**
* Not a regression since this is reproducible since this feature was first enabled in Nightly.

**Additional Notes**
* This is not reproducible if the PDF is opened in another browser (Chrome, Safari, IE, Firefox, SumatraPDF).
* Same thing if saving the PDF inside chrome and opening in Adobe.

**Suggested severity**
* I my opinion this should have the S2 severity since data loss is involved.

calixteman:
This file contains XFA and Acroform and Acrobat (probably Foxit too) uses data from XFA so since we don't update them then nothing is displayed.
After investigations, getting "datasets" stream from XFA array, updating it (using the T value from Acroform to get field id) with the value and appending the new stream at end fixes the issue.
I'll do a patch in the next days.

timvandermeij:
Fixed by the PR above.

PR #12344, opened 2020-09-08
Save form data in XFA datasets when pdf is a mix of acroforms and xfa 

calixteman:
This patch aims to fix issue  #12319.


src/core/worker.js
```diff
@@ -545,6 +548,15 @@ class WorkerMessageHandler {
           return stream.bytes;
         }
 
+        acroForm = isDict(acroForm) ? acroForm : Dict.empty;
+        const xfa = acroForm.get("XFA") || [];
+        let xfaDatasets = null;
+        for (let i = 0, ii = xfa.length; i < ii; i += 2) {
+          if (xfa[i] === "datasets") {
```
timvandermeij, line 555: I don't think this always works. XFA can not only be an array, but also a stream; see https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=439&zoom=auto,-215,230 and and https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=473&zoom=auto,-215,214.

src/shared/xml_parser.js
```diff
@@ -326,14 +328,99 @@ class SimpleDOMNode {
   hasChildNodes() {
     return this.childNodes && this.childNodes.length > 0;
   }
+
+  searchNode(paths, pos) {
```
brendandahl, line 332: Can you add a comment for what this does?

src/core/writer.js
```diff
@@ -14,8 +14,14 @@
  */
 /* eslint no-var: error */
 
-import { bytesToString, escapeString } from "../shared/util.js";
+import {
+  bytesToString,
+  escapeString,
+  parseXFAPath,
```
timvandermeij, line 20: This only seems to be used in the core, so instead of putting it in `src/shared/util.js` it should be moved to `src/core/core_utils.js` (and its test).

timvandermeij:
@calixteman Now that this is merged, it would be great if you could make a follow-up PR that addresses the two comments above. Thanks!

src/core/worker.js
calixteman: If I understand correctly the specs, then it means that we won't have a reference for the datasets stream and so it'll be a way harder to update it without rewriting all the file...

src/core/writer.js
```diff
@@ -123,7 +129,55 @@ function computeMD5(filesize, xrefInfo) {
...
+  const data =
+    `${datasetsRef.num} ${datasetsRef.gen} obj\n` +
+    `<< /Type /EmbeddedFile /Length ${updatedXml.length}>>\nstream\n` +
+    updatedXml +
+    "\nendstream\nendobj\n";
+
+  newRefs.push({ ref: datasetsRef, data });
+}
+
+function incrementalUpdate(originalData, xrefInfo, newRefs, xref, datasetsRef) {
```
Snuffleupagus, line 178: Also, the number of parameters here are now large enough to warrant changing this function to take a parameter object instead (at least in my opinion), to aid readability of the call-sites:
```javascript
function incrementalUpdate({
  originalData,
  xrefInfo,
  newRefs,
  xref = null,
  datasetsRef = null,
}) {
```

PR #12345, opened 2020-09-08
Don't try to save something for a button which is neither a checkbox nor a radio

calixteman:


test/unit/annotation_spec.js
```diff
@@ -2317,6 +2317,28 @@ describe("annotation", function () {
...
+        pdfManagerMock,
+        idFactoryMock
+      )
+        .then(annotation => {
+          return annotation.save(partialEvaluator, task, {});
+        }, done.fail)
+        .then(data => {
+          expect(data).toEqual(null);
+          done();
+        }, done.fail);
```
Snuffleupagus, line 2340: Nit: Do you actually need to repeat the `done.fail` part multiple times here?
Unless I'm missing something obvious here, it ought to be possible to simply add one `catch` at the end to deal with failures:
```suggestion
        .then(annotation => {
          return annotation.save(partialEvaluator, task, {});
        })
        .then(data => {
          expect(data).toEqual(null);
          done();
        })
        .catch(done.fail);
```

timvandermeij:
Looks good; thanks!

PR #12352, opened 2020-09-09
Remove CSS variables feature-testing from `PDFSidebarResizer`

Snuffleupagus:
CSS variables are now supported in all reasonably modern browsers, according to:
 - https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#Browser_compatibility
 - https://caniuse.com/css-variables

### DIFF
--- a/web/pdf_sidebar_resizer.js
+++ b/web/pdf_sidebar_resizer.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import { clamp, NullL10n } from "./ui_utils.js";
+import { NullL10n } from "./ui_utils.js";
 
 const SIDEBAR_WIDTH_VAR = "--sidebar-width";
 const SIDEBAR_MIN_WIDTH = 200; // pixels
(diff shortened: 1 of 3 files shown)


web/pdf_sidebar_resizer.js
```diff
@@ -34,7 +34,7 @@ class PDFSidebarResizer {
    * @param {IL10n} l10n - Localization service.
    */
   constructor(options, eventBus, l10n = NullL10n) {
-    this.enabled = false;
+    this.enabled = true;
```
timvandermeij, line 37: I would say we can get rid of `this.enabled` altogether. It's always `true` now and we don't have similar toggles in other components.

Snuffleupagus: > I would say we can get rid of `this.enabled` altogether. It's always `true` now and we don't have similar toggles in other components.

That's a good point; it's gone now :-)

timvandermeij (approved):

timvandermeij:
Nice to get rid of this code. Thanks!

Issue #11626, opened 2020-02-24
Use #viewerContainer without position: absolute and overflow: auto


hrst:
I'm currently testing pdf.js and stuck with the components examples such as the [simpleviewer](https://github.com/mozilla/pdf.js/blob/master/examples/components/simpleviewer.html). These work great with the #viewerContainer div having position: absolute and overflow: auto. 

I understand this makes a lot of sense for a setup such as the default viewer, however, I'd like to display PDFs by adding it (basically) to the main body, i.e. utilizing the main browser window scrollbar. When I remove position: absolute and overflow: auto, the viewer works but loads all pages at once (I'm assuming because it cannot correctly detect which pages are visible to the user anymore). 

I also tried using the [pageviewer.js example](https://github.com/mozilla/pdf.js/blob/master/examples/components/pageviewer.js), which again works great but here I did not really find good examples on how to, after loading the first page, continue loading pages on demand when they are visible to the user (and unloading them afterwards), and so on, i.e. lots of functionality that is already perfectly implemented in the default viewer. I tried extracting the relevant parts (as I do not need 95% of the other stuff such as search, annotation, thumbnails, etc.), but this obviously got out of hand quickly and I felt as if was on the wrong path.

Is there any basic example on how to use pdf.js in a div without position: absolute and overflow: auto, and loading and unloading pages on demand as they are viewed by the user?

Any help is highly appreciated, I really tried to work with the examples and documentation but I'm stuck.

burtonator:
I'm having the same issue... I still can't get it to auto-render on scroll and it just renders all the pages.  even with overflow auto and position absolute.

burtonator:
When using the rendering queue it doesn't render anything... even with overflow auto and position: 0

burtonator:
If anyone has any suggestions where to start looking I'm going to try to fix this bug... It seems it only works with bot position AND overflow

ChrisHSandN:
After [getting caught out by this `position: absolute|relative` requirement](https://github.com/mozilla/pdf.js/issues/12351) I would like to add that if the library requires it (and will silently fail to work correctly if not set) then it becomes more than just a presentation attribute to be found in the CSS and feels like it should either:

1. Be automatically set by the pdf_viewer on initialisation
2. Be checked and an exception thrown if the `position` value is is not set to an acceptable value (possibly when you first try to scroll programmatically)

timvandermeij:
[...]

PR #12354, opened 2020-09-09
Ensure that the `container` div, on `BaseViewer`-instances, is absolutely positioned

Snuffleupagus:
The `getVisibleElements` helper function currently requires the viewerContainer to be absolutely positioned; possibly fixing this is tracked in issue #11626.

Without `position: absolute;` set, in the CSS, there's a number of things that won't work correctly such as e.g.
 - Determining which pages are currently visible, thus forcing all of them to render on load and increasing resource usage significantly; note https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#allthepages

 - Scrolling pages into view, by using the `BaseViewer.currentPageNumber` setter or similar.

Based on the number of opened issues over the years, the fact that `position: absolute;` is required has shown to be something that users can very easily overlook unless they follow e.g. the `simpleviewer` example to the letter.
Hence, to improve things until such a time that issue #11626 is fixed, we'll now refuse to initialize a `BaseViewer` instance unless the `container` has the required CSS set. (Forcibly setting `position: absolute;` on the viewerContainer element is bound to cause significantly more issues/confusion, hence the current approach of throwing an Error.)

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -145,14 +145,21 @@ class BaseViewer {
     this.viewer = options.viewer || options.container.firstElementChild;
 
     if (
-      (typeof PDFJSDev === "undefined" ||
-        PDFJSDev.test("!PRODUCTION || GENERIC")) &&
-      !(
-        this.container instanceof HTMLDivElement &&
-        this.viewer instanceof HTMLDivElement
-      )
+      typeof PDFJSDev === "undefined" ||
+      PDFJSDev.test("!PRODUCTION || GENERIC")
     ) {
-      throw new Error("Invalid `container` and/or `viewer` option.");
+      if (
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Given the confusion, this seems like a good idea to me too. Thanks!

kno10:
This broke my *working* code when updating from 2.4.456.
I only display one page "full screen" at a time using `SinglePageViewer` in a flex layout, and do not use scrolling at all.

Wouldn't a *warning* on the console have been sufficient?

I was able to fix this by nesting another position:absolute div in my existing container, but I dislike changes that break backwards compatibility when a warning would likely have been sufficient.

PR #12356, opened 2020-09-09
Set the modification date to the current day when saving

calixteman:


Snuffleupagus:
Perhaps slightly unrelated here, but isn't the default value in the `getModificationDate` function overly verbose?
Rather than https://github.com/mozilla/pdf.js/blob/e51e9d1f33bd6e404bfa795799f1cc035537091e/src/shared/util.js#L835 wouldn't just
```javascript
function getModificationDate(date = new Date()) {
```
suffice according to the "No parameters" case in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#Syntax

calixteman:
Fixed, thx

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

Issue #12355, opened 2020-09-09
Some checkboxes are not ticked inside 3 party apps if saved from Firefox


bogdanmaris:
**Affected versions**
* Latest Nightly 82.0a1 that has the fix from https://github.com/mozilla/pdf.js/issues/12319
(https://treeherder.mozilla.org#/jobs?repo=mozilla-central&revision=6cc2266faca2a2301c81f3442d6d893ad3fc9fdf)

**Affected platforms**
* Ubuntu 18.04 64bit
* MacOS 10.15
* Windows 10 64bit

**Steps to reproduce**
1. Open the following https://www.nta.go.jp/taxes/tetsuzuki/shinsei/annai/joyaku/annai/pdf2/250.pdf
2. Check the checkboxes from top right or scroll down to `Indirect Ownership` column inside the table from 6.
3. Click inside the checkboxes
4. Save the pdf by download button, save as or open with
5. Open the saved pdf in Foxit Reader.

**Expected result**
* Checkboxes are ticked when opening the PDF in Foxit.

**Actual result**
* Checkboxes are empty.

**Regression range**
* Not a regression since this did not work before as well.

**Additional Notes**
* I also saw one of the checkboxes missing from `Indirect Ownership` when opening in Adobe reader as well.

![image(1)](https://user-images.githubusercontent.com/69848942/92620347-3635cd80-f2cb-11ea-8282-8aa0a49ac035.png)
 
**Suggested severity**
* I my opinion this should have the S2 severity since this is data loss related, not sure how many pdfs are affected though.

calixteman:
It's because the names must be escaped when dumped (see 7.3.5)

timvandermeij:
Fixed by the PR above.

PR #12357, opened 2020-09-09
PDF names need to be escaped when saving

calixteman:
This patch aims to fix:
https://github.com/mozilla/pdf.js/issues/12355

### DIFF
--- a/src/core/core_utils.js
+++ b/src/core/core_utils.js
@@ -185,7 +185,33 @@ function parseXFAPath(path) {
   });
 }
 
+function escapePDFName(str) {
+  const buffer = [];
+  let start = 0;
+  for (let i = 0, ii = str.length; i < ii; i++) {
+    const char = str.charCodeAt(i);
+    if (char < 0x21 || char > 0x7e || char === 0x23) {
+      if (start < i) {
+        buffer.push(str.substring(start, i));
+      }
+      buffer.push(`#${char.toString(16)}`);
+      start = i + 1;
+    }
+  }
+
+  if (buffer.length === 0) {
+    return str;
+  }
+
+  if (start < str.length) {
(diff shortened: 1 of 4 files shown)


src/core/core_utils.js
```diff
@@ -165,7 +165,33 @@ function isWhiteSpace(ch) {
   return ch === 0x20 || ch === 0x09 || ch === 0x0d || ch === 0x0a;
 }
 
+function escapePDFName(str) {
```
brendandahl, line 168: There needs to be a special case for `#` to escape it to `#23`.

test/unit/core_utils_spec.js
```diff
@@ -211,4 +212,15 @@ describe("core_utils", function () {
       expect(isWhiteSpace(undefined)).toEqual(false);
     });
   });
+
+  describe("escapePDFName", function () {
+    it("should escape PDF name", function () {
```
brendandahl, line 217: Add test for `#`

brendandahl (changes requested):

src/core/writer.js
```diff
@@ -77,7 +78,7 @@ function numberToString(value) {
 
 function writeValue(value, buffer, transform) {
   if (isName(value)) {
-    buffer.push(`/${value.name}`);
+    buffer.push(`/${escapePDFName(value.name)}`);
```
timvandermeij, line 81: Could you also add a test in the writer spec so we know that this is in fact called correctly? I think changing https://github.com/mozilla/pdf.js/blob/master/test/unit/writer_spec.js#L68 to a value that needs escaping is already enough since `writeValue` is called by `writeDict` that is being tested there.

brendandahl (approved):

Issue #12359, opened 2020-09-10
First characters are not displayed when focus is lost from an overpopulated field


bogdanmaris:
**Affected versions**
* Latest Nightly 82.0a1
* Firefox 81.0b8

**Affected platforms**
* Ubuntu 18.04 64bit
* MacOS 10.15
* Windows 10 64bit

**Steps to reproduce**
1. Open the following http://foersom.com/net/HowTo/data/OoPdfFormExample.pdf
2. Click inside the Given Name field and overpopulate the field so the first characters are hidden
3. Click outside of the field

**Expected result**
* First characters are displayed at the beginning of the field when the focus is off that field.

**Actual result**
* Characters entered at the start of the field are hidden from view, although printing or saving the pdf will display the first characters as expected.

**Regression range**
* Not a regression since this did not work on a build where this feature was enabled in Nightly.

**Additional Notes**
* Other form editors don't suffer from this issue (Acrobat, Foxit, Chrome).
* Gif showing the issue:

![Focus jump](https://user-images.githubusercontent.com/69848942/92699666-8efe7800-f356-11ea-881f-b6d03d5297ca.gif)

**Suggested severity**
* I my opinion this should have the S4 severity, it is an inconvenience when editing the pdf, but it's not visible after printing, saving the pdf.

timvandermeij:
Fixed by the PR above.

cetinsert:
This has caused a regression for Safari currently being tracked here: https://github.com/mozilla/pdf.js/pull/13224#issuecomment-817613652

PR #12360, opened 2020-09-10
Reset cursor position when focus is out of text field

calixteman:
This patch aims to fix issue #12359.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12361, opened 2020-09-10
Ensure that all necessary /Font resources are included when saving a `WidgetAnnotation`-instance (issue 12294)

Snuffleupagus:
This patch contains a possible approach for fixing issue #12294, which compared to other PRs is purposely limited to the affected `WidgetAnnotation` code.

As mentioned elsewhere, considering that we're (at least for now) trying to fix *one specific* case, I think that we should avoid modifying the `Dict` primitive[1] and/or avoid a solution that (indirectly) modifies an existing `Dict`-instance[2].
This patch simply fixes the issue at hand, since that seems easiest for now, and I'd suggest that we worry about a more general approach if/when that actually becomes necessary.

Hence the solution implemented here, for `WidgetAnnotation`, is to simply use a combination of the local *and* AcroForm /DR resources during OperatorList-parsing to ensure that things work correctly regardless of where a particular /Font resource is found.
For saving of form-data, on the other hand, we want to avoid increasing the file-size unnecessarily and need to be smarter than just merging all of the available resources. To achive this, a new `WidgetAnnotation._getSaveFieldResources` method will when necessary produce a combined resources `Dict` with only the minimum amount of data from the AcroForm /DR resources included.

---
[1] You want to avoid anything that could cause the general `Dict` implementation to become slower, or more complex, just for handling an edge-case in my opinion.

[2] If an existing `Dict`-instance is modified unexpectedly, that could very easily lead to problems elsewhere since e.g. `Dict`-instances created during parsing are not expected to be changed.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -885,10 +885,18 @@ class WidgetAnnotation extends Annotation {
       "";
     const fieldType = getInheritableProperty({ dict, key: "FT" });
     data.fieldType = isName(fieldType) ? fieldType.name : null;
-    this.fieldResources =
-      getInheritableProperty({ dict, key: "DR" }) ||
-      params.acroForm.get("DR") ||
-      Dict.empty;
+
+    const localResources = getInheritableProperty({ dict, key: "DR" });
+    const acroFormResources = params.acroForm.get("DR");
+    this._fieldResources = {
+      localResources,
+      acroFormResources,
+      mergedResources: Dict.merge({
+        xref: params.xref,
+        dictArray: [localResources, acroFormResources],
+        mergeSubDicts: true,
+      }),
+    };
 
     data.fieldFlags = getInheritableProperty({ dict, key: "Ff" });
     if (!Number.isInteger(data.fieldFlags) || data.fieldFlags < 0) {
@@ -1043,7 +1051,7 @@ class WidgetAnnotation extends Annotation {
           .getOperatorList({
             stream,
             task,
-            resources: this.fieldResources,
+            resources: this._fieldResources.mergedResources,
             operatorList,
           })
           .then(function () {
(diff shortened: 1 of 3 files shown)


timvandermeij:
The functionality here not only works for saving, but also for printing, so could you add a reference test for the document from the issue? (Bonus points if it can be reduced, but if that's difficult a linked test case will definitely suffice.)

Moreover, I entered "test2" in the "Full name" field of the document from the issue and printed it to a file. The result is here: 
[out.pdf](https://github.com/mozilla/pdf.js/files/5211586/out.pdf)

Is this expected? If I copy/paste some of the characters from the PDF file itself, I get: 
[out2.pdf](https://github.com/mozilla/pdf.js/files/5211591/out2.pdf)

Since I'm not familiar with this language, I'm not really sure how to test this. Do you have any suggestions?

calixteman:
@timvandermeij, I tested this patch on top of https://github.com/mozilla/pdf.js/pull/12292 and it works well (print & save).
The patch in #12292 aims to fix the encoding issue you met.
Anyway r+ for me.

timvandermeij:
Alright, that's good to know. I think it would still be good to add a reference test in this PR to test the added logic, and it will give us confirmation that the other patch also works as expected since that _should_ improve the reference test added here.

Snuffleupagus:
> Bonus points if it can be reduced, but if that's difficult a linked test case will definitely suffice.

While that'd have been *technically* possible, it would have meant a whole lot of hand-editing of PDF data to make it work. All-in-all, it really didn't seem worth the time/effort; hence a regular linked test-case is included here.

Furthermore, I'm not completely sure if the `annotationStorage` data used here makes sense and/or if it needs to be encoded differently. Given that the encoding issues will be handled elsewhere, I hope that we can just leave this as-is for now.

timvandermeij (approved):

timvandermeij:
Nice work!

/botio makeref

PR #12363, opened 2020-09-11
canvas: Properly restore all the remaining items in stateStack in endDrawing

emilio:
We were correctly finishing the SMask group but not restoring all the extra transformations applied in stateStack, so if somebody ends up drawing to the same context after canceling mid-draw we'd get artifacts.

This fixes Mozilla bug 1664178[1].

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1664178

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -1009,8 +1009,8 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
 
     endDrawing: function CanvasGraphics_endDrawing() {
       // Finishing all opened operations such as SMask group painting.
-      if (this.current.activeSMask !== null) {
-        this.endSMaskGroup();
+      while (this.stateStack.length || this.current.activeSMask !== null) {
+        this.restore();
       }
 
       this.ctx.restore();


Snuffleupagus:
Unfortunately, as can be seen above, this causes a number of outright regressions in existing reference tests.

Please see https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing for information about running tests locally.

emilio:
Ah, thank you! I naively just ran `npm test` before submitting.

It seems we reset to default also to draw annotations, so that change is probably not quite correct... will look into the failures. Thanks @Snuffleupagus!

emilio:
(And I can try to add a test for this too :P)

emilio:
@Snuffleupagus can you run automation on the current PR? I'm running `gulp makeref`, but it's taking forever downloading PDFs so this may be faster :)

Snuffleupagus:
Please remove the unwanted commit properly, using something like `git rebase -i HEAD~3` and make the necessary edits, rather than adding a "revert" commit to the PR.

> can you run automation on the current PR?

Let's try once more; but if there's still issues you'll unfortunately have to bit the bullet and run test locally, since developing using the bots aren't exactly practical. 

> I'm running `gulp makeref`, but it's taking forever downloading PDFs so this may be faster :)

That's an unfortunate side-effect of not being able, for copyright reasons, to include all test PDF files in the repository. (Often times, creating reduced test-cases are unfortunately not that simple.)

Snuffleupagus:
@emilio If you're still having trouble downloading the PDF files for testing, send me an email (the address is in my Github profile) and I can provide all of them through a private Dropbox link.

emilio:
> Please remove the unwanted commit properly, using something like `git rebase -i HEAD~3` and make the necessary edits, rather than adding a "revert" commit to the PR.

Yup, I just didn't want to make destructive changes without knowing if that commit was the cuplrit, done now :)

> > can you run automation on the current PR?
> 
> Let's try once more; but if there's still issues you'll unfortunately have to bite the bullet and run tests locally, since developing using the bots aren't exactly practical.

Yeah, I understand, thanks!

> > I'm running `gulp makeref`, but it's taking forever downloading PDFs so this may be faster :)
> 
> That's an unfortunate side-effect of not being able, for copyright reasons, to include all test PDF files in the repository. (Often times, creating reduced test-cases are unfortunately not that simple.)

I see, that is indeed unfortunate... Anyhow I left my laptop downloading yesterday and now I have all the test files locally :)

The image differences above all seem like minor subpixel differences, unless I'm missing something? Which are not totally unexpected, I guess, as they are from tests that use this cancellation machinery which was leaving some existing state before this patch.

Do you know how should I update the references? Or I guess that happens automatically after merging the change anyways...

timvandermeij:
I have checked the unit test and reference test failures. You're right that your patch didn't cause any regressions anymore and these unit/reference test failures are unfortunately intermittent and being dealt with in #12123, so for this patch it looks good.

Updating references is done by our bot on command when we merge the PR, so you don't have to do anything for that.

timvandermeij (approved):

timvandermeij:
Looks good. Thanks!

/botio makeref

timvandermeij:
Hm, not sure why this didn't fail in the test run, but https://github.com/mozilla/pdf.js/blob/master/test/pdfs/bug852992_reduced.pdf now shows `Uncaught (in promise) TypeError: ctx is undefined` in the console :(

emilio:
Thanks, looking.

PR #12364, opened 2020-09-11
Dict keys need to be escaped too when saving

calixteman:


calixteman:
@timvandermeij,  @Snuffleupagus could you merge it asap please ?
We want to uplift it in beta.
Thx a lot

PR #12369, opened 2020-09-12
canvas: fix restore() with existing SMask groups and re-land #12363.

### LINKED ISSUE #12367: Fix or revert PR #12363
See https://github.com/mozilla/pdf.js/pull/12363#issuecomment-691488345 for context. If we can easily fix this I'd prefer that, but otherwise we can revert it for now.

/cc @emilio

emilio:
This fixes #12367. When we end the SMask group and stateStack.length is zero,
nothing updates this.current to reflect it.

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -1192,7 +1192,7 @@ var CanvasGraphics = (function CanvasGraphicsClosure() {
       groupCtx.clearRect(0, 0, groupCtx.canvas.width, groupCtx.canvas.height);
       groupCtx.restore();
     },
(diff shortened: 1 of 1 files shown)


emilio:
Seems like the same kind of subpixel differences as in #12369.

timvandermeij (approved):

timvandermeij:
Thank you! I have verified that the original issue that caused the backout is fixed now, and the code looks better like this.

/botio makeref

emilio:
Np, thanks a lot for reviewing!

PR #12370, opened 2020-09-12
Implement resetting of created streams for annotations

### LINKED ISSUE #12337: Highlight disappears when page is re-rendered after being out of view
Steps to reproduce:

1. Go to https://mozilla.github.io/pdf.js/web/viewer.html.
2. Download https://drive.google.com/uc?export=download&id=0B6YhVF9p-y2AU0RlYjRKakFQRjQ and open it with the viewer.
3. Notice that the highlight (without appearance stream) is rendered correctly on the first page.
4. Scroll to the last page.
5. Scroll to the first page.
6. Notice that the highlight is gone.

I don't directly have an idea how this can happen. As far as I know, the complete page is re-rendered after being out of view, so we _should_ also re-render the annotations, but apparently we don't. This is not intermittent and reproduces 100% of the time.

/cc @calixteman Perhaps you have any ideas why the second time the appearance stream is not generated/rendered?

timvandermeij:
Fixes #12337.

Note that I tried to use a non-caching getter here, but I did not manage to get that to work. I found that if it were to work it would require more code changes and I found it not that great that we would not cache the appearance stream on the instance because it would lead to recalculation every time the annotation goes out of view. Therefore, after checking the alternatives, this solution seemed the simplest.


src/core/annotation.js
```diff
@@ -563,6 +566,21 @@ class Annotation {
...
+  /**
+   * Reset the annotation.
+   *
+   * This involves resetting the various streams that are either cached on the
+   * annotation instance or created during its construction.
+   *
+   * @public
+   * @memberof Annotation
+   */
+  reset() {
```
Snuffleupagus, line 579: One thing that slightly worries me here, with respect to any future code changes, is the need to remember to *manually* add these streams in order for things to work correctly.

To reduce the risk of future bugs, can we at least ensure that `this.appearance` will always be reset here? Something like the following may not be perfect, but should at least help prevent trivial bugs during development:
```suggestion
  reset() {
    if (
      (typeof PDFJSDev === "undefined" ||
        PDFJSDev.test("!PRODUCTION || TESTING")) &&
      this.appearance && !this._streams.includes(this.appearance)
    ) {
      unreachable(/* Error message here... */);
    }
```

timvandermeij: I think that's a valid concern and we might be able to improve this by unconditionally including `this.appearance` so we don't need to do it at the various call sites.

timvandermeij: I pushed a new version that centralizes the `this.appearance` pushing in the `Annotation` constructor as much as possible instead of in various places. Would you still like to have this `PDFJSDev` addition or does the current solution suffice?

Snuffleupagus: > I pushed a new version that centralizes the `this.appearance` pushing in the `Annotation` constructor as much as possible instead of in various places.

While this obviously *reduces* the risk of future issues, as far as I can tell it doesn't completely eliminate it.

> Would you still like to have this `PDFJSDev` addition or does the current solution suffice?

Yes, please add those checks as well here!
They'll *only* apply in non-PRODUCTION or TESTING builds, i.e. won't affect the run-time in releases, and seem like an easy way to prevent future bugs.

Snuffleupagus:
Would it be possible to add a `fbf`-test for this, since unless I'm completely misunderstanding how those work they should currently trigger the bug!?

src/core/annotation.js
timvandermeij: Done in the new commit, including the FBF test which failed on master but passes with this patch applied.

test/test_manifest.json
```diff
@@ -4702,5 +4702,11 @@
       "md5": "f84ce8b7414f6a18e75a6ce69c902501",
       "rounds": 1,
       "type": "eq"
+    },
+    {  "id": "issue12337",
+       "file": "pdfs/issue12337.pdf",
+       "md5": "9165772d5b860bcbcc2478f32e311eb0",
+       "rounds": 2,
+       "type": "fbf"
```
Snuffleupagus, line 4710: Do you actually need to test all pages, since I'd imagine that using `lastPage: 1` should work!?

Given that `page.cleanup` is always run between pages, note [the `completeRender` function](https://github.com/mozilla/pdf.js/blob/741ce4f7fc451473c45b7a42ccc4138c94234e14/test/driver.js#L625-L645), that *should* be enough to trigger the bug since it's indirectly what's happening when the page is evicted from the `PDFPageViewBuffer` in the viewer.

Snuffleupagus (approved):
Looks good to me, with one final comment/question; thank you!

test/test_manifest.json
timvandermeij: You're right, that works!

Issue #12371, opened 2020-09-12
Investigate forward-back-forward test failures


timvandermeij:
In https://github.com/mozilla/pdf.js/commit/6ab3ba95193e7ff287b7d982cc011742603c3c61 we already disabled the `fbf` tests for Chrome, but all of a sudden they also fail for Firefox now in https://github.com/mozilla/pdf.js/pull/12368#issuecomment-691524124. I'd prefer we fix this, but otherwise we'll have to disable them for Firefox too to make `makeref` pass again. This issue tracks the problem for both browsers.

baloone:
Maybe we could allow getting a small differences between the first and the second round ?

timvandermeij:
Yes, perhaps #12165 might address this too.

Snuffleupagus:
Should we perhaps track this in issue #13061 instead now, since the `fbf` failures are also intermittent?

timvandermeij:
Yes, I have updated that issue so we can close this one.

PR #12372, opened 2020-09-13
Skip failing FBF tests, when running `makeref`, in Firefox as well

Snuffleupagus:
This will allow `makeref` to run "successfully" on the bots, since in the current state testing/makeref is just overall broken.
Obviously we still need to figure what's causing the intermittent failures, and fix them, but let's at least unblock things for now; see issue #12371.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12375, opened 2020-09-15
Ensure that the empty dictionary won't be accidentally modified, and slightly improve the "SaveDocument" handler in `src/core/worker.js`

Snuffleupagus:
- Ensure that the empty dictionary won't be accidentally modified
   
   Currently there's nothing that prevents modification of the `Dict.empty` primitive, which obviously needs to be *truly* empty to prevent any future (hard to find) bugs.

 - A couple of small improvements in the "SaveDocument" handler in `src/core/worker.js`

   - Check that the "Info"-entry, in the XRef-trailer, is actually a dictionary before accessing it. This is similar to the `PDFDocument.documentInfo` method and follows the general principal of validating data carefully before accessing it, given how often PDF-software may create corrupt PDF files.

   - Slightly simplify the "XFA"-lookup, since there's no point in trying to fetch something from the empty dictionary.

### DIFF
--- a/src/core/primitives.js
+++ b/src/core/primitives.js
@@ -171,7 +171,14 @@ var Dict = (function DictClosure() {
     },
   };
 
-  Dict.empty = new Dict(null);
+  Dict.empty = (function () {
+    const emptyDict = new Dict(null);
+
+    emptyDict.set = (key, value) => {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #12377, opened 2020-09-15
Set parent of radio annotation even if there is no 'V' field

### LINKED ISSUE #12376: Status of radio buttons not displayed in Firefox after saving the pdf file from Firefox
**Affected versions**
* Latest Nightly 82.0a1
* Firefox 81.0b9

**Affected platforms**
* Ubuntu 18.04 64bit
* MacOS 10.15
* Windows 10 64bit

**Steps to reproduce**
1. Open the following https://royalegroupnyc.com/wp-content/uploads/seating_areas/sample_pdf.pdf
2. Click on a radio button (remember it)
3. Save the .pdf locally
4. Open the .pdf inside Firefox

**Expected result**
* Radio buttons previously saved are displayed in Firefox.

**Actual result**
* Radio buttons that were previously saved are not displayed in Firefox.

**Regression range**
* Not a regression since this did not work before as well.
 
**Additional notes**
* The radio button is displayed in Firefox if pdfjs.renderInteractiveForms is set as false.

**Suggested severity**
* Not sure about the severity here, I would say S2-S3, I would say though that it would be bad if the data saved in Firefox won't be displayed in Firefox but all other pdf editors.

calixteman:
It aims to fix: #12376.
I added a test which fails without the patch.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -1741,10 +1741,10 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     // The parent field's `V` entry holds a `Name` object with the appearance
     // state of whichever child field is currently in the "on" state.
     const fieldParent = params.dict.get("Parent");
-    if (isDict(fieldParent) && fieldParent.has("V")) {
+    if (isDict(fieldParent)) {
+      this.parent = params.dict.getRaw("Parent");
       const fieldParentValue = fieldParent.get("V");
       if (isName(fieldParentValue)) {
-        this.parent = params.dict.getRaw("Parent");
         this.data.fieldValue = this._decodeFormValue(fieldParentValue);
       }
     }
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -2318,6 +2318,68 @@ describe("annotation", function () {
         }, done.fail);
     });
 
+    it("should save radio buttons without a field value", function (done) {
+      const appearanceStatesDict = new Dict();
+      const normalAppearanceDict = new Dict();
+
+      normalAppearanceDict.set("Checked", Ref.get(314, 0));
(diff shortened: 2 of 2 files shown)


src/core/annotation.js
```diff
@@ -1741,11 +1741,13 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
-        this.parent = params.dict.getRaw("Parent");
-        this.data.fieldValue = this._decodeFormValue(fieldParentValue);
+    if (isDict(fieldParent)) {
+      this.parent = params.dict.getRaw("Parent");
+      if (fieldParent.has("V")) {
+        const fieldParentValue = fieldParent.get("V");
+        if (isName(fieldParentValue)) {
+          this.data.fieldValue = this._decodeFormValue(fieldParentValue);
+        }
       }
```
Snuffleupagus, line 1751: Can't this be simplified a little bit, since having a separate `fieldParent.has("V")` check seems redundant now?
```suggestion
      const fieldParentValue = fieldParent.get("V");
      if (isName(fieldParentValue)) {
        this.data.fieldValue = this._decodeFormValue(fieldParentValue);
      }
```

test/unit/annotation_spec.js
```diff
@@ -2318,6 +2318,67 @@ describe("annotation", function () {
...
+        idFactoryMock
+      )
+        .then(annotation => {
+          const annotationStorage = {};
+          annotationStorage[annotation.data.id] = true;
+          return Promise.all([
+            annotation,
+            annotation.save(partialEvaluator, task, annotationStorage),
+          ]);
+        }, done.fail)
```
Snuffleupagus, line 2362: Rather than adding `done.fail` in multiple places, can't you simply add `.catch(done.fail);` *once* at the end of the promise-chain instead?

PR #12378, opened 2020-09-15
Fix issue with src="about:blank" on chrome v >= 76

romicaiarca:
Fix issue with src="about:blank" on chrome version >= 76


timvandermeij:
The commit message is not descriptive enough unfortunately. Which issue is being fixed here? Why is this necessary?

/cc @Rob--W since this involves the Chrome extension

Rob--W (changes requested):
Could you explain why you feel that this change is necessary?

timvandermeij:
Closing since it's not at all clear what the problem is that this fixes and why the fix works. We can reopen it once more clarity is provided.

PR #12388, opened 2020-09-17
Refactor the container/viewer checks in the BaseViewer constructor

### LINKED ISSUE #12385: Viewer Constructor blocks remote rendering

Configuration:
- Web browser and its version: all
- Operating system and its version: all
- PDF.js version: 2.6.347
- Is a browser extension: no

Construct the viewer and give it a div hosted in another frame... when it was working, it now throws 
```
Invalid `container` and/or `viewer` option.
```

This broke it: https://github.com/mozilla/pdf.js/commit/89f6bb0fc6b76009bff922ac5d5c3c4ddbc21ae2

Allow constructing a viewer that renders to an alternate window.

jsg2021:
Fixes #12385

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -150,8 +150,10 @@ class BaseViewer {
     ) {
       if (
         !(
-          this.container instanceof HTMLDivElement &&
-          this.viewer instanceof HTMLDivElement
+          this.container &&
+          this.container.tagName.toUpperCase() === "DIV" &&
+          this.viewer &&
+          this.viewer.tagName.toUpperCase() === "DIV"
         )
       ) {
         throw new Error("Invalid `container` and/or `viewer` option.");


Snuffleupagus (changes requested):
The solution implemented here feels way too complex, considering that the purpose of these checks were simply to catch outright user error. Can this be implemented simpler, e.g. by just checking the `tagName` instead?

Furthermore, note that the existing checks are only relevant for *one* particular build and consequently we do *not* want to add this code in e.g. `MOZCENTRAL` builds.

jeffri123 (approved):

web/base_viewer.js
```diff
@@ -149,10 +149,10 @@ class BaseViewer {
...
     ) {
       if (
-        !(
-          this.container instanceof HTMLDivElement &&
-          this.viewer instanceof HTMLDivElement
-        )
+        !this.container ||
+        !/div/i.test(this.container.tagName) ||
+        !this.viewer ||
+        !/div/i.test(this.viewer.tagName)
```
Snuffleupagus, line 155: No regular expressions, please!

```suggestion
        !(
          this.container && this.container.tagName.toUpperCase() === "DIV") &&
          this.viewer && this.viewer.tagName.toUpperCase() === "DIV")
         )
```

Snuffleupagus (changes requested):
Please improve the commit message, since it's currently a bit difficult to tell what the patch does without looking at the code and the linked issue.

For example, something like "Refactor the `container`/`viewer` checks in the `BaseViewer` constructor" might work. Furthermore, please *also* explain in the commit message *why* these changes are necessary; by mentioning the iframe use-case.

jsg2021:
Sure thing! :)

web/base_viewer.js
```diff
@@ -150,9 +150,9 @@ class BaseViewer {
     ) {
       if (
         !(
-          this.container instanceof HTMLDivElement &&
-          this.viewer instanceof HTMLDivElement
-        )
+          this.container && this.container.tagName.toUpperCase() === "DIV") &&
```
timvandermeij, line 153: The Travis CI build fails because parentheses are missing at the start of this line and the next line. Moreover, there is extra space on line 155. Please fix that and check if the final patch runs correctly.

jsg2021: Doh! Assumed a suggested change was safe... should have checked it locally.

timvandermeij (approved):

timvandermeij:
Looks good to me. Thanks!

PR #12391, opened 2020-09-18
Split only on the first `=` found in a key-value segment when parsing the parameters passed to the web viewer.

### LINKED ISSUE #12390: Parameters passed to web viewer parsed incorrectly when more than one `=` is present in a single key-value segment
Take the following URL for example:
`https://mozilla.github.io/pdf.js/web/viewer.html?file=https://cors-anywhere.herokuapp.com/https://drive.google.com/uc?id=1kTywHvNtS2I43rvhDZDaCSZw6Poz7MBX#page=1`

The URL points to a dummy PDF that is a copy of the one found on [w3c's website](https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf). Opening it in Google Chrome 85.0.4183.102 (Official Build) (64-bit) results in:
image

As you can see, the error (400 Bad Request) shown in the console is due to the file URL being truncated at the second `=`.

The relevant code that parse the key-value pairs is found here:

https://github.com/mozilla/pdf.js/blob/26ae7ba2ad97b50fedfbd714d703c3d0a63c53a3/web/ui_utils.js#L211

Changing it to split on `/=(.*)/` resolves the issue.

```
const param = parts[i].split(/=(.*)/);
```

zhangxiubo:
Fixes #12390.

### DIFF
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -208,7 +208,7 @@ function parseQueryString(query) {
   const parts = query.split("&");
   const params = Object.create(null);
   for (let i = 0, ii = parts.length; i < ii; ++i) {
-    const param = parts[i].split("=");
+    const param = parts[i].split(/=(.*)/);
     const key = param[0].toLowerCase();
     const value = param.length > 1 ? param[1] : null;
     params[decodeURIComponent(key)] = decodeURIComponent(value);


Snuffleupagus (changes requested):
See https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#file, which contains the following sentence:
> In the latter case, the PDF path/URL must be encoded using `encodeURIComponent()`.

Hence this is *not* a change that we can/will accept, and the linked issue is thus invalid.

PR #12393, opened 2020-09-19
Prevent errors if the `InkList` property, in InkAnnotations, is missing and/or not an Array (issue 12392)

Snuffleupagus:
To prevent a future bug, the `Vertices` property in PolylineAnnotations are handled the same way.

Fixes #12392

Issue #12392, opened 2020-09-19
pdfjs-2.5.207-es5-dis can not display INK in pdf file。


starryfunj:
Attach (recommended) or 
[product1.pdf](https://github.com/mozilla/pdf.js/files/5249872/product1.pdf)
Link to PDF file here:

Configuration:
-  chrome 80.0.3987.149
- windows 10 professional edition:
- PDF.js version:pdfjs-2.5.207-es5-dist,pdfjs-1.8.188-dist
- Is a browser extension: no

Steps to reproduce the problem:
1. use pdfjs-1.8.188-dist ,chrome can display Ink in the first page  in product1.pdf.
2. use pdfjs-2.5.207-es5-dist ,chrome can not display Ink in the first page  in product1.pdf.

What is the expected behavior? (add screenshot)
![image](https://user-images.githubusercontent.com/71555076/93668018-7b70b100-fabc-11ea-873f-1850b4c4a641.png)

What went wrong? (add screenshot)
![image](https://user-images.githubusercontent.com/71555076/93668001-5a0fc500-fabc-11ea-8d0b-40efc8728811.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

timvandermeij:
The pull request above will fix this, but do note that the PDF generator you're using (iText 7.0.3) is generating _invalid_ PDF files since the `InkList` property for ink annotations is required and must be an array according to the PDF specification; see https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=413&zoom=auto,-215,569 section 12.5.6.13. Consider filing a bug report for iText too.

PR #12394, opened 2020-09-19
In a text widget, Font resources can be in the appearance

### LINKED ISSUE #12398: Font resources for a text widget can be in appearance
In file https://github.com/mozilla/pdf.js/files/5249872/product1.pdf, the fifth text widget has a default appearance using font F1 and this one is neither in DR nor in AcroForm but in AP::N::Resources.

calixteman:
In file https://github.com/mozilla/pdf.js/files/5249872/product1.pdf, the fifth text widget has a default appearance using font F1 and this one is neither in DR nor in AcroForm but in AP::N::Resources.

Fixes #12398.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -998,12 +998,16 @@ class WidgetAnnotation extends Annotation {
 
     const localResources = getInheritableProperty({ dict, key: "DR" });
     const acroFormResources = params.acroForm.get("DR");
+    const appearanceResources =
+      this.appearance && this.appearance.dict.get("Resources");
+
     this._fieldResources = {
       localResources,
       acroFormResources,
+      appearanceResources,
       mergedResources: Dict.merge({
         xref: params.xref,
-        dictArray: [localResources, acroFormResources],
+        dictArray: [localResources, appearanceResources, acroFormResources],
         mergeSubDicts: true,
       }),
     };
@@ -1450,17 +1454,25 @@ class WidgetAnnotation extends Annotation {
         "Expected `_getAppearance()` to have been called."
       );
     }
-    const { localResources, acroFormResources } = this._fieldResources;
+    const {
+      localResources,
+      acroFormResources,
+      appearanceResources,
+    } = this._fieldResources;
 
     if (!this._fontName) {
       return localResources || Dict.empty;
     }
-    if (localResources instanceof Dict) {
-      const localFont = localResources.get("Font");
-      if (localFont instanceof Dict && localFont.has(this._fontName)) {
-        return localResources;
+
+    for (const resources of [localResources, appearanceResources]) {
+      if (resources instanceof Dict) {
+        const localFont = resources.get("Font");
+        if (localFont instanceof Dict && localFont.has(this._fontName)) {
+          return resources;
+        }
       }
     }
+
     if (acroFormResources instanceof Dict) {
       const acroFormFont = acroFormResources.get("Font");
       if (acroFormFont instanceof Dict && acroFormFont.has(this._fontName)) {
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -4410,6 +4410,18 @@
        "lastPage": 1,
        "type": "eq"
     },
+    {  "id": "issue12394",
+       "file": "pdfs/issue12392.pdf",
+       "md5": "76c3a34c6520940c45c66c92f7df2de5",
+       "rounds": 1,
+       "link": true,
+       "lastPage": 1,
+       "type": "eq",
+       "print": "true",
+       "annotationStorage": {
+         "35R": "收文"
+       }
+    },
     {  "id": "issue4883",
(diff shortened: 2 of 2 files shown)


timvandermeij:
Is there test coverage for this or should a test be added? The file itself is already included in https://github.com/mozilla/pdf.js/pull/12393/files, but do we need a print test here?

calixteman:
I can add a test.

timvandermeij:
Looks good to me with a reference test added. Most likely this can be done in the same way as in https://github.com/mozilla/pdf.js/pull/12361/files.

src/core/annotation.js
```diff
@@ -922,12 +922,17 @@ class WidgetAnnotation extends Annotation {
 
     const localResources = getInheritableProperty({ dict, key: "DR" });
     const acroFormResources = params.acroForm.get("DR");
+    const appearanceResources = this.appearance
+      ? this.appearance.dict.get("Resources")
+      : null;
+
```
Snuffleupagus, line 928: Do you actually need the ternary expression here?
```javascript
const appearanceResources =
  this.appearance && this.appearance.dict.get("Resources")`;
```
ought to suffice, given that all relevant code already contains `Dict` checks and having `appearanceResources` be `null` instead of `undefined` really shouldn't matter.

automatedbugreportingfacility: In the future, when all supported browsers handle optional chaining, this can be simplified further to:
```
const appearanceResources = this.appearance?.dict.get("Resources");
```

test/test_manifest.json
```diff
@@ -4243,6 +4243,18 @@
        "lastPage": 1,
        "type": "eq"
     },
+    {  "id": "issue12394",
+       "file": "pdfs/issue12392.pdf",
```
timvandermeij, line 4247: This file is missing in the test folder. You'll need something similar to https://github.com/mozilla/pdf.js/pull/12361/files#diff-eba194d158906ef5ebd771295e400e6a.

timvandermeij:
@calixteman If you have time, could you address the comments in this and the other PRs so we can get some of them merged?

test/test_manifest.json
calixteman: I'm reusing the test file for #12392:
https://github.com/mozilla/pdf.js/blob/master/test/pdfs/issue12392.pdf.link
So it should be ok, isn't it ? or need I to duplicate the pdf ?

Snuffleupagus: > So it should be ok, isn't it ?

Of course it's fine to re-use one PDF file for multiple tests, we already do that in other tests.

test/test_manifest.json
```diff
@@ -4243,6 +4243,18 @@
...
+       "md5": "76c3a34c6520940c45c66c92f7df2de5",
+       "rounds": 1,
+       "link": true,
+       "lastPage": 1,
+       "type": "eq",
+       "print": "true",
+       "annotationStorage": {
+           "35R": "收文"
```
Snuffleupagus, line 4255: Nit: A little bit too much indentation here.

src/core/annotation.js
```diff
@@ -922,12 +922,16 @@ class WidgetAnnotation extends Annotation {
...
+      this.appearance && this.appearance.dict.get("Resources");
+
     this._fieldResources = {
       localResources,
       acroFormResources,
+      appearanceResources,
       mergedResources: Dict.merge({
         xref: params.xref,
-        dictArray: [localResources, acroFormResources],
+        dictArray: [localResources, acroFormResources, appearanceResources],
```
Snuffleupagus, line 934: Just a small questions here: Do you want this ordering, since that'd prefer the `appearanceResources` *last* when merging the dictionaries, since `Dict.merge` iterates through this array left-to-right?
Or would swapping the order of `acroFormResources` and `appearanceResources` here make sense?

It probably doesn't matter, since this situation probably is an an edge-case anyway, but figured I'd ask anyway :-)

calixteman: Probably a edge case, but I think it's better to swap.

test/test_manifest.json
timvandermeij: I completely missed this since no file was added and the issue numbers looked similar at first sight. Reusing the file is definitely OK; sorry for the noise!

timvandermeij:
Opening the `issue12392.pdf` file in the preview above or locally with this code checked out, it doesn't print anything I enter in the fields anymore for me while this works with https://mozilla.github.io/pdf.js/web/viewer.html. Could you check why this happens?

calixteman:
In order to see the "effect" of this patch on file `issue12392.pdf`, it's required to have https://github.com/mozilla/pdf.js/pull/12292 before.

timvandermeij:
Could you rebase this one now that #12292 is merged so we can also look into landing this one?

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

PR #12395, opened 2020-09-19
Render not displayed annotations in using normal appearance when printing

calixteman:
A pdf with several pages containing some text or checked radio/checkbox not on page 1/2 were not printed.
Since they were not rendered in annotation_layer, then the annotationStorage has no entries for such widget.
So the goal of the patch is to fix that in using normalAppearance when the annotationStorage doesn't contain the widget id.


src/core/annotation.js
```diff
@@ -1097,7 +1097,14 @@ class WidgetAnnotation extends Annotation {
   }
 
   async save(evaluator, task, annotationStorage) {
-    if (this.data.fieldValue === annotationStorage[this.data.id]) {
+    const value = annotationStorage[this.data.id];
+    if (
+      this.data.fieldValue === value ||
+      value === null ||
```
timvandermeij, line 1103: I would expect the value to either be set (the default value or a changed value) or be `undefined` (not rendered). Therefore, why do have to handle `null` as well since I don't see code that sets it to `null`? (Applies to the changes below as well.)

test/unit/annotation_spec.js
```diff
@@ -2280,6 +2387,65 @@ describe("annotation", function () {
         }, done.fail);
     });
 
+    it("should render radio for printing using normal appearance", function (done) {
```
timvandermeij, line 2390: `should render radio buttons for printing using normal appearance` ("buttons" should be added).

src/core/annotation.js
```diff
@@ -1182,6 +1188,14 @@ class WidgetAnnotation extends Annotation {
       return null;
     }
     const value = annotationStorage[this.data.id];
+    if (value === null || value === undefined) {
+      // value === null when a choice widget
```
timvandermeij, line 1192: `choice widget` -> `widget` since this is the common widget class.

timvandermeij (commented):
Do you have a PDF file to test this with? Even though the unit tests cover the cases and adding a reference test is not really possible, it would be good for me to verify this in the viewer itself.

calixteman:
> Do you have a PDF file to test this with? Even though the unit tests cover the cases and adding a reference test is not really possible, it would be good for me to verify this in the viewer itself.

https://www.canada.ca/content/dam/ircc/migration/ircc/english/passport/forms/pdf/pptc155.pdf

Save something on page 3 and reopen the newly created file and print.

src/core/annotation.js
calixteman: https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js#L681
I was wondering if we should replace the null by a undefined... 
Wdyt ?

timvandermeij: I noticed that I changed that in https://github.com/mozilla/pdf.js/commit/36e149800e17ae72463307f3e56fd4723c2373e5. The `null` value is only there to indicate that no field value is present, so `undefined` should be fine here as well. That way the code is more consistent: the value is either undefined (no key in storage due to rendering not happening or key in storage with no value due to missing data) or any other value (default/changed valid value).

src/core/annotation.js
```diff
@@ -1097,7 +1097,14 @@ class WidgetAnnotation extends Annotation {
   }
 
   async save(evaluator, task, annotationStorage) {
-    if (this.data.fieldValue === annotationStorage[this.data.id]) {
+    const value = annotationStorage[this.data.id];
+    if (
+      this.data.fieldValue === value ||
```
Snuffleupagus, line 1102: Nit: For improved readability, please change this to `value === this.data.fieldValue ||` such that all of the comparisons are sorted consistently in one `if` condition.

src/core/annotation.js
```diff
@@ -1097,7 +1097,14 @@ class WidgetAnnotation extends Annotation {
...
+      value === null ||
+      value === undefined
+    ) {
+      // value === null when a choice widget
+      // hasn't been changed
```
Snuffleupagus, line 1107: Nit: It looks like this comment will fit on *one* line, and please remember to end comments with a period as well.

This applies to the comments below as well.

src/core/annotation.js
```diff
@@ -1182,6 +1188,14 @@ class WidgetAnnotation extends Annotation {
...
+      // value === null when a choice widget
+      // hasn't been changed.
+      // The annotation hasn't been rendered
+      // so use the appearance
+      return null;
+    }
+
     if (value === "") {
       return "";
     }
```
Snuffleupagus, line 1201: Wait, why is this pre-existing case special?

If it's indeed necessary to return an empty string here, a comment explaining why would seem like a very good idea!

Snuffleupagus:
> [...] and adding a reference test is not really possible,

Why is that though, since based on the description in https://github.com/mozilla/pdf.js/pull/12395#issue-489806229 I'd sort of expect that an `eq` test of even *a single* page ought to trigger this if all of the following holds:
 - Printing is tested, by specifying `"print": true,` in the manifest.
 - An *empty* `AnnotationStorage`-instance is provided, by specifying `"annotationStorage": {},` in the manifest.

timvandermeij:
> Why is that though

I thought so because of the "since they were not rendered in annotation_layer" part. If we add a reference test we have to render the page and I thought this bug would not show up then. I might be wrong here and I see that a test is now added, so we should verify if this indeed catches the bug.

timvandermeij:
This PR needs a rebase.

test/test_manifest.json
```diff
@@ -3659,6 +3659,16 @@
...
       "about": "CFF font that is drawn with clipping."
     },
+    {  "id": "prefilled_f1040",
+       "file": "pdfs/prefilled_f1040.pdf",
+       "md5": "6c45b2c43e365bb7c88b9cddca0a4865",
+       "rounds": 1,
+       "type": "eq",
+       "print": true,
+       "annotationStorage": {
+         "1605R": true
```
timvandermeij, line 3669: I verified that this patch itserlf fixes the reported issue, but I can't get this reference test working when checking out this patch and I also don't see how this is supposed to work. The prefilled file seems to only have prefilled content on page 1 and this page is always rendered, so there is an annotation layer and the content is simply in the annotation storage. Checkbox 1605R is also on the first page, so that will also work. I don't see any changes in the output before/after this patch, which confirms my suspicion that this most likely doesn't work as expected.

Did you try the suggestion from https://github.com/mozilla/pdf.js/pull/12395#issuecomment-695767556? You can run the reference tests locally before/after applying this patch to check if the test indeed works as expected.

src/core/annotation.js
```diff
@@ -1602,9 +1617,13 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
   }
 
   async _saveCheckbox(evaluator, task, annotationStorage) {
-    const defaultValue = this.data.fieldValue && this.data.fieldValue !== "Off";
     const value = annotationStorage[this.data.id];
+    if (value === undefined) {
+      // No change
```
timvandermeij, line 1622: `undefined` actually means "not rendered" because otherwise the value would be false/true depending on the checked state. Given that we also don't have this comment in `WidgetAnnotation.save`, I'd say we can just remove this line here and in `_saveRadioButton` below.

timvandermeij (commented):
Finally had some time to review this; sorry for the delay. Looks good with these final comments resolved, the most important one being the reference test that doesn't seem to work.

timvandermeij:
@calixteman Since form support is getting close to being enabled upstream, it would be great if we can get this PR merged. For the reference test, I recommend to run `gulp makeref` locally with this branch checked out, with all entries in `test/test_manifest.json` removed except for the one added in this patch so you can iterate quickly, and then looking at the reference image in the `test/ref` folder to see what I currently see and if a fix has effect.

calixteman:
> @calixteman Since form support is getting close to being enabled upstream, it would be great if we can get this PR merged. For the reference test, I recommend to run `gulp makeref` locally with this branch checked out, with all entries in `test/test_manifest.json` removed except for the one added in this patch so you can iterate quickly, and then looking at the reference image in the `test/ref` folder to see what I currently see and if a fix has effect.

It should be ok now, sorry for the delay.
I ran the test without the patch and nothing is printed on page 2 and it's ok with it.

timvandermeij:
`WARNING: MD5 of file "pdfs/prefilled_f1040.pdf" does not match file. Expected "6c45b2c43e365bb7c88b9cddca0a4865" computed "2335da66fb7c2c3b84971597f27785e2"`

Did the PDF file change? If so, the hash also needs to be changed in the test manifest.

calixteman:
> `WARNING: MD5 of file "pdfs/prefilled_f1040.pdf" does not match file. Expected "6c45b2c43e365bb7c88b9cddca0a4865" computed "2335da66fb7c2c3b84971597f27785e2"`
> 
> Did the PDF file change? If so, the hash also needs to be changed in the test manifest.

My bad, I fixed it, sorry.
Anyway the CI is failing but I guess that isn't my fault, else tell me.

timvandermeij:
Oh, that's indeed not the fault of this PR. We'll look into that.

timvandermeij:
Node.js 14 became LTS today, causing the CI builds to fail because the caches still contained packages for the previous LTS version. Purging the cache and retriggering the CI build fixed the issue.

timvandermeij (approved):

timvandermeij:
Thank you!

/botio-linux makeref

PR #12397, opened 2020-09-20
Update the `webpack-stream` dependency (issue 11996)

Snuffleupagus:
Given that the long-standing `webpack-stream` [issue 201](https://github.com/shama/webpack-stream/issues/201) was recently fixed in [PR 207](https://github.com/shama/webpack-stream/pull/207), and a new version released, we should now finally be able to update the dependency.
However, depending on if/when `webpack-stream` gets support for Webpack 5 (which is currently in beta) we may still want remove our `webpack-stream` dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16772,22 +16772,38 @@
       }
     },
     "webpack-stream": {
-      "version": "5.0.0",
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Fortunately we can update this now. Thanks!

PR #12405, opened 2020-09-23
Fixed keydown event handling problem with shadow DOM.

nickyc975:
Editable elements in shadow DOMs can not be detected in old version.

### DIFF
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -985,6 +985,28 @@ function moveToEndOfArray(arr, condition) {
   }
 }
 
+/**
+ * Get the active or focused element in current DOM.
+ *
+ * Recursively search for the truly active or focused element in case there are
+ * shadow DOMs.
+ *
+ * @returns {Element} the truly active or focused element.
+ */
+function getActiveOrFocusedElement() {
+  let curRoot = document;
+  let curActiveOrFocused =
+    curRoot.activeElement || curRoot.querySelector(":focus");
+
+  while (curActiveOrFocused && curActiveOrFocused.shadowRoot) {
+    curRoot = curActiveOrFocused.shadowRoot;
+    curActiveOrFocused =
+      curRoot.activeElement || curRoot.querySelector(":focus");
+  }
+
+  return curActiveOrFocused;
+}
+
 export {
   AutoPrintRegExp,
   CSS_UNITS,
@@ -1026,4 +1048,5 @@ export {
   WaitOnType,
   waitOnEventOrTimeout,
   moveToEndOfArray,
+  getActiveOrFocusedElement,
 };
--- a/web/app.js
+++ b/web/app.js
@@ -19,6 +19,7 @@ import {
   AutoPrintRegExp,
   DEFAULT_SCALE_VALUE,
   EventBus,
+  getActiveOrFocusedElement,
   getPDFFileNameFromURL,
   isValidRotation,
   isValidScrollMode,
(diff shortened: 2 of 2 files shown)


Snuffleupagus:
> Editable elements in shadow DOMs can not be detected in old version.

When does this actually happen, since as-is there's not really enough information provided to make it possible to understand in *exactly* what situations these changes are needed. Please provide additional information, and preferably a *runnable example* such that it's possible to actually test/review these changes!

Furthermore, given that the default viewer was originally written for usage in Firefox, it's somewhat difficult to even understand why these any changes are necessary here. For example, why can't these changes simply be included in a custom implementation of the viewer; please note http://mozilla.github.io/pdf.js/getting_started/#introduction (emphasis mine):

nickyc975:
Sorry for the information missing. 

I am developing a [translator extension](https://github.com/EdgeTranslate/EdgeTranslate) for multiple browsers and I used pdf.js as a built-in pdf viewer in my extension. When I am trying to injecting a shadow DOM with an input box into the pdf viewer, I found that I could not input space in the input box. When I clicked the space key, the keydown event was catched by pdf.js and resulted in page scrolling instead of inputting a space in the input box.

After digging into the code I found that although pdf.js had considered the input case, the original solution could not handle DOMs with shadow DOM element. I also found that the problem with text editors has been discussed in #7156 and #11253 provided the incomplete solution. So I thought that maybe you need a patch for that.

web/ui_utils.js
```diff
@@ -985,6 +985,32 @@ function moveToEndOfArray(arr, condition) {
   }
 }
 
+/**
+ * Get the active or focused element in current DOM.
+ *
+ * Recursively search for the truly active or focused element incase there are
```
timvandermeij, line 991: `incase` -> `in case`

web/ui_utils.js
```diff
@@ -985,6 +985,32 @@ function moveToEndOfArray(arr, condition) {
...
+ * shadow DOMs.
+ *
+ * @author Chen Jinlong <chenjinlong2016@outlook.com>
```
timvandermeij, line 994: We never use the `@author` tag in the codebase, so this should be removed.

web/ui_utils.js
```diff
@@ -985,6 +985,32 @@ function moveToEndOfArray(arr, condition) {
...
+ *
+ * @param {DocumentOrShadowRoot} root the root element.
```
timvandermeij, line 996: `@param {DocumentOrShadowRoot} [root] - The root element.`

`root` should be in brackets to indicate that it's an optional argument.

We can also leave this out since no code currently calls this function with an argument?

web/ui_utils.js
```diff
@@ -985,6 +985,32 @@ function moveToEndOfArray(arr, condition) {
...
+ *
```
timvandermeij, line 997: This newline can be removed.

timvandermeij (commented):
Is there an easy way to verify for us that this works? The code itself looks good with the comments addressed.

nickyc975:
Sorry for the late replying. I'll fix the comment problems and make a package for you to verify the patch.

nickyc975:
I made 2 packages for you to verify the patch. One is built from the fixed branch and the other is built from the original branch. Download the packages and extract them to somewhere in you computer respectively. Follow the following steps to run the packages.

Run original:

1. `cd` to the original folder and install gulp by `npm install gulp` (maybe other dependencies needed too);
2. Run `gulp server` to start the server;
3. Navigate to [http://localhost:8888/web/viewer.html](http://localhost:8888/web/viewer.html);
4. You'll see an extra editable element with `text content` as its content on top of the original tool bar;
5. Inputting space in the element will cause the page scrolling instead of puting space in the element.

Run fixed:

Steps 1-4 is the same (replace `original` with `fixed`).
The difference is that inputing space in the element just puts spaces into the content, not causes scrolling or other strange behavior.

Here are the packages:

fixed: [fixed.zip](https://github.com/mozilla/pdf.js/files/5272977/fixed.zip)

original: [original.zip](https://github.com/mozilla/pdf.js/files/5272979/original.zip)

timvandermeij (approved):

timvandermeij:
Seems to work just fine. Thanks!

Issue #12406, opened 2020-09-23
Links with bounds outside the page appear in the wrong place


nvinen:
I'm using the current beta version (pdfjs-2.6.347-dist).

I traced the bug to incorrect code for negating numbers by pasting a hyphen on the front of the string, rather than actually negating the numbers. Pasting a hyphen in front of a negative number like -3 gives you --3 instead of 3.

It's certainly possible for link (and other annotation) bounds can be negative (ie, beyond the top/left edge of the page). The current pdf.js code offsets such regions significantly from where they should be. The following patch fixes this bug and a related bug, and also adds a few lines to clip the regions so that they are within the page so they look neater.

--- pdf.js      (revision 1225)
+++ pdf.js      (working copy)
@@ -9374,14 +9374,18 @@
           page = this.page,
           viewport = this.viewport;
     const container = document.createElement("section");
-    let width = data.rect[2] - data.rect[0];
-    let height = data.rect[3] - data.rect[1];
     container.setAttribute("data-annotation-id", data.id);

     const rect = _util.Util.normalizeRect([data.rect[0], page.view[3] - data.rect[1] + page.view[1], data.rect[2], page.view[3] - data.rect[3] + page.view[1]]);
+    if( rect[0] < page.view[0] ) rect[0] = page.view[0]; else if( rect[0] > page.view[2] ) rect[0] = page.view[2];
+    if( rect[1] < page.view[1] ) rect[1] = page.view[1]; else if( rect[1] > page.view[3] ) rect[1] = page.view[3];
+    if( rect[2] < page.view[0] ) rect[2] = page.view[0]; else if( rect[2] > page.view[2] ) rect[2] = page.view[2];
+    if( rect[3] < page.view[1] ) rect[3] = page.view[1]; else if( rect[3] > page.view[3] ) rect[3] = page.view[3];
+    let width = rect[2] - rect[0];
+    let height = rect[3] - rect[1];

     container.style.transform = `matrix(${viewport.transform.join(",")})`;
-    container.style.transformOrigin = `-${rect[0]}px -${rect[1]}px`;
+    container.style.transformOrigin = `${-rect[0]}px ${-rect[1]}px`;

     if (!ignoreBorder && data.borderStyle.width > 0) {
       container.style.borderWidth = `${data.borderStyle.width}px`;
@@ -9801,7 +9805,7 @@
     });
     const parentLeft = parseFloat(parentElement.style.left);
[...]

PR #12409, opened 2020-09-24
Compute the `transformOrigin` correctly, for negative values, when rendering `AnnotationElement`s (bug 1627030)

Snuffleupagus:
This changes the `transformOrigin` calculations in `AnnotationElement._createContainer` and `PopupAnnotationElement.render`, to ensure that e.g. the clickable area of annotations and/or popups are both positioned correctly.

The problem occurs for *negative* values, since they're not negated correctly because of how the `transformOrigin` strings were build; see issue #12406 for a more in-depth explanation. Previously, for negative values, the `transformOrigin` strings would thus be ignored since they're not valid.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1627030

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -177,7 +177,7 @@ class AnnotationElement {
     ]);
 
     container.style.transform = `matrix(${viewport.transform.join(",")})`;
-    container.style.transformOrigin = `-${rect[0]}px -${rect[1]}px`;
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Interesting bug; thanks for fixing this!

/botio makeref

PR #12414, opened 2020-09-24
Need to reset the streams when printing

calixteman:
It aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1666753.


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

PR #12420, opened 2020-09-28
Improve webkit ui rendering

janbrummer:
Add -webkit-appearance: none;
to

.dropdownToolbarButton > select
and
.toolbarField.pageNumber

to fix frame rendering in webkit based browsers.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -918,6 +918,7 @@ html[dir="rtl"] .dropdownToolbarButton::after {
 }
 
 .dropdownToolbarButton > select {
+  -webkit-appearance: none;
   width: 162px;
   height: 28px;
   font-size: 12px;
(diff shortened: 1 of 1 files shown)


timvandermeij:
It's not clear what this actually fixes without screenshots before and after. Moreover, we use Autoprefixer for release builds, which should already take care of this since we don't want to insert all browser-specific prefixes ourselves...

Snuffleupagus:
> It's not clear what this actually fixes without screenshots before and after.

Testing this locally in Google Chrome 86 (the current beta version), the patch removes (part of) the intended `left-padding` for the zoom-dropdown which result in things looking inconsistent between browsers; as such I don't think that we can accept this patch:
 - With `master` master

 - With this patch patch

PR #12422, opened 2020-09-29
Don't trigger searching, in the `GENERIC` viewer, when <kbd>Shift</kbd> is used together with the regular keyboard shortcut (issue 12421)

### LINKED ISSUE #12421: Mac only: "Command F" vs "Command Shift F"

Configuration:
- Web browser and its version: Chrome 85.0.4183.121 (Official Build) (64-bit)
- Operating system and its version: Mac 10.15.4
- PDF.js version: 2.3.164
- Is a browser extension: Yes

1. Open a PDf
2. Press "Command + Shift + F"

Chrome typically uses this shortcut for "Always Show Toolbar in Full Screen".

In PDF.js, this shortcut is captured by the extension as "Find". The shortcut "Command + F" is also captured and serves this purpose, so I don't believe "Command + Shift + F" should ALSO be captured for "Find". There is no way to configure the keyboard shortcuts captured by this extension in its options.

I prefer PDF.js to the built-in Chrome PDF viewer (mostly for its ability to open PDF documents to the previous location), but this makes it hard to use.

Snuffleupagus:
Please note that this change won't affect the Firefox built-in PDF Viewer, since it uses the native browser findbar. However, it brings the `GENERIC` viewer inline with how searching is normally triggered in browsers.

Note that in Firefox searching is only triggered by <kbd>Ctrl</kbd>+<kbd>F</kbd> (or <kbd>command</kbd>+<kbd>F</kbd> on Mac), but <kbd>Shift</kbd> isn't used; see https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly#w_search

Furthermore, looking at the "Miscellaneous" section of https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly#firefox:mac:fx81 the <kbd>command</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd> shortcut is even reserved for the "Toggle Full Screen" functionality on Mac.

Fixes #12421

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2714,7 +2714,7 @@ function webViewerKeyDown(evt) {
     // either CTRL or META key with optional SHIFT.
     switch (evt.keyCode) {
       case 70: // f
-        if (!PDFViewerApplication.supportsIntegratedFind) {
+        if (!PDFViewerApplication.supportsIntegratedFind && !evt.shiftKey) {
           PDFViewerApplication.findBar.open();
           handled = true;
         }


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12423, opened 2020-09-29
Update `escodegen` to version 2.0 

Snuffleupagus:
- Update `escodegen` to version 2.0

   This release contains support for additional modern ECMAScript features, such as e.g. the nullish coalescing operator `??` and the optional chaining operator `?.`.
 
 - Add basic support for the optional chaining operator `?.`

   For now we need to use a Babel-plugin, since part of our build system doesn't support this fully (e.g. Babel-loader, Webpack 4.x, and SystemJS).

   While the `?.` operator will thus always be transpiled by Babel, even in modern builds, simply supporting it for development purposes seems like a step in the right direction.

   https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -224,6 +224,12 @@ function createWebpackConfig(defines, output) {
                   loose: true,
                 },
               ],
+              [
+                "@babel/plugin-proposal-optional-chaining",
+                {
+                  loose: true,
+                },
+              ],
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

PR #12426, opened 2020-09-30
Re-factor how printing is triggered in the default viewer

Snuffleupagus:
This adds a new `PDFViewerApplication.triggerPrinting` method, which takes care of checking that printing is actually supported before calling `window.print`, to remove the need to duplicate that code in multiple places.

Also, removes the `PDFViewerApplication.printing` getter since it's not really necessary any more.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -593,10 +593,6 @@ const PDFViewerApplication = {
     this.pdfViewer.currentPageNumber = val;
   },
 
-  get printing() {
-    return !!this.printService;
-  },
-
   get supportsPrinting() {
     return PDFPrintServiceFactory.instance.supportsPrinting;
   },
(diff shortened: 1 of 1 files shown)


web/app.js
```diff
@@ -1378,12 +1374,9 @@ const PDFViewerApplication = {
       }
     }
 
-    if (!this.supportsPrinting) {
-      return;
-    }
     if (triggerAutoPrint) {
-      setTimeout(function () {
-        window.print();
+      setTimeout(() => {
```
Snuffleupagus, line 1378: Note that I'm not actually sure if/why `setTimeout` is (still) necessary here!?

As far as I can tell, this was present already in PR #2839 which originally added this auto-print feature, hence there's probably a good reason for this.

timvandermeij (approved):

timvandermeij:

PR #12429, opened 2020-09-30
[api-minor] Add the possibility to collect Javascript actions

calixteman:

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -14,12 +14,14 @@
  */
 
 import {
+  AnnotationActionEventType,
   AnnotationBorderStyleType,
   AnnotationFieldFlag,
   AnnotationFlag,
   AnnotationReplyType,
   AnnotationType,
   assert,
+  bytesToString,
   escapeString,
   getModificationDate,
   isString,
@@ -30,7 +32,15 @@ import {
   warn,
 } from "../shared/util.js";
 import { Catalog, FileSpec, ObjectLoader } from "./obj.js";
-import { Dict, isDict, isName, isRef, isStream, Name } from "./primitives.js";
+import {
+  Dict,
+  isDict,
+  isName,
+  isRef,
+  isStream,
+  Name,
+  RefSet,
+} from "./primitives.js";
 import { ColorSpace } from "./colorspace.js";
 import { getInheritableProperty } from "./core_utils.js";
 import { OperatorList } from "./operator_list.js";
@@ -570,6 +580,20 @@ class Annotation {
   }
 
   /**
+   * Get field data for usage in JS sandbox.
+   *
+   * Field object is defined here:
+   * https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=16
+   *
+   * @public
+   * @memberof Annotation
+   * @returns {Object | null}
+   */
+  getFieldObject() {
+    return null;
+  }
+
+  /**
    * Reset the annotation.
    *
    * This involves resetting the various streams that are either cached on the
@@ -903,6 +927,7 @@ class WidgetAnnotation extends Annotation {
 
     data.annotationType = AnnotationType.WIDGET;
     data.fieldName = this._constructFieldName(dict);
+    data.actions = this._collectActions(params.xref, dict);
 
     const fieldValue = getInheritableProperty({
       dict,
@@ -937,6 +962,7 @@ class WidgetAnnotation extends Annotation {
     }
 
     data.readOnly = this.hasFieldFlag(AnnotationFieldFlag.READONLY);
+    data.hidden = this.hasFieldFlag(AnnotationFieldFlag.HIDDEN);
 
     // Hide signatures because we cannot validate them, and unset the fieldValue
     // since it's (most likely) a `Dict` which is non-serializable and will thus
@@ -944,6 +970,7 @@ class WidgetAnnotation extends Annotation {
     if (data.fieldType === "Sig") {
       data.fieldValue = null;
       this.setFlags(AnnotationFlag.HIDDEN);
+      data.hidden = true;
     }
   }
 
@@ -1366,6 +1393,87 @@ class WidgetAnnotation extends Annotation {
     }
     return localResources || Dict.empty;
   }
+
+  _collectJS(entry, xref, list, parents) {
+    if (!entry) {
+      return;
+    }
+
+    let parent = null;
+    if (isRef(entry)) {
+      if (parents.has(entry)) {
+        // If we've already found entry then we've a cycle.
+        return;
+      }
+      parent = entry;
+      parents.put(parent);
+      entry = xref.fetch(entry);
+    }
+    if (Array.isArray(entry)) {
+      for (const element of entry) {
+        this._collectJS(element, xref, list, parents);
+      }
+    } else if (entry instanceof Dict) {
+      if (isName(entry.get("S"), "JavaScript") && entry.has("JS")) {
+        const js = entry.get("JS");
+        let code;
+        if (isStream(js)) {
+          code = bytesToString(js.getBytes());
+        } else {
+          code = js;
+        }
+        code = stringToPDFString(code);
+        if (code) {
+          list.push(code);
+        }
+      }
+      this._collectJS(entry.getRaw("Next"), xref, list, parents);
+    }
+
+    if (parent) {
+      parents.remove(parent);
+    }
+  }
+
+  _collectActions(xref, dict) {
+    const actions = Object.create(null);
+    if (dict.has("AA")) {
+      const additionalActions = dict.get("AA");
+      for (const key of additionalActions.getKeys()) {
+        if (key in AnnotationActionEventType) {
+          const actionDict = additionalActions.getRaw(key);
+          const parents = new RefSet();
+          const list = [];
+          this._collectJS(actionDict, xref, list, parents);
+          if (list.length > 0) {
+            actions[AnnotationActionEventType[key]] = list;
+          }
+        }
+      }
+    }
+    // Collect the Action if any (we may have one on pushbutton)
+    if (dict.has("A")) {
+      const actionDict = dict.get("A");
+      const parents = new RefSet();
+      const list = [];
+      this._collectJS(actionDict, xref, list, parents);
+      if (list.length > 0) {
+        actions.Action = list;
+      }
+    }
+    return actions;
+  }
+
+  getFieldObject() {
+    if (this.data.fieldType === "Sig") {
+      return {
+        id: this.data.id,
+        value: null,
+        type: "signature",
+      };
+    }
+    return null;
+  }
 }
 
 class TextWidgetAnnotation extends WidgetAnnotation {
@@ -1515,6 +1623,23 @@ class TextWidgetAnnotation extends WidgetAnnotation {
     }
 
     return chunks;
+  }
+
+  getFieldObject() {
+    return {
+      id: this.data.id,
+      value: this.data.fieldValue,
+      multiline: this.data.multiLine,
+      password: this.hasFieldFlag(AnnotationFieldFlag.PASSWORD),
+      charLimit: this.data.maxLen,
+      comb: this.data.comb,
+      editable: !this.data.readOnly,
+      hidden: this.data.hidden,
+      name: this.data.fieldName,
+      rect: this.data.rect,
+      actions: this.data.actions,
+      type: "text",
+    };
   }
 }
 
@@ -1793,6 +1918,28 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       docBaseUrl: params.pdfManager.docBaseUrl,
     });
   }
+
+  getFieldObject() {
+    let type = "button";
+    let value = null;
+    if (this.data.checkBox) {
+      type = "checkbox";
+      value = this.data.fieldValue && this.data.fieldValue !== "Off";
+    } else if (this.data.radioButton) {
+      type = "radiobutton";
+      value = this.data.fieldValue === this.data.buttonValue;
+    }
+    return {
+      id: this.data.id,
+      value,
+      editable: !this.data.readOnly,
+      name: this.data.fieldName,
+      rect: this.data.rect,
+      hidden: this.data.hidden,
+      actions: this.data.actions,
+      type,
+    };
+  }
 }
 
 class ChoiceWidgetAnnotation extends WidgetAnnotation {
@@ -1842,6 +1989,23 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
     this.data.combo = this.hasFieldFlag(AnnotationFieldFlag.COMBO);
     this.data.multiSelect = this.hasFieldFlag(AnnotationFieldFlag.MULTISELECT);
     this._hasText = true;
+  }
+
+  getFieldObject() {
+    const type = this.data.combo ? "combobox" : "listbox";
+    const value =
+      this.data.fieldValue.length > 0 ? this.data.fieldValue[0] : null;
+    return {
+      id: this.data.id,
+      value,
+      editable: !this.data.readOnly,
+      name: this.data.fieldName,
+      rect: this.data.rect,
+      multipleSelection: this.data.multiSelect,
+      hidden: this.data.hidden,
+      actions: this.data.actions,
+      type,
+    };
   }
 }
 
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -708,7 +708,7 @@ class PDFDocument {
   }
 
   get formInfo() {
-    const formInfo = { hasAcroForm: false, hasXfa: false };
+    const formInfo = { hasAcroForm: false, hasXfa: false, fields: null };
     const acroForm = this.catalog.acroForm;
     if (!acroForm) {
       return shadow(this, "formInfo", formInfo);
@@ -736,6 +736,9 @@ class PDFDocument {
       const hasOnlyDocumentSignatures =
         !!(sigFlags & 0x1) && this._hasOnlyDocumentSignatures(fields);
       formInfo.hasAcroForm = hasFields && !hasOnlyDocumentSignatures;
+      if (hasFields) {
+        formInfo.fields = fields;
+      }
     } catch (ex) {
       if (ex instanceof MissingDataException) {
         throw ex;
@@ -934,6 +937,73 @@ class PDFDocument {
     return this.catalog
       ? this.catalog.cleanup(manuallyTriggered)
       : clearPrimitiveCaches();
+  }
+
+  _collectFieldObjects(name, fieldRef, promises) {
+    const field = this.xref.fetchIfRef(fieldRef);
+    if (field.has("T")) {
+      const partName = stringToPDFString(field.get("T"));
+      if (name === "") {
+        name = partName;
+      } else {
+        name = `${name}.${partName}`;
+      }
+    }
+
+    if (!(name in promises)) {
+      promises.set(name, []);
+    }
+    promises.get(name).push(
+      AnnotationFactory.create(
+        this.xref,
+        fieldRef,
+        this.pdfManager,
+        this._localIdFactory
+      )
+        .then(annotation => annotation.getFieldObject())
+        .catch(function (reason) {
+          warn(`_collectFieldObjects: "${reason}".`);
+          return null;
+        })
+    );
+
+    if (field.has("Kids")) {
+      const kids = field.get("Kids");
+      for (const kid of kids) {
+        this._collectFieldObjects(name, kid, promises);
+      }
+    }
+  }
+
+  get fieldObjects() {
+    const formInfo = this.formInfo;
+    if (!formInfo.fields) {
+      return shadow(this, "fieldObjects", Promise.resolve(null));
+    }
+
+    const allFields = Object.create(null);
+    const fieldPromises = new Map();
+    for (const fieldRef of formInfo.fields) {
+      this._collectFieldObjects("", fieldRef, fieldPromises);
+    }
+
+    const allPromises = [];
+    for (const [name, promises] of fieldPromises.entries()) {
+      allPromises.push(
+        Promise.all(promises).then(fields => {
+          fields = fields.filter(field => field !== null);
+          if (fields.length > 0) {
+            allFields[name] = fields;
+          }
+        })
+      );
+    }
+
+    return shadow(
+      this,
+      "fieldObjects",
+      Promise.all(allPromises).then(() => allFields)
+    );
   }
 }
 
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -516,6 +516,10 @@ class WorkerMessageHandler {
       });
     });
 
+    handler.on("GetFieldObjects", function (data) {
+      return pdfManager.ensureDoc("fieldObjects");
+    });
+
     handler.on("SaveDocument", function ({
       numPages,
       annotationStorage,
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -876,6 +876,14 @@ class PDFDocumentProxy {
   saveDocument(annotationStorage) {
     return this._transport.saveDocument(annotationStorage);
   }
+
+  /**
+   * @returns {Promise<Array<Object>>} A promise that is resolved with an
+   *   {Array<Object>} containing field data for the JS sandbox.
+   */
+  getFieldObjects() {
+    return this._transport.getFieldObjects();
+  }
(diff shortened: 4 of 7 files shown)


src/shared/util.js
```diff
@@ -310,6 +310,28 @@ const PasswordResponses = {
   INCORRECT_PASSWORD: 2,
 };
 
+const EVENT_MAP = {
```
Snuffleupagus, line 313: Can we please use a more descriptive, and less general, name here?

How about something like e.g. `AnnotationActionEventType` or something, which is more consistent with other Annotation types defined earlier in this file?

src/core/annotation.js
```diff
@@ -1367,6 +1375,71 @@ class WidgetAnnotation extends Annotation {
...
+      if (parents.has(entry)) {
+        // If we've already met entry then we've a cycle
+        return;
+      }
+      parent = entry;
+      parents.add(parent);
+      entry = xref.fetch(entry, true);
+    }
+    if (Array.isArray(entry)) {
+      for (const e of entry) {
```
Snuffleupagus, line 1395: Nit: Please no one-letter variable names, use e.g. `element` instead,

src/core/annotation.js
```diff
@@ -1367,6 +1375,71 @@ class WidgetAnnotation extends Annotation {
...
+        this._collectJS(e, xref, list, parents);
+      }
+    } else if (entry instanceof Dict) {
+      if (entry.get("S") === Name.get("JavaScript") && entry.has("JS")) {
```
Snuffleupagus, line 1399: Sorry, but the `Name` comparison-part needs to read `isName(entry.get("S"), "JavaScript")` instead, since this could potentially break once the `Name`-cache has been reset.

src/core/document.js
```diff
@@ -227,6 +227,27 @@ class Page {
     return stream;
   }
 
+  getAnnotationObjects() {
+    // Fetch the page's annotations and get annotation data
+    // to be used in JS sandbox
```
Snuffleupagus, line 232: Nit: Missing a period at the end of the sentence.

src/core/annotation.js
```diff
@@ -1367,6 +1375,71 @@ class WidgetAnnotation extends Annotation {
...
+
+  _collectJS(entry, xref, list, parents) {
+    if (!entry) {
+      return;
+    }
+
+    let parent = null;
+    if (isRef(entry)) {
+      if (parents.has(entry)) {
+        // If we've already met entry then we've a cycle
```
Snuffleupagus, line 1387: Nit: `met` -> `found` sounds better, and please add a period at the end of the sentence.

src/core/annotation.js
```diff
@@ -1791,6 +1882,29 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
+      type = "checkbox";
+      value = this.data.fieldValue && this.data.fieldValue !== "Off";
+    } else if (this.data.radioButton) {
+      type = "radiobutton";
+      value = this.data.fieldValue === this.data.buttonValue;
+    }
+    return {
+      id: this.data.id,
+      value,
+      valueAsString: value.toString(),
```
Snuffleupagus, line 1899: Can `value` still be `null` here, because `null.toString()` will throw!?

src/core/annotation.js
```diff
@@ -1517,6 +1590,24 @@ class TextWidgetAnnotation extends WidgetAnnotation {
 
     return chunks;
   }
+
+  async getAnnotationObject() {
+    return {
+      id: this.data.id,
+      value: this.data.fieldValue,
+      valueAsString: this.data.fieldValue.toString(),
```
Snuffleupagus, line 1598: Is `this.data.fieldValue` guaranteed to be of a type that has a `toString()` method, because otherwise there'll be errors!?

src/core/annotation.js
```diff
@@ -1841,6 +1955,24 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
...
   }
+
+  async getAnnotationObject() {
+    const type = this.data.combo ? "combobox" : "listbox";
+    const value =
+      this.data.fieldValue.length > 0 ? this.data.fieldValue[0] : null;
+    return {
+      id: this.data.id,
+      value,
+      valueAsString: value.toString(),
```
Snuffleupagus, line 1966: Since `value` can be `null`, according to the definition above, this can throw.

src/display/api.js
```diff
@@ -2550,6 +2558,18 @@ class WorkerTransport {
...
 
+  getAnnotationObjects() {
+    return this.messageHandler
+      .sendWithPromise("GetAnnotationObjects", {
+        numPages: this._numPages,
+      })
+      .then(data => {
+        return data
+          .reduce((acc, x) => acc.concat(x), [])
+          .filter(x => x !== null);
```
Snuffleupagus, line 2569: This feels overly terse, and it's honesty a bit difficult to understand exactly what this code actually accomplishes.
Can this be improved somewhat? And please also use longer/more descriptive variable names here.

Finally, any reason not to do these manipulations on the worker-side before sending the data to the API?

src/core/annotation.js
```diff
@@ -1367,6 +1375,71 @@ class WidgetAnnotation extends Annotation {
...
+  }
+
+  _collectActions(xref, dict) {
+    const actions = Object.create({});
+    if (dict.has("AA")) {
+      const additionalAction = dict.get("AA");
+      for (const key of additionalAction.getKeys()) {
+        if (key in EVENT_MAP) {
+          const actionDict = additionalAction.getRaw(key);
+          const parents = new Set();
```
Snuffleupagus, line 1420: Given how this is actually used in `_collectJS`, for storing of `Ref`s, you really want a `RefSet` instead here to prevent issues related to caching.

src/core/annotation.js
```diff
@@ -1367,6 +1375,71 @@ class WidgetAnnotation extends Annotation {
...
+
+    let parent = null;
+    if (isRef(entry)) {
+      if (parents.has(entry)) {
+        // If we've already met entry then we've a cycle
+        return;
+      }
+      parent = entry;
+      parents.add(parent);
+      entry = xref.fetch(entry, true);
```
Snuffleupagus, line 1392: It's extremely rare that you need to manually set the `suppressEncryption` parameter when calling `XRef.fetch`, and in general it *may* not even be correct to use it unless you're fetching top-level (in the PDF document tree) data.

Is there an actual test-case that necessitated this?

Snuffleupagus (changes requested):
Based on a *very* cursory look, I've added a few quick comments.

timvandermeij:
I'm missing some context for this patch. Which issue does this solve? Why are we interested in collecting this information, and what is it used for?

calixteman:
> I'm missing some context for this patch. Which issue does this solve? Why are we interested in collecting this information, and what is it used for?

I'll fix the nits tomorrow: thanks for that.
We plan to add support to exectute js stuff embedded in the PDF.
A first step has been made here: https://phabricator.services.mozilla.com/D91746 for the m-c version.
And we plan to add something else (probably quickjs in using emscripten) for the extension version.
Anyway if there is an open issue for js stuff, then please assign me.

src/core/worker.js
```diff
@@ -515,6 +515,18 @@ class WorkerMessageHandler {
...
+    handler.on("GetAnnotationObjects", function ({ numPages }) {
+      const promises = [];
+      for (let pageIndex = 0; pageIndex < numPages; pageIndex++) {
+        promises.push(
+          pdfManager.getPage(pageIndex).then(function (page) {
+            return page.getAnnotationObjects();
+          })
+        );
+      }
+      return Promise.all(promises).then(data => data.flat());
```
Snuffleupagus, line 527: Given that there's not been any recent decisions, as far as I'm aware, of the lowest supported Chrome/Safari versions you'll need to add a polyfill for this Array method in `src/shared/compatibility.js`; similar to e.g. https://github.com/mozilla/pdf.js/blob/88f72d6b1c37dfc0dbb6715685bf8cb4ae516cd7/src/shared/compatibility.js#L81-L97

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#Browser_compatibility

test/unit/annotation_spec.js
```diff
@@ -1863,6 +1863,92 @@ describe("annotation", function () {
...
+        xref,
+        textWidgetRef,
+        pdfManagerMock,
+        idFactoryMock
+      )
+        .then(annotation => {
+          return annotation.getAnnotationObject();
+        })
+        .then(object => {
+          console.log(object.actions);
```
Snuffleupagus, line 1938: This looks like a left-over debug statement, so please remove it :-)

src/core/annotation.js
```diff
@@ -1367,6 +1384,76 @@ class WidgetAnnotation extends Annotation {
...
+      this._collectJS(entry.getRaw("Next"), xref, list, parents);
+    }
+
+    if (parent) {
+      parents.remove(parent);
+    }
+  }
+
+  _collectActions(xref, dict) {
+    const actions = Object.create({});
```
Snuffleupagus, line 1429: Why not simply `Object.create(null);` here?

calixteman: isn't the same ? (I'm not a JS expert)

Snuffleupagus: Well, they might not be identical; but why do you *need* that special format here?

calixteman: dunno ;)
I'll change it.

Snuffleupagus (changes requested):
Sorry, but the API implemented here is really not correct!

Similar to the regular `getAnnotations` method, which lives on the *page* rather than the document, this new API-method *must* also be on the page-level.

Please keep in mind that the PDF.js library is written such that data can be loaded when needed, using e.g. range requests and streaming, to explicitly *not* require that the entire document is loaded in order for rendering to start (see e.g. the `disableAutoFetch` option).

Hence we *really* can't/shouldn't add an API-method that essentially forces the *entire* document to load and *all* pages to be parsed, which would be particularly bad when invoked from the viewer. (It would force *a lot* of data to be requested very early on, possibly affecting general viewer performance negatively when loading longer documents.)

calixteman:
I thought about that of course and the problem is when a script is getting a value from a Field which hasn't been rendered: we can't await for its value.
Setting a value shouldn't be a problem: we can put it in annotationStorage if the Field hasn't been rendered and use this value when rendered for this first time.

What we can do:
 - only get all the annotations and then run the sandbox when at least one contains some JS
 - make the hypothesis that no script is trying to get a value from a non-rendered field (I guess it should be pretty rare but tbh I don't know) and then use a by-page approach.
 - or another good idea I didn't think about: any suggestion will be appreciated.

calixteman:
I moved the data collection stuff in PDFDocument in using Fields entry from Acroform dictionary.
So this way, we don't parse all the pages.
@Snuffleupagus does it sound better ?

src/core/annotation.js
```diff
@@ -1516,6 +1603,23 @@ class TextWidgetAnnotation extends WidgetAnnotation {
 
     return chunks;
   }
+
+  getAnnotationObject() {
```
timvandermeij, line 1607: Why do we need this `getAnnotationObject` method? The data passed between the core and display layers is all in the `this.data` attribute, and most of that information is already in there, so why can't we just use that?

calixteman: Because in this case we'll get only the data for the rendered Fields which should be fine most of the time.
But if a script has something like `var foo = this.getField("foo");do_whatever_with(foo);` and if foo hasn't been rendered then foo will be undefined. So in this case we could get lazily foo but it means that we should wait for an async function.
So the only solution I found (which doesn't mean that's the only solution) is to get all Fields.
To avoid to much redundancy, we could add a cache (if none) for the annotations (at least for the Fields): I guess we could expect that a user will read all the form just to be sure that no data are missing, it'll add some overhead at document loading but annotation rendering will be then slightly faster since we've a cache (but we use more memory...).
If you've any ideas on how to do that nicely then feel free to share them, I'm a kinda stuck here.

Snuffleupagus:
> I moved the data collection stuff in PDFDocument in using Fields entry from Acroform dictionary.
> So this way, we don't parse all the pages.
> @Snuffleupagus does it sound better ?

I suppose I still don't really understand why this, similar to annotations in general, cannot be parsed on a page-by-page basis!?
Are you perhaps expecting that JavaScript actions on one page will affect fields on *other* pages?

(Generally speaking, given the lack of context/information in the commit message it's quite difficult to get a good overview of the entire situation and e.g. understand why you went with a particular approach.)

calixteman:
> > I moved the data collection stuff in PDFDocument in using Fields entry from Acroform dictionary.
> > So this way, we don't parse all the pages.
> > @Snuffleupagus does it sound better ?
> 
> I suppose I still don't really understand why this, similar to annotations in general, cannot be parsed on a page-by-page basis!?
> Are you perhaps expecting that JavaScript actions on one page will affect fields on _other_ pages?

Yes exactly

> 
> (Generally speaking, given the lack of context/information in the commit message it's quite difficult to get a good overview of the entire situation and e.g. understand why you went with a particular approach.)

My bad, you're right, sorry about that.
Anyway I gave more explanations in the different comments I made (e.g. https://github.com/mozilla/pdf.js/pull/12429#discussion_r500425603).

src/core/annotation.js
timvandermeij: Right, this explanation does help to get a bit more context here. Do we actually have examples of PDF files that do that or is it more a theoretical problem (is it even allowed by the spec to access fields from other pages)? If you really need all fields, I suppose that fetching the information from the `AcroForm` dictionary is the best as it doesn't require rendering all pages, which I think would be a no-go. I'm not sure though if the `AcroForm` dictionary is guaranteed to contain all form fields (some PDFs don't have the `AcroForm` dictionary, or work with `XFA`)...

timvandermeij: For me this still leaves the question why this method is necessary instead of just using `this.data` because in `document.js` you're using `AnnotationFactory.create` which will still create the entire annotation object anyway. Instead of calling `annotation.getAnnotationObject()` you might as well use `annotation.data` as far as I can tell. This would at least make sure we have one and only one way of exposing the data, and practically all fields in this method are already in `this.data` anyway...

calixteman: > Right, this explanation does help to get a bit more context here. Do we actually have examples of PDF files that do that or is it more a theoretical problem (is it even allowed by the spec to access fields from other pages)?

I don't have such a pdf but I guess it's pretty rare but having some fields to fill on page 1 and a field with the total sum on page 2 is likely existing (at least in France, forms for taxes are pretty long).
The method `Doc.getField` (https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=301) allows to get a `Field` by its name (see page 413 too) and I don't see any mention that the `Field` must have been rendered.

> If you really need all fields, I suppose that fetching the information from the `AcroForm` dictionary is the best as it doesn't require rendering all pages, which I think would be a no-go. I'm not sure though if the `AcroForm` dictionary is guaranteed to contain all form fields (some PDFs don't have the `AcroForm` dictionary, or work with `XFA`)...

I don't know too. It should possible to build `AcroForm::Fields` using all the found annotations and their `T` field but in this case it'd require to read all the pages...
And about XFA, we plane to do that later.

calixteman: > For me this still leaves the question why this method is necessary instead of just using `this.data` because in `document.js` you're using `AnnotationFactory.create` which will still create the entire annotation object anyway. Instead of calling `annotation.getAnnotationObject()` you might as well use `annotation.data` as far as I can tell. This would at least make sure we have one and only one way of exposing the data, and practically all fields in this method are already in `this.data` anyway...

Agreed, I thought doing that but to have correct names it'd have required more change and I wanted to limit the changes in this patch. And by the way they're some "useless" (because they aren't in `Field` object) props in data.
I think we can do that in a follow-up (even in this one if we've an agreement on the way to do). 
An idea is to put these data (with correct names) in `AnnotationStorage` since I use this storage to store data coming from JS in case the annotation hasn't been rendered. Wdyt ?

calixteman: After thinking again about that, we can change names in `annotation.data` but it means that we potentially break something (api compat) and I'll still need to strip out useless properties at some point.

timvandermeij: You're right that we should break the API compatibility. How about instead of changing anything in `this.data` just adding a new key with the data you need for the JavaScript actions? In that case you can just augment `this.data` with a new key, say `javaScriptActions` with has an object as value, put your key/value pairs in there and when fetching the data only taking that key and ignoring the existing fields.

calixteman: I can do that but it means that we'll keep a ref on these data and once they have been consumed by the sandbox we don't need them anymore. So we'll consume memory for nothing.
But if your purpose is to expose that stuff, I'm fine to change that of course.

calixteman:
Can we move forward on this patch please ?

test/unit/annotation_spec.js
```diff
@@ -1863,6 +1863,91 @@ describe("annotation", function () {
...
+      ]);
+
+      const JS = Name.get("JavaScript");
+      const additionalActionDict = new Dict();
+      const eDict = new Dict();
+      eDict.set("JS", "hello()");
+      eDict.set("S", JS);
+      additionalActionDict.set("E", eDict);
+
+      xDict.set("JS", "world()");
```
timvandermeij, line 1899: Let's add a comment above this line saying `// Note that we also test the cycle detection logic here.` or something similar to indicate that the `xDictRef` in the `Next` key is put there on purpose.

timvandermeij (commented):
Now that it's finally clear to me why the extra method is needed, I think this approach can be used now with these comments addressed, in particular to add some documentation on why they are there (see the comment below).

src/core/annotation.js
timvandermeij: See the new comment below. I only now understood why you actually need this.

src/core/annotation.js
```diff
@@ -1366,6 +1383,76 @@ class WidgetAnnotation extends Annotation {
...
+            actions[AnnotationActionEventType[key]] = list;
+          }
+        }
+      }
+    }
+    return actions;
+  }
+
+  getAnnotationObject() {
+    if (this.data.fieldType === "Sig") {
```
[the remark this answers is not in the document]
calixteman: It's just to avoid to return an `undefined` if this object is used in JS.

src/core/worker.js
```diff
@@ -516,6 +516,10 @@ class WorkerMessageHandler {
       });
     });
 
+    handler.on("GetFieldObjects", function (data) {
+      return pdfManager.pdfDocument.getFieldObjects();
```
Snuffleupagus, line 520: I missed this previously, but please note that it's *generally not correct* to access data like this *anywhere* in this file. There may of course be exceptions, but unless you're using `pdfManager.onLoadedStream()` (which should be used very carefully/sparingly) you should generally not do this!

Please note that before data has been loaded, there's a very real risk that `MissingDataException`s are thrown during parsing and as-is this code would thus incorrectly return a rejected promise to the main-thread and break parsing; hence please change this to (and please keep it in mind for the future):
```suggestion
      return pdfManager.ensureDoc("getFieldObjects");
```

src/core/document.js
```diff
@@ -935,6 +938,65 @@ class PDFDocument {
...
+        this._collectFieldObjects(name, kid, promises);
+      }
+    }
+  }
+
+  async getFieldObjects() {
+    const formInfo = this.formInfo;
+    const allFields = Object.create(null);
+    if (!formInfo.hasAcroForm) {
+      return allFields;
```
Snuffleupagus, line 982: Would it not make sense to return `null` instead in this case, since that's consistent with most existing API-methods!?

src/core/document.js
```diff
@@ -935,6 +938,65 @@ class PDFDocument {
...
+    }
+
+    const fieldPromises = Object.create(null);
```
Snuffleupagus, line 985: Nit: The code might become slightly simpler if you use a `Map` instead here?

src/core/document.js
```diff
@@ -935,6 +938,65 @@ class PDFDocument {
...
+
+    if (field.has("Kids")) {
+      const kids = field.get("Kids");
+      for (const kid of kids) {
+        this._collectFieldObjects(name, kid, promises);
+      }
+    }
+  }
+
+  async getFieldObjects() {
```
Snuffleupagus, line 978: Any reason to not simply make this a getter instead?
I.e. `async get fieldObjects() {` instead, since then you could also `shadow` the return values similar to most existing methods in `PDFDocument`.

calixteman: I did that but `async get foo` is not allowed and then I thought that it was useless to keep in memory the `fieldObjects` because we'll use it only one time.

src/core/worker.js
calixteman: Thx for the tip

src/core/document.js
```diff
@@ -226,6 +226,25 @@ class Page {
     return stream;
   }
 
+  getAnnotationObjects() {
+    // Fetch the page's annotations and get annotation data
+    // to be used in JS sandbox.
+    return this._parsedAnnotations.then(function (annotations) {
+      const results = [];
+      for (const annotation of annotations) {
+        if (!isAnnotationRenderable(annotation, "print")) {
```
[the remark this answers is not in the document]
calixteman: Useless code...

src/shared/util.js
```diff
@@ -144,6 +144,28 @@ const AnnotationBorderStyleType = {
...
 };
 
+const AnnotationActionEventType = {
+  E: "MouseEnter",
+  X: "MouseExit",
+  D: "MouseDown",
+  U: "MouseUp",
+  Fo: "Focus",
+  Bl: "Blur",
+  PO: "PageOpen",
```
timvandermeij, line 154: `PageOpened` would be more consistent with `PageClosed` below. Is there a spec for these event names? In https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=16 I only find _some_ of them.

test/unit/annotation_spec.js
```diff
@@ -1863,6 +1863,92 @@ describe("annotation", function () {
...
+        { ref: next1Ref, data: next1Dict },
+        { ref: next2Ref, data: next2Dict },
+      ]);
+
+      const JS = Name.get("JavaScript");
+      const additionalActionDict = new Dict();
+      const eDict = new Dict();
+      eDict.set("JS", "hello()");
+      eDict.set("S", JS);
+      additionalActionDict.set("E", eDict);
```
timvandermeij, line 1897: `additionalActionsDict` (same rename as above)

timvandermeij (approved):
Approved pending these final comments now that I did a full review, and passing tests afterwards (ignore the intermittent one above). Thanks!

src/shared/util.js
calixteman: I changed `PageClosed` into `PageClose`:
https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=387

src/core/annotation.js
```diff
@@ -1516,6 +1623,23 @@ class TextWidgetAnnotation extends WidgetAnnotation {
 
     return chunks;
   }
+
+  getFieldObject() {
+    return {
+      id: this.data.id,
+      value: this.data.fieldValue,
+      multiline: this.data.multiline,
```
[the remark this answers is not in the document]
Snuffleupagus: For consistency, shouldn't the key have the same capitalization as well (i.e. `multiLine`)?
Having inconsistent casing does seem like a good way of *accidentally* introducing bugs elsewhere.

Snuffleupagus (approved):
Seems fine to me too, with just one more question; please see https://github.com/mozilla/pdf.js/pull/12429#discussion_r504595771

src/core/annotation.js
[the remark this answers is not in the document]
calixteman: On js side multiline property is lowercase:
https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=433
So here the keys reflect what we've in the js api.
I'd prefer changing the key names in `this.data` for consistency, but it'd broke the api (pdf.js' one)...
@Snuffleupagus, I probably made this typo because of this mismatch so I strongly agree with you.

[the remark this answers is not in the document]
timvandermeij: It's indeed not possible to change this now because it's an API-breaking change.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #12436, opened 2020-10-02
Enable the ESLint `no-var` rule in the `src/shared/` folder

Snuffleupagus:
Previously this rule has been enabled in the `web/` folder, and in select files in the `src/` sub-folders.
In this case, enabling of this rule didn't actually require any further code changes.

### DIFF
--- a/src/shared/message_handler.js
+++ b/src/shared/message_handler.js
@@ -12,7 +12,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint no-var: error, prefer-const: error */
 
 import {
   AbortException,
--- a/src/shared/.eslintrc
+++ b/src/shared/.eslintrc
@@ -1,0 +1,10 @@
+{
+  "extends": [
+    "../../.eslintrc"
+  ],
+
+  "rules": {
+    // ECMAScript 6
+    "no-var": "error",
+  },
+}
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -12,7 +12,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint no-var: error */
 
 import { isNodeJS } from "./is_node.js";
 
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -12,7 +12,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint no-var: error */
 
 import "./compatibility.js";
 


src/shared/cffStandardStrings.js
```diff
@@ -12,6 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* eslint-disable no-var */
```
timvandermeij, line 15: Why disable this here? It looks like `var` can be made `const` in this file without problems?

src/shared/message_handler.js
```diff
@@ -12,7 +12,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint no-var: error, prefer-const: error */
```
timvandermeij, line 15: Is the removal of `prefer-const` intended?

src/shared/fonts_utils.js
```diff
@@ -12,6 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* eslint-disable no-var */
```
timvandermeij, line 15: Is ESLint not able to fix this file for us?

src/shared/message_handler.js
Snuffleupagus: Yes, since it's enabled globally now; please see https://github.com/mozilla/pdf.js/blob/88f72d6b1c37dfc0dbb6715685bf8cb4ae516cd7/.eslintrc#L190

src/shared/cffStandardStrings.js
Snuffleupagus: Given that this file isn't actually used for anything, I figured that ignoring it makes sense.
However, I missed that it's already excluded from linting so this line can just be removed; see https://github.com/mozilla/pdf.js/blob/88f72d6b1c37dfc0dbb6715685bf8cb4ae516cd7/.eslintignore#L10-L11

src/shared/fonts_utils.js
Snuffleupagus: Given that this file isn't actually used for anything, I figured that ignoring it makes sense.
However, I missed that it's already excluded from linting so this line can just be removed; see https://github.com/mozilla/pdf.js/blob/88f72d6b1c37dfc0dbb6715685bf8cb4ae516cd7/.eslintignore#L10-L11

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12437, opened 2020-10-02
Enable the ESLint `no-var` rule in the `src/display/` folder 

Snuffleupagus:
Previously this rule has been enabled in the `web/` folder, and in select files in the `src/` sub-folders.
Note that a number of the files in the `src/display/` folder were already enforcing the `no-var` rule, and thanks to Prettier the necessary re-writing will be (mostly) handled automatically.

### DIFF
--- a/src/display/.eslintrc
+++ b/src/display/.eslintrc
@@ -1,0 +1,10 @@
+{
+  "extends": [
+    "../../.eslintrc"
+  ],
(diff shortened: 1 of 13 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for doing this!

PR #12438, opened 2020-10-03
Implement viewer option for changing the reading direction

### LINKED ISSUE #11154: Function suggestion: Horizontal Scrolling from right to left

baloone:
Fixes #11154.


baloone:
What remains to do is add translations and maybe a little icon next to the buttons

timvandermeij:
Could you squash the commits into one for easier review? Refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits if you're not familiar with that.

/botio-linux preview

timvandermeij:
Seems to work nicely at a quick first glance!

For the translations, you only have to add the new strings to https://github.com/mozilla/pdf.js/blob/master/l10n/en-US/viewer.properties. If this PR is merged, the Mozilla translation teams will automatically get a ping to translate it to their languages.

For the icons, maybe @utopianknight has time to provide two SVG icons for this?

Finally, please change the commit message to `Implement viewer option for changing the reading direction` or something similar like I've done for this PR so it's immediately clear what the commit/PR does from the commit history without having to look up the issue.

baloone:
Thank you very much for the review, I will make the changes you suggest.

Snuffleupagus (changes requested):
There's a couple of things that we need to agree upon/fix, before we can consider merging this:
 - Figure out how to handle RTL locales, where the entire viewer layout is reversed, to avoid inconsistency issues.
 - Ensure that the `getVisibleElements` helper function works correctly, when the direction is changed.
 - Add icons, as mentioned elsewhere. (I think having *two* buttons makes a lot of sense, given the viewer-RTL point mentioned above.)
 - Improve the overall implementation, since there's a number of things that based on a very quick look must be fixed.

baloone:
> There's a couple of things that we need to agree upon/fix, before we can consider merging this:
> 
> * Figure out how to handle RTL locales, where the entire viewer layout is reversed, to avoid inconsistency issues.
> * Ensure that the `getVisibleElements` helper function works correctly, when the direction is changed.
> * Add icons, as mentioned elsewhere. (I think having _two_ buttons makes a lot of sense, given the viewer-RTL point mentioned above.)
> * Improve the overall implementation, since there's a number of things that based on a very quick look must be fixed.

I added an icon. I don't know how to test the `getVisibleElements` function but I believe it works since the correct pages are loaded. At first I used css's `direction: rtl` but the pages weren't loading, then I found a workaround: I use` scaleX (-1) ` twice, on the container and the pages.

baloone:
I also reduced the two buttons to just one because I hadn't read your message.

baloone:
@Snuffleupagus Is it ok ?

utopianknight:
This is how it looks like: 

Screenshot 2020-10-07 094728
Screenshot 2020-10-07 094753

secondaryToolbarButton-toggleReadingDirection.svg
`<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg"><path d="M10.97.008c-.743.049-1.35 1.002-.668 1.709L12.113 3H3a1 1 0 00-1 1 1 1 0 001 1h9.113l-1.811 1.283c-.91.943.472 2.324 1.414 1.414l2.99-2.99a1 1 0 000-1.414l-2.99-2.99c-.235-.228-.498-.311-.746-.295zm-5.941 8c-.248-.016-.51.067-.746.295l-2.99 2.99a1 1 0 000 1.414l2.99 2.99c.942.91 2.324-.471 1.414-1.414L3.886 13h9.113a1 1 0 001-1 1 1 0 00-1-1H3.886l1.81-1.283c.683-.707.076-1.66-.667-1.71z"/></svg>`

secondaryToolbarButton-toggleReadingDirection-dark.svg
`<svg width="16" height="16" fill="rgba(255,255,255,1)" xmlns="http://www.w3.org/2000/svg"><path d="M10.97.008c-.743.049-1.35 1.002-.668 1.709L12.113 3H3a1 1 0 00-1 1 1 1 0 001 1h9.113l-1.811 1.283c-.91.943.472 2.324 1.414 1.414l2.99-2.99a1 1 0 000-1.414l-2.99-2.99c-.235-.228-.498-.311-.746-.295zm-5.941 8c-.248-.016-.51.067-.746.295l-2.99 2.99a1 1 0 000 1.414l2.99 2.99c.942.91 2.324-.471 1.414-1.414L3.886 13h9.113a1 1 0 001-1 1 1 0 00-1-1H3.886l1.81-1.283c.683-.707.076-1.66-.667-1.71z"/></svg>`

Snuffleupagus:
> At first I used css's `direction: rtl` but the pages weren't loading, then I found a workaround: I use`scaleX (-1)` twice, on the container and the pages.

That would point to a pre-existing bug in `getVisibleElements` for RTL-mode which we really need to fix *first*, before moving forward with this PR (since it's certainly possible that a simpler solution is available once that is addressed).

Furthermore, as-is, this PR would introduce *really* inconsistent and confusing behaviour for RTL-locales, which was the very first point mentioned in https://github.com/mozilla/pdf.js/pull/12438#pullrequestreview-501623289.

baloone:
> Furthermore, as-is, this PR would introduce _really_ inconsistent and confusing behaviour for RTL-locales, which was the very first point mentioned in [#12438 (review)](https://github.com/mozilla/pdf.js/pull/12438#pullrequestreview-501623289).

The fact that I replaced left to right and right to left buttons by a toggle button fixes that problem.

baloone:
@Snuffleupagus  Now that the problem with getvisibleelements is solved, could we reconsider this PR

web/viewer.css
```diff
@@ -1899,3 +1905,13 @@ html[dir="rtl"] #documentPropertiesOverlay .row > * {
     display: none;
   }
 }
+
+.dir-rev {
+  transform: scaleX(-1);
```
timvandermeij, line 1910: Now that the RTL issue is fixed, if you rebase this PR can't we remove this and the one below then, since you indicated it was a work-around?

baloone: We have three possible solutions : 
-Get back the two buttons "Left to right" and "right to left"
-Use getcomputedstyle to toggle the document direction.
-Leave it like that to avoid using getcomputedstyle

baloone:
I replaced `transform: scaleX(-1);` by this:
```css
html[dir=ltr] .dir-rev {
  direction: rtl;
}
html[dir=rtl] .dir-rev {
  direction: ltr;
}
```

timvandermeij:
Looks like a better solution. Let's make a new build to check the current status of this patch. I wanted to get back to this sooner, but we have been very busy with other patches having priority first. Our contributors are simply working way too hard ;-)

/botio-linux preview

baloone:
I'm sorry i forgot to run the css linter, it is fixed now

baloone:
any news ?

timvandermeij:
There's a merge conflict in the preferences schema, so this needs a rebase. @Snuffleupagus Could you check if the updated patch addresses your initial points sufficiently or indicate which changes are still needed from your point of view?

Snuffleupagus:
> Could you check if the updated patch addresses your initial points sufficiently or indicate which changes are still needed from your point of view?

This still needs to use *two* buttons (as in the first version of the patch), such that it's possible for the user to easily determine the current reading direction. So once more, please implement this using *two* buttons as asked for previously :-)

The reason for request is really simple: It's possible for a PDF document itself to control the reading direction, which as a *follow-up* PR to this one we'll probably want to look into supporting. When that's done, in order to the reduce user confusion and (invalid) bug reports, we thus need to be able to show in the UI what the *current* reading direction actually is (hence the need for two buttons).

timvandermeij:
> It's possible for a PDF document itself to control the reading direction

Ah, I didn't know that was possible. In that case I agree two buttons would be better here.

baloone:
Thank you for the review, I got back the two buttons, please let me know if further changes are needed

web/app.js
```diff
@@ -1765,6 +1772,10 @@ const PDFViewerApplication = {
 
     setViewerModes(scrollMode, spreadMode);
 
+    if (Object.values(ReadingDirection).includes(direction)) {
+      this.pdfViewer.direction = direction;
+    }
+
```
Snuffleupagus, line 1778: Please follow the *same* pattern as with the other options here, by adding the following helper function in https://github.com/mozilla/pdf.js/blob/master/web/ui_utils.js#L693
```javascript
function isValidReadingDirection(direction) {
  return (
    Number.isInteger(mode) &&
    Object.values(ReadingDirection).includes(direction) &&
    direction !== ReadingDirection.UNKNOWN
  );
}
```
and use that here by adding function such as
```javascript
const setReadingDirection = direction => {
  if (isValidReadingDirection(direction)) {
    this.pdfViewer.readingDirection = direction;
  }
};
```
and finally replacing the block above with 
```suggestion
  setReadingDirection(readingDirection);
```

web/base_viewer.js
```diff
@@ -1476,6 +1478,35 @@ class BaseViewer {
     this._setCurrentPageNumber(pageNumber, /* resetCurrentPageView = */ true);
     this.update();
   }
+
+  /**
+   * @type {number} One of the values in {ReadingDirection}.
+   */
+  get direction() {
+    return this._direction;
+  }
```
Snuffleupagus, line 1487:
```suggestion
  get readingDirection() {
    return this._readingDirection;
  }
```
(And similar in the getter below.)

web/base_viewer.js
```diff
@@ -1476,6 +1478,35 @@ class BaseViewer {
...
+
+  /**
+   * @param {number} dir
```
Snuffleupagus, line 1490: This requires an actual description of the setter, please see how the other ones are documented.

web/base_viewer.js
```diff
@@ -1476,6 +1478,35 @@ class BaseViewer {
...
+   */
+  set direction(direction) {
+    if (this._direction === direction) {
+      return;
+    }
+    if (!Object.values(ReadingDirection).includes(direction)) {
```
Snuffleupagus, line 1496: The validation needs to be done using the `isValidReadingDirection` helper function (see other review comment).

web/viewer.html
```diff
@@ -209,9 +209,18 @@
               <span data-l10n-id="spread_even_label">Even Spreads</span>
             </button>
 
-            <div class="horizontalToolbarSeparator spreadModeButtons"></div>
+            <div class="horizontalToolbarSeparator"></div>
+            
+            <button id="ltrReadingDirection" class="secondaryToolbarButton readingDirection" title="Set reading direction from left to right" tabindex="68" data-l10n-id="ltr_reading_direction">
```
Snuffleupagus, line 214: The id should be `readingDirectionLtr` instead, with the `data-l10n-id`s updated accordingly as well.

web/viewer.html
```diff
@@ -209,9 +209,18 @@
...
+              <span data-l10n-id="ltr_reading_direction_label">Left to right</span>
+            </button>
+            <button id="rtlReadingDirection" class="secondaryToolbarButton readingDirection" title="Set reading direction from right to left" tabindex="68" data-l10n-id="rtl_reading_direction">
```
Snuffleupagus, line 217: The id should be `readingDirectionRtl` instead, with the `data-l10n-id`s updated accordingly as well.

web/viewer.html
```diff
@@ -209,9 +209,18 @@
               <span data-l10n-id="spread_even_label">Even Spreads</span>
             </button>
 
-            <div class="horizontalToolbarSeparator spreadModeButtons"></div>
```
Snuffleupagus, line 212: Changing this line will cause things to break, please don't do that!

web/viewer.html
```diff
@@ -209,9 +209,18 @@
...
+            <div class="horizontalToolbarSeparator"></div>
+            
+            <button id="ltrReadingDirection" class="secondaryToolbarButton readingDirection" title="Set reading direction from left to right" tabindex="68" data-l10n-id="ltr_reading_direction">
+              <span data-l10n-id="ltr_reading_direction_label">Left to right</span>
+            </button>
+            <button id="rtlReadingDirection" class="secondaryToolbarButton readingDirection" title="Set reading direction from right to left" tabindex="68" data-l10n-id="rtl_reading_direction">
+              <span data-l10n-id="rtl_reading_direction_label">Right to left</span>
+            </button>
+
+            <div class="horizontalToolbarSeparator"></div>
```
Snuffleupagus, line 221:
```suggestion
            <div class="horizontalToolbarSeparator readingDirectionButtons"></div>
```

web/viewer.js
```diff
@@ -112,6 +112,8 @@ function getViewerConfiguration() {
       spreadNoneButton: document.getElementById("spreadNone"),
       spreadOddButton: document.getElementById("spreadOdd"),
       spreadEvenButton: document.getElementById("spreadEven"),
+      ltrReadingDirectionButton: document.getElementById("ltrReadingDirection"),
+      rtlReadingDirectionButton: document.getElementById("rtlReadingDirection"),
```
Snuffleupagus, line 116: Similar to other comments, please re-order the names here.

web/secondary_toolbar.js
```diff
@@ -129,6 +134,18 @@ class SecondaryToolbar {
         eventDetails: { mode: SpreadMode.EVEN },
         close: true,
       },
+      {
+        element: options.ltrReadingDirectionButton,
+        eventName: "changedirection",
```
Snuffleupagus, line 139:
```suggestion
      {
        element: options.readingDirectionLtrButton,
        eventName: "switchreadingdirection",
```
(With *similar* changes below as well.)

web/viewer.css
```diff
@@ -1899,3 +1907,10 @@ html[dir="rtl"] #documentPropertiesOverlay .row > * {
     display: none;
   }
 }
+
+.dir-rtl {
+  direction: rtl;
+}
+.dir-ltr {
+  direction: ltr;
+}
```
Snuffleupagus, line 1916: These rules need to go into the https://github.com/mozilla/pdf.js/blob/master/web/pdf_viewer.js file instead.

web/secondary_toolbar.js
```diff
@@ -300,6 +318,20 @@ class SecondaryToolbar {
...
+      buttons.ltrReadingDirectionButton.classList.toggle(
+        "toggled",
+        direction === ReadingDirection.LTR
+      );
+      buttons.rtlReadingDirectionButton.classList.toggle(
+        "toggled",
+        direction === ReadingDirection.RTL
+      );
+    }
+    this.eventBus._on("directionchanged", readingDirectionChanged);
```
Snuffleupagus, line 332:
```suggestion
    this.eventBus._on("readingdirectionchanged", readingDirectionChanged);
```

web/secondary_toolbar.js
```diff
@@ -157,6 +174,7 @@ class SecondaryToolbar {
     this._bindCursorToolsListener(options);
     this._bindScrollModeListener(options);
     this._bindSpreadModeListener(options);
+    this._bindReadingDirectionListener(options);
```
Snuffleupagus, line 177: Please note that you also need add
```css
#secondaryToolbarButtonContainer.hiddenReadingDirectionButtons > .readingDirectionButtons
```
to https://github.com/mozilla/pdf.js/blob/master/web/viewer.css#L638 and then add the appropriate lines in the "baseviewerinit" event listener a few lines below here.

web/base_viewer.js
```diff
@@ -1476,6 +1478,35 @@ class BaseViewer {
...
+    this.container.classList.toggle(
+      "dir-ltr",
+      ReadingDirection.LTR === direction
+    );
+    this.container.classList.toggle(
+      "dir-rtl",
+      ReadingDirection.RTL === direction
+    );
+    this.eventBus.dispatch("directionchanged", { source: this, direction });
+  }
```
Snuffleupagus, line 1509: Please change this to
```suggestion
    this._readingDirection = direction;
    this.eventBus.dispatch("readingdirectionchanged", { source: this, direction });
    
    this._updateReadingDirection();
  }
```
by removing the `this.container.classList.toggle(...)` lines from here and instead adding them to a new `_updateReadingDirection` helper method placed just below instead.
Then, in https://github.com/mozilla/pdf.js/blob/master/web/pdf_single_page_viewer.js, please add an *empty* `_updateReadingDirection() {}` line at the end of the file.

web/base_viewer.js
```diff
@@ -1476,6 +1478,35 @@ class BaseViewer {
...
+   * @type {number} One of the values in {ReadingDirection}.
+   */
+  get direction() {
+    return this._direction;
+  }
+
+  /**
+   * @param {number} dir
+   */
+  set direction(direction) {
```
Snuffleupagus, line 1492:
```suggestion
  set readingDirection(direction) {
```

web/viewer.css
```diff
@@ -90,6 +90,7 @@
   --secondaryToolbarButton-spreadNone-icon: url(images/secondaryToolbarButton-spreadNone.svg);
   --secondaryToolbarButton-spreadOdd-icon: url(images/secondaryToolbarButton-spreadOdd.svg);
   --secondaryToolbarButton-spreadEven-icon: url(images/secondaryToolbarButton-spreadEven.svg);
+  --secondaryToolbarButton-toggleReadingDirection-icon: url(images/secondaryToolbarButton-toggleReadingDirection.svg);
```
Snuffleupagus, line 93: Please change the filenames/references according to:
```suggestion
  --secondaryToolbarButton-readingDirectionToggle-icon: url(images/secondaryToolbarButton-readingDirectionToggle.svg);
```

Snuffleupagus (changes requested):
I've done an initial pass over the code, and based on that I unfortunately have to request a number of changes; sorry about the number of comments here!

Most of my comments are regarding consistent *naming* of properties/variables (given pre-existing conventions in the code), but there's also some additional validation that needs to happen in various places.

web/base_viewer.js
```diff
@@ -1476,6 +1478,35 @@ class BaseViewer {
...
+    if (this._direction === direction) {
+      return;
+    }
+    if (!Object.values(ReadingDirection).includes(direction)) {
+      throw new Error(`Invalid document direction: ${direction}`);
+    }
```
Snuffleupagus, line 1498: The validation needs to be done using the `isValidReadingDirection` helper function (see other review comment), and I don't *think* that we need to support setting this until the document has loaded (since that simplifies things overall):
```suggestion
  set readingDirection(direction) {
    if (!isValidReadingDirection(direction)) {
      throw new Error(`Invalid reading direction: ${direction}`);
    }
    if (!this.pdfDocument) {
      return;
    }
    if (this._readingDirection === direction) {
      return; // The reading direction didn't change.
    }
```

Snuffleupagus (changes requested):

web/viewer.html
Snuffleupagus: Also, this button should have the `toggled` CSS class set as well.

Snuffleupagus (changes requested):
Please ensure that the state of the secondaryToolbar is initialized correctly, and the viewer reset as intended when changing documents, something like the following diff might work (although untested):
```diff
diff --git a/web/app.js b/web/app.js
index 9c4f4e24c..dbec5c3ec 100644
--- a/web/app.js
+++ b/web/app.js
@@ -539,7 +539,8 @@ const PDFViewerApplication = {
     this.secondaryToolbar = new SecondaryToolbar(
       appConfig.secondaryToolbar,
       container,
-      eventBus
+      eventBus,
+      this.l10n
     );
 
     if (this.supportsFullscreen) {
diff --git a/web/base_viewer.js b/web/base_viewer.js
index 24579f4aa..991494beb 100644
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -661,6 +661,8 @@ class BaseViewer {
     this.viewer.textContent = "";
     // ... and reset the Scroll mode CSS class(es) afterwards.
     this._updateScrollMode();
+
+    this._updateReadingDirection();
   }
 
   _scrollUpdate() {
diff --git a/web/secondary_toolbar.js b/web/secondary_toolbar.js
index 777d0205f..456867692 100644
--- a/web/secondary_toolbar.js
+++ b/web/secondary_toolbar.js
@@ -14,6 +14,7 @@
  */
 
 import {
+  NullL10n,
   ReadingDirection,
   SCROLLBAR_PADDING,
   ScrollMode,
@@ -59,8 +60,9 @@ class SecondaryToolbar {
    * @param {SecondaryToolbarOptions} options
    * @param {HTMLDivElement} mainContainer
    * @param {EventBus} eventBus
+   * @param {IL10n} l10n - Localization service.
    */
-  constructor(options, mainContainer, eventBus) {
+  constructor(options, mainContainer, eventBus, l10n = NullL10n) {
     this.toolbar = options.toolbar;
     this.toggleButton = options.toggleButton;
     this.toolbarButtonContainer = options.toolbarButtonContainer;
@@ -161,6 +163,7 @@ class SecondaryToolbar {
 
     this.mainContainer = mainContainer;
     this.eventBus = eventBus;
+    this.l10n = l10n;
 
     this.opened = false;
     this.containerHeight = null;
@@ -330,6 +333,17 @@ class SecondaryToolbar {
       );
     }
     this.eventBus._on("directionchanged", readingDirectionChanged);
+
+    this.eventBus._on("secondarytoolbarreset", evt => {
+      if (evt.source === this) {
+        this.l10n.getDirection().then(dir => {
+          readingDirectionChanged({
+            direction:
+              dir === "rtl" ? ReadingDirection.RTL : ReadingDirection.LTR,
+          });
+        });
+      }
+    });
   }
 
   open() {
```

timvandermeij:
Closing since this PR is 2.5 years old now, quite outdated and not a feature that's currently planned. Thanks.

PR #12439, opened 2020-10-03
Add `version`/`build` info at the top of the *built* `web/viewer.js` file

Snuffleupagus:
This should be helpful to easily determine the *exact* version of the viewer itself, when looking at a *built* `web/viewer.js` file.
Note that we're already including this information in other built files, such as e.g. `pdf.js`, `pdf.worker.js`, `pdf_viewer.js`, and `pdf.image_decoders.js`.

### DIFF
--- a/web/viewer.js
+++ b/web/viewer.js
@@ -15,6 +15,13 @@
 
 "use strict";
 
+/* eslint-disable-next-line no-unused-vars */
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

PR #12441, opened 2020-10-03
Re-factor how the ESLint `no-var` rule is enabled in the `src/` folder

Snuffleupagus:
This simplifies/consolidates the ESLint configuration slightly in the `src/` folder, and prevents the addition of any new files where `var` is being used.[1]
Hence we no longer need to manually add `/* eslint no-var: error */` in files, which is easy to forget, and can instead disable the rule in the `src/core/` files where `var` is still in use.

---
[1] Obviously the `no-var` rule can, in the same way as every other rule, be disabled on a case-by-case basis where actually necessary.


timvandermeij (approved):

timvandermeij:
Good idea; thanks!

PR #12442, opened 2020-10-03
Upgrade Puppeteer to version 5.3.1

### LINKED ISSUE #11998: Remove extraction log suppression code

timvandermeij:
Fixes #11998.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1838,9 +1838,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "14.0.13",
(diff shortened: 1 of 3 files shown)


Snuffleupagus (approved):
Nice that we can finally do this now; thanks for looking into this again!

Snuffleupagus:
It seems that this also reduced the overall test run-times, on *both* bots, which is really nice :-)

PR #12445, opened 2020-10-04
Upgrade minor/major versions of dependencies that don't require code changes

timvandermeij:


package.json
```diff
@@ -10,12 +10,12 @@
...
     "@babel/runtime": "^7.11.2",
     "acorn": "^7.4.0",
-    "autoprefixer": "^9.8.6",
+    "autoprefixer": "^10.0.1",
     "babel-loader": "^8.1.0",
     "canvas": "^2.6.1",
     "core-js": "^3.6.5",
     "cross-env": "^7.0.2",
-    "es-module-shims": "^0.4.7",
+    "es-module-shims": "^0.6.0",
```
Snuffleupagus, line 18: Given that this dependency is only used for development, and isn't covered by the test-suite, please confirm that the following things work locally (after running `gulp server` of course):
 - The viewer itself, and the default tracemonkey document, loads successfully when opened with http://localhost:8888/web/viewer.html 
 - The unit-tests completes successfully (with the possible exception of one Worker-related test failing), when run with http://localhost:8888/test/unit/unit_test.html

Snuffleupagus (approved):
r=me, with the comment addressed and all tests passing of course :-)

Thanks for updating all of these dependencies!

package.json
timvandermeij: I have verified that this indeed works as described. Thank you for reviewing this, and the other patch!

Snuffleupagus:
> Now only `acorn`, `systemjs`, `terser` and `yargs` are not the latest versions because they require more work.

Given that the intention is to simply remove `systemjs` altogether, which is unfortunately still blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1247687, we can probably just ignore that part (especially since the upgrade is non-trivial).

For the other ones, *especially* `acorn` which is an integral part of the build system, we should try to upgrade as soon as possible to e.g. have better support for modern ECMAScript features.

Would you happen to have *some* information about what the upstream changes are that prevents upgrading the `acorn`, `terser`, and `yargs` versions?

timvandermeij:
Yes, for `acorn` we require setting the `ecmaVersion` property explicitly when calling `acorn.parse()` for instance. This should be easy, but I didn't manage to get it working at a first try, so I left it out of this PR for now.

`terser` made the `minify` call asynchronous, which mostly means some rewriting, most likely also relatively easy to do.

`yargs` mainly changed the validation functions for the argument parser. I looked into this briefly before, but didn't get it working easily.

In summary, those three should be doable for us, it's just that I haven't had the time before to really look into them.

PR #12448, opened 2020-10-05
Enable the ESLint `no-debugger` and `no-alert` rules

Snuffleupagus:
The `debugger`-statement would only, potentially, make sense during development and we thus want to prevent it from being accidentally included when landing code.
The `alert`, `confirm`, and `prompt` functions should generally be avoided, with the few intended cases manually allowed.

Please find additional details about the ESLint rules at:
 - https://eslint.org/docs/rules/no-debugger
 - https://eslint.org/docs/rules/no-alert


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12454, opened 2020-10-07
Upgrade `terser` to version 5

Snuffleupagus:
The only significant change is that the `minify` command is now asynchronous, which we can handle easily by simply making the containing functions `async`.
Based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Browser_compatibility, using `async`/`await` in the gulpfile should no longer be an issue as far as I can tell.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -916,7 +916,7 @@ gulp.task(
   })
 );
 
-function parseMinified(dir) {
+async function parseMinified(dir) {
   var pdfFile = fs.readFileSync(dir + "/build/pdf.js").toString();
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Nice, another one done!

Snuffleupagus:
Thanks for landing this, and the other one too!

I'm not going to attempt the `yargs` upgrade though, since compared to the others I cannot *immediately* see what changes are necessary and because this dependency isn't part of the build process (and thus slightly less important IMHO).

Issue #8473, opened 2017-06-01
Colors shift and elements disappear


mawmawmawm:
Hi all,
PDF.js renders print files wrongly, elements disappear or have the wrong color.

Configuration:
- Web browser and its version: Chrome 58
- Operating system and its version: macOS 10.12 and Windows 7
- PDF.js version: 1.4.20 as well as 1.8.408
- Is an extension: No

Steps to reproduce the problem:
1. Load PDF file and compare against Adobe Acrobat or macOS PDF viewer

File "stepped":
![2017-06-01 at 14 42](https://cloud.githubusercontent.com/assets/9080951/26702291/4067449c-46d9-11e7-94ad-4b1b3d5d2ec5.png)
PDF.js left, Acrobat right

File "proof":
![2017-06-01 at 14 02](https://cloud.githubusercontent.com/assets/9080951/26702310/54f06e5c-46d9-11e7-9018-34099a4e11df.png)
PDF.js left, Acrobat right

Original files:
[stepped.pdf](https://github.com/mozilla/pdf.js/files/1046131/stepped.pdf)
[proof.pdf](https://github.com/mozilla/pdf.js/files/1046130/proof.pdf)

vikdet:
Hi, I want to add some examples:  
Result:
![image007](https://user-images.githubusercontent.com/3327188/95074803-d25ec300-0717-11eb-9fe6-1d5ba75d1ed0.png)
Expected:
![image004](https://user-images.githubusercontent.com/3327188/95074849-e276a280-0717-11eb-89a9-7fbf7d59c72c.png)
[Webinar Hub.pdf](https://github.com/mozilla/pdf.js/files/5327057/Webinar.Hub.pdf)

Issue #2765, opened 2013-02-21
TilingType


waddlesplash:
http://pubs.usgs.gov/sim/3067/pdf/sim3067sheet-2.pdf does not have black lines where it should in pdf.js. This is probably because of the warning in the console:

> Warning: TODO: TilingType: 3

... that gets printed 125 times in a row.

Since enough other PDFs want this, perhaps it's finally time to attempt a _complete_ implementation?

Rob--W:
Is this bug report still relevant? I don't see any issues with the provided PDFs and descriptions.

timvandermeij:
I'm going to keep only the performance label here as both PDFs are not very performant (especially the first one). The image issues appear to be resolved as far as I can tell.

PR #12458, opened 2020-10-08
Add local caching of TilingPatterns in `PartialEvaluator.getOperatorList` (issue 2765 and 8473)

Snuffleupagus:
In practice it's not uncommon for PDF documents to re-use the same TilingPatterns more than once, and parsing them is essentially equal to parsing of a (small) page since a `getOperatorList` call is required.

By caching the internal TilingPattern representation we can thus avoid having to re-parse the same data over and over, and there's also *less* asynchronous parsing required for repeated TilingPatterns.

Initially I had intended to include (standard) benchmark results with this patch, however it's not entirely clear that this is actually necessary here given the preliminary results.
When testing this manually in the development viewer, using `pdfBug=Stats`, the following (approximate) reduction in rendering times were observed when comparing `master` against this patch:
 - http://pubs.usgs.gov/sim/3067/pdf/sim3067sheet-2.pdf (from issue #2765): `6800 ms` -> `4100 ms`.
 - https://github.com/mozilla/pdf.js/files/1046131/stepped.pdf (from issue #8473): `54000 ms` -> `13000 ms`
 - https://github.com/mozilla/pdf.js/files/1046130/proof.pdf (from issue #8473): `5900 ms` -> `2500 ms`

As always, whenever you're dealing with documents which are "slow", there's usually a certain level of subjectivity involved with regards to what's deemed acceptable performance.
Hence it's not clear to me that we want to regard any of the referenced issues as fixed, however the improvements are significant enough to warrant caching of TilingPatterns in my opinion.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -81,6 +81,7 @@ import {
   LocalColorSpaceCache,
   LocalGStateCache,
   LocalImageCache,
+  LocalTilingPatternCache,
 } from "./image_utils.js";
 import { bidi } from "./bidi.js";
 import { ColorSpace } from "./colorspace.js";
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Nice work! I'd also say they are not completely fixed yet, but it's _much_ better now.

/botio makeref

PR #12464, opened 2020-10-09
Fix getVisibleElements helper in RTL-locales

### LINKED ISSUE #12453: [Regression] The `getVisibleElements` helper function is broken in RTL-locales
Attach (recommended) or Link to PDF file here: *Not applicable*

Configuration:
- Web browser and its version: All
- Operating system and its version: All
- PDF.js version: All versions since `2.0.943`, including the current `master`.
- Is a browser extension: *Not applicable*

1. Open http://localhost:8888/web/viewer.html#locale=he (or another RTL-locale)
2. Choose "Horizontal Scrolling"-mode in the Secondary Toolbar
3. Scroll to the end of the document.

All pages, including the last one, should render successfully.

All pages, except the first ones, are completely blank.

---

The bug is, as far as I can tell, caused by the [`getVisibleElements` helper function](https://github.com/mozilla/pdf.js/blob/fd1d9cc85f631ee2c871c79cb10cc511c1e8d8d4/web/ui_utils.js#L414-L502) not handling the viewer being used in RTL-locales (i.e. when `dir="rtl"` is set on the containing HTML-element).

While this very clearly effects the "Horizontal Scrolling"-mode, it's also possible that both the "Wrapped Scrolling"-mode and the "Odd/Even Spread"-modes are also affected (but to a lesser extent) in RLT-locales.

---

I'd tentatively suggest that we label this as a regression, since it appears to have been broken in RTL-locales ever since PR #9208 which originally added the "Horizontal Scrolling"-mode (and others).

baloone:
This PR fixes #12453. I'm not sure about the name of the auxiliary function `isElementNextAfterViewHorizontally`.

### DIFF
--- a/test/unit/ui_utils_spec.js
+++ b/test/unit/ui_utils_spec.js
@@ -639,12 +639,12 @@ describe("ui_utils", function () {
     // This function takes a fixed layout of pages and compares the system under
     // test to the slower implementation above, for a range of scroll viewport
     // sizes and positions.
-    function scrollOverDocument(pages, horizontally = false) {
+    function scrollOverDocument(pages, horizontally = false, rtl = false) {
       const size = pages.reduce(function (max, { div }) {
         return Math.max(
           max,
           horizontally
-            ? div.offsetLeft + div.clientLeft + div.clientWidth
+            ? Math.abs(div.offsetLeft + div.clientLeft + div.clientWidth)
             : div.offsetTop + div.clientTop + div.clientHeight
         );
       }, 0);
@@ -652,7 +652,7 @@ describe("ui_utils", function () {
       // make scrollOverDocument tests faster, decrease them to make the tests
       // more scrupulous, and keep them coprime to reduce the chance of missing
       // weird edge case bugs.
-      for (let i = 0; i < size; i += 7) {
+      for (let i = -size; i < size; i += 7) {
         // The screen height (or width) here (j - i) doubles on each inner loop
         // iteration; again, this is just to test an interesting range of cases
         // without slowing the tests down to check every possible case.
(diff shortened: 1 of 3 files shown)


timvandermeij:
@Snuffleupagus If you have time, could you take a look at this PR? It looks reasonable to me, but my experience with the (not so trivial) `getVisibleElements` function is limited.

timvandermeij (commented):
I just noticed that we have unit tests for the `getVisibleElements` function in `test/unit/ui_utils_spec.js`. I don't know how difficult this it to test, but it would be great if you could extend the unit tests for this PR so that we prevent this from regressing again in the future.

Snuffleupagus (changes requested):
As mentioned in https://github.com/mozilla/pdf.js/pull/12464#pullrequestreview-506177403, we need at least a *basic* unit-test for this (called e.g. `handles horizontal scrolling with RTL-documents correctly`).
I'd suggest defining a suitable `scrollEl`, similar but obviously not identical to https://github.com/mozilla/pdf.js/blob/49791f55b3cb4ae1cef68a6573e764aa0db3b5fb/test/unit/ui_utils_spec.js#L740-L746 with pages defined e.g. as in https://github.com/mozilla/pdf.js/blob/49791f55b3cb4ae1cef68a6573e764aa0db3b5fb/test/unit/ui_utils_spec.js#L730-L736 and with the test doing something like
```javascript
const visibleSorted = getVisibleElements(
  scrollEl,
  views,
  /* sortByVisibility = */ true,
  /* horizontal = */ true,
  /* rtl = */ true
);
```
and then checking that `visibleSorted` has the expected order (see how that's implemented in the existing test).

baloone:
I added the two suggested changes and a test

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #12470, opened 2020-10-11
Upgrade `webpack` to version 5

Snuffleupagus:
The only noticeable changes are that the built files are now *slightly* smaller, and that Webpack now supports optional chaining and nullish coalescing without the need for Babel plugins.


timvandermeij (approved):

timvandermeij:
Nice, slightly smaller files and fewer direct dependencies. Thanks!

PR #12478, opened 2020-10-15
Convert `PartialEvaluator.translateFont` to an `async` method

Snuffleupagus:
This allows us to make a slight simplification in `PartialEvaluator.loadFont`, which thus removes an old TODO-comment from the method.
Furthermore, in `PartialEvaluator.translateFont`, the CMap-handling is now limited to only *composite* fonts to avoid having to wait for a "dummy"-Promise for most fonts.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1111,15 +1111,7 @@ class PartialEvaluator {
 
     font.translated = fontCapability.promise;
 
-    // TODO move promises into translate font
-    var translatedPromise;
-    try {
-      translatedPromise = this.translateFont(preEvaluatedFont);
-    } catch (e) {
-      translatedPromise = Promise.reject(e);
-    }
-
-    translatedPromise
+    this.translateFont(preEvaluatedFont)
       .then(translatedFont => {
         if (translatedFont.fontType !== undefined) {
           var xrefFontStats = xref.stats.fontTypes;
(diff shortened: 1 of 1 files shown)


src/core/evaluator.js
```diff
@@ -3465,36 +3457,30 @@ class PartialEvaluator {
...
+      newProperties => {
         this.extractWidths(dict, descriptor, newProperties);
 
         if (type === "Type3") {
           newProperties.isType3Font = true;
         }
         return new Font(fontName.name, fontFile, newProperties);
-      });
+      }
+    );
```
SamyCookie, line 3483: Should this be better like this ?
```javascript
    const newProperties = await this.extractDataStructures(dict, baseDict, properties);
    this.extractWidths(dict, descriptor, newProperties);

    if (type === "Type3") {
      newProperties.isType3Font = true;
    }
    return new Font(fontName.name, fontFile, newProperties);
```

Snuffleupagus: Well, I don't think that what you're suggesting is necessarily *better*, but rather only *different*.
Given that it'd only serve to increase the total size of the diff, since there's more than one such "block", I'm not convinced that it's a necessary change.

SamyCookie: You're right, it's not necessary but seems to me more readable in only one block.
It was just a suggestion, you can skip my comment ;)

timvandermeij (approved):

timvandermeij:
This is indeed a bit simpler. Thanks!

PR #12479, opened 2020-10-15
Fix the "should get form info when AcroForm is present" unit-test

Snuffleupagus:
The last unit-test didn't work correctly, since an error was thrown in `PDFDocument._hasOnlyDocumentSignatures` because the mocked `XRef`-instance wasn't actually being set correctly.

Also, updates the `XRefMock` to use `async` methods where appropriate.


timvandermeij (approved):

timvandermeij:
Oh wow, nice catch!

Issue #12475, opened 2020-10-13
Some urls behind pushbutton are not extracted from the AA dictionary 


calixteman:
Steps to reproduce the problem:
1. Open https://www.guidedesimpots.lu/wp-content/uploads/2019/02/160F-2019.pdf
2. The links behind FD, FO, DS, ... are not usable

These annotations correspond to a PushButton because their parent is and the action is in the AA dictionary associated to D (MouseDown).
So once https://github.com/mozilla/pdf.js/pull/12429 is merged we can easily extract url from the action.
Since for security reasons we won't implement `app.launchURL` we must handle that stuff in same way as we do for A dict.

calixteman:
The issue in #9776 is slightly different since the actions in `AA::D` are `/GoTo` and here they're `/JavaScript`.
But finally thanks to your comment, I think I've a way better (and simpler) patch to handle the different cases.

timvandermeij:
Fixed by the PR above.

Issue #9776, opened 2018-06-04
Links not working


pchiriot:
Attach (recommended) or Link to PDF file here: https://drive.google.com/file/d/1W9SybAqqKg2MDkgT1yp6hnmNGLrquj6r/view?usp=sharing

Configuration:
- Web browser and its version: Firefox 60.0.1
- Operating system and its version: Window 10
- PDF.js version:  
       PDFJS.version = '1.8.188';
       PDFJS.build = 'ad1023ff';

- Is a browser extension: No

Steps to reproduce the problem:
1.  Open the PDF with PDFJS
2. Click on a link (page 4 or 5)

What is the expected behavior? (add screenshot)
You should have move to the correct page. Open the PDF on adobe acrobat reader to see the expected behavior

What went wrong? (add screenshot)
Most of the link are not clickable. If they are, you are redirected to the page page 1

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

Thanks in advance.

timvandermeij:
I looked into this and found that the links are not link annotations, but actually button widget annotations. The console shows the following warnings:

```
Warning: Push buttons without action dictionaries are not supported
Warning: parseDestDictionary: Unsupported Action type "Hide".
```

timvandermeij:
I debugged this a bit and found that the annotations don't have the `A` dictionary, which is why we stop parsing. However, they do have the `AA` (additional actions) dictionary, which we don't support but does contain the actions.

timvandermeij:
Fixed by the PR above.

PR #12481, opened 2020-10-15
Get urls if any in AA::D dictionary for pushbuttons

calixteman:
This patch aims to fix isssue #12475 and issue #9776.


timvandermeij (approved):

timvandermeij:
Nice work fixing two issues at once!

PR #12485, opened 2020-10-16
[api-minor] JS - Add a function in api to get the fields ids in AcroForm::CO

calixteman:
Each time a Field is validate then some `Calculate` actions may be ran just after.
In order to know in which order these actions are ran we must get this information from AcroForm::CO.
See the top of page 365 (and bottom of 364 of course):
https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=365

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -37,6 +37,7 @@ import {
   Dict,
   isDict,
   isName,
+  isRef,
   isStream,
   Ref,
 } from "./primitives.js";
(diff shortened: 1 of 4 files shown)


src/display/api.js
```diff
@@ -884,6 +884,15 @@ class PDFDocumentProxy {
   getFieldObjects() {
     return this._transport.getFieldObjects();
   }
+
+  /**
```
timvandermeij, line 888: Let's replace this block with the following block that contains more information about the `null` case, which makes it consistent with the recently changed https://github.com/mozilla/pdf.js/pull/12483/files#diff-082d6b37ad01db7ac97cc07c6ddb0dc52040484c5ef91b110b072f50144d9f39R881-R883:

```javascript
  /**
   * @returns {Promise<Array<string> | null>} A promise that is resolved with an
   *   {Array<string>} containing IDs of annotations that have a calculation
   *   action, or `null` when no such annotations are present in the PDF file.
   */
```

src/core/document.js
```diff
@@ -1005,6 +1006,21 @@ class PDFDocument {
       Promise.all(allPromises).then(() => allFields)
     );
   }
+
+  get calculationOrderIds() {
```
timvandermeij, line 1010: This getter needs a unit test on `document_spec.js`. I think it can be very similar to https://github.com/mozilla/pdf.js/blob/master/test/unit/document_spec.js#L108-L118 where we provide an `AcroForm` dictionary with first no valid `CO` entry to test the `null` case and then with a valid `CO` entry to assert that the IDs are returned properly.

(Thinking about this, the `fieldObjects` getter should have gotten a similar unit test. If that can be added in a separate PR that would be great.)

timvandermeij (commented):
Looks good with these two comments addressed. Thank you for making this PR small so it's easier to oversee and we can iterate faster.

timvandermeij (approved):

timvandermeij:
Looks good!

PR #7347, opened 2016-05-23
Slightly refactor the `fontRef` handling in `PartialEvaluator_loadFont` (issue 7403 and issue 7402)

Snuffleupagus:
Originally, I was just going to change this code to use `Ref_toString` in a couple more places. When I started reading the code, I figured that it wouldn't hurt to clean up a couple of comments. While doing this, I noticed that the logic for the (rare) `isDict(fontRef)` case could do with a few improvements.

There should be no functional changes with this patch, but given the added reference checks, we will now avoid bogus `Ref`s when resolving font aliases. ~~In practice, I don't think the current code has caused issues, but it seems good to clean this up to prevent any future problems.~~ _As issue #7403 shows, there're issues with the current code._

Note that the only thing that this patch will change, is the `font.loadedName` in the case where a `fontRef` is a reference _and_ the font doesn't have a descriptor. Previously for `fontRef = Ref(4, 0)` we'd get `font.loadedName = 'g_d0_f4_0'`, and with this patch `font.loadedName = g_d0_f4R`, which is actually one character shorted in most cases. (Given that `Ref_toString` contains an optimization for the `gen === 0` case, which is by far the most common `gen` value.)

In the already existing fallback case, where the `fontName` is used to when creating the `font.loadedName`, we allow any alphanumeric character. Hence I don't see how (as mentioned above) e.g. `font.loadedName = g_d0_f4R` would be an issue here.

**Edit:** This also fixes #7403 and #7402.

PR #12486, opened 2020-10-16
Stop caching the *parsed* Font data on its `Dict` object (PR 7347 follow-up)

Snuffleupagus:
Given that *all* fonts are, ever since PR #7347, now cached in the "normal" `fontCache` there's actually no reason for the special `font.translated` construction. (Given how Objects in JavaScript are references, rather than raw values, the old code shouldn't have caused any significant memory overhead.)

Instead we can simply store the `cacheKey`, which is a simple string, on only the Font `Dict`s where it's needed and thus look-up all fonts using the `fontCache` instead.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -979,15 +979,13 @@ class PartialEvaluator {
   }
 
   loadFont(fontName, font, resources) {
-    const errorFont = () => {
-      return Promise.resolve(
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

Issue #12431, opened 2020-10-01
Pdf.js v2.5.207 and v2.4.456 doesn't not render on IOS Safari 14.0.1 (Renders white or with errors).


Brendon-F:
Pdf.js v2.5.207 stable prebuilt (ES5-compatible) version does not render pdf's on iOS safari 14.0.1. The pages are either white or render the following error.

Have now tested 2.4.456-es5-dist with same issues on iOS safari 14.0.1.

Error message:
<img width="693" alt="Screenshot 2020-10-01 at 13 14 38" src="https://user-images.githubusercontent.com/72203963/94807852-22701980-03e8-11eb-81ed-f908fd6c46b1.png">

White page:
<img width="699" alt="Screenshot 2020-10-01 at 17 17 18" src="https://user-images.githubusercontent.com/72203963/94835751-febdcb00-0409-11eb-8261-8213a379f124.png">

The link to a test file can be found here for version 2.4.456:

https://www.designoriginal.co.uk/test/web/viewer.html?file=../pdfs/PRISMA-Core-Datasheet.pdf
https://www.designoriginal.co.uk/test/web/viewer.html?file=../pdfs/VSPP-Datasheet.pdf

To replicate the issue use an apple device with the latest iOS version 14.0.1

Brendon Fear

astehlik:
I'm not sure if this is the cause but some Safari or Samsung Internet versions seems to have a problem with the [Nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) `??` and [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) `?.`

It seems that those are not transpiled by babel currently, see for example:

https://github.com/mozilla/pdf.js/blob/gh-pages/web/viewer.js#L1142

Snuffleupagus:
> The link to a test file can be found here for version 2.4.456:
> 
> https://www.designoriginal.co.uk/test/web/viewer.html?file=../pdfs/PRISMA-Core-Datasheet.pdf
> https://www.designoriginal.co.uk/test/web/viewer.html?file=../pdfs/VSPP-Datasheet.pdf

[...]

PR #12488, opened 2020-10-16
Add back babel plugins to fix browser compatibility

astehlik:
The babel plugins

plugin-proposal-nullish-coalescing-operator
plugin-proposal-optional-chaining

are added back to the build process to fix problems in various
browsers.

Related issue: #12431

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -217,6 +217,18 @@ function createWebpackConfig(defines, output) {
           options: {
             presets: skipBabel ? undefined : ["@babel/preset-env"],
             plugins: [
+              [
+                "@babel/plugin-proposal-nullish-coalescing-operator",
+                {
+                  loose: true,
+                },
+              ],
+              [
+                "@babel/plugin-proposal-optional-chaining",
+                {
+                  loose: true,
+                },
+              ],
               "@babel/plugin-transform-modules-commonjs",
               [
                 "@babel/plugin-transform-runtime",
--- a/package-lock.json
+++ b/package-lock.json
@@ -890,9 +890,9 @@
       }
     },
     "@babel/plugin-proposal-nullish-coalescing-operator": {
-      "version": "7.10.4",
-      "resolved": "https://registry.npmjs.org/@babel/plugin-proposal-nullish-coalescing-operator/-/plugin-proposal-nullish-coalescing-operator-7.10.4.tgz",
-      "integrity": "sha512-wq5n1M3ZUlHl9sqT2ok1T2/MTt6AXE0e1Lz4WzWBr95LsAZ5qDXe4KnFuauYyEyLiohvXFMdbsOTMyLZs91Zlw==",
+      "version": "7.12.1",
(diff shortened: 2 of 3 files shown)


astehlik:
This reverts parts of pull request #12470

Snuffleupagus (changes requested):
Sorry, but unfortunately we cannot accept this PR.

For older browsers, you need to use an `es5`-compatible build; see e.g. the "Older browsers" section of https://github.com/mozilla/pdf.js#online-demo or the `-es5-dist.zip` builds at https://github.com/mozilla/pdf.js/releases (as outlined at https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support).

brendandahl:
Someone on matrix was saying the es5 build is broken on IE11, maybe this is related?

astehlik:
Thank you for the quick reply.

I would not consider the latest version of Samsung Internet an "older" browser. It's based on Chromium and crashes with both feature proposals.

When you look at the documentation

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

you can see that also Firefox Mobile does not support them yet :(

I hope you reconsider this decision.

Snuffleupagus:
> Someone on matrix was saying the es5 build is broken on IE11, maybe this is related?

That should be completely unrelated here; please note https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support (emphasis mine):
> While the library and the default viewer should still work in IE 11/Edge (non-Chromium-based) for the time being, some functionality/features may not be available and the performance will be worse compared to modern browsers. ***However, this support will be dropped after stable version `2.6.347`.***

---

> I hope you reconsider this decision.

Why though, since we purposely provide `es5`-compatible builds to address these kind of situations!?

astehlik:
> Why though, since we purposely provide `es5`-compatible builds to address these kind of situations!?

As far as I understand the es5 version has a quite significant worse performance than the modern build.

And since those two syntaxes are the only thing that is causing problems in these browsers it might be worth the slightly larger build size.

Snuffleupagus:
> As far as I understand the es5 version has a quite significant worse performance than the modern build.

Huh, that's actually the first time that I can remember seeing such claims...

> And since those two syntaxes are the only thing that is causing problems in these browsers

As of right now, it may perhaps only be those two things. However we want to be able to utilize modern EcmaScript features, as they become available, and that could thus change (quickly) in the future; hence the reason for offering transpiled/non-transpiled builds.

(Furthermore, we want to avoid unnecessarily transpiling of code intended for the Firefox PDF Viewer.)

astehlik:
> Huh, that's actually the first time that I can remember seeing such claims...

OK, maybe my memory is wrong ;)

Well, thank you very much for taking you time to explain it. An thank you for all the great work!

Have a nice weekend :)

PR #12489, opened 2020-10-17
Add a test for pdfDocument::fieldObjects

calixteman:


test/unit/document_spec.js
```diff
@@ -174,5 +188,61 @@ describe("document", function () {
...
       expect(pdfDocument.calculationOrderIds).toEqual(null);
     });
+
+    it("should get field objects array or null", function (done) {
+      const acroForm = new Dict();
+
+      let pdfDocument = getDocument(acroForm);
+      pdfDocument.fieldObjects.then(fields => {
+        expect(fields).toEqual(null);
+        done();
```
timvandermeij, line 198: I don't think this (and the one below) is correct. If Jasmine reaches `done`, the `it` scope is considered completed, so if that happens soon the rest of the test it not actually run.

I think this can be avoided by simply using `async`/`await`; see https://jasmine.github.io/tutorials/async#asyncawait. Something like this should then work so you can remove `done` completely like in the other tests:

```javascript
const fields = await pdfDocument.fieldObjects;
expect(fields).toEqual(null);

[remaining tests in the same way]
```

timvandermeij (commented):
Looks good with this comment addressed.

timvandermeij (approved):

timvandermeij:
Thank you for the follow-up!

PR #12493, opened 2020-10-18
Support adding pages, in addition to regular destinations, to the browser history and use it with thumbnails (issue 12440)

### LINKED ISSUE #12440: Back action navigates to wrong page after thumbnail is clicked

Configuration:
- Web browser and its version: Chrome Version 87.0.4280.6 (Official Build) canary (64-bit)
- Operating system and its version: Windows 10 Home 10.0.0.18363
- PDF.js version: 2.6.12
- Is a browser extension: Yes

1. Open https://www.ijcai.org/Proceedings/2019/0241.pdf
2. Click on Sidebar icon
3. Click on e.g. last page's thumbnail
4. Click on Back arrow

The browser should navigate to 1st page of the document

The browser navigates to the page from which the document was opened (e.g., the Home page)

Note: Intrestingly enough, clicking on the Outline works correctly.

Snuffleupagus:
While the referenced issue could very well be seen as an edge-case, this patch adds support for updating of the browser history when interacting with the thumbnails in the sidebar (assuming we want to do this).

The main reason for adding the history implementation in the first place, was to simplify navigating back to a previous position in the document when named/explicit destinations are used (e.g. when clicking on "links" or when using the outline in the sidebar).
As such, it never really crossed by mind to update the browser history when the thumbnails are used. However, a user clicking on thumbnails could be regarded as a pretty strong indication of user-intent w.r.t. navigation in the document, hence I suppose that updating the browser history in this particular case probably won't hurt.

Fixes #12440

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -344,7 +344,7 @@ class LinkAnnotationElement extends AnnotationElement {
     link.href = this.linkService.getDestinationHash(destination);
     link.onclick = () => {
       if (destination) {
-        this.linkService.navigateTo(destination);
+        this.linkService.goToDestination(destination);
       }
       return false;
     };
--- a/web/interfaces.js
+++ b/web/interfaces.js
@@ -54,9 +54,9 @@ class IPDFLinkService {
   set externalLinkEnabled(value) {}
 
   /**
-   * @param dest - The PDF destination object.
+   * @param {string|Array} dest - The named, or explicit, PDF destination.
    */
-  navigateTo(dest) {}
+  async goToDestination(dest) {}
 
   /**
    * @param dest - The PDF destination object.
--- a/web/interfaces.js
+++ b/web/interfaces.js
@@ -59,6 +59,11 @@ class IPDFLinkService {
   async goToDestination(dest) {}
 
   /**
+   * @param {number} pageNumber - The page number.
+   */
+  goToPage(pageNumber) {}
+
+  /**
    * @param dest - The PDF destination object.
    * @returns {string} The hyperlink to the PDF object.
    */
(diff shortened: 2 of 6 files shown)


timvandermeij:
Wouldn't we then want to do this in more situations? Clicking a thumbnail changes the page number, so then I'd expect this to also work when entering a page number in the textbox or via the hash parameters, and also when clicking an outline item. I'd say those are all signs of navigation within the document. What do you think?

Snuffleupagus:
> so then I'd expect this to also work when entering a page number in the textbox

That's somewhat more difficult to implement well, given that we support page *labels* in addition to just page numbers; hence that's not something I want to worry about here (or even at all).

> via the hash parameters,

But, that's been supported from the start[1]...

> and also when clicking an outline item.

... and so has this, since those are just destinations, given that it's the *primary* reason for `PDFHistory` existing in the first place :-)

---
[1] To the extent that doing so is feasible, since the old `PDFHistory` implementation showed that you want to steer well clear of trying to navigate back-and-forth between previous history entries in order to edit them.

timvandermeij:
> But, that's been supported from the start

You're completely right; now that I tried it clearly works. Not sure how I tried it yesterday, but sorry for the noise.

The patch seems to work just fine, so let's do this. Thanks!

timvandermeij (approved):

Snuffleupagus:
> Not sure how I tried it yesterday, but sorry for the noise.

It's *very* possible that your initial testing actually uncovered a bug, since all of this is unfortunately pretty complicated[1]. If so, please let me know if you manage to come up with a way to reproduce it.

No worries; thanks for taking the time to test this and think about the various cases involved.

---
[1] Unless you simply found an edge-case, which is purposely not implemented because doing so would complicate and/or break the primary functionality of `PDFHistory`; it'd still be good to know though!

PR #12503, opened 2020-10-21
[api-minor] Invalidate an annotation with no quadPoints (when it's required)

calixteman:
Some pdf softwares don't remove highlight annotations but make the QuadPoints array empty.
And the Rect for the annotation can be [-32768, -32768, 32768, 32768] so it leads to have a giant div which catches all the mouse events and make the pdf unusable when there are some forms elements.
I aims to fix issue #12506.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -194,7 +194,11 @@ function getQuadPoints(dict, rect) {
   // The region is described as a number of quadrilaterals.
   // Each quadrilateral must consist of eight coordinates.
   const quadPoints = dict.getArray("QuadPoints");
-  if (!Array.isArray(quadPoints) || quadPoints.length % 8 > 0) {
+  if (
+    !Array.isArray(quadPoints) ||
+    quadPoints.length === 0 ||
+    quadPoints.length % 8 > 0
+  ) {
     return null;
   }
 
@@ -312,6 +316,9 @@ class Annotation {
    * @type {boolean}
    */
   get viewable() {
+    if (this.data.quadPoints === null) {
+      return false;
+    }
     if (this.flags === 0) {
       return true;
     }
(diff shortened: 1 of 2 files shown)


src/core/annotation.js
```diff
@@ -2245,6 +2250,10 @@ class HighlightAnnotation extends MarkupAnnotation {
           },
         });
       }
+    } else {
+      // QuadPoints is required
+      this._isInvalid = true;
```
Snuffleupagus, line 2255: This name feels very "generic", while it's however only used together with QuadPoints.

Couldn't we simply set `this.data.quadPoints = null;` explicitly here, and then add a `this.data.quadPoints !== null` check to the `_isViewable`/`_isPrintable` helpers instead?

Snuffleupagus (commented):
Is there a PDF file that can be used to (manually) test this patch?

Also, please provide slightly more complete *commit messages*, such that it's easier to understand what a patch does (and why) e.g. when looking at it using the `git` command line.
There's a bunch more information in https://github.com/mozilla/pdf.js/pull/12503#issue-507413628, so it'd be nice if that was included in the *commit message* too.

src/core/annotation.js
calixteman: So for any reason, I can't attach a file here so I filed an issue #12506.

timvandermeij:
I don't think this will always work. According to https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=403&zoom=auto,-215,524 the `QuadPoints` array is optional, so while I'm fine with discarding it if it's empty we should then fall back to the `Rect` according to the specification. How does Adobe Reader handle this particular PDF file?

In any case, if there are no quadpoints this patch will hide the popup, which is not correct because quadpoints are optional and the popup works fine for annotations that are not rendered with quadpoints but just with the rectangle.

timvandermeij (approved):

timvandermeij:
Looks good!

Issue #12504, opened 2020-10-21
Highlight annotation div can be on top of a textfield


calixteman:
Attach (recommended) or Link to PDF file here:
[plop.pdf](https://github.com/mozilla/pdf.js/files/5415411/plop.pdf)

Here the highlight has two rectangles and only `Annotation::Rect` is used when creating the `div` in the annotation layer and so the field is almost unusable.

timvandermeij:
Fixed the PR above.

PR #12505, opened 2020-10-21
[api-minor] Split highlight annotation div into multiple divs

calixteman:
This patch aims to fix issue #12504.
Instead of using `Annotation::Rect` for the div dimensions, we use several divs where dimensions are in `QuadPoints` and then set mouse handlers on each of them.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -240,6 +240,35 @@ class AnnotationElement {
   }
 
   /**
+   * Create quadrilaterals for the quadPoints.
+   *
+   * @private
+   * @param {boolean} ignoreBorder
+   * @memberof AnnotationElement
+   * @returns {HTMLSectionElement}
+   */
+  _createQuadrilaterals(ignoreBorder = false) {
+    if (!this.data.quadPoints) {
+      return null;
+    }
+
+    const quadrilaterals = [];
+    const savedRect = this.data.rect;
+    for (const quadPoint of this.data.quadPoints) {
+      const rect = [
+        quadPoint[2].x,
+        quadPoint[2].y,
+        quadPoint[1].x,
+        quadPoint[1].y,
+      ];
+      this.data.rect = rect;
+      quadrilaterals.push(this._createContainer(ignoreBorder));
+    }
+    this.data.rect = savedRect;
+    return quadrilaterals;
+  }
+
+  /**
    * Create a popup for the annotation's HTML element. This is used for
    * annotations that do not have a Popup entry in the dictionary, but
    * are of a type that works with popups (such as Highlight annotations).
@@ -789,14 +818,14 @@ class PopupAnnotationElement extends AnnotationElement {
     }
 
     const selector = `[data-annotation-id="${this.data.parentId}"]`;
-    const parentElement = this.layer.querySelector(selector);
-    if (!parentElement) {
+    const parentElements = this.layer.querySelectorAll(selector);
+    if (parentElements.length === 0) {
       return this.container;
     }
 
     const popup = new PopupElement({
       container: this.container,
-      trigger: parentElement,
+      trigger: Array.from(parentElements),
       color: this.data.color,
       title: this.data.title,
       modificationDate: this.data.modificationDate,
@@ -805,12 +834,18 @@ class PopupAnnotationElement extends AnnotationElement {
 
     // Position the popup next to the parent annotation's container.
     // PDF viewers ignore a popup annotation's rectangle.
-    const parentTop = parseFloat(parentElement.style.top),
-      parentLeft = parseFloat(parentElement.style.left),
-      parentWidth = parseFloat(parentElement.style.width);
-    const popupLeft = parentLeft + parentWidth;
+    const page = this.page;
+    const rect = Util.normalizeRect([
+      this.data.parentRect[0],
+      page.view[3] - this.data.parentRect[1] + page.view[1],
+      this.data.parentRect[2],
+      page.view[3] - this.data.parentRect[3] + page.view[1],
+    ]);
+    const popupLeft =
+      rect[0] + this.data.parentRect[2] - this.data.parentRect[0];
+    const popupTop = rect[1];
 
-    this.container.style.transformOrigin = `${-popupLeft}px ${-parentTop}px`;
+    this.container.style.transformOrigin = `${-popupLeft}px ${-popupTop}px`;
     this.container.style.left = `${popupLeft}px`;
 
     this.container.appendChild(popup.render());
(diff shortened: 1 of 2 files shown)


src/display/annotation_layer.js
```diff
@@ -239,6 +239,38 @@ class AnnotationElement {
     return container;
   }
 
+  /**
+   * Create areas for the quadPoints
```
Snuffleupagus, line 243: Nit: Here, and elsewhere below, please remember to end sentences with a period.

src/display/annotation_layer.js
```diff
@@ -239,6 +239,38 @@ class AnnotationElement {
...
+   *
+   * @private
+   * @memberof AnnotationElement
+   */
+  _createAreas() {
```
Snuffleupagus, line 248: Should this not take a `ignoreBorder` parameter, similar to the `createContainer` method?

src/display/annotation_layer.js
```diff
@@ -239,6 +239,38 @@ class AnnotationElement {
...
+      let first = true;
+      for (const quadPoint of this.data.quadPoints) {
+        const rect = [
+          quadPoint[2].x,
+          quadPoint[2].y,
+          quadPoint[1].x,
+          quadPoint[1].y,
+        ];
+        this.data.rect = rect;
+        const area = this._createContainer(true);
```
Snuffleupagus, line 261: See the comment above, regarding `ignoreBorder`.

src/display/annotation_layer.js
```diff
@@ -239,6 +239,38 @@ class AnnotationElement {
...
   }
 
+  /**
+   * Create areas for the quadPoints
+   *
+   * @private
+   * @memberof AnnotationElement
+   */
+  _createAreas() {
+    if (this.data.quadPoints) {
```
Snuffleupagus, line 249: You can reduce the overall indentation by using an early return instead here, i.e.
```javascript
if (!this.data.quadPoints) {
  return;
}
// The rest of the method goes here...
```

src/display/annotation_layer.js
```diff
@@ -758,28 +790,38 @@ class PopupAnnotationElement extends AnnotationElement {
...
-      trigger: parentElement,
+      trigger: Array.from(parentElement),
       color: this.data.color,
       title: this.data.title,
       modificationDate: this.data.modificationDate,
       contents: this.data.contents,
     });
 
+    // Find the main container if any
+    let main;
```
Snuffleupagus, line 808: Any reason to not just initialize this here, i.e. `let main = parentElement.item(0);`, since there's no validation below the loop (but rather only a null-check)?

src/display/annotation_layer.js
```diff
@@ -1550,13 +1606,16 @@ class AnnotationLayer {
...
       );
-      if (element) {
-        element.style.transform = `matrix(${parameters.viewport.transform.join(
-          ","
-        )})`;
+      if (elements) {
+        elements.forEach(element => {
+          const transform = `matrix(${parameters.viewport.transform.join(
+            ","
+          )})`;
```
Snuffleupagus, line 1616: It looks like the `transform`-string is constant for all elements, hence please move it out out the loop.

Snuffleupagus (commented):
I've added a couple of quick comments; please also add more context to the *commit message* here as well :-)

src/display/annotation_layer.js
```diff
@@ -239,6 +239,40 @@ class AnnotationElement {
     return container;
   }
 
+  /**
+   * Create areas for the quadPoints.
+   *
+   * @private
```
timvandermeij, line 245: After this line add: `@param {boolean} ignoreBorder` (just like `_createContainer`).

src/display/annotation_layer.js
```diff
@@ -239,6 +239,40 @@ class AnnotationElement {
...
+   * @memberof AnnotationElement
+   */
+  _createAreas(ignoreBorder = false) {
```
timvandermeij, line 248: `_createQuadrilaterals` (that's what the specification calls the "areas"), so let's rename all "area" occurrences to "quadrilateral".

src/display/annotation_layer.js
```diff
@@ -239,6 +239,40 @@ class AnnotationElement {
     return container;
   }
 
+  /**
+   * Create areas for the quadPoints.
```
timvandermeij, line 243: `Create quadrilaterals from the quadpoints.`

src/display/annotation_layer.js
```diff
@@ -758,28 +792,37 @@ class PopupAnnotationElement extends AnnotationElement {
     }
 
     const selector = `[data-annotation-id="${this.data.parentId}"]`;
-    const parentElement = this.layer.querySelector(selector);
-    if (!parentElement) {
+    const parentElement = this.layer.querySelectorAll(selector);
+    if (!parentElement || parentElement.length === 0) {
```
timvandermeij, line 796: If it's a list now, we should call it `parentElements` (plural) for clarity. Moreover, `querySelectorAll` always returns a list, so the `!parentElement` case can no longer occur.

src/display/annotation_layer.js
```diff
@@ -758,28 +792,37 @@ class PopupAnnotationElement extends AnnotationElement {
...
-      trigger: parentElement,
+      trigger: Array.from(parentElement),
       color: this.data.color,
       title: this.data.title,
       modificationDate: this.data.modificationDate,
       contents: this.data.contents,
     });
 
+    // Find the main container if any
+    let main = parentElement.item(0);
```
timvandermeij, line 810: I don't see why we need this and the `first` logic above. The main container should always be `this.container` and you can access it everywhere. Moreover, we should not be storing a non-default property like `mainContainer` on a standard element.

src/display/annotation_layer.js
```diff
@@ -239,6 +239,40 @@ class AnnotationElement {
     return container;
   }
 
+  /**
+   * Create areas for the quadPoints.
+   *
+   * @private
+   * @memberof AnnotationElement
```
timvandermeij, line 246: `_createContainer` returns the container, with `@returns {HTMLSectionElement}` added here, but this function set the member variable. For consistency it would be better to also let this method return a list of quadrilaterals, or `null` if not applicable.

timvandermeij (commented):
Looks good with these comments addressed. This is most likely related to #6811; note that it applies not only to highlights, but to all markup annotations and link annotations. That can be done in a follow-up though.

src/display/annotation_layer.js
calixteman: Popup was positioned according the highlight container but now the highlights are splited. So my goal was to position the Popup using the parent rect finally.
So instead of doing that, I added parent rect in `Popup.data` (in `annotation.js`).

brendandahl (approved):

timvandermeij:
Unfortunately the reference tests above show a regression for `firefox-annotation-highlight-without-appearance-page1`. I'll open a new issue for this.

timvandermeij:
Aside from the things mentioned in follow-up issue #12576, this patch does look exactly like what I would expect code-wise and works fine for the original issue and most other files, so thank you for that!

PR #12508, opened 2020-10-21
Fallback font for buttons must be ZapfDingbats.

calixteman:
Fix bug https://bugzilla.mozilla.org/show_bug.cgi?id=1669099.


Snuffleupagus (commented):
This seems like the kind of patch where you probably want to add an `eq` test as well, in addition to any unit-tests, since those are much more difficult to accidentally regress with any future changes.

src/core/evaluator.js
```diff
@@ -978,7 +986,7 @@ class PartialEvaluator {
     });
   }
 
-  loadFont(fontName, font, resources) {
+  loadFont(fontName, font, resources, isButtonContext) {
```
timvandermeij, line 989: I'm not really a big fan of this annotation detail leaking through to the generic evaluator. Would it perhaps be slightly better to let the caller (the annotation code) provide a `fallbackFont` dictionary explicitly and pass that instead of this boolean? By default this parameter can be `null` to indicate that we want to use the existing `fallbackFontDict`, but if provided it allows the caller to override the default fallback font.

What do you think? /cc @Snuffleupagus to check if this is a good idea or not.

Snuffleupagus: > I'm not a big fan of this annotation detail leaking through to the generic evaluator.

Yeah, I completely agree with this sentiment; it just feels all-around weird to special-case the `PartialEvaluator` like that.

> Would it perhaps be better to let the caller (the annotation code) provide a `fallbackFont` dictionary parameter explicitly and pass that instead of this boolean?

That's a really nice idea, and one that I completely support!

calixteman: Is it really worth creating an object for such a corner case ? pr maybe we could then have a static fallback.

timvandermeij: Well, it's a more generic solution. The current solution is very specific to this single case, but puts annotation-specific details in the generic evaluator. Moving the fallback font dictionary itself to the annotation code, to which it belongs because it's specific to annotations, moves the edge case to there and ensures that the generic evaluator also gets a generic parameter, namely the fallback font dictionary to use. If it's not provided (by default), the existing fallback font dictionary is simply used.

src/core/evaluator.js
```diff
@@ -777,7 +777,15 @@ class PartialEvaluator {
...
 
-  handleSetFont(resources, fontArgs, fontRef, operatorList, task, state) {
+  handleSetFont(
+    resources,
+    fontArgs,
+    fontRef,
+    operatorList,
+    task,
+    state,
+    fallbackFontDict
```
Snuffleupagus, line 787: Please indicate that the parameter is optional here, i.e.
```suggestion
    fallbackFontDict = null
```

src/core/evaluator.js
```diff
@@ -978,7 +986,7 @@ class PartialEvaluator {
     });
   }
 
-  loadFont(fontName, font, resources) {
+  loadFont(fontName, font, resources, fallbackFontDict) {
```
Snuffleupagus, line 989: Please indicate that the parameter is optional here, i.e.
```suggestion
  loadFont(fontName, font, resources, fallbackFontDict = null) {
```

timvandermeij (approved):

timvandermeij:
Looks good; thanks!

/botio-linux makeref

PR #12515, opened 2020-10-22
Enable the `unicorn/no-abusive-eslint-disable` ESLint plugin rule

Snuffleupagus:
This rule prevents *accidental* disabling of ESLint completely, be ensuring that inline ESLint disable comments always list the affected rules explicitly.

Please see https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-abusive-eslint-disable.md for additional information.


timvandermeij (approved):

timvandermeij:
Useful addition; thanks!

PR #12516, opened 2020-10-22
Prevent issues, in `PDFDocument.fieldObjects`, for invalid Annotations

Snuffleupagus:
For an invalid Annotation, there's one code-path where `undefined` is returned from `AnnotationFactory._create`. That'd currently, incorrectly, trigger an error during the `PDFDocument._collectFieldObjects` parsing which thus seem good to avoid.
Along these lines, the filtering in `PDFDocument.fieldObjects` is also updated to handle both `null` and `undefined` the same way.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -969,7 +969,7 @@ class PDFDocument {
         this.pdfManager,
         this._localIdFactory
       )
-        .then(annotation => annotation.getFieldObject())
+        .then(annotation => annotation && annotation.getFieldObject())
         .catch(function (reason) {
           warn(`_collectFieldObjects: "${reason}".`);
           return null;
@@ -999,7 +999,7 @@ class PDFDocument {
     for (const [name, promises] of fieldPromises) {
       allPromises.push(
         Promise.all(promises).then(fields => {
-          fields = fields.filter(field => field !== null);
+          fields = fields.filter(field => !!field);
           if (fields.length > 0) {
             allFields[name] = fields;
           }


calixteman:
Good catch.

timvandermeij:
Are you sure this can actually happen? `undefined` is returned at https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L81, which can only happen if the `ref` in https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L79 is not a dictionary. However, _if_ that were the case, it would break even before https://github.com/mozilla/pdf.js/pull/12516/files#diff-f0385d138acfe65f04aec809ad0522722883e56b6420ccb8f1d2e16b50631c1eR972, namely at https://github.com/mozilla/pdf.js/pull/12516/files#diff-f0385d138acfe65f04aec809ad0522722883e56b6420ccb8f1d2e16b50631c1eR952-R953 as far as I can tell.

That's because the `fieldRef` we give to `AnnotationLayer.create` is the same `fieldRef` that we use just above to also fetch the `T` property, so if it's not a dictionary we wouldn't be able to call `.has("T")` on it.

Snuffleupagus:
> Are you sure this can actually happen?

It's more of a better safe than sorry sort of patch, since I think that it first of all makes *general* sense and secondly who knows what future code changes may bring. Good point though!

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12519, opened 2020-10-23
Test the windows bot.

brendandahl:


brendandahl:
It seems some sub-processes of firefox keep crashing, but unfortunately there's no crash reports

brendandahl:
The crash is https://bugzilla.mozilla.org/show_bug.cgi?id=1673194

Snuffleupagus:
> I see `firefox-issue12233-print-page1` failed

Is it actually a regression though, since it rather *looks* like an improvement unless I'm completely misunderstanding this.

> did a regression land or did we not update ref images?

I'd say the latter, since a number of PRs have landed while the Windows-bot was down; hopefully Tim can confirm this.

timvandermeij:
This reference image is indeed an improvement from #12395; before the image was incorrect, so a makeref is safe here.

timvandermeij:
I verified that the Windows bot is working properly again now. Thanks!

/botio makeref

PR #12522, opened 2020-10-23
Fix a couple of edge-cases in `PDFViewerApplication._initializeJavaScript` (PR 12432 follow-up)

Snuffleupagus:
- Return early in `PDFViewerApplication._initializeJavaScript` for PDF documents without any `fieldObjects`, which is the vast majority of all documents, to prevent errors when trying to parse a non-existent object.

 - Similar to the other `PDFViewerApplication._initialize*` methods, ignore the `fieldObjects` if the document was closed before the data resolved.

 - Fix the JSDoc comment for the `generateRandomStringForSandbox` helper function, since there's currently a bit too much copy-and-paste going on :-)

 - Change `FirefoxScripting` to a class with static methods, which is consistent with the surrounding code in `web/firefoxcom.js`.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1345,10 +1345,14 @@ const PDFViewerApplication = {
    * @private
    */
   async _initializeJavaScript(pdfDocument) {
-    if (!AppOptions.get("enableScripting")) {
+    const objects = await pdfDocument.getFieldObjects();
+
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:

PR #12525, opened 2020-10-24
[api-minor] Implement API to get MarkInfo from the catalog.

brendandahl:
Trying to get more data so we can decide whether to implement tagged pdfs.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -152,6 +152,47 @@ class Catalog {
     return shadow(this, "metadata", metadata);
   }
 
+  get markInfo() {
+    let markInfo = null;
+    try {
+      markInfo = this._readMarkInfo();
+    } catch (ex) {
+      if (ex instanceof MissingDataException) {
+        throw ex;
+      }
+      warn("Unable to read mark info.");
+    }
+    return shadow(this, "markInfo", markInfo);
+  }
+
+  /**
+   * @private
+   */
+  _readMarkInfo() {
+    const obj = this._catDict.get("MarkInfo");
+    if (!isDict(obj)) {
+      return null;
+    }
+
+    const markInfo = Object.assign(Object.create(null), {
+      Marked: false,
+      UserProperties: false,
+      Suspects: false,
+    });
+    for (const key in markInfo) {
+      if (!obj.has(key)) {
+        continue;
+      }
+      const value = obj.get(key);
+      if (!isBool(value)) {
+        continue;
+      }
+      markInfo[key] = value;
+    }
+
+    return markInfo;
+  }
+
   get toplevelPagesDict() {
     const pagesObj = this._catDict.get("Pages");
     if (!isDict(pagesObj)) {
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -806,6 +806,23 @@ class PDFDocumentProxy {
   }
 
   /**
+   * @typedef {Object} MarkInfo
+   * Properties correspond to Table 321 of the PDF 32000-1:2008 spec.
+   * @property {boolean} Marked
+   * @property {boolean} UserProperties
+   * @property {boolean} Suspects
+   */
+
+  /**
+   * @returns {Promise<MarkInfo | null>} A promise that is resolved with
+   *   a {MarkInfo} object that contains the MarkInfo flags for the PDF
+   *   document, or `null` when no MarkInfo values are present in the PDF file.
+   */
+  getMarkInfo() {
+    return this._transport.getMarkInfo();
+  }
+
+  /**
    * @returns {Promise<TypedArray>} A promise that is resolved with a
    *   {TypedArray} that has the raw data from the PDF.
    */
@@ -2644,6 +2661,10 @@ class WorkerTransport {
             : null,
         };
       });
+  }
+
+  getMarkInfo() {
+    return this.messageHandler.sendWithPromise("GetMarkInfo", null);
   }
 
   getStats() {
--- a/web/app.js
+++ b/web/app.js
@@ -242,6 +242,7 @@ const PDFViewerApplication = {
   triggerDelayedFallback: null,
   _saveInProgress: false,
   _wheelUnusedTicks: 0,
+  _idleCallbacks: new Set(),
 
   // Called once when the document is loaded.
   async initialize(appConfig) {
@@ -738,6 +739,10 @@ const PDFViewerApplication = {
     this.contentDispositionFilename = null;
     this.triggerDelayedFallback = null;
     this._saveInProgress = false;
+    for (const callback of this._idleCallbacks) {
+      window.cancelIdleCallback(callback);
+    }
+    this._idleCallbacks.clear();
 
     this.pdfSidebar.reset();
     this.pdfOutlineViewer.reset();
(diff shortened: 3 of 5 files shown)


src/core/obj.js
```diff
@@ -152,6 +152,47 @@ class Catalog {
...
+    const markInfo = Object.create({
+      Marked: null,
+      UserProperties: null,
+      Suspects: null,
+    });
+    for (const key in markInfo) {
+      if (!obj.has(key)) {
+        continue;
+      }
+      const prop = obj.get(key);
```
timvandermeij, line 186: I think `value` would be a better name given that we use `key` just above.

src/display/api.js
```diff
@@ -805,6 +805,24 @@ class PDFDocumentProxy {
...
+   * Properties correspond to Table 321 of the PDF 32000-1:2008 spec. Values
+   * are null if not defined by the PDF.
+   * @property {boolean | null} Marked
+   * @property {boolean | null} UserProperties
+   * @property {boolean | null} Suspects
+   */
+
+  /**
+   * @returns {Promise<MarkInfo | null>} A promise that is resolved with
+   *   a {MarkInfo} that contains the MarkInfo flags for the PDF document, or
```
timvandermeij, line 819: Shouldn't this be `a {MarkInfo} object that...` here?

src/display/api.js
```diff
@@ -805,6 +805,24 @@ class PDFDocumentProxy {
     return this._transport.getMetadata();
   }
 
+  /**
+   * @typedef {Object} MarkInfo
+   * Properties correspond to Table 321 of the PDF 32000-1:2008 spec. Values
+   * are null if not defined by the PDF.
+   * @property {boolean | null} Marked
```
timvandermeij, line 812: I looked at https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=564&zoom=auto,-215,328 and all those fields should have the default value `false`, so I think we should implement that as well. Moreover, then we can always have a boolean here, which also seems like a better API and more logical when looking at the unit test. Finally, I think it's also necessary for https://github.com/mozilla/pdf.js/pull/12525/files#diff-e24e4bd32090b0c400a70f02d1b844df95c9bbe62cd0b280f4397e9f4194fe44R1352 to work properly, since we'd now be reporting `null` if there is `MarkInfo` without a `Marked` flag.

web/app.js
```diff
@@ -1329,12 +1329,33 @@ const PDFViewerApplication = {
...
 
+  /**
+   * A place to fetch data for telemetry after one page is rendered and the
+   * viewer is idle.
+   */
+  async _collectTelemetry() {
+    if (!this.pdfDocument) {
+      return;
+    }
+    let markInfo = await this.pdfDocument.getMarkInfo();
```
timvandermeij, line 1351: This variable and the next one can be `const` (I believe this is also why the CI failed).

timvandermeij (commented):
Looks good with these comments addressed. I think the `requestIdleCallback` usage is quite a nice solution here!

src/display/api.js
```diff
@@ -805,6 +805,24 @@ class PDFDocumentProxy {
     return this._transport.getMetadata();
   }
 
+  /**
+   * @typedef {Object} MarkInfo
+   * Properties correspond to Table 321 of the PDF 32000-1:2008 spec. Values
+   * are null if not defined by the PDF.
```
timvandermeij, line 811: `null` -> `false`

timvandermeij (approved):
Looks good and can be merged with the final comment addressed!

timvandermeij:
`TEST-UNEXPECTED-FAIL | gets markInfo | in firefox | Expected undefined to equal false. Expected undefined to equal false. `

Snuffleupagus:
> Ah, it seems that `Object.create` creates an empty object instead of the prefilled one (tried in the web console).

The argument is used as the *prototype*, which doesn't survive the structured cloning algorithm.
If you actually want a `null` principal, which I assume is indeed the case here, then the following is probably the way to go:
```javascript
const markInfo = Object.assign(Object.create(null), {
  Marked: false,
  UserProperties: false,
  Suspects: false,
});
```

> Not sure why Travis does like it. Maybe it does work in Node.js?

Probably because those tests run with workers *disabled*.

web/app.js
```diff
@@ -1329,12 +1329,33 @@ const PDFViewerApplication = {
...
+      }
     });
 
     this._initializePageLabels(pdfDocument);
     this._initializeMetadata(pdfDocument);
   },
 
+  /**
+   * A place to fetch data for telemetry after one page is rendered and the
+   * viewer is idle.
```
Snuffleupagus, line 1345: Nit: Add a `* @private` line here.

web/app.js
```diff
@@ -1329,12 +1329,33 @@ const PDFViewerApplication = {
       pdfViewer.optionalContentConfigPromise.then(optionalContentConfig => {
         this.pdfLayerViewer.render({ optionalContentConfig, pdfDocument });
       });
+      if ("requestIdleCallback" in window) {
+        window.requestIdleCallback(() => {
+          this._collectTelemetry();
+        });
```
Snuffleupagus, line 1335: Also, I think that you need to track this callback and ensure that it's cancelled when/if the document is closed, to avoid the possibility of more than one of these running at any one time.

Basically something like this (assuming that you add an `_idleCallbacks: new Set(),` after line https://github.com/mozilla/pdf.js/blob/master/web/app.js#L249):
```suggestion
        const callback = window.requestIdleCallback(() => {
          this._collectTelemetry(pdfDocument);
          
          this._idleCallback.delete(callback);
        });
        this._idleCallbacks.add(callback);
```

Then, e.g. after line https://github.com/mozilla/pdf.js/blob/master/web/app.js#L745 add something like
```javascript
    for (const callback of this._idleCallbacks) {
      window.cancelIdleCallback(callback);
    }
    this._idleCallbacks.clear();
```

web/app.js
```diff
@@ -1329,12 +1329,33 @@ const PDFViewerApplication = {
...
 
+  /**
+   * A place to fetch data for telemetry after one page is rendered and the
+   * viewer is idle.
+   */
+  async _collectTelemetry() {
+    if (!this.pdfDocument) {
+      return;
+    }
+    const markInfo = await this.pdfDocument.getMarkInfo();
```
Snuffleupagus, line 1351: I'd suggest the following here, which is consistent with the various `this._initializeX` methods that exists below:
```suggestion
  async _collectTelemetry(pdfDocument) {
    const markInfo = await pdfDocument.getMarkInfo();
    
    if (pdfDocument !== this.pdfDocument) {
      return; // The document was closed while the MarkInfo data resolved.
    }
```

web/app.js
```diff
@@ -1329,12 +1329,33 @@ const PDFViewerApplication = {
...
+    const tagged = markInfo ? markInfo.Marked : false;
```
Snuffleupagus, line 1352: Nit: If you want to, you can now use optional chaining here[1] instead:
```suggestion
    const tagged = markInfo?.Marked || false;
```

---
[1] With the *exception* of the `src/core/` and `src/shared/` folders, you should now be free to use both optional changing and nullish coalescing where it makes sense.

Snuffleupagus (changes requested):

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12527, opened 2020-10-25
Remove SystemJS usage from the development viewer and the unit-tests

Snuffleupagus:

It's highly recommended to review this *one* commit at a time, perhaps also ignoring whitespace-only changes where appropriate.
The second commit is probably more-or-less impossible to actually review, so we're probably going to have to trust our test-suite and that I wrote proper regular expressions.

This still leaves SystemJS usage in the font-tests, however fixing that will require larger changes and it thus seemed best to defer that to later since this PR is already very large.


gulpfile.js
```diff
@@ -185,6 +185,19 @@ function createWebpackConfig(defines, output) {
...
+  // `src/core/{glyphlist,unicode}.js` (Babel is too slow for those when
+  // source-maps are enabled) should be excluded from processing.
+  const babelExcludes = [
+    "node_modules[\\\\\\/]core-js",
+    "node_modules[\\\\\\/]web-streams-polyfill",
+  ];
+  if (enableSourceMaps) {
+    babelExcludes.push("src[\\\\\\/]core[\\\\\\/](glyphlist|unicode)");
+  }
+  const babelExcludeRegExp = new RegExp("(" + babelExcludes.join("|") + ")");
```
timvandermeij, line 199: I think we can use a template string here for slightly improved readability.

gulpfile.js
```diff
@@ -185,6 +185,19 @@ function createWebpackConfig(defines, output) {
     !bundleDefines.TESTING;
   var skipBabel = bundleDefines.SKIP_BABEL;
 
+  // `core-js` (see https://github.com/zloirock/core-js/issues/514),
+  // `web-streams-polyfill` (already using a transpiled file), and
+  // `src/core/{glyphlist,unicode}.js` (Babel is too slow for those when
+  // source-maps are enabled) should be excluded from processing.
+  const babelExcludes = [
+    "node_modules[\\\\\\/]core-js",
```
timvandermeij, line 193: I'm going to assume that, like the second commit, the number of backslashes here is correct, although it looks a bit much :-)

test/unit/jasmine-boot.js
```diff
@@ -197,26 +199,26 @@ function initializePDFJS(callback) {
...
-  // Replace the browser window's `onload`, ensure it's called, and then run
-  // all of the loaded specs. This includes initializing the `HtmlReporter`
-  // instance and then executing the loaded Jasmine environment.
-  var currentWindowOnload = window.onload;
-
-  window.onload = function () {
-    if (currentWindowOnload) {
-      currentWindowOnload();
+  function extend(destination, source) {
+    for (var property in source) {
```
timvandermeij, line 203: `var` -> `const` now that my PR just landed (which created a merge conflict that needs to be addressed; sorry for that!)

timvandermeij (approved):
Looks good with these comments addressed. Nice to see yet another part of SystemJS go!

gulpfile.js
Snuffleupagus: Sure, why not.

test/unit/jasmine-boot.js
Snuffleupagus: No worries!

gulpfile.js
Snuffleupagus: For `gulp mozcentral` the resulting RegExp looks as follows, which is consistent with the old format:
```
/(node_modules[\\\/]core-js|node_modules[\\\/]web-streams-polyfill)/
```
For `gulp generic` the resulting RegExp looks as follows, which is also consistent with the old format:
```
/(node_modules[\\\/]core-js|node_modules[\\\/]web-streams-polyfill|src[\\\/]core[\\\/](glyphlist|unicode))/
```

Since we now build the regular expressing using "normal" strings, we need to do more escaping to not lose part of it.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #12528, opened 2020-10-25
Convert `var` to `const`/`let` in the `test/unit` folder 

timvandermeij:
The commit messages contain more information about the individual commits.

The `test/unit` folder counts for about 50% of the `var` usage in the entire `test` folder and used a mix of `var` and `const`/`let`. This commit series makes the entire `test/unit` folder use `const`/`let` consistently and allows us to enforce that unit tests are written with modern syntax, bringing us one step closer to enabling the ESLint `no-var` rule globally.

### DIFF
--- a/test/unit/.eslintrc
+++ b/test/unit/.eslintrc
@@ -1,0 +1,10 @@
+{
+  "extends": [
+    "../.eslintrc"
+  ],
+
+  "rules": {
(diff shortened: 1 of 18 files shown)


Snuffleupagus (approved):
Looks good to me, thanks for doing this!

PR #12529, opened 2020-10-25
Tweak the `pdf.scripting.js` bundling, to improve overall consistency

Snuffleupagus:
This brings the new `pdf.scripting.js` bundling more in-line with the pre-existing handling for the  `pdf.js`/`pdf.worker.js` files:
 - Add a new `src/pdf.scripting.js` file as the entry-point for the build scripts.

 - Add the version/build numbers at the top of the *built* `pdf.scripting.js` files, since all other built files include that information given that it's often helpful to be able to easily determine the *exact* version.

 - Tweak the `createScriptingBundle` in the gulp-file, since it looks like a little bit too much copy-and-paste in the variable names.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -317,20 +317,20 @@ function createMainBundle(defines) {
 }
 
 function createScriptingBundle(defines) {
-  var mainAMDName = "pdfjs-dist/build/pdf.scripting";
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12530, opened 2020-10-26
JS -- Add 'util' object

calixteman:
This patch provides an implementation of the util object as described:
 * https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=716

### DIFF
--- a/src/scripting_api/util.js
+++ b/src/scripting_api/util.js
@@ -20,10 +20,561 @@ class Util extends PDFObject {
     super(data);
 
     this._crackURL = data.crackURL;
+    this._scandCache = new Map();
+    this._months = [
+      "January",
+      "February",
+      "March",
+      "April",
+      "May",
+      "June",
+      "July",
+      "August",
+      "September",
+      "October",
+      "November",
+      "December",
+    ];
+    this._days = [
+      "Sunday",
+      "Monday",
+      "Tuesday",
+      "Wednesday",
+      "Thursday",
+      "Friday",
+      "Saturday",
+    ];
+    this.MILLISECONDS_IN_DAY = 86400000;
+    this.MILLISECONDS_IN_WEEK = 604800000;
   }
 
   crackURL(cURL) {
     return this._crackURL(cURL);
+  }
+
+  printf(...args) {
+    if (args.length === 0) {
+      throw new Error("Invalid number of params in printf");
+    }
+
+    if (typeof args[0] !== "string") {
+      throw new TypeError("First argument of printf must be a string");
+    }
+
+    const pattern = /%(,[0-4])?([\+ 0#]+)?([0-9]+)?(\.[0-9]+)?(.)/g;
+    const PLUS = 1;
+    const SPACE = 2;
+    const ZERO = 4;
+    const HASH = 8;
+    let i = 0;
+    return args[0].replace(pattern, function (
+      match,
+      nDecSep,
+      cFlags,
+      nWidth,
+      nPrecision,
+      cConvChar
+    ) {
+      // cConvChar must be one of d, f, s, x
+      if (
+        cConvChar !== "d" &&
+        cConvChar !== "f" &&
+        cConvChar !== "s" &&
+        cConvChar !== "x"
+      ) {
+        const buf = ["%"];
+        for (const str of [nDecSep, cFlags, nWidth, nPrecision, cConvChar]) {
+          if (str) {
+            buf.push(str);
+          }
+        }
+        return buf.join("");
+      }
+
+      i++;
+      if (i === args.length) {
+        throw new Error("Not enough arguments in printf");
+      }
+      const arg = args[i];
+
+      if (cConvChar === "s") {
+        return arg.toString();
+      }
+
+      let flags = 0;
+      if (cFlags) {
+        for (const flag of cFlags) {
+          switch (flag) {
+            case "+":
+              flags |= PLUS;
+              break;
+            case " ":
+              flags |= SPACE;
+              break;
+            case "0":
+              flags |= ZERO;
+              break;
+            case "#":
+              flags |= HASH;
+              break;
+          }
+        }
+      }
+      cFlags = flags;
+
+      if (nWidth) {
+        nWidth = parseInt(nWidth);
+      }
+
+      let intPart = Math.trunc(arg);
+
+      if (cConvChar === "x") {
+        let hex = Math.abs(intPart).toString(16).toUpperCase();
+        if (nWidth !== undefined) {
+          hex = hex.padStart(nWidth, cFlags & ZERO ? "0" : " ");
+        }
+        if (cFlags & HASH) {
+          hex = `0x${hex}`;
+        }
+        return hex;
+      }
+
+      if (nPrecision) {
+        nPrecision = parseInt(nPrecision.substring(1));
+      }
+
+      nDecSep = nDecSep ? nDecSep.substring(1) : "0";
+      const separators = {
+        0: [",", "."],
+        1: ["", "."],
+        2: [".", ","],
+        3: ["", ","],
+        4: ["'", "."],
+      };
+      const [thousandSep, decimalSep] = separators[nDecSep];
+
+      let decPart = "";
+      if (cConvChar === "f") {
+        if (nPrecision !== undefined) {
+          decPart = (arg - intPart).toFixed(nPrecision);
+        } else {
+          decPart = (arg - intPart).toString();
+        }
+        if (decPart.length > 2) {
+          decPart = `${decimalSep}${decPart.substring(2)}`;
+        } else if (cFlags & HASH) {
+          decPart = ".";
+        } else {
+          decPart = "";
+        }
+      }
+
+      let sign = "";
+      if (intPart < 0) {
+        sign = "-";
+        intPart = -intPart;
+      } else if (cFlags & PLUS) {
+        sign = "+";
+      } else if (cFlags & SPACE) {
+        sign = " ";
+      }
+
+      if (thousandSep && intPart >= 1000) {
+        const buf = [];
+        while (true) {
+          buf.push((intPart % 1000).toString().padStart(3, "0"));
+          intPart = Math.trunc(intPart / 1000);
+          if (intPart < 1000) {
+            buf.push(intPart.toString());
+            break;
+          }
+        }
+        intPart = buf.reverse().join(thousandSep);
+      } else {
+        intPart = intPart.toString();
+      }
+
+      let n = `${intPart}${decPart}`;
+      if (nWidth !== undefined) {
+        n = n.padStart(nWidth - sign.length, cFlags & ZERO ? "0" : " ");
+      }
+
+      return `${sign}${n}`;
+    });
+  }
+
+  iconStreamFromIcon() {
+    /* Not implemented */
+  }
+
+  printd(cFormat, oDate) {
+    switch (cFormat) {
+      case 0:
+        return this.printd("D:yyyymmddHHMMss", oDate);
+      case 1:
+        return this.printd("yyyy.mm.dd HH:MM:ss", oDate);
+      case 2:
+        return this.printd("m/d/yy h:MM:ss tt", oDate);
+    }
+
+    const handlers = {
+      mmmm: data => {
+        return this._months[data.month];
+      },
+      mmm: data => {
+        return this._months[data.month].substring(0, 3);
+      },
+      mm: data => {
+        return (data.month + 1).toString().padStart(2, "0");
+      },
+      m: data => {
+        return (data.month + 1).toString();
+      },
+      dddd: data => {
+        return this._days[data.dayOfWeek];
+      },
+      ddd: data => {
+        return this._days[data.dayOfWeek].substring(0, 3);
+      },
+      dd: data => {
+        return data.day.toString().padStart(2, "0");
+      },
+      d: data => {
+        return data.day.toString();
+      },
+      yyyy: data => {
+        return data.year.toString();
+      },
+      yy: data => {
+        return (data.year % 100).toString().padStart(2, "0");
+      },
+      HH: data => {
+        return data.hours.toString().padStart(2, "0");
+      },
+      H: data => {
+        return data.hours.toString();
+      },
+      hh: data => {
+        return (1 + ((data.hours + 11) % 12)).toString().padStart(2, "0");
+      },
+      h: data => {
+        return (1 + ((data.hours + 11) % 12)).toString();
+      },
+      MM: data => {
+        return data.minutes.toString().padStart(2, "0");
+      },
+      M: data => {
+        return data.minutes.toString();
+      },
+      ss: data => {
+        return data.seconds.toString().padStart(2, "0");
+      },
+      s: data => {
+        return data.seconds.toString();
+      },
+      tt: data => {
+        return data.hours < 12 ? "am" : "pm";
+      },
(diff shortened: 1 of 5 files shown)


src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
     super(data);
 
     this._crackURL = data.crackURL;
+    this._scandCache = Object.create(null);
```
Snuffleupagus, line 23: Perhaps make this a `Map` instead, since that avoids e.g. using `in` to check if things exist in the cache?

src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
...
+        return this.scand("D:yyyymmddHHMMss", cDate);
+      case "1":
+        return this.scand("yyyy.mm.dd HH:MM:ss", cDate);
+      case "2":
+        return this.scand("m/d/yy h:MM:ss tt", cDate);
+    }
+
+    if (!(cFormat in this._scandCache)) {
+      const _months = this._months;
+      const _days = this._days;
```
Snuffleupagus, line 557: Note that there's not really any precedent in the PDF.js code-base for having *local variable* names start with an underscore. (That's usually limited to indicate "private" properties/methods on e.g. objects and classes.)

src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
...
+
+      const handlers = {
+        mmmm: {
+          pat: `(${_months.join("|")})`,
```
Snuffleupagus, line 561: What is this short for, `pattern` perhaps?
If so, it'd probably not hurt to just spell that out here to avoid confusion.

src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
...
+  printd(cFormat, cDate, bXFAPicture = false) {
+    if (bXFAPicture) {
+      return this._printdXFAPicture(cFormat, cDate);
+    }
+
+    switch (cFormat) {
+      case "0":
+        return this.printd("D:yyyymmddHHMMss", cDate);
+      case "1":
+        return this.printd("yyyy.mm.dd HH:MM:ss", cDate);
```
timvandermeij, line 196: Unless I'm missing something here, these formats don't match the examples at https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=718&zoom=auto,-207,541. Value 1 starts with `D:` for example, while this format doesn't, and the one above can also have a timezone offset specifier. The one below is yyyy/mm/dd in the spec.

src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
...
+    const year = cDate.getFullYear();
+    const month = cDate.getMonth();
+    const day = cDate.getDate();
+    const dayOfWeek = cDate.getDay();
+    const hours = cDate.getHours();
+    const minutes = cDate.getMinutes();
+    const seconds = cDate.getSeconds();
+    const data = { year, month, day, dayOfWeek, hours, minutes, seconds };
+
+    const pattern = /(mmmm|mmm|mm|m|dddd|ddd|dd|d|yyyy|yy|HH|H|hh|h|MM|M|ss|s|tt|t|\\.)/g;
```
timvandermeij, line 273: Why is the `\\.` part here?

src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
...
+    return cFormat.replace(pattern, function (match, p1) {
```
timvandermeij, line 274: `p1` -> `pattern`

timvandermeij (commented):
I have reviewed everything from the top of the file until `_getTZ` so far, so this is the first set of comments. It's quite a lot of code and requires careful matching with the spec, so I'll do it in chunks. Nice work so far!

src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
...
+  printf(...args) {
+    if (args.length === 0) {
+      throw new Error("Invalid number of params in printf");
+    }
+
+    if (typeof args[0] !== "string") {
+      throw new TypeError("First argument of printf must be a string");
+    }
+
+    const pattern = /%(,[0-9])?([\+ 0#]+)?([0-9]+)?(\.[0-9]+)?([dfsx])/g;
```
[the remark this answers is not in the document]
calixteman: Yep agreed and I did that but:
- look at function `AFPercent_Format` in http://www.sfu.ca/~wcs/ForGraham/Aladdin%20stuff/Acrobat%20Reader%205.0/Contents/MacOS/JavaScripts/AForm.js, the sepStyle argument is used as the nDecSep part of the format.
- look at the patch here: https://pdfium.googlesource.com/pdfium/+/refs/heads/chromium/3716

so I changed the `[0-3]` into a `[0-9]` and handle that here:
https://github.com/mozilla/pdf.js/pull/12530/files#diff-cffce7b4153238cbf3dc690777f5f5890edf1eaa4f2fcd9e42a4a603266efd48R131

and looking at pdfium sources: the `nDecSep` part is not handled.

src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
...
+
+      let intPart = Math.trunc(arg);
+
+      if (cConvChar === "x") {
+        let hex = Math.abs(intPart).toString(16).toUpperCase();
+        if (nWidth !== undefined) {
+          hex = hex.padStart(nWidth, cFlags & ZERO ? "0" : " ");
+        }
+        if (cFlags & HASH) {
+          hex = `0x${hex}`;
```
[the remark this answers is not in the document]
calixteman: I found that here:
https://en.wikipedia.org/wiki/Printf_format_string#Flags_field

src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
...
+      }
+
+      const separators = {
+        0: [",", "."],
+        1: ["", "."],
+        2: [".", ","],
+        3: ["", ","],
+      };
+      const [thousandSep, decimalSep] =
+        nDecSep in separators ? separators[nDecSep] : ["'", "."];
```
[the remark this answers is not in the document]
calixteman: yes that's true but as explained above it's a workaround for `sepStyle`

src/scripting_api/util.js
calixteman: The spec is pretty quite strange here:
 - 0 & 1 are the same
 - in the `cFormat` there is nothing related to the timezone offset
 So I took the format in pdfium:
https://pdfium.googlesource.com/pdfium/+/refs/heads/master/fxjs/cjs_util.cpp#187

But in looking at 7.9.4 in pdf spec a date has the format `D:YYYYMMDDHHmmSSOHH'mm` so it could have an offset specifier.
And maybe "Universal" (1) should be `YYYY-MM-DDTHH:mm:SSZHH(Z|[+-]]HHmm)`.
Wdyt ?

src/scripting_api/util.js
calixteman: see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=719
...
`\` use as an escape character

src/scripting_api/util.js
calixteman: I just checked in Acrobat (Mac version) and the display is exactly the same as the one defined in pdfium source.
So I guess the js specs are incorrect here.

src/scripting_api/util.js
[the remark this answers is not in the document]
calixteman: I checked in Acrobat and `,4` is ok. In this case decimal sep is `.` and thousand sep is `'`.

src/scripting_api/util.js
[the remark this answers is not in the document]
calixteman: Ok, I removed that stuff and added a case for 4.

src/scripting_api/util.js
```diff
@@ -20,11 +20,724 @@ class Util extends PDFObject {
...
+      "Saturday",
+    ];
   }
 
   crackURL(cURL) {
     return this._crackURL(cURL);
   }
+
+  printf(...args) {
+    if (args.length === 0) {
```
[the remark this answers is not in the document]
calixteman: The spec is quite confusing: in `Parameters` section we've `The number of optional arguments must be the same as the number of % tags`.
I checked in Acrobat and `util.printf('hello world')` is ok.

src/scripting_api/util.js
```diff
@@ -20,11 +20,734 @@ class Util extends PDFObject {
...
+      "Friday",
+      "Saturday",
+    ];
   }
 
   crackURL(cURL) {
     return this._crackURL(cURL);
   }
+
+  printf(...args) {
```
brendandahl, line 53: Is the PDF's printf different than c's printf? It doesn't look like below handles escaping. For example in c `printf("%%s")` should print `%s`

calixteman: `printf("%%s", "foo")` and `printf("%%s")` both print `%%s` in Acrobat.
So I must change my approach.

src/scripting_api/util.js
timvandermeij: Ugh, Adobe's specs are indeed sometimes confusing, incomplete or simply don't match their reference implementation (Adobe Reader/Acrobat), so unfortunately I'm not really surprised. Thank you for figuring this all out!

src/scripting_api/util.js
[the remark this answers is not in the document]
timvandermeij: Alright. Can we then add the `util.printf('hello world')` to the unit tests to make sure it's documented that we expect this to work?

src/scripting_api/util.js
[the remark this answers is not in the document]
timvandermeij: Nice, looks much better like this!

src/scripting_api/util.js
```diff
@@ -20,11 +20,734 @@ class Util extends PDFObject {
...
+      day,
+      dayOfWeek,
+      hours,
+      minutes,
+      seconds,
+      milliseconds,
+      cDate,
+    };
+
+    const pattern = /(MMMM|MMM|MM|M|DD|D|YYYY|YY|JJJ|J|EEEE|EEE|E|G|w|WW|hh|h|kk|k|HH|H|KK|K|mm|m|SS|S|FFF|A|Z|zz|z|\\.)/g;
```
timvandermeij, line 459: Which specification did you use for this? I could only find http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.119.914&rep=rep1&type=pdf#page=842&zoom=auto,-207,487 but there for example the symbol `e` is also allowed and it's not in this regex. I'm not sure if this is the right spec though.

calixteman: It's likely the right spec and mine was older than yours: I added the `e`.
Anyway there is something wrong with this spec: `MM` is used for month and minutes, and a js Date is a date+time... so I don't know how to handle that correctly. So for now I just replaced the `MM` for minutes by `mm`.

src/scripting_api/util.js
```diff
@@ -20,11 +20,753 @@ class Util extends PDFObject {
...
+      }
+      switch (command) {
+        case "?":
+          buf.push(currCase(cSource.charAt(i++)));
+          break;
+        case "X":
+          while (i < ii) {
+            const code = cSource.charCodeAt(i++);
+            if (
+              (LA <= code && code <= LZ) ||
```
timvandermeij, line 519: Wouldn't it be possible to use `chatAt` instead so this can be simplified to `code >= 'a' && code <= 'z'`? Then you can remove the `limits` variable above and it seems more readable to me to remove the charcode conversion if it's not necessary.

timvandermeij (commented):
I have now reviewed everything up until `scand`.

src/scripting_api/util.js
```diff
@@ -20,11 +20,753 @@ class Util extends PDFObject {
...
+      k: data => {
+        return (data.hours % 12).toString();
+      },
+      KK: data => {
+        return (1 + ((data.hours + 23) % 24)).toString().padStart(2, "0");
+      },
+      K: data => {
+        return (1 + ((data.hours + 23) % 24)).toString();
+      },
+      mm: data => {
```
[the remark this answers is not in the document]
calixteman: It's conflicting with he `MM` for month... I don't really know how to handle that.

src/scripting_api/util.js
```diff
@@ -20,11 +20,753 @@ class Util extends PDFObject {
...
+      day,
+      dayOfWeek,
+      hours,
+      minutes,
+      seconds,
+      milliseconds,
+      cDate,
+    };
+
+    const patterns = /(MMMM|MMM|MM|M|DD|D|YYYY|YY|JJJ|J|EEEE|EEE|E|e|G|w|WW|hh|h|kk|k|HH|H|KK|K|mm|m|SS|S|FFF|A|Z|zz|z|\\.)/g;
```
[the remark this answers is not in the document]
calixteman: Ok for the backslash, but chars `*,?,+` are used when parsing and here we're just printing a date according to a format so I guess it's unrelated.

src/scripting_api/util.js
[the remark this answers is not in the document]
calixteman: I tried to check how acrobat (last version) is dealing with that stuff and astonishingly it doesn't handle the 3rd argument (i.e. bXFAPicture), so it's quite confusing.

calixteman:
Do you know why CI is broken ?

Snuffleupagus:
> Do you know why CI is broken ?

Because the new files aren't included in the `lib`/`lib-es5` build targets; the following should fix it:
```diff
   return merge([
     gulp.src(
       [
-        "src/{core,display,shared}/*.js",
+        "src/{core,display,scripting_api,shared}/*.js",
         "!src/shared/{cffStandardStrings,fonts_utils}.js",
-        "src/{pdf,pdf.worker}.js",
+        "src/{pdf,pdf.scripting,pdf.worker}.js",
       ],
       { base: "src/" }
     ),
```

src/scripting_api/util.js
brendandahl: I think there are still some issues:
`%%%%s` should be `%%s`
or `"%%%s", "a"` should be `%a`

IIRC, printf format is not a regular grammar so a regex can't be used to parse it.

calixteman: Did you try it in acrobat ?
According to the spec, only `printf` function is pretty basic:
https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=716
or I missed something

brendandahl: Tried in acrobat, you're right. I was going off of how printf in c actually works. I guess it's not the same cFormat.

src/scripting_api/util.js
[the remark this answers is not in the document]
timvandermeij: If Acrobat isn't handling it at all and there is confusion in the specification, I'd really like to remove it completely from this patch. It will be more consistent with Acrobat (our reference implementation) and that avoids us from filling in any blanks we can't verify. Moreover, I don't think this functionality is actually used often, and it would make the overall patch considerable smaller. It can always be added again in a follow-up PR.

timvandermeij:
Looks good to me too. Thanks!

PR #12534, opened 2020-10-26
Ensure that `MurmurHash3_64.update` handles `ArrayBuffer` input correctly, to avoid hash-collisions (issue 12533)

### LINKED ISSUE #12533: Large characters rendered overlapping on PDF.js in Firefox 82
Example pdf: https://www.mass.gov/doc/covid-19-dashboard-october-25-2020/download -- and most of the previous versions (https://www.mass.gov/info-details/archive-of-covid-19-cases-in-massachusetts) as well

Configuration:
- Web browser and its version: Firefox v82.0
- PDF.js version: whatever's bundled with Firefox
- Is a browser extension: no

1. View pages 18 or 20 of the PDF for examples

Actual results:
image

Expected results (in Chrome):
image

Snuffleupagus:
Different fonts incorrectly end up with *identical* hashes, despite having different /ToUnicode data.
The issue, and it's very interesting that we've apparently not seen it before, appears to be caused by the fact that different /ToUnicode entries share the *same* underlying `ArrayBuffer`, which thus becomes problematic at the `const dataUint32 = new Uint32Array(data.buffer, 0, blockCounts);` line. The simplest solution thus seem to be to just *copy* the input, when it's an `ArrayBuffer`, rather than using it as-is. (Note that if we'd stringified the input, when calling `MurmurHash3_64.update`, the issue would also have been fixed. In this case, we're already creating an unique TypedArray.)

Fixes #12533

### DIFF
--- a/src/core/murmurhash3.js
+++ b/src/core/murmurhash3.js
@@ -45,7 +45,7 @@ class MurmurHash3_64 {
         }
       }
     } else if (isArrayBuffer(input)) {
-      data = input;
+      data = input.slice();
       length = data.byteLength;
     } else {
       throw new Error(
--- a/test/pdfs/issue12533.pdf.link
+++ b/test/pdfs/issue12533.pdf.link
@@ -1,0 +1,1 @@
+https://web.archive.org/web/20201026152003/https://www.mass.gov/doc/covid-19-dashboard-october-25-2020/download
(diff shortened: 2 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Nice find; that's a really interesting one!

/botio-linux makeref

PR #12540, opened 2020-10-27
Ensure that the same version of PDF.js is used in both the API and the Viewer (PR 8959 follow-up)

Snuffleupagus:
Given that we're now accessing certain API-functionality *directly* in this file, e.g. the AnnotationStorage and Optional Content configuration, ensuring that there's not a version mismatch definitely seem like a good idea to prevent any *subtle* future bugs.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -13,6 +13,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

PR #12542, opened 2020-10-28
Add a `MurmurHash3_64.update` unit-test for TypedArrays which share the same underlying ArrayBuffer (PR 12534 follow-up)

Snuffleupagus:
This probably ought to have been included in PR #12534, but better late than never I suppose, since it helps to more clearly demonstrate the bug in a way that a reference-test alone just cannot do.

When writing this unit-test I also noticed that it required a certain amount of "luck" to actually trigger the bug, prior to the patch, since it seems that the bug only reproduced for certain *unfortunate* sequences of TypedArray data. (The added unit-test contains one such, purposely simple, example.)


timvandermeij (approved):

timvandermeij:
Nice; thanks for the follow-up!

PR #12543, opened 2020-10-28
Use a `Map`, rather than an `Object`, internally in the `Catalog.openAction` getter (PR 11644 follow-up)

Snuffleupagus:
This provides a work-around to avoid having to conditionally try to initialize the `openAction`-object in multiple places; this indirectly fixes https://github.com/mozilla/pdf.js/pull/11644#issuecomment-635396349.
Given that `Object.fromEntries` doesn't seem to *guarantee* that a `null` prototype is used, we thus hack around that by using `Object.assign` with `Object.create(null)`.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -25,6 +25,7 @@ import {
   isBool,
   isNum,
   isString,
+  objectFromEntries,
   PermissionFlag,
   shadow,
   stringToPDFString,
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #12546, opened 2020-10-28
[api-minor] JS -- Add listener for sandbox events only if there are some actions

calixteman:
* When no actions then set it to null instead of empty object
* Even if a field has no actions, it needs to listen to events from the sandbox in order to be updated if an action changes something in it.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -1020,6 +1020,21 @@ class PDFDocument {
     );
   }
 
+  get hasJSActions() {
+    return shadow(
+      this,
+      "hasJSActions",
+      this.fieldObjects.then(fieldObjects => {
+        return (
+          fieldObjects !== null &&
+          Object.values(fieldObjects).some(fieldObject =>
+            fieldObject.some(object => object.actions !== null)
+          )
+        );
+      })
+    );
+  }
+
   get calculationOrderIds() {
     const acroForm = this.catalog.acroForm;
     if (!acroForm || !acroForm.has("CO")) {
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -43,6 +43,8 @@ import { AnnotationStorage } from "./annotation_storage.js";
  *   for annotation icons. Include trailing slash.
  * @property {boolean} renderInteractiveForms
  * @property {Object} svgFactory
+ * @property {boolean} [enableScripting]
+ * @property {boolean} [hasJSActions]
  */
 
 class AnnotationElementFactory {
@@ -142,6 +144,8 @@ class AnnotationElement {
     this.renderInteractiveForms = parameters.renderInteractiveForms;
     this.svgFactory = parameters.svgFactory;
     this.annotationStorage = parameters.annotationStorage;
+    this.enableScripting = parameters.enableScripting;
+    this.hasJSActions = parameters.hasJSActions;
 
     if (isRenderable) {
       this.container = this._createContainer(ignoreBorder);
@@ -507,7 +511,7 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         event.target.setSelectionRange(0, 0);
       });
 
-      if (this.data.actions) {
+      if (this.enableScripting && this.hasJSActions) {
         element.addEventListener("updateFromSandbox", function (event) {
           const data = event.detail;
           if ("value" in data) {
@@ -517,21 +521,23 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
           }
         });
 
-        for (const eventType of Object.keys(this.data.actions)) {
-          switch (eventType) {
-            case "Format":
-              element.addEventListener("blur", function (event) {
-                window.dispatchEvent(
-                  new CustomEvent("dispatchEventInSandbox", {
-                    detail: {
-                      id,
-                      name: "Format",
-                      value: event.target.value,
-                    },
-                  })
-                );
-              });
-              break;
+        if (this.data.actions !== null) {
+          for (const eventType of Object.keys(this.data.actions)) {
+            switch (eventType) {
+              case "Format":
+                element.addEventListener("blur", function (event) {
+                  window.dispatchEvent(
+                    new CustomEvent("dispatchEventInSandbox", {
+                      detail: {
+                        id,
+                        name: "Format",
+                        value: event.target.value,
+                      },
+                    })
+                  );
+                });
+                break;
+            }
           }
         }
       }
@@ -1562,6 +1568,9 @@ class FileAttachmentAnnotationElement extends AnnotationElement {
  * @property {string} [imageResourcesPath] - Path for image resources, mainly
  *   for annotation icons. Include trailing slash.
  * @property {boolean} renderInteractiveForms
+ * @property {boolean} [enableScripting] - Enable embedded script execution.
+ * @property {boolean} [hasJSActions] - Some fields have JS actions.
+ *   The default value is `false`.
  */
 
 class AnnotationLayer {
@@ -1608,6 +1617,8 @@ class AnnotationLayer {
         svgFactory: new DOMSVGFactory(),
         annotationStorage:
           parameters.annotationStorage || new AnnotationStorage(),
+        enableScripting: parameters.enableScripting,
+        hasJSActions: parameters.hasJSActions,
       });
       if (element.isRenderable) {
         const rendered = element.render();
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -904,6 +904,14 @@ class PDFDocumentProxy {
   }
 
   /**
+   * @returns {Promise<boolean>} A promise that is resolved with `true`
+   * if some /AcroForm fields have JavaScript actions.
+   */
+  hasJSActions() {
+    return this._transport.hasJSActions();
+  }
+
+  /**
    * @returns {Promise<Array<string> | null>} A promise that is resolved with an
    *   {Array<string>} containing IDs of annotations that have a calculation
    *   action, or `null` when no such annotations are present in the PDF file.
@@ -2566,6 +2574,10 @@ class WorkerTransport {
 
   getFieldObjects() {
     return this.messageHandler.sendWithPromise("GetFieldObjects", null);
+  }
+
+  hasJSActions() {
+    return this.messageHandler.sendWithPromise("HasJSActions", null);
   }
 
   getCalculationOrderIds() {
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -77,6 +77,8 @@ const DEFAULT_CACHE_SIZE = 10;
  *   total pixels, i.e. width * height. Use -1 for no limit. The default value
  *   is 4096 * 4096 (16 mega-pixels).
  * @property {IL10n} l10n - Localization service.
+ * @property {boolean} [enableScripting] - Enable embedded script execution.
+ *   The default value is `false`.
  */
 
 function PDFPageViewBuffer(size) {
@@ -187,6 +189,7 @@ class BaseViewer {
     this.useOnlyCssZoom = options.useOnlyCssZoom || false;
     this.maxCanvasPixels = options.maxCanvasPixels;
     this.l10n = options.l10n || NullL10n;
+    this.enableScripting = options.enableScripting || false;
 
     this.defaultRenderingQueue = !options.renderingQueue;
     if (this.defaultRenderingQueue) {
@@ -527,6 +530,7 @@ class BaseViewer {
             useOnlyCssZoom: this.useOnlyCssZoom,
             maxCanvasPixels: this.maxCanvasPixels,
             l10n: this.l10n,
+            enableScripting: this.enableScripting,
           });
           this._pages.push(pageView);
         }
@@ -1208,6 +1212,7 @@ class BaseViewer {
    *   for annotation icons. Include trailing slash.
    * @param {boolean} renderInteractiveForms
    * @param {IL10n} l10n
+   * @param {boolean} [enableScripting]
    * @returns {AnnotationLayerBuilder}
    */
   createAnnotationLayerBuilder(
@@ -1216,7 +1221,8 @@ class BaseViewer {
     annotationStorage = null,
     imageResourcesPath = "",
     renderInteractiveForms = false,
-    l10n = NullL10n
+    l10n = NullL10n,
+    enableScripting = false
   ) {
     return new AnnotationLayerBuilder({
       pageDiv,
@@ -1227,6 +1233,10 @@ class BaseViewer {
       linkService: this.linkService,
       downloadManager: this.downloadManager,
       l10n,
+      enableScripting,
+      hasJSActionsPromise: enableScripting
+        ? this.pdfDocument.hasJSActions()
+        : Promise.resolve(false),
     });
   }
 
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -63,6 +63,8 @@ import { viewerCompatibilityParams } from "./viewer_compatibility.js";
  *   total pixels, i.e. width * height. Use -1 for no limit. The default value
  *   is 4096 * 4096 (16 mega-pixels).
  * @property {IL10n} l10n - Localization service.
+ * @property {boolean} [enableScripting] - Enable embedded script execution.
+ *   The default value is `false`.
  */
 
 const MAX_CANVAS_PIXELS = viewerCompatibilityParams.maxCanvasPixels || 16777216;
@@ -109,6 +111,7 @@ class PDFPageView {
     this.renderer = options.renderer || RendererType.CANVAS;
     this.enableWebGL = options.enableWebGL || false;
     this.l10n = options.l10n || NullL10n;
+    this.enableScripting = options.enableScripting || false;
 
     this.paintTask = null;
     this.paintedViewportMap = new WeakMap();
@@ -549,7 +552,8 @@ class PDFPageView {
           this._annotationStorage,
           this.imageResourcesPath,
           this.renderInteractiveForms,
-          this.l10n
+          this.l10n,
+          this.enableScripting
         );
       }
       this._renderAnnotationLayer();
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -525,6 +525,10 @@ class WorkerMessageHandler {
       return pdfManager.ensureDoc("fieldObjects");
     });
 
+    handler.on("HasJSActions", function (data) {
+      return pdfManager.ensureDoc("hasJSActions");
+    });
+
     handler.on("GetCalculationOrderIds", function (data) {
       return pdfManager.ensureDoc("calculationOrderIds");
     });
--- a/src/scripting_api/field.js
+++ b/src/scripting_api/field.js
@@ -73,12 +73,14 @@ class Field extends PDFObject {
     // Private
     this._actions = Object.create(null);
     const doc = (this._document = data.doc);
-    for (const [eventType, actions] of Object.entries(data.actions)) {
-      // This code is running in a sandbox so it's safe to use Function
-      this._actions[eventType] = actions.map(action =>
-        // eslint-disable-next-line no-new-func
-        Function("event", `with (this) {${action}}`).bind(doc)
-      );
+    if (data.actions !== null) {
+      for (const [eventType, actions] of Object.entries(data.actions)) {
+        // This code is running in a sandbox so it's safe to use Function
+        this._actions[eventType] = actions.map(action =>
+          // eslint-disable-next-line no-new-func
+          Function("event", `with (this) {${action}}`).bind(doc)
+        );
+      }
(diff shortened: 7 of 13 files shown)


web/base_viewer.js
```diff
@@ -497,8 +500,8 @@ class BaseViewer {
 
     // Fetch a single page so we can get a viewport that will be the default
     // viewport for all pages
-    firstPagePromise
-      .then(firstPdfPage => {
+    Promise.all([firstPagePromise, hasJSActionsPromise])
```
brendandahl, line 503: This section is pretty important to display things as fast as possible. Could we pass in the hasJSActionsPromise into the page view instead so it's resolved a a later time and we don't have to wait for it to starting showing pages?

brendandahl (changes requested):
Looks good overall, but I'd like to hear if it's possible to address the comment.

Snuffleupagus (changes requested):
Can't we actually delay this *completely* until Annotation parsing actually occurs, since in most documents there's either no annotations present or at least no JavaScript actions present!?

That'd probably reduce the overall size of this, and also simplify the implementation (especially on the viewer side). One quick idea that I had, was to simply include this property in the annotation data *itself* such that you'll only need to pass around the `enableScripting`-value to the various viewer-components.

To illustrate my idea, perhaps something along these lines (purposely incomplete, but only to illustrate my idea):
```diff
diff --git a/src/core/annotation.js b/src/core/annotation.js
index 252e1529b..f28806726 100644
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -62,13 +62,17 @@ class AnnotationFactory {
    *   instance.
    */
   static create(xref, ref, pdfManager, idFactory) {
-    return pdfManager.ensureCatalog("acroForm").then(acroForm => {
+    return Promise.all([
+      pdfManager.ensureCatalog("acroForm"),
+      pdfManager.ensureDoc("hasJSActions"),
+    ]).then(([acroForm, hasJSActions]) => {
       return pdfManager.ensure(this, "_create", [
         xref,
         ref,
         pdfManager,
         idFactory,
         acroForm,
+        hasJSActions,
       ]);
     });
   }
@@ -76,7 +80,14 @@ class AnnotationFactory {
   /**
    * @private
    */
-  static _create(xref, ref, pdfManager, idFactory, acroForm) {
+  static _create(
+    xref,
+    ref,
+    pdfManager,
+    idFactory,
+    acroForm,
+    hasJSActions = false
+  ) {
     const dict = xref.fetchIfRef(ref);
     if (!isDict(dict)) {
       return undefined;
@@ -97,6 +108,7 @@ class AnnotationFactory {
       id,
       pdfManager,
       acroForm: acroForm instanceof Dict ? acroForm : Dict.empty,
+      hasJSActions,
     };
 
     switch (subtype) {
@@ -281,6 +293,7 @@ class Annotation {
       modificationDate: this.modificationDate,
       rect: this.rectangle,
       subtype: params.subtype,
+      documentHasJSActions: params.hasJSActions,
     };
 
     this._fallbackFontDict = null;
diff --git a/src/display/annotation_layer.js b/src/display/annotation_layer.js
index effa122b5..4e07676fe 100644
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -478,7 +478,8 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         event.target.setSelectionRange(0, 0);
       });
 
-      if (this.data.actions) {
+      // Assuming that the `enableScripting`-value is passed in:
+      if (this.enableScripting && this.data.documentHasJSActions) {
         element.addEventListener("updateFromSandbox", function (event) {
           const data = event.detail;
           if ("value" in data) {
```

calixteman:
I think this is a very good idea: I like it, thank you.

brendandahl (approved):

src/core/annotation.js
```diff
@@ -59,25 +59,39 @@ class AnnotationFactory {
    * @param {Object} ref
    * @param {PDFManager} pdfManager
    * @param {Object} idFactory
+   * @param {boolean} checkJSActions
    * @returns {Promise} A promise that is resolved with an {Annotation}
    *   instance.
    */
-  static create(xref, ref, pdfManager, idFactory) {
-    return pdfManager.ensureCatalog("acroForm").then(acroForm => {
+  static create(xref, ref, pdfManager, idFactory, checkJSActions = true) {
```
Snuffleupagus, line 66: Can't we make `false` the default value instead, and explicitly enable it where necessary?

src/core/annotation.js
```diff
@@ -59,25 +59,39 @@ class AnnotationFactory {
...
+    return Promise.all([
+      pdfManager.ensureCatalog("acroForm"),
+      checkJSActions
+        ? pdfManager.ensureDoc("hasJSActions")
+        : Promise.resolve(false),
```
Snuffleupagus, line 71: I believe that the following ought to suffice, since non-Promise values should be automatically wrapped in a Promise:
```suggestion
      checkJSActions && pdfManager.ensureDoc("hasJSActions"),
```

src/core/document.js
```diff
@@ -962,12 +962,14 @@ class PDFDocument {
     if (!(name in promises)) {
       promises.set(name, []);
     }
+
```
Snuffleupagus, line 965: Nit: Looks like an unrelated change!?

src/display/api.js
```diff
@@ -903,6 +903,14 @@ class PDFDocumentProxy {
...
   }
 
+  /**
+   * @returns {Promise<Boolean>} A promise that is resolved with true if
+   * there are some actions in JavaScript.
+   */
+  hasJSActions() {
+    return this._transport.hasJSActions();
+  }
+
```
Snuffleupagus, line 913: It *looks* like this method is now completely unused/untested in the latest version of the patch.

Do we actually need to add this to the API, or could we instead wait until we see if there's an actual use for this? Generally speaking, it feels best to avoid adding unused things to the official API since we then have to maintain something that's unused/untested.

Hence I'd suggest reverting the changes in both the `src/display/api.js` and `src/core/worker.js` files for now.

web/base_viewer.js
```diff
@@ -77,6 +77,7 @@ const DEFAULT_CACHE_SIZE = 10;
  *   total pixels, i.e. width * height. Use -1 for no limit. The default value
  *   is 4096 * 4096 (16 mega-pixels).
  * @property {IL10n} l10n - Localization service.
+ * @property {boolean} [enableScripting] - Enable JS evaluation.
```
Snuffleupagus, line 80: Please mention the default value here, and make sure to use the *same* description in both `BaseViewer` and `PDFPageView`, i.e.
```suggestion
 * @property {boolean} [enableScripting] - Enable embedded script execution.
 *   The default value is `false`.
```

web/pdf_page_view.js
```diff
@@ -63,6 +63,7 @@ import { viewerCompatibilityParams } from "./viewer_compatibility.js";
  *   total pixels, i.e. width * height. Use -1 for no limit. The default value
  *   is 4096 * 4096 (16 mega-pixels).
  * @property {IL10n} l10n - Localization service.
+ * @property {boolean} enableScripting - Enable embedded script execution.
```
Snuffleupagus, line 66: Please mention the default value here as well:
```javascript
```
`*   The default value is `false`.`

src/display/annotation_layer.js
```diff
@@ -1565,6 +1568,7 @@ class AnnotationLayer {
         svgFactory: new DOMSVGFactory(),
         annotationStorage:
           parameters.annotationStorage || new AnnotationStorage(),
+        enableScripting: parameters.enableScripting,
```
Snuffleupagus, line 1571: You also need to update the `AnnotationElementParameters`-typedef at the top of this file, hence please add the following there:
```javascript
 * @property {boolean} [enableScripting] - Enable embedded script execution.
```

Snuffleupagus (commented):
Looks much better this way; I've added a few comments, after which this should be good to go (assuming all tests pass of course :-)

src/core/document.js
```diff
@@ -1020,6 +1021,26 @@ class PDFDocument {
...
+  get hasJSActions() {
+    return shadow(
+      this,
+      "hasJSActions",
+      this.fieldObjects.then(objects => {
+        if (objects === null) {
+          return false;
+        }
+        for (const object of Object.values(objects)) {
+          for (const obj of object) {
```
timvandermeij, line 1033: It might be more clear to rename `objects` -> `fieldObjects`, `object` -> `fieldObject` and `obj` -> `object`.

timvandermeij (commented):
Looks good but I'd really like to see the property comment addressed since it doesn't feel like the right place to put it.

src/core/annotation.js
```diff
@@ -282,6 +295,7 @@ class Annotation {
       modificationDate: this.modificationDate,
       rect: this.rectangle,
       subtype: params.subtype,
+      documentHasJSActions: params.hasJSActions,
```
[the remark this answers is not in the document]
Snuffleupagus: > I don't really see why we need to and would want to store a document-wide property on every individual annotation, especially since this is then also an API change. 

Sorry, but how is adding a *new* boolean to the data returned by `getAnnotations` a problem? :-)
The data in question is also cached, on the worker-side, so that shouldn't be much of an issue.

The reason why I suggested the previous approach, is that doing it the way you suggest (and what the patch originally did), has *really* undesirable consequences for the viewer since that first of all gives a number of the viewer-components (e.g. `PDFPageView`) *very* inconsistent API-surfaces!
Note also the point below, about unnecessary things API-calls running too early.

> It would be much better to use the same logic as enableScripting, which is also a document-wide property, where we calculate the property once and pass it into the AnnotationLayer class.

Sorry, but I don't agree with this at all, since it ends up forcing an unnecessary API-call early on, *before* rendering has started. That's why I suggested the previous approach, since it delays these things until the Annotations are actually fetched.
You absolutely want to avoid *all* strictly unnecessary API-calls *before* rendering has started!

---

Given that we apparently have very different positions/perspectives here, please give me some time to suggest a compromise here!

web/pdf_page_view.js
```diff
@@ -541,26 +547,30 @@ class PDFPageView {
...
+      div.setAttribute("data-loaded", true);
 
-    this.eventBus.dispatch("pagerender", {
-      source: this,
-      pageNumber: this.id,
+      this.eventBus.dispatch("pagerender", {
+        source: this,
+        pageNumber: this.id,
+      });
+      return resultPromise;
```
Snuffleupagus, line 572: You *absolutely* can't delay this event, since that will affect things like clean-up (and thus resource usage) in the viewer negatively.

Snuffleupagus (changes requested):

src/core/annotation.js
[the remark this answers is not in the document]
Snuffleupagus: I'd suggest the following approach, to ensure that `hasJSActions` isn't called *before* rendering has started and avoid `BaseViewer`/`PDFPageView` diverging unnecessarily:
```diff
diff --git a/src/display/annotation_layer.js b/src/display/annotation_layer.js
index 705568575..b8e1c90fe 100644
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -43,6 +43,8 @@ import { AnnotationStorage } from "./annotation_storage.js";
  *   for annotation icons. Include trailing slash.
  * @property {boolean} renderInteractiveForms
  * @property {Object} svgFactory
+ * @property {boolean} [enableScripting]
+ * @property {boolean} [hasJSActions]
  */
 
 class AnnotationElementFactory {
diff --git a/src/display/api.js b/src/display/api.js
index 8ffcc1495..b58830463 100644
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -904,8 +904,8 @@ class PDFDocumentProxy {
   }
 
   /**
-   * @returns {Promise<boolean>} A promise that is resolved with true
-   * if some fields have some JS actions.
+   * @returns {Promise<boolean>} A promise that is resolved with `true` if some
+   *   /AcroForm fields has JavaScript actions.
    */
   hasJSActions() {
     return this._transport.hasJSActions();
diff --git a/web/annotation_layer_builder.js b/web/annotation_layer_builder.js
index e37ee831f..c4d47112a 100644
--- a/web/annotation_layer_builder.js
+++ b/web/annotation_layer_builder.js
@@ -28,7 +28,8 @@ import { SimpleLinkService } from "./pdf_link_service.js";
  * @property {IPDFLinkService} linkService
  * @property {DownloadManager} downloadManager
  * @property {IL10n} l10n - Localization service.
- * @property {boolean} enableScripting
+ * @property {boolean} [enableScripting]
+ * @property {boolean} [hasJSActionsPromise]
  */
 
 class AnnotationLayerBuilder {
@@ -45,7 +46,7 @@ class AnnotationLayerBuilder {
     renderInteractiveForms = true,
     l10n = NullL10n,
     enableScripting = false,
-    hasJSActions = false,
+    hasJSActionsPromise = null,
   }) {
     this.pageDiv = pageDiv;
     this.pdfPage = pdfPage;
@@ -56,7 +57,7 @@ class AnnotationLayerBuilder {
     this.l10n = l10n;
     this.annotationStorage = annotationStorage;
     this.enableScripting = enableScripting;
-    this.hasJSActions = hasJSActions;
+    this._hasJSActionsPromise = hasJSActionsPromise;
 
     this.div = null;
     this._cancelled = false;
@@ -69,7 +70,10 @@ class AnnotationLayerBuilder {
    *   annotations is complete.
    */
   render(viewport, intent = "display") {
-    return this.pdfPage.getAnnotations({ intent }).then(annotations => {
+    return Promise.all([
+      this.pdfPage.getAnnotations({ intent }),
+      this._hasJSActionsPromise,
+    ]).then(([annotations, hasJSActions = false]) => {
       if (this._cancelled) {
         return;
       }
@@ -88,7 +92,7 @@ class AnnotationLayerBuilder {
         downloadManager: this.downloadManager,
         annotationStorage: this.annotationStorage,
         enableScripting: this.enableScripting,
-        hasJSActions: this.hasJSActions,
+        hasJSActions,
       };
 
       if (this.div) {
@@ -133,8 +137,8 @@ class DefaultAnnotationLayerFactory {
    *   for annotation icons. Include trailing slash.
    * @param {boolean} renderInteractiveForms
    * @param {IL10n} l10n
-   * @param {boolean} enableScripting
-   * @param {boolean} hasJSActions
+   * @param {boolean} [enableScripting]
+   * @param {boolean} [hasJSActionsPromise]
    * @returns {AnnotationLayerBuilder}
    */
   createAnnotationLayerBuilder(
@@ -145,7 +149,7 @@ class DefaultAnnotationLayerFactory {
     renderInteractiveForms = true,
     l10n = NullL10n,
     enableScripting = false,
-    hasJSActions = false
+    hasJSActionsPromise = null
   ) {
     return new AnnotationLayerBuilder({
       pageDiv,
@@ -156,7 +160,7 @@ class DefaultAnnotationLayerFactory {
       l10n,
       annotationStorage,
       enableScripting,
-      hasJSActions,
+      hasJSActionsPromise,
     });
   }
 }
diff --git a/web/base_viewer.js b/web/base_viewer.js
index cf2c8552a..966eaf342 100644
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -468,7 +468,6 @@ class BaseViewer {
 
     const annotationStorage = pdfDocument.annotationStorage;
     const optionalContentConfigPromise = pdfDocument.getOptionalContentConfig();
-    const hasJSActionsPromise = pdfDocument.hasJSActions();
 
     this._pagesCapability.promise.then(() => {
       this.eventBus.dispatch("pagesloaded", {
@@ -532,7 +531,6 @@ class BaseViewer {
             maxCanvasPixels: this.maxCanvasPixels,
             l10n: this.l10n,
             enableScripting: this.enableScripting,
-            hasJSActionsPromise,
           });
           this._pages.push(pageView);
         }
@@ -1214,8 +1212,7 @@ class BaseViewer {
    *   for annotation icons. Include trailing slash.
    * @param {boolean} renderInteractiveForms
    * @param {IL10n} l10n
-   * @param {boolean} enableScripting
-   * @param {boolean} hasJSActions
+   * @param {boolean} [enableScripting]
    * @returns {AnnotationLayerBuilder}
    */
   createAnnotationLayerBuilder(
@@ -1225,8 +1222,7 @@ class BaseViewer {
     imageResourcesPath = "",
     renderInteractiveForms = false,
     l10n = NullL10n,
-    enableScripting = false,
-    hasJSActions = false
+    enableScripting = false
   ) {
     return new AnnotationLayerBuilder({
       pageDiv,
@@ -1238,7 +1234,7 @@ class BaseViewer {
       downloadManager: this.downloadManager,
       l10n,
       enableScripting,
-      hasJSActions,
+      hasJSActionsPromise: enableScripting && this.pdfDocument.hasJSActions(),
     });
   }
 
diff --git a/web/interfaces.js b/web/interfaces.js
index 062369362..41a4fe571 100644
--- a/web/interfaces.js
+++ b/web/interfaces.js
@@ -181,8 +181,8 @@ class IPDFAnnotationLayerFactory {
    *   for annotation icons. Include trailing slash.
    * @param {boolean} renderInteractiveForms
    * @param {IL10n} l10n
-   * @param {boolean} enableScripting
-   * @param {boolean} hasJSActions
+   * @param {boolean} [enableScripting]
+   * @param {boolean} [hasJSActionsPromise]
    * @returns {AnnotationLayerBuilder}
    */
   createAnnotationLayerBuilder(
@@ -193,7 +193,7 @@ class IPDFAnnotationLayerFactory {
     renderInteractiveForms = true,
     l10n = undefined,
     enableScripting = false,
-    hasJSActions = false
+    hasJSActionsPromise = null
   ) {}
 }
 
diff --git a/web/pdf_page_view.js b/web/pdf_page_view.js
index 2d28d9d8c..72b338764 100644
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -63,10 +63,8 @@ import { viewerCompatibilityParams } from "./viewer_compatibility.js";
  *   total pixels, i.e. width * height. Use -1 for no limit. The default value
  *   is 4096 * 4096 (16 mega-pixels).
  * @property {IL10n} l10n - Localization service.
- * @property {boolean} enableScripting - Enable embedded script execution.
+ * @property {boolean} [enableScripting] - Enable embedded script execution.
  *   The default value is `false`.
- * @property {Promise<boolean>} hasJSActionsPromise - A promise that is resolved
- * with true if some fields have some JS actions.
  */
 
 const MAX_CANVAS_PIXELS = viewerCompatibilityParams.maxCanvasPixels || 16777216;
@@ -114,7 +112,6 @@ class PDFPageView {
     this.enableWebGL = options.enableWebGL || false;
     this.l10n = options.l10n || NullL10n;
     this.enableScripting = options.enableScripting || false;
-    this.hasJSActionsPromise = options.hasJSActionsPromise;
 
     this.paintTask = null;
     this.paintedViewportMap = new WeakMap();
@@ -547,30 +544,27 @@ class PDFPageView {
       }
     );
 
-    return this.hasJSActionsPromise.then(hasJSActions => {
-      if (this.annotationLayerFactory) {
-        if (!this.annotationLayer) {
-          this.annotationLayer = this.annotationLayerFactory.createAnnotationLayerBuilder(
-            div,
-            pdfPage,
-            this._annotationStorage,
-            this.imageResourcesPath,
-            this.renderInteractiveForms,
-            this.l10n,
-            this.enableScripting,
-            hasJSActions
-          );
-        }
-        this._renderAnnotationLayer();
+    if (this.annotationLayerFactory) {
+      if (!this.annotationLayer) {
+        this.annotationLayer = this.annotationLayerFactory.createAnnotationLayerBuilder(
+          div,
+          pdfPage,
+          this._annotationStorage,
+          this.imageResourcesPath,
+          this.renderInteractiveForms,
+          this.l10n,
+          this.enableScripting
+        );
       }
-      div.setAttribute("data-loaded", true);
+      this._renderAnnotationLayer();
+    }
+    div.setAttribute("data-loaded", true);
 
-      this.eventBus.dispatch("pagerender", {
-        source: this,
-        pageNumber: this.id,
-      });
-      return resultPromise;
+    this.eventBus.dispatch("pagerender", {
+      source: this,
+      pageNumber: this.id,
     });
+    return resultPromise;
   }
 
   paintOnCanvas(canvasWrapper) {
```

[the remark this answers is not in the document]
timvandermeij: Oh, I'm perfectly fine with this approach of delaying `hasJSActions`. My concerns here were only with the changes in `src/core/annotation.js` since it doesn't feel right to add a document-wide property on annotation instances (not the boolean itself being the problem, but merely the location of the data being out of place). Annotation instances should only contain information about the annotations themselves, and `hasJSActions` is not a property of the annotation but of the document. A solution along the lines of `enableScripting`, which also isn't included in every annotation instance, was merely a direction that could be thought about to move that data out of the annotation instances, but that wasn't really clear from my side; sorry for that.

timvandermeij (approved):

timvandermeij:
Looks good to me too now. Thank you!

PR #12552, opened 2020-10-29
Miscellaneous (small) improvements in `src/core/annotation.js`

Snuffleupagus:

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -25,6 +25,7 @@ import {
   escapeString,
   getModificationDate,
   isString,
+  objectSize,
   OPS,
   shadow,
   stringToPDFString,
@@ -1460,18 +1461,20 @@ class WidgetAnnotation extends Annotation {
     if (dict.has("AA")) {
       const additionalActions = dict.get("AA");
       for (const key of additionalActions.getKeys()) {
-        if (key in AnnotationActionEventType) {
-          const actionDict = additionalActions.getRaw(key);
-          const parents = new RefSet();
-          const list = [];
-          this._collectJS(actionDict, xref, list, parents);
-          if (list.length > 0) {
-            actions[AnnotationActionEventType[key]] = list;
-          }
+        const action = AnnotationActionEventType[key];
+        if (!action) {
+          continue;
+        }
+        const actionDict = additionalActions.getRaw(key);
+        const parents = new RefSet();
+        const list = [];
+        this._collectJS(actionDict, xref, list, parents);
+        if (list.length > 0) {
+          actions[action] = list;
         }
       }
     }
-    // Collect the Action if any (we may have one on pushbutton)
+    // Collect the Action if any (we may have one on pushbutton).
     if (dict.has("A")) {
       const actionDict = dict.get("A");
       const parents = new RefSet();
@@ -1481,7 +1484,7 @@ class WidgetAnnotation extends Annotation {
         actions.Action = list;
       }
     }
-    return actions;
+    return objectSize(actions) > 0 ? actions : null;
   }
 
   getFieldObject() {
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -1326,7 +1326,7 @@ class WidgetAnnotation extends Annotation {
 
   _computeFontSize(font, fontName, fontSize, height) {
     if (fontSize === null || fontSize === 0) {
-      const em = font.charsToGlyphs("M", true)[0].width / 1000;
+      const em = font.charsToGlyphs("M")[0].width / 1000;
       // According to https://en.wikipedia.org/wiki/Em_(typography)
       // an average cap height should be 70% of 1em
       const capHeight = 0.7 * em;
@@ -1597,7 +1597,7 @@ class TextWidgetAnnotation extends WidgetAnnotation {
     }
 
     const scale = fontSize / 1000;
-    const whitespace = font.charsToGlyphs(" ", true)[0].width * scale;
+    const whitespace = font.charsToGlyphs(" ")[0].width * scale;
     const chunks = [];
 
     let lastSpacePos = -1,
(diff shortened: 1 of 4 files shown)


src/core/annotation.js
```diff
@@ -1456,32 +1457,34 @@ class WidgetAnnotation extends Annotation {
   }
 
   _collectActions(xref, dict) {
-    const actions = Object.create(null);
+    const actionsMap = new Map();
```
calixteman, line 1460: I already added the null actions in this patch:
https://github.com/mozilla/pdf.js/pull/12546

Anyway, I think we mustn't change the type of actions because this object must be able to go into the sandbox.
It's fine with the firefox `Cu.Sandbox` but for the quickjs one (or whatever) we'll need to serialize the data to pass them into the sandbox as a string, my idea is to use `JSON.stringify` and so in this case a Map won't be serialized.

Snuffleupagus: > I already added the null actions in this patch:
> #12546 

Sorry, I hadn't seen that one yet!
However, personally I'd very much prefer a solution that doesn't require *manually* tracking whether or not something has been added to an Object.

Also, based on a quick look I'm not entirely crazy about the implementation there, since rendering starting in the default viewer is now conditional on something *completely unrelated* to the actual rendering of pages. Let me think about a possible way that we can avoid that!

> Anyway, I think we mustn't change the type of actions because this object must be able to go into the sandbox.

But it doesn't actually change the return type, see https://github.com/mozilla/pdf.js/pull/12552/files#diff-a7e409abd3d3209d32d3adc8457c5e23feb4c06113a2e065b0c6a9322a3a3369R1487 where either an Object or `null` is now returned instead.

calixteman: > > I already added the null actions in this patch:
> > #12546
> 
> Sorry, I hadn't seen that one yet!
> However, personally I'd very much prefer a solution that doesn't require _manually_ tracking whether or not something has been added to an Object.
> 
> Also, based on a quick look I'm not entirely crazy about the implementation there, since rendering starting in the default viewer is now conditional on something _completely unrelated_ to the actual rendering of pages. Let me think about a possible way that we can avoid that!
> 
The goal is to avoid adding useless listeners and I've another patch of the same kind to avoid adding some listeners when the calculationOrder array is empty but let chat about that in the PR.

> > Anyway, I think we mustn't change the type of actions because this object must be able to go into the sandbox.
> 
> But it doesn't actually change the return type, see https://github.com/mozilla/pdf.js/pull/12552/files#diff-a7e409abd3d3209d32d3adc8457c5e23feb4c06113a2e065b0c6a9322a3a3369R1487 where either an Object or `null` is now returned instead.

pfff... I read the patch too quickly, I missed that, sorry.
Isn't it overkill to create and fill a Map to then create and fill an Object ? and in term of perfs ? I've the impress that's almost duplicating work.

Snuffleupagus: > Isn't it overkill to create and fill a Map to then create and fill an Object ? and in term of perfs ?

I was under the impression that `Map`s should be reasonable efficient to use, and in *most* PDF documents there probably won't be any Annotations with JS actions anyway. It's very possible that this is indeed slightly less efficient, but the question is how much (if at all) that matters in the grand scheme of things!?
(As mentioned, I simply wanted a solution that didn't require manually tracking things being inserted into an Object.)

> I've the impress that's almost duplicating work.

You could very well be right, I don't know :-)

calixteman: I ran the following code:
```javascript
var tests=100000;
function m1() {
  var m=new Map();
  for (const c of "abcdefg") {
    m.set(c, c);
  }
  return Object.assign(Object.create(null), Object.fromEntries(m));
};
function m2() {
  var m=Object.create(null);
  for (const c of "abcdefg") {
      m[c]=c;
  }
  return m;
};
console.time('m1');
var i=tests;
while (i--) {
  [1,2,3].forEach(function() {
    return m1();
  });
}
console.timeEnd('m1');
console.time('m2');
i=tests;
while (i--) {
  [1,2,3].forEach(function() {
    return m2();
  });
}
console.timeEnd('m2');
```
on https://playcode.io and the Map m1 is around 5 times slower than m2 with Firefox nightly and around 4 times slower with chrome.
I agree on the fact that there aren't so many js actions and this code isn't called so many times... so it isn't really time consuming.
So if the code clarity matters more than perfs and resource consumption, then let's do that.
My opinion is that every nanosec we win is either useful to do something else or useful to not waste resources.

Snuffleupagus: > [...] and the Map m1 is around 5 times slower than m2 with Firefox nightly and around 4 times slower with chrome.

That's very interesting, the difference is a lot larger than I imagined it would be.

Given that I still would like to avoid *manually* tracking things in these types of situations, since it feels way to easy to accidentally introduce bugs that way, I'm thus proposing an alternative solution whereby we still use Objects but check their "size" before returning the data.

> I agree on the fact that there aren't so many js actions and this code isn't called so many times... so it isn't really time consuming.

That was my idea, until I saw the performance difference.

> So if the code clarity matters more than perfs and resource consumption, then let's do that.

Obviously both matter, but with the updated patch I think that the trade-off is now OK.

---

I've tested the following code, using the same service as you did, and the difference is now small enough that I think it's acceptable.
```javascript
function objectSize(obj) {
  return Object.keys(obj).length;
}

var tests=100000;
function m1() {
  const m=Object.create(null);
  for (const c of "abcdefg") {
      m[c]=c;
  }
  return objectSize(m) > 0 ? m : null;
};
function m2() {
  const m=Object.create(null);
  for (const c of "abcdefg") {
      m[c]=c;
  }
  return m;
};
console.time('m1');
var i=tests;
while (i--) {
  [1,2,3].forEach(function() {
    return m1();
  });
}
console.timeEnd('m1');
console.time('m2');
i=tests;
while (i--) {
  [1,2,3].forEach(function() {
    return m2();
  });
}
console.timeEnd('m2');
```

timvandermeij (approved):

timvandermeij:
Looks good!

PR #12555, opened 2020-10-30
Replace css color rgb(...) by #...

calixteman:
* it's faster to generate the color code in using a table for components
* it's very likely a way faster to parse (when setting the color in the canvas)

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -618,16 +618,13 @@ const IsEvalSupportedCached = {
   },
 };
 
-const rgbBuf = ["rgb(", 0, ",", 0, ",", 0, ")"];
+const hexNumbers = [...Array(256).keys()].map(n =>
+  n.toString(16).padStart(2, "0")
+);
 
 class Util {
-  // makeCssRgb() can be called thousands of times. Using ´rgbBuf` avoids
-  // creating many intermediate strings.
-  static makeCssRgb(r, g, b) {
-    rgbBuf[1] = r;
-    rgbBuf[3] = g;
-    rgbBuf[5] = b;
-    return rgbBuf.join("");
+  static makeHexColor(r, g, b) {
+    return `#${hexNumbers[r]}${hexNumbers[g]}${hexNumbers[b]}`;
   }
 
   // Concatenates two transformation matrices together and returns the result.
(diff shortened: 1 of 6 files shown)


Snuffleupagus:
Interesting find; should the cases in https://github.com/mozilla/pdf.js/blob/master/src/display/svg.js be converted as well and possibly the `Util.makeCssRgb` method removed if becomes unused?

src/shared/util.js
```diff
@@ -618,16 +618,13 @@ const IsEvalSupportedCached = {
...
 class Util {
-  // makeCssRgb() can be called thousands of times. Using ´rgbBuf` avoids
-  // creating many intermediate strings.
-  static makeCssRgb(r, g, b) {
-    rgbBuf[1] = r;
-    rgbBuf[3] = g;
-    rgbBuf[5] = b;
-    return rgbBuf.join("");
+  static makeHexColor(r, g, b) {
+    return "#" + hexNumbers[r] + hexNumbers[g] + hexNumbers[b];
```
timvandermeij, line 627: I think we should then change this to a template string because currently this creates quite a few intermediate strings everytime it's called, so this:

```javascript
return `#${hexNumbers[r]}${hexNumbers[g]}${hexNumbers[b]}`;
```

should fix that.

timvandermeij (commented):
Looks good, but this really needs a benchmark since this code is called very often and the question is what the effect of this change is; see https://github.com/mozilla/pdf.js/wiki/Benchmarking-your-changes for how to do that, and make sure to use different PDF files from the `test/pdf` folder for good coverage (especially image-heavy ones or other files that trigger this utility function often).

calixteman:
I ran benchmarks with this file:
[foo.pdf](https://github.com/mozilla/pdf.js/files/5469056/foo.pdf)
I generated using a python script.
In this file, `makeHexColors` is called 406,890 times and the perf comparison before/after the change:
```
firefox | Overall      |    50 |         4578 |        4509 | -69 | -1.51 |        faster
firefox | Page Request |    50 |            6 |           8 |   2 | 25.81 |              
firefox | Rendering    |    50 |         4571 |        4501 | -71 | -1.54 |        faster
```

timvandermeij:
Good improvement! If you can fix the comment above and re-run the benchmark with that fix applied, I'll also trigger the tests here.

calixteman:
I already ran the benchmark with the template string.

timvandermeij (approved):

timvandermeij:
Looks good!

/botio makeref

SamyCookie:
just for information see also this https://stackoverflow.com/a/55200387 that also approve the idea to make a pre created table for performance.

PR #12559, opened 2020-10-31
Also update the browser history when the user *manually* change pages using the pageNumber-input (PR 12493 follow-up)

Snuffleupagus:
This patch addresses a review comment, which pointed out that we should *also* handle the pageNumber-input, from PR https://github.com/mozilla/pdf.js/pull/12493#issuecomment-712405285.

Given that a user *manually* changing pages using the pageNumber-input, on the toolbar, could be regarded as a pretty strong indication of user-intent w.r.t. navigation in the document, hence I suppose that updating the browser history in this case as well probably won't hurt.


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

Issue #11965, opened 2020-06-03
Go over findings from static analyzers


timvandermeij:
https://lgtm.com/projects/g/mozilla/pdf.js?mode=list is interesting and might provide PDF.js with some opportunities to improve the code.

Snuffleupagus:
> https://lgtm.com/projects/g/mozilla/pdf.js?mode=list

The problem with that one, the last time I looked, was that it's fairly opinionated and that there doesn't appear to be a great way of ignoring *valid* patterns on e.g. a case-by-case basis.

Snuffleupagus:
Recently LGTM has actually been helpful in flagging a couple of *valid* logical mistakes in new code, so I'm wondering if we should actually reconsider the decision to use its PR integration?

Looking e.g. at https://lgtm.com/help/lgtm/github-apps-integration, the setup doesn't look particularly difficult.

Snuffleupagus:
> so I'm wondering if we should actually reconsider the decision to not use its PR integration?

Given that no objections have been raised here, I'll try to get the LGTM PR-integration setup sometime this week (assuming I've got the required permissions).

timvandermeij:
Sorry, I forgot to reply. I agree that LGTM can be useful to integrate now that it has given some useful results and the false positives are (largely) silenced.

Snuffleupagus:
As I (somewhat) suspected, enabling the LGTM PR-integration seem to require repository *ownership* since simple commit access doesn't appear to be enough.

@brendandahl When you've got time, mind following https://lgtm.com/help/lgtm/github-apps-integration to setup the LGTM PR-integration for this repository.

Snuffleupagus:
Looking e.g. at https://github.com/mozilla/pdf.js/pull/13761/checks or https://github.com/mozilla/pdf.js/pull/13754/checks, it seems that LGTM has now been setup and that it's working :-)

PR #12561, opened 2020-11-01
Fix some static analyzer warnings (issue 11965)

Snuffleupagus:
This fixes only those warnings, as reported by https://lgtm.com/projects/g/mozilla/pdf.js?mode=list, that make sense (as far as I'm concerned).

Hence this patch leaves the following things unaddressed:
 - The "recommendation"-category, since it only complains about unused variables. However, note that all of those cases are purposely included and that there's thus ESLint-disable comments added to explictly allow them.
 - The "warning"-category, which still contains two complaints. However, as far as I can tell, they are both false positives.

Given first of all the false positives of the LGTM static analyzer, and secondly that we'd need to add (essentially duplicated) disable-comments for the unused variable cases, it's not entirely clear to me if we actually want to work towards including LGTM in the PDF.js project (e.g. running alongside Travis) or if we should just close issue #11965.

### DIFF
--- a/src/shared/fonts_utils.js
+++ b/src/shared/fonts_utils.js
@@ -120,7 +120,7 @@ function readFontDictData(aString, aMap) {
   var fontDictDataTokens = [];
 
   var count = aString.length;
-  for (var i = 0; i < count; i) {
+  for (var i = 0; i < count; ) {
     var value = aString[i++] | 0;
     var token = null;
 
--- a/web/chromecom.js
+++ b/web/chromecom.js
@@ -418,7 +418,7 @@ class ChromeExternalServices extends DefaultExternalServices {
   }
 
   static createDownloadManager(options) {
-    return new DownloadManager(options);
+    return new DownloadManager();
   }
 
   static createPreferences() {
--- a/web/debugger.js
+++ b/web/debugger.js
@@ -381,7 +381,6 @@ var Stepper = (function StepperClosure() {
         line.appendChild(c("td", JSON.stringify(simplifyArgs(decArgs))));
       }
       if (operatorsToDisplay < operatorList.fnArray.length) {
-        line = c("tr");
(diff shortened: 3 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Thank you! I'd say this is enough to close the issue. If we ever want to do more with this, we can always do so.

automatedbugreportingfacility:
> However, as far as I can tell, they are both false positives.

I don't think the useless comparison test in `/src/display/pattern_helper.js` is a false positive. At line 121, `y1 === y2` is evaluated only when both these conditions are met:

1) At line 117, `y < y2` evaluates to true
2) At line 119, `y < y1` evaluates to false

If `y` is lower than `y2` and not lower than `y1`, then `y1` cannot be equal to `y2` (unless any of those variables can be modified between the comparisons, which is not the case here).

Snuffleupagus:
> 
> 
> > However, as far as I can tell, they are both false positives.
> 
> I don't think the useless comparison test in `/src/display/pattern_helper.js` is a false positive. At line 121, `y1 === y2` is evaluated only when both these conditions are met:
> 
>     1. At line 117, `y < y2` evaluates to true
> 
>     2. At line 119, `y < y1` evaluates to false
> 
> 
> If `y` is lower than `y2` and not lower than `y1`, then `y1` cannot be equal to `y2` (unless any of those variables can be modified between the comparisons, which is not the case here).

That's certainly an interesting find; thank you!

However, since that we're currently not intending on integrating LGTM officially for this repo, we can probably just leave this as-is for now. Given that there's *similar* code in the `else`-branch below (with a valid `y2 === y3` check), I do wonder if there's perhaps a bug in the code as-is, however it seems better to just leave it alone for now since we're (currently) not aware of any PDF documents that'd benefit from (general) changes to this code.

PR #12564, opened 2020-11-03
Make sure that Popup is rendered next to trigger for textAnnotation

hakubo:
Popup for textAnnotations can be rendered in a wrong position. 
Example: 
[abc (2).pdf](https://github.com/mozilla/pdf.js/files/5479588/abc.2.pdf)

After this fix all popup annotations are rendered in a correct place.


timvandermeij:
Looks good, but please add a reference test for this to prevent future regressions. You can use https://github.com/mozilla/pdf.js/commit/34918a66663f0da452288e286de0dd5496b383e2 as an example for how to do that. In short:

- Put the file from the PR description in the `test/pdfs` folder as `pr12564.pdf`.
- Add this same file to `test/pdfs/.gitignore` to be able to commit it.
- Add an entry to `test/test_manifest` that describes your test, similar to https://github.com/mozilla/pdf.js/commit/34918a66663f0da452288e286de0dd5496b383e2#diff-9bc50ce4ba8a4ba8d322114a3cc05fa62005a40b52223059200ca07087dd5a61. In you case also add `"firstPage": 1` and `"lastPage": 1` to the entry because we only need to test the first page since the problem already occurs there.

hakubo:
@timvandermeij sure!
I compressed the pdf a bit so it's a bit smaller in the repo.

Thanks for looking into it.

test/test_manifest.json
```diff
@@ -4784,5 +4784,14 @@
...
        "lastPage": 1,
        "type": "fbf"
-    }
+    },
+    {  "id": "pr12564",
+      "file": "pdfs/pr12564.pdf",
+      "md5": "8033857922eb03d050f6a04c2f5e9851",
+      "rounds": 1,
+      "firstPage": 1,
+      "lastPage": 1,
```
Snuffleupagus, line 4793: When you're adding the test-case to the repository (as opposed to linking it), and it *only* contains one page, you don't really need to add those lines :-)

Snuffleupagus (commented):
Also, please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

test/test_manifest.json
hakubo: sure. Fixed.

hakubo:
@Snuffleupagus squished to one commit! :)

test/test_manifest.json
```diff
@@ -4784,5 +4784,12 @@
...
        "type": "fbf"
-    }
+    },
+    {  "id": "pr12564",
+      "file": "pdfs/pr12564.pdf",
+      "md5": "8033857922eb03d050f6a04c2f5e9851",
+      "rounds": 1,
+      "type": "eq",
+      "annotations": true
+   },
```
Snuffleupagus, line 4794: This comma causes the tests to fail, since it's not valid JSON:
```
### Running unit tests
undefined:4795
]
^

SyntaxError: Unexpected token ] in JSON at position 129429
    at JSON.parse (<anonymous>)
    at getTestManifest (C:\pdfjs\botio-files-pdfjs\private\739c7c0473bff57\test\test.js:329:23)
    at ensurePDFsDownloaded (C:\pdfjs\botio-files-pdfjs\private\739c7c0473bff57\test\test.js:888:18)
    at main (C:\pdfjs\botio-files-pdfjs\private\739c7c0473bff57\test\test.js:918:5)
    at Object.<anonymous> (C:\pdfjs\botio-files-pdfjs\private\739c7c0473bff57\test\test.js:935:1)
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
```

Please make sure that patches pass all tests *locally*, before submitting a patch; please find additional details at https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing

hakubo: omg, sorry about that. 
Fixed.

hakubo:
looks I don't have that kind of power to run tests :)

Snuffleupagus:
Once more, please make sure that your patch passes *all tests locally*; more information at https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing

There's no point in attempting to run tests here again, until you've confirmed that your patch passes all tests locally!

timvandermeij:
This PR needs a rebase onto the current master. Unfortunately another patch also touched the same code, but on a positive note it already added the `popupTop` variable, which should make this patch even smaller.

Moreover, I think we should also add the top change to https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js#L303 so it also works properly for annotations without explicit popup annotations.

hakubo:
@timvandermeij rebased on current master.

If I get it right, https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js#L303 this renders popup within annotation section element. To these are positioned correctly as they will be positioned relative to the annotation section element.

timvandermeij (approved):

timvandermeij:
Thank you for your contribution! (The reference test failures are not relevant to this PR.)

/botio makeref

PR #12567, opened 2020-11-03
[api-minor] JS -- hidden annotations must be built in case a script show them

calixteman:
* in some pdf, there are actions with "event.source.hidden = ..."
 * in order to handle visibility when printing, annotationStorage is extended to store multiple properties (value, hidden, editable, ...)


Snuffleupagus:
Given that this changes the format used internally in the `AnnotationStorage`, it's probably best to tag this PR with `[api-minor]` just in case someone has started to depend on it; that way it's easier to spot PRs with breaking changes when looking at https://github.com/mozilla/pdf.js/releases

brendandahl:
Are you expecting the changes in `issue4914-page1` and `ZapfDingbats-page1`? I'm not sure what behavior we want since chrome/acrobat/preview all seem to show different things.

PR #12568, opened 2020-11-03
[api-minor] JS -- Add default value in annotation data

calixteman:
These values are used when a form is reset.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -951,6 +951,13 @@ class WidgetAnnotation extends Annotation {
     });
     data.fieldValue = this._decodeFormValue(fieldValue);
 
+    const defaultFieldValue = getInheritableProperty({
+      dict,
+      key: "DV",
+      getArray: true,
+    });
+    data.defaultFieldValue = this._decodeFormValue(defaultFieldValue);
+
     data.alternativeText = stringToPDFString(dict.get("TU") || "");
     data.defaultAppearance =
       getInheritableProperty({ dict, key: "DA" }) ||
@@ -1652,6 +1659,7 @@ class TextWidgetAnnotation extends WidgetAnnotation {
     return {
       id: this.data.id,
       value: this.data.fieldValue,
+      defaultValue: this.data.defaultFieldValue,
       multiline: this.data.multiLine,
       password: this.hasFieldFlag(AnnotationFieldFlag.PASSWORD),
       charLimit: this.data.maxLen,
@@ -1976,6 +1984,7 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     return {
       id: this.data.id,
       value,
+      defaultValue: this.data.defaultFieldValue,
       editable: !this.data.readOnly,
       name: this.data.fieldName,
       rect: this.data.rect,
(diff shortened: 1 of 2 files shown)


src/core/annotation.js
```diff
@@ -1652,6 +1659,7 @@ class TextWidgetAnnotation extends WidgetAnnotation {
     return {
       id: this.data.id,
       value: this.data.fieldValue,
+      defaultValue: this.data.defaultFieldValue || "",
```
timvandermeij, line 1662: Like the other cases below, the `|| ""` part should be removed here. It's optional according to the spec and will be `null` if it's not available, which is consistent with the API.

src/core/annotation.js
```diff
@@ -951,6 +951,13 @@ class WidgetAnnotation extends Annotation {
     });
     data.fieldValue = this._decodeFormValue(fieldValue);
 
+    const defaultFieldValue = getInheritableProperty({
+      dict,
+      key: "DV",
+      getArray: true,
+    });
+    data.defaultFieldValue = this._decodeFormValue(defaultFieldValue);
```
timvandermeij, line 959: Please add unit test coverage for this property. I'd recommend to simply extend the existing unit tests for `fieldValue` to also check `defaultFieldValue`, and set a value for it per annotation type tested.

src/core/annotation.js
```diff
@@ -1966,16 +1974,21 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
   getFieldObject() {
     let type = "button";
     let value = null;
+    let exportValues;
     if (this.data.checkBox) {
       type = "checkbox";
       value = this.data.fieldValue && this.data.fieldValue !== "Off";
+      exportValues = this.data.exportValue;
```
timvandermeij, line 1981: I believe this is incorrect, since according to https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=429&zoom=auto,-207,718 the export values for the field object must always be an array of strings. I recomment to change `let exportValues;` above to `let exportValues = []` since it doesn't seem to be optional and in these branches simply push the applicable values to the array (they are guaranteed to be strings). That way this field is also valid (empty) for e.g., push buttons.

Moreover, from the spec descrption I _think_ for checkboxes the list needs to include _all_ export values. `this.data.exportValue` is only the "not Off" value, so shouldn't this be `exportValues = ["Off", this.data.exportValue]`? Please check what e.g. Adobe Reader expects here.

Finally, please move the export value logic to a second commit since it's separate from adding default values.

timvandermeij (commented):

src/display/annotation_layer.js
```diff
@@ -468,6 +468,8 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         element.setAttribute("value", textContent);
       }
 
+      element.style.transform = "rotate(-90deg)";
```
timvandermeij, line 471: The additions in this file don't seem correct. Why was this file changed?

timvandermeij (approved):
Looks good with this final commit addressed, and passing tests.

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #12569, opened 2020-11-03
JS -- Fix events dispatchment and add tests

calixteman:
* dispatch event to take into account calculation order
 * use a map for actions in Field

### DIFF
--- a/src/scripting_api/event.js
+++ b/src/scripting_api/event.js
@@ -26,14 +26,14 @@ class Event {
     this.richChange = data.richChange || [];
     this.richChangeEx = data.richChangeEx || [];
     this.richValue = data.richValue || [];
-    this.selEnd = data.selEnd || 0;
-    this.selStart = data.selStart || 0;
+    this.selEnd = data.selEnd || -1;
+    this.selStart = data.selStart || -1;
     this.shift = data.shift || false;
     this.source = data.source || null;
     this.target = data.target || null;
-    this.targetName = data.targetName || "";
+    this.targetName = "";
     this.type = "Field";
-    this.value = data.value || null;
+    this.value = data.value || "";
     this.willCommit = data.willCommit || false;
   }
 }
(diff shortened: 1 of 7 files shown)


src/scripting_api/initialization.js
```diff
@@ -21,7 +21,7 @@ import { Field } from "./field.js";
 import { ProxyHandler } from "./proxy.js";
 import { Util } from "./util.js";
 
-function initSandbox(data, extra, out) {
+function initSandbox(data, extra, out, testMode = false) {
```
Snuffleupagus, line 24: Given the increasing number of parameters here, it might not be a bad idea to change this to take a parameter object instead!?

src/scripting_api/initialization.js
```diff
@@ -52,6 +52,10 @@ function initSandbox(data, extra, out) {
       out[name] = aform[name].bind(aform);
     }
   }
+
+  if (testMode) {
+    out._app = app;
+  }
```
Snuffleupagus, line 58: Do you perhaps want to *explicitly* prevent this from being set *unless* testing is running?
If so, the following prevent the parameter being respected in "regular" builds:
```suggestion
  if (
    (typeof PDFJSDev === "undefined" ||
      PDFJSDev.test("!PRODUCTION || TESTING")) &&
    testMode
  ) {
    out._app = app;
  }
```

src/scripting_api/event.js
```diff
@@ -56,23 +71,115 @@ class EventDispatcher {
...
     event.name = eventName;
+    event.targetName = target.obj.name;
     event.rc = true;
-    if (!target.obj._runActions(event)) {
-      return true;
+
+    return target.obj._runActions(event);
+  }
+
+  calculateNow() {
```
brendandahl, line 152: Is this used?

brendandahl (changes requested):

src/scripting_api/field.js
```diff
@@ -89,31 +82,56 @@ class Field extends PDFObject {
...
+            // eslint-disable-next-line no-new-func
+            Function("event", `with (this) {${action}}`).bind(doc)
+          )
+        );
+      }
+    }
+    return actionsMap;
+  }
+
+  _isButton() {
```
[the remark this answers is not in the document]
calixteman: Yes especially for radio buttons.

src/scripting_api/event.js
calixteman: From here https://github.com/mozilla/pdf.js/blob/master/src/scripting_api/doc.js#L572.
and for the spec:
https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=277

PR #12572, opened 2020-11-04
Change the `getVisibleElements` helper function to take a parameter object

Snuffleupagus:
Given the number of parameters, and the fact that many of them are booleans, the call-sites are no longer particularly easy to read and understand. Furthermore, this slightly improves the formatting of the JSDoc-comment, since it needed updating as part of these changes anyway.

Finally, this removes an unnecessary `numViews === 0` check from `getVisibleElements`, since that should be *very* rare and more importantly that the `binarySearchFirstItem` function already has a fast-path for that particular case.


timvandermeij (approved):

timvandermeij:
Thanks! I really like this improvement.

PR #12575, opened 2020-11-04
Fix intermittent reference failures.

brendandahl:
It's not ideal to use a new canvas for every PDF, but this makes the tests
reliable which I think is more important.


brendandahl:
The above is missing a few things that causes blank canvases. Will close for now to until I have time to update.

PR #12582, opened 2020-11-05
JS -- Implement doc object

calixteman:
* https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=335
 * it has all the properties/methods defined in the spec
 * unimplemented methods are there but with an empty body to avoid exception when calling an undefined method
 * implement zoom, zoomType, layout, pageNum, ...

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1350,11 +1350,11 @@ const PDFViewerApplication = {
         );
         this._idleCallbacks.add(callback);
       }
+      this._initializeJavaScript(pdfDocument);
     });
 
     this._initializePageLabels(pdfDocument);
     this._initializeMetadata(pdfDocument);
-    this._initializeJavaScript(pdfDocument);
   },
 
   /**
@@ -1370,25 +1370,46 @@ const PDFViewerApplication = {
       return;
     }
     const scripting = this.externalServices.scripting;
+    const {
+      info,
+      metadata,
+      contentDispositionFilename,
+    } = await pdfDocument.getMetadata();
 
-    window.addEventListener("updateFromSandbox", function (event) {
+    window.addEventListener("updateFromSandbox", event => {
       const detail = event.detail;
       const id = detail.id;
       if (!id) {
         switch (detail.command) {
-          case "println":
-            console.log(detail.value);
-            break;
-          case "clear":
-            console.clear();
-            break;
           case "alert":
             // eslint-disable-next-line no-alert
             window.alert(detail.value);
             break;
+          case "clear":
+            console.clear();
+            break;
           case "error":
             console.error(detail.value);
             break;
+          case "layout":
+            this.pdfViewer.spreadMode = apiPageLayoutToSpreadMode(detail.value);
+            return;
+          case "page-num":
+            this.pdfViewer.currentPageNumber = detail.value + 1;
+            return;
+          case "print":
+            this.triggerPrinting();
+            return;
+          case "println":
+            console.log(detail.value);
+            break;
+          case "zoom":
+            if (typeof detail.value === "string") {
+              this.pdfViewer.currentScaleValue = detail.value;
+            } else {
+              this.pdfViewer.currentScale = detail.value;
+            }
+            return;
         }
         return;
       }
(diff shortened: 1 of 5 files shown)


web/app.js
```diff
@@ -1410,7 +1433,18 @@ const PDFViewerApplication = {
...
+    scripting.createSandbox({
+      objects,
+      dispatchEventName,
+      calculationOrder,
+      docInfo: {
+        ...info,
+        metadata,
+        filename: contentDispositionFilename,
+        numPages: pdfDocument.numPages,
+        URL: window.location.href,
```
Snuffleupagus, line 1445: Does this actually do what you want/expect across different types of PDF.js builds?

web/app.js
```diff
@@ -1388,6 +1393,24 @@ const PDFViewerApplication = {
           case "error":
             console.error(detail.value);
             break;
+          case "print":
+            setTimeout(() => {
```
Snuffleupagus, line 1397: Why do we need the `setTimeout` here?

web/app.js
```diff
@@ -1369,8 +1369,13 @@ const PDFViewerApplication = {
...
     }
     const scripting = this.externalServices.scripting;
+    const {
+      info,
+      metadata,
+      contentDispositionFilename,
+    } = await pdfDocument.getMetadata();
 
-    window.addEventListener("updateFromSandbox", function (event) {
+    window.addEventListener("updateFromSandbox", event => {
```
Snuffleupagus, line 1378: Generally speaking, given that the `case`s you're adding below *could* interfere with initial loading of the document and setting of the initial position, you should ensure that none of this code will run *until* the document has loaded at enough that it's possible to interact safely with it.
Hence, please add the following type of check here (unless we can delay the entire `this._initializeJavaScript(pdfDocument)` call to make this point moot):
```javascript
if (!this.isInitialViewSet) {
  return;
}
```

Snuffleupagus (commented):
More generally, do we actually need to run `this._initializeJavaScript(pdfDocument);` as early as is done currently?
Is it absolutely impossible for things to work correctly if we e.g. waited for `onePageRendered`, as we do for a number of other API-calls, to prevent them running very early and thus possibly delay initial rendering?

For example: If you're able to delay `this._initializeJavaScript(pdfDocument);` slightly you'd not need to re-fetch the "metadata" again and could instead use the already cached data.

calixteman:
> More generally, do we actually need to run `this._initializeJavaScript(pdfDocument);` as early as is done currently?
> Is it absolutely impossible for things to work correctly if we e.g. waited for `onePageRendered`, as we do for a number of other API-calls, to prevent them running very early and thus possibly delay initial rendering?

Good point, as far as I understand the specs, there is nothing which indicates that some actions must be ran before rendering (some of them must ran on "Page open" or "Document open" events).
Anyway I guess that the delay between end of rendering and the moment the user begins to interact with the document is a way enough to load the sandbox stuff. 
@Snuffleupagus is it ok for you to move `this._initializeJavaScript(pdfDocument);` at the end of https://github.com/mozilla/pdf.js/blob/8b652d685b4e66ea95dc036bd7aa2a200766ceb5/web/app.js#L1330 ?

> 
> For example: If you're able to delay `this._initializeJavaScript(pdfDocument);` slightly you'd not need to re-fetch the "metadata" again and could instead use the already cached data.

web/app.js
calixteman: Copy/paste of:
https://github.com/mozilla/pdf.js/blob/8b652d685b4e66ea95dc036bd7aa2a200766ceb5/web/app.js#L1473

Snuffleupagus:
> is it ok for you to move `this._initializeJavaScript(pdfDocument);` at the end of
> 
> https://github.com/mozilla/pdf.js/blob/8b652d685b4e66ea95dc036bd7aa2a200766ceb5/web/app.js#L1330

That's most definitely the preferred solution here, assuming if works of course :-)

web/app.js
```diff
@@ -1388,6 +1393,22 @@ const PDFViewerApplication = {
...
+            } else {
+              this.pdfViewer.currentScale = detail.value;
+            }
+            return;
+          case "layout":
+            this.pdfViewer.spreadMode = apiPageLayoutToSpreadMode(detail.value);
+            return;
+          case "page-num":
+            this.pdfViewer.currentPageNumber = detail.value + 1;
+            return;
```
Snuffleupagus, line 1411: Please make sure that you sort all of these `case`s *alphabetically*, since that will make it much easier to find things as the number of cases grow.

web/app.js
```diff
@@ -1410,7 +1431,19 @@ const PDFViewerApplication = {
...
     const calculationOrder = [];
-    scripting.createSandbox({ objects, dispatchEventName, calculationOrder });
+    scripting.createSandbox({
+      objects,
+      dispatchEventName,
+      calculationOrder,
+      docInfo: {
+        ...info,
+        metadata,
+        filesize: this.pdfDocumentProperties.maybeFileSize,
```
Snuffleupagus, line 1441: This property is first of all not guaranteed to even be set at this point in time and/or it may not actually be correct. Secondly, it's not really correct to reach into that component in this way, since that property was never intended to be "public" in the first place.

Snuffleupagus (changes requested):

brendandahl (approved):

brendandahl:
Looks like a good starting point. Now that we are interacting with the viewer I'd really like to see some tests when we have quickjs wired up.

Issue #6961, opened 2016-02-08
Major performance issues for a simple PDF file


timvandermeij:
I have created a simple PDF file using Scribus 1.5.0svn. Notice that the file size is large for a two-page PDF file, so I can only suspect that Scribus is doing something really unefficient when exporting the PDF file. Nevertheless, the PDF file below renders instantly with both Adobe Acrobat Reader DC and Foxit Reader (within 0.5 seconds), however PDF.js takes *27 seconds* to render only the first page of this file. I have no idea why PDF.js is taking such an excessive amount of time, but we need to do better here, given that other viewers do not have any problems with this file. Are there perhaps inefficient patterns in this file that the optimizer could remove? I notice that the PDF file contains an excessive amount of resources in the Resources dictionary of each page (XObject, Font, Pattern and ExtGState).

Below is the PDF file. I made this myself, so anyone is free to use this as a test case in a PR that addresses this issue:
[test.pdf](https://github.com/mozilla/pdf.js/files/121712/test.pdf)

Rob--W:
The CPU profiler shows that most of the time is spent in `PartialEvaluator_hasBlendModes`: 
https://github.com/mozilla/pdf.js/blob/a0aa781c394a8e5dedbf14c5db0195ab72e8ff46/src/core/evaluator.js#L144-L200

To profile (using Chrome), simply:

1. Download test.pdf from above.
2. Visit https://mozilla.github.io/pdf.js/web/viewer.html?file=
3. Open the developer tools, go to the Profiles tab.
4. Open test.pdf with PDF.js
5. Select "Target: pdf.worker.js" and click on Start to profile
6. Whenever you feel ready (e.g. when the CPU is idle), stop the capture.
7. Analyse the results.

timvandermeij:
Thank you for profiling this! This code seems to loop over (potentially) all ExtGState and XObject dictionaries, and there are *a lot* of them in this PDF file. I'm afraid the `in` operator might cause delays here.

timvandermeij:
I have narrowed down the flaw a bit. For the first page, https://github.com/mozilla/pdf.js/blob/a0aa781c394a8e5dedbf14c5db0195ab72e8ff46/src/core/evaluator.js#L182 is trigger almost 31000 times, meaning that we attempt to process *a lot* of XObjects that we already processed before.

Snuffleupagus:
[...]

PR #12583, opened 2020-11-05
Add global caching, for /Resources without blend modes, and use it to reduce repeated fetching/parsing in `PartialEvaluator.hasBlendModes`

Snuffleupagus:
The `PartialEvaluator.hasBlendModes` method is necessary to determine if there's any blend modes on a page, which unfortunately requires *synchronous* parsing of the /Resources of each page before its rendering can start (see the "StartRenderPage"-message).
In practice it's not uncommon for certain /Resources-entries to be found on more than one page (referenced via the XRef-table), which thus leads to unnecessary re-fetching/re-parsing of data in `PartialEvaluator.hasBlendModes`.

To improve performance, especially in pathological cases, we can cache /Resources-entries when it's absolutely clear that they do not contain *any* blend modes at all[1]. This way, subsequent `PartialEvaluator.hasBlendModes` calls can be made significantly more efficient.

This patch was tested using the PDF file from issue #6961, i.e. https://github.com/mozilla/pdf.js/files/121712/test.pdf:
```
[
    {  "id": "issue6961",
       "file": "../web/pdfs/issue6961.pdf",
       "md5": "a80e4357a8fda758d96c2c76f2980b03",
       "rounds": 100,
       "type": "eq"
    }
]
```

```
-- Grouped By browser, page, stat --
browser | page | stat         | Count | Baseline(ms) | Current(ms) |  +/- |     %  | Result(P<.05)
------- | ---- | ------------ | ----- | ------------ | ----------- | ---- | ------ | -------------
firefox | 0    | Overall      |   100 |         1034 |         555 | -480 | -46.39 |        faster
firefox | 0    | Page Request |   100 |          489 |           7 | -482 | -98.67 |        faster
firefox | 0    | Rendering    |   100 |          545 |         548 |    2 |   0.45 |
firefox | 1    | Overall      |   100 |          912 |         428 | -484 | -53.06 |        faster
firefox | 1    | Page Request |   100 |          487 |           1 | -486 | -99.77 |        faster
firefox | 1    | Rendering    |   100 |          425 |         427 |    2 |   0.51 |
```

---
[1] In the case where blend modes *are* found, it becomes a lot more difficult to know if it's generally safe to skip /Resources-entries. Hence we don't cache anything in that case, however note that most document/pages do not utilize blend modes anyway.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -76,6 +76,7 @@ class Page {
     fontCache,
     builtInCMapCache,
     globalImageCache,
+    nonBlendModesSet,
   }) {
     this.pdfManager = pdfManager;
     this.pageIndex = pageIndex;
@@ -85,6 +86,7 @@ class Page {
     this.fontCache = fontCache;
     this.builtInCMapCache = builtInCMapCache;
     this.globalImageCache = globalImageCache;
+    this.nonBlendModesSet = nonBlendModesSet;
     this.evaluatorOptions = pdfManager.evaluatorOptions;
     this.resourcesPromise = null;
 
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thank you for doing this!

PR #12585, opened 2020-11-05
Fix popup for highlights without popup (follow-up of #12505)

### LINKED ISSUE #12576: Fix highlight annotation regressions
Unfortunately there was an oversight in PR #12505 where some regressions were missed:

- The `annotation-highlight-without-appearance`, `issue7014` and `issue9552` reference tests no longer show the popup, and I can confirm that it also doesn't render in the viewer anymore. They most likely share the same root cause.
   - I have debugged this and found that, at least in the `annotation-highlight-without-appearance` case, the `_createPopup` line in `HighlightAnnotationElement` is triggered. The popup is in fact added to the container, but the annotation also has quadrilaterals and in that case those are returned/rendered and _not_ the container. The problem is therefore limited to annotations without an explicit popup annotation, otherwise it works because that logic queries all quadrilateral elements. Therefore, the solution here seems to be to change `_createPopup` to also take the quadrilaterals into account, similar to the other code.
- The quadrilaterals don't have the proper class name for highlight annotations anymore, causing them not to have a cursor pointer on hover. 
  - I have found the cause of this to be on line 1379 in `src/display/annotation_layer.js` where the class name is only set on the container and not on all quadrilaterals (because we only had the container before).

Once this is fixed, please add the PDF file from #12504 to the test suite as well to ensure that that case is also still working.

/cc @calixteman

calixteman:
* remove 1st param of _createPopup (almost useless for a method)
 * prepend popup div to avoid to have them on top of some highlights (and so "disable" partially mouse events)
 * add a ref test for issue #12504

Fixes #12576.


src/display/annotation_layer.js
```diff
@@ -1615,7 +1622,12 @@ class AnnotationLayer {
             parameters.div.appendChild(renderedElement);
           }
         } else {
-          parameters.div.appendChild(rendered);
+          if (element.popup) {
```
Snuffleupagus, line 1625: *Drive-by-comment:* I don't think you need to add a boolean for this, since `element instanceof PopupAnnotationElement` ought to work just as well :-)

src/display/annotation_layer.js
```diff
@@ -262,7 +262,9 @@ class AnnotationElement {
         quadPoint[1].y,
       ];
       this.data.rect = rect;
-      quadrilaterals.push(this._createContainer(ignoreBorder));
+      const quad = this._createContainer(ignoreBorder);
+      quad.className = "highlightAnnotation";
```
timvandermeij, line 266: The `_createQuadrilaterals` method is generic since multiple annotation types can use quadrilaterals, so this class name should not be set here. Fortunately, it can simply be fixed locally in the `HighlightAnnotationElement` class below.

Hence I'd suggest to revert all changes in this method and instead simply change the `render` method in the `HighlightAnntationElement` class to the following (untested):

```javascript
  render() {
    if (!this.data.hasPopup) {
      this._createPopup(null, this.data);
    }

    if (this.quadrilaterals) {
      this.quadrilaterals.forEach(quadrilateral => {
        quadrilateral.className = "highlightAnnotation";
      });
      return this.quadrilaterals;
    }

    this.container.className = "highlightAnnotation";
    return this.container;
  }
```

src/display/annotation_layer.js
```diff
@@ -885,7 +892,6 @@ class PopupElement {
     // disappear too. In that special case, hiding the wrapper suffices.
     this.hideElement = this.hideWrapper ? wrapper : this.container;
     this.hideElement.setAttribute("hidden", true);
-
```
timvandermeij, line 888: This change is unrelated, so please revert it (the whitespace here it intentional to separate the text annotation logic from the rest).

src/display/annotation_layer.js
```diff
@@ -1615,7 +1621,12 @@ class AnnotationLayer {
             parameters.div.appendChild(renderedElement);
           }
         } else {
-          parameters.div.appendChild(rendered);
+          if (element instanceof PopupAnnotationElement) {
+            // popup div mustn't be on top of an annotation
```
timvandermeij, line 1625: Let's change this comment to the following since it's slightly more descriptive as to why this situation should not happen: `Popup annotation elements should not be on top of other annotation elements to prevent interfering with mouse events.`

timvandermeij (commented):

Snuffleupagus:
Given that this fixes a regression, we probably want to make sure this lands before the next "Merge Date" (which according to https://wiki.mozilla.org/Release_Management/Calendar is in less than one week); @calixteman Do you have time to address the outstanding comments here?

src/display/annotation_layer.js
timvandermeij: @calixteman Hm, it looks like this comment was forgotten to be addressed. Would you mind following this up in a new PR to move this specific class name setting out of the generic function?

brendandahl: Sorry misread the comment, thought it had been addressed. Filed #12585

PR #12591, opened 2020-11-06
Improve the Pattern-detection in `CanvasGraphics.stroke`

Snuffleupagus:
The vast majority of the time, unless a Pattern is active, the `strokeColor`-property contains a "simple" colour value represented by a String. Hence it seems somewhat ridiculous to do a `hasOwnProperty` check on a String, and it's should thus be possible to improve things a tiny bit here.

Unfortunately using a simple `instanceof` check would only work for `TilingPattern`s, but not for the `ShadingIRs` given how they are implemented; see `src/display/pattern_helper.js`. (While that file could probably do with some clean-up, given the age of some of its code, that probably shouldn't happen here.)

Finally, the `this.type = "Pattern"`-property of the various Shadings/TilingPatterns were removed, since I cannot see why it's necessary when we can simply check for a `getPattern` method instead. Note that part of this code even pre-dates the main/worker-thread split, which probably in part explains why it looks the way it does.

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -1350,11 +1350,7 @@ const CanvasGraphics = (function CanvasGraphicsClosure() {
       // stroking alpha.
       ctx.globalAlpha = this.current.strokeAlpha;
       if (this.contentVisible) {
-        if (
-          strokeColor &&
-          strokeColor.hasOwnProperty("type") &&
-          strokeColor.type === "Pattern"
-        ) {
+        if (typeof strokeColor === "object" && strokeColor?.getPattern) {
           // for patterns, we transform to pattern space, calculate
           // the pattern, call stroke, and restore to user space
           ctx.save();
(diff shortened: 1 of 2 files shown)


src/display/canvas.js
```diff
@@ -1350,11 +1350,7 @@ const CanvasGraphics = (function CanvasGraphicsClosure() {
       // stroking alpha.
       ctx.globalAlpha = this.current.strokeAlpha;
       if (this.contentVisible) {
-        if (
-          strokeColor &&
-          strokeColor.hasOwnProperty("type") &&
-          strokeColor.type === "Pattern"
-        ) {
+        if (typeof strokeColor === "object" && strokeColor?.getPattern) {
```
Snuffleupagus, line 1353: The optional chaining used here replaces an explicit `strokeColor !== null` check.

In practice there shouldn't ever be a situation when the `strokeColor` is `null`, as far as I can tell, but nonetheless I don't think the above will hurt.

timvandermeij (approved):

timvandermeij:

PR #12594, opened 2020-11-07
Enable the ESLint `no-useless-escape` rule (PR 12551 follow-up)

Snuffleupagus:
Note that a number of these cases are covered by existing unit-tests, and a few others only matter for the development/build scripts.
Furthermore, I've also tried to the best of my ability to test each case *manually* to hopefully further reduce the likelihood of this patch introducing any bugs.

Please find additional details about the ESLint rule at https://eslint.org/docs/rules/no-useless-escape

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -111,6 +111,7 @@
     "no-useless-call": "error",
     "no-useless-catch": "error",
     "no-useless-concat": "error",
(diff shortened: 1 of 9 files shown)


timvandermeij (approved):

timvandermeij:
Nice improvement!

PR #12595, opened 2020-11-07
Convert files in the `src/display/`-folder to use optional chaining where possible

Snuffleupagus:
By using optional chaining, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining, it's possible to reduce unnecessary code-repetition in many cases.
Note that these changes also reduce the size of the *built* `pdf.js` file, when `SKIP_BABEL == true` is set, and for the `MOZCENTRAL` build-target that result in a `0.1%` filesize reduction from a simple and mostly mechanical code change.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1566,9 +1566,7 @@ class PDFPageProxy {
         return;
       }
     }
-    intentState.streamReader.cancel(
-      new AbortException(reason && reason.message)
-    );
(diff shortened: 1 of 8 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for doing this!

PR #12598, opened 2020-11-07
Fail early, in modern `GENERIC` builds, if `globalThis` isn't available (PR 11799 follow-up, issue 12596)

Snuffleupagus:
It probably doesn't hurt to explicitly check for `globalThis` as well, in addition to the existing checks.


timvandermeij (approved):

timvandermeij:
Looks like a good additional safeguard to me. Thanks!

PR #12600, opened 2020-11-08
working multi page pdf viewer example with pdfjs-dist

zacharykapank:


timvandermeij:
Unfortunately we can't accept this PR. We had such examples before, but they always ended up becoming out of date because we simply can't support them since our focus is on vanilla JavaScript and not on the many JavaScript frameworks out there. There was even a React example, but we had to remove it in #11864 for the reasons mentioned there.

Finally, there are projects like https://github.com/wojtekmaj/react-pdf that already provide React integration of PDF.js and can probably support this much better than we can. Thanks.

PR #12604, opened 2020-11-09
JS -- Add a sandbox based on quickjs

calixteman:
* quickjs-eval.js has been generated using https://github.com/calixteman/pdf.js.quickjs/
 * lazy load of sandbox code


calixteman:
We need to decide where https://github.com/calixteman/pdf.js.quickjs/ should live.

web/app.js
```diff
@@ -182,6 +182,31 @@ class DefaultExternalServices {
...
+      const promise = new Promise((resolve, reject) => {
+        const sandboxScript = document.createElement("script");
+        sandboxScript.src = "../build/pdf.sandbox.js";
+        sandboxScript.type = "text/javascript";
+        sandboxScript.async = true;
+        sandboxScript.onload = () => {
+          resolve();
+        };
+        document.head.appendChild(sandboxScript);
+      }).then(() => {
```
Snuffleupagus, line 195: Please use [`loadScript`](https://github.com/mozilla/pdf.js/blob/55f55f58594b9a6947fecaabf8ef4e3b02002023/src/display/display_utils.js#L531-L546) rather than essentially re-implementing that manually.

gulpfile.js
```diff
@@ -493,6 +521,20 @@ function makeRef(done, bot) {
   });
 }
 
+gulp.task("sandbox", function (done) {
```
Snuffleupagus, line 524: I'd suggest separating the `createScriptingBundle` from this gulp-task, and possible even remove this completely.

For the `mozcentral` task you added the `createScriptingBundle` to list existing list of tasks in https://github.com/mozilla/pdf.js/blob/55f55f58594b9a6947fecaabf8ef4e3b02002023/gulpfile.js#L1065-L1067
hence I suggest adding `createScriptingBundle` *and* `createSandboxBundle` to `buildGeneric` instead: https://github.com/mozilla/pdf.js/blob/55f55f58594b9a6947fecaabf8ef4e3b02002023/gulpfile.js#L723-L729

.eslintignore
```diff
@@ -7,6 +7,7 @@ external/webL10n/
 external/cmapscompress/
 external/builder/fixtures/
 external/builder/fixtures_esprima/
+src/scripting_api/quickjs-eval.js
```
Snuffleupagus, line 10: Unless we can somehow add an external dependency, rather than simply dumping the code into PDF.js repository, anything *specific* to quickjs should probably go into a `external/quickjs` (or similar) folder instead.

Snuffleupagus (commented):
Besides a couple of comments based on a quick look, I've got a few more high-level questions:
 1. How's the quickjs-project licensed? Is it compatible with "Apache License 2.0", as that's what the PDF.js library uses?
 2. How will the quickjs-files, once added to the PDF.js library, be updated?
The reason for asking, is that we used to have loads of dependencies in the `external/` folder, which made updating and maintaining them very difficult and time-consuming (why we switched to `npm` for most dependencies).
 3. Would it be at all possible to pull in the quickjs-files as a dependency, rather than simply dumping them into the PDF.js repository as-is?

calixteman:
> Besides a couple of comments based on a quick look, I've got a few more high-level questions:
> 
>     1. How's the quickjs-project licensed? Is it compatible with "Apache License 2.0", as that's what the PDF.js library uses?

According to https://raw.githubusercontent.com/bellard/quickjs/master/quickjs.c it's a MIT licence so it's ok with Apache License 2.0.

> 
>     2. How will the quickjs-files, once added to the PDF.js library, be updated?
>        The reason for asking, is that we used to have loads of dependencies in the `external/` folder, which made updating and maintaining them very difficult and time-consuming (why we switched to `npm` for most dependencies).

The easiest solution is the best here... so tell me what you prefer having.
I put the build stuff in a Docker image to avoid to have to set `emsdk` to build it.
I guess we can change the script [build.py](https://github.com/calixteman/pdf.js.quickjs/blob/main/build.py) into js one using node.
For now it's quite easy to run: `python3 build.py -c -o ../pdf.js.calixteman/src/scripting_api`.
And to update quickjs, just need to change https://github.com/calixteman/pdf.js.quickjs/blob/main/Dockerfile#L5 and then `python3 build.py -C -c -o ../pdf.js.calixteman/src/scripting_api` (-C is to update the docker image).
> 
>     3. Would it be at all possible to pull in the quickjs-files as a dependency, rather than simply dumping them into the PDF.js repository as-is?

I don't know enough to answer here.

The way it's built now is:
 - create a bundle for scripting stuff (initialization.js)
 - remove comments to decrease the size
 - stringify that stuff (it explains why I replace the \n,\t,\*)
 - create a bundle for the sandbox (quickjs-sandbox.js)
 - insert the string we created above in the code (since this code will be loaded (eval-ed) in the sandbox)
 - and at the end we've a file `pdf.sandbox.js` which contains everything (sandbox code+initialization code)

If you think about a better way to do, feel free to tell it or amend the patch or whatever.

gulpfile.js
```diff
@@ -493,6 +521,20 @@ function makeRef(done, bot) {
...
+  const defines = builder.merge(DEFINES, { GENERIC: true });
+  const scriptingDefines = builder.merge(defines, { NO_SOURCE_MAP: true });
+  return createScriptingBundle(scriptingDefines)
+    .pipe(stripComments())
```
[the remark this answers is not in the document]
calixteman: I thought the same but it seems that it's the case so I added that to do it.

gulpfile.js
calixteman: It's on purpose.
The idea is to get a bundle from `initialization.js` and the convert the bundle content into a string and inject this string in the sandbox bundle. This string will be eval-ed in the sandbox in order to create all what we need for the JS api.
But of course, if you see a better way to do, please tell me.

calixteman:
I moved my quickjs stuff under mozilla's umbrella:
https://github.com/mozilla/pdf.js.quickjs/

web/genericcom.js
```diff
@@ -49,6 +51,25 @@ class GenericExternalServices extends DefaultExternalServices {
   static createL10n({ locale = "en-US" }) {
     return new GenericL10n(locale);
   }
+
+  static get scripting() {
+    const promise = loadScript(AppOptions.get("scriptingSrc")).then(() => {
+      return window.pdfjsSandbox.QuickJSSandbox();
+    });
+    const sandbox = {
```
brendandahl, line 59: Is there a benefit to having all of these methods using `promise.then` vs have `scripting` return a promise and then have all the methods be direct calls? e.g.
```
get scripting() {
  return loadScript(...).then((sbx) => {
    return {
      createSandbox(data) {
         sbx.create(data);
      },
...
```

brendandahl (approved):
r+ depending on comment

web/genericcom.js
calixteman: It's for consistency with:
https://github.com/mozilla/pdf.js/blob/d3936ac9d2f7d11a8eeca6bdf86c3cd77cfbf02f/web/firefoxcom.js#L257-L269

Snuffleupagus:
At the very least, I would have expected e.g. a README file in https://github.com/mozilla/pdf.js/tree/master/external/quickjs mentioning things like:
 - What LICENSE applies to the code.
 - From where the code was obtained (original repository).
 - The *exact* build steps, and any necessary setup info, required to generate the code that's in the `quickjs-eval.js` file.
Without that, it can become very difficult for anyone to update that in e.g. six months or a years time (having to manually go through old PRs to find the information doesn't seem great).

---

Also, https://github.com/mozilla/pdf.js/blob/c88e8058707895e30abe9b91751c6ed5ffaa499f/web/app_options.js#L226-L233 is *really* unfortunate since it breaks the fact that the *development viewer* should be usable without manually building anything.
Furthermore, that block was placed among the API-options despite being a *viewer*-option!

PR #12617, opened 2020-11-13
Fix #9820 Thumbnail scale

### LINKED ISSUE #9820: Scaling of thumbnails does not work with larger files
Attach (recommended) or Link to PDF file here:  [pdf.js.test.7.pdf](https://github.com/mozilla/pdf.js/files/2107559/pdf.js.test.7.pdf)

Configuration:
- Web browser and its version: Google Chrome Version 67.0.3396.87 (Official Build) (64-bit)
- Operating system and its version: Windows 10 Version 1803 (OS Build 17134.112)
- PDF.js version: The 'demo' version at https://mozilla.github.io/pdf.js/web/viewer.html
- Is a browser extension: No

This does not seem to be platform / browser specific - I reproduced it on:
- Android Google Chrome
- Windows Internet Explorer

1. Use a device with a device pixel ratio > 1
2. Upload the attached file somewhere publicly accessible with appropriate CORS headers
3. View the attached document via demo URL https://mozilla.github.io/pdf.js/web/viewer.html?file=blah
4. Open the sidebar to view the thumbnails - some will be the correct size, other will not have been scaled properly

My desktop has a device pixel ratio of 1.25. 

The resultant output looks like this: thumbnail_scaling_error

The pages are all black. I have added an arrow to show the white area that indicates that the thumbnail image has not been scaled to fit the space.

My Android phone has a device pixel ratio of 2. The result looks even worse!

This isn't a CSS or styling issue - the generated dataURL appears to be wrong, as though the scaling has been ignored. Here's one of the dodgy thumbnails:
thumbnail_with_scaling_error

YipingRuan:
#9820

To reproduce the bug
- Set monitor DPI/Scale beyond 100%
- Thumbnails of pages out of view (not in the getVisibleElements from ui_utils.js)

image
image

### DIFF
--- a/web/pdf_thumbnail_view.js
+++ b/web/pdf_thumbnail_view.js
@@ -128,7 +128,8 @@ class PDFThumbnailView {
 
     this.canvasWidth = THUMBNAIL_WIDTH;
     this.canvasHeight = (this.canvasWidth / this.pageRatio) | 0;
-    this.scale = this.canvasWidth / this.pageWidth;
+    this.scale =
+      (this.canvasWidth / this.pageWidth) * (window.devicePixelRatio || 1);
 
     this.l10n = l10n;
 
@@ -177,7 +178,8 @@ class PDFThumbnailView {
     this.pageRatio = this.pageWidth / this.pageHeight;
 
     this.canvasHeight = (this.canvasWidth / this.pageRatio) | 0;
-    this.scale = this.canvasWidth / this.pageWidth;
+    this.scale =
+      (this.canvasWidth / this.pageWidth) * (window.devicePixelRatio || 1);
 
     this.div.removeAttribute("data-loaded");
     const ring = this.ring;


Snuffleupagus (changes requested):
Why is this necessary though, since the use of `getOutputScale` in `_getPageDrawContext` *should* already handle things?
Can you please compare the `PDFPageView` and `PDFThumbnailView` implementations to figure out *exactly* where they differ? As-is, it's not entirely clear (at least to me) where the bug actually is and why the current solution is correct.

Finally, please remember to write a good *commit message* as well (only including information in the PR description isn't enough).

YipingRuan:
Apparently it was not handled and the thumbnails have wrong scale at high DPI. 

My guess is in pdf_page_view.js it use:

```
const outputScale = getOutputScale(ctx);
```

Snuffleupagus:
> My guess is in pdf_page_view.js it use:
> 
> ```
> const outputScale = getOutputScale(ctx);
> ```
> 
> Not the case for "PDFThumbnailView".

Huh, it's most definitely used there as well; please see
https://github.com/mozilla/pdf.js/blob/59b35600be8c6ac2a00504a1813c52925e85c511/web/pdf_thumbnail_view.js#L245
The question is why it's not working as-is, and looking at that method in detail I've got an idea for the *actual* underlying bug.

YipingRuan:
Ok great! Leave to expert then.

Other observation is "this.scale" here was always 1

https://github.com/mozilla/pdf.js/blob/59b35600be8c6ac2a00504a1813c52925e85c511/web/pdf_thumbnail_view.js#L337

timvandermeij:
Let's close this PR, but keep the issue open so that the proper solution can be discussed. We do want to fix the underlying problem instead of potentially working around the actual bug. Thanks.

Issue #12290, opened 2020-08-27
Always in light mode even when using a dark mode firefox theme


bjorn3:
Attach (recommended) or Link to PDF file here: Any pdf

Configuration:
- Web browser and its version: Firefox Developer Edition 81.0b1 (64-bit)
- Operating system and its version: Debian 10
- PDF.js version: PDF.js: 2.6.276
- Is a browser extension: Firefox builtin pdf reader

Steps to reproduce the problem:
1. Set builtin dark theme for Firefox.
2. Open any pdf in Firefox.

What is the expected behavior? (add screenshot)

pdf.js uses dark colors

What went wrong? (add screenshot)

![image](https://user-images.githubusercontent.com/17426603/91433024-e71b8180-e862-11ea-8123-17be8bb59324.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): N/A

timvandermeij:
@utopianknight From your screenshots you did manage to get the viewer in the dark theme, and the code is also there. Do you perhaps have an idea why this might not work here?

brendandahl:
Strange, dark mode works fine for web version, but not built in version. Looks like there's some code to override this in Firefox. I'll open an upstream bug.

brendandahl:
One thing to clarify, using a dark mode theme in Firefox doesn't change everything to dark mode that uses the CSS `prefers-color-scheme: dark`. That has to be set at the operating system level. 

Though even if I set the OS to dark mode, the built in PDF viewer doesn't change themes.

utopianknight:
This seems to be a problem with Firefox, not PDF.js. Open these and play around with it in Firefox and in a chromium browser:
- https://codepen.io/jonneal/pen/vzPwWo
- https://color-scheme-demo.glitch.me/

Until this is fixed and as a temporary workaround, I can make the dark theme the default since it's what Firefox users have been accustomed to.

utopianknight:
Changing the theme with Menu > Customize > Themes > Light/Dark does not work.

But changing the theme in the dev tools (with devtools.inspector.color-scheme-simulation.enabled set to true in about:config) does work.

bjorn3:
> One thing to clarify, using a dark mode theme in Firefox doesn't change everything to dark mode that uses the CSS prefers-color-scheme: dark. That has to be set at the operating system level.

Unfortunately setting a dark mode theme for the xfce desktop is not detected by firefox.

[...]

PR #12625, opened 2020-11-15
Add a new preference, `viewerCssTheme`, to allow forcing the use of the light/dark viewer CSS themes (issue 12290)

Snuffleupagus:
While this does work pretty well in my quick testing, it's *very much* a hack since as far as I can tell there's no support in the CSS specification for using e.g. a CSS variable to override a `@media (prefers-color-scheme: dark) {...}` block.

The solution implemented here is thus to *edit* the viewer CSS, by either removing the entire `@media ...` block in light-mode or by ensuring that its rules become *unconditionally* applied in dark-mode.
To simplify the overall implementation, since all of this does seem like somewhat of an edge-case, the `viewerCssTheme` preference will *only* be read during viewer initialization. (Similar to many other existing preferences, a reload is thus required when changing it.)

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -198,6 +198,15 @@
         2
       ],
       "default": -1
+    },
+    "viewerCssTheme": {
+      "type": "integer",
+      "enum": [
+        0,
+        1,
+        2
+      ],
+      "default": 0
     }
   }
 }
--- a/web/app.js
+++ b/web/app.js
@@ -90,6 +90,12 @@ const ViewOnLoad = {
   INITIAL: 1,
 };
 
+const ViewerCssTheme = {
+  AUTOMATIC: 0, // Default value.
+  LIGHT: 1,
+  DARK: 2,
+};
+
 // Keep these in sync with mozilla-central's Histograms.json.
 const KNOWN_VERSIONS = [
   "1.0",
@@ -256,6 +262,7 @@ const PDFViewerApplication = {
 
     await this._readPreferences();
     await this._parseHashParameters();
+    this._forceCssTheme();
     await this._initializeL10n();
 
     if (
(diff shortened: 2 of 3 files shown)


timvandermeij:
I might be doing something wrong here, but when running `PDFViewerApplication.preferences.set("viewerCssTheme", 2);` in the console and reloading the page, I get the following stack trace and the viewer doesn't load:

```
Uncaught (in promise) Error: Permission denied to access property 0
    _forceCssTheme app.js:425
    initialize app.js:259
    run app.js:593
    webViewerLoad viewer.js:230
    __webpack_modules__ viewer.js:240
    Webpack 3
app.js:425:8
```

Snuffleupagus:
> I might be doing something wrong here, but when running `PDFViewerApplication.preferences.set("viewerCssTheme", 2);` in the console and reloading the page, I get the following stack trace and the viewer doesn't load:

That's really weird, and unfortunately I can't reproduce it with Firefox Nightly 84.0a1 (2020-11-15) (64-bit); however I've submitted a patch which hopefully *should* prevent that particular error.
Does that improve things at all, i.e. is the error gone and does the theme switch as expected now?

timvandermeij:
The reason for the error is at least more clear now: `_forceCssTheme: "Permission denied to access property 0".` Not sure why this happens, because it only happens when the console is open during the reload. If I close it, it's all good (using Firefox 82.0.3 on Arch Linux). This very much looks like the open browser bug like https://bugzilla.mozilla.org/show_bug.cgi?id=1423300, and given that this works fine without the console involved we can land this. Thanks!

timvandermeij (approved):

Snuffleupagus:
> This very much looks like the open browser bug https://bugzilla.mozilla.org/show_bug.cgi?id=1423300,

That one also references https://bugzilla.mozilla.org/show_bug.cgi?id=1673199, a regression in Firefox 82, which was fixed in Firefox 84 (and uplifted to Firefox 83); so there's a chance that it actually fixed :-)

Thanks for landing this; hopefully it'll help users work-around some of the problems described in #12290.

Snuffleupagus:
Given issues/bugs such as #12290 and [bug 1660650](https://bugzilla.mozilla.org/show_bug.cgi?id=1660650), which show that a fair number users clearly want to enforce the dark-theme for the PDF Viewer, I'm wondering if this is the sort of patch that we should consider uplifting to previous Firefox versions (once it's landed in Nightly of course)? /cc @brendandahl 

As far as I'm concerned, it should be fairly safe to uplift this since all code which could *potentially* throw is wrapped in `try...catch` to prevent the viewer from breaking.

qwer1304:
Why isn't this preference exported to Options?

Snuffleupagus:
> Why isn't this preference exported to Options?

Huh; please actually provide enough context with your question to make it easily understandable, since I've got no idea what "Options" you're referring to here!

Note that this is, in the `master`-branch (since it's not been part of an official release yet), already available through both the `AppOptions` and the `Preferences`, so I'm not sure what other "Options" there are here!?

qwer1304:
I was referring to the 'Extension options' (I have built the extension from master and can set the cssTheme from preferences with the desired effect, but don't see anything in Extension options'.)

Issue #12460, opened 2020-10-09
RTL Text Contains Strange Characters


henrywarhurst:
Attach (recommended) or Link to PDF file here: Fiddle: https://jsfiddle.net/47crfxqa/35/

Configuration:
- Web browser and its version: Chrome Version 85.0.4183.121
- Operating system and its version: MacOSX 10.14.6
- PDF.js version: v2.6.347
- Is a browser extension: no

Steps to reproduce the problem:
1. View the extracted text vs actual pdf content in https://jsfiddle.net/47crfxqa/35/
2. **Some of the rtl text has strange punctuation characters in it after parsing by PDFJS.**
3. All the rtl text seems to be backwards (i.e. the raw strings are ltr). I'm not sure if this is expected. The main issue here is point 2. 

The PDF I'm using in that fiddle is 'https://raw.githubusercontent.com/henrywarhurst/pdf-bug/add13fca1d6bc8a167ba95546f64633f2c962f84/rtl.pdf'

Snuffleupagus:
> Some of the rtl text has strange punctuation characters in it after parsing by PDFJS.

Without knowing the languages in question (looks like maybe Hebrew and Arabic), it's slightly difficult to tell what the problem is.

If it is the fact that some characters are copied as numbers, then similar behaviour can be observed in other PDF viewers as well (e.g. Adobe Reader and PDFium in Google Chrome). That is a bug in the PDF document *itself*, which is caused by incomplete/incorrect `/ToUnicode` data in the embedded font, rather than a problem in the PDF.js library.

henrywarhurst:
@Snuffleupagus thanks for the fast response. Ok it appears that if I try to copy the first piece of text from https://raw.githubusercontent.com/henrywarhurst/pdf-bug/add13fca1d6bc8a167ba95546f64633f2c962f84/rtl.pdf in Preview on OSX I also have the dollar sign, so happy to agree that isn't a PDFJS-specific issue.

I do still find it an issue that the RTL text is backwards when selected and copied though. In my app I have to reverse text in RTL languages manually when it comes back from PDFJS's `page.getTextContent()` which seems like a bug to me. For example in https://raw.githubusercontent.com/henrywarhurst/pdf-bug/add13fca1d6bc8a167ba95546f64633f2c962f84/rtl.pdf

[...]

PR #12627, opened 2020-11-16
Correctly render RTL text chunks

henrywarhurst:
Fixes an issue from https://github.com/mozilla/pdf.js/issues/12460 where RTL languages would appear in LTR order after a selection was made.

**Before (pasted text is backwards):**
pdfjs-before

**After:**
pdfjs-after


src/core/evaluator.js
```diff
@@ -2190,7 +2191,14 @@ class PartialEvaluator {
         }
         textState.translateTextMatrix(tx, ty);
 
-        textChunk.str.push(glyphUnicode);
+        // remark(henrywarhurst): Add glyph to the beginning of the string for
```
timvandermeij, line 2194: You can summarize this as:

```javascript
// Add the glyph to the beginning of the string for RTL languages,
// and to the end for LTR languages (fixes issue 12460).
```

This is a bit more compact and the common style for comments within the codebase. We don't use a username mention style.

src/core/evaluator.js
```diff
@@ -2190,7 +2191,14 @@ class PartialEvaluator {
...
+        // RTL languages, and to the end for LTR languages.
+        // See https://github.com/mozilla/pdf.js/issues/12460 for more info.
+        if (isRTLRangeFor(glyphUnicode.charCodeAt(0))) {
+          textChunk.str.unshift(glyphUnicode);
```
timvandermeij, line 2198: We should add a test for this to prevent future regressions, and I think we can test this with a reference test of the "text" type. If the PDF file from the JSFiddle in #12460 is yours, you can add it as a test case to this repository as it's a nice small reduced test case. To do this, the following is necessary:

1. Put the PDF file in the `test/pdfs` folder as `issue12460.pdf`.
2. Put `issue12460.pdf` in the `test/pdfs/.gitignore` file to allow it to be committed (see https://github.com/mozilla/pdf.js/blob/master/test/pdfs/.gitignore#L403 for example).
3. Include a new "text" test case in the test manifest in `test/test_manifest.json`, similar to an existing one like https://github.com/mozilla/pdf.js/blob/master/test/test_manifest.json#L56-L62.

timvandermeij (commented):
The approach looks good to me with these comments addressed.

src/core/unicode.js
```diff
@@ -1630,7 +1630,7 @@ var getNormalizedUnicodes = getArrayLookupTableFactory(function () {
 
 function reverseIfRtl(chars) {
   var charsLength = chars.length;
-  // Reverse an arabic ligature.
```
Snuffleupagus, line 1633: Please don't remove existing comments!

src/core/evaluator.js
Snuffleupagus: Also, shouldn't the existing [`reverseIfRtl` call above](https://github.com/mozilla/pdf.js/pull/12627/files#diff-cf40014d9c7d352f3ae212e7fffe5386decf96a10e53066a4d900c3eb8fc559eR2170) already handle this? Why is that not sufficient?

Furthermore, what's the performance impact of doing an extra `isRTLRangeFor` check for *every* single piece of text in a document?

src/core/unicode.js
henrywarhurst: Sorry! Could I please move the comment up to line 1630 and change it to say `// Reverse an RTL (right-to-left) ligature`? I removed it because the line it's above DOES NOT actually reverse a ligature, and the method as a whole deals with other RTL languages in addition to Arabic

src/core/evaluator.js
henrywarhurst: Hey @Snuffleupagus I actually think that line does nothing. Each glyph is an individual character, and in `reverseIfRtl()`, the function reverses strings, not individual characters

henrywarhurst: I'm happy to remove it if you agree? In my testing on various Hebrew and Arabic strings, removing https://github.com/mozilla/pdf.js/blob/f39d87bff1459b08f563db793d21f27ef843baab/src/core/evaluator.js#L2169 made no difference

timvandermeij: I didn't see it before, but I agree with @Snuffleupagus  that `reverseIfRtl` _should_ work. If it doesn't, then the bug most likely needs to be fixed in there since that function is already called here. Looking at https://github.com/mozilla/pdf.js/blob/2251677e5aad9f6db8635c91899fe5225df885d0/test/unit/unicode_spec.js#L114-L149, there are unit tests for it, which suggests that the functionality itself should be working, but it might be that it simply doesn't handle all cases correctly. It might help to find the root cause of this issue by adding a unit test for the case(s) from this particular document and seeing if it fails, and then try to change the function and see if all test cases then work.

henrywarhurst: @timvandermeij I think the difference between the pre-existing reverseIfRtl and what I have added is that what was there before reversed on a per character basis. This seems to be required in languages like arabic that have combination unicode characters like Alef. However, an RTL string still needs to be built in an RTL direction. The pre-existing code will always build the string in a LTR fashion, resulting in backwards RTL strings

Snuffleupagus: > I actually think that line does nothing. Each glyph is an individual character,

It most certainly does something! The assumption here unfortunately isn't correct, please note what the `getNormalizedUnicodes` function returns: https://github.com/mozilla/pdf.js/blob/a06f487bd40539cb44adaf9e6031d3beb3248c98/src/core/unicode.js#L246-L248

> I'm happy to remove it if you agree? In my testing on various Hebrew and Arabic strings, removing
> 
> https://github.com/mozilla/pdf.js/blob/f39d87bff1459b08f563db793d21f27ef843baab/src/core/evaluator.js#L2169

That'd break existing cases, and is (obviously) not correct. Most likely, your testing simply didn't come across those cases where it does matter.

---

> However, an RTL string still needs to be built in an RTL direction. The pre-existing code will always build the string in a LTR fashion, resulting in backwards RTL strings

Yes, and that's why https://github.com/mozilla/pdf.js/blob/a06f487bd40539cb44adaf9e6031d3beb3248c98/src/core/evaluator.js#L2124 takes care of flipping strings on an as-needed basis; see https://github.com/mozilla/pdf.js/blob/master/src/core/bidi.js

Having though more about this patch, I'm now pretty convinced that it will break existing cases and that it in its current form cannot be accepted unfortunately.
Basically, the changes you're proposing here are likely to "pull the rug out from under" the `bidi` function (and possibly more).

Was this patch *successfully* tested locally, see https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing, since I'd sort-of expect that it will cause regressions!?
(We probably don't have a great number of `text`-tests for RTL documents, but there should hopefully be *something* there.)

Snuffleupagus:
Closing for now, since there's not been any updates for quite some time and as-is this patch would cause regressions; issue #12460 remains open to track this.

PR #12630, opened 2020-11-16
Reduce, now unnecessary, asynchronicity in the `BasePreferences` constructor

Snuffleupagus:
Originally the default preferences were defined in a JSON-file checked into the repository, which was loaded using SystemJS in development mode.
Over the years a number of changes have been made to this code, most notably:
 - The preferences JSON-file is now generated automatically, during building, from the `AppOptions` abstraction.
 - All SystemJS usage has been removed from the development viewer.

Hence the default preferences are now available *synchronously* even in the development viewer, and it's thus no longer necessary to defer to the microtask queue (since `getDefaultPreferences` is async) just to get the default preferences.

While the effect of these changes is quite small, it *does* reduces the time it takes for the preferences to be fully initialized. Given the amount of asynchronous code during viewer initialization, every bit of time that we can save should thus help.

*Smaller/simpler diff with https://github.com/mozilla/pdf.js/pull/12630/files?w=1*

### DIFF
--- a/web/preferences.js
+++ b/web/preferences.js
@@ -15,22 +15,6 @@
 
 import { AppOptions, OptionKind } from "./app_options.js";
 
-let defaultPreferences = null;
-function getDefaultPreferences() {
-  if (!defaultPreferences) {
-    if (typeof PDFJSDev === "undefined" || !PDFJSDev.test("PRODUCTION")) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice simplification!

PR #12631, opened 2020-11-16
JS -- Implement app object

calixteman:
* https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/AcrobatDC_js_api_reference.pdf
 * Add color, fullscreen objects + few constants.

### DIFF
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -635,6 +635,59 @@ class PDFDateString {
   }
 }
 
+function makeColorComp(n) {
+  return Math.floor(Math.max(0, Math.min(1, n)) * 255)
+    .toString(16)
+    .padStart(2, "0");
+}
+
+const ColorConverters = {
+  // PDF specifications section 10.3
+  CMYK_G([c, y, m, k]) {
+    return ["G", 1 - Math.min(1, 0.3 * c + 0.59 * m + 0.11 * y + k)];
+  },
+  G_CMYK([g]) {
+    return ["CMYK", 0, 0, 0, 1 - g];
+  },
+  G_RGB([g]) {
+    return ["RGB", g, g, g];
+  },
+  G_HTML([g]) {
+    const G = makeColorComp(g);
+    return `#${G}${G}${G}`;
+  },
+  RGB_G([r, g, b]) {
+    return ["G", 0.3 * r + 0.59 * g + 0.11 * b];
+  },
+  RGB_HTML([r, g, b]) {
+    const R = makeColorComp(r);
+    const G = makeColorComp(g);
+    const B = makeColorComp(b);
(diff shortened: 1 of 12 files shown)


src/display/display_utils.js
```diff
@@ -635,6 +635,67 @@ class PDFDateString {
...
+function makeColorComp(n) {
+  return Math.floor(Math.max(0, Math.min(1, n)) * 255)
+    .toString(16)
+    .padStart(2, "0");
+}
+
+const ColorConverters = {
+  // PDF specifications section 10.3
+  CMYK_G(components) {
+    const [c, y, m, k] = components;
```
Snuffleupagus, line 647: Here, and everywhere below, is there a reason for not using the following (more compact) pattern instead?
```suggestion
  CMYK_G([c, y, m, k]) {
```

calixteman: Thx and fixed.

src/display/display_utils.js
```diff
@@ -635,6 +635,59 @@ class PDFDateString {
...
 }
 
+function makeColorComp(n) {
+  return Math.floor(Math.max(0, Math.min(1, n)) * 255)
+    .toString(16)
+    .padStart(2, "0");
+}
+
+const ColorConverters = {
+  // PDF specifications section 10.3
```
brendandahl, line 645: I don't think it matters that much for how this will be used, but we do have a bunch of colorspace conversion code that does a better job than the simplistic conversions given by the spec: https://github.com/mozilla/pdf.js/blob/master/src/core/colorspace.js

src/scripting_api/app.js
```diff
@@ -56,6 +411,160 @@ class App extends PDFObject {
...
+      );
+    }
+
+    const id = this._setInterval(cExpr, nMilliseconds);
+    const timeout = Object.create(null);
+    this._registerTimeout(timeout, id, true);
+    return timeout;
+  }
+
+  setTimeOut(cExpr, nMilliseconds) {
```
brendandahl, line 547: I wonder how they ended up with a different case than JS. :(

src/scripting_api/color.js
```diff
@@ -0,0 +1,129 @@
...
+      case "CMYK":
+        if (colorArray.length !== 5) {
+          return false;
+        }
+        break;
+      default:
+        return false;
+    }
+
+    return colorArray
```
brendandahl, line 81: Is this called very often? Allocating a new array and doing every seems excessive vs for loop.

src/display/display_utils.js
calixteman: I saw that but afaik the end goal is to be able to set a color in a textfield or whatever I think too that it doesn't matter.
FYI, pdfium code does the same here.

PR #12635, opened 2020-11-18
JS -- Send events to the sandbox from annotation layer

calixteman:


brendandahl:
Will this be tested in another patch or are there some manual tests for now?

brendandahl (approved):
r+ with update for the window event

PR #12637, opened 2020-11-18
JS -- Add support for buttons

calixteman:
* radio buttons
 * checkboxes


brendandahl:
Needs a rebase. Did a quick test with it rebased and radio buttons are working as expected, but radio checkboxes are not. http://www.windjack.com/DownLoads/RadioCheckBox_AcroForm.pdf

Would be good if we could get the puppeteer testing added, so then we can add tests for this first.

calixteman:
The radio checkboxes have no JS here so there is a bug in annotation.js itself (anyway there is a bug in my code to manage such a situation).

brendandahl:
Can we add a test for this code now too?

test/integration/scripting_spec.js
```diff
@@ -191,4 +191,79 @@ describe("Interaction", () => {
...
+          const expected = [
+            ["#\\36 8R", "Group1=Choice1::1"],
+            ["#\\36 9R", "Group1=Choice2::2"],
+            ["#\\37 0R", "Group1=Choice3::3"],
+            ["#\\37 1R", "Group1=Choice4::4"],
+          ];
+          await page.waitForTimeout(100);
+          for (const [selector, expectedText] of expected) {
+            await page.click(selector);
+            await page.waitForTimeout(100);
```
brendandahl, line 218: As we've seen in other tests, using timeouts is a big cause of intermittent failures. Can we add an annotationchanged event or something instead?

test/integration/scripting_spec.js
```diff
@@ -191,4 +191,79 @@ describe("Interaction", () => {
...
+    it("must show values in a text input when clicking on radio buttons", async () => {
+      await Promise.all(
+        pages.map(async ([browserName, page]) => {
+          const expected = [
+            ["#\\36 8R", "Group1=Choice1::1"],
+            ["#\\36 9R", "Group1=Choice2::2"],
+            ["#\\37 0R", "Group1=Choice3::3"],
+            ["#\\37 1R", "Group1=Choice4::4"],
+          ];
+          await page.waitForTimeout(100);
```
brendandahl, line 215: Same here, why is this timeout needed is there an event that can be used instead?

brendandahl (changes requested):

PR #12642, opened 2020-11-20
[api-minor] Add "contentLength" to the information returned by the `getMetadata` method

Snuffleupagus:
Given that we already include the "Content-Disposition"-header filename, when it exists, it shouldn't hurt to also include the information from the "Content-Length"-header.
For PDF documents opened via a URL, which should be a very common way for the PDF.js library to be used, this will[1] thus provide a way of getting the PDF filesize without having to wait for the `getDownloadInfo`-promise to resolve[2].

With these API improvements, we can also simplify the filesize handling in the `PDFDocumentProperties` class.

*Smaller/simpler diff with https://github.com/mozilla/pdf.js/pull/12642/files?w=1*

---
[1] Assuming that the server is correctly configured, of course.

[2] Since that's not *guaranteed* to happen in general, with e.g. `disableAutoFetch = true` set.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2654,9 +2654,8 @@ class WorkerTransport {
         return {
           info: results[0],
           metadata: results[1] ? new Metadata(results[1]) : null,
-          contentDispositionFilename: this._fullReader
-            ? this._fullReader.filename
-            : null,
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Nice work!

PR #12645, opened 2020-11-21
Convert the `PDFDocumentProperties.open` method to be async

Snuffleupagus:
By using `await`, rather than chaining promises, this method becomes more compact and slightly easier to reason about (at least in my opinion).

*Smaller/simpler diff with https://github.com/mozilla/pdf.js/pull/12645/files?w=1*


timvandermeij (approved):

timvandermeij:
Thank you!

PR #12647, opened 2020-11-22
Update Puppeteer to version 5.5.0

timvandermeij:
This might help to reduce the number of intermittent failures in Chrome, but we'll see...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2912,9 +2912,9 @@
       }
     },
     "buffer": {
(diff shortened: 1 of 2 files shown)


Snuffleupagus (approved):
r=me, let's hope this helps!

timvandermeij:
Unfortuntely it didn't do as much as I had hoped, but at the very least it can't hurt to use the latest Chromium, so let's merge this anyway.

PR #12655, opened 2020-11-24
Parenthesis in names are not escaped when saving

### LINKED ISSUE #12654: Wrong display and invalid values in AP dictionary of radio button group when filled/saved
[Subscription-tekMag.pdf](https://github.com/mozilla/pdf.js/files/5591356/Subscription-tekMag.pdf)
[Subscription-tekMag-filled-by-pdfjs.pdf](https://github.com/mozilla/pdf.js/files/5591357/Subscription-tekMag-filled-by-pdfjs.pdf)

Configuration:
- Web browser and its version: FF 83.0
- Operating system and its version: Windows 10 Pro 
- PDF.js version: 2.7.99 / 2.7.295
- Is a browser extension: yes / [no](https://mozilla.github.io/pdf.js/web/viewer.html)

1. Open document [Subscription-tekMag.pdf](https://github.com/mozilla/pdf.js/files/5591356/Subscription-tekMag.pdf)
2. See the wrong rendered appearances
3. Choose a subscription through the radio button group (Choose your Subscription)
4. Save the document (STRG+S)

grafik

What went wrong? 
grafik

Also when the file is saved the AP dictionary has invalid data and the radio button groups are not useable anymore. E.g.:

    /N << /2#20years,#2024#20issues#20(digital#20download#20only) 162 0 R>>

Looks like the name is not escaped correctly.

(The checkbox at the end has also a strange behavior (always locked when saved) - ping me, if I should open an additional issue for this one)

calixteman:
it aims to fix #12654.

### DIFF
--- a/src/core/core_utils.js
+++ b/src/core/core_utils.js
@@ -205,7 +205,22 @@ function escapePDFName(str) {
   let start = 0;
   for (let i = 0, ii = str.length; i < ii; i++) {
     const char = str.charCodeAt(i);
-    if (char < 0x21 || char > 0x7e || char === 0x23) {
+    // Whitespace or delimiters aren't regular chars, so escape them.
+    if (
+      char < 0x21 ||
+      char > 0x7e ||
+      char === 0x23 /* # */ ||
+      char === 0x28 /* ( */ ||
+      char === 0x29 /* ) */ ||
+      char === 0x3c /* < */ ||
+      char === 0x3e /* > */ ||
+      char === 0x5b /* [ */ ||
+      char === 0x5d /* ] */ ||
+      char === 0x7b /* { */ ||
+      char === 0x7d /* } */ ||
+      char === 0x2f /* / */ ||
+      char === 0x25 /* % */
+    ) {
       if (start < i) {
         buffer.push(str.substring(start, i));
       }
--- a/test/unit/core_utils_spec.js
+++ b/test/unit/core_utils_spec.js
@@ -237,6 +237,9 @@ describe("core_utils", function () {
         "#fehe#fell#ffo#ff"
       );
       expect(escapePDFName("#h#e#l#l#o")).toEqual("#23h#23e#23l#23l#23o");
+      expect(escapePDFName("#()<>[]{}/%")).toEqual(
+        "#23#28#29#3c#3e#5b#5d#7b#7d#2f#25"
+      );
     });
   });
 });


src/core/core_utils.js
```diff
@@ -205,7 +205,13 @@ function escapePDFName(str) {
...
   for (let i = 0, ii = str.length; i < ii; i++) {
     const char = str.charCodeAt(i);
-    if (char < 0x21 || char > 0x7e || char === 0x23) {
+    if (
+      char < 0x21 ||
+      char > 0x7e ||
+      char === 0x23 /* # */ ||
+      char === 0x28 /* ( */ ||
+      char === 0x29 /* ) */
+    ) {
```
JanSlabon, line 214: A Name object should escape ALL possible delemitter and not only these 3. So I would say that these are still missing: `<>[]{}\%`

calixteman: That's correct, I'm fixing that, thx.

timvandermeij: For our reference, is this mentioned somewhere in the PDF specification so we can verify this patch? The change itself looks good if for all characters a unit test is available, but it would be good to check with the specification too.

JanSlabon: In PDF 32000-1:2008 it is 7.5.3:

> a) A NUMBER SIGN (23h) (#) in a name shall be written by using its 2-digit hexadecimal code (23), preceded by the NUMBER SIGN.
b ) Any character in a name that is a regular character (other than NUMBER SIGN) shall be written as itself or by using its 2-digit hexadecimal code, preceded by the NUMBER SIGN.
c) Any character that is not a regular character shall be written using its 2-digit hexadecimal code, preceded by the NUMBER SIGN only.

> White space used as part of a name shall always be coded using the 2-digit hexadecimal notation and no white space may intervene between the SOLIDUS and the encoded name.
> Regular characters that are outside the range EXCLAMATION MARK(21h) (!) to TILDE (7Eh) (~) should be written using the hexadecimal notation.

The quoted characters are all delemiter characters (see 7.2.2 Character Set). If you would not escape them a new object will begin ;-)

src/core/core_utils.js
```diff
@@ -205,7 +205,23 @@ function escapePDFName(str) {
   let start = 0;
   for (let i = 0, ii = str.length; i < ii; i++) {
     const char = str.charCodeAt(i);
-    if (char < 0x21 || char > 0x7e || char === 0x23) {
+    // whites or delimiter aren't regular chars
+    // so escape them
```
Snuffleupagus, line 209: "whites", did you actually mean "whitespace" here?

Also, please improve the formatting of this comment, since the line-break is just overall confusing (i.e. is this intended to be two sentences, or not):
```suggestion
    // Whitespace or delimiters aren't regular chars, so escape them.
```

src/core/core_utils.js
timvandermeij: For future reference: it's in section 7.3.5, not 7.5.3, but it does allow us to check this patch against. Thanks!

JanSlabon: ups, sorry... sure 7.3.5! It was late :-)

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #12666, opened 2020-11-29
Use the same SVG images, in the default viewer, regardless of the CSS theme

Snuffleupagus:
As mentioned in https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support, PDF.js version `2.6.347` is the last release with IE 11/Edge support.
Hence we should now be able to reduce unnecessary duplication in the default viewer image resources, note the files in the `web/images/` folder with a `-dark` suffix, by using only *one* SVG-image for each icon and letting the `background-color` depend on the CSS theme instead.

For the `gulp mozcentral` build-target, the resulting `web/images/` folder is reduced from `43 997` to `28 566` bytes (~35 percent).

*Please note:* I don't really know if this implementation is necessarily the *best* solution, but it seems to work well enough in e.g. Firefox Nightly and Google Chrome Beta as far as my testing goes.


web/viewer.css
```diff
@@ -128,40 +130,6 @@
     --overlay-button-hover-color: rgba(115, 115, 115, 1);
 
     --loading-icon: url(images/loading-dark.svg);
```
Snuffleupagus, line 132: The approach used for the buttons unfortunately doesn't seem to work here, since this one is applied to `<input>` elements which as far as I can tell doesn't support using `::before`/`::after`.

Most likely there's a clever work-around, but so far I've been unable to come up with anything here.
Given that this patch improves things overall, with respect to the size of the image resources in the default viewer, do we perhaps want to consider this patch even as-is?

timvandermeij: I'm fine with that, but in that case it would be good to add a comment to this one left-over dark image stating why it's still there.

web/viewer.css
```diff
@@ -903,12 +871,11 @@ html[dir="rtl"] #toolbarViewerLeft > .toolbarButton:first-child {
...
 }
 .dropdownToolbarButton::after {
-  position: absolute;
-  display: inline-block;
   top: 6px;
-  content: var(--toolbarButton-menuArrow-icon);
   pointer-events: none;
-  max-width: 16px;
+
+  -webkit-mask-image: var(--toolbarButton-menuArrow-icon);
```
timvandermeij, line 877: Doesn't Autoprefixer handle this for us so we don't have to worry about vendor prefixes anymore (same below)?

Snuffleupagus: Annoyingly enough, Chromium doesn't support the *un*prefixed version of this CSS property and I thus opted to include this to make sure that the *development* viewer doesn't look completely broken locally in Chromium-browsers.

I should obviously have mentioned, sorry about that!

web/viewer.css
Snuffleupagus: Really good point, I've added a comment about this; thank you!

timvandermeij (approved):

timvandermeij:
Looks good to me. Thanks!

PR #12668, opened 2020-11-30
Add some integration tests using puppeteer

calixteman:
* run with `gulp integrationtest`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -98,6 +98,7 @@ const DEFINES = Object.freeze({
   PRODUCTION: true,
   SKIP_BABEL: true,
   TESTING: false,
+  ENABLE_SCRIPTING: false,
   // The main build targets:
   GENERIC: false,
   MOZCENTRAL: false,
@@ -303,14 +304,26 @@ function checkChromePreferencesFile(chromePrefsPath, webPrefsPath) {
     return false;
   }
   if (webPrefsKeys.length !== chromePrefsKeys.length) {
+    console.log("Warning: Prefs objects haven't the same length");
     return false;
   }
-  return webPrefsKeys.every(function (value, index) {
-    return (
-      chromePrefsKeys[index] === value &&
-      chromePrefs.properties[value].default === webPrefs[value]
-    );
-  });
+
+  let ret = true;
+  for (let i = 0, ii = webPrefsKeys.length; i < ii; i++) {
+    const value = webPrefsKeys[i];
+    if (chromePrefsKeys[i] !== value) {
+      ret = false;
+      console.log(
+        `Warning: not the same keys: ${chromePrefsKeys[i]} !== ${value}`
+      );
+    } else if (chromePrefs.properties[value].default !== webPrefs[value]) {
+      ret = false;
+      console.log(
+        `Warning: not the same values: ${chromePrefs.properties[value].default} !== ${webPrefs[value]}`
+      );
+    }
+  }
+  return ret;
 }
 
 function replaceWebpackRequire() {
@@ -515,6 +528,9 @@ function createTestSource(testsName, bot) {
       case "font":
         args.push("--fontTest");
         break;
+      case "integration":
+        args.push("--integration");
+        break;
       default:
         this.emit("error", new Error("Unknown name: " + testsName));
         return null;
@@ -646,6 +662,7 @@ gulp.task("default_preferences-pre", function () {
       LIB: true,
       BUNDLE_VERSION: 0, // Dummy version
       BUNDLE_BUILD: 0, // Dummy build
+      ENABLE_SCRIPTING: process.env.ENABLE_SCRIPTING === "true",
     }),
     map: {
       "pdfjs-lib": "../pdf",
@@ -1507,27 +1524,46 @@ gulp.task("testing-pre", function (done) {
   done();
 });
 
+gulp.task("enable-scripting", function (done) {
+  process.env.ENABLE_SCRIPTING = "true";
+  done();
+});
+
 gulp.task(
   "test",
-  gulp.series("testing-pre", "generic", "components", function () {
-    return streamqueue(
-      { objectMode: true },
-      createTestSource("unit"),
-      createTestSource("browser")
-    );
-  })
+  gulp.series(
+    "enable-scripting",
+    "testing-pre",
+    "generic",
+    "components",
+    function () {
+      return streamqueue(
+        { objectMode: true },
+        createTestSource("unit"),
+        createTestSource("browser"),
+        createTestSource("integration")
+      );
+    }
+  )
 );
 
 gulp.task(
   "bottest",
-  gulp.series("testing-pre", "generic", "components", function () {
-    return streamqueue(
-      { objectMode: true },
-      createTestSource("unit", true),
-      createTestSource("font", true),
-      createTestSource("browser (no reftest)", true)
-    );
-  })
+  gulp.series(
+    "enable-scripting",
+    "testing-pre",
+    "generic",
+    "components",
+    function () {
+      return streamqueue(
+        { objectMode: true },
+        createTestSource("unit", true),
+        createTestSource("font", true),
+        createTestSource("browser (no reftest)", true),
+        createTestSource("integration")
+      );
+    }
+  )
 );
 
 gulp.task(
@@ -1542,6 +1578,13 @@ gulp.task(
   gulp.series("testing-pre", "generic", "components", function () {
     process.env.TZ = "UTC";
     return createTestSource("unit");
+  })
+);
+
+gulp.task(
+  "integrationtest",
+  gulp.series("enable-scripting", "testing-pre", "generic", function () {
+    return createTestSource("integration");
   })
 );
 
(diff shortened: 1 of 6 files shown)


test/test.js
```diff
@@ -768,7 +803,20 @@ function unitTestPostHandler(req, res) {
...
+async function runIntegrationTest(browserName, baseUrl) {
+  const browser = await startBrowser(browserName);
+  const { runMe } = require("./unit/integration.js");
+  try {
+    await runMe(baseUrl, browser);
+  } catch (err) {
+    console.log(err);
+  } finally {
+    browser.close();
+    process.exit(0);
```
timvandermeij, line 815: I think the exit code should be nonzero in the case of an error.

test/test.js
```diff
@@ -790,14 +838,29 @@ async function startBrowser(browserName, startUrl) {
...
+  const options = {
     product: browserName,
     headless: false,
     defaultViewport: null,
-  });
-  const pages = await browser.pages();
-  const page = pages[0];
-  await page.goto(startUrl, { timeout: 0 });
+  };
+  if (browserName === "chrome") {
```
timvandermeij, line 846: Could you add a comment why these additional Chrome and Firefox preferences are necessary?

test/unit/integration.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+const assert = require("assert");
+
+const itCallbacks = [];
+function it(desc, callback) {
+  itCallbacks.push([desc, callback]);
+}
+
+const testsCallback = [];
```
timvandermeij, line 23: Maybe call this `describeCallbacks` similar to the one above?

test/unit/integration.js
```diff
@@ -0,0 +1,74 @@
...
+function describe(desc, filename, id, callback) {
+  testsCallback.push([desc, filename, id, callback]);
+}
+
+function selector(id) {
+  return `#\\3${id.charAt(0)} ${id.slice(1)}`;
+}
+
+async function runMe(baseUrl, browser) {
```
timvandermeij, line 32: Maybe call this entrypoint `runTests` instead so it's a bit more descriptive?

test/unit/integration.js
```diff
@@ -0,0 +1,74 @@
...
+        await page.waitForSelector(selector(id), {
+          timeout: 0,
+        });
+        itCallbacks.push(["", async () => page.bringToFront()]);
+        cb(page);
+      })()
+    )
+  );
+
+  // Execute it sequencially
```
timvandermeij, line 48: `sequencially` -> `sequentially`

test/unit/integration.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function selector(id) {
+  return `#\\3${id.charAt(0)} ${id.slice(1)}`;
+}
+
+async function runMe(baseUrl, browser) {
+  // Create the pages
+  await Promise.all(
+    testsCallback.map(([_, filename, id, cb]) =>
```
timvandermeij, line 35: We don't seem to be using the `id` here at all. Removing that would also make this helper more generally useful for other integration tests.

timvandermeij (commented):
Nice idea! I primarily have some comments that should make this more generally usable. We should think about if we want to make this part of the bot/Travis tests already or to do that later.

test/test.js
calixteman: About option for Chrome, they're required at least on my machine because without them chrome is crashing!
And as a side effect I just discovered that unit tests are running on chrome too.

test/unit/integration.js
calixteman: the id is used as a selector to select the element to wait for before running the test.

test/integration/integration-boot.js
```diff
@@ -0,0 +1,88 @@
...
+
+function loadTestFiles() {
+  ["./scripting.js", "./annotation.js"].map(function (moduleName) {
+    return require(moduleName);
+  });
+}
+
+const describes = [];
+
+global.describe = (desc, filename, selector, callback) => {
```
brendandahl, line 26: Instead of adding a new test framework, can we load jasmine (which we already use)? I was hoping to use the integration tests for more than just form tests which this is pretty tailored too. This current method would then be a helper that a test would use.

brendandahl (changes requested):
For getting various events when a page is rendered and forms are ready we can enable the dom dispatch. It's currently only enabled for development and mozcentral builds, so we'd need to add a new build flag or option to https://github.com/mozilla/pdf.js/blob/a618b02e6274c173c414855d8311930c43b99c00/web/ui_utils.js#L806

Snuffleupagus:
> For getting various events when a page is rendered and forms are ready we can enable the dom dispatch.

While we *used* to dispatch events to the DOM, that was *purposely* removed since it's not a great API in the general case (and only existed for backwards compatibility); hence can we please *not* re-add that sort of functionality here!

If it's turns out to be *absolutely* necessary, with no way around it, please at least limit this to *only* being enabled (using pre-processor statements) for `TESTING`-builds.

> It's currently only enabled for development and mozcentral builds, so we'd need to add a new build flag or option to

That's not entirely accurate, since it's *only* enabled for mozilla-central (and nowhere else) to not break the existing tests in https://searchfox.org/mozilla-central/source/toolkit/components/pdfjs/test

timvandermeij:
Related: https://github.com/mozilla/botio-files-pdfjs/pull/28

brendandahl:
For the windows failure we need to increase the timeout for the beforeAll (second argument). There's suppose to be a way to change it for everything with jasmine.DEFAULT_TIMEOUT_INTERVAL but I can't seem to get that to apply.

gulpfile.js
```diff
@@ -1502,6 +1519,31 @@ gulp.task(
...
+
+      rimraf.sync(sandboxDir);
+
+      return createSandboxBundle(defines, {
+        disableVersionInfo: true,
+      }).pipe(gulp.dest(sandboxDir));
+    }
+  )
+);
+
```
Snuffleupagus, line 1546: Is there a reason for moving all of this code?

gulpfile.js
```diff
@@ -1545,6 +1589,24 @@ gulp.task(
...
+  done();
+});
+
+gulp.task(
+  "integrationtest",
+  gulp.series(
+    "integration-pre",
+    "testing-pre",
+    "generic",
+    "components",
```
Snuffleupagus, line 1603: Are you actually using the output of the "components" build for anything here?

test/integration/annotation_spec.js
```diff
@@ -0,0 +1,63 @@
...
+            `${global.integrationBaseUrl}?file=/test/pdfs/annotation-highlight.pdf`
+          );
+          await page.bringToFront();
+          await page.waitForSelector("[data-annotation-id='19R']", {
+            timeout: 0,
+          });
+          return page;
+        })
+      );
+    }, 1e5);
```
Snuffleupagus, line 34: Having a *seemingly* arbitrary constant here, and in other files, is neither very clear nor maintainable in the future. Please model the new `test/integration-boot.js` file after the "boot"-files used with the existing unit- and font-tests instead, note in particular https://github.com/mozilla/pdf.js/blob/93b3ba2abb4e2eb760589899ff38cb99e8f0cbae/test/unit/jasmine-boot.js#L195-L196 and
https://github.com/mozilla/pdf.js/blob/93b3ba2abb4e2eb760589899ff38cb99e8f0cbae/test/font/jasmine-boot.js#L136-L137

Snuffleupagus (changes requested):

web/app_options.js
```diff
@@ -260,7 +260,8 @@ if (
   defaultOptions.sandboxBundleSrc = {
     /** @type {string} */
     value:
-      typeof PDFJSDev === "undefined" || !PDFJSDev.test("PRODUCTION")
+      typeof PDFJSDev === "undefined" ||
+      PDFJSDev.test("!PRODUCTION && !INTEGRATIONTEST")
```
Snuffleupagus, line 264: Also, are these changes actually necessary since there *should* be no build-target setting `PRODUCTION = false` anyway?

gulpfile.js
calixteman: Correct, it's useless now

gulpfile.js
calixteman: No

web/app_options.js
calixteman: Good question, there was a `!PDFJSDev.test("PRODUCTION")`, is this value different of  `PDFJSDev.test("!PRODUCTION")` ?
And there's a similar use here:
https://github.com/mozilla/pdf.js/blob/master/web/app_options.js#L248

test/integration/annotation_spec.js
calixteman: I tried to play with `jasmine.DEFAULT_TIMEOUT_INTERVAL` and it never worked.
Finally I found how to set it correctly: `jasmine.jasmine.DEFAULT_TIMEOUT_INTERVAL`.

web/app_options.js
Snuffleupagus: > there was a `!PDFJSDev.test("PRODUCTION")`, is this value different of `PDFJSDev.test("!PRODUCTION")` ?

Those two formats are equivalent.

brendandahl (approved):
Let's merge this if the full test suite runs fine, so we can unbreak tests for other PR's.

timvandermeij:
Thank you for setting this up! This will be very useful so we can actually start writing integration tests for the viewer now.

PR #12673, opened 2020-12-01
Split underline, strikeout, squiggly annotions div into multiple divs

calixteman:
* Follow up of #12505
 * Fix bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1679696

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -1462,6 +1462,7 @@ class UnderlineAnnotationElement extends AnnotationElement {
       parameters.data.contents
     );
     super(parameters, isRenderable, /* ignoreBorder = */ true);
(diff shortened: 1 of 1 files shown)


calixteman:
The user (in bug 1679696) shared privately the PDF and this patch fix the issue.
The div for strikeout is over the div for some highlights, so splitting them fix the issue.

timvandermeij (approved):

timvandermeij:
Nice work! I actually had this exact same patch stashed locally, but got distracted with another edge case that we need to fix. I'll make a patch for that later which will also include a minimal reference test case for all these annotations, but I can confirm that this patch indeed also fixes the normal cases in that file.

/botio makeref

Issue #6854, opened 2016-01-11
Make linked test cases point to the Internet Archive


timvandermeij:
We have quite a number of linked test cases in the https://github.com/mozilla/pdf.js/tree/master/test/pdfs folder. Some of them point to sources other than the Internet Archive or Bugzilla and might become unavailable, which is to be avoided.

It is better to fetch the files from the Internet Archive to be sure that they remain available. This should always be possible if the current source exists because once you input the link on the Internet Archive website it will archive the file automatically.

Below is a list of linked test cases that point to external sources and should be updated (obtained with `grep -L "bugzilla.mozilla.org\|archive.org" *.pdf.link`). Make sure that the hash of the Internet Archive file is the same as the hash in the test manifest.

- [x] 20130226130259.pdf.link
- [x] JST2007-5.pdf.link
- [x] P020121130574743273239.pdf.link
- [x] SFAA_Japanese.pdf.link
- [x] TaroUTR50SortedList112.pdf.link
- [x] aboutstacks.pdf.link
- [x] artofwar.pdf.link
- [x] bpl13210.pdf.link
- [x] bug766138.pdf.link
- [x] bug808084.pdf.link
- [x] bug887152.pdf.link
- [x] cable.pdf.link
- [x] ecma262.pdf.link
- [x] fips197.pdf.link
- [x] fit11-talk.pdf.link
- [x] geothermal.pdf.link
- [x] hmm.pdf.link
- [x] html5checker.pdf.link
- [x] hudsonsurvey.pdf.link
- [x] ichiji.pdf.link
- [x] issue1010.pdf.link
- [x] issue1015.pdf.link
- [x] issue1049.pdf.link
- [x] issue1096.pdf.link
- [x] issue1127.pdf.link
- [x] issue1133.pdf.link
- [x] issue1155.pdf.link
- [x] issue1169.pdf.link
- [x] issue1233.pdf.link
- [x] issue1257.pdf.link
- [x] issue1309.pdf.link
- [x] issue1317.pdf.link
- [x] issue1419.pdf.link
- [x] issue1466.pdf.link
- [x] issue1597.pdf.link
- [x] issue1629.pdf.link
- [x] issue1658.pdf.link
- [x] issue1685.pdf.link
- [x] issue1687.pdf.link
- [x] issue1709.pdf.link
- [x] issue1721.pdf.link
- [x] issue1729.pdf.link
- [x] issue1796.pdf.link
- [x] issue1810.pdf.link
[...]

PR #12681, opened 2020-12-02
Update the link for the "pr8808" test-case (issue 12680)

### LINKED ISSUE #12680: pr8808.pdf.link is not accessible in my country.
The link reports the following message.
```
Statens It har blokeret denne hjemmeside
Du har forsoegt at tilgaa en hjemmeside, der udgoer en sikkerhedsmaessig risiko, derfor har Statens It blokeret siden.
Har du et arbejdsbetinget behov for at tilgaa denne side, saa bedes du kontakte din it-sikkerhedskoordinator
```

Snuffleupagus:
This seems like a very minor issue, since in general we can't really help if domains are blocked from certain networks, however in this particular case I suppose that using the Internet Archive should work.

Fixes #12680

### DIFF
--- a/test/pdfs/pr8808.pdf.link
+++ b/test/pdfs/pr8808.pdf.link
@@ -1,1 +1,1 @@
-https://dokument.plandata.dk/20_1057347_APPROVED_1193694109382.pdf
+https://web.archive.org/web/20201202134646/https://dokument.plandata.dk/20_1057347_APPROVED_1193694109382.pdf


timvandermeij (approved):

timvandermeij:
Thanks! I verified that the MD5 hashes of both files are equal. Interesting because I thought we replaced all direct links with Internet Archive links already, but I guess this one slipped though.

Snuffleupagus:
> I verified that the MD5 hashes of both files are equal.

Thanks for double checking this!

> Interesting because I thought we replaced all direct links with Internet Archive links already,

Indeed that was done in issue #6854 (back in 2016), but PR #8808 is from 2017 so that should explain it :-)

PR #12682, opened 2020-12-02
[Generic viewer] Re-factor how the `notFound` appearance is set on the "findInput" in the `PDFFindBar`

Snuffleupagus:
Rather than having two slightly different ways of setting the pending/notFound appearance on the "findInput", we can simply use "data-status" in both cases since they're obviously mutually exclusive.


timvandermeij (approved):

timvandermeij:
Looks good!

Issue #12684, opened 2020-12-02
Possible typo/error in the `src/scripting_api/field.js` file


Snuffleupagus:
The following code looks really strange, since I cannot imagine that using an Array together with a bitwise OR operator (`|`) makes much sense in general!?
Shouldn't these lines simply use a "normal" OR operator (`||`) instead?
https://github.com/mozilla/pdf.js/blob/a618b02e6274c173c414855d8311930c43b99c00/src/scripting_api/field.js#L75-L77

/cc @calixteman

calixteman:
my bad... you're right it should be a `||`.

PR #12685, opened 2020-12-02
Fix issue #12684: replace bitwise ORs by ORs

calixteman:


Snuffleupagus (approved):
Thanks for the quick follow-up!

PR #8338, opened 2017-04-27
Store image dictionary interpolate entry. Apply it as smoothing in pa…

davehouse:
I would like to have the image dictionary interpolate boolean entry applied to image painting so that hard pixel edges are possible. This uses canvas context imageSmoothingEnabled (https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled).

Fixes issue #5757

Test PDF:
[issue5757.pdf](https://github.com/mozilla/pdf.js/files/960522/issue5757.pdf)

Before:
image

After:
image

Acrobat Reader:
image

PR #12688, opened 2020-12-03
Disable interpolation of XObject when Interpolate entry is not true (issue9713)

ghost:
According pdf reference section 4.8, image should not be interpolated unless Interpolate entry is True.

> When the resolution of a source image is significantly lower than that of the out-put device, each source sample covers many device pixels. This can cause imagestoappear  “jaggy”  or  “blocky.”  These  visual  artifacts  can  be  reduced  by  applyingan image interpolationalgorithm during rendering. Instead of painting all pixelscovered by a source sample with the same color, image interpolation attempts toproduce a smooth transition between adjacent sample values. Because it may in-crease the time required to render the image, image interpolation is disabled bydefault; setting the Interpolate entry in the image dictionary to true enables it.


Snuffleupagus:
This looks, for all intents and purposes, like a duplicate of PR #8338; please note that that one was closed because it caused *way too many* outright regressions in existing test-cases.
Did this patch actually pass *all* existing tests locally, please refer to https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing additional information?

Kyroi:
Yes, it is duplicated, and it didn't pass the test-cases.

PR #12689, opened 2020-12-03
In order to simplify m-c code, move some in pdf.js

calixteman:
* move set/clear|Timeout/Interval and crackURL code in pdf.js
* remove the "backdoor" in the proxy (used to dispatch event) and so return the dispatch function in the initializer
* remove listeners if an error occured during sandbox initialization
* add support for alert and prompt in the sandbox
* add a function to eval in the global scope


Snuffleupagus:
Why is this removing the `destroySandbox`-functionality, since I cannot imagine that it'll actually work correctly in the *general* case?

Please keep in mind that the `GENERIC` default viewer supports opening more than *one* document, and I'd thus expect that you need a unique scripting-instance for each document (the same way that basically everything in the default viewer is *document specific*); see PR #12695 which tries to improve a number of things related to both `pdf.sandbox.js` building and how it's used in the viewer.

brendandahl (approved):
I did a quick skim and seems reasonable. I'll let @Rob--W review the parts he suggested.

Rob--W:
> Why is this removing the `destroySandbox`-functionality, since I cannot imagine that it'll actually work correctly in the _general_ case?
> 
> Please keep in mind that the `GENERIC` default viewer supports opening more than _one_ document, and I'd thus expect that you need a unique scripting-instance for each document (the same way that basically everything in the default viewer is _document specific_); see PR #12695 which tries to improve a number of things related to both `pdf.sandbox.js` building and how it's used in the viewer.

I pointed out that the sandbox destruction logic wasn't called, at https://phabricator.services.mozilla.com/D91746?id=370264#inline-555164 with the ending note:

> Make this sandbox destruction part of the window. There is already an unload listener, you could call `destroySandbox` from there (in addition to any other existing calls in PDF.js code).

By this, I meant to call it at least twice: as part of the privileged unload handler (which always runs), and as usual from the destruction logic in PDF.js. @Snuffleupagus has already fixed the latter in #12695, so what remains is only on the m-c side (and also there already).

src/pdf.sandbox.js
```diff
@@ -102,8 +104,19 @@ class Sandbox {
 }
 
 function QuickJSSandbox() {
+  const handlers = {
+    callFunction(name, args) {},
+    clone(obj) {
+      return window.JSON.stringify(obj);
+    },
+  };
+  window.sandboxExtra = buildExtra(window, handlers);
```
Rob--W, line 113: Without a comment, it wasn't obvious at all why the implementation is like this.

But here it became apparent....: https://github.com/mozilla/pdf.js.quickjs/blob/f927cb7527c7cf0087872720aa151e7d4cf43460/src/myjs.js#L37-L46

Please add an explanatory comment when something non-obvious is done.

src/pdf.sandbox.js
```diff
@@ -24,66 +25,67 @@ const TESTING =
...
+    const code = [
       // Next line is replaced by code from initialization.js
       // when we create the bundle for the sandbox.
       PDFJSDev.eval("PDF_SCRIPTING_JS_SOURCE"),
-      `data = ${sandboxData};`,
-      `module.exports.initSandbox({ data, extra: {${extraStr}}, out: this});`,
-      "delete exports;",
-      "delete module;",
-      "delete data;",
+      `pdfjsScripting.initSandbox({ data: ${sandboxData}, extra: {${extraStr}}, out: this})`,
```
Rob--W, line 47: Is your intention of this to hide these extra helpers from the script? That is not accomplished by this implementation. Your code basically becomes:

```
pdfjsScripting.initSandbox({ // ...
  extra: {
    send, globalEval, ...
  },
})
```
Above syntax is "shorthand property names" - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015
In ES5 syntax, it is:
```
pdfjsScripting.initSandbox({ // ...
  extra: {
    send: send,
    globalEval: globalEval,
    // ...
  },
})
```

Now it should be more obvious that this basically reading off the global (since `send`, `globalEval`, etc. aren't default global variables.)

Rob--W (changes requested):
It took a while to review this because the patch is very large, with many more changes than strictly needed to provide the sandboxing functionality. The review was further complicated by the lack of documentation and references to relevant sources.

To support my review task, I composed the following list of locations that are relevant to the feature that I was asked to review (a sandbox mechanism). I also reviewed the other code since examples of actual usage is a good verification of the effectiveness of the API surrounding the sandbox.

- https://phabricator.services.mozilla.com/D91746 - patch for mozilla-central, with Firefox's implementation.
- https://github.com/mozilla/pdf.js.quickjs - PDF.js's bindings/glue for the quickjs sandbox
- https://github.com/mozilla/pdf.js/pull/12689 (this patch)
- https://github.com/mozilla/pdf.js/tree/master/src/scripting_api
- https://github.com/mozilla/pdf.js/blob/master/src/pdf.sandbox.js

There is no clear authorative reference for JavaScript in PDFs. The following are available (with an incomplete changelog).

- April 2007 - Adobe Acrobat SDK version 8.1
  https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf
- May 2015 - Adobe Acrobat DC SDK
  https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/AcrobatDC_js_api_reference.pdf
- HTML version of the 2015 document, e.g. specifically the `util.crackURL` method that you're adding to this patch:
  https://help.adobe.com/en_US/acrobat/acrobat_dc_sdk/2015/HTMLHelp/Acro12_MasterBook/JS_API_AcroJS/util_methods.htm

The last published reference is from 5 years ago, so they may be out of date. Interestingly, Chrome's PDF implementation has a README that references the 2007 implementation, at https://source.chromium.org/chromium/chromium/src/+/master:third_party/pdfium/fxjs/README;l=40;drc=d9d3ed281f87f706ecc442c7902abfa8ef5c4928

Could you also add a README.md to the `scripting_api` repository, to allow other readers to know what you're using as the definition of the API, and how everything fits together?

src/scripting_api/util.js
```diff
@@ -52,7 +52,44 @@ class Util extends PDFObject {
     if (typeof cURL !== "string") {
       throw new TypeError("First argument of util.crackURL must be a string");
     }
-    return this._crackURL(cURL);
+
+    const url = this._parseURL(cURL);
```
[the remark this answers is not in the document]
calixteman: I removed that stuff from this PR and I'll add it back in a follow-up.

src/scripting_api/initialization.js
```diff
@@ -35,47 +35,51 @@ import { Field } from "./field.js";
...
+     a functional api:
+     - the functions in ./extra.js
+     - send: to send data in the main window through an event
+     - globalEval: an eval function to evaluate a string in the global scope.
+     Unfortunatelly, something like:
+     globalThis.eval.call(globalThis, "const foo = 123;")
+     doesn't create a variable foo in the global scope.
+   */
+  const { data, extra, out } = params;
+  const proxyHandler = new ProxyHandler();
```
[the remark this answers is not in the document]
calixteman: Let's do that in an other PR.
Anyway it could be useful to add some telemetry.

src/pdf.sandbox.js
calixteman: These functions are global variables in the sandbox, for example:
https://github.com/mozilla/pdf.js.quickjs/blob/895a07683ae7dc71484ce59f493a5c2e52900a7b/src/quickjs_extra.c#L61-L63
These functions are wrappers to call their counterpart in pdf.sandbox.external.js.

src/pdf.sandbox.js
```diff
@@ -24,86 +25,124 @@ const TESTING =
...
-      "crackURL",
-    ];
-    const extraStr = extra.join(",");
-    let code = [
-      "exports = Object.create(null);",
-      "module = Object.create(null);",
+    const sandboxData = Sandbox.cloneIn(data);
+    // In the sandox, external functions like send, setTimeout, ... exist
+    // as global variables.
+    const external = `{ ${this._externalNames.join(",")} }`;
```
Rob--W, line 39: So in order to define an external X, you currently have to do:

1. add implementation named X to pdf.sandbox.external.js
2. declare X in `myjs.js`, twice: first the global, then again in `$externals` - https://github.com/mozilla/pdf.js.quickjs/blob/895a07683ae7dc71484ce59f493a5c2e52900a7b/src/myjs.js#L23
3. declare C function with name X for quickjs in quickjs_extra.c - https://github.com/mozilla/pdf.js.quickjs/blob/895a07683ae7dc71484ce59f493a5c2e52900a7b/src/quickjs_extra.c
4. reference externals.X from the actual JS implementation.

2 and 3 is an unfortunate amount of repetition, each with potential for bugs. Step 3 already has bugs: You're returning `JS_EXCEPTION` several times without setting an exception first, i.e. by `JS_ThrowTypeError(ctx, "message here");`. The effect of this is that the function won't actually throw an exception.

2 and 3 can be reduced to a zero-maintenance overhead + behavior equivalent to m-c, by adding a way to make generic functions from inside the sandbox to outside the sandbox, along with a good choice of a serialization and deserialization function. I suggest JSON serialization because a serializer is already available in quickjs's C code, and the corresponding deserializer is available in normal JS. Without JSON serialization, you would need considerably more code (like you have right now...) for no apparent benefit.

Here is an example where you'll maintain a table of callable methods in the `pdf.sandbox.js`, and allow the sandboxed code to call any of the methods in the table. This is not the only way to implement this concept, but I want to show at least one example for clarification.

```c
static JSValue js_callExternalFunc(JSContext *ctx, JSValueConst this_val,
                                int argc, JSValueConst *argv) {
  // TODO: function name (char*) from argv
  // TODO: rest of arguments OR just argv[1] as array, JS_JSONStringify it.
  // TODO: call external function with function name + args
  // TODO: Consider optional exception handling.
}
```

In myjs.js, you'll only need one export - that generic external function handler.

In pdf.sandbox.js, make that generic external function available. ensure that name is valid, parse args from JSON and call `externals[name](...args)`.

What I described above is the rudimentary version of how m-c's sandbox handles function calls of exported functions. The difference is that in my implementation (which matches how you're currently using it) is that values are copied rather than passed by a (mutable) reference. Since you're only passing numbers and strings, this distinction doesn't matter at all.

With the above logic, you could do something like 

`callExternalFunction("alert", ["msg"])` or
`callExternalFunction("alert", "msg")`, depending on your signature.

or specifically in this context:

```
var externals = {};
for (let ext of ${externals}) {
  externals[ext] = callExternalFunction.bind(null, ext);
}
delete callExternalFunction; // TODO: does this work? otherwise = null.
```

This can probably be made prettier, but at least it works with less maintenance cost than what you have right now.

src/pdf.sandbox.js
```diff
@@ -24,86 +25,124 @@ const TESTING =
...
-      "clearTimeout",
-      "setInterval",
-      "clearInterval",
-      "crackURL",
-    ];
-    const extraStr = extra.join(",");
-    let code = [
-      "exports = Object.create(null);",
-      "module = Object.create(null);",
+    const sandboxData = Sandbox.cloneIn(data);
```
Rob--W, line 36: This should have been `JSON.stringify`, not `Sandbox.cloneIn`. The result is used to concatenate with a string for JS code, not to pass as a parameter to sandboxed function.

src/pdf.sandbox.js
```diff
@@ -24,86 +25,124 @@ const TESTING =
...
   }
 }
 
 function QuickJSSandbox() {
+  const handlers = {
+    callFunction(name, args) {},
+    cloneIn: Sandbox.cloneIn,
+    cloneOut: Sandbox.cloneOut,
+  };
+  window.sandboxExternals = buildExternals(window, handlers);
```
Rob--W, line 130: This works, but I was wondering if there is another way to share values with the module without temporarily polluting the global namespace.

src/pdf.sandbox.js
```diff
@@ -24,86 +25,124 @@ const TESTING =
...
-      "delete module;",
-      "delete data;",
+      `pdfjsScripting.initSandbox({ data: ${sandboxData}, external: ${external} })`,
     ];
     if (!TESTING) {
-      code = code.concat(extra.map(name => `delete ${name};`));
       code.push("delete debugMe;");
+      // Don't expose external functions
+      code.push(this._externalNames.map(name => `delete ${name};`).join(";"));
+      delete this._externalNames;
```
Rob--W, line 50: This line doesn't do anything bad, but it's not doing anything useful either, except potentially preventing this `create` sandbox from being called twice (because a second call would trigger a runtime error when `.join()` is called). Is that what you've meant with adding this code? With all `delete` and so on around here, I just want to double-check that the code matches with your intentions.

Rob--W (changes requested):
I looked closely at the interaction of the sandbox and quickjs, a bit at the timer, but didn't look at the rest in detail. Since there is already something actionable, I'm returning it for review now.

web/generic_scripting.js
```diff
@@ -28,7 +28,7 @@ class GenericScripting {
 
   async createSandbox(data) {
     const sandbox = await this._ready;
-    sandbox.create(data);
+    return sandbox.create(data);
```
Snuffleupagus, line 31: The `pdf.sandbox.js` file *can* fail to load, if e.g. the option is incorrectly set or the file isn't available, and in that case the `this._ready` promise will be rejected.
That will cause the `if (!success) { ... }` check added in the `web/app.js` file to not be hit, since an error is thrown, and things won't then be clean-up as expected. To avoid this situation, I suggest modifying this method as follows:
```javascript
    try {
      const sandbox = await this._ready;
      return sandbox.create(data);
    } catch (ex) {
      return false;
    }
```

src/scripting_api/app.js
```diff
@@ -66,50 +79,58 @@ class App extends PDFObject {
...
+
+    if (expr) {
+      this._globalEval(expr);
     }
-    this._timeoutIds.set(timeout, [id, interval]);
+  }
+
+  _registerTimeout(callbackId, interval) {
+    const timeout = Object.create(null);
+    const id = { callbackId, interval };
```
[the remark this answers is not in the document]
calixteman: The two maps don't have the same goal:
 i) `_timeoutCallbackIds` is a Map::<int, string> which contains id and expression to evaluate;
 ii) `_timeoutIds` is a WeakMap::<Object, {int, bool}> which contains the "id" returned in a js script and the a callback id + is_interval to be able to kill & remove timeouts (or intervals).

src/pdf.sandbox.external.js
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+// In mozilla-central, this file is loaded as non-module script,
+// so it mustn't have any dependencies.
+
+class SandboxSupportBase {
+  /**
+   * @param {DOMWindow} win
+   */
+  constructor(win, sandbox) {
```
Rob--W, line 23: The `sandbox` parameter is not documented. By trying to document you may discover that it's not really straightforward to document it because the `sandbox` object is different here in quickjs and over there in m-c.

You're currently calling `this.sandbox.callFunction` directly, but the problem with this is that it forces users of `SandboxSupportBase` to declare a `callFunction` method on an undocumented parameter. This `SandboxSupportBase` class should not need to know anything about the sandbox implementation, as that could indicate a violation of the separation of concerns.

In my initial suggestion at https://github.com/mozilla/pdf.js/pull/12689#discussion_r542377524, I intentionally excluded `sandbox` from the signature of the `SandboxSupportBase`, so that users/subclasses can change the implementation without depending on `SandboxSupportBase` (and vice versa).

I suggest to do the same, by replacing `self.sandbox.callFunction` with `self.callSandboxFunction`, and letting subclasses defining their own implementation of `callSandboxFunction`, which could look like this:

```
constructor(win) {
  super(win);
  this.commFunc = null; // Will be assigned after the sandbox is initialized
}
callSandboxFunction(name, args) {
  args = this.support.exportValueToSandbox(args);
  try {
    this.commFunc(name, args);
  } catch (e) {
    Cu.reportError(e); // in m-c, or console.error(e) in quickjs
  }
}
```

note: I removed the typeof `name` check and the `commFunc` null check, which is present in your current implementation, because 1) name should never not be a string and 2) the function should never be called when `commFunc` is not set. Adding checks for these and silently ignoring them results in less robust code, because errors go unnoticed.

src/pdf.sandbox.js
```diff
@@ -23,87 +24,114 @@ const pdfjsBuild = PDFJSDev.eval("BUNDLE_BUILD");
...
 class Sandbox {
-  constructor(module) {
-    this._evalInSandbox = module.cwrap("evalInSandbox", null, [
-      "string",
-      "int",
-    ]);
-    this._dispatchEventName = null;
+  constructor(win, module) {
+    this.support = new SandboxSupport(win, this);
+    module.externalCall = this.support.createSandboxExternals();
```
Rob--W, line 43: PLEASE document this line here! I have asked multiple times for documentation of the non-obvious, and this is yet again another instance where this applies. Nobody who reads the code in this repository for the first time will have any clue about how this line connects to the code inside the sandbox because it is far from obvious.

Please add a comment from which it is obvious that this is used by https://github.com/mozilla/pdf.js.quickjs/blob/6836f77e568eba01e4a8f7f6e7bd666f73097892/src/myjs.js#L23-L36

This line can also benefit from some documentation:
https://github.com/mozilla/pdf.js.quickjs/blob/6836f77e568eba01e4a8f7f6e7bd666f73097892/src/pre.js#L21
... maybe even with a `console.error("externalCall called before being defined");` call to log an error to catch unexpected bugs.

and just wondering - without pre.js, would Emscripten eliminate "dead code" and prevent it from working? Or does it even work without the pre.js?

src/pdf.sandbox.external.js
```diff
@@ -0,0 +1,174 @@
...
+            result = externals[name](...args);
+          }
+        } else {
+          result = externals[name]();
+        }
+        if (!result) {
+          return null;
+        }
+      } catch (err) {
+        result = { error: err.message };
```
Rob--W, line 161: The current logic is prone to type confusion, as a return value of `{ error: ... , ... }` is indistinguishable from a caught actual error.

I don't see any matching error-handling (e.g. extracting + re-throwing an error) inside the sandbox either.

In m-c, you can instantiate an error object from the `Cu.andbox` global and throw it (and it will propagate back up into the sandbox).

In quickjs, if you can't do the same, then you'd be forced to do something else, for example serializing and deserializing it.

src/pdf.sandbox.js
```diff
@@ -23,87 +24,114 @@ const pdfjsBuild = PDFJSDev.eval("BUNDLE_BUILD");
...
-      "clearTimeout",
-      "setInterval",
-      "clearInterval",
-      "crackURL",
-    ];
-    const extraStr = extra.join(",");
-    let code = [
-      "exports = Object.create(null);",
-      "module = Object.create(null);",
+    const sandboxData = window.JSON.stringify(data);
```
Rob--W, line 51: Use `JSON` instead of `window.JSON`. `JSON` is part of the language definition. `window.xxxx` is more appropriate for host objects such as DOM APIs.

Rob--W (changes requested):
It's getting in a good shape. I do have a few comments, but the overall structure looks sane sane.

src/pdf.sandbox.external.js
calixteman: The only function which could throw here is `parseURL` and so it's planned to use the error here and rethrow an exception from its counterpart `crackURL` and then have a correct stack.

src/pdf.sandbox.js
Rob--W: You had marked this as resolved but didn't replace `window.JSON` with `JSON`. Why?

src/pdf.sandbox.js
```diff
@@ -23,87 +24,117 @@ const pdfjsBuild = PDFJSDev.eval("BUNDLE_BUILD");
...
   evalForTesting(code, key) {
     if (TESTING) {
-      this._evalInSandbox(
-        `try {
-           send({ id: "${key}", result: ${code} });
-         } catch (error) {
-           send({ id: "${key}", result: error.message });
-         }`,
-        this._alertOnError
+      this._evalInSandbox = this._module.ccall(
```
Rob--W, line 118: Remove `this._evalInSandbox = `. You're using `ccall`, not `cwrap`, so there is no return value.

web/app.js
```diff
@@ -1556,9 +1549,8 @@ const PDFViewerApplication = {
     const filename =
       this._contentDispositionFilename || getPDFFileNameFromURL(this.url);
 
-    scripting.createSandbox({
+    const success = await scripting.createSandbox({
```
Rob--W, line 1552: `createSandbox` is expected to succeed under normal circumstances. A try-catch or a `.catch` (or equivalently a `.then` handler would be more idiomatic than a single boolean.

src/pdf.sandbox.external.js
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * @param {Object} val - Import a value from the sandbox.
+   */
+  callSandboxFunction(name, args) {
+    try {
+      args = this.exportValueToSandbox(args);
+      this.commFun(name, args);
+    } catch (e) {
+      this.reportError(e);
```
Rob--W, line 66: Doesn't just `console.error(e);` work for both m-c and the web app? Let's do that instead. Keeping this file minimal makes it easier to audit.

src/pdf.sandbox.js
```diff
@@ -23,87 +24,117 @@ const pdfjsBuild = PDFJSDev.eval("BUNDLE_BUILD");
...
+    const code = [
       // Next line is replaced by code from initialization.js
       // when we create the bundle for the sandbox.
       PDFJSDev.eval("PDF_SCRIPTING_JS_SOURCE"),
-      `data = ${sandboxData};`,
-      `module.exports.initSandbox({ data, extra: {${extraStr}}, out: this});`,
-      "delete exports;",
-      "delete module;",
-      "delete data;",
+      `pdfjsScripting.initSandbox({ data: ${sandboxData}, keepCallExternalFunction: ${TESTING} })`,
```
Rob--W, line 63: Remove unused ` keepCallExternalFunction`

src/pdf.sandbox.external.js
Rob--W: I don't see how your reply addresses the type confusion that I pointed out.

If you're not doing anything with it right now, I suggest to change it to `throw err;` to get the patch in a reasonable state for merging.

Rob--W (approved):
r+ with comments addressed.

src/pdf.sandbox.js
calixteman: I just forgot.

PR #12693, opened 2020-12-04
[Regression] Prevent the *built* `pdf.scripting.js`/`pdf.sandbox.js` files from accidentally including most of the main-thread code (PR 12631 follow-up)

Snuffleupagus:
*This is a recent regression, which I stumbled upon while working on cleaning-up the gulpfile related to `pdf.sandbox.js` building.*

By placing the `ColorConverters` functionality in the `src/display/display_utils.js` file, you end up including a *significant* chunk of the `pdf.js` file in the built `pdf.scripting.js`/`pdf.sandbox.js` files.
Given that I cannot imagine that this was actually intended, since it inflates the built files with unnecessary/unused code, this moves `ColorConverters` to a new file instead (thus breaking the dependencies).

To hopefully reduce the risk future bugs, along these lines, a big comment is also placed at the top of the new file.
Finally, the `ColorConverters` is converted to a class with static methods, since this felt slightly cleaner overall.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -15,7 +15,6 @@
 
 import {
   addLinkAttributes,
-  ColorConverters,
   DOMSVGFactory,
   getFilenameFromUrl,
   LinkTarget,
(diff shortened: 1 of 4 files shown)


calixteman (approved):
Thanks for doing that.
I supposed that unused code was striped out, TIL.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12696, opened 2020-12-05
Fix non-standard quadpoints orders for annotations

### LINKED ISSUE #6811: Support quad points for annotations

### LINKED ISSUE #12675: Fix non-standard quadpoints orders for annotations
I have made a PDF file that serves as a minimal reduced test case for all annotation types with quadpoints support. Currently all markup annotations are supported for quadpoints (highlight, underline, strikethrough and squiggly annotations), but link annotations can also have quadpoints which we don't support yet. This can be done similar to the existing annotation types, but in my test file I actually ran into https://stackoverflow.com/questions/9855814/pdf-spec-vs-acrobat-creation-quadpoints which should be addressed first before we can do that.

In short:
- Handle https://stackoverflow.com/questions/9855814/pdf-spec-vs-acrobat-creation-quadpoints properly.
- Implement quadpoints support for link annotations.
- Include the reduced test case for all types.

timvandermeij:

Fixes #12675.
Fixes #6811.


src/core/annotation.js
```diff
@@ -227,7 +227,31 @@ function getQuadPoints(dict, rect) {
...
+  // mentioned on https://github.com/highkite/pdfAnnotate#QuadPoints), so
+  // this is the actual order we should work with. However, the situation is
+  // even worse since Adobe's own applications and other applications violate
+  // the specification and create annotations with other orders, namely top
+  // left, top right, bottom left and bottom right or even top left, top right,
+  // bottom right and bottom left. To avoid inconsistency and broken rendering,
+  // we normalize all lists to put the quadpoints in the same standard order
+  // (see https://stackoverflow.com/a/10729881).
+  return quadPointsLists.map(quadPointsList => {
+    const minX = Math.min(...quadPointsList.map(quadPoint => quadPoint.x));
```
calixteman, line 244: I think you could use something like
```js
quadPointsList.reduce(([mX, MX, mY, MY], p) => [Math.min(mX, p.x), Math.max(mY, p.y), Math.min(mY, p.y), Math.max(mY, p.y)]);
```
is slightly faster and more concise.
I discover few weeks ago that `reduce` is a way faster than `min+spread` (seef https://bugzilla.mozilla.org/show_bug.cgi?id=1674143)

src/core/annotation.js
```diff
@@ -227,7 +227,31 @@ function getQuadPoints(dict, rect) {
...
+  // and top left. However, in practice PDF files use a different order,
+  // namely bottom left, bottom right, top left and top right (this is also
+  // mentioned on https://github.com/highkite/pdfAnnotate#QuadPoints), so
+  // this is the actual order we should work with. However, the situation is
+  // even worse since Adobe's own applications and other applications violate
+  // the specification and create annotations with other orders, namely top
+  // left, top right, bottom left and bottom right or even top left, top right,
+  // bottom right and bottom left. To avoid inconsistency and broken rendering,
+  // we normalize all lists to put the quadpoints in the same standard order
+  // (see https://stackoverflow.com/a/10729881).
```
Snuffleupagus, line 242: Really good comment!

However, the fact that all of this special-casing is even necessary feels somewhat sad; why is it that Annotations in particular seem to deviate *so much* from the specification in many cases :-(

Snuffleupagus (approved):
Looks good to me, thank you!
Also, thanks for re-factoring the existing code, since this looks much better overall :-)

src/core/annotation.js
timvandermeij: I have changed this in the new commit. Thanks!

src/core/annotation.js
timvandermeij: I feel the same. I can only guess that implementation errors were made before and fixing it properly would mean breaking existing PDF files, so they had to go with a backwards compatible solution, forcing other tools to also consider those cases. However, the fact that Adobe applications violate their own specification also gets a sad face from me :-(

PR #12698, opened 2020-12-06
DnD a text from a text field doesn't remove text in the source

calixteman:
* aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1679872
 * Fix a bug in generic build: it wasn't possible to drop a text in a form

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -632,13 +632,26 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         storage.setValue(id, { value: event.target.value });
       });
 
+      let hasDragStarted = false;
       let blurListener = event => {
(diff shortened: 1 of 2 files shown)


timvandermeij:
I forgot about this patch, but given that this seems like a clear bug in Firefox, is also tracked upstream, and is pretty minor in my opinion, is it worth working around this here? I'm a bit afraid that such workarounds might not behave correctly in e.g., other browsers, and in general we try to keep browser-specific hacks to the bare minimum...

Snuffleupagus:
I completely agree with https://github.com/mozilla/pdf.js/pull/12698#issuecomment-860209657, since this really ought to be fixed in Firefox rather than worked-around in PDF.js; closing this PR for now.

PR #12700, opened 2020-12-06
Fix automatic zoom under spread mode

### LINKED ISSUE #12699: Automatic zoom not correct with odd or even spreads

Configuration:

1. open https://mozilla.github.io/pdf.js/web/viewer.html 
2. set spread to odd or even and zoom to automatic

image

image

calixteman:
Fixes #12699.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -725,6 +725,20 @@ class BaseViewer {
     }
   }
 
+  /**
+   * @private
+   */
+  get _pageWidthScaleFactor() {
+    if (
+      this.spreadMode !== SpreadMode.NONE &&
+      this.scrollMode !== ScrollMode.HORIZONTAL &&
+      !this.isInPresentationMode
+    ) {
+      return 2;
+    }
+    return 1;
+  }
+
   _setScale(value, noScroll = false) {
     let scale = parseFloat(value);
 
@@ -743,8 +757,9 @@ class BaseViewer {
         [hPadding, vPadding] = [vPadding, hPadding]; // Swap the padding values.
       }
       const pageWidthScale =
-        ((this.container.clientWidth - hPadding) / currentPage.width) *
-        currentPage.scale;
+        (((this.container.clientWidth - hPadding) / currentPage.width) *
+          currentPage.scale) /
+        this._pageWidthScaleFactor;
       const pageHeightScale =
         ((this.container.clientHeight - vPadding) / currentPage.height) *
         currentPage.scale;
@@ -1472,6 +1487,9 @@ class BaseViewer {
 
     if (!pageNumber) {
       return;
+    }
+    if (this._currentScaleValue && isNaN(this._currentScaleValue)) {
+      this._setScale(this._currentScaleValue, true);
     }
     this._setCurrentPageNumber(pageNumber, /* resetCurrentPageView = */ true);
     this.update();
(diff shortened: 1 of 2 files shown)


web/base_viewer.js
```diff
@@ -1475,6 +1481,7 @@ class BaseViewer {
     }
     this._setCurrentPageNumber(pageNumber, /* resetCurrentPageView = */ true);
     this.update();
+    this._setScale(this._currentScaleValue, true);
```
Snuffleupagus, line 1484: Please do exactly that same as in https://github.com/mozilla/pdf.js/blob/d784af3f38d5604b653c06fb40a6fe6a021d45d3/web/base_viewer.js#L1412-L1416 here as well, since this call shouldn't be needed unconditionally!

web/base_viewer.js
```diff
@@ -735,6 +735,11 @@ class BaseViewer {
       if (!currentPage) {
         return;
       }
+      const factor =
+        this.spreadMode === SpreadMode.NONE || this._isScrollModeHorizontal
+          ? 1
+          : 2;
```
Snuffleupagus, line 741: I don't believe that the above condition is correct in *all* situations, what about e.g. PresentationMode and/or the `PDFSinglePageViewer`?

To address all of this, let's add a new (private) getter above, and use that instead, e.g. something like:
```javascript
  /**
   * @private
   */
  get _pageWidthScaleFactor() {
    if (
      this.spreadMode !== SpreadMode.NONE &&
      this.scrollMode !== ScrollMode.HORIZONTAL &&
      !this.isInPresentationMode()
    ) {
      return 2;
    }
    return 1;
  }
```
and also the following dummy in `web/pdf_single_page_viewer.js`:
```javascript
  get _pageWidthScaleFactor() {
    return 1;
  }
```

Snuffleupagus (changes requested):

timvandermeij:
Those commits look similar. Should they be squashed into one?

calixteman:
> Those commits look related to each other. Should they be squashed into one?

Yes they can.

web/base_viewer.js
```diff
@@ -1475,6 +1491,9 @@ class BaseViewer {
     }
     this._setCurrentPageNumber(pageNumber, /* resetCurrentPageView = */ true);
     this.update();
+    if (this._currentScaleValue && isNaN(this._currentScaleValue)) {
```
timvandermeij, line 1494: If I look at https://github.com/mozilla/pdf.js/blob/d784af3f38d5604b653c06fb40a6fe6a021d45d3/web/base_viewer.js#L1412-L1416 this happens _before_ `_setCurrentPageNumber` and `update`. Shouldn't that be done here as well?

timvandermeij (commented):
Looks good to me with this final comment addressed.

web/base_viewer.js
Snuffleupagus: > Shouldn't that be done here as well?

That's correct, since otherwise there's a risk that the active page changes.

web/base_viewer.js
```diff
@@ -735,6 +749,7 @@ class BaseViewer {
       if (!currentPage) {
         return;
       }
+
```
Snuffleupagus, line 752: Nit: Unrelated change, please revert.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #12701, opened 2020-12-07
Add events pageopen & pageclosed to be used with JS actions (PageOpen and PageClose)

calixteman:
…

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -207,6 +207,7 @@ class BaseViewer {
     this.scroll = watchScroll(this.container, this._scrollUpdate.bind(this));
     this.presentationModeState = PresentationModeState.UNKNOWN;
     this._onBeforeDraw = this._onAfterDraw = null;
+    this._onAfterPageDraw = null;
     this._resetView();
 
     if (this.removePageBorders) {
@@ -281,6 +282,7 @@ class BaseViewer {
     if (!(0 < val && val <= this.pagesCount)) {
       return false;
     }
+
     this._currentPageNumber = val;
 
     this.eventBus.dispatch("pagechanging", {
@@ -292,6 +294,25 @@ class BaseViewer {
     if (resetCurrentPageView) {
       this._resetCurrentPageView();
     }
+
+    if (this._pages[val - 1].renderingState === RenderingStates.FINISHED) {
+      this.eventBus.dispatch("pageopen", {
+        source: this,
+        pageNumber: val,
+      });
+    } else if (!this._onAfterPageDraw) {
+      this._onAfterPageDraw = evt => {
+        if (evt.pageNumber !== this._currentPageNumber) {
+          return;
+        }
+        this.eventBus.dispatch("pageopen", {
+          source: this,
+          pageNumber: this._currentPageNumber,
+        });
+      };
+      this.eventBus._on("pagerendered", this._onAfterPageDraw);
+    }
+
     return true;
   }
 


web/base_viewer.js
```diff
@@ -292,6 +298,25 @@ class BaseViewer {
...
+      this.eventBus.dispatch("pageopen", pageOpenEvent);
+    } else {
+      const onAfterDraw = evt => {
+        if (evt.pageNumber !== val) {
+          return;
+        }
+        this.eventBus._off("pagerendered", onAfterDraw);
+        this.eventBus.dispatch("pageopen", pageOpenEvent);
+      };
+      this.eventBus._on("pagerendered", onAfterDraw);
```
Snuffleupagus, line 317: This is generally problematic, since there's absolutely no guarantee that the "new" page will ever be rendered; e.g. when scrolling *quickly* through a long document.

Hence you can easily end up in situation where you're have a *very large* number of "pagerendered" event listeners that won't ever get removed.

Snuffleupagus (changes requested):
Is it really not possible to use the *existing* events instead, rather than adding a bunch of new ones?

If would be most helpful with additional context to really understand *exactly* what you're attempting to achieve here!

calixteman:
> Is it really not possible to use the _existing_ events instead, rather than adding a bunch of new ones?
> 
> If would be most helpful with additional context to really understand _exactly_ what you're attempting to achieve here!

My goal to execute actions on PageOpen and PageClose events:
https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/AcrobatDC_js_api_reference.pdf#page=243

So the PageClose occurs just after the page number changed and PageOpen after the drawing of the new page.

web/base_viewer.js
calixteman: Any hints or ideas to avoid such situation ?

web/base_viewer.js
```diff
@@ -281,6 +281,12 @@ class BaseViewer {
     if (!(0 < val && val <= this.pagesCount)) {
       return false;
     }
+
+    this.eventBus.dispatch("pageclose", {
+      source: this,
+      pageNumber: this._currentPageNumber,
+    });
+
```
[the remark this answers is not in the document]
calixteman: I did that but I thought it was more clear to have a separate event.
I'm fine to append a previous page number.

calixteman:
@Snuffleupagus, in order to not introduce a new event (i.e. a pageopen dispatched after the page has been rendered) I moved the dispatch of pagechanging (just few lines after). 
Is it breaking something ?

Snuffleupagus:
> Is it breaking something ?

Yes, unfortunately this is way worse than the previous patch:
 - First of all, note that `_previousPageNumber` is already used by `PDFSinglePageViewer`, in a different way, and what you're doing here thus breaks that completely.
 - Secondly, potentially delaying "pagechanging" until after rendering (or potentially not firing it at all) is bound to break any number of things, e.g. the toolbar will no longer update when it's supposed to and who known what else.

There's been less than 24 hours since I first saw this patch, and I've not yet had time to come up with a better approach (being that I'm doing PDF.js development in my spare time).
Please give me some time to think this through properly, such that we cover all cases correctly and don't regress something else in the process. (Looking at the specification I suppose that we still may need new events here, but exactly how this best fits into the existing PDF.js code-base I'm not yet sure of.)

Snuffleupagus:
Sorry about the delay here; I've been thinking about this a bit more, and based on looking at the specification I'm no longer that opposed to adding new events here. (My previous objection was based on the initial lack of context, since it's difficult to really understand the problem-space.)

However, I'd suggest a slightly modified implementation; please see my suggesting at https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:pageopen-pageclose

calixteman:
@Snuffleupagus I think your patch is a way better than mine, so please make a PR, thanks.

PR #12702, opened 2020-12-07
JS - Collect and execute actions at doc level

calixteman:
* the goal is to execute actions like Open or OpenAction
 * can be tested with issue6106.pdf (auto-print)
 * once #12701 is merged, we can add page actions

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -19,6 +19,7 @@ import {
   bytesToString,
   createPromiseCapability,
   createValidAbsoluteUrl,
+  DocumentActionEventType,
   FormatError,
   info,
   InvalidPDFException,
@@ -47,13 +48,14 @@ import {
   RefSet,
   RefSetCache,
 } from "./primitives.js";
-import { Lexer, Parser } from "./parser.js";
 import {
+  collectActions,
   MissingDataException,
   toRomanNumerals,
   XRefEntryException,
   XRefParseException,
 } from "./core_utils.js";
+import { Lexer, Parser } from "./parser.js";
 import { CipherTransformFactory } from "./crypto.js";
 import { ColorSpace } from "./colorspace.js";
 import { GlobalImageCache } from "./image_utils.js";
@@ -873,11 +875,11 @@ class Catalog {
     return shadow(this, "attachments", attachments);
   }
 
-  get javaScript() {
+  _collectJavaScript() {
     const obj = this._catDict.get("Names");
 
     let javaScript = null;
-    function appendIfJavaScriptDict(jsDict) {
+    function appendIfJavaScriptDict(name, jsDict) {
       const type = jsDict.get("S");
       if (!isName(type, "JavaScript")) {
         return;
@@ -890,10 +892,10 @@ class Catalog {
         return;
       }
 
-      if (!javaScript) {
-        javaScript = [];
+      if (javaScript === null) {
+        javaScript = Object.create(null);
       }
-      javaScript.push(stringToPDFString(js));
+      javaScript[name] = stringToPDFString(js);
     }
 
     if (obj && obj.has("JavaScript")) {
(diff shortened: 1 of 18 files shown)


src/display/api.js
```diff
@@ -753,6 +753,15 @@ class PDFDocumentProxy {
     return this._transport.getJavaScript();
   }
 
+  /**
+   * @returns {Promise<Object | null>} A promise that is resolved with
+   *   an {Object} with the JavaScript actions, or `null`
+   *   if no JavaScript exists.
+   */
+  getJSActions() {
```
Snuffleupagus, line 761: Given how similar this is to the existing `getJavaScript` method just above, there should probably be some kind of (brief) explanation of the differences here.

src/display/api.js
```diff
@@ -1124,6 +1133,18 @@ class PDFPageProxy {
     return this.annotationsPromise;
   }
 
+  /**
+   * @param {GetAnnotationsParameters} params - Annotation parameters.
+   * @returns {Promise<Array<any>>} A promise that is resolved with an
+   *   {Array} of the annotation objects.
+   */
+  getJSActions() {
+    if (!this.jsActionsPromise) {
```
Snuffleupagus, line 1142: This should be `this._jsActionsPromise` instead, and please also remember to reset it with `this._jsActionsPromise = null;`-lines after all existing `this.annotationsPromise = null;`-lines in this file.

web/app.js
```diff
@@ -1009,12 +1009,26 @@ const PDFViewerApplication = {
       return;
     }
 
+    if (AppOptions.get("enableScripting") && this._scriptingInstance) {
```
Snuffleupagus, line 1012: Here, and elsewhere, it should suffice to check `if (this._scriptingInstance) {` since that cannot happen without the "enableScripting"-option being set.

web/app.js
```diff
@@ -1551,8 +1569,36 @@ const PDFViewerApplication = {
...
       },
     });
+
+    scripting.dispatchEventInSandbox({
+      id: "doc",
+      name: "Open",
+    });
+  },
+
+  execPageActions(pageNumber, isClosed) {
```
Snuffleupagus, line 1582: Nothing seems to call this method, is this actually necessary?

src/core/obj.js
```diff
@@ -877,7 +879,7 @@ class Catalog {
     const obj = this._catDict.get("Names");
 
     let javaScript = null;
-    function appendIfJavaScriptDict(jsDict) {
+    function appendIfJavaScriptDict(name, jsDict) {
```
Snuffleupagus, line 882: Unless I'm mistaken, the changes here will affect the existing format used for the `getJavaScript` method in the API; which constitutes a *breaking* API-change and that's very much not allowed!

The fact that you had to change an existing unit-test is definitely a red-flag, and you've most likely also broken the `_initializeAutoPrint` method in the viewer with these changes.

web/app.js
```diff
@@ -1385,9 +1399,11 @@ const PDFViewerApplication = {
...
     });
 
-    pagesPromise.then(() => {
-      this._initializeAutoPrint(pdfDocument, openActionPromise);
-    });
+    if (!AppOptions.get("enableScripting")) {
+      pagesPromise.then(() => {
+        this._initializeAutoPrint(pdfDocument, openActionPromise);
+      });
+    }
```
Snuffleupagus, line 1406: This doesn't look correct, since that will also end up ignoring the openAction-data.

Please revert these changes, and instead change https://github.com/mozilla/pdf.js/blob/d784af3f38d5604b653c06fb40a6fe6a021d45d3/web/app.js#L1592 to
```javascript
    if (javascript && !AppOptions.get("enableScripting")) {
```

---

Also, something I just realized, if you're now relying on actually executing JavaScript to *trigger printing on load* you probably want to change https://github.com/mozilla/pdf.js/blob/d784af3f38d5604b653c06fb40a6fe6a021d45d3/web/app.js#L1480 to 
```javascript
  this.pdfViewer.pagesPromise.then(() => {
    this.triggerPrinting();
  }
```
since otherwise there's a risk that it will just be ignored.

Snuffleupagus (changes requested):

brendandahl (approved):

PR #12703, opened 2020-12-07
Ensure that the `pdf.sandbox.js` is removed from the DOM on destroy, and unbreak the Chromium-extension (PR 12695 follow-up)

Snuffleupagus:


brendandahl:
I think this will collide with #12689.

web/generic_scripting.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+import { AppOptions } from "./app_options.js";
+import { loadScript } from "pdfjs-lib";
+
+class GenericScripting {
+  constructor() {
+    this._scriptElement = null;
+    this._ready = loadScript(AppOptions.get("sandboxBundleSrc")).then(evt => {
+      this._scriptElement = evt?.target;
```
Rob--W, line 23: What's the purpose of saving the script reference? The existence of a `<script>` element in the DOM is only significant between insertion and execution. After a script has loaded, there is not really a purpose for it any more.

The only situation where it could make a difference is if the script grabs a reference to the `<script>` tag (e.g. `document.currentScript` at the start of the script) and actively watches the node. That's not likely.

Rob--W (commented):
This change looks reasonable and smaller than #12689, so I think that it would be better to land this first. The other patch is not close to landing yet.

web/generic_scripting.js
Snuffleupagus: > What's the purpose of saving the script reference?

Given that the `loadScript` function inserts the script into the DOM, see https://github.com/mozilla/pdf.js/blob/b194c820bfbf30734589bb37f7f632bcc7790352/src/display/display_utils.js#L531-L546 my (possibly flawed) thinking was that we should make sure to clean-up that up as well.

> After a script has loaded, there is not really a purpose for it any more.

Unless I'm completely misunderstanding you, are you actually suggesting that we can simply remove the script from the DOM already in the `onload` handler?

Rob--W: > Unless I'm completely misunderstanding you, are you actually suggesting that we can simply remove the script from the DOM already in the `onload` handler?

Yep exactly. The script has run, it does not serve any other purpose after that, other than the potential case that I described before, or to help those who inspect the DOM to see that the script has been inserted. In other words, removing it from `onload` in `loadScript` should be fine.

Snuffleupagus: Thanks for your help!

I've updated the PR to address this, however not unconditionally since I didn't fancy going through all existing `loadScript` call-sites.

timvandermeij (approved):

timvandermeij:
Looks like a good set of improvements to me too. Thanks!

PR #12711, opened 2020-12-09
Renderer - Draw text if a font can't be loaded

dsherret:
Wanted to at least get the conversation started about #4244. I don't really understand the code in this area, but here's a patch that at least draws text if a font can't be loaded which was important for my purposes.

### DIFF
--- a/src/display/font_loader.js
+++ b/src/display/font_loader.js
@@ -354,7 +354,7 @@ class FontFaceObject {
       fontRegistry = null,
     }
   ) {
-    this.compiledGlyphs = {};
+    this.compiledGlyphs = Object.create(null);
     // importing translated data
     for (const i in translatedData) {
       this[i] = translatedData[i];
@@ -394,7 +394,7 @@ class FontFaceObject {
   }
 
   tryGetPathGenerator(objs, character) {
-    if (this.compiledGlyphs.hasOwnProperty(character)) {
+    if (this.compiledGlyphs[character] !== undefined) {
       return this.compiledGlyphs[character];
     }
 
@@ -412,7 +412,7 @@ class FontFaceObject {
       }
       warn(`tryGetPathGenerator - ignoring character: "${ex}".`);
 
-      return (this.compiledGlyphs[character] = undefined);
+      return (this.compiledGlyphs[character] = null);
     }
 
     // If we can, compile cmds into JS for MAXIMUM SPEED...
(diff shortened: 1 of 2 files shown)


src/display/font_loader.js
```diff
@@ -354,7 +354,7 @@ class FontFaceObject {
       fontRegistry = null,
     }
   ) {
-    this.compiledGlyphs = Object.create(null);
```
timvandermeij, line 357: `Object.create(null)` is used on purpose to get an object without prototype, so please keep that.

src/display/font_loader.js
```diff
@@ -393,8 +393,8 @@ class FontFaceObject {
     return rule;
   }
 
-  getPathGenerator(objs, character) {
-    if (this.compiledGlyphs[character] !== undefined) {
+  tryGetPathGenerator(objs, character) {
+    if (this.compiledGlyphs.hasOwnProperty(character)) {
```
timvandermeij, line 397: `hasOwnProperty` checks tend to be more expensive, and `paintChar` where this is called seems to be called a lot, so let's try to avoid using that.

timvandermeij (commented):
Some drive-by comments, but I don't know enough about this particular part of the codebase to review it completely.

src/display/font_loader.js
dsherret: Sure. I reverted to the old way, but I think it is less clear than using `hasOwnProperty`. I believe performance-wise it doesn't matter here which is used https://www.measurethat.net/Benchmarks/Show/1303/0/objecthasownproperty-vs-object-in-vs-object-vs-arrayind#latest_results_block

src/display/font_loader.js
dsherret: I think it doesn't matter performance-wise (since many fonts won't be created... or is there some other reason for doing this?), but I reverted since I changed the rest of the code to use a way for `hasOwnProperty` to not be necessary.

timvandermeij: One of the reasons this is done is because PDF.js is often used in combination with other libraries which sometimes alter/pollute the object's prototype, which has caused bugs before. Even though it's those libraries' fault for polluting the object's prototype, PDF.js is sometimes blamed for it since "it doesn't work with library X". We don't need the prototype in these cases, so it's best not to use it to avoid problems.

dsherret: I see and makes sense. That sucks though haha

PR #12712, opened 2020-12-09
Attempt to handle collapsed outline items, in the default viewer, according to the specification (issue 12704, PR 10890 follow-up)

### LINKED ISSUE #12704: Outline items are not opened/closed correctly
Attach (recommended) or Link to PDF file here: [outline.pdf](https://github.com/mozilla/pdf.js/files/5658970/outline.pdf)

Configuration:
- Web browser and its version: Firefox (83.0)
- PDF.js version: 2.7.345
- Is a browser extension: no (but should be independent)

1. Open the PDF in PDF.js Viewer
2. Open the bookmark outline

grafik

What went wrong? 
grafik
`C.A` should be closed.

For me it looks like an issue which is caused only if the node that should be closed indcludes "open" nodes. This looks like a wrong interpretation of the `/Count` entry in outline items.

Snuffleupagus:
This patch *attempts* to actually implement what's described for the `Count`-entry in the PDF specification, see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G11.2095911, which I mostly ignored back in PR #10890 since it seemed unnecessarily complicated[1].

Besides issue #12704, I've also tested a couple of other documents (e.g. the PDF specification) and these changes don't *seem* to break anything else; additional testing would be helpful though!

Fixes #12704

---
[1] At the time, all PDF documents that I tested worked even with a very simple approach and I thus hoped that it'd would suffice.

### DIFF
--- a/web/pdf_outline_viewer.js
+++ b/web/pdf_outline_viewer.js
@@ -95,7 +95,23 @@ class PDFOutlineViewer extends BaseTreeViewer {
    * @private
    */
   _addToggleButton(div, { count, items }) {
-    const hidden = count < 0 && Math.abs(count) === items.length;
+    let hidden = false;
+    if (count < 0) {
+      let totalCount = items.length;
+      if (totalCount > 0) {
+        const queue = [...items];
+        while (queue.length > 0) {
+          const { count: nestedCount, items: nestedItems } = queue.shift();
+          if (nestedCount > 0 && nestedItems.length > 0) {
+            totalCount += nestedItems.length;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me, also in my testing with some other PDF files. I did have to re-read the specification a couple of times to actually understand what they mean, so I share the feeling that the idea of the `Count` property is perhaps more complex than strictly necessary :-)

PR #12718, opened 2020-12-10
Ignore color-operators in Type3 glyphs beginning with a `d1` operator (issue 12705)

### LINKED ISSUE #12705: Type3 font is not rendered
Attach (recommended) or Link to PDF file here: [issue.pdf](https://github.com/mozilla/pdf.js/files/5659516/issue.pdf)

Configuration:
- Web browser and its version: Firefox (83.0)
- PDF.js version: 2.7.345

1. Open attached PDF
2. You see a white page

grafik

Nothing is rendered but the text layer (selected for demonstration only):
grafik

Snuffleupagus:
Please refer to the PDF specification at https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G8.1977497 and https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G7.3998470

This patch removes the color-operators in the evaluator, since that should be more efficient than doing it repeatedly in the main-thread when rendering the Type3 glyphs.

Fixes #12705


src/core/evaluator.js
```diff
@@ -3612,7 +3612,16 @@ class TranslatedFont {
...
-          .then(function () {
+          .then(() => {
+            // According to the PDF specification, section "9.6.5 Type 3 Fonts"
+            // and "Table 113":
+            //  "A glyph description that begins with the d1 operator should
+            //   not execute any operators that set the colour (or other
+            //   colour-related parameters) in the graphics state;
+            //   any use of such operators shall be ignored."
+            if (operatorList.fnArray[0] === OPS.setCharWidthAndBounds) {
+              this._removeType3ColorOperators(operatorList);
```
brendandahl, line 3623: The spec also mentions we should ignore some `gs` entries. It looks like the only one we support is `TR`, but it shouldn't be too hard to add an option to the evaluator to skip that or remove it here.

brendandahl (changes requested):

Ridho-Ilahi:
Why am i always get this email?

Pada tanggal Kam, 10 Des 2020 20:50, pdf.js test bot <
notifications@github.com> menulis:

src/core/evaluator.js
Snuffleupagus: > The spec also mentions we should ignore some `gs` entries.

The reason that I didn't add this originally was because I had no test-case where it mattered, and it seemed easier to just ignore it.

> It looks like the only one we support is `TR`, but it shouldn't be too hard to add an option to the evaluator to skip that or remove it here.

The updated patch ignores all listed types, not only `TR`, to prevent a future bug if/when we support more of these.

Also, given that this only applies to Type3 glyphs I didn't want add special-casing to the general GState handling, which is why I opted to simply remove them after the fact here.

brendandahl (approved):

timvandermeij (approved):

timvandermeij:
Looks good!

PR #12719, opened 2020-12-10
JS -- add function eMailValidate used to validate an email address

calixteman:
This function is used in this pdf:
https://www2.illinois.gov/rev/forms/reg/Documents/reg-1-mr.pdf

### DIFF
--- a/src/scripting_api/aform.js
+++ b/src/scripting_api/aform.js
@@ -37,6 +37,14 @@ class AForm {
       "m/d/yy HH:MM",
     ];
     this._timeFormats = ["HH:MM", "h:MM tt", "HH:MM:ss", "h:MM:ss tt"];
+
+    // The e-mail address regex below originates from:
+    // https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
+    this._emailRegex = new RegExp(
+      "^[a-zA-Z0-9.!#$%&'*+\\/=?^_`{|}~-]+" +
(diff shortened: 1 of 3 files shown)


src/scripting_api/aform.js
```diff
@@ -37,6 +37,11 @@ class AForm {
       "m/d/yy HH:MM",
     ];
     this._timeFormats = ["HH:MM", "h:MM tt", "HH:MM:ss", "h:MM:ss tt"];
+
+    // https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
+    // The following regex has been copy/paste from the above link.
```
timvandermeij, line 42: I think the following would be a bit easier to read, with the URL below the description:

```javascript
# The e-mail address regex below originates from:
# https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
```

test/unit/scripting_spec.js
```diff
@@ -1202,5 +1202,21 @@ describe("Scripting", function () {
         }
       });
     });
+
+    describe("eMailValidate", function () {
+      it("should validate an email address", function (done) {
```
timvandermeij, line 1207: `email` -> `e-mail`

timvandermeij (commented):
Looks good to me with these comments addressed.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12722, opened 2020-12-10
JS -- fix printd issue with negative number

calixteman:


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #12707, opened 2020-12-08
Checkboxes with the same name must behave like a radio buttons group

calixteman:
Fixes #12706.

PR #12724, opened 2020-12-10
Follow-up of #12707: Add an integration test for checkboxes as radio …

calixteman:
…buttons

 * Integration tests: Add a function to load a pdf and wait for a selected element
 * Integration tests: Add a function to close all the open pages


test/integration/commons.js
```diff
@@ -0,0 +1,36 @@
+/* Copyright 2020 Mozilla Foundation
```
timvandermeij, line 1: Let's rename this file to `test_utils.js` to match the same kind of file we already have for the unit tests. It's also a bit more descriptive that it contains utility functions for the tests themselves.

test/integration/annotation_spec.js
```diff
@@ -13,49 +13,72 @@
...
     it("must show a popup on mouseover", async () => {
       await Promise.all(
-        pages.map(async page => {
+        pages.map(async ([name, page]) => {
           let hidden = await page.$eval(
             "[data-annotation-id='21R']",
             el => el.hidden
           );
-          expect(hidden).toEqual(true);
+          expect(hidden).withContext(`In ${name}`).toEqual(true);
```
timvandermeij, line 40: What does `withContext` do, and what is the advantage of using it here?

test/integration/annotation_spec.js
```diff
@@ -13,49 +13,72 @@
...
+
+    beforeAll(async () => {
+      pages = await loadAndWait("issue12706.pdf", "[data-annotation-id='63R']");
+    });
+
+    afterAll(async () => {
+      await closeAll(pages);
+    });
+
+    it("must click on different checkboxes and they must behave like radio buttons", async () => {
```
timvandermeij, line 66: `must let checkboxes with the same name behave like radio buttons` is a bit shorter and also tells us when this behavior happens.

test/integration/annotation_spec.js
```diff
@@ -13,49 +13,72 @@
...
+      const selectors = [63, 70, 79].map(n => `[data-annotation-id='${n}R']`);
+      await Promise.all(
+        pages.map(async ([name, page]) => {
+          for (const sel of selectors) {
```
timvandermeij, line 70: `sel` -> `selector` and `otherSel` -> `otherSelector` for readability in this function.

test/integration/commons.js
```diff
@@ -0,0 +1,36 @@
...
+      );
+      await page.bringToFront();
+      await page.waitForSelector(selector, {
+        timeout: 0,
+      });
+      return [session.name, page];
+    })
+  );
+
+exports.closeAll = pages =>
```
timvandermeij, line 31: Maybe call this `closePages` and the other one above `loadPagesAndWait` so they have more similar names and it's clear it involves the pages.

test/integration/annotation_spec.js
calixteman: this function will prepend the string in case of error in the output.
So here it allows to get the name of the browser in case of error.

test/integration/annotation_spec.js
```diff
@@ -13,49 +13,72 @@
...
+
+    beforeAll(async () => {
+      pages = await loadAndWait("issue12706.pdf", "[data-annotation-id='63R']");
+    });
+
+    afterAll(async () => {
+      await closePages(pages);
+    });
+
+    it("must let checkboxes with the same browserName behave like radio buttons", async () => {
```
timvandermeij, line 66: The `browserName` here seems like a find/replace gone wrong since it's the element names of the radio buttons that should be the same here, so `browserName` -> `name` is needed here.

timvandermeij (approved):

timvandermeij:
Nice improvements!

PR #12725, opened 2020-12-10
Use widths defined by font for standard fonts.

brendandahl:
There doesn't seem to be anything definitive about this in
the spec, but from experimenting, it seems acrobat lets
PDFs override the widths of the standard fonts.


Snuffleupagus (approved):
Lots of movement in existing tests, but it seems to split into two categories:
 - Changes which are *virtually* imperceivable to the naked eye (most cases fall into this category).
 - Small rendering improvements.

Looks good to me, thank you!

timvandermeij (approved):

timvandermeij:
Thanks!

/botio makeref

PR #12726, opened 2020-12-11
[api-minor] Include and use the 14 standard font files.

brendandahl:
Fixes #4244 (at least for the standard font issue).

I need to do some more testing to see if this impacts performance for the browser use case. I'm also considering disabling this feature for browsers so we use the builtin fonts. Though, we may want to use the various symbol fonts since those seem to vary per platform.


src/core/evaluator.js
```diff
@@ -387,6 +388,48 @@ class PartialEvaluator {
     return data;
   }
 
+  async fetchStandardFontData(name) {
```
Snuffleupagus, line 391: Similar to the `fetchBuiltInCMap` above, please cache the *raw* font data using a `standardFontCache` (see the `builtInCMapCache` implementation) to avoid having to repeatedly fetch the same font data over and over.

src/core/evaluator.js
```diff
@@ -3376,13 +3419,21 @@ class PartialEvaluator {
...
           flags,
           firstChar,
           lastChar,
         };
         const widths = dict.get("Widths");
+        const standardFontName = getStandardFontName(baseFontName);
+        let file = null;
+        if (standardFontName) {
+          const data = await this.fetchStandardFontData(standardFontName);
+          file = new Stream(data);
```
Snuffleupagus, line 3434: Please let `fetchStandardFontData` return the data as a `Stream`, rather than doing that manually at the call-sites.

src/core/fonts.js
```diff
@@ -243,6 +243,12 @@ function getFontType(type, subtype) {
   }
 }
 
+function getStandardFontName(name) {
+  const fontName = name.replace(/[,_]/g, "-").replace(/\s/g, "");
```
Snuffleupagus, line 247: Since you're using the same regular expression at https://github.com/mozilla/pdf.js/blob/00b4f86db301546d21d5dc79c61e2a9a829667a4/src/core/fonts.js#L1319
can you please extract the regexp-part to a new constant, used by both call-sites, to avoid these getting out of sync?

src/display/api.js
```diff
@@ -2085,6 +2096,10 @@ class WorkerTransport {
       isCompressed: params.cMapPacked,
     });
 
+    this.StandardFontDataFactory = new params.StandardFontDataFactory({
+      baseUrl: params.standardFontDataUrl,
+    });
```
Snuffleupagus, line 2101: The new parameters, i.e. `standardFontDataUrl` and `StandardFontDataFactory`, should be added to https://github.com/mozilla/pdf.js/blob/00b4f86db301546d21d5dc79c61e2a9a829667a4/src/display/api.js#L109-L183

src/display/display_utils.js
```diff
@@ -125,56 +175,44 @@ class BaseCMapReaderFactory {
...
-        let cMapData;
-        if (this.isCompressed) {
-          cMapData = new Uint8Array(await response.arrayBuffer());
-        } else {
-          cMapData = stringToBytes(await response.text());
-        }
-        return { cMapData, compressionType };
-      });
-    }
+    return fetchData(url, this.isCompressed).then(data => {
```
Snuffleupagus, line 178: Nit: 
```suggestion
    return fetchData(url, /* asTypedArray = */ this.isCompressed).then(data => {
```

src/display/display_utils.js
```diff
@@ -125,56 +175,44 @@ class BaseCMapReaderFactory {
...
-      request.send(null);
+    return this._fetchData(url).catch(reason => {
+      throw new Error(`Unable to load font data at: ${url}`);
     });
   }
 }
 
+class DOMStandardFontDataFactory extends BaseStandardFontDataFactory {
+  _fetchData(url) {
+    return fetchData(url, true);
```
Snuffleupagus, line 212: Nit: 
```suggestion
    return fetchData(url, /* asTypedArray = */ true);
```

web/app_options.js
```diff
@@ -178,6 +178,14 @@ const defaultOptions = {
         : "../web/cmaps/",
     kind: OptionKind.API,
   },
+  standardFontDataUrl: {
```
Snuffleupagus, line 181: Please move this further down, since the parameters for *each* `OptionKind` should be sorted alphabetically.

Snuffleupagus (changes requested):
Overall this seems like a very reasonable approach, and the re-use between the CMap/StandardFont factories is really nice.
Besides a few inline comments, based on a quick look at the code, I've got one "bigger picture" suggestion below.

> I'm also considering disabling this feature for browsers so we use the builtin fonts. Though, we may want to use the various symbol fonts since those seem to vary per platform.

My *general* train of thought here (ignoring the symbol fonts for now), is that we'd only fetch/use the standard font files when *actually* necessary. In practice, that'd mean:
 - When the `disableFontFace` API-parameter is `true`.
 - When handling Patterns, see e.g. https://github.com/mozilla/pdf.js/blob/00b4f86db301546d21d5dc79c61e2a9a829667a4/src/core/evaluator.js#L855
 - When failing to load a font and we thus fallback, see e.g. https://github.com/mozilla/pdf.js/blob/00b4f86db301546d21d5dc79c61e2a9a829667a4/src/core/evaluator.js#L856

---

*Edit:* Thinking about my suggestion a bit more, it seems that conditional loading will probably be somewhat difficult to implement well given the existing font-parsing code and the multitude of cases to consider. However, with caching implemented in the worker (note the inline comment), I'm not sure how much of a problem always loading these font files would really be!?

(Back when I added worker-thread caching of CMap-files, overall performance improved significantly for affected documents.)

src/core/evaluator.js
```diff
@@ -387,6 +388,48 @@ class PartialEvaluator {
...
+      "Helvetica-Bold": "FoxitSansBold",
+      "Helvetica-BoldOblique": "FoxitSansBoldItalic",
+      "Helvetica-Oblique": "FoxitSansItalic",
+      "Times-Roman": "FoxitSerif",
+      "Times-Bold": "FoxitSerifBold",
+      "Times-BoldItalic": "FoxitSerifBoldItalic",
+      "Times-Italic": "FoxitSerifItalic",
+      Symbol: "FoxitSymbol",
+      ZapfDingbats: "FoxitDingbats",
+    };
```
Snuffleupagus, line 407: It might not hurt to move this to https://github.com/mozilla/pdf.js/blob/master/src/core/standard_fonts.js instead!?

Snuffleupagus:
Given the sheer number of long-standing issues/bugs that this PR would solve, this seems like something that would be *really* nice to try and fix-up and land sooner rather than later :-)
@brendandahl Do you have time to finish this, such that we can land the PR?

---

Also, we're probably overdue for a new PDF.js release at this point; however I'd suggest that we hold off on that a little while longer since *this* PR would be really nice to include in the next release. /cc @timvandermeij

timvandermeij:
I agree, this would be a _really_ nice one to include.

brendandahl:
Still waiting on some Firefox talos runs for windows, but locally this patch is much slower loading the first page of unix01.pdf in the pdfpaint benchmark. 20 runs median 238.43ms vs 332.51ms

brendandahl:
Windows talos runs also show a slow down:
base: 257
new: 313
https://treeherder.mozilla.org/perfherder/compare?originalProject=try&originalRevision=2f45f4bcccb0fc60c48180aaaa432e5f1d96424d&newProject=try&newRevision=5cac140a3074ed804b1d72413384f2f065ed83d7&framework=1&showOnlyImportant=1

Snuffleupagus:
> Windows talos runs also show a slow down:
> base: 257
> new: 313
> https://treeherder.mozilla.org/perfherder/compare?originalProject=try&originalRevision=2f45f4bcccb0fc60c48180aaaa432e5f1d96424d&newProject=try&newRevision=5cac140a3074ed804b1d72413384f2f065ed83d7&framework=1&showOnlyImportant=1

I suppose that it's not entirely surprising if *some* slowdown is observed, since we're now having to actually *parse* font data that we previously didn't have to. The question is obviously how large of a regression we can accept!?

Looking briefly at the tested patch, it *appears* that it doesn't actually implement worker-thread caching of the standard font data (something that I mentioned as a potential problem above). If that's indeed the case, I'd not be entirely surprised if that at least in *part* contributes to the observed regression.

What would be very interesting to know, is more precisely *where* the regression comes from:
 - If it's *only* caused by `src/core/cff_parser.js` being used a lot more, then a regression would seem entirely acceptable since we obviously need to parse more font data with this patch.
 - If, on the other hand, a significant part of the regression is related to the additional data loading *itself* then I'd suggest that worker-thread caching might help. A related idea could perhaps be to pre-load the font data *early* during document loading, to cut down on the waiting time during the actual font-parsing!?

All-in-all, I suppose that it's going to take more work to get this landed than I'd initially hoped.

Snuffleupagus:
Having played around with this patch locally, focusing on the actual standard font data loading (using `console.time/timeEnd`), it seems that a *very significant* part of the regression comes from just waiting for the font data to become available in the worker.

Based on that observation I have one idea, however it's not a great one: What if we simply bundled the standard font data *into* the `pdf.worker.js` file, such that it's *synchronously* available?
Even suggesting this feel all kinds of wrong, but I'm still wondering it this would be an acceptable trade-off here?

It's difficult to not compare with the CMaps, but I do believe that this situation is slightly different here:
 - A *very large* number of existing PDF files use non-embedded standard fonts, whereas CMaps are much less rare in practice (e.g. mostly non-Latin documents). Hence it's going to be *very* common to have to access the standard font data.
 - The standard font data files included in this patch take up `266 616` bytes, whereas the CMap files are `1 167 756` bytes in total. Obviously the `pdf.worker.js` is already large, but that kind of size increase doesn't seem completely unacceptable in *this* case.
 - We've seen again and again how people report bugs, in custom PDF.js implementations, with broken font rendering where the culprit is that the CMap-parameters weren't provided when calling `getDocument`[1]; bundling the standard font data into `pdf.worker.js` would prevent similar issues there.
 - Given how many duplicate issues/bugs we've seen regarding missing standard fonts in PDF.js, it seems *very* unlikely that users would *not* want to use this feature anyway.

---
[1] Many user seem to just ignore the errors/warnings, mentioning the problem, printed in the console.

brendandahl:
I also tried fetching the font data from the worker. It was a bit faster but not much. 

I'd be curious to try directly embedding the files in the worker, but 266,616 bytes seems like a big increase for something that is not really needed in the common case.

> Looking briefly at the tested patch, it _appears_ that it doesn't actually implement worker-thread caching of the standard font data (something that I mentioned as a potential problem above). If that's indeed the case, I'd not be entirely surprised if that at least in _part_ contributes to the observed regression.

For the talos test, the web page is reloaded so the cache wouldn't help there. Also, I only see the font data being loaded once right now since we cache fonts, so I'm not sure this would help in the re-render case either.

I'm planning to put this patch on the back burner though as we have some higher priority stuff for firefox (XFA and tagged PDFs). I'll push up my changes with comments addressed, then we could add a pref to have it disabled by default for the time being.

Snuffleupagus:
> I'd be curious to try directly embedding the files in the worker, but 266,616 bytes seems like a big increase for something that is not really needed in the common case.

One more idea here: How about generating a separate bundle with the standard font data, and then using `importScripts` to load it into the worker-thread when needed?

Snuffleupagus:
I've tried implementing the second idea in https://github.com/mozilla/pdf.js/pull/12726#issuecomment-769260686, i.e. creating a bundle with all of the standard font data such that we only need to fetch *a single* file.
Despite the bundle being fairly large (~260 kb), this is a *significant* performance improvement for the `unix01.pdf` document (~33 percent faster than *this* PR). Obviously it's still not as fast as the `master` branch, but the remaining slowdown is unavoidable since there's more parsing happening in `src/core/fonts.js` and `src/core/cff_parser.js`.

Please see my WIP-patch at https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:standard-fonts

@brendandahl This is probably the best that we can do, without inlining the standard font data in the `pdf.worker.js` file; could we perhaps consider moving forward with our combined patches here?

Snuffleupagus:
> [...] as we have some higher priority stuff for firefox (XFA and tagged PDFs).

I *obviously* don't question the value of either of those features.
However, I'm willing to bet that including standard font data in PDF.js will fix *a lot* more things than the above features *combined* :-)

Snuffleupagus:
@brendandahl Ping, how about my idea/implementation mentioned in https://github.com/mozilla/pdf.js/pull/12726#issuecomment-770431639?

Given that it's going to be *impossible* to add this functionality without any sort of performance overhead, since we now need to parse more font data in affected cases, the only thing that we can realistically do here is to limit the necessary data fetching/loading by only having *one* (larger) font data bundle.

brendandahl:
> I've tried implementing the second idea in [#12726 (comment)](https://github.com/mozilla/pdf.js/pull/12726#issuecomment-769260686), i.e. creating a bundle with all of the standard font data such that we only need to fetch _a single_ file.
> Despite the bundle being fairly large (~260 kb), this is a _significant_ performance improvement for the `unix01.pdf` document (~33 percent faster than _this_ PR). Obviously it's still not as fast as the `master` branch, but the remaining slowdown is unavoidable since there's more parsing happening in `src/core/fonts.js` and `src/core/cff_parser.js`.
> 
> Please find my patch at [master...Snuffleupagus:standard-fonts](https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:standard-fonts)
> 
> @brendandahl This is probably the best that we can do, without inlining the standard font data in the `pdf.worker.js` file; could we perhaps consider moving forward with our combined patches here?

I imagine that will slow down the case where there's only one font used (especially for the generic version of the viewer)?

Another thought, we could try to translate the fonts so they don't need to go through the full font parser. I imagine this would be quite a bit of worker though. It'd probably be good to profile some more though, and see if the font translation is slowing it down or just loading the font in the browser is the slow part.

Snuffleupagus:
> I imagine that will slow down the case where there's only one font used (especially for the generic version of the viewer)?

At least locally, the difference in loading time is quite small between loading *one* font file and loading the (larger) bundle. It seems to me that the overhead related to simply opening the connection, fetching the data, and finally transferring it to the worker-thread is what dominates overall.
However, it would be quite simple to *by default* only use the bundle in `MOZCENTRAL` builds, but use the individual files for the `GENERIC` build. I purposely modified the relevant code such that both "file formats" would still be supported :-)

> Another thought, we could try to translate the fonts so they don't need to go through the full font parser. I imagine this would be quite a bit of worker though.

I've also had that idea, but thinking about the practical implications of that made me quickly abandon it. Besides the overall complexity, we'd also need to add a way to automatically re-build the translated fonts for any code-changes.

> [...] and see if the font translation is slowing it down or just loading the font in the browser is the slow part.

With the bundle approach, it's mostly the actual font conversion code that contributes to the "regression" here.

*However, I do believe that the comparison is perhaps not entirely fair here:*
In an ideal world every PDF document would simply embed *all* of its fonts, and we'd thus always have to run our font conversion code.
Hence it *may* be slightly more fair to compare the performance for a PDF document which has embedded *standard* fonts, but instead forcibly load "our" standard font data!?

Snuffleupagus:
Another thing that I *think* this patch would also improve is rendering of standard fonts on Linux, since we've a few bugs/issues reported where non-embedded standard fonts render badly on Linux because of poor font substitution.

Snuffleupagus:
> I'd be curious to try directly embedding the files in the worker, but 266,616 bytes seems like a big increase for something that is not really needed in the common case.

Well, with recent development I don't really think that argument holds any more :-)

The XFA functionality, which just landed, increased the size of the *built* `pdf.worker.js` file alone by `214 903` bytes in the MOZCENTRAL build. In my experience, XFA is quite *uncommon* whereas every other PDF file would benefit from us bundling standard fonts.
Hence I think that if we're willing to accept such a large size increase for a *niche* feature such as XFA, it really ought to be acceptable for non-embedded standard fonts given how just how common they are.

rtestard:
> The XFA functionality, which just landed, increased the size of the _built_ `pdf.worker.js` file alone by `214 903` bytes in the MOZCENTRAL build. In my experience, XFA is quite _uncommon_ whereas every other PDF file would benefit from us bundling standard fonts.
> Hence I think that if we're willing to accept such a large size increase for a _niche_ feature such as XFA, it really ought to be acceptable for non-embedded standard fonts given just how common they are.

XFA is a lot less frequent than Acroform but we see about a million unique monthly active users encountering them in our desktop telemetry. I know at least of the Canadian Gov website using them. This is a significant number of users who will benefit from this change.

Snuffleupagus:
> XFA is a lot less frequent than Acroform but we see about a million unique monthly active users encountering them in our desktop telemetry.

In all fairness, then we should *also* measure how often users encounter non-embedded standard fonts in PDF documents; I'd be extremely surprised if that number isn't *at least* one order of magnitude larger :-)

marco-c:
> I know at least of the Canadian Gov website using them

There are some XFA forms on the UK and USA gov website too.

Snuffleupagus:
> XFA is a lot less frequent than Acroform but we see about a million unique monthly active users encountering them in our desktop telemetry. I know at least of the Canadian Gov website using them. This is a significant number of users who will benefit from this change.

And my claim, based on *years* of contributing to the PDF.js library, is that non-embedded standard fonts are *very* common (much more-so than both AcroForm and XFA).

---

@brendandahl Can't we add telemetry, to measure how often non-embedded standard fonts are actually being used?

brendandahl:
> @brendandahl Can't we add telemetry, to measure how often non-embedded standard fonts are actually being used?

We can, but I think we're going to land this regardless of the telemetry. I just need to update it and switch to loading data from the worker. FWIW, we do currently collect some [font info](https://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2021-04-24&include_spill=0&keys=__none__!__none__!__none__&max_channel_version=nightly%252F90&measure=PDF_VIEWER_FONT_TYPES_2&min_channel_version=null&processType=*&product=Firefox&sanitize=1&sort_by_value=0&sort_keys=submissions&start_date=2021-04-19&table=0&trim=1&use_submission_date=0) on telemetry, but we don't capture if the font was embedded or not.

Snuffleupagus:
> I just need to update it and switch to loading data from the worker.

Without knowing *exactly* what you have in mind here, won't that cause trouble for the GENERIC library (e.g. in Node.js) though?
I.e. are you intending to *bundle* the font data into the `pdf.worker.js` file, or actually fetching the data from the worker?

As far as I can remember *originally* the CMap-data was loaded on the worker-thread, but it was re-factored to be loaded on the main-thread instead since that allowed users to provide a *custom* factory for loading the data in special situations; such as e.g. in Node.js or in order to deal with CSPs in browsers.

brendandahl:
I was planning to do a fetch in the worker thread. The PDF that I used for testing is kind of a bad example as it uses multiple standard fonts whereas it seems most only use one or two of them, so it's not worth preloading all the fonts in JS.

We could still support loading from the main thread, but have an option for loading from the worker if it's supported. It sucks that it will give us two paths for loading, but I don't think we should hold back the general use case (Firefox and self hosted) that should give better performance.

web/app_options.js
```diff
@@ -133,6 +133,14 @@ const defaultOptions = {
...
     kind: OptionKind.VIEWER + OptionKind.PREFERENCE,
   },
+  standardFontDataUrl: {
+    /** @type {string} */
+    value:
+      typeof PDFJSDev === "undefined" || !PDFJSDev.test("PRODUCTION")
+        ? "../external/standard_fonts/"
+        : "../web/standard_fonts/",
+    kind: OptionKind.API,
+  },
```
Snuffleupagus, line 143: Given that this uses `OptionKind.API`, and that the different kinds are grouped together, please move this option down to just after this line instead: https://github.com/mozilla/pdf.js/blob/239d0097fa5c717ba8dde56e175ae006c9053df1/web/app_options.js#L219

src/display/api.js
```diff
@@ -472,6 +479,8 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
       isEvalSupported: source.isEvalSupported,
       fontExtraProperties: source.fontExtraProperties,
       enableXfa: source.enableXfa,
+      standardFontDataUrl: source.standardFontDataUrl,
+      standardFontDataWorkerFetch: source.standardFontDataWorkerFetch,
```
Snuffleupagus, line 483: Given that the JSDocs for the `standardFontDataUrl` options says "Include the trailing slash." it thus cannot be empty and I think it should suffice to only send *one* parameter to the worker:
```javascript
standardFontDataUrl:
  source.standardFontDataWorkerFetch ? source.standardFontDataUrl : null,
```

src/core/worker.js
```diff
@@ -411,6 +411,8 @@ class WorkerMessageHandler {
         ignoreErrors: data.ignoreErrors,
         isEvalSupported: data.isEvalSupported,
         fontExtraProperties: data.fontExtraProperties,
+        standardFontDataUrl: data.standardFontDataUrl,
```
Snuffleupagus, line 414: When you're adding options here, please also remember to provide default values in https://github.com/mozilla/pdf.js/blob/239d0097fa5c717ba8dde56e175ae006c9053df1/src/core/evaluator.js#L91-L97

src/core/evaluator.js
```diff
@@ -381,10 +381,21 @@ class PartialEvaluator {
...
     const standardFontNameToFileName = getStdFontNameToFileMap();
+    const filename = standardFontNameToFileName[name];
+    if (this.options.standardFontDataWorkerFetch) {
+      const url = `${this.options.standardFontDataUrl}${filename}.pfb`;
+      const response = await fetch(url);
+      if (!response.ok) {
+        throw new Error(response.statusText);
+      }
+      const data = new Uint8Array(await response.arrayBuffer());
+      return new Stream(data);
```
Snuffleupagus, line 392: I believe that you'll need to wrap all of this is a `try...catch` block, and fallback to main-thread data loading if there's any errors, since otherwise there's a very real risk that font data loading will break in GENERIC-builds.

Given that, in the API, you're setting the option to *by default* (in browsers) try and load font data in the worker-thread, there could thus be errors if people have implemented a *custom* `StandardFontDataFactory` e.g. to deal with some CORS issue, or if the PDF.js library is used in an environment and/or with a protocol where `fetch` isn't properly supported.

src/display/api.js
```diff
@@ -2262,9 +2271,11 @@ class WorkerTransport {
       isCompressed: params.cMapPacked,
     });
 
-    this.StandardFontDataFactory = new params.StandardFontDataFactory({
-      baseUrl: params.standardFontDataUrl,
-    });
+    if (!params.standardFontDataWorkerFetch) {
```
Snuffleupagus, line 2274: `this.StandardFontDataFactory` needs to be defined *unconditionally*, please note the comment in the `src/core/evaluator.js` file.

Snuffleupagus (changes requested):
(As a follow-up, would it make sense to implement the same kind of feature for the CMap-data?)

src/display/api.js
```diff
@@ -455,6 +479,8 @@ function _fetchDocument(worker, source, pdfDataRangeTransport, docId) {
       isEvalSupported: source.isEvalSupported,
       fontExtraProperties: source.fontExtraProperties,
       enableXfa: source.enableXfa,
+      standardFontDataUrl: source.standardFontDataUrl,
+      standardFontDataWorkerFetch: source.standardFontDataWorkerFetch,
```
Snuffleupagus, line 483: As mentioned in https://github.com/mozilla/pdf.js/pull/12726/files#r634076823, please change this to the following instead to remove the need to send *two* very closely related properties:
```suggestion
      standardFontDataUrl:
        source.standardFontDataWorkerFetch ? source.standardFontDataUrl : null,
```

src/core/fonts_utils.js
```diff
@@ -78,9 +78,12 @@ const MacStandardGlyphOrdering = [
   "threequarters", "franc", "Gbreve", "gbreve", "Idotaccent", "Scedilla",
   "scedilla", "Cacute", "cacute", "Ccaron", "ccaron", "dcroat"];
 
-function getFontType(type, subtype) {
+function getFontType(type, subtype, isStandardFont) {
```
Snuffleupagus, line 81:
```suggestion
function getFontType(type, subtype, isStandardFont = false) {
```
since there's a couple of call-sites where this new third parameter isn't provided.

src/core/evaluator.js
```diff
@@ -377,6 +381,34 @@ class PartialEvaluator {
...
 
+  async fetchStandardFontData(name) {
+    const standardFontNameToFileName = getStdFontNameToFileMap();
+    const filename = standardFontNameToFileName[name];
+    if (this.options.standardFontDataWorkerFetch) {
+      const url = `${this.options.standardFontDataUrl}${filename}.pfb`;
+      const response = await fetch(url);
+      if (response.ok) {
+        const data = new Uint8Array(await response.arrayBuffer());
+        return new Stream(data);
```
Snuffleupagus, line 392: Please note that the `Stream` constructor will wrap an `ArrayBuffer` into a `Uint8Array`: https://github.com/mozilla/pdf.js/blob/3da9f077be33d8689129d5ddb7915d7bf1f2a5fa/src/core/stream.js#L23-L26
Hence this can be changed to simply:
```suggestion
        return new Stream(await response.arrayBuffer());
```

src/core/evaluator.js
```diff
@@ -377,6 +381,34 @@ class PartialEvaluator {
...
+      }
+      warn(
+        `fetchStandardFontData failed to fetch file "${url}" with "${response.statusText}".`
+      );
```
Snuffleupagus, line 396: You'll *also* need to wrap all of this code in a `try...catch` block to avoid breaking errors.
For example: `fetch` itself might be unsupported by the browser/environment, the protocol might not be supported by fetch, there might be CORS issues preventing file loading, and/or any other possible issues.

src/display/api.js
```diff
@@ -304,6 +325,9 @@ function getDocument(src) {
   if (!Number.isInteger(params.maxImageSize)) {
     params.maxImageSize = -1;
   }
+  if (typeof params.standardFontDataWorkerFetch !== "boolean") {
+    params.standardFontDataWorkerFetch = !isNodeJS;
```
Snuffleupagus, line 329: In the event that the user has provided a *custom* StandardFontDataFactory, it probably doesn't make sense to even attempt to load the font data in the worker-thread. Hence it might not be a bad idea to modify *this* default value as follows?
```suggestion
    params.standardFontDataWorkerFetch = 
      !isNodeJS &&
      params.StandardFontDataFactory instanceof DefaultStandardFontDataFactory;
```

test/unit/api_spec.js
```diff
@@ -1500,18 +1500,20 @@ describe("api", function () {
       const pdfPage = await pdfDoc.getPage(1);
       const { items, styles } = await pdfPage.getTextContent();
       expect(items.length).toEqual(1);
-      expect(Object.keys(styles)).toEqual(["Times"]);
+      // Font name will random object id.
```
Snuffleupagus, line 1503:
```suggestion
      // Font name will be a random object id.
```

test/unit/annotation_spec.js
```diff
@@ -68,13 +76,20 @@ describe("annotation", function () {
     }
   }
 
+  const FontDataReader = new DefaultStandardFontDataFactory({
```
Snuffleupagus, line 79: Nit: Since this is an *instance*, let's use a lower-case initial letter here.
```suggestion
  const fontDataReader = new DefaultStandardFontDataFactory({
```

src/core/evaluator.js
```diff
@@ -3709,6 +3742,12 @@ class PartialEvaluator {
           toUnicode,
         };
         const widths = dict.get("Widths");
+        const standardFontName = getStandardFontName(baseFontName);
+        let file = null;
+        if (standardFontName) {
+          file = await this.fetchStandardFontData(standardFontName);
+          properties.isStandardFont = true;
```
Snuffleupagus, line 3749: If loading of the standard font file failed, and thus `file === null`, it probably doesn't make sense to set the `isStandardFont` property since that affects glyph mapping!?

```suggestion
          properties.isStandardFont = !!file;
```

src/core/evaluator.js
```diff
@@ -3787,6 +3827,12 @@ class PartialEvaluator {
         length2 = fontFile.dict.get("Length2");
         length3 = fontFile.dict.get("Length3");
       }
+    } else if (type === "Type1") {
+      const standardFontName = getStandardFontName(fontName.name);
+      if (standardFontName) {
+        fontFile = await this.fetchStandardFontData(standardFontName);
+        isStandardFont = true;
```
Snuffleupagus, line 3834: As above:
```suggestion
        isStandardFont = !!fontFile;
```

Snuffleupagus (changes requested):

test/unit/test_utils.js
```diff
@@ -26,6 +26,10 @@ const CMAP_PARAMS = {
   cMapPacked: true,
 };
 
+const STANDARD_FONT_DATA_URL = isNodeJS
+  ? "./external/standard_fonts/"
+  : "/build/generic/web/standard_fonts/";
```
Snuffleupagus, line 31: Shouldn't this be `"../../external/standard_fonts/"` instead here, such that the unit-tests also work when run using e.g. http://localhost:8888/test/unit/unit_test.html?

test/unit/test_utils.js
```diff
@@ -61,6 +65,8 @@ function buildGetDocumentParams(filename, options) {
   params.url = isNodeJS
     ? TEST_PDFS_PATH + filename
     : new URL(TEST_PDFS_PATH + filename, window.location).href;
+  params.standardFontDataUrl = STANDARD_FONT_DATA_URL;
+  params.standardFontDataWorkerFetch = !isNodeJS;
```
Snuffleupagus, line 69: This line shouldn't be necessary, given the default value set in `src/display/api.js`, as far as I can tell.

Snuffleupagus (changes requested):

test/unit/test_utils.js
brendandahl: That relative path won't work for the packaged version of the unitests since it loads relative to the worker js file. 
I've been meaning to look into this, but the unpackaged unit tests in the browser are kind of broken arleady and are loading the packaged version of the worker. With this path above they both still work.

src/core/evaluator.js
brendandahl: I've been going back and forth on this, but I'm leaning towards not automatically falling back to getting data on the main thread if fetch fails as it really is an error that the user of the library should handle either by disabling fetch or putting the font files in the correct location. Also, if we fallback automatically this may silently regress and fetch in the worker has been supported for quite awhile now.

src/core/evaluator.js
```diff
@@ -377,6 +380,34 @@ class PartialEvaluator {
...
+    if (this.options.standardFontDataUrl !== null) {
+      const url = `${this.options.standardFontDataUrl}${filename}.pfb`;
+      const response = await fetch(url);
+      if (!response.ok) {
+        warn(
+          `fetchStandardFontData failed to fetch file "${url}" with "${response.statusText}".`
+        );
+        return null;
+      }
+      return new Stream(await response.arrayBuffer());
```
Snuffleupagus, line 395: Sorry, but I still think that you need to properly catch any errors and always fallback to a main-thread fetch here to prevent any future breakage in the GENERIC viewer; something along these lines:
```suggestion
      try {
        const response = await fetch(url);
        if (!response.ok) {
          throw new Error(response.statusText);
        }
        return new Stream(await response.arrayBuffer());
      } catch (e) {
        warn(
          `fetchStandardFontData failed to fetch url "${url}" with "${e}".`
        );
      }
```

src/display/api.js
```diff
@@ -143,6 +152,16 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *   reading built-in CMap files. Providing a custom factory is useful for
  *   environments without Fetch API or `XMLHttpRequest` support, such as
  *   Node.js. The default value is {DOMCMapReaderFactory}.
+ * @property {string} [standardFontDataUrl] - The URL where the standard font
+ *   files are located. Include the trailing slash.
+ * @property {boolean} [standardFontDataWorkerFetch] - Enable fetching the font
+ *   data from the worker thread. When `true`, StandardFontDataFactory will be
+ *   ignored. The default value is `true` in web environment and `false` for
```
Snuffleupagus, line 159: Let's remove the ```When `true`, StandardFontDataFactory will be ignored.``` sentence here, since we should always fallback to a main-thread fetch if all else fails.

test/unit/annotation_spec.js
```diff
@@ -107,6 +125,9 @@ describe("annotation", function () {
       idFactory: createIdFactory(/* pageIndex = */ 0),
       fontCache: new RefSetCache(),
       builtInCMapCache,
+      options: {
+        standardFontDataUrl: null,
+      },
```
Snuffleupagus, line 130: Given that `standardFontDataUrl: null,` is the default value, I don't think you should need this block now?

test/unit/annotation_spec.js
```diff
@@ -2281,7 +2302,10 @@ describe("annotation", function () {
         { ref: buttonWidgetRef, data: buttonWidgetDict },
       ]);
       const task = new WorkerTask("test print");
-      partialEvaluator.options = { ignoreErrors: true };
+      partialEvaluator.options = {
+        ignoreErrors: true,
+        standardFontDataUrl: null,
```
Snuffleupagus, line 2307: Since `standardFontDataUrl: null,` is the default value, is this line necessary?

Snuffleupagus (approved):
r=me, with the last few comments addressed; thank you for fixing this long-standing issue!

src/core/evaluator.js
brendandahl: Did you see my [comment above](https://github.com/mozilla/pdf.js/pull/12726#discussion_r641715489) (may have been lost in all the updates). If you did, and think we should still fallback, we could alternatively only fallback for the generic viewer.

Snuffleupagus: Sorry, I completely missed that!

Well, you obviously do have a point about both `fetch`-support and silent failures, and users can of course disable this easily.
Let's just ignore my comment for now, and we can always revisit this (e.g. for `GENERIC`-builds only) if it turns out to cause problems in practice :-)

Snuffleupagus (changes requested):
Really sorry, but after going over all of the reference-test changes in some detail I've now spotted a couple of regressions that I think we need to fix before landing this:
 - issue11385
 - issue6019-text
 - operator-in-TJ-array

Snuffleupagus:
While I still don't know how to fix rendering regression in `issue11385`, the textLayer regressions mentioned in https://github.com/mozilla/pdf.js/pull/12726#pullrequestreview-673980193 should however be easy to fixed with the following change:
```diff
diff --git a/src/core/fonts.js b/src/core/fonts.js
index 1eccf7b5b..618b6f472 100644
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -131,6 +131,9 @@ function adjustWidths(properties) {
 }

 function adjustToUnicode(properties, builtInEncoding) {
+  if (properties.isStandardFont) {
+    return;
+  }
   if (properties.hasIncludedToUnicodeMap) {
     return; // The font dictionary has a `ToUnicode` entry.
   }

```

brendandahl:
Thanks for catching those. I'm also not sure about `issue11385` the glyph `č` is not in the standard font files. If I'm following the pdfium code correctly, they don't actually use those standard font files if the system has the fonts available.

brendandahl:
Calxite proposed another option for fixing issue11385 would be to use fonts that define more glyphs such the liberation fonts (we'll need them for XFA anyway). This does fix the issue, but the bigger fonts even further slow things down:

first page of example_055.pdf from tcpdf:
```
214-250ms - liberation
171-180ms - foxit
110-120ms - system
```

Another option I proposed: if we see that a glyph is missing from the standard font then load a bigger standard font with more glyphs or fallback to the system font. Both of those seem father complicated though.

Since this work is blocking Calixte's work on XFA, I propose we land with the following and iterate (which is similar to what @Snuffleupagus  was [thinking above](https://github.com/mozilla/pdf.js/pull/12726#pullrequestreview-550198199)):
 1) only use foxit fonts for symbol fonts
 2) use system fonts for times, helvetica, and courier

In follow ups:
1) use foxit fonts when disableFontFace is true
2) use foxit fonts when we need the path for the font

Snuffleupagus:
> Calxite proposed another option for fixing issue11385 would be to use fonts that define more glyphs such the liberation fonts (we'll need them for XFA anyway). This does fix the issue, but the bigger fonts even further slow things down:

A quick look at https://github.com/liberationfonts/liberation-fonts seems to suggest that the liberation fonts are provided as TrueType fonts, which if I recall correctly are generally speaking larger than a similar CFF file (which is what the foxit fonts are).

My, probably stupid, question would then be if it's possible to convert the liberation fonts to CFF and get files which are smaller and should thus (hopefully) be faster to both fetch and parse?

> This does fix the issue, but the bigger fonts even further slow things down:
>     1. only use foxit fonts for symbol fonts
>     2. use system fonts for times, helvetica, and courier
> 
> In follow ups:
> 
>     1. use foxit fonts when disableFontFace is true
>     2. use foxit fonts when we need the path for the font

I'm slightly confused here, where does the need for liberation fonts (for the XFA use-case) fit into the plan above?
Should we not instead load the liberation fonts *conditionally* rather than the foxit ones, since I'm assuming that we don't want to include basically "duplicate" fonts for different use-cases?

src/core/evaluator.js
```diff
@@ -381,6 +385,43 @@ class PartialEvaluator {
...
+  async fetchStandardFontData(name) {
+    // The symbol fonts are not consistent across platforms, always load the
+    // font data for them.
+    if (
+      this.options.useSystemFonts &&
+      name !== "Symbol" &&
+      name !== "ZapfDingbats"
+    ) {
+      return null;
+    }
```
Snuffleupagus, line 397: > In follow ups:
> use foxit fonts when disableFontFace is true

I believe that you could quite easily support at least *this* part immediately, simply by doing:
```suggestion
    if (!this.options.disableFontFace) {
      // The symbol fonts are not consistent across platforms, always load the
      // font data for them.
      if (
        this.options.useSystemFonts &&
        name !== "Symbol" &&
        name !== "ZapfDingbats"
      ) {
        return null;
      }
    }
```

brendandahl:
I did a quick test converting one of the liberation fonts to cff using `tx`. It did bring the size down:
```
248K	out.cff
404K	LiberationSans-Regular.ttf
```
Still much bigger than the foxit font of course and I think we loose hinting when we convert (though we may be stripping that out anyway in fonts.js).

As for how this interacts with liberation and XFA, up above I should have wrote "internal" instead of "foxit". I think we should replace the foxit courier, times, helvetica with the liberation fonts when we land the XFA changes. I'd like to first land with the foxit fonts, so we can then run tests to make sure the glyphs still match up once we switch.

We could have the option to use foxit or liberation, but that seems like a pain to support yet another configuration.

src/display/api.js
```diff
@@ -143,6 +152,19 @@ function setPDFNetworkStreamFactory(pdfNetworkStreamFactory) {
  *   reading built-in CMap files. Providing a custom factory is useful for
  *   environments without Fetch API or `XMLHttpRequest` support, such as
  *   Node.js. The default value is {DOMCMapReaderFactory}.
+ * @property {boolean} [useSystemFonts] - When true, fonts that aren't embedded
+ *   in the PDF will fallback to a system font. Defaults to true for web
+ *   environments and false for node.
+ * @property {string} [standardFontDataUrl] - The URL where the standard font
+ *   files are located. Include the trailing slash.
+ * @property {boolean} [standardFontDataWorkerFetch] - Enable fetching the font
```
brendandahl, line 160: @Snuffleupagus You had mentioned using worker fetch for cmaps too. What are you thoughts on changing this to `useWorkerFetch`, so we just have one option to enable it?

Snuffleupagus:
> Still much bigger than the foxit font of course and I think we loose hinting when we convert (though we may be stripping that out anyway in fonts.js).

Looking quickly at the code in https://github.com/mozilla/pdf.js/blob/master/src/core/cff_parser.js, there seem to be *some* (possibly limited) handling of hint data as far as I can tell!?
*Ninja-edit:* Although, I suppose that you mean that the TrueType -> CFF conversion itself strips the hinting data?

> I think we should replace the foxit courier, times, helvetica with the liberation fonts when we land the XFA changes.

So, just to make sure that I understand your plan here:
 1. Land an initial implementation, using the foxit fonts.
 2. Later change to use liberation fonts, as part of the XFA-work, regardless of the resulting font file sizes.

> We could have the option to use foxit or liberation, but that seems like a pain to support yet another configuration.

I completely agree, supporting multiple types seems really annoying and error-prone!

src/display/api.js
Snuffleupagus: > What are you thoughts on changing this to `useWorkerFetch`, so we just have one option to enable it?

That's a really *excellent* idea, the fewer new options we can add the better in my opinion (for general testability and maintainability reasons).

brendandahl:
> _Ninja-edit:_ Although, I suppose that you mean that the TrueType -> CFF conversion itself strips the hinting data?

Yeah, I thought I remember reading the hinting was lost, but now I don't see anything in the tx docs.
 
> So, just to make sure that I understand your plan here:
> 
>     1. Land an initial implementation, using the foxit fonts.
> 
>     2. Later change to use liberation fonts, as part of the XFA-work, regardless of the resulting font file sizes.
> 

That's the general gist. We could do some work to reduce the font size and use cff, but calixte's patch to rescale glyphs only currently works with true type fonts. We could also probably speed it up by no running the glyphs to the pdf.js santizer.

Snuffleupagus:
@brendandahl Can you please squash the commits (since there's now a little bit of back-and-forth in there), and re-run all tests before we try landing this?

src/core/cff_font.js
```diff
@@ -74,7 +74,10 @@ class CFFFont {
       return charCodeToGlyphId;
     }
 
-    const encoding = cff.encoding ? cff.encoding.encoding : null;
+    let encoding = cff.encoding ? cff.encoding.encoding : null;
+    if (properties.isStandardFont) {
```
Snuffleupagus, line 78: Wait, shouldn't this actually be `properties.isInternalFont` now given the change to this parameter?
This might be related to the movement in the reference tests!?

src/core/fonts_utils.js
```diff
@@ -135,7 +138,17 @@ function type1FontGlyphMapping(properties, builtInEncoding, glyphNames) {
   let glyphId, charCode, baseEncoding;
   const isSymbolicFont = !!(properties.flags & FontFlags.Symbolic);
 
-  if (properties.baseEncodingName) {
+  if (properties.isStandardFont) {
```
Snuffleupagus, line 141: Wait, shouldn't this actually be `properties.isInternalFont` now given the change to this parameter?
This might be related to the movement in the reference tests!?

Snuffleupagus (changes requested):

src/core/cff_font.js
brendandahl: I don't think this should matter, since standard fonts should use the base encoding, but we can change it for now. 

As for the ref test changes, I too was confused by that. It turns out it was because I added several of the standard fonts to `getStdFontMap` that were not in there before. This caused `isStandardFont` to be `true` in https://github.com/mozilla/pdf.js/blob/8b4acb4e3683e544f3b4840b3b9a8bfa70f36102/src/core/fonts.js#L967 and subsequently changed the value of `remeasure`. It doesn't seem to change much on the bots, but on macos the fonts seem to look better now for unix01.

Snuffleupagus (approved):

kmcclellan:
Could we get a new build of `pdfjs-dist` to npm please?

Snuffleupagus:
> Could we get a new build of `pdfjs-dist` to npm please?

Not immediately, since this *just* landed and we (obviously) need some time to handle any possible regressions and there's also various follow-up work that should happen first. (Also, the last release was less than two weeks ago.)

Snipeye:
I don't github (verb) very often, so I don't know if this is correct/appropriate (or if I'm about to ping a million people, in which case I am deeply sorry) but I just read through this and wanted to add my "thank you 1e6" for the effort to get this in; at least in my case it dramatically increases the usability.

ceztko:
Sorry for late comment. I just wanted to say that the fonts taken from PDFium are possibly misnamed, or their format is not immediately clear. `pfb` extension seems to be mostly taken for [Printer Font Binary](https://en.wikipedia.org/wiki/PostScript_fonts#Printer_Font_Binary), which is basically the encrypted part of a `pfa` font. `pfb` can't be used alone but must be coupled with a `pfm` file that just contains the metric. The fonts taken from PDFium are neither  canonical `pfb` or `pfa`, but they seem to be a raw OpenType compatible [CFF](https://en.wikipedia.org/wiki/PostScript_fonts#Compact_Font_Format) representation, and I can confirm they are opened by the CFF driver in Freetype. I don't know if there's a better extension for these. At least their actual format could be better documented.

PR #12728, opened 2020-12-11
[gulpfile.js] Move the time-zone hack to the `testing-pre` task, such that *all* tests work regardless of the current time-zone

Snuffleupagus:
Currently only the `gulp unittest` task actually set the time-zone, which means that locally I'm now getting failures with e.g. `gulp test`.

*Please note:* I firmly believe that the unit-tests in question should be re-written, since even with this patch applied there's failures when running http://localhost:8888/test/unit/unit_test.html directly in a browser.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1521,6 +1521,10 @@ gulp.task(
 
 gulp.task("testing-pre", function (done) {
   process.env.TESTING = "true";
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12732, opened 2020-12-12
Call `done.fail` correctly in the `scripting_spec.js` unit-tests

Snuffleupagus:
The `done.fail` method should *always* be called with a reason, to ensure that any errors are propagated as intended to the test results.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12736, opened 2020-12-14
Remove the remaining IE 11 polyfills

Snuffleupagus:
We really ought to settle on the *lowest* supported versions of various browsers[1], since that should allow even more clean-up, but given that https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support *explicitly* lists IE 11 as unsupported after PDF.js version `2.6.347` there's a number of polyfills that are no longer needed.

Based on the MDN compatibility data, this patch thus removes the following polyfills:
 - `String.prototype.startsWith()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith#Browser_compatibility

 - `String.prototype.endsWith()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith#Browser_compatibility

  - `String.prototype.includes()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Browser_compatibility

 - `Array.prototype.includes()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility

  - `Array.from()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Browser_compatibility

  - `Object.assign()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility

  - `Math.log2()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/log2#Browser_compatibility

  - `Number.isNaN()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN#Browser_compatibility

  - `Number.isInteger()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#Browser_compatibility

  - `Map.prototype.entries()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries#Browser_compatibility

  - `Set.prototype.entries()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/entries#Browser_compatibility

  - `WeakMap`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#Browser_compatibility

  - `WeakSet`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet#Browser_compatibility

  - `Symbol`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Browser_compatibility

Finally, this patch also attempts to update the compatibility information for the remaining polyfills.

---
[1] For example: It's questionable if Safari 9 should be listed as supported, given that the last release from that branch was in 2016.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -84,7 +84,6 @@ var AUTOPREFIXER_CONFIG = {
     "Chrome >= 49", // Last supported on Windows XP
     "Firefox >= 52", // Last supported on Windows XP
     "Firefox ESR",
-    "IE >= 11",
     "Safari >= 9",
     "> 0.5%",
     "not dead",
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -21,7 +21,7 @@ if (
   (typeof globalThis === "undefined" || !globalThis._pdfjsCompatibilityChecked)
 ) {
   // Provides support for globalThis in legacy browsers.
-  // Support: IE11/Edge, Opera
+  // Support: Firefox<65, Chrome<71, Safari<12.1
   if (typeof globalThis === "undefined" || globalThis.Math !== Math) {
     // eslint-disable-next-line no-global-assign
     globalThis = require("core-js/es/global-this");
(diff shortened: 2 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me, and given that we already announced IE11 as being unsupported this change seems fine. Thanks!

PR #12737, opened 2020-12-14
Change the minimum "supported" version of the Safari-browser to Safari 10

Snuffleupagus:
According to https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support, Safari 9 is still listed as "mostly supported".

Given that the *last* release from the Safari 9 branch was on [September 1, 2016](https://en.wikipedia.org/wiki/Safari_version_history#Safari_9), it's questionable at least to me if it actually makes sense for us to even pretend to "support" such an old browser.
Especially when the *first* release from the Safari 10 branch was on [September 20, 2016](https://en.wikipedia.org/wiki/Safari_version_history#Safari_10), which is now over four years ago.

Based on the MDN compatibility data, this patch thus removes the following polyfills:
 - `TypedArray.prototype.slice()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice#Browser_compatibility
 - `String.prototype.codePointAt()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt#Browser_compatibility
 - `String.fromCodePoint()`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint#Browser_compatibility

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -84,7 +84,7 @@ var AUTOPREFIXER_CONFIG = {
     "Chrome >= 49", // Last supported on Windows XP
     "Firefox >= 52", // Last supported on Windows XP
     "Firefox ESR",
-    "Safari >= 9",
+    "Safari >= 10",
     "> 0.5%",
     "not dead",
   ],
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -59,15 +59,6 @@ if (
     require("core-js/es/object/from-entries.js");
   })();
 
-  // Provides support for TypedArray.prototype.slice in legacy browsers.
-  // Support: Safari<10
-  (function checkTypedArraySlice() {
-    if (Uint8Array.prototype.slice) {
-      return;
-    }
-    require("core-js/es/typed-array/slice");
-  })();
-
   // Provides support for *recent* additions to the Promise specification,
   // however basic Promise support is assumed to be available natively.
   // Support: Firefox<71, Chrome<76, Safari<13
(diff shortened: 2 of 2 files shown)


timvandermeij:
I fully agree with this, and anything higher than Safari 9 (which is way too old in my opinion as well) seems reasonable to me. Let's indeed stick with `>= 10` for now, and we can always increase the number later on if the need arises, but this seems like a small change that will give us some clean-up.

Snuffleupagus:
Something else, which I just spotted the other day, is that the update to Webpack 5 changed some *default values*. Note how there's now an *Arrow function* being used as part of the Webpack-header even in our `es5`-files, see e.g. https://github.com/mozilla/pdf.js/blob/917fbf0f458898bc7b3c09b88da518ce154a08b3/es5/build/pdf.js#L33

This can obviously be fixed by tweaking some configuration options, however looking at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Browser_compatibility I'm not convinced that we need to do that *if* we accept this patch since all browsers that we support should thus have native *Arrow function* support.

timvandermeij (approved):

timvandermeij:
Looks good to me. I'll update the FAQ page.

Snuffleupagus:
> I'll update the FAQ page.

Thanks for doing this!

Looking more closely on that list, I do wonder if we actually need to keep the "Safari 9 and below"-line since it honestly seems redundant now; I'd suggest just removing it.
Also, the "Android 4 and below"-line seems useless as well given that e.g. Android 5 is *way too old* to be supported; hence I'd suggest just removing this one too.

timvandermeij:
I agree; it's removed now.

PR #12741, opened 2020-12-16
JS -- Actions must be evaluated in global scope

calixteman:
* All the public properties of doc are injected into globalThis, in order to make them available through `this`
 * Put event in the global scope too.

### DIFF
--- a/src/scripting_api/initialization.js
+++ b/src/scripting_api/initialization.js
@@ -72,6 +72,7 @@ function initSandbox(params) {
     obj.send = send;
     obj.globalEval = globalEval;
     obj.doc = _document.wrapped;
+    obj.globalEval = globalEval;
     const field = new Field(obj);
     const wrapped = new Proxy(field, proxyHandler);
     doc._addField(name, wrapped);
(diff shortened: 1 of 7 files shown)


src/scripting_api/initialization.js
```diff
@@ -68,34 +71,57 @@ function initSandbox({ data, extra, out }) {
...
+  globalThis.zoomtype = ZoomType;
 
   for (const name of Object.getOwnPropertyNames(AForm.prototype)) {
     if (name !== "constructor" && !name.startsWith("_")) {
-      out[name] = aform[name].bind(aform);
+      globalThis[name] = aform[name].bind(aform);
+    }
+  }
+
+  // The doc properties must live in the global scope too
```
brendandahl, line 106: I'm guessing it's not possible to make the global scope be the document?

brendandahl: For example when I do `console.println(this)` in acrobat, i get `[object Doc]`

calixteman: I think that in a way this code change globalThis into a kind of doc.
About the second point, we can change the implementation of `toString`.
In chrome, it displays `[object global]`.

brendandahl (approved):

PR #12747, opened 2020-12-17
Add new "pageopen"/"pageclose" events for usage with JavaScript actions 

Snuffleupagus:
*Please refer to the individual commit messages for additional information.*

https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/AcrobatDC_js_api_reference.pdf#G5.1963437

@timvandermeij I really hoped to be able to use e.g. the existing "pagechanging" events, but based on the requirements outlined in the specification I just don't see a better way than adding new events here; please let me know if you have a better idea!

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -460,6 +460,8 @@ class BaseViewer {
    */
   setDocument(pdfDocument) {
     if (this.pdfDocument) {
+      this.eventBus.dispatch("pagesdestroy", { source: this });
+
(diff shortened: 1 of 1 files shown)


calixteman (approved):
WFM, thanks a lot.
I'll add an integration test in a followup.

timvandermeij (approved):

timvandermeij:
Looks like a good solution to me too!

PR #12748, opened 2020-12-17
Update the events, used with scripting, to use lower-case names and avoid using DOM events internally in the viewer + misc scripting-related tweaks

Snuffleupagus:
*Please refer to the individual commit messages for additional information.*

Also, ignoring white-space changes when viewing the larger patches should be helpful.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -478,14 +478,13 @@ class LinkAnnotationElement extends AnnotationElement {
         continue;
       }
       link[jsName] = () => {
-        window.dispatchEvent(
-          new CustomEvent("dispatchEventInSandbox", {
-            detail: {
-              id: data.id,
-              name,
-            },
-          })
-        );
+        this.linkService.eventBus?.dispatch("dispatcheventinsandbox", {
+          source: this,
+          detail: {
+            id: data.id,
+            name,
+          },
+        });
         return false;
       };
     }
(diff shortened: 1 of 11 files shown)


src/display/annotation_layer.js
```diff
@@ -544,40 +544,42 @@ class WidgetAnnotationElement extends AnnotationElement {
...
-              },
-            })
-          );
+    if (this.data.actions[eventName.replace(" ", "")] === undefined) {
+      return;
+    }
+    if (baseName.includes("mouse")) {
+      // Mouse events
+      element.addEventListener(baseName, event => {
+        this.linkService.eventBus?.dispatch("dispatcheventinsandbox", {
```
calixteman, line 553: I think we should have at least a warning in the console if eventBus is null.
Or if we expect that eventBus is neither null nor undefined, then you can remove the `?` and we'll get an exception.

src/display/annotation_layer.js
```diff
@@ -717,39 +718,37 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
...
+                commitKey,
+                selStart: event.target.selectionStart,
+                selEnd: event.target.selectionEnd,
+              },
+            });
+          });
+          const _blurListener = blurListener;
+          blurListener = null;
+          element.addEventListener("blur", event => {
+            if (this._mouseState?.isDown) {
```
calixteman, line 737: `_mouseState` shoudn't be null here else the behaviour won't be correct, so you should report something (a warning or throw an exception).

src/display/annotation_layer.js
Snuffleupagus: That's not really a problem though, since the `AnnotationLayer` *must* be initialized with either a `PDFLinkService` or a `SimpleLinkService` instance (the latter having only dummy-methods).
Note how the existing code simply assumes that *either* is available, and the optional chaining added here (and elsewhere) is simply intended to prevent errors in the `SimpleLinkService`-case.

src/display/annotation_layer.js
Snuffleupagus: We're not currently warning/throwing for *any* of the parameters used with the `AnnotationLayer`, which assuming we even care enough to fix this (which is somewhat doubtful to me), should then probably be fixed all at once rather than for just one parameter.

Please note also that it's perfectly fine for the `mouseState` parameter to be optional in *some* cases:
 - The `AnnotationLayer` is also utilized with the viewer-components, which can be used build a simplified viewer without all the bells and whistles of the full default viewer, and in that case we *really* shouldn't require all of the scripting-related options to be provided by the user.
 - Furthermore, the *existing* viewer JSDocs specifically listed the `mouseState`-parameter as optional in e.g. the `BaseViewer` and consequently it should be optional *everywhere* including the `AnnotationLayer`.

However, I've changed things slightly to ensure that `this._mouseState` is always defined with the expected format, such that we don't need to use inline checks in (potentially) multiple places.

calixteman:
The failure in integration test is because the open action is executed after trying to get the focused element.
I'm figuring out a way to fix it.

calixteman:
@Snuffleupagus I think this patch fix the issue: https://github.com/mozilla/pdf.js/pull/12755

timvandermeij (approved):

timvandermeij:
Thank you for improving the overall consistency here!

Issue #12750, opened 2020-12-18
Invalid DA avoids to save or print


calixteman:
Attach (recommended) or Link to PDF file here:
[da_not_a_string.pdf](https://github.com/mozilla/pdf.js/files/5715326/da_not_a_string.pdf)

The problem is that the DA for annotation 12 is a Name instead being a string and it prevents from saving or printing because an exception is thrown when trying to get font information from the DA.

PR #12751, opened 2020-12-18
Add a default DA for textfield to avoid issues when printing or saving

calixteman:
* it aims to fix issue #12750


src/core/annotation.js
```diff
@@ -1282,6 +1285,15 @@ class WidgetAnnotation extends Annotation {
     const totalHeight = this.data.rect[3] - this.data.rect[1];
     const totalWidth = this.data.rect[2] - this.data.rect[0];
 
+    if (!this.data.defaultAppearance) {
+      // The DA is required and must be a string.
+      // There is no reason to have a font named Helvetica in
```
timvandermeij, line 1290: Why not? The PDF file might just have a font named `Helvetica` in the dictionary. I think this whole sentence should be rephrased to: `If there is no font named Helvetica in the resource dictionary, the evaluator will fall back to a default font. Doing so prevents exceptions and allows saving/printing the file as expected.`

src/core/annotation.js
```diff
@@ -1282,6 +1285,15 @@ class WidgetAnnotation extends Annotation {
...
+      // the resource dictionary so the evaluator will fallback on
+      // a default one: it isn't ideal but at least it prevents from
+      // exceptions and it allows to save or print the file as expected.
+      this.data.defaultAppearance = "/Helvetica 10 Tf 0 g";
```
timvandermeij, line 1294: Looking at https://stackoverflow.com/questions/17503349/da-pdfbox-how-to-create, would using 0 perhaps make sense here so it's auto-scaled?

PR #12754, opened 2020-12-18
Dispatch an event on sandbox creation

calixteman:
* the goal is to be able to know when the sandbox is ready for mochitest in m-c


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12755, opened 2020-12-18
JS -- in integration test, wait for open actions

calixteman:
Aims to fix intermittent failures in integration tests

### DIFF
--- a/test/integration/scripting_spec.js
+++ b/test/integration/scripting_spec.js
@@ -30,9 +30,12 @@ describe("Interaction", () => {
     it("must check that first text field has focus", async () => {
       await Promise.all(
         pages.map(async ([browserName, page]) => {
+          await page.waitForFunction(
+            "window.PDFViewerApplication.opendocfinished === true"
+          );
+
           // The document has an open action in order to give
           // the focus to 401R.
-          await page.waitForTimeout(1000);
           const id = await page.evaluate(
             () => window.document.activeElement.id
           );
--- a/web/app.js
+++ b/web/app.js
@@ -1525,6 +1525,11 @@ const PDFViewerApplication = {
               this.pdfViewer.currentScale = value;
             }
             break;
+          case "opendocfinished":
+            if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("TESTING")) {
+              this.opendocfinished = true;
+            }
+            break;
         }
         return;
       }
--- a/src/scripting_api/doc.js
+++ b/src/scripting_api/doc.js
@@ -109,6 +109,7 @@ class Doc extends PDFObject {
         }
       }
       this._runActions("OpenAction");
+      this._send({ command: "opendocfinished" });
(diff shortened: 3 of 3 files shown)


Snuffleupagus (changes requested):
Is there really not an *existing* event you could wait for instead, rather than hacking the viewer in this way!?

I wouldn't at all mind if the following method [were added here](https://github.com/mozilla/pdf.js/blob/e6e280982536d27582b55ee41b2c200f29e79ac5/web/genericcom.js#L57) to give you access to all the internal viewer-events in the DOM during tests:
```javascript
static get isInAutomation() {
  return shadow(this, "isInAutomation", typeof PDFJSDev !== "undefined" && PDFJSDev.test("TESTING"));
}
```

calixteman:
I would prefer too but I didn't find any way to achieve this using puppeteer (https://github.com/puppeteer/puppeteer/blob/main/docs/api.md). But of course feel free to find a better solution.

calixteman:
We could a solution based on `page.exposeFunction` but it doesn't work in firefox:
https://github.com/puppeteer/puppeteer/issues/6116

calixteman:
Anyway we can merge it in waiting for a better solution, I think it's better than adding a timeout.

web/app.js
```diff
@@ -1525,6 +1525,11 @@ const PDFViewerApplication = {
               this.pdfViewer.currentScale = value;
             }
             break;
+          case "opendocfinished":
+            if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("TESTING")) {
+              this.opendocfinished = true;
```
timvandermeij, line 1530: Name this `this._openDocFinished` to use same casing as the other member variables and to indicate that it's private, especially since this is only used for testing.

Moreover, since I agree with @Snuffleupagus that this is a bit hacky (although more solid than fixed timeout that may or may not be enough), I'd at least like to see a short comment here explaining why this is there; something like:

`// Signal the integration tests that open actions have run.`

test/integration/scripting_spec.js
```diff
@@ -30,9 +30,12 @@ describe("Interaction", () => {
     it("must check that first text field has focus", async () => {
       await Promise.all(
         pages.map(async ([browserName, page]) => {
+          await page.waitForFunction(
+            "window.PDFViewerApplication.opendocfinished === true"
```
timvandermeij, line 34: I think we don't need the `=== true` check here since if it's not defined `window.PDFViewerApplication.opendocfinished` should be undefined (falsy).

Snuffleupagus:
Hopefully this is no longer necessary, since the integration-tests are now passing consistently with the last patch from PR #12748 :-)

brendandahl:
@Snuffleupagus are we good to close this?

Snuffleupagus:
> @Snuffleupagus are we good to close this?

I'd say so, since this no longer fails with recent changes.

PR #12759, opened 2020-12-19
Annotation -- Don't compute appearance when nothing has changed

calixteman:
* don't set a value in annotationStorage by default:
   - having an undefined when the annotation is rendered for saving/printing means nothing has changed so use normal appearance
   - aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1681687
 * change the way to compute font size when this one is null in DA:
   - make fontSize proportional to line height
   - in multiline case, take into account the number of lines for text entered to adapt the font size

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -1268,16 +1268,23 @@ class WidgetAnnotation extends Annotation {
     if (!annotationStorage || isPassword) {
       return null;
     }
-    const value =
+    let value =
       annotationStorage[this.data.id] && annotationStorage[this.data.id].value;
     if (value === undefined) {
       // The annotation hasn't been rendered so use the appearance
       return null;
     }
 
+    value = value.trim();
+
     if (value === "") {
       // the field is empty: nothing to render
       return "";
+    }
+
+    let lineCount = -1;
+    if (this.data.multiLine) {
+      lineCount = value.split(/\r\n|\r|\n/).length;
     }
 
     const defaultPadding = 2;
@@ -1297,8 +1304,12 @@ class WidgetAnnotation extends Annotation {
       );
     }
 
+    const [defaultAppearance, fontSize] = this._computeFontSize(
+      totalHeight,
+      lineCount
+    );
+
     const font = await this._getFontData(evaluator, task);
-    const fontSize = this._computeFontSize(font, totalHeight);
 
     let descent = font.descent;
     if (isNaN(descent)) {
@@ -1306,7 +1317,6 @@ class WidgetAnnotation extends Annotation {
     }
 
     const vPadding = defaultPadding + Math.abs(descent) * fontSize;
-    const defaultAppearance = this.data.defaultAppearance;
     const alignment = this.data.textAlignment;
 
     if (this.data.multiLine) {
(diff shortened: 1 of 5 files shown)


src/core/annotation.js
```diff
@@ -1359,26 +1369,41 @@ class WidgetAnnotation extends Annotation {
...
+  _computeFontSize(font, fontName, fontSize, height, lineCount) {
     if (fontSize === null || fontSize === 0) {
-      const em = font.charsToGlyphs("M")[0].width / 1000;
-      // According to https://en.wikipedia.org/wiki/Em_(typography)
-      // an average cap height should be 70% of 1em
-      const capHeight = 0.7 * em;
-      // 1.5 * capHeight * fontSize seems to be a good value for lineHeight
-      fontSize = Math.max(1, Math.floor(height / (1.5 * capHeight)));
+      if (lineCount === -1) {
+        fontSize = Math.round(0.8 * height * 10) / 10;
```
timvandermeij, line 1375: The value 0.8 used quite some times in this function should become a constant with a descriptive name of what it does. Just looking at the code doesn't allow me to figure out what its purpose is and why it's 0.8. In general this function is a bit "magic" due to the magic numbers but also because of for example this calculation, so maybe a comment would be good here to explain the logic in the various branches.

src/core/annotation.js
```diff
@@ -1359,26 +1369,41 @@ class WidgetAnnotation extends Annotation {
...
+      } else {
+        // Hard to guess how many lines there are
```
timvandermeij, line 1377: I don't understand this comment: we know exactly how many lines there are, namely `lineCount`?

src/core/annotation.js
```diff
@@ -1359,26 +1369,41 @@ class WidgetAnnotation extends Annotation {
...
+        // let's say that fonsize is 10.
+        fontSize = 10;
+        let lineHeight = fontSize / 0.8;
+        let numberOfLines = Math.round(height / lineHeight);
```
timvandermeij, line 1381: Shouldn't we use `Math.ceil` here? If there are 5.05 lines, this would round it down to 5, causing a part of the line to be cut off?

src/core/annotation.js
```diff
@@ -1359,26 +1369,41 @@ class WidgetAnnotation extends Annotation {
...
-      // According to https://en.wikipedia.org/wiki/Em_(typography)
-      // an average cap height should be 70% of 1em
-      const capHeight = 0.7 * em;
-      // 1.5 * capHeight * fontSize seems to be a good value for lineHeight
-      fontSize = Math.max(1, Math.floor(height / (1.5 * capHeight)));
+      if (lineCount === -1) {
+        fontSize = Math.round(0.8 * height * 10) / 10;
+      } else {
+        // Hard to guess how many lines there are
+        // let's say that fonsize is 10.
```
timvandermeij, line 1378: Is this not defined anywhere in the PDF data? This is an assumption that surely doesn't always apply.

src/core/annotation.js
```diff
@@ -1359,26 +1369,41 @@ class WidgetAnnotation extends Annotation {
...
+        fontSize = 10;
+        let lineHeight = fontSize / 0.8;
+        let numberOfLines = Math.round(height / lineHeight);
+        // multiline so there are at least 2 lines
```
timvandermeij, line 1382: Why though? Even though it's a multiline field, that's only a flag and doesn't necessarily indicate that the entered value is in fact split over multiple lines, i.e., it can still just be a single line.

src/core/annotation.js
```diff
@@ -1359,26 +1369,41 @@ class WidgetAnnotation extends Annotation {
...
+        numberOfLines = Math.max(2, numberOfLines, lineCount);
+        // finally we have lineHeight
```
timvandermeij, line 1384: This comment and the one below can be removed as the code is clear enough here so that these comments don't add anything.

test/unit/annotation_storage_spec.js
```diff
@@ -50,12 +54,10 @@ describe("AnnotationStorage", function () {
         called = true;
       };
       annotationStorage.onSetModified = callback;
-      annotationStorage.getOrCreateValue("asdf", { value: "original" });
-      expect(called).toBe(false);
 
       // not changing value
       annotationStorage.setValue("asdf", { value: "original" });
-      expect(called).toBe(false);
+      expect(called).toBe(true);
```
timvandermeij, line 60: We're not testing the `false` case anymore now, and the comment above this test don't match with the test anymore.

src/core/annotation.js
calixteman: Yeah agreed with the "magic" aspect. It just looks good with this value (at least for my eyes).
The current way to produce obviously too big font and I guess the 1.5 I found is line height in a paragraph and not for an input.

src/core/annotation.js
calixteman: The problem here is that if the textarea is "designed" to have 3 lines and the user enters text on 1 line then the font will be too big.
It's exactly what happens in the example gave by the user in the bug.

src/core/annotation.js
calixteman: Sure but need to start from something...
As said I've no idea on the number of lines we've for this textarea and we can't rely on the number we've from the user input.
Any idea is welcome.
I thought that maybe we could try to infer it using the font size used in the page itself, if all the text is 13 for example then it can make sense to have 13 for the input too. But if the page contains different fonts at different sizes....

src/core/annotation.js
calixteman: Agreed too... I've no idea on what the pdf creator had in mind...
So we consider that it should one line and we've a height of 100 then it will result in a too big font probably.
With the pdf from the bug, pdfium is rendering "correctly", I mean it considers that the textarea has 3 lines.
I discovered that pdfium computes the font size using field width too (to be sure that the entered text doesn't overflow the field).

src/core/annotation.js
```diff
@@ -1371,26 +1381,52 @@ class WidgetAnnotation extends Annotation {
...
     if (fontSize === null || fontSize === 0) {
-      const em = font.charsToGlyphs("M")[0].width / 1000;
-      // According to https://en.wikipedia.org/wiki/Em_(typography)
-      // an average cap height should be 70% of 1em
-      const capHeight = 0.7 * em;
-      // 1.5 * capHeight * fontSize seems to be a good value for lineHeight
-      fontSize = Math.max(1, Math.floor(height / (1.5 * capHeight)));
+      // A zero value for size means that the font shall be auto-sized:
+      // its size shall be computed as a function of the height of the
+      // annotation rectangle (see 12.7.3.3).
```
dhufnagel, line 1388: Shouldn't auto-sized also be calculated to fit all text inside the box? So if the text would be too long to fit inside the box, the font-size should be calculated to fit.

calixteman: AFAIK there is nothing about that in the pdf spec but pdfium is doing that so maybe we should too, but in an other patch.

src/core/annotation.js
```diff
@@ -1264,18 +1264,25 @@ class WidgetAnnotation extends Annotation {
...
     if (value === undefined) {
       // The annotation hasn't been rendered so use the appearance
       return null;
     }
 
+    value = value.trim();
+
     if (value === "") {
       // the field is empty: nothing to render
       return "";
```
Snuffleupagus, line 1278: A pre-existing issue, but why are we simply not returning `null` here if there's in fact nothing to render?

That's what we do in the `value === undefined` case just above, so what's special about *this* code-path?

src/display/annotation_storage.js
```diff
@@ -33,20 +33,19 @@ class AnnotationStorage {
...
+   * or return the default value
    *
    * @public
    * @memberof AnnotationStorage
    * @param {string} key
    * @param {Object} defaultValue
    * @returns {Object}
    */
-  getOrCreateValue(key, defaultValue) {
+  getOrDefaultValue(key, defaultValue) {
```
Snuffleupagus, line 44: Any reason to not simply naming this `getValue` instead now, since it would mirror the existing `setValue` method nicely?

src/core/annotation.js
calixteman: In this case fontSize will be readjusted to for these 5 lines.

src/core/annotation.js
calixteman: Returning `null` means that we'll use `AP::N` to render the field.
Here `value===""` means that the field is empty so it could be rendered using  `/Font 10 Tf 0 g () Tj` which is useless.

PR #12762, opened 2020-12-20
Await manually triggered `dispatchEventInSandbox` calls in the viewer

Snuffleupagus:
- Await manually triggered `dispatchEventInSandbox` calls in the viewer

   *As usual, this only occured to me after the recent round of scripting-related patches had landed.*

   Given that the `dispatchEventInSandbox` method (on the scripting-classes) is asynchronous, there's a very real risk that things such as e.g. "WillPrint" and "WillSave" are dispatched/handled *after* the associated functionality has run.
   To reduce the likelihood of that happening, we can simply `await` the `dispatchEventInSandbox` calls as necessary. A couple of methods are now marked as `async` to support these changes, however that shouldn't be a problem as far as I can tell.

 - Pass in the "sandboxBundleSrc" option when calling `DefaultExternalServices.createScripting`

   Similar to e.g. the "locale" option, this in *only* done for those build-targets where the "sandboxBundleSrc" is actually defined.
   With these changes we can remove an `AppOptions` dependency from the `web/generic_scripting.js` file, thus limiting *direct* `AppOptions` usage in the default viewer files.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1013,7 +1013,7 @@ const PDFViewerApplication = {
       .catch(downloadByUrl); // Error occurred, try downloading with the URL.
   },
 
-  save({ sourceEventType = "download" } = {}) {
+  async save({ sourceEventType = "download" } = {}) {
     if (this._saveInProgress) {
       return;
     }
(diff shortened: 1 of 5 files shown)


Snuffleupagus:
Note that the [one failure above](https://github.com/mozilla/pdf.js/pull/12762#issuecomment-748597221) is unrelated to the recently fixed one, and is a pre-existing (but rare) intermittent failure caused by *another* arbitrary timeout in https://github.com/mozilla/pdf.js/blob/c3730c177a2ff7b9ba12b81027d0a2bc1252044f/test/integration/annotation_spec.js#L33-L50

PR #12773, opened 2020-12-23
Pass in the "sandboxBundleSrc" option when calling `DefaultExternalServices.createScripting`

Snuffleupagus:
Similar to e.g. the "locale" option, this in *only* done for those build-targets where the "sandboxBundleSrc" is actually defined.
With these changes we can remove an `AppOptions` dependency from the `web/generic_scripting.js` file, thus limiting *direct* `AppOptions` usage in the default viewer files.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -167,7 +167,7 @@ class DefaultExternalServices {
     throw new Error("Not implemented: createL10n");
   }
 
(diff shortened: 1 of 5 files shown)


calixteman (approved):
Thanks for doing this.

PR #12774, opened 2020-12-23
JS -- Add tests for print/save actions

calixteman:
* change PDFDocument::hasJSActions to return true when there are JS actions in catalog.


test/pdfs/.gitignore
```diff
@@ -225,6 +226,7 @@
 !issue4875.pdf
 !issue11740_reduced.pdf
 !issue12705.pdf
+!doc_actions.pdf
```
Snuffleupagus, line 229: This line probably doesn't need to be duplicated :-)

test/test_manifest.json
```diff
@@ -1015,6 +1015,12 @@
        "firstPage": 2,
        "type": "eq"
     },
+    {  "id": "doc_actions",
```
Snuffleupagus, line 1018: There's already a file with an *almost* identical name, see `docactions.js`.
Can we perhaps use a that file for these new tests as well, or at least use a bit less similar filenames please?

timvandermeij (approved):

timvandermeij:
Thank you for extending the test coverage!

test/integration/scripting_spec.js
```diff
@@ -280,4 +289,76 @@ describe("Interaction", () => {
...
+    afterAll(async () => {
+      await closePages(pages);
+    });
+
+    it("must execute WillPrint and DidPrint actions", async () => {
+      await Promise.all(
+        pages.map(async ([browserName, page]) => {
+          if (process.platform === "win32" && browserName === "firefox") {
+            // Doesn't work because of bug 1662471
+            return;
```
Snuffleupagus, line 309: Sorry, I just noticed this now and that's *really not* how you *temporarily* disable a test, since that pretty much guarantees that no one will ever fix it.
The correct way, since we're using Jasmine, is to add a `pending()` statement; i.e.
```javascript
pending("Disabled in Firefox on Windows, because of bug 1662471.");
```

timvandermeij: Good point, I also missed this one. @calixteman Let's address this one in a follow-up PR.

PR #12777, opened 2020-12-25
Add support for finding/highlighting the outlineItem, corresponding to the currently visible page, in the sidebar (issue 7557, bug 1253820, bug 1499050)

### LINKED ISSUE #7557: outline sidebar and current position in sync
What I mean is a visual indication on the outline sidebar of where you're currently browsing the document.  I find this very useful information and its a common behavior among other pdf viewers.

Snuffleupagus:
This implementation is inspired by the behaviour in (recent versions of) Adobe Reader, since it leads to reasonably simple and straightforward code as far as I'm concerned.
*Specifically:* We'll only consider *one* destination per page when finding/highlighting the current outline item, which is similar to e.g. Adobe Reader, and we choose the *first* outline item at the *lowest* level of the outline tree.

Given that this functionality requires not only parsing of the `outline`, but looking up *all* of the destinations in the document, this feature can when initialized have a non-trivial performance overhead for larger PDF documents.
In an attempt to reduce the performance impact, the following steps are taken here:

 - The "find current outline item"-functionality will only be enabled once *one* page has rendered and *all* the pages have been loaded[1], to prevent it interfering with data regular fetching/parsing early on during document loading and viewer initialization.

 - With the exception of a couple of small and simple `eventBus`-listeners, in `PDFOutlineViewer`, this new functionality is initialized *lazily* the first time that the user clicks on the `currentOutlineItem`-button.

 - The entire "find current outline item"-functionality is disabled when `disableAutoFetch = true` is set, since it can easily lead to the setting becoming essentially pointless[2] by triggering *a lot* of data fetching from a relatively minor viewer-feature.

 - Fetch the destinations *individually*, since that's generally more efficient than using `PDFDocumentProxy.getDestinations` to fetch them all at once. Despite making the overall parsing code *more* asynchronous, and leading to a lot more main/worker-thread message passing, in practice this seems faster for larger documents.

Finally, we'll now always highlight an outline item that the user manually clicked on, since only highlighting when the new "find current outline item"-functionality is used seemed inconsistent.

Fixes #7557
Fixes [bug 1253820](https://bugzilla.mozilla.org/show_bug.cgi?id=1253820)
Fixes [bug 1499050](https://bugzilla.mozilla.org/show_bug.cgi?id=1499050)

---
[1] Keep in mind that the `outline` itself already isn't fetched/parsed until at least *one* page has been rendered in the viewer.

[2] And also quite slow, since it can take a fair amount of time to fetch all of the necessary `destinations` data when `disableAutoFetch = true` is set.


timvandermeij:
Could we move the second commit to its own PR to make this PR a bit smaller?

Moreover, am I correct in that the intention here is not to let the sidebar follow the document during scrolling automatically but only when pressing the button? I assume that's also what Adobe Reader does?

Finally, this needs a different image since the current one seems equal to the outline item SVG. What icon does Adobe Reader use for this button? Maybe that can give some inspiration as to what it should look like.

Snuffleupagus:
> Could we move the second commit to its own PR to make this PR a bit smaller?

Of course, please see PR #12796.

> Moreover, am I correct in that the intention here is not to let the sidebar follow the document during scrolling automatically but only when pressing the button?

Correct, since the amount of DOM-lookups and outline scrolling that this would result in when the user scrolls through the document seems generally undesirable (for a minor feature). With the current implementation, we avoid this feature having a performance impact/overhead by default.

> I assume that's also what Adobe Reader does?

Yes, there's a button to find the current item (which served as inspiration for this patch).

> Finally, this needs a different image since the current one seems equal to the outline item SVG.

Of course, but UI-design isn't exactly my strong point and I simply wanted *something* for testing purposes :-)

> What icon does Adobe Reader use for this button? Maybe that can give some inspiration as to what it should look like.

Please see the screen-shot below, with the button outlined, however that's probably not entirely desirable for PDF.js since we use an "outline" icon whereas Adobe Reader uses a "bookmark" one.
Perhaps we want *some* variant of our existing "outline" icon to be used here!?

find

janpe2:
Maybe the icon could be like this?
[svg_icon.zip](https://github.com/mozilla/pdf.js/files/5765225/svg_icon.zip)
kuva

Snuffleupagus:
> Maybe the icon could be like this?
> [svg_icon.zip](https://github.com/mozilla/pdf.js/files/5765225/svg_icon.zip)

That looks *great*, thanks a lot for your help!
(I've added the icon in a separate commit, to credit you as the author.)

web/pdf_sidebar.js
```diff
@@ -62,6 +53,10 @@ const SidebarView = {
  *   the attachments are placed.
  * @property {HTMLDivElement} layersView - The container in which
  *   the layers are placed.
+ * @property {HTMLDivElement} outlineOptionsContiner - The container in which
```
timvandermeij, line 56: `Continer` -> `Container`

web/viewer.html
```diff
@@ -76,19 +76,31 @@
...
+              </button>
+            </div>
+          </div>
+
+          <div id="toolbarSidebarRight">
+            <div id="outlineOptionsContainer" class="hidden">
+              <div class="verticalToolbarSeparator"></div>
+
+              <button id="currentOutlineItem" class="toolbarButton" disabled="disabled" title="Find Current Outline Item" tabindex="6" data-l10n-id="current_outline_item">
+                <span data-l10n-id="current_outline_item">Current Outline Item</span>
```
timvandermeij, line 101: This should be `current_outline_item_label` for the localization to work.

timvandermeij (commented):
Looks good with these comments addressed. Thanks!

timvandermeij:
If you'd like to add the Dutch translations while we're making changes here, feel free to use these:

```
current_outline_item.title=Huidige positie in documentoverzicht selecteren
current_outline_item_label=Huidige positie in documentoverzicht
```

web/pdf_sidebar.js
Snuffleupagus: D'oh; good eye!

web/viewer.html
Snuffleupagus: Well spotted, thanks!

Snuffleupagus:
> If you'd like to add the Dutch translations while we're making changes here,

Thank you, those l10n strings have been included :-)

As always, thanks for helping out with review this!

timvandermeij (approved):

timvandermeij:
Looks great; thanks!

PR #12788, opened 2020-12-28
Include the `state` in the "presentationmodechanged" event, and remove the separate `active`/`switchInProgress` properties

Snuffleupagus:
Given that we already have a `PresentationModeState`-enumeration, we should use that with the "presentationmodechanged" event rather than including separate properties. Note that this new behaviour, of including an enumeration-value in the event, is consistent with lots of other existing viewer-events.

To hopefully avoid issues in custom implementations of the default viewer, any attempt to access the removed properties will now throw.

### DIFF
--- a/web/pdf_presentation_mode.js
+++ b/web/pdf_presentation_mode.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import { normalizeWheelEventDelta } from "./ui_utils.js";
+import { normalizeWheelEventDelta, PresentationModeState } from "./ui_utils.js";
 
 const DELAY_BEFORE_RESETTING_SWITCH_IN_PROGRESS = 1500; // in ms
 const DELAY_BEFORE_HIDING_CONTROLS = 3000; // in ms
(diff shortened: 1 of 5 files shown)


web/pdf_presentation_mode.js
```diff
@@ -183,11 +183,34 @@ class PDFPresentationMode {
...
-      switchInProgress: !!this.switchInProgress,
-    });
+    let state = PresentationModeState.NORMAL;
+    if (this.switchInProgress) {
+      state = PresentationModeState.CHANGING;
+    } else if (this.active) {
+      state = PresentationModeState.FULLSCREEN;
+    }
+
+    if (typeof PDFJSDev === "undefined" || !PDFJSDev.test("MOZCENTRAL")) {
```
timvandermeij, line 193: Shouldn't this be `typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")` instead since the throwing is only for third-party usage and therefore not needed if we are targeting mozcentral?

web/pdf_presentation_mode.js
```diff
@@ -183,11 +183,34 @@ class PDFPresentationMode {
...
+        source: this,
+        state,
+      });
+    } else {
+      this.eventBus.dispatch("presentationmodechanged", {
+        source: this,
+        state,
+        get active() {
+          throw new Error(
+            "Removed parameter: `active`, please use `state` instead."
```
timvandermeij, line 204: Would it be useful to use the word `Deprecated` here and below for easier searching later on to remove it?

timvandermeij (commented):

web/pdf_presentation_mode.js
Snuffleupagus: D'oh; of course it should :-)

Embarrassingly enough I managed to copy-paste the wrong thing here, thanks for spotting this!

web/pdf_presentation_mode.js
Snuffleupagus: Of course that's better, fixed now :-)

timvandermeij (approved):

timvandermeij:
Looks good!

PR #12793, opened 2020-12-29
Support the `once` option, when registering `EventBus` listeners

Snuffleupagus:
This follows the same principle as the `once` option that exists in the native `addEventListener` method, and will thus automatically remove an `EventBus` listener when it's invoked; see https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Parameters

Finally, this patch also tweaks some the existing `EventBus`-code to use modern features such as optional chaining and logical assignment operators.

### DIFF
--- a/test/unit/ui_utils_spec.js
+++ b/test/unit/ui_utils_spec.js
@@ -312,6 +312,30 @@ describe("ui_utils", function () {
       expect(count).toEqual(2);
     });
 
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #12800, opened 2020-12-30
Convert `DefaultExternalServices.fallback` to an asynchronous method

Snuffleupagus:
This method currently accepts a callback-function, which does feel a bit old fashioned now. At the time that this code was introduced, native Promises didn't exist yet and there's a custom Promise-implementation used instead.

However, today with Promises and async/await being used *a lot* it seems reasonable to change `DefaultExternalServices.fallback` to an `async` method instead such that the callback-function can be removed.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -150,7 +150,7 @@ class DefaultExternalServices {
 
   static initPassiveLoading(callbacks) {}
 
-  static fallback(data, callback) {}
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks much better. Thanks!

PR #12802, opened 2021-01-01
Add a new `FirefoxCom.requestAsync` method, to simplify the code in `web/firefoxcom.js`

Snuffleupagus:
*Please note:* I've tested this patch in a *local* Firefox build, to ensure that I didn't break anything here :-)

 - Convert `FirefoxCom` to a class, with `static` methods

   *Please note:* It's highly recommended to ignore whitespace-only changes when looking at this patch.

   Besides modernizing this code, by converting it to a standard class, the existing JSDoc comments are updated to actually agree better with the way that this functionality is used now. (The next patch will reduce usage of `FirefoxCom.request` significantly, hence the JSDocs for the optional `callback` is removed to not unnecessarily advertise that functionality.)

   Finally, the unnecessary/unused `return` statement at the end of `FirefoxCom.request` is also removed.

 - Add a new `FirefoxCom.requestAsync` method, to simplify the code in `web/firefoxcom.js`

   There's a fair number of cases where `FirefoxCom.request`-calls are manually wrapped in a Promise to make it asynchronous. We can reduce the amount of boilerplate code in these cases by introducing a new `FirefoxCom.requestAsync` method instead.

   Furthermore, a couple of `FirefoxCom.request`-calls in the `DownloadManager` are also changed to be asynchronous rather than using callback-functions.
   With this patch, we're thus able to replace a lot of *direct* usages of `FirefoxCom.request` with the new `FirefoxCom.requestAsync` method instead.

### DIFF
--- a/web/firefoxcom.js
+++ b/web/firefoxcom.js
@@ -29,8 +29,8 @@ class FirefoxCom {
   /**
    * Creates an event that the extension is listening for and will
    * synchronously respond to.
-   * NOTE: It is recommended to use request() instead since one day we may not
-   * be able to synchronously reply.
+   * NOTE: It is recommended to use requestAsync() instead since one day we may
+   *       not be able to synchronously reply.
    * @param {string} action - The action to trigger.
    * @param {Object|string} [data] - The data to send.
    * @returns {*} The response.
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for modernizing this code!

PR #12804, opened 2021-01-02
Remove the `DownloadManager.onerror` functionality, since its only usage is unlikely to be helpful

Snuffleupagus:
- Add a new helper method, on `PDFViewerApplication`, to determine the document filename

   Currently this code is duplicated no less than three times in the `web/app.js` file, and by introducing a helper method we can avoid unnecessary repetition.

 - Remove the `DownloadManager.onerror` functionality, since its only usage is unlikely to be helpful

   Note how the `onerror` functionality is not being used in the GENERIC `DownloadManager`, since we have no way of knowing if downloading succeeded.
   Hence this functionality is only *possibly* useful in MOZCENTRAL builds, however as outlined in the existing comments it's unlikely to be helpful in practice. Generally speaking, if downloading failed once in [`PdfStreamConverter.jsm`](https://searchfox.org/mozilla-central/rev/809ac3660845fef6faf18ec210232fdadc0f1ad9/toolkit/components/pdfjs/content/PdfStreamConverter.jsm#294-406) it seems very likely that it would fail again; all-in-all I'm thus suggesting that we just remove the `onerror` functionality altogether here.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -763,6 +763,12 @@ const PDFViewerApplication = {
     document.title = title;
   },
 
+  get _docFilename() {
+    // Use `this.url` instead of `this.baseUrl` to perform filename detection
+    // based on the reference fragment as ultimate fallback if needed.
+    return this._contentDispositionFilename || getPDFFileNameFromURL(this.url);
+  },
+
   /**
    * @private
    */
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Nice, thanks!

PR #12805, opened 2021-01-02
Remove unnecessary toolbarButton icon-flipping in RTL mode (PR 11077 follow-up)

Snuffleupagus:
With the updated default viewer UI, a couple of the toolbarButton icons are now *vertically* symmetrical; hence we can remove some now unneeded `transform: scaleX(-1);` rules from the viewer CSS.


timvandermeij (approved):

timvandermeij:

PR #12807, opened 2021-01-03
Remove unnecessary `dir`-dependent CSS rules, and add a missing CSS variable (PR 11077 follow-up)

Snuffleupagus:
With the updated default viewer UI, a couple of `dir`-dependent CSS rules have now become redundant since *identical* rules are being specified for both LTR and RTL mode.

Furthermore, there's also some unnecessary re-defining of the `toolbarButton`/`secondaryToolbarButton`-icon related CSS rules.

Finally, for the toggle-buttons there's a particular styling applied to the `:hover:active` state, however the color wasn't defined with CSS variables.


timvandermeij (approved):

timvandermeij:
Thanks!

Issue #12295, opened 2020-08-27
Line drawing  (Page 16)   does not have the correct line thickness when using the demo application on Google Chrome or any brower


Marsrobot2008:
Steps to reproduce the problem:
1. When using the demo application - https://mozilla.github.io/pdf.js/web/viewer.html, load  the document 5311693-1.pdf

[5311693-1.pdf](https://github.com/mozilla/pdf.js/files/5137225/5311693-1.pdf)

2. go to Page 16 of the document.

3.  The strip of  EKG data  appears to have an incorrect line width   under the  Multi-color heat rate chart.

4.  When the same document is generated by pdffactory ,  the artifact are not visible.

What is the expected behavior? (add screenshot)
![expected](https://user-images.githubusercontent.com/3577617/91461157-4622cb80-e84e-11ea-8e5d-750c7b4e64ae.PNG)

What went wrong? (add screenshot)
The EKG line strip  has an incorrect thickness.  the x-transform seems to stretch the line thickness.  See the  stri in the RED rectagngle of the attached behavior versus the GREEN rectangle

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):
 https://mozilla.github.io/pdf.js/web/viewer.html

THausherr:
I hope that Halima A. agreed to have her cardio health data in this ticket!

waluyo102:
Is good ilike

janpe2:
This issue has been fixed by pull request #12812

PR #12812, opened 2021-01-04
Enforce line width to be at least 1px after applied transform

### LINKED ISSUE #12810: Rendering Issue in Chrome: Lose some green dots in rendering one PDF exported from AutoCAD
[Attach](url
[Test1.pdf](https://github.com/mozilla/pdf.js/files/5763185/Test1.pdf)
)

Configuration:
- Browser: Chrome 87.0.4280.88
- OS: MacOS 10.15.4
- PDF.js version: 2.5.207

1. Open the viewer.html in above pdf.js package
2. Open the attached **_Test1.pdf_** in the viewer.html

All contents should be rendered properly, just like the attached **_right.png_**
right

What went wrong? 
Some green dots in left side are lost, as highlighted in the **_wrong.png_**. And this issue only exists in the Chrome browser, it renders properly in safari and Firefox.
wrong

calixteman:
* add a comment to explain how minimal linewidth is computed.
* when context.linewidth < 1 after transform, firefox and chrome don't render in the same way (issue #12810).
* set lineWidth to 1 after transform and before stroking 
  - aims fix issue #12295   
  - a pixel can be transformed into a rectangle with both heights < 1. A single rescale leads to a rectangle with dim equals to 1 and      the other to something greater than 1. 
* change the way to render rectangle with null dimensions: 
  - right now we rely on the lineWidth set before "re" but      it can be set after "re" and before "S" and in this case the rendering      will be wrong.   
  - render such rectangles as a single line.


src/display/canvas.js
```diff
@@ -38,8 +38,8 @@ const MIN_FONT_SIZE = 16;
 const MAX_FONT_SIZE = 100;
 const MAX_GROUP_SIZE = 4096;
 
-// Heuristic value used when enforcing minimum line widths.
-const MIN_WIDTH_FACTOR = 0.65;
+// Enforce line to have a minimal width of 1px when transform is applied
```
Snuffleupagus, line 41: This is a sentence without a period at the end :-)

src/display/canvas.js
```diff
@@ -38,8 +38,8 @@ const MIN_FONT_SIZE = 16;
...
+const MIN_WIDTH_FACTOR = 1.000001;
```
Snuffleupagus, line 42: Also, given all the different `getSinglePixelWidth` and `MIN_WIDTH_FACTOR` usages throughout this file; are these changes generally correct/safe (as in not regressing existing tests)?

Unfortunately PR #4615 landed without a test-case included, so please also *manually* test the issues/bugs listed there to ensure that you're not regressing anything here!

src/shared/util.js
```diff
@@ -710,19 +710,27 @@ class Util {
   // The SVD can be represented with formula A = USV. We are interested in the
   // matrix S here because it represents the scale values.
   static singularValueDecompose2dScale(m) {
-    const transpose = [m[0], m[2], m[1], m[3]];
+    if (Math.abs(m[1]) <= Number.EPSILON && Math.abs(m[2]) <= Number.EPSILON) {
```
Snuffleupagus, line 713: The changes in this file probably belongs in a *separate commit*, since they seem orthogonal to the rest of the changes.

src/shared/util.js
```diff
@@ -710,19 +710,27 @@ class Util {
...
+    const c = m[2] * m[0] + m[3] * m[1];
+    const d = m[2] * m[2] + m[3] * m[3];
 
     // Solve the second degree polynomial to get roots.
-    const first = (a + d) / 2;
-    const second = Math.sqrt((a + d) * (a + d) - 4 * (a * d - c * b)) / 2;
-    const sx = first + second || 1;
-    const sy = first - second || 1;
+    const sum_a_d = a + d;
+    const second = Math.sqrt(sum_a_d * sum_a_d - 4 * (a * d - c * b));
```
Snuffleupagus, line 731:
```suggestion
    const second = Math.sqrt(sum_a_d ** 2 - 4 * (a * d - c * b));
```

Also, the variable name `second` looks slightly out-of-place here now.

src/shared/util.js
```diff
@@ -710,19 +710,27 @@ class Util {
...
-    const transpose = [m[0], m[2], m[1], m[3]];
+    if (Math.abs(m[1]) <= Number.EPSILON && Math.abs(m[2]) <= Number.EPSILON) {
+      // Diagonal matrix
+      return [Math.abs(m[0]) || 1, Math.abs(m[3]) || 1];
+    }
+
+    if (Math.abs(m[0]) <= Number.EPSILON && Math.abs(m[3]) <= Number.EPSILON) {
+      // Anti-diagonal matrix: scale composed with rotation of +/- 90°
+      return [Math.abs(m[2]) || 1, Math.abs(m[1]) || 1];
+    }
```
Snuffleupagus, line 721: Given the additional checks added here, how much of a performance increase are these changes bringing in the general case (do you at least have some ballpark numbers)?

Snuffleupagus (changes requested):
Besides some inline comment; can we add a reference-test for this?

janpe2:
Does this PR improve also the rendering of issue #12295? The problems in that issue are caused by the existing implementation of `getSinglePixelWidth()` because the scaling factors differ greatly in the horizontal and vertical direction.

src/display/canvas.js
calixteman: I checked all the pdf mentioned in the above PR and they all looks good.
I ran locally ref tests and the differences are minimal.
I'd say that with my eyes the rendering on firefox-linux and on acrobat-windows are very close (same screen).

Issue #12813, opened 2021-01-04
webpack instructions can't work with last worker-loader versions


fdaveine:
I followed the recommendations [given here](https://github.com/mozilla/pdf.js/blob/master/examples/webpack/README.md) to use pdf.js `2.6.347` with webpack.

It mention that:
> ...(configuring worker-loader is not necessary; just installing it is sufficient).

It's no longer the case with the last version of `worker-loader` (`v3.0.7`), and since the version [v3.0.0](https://github.com/webpack-contrib/worker-loader/releases/tag/v3.0.0) I believe, due to a breaking change in this package:

![image](https://user-images.githubusercontent.com/12005932/103556603-56a91980-4eb2-11eb-98aa-ad8455e7edba.png)

The `filename` option now [default to [name].worker.js](https://github.com/webpack-contrib/worker-loader#filename), which cause webpack to generate a `pdf.worker.worker.js` file instead of the expected `pdf.worker.js`.

To make webpack generate the proper file, I had to update [a line inside /external/dist/webpack.js](https://github.com/mozilla/pdf.js/blob/master/external/dist/webpack.js#L19) as mention bellow:

```javascript
var PdfjsWorker = require("worker-loader?esModule=false&filename=[name].js!pdfjs-dist/build/pdf.worker.js");
```
(I added `&filename=[name].js` option, for `worker-loader` not to add its new suffix)

fdaveine:
Let me know if it would be helpful that I make a PR for this.

timvandermeij:
A PR for this would be very welcome, most likely together with a small change to the README file to say that at least `worker-loader` version 3.0.0 or higher is required now; thanks!

PR #12815, opened 2021-01-05
Update webpack example

Snuffleupagus:
- Update the webpack-versions used in `examples/webpack`

   Once the next PDF.js release is made, the `webpack` example will no longer work since the non-translated builds now use ECMAScript features not supported by older `webpack`-versions.

 - Try to re-add the `worker-loader` package, in pdfjs-dist, using `peerDependencies` (PR 11474 follow-up)

   There's been a number of (somewhat) recent issues where people are having trouble using pdfjs-dist together with `webpack`, since that library purposely doesn't declare any dependencies; refer to PR #11474 for additional context.

   In an *attempt*, although I don't know how much this will actually help in practice (given my limited `webpack` experience), let's try to list `worker-loader` as a *peer*-dependency to see if that helps. This should, unless I'm completely misunderstanding https://docs.npmjs.com/cli/v6/configuring-npm/package-json#peerdependencies, prevent `worker-loader` from being installed by default for *all* pdfjs-dist users while still indicating the dependency for those who need it.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1969,6 +1969,9 @@ function packageBowerJson() {
     homepage: DIST_HOMEPAGE,
     bugs: DIST_BUGS_URL,
     license: DIST_LICENSE,
+    peerDependencies: {
+      "worker-loader": "^3.0.7", // Used in `external/dist/webpack.js`.
+    },
     browser: {
       canvas: false,
       fs: false,
--- a/examples/webpack/package.json
+++ b/examples/webpack/package.json
@@ -5,8 +5,8 @@
     "build": "webpack"
   },
   "devDependencies": {
-    "webpack": "4.21.0",
-    "webpack-cli": "^3.1.2",
+    "webpack": "^5.11.1",
+    "webpack-cli": "^4.3.1",
     "pdfjs-dist": "../../node_modules/pdfjs-dist"
   }
 }


gulpfile.js
```diff
@@ -1969,6 +1969,9 @@ function packageBowerJson() {
     homepage: DIST_HOMEPAGE,
     bugs: DIST_BUGS_URL,
     license: DIST_LICENSE,
+    peerDependencies: {
+      "worker-loader": "^3.0.7", // Used in `external/dist/webpack.json`.
```
timvandermeij, line 1973: `json` -> `js` ;-)

timvandermeij (commented):
I'm also not really familiar with the concept of peer dependencies, but this suggestion has been made before, so I'm going to assume that this is correct. Did you check that this doesn't re-introduce #9248 for example?

gulpfile.js
Snuffleupagus: That's what I get for copy-pasting from old code :-)

Snuffleupagus:
> Did you check that this doesn't re-introduce #9248 for example?

I'm honestly not entirely sure how to do test this locally, unfortunately!
However, when that issue was filed `worker-loader` was listed as a *regular* dependency and now we only list it as a *peer* dependency which should hopefully help; so that exact problem shouldn't be re-introduced as far as I can tell.

fdaveine:
@Snuffleupagus I was going to create a PR for #12813 but then I saw this PR, which is related to my issue, should I wait for this PR to be merged to create mine?

Snuffleupagus:
> should I wait for this PR to be merged to create mine?

As far as I can tell, you're not going to touch the same files and consequently there shouldn't be any conflicts (not that those are difficult to resolve anyway); hence I don't really see why you'd need to wait :-)

timvandermeij (approved):

timvandermeij:
Thank you!

PR #12816, opened 2021-01-05
JS -- Plug PageOpen and PageClose actions

calixteman:


web/app.js
```diff
@@ -1542,6 +1542,33 @@ const PDFViewerApplication = {
     };
     internalEvents.set("updatefromsandbox", updateFromSandbox);
 
+    const visitedPages = new Set();
+    const pageOpen = async ({ pageNumber }) => {
+      let actions = null;
+      if (!visitedPages.has(pageNumber)) {
+        // Get actions only the first time the page is visited
```
Snuffleupagus, line 1549: The comment is misleading here, since you don't actually prevent multiple fetches of the same data until it's resolved at least *once*.
Furthermore, there's also generally speaking a risk that this data resolves *after* the page has been "closed" , meaning that the event-ordering is not guaranteed. Is that not an issue?

Also, you shouldn't generally need to re-fetch the `PDFPageProxy` instance since it will usually be available already here.

web/app.js
```diff
@@ -1542,6 +1542,33 @@ const PDFViewerApplication = {
...
+        const page = await pdfDocument.getPage(pageNumber);
+        actions = await page.getJSActions();
+        visitedPages.add(pageNumber);
+      }
```
Snuffleupagus, line 1553: If you want to prevent *some* ordering issues, since this event-handler is currently asynchronous in some cases and not in others, something like the following could work I think:
```suggestion
    const visitedPages = new Map();
    const pageOpen = async ({ pageNumber }) => {
      if (!visitedPages.has(pageNumber)) {
        // Get actions only the first time the page is visited.
        const promise = new Promise(resolve => {
          const pageView = this.pdfViewer.getPageView(/* index = */ pageNumber - 1);
          if (pageView?.pdfPage) {
            pageView.pdfPage.getJSActions().then(resolve);
            return;
          }
          pdfDocument.getPage(pageNumber).then(pdfPage => {
            pdfPage.getJSActions().then(resolve);
          });
        });
        visitedPages.set(pageNumber, promise);
      }
      const actions = await visitedPages.get(pageNumber);
      
      if (pdfDocument !== this.pdfDocument) {
        return; // The document was closed while the actions resolved.
      }
```

---
This, however, still leaves the issue of all of this possibly happening *after* the associated "pageclose" event though.

Snuffleupagus (changes requested):

web/app.js
```diff
@@ -1542,6 +1542,33 @@ const PDFViewerApplication = {
...
+        actions,
+      });
+    };
+    const pageClose = ({ pageNumber }) => {
+      this._scriptingInstance.scripting.dispatchEventInSandbox({
+        id: "page",
+        name: "PageClose",
+        pageNumber,
+      });
+    };
```
Snuffleupagus, line 1568: To ensure that a "pageclose" event doesn't happen *before* "pageopen", since that doesn't seem reasonable (at least to me), the following should prevent that (based on the suggestion above):
```suggestion
    const pageClose = async ({ pageNumber }) => {
      if (!visitedPages.has(pageNumber)) {
        // Ignore the "pageclose" event if there wasn't a corresponding
        // "pageopen" event first.
        return;
      }
      await visitedPages.get(pageNumber);
      
      if (pdfDocument !== this.pdfDocument) {
        return; // The document was closed while the actions resolved.
      }
      
      await this._scriptingInstance.scripting.dispatchEventInSandbox({
        id: "page",
        name: "PageClose",
        pageNumber,
      });
    };
```

web/app.js
calixteman: Thx for the patch: I almost copied/pasted it.
My goal was to not send actions when a page was visited again, so I just adapted your patch to this.

web/app.js
calixteman: `visitedPages` is used to avoid to send actions again and again.
So I adapted the code, anyway thx for pointing this.

web/app.js
Snuffleupagus: > My goal was to not send actions when a page was visited again,

But why does sending it again matter though? You've already added code that should ignore that case in https://github.com/mozilla/pdf.js/pull/12816/files#diff-cd84875e4422389162fdb6971a6766a7666b89af6389e9039560ac32b342fae5R120-R122

The "issue" with the latest patch is, as far as I'm concerned, is that it's a lot less readable/understandable overall!

calixteman: Agreed, but at least for the quickjs version, we need to serialize/deserialize all the data which are evaluated in the sandbox.
So if we've some big scripts (do we irl ??), it's time/resource consuming.

Snuffleupagus: > Agreed, but at least for the quickjs version, we need to serialize/deserialize all the data which are evaluated in the sandbox.

I'm not overly worried about *that* part, as long as the mozilla-central version works OK; and I do think that we need easily readable code here (since otherwise it's too difficult to reason about it).

However, can you then instead please do what I suggested in my comments above/below, but with e.g. the following addendum here (it's possible this can become even shorter and more compact, but this should do the trick):
```javascript
const actions = await visitedPages.get(pageNumber);
if (actions) {
  // Avoid sending, and thus serializing, the `actions` data for each event.
  visitedPages.set(pageNumber, Promise.resolve(null));
}
```

web/app.js
```diff
@@ -1542,6 +1542,68 @@ const PDFViewerApplication = {
     };
     internalEvents.set("updatefromsandbox", updateFromSandbox);
 
+    const lastOpenPages = new Map();
```
brendandahl, line 1545: Can you add a comment explaining this section. The async function with double nested promises is rather confusing.

Snuffleupagus: Please see https://github.com/mozilla/pdf.js/pull/12816#discussion_r552157861, where I've added a suggestion that removes the need for the added complexity the the `lastOpenPages` brings.

calixteman:
The test error on windows is fixed with:
https://github.com/mozilla/pdf.js/pull/12819

PR #12819, opened 2021-01-06
Click on download and print buttons using element.click function in scripting integration tests

calixteman:
If for any reason the save/print buttons aren't on screen (no height), the `page.click` is failing.
So replace `page.click` by `document.querySelector(...).click()`.

### DIFF
--- a/test/integration/scripting_spec.js
+++ b/test/integration/scripting_spec.js
@@ -312,6 +312,9 @@ describe("Interaction", () => {
           await page.waitForFunction(
             "window.PDFViewerApplication.scriptingReady === true"
           );
+          await page.waitForSelector("#print", {
+            timeout: 0,
+          });
 
           await clearInput(page, "#\\34 7R");
           await page.evaluate(_ => {
@@ -322,7 +325,7 @@ describe("Interaction", () => {
           );
 
           let text = await actAndWaitForInput(page, "#\\34 7R", async () => {
-            await page.click("#print");
+            await page.evaluate(() => document.querySelector("#print").click());
           });
           expect(text).withContext(`In ${browserName}`).toEqual("WillPrint");
 
@@ -354,6 +357,9 @@ describe("Interaction", () => {
           await page.waitForFunction(
             "window.PDFViewerApplication.scriptingReady === true"
           );
+          await page.waitForSelector("#download", {
+            timeout: 0,
+          });
 
           try {
             // Disable download in chrome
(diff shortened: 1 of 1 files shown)


timvandermeij:
I'm a bit confused since the integration test that apparently broke in the other PR seems to just pass (as expected) now when I look at https://github.com/mozilla/pdf.js/pull/12394#issuecomment-756738845. Is this patch really still necessary given this?

calixteman:
When the test was failing, I had an open connection through rdp to the windows bot and it's probably the root cause.
I guess that because of a too small window, the download button had was not rendered.
So I think that this patch is not so bad because if for any reason in the future the layout change and the button is no more visible we'll handle the case.
If you think that's useless to have it, close it, I'm fine with that too.

calixteman:
I tried to wait on `button.height !== 0` and the test failed with timeout (30s): it's the reason why I think the button isn't visible.
When the window is too small then the buttons go into the `>>` menu and so you scroll or do whatever you want you cannot "physically" click on the button it's why using `button.click()` is working.

timvandermeij:
In that case, wouldn't setting the viewport as stated on https://stackoverflow.com/questions/51857070/puppeteer-in-nodejs-reports-error-node-is-either-not-visible-or-not-an-htmlele (using `page.setViewport`) be a better fix here (and in general for all tests, since you want the window to have a reasonable size) or maybe even simply maximizing the entire browser during browser setup?

calixteman:
Yeah, I thought about that too but I almost don't know how the bot works (I'm not a windows expert) and what are the available screen size... and I don't really understand why opening the bot through rdp and remmina can change the window size...
So I just had a better feeling in clicking rather than fixing the window size (and for these tests (print and save) I don't care so much to make a "real" click on download/print buttons).
Anyway, I'll do a patch to maximize the browser at startup.

calixteman:
@timvandermeij unfortunately we're starting the browser using `defaultViewport: null`:
https://github.com/mozilla/pdf.js/blob/a10dc1cb6e24b83d1b713e3fd3dc9a44788541d1/test/test.js#L835.

timvandermeij:
Ugh, it seems like this re-appeared in #12875 and other PRs, so we'll need some solution here. Perhaps we can do this for now, with a clear comment added as to why we do this instead of `page.click()`, but it doesn't feel like the proper solution. If we do this for now, we should also try to file an upstream issue for Puppeteer since `page.click()` _should_ work according to the documentation.

calixteman:
I'm pretty sure that's because the window is too small and then the button to save goes to the `>>` menu and so it's impossible for puppeteer to make the button visible in scrolling so it isn't a problem puppeteer for me but maybe with remmina: I used it friday to restart the bot and problems with integration test came after.

timvandermeij:
Yes, if that's the case then your explanation makes sense. If the window is too small, the save button is indeed pushed into the secondary toolbar. Since we didn't really have this before, let's indeed keep using `page.click()` as intended by Puppeteer, but we just need to be careful with bot restarts then. If this really becomes annoying, we can always reconsider something like this PR.

PR #12820, opened 2021-01-06
Don't dispatch "pageclose" events if a "pageopen" wasn't dispatched for the page (PR 12747 follow-up)

Snuffleupagus:
Given that "pageopen" events are not guaranteed to occur, if the page becomes inactive *before* it finishes rendering, we should probably also avoid dispatching a "pageclose" event in that case to avoid confusing/inconsistent state in any event handlers.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -654,7 +654,7 @@ class BaseViewer {
     this._pagesCapability = createPromiseCapability();
     this._scrollMode = ScrollMode.VERTICAL;
     this._spreadMode = SpreadMode.NONE;
-    this._pageOpenPendingSet = null;
+    this._pageOpenPendingSet?.clear();
 
     if (this._onBeforeDraw) {
       this.eventBus._off("pagerender", this._onBeforeDraw);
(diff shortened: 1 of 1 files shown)


web/base_viewer.js
```diff
@@ -1517,6 +1517,9 @@ class BaseViewer {
     const { eventBus } = this;
 
     const dispatchPageClose = pageNumber => {
+      if (this._pageOpenPendingSet?.has(pageNumber)) {
```
calixteman, line 1520: The patch looks good.
An unrelated question: why do you lazily initialize `_pageOpenPendingSet` ?
Setting it in `_initializeScriptingEvent` will avoid few checks but maybe it's pretty rare to have a non null `_pageOpenPendingSet`.

Snuffleupagus: > but maybe it's pretty rare to have a non null `_pageOpenPendingSet`.

That was my idea when writing the code, since that's what I observed in practice early during development.

However, to initialize things correctly in the general case, the final code also dispatches a "pageopen" event on the pre-existing "pagesinit" event and with those changes `_pageOpenPendingSet` is now always created. While the extra checks probably don't matter much, I'll fix this is a separate commit; thank you!

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12822, opened 2021-01-07
Improve the `PDFViewerApplication._contentLength` handling in the viewer, related mostly to scripting

Snuffleupagus:
Ensure that `PDFViewerApplication._contentLength` is always updated with the *correct* length, as returned by `PDFDocumentProxy.getDownloadInfo`, and only let the `PDFViewerApplication._initializeMetadata` method overwrite if it's not already been set.

Finally, in `PDFViewerApplication._initializeJavaScript`, the fallback `_contentLength` handling is now moved to just after the fallback `documentInfo` handling, such that all the fallback code is in one place within the method.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1242,7 +1242,8 @@ const PDFViewerApplication = {
   load(pdfDocument) {
     this.pdfDocument = pdfDocument;
 
-    pdfDocument.getDownloadInfo().then(() => {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #12825, opened 2021-01-07
An option is missing for last versions of worker-loader to use pdf.js with webpack

### LINKED ISSUE #12813: webpack instructions can't work with last worker-loader versions

fdaveine:
Fix #12813

### DIFF
--- a/examples/webpack/README.md
+++ b/examples/webpack/README.md
@@ -23,7 +23,7 @@ the worker code, and the `workerSrc` path shall be set to the latter file.
 If you are getting the `Setting up fake worker` warning, make sure you are
 importing `pdfjs-dist/webpack` which is the zero-configuration method for
 Webpack users. You will need to install
-[worker-loader](https://github.com/webpack-contrib/worker-loader) as a
+[worker-loader](https://github.com/webpack-contrib/worker-loader) (version 3.0.0 or higher is required), as a
 dependency in your project in order to use `pdfjs-dist/webpack` (configuring
 `worker-loader` is not necessary; just installing it is sufficient).
 
--- a/external/dist/webpack.js
+++ b/external/dist/webpack.js
@@ -16,7 +16,7 @@
 "use strict";
 
 var pdfjs = require("./build/pdf.js");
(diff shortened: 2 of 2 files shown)


Snuffleupagus (changes requested):
These changes are small enough, and sufficiently connected, that a *single* commit will suffice here; please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

fdaveine:
@Snuffleupagus fixed

Snuffleupagus (changes requested):
The commit message contains "work-loader" (more than once), which looks like a typo; please fix that as well.

fdaveine:
@Snuffleupagus fixed too :+1:

Snuffleupagus:
> fixed too +1

Not quite, since you *still* have the incorrect package name in your commit message :-)

fdaveine:
> Not quite, since you _still_ have the incorrect package name in your commit message :-)

Sorry I misread your previous comment, should be fixed now...

timvandermeij (approved):

timvandermeij:

PR #12828, opened 2021-01-07
[api-minor] Set font size and color for text widget annotations

### LINKED ISSUE #7367: Form field text values too large
Link to PDF file (or attach file here): [detect-form-fields.pdf](https://github.com/mozilla/pdf.js/files/287227/detect-form-fields.pdf)

Configuration:
- Web browser and its version: Firefox 46.0.1
- Operating system and its version: Ubuntu 16.04
- PDF.js version: 1.3.231
- Is an extension: No

1. Open PDF
2. Look at entered text in form sections 8a and 9a

What is the expected behavior? Text is correct size:

screenshot from 2016-05-27 12-32-04

What went wrong? Text is very large:

screenshot from 2016-05-27 12-30-58

PdfBug console output:

```
11:56:13.291 PDF ef2293a3b9f4e0accbef97ee38798a2c [1.3 Mac OS X 10.11.1 Quartz PDFContext / Amgraf OneForm Designer Plus] (PDF.js: 1.3.231)1 viewer.js:6800:7
11:56:13.292 Warning: AcroForm/XFA is not supported1 viewer.js:6824:9
11:56:13.424 Warning: Unimplemented annotation type "FreeText", falling back to base annotation1 pdf.worker.js:9407:5
11:56:13.806 Warning: fontRes not available1 pdf.worker.js:9407:5 (logged 54 times)
11:56:13.898 Warning: Error during font loading: Font Helvetica-Bold is not available1 pdf.js:368:5
```

### LINKED ISSUE #11199: Form Field doesn't have the correct font size
https://drive.google.com/open?id=1UjLxnFw2g3gn3tvNLRYLdmG6mFdFKCcr

Configuration:
- Web browser and its version: Chrome 77.0.3865.90
- Operating system and its version: Mac mojave 10.14.6
- PDF.js version: 2.3.194
- Is a browser extension: Nope

1. Make sure to set `renderInteractiveForms` to `true`
2. The read-only form field(Sample title) doesn't honor it's proper font-size

write

What went wrong?
wrong

dhufnagel:
The display layer for annotation did not consider the font size in input fields.

I added the parsing and assigning of the font size of the input field. Aside from my own issues of PDFs not beeing correctly rendered on Screen, this fixes the too large rendered text fields in issue #7367 as well as the too small rendered text field in issue #11199 .

I did not change any rendering steps, as the font size seems to be considered there. One remaining issue are fields with font size "auto" which corresponds to size 0.

Fixes #7367.
Fixes #11199.


dhufnagel:
If #12831 gets merged, i would rewrite my changes to use the defaultAppearance parser and font-size calculations for auto-size provided in that PR

dhufnagel:
I've rewritten my old commit and now use the parsed default appearance data which was introduced by @calixteman in #12831 .
As this is my first PR, maybe someone can guide me how to go on from now? Just wait till someone is going to review this PR or is there anything to do now?

src/display/annotation_layer.js
```diff
@@ -858,14 +846,37 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
...
+
+    if (this.data.textAlignment !== null) {
+      element.style.textAlign = TEXT_ALIGNMENT[this.data.textAlignment];
+    }
+
+    let font = null;
+    if (
+      fontName &&
+      fontName.name &&
+      this.data.fontRefName &&
```
timvandermeij, line 875: I think this line can be removed because of https://github.com/mozilla/pdf.js/issues/12226#issuecomment-682163009. As far as I could tell, but please check again, nothing actually sets this `fontRefName` variable, so it should be dead code. I believe `fontName.name` that is used here is the replacement for this.

src/core/annotation.js
```diff
@@ -1000,6 +1000,7 @@ class WidgetAnnotation extends Annotation {
     this._defaultAppearanceData = parseDefaultAppearance(
       data.defaultAppearance
     );
+    this.data.defaultAppearanceData = this._defaultAppearanceData;
```
timvandermeij, line 1003: You can shorten this to `data.defaultAppearanceData = ...` since `data` is a local variable reference to `this.data` here (similar to the lines above). I also think we can then remove usage of `this._defaultApperanceData` and replace it with usage of `this.data.defaultAppearanceData` so it's only tracked once in the data field instead of also as a member variable here.

Since this is an API addition this patch gets the label `[api-minor]`, and I'm only thinking if there is perhaps a shoter/better name here from an API perspective: `fontData`/`fontInformation` maybe, or is the data in default appearances not necessarily related to fonts?

timvandermeij (commented):
Looks good to me with these comments addressed, and I like how this uses the new parser.

src/display/annotation_layer.js
```diff
@@ -858,14 +846,37 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
    *
    * @private
    * @param {HTMLDivElement} element
-   * @param {Object} font
    * @memberof TextWidgetAnnotationElement
    */
-  _setTextStyle(element, font) {
-    // TODO: This duplicates some of the logic in CanvasGraphics.setFont().
+  _setTextStyle(element) {
```
timvandermeij, line 851: This PR should have a reference test attached to it so we can avoid future regressions. One of the PDF files from the issues you mentioned will suffice, or another file you may have that is smaller and shows the same issue that this PR fixes. Since this shows up when rendering forms, you can add a test similar to https://github.com/mozilla/pdf.js/commit/adf0972ca57e66202acb489b8782a3b73d510bf5#diff-bbeef90df04bff7617f7ef59e3f37cac92192c5169bc500d857bd6fb107fcabd: add the PDF file, add a gitignore entry and add a manifest entry with `"forms": true` to enable form rendering for your reference test.

src/display/annotation_layer.js
```diff
@@ -858,14 +846,37 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
...
+    const { fontName, fontSize, fontColor } = this.data.defaultAppearanceData;
     const style = element.style;
-    style.fontSize = `${this.data.fontSize}px`;
-    style.direction = this.data.fontDirection < 0 ? "rtl" : "ltr";
+
+    /**
+     *  TODO: If font size is 0, calculate the font-size based on the height and
+     *          width of the element
+     *  For now, just use the default font size
+     */
```
Snuffleupagus, line 860: This kind of formatting is *only* used for JSDoc comments, hence please use only `//` for comments like this.
Furthermore, please remember to that comments should contain *complete* sentences for improved readability, i.e. there should be a *period* at the end of every sentence.

src/display/annotation_layer.js
```diff
@@ -858,14 +846,37 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
...
+    if (fontSize) {
+      style.fontSize = `${fontSize}px`;
+    }
+
+    style.color = `rgb(${fontColor[0]}, ${fontColor[1]}, ${fontColor[2]})`;
+
+    if (this.data.textAlignment !== null) {
+      element.style.textAlign = TEXT_ALIGNMENT[this.data.textAlignment];
```
Snuffleupagus, line 868: Remove the `element.` part here.

src/display/annotation_layer.js
```diff
@@ -858,14 +846,37 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
...
+    /**
+     *  TODO: If font size is 0, calculate the font-size based on the height and
+     *          width of the element
+     *  For now, just use the default font size
+     */
+    if (fontSize) {
+      style.fontSize = `${fontSize}px`;
+    }
+
+    style.color = `rgb(${fontColor[0]}, ${fontColor[1]}, ${fontColor[2]})`;
```
Snuffleupagus, line 865: You must use `Util.makeHexColor` rather than manually doing this.

Also, the commit message should be updated to mention that you also set the colour now.

src/display/annotation_layer.js
Snuffleupagus: This whole block is dead code, and furthermore there's no way that this actually works (it looks untested):
 - First of all, the `data.fontRefName` property is never actually set.
 - Secondly, the `fontName` cannot be used to lookup a font given that we index fonts using an *internal* PDF.js-specific name rather than what exists in the `/DA` of the annotation.

All-in-all, this code-block should probably just be remove since there's no way that it works as-is (and getting it to work would require a lot more changes and testing).

Snuffleupagus (changes requested):

src/display/annotation_layer.js
dhufnagel: I agree, I reused this block from what was there. I did test it with other fonts, but it did not work as expected, so I will remove handling the font name.

src/core/annotation.js
dhufnagel: The specification mentions, that there might also be a text transformation matri inside the default appearance stream. So I am not sure if this would fit to  `fontData/fontInformation` . With the current implementation of the defaultAppearance parser, I would agree to rename this property, but on in the long term, this might not be the best choice.

src/display/annotation_layer.js
```diff
@@ -858,32 +846,25 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
...
+    const { fontSize, fontColor } = this.data.defaultAppearanceData;
     const style = element.style;
-    style.fontSize = `${this.data.fontSize}px`;
-    style.direction = this.data.fontDirection < 0 ? "rtl" : "ltr";
 
-    if (!font) {
-      return;
+    // TODO: If font size is 0, calculate the font-size based on the height and
+    //         width of the element.
+    // Not setting style.fontSize will use the default font size for now.
```
Snuffleupagus, line 858: Let's be more consistent here w.r.t. to "font size" and "font-size", and also fix the indentation above:
```suggestion
    // TODO: If the font-size is zero, calculate it based on the height and
    //       width of the element.
    // Not setting `style.fontSize` will use the default font-size for now.
```

test/test_manifest.json
```diff
@@ -5005,5 +5005,13 @@
...
       "type": "eq",
       "annotations": true
+    },
+    {
+      "id": "pr12828",
+      "file": "pdfs/pr12828.pdf",
+      "md5": "e44d364fba2f146aed04f9d9abbb0f28",
+      "rounds": 1,
+      "type": "eq",
+      "annotations": true
```
Snuffleupagus, line 5015: Is this actually sufficient/correct, since another review comment mentioned using `"forms": true`?

Snuffleupagus (changes requested):

Snuffleupagus:
See also https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

Snuffleupagus:
Please also follow https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

test/test_manifest.json
```diff
@@ -5005,5 +5005,13 @@
       "rounds": 1,
       "type": "eq",
       "annotations": true
+    },
+    {
+      "id": "pr12828",
```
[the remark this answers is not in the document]
dhufnagel: I choosed pr12828 as I created this pdf for this specific PR wich also includes test for font color. Should I rename it according to one of the two addressed issues?

test/test_manifest.json
dhufnagel: You're right.

src/display/annotation_layer.js
dhufnagel: Thanks!

test/test_manifest.json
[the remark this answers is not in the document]
timvandermeij: I'd say this is fine given that it does a bit more and solves two issues. We also already have this format in the manifest, for example `pr7352` and `pr4922` (and more).

src/core/annotation.js
timvandermeij: Let's indeed keep it like this; thanks.

timvandermeij:
Looks good to me once the merge conflict is resolved (happened because we just merged #12886 that also touched the same code, but simplified it). We'll run a preview and the tests after this is done.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

PR #12830, opened 2021-01-07
JS -- Fix doc.getField and add missing field methods

calixteman:
- getField("foo") was wrongly returning a field named "foobar";
 - field object had few missing unimplemented methods

### DIFF
--- a/src/scripting_api/doc.js
+++ b/src/scripting_api/doc.js
@@ -828,13 +828,48 @@ class Doc extends PDFObject {
       return searchedField;
     }
 
+    const parts = cName.split("#");
+    let childIndex = NaN;
+    if (parts.length === 2) {
+      childIndex = Math.floor(parseFloat(parts[1]));
+      cName = parts[0];
+    }
+
     for (const [name, field] of this._fields.entries()) {
-      if (name.includes(cName)) {
+      if (name.endsWith(cName)) {
+        if (!isNaN(childIndex)) {
+          const children = this._getChildren(name);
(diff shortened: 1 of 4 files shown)


test/unit/scripting_spec.js
```diff
@@ -159,6 +159,64 @@ describe("Scripting", function () {
         done.fail(ex);
       }
     });
+
+    it("should get field using a path", async function (done) {
```
brendandahl, line 163: Why async and done?

src/scripting_api/doc.js
```diff
@@ -828,15 +828,50 @@ class Doc extends PDFObject {
       return searchedField;
     }
 
+    const parts = cName.split("#");
```
brendandahl, line 831: It looks like fields should be a tree according to the spec. Why not store the fields as a tree and walk it to find the field?

brendandahl (changes requested):

test/unit/scripting_spec.js
calixteman: Yeah it's useless.
Let's fixing that in a follow-up.

brendandahl: Is there a reason this needs to be done in a follow up? Seems like a good time to fix it here.

calixteman: Done here: https://github.com/mozilla/pdf.js/pull/12830/commits/ccef734ebb80e50b0ed15ddc0f4c26cc1619f1ad

PR #12831, opened 2021-01-07
Add a parser to get font data from the default appearance

calixteman:
- pdfium & poppler use a special parser too to get these info.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -35,6 +35,10 @@ import {
 } from "../shared/util.js";
 import { Catalog, FileSpec, ObjectLoader } from "./obj.js";
 import { collectActions, getInheritableProperty } from "./core_utils.js";
+import {
+  createDefaultAppearance,
+  parseDefaultAppearance,
+} from "./default_appearance.js";
 import { Dict, isDict, isName, isRef, isStream, Name } from "./primitives.js";
 import { ColorSpace } from "./colorspace.js";
 import { OperatorList } from "./operator_list.js";
@@ -993,6 +997,10 @@ class WidgetAnnotation extends Annotation {
     data.defaultAppearance = isString(defaultAppearance)
       ? defaultAppearance
       : "";
+    this._defaultAppearanceData = parseDefaultAppearance(
+      data.defaultAppearance
+    );
+
     const fieldType = getInheritableProperty({ dict, key: "FT" });
     data.fieldType = isName(fieldType) ? fieldType.name : null;
 
@@ -1288,12 +1296,14 @@ class WidgetAnnotation extends Annotation {
       // Doing so prevents exceptions and allows saving/printing
       // the file as expected.
       this.data.defaultAppearance = "/Helvetica 0 Tf 0 g";
+      this._defaultAppearanceData = parseDefaultAppearance(
+        this.data.defaultAppearance
+      );
     }
 
-    const fontInfo = await this._getFontData(evaluator, task);
-    const [font, fontName] = fontInfo;
-    const fontSize = this._computeFontSize(...fontInfo, totalHeight);
-    this._fontName = fontName;
+    const font = await this._getFontData(evaluator, task);
+    const fontSize = this._computeFontSize(font, totalHeight);
+    this._fontName = this._defaultAppearanceData.fontName.name;
 
     let descent = font.descent;
     if (isNaN(descent)) {
@@ -1364,27 +1374,30 @@ class WidgetAnnotation extends Annotation {
   async _getFontData(evaluator, task) {
     const operatorList = new OperatorList();
     const initialState = {
-      fontSize: 0,
       font: null,
-      fontName: null,
       clone() {
         return this;
       },
     };
 
-    await evaluator.getOperatorList({
-      stream: new StringStream(this.data.defaultAppearance),
-      task,
-      resources: this._fieldResources.mergedResources,
+    const { fontName, fontSize } = this._defaultAppearanceData;
+    await evaluator.handleSetFont(
+      this._fieldResources.mergedResources,
+      [fontName, fontSize],
+      /* fontRef = */ null,
       operatorList,
+      task,
       initialState,
-    });
+      /* fallbackFontDict = */ null
+    );
 
-    return [initialState.font, initialState.fontName, initialState.fontSize];
+    return initialState.font;
   }
 
-  _computeFontSize(font, fontName, fontSize, height) {
-    if (fontSize === null || fontSize === 0) {
+  _computeFontSize(font, height) {
+    let fontSize = this._defaultAppearanceData.fontSize;
+    if (!fontSize) {
+      const { fontColor, fontName } = this._defaultAppearanceData;
       let capHeight;
       if (font.capHeight) {
         capHeight = font.capHeight;
@@ -1403,15 +1416,11 @@ class WidgetAnnotation extends Annotation {
       // 1.5 * capHeight * fontSize seems to be a good value for lineHeight
       fontSize = Math.max(1, Math.floor(height / (1.5 * capHeight)));
 
-      let fontRegex = new RegExp(`/${fontName}\\s+[0-9.]+\\s+Tf`);
-      if (this.data.defaultAppearance.search(fontRegex) === -1) {
-        // The font size is missing
-        fontRegex = new RegExp(`/${fontName}\\s+Tf`);
-      }
-      this.data.defaultAppearance = this.data.defaultAppearance.replace(
-        fontRegex,
-        `/${fontName} ${fontSize} Tf`
-      );
+      this.data.defaultAppearance = createDefaultAppearance({
+        fontSize,
+        fontName,
+        fontColor,
+      });
     }
     return fontSize;
   }
--- a/src/core/default_appearance.js
+++ b/src/core/default_appearance.js
@@ -14,10 +14,10 @@
  */
 
 import { isName, Name } from "./primitives.js";
+import { OPS, warn } from "../shared/util.js";
 import { ColorSpace } from "./colorspace.js";
 import { escapePDFName } from "./core_utils.js";
 import { EvaluatorPreprocessor } from "./evaluator.js";
-import { OPS } from "../shared/util.js";
 import { StringStream } from "./stream.js";
 
 class DefaultAppearanceEvaluator extends EvaluatorPreprocessor {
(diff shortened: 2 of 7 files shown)


src/core/defaultappearance.js
```diff
@@ -0,0 +1,162 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { isName, Name } from "./primitives.js";
+import { escapePDFName } from "./core_utils.js";
+
+// Parse DA to extract font and color informations
+function parseDefaultAppearance(str) {
```
Snuffleupagus, line 20: Given just how *simplistic* this algorithm is, compared to the existing `PartialEvaluator`/`Parser`/`Lexer`, isn't there a very real risk that it won't cope with all the kinds of weirdness seem in real-world PDF document!?

Can't you *at least* re-use the `Parser`/`Lexer` instead here, rather than "re-inventing the wheel" so to speak, since I believe that should be slightly safer (and with less risk of future bugs) overall?

src/core/defaultappearance.js
```diff
@@ -0,0 +1,162 @@
...
+    switch (token.charAt(0)) {
+      case "/":
+        // we've a name
+        const name = token
+          .slice(1)
+          .replace(/#([0-9a-fA-F]{2})/, (match, n) =>
+            String.fromCharCode(n.parseInt(16))
+          );
+        stack.push(Name.get(name));
+        break;
```
Snuffleupagus, line 46: Are you sure that this correctly handles all cases of malformed `Name`-data?
Note the parsing/validation, which is another argument for re-using that functionality here, in e.g. https://github.com/mozilla/pdf.js/blob/35845d1bbb7d6b581e48ddd8d8f1a96ab9a931d2/src/core/parser.js#L1075-L1120

src/core/defaultappearance.js
```diff
@@ -0,0 +1,162 @@
...
+    color: null,
+  };
+
+  if (!str) {
+    return result;
+  }
+
+  const stack = [];
+
+  for (const token of str.split(/\s+/)) {
```
Snuffleupagus, line 33: In practice it's not *always* guaranteed that operators are properly separated by whitespace, since bad PDF generators sometimes "glue" them together; see examples in these unit-tests:
https://github.com/mozilla/pdf.js/blob/master/test/unit/evaluator_spec.js

Again, re-using the `Parser`/`Lexer´ might help here!

src/core/annotation.js
```diff
@@ -1284,10 +1292,9 @@ class WidgetAnnotation extends Annotation {
       this.data.defaultAppearance = "/Helvetica 0 Tf 0 g";
     }
 
-    const fontInfo = await this._getFontData(evaluator, task);
-    const [font, fontName] = fontInfo;
-    const fontSize = this._computeFontSize(...fontInfo, totalHeight);
-    this._fontName = fontName;
+    const font = await this._getFontData(evaluator, task);
+    const fontSize = this._computeFontSize(font, totalHeight);
+    this._fontName = this._defaultAppearanceData.fontName.name;
```
Snuffleupagus, line 1297: Given that `fontName` is null by default, this can throw!

Snuffleupagus (changes requested):
There appears to be one regression, in the `issue12750` test-case.
Furthermore, I'd strongly recomment making the overall parsing a lot more robust by utilizing the `Parser`/`Lexer` instead; please see inline comments.

src/core/annotation.js
```diff
@@ -34,6 +34,10 @@ import {
 import { Catalog, FileSpec, ObjectLoader } from "./obj.js";
 import { collectActions, getInheritableProperty } from "./core_utils.js";
 import { Dict, isDict, isName, isRef, isStream, Name } from "./primitives.js";
+import {
+  dumpDefaultAppearance,
+  parseDefaultAppearance,
+} from "./defaultappearance.js";
```
timvandermeij, line 40: Aside from the already added comments that I agree with, we should call this file `default_appearance.js`.

src/core/defaultappearance.js
Snuffleupagus: Thinking about this more, you should probably just use the [`EvaluatorPreprocessor`](https://github.com/mozilla/pdf.js/blob/f224a0eb9f5cb4798d9505f8b618934e12b19885/src/core/evaluator.js#L3856-L4127) for this parsing as well. That way you'd use the *same* exact parsing as the "regular" evaluator uses, and thus benefit from all of the existing handling of various edge-cases (you'd simply need to `export` it such that it can be used from this file as well).

src/core/default_appearance.js
```diff
@@ -0,0 +1,82 @@
...
+      fontName: Name.get(""),
+      fontColor: new Uint8ClampedArray([0, 0, 0]) /* black */,
+    };
+
+    while (this.read(operation)) {
+      const { fn, args } = operation;
+      switch (fn | 0) {
+        case OPS.setFont:
+          result.fontSize = args[1];
+          result.fontName = args[0];
```
Snuffleupagus, line 45: Can we re-order these lines, since it feels slightly more logical accessing the array elements in order.

src/core/default_appearance.js
```diff
@@ -0,0 +1,82 @@
...
+import { escapePDFName } from "./core_utils.js";
+import { EvaluatorPreprocessor } from "./evaluator.js";
+import { Name } from "./primitives.js";
+import { OPS } from "../shared/util.js";
+import { StringStream } from "./stream.js";
+
+class DefaultAppearanceEvaluator extends EvaluatorPreprocessor {
+  constructor(str) {
+    const stream = new StringStream(str);
+    super(stream, null);
```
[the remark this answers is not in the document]
Snuffleupagus: If there's ever a case where a defaultAppearance contains save/restore operators, won't there be errors thrown?

The simplest way of addressing that, I think, is probably to just given the relevant classes reasonable default values: perhaps something like this:
 1. Change https://github.com/mozilla/pdf.js/blob/631bada0df53acd221d49e8cea83a1700abb7c05/src/core/evaluator.js#L3988 to
    ```javascript
    constructor(stream, xref, stateManager = new StateManager()) { 
    ```
 2. And change https://github.com/mozilla/pdf.js/blob/631bada0df53acd221d49e8cea83a1700abb7c05/src/core/evaluator.js#L3717 to
    ```javascript
    constructor(initialState = new EvalState()) { 
    ```

src/core/default_appearance.js
Snuffleupagus: Also, given that the rest of the font-handling in `src/core/evaluator.js` normally takes care of (indirectly) validating these parameters, you probably want to check here that they are of the expected types (`Name` respectively a positive number).

test/unit/jasmine-boot.js
```diff
@@ -60,6 +60,7 @@ async function initializePDFJS(callback) {
       "pdfjs-test/unit/core_utils_spec.js",
       "pdfjs-test/unit/crypto_spec.js",
       "pdfjs-test/unit/custom_spec.js",
+      "pdfjs-test/unit/default_appearance_spec.js",
```
Snuffleupagus, line 63: Please add this to https://github.com/mozilla/pdf.js/blob/master/test/unit/clitests.json as well.

Snuffleupagus (changes requested):
This looks *much* nicer already, however I've got few more suggestions here.

src/core/default_appearance.js
```diff
@@ -0,0 +1,91 @@
...
+          break;
+        case OPS.setFillGray:
+          ColorSpace.singletons.gray.getRgbItem(args, 0, result.fontColor, 0);
+          break;
+        case OPS.setFillColorSpace:
+          ColorSpace.singletons.cmyk.getRgbItem(args, 0, result.fontColor, 0);
+          break;
+      }
+      args.length = 0;
+    }
```
Snuffleupagus, line 66: Given that the `Parser`/`Lexer` will throw on invalid data, which would cause Annotation-parsing to fail, please wrap all of this code in a try-catch block as below:

```suggestion
    try {
      while (this.read(operation)) {
        if (this.stateManager.stateStack.length !== 0) {
          // Don't get info in save/restore sections.
          continue;
        }
        const { fn, args } = operation;
        switch (fn | 0) {
          case OPS.setFont:
            const [fontName, fontSize] = args;
            if (isName(fontName)) {
              result.fontName = fontName;
            }
            if (typeof fontSize === "number" && fontSize > 0) {
              result.fontSize = fontSize;
            }
            break;
          case OPS.setFillRGBColor:
            ColorSpace.singletons.rgb.getRgbItem(args, 0, result.fontColor, 0);
            break;
          case OPS.setFillGray:
            ColorSpace.singletons.gray.getRgbItem(args, 0, result.fontColor, 0);
            break;
          case OPS.setFillColorSpace:
            ColorSpace.singletons.cmyk.getRgbItem(args, 0, result.fontColor, 0);
            break;
        }
        args.length = 0;
      }
    } catch (reason) {
      warn(`parseDefaultAppearance - ignoring errors: "${reason}".`);
    }
```

Snuffleupagus (commented):
We should probably get the Windows-bot restarted, such that we can actually run all tests *before* landing this.

This looks good to me, with the final comment addressed and all test run.

src/core/default_appearance.js
```diff
@@ -0,0 +1,95 @@
...
+      fontName: Name.get(""),
+      fontColor: new Uint8ClampedArray([0, 0, 0]) /* black */,
+    };
+
+    try {
+      while (this.read(operation)) {
+        if (this.stateManager.stateStack.length !== 0) {
+          // Don't get info in save/restore sections.
+          continue;
+        }
```
Snuffleupagus, line 44: Sorry, one more thing here: please also set `args.length = 0;` in this branch since *if* this branch is hit (however unlikely that may be), we need to reset the `args` to prevent issues.

calixteman: oh correct, thx

calixteman:
> We should probably get the Windows-bot restarted, such that we can actually run all tests _before_ landing this.
> 
I restarted it (error was: `shell.js: internal error
Error: EPERM: operation not permitted, unlink 'C:\pdfjs\botio-files-pdfjs\private\165fb7f6ccb2470/node_modules/puppeteer/.local-firefox/win64-86.0a1/firefox/browser/omni.ja'`)

> This looks good to me, with the final comment addressed and all test run.

calixteman:
The integration test failure is unrelated to this patch (maybe because I opened a remote desktop to restart the bot).

Snuffleupagus (approved):
r=me, with the commits squashed :-)

PR #12834, opened 2021-01-08
Try to fix TypeScript definitions for the es5-build in pdfjs-dist (issue 12872)

### LINKED ISSUE #12827: Typescript types definition missing for es5 build
The current type definition did only work if pdfjs is directly imported, but not if the es5 build is used (when used with node).

I had to add a d.ts file to make it work. But I think this should be directly included in the release.

```typescript
declare module "pdfjs-dist/es5/build/pdf.js" {
  export * from "pdfjs-dist";
}
```

Snuffleupagus:
Given that we're already using the `external/dist/` folder for things we simply want to copy to pdfjs-dist during building, this patch *should* hopefully work since it's based on the suggestion in https://github.com/mozilla/pdf.js/issues/12827#issuecomment-756471707.

As long as this only requires a *single/small* file, to fix the TypeScript definitions in es5-builds, this solution seem acceptable as far as I'm concerned. (Although, please note that I don't know enough about TypeScript to actually test the patch.)

Fixes #12827

### DIFF
--- a/external/dist/es5/build/pdf.d.ts
+++ b/external/dist/es5/build/pdf.d.ts
@@ -1,0 +1,1 @@
+export * from "pdfjs-dist";


timvandermeij (approved):

timvandermeij:
Let's do this; thanks!

simonhaenisch:
Somehow this file is not part of `pdfjs-dist`, or at least not anymore:

```
❯ ls node_modules/pdfjs-dist/es5/build
pdf.js
pdf.js.map
pdf.min.js
pdf.worker.entry.js
pdf.worker.js
pdf.worker.js.map
pdf.worker.min.js
```

(see https://github.com/mozilla/pdfjs-dist/tree/7bfceb110b9b27a09cc3d69f4907c59fd5561f0c/es5/build)

I couldn't figure out where the build script for `pdfjs-dist` lives but is it possible that it's just forgetting to include this file/file extension in the output?

---

FWIW, re the original issue #12827: there's also the official `browser` field in `package.json` (see https://docs.npmjs.com/cli/v7/configuring-npm/package-json#browser), so it could be

```json
{
  "main": "es5/build/pdf.js",
  "browser": "build/pdf.js",
  "types": "types/pdf.d.ts",
  "//": "..."
}
```

However I'm not totally sure about the implications of this, i. e. how well `browser` is supported by other tools (should be pretty good support though).

simonhaenisch:
Hm actually never mind I just figured out that the latest stable release (2.6.347) was from 5 months ago... the file is there now that I installed `pdfjs-dist@next` (2.7.570).

PR #12837, opened 2021-01-08
Remove unncessary `CanvasFactory`/`CMapReaderFactory`/`FileReaderFactory` duplication in unit-tests

Snuffleupagus:
Given that the API will now, after PR #12039, automatically pick the correct factories to use depending on the environment (browser vs. Node.js), we can utilize that in the unit-tests as well. This way we don't have to manually repeat the same initialization code in *multiple* unit-tests.
*Note:* The *official* PDF.js API is defined in `src/pdf.js`, hence the new exports in `src/display/api.js` will not affect that.

Also, updates the unit-test `FileReaderFactory` helpers similarily.

*Drive-by change:* Fix the `CMapReaderFactory` usage in the annotation unit-tests, since the cache should only contain raw data and not a Promise. While this obviously works as-is, having unit-tests that "abuse" the intended data format can easily lead to unnecessary failures if changes are made to the relevant `src/core/` code.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -3060,4 +3060,6 @@ export {
   setPDFNetworkStreamFactory,
   version,
   build,
+  DefaultCanvasFactory,
+  DefaultCMapReaderFactory,
 };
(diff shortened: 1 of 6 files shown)


timvandermeij (approved):

timvandermeij:
Nice deduplication!

Issue #12619, opened 2020-11-13
Doc.info in scripting API being currently implemented is not spec-compliant


AdobeScripter:
Don't get me wrong, it's great that @calixteman wanted to implement that — JS should definitely be supported! I don't want to come off as rude or attacking.

For starters, JS reference specified in the [commit comment](https://github.com/mozilla/pdf.js/commit/8de98079ca05f8f7001ee077771055a57dc0e948) is outdated. It's version 8.1 from 2007… We're more than decade past that now. We should be implementing [Acrobat DC SDK](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/AcrobatDC_js_api_reference.pdf) ([here](https://help.adobe.com/en_US/acrobat/acrobat_dc_sdk/2015/HTMLHelp/#t=Acro12_MasterBook%2FJS_API_AcroJS%2FJavaScript_API.htm) in a nice HTML version).

Other than that, the Doc object now is only implemented with `info` property with standard fields (`Title`, `Author`, `Subject`, `Keywords`, `Creator`, `Producer`, `CreationDate`, `ModDate` and `Trapped`, case insensitive). But [according to the spec](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/AcrobatDC_js_api_reference.pdf#page=243) (even [in that old version](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf#page=243)), this object can contain other, non-standard properties, that can be set both in Acrobat interface and in JS (which are case-sensitive then).

Acrobat 9.0 introduced another field to that property, `Authors`, containing semicolon-delimited list of authors.

There are probably other problems with that code due to the outdated spec @calixteman was apparently working on.

calixteman:
@AdobeScripter thanks for pointing out the correct specifications to use.
Anyway, we focus on having conformance with Adobe Reader so for now `doc.info` will stay read-only, but feel free to file a bug to ask for this feature.

AdobeScripter:
Even if it is read-only, the non-standard fields should be read from the PDF file and provided in that property. And there is still the `Authors` field that is missing.

AdobeScripter:
I'm sorry to ask here, since it's not completely on the topic, but on the other hand I don't really want to create a new issue just for this question, so I'll just go with it.

When do you plan to release first version (embedded in Firefox) to support JavaScript?

timvandermeij:
[...]

PR #12838, opened 2021-01-08
[api-minor] Change the "dc:creator" Metadata field to an Array

calixteman:
- add scripting support for doc.info.authors
- doc.info.metadata is the raw string with xml code
- aims to fix issue #12619


Snuffleupagus (changes requested):
This is a breaking API-change, as evident by the updated unit-tests; please note that it's *not* correct to just change an existing API in this way!

calixteman:
What's the correct way to break an api ?
Do you prefer I let `dc:creator` as it worked and so I can add a new property `authors` for example ?
Anyway according to xmp spec it seems that `dc:creator` is expected to be an array.

Snuffleupagus:
> Anyway according to xmp spec it seems that `dc:creator` is expected to be an array.

Can you please link relevant section of the XMP specification here?
If we're *explicitly* breaking an existing specification, I suppose that changes things slightly as far as I'm concerned.

Given that what we currently do is apparently wrong, the question here I suppose is how often we expect that the Metadata is used by third-party users and how badly we'd break them if we simply made the change?

Thinking about this a bit, my hunch would be that the Metadata probably isn't depended upon a whole lot in practice; hence we might be OK with this *small* breaking changing if we clearly label it as such (so that it's visible in the release notes).
I.e. you should update the PR title, and commit message, to *start* with something like:
`[api-minor] Change the "dc:creator" Metadata field to an Array; add scripting support for doc.info.authors`

/cc @timvandermeij What's your opinion, can we change the existing "dc:creator" Metadata field in this way?

src/display/api.js
```diff
@@ -2705,6 +2705,7 @@ class WorkerTransport {
       .then(results => {
         return {
           info: results[0],
+          rawMetadata: results[1],
```
Snuffleupagus, line 2708: Can you please remove this property, and instead do something like this:
 1. Add `this._data = data;` after line https://github.com/mozilla/pdf.js/blob/c0a6d6cd212436a2ed4e796815fc11d04d94c638/src/display/metadata.js#L34
 2. Add a new method, e.g.
    ```javascript
    getRaw() {
      return this._data;
    }
    ```
    after line https://github.com/mozilla/pdf.js/blob/c0a6d6cd212436a2ed4e796815fc11d04d94c638/src/display/metadata.js#L126

src/display/metadata.js
```diff
@@ -103,10 +103,31 @@ class Metadata {
       }
 
       for (let j = 0, jj = desc.childNodes.length; j < jj; j++) {
-        if (desc.childNodes[j].nodeName.toLowerCase() !== "#text") {
-          const entry = desc.childNodes[j];
+        const entry = desc.childNodes[j];
+        if (entry.nodeName.toLowerCase() !== "#text") {
```
Snuffleupagus, line 107: Given the overall indentation level and the size of the code below, things would be slightly more readable with this condition inverted like so:
```suggestion
        if (entry.nodeName.toLowerCase() === "#text") {
          continue;
        }
```

web/app.js
```diff
@@ -1765,6 +1767,7 @@ const PDFViewerApplication = {
       return; // The document was closed while the metadata resolved.
     }
     this.documentInfo = info;
+    this.rawMetadata = rawMetadata;
```
Snuffleupagus, line 1770: With the suggested changes in `src/display/api.js` and `src/display/metadata.js`, this is no longer needed.

web/app.js
```diff
@@ -1655,7 +1655,8 @@ const PDFViewerApplication = {
           baseURL: this.baseUrl,
           filesize: this._contentLength,
           filename: this._docFilename,
-          metadata: this.metadata,
+          metadata: this.rawMetadata,
```
Snuffleupagus, line 1658: With the suggested changes in `src/display/api.js` and `src/display/metadata.js`, this needs to be
```suggestion
          metadata: this.metadata?.getRaw(),
```

calixteman:
> > Anyway according to xmp spec it seems that `dc:creator` is expected to be an array.
> 
> Can you please link relevant section of the XMP specification here?
https://wwwimages2.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart1.pdf#page=33
According to this doc, `dc:creator` should be a `Seq` but I got a `Bag` using Acrobat DC to generate the pdf file in this patch. 

> If we're _explicitly_ breaking an existing specification, I suppose that changes things slightly as far as I'm concerned.
> 
> Given that what we currently do is apparently wrong, the question here I suppose is how often we expect that the Metadata is used by third-party users and how badly we'd break them if we simply made the change?
> 
> Thinking about this a bit, my hunch would be that the Metadata probably isn't depended upon a whole lot in practice; hence we might be OK with this _small_ breaking changing if we clearly label it as such (so that it's visible in the release notes).
> I.e. you should update the PR title, and commit message, to _start_ with something like:
> `[api-minor] Change the "dc:creator" Metadata field to an Array; add scripting support for doc.info.authors`
+1
 
> /cc @timvandermeij What's your opinion, can we change the existing "dc:creator" Metadata field in this way?

web/app.js
```diff
@@ -1655,7 +1655,8 @@ const PDFViewerApplication = {
...
+          authors: this.metadata.get("authors") || null,
```
Snuffleupagus, line 1659: Note that `this.metadata` can be null, hence this can throw!
Furthermore, the `get` method already fallback to returning `null`.

All in all, this line should read
```suggestion
          authors: this.metadata?.get("authors"),
```

Snuffleupagus (changes requested):
We might be OK with changing the format of "dc:creator", see https://github.com/mozilla/pdf.js/pull/12838#issuecomment-757134155

However, I've also added a couple of comments on the overall implementation here.

timvandermeij:
If I'm reading https://wwwimages2.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart1.pdf#page=33 correctly, `dc:creator` should indeed be a list (even though the name implies otherwise, since I'd have expected `dc:creators` in that case, but it's not the first time this happens in the spec), according to the "XMP usage is a list of creators" part. While it's indeed an API change, the behavior before was incorrect, so it's more like a bugfix to me, hence I'm OK with this change but only if it's clearly indicated as `[api-minor]`.

src/display/metadata.js
```diff
@@ -79,40 +80,72 @@ class Metadata {
...
       rdf = rdf.firstChild;
-      while (rdf && rdf.nodeName.toLowerCase() !== "rdf:rdf") {
+      while (rdf && rdf.nodeName !== "rdf:rdf") {
         rdf = rdf.nextSibling;
       }
     }
 
-    const nodeName = rdf ? rdf.nodeName.toLowerCase() : null;
+    const nodeName = rdf ? rdf.nodeName : null;
     if (!rdf || nodeName !== "rdf:rdf" || !rdf.hasChildNodes()) {
```
Snuffleupagus, line 123: With the lower-casing now happening in the XML-parser, you can inline the `nodeName` check here instead:
```suggestion
    if (!rdf || rdf.nodeName !== "rdf:rdf" || !rdf.hasChildNodes()) {
```

test/test_manifest.json
```diff
@@ -3213,6 +3213,12 @@
       "rounds": 1,
       "type": "eq"
     },
+    {  "id": "js-authors",
+       "file": "pdfs/js-authors.pdf",
+       "md5": "e892f290ae209286a29d3c17dfeab226",
+       "rounds": 1,
+       "type": "eq"
+    },
```
Snuffleupagus, line 3221: How much value does this `eq` test bring, since the functionality that you're adding here doesn't really show up in a reference-test as far as I can tell?

Snuffleupagus (approved):
With the final comments addressed, this looks good to me :-)

PR #12845, opened 2021-01-10
Remove unnecessary `dir`-dependent CSS rules, and simplify som findbar-button rules (PR 11077 follow-up)

Snuffleupagus:
With the updated default viewer UI, some `dir`-dependent CSS rules are now redundant since *identical* rules are being specified for both LTR and RTL mode; after PR #12807 landed I've found even more of these cases.

Note in particular that the findbar-button rules can be simplified quite a bit, since there's a fair amount of unnecessary duplication in the CSS.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -367,15 +367,6 @@ html[dir="rtl"] #sidebarResizer {
   position: relative;
   height: 32px;
   background-color: var(--toolbar-bg-color);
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for simplifying this even more!

PR #12846, opened 2021-01-10
Enable the Stylelint `length-zero-no-unit` rule

Snuffleupagus:
Note that these changes were done automatically, using `gulp lint --fix`.
With this rule, we'll thus enforce a *consistent* formatting of zero-lengths in our CSS files.

Please find additional details about the Stylelint rule at https://stylelint.io/user-guide/rules/length-zero-no-unit

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -10,5 +10,8 @@
 
   "rules": {
     "block-no-empty": true,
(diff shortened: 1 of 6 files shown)


timvandermeij (approved):

timvandermeij:
Nice consistency improvement!

PR #12848, opened 2021-01-10
Correctly align annotation content in horizontal/spead scrolling modes

### LINKED ISSUE #12847: Annotation text does not wrap and overflows popup
I am unable to share the particular PDF at this time. If it is really necessary, I can get a file with an annotation, but I doubt it is a specific problem.

Configuration:
- Web browser and its version: Mozilla Firefox for Ubuntu 84.0.2 (64-bit)
- Operating system and its version: Ubuntu 20.04
- PDF.js version: unknown
- Is a browser extension: no

The text should be contained within the annotation popup like so:

image

Instead, it overflows.

image

The problem seems to be caused by the `.pdfViewer.scrollHorizontal, .spread` CSS rule in `resource://pdf.js/web/viewer.css` which specifies `white-space: nowrap;`. At least I am able to fix it by adding `white-space: normal` to the `.annotationLayer .popup` rule in the DOM inspector.

timvandermeij:
Not only was long text in popups no longer wrapped correctly, the alignment was also center instead of left (or right, depending on the locale used) for both text in popups and the other parts within the annotation's section, such as the icon.

In order to test this, I made [annotation-popup-long.pdf](https://github.com/mozilla/pdf.js/files/5792830/annotation-popup-long.pdf) that, when opened with the current viewer, will show the issue when the horizontal/spread scrolling modes are used and when opened with the new viewer will no longer have this issue. I have checked the other files in the test suite as well to make sure that this works for all of them.

Fixes #12847.

### DIFF
--- a/web/annotation_layer_builder.css
+++ b/web/annotation_layer_builder.css
@@ -15,6 +15,7 @@
 
 .annotationLayer section {
   position: absolute;
+  text-align: initial;
 }
 
 .annotationLayer .linkAnnotation > a,
@@ -165,6 +166,7 @@
   cursor: pointer;
   font: message-box;
   font-size: 9px;
+  white-space: normal;
   word-wrap: break-word;
 }
 


Snuffleupagus (approved):
Looks good to me, thank you!

Also, note that this fixes the same problem when the *spread*-modes are used, in addition to horizontal scrolling.

timvandermeij:
I have amended the commit message and PR title/text to also instead the spread modes.

PR #12861, opened 2021-01-14
Always re-measure non-embedded ArialNarrow fonts (bug 1671312, PR 12725 follow-up)

Snuffleupagus:
While PR #12725 fixed [bug 1671312](https://bugzilla.mozilla.org/show_bug.cgi?id=1671312) as reported, i.e. the "In the upper right corner "Purposes' has bad kerning."-part, it however broke other parts of the text rendering.
Note in particular the tables, e.g. on page 2 and beyond, where the glyphs are now rendered too close together. The reason for this is that the fonts in question are non-embedded ArialNarrow, which we just replace with Helvetica which obviously is not narrow. Given that the font replacement isn't a perfect fit for non-embedded ArialNarrow, we still need to re-measure the glyph widths in this case.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -1335,8 +1335,13 @@ var Font = (function FontClosure() {
       // name ArialBlack for example will be replaced by Helvetica.
       this.black = name.search(/Black/g) !== -1;
 
+      // Use 'name' instead of 'fontName' here because the original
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

/botio makeref

PR #12863, opened 2021-01-15
Remove a duplicated reference test (PR 12812 follow-up)

Snuffleupagus:
- Remove a *duplicated* reference test, see "issue12810", from the manifest.

 - Improve the spelling in a couple of comments in `src/core/canvas.js`, most notable of the word "parallelogram".

 - Update a comment, also in `src/core/canvas.js`, to actually agree with the value used to reduce confusion when reading the code.


calixteman (approved):
Thank you for the follow-up.

Issue #6127, opened 2015-06-18
Underscore too bold


automatedbugreportingfacility:
This is a regression from PR #4615. The underscore for the last 2 chars in "NUMERO D’IMMATRICULATION" is too bold.

URL: http://www.ameli.fr/fileadmin/user_upload/formulaires/S3706.pdf
1.1.215: https://i.imgur.com/up8Y33j.png
0.8.990: https://i.imgur.com/UioZcLB.png

fc1e2483b433134a538cce0acd0ccc76e335cf5f is the first bad commit
commit fc1e2483b433134a538cce0acd0ccc76e335cf5f
Author: Jonas Jenwald
Date:   Fri Apr 11 20:19:39 2014 +0200

```
Enforce minimum line widths
```

Snuffleupagus:
That looks "fun". Compared to the recent issue #6117, this is *unfortunately not* related to the fact that we fail to invalidate the `getSinglePixelWidth` cache. Rather the issue seems to be that, for that particular line, the actual `getSinglePixelWidth` method returns an "incorrect" value.

Unfortunately I've no idea how to address that :-(
Given the number of issues/bugs fixed by PR #4615, and considering how broken the rendering of some of those PDF files were before that PR, backing it out isn't really an option in my opinion.

automatedbugreportingfacility:
Agreed, I've got another regression where getSinglePixelWidth might be falling short, and it's NOT from PR #4615, but PR #4521. If you're interested in looking into that, I can file a separate issue or paste it here.

Snuffleupagus:
> If you're interested in looking into that, I can file a separate issue or paste it here.

Please open a new issue; if it turns out that the issues are in fact *the same*, we can always close one as a duplicate of the other.

timvandermeij:
Closing since the link is dead and it's also not available in the Web Archive.

Snuffleupagus:
> Closing since the link is dead and it's also not available in the Web Archive.

Since the issue still persists, I'm attaching the file here: [issue6127.pdf](https://github.com/mozilla/pdf.js/files/2321540/issue6127.pdf)

Snuffleupagus:
@calixteman It seems that PR #12812 actually regressed this even further, can you please take a look?

PR #12864, opened 2021-01-15
Enforce linewidth to 1px when at least one of scale factor is lower than 1

calixteman:
It aims to fix issue #6127.


test/test_manifest.json
```diff
@@ -4863,6 +4863,12 @@
          "107R": {
            "value": false
          },
+    {  "id": "issue6127",
+       "file": "pdfs/issue6127.pdf",
+       "md5": "65ef9572bddef5fddfaf7172f10a0df2",
+       "rounds": 1,
+       "type": "eq"
+    },
```
Snuffleupagus, line 4871: The placement of these lines caused the tests to fail, since the manifest file is not valid.

Also, it should suffice to test *only* the first page rather than the entire document; i.e. you want to add a `"lastPage": 1,` line to this block.

Snuffleupagus (changes requested):

test/test_manifest.json
calixteman: oh, I just `node test/add_test.js ...` and I didn't check that's ok.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

Issue #12868, opened 2021-01-16
Underscore too thin when the document is zoomed in


automatedbugreportingfacility:
This is a regression from #12864. The underscore is now too thin when the document is zoomed in.

URL: https://github.com/mozilla/pdf.js/files/2321540/issue6127.pdf
PDF.js 2.7.526: https://i.imgur.com/T4mIpct.png
PDF.js 2.7.510: https://i.imgur.com/nbwEpGT.png
Adobe Acrobat Reader DC 2020.013.20074: https://i.imgur.com/wDFBf3R.png

The screenshots were taken at 400% zoom.

calixteman:
The problem is the following: at 400% the transformation matrix (in the canvas: current pdf transform + zoom) is `[ 0.6728517413139343, 0, 0, -5.3331708908081055]` and the linewidth is `1.158` so when matrix is apply to linewidth we get a rectangle pixel `0.78x6.17` this is why it's replaced by a 1px linewidth.
According to the screenshot with Acrobat, I've the impress that zoom scale is not taken into account when fixing the linewidth, I mean only the transformation in the pdf is taken into account, linewidth is then fixed and finally the zoom is applied.
@Snuffleupagus, @timvandermeij is there a way to get the zoom factor when drawing in the canvas ?

calixteman:
This info is available in `viewport.transform` in `canvas.js`.

PR #12869, opened 2021-01-16
Fix zoom issue with too thin lines

calixteman:
- aims to fix issue #12868: apply zoom factor to linewidth after setting it to 1.
 - only apply 1px-width when required
 - the sign of getSinglePixelWidth is used to know if 1px-width is required

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -923,6 +923,10 @@ const CanvasGraphics = (function CanvasGraphicsClosure() {
       this.ctx.transform.apply(this.ctx, viewport.transform);
 
       this.baseTransform = this.ctx.mozCurrentTransform.slice();
+      this._combinedScaleFactor = Math.hypot(
+        this.baseTransform[0],
+        this.baseTransform[2]
+      );
 
       if (this.imageLayer) {
         this.imageLayer.beginLayout();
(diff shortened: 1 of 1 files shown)


src/display/canvas.js
```diff
@@ -922,6 +922,11 @@ const CanvasGraphics = (function CanvasGraphicsClosure() {
       }
       this.ctx.transform.apply(this.ctx, viewport.transform);
 
+      this.viewportTransform = viewport.transform;
+      // Square of viewport scale factor.
+      this.viewportSqScaleFactor =
+        viewport.transform[0] ** 2 + viewport.transform[2] ** 2;
+      this.viewportScaleFactor = Math.sqrt(this.viewportSqScaleFactor);
```
Snuffleupagus, line 929: This doesn't actually take the optional `transform` argument into account, so I'd expect that you actually want to use `this.baseTransform` in your calculations here?

Furthermore, you're not actually using most of these properties anywhere but here, hence please use local variables where appropriate!

It should be possible to simplify all of this code, after moving it down a couple of lines, to e.g. the following:
```suggestion
      this._combinedScaleFactor =
        Math.sqrt(this.baseTransform[0] ** 2 + this.baseTransform[2] ** 2);
```

Snuffleupagus (changes requested):

src/display/canvas.js
calixteman: I forgot to remove them.
Anyway, `baseTransform` is likely better than `viewport.transform`: thanks for the tip.

src/display/canvas.js
```diff
@@ -1357,22 +1361,24 @@ const CanvasGraphics = (function CanvasGraphicsClosure() {
...
           ctx.restore();
         } else {
           const lineWidth = this.getSinglePixelWidth();
-          if (lineWidth === -1) {
-            // The current transform will transform a square pixel into a
-            // parallelogram where both heights are lower than 1 and not equal.
+          if (lineWidth < 0 && -lineWidth >= this.current.lineWidth) {
+            // The current transform will transform a square pixel into
+            // a parallelogram where both heights are lower than 1 and
+            // not equal.
```
Snuffleupagus, line 1378: Nit: This looks like an unrelated change, please revert it.

src/display/canvas.js
```diff
@@ -2664,25 +2670,31 @@ const CanvasGraphics = (function CanvasGraphicsClosure() {
...
+          // In this case, we must reset the transform, set linewidth to 1
           // and then stroke.
-          this._cachedGetSinglePixelWidth = -1;
-        } else if (sqDet > Number.EPSILON ** 2) {
-          // The multiplication by the constant 1.0000001 is here to have
+          this._cachedGetSinglePixelWidth = -(
+            this._combinedScaleFactor * pixelHeight
+          );
+        } else if (absDet > Number.EPSILON) {
+          // The multiplication by the constant 1.000001 is here to have
```
Snuffleupagus, line 2695: The constant doesn't actually agree with the code now, please revert this change.

Snuffleupagus (changes requested):

PR #12874, opened 2021-01-19
Put less emphasis on "ES5"/"ES6" in the README and other documentation

Snuffleupagus:
- Don't explicitly mention "ES5" in https://mozilla.github.io/pdf.js/getting_started/#download, since that's no longer accurate. As-is, this will more than likely give users the wrong impression regarding our *actual* browser support.

 - Reword the "Getting the Code"-section of the README, since an ES6-compatible browser is no longer sufficient for development purposes given our usage of modern ECMAScript features.

 - Update the "Building PDF.js"-section of the README, to explicitly mention both the `gulp generic` and `gulp generic-es5` commands.

Hopefully this will help reduce *some* user confusion, but we should perhaps also consider changing the "-es5"-suffix in our build-scripts and build-folders. (That would however require coming up with a "better" suffix, which might not be trivial.)


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #12875, opened 2021-01-19
Replace some ternary operators with optional chaining, and nullish coalescing, in the `src/display/`-folder

Snuffleupagus:
This way, we can further reduce unnecessary code-repetition in some cases.


Snuffleupagus:
Given that PR #12874 has the *same exact* integration-test failure, and that PR only touches README-files, I don't think the failure is relevant here either.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12879, opened 2021-01-20
XFA - Add a parser for XFA files

calixteman:
- the parser is base on a class extending XMLParserBase
 - it handle xml namespaces:
   * each namespace is assocated with a builder
   * builder builds nodes belonging to the namespace
   * when a node is inserted in the parent namespace compatibility is checked (if required)
 - to avoid name collision between xml names and object properties, use Symbol.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1431,7 +1431,7 @@ function buildLib(defines, dir) {
   return merge([
     gulp.src(
       [
-        "src/{core,display,shared}/*.js",
+        "src/{core,display,shared}/**/*.js",
         "!src/shared/{cffStandardStrings,fonts_utils}.js",
         "src/{pdf,pdf.worker}.js",
       ],
--- a/src/core/xfa/builder.js
+++ b/src/core/xfa/builder.js
@@ -1,0 +1,148 @@
+/* Copyright 2021 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { $buildXFAObject, NamespaceIds } from "./namespaces.js";
+import { $cleanup, $onChild, XFAObject } from "./xfa_object.js";
+import { NamespaceSetUp } from "./setup.js";
+import { UnknownNamespace } from "./unknown.js";
+import { warn } from "../../shared/util.js";
+
+class Root extends XFAObject {
+  constructor() {
+    super(-1, "root", Object.create(null));
+    this.element = null;
+  }
+
+  [$onChild](child) {
+    this.element = child;
+  }
+}
+
+class Empty extends XFAObject {
+  constructor() {
+    super(-1, "", Object.create(null));
+  }
+
+  [$onChild](_) {}
+}
+
+class Builder {
+  constructor() {
+    this._namespaceStack = [];
+
+    // Each prefix has its own stack
+    this._namespacePrefixes = new Map();
+    this._namespaces = new Map();
+    this._nextNsId = Math.max(
+      ...Object.values(NamespaceIds).map(({ id }) => id)
+    );
+    this._currentNamespace = new UnknownNamespace(++this._nextNsId);
+  }
+
+  buildRoot() {
+    return new Root();
+  }
+
+  build({ nsPrefix, name, attributes, namespace, prefixes }) {
+    const hasNamespaceDef = namespace !== null;
+    if (hasNamespaceDef) {
+      // Define the current namespace to use.
+      this._namespaceStack.push(this._currentNamespace);
+      this._currentNamespace = this._searchNamespace(namespace);
+    }
+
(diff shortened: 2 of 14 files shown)


test/unit/jasmine-boot.js
```diff
@@ -84,6 +84,7 @@ async function initializePDFJS(callback) {
       "pdfjs-test/unit/unicode_spec.js",
       "pdfjs-test/unit/util_spec.js",
       "pdfjs-test/unit/writer_spec.js",
+      "pdfjs-test/unit/xfa_parser_spec.js",
```
Snuffleupagus, line 87: Unless these unit-tests can't for some reason be run in Node.js, please don't forget to update https://github.com/mozilla/pdf.js/blob/master/test/unit/clitests.json as well :-)

calixteman:
The objects in `config.js` or in `xdp.js` reflect what we've in the documentation, for example the `acrobat` element documentation is:
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=1298&zoom=auto,-207,679
and the corresponding class:
https://github.com/mozilla/pdf.js/pull/12879/files#diff-0fde7327b01926a0a21537877b58611fa916d5cd228e33673678bc37565db84aR26

gulpfile.js
```diff
@@ -1447,6 +1447,7 @@ function buildLib(defines, dir) {
     gulp.src(
       [
         "src/{core,display,shared}/*.js",
+        "src/core/xfa/*.js",
```
Snuffleupagus, line 1450: I haven't tested, but wouldn't it suffice to simply do:
```suggestion
        "src/{core,display,shared}/**/*.js",
```

src/core/xfa/builder.js
```diff
@@ -0,0 +1,148 @@
...
+import { UnknownNamespace } from "./unknown.js";
+import { warn } from "../../shared/util.js";
+
+class Root extends XFAObject {
+  constructor() {
+    super(-1, "root", Object.create(null));
+    this.element = null;
+  }
+
+  [$onChild](child) {
```
timvandermeij, line 28: How does this syntax work? It doesn't look like a valid function name with these special characters in it.

calixteman: `$onChild` is defined in `xfa_object` like this: `const $onChild = Symbol();`
And the way to use this symbol as a function name is to use brackets:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#symbol_wrapper_objects_as_property_keys
The only way to call this function is to have the symbol so there's no access by name (as a string).

timvandermeij: Cool, learned something new then :-)

It's not entirely clear yet as to why this is needed exactly (i.e., why access by name causes conflicts here), but that requires a more in-depth look. This syntax was just something I noticed and didn't know; thanks for explaining this!

calixteman: In the specs there is something call SOM expressions to select nodes:
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=101&zoom=auto,-207,766
These expressions seem to be used in different places.
And so in order to make easy their interpretation, each part of the expression can be searched as a own property of the object without the need to keep a flag somewhere to check if the property exists by spec or if it is an implementation detail.
And I guess it'll avoid any bad use an attacker could do with specific expressions.
So my idea was to have objects which are the exact reflection of specs in hiding implementations details and thx to that we can enjoy the object model stuff like overriding `$onChildCheck` method when a specific node can accept a node from another namespace.
So from my pov, it makes the implementation safer, simpler and reduce memory use (no need to track spec properties...).
And it's unusual I added this `$` symbols to help to see them.

src/core/xfa/xfa_object.js
```diff
@@ -0,0 +1,253 @@
...
+      builder.clean(this[$cleanup]);
+      delete this[$cleanup];
+    }
+  }
+
+  [$isTransparent]() {
+    return this.name === "";
+  }
+
+  [$dump](buf = [], level = 0) {
```
brendandahl, line 84: Is there an advantage to using our own output format vs json stringifying an object?

src/core/xfa/xfa_object.js
```diff
@@ -0,0 +1,253 @@
...
+}
+
+class Option10 extends IntegerObject {
+  constructor(nsId, name) {
+    super(nsId, name, 1, n => n === 0);
+  }
+}
+
+export {
+  $clean,
```
brendandahl, line 234: Are there going to be many more of these? I like the idea of keeping things separate using symbols, but having to export every method seems very awkward vs prefxing with a `$` and ensuring those properties aren't overwritten.

brendandahl (changes requested):

src/core/xfa/xfa_object.js
calixteman: The other advantage of having a `Symbol` is that they aren't returned by `Object.getOwnPropertyNames` and so it avoid to check if name starts with a `$` when evaluating a SOM expression:
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=86&zoom=auto,-207,138
Anyway, you're right: the drawback that we need to import symbols.

src/core/xfa/xfa_object.js
calixteman: I just discover `toJSON`: I'll update the patch with.
Thx for the tip.

src/core/xfa/xfa_object.js
brendandahl: As discussed, we'll go with symbol since it makes SOM stuff easier.

brendandahl (approved):

PR #12885, opened 2021-01-22
Simplify the `PDFFunctionFactory._localFunctionCache` initialization (PR 12034 follow-up); Fix the `gStateObj` lookup in `TranslatedFont._removeType3ColorOperators` (PR 12718 follow-up)

Snuffleupagus:
These two commits are not directly related, but given their small size I figured that one PR won't hurt since that reduces bot usage slightly.

 - Simplify the `PDFFunctionFactory._localFunctionCache` initialization (PR 12034 follow-up)

   By changing this a `shadow`ed getter, we can simply access it directly and not worry about it being initialized. I have no idea why I didn't just implement it this way in the first place.

 - Fix the `gStateObj` lookup in `TranslatedFont._removeType3ColorOperators` (PR 12718 follow-up)

   As can be seen in https://github.com/mozilla/pdf.js/blob/2cba29036180b420778bd2c91d21d71bd207c146/src/core/evaluator.js#L986 the `gStateObj` (which is actually an Array despite its name), is wrapped in Array when it's inserted into the OperatorList. Hence we obviously need to take this into account when accessing it in `TranslatedFont._removeType3ColorOperators`; this mistake happened because we don't have any test-cases for this particular code-path as far as I know.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -3683,7 +3683,7 @@ class TranslatedFont {
           continue;
 
         case OPS.setGState:
-          const gStateObj = operatorList.argsArray[i];
+          const [gStateObj] = operatorList.argsArray[i];
           let j = 0,
             jj = gStateObj.length;
           while (j < jj) {
--- a/src/core/function.js
+++ b/src/core/function.js
@@ -20,6 +20,7 @@ import {
   info,
   isBool,
   IsEvalSupportedCached,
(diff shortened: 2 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

/botio makeref (since it looks like some browsers got updated)

PR #12886, opened 2021-01-22
Use `_defaultAppearanceData` directly in `WidgetAnnotation._getSaveFieldResources` (PR 12831 follow-up)

Snuffleupagus:
With the changes in PR #12831, it's no longer necessary to keep track of the `fontName`-string separately since it's available through the `_defaultAppearanceData`-property as well.


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #12887, opened 2021-01-22
Keep logical assignment operators in `MOZCENTRAL`/`TESTING`-builds

Snuffleupagus:
Given how recently logical assignment operators support were added to browsers, we're intentionally translating them even in `SKIP_BABEL = true` builds for the time being.
However, in either `MOZCENTRAL` or `TESTING` builds it should be OK to simply leave logical assignment operators intact.


timvandermeij (approved):

timvandermeij:
Looks like a good idea!

PR #12890, opened 2021-01-22
Enable ESLint rules that no longer need to be disabled on a directory/file-basis

Snuffleupagus:
Given that browsers/environments without native support for both arrow functions and object shorthand properties are no longer supported in PDF.js, please refer to the compatibility information below, we can now enable a fair number of ESLint rules and also simplify/remove some `.eslintrc` files.

With the exception of the `no-alert` cases, all code changes were made automatically by using `gulp lint --fix`.

 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#browser_compatibility

### DIFF
--- a/examples/.eslintrc
+++ b/examples/.eslintrc
@@ -3,23 +3,9 @@
     "../.eslintrc"
   ],
 
-  "parserOptions": {
-    "ecmaVersion": 6,
-  },
-
-  "env": {
-    "es6": false,
-  },
-
   "globals": {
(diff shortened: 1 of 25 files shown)


timvandermeij (approved):

timvandermeij:
Nice work!

PR #12893, opened 2021-01-23
Enable, a basic version of, the ESLint `no-use-before-define` rule

Snuffleupagus:
Given how our existing code-base is written, we simply can't (and probably don't want to) enable *all* of this ESLint rule. The reason is that we have a lot of code, especially related to functions, which is fine at runtime but might to a static analyzer *look* like a problem.
However, even in its most *basic* version this ESLint rule can help catch some outright bugs; see PR #12891.

Please find additional details about the ESLint rule at https://eslint.org/docs/rules/no-use-before-define

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -139,6 +139,11 @@
       "vars": "local",
       "args": "none",
     }],
+    "no-use-before-define": ["error", {
+      "functions": false,
+      "classes": false,
+      "variables": false,
+    }],
 
     // Node.js and CommonJS
     "no-buffer-constructor": "error",


calixteman:
Maybe we could run at least such checkers on the changed lines in a patch.

Snuffleupagus:
> Maybe we could run at least such checkers on the changed lines in a patch.

I don't this ESLint is easily able to do that. Anyway, this patch on its own is sufficient to catch the bug.

timvandermeij:
Looks good to me with a rebase onto the current master to see if the failure disappears.

Snuffleupagus:
> Looks good to me with a rebase onto the current master to see if the failure disappears.

As expected, it did :-)

timvandermeij (approved):

timvandermeij:
Thank you!

PR #12896, opened 2021-01-23
Modifiy the way to compute baseline to have a better match between canvas and text layer

calixteman:
- compute ascent as a ratio of font height
 - in case the font doesn't have ascent/descent/bbox info:
   * use TextMetrics.fontBoundingBoxAscent if available
   * use a basic heuristic to guess values in drawing char on a canvas

In showing text layer, before the patch:
Screenshot_2021-01-23 tracemonkey pdf

and after the patch:

image

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -54,6 +54,9 @@ import {
  */
 const renderTextLayer = (function renderTextLayerClosure() {
   const MAX_TEXT_DIVS_TO_RENDER = 100000;
+  const DEFAULT_FONT_SIZE = 30;
+  const DEFAULT_FONT_ASCENT = 0.8;
+  const ascentCache = new Map();
 
   const NonWhitespaceRegexp = /\S/;
 
@@ -61,7 +64,70 @@ const renderTextLayer = (function renderTextLayerClosure() {
     return !NonWhitespaceRegexp.test(str);
   }
 
-  function appendText(task, geom, styles) {
+  function getAscent(fontFamily, ctx) {
+    const cachedAscent = ascentCache.get(fontFamily);
+    if (cachedAscent) {
+      return cachedAscent;
+    }
+
+    ctx.save();
+    ctx.font = `${DEFAULT_FONT_SIZE}px ${fontFamily}`;
+    const metrics = ctx.measureText("");
+
+    // Both properties aren't available by default in Firefox.
+    let ascent = metrics.fontBoundingBoxAscent;
+    let descent = Math.abs(metrics.fontBoundingBoxDescent);
+    if (ascent) {
+      ctx.restore();
+      const ratio = ascent / (ascent + descent);
+      ascentCache.set(fontFamily, ratio);
+      return ratio;
+    }
+
+    // Try basic heuristic to guess ascent/descent.
+    // Draw a g with baseline at 0,0 and then get the line
+    // number where a pixel has non-null red component (starting
+    // from bottom).
+    ctx.strokeStyle = "red";
+    ctx.clearRect(0, 0, DEFAULT_FONT_SIZE, DEFAULT_FONT_SIZE);
+    ctx.strokeText("g", 0, 0);
+    let pixels = ctx.getImageData(0, 0, DEFAULT_FONT_SIZE, DEFAULT_FONT_SIZE)
+      .data;
+    descent = 0;
+    for (let i = pixels.length - 1 - 3; i >= 0; i -= 4) {
+      if (pixels[i] > 0) {
+        descent = Math.ceil(i / 4 / DEFAULT_FONT_SIZE);
+        break;
+      }
+    }
+
+    // Draw an A with baseline at 0,DEFAULT_FONT_SIZE and then get the line
+    // number where a pixel has non-null red component (starting
+    // from top).
+    ctx.clearRect(0, 0, DEFAULT_FONT_SIZE, DEFAULT_FONT_SIZE);
+    ctx.strokeText("A", 0, DEFAULT_FONT_SIZE);
+    pixels = ctx.getImageData(0, 0, DEFAULT_FONT_SIZE, DEFAULT_FONT_SIZE).data;
+    ascent = 0;
+    for (let i = 0, ii = pixels.length; i < ii; i += 4) {
+      if (pixels[i] > 0) {
+        ascent = DEFAULT_FONT_SIZE - Math.floor(i / 4 / DEFAULT_FONT_SIZE);
+        break;
+      }
+    }
+
+    ctx.restore();
+
+    if (ascent) {
+      const ratio = ascent / (ascent + descent);
+      ascentCache.set(fontFamily, ratio);
+      return ratio;
+    }
+
+    ascentCache.set(fontFamily, DEFAULT_FONT_ASCENT);
+    return DEFAULT_FONT_ASCENT;
+  }
+
+  function appendText(task, geom, styles, ctx) {
     // Initialize all used properties to keep the caches monomorphic.
     const textDiv = document.createElement("span");
     const textDivProperties = {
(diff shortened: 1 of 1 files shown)


src/core/evaluator.js
```diff
@@ -2034,11 +2034,23 @@ class PartialEvaluator {
       }
       var font = textState.font;
       if (!(font.loadedName in seenStyles)) {
+        let ascent = font.ascent;
+        let descent = font.descent;
+        if (font.bbox) {
+          ascent = ascent || font.bbox[3];
+          descent = descent || -Math.abs(font.bbox[1]);
+        }
+        if (ascent && !isNaN(descent)) {
```
Snuffleupagus, line 2043: Don't you need to *also* guard against the divisor becoming zero here, since an ascent value of `Infinity` seems less than useful?
```suggestion
        if (ascent && !isNaN(descent) && ascent !== descent) {
```

src/display/text_layer.js
```diff
@@ -57,11 +57,66 @@ const renderTextLayer = (function renderTextLayerClosure() {
 
   const NonWhitespaceRegexp = /\S/;
 
+  const DEFAULT_FONT_SIZE = 30;
+  const DEFAULT_FONT_ASCENT = 0.8;
+
```
Snuffleupagus, line 62: Can you please move these lines up, such that `NonWhitespaceRegexp` and `isAllWhitespace` are still "together" here?

src/display/text_layer.js
```diff
@@ -57,11 +57,66 @@ const renderTextLayer = (function renderTextLayerClosure() {
...
+      .data;
+    descent = 0;
+    for (let i = pixels.length - 1 - 3; i >= 0; i -= 4) {
+      if (pixels[i] > 0) {
+        descent = Math.ceil(i / 4 / DEFAULT_FONT_SIZE);
+        break;
+      }
+    }
+
+    // Draw a A with baseline at 0,DEFAULT_FONT_SIZE and then get the line
```
Snuffleupagus, line 96: Nit: "Draw **an** A with ..."

src/display/text_layer.js
```diff
@@ -57,11 +57,66 @@ const renderTextLayer = (function renderTextLayerClosure() {
...
-  function appendText(task, geom, styles) {
+  function getAscent(fontFamily, ctx) {
+    const savedFont = ctx.font;
+    ctx.font = `${DEFAULT_FONT_SIZE}px ${fontFamily}`;
+    const metrics = ctx.measureText("");
+
+    // Both properties aren't available by default in Firefox.
+    let ascent = metrics.fontBoundingBoxAscent;
+    let descent = Math.abs(metrics.fontBoundingBoxDescent);
+    if (ascent) {
```
Snuffleupagus, line 75: As above, please avoid the zero-divisor here.

src/display/text_layer.js
```diff
@@ -57,11 +57,66 @@ const renderTextLayer = (function renderTextLayerClosure() {
...
+    for (let i = 0, ii = pixels.length; i < ii; i += 4) {
+      if (pixels[i] > 0) {
+        ascent = DEFAULT_FONT_SIZE - Math.floor(i / 4 / DEFAULT_FONT_SIZE);
+        break;
+      }
+    }
+
+    ctx.font = savedFont;
+
+    if (ascent) {
```
Snuffleupagus, line 112: As above, please avoid the zero-divisor here.

src/display/text_layer.js
```diff
@@ -91,11 +146,10 @@ const renderTextLayer = (function renderTextLayerClosure() {
...
     const fontHeight = Math.sqrt(tx[2] * tx[2] + tx[3] * tx[3]);
     let fontAscent = fontHeight;
-    if (style.ascent) {
-      fontAscent = style.ascent * fontAscent;
-    } else if (style.descent) {
-      fontAscent = (1 + style.descent) * fontAscent;
+    if (!style.ascent) {
+      style.ascent = getAscent(style.fontFamily, ctx);
     }
+    fontAscent *= style.ascent;
```
Snuffleupagus, line 152: If you want to, you can probably simplify this a little bit now that you always have a `style.ascent` value:
```javascript
    if (!style.ascent) {
      style.ascent = getAscent(style.fontFamily, ctx);
    }
    const fontAscent = fontHeight * style.ascent;
```

Snuffleupagus (changes requested):

src/display/text_layer.js
calixteman: These numbers are a distance so they must be positive or zero:
https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics/actualBoundingBoxRight
so if ascent is non-zero, divisor is non-zero too.

src/display/text_layer.js
calixteman: The same as above: these numbers are positive or zero.

src/core/evaluator.js
calixteman: Isn't the descent supposed to be a negative or zero number ? And the ascent a positive or zero one ?

Snuffleupagus: Generally speaking that's obviously how these things *should* behave, but given all the stupid things that PDF generators tend to do I wanted to at least ask about this :-)

timvandermeij:
The Windows bot seems down again. Aside from that, while this seems to work nicely for the Tracemonkey file, it seems to regress quite a bit of other files unfortunately, so this needs some more work. Examples of noticeable movement are `firefox-arabiccidtruetype-text-page1`, `firefox-issue4550-text-page2` and `firefox-extgstate-text-page1`.

calixteman:
So I reboot the windows bot.
The new version of the patch doesn't take anymore into account the ascent/descent of the font in the pdf to position the baseline of the fallback font in the text layer but use ones from the fallback font which makes sense finally.

calixteman:
I'm not sure to understand how to interpret some results of unit tests.
For example with `firefox-taro-text-page1` (i.e. `TaroUTR50SortedList112.pdf` ), the text in text_layer is better positioned with patch locally but it isn't the case in reftest-analyzer where it seems to be a way better in the reference.
And the same for tracemonkey.pdf: it looks ok locally but neither in the test nor in the reference.

Snuffleupagus:
> [...] the text in text_layer is better positioned with patch locally but it isn't the case in reftest-analyzer where it seems to be a way better in the reference.

In *theory* the output of the `text` tests should agree with the development viewer, when it's opened with http://localhost:8888/web/viewer.html#textLayer=visible, so if that's not actually the case this suggests a *pre-existing* bug in our test-suite.

The relevant test-code is found in https://github.com/mozilla/pdf.js/blob/6249ef517d3aaacc9aa6c9e1f5377acfaa4bc2a7/test/driver.js#L27-L107 and the CSS-file that it loads is found in https://github.com/mozilla/pdf.js/blob/master/test/text_layer_test.css

---

Unfortunately we're going to need to somehow debug/fix the underlying issue *first*, since it's virtually impossible to review this patch if we cannot actually trust the test-suite.

timvandermeij:
Now that the issue mentioned above should be fixed, could you rebase this and retrigger the tests to see how this goes now?

Snuffleupagus:
Looking at the latest test results, it seems to be pretty even split between test-cases that improve respectively regress with this patch; hence it's unfortunately not really clear (at least to me) that this is an overall improvement as-is.

timvandermeij:
Yes, for example `chrome-arabiccidtruetype-text-page1` is clearly improved and `chrome-issue6387-text-page1` is even quite a dramatic improvement, but I also see `firefox-issue8229-page1` for example which looks a bit worse. In general my feeling is that overall this is an improvement, but it would be good to have an additional look at the ones that look a bit worse here to see if it can be improved further.

calixteman:
@timvandermeij I've the impress that overall this is an improvement too even if it's a bit worse for some of them.

brendandahl:
/botio test
(re-running to see with chrome updates)

timvandermeij:
Hm, looks like the Windows bot became unreachable here again.

timvandermeij (approved):

timvandermeij:
I have looked at all reference test differences with this version of the patch and I'd say this is an overall improvement. Some files are improved quite dramatically, others have minor improvements and for a few it's a bit worse, but given that we have to rely on a heuristic here I don't think that will ever be perfect, so this will remain to be a trade-off. In this case the majority of cases is an improvements, so let's go with this and we can always tweak it further once more test cases are presented to us.

Thank you for putting effort into this!

timvandermeij:
This closed quite a few issues, some of them mentioned above, so nice work!

minhan12397:
@timvandermeij  I see that the problem has not been thoroughly resolved.

Issue #12895, opened 2021-01-23
Possible typo/error in `/src/display/annotation_layer.js` with the "MouseUp" eventName


Snuffleupagus:
In the following spots a "MouseUp" eventName is being used, which does look inconsistent considering that the "Mouse Down"/"Mouse Enter"/"Mouse Exit" eventNames all have a space in them.
Is this inconsistency actually expected, or is it a simple typo? If the former, then a comment explaining *why* would be appropriate; otherwise we should just fix this :-)

 - https://github.com/mozilla/pdf.js/blob/25b84ce84c95ba06d1ee57e578672048269c39c1/src/display/annotation_layer.js#L806

 - https://github.com/mozilla/pdf.js/blob/25b84ce84c95ba06d1ee57e578672048269c39c1/src/display/annotation_layer.js#L945

 - https://github.com/mozilla/pdf.js/blob/25b84ce84c95ba06d1ee57e578672048269c39c1/src/display/annotation_layer.js#L1034

 - https://github.com/mozilla/pdf.js/blob/25b84ce84c95ba06d1ee57e578672048269c39c1/src/display/annotation_layer.js#L1168

/cc @calixteman

PR #12897, opened 2021-01-23
JS - Fix mouse event names

calixteman:
- fix issue #12895

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -370,8 +370,8 @@ class LinkAnnotationElement extends AnnotationElement {
       parameters.data.isTooltipOnly ||
       (parameters.data.actions &&
         (parameters.data.actions.Action ||
-          parameters.data.actions.MouseUp ||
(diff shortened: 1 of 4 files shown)


src/display/annotation_layer.js
```diff
@@ -577,8 +579,7 @@ class WidgetAnnotationElement extends AnnotationElement {
     for (const [baseName, eventName] of names) {
       if (
         eventName === "Action" ||
-        (this.data.actions &&
-          this.data.actions[eventName.replace(" ", "")] !== undefined)
+        (this.data.actions && this.data.actions[eventName] !== undefined)
```
Snuffleupagus, line 582: Would `this.data.actions?.[eventName]` suffice here?

Snuffleupagus, line 582: Would is suffice to just check the following instead?
```suggestion
        this.data.actions?.[eventName]
```

timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #12899, opened 2021-01-24
Update the ESLint `env` to use "es2021"

Snuffleupagus:
Currently this is inconsistent, since we're using ECMAScript 2021 in the parser options.


timvandermeij (approved):

timvandermeij:
Thanks!

Issue #12098, opened 2020-07-15
PDFs generated by selectpdf.com slow to render in pdf.js


zoetyu:
Attach (recommended) or Link to PDF file here:
Problem PDF: [wikipedia-select-net.pdf](https://github.com/mozilla/pdf.js/files/4922949/wikipedia-select-net.pdf)
Good PDF for contrast: [Wikipedia-chrome.pdf](https://github.com/mozilla/pdf.js/files/4922950/Wikipedia-chrome.pdf)

Configuration:
- Web browser and its version: Chrome 83.0.4103.116
- Operating system and its version: MacOS 10.14.6
- PDF.js version: 2.5.207, discovered in 2.1.266
- Is a browser extension: No

Steps to reproduce the problem:
1. Use attached pdf "wikipedia-select-net.pdf" above or generate your own from https://selectpdf.com/
    * This pdf was generated from selectpdf.com. For contrast, a PDF generated using the chrome native HTML to PDF converter is also provided; this pdf has no slowness issues.
2. Open PDF in the demo, scroll around, see that the PDF is slow to load

What is the expected behavior? (add screenshot)
PDF pages should render quickly for PDFs created via Selectnet. The load time should more closely resemble the Chrome PDF.

I am filing this as a pdf.js bug because Selectnet PDFs are able to render quickly in other applications, e.g. Preview or Chrome

What went wrong? (add screenshot)
PDF takes excessive time to load

![selectnet-slow-render](https://user-images.githubusercontent.com/35852738/87507844-19b74500-c63c-11ea-85c5-642bff3040c1.gif)

Related to the render time, is the excessive amount of memory consumed by the selectnet PDF. Example of memory comparison between selectnet and chrome PDF (see tabs for title)
![image](https://user-images.githubusercontent.com/35852738/87508490-8121c480-c63d-11ea-8577-1654df59d201.png)
My coworker found that:

>  the textLayer HTML element for each individual page contains the full text for the entire document, but with the positioning such that text elements not for that page are outside the viewport of that page. This is not the case for the PDF generated from Chrome. This would appear to explain the excessive memory use while viewing these PDFs because if you have an N page document, then the full content of those N pages is included in the textLayer of each of those N pages leading to an N^2 growth in memory consumption.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

amit777:
[...]

PR #12903, opened 2021-01-25
Improve global image caching for small images (PR 11912 follow-up, issue 12098)

Snuffleupagus:
When implementing the `GlobalImageCache` functionality I was mostly worried about the effect of *very large* images, hence the maximum number of cached images were purposely kept quite low[1].
However, there's one fairly obvious problem with that approach: In documents with hundreds, or even thousands, of *small* images the `GlobalImageCache` as implemented becomes essentially pointless.

Hence this patch, where the `GlobalImageCache`-implementation is changed in the following ways:
 - We're still guaranteed to be able to cache a *minimum* number of images, set to `10` (similar as before).
 - If the *total* size of all the cached image data is below a threshold[2], we're allowed to cache additional images.

This patch thus *improve*, but doesn't completely fix, issue #12098. Note that that document is created by a *very poor* PDF generator, since every single page contains the *entire* document (with all of its /Resources) and to create the individual pages clipping is used.[3]

---
[1] Currently set to `10` images; imagine what would happen to overall memory usage if we encountered e.g. 50 images each 10 MB in size.

[2] This value was chosen, somewhat randomly, to be `40` megabytes; basically five times the [maximum individual image size per page](https://github.com/mozilla/pdf.js/blob/6249ef517d3aaacc9aa6c9e1f5377acfaa4bc2a7/src/display/api.js#L2483-L2484).

[3] This surely has to be some kind of record w.r.t. how badly PDF generators can mess things up...

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -609,6 +609,9 @@ class PartialEvaluator {
       .then(imageObj => {
         imgData = imageObj.createImageData(/* forceRGBA = */ false);
 
+        if (cacheKey && imageRef && cacheGlobally) {
+          this.globalImageCache.addByteSize(imageRef, imgData.data.length);
+        }
         return this._sendImgData(objId, imgData, cacheGlobally);
       })
       .catch(reason => {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

PR #12904, opened 2021-01-25
fix initial state of checkboxes in display layer

### LINKED ISSUE #12716: Radio-Button group shows all buttons as checked while only one is
Attach (recommended) or Link to PDF file here: [form-cms1500.pdf](https://github.com/mozilla/pdf.js/files/5670936/form-cms1500.pdf) (Original opened in Acrobat and checked one radio-button)
Original Source: https://www.cigna.com/static/www-cigna-com/docs/health-care-providers/form-cms1500.pdf

Configuration:
- Web browser and its version: FF 83.0
- PDF.js version: 2.7.355

1. Open the PDF here: https://mozilla.github.io/pdf.js/web/viewer.html
2. See that all radio-buttons are checked, which is wrong.

grafik

grafik

This radio-button field can be found in object `45 0 obj`.

dhufnagel:
When having multiple checkboxes with the same name, they behave like radiobuttons and the value "V" is equal to the export value of the checked checkbox.

Fixes #12716.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -876,7 +876,10 @@ class CheckboxWidgetAnnotationElement extends WidgetAnnotationElement {
     const data = this.data;
     const id = data.id;
     const value = storage.getOrCreateValue(id, {
-      value: data.fieldValue && data.fieldValue !== "Off",
+      value:
+        data.fieldValue &&
+        ((data.exportValue && data.exportValue === data.fieldValue) ||
+          (!data.exportValue && data.fieldValue !== "Off")),
     }).value;
 
     this.container.className = "buttonWidgetAnnotation checkBox";
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -5013,5 +5013,15 @@
       "rounds": 1,
       "type": "eq",
       "forms": true
+    },
+    {
+      "id": "issue12716",
+      "file": "pdfs/issue12716.pdf",
+      "md5": "9bdc9c552bcfccd629f5f97385e79ca5",
+      "rounds": 1,
+      "link": true,
+      "type": "eq",
+      "forms": true,
+      "lastPage": 1
     }
 ]
--- a/test/pdfs/issue12716.pdf.link
+++ b/test/pdfs/issue12716.pdf.link
@@ -1,0 +1,1 @@
+https://github.com/mozilla/pdf.js/files/5670936/form-cms1500.pdf


test/pdfs/.gitignore
```diff
@@ -419,3 +419,4 @@
 !issue12337.pdf
 !pr12564.pdf
 !pr12828.pdf
+!issue12716.pdf
```
Snuffleupagus, line 422: This should be a *linked* test-case, given that we don't have explicit permission to include it in the repository; please just link to https://github.com/mozilla/pdf.js/files/5670936/form-cms1500.pdf instead.

test/test_manifest.json
```diff
@@ -5013,5 +5013,13 @@
       "rounds": 1,
       "type": "eq",
       "forms": true
+    },
+    {
+      "id": "issue12716",
+      "file": "pdfs/issue12716.pdf",
+      "md5": "b91ee4c41a5902710fcdd27166416fd6",
+      "rounds": 1,
+      "type": "eq",
```
Snuffleupagus, line 5022: Unless actually necessary, please test only *one* page to save resources; adding `"lastPage": 1,` here will limit testing to just the *first* page.

Snuffleupagus (changes requested):

test/pdfs/.gitignore
dhufnagel: Thank you for the hint. Didn't know this is possible.

src/display/annotation_layer.js
```diff
@@ -876,7 +876,10 @@ class CheckboxWidgetAnnotationElement extends WidgetAnnotationElement {
     const data = this.data;
     const id = data.id;
     const value = storage.getOrCreateValue(id, {
-      value: data.fieldValue && data.fieldValue !== "Off",
+      value:
```
calixteman, line 879: It should be enough to have `data.fieldValue === data.exportValue`.

dhufnagel: I first thought too, but the data.exportValue is defined in [https://github.com/mozilla/pdf.js/blob/6249ef517d3aaacc9aa6c9e1f5377acfaa4bc2a7/src/core/annotation.js#L1923](here). So there might be cases where this can be undefined?!

dhufnagel: Hi @calixteman . Do you think this should be enough because the pdf might be broken if the exportValue is not set? Should I change it according to your suggestion?

calixteman: It's likely better to be defensive so +1.

timvandermeij:
The tests failed because the MD5 in the manifest seems incorrect:

```WARNING: MD5 of file "pdfs/issue12716.pdf" does not match file. Expected "b91ee4c41a5902710fcdd27166416fd6" computed "9bdc9c552bcfccd629f5f97385e79ca5"```

dhufnagel:
Yep, my fault. I had an local file with some other boxes checked. Fixed it.

dhufnagel:
Am I supposed to do anything because of the failed regression tests? It seems none of them relate to forms, so I don't know what the problem may be.

timvandermeij:
Those regression test failures are indeed unrelated, so you don't have to do anything for that. I'll retrigger the tests since we recently updated the test bots.

/botio test

calixteman:
Thank you for fixing this.

PR #12908, opened 2021-01-25
Slightly rescale lineWidth to workaround chrome rendering issue

### LINKED ISSUE #11918: Rendering issue with some objects on Chrome
[Test file.pdf](https://github.com/mozilla/pdf.js/files/4656226/Test.file.pdf)

Configuration:
- Web browser and its version: Chrome Version 81.0.4044.138 (Official Build) (64-bit)
- PDF.js version: latest
- Is a browser extension: no

1. Open [https://mozilla.github.io/pdf.js/web/viewer.html](https://mozilla.github.io/pdf.js/web/viewer.html)
2. Load the provided .pdf file

image

image

calixteman:
It aims to fix #11918.

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -42,6 +42,13 @@ const MAX_SIZE_TO_COMPILE = 1000;
 
 const FULL_CHUNK_HEIGHT = 16;
 
+// Because of https://bugs.chromium.org/p/chromium/issues/detail?id=1170396
+// some curves aren't rendered correctly.
+// Multiplying lineWidth by this factor should help to have "correct"
+// rendering with no artifacts.
+// Once the bug is fixed upstream, we can remove this constant and its use.
+const LINEWIDTH_SCALE_FACTOR = 1.000001;
+
 function addContextCurrentTransform(ctx) {
   // If the context doesn't expose a `mozCurrentTransform`, add a JS based one.
   if (!ctx.mozCurrentTransform) {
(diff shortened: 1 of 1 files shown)


src/display/canvas.js
```diff
@@ -42,6 +42,12 @@ const MAX_SIZE_TO_COMPILE = 1000;
 
 const FULL_CHUNK_HEIGHT = 16;
 
+// Because of https://bugs.chromium.org/p/chromium/issues/detail?id=1170396
+// some curves aren't rendered correctly.
+// Multiplying lineWidth by this factor should help to have "correct"
+// rendering with no artifacts.
```
timvandermeij, line 48: Should we mention here that this constant can be removed entirely once the upstream bug is fixed? Or should this patch basically be reverted in that case, given that the 1.000001 number was already in use before this patch if `absDet > Number.EPSILON`?

src/display/canvas.js
```diff
@@ -144,6 +151,12 @@ function addContextCurrentTransform(ctx) {
       ctx._originalSetTransform(a, b, c, d, e, f);
     };
 
+    ctx.resetTransform = function ctxResetTransform(a, b, c, d, e, f) {
```
timvandermeij, line 154: Given that this, and all multiplications below now happen unconditionally for every PDF even though this is a Chrome-specific issue, I'd really like to see a benchmark here to make sure performance of larger PDF files from the test suite doesn't suffer. See https://github.com/mozilla/pdf.js/wiki/Benchmarking-your-changes for how to do this.

timvandermeij (changes requested):

src/display/canvas.js
Snuffleupagus: I think this is fine, since it does *not* apply to Firefox given that all of this code is within the `if (!ctx.mozCurrentTransform) { ... }` block

src/display/canvas.js
calixteman: In the updated patch the multiplication by the constant is chrome-only.

src/display/canvas.js
calixteman: The stuff here is chrome-only so there is no impact for firefox.

src/display/canvas.js
```diff
@@ -52,10 +59,33 @@ function addContextCurrentTransform(ctx) {
     ctx._originalTranslate = ctx.translate;
     ctx._originalTransform = ctx.transform;
     ctx._originalSetTransform = ctx.setTransform;
+    ctx._originalResetTransform = ctx.resetTransform;
 
     ctx._transformMatrix = ctx._transformMatrix || [1, 0, 0, 1, 0, 0];
     ctx._transformStack = [];
 
+    try {
+      // The call to getOwnPropertyDescriptor throws an exception with node:
```
timvandermeij, line 68: Where does it do this exactly? This code is supposed to only run in Chrome, but if I call `getOwnPropertyDescriptor` in Firefox I don't get this exception, so that would mean that this is also executed in Firefox...

Snuffleupagus: > In which browser does it throw this exception exactly?

The comment mentions "node", so I'd guess that this is only required when the unit-tests are run in Node.js environments.
Perhaps the comment would be *slightly* clearer if this line was reworded to:
```suggestion
      // The call to getOwnPropertyDescriptor throws an exception in Node.js:
```

> [...] but if I call `getOwnPropertyDescriptor` in Firefox I don't get this exception, so that would mean that this is also executed in Firefox...

This *entire* block is guarded by a `if (!ctx.mozCurrentTransform) {` check above, hence none of this code should run in Firefox as far as I can tell.

timvandermeij: Ah, I missed the `mozCurrentTransform` guard. In that case it looks good to me. Rewording the comment would indeed have helped since I was thinking about a DOM node since it was written in lower case.

timvandermeij:
@calixteman Could you perhaps address the comment above? I think we can land this once that's done.

calixteman:
Sorry, I postponed and forgot.

timvandermeij (approved):

timvandermeij:
No worries! Looks good like this; thanks!

/botio makeref

PR #12910, opened 2021-01-25
Add back dir property in spans in text layer

### LINKED ISSUE #12909: Text layer doesn't use dir property anymore
Some characters in:
[hebrew.pdf](https://github.com/mozilla/pdf.js/files/5869562/hebrew.pdf)
are not correct in the text layer (such as question marks).

We get text direction in `evaluator.js` but we don't use it.
It was used in the past, but it seems it has been removed:
https://github.com/mozilla/pdf.js/commit/5bd8a83c9bd7c23003b09e1cfd71422c6b083826#diff-579f2fb8055052aa97906d5fc75aeee84aa7dd9f8efcca24d8a5460d828e1de8L170

calixteman:
- aims to fix #12909

### DIFF
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -4526,6 +4526,14 @@
        "link": false,
        "type": "eq"
     },
+    {  "id": "issue12909",
+       "file": "pdfs/issue12909.pdf",
+       "md5": "0faa1cb029c3323c168595cccbe46ec0",
+       "link": true,
+       "rounds": 1,
+       "lastPage": 1,
+       "type": "text"
+    },
     {  "id": "issue5592",
        "file": "pdfs/issue5592.pdf",
        "md5": "a0750f95afa80c880f7966df7062616c",
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -113,6 +113,9 @@ const renderTextLayer = (function renderTextLayerClosure() {
     textDiv.style.fontFamily = style.fontFamily;
 
     textDiv.textContent = geom.str;
+    // geom.dir may be 'ttb' for vertical texts.
+    textDiv.dir = geom.dir;
+
     // `fontName` is only used by the FontInspector, and we only use `dataset`
     // here to make the font name available in the debugger.
     if (task._fontInspectorEnabled) {
--- a/test/pdfs/issue12909.pdf.link
+++ b/test/pdfs/issue12909.pdf.link
@@ -1,0 +1,1 @@
+https://web.archive.org/web/20210126084513/https://www.machonso.org/uploads/images/%D7%95%D7%95%D7%9C%D7%A3.pdf


test/pdfs/.gitignore
```diff
@@ -70,6 +70,7 @@
 !bug1650302_reduced.pdf
 !issue8570.pdf
 !issue8697.pdf
+!issue12909.pdf
```
Snuffleupagus, line 73: Shouldn't this be a linked test-case instead?

test/test_manifest.json
```diff
@@ -4526,6 +4526,12 @@
        "link": false,
        "type": "eq"
     },
+    {  "id": "issue12909",
+       "file": "pdfs/issue12909.pdf",
+       "md5": "0faa1cb029c3323c168595cccbe46ec0",
+       "rounds": 1,
```
Snuffleupagus, line 4532: Do you need to test all pages, or can we limit it to just *one* page?

Snuffleupagus (changes requested):

test/pdfs/.gitignore
calixteman: I don't know.
How do you decide this in general ?

brendandahl: I don't think we have a rule, but generally I follow:
- reduced test case - include in repo
- small file author has given us permission to use - include in repo
- linked that will likely keep working (e.g. github, bugzilla) - linked test case
- linked file from random site - try to create reduced test case or use way back machine

Snuffleupagus, line 73: Please remove this line, since you're now adding a linked test.

Snuffleupagus (approved):
Looks good to me, with the final comment addressed *and* test run on the Windows-bot as well (just to be safe).

calixteman:
I see that in the log:
`WARNING: MD5 of file "pdfs/issue12909.pdf" does not match file. Expected "0faa1cb029c3323c168595cccbe46ec0" computed "d41d8cd98f00b204e9800998ecf8427e"`
I checked the md5 and it's ok:
```bash
wget "https://web.archive.org/web/20210126084513/https://www.machonso.org/uploads/images/%D7%95%D7%95%D7%9C%D7%A3.pdf" -qO - | md5sum
0faa1cb029c3323c168595cccbe46ec0  -
```
and I get the second md5 with an empty file:
```bash
touch foo && md5sum foo
d41d8cd98f00b204e9800998ecf8427e  foo
```
Could it explain why regression tests aren't ran ?

Snuffleupagus:
Given the previous errors in https://github.com/mozilla/pdf.js/pull/12910#issuecomment-767472177, I'd suggest removing the `issue12909.pdf` *and* `issue12909.pdf.error` files from the Windows bot and trying again.

Snuffleupagus:
Huh, what happened to the fonts in the `eq`-tests?

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

PR #12914, opened 2021-01-26
JS - QuickJS sandbox initialization must be the last evaluated string

calixteman:
- aims to fix issue #12912

PR #12916, opened 2021-01-26
Include `pdf.sandbox.js` when building the pdfjs-dist files

Snuffleupagus:
Currently it's missing in pdfjs-dist, however that's not a big issue considering that it's broken in non-TESTING builds before PR #12914 anyway.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12922, opened 2021-01-27
Ignore globally cached images in `PartialEvaluator.getTextContent` (PR 11930 follow-up)

Snuffleupagus:
Given that we'll only cache `/XObject`s of the `Image`-type globally, we can utilize that in `PartialEvaluator.getTextContent` as well. This way, in cases such as e.g. issue #12098, we can avoid having to fetch/parse `/XObject`s that we already know to be `Image`s. This is helpful, since `Stream`s are not cached on the `XRef` instance (given their potential size) and the lookup can thus be somewhat expensive in general.

Also, skip a redundant `RefSetCache.has` check in the `GlobalImageCache.getData` method.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -2515,6 +2515,15 @@ class PartialEvaluator {
                     return;
                   }
 
+                  const globalImage = self.globalImageCache.getData(
+                    xobj,
+                    self.pageIndex
+                  );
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good idea; thanks!

PR #12924, opened 2021-01-27
Fix font data clone error when pdfBug is enabled.

brendandahl:
The widths property should be an object to match what metrics returns.

In ZapfDingbats.pdf I was getting a data clone error with pdfBug enabled.
In buildCharCodeToWidth() there was an encoding with the name "at" which
is also the name of a method on an array. buildCharCodeToWidth assumes an
object is passed in, so when it checked for the "at" property, it found the
method and copied it over.

This only seemed to affect Firefox.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -3192,7 +3192,7 @@ class PartialEvaluator {
 
   getBaseFontMetrics(name) {
     var defaultWidth = 0;
-    var widths = [];
+    var widths = Object.create(null);
     var monospace = false;
     var stdFontMap = getStdFontMap();
     var lookupName = stdFontMap[name] || name;


Snuffleupagus:
> so when it checked for the "at" property, it found the
> method and copied it over.
> 
> This only seemed to affect Firefox.

That's probably because `at` isn't generally available in browsers yet, with the exception of Firefox Nightly, given that it's a very recent addition to the ECMAScript specification. There's also currently some web compatibility questions regarding the name itself; see https://github.com/tc39/proposal-relative-indexing-method/issues/41

Snuffleupagus (approved):
Looks good to me, thank you!

timvandermeij (approved):

timvandermeij:
Thanks!

PR #12932, opened 2021-01-30
Stop including `-ms` prefixed CSS rules in GENERIC builds

Snuffleupagus:
Given that IE 11/Edge is now unsupported in PDF.js, and that Microsoft Edge is now a Chromium-browser, we can avoid (some) unnecessary bloat in the built CSS files.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12933, opened 2021-01-30
Stop polyfilling CSS variables in GENERIC builds

Snuffleupagus:
At this point in time, all browsers that we support have native support for CSS variables; please see https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#browser_compatibility and https://github.com/mozilla/pdf.js/blob/9af8501e6c44eb32bea905c3b56b4c1e6c5f5ba5/gulpfile.js#L79-L91

This reduces the size of the *built* `viewer.css` file, in GENERIC builds, from `93 814` to `55 285` bytes (~41 percent).

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -19,7 +19,6 @@
 
 var autoprefixer = require("autoprefixer");
 var calc = require("postcss-calc");
-var cssvariables = require("postcss-css-variables");
 var fs = require("fs");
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Nice, that's a pretty significant reduction!

PR #12935, opened 2021-01-31
Remove prefixed fullscreen properties/methods from the MOZCENTRAL builds

Snuffleupagus:
The unprefixed version of the fullscreen API has been enabled for quite some time in Firefox, see below, hence we can (slightly) clean-up the relevant code.

 - https://developer.mozilla.org/en-us/docs/Web/API/Document/fullscreenEnabled#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen#browser_compatibility

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -657,26 +657,22 @@ const PDFViewerApplication = {
   },
 
   get supportsFullscreen() {
-    let support;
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12936, opened 2021-01-31
XFA - Add a lexer/parser for FormCalc language

calixteman:
- the language specifications are: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=1049
 - it can be used to:
   * as a scripting language for calculation, validations, ...
   * in SOM expressions to select nodes: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=101

### DIFF
--- a/src/core/xfa/formcalc_lexer.js
+++ b/src/core/xfa/formcalc_lexer.js
@@ -1,0 +1,385 @@
+/* Copyright 2021 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const KEYWORDS = new Set([
+  "and",
+  "break",
+  "continue",
+  "do",
+  "downto",
+  "else",
+  "elseif",
+  "end",
+  "endfor",
+  "endfunc",
+  "endif",
+  "endwhile",
+  "eq",
+  "exit",
+  "for",
+  "foreach",
+  "func",
+  "ge",
+  "gt",
+  "if",
+  "in",
+  "infinity",
+  "le",
+  "lt",
+  "nan",
+  "ne",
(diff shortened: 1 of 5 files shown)


calixteman:
For the scripting stuff, an idea is to generate some javascript from the generated AST and then execute it in the js sandbox.
It's the way chose by pdfium:
https://pdfium.googlesource.com/pdfium/+/refs/heads/master/xfa/fxfa/fm2js/cxfa_fmexpression.cpp

benbro:
Will it be possible to disable scripting?

timvandermeij:
It's a preference now and it's disabled by default. There are no plans to remove this preference because scripting is a feature that we can imagine that people want to disable.

This question is unrelated to this PR though, so let's leave it at this.

src/core/xfa/formcalc_lexer.js
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    if (0x30 /* = 0 */ <= char && char <= 0x39 /* = 9 */) {
+      this.pos++;
+      const match = this.data.substring(this.pos).match(dotNumberPattern);
+      if (!match) {
+        return new Token(TOKEN.number, (char - 0x30) /* = 0 */ / 10);
+      }
+      const end = this.pos + match[0].length;
+      const number = parseFloat(this.data.substring(this.pos - 2, end));
```
brendandahl, line 295: What should happen in this language if we parse a really large number that would be infinity in JS? e.g. "9e99999"

brendandahl (changes requested):

src/core/xfa/formcalc_lexer.js
```diff
@@ -0,0 +1,372 @@
...
+      return new Token(alt1);
+    }
+    return new Token(alt2);
+  }
+
+  getLower() {
+    const char = this.data.charCodeAt(this.pos);
+    if (char === 0x3d /* = = */) {
+      this.pos++;
+      return new Token(TOKEN.le);
```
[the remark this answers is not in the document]
calixteman: good idea

src/core/xfa/formcalc_lexer.js
calixteman: Numbers are following IEEE754 format: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=1056
like in JS: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Numbers_and_dates
So using `parseFloat` should give the correct number.

brendandahl (approved):
r+ with tests passing

PR #12937, opened 2021-02-01
making file more readable by shifting curly braces

aman282571:
I have made some indentation. Can anyone confirm so that i make whole file more readable.
@timvandermeij

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,28 +3,34 @@
   "version": "2.0.0",
   "lockfileVersion": 1,
   "requires": true,
-  "dependencies": {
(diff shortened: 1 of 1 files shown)


Snuffleupagus (changes requested):
> [...] so that i make whole file more readable.

In this case, none of the changes constitutes any sort of readability improvement here. And this type of file isn't intended to be read be humans anyway.

---

These types of changes are *not at all* correct, since this file is created automatically by Node.js when installing/updating packages.

Unfortunately we cannot accept patches such as these!

PR #12938, opened 2021-02-01
fix: removed comma from last key-value pair(ES6 notation) in import and export statement

aman282571:
removed comma from last key-value pair(ES6 notation) in import and export statement and indented the file

### DIFF
--- a/src/pdf.js
+++ b/src/pdf.js
@@ -22,7 +22,7 @@ import {
   LinkTarget,
   loadScript,
   PDFDateString,
-  RenderingCancelledException,
+  RenderingCancelledException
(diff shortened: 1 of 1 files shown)


aman282571:
can anyone tell me why all checks are not successful

Snuffleupagus (changes requested):
> can anyone tell me why all checks are not successful

Because the changes you're making here explicitly violates the ESLint and/or Prettier styling that's automatically enforced for all code.
Generally, there's also a very good reason for having commas even at the last line, since that reduces the amount of code churn (and changing blame) when making future code changes.

---

Unfortunately we cannot accept patches such as these!

PR #12939, opened 2021-02-01
removed body element because its empty

aman282571:


docs/contents/css/main.css
```diff
@@ -1,5 +1,4 @@
-body {
-}
+
```
Snuffleupagus, line 1: Why leave a blank line here?

Snuffleupagus (changes requested):
Have you successfully tested this locally?

---

It's not entirely clear to me that this is an overall necessary change.
But if we do end up accepting the patch, it definitely needs a better commit message since it's really impossible to tell what's being changed here without reading the code.

timvandermeij:
Closing since there was no response. This can be reopened if the comments above are both addressed.

PR #12941, opened 2021-02-01
fix(a11y): resolve sidebar, find, toolbar missing aria-expanded and aria-controls state

justinribeiro:
In testing, screen readers such as JAWS have trouble understanding the expanded state of the buttons that expand hidden menus due to lacking `aria-expanded` attribute. Also, given that the buttons do not contain the controlled/shown element, they should also define the `aria-controls` attribute with associated element id per https://www.w3.org/TR/wai-aria-1.1/#aria-expanded

This fixes adds these requirements for the sidebar, find, and secondary toolbar buttons.

### DIFF
--- a/web/pdf_find_bar.js
+++ b/web/pdf_find_bar.js
@@ -211,6 +211,7 @@ class PDFFindBar {
     if (!this.opened) {
       this.opened = true;
       this.toggleButton.classList.add("toggled");
+      this.toggleButton.setAttribute("aria-expanded", "true");
       this.bar.classList.remove("hidden");
     }
     this.findField.select();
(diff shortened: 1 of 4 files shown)


Snuffleupagus (changes requested):
Please keep in mind that the *commit message* itself should contain all of the information found in https://github.com/mozilla/pdf.js/pull/12941#issue-565523643, such that it's also available e.g. on the Git command line.

timvandermeij:
Looks good to me with an amended commit message.

justinribeiro:
@Snuffleupagus per @timvandermeij commit message amended. Cheers!

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #12952, opened 2021-02-03
Don't focus the `PasswordPrompt` input-field on load, when the viewer is embedded in e.g. an `iframe` (issue 12951)

### LINKED ISSUE #12951: Viewer: Password secured PDF focussing issue
Tested in latest Chrome and FF:

If you have multiple PDF files embedded as Iframe (via viewer.html) and at least one of them is password protected, the browser automatically jumps to the (first) iframe that shows the password prompt.

Imho, the iframe shouldn't modify the behavior of it's parent in any way.

See https://support.softcreatr.com/thread/4360-pdf-test/ for a demo.

Snuffleupagus:
Given that we don't focus the viewer *itself* (among other things) when the viewer is embedded, I suppose that it makes some sense to not focus the `PasswordPrompt` input-field either on load.
In order to improve the overall UX here, if an *incorrect* password was provided we'll still focus the input-field.

Fixes #12951 (assuming we care to do so, of course).

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -558,7 +558,8 @@ const PDFViewerApplication = {
     this.passwordPrompt = new PasswordPrompt(
       appConfig.passwordOverlay,
       this.overlayManager,
-      this.l10n
+      this.l10n,
+      this.isViewerEmbedded
     );
 
     this.pdfOutlineViewer = new PDFOutlineViewer({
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
I agree that it makes sense, and it's a small change, so I'm fine with this. Thanks!

PR #12961, opened 2021-02-05
Use optional chaining, where possible, in the `web/`-folder

Snuffleupagus:
- Slightly re-factor how the `BaseViewer`/`PDFThumbnailViewer` handle page labels internally, to make the `null` default value clearer

   Currently it's not *immediately* clear from the code itself, unless you look at the definition of `this._pageLabels`, that the default value is `null`.[1]
   We can improve this, and also reduce the amount of code, by using modern ECMAScript features such as optional chaining and nullish coalescing.

   ---
   [1] Keep in mind that an *empty* string is actually a valid page label, according to the PDF specification.

 - Use optional chaining, where possible, in the `web/`-folder

   By using optional chaining, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining, it's possible to reduce unnecessary code-repetition in many cases.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1326,7 +1326,7 @@ const PDFViewerApplication = {
           this._initializePdfHistory({
             fingerprint: pdfDocument.fingerprint,
             viewOnLoad,
-            initialDest: openAction && openAction.dest,
+            initialDest: openAction?.dest,
           });
           const initialBookmark = this.initialBookmark;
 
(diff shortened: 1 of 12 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #12962, opened 2021-02-05
Handle errors gracefully, in `PartialEvaluator.translateFont`, when fetching the font file (issue 9462)

### LINKED ISSUE #9462: First page is mixed with the last page
Could you take a look on this file. The first page is mixed with the last page.

[cb7e7bd2362328f37fe97dc8dff151a3.pdf](https://github.com/mozilla/pdf.js/files/1706865/cb7e7bd2362328f37fe97dc8dff151a3.pdf)

Configuration:
- Firefox Nightly 60
- Windows 7

Snuffleupagus:
The *third* page of the referenced PDF document currently fails to render completely, since one of its font files fail to load.
Since that error isn't handled, a large part of the text is thus missing which looks quite bad. By "replacing" the font data with an *empty* stream, we'll thus be able to fallback to rendering the text with a standard font (instead of using `ErrorFont`). While there's obviously no guarantee that things will look perfect, actually rendering the text at all should be an improvement in general.

Also, print a warning in `PartialEvaluator.loadFont` when the `PartialEvaluator.translateFont` method rejects, since that'd have helped debug/fix the issue faster.

Fixes #9462

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -47,6 +47,7 @@ import {
   Ref,
   RefSet,
 } from "./primitives.js";
+import { DecodeStream, NullStream } from "./stream.js";
 import {
   ErrorFont,
   Font,
@@ -85,7 +86,6 @@ import {
 } from "./image_utils.js";
 import { bidi } from "./bidi.js";
 import { ColorSpace } from "./colorspace.js";
-import { DecodeStream } from "./stream.js";
 import { getGlyphsUnicode } from "./glyphlist.js";
 import { getMetrics } from "./metrics.js";
 import { MurmurHash3_64 } from "./murmurhash3.js";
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
*Oh great*, this fails both on the bots and *locally* in Google Chrome even when testing against the `master` branch (but not if the document is loaded/rendered "in order").
This all suggests a (somewhat) serious bug in Google Chrome, since the rejected Promise from `PartialEvaluator.translateFont` *somehow* isn't caught by https://github.com/mozilla/pdf.js/blob/a392082e302366a25a17d1ba35551a2acd9e5195/src/core/evaluator.js#L1153-L1182 and we thus end up propagating the error to the main-thread.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

Issue #12963, opened 2021-02-06
Warning: Unimplemented widget field type "null", falling back to base field type.


evlgromov:
When uploading a pdf file, I get Warning: Unimplemented widget field type "null", falling back to base field type at pdf.worker.js. The file is not loading.
I'm using the chrome browser. All other files are processed fine, but this one is not. Help solve the problem.
[usn2020.pdf](https://github.com/mozilla/pdf.js/files/5936611/usn2020.pdf)

calixteman:
Annotation 19R is equal to its parent so infinite loop.

evlgromov:
> Annotation 19R is equal to its parent so infinite loop.

how can i solve the problem

calixteman:
I don't know.
This file contains a lot of objects which have references to themselves (modulo a chain).

PR #12964, opened 2021-02-06
Avoid infinite loop when getting annotation field name

calixteman:
- aims to fix issue #12963;
 - use a Set to track already visited objects;
 - remove the loop limit in getInheritableProperty and use a Set too.


src/core/annotation.js
```diff
@@ -1066,14 +1066,21 @@ class WidgetAnnotation extends Annotation {
     }
 
     let loopDict = dict;
+    const visited = new Set([this.data.id]);
```
Snuffleupagus, line 1069: First of all, please use a `RefSet` instead. (Ideally the Dictionary instances would actually use a `Ref` directly, rather than their string-format, so let's avoid future surprises here.)

Secondly, `this.data.id` isn't guaranteed to contain the string-format of a `Ref`; please see https://github.com/mozilla/pdf.js/blob/a392082e302366a25a17d1ba35551a2acd9e5195/src/core/annotation.js#L82, hence you should use `dict.objId` instead.

src/core/core_utils.js
```diff
@@ -90,11 +89,13 @@ function getInheritableProperty({
   getArray = false,
   stopWhenFound = true,
 }) {
-  const LOOP_LIMIT = 100;
-  let loopCount = 0;
   let values;
+  const visited = new Set();
```
Snuffleupagus, line 93: Again, use `RefSet` please.

src/core/annotation.js
```diff
@@ -1066,14 +1066,21 @@ class WidgetAnnotation extends Annotation {
...
     while (loopDict.has("Parent")) {
       loopDict = loopDict.get("Parent");
-      if (!isDict(loopDict)) {
+      if (!isDict(loopDict) || visited.has(loopDict.objId)) {
```
Snuffleupagus, line 1072:
```suggestion
      if (
        !(loopDict instanceof Dict) ||
        (loopDict.objId && visited.has(loopDict.objId))
      ) {
```

src/core/core_utils.js
```diff
@@ -90,11 +89,13 @@ function getInheritableProperty({
...
 
-  while (dict) {
+  while (isDict(dict) && !visited.has(dict.objId)) {
```
Snuffleupagus, line 95:
```suggestion
  while (dict instanceof Dict && !(dict.objId && visited.has(dict.objId))) {
```

src/core/core_utils.js
```diff
@@ -105,10 +106,6 @@ function getInheritableProperty({
       }
       values.push(value);
     }
-    if (++loopCount > LOOP_LIMIT) {
-      warn(`getInheritableProperty: maximum loop count exceeded for "${key}"`);
-      break;
-    }
```
Snuffleupagus, line 111: Given that it's *possible* for a Dictionary to not have an `objId` property, this code could still be useful to avoid an infinite loop in the general case. Hence, please restore this code and the related unit-test!

Snuffleupagus (changes requested):

Snuffleupagus (changes requested):
Also, this needs a reference test of the `eq` and `annotations` kind (testing only the *first* page should suffice).

src/core/core_utils.js
calixteman: Yes I thought about that and it's why I added the `if (dict.objId)` above.
My assertion (tell me if I'm wrong) is that we can have no objId when the parent is in the object: ` << /Parent << /Parent << ....` but in such a case we can't have an infinite loop, so the only way to have an infinite loop is to have objId.

src/core/core_utils.js
calixteman: Why not `isDict` ? (I don't care but just to understand)

src/core/annotation.js
calixteman: It won't work if `objId` is null which should be the case if the parent is inside the object (no ref).

src/core/core_utils.js
calixteman: And what happen if a pdf has an object with more than `LOOP_LIMIT` ancestors (which should be pretty rare I guess) ?

src/core/annotation.js
calixteman: I don't see any advantage to use a `RefSet` which seems to be useful to store a `Ref`. Here `objId` is a string or null (correct ?).

Snuffleupagus: Sorry about the delay here, I got sidetracked with other things lately!

---

The fact that `Dictionary`-instance uses a string for the `objId`, rather than simply storing the `Ref` directly, is the result of some *really* old implementation decisions. Hence we might, at some point, actually want to fix that and then it's simpler if all code actually handles (what should be) `Ref`s consistently.

src/core/core_utils.js
Snuffleupagus: Sorry, it probably doesn't matter here given that this code shouldn't be that hot.

---

However, in e.g. some parts of the `Parser`/`Lexer` and for some complex PDF documents, there's been cases where inlining these kind of checks and thus being guaranteed to skip unnecessary function calls have helped performance.

src/core/core_utils.js
Snuffleupagus: > My assertion (tell me if I'm wrong) is that we can have no objId when the parent is in the object

Oh, you're probably correct here. The kind of structure required for an infinite loop to occur ought to be essentially impossible to create with only "direct" objects.

> And what happen if a pdf has an object with more than `LOOP_LIMIT` ancestors (which should be pretty rare I guess) ?

We simply broke out of the loop, since the intention was simply to avoid permanently hanging the worker-thread.

timvandermeij:
We'll have another look at this once the comments above are addressed, but the idea here looks good.

calixteman:
I changed `Set` for `RefSet`. But I didn't add back the `LOOP_LIMIT` stuff because it's impossible to have an infinite loop now.
Anyway if for perf reason or whatever, I'm fine to add it back, just tell me.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

PR #12969, opened 2021-02-07
[Firefox] Unblock load event on error.

emilio:
We can get this when loading a PDF with no data, etc.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -721,6 +721,7 @@ const PDFViewerApplication = {
         PDFViewerApplication.open(file, args);
       },
       onError(err) {
+        PDFViewerApplication._unblockDocumentLoadEvent();
         PDFViewerApplication.l10n
           .get(
             "loading_error",


Snuffleupagus:
While this will obviously work for one *particular* error path, having looked more thoroughly at the code (which I should have done beforehand, sorry!) I'm afraid that there's still other ways that errors can occur.

Generally speaking we probably need to do this before *most* (but not all) `PDFViewerApplication.error` calls, however having to add this in multiple places seems quite error prone to me.
It might be possible to fix all the relevant cases at once, by doing a small bit of refactoring, so unless you need to landed *immediately* please given me the rest of the evening to look into this.

emilio:
Sure, sounds good, there's no rush.

Snuffleupagus:
Please see PR #12970, which should help to ensure that regardless of where a document loading/parsing error occurs we'll now always unblock the "load" event.

emilio:
Sounds good, let's close in favor of that.

PR #12970, opened 2021-02-07
Split `PDFViewerApplication.error` into two methods, for PDF document loading/parsing errors vs other errors (PR 11647 follow-up)

Snuffleupagus:
With these changes, we can easily unblock the "load" event regardless of where an error occurred.

/cc @emilio

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -728,7 +728,7 @@ const PDFViewerApplication = {
             "An error occurred while loading the PDF."
           )
           .then(msg => {
-            PDFViewerApplication.error(msg, err);
+            PDFViewerApplication._documentError(msg, err);
           });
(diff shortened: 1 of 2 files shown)


web/app.js
```diff
@@ -812,6 +812,8 @@ const PDFViewerApplication = {
    *                      destruction is completed.
    */
   async close() {
+    this._unblockDocumentLoadEvent();
```
Snuffleupagus, line 815: While this *really shouldn't* matter for the Firefox built-in PDF Viewer use-case, since only the GENERIC viewer has openFile-support, in nonetheless can't hurt to ensure that closing a partially loaded document will actually fire the "load" event.

PR #12971, opened 2021-02-07
Use DOM hidden property instead of attribute methods

nt1m:
`.setAttribute("hidden", "true")` or `.setAttribute("hidden", true)` are misleading, because `.setAttribute("hidden", "false")` also hides the element. This is because the HTML UA sheet matches `[hidden]`.

`.setAttribute("hidden", "hidden");` or `.hidden = true` would be correct ways to do it. The latter is more future-proof though.

The reason I'm doing this is because this showed up in the XUL boolean attribute matches in mozilla-central (XUL considers `hidden=true` as correct unlike HTML, and I'm looking into fixing that).

### DIFF
--- a/examples/mobile-viewer/viewer.js
+++ b/examples/mobile-viewer/viewer.js
@@ -135,7 +135,7 @@ var PDFViewerApplication = {
    */
   close() {
     var errorWrapper = document.getElementById("errorWrapper");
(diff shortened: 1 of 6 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #12973, opened 2021-02-09
Use `Math.hypot`, instead of `Math.sqrt` with manual squaring

Snuffleupagus:
When the PDF.js project started `Math.hypot` didn't exist yet, and until recently we still supported browsers (IE 11) without a native `Math.hypot` implementation; please see this compatibility information: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot#browser_compatibility

Furthermore, somewhat recently there were performance improvements of `Math.hypot` in Firefox; see https://bugzilla.mozilla.org/show_bug.cgi?id=1648820

Finally, this patch also replaces a couple of multiplications with the exponentiation operator.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -1364,7 +1364,7 @@ const LabCS = (function LabCSClosure() {
   function fn_g(x) {
     let result;
     if (x >= 6 / 29) {
-      result = x * x * x;
+      result = x ** 3;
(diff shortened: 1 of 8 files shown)


calixteman (approved):
LGTM.
Thanks for doing this.

PR #12975, opened 2021-02-09
Remove the `contentmenu` usage, from PresentationMode, since it's no longer working

Snuffleupagus:
This feature was Firefox-specific, and it's now been removed from the HTML specification and it's disabled by default starting with Firefox 85. Hence it seems completely unnecessary to keep this code in the default viewer.

Please refer to https://groups.google.com/g/mozilla.dev.platform/c/tc11BCenm2c and the resources that it links to.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #12978, opened 2021-02-09
[api-minor] Rename `-es5` to `-legacy`, to reduce confusion over what's actually supported  (issue 12976)

### LINKED ISSUE #12976: Rename es5 version
This is very confusing as seen in #12974 or #12906. My suggestion would be "legacy."

Snuffleupagus:
*Please note that this will also require some edits of the Wiki.*

Fixes #12976


docs/contents/getting_started/index.md
```diff
@@ -54,8 +54,8 @@ Before downloading PDF.js please take a moment to understand the different layer
       Includes the generic build of PDF.js and the viewer.
     </p>
     <span class="GROUP_CLASS">
-      <a type="button" class="btn btn-primary" href="https://github.com/mozilla/pdf.js/releases/download/vSTABLE_VERSION/pdfjs-STABLE_VERSION-es5-dist.zip">Stable (vSTABLE_VERSION)</a>
-      <a type="button" class="btn btn-warning HIDDEN_CLASS" href="https://github.com/mozilla/pdf.js/releases/download/vBETA_VERSION/pdfjs-BETA_VERSION-es5-dist.zip">Beta (vBETA_VERSION)</a>
+      <a type="button" class="btn btn-primary" href="https://github.com/mozilla/pdf.js/releases/download/vSTABLE_VERSION/pdfjs-STABLE_VERSION-legacy-dist.zip">Stable (vSTABLE_VERSION)</a>
```
timvandermeij, line 57: This might be problematic given that these archives have already been published with the `-es5` name. We might have to make this conditional, like we did before for the introduction of the ES5 archives, otherwise these links will be broken.

timvandermeij:
Note: when this gets merged, also merge https://github.com/mozilla/botio-files-pdfjs/pull/29.

docs/contents/getting_started/index.md
Snuffleupagus: That's a very good point, and something that I'd completely forgot about; thank you!

I've pushed an updated patch which, assuming that I didn't mess up the logic of course, should handle this automatically for us and thus reduce the number of manual edits required in this file.

timvandermeij:
Looks good to me too. Thanks!

PR #12979, opened 2021-02-10
XFA - Add support for prototypes

calixteman:
- specifications: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=225&zoom=auto,-207,784
 - add a clone method on nodes in order to be able to clone a proto;
 - support ids in template namespace;
 - prevent from cycle when applying protos.


brendandahl (approved):
This is giving me flashbacks to XUL overlays.

PR #12981, opened 2021-02-11
Collect the l10n error/warning message lookup, in `web/app.js`, in a new helper method

Snuffleupagus:
Some of the localization strings (e.g. "loading_error") are repeated multiple times throughout the `web/app.js` file, which means that we need to duplicate the fallback strings as well. Furthermore, the signature of the `IL10n.get` method makes the call-sites quite verbose.

By adding a new helper method, in `PDFViewerApplication`, we're able to gather the localization fallback strings in one central spot in `web/app.js` and also make the lookup of the error/warning messages more compact.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -702,13 +702,13 @@ const PDFViewerApplication = {
       throw new Error("Not implemented: initPassiveLoading");
     }
     this.externalServices.initPassiveLoading({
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks much less verbose overall. Thanks!

PR #12984, opened 2021-02-11
Replace a few `new Date().getTime()` instances with `Date.now()`

Snuffleupagus:
The former format is not only more verbose, but it's also *slightly* less efficient since it creates a new `Date` object.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #2719, opened 2013-02-13
Implement progressive loading of PDFs

mduan:
This implements progressive loading of the PDF using an implementation similar to that described in #1108.

**Note:** This is still a work in progress, so it's incomplete and a bit messy. It is **NOT** ready to be merged.

Some of the things that still have to be done are:
- [x] Integrate w/ extension
- [x] For addon: fallback to fetching entire PDF if range requests are not supported
- [x] For viewer: fallback to fetching entire PDF if range requests are not supported
- [x] When there are no PDF chunks that are actively being requested, continue fetching PDF chunks sequentially
- [x] Fetching the current page first when the user changes pages
- [x] Optimization of parsing XRef tables/streams
- [x] Optimization of fetching of page tree
- [x] <del>Possibly more optimizations specifically for linearized PDFs</del>
- [x] Code cleanup
- [x] Unit tests, regression tests, etc...

PR #12986, opened 2021-02-12
Remove the unused "loading_error_indicator" l10n string (PR 2719 follow-up)

Snuffleupagus:
This l10n string has been completely unused ever since PR #2719, which is now *eight* years ago.


timvandermeij (approved):

timvandermeij:
Thank you!

Issue #12142, opened 2020-07-30
addFakeSpaces can easily trigger out-of-memory and cause Chrome to crash


igorbednar:
Configuration:
- Chrome 84.0.4147.105
- Windows 10 1909
- PDF.js version: latest, reproduces in demo website with version 2.6.160
- Is a browser extension: no

Steps to reproduce the problem:
1.  Load attached document.pdf in https://mozilla.github.io/pdf.js/web/viewer.html
[document.pdf](https://github.com/mozilla/pdf.js/files/5001194/document.pdf)

Expected:
Pdf should be displayed
Actual:
![image](https://user-images.githubusercontent.com/20933823/88923084-d1be3200-d279-11ea-9d8d-3e9dee4512ba.png)

Chrome tab crashes with "Aw snap" error

Rob--W:
Minimal reproduction: [setCharSpacingInsane.pdf](https://github.com/mozilla/pdf.js/files/5004175/setCharSpacingInsane.pdf)



Chrome crashes because the page runs out of memory.
Firefox does not crash but tries to use lots of memory (Gigabytes of memory, high CPU usage, lots of time spent on garbage collection, eventually an "out of memory" error in getTextContent in the console).
Here is the profiler data: https://profiler.firefox.com/public/zsq9tz8bssr3wdky41djv0caafkv2kxntns91r0/flame-graph/?globalTrackOrder=0-1-2-3-4&localTrackOrderByPid=143552-0~143652-0~&thread=3&v=5

According to the profile from Firefox (and the call stack of Chrome's crash), something attempts to repeatedly grow an array.
I also see that lots of time is spent in `addFakeSpaces` :
https://github.com/mozilla/pdf.js/blob/403816040ef64ae17758439948771c2920f96ac4/src/core/evaluator.js#L1999-L2011

I attached a debugger and saw that `width = 2147483649.949` and `fakeSpaceMin = 6.831`, i.e. `fakeSpaces = 314373247` = 300M+. This is a LOT, and easily explains the crash.

This large value for `width` originates from `textContentItem.lastAdvanceWidth` (as a very large negative value) @ https://github.com/mozilla/pdf.js/blob/403816040ef64ae17758439948771c2920f96ac4/src/core/evaluator.js#L2121
from `width` @ https://github.com/mozilla/pdf.js/blob/403816040ef64ae17758439948771c2920f96ac4/src/core/evaluator.js#L1989
[...]

Issue #7226, opened 2016-04-20
Textlayer is not correctly on top


mangei:
## Link to PDF file (or attach file here):

[testdocument.pdf](https://github.com/mozilla/pdf.js/files/227533/testdocument.pdf)
## Configuration:
- Web browser and its version: Firefox 45.0.2, Chrome  49.0.2623
- Operating system and its version: Microsoft Windows 10
- PDF.js version: 1.4.175
- Is an extension: usage as a library in a web application or normal viewer in Firefox
## Steps to reproduce the problem:
1. Open Document (see attached testdocument.pdf)
2. Select a paragraph
## What is the expected behavior? (add screenshot)

The text should be correctly on top, such that e.g. selecting works fine.

**Adobe Acrobat Reader DC**
![2016-04-20 103451](https://cloud.githubusercontent.com/assets/1218033/14668660/ae737170-06e5-11e6-9b14-412fcbc2b723.png)
## What went wrong? (add screenshot)

The text overlay is not correctly positioned and scaled.

**PDF.js in Firefox**
![2016-04-20 103326](https://cloud.githubusercontent.com/assets/1218033/14668671/bc64d710-06e5-11e6-8dc3-0eb1601274ed.png)

**Web application**
![2016-04-20 104803](https://cloud.githubusercontent.com/assets/1218033/14668658/aa95abcc-06e5-11e6-9e58-66576630c4c5.png)

mangei:
related to #5558, #5387, #3713

timvandermeij:
The text layer alignment got fixed by #12896 and other work, but the lines are still too short.

timvandermeij:
Fixed by #13257.

PR #12989, opened 2021-02-13
[api-minor] Fix the way to chunk the strings

### LINKED ISSUE #7310: missing all spaces on page 1 (text selection)

Configuration:
- Web browser and its version: Chrome 51.0.2704.36 beta (64-bit)
- Operating system and its version: Mac OS X El Capitan 10.11.3 (15D21)
- PDF.js version: online viewer (May 10, 2016)
- Is an extension:

1. copy text from page 1
2. paste somewhere else: all words touch, without any space

need to add spaces between words.
[jansen_public_queries.pdf](https://github.com/mozilla/pdf.js/files/257334/jansen_public_queries.pdf)

no spaces were found between words.

calixteman:
- Improve chunking in order to fix some bugs where the spaces aren't here;
 - add some breaks in order to get lines. 
 - aims to fix: #7310, #7226, #12142 and probably few others.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1985,7 +1985,8 @@ class PartialEvaluator {
     resources = resources || Dict.empty;
     stateManager = stateManager || new StateManager(new TextState());
 
-    var WhitespaceRegexp = /\s/g;
+    const WhitespaceRegexp = /\s/g;
+    const AllWhitespaceRegexp = /^\s+$/g;
 
     var textContent = {
       items: [],
@@ -1997,23 +1998,39 @@ class PartialEvaluator {
       width: 0,
       height: 0,
       vertical: false,
+      lastCharSize: 0,
       lastAdvanceWidth: 0,
       lastAdvanceHeight: 0,
       textAdvanceScale: 0,
       spaceWidth: 0,
-      fakeSpaceMin: Infinity,
-      fakeMultiSpaceMin: Infinity,
-      fakeMultiSpaceMax: -0,
-      textRunBreakAllowed: false,
+      spaceInFlowMin: 0,
+      spaceInFlowMax: 0,
+      trackingSpaceMin: Infinity,
       transform: null,
       fontName: null,
+      isSpace: false,
+      hasEOL: false,
     };
-    var SPACE_FACTOR = 0.3;
-    var MULTI_SPACE_FACTOR = 1.5;
-    var MULTI_SPACE_FACTOR_MAX = 4;
+
+    // Used in addFakeSpaces.
+    // wsw stands for whitespace width.
+
+    // A white <= wsw * TRACKING_SPACE_FACTOR is a tracking space
+    // so it doesn't count as a space.
+    const TRACKING_SPACE_FACTOR = 0.3;
+
+    // A white with a width in [wsw * MIN_FACTOR; wsw * MAX_FACTOR]
+    // is a space which will be inserted in the current flow of words.
+    // If the width is outside of this range then the flow is broken
+    // (which means a new span in the text layer).
+    // It's useful to adjust the best as possible the span in the layer
+    // to what is displayed in the canvas.
+    const SPACE_IN_FLOW_MIN_FACTOR = 0.3;
+    const SPACE_IN_FLOW_MAX_FACTOR = 1.1;
 
     var self = this;
     var xref = this.xref;
+    const showSpacedTextBuffer = [];
 
     // The xobj is parsed iff it's needed, e.g. if there is a `DO` cmd.
     var xobjs = null;
@@ -2024,24 +2041,10 @@ class PartialEvaluator {
 
     var textState;
 
-    function ensureTextContentItem() {
-      if (textContentItem.initialized) {
-        return textContentItem;
-      }
-      var font = textState.font;
-      if (!(font.loadedName in seenStyles)) {
-        seenStyles[font.loadedName] = true;
-        textContent.styles[font.loadedName] = {
-          fontFamily: font.fallbackName,
-          ascent: font.ascent,
-          descent: font.descent,
-          vertical: font.vertical,
-        };
-      }
-      textContentItem.fontName = font.loadedName;
-
+    function getCurrentTextTransform() {
       // 9.4.4 Text Space Details
-      var tsm = [
+      const font = textState.font;
+      const tsm = [
         textState.fontSize * textState.textHScale,
         0,
         0,
@@ -2061,11 +2064,28 @@ class PartialEvaluator {
         }
       }
 
-      var trm = Util.transform(
+      return Util.transform(
         textState.ctm,
         Util.transform(textState.textMatrix, tsm)
       );
-      textContentItem.transform = trm;
+    }
+
+    function ensureTextContentItem() {
+      if (textContentItem.initialized) {
+        return textContentItem;
+      }
+      var font = textState.font;
+      if (!(font.loadedName in seenStyles)) {
+        seenStyles[font.loadedName] = true;
+        textContent.styles[font.loadedName] = {
+          fontFamily: font.fallbackName,
+          ascent: font.ascent,
+          descent: font.descent,
+          vertical: font.vertical,
+        };
+      }
+      textContentItem.fontName = font.loadedName;
+      const trm = (textContentItem.transform = getCurrentTextTransform());
       if (!font.vertical) {
         textContentItem.width = 0;
         textContentItem.height = Math.hypot(trm[2], trm[3]);
@@ -2082,24 +2102,24 @@ class PartialEvaluator {
       );
       const scaleCtmX = Math.hypot(textState.ctm[0], textState.ctm[1]);
       textContentItem.textAdvanceScale = scaleCtmX * scaleLineX;
-      textContentItem.lastAdvanceWidth = 0;
-      textContentItem.lastAdvanceHeight = 0;
+      textContentItem.lastAdvanceWidth = textContentItem.lastAdvanceWidth || 0;
+      textContentItem.lastAdvanceHeight =
+        textContentItem.lastAdvanceHeight || 0;
+      textContentItem.lastCharSize = textContentItem.lastCharSize || 0;
 
       var spaceWidth = (font.spaceWidth / 1000) * textState.fontSize;
       if (spaceWidth) {
         textContentItem.spaceWidth = spaceWidth;
-        textContentItem.fakeSpaceMin = spaceWidth * SPACE_FACTOR;
-        textContentItem.fakeMultiSpaceMin = spaceWidth * MULTI_SPACE_FACTOR;
-        textContentItem.fakeMultiSpaceMax = spaceWidth * MULTI_SPACE_FACTOR_MAX;
-        // It's okay for monospace fonts to fake as much space as needed.
-        textContentItem.textRunBreakAllowed = !font.isMonospace;
+        textContentItem.trackingSpaceMin = spaceWidth * TRACKING_SPACE_FACTOR;
+        textContentItem.spaceInFlowMin = spaceWidth * SPACE_IN_FLOW_MIN_FACTOR;
+        textContentItem.spaceInFlowMax = spaceWidth * SPACE_IN_FLOW_MAX_FACTOR;
       } else {
         textContentItem.spaceWidth = 0;
-        textContentItem.fakeSpaceMin = Infinity;
-        textContentItem.fakeMultiSpaceMin = Infinity;
-        textContentItem.fakeMultiSpaceMax = 0;
-        textContentItem.textRunBreakAllowed = false;
+        textContentItem.trackingSpaceMin = Infinity;
       }
+
+      textContentItem.isSpace = false;
+      textContentItem.hasEOL = false;
 
       textContentItem.initialized = true;
       return textContentItem;
@@ -2119,17 +2139,20 @@ class PartialEvaluator {
     }
 
     function runBidiTransform(textChunk) {
-      var str = textChunk.str.join("");
-      var bidiResult = bidi(str, -1, textChunk.vertical);
+      const text = textChunk.str.join("");
+      const bidiResult = bidi(text, -1, textChunk.vertical);
+      const str = normalizeWhitespace
+        ? replaceWhitespace(bidiResult.str)
+        : bidiResult.str;
       return {
-        str: normalizeWhitespace
-          ? replaceWhitespace(bidiResult.str)
-          : bidiResult.str,
+        str,
         dir: bidiResult.dir,
         width: textChunk.width,
         height: textChunk.height,
         transform: textChunk.transform,
         fontName: textChunk.fontName,
+        isSpace: AllWhitespaceRegexp.test(str),
+        hasEOL: textChunk.hasEOL,
       };
     }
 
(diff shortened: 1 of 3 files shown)


src/core/evaluator.js
```diff
@@ -1997,23 +1997,21 @@ class PartialEvaluator {
...
-      fakeMultiSpaceMax: -0,
-      textRunBreakAllowed: false,
       transform: null,
       fontName: null,
     };
-    var SPACE_FACTOR = 0.3;
-    var MULTI_SPACE_FACTOR = 1.5;
-    var MULTI_SPACE_FACTOR_MAX = 4;
+    const SPACE_ERROR = 0.01;
+    var SPACE_FACTOR = 0.25;
```
timvandermeij, line 2010: This can be `const`.

Snuffleupagus:
*Looking briefly at the test results, there's one thing that jumps out as a potential performance problem and will thus require careful analysis:*
By adding *separate* `<span>`s for the spaces, the number of DOM-elements will thus increase *a lot* in many cases (even e.g. `tracemonkey.pdf`) which can have quite undesirable effects on both memory usage and e.g. scrolling performance in many documents.

Please note that years ago, before the work to try and coalesce individual `<span>`s (or `<div>s` at that time) into text runs, overall performance often suffered considerably in many documents because of the sheer number of DOM elements.

calixteman:
Yeah I saw the discussion about perf and memory.
IIRC, there was one letter by span, here there is at least one word (in general) by span and depending of the font sometimes some sentences are in the span.
But this way the text selection is a way better because most of the time the span fits the box in the canvas  and there are exceptions when spans contains several words.
So I guess it's a compromise, I'm fine to revert to the default behaviour, but maybe Firefox is better than it was (no opinion on that subject) about scroll perf or memory: I think it's worth having a look on that stuff.
Anyway on some documents text selection is bad, maybe we could try to find some heuristics to decide of a strategy, maybe it's possible to guess something from metrics for the canvas font and ones from text layer font.

Snuffleupagus:
> But this way the text selection is a way better because most of the time the span fits the box in the canvas and there are exceptions when spans contains several words.

Generally speaking, the `textLayer` will never be *perfect* and most of the time we simply want it to be "good enough" when compared to the canvas.
Would it perhaps be possible to e.g. tweak the constants ever so slightly, to make the new heuristics a bit less aggressive?

---

However, I think that it'd be very good to have some actual data here. Hence can you please provide some statistics, on a per page basis for the affected test-cases, that compared the number of `textLayer` elements with/without this patch?

calixteman:
> > But this way the text selection is a way better because most of the time the span fits the box in the canvas and there are exceptions when spans contains several words.
> 
> Generally speaking, the `textLayer` will never be _perfect_ and most of the time we simply want it to be "good enough" when compared to the canvas.
> Would it perhaps be possible to e.g. tweak the constants ever so slightly, to make the new heuristics a bit less aggressive?
> 

Playing with the new constant `SPACE_ERROR` is enough to change the number of spans. Currently with a value of `0.01` then only spaces with width in `[0.99*font_space_width, 1.01*font_space_width]` are rendered with a true whitespace and the others with an extensible span.
So changing this value to 1 for (likely too aggressive) remove almost all the extensible whites.

TamarLdad:
hi we are really looking fwd for this fix. Thanks for working on it! Do you think it will  be merged soon?

calixteman:
@brendandahl hacked a talos test used to test scrolling performance to display the results
```diff
diff --git a/testing/talos/talos/pageloader/chrome/tscroll.js b/testing/talos/talos/pageloader/chrome/tscroll.js
--- a/testing/talos/talos/pageloader/chrome/tscroll.js
+++ b/testing/talos/talos/pageloader/chrome/tscroll.js
@@ -6,16 +6,18 @@
 //       unprivileged code.
 // - Please make sure that any changes apply cleanly to all use cases.
 
 function testScroll(target, stepSize, opt_reportFunc, opt_numSteps) {
   var win;
   if (target == "content") {
     target = content.wrappedJSObject;
     win = content;
+    target = win.document.getElementById("viewerContainer");
+    dump(`>>> target=${target}`);
   } else {
     win = window;
   }
 
   var result = {
     names: [],
     values: [],
   };
```
I opened: `viewer.html?file=/test/pdfs/tracemonkey.pdf#page=1&zoom=5` to have all the pages.
With master:
`2436` spans are created
Scroll perf: values are `1 / fps`
```log
15:59:34     INFO -  PID 17278 | [#0] tracemonkey.pdf  Cycles:12  Average:3.79  Median:3.92  stddev:0.61 (15.5%)  stddev-sans-first:0.34
15:59:34     INFO -  PID 17278 | Values: 2.1  4.3  3.7  3.6  3.9  3.9  3.9  4.4  4.3  4.3  3.5  3.6
```

and with this PR:
`23747` spans are created
Scroll perf: values are `1 / fps`
```log
16:03:09     INFO -  PID 18348 | [#0] tracemonkey.pdf  Cycles:12  Average:8.51  Median:8.58  stddev:0.62 (7.2%)  stddev-sans-first:0.59
16:03:09     INFO -  PID 18348 | Values: 7.7  8.2  8.2  9.1  9.0  8.9  8.1  8.3  7.8  9.6  9.2  8.1
```

So there's a memory inflation but as far as I can tell the text in the layer a way better fits the one on the canvas and the scroll xp is doesn't change so much (at least on my machine).

src/display/text_layer.js
```diff
@@ -679,8 +666,11 @@ const renderTextLayer = (function renderTextLayerClosure() {
         }
         textDiv.style.transform = transform;
       }
-      this._textDivProperties.set(textDiv, textDivProperties);
+
       this._container.appendChild(textDiv);
+      if (textDivProperties.hasEOL) {
+        textDiv.appendChild(this._document.createElement("br"));
```
Snuffleupagus, line 672: Why not just `textDiv.appendChild(document.createElement("br"))` instead, since this is only creating a *simple* element (which isn't accessed further)?

src/core/evaluator.js
```diff
@@ -2304,29 +2347,46 @@ class PartialEvaluator {
...
+                    const isSplited = addFakeSpaces(advance);
+                    textState.translateTextLineMatrix(args[0], args[1]);
+                    if (isSplited) {
+                      textState.textMatrix = textState.textLineMatrix.slice();
+                    } else {
+                      textContentItem.width += advance;
+                    }
+                    break;
+                  }
+                } else if (advance < -2 * textContentItem.lastCharSize) {
```
Snuffleupagus, line 2386: Another magical value (i.e. `-2`), please at least add a comment explaining what it is/does.

src/display/text_layer.js
```diff
@@ -133,7 +127,8 @@ const renderTextLayer = (function renderTextLayerClosure() {
     const textDivProperties = {
       angle: 0,
       canvasWidth: 0,
-      isWhitespace: false,
+      hasEOL: geom.hasEOL,
+      isWhitespace: geom.isSpace,
```
Snuffleupagus, line 131: Is the `isWhitespace` property still necessary?

I know that it's being used in `expandTextDivs` below, however I believe that with the other changes in this patch you'll now need to *also* expand the spaces for things to work as intended (based on the test results).
Hence the `isWhitespace` property, and *all* of its uses, should probably be removed now!?

src/core/evaluator.js
```diff
@@ -2407,14 +2469,16 @@ class PartialEvaluator {
...
 
+            let lastAdvance = 0;
+            const spaceFactor =
+              ((textState.font.vertical ? 1 : -1) * textState.fontSize) / 1000;
+            const elements = args[0];
+            for (let i = 0, ii = elements.length; i < ii - 1; i++) {
+              const item = elements[i];
+              if (typeof item === "string") {
+                showSpacedTextBuffer.push(item);
+              } else if (isNum(item) && item !== 0) {
```
Snuffleupagus, line 2481: Nit: This is obviously pre-existing code, but I can't help wonder if replacing the `isNum` call with an inline `typeof time === "number"` instead wouldn't be a tiny bit more efficient!?

src/core/evaluator.js
```diff
@@ -1997,23 +1998,27 @@ class PartialEvaluator {
...
+      isSpace: false,
+      hasEOL: false,
     };
-    var SPACE_FACTOR = 0.3;
-    var MULTI_SPACE_FACTOR = 1.5;
-    var MULTI_SPACE_FACTOR_MAX = 4;
+
+    // Used in addFakeSpaces
+    // A white = whiteSpace +/- (100*SPACE_ERROR)% is a white space.
+    const SPACE_ERROR = 0.1;
```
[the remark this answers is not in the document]
calixteman: `0.2` doesn't change the order of magnitude.
Anyway I split the `SPACE_ERROR` value into two: a lower and an upper bounds.
Setting `SPACE_IN_FLOW_MIN_FACTOR` to `0.25` leads to `2901` spans for all the document which is almost what we've right now.
As far as I understood, "enhanced" text-selection is mainly a chrome feature, correct ? if yes then we can tune the constant above to avoid perf regression when enhanced text-selection is enabled.

src/display/text_layer.js
calixteman: I checked with `tracemonkey.pdf` but it doesn't seem necessary to do it.

src/core/evaluator.js
```diff
@@ -1997,23 +1998,39 @@ class PartialEvaluator {
...
       lastAdvanceHeight: 0,
       textAdvanceScale: 0,
       spaceWidth: 0,
-      fakeSpaceMin: Infinity,
-      fakeMultiSpaceMin: Infinity,
-      fakeMultiSpaceMax: -0,
-      textRunBreakAllowed: false,
+      spaceInFlowMin: 0,
+      spaceInFlowMax: 0,
+      kerningSpaceMin: Infinity,
```
Snuffleupagus, line 2008: Doesn't this need to be `trackingSpaceMin` instead, given the latest changes below?

timvandermeij:
It looks like there are some regressions, such as `firefox-issue6387-text-page1` where the `-50` text is moved way too much to the left.

calixteman:
> It looks like there are some regressions, such as `firefox-issue6387-text-page1` where the `-50` text is moved way too much to the left.

Fixed.

src/core/evaluator.js
```diff
@@ -2143,80 +2160,122 @@ class PartialEvaluator {
...
+        textChunk.str.push(glyphUnicode);
+
+        const glyphWidth =
+          font.vertical && glyph.vmetric ? glyph.vmetric[0] : glyph.width;
 
-        var glyphUnicode = glyph.unicode;
-        var NormalizedUnicodes = getNormalizedUnicodes();
-        if (NormalizedUnicodes[glyphUnicode] !== undefined) {
-          glyphUnicode = NormalizedUnicodes[glyphUnicode];
+        let t = glyphWidth * scale;
```
Snuffleupagus, line 2189: Can we, pretty please, avoid one-letter variable names for improved readability?

src/core/evaluator.js
```diff
@@ -2143,80 +2160,122 @@ class PartialEvaluator {
...
+        if (!font.vertical) {
+          t *= textState.textHScale;
+          textState.translateTextMatrix(t, 0);
+        } else {
+          textState.translateTextMatrix(0, t);
         }
-        glyphUnicode = reverseIfRtl(glyphUnicode);
+        t = font.vertical ? Math.abs(t) : t;
```
Snuffleupagus, line 2196: Can the `Math.abs(...)`-part be moved into the `else`-branch *just* above instead?

marcenacp:
Really interesting PR. When will it merged? :)

calixteman:
My last update is just to set `SPACE_IN_FLOW_MIN_FACTOR` to the same value as `TRACKING_SPACE_FACTOR` so this way we've almost one span by line and not one span by word.
This way, there still are some issues between canvas and text layer but at least almost all the issues with text selection (from a content point of view) should be fixed.
And we can discuss of this solution (one span by word) or another one (based on svg for example) in an other issue/patch.

src/core/evaluator.js
```diff
@@ -2109,27 +2129,24 @@ class PartialEvaluator {
...
       // alignment issues between the textLayer and the canvas if the text
       // contains e.g. tabs (fixes issue6612.pdf).
-      var i = 0,
-        ii = str.length,
-        code;
-      while (i < ii && (code = str.charCodeAt(i)) >= 0x20 && code <= 0x7f) {
-        i++;
-      }
-      return i < ii ? str.replace(WhitespaceRegexp, " ") : str;
+      return str.replace(BeforeSpaceRegexp, " ");
```
Snuffleupagus, line 2132: Wait, is this actually equivalent to the old code though?

The old `WhitespaceRegexp = /\s/g` includes *a lot* more things than the new `BeforeSpaceRegexp` does, according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet:
> Matches a single white space character, including space, tab, form feed, line feed, and other Unicode spaces. Equivalent to [ \f\n\r\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]. For example, /\s\w*/ matches " bar" in "foo bar".

I'd suggest reverting this change, to avoid unforeseen regressions!

Snuffleupagus (changes requested):

cetinsert:
> #### From: Bot.io (Windows)
> # Failed
> Full output at http://3.101.106.178:8877/72f715db5b9b1fb/output.txt
> 
> Total script time: 28.99 mins
> 
> * **Font tests:** Passed
> * **Unit tests:** Passed
> * **Integration Tests:** Passed
> * **Regression tests:** FAILED
> 
> Image differences available at: http://3.101.106.178:8877/72f715db5b9b1fb/reftest-analyzer.html#web=eq.log

For many pages like **chrome-preistabelle-text-page1**, **Test** creates a lot of non-image(canvas)-backed ghost space-only text layer elements compared to **Reference**, the overall match seems worse.

Also a 10x DOM explosion may not seem too much for a single page but when viewing

* multiple pages on a large 4K/8K screen at once or
* on a small device for using browser-integrated search

that jump from 2K to 20K is bound to cause issues with performance.

Overall, I am not convinced this is worth pursuing unless kept behind an option or the number of elements get reduced considerably.

I remember a suggestion by the same developer to use `svg` `tspan`s instead as they seem to offer more precise control over spacing. Perhaps this commit should be replaced with an `svg` `tspan`-based solution altogether (or be kept behind an option or abandoned for now.)

calixteman:
> > #### From: Bot.io (Windows)
> > # Failed
> > Full output at http://3.101.106.178:8877/72f715db5b9b1fb/output.txt
> > Total script time: 28.99 mins
> > 
> > * **Font tests:** Passed
> > * **Unit tests:** Passed
> > * **Integration Tests:** Passed
> > * **Regression tests:** FAILED
> > 
> > Image differences available at: http://3.101.106.178:8877/72f715db5b9b1fb/reftest-analyzer.html#web=eq.log
> 
> For many pages like **chrome-preistabelle-text-page1**, **Test** creates a lot of non-image(canvas)-backed ghost space-only text layer elements compared to **Reference**, the overall match seems worse.
> 

The text layer positioning is slightly worse and it's because there are less spans!
For example take `Anrufe ins deutsche Festnetz` (first cell in the tabular), with master there are 4 spans when there is only one with the patch. To be clear the less spans you've the more differences you'll get between text layer and canvas. To not have any differences, we must have one span per character. 
You can see few extra empty spans around `Preistabelle`: there are useful to improve copying/pasting. To see it action: select from `Preistabelle` to `FYVE` (in green part), copy and paste, with master you'll get `PreistabelleStand: Oktober 2010FYVE-PreislisteSeite 1 von 7FYVE` and with the patch you'll get `Preistabelle     Stand: Oktober 2010 FYVE - Preisliste Seite 1 von 7 FYVE` 
  

> Also a 10x DOM explosion may not seem too much for a single page but when viewing
> 
>     * multiple pages on a large 4K/8K screen at once or
> 
>     * on a small device for using browser-integrated search
> 
> 
> that jump from 2K to 20K is bound to cause issues with performance.
> 

Agreed and it's why I changed a constant to have almost the same number of span as currently.

> Overall, I am not convinced this is worth pursuing unless kept behind an option or the number of elements get reduced considerably.
> 

I really think it's worth pursuing.
Initially this patch aims to fix some issues when copying/pasting some text:
https://github.com/mozilla/pdf.js/issues/7310
and as far as I know it really improves things so it would be a pity to not have it.

> I remember a suggestion by the same developer to use `svg` `tspan`s instead as they seem to offer more precise control over spacing. Perhaps this commit should be replaced with an `svg` `tspan`-based solution altogether (or be kept behind an option or abandoned for now.)

I guess you missed my comment here:
https://github.com/mozilla/pdf.js/pull/12989#issuecomment-811370665

src/display/text_layer.js
Snuffleupagus: Looking at the *latest* test results, e.g. the `tracemonkey-text-enhance` tests, it seems pretty clear that you'll actually need to expand spaces as well now (since the point is that the textLayer elements should fill the available area in `enhanced`-mode).

Hence I still believe that my previous comment, about simply removing `isWhitespace` and its remaining usage, still applies here.

Snuffleupagus:
Besides https://github.com/mozilla/pdf.js/pull/12989#discussion_r605480343 which I still think applies; this PR should be tagged "[api-minor]" given the changes being made here.

src/display/text_layer.js
```diff
@@ -189,7 +178,7 @@ const renderTextLayer = (function renderTextLayerClosure() {
     // little effect on text highlighting. This makes scrolling on docs with
     // lots of such divs a lot faster.
     let shouldScaleText = false;
-    if (geom.str.length > 1) {
+    if (geom.str.length > 1 || geom.isSpace) {
```
Snuffleupagus, line 181: Is it necessary to do this for *all* spaces, or was it only necessary for the `enhanced`-mode?
If the latter is true, then the following would probably be a good idea: 
```javascript
if (geom.str.length > 1 || (task._enhanceTextSelection && geom.isSpace)) {
```

calixteman:
I think I've a better patch to fix these issues so to avoid to have a too long thread I'm closing this PR and I'll open a new one.

calixteman:
New PR: https://github.com/mozilla/pdf.js/pull/13257

PR #12993, opened 2021-02-14
[api-minor] Change the `dc:subject` Metadata field to an Array

### LINKED ISSUE #12990: Multiple authors squished together in dc: creator field
http://mariovalle.name/SciViz/1-s2.0-S0097849310001846-main.pdf

Configuration:
- Web browser and its version: Using pdf.js from Node.js through npm package `pdfjs-dist`
- Operating system and its version: Windows 10 64bits
- PDF.js version: 2.6.347
- Is a browser extension: No

1. Load the file in acrobat
2. In document properties look at the Author field: it is `Samuel Silva; Beatriz Sousa Santos; Joaquim Madeira`
3. Look at  Additional metadata > advanced > dc:creator. It is marked as `seq container` and contains the 3 authors.
4. Using the code below gives: `"Author": "Samuel Silva"` that is, only the first one. And `"dc:creator": "Samuel SilvaBeatriz Sousa SantosJoaquim Madeira"`, that is the three authors squashed together without separators.

In both fields the three authors as a single string with some sort of separator between them. Note that other pdf files that contains multiple authors already have them as a single string with names separated by commas.

The code used:
```
var pdfjsLib = require("pdfjs-dist/es5/build/pdf.js");
var pdfPath = "http://mariovalle.name/SciViz/1-s2.0-S0097849310001846-main.pdf";
var loadingTask = pdfjsLib.getDocument(pdfPath);
loadingTask.promise
  .then(function (doc) {
    var lastPromise;
    lastPromise = doc.getMetadata().then(function (data) {
      console.log(JSON.stringify(data.info, null, 2));
      if (data.metadata) {
        console.log(JSON.stringify(data.metadata.getAll(), null, 2));
      }
    });
  });
```

Snuffleupagus:
This patch simply extends the existing handling of the `dc:creator` field, which should hopefully suffice here; please refer to https://wwwimages2.adobe.com/content/dam/acom/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart1.pdf#page=34

Fixes https://github.com/mozilla/pdf.js/issues/12990#issuecomment-778787111

### DIFF
--- a/src/display/metadata.js
+++ b/src/display/metadata.js
@@ -89,23 +89,18 @@ class Metadata {
     return entry.childNodes.filter(node => node.nodeName === "rdf:li");
   }
 
-  _getCreators(entry) {
-    if (entry.nodeName !== "dc:creator") {
-      return false;
-    }
+  _parseArray(entry) {
     if (!entry.hasChildNodes()) {
-      return true;
+      return;
     }
-
     // Child must be a Bag (unordered array) or a Seq.
-    const seqNode = entry.childNodes[0];
-    const authors = this._getSequence(seqNode) || [];
+    const [seqNode] = entry.childNodes;
+    const sequence = this._getSequence(seqNode) || [];
+
     this._metadataMap.set(
       entry.nodeName,
-      authors.map(node => node.textContent.trim())
+      sequence.map(node => node.textContent.trim())
     );
-
-    return true;
   }
 
   _parse(xmlDocument) {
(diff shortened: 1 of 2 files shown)


src/display/metadata.js
```diff
@@ -89,22 +89,24 @@ class Metadata {
     return entry.childNodes.filter(node => node.nodeName === "rdf:li");
   }
 
-  _getCreators(entry) {
-    if (entry.nodeName !== "dc:creator") {
-      return false;
-    }
+  _getArray(entry) {
+    const name = entry.nodeName;
+    assert(
```
calixteman, line 94: This `assert` is useless since the name has already been checked in the `switch`.

Snuffleupagus: I added that as an extra level of security, to avoid a future accidental call of `_getArray` for an unsupported name.
However, as you pointed it out, it's not entirely necessary and I've removed it now.

Snuffleupagus: Since the return value is now consistent, I've renamed/rewritted the method to `_parseArray` instead.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #12995, opened 2021-02-16
Stop showing the fallback bar for "errorFontMissing" errors (PR 11218 follow-up)

Snuffleupagus:
*This is somewhat similar to PR #12931.*

For PDF documents where fonts are completely missing in the /Resources dictionaries, there's basically no "correct" way of rendering the document.
Hence it's very unlikely that another PDF viewer will do a better job than PDF.js in these cases, and consequently it seems highly questionable if the fallback bar really helps here.


timvandermeij (approved):

timvandermeij:
That seems fair to me. Thanks!

PR #12998, opened 2021-02-17
Simplify the default value handling of `renderInteractiveForms` in the viewer components

Snuffleupagus:
I happened to look at this code, and I can't for the life of me figure out why I didn't just implement it like this patch in the first place (since the current format feels overly verbose).


timvandermeij (approved):

timvandermeij:
Looks good!

PR #9777, opened 2018-06-04
Various `PageViewport` and `MessageHandler` cleanup

Snuffleupagus:
*Split off from PR #9729, as requested in https://github.com/mozilla/pdf.js/pull/9729#issuecomment-393698402.* /cc @timvandermeij

For easier reviewing I'd suggest first of all to look at one commit at a time, and second of all to use the `?w=1` flag to reduce the sizes of some of the diffs.

PR #12999, opened 2021-02-17
[api-minor] Remove support for synchronous event dispatching in `LoopbackPort`

Snuffleupagus:
*Please note:* The `defer` parameter has been enabled by default ever since PR #9777 (in 2018), which first shipped in PDF.js release `2.0.943`.
With workers *disabled*, e.g. in Node.js environments, this has been used ever since without any problems reported[1].

The impetus for this change was that I happened to notice that *if* the `LoopbackPort` was used with synchronous event dispatching, we'd simply send that data as-is to the listeners. This created an inconsistency in the data returned from the `pdf.worker.js` file, since `postMessage` used with *actual* workers (or the `LoopbackPort` with `defer = true`) will ignore/throw when encountering unclonable data.
Originally my intention was simply to just call `cloneValue` regardless of the event dispatching used in `LoopbackPort`, however looking at the use-cases (or lack thereof) of the `LoopbackPort` it seemed reasonable to simply remove the `defer` parameter instead.

This patch is tagged "[api-minor]" since the `LoopbackPort` is still exposed in the API, although I really hope that no third-party is using this (since disabling workers leads to bad performance).

Finally, this patch changes a `forEach` loop to `for...of` and makes uses of optional changing in existing code.

---
[1] As evident by the `npm test` command run by Github Actions, and previously by Travis.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1624,9 +1624,8 @@ class PDFPageProxy {
 }
 
 class LoopbackPort {
-  constructor(defer = true) {
+  constructor() {
     this._listeners = [];
-    this._defer = defer;
     this._deferred = Promise.resolve(undefined);
   }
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
I also don't see a particular good use case for this. Let's remove it; thanks!

PR #13001, opened 2021-02-18
Send the `AnnotationStorage`-data to the worker-thread as a `Map`

Snuffleupagus:
Rather than converting the `AnnotationStorage`-data to an Object, before sending it to the worker-thread, we should be able to simply send the internal `Map` directly.
The "structured clone algorithm" doesn't have a problem with `Map`s, however the `LoopbackPort` used when workers are *disabled* (e.g. in Node.js environments) didn't use to support them. With PR #12997 having lifted that restriction, we should now be able to simply send the `AnnotationStorage`-data as-is rather than having to iterate through it to first create an Object.

*Please note:* The changes in `src/core/annotation.js` could have been a lot more compact if we were able to use optional chaining in the `src/core` folder. Unfortunately that's still not possible, since SystemJS is being used in the development viewer (i.g. `gulp server`) and fixing that is *still* blocked by [bug 1247687](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687).

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -348,9 +348,10 @@ class Annotation {
   }
 
   isHidden(annotationStorage) {
-    const data = annotationStorage && annotationStorage[this.data.id];
-    if (data && "hidden" in data) {
-      return data.hidden;
+    const storageEntry =
+      annotationStorage && annotationStorage.get(this.data.id);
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Nice work; thanks!

PR #13004, opened 2021-02-20
Add a `this`-bound method for `InternalRenderTask.cancel`

Snuffleupagus:
This is similar to the other methods, and the only reason for this not having been done originally is that the `cancel` functionality is a later addition.


timvandermeij (approved):

timvandermeij:
Thanks for making this more consistent!

PR #13006, opened 2021-02-20
Annotation - Use border and background colors from MK dictionary

### LINKED ISSUE #13003: PDF rendering problem: Checkboxes have black background
[Erstattung von Arbeitgeberaufwendungen nach __ 56 Abs. 1 und 57, 58 Infektionsschutzgesetz - Antrag(1).pdf](https://github.com/mozilla/pdf.js/files/6009589/Erstattung.von.Arbeitgeberaufwendungen.nach.__.56.Abs.1.und.57.58.Infektionsschutzgesetz.-.Antrag.1.pdf)

Also available at: [formularserver.bayern.de](https://formularserver.bayern.de/intelliform/forms/stmi+regierungen/rof/b5/55.2/rof_55.2-086/index)

Configuration:
- Web browser and its version: FF 85.0.2 and FF 86.0b9
- PDF.js version: 
FF 85.0.2:
const pdfjsVersion = '2.7.345';
const pdfjsBuild = 'b194c820b';

FF 86.0b9:
const pdfjsVersion = '2.7.510';
const pdfjsBuild = '631bada0d';

- Is a browser extension:
no
1. load pdf document

This is with latest Chrome:
grafik

The checkboxes for e.g. "Geschlecht" are black and you cannot see which one is checked. But this is true for every checkbox in that document.
grafik

calixteman:
- aims to fix #13003;
 - pass the colors to js sandbox;
 - fix checkbox color.


src/core/annotation.js
```diff
@@ -451,46 +454,59 @@ class Annotation {
...
+  /**
+   * Set the color for background and border if any.
+   * The default values are transparent.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {Dict}  - The MK dictionary
+   */
+  setBorderAndBackgroundColors(mk) {
+    if (mk) {
```
Snuffleupagus, line 507: *Drive-by comment, to prevent bugs in corrupt documents:*
```suggestion
    if (mk instanceof Dict) {
```

calixteman: You're so right...

src/core/annotation.js
```diff
@@ -451,46 +454,59 @@ class Annotation {
...
+   * The default values are transparent.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {Dict}  - The MK dictionary
+   */
+  setBorderAndBackgroundColors(mk) {
+    if (mk instanceof Dict) {
+      this.borderColor = this._getColor(mk.get("BC"), undefined);
+      this.backgroundColor = this._getColor(mk.get("BG"), undefined);
```
Snuffleupagus, line 509: *Sorry, one more thing here:* As far as I understand, these entries are Arrays right?
If so, then please use `getArray(...)` rather than `get(...)` to prevent issues if any of the Array entries contains `Ref`s.

---

Note that similar things have caused issues in other parts of the code-base, which prompted me to add the `getArray` method years ago.

timvandermeij:
This PR needs a rebase. I'm also not really sure what to make of some of the reference test failures, mainly `firefox-issue12716-page1`. Does it also render like that in Adobe Reader/Acrobat?

calixteman:
I played with acrobat DC and the background color is used when the widget has the focus.
When unfocused, an opaque (blue/violet) is used when the bg color is opaque and a transparent (still blue/violet) one when the bg color is transparent.

Snuffleupagus:
Sorry to step in here, but that last version of this patch feels somewhat unfortunate in general, since:
 - It moves appearance from the CSS files and into the JS files, which is less maintainable.
 - It requires no less than *two* event listeners per element, just to manage the colours correctly.

I suppose that it might make sense to take a step back and consider how other PDF viewers (e.g. Adobe Reader and PDFium) handles fillable fields in *general*, since it seems that they don't necessarily always outline all fields the way that we currently do (and it rather seem to depend on the PDF file itself)!?

calixteman:
In either acrobat dc or pdfium, the bg color of an unfocused rendered input has alpha or not depending if bg color of the annotation has alpha. So I don't see how avoid to have default colors in js instead of css, please tell me if it's possible. Your point is fair and I tried to figure out a solution to avoid that but I didn't manage to.
About listeners, I agree too but I don't see how to do differently: any good ideas are welcome.
Anyway I can remove listeners and remove the color in js but then as pointed by Tim the rendering is different from what we've in Acrobat: some annotations are blue and some others are not.

brendandahl:
Not sure if this is really better, but we could leave defaults in css and use getComputedStyle. @Snuffleupagus any thoughts on a better way to do this?

Snuffleupagus:
When looking at how e.g. Adobe Reader handles forms, it seems (at least to me) that things like e.g. the background of form fields are generally controlled by the PDF document *itself* rather than the viewer.

Hence my suggestion would be that we could perhaps simplify things overall by doing that in PDF.js as well, and stop setting e.g. a default `background-color` (in `web/annotation_layer_builder.css` or elsewhere) and instead defer that to the PDF document!?

calixteman:
For me the "blue" background is just added by the reader to indicate that there is a fillable form element so I think it's useful for the user.
@timvandermeij what do you think ?

Snuffleupagus:
How about the following compromise, then:
 - Leave the *default styling* in the CSS, as before.
 - Let PDF documents with field colours, set in /MK dictionaries, simply *override* the defaults when present.

That should hopefully reduce the overall complexity *somewhat* here, and also guarantee that PDF documents renders as intended (by the document author).

timvandermeij:
I'll have a look at this once it's rebased, since then the patch will be simpler as all color function related changes have already been made in the square/circle annotation PR that was just merged, and with the comment above implemented.

src/core/annotation.js
```diff
@@ -489,7 +492,22 @@ class Annotation {
...
+   * Set the color for background and border if any.
+   * The default values are transparent.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {Dict}  - The MK dictionary
+   */
+  setBorderAndBackgroundColors(mk) {
+    if (mk instanceof Dict) {
+      this.borderColor = getRgbColor(mk.getArray("BC"), undefined);
```
timvandermeij, line 508: Transparent is indicated by `null` instead of `undefined` as indicated on line 202.

src/display/annotation_layer.js
```diff
@@ -976,6 +1016,15 @@ class CheckboxWidgetAnnotationElement extends WidgetAnnotationElement {
       );
     }
 
+    this._setBorderAndBackgroundColors(element);
+
+    const { fontColor } = this.data.defaultAppearanceData;
```
timvandermeij, line 1021: Why is this necessary only here and not elsewhere? Moreover, checkbox widgets don't contain text AFAIK, so what does this do?

calixteman:
> How about the following compromise, then:
> 
>     * Leave the _default styling_ in the CSS, as before.
> 
>     * Let PDF documents with field colours, set in /MK dictionaries, simply _override_ the defaults when present.
> 
> 
> That should hopefully reduce the overall complexity _somewhat_ here, and also guarantee that PDF documents renders as intended (by the document author).

Ok so if I understand correctly:
 - no bg color in MK: use "blue" one when unfocused and white when focused (current behaviour)
 - else: use the color as bg when unfocused and focused.

But if bg color is white or transparent and there is no border, there is no way for the user that there is something modifiable.
And we'll probably have the same issue as pointed by @timvandermeij: https://github.com/mozilla/pdf.js/pull/13006#issuecomment-782861509.

My goal was to have exactly the same behaviour as in either Acrobat or chrome.

Snuffleupagus:
> But if bg color is white or transparent and there is no border, there is no way for the user that there is something modifiable.

In that case the PDF document *itself* is responsible for that behaviour, so I don't really see this as a big problem.
Hence my suggestion above to *not* mess with "default" colours in the JS files, to avoid e.g. the focus/blur event listeners.

> My goal was to have exactly the same behaviour as in either Acrobat or chrome.

Of course I understand that, but there's something to be said for simplicity of the overall implementation as well (in my opinion).

src/core/annotation.js
calixteman: I didn't find in the spec what's the default color (transparent or whatever), so `undefined` stands for "use HTML default".

src/display/annotation_layer.js
calixteman: Checkboxes contains text for the `on` state: most of the time one char using `ZapfDingbats` font.

src/core/annotation.js
Snuffleupagus: Wait, does that mean that we now use both `undefined` *and* `null`, but they have very slightly different meaning in the display-layer?
If so, that sounds like a recipe for confusion...

calixteman:
> > But if bg color is white or transparent and there is no border, there is no way for the user that there is something modifiable.
> 
> In that case the PDF document _itself_ is responsible for that behaviour, so I don't really see this as a big problem.

Yes you're right.
But when a form has been created and tested with Acrobat, some creators saw the blue background and so maybe for this reason they didn't set a background and/or border.
I've the impress (maybe I'm wrong) that removing this blue bg can diminish the quality of the user xp.
I really understand your concern and I don't want to bikeshed but I'm really not sure it's good for the user.

> Hence my suggestion above to _not_ mess with "default" colours in the JS files, to avoid e.g. the focus/blur event listeners.
> 
> > My goal was to have exactly the same behaviour as in either Acrobat or chrome.
> 
> Of course I understand that, but there's something to be said for simplicity of the overall implementation as well (in my opinion).

Snuffleupagus:
> But when a form has been created and tested with Acrobat, some creators saw the blue background and so maybe for this reason they didn't set a background and/or border.

Please keep in mind that the "default" blue background, in Adobe Reader, mentioned is actually a *setting* which can be toggled on/off in the options (and that the particular colour used is also customizable); hence I'm not *entirely* convinced by this argument.

I'm mostly worried about readability/maintainability of the solution, especially since these types of changes feel like they have the potential to require additional fixes later on for various edge-cases in "weird" PDF documents.
(Since especially for Annotations it seems that many PDF generators aren't really following the specification all that well.)

calixteman:
I'm closing this PR and I'll open a new one.

PR #13015, opened 2021-02-24
JS - Avoid a popup to ask for updating Acrobat.

calixteman:
- this popup appears because js is enabled;
  - and because the pdf contains some unsupported features (e.g. XFA);
  - can be tested with: https://www.cbsa-asfc.gc.ca/publications/forms-formulaires/a10.pdf.

### DIFF
--- a/src/scripting_api/initialization.js
+++ b/src/scripting_api/initialization.js
@@ -115,6 +115,12 @@ function initSandbox(params) {
   globalThis.trans = Trans;
   globalThis.zoomtype = ZoomType;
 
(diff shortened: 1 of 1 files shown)


brendandahl:
Could this cause issues with other PDFs where we say something is supported but it isn't?

calixteman:
Good question...
I get it here too: https://www.canada.ca/content/dam/ircc/migration/ircc/english/pdf/kits/forms/imm0008enu_2d.pdf
Anyway, when js wasn't supported, the code to display this popup was never executed so it isn't really a regression to disable that stuff.

PR #13018, opened 2021-02-24
XFA - Create Form DOM in merging template and data trees

calixteman:
- Spec: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=171;
  - support for the 2 ways of merging: consumeData and matchTemplate;
  - create additional nodes in template DOM when occur node allows it;
  - support for global values in data DOM.


brendandahl (changes requested):
Overall looks good. It could use some more comments in parts that are algorithms out of the spec even if it's just paraphrasing the spec.

PR #13019, opened 2021-02-25
Remove the `findResultsCount`-check from `PDFFindBar.updateResultsCount`, and unnecessary defaults from the constructor

Snuffleupagus:
Given that `PDFFindBar` is written *specifically* for the default viewer, rather than general usage (as opposed to the `PDFFindController`), we should be able to simply assume that the `findResultsCount` DOM-element is always present. Even more so, when we're purposely not doing any similar checks for other DOM-elements in this code.

Also, remove unnecessary `null` defaults for the various DOM-element options in the constructor, since the code simply assumes that all of the relevant DOM-elements are in fact available.

### DIFF
--- a/web/pdf_find_bar.js
+++ b/web/pdf_find_bar.js
@@ -28,16 +28,16 @@ class PDFFindBar {
   constructor(options, eventBus, l10n = NullL10n) {
     this.opened = false;
 
-    this.bar = options.bar || null;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks better; thanks!

PR #13020, opened 2021-02-25
Remove the `NullL10n` default value from viewer components not included in the `COMPONENTS`-bundle

Snuffleupagus:
For any viewer component not listed in `web/pdf_viewer.component.js`, it shouldn't be necessary to provide a default value for the `l10n`-parameters.
Note also that these *specific* components are heavily tailored towards the default viewer use-case, rather than for general usage.

### DIFF
--- a/web/password_prompt.js
+++ b/web/password_prompt.js
@@ -13,7 +13,6 @@
  * limitations under the License.
  */
 
-import { NullL10n } from "./ui_utils.js";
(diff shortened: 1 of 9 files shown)


timvandermeij (approved):

timvandermeij:
I also don't see a good reason to keep this around. Thanks!

PR #13026, opened 2021-02-26
Convert code in `src/core/crypto.js` to use "normal" classes

Snuffleupagus:
All of this code predates the existence of native JS classes, however we can now clean this up a bit. This patch thus let us remove some variable "shadowing" from the code.

*Much smaller diff with https://github.com/mozilla/pdf.js/pull/13026/files?w=1*


timvandermeij (approved):

timvandermeij:
Nice readability improvement!

PR #13029, opened 2021-02-27
Enable the `no-var` linting rule in more core files

timvandermeij:
This is mostly done using the `gulp lint --fix` command, and manual changes in a small number of cases where ESLint couldn't determine what to use, in an effort to reduce the number of disabled ESLint rules in the core folder and to use modern syntax.

### DIFF
--- a/src/core/glyphlist.js
+++ b/src/core/glyphlist.js
@@ -13,11 +13,10 @@
  * limitations under the License.
  */
 /* no-babel-preset */
-/* eslint-disable no-var */
 
 import { getArrayLookupTableFactory } from "./core_utils.js";
 
-var getGlyphsUnicode = getArrayLookupTableFactory(function () {
+const getGlyphsUnicode = getArrayLookupTableFactory(function () {
   // prettier-ignore
   return [
     "A", 0x0041,
@@ -4347,7 +4346,7 @@ var getGlyphsUnicode = getArrayLookupTableFactory(function () {
   ];
 });
 
-var getDingbatsGlyphsUnicode = getArrayLookupTableFactory(function () {
+const getDingbatsGlyphsUnicode = getArrayLookupTableFactory(function () {
   // prettier-ignore
   return [
     "space", 0x0020,
--- a/src/core/unicode.js
+++ b/src/core/unicode.js
@@ -13,7 +13,6 @@
  * limitations under the License.
  */
 /* no-babel-preset */
-/* eslint-disable no-var */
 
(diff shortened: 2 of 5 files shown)


src/core/glyphlist.js
```diff
@@ -13,11 +13,10 @@
  * limitations under the License.
  */
 /* no-babel-preset */
-/* eslint-disable no-var */
```
Snuffleupagus, line 16: *Including this here for future reference, to clarify that we've considered the implications of these changes:*

Here, and in `src/core/unicode.js`, we include these files as-is in the "legacy"-builds because of issues with Babel; please note https://github.com/mozilla/pdf.js/blob/052db45e7a26fead95b5233e3425008784daa8cf/gulpfile.js#L198-L200

Hence this PR will result in `let`/`const` being included even in the "legacy"-builds, however that should be fine given the following compatibility information:
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#browser_compatibility

Note also that IE 11 support is now explicitly removed, and this CSS-compatibility configuration for the `GENERIC`-builds: https://github.com/mozilla/pdf.js/blob/052db45e7a26fead95b5233e3425008784daa8cf/gulpfile.js#L78-L90

src/core/unicode.js
```diff
@@ -1629,13 +1628,13 @@ var getNormalizedUnicodes = getArrayLookupTableFactory(function () {
...
   if (charsLength <= 1 || !isRTLRangeFor(chars.charCodeAt(0))) {
     return chars;
   }
-  var s = "";
-  for (var ii = charsLength - 1; ii >= 0; ii--) {
+  let s = "";
+  for (let ii = charsLength - 1; ii >= 0; ii--) {
     s += chars[ii];
   }
   return s;
```
Snuffleupagus, line 1640: *Nit:* Can we please change this to use a buffer, rather than string concatenation, since that should be more efficient for short strings (which is what we're dealing with here). 

*Obviously making these changes in a separate commit would be fine.*

Snuffleupagus (approved):
Looks great, thank you!

I've added one (small) suggestion, but it's obviously not a blocker :-)

src/core/unicode.js
timvandermeij: Done in a new commit. Good idea!

PR #13030, opened 2021-02-27
Modernize some of the code in `src/core/cmap.js` by using classes and async/await

Snuffleupagus:
This converts a couple of our old "classes" to proper ECMAScript classes, and replaces a lot of manual Promise-wrapping with async/await instead.

*Smaller diff with https://github.com/mozilla/pdf.js/pull/13030/files?w=1*


timvandermeij (approved):

timvandermeij:
Nice improvement!

PR #13031, opened 2021-02-27
Implement rendering square/circle annotations without appearance stream

timvandermeij:
Fixes another part of #6810.


src/core/annotation.js
```diff
@@ -2278,6 +2291,43 @@ class SquareAnnotation extends MarkupAnnotation {
     super(parameters);
 
     this.data.annotationType = AnnotationType.SQUARE;
+
+    if (!this.appearance) {
```
Snuffleupagus, line 2295: *Just to help me understand the differences between Annotation types:*

It seems that a few of existing `.... extends MarkupAnnotation`-cases only does this when `quadPoints` is defined, note [e.g. `HighlightAnnotation`](https://github.com/mozilla/pdf.js/blob/8e74278b65d2abf7ee720aa615af87757b05626e/src/core/annotation.js#L2364-L2370), hence I wonder what's different between the various Annotation types here?

timvandermeij: Good question! I looked into the specification and found that the existing annotation types that support this, namely highlight/squiggly/underline/strikethrough, are text markup annotations and those are _required_ to have quadpoints; see https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=411&zoom=auto,-215,424. Circle and square annotations on the other hand are regular markup annotations and for those quadpoints are not defined; see https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=411&zoom=auto,-215,424.

In our implementation we don't make this distinction because for the way we work with them this difference is not relevant. The overview at https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=398&zoom=auto,-215,775 shows which annotation types are markup annotations and it also shows that highlight/squiggly/underline/strikethrough annotations belong to the text markup annotation "subcategory". The reason is that they are always applied to text, whereas squares/circles/other markup annotations are not text-related and can be used anywhere in the document.

Snuffleupagus: > Hopefully this helps a bit to clear things up :-)

That's indeed *very* helpful, thank you!

src/core/annotation.js
```diff
@@ -2286,6 +2336,67 @@ class CircleAnnotation extends MarkupAnnotation {
...
+      this._setDefaultAppearance({
+        xref: parameters.xref,
+        extra: `${this.borderStyle.width} w`,
+        strokeColor,
+        fillColor,
+        pointsCallback: (buffer, points) => {
+          // Circles are approximated by Bézier curves with four segments since
+          // there is no circle primitive in the PDF specification. For the
+          // control points distance, see https://stackoverflow.com/a/27863181.
+          const controlPointsDistance = (4 / 3) * Math.tan(Math.PI / (2 * 4));
```
Snuffleupagus, line 2365: Nit: Maybe move this out of the callback-function, since it's a proper constant?

src/core/annotation.js
```diff
@@ -2278,6 +2291,43 @@ class SquareAnnotation extends MarkupAnnotation {
...
+          const width = points[3].x - points[2].x - this.borderStyle.width;
+          const height = points[1].y - points[3].y - this.borderStyle.width;
+          buffer.push(`${x} ${y} ${width} ${height} re`);
+          if (fillColor) {
+            buffer.push("B");
+          } else {
+            buffer.push("S");
+          }
+          return [points[0].x, points[1].x, points[3].y, points[1].y];
+        },
```
Snuffleupagus, line 2328: Testing this, with the preview, there seem to be some *small* discrepancy between the rendered circle border and clickable border. This is especially noticeable at larger zoom levels, and I'm not entirely sure if it's the new code above or the existing one in the display-layer that needs adjusting.

(Note that for the SQUARE annotation, the rendered/clickable border agrees very well.)

Snuffleupagus (changes requested):

src/core/annotation.js
timvandermeij: Done in the new commit.

src/core/annotation.js
timvandermeij: Well spotted! I indeed made a mistake with including the border width in the calculations. Fixed in the new commit.

Snuffleupagus (approved):
r=me, with passing tests.

This looks great, thanks for improving the "no appearance stream"-support!

PR #13032, opened 2021-02-28
Don't warn about actions that require scripting support in `Catalog.parseDestDictionary`

Snuffleupagus:
Now that we have scripting support, warning about e.g. JavaScript actions doesn't seem necessary anymore. Especially considering that scripting-related actions are/will not be parsed by the `Catalog.parseDestDictionary` method anyway, since it's intended for handling "simple" actions.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -1366,7 +1366,16 @@ class Catalog {
           }
         /* falls through */
         default:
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

PR #13033, opened 2021-02-28
Implement rendering line annotations without appearance stream

timvandermeij:
Fixes another part of #6810.


Snuffleupagus (approved):
r=me, with passing tests.

Looks simple enough, thank you!

PR #13037, opened 2021-03-01
fixing checkboxes in presentation mode (issue 12232)

rezzaghi:
This avoids checking a checkbox in presentation mode.
It is my first time contributing to open source, please let me know if i made some novice mistake or something like that.


Snuffleupagus (changes requested):
Sorry, but this is *way* too complex and it really ought to be possible to fix it with just CSS rules instead.

timvandermeij:
I agree, this should be addressed with CSS rules specifically for presentation mode.

PR #13050, opened 2021-03-04
Collect all l10n fallback strings, used in the viewer, in one helper function (PR 12981 follow-up)

Snuffleupagus:
Rather than having to spell out the English fallback strings at *every* single `IL10n.get` call-site throughout the viewer, we can simplify things by collecting them in *one* central spot.
This provides a much better overview of the fallback l10n strings used, which makes future changes easier and ensures that fallback strings occuring in multiple places cannot accidentally get out of sync.
Furthermore, by making the `fallback` parameter of the `IL10n.get` method *optional*[1] many of the call-sites (and their surrounding code) become a lot less verbose.

---
[1] It's obviously still possible to pass in a fallback string, it's just not required.

### DIFF
--- a/web/annotation_layer_builder.js
+++ b/web/annotation_layer_builder.js
@@ -14,7 +14,7 @@
  */
 
 import { AnnotationLayer } from "pdfjs-lib";
-import { NullL10n } from "./ui_utils.js";
+import { NullL10n } from "./l10n_utils.js";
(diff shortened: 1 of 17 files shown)


timvandermeij (approved):

timvandermeij:
Looks really clean; thanks!

PR #13053, opened 2021-03-04
Enable scripting by default in the development viewer; remove the standalone `acroforms` example

Snuffleupagus:
- Enable scripting by default in the development viewer

   Given that scripting is now enabled in Firefox Nightly (but only there), it seems weird to not have scripting enabled by default in `gulp server` mode.

 - Remove the standalone `acroforms` example

   At this point in time the standalone `acroforms` example is no longer needed, since both the `simpleviewer` and `singlepageviewer` examples already support AcroForms by default.
Moreover, the `acroforms` example no longer illustrate best practices, given the direct usage of `PDFPageView` for a multipage document.


Snuffleupagus:
*Please note:* The `simpleviewer`/`singlepageviewer` examples will be updated in PR #13042 to mention both AcroForms and scripting support

timvandermeij (approved):

timvandermeij:
Looks good!

PR #13054, opened 2021-03-05
Move handling of the PageOpen/PageClose events into the `PDFScriptingManager` (PR 13042 follow-up)

Snuffleupagus:
By moving this code from the `BaseViewer` and into `PDFScriptingManager`, all of the scripting initialization/handling code is now limited to just one file/class which help overall readability (in my opinion). Also, this patch is a *net reduction* in number of lines of code which can never hurt.

As part of these changes, the intermediary "pageopen"/"pageclose" events are now removed in favor of using the "regular" viewer events directly in `PDFScriptingManager`. Hence this removes some (strictly unnecessary) indirection in the current code, when handling PageOpen/PageClose events, which leads to overall fewer function calls in this part of the code.

*Slightly smaller/easier diff with https://github.com/mozilla/pdf.js/pull/13054/files?w=1*

### DIFF
--- a/web/pdf_scripting_manager.js
+++ b/web/pdf_scripting_manager.js
@@ -15,6 +15,7 @@
 
 import { createPromiseCapability, shadow } from "pdfjs-lib";
 import { apiPageLayoutToSpreadMode } from "./ui_utils.js";
+import { RenderingStates } from "./pdf_rendering_queue.js";
 
 /**
  * @typedef {Object} PDFScriptingManagerOptions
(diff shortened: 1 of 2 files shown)


web/pdf_scripting_manager.js
```diff
@@ -96,17 +98,33 @@ class PDFScriptingManager {
     this._scripting = this._createScripting();
 
     this._internalEvents.set("updatefromsandbox", event => {
+      if (event?.source !== window) {
+        return;
+      }
```
Snuffleupagus, line 103: While this is probably not *necessary*, it a "better safe than sorry"-type of check I forgot to add in PR #13042

timvandermeij (approved):

timvandermeij:

PR #13055, opened 2021-03-06
JS - reset correctly radio buttons

calixteman:


src/display/annotation_layer.js
```diff
@@ -1016,19 +1015,16 @@ class RadioButtonWidgetAnnotationElement extends WidgetAnnotationElement {
...
             for (const radio of document.getElementsByName(event.target.name)) {
               const radioId = radio.getAttribute("id");
-              if (fieldValue === radio.getAttribute("pdfButtonValue")) {
-                radio.setAttribute("checked", true);
-                storage.setValue(radioId, { value: true });
-              } else {
-                storage.setValue(radioId, { value: false });
-              }
+              radio.checked = radioId === id && checked;
+              storage.setValue(id, { value: radio.checked });
```
Snuffleupagus, line 1027: Unless I'm completely misunderstanding this line, won't it simply keep overwriting the *same* AnnotationStorage entry over and over for each loop iteration?

Is that actually intended, or should it still say `radioId` here?

calixteman: pfff... thx

timvandermeij:
It looks like there is a regression in the `annotation-button-widget-forms-page1` test. I can only guess that this has something to do with the `setAttribute` change, although I don't really understand why, since the other changes are related to scripting which that file doesn't use...

timvandermeij:
It looks like there is a regression in some checkbox reference tests. I can only guess that this has something to do with the setAttribute change, although I don't really understand why, since the other changes are related to scripting which that file doesn't use...

Snuffleupagus:
>  I can only guess that this has something to do with the `setAttribute` change,

If that's case, perhaps this comment applies here as well (meaning that we need to somehow fix/update the code in `test/driver.js`): https://github.com/mozilla/pdf.js/blob/709c57a66b176d17db4b2e034d16daddd3860f33/src/display/annotation_layer.js#L636-L638

calixteman:
> It looks like there is a regression in some checkbox reference tests. I can only guess that this has something to do with the `setAttribute` change, although I don't really understand why, since the other changes are related to scripting which that file doesn't use...

I replaced `radio.setAttribute("checked"...` in the listener because else it doesn't work and so to be consistent I changed the other one too.

src/display/annotation_layer.js
```diff
@@ -915,7 +915,7 @@ class CheckboxWidgetAnnotationElement extends WidgetAnnotationElement {
     element.type = "checkbox";
     element.name = this.data.fieldName;
     if (value) {
-      element.setAttribute("checked", true);
+      element.setAttributes("checked", true);
```
Snuffleupagus, line 918: There's a extra "s" added at the end of the method, which probably explains why not all tests run with the test logs containing:
```
TEST-UNEXPECTED-FAIL | test failed firefox has not responded in 120s
TEST-UNEXPECTED-FAIL | test failed chrome has not responded in 120s
```

src/display/annotation_layer.js
```diff
@@ -1000,9 +1000,8 @@ class RadioButtonWidgetAnnotationElement extends WidgetAnnotationElement {
     element.type = "radio";
     element.name = data.fieldName;
     if (value) {
-      element.setAttribute("checked", true);
+      element.setAttributes("checked", true);
```
Snuffleupagus, line 1003: As above.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #13056, opened 2021-03-06
Dispatch a "Doc/WillClose" event, when scripting is enabled, when closing the document

Snuffleupagus:
This event is mentioned in the specification, see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/AcrobatDC_js_api_reference.pdf#G5.1963176, and it's also listed as a valid type in https://github.com/mozilla/pdf.js/blob/fb87704b386bdc940ceef4eff31b1374432a6122/src/shared/util.js#L164-L170

Since we're already dispatching a "PageClose" event in this case, it seems reasonable to *also* dispatch a "Doc/WillClose" event as well.

### DIFF
--- a/web/pdf_scripting_manager.js
+++ b/web/pdf_scripting_manager.js
@@ -128,6 +128,11 @@ class PDFScriptingManager {
     this._internalEvents.set("pagesdestroy", async event => {
       await this._dispatchPageClose(this._pdfViewer.currentPageNumber);
 
+      await this._scripting?.dispatchEventInSandbox({
+        id: "doc",
+        name: "WillClose",
+      });
+
       this._closeCapability?.resolve();
     });
 


calixteman:
Is there any guarantee that the sandbox won't be destroyed before the end of the callback execution in the sandbox ?
If yes we should add a special treatment for this event just to signal the sandbox can be safely destroyed.

calixteman:
@Snuffleupagus do you have the possibility to create a test file ? If no please tell me and I'll create one.

Snuffleupagus:
> Is there any guarantee that the sandbox won't be destroyed before the end of the callback execution in the sandbox ?

Not really, as far as I know...

> If no we should add a special treatment for this event just to signal that the sandbox can be safely destroyed.

Well, now that I look at the "bigger" picture I'm reminded of the fact that that we *normally* don't run any of the closing/destruction code in the default viewer as-is.
Basically, in the Firefox PDF Viewer (or even the GENERIC viewer) when the user navigates to a new web-page we don't do anything special at all. To actually get any of this code to run, we'd need to call (and wait for the *async* completion of) the `PDFViewerApplication.close` method which seems *very* messy and potentially both quite slow and annoying for the user.

Hence, in general, I'm not sure if we can actually guarantee that e.g. any of the sandbox-destruction even runs without severely degrading the overall UX!?

Snuffleupagus:
The last version of this PR waits for "Doc/WillClose" before destroying the scripting/sandbox, but as outline in https://github.com/mozilla/pdf.js/pull/13056#issuecomment-792054560 we're *really* not going to be able (in general) to enforce that the destruction code actually runs when navigation happens in the browser unfortunately. 

/botio integrationtest

timvandermeij:
Looks good to me like this. Are there any more concerns here or is this ready to land?

Snuffleupagus:
> Are there any more concerns here or is this ready to land?

Well, the patch as-is should be fine now.
(However, as outlined above, it's not really going to be possible to block navigation in the viewer on sandbox destruction completing.)

timvandermeij (approved):

timvandermeij:
Thanks!

PR #13062, opened 2021-03-07
Remove the `test/features` folder, since it's very out of date (issue 11954)

### LINKED ISSUE #11954: Verify correctness of the required browser features pages

Snuffleupagus:
These tests, and their [accompanying Wiki page](https://github.com/mozilla/pdf.js/wiki/Required-Browser-Features), haven't received any real updates for *many years* and are sufficiently out of date to be effectively useless now.
Providing *irrelevant* compatibility information seems overall worse than not providing any information at all, and as suggested in the issue it'd probably be better to use https://github.com/mozilla/pdf.js#online-demo for checking if a particular platform/browser is supported.

Thanks to version control, it's easy to restore these files should the need ever arise. However, re-introducing these tests would essentially require updating every single test-case *and* a commitment to keeping them up to date with future code changes.

Fixes #11954


timvandermeij (approved):

timvandermeij:
Looks good to me. I have also removed the wiki page and the links to the page now.

PR #13063, opened 2021-03-07
Use more optional chaining in the web/-folder (PR 12961 follow-up)

Snuffleupagus:
- Avoid unnecessary `size_kb` calculation, for large files, in `PDFDocumentProperties._parseFileSize` (PR 13050 follow-up)

 - Use more optional chaining in the `web/`-folder (PR 12961 follow-up)

   I overlooked these cases previously, but there's no reason why optional chaining (and nullish coalescing) cannot be used here as well.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13066, opened 2021-03-08
Revoke blob URL's.

koenvd:
Wrong PR


koenvd:
This was issue on our fork only.

PR #13069, opened 2021-03-09
[api-minor]XFA - Add a layer to display XFA forms

calixteman:
- add an option to enable XFA rendering if any;
  - for now, let the canvas layer: it could be useful to implement XFAF forms (embedded pdf in xml stream for the background and xfa form for the foreground);
  - ui elements in template DOM are pretty close to their html counterpart so we generate a fake html DOM from template one:
    - it makes easier to translate template properties to html ones;
    - it makes faster the creation of the html element in the main thread.


web/app_options.js
```diff
@@ -230,6 +230,11 @@ const defaultOptions = {
     value: 1,
     kind: OptionKind.API,
   },
+  enableXfa: {
+    /** @type {boolean} */
+    value: false,
+    kind: OptionKind.API,
+  },
```
Snuffleupagus, line 237: Please take care to sort options *alphabetically* within each OptionKind section in this file.

Snuffleupagus (changes requested):

web/xfa_layer_builder.js
```diff
@@ -0,0 +1,149 @@
...
+   *   annotations is complete.
+   */
+  render(viewport, intent = "display") {
+    return Promise.all([this.pdfPage.getXfa()]).then(([xfa]) => {
+      if (this._cancelled) {
+        return;
+      }
+      /* if (annotations.length === 0) {
+        return;
+      } */
```
Snuffleupagus, line 75: Please remember to remove, rather than comment out, unused code.

web/xfa_layer_builder.js
```diff
@@ -0,0 +1,149 @@
...
+ * @property {PDFPage} pdfPage
+ * @property {AnnotationStorage} [annotationStorage]
+ * @property {boolean} renderInteractiveForms
+ * @property {IPDFLinkService} linkService
+ * @property {DownloadManager} downloadManager
+ * @property {IL10n} l10n - Localization service.
+ * @property {boolean} [enableScripting]
+ * @property {Promise<boolean>} [hasJSActionsPromise]
+ * @property {Object} [mouseState]
+ */
```
Snuffleupagus, line 32: The `XfaLayer` doesn't seem to use even a fraction of these options, please remember to remove the ones that you're not actually using both from the JSDoc comments *and* from the actual code (in this file and elsewhere).

web/base_viewer.js
```diff
@@ -1311,6 +1315,38 @@ class BaseViewer {
...
+  /**
+   * @param {HTMLDivElement} pageDiv
+   * @param {PDFPage} pdfPage
+   * @param {AnnotationStorage} [annotationStorage] - Storage for annotation
+   *   data in forms.
+   * @param {boolean} renderInteractiveForms
+   * @param {IL10n} l10n
+   * @param {boolean} [enableScripting]
+   * @param {Promise<boolean>} [hasJSActionsPromise]
+   * @returns {AnnotationLayerBuilder}
```
Snuffleupagus, line 1327: Too much copy-and-paste here :-)

web/xfa_layer_builder.js
```diff
@@ -0,0 +1,149 @@
...
+class DefaultXfaLayerFactory {
+  /**
+   * @param {HTMLDivElement} pageDiv
+   * @param {PDFPage} pdfPage
+   * @param {AnnotationStorage} [annotationStorage]
+   * @param {IL10n} l10n
+   * @param {boolean} [enableScripting]
+   * @param {Promise<boolean>} [hasJSActionsPromise]
+   * @returns {AnnotationLayerBuilder}
```
Snuffleupagus, line 127: Too much copy-and-paste here :-)

web/xfa_layer_builder.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * @param {PageViewport} viewport
+   * @param {string} intent (default value is 'display')
+   * @returns {Promise<void>} A promise that is resolved when rendering of the
+   *   annotations is complete.
+   */
+  render(viewport, intent = "display") {
+    return Promise.all([this.pdfPage.getXfa()]).then(([xfa]) => {
```
Snuffleupagus, line 69: Unless you're going to add more Promises here, the `Promise.all` part should be removed.

src/display/xfa_layer.js
```diff
@@ -0,0 +1,89 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class XfaLayer {
+  static setAttributes(html, attrs) {
+    for (const [key, value] of Object.entries(attrs)) {
```
brendandahl, line 18: Just to confirm, only the predefined attributes from the various toHTML can be set here right? I just want to confirm we don't need need to remove any potentially malicious attributes from XFA (e.g. onclick).

src/core/worker.js
```diff
@@ -487,6 +491,16 @@ class WorkerMessageHandler {
       });
     });
 
+    handler.on("GetPageXfa", function wphSetupGetXfa({ pageIndex }) {
+      return pdfManager.getPage(pageIndex).then(function (page) {
+        return page.xfaData;
```
brendandahl, line 496: I think we need an ensure `pdfManager.ensure(page, xfaData` since this could potentially throw if the data isn't available.

src/display/xfa_layer.js
calixteman: Yeah, normally everything is under control:
  - only nodes and attributes defined in spec are in the final form DOM;
  - the various `toHTML` functions will generate some HTML based on those attributes and nodes.
So it's impossible to have some injections.
XFA can contain some xhtml, by spec nodes and attributes are limited too. And for xhtml nodes style is parsed to filter out any forbidden property:
https://github.com/mozilla/pdf.js/blob/master/src/core/xfa/xhtml.js#L63
So it should be safe.

src/display/api.js
```diff
@@ -1165,6 +1176,14 @@ class PDFPageProxy {
     ));
   }
 
+  /**
+   * @returns {Promise<Object | null>} A promise that is resolved with
+   *   an {Object} with a fake DOM object, or `null` if no XFA exists.
```
brendandahl, line 1181: We should describe what `fake DOM object` means.

brendandahl (approved):
r+ with more documentation on the public API part.

Issue #13075, opened 2021-03-10
pdfjs fails to read a Buffer.from or buffer from fs.readFileSync


yrambler2001:
Attach (recommended) or Link to PDF file here:
[issue13075.pdf](https://github.com/mozilla/pdf.js/files/6117265/issue13075.pdf)

Configuration:
- Web browser and its version: Node.js 14.15.5
- Operating system and its version: Mac OS 11.2.1
- PDF.js version: master/2.7.570/2.6.347/1.10.100
- Is a browser extension: false

Steps to reproduce the problem:
1. const fileBuffer = fs.readFileSync('./pdf')
2. PDFJS.getDocument(fileBuffer)
3. fileBuffer.buffer has junk data at the start and at the end. Randomly, pdfjs can not read a buffer, whose buffer.buffer contains random memory junk data in fileBuffer.buffer

What is the expected behavior? pdfjs ignores a buffer.buffer and reads only buffer

What went wrong? (add screenshot)
I've catched that random error and wrote fileBuffer.buffer of the error buffer to a file named issue13075.pdf
I've recreated a buffer that has the wrong buffer.buffer (see line 6). Pdfjs fails to read that buffer (console.log('Fail')).
The Uint8Array of that buffer, pdfjs correctly reads (console.log('OK')).
![image](https://user-images.githubusercontent.com/8398353/110660366-59f1a780-8178-11eb-94c2-29be826c2118.png)

codesandbox:
https://codesandbox.io/s/pdfjs-loading-from-buffer-issue-9h4fd?file=/src/index.js

Snuffleupagus:
> const fileBuffer = fs.readFileSync('./pdf')
> PDFJS.getDocument(fileBuffer)

As can be seen in the docs, `getDocument` does *not* accept any Node.js-specific data types: https://github.com/mozilla/pdf.js/blob/5e3af62d582133d1f3e2550768e87680c8c2076c/src/display/api.js#L195-L207

Hence you need to convert your `fileBuffer` into a *standard* `TypedArray`, otherwise it simply cannot work!

> codesandbox:
> https://codesandbox.io/s/pdfjs-loading-from-buffer-issue-9h4fd?file=/src/index.js

That gives a "502: Bad Gateway" error when I load it (in both Firefox and Chrome browser).

yrambler2001:
> That gives a "502: Bad Gateway" error when I load it (in both Firefox and Chrome browsers).
[...]

PR #13076, opened 2021-03-10
pdfjs fails to read a Buffer.from or buffer from fs.readFileSync

yrambler2001:
#13075 issue


Snuffleupagus (changes requested):
As mentioned in https://github.com/mozilla/pdf.js/issues/13075#issuecomment-795742644, I don't think that we should accept this:
> Sorry, but I *really* don't like the changes being that are suggested there. Please note that while the change may *seem* minor, you're suggesting to *very subtly* change the behaviour of a very old (and completely internal) library method.
It's thus very possible that some code, somewhere where our test-coverage isn't great, could break as a result of this modifying a code-invariant that's now a decade old.

timvandermeij:
Closing since this is fixed by #13079 instead.

Issue #13077, opened 2021-03-10
Re-enable unit test "creates pdf doc from non-existent URL"


brendandahl:
Test fail intermittently on the linux bot. I have yet to reproduce locally on linux.

Snuffleupagus:
Already tracked in #13061, hence we can probably close this one.

PR #13078, opened 2021-03-10
Disable intermittent unit test "creates pdf doc from non-existent URL"

brendandahl:
Disable this test so we don't have to manually review unit test
failure log all the time.

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -169,6 +169,10 @@ describe("api", function () {
         });
     });
     it("creates pdf doc from non-existent URL", function (done) {
+      if (!isNodeJS) {
+        // re-enable https://github.com/mozilla/pdf.js/issues/13061
+        pending("Fails intermittently on linux in browsers.");
+      }
       const loadingTask = getDocument(
         buildGetDocumentParams("non-existent.pdf")
       );


brendandahl:
Filed #13077 to re-enable at some point.

test/unit/api_spec.js
```diff
@@ -168,19 +167,20 @@ describe("api", function () {
...
-        .then(function (error) {
-          done.fail("shall fail loading");
-        })
-        .catch(function (error) {
-          expect(error instanceof MissingPDFException).toEqual(true);
-          loadingTask.destroy().then(done);
-        });
-    });
+    // TODO: re-enable https://github.com/mozilla/pdf.js/issues/13077
+    // it("creates pdf doc from non-existent URL", function (done) {
```
Snuffleupagus, line 171: Given that this doesn't seem to fail when the tests are run with Github Actions, can't we keep this enabled *conditionally* instead of just commenting it out?

Something along these lines instead:
```javascript
if (!isNodeJS) {
  pending(...);
}
```

Snuffleupagus (changes requested):

Snuffleupagus (approved):

PR #13081, opened 2021-03-11
Replace the `objectFromEntries` helper function with an `objectFromMap` one instead, and simplify the data lookup in the AnnotationStorage.getValue method

Snuffleupagus:
- Replace the `objectFromEntries` helper function with an `objectFromMap` one instead

   Given that it's only used with `Map`s, and that it's currently implemented in such a way that we (indirectly) must iterate through the data *twice*, some simplification cannot hurt here.
   Note that the only reason that we're not using `Object.fromEntries(...)` directly, at each call-site, is that that one won't guarantee that a `null` prototype is being used.

 - Simplify the data lookup in the `AnnotationStorage.getValue` method

   Rather than first checking if data exists before fetching it from storage, we can simply do the lookup directly and then check its value.
   Note that this follows the same pattern as utilized in the `AnnotationStorage.setValue` method.

### DIFF
--- a/src/display/annotation_storage.js
+++ b/src/display/annotation_storage.js
@@ -14,7 +14,7 @@
  */
 
 import { deprecated } from "./display_utils.js";
-import { objectFromEntries } from "../shared/util.js";
+import { objectFromMap } from "../shared/util.js";
 
 /**
  * Key/value storage for annotation data in forms.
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13085, opened 2021-03-12
Ensure that *all* errors are handled in `rasterizeTextLayer`/`rasterizeAnnotationLayer`

Snuffleupagus:
Currently errors occurring within the `src/display/{text_layer, annotation_layer}.js` files are not being handled properly by the test-suite, and the tests simply time out rather than failing as intended.
This makes it *very* easy to accidentally overlook a certain type of errors, see e.g. https://github.com/mozilla/pdf.js/pull/13055#discussion_r589005041, which this patch will thus prevent.

*Much simpler/smaller diff with https://github.com/mozilla/pdf.js/pull/13085/files?w=1*

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -74,32 +74,39 @@ var rasterizeTextLayer = (function rasterizeTextLayerClosure() {
       div.className = "textLayer";
       foreignObject.appendChild(div);
 
-      // Rendering text layer as HTML.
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

PR #13093, opened 2021-03-13
Enable the ESLint `no-var` rule in the `external/builder/` folder 

Snuffleupagus:
As part of testing this, I've diffed the output of `gulp mozcentral` with/without this patch and the *only* difference is the incremented `version`/`build` numbers.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13094, opened 2021-03-13
Enable the `no-var` linting rule for `src/core/{operator_list, pattern}.js`

timvandermeij:
The commit messages contain more information about the individual commits.

### DIFF
--- a/src/core/operator_list.js
+++ b/src/core/operator_list.js
@@ -12,15 +12,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint-disable no-var */
 
 import { assert, ImageKind, OPS, warn } from "../shared/util.js";
 
-var QueueOptimizer = (function QueueOptimizerClosure() {
+const QueueOptimizer = (function QueueOptimizerClosure() {
   function addState(parentState, pattern, checkFn, iterateFn, processFn) {
-    var state = parentState;
-    for (var i = 0, ii = pattern.length - 1; i < ii; i++) {
-      var item = pattern[i];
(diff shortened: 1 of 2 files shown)


src/core/operator_list.js
```diff
@@ -106,15 +106,15 @@ var QueueOptimizer = (function QueueOptimizerClosure() {
...
-      var map = [],
-        maxLineHeight = 0;
-      var currentX = IMAGE_PADDING,
+      let maxX = 0;
+      const map = [];
+      let maxLineHeight = 0;
+      let currentX = IMAGE_PADDING,
         currentY = IMAGE_PADDING;
-      var q;
+      let q;
```
Snuffleupagus, line 114: *Optional nit:* I believe that you can simply inline this definition in the loop headers below, since it doesn't seem to ever be accessed outside of the loops.

src/core/operator_list.js
```diff
@@ -226,10 +226,10 @@ var QueueOptimizer = (function QueueOptimizerClosure() {
         return i - ((i - iFirstSave) % 4);
       }
 
-      var q;
-      var isSameImage = false;
-      var iTransform, transformArgs;
-      var firstPIMXOArg0 = argsArray[iFirstPIMXO][0];
+      let q;
```
Snuffleupagus, line 229: *Optional nit:* I believe that you can simply inline this definition in the loop headers below, since it doesn't seem to ever be accessed outside of the loops.

Snuffleupagus (approved):
I've added two *optional* comments, but other than that this looks good to me; thank you!

PR #13095, opened 2021-03-13
Enable linting of the `external/cmapscompress/` folder

Snuffleupagus:
Given that this our "own" code, there's no good reason as far as I'm concerned to not lint it.


timvandermeij (approved):

timvandermeij:
Thank you for bringing this code up-to-date with the coding style!

PR #13099, opened 2021-03-14
print: A couple print fixes.

emilio:
See each commit for details. This fixes:

 * https://bugzilla.mozilla.org/show_bug.cgi?id=1697778
 * https://bugzilla.mozilla.org/show_bug.cgi?id=1698136

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -1336,6 +1336,7 @@ class BaseViewer {
       return {
         width: viewport.width,
         height: viewport.height,
+        scale: pageView.scale,
         rotation: viewport.rotation,
       };
     });
--- a/web/firefox_print_service.js
+++ b/web/firefox_print_service.js
@@ -39,6 +39,8 @@ function composePage(
 
   const canvasWrapper = document.createElement("div");
   canvasWrapper.appendChild(canvas);
+  canvasWrapper.style.width = canvas.style.width;
+  canvasWrapper.style.height = canvas.style.height;
   printContainer.appendChild(canvasWrapper);
 
   // A callback for a given page may be executed multiple times for different
--- a/web/firefox_print_service.js
+++ b/web/firefox_print_service.js
@@ -34,8 +34,8 @@ function composePage(
   canvas.height = Math.floor(size.height * PRINT_UNITS);
 
   // The physical size of the canvas as specified by the PDF document.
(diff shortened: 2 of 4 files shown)


Snuffleupagus (changes requested):

web/base_viewer.js
```diff
@@ -1336,6 +1336,7 @@ class BaseViewer {
       return {
         width: viewport.width,
         height: viewport.height,
+        scale: pageView.scale,
```
Snuffleupagus, line 1339: The `pageView.scale` property simply contains the *current* zoom-factor used in the viewer itself, hence I don't really understand how this is relevant for printing!?

Unless I'm completely misremembering/misunderstanding things printing were supposed to be *separate* from the rendering in the viewer, and having the printing output connected to the current viewer rendering thus seem strange.

Snuffleupagus (changes requested):

web/base_viewer.js
emilio: Hmm, so this does fix it, so something is going wrong. Maybe Firefox is copying the zoom factor over to the printed page, let me check.

emilio: Alright, so yeah, this is not the right thing to do.

This patch does fix it at the "automatic zoom level" of course, but the behavior of course starts depending on the zoom factor which is obviously broken.

Do you know where the automatic zoom level is being computed?

Snuffleupagus: > Do you know where the automatic zoom level is being computed?

I'm not sure how helpful it is, but see below; as mentioned it seems strange to connect the printing output to the current viewer state (since they should be independent), so it'd be good to understand *exactly* what goes wrong in the current code.
https://github.com/mozilla/pdf.js/blob/7fef896a9fba6366b22da70747d8eff273ba47e9/web/base_viewer.js#L791-L798

emilio: Yes, agreed. I'll split my other fix in another PR, since they should be independent.

emilio: Ok, so I see what's going wrong. The issue is that the PDF is for an A5 page, rather than A4, but Firefox doesn't support `@page { size }` so pdf.js can't set the right page size for the user.

emilio: So this is invalid, afaict, #13100 is still right though.

PR #13100, opened 2021-03-14
print: Ensure print containers have the right size and don't create overflow.

emilio:
First, there's just no need to do something like this, this is simpler and
closer to what the screen renderer does.

Second, this causes overflow, which Firefox tries to compensate for when
fitting to page width, and fails at it. That is tracked in:

https://bugzilla.mozilla.org/show_bug.cgi?id=1698136

But this bug works around it by not causing overflow.

For modern browsers, we could avoid the duplication setting the style attribute
by using something like width: min/max-content, but this is not a big deal I
think, let me know if you'd prefer that.

Also I had to add a max-height for Chromium not to create extra pages. This
is harmless in Firefox and workarounds the Chromium bug, so so be it.

### DIFF
--- a/web/firefox_print_service.js
+++ b/web/firefox_print_service.js
@@ -39,6 +39,8 @@ function composePage(
 
   const canvasWrapper = document.createElement("div");
   canvasWrapper.appendChild(canvas);
+  canvasWrapper.style.width = canvas.style.width;
+  canvasWrapper.style.height = canvas.style.height;
   printContainer.appendChild(canvasWrapper);
 
   // A callback for a given page may be executed multiple times for different
--- a/web/pdf_print_service.js
+++ b/web/pdf_print_service.js
@@ -206,6 +206,8 @@ PDFPrintService.prototype = {
 
     const wrapper = document.createElement("div");
     wrapper.appendChild(img);
+    wrapper.style.width = img.style.width;
+    wrapper.style.height = img.style.height;
     this.printContainer.appendChild(wrapper);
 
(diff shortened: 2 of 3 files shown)


emilio:
Split out from #13099 since this should still help with one of the bugs and should be uncontroversial.

r? @timvandermeij / @Snuffleupagus / @brendandahl

timvandermeij:
First of all, thank you for looking into some printing bugs!

> Also I had to add a max-height for Chromium not to create extra pages.

This might also resolve some other issues we have open regarding Chrome and extra blank pages during printing.

emilio:
Yeah, if that doesn't solve other issues, we should just add overflow: hidden to that rule. And if _that_ doesn't solve it then I'm out of ideas, but pretty sure it should.

timvandermeij (approved):

timvandermeij:
Looks good to me and I didn't find any issues during testing in various browsers. It doesn't solve all extra blank page issues we have open (see https://github.com/mozilla/pdf.js/issues?q=is%3Aopen+is%3Aissue+label%3A4-printing+blank) but it does seem like a good improvement, and also simpler code. Thank you!

timvandermeij:
(Side note: I cleaned up the issues we have with the `4-printing` label so only the relevant ones are left; see https://github.com/mozilla/pdf.js/issues?q=is%3Aopen+is%3Aissue+label%3A4-printing in case you're looking for more documents to test with / issues to solve)

emilio:
Great, thanks @timvandermeij!

PR #13101, opened 2021-03-14
print: Remove invalid @supports condition.

emilio:
@supports() is not supposed to report support for page descriptors, this is
depending on a Chromium bug, which doesn't treat as invalid:

```
<div style="size: 1pt 1pt">
```

Even though it should. That is
https://bugs.chromium.org/p/chromium/issues/detail?id=1079214

There's no need to use @supports for this. If the descriptor is not accepted it
will just be ignored.

That way, when Firefox implements @page { size }, which is in progress, it will
get the right behavior.

### DIFF
--- a/web/pdf_print_service.js
+++ b/web/pdf_print_service.js
@@ -120,15 +120,7 @@ PDFPrintService.prototype = {
     this.pageStyleSheet = document.createElement("style");
     const pageSize = this.pagesOverview[0];
     this.pageStyleSheet.textContent =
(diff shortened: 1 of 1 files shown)


timvandermeij:
@emilio The CI failed due to a linting error. You can most likely fix it automatically by running `gulp lint --fix`.

emilio:
Thanks, fixed!

timvandermeij (approved):

timvandermeij:
Looks good and successfully tested in various browsers to check that nothing changed about the printing output. Thank you!

PR #13102, opened 2021-03-14
print: RFC: Center when printing a PDF with smaller size than the output page.

emilio:
This builds on top of #13100, but this changes printing behavior intentionally
so I thought it was worth discussing separately, to improve the rendering on
test-cases like the one in https://bugzil.la/1697778.

This matches what e.g.  Evince does when you print the PDF in there on an A4
printer.

We use margins to center horizontally, and flex to center vertically. The
reasoning for this is that it should have better browser support (though maybe
pdf.js no longer supports browsers without flex support?) and it's just as
simple.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -1763,6 +1763,15 @@ html[dir="rtl"] #documentPropertiesOverlay .row > * {
     max-height: 100%;
     page-break-after: always;
     page-break-inside: avoid;
+
+    /* If we're smaller than the page, center the canvas horizontally and
+     * vertically */
+    min-height: 100%;
+    min-width: 100%;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
   }
   #printContainer canvas,
   #printContainer img {


emilio:
* [Before](https://user-images.githubusercontent.com/1323194/111074060-a70bab80-84e1-11eb-9e80-41eee22e3229.png)
* [After](https://user-images.githubusercontent.com/1323194/111074071-b12daa00-84e1-11eb-9963-3560a72fdf99.png)

Snuffleupagus:
> We use margins to center horizontally, and flex to center vertically. The
> reasoning for this is that it should have better browser support (though maybe
> pdf.js no longer supports browsers without flex support?) and it's just as
> simple.

I *think* it's safe to assume that `flex`-related properties can be used without problems in PDF.js now, based on this compatibility information:
 - https://developer.mozilla.org/en-US/docs/Web/CSS/flex#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content#browser_compatibility

However it seems that Firefox-support is *slightly* limited, but I assume that's not a problem?

emilio:
> However it seems that Firefox-support is slightly limited, but I assume that's not a problem?

Limited in what sense? I wasn't able to infer it from the links :)

emilio:
Anyhow using margins for horizontal centering should be fine too, and as I said is about the same amount of complexity.

emilio:
(But happy to change to flex if you think it's better, it's a matter of setting `min-width: 100%` and `align-content: center` instead of `margin-left` / `margin-right`.

Snuffleupagus:
> > However it seems that Firefox-support is slightly limited, but I assume that's not a problem?
> 
> Limited in what sense? I wasn't able to infer it from the links :)

https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction#browser_compatibility lists Firefox as "Partial support", hence why I mentioned it.

emilio:
Ah:

> Before Firefox 81, overflow with *-reverse is unsupported. See bug 1042151.

So that was about a particular bug that has been addressed a while ago.

timvandermeij:
I think centering the page looks much better, and this also makes sense given that other viewers seem to do this as well. I also think using `display: flex` is not a problem anymore given that we dropped support for IE completely and all modern browsers support it well. I do think it would be slightly better here to use `display: flex` also for centering the content horizontally so that the same mechanism is used.

emilio:
Thanks, updated to use flex to center horizontally too.

PR #13104, opened 2021-03-15
Ensure that printing, triggered from scripting, won't accidentally throw in `PDFScriptingManager._updateFromSandbox`

Snuffleupagus:
The issue that this patch fixes is extremely unlikely, but still theoretically possible, and I really should've caught this earlier.
Note how `BaseViewer.pagesPromise` will only be defined when a document is active, see below, and that if a printing event (triggered from scripting) arrives while the document is been closed there's a small chance that the promise isn't defined.

https://github.com/mozilla/pdf.js/blob/eb92ed12f202de4cc8b0060aac1f1a06b26f83a4/web/base_viewer.js#L426-L428

### DIFF
--- a/web/pdf_scripting_manager.js
+++ b/web/pdf_scripting_manager.js
@@ -260,7 +260,7 @@ class PDFScriptingManager {
   /**
    * @private
    */
-  _updateFromSandbox(detail) {
+  async _updateFromSandbox(detail) {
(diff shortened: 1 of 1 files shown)


calixteman (approved):
LGTM.
Thank you for fixing this.

PR #13105, opened 2021-03-15
Improve memory usage around the `BasePdfManager.docBaseUrl` parameter (PR 7689 follow-up)

Snuffleupagus:
While there is nothing *outright* wrong with the existing implementation, it can however lead to increased memory usage in one particular case (that I completely overlooked when implementing this):
For "data:"-URLs, which by definition contains the entire PDF document and can thus be arbitrarily large, we obviously want to avoid sending, storing, and/or logging the "raw" docBaseUrl in that case.

To address this, this patch makes the following changes:
 - Ignore any non-string in the `docBaseUrl` option passed to `getDocument`, since those are unsupported anyway, already on the main-thread.

 - Ignore "data:"-URLs in the `docBaseUrl` option passed to `getDocument`, to avoid having to send what could potentially be a *very* long string to the worker-thread.

 - Parse the `docBaseUrl` option *directly* in the `BasePdfManager`-constructors, on the worker-thread, to avoid having to store the "raw" docBaseUrl in the first place.

### DIFF
--- a/src/core/pdf_manager.js
+++ b/src/core/pdf_manager.js
@@ -13,16 +13,22 @@
  * limitations under the License.
  */
 
-import {
-  createValidAbsoluteUrl,
-  shadow,
-  unreachable,
-  warn,
-} from "../shared/util.js";
+import { createValidAbsoluteUrl, unreachable, warn } from "../shared/util.js";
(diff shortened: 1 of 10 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #13106, opened 2021-03-15
Support `LineAnnotation`s with empty /Rect-entries (issue 6564)

Snuffleupagus:
This extends PR #13033 slightly, with a heuristic to support corrupt PDF documents where the `LineAnnotation`s have an empty /Rect-entry. Please note that while I have no idea if this is "correct", this patch at least makes us output the same /BBox as re-saving in Adobe Reader does.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -23,6 +23,7 @@ import {
   assert,
   escapeString,
   getModificationDate,
+  isArrayEqual,
   isAscii,
   isString,
   OPS,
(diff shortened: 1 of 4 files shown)


src/core/annotation.js
```diff
@@ -2291,6 +2292,17 @@ class LineAnnotation extends MarkupAnnotation {
 
       const borderWidth = this.borderStyle.width;
 
+      // If the /Rect-entry is empty, create a fallback rectangle such that we
+      // get similar rendering/highlighting behaviour as in Adobe Reader.
+      if (isArrayEqual(this.rectangle, [0, 0, 0, 0])) {
+        this.rectangle = [
+          this.data.lineCoordinates[0] - 2 * borderWidth,
```
timvandermeij, line 2299: While it's certainly an improvement that we now actually render the line, I can't help but notice that the canvas rendering is a quite off when compared to the annotation layer rendering. Do you have any idea what could cause that?

Snuffleupagus: Yes, it's a tiny bit off; however it's *not* perfect in Adobe Reader either and given that now we create the same /BBox I think that this is good enough all things considered (since it's the PDF document *itself* which is at fault here).

This is really the best I can come up with, but you're more than welcome to improve it :-)

timvandermeij: Ah, ok. I already expected that a bit, but I figured it couldn't hurt to ask since I can't check how Adobe Reader renders it. Thanks!

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

/botio makeref

src/core/annotation.js
Snuffleupagus: Trust me, I really tried to make it fit *better* but I wasn't able to improve the AnnotationLayer-part without impacting the actual *rendering* negatively :-(
At some point, I just gave up and said to myself that it'll have to be "good enough" for now!

timvandermeij: It's definitely an improvement over rendering nothing, and given that the PDF document itself is corrupted I think this is more than fine for now :-)

PR #13113, opened 2021-03-17
Ignore some *scripting* events which don't make sense in PresentationMode

Snuffleupagus:
A number of the currently supported *scripting* events only make sense in the "normal" viewer mode, and not when PresentationMode is active. For example:
 - Changing the zoom-level will outright break rendering in PresentationMode, since it relies on "page-fit" being used.
 - Focusing a particular (AcroForm) element won't work, and could break keyboard navigation, since forms should not be editable in PresentationMode (see issue 12232).

### DIFF
--- a/web/pdf_scripting_manager.js
+++ b/web/pdf_scripting_manager.js
@@ -261,6 +261,11 @@ class PDFScriptingManager {
    * @private
    */
   async _updateFromSandbox(detail) {
+    // Ignore some events, see below, that don't make sense in PresentationMode.
+    const isInPresentationMode =
+      this._pdfViewer.isInPresentationMode ||
+      this._pdfViewer.isChangingPresentationMode;
+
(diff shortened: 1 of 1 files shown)


calixteman:
Maybe we should disable scripting when in presentation mode, wdyt ?

Snuffleupagus:
> Maybe we should disable scripting when in presentation mode, wdyt ?

My initial though was exactly the same, but there *may* be some cases where scripting actually make sense even in PresentationMode.

See e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1100769, which is currently not working because of an error being thrown in the sandbox, but is probably something that users would expect to work *also* in PresentationMode.

timvandermeij (approved):

timvandermeij:
Looks good to me; thanks!

PR #13114, opened 2021-03-17
built in Text to speech (TTS)

samharwood:
I made this because at my work we have a need for a simple UI to speak PDFs. This is surprising difficult to find. So I had a go at making my own!

I do not expect this to be merged at all as it has many problems i know! 
(a large feature, built on experimental standard! not core PDF functionality, adds UI, hacked together! etc)
but thought I would submit here for discoverability for others searching who have similar needs, and am curious if ppl think something like this belongs and should be considered for the PDF viewer? 

I saw MS Edge has implemented Read Aloud in their PDF reader. But to my frustration it does not work well. (ok if you want to read a document top to bottom none stop. Not so good if you want to select and read a portion of text, which is usually the case.)
I've already tried all the extensions I could find that read webpages, and of those that worked, none were very user friendly (because every interaction requires first clicking the extension button, to open a menu, to click play/pause, etc.)

This is my first attempt at a pull request. Please be gentle!

I put up a demo version here so you can see it in action
https://pdf-tts-demo.azurewebsites.net/pdfjs_tts/web/viewer.html 
Play/Pause on Toolbar
Voice options in Sidebar


Snuffleupagus (changes requested):
> I do not expect this to be merged at all as it has many problems i know!
> (a large feature, built on experimental standard! not core PDF functionality, adds UI, hacked together! etc)

While I totally get that something like this could make sense in a custom implementation, it might not be a great fit for the PDF.js default viewer as-is; especially considering the "built on experimental standard"-part mentioned above.

Note also that *if*, at some (future) point, something like this would be considered it'd require a *fair bit* of additional work to improve the overall implementation (of both the UI and JavaScript parts).

samharwood:
Yep totally agree and understand the scale of work it would require :) Will close this now

PR #13115, opened 2021-03-17
XFA - Add support for few ui elements

calixteman:
- input;
   - layout;   
   - border;   
   - margin;  
   -  color.

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -51,7 +51,7 @@ import {
   getRelevant,
   getStringOption,
 } from "./utils.js";
-import { measureToString, toStyle } from "./html_utils.js";
+import { layoutClass, measureToString, toStyle } from "./html_utils.js";
 import { Util, warn } from "../../shared/util.js";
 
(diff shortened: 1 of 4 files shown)


src/core/xfa/template.js
```diff
@@ -802,6 +1071,10 @@ class Decimal extends ContentObject {
     const number = parseFloat(this[$content].trim());
     this[$content] = isNaN(number) ? null : number;
   }
+
+  [$toHTML]() {
+    return this[$content].toString();
```
brendandahl, line 1076: Looks like this[$content] can be `null`. Would an exception be okay here?

src/core/xfa/template.js
```diff
@@ -1826,6 +2261,10 @@ class Integer extends ContentObject {
     const number = parseInt(this[$content].trim(), 10);
     this[$content] = isNaN(number) ? null : number;
   }
+
+  [$toHTML]() {
+    return this[$content].toString();
```
brendandahl, line 2266: Same null issue here.

brendandahl (approved):
r+ with null handled in the two spots

PR #13116, opened 2021-03-17
Enable scripting by default in the viewer (PR 13053 follow-up)

Snuffleupagus:
Given that https://bugzilla.mozilla.org/show_bug.cgi?id=1699219 has enabled scripting for all Firefox-channels, it seems reasonable to simply set `enableScripting = true` unconditionally in the viewer preferences/options.

For now, this patch leaves the standalone viewer-components alone (such as e.g. `BaseViewer`), and if those are used scripting will thus have to be manually enabled (see e.g. the "simpleviewer"/"singlepageviewer" examples).

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -153,7 +153,7 @@
     },
     "enableScripting": {
       "type": "boolean",
-      "default": false
+      "default": true
     },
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Let's follow upstream here. Thanks!

PR #13117, opened 2021-03-18
Re-factor the default preferences generation to support build targets (PR 10548)

Snuffleupagus:
Originally the default preferences where simply placed in a JSON-file, checked into the repository, which over time became impractical, annoying, and error-prone to maintain; please see PR #10548.
While that improved the overall situation a fair bit, it however inherited one quite unfortunate property of the old JSON-based solution[1]: It's still not possible for *different* build targets to specify their *own* default preference values.

With some preferences, such as e.g. `enableScripting`, it's not inconceivable that you'd want to (at least) support build-specific default preference values. Currently that's not really possible, which is why this PR re-factors the default preferences generation to support this.

*Much smaller/simpler diff with https://github.com/mozilla/pdf.js/pull/13117/files?w=1*

---
[1] This fact isn't really clear from the `AppOptions` implementation, unless you're familiar with the `gulpfile.js` code, which could lead to some confusion for those new to this part of the code-base.

### DIFF
--- a/extensions/firefox/content/PdfJsDefaultPreferences.jsm
+++ b/extensions/firefox/content/PdfJsDefaultPreferences.jsm
@@ -18,5 +18,5 @@
 var EXPORTED_SYMBOLS = ["PdfJsDefaultPreferences"];
 
 var PdfJsDefaultPreferences = Object.freeze(
-  PDFJSDev.json("$ROOT/build/default_preferences.json")
+  PDFJSDev.eval("DEFAULT_PREFERENCES")
 );
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -669,49 +669,17 @@ function buildDefaultPreferences(defines, dir) {
     TESTING: defines.TESTING || process.env.TESTING === "true",
   });
 
-  // Refer to the comment in the 'lib' task below.
-  function babelPluginReplaceNonWebPackRequire(babel) {
-    return {
-      visitor: {
-        Identifier(curPath, state) {
-          if (curPath.node.name === "__non_webpack_require__") {
-            curPath.replaceWith(babel.types.identifier("require"));
-          }
-        },
-      },
-    };
-  }
-  function preprocess(content) {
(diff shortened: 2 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks like a really good idea; thank you for doing this!

Snuffleupagus:
> Looks like a really good idea; thank you for doing this!

As always, thank *you* for review/landing my patches :-)

---

This now leaves only the https://github.com/mozilla/pdf.js/blob/master/extensions/chromium/preferences_schema.json file as a slight annoyance, since it still requires *manual* updates whenever the `AppOptions` change; although I'm not sure if trying to (somehow) generate that one during building is worth the effort.

Looking at https://chrome.google.com/webstore/detail/pdf-viewer/oemmndcbldboiebfnladdacbdfmadadm/ I'm also starting to wonder if the Chromium extension is still being maintained, since the last update is listed as "13 september 2019" which is now a year and a half ago.

timvandermeij:
@Rob--W Do you have more information regarding the Chrome extension? It seems to very out of date; is it still maintained?

Rob--W:
I still support it, but an update is indeed overdue. Due to the age and updates (appearance, experimental JS support), I need to do some testing and perhaps an update of defaults (e.g. enabling dark theme by default) to make sure that users still recognize the UI.

Snuffleupagus:
> experimental JS support

Just a FYI: Scripting will be enabled by default for all user starting in Firefox 88; please see https://bugzilla.mozilla.org/show_bug.cgi?id=1699219

> e.g. enabling dark theme by default

That should now be easy to change for the Chromium extension, by updating https://github.com/mozilla/pdf.js/blob/d426ffdad98106092a9fe138d7bce3aeda36fb69/web/app_options.js#L159 to something like
```javascript
value: typeof PDFJSDev !== "undefined" && PDFJSDev.test("CHROME") ? 2 : 0,
```

PR #13118, opened 2021-03-19
Add landmark region and aria-label for each page.

brendandahl:
Allows screen readers to jump landmarks and announce
the page. Since landmarks must have content, I also added
aria labels for the loading images.

### DIFF
--- a/l10n/en-US/viewer.properties
+++ b/l10n/en-US/viewer.properties
@@ -149,8 +149,8 @@ findbar.title=Find in Document
 findbar_label=Find
 
 additional_layers=Additional Layers
-# LOCALIZATION NOTE (page_canvas): "{{page}}" will be replaced by the page number.
-page_canvas=Page {{page}}
+# LOCALIZATION NOTE (page_landmark): "{{page}}" will be replaced by the page number.
+page_landmark=Page {{page}}
 # Thumbnails panel item (tooltip and alt text for images)
 # LOCALIZATION NOTE (thumb_page_title): "{{page}}" will be replaced by the page
 # number.
@@ -223,6 +223,7 @@ page_scale_actual=Actual Size
 page_scale_percent={{scale}}%
 
 # Loading indicator messages
+loading=Loading…
 loading_error=An error occurred while loading the PDF.
 invalid_file_error=Invalid or corrupted PDF file.
 missing_file_error=Missing PDF file.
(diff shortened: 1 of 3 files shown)


web/pdf_page_view.js
```diff
@@ -122,6 +122,12 @@ class PDFPageView {
     div.style.width = Math.floor(this.viewport.width) + "px";
     div.style.height = Math.floor(this.viewport.height) + "px";
     div.setAttribute("data-page-number", this.id);
+    div.setAttribute("role", "region");
+    this.l10n
+      .get("page_canvas", { page: this.id }, "Page {{page}}")
```
Snuffleupagus, line 127: Is it "correct" to re-use the existing `page_canvas` string here, since this isn't actually a canvas?
Furthermore, with these changes should we still keep setting https://github.com/mozilla/pdf.js/blob/24e598a895b712cbfa1f34deb12bb2c0f3575fee/web/pdf_page_view.js#L620-L622
or can that now become confusing to accessibility software with the *same* aria label being duplicated for *different* elements?

Finally, please remove the "Page {{page}}" part here since it's already handled by https://github.com/mozilla/pdf.js/blob/24e598a895b712cbfa1f34deb12bb2c0f3575fee/web/l10n_utils.js#L48

l10n/en-US/viewer.properties
```diff
@@ -223,6 +223,7 @@ page_scale_actual=Actual Size
 page_scale_percent={{scale}}%
 
 # Loading indicator messages
+loading=Loading…
```
Snuffleupagus, line 226: You need to add this new l10n string *before* line https://github.com/mozilla/pdf.js/blob/24e598a895b712cbfa1f34deb12bb2c0f3575fee/web/l10n_utils.js#L73 as well.

Snuffleupagus (changes requested):

web/pdf_page_view.js
brendandahl: Not really correct to re-use, but we'll end up re-triggering localization of it. I guess that's fine.

With the page div and canvas labeled the screen reader ends up reading it twice which isn't ideal.

brendandahl: I thought i removed the page_canvas, must have missed that change...

Snuffleupagus (approved):
The localization-part looks good to me now, thanks!

Unfortunately I don't know enough about a11y to really review that part, so I'm assuming that the attribute usage is correct :-)

timvandermeij (approved):

PR #13119, opened 2021-03-19
Actually reset the `PDFPageProxy._xfaPromise` property as intended (PR 13069 follow-up)

Snuffleupagus:
Similar to the existing `annotationsPromise` and `_jsActionsPromise` properties, the new `_xfaPromise` should obviously also be reset, since otherwise you might end up holding onto a lot of data for pages that are no longer active.

(That caching wasn't present in the original version of PR #13069, which is why I didn't spot it until now.)


calixteman (approved):
Thank you for fixing this.

PR #13120, opened 2021-03-19
Rotate landscape pages, during printing, by default in the viewer (`enablePrintAutoRotate = true`)

Snuffleupagus:
While this will perhaps not be perfect for *every* PDF document with mixed page orientation, based on the large number of bugs/issues seen over the years I'm however pretty convinced that it'll be an overall improvement in a majority of cases.

In order to improve things further, we'd probably need Firefox to support e.g. `@page` such that the viewer can provide better information to the print engine.


timvandermeij (approved):

timvandermeij:
Good idea!

PR #13125, opened 2021-03-20
Don't provide the `enableXfa` parameter to the `BaseViewer` constructor, and avoid the fallback bar with `enableXfa = true` set (PR 13069 follow-up)

Snuffleupagus:
Given that the `enableXfa` parameter must to be passed to the API/Worker, and thus included in the `getDocument` call, it's not necessary to include it when initializing the `PDFViewer`-instance used in the default viewer. (Also, in `AppOptions`, the parameter is clearly marked with `OptionKind.API`.)

Furthermore, we probably don't want to display the fallback bar (in Firefox) for XFA documents when `enableXfa = true` is set.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -518,7 +518,6 @@ const PDFViewerApplication = {
       useOnlyCssZoom: AppOptions.get("useOnlyCssZoom"),
       maxCanvasPixels: AppOptions.get("maxCanvasPixels"),
       enableScripting: AppOptions.get("enableScripting"),
-      enableXfa: AppOptions.get("enableXfa"),
     });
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Wait, looking once more at this I suppose that line https://github.com/mozilla/pdf.js/blob/c6b44d1516b9b7dd6d80ee2304477e6483e0cbdb/web/app.js#L521 is what should actually be removed here?
Apparently the parameter is necessary already in the `getDocument` call, so I suppose that the PR should be updated to *only* remove that line instead?

timvandermeij (approved):

timvandermeij:
Looks good!

PR #13128, opened 2021-03-22
Re-factor how the `BasePreferences.prefs`-property is initialized; remove some *indirect* loops

Snuffleupagus:
- Re-factor how the `BasePreferences.prefs`-property is initialized

   Looking at this now, I cannot understand why we'd need to initialize `this.prefs` with all of the values from `this.defaults`.
   Not only does this *indirectly* require one extra loop, via the `Object.assign`-call, but it also means that in GENERIC-builds changes to default-preference values might not be picked-up unless the the existing user-prefs are cleared (if the user had *manually* set prefs previously).

 - Remove some *indirect* loops in the `BasePreferences.getAll`-method

   In the `getAll`-method, we can have just one *explicit* loop rather than two indirect ones via the old `Object.assign`-call.

   Also, changes the `get`-method to be slightly more compact (while keeping the logic intact).

### DIFF
--- a/web/preferences.js
+++ b/web/preferences.js
@@ -125,18 +125,13 @@ class BasePreferences {
    */
   async get(name) {
     await this._initializedPromise;
-    const defaultValue = this.defaults[name];
+    const defaultValue = this.defaults[name],
+      prefValue = this.prefs[name];
 
     if (defaultValue === undefined) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13133, opened 2021-03-23
JS - Handle correctly hierarchy of fields

### LINKED ISSUE #13132: JS - actions are not executed in a form
Original pdf:
https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/866023/IHT205_2004-2006.pdf

A copy: [IHT205_2004-2006.pdf](https://github.com/mozilla/pdf.js/files/6191219/IHT205_2004-2006.pdf)

Configuration:
- Web browser and its version: firefox nightly
- Operating system and its version: linux
- PDF.js version: 
- Is a browser extension:

1. Put some numbers into fields for item 11 at page 2

Field `A` under `11.11` should contain the sum of numbers entered in field `11.*`

Nothing in field `A`.

calixteman:
- it aims to fix #13132;
  - annotations can inherit their actions from the parent field;
  - there are some fields which act as a container for other fields:
    - they can be access through js so need to add them with an empty type (nothing in the spec about that but checked in Acrobat);
    - calculation order list (CO) can reference them so need make them through this.getField;
    - getArray method must return kids.
  - field values are number, string, ... depending of their type but nothing in the spec on how to know what's the type:
    - according to the comment for Canonical Format: https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=461
    - it seems that this "type" can be guessed from js action Format (when setting a type in Acrobat DC, the only affected thing is this action).
  - util.scand with an empty string returns the current date.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -64,10 +64,11 @@ class AnnotationFactory {
    * @param {Object} ref
    * @param {PDFManager} pdfManager
    * @param {Object} idFactory
+   * @param {boolean} collectFields
    * @returns {Promise} A promise that is resolved with an {Annotation}
    *   instance.
    */
-  static create(xref, ref, pdfManager, idFactory) {
+  static create(xref, ref, pdfManager, idFactory, collectFields) {
     return pdfManager.ensureCatalog("acroForm").then(acroForm => {
       return pdfManager.ensure(this, "_create", [
         xref,
@@ -75,6 +76,7 @@ class AnnotationFactory {
         pdfManager,
         idFactory,
         acroForm,
+        collectFields,
       ]);
     });
   }
@@ -82,7 +84,7 @@ class AnnotationFactory {
   /**
    * @private
    */
-  static _create(xref, ref, pdfManager, idFactory, acroForm) {
+  static _create(xref, ref, pdfManager, idFactory, acroForm, collectFields) {
     const dict = xref.fetchIfRef(ref);
     if (!isDict(dict)) {
       return undefined;
@@ -103,6 +105,7 @@ class AnnotationFactory {
       id,
       pdfManager,
       acroForm: acroForm instanceof Dict ? acroForm : Dict.empty,
+      collectFields,
     };
 
     switch (subtype) {
@@ -178,15 +181,17 @@ class AnnotationFactory {
         return new FileAttachmentAnnotation(parameters);
 
       default:
-        if (!subtype) {
-          warn("Annotation is missing the required /Subtype.");
-        } else {
-          warn(
-            'Unimplemented annotation type "' +
-              subtype +
-              '", ' +
-              "falling back to base annotation."
-          );
+        if (!collectFields) {
+          if (!subtype) {
+            warn("Annotation is missing the required /Subtype.");
+          } else {
+            warn(
+              'Unimplemented annotation type "' +
+                subtype +
+                '", ' +
+                "falling back to base annotation."
+            );
+          }
         }
         return new Annotation(parameters);
     }
@@ -344,6 +349,31 @@ class Annotation {
       rect: this.rectangle,
       subtype: params.subtype,
     };
+
+    if (params.collectFields) {
+      // Fields can act as container for other fields and have
+      // some actions even if no Annotation inherit from them.
+      // Those fields can be referenced by CO (calculation order).
+      const kids = dict.get("Kids");
+      if (Array.isArray(kids)) {
+        const kidIds = [];
+        for (const kid of kids) {
+          if (isRef(kid)) {
+            kidIds.push(kid.toString());
+          }
+        }
+        if (kidIds.length !== 0) {
+          this.data.kidIds = kidIds;
+        }
+      }
+
+      this.data.actions = collectActions(
+        params.xref,
+        dict,
+        AnnotationActionEventType
+      );
+      this.data.fieldName = this._constructFieldName(dict);
+    }
 
     this._fallbackFontDict = null;
   }
@@ -644,6 +674,15 @@ class Annotation {
    * @returns {Object | null}
    */
   getFieldObject() {
+    if (this.data.kidIds) {
+      return {
+        id: this.data.id,
+        actions: this.data.actions,
+        name: this.data.fieldName,
+        type: "",
+        kidIds: this.data.kidIds,
+      };
+    }
     return null;
   }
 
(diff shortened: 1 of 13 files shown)


src/core/annotation.js
```diff
@@ -332,6 +332,16 @@ class Annotation {
       this._streams.push(this.appearance);
     }
 
+    const kids = dict.getRaw("Kids");
+    const kidsId = [];
```
Snuffleupagus, line 336: Can we please initialize this to `null` instead here, and add `if (!kidsId) { kidsId = []; }` before pushing below, to avoid having to include an empty Array in the data for *every* single Annotation?

src/core/core_utils.js
```diff
@@ -287,19 +287,26 @@ function _collectJS(entry, xref, list, parents) {
...
-      _collectJS(actionDict, xref, list, parents);
-      if (list.length > 0) {
-        actions[action] = list;
+  const additionalActionsDicts = getInheritableProperty({
+    dict,
+    key: "AA",
+    stopWhenFound: false,
+  });
+  if (additionalActionsDicts) {
+    for (let i = additionalActionsDicts.length - 1; i >= 0; i--) {
```
Snuffleupagus, line 296: I'm assuming that there's a particular reason for iterating through this array in *reverse* order, so small comment mentioning why would probably not hurt here (to avoid something changing that later thinking it won't matter).

src/core/core_utils.js
```diff
@@ -287,19 +287,26 @@ function _collectJS(entry, xref, list, parents) {
...
+      const additionalActions = additionalActionsDicts[i];
```
Snuffleupagus, line 297: *Most a pre-existing issue, but:* You probably want to also wrap the loop below in a `if (additionalActions instanceof Dict) { ... }`-check to avoid throwing if the data isn't a Dictionary (e.g. if the PDF is somehow corrupt).

src/core/annotation.js
```diff
@@ -670,6 +692,65 @@ class Annotation {
       stream.reset();
     }
   }
+
+  /**
+   * Construct the (fully qualified) field name from the (partial) field
+   * names of the field and its ancestors.
+   *
+   * @private
+   * @memberof WidgetAnnotation
```
Snuffleupagus, line 701: This no longer looks correct.

src/core/annotation.js
```diff
@@ -343,6 +353,9 @@ class Annotation {
       modificationDate: this.modificationDate,
       rect: this.rectangle,
       subtype: params.subtype,
+      actions: collectActions(params.xref, dict, AnnotationActionEventType),
+      fieldName: this._constructFieldName(dict),
```
Snuffleupagus, line 357: Generally speaking, is it "correct" to always do this for *every* single Annotation since it probably doesn't make sense for a lot of them (e.g. LinkAnnotations. and probably more)?

src/core/annotation.js
```diff
@@ -670,6 +692,65 @@ class Annotation {
...
+   * @private
+   * @memberof WidgetAnnotation
+   * @param {Dict} dict - Complete widget annotation dictionary
+   * @returns {string}
+   */
+  _constructFieldName(dict) {
+    // Both the `Parent` and `T` fields are optional. While at least one of
+    // them should be provided, bad PDF generators may fail to do so.
+    if (!dict.has("T") && !dict.has("Parent")) {
+      warn("Unknown field name, falling back to empty field name.");
```
Snuffleupagus, line 709: *Connect the the comments above:* By always running this method, for every Annotation rather than just WidgetAnnotations, won't this warning now be printed *a lot* for cases where it shouldn't?
Obviously we could remove it and/or change it to `info` instead, but I do wonder about these changes generally.

Snuffleupagus (changes requested):

test/pdfs/.gitignore
```diff
@@ -314,6 +314,7 @@
 !issue10339_reduced.pdf
 !issue4934.pdf
 !issue4650.pdf
+!issue13132.pdf
```
Snuffleupagus, line 317: Also, given that this isn't a reduced test-case but a full PDF document (with unclear copyright situation), this should probably be a *linked* test-case instead.

src/core/annotation.js
```diff
@@ -332,6 +337,30 @@ class Annotation {
...
     }
 
+    let actions = null;
+    let fieldName = null;
+    let kidsId = null;
+    if (params.collectFields) {
+      // Fields can act as container for other fields and have
+      // some actions even if no Annotation inherit from them.
+      // Those fields can be referenced by CO (calculation order).
+      const kids = dict.getRaw("Kids");
```
Snuffleupagus, line 347: Something I didn't think of before, but are you *absolutely* sure that the /Kids entry itself cannot ever be a Reference?

Looking at the PDF specification, specifically https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#page=441, it doesn't say anything about indirect objects but there a risk that PDF documents with such data could exist.

Basically, I'd suggest using `const kids = dict.get("Kids");` instead here to prevent any possible errors. (Note that that method will *not* resolve the elements within the Array, only the Array itself.)

calixteman: Thx and fixed.

src/core/annotation.js
```diff
@@ -332,6 +337,30 @@ class Annotation {
       this._streams.push(this.appearance);
     }
 
+    let actions = null;
+    let fieldName = null;
+    let kidsId = null;
```
timvandermeij, line 342: Given that this is an array of IDs, I'd expect a name like `kidsIds` or maybe `kidIds`, but at least IDs in plural.

src/core/annotation.js
```diff
@@ -343,6 +372,9 @@ class Annotation {
       modificationDate: this.modificationDate,
       rect: this.rectangle,
       subtype: params.subtype,
+      actions,
+      fieldName,
+      kidsId,
```
timvandermeij, line 377: This is all exposed in the API now if we add it here, and that's maybe unwanted here given that this is really only interesting for the scripting code and not for annotation API consumers in general?

Snuffleupagus:
@brendandahl If we need a way to add *linked* test-cases for unit/integration-tests, without also *forcing* them to run as part of the reference-tests, could the following patch be useful?

https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:test-type-other

I've had that locally for awhile, and recently (for unrelated reasons) figured it might make sense to fix-up and submit a PR.

brendandahl:
Yeah, that would be useful. Do you know off hand what happens when the test type doesn't match the current expected ones? Does it default to a load test?

Snuffleupagus:
> Do you know off hand what happens when the test type doesn't match the current expected ones? Does it default to a load test?

Essentially an `eq`-test, as far as I can tell, given https://github.com/mozilla/pdf.js/blob/1a2cdaffc513385fc2f17d0e53d2643f92d5fa81/test/driver.js#L570 where everything that's *not* a `text`-test ends up.

However, in https://github.com/mozilla/pdf.js/blob/1a2cdaffc513385fc2f17d0e53d2643f92d5fa81/test/test.js#L574 we end up throwing since the type isn't a valid one.

PR #13138, opened 2021-03-25
[PDFHistory] Correctly simulate an internal destination in the `pushPage`-method (PR 12493 follow-up)

Snuffleupagus:
The intention, in PR #12493, was that the page we're adding to the browser history should behave as if it were a "regular" internal destination (to properly convey user intent).
Unfortunately, since I didn't consider all the edge-cases correctly, it ended up behaving like a URL-hash instead which obviously wasn't intended. Note that currently this isn't a problem, however it can become an issue (in some cases) with upcoming re-factoring around `PDFHistory` and OpenAction support[1].

---
[1] I've started working on fixing the following TODO, which will require a couple of smaller tweaks here and there: https://github.com/mozilla/pdf.js/blob/9d0ce6e79fc70b1b47d6320c8ee18fd6cbe8467d/web/app.js#L1680-L1681

### DIFF
--- a/web/pdf_history.js
+++ b/web/pdf_history.js
@@ -292,6 +292,8 @@ class PDFHistory {
     }
 
     this._pushOrReplaceState({
+      // Simulate an internal destination, for `this._tryPushCurrentPosition`:
+      dest: null,
       hash: `page=${pageNumber}`,
       page: pageNumber,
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #13140, opened 2021-03-25
[GrabToPan] Remove more IE-specific code, and remove *unconditional* browser sniffing

Snuffleupagus:
- Remove the remaining IE-specific checks from `web/grab_to_pan.js` (PR 12328 follow-up)

   Given that all versions of Internet Explorer are now *explicitly* unsupported, we can remove some more dead code in this file.

 - Remove *unconditional* browser sniffing in `web/grab_to_pan.js`

   According to https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons#browser_compatibility, and manual testing in an up-to-date version of Google Chrome, the *unconditional* browser sniffing in `web/grab_to_pan.js` is no longer necessary.
   In particular, this code is definitely not necessary in MOZCENTRAL-builds.

### DIFF
--- a/web/grab_to_pan.js
+++ b/web/grab_to_pan.js
@@ -194,7 +194,6 @@ let matchesSelector;
 
 // Browser sniffing because it's impossible to feature-detect
 // whether event.which for onmousemove is reliable
-const isNotIEorIsIE10plus = !document.documentMode || document.documentMode > 9;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice find!

PR #13141, opened 2021-03-25
XFA -- Display text content

calixteman:
- display xhtml;
  - allow spaces in xhtml (xfa-spacerun:yes);
  - support column layout;
  - fix some border issues.


Snuffleupagus:
It looks like the `xfa_tohtml_spec.js` test isn't listed in https://github.com/mozilla/pdf.js/blob/master/test/unit/clitests.json, is this intentional or just a simple oversight (since other xfa-tests are present there)?

src/core/xfa/parser.js
```diff
@@ -14,6 +14,7 @@
  */
 
 import {
+  $acceptWhites,
```
brendandahl, line 17: I assume whites = whitespace? If so, let's change to full word here and elsewhere.

src/core/xfa/template.js
```diff
@@ -2053,15 +2088,18 @@ class Font extends XFAObject {
 
   [$toStyle]() {
     const style = toStyle(this, "fill");
-    if (style.color) {
-      if (!style.color.startsWith("#")) {
+    const color = style.background;
+    if (color) {
+      if (color === "#000000") {
+        delete style.background;
```
brendandahl, line 2094: Why do we delete the background if it's black?

src/core/xfa/template.js
```diff
@@ -3520,8 +3595,61 @@ class Text extends ContentObject {
...
+        name: "span",
+        attributes: {
+          class: "xfaRich",
+          style: {},
+        },
+        value: this[$content],
+      };
+
+      if (this[$content].includes("\u2029")) {
+        html.name = "div";
```
brendandahl, line 3610: comment on what we're doing in this block

src/core/xfa/template.js
```diff
@@ -3520,8 +3595,61 @@ class Text extends ContentObject {
...
+            }, [])
+          )
+          .forEach(lines => {
+            html.children.push({
+              name: "p",
+              children: lines,
+            });
+          });
+      } else if (/[\u2028\n]/.test(this[$content])) {
+        html.name = "div";
```
brendandahl, line 3635: comment on what we're doing on this block

test/unit/xfa_parser_spec.js
```diff
@@ -323,16 +323,16 @@ describe("XFAParser", function () {
   </template>
 </xdp:xdp>
       `;
-      const root = new XFAParser().parse(xml)[$dump]();
+      const root = new XFAParser().parse(xml)[$dump](); // console.log(root);window.alert("");
```
brendandahl, line 326: Remove comment.

web/xfa_layer_builder.css
```diff
@@ -123,6 +129,17 @@
   background: Highlight;
 }
 
+.xfaRich {
+  /*width: 100%;*/
```
brendandahl, line 133: remove

brendandahl (changes requested):
Some minor nits.

src/core/xfa/template.js
calixteman: Because it's the default color in the css:
```css
.xfaFont {
  color: black;
```
so there is no need to have this value.

src/core/xfa/xhtml.js
```diff
@@ -65,99 +141,163 @@ function checkStyle(style) {
     .join(";");
 }
 
-class A extends XmlObject {
+const NoWhites = new Set(["body", "html"]);
```
Snuffleupagus, line 144: This should be `NoWhites` -> `NoWhitespace` instead, right?

(I'd suggest just folding this change into a pending XFA-patch, to reduce unnecessary code churn.)

PR #13142, opened 2021-03-25
[GrabToPan] Prefer the standard, rather than a prefixed, `matchesSelector` value

Snuffleupagus:
Given how these checks are currently implemented, they will *always* prefer the prefixed versions over the standard one which does seem quite unfortunate.
This looks especially bad in MOZCENTRAL-builds, since Firefox nowadays support a number of `webkit`-prefixes/methods to improve web-compatiblity, where we thus end up using `webkitMatchesSelector`.

To address this I'm thus proposing that we simply *reverse* the order in which we check for the `matchesSelector` value, which according to the following compatibility information ought to work fine in general:
https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#browser_compatibility

### DIFF
--- a/web/grab_to_pan.js
+++ b/web/grab_to_pan.js
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { shadow } from "pdfjs-lib";
+
 /**
  * Construct a GrabToPan instance for a given HTML element.
  * @param options.element {Element}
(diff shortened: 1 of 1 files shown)


web/grab_to_pan.js
```diff
@@ -176,21 +178,27 @@ GrabToPan.prototype = {
...
-  name += "Selector";
-  if (name in document.documentElement) {
-    matchesSelector = name;
-  }
-  return matchesSelector; // If found, then truthy, and [].some() ends.
-});
+  /**
+   * @private
+   */
+  get _matchesSelector() {
```
Snuffleupagus, line 185: We *could* perhaps even add a fast-path for MOZCENTRAL here, e.g. as outlined below, however given that this code no longer runs *unconditionally* that's probably not necessary.
```javascript
if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")) {
  return shadow(this, "matchesSelector", "matches");
}
```

Snuffleupagus:
> [...] which according to the following compatibility information ought to work fine in general:
> https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#browser_compatibility

Actually, looking at that table a bit closer now, I'm starting to wonder if we could simply remove the `matchesSelector` logic altogether and just use `Element.matches()` directly!?

Snuffleupagus:
Closing in favour of PR #13144.

PR #13144, opened 2021-03-25
[GrabToPan] Use `Element.matches()` rather than prefixed `...matchesSelector` versions

Snuffleupagus:
At this point in time all of the browsers that PDF.js support now implements, with some margin based on the version numbers, `Element.matches()` according to the following compatibility information:
 - https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#browser_compatibility
 - https://caniuse.com/matchesselector


timvandermeij:
Looks good with a rebase.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #13145, opened 2021-03-26
print: Improve rendering of oversized pages.

emilio:
This improves and simplifies #13102 in order to make printing of test-cases
like the one in [bug 1698414](https://bugzil.la/1698414) (where the real page is bigger than the target
page) much better, see incoming screenshots.

The reason why we need to stop setting .style.width / .style.height is to get
the right auto-sizing behavior in both axes. This shouldn't change behavior as
long as the print resolution is >= the CSS resolution, which seems like a
reasonable assumption.

If you try to print with a lower resolution than CSS, then instead of an
stretched canvas, you'd get a centered CSS-quality canvas, which seems
sensible. This could maybe be fixed with some CSS hackery (some combination of
min / max and viewport units perhaps?), but I think it's more trouble than it's
worth.

### DIFF
--- a/web/firefox_print_service.js
+++ b/web/firefox_print_service.js
@@ -14,7 +14,6 @@
  */
 
 import { RenderingCancelledException, shadow } from "pdfjs-lib";
-import { CSS_UNITS } from "./ui_utils.js";
 import { PDFPrintServiceFactory } from "./app.js";
 
 // Creates a placeholder with div and canvas with right size for the page.
@@ -33,14 +32,8 @@ function composePage(
   canvas.width = Math.floor(size.width * PRINT_UNITS);
   canvas.height = Math.floor(size.height * PRINT_UNITS);
 
-  // The physical size of the canvas as specified by the PDF document.
-  canvas.style.width = Math.floor(size.width * CSS_UNITS) + "px";
-  canvas.style.height = Math.floor(size.height * CSS_UNITS) + "px";
-
   const canvasWrapper = document.createElement("div");
   canvasWrapper.appendChild(canvas);
-  canvasWrapper.style.width = canvas.style.width;
-  canvasWrapper.style.height = canvas.style.height;
   printContainer.appendChild(canvasWrapper);
 
   // A callback for a given page may be executed multiple times for different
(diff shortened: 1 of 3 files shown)


emilio:
* before: In order to print the whole thing you need to manually choose the right scale.
 * after

web/firefox_print_service.js
```diff
@@ -33,14 +32,8 @@ function composePage(
   canvas.width = Math.floor(size.width * PRINT_UNITS);
   canvas.height = Math.floor(size.height * PRINT_UNITS);
 
-  // The physical size of the canvas as specified by the PDF document.
-  canvas.style.width = Math.floor(size.width * CSS_UNITS) + "px";
-  canvas.style.height = Math.floor(size.height * CSS_UNITS) + "px";
-
```
Snuffleupagus, line 39: Given that this is being removed, I cannot help but wonder if we still need to keep the following code around?
https://github.com/mozilla/pdf.js/blob/63471bcbbeee3fc1c6b1df988174640a124f097a/web/pdf_print_service.js#L39-L43

---

Obviously I see that those values are being [returned below](https://github.com/mozilla/pdf.js/blob/63471bcbbeee3fc1c6b1df988174640a124f097a/web/pdf_print_service.js#L63-L68), but I don't think that we use them now that you're also removing https://github.com/mozilla/pdf.js/blob/63471bcbbeee3fc1c6b1df988174640a124f097a/web/pdf_print_service.js#L184-L185 in this patch.

emilio: Yup, you're right this can be cleaned up, good catch!

Snuffleupagus (approved):
Looks good to me, thank you!

timvandermeij (approved):

timvandermeij:
Thank you for fixing this, and simplifying the code!

emilio:
Pleasure, thanks both for the review!

dholbert:
It looks like this bug is the suspect for having caused regression https://bugzilla.mozilla.org/show_bug.cgi?id=1711659 , FWIW.

(I ran across this because I noticed that testcase 1 and testcase 2 in https://bugzilla.mozilla.org/show_bug.cgi?id=1693989 fail to actually shrink in Firefox Nightly's print-preview UI and in actual printed output, when I set e.g. `scale: 90%` in the print dialog.  This seems to go back to the PDF.js update https://bugzilla.mozilla.org/show_bug.cgi?id=1702796 , and it seems that bug 1711659 is filed on the regression with a theory that this commit (for bug 1698414) was the relevant change.)

PR #13146, opened 2021-03-26
XFA -- Load fonts permanently from the pdf

calixteman:
- Different fonts can be used in xfa and some of them are embedded in the pdf.
  - Load all the fonts in window.document.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -42,6 +42,7 @@ import {
   isName,
   isRef,
   isStream,
+  Name,
   Ref,
 } from "./primitives.js";
 import {
@@ -49,6 +50,7 @@ import {
   getInheritableProperty,
   isWhiteSpace,
   MissingDataException,
+  validateCSSFont,
   XRefEntryException,
   XRefParseException,
 } from "./core_utils.js";
@@ -852,6 +854,71 @@ class PDFDocument {
 
   get isPureXfa() {
     return this.xfaFactory !== null;
+  }
+
+  async loadXfaFonts(handler, task) {
+    const acroForm = await this.pdfManager.ensureCatalog("acroForm");
+
+    const resources = await acroForm.getAsync("DR");
+    if (!(resources instanceof Dict)) {
+      return;
+    }
+    const objectLoader = new ObjectLoader(resources, ["Font"], this.xref);
+    await objectLoader.load();
+
+    const fontRes = resources.get("Font");
+    if (!(fontRes instanceof Dict)) {
+      return;
+    }
+
+    const partialEvaluator = new PartialEvaluator({
+      xref: this.xref,
+      handler,
+      pageIndex: -1,
+      idFactory: this._globalIdFactory,
+      fontCache: this.catalog.fontCache,
+      builtInCMapCache: this.catalog.builtInCMapCache,
+    });
+    const operatorList = new OperatorList();
+    const initialState = {
+      font: null,
+      clone() {
+        return this;
+      },
+    };
+
+    const fonts = new Map();
+    fontRes.forEach((fontName, font) => {
+      fonts.set(fontName, font);
+    });
+    const promises = [];
+
+    for (const [fontName, font] of fonts) {
+      const descriptor = font.get("FontDescriptor");
+      if (descriptor instanceof Dict) {
+        const fontFamily = descriptor.get("FontFamily");
+        const fontWeight = descriptor.get("FontWeight");
+        const italicAngle = descriptor.get("ItalicAngle");
+        const cssFontInfo = { fontFamily, fontWeight, italicAngle };
+
+        if (!validateCSSFont(cssFontInfo)) {
+          continue;
+        }
+
+        const promise = partialEvaluator.handleSetFont(
+          resources,
+          [Name.get(fontName), 1],
+          /* fontRef = */ null,
+          operatorList,
+          task,
+          initialState,
+          /* fallbackFontDict = */ null,
+          /* cssFontInfo = */ cssFontInfo
+        );
+        promises.push(promise.catch(() => {}));
+      }
+    }
+    await Promise.all(promises);
   }
 
   get formInfo() {
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -95,6 +95,7 @@ const EXPORT_DATA_PROPERTIES = [
   "bold",
   "charProcOperatorList",
   "composite",
+  "cssFontInfo",
   "data",
   "defaultVMetrics",
   "defaultWidth",
@@ -565,6 +566,7 @@ var Font = (function FontClosure() {
     this.loadedName = properties.loadedName;
     this.isType3Font = properties.isType3Font;
     this.missingFile = false;
+    this.cssFontInfo = properties.cssFontInfo;
 
     this.glyphCache = Object.create(null);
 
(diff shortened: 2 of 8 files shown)


src/core/document.js
```diff
@@ -826,6 +827,53 @@ class PDFDocument {
...
+          const fontFamily = descriptor.get("FontFamily");
+          const fontWeight = descriptor.get("FontWeight");
+          const italicAngle = descriptor.get("ItalicAngle");
+
+          partialEvaluator.handleSetFont(
+            resources,
+            [Name.get(pdfFontName), 1],
+            null,
+            operatorList,
+            "",
```
Snuffleupagus, line 867: This doesn't really seem like a good idea, since leaving the `task` both empty and of the wrong type could (with future changes) lead to errors in the `evaluator.js`-code.

At the very least, I think that you should instead construct a "dummy"-task with all of the expected methods/properties (although they can of course be empty).

src/core/fonts.js
```diff
@@ -2928,23 +2930,31 @@ var Font = (function FontClosure() {
...
-      if (!tables["OS/2"] || !validateOS2Table(tables["OS/2"], font)) {
-        tables["OS/2"] = {
-          tag: "OS/2",
-          data: createOS2Table(
-            properties,
-            newMapping.charCodeToGlyphId,
-            metricsOverride
-          ),
+      // When cssFontInfo the font is used to render text in html view
+      // (e.g. with Xfa) so nothing must be moved in private area.
```
Snuffleupagus, line 2934: It feels like there's (some) missing words here, so perhaps instead:
```javascript
      // When `cssFontInfo` is set, the font is used to render text in the HTML
      // view (e.g. with Xfa) so nothing must be moved to the private use area.
```

src/display/api.js
```diff
@@ -2785,7 +2789,7 @@ class WorkerTransport {
     return this.messageHandler.sendWithPromise("GetStats", null);
   }
 
-  startCleanup() {
+  startCleanup(cleanupFontLoader) {
```
Snuffleupagus, line 2792: I want the *default value* to still be to remove the fonts, in order to prevent future issues, hence please change this to `keepFonts = false` instead.

src/display/api.js
```diff
@@ -2800,7 +2804,9 @@ class WorkerTransport {
         }
       }
       this.commonObjs.clear();
-      this.fontLoader.clear();
+      if (cleanupFontLoader) {
```
Snuffleupagus, line 2807: Following the above:
```suggestion
      if (!keepFonts) {
```

src/display/api.js
```diff
@@ -910,7 +910,11 @@ class PDFDocumentProxy {
    * @returns {Promise} A promise that is resolved when clean-up has finished.
    */
   cleanup() {
-    return this._transport.startCleanup();
+    // Don't cleanup the font loader with Xfa since fonts are used in html
```
Snuffleupagus, line 913: Nit: `html` -> `HTML`.

src/display/api.js
```diff
@@ -910,7 +910,11 @@ class PDFDocumentProxy {
...
+    // to display Xfa content.
+    return this._transport.startCleanup(
+      /* cleanupFontLoader */ !this.isPureXfa
```
Snuffleupagus, line 916: With the other changes:
```suggestion
      /* keepFonts = */ this.isPureXfa
```

src/display/api.js
Snuffleupagus: Also, is it not an issue that the fonts will be clean-up in the worker?

Snuffleupagus (changes requested):

src/core/document.js
```diff
@@ -826,6 +827,53 @@ class PDFDocument {
...
+    const fonts = resources.get("Font");
+    if (!(fonts instanceof Dict)) {
+      return Promise.resolve(null);
+    }
+
+    const objectLoader = new ObjectLoader(resources, ["Font"], this.xref);
+    const partialEvaluator = new PartialEvaluator({
+      xref: this.xref,
+      handler,
+      pageIndex: 0,
```
Snuffleupagus, line 841: I do wonder if it'd be better to send a proper "dummy" value, such as e.g. `-1`, given that this isn't actually the first page you're parsing here!?

src/display/api.js
calixteman: We don't need to have them in the worker so it should be fine.

timvandermeij:
PR #13172 got merged today which provides the `keepLoadedFonts` parameter for `PDFDocumentProxy.cleanup` so that this PR can simply use that to make this one smaller.

src/display/font_loader.js
```diff
@@ -385,7 +400,16 @@ class FontFaceObject {
     const data = bytesToString(new Uint8Array(this.data));
     // Add the @font-face rule to the document.
     const url = `url(data:${this.mimetype};base64,${btoa(data)});`;
-    const rule = `@font-face {font-family:"${this.loadedName}";src:${url}}`;
+    let rule;
+    if (!this.cssFontInfo) {
+      rule = `@font-face {font-family:"${this.loadedName}";src:${url}}`;
+    } else {
+      let css = `font-weight: ${this.cssFontInfo.fontWeight};`;
```
brendandahl, line 407: fontWeight, italicAngle, and fontFamily should all be sanitized/validated if we're directly injecting them into css

brendandahl (changes requested):

src/core/document.js
```diff
@@ -827,6 +828,53 @@ class PDFDocument {
...
+            task,
+            initialState,
+            /* fallbackFontDict = */ null,
+            /* cssFontInfo = */ { fontFamily, fontWeight, italicAngle }
+          );
+        }
+      });
+    });
+  }
+
```
Snuffleupagus, line 877: This method is doing a bunch of *synchronous* data lookups, all of which can throw `MissingDataException`s that will break PDF document loading.
Since this method is intended to be called asynchronously, it really needs to guard all of it's data accesses properly and below I'm suggesting a few (untested) changes to prevent issues later on.
```suggestion
  async loadXfaFonts(handler, task) {
    // (Remove this comment) While this is currently *indirectly* guarded by where you're calling the `loadXfaFonts` method from, that's a brittle approach which we should avoid depending on.
    const acroForm = await this.pdfManager.ensureCatalog("acroForm");
    
    const resources = await acroForm.getAsync("DR");
    if (!(resources instanceof Dict)) {
      return;
    }
    const objectLoader = new ObjectLoader(resources, ["Font"], this.xref);
    await objectLoader.load();
    
    const fontRes = resources.get("Font");
    if (!(fontRes instanceof Dict)) {
      return;
    }

    const partialEvaluator = new PartialEvaluator({
      xref: this.xref,
      handler,
      pageIndex: -1,
      idFactory: this._globalIdFactory,
      fontCache: this.catalog.fontCache,
      builtInCMapCache: this.catalog.builtInCMapCache,
    });
    const operatorList = new OperatorList();
    const initialState = {
      font: null,
      clone() {
        return this;
      },
    };

    const fonts = new Map();
    fontRes.forEach((fontName, font) => {
      fonts.set(fontName, font);
    });
    const promises = [];
    
    for ([fontName, font] of fonts) {
      const descriptor = font.get("FontDescriptor");
      if (descriptor instanceof Dict) {
        const fontFamily = descriptor.get("FontFamily");
        const fontWeight = descriptor.get("FontWeight");
        const italicAngle = descriptor.get("ItalicAngle");

        // (Remove this comment) You also need to be able to wait for all fonts to be loaded before returning.
        const promise = partialEvaluator.handleSetFont(
          resources,
          [Name.get(fontName), 1],
          /* fontRef = */ null,
          operatorList,
          task,
          initialState,
          /* fallbackFontDict = */ null,
          /* cssFontInfo = */ { fontFamily, fontWeight, italicAngle }
        );
        // (Remove this comment) If necessary, you could do `promises.push(promise.catch(() => {});` here to prevent one bad font from breaking all font loading.
        promises.push(promise);
      }
    });
    await Promise.all(promises);
  }

```

src/core/worker.js
```diff
@@ -193,6 +193,14 @@ class WorkerMessageHandler {
         pdfManager.ensureDoc("fingerprint"),
         pdfManager.ensureDoc("isPureXfa"),
       ]);
+
+      if (isPureXfa) {
+        const task = new WorkerTask("Load fonts for Xfa");
+        startWorkerTask(task);
+        await pdfManager
+          .loadXfaFonts(handler, task)
+          .finally(() => finishWorkerTask(task));
```
Snuffleupagus, line 202: If the `loadXfaFonts` method would ever *reject*, I believe that this will break all PDF document loading (since `finally` doesn't stop error-propagation); hence please do:
```suggestion
        await pdfManager
          .loadXfaFonts(handler, task)
          .catch(reason => {
            // Ignore errors, to allow the document to load.
          })
          .then(() => finishWorkerTask(task));
```

src/display/font_loader.js
Snuffleupagus: The validation should happen on the worker-side, right?

Snuffleupagus (changes requested):

src/core/document.js
```diff
@@ -854,6 +856,71 @@ class PDFDocument {
...
+    for (const [fontName, font] of fonts) {
+      const descriptor = font.get("FontDescriptor");
+      if (descriptor instanceof Dict) {
+        const fontFamily = descriptor.get("FontFamily");
+        const fontWeight = descriptor.get("FontWeight");
+        const italicAngle = descriptor.get("ItalicAngle");
+        const cssFontInfo = { fontFamily, fontWeight, italicAngle };
+
+        if (!validateCSSFont(cssFontInfo)) {
+          break;
```
Snuffleupagus, line 905: Isn't `continue` what you want here, since otherwise *one* broken font could prevent all fonts from loading?

calixteman: Of course... Thx

Issue #13147, opened 2021-03-26
Uncatchable error thrown when embedded font is broken


keroxp:
Attach (recommended) or Link to PDF file here:

[broken_font.pdf](https://github.com/mozilla/pdf.js/files/6210602/broken_font.pdf)

Configuration:
- Web browser and its version: Firefox@87, Chrome@89, Safari@14
- Operating system and its version: Mac OS@11.2.3
- PDF.js version: 2.7.570
- Is a browser extension: No

Steps to reproduce the problem:
1.  To render page by PDFPageProxy.render()
2. Uncatchable error thrown.

What is the expected behavior? (add screenshot)

- The error should be able to be caught.
- Or correctly rendered.

What went wrong? (add screenshot)

This is console log when opening that pdf by Firefox@87. 

![image](https://user-images.githubusercontent.com/927286/112610038-99e6a900-8e5f-11eb-8aca-05ddba50d3a8.png)

I met identical errors when opening it via pdfjs's API. Well, that file seems to contain the font that is wrongly encoded, but other pdf viewer can show the content.  The message `Warning: Failed to load font 'g_d2_f1': 'SyntaxError: Invalid font data in ArrayBuffer.'.` was logged from [here](https://github.com/mozilla/pdf.js/blob/master/src/display/font_loader.js#L93).

Snuffleupagus:
[issue13147_reduced.pdf](https://github.com/mozilla/pdf.js/files/6211549/issue13147_reduced.pdf)

The CFF in question font has weird/broken header data, since we get bogus data already in https://github.com/mozilla/pdf.js/blob/63471bcbbeee3fc1c6b1df988174640a124f097a/src/core/cff_parser.js#L234-L240

calixteman:
The error comes from:
https://searchfox.org/mozilla-central/source/gfx/ots/src/cff.cc#1194
A dirty (??) workaround is to set `this.bytes[2] = 4;` in the CFFParser and then the display is fine.
Maybe we could set it to 4 (which is the correct value for the header size for this version of CFF) when it isn't 4 and when `major==1 && minor==0` and then print a warning in console.
@Snuffleupagus, any thoughts on that ?

Snuffleupagus:
> Maybe we could set it to 4 (which is the correct value for the header size for this version of CFF) when it isn't 4

[...]

PR #13149, opened 2021-03-26
Set CFF header to 4 when writing it because it contains 4 elements

calixteman:
it aims to fix issue #13147.


src/core/cff_parser.js
```diff
@@ -1556,7 +1556,9 @@ class CFFCompiler {
   }
 
   compileHeader(header) {
-    return [header.major, header.minor, header.hdrSize, header.offSize];
+    // `header.hdrSize` can be any value but we only write 4 values
+    // so header size is 4.
```
Snuffleupagus, line 1560: Please mention that we're doing this to appease OTS, hence the following would be a nice addition to this comment:
```javascript
    // so the header size must be 4 (prevents OTS from rejecting the font).
```

Snuffleupagus (approved):
r=me, with the comment addressed *and* all tests passing; thank you!

PR #13152, opened 2021-03-27
Skip extra objects in object stream in using offsets

### LINKED ISSUE #13130: Rendering font error
Attach (recommended) or Link to PDF file here:  [test.pdf](https://akmer.aikucun.com/a94a8fe5ccb19ba61c4c0873d391e987982fbbd3_1616477914120_83.pdf)

Configuration:
- Web browser and its version: chrome 87.0.4280.88
- Operating system and its version: windows10
- PDF.js version: 2.6.347
- Is a browser extension:

1. just open 

image

calixteman:
It aims to fix #13130.

### DIFF
--- a/src/core/obj.js
+++ b/src/core/obj.js
@@ -2151,12 +2151,13 @@ var XRef = (function XRefClosure() {
           "invalid first and n parameters for ObjStm stream"
         );
       }
-      const parser = new Parser({
+      let parser = new Parser({
         lexer: new Lexer(stream),
         xref: this,
         allowStreams: true,
       });
       const nums = new Array(n);
+      const offsets = new Array(n);
       // read the object numbers to populate cache
       for (let i = 0; i < n; ++i) {
         const num = parser.getObj();
@@ -2172,17 +2173,27 @@ var XRef = (function XRefClosure() {
           );
         }
         nums[i] = num;
+        offsets[i] = offset;
       }
+
+      const start = (stream.start || 0) + first;
       const entries = new Array(n);
       // read stream objects for cache
       for (let i = 0; i < n; ++i) {
+        const length = i < n - 1 ? offsets[i + 1] - offsets[i] : undefined;
+        if (length < 0) {
+          throw new FormatError("Invalid offset in the ObjStm stream.");
+        }
+        parser = new Parser({
+          lexer: new Lexer(
+            stream.makeSubStream(start + offsets[i], length, stream.dict)
+          ),
+          xref: this,
+          allowStreams: true,
+        });
+
         const obj = parser.getObj();
         entries[i] = obj;
-        // The ObjStm should not contain 'endobj'. If it's present, skip over it
-        // to support corrupt PDFs (fixes issue 5241, bug 898610, bug 1037816).
-        if (parser.buf1 instanceof Cmd && parser.buf1.cmd === "endobj") {
-          parser.shift();
-        }
         if (isStream(obj)) {
           continue;
         }
--- a/src/core/stream.js
+++ b/src/core/stream.js
@@ -259,9 +259,15 @@ var DecodeStream = (function DecodeStreamClosure() {
       return bytes;
     },
     makeSubStream: function DecodeStream_makeSubStream(start, length, dict) {
-      var end = start + length;
-      while (this.bufferLength <= end && !this.eof) {
-        this.readBlock();
+      if (length === undefined) {
+        while (!this.eof) {
+          this.readBlock();
+        }
+      } else {
+        var end = start + length;
+        while (this.bufferLength <= end && !this.eof) {
+          this.readBlock();
+        }
       }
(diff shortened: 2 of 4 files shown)


src/core/stream.js
```diff
@@ -123,6 +123,7 @@ var Stream = (function StreamClosure() {
       this.start = this.pos;
     },
     makeSubStream: function Stream_makeSubStream(start, length, dict) {
+      length = length !== undefined ? length : this.bytes.buffer.length;
```
Snuffleupagus, line 126: Can we please use more "consistent" formatting between the various `makeSubStream`-methods?
I.e. let's change this to the follow, since you also don't need to update the `length` when it's already set.
```suggestion
      if (length === undefined) {
        length = this.bytes.buffer.length;
      }
```

src/core/obj.js
```diff
@@ -2172,10 +2173,25 @@ var XRef = (function XRefClosure() {
...
+        offsets[i] = offset;
       }
+
+      const start = (stream.start || 0) + first;
       const entries = new Array(n);
       // read stream objects for cache
       for (let i = 0; i < n; ++i) {
+        const length = i < n - 1 ? offsets[i + 1] - offsets[i] : undefined;
+        if (length < 0) {
+          throw new FormatError("invalid offset in the ObjStm stream");
```
Snuffleupagus, line 2185: Nit: 
```suggestion
          throw new FormatError("Invalid offset in the ObjStm stream.");
```

src/core/obj.js
```diff
@@ -2172,10 +2173,25 @@ var XRef = (function XRefClosure() {
...
+          lexer: new Lexer(
+            stream.makeSubStream(start + offsets[i], length, stream.dict)
+          ),
+          xref: this,
+          allowStreams: true,
+        });
+
         const obj = parser.getObj();
         entries[i] = obj;
         // The ObjStm should not contain 'endobj'. If it's present, skip over it
```
Snuffleupagus, line 2197: This will obviously require testing, but given that you're now *replacing* the parser for each loop iteration my question is if the `endobj` special-case here still necessary with this patch?

test/pdfs/.gitignore
```diff
@@ -36,6 +36,7 @@
 !issue7115.pdf
 !issue7180.pdf
 !issue7769.pdf
+!issue13130.pdf
```
Snuffleupagus, line 39: This ought to be a `link`ed test-case instead, since it's not a reduced one (and the copyright situation is unclear).

test/test_manifest.json
```diff
@@ -18,6 +18,12 @@
        "type": "eq",
        "renderTaskOnContinue": true
     },
+    {  "id": "issue13130",
+       "file": "pdfs/issue13130.pdf",
+       "md5": "318518299132fe3b52252ca43a69a23e",
+       "rounds": 1,
+       "type": "eq"
+    },
```
Snuffleupagus, line 26: You probably don't need to test *all* pages, hence adding `"firstPage": 2,` and `"lastPage": 2,` here would suffice as far as I can tell.

Snuffleupagus (commented):
This looks reasonable, based on a cursory glance at the code, but I've not tested it yet :-)

I've added a couple of comments, but given the nature of the code I'd also like a little time to really think this through before signing off on it.

src/core/stream.js
Snuffleupagus: Actually, is it even necessary to handle the `length === undefined` case at all in *this* method?
Given that https://github.com/mozilla/pdf.js/blob/ba76dd4000c0efa570558dd56d551b773bb4e856/src/core/stream.js#L36 ought to assign a correct `this.end` value already, since `start + undefined === NaN` and we'd fallback to `this.bytes.length`.

src/core/obj.js
calixteman: I looked into `bug1037816.pdf` and yes we have these `endobj` as  garbage at the end of the chunk.
So having a new Parser with the correct starting offset helps to skip those.
So it shouldn't be a regression to remove that stuff.

Snuffleupagus: I've just tested all of the files listed in https://github.com/mozilla/pdf.js/blob/ba76dd4000c0efa570558dd56d551b773bb4e856/src/core/obj.js#L2182 and all of them still work fine with this code removed :-)

---

Can you please also change https://github.com/mozilla/pdf.js/blob/ba76dd4000c0efa570558dd56d551b773bb4e856/test/test_manifest.json#L2040 to an `eq` test instead, as part of this patch, since it's not generally *guaranteed* that just a `load` test catches all kind of /XRef-parsing errors?

Snuffleupagus (approved):
r=me, with the final comment addressed and all tests passing; thank you!

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

PR #13157, opened 2021-03-28
Remove the `file://`-URL special-case from `webViewerOpenFileViaURL`

Snuffleupagus:
Given that *all* data has been loaded on the main-thread, and then transferred to the worker-thread, ever since PR #8617 (almost four years ago) it should no longer be necessary to keep this special-case around.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2274,10 +2274,9 @@ function webViewerInitialized() {
   }
 }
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:

PR #13160, opened 2021-03-30
Improve handling of *linked* test-cases for the unit/integration suites

Snuffleupagus:
- Actually support *linked* test-cases in the integration-tests (in the same way as the unit-tests).

 - Add a new `"type": "other"`-kind to the test-manifest, to support *linked* test-cases in the unit/integration-tests without requiring the PDF document in question to also be a reference-test.

*Please note:* I'm purposely not adding a new case below, since the whole purpose of `other`-tests is that they shouldn't run at all during reference tests, and if we'd ever get that far it's an error.
https://github.com/mozilla/pdf.js/blob/1a2cdaffc513385fc2f17d0e53d2643f92d5fa81/test/test.js#L562-L575

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -392,6 +392,20 @@ var Driver = (function DriverClosure() {
         task.pageNum = task.firstPage || 1;
         task.stats = { times: [] };
 
+        // Support *linked* test-cases for the other suites, e.g. unit- and
+        // integration-tests, without needing to run them as reference-tests.
+        if (task.type === "other") {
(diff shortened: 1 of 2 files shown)


calixteman (approved):
I tested locally with https://github.com/mozilla/pdf.js/pull/13133 and it works well.
Thank you for doing this.

Dragonball39 (commented):
หมายความว่าอะไร แล้วเปนการเเจ้งเตือน ถึงใคร

PR #13164, opened 2021-03-31
Fix intermittent failure with integration test on linux

calixteman:


timvandermeij:
It would be great if this one could get fixed since this color integration test breaks almost all the time now on the bots. If we cannot figure it out we should back it out and make a ticket to re-enable it later on to avoid too much noise in the bot output.

timvandermeij:
Interesting, since a couple of days I haven't seen the integration tests fail anymore on the bots. Maybe browser updates fixed this? Let's try it for a couple more days and if it indeed doesn't fail anymore I guess this is fixed.

calixteman:
Funny, I was thinking the same this morning.

timvandermeij:
I have not seen this fail anymore, so let's close this since it was most likely a browser bug that got fixed.

PR #13165, opened 2021-03-31
[api-minor] Stop exposing the *raw* `defaultAppearance`-string on Annotation-instances

Snuffleupagus:
The reasons for making this change are:
 - This property is not, nor has it ever been, used anywhere in the PDF.js display-layer.
 - Related to the previous point, the format of the `defaultAppearance`-string is such that it'd be difficult to use it as-is in the display-layer anyway.
 - It (usually) contains the "raw" appearance-string, from the PDF document, which is neither parsed nor validated and could thus be bogus.
 - We now expose a `defaultAppearanceData`-property, which is first of all used in the display-layer and secondly contains actually parsed/validated data.
 - In the event that a third-party implementation needs the `defaultAppearance`-string, it could be easily constructed from the recently added `defaultAppearanceData`-property.

All-in-all, I'm thus suggesting that we stop exposing an unused and unnecessary property on all Annotation-instances.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -1115,14 +1115,15 @@ class WidgetAnnotation extends Annotation {
     data.defaultFieldValue = this._decodeFormValue(defaultFieldValue);
 
     data.alternativeText = stringToPDFString(dict.get("TU") || "");
+
     const defaultAppearance =
-      getInheritableProperty({ dict, key: "DA" }) ||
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks! It was indeed never really intended to expose this in the first place.

PR #13169, opened 2021-04-01
[api-minor] Change the format of the `fontName`-property, in `defaultAppearanceData`, on Annotation-instances (PR 12831 follow-up)

Snuffleupagus:
Currently the `fontName`-property contains an actual /Name-instance, which is a problem given that its fallback value is an empty string; see https://github.com/mozilla/pdf.js/blob/ca7f546828603d15ac0975f6131669321bfccceb/src/core/default_appearance.js#L35
The reason that this is a problem can be seen in https://github.com/mozilla/pdf.js/blob/ca7f546828603d15ac0975f6131669321bfccceb/src/core/primitives.js#L30-L34, since an empty string short-circuits the cache. Essentially, in PDF documents, a /Name-instance cannot be empty and the way that the `DefaultAppearanceEvaluator` does things is unfortunately not entirely correct.

Hence the `fontName`-property is changed to instead contain a string, rather than a /Name-instance, which simplifies the code overall.

*Please note:* I'm tagging this patch with "[api-minor]", since PR #12831 is included in the current pre-release (although we're not using the `fontName`-property in the display-layer).

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -1470,7 +1470,7 @@ class WidgetAnnotation extends Annotation {
     const { fontName, fontSize } = this.data.defaultAppearanceData;
     await evaluator.handleSetFont(
       this._fieldResources.mergedResources,
-      [fontName, fontSize],
+      [fontName && Name.get(fontName), fontSize],
       /* fontRef = */ null,
       operatorList,
       task,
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13172, opened 2021-04-02
[api-minor] Add an option, in `PDFDocumentProxy.cleanup`, to allow fonts to remain attached to the DOM

Snuffleupagus:
As mentioned in the JSDoc comment, this should not be used unless you know what you're doing, since it will lead to increased memory usage. However, in some situations (e.g. SVG-rendering), we still want to be able to run general clean-up on both the main/worker-thread while keeping loaded fonts attached to the DOM.[1]

As part of these changes, `WorkerTransport.startCleanup` is converted to an async method and we'll also skip clean-up when destruction has started (since it's redundant).

---
[1] The SVG-rendering mode is obviously not officially supported, since it's both rather incomplete and inherently slower. However with recent changes, whereby we cache repeated images on the document rather than the page level, memory usage can be *a lot* worse than before if we never attempt to release e.g. cached image-data when the viewer is in SVG-rendering mode.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -921,10 +921,13 @@ class PDFDocumentProxy {
    * NOTE: Do not, under any circumstances, call this method when rendering is
    * currently ongoing since that may lead to rendering errors.
    *
+   * @param {boolean} [keepLoadedFonts] - Let fonts remain attached to the DOM.
+   *   NOTE: This will increase persistent memory usage, hence don't use this
+   *   option unless absolutely necessary. The default value is `false`.
    * @returns {Promise} A promise that is resolved when clean-up has finished.
    */
-  cleanup() {
-    return this._transport.startCleanup();
+  cleanup(keepLoadedFonts = false) {
+    return this._transport.startCleanup(keepLoadedFonts || this.isPureXfa);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
This is inspired by, and extends, the functionality implemented in https://github.com/mozilla/pdf.js/pull/13146/files#diff-082d6b37ad01db7ac97cc07c6ddb0dc52040484c5ef91b110b072f50144d9f39 since as outlined above this makes general sense to expose in the API.
Depending on how soon that PR will land, we should perhaps hold off on *this one* to avoid unnecessary merge conflicts!?

Snuffleupagus:
This is inspired by, and extends, the functionality implemented in https://github.com/mozilla/pdf.js/pull/13146/files#diff-082d6b37ad01db7ac97cc07c6ddb0dc52040484c5ef91b110b072f50144d9f39 since as outlined above this makes general sense to expose in the API.
Depending on how soon that PR will land, we should perhaps hold off on *this one* to avoid unnecessary merge conflicts!?

timvandermeij (approved):

timvandermeij:
Looks good! Given that a new release with `[api-minor]` is coming up, I would like to have this included.

PR #13175, opened 2021-04-03
[GENERIC viewer] Avoid data loss in forms, by triggering saving when the document is closed (issue 12257)

### LINKED ISSUE #12257: Notify the user about form data loss when another document is opened
Since #12241 we have notifications for this when the user navigates away from the viewer, but we do not yet warn when switching documents, which means that if you call e.g. `PDFViewerApplication.open` with a new document (or directly call `BaseViewer.setDocument` when using the viewer components standalone) there won't be any warning about losing form data. This should be addressed.

Snuffleupagus:
As discussed in the issue, this is a small/simple patch that should help to prevent *outright* data loss in forms when a new document is opened in the GENERIC viewer.

While the implementation is perhaps a bit "simplistic", it does seem to work and should be fine given that this is an edge-case only relevant for the GENERIC viewer.

Fixes #12257

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -802,7 +802,19 @@ const PDFViewerApplication = {
     }
 
     if (!this.pdfLoadingTask) {
-      return undefined;
+      return;
+    }
+    if (
+      (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) &&
+      this.pdfDocument?.annotationStorage.size > 0 &&
+      this._annotationStorageModified
+    ) {
+      try {
+        // Trigger saving, to prevent data loss in forms; see issue 12257.
+        await this.save({ sourceEventType: "save" });
+      } catch (reason) {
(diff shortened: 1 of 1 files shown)


web/app.js
```diff
@@ -802,7 +802,15 @@ const PDFViewerApplication = {
     }
 
     if (!this.pdfLoadingTask) {
-      return undefined;
+      return;
+    }
+    if (
+      (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) &&
+      this.pdfDocument?.annotationStorage.size > 0 &&
```
Snuffleupagus, line 809: Technically the following check *should* be sufficient here, but this line cannot hurt and should help avoid triggering saving unnecessarily.

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #13184, opened 2021-04-05
Convert objects to sets in places where we only track keys

timvandermeij:


Snuffleupagus (approved):
Looks good to me, thank you!

Issue #7026, opened 2016-02-24
Some PDFs trigger canvas fingerprinting warning false-positives


xcolour:
We have observed that some PDFs cause canvas fingerprinting (https://en.wikipedia.org/wiki/Canvas_fingerprinting) warnings to trigger in certain browsers. This issue seems to trigger specifically in image-heavy PDFs.

For example: https://www.aclu.org/foia-document/some-key-sso-cyber-milestone-dates-fall-2005.

We have seen warnings trigger in the most recent version of the Tor Browser Bundle (5.5.2) as well as with the Chrome extension Canvas Fingerprint Block (https://chrome.google.com/webstore/detail/canvasfingerprintblock/ipmjngkmngdcdpmgmiebdmfbkcecdndc).

This issue was noted a while back by the Tor project (https://trac.torproject.org/projects/tor/ticket/10570). They appear to have resolved it by whitelisting the locally-available version of PDF.js provided by Firefox, but that fix doesn't seem to apply to PDF.js when it's served by the website.

We use the PDF.js provided via the "PDF" Drupal extension version "7.x-1.6" with PDF.js version "1.1.215".

I understand that canvas fingerprinting is an intentionally obfuscated de-anonymization technique and that detection is necessarily heuristic. It may be that PDF.js is doing something that is legitimately indistinguishable from canvas fingerprinting, but it would be nice if there were a way to avoid tripping the detection algorithms. Some of our more privacy-conscious site visitors have noted that it's a bit creepy to be warned by your browser that you're being tracked while trying to read FOIA docs!

Thanks for reading and let me know if this is too open-ended for your issue tracker. I'll be happy to re-submit to your dev mailing list.

I've opened a parallel ticket for the Tor Browser team: https://trac.torproject.org/projects/tor/ticket/18390#ticket.

yurydelendik:
To understand the issue we need to know algorithm for detecting of canvas fingerprinting. If it's just a generic CANVAS access there is nothing we can do. However if it's just access to some specific function such as getImageData or toDataURL then there might be change to fix that. We are using toDataURL to generate thumbs. I wonder if we can detect that and disable this optimization for tor.

xcolour:
I got a brief response from the Tor devs, and it sounds like any attempt to peek at the contents of the canvas will trigger their check. I haven't had a chance to dig into the code yet, but they specifically mentioned `getImageData` and `to{Blob,Url,File,...}`. I would guess there are more. I'll try to get you a complete list.

[...]

PR #13197, opened 2021-04-07
Improve the image quality of thumbnails rendered by `PDFThumbnailView.draw` (issue 8233)

### LINKED ISSUE #8233: Thumbnail downsampling is bad if thumbnail is rendered before full view of that page
Document is copyright protected

Configuration:
- Web browser and its version: Chrome 54.0.2840.71
- Operating system and its version: Windows 7 Build 7061
- PDF.js version: 1.7.225
- Is an extension: no

1. Load file in pdf.js
2. View thumbnails

It works fine in OS X (though image downsampling looks better on the first two pages) :

mac

In Windows the first thumbnails (the pages that have been rendered in full view) are fine. The following thumbnails which are not rendered in full view look like the downsampling didnt work.

screen shot 2017-04-04 at 10 41 02

screen shot 2017-04-04 at 10 41 20

Snuffleupagus:
The reason for the fairly large discrepancy, in the thumbnail quality, between the `draw`/`setImage`-methods is that in the former case we *directly* render the thumbnails at the final size that they'll appear at in the sidebar. In the latter case, we instead downsize the (generally) much larger "regular" pages.

To address this, I'm thus proposing that we let `PDFThumbnailView.draw` render thumbnails at *twice* their intended size and then downsize them to the final size.
Obviously this will increase *peak* memory usage during thumbnail rendering in `PDFThumbnailView.draw`, since doubling the width/height of a `canvas` will lead to its pixel-count increasing by a factor of `4`. Furthermore, since you need four components per pixel (given that it's RGBA-data), this will thus lead to the *temporary* thumbnail `canvas`-sizes increasing by a factor of `16` during rendering. Hence why rendering thumbnails at their "original" scale, i.e. using something like `PDFPageProxy.getViewport({ scale: 1 });`, would be an absolutely terrible idea!

To reduce the size and scope of these changes, I've tried to re-factor and re-use as much of the existing downsizing-implementation already present in `PDFThumbnailView` as possible.

While this will generally *not* make thumbnails rendered by `PDFThumbnailView.draw` look *identical* to those based on the rendered pages (via `PDFThumbnailView.setImage`), it's a considerable improvement as far as I'm concerned and enough to call the issue fixed.

*Please note:* This patch will not lead to *any* additional overhead, in either memory usage or parsing, for thumbnails which are based on the rendered pages.

Fixes #8233

### DIFF
--- a/web/pdf_thumbnail_view.js
+++ b/web/pdf_thumbnail_view.js
@@ -181,10 +181,7 @@ class PDFThumbnailView {
 
     this.div.removeAttribute("data-loaded");
     const ring = this.ring;
-    const childNodes = ring.childNodes;
-    for (let i = childNodes.length - 1; i >= 0; i--) {
-      ring.removeChild(childNodes[i]);
-    }
+    ring.textContent = ""; // Remove the thumbnail from the DOM.
     const borderAdjustment = 2 * THUMBNAIL_CANVAS_BORDER_WIDTH;
     ring.style.width = this.canvasWidth + borderAdjustment + "px";
     ring.style.height = this.canvasHeight + borderAdjustment + "px";
--- a/web/pdf_thumbnail_view.js
+++ b/web/pdf_thumbnail_view.js
@@ -17,6 +17,7 @@ import { getOutputScale } from "./ui_utils.js";
 import { RenderingCancelledException } from "pdfjs-lib";
 import { RenderingStates } from "./pdf_rendering_queue.js";
 
+const DRAW_UPSCALE_FACTOR = 2; // See comment in `PDFThumbnailView.draw` below.
 const MAX_NUM_SCALING_STEPS = 3;
 const THUMBNAIL_CANVAS_BORDER_WIDTH = 1; // px
 const THUMBNAIL_WIDTH = 98; // px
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
I needed some time to think through all of the code-paths involved here (hence why it's submitted as a "draft"), and hopefully this patch now covers all the edge-cases involved here. As mentioned above this isn't *perfect*, but I hope it's deemed good enough since we need to consider the memory usage as well.

/botio-linux preview

Snuffleupagus:
*Off-topic for this PR, but:* I'm wondering if we should simply remove https://github.com/mozilla/pdf.js/blob/6429ccc0028ecfadca4a4db3a0baf39146325379/web/pdf_thumbnail_view.js#L35-L37 since it is/was never actually used and we don't actually expose a way to enable it (e.g. through `AppOptions`) besides manually editing the code?

In cases where access to `canvas`-data is blocked, often done to reduce fingerprinting, we'll have more general rendering issues (affecting pages) given some of the `canvas`-accesses in e.g. [src/display/canvas.js](https://github.com/mozilla/pdf.js/blob/master/src/display/canvas.js) and [src/display/text_layer.js](https://github.com/mozilla/pdf.js/blob/master/src/display/text_layer.js).

timvandermeij (approved):

timvandermeij:
Looks better to me too. Thank you for looking into this!

Regarding `disableCanvasToImageConversion`, I _believe_ this was introduced indeed for avoiding canvas fingerprinting, in particular in the Tor browser (Firefox fork), but I don't know if it was actually used. All I can find is https://github.com/mozilla/pdf.js/issues/7026.

PR #13198, opened 2021-04-07
Fixed the dark mode when using chrome

JovaniTarnowski:
Implemnented pdf.js in my application, but when trying to open the application on chrome if my browser is in Dark mode the pdf viewer is in light mode and the orther way around. This is fixed getting rid of "prefers-color-scheme: dark".

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -97,7 +97,7 @@
   --secondaryToolbarButton-documentProperties-icon: url(images/secondaryToolbarButton-documentProperties.svg);
 }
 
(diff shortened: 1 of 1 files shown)


timvandermeij:
This is unfortunately not something we can accept since this is the primary mechanism to switch between light and dark theme, and this has been confirmed to work correctly. Moreover, removing this would cause the dark theme to _always_ be used. If Chrome or your application is doing something different, that's either a browser bug or an application bug.

PR #13209, opened 2021-04-09
Remove the `enableScripting` option from the `PDFPageView` constructor

Snuffleupagus:
Scripting, as implemented, requires access to a complete document/viewer in order to work. Hence it doesn't really make sense to keep the `enableScripting`-option on `PDFPageView`-instances.[1]

---
[1] Note that there's the `PDFSinglePageViewer`, which can be used in cases where you want access to all features/functionality of the viewer but only display *one* page at a time.


timvandermeij (approved):

timvandermeij:
Looks good to me; thanks!

PR #13210, opened 2021-04-09
Cache babel source map.

brendandahl:
When source is already cached and you reload,
the source map is lost which makes debugging
async functions difficult.

### DIFF
--- a/external/systemjs/plugin-babel-cached.js
+++ b/external/systemjs/plugin-babel-cached.js
@@ -43,7 +43,7 @@ function getDb() {
   return dbPromise;
 }
 
-function storeCache(address, hashCode, translated, format) {
+function storeCache(address, hashCode, translated, format, sourceMap) {
   return getDb().then(function (db) {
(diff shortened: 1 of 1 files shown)


Snuffleupagus (approved):
WFM, thank you!

*Off-topic, but:* I really hope that [bug 1247687](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687) is fixed sooner rather than later, since that's the only thing preventing us from removing the last usage of SystemJS from the development viewer.

timvandermeij (approved):

timvandermeij:
Thank you!

brendandahl:
> _Off-topic, but:_ I really hope that [bug 1247687](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687) is fixed sooner rather than later, since that's the only thing preventing us from removing the last usage of SystemJS from the development viewer.

That would be nice! Unfortunately, last time I asked (month or two ago) no one was actively working on it.

PR #13212, opened 2021-04-10
Check that the correct `pdfDocument` is still active, before rendering the outline/attachments/layers

Snuffleupagus:
*This patch fixes some technical debt in the viewer.*

Given that most API methods are (purposely) asynchronous, there's always a risk that the viewer could have been `close`d before the requested data arrives.
Lately we've started to check this case before using the data, to prevent errors and/or inconsistent state, however the outline/attachments/layers fetching and rendering is old enough that it pre-dates those checks.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1383,14 +1383,23 @@ const PDFViewerApplication = {
 
     onePageRendered.then(() => {
       pdfDocument.getOutline().then(outline => {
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13214, opened 2021-04-10
Display widget signature

calixteman:
- but don't validate them for now;
  - Firefox will display a bar to warn that the signature validation is not supported (see https://bugzilla.mozilla.org/show_bug.cgi?id=854315)
  - almost all (all ?) pdf readers display signatures;
  - validation is done in edge but for now it's behind a pref.


l10n/en-US/chrome.properties
```diff
@@ -15,5 +15,6 @@
 # Chrome notification bar messages and buttons
 unsupported_feature=This PDF document might not be displayed correctly.
 unsupported_feature_forms=This PDF document contains forms. The filling of form fields is not supported.
+unsupported_feature_signatures=This PDF document contains digital signatures. The signatures validation is not supported.
```
Snuffleupagus, line 18: While I'm obviously not a native English speaker, the second sentence sounds a bit weird to me. How about the following instead?
```suggestion
unsupported_feature_signatures=This PDF document contains digital signatures. Validation of signatures is not supported.
```

src/core/annotation.js
```diff
@@ -1201,7 +1195,7 @@ class WidgetAnnotation extends Annotation {
   getOperatorList(evaluator, task, renderForms, annotationStorage) {
     // Do not render form elements on the canvas when interactive forms are
     // enabled. The display layer is responsible for rendering them instead.
-    if (renderForms) {
+    if (!this._isSignature && renderForms) {
```
Snuffleupagus, line 1198: Do we really need a boolean for this, i.e. can't we remove that property and instead do the following (to avoid having to track the kind twice)?
```suggestion
    if (renderForms && !(this instanceof SignatureWidgetAnnotation)) {
```

src/core/document.js
```diff
@@ -894,6 +901,7 @@ class PDFDocument {
       IsAcroFormPresent: this.formInfo.hasAcroForm,
       IsXFAPresent: this.formInfo.hasXfa,
       IsCollectionPresent: !!this.catalog.collection,
+      AreSignaturesPresent: this.formInfo.hasSignatures,
```
Snuffleupagus, line 904: For consistency with the other properties, since this is mostly used internally anyway, can we please call this `IsSignaturesPresent` instead (despite that being grammatically wrong)?

Also, please add the new property to the *existing* tests in https://github.com/mozilla/pdf.js/blob/master/test/unit/api_spec.js

web/app.js
```diff
@@ -1581,6 +1581,9 @@ const PDFViewerApplication = {
     ) {
       console.warn("Warning: Interactive form support is not enabled");
       this._delayedFallback(UNSUPPORTED_FEATURES.forms);
+    } else if (info.AreSignaturesPresent) {
```
Snuffleupagus, line 1584: Shouldn't this always be done when signatures are present, since as-is the warning might not actually display otherwise in some cases?

Snuffleupagus (changes requested):

Snuffleupagus (commented):
Do we have any existing reference-tests which will cover these changes? If not, can you please add one?

(We probably want to defer this to a follow-up, but we should perhaps consider if/how we want to display a warning in the GENERIC-viewer as well.)

Snuffleupagus (approved):

test/test_manifest.json
```diff
@@ -389,6 +389,12 @@
        "rounds": 1,
        "type": "load"
     },
+    {  "id": "bug854315",
+       "file": "pdfs/bug854315.pdf",
+       "md5": "675754c07f71c068d2997905a5456f05",
+       "rounds": 1,
+       "type": "eq"
+    },
```
timvandermeij, line 397: This needs `"annotations": true` too I think since otherwise those are not rendered right now.

Snuffleupagus: I'm probably missing something trivial here, but shouldn't signatures render *automatically* on the canvas now?
Given this code: https://github.com/mozilla/pdf.js/blob/86374d4f75ef1ac599ee3a1431311d614075faed/src/core/document.js#L361-L407

---

Again I might be way off, but that `annotations: true` is intended for things normally rendered in the AnnotationLayer (in the viewer) which shouldn't apply here!?

Snuffleupagus: I've *just* tested locally, and a "normal" `eq`-test seems sufficient here!

timvandermeij:
@calixteman Unfortunately I don't have enough access rights to be able to comment on the upstream patch in Phabricator, but I found a bug in there: `} else if ("signatures") {` always evaluates to true. I think you meant `} else if (featureId === "signatures") {` there, but the `featureId ===` part fell off.

test/test_manifest.json
calixteman: @Snuffleupagus I just tested locally too and same result.

timvandermeij: Whoops, that's indeed true. In that case it's indeed not necessary.

timvandermeij (approved):

timvandermeij:
Looks good to me too; thanks! Let's do this, especially since we can now properly warn users about the missing validation and this has been requested very often.

/botio makeref

l10n/en-US/chrome.properties
```diff
@@ -15,5 +15,6 @@
 # Chrome notification bar messages and buttons
 unsupported_feature=This PDF document might not be displayed correctly.
 unsupported_feature_forms=This PDF document contains forms. The filling of form fields is not supported.
+unsupported_feature_signatures=This PDF document contains digital signatures. Validation of signatures is not supported.
```
sylvestre, line 18: "not yet" ?

cetinsert:
How can this be implemented for the SVG renderer?

cetinsert:
@calixteman How can this be implemented for the SVG renderer?

Snuffleupagus:
> Can this be implemented for the SVG renderer as well?

Please note https://github.com/mozilla/pdf.js/issues/13219#issuecomment-817267765

stephanrauh:
@calixteman @timvandermeij Awesome! You've merged the PR roughly the same time when you've published the new version, so I'd like to ask if this is a feature of version 2.9? Or did you already include it in version 2.8?

Snuffleupagus:
> You've merged the PR roughly the same time when you've published the new version, so I'd like to ask if this is a feature of version 2.9? Or did you already include it in version 2.8?

This patch is not included in any official release yet, since it landed *after* the last release; see https://github.com/mozilla/pdf.js/releases and compare the date of this patch being merged with the date of the latest release :-)

Also, please keep in mind that *all* PDF.js versions consist of *three* numbers, e.g. the current pre-release is `2.8.335` and the stable release is `2.7.570`. (In order to reduce confusion, it's always best to use full version numbers.)

PR #13216, opened 2021-04-10
Remove obsolete done callbacks from the unit tests

timvandermeij:
The done callbacks are an outdated mechanism to signal Jasmine that a unit test is done, mostly in cases where a unit test needed to wait for an asynchronous operation to complete before doing its assertions. Nowadays a much better mechanism is in place for that, namely simply passing an asynchronous function to Jasmine, so we don't need callbacks anymore (which require more code and may be more difficult to reason about).

In these particular cases though the done callbacks never had any real use since nothing asynchronous happens in these places. Synchronous functions don't need to use done callbacks since Jasmine simply knows it's done when the function reaches its normal end, so we can safely get rid of these callbacks. The telltale sign is if the done callback is used unconditionally at the end of the function.

This is all done in an effort to over time get rid of all callbacks in the unit test code.

### DIFF
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -194,10 +194,9 @@ describe("annotation", function () {
   describe("getQuadPoints", function () {
     let dict, rect;
 
-    beforeEach(function (done) {
+    beforeEach(function () {
       dict = new Dict();
       rect = [];
-      done();
     });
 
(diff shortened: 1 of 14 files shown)


Snuffleupagus (commented):
Looks good to me, thank you for doing this!

PR #13217, opened 2021-04-10
[GENERIC viewer] Display a warning message, using the `errorWrapper`, for documents with signatures (PR 13214 follow-up, issue 13215)

### LINKED ISSUE #13215: Implement a warning for unvalidated digital signatures in the generic viewer
In PR #13214 we implemented rendering digital signature annotations, but they are not validated yet. For mozilla-central we display the fallback bar with a clear warning that the signatures are not validated so the user is aware of this, but we have no such mechanism for the generic viewer yet.

I'm thinking we could implement something similar to the fallback bar, which might even be more generally useful for other kinds of messages we now display in the fallback bar. For mozilla-central we'd then use the current browser fallback bar, and for other targets we would display the PDF.js fallback bar instead.

Alternatively we could perhaps add a semi-transparent overlay over the signature.

Snuffleupagus:
To simplify the overall implementation, given that it only applies to the GENERIC-viewer, this patch purposely re-uses the existing `errorWrapper`-functionality to display the message.
While that one is mostly intended for actual *errors*, by re-using it here we considerably reduce the amount of code/complexity necessary for supporting this new warning. It's obviously possible to re-factor/improve this later on, but the patch should do just fine here since it'll indeed inform users (of the GENERIC-viewer) about unverified signatures.

Finally this patch also tweaks the background-color of the `errorWrapper`, making it 20 percent lighter respectively darker (depending on the theme) to make it "stand out" a little bit *less*.[1] While it may perhaps be useful to re-style/re-factor the `errorWrapper`, this patch probably isn't the right place for doing that.

Fixes #13215

---
[1] Note how in the MOZCENTRAL-viewer, which instead uses the browser notification-bar, we're purposely using a neutral colour to not draw too much attention to the notification-bar.

### DIFF
--- a/l10n/en-US/viewer.properties
+++ b/l10n/en-US/viewer.properties
@@ -246,3 +246,6 @@ password_cancel=Cancel
 printing_not_supported=Warning: Printing is not fully supported by this browser.
 printing_not_ready=Warning: The PDF is not fully loaded for printing.
 web_fonts_disabled=Web fonts are disabled: unable to use embedded PDF fonts.
+# LOCALIZATION NOTE (unsupported_feature_signatures): Should contain the same
+# exact string as in the `chrome.properties` file.
+unsupported_feature_signatures=This PDF document contains digital signatures. Validation of signatures is not supported.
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for doing this! The error wrapper is actually better than I initially remembered in terms of styling, so I don't think we'll have to do anything about it for now.

PR #13220, opened 2021-04-11
Ensure that `currentOutlineButton` enabling/disabling won't depend on the exact timing of the "pagesloaded" event (PR 13212 follow-up)

Snuffleupagus:
I (unsurprisingly) managed to forget about handling the case where a "pagesloaded" event arrives *before* the outline has been parsed, in which case we'd not actually enable the `currentOutlineButton` as intended.
Also, in the "pagesloaded" event handler, we should ensure that there's actually any pages loaded since otherwise the "find current outlineItem"-feature doesn't make any sense.


timvandermeij (approved):

timvandermeij:
Thanks for the follow-up!

PR #13221, opened 2021-04-11
[api-minor] A couple of smaller `PDFPageProxy.getStructTree` fixes (PR 13171 follow-up)

Snuffleupagus:
- Ensure that `getStructTree` won't break with `disableAutoFetch = true` set (PR 13171 follow-up)

   Open http://localhost:8888/web/viewer.html?file=/test/pdfs/pdf.pdf#disableStream=true&disableAutoFetch=true and observe the following message in the console (repeated for each page of the document):
   ```
   Uncaught (in promise)
   Object { message: "Missing data [19787293, 19787294)", name: "UnknownErrorException", details: "MissingDataException: Missing data [19787293, 19787294)", stack: "BaseExceptionClosure@http://localhost:8888/src/shared/util.js:458:29\n@http://localhost:8888/src/shared/util.js:462:3\n" }
   ```

- [api-minor] Let `PDFPageProxy.getStructTree` return `null`, rather than an empty structTree, for documents without any accessibility data (PR 13171 follow-up)

   This is first of all consistent with the existing API, where we return `null` when the data in question doesn't exist. Secondly, it should also be (slightly) more efficient since there's less dummy-data that we need to transfer between threads.
   Finally, this prevents us from adding an empty/unnecessary span to *every* single page even in documents without any structure tree data.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -454,6 +454,13 @@ class Page {
     const structTreeRoot = await this.pdfManager.ensureCatalog(
       "structTreeRoot"
     );
+    return this.pdfManager.ensure(this, "_parseStructTree", [structTreeRoot]);
+  }
+
+  /**
+   * @private
+   */
+  _parseStructTree(structTreeRoot) {
     const tree = new StructTreePage(structTreeRoot, this.pageDict);
     tree.parse();
     return tree;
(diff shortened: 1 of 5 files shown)


Snuffleupagus:
*Add a couple of missing __API__ unit-tests for the structure tree functionality.*

/botio unittest

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #13222, opened 2021-04-11
Convert done callbacks to async/await in the smaller unit test files

timvandermeij:
This commit series is the next step (after PR #13216) in getting rid of the outdated done callbacks by converting to async/await instead. Not using callbacks and having less indentation/code makes the overall code a bit easier to read. I have purposely done the smaller files first to make it easier to review.

### DIFF
--- a/test/unit/ui_utils_spec.js
+++ b/test/unit/ui_utils_spec.js
@@ -178,9 +178,9 @@ describe("ui_utils", function () {
       expect(onceCount).toEqual(1);
     });
 
-    it("should not re-dispatch to DOM", function (done) {
+    it("should not re-dispatch to DOM", async function () {
       if (isNodeJS) {
-        pending("Document in not supported in Node.js.");
+        pending("Document is not supported in Node.js.");
       }
       const eventBus = new EventBus();
       let count = 0;
@@ -189,18 +189,17 @@ describe("ui_utils", function () {
         count++;
       });
       function domEventListener() {
-        done.fail("shall not dispatch DOM event.");
+        // Shouldn't get here.
+        expect(false).toEqual(true);
       }
       document.addEventListener("test", domEventListener);
 
       eventBus.dispatch("test");
 
-      Promise.resolve().then(() => {
-        expect(count).toEqual(1);
+      await Promise.resolve();
+      expect(count).toEqual(1);
 
-        document.removeEventListener("test", domEventListener);
-        done();
-      });
+      document.removeEventListener("test", domEventListener);
     });
   });
 
@@ -265,13 +264,14 @@ describe("ui_utils", function () {
       eventBus = null;
     });
 
-    it("should reject invalid parameters", function (done) {
+    it("should reject invalid parameters", async function () {
       const invalidTarget = waitOnEventOrTimeout({
         target: "window",
         name: "DOMContentLoaded",
       }).then(
         function () {
-          throw new Error("Should reject invalid parameters.");
+          // Shouldn't get here.
+          expect(false).toEqual(true);
         },
         function (reason) {
           expect(reason instanceof Error).toEqual(true);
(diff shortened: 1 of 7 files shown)


test/unit/util_spec.js
```diff
@@ -289,33 +289,31 @@ describe("util", function () {
...
+    it("should reject with correct reason", async function () {
       const promiseCapability = createPromiseCapability();
       expect(promiseCapability.settled).toEqual(false);
 
       promiseCapability.reject(new Error("reason"));
 
-      promiseCapability.promise.then(done.fail, function (reason) {
+      try {
+        await promiseCapability.promise;
+        throw new Error("Promise should be rejected.");
```
Snuffleupagus, line 311: *Looking at this I realize that I've made the same mistake recently, in the API unit-tests, so we probably need to figure out a proper solution here.*

Won't this `throw new Error(...)` actually be handled by the `catch` below, which is decidedly *not* what you actually want here!?
My idea, and it's not perfect, is that we could deal with these types of situations by adding a *custom* Error-type for the unit-tests (placed in test/unit/test_utils.js) and inheriting from `BaseException`.
Maybe name it `TestShouldFailException`, better suggestions are *very* welcome, and instead do something like this (similar in other places):
```javascript
try {
  await promiseCapability.promise;
  throw new TestShouldFailException("Promise should be rejected.");
} catch (reason) {
  if (reason instanceof TestShouldFailException) {
    throw reason;
  }
  // The rest of the catch goes here...
}
```

test/unit/ui_utils_spec.js
```diff
@@ -189,18 +189,16 @@ describe("ui_utils", function () {
         count++;
       });
       function domEventListener() {
-        done.fail("shall not dispatch DOM event.");
+        throw new Error("Shall not dispatch DOM event.");
```
Snuffleupagus, line 192: Have you checked that an `Error` being thrown *inside* of the event handler is enough to propagate it to the "global" level?

Snuffleupagus (changes requested):
Thanks for taking this on!

I've only looked briefly at the first two commits, and found two *possible* bigger picture issues; hence I'm holding off on doing a full review since I guess that these problems might occur elsewhere too.

test/unit/ui_utils_spec.js
timvandermeij: Yes, I tested this by explicitly dispatching the event on the document and seeing that this exception is raised then.

test/unit/util_spec.js
timvandermeij: I'm a bit on the fence here. On one hand this is explicit, on the other hand it's extra code that we'll most likely have to repeat quite a number of times without too much added value since even if it's handled by the current `catch` the test will fail because the message doesn't agree...

Snuffleupagus: > [...] on the other hand it's extra code that we'll most likely have to repeat quite a number of times without too much added value

That's sort of the point here :-)
I see *a lot* of added value in being explicit, since it's then entirely clear what the expected behaviour is and it will *definitely* help prevent future bugs.

> [...] even if it's handled by the current `catch` the test will fail because the message doesn't agree...

While it currently "works", it's however *very* brittle and there's a very real risk that a unit-test could start failing without us noticing.
I think that requiring the `message` matching is an anti-pattern, since we cannot expect that every test will actually want to compare the messages; and in these cases we're *supposed* to fail immediately (as we're currently doing).

timvandermeij: That's fair indeed. Another approach I have been before is this: https://github.com/mozilla/pdf.js/blob/ebeb3f79996df5f6ff411e38e10a572157656f96/test/unit/api_spec.js#L242-L243

It's also not the best, but it would fail immediately without having to add new exception types and handling them.

Snuffleupagus: Oh, that does seem much more light-weight and it conveys the same intent (I'd forgotten about those cases); please use that pattern instead :-)

Snuffleupagus (approved):
This isn't just less code, but it's *much* nicer to look at compared to the callback functions everywhere :-)

Thank you for doing this, and please continue with the rest of the unit-tests when you've got the time!

timvandermeij:
Will do, thanks!

PR #13223, opened 2021-04-12
Remove the unused "GetIsPureXfa" message handler; and avoid unnecessary parsing when no structTree is available (PR 13069 follow-up, PR 13221 follow-up)

Snuffleupagus:
- Remove the unused "GetIsPureXfa" message handler in the worker (PR 13069 follow-up)

   Looking at the API, there's no code which actually sends this message. Most likely it's a left-over from a previous version of PR 13069, since the `isPureXfa` parameter is being included in the "GetDoc" message.

 - Avoid unnecessary parsing, in `Page.GetStructTree`, when no structTree is available (PR 13221 follow-up)

   It's obviously (a bit) more efficient to return early in `Page.getStructTree`, rather than trying to first "parse" an *empty* structTree-root.

   *Somehow I didn't think of this yesterday, but this feels like a much better solution overall; sorry about the churn here!*

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -454,7 +454,13 @@ class Page {
     const structTreeRoot = await this.pdfManager.ensureCatalog(
       "structTreeRoot"
     );
-    return this.pdfManager.ensure(this, "_parseStructTree", [structTreeRoot]);
+    if (!structTreeRoot) {
+      return null;
+    }
+    const structTree = await this.pdfManager.ensure(this, "_parseStructTree", [
+      structTreeRoot,
+    ]);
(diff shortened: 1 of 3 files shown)


Sagar-SalesforceTechnicalArchitech:
Hi Team,
I am implementing PDFJS in salesforce. Using version 2.7.5. and I am passing base64 to it as shown in example of pdfJS. I am getting error "cannot read property document of undefined" which is globalThis.document. I need a urgent help

timvandermeij (approved):

timvandermeij:
Thanks!

Issue #13191, opened 2021-04-07
Cannot switch between annotation layer form fields on Safari (macOS/iOS)


cetinsert:
Attach (recommended) or Link to PDF file here: [OoPdfFormExample.pdf](https://github.com/mozilla/pdf.js/files/6268923/OoPdfFormExample.pdf)

Configuration:
- Web browser and its version: **Safari** &ndash; v14.0.3
- Operating system and its version: **macOS** &ndash; Big Sur v11.2.3

(the issue persists in latest versions of Safari on iOS as well)

- PDF.js version: https://mozilla.github.io/pdf.js/web/viewer.html
- Is a browser extension: no

Steps to reproduce the problem:
1. open attached PDF
2. click/tap into the **Given Name** field
3. click/tap into any other field

What is the expected behavior? (add screenshot)
Focus switches to the field from step 3

What went wrong? (add screenshot)
Focus stays stuck in the Given Name field.
Tab button does not work either.

Snuffleupagus:
WFM, using PDF.js version `2.8.320` in Firefox Nightly.

> Web browser and its version: **Safari** – v14.0.3

If this is specific to Safari, that'd suggest a bug in the browser *itself* rather than the PDF.js library.

---

For anyone trying to debug/fix this, please keep in mind that we'll generally not accept browser-specific compatibility hacks placed in the main code-base and such things should usually live in https://github.com/mozilla/pdf.js/blob/master/src/shared/compatibility.js

cetinsert:
This is specific to Safari - to which there is no alternative on iOS (tablets, phones).

cetinsert:
- Removing contents of the entire `viewer.css` does not help.
- Clicking into developer tools (browser chrome (such as address bar) does not work) and back into an unfocused field does switch focus.

cetinsert:
> Steps to reproduce the problem:
> 1. open attached PDF
> 2. click/tap into the **Given Name** field
> 3. click/tap into any other field

in step 3. other viewer inputs (such as `#pageNumber` and `#findInput`) cannot be focused into either.

cetinsert:
- issue persists in other documents with forms

Snuffleupagus:
Looks, for all intents and purposes, like a duplicate of #12592

Snuffleupagus:
This looks, for all intents and purposes, like a duplicate of #12592.

Issue #12592, opened 2020-11-06
Strange input field behavior in Safari


astehlik:
Attach (recommended) or Link to PDF file here: 
[pdf_form_error_test.pdf](https://github.com/mozilla/pdf.js/files/5500961/pdf_form_error_test.pdf)

Configuration:
- Web browser and its version: Safari versions 13.1 and older
- Operating system and its version: Mac Catalina and iOS (tested with 11 to 14)
- PDF.js version: Current Demo version

Steps to reproduce the problem:
1. Open the attached PDF in Safari (or Chrome on iOS)
2. Try entering a value in the right form field

What is the expected behavior?
When clicking the right field it should be activated.

What went wrong?
The left field ist activated. There is no way to enter a value in the right field.

![safari_bug-2020-11-06_12 09 47](https://user-images.githubusercontent.com/956513/98360675-8494ed80-202a-11eb-9713-d54ac3252cd8.gif)

I did some testing with the Console: when I set a value to the right field using JavaScript I can activate the field and edit the value. But then I can not select the left field any more.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):
https://mozilla.github.io/pdf.js/web/viewer.html?file=https://www.intera.de/fileadmin/pdftest/pdf_form_error_test.pdf

Snuffleupagus:
WFM, when using either: Firefox Nightly 84.0a1 or Google Chrome 87.0.4280.47, on Windows.

Unfortunately this sounds like a platform/browser bug, rather than a PDF.js one, and as such it's really quite difficult to see that it's something that can be fixed here.

astehlik:
@timvandermeij Thanks for the feedback. I'm also thinking it's a browser bug but IMO before I can report a Safari bug I need to find out what exactly is breaking.

I tried reproducing the behavior by building a static HTML page using exactly the same HTML / CSS that is generated by PDF.js (without the JavaScript). The fields behaved normally.

So my conclusion is that is has to be related to the JavaScript of PDF.js. Can you point me to the JS that is related to the annotation overlay so I can dig a little deeper and hopefully find the Code that is breaking?

Cheers
Alex

timvandermeij:
The fields are text widget annotations, for which the display (rendering) code lives here: https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js#L462-L575

cetinsert:
[...]

PR #13224, opened 2021-04-12
Fixes recent Safari TextWidgetAnnotation regression causing #13191 and #12592, ...

CetinSert:
https://github.com/mozilla/pdf.js/commit/44b24fcc2936e1d75c249aa161970ab8fa83e4d2 causes issues on Safari (see #12592 and #13191). Just removing the `event.target.setSelection(0, 0);`  line fixes the issues and no difference / regression in behavior was observed without the `event.target.setSelection(0, 0);` bit in other browsers (including Firefox). The blur (switching focus out of the control) does effectively achieve the same thing in all browsers I have tested on.

Tested specifically in Firefox with

1. inspect a `TextWidgetAnnotation` `input` element
2. `$0.setSelectionRange = (a, b) => console.warn(a, b);`
3. type `wow` and select `wow` in the input element
4. click/tab/tap outside the input element
5. <kbd>CTRL</kbd> + <kbd>C</kbd> ➡ no `wow` ✔ copied
6. <kbd>CTRL</kbd> + <kbd>V</kbd> ➡ confirmed

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -668,7 +668,6 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         if (elementData.formattedValue) {
           event.target.value = elementData.formattedValue;
         }
-        event.target.setSelectionRange(0, 0);
         elementData.beforeInputSelectionRange = null;
       };
 


Snuffleupagus (changes requested):
> [...] and no difference / regression in behavior was observed without the `event.target.setSelection(0, 0);`

Sorry, but this isn't actually true; it does appear that you didn't test the *correct* thing here.

---

You linked to a particular commit, https://github.com/mozilla/pdf.js/commit/44b24fcc2936e1d75c249aa161970ab8fa83e4d2, which points to PR #12360 and that one clearly mentions that this line was necessary to fix issue #12359.

Your patch regresses that issue, and thus we unfortunately cannot accept this patch!

cetinsert:
@Snuffleupagus thank you for the link. I must have been blind. I will take / re-work it from here.

cetinsert:
https://bugzilla.mozilla.org/show_bug.cgi?id=860329 hmm, Firefox 🔥🦊 seems to be alone in this o__O.

cetinsert:
```js
if (event.target.selectionStart > 0)
  event.target.setSelection(0, 0);
```

improves the situation for Safari 🧭. 

Now the second click/tap/tab does get one out: http://jsfiddle.net/mv08ekay/2/

cetinsert:
@calixteman - looking for answers here:
https://stackoverflow.com/questions/67055218/set-input-caret-position-on-blur-in-safari-without-refocusing-input

PR #13227, opened 2021-04-12
can search accent insensitive

Stix89:
be able to do a search without taking into account the accents on the letters


Snuffleupagus (changes requested):
Thanks for submitting this patch; there's a number of things we need to discuss/have answers to before this can be reviewed in any detail:

 - When submitting a patch, please remember to write a *good* commit message outlining/explaining your changes in some detail. Keep in mind that a single sentence isn't really enough...
 - Looking at the complexity of the diacritics matching implemented for regular find operations in Firefox, see e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=202251, how "complete" is the normalization done here?
   Is only *some* things expected to be normalized with this patch, and not others? Please provide some details here (this ties into the point above about a better commit message).
 - How will the normalization perform in longer/more complex documents? E.g. what about a document with 1000+ pages?
 - Will the normalization done here *potentially* split one character into multiple ones? If so, that will most definitely break highlighting of search results (and might not be entirely simple to address). Can we ignore/filter out such cases (if they exists), to keep the initial implementation simpler?

Stix89:
- To search without accent, replace the following characters "Iлｔèｒｎåｔïｏｎɑｌíƶａｔï߀ԉ ą ć ę ł ń ó ś ź ż ä ö ü ß" by "Iлｔeｒｎaｔiｏｎɑｌiƶａｔi߀ԉ a c e l n o s z z a o u ß"

- There is no more normalization, only a remove accent
- The normalization will only remove accent and never split one character into multiple ones

web/pdf_find_controller.js
```diff
@@ -62,6 +62,421 @@ function normalize(text) {
...
+  Â: "A",
+  Ã: "A",
+  Ä: "A",
+  Å: "A",
+  Ấ: "A",
+  Ắ: "A",
+  Ẳ: "A",
+  Ẵ: "A",
+  Ặ: "A",
+  Æ: "AE",
```
Snuffleupagus, line 77: As mentioned previously, "expanding" a character in this way will break highlighting of search results.

Snuffleupagus (changes requested):
(My earlier point, about better a commit/more readable commit message still stands.)

Please keep in mind that this library exists *primarily* to develop the built-in Firefox PDF Viewer, and consequently this kind of feature *must* be implemented in such a way that it fits into the native browser findbar (which has a "Match Diacritics" feature).

Hence it's probably going to be strange, and cause user confusion/bug reports, if this feature does something else in the PDF Viewer compared to searching on an HTML-page.
Note that in my previous comment, I *assumed* (likely incorrectly) that it'd be possible to simply tweak the original normalization-approach, but in its current form this might be a bit too far from working (considering the Firefox-integration for the PDF viewer.) 

Unfortunately this might not be a good beginner bug, all things considered.

PR #13229, opened 2021-04-13
Fix annotation input focus trap regression in Safari

CetinSert:
`setSelectionRange(0, 0)` added in https://github.com/mozilla/pdf.js/commit/44b24fcc2936e1d75c249aa161970ab8fa83e4d2 for #12359, required only by Firefox ([bug](https://bugzilla.mozilla.org/show_bug.cgi?id=860329)), causes issues mozilla#13191, mozilla#12592 in Safari.
`scrollLeft = 0` is a fix that breaks the focus trap in Safari while **keeping Firefox behavior same for #12359**.

@calixteman

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -668,6 +668,7 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         if (elementData.formattedValue) {
           event.target.value = elementData.formattedValue;
         }
+        // Reset the cursor position to the start of the field (issue 12359).
         event.target.scrollLeft = 0;
         elementData.beforeInputSelectionRange = null;
       };
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -667,6 +667,7 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
       let blurListener = event => {
         if (elementData.formattedValue) {
           event.target.value = elementData.formattedValue;
+          event.target.scrollLeft = 0;
         }
         elementData.beforeInputSelectionRange = null;
       };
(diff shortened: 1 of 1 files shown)


cetinsert:
@Snuffleupagus - please see files changed: 1 LoC. Sorry for yesterday, I honestly failed to see the link to #12359.
Minimal case here: http://jsfiddle.net/mv08ekay/4/ (10 LoC - no PDF.js)

Answer from:

- https://stackoverflow.com/questions/67055218/set-input-caret-position-on-blur-in-safari-without-refocusing-input
- https://bugzilla.mozilla.org/show_bug.cgi?id=860329

Safari bug report for `setSelectionRange()`:

- https://bugs.webkit.org/show_bug.cgi?id=224425

src/display/annotation_layer.js
```diff
@@ -668,7 +668,7 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         if (elementData.formattedValue) {
           event.target.value = elementData.formattedValue;
         }
-        event.target.setSelectionRange(0, 0);
+        event.target.scrollLeft = 0;
```
Snuffleupagus, line 671: This *really* needs a good comment, referencing the original issue, since it's less clear exactly what this does than in the previous code; I'd suggest: 
```javascsript
        // Reset the cursor position to the start of the field (issue 12359).
        event.target.scrollLeft = 0;
```

When fixing this, please remember https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

Snuffleupagus (changes requested):

Snuffleupagus (changes requested):
As mentioned previously, you *need* to follow https://github.com/mozilla/pdf.js/wiki/Squashing-Commits since we cannot consider landing this in its current state.

Note also that it's not advisable to do development directly against the `master` branch, please see https://github.com/mozilla/pdf.js/wiki/Contributing

cetinsert:
@Snuffleupagus great, so this will now take me at least a few more days to my workflow up properly.

cetinsert:
@Snuffleupagus see https://github.com/mozilla/pdf.js/pull/13232. Thank you for your prompt and patient guidance!

PR #13235, opened 2021-04-13
Split the classes in `src/core/obj.js` into separate files

Snuffleupagus:
The size of the `src/core/obj.js` file has increased slowly over the years, and it also contains a fair amount of *distinct* functionality.
In order to improve readability and make it easier to navigate through the code, these patches move (and re-factor) the different classes into their own files.

*In order to simplify reviewing:* I'd suggest looking at one commit at a time, and using the `?w=1` flag to reduce the size of the diffs.


timvandermeij:
I think this is a good idea now that the size and functionality of `obj.js` has increased. I'll have a more detailed look once this is rebased.

timvandermeij (approved):
r=me if the tests pass. Looks much better like this; thank you for doing this!

timvandermeij:
Nice work!

PR #13240, opened 2021-04-14
Update `eslint-config-prettier`/`eslint-plugin-unicorn` packages to their latest versions

Snuffleupagus:
Both of these ESLint "helpers" has increased their *major* version numbers, which `npm update` doesn't handle automatically (since that could cause errors), hence this patch which updates them manually.

https://www.npmjs.com/package/eslint-config-prettier
https://www.npmjs.com/package/eslint-plugin-unicorn


timvandermeij (approved):

timvandermeij:
Thank you for updating this!

Snuffleupagus:
There's now only two packages left where we're on old *major* versions:
 - SystemJS, which we don't really care about since the intention is to get rid of it completely as soon as https://bugzilla.mozilla.org/show_bug.cgi?id=1247687 is fixed.
 - yargs, which we however *should* updated. I've looked at that briefly, at least twice, and wasn't really able to immediately figure out exactly what changes were necessary.

PR #13243, opened 2021-04-14
Implement visibility expressions for optional content

### LINKED ISSUE #12097: Support optional content group visibility expressions
See `8.11.2.2 Optional Content Membership Dictionaries` -> `VE` entry. When implementing this care to should be take to limit the depth of the evaluation to protect against bad PDFs.

janpe2:
Fixes #12097


src/core/evaluator.js
```diff
@@ -1324,6 +1324,56 @@ class PartialEvaluator {
         id: optionalContent.objId,
       };
     } else if (optionalContentType === "OCMD") {
+      const expression = optionalContent.get("VE");
+      if (Array.isArray(expression)) {
+        const MAX_NESTING = 10;
+
+        const parseExpression = (array, nestingCounter, currentResult) => {
```
Snuffleupagus, line 1331: Can we please move this helper function to the top-level of the file, rather than defining it inline?

src/display/optional_content_config.js
```diff
@@ -65,10 +65,46 @@ class OptionalContentConfig {
       }
       return this._groups.get(group.id).visible;
     } else if (group.type === "OCMD") {
-      // Per the spec, the expression should be preferred if available. Until
-      // we implement this, just fallback to using the group policy for now.
+      // Per the spec, the expression should be preferred if available.
       if (group.expression) {
-        warn("Visibility expression not supported yet.");
+        const evaluateExpression = array => {
```
Snuffleupagus, line 70: Can we please move this helper function to a "private" method, rather than defining it inline?

Snuffleupagus (changes requested):
Thanks for taking this on!

I've not really reviewed the patch yet, but I've got two smaller suggestions to (maybe) improve overall readability.

src/core/evaluator.js
```diff
@@ -1324,6 +1365,16 @@ class PartialEvaluator {
         id: optionalContent.objId,
       };
     } else if (optionalContentType === "OCMD") {
+      const expression = optionalContent.get("VE");
+      if (Array.isArray(expression)) {
+        const result = [];
+        this._parseVisibilityExpression(expression, 0, result);
```
Snuffleupagus, line 1371: Is there ever a risk, in a badly generated PDF document, that the `result` array is *empty* here?
If that were to happen: is it correct to use the empty array here, or should we fallback to the "regular" parsing below?

src/core/evaluator.js
```diff
@@ -1306,6 +1306,47 @@ class PartialEvaluator {
     throw new FormatError(`Unknown PatternName: ${patternName}`);
   }
 
+  _parseVisibilityExpression(array, nestingCounter, currentResult) {
+    const MAX_NESTING = 10;
```
Snuffleupagus, line 1310: I think it'd be *generally* safer to the do nesting check here, since that guarantees that it cannot be missed if any future changes are made to this method; hence something like this (which also removes the need to manually increment the counter in the call below):
```javascript
    if (++nestingCounter > MAX_NESTING) {
      warn("Visibility expression is too deeply nested");
      return;
    }
```

Snuffleupagus (changes requested):

src/core/evaluator.js
janpe2: It's probably safer to return the result only if its length > 0

src/core/evaluator.js
janpe2: That's a good idea

Snuffleupagus (approved):
r=me, with passing tests; thank you for implementing this!

timvandermeij (approved):

timvandermeij:
Nice clean implementation; thanks!

/botio makeref

PR #13244, opened 2021-04-14
Update the `postcss-calc` package to the latest version

Snuffleupagus:
I've successfully run `gulp mozcentral` and `gulp generic` locally, with/without this patch and diff-ed the *built* `web/viewer.css` files. There were no changes, in the build-output, caused by this update.

https://www.npmjs.com/package/postcss-calc

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14703,58 +14703,13 @@
       }
     },
     "postcss-calc": {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13246, opened 2021-04-14
Convert done callbacks to async/await in more unit test files

timvandermeij:
This is the follow-up PR after #13222 that finalizes the conversion for the smaller and medium sized unit test files. Only three unit test files remain with done callbacks after this PR, but since they are bigger they will be done in a follow-up PR.

_Using the `?w=1` parameter will help during review for most commits._


Snuffleupagus (approved):
Looks good to me, thank you!

PR #13247, opened 2021-04-15
Update the `yargs` package to the latest version

Snuffleupagus:
While I wasn't able to figure out *exactly* why the old format didn't work, re-factoring the `parseOptions` function to use `yargs` differently "just worked" so that's hopefully good enough here.
With these changes everything related to a *particular* option now appears in one place, rather than being spread out, which aids readability in my opinion. Also, the options are now sorted alphabetically, to make it easier to find a particular one.

https://www.npmjs.com/package/yargs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9670,15 +9670,6 @@
         "kind-of": "^6.0.2"
       }
     },
-    "map-age-cleaner": {
-      "version": "0.1.3",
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks much nicer. Thank you for bringing this up-to-date!

PR #13249, opened 2021-04-16
[GENERIC viewer] Don't display the `errorWrapper`, for documents with signatures (PR 13217 follow-up)

Snuffleupagus:
According to a decision by UX and PM, please see https://bugzilla.mozilla.org/show_bug.cgi?id=1705060#c2, in Firefox we should first of all *not* display the notification-bar for signatures. Secondly, as can also be seen there, we shouldn't display the notification-bar *at all* and it's thus disabled in https://bugzilla.mozilla.org/show_bug.cgi?id=1705327.

If we purposely don't display a notification, for documents with signatures, in the *built in* Firefox PDF Viewer then it cannot be necessary in the GENERIC viewer either.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13253, opened 2021-04-16
Convert done callbacks to async/await in `test/unit/api_spec.js`

timvandermeij:
This is one of the three remaining files, and since it's quite big it's done separately from the other ones for easier review. It's recommended to use the `?w=1` flag to make reviewing much easier because of all indentation changes.

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -101,7 +101,7 @@ describe("api", function () {
       await loadingTask.destroy();
     });
 
-    it("creates pdf doc from URL", function (done) {
+    it("creates pdf doc from URL", async function () {
       const loadingTask = getDocument(basicApiGetDocumentParams);
 
       const progressReportedCapability = createPromiseCapability();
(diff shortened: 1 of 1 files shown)


test/unit/api_spec.js
```diff
@@ -113,107 +113,110 @@ describe("api", function () {
...
+    it("creates pdf doc from typed array", async function () {
       const typedArrayPdfPromise = DefaultFileReaderFactory.fetch({
         path: TEST_PDFS_PATH + basicApiFileName,
       });
 
-      typedArrayPdfPromise
-        .then(typedArrayPdf => {
-          // Sanity check to make sure that we fetched the entire PDF file.
-          expect(typedArrayPdf.length).toEqual(basicApiFileLength);
+      const typedArrayPdf = await typedArrayPdfPromise;
```
Snuffleupagus, line 160: Nit: You could probably simplify this a bit further, if you wanted to.
```javascript
const typedArrayPdf = await DefaultFileReaderFactory.fetch({
  path: TEST_PDFS_PATH + basicApiFileName,
});
```

test/unit/api_spec.js
```diff
@@ -798,135 +742,103 @@ describe("api", function () {
...
 
-      loadingTask.promise
-        .then(function (pdfDoc) {
-          return pdfDoc.getPageLayout();
-        })
-        .then(function (mode) {
-          expect(mode).toEqual("");
+      const pageLayout = await loadingTask.promise.then(function (pdfDoc) {
+        return pdfDoc.getPageLayout();
+      });
```
Snuffleupagus, line 771: Nit: Here, and in *a lot* of similar cases below (where there's only one promise involved), you could just do e.g. the following instead.
```javascript
const pdfDoc = await loadingTask.promise;
const pageLayout = await pdfDoc.getPageLayout();
```

test/unit/api_spec.js
```diff
@@ -2260,132 +2073,110 @@ describe("api", function () {
...
+      expect(pdfPage.rotate).toEqual(0);
+      expect(fetches).toEqual(1);
 
-          waitSome(function () {
-            loadingTask.destroy().then(done);
-          });
-        })
-        .catch(done.fail);
+      waitSome(async function () {
+        await loadingTask.destroy();
```
Snuffleupagus, line 2148: Are you *absolutely sure* that this works as intended, since it looks like the test will now finish *before* destruction is complete. I'd expect that what you actually want is something like:
```javascript
await new Promise(resolve => {
  waitSome(resolve);
});
await loadingTask.destroy();
```

Snuffleupagus (changes requested):
That's a lot of code, but hopefully I've not missed anything :-)

This looks mostly good, but I think that there's some additional clean-up possible if you're touching most of the file anyway.
I'd suggest re-factoring the promise usage a bit more while you're at it, please see the inline comment which shows *one* such example.

Also, can you please fix the mistake I made in https://github.com/mozilla/pdf.js/blob/master/test/unit/api_spec.js#L2116?

timvandermeij:
The review comments have been addressed. Given the size of the file, it might be easiest to review the interdiff and ignoring all changes not in `test/unit/api_spec` (from the rebase onto master); please see https://github.com/mozilla/pdf.js/compare/9f8272faf75bca4d2c6d88ed6c8cfaf85fe24146..130c2547ec8ecbdd85988d6720eee89441d362cd.

test/unit/api_spec.js
timvandermeij: Good point; fixed now.

test/unit/api_spec.js
timvandermeij: Good idea! This increases readability and saves another 30-ish lines in the file.

timvandermeij:
I have also fixed the mistake you mentioned now.

/botio unittest

test/unit/api_spec.js
```diff
@@ -1053,112 +937,89 @@ describe("api", function () {
...
-            PageClose: [`this.getField("Text2").value = "PageClose 1";`],
-          });
-          expect(page3Actions).toEqual({
-            PageOpen: [`this.getField("Text5").value = "PageOpen 3";`],
-            PageClose: [`this.getField("Text6").value = "PageClose 3";`],
-          });
-          loadingTask.destroy().then(done);
-        })
-        .catch(done.fail);
+      const [docActions, page1Actions, page3Actions] = await promise;
```
Snuffleupagus, line 940: Nit: In *this* particular case, given how the pre-existing code is written above, I'm not sure that we actually need the temporary `promise` for anything?

*Normally* we usually look-up the various resources in *parallel*, but in this case everything is done serially and the following should thus be equivalent:
```javascript
const pdfDoc = await loadingTask.promise;
const docActions = await pdfDoc.getJSActions();
const page1 = await pdfDoc.getPage(1);
const page1Actions = await page1.getJSActions();
const page3 = await pdfDoc.getPage(3);
const page3Actions = await page3.getJSActions();
```

Snuffleupagus (approved):
That's a lot of boilerplate being removed here :-)

r=me, with one final nit; thank you for doing this clean-up!

PR #13254, opened 2021-04-17
A couple of small scripting/XFA-related tweaks in the worker-code

Snuffleupagus:
- Use `PDFManager.ensureDoc`, rather than `PDFManager.ensure`, in a couple of spots in the code. If there exists a short-hand format, we should obviously use it wherever possible.

 - Fix a unit-test helper, to account for the previous changes. (Also, converts a function to be `async` instead.)

 - Add one more exists-check in `PDFDocument.loadXfaFonts`, which I missed to suggest in PR 13146, to prevent any future errors if the method is ever called in a situation where it shouldn't be.
  Also, print a warning if the actual font-loading fails since that could help future debugging. (Finally, reduce overall indentation in the loop.)

 - Slightly unrelated, but make a small tweak of a comment in `src/core/fonts.js` to reduce possible confusion.

*Slightly smaller/simpler diff with https://github.com/mozilla/pdf.js/pull/13254/files?w=1*

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -865,7 +865,9 @@ class PDFDocument {
 
   async loadXfaFonts(handler, task) {
     const acroForm = await this.pdfManager.ensureCatalog("acroForm");
-
+    if (!acroForm) {
+      return;
+    }
     const resources = await acroForm.getAsync("DR");
     if (!(resources instanceof Dict)) {
       return;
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13256, opened 2021-04-17
Convert done callbacks to async/await in the last two unit test files

timvandermeij:
It's highly recommended to use the `?w=1` flag during review because of many indentation changes.

### DIFF
--- a/test/unit/evaluator_spec.js
+++ b/test/unit/evaluator_spec.js
@@ -37,24 +37,16 @@ describe("evaluator", function () {
     },
   };
 
-  function runOperatorListCheck(evaluator, stream, resources, callback) {
-    const result = new OperatorList();
+  async function runOperatorListCheck(evaluator, stream, resources) {
+    const operatorList = new OperatorList();
     const task = new WorkerTask("OperatorListCheck");
-    evaluator
-      .getOperatorList({
-        stream,
-        task,
-        resources,
-        operatorList: result,
-      })
-      .then(
-        function () {
-          callback(result);
-        },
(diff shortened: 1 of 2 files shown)


test/unit/evaluator_spec.js
```diff
@@ -377,54 +341,58 @@ describe("evaluator", function () {
...
+        // Shouldn't get here.
+        expect(false).toEqual(true);
+      } catch (_) {
+        expect(!!result.fnArray && !!result.argsArray).toEqual(true);
+        expect(result.fnArray.length).toEqual(0);
+      }
     });
-    it("should abort text parsing parsing", function (done) {
+
+    it("should abort text parsing parsing", async function () {
```
Snuffleupagus, line 378: Nit: Pre-existing issue, but there's probably one "parsing" too much in the test name :-)
I'm guessing that it *should* actually be:
```suggestion
    it("should abort text content parsing", async function () {
```

Snuffleupagus (approved):
Well, that certainly was *challenging* to try and review; I can only image that actually making the changes was even worse :-)

Looks great to me, and using `async`/`await` really is a much nicer format; thank you for modernizing the unit-tests!

timvandermeij:
Thank you for the reviews!

Snuffleupagus:
@timvandermeij Do you want to fix the font-tests as well, or should I do that?

timvandermeij:
If you've already started on it, feel free to finish it. I won't have too much time the coming days, so otherwise it'll have to wait a couple of days.

PR #13257, opened 2021-04-17
[api-minor] Fix the way to chunk the strings

calixteman:
- Improve chunking in order to fix some bugs where the spaces aren't here:
    * track the last position where a glyph has been drawn;
    * when a new glyph (first glyph in a chunk) is added then compare its position with the last saved one and add a space or break:
      - there are multiple ways to move the glyphs and to avoid to have to deal with all the different possibilities it's a way easier to just compare positions;
      - and so there is now one function (i.e. "compareWithLastPosition") where all the job is done.
  - Add some breaks in order to get lines;
  - Remove the multiple whites spaces:
    * some spaces were filled with several whites spaces and so it makes harder to find some sequences of words using the search tool;
    * other pdf readers replace spaces by one white space.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -2042,7 +2042,7 @@ class PartialEvaluator {
     resources = resources || Dict.empty;
     stateManager = stateManager || new StateManager(new TextState());
 
-    var WhitespaceRegexp = /\s/g;
+    const WhitespaceRegexp = /\s/g;
 
     var textContent = {
       items: [],
@@ -2051,26 +2051,43 @@ class PartialEvaluator {
     var textContentItem = {
       initialized: false,
       str: [],
+      totalWidth: 0,
+      totalHeight: 0,
       width: 0,
       height: 0,
       vertical: false,
-      lastAdvanceWidth: 0,
-      lastAdvanceHeight: 0,
+      lastCharSize: 0,
+      prevTransform: null,
       textAdvanceScale: 0,
       spaceWidth: 0,
-      fakeSpaceMin: Infinity,
-      fakeMultiSpaceMin: Infinity,
-      fakeMultiSpaceMax: -0,
-      textRunBreakAllowed: false,
+      spaceInFlowMin: 0,
+      spaceInFlowMax: 0,
+      trackingSpaceMin: Infinity,
       transform: null,
       fontName: null,
+      hasEOL: false,
+      isLastCharWhiteSpace: false,
     };
-    var SPACE_FACTOR = 0.3;
-    var MULTI_SPACE_FACTOR = 1.5;
-    var MULTI_SPACE_FACTOR_MAX = 4;
+
+    // Used in addFakeSpaces.
+    // wsw stands for whitespace width.
+
+    // A white <= wsw * TRACKING_SPACE_FACTOR is a tracking space
+    // so it doesn't count as a space.
+    const TRACKING_SPACE_FACTOR = 0.3;
+
+    // A white with a width in [wsw * MIN_FACTOR; wsw * MAX_FACTOR]
+    // is a space which will be inserted in the current flow of words.
+    // If the width is outside of this range then the flow is broken
+    // (which means a new span in the text layer).
+    // It's useful to adjust the best as possible the span in the layer
+    // to what is displayed in the canvas.
+    const SPACE_IN_FLOW_MIN_FACTOR = 0.3;
+    const SPACE_IN_FLOW_MAX_FACTOR = 1.3;
 
     var self = this;
     var xref = this.xref;
+    const showSpacedTextBuffer = [];
 
     // The xobj is parsed iff it's needed, e.g. if there is a `DO` cmd.
     var xobjs = null;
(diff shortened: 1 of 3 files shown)


src/core/evaluator.js
```diff
@@ -2120,18 +2123,41 @@ class PartialEvaluator {
...
+      if (!(font.loadedName in seenStyles)) {
+        seenStyles[font.loadedName] = true;
+        textContent.styles[font.loadedName] = {
+          fontFamily: font.fallbackName,
+          ascent: font.ascent,
+          descent: font.descent,
+          vertical: font.vertical,
+        };
+      }
+      textContentItem.fontName = font.loadedName;
```
Snuffleupagus, line 2146: This looks like a bug introduced *accidentally* during a re-basing, since the upstream code changed in the meantime and this won't actually work as intended now; hence you want the following here:
```suggestion
      const font = textState.font,
        loadedName = font.loadedName;
      if (!seenStyles.has(loadedName)) {
        seenStyles.add(loadedName);

        textContent.styles[loadedName] = {
          fontFamily: font.fallbackName,
          ascent: font.ascent,
          descent: font.descent,
          vertical: font.vertical,
        };
      }
      textContentItem.fontName = loadedName;
      
```

src/core/evaluator.js
```diff
@@ -2178,17 +2229,20 @@ class PartialEvaluator {
...
-          : bidiResult.str,
+        str,
         dir: bidiResult.dir,
-        width: textChunk.width,
-        height: textChunk.height,
+        width: textChunk.totalWidth,
+        height: textChunk.totalHeight,
         transform: textChunk.transform,
         fontName: textChunk.fontName,
+        isSpace: AllWhitespaceRegexp.test(str),
```
Snuffleupagus, line 2244: Unless I'm missing something obvious here, in the latest version of the patch it appears that this property is *only* being using (on the main-thread) for the `task._enhanceTextSelection` mode?

If so, it feels a bit unnecessary to run a regular expression on *all* text (for a mode not used by default), and my question is if we cannot instead run the regex-check in the `src/display/text_layer.js` file only when `task._enhanceTextSelection` is actually set?

src/core/evaluator.js
```diff
@@ -2202,91 +2256,257 @@ class PartialEvaluator {
...
+      } else {
+        appendEOL();
+      }
+    }
 
-        var glyphUnicode = glyph.unicode;
-        var NormalizedUnicodes = getNormalizedUnicodes();
-        if (NormalizedUnicodes[glyphUnicode] !== undefined) {
-          glyphUnicode = NormalizedUnicodes[glyphUnicode];
+    function buildTextContentItem(chars, extraSpacing, isFirstChunk) {
```
Snuffleupagus, line 2339: Nit: You may want to consider changing this function to take a parameter object instead, which improves the self-documention at the call-sites, since some of the calls below are a little hard read as-is.
(That way you could also make use of default values, removing the need to pass all parameters everywhere.)

One example is the  `buildTextContentItem(args[0], 0, true);` calls, where it's not really obvious what the arguments represent without going back to the function definition.

Snuffleupagus (changes requested):

src/core/evaluator.js
```diff
@@ -2399,40 +2587,10 @@ class PartialEvaluator {
             textState.textMatrix = textState.textLineMatrix.slice();
             break;
           case OPS.nextLine:
-            flushTextContentItem();
+            appendEOL();
             textState.carriageReturn();
             break;
           case OPS.setTextMatrix:
-            // Optimization to treat same line movement as advance.
-            advance = textState.calcTextLineMatrixAdvance(
```
Snuffleupagus, line 2407: If the patch results in this method becoming unused, and you don't plan on using it for any (possibly) upcoming changes, please remove this code as well: https://github.com/mozilla/pdf.js/blob/c86e70ba08bcde6b2b510bbfd60dc5567ac5285b/src/core/evaluator.js#L3890-L3915

---

Also, it appears that this method has been unused for *years* and can also be removed: https://github.com/mozilla/pdf.js/blob/c86e70ba08bcde6b2b510bbfd60dc5567ac5285b/src/core/evaluator.js#L3917-L3928

Snuffleupagus (commented):
Looking at the `bug1130815-text` results, it appears that there's now *multiple* spaces being added in the footers on the pages. (Obviously that particular PDF document is *corrupt*, but I figured it cannot hurt to mention this.)

calixteman:
> Looking at the `bug1130815-text` results, it appears that there's now _multiple_ spaces being added in the footers on the pages. (Obviously that particular PDF document is _corrupt_, but I figured it cannot hurt to mention this.)

`Fiche technique...` is followed by a white space character and then there is an empty space before `Mise a jour` so the empty space is replaced by a white space thanks to this patch.
When the last char in the chunk is a white space then we could not add a new one for an empty space, wdyt ?

Snuffleupagus:
> When the last char in the chunk is a white space then we could not add a new one for an empty space, wdyt ?

That sounds like a nice idea, assuming it's reasonably straightforward to implement, since this situation might not be too uncommon in practice.

src/display/text_layer.js
```diff
@@ -142,12 +138,8 @@ const renderTextLayer = (function renderTextLayerClosure() {
       scale: 1,
     };
 
+    textDiv.textContent = geom.str;
```
Snuffleupagus, line 141: This looks like a mistake, since there's already an *identical* line below; please remove this.

src/display/text_layer.js
```diff
@@ -177,6 +169,7 @@ const renderTextLayer = (function renderTextLayerClosure() {
     textDiv.setAttribute("role", "presentation");
 
     textDiv.textContent = geom.str;
+
```
Snuffleupagus, line 172: Nit: Unrelated change.

Snuffleupagus (changes requested):

brendandahl (approved):
Working nicely. One minor change needed. 

It doesn't have to be done in this PR, but it would be good to add some integration tests that select the text of a page and compare it to what we expect so we don't regress this in the future.

Snuffleupagus:
Huh, I'm guessing that `makeref` failed because of *how* this PR was merged; I'll fix that separately.

Please note that the default, in the PDF.js project, has always been to use *merge commits* (rather than landing it as-is), in which case `makeref` should just work.

timvandermeij:
Nice work! It closed a lot of issues already, and there might even be more.

Issue #10036, opened 2018-09-04
Support regex search


tk3369:
Hi, great project!  
Currently, pdf.js seems to support exact matches only.  Can it be enhanced with regular expression search function?

yurydelendik:
IMHO is it too complex for end user and hardly adds any value to the demo viewer

timvandermeij:
You could update the find controller (https://github.com/mozilla/pdf.js/blob/master/web/pdf_find_controller.js) in a custom build to support this, which is most likely not too difficult because you have the query handling and match calculation logic there, but we don't plan on adding such functionality in the default viewer since most users won't use this.

lixaotec:
That would be a great feature, not for end user, but as api, developers need. I also look foward this behavior.

PR #13261, opened 2021-04-18
[api-minor] Support search with or without diacritics  (bug 1508345, bug 916883, bug 1651113)

### LINKED ISSUE #4742: Can't search across lines in some PDFs (like the demo PDF)
pdf.js is not able to search for text across lines in some PDFs (like the demo PDF). For example if you search for the phrase "code that can handle" the find dialog states "Phrase not found"; however, this phrase does exist in the 1st paragraph (3rd line) wrapping to the next line.

### LINKED ISSUE #7355: String.indexOf() cannot match phrases with variable whitespace
`String.indexOf()` is being used for phrase search. There are two scenarios where this will fail because of whitespace:
1. Justified text, very common in legal documents, which can result in variable amount of whitespace between words.
2. Currently spaces are not added between lines therefore phrases that span multiple lines will not be found. See #5783 for more information.

Even if 2 is fixed, we still have to deal with 1 for phrase search. I have written a JS function that is equivalent to `String.indexOf()` but ignores `" "` while matching. This function also returns the end index since the length of `searchValue` cannot be trusted because the amount of whitespace between `text` and `searchValue` may differ. It could be modified to handle all whitespace characters, but I think it is unnecessary in this case since the text layer converts all whitespace to `" "`.

``` js
function indexOfIgnoreSpace(text, searchValue, fromIndex) {

  var begin = 0;
  var end = -1;
  var fragment = searchValue.slice(0).split(' ').join(''); // remove whitespace

  if (!fromIndex) {
    fromIndex = 0;
  }

  for (var i = fromIndex; i < text.length; i++) {

    // do not start matching on whitespace
    if (text[i] === ' ') {
      continue;
    }

    var index = 0;
    begin = i;

    for (var j = i; j < text.length; j++) {

      if (text[j] === ' ') {
          continue;
      }

      if (text[j] !== fragment[index]) {
          break;
      }

      index++;
      if (index === fragment.length) {
          end = j + 1;
          break;
      }
    }

    if (end !== -1) {
      return {
        begin: begin,
        end: end
      };
    }
  }

  return null;
}
```

### LINKED ISSUE #11752: Can't find words hyphenated across lines
**Attach PDF file here:** [1503.01817v2.pdf](https://github.com/mozilla/pdf.js/files/4387305/1503.01817v2.pdf) (from https://arxiv.org/pdf/1503.01817v2.pdf)

**Configuration:**
- Web browser version: Chrome 80.0.3987.158
- Operating system version: Chrome OS 80.0.3987.158
- PDF.js version: 2.5.40 https://mozilla.github.io/pdf.js/web/viewer.html
- Is a browser extension: no (but also reproduces on https://chrome.google.com/webstore/detail/pdf-viewer/oemmndcbldboiebfnladdacbdfmadadm version 2.3.164)

1. Open the attached PDF.
2. Search for "caption".

**What is the expected behavior?**

There should be 2 matches, one on the line "**caption**s extracted from the videos", and another straddling the two lines "boxes, segmentations of objects and faces, and image **cap-**" and "**tion**s are not yet available for the YFCC100M at present,". This is what you get in Chrome's built-in PDF viewer.

**What went wrong?**

Only the first match is found, as you can see in this screenshot:
>  screenshot

**Analysis**

This seems to happen because pdf.js doesn't join lines as well as Chrome's PDF viewer. If you copy the two lines using pdf.js, you get the following string, with an extraneous dash between "cap" and "tions":
```
boxes, segmentations of objects and faces, and image cap-tions  are  not  yet  available  for  the  YFCC100M  at  present,
```

whereas if you copy the same text using Chrome's PDF viewer, you get the following string without a hyphen:
```
boxes, segmentations of objects and faces, and image captions are not yet available for the YFCC100M at present,
```

(If you search for "cap-tion" with a hyphen in pdf.js, it does match, but that's not particularly useful.)

**Possible solutions**

The cleanest solution would be to remove line break hyphens in the middle of words when joining lines together, like Chrome's pdfium. This would fix both copy/paste and Find in Document.

A hackier alternative would be to just make Find in Document ignore hyphens (so "caption" would match "cap-tion"), but then there would still be extraneous hyphens when you copy text that was broken across lines midword.

**Additional information**

Related to #4742 (Can't search across lines in some PDFs) and its PR #5783 (Improve Copy/Paste).

The PDF was generated with standard software for scientific papers: [1.5 pdfTeX-1.40.12 / LaTeX with hyperref package]

calixteman:
- get original index in using a dichotomic seach instead of a linear one;
  - normalize the text in using NFD;
  - convert the query string into a RegExp;
  - replace whitespaces in the query with \s+;
  - handle hyphens at eol use to break a word;
  - add some \s* around punctuation signs


calixteman:
The PR is almost ready but highlights are wrong with RTL languages.

test/unit/pdf_find_controller_spec.js
```diff
@@ -271,5 +271,90 @@ describe("pdf_find_controller", function () {
       pageMatches: [[19, 48, 66]],
       pageMatchesLength: [[8, 8, 8]],
     });
+
+    await testSearch({
+      eventBus,
+      pdfFindController,
+      parameters: {
+        query: "1/2",
```
Snuffleupagus, line 279: Would it perhaps not be more "interesting" to search for `½` instead here?

web/viewer.html
```diff
@@ -137,8 +137,13 @@
             <input type="checkbox" id="findMatchCase" class="toolbarField" tabindex="95">
             <label for="findMatchCase" class="toolbarLabel" data-l10n-id="find_match_case_label">Match case</label>
           </div>
+          <div id="findbarOptionsThreeContainer">
```
Snuffleupagus, line 140: It looks like you flipped the names of the containers here, since two comes before three :-)

More generally, you now have a container with only *one* input element which doesn't really make all that much sense. I'd expect the following format instead:
```javascript
          <div id="findbarOptionsTwoContainer">
            <input type="checkbox" id="findMatchDiacritics" class="toolbarField" tabindex="96">
            <label for="findMatchDiacritics" class="toolbarLabel" data-l10n-id="find_match_diacritics_label">Match Diacritics</label>
            <input type="checkbox" id="findEntireWord" class="toolbarField" tabindex="97">
            <label for="findEntireWord" class="toolbarLabel" data-l10n-id="find_entire_word_label">Whole words</label>
          </div>
          <div id="findbarOptionsThreeContainer">
            <span id="findResultsCount" class="toolbarLabel hidden"></span>
          </div>
```

web/viewer.html
```diff
@@ -137,8 +137,13 @@
...
+            <input type="checkbox" id="findMatchDiacritics" class="toolbarField" tabindex="96">
+            <label for="findMatchDiacritics" class="toolbarLabel" data-l10n-id="find_match_diacritics_label">Match diacritics</label>
```
Snuffleupagus, line 142: `Match diacritics` -> `Match Diacritics` such that the capitalization matches the Firefox findbar.

Also, please remember to add `find_match_diacritics_label=Match Diacritics` *after* this line as well (such that it can actually be localized): https://github.com/mozilla/pdf.js/blob/fd82adccfa409f9b2cc72ad2cb8555bd2aa68b36/l10n/en-US/viewer.properties#L170

web/viewer.js
```diff
@@ -155,6 +155,7 @@ function getViewerConfiguration() {
       highlightAllCheckbox: document.getElementById("findHighlightAll"),
       caseSensitiveCheckbox: document.getElementById("findMatchCase"),
       entireWordCheckbox: document.getElementById("findEntireWord"),
+      diacriticsCheckbox: document.getElementById("findMatchDiacritics"),
```
Snuffleupagus, line 158: Nit: `diacriticsCheckbox` -> `matchDiacriticsCheckbox` for consistency.

web/pdf_find_controller.js
```diff
@@ -43,46 +43,86 @@ const CHARACTERS_TO_NORMALIZE = {
 };
 
 let normalizationRegex = null;
-function normalize(text) {
+function normalize(text, matchDiacritics) {
```
Snuffleupagus, line 46: Perhaps add a default value, i.e.
```javascript
function normalize(text, matchDiacritics = false) {
```

web/pdf_find_controller.js
```diff
@@ -43,46 +43,86 @@ const CHARACTERS_TO_NORMALIZE = {
...
+    let regexStr = `[${replace}]`;
+    let option = "g";
+    if (!matchDiacritics) {
+      regexStr += "|\\p{Mn}+";
+      option += "u";
+    }
+    normalizationRegex = new RegExp(regexStr, option);
+  }
+
+  // The goal of this fonction is to normalize the string and
```
Snuffleupagus, line 59: Spelling-nit: `fonction` -> `function`.

web/pdf_find_controller.js
```diff
@@ -43,46 +43,86 @@ const CHARACTERS_TO_NORMALIZE = {
...
+  // substitution at index y (old string) which will replace o chars by n chars.
+  // Firstly, if o === n, then no need to add a new entry: the shift is
+  // the same.
+  // Secondly, if o < n, then we push the n - o elements:
+  // [y - (s - 1), s - 1], [y - (s - 2), s - 2], ...
+  // Thirdly, if o > n, then we push the element: [y - (s - n), o + s - n]
+  const positions = [[0, 0]];
+  let shift = 0;
+  const normalized = text.replace(normalizationRegex, (match, i) => {
+    const rep = CHARACTERS_TO_NORMALIZE[match];
```
Snuffleupagus, line 88: Nit: `rep` -> `replacement`?

web/pdf_find_controller.js
```diff
@@ -381,11 +425,12 @@ class PDFFindController {
     return true;
   }
 
-  _calculatePhraseMatch(query, pageIndex, pageContent, pageDiffs, entireWord) {
-    const matches = [],
-      matchesLength = [];
+  _calculatePhraseMatch(query, pageIndex, pageContent, entireWord) {
+    const matches = [];
+    const matchesLength = [];
```
Snuffleupagus, line 430: Looks like an unrelated change...

web/pdf_find_controller.js
```diff
@@ -381,11 +425,12 @@ class PDFFindController {
...
     const queryLen = query.length;
-
+    const diffs = this._pageDiffs[pageIndex];
+    const hasDiffs = !!diffs?.length;
```
Snuffleupagus, line 433: Nit: can we keep a blank line here, for readability?

web/pdf_find_controller.js
```diff
@@ -457,36 +514,33 @@ class PDFFindController {
...
+    } = this._state;
 
     if (query.length === 0) {
       // Do nothing: the matches should be wiped out already.
       return;
     }
 
+    if (!matchDiacritics) {
+      query = stripOutDiacritics(query);
+      pageContent = stripOutDiacritics(pageContent, true);
```
Snuffleupagus, line 532: Huh, but this function only accepts *one* parameter!?

web/pdf_find_controller.js
```diff
@@ -232,6 +276,7 @@ class PDFFindController {
     this._pdfDocument = null;
     this._pageMatches = [];
     this._pageMatchesLength = [];
+    this._pageDiffs = [];
```
Snuffleupagus, line 279: Why move this?

web/pdf_find_controller.js
```diff
@@ -166,6 +206,10 @@ class PDFFindController {
       ) {
         return;
       }
+      if (cmd === "finddiacriticmatchingchange") {
+        this._pageDiffs.length = 0;
```
Snuffleupagus, line 210: Unless I'm overlooking something obvious here, it doesn't actually seem that you're guaranteed to ever re-populate `this._pageDiffs` after it's been cleared here!?

Won't that cause bugs, if the only thing changing is the "Match Diacritics" setting (for the same search term)?

web/pdf_find_controller.js
```diff
@@ -548,7 +602,6 @@ class PDFFindController {
               );
               // Page error -- assuming no text content.
               this._pageContents[i] = "";
-              this._pageDiffs[i] = null;
```
Snuffleupagus, line 551: Why remove this?

Snuffleupagus (changes requested):

test/unit/pdf_find_controller_spec.js
calixteman: I'll add one for `½` but I think the `1/2` is good too because the query is 3 chars when the highlight itself contains one char.

web/viewer.html
calixteman: About the case on `Diacritics`, I just followed what we already have in this file (lower case for the second word).
I just checked and in firefox, we've "Highlight All", "Match Case" and "Whole Words".
We can fix in a follow-up.

calixteman:
For now it doesn't work with RTL languages and I'll do that in an other patch.

web/text_layer_builder.css
```diff
@@ -24,7 +24,7 @@
   line-height: 1;
 }
 
-.textLayer span {
+.textLayer > span {
```
Snuffleupagus: Is this change intended and correct, or is it perhaps a rebase error?

web/pdf_find_controller.js
```diff
@@ -533,12 +554,14 @@ class PDFFindController {
...
+                  strBuf.push("\n");
+                }
               }
 
               // Store the normalized page content (text items) as one string.
-              [this._pageContents[i], this._pageDiffs[i]] = normalize(
-                strBuf.join("")
-              );
+              const query = strBuf.join("");
+              [this._pageContents[i], this._pageDiffs[i]] = normalize(query);
```
Snuffleupagus, line 564: Looks like an unrelated change, and the `query` variable name looks strange here (since it's actually page-contents); please just revert these changes.

web/pdf_find_controller.js
```diff
@@ -358,136 +448,67 @@ class PDFFindController {
...
-      const limit = content.charCodeAt(endIdx + 1);
-      if (getCharacterType(last) === getCharacterType(limit)) {
-        return false;
-      }
-    }
-    return true;
-  }
-
-  _calculatePhraseMatch(query, pageIndex, pageContent, pageDiffs, entireWord) {
+  _calculateRegExMatch(query, pageIndex, pageContent) {
```
Snuffleupagus, line 451: `_calculateRegExpMatch` to be consistent with the other helper method.

web/pdf_find_controller.js
```diff
@@ -152,6 +240,7 @@ class PDFFindController {
     if (this._state === null || this._shouldDirtyMatch(cmd, state)) {
       this._dirtyMatch = true;
     }
+
```
Snuffleupagus, line 243: Nit: Looks unrelated, please revert.

web/pdf_find_controller.js
```diff
@@ -166,6 +255,7 @@ class PDFFindController {
       ) {
         return;
       }
+
```
Snuffleupagus, line 258: Nit: Looks unrelated, please revert.

web/pdf_find_controller.js
```diff
@@ -42,47 +41,136 @@ const CHARACTERS_TO_NORMALIZE = {
...
+        const jj = replacement.length;
+        for (let j = 1; j < jj; j++) {
+          positions.push([i - shift + j, shift - j]);
+        }
+        shift -= jj - 1;
+        return replacement;
+      }
+
+      if (p2) {
+        // -\n is removed because a - at eol is because of an hyphenation.
```
Snuffleupagus, line 118: This sentence is a little bit hard to read, can it perhaps be re-written/re-worded slightly?

Snuffleupagus (changes requested):

web/pdf_find_controller.js
```diff
@@ -358,136 +449,80 @@ class PDFFindController {
...
-    // Sort `matchesWithLength`, remove intersecting terms and put the result
-    // into the two arrays.
-    this._prepareMatches(
-      matchesWithLength,
-      this._pageMatches[pageIndex],
-      this._pageMatchesLength[pageIndex]
-    );
+    if (entireWord) {
+      query = `\\b${query}\\b`;
+    }
```
Snuffleupagus, line 499: Are you *absolutely sure* that this catches *all* of the cases that the old `getCharacterType` functionality did?
Instinctively this does "feel" a lot less complete than the previous code, which compared different character classes in some detail; that implementation was borrowed from https://searchfox.org/mozilla-central/rev/083983b7f09b00cdfe4f70396e39ea4f8a1735e1/intl/lwbrk/WordBreaker.cpp#44-120

Snuffleupagus (changes requested):

web/pdf_find_controller.js
calixteman: I thought `\b` was for unicode too... so I'll fix that, thanks for catching that.

Snuffleupagus:
> * remove pdf_find_utils.js.

You probably want to remove this from the commit message (and PR description) now :-)

---

Given that this implementation uses a lot more, and more complex, regular expressions during both initial text-parsing and subsequent searching: What sort of performance impact, if any, does this patch have for larger and/or more complex documents?

For example, what about e.g. the `pdf.pdf` document (in the test-suite) or perhaps [kjv.pdf](https://github.com/mozilla/pdf.js/files/876321/kjv.pdf) (which is even longer)?

calixteman:
Good questions.
I used some `perfomance.now` calls at the beginning and at the end of functions `normalize` and `_calculateRegExpMatch` and I used file `kjv.pdf` (and set `privacy.reduceTimerPrecision` to false):
Normalization (ms): 69, 63, 78, 60,
Search for `"a"` (ms): 316, 296, 306, 253

Normalization (ms): 71, 63, 77, 77
Search for `"and  thou  wast  not  spoiled;"` (ms): 32, 32, 36, 33

And the same in master with `normalize` and `_calculatePhraseMatch`:
Normalization (ms): 25, 25, 22, 20,
Search for `"a"` (ms): 41, 41, 40, 40

Normalization (ms): 29, 29, 27, 20
Search for `"and  thou  wast  not  spoiled;"` (ms): 7, 7, 7, 7

From a user pov, I don't see any differences with both searches and so I think that this perf regression is acceptable.

I added some code to remove the diacritics stuff in the query regexp when there are no diacritics on the page (which is the case in kjv.pdf) and there are no significant difference in time for both searches, so in the search part we pay for the use of regexp instead of  `indexOf` and not so much because of diacritics.
One big advantage of regexp over `indexOf` is that it's very simple to search for `foo\s+bar` (because the pdf can contain multiple spaces between foo and bar) and we don't have to re-normalize pageContent when matchDiacritics is turned on/off or to change pageContent case when caseSensitive is turned on/off.  

And as usual, I'm open to any good idea to improve this.

calixteman:
@timvandermeij, @Snuffleupagus do you have any objections for landing that stuff ? or any idea to improve perf or whatever ?

Snuffleupagus:
> do you have any objections for landing that stuff ?

I've been a little bit short on time to really review this properly, and have only looked *briefly* at the implementation, so it'd probably be very good idea to actually do a "full" review before landing it since this is pretty fundamentally change to the find implementation.

timvandermeij:
Same here. I think it would be good to, aside from the full review, wait until #13418 is done before merging since it's quite a change to the implementation and it would be good to get the release out first to avoid any risks. (We usually do this for significant changes, such as the text layer and struct tree PRs.)

lixaotec:
Is this related to #10036 anyhow?

I look foward regex search for multiple terms, and also advanced highlight with custom html tags. 

Is there any PR or roadmap improvments on that way?

thanks

web/viewer.js
```diff
@@ -154,6 +154,7 @@ function getViewerConfiguration() {
       highlightAllCheckbox: document.getElementById("findHighlightAll"),
       caseSensitiveCheckbox: document.getElementById("findMatchCase"),
       entireWordCheckbox: document.getElementById("findEntireWord"),
+      matchDiacriticsCheckbox: document.getElementById("findMatchDiacritics"),
```
Snuffleupagus, line 157: Nit: Let's move this up *one* line, since the new checkbox is placed *before* the entireWord-one in the findbar.

web/pdf_find_bar.js
```diff
@@ -33,6 +33,7 @@ class PDFFindBar {
     this.highlightAll = options.highlightAllCheckbox;
     this.caseSensitive = options.caseSensitiveCheckbox;
     this.entireWord = options.entireWordCheckbox;
+    this.matchDiacritics = options.matchDiacriticsCheckbox;
```
Snuffleupagus, line 36: Nit: Let's move this up one line, since the new checkbox is placed before the entireWord-one in the findbar (similar with the addEventListener placement below).

web/pdf_find_controller.js
```diff
@@ -43,47 +43,144 @@ const CHARACTERS_TO_NORMALIZE = {
   "\u00BE": "3/4", // Vulgar fraction three quarters
 };
 
+const diacriticsRegExp = /\p{Mn}+/gu;
+const escapeRegExp = /[.*+\-?^${}()|[\]\\]/g;
+const prepareNoDiacriticsRegExp = /(\p{Mn})|(\p{L})/gu;
+const whitespacesRegExp = /\s+/g;
+const notDiacriticFromEndRegExp = /([^\p{Mn}])\p{Mn}*$/u;
+const notDiacriticFromStartRegExp = /^\p{Mn}*([^\p{Mn}])/u;
```
Snuffleupagus, line 51: Since these are true constants, please use the upper-case format instead (e.g. `DIACRITICS_REG_EXP` and similar).

Snuffleupagus (commented):
This is going to take me some time to try and understand, w.r.t. all the `web/pdf_find_controller.js` changes, so just leaving a couple of small nits in the meantime.

web/pdf_find_controller.js
```diff
@@ -43,47 +43,144 @@ const CHARACTERS_TO_NORMALIZE = {
...
+        for (let j = 1; j < jj; j++) {
+          positions.push([i - shift + j, shift - j]);
+        }
+        shift -= jj - 1;
+        return replacement;
+      }
+
+      if (p2) {
+        // "-\n" is removed because an hypen at the end of a line
+        // is likely here to mark a break in a word.
```
Snuffleupagus, line 127: Unless I'm misunderstanding this part, there's a *potential* false positive here.
 - As-is this is intended to replace the e.g. the following text:
   ```
   trace-
   monkey
   ```
   with "tracemonkey".
 - However, could it not also lead e.g. to:
   ```
   25 -
   5
   ```
   being replaced with "25 5" which would be wrong? This would not be limited to just numbers, since it'd affect words as well, but it's going to be especially noticeable with numbers.

Hence my question: Should we also check that the "-\n" part isn't preceded by a space here?

web/pdf_find_controller.js
```diff
@@ -43,47 +43,144 @@ const CHARACTERS_TO_NORMALIZE = {
...
+  let eol = 0;
+  let hasDiacritics = false;
+
+  normalized = normalized.replace(
+    normalizationRegex,
+    (match, p1, p2, p3, p4, i) => {
+      i -= shiftOrigin;
+      if (p1) {
+        // Fractions...
+        const replacement = CHARACTERS_TO_NORMALIZE[match];
```
Snuffleupagus, line 116: The comment should perhaps say "Maybe fractions..." instead, since `CHARACTERS_TO_NORMALIZE` contains mostly non-fraction entries?

Snuffleupagus (commented):
A more general question, since the concepts outlined in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes are somewhat unfamiliar to me:

Looking through some of the links on that MDN page, I noticed https://unicode.org/reports/tr44/#General_Category_Values which outlines the three categories Mn, Mc, Me. In this patch, you're only using the nonspacing-mark (i.e. `Mn`), which I'm assuming is on purpose?

calixteman:
> A more general question, since the concepts outlined in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes are somewhat unfamiliar to me:
> 
> Looking through some of the links on that MDN page, I noticed https://unicode.org/reports/tr44/#General_Category_Values which outlines the three categories Mn, Mc, Me. In this patch, you're only using the nonspacing-mark (i.e. `Mn`), which I'm assuming is on purpose?

That's a fair question.
I checked, few month ago, if spacing marks were handled in Firefox in using:
https://en.wikipedia.org/wiki/Visarga
and I searched for "विसर्ग" (which I created in using `"विसर्गः".normalize("NFD").split("").slice(0, 6).join("")`) and the string + `visarga` is matched whatever "Match diacritics" is checked or not. So I considered that this kind of diacritics where likely not covered by the search stuff in Firefox. 
Anyway, after thinking again about that, maybe I should reconsider this choice because it's likely a bug in Firefox.

web/pdf_find_controller.js
```diff
@@ -47,47 +47,177 @@ const CHARACTERS_TO_NORMALIZE = {
...
+  // https://www.compart.com/fr/unicode/combining/129
+  0x0f71,
+  // 130
+  // https://www.compart.com/fr/unicode/combining/130
+  0x0f72, 0x0f7a, 0x0f7b, 0x0f7c, 0x0f7d, 0x0f80,
+  // 132
+  // https://www.compart.com/fr/unicode/combining/132
+  0x0f74,
+]);
+const DIACRITICS_EXCEPTION_STR = Array.from(DIACRITICS_EXCEPTION.values())
```
Snuffleupagus, line 79: Nit: I'm sort-of partial to the slightly shorter `[...DIACRITICS_EXCEPTION]` format.

web/pdf_find_controller.js
```diff
@@ -331,193 +462,140 @@ class PDFFindController {
...
-  }
-
   /**
    * Determine if the search query constitutes a "whole word", by comparing the
    * first/last character type with the preceding/following character type.
    */
   _isEntireWord(content, startIdx, length) {
-    if (startIdx > 0) {
+    let match = content
+      .slice(0, startIdx)
```
Snuffleupagus, line 471: Here, and similar below, would it make sense to check that `match` isn't empty *before* running a then unnecessary regular expression?

web/pdf_find_controller.js
```diff
@@ -331,193 +462,140 @@ class PDFFindController {
...
+        match,
+        p1 /* to escape */,
+        p2 /* punctuation */,
+        p3 /* whitespaces */,
+        p4 /* diacritics */,
+        p5 /* letters */
+      ) => {
+        if (p1) {
+          // Escape characters like *+?... to not interfer with regexp syntax.
+          return `\\s*\\${p1}\\s*`;
```
Snuffleupagus, line 535: *This is potentially a case of a useless micro-optimization, but I wanted to ask nonetheless:*
Given that we call `getTextContent` with the `normalizeWhitespace = true` option set, all whitespaces are thus normal spaces (i.e. 0x20), hence I wonder if it would perhaps be more efficient to only check for that character rather than arbitrary white-space in these regular expressions?

web/pdf_find_controller.js
```diff
@@ -331,193 +462,140 @@ class PDFFindController {
...
-      this._pageMatches[pageIndex],
-      this._pageMatchesLength[pageIndex]
+        // A letter has been matched and it can be followed by any diacritics
+        // in normalized text.
+        return hasDiacritics ? `${p5}\\p{M}*` : p5;
+      }
     );
+
+    if (matchDiacritics) {
+      // aX musn't match aXY.
```
Snuffleupagus, line 561: Spelling-nit: `musn't` -> `mustn't`, or perhaps just `must not` which is a tiny bit easier to read.

web/pdf_find_controller.js
```diff
@@ -47,47 +47,177 @@ const CHARACTERS_TO_NORMALIZE = {
...
+  return [normalized, positions, hasDiacritics];
 }
 
-// Determine the original, non-normalized, match index such that highlighting of
-// search results is correct in the `textLayer` for strings containing e.g. "½"
-// characters; essentially "inverting" the result of the `normalize` function.
-function getOriginalIndex(matchIndex, diffs = null) {
-  if (!diffs) {
-    return matchIndex;
+function getOriginalIndex(positions, pos, len) {
```
Snuffleupagus, line 207: Any reason for renaming `diffs` to `positions` here?

Also, please note that the old function handled the `diffs` being `null` which can still happen for pages where text-extraction fails. Hence you'll most likely need to *still* handle that case here, to prevent any errors being thrown and breaking searching; please see the old function.

web/ui_utils.js
```diff
@@ -227,8 +227,8 @@ function removeNullCharacters(str, replaceInvisible = false) {
  * @returns {number} Index of the first array element to pass the test,
  *                   or |items.length| if no such element exists.
  */
-function binarySearchFirstItem(items, condition) {
-  let minIndex = 0;
+function binarySearchFirstItem(items, condition, start = 0) {
+  let minIndex = start;
```
Snuffleupagus, line 231: These changes probably deserve extending the existing unit-tests to also cover the new functionality; see https://github.com/mozilla/pdf.js/blob/master/test/unit/ui_utils_spec.js

web/pdf_find_controller.js
```diff
@@ -47,47 +47,177 @@ const CHARACTERS_TO_NORMALIZE = {
...
+        shift -= 1;
+        shiftOrigin += 1;
+        eol += 1;
+        return " ";
+      }
+
+      // Diacritics.
+      hasDiacritics = true;
+      let jj = p4.length;
+      if (i + eol === rawDiacriticsPositions?.[k]?.[1]) {
```
Snuffleupagus, line 185: Isn't `rawDiacriticsPositions` always defined here? If so, then I believe that the *first* optional chaining is probably not necessary here:
```javascript
      if (i + eol === rawDiacriticsPositions[k]?.[1]) {
```

Snuffleupagus (commented):
Sorry that this dropped off my TODO list previously, but it's a *bunch* of quite complicated code to try and make sense of.
I've just done a new pass through the code, and added some new questions and comments; given the size/scope of these change I also wonder if it'd make sense to get have @brendandahl review this as well?

web/pdf_find_controller.js
calixteman: Do you mean checking that `startIdx` is not zero ?
If yes, I'd say it's useless since content is the page content and very likely startIdx is not zero, and I hope there is an optimization to immediately return an empty match in case the string is empty.
But of course if you think it's better to check, no problem, I'll do.

brendandahl (approved):
I don't claim to understand everything here, but the changes seem reasonable, there are tests, and my concerns with escaping seem to be taken care of.

@Snuffleupagus  did you have anything else before landing? I'm hoping to do an update before the next Firefox soft freeze on 2022-02-03

calixteman:
I just changed the commit message.

Snuffleupagus (changes requested):
In the preview, using the default `tracemonkey.pdf` document, searching for any word containing a dash (e.g. "trace-" or "col-") breaks with a *bunch* of errors like this:
```
Uncaught (in promise) SyntaxError: invalid identity escape in regular expression
    _calculateMatch pdf_find_controller.js:605
    _nextMatch pdf_find_controller.js:732
    promise callback*_nextMatch pdf_find_controller.js:730
    _findTimeout pdf_find_controller.js:342
    setTimeout handler*__webpack_modules__</_onFind/< pdf_find_controller.js:342
    promise callback*_onFind pdf_find_controller.js:321
    dispatch event_utils.js:118
    dispatchEvent pdf_find_bar.js:98
    PDFFindBar pdf_find_bar.js:50
```

---

Another regression that I spotted, is that searching for *only* "col" now highlights the "col-" in the textLayer (i.e. one character too many).

calixteman:
```js
a="-"; new RegExp(`\\${a}`, "u").test("-")
```
doesn't really work as I expected. Anyway in removing the minus from the regex of the chars to escape it's fine.
Since the minus is only for range and between brackets, I think we don't have to escape it.
Anyway I'll add a test for that.

About the "col-":
 - in Acrobat, "col" leads to highlight "Col-laboration",
 - in Chrome, it leads to highlight "Col-" and "collaboration" leads to highlight "Col-laboration"
 - in Foxit, it leads to hightlight "Col" but "collaboration" doesn't produce anything.

So we have the same behaviour as in Chrome and almost the same as in Acrobat.

test/unit/pdf_find_controller_spec.js
```diff
@@ -264,5 +265,172 @@ describe("pdf_find_controller", function () {
...
+      },
+      matchesPerPage: [1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+      selectedMatch: {
+        pageIndex: 0,
+        matchIndex: 0,
+      },
+    });
+  });
+
+  it("performs a search where the result is on two lines with a punctution at eol", async function () {
```
Snuffleupagus, line 416: Spelling-nit: `punctution` -> `punctuation`.

web/app.js
```diff
@@ -2807,6 +2808,7 @@ function webViewerKeyDown(evt) {
               source: window,
               type: "again",
               findPrevious: cmd === 5 || cmd === 12,
+              matchDiacritics: state.matchDiacritics,
```
Snuffleupagus, line 2811: This line shouldn't be necessary now, right?
I'm guessing that this might be fallout from rebasing on top of other changes.

calixteman:
And for information, neither Acrobat nor Chrome finds something for "col-".

web/pdf_find_controller.js
```diff
@@ -331,193 +469,161 @@ class PDFFindController {
...
-        pageIndex,
-        pageContent,
-        pageDiffs,
-        entireWord
-      );
+      // Words are sorted in reverse order to be sure that "foobar" is matched
+      // before "foo" in case the query is "foobar foo".
+      query = query
+        .match(/\S+/g)
+        .sort()
```
Snuffleupagus, line 606: Would it be possible for `query` to contain only whitespaces here?
Because in that case `String.prototype.match()` will return `null` which will thus cause the `sort()`-call to throw an Error.

web/pdf_find_controller.js
```diff
@@ -331,193 +469,161 @@ class PDFFindController {
...
+        .reverse()
+        .map(q => {
+          const [isUnicodePart, partialQuery] = this._convertToRegExpString(
```
Snuffleupagus, line 609: Nit: For consistency, how about `partialQuery` -> `queryPart` instead?

web/pdf_find_controller.js
calixteman: Good catch, thank you.

test/unit/pdf_find_controller_spec.js
```diff
@@ -264,5 +265,327 @@ describe("pdf_find_controller", function () {
...
+      pageMatches: [[1946]],
+      pageMatchesLength: [[query.length]],
+    });
+  });
+
+  it("performs a search with a dot in the query and a missing whitespace", async function () {
+    const { eventBus, pdfFindController } = await initPdfFindController();
+
+    // The whitespace after the dot must be matched.
+    const query = "complex applications.J";
```
Snuffleupagus, line 570: Sorry, but I *really* don't understand how this makes sense!

Looking at the contents of the textLayer, there's clearly a whitespace between the dot and the following "J" which means that this query should *not* match.
What should however match is "complex applications. J" (note the added space), so what's happening here and more importantly why?

calixteman: In Acrobat (but not in Chrome), this query leads to highlight "complex applications. JavaScript" so the goal is to have something similar.
I guess that in some cases we can have a space in the text but the rendering doesn't strongly show it and so introducing of fuzziness in the query helps the user to find what he wants, as you know whitespace in pdf is a bit weird concept...

Snuffleupagus: Does the query "complex applications. J" (with a space) still work as expected?
If so that mostly alleviates my concerns (even though this adds *even more* complexity to an already very complex patch); can you please amend this unit-test to also check that case as well?

test/unit/pdf_find_controller_spec.js
```diff
@@ -264,5 +265,327 @@ describe("pdf_find_controller", function () {
...
+        phraseSearch: true,
+        findPrevious: false,
+      },
+      matchesPerPage: [1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+      selectedMatch: {
+        pageIndex: 0,
+        matchIndex: 0,
+      },
+      pageMatches: [[1946]],
+      pageMatchesLength: [[query.length]],
```
Snuffleupagus, line 562: Here, and elsewhere below, it feels slightly "safer" to *explicitly* specify the expected values since that reduces the risk of any future changes accidentally breaking something.

test/unit/pdf_find_controller_spec.js
```diff
@@ -264,5 +265,327 @@ describe("pdf_find_controller", function () {
...
+
+    await testSearch({
+      eventBus,
+      pdfFindController,
+      state: {
+        query: "1/2",
+        caseSensitive: false,
+        entireWord: false,
+        phraseSearch: true,
+        findPrevious: false,
```
Snuffleupagus, line 277: Nit: Here, and elsewhere below, note that you now only need to specify those options that *differ* from the default values in the `testSearch` function.
The advantage of using a more compact format in the individual test-cases is that it becomes easier to spot what's "special" about each test-case.

web/pdf_find_controller.js
```diff
@@ -331,193 +469,173 @@ class PDFFindController {
...
-        matchesWithLength.push({
-          match: originalMatchIdx,
-          matchLength: originalQueryLen,
-          skipped: false,
-        });
+        return p5;
       }
+    );
+
+    if (query.endsWith("[ ]*")) {
```
Snuffleupagus, line 578: Maybe add a `const trailingSpaces = "[ ]*";` (assuming that's a good name) just above this line and use that instead to avoid having to specify that string repeatedly?

web/pdf_find_controller.js
```diff
@@ -331,193 +469,173 @@ class PDFFindController {
...
 
-    if (!caseSensitive) {
-      pageContent = pageContent.toLowerCase();
-      query = query.toLowerCase();
-    }
+    const { caseSensitive, entireWord, phraseSearch } = this._state;
+    const pageContent = this._pageContents[pageIndex];
+    const hasDiacritics = this._hasDiacritics[pageIndex];
 
+    let flags = caseSensitive ? "g" : "gi";
```
Snuffleupagus, line 607: Maybe move this definition down to *just above* the `if (isUnicode) { ... }` block instead, to have all of that logic in one place?

calixteman:
@Snuffleupagus, do you see something else to fix ? or do you have any other concern ?
For your information, we'd like to have this feature in the next nightly cycle (the soft freeze is next week) and we'll ask to QA to check that everything is fine. If the quality is not good enough we'll backout and it I'll work on it to improve what is needed.

Snuffleupagus (approved):
> For your information, we'd like to have this feature in the next nightly cycle (the soft freeze is next week) and we'll ask to QA to check that everything is fine. 

OK, let's try landing this and see how it goes :-)
I've tried to do some manual testing, but that's obviously quite limited and given the scope/size of these changes we obviously need more widespread testing.

> If the quality is not good enough we'll backout and it I'll work on it to improve what is needed.

Unless there's *really* big problems, it might be easier to fix in place (and uplift patches) as necessary since trying to back-out a PR this size could very quickly become difficult.

PR #13263, opened 2021-04-19
Replace `done` callbacks in the font-tests with async/await instead

Snuffleupagus:


test/font/fontutils.js
```diff
@@ -65,24 +65,26 @@ function encodeFontData(data) {
   return buffer;
 }
 
-function ttx(data, callback) {
-  const xhr = new XMLHttpRequest();
-  xhr.open("POST", "/ttx");
+function ttx(data) {
+  return new Promise((resolve, reject) => {
```
timvandermeij, line 69: Can't we just make this function `async` instead so we don't need to wrap it in a `Promise` manually?

timvandermeij (commented):
Looks good to me with the comment addressed.

test/font/fontutils.js
Snuffleupagus: Unfortunately I cannot see how that'd work, given that `xhr.onreadystatechange` is a *callback* function that we cannot simply `await` :-(

We could *maybe* re-factor this function to use `fetch` instead, assuming this isn't using e.g. a `file:`-URL, but that's probably better left for a follow-up.

Snuffleupagus: It seems that using `fetch` instead worked just fine, so I've pushed a separate commit that fixes this :-)

timvandermeij (approved):

timvandermeij:
Looks really good, and using the fetch API is also much cleaner. Thank you for doing this!

PR #13272, opened 2021-04-20
Update all the text widgets having the same name with the same value

### LINKED ISSUE #13271: Text widgets with the same name must share the same value
[multiple.pdf](https://github.com/mozilla/pdf.js/files/6344777/multiple.pdf)

1. Write something in a field

The 3 fields must have the same value

The 2 other fields are empty

The three fields share the same name `Text3`.
It works as expected in Adobe and Chrome.

calixteman:
Fixes #13271.


test/integration/annotation_spec.js
```diff
@@ -87,3 +87,38 @@ describe("Checkbox annotation", () => {
...
+
+    beforeAll(async () => {
+      pages = await loadAndWait("issue13271.pdf", "[data-annotation-id='24R']");
+    });
+
+    afterAll(async () => {
+      await closePages(pages);
+    });
+
+    it("must all the fields with the same value", async () => {
```
Snuffleupagus, line 103: This reads a little bit weird, how about "all fields must have the same value" instead?

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #13280, opened 2021-04-21
Ensure that `saveDocument` works if there's no /ID-entry in the PDF document (issue 13279)

### LINKED ISSUE #13279: Crash on form saving if at least one form field is filled
Attach (recommended) or Link to PDF file here: [ACORD 0211 2016-09 Acroform BLANK.pdf](https://github.com/mozilla/pdf.js/files/6353664/ACORD.0211.2016-09.Acroform.BLANK.pdf)

Configuration:
- Web browser and its version: Google Chrome Version 90.0.4430.72 (Official Build) (x86_64)
- Operating system and its version: Windows 10 Pro 20H2
- PDF.js version: 2.7.570
- Is a browser extension: No

1. Load attached form into PDFViewer
2. Call saveDocument method on loaded document proxy

What is the expected behavior: should not crash

It crashes in writer.js. Line 273 as of 21 Apr 2021.
image

Line that crashes: 
image

Potential fix: 
image

Potential fix is based on this line if worker.js file: 
image

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): not hosted

Snuffleupagus:
First of all, while it should be very unlikely that the /ID-entry is an *indirect* object, note how we're using `Dict.get` when parsing it e.g. in `PDFDocument.fingerprint`. Hence we definitely should be consistent here, since if the /ID-entry is an *indirect* object the existing code in `src/core/writer.js` would already fail.
Secondly, to fix the referenced issue, we also need to check that the /ID-entry actually is an Array before attempting to access its contents in `src/core/writer.js`.

*Drive-by change:* In the `xrefInfo` object passed to the `incrementalUpdate` function, re-name the `encrypt` property to `encryptRef` since its data is fetched using `Dict.getRaw` (given the names of the other properties fetched similarly).

Fixes #13279

### DIFF
--- a/test/unit/writer_spec.js
+++ b/test/unit/writer_spec.js
@@ -32,7 +32,7 @@ describe("Writer", function () {
         fileIds: ["id", ""],
         rootRef: null,
         infoRef: null,
-        encrypt: null,
+        encryptRef: null,
         filename: "foo.pdf",
         info: {},
       };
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
While PDF document *without* /ID is quite rare, as the issue shows they unfortunately do exist.
Given that the patch is small, and in my opinion quite safe, this might be the sort of thing that we should consider uplifting (once it's landed of course)?

test/unit/writer_spec.js
```diff
@@ -59,6 +59,41 @@ describe("Writer", function () {
...
+        "\x00\x01\xff\xff" +
+        "\x01\x01\x00\x2d" +
+        "\x01\x05\x00\x00\n" +
+        "endstream\n" +
+        "endobj\n" +
+        "startxref\n" +
+        "5\n" +
+        "%%EOF\n";
+
+      console.log(data);
```
calixteman, line 94: you probably forgot to remove that.

Snuffleupagus: Whoops; fixed now :-)

calixteman (approved):
LGTM, thanks for doing this.

PR #13283, opened 2021-04-22
Change `NameOrNumberTree.getAll` to return a `Map` rather than an Object

Snuffleupagus:
Given that we're (almost) always iterating through the result of the `getAll`-calls, using a `Map` seems nicer overall since it's more suited to iteration compared to a regular Object.

Also, add a couple of `Dict`-checks in existing code touched by this patch, since it really cannot hurt to prevent *potential* errors in a corrupt PDF document.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -894,18 +894,20 @@ class Catalog {
 
   _collectJavaScript() {
     const obj = this._catDict.get("Names");
-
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

Issue #12593, opened 2020-11-06
Impossibility to render correctly a form-fillable PDF


Luni-4:
Attach (recommended) or Link to PDF file here:  [form.pdf](https://github.com/mozilla/pdf.js/files/5556500/form.pdf)

Configuration:
- Web browser and its version: 82.0.2
- Operating system and its version: Ubuntu 20.10
- PDF.js version:
- Is a browser extension: no

Steps to reproduce the problem:
1. Open one of the PDF files attached above with Mozilla 82.0.2
2. Only the basic sheet is rendered, but it is not possible to fill the forms and view the JavaScript menu that allows to generate additional pages

What is the expected behavior? (add screenshot)

The possibility to fill the sheet, in addition to view and use the JavaScript menu that allows to generate the additional pages.

The JavaScript code present within the PDF file is open-source and you can find it [here](https://github.com/morepurplemorebetter/MPMBs-Character-Record-Sheet). Since that code makes use of the [AcroJS API](https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/js_api_reference.pdf), the PDF file could be rendered correctly using the last free version of Adobe Acrobat Reader.

Thanks in advance for your help!

Snuffleupagus:
> Attach (recommended) or Link to PDF file here: PDF files can be downloaded [here](https://www.flapkan.com/#download)

Please actually *attach* the test-case here, since that generally helps save everyone's time, given that the link doesn't point directly to a PDF file and even more importantly that it also also seems to require *payment* to even access the files.

As-is, this issue unfortunately isn't actionable and should probably be closed as a duplicate of one (or more) of the already open issues in the https://github.com/mozilla/pdf.js/labels/4-form-xfa category.

Luni-4:
> > Attach (recommended) or Link to PDF file here: PDF files can be downloaded [here](https://www.flapkan.com/#download)
> 
> Please actually _attach_ the test-case here, since that generally helps save everyone's time, given that the link doesn't point directly to a PDF file and even more importantly that it also also seems to require _payment_ to even access the files.

Oh, sorry about that, I didn't want to waste your time, but I don't think those files can be re-uploaded somewhere else, that's why I posted the link. Anyway, it is not necessary to require a payment to access them, you can download them setting up the offer value to 0, it is just a voluntary offer to help developers.

> 
[...]

PR #13286, opened 2021-04-23
[JS] Use heap allocation when initializing quickjs sandbox

calixteman:
- In case of large string the sandbox initialization failed because of an OOM
    * so allocate a new string in the heap
    * and free it after use.
  - it requires a quickjs update since we need to export some symbols (stringToNewUTF8 and free).


Snuffleupagus:
Do you have an example of a PDF document, that you can share, which triggers this behaviour?

calixteman:
Sure, the pdf attached to this issue:
https://github.com/mozilla/pdf.js/issues/12593
contains a lot of JS and the final string (sent to the sandbox) is around 6Mb.

Snuffleupagus (approved):
r=me, since sandbox initialization now works in the GENERIC viewer.

PR #13291, opened 2021-04-24
Replace a bunch of `Array.prototype.forEach()` cases with `for...of` loops instead

Snuffleupagus:
Using `for...of` is a modern and generally much nicer pattern, since it gets rid of unnecessary callback-functions. (In a couple of spots, a "regular" `for` loop had to be used.)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13294, opened 2021-04-25
Enable the `no-var` linting rule in `src/core/{cmap,image,worker}.js`

timvandermeij:

### DIFF
--- a/src/core/cmap.js
+++ b/src/core/cmap.js
@@ -12,7 +12,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint-disable no-var */
 
 import {
   CMapCompressionType,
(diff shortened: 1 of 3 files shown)


src/core/cmap.js
```diff
@@ -875,37 +873,37 @@ var CMapFactory = (function CMapFactoryClosure() {
...
     if (isName(obj) && isString(obj.name)) {
       cMap.name = obj.name;
     }
   }
 
   async function parseCMap(cMap, lexer, fetchBuiltInCMap, useCMap) {
-    var previous;
-    var embeddedUseCMap;
+    let previous;
+    let embeddedUseCMap;
```
Snuffleupagus, line 903: Nit: I *think* that you can change this into a one-liner, since the following should be allowed:
```suggestion
    let previous, embeddedUseCMap;
```

Snuffleupagus (approved):
Looks good to me, thank you!

PR #13297, opened 2021-04-25
Add a note about minification to the webpack-example README (issue 13290)

### LINKED ISSUE #13290: DocException problem with webpack in production mode
This is not related to a specific pdf file.

## Configuration:

  Web browser and its version: Any (latest firefox and chrome tested)
  Operating system and its version: macOS 11.2
  PDF.js version: latest
  Is a browser extension: no (using the `examples/webpack` in this repo, and pdfjs-dist in a separate project using default optimize options for webpack)

## Issue Description:

pdf.js [uses `this.constructor.name` in error objects](https://github.com/mozilla/pdf.js/blob/da0e7ea96984539419b5cc38adb378e7dfd1f1fb/src/shared/util.js#L456), and tries to [use that name in a switch case](https://github.com/mozilla/pdf.js/blob/da0e7ea96984539419b5cc38adb378e7dfd1f1fb/src/display/api.js#L2468). Webpack's default minimizer, Terser, removes these names in default production mode, so the switch case never matches, causing the error to be lost in production builds.

I think one solution could be to refactor the error classes to use explicitly declared name strings instead of relying on their constructor names. I'd take a stab at implementing this if you would like me to.

1. In the examples/webpack/main.js, change the pdfPath to a nonexistent pdf, e.g. 

    ```javascript
    const pdfPath = "../learning/404.pdf";
    ```
2. Follow the instructions in examples/webpack/README.md to run the webpack example, except use production webpack on the 4th step:
    
    ```bash
    ./node_modules/webpack/bin/webpack.js --mode=production
    ```

image

image

## Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): 

N/A

Snuffleupagus:
Since we really don't want to let a particular Webpack-mode dictate how we can/can't write code, let's add a note in the webpack-example README about minification instead.

Fixes #13290

---

I'd also suggest that we amend the FAQ-entry, i.e. https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#minified, with something like:
```
If you are not using `gulp minified`, please note that you *must* configure the minifier to keep original class/function names intact; otherwise the build is not guaranteed to work correctly.
```

### DIFF
--- a/examples/webpack/README.md
+++ b/examples/webpack/README.md
@@ -18,6 +18,12 @@ Refer to the `main.js` and `webpack.config.js` files for the source code.
 Note that PDF.js packaging requires packaging of the main application and
 the worker code, and the `workerSrc` path shall be set to the latter file.
 
+### Minification
+
+If you are configuring Webpack to output a minified build, please note that you
+*must* configure the minifier to keep original class/function names intact;
+otherwise the build is not guaranteed to work correctly.
+
 ## Worker loading
 
 If you are getting the `Setting up fake worker` warning, make sure you are


timvandermeij (approved):

timvandermeij:
Thanks! The FAQ is also updated now.

PR #13300, opened 2021-04-26
Convert the code in `src/display/canvas.js` to use standard classes

Snuffleupagus:
This gets rid of *a lot* of boilerplate that stems from our old way of simulating classes, and it actually reduces the filesize noticeably.
For e.g. `gulp mozcentral`, the *built* `pdf.js` files decreases from `318 404` to `314 722` bytes (~1 percent) with this patch.

*Much smaller/simpler diff with https://github.com/mozilla/pdf.js/pull/13300/files?w=1*


timvandermeij (approved):

timvandermeij:
Nice work!

PR #13301, opened 2021-04-26
Support `InkAnnotation`s without appearance streams (issue 13298)

### LINKED ISSUE #13298: Ink-Annotations not shown
[output.pdf](https://github.com/mozilla/pdf.js/files/6374987/output.pdf)

Configuration:
- Web browser and its version: Firefox 88.0
- Operating system and its version: macOS Big Sur 11.2.3 (20D91)
- PDF.js version: 2.8.243
- Is a browser extension: no/Firefox built-in

1.  add Ink-Annotations to a pdf (i use https://github.com/highkite/pdfAnnotate, Annotations with appearance streams are shown correctly)
2. open output file in firefox

The annotations should be visible.
Bildschirmfoto 2021-04-26 um 09 15 48

What went wrong? 
The annotations are added to the dom but invisible.
Bildschirmfoto 2021-04-26 um 09 18 26

Tested successfull with FoxitReader, AcrobatReader, Chrome and Safari.

Snuffleupagus:
For now, we keep things purposely simple by using straight lines (rather than curves); please see https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G11.2096579

Fixes #13298


src/core/annotation.js
```diff
@@ -2552,6 +2552,36 @@ class InkAnnotation extends MarkupAnnotation {
...
+      // The default stroke color is black.
+      const strokeColor = this.color
+        ? Array.from(this.color).map(c => c / 255)
+        : [0, 0, 0];
+
+      const borderWidth = this.borderStyle.width || 1;
+
+      this._setDefaultAppearance({
+        xref: parameters.xref,
+        extra: `${borderWidth} w`,
```
calixteman, line 2566: After reading the specs for Ink annotation, I think you forgot to handle dash patterns.

Snuffleupagus: In this patch I simply implemented the bare minimum required to fix the referenced issue, and I think that we'd need *real world* test-cases to be able to implement dash patterns correctly here (and that that could perhaps be done in a follow-up instead).

Snuffleupagus: Looking at the PDF document in the issue, it contains a `[3]` dashArray (i.e. the default value) which according to https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#G7.3881971 thus would suggest that there should be gaps in the rendered InkAnnotation.

*However*, if you actually look at the rendering and/or Appearance stream that Adobe Reader generates for that PDF document (or even the actual rendering in PDFium), there's *no* dashes present and only solid lines being rendered.
Furthermore, I've tried modifying the reduced test-case included in this patch to specify a non-standard dashArray, and both Adobe Reader and PDFium seem to outright ignore those values and only render the InkAnntations with solid lines.

All-in-all, ignoring the dashArray seems OK in practice here as far as I can tell.

calixteman (approved):
LGTM.

src/core/annotation.js
marco-c: @Snuffleupagus if you want I can try to find usage of dash patterns in my corpus of PDFs, I "just" need a way to detect them (like https://github.com/marco-c/pdf-finder/blob/f045328bbda0bf3a371e992e5c843ba23c02d502/run.py#L20-L37).

timvandermeij:

PR #13303, opened 2021-04-27
Add an abstract base-class, which all the various Stream implementations inherit from

Snuffleupagus:
By having an abstract base-class, it becomes a lot clearer exactly which methods/getters are expected to exist on all Stream instances.
Furthermore, since a number of the methods are *identical* for all Stream implementations, this reduces unnecessary code duplication in the `Stream`, `DecodeStream`, and `ChunkedStream` classes.

For e.g. `gulp mozcentral`, the *built* `pdf.worker.js` files decreases from `1 619 329` to `1 616 115` bytes with this patch-series.

*The only way to review this is probably __one commit at a time__, with the `?w=1` flag appended to the URL.*


src/core/base_stream_class.js
```diff
@@ -0,0 +1,99 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { shadow, unreachable } from "../shared/util.js";
+
+class BaseStreamClass {
```
timvandermeij, line 18: I think it would be better to remove `Class` from the class name and the file name. It's already visible that it's a class and we don't do this elsewhere (compare e.g., `BasePreferences`, `BaseLocalCache`, `BaseCanvasFactory` and more).

timvandermeij (commented):
Would it be an idea to maybe split the first five commits off into their own PR since they are only about converting existing files to standard classes, while the other commits are also creating and organizing new files? That would make this PR a bit more manageable.

src/core/base_stream_class.js
Snuffleupagus: I was going back and forth quite on lot on the naming here, since I was a little bit worried about `BaseStream` sounding just like any other `...Stream`. However, given that that one cannot create an instance of it, but only extend it, I suppose that worry is unfounded; I'll change the name as suggested :-)

timvandermeij:
Looks good so far. I'll try to do a full review this weekend since I'll need some time to carefully go over all commits :-)

timvandermeij (approved):

timvandermeij:
Thank you for all this work; looks much better!

Snuffleupagus:
@timvandermeij Thanks a lot for reviewing this patch series; I realize that it cannot have been all that easy, or even fun for that matter!

PR #13314, opened 2021-04-28
For mozcentral use Firefox color theme instead of system theme.

brendandahl:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1701691

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -424,20 +424,23 @@ const PDFViewerApplication = {
     try {
       const styleSheet = document.styleSheets[0];
       const cssRules = styleSheet?.cssRules || [];
+      const mediaMatcher =
+        typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")
+          ? "-moz-toolbar-prefers-color-scheme"
+          : "prefers-color-scheme";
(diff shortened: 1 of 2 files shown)


web/app.js
```diff
@@ -424,20 +424,23 @@ const PDFViewerApplication = {
     try {
       const styleSheet = document.styleSheets[0];
       const cssRules = styleSheet?.cssRules || [];
+      let mediaMatcher = "prefers-color-scheme";
+      if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")) {
+        mediaMatcher = "-moz-toolbar-prefers-color-scheme";
+      }
```
Snuffleupagus, line 430: Nit: I'd probably suggest just setting the value *directly*, rather than first defining it and then immediately overwriting it here:
```javascript
const mediaMatcher =
  typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")
    ? "-moz-toolbar-prefers-color-scheme"
    : "prefers-color-scheme";
```

Snuffleupagus (approved):
Looks reasonable enough to me, with one small nit; thank you!

PR #13319, opened 2021-04-30
JS -- add support for page property in field

calixteman:
Page property is the page where the field is.
Currently the page is `undefined` which can lead to errors (see bug [1708641](https://bugzilla.mozilla.org/show_bug.cgi?id=1708641)).


Snuffleupagus (changes requested):

src/core/document.js
```diff
@@ -1241,29 +1243,35 @@ class PDFDocument {
...
-      this._collectFieldObjects("", fieldRef, fieldPromises);
-    }
+    const fieldObjectsPromise = this.pdfManager
+      .ensureCatalog("toplevelPagesDict")
+      .then(pages => {
+        let kids = pages.getArray("Kids");
+        kids = Array.isArray(kids) ? kids : [];
+        const pageRefs = new Map(
+          kids.map((x, i) => [isRef(x) ? x.toString() : "", i])
+        );
```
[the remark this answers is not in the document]
Snuffleupagus: > I was thinking to something similar but with a cache to avoid too much calls to `getPageIndex`

How much of an issue is this going to be in real-world documents, I wonder? Obviously we can add a cache, possibly in `Catalog.getPageIndex`, but I'm not sure how necessary it actually is right now.

 > finally I was wondering if it'd be better to just make one loop to generate a map pageRef => pageIndex which could be useful in `getPageIndex` itself.

Personally, I'd strongly suggest that we do *NOT* do that!
There's no guarantee that every page of a document has form elements, and generally with all the lazy initialization/parsing that we do it seems *wrong* to force the *entire* /Pages-tree to be loading/parsed upfront like that (and it'd basically end up "breaking" to whole idea of e.g. `disableAutoFetch`).

src/core/document.js
```diff
@@ -507,7 +507,8 @@ class Page {
               annotationRef,
               this.pdfManager,
               this._localIdFactory,
-              /* collectFields */ false
+              /* collectFields */ false,
+              null
```
Snuffleupagus, line 511: This is no longer used, please revert this change.

src/core/document.js
```diff
@@ -1241,29 +1242,30 @@ class PDFDocument {
...
+        }
+        return Promise.all(allPromises).then(() => allFields);
+      });
 
-    return shadow(
-      this,
-      "fieldObjects",
-      Promise.all(allPromises).then(() => allFields)
-    );
+    return shadow(this, "fieldObjects", fieldObjectsPromise);
```
Snuffleupagus, line 1268: None of this ought to be necessary any more, please revert these changes.

Snuffleupagus (approved):
r=me, with the final comments addressed and all tests passing; thank you!

PR #13320, opened 2021-05-01
Add a new `BaseStream.getString(...)` method to replace manual `bytesToString(BaseStream.getBytes(...))` calls

Snuffleupagus:
Given that the `bytesToString(BaseStream.getBytes(...))` pattern is somewhat common throughout the `src/core/` code, it cannot hurt to add a new `BaseStream`-method which handles that case internally.

### DIFF
--- a/src/core/base_stream.js
+++ b/src/core/base_stream.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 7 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13321, opened 2021-05-01
Enable the `no-var` linting rule in `src/core/{crypto,function}.js`

timvandermeij:

### DIFF
--- a/src/core/crypto.js
+++ b/src/core/crypto.js
@@ -12,7 +12,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint-disable no-var */
 
 import {
   bytesToString,
@@ -32,8 +31,8 @@ class ARCFourCipher {
   constructor(key) {
     this.a = 0;
     this.b = 0;
-    var s = new Uint8Array(256);
-    var i,
+    const s = new Uint8Array(256);
+    let i,
       j = 0,
       tmp,
       keyLength = key.length;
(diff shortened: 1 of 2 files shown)


src/core/crypto.js
```diff
@@ -32,15 +32,14 @@ class ARCFourCipher {
     this.a = 0;
     this.b = 0;
     const s = new Uint8Array(256);
-    let i,
-      j = 0,
-      tmp,
-      keyLength = key.length;
-    for (i = 0; i < 256; ++i) {
+    const keyLength = key.length;
+    let j = 0;
```
Snuffleupagus, line 36: Nit: You could *maybe* just move this definition to the second loop below (at line 41), since it's only used within that loop unless I'm missing something here:
```javascript
for (let i = 0, j = 0; i < 256; ++i) {
```

Snuffleupagus (approved):
r=me, looks good to me; thanks for doing this!

Snuffleupagus:
*Just a FYI, so that we duplicate work here:* I've started looking at splitting the `src/core/fonts.js` file a bit, similar to (but not as invasive as) the recent Streams re-factoring. As part of those changes I'd obviously fix the `no-var` errors as well, so feel free to skip that file :-)

---

Also, a warning if you feel like trying to fix `src/core/evaluator.js`, since I looked at that one awhile back and decided to ignore it for now:
There's quite a few cases in that file which cannot be fixed by ESLint, and a fair number of them were *surprisingly difficult* to fix manually. Particularly around the `Promise`s and `while`-loops in the `getOperatorList`/`getTextContent` methods, there's some quite tricky scope-issues which aren't totally obvious at first sight. If I recall correctly, it's easy to make manual changes which ESLint happily accepts but which however leads to run-time errors in the viewer/tests.

timvandermeij:
Ah, thanks, I won't touch those files for now then. For the evaluator it might be best to first do some refactoring then before applying ESLint to make the process less error-prone.

PR #13324, opened 2021-05-02
Fix highlighting of search results when the `textLayer` contains `br`-elements (PR 13257 follow-up, issue 13323)

### LINKED ISSUE #13323: Highlight when searching can be misplaced
https://github.com/mozilla/pdf.js/files/1426655/Sample.PDF.document.pdf

Configuration:
https://mozilla.github.io/pdf.js/web/viewer.html in Firefox nightly (90).

1. Open the pdf
2. Search for "acting  as  single  judge" (whitespaces must be doubled)

Highlights on match should be well positioned.

Highlight for "as" is too low.

It's very likely an other regression caused by https://github.com/mozilla/pdf.js/pull/13171 (in the css).

Snuffleupagus:
Apparently we need to layout `br`-elements in the same *exact* way as the regular `span`-elements which contain the text-content.

Fixes #13323

### DIFF
--- a/test/text_layer_test.css
+++ b/test/text_layer_test.css
@@ -23,7 +23,8 @@
   bottom: 0;
   line-height: 1;
 }
-.textLayer span {
+.textLayer span,
+.textLayer br {
   position: absolute;
   white-space: pre;
   -webkit-transform-origin: 0% 0%;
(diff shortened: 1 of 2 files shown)


calixteman:
Both issues in #13323 are fixed thanks to this patch.

calixteman (approved):
LGTM, thanks for doing this.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #13327, opened 2021-05-03
Split the functionality in `src/core/fonts.js` into multiple files, and use standard classes

Snuffleupagus:
Obviously the `Font`-class is still *very* large, given particularly how TrueType fonts are handled, however this patch-series at least improves things by moving a number of functions/classes into their own files.
As a follow-up it might make sense to try and re-factor/extract the TrueType parsing into its own file, since all of this code is quite old, however that's probably best left for another time.

For e.g. `gulp mozcentral`, the *built* `pdf.worker.js` files decreases from `1 620 332` to `1 617 466` bytes with this patch-series.


timvandermeij (approved):

timvandermeij:
Nice work!

PR #13328, opened 2021-05-03
JS - Add support for display property

calixteman:
With this property, a field can be visible or not on screen or when printing.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -439,13 +439,40 @@ class Annotation {
     );
   }
 
-  isHidden(annotationStorage) {
+  /**
+   * Check if the annotation must be displayed by taking into account
+   * the value found in the annotationStorage which may have been set
+   * through JS.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {AnnotationStorage} [annotationStorage] - Storage for annotation
+   */
+  mustBeViewed(annotationStorage) {
     const storageEntry =
       annotationStorage && annotationStorage.get(this.data.id);
     if (storageEntry && storageEntry.hidden !== undefined) {
-      return storageEntry.hidden;
+      return !storageEntry.hidden;
     }
-    return this._hasFlag(this.flags, AnnotationFlag.HIDDEN);
+    return this.viewable && !this._hasFlag(this.flags, AnnotationFlag.HIDDEN);
+  }
+
+  /**
+   * Check if the annotation must be printed by taking into account
+   * the value found in the annotationStorage which may have been set
+   * through JS.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {AnnotationStorage} [annotationStorage] - Storage for annotation
+   */
+  mustBePrinted(annotationStorage) {
+    const storageEntry =
+      annotationStorage && annotationStorage.get(this.data.id);
+    if (storageEntry && storageEntry.print !== undefined) {
+      return storageEntry.print;
+    }
+    return this.printable;
   }
 
   /**
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -68,13 +68,6 @@ import { XRef } from "./xref.js";
 const DEFAULT_USER_UNIT = 1.0;
 const LETTER_SIZE_MEDIABOX = [0, 0, 612, 792];
 
-function isAnnotationRenderable(annotation, intent) {
-  return (
-    (intent === "display" && annotation.viewable) ||
-    (intent === "print" && annotation.printable)
-  );
-}
-
 class Page {
   constructor({
     pdfManager,
@@ -274,7 +267,7 @@ class Page {
     return this._parsedAnnotations.then(function (annotations) {
       const newRefsPromises = [];
       for (const annotation of annotations) {
-        if (!isAnnotationRenderable(annotation, "print")) {
+        if (!annotation.mustBePrinted(annotationStorage)) {
           continue;
         }
         newRefsPromises.push(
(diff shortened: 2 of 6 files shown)


src/core/annotation.js
```diff
@@ -439,13 +439,22 @@ class Annotation {
...
+  isPrintable(annotationStorage) {
+    const storageEntry =
+      annotationStorage && annotationStorage.get(this.data.id);
+    if (storageEntry && storageEntry.print !== undefined) {
+      return storageEntry.print;
+    }
+    return this.printable;
+  }
+
+  isRenderable(annotationStorage) {
```
Snuffleupagus, line 451: Nit: `isViewable` instead, such that it matches the `viewable` property used below?

src/core/document.js
```diff
@@ -482,7 +479,9 @@ class Page {
     return this._parsedAnnotations.then(function (annotations) {
       const annotationsData = [];
       for (let i = 0, ii = annotations.length; i < ii; i++) {
-        if (!intent || isAnnotationRenderable(annotations[i], intent)) {
+        // Get the annotation even if it's hidden because
+        // JS can change its display.
+        if (annotations[i].viewable) {
```
Snuffleupagus, line 484: You probably want to keep `!intent || ...` here, since otherwise it'd be an (indirect) API change as.

Snuffleupagus, line 484: Won't this be an observable API change as-is, since you're now return all viewable annotations *regardless* of the `intent`-value?
I believe that what you actually want here is something like this:
```javascript
if (!intent || (intent === "display" && annotations[i].viewable)) {
```

src/core/document.js
```diff
@@ -376,10 +376,7 @@ class Page {
         // is resolved with the complete operator list for a single annotation.
         const opListPromises = [];
         for (const annotation of annotations) {
-          if (
-            isAnnotationRenderable(annotation, intent) &&
-            !annotation.isHidden(annotationStorage)
-          ) {
+          if (isAnnotationRenderable(annotation, annotationStorage, intent)) {
```
Snuffleupagus, line 379: *Sorry about the back-and-forth here, but all of this looks much more complex than it first seems (with lots of special-cases to consider).*

Previously we did the `isHidden` check *regardless* of the `intent`-value, whereas now it only applies to "display" as far as I can tell, is that intended?

---

I almost wonder if, at this point in time, we should remove the old `isAnnotationRenderable`-helper and simply spell out *exactly* what we want at each call-site instead!?

calixteman: yeah it's intended: a field can be not on screen but printed (from js specs: the link I used to use is dead now)
```
Field is visible on screen and in print               display.visible
Field is hidden on screen and in print              display.hidden
Field is visible on screen but does not print     display.noPrint
Field is hidden on screen but prints                 display.noView
```
And it was likely a bug to not print an hidden field with a flag print.
And yes I think it'd be better to remove `isAnnotationRenderable` helper.

Snuffleupagus: > (from js specs: the link I used to use is dead now)

It seems that Adobe recently broke a whole bunch of links, but fortunately the Internet Archive is helpful as usual :-)

The following might be helpful, for future linking purposes, assuming it's the document you're thinking of: https://web.archive.org/web/20201112021418/https://www.adobe.com/content/dam/acom/en/devnet/acrobat/pdfs/AcrobatDC_js_api_reference.pdf

src/display/annotation_layer.js
```diff
@@ -725,6 +725,17 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
               // tooltip
               event.target.title = detail.userName;
             },
+            display() {
```
brendandahl, line 728: It looks like display and print are repeated, can they live on WidgetAnnoationElement or are they not valid for all annotation types? Alternatively, would it make sense to add another class for form widgets to extend from?

brendandahl (changes requested):

src/core/annotation.js
```diff
@@ -439,13 +439,22 @@ class Annotation {
...
+  isPrintable(annotationStorage) {
+    const storageEntry =
+      annotationStorage && annotationStorage.get(this.data.id);
+    if (storageEntry && storageEntry.print !== undefined) {
+      return storageEntry.print;
+    }
+    return this.printable;
+  }
+
+  isDisplayable(annotationStorage) {
```
timvandermeij, line 451: I would call this `isViewable` to match the existing naming scheme.

src/core/annotation.js
```diff
@@ -439,13 +439,22 @@ class Annotation {
     );
   }
 
-  isHidden(annotationStorage) {
+  isPrintable(annotationStorage) {
```
timvandermeij, line 442: It's a bit confusing now that we have both a public and a private method with this name. If there is no better name, can we maybe add some explanation on how they are different?

src/core/annotation.js
```diff
@@ -439,13 +439,40 @@ class Annotation {
     );
   }
 
-  isHidden(annotationStorage) {
+  /**
+   * Check if the annotation must be printed in taking into account
```
Snuffleupagus, line 443: Nit: Here, and in the comment below, let's replace the word `in` with `by` to make this sentence flow a little easier:
```javascript
/**
 * Check if the annotation must be printed by taking into account the value
 * found in the annotationStorage which may have been set through JS.
 * 
```

src/core/annotation.js
```diff
@@ -439,13 +439,40 @@ class Annotation {
...
+   * the value found in the annotationStorage which may have been set
+   * through JS.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {AnnotationStorage} [annotationStorage] - Storage for annotation
+   */
+  mustBePrinted(annotationStorage) {
```
Snuffleupagus, line 451: Nit: In the existing code, all "view" methods/getters always occur *before* the "print" ones.
Can we please re-order these two new methods similarity, such that `mustBeViewed` is place before `mustBePrinted` for consistency?

PR #13332, opened 2021-05-04
fix: set the request to be asynchronous explicitly

z-zodiac:
Explicitly set the XMLHttpRequest request to be asynchronous

### DIFF
--- a/src/display/network.js
+++ b/src/display/network.js
@@ -81,7 +81,8 @@ class NetworkManager {
       xhr,
     });
 
-    xhr.open("GET", this.url);
+    // The third parameter is required for avoiding side-effect
+    xhr.open("GET", this.url, true);
     xhr.withCredentials = this.withCredentials;
     for (const property in this.httpHeaders) {
       const value = this.httpHeaders[property];


Snuffleupagus (changes requested):
Why is this change necessary, since you've not provided *any* explanation/motivation as to exactly *why* the PDF.js library should make this change!?

Note that https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open#parameters clearly lists the parameter as *optional and defaulting to* `true` if not provided.

Furthermore, note that by default the Fetch API is being used in the PDF.js library and this code is thus usually not even invoked.

All-in-all, without a lot more context I don't think this is a change that we want/need to make here!

isGuard:
This problem appeared when I tried vue.pdf
https://github.com/FranckFreiburger/vue-pdf/blob/master/src/pdfjsWrapper.js
In line 35 of this file
var loadingTask = PDFJS.getDocument(source); //source = {url:'blob:http://localhost/...'}
After the request, Chrome will show a warning of `Failed to set the'responseType' property on'XMLHttpRequest': The response type cannot be changed for synchronous requests made from a document`, causing the PDF file to not be displayed

isGuard:
> 这是为什么改变必要的，因为你没有提供_任何_解释/动机，究竟_为什么_在PDF.js库应该让这种变化！？
> 
> 请注意，[https：//developer.mozilla.org/zh-CN/docs/Web/API/XMLHttpRequest/open#parameters](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open#parameters)明确列出了该参数为_可选_ 参数_，_`true`如果未提供则_默认_为该参数。
> 
> 此外，请注意，默认情况下，PDF.js库中使用的是Fetch API，因此通常甚至不会调用此代码。
> 
> 总之，在没有更多上下文的情况下，我认为这不是我们希望/需要在此处进行的更改！

isGuard:
This problem appeared when I tried vue.pdf
https://github.com/FranckFreiburger/vue-pdf/blob/master/src/pdfjsWrapper.js
In line 35 of this file
var loadingTask = PDFJS.getDocument(source); //source = {url:'blob:http://localhost/...'}
After the request, Chrome will show a warning of `Failed to set the'responseType' property on'XMLHttpRequest': The response type cannot be changed for synchronous requests made from a document`, causing the PDF file to not be displayed

Snuffleupagus:
> This problem appeared when I tried vue.pdf
> https://github.com/FranckFreiburger/vue-pdf/blob/master/src/pdfjsWrapper.js

That's a different project, which we cannot support here and according to https://github.com/FranckFreiburger/vue-pdf/blob/e5d4231f21481d2b11144a007c7c9de056eba78d/package.json#L27 it uses a now outdated/unsupported version of the PDF.js library.

> After the request, Chrome will show a warning of `Failed to set the'responseType' property on'XMLHttpRequest': The response type cannot be changed for synchronous requests made from a document`, causing the PDF file to not be displayed

Given that the *default value* should be `true`, that'd possibly even suggest a bug in Google Chrome rather than here. Which Google Chrome version are you actually using?

---

Unless you can reproduce this issue when using the PDF.js library *directly* (and in an up-to-date version of Google Chrome), without another project on top of it, we unfortunately cannot consider a patch such as this.

isGuard:
Ok
vue.pdf uses this library file, so I want to try
Google Chrome
Version 90.0.4430.93 (official version) (64-bit)
This problem does not appear during local testing
Will appear after packaging

Issue #13051, opened 2021-03-04
File renders white while looks fine on Chrome


torakiki:
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version: Firefox 86.0 (64 bit)
- Operating system and its version: Linux Mint 20.1
- PDF.js version: 2.7.510

Steps to reproduce the problem:
This file renders completely white while I can see it using Chrome or Xreader
[njp.32101064480005_page_013(2).pdf](https://github.com/mozilla/pdf.js/files/6083044/njp.32101064480005_page_013.2.pdf)

What is the expected behavior? 
![immagine](https://user-images.githubusercontent.com/3884479/109954440-a143e800-7ce1-11eb-85b4-6ed83cb84021.png)

Snuffleupagus:
```
PDF d0b251002ea92fc25f6c12d7709c33e8 [1.5 HathiTrust Image Server / PDF::API2 / HathiTrust] (PDF.js: 2.8.126) viewer.js:1758:13
Warning: Unsupported header type 1970628964 (uuid) pdf.worker.js:1052:13
Warning: JPX: Unsupported COD options (resetContextProbabilities, terminationOnEachCodingPass, verticallyStripe, predictableTermination). pdf.worker.js:1052:13
Warning: Unable to decode image "img_p0_1": "JpxError: JPX error: JPX error: Out of packets". pdf.worker.js:1052:13
Warning: Dependent image isn't ready yet pdf.js:922:13
```

PR #13337, opened 2021-05-05
[Regression] Move the `super`-call in the `PredictorStream`-constructor to prevent errors (PR 13303)

Snuffleupagus:
*My apologies for breaking this; thankfully PR #13303 hasn't reach mozilla-central yet.*

It's (obviously) necessary to initialize a `PredictorStream`-instance fully, since otherwise breakage may occur if there's errors during the actual stream parsing.
To reproduce this issue, try opening the PDF document from issue #13051 locally and observe the following message in the console:
```
Warning: Invalid stream: "ReferenceError: this hasn't been initialised - super() hasn't been called"
```

### DIFF
--- a/src/core/predictor_stream.js
+++ b/src/core/predictor_stream.js
@@ -19,10 +19,12 @@ import { isDict } from "./primitives.js";
 
 class PredictorStream extends DecodeStream {
   constructor(str, maybeLength, params) {
+    super(maybeLength);
+
     if (!isDict(params)) {
       return str; // no prediction
     }
-    const predictor = params.get("Predictor") || 1;
+    const predictor = (this.predictor = params.get("Predictor") || 1);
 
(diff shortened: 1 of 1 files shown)


calixteman:
Is it possible to add a test using a `PredictorSteam` ?

Snuffleupagus:
It actually turns out that I cannot reproduce this in any *build*, but only locally in `gulp server` mode which suggests that this is some kind of SystemJS/Babel related problem.
Hence this is a) a lot less severe than I initially thought, and b) not really possible/necessary to test.

---

Note that the intention is to get rid of SystemJS for local development, see PR #12563, however that's unfortunately still blocked by https://bugzilla.mozilla.org/show_bug.cgi?id=1247687

calixteman (approved):
Thanks.

PR #13339, opened 2021-05-05
Remove unnecessary closure in `src/display/text_layer.js`, and use standard classes

Snuffleupagus:
With modern JavaScript modules, where you explicitly list the properties that should be exported, it's no longer necessary to wrap all of the code in a closure.[1]

This patch also tries to clean-up/improve a couple of the existing JSDoc-comments.

---
[1] This reduces the size, even of the *built* `pdf.js` file, since there's now a lot less unnecessary whitespace.

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -42,536 +42,516 @@ import {
  *   selection enhancement.
  */
 
-/**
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:

PR #13341, opened 2021-05-05
Remove unnecessary closure in the `src/core/font_renderer.js` file

Snuffleupagus:
With modern JavaScript modules, where you explicitly list the properties that should be exported, it's no longer necessary to wrap *all* of the code within one file into a top-level closure.[1]

This patch reduces the size, of even the *built* `pdf.worker.js` file, since there's now a lot less unnecessary whitespace.

---
[1] For files which contain *different* functionality, some closures may however still make sense in order to separate the code.
It might be possible to remove some of those cases later, e.g. once private class fields becomes generally available/usable in browsers.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13344, opened 2021-05-06
Enable the `no-var` rule in the `src/core/evaluator.js` file

Snuffleupagus:
The only *slight* complication here were some of the `switch`-cases, in `getOperatorList`/`getTextContent`, where the parsing is done asynchronously.
However, those cases are easy to deal with by wrapping the code within its own block; please see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#block-scope_variables_within_switch_statements

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -12,7 +12,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* eslint-disable no-var */
 
 import {
   AbortException,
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
This wasn't as difficult as I previously thought, probably thanks to more experience than last time I looked at this :-)
(With this PR, and my pending one for the image-decoders, I think that I'm done with `no-var` enabling for now.)

/botio test

timvandermeij (approved):

timvandermeij:
Thanks!

PR #13347, opened 2021-05-07
Take the `W` array into account when computing the hash, in `PartialEvaluator.preEvaluateFont`, for composite fonts (issue 13343)

### LINKED ISSUE #13343: The text is cut off on Japanese documents.
[jpa2021506744.pdf](https://github.com/mozilla/pdf.js/files/6431655/jpa2021506744.pdf)

Configuration:
- Web browser and its version: Google Chrome v89.0.4389.114 /IE ver 20H2 (OS build 19042.928)/Microsoft Edge 89.0.774.77 (64 bit)
- Operating system and its version: Windows 10 Pro 20H2 19042.928
- PDF.js version: 2.7.570
- Is a browser extension: No

As you can see from the attached jpa2021506744 [jpa2021506744.pdf](https://github.com/mozilla/pdf.js/files/6431655/jpa2021506744.pdf)  page 15,
When using PDF.js, the  text is cut off on Japanese documents.
What went wrong? I hope to solve this problem.

I tried to improve after seeing these pages.

https://github.com/mozilla/pdf.js/issues/12110
https://github.com/mozilla/pdf.js/pull/13067

Adobe Reader does not have this problem.

jpa_2021_506744_OK

jpa_2021_506744_NG

Snuffleupagus:
Without this some *composite* fonts may incorrectly end up with matching `hash`es, thus breaking rendering since we'll not actually try to load/parse some of the fonts.

*Please note:* Given that the document, in the referenced issue, doesn't embed *any* of its fonts there's no guarantee that it renders correctly in all configurations even with this patch.

Fixes #13343 (to the extent that doing so is possible, since the document uses non-embedded and non-standard fonts).


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

/botio makeref

PR #7029, opened 2016-02-25
[PDFThumbnailView] Re-factor the `canvas` to `image` conversion such that we always render to a `canvas`, and then replace it with an `image` once rendering is done

Snuffleupagus:
_This is a follow-up to PRs #6299 and #6441._

The patch also adds an option to `PDFThumbnailView`, that disables the canvas-to-image conversion entirely, which might be useful in the context of issue #7026.

PR #13357, opened 2021-05-09
Remove the `disableCanvasToImageConversion` functionality from `PDFThumbnailView` (PR 7029 follow-up)

Snuffleupagus:
This functionality was originally implemented in PR #7029; however it's not, nor has it ever been, used as far as I can tell.[1]
Note in particular that the default viewer does not expose either a preference or even an option with which `disableCanvasToImageConversion` can be toggled, and source-code modification is thus required.

Furthermore, note also that we have multiple other instances of `canvas`-data accesses in both the `src/display/canvas.js` and `src/display/text_layer.js` files. If any of those are blocked, by e.g. browser settings, there will be outright rendering bugs and non-working thumbnails thus seem like a very small issue in the grand scheme of things; hence why I'm suggesting that we remove the unused `disableCanvasToImageConversion` functionality.

---
[1] For the Tor use-case mentioned in issue #7026, I *believe* that the solution was to white-list `canvas`-data accesses for its built-in PDF Viewer.

### DIFF
--- a/web/pdf_thumbnail_view.js
+++ b/web/pdf_thumbnail_view.js
@@ -33,9 +33,6 @@ const THUMBNAIL_WIDTH = 98; // px
  * @property {IPDFLinkService} linkService - The navigation/linking service.
  * @property {PDFRenderingQueue} renderingQueue - The rendering queue object.
  * @property {function} checkSetImageDisabled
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks like a good plan to me. Thanks!

PR #13358, opened 2021-05-09
[api-minor] Remove the WebGL implementation

### LINKED ISSUE #5161: Enable WebGL by default?
I was wondering if WebGL is considered stable enough to be enabled by default. It makes some PDFs render much faster and it has been off by default for a while now, so we have had quite some time to test it.

If we decide to enable WebGL by default, we should not forget to update https://github.com/mozilla/pdf.js/blob/master/extensions/chromium/preferences_schema.json#L39 as well.

/cc @yurydelendik 
/cc @Snuffleupagus (because IIRC you were testing WebGL too)

Snuffleupagus:
Reasons for the removal include:
 - This functionality was always somewhat experimental and has never been enabled by default, partly because of worries about rendering bugs caused by e.g. bad/outdated graphics drivers.

 - After the initial implementation, in PR #4286 (back in 2014), no additional functionality has been added to the WebGL implementation.

 - The vast majority of all documents do not benefit from WebGL rendering, since only a couple of *specific* features are supported (e.g. some Soft Masks and Patterns).

 - There is, and has always been, *zero* test-coverage for the WebGL implementation.

 - Overall performance, in the PDF.js library, has improved since the experimental WebGL implementation was added.

Rather than shipping unused *and* untested code, it seems reasonable to simply remove the WebGL implementation for now; thanks to version control it's always possible to bring back the code should the need ever arise.

Fixes #5161

---

*Edit:* A new version was pushed to fix typos in the commit message, no actual code changes were made.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13359, opened 2021-05-10
Add the possibility to rescale each glyph in a font

calixteman:
- a lot of xfa files are using Myriad pro or Arial fonts without embedding them and some containers have some dimensions based on those font metrics. So not having the exact same font leads to a wrong display.
  - since it's pretty hard to find a replacement font with the exact same metrics, this patch gives the possibility to read glyf table, rescale each glyph and then write a new table.
  - so once PR #12726 is merged we could rescale for example Helvetica to replace Myriad Pro.

### DIFF
--- a/src/core/glyf.js
+++ b/src/core/glyf.js
@@ -1,0 +1,708 @@
+/* Copyright 2021 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const ON_CURVE_POINT = 1 << 0;
+const X_SHORT_VECTOR = 1 << 1;
+const Y_SHORT_VECTOR = 1 << 2;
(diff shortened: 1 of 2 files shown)


src/core/glyf.js
```diff
@@ -0,0 +1,714 @@
...
+const WE_HAVE_AN_X_AND_Y_SCALE = 1 << 6;
+const WE_HAVE_A_TWO_BY_TWO = 1 << 7;
+const WE_HAVE_INSTRUCTIONS = 1 << 8;
+// const USE_MY_METRICS = 1 << 9;
+// const OVERLAP_COMPOUND = 1 << 10;
+// const SCALED_COMPONENT_OFFSET = 1 << 11;
+// const UNSCALED_COMPONENT_OFFSET = 1 << 12;
+
+/*
+  GlypfTable object represents a glyf table containing glyph information:
```
Snuffleupagus, line 38: Typo: `GlypfTable` -> `GlyfTable`.

Also, please format these block-comments in the same way as elsewhere in the code-base; note e.g. https://github.com/mozilla/pdf.js/blob/f07d50f8eeced1cd9cc967e938485893d584fc32/src/display/api.js#L500-L504

src/core/glyf.js
```diff
@@ -0,0 +1,714 @@
...
+
+  A full description of glyf table can be found here
+  https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6glyf.html
+*/
+class GlyfTable {
+  constructor(glyphs) {
+    this.glyphs = glyphs;
+  }
+
+  static parse({ glyfTable, isGlyphLocationsLong, locaTable, numGlyphs }) {
```
Snuffleupagus, line 55: Can't you simply let the `GlyfTable` constructor do this, and remove the static method, instead?

Snuffleupagus (changes requested):

Issue #11473, opened 2020-01-04
Wrong rendering of pattern


MichaelEischer:
[pattern.pdf](https://github.com/mozilla/pdf.js/files/4022310/pattern.pdf)

Configuration:
- Firefox Nightly 73.0a1 (2020-01-04)
- macOS Mojave 10.14.6
- PDF.js version: 2.4.254
- Is a browser extension: built-in

Steps to reproduce the problem:
1. Open the PDF, zoom in to 200% for a more detailed view

What is the expected behavior?
<img width="736" alt="pattern-acrobat" src="https://user-images.githubusercontent.com/9106997/71770714-34dc2600-2f30-11ea-958d-4fea6f2b53c2.png">

What went wrong?
<img width="563" alt="pattern-pdfjs" src="https://user-images.githubusercontent.com/9106997/71770723-4a515000-2f30-11ea-8a92-cef4cac2a8f8.png">

The crosshatch pattern is rendered by pdf.js using the wrong pattern style. The other pattern style just seem to have a broken redering.

TeX source of PDF:
```tex
\documentclass[a4paper,11pt]{scrartcl}
\usepackage{tikz}
\usetikzlibrary{patterns}

\begin{document}
crosshatch pattern:
\begin{tikzpicture}
\draw[draw=black, fill=black!10, pattern=crosshatch, pattern color=black] (0,0) rectangle (1,1);
\end{tikzpicture}
\\

north east lines pattern:
\begin{tikzpicture}
\draw[draw=black, fill=black!10, pattern=north east lines, pattern color=black] (0,0) rectangle (1,1);
\end{tikzpicture}
\\

north west lines pattern:
\begin{tikzpicture}
\draw[draw=black, fill=black!10, pattern=north west lines, pattern color=black] (0,0) rectangle (1,1);
\end{tikzpicture}
\\

grid pattern:
\begin{tikzpicture}
\draw[draw=black, fill=black!10, pattern=grid, pattern color=black] (0,0) rectangle (1,1);
\end{tikzpicture}
\end{document}
```

SamyCookie:
Since #11473, there is some tiny glitches on lines, should we reopen this ticket until rendering is perfectly effected ?

Issue #10955, opened 2019-07-08
Pattern/Shading transformation matrix issues


brendandahl:
Currently, when we paint patterns/shadings we apply a base transform then apply the pattern's matrix. This isn't technically correct, instead we should leave the current transform applied and instead use pattern.setTransform. Something like:

```
matrix =  ctx.mozCurrentTransformInverse;
if (!svgElement) {
  svgElement = document.createElementNS("http://www.w3.org/2000/svg", "svg");
}
let mtx = svgElement.createSVGMatrix();
mtx.a = matrix[0];
mtx.b = matrix[1];
mtx.c = matrix[2];
mtx.d = matrix[3];
mtx.e = matrix[4];
mtx.f = matrix[5];
mtx = mtx.translate(clipped.offsetX, clipped.offsetY);
pattern.setTransform(mtx); 
```

brendandahl:
A good PDF for testing this:
https://github.com/mozilla/pdf.js/issues/6769#issuecomment-165476302
And also another PDF that is the same as above but with a pattern matrix of [2, 0, 0, 2, 0, 0] that illuminates some other issues as well:
[issue6769_no_matrix.pdf](https://github.com/mozilla/pdf.js/files/3370236/issue6769_no_matrix.pdf)

Issue #11597, opened 2020-02-14
PDF Render Issue in 2.3.200


blackmambahk:
Attach (recommended) or Link to PDF file here:
[CLIENT_TEST.pdf](https://github.com/mozilla/pdf.js/files/4204251/CLIENT_TEST.pdf)

Configuration:
- Web browser and its version:Chrome 79.0.3945.130, Firefox 72.0.2
- Operating system and its version:Windows 10
- PDF.js version: 2.3.200
- Is a browser extension: no

Steps to reproduce the problem:
1. Open the attached PDF, 
2. page 2 bottom left corner there is a callout with text "UV Blocking"
3. Issue relates to the width of the border and the gradient color.

What is the expected behavior? (add screenshot)
In Acrobat the border is thin and the gradient fades from the center to left and right
In Firefox built in pdfjs the border is a little bit thicker and the gradient fades left to right
In our own app using 2.2.228 the border render the same as with FireFox builtin. 
In our own app using 2.3.200 the border is very thick and fades left to right

What went wrong? (add screenshot)
PDF rendered differently between 2.3.200 and 2.2.228, border is double thickness

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

janpe2:
Maybe a duplicate of #11018. A shading pattern is used as the "color" when a path is stroked.

Issue #11018, opened 2019-07-29
Gradients do not render properly sometimes


izolenta:
[lines_issue.pdf](https://github.com/mozilla/pdf.js/files/3441553/lines_issue.pdf)

Configuration:
- Chrome (latest)
- Mac Os X Mojac
- PDF.js version: latest (reproduced on pdf.js demo page)

Steps to reproduce the problem:
1. Open attached pdf in pdf.js demo viewer
2. Please note that thin line in the middle is not affected by gradient, but it should be (see screenshot below taken from Mac Os X Preview)

What is the expected behavior? (add screenshot)

![image](https://user-images.githubusercontent.com/430887/62041871-5d375a00-b205-11e9-9811-97c1643ae5b9.png)

What went wrong? (add screenshot)

![image](https://user-images.githubusercontent.com/430887/62041851-51e42e80-b205-11e9-97a3-5a1ef1b20245.png)

MitchDizzle:
I believe I'm seeing this issue also in Firefox. Tried to look through the code and didn't find anything regarding to gradients/shading/stroke. The lines in my document are using gradients on the stroke, and fill is set to none. 
Renders properly in Microsoft Edge also.

PR #13361, opened 2021-05-11
Fix several issues with radial/axial shadings and tiling patterns.

### LINKED ISSUE #13325: Graphics problems on PDF in web viewer

Configuration:
- Web browser and its version: Chrome 90.0.4430.93
- Operating system and its version: macOs Big Sur 11.2.3
- PDF.js version: 2.7.570
- Is a browser extension: No

1. Load the attached pdf in https://mozilla.github.io/pdf.js/web/viewer.html
2. Go to page 46. Dotted lines are missing in the graphics (see screenshot from acrobat reader)

What is the expected behavior? (add screenshot) See the screenshot
[Technical_Guide_UK.pdf](https://github.com/mozilla/pdf.js/files/6413001/Technical_Guide_UK.pdf)
Screenshot 2021-05-03 at 07 46 09

brendandahl:
Previously, we set the base transformation and pattern matrix
directly to the main rendering ctx of the page, however doing this
caused the current transform to be lost. This would cause issues
with things like shear missing so the pattern was misaligned or when
stroke was used the scale of the line width or dash would be wrong.
Instead we should leave the current transform and use setTransfrom
on the pattern so it is applied correctly. For axial and radial shadings I had
to create a temporary canvas to draw the shading so I could in turn
use setTransform.

Fixes: #13325, #6769, #7847, #11018, #11597, #11473

The following already in the corpus are improved:
issue8078-page1
issue1877-page1


src/display/pattern_helper.js
```diff
@@ -37,21 +37,57 @@ ShadingIRs.RadialAxial = {
...
           const c = colorStops[i];
           grad.addColorStop(c[0], c[1]);
         }
-        return grad;
+
+        tmpCtx.fillStyle = grad;
+        tmpCtx.fill();
+
+        const pattern = ctx.createPattern(tmpCanvas.canvas, "repeat");
+        pattern.setTransform(new DOMMatrix(ctx.mozCurrentTransformInverse));
```
Snuffleupagus, line 89: I *really* don't want to complicate this patch in any way (especially for the `MOZCENTRAL` build), given how much it improves the pattern handling, but are we worried about people using the general PDF.js library in older Firefox versions?

---

According to https://developer.mozilla.org/en-US/docs/Web/API/CanvasPattern/setTransform#browser_compatibility, using a `DOMMatrix` as the argument has only been supported since Firefox 79 (the current ESR is still on the 78 branch).

Snuffleupagus (approved):
Looks great to me, thank you!

brendandahl:
Now also fixes #11473.

Snuffleupagus (approved):

Snuffleupagus:
Fixes #10955 as well, right?

timvandermeij (approved):

timvandermeij:
Thanks a lot for fixing this!

/botio makeref

PR #13363, opened 2021-05-11
Prevent LGTM from complaining about useless assignments (PR 12562 follow-up)

Snuffleupagus:
Given that we're using ESLint, which is fine with the code as-is, let's just silence the warnings; this is similar to PR #12562.


timvandermeij:
LGTM ;-)

PR #13364, opened 2021-05-11
Clean-up usage of the `TESTING`-define in `src/pdf.sandbox.js`

Snuffleupagus:
This patch moves the `PDFJSDev`-checks *inline*, similar to the rest of the code-base, such that the code in question is actually being removed from the *built* files in e.g. the official releases.


calixteman (approved):
LGTM. Thanks.

PR #13365, opened 2021-05-11
Delete LICENSE

meto9946:


timvandermeij:
The license is important and needs to be provided.

PR #13366, opened 2021-05-12
Convert the remaining functions in `src/core/primitives.js` to use standard classes

Snuffleupagus:
This patch was tested using the PDF file from issue #2618, i.e. https://bug570667.bugzilla-attachments.gnome.org/attachment.cgi?id=226471, with the following manifest file:
```
[
    {  "id": "issue2618",
       "md5": "",
       "rounds": 50,
       "type": "eq"
    }
]
```

```
browser | stat         | Count | Baseline(ms) | Current(ms) | +/- |   %  | Result(P<.05)
------- | ------------ | ----- | ------------ | ----------- | --- | ---- | -------------
firefox | Overall      |    50 |         3417 |        3426 |   9 | 0.27 |
firefox | Page Request |    50 |            1 |           1 |   0 | 5.41 |
firefox | Rendering    |    50 |         3416 |        3426 |   9 | 0.27 |
```

Based on these results, there's no significant performance regression from using standard classes and this patch should thus be OK.

### DIFF
--- a/src/core/primitives.js
+++ b/src/core/primitives.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-import { assert, unreachable } from "../shared/util.js";
+import { assert, shadow, unreachable } from "../shared/util.js";
 import { BaseStream } from "./base_stream.js";
 
 const EOF = {};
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks much nicer!

PR #13368, opened 2021-05-12
Fix the `Jbig2Image` export for the `gulp image_decoders` build (PR 9729 follow-up, issue 13367)

### LINKED ISSUE #13367: JBIG2 image decoder does not work at all
I am using JPG, JPX and JBIG2 decoders from here: https://github.com/mozilla/pdfjs-dist/tree/master/image_decoders

The JBIG2 decoder does not work at all. I think it is because of the typo that you made: you wrote "Jbig2mage" instead of "Jbig2Image".

https://github.com/mozilla/pdfjs-dist/blob/master/image_decoders/pdf.image_decoders.js#L8603

I found a demo here: https://github.com/mozilla/pdf.js/tree/master/examples/image_decoders , but it parses only a JPG image. Could you add a demo code for parsing JBIG2 and JPX?

Snuffleupagus:
Fixes #13367

### DIFF
--- a/src/pdf.image_decoders.js
+++ b/src/pdf.image_decoders.js
@@ -14,7 +14,7 @@
  */
 
 import { getVerbosityLevel, setVerbosityLevel } from "./shared/util.js";
-import { Jbig2mage } from "./core/jbig2.js";
+import { Jbig2Image } from "./core/jbig2.js";
 import { JpegImage } from "./core/jpg.js";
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Whoops, that was a very subtle typo. Thank you for fixing this!

Issue #9627, opened 2018-04-03
contents display incorrectly.


ntanh005:
https://github.com/ntanh005/ntanhGit1990/blob/master/untitled.pdf
Configuration:
- Web browser and its version: IE 11
- Window 10
- PDF.js version: newest
- Is a browser extension: No

Steps to reproduce the problem:
1.  open the pdf file attached above

What is the expected behavior? (add screenshot)
- The text **"helvetica 14 Point"** is positioned correctly.

What went wrong? (add screenshot)
- The text **"helvetica 14 Point"** is positioned incorrectly.

timvandermeij:
The text in green is mirrored and in the wrong position. Moreover, it misses the black pattern in PDF.js.

Snuffleupagus:
Fixed by PR #13361

THausherr:
The original file is fixed, but not the cinnebar file (hard to see), gs-bugzilla690297.pdf and pattyp2.pdf from my own comment. I can of course open a separate issue but I believe it's all related.

Snuffleupagus:
> The original file is fixed, but not the cinnebar file (hard to see), gs-bugzilla690297.pdf and pattyp2.pdf from my own comment. I can of course open a separate issue but I believe it's all related.

Please note that as mentioned in https://github.com/mozilla/pdf.js/blob/master/.github/CONTRIBUTING.md, there should only be *one* issue reported by bug since otherwise tracking can become a bit of a nightmare :-)

It's always possible to close issues as duplicates of each other, if that turns out to actually be the case, but have (subtly) different bugs reported in the same issue means that it could in the general case become impossible to ever close some issues (e.g. if users kept adding new files to an old issue).

timvandermeij:
Closing this one; please open a new issue if there is still a problem with a file so we can track that separately. Thanks!

PR #13371, opened 2021-05-13
Revert "Fix the remaining `no-var` failures, which couldn't be handled automatically, in the `src/core/evaluator.js` file" (PR 13344 follow-up)

Snuffleupagus:
This reverts commit 0ef9b5aafc88094f19fec793c174c622e7e15542, since it cases a lot of warnings (see below) *locally* with e.g. the document from issue #9627.
Strangely enough, this only occurs with `gulp server`-mode and the actual builds are apparently fine. It seems that this *may* be some unfortunate interaction with the old Babel-plugin that's used together with SystemJS.

```
Warning: getTextContent - ignoring ExtGState: "FormatError: ExtGState should be a dictionary.".
```

Rather than taking the risk that this could actually cover a more serious bug, and since I cannot immediately figure out what's wrong, it thus seem safest to revert this for now and we can (carefully) revisit this once SystemJS has been removed (see PR #12563).

*Reverts part of PR #13344.*

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -12,6 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* eslint-disable no-var */
 
 import {
   AbortException,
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Good call. It also looks like this should just work; very strange indeed.

PR #13375, opened 2021-05-14
Improve the `Page.content` and `Page.getContentStream` methods

Snuffleupagus:
First of all, by using `Dict.getArray` in the `Page.content` getter we remove the need to manually iterate through and fetch the sub-streams (when they exist) in the `Page.getContentStream` method.
Secondly, we can simplify the code in `Page.{getOperatorList, extractTextContent}` by letting `Page.getContentStream` ensure that `content` is available and returning a Promise instead.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13376, opened 2021-05-14
Replace command with not enough args by an endchar in CFF font

### LINKED ISSUE #6132: Many characters in this PDF are not drawn
[`http://www.ei.ruhr-uni-bochum.de/media/crypto/veroeffentlichungen/2010/08/08/baileypaaroefjc.pdf`](http://www.ei.ruhr-uni-bochum.de/media/crypto/veroeffentlichungen/2010/08/08/baileypaaroefjc.pdf) is severely mis-rendered by the version of pdf.js bundled with Firefox 38.0.5:

screen shot of misrendered PDF using pdf.js

You can see that practically every word has some blanks where letters should be.

Note that Adobe Reader (11.0.09) does a _better_ job of rendering but still has many missing characters, and throws up an error message about a font that it "cannot extract", so the PDF itself may be ill-formed (in which case pdf.js should give an error message too).

screen shot of the same PDF with Adobe Reader 11.0.09

calixteman:
- Right now, a glyph with an erroneous outline is replaced by an empty glyph
    if the error is far enough from the start there's likely something to render
    so the idea is to replace a command with args by an endchar when no args are
    on the stack: this way OTS is likely happy (no remaining args on stack) and we
    can draw something which is likely better than nothing.
  - it aims to fix #6132


calixteman:
Interestingly, I disabled charStrings validation and OTS in Firefox complained about wrong charStrings but the pdf was well-rendered and fontForge doesn't complain about those fonts.

PR #13377, opened 2021-05-14
Re-factor and convert the code in `src/core/pattern.js` to use standard classes

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Looks a lot better, thanks!

PR #13378, opened 2021-05-14
Replace terminal null char by a endchar command in CFF charstrings to make OTS happy

### LINKED ISSUE #10544: Weird font misrendering in PDF
Attach (recommended) or Link to PDF file here: [pet.pdf](https://github.com/mozilla/pdf.js/files/2852697/pet.pdf)

Configuration:
- Operating system and its version: Fedora 28
- PDF.js version: tip
- Is a browser extension: testing using `gulp server` and picking the file off the local file system

1. Load the PDF!

screenshot from 2019-02-11 11-31-09

What went wrong?
screenshot from 2019-02-11 11-30-48

@arai-a notes on IRC that there are some possibly potentially likely relevant console warnings:

> PDF 2dbe539840cea24147791d677c5928ba [1.4 Nuance PDF Create 8 / Microsoft Word - Davis petition 10.18.docx] (PDF.js: -) app.js:1084:6
> Warning: Failed to load font "g_d1_f1": SyntaxError: An invalid or illegal string was specified util.js:306:4
> Warning: Failed to load font "g_d1_f3": SyntaxError: An invalid or illegal string was specified util.js:306:4
> Warning: Failed to load font "g_d1_f4": SyntaxError: An invalid or illegal string was specified util.js:306:4

Could be, for all I know, that the PDF itself is not quite up to snuff and all.  Yet evince views it correctly and someone thought the PDF actually worked correctly enough to submit it to the United States Supreme Court with serious legal consequences if it didn't work correctly, so... ¯\\\_(ツ)_/¯

calixteman:
It aims to fix #10544.

### DIFF
--- a/src/core/cff_parser.js
+++ b/src/core/cff_parser.js
@@ -599,6 +599,12 @@ const CFFParser = (function CFFParserClosure() {
         } else if (value === 11) {
           state.stackSize = stackSize;
           return true;
+        } else if (value === 0 && j === data.length) {
+          // Operator 0 is not used according to the current spec and
+          // it's the last char and consequently it's likely a terminator.
+          // So just replace it by endchar command to make OTS happy.
+          data[j - 1] = 14;
+          validationCommand = CharstringValidationData[14];
         } else {
           validationCommand = CharstringValidationData[value];
         }


timvandermeij:
Is it possible to add a reference test for this bug, or is the rendering equal and only the OTS warnings are fixed?

calixteman:
> Is it possible to add a reference test for this bug, or is the rendering equal and only the OTS warnings are fixed?
Yeah rendering is the same but it fixes OTS warnings.
If you've an idea of possible corrupted pdf where rendering could be fixed by this patch, please tell me and I'll add it in test suite.

timvandermeij:
I'm not sure, but if `ttx` complained about it before, a font test might work; see https://github.com/mozilla/pdf.js/tree/master/test/font. If that's also not the case, I indeed don't think this can really be tested automatically.

Snuffleupagus (approved):
Looks reasonable to me, thank you!

PR #13379, opened 2021-05-15
[api-minor] Update minimum supported browser versions (PR 13361 follow-up)

Snuffleupagus:
With the changes in PR #13361, we're now using the `CanvasPattern.setTransform()` method when rendering certain Shadings/Patterns.
Note that while `CanvasPattern` itself has been supported since basically "forever", its `setTransform` method is a slightly newer addition to the specification; please refer to https://developer.mozilla.org/en-US/docs/Web/API/CanvasPattern#browser_compatibility

Rather than trying to re-write PR #13361 to not use, or possibly spending time/effort (if possible) polyfilling, `CanvasPattern.setTransform()` this patch thus suggests that we simply update the *minimum* supported browser versions instead.

According to the compatibility data linked above, the *minimum* supported browser versions in the PDF.js library are now as follows:
 - Chrome >= 68, which was released on 2018-07-24.[1]
 - Firefox ESR, see https://wiki.mozilla.org/Release_Management/Calendar.
 - Safari >= 11.1, which was release on 2018-03-29.[2]

(Given that the PDF.js contributors cannot realistically test a bunch of old browsers, it's not unimaginable that some older browser versions are already not working with the PDF.js library.)

Based on these changes, which we should ensure are reflected in the Wiki as well, we can also remove a number of now redundant polyfills. Furthermore we'll no longer "claim" to support Windows XP, note the `gulpfile.js` changes, which should definitely *not* be an issue given that it's no longer officially supported.[3]

---
[1] According to https://en.wikipedia.org/wiki/Google_Chrome_version_history

[2] According to https://en.wikipedia.org/wiki/Safari_version_history#Safari_11

[3] According to https://en.wikipedia.org/wiki/Windows_XP#End_of_support

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -79,10 +79,9 @@ const config = JSON.parse(fs.readFileSync(CONFIG_FILE).toString());
 const AUTOPREFIXER_CONFIG = {
   overrideBrowserslist: [
     "last 2 versions",
-    "Chrome >= 49", // Last supported on Windows XP
-    "Firefox >= 52", // Last supported on Windows XP
+    "Chrome >= 68",
     "Firefox ESR",
-    "Safari >= 10",
+    "Safari >= 11.1",
     "> 0.5%",
     "not IE > 0",
     "not dead",
@@ -1321,7 +1320,7 @@ gulp.task(
         ),
         preprocessCSS("web/viewer.css", "chrome", defines, true)
           .pipe(
-            postcss([autoprefixer({ overrideBrowserslist: ["chrome >= 49"] })])
+            postcss([autoprefixer({ overrideBrowserslist: ["chrome >= 68"] })])
           )
           .pipe(gulp.dest(CHROME_BUILD_CONTENT_DIR + "web")),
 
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
This looks very reasonable to me. Those versions are significantly old enough to not be supported anymore.

timvandermeij:
The browser support FAQ entry is also updated now.

Snuffleupagus:
> The browser support FAQ entry is also updated now.

Thank you, that looks much better already!

If I may suggest some additional clean-up of the Wiki:
 - Since I don't believe that we know all that much about the compatibility situation of iOS, except that there may be some issues e.g. related to canvas-sizes, I wonder if we should drop the `For mobile only iOS 10 and higher are supported.` sentence?
 - Should we perhaps also remove the `Support was dropped after stable version 2.6.347.` sentence, from the IE/Edge section, since I don't believe that we want to actively steer users towards older releases?

timvandermeij:
Good points! I have made both changes now.

PR #13380, opened 2021-05-15
Re-factor and convert the code in `src/display/pattern_helper.js` to use standard classes 

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Thank you for modernizing this code; looks much better like this!

Issue #7625, opened 2016-09-12
PDF text not correctly rendered


Woodgnome:
[Er æstetik en løftestang for læring.pdf](https://github.com/mozilla/pdf.js/files/467169/Er.aestetik.en.loftestang.for.laering.pdf)

**Configuration:**
- Web browser and its version: Google Chrome 52.0.2743.116
- Operating system and its version: Microsoft Windows 10 v1607
- PDF.js version: 1.5.452

**Steps to reproduce the problem:**
1. Open attached PDF in latest public demo of PDF.js

**What is the expected behavior?**
As rendered in Google Chrome PDF viewer: http://i.imgur.com/LXDR9VD.png

**What went wrong?**
As rendered in PDF.js: http://i.imgur.com/r9FwrN5.png

I should probably add that my Acrobat is not able to correctly render the PDF either, so it may be a case of the PDF document simply being *too* mangled to fix (or maybe a missing font).

Snuffleupagus:
```
PDF 86084efb634c2241847af0c784588e26 [1.6 Adobe PDF Library 9.9 / Adobe InDesign CS5 (7.0)] (PDF.js: 1.5.452)  viewer.js:7984:7
Warning: Not enough parameters for hhcurveto; actual: 0, expected: 4  pdf.worker.js:1263:5
Warning: Not enough parameters for hvcurveto; actual: 3, expected: 4 pdf.worker.js:1263:5
Warning: Not enough parameters for hvcurveto; actual: 2, expected: 4  pdf.worker.js:1263:5
Warning: Not enough parameters for hvcurveto; actual: 3, expected: 4 pdf.worker.js:1263:5
Warning: Not enough parameters for rcurveline; actual: 2, expected: 8  pdf.worker.js:1263:5
Warning: Found too many parameters for stack-clearing command  pdf.worker.js:1263:5
Warning: Not enough parameters for hstem; actual: 1, expected: 2  pdf.worker.js:1263:5
Warning: Not enough parameters for rcurveline; actual: 3, expected: 8  pdf.worker.js:1263:5
Warning: Not enough parameters for hvcurveto; actual: 3, expected: 4 pdf.worker.js:1263:5
Warning: Not enough parameters for vhcurveto; actual: 2, expected: 4  pdf.worker.js:1263:5
Warning: Not enough parameters for hhcurveto; actual: 0, expected: 4  pdf.worker.js:1263:5
[...]

PR #13381, opened 2021-05-15
Handle errors gracefully, in PartialEvaluator.buildFontPaths, when glyph path building fails

Snuffleupagus:
The building of glyph paths, in the `FontRendererFactory`, can fail in various ways for corrupt font data. However, we're currently not attempting to handle any such errors in the evaluator, which means that a single broken glyph *can* prevent an entire page from rendering.

To address this we simply have to pass along, and check, the existing `ignoreErrors` option in `PartialEvaluator.buildFontPaths` similar to the rest of the `PartialEvaluator` code.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -825,7 +825,7 @@ class PartialEvaluator {
               loadedName: "g_font_error",
               font: new ErrorFont(`Type3 font load error: ${reason}`),
               dict: translated.font,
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
Note that I stumbled upon this missing error handling when testing PR #13376 locally, using the PDF document from issue #7625.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #13383, opened 2021-05-16

Snuffleupagus:


Snuffleupagus:
These lint errors are really weird, since `gulp lint`/`gulp lint --fix` both run just fine locally.
If I were to guess, the problem *might* be an outdated Node.js version on the bots since `Object.fromEntries` requires at least Node.js version `12.0.0` [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries#browser_compatibility).

Assuming that GitHub actions start working again, since currently everything is listed as "Queued", we could perhaps move forward with this PR anyway!?

timvandermeij (approved):

timvandermeij:
Thank you for the update!

PR #13385, opened 2021-05-16
Convert `src/core/operator_list.js` to use standard classes

Snuffleupagus:
With modern JavaScript modules, where only *explicitly* exported properties are visible to the outside, the `QueueOptimizerClosure` should no longer be necessary.

Furthermore, to reduce the possibility of `NullOptimizer` and `QueueOptimizer` getting out of sync (note e.g. the inconsistency fixed in PR #10784), we now let the latter extend the former one.

### DIFF
--- a/src/core/operator_list.js
+++ b/src/core/operator_list.js
@@ -13,487 +13,498 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13387, opened 2021-05-16
Fix a few *safe* ESLint no-var failures in `src/core/evaluator.js` (13371 follow-up)

Snuffleupagus:
As can be seen in PR #13371, some of the `no-var` changes in the `PartialEvaluator.{getOperatorList, getTextContent}` methods caused errors in `gulp server`-mode.
However, there's a handful of instances of `var` in other methods which should be completely *safe* to convert since there's no strange scope-issues present in that code.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -394,7 +394,8 @@ class PartialEvaluator {
     } else {
       bbox = null;
     }
-    let optionalContent = null;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13394, opened 2021-05-18
Handle PI with no value in xml parser

calixteman:
- an XML PI contains a target and optionally some content (see https://en.wikipedia.org/wiki/Processing_Instruction)
  - the parser expected to always have some content and so it could lead to wrong parsing.


Snuffleupagus (approved):
Looks good to me, thank you!

PR #13401, opened 2021-05-19
Remove some, with Prettier `2.3.0`, unnecessary `// prettier-ignore` comments

Snuffleupagus:
To get the maximum benefit from something like Prettier, you obviously don't want to disable the automatic formatting unless absolutely necessary. When we added Prettier there were a number of cases, mostly involving larger Arrays, which required disabling of the automatic formatting for overall readability and/or to not break inline comments.

With changes in Prettier version `2.3.0`, see [the release notes](https://prettier.io/blog/2021/05/09/2.3.0.html#concise-formatting-of-number-only-arrays-10106httpsgithubcomprettierprettierpull10106-10160httpsgithubcomprettierprettierpull10160-by-thorn0httpsgithubcomthorn0), there's now better formatting support for Arrays containing only numbers. Hence we can now remove a number of `// prettier-ignore` comments, and thus get the benefit of automatic formatting in (slightly) more of the code-base.


timvandermeij (approved):

timvandermeij:
Thank you for doing this!

Snuffleupagus:
@timvandermeij As always, thank you for landing my patches :-)

---

Somewhat unrelated, but is it too soon to start thinking about creating a new release now?

Historically we've probably, more than once, gone a bit too long between releases at times. Having releases happening around e.g. every 6-8 weeks[1] would mean that users get access to bug fixes and new features much quicker (since many users seem to prefer the *stable* versions).
Also, in recent months the development pace has picked up nicely, and having e.g. fewer new features and overall changes in each release might help reduce (possible) bugs in any one release!?

---
[1] Note how the release cycle of e.g. Firefox is even shorter nowadays, see https://wiki.mozilla.org/Release_Management/Calendar

PR #13402, opened 2021-05-19
XFA - Fix lot of layout issues

calixteman:
- I thought it was possible to rely on browser layout engine to handle layout stuff but it isn't possible
    - mainly because when a contentArea overflows, we must continue to layout in the next contentArea
    - when no more contentArea is available then we must go to the next page...
    - we must handle breakBefore and breakAfter which allows to "break" the layout to go to the next container
  - Sometimes some containers don't provide their dimensions so we must compute them in order to know where to put
    them in their parents but to compute those dimensions we need to layout the container itself...
  - See top of file layout.js for more explanations about layout.
  - fix few bugs in other places I met during my work on layout.

### DIFF
--- a/src/core/xfa/layout.js
+++ b/src/core/xfa/layout.js
@@ -1,0 +1,190 @@
+/* Copyright 2021 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { $extra, $flushHTML } from "./xfa_object.js";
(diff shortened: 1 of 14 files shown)


src/core/xfa/layout.js
```diff
@@ -0,0 +1,188 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { $extra, $flushHTML } from "./xfa_object.js";
+import { measureToString } from "./html_utils.js";
+
+// Subform and ExclGroup have a layout so they share these functions.
+
+/*
```
Snuffleupagus, line 21: Nit: Please remember to use the following comment formatting, for overall consistency in the code-base.
```javascript 
/**
 * How layout works?
 *
 * ...
 */
```

src/core/xfa/template.js
```diff
@@ -1683,29 +1925,154 @@ class ExclGroup extends XFAObject {
...
+    attributes.style = style;
+    attributes.class = clazz.join(" ");
+
+    if (this.name) {
+      attributes.xfaName = this.name;
+    }
+
+    let failure;
+    if (this.layout === "lr-tb" || this.layout === "rl-tb") {
+      for (; this[$extra].attempt < 3; this[$extra].attempt++) {
```
brendandahl, line 2020: Make number of attempts a constant.

brendandahl (changes requested):

brendandahl (approved):
r+ needs the comment in layout.js updated though.

PR #13408, opened 2021-05-20
Remove focus from the zoom dropdown, when a mouse is used (bug 1300525, issue 4923)

### LINKED ISSUE #4923: Issue browsing the document after setting the zoom level using the drop-down menu
After selecting for example the zoom level "page fit", if would be intuitive and desirable to be able to navigate the document using the arrow keys. Right now pressing an arrow key (even and especially left/right) changes the zoom level, although you just selected the desirable zoom level manually.

EDIT: the same thing holds for the page up and page down keys.

Snuffleupagus:
This patch fixes the referenced bugs/issues, in a way that won't interfere with keyboard users, assuming that we actually want to fix these old bugs/issues. (If not, we should close them as WONTFIX.)

Fixes [bug 1300525](https://bugzilla.mozilla.org/show_bug.cgi?id=1300525)
Fixes #4923

### DIFF
--- a/web/toolbar.js
+++ b/web/toolbar.js
@@ -155,6 +155,19 @@ class Toolbar {
         value: this.value,
       });
     });
+    // Here we depend on browsers dispatching the "click" event *after* the
+    // "change" event, when the <select>-element changes.
+    scaleSelect.addEventListener("click", function (evt) {
+      const target = evt.target;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks easy enough to include. Thanks!

PR #13410, opened 2021-05-20
Allow consumers of the text layer builder and render task to specify element type of the text div.

gone-skiing:
**Background** 
Once upon a time text layer used to generate `div` text elements (the variable name is still `textDiv`). It was later changed to `span`. For our use case, it is a breaking change and we have been maintaining our own fork that still generates div(s).

**Summary** 
This pull request adds the ability to override element type for text builder and rendering task constructors. The current behavior of using span is preserved as the default value.

### DIFF
--- a/src/display/text_layer.js
+++ b/src/display/text_layer.js
@@ -40,6 +40,8 @@ import {
  *   text runs occurs.
  * @property {boolean} [enhanceTextSelection] - Whether to turn on the text
  *   selection enhancement.
+ * @property {"span" | "div" } [textDivElementType] - Allows to specify type
+ * of text elements for backwards compatibility. Defaults to span.
  */
 
 const MAX_TEXT_DIVS_TO_RENDER = 100000;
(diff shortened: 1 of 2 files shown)


Snuffleupagus (changes requested):
Sorry, but I *really* don't think this is something that the PDF.js library should support out-of-the-box.

Please note that using `<div>`s instead of `<span>`s will result in *a lot* worse text-selection behaviour, with line-breaks added basically everywhere and undoing the last couple of years worth of improvements in this area.
Hence we really don't want to provide an option which knowingly degrades the text-selection UX, and without additional (CSS-related) changes this option would also break e.g. the viewer quite badly as-is. (If you have to modify the CSS anyway, why not the JS code as well?)

Furthermore, we really don't want to add options that are both unused *and* untested, since doing so adds future maintenance burden and this kind of option isn't the sort of thing that we want to commit to supporting.

All-in-all, this unfortunately seems like the sort of thing that belongs in a *fork* rather than the main PDF.js library.

gone-skiing:
@Snuffleupagus thank you for the quick response. I can definitely see where you are coming from on this, my only source of hope was that it was fairly limited change and it provided backward compatibility for something that had been changed previously. That said it is something that most likely matters to very few library users (we have xpaths stored in test configurations in some cases so div vs span change matters).

PR #13411, opened 2021-05-20
XFA - Add support to print XFA forms

calixteman:


web/base_viewer.js
```diff
@@ -581,21 +581,35 @@ class BaseViewer {
...
                   pageView.setPdfPage(pdfPage);
                 }
+                if (isPureXfa) {
+                  // Force the rendering of all pages (in case the pdf is
+                  // printed). It isn't really an issue since the document
+                  // has been fully layed out in worker in order to just
+                  // get the total page number. For now, in order to print
+                  // a filled form, we print the generated html as the user
+                  // can see it. It could lead to a memory issue in case
+                  // the document as a huge number of pages which is pretty
```
Snuffleupagus, line 609: Nit: as -> has

web/base_viewer.js
```diff
@@ -581,21 +581,35 @@ class BaseViewer {
...
+                  // unlikely since normally XFA is used to render forms.
+                  xfaDrawPromises.push(pageView.draw());
```
Snuffleupagus, line 611: *I'm finding this, as a matter of principle, somewhat annoying/bad to have to do just to enable printing; especially given e.g. the cases mentioned below.*
Is it really not possible to have the print-functionality generate any non-rendered XFA pages, similarly to how it creates normal canvases during printing?

For one thing, this short-circuits the caching of pages in the viewer in ways that will lead to weird bugs, too much memory usage (in a longer document), an inability to clean-up any allocated resourced, and just overall *very* inconsistent state.
Note that an entire document was *not* intended to be rendered at the same time, with the exception of one page documents, and that this breaks a lot of assumptions elsewhere in the viewer code.

Snuffleupagus (changes requested):

web/base_viewer.js
```diff
@@ -612,6 +626,11 @@ class BaseViewer {
               }
             );
           }
+
+          if (isPureXfa) {
+            return Promise.all(xfaDrawPromises);
```
Snuffleupagus, line 631: The return value of this function isn't used for anything, hence I don't think this actually does what you think here. Basically, I'm assuming that you'd want to delay `this._pagesCapability.resolve();` in the XFA-case (however please note the comment regarding the overall implementation.)

calixteman:
@Snuffleupagus, I tried to implement print stuff in `firefox_print_service.js` but I'm kind of stuck.
The only idea I had is to try to fallback on `pdf_print_servive.js` in case of xfa document since I don't render xfa on a canvas: is it a good idea or not ?

Snuffleupagus:
>  I tried to implement print stuff in `firefox_print_service.js` but I'm kind of stuck.

Stuck how, what did/didn't work?

> is it a good idea or not ?

Not really, since that code isn't actually bundled in MOZCENTRAL-builds (and switching between two *separate* implementations at runtime feels messy).

web/pdf_print_service.js
```diff
@@ -27,8 +29,28 @@ function renderPage(
...
+  optionalContentConfigPromise,
+  printContainer
 ) {
+  if (pdfDocument.isPureXfa) {
+    const page = document.createElement("div");
+    page.setAttribute("class", "xfaPrintedPage");
+    printContainer.appendChild(page);
+
+    return pdfDocument.getPage(pageNumber).then(pdfPage => {
+      const builder = new DefaultXfaLayerFactory().createXfaLayerBuilder(
```
Snuffleupagus, line 41: I know that this is still a WIP, but for future reference:
We should probably initialize/re-use just *one* instance of `DefaultXfaLayerFactory` for all pages.

web/pdf_print_service.js
```diff
@@ -27,8 +29,28 @@ function renderPage(
...
+        page,
+        pdfPage,
+        pdfDocument.annotationStorage
+      );
+      const viewport = pdfPage.getViewport({
+        scale: CSS_UNITS,
+        rotation: size.rotation,
+      });
+      return builder.render(viewport, "print").promise;
```
Snuffleupagus, line 50: I know that this is still a WIP, but for future reference:
As far as I can tell, the `intent` parameter is currently being ignored in the `render` method.

calixteman:
> > I tried to implement print stuff in `firefox_print_service.js` but I'm kind of stuck.
> 
> Stuck how, what did/didn't work?
>

I copied/pasted the code I wrote for `pdf_print_service.js` (in `renderPage`) at the beginning of `composePage` but the printed pages are white.
My guess is that the "rendering" is not sync so the printed html doesn't contain the html for each page.
So ideally we should block until all the promises are resolved before returning from `composePage` but afaik it isn't possible.
 
> > is it a good idea or not ?
> 
> Not really, since that code isn't actually bundled in MOZCENTRAL-builds (and switching between two _separate_ implementations at runtime feels messy).

web/ui_utils.js
```diff
@@ -994,6 +997,29 @@ function apiPageModeToSidebarView(mode) {
...
+  const factory = new DefaultXfaLayerFactory();
+  const scale = Math.round(CSS_UNITS * 100) / 100;
+  for (const xfaPage of xfaHtml.children) {
+    const page = document.createElement("div");
+    page.setAttribute("class", "xfaPrintedPage");
+    printContainer.appendChild(page);
+
+    const { width, height } = xfaPage.attributes.style;
+    const viewBox = [0, 0, parseInt(width), parseInt(height)];
+    const viewport = new PageViewport({ viewBox, scale, rotation: 0 });
```
Snuffleupagus, line 1011: Can we pretty please add an API method which returns the necessary `PageViewport` and use that instead here, rather than exposing this class directly in the *public API*?

Unless absolutely necessary, this class isn't the sort of thing that we ideally want to expose publicly since it makes it more difficult to e.g. make any future changes to that code.

web/pdf_print_service.js
```diff
@@ -157,8 +163,12 @@ PDFPrintService.prototype = {
         this._printResolution,
         this._optionalContentConfigPromise
       )
-        .then(this.useRenderedPage.bind(this))
-        .then(function () {
+        .then(() => {
+          if (!this.pdfDocument.isPureXfa) {
```
Snuffleupagus, line 167: Is this check still relevant, since I don't believe that we can get here now given the check at the start of this method?

brendandahl (approved):

PR #13412, opened 2021-05-20
Keep all rendered pages active, in the viewer, for XFA-documents

Snuffleupagus:
*Please note:* This will obviously have a negative impact on e.g. overall performance and memory usage for XFA-documents, by keeping all rendered pages active regardless if they're visible or not.

Hopefully most XFA-documents aren't that large/long, and if we *need* to keep all pages active this patch provides a fairly small/self-contained way of doing that. (Assuming I didn't overlook some edge-case, of course...)

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -517,6 +517,12 @@ class BaseViewer {
     };
     this.eventBus._on("pagerendered", this._onAfterDraw);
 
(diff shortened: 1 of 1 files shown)


calixteman:
I think we should avoid to break made assumptions.
So I'm figuring out a correct way to deal with print stuff for xfa.

Issue #13266, opened 2021-04-19
Javascipt: Submit/Export Form action doesn't work in Firefox


harrysingh1009:
Attach (recommended) or Link to PDF file here: 
[Sample PDF.pdf](https://github.com/mozilla/pdf.js/files/6338434/Sample.PDF.pdf)

Configuration:
- Web browser and its version: Firefox v88
- Operating system and its version: Windows 10 
- PDF.js version: Latest
- Is a browser extension: yes

Steps to reproduce the problem:
1. Open the PDF file attached to this ticket in Firefox.
2. Try clicking on the "Submit" or "Export" buttons at the end of the document.

What is the expected behavior? (add screenshot)

Clicking Submit should submit form data to the URL specified. (in this case, google.com)
Clicking export button should pop up a window to ask user to save the file.

What went wrong? (add screenshot)

Clicking them doesn't perform the respective action of Submit or Export.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

Snuffleupagus:
- https://github.com/mozilla/pdf.js/blob/fd82adccfa409f9b2cc72ad2cb8555bd2aa68b36/src/scripting_api/doc.js#L1161-L1163
- https://github.com/mozilla/pdf.js/blob/fd82adccfa409f9b2cc72ad2cb8555bd2aa68b36/src/scripting_api/doc.js#L770-L772

harrysingh1009:
@Snuffleupagus Thanks for your response. So, does that mean that the functions/APIs "exportAsXFDF" and "submitForm" have not been implemented yet and I cannot use it?
Or there is a way to implement it from my side?

Thanks

marco-c:
@calixteman maybe we should show a message saying something like "This action is not supported"?

calixteman:
For now there are no plan to support `submitForm` function since no function involving network are supported.
@marco-c, why not.

tbruckmaier:
Is this something on the roadmap? Or is it a design choice to not support network functions?

calixteman:
We need to see with security team if there are any concerns to implement functions involving network.
For now it has been decided to not implement them.

dr-death2:
From my point of view implementing network functions **should be a no-go**.

[...]

PR #13417, opened 2021-05-22
[XFA] Send URLs as strings, rather than objects (issue 1773)

Snuffleupagus:
Given that `URL`s aren't supported by the structured clone algorithm, see https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm, the document in issue #1773 will cause the browser to throw `DataCloneError: The object could not be cloned.`-errors and nothing will render.
To fix this, we'll instead simply send the stringified version of the `URL` to prevent these errors from occuring.

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -2257,7 +2257,7 @@ class Image extends StringObject {
     };
 
     if (this.href) {
-      html.attributes.src = new URL(this.href);
+      html.attributes.src = new URL(this.href).href;
       return html;
     }
 
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1800,7 +1800,10 @@ class LoopbackPort {
         }
         return result;
       }
-      result = Array.isArray(value) ? [] : {};
+      if (value instanceof URL) {
+        throw new Error(`LoopbackPort.postMessage - cannot clone: ${value}`);
+      }
+      result = Array.isArray(value) ? [] : Object.create(null);
       cloned.set(value, result); // Adding to cache now for cyclic references.
       // Cloning all value and object properties, however ignoring properties
       // defined via getter.


src/core/xfa/template.js
```diff
@@ -2266,7 +2266,7 @@ class Image extends StringObject {
         c.charCodeAt(0)
       );
```
Snuffleupagus, line 2267: Somewhat unrelated, but I cannot help wonder why we're sort-of "reinventing the wheel" rather than using the existing `stringToBytes` utility function (from `src/shared/util.js`) here?

src/core/xfa/template.js
```diff
@@ -2266,7 +2266,7 @@ class Image extends StringObject {
...
       const blob = new Blob([buffer], { type: this.contentType });
-      html.attributes.src = URL.createObjectURL(blob);
+      html.attributes.src = URL.createObjectURL(blob).href;
```
calixteman, line 2269: It isn't correct.
You can try with this form: https://www.canada.ca/content/dam/ircc/migration/ircc/english/pdf/kits/forms/imm1344e.pdf, there is a picture in top-left corner with a canadian flag.
When dumping `html.attributes.src`, I got an `undefined`.
Anyway the picture is correctly displayed so it likely means that the URL is correctly cloned (or I missed something).
The failing URL in the document you mentioned is a local path `Q:\hrmgmt\New HR NEU Logos\NortheasternUniv-w-Seal-Red.jpg`.

src/core/xfa/template.js
calixteman: Just because the wheel is so small here but you're right I'll replace it.

src/core/xfa/template.js
Snuffleupagus: Sorry, that was a bit too much copy-and-paste which I've fixed now; thank you!

Obviously `URL.createObjectURL` returns a string, but the rest of the patch is necessary to prevent breaking errors.

src/core/xfa/template.js
```diff
@@ -2257,7 +2257,7 @@ class Image extends StringObject {
     };
 
     if (this.href) {
-      html.attributes.src = new URL(this.href);
+      html.attributes.src = new URL(this.href).href;
```
calixteman, line 2260: I think that for now we shouldn't create an image in case of `this.href !== null`:
  - the ref can be a path (a name) to an internal picture which is not in the xml stream and we'll handle this case
  - it can be a real `http://` link or whatever. In this case, what's the policy ? especially if the url has the file protocol.
  - it can be garbage as you saw it in the mentioned pdf.
And I've the same question for `a` tag in `xhtml.js`.

Snuffleupagus: The *main* objective of the PR was simply to avoid this line completely breaking everything, because of structured cloning errors (since a `URL` can't be cloned), and further improvements can probably happen elsewhere?

> * it can be a real `http://` link or whatever. In this case, what's the policy ? especially if the url has the file protocol.

I'm not sure that we should even attempt to support that case at all, given the possibly security implications of loading arbitrary content; this feels a bit similar to e.g. https://github.com/mozilla/pdf.js/issues/13266#issuecomment-824046524.

calixteman (approved):
LGTM

PR #13419, opened 2021-05-22
Use the `stringToBytes` helper function in more places

Snuffleupagus:
Rather than manually reimplementing, more-or-less, this functionality in a few spots we can simply use the existing helper function instead.


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #13420, opened 2021-05-22
Drop obsolete logic from the `downloadFile` function in `test/downloadutils.js`

timvandermeij:
_This patch is a first step to refactor the testing code to make it more manageable. This is needed in order to, over time, enable the `no-var` rule there too, which is now made difficult by lots of global variable and callback usage._

This code is old and predates the improvements we made to the test manifest to only contain working URLs (either Web Archive or GitHub/Bugzilla links), so the fallback logic to try the Web Archive is no longer necessary. This greatly simplifies the function and also makes sure that we fail directly in case a bad URL is added to the manifest, instead of having it work "accidentally" because of this logic, since we want the manifest to be correct at all times (and otherwise fail loudly).

To verify that this works I removed my local `test/pdfs` folder, did a clean checkout of it and ran the tests, which forced a re-download of all files. All files downloaded correctly, there were no MD5 checksum mismatches and the files could also be opened correctly.

### DIFF
--- a/test/downloadutils.js
+++ b/test/downloadutils.js
@@ -38,11 +38,9 @@ function rewriteWebArchiveUrl(url) {
 function downloadFile(file, url, callback, redirects) {
   url = rewriteWebArchiveUrl(url);
 
-  var completed = false;
   var protocol = /^https:\/\//.test(url) ? https : http;
   protocol
     .get(url, function (response) {
-      var redirectTo;
       if (
         response.statusCode === 301 ||
         response.statusCode === 302 ||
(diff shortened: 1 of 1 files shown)


Snuffleupagus (approved):
> This greatly simplifies the function and also makes sure that we fail directly in case a bad URL is added to the manifest,

Failing immediately/explicitly is much nicer indeed, since there's nothing worse than a silent failure when trying to debug something!

> instead of having it work "accidentally" because of this logic, since we want the manifest to be correct at all times (and otherwise fail loudly).

Honestly, the fallback logic was probably never all that helpful to begin with, since it's likely that the md5-hashes didn't match and in that case you might not even test the intended thing anyway.

---

Looks good to me, whatever can be done to modernize/improve the test-code is a welcome change; thank you!

PR #13421, opened 2021-05-22
XFA - Don't display images with a href

calixteman:


timvandermeij (approved):

timvandermeij:
This seems best for now indeed. Thanks!

PR #13424, opened 2021-05-23
[api-minor] Fix issues in text selection

### LINKED ISSUE #6705: Space inserted between each letter in textLayer
PDF file:
http://www.filedropper.com/extraspaces

The text layer is rendered with extra spaces between each letter, for example:

Correct text: "VELKOMMEN TIL ALMINDINGEN"
Rendered text: "V E L K O M M E N   T I L   A L M I N D I N G E N"

### LINKED ISSUE #7833: Copy text selection from PDF adds extra line breaks where they don't belong
http://research.microsoft.com/en-us/um/people/hiballan/pubs/ccs08-staledns.pdf

Configuration:
- Web browser and its version: Chrome 54.0.2840.98 (64-bit) and Firefox 49.0.2
- Operating system and its version: Mac OSX Sierra
- PDF.js version: 1.6.210
- Is an extension: no

1. Open the referenced PDF
2.Try copying text from the article (I've just tried on the first page)
3. Paste somewhere else, and spaces are missing between most words.

Copy the text from the pdf with spaces between words: EG first line in paper should copy as
This paper considers DoS attacks on DNS wherein attackers flood

This is what is copied
This
paper
consider
s DoS
attac
ks
on
DNS
wher
ein
attac
kers flood

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): Reproduced this using the github hosted version.

I have a pull request to fix this problem in the works.

### LINKED ISSUE #9247: PDF.js Not finding all matches from pdf
[Hong Kong Aircraft Engineering Company Limited_1511417616.1 2.pdf](https://github.com/mozilla/pdf.js/files/1539395/Hong.Kong.Aircraft.Engineering.Company.Limited_1511417616.1.2.pdf)

Configuration:
- Web browser and its version: Chrome: 63 
- Operating system and its version: windows and macOs
- PDF.js version: 2.0.185
- Is a browser extension: No

1. Upload the above PDF and search for "article 1"
2. PDF.js showing only **2** match But actually **4** times available.   

screen shot 2017-12-07 at 7

### LINKED ISSUE #9998: Extra spaces inside words when using getTextContent()
[page226-simple.pdf](https://github.com/mozilla/pdf.js/files/2309555/page226-simple.pdf)

Configuration:
- Web browser and its version: Firefox Quantum 61.0.2 (64-bit) and Chrome 68.0.3440.106
- PDF.js version: latest version from http://mozilla.github.io/pdf.js/build/pdf.js
- Is a browser extension: No

1. Copy this html in the same folder as the pdf and open it in a browser.

```
<!doctype html>
<html>
<head>
	<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
	<script src="http://mozilla.github.io/pdf.js/build/pdf.js"></script>
	<script>
		function pdfToPlainText(pdfData) {
			pdfjsLib.disableWorker = true;
			var pdf = pdfjsLib.getDocument(pdfData);
			pdf.then(getPages);
		}
		function getPages(pdf) {
			var pagePromiseArr = [];
			for (var i = 0; i < pdf.numPages; i++) {
				pagePromiseArr.push(pdf.getPage(i + 1));
			}
			Promise.all(pagePromiseArr).then(getPagesText);
		}
		function getPagesText(pageArr) {
			var textContentPromiseArr = [];
			for(var j=0; j<pageArr.length; j++)
			{
				var page = pageArr[j];
				textContentPromiseArr.push(page.getTextContent());
				
			}
			Promise.all(textContentPromiseArr).then(function(textContentArr) {
				
				for(var j=0; j<textContentArr.length; j++)
				{
					var textContent = textContentArr[j];
					textContent.items.forEach(function(o) {
						$("#pdf").append(o.str);
						$("#pdf").append("<br>");
					});	

				}
			});
		}
	</script>
</head>
<body onload="pdfToPlainText('page226-simple.pdf')">
	<div id="pdf" style='font-family: "Courier New", monospace; font-size: 10px'></div>
</body>
</html>
```

The text should be (this is the text that you can copy and paste from Acrobat Reader or Chrome's PDF viewer):

```
Vehicles Involved and Human Injuries
fields should be left blank by default until
operator inputs an accurate value.
```

What went wrong?

The text has extra spaces inside most of the words:

```
Vehi cl es Invol ved and Human Inj uri es
fi el ds s houl d be l eft bl ank by defaul t unti l
operator i nputs an accurate val ue. 
```

The same bug happens in Firefox default PDF viewer (which uses pdf.js):
* If you open the file in Firefox
* select all, 
* copy 
* and paste the text in Notepad for example.

### LINKED ISSUE #10448: highlight postion is not corretly when search a word，nor copy a word
image

Configuration:
- Web browser and its version:chrome 71.0.3578.98
- Operating system and its version:mac os 10.12.6
- PDF.js version:I use online demo in github
- Is a browser extension:no

1. search "中华" or "电力" or so on,the highlight position is not right

image

image

### LINKED ISSUE #10640: No spaces between words when copying text
PDF files:
[1.pdf](https://github.com/mozilla/pdf.js/files/2955999/1.pdf)
[2.pdf](https://github.com/mozilla/pdf.js/files/2956000/2.pdf)
[3.pdf](https://github.com/mozilla/pdf.js/files/2956002/3.pdf)

Configuration:
- Web browser and its version: Chrome, Firefox
- Operating system and its version: Ubuntu, macOS
- PDF.js version: 2.2.42, but the same was with 2.1.x, and with https://mozilla.github.io/pdf.js/web/viewer.html
- Is a browser extension: no

1. Select text
2. Copy text

What went wrong?

Getting this:
`OpenSansisahumanistsansseriftypefacedesignedbySteveMatteson.OpenSanswasdesignedwithanuprightstress,openformsandaneu-tral,yetfriendlyappearance.Itwasoptimizedforprint,web,andmobileinterfaces,andhasexcellentlegibilitycharacteristicsinitsletterforms(seefigureonthefollowingpage).ThisfontisavailablefromtheGoogleFontDirectory[]asTrueTypefileslicensedundertheApacheLicenseversion..`

Instead of this:
`Open Sans is a humanist sans serif typeface designed by Steve Matteson.
Open Sans was designed with an upright stress, open forms and a neutral,
yet friendly appearance. It was optimized for print, web, and mobile
interfaces, and has excellent legibility characteristics in its letterforms (see
figure 1 on the following page). This font is available from the Google Font
Directory [1] as TrueType files licensed under the Apache License version 2.0.`

This is quite significant problem, because I would say 1/20 of scientific papers encounter this problem.

### LINKED ISSUE #10900: Investigate showSpacedText - getTextContent returning wrong widths
Attach (recommended) or Link to PDF file here (unable to attach as PDF falls under NDA).  Here is a relevant section:

image

Configuration:
- Web browser and its version: Node v10.11.0
- Operating system and its version: MacOS 10.14.5
- PDF.js version: v2.1.266
- Is a browser extension: No

1. Calling `getTextContent` on the page is returning inappropriate widths for text items with appended with spaces.

The expected behavior is that the coordinates for the string "3" are above the coordinates for the string "27.0" coordinates PLUS the width of the "27.0" string.  This should make sense as the numbers are on top of each other.

What went wrong?
image

Unfortunately, the width is calculated wrong and "27.0" has a width of `2.17...` which does not make sense in this context.  The string "3" also has a width of `2.15...` which means that these two strings are off by `.02...` units which is not the case.

I dug into the source code of pdfjs and found commenting out [this line](https://github.com/mozilla/pdf.js/blob/c8c937c2579812798fddc13af9d8c1df1e7ce30b/src/core/evaluator.js#L1657) resolved the issue for me.  I bet this line is necessary though, so would like to start an investigation (happy to help!) into what calculations are going wrong to cause this wrongly reported width.

### LINKED ISSUE #11913: Search issue

Configuration:
- Web browser and its version: Version 81.0.4044.138 (Official Build) (32-bit)
- PDF.js version: latest
- Is a browser extension: no

1. Open https://mozilla.github.io/pdf.js/web/viewer.html
2. Open the provided document
3. ctrl + F  and search for 'costs' .
4. Search for 'costs' - match NOT found.
5. Click the 'Download' and open the file in Adobe Reader, and search for costs - match is found.

The word 'costs' to be returned by search

No results found

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):  https://mozilla.github.io/pdf.js/web/viewer.html
[2001_IESC_63_5_v6.pdf](https://github.com/mozilla/pdf.js/files/4644486/2001_IESC_63_5_v6.pdf)

### LINKED ISSUE #13201: Unable to search for phrases in justified text
https://bitcoin.org/bitcoin.pdf

Configuration:
- Web browser and its version: all
- Operating system and its version: all
- PDF.js version: tested with 2.9.11, but most likely all versions
- Is a browser extension: no

1. Type "a purely" (single space between "a" and "purely") - doesn't find anything
2. Type "a  purely" (two spaces between "a" and "purely") - finds the phrase

Sometimes even 3 spaces are necessary.

Screenshot from 2021-04-08 13-00-03

Screenshot from 2021-04-08 13-01-53

This affects many PDFs because justify alignment is common enough.

### LINKED ISSUE #13226: German text incorrectly recognized
https://boersengefluester.de/wp-content/uploads/assets/annuals/2019/578560.pdf

Configuration:
- PDF.js version: 2.7.570

1. Load the linked PDF
2. On the second page, highlight the word "Mitarbeiterinnen" by double-clicking it

The word should be recognized as a single word and highlighted as such.

What went wrong?
Instead, it's recognized as 3 words: "Mit", "arbei", "terinnen"
image

This happens for many words in this PDF.

Is there anything that can be changed in how I use pdfjs-dist to make it recognize this as a single word? Just looking at the PDF, I can't tell what would cause this behavior.

calixteman:
- PR #13257 fixed a lot of issues but not all and this patch aims to fix almost all remaining issues.
  - the idea in this new patch is to compare position of new glyph with the last position where a glyph has been drawn;
    - no space are "drawn": it just moves the cursor but they aren't added in the chunk;
    - so this way a space followed by a cursor move can be treated as only one space: it helps to merge all spaces into one.
  - to make difference between real spaces and tracking ones, we used a factor of the space width (from the font)
    - it was a pretty good idea in general but it fails with some fonts where space was too big:
    - in Poppler, they're using a factor of the font size: this is an excellent idea (<= 0.1 * fontSize implies tracking space).

Fixes #6705
Fixes #7833
Fixes #9247
Fixes #9998
Fixes #10448
Fixes #10640
Fixes #10900
Fixes #11913
Fixes #13201
Fixes #13226


test/test_manifest.json
```diff
@@ -97,6 +97,12 @@
        "link": false,
        "type": "text"
     },
+    {  "id": "issue10900",
+       "file": "pdfs/issue10900.pdf",
+       "md5": "78e598c748076a1dc9e67bbec1a1b666",
+       "rounds": 1,
+       "type": "other"
```
Snuffleupagus, line 104: Please note that "other" was added to be able to use *linked* files in e.g. unit/integration-tests, hence there's no need add to update `test_manifest` in this case as far as I can tell.

test/unit/api_spec.js
```diff
@@ -1500,13 +1504,20 @@ describe("api", function () {
       });
 
       const data = await Promise.all([defaultPromise, parametersPromise]);
+      // console.log(data);window.alert("");
```
Snuffleupagus, line 1507: Nit: Looks like a bit of left-over debugging.

test/unit/api_spec.js
```diff
@@ -1539,6 +1550,107 @@ describe("api", function () {
...
+      ).toEqual(true);
+      expect(
+        text.includes(
+          "argued in this Court the appeal raises narrower, but important, issues which may be summarised as follows:-"
+        )
+      ).toEqual(true);
+      await loadingTask.destroy();
+    });
+
+    it("gets text content, with merged (issue 10900)", async function () {
```
Snuffleupagus, line 1613: Nit: It looks like there's one and/or more words missing after "merged" here.

test/unit/api_spec.js
```diff
@@ -1539,6 +1550,107 @@ describe("api", function () {
...
+633.6 727.8 789.9 796.2
+1,485.1 1,582.7 1,629.2 1,633.7
+114.2 121.7 125.3 130.7
+13.0x 13.0x 13.0x 12.5x`)
+      ).toEqual(true);
+
+      await loadingTask.destroy();
+    });
+
+    it("gets text content, with spaces", async function () {
```
Snuffleupagus, line 1632: Nit: Perhaps add ` (issue10640)` at the end, like all the other tests?

Snuffleupagus (changes requested):
Unfortunately this patch causes rendering regressions in `eq`-tests, and those tests should definitely *not* be changing for a text-selection only patch (see a possibly incomplete list below). My guess is that the `src/core/fonts.js` changes are not generally appropriate and/or correct here.
 - bug921760
 - issue9949
 - issue6127
 - font_ascent_descent (slight movement)
 - issue6737 (slight movement)

There's also some `text`-tests which looks slightly worse than before, some examples include:
 - tracemonkey-text-page14
 - issue11713
 - issue6019
 - issue5972
 - issue3925

src/core/fonts.js
```diff
@@ -2903,6 +2903,7 @@ class Font {
       unicode = String.fromCharCode(unicode);
     }
 
+    const isSpace = unicode === " ";
```
Snuffleupagus, line 2906: Given the rendering regressions, which I'm pretty certain are caused by the `isSpace` change, a naive idea would be to revert this and instead add a new `isUnicodeSpace` property to `Glyph`-instances for the `unicode === " ";`-case necessary for the text-selection logic.

calixteman: I don't know how I missed the failing tests... likely I was too focused on spaces.

calixteman:
> Unfortunately this patch causes rendering regressions in `eq`-tests, and those tests should definitely _not_ be changing for a text-selection only patch (see a possibly incomplete list below). My guess is that the `src/core/fonts.js` changes are not generally appropriate and/or correct here.
> 
>     * bug921760
> 
>     * issue9949
> 
>     * issue6127
> 
>     * font_ascent_descent (slight movement)
> 
>     * issue6737 (slight movement)
>

They should be fixed now (I reverted changes in `fonts.js`).
 
> 
> There's also some `text`-tests which looks slightly worse than before, some examples include:
> 
>     * tracemonkey-text-page14
>

There's slightly less spans so it implies a shift between canvas and text layer.
 
>     * issue11713
> 

There no font change between sigmas but scale factor changes so chunks are not flushed and glyphes with different scale factor can be merged.

>     * issue6019
> 

I handled this in using the rotation angle from document: it's ok now.

>     * issue5972
> 

I'm not sure to know if it's worth or better.
Anyway some chunks are flushed because of https://github.com/mozilla/pdf.js/pull/13424/files#diff-cf40014d9c7d352f3ae212e7fffe5386decf96a10e53066a4d900c3eb8fc559eR2339.
We can change the factor in `HALF_LAST_CHAR` (and the name too) to have all the chars in the same chunk, what do you think ?

>     * issue3925

I fixed the big space things but some shift are due to less spans.

src/core/fonts.js
```diff
@@ -2884,7 +2884,7 @@ class Font {
   /**
    * @private
    */
-  _charToGlyph(charcode, isSpace = false) {
+  _charToGlyph(charcode, isSpace) {
```
Snuffleupagus, line 2887: Nit: Unrelated change, please revert.

src/core/evaluator.js
```diff
@@ -2264,36 +2260,62 @@ class PartialEvaluator {
...
+      let lastPosX = textContentItem.prevTransform[4];
+      let lastPosY = textContentItem.prevTransform[5];
 
       if (lastPosX === posX && lastPosY === posY) {
         return;
       }
 
-      const advanceX = (posX - lastPosX) / textContentItem.textAdvanceScale;
-      const advanceY = (posY - lastPosY) / textContentItem.textAdvanceScale;
+      if (rotate !== 0) {
```
Snuffleupagus, line 2272: I'm not entirely sure that I understand why the *page* rotation really matters here, since a PDF document can easily specify the current transformation matrix and/or the text matrix such all text is rendered at an (arbitrary) angle to the page itself?

calixteman: I added that to handle `issue6019.pdf`: visually we've LTR strings but if we omit the rotation it's TTB (or BTT, I don't remember) strings.
So here I finally just "apply" the rotation matrix to the current text matrix.
It was working ok before my patch because we just got chars from a stream without really thinking about their positions.
Anyway, from a user pov, there are LTR strings.

src/core/fonts.js
Snuffleupagus: As mentioned previously, please revert this change!

src/core/evaluator.js
```diff
@@ -2162,13 +2162,16 @@ class PartialEvaluator {
     includeMarkedContent = false,
     sink,
     seenStyles = new Set(),
+    rotate = 0,
```
Snuffleupagus, line 2165: Don't you want to include this parameter in the `getTextContent` invocation below as well?

https://github.com/mozilla/pdf.js/blob/f5b79be0b7fc2e56bdc9829b2c98369afe2e4e8e/src/core/evaluator.js#L2969-L2980

Snuffleupagus (changes requested):
Looking at the test results, I do believe that there's too much movement (and some outright regressions) in existing test-cases. The following is a non-exhaustive list (focusing on Firefox), so please go through the test results and see what can be done to improve (some of) these cases:
 - tracemonkey-text-page1, one case of worse agreement between the canvas/textLayer.
 - issue3925-page1, many cases of disagreement between the canvas/textLayer.
 - taro-text-page1..4, worse agreement between the canvas/textLayer for the vertical text.
 - issue11713-page1, the sizing looks off for the last glyph.
 - issue1127-text-page1, there's both improvements and regressions,
 - issue4684-text-page1, lots of separate spaces being inserted.
 - issue5896-text-page1, another possible issue with space handling (based on the reference images).
 - mao-text-page1, slightly worse agreement between the canvas/textLayer in places.
 - issue6605-page1, all spaces seem to be missing in the textLayer (causing bad alignment).

calixteman:
> Looking at the test results, I do believe that there's too much movement (and some outright regressions) in existing test-cases. The following is a non-exhaustive list (focusing on Firefox), so please go through the test results and see what can be done to improve (some of) these cases:
> 
>     * tracemonkey-text-page1, one case of worse agreement between the canvas/textLayer.

Looking the "Category..." sentence: it looks like a font artifact, because locally chars are almost at the same positions.
Looking at "Copyright...": the 3 chars "ACM" are larger in the pdf than in text layer and everything is shifted.

> 
>     * issue3925-page1, many cases of disagreement between the canvas/textLayer.

Clearly, this one a bit worse...
It's mainly because of:
https://github.com/calixteman/pdf.js/blob/a1add12e187b5dde1e2e8206c47888de80b15bb0/src/core/evaluator.js#L2212
If I decrease it to 0.3 I get almost the same result as before, but at the end the text layer contains more spans.

> 
>     * taro-text-page1..4, worse agreement between the canvas/textLayer for the vertical text.
> 

We must take into account the rotation we've in the current transform.
And when there is a space between two chars (non-spaces) we must add it in 

>     * issue11713-page1, the sizing looks off for the last glyph.

The two final chars with a space between are in the same chunk because of the `MAX_FACTOR`.
It's possible to break the flow when a scale occurs, but I'm not sure that's a good idea to add some code to handle such a corner case.
> 
>     * issue1127-text-page1, there's both improvements and regressions,
> 

It's still a matter of `MAX_FACTOR`.

>     * issue4684-text-page1, lots of separate spaces being inserted.

It's a bug: when a space is between two non-spaces chars and in the same chunk `(xxxSPACExxx) TJ`, we must add a space in the flow.

> 
>     * issue5896-text-page1, another possible issue with space handling (based on the reference images).

It's just a matter of rescale on a span with one space: I'll fix it.

> 
>     * mao-text-page1, slightly worse agreement between the canvas/textLayer in places.
> 

Again because of `MAX_FACTOR`.

>     * issue6605-page1, all spaces seem to be missing in the textLayer (causing bad alignment).

Same issue as for  issue4684-text-page1

Snuffleupagus:
That's certainly an improvement, however I still feel like there's too much movement in e.g. `issue3925-page1`. While I don't know how representative that particular document is, given that we don't have loads of `text`-tests I'm a bit worried about new regressions.

Based on https://github.com/mozilla/pdf.js/pull/13424#issuecomment-939317829 it seems that most (if not all) remaining "regressions" are related to `MAX_FACTOR`, hence my question is what the overall effect would be if that constant is tweaked slightly?

src/display/text_layer.js
```diff
@@ -185,10 +185,11 @@ function appendText(task, geom, styles, ctx) {
   let shouldScaleText = false;
   if (
     geom.str.length > 1 ||
-    (task._enhanceTextSelection && AllWhitespaceRegexp.test(geom.str))
+    (task._enhanceTextSelection &&
+      (geom.isOneSpace || AllWhitespaceRegexp.test(geom.str)))
   ) {
     shouldScaleText = true;
-  } else if (geom.transform[0] !== geom.transform[3]) {
+  } else if (!geom.isOneSpace && geom.transform[0] !== geom.transform[3]) {
```
Snuffleupagus, line 192: Again, I might be missing something, but do we really need this new `geom.isOneSpace` property?
Could we simply not replace that part with a `geom.str !== " " && ...` check, and avoid having to add a new property to each textItem?

Snuffleupagus (commented):
I'm honestly not really sure what to make of the test results, since regardless of the `MAX_FACTOR` value there's going to be *some* regressions.
/cc @brendandahl What's your opinion here, what sort of movement (in existing test-cases) are we willing to trade for a handful of fixed issues?

brendandahl:
Calixte mentioned .5 seemed to be the best compromise, so I looked through those ones. As mentioned above there are few regressions but there also some improvements (issue6019-text-page1, taro-text-page1).  IIRC the PDF in `issue3925-page1` is a really weird one, so I'm not too worried about regressing it. 

One I'm a bit concerned about is `mao-text-page1`, we have very few CJK text layer tests. Is there something we should do differently for CJK fonts?

If we can fix mao, I'd be fine with taking a few minor regressions for all the copy/paste bugs this fixes.

calixteman:
I set the MAX_FACTOR to 0.6 and fixed few issues in mao test.
/botio test

Snuffleupagus (approved):

PR #13427, opened 2021-05-24
XFA - Add a storage to save fields values

calixteman:
- this is required to be able to print (or save) a document. Some pages can be unloaded (because pdf.js is lazy) and this storage will help to save their data in order to resuse them when printing or just when displaying a page again.

### DIFF
--- a/src/display/xfa_layer.js
+++ b/src/display/xfa_layer.js
@@ -14,9 +14,60 @@
  */
 
 class XfaLayer {
-  static setAttributes(html, attrs) {
-    for (const [key, value] of Object.entries(attrs)) {
-      if (value === null || value === undefined) {
+  static setupStorage(html, fieldId, element, storage) {
+    const storedData = storage.getValue(fieldId, { value: null });
+    switch (element.name) {
+      case "textarea":
+        html.textContent = storedData.value !== null ? storedData.value : "";
+        html.addEventListener("input", event => {
+          storage.setValue(fieldId, { value: event.target.value });
+        });
+        break;
+      case "input":
+        if (storedData.value !== null) {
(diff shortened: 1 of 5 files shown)


src/display/xfa_layer.js
```diff
@@ -30,13 +81,20 @@ class XfaLayer {
         Object.assign(html.style, value);
       }
     }
+
+    // Set the value after the others to be sure overwrite
+    // any other values.
+    if ("fieldId" in attributes && storage) {
```
Snuffleupagus, line 87: Nit: It's probably a tiny bit cheaper to re-order this as `if (storage && attributes.fieldId !== undefined) {` instead.

src/display/xfa_layer.js
```diff
@@ -30,13 +81,20 @@ class XfaLayer {
...
+      XfaLayer.setupStorage(html, attributes.fieldId, element, storage);
```
Snuffleupagus, line 88: Nit: Here, and elsewhere in this file, note that `static` methods should be able to call each other directly; hence `this.setupStorage(...)` ought to work.

brendandahl:
Will this be needed if we keep all pages in cache?

Snuffleupagus:
> Will this be needed if we keep all pages in cache?

That would be breaking some very old/basic assumptions in the viewer, since only one-page documents are ever guaranteed to be fully rendered at all times. Relaxing that assumption could be quite bad in the general case, for e.g. memory usage and performance, for anything except very short documents and as far as I can tell there's no *guarantee* that all XFA-documents are indeed "short".

brendandahl (approved):

PR #13435, opened 2021-05-25
Enable the `unicorn/no-array-push-push` ESLint plugin rule

Snuffleupagus:
There's generally speaking no need to use multiple consecutive `Array.prototype.push()` calls, since that method accepts multiple arguments, and this ESLint rule helps enforce that pattern.

Please see https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-push-push.md for additional information.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -43,6 +43,7 @@
       "ignoreCase": true,
     }],
     "unicorn/no-abusive-eslint-disable": "error",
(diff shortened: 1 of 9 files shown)


timvandermeij (approved):

timvandermeij:
This looks like a nice addition to me; thank you!

PR #13436, opened 2021-05-25
Re-factor FontFaceObject.getPathGenerator to use Arrays instead of strings

Snuffleupagus:
This is similar to a lot of other code, where we use "Array + join" rather than repeated string concatenation.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13437, opened 2021-05-25
XFA - Move the fake HTML representation of XFA from the worker to the main thread

calixteman:
- the only goal of this patch is to be able to get synchronously the fake html when printing from firefox:
    - in order to print we need to inject some html in beforeprint callback but we cannot block in waiting for all the pages.
  - from a memory point of view: it doesn't change anything since the fake HTML is deleted in the worker;
  - this way we don't break any assumptions.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -146,8 +146,7 @@ class Page {
 
   _getBoundingBox(name) {
     if (this.xfaData) {
-      const { width, height } = this.xfaData.attributes.style;
-      return [0, 0, parseInt(width), parseInt(height)];
+      return this.xfaData.bbox;
     }
 
     const box = this._getInheritableProperty(name, /* getArray = */ true);
@@ -241,7 +240,9 @@ class Page {
 
   get xfaData() {
     if (this.xfaFactory) {
-      return shadow(this, "xfaData", this.xfaFactory.getPage(this.pageIndex));
+      return shadow(this, "xfaData", {
+        bbox: this.xfaFactory.getBoundingBox(this.pageIndex),
+      });
     }
     return shadow(this, "xfaData", null);
   }
@@ -851,8 +852,11 @@ class PDFDocument {
     return shadow(this, "xfaFaxtory", null);
   }
 
-  get isPureXfa() {
-    return this.xfaFactory !== null;
+  get htmlForXfa() {
+    if (this.xfaFactory) {
+      return this.xfaFactory.getPages();
+    }
+    return null;
   }
 
   async loadXfaFonts(handler, task) {
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -695,7 +695,15 @@ class PDFDocumentProxy {
    * @type {boolean} True if only XFA form.
    */
   get isPureXfa() {
-    return this._pdfInfo.isPureXfa;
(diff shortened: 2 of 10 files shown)


src/core/document.js
```diff
@@ -855,6 +856,13 @@ class PDFDocument {
     return this.xfaFactory !== null;
   }
 
+  get htmlForXfa() {
+    if (this.isPureXfa) {
```
Snuffleupagus, line 860: Is the `isPureXfa` getter still necessary now, since just `if (this.xfaFactory) {` ought to suffice (which is also consistent with other parts of this file)?

src/display/api.js
```diff
@@ -698,13 +698,21 @@ class PDFDocumentProxy {
     return this._pdfInfo.isPureXfa;
   }
 
+  /**
+   * @type {Object} An object representing a HTML tree structure to render
+   * the XFA.
```
Snuffleupagus, line 703:
```javascript
* @type {Object | null} An object representing the HTML tree structure to render
*   the XFA, or `null` when no XFA form exists.
```

src/display/api.js
```diff
@@ -698,13 +698,21 @@ class PDFDocumentProxy {
...
+  }
+
   /**
    * @param {number} pageNumber - The page number to get. The first page is 1.
    * @returns {Promise<PDFPageProxy>} A promise that is resolved with
    *   a {@link PDFPageProxy} object.
    */
   getPage(pageNumber) {
-    return this._transport.getPage(pageNumber);
+    return this._transport.getPage(pageNumber, this._pdfInfo.htmlForXfa);
```
Snuffleupagus, line 715: I'd suggest just adding `this._htmlForXfa = pdfInfo.htmlForXfa;` after line https://github.com/mozilla/pdf.js/blob/3538ef017fa05787d2e0c2b6bdf2d904753da489/src/display/api.js#L2458 instead, to avoid having to pass this property around as much.

src/display/api.js
```diff
@@ -1256,7 +1272,7 @@ class PDFPageProxy {
    *   children, very similar to a HTML DOM tree), or `null` if no XFA exists.
    */
   getXfa() {
-    return (this._xfaPromise ||= this._transport.getPageXfa(this._pageIndex));
+    return (this._xfaPromise ||= Promise.resolve(this._xfaHtml));
```
Snuffleupagus, line 1275: With this being *synchronously* available, I'm not sure that caching a Promise is still necessary!?
Can't you simply remove `this._xfaPromise` completely and instead do e.g.
```javascript
async getXFA() {
  return this._transport._htmlForXfa;
```

src/display/api.js
```diff
@@ -2812,12 +2829,6 @@ class WorkerTransport {
     });
   }
 
-  getPageXfa(pageIndex) {
-    return this.messageHandler.sendWithPromise("GetPageXfa", {
```
Snuffleupagus, line 2816: Also, if you're removing sending of this message the following handler looks like dead code: https://github.com/mozilla/pdf.js/blob/3538ef017fa05787d2e0c2b6bdf2d904753da489/src/core/worker.js#L504-L508

brendandahl (approved):

PR #13445, opened 2021-05-26
Fix Postscript name in font to avoid bug when saving in pdf

calixteman:
- for xfa rendering, fonts are loaded and used in html;
  - when printed and saved in pdf, on linux, Firefox uses cairo backend
  - when subsetting a font, cairo uses the font postscript name and when this one is empty that leads to a bug
    (the append at https://github.com/freedesktop/cairo/blob/63f0d6268469dcd12316ea405a026f4be900dd79/src/cairo-cff-subset.c#L2049 is failing because of null length)
  - so this patch adds a postscript name to the font to make cairo happy.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -722,6 +722,11 @@ function createPostTable(properties) {
   ); // maxMemType1
 }
 
+function createPostscriptName(name) {
+  // See https://docs.microsoft.com/en-us/typography/opentype/spec/recom#name.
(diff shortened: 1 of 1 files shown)


src/core/fonts.js
```diff
@@ -722,6 +722,11 @@ function createPostTable(properties) {
   ); // maxMemType1
 }
 
+function makePostscriptName(name) {
```
Snuffleupagus, line 725: Nit: I'd probably call this `createPostscriptName` for improved consistency, since all of the surrounding functions use that pattern.

Snuffleupagus (approved):

Issue #13447, opened 2021-05-27
Missing PDF Annotations when rendered to PdfJs


roneytham-ideagen:
Attach (recommended) or Link to PDF file here: 
[A Sample PDF Clean 0 degree with all annotations-original-rotated.pdf](https://github.com/mozilla/pdf.js/files/6551789/A.Sample.PDF.Clean.0.degree.with.all.annotations-original-rotated.pdf)

Configuration:
- Web browser and its version: all browsers, tested with Chrome and Firefox latest versions.
- Operating system and its version: tested in Windows 10
- PDF.js version: current (online version)

Steps to reproduce the problem:
1. Visit https://mozilla.github.io/pdf.js/web/viewer.html
2. Upload the file above
3. Notice that several annotations such as arrow, freehand & polyline are missing (compared to Acrobat Reader)
![image](https://user-images.githubusercontent.com/45581023/119781014-11d35d00-befd-11eb-8daa-64f15374cf5b.png)

What is the expected behavior?
- Should display all annotations accordingly to Acrobat Reader

What went wrong? 
- Several annotations are missing such as arrow, freehand & polyline
- Opacity property is not supported and cannot be retrieved via getAnnotations

Snuffleupagus:
This looks, for all intents and purposes, like another duplicate of issue #6810.

roneytham-ideagen:
Alright got it, and we was wondering when will these changes release so that we can update?
https://github.com/mozilla/pdf.js/pull/13301
https://github.com/mozilla/pdf.js/pull/13305
https://github.com/mozilla/pdf.js/pull/13033

timvandermeij:
The pull request above fixed most remaining issues here with the exception of the vertical strikeout annotations and the custom line endings. However, given that all the rest renders like in Adobe Reader, the reference implementation, now that's a minor issue that can be picked up later if need be.

roneytham-ideagen:
@timvandermeij Thanks for the input Tim.

I upgraded to version 2.9.359, everything works well, it just that there are a few missing arrow heads, are those the custom line endings you mentioned above?
![image](https://user-images.githubusercontent.com/45581023/120154326-58d59100-c222-11eb-86f9-416da8136344.png)

Anyhow, thanks for getting these annotations works 👍

timvandermeij:
Indeed, the arrow heads still remain to be fixed here.

Snuffleupagus:
[...]

PR #13448, opened 2021-05-27
Support strokeAlpha/fillAlpha when creating a fallback appearance stream (issue 6810)

Snuffleupagus:
Fixes *most* of issue #13447; with only two things remaining (that I won't attempt here):
 - Support for non-default `/LE`-entries (line endings) in LineAnnotations, assuming we care about that.
 - The "weird" StrikeOutAnnotations with *vertical* lines, again assuming we want to fix that.

With this patch, while we still don't match Adobe Reader perfectly, we're doing a much better job overall than e.g. PDFium since all Annotations are now at least visible in the viewer.
For some Annotation-types without an included appearance stream, it definitely feels like we're getting into the territory of diminishing returns w.r.t. adding even more special-cases. At some point, it's feels *very* tempting to simply blame PDF generators for using more "exotic" Annotations and failing to including an appearance stream.


timvandermeij (approved):

timvandermeij:
Thank you for fixing this, also for the other markup annotation types that according to the spec also support the CA property! I forgot about CA before since I had no test cases that actually used it and I also postponed custom line endings because they seemed very rare.

/botio makeref

PR #13450, opened 2021-05-27
Fix the JSDocs for `PDFDocumentProxy.getPageIndex` (issue 13449)

### LINKED ISSUE #13449: Return type signature of getPageIndex is incorrect
https://github.com/mozilla/pdf.js/blob/0df1a5661958df0c4ed2f73b0f3266d825c5c280/src/display/api.js#L710-L719
The return type signature `@returns {Promise<{num: number, gen: number}>}` is mismatched with its definition. (should be `@returns {Promise<number>}`)

Snuffleupagus:
Fixes #13449

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -708,11 +708,15 @@ class PDFDocumentProxy {
   }
 
   /**
-   * @param {{num: number, gen: number}} ref - The page reference. Must have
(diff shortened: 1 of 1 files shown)


calixteman (approved):
Thanks.

PR #13455, opened 2021-05-28
Italic angle is defined clockwise in CSS when it's counterclockwise in PDF

calixteman:


Snuffleupagus (approved):
Looks good to me, thanks!

PR #13457, opened 2021-05-28
Work-around for HighlightAnnotations without a top-level /ExtGState-entry (issue 13242)

### LINKED ISSUE #13242: markings/highlights in pdf file cover up text
[Testfile v1.5 to PDFA2.pdf](https://github.com/mozilla/pdf.js/files/6310752/Testfile.v1.5.to.PDFA2.pdf)

Configuration:
- Web browser and its version: Version 89.0.4389.114 (Offizieller Build) (x86_64)
- Operating system and its version: Windows 10 / MacOS BigSur
- PDF.js version: 2.5.207
- Is a browser extension: no 

1.  Open Testfile v1.5 to PDFA2.pdf in pdf.js viewer demo
2. text is covered up by markings / highlights which are stored in pdf itself

Should be displayed as in Acrobat Reader
Bildschirmfoto 2021-04-14 um 14 24 41

It's how it is displayed in pdf.js viewer
Bildschirmfoto 2021-04-14 um 14 24 16

Snuffleupagus:
For HighlightAnnotations with a built-in appearance stream, we still rely on it to specify the opacity correctly via a suitable blend mode. However, if the Annotation-drawing operators are placed *within* a /XObject of the /Form-type, the /ExtGState won't apply to the final rendering and the result is that the highlighting obscures the underlying text.

The more *correct* and general solution would likely be to somehow modify the implementation in `src/display/canvas.js`, to special-case handling of /Form-type /XObjects when rendering Annotations. Since we can very easily work-around this problem for now by using the "no appearance stream" code-path, doing *something* here ought to be preferable.

This patch is (obviously) merely a work-around, but given that the referenced issue is (as far as I know) the first case we've seen of this problem a simple solution will hopefully suffice for now.

Fixes #13242


timvandermeij (approved):

timvandermeij:
Thank you! This does seem like the simplest solution, especially since this is the first case we have seen so far.

PR #13460, opened 2021-05-29
Update webpack import instructions

nikolas:
This resolves an issue I had when loading pdf.js with webpack. `pdfjsLib` was always `undefined` after the import, and I dug deep into pdf.js's `webpack.js` file to sort out this issue, making all sorts of changes.

I figured out that there was just something wrong with the export/import procedure. After making this change, I now have the full pdfjs object in my webpack application.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #13461, opened 2021-05-29
Improve text-selection for Type3 fonts with empty /FontBBox-entries (issue 6605)

### LINKED ISSUE #6605: very high marking of type 3 font glyphs with 0 0 0 0 BBox
Text marking with this file
https://issues.apache.org/jira/secure/attachment/12769999/type3%20zero%20font%20size.pdf
has a giant size.

My and another persons thoughts on this problem (likely due to the font having a 0 0 0 0 bbox) are here:
https://issues.apache.org/jira/browse/PDFBOX-3076

Snuffleupagus:
For Type3 fonts where the /CharProcs-streams of the individual glyphs start with a `d1` operator, we can use that to build a fallback bounding box for the font and thus improve text-selection in some cases.

Fixes #6605


timvandermeij (approved):

timvandermeij:
Looks good; thank you for fixing this!

/botio makeref

PR #13472, opened 2021-05-31
Don't change options of the globally used `PartialEvaluator` in the "should render checkbox with fallback font for printing" unit-test

Snuffleupagus:
Given that the same `PartialEvaluator`-instance is used for a lot of these unit-tests, manually changing the options in any one test-case could lead to intermittently failing unit-tests since they're run in a random order.
To fix this, we simply have to use the existing method to clone the `PartialEvaluator`-instance but with the custom options.


calixteman (approved):
Thanks for doing this.

PR #13306, opened 2021-04-27
Fix position of highlighted all text.

brendandahl:
Adds a new integration test to ensure we don't
regress this again.

Fixes #13292.

PR #13478, opened 2021-06-01
Fix find highlighting regression from #13306.

### LINKED ISSUE #13345: [Regression] Highlighting of search results broken in some cases, after PR 13306
Attach (recommended) or Link to PDF file here: https://github.com/tesseract-ocr/tesseract/files/947838/2.pdf

Configuration:
- Web browser and its version: N/A
- Operating system and its version: N/A
- PDF.js version: `2.9.223`, or any version after PR #13306 landed.
- Is a browser extension: N/A

1. Open the document above, e.g. with https://mozilla.github.io/pdf.js/web/viewer.html
2. Search for "of the r"

Correct highlighting, please see:

correct

Incorrect highlighting, please see:

incorrect

/cc @brendandahl

brendandahl:
When we insert extra spans for highlighting we want
them to be positioned normally instead of absolute or
relative.

Fixes #13345.

### DIFF
--- a/web/text_layer_builder.css
+++ b/web/text_layer_builder.css
@@ -34,11 +34,14 @@
 }
 
 .textLayer .highlight {
-  position: relative;
   margin: -1px;
   padding: 1px;
   background-color: rgba(180, 0, 170, 1);
   border-radius: 4px;
+}
+
+.textLayer .highlight.appended {
+  position: initial;
 }
 
 .textLayer .highlight.begin {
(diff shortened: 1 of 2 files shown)


brendandahl:
The highlight still looks a bit wonky, but for this somewhat rare case I think it's okay.
image

Snuffleupagus (approved):
Can we add an integration-test for this, similar to the existing one in `find_spec.js`, to avoid it regressing again?

r=me, thank you for fixing this!

Snuffleupagus:
We can always land tests later, however I'm going to land this as-is (with passing tests) just so that the regression is fixed sooner rather than later.

/botio integrationtest

PR #13480, opened 2021-06-02
[GENERIC viewer] Fix printing regression from PR 13411

Snuffleupagus:
I missed this during review, since some of the changes in `web/pdf_print_service.js` broke printing.

Also, as part of these changes the patch replaces what looks like unnecessary `setAttribute` usage with "regular" `className` assignment and finally updates a couple of the CSS-rules to be more consistent.


calixteman (approved):
My bad... thanks for fixing this.

PR #13489, opened 2021-06-04
Add `hasEOL` to the `TextItem` typedef in the API (PR 13257 follow-up)

Snuffleupagus:


calixteman (approved):
Thanks for doing that.

PR #13490, opened 2021-06-04
Miscellaneous improvements for `gulpfile.js` (issue 10362)

### LINKED ISSUE #10362: Improve the Gulpfile
Recently we upgraded to Gulp 4. However, the following things still need to be addressed in some way:

- If a task series is executed with an anonymous function inside it, we'll see `Starting '<anonymous>'` in the output, which isn't that great. We need to either suppress it or use named functions.
- If a task is private, it should be hidden in some way. For example, `gh-pages-prepare` is only used in the public `web` target and should not be used on its own, however currently it shows in the default `gulp` output since it's a regular task. Perhaps we should mark them with an underscore?

Snuffleupagus:
- Replace anonymous functions in `gulpfile.js` with named functions (issue 10362)

   This replaces all of the `Starting '<anonymous>'...` lines in the output, when running various `gulp` tasks, with more meaningful function names.

 - Clean-up `gulpfile.js` by removing most "-pre" tasks (issue 10362)

   Most of the `gulp` tasks with a "-pre" suffix, especially those related to the main build-targets, are no longer needed and can simply be inlined in their corresponding tasks. This is likely a left-over from older Gulp-versions, but currently it only adds unnecessary indirection.

   Furthermore, a number of the `gulp` tasks which are only used *internally* are converted to standard functions instead. This prevents them from being exposed "globally" and e.g. listed by `gulp default`.

 - Remove `gulp makefile`, since it's broken and unused

   Looking at the history of the code, this `gulp` task was *accidentally* broken in the Gulp 4 upgrade in PR #10266 (all the way back in 2018).
   Given that no one has either noticed and/or complained about this since, it's safe to assume that it's completely unused, hence why I suggest that we just remove this `gulp` task (rather than fixing it).

Fixes #10362

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -829,14 +829,14 @@ gulp.task(
   gulp.series(
     "buildnumber",
     "locale",
-    function scripting() {
+    function scriptingGeneric() {
       const defines = builder.merge(DEFINES, { GENERIC: true });
       return merge([
         buildDefaultPreferences(defines, "generic/"),
         createTemporaryScriptingBundle(defines),
       ]);
     },
-    function () {
+    function createGeneric() {
       console.log();
       console.log("### Creating generic viewer");
       const defines = builder.merge(DEFINES, { GENERIC: true });
@@ -853,7 +853,7 @@ gulp.task(
   gulp.series(
     "buildnumber",
     "locale",
-    function scripting() {
+    function scriptingGenericLegacy() {
       const defines = builder.merge(DEFINES, {
         GENERIC: true,
         SKIP_BABEL: false,
@@ -863,7 +863,7 @@ gulp.task(
         createTemporaryScriptingBundle(defines),
       ]);
     },
-    function () {
+    function createGenericLegacy() {
       console.log();
       console.log("### Creating generic (legacy) viewer");
       const defines = builder.merge(DEFINES, {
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
With this PR, `gulp default` gives the output below. As far as can tell all of these tasks makes sense to expose, since they're either actual build-targets or at least useful during development and/or testing.
```
$ gulp default
[13:30:33] Using gulpfile ~\Git\pdf.js\gulpfile.js
[13:30:33] Starting 'default'...
Available tasks:
  botmakeref
  bottest
  browsertest
  chromium
  clean
  cmaps
  components
  components-legacy
  default
  dev-sandbox
  dist
  dist-install
  externaltest
  fonttest
  generic
  generic-legacy
  image_decoders
  image_decoders-legacy
  importl10n
  integrationtest
  jsdoc
  lib
  lib-legacy
  lint
  lint-chromium
  locale
  makeref
  minified
  minified-legacy
  mozcentral
  mozcentralbaseline
  mozcentraldiff
  npm-test
  publish
  server
  test
  types
  typestest
  unittest
  unittestcli
  web
  wintersmith
[13:30:33] Finished 'default' after 19 ms
```

timvandermeij (approved):

timvandermeij:
Looks much better; thank you for fixing this!

PR #13492, opened 2021-06-04
Add normalization for Hyphen -> Hyphen-minus

MMeent:
Previously these two characters were not searchable interchangably, even when Hyphen-Minus is being changed to Hyphen in some text to PDF pipelines.

example of what it fixes: The CA/B Forum Baseline Requirements documents (v1.7.5, https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.7.5.pdf)  currently contain "SHA-1" on page 15 (search 'Subordinate CA certificates using the SHA"), but you won't find matches for "SHA-1" due to it using Hyphen (\u2010) instead of Hyphen-Minus (\u002D).


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #13494, opened 2021-06-04
Add more info for showText operator in stepper.

brendandahl:
Adds a table that shows original char code, font char code, and unicode.

### DIFF
--- a/src/core/fonts.js
+++ b/src/core/fonts.js
@@ -159,6 +159,7 @@ function adjustToUnicode(properties, builtInEncoding) {
 
 class Glyph {
   constructor(
+    originalCharCode,
     fontChar,
     unicode,
     accent,
@@ -168,6 +169,7 @@ class Glyph {
     isSpace,
     isInFont
   ) {
+    this.originalCharCode = originalCharCode;
     this.fontChar = fontChar;
     this.unicode = unicode;
     this.accent = accent;
@@ -179,6 +181,7 @@ class Glyph {
   }
 
   matchesForCache(
+    originalCharCode,
     fontChar,
     unicode,
     accent,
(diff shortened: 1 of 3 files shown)


brendandahl:
I've had to manually figure this out so many times. I think it's time to make it easier.

image

src/core/fonts.js
```diff
@@ -159,6 +159,7 @@ function adjustToUnicode(properties, builtInEncoding) {
 
 class Glyph {
   constructor(
+    originalCharCode,
```
Snuffleupagus, line 162: Don't you need to update the `matchesForCache` method and its call-site as well, to avoid inconsistent data?
 - https://github.com/mozilla/pdf.js/blob/425e58431a354673817f4687b6cf27e7350c4945/src/core/fonts.js#L181
 - https://github.com/mozilla/pdf.js/blob/425e58431a354673817f4687b6cf27e7350c4945/src/core/fonts.js#L2958

web/viewer.css
```diff
@@ -1699,6 +1699,15 @@ html[dir="rtl"] #documentPropertiesOverlay .row > * {
...
   font-size: 10px;
 }
+#PDFBug table.showText {
+  border-collapse: collapse;
+  text-align: center;
+}
+#PDFBug table.showText,
+#PDFBug table.showText tr,
+#PDFBug table.showText td {
+  border: 1px solid black;
```
Snuffleupagus, line 1709: I'd suggest adding e.g. `padding: 1px` here as well, since that'd improve readability of the table data a little bit (especially for the `originalCharCode` values).

Snuffleupagus (approved):
r=me, with a couple of small comments; thank you!

src/core/fonts.js
```diff
@@ -2967,6 +2971,7 @@ class Font {
       )
     ) {
       glyph = new Glyph(
+        charcode,
```
Snuffleupagus, line 2974: Please note that you still need to insert `charCode` before line 2963 above, otherwise this won't work as intended.

brendandahl: 🤦

PR #13497, opened 2021-06-04
Initialize `HTMLResult.{FAILURE, EMPTY}` lazily

Snuffleupagus:
While these objects aren't exactly that big and/or complex, they are nonetheless *only* necessary for XFA documents.
However, currently these objects are initialized *eagerly* for all PDF documents. By using the same pattern as elsewhere in the code-base, it's very easy to make these lazily initialized; so let's just do that instead :-)

### DIFF
--- a/src/core/xfa/utils.js
+++ b/src/core/xfa/utils.js
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #13505, opened 2021-06-07
Use the `DOMSVGFactory`, rather than manually creating the SVG-element, in `createMatrix` (PR 13361 follow-up)

Snuffleupagus:
Generally, in the `src/display/` folder, we utilize `DOMSVGFactory` rather than manually creating an SVG-element; hence let's do the same thing in `src/display/pattern_helper.js` as well.


calixteman (approved):
LGTM.

PR #13506, opened 2021-06-07
XFA - Add support for reftests

calixteman:
In order to render correctly embedded fonts, we need hack the font loader to get font rules to inject them in the svg.
The idea is to pass a style element through the api to the font loader, load font rules in this element and finally get rules as string in the driver.

### DIFF
--- a/src/display/font_loader.js
+++ b/src/display/font_loader.js
@@ -29,6 +29,8 @@ class BaseFontLoader {
     docId,
     onUnsupportedFeature,
     ownerDocument = globalThis.document,
+    // For testing only.
+    styleElement = null,
   }) {
     if (this.constructor === BaseFontLoader) {
       unreachable("Cannot initialize BaseFontLoader.");
@@ -38,7 +40,10 @@ class BaseFontLoader {
     this._document = ownerDocument;
 
     this.nativeFontFaces = [];
-    this.styleElement = null;
+    this.styleElement =
+      typeof PDFJSDev === "undefined" || PDFJSDev.test("!PRODUCTION || TESTING")
+        ? styleElement
+        : null;
   }
 
   addNativeFontFace(nativeFontFace) {
@@ -55,7 +60,6 @@ class BaseFontLoader {
         .getElementsByTagName("head")[0]
         .appendChild(styleElement);
     }
-
     const styleSheet = styleElement.sheet;
     styleSheet.insertRule(rule, styleSheet.cssRules.length);
   }
@@ -121,7 +125,18 @@ class BaseFontLoader {
   }
 
   get isFontLoadingAPISupported() {
-    return shadow(this, "isFontLoadingAPISupported", !!this._document?.fonts);
+    const hasFonts = !!this._document?.fonts;
+    if (
+      typeof PDFJSDev === "undefined" ||
+      PDFJSDev.test("!PRODUCTION || TESTING")
+    ) {
+      return shadow(
+        this,
+        "isFontLoadingAPISupported",
+        hasFonts && !this.styleElement
+      );
+    }
+    return shadow(this, "isFontLoadingAPISupported", hasFonts);
   }
 
   // eslint-disable-next-line getter-return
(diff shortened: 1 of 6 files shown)


test/driver.js
```diff
@@ -278,6 +276,83 @@ var rasterizeAnnotationLayer = (function rasterizeAnnotationLayerClosure() {
...
+      stylePromise
+        .then(async cssRules => {
+          style.textContent = fontRules + "\n" + cssRules;
+          // console.log(style);window.alert();
+
+          pdfjsLib.XfaLayer.render({
+            xfa,
+            div,
+            viewport: viewport.clone({ dontFlip: true }),
+            intent: "print",
```
Snuffleupagus, line 328: This should be `display` instead, right?
Otherwise you're not actually testing the *exact* same code-paths as used when rendering in e.g. the viewer.
Furthermore, note that we also use a `print` flag in some reference tests, so having this hard-coded could also lead to inconsistencies elsewhere.

test/pdfs/.gitignore
```diff
@@ -278,6 +278,7 @@
 !issue5070.pdf
 !issue5238.pdf
 !issue5244.pdf
+!hsbc.pdf
```
Snuffleupagus, line 281: This should *probably* be a linked test-case instead, looking at the file contents.

Snuffleupagus (changes requested):
I've left a couple of comments, based on a quick first look; although I still haven't looked through the changes in `test/driver.js` in any detail.

*Please note:* This feels like the sort of changes where *some* context (in the commit message and/or PR description) would definitely have helped when going over the code.
For example: Given that you apparently needed to "hack" the API and FontLoader code to support XFA-testing, having explained why upfront would have saved me a little bit of time here.

src/display/font_loader.js
```diff
@@ -121,6 +124,16 @@ class BaseFontLoader {
...
 
   get isFontLoadingAPISupported() {
+    if (
+      typeof PDFJSDev === "undefined" ||
+      PDFJSDev.test("!PRODUCTION || TESTING")
+    ) {
+      return shadow(
+        this,
+        "isFontLoadingAPISupported",
+        !!this._document?.fonts && !this.styleElement
```
brendandahl, line 134: Pull out the `!!this._document?.fonts` from here and below to a variable above so we don't have to keep these in sync.

src/display/font_loader.js
```diff
@@ -29,6 +29,7 @@ class BaseFontLoader {
     docId,
     onUnsupportedFeature,
     ownerDocument = globalThis.document,
+    styleElement = null,
```
brendandahl, line 32: Add inline comment, something like "For testing only."

brendandahl (approved):
r+ with some minor nits

test/driver.js
```diff
@@ -278,6 +265,67 @@ var rasterizeAnnotationLayer = (function rasterizeAnnotationLayerClosure() {
...
+
+  // eslint-disable-next-line no-shadow
+  function rasterizeXfaLayer(ctx, viewport, xfa, fontRules) {
+    return new Promise(function (resolve, reject) {
+      // Building SVG with size of the viewport.
+      const svg = document.createElementNS(SVG_NS, "svg:svg");
+      svg.setAttribute("width", viewport.width + "px");
+      svg.setAttribute("height", viewport.height + "px");
+
+      // Adding element to host our HTML (style + text layer div).
```
Snuffleupagus, line 291: Nit: Too much copy-and-paste here :-)
`// Adding element to host our HTML (style + xfa layer div).`

test/pdfs/Contestation-Transaction-Carte-Bancaire.pdf.link
```diff
@@ -0,0 +1 @@
+https://web.archive.org/save/https://www.hsbc.fr/content/dam/hsbc/fr/docs/pib/Contestation-Transaction-Carte-Bancaire.pdf
```
Snuffleupagus, line 1: This doesn't look like a working Internet Archive URL, and the tests failed as a result.

You'd need to use e.g. https://web.archive.org/web/20210607145115/https://www.hsbc.fr/content/dam/hsbc/fr/docs/pib/Contestation-Transaction-Carte-Bancaire.pdf instead (also with a new-line at the end to stop GitHub complaining), but given that it's already cached on the bots things might not work with only that change.
Hence the simplest solution might be to re-name this file to `test/pdfs/hsbc.pdf.link` instead, and update the manifest as well.

Snuffleupagus (approved):

Issue #11399, opened 2019-12-09
μ (mu) gets rendered as α (alpha)


vale981:
In 
https://tu-dresden.de/mn/physik/iktp/ressourcen/dateien/studium/praktikum/LM_Anleitung_komplett.pdf?lang=en
μ (mu) gets rendered as α (alpha).
This doesn't usually happen so I figure that this is a font issue, maybe even specific to my machine.

Configuration:
- Web browser and its version: Firefox 71
- Operating system and its version: Arch Linux, Kernel Linux 5.1.12
- PDF.js version:
```js
var pdfjsVersion = '2.4.43';
var pdfjsBuild = '16ae7c69';
```
- Is a browser extension: no

Steps to reproduce the problem:
 - Just open the abovementioned pdf.

What is the expected behavior?
![image](https://user-images.githubusercontent.com/4025991/70468259-13218780-1ac7-11ea-89c6-ae3cf110c9c0.png)
In evince, the standard gnome pdf reader.

What went wrong?
![image](https://user-images.githubusercontent.com/4025991/70468282-1ddc1c80-1ac7-11ea-83aa-73370f295a98.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

Snuffleupagus:
That's a non-embedded "Symbol" font; the file renders correctly in Firefox Nightly 73 on Windows 7.

timvandermeij:
I can also reproduce this using Arch Linux, so since it's a non-embedded font it's most likely OS/configuration specific.

Snuffleupagus:
Duplicate of #4244

PR #13515, opened 2021-06-08
Cache the "raw" standard font data in the worker-thread (PR 12726 follow-up)

Snuffleupagus:
*This implementation is basically a copy of the pre-existing `builtInCMapCache` implementation.*

For some, badly generated, PDF documents it's possible that we'll end up having to fetch the *same* standard font data over and over (which is obviously inefficient).
While not common, it's certainly possible that a PDF document uses *custom* font names where the actual font then references one of the standard fonts; see e.g. issue #11399 for one such example. ***Edit:*** Loading all pages of that document currently causes the `FoxitSymbol.pfb` file to be loaded *thirteen* times.

Note that I did suggest adding worker-thread caching of standard font data in PR #12726, however it wasn't deemed necessary at the time. Now that we have a real-world example that benefit from caching, I think that we should simply implement this now.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -71,6 +71,7 @@ class Catalog {
 
     this.fontCache = new RefSetCache();
     this.builtInCMapCache = new Map();
+    this.standardFontDataCache = new Map();
     this.globalImageCache = new GlobalImageCache();
     this.pageKidsCountCache = new RefSetCache();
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me; thank you for doing this!

PR #13517, opened 2021-06-08
XFA - Add Liberation-Sans font as a substitution for some missing fonts

calixteman:
- Some js files contain scale factors for each glyph in order to rescale Liberation to have a final font with the correct width.   
  - A lot of XFA have some containers where their dimensions are based on their text content, so using default font from browser can lead to an almost unreadable pdf.


src/core/calibri_bold.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-disable max-len */
+/* prettier-ignore */
```
Snuffleupagus, line 17: Here, and elsewhere, can't we just let Prettier autoformat the code rather than disabling that completely?

(Please note that in the *built* files, this formatting isn't kept anyway.)

src/display/display_utils.js
```diff
@@ -199,7 +199,8 @@ class BaseStandardFontDataFactory {
     if (!filename) {
       throw new Error("Font filename must be specified.");
     }
-    const url = this.baseUrl + filename + ".pfb";
+
```
Snuffleupagus, line 202: Nit: Unnecessary new-line.

src/core/standard_fonts.js
```diff
@@ -793,9 +892,14 @@ const getSupplementalGlyphMapForCalibri = getLookupTableFactory(function (t) {
   t[1086] = 45;
 });
 
-function getStandardFontName(name) {
+function getStandardFontName(name, xfa = false) {
```
brendandahl, line 895: Instead of of adding a param, can we make this a different function instead? It returns a different type for xfa fonts and they aren't really standard fonts. Maybe `getXfaFontInfo`?

src/core/standard_fonts.js
```diff
@@ -90,21 +106,104 @@ const getStdFontMap = getLookupTableFactory(function (t) {
   t["TimesNewRomanPSMT-Italic"] = "Times-Italic";
 });
 
+const getStdXFAFontMap = getLookupTableFactory(function (t) {
```
brendandahl, line 109: Same thing there, not realy a std font. `getXFAFontMap`?

src/core/standard_fonts.js
```diff
@@ -90,21 +106,104 @@ const getStdFontMap = getLookupTableFactory(function (t) {
...
+    name: "LiberationSans-Italic",
+    factors: HelveticaItalicFactors,
+  };
+  t["Helvetica-BoldItalic"] = {
+    name: "LiberationSans-BoldItalic",
+    factors: HelveticaBoldItalicFactors,
+  };
+});
+
 const getStdFontNameToFileMap = getLookupTableFactory(function (t) {
```
brendandahl, line 188: Change this to just `getFontNameToFileMap`.

brendandahl (approved):
A few refactors below. We'll probably want to do a bit more in some follow ups to avoid confusing these XFA fonts with the  PDF standard 14 fonts.

src/core/document.js
```diff
@@ -876,13 +876,17 @@ class PDFDocument {
       return;
     }
 
+    const options = Object.assign({}, this.pdfManager.evaluatorOptions);
```
Snuffleupagus, line 879: Nit: `{}` -> `Object.create(null)`

PR #13521, opened 2021-06-09
[XFA] Use `align-items: flex-start;` to improve browser compatibility (issue 13518)

### LINKED ISSUE #13518: Autoprefixer warnings about the partially supported `align-items: start;` CSS rule
After PR #13479 landed, building the PDF.js library now produces warnings from the Autoprefixer package.
For example, with `gulp generic` the following is printed in the console:
```
[21:33:58] gulp-postcss: viewer.css
autoprefixer: C:\Users\Jonas\Git\pdf.js\viewer.css:383:3: start value has mixed support, consider using flex-start instead
autoprefixer: C:\Users\Jonas\Git\pdf.js\viewer.css:393:3: start value has mixed support, consider using flex-start instead
```
with those warnings originating at
 - https://github.com/mozilla/pdf.js/blob/2a6111044e0ad1ee70d6430b073ff39843f7b9d9/web/xfa_layer_builder.css#L121
 - https://github.com/mozilla/pdf.js/blob/2a6111044e0ad1ee70d6430b073ff39843f7b9d9/web/xfa_layer_builder.css#L131

According to MDN, please see https://developer.mozilla.org/en-US/docs/Web/CSS/align-items#browser_compatibility, only Firefox supports `align-items: start;` currently.

@calixteman Can we make the change suggested by Autoprefixer (in which case I'd happily submit a patch), or would that break the layout of XFA documents (in which case I'm guessing this is wontfix)?

Or, could it *possibly* make sense to at least amend the code with the following in those two cases?
```css
align-items: flex-start;
align-items: start;
```

Snuffleupagus:
https://developer.mozilla.org/en-US/docs/Web/CSS/align-items#browser_compatibility

Fixes #13518

### DIFF
--- a/web/xfa_layer_builder.css
+++ b/web/xfa_layer_builder.css
@@ -118,7 +118,7 @@
 .xfaTop {
   display: flex;
   flex-direction: column;
-  align-items: start;
+  align-items: flex-start;
 }
 
 .xfaTop > .xfaCaption {
@@ -128,7 +128,7 @@
 .xfaBottom {
   display: flex;
   flex-direction: column-reverse;
-  align-items: start;
+  align-items: flex-start;
 }
 
 .xfaBottom > .xfaCaption {


calixteman (approved):
Thank you.

PR #13523, opened 2021-06-09
Annotation - Use default values instead of default appearance when printing

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1715476.


marco-c:
@Snuffleupagus would you be interested in finishing this up?

calixteman:
Yep you're right.

PR #13524, opened 2021-06-09
XFA - Don't print arrow in select

calixteman:


web/xfa_layer_builder.css
```diff
@@ -297,4 +297,11 @@
   .xfaSelect {
     background-color: transparent;
   }
+
+  .xfaSelect {
+    -webkit-appearance: none;
+    -moz-appearance: none;
```
Snuffleupagus, line 303: I believe that just `appearance: none;` *should* suffice here, since Autoprefixer will take care of adding the prefixed versions as necessary.

Note e.g. https://github.com/mozilla/pdf.js/blob/b4c96f3bc890a63c6dada9145a0ed397973b4e78/web/annotation_layer_builder.css#L147-L148 which in the *built* file becomes https://github.com/mozilla/pdf.js/blob/8a41c11d4a1a488cfedb14846f4dbd8ab69c02bc/web/viewer.css#L221-L224

Snuffleupagus (approved):
r=me, with the comment addressed.

PR #13525, opened 2021-06-09
[api-minor] Re-factor the `disableFontFace` fallback value, and skip initializing factories with `useWorkerFetch` set

Snuffleupagus:
- [api-minor] Set the `disableFontFace` fallback value directly in the API

   At this point in time, the `apiCompatibilityParams` is essentially unused with the sole exception of the `disableFontFace` handling for Node.js environments.
   Given that `isNodeJS` is a constant now (originally it was a function), we can simply set the correct fallback value for `disableFontFace` directly in the API and clean-up the code a bit here.

 - Don't initialize `CMapReaderFactory`/`StandardFontDataFactory` when the `useWorkerFetch` API option is set

   Given that there's no fallback on the worker-thread, it shouldn't be necessary to initialize `CMapReaderFactory`/`StandardFontDataFactory` when `useWorkerFetch = true` is set.

   Slightly unrelated, but this patch also ensures that the `useSystemFonts` default value only does the `isNodeJS` check in builds where that's actually necessary.


timvandermeij (approved):

timvandermeij:

PR #13527, opened 2021-06-09
XFA - Avoid infinite loop when creating some nodes in data

calixteman:


brendandahl:
[Linux refs failed to update in another PR](https://github.com/mozilla/pdf.js/pull/13517#issuecomment-857884339)

PR #13529, opened 2021-06-09
Bundle the license-file for the Liberation fonts in the builds (PR 13517 follow-up)

Snuffleupagus:
- Bundle the license-file for the Liberation fonts in the builds (PR #13517 follow-up)

   Currently only the Foxit license-file is included, which is most likely just an oversight as far as I can tell.
   Furthermore, to be able the tell the two license-files apart, the Foxit one is also renamed slightly.

 - [gulpfile.js] Reduce unnecessary duplication when bundling CMap/StandardFont files

   Rather than repeating the exact same code in multiple `gulp`-tasks, we can extract it into two helper functions instead.


calixteman (approved):
Thank you for doing that.

PR #13530, opened 2021-06-10
Add a `DOMMatrix` polyfill for Node.js environments (PR 13361 follow-up)

Snuffleupagus:
Given that `DOMMatrix` is, unsurprisingly, not supported in Node.js the `createMatrix` helper function in `src/display/pattern_helper.js` is most likely broken in Node.js environments. It will obviously try to fallback to the `DOMSVGFactory`, however that isn't intended for Node.js usage and errors will be thrown.

Rather than trying to implement a `NodeSVGFactory`, this patch takes easier route of just adding a `DOMMatrix` polyfill using: https://www.npmjs.com/package/dommatrix
This isn't done only for simplicity, but it'll become necessary anyway since the `createMatrix` helper function is only temporary and will be removed in the future.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1380,6 +1380,7 @@ gulp.task("types", function (done) {
     "esModuleInterop",
     "forceConsistentCasingInFileNames",
     "emitDeclarationOnly",
+    "moduleResolution node",
   ].join(" --");
   exec(`"node_modules/.bin/tsc" --${args} src/pdf.js`, done);
 });
--- a/package-lock.json
+++ b/package-lock.json
@@ -2160,6 +2160,45 @@
         "fastq": "^1.6.0"
       }
     },
+    "@rollup/plugin-buble": {
+      "version": "0.21.3",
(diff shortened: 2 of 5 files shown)


gulpfile.js
```diff
@@ -1380,6 +1380,7 @@ gulp.task("types", function (done) {
     "esModuleInterop",
     "forceConsistentCasingInFileNames",
     "emitDeclarationOnly",
+    "skipLibCheck",
```
Snuffleupagus, line 1383: Please note that I've got no idea why adding this "just works", and whether or not it'll impact the generated types negatively, but some quick searching suggested using this and it worked :-)

(This kind of problem, with a cryptic error thrown by `tsc`, is what I was afraid of when support for generation of types was added. At least in *this* case, a simple work-around was at least available.)

brendandahl: Alternatively, I think we can add ` --moduleResolution node`

brendandahl (approved):
r+ with the change below

PR #13540, opened 2021-06-11
XFA - Return html element for the different possible value

### LINKED ISSUE #13536: HSBC XFA "Compte" field not pre-filled
https://web.archive.org/web/20210607145115/https://www.hsbc.fr/content/dam/hsbc/fr/docs/pib/Contestation-Transaction-Carte-Bancaire.pdf
(in the test corpus)

image

image

calixteman:
- it aims to fix #13536.

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -106,6 +106,17 @@ function getRoot(node) {
   return parent;
 }
 
+function valueToHtml(value) {
+  return HTMLResult.success({
+    name: "span",
+    attributes: {
+      class: ["xfaRich"],
+      style: Object.create(null),
+    },
+    value,
+  });
+}
+
 function getTransformedBBox(node) {
   // Take into account rotation and anchor the get the
   // real bounding box.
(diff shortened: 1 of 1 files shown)


src/core/xfa/template.js
```diff
@@ -615,7 +615,14 @@ class BooleanElement extends Option01 {
...
   [$toHTML](availableSpace) {
-    return HTMLResult.success(this[$content] === 1);
+    return HTMLResult.success({
+      name: "span",
+      attributes: {
+        class: ["xfaRich"],
+        style: {},
+      },
+      value: this[$content] === 1,
+    });
```
marco-c, line 625: this code is all very similar between the different objects, what if you moved it to a base class and made all classes inherit this method?
or even just extract the code to a method and call it in all places

src/core/xfa/template.js
```diff
@@ -4782,12 +4820,20 @@ class Time extends StringObject {
   }
 
   [$finalize]() {
-    // TODO
-    this[$content] = new Date(this[$content]);
+    // TODO.
```
marco-c, line 4823: What's left TODO here?

marco-c (approved):

Issue #13213, opened 2021-04-10
CERFA 12156*05 (pdf with forms) asks to be opened with Adobe reader


ggramaize:
Attach (recommended) or Link to PDF file here: 
[cerfa_12156-05.pdf](https://github.com/mozilla/pdf.js/files/6290046/cerfa_12156-05.pdf)

Configuration:
- Web browser and its version: Firefox Nightly 89.0a1 (2021-04-03) (64 bits)
- Operating system and its version: Linux 4.19.0-1-amd64 #1 SMP Debian 4.19.12-1 (2018-12-22)
- PDF.js version: firefox built-in, but reproducible on the [online demo](https://mozilla.github.io/pdf.js/web/viewer.html)
- Is a browser extension: 

Steps to reproduce the problem:
1. Open cerfa_12156-05.pdf with pdf.js

What is the expected behavior? (add screenshot)
![Capture d’écran du 2021-04-10 16-53-58](https://user-images.githubusercontent.com/5450075/114274150-7013b180-9a1d-11eb-821d-a7c620f7f0f8.png)

What went wrong? (add screenshot)
![Capture d’écran du 2021-04-10 16-54-23](https://user-images.githubusercontent.com/5450075/114274162-7ace4680-9a1d-11eb-99e2-8662bc254dbf.png)

calixteman:
The form is a XFA and we're working to support them.
Anyway it's dup of #2373.

PR #13547, opened 2021-06-11
XFA - Handle correctly subformSet

### LINKED ISSUE #13543: Error when loading XFA PDF
PDF from #13213

1. Open the PDF in Firefox

The PDF is loaded and a form is shown.

What went wrong?
The PDF fails to load.

calixteman:
- it aims to avoid to loop forever when opening pdf in #13213;
  - the idea is to consider subformSet as inexistent when running in the tree. So if we've subformA > subformSet > subformB then subformB will be visited as a direct child of subformA.

### DIFF
--- a/src/core/xfa/html_utils.js
+++ b/src/core/xfa/html_utils.js
@@ -16,6 +16,7 @@
 import {
   $extra,
   $getParent,
+  $getSubformParent,
   $nodeName,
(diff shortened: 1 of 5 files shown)


calixteman:
There few errors when rendering the pdf: I'll fix them in some follow-up.

PR #13549, opened 2021-06-12
Remove the `isFetchSupported` function since the Fetch API is available in all supported browsers

Snuffleupagus:
The currently supported browsers, note the minimum versions [listed here](https://github.com/mozilla/pdf.js/blob/5a4e06af2d1f5bf1a4281b4ec3ec18bccd2adebb/gulpfile.js#L78-L88), should now have native support for all of the features checked in the `isFetchSupported` function:

 - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/API/Response#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/API/Body/body#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#browser_compatibility

Hence this function can now be removed, and the code can thus be simplified a little bit.

### DIFF
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -49,7 +49,7 @@ class DOMCanvasFactory extends BaseCanvasFactory {
 async function fetchData(url, asTypedArray = false) {
   if (
     (typeof PDFJSDev !== "undefined" && PDFJSDev.test("MOZCENTRAL")) ||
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #13554, opened 2021-06-13
XFA - Add support for overflow element

calixteman:
- and fix few bugs:
    - avoid infinite loop when layout the document;
    - avoid confusion between break and layout failure;
    - don't add margin width in tb layout when getting available space.


brendandahl:
@calixteman looks like some ref tests failed with this PR

brendandahl:
Wiped out the bad cached file, let's try this again.
/botio-windows test

Snuffleupagus:
@brendandahl It looks like this regressed *at least* `xfa_imm1344e-page1` in all browsers/platforms, and possibly the rest of that file as well, why was this landed as-is?

brendandahl:
I mixed up the refs for that one (thought it was improving). Calixte has a fix in progress, so we'll just re-run with the fix.

PR #13555, opened 2021-06-13
XFA - Value in field can be html

calixteman:
I'll add a ref test later.


marco-c (approved):
> I'll add a ref test later.

In this PR or in a separate one?

calixteman:
In a separate one.

marco-c:
OK, when you do, please make a link with this PR.

PR #13558, opened 2021-06-14
Harden file system access

### LINKED ISSUE #13512: Program files created world writable

Some program files and directories are created world writable. On a multiuser system this may allow an attacker to control behavior by modifying the files (eg a config file) or creating files in the directory (eg in a config directory). On Linux the umask should usually prevent this but it's not guaranteed. [Secure coding standards](https://wiki.sei.cmu.edu/confluence/display/c/FIO06-C.+Create+files+with+appropriate+access+permissions) recommend always specifying a restricted mode so files are secure in any environment. MITRE calls this [CWE-732](https://cwe.mitre.org/data/definitions/732.html), one of the [top 25 most dangerous](https://cwe.mitre.org/top25/archive/2020/2020_cwe_top25.html) vulnerabilities.

Findings are listed here:
https://lgtm.com/query/308122751628928618/
https://lgtm.com/query/8429440007276116881/

[This line](https://lgtm.com/projects/g/mozilla/pdf.js/snapshot/3525d09d63ce145bf6745daa2bc4346859b4017b/files/external/builder/builder.js?sort=name&dir=ASC&mode=heatmap#L197) in the preprocessor seems to output code world writable, possibly allowing an attacker to inject their own code. If any results in the gulpfile are creating config resources, they could allow an attacker to control behavior after install.

This came out of research with CodeQL.

Program files are created with secure permissions.

ghost:
Defaults all file system functions to secure modes when no mode is provided.

I handled this with a wrapper layer because it seemed like the least disruptive way to get it everywhere. If you'd prefer isolated functions I can update.

```js
const fs = require("./src/shared/fs.js");
```

Snapshots produced by `makeref` now use tightened permissions.

Before:

```
$ ls -l test/ref/linux/firefox/*
test/ref/linux/firefox/160F-2019:
total 152
-rw-rw-r-- 1 user user 153998 Jun 12 05:08 1.png

test/ref/linux/firefox/20130226130259:
total 144
-rw-rw-r-- 1 user user 144916 Jun 12 05:08 1.png

test/ref/linux/firefox/High-Pressure-Measurement-WP-001287:
total 92
-rw-rw-r-- 1 user user 92665 Jun 12 05:08 3.png
..
```

After:

```
$ ls -l test/ref/linux/firefox/*
test/ref/linux/firefox/160F-2019:
total 152
-rw------- 1 user user 153998 Jun 14 01:18 1.png

test/ref/linux/firefox/20130226130259:
total 144
-rw------- 1 user user 144916 Jun 14 01:18 1.png

test/ref/linux/firefox/High-Pressure-Measurement-WP-001287:
total 92
-rw------- 1 user user 92665 Jun 14 01:18 3.png
..
```

Build seems to apply it only partially, I think because some internal `gulp` functions are doing creations with default mode.

```
$ ls -l build/dist/build
total 9284
-rw-rw-r-- 1 user user  392307 Jun 14 02:17 pdf.js
-rw-rw-r-- 1 user user  799981 Jun 14 02:18 pdf.js.map
-rw------- 1 user user  219299 Jun 14 02:20 pdf.min.js
-rw-rw-r-- 1 user user  628758 Jun 14 02:17 pdf.sandbox.js
-rw-rw-r-- 1 user user  541393 Jun 14 02:17 pdf.sandbox.js.map
-rw------- 1 user user  624659 Jun 14 02:20 pdf.sandbox.min.js
-rw-rw-r-- 1 user user     693 Jun 12 23:40 pdf.worker.entry.js
-rw-rw-r-- 1 user user 1791297 Jun 14 02:18 pdf.worker.js
-rw-rw-r-- 1 user user 3520000 Jun 14 02:18 pdf.worker.js.map
-rw------- 1 user user  966460 Jun 14 02:20 pdf.worker.min.js
```

Closes #13512.

### DIFF
--- a/examples/node/pdf2png/pdf2png.js
+++ b/examples/node/pdf2png/pdf2png.js
@@ -15,7 +15,7 @@
 
 const Canvas = require("canvas");
 const assert = require("assert").strict;
-const fs = require("fs");
+const fs = require("../../../src/shared/fs.js");
 
 function NodeCanvasFactory() {}
 NodeCanvasFactory.prototype = {
--- a/examples/node/pdf2svg.js
+++ b/examples/node/pdf2svg.js
@@ -5,7 +5,7 @@
 // Node tool to dump SVG output into a file.
 //
 
-const fs = require("fs");
+const fs = require("../../src/shared/fs.js");
 const util = require("util");
 const path = require("path");
 const stream = require("stream");
--- a/external/builder/builder.js
+++ b/external/builder/builder.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const fs = require("fs"),
+const fs = require("../../src/shared/fs.js"),
   path = require("path"),
   vm = require("vm");
 
--- a/external/builder/preprocessor2.js
+++ b/external/builder/preprocessor2.js
@@ -3,7 +3,7 @@
 const acorn = require("acorn");
 const escodegen = require("escodegen");
 const vm = require("vm");
-const fs = require("fs");
+const fs = require("../../src/shared/fs.js");
 const path = require("path");
 
 const PDFJS_PREPROCESSOR_NAME = "PDFJSDev";
--- a/external/builder/test-fixtures.js
+++ b/external/builder/test-fixtures.js
@@ -1,7 +1,7 @@
 "use strict";
 
 const builder = require("./builder.js");
-const fs = require("fs");
+const fs = require("../../src/shared/fs.js");
 const path = require("path");
 
 let errors = 0;
--- a/external/builder/test-fixtures_esprima.js
+++ b/external/builder/test-fixtures_esprima.js
@@ -1,7 +1,7 @@
 "use strict";
 
 const p2 = require("./preprocessor2.js");
-const fs = require("fs");
+const fs = require("../../src/shared/fs.js");
 const path = require("path");
 
 let errors = 0;
--- a/external/cmapscompress/compress.js
+++ b/external/cmapscompress/compress.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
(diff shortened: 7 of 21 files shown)


Snuffleupagus (changes requested):
Sorry, but I don't think that this is a desirable change as-is (or possibly at all) for a number of reasons:
 - First of all, this type of code *really does not* belong in the `src/` folder. A more appropriate place would be somewhere in `external/` instead.
 - Issue #13512 talks about `fs.writeFileSync` in particular, however this PR seems to basically "re-implement" a significant chunk of the entire Node.js `fs` module. Note also https://github.com/mozilla/pdf.js/issues/13512#issuecomment-857975772: 
    > [...] most likely by making a simple wrapper function like we already have in the test utilities file.

    Doing this adds, as far as I'm concerned, way too much of a maintenance burden for the core PDF.js contributors, since we'd need to *manually* ensure that this code is kept up-to-date with any changes in future Node.js versions.
 - There's nothing here which prevents *direct* `fs`-usage being re-introduced in the future, and enforcing this thus puts additional load on reviewers by forcing them to remember to *manually* check any `fs` accesses in new code.
 - > Build seems to apply it only partially, I think because some internal gulp functions are doing creations with default mode.

    The fact that, even with this patch, the intended file permissions apparently still aren't applied *everywhere* provides even less of an incentive for this type of change (in my opinion). 
 - As mentioned in the issue, this can be seen as a Node.js "issue" rather than a PDF.js one, and please note that the PDF.js library is *primarily* intended for usage in *browsers* where this shouldn't matter.


All-in-all, I'm sorry to say that this *unfortunately* doesn't look like a change we want to make here.

dittyroma:
Thank you for considering.

PR #13560, opened 2021-06-14
XFA - Add the possibily to layout and measure text

### LINKED ISSUE #13500: XFA use of CDATA in exData section causes unhandled exception
[Acrobat.pdf](https://github.com/mozilla/pdf.js/files/6602628/Acrobat.pdf)

Configuration:
- Web browser and its version: Firefox 89.0 x64
- Operating system and its version: Windows 10.0.19043.985
- PDF.js version: pr/13479 at ec1dd06c
- Is a browser extension: No

1. enable XFA support
2. Open Acrobat.pdf

PDF renders 
image

Receive console error. PDF is not loaded
```
TypeError: this.exData[$content][$text] is not a function
```
This appears to be caused by the use of `CDATA` in `exData` tags containing XHTML content. For example:
```xml
<exData contentType="text/html"><![CDATA[<body xfa:APIVersion="2.7.0.0" xmlns="http://www.w3.org/1999/xhtml" xmlns:xfa="http://www.xfa.org/schema/xfa-data/1.0/"><p style="letter-spacing:0in;text-align:left"><span style="xfa-spacerun:yes;font-family:Arial Black;kerning-mode:pair;font-size:8.742pt;color:#e44b25"> </span></p><p style="letter-spacing:0in;text-align:left"><span style="font-family:Arial Black;kerning-mode:pair;font-size:8.742pt;color:#e44b25;xfa-spacerun:yes">CLIENT INFORMATION</span></p></body>]]></exData>
```
(Removal of CDATA wrapper in a simpler example made the error go away)

calixteman:
- some containers doesn't always have their 2 dimensions and those dimensions re based on contents;
  - so in order to measure text, we must get the glyph widths (for the xfa fonts) before starting the layout;
  - implement a word-wrap algorithm;
  - handle font change during text layout.

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -187,13 +187,8 @@ class WorkerMessageHandler {
         await pdfManager.ensureDoc("checkFirstPage");
       }
 
-      const [numPages, fingerprint, htmlForXfa] = await Promise.all([
-        pdfManager.ensureDoc("numPages"),
-        pdfManager.ensureDoc("fingerprint"),
-        pdfManager.ensureDoc("htmlForXfa"),
-      ]);
-
-      if (htmlForXfa) {
+      const isPureXfa = await pdfManager.ensureDoc("isPureXfa");
+      if (isPureXfa) {
         const task = new WorkerTask("loadXfaFonts");
         startWorkerTask(task);
         await pdfManager
@@ -203,6 +198,17 @@ class WorkerMessageHandler {
           })
           .then(() => finishWorkerTask(task));
       }
+
+      const [numPages, fingerprint] = await Promise.all([
+        pdfManager.ensureDoc("numPages"),
+        pdfManager.ensureDoc("fingerprint"),
+      ]);
+
+      // Get htmlForXfa after numPages to avoid to create HTML twice.
+      const htmlForXfa = isPureXfa
+        ? await pdfManager.ensureDoc("htmlForXfa")
+        : null;
+
       return { numPages, fingerprint, htmlForXfa };
     }
 
--- a/src/core/xfa/factory.js
+++ b/src/core/xfa/factory.js
@@ -13,8 +13,9 @@
  * limitations under the License.
  */
 
-import { $toHTML } from "./xfa_object.js";
+import { $fonts, $toHTML } from "./xfa_object.js";
 import { Binder } from "./bind.js";
+import { warn } from "../../shared/util.js";
 import { XFAParser } from "./parser.js";
 
 class XFAFactory {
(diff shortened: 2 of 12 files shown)


src/core/xfa/factory.js
```diff
@@ -22,28 +22,61 @@ class XFAFactory {
...
-      return [0, 0, parseInt(width), parseInt(height)];
-    });
+    try {
+      this.pages = this.form[$toHTML]();
+      this.dims = this.pages.children.map(c => {
+        const { width, height } = c.attributes.style;
+        return [0, 0, parseInt(width), parseInt(height)];
+      });
+    } catch (e) {
+      console.log(e);
```
Snuffleupagus, line 42: Please use `warn` instead, since we should never use `console.log(...)` *directly* in code residing in the `src/` folder.

marco-c:
I'm assuming this is going to fix #13500?

marco-c:
There seems to be a regression at least in firefox-hsbc_closure-page1.

calixteman:
There are still an issue with `hsbc_closure-page3` so I slightly increase the width factor from 1.1 to 1.2.

marco-c:
> There is still an issue with `hsbc_closure-page3` so I slightly increase the width factor from 1.1 to 1.2.

There's still a minor issue ("IBAN:" the ":" is on the following line)

calixteman:
I don't understand why the issue is still there (`hsbc_closure-page3`)...
Normally the font used to render the "Non" (bottom left) is Arial and is embedded in the pdf.
And on my screen at least the "o" is almost a circle when I load the pdf locally when it isn't really a circle on the images reftest.
So maybe something is wrong with fonts in xfa reftests, I need to investigate.

calixteman:
And of course I cannot reproduce locally, it seems that the font is correct.
The "O" are not the same from this png I generated locally.

calixteman:
Finally tests are ok...

src/core/worker.js
```diff
@@ -203,6 +198,17 @@ class WorkerMessageHandler {
           })
           .then(() => finishWorkerTask(task));
       }
+
+      const [numPages, fingerprint] = await Promise.all([
+        pdfManager.ensureDoc("numPages"),
+        pdfManager.ensureDoc("fingerprint"),
+      ]);
+
+      // Get htmlForXfa after numPages to avoid to create HTML 2 times.
```
marco-c, line 207: Nit: 
```suggestion
      // Get htmlForXfa after numPages to avoid to create HTML twice.
```

src/core/xfa/template.js
```diff
@@ -2622,7 +2660,7 @@ class Font extends XFAObject {
     ]);
     this.posture = getStringOption(attributes.posture, ["normal", "italic"]);
     this.size = getMeasurement(attributes.size, "10pt");
-    this.typeface = attributes.typeface || "";
+    this.typeface = attributes.typeface || "Myriad Pro";
```
marco-c, line 2663: Why "Myriad Pro" as the default?

marco-c (approved):

PR #13563, opened 2021-06-15
Add basic linting of JSON files using `eslint-plugin-json`

Snuffleupagus:
By adding basic linting of JSON files, we can ensure that they're actually valid and prevent e.g. test-failures caused by *accidental* errors when editing the `test/test_manifest.json` file (something that I've done *many* times myself).

For now this simply uses the `recommended` configuration, but we can obviously tweak this later if/when needed. Please find additional information at https://github.com/azeemba/eslint-plugin-json


calixteman (approved):
Useful stuff: thanks.

PR #13566, opened 2021-06-16
XFA - Fix layout issues

### LINKED ISSUE #13567: xfa_imm1344e-page1 regression after #13554
This issue is tracking the regression mentioned in https://github.com/mozilla/pdf.js/pull/13554#issuecomment-861661644.

calixteman:
- PR #13554 is buggy, so this patch aims to fix bugs.
  - check if a component fits into its parent in taking into account the parent layout.
  - introduce method isSplittable for template nodes to know if a component can be splitted in case of overflow.

### DIFF
--- a/src/core/xfa/html_utils.js
+++ b/src/core/xfa/html_utils.js
@@ -102,24 +102,12 @@ const converters = {
       style.width = measureToString(width);
     } else {
       style.width = "auto";
-      if (node.maxW > 0) {
-        style.maxWidth = measureToString(node.maxW);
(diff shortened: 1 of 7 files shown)


calixteman:
Something is wrong on `firefox-xfa_imm1295e-page3`.

marco-c (approved):
Can you make sure the differences in `firefox-xfa_annual_voting_survey` are expected?
And can you double check the differences in `firefox-xfa_imm1295e-page2`?

PR #13569, opened 2021-06-16
XFA - Container wrapper must take the visibility of the wrapped content

calixteman:


calixteman:
We already have a ref test for this case: `imm1295`.

marco-c (approved):

PR #13570, opened 2021-06-16
XFA - By default a text ui has only one line when in a field element

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1716809.


calixteman:
The reftest is not useful for the issue this patch aims to fix but I added it to have a good corpus.

marco-c (approved):

PR #13573, opened 2021-06-17
XFA - Fix the way to select page on breaking

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1716838.
  - some fonts in the pdf in the bug where bold when they shouldn't so write the font properties in the html to avoid to use some wrong inherited ones.

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -19,6 +19,7 @@ import {
   $appendChild,
   $childrenToHTML,
   $clean,
+  $cleanPage,
   $content,
   $extra,
   $finalize,
@@ -37,6 +38,7 @@ import {
   $isCDATAXml,
   $isSplittable,
   $isTransparent,
+  $isUsable,
   $namespaceId,
   $nodeName,
   $onChild,
@@ -2396,8 +2398,10 @@ class Field extends XFAObject {
 
     const caption = this.caption ? this.caption[$toHTML]().html : null;
     if (!caption) {
-      // Even if no caption this class will help to center the ui.
-      ui.attributes.class.push("xfaLeft");
+      if (ui.attributes.class) {
+        // Even if no caption this class will help to center the ui.
+        ui.attributes.class.push("xfaLeft");
+      }
       return HTMLResult.success(createWrapper(this, html), bbox);
     }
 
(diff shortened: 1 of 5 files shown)


brendandahl:
Looks like there's regressions on imm1344e and imm1295e.

calixteman:
The diffs on `imm1344e` and `imm1295e` are expected: the first pageArea has a max set to 1 so the other pageArea must be used for other pages.

marco-c:
> The diffs on imm1344e and imm1295e are expected: the first pageArea has a max set to 1 so the other pageArea must be used for other pages.

In page 2 and 4 of imm1295e there seems to be a regression. The header and footer of page 2 and page 4 should be like on page 5.

calixteman:
There are some diff with `xfa_bug1716816.pdf` but it seems that this pdf is wrongly displayed (even before the patch).
I'll fix this pdf later.

marco-c (approved):

Snuffleupagus:
Sorry, but why was this landed with a bunch of breaking error (which is also why `makeref` failed) such as e.g. the following in multiple tests!? 
```
TEST-UNEXPECTED-FAIL | test failed xfa_candidate_petitions | in chrome | page1 round 1 | Loading PDF document: UnknownErrorException: Cannot read property 'length' of undefined
```

This should probably be backed out, and then re-landed once fixed, to prevent testing being completely busted in the meantime!
(I'd suggest looking at not just the reference images, but *also* the actual test output when e.g. reviewing :-P)

calixteman:
> (I'd suggest looking at not just the reference images, but _also_ at the actual test output when e.g. reviewing :-P)
It's the reason why I missed the failures... sorry for that.

Snuffleupagus:
Since it's currently way too easy to *accidentally* overlook breaking errors in the test-output, we might want to include more information directly in the botio output; in case it's deemed helpful, I've just submitted https://github.com/mozilla/botio-files-pdfjs/pull/30

marco-c:
> Since it's currently way too easy to accidentally overlook breaking errors in the test-output, we might want to include more information directly in the botio output

Totally agree!

PR #13575, opened 2021-06-17
XFA - Fix reftest for xfa_issue13500

calixteman:


calixteman:
My bad, sorry for that.

marco-c (approved):

PR #13577, opened 2021-06-17
Remove the internal `PDFScriptingManager._pageEventsReady` boolean (PR 13074 follow-up)

Snuffleupagus:
With the introduction of `PDFScriptingManager._closeCapability` in PR #13074, the pre-existing `PDFScriptingManager._pageEventsReady` boolean essentially became redundant.
Given that you always want to avoid tracking closely related state *separately*, since it's easy to introduce subtle bugs that way, we should just remove `PDFScriptingManager._pageEventsReady` now.

Obviously I *should* have done this already back in PR #13074, sorry about the churn here!


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13587, opened 2021-06-19
Set the default value of `useSystemFonts` correctly, depending on `disableFontFace`, in the API (PR 13516 follow-up)

Snuffleupagus:
*Sorry about the churn here, since the change that I made in PR #13516 was not very smart.*

With the current code, it's now *impossible* for a user to actually control the `useSystemFonts` option manually. To prevent outright breakage we obviously still need to default to setting `useSystemFonts = false` when `disableFontFace === true`, however that should be possible for an API consumer to override.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -397,16 +397,14 @@ class PartialEvaluator {
       return new Stream(cachedData);
     }
 
-    if (!this.options.disableFontFace) {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

PR #13588, opened 2021-06-19
Add a (basic) TypeScript definitions for the viewer components (issue 13267)

### LINKED ISSUE #13267: Type declaration are missing in prebuilt files for the viewer API
The prebuilt files include `web/pdf_viewer.js` with some viewer APIs. Those are missing from type declarations.

Snuffleupagus:
Please note that while the `gulp types`/`gulp typestest` tasks (obviously) still work with this patch, I've got no idea if the output is first of all even useable and secondly if it's actually useful in practice.
However, in the interest of closing some (seemingly simple) issues, I suppose that this probably shouldn't hurt (and we'd need TypeScript users to help improve things here).

Fixes #13267

### DIFF
--- a/external/dist/legacy/web/pdf_viewer.d.ts
+++ b/external/dist/legacy/web/pdf_viewer.d.ts
@@ -1,0 +1,1 @@
+export * from "pdfjs-dist/types/web/pdf_viewer.component.d.ts";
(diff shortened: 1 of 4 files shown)


systemespectr (approved):

timvandermeij (approved):

timvandermeij:
Let's add this; thanks!

PR #13590, opened 2021-06-20
Stop encoding the value in the `DOMElement.setAttribute` method (issue 8558)

### LINKED ISSUE #8558: node svg example double-encodes element attributes
https://github.com/mozilla/pdf.js/blob/9bed695ebde488b49d39290087f72ba224bd1078/examples/node/domstubs.js#L84-L87
```javascript
  setAttributeNS: function DOMElement_setAttributeNS(NS, name, value) {
    value = value || '';
    value = xmlEncode(value);
    this.attributes[name] = value;
```

https://github.com/mozilla/pdf.js/blob/9bed695ebde488b49d39290087f72ba224bd1078/examples/node/domstubs.js#L105

```javascript
      buf.push(' ' + i + '="' + xmlEncode(this.attributes[i]) + '"');
```

One of the two should be removed. Most likely the first one, because `getAttributeNS`'s implementation just returns `this.attributes[i]`, and that should not be encoded.

Snuffleupagus:
This patch is an attempt at closing an old, and seemingly trivial, issue and the SVG-files created by the `pdf2svg.js` examples still appear to work just fine when opened in browsers (tested with Firefox Nightly and Google Chrome Beta).

Fixes #8558

### DIFF
--- a/examples/node/domstubs.js
+++ b/examples/node/domstubs.js
@@ -93,9 +93,7 @@ DOMElement.prototype = {
   },
 
   setAttribute: function DOMElement_setAttribute(name, value) {
-    value = value || "";
-    value = xmlEncode(value);
-    this.attributes[name] = value;
+    this.attributes[name] = value || "";
   },
 
   setAttributeNS: function DOMElement_setAttributeNS(NS, name, value) {


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #13591, opened 2021-06-20
XFA - Match font family correctly

calixteman:
- partial fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1716980;
  - some pdf can contain an invalid font family (e.g. 'Windings 3') so in this case remove the space;
  - the font family in typeface attribute doesn't always match the one defined in the FontDescriptor dictionary.

### DIFF
--- a/src/core/xfa/fonts.js
+++ b/src/core/xfa/fonts.js
@@ -1,0 +1,157 @@
+/* Copyright 2021 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


src/core/xfa/fonts.js
```diff
@@ -0,0 +1,161 @@
...
+            .startsWith(name)
+        ) {
+          maybe.push(pdfFont);
+        }
+      }
+    }
+
+    if (maybe.length >= 1) {
+      if (maybe.length !== 1) {
+        if (mustWarn) {
```
Snuffleupagus, line 143: I'm guessing that *one* if, i.e. `if (maybe.length !== 1 && mustWarn) {`, would suffice here?

src/core/xfa/fonts.js
```diff
@@ -0,0 +1,161 @@
...
+          warn(
+            `XFA - Too much choices to guess the the correct font: ${fontName}`
```
Snuffleupagus, line 145: Grammar-nits: `much` -> `many` and `the the` -> `the`.

marco-c (approved):

PR #13592, opened 2021-06-20
XFA - Don't display print-only elements

calixteman:
- partial fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1716980.

### DIFF
--- a/web/xfa_layer_builder.css
+++ b/web/xfa_layer_builder.css
@@ -13,6 +13,10 @@
  * limitations under the License.
  */
 
+.xfaPrintOnly {
+  display: none;
+}
+
 .xfaLayer {
   position: absolute;
   top: 0;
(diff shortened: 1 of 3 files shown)


web/xfa_layer_builder.css
```diff
@@ -13,6 +13,10 @@
  * limitations under the License.
  */
 
+.xfaPrintOnly {
+  display: none;
```
marco-c, line 17: Shouldn't these elements be displayed when the user wants to print the PDF?

calixteman: I need to check if it's relevant or not in general. For example in the pdf in the above bug, I'm not sure if it's correct correct to print these elements when maybe they should be dynamically modified.
So for now we just print what the user see.

marco-c: OK, can you file an issue about this so we don't forget?

marco-c (approved):

PR #13594, opened 2021-06-20
XFA - Add margins if needed after having layout some text

calixteman:
- it aims to fix the text misplacement found in https://bugzilla.mozilla.org/show_bug.cgi?id=1716380.

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -1457,6 +1457,13 @@ class Draw extends XFAObject {
     fixDimensions(this);
 
     if ((this.w === "" || this.h === "") && this.value) {
+      let marginH = 0;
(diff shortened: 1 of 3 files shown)


marco-c (approved):
Can you run the reftests and ensure all changes are OK?

calixteman:
> Can you run the reftests and ensure all changes are OK?

My bad, I thought we already have the file in the bug in the suite.

brendandahl (approved):
r+ if you're fine with the changes in `xfa_issue13500`. They seem mostly better, but I'm not sure if you're expecting that.

calixteman:
I think the diff is correct but this pdf has other errors I need to fix.

PR #13600, opened 2021-06-21
Take the position of the `selected` element into account when scrolling matches (issue 13596)

### LINKED ISSUE #13596: Word search does not scroll a document to the exact text location

Configuration:
- Web browser and its version: **Chrome 91.0.4472.106**
- Operating system and its version: **Windows 10**
- PDF.js version: **Same as in your official demo** https://mozilla.github.io/pdf.js/web/viewer.html
- Is a browser extension: **No**

1. Open your official demo https://mozilla.github.io/pdf.js/web/viewer.html
2. Zoom a document in a way right side of the document is not visible
3. Scroll the document to the left top
4. Perfrom a search by the word `expressions`

Document is scrolled in a way the word `expressions` is visible

Document is scrolled only to the line with the word `expressions`. But the word `expressions` remains invisible

image

Snuffleupagus:
Note that as far as I can tell, this is *not* a regression but rather a bug which has existed since basically "forever".

**In order to reproduce this easily:**
 - Open the viewer.
 - Set the zoom level to `400%`,
 - Search for "expression".

The problem here is that when scrolling matches into view, we're scrolling to the start of the *containing* `textLayer` element rather than the start of the highlighted match itself.[1] When the entire width (or at least most) of the page is visible in the viewer, that doesn't really matter though which is likely why this bug has gone unnoticed for so long.[2]
Given that the highlighted match can be placed anywhere, e.g. even at the very end, within its `textLayer` element it's quite easy to see why the current implementation becomes a problem at higher zoom levels. All of this is then *further* exacerbated by `PDFFindController.scrollMatchIntoView` using a negative left offset, to ensure that the current match has some (visible) context available once scrolled into view.

In order to address this long-standing bug, we'll determine the (left) offset of the `selected` match and use that to modify the final position scrolled to in `PDFFindController.scrollMatchIntoView` such that the match is visible regardless of zoom level.

---
[1] Unfortunately we cannot directly scroll to the `selected` match, since it's not absolutely positioned and changing that would cause other bugs/regressions (note recent patches in that area).

[2] I did actually stumble upon this problem a little while ago, while working on PR 13482, but forgot to look into this again until I saw the new issue.

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -210,7 +210,12 @@ class PDFFindController {
     });
   }
 
-  scrollMatchIntoView({ element = null, pageIndex = -1, matchIndex = -1 }) {
+  scrollMatchIntoView({
+    element = null,
+    selectedLeft = 0,
+    pageIndex = -1,
+    matchIndex = -1,
+  }) {
     if (!this._scrollMatches || !element) {
       return;
     } else if (matchIndex === -1 || matchIndex !== this._selected.matchIdx) {
@@ -222,7 +227,7 @@ class PDFFindController {
 
     const spot = {
       top: MATCH_SCROLL_OFFSET_TOP,
-      left: MATCH_SCROLL_OFFSET_LEFT,
+      left: selectedLeft + MATCH_SCROLL_OFFSET_LEFT,
     };
     scrollIntoView(element, spot, /* scrollMatches = */ true);
   }
--- a/web/text_layer_builder.js
+++ b/web/text_layer_builder.js
@@ -227,7 +227,7 @@ class TextLayerBuilder {
     function beginText(begin, className) {
       const divIdx = begin.divIdx;
       textDivs[divIdx].textContent = "";
-      appendTextToDiv(divIdx, 0, begin.offset, className);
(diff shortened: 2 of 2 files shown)


timvandermeij:
I can't reproduce this issue using Firefox on Linux, so I'm wondering if this is somehow Windows-specific? The steps to reproduce above and in the issue itself don't lead to the word "expression" ever being out of view for me. Once the word is found, the document is scrolled to the right position, even when it's out of view... :-s

Snuffleupagus:
> so I'm wondering if this is somehow Windows-specific?
> [...]
> This is right after I opened the viewer (on automatic zoom) and changed it to 400%:

I don't think it's Windows specific, but looking at your screen-shot (which is very helpful) I've got an idea that might explain this :-)

Most likely it's *also* connected to the width of the browser window, since that obviously *also* affects things. Could you perhaps try to increase the zoom level even more, to e.g. `1000%`, and/or reduce the width of the browser window?

timvandermeij (approved):

timvandermeij:
I tried again with a much smaller window and that worked; thanks!

PR #13603, opened 2021-06-21
Revert "Revert "XFA - Fix the way to select page on breaking"" 

calixteman:
- and fix the error which caused the backout: add an $extra property when creating html.
  - switch to next content area when breaking on page area.


marco-c:
@calixteman there seems to be some regressions

marco-c (approved):

brendandahl:
killed the bots accidentally
/botio makeref

PR #13604, opened 2021-06-21
XFA - A prototype can have a property which needs itself to resolve a proto

calixteman:


marco-c:
The changes in `xfa_bug1716816` all seem improvements.

marco-c (approved):

PR #13606, opened 2021-06-21
Correctly align the `textLayer` content with horizontal/spread scrolling modes (issue 13605)

Snuffleupagus:
This is *very* similar to PR 12848, however for this `textLayer`-case it appears to only be an issue in documents with marked content.

PR #13609, opened 2021-06-22
Correctly align the `xfaLayer` content with horizontal scrolling/spread modes

Snuffleupagus:
This extends the approach in PRs #12848 and #13606 to also apply to the `xfaLayer`, since otherwise XFA forms will be similarly broken in most non-default scroll/spread modes.


calixteman (approved):
Thanks for doing this.

PR #13612, opened 2021-06-22
Support corrupt documents with *empty* `Name`-entries (issue 13610)

### LINKED ISSUE #13610: Header line of document wasn't showed
[Unitfactuur_18707277_KORTELAND_PRINTEN&INSCANNEN.pdf](https://github.com/mozilla/pdf.js/files/6694277/Unitfactuur_18707277_KORTELAND_PRINTEN.INSCANNEN.pdf)

Configuration:
Web browser and its version:  Chrome Version 91.0.4472.106
- PDF.js version: latest (that is used in demo)
- Is a browser extension: no

1.  Upload attached document to demo 

What is the expected behavior? Header of document is showed
image

What went wrong? Header of document isn't showed, some warnings were showed in console
Warning: getOperatorList - ignoring XObject: "FormatError: XObject must be referred to by name.".
util.js:369 
Warning: getTextContent - ignoring XObject: "FormatError: XObject must be referred to by name.".
image

Snuffleupagus:
Apparently some really bad PDF software can create documents with *empty* `Name`-entries, which we thus need to somehow deal with.
While I don't know if this patch is necessarily the best solution, it should at least ensure that the *empty* `Name`-instance cannot accidentally match a proper `Name`-instance (and it doesn't require changes to a lot of existing code).[1]

---
[1] I briefly considered using a `Symbol` rather than an Object, but quickly decided against that since the former one [is not clonable](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types) and `Name`-instances may be sent to the API.

### DIFF
--- a/src/core/primitives.js
+++ b/src/core/primitives.js
@@ -33,6 +33,12 @@ const Name = (function NameClosure() {
       return nameValue ? nameValue : (nameCache[name] = new Name(name));
     }
 
+    static get empty() {
+      // eslint-disable-next-line no-restricted-syntax
+      const emptyName = new Name({ empty: true });
+      return shadow(this, "empty", emptyName);
+    }
+
     static _clearCache() {
       nameCache = Object.create(null);
     }
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -1113,6 +1113,9 @@ class Lexer {
     }
     if (strBuf.length > 127) {
       warn(`Name token is longer than allowed by the spec: ${strBuf.length}`);
+    } else if (strBuf.length === 0) {
+      warn("Name token is empty.");
+      return Name.empty;
     }
     return Name.get(strBuf.join(""));
   }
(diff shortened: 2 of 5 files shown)


Snuffleupagus:
For once I've not run the entire test-suite locally, since it's already hot enough around here, so hopefully there won't be any regressions with this patch :-)

/botio test

src/core/primitives.js
```diff
@@ -33,6 +33,12 @@ const Name = (function NameClosure() {
       return nameValue ? nameValue : (nameCache[name] = new Name(name));
     }
 
+    static get empty() {
+      // eslint-disable-next-line no-restricted-syntax
+      const emptyName = new Name({ empty: true });
+      return shadow(this, "empty", emptyName);
+    }
```
Snuffleupagus, line 40: The other possible solution, which I suppose could be seen as overall preferable, would be just remove [checks such as this one](https://github.com/mozilla/pdf.js/blob/9441245320ba80b91cc9e94da5205a394a165f7b/src/core/evaluator.js#L1571-L1573) and thus allow empty `Name`s in the "normal" way!?

timvandermeij (approved):

timvandermeij:
Let's do this; thanks!

/botio makeref

PR #13614, opened 2021-06-23
XFA - Get line height from the font

calixteman:
- when the CSS line-height property is set to 'normal' then the value depends of the user agent. So use a line height based on the font itself and if for any reasons this value is not available use 1.2 as default.
  - it's a partial fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1717681.

### DIFF
--- a/src/core/xfa/fonts.js
+++ b/src/core/xfa/fonts.js
@@ -154,4 +154,17 @@ class FontFinder {
   }
 }
 
-export { FontFinder };
+function selectFont(xfaFont, typeface) {
+  if (xfaFont.posture === "italic") {
+    if (xfaFont.weight === "bold") {
+      return typeface.bolditalic;
+    }
+    return typeface.italic;
+  } else if (xfaFont.weight === "bold") {
+    return typeface.bold;
+  }
+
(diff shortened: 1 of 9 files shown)


calixteman:
Almost all the xfa reftests should failed but overall I'd say it's an improvement.

/botio test

src/core/xfa/fonts.js
```diff
@@ -154,4 +154,17 @@ class FontFinder {
...
 }
 
-export { FontFinder };
+function selectFont(xfaFont, typeface) {
+  if (xfaFont.posture === "italic") {
+    if (xfaFont.weight === "bold") {
+      return typeface.bolditalic;
+    }
+    return typeface.italic;
+  } else if (xfaFont.weigth === "bold") {
```
Snuffleupagus, line 163: `weigth` -> `weight`

marco-c:
There is a regression in `firefox-xfa_imm1295e-page2` and `firefox-xfa_imm1344e-page2`.

marco-c (approved):
Reftests are looking good now.

PR #13616, opened 2021-06-23
Add PDFs from #1773 to the reftests

### LINKED ISSUE #1773: PDF doesn't display anything, complains about my pdf viewer

marco-c:


test/pdfs/MBTA-pretax-form-July2012.pdf.link
```diff
@@ -0,0 +1 @@
+https://web.archive.org/web/20121105185256if_/http://www.northeastern.edu/hrm/pdfs/resources/benefits/MBTA-pretax-form-July2012.pdf
```
Snuffleupagus, line 1: The `if_` part should be removed here, since that will be added *automatically* by https://github.com/mozilla/pdf.js/blob/master/test/downloadutils.js

Furthermore, please add a blank line at the end of the file (to avoid the big red marker here in the Github diff).

test/pdfs/xfa_imm5257e.pdf.link
```diff
@@ -0,0 +1 @@
+https://web.archive.org/web/20210521163640/https://www.canada.ca/content/dam/ircc/migration/ircc/english/pdf/kits/forms/imm5257e.pdf
```
Snuffleupagus, line 1: Please add a blank line at the end of the file (to avoid the big red marker here in the Github diff).

Snuffleupagus (commented):
Are both of these new test-cases actually checking *specific* things not already covered by existing tests?

test/pdfs/MBTA-pretax-form-July2012.pdf.link
marco-c: Fixed.

test/pdfs/xfa_imm5257e.pdf.link
marco-c: Fixed.

marco-c:
> Are both of these new test-cases actually checking specific things not already covered by existing tests?

There is no easy way to tell without code coverage information, but since we still have relatively few reftests for XFA, I'm inclined to think these are likely to cover at least something that isn't already covered.

calixteman:
And layout stuff is pretty touchy so having more reftests will help to avoid future layout regressions.

PR #13617, opened 2021-06-23
Reset the `IPDFLinkService.externalLinkEnabled` property on document closing

Snuffleupagus:
Given that this property is only used with password protected documents, and is consequently document-specific rather than viewer-specific, ensure that `IPDFLinkService.externalLinkEnabled` is actually being reset by `PDFViewerApplication.close`.

To make things less confusing/inconsistent, remove the *undocumented* `externalLinkEnabled` property from the `PDFLinkService` constructor and force it to always be manually set when needed.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -828,6 +828,7 @@ const PDFViewerApplication = {
       this.pdfDocumentProperties.setDocument(null);
     }
     webViewerResetPermissions();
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #13619, opened 2021-06-23
XFA - Add back empty subforms (which can have a background)

### LINKED ISSUE #13579: XFA background color not drawn in issue13500
issue13500 (in test corpus)

This regressed when landing #13566. 

image

image

calixteman:
It aims to fix issue #13579.

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -2154,10 +2154,6 @@ class ExclGroup extends XFAObject {
       return HTMLResult.FAILURE;
     }
 
-    if (children.length === 0) {
-      return HTMLResult.EMPTY;
-    }
-
     let marginH = 0;
     let marginV = 0;
     if (this.margin) {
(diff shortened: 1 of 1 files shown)


src/core/xfa/template.js
```diff
@@ -4649,7 +4641,8 @@ class Template extends XFAObject {
       const flush = index => {
         const html = root[$flushHTML]();
         if (html) {
-          hasSomething = true;
+          hasSomething =
+            hasSomething || (html.children && html.children.length !== 0);
```
Snuffleupagus, line 4645: Nit: Here, and similarly below, this can be a bit more compact by instead formatting it as:
```javascript
hasSomething ||= (html.children && html.children.length !== 0);
```

calixteman: I wrote that but the linter failed. And I tried again with parenthesis and it failed again. Do I miss something ?

marco-c (approved):

PR #13627, opened 2021-06-25
XFA - Save filled data in the pdf when downloading the file (Bug 1716288)

calixteman:
- when binding (after parsing) we get a map between some template nodes and some data nodes;
  - so set user data in input handlers in using data node uids in the annotation storage;
  - to save the form, just put the value we have in the storage in the correct data nodes, serialize the xml as a string and then write the string at the end of the pdf using src/core/writer.js;
  - fix few bugs around data bindings:
    - the "Off" issue in Bug 1716288.


brendandahl (approved):
escaping fixes look good thx

PR #13641, opened 2021-06-27
XFA - Implement aspect property on image element

### LINKED ISSUE #13634: Image not scaled proportionality
https://guichet.public.lu/dam-assets/catalogue-formulaires/tva/tva-option-aic-pers-morales/tva-option-aic-pers-morale-DE.pdf

Configuration:
- Web browser and its version: Firefox Nightly 91.0a1 (2021-06-26) (64-bit)
- Operating system and its version: Windows 10.0.19043.1052
- PDF.js version: N/A

1. enableXFA in Firefox Nightly
2. Open form

Header image is scaled properly
image

Image is scaled to fit
image

calixteman:
- it aims to fix issue #13634;
  - move some img-related functions in test/drivers.js in order to have images in xfa reftests.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -66,6 +66,54 @@ function writeSVG(svgElement, ctx, resolve, reject) {
   };
 }
 
+function inlineImages(images) {
+  const imagePromises = [];
+  for (let i = 0, ii = images.length; i < ii; i++) {
+    imagePromises.push(
+      new Promise(function (resolve, reject) {
+        const xhr = new XMLHttpRequest();
+        xhr.responseType = "blob";
+        xhr.onload = function () {
+          const reader = new FileReader();
+          reader.onloadend = function () {
+            resolve(reader.result);
+          };
+          reader.readAsDataURL(xhr.response);
+        };
+        xhr.onerror = function (e) {
+          reject(new Error("Error fetching inline image " + e));
+        };
+        xhr.open("GET", images[i].src);
+        xhr.send();
+      })
+    );
+  }
+  return Promise.all(imagePromises);
+}
+
+async function resolveImages(node, silentErrors = false) {
+  const images = node.getElementsByTagName("img");
+  const data = await inlineImages(images);
+
+  const loadedPromises = [];
+  for (let i = 0, ii = data.length; i < ii; i++) {
+    images[i].src = data[i];
+    loadedPromises.push(
+      new Promise(function (resolveImage, rejectImage) {
+        images[i].onload = resolveImage;
+        images[i].onerror = function (e) {
+          if (silentErrors) {
+            resolveImage();
+          } else {
+            rejectImage(new Error("Error loading image " + e));
+          }
+        };
+      })
+    );
+  }
+  await Promise.all(loadedPromises);
+}
+
 /**
  * @class
  */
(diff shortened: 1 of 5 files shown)


calixteman:
A lot of xfa contains image so a lot of failure in reftests are expected.

/botio test

test/driver.js
```diff
@@ -66,6 +66,51 @@ function writeSVG(svgElement, ctx, resolve, reject) {
...
+      xhr.responseType = "blob";
+      xhr.onload = function () {
+        const reader = new FileReader();
+        reader.onloadend = function () {
+          resolve(reader.result);
+        };
+        reader.readAsDataURL(xhr.response);
+      };
+      xhr.onerror = function (e) {
+        reject(new Error("Error fetching inline annotation image " + e));
```
Snuffleupagus, line 83: Nit: Let's remove the "annotation " part, since the function is now used more generally.

test/driver.js
```diff
@@ -66,6 +66,51 @@ function writeSVG(svgElement, ctx, resolve, reject) {
...
+    imagePromises.push(imagePromise);
+  }
+  return imagePromises;
+}
+
+async function resolveImages(node) {
+  const images = node.getElementsByTagName("img");
+  const imagePromises = inlineImages(images);
+
+  await Promise.all(imagePromises).then(function (data) {
```
marco-c, line 97: Nit: since you are in an async function, you could avoid the `then' and so reduce the indentation level:
```
await Promise.all(imagePromises);
const loadedPromises = [];
...
```

test/driver.js
```diff
@@ -233,25 +254,7 @@ var rasterizeAnnotationLayer = (function rasterizeAnnotationLayerClosure() {
...
-                  images[i].onerror = function (e) {
-                    rejectImage(new Error("Error loading image " + e));
-                  };
-                })
-              );
-            }
-            return loadedPromises;
-          });
-
+          await resolveImages(div);
```
marco-c, line 257: `await resolveImages(div)` is a list of promises, shouldn't you await them via `await Promise.all(await resolveImages(div))`?

marco-c: Or, you could add `await Promise.all(loadedPromises)` at the end of `resolveImages` and keep the code here as it is.

test/driver.js
```diff
@@ -66,6 +66,51 @@ function writeSVG(svgElement, ctx, resolve, reject) {
   };
 }
 
+function inlineImages(images) {
+  const imagePromises = [];
+  for (let i = 0, ii = images.length; i < ii; i++) {
+    const imagePromise = new Promise(function (resolve, reject) {
```
marco-c, line 72: Nit: 
```suggestion
    imagePromises.push(new Promise(function (resolve, reject) {
      ...
    }));
```
so you don't need the intermediate `imagePromise` variable.

test/driver.js
Snuffleupagus: You probably mean `const data = await Promise.all(imagePromises);` here, since otherwise this function would be broken :-)

test/driver.js
Snuffleupagus: The *second* suggestion is definitely the preferred way of doing things here!

---

Given that this patch only moves pre-existing code, it's actually quite interesting that old issue this hasn't caused any bugs (as far as we know).

test/driver.js
marco-c: Yep :)

test/driver.js
calixteman: I locally got some errors because tiff images are not rendered by neither firefox nor chrome which is expected.
So I disabled promise rejection for image failures for the xfa reftests.

marco-c (approved):

PR #13644, opened 2021-06-28
XFA - Support non-embedded fonts without a Widths entry

calixteman:
- some pdf use some fonts which are not embedded or they don't have any width array or don't have any css info (e.g. for standard fonts or Arial).
  - so add widths arrays for Liberation fonts in order to compute the ones for other fonts in using scale factors array.


marco-c (approved):

src/core/liberation_bold_italic_widths.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Widths of glyphes in LiberationSans-BoldItalic.ttf.
+const LiberationBoldItalicWidths = [
```
Snuffleupagus, line 17: Would it not be better to call this `LiberationSansBoldItalicWidths`, and similarly for the other fonts below, to reduce the risk that we'd need to rename everything in the future if e.g. support for *serif* Liberation fonts is added?

Snuffleupagus: Also, given that these arrays aren't *that* large, how about simply placing all of them in just one file instead (e.g. `liberation_sans_widths.js`)?

PR #13648, opened 2021-06-30
XFA - Don't fill when the fill element is not visible (bug 1718735)

calixteman:


marco-c (approved):

calixteman:
The failure is expected.

PR #13649, opened 2021-07-01
Merge the supplemental font data files used with XFA documents

Snuffleupagus:
When XFA support was added, the size of the *built* `pdf.worker.js` file increased quite a bit. Hence I think that it makes sense to, where easily possible, do what we can to (slightly) reduce the size of the PDF.js library.

The supplemental font data files (added for XFA rendering), containing rescale-factors respectively widths, seem like an excellent candidate here since they're not particularly large in either line-count or file sizes.
In this patch these files are instead merged into a *single* file per font, rather than four different ones, and even with these changes the resulting source files don't become all that large.[1]

For e.g. the `gulp mozcentral` build, this reduces the size of the *built* `pdf.worker.js` file by more than `3 kB`. Given the overall simplicity of the patch, that kind of size decrease definitely seem worthwhile to me.

---
[1] Especially when compared to truly large files such as e.g. `glyphlist.js`, `metrics.js`, and `unicode.js`.

PR #13653, opened 2021-07-02
XFA - Improve text layout

calixteman:
- support paragraph margins, line height, letter spacing, ...
  - compute missing dimensions from fields based almost on the dimensions of caption contents.

### DIFF
--- a/src/core/xfa/html_utils.js
+++ b/src/core/xfa/html_utils.js
@@ -14,11 +14,15 @@
  */
 
 import {
+  $content,
   $extra,
   $getParent,
   $getSubformParent,
+  $getTemplateRoot,
+  $globalData,
   $nodeName,
   $pushGlyphs,
+  $text,
   $toStyle,
   XFAObject,
 } from "./xfa_object.js";
@@ -191,8 +195,8 @@ function setMinMaxDimensions(node, style) {
   }
 }
 
-function layoutText(text, xfaFont, fontFinder, width) {
-  const measure = new TextMeasure(xfaFont, fontFinder);
+function layoutText(text, xfaFont, margin, lineHeight, fontFinder, width) {
+  const measure = new TextMeasure(xfaFont, margin, lineHeight, fontFinder);
   if (typeof text === "string") {
     measure.addString(text);
   } else {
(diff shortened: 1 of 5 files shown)


calixteman:
It's expected to have a lot of test failures.
Overall, the rendering looks better (there is one exception in an Italian document where some different texts are not correctly vertically aligned: it's a matter of one pixel and I'll investigate that later).

/botio  test

marco-c (approved):

Snuffleupagus:
This might be a known issue, but looking at e.g. `xfa_bug17176688_3` and comparing with Adobe Reader some of the changes with this patch are improvements and others are regressions.

marco-c:
> This might be a known issue, but looking at e.g. `xfa_bug17176688_3` and comparing with Adobe Reader some of the changes with this patch are improvements and others are regressions.

Most of the changes there looked irrelevant to me, do you have an example of a regression?

calixteman:
I know we've still some issues with line height.

Snuffleupagus:
> Most of the changes there looked irrelevant to me, do you have an example of a regression?

One example: Compare the *bottom* of the first page and the *top* of the second page, in PDF.js and Adobe Reader.

(If necessary I can attach screen-shots later, but based on https://github.com/mozilla/pdf.js/pull/13653#issuecomment-874068427 I guess this is indeed a know issue.)

calixteman:
sadly we had an intermittent failure here...
The message `TEST-UNEXPECTED-FAIL | test failed firefox has not responded in 120s` likely means that firefox crashed, is it correct ?
In comparing logs from the two bots, I've the impress the first test not appearing is `xfa_bug1716980` which could be the one which led to the issue.
Unfortunately I don't manage to reproduce locally.

Snuffleupagus:
> The message `TEST-UNEXPECTED-FAIL | test failed firefox has not responded in 120s` likely means that firefox crashed, is it correct ?

Possibly it *could* also mean that the browser hangs, but I also don't really know :-(

This type of error looks *very* similar, or even identical, to the intermittent ones that made me abandon PR #13649.

> In comparing logs from the two bots, I've the impress the first test not appearing is `xfa_bug1716980` which could be the one which led to the issue.

In the PR mentioned above, initially the failures occurred for another test-case; hence I'm not sure if the problem is with a specific file. However, it seems to me that it's the XFA test-cases which are (intermittently) affected by this. This is pure speculation on my part, but I cannot help wonder if *maybe* there's cases where some of the image resources somehow fail to load (without an error being throw), causing the `resolveImages` helper (in `test/driver.js`) to never resolve!?

> Unfortunately I don't manage to reproduce locally.

Me neither, and I really tried in PR #13649.
I wonder if adding more logging around the `resolveImages`, for the `enableXfa` case, could maybe help?

PR #13654, opened 2021-07-02
XFA - An image can be a stream in the pdf (bug 1718521)

calixteman:
- hrefs can be found in catalog > Names > XFAImages


src/core/catalog.js
```diff
@@ -915,6 +915,22 @@ class Catalog {
...
 
+  get xfaImages() {
+    const obj = this._catDict.get("Names");
+    let xfaImages = null;
+
+    if (obj instanceof Dict && obj.has("XFAImages")) {
+      const nameTree = new NameTree(obj.getRaw("XFAImages"), this.xref);
+      for (const [key, value] of nameTree.getAll()) {
+        if (!xfaImages) {
+          xfaImages = new Dict(null);
```
Snuffleupagus, line 926: Are you absolutely sure that these Streams cannot ever contain indirect objects anywhere (given all the weird things some PDF generators do)?
If not, I'd definitely suggest just using `new Dict(this.xref);` instead to avoid any future bugs (however unlikely).

marco-c (approved):

src/core/catalog.js
```diff
@@ -915,6 +915,22 @@ class Catalog {
...
 
+  get xfaImages() {
+    const obj = this._catDict.get("Names");
+    let xfaImages = null;
+
+    if (obj instanceof Dict && obj.has("XFAImages")) {
+      const nameTree = new NameTree(obj.getRaw("XFAImages"), this.xref);
+      for (const [key, value] of nameTree.getAll()) {
+        if (!xfaImages) {
+          xfaImages = new Dict(this.ref);
```
Snuffleupagus, line 926: Typo: `this.ref` -> `this.xref`

Snuffleupagus:
@calixteman This is causing failures all over the place, looking at the `makeref` logs.
Please back this out for now, and re-land once that's fixed *and* the tests have actually been run (since `makeref` without running tests first is always risky).

calixteman:
> @calixteman This is causing failures all over the place, looking at the `makeref` logs.
> Please back this out for now, and re-land once that's fixed _and_ the tests have actually been run (since `makeref` without running tests first is always risky).

Sorry about that I thought I ran the tests...

PR #13657, opened 2021-07-03
Replace `instanceof Object` with `typeof` checks

Snuffleupagus:
Using `instanceof Object` is generally problematic, since it's not guaranteed to always do the right thing for all Objects.
(I stumbled upon this while working on another patch, when I noticed that the `outlineView` was broken with workers disabled.)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -167,6 +167,10 @@
     "no-new-object": "error",
     "no-restricted-syntax": ["error",
       {
+        "selector": "BinaryExpression[operator='instanceof'][right.name='Object']",
+        "message": "Use `typeof` rather than `instanceof Object`.",
+      },
+      {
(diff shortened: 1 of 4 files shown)


timvandermeij:
I couldn't immediately figure out from the given description in which cases this would fail, but some Googling showed me that this happens for objects without prototypes, which we use. From the console:

```javascript
> const foo = Object.create(null);
undefined
> foo instanceof Object
false
> typeof foo === "object" && foo !== null
true
```

I'm adding this for future reference.

timvandermeij (approved):

timvandermeij:
Thanks!

Snuffleupagus:
> I couldn't immediately figure out from the given description in which cases this would fail,

My apologies; I should have included more details in the commit message!

PR #13658, opened 2021-07-03
Don't attempt to structure clone unsupported types with workers disabled

Snuffleupagus:
Please refer to https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

Based on that information, and manually testing our code, the implementation in `cloneValue` has the following shortcomings:
 - Attempting to clone `function`s is only prevented when they're part of an Object, but is currently allowed when they occur standalone.
 - Cloning of `Symbol`s is currently not prevented, which it should be since the native structured clone algorithm throws.
 - Any disallowed types should be checked first, to reduce the risk of future changes accidentally allowing something that shouldn't be supported.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1805,6 +1805,16 @@ class LoopbackPort {
     function cloneValue(value) {
       // Trying to perform a structured clone close to the spec, including
       // transfers.
+      if (
+        typeof value === "function" ||
+        typeof value === "symbol" ||
+        value instanceof URL
+      ) {
+        throw new Error(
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Given that only the Node.js tests, as run by e.g. GitHub Actions, actually trigger these code-paths running the *full* test-suite seem mostly like a waste of time.

/botio unittest

timvandermeij (approved):

timvandermeij:
Looks good!

PR #13660, opened 2021-07-03
XFA - Disable xfa rendering for XFAF pdfs

calixteman:
- we'll implement XFAF support later.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -86,6 +86,18 @@ class Catalog {
     return shadow(this, "version", version.name);
   }
 
+  /**
+   * @type {boolean} `true` for pure XFA documents,
+   *   `false` for XFA Foreground documents.
+   */
+  get needsRendering() {
+    const needsRendering = this._catDict.get("NeedsRendering");
+    if (!isBool(needsRendering)) {
+      return shadow(this, "needsRendering", false);
+    }
+    return shadow(this, "needsRendering", needsRendering);
+  }
+
   get collection() {
     let collection = null;
     try {
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -849,6 +849,7 @@ class PDFDocument {
   get xfaFactory() {
     if (
       this.pdfManager.enableXfa &&
+      this.catalog.needsRendering &&
       this.formInfo.hasXfa &&
       !this.formInfo.hasAcroForm
     ) {


src/core/catalog.js
```diff
@@ -86,6 +86,14 @@ class Catalog {
     return shadow(this, "version", version.name);
   }
 
+  get needsRendering() {
```
Snuffleupagus, line 89: Can we please just document this getter instead, rather than adding a comment below, since that seems more helpful overall:
```javascript
  /**
   * @type {boolean} `true` for pure XFA documents,
   *   `false` for XFA Foreground documents.
   */
  get needsRendering() {
```

src/core/document.js
```diff
@@ -847,8 +847,12 @@ class PDFDocument {
   }
 
   get xfaFactory() {
+    // NeedsRendering is set to true for pure XFA.
+    // When the pdf uses XFAF then it's false and there's
+    // a XFA entry.
```
Snuffleupagus, line 852: This can be removed, with my suggestion above addressed.

Snuffleupagus (approved):
r=me, with the comments addressed and passing tests; thank you!

Snuffleupagus (changes requested):
With the change in https://github.com/mozilla/pdf.js/pull/13660#event-4975340687, you're no longer testing the code added in this patch as far as I understand!?
In order to address the test failures, would it not be more appropriate to update the test-driver to fix this? I've not tested it, but would adding something like the below *before* https://github.com/mozilla/pdf.js/blob/62808cb3c232ade45cc19495d68a70410daf1682/test/driver.js#L712 work?
```javascript
if (!xfa) {
  annotationLayerCanvas = null;
  return;
}
```

marco-c (dismissed):

brendandahl:
@Snuffleupagus I ran this locally and it looks like we're generating the correct refs for xfaf_issue13647. Is there anything else that needs to be done here or is this good to go?

Snuffleupagus:
> I ran this locally and it looks like we're generating the correct refs for xfaf_issue13647.

As mentioned previously, given the current `test_manifest.json` entry added in this patch we're not actually testing the new code as far as I understand!?

Without a `"enableXfa": true,` line in the manifest, we should be returning already at line https://github.com/mozilla/pdf.js/pull/13660/files#diff-f0385d138acfe65f04aec809ad0522722883e56b6420ccb8f1d2e16b50631c1eR851 as far as I can tell.

brendandahl:
Oh I see. This one is a weird one to test right now since: 1) with xfa disabled it works as expected in the parent version and 2) our test framework doesn't really behave the same way as the viewer. It seems we should really be enabling xfa and making sure the canvas is blank, but that's also problematic. I'd really like to get this into nightly, so I think I'll drop the test and file a follow up issue. Maybe a integration test would be better here.

Snuffleupagus:
> so I think I'll drop the test and file a follow up issue.

That'd be fine by me!
I simply wanted to avoid us adding what's *currently* a useless test-case.

Issue #13633, opened 2021-06-26
Field missing text comb


JDziurlaj:
Attach (recommended) or Link to PDF file here:
https://www.jobcenter-remscheid.de/files/jobcenter-remscheid/download/anzeige-kug101_ba013134.pdf
Configuration:
- Web browser and its version: Firefox Nightly 91.0a1 (2021-06-25) (64-bit)
- Operating system and its version: Windows 10.0.19043.1052
- PDF.js version: N/A

Steps to reproduce the problem:
1. enableXFA in Firefox Nightly
2. Open form

What is the expected behavior? (add screenshot)
Text in field is combed (vertical bars)
![image](https://user-images.githubusercontent.com/7104454/123525595-df558500-d69f-11eb-8f59-e2c70d23ba09.png)
(numbers added for illustration)

What went wrong? (add screenshot)
No comb appears.
![image](https://user-images.githubusercontent.com/7104454/123525586-d5338680-d69f-11eb-93a1-55983f26a666.png)

symtalha14:
https://github.com/mozilla/pdf.js/blob/64f86de5cb70928e748d87dbded99cf0992ba1d7/src/core/document.js#L1266

symtalha14:
https://github.com/mozilla/pdf.js/blob/64f86de5cb70928e748d87dbded99cf0992ba1d7/src/core/document.js#L1266

I think there's some issue with the pageDict object being passed to Page Class in case of xfaFactory. We are sending an empty dict. Due to which:

The parsedAnnotations function ensures annotations but the call to geAnnotations() return an empty array. 
https://github.com/mozilla/pdf.js/blob/64f86de5cb70928e748d87dbded99cf0992ba1d7/src/core/document.js#L496

Here, we are trying to get "annots" and passing the key to the _getInheritableProperty.
https://github.com/mozilla/pdf.js/blob/64f86de5cb70928e748d87dbded99cf0992ba1d7/src/core/document.js#L490

Now the getInheritableProperty function in core/core_utils.js receives an empty dict and sends back an empty array.
https://github.com/mozilla/pdf.js/blob/64f86de5cb70928e748d87dbded99cf0992ba1d7/src/core/core_utils.js#L85

Here if you see the getInheritableProperty simply returns an empty array because the Dict is sent empty.
https://github.com/mozilla/pdf.js/blob/64f86de5cb70928e748d87dbded99cf0992ba1d7/src/core/document.js#L119

[...]

PR #13662, opened 2021-07-03
XFA - Don't use system font when a font is not embeded but there is a substitution

calixteman:
- always use a font coming from pdf.js when there is one: this way we don't use a system font which could looks wrong.


calixteman:
The problem is visible on the screenshot in issue #13633: space between glyphes is a bit strange (see the space after "k" in "bekannt").
I can reproduce locally (and it's worst on my windows ws) and with this patch the pdf in the above issue looks a way better.

Snuffleupagus (approved):
Looks like a good improvement overall, thank you!

PR #13664, opened 2021-07-03
Remove the `version` field from the `package.json` file

Snuffleupagus:
According to https://docs.npmjs.com/cli/v7/configuring-npm/package-json#version, the "version" field is not required[1]:
> If you don't plan to publish your package, the name and version fields are optional.

Hence it shouldn't be necessary to have a "dummy" `version` field in the `package.json` file, and it seems quite unfortunate to have an essentially meaningless entry in that file.[2]
Furthermore, I'd even go as far as suggesting that it's actually doing more harm than good in practice, since it's not uncommon for people to open issues where they simply quote the `package.json`-entry when filling out the ISSUE_TEMPLATE thus causing confusion as to which *exact* version is actually used.

Unless I'm misremembering, I believe that the only reason for adding the `version` field was that is was necessary in order for things to work back when testing was run on Travis.
Now we're using GitHub Actions instead, where things seem to work just find even without a `version` field; hence why I think it makes sense to remove this.

---
[1] Please note that this patch doesn't affect the `pdfjs-dist` package, since the `package.json` file used there is created in `gulpfile.js` during building.

[2] Trying to, automatically, update the `version` field on *every* commit really doesn't seem worth it to me.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7 @@
 {
   "name": "pdf.js",
-  "version": "2.0.0",
-  "lockfileVersion": 1,
   "requires": true,
+  "lockfileVersion": 1,
   "dependencies": {
     "@babel/code-frame": {
       "version": "7.5.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,5 @@
 {
   "name": "pdf.js",
-  "version": "2.0.0",
   "devDependencies": {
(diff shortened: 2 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Seems like a good idea to me!

PR #13666, opened 2021-07-04
Enable the ESLint `operator-assignment` rule

Snuffleupagus:
- Enable the ESLint `operator-assignment` rule

   This patch was generated automatically, using the `gulp lint --fix` command.

   Please find additional details about the ESLint rule at https://eslint.org/docs/rules/operator-assignment

 - Fix the remaining ESLint `operator-assignment` errors

### DIFF
--- a/external/cmapscompress/compress.js
+++ b/external/cmapscompress/compress.js
@@ -204,7 +204,7 @@ function parseCMap(binaryData) {
         bufferSize = 0;
       while (s.length < lengthInChars) {
         while (bufferSize < 4 && stack.length > 0) {
-          buffer = (stack.pop() << bufferSize) | buffer;
+          buffer |= stack.pop() << bufferSize;
           bufferSize += 7;
         }
         s = toHexDigit(buffer & 15) + s;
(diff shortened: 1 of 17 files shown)


external/cmapscompress/compress.js
```diff
@@ -204,6 +204,7 @@ function parseCMap(binaryData) {
         bufferSize = 0;
       while (s.length < lengthInChars) {
         while (bufferSize < 4 && stack.length > 0) {
+          // eslint-disable-next-line operator-assignment
```
timvandermeij, line 207: Why do we need these ignore lines in this commit? The bitwise OR and XOR operations used here should be commutative and therefore the order doesn't matter, so I think we can also use `buffer |= (stack.pop() << bufferSize)` here?

Snuffleupagus: > Why do we need these ignore lines in this commit? The bitwise OR and XOR operations used here should be commutative and therefore the order doesn't matter,

Because I was being dumb and didn't realize that those are in fact commutative :-)

timvandermeij (approved):

timvandermeij:
Thank you!

PR #13671, opened 2021-07-05
XFA - Fix typo which cause test failures

calixteman:


marco-c (approved):

Snuffleupagus (changes requested):
While this typo should obviously be fixed, I do *not* believe that it's in any way related to failures (since `this.ref === undefined`, which the `Dict` will handle).
Opening e.g. `xfa_annual_expense_report.pdf` locally with XFA enabled prints the following, which suggests that the actual error is elsewhere:
```
Warning: XFA - an error occured during layout: TypeError: can't access property "get", this[$globalData].images is undefined
```

*Edit:* The actual bug is at line https://github.com/mozilla/pdf.js/blob/b370d4714f6713007af9f0b0f947a949a947d0c1/src/core/xfa/template.js#L2935, which I suppose should read:
```javascript
 let buffer =
      this[$globalData].images && this[$globalData].images.get(this.href);
```

PR #13673, opened 2021-07-05
XFA - An image can be a stream in the pdf (bug 1718521) 

calixteman:
- hrefs can be found in catalog > Names > XFAImages


t
```diff
@@ -0,0 +1,5707 @@
+[1mdiff --git a/src/core/xfa/template.js b/src/core/xfa/template.js[m
```
marco-c, line 1: I guess this file should not be in the PR ;)

marco-c (approved):

PR #13675, opened 2021-07-05
[test/driver.js] Ensure that Image `src` is set *after* the callbacks in `resolveImages`

Snuffleupagus:
*While I cannot guarantee that this will fix the recent intermittents, this patch really shouldn't hurt.*

By setting the Image `src` first, there's a small possibility that the Image is loaded *before* we've had a change to attach the `onload`/`onerror` callbacks which may cause the Promise to remain in a pending state.
Note that prior to PR #13641 we didn't correctly await all image resources to actually load, which could explain the very recent intermittent test-failures.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -97,7 +97,6 @@ async function resolveImages(node, silentErrors = false) {
 
   const loadedPromises = [];
   for (let i = 0, ii = data.length; i < ii; i++) {
-    images[i].src = data[i];
     loadedPromises.push(
       new Promise(function (resolveImage, rejectImage) {
         images[i].onload = resolveImage;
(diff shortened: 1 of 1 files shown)


calixteman (commented):
I don't know either if it'll fix the issues but it's definitely an interesting find

calixteman (approved):

Snuffleupagus:
Apparently this didn't help :-(
Should we actually bother to land this, or just abandon the patch?

marco-c:
>  Should we actually bother to land this, or just abandon the patch?

I'd say landing it doesn't hurt and might be useful, so... let's land it?

PR #13678, opened 2021-07-05
Log some possible issues when resolving images in tests

calixteman:


Snuffleupagus:
When you actually *want* the tests to fail, they obviously seem to be working just fine :-)

/botio test

Snuffleupagus:
It looks like we don't need this any more, since the intermittent failures stopped :-)

PR #13680, opened 2021-07-06
XFA - Layout correctly a subform with row layout (bug 1718740)

calixteman:
- Fix issues with subformSet elements which are not a real container.


marco-c:
`firefox-xfa_issue13213-page2` and other pages of the same PDF lost the border on the right.

calixteman:
Some lines are pretty small so sometimes depending on the zoom level and rounding errors the lines are drawn or not.
I checked for the one above and it's the case.

marco-c (approved):

PR #13681, opened 2021-07-06
Added page number indicators to the thumbnail views in the sidebar

symtalha14:
I have done a slight change to the user interface. If you see the sidebar which shows the thumbnails, it does not have page number indicators as of now. Only when you hover over a thumbnail, you can see the page number. Sometimes, It can be a little irritating for the user to identify which page he is currently viewing. Many pdf readers like Acrobat etc. also have page numbers at the sidebar. I have added the page indicator to the sidebar thumbnails. Please check it out once.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,22559 @@
 {
   "name": "pdf.js",
   "version": "2.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "2.0.0",
+      "hasInstallScript": true,
+      "license": "Apache-2.0",
+      "devDependencies": {
+        "@babel/core": "^7.14.5",
+        "@babel/plugin-proposal-logical-assignment-operators": "^7.14.5",
+        "@babel/plugin-transform-modules-commonjs": "^7.14.5",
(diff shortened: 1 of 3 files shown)


Snuffleupagus (changes requested):
There's *a lot* of unrelated changes in this patch; looking at the actual changes I'm however not convinced that we want to make this kind of change unfortunately.

Please note that the sidebar follows the initial design specification, from Firefox UI/UX designers, which didn't call for page numbers/labels in the sidebar (and as you mention the page number/label is indeed visible on hover).

Furthermore, covering part of the thumbnails like this seem *generally* undesirable since it's obviously impossible to know in arbitrary documents where on the pages potentially relevant details are placed. (A potentially better option could *maybe* be to follow e.g. Adobe Reader closer, and add page numbers/labels below the thumbnail instead. However doing so would increase the height of the sidebar a fair amount in longer documents thus forcing the user to scroll much more.)
Finally, note that as-is this patch also wouldn't handle documents with page labels correctly.

All-in-all, this unfortunately doesn't seem like a change that we need to make here.

symtalha14:
Hello,
Thanks for the quick response to my pull request. Just to clarify, does
that mean we cannot make any fresh changes to the UI/UX?

Thank you

Issue #13679, opened 2021-07-06
XFA - Cannot open a pdf because of a typo in xfa/config.js


calixteman:
Attach (recommended) or Link to PDF file here:
[Form_MGT-7.pdf](https://github.com/mozilla/pdf.js/files/6768253/Form_MGT-7.pdf)

When we try to open the above pdf, we get in the console:
```
Warning: XFA - an error occured during parsing and binding: TypeError: x.include is not a function
```
which is caused by a typo:
https://github.com/mozilla/pdf.js/blob/b9e84ba70ec5a4a30333b9092f316c9ec7b7f079/src/core/xfa/config.js#L174
 
So fixing the typo allows to open the pdf but there is only 1 page when we have 9 in using masterpdfeditor.

Snuffleupagus:
I'm assuming that the `!!`-part could also be simplified, since the use of AND should already (automatically) be coercing the value to a boolean? Hence `.filter(x => x && x.includes(":"))` ought to suffice as far as I can tell.

marco-c:
@calixteman is this fixed by #13684?

marco-c:
Yes, it's fixed.

PR #13684, opened 2021-07-06
XFA - Default background in rectangle is white

calixteman:
- Fix a typo in order to open the pdf in issue #13679
  - After fixing the fill default color there wer some regressions because of z-index
    and when fixing z-index there were some regressions because of borders
  - So fix the borders rendering.

### DIFF
--- a/src/core/xfa/config.js
+++ b/src/core/xfa/config.js
@@ -171,7 +171,7 @@ class BehaviorOverride extends ContentObject {
       this[$content]
         .trim()
         .split(/\s+/)
-        .filter(x => !!x && x.include(":"))
+        .filter(x => x.includes(":"))
         .map(x => x.split(":", 2))
     );
   }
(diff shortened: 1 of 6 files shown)


calixteman:
It's expected to have a lot of regressions in reftests but overall borders are a way better rendered thanks to this patch.

/botio test

marco-c (approved):

marco-c:
The PDF from https://github.com/mozilla/pdf.js/issues/13634 is better than before (more rectangles are white), but there are still some rectangles that should be white but aren't.

brendandahl:
It looks like  `xfa_issue13213` has some content that is now missing. Is this expected?

marco-c:
> It looks like xfa_issue13213 has some content that is now missing. Is this expected?

Ooh, I had missed that, too many pages!

calixteman:
There is an error on page 4 of `xfa_issue13213` but for the others it is expected: depending of the value of a checkbox theses white rectangle on top of form elements are hidden or not.

brendandahl:
I noticed one more `xfa_bug1718740-page13`. Though the button is in the wrong place to start, so I'm not sure that's a regression.

PR #13687, opened 2021-07-07
XFA - Don't fail xfa loading because of a JS subexpression in SOM expressions

calixteman:
- Fix for one pdf in bug 1717668 (PDFIUM-292-0.pdf).

### DIFF
--- a/src/core/xfa/bind.js
+++ b/src/core/xfa/bind.js
@@ -545,6 +545,12 @@ class Binder {
           // to have something to match with the given expression.
           // See http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=199
           match = createDataNode(this.data, dataNode, ref);
+          if (!match) {
+            // For example if the node contains a .(...) then it isn't
+            // findable.
+            // TODO: remove this when .(...) is implemented.
+            continue;
+          }
           if (this._isConsumeData()) {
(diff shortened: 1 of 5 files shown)


src/core/xfa/bind.js
```diff
@@ -545,6 +545,12 @@ class Binder {
           // to have something to match with the given expression.
           // See http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.364.2157&rep=rep1&type=pdf#page=199
           match = createDataNode(this.data, dataNode, ref);
+          if (!match) {
+            // For example if the node contains a .(...) then it isn't
+            // findable.
+            // TODO: remove this when .(...) is implemented.
```
marco-c, line 551: Can you file an issue to track this TODO (and attach a PDF that uses it)?

calixteman: https://github.com/mozilla/pdf.js/issues/13688

marco-c (approved):

calixteman:
@Snuffleupagus, do you know if there is a command to cancel the last running job ?

Snuffleupagus:
> @Snuffleupagus, do you know if there is a command to cancel the last running job ?

No, not as far as I know unfortunately.

You'd have to directly access the bots and *manually* abort the running processes, so it's probably easier to just let them run to completion to avoid the hassle.

PR #13691, opened 2021-07-07
XFA - Handle correctly nested containers with lr-tb layout (bug 1718670)

calixteman:
- and avoid to push a field with no dimensions when we have some available space in width in a parent.

### DIFF
--- a/src/core/xfa/layout.js
+++ b/src/core/xfa/layout.js
@@ -19,6 +19,7 @@ import {
   $getSubformParent,
   $getTemplateRoot,
   $isSplittable,
+  $isThereMoreWidth,
 } from "./xfa_object.js";
 import { measureToString } from "./html_utils.js";
 
@@ -75,6 +76,7 @@ function flushHTML(node) {
 
   node[$extra].children = [];
   delete node[$extra].line;
+  node[$extra].numberInLine = 0;
 
   return html;
 }
@@ -83,9 +85,9 @@ function addHTML(node, html, bbox) {
   const extra = node[$extra];
   const availableSpace = extra.availableSpace;
 
+  const [x, y, w, h] = bbox;
   switch (node.layout) {
     case "position": {
-      const [x, y, w, h] = bbox;
       extra.width = Math.max(extra.width, x + w);
       extra.height = Math.max(extra.height, y + h);
       extra.children.push(html);
@@ -102,16 +104,17 @@ function addHTML(node, html, bbox) {
           children: [],
         };
         extra.children.push(extra.line);
+        extra.numberInLine = 0;
       }
+
+      extra.numberInLine += 1;
       extra.line.children.push(html);
 
       if (extra.attempt === 0) {
         // Add the element on the line
-        const [, , w, h] = bbox;
         extra.currentWidth += w;
         extra.height = Math.max(extra.height, extra.prevHeight + h);
       } else {
-        const [, , w, h] = bbox;
         extra.currentWidth = w;
         extra.prevHeight = extra.height;
         extra.height += h;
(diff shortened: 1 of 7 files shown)


marco-c:
There is a regression in `firefox-xfa_issue13500-page2`.

calixteman:
I'm sad here because I'd say that this "yellow pages" on 2 lines is not so bad...
The parent container as a layout `lr-tb` and the field "yellow..." has no dimensions. And in this case, we have enough space to put the field at the end of the line.

marco-c:
I'm OK with accepting the regression, @brendandahl wdyt?

calixteman:
I think it's better to fix: we could have strange things because of that.
My idea is when it's possible always "prefer" a non-broken text.

src/core/xfa/template.js
```diff
@@ -1612,8 +1613,18 @@ class Draw extends XFAObject {
...
     const savedH = this.h;
-    const [w, h] = layoutNode(this, availableSpace);
+    const { w, h, isBroken } = layoutNode(this, availableSpace);
     if (w && this.w === "") {
+      // If the parent layout is lr-tb with a w=100 and we already have a child
+      // which takes 90 on the current line.
+      // If we have a text with a length (in px) equal to 100 then it'll be
+      // splitted into almost 10 chunks: so it won't be nice.
+      // So if we've potentially more width to provide in some parent containers
+      // let's do increase it to give a chance to have a better rendering.
```
marco-c, line 1623:
```suggestion
      // let's increase it to give a chance to have a better rendering.
```

src/core/xfa/template.js
```diff
@@ -2232,33 +2266,23 @@ class ExclGroup extends XFAObject {
...
-        if (result.isBreak()) {
-          return result;
-        }
+    const maxRun =
+      this.layout === "lr-tb" || this.layout === "rl-tb"
+        ? MAX_ATTEMPTS_FOR_LRTB_LAYOUT
+        : 1;
+    for (; this[$extra].attempt < maxRun; this[$extra].attempt++) {
+      if (this[$extra].attempt === MAX_ATTEMPTS_FOR_LRTB_LAYOUT - 1) {
+        // Try on the next line so the current one if finished.
```
marco-c, line 2275: ?

src/core/xfa/template.js
```diff
@@ -4600,6 +4660,10 @@ class Subform extends XFAObject {
         ? MAX_ATTEMPTS_FOR_LRTB_LAYOUT
         : 1;
     for (; this[$extra].attempt < maxRun; this[$extra].attempt++) {
+      if (this[$extra].attempt === MAX_ATTEMPTS_FOR_LRTB_LAYOUT - 1) {
+        // Try on the next line so the current one if finished.
```
marco-c, line 4664: ?

marco-c (approved):

src/core/xfa/layout.js
```diff
@@ -325,18 +335,27 @@ function checkDimensions(node, space) {
       // is breakable then we can return true.
       if (node.h !== "" && !node[$isSplittable]()) {
         [, , , h] = getTransformedBBox(node);
-        return Math.round(h - space.height) <= 1;
+        return Math.round(h - space.height) <= ERROR;
       }
       // Else wait and see: this node will be layed out itself
       // in the provided space and maybe a children won't fit.
+
+      if (node.w === "" || Math.round(w - space.width) <= ERROR) {
```
Snuffleupagus, line 343: @calixteman It seems that LGTM complains about this line, since `w === undefined` always holds here; which I'm guessing is probably not what you actually want?

Could we simplify all of the code in the various `case`s by replacing https://github.com/mozilla/pdf.js/blob/c33bf0b5e8dc09ed010d9c4a1dcdaa98975b6378/src/core/xfa/layout.js#L271
with unconditionally doing `const [, y, w, h] = getTransformedBBox(node);` instead?

If, so I can submit a patch to fix that!

PR #13692, opened 2021-07-07
XFA - Correctly bind global data (bug 1718725)

calixteman:


calixteman:
I checked with masterpdfeditor and in clicking on a button I managed to get one of the page here and the string is not rendered too.
So from my pov, the diff is ok.

marco-c (approved):

PR #13693, opened 2021-07-07
XFA - Enable disabled fields (bug 1719464)

calixteman:
- it's a workaround in waiting for JS implementation to let the use fill manually some fields.

### DIFF
--- a/src/display/xfa_layer.js
+++ b/src/display/xfa_layer.js
@@ -161,13 +161,24 @@ class XfaLayer {
       }
     }
 
+    /**
+     * TODO: re-enable that stuff once we've JS implementation.
+     * See https://bugzilla.mozilla.org/show_bug.cgi?id=1719465.
+     *
+     * for (const el of rootDiv.querySelectorAll(
+     * ".xfaDisabled input, .xfaDisabled textarea"
+     * )) {
+     * el.setAttribute("disabled", true);
+     * }
+     * for (const el of rootDiv.querySelectorAll(
+     * ".xfaReadOnly input, .xfaReadOnly textarea"
+     * )) {
+     * el.setAttribute("readOnly", true);
+     * }
+     */
+
     for (const el of rootDiv.querySelectorAll(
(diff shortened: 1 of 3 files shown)


src/display/xfa_layer.js
```diff
@@ -161,13 +161,22 @@ class XfaLayer {
       }
     }
 
+    /*
+       TODO: re-enable that stuff once we've JS implementation.
```
marco-c, line 165: Can you mention https://bugzilla.mozilla.org/show_bug.cgi?id=1719465 in this comment?

src/display/xfa_layer.js
```diff
@@ -161,13 +161,22 @@ class XfaLayer {
...
+       ".xfaDisabled input, .xfaDisabled textarea"
+       )) {
+       el.setAttribute("disabled", true);
+       }
+       for (const el of rootDiv.querySelectorAll(
+       ".xfaReadOnly input, .xfaReadOnly textarea"
+       )) {
+       el.setAttribute("readOnly", true);
+       }
+     */
```
Snuffleupagus, line 176: Nit: Please format block comments as follows
```javascript
/**
 * Comment here...
 */
```

calixteman: Could we add a rule in the linter for that ? (personally I always thought that this kind of comment (the ones you mentioned) were for function documentation (javadoc style)).

marco-c (approved):

src/display/xfa_layer.js
Snuffleupagus: > Could we add a rule in the linter for that ?

I'll try to look into that later, when I've got some spare time.

>  (personally I always thought that this kind of comment (the ones you mentioned) were for function documentation (javadoc style)).

You're correct that we mostly use that kind of comments for JSDocs, and *usually* multi-line comments in the code-base are formatted as:
```javascript
// This is a 
// multi-line
// comment.
```

My point, purely for consistency, was that the *particular* comment formatting used in the earlier version of this patch looked a little "uncommon" for the code-base :-)

PR #13694, opened 2021-07-07
XFA - remove unnecessary check in the `handleBreak` function (PR 13687 follow-up)

Snuffleupagus:
With the changes in PR #13687 we're now checking if `target` is defined *twice* in a row, which shouldn't be necessary :-)

(I noticed this when glancing at the unofficial LGTM results; maybe we should re-evalute the decision to not integrate that into the CI.)


calixteman (approved):
LGTM, thanks.

PR #13696, opened 2021-07-08
[GENERIC viewer] Try to fixup "incomplete" language codes (issue 13689)

### LINKED ISSUE #13689: Localization does not work for sv (Swedish) in Edge (but does in Chrome that reports sv-SE).
_Configuration:_
   - **Edge**: 91.0.864.64 
   - **Chrome**: 91.0.4472.124
- PDF.js version:
   - Latest demo version: https://mozilla.github.io/pdf.js/web/viewer.html

_Steps to reproduce the problem:_
1. Open the demo viewer in Chrome
2. Open the demo viewer in Edge

_What is the expected behavior?_ 

The viewer should be localized, i.e. translated to Swedish.

_What went wrong? (add screenshot)_

The viewer is translated to Swedish in Chrome, but not in Edge.

_Request headers:_
  - Chrome: `Accept-Language: sv-SE,sv;q=0.9,en-US;q=0.8,en;q=0.7`
  - Edge: `Accept-Language: sv,en;q=0.9,en-GB;q=0.8,en-US;q=0.7`

The problem seems to be that the `l10n` folder does not contain an `sv` folder, but only an `sv-SE`.

Snuffleupagus:
While I don't know if it's technically correct to even do this, it could provide a slightly better out-of-the-box behaviour in browsers that specify (from the PDF.js `l10n`-folder perspective) "incomplete" language codes.

Rather than immediately falling back to English, we'll use a white-list to try and re-write a "partial" language code to a (hopefully) suitable one that matches an existing `l10n`-folder. The disadvantage of this solution is that the list needs to be kept *manually* up-to-date with any changes in the `l10n`-folder, however new locales are added infrequently enough that this should be acceptable.

Fixes #13689 (assuming we actually want/care to do so, otherwise we should just WONTFIX the issue).

### DIFF
--- a/web/genericl10n.js
+++ b/web/genericl10n.js
@@ -14,7 +14,7 @@
  */
 
 import "../external/webL10n/l10n.js";
-import { getL10nFallback } from "./l10n_utils.js";
+import { fixupLangCode, getL10nFallback } from "./l10n_utils.js";
 
 const webL10n = document.webL10n;
 
@@ -22,7 +22,7 @@ class GenericL10n {
   constructor(lang) {
     this._lang = lang;
     this._ready = new Promise((resolve, reject) => {
-      webL10n.setLanguage(lang, () => {
+      webL10n.setLanguage(fixupLangCode(lang), () => {
         resolve(webL10n);
       });
     });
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thanks! I'm fine with this small change indeed if it improves the overall out-of-the-box behavior of the viewer.

Issue #11957, opened 2020-06-02
Warning: JPX: Trying to recover from: JPX error: Out of packets


moxun33:
Attach (recommended) or Link to PDF file here:
https://drive.google.com/file/d/1V3UKoqa3-WRiUlC1Mcg-iVQYib8bGACF/view?usp=sharing

Configuration:
- Web browser and its version: chrome 83.0
- Operating system and its version: Windows10 1909
- PDF.js version: latest
- Is a browser extension: no

Steps to reproduce the problem:
1. view pdf on [https://mozilla.github.io/pdf.js/web/viewer.html](https://mozilla.github.io/pdf.js/web/viewer.html)
2. fuzzy render  

What is the expected behavior? (add screenshot)
[![ttX51U.md.png](https://s1.ax1x.com/2020/06/02/ttX51U.md.png)](https://imgchr.com/i/ttX51U)

What went wrong? (add screenshot)
[![ttXIcF.md.png](https://s1.ax1x.com/2020/06/02/ttXIcF.md.png)](https://imgchr.com/i/ttXIcF)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome [extension): [https://mozilla.github.io/pdf.js/web/viewer.html](https://mozilla.github.io/pdf.js/web/viewer.html)

yangwen1020053105:
How to solve this problem,please!

timvandermeij:
This file is broken in Firefox too, with the following console log:

`Warning: JPX: Trying to recover from: JPX error: Out of packets`

Issue #13487, opened 2021-06-04
how to make this pdf file work well


1554312244:
Attach (recommended) or Link to PDF file here:
https://pdf.dfcfw.com/pdf/H2_AN202104301488768931_1.pdf?1619977027000.pdf
Configuration:
- Web browser and its version:Google Chrome  91.0.4472.77
- Operating system and its version:
- PDF.js version: PDF.js: 2.10.22
- Is a browser extension:

Steps to reproduce the problem:
1. open https://mozilla.github.io/pdf.js/web/viewer.html
2. upload the file https://pdf.dfcfw.com/pdf/H2_AN202104301488768931_1.pdf?1619977027000.pdf
3. the page 8 looks not well .

What is the expected behavior? (add screenshot)

What went wrong? (add screenshot)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

1554312244:
![image](https://user-images.githubusercontent.com/12227998/120758759-cd6f3f00-c544-11eb-8077-0569e4895c99.png)

Snuffleupagus:
> PDF.js version:latest

You should always provide an actual version *number*, since "latest" is ambiguous and thus of no use.

1554312244:
> > PDF.js version:latest
> 
> You should always provide an actual version _number_, since "latest" is ambiguous and thus of no use.

the version is '2.10.22'

Snuffleupagus:
Duplicate of #11957

linfangrong:
@1554312244

Issue #12579, opened 2020-11-05
firefox open this pdf file has wrong with picture in the file


wowplayer01:
Attach (recommended) or Link to PDF file here:
[（3）2020年中心组学习记录情况资料(1).pdf](https://github.com/mozilla/pdf.js/files/5492304/3.2020.1.pdf)

Configuration:
- Web browser and its version:  firefox  82.0.2
- Operating system and its version: windows 10
- PDF.js version: 2.0
- Is a browser extension: 

Steps to reproduce the problem:
1. 
2.

What is the expected behavior? (add screenshot)
![Dingtalk_20201105142744](https://user-images.githubusercontent.com/42967421/98205437-222ce600-1f73-11eb-8aa0-dabe8d7aa4af.jpg)

![Dingtalk_20201105141020](https://user-images.githubusercontent.com/42967421/98204294-b9dd0500-1f70-11eb-93ee-1f25e6201918.jpg)

What went wrong? (add screenshot)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

SamyCookie:
seems to be a duplicate of #11957

PR #13628, opened 2021-06-25
Check that TrueType (3, 0) cmap tables, for symbolic fonts, are sorted correctly (issue 13626)

Snuffleupagus:
- According to a comment in `readCmapTable`, we're assuming that the cmap tables (when more than one exist) are sorted in ascending order. If that's not the case, keep checking the following cmap tables in order to fix the referenced issue.

   Fixes #13626

 - Suppress OTS warnings about the caretOffset in the hhea-table

    - https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6hhea.html
    - https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6head.html

PR #13698, opened 2021-07-08
Fix JPX tag tree decode (issue 11957)

linfangrong:
fix jpx tag tree decode
error message like:
`Warning: JPX: Trying to recover from: JPX error: Out of packets`

the relevant section of the JPEG 2000 specification:
[J2K_Guide.pdf](https://gwg.nga.mil/ntb/baseline/docs/bwcj2k/J2K_Guide_WD1.pdf)  Page 79  (B.10.8)

Since code-block 0,0 is included in this layer for the first time, we need to code the number of zero bit-planes that were skipped.  For this, we use the zero bit-planes tag tree, which is a separate tag tree from the inclusion information tag tree, with its own state.  In this tag tree, we have not coded any partial information yet, so we start at the root and code all nodes on the branch to the leaf for code-block 0,0.  The root q0(0,0) is 3, but its initial state was 0, so we need to increment by 3; i.e. we code three 0s followed by a 1 to lock in the value.  The intermediate node q1(0,0) is also 3, so there is no increment to its parent’s value, and we code the 1 to lock in its value.  The leaf node q2(0,0) is also 3, so again we simply code a 1.  Thus, the code word is 000111.

Fixes 
https://github.com/mozilla/pdf.js/issues/11957
https://github.com/mozilla/pdf.js/issues/12579
https://github.com/mozilla/pdf.js/issues/13487

### DIFF
--- a/src/core/jpx.js
+++ b/src/core/jpx.js
@@ -1203,6 +1203,11 @@ function parseTilePackets(context, data, offset, dataLength) {
           zeroBitPlanesTree = new TagTree(width, height);
           precinct.inclusionTree = inclusionTree;
           precinct.zeroBitPlanesTree = zeroBitPlanesTree;
+          for (let l = 0; l < layerNumber; l++) {
+            if (readBits(1) !== 0) {
+              throw new JpxError("Invalid tag tree");
+            }
+          }
         }
 
         if (inclusionTree.reset(codeblockColumn, codeblockRow, layerNumber)) {
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -3987,6 +3987,15 @@
        "lastPage": 1,
        "type": "eq"
     },
+    {  "id": "issue12579",
+       "file": "pdfs/issue12579.pdf",
+       "md5": "a56f6c7b6ebe0008bd6215b9a4dc3cca",
+       "link": true,
+       "firstPage": 3,
+       "lastPage": 3,
+       "rounds": 1,
+       "type": "eq"
+    },
     {  "id": "issue11330",
        "file": "pdfs/issue11330.pdf",
        "md5": "03a8a53d4b0dc825e08554f5c0178308",
--- a/test/pdfs/issue12579.pdf.link
+++ b/test/pdfs/issue12579.pdf.link
@@ -1,0 +1,1 @@
+https://github.com/mozilla/pdf.js/files/5492304/3.2020.1.pdf


src/core/jpx.js
```diff
@@ -192,7 +192,6 @@ class JpxImage {
                 XRsiz: data[j + 1],
                 YRsiz: data[j + 2],
               };
-              j += 3;
```
Snuffleupagus, line 195: Sorry, but this is definitely *not* correct since it'll incorrectly use the same data for each component; please revert this.

src/core/jpx.js
```diff
@@ -1203,6 +1202,11 @@ function parseTilePackets(context, data, offset, dataLength) {
           zeroBitPlanesTree = new TagTree(width, height);
           precinct.inclusionTree = inclusionTree;
           precinct.zeroBitPlanesTree = zeroBitPlanesTree;
+          for (let l = 0; l < layerNumber; l++) {
+            if (readBits(1) !== 0) {
+              throw new Error('Invalid tag tree');
```
Snuffleupagus, line 1207: Please use `JpxError`, rather than a plain `Error`.

Snuffleupagus (changes requested):
First of all, please link to the relevant section of the JPEG 2000 specification such that it's possible to understand/review the patch.

Secondly, please add a *linked* test-case to the patch; https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing may be generally useful.

Finally, please provide a more descriptive *commit message* actually outlining the changes in some detail (e.g. with a reference to the specification).

---

When updating the patch, please keep https://github.com/mozilla/pdf.js/wiki/Squashing-Commits in mind.

linfangrong:
92fc1584-a0c6-4628-9b60-f26489dfa229

src/core/jpx.js
linfangrong: yeah，I see

linfangrong (commented):
https://gwg.nga.mil/ntb/baseline/docs/bwcj2k/J2K_Guide_WD1.pdf

Page 79    B.10.8

test/pdfs/.gitignore
```diff
@@ -392,6 +392,7 @@
 !issue6298.pdf
 !issue6889.pdf
 !issue11473.pdf
+!issue13487.pdf
```
Snuffleupagus, line 395: As mentioned previously, please use a *linked* test-case instead since for copyright reasons we cannot include arbitrary PDF files in the repository; please look at e.g. https://github.com/mozilla/pdf.js/pull/13628/files for an example of how that is done.

Furthermore, please pick a *smaller* PDF document as the test-case (there's a bunch of duplicate issues listed in #11957), since we really don't want to use a 10 MB file unless no other option is available.

test/test_manifest.json
```diff
@@ -5226,6 +5226,12 @@
        "type": "eq",
        "annotations": true
     },
+    {  "id": "issue13487",
+       "file": "pdfs/issue13487.pdf",
+       "md5": "80044ae1ce3fb8c748d6a72f868d5de1",
+       "rounds": 1,
```
Snuffleupagus, line 5232: When you've addressed the previous comment, please ensure that you're only testing *one* page rather than the entire document; make use of the `firstPage`/`lastPage` properties as necessary (see examples of that in the `test_manifest.json` file).

Snuffleupagus (changes requested):
Again, please remember to update the *commit message* with at least the same information that you've now included in https://github.com/mozilla/pdf.js/pull/13698#issue-685998397.

test/pdfs/.gitignore
linfangrong: OK

Snuffleupagus (changes requested):
There's no longer *any* test-case included here, did you forget to add the new linked test-case (and the associated `test_manifest.json` changes)?

Also, again, please write a proper commit message!
I'm also finding it slightly difficult to interpret the specification in relation to the patch, so it would *definitely* be a big help and speed things up considerably if you could (in some detail) actually explain your changes a little bit more.

test/pdfs/.gitignore
```diff
@@ -257,6 +257,7 @@
 !glyph_accent.pdf
 !personwithdog.pdf
 !find_all.pdf
+!issue13487.pdf
```
Snuffleupagus, line 260: This is still the 10 MB file, please don't use that one since it's way too large!

test/test_manifest.json
```diff
@@ -1507,6 +1507,15 @@
        "rounds": 1,
        "type": "load"
     },
+    {  "id": "issue13487",
+       "file": "pdfs/issue13487.pdf",
+       "md5": "80044ae1ce3fb8c748d6a72f868d5de1",
+       "link": true,
+       "firstPage": 7,
+       "lastPage": 8,
```
Snuffleupagus, line 1515: The firstPage/lastPage should be the *same* number to limit the test to only one page.

Snuffleupagus (changes requested):
In order to save both of our time, please address *all* of the previously mentioned issues *before* asking for a new review.

test/pdfs/.gitignore
Snuffleupagus: Also, with a *linked* test-case there's no need for editing this file.

Snuffleupagus (changes requested):
I'm having some trouble "connecting the dots" (so to speak) between the specification and the PDF.js implementation here, so in order for me to be able to review this (as I've asked for twice before):
Please write a couple of sentences explaining how to interpret this change, for example *quoting specific* parts of the specification text and explain how they relate to the particular PDF.js code being modified here.

Since you must have read the specification in some detail here, simply explaining your train of thought would help immensely!

---

Finally, please add that information to the the *commit message* of the patch (since it currently contains nothing but "[FIX] jpx distortion" which isn't really helpful when using e.g. the Git command line).

Snuffleupagus:
@brendandahl Would you be able to take over the review here, since I unfortunately (even after some trying) don't really understand the JPEG 2000 image format well enough to properly attempt to review this?

linfangrong:
For Squashing Commits

brendandahl:
/botio-linux test 
We're having some intermittent failures (probably not your patch).

brendandahl:
@linfangrong Thanks for the patch! If you're interested, it would be great to have your help on some more [jpeg2000 bugs](https://github.com/mozilla/pdf.js/issues?q=is%3Aopen+is%3Aissue+label%3A4-image-jpx).

PR #13705, opened 2021-07-09
XFA - Fix text positions (bug 1718741)

calixteman:
- font line height is taken into account by acrobat when it isn't with masterpdfeditor: I extracted a font from a pdf, modified some ascent/descent properties thanks to ttx and the reinjected the font in the pdf: only Acrobat is taken it into account. So in this patch, line heights for some substituted fonts are added.
  - it seems that Acrobat is using a line height of 1.2 when the line height in the font is not enough (it's the only way I found to fix correctly bug 1718741).
   - don't use flex in wrapper container (which was causing an horizontal overflow in the above bug).
   - consequently, the above fixes introduced a lot of small regressions, so in order to see real improvements on reftests, I fixed the regressions in this patch:
     - replace margin by padding in some case where padding is a part of a container dimensions;
     - remove some flex display: some containers are wrongly sized when rendered;
     - set letter-spacing to 0.01px: it helps to be sure that text is not broken because of not enough width in Firefox.

### DIFF
--- a/src/core/calibri_factors.js
+++ b/src/core/calibri_factors.js
@@ -81,6 +81,7 @@ const CalibriBoldFactors = [
   0.95794, 0.95794, 0.82616, 0.86513, 0.85162, 0.85162, 0.85162, 0.85162,
   0.91133, 0.85162, 0.79492, 0.79492, 0.79492, 0.79492, 0.91133, 0.79109,
 ];
+const CalibriBoldLineHeight = 1.2207;
 
 // Factors to rescale LiberationSans-BoldItalic.ttf to have the same
 // metrics as calibriz.ttf.
@@ -152,6 +153,7 @@ const CalibriBoldItalicFactors = [
   0.84548, 0.84548, 0.91133, 0.84548, 0.79492, 0.79492, 0.79492, 0.79492,
   0.91133, 0.74081,
 ];
+const CalibriBoldItalicLineHeight = 1.2207;
 
 // Factors to rescale LiberationSans-Italic.ttf to have the same
 // metrics as calibrii.ttf.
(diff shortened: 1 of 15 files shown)


calixteman:
Almost all the reftests are broken: I checked locally all of them and overall it's an improvement, I compared a lot of output with Acrobat and for most of them we've very close rendering.

/botio test

marco-c:
A few observations:
- In `firefox-xfa_bug1718670_1-page1`, towards the end of the page there is some text moving a bit behind a field.
- In `firefox-hsbc_closure-page1`, towards the bottom of the page, the "IBAN" text disappeared.
- In "firefox-xfa_bug1716816-page1", the bottom of the red rectangle is now behind "Numero de Referencia".
- In "firefox-xfa_bug1717668_4-page1" the header of the table at the bottom is repeated on page 2 too.

calixteman:
about:
  - `firefox-xfa_bug1718670_1-page1`, it's firefox-linux specific
  - `firefox-hsbc_closure-page1`: I think you made a confusion: "IBAN" disappeared in the previous version
  - `firefox-xfa_bug1717668_4-page1`: it's by design: when there is an overflow there is a leader. I open this doc in acrobat and the delete is on the first page because there is no delete in table I.B. (thanks to JS).
  - `firefox-xfa_bug1716816-page1`: I noticed that too.

Since we've an overall improvement, I'd prefer fixing the 2 small regressions in a follow-up: it'd make my life easier.

marco-c:
> Since we've an overall improvement, I'd prefer fixing the 2 small regressions in a follow-up: it'd make my life easier.

WFM, can you file bugs for the regressions?

marco-c (approved):

PR #13706, opened 2021-07-09
XFA - Keep xfa layer on top of the others (bug 1719629)

calixteman:


marco-c (approved):

web/pdf_page_view.js
```diff
@@ -496,6 +496,11 @@ class PDFPageView {
         div.appendChild(textLayerDiv);
       }
 
+      if (this.xfaLayer?.div) {
+        // The xfa layer needs to stay on top.
+        div.appendChild(this.xfaLayer.div);
+      }
```
Snuffleupagus, line 502: If I'd actually been given a chance to review this *before* it landed, I'd have suggested to move this block and place it around line 513 instead rather than here since that'd seem generally "safer"!

With this patch, you're not actually guaranteed a correct placement of the XFALayer in the DOM in the event that the textLayer is disabled. This probably won't matter in practice, since I'm hoping that XFA documents don't contain a separate AnnotationLayer (although how knows with XFA Foreground documents).

(Since as was asked for review this, and the patch was opened a mere two hours ago, I guess that I don't really understand the big rush here!?)

marco-c: > (Since as was asked for review this, and the patch was opened a mere two hours ago, I guess that I don't really understand the big rush here!?)

We're trying to land as many XFA fixes as possible as we hope to enable XFA in Firefox 91.

PR #13712, opened 2021-07-10
XFA - Always compute the transformed BBox values in `checkDimensions` (PR 13691 follow-up)

Snuffleupagus:
This way we ensure that these BBox values are *always* defined as expected for every `case`-block, and we also don't need to duplicate the lookup in multiple places. (Also, the patch removes a couple of unnecessary line-breaks in existing comments.)

Fixes https://github.com/mozilla/pdf.js/pull/13691#pullrequestreview-702356627, which was flagged by LGTM.

### DIFF
--- a/src/core/xfa/layout.js
+++ b/src/core/xfa/layout.js
@@ -191,8 +191,7 @@ function getAvailableSpace(node) {
 }
 
 function getTransformedBBox(node) {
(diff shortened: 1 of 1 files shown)


calixteman (approved):
LGTM. Thanks.

PR #13714, opened 2021-07-10
XFA - Use fake MyriadPro as a fallback for missing fonts

### LINKED ISSUE #13597: XFA Only some pages render
http://help.adobe.com/en_US/livecycle/11.0/SampleForms/941/Form.pdf

- Web browser and its version: Firefox 89.0.1
- Operating system and its version: Windows 10 10.0.19043.1052
- PDF.js version: commit bdfcdd6f8326f22c135edfa29a341ea4c558f3d1
- Is a browser extension: No

1. Enable XFA support
2. Open form

Page 3 (of 4) renders
image

Only 3 pages total render in pdf.js

calixteman:
- aims to fix #13597.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -993,7 +993,22 @@ class PDFDocument {
     promises.length = 0;
     pdfFonts.length = 0;
 
+    const reallyMissingFonts = new Set();
     for (const missing of missingFonts) {
+      if (!getXfaFontWidths(`${missing}-Regular`)) {
+        // No substitution available: we'll fallback on Myriad.
+        reallyMissingFonts.add(missing);
+      }
+    }
+
+    if (reallyMissingFonts.size) {
+      missingFonts.push("PdfJS-Fallback");
+    }
+
+    for (const missing of missingFonts) {
+      if (reallyMissingFonts.has(missing)) {
+        continue;
+      }
       for (const fontInfo of [
         { name: "Regular", fontWeight: 400, italicAngle: 0 },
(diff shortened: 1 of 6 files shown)


src/core/document.js
```diff
@@ -993,7 +993,23 @@ class PDFDocument {
...
+
     for (const missing of missingFonts) {
+      if (
+        missing === "MyriadPro" &&
+        !hasMyriad &&
+        reallyMissingFonts.size === 0
+      ) {
+        break;
+      }
+
```
Snuffleupagus, line 1012: I'm finding this, and the new code above, slightly confusing to reason about; any chance that it could be simplified (or re-factored) somewhat?

calixteman: The idea was to load Myriad only if Myriad is present in the list of the missing fonts (=> hasMyriad) or if a font has no substitution (.size === 0) and for the latter we need to review all the missing fonts so it's why I put the MyriadPro at the of the list.
Anyway you're right: I'm pretty sure that I'll find myself pretty painful to read this code in 2 weeks...

Snuffleupagus: Thank you; the updated patch is *a lot* clearer and easier to follow!

Snuffleupagus:
It seems that these changes caused a couple of fonts to become *italic*, which I'm not sure is correct when comparing against Adobe Reader; for example `xfa_bug1717668_1` and `xfa_issue13679`.

Snuffleupagus (approved):

PR #13715, opened 2021-07-11

Snuffleupagus:


timvandermeij:
I read about a possible regression in Puppeteer 10.1.0, see https://github.com/puppeteer/puppeteer/pull/7414, so we'll have to see if it affects us as well.

/botio test

timvandermeij:
While it does look like that PR is for functionality in Puppeteer that we don't use, namely connecting to a _remote_ browser instance, it can't hurt to make sure :-)

Snuffleupagus:
> I read about a possible regression in Puppeteer 10.1.0, see [puppeteer/puppeteer#7414](https://github.com/puppeteer/puppeteer/pull/7414), so we'll have to see if it affects us as well.
> 
> While it does look like that PR is for functionality in Puppeteer that we don't use, namely connecting to a _remote_ browser instance, it can't hurt to make sure :-)

All tests pass locally with this patch, so I believe that we should be good here. (I was just waiting on tests to finish locally, before triggering them here :-)

timvandermeij (approved):

timvandermeij:
Thanks!

Issue #13668, opened 2021-07-04
Large amount of text missing


JDziurlaj:
Attach (recommended) or Link to PDF file here:
https://www.placementsmondiauxsunlife.com/content/dam/sunlife/regional/canada/documents/slgi/4839-I-F.pdf

Configuration:

- Web browser and its version: 91.0a1 (2021-07-03) (64-bit)
- Operating system and its version: Windows 10

Steps to reproduce the problem:
1. enableXfa = true
2. Open form

What is the expected behavior? (add screenshot)

![image](https://user-images.githubusercontent.com/7104454/124389449-1ede2b80-dcb5-11eb-98ba-01e1f3d6c6f5.png)

What went wrong? (add screenshot)

![image](https://user-images.githubusercontent.com/7104454/124389460-2c93b100-dcb5-11eb-9768-98ddd2c8f1e7.png)

PR #13716, opened 2021-07-11
XFA - Avoid to have containers not pushed in the html

calixteman:
- it aims to fix issue #13668.

### DIFF
--- a/src/core/xfa/layout.js
+++ b/src/core/xfa/layout.js
@@ -52,10 +52,21 @@ import { measureToString } from "./html_utils.js";
  * returning.
  */
 
+function createLine(node, children) {
+  return {
+    name: "div",
+    attributes: {
+      class: [node.layout === "lr-tb" ? "xfaLr" : "xfaRl"],
+    },
+    children,
+  };
+}
+
 function flushHTML(node) {
   if (!node[$extra]) {
     return null;
   }
+
   const attributes = node[$extra].attributes;
   const html = {
     name: "div",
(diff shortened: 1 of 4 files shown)


marco-c (approved):

Snuffleupagus:
Hmm; given that *both* browsers timed-out on *both* of the bots during the tests above, see https://github.com/mozilla/pdf.js/pull/13716#issuecomment-877836304 and https://github.com/mozilla/pdf.js/pull/13716#issuecomment-877836572, how do we know that this doesn't cause a bunch of breakage/regressions!?

calixteman:
My bad, I missed that... I checked locally that everything was good and it was.
Anyway, do you prefer I revert and then push again, test, ... ?

marco-c:
> Hmm; given that both browsers timed-out on both of the bots during the tests above, see #13716 (comment) and #13716 (comment), how do we know that this doesn't cause a bunch of breakage/regressions!?

Where can I see an indication that they timed out?

Snuffleupagus:
> I checked locally that everything was good and it was.

Strange, for me this patch actually times-out locally as well.

> Where can I see an indication that they timed out?

There's a `errors: 1238` respectively `errors: 1226` printed in https://github.com/mozilla/pdf.js/pull/13716#issuecomment-877836304 respectively https://github.com/mozilla/pdf.js/pull/13716#issuecomment-877836572, and those kind of (general) errors should never appear when testing a working patch. After that, following the "Full output at..." links, the reference tests end with:
```
TEST-UNEXPECTED-FAIL | test failed firefox has not responded in 120s
TEST-UNEXPECTED-FAIL | test failed chrome has not responded in 120s
```

marco-c:
> There's errors: 1238 respectively errors: 1226 printed in #13716 (comment) respectively #13716 (comment), and those kind of (general) errors should never appear when testing a working patch. After that, following the "Full output at..." links, the reference tests end with:

Could the bot surface the "has not responded" errors in the comment or the HTML output?

Snuffleupagus:
@calixteman I've tested this patch locally, in the viewer (with `enableXfa = true` set), and `xfa_bug1717668_2` causes the browser (or tab rather) to hang (with full CPU utilization of one core) and the document doesn't render.

---

> Could the bot surface the "has not responded" errors in the comment or the HTML output?

But as mentioned this is already indirectly visible, via the `errors: {number here}` lines, since those should *never* appear and that's thus a clear indication to look at the full logs.

calixteman:
The same for me locally... I was sure I tested (and I'm still sure) but obviously I didn't.

PR #13721, opened 2021-07-12
XFA - Avoid to have containers not pushed in the html

### LINKED ISSUE #13668: Large amount of text missing

calixteman:
- it aims to fix issue #13668.

### DIFF
--- a/src/core/xfa/layout.js
+++ b/src/core/xfa/layout.js
@@ -52,10 +52,21 @@ import { measureToString } from "./html_utils.js";
  * returning.
  */
 
+function createLine(node, children) {
+  return {
+    name: "div",
+    attributes: {
+      class: [node.layout === "lr-tb" ? "xfaLr" : "xfaRl"],
+    },
+    children,
+  };
+}
+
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
Compared with Adobe Reader, the `issue13611` test-case seems to regress a little bit.

calixteman:
The issue fixed by this patch is a way worse than the regression.
So if you're ok, we can merge this PR and then we could update the m-c version (once https://bugzilla.mozilla.org/show_bug.cgi?id=1720159 is fixed too) and I'll fix the regression in a follow-up.
@marco-c, are you ok ?

calixteman:
Finally the fix was pretty easy.

/botio test

marco-c (approved):

PR #13723, opened 2021-07-12
XFA - Only the first page was printed (bug 1720159)

calixteman:


web/viewer.css
```diff
@@ -1844,6 +1844,10 @@ html[dir="rtl"] #documentPropertiesOverlay .row > * {
     align-items: center;
   }
 
+  #printContainer .xfaPage {
+    position: initial;
+  }
+
```
Snuffleupagus, line 1850: Looking at the canvas-related printing rules further down, do you perhaps want to do something like `#printContainer > .xfaPrintedPage .xfaPage` instead here (e.g. for consistency)?

web/viewer.css
```diff
@@ -1844,6 +1844,10 @@ html[dir="rtl"] #documentPropertiesOverlay .row > * {
     align-items: center;
   }
 
+  #printContainer .xfaPage {
```
Snuffleupagus, line 1847: Looking at the canvas-related printing rules further down, do you perhaps want to do something like `#printContainer > .xfaPrintedPage .xfaPage` instead here (e.g. for consistency)?

marco-c (approved):

PR #13728, opened 2021-07-14
Replace recursive call to ensure by a loop

calixteman:
- it could potentially avoid to have some recursion errors and so avoid some failures in reftests.

### DIFF
--- a/src/core/pdf_manager.js
+++ b/src/core/pdf_manager.js
@@ -178,18 +178,19 @@ class NetworkPdfManager extends BasePdfManager {
   }
 
   async ensure(obj, prop, args) {
-    try {
-      const value = obj[prop];
-      if (typeof value === "function") {
-        return value.apply(obj, args);
+    while (true) {
+      try {
+        const value = obj[prop];
+        if (typeof value === "function") {
+          return value.apply(obj, args);
+        }
+        return value;
+      } catch (ex) {
+        if (!(ex instanceof MissingDataException)) {
+          throw ex;
+        }
+        await this.requestRange(ex.begin, ex.end);
       }
-      return value;
-    } catch (ex) {
-      if (!(ex instanceof MissingDataException)) {
-        throw ex;
-      }
-      await this.requestRange(ex.begin, ex.end);
-      return this.ensure(obj, prop, args);
     }
   }
 


calixteman:
I've the impress that these last times, I've a little more recursion errors locally and I was wondering if it could be the root cause of intermittent failures in reftests.

Snuffleupagus:
My general question, given how basically *everything* goes throw that method, is what (if any) performance impact these sort of changes will have in general?
Given that the recent failures all seem connected to the XFA test-cases, I do wonder what's "special" about the XFA documents (compared to regular PDF documents) that cause them to trigger these problems?

Anyway, let's start with running the test-suite a few times to see if this patch actually helps.

/botio test

calixteman:
I'd be surprised if there is any optimization for a tail call which in the catch handler, so I'd say that we've normal calls and stuff is put on the stack at each call. So it should be slightly faster to run that stuff in a loop.
My concern would be: thanks to the stack size itself we've a limit so it isn't possible to loop for ever, here in this patch we've a `while (true)`, so is it possible to have a call which throw a `MissingDataException` for ever ? if not then I think it cannot hurt to in general to have this.
Anyway as you said I don't see either where is the connection with xfa...

calixteman:
Anyway from time to time I see locally errors like
```
Error: http://localhost:8888/src/core/calibri_factors.js: too much recursion
  Instantiating http://localhost:8888/src/core/calibri_factors.js
  Loading http://localhost:8888/src/core/xfa_fonts.js
  Loading pdfjs/core/worker.js
```
in `system.js: 4`.
Would it be possible to have similar errors in reftests ?

Snuffleupagus:
> I'd be surprised if there is any optimization for a tail call which in the catch handler, so I'd say that we've normal calls and stuff is put on the stack at each call. So it should be slightly faster to run that stuff in a loop.

Good to know, thank you!

> so is it possible to have a call which throw a `MissingDataException` for ever ?

That should *not* be possible (famous last words), since the `requestRange`-call will request the specific byte-range from the server and this code should simply wait until the data (eventually) becomes available.
In the event that the server never responds the Promise returned by the `ensure` method will simply never settle, which is already the case in the existing code.

Also, for consistency it'd be nice to update this method in the same way:
https://github.com/mozilla/pdf.js/blob/a17bd13023b476db874e8ebe240e4d8969cb793e/src/core/xref.js#L858-L868

---

> Anyway from time to time I see locally errors like
> 
> ```
> Error: http://localhost:8888/src/core/calibri_factors.js: too much recursion
>   Instantiating http://localhost:8888/src/core/calibri_factors.js
>   Loading http://localhost:8888/src/core/xfa_fonts.js
>   Loading pdfjs/core/worker.js
> ```
> 
> in `system.js: 4`.
> Would it be possible to have similar errors in reftests ?

That *should* be unrelated here, since nowadays SystemJS is only being used in the development viewer[1] and not e.g. in the test-suite.

---
[1] Note that I want to remove SystemJS completely, see PR #12563, however it's still blocked on [bug 1247687](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687) being fixed.

Snuffleupagus:
@calixteman Do we still want to make these change, since the PR unfortunately doesn't seem to improve the reference tests?
If yes, please also update the `XRef.fetchAsync`-method as mentioned in https://github.com/mozilla/pdf.js/pull/13728#issuecomment-879798554.

Snuffleupagus:
Closing this PR for now; please re-open if this change is still deemed necessary (as mentioned previously `XRef.fetchAsync` should also be updated in that case).

PR #13732, opened 2021-07-14
XFA - A rectangle must have the width of its parent but without inner margins

### LINKED ISSUE #13584: XFA Geometry exceeds containers
http://education.ohio.gov/getattachment/Topics/Special-Education/Preschool-Special-Education/ETR-Preschool-Planning-Form.pdf.aspx

- Web browser and its version: Firefox 89.0.1
- Operating system and its version: Windows 10 10.0.19043.1052
- PDF.js version: commit 5d251a3a3e7bfaa8fbfa07f0b7f9901eb3f999d9
- Is a browser extension: No

1. Enable XFA support
2. Open form

Shapes stay inside table cells
image

Shapes fall outside cells (among other issues)
image

calixteman:
- it aims to fix #13584;
  - to avoid bad rendering because of clipping just set overflow to visible on SVG element.

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -115,6 +115,16 @@ const MAX_ATTEMPTS_FOR_LRTB_LAYOUT = 2;
 // the loop after having MAX_EMPTY_PAGES empty pages.
 const MAX_EMPTY_PAGES = 3;
 
+function hasMargin(node) {
+  return (
+    node.margin &&
+    (node.margin.topInset ||
+      node.margin.rightInset ||
+      node.margin.bottomInset ||
+      node.margin.leftInset)
+  );
+}
+
 function _setValue(templateNode, value) {
   if (!templateNode.value) {
     const nodeValue = new Value({});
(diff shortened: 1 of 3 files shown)


calixteman:
Something is wrong with dhl test.

marco-c:
firefox-xfa_bug1716380-page2 and firefox-xfa_bug1716809-page1 look a bit worse

calixteman:
For example, in `xfa_bug1716809.pdf` the rectangle around "Interesado" is drawn thanks to a border when the content just after are inside a svg rect and it seems that they rendered differently at least for non-integer stroke widths (0.5px here).
I tested locally in the viewer and I'd say that the diff is acceptable: before the patch some lines disappeared at some zoom levels (the parent div was clipping the svg rect which has non-integer stroke width).
Maybe we should either use a border to render a rect or use a rect to render a border to have at least something consistent when drawing some lines.

marco-c:
The regressions are minor, so we will accept them.

marco-c (approved):

marco-c:
@calixteman can you file an issue to track the (minor) regression?

PR #13733, opened 2021-07-14
XFA - Textfield must a white background by default

calixteman:


src/display/xfa_layer.js
```diff
@@ -101,6 +102,7 @@ class XfaLayer {
 
     // Set the value after the others to be sure overwrite
     // any other values.
+    console.log("COUCUO", attributes.dataId);
```
marco-c, line 105: I guess these console.log should be removed :P

src/core/xfa/bind.js
```diff
@@ -86,7 +86,6 @@ class Binder {
     // Here we make the link between form node and
     // data node (through $data property): we'll use it
     // to save form data.
-
```
marco-c, line 89: Nit: unrelated newline removal

marco-c:
Can you run the reftests?

calixteman:
There is an error in `firefox-xfa_imm5257e-page3`: the textedit is drawn after the string so it's fine when the background is transparent but it isn't when it's white.
Maybe something is wrong with the textedit size.

PR #13734, opened 2021-07-14
XFA - Cannot print fields with no names

calixteman:
- it was not possible to print pdf file in issue #13500.


marco-c (approved):
Please make sure there are no reftest regressions.

PR #13735, opened 2021-07-14
Ensure that the field value, for checkboxes, refers to an existing appearance state (bug 1720411)

Snuffleupagus:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1720411

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -2075,6 +2075,11 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       // The /Off appearance is optional.
       exportValues.push("Off");
     }
+    // Don't use a "V" entry pointing to a non-existent appearance state,
+    // see e.g. bug1720411.pdf where it's an *empty* Name-instance.
+    if (!exportValues.includes(this.data.fieldValue)) {
+      this.data.fieldValue = null;
+    }
     if (exportValues.length !== 2) {
       return;
     }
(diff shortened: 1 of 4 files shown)


marco-c:
@Snuffleupagus do you know what caused the regression?

Snuffleupagus:
> do you know what caused the regression?

Are we sure that this actually is a regression, since there wasn't an *exact* provided regression range in the bug (and I've not had time to actually bisect it)?

However, assuming that it's a regression, a potential candidate would be PR #13274 since the changes in https://github.com/mozilla/pdf.js/pull/13274/files#diff-d7b040092cf6be09a276b785ba03c46e79bc8342cf4ae4fb1bc139dea8513425R931-R935 are unlikely to work all that well with broken/corrupt PDF documents that contain *empty* `Name`-instances.

I suppose that we could also have added a work-around on the main-thread, however it seems to me that trying to reference a non-existent appearance state isn't generally correct; at least that's how I'm reading [the PDF specification](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#G11.1952119):
> The **V** entry in the field dictionary (see Table 220) holds a name object representing the check box’s appearance state, which shall be used to select the appropriate appearance from the appearance dictionary.

marco-c:
All right, I'm running mozregression to double check!

marco-c:
https://bugzilla.mozilla.org/show_bug.cgi?id=1711106 is the cause, so I guess it's #13274 like you suggested.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

/botio makeref

PR #13737, opened 2021-07-15
Add GitHub Advance Security and CodeQL

GeekMasher:
Hi team, I've created this PR to add security results from GitHub Advance Security / CodeQL into your workflow and thought the team might appreciate having a look at the results of the security testing directly in GitHub. Currently, it will also scan on all new PR's that are created to check for security issues and it's scheduled to run every week on Tuesday to make sure you are running the latest and greatest rules from CodeQL.

I have ignored the `test` path as it was reporting a few False Positives in the Unit tests.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,4 +1,1 @@
 name: "CodeQL Configuration"
-
-paths-ignore:
-  - 'test/**'
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,4 @@
+name: "CodeQL Configuration"
+
+paths-ignore:
+  - 'test/**'
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -5,8 +5,6 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
-  schedule:
-    - cron: '26 1 * * 2'
 
 jobs:
   analyze:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,40 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron: '26 1 * * 2'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
+        queries: security-extended
+        config-file: ./.github/codeql/codeql-config.yml
+
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v1
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v1


Snuffleupagus:
We're already planning adding the PR-integration of LGTM (see #11965), which as far as I understand uses CodeQL internally, so my question is how this is different and if really need both?

Snuffleupagus:
There's also the question of false positives, which all tools will suffer from[1], and specifically how to ignore those?

---
[1] For example, we're using ESLint extensively and there it's very easy to white-list (on an individual basis) false positives.

GeekMasher:
@Snuffleupagus Both LGTM and GitHub use the same technology (CodeQL) but [GitHub Advance security is built into GitHub](https://github.com/features/security) and the next generation of the tool. You can mark False Positives in GitHub Security tab itself (see screenshot below).

image

*Note: this screenshot is from another project as an example and not based on the results from this project.*

From what I saw of my scan, you only have a few issues that were reported. I removed some false positives already using the configuration file.

FYI, you can also push ESLint results into GitHub Advance Security 👀

timvandermeij:
Interesting. From what I'm reading GitHub Advanced Security/CodeQL is more specifically related to finding potential security issues whereas LGTM is more general about finding anyt defects in code using static analysis (not necessarily security-related). I agree that the advantage of GitHub Advanced Security/CodeQL is that GitHub itself provides it, so it's not really an external tool. If GitHub Advanced Security/CodeQL and LGTM indeed differ in this way (I'm not 100% sure, so please correct me if I'm wrong here), it might make sense to enable them both?

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,4 @@
+name: "CodeQL Configuration"
+
+paths-ignore:
+  - 'test/**'
```
Snuffleupagus, line 4: I don't think that we should be blanket-disabling an *entire* folder like this, even if this is "only" test-code, but white-list false positives individually instead.

Also, is there a reason for mixing single and double quotes in these files?
(In the source code we use *double* quotes, so that would seem preferable here as well.)

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,40 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron: '26 1 * * 2'
```
Snuffleupagus, line 9: Do we really need this, and if so why on Tuesdays specifically, since there's easily *at least* one PR opened/landed every week hence this analysis will run very frequently anyway?

GeekMasher:
@timvandermeij Not necessarily, you can enable all of the [Query Suites](https://codeql.github.com/docs/codeql-cli/creating-codeql-query-suites/) (list of rules) LGTM does in GitHub Advance security (security and quality) but I didn't know if that was something we wanted to start off with.

This can be enabled by using the `security-and-quality` query suite, see [this guide](https://docs.github.com/en/code-security/secure-coding/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#running-additional-queries).

.github/codeql/codeql-config.yml
GeekMasher: @Snuffleupagus I can agree with that and have removed with latest commit. I decided not to remove the configuration file as you might want to add content in the future.

.github/workflows/codeql-analysis.yml
GeekMasher: @Snuffleupagus that works for me. The date is generated automatically by GitHub so Thursday was just random.

Snuffleupagus:
Please note that we *generally* don't use separate fixup commits, to keep the history cleaner and to ensure that all patches are complete and fully working on their own; hence please squash the commits together such that we can test this!

timvandermeij:
Aside from squashing the commits, would it perhaps make sense to enable the `security-and-quality` suite too so we have both security and quality checks managed by GitHub? This might make LGTM integration obsolete, provided that it provides more-or-less the same functionality of course (plus it's likely easier to integrate given that external services need additional steps from project admins to enable them).

Snuffleupagus:
Given that LGTM has been enabled already, and in order to avoid too much scope-creep in *this* PR, maybe we could defer https://github.com/mozilla/pdf.js/pull/13737#issuecomment-882081947 to another time?

timvandermeij:
Yes, good idea. I didn't see that LGTM was already enabled. In that case the current PR should be fine (i.e., the security checks only). @GeekMasher Could you squash the commits into one so we can give this a try?

timvandermeij:
Closing since this is fixed up and merged in #13828 now. Thanks!

GeekMasher:
@timvandermeij apologies for the delay in my replay, I was out for holiday and limited access.

Glad to see the code merged and I hope this is useful going forward

timvandermeij:
No worries! Thank you for introducing CodeQL to us; we're now using it in our build process, configured it to ignore the false positives/irrelevant alerts and fixed the ones that it flagged correctly, so overall it's a very welcome addition.

PR #13738, opened 2021-07-15
Re-factor the handling of *empty* `Name`-instances (PR 13612 follow-up)

Snuffleupagus:
When working on PR #13612, I mostly prioritized a simple solution that didn't require touching a lot of code. However, while working on PR #13735 I started to realize that the static `Name.empty` construction really wasn't a good idea.

In particular, having a special `Name`-instance where the `name`-property isn't actually a String is confusing (to put it mildly) and can easily lead to issues elsewhere. The only reason for not simply allowing the `name`-property to be an *empty* string, in PR #13612, was to avoid having to touch a lot of existing code. However, it turns out that this is only limited to a few methods in the `PartialEvaluator` and a few of the `BaseLocalCache`-implementations, all of which can be easily re-factored to handle *empty* `Name`-instances.

All-in-all, I think that this patch is even an *overall* improvement since we're now validating (what should always be) `Name`-data better in the `PartialEvaluator`.
This is what I ought to have done from the start, sorry about the code churn here!

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -1540,7 +1540,7 @@ class PartialEvaluator {
       timeSlotManager.reset();
 
       const operation = {};
-      let stop, i, ii, cs, name;
+      let stop, i, ii, cs, name, isValidName;
       while (!(stop = timeSlotManager.check())) {
         // The arguments parsed by read() are used beyond this loop, so we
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #6374, opened 2015-08-21
Update viewer.html for Internet Explorer Compatibility

WilliamRClark:
If your Internet Explorer 11 default compatibility settings are set to "I.E 7 Compatibility", the PDF plugin will not load.  This fix is the same one used by AngularJS to force the browser to use HTML5 mode.

PR #13739, opened 2021-07-15
Remove `<meta>` tag only necessary for IE-compatibility

Snuffleupagus:
Given that Internet Explorer is, since some time now, no longer supported in the PDF.js library this `<meta>` tag can also be removed (added in PR #6374).


timvandermeij (approved):

timvandermeij:
Nice find; thanks!

PR #13743, opened 2021-07-16
XFA - Checkboxes must be printed (bug 1720182)

calixteman:
- to avoid future regressions, annotationStorage is passed to the xfa render in reftests.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -288,7 +288,14 @@ var rasterizeXfaLayer = (function rasterizeXfaLayerClosure() {
   }
 
   // eslint-disable-next-line no-shadow
-  function rasterizeXfaLayer(ctx, viewport, xfa, fontRules) {
+  function rasterizeXfaLayer(
+    ctx,
+    viewport,
+    xfa,
+    fontRules,
+    annotationStorage,
+    isPrint
+  ) {
     return new Promise(function (resolve, reject) {
       // Building SVG with size of the viewport.
       const svg = document.createElementNS(SVG_NS, "svg:svg");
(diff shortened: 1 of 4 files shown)


test/test_manifest.json
```diff
@@ -952,6 +952,25 @@
        "enableXfa": true,
        "type": "eq"
     },
+    {  "id": "xfa_bug1720182",
+       "file": "pdfs/xfa_bug1720182.pdf",
+       "md5": "1351f816f0509fe750ca61ef2bd40872",
+       "link": true,
+       "rounds": 1,
+       "enableXfa": true,
+       "type": "eq",
```
Snuffleupagus, line 961: Given that the bug is about printing, shouldn't this also use `"print": true,` as well?

test/test_manifest.json
```diff
@@ -952,6 +952,25 @@
...
+       "annotationStorage": {
+         "RadioButtonList2707": {
+           "value": "1"
+         },
+         "ComplainantFirstname2710": {
+             "value": "Foo"
```
Snuffleupagus, line 967: Nit: Here, and below, there's some inconsistent indentation.

test/driver.js
```diff
@@ -316,6 +316,7 @@ var rasterizeXfaLayer = (function rasterizeXfaLayerClosure() {
             xfa,
             div,
             viewport: viewport.clone({ dontFlip: true }),
+            annotationStorage,
```
Snuffleupagus, line 319: Similar to the comment in `test_manifest.json`, shouldn't this provide the "correct" intent for a print test?
Note that we're doing that for the "regular" rendering tests, when `"print": true,` is set in the manifest.

marco-c (approved):

PR #13745, opened 2021-07-16
Enable XFA by default in the development viewer

Snuffleupagus:
Given that https://bugzilla.mozilla.org/show_bug.cgi?id=1720402 has just landed, enabling XFA in the development viewer probably cannot hurt now.


timvandermeij (approved):

timvandermeij:
Good idea!

Issue #10165, opened 2018-10-19
Implement beginAnnotations and endAnnotations for SVG


valgussev:
Configuration:
- PDF.js version: 2.0.550

Steps to reproduce the problem:
1. Render document in SVG mode `renderMode="svg"`

What is the expected behavior? (add screenshot)
Rendering annotations same way as they rendered on canvas

What went wrong? (add screenshot)
```
Warning: Unimplemented operator beginAnnotations
Warning: Unimplemented operator endAnnotations
```

mishase:
Is there any plans to implement these operators soon?

PR #13746, opened 2021-07-16
[api-minor] Add `intent` support to the `PDFPageProxy.getOperatorList` method (issue 13704)

### LINKED ISSUE #13704: Unable to get the image of an image form element

Configuration:
- Web browser and its version: Chrome 91.0.4472.124
- PDF.js version: 2.8.335
- Is a browser extension: Nope.

I'm trying to list and extract data form each element of a form created in a PDF file.
I have some image elements detected as Btn (button) and without any indice about the image.

`const items = await pdfPage.getAnnotations();

for (const item of items) {

    // With 'TestImage' as the name of one on my image elements
    if (item.fieldName === 'TestImage') {
        console.log('TestImage', item);
    }
}`

Gives:

TestImage 
{annotationFlags: 4, borderStyle: {…}, color: Uint8ClampedArray(3), contents: "", hasAppearance: true, …}
actions: {Action: Array(1)}
alternativeText: ""
annotationFlags: 4
annotationType: 20
borderStyle:
    dashArray: [3]
    horizontalCornerRadius: 0
    style: 1
    verticalCornerRadius: 0
    width: 0
checkBox: false
color: 
Uint8ClampedArray(3) 
[0, 0, 0]
contents: ""
defaultAppearanceData:
    fontColor: Uint8ClampedArray(3) [0, 0, 0]
    fontName: "HeBo"
    fontSize: 12
defaultFieldValue: null
fieldFlags: 65536
**fieldName: "TestImage"**
**fieldType: "Btn"**
fieldValue: null
hasAppearance: true
hidden: false
id: "57R"
isTooltipOnly: false
modificationDate: null
pushButton: true
radioButton: false
readOnly: false
rect: (4) [383.563, 623.486, 527.808, 744.176]
subtype: "Widget"

Any idea?

In front of that, I'm able to get images from regular image fields (out of Forms) without any context (the  location and the size used in the PDF):

```
// Potential image types
const imagesObjectsTypes = [
    pdfjsLib.OPS.paintImageMaskXObject,
    pdfjsLib.OPS.paintImageMaskXObjectGroup,
    pdfjsLib.OPS.paintImageXObject, // <- 85
    pdfjsLib.OPS.paintInlineImageXObject,
    pdfjsLib.OPS.paintInlineImageXObjectGroup,
    pdfjsLib.OPS.paintImageXObjectRepeat,
    pdfjsLib.OPS.paintImageMaskXObjectRepeat,   
];

const operatorsList = await pdfPage.getOperatorList();

for (let i = 0; i < operatorsList.fnArray.length; i++) {

    const type = operatorsList.fnArray[i];

    if (imagesObjectsTypes.indexOf(type) >= 0) {

        const image = pdfPage.objs.get(operatorsList.argsArray[i][0]);

        console.log('Image detected:')
        console.log(image);
        console.log(operatorsList.argsArray[i]);
    }
}

```
Gives

Image detected:
width: 1067,
height: 600,
kind: 2, data: Uint8ClampedArray(1920600)} data: Uint8ClampedArray(1920600) [ …],
height: 600,
kind: 2,
width: 1067,

0: "img_p0_3"
1: 1067
2: 600

Thank you :)

Snuffleupagus:
With this patch, the `PDFPageProxy.getOperatorList` method will now return `PDFOperatorList`-instances that also include Annotation-operatorLists (when those exist). Hence this closes a small, but potentially confusing, gap between the `render` and `getOperatorList` methods.

Previously we've been somewhat reluctant to do this, as explained below, but given that there's actual use-cases where it's required probably means that we'll *have* to implement it now.
Since we still need the ability to separate "normal" rendering operations from direct `getOperatorList` calls in the worker-thread, this API-change unfortunately causes the *internal* renderingIntent to become a bit "messy" which is indeed unfortunate (note the `"oplist-"` strings in various spots). As-is I suppose that it's not all that bad, but we may want to consider changing the *internal* renderingIntent to e.g. a bitfield in the future.

Besides fixing issue #13704, this patch would also be necessary if someone ever tries to implement e.g. issue #10165 (since currently `PDFPageProxy.getOperatorList` doesn't include Annotation-operatorLists).

*Please note:* This patch is *also* tagged "api-minor" for a second reason, which is that we're now including the Annotation-id in the `beginAnnotation` argument. The reason for this is to allow correlating the Annotation-data returned by `PDFPageProxy.getAnnotations`, with its corresponding operatorList-data (for those Annotations that have it).

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -702,7 +702,13 @@ class Annotation {
 
     return resourcesPromise.then(resources => {
       const opList = new OperatorList();
-      opList.addOp(OPS.beginAnnotation, [data.rect, transform, matrix]);
+      opList.addOp(OPS.beginAnnotation, [
+        data.id,
+        data.rect,
+        transform,
+        matrix,
+      ]);
+
       return evaluator
         .getOperatorList({
           stream: appearance,
@@ -1307,6 +1313,7 @@ class WidgetAnnotation extends Annotation {
 
         const transform = getTransformMatrix(this.data.rect, bbox, matrix);
         operatorList.addOp(OPS.beginAnnotation, [
+          this.data.id,
           this.data.rect,
           transform,
           matrix,
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -366,12 +366,16 @@ class Page {
 
         // Collect the operator list promises for the annotations. Each promise
         // is resolved with the complete operator list for a single annotation.
+        const annotationIntent = intent.startsWith("oplist-")
+          ? intent.split("-")[1]
+          : intent;
         const opListPromises = [];
         for (const annotation of annotations) {
           if (
-            (intent === "display" &&
+            (annotationIntent === "display" &&
               annotation.mustBeViewed(annotationStorage)) ||
-            (intent === "print" && annotation.mustBePrinted(annotationStorage))
+            (annotationIntent === "print" &&
+              annotation.mustBePrinted(annotationStorage))
           ) {
             opListPromises.push(
               annotation
--- a/src/core/operator_list.js
+++ b/src/core/operator_list.js
@@ -594,14 +594,14 @@ class OperatorList {
 
   // Close to chunk size.
   static get CHUNK_SIZE_ABOUT() {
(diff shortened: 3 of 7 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13754, opened 2021-07-18
Remove the IR (internal representation) part of the `PDFFunction` parsing

Snuffleupagus:
- Remove the IR (internal representation) part of the `PDFFunction` parsing

   *This follows the exact same princial as PR #12083, but for the `PDFFunction` parsing instead.*

   Given that the IR format is completely unused now, all that the current code does is add a bunch of unnecessary indirection/overhead to the handling of PDF-functions.

 - Convert `PDFFunction` to a standard class with `static` methods

   For e.g. `gulp mozcentral`, the *built* `pdf.worker.js` file decreases from `1 837 608` to `1 834 907` bytes with this patch-series.
   The improvement comes first of all from less overall indentation in `PDFFunction`, and secondly from the removal of (now) unnecessary indirection in the code.

### DIFF
--- a/src/core/function.js
+++ b/src/core/function.js
@@ -132,11 +132,6 @@ function toNumberArray(arr) {
 }
 
 const PDFFunction = (function PDFFunctionClosure() {
-  const CONSTRUCT_SAMPLED = 0;
-  const CONSTRUCT_INTERPOLATED = 2;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Much simpler; thanks!

Issue #13756, opened 2021-07-18
XFA Binding error in Form


JDziurlaj:
Attach (recommended) or Link to PDF file here:
https://services.groups.be/docgenerator/Docgen/FormRequest.aspx?noDocument=90101&noGroupe=10&cdLangue=1
Configuration:

- Web browser and its version: 92.0a1 (2021-07-18) (64-bit)
- Operating system and its version: Windows 10

Steps to reproduce the problem:
1. enableXfa = true
2. Open form

What is the expected behavior? (add screenshot)

What went wrong? (add screenshot)
![image](https://user-images.githubusercontent.com/7104454/126066521-68337be2-1429-403f-a4e5-06dd4b88a457.png)

Document does not load.
`Warning: XFA - an error occurred during parsing and binding: TypeError: can't access property Symbol.iterator
Warning: XFA - an error occurred during layout: TypeError: can't access property Symbol(), this.form is undefined`

Snuffleupagus:
> Document does not load.
> `Warning: XFA - an error occurred during parsing and binding: TypeError: can't access property Symbol.iterator Warning: XFA - an error occurred during layout: TypeError: can't access property Symbol(), this.form is undefined`

Fixing the breaking error *specifically* looks pretty trivial, and I can submit a patch for that in case it's deemed helpful.
(Although there's likely other things wrong with this document and those bugs I probably cannot fix.)

calixteman:
@Snuffleupagus, feel free to fix whatever you want ;): your help is greatly appreciated.
Anyway, the binding ref make a reference to `$record`: as far as I understand specs, it's related to data coming from an external source (see chapter `Exchanging Data Between an External Application and a Basic XFA Form`) so we won't implement anything related to that stuff but if I misunderstood specs please tell me.
And there are few layout issues...

JDziurlaj:
$record allows for windowing a larger dataset, including XML embedded in the xfa data dom. For example, a subset of data can be rendered on to the form at once.

calixteman:
@JDziurlaj, +1 for an example.

PR #13761, opened 2021-07-19
XFA - Prevent breaking errors in `Binder`, when `searchNode` doesn't return data (issue 13756)

Snuffleupagus:
As can be seen in the code (see below), the `searchNode` helper function will return `null` in some cases and all of its call-sites should protect against that before attempting to access the returned data.
While only one of these changes were necessary to fix the breaking errors in issue #13756, in order to prevent future bugs I've added similar defensive code throughout this file.

 - https://github.com/mozilla/pdf.js/blob/07955fa1d31df62fd668a15f58cd40b29a82bd63/src/core/xfa/som.js#L169
 - https://github.com/mozilla/pdf.js/blob/07955fa1d31df62fd668a15f58cd40b29a82bd63/src/core/xfa/som.js#L239
 - https://github.com/mozilla/pdf.js/blob/07955fa1d31df62fd668a15f58cd40b29a82bd63/src/core/xfa/som.js#L254

### DIFF
--- a/src/core/xfa/bind.js
+++ b/src/core/xfa/bind.js
@@ -198,34 +198,36 @@ class Binder {
         continue;
       }
 
-      const [node] = searchNode(
+      const nodes = searchNode(
         this.root,
         dataNode,
         ref,
         false /* = dotDotAllowed */,
         false /* = useCache */
       );
-      if (!node) {
+      if (!nodes) {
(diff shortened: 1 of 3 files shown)


calixteman (approved):
LGTM.
Thanks for doing that.

PR #13762, opened 2021-07-19
Hide a couple of LGTM warnings that we don't care about

Snuffleupagus:
- The `web/ui_utils.js` one is *definitely* a false positive, since we've got e.g. unit-tests that trigger this code-path (and so does the "viewer component" examples).

 - The `src/core/file_spec.js` one seem to be another false positive (at least to me, and we have similar regular expressions elsewhere in the code-base).

 - The `src/display/pattern_helper.js` one is technically correct, but given that we have very similiar code elsewhere in this file I figured that it cannot hurt to leave it as-is.

Note that since Prettier doesn't allow trailing comments on lines ending with a curly bracket, I unfortunately had to use a slightly ugly formatting for some of these LGTM comments.

### DIFF
--- a/src/core/file_spec.js
+++ b/src/core/file_spec.js
@@ -67,7 +67,7 @@ class FileSpec {
     if (!this._filename && this.root) {
       const filename = pickPlatformItem(this.root) || "unnamed";
       this._filename = stringToPDFString(filename)
-        .replace(/\\\\/g, "\\")
+        .replace(/\\\\/g, "\\") // lgtm [js/double-escaping]
         .replace(/\\\//g, "/")
         .replace(/\\/g, "/");
     }
--- a/src/display/pattern_helper.js
+++ b/src/display/pattern_helper.js
@@ -192,7 +192,7 @@ function drawTriangle(data, context, p1, p2, p3, c1, c2, c3) {
       let k;
       if (y < y1) {
         k = 0;
-      } else if (y1 === y2) {
+      } else if (y1 === y2 /* lgtm [js/useless-comparison-test] */) {
         k = 1;
       } else {
         k = (y1 - y) / (y1 - y2);
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -779,7 +779,7 @@ class EventBus {
         this._off(eventName, listener);
       }
       if (external) {
-        (externalListeners ||= []).push(listener);
+        (externalListeners ||= new Set()).add(listener);
         continue;
       }
       listener.apply(null, args);


timvandermeij:
Could you elaborate a bit on why this is closed? I noticed that this didn't change anything in the LGTM report, but I wonder why since I think this should work and moreover we would like to get rid of those false positives to have a clean report (so we have a clean baseline and any new things show up).

Snuffleupagus:
> Could you elaborate a bit on why this is closed?

Besides the mistake outlined below, I'm also really disliking the format that I had to resort to for the comments.

> I noticed that this didn't change anything in the LGTM report, but I wonder why since I think this should work

That's my though as well, but now that I'm actually reading https://lgtm.com/help/lgtm/alert-suppression properly there's this:
> If you've enabled automated code review of pull requests for your projects, it's important to note that alert suppression changes made in a pull request are only applied when that pull request is merged.

So now I'm feeling stupid :-)

> and moreover we would like to get rid of those false positives to have a clean baseline.

For the PR-integration itself I'm not sure if it matters to have the warnings at exactly zero, since what it reports is simply the *delta* between `master` and the patch. However, I suppose that getting it to zero might be a goal in itself!?

Another thing that I noticed is that by default a lot of things looking like external- and/or test-code is excluded from the LGTM report; see https://lgtm.com/projects/g/mozilla/pdf.js/latest/files/?sort=name&dir=ASC&mode=heatmap&showExcluded=true for a more complete picture.
Hence I'm thinking that we probably want to make sure that at least *some* of our test-code is covered as well, since logical errors e.g. in a unit-test would likely be something that you'd want to know about.

---

I'll probably try to enable the LGTM reports for a bit more of the code-base first, and then come back to this PR later!

Snuffleupagus:
As I somewhat suspected here, the existing warning in the `web/ui_utils.js` file is actually a bug in CodeQL/LGTM since it's apparently unable to correctly deal with Logical OR assignment (`||=`) operators.

calixteman:
About the comparison: we've `y<y2` and `y1<=y` (because we're in the else) so `y1<y2` and so `y1 !== y2`. So I think LGTM is good here or am I missing something ?

Snuffleupagus:
> So I think LGTM is good here or am I missing something ?

Yes, that one is correct, however I'm not (yet) sure if that points a bug in our code (i.e. the condition should look differently) or if we can just remove that branch and not worry about it!?

timvandermeij:
If the condition is clearly incorrect and/or confusing, I'd say we remove the branch, provided of course that no test cases fall over. I've also been struggling to get my head around that statement after LGTM reported it, so at the very least I'd say it's confusing code.

Issue #13748, opened 2021-07-17
XFA Form fails to load with data error


JDziurlaj:
Attach (recommended) or Link to PDF file here: [IFT-3R(7)_v1-1E.pdf](https://github.com/mozilla/pdf.js/files/6849364/IFT-3R.7._v1-1E.pdf)
http://www.finanse.mf.gov.pl/documents/766655/1481810/IFT-3R(7)_v1-1E.pdf
Configuration:

- Web browser and its version: 92.0a1 (2021-07-17) (64-bit)
- Operating system and its version: Windows 10

Steps to reproduce the problem:
1. enableXfa = true
2. Open form

What is the expected behavior? (add screenshot)
PDF Loads
![image](https://user-images.githubusercontent.com/7104454/126037353-c641793d-c77f-47de-bdab-af003d1fc275.png)

What went wrong? (add screenshot)
console error `data[_xfa_object.$getDataValue] is not a function`

PR #13764, opened 2021-07-20
XFA - Add a missing method to `XFAAttribute`, to prevent breaking errors (issue 13748)

Snuffleupagus:
*This is yet another case where I've got no idea if the patch is correct, but it does at least fix a breaking error :-)*

Note how in the [`Binder._bindValue` method](https://github.com/mozilla/pdf.js/blob/683ce66a48cd54fa48e96e1925277c134859e76c/src/core/xfa/bind.js#L92-L93), we're assuming that if a `data`-value exists then it'll also be possible to actually access it. For the `XFAAttribute`-implementation however, the second method is missing and that's what causes the breaking errors in issue #13748.

Please note that another possible way of "fixing" the error wouldn't been to simply change the exists-check to return `false`, and I could see that being a preferred solution.
However, the reason for submitting the current patch is that we get *fewer* warnings about Nodes with mis-matched types this way.

### DIFF
--- a/src/core/xfa/xfa_object.js
+++ b/src/core/xfa/xfa_object.js
@@ -745,6 +745,10 @@ class XFAAttribute {
     return true;
   }
 
+  [$getDataValue]() {
+    return this[$content].trim();
+  }
+
   [$setValue](value) {
     value = value.value || "";
     this[$content] = value.toString();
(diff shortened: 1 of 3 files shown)


calixteman (approved):
In the pdf in the issue, the failing ref is `$.Naglowek.KodFormularza.kodSystemowy` and the node in data tree is `<KodFormularza kodSystemowy="IFT-3R (7)".../>`.
So this patch is the right thing to do.
Thank you.

marco-c:
Is this fixing #13748 completely?

Snuffleupagus:
> Is this fixing #13748 completely?

Unfortunately not, since there's a bunch of scripting in that document as well (I've updated the tags accordingly).

Issue #13751, opened 2021-07-17
XFA Form causes layout error


JDziurlaj:
Attach (recommended) or Link to PDF file here:
[Oswiadczenie_o_statusie_FATCA_i_CRS-1.pdf](https://github.com/mozilla/pdf.js/files/6834770/Oswiadczenie_o_statusie_FATCA_i_CRS-1.pdf)

Configuration:

- Web browser and its version: 92.0a1 (2021-07-17) (64-bit)
- Operating system and its version: Windows 10

Steps to reproduce the problem:
1. enableXfa = true
2. Open form

What is the expected behavior? (add screenshot)
PDF Loads
![image](https://user-images.githubusercontent.com/7104454/126038784-2b81e396-f31b-48aa-b2a3-061441c017d1.png)

What went wrong? (add screenshot)
Console error of `Warning: XFA - an error occurred during layout: TypeError: can't access property "w", contentArea is undefined` (nothing loads)

PR #13766, opened 2021-07-21
XFA - Handle `startIndex` correctly in the `Template.$toHTML` method (issue 13751)

Snuffleupagus:
*Please note:* The PDF document in issue #13751 is *dynamically* created (in e.g. Adobe Reader), with pages added when certain buttons are clicked, hence this patch simply fixes the breaking error and nothing more.

It looks like the current code contains a little bit too much copy-and-paste from the *similar* `index` branch above, since we cannot set the `startIndex` to a negative value. Note how it's being used to initialize the loop-variable, which is then used to lookup values in an Array and accessing the `-1`th element of an Array obviously makes no sense.

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -5288,13 +5288,14 @@ class Template extends XFAObject {
           } else if (target instanceof ContentArea) {
             const index = contentAreas.findIndex(e => e === target);
             if (index !== -1) {
+              // In the next loop iteration `i` will be incremented, note the
+              // `continue` just below, hence we need to subtract one here.
               i = index - 1;
             } else {
               targetPageArea = target[$getParent]();
-              startIndex =
-                targetPageArea.contentArea.children.findIndex(
-                  e => e === target
-                ) - 1;
+              startIndex = targetPageArea.contentArea.children.findIndex(
+                e => e === target
+              );
             }
           }
           continue;
--- a/test/pdfs/issue13751.pdf.link
+++ b/test/pdfs/issue13751.pdf.link
@@ -1,0 +1,1 @@
+https://github.com/mozilla/pdf.js/files/6834770/Oswiadczenie_o_statusie_FATCA_i_CRS-1.pdf
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -1356,6 +1356,15 @@
        "enableXfa": true,
        "type": "eq"
     },
+    {  "id": "issue13751",
+       "file": "pdfs/issue13751.pdf",
+       "md5": "b9960bfa99be3573d9c32a9d55e5b7da",
+       "link": true,
+       "rounds": 1,
+       "lastPage": 1,
+       "enableXfa": true,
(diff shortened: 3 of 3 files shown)


src/core/xfa/template.js
```diff
@@ -5288,13 +5288,17 @@ class Template extends XFAObject {
...
             } else {
               targetPageArea = target[$getParent]();
-              startIndex =
-                targetPageArea.contentArea.children.findIndex(
-                  e => e === target
-                ) - 1;
+              const childIndex = targetPageArea.contentArea.children.findIndex(
+                e => e === target
+              );
+              if (childIndex !== -1) {
```
calixteman, line 5299: This check should be useless: the target (contentArea) is necessarly a children of its parent.

calixteman: And in case it's possible to have -1 I'd prefer having an error, but I'm sure it can't.

Snuffleupagus: > This check should be useless: the target (contentArea) is necessarly a child of its parent.

The reason for this added check, based on no actual knowledge of either this code or the XFA specification was:
 - To prevent issues if the `children` Array would ever be empty; although I'm not sure if that can ever happen for real-life XFA documents (e.g. corrupt ones)?
 - To prevent issues if the specific `target` cannot be found in the `children` Array; again I'm not sure if that can happen for real-life XFA documents (e.g. corrupt ones)?

If neither of those error-case can ever happen, I'll just remove it this check; otherwise throwing a `FormatError` is probably appropriate.
@calixteman Please let me know what you prefer, thank you!

calixteman: A `contentArea` can only be under a `pageArea`. I mean it isn't possible to have in the JS representation of a node something which isn't defined in the constructor, so even with ill-formatted xml it's impossible to have a `contentArea` not under a `pageArea`. The constructor of each class in `template.js` must be the exact reflection of the xml node as we have in the specs: so normally, I mean modulo bugs, the xml is validated. 
So in this case the `$getParent` will always return a `pageArea` and this one must contain one of its child (i.e. the `contentArea`). The parent is attached to its child when the child is accepted by its future parent:
https://github.com/mozilla/pdf.js/blob/6c9b6bc59999cfe7253ecc1323f38210b8739cd0/src/core/xfa/xfa_object.js#L132
So having a -1 here means that the parent is not really the parent and something is really wrong in the code but not in the pdf.
So I'd just remove the check but if you feel that's better to throw an exception in this case I'm fine with that too.

Snuffleupagus: Thank you for the in-depth explanation!

> So having a -1 here means that the parent is not really the parent and something is really wrong in the code but not in the pdf.

Given that this suggests a bug in our code, and not simply a corrupt XFA document, I'm completely fine with just setting the `startIndex` and not worry about the `-1` case. I'll update the patch accordingly, thanks!

calixteman (approved):
LGTM. Thanks.

PR #13768, opened 2021-07-21
XFA - Remove namespace from nodes under xfa:data node

calixteman:
- in real life some xfa contains xml like \<xfa:data>\<xfa:Foo>\<xfa:Bar>...</xfa:data>
    since there are no Foo or Bar in the xfa namespace the JS representation are empty
    and that leads to errors.
  - so the idea is to make all nodes under xfa:data namespace agnostic which means
    that ns are removed from nodes in the parser but only for xfa:data descendants.

### DIFF
--- a/src/core/xfa/builder.js
+++ b/src/core/xfa/builder.js
@@ -18,6 +18,7 @@ import {
   $cleanup,
   $finalize,
   $ids,
+  $isNsAgnostic,
   $nsAttributes,
   $onChild,
   $resolvePrototypes,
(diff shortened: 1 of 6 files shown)


src/core/xfa/builder.js
```diff
@@ -67,6 +68,7 @@ class Empty extends XFAObject {
 class Builder {
   constructor() {
     this._namespaceStack = [];
+    this._nsAgnosticStack = [];
```
Snuffleupagus, line 71: *This is possibly a stupid question on my part, however:*
Are you planning of adding anything except simply boolean values to this Array?

If not, it may not be necessary to use an Array and an integer (e.g. `this._nsAgnosticLevel` or similar) that you just increment/decrement might work just as well!?

Snuffleupagus (approved):
Looks good to me, thanks!

PR #13770, opened 2021-07-21
Improve performance of reused patterns.

brendandahl:
Bug 1721218 has a shading pattern that was used thousands of times.
To improve performance of this PDF:
 - add a cache for patterns in the evaluator and only send the IR form once
   to the main thread (this also makes caching in canvas easier)
 - cache the created canvas radial/axial patterns
 - for shading fill radial/axial use the pattern directly instead of creating temporary
   canvas


src/core/evaluator.js
```diff
@@ -1371,16 +1401,15 @@ class PartialEvaluator {
...
             resources,
-            this.handler,
-            this._pdfFunctionFactory,
-            localColorSpaceCache
-          );
-          operatorList.addOp(fn, pattern.getIR());
+            localColorSpaceCache,
+            localPatternCache,
+          });
+          operatorList.addOp(fn, ["Shading", objId]);
```
Snuffleupagus, line 1412: To prevent any future issues, I believe that you also want to add `operatorList.addDependency(objId);` *before* this line since you're no longer sending the data inline here.

src/core/evaluator.js
```diff
@@ -1909,18 +1941,14 @@ class PartialEvaluator {
...
-              self.handler,
-              self._pdfFunctionFactory,
-              localColorSpaceCache
-            );
-            var patternIR = shadingFill.getIR();
-            args = [patternIR];
+              localColorSpaceCache,
+              localPatternCache,
+            });
+            args = [patternId];
```
Snuffleupagus, line 1951: As mentioned above, you probably want to add a `operatorList.addDependency(patternId);` before this line.

src/core/evaluator.js
```diff
@@ -1513,6 +1542,7 @@ class PartialEvaluator {
     const localColorSpaceCache = new LocalColorSpaceCache();
     const localGStateCache = new LocalGStateCache();
     const localTilingPatternCache = new LocalTilingPatternCache();
+    const localPatternCache = new Map();
```
Snuffleupagus, line 1545: Would it perhaps help to implement a more "proper" cache here, see the existing ones in https://github.com/mozilla/pdf.js/blob/master/src/core/image_utils.js, since that may help improve overall caching since you could then cache by e.g. actual `Ref`s instead?

(This may, or may not, help all that much for the particular data in question, but I wanted to ask.)

src/display/pattern_helper.js
```diff
@@ -56,41 +56,20 @@ class RadialAxialShadingPattern extends BaseShadingPattern {
...
-      tmpCtx.setTransform.apply(tmpCtx, owner.baseTransform);
-      if (this._matrix) {
-        tmpCtx.transform.apply(tmpCtx, this._matrix);
-      }
-    } else {
-      tmpCtx.setTransform.apply(tmpCtx, ctx.mozCurrentTransform);
-    }
-    applyBoundingBox(tmpCtx, this._bbox);
-
+  createGradient(ctx) {
```
Snuffleupagus, line 62: Nit: Perhaps make this "private" with a leading underscore in the name?

src/core/evaluator.js
Snuffleupagus: Also, perhaps `localShadingPatternCache` instead to provide a slightly clearer distinction between this new cache and the existing one just above?

src/core/evaluator.js
```diff
@@ -1314,6 +1314,35 @@ class PartialEvaluator {
     });
   }
 
+  parsePattern({
```
Snuffleupagus, line 1317: Nit: Maybe `parseShading` instead.

src/core/evaluator.js
brendandahl: I started out using the localTilingPatternCache, but found some issues with it and and was planning to file some bugs. The patterns can't be cached by name alone since they could be in different resources dictionaries with duplicate names (issue6737.pdf has an example, but with radial gradients). There also seemed to be a number of pdfs where the pattern/shading dictionary is inline and caching by ref can't be used. 

As for perf, I don't think caching by ref should speed it up much, the shading/pattern dict should be in the xref's cache map, so it should be a pretty fast lookup.

Snuffleupagus: > I started out using the localTilingPatternCache,

You probably would've needed to implement a new and *separate* cache, rather than simply re-using the existing `localTilingPatternCache` for the Shadings-case that this patch is about.

> but found some issues with it and and was planning to file some bugs.

What sort of bugs did you encounter?

> There also seemed to be a number of pdfs where the pattern/shading dictionary is inline and caching by ref can't be used.
>
> As for perf, I don't think caching by ref should speed it up much, the shading/pattern dict should be in the xref's cache map, so it should be a pretty fast lookup.

I wasn't really worried about the `Dict` lookups themselves, since as you say the data is being cached, but was mostly wondering if caching by `Ref` could possibly have improved the hit/miss ratio in the cache. It doesn't sound like that'd been the case here though :-)

brendandahl: Tiling patterns and shadings can both be patterns, so I should have been able to re-use that cache. I filed #13780 to further explain the issue.

src/core/evaluator.js
brendandahl: The pattern should always be sent before the id is created, so a dependency pause shouldn't be needed. If it isn't sent first there will be an error on the main thread when we try to fetch the object (I did this accidentally).

Snuffleupagus (approved):
Looks good to me, thank you!

PR #13773, opened 2021-07-22
[api-minor] Exposed properties needed for geo spatial extension

m0ose:
GeoPDFs are a popular and attractive way to distribute maps. They are full featured, self contained, and easy to use. Unfortunately there are not many if any good javascript viewers. 

In order to read the geo-spatial part of GeoPDFs one must have access to the "measure" property of the document, and the "Bbox" property in "VP". This code exposes the properties defined by the geo spatial extensions from section 8.8.1 of the Adobe ISO32000 supplement, ~~https://www.adobe.com/content/dam/acom/en/devnet/pdf/adobe_supplement_iso32000.pdf~~.   
https://web.archive.org/web/20211202175942/https://www.adobe.com/content/dam/acom/en/devnet/pdf/adobe_supplement_iso32000.pdf

An example html page is included, called geo.html, which I am sure is in the wrong directory. Here is a live version to try out https://simtable-static-web-apps.s3.us-west-1.amazonaws.com/pdf.js/examples/geo.html .

I am new to this code base. It turns out references could not transfer from the worker so I introduced a function called _serializeReferences which did some recursive dereferencing. I am wondering if there is an existing way to do this.


examples/geo.html
```diff
@@ -0,0 +1,108 @@
+<html>
```
brendandahl, line 1: Instead of adding an example, it would be better to add a test PDF and unit test for this new API. See `api_spec.js`.

brendandahl (changes requested):
Please revert your changes to package-lock.json and add tests.

Snuffleupagus (changes requested):
In addition the previous comments, I'm also a bit worried about the fact that this adds a bunch of code which is completely unused/untested in the PDF.js library/viewer (since it's easy to add code, but then it's got to be maintained).

Furthermore, I'm sorry to say that as-is the implementation unfortunately doesn't really look correct or acceptable. Basically, all the existing API-methods will return carefully validated data from the worker-thread, whereas this simply tries to serialize "anything" and doesn't properly fetch *or* validate any of the data. (The latter can/will lead to all sorts of issues, e.g. with unclonable data when sending it from the worker-thread to the main-thread. There's also the potential issues with cyclical references, which can be notoriously difficult to avoid. Furthermore, in a corrupt PDF document you *could* end up trying to serialize/send basically the *entire* document like this.)

Assuming that we agree to add this kind of functionality to the PDF.js API, the worker-thread implementation would need to be more-or-less completely re-written to both fetch *and* carefully validate all of the expected data in accordance with the specification, to prevent any issues there.

m0ose:
Thank you for the feedback. Honestly, I expected something along these lines. I will work on these changes and get back to you.

m0ose:
I reverted the package.json and thoroughly validated the data as requested.
I may have gone a bit overboard in the data validation. It can now also return the rectilinear measurement, which is in the spec but apparently rarely used. The code size grew a bit so I put it into a separate file.

m0ose:
@Snuffleupagus How does this look?

timvandermeij:
Closing since this PR is over 1.5 years old now, quite outdated and not a feature that's currently planned. Thanks.

PR #13775, opened 2021-07-22
Remove some duplication in the `Dict.merge` method

Snuffleupagus:
Currently the `!mergeSubDicts` code-path is essentially just duplicated code, which we can easily avoid by simply moving that check. (This may lead to ever so slightly more parsing for this case, but the difference ought to be negligible in practice.)


timvandermeij (approved):

timvandermeij:

PR #13776, opened 2021-07-22
Include most of the `test`-folder in the LGTM report (PR 13772 follow-up)

Snuffleupagus:
Given that PR #13772 seem to have worked as intended, it probably cannot hurt to enable the LGTM report for *most* of the remaining code in the `test`-folder. (The one remaining exception is code which originated *outside* of the PDF.js project.)

*Please note:* We'll need to land this patch to actually see any difference in the LGTM results.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,8 +1,7 @@
 path_classifiers:
   test:
-    - exclude: test/font
-    - exclude: test/integration
-    - exclude: test/unit
+    - exclude: test
+    - test/resources
 
 queries:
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
I'm going to make an exception and land this without waiting for review, since:
 - We cannot tell if this actually works as intended, without landing the patch.
 - The patch is very small and simple, and only applies to the LGTM-integration.
 - Finally, and most importantly, none of this code applies to the built PDF.js library and cannot possibly affect rendering, the viewer, or any other even remotely critical functionality.

PR #13777, opened 2021-07-22
XFA - SVG elements mustn't get any pointer events (bug 1721589)

calixteman:
- a rectangle can be defined after a field and so from a z-index pov, it's on top of the field, which means that the rectangle avoid to have some mouse events in the fields;
  - so this patch just disable pointer events for all elements under svg (included).

### DIFF
--- a/web/xfa_layer_builder.css
+++ b/web/xfa_layer_builder.css
@@ -55,6 +55,14 @@
   pointer-events: none;
 }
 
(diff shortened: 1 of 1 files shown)


marco-c:
I've tested it with the PDF from bug 1721589, it's working well.

marco-c (approved):

PR #13778, opened 2021-07-22
XFA - A field without an ui must provide a default one (bug 1718245)

calixteman:


src/core/xfa/template.js
```diff
@@ -2548,6 +2548,34 @@ class Field extends XFAObject {
...
+      //   the application chooses a default user interface for the
+      //   container, based on the type of the container's content.
+
+      this.ui = new Ui({});
+      this.ui[$globalData] = this[$globalData];
+      this[$appendChild](this.ui);
+      let node;
+
+      switch (this.items.children.length) {
+        case 0:
```
marco-c, line 2563: Can you explain your reasoning about the choice of the field with a comment?

Snuffleupagus:
Also, please note that as outlined in https://github.com/mozilla/pdf.js/pull/13778#issuecomment-885129291 there's unfortunately a number of breaking errors in this patch; one such example being:
```
TEST-UNEXPECTED-FAIL | test failed xfa_bug1718725 | in firefox | page1 round 1 | Loading PDF document: UnknownErrorException: can't access property "length", this.dims is undefined
```

marco-c (approved):

PR #13782, opened 2021-07-23
[api-minor] XFA - Remove the `xfaLayer` from the DOM when resetting pages (bug 1721977, PR 13427 follow-up)

Snuffleupagus:
Originally the `xfaLayer` wasn't implemented in such a way that it supported being removed from the DOM when pages were evicted from the cache, however this limitation was lifted in PR #13427 and the `xfaLayer` should thus be handled similar to e.g. the `annotationLayer`.

In addition to removing the `xfaLayer` from the DOM, this patch *also* implements proper rendering/hiding-handling for it (mirroring the `annotationLayer`-code).

*Please note:* This patch is tagged API-minor just in case[1], since it changes the signatures of a couple of `PDFPageView`-methods to improve readability of the code.

---
[1] Although users are *hopefully* not directly accessing any of the affected methods, and are rather using e.g. `PDFViewer` in which case none of these changes will matter.

### DIFF
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -207,33 +207,36 @@ class PDFPageView {
     this.zoomLayer = null;
   }
 
-  reset(keepZoomLayer = false, keepAnnotations = false) {
-    this.cancelRendering(keepAnnotations);
+  reset({
+    keepZoomLayer = false,
+    keepAnnotationLayer = false,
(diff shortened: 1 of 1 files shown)


calixteman (approved):
The way to handle annotation & xfa layers is now consistent thanks to this patch: r+.

PR #13786, opened 2021-07-24
Remove a couple of small closures in `src/core/` code

Snuffleupagus:
This patch-series decreases the size of the *built* `pdf.worker.js` file slightly, and we've been making similar changes elsewhere in the code-base.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13787, opened 2021-07-24
Fix (most) LGTM warnings

Snuffleupagus:
Most of the warnings we don't really care about, and those are simply white-listed using inline comments; however two cases prompted actual code changes:

 - In `src/display/pattern_helper.js` the branch in question is indeed unreachable, and should thus be safe to remove. (This code originated in PR #4192, which is now over seven years ago.)

 - In `test/test.js`, the function in question indeed doesn't accept any arguments. (The patch also re-formats a string just above, which didn't seem worthy of a separated patch.)

This now leaves only *one* warning in the LGTM report, however that one is a false positive that we'll need to report upstream.

### DIFF
--- a/src/display/pattern_helper.js
+++ b/src/display/pattern_helper.js
@@ -192,8 +192,6 @@ function drawTriangle(data, context, p1, p2, p3, c1, c2, c3) {
       let k;
       if (y < y1) {
         k = 0;
-      } else if (y1 === y2) {
-        k = 1;
       } else {
         k = (y1 - y) / (y1 - y2);
       }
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Nice work!

src/display/pattern_helper.js
```diff
@@ -192,8 +192,6 @@ function drawTriangle(data, context, p1, p2, p3, c1, c2, c3) {
       let k;
       if (y < y1) {
         k = 0;
-      } else if (y1 === y2) {
```
calixteman, line 195: My feeling is that this guard has been added just to avoid to divide by zero.

timvandermeij:
It looks like false positives in CodeQL (the "backend" for LGTM and GitHub Advanced Security) can be reported here: https://github.com/github/codeql

The original Semmle repository (creators of LGTM) now also redirects to that repository after being acquired by GitHub (see https://github.com/Semmle/ql/).

timvandermeij:
I have reported the false positive to CodeQL at https://github.com/github/codeql/issues/6362 now.

PR #13788, opened 2021-07-24
Enable the `no-empty-character-class` and `no-template-curly-in-string` linting rules

timvandermeij:
The former prevents unintended errors in regexes. The latter is something I ran into a couple of times before editors had proper support for syntax highlighting in template literals, and it's an easy mistake to make, especially when converting older code to use template literals. Nowadays most editors display template literals in different colors, but nevertheless it seems like a good idea to enable this rule.

Fortunately both rules don't require any code changes, so it's more for prevention.

Refer to https://eslint.org/docs/rules/no-empty-character-class and https://eslint.org/docs/rules/no-template-curly-in-string for more information about those rules.


Snuffleupagus (approved):
More linting is always a good idea; thank you!

PR #13790, opened 2021-07-25

Snuffleupagus:


Snuffleupagus:
@timvandermeij What do you think about creating a new release now, since the last one was almost two months ago and there's been a steady stream of bug fixes and improvements lately?

/botio test

timvandermeij (approved):

timvandermeij:
Thanks!

Making a release is probably a good idea indeed. I'll see if I have time for that today.

PR #13796, opened 2021-07-26
Allow `StreamsSequenceStream.readBlock` to skip sub-streams with errors (issue 13794)

### LINKED ISSUE #13794: PDF missing image
[Scan-to-Mail-PDF1_(1).pdf](https://github.com/mozilla/pdf.js/files/6876708/Scan-to-Mail-PDF1_.1.pdf)

Configuration:
- Web browser and its version: Firefox Developer 91.0b7 (64-Bit)
- Operating system and its version: Windows 10 (21H1)
- PDF.js version: 2.10.339 (also tested with the online demo, Version 2.11.5)
- Is a browser extension:

1. Open the attached pdf
2. Scroll to page 3

image

image

Snuffleupagus:
This patch makes use of the existing `ignoreErrors` option, thus allowing a page to continue parsing/rendering even if (some of) its sub-streams are corrupt. Obviously this may cause *part* of a page to be broken/missing, however it should be better than (potentially) rendering nothing.
Also, to the best of my knowledge, this is the first bug of its kind that we've encountered.

To avoid having to pass in a bunch of, for a `BaseStream`-instance, mostly unrelated parameters when initializing a `StreamsSequenceStream`-instance, I settled on utilizing a callback function instead to allow conditional Error-suppression.
Note that the `StreamsSequenceStream`-class is a *special* stream-implementation that we only use when the `/Contents`-entry, in the `/Page`-dictionary, consists of an Array with streams.

### DIFF
--- a/src/core/decode_stream.js
+++ b/src/core/decode_stream.js
@@ -127,7 +127,7 @@ class DecodeStream extends BaseStream {
 }
 
 class StreamsSequenceStream extends DecodeStream {
-  constructor(streams) {
+  constructor(streams, onError = null) {
     let maybeLength = 0;
     for (const stream of streams) {
       maybeLength +=
@@ -138,6 +138,7 @@ class StreamsSequenceStream extends DecodeStream {
     super(maybeLength);
 
     this.streams = streams;
+    this._onError = onError;
   }
 
   readBlock() {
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

/botio makeref

PR #13804, opened 2021-07-27
Move the `compatibilityParams` into the `web/app_options.js` file

Snuffleupagus:
Given that we've over time been reducing the number of `compatibilityParams` in use, there's now few enough left that I think it makes sense to simply inline them directly in the `web/app_options.js` file.
Note that we recently inlined/removed the separate `src/display/api_compatibility.js` file, see PR #13525, and that it (in my opinion) thus makes sense to do the same in the `web/`-folder. This patch will also slightly reduce the size of *built* `web/viewer.js` file, which cannot hurt.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13806, opened 2021-07-27
XFA - Fix auto-sized fields (bug 1722030)

### LINKED ISSUE #13747: XFA Subforms incorrectly colored
https://www.nbp.pl/badania/projekty_badawcze/4kbe/formularz_zgloszeniowy_zespolowy.pdf

Configuration:

- Web browser and its version: 92.0a1 (2021-07-17) (64-bit)

1. enableXfa = true
2. Open form

Subforms (not including fields) are grey
image

Everything is grey (underneath field highlight overlay) 
image

calixteman:
- In order to better compute text fields size, use line height with no gaps (and consequently guessed height for text are slightly better in general).
  - Fix default background color in fields.

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -2926,16 +2926,22 @@ class Fill extends XFAObject {
     const ggrandpa = grandpa[$getParent]();
     const style = Object.create(null);
 
+    // Use for color, i.e. #...
     let propName = "color";
+
+    // Use for non-color, i.e. gradient, radial-gradient...
+    let altPropName = propName;
+
     if (parent instanceof Border) {
-      propName = "background";
+      propName = "background-color";
+      altPropName = "background";
       if (ggrandpa instanceof Ui) {
         // The default fill color is white.
-        style.background = "white";
+        style.backgroundColor = "white";
       }
     }
     if (parent instanceof Rectangle || parent instanceof Arc) {
-      propName = "fill";
+      propName = altPropName = "fill";
       style.fill = "white";
     }
 
@@ -2948,12 +2954,16 @@ class Fill extends XFAObject {
         continue;
       }
 
-      style[propName] = obj[$toStyle](this.color);
+      const color = obj[$toStyle](this.color);
+      if (color) {
+        style[color.startsWith("#") ? propName : altPropName] = color;
+      }
       return style;
     }
 
-    if (this.color) {
-      style[propName] = this.color[$toStyle]();
+    if (this.color && this.color.value) {
+      const color = this.color[$toStyle]();
+      style[color.startsWith("#") ? propName : altPropName] = color;
     }
 
     return style;
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -97,6 +97,7 @@ import {
   HTMLResult,
 } from "./utils.js";
 import { stringToBytes, Util, warn } from "../../shared/util.js";
+import { getMetrics } from "./fonts.js";
 import { searchNode } from "./som.js";
 
 const TEMPLATE_NS_ID = NamespaceIds.template.id;
@@ -114,6 +115,30 @@ const MAX_ATTEMPTS_FOR_LRTB_LAYOUT = 2;
 // so in order to avoid that (and avoid a OOM crash) we break
 // the loop after having MAX_EMPTY_PAGES empty pages.
 const MAX_EMPTY_PAGES = 3;
+
+function getBorderDims(node) {
+  if (!node || !node.border) {
+    return { w: 0, h: 0 };
+  }
+
+  const borderExtra = node.border[$getExtra]();
+  if (!borderExtra) {
+    return { w: 0, h: 0 };
+  }
+
+  return {
+    w:
+      borderExtra.widths[0] +
+      borderExtra.widths[2] +
+      borderExtra.insets[0] +
+      borderExtra.insets[2],
+    h:
+      borderExtra.widths[1] +
+      borderExtra.widths[3] +
+      borderExtra.insets[1] +
+      borderExtra.insets[3],
+  };
+}
 
 function hasMargin(node) {
   return (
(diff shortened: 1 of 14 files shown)


calixteman:
A lot of failures are expected, overall it's an improvement, but of course if something is really bad, tell me.

/botio test

src/core/xfa/template.js
```diff
@@ -773,31 +798,38 @@ class Border extends XFAObject {
...
+        insets[2] = this.margin.bottomInset;
+        insets[3] = this.margin.leftInset;
+      }
+      this[$extra] = { widths, insets, edges };
     }
+    return this[$extra];
+  }
 
+  [$toStyle]() {
+    // TODO: incomplete.
```
marco-c, line 825: Can you expand?

marco-c:
There are some timeouts.

marco-c:
Still some errors

marco-c:
There are some errors on Windows, maybe intermittent?

On `firefox-xfa_bug1718670_1-page3`, there is a field which was highlighted and isn't anymore, is it still fillable?
Same for `firefox-xfa_imm5257e-page1`, `firefox-xfa_bug1716980-page1` and following pages of the same PDF, `firefox-xfa_annual_expense_report-page5`, `firefox-xfa_issue13556-page3`, `firefox-xfa_issue13213-page5`, `firefox-xfa_issue13213-page6`, `firefox-xfa_imm1295e-page1`, `firefox-xfa_issue13500-page1`, `firefox-xfa_issue13500-page2`.

A button at the end of `firefox-issue13751-page1` disappeared.

Some text in `firefox-xfa_issue13500-page1` is now hidden behind a field.

calixteman:
About the field colors:
  - some fields have no `<border><fill...` which means that by default they're transparent and in this case the bg color is taken from the css (unfocused: violet & focused: transparent);
  - some fields have `<border><fill/>...` which means that the background color is white by default and then the bg color is set through the style attribute. Then colors from the css are overwritten and there are no more distinction between focused/unfocused.

There are several solutions here:
  - do nothing: it's the actual state
  - add some js callbacks to change the colors on focus/blur events;
  - add a `background-image`: 1x1 with the violet-transparent color and remove it when the field has the focus;
  - other ideas ??

About `firefox-issue13751-page1`: the button moved on the next page but the test is only with first page (see lastPage attribute in test_manifest.json). Looking at this pdf, it has some layout issues but unrelated to this patch.
About `firefox-xfa_issue13500-page1`: it's on purpose: I checked in Acrobat & Masterpdf: some checkboxes are behind the text field.

Snuffleupagus:
> There are several solutions here:
> * do nothing: it's the actual state

That'd be *my* definite preference, since it seems like the simplest approach; unless this is a widespread problem in practice.

> * add some js callbacks to change the colors on focus/blur events;
> * add a `background-image`: 1x1 with the violet-transparent color and remove it when the field has the focus;

All of these ideas sound like they'd easily add a bunch of complexity (and overhead), thus negatively impacting code readability/maintainability.

marco-c:
> That'd be my definite preference, since it seems like the simplest approach; unless this is a widespread problem in practice.

My worry is that if some fields have it and some fields don't, users might think the white ones are not fillable. It seems to be occurring in many PDFs (Canadian ones included, and they're between the most important we want to support).

calixteman:
My opinion as a user is that we must be consistent.
So either we put this violet bg on every fillable field in trying to minimize the cost (memory, perf) or we remove it from everywhere.
But having a mix is a bit confusing: it gives the impress that fields haven't the same role, behaviour, whatever...
We already had this discussion: https://github.com/mozilla/pdf.js/pull/13006 (which didn't land).
About the background-image idea: it should pretty simple (I didn't try it for now):
```css
.xfaTextfield {
   background-image: url('a.png'); /* a.png is a 1x1 image with a transparent blue */
}
.xfaTextfield:focus {
   background-image: none;
}
```
I have no idea of the impact on memory/perf but from a code pov I'd say it isn't that painful to maintain

Snuffleupagus:
> We already had this discussion: #13006 (which didn't land).
> [...]
> I have no idea of the impact on memory/perf but from a code pov I'd say it isn't that painful to maintain

Agreed that a CSS-only solution seems OK, in all aspects. (The previous discussion referenced, if I recall correctly, was centred around using a JS-solution which didn't feel great.)

web/xfa_layer_builder.css
```diff
@@ -130,24 +146,28 @@
...
 .xfaTextfield,
 .xfaSelect {
-  width: 100%;
   height: 100%;
-  flex: 1 1 0;
+  width: 100%;
+  flex: 1 1 auto;
   border: none;
   resize: none;
+  background-image: url("./images/unfocused-field.png");
```
Snuffleupagus, line 163: Can you please use a variable instead, defined at the top, e.g. `--unfocused-field-background: url(images/unfocused-field.png);` similar to how we handle images in https://github.com/mozilla/pdf.js/blob/master/web/viewer.css.

Also, is the test-suite load able to load this new image as-is?

calixteman: > Can you please use a variable instead, defined at the top, e.g. `--unfocused-field-background: url(images/unfocused-field.png);` similar to how we handle images in https://github.com/mozilla/pdf.js/blob/master/web/viewer.css.
> 
> Also, is the test-suite able to load this new image as-is?
Good point: it doesn't. I'll fix that

calixteman:
I replaced png image by an inlined svg one: this way we can easily modify it if we need:
https://github.com/mozilla/pdf.js/pull/13806/commits/e64078cbd465f280f215c407aa28b40dd661bf1d#diff-3047fd328b0593becf1d475dceb17eed893613549d26c537cb9d105a113106afR17

Thanks to that change, the reftest are fine with one exception (at least locally): the one for `xfa_filled_imm1344e` has a bg color issue but only for the first page and only with Firefox.
Maybe it's because it's the first test ran...

marco-c:
> Thanks to that change, the reftest are fine with one exception (at least locally): the one for `xfa_filled_imm1344e` has a bg color issue but only for the first page and only with Firefox.
> Maybe it's because it's the first test ran...

It is happening with firefox-xfa_imm1295e-page1 too, is it only a reftest issue or is it occurring in the actual viewer too?

marco-c:
There are also a few errors.

calixteman:
> > Thanks to that change, the reftest are fine with one exception (at least locally): the one for `xfa_filled_imm1344e` has a bg color issue but only for the first page and only with Firefox.
> > Maybe it's because it's the first test ran...
> 
> It is happening with firefox-xfa_imm1295e-page1 too, is it only a reftest issue or is it occurring in the actual viewer too?

Locally: there are no problem... pffff
And it's ok with chrome.
I dumped the generated svg for all the page 1, copy/paste it in a file and then open it in Firefox: no problem at all.

marco-c:
@calixteman can you check `firefox-hsbc-page1` too? It also has white fields.

marco-c (approved):

calixteman:
> @calixteman can you check `firefox-hsbc-page1` too? It also has white fields.

The chrome snapshot is ok for this one and locally there are no problems.

PR #13808, opened 2021-07-28
Improve caching of shading patterns. (bug 1721949)

brendandahl:
The PDF in bug 1721949 uses many unique pattern objects
that references the same shading many times. This caused
a new canvas pattern to be created and cached many times
driving up memory use.

To fix, I've changed the cache in the worker to key off the
shading object and instead send the shading and matrix
separately. While that worked well to fix the above bug,
there could be PDFs that use many shading that could
cause memory issues, so I've also added a LRU cache
on the main thread for canvas patterns. This should prevent
memory use from getting too high.

### DIFF
--- a/src/display/pattern_helper.js
+++ b/src/display/pattern_helper.js
@@ -46,7 +46,7 @@ class BaseShadingPattern {
 }
 
 class RadialAxialShadingPattern extends BaseShadingPattern {
-  constructor(IR) {
+  constructor(IR, cachedCanvasPatterns) {
     super();
     this._type = IR[1];
     this._bbox = IR[2];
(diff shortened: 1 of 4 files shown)


src/display/pattern_helper.js
```diff
@@ -87,8 +87,8 @@ class RadialAxialShadingPattern extends BaseShadingPattern {
 
   getPattern(ctx, owner, inverse, shadingFill = false) {
     let pattern;
-    if (this._patternCache) {
-      pattern = this._patternCache;
+    if (!shadingFill && this.cachedCanvasPatterns.has(this)) {
+      pattern = this.cachedCanvasPatterns.get(this);
     } else {
       if (!shadingFill) {
         const tmpCanvas = owner.cachedCanvases.getCanvas(
```
Snuffleupagus, line 94: Nit: Unless I'm completely misreading this code, it'd feel slightly easier to reason about the conditions with the following small change:
```suggestion
    if (!shadingFill) {
      if (this.cachedCanvasPatterns.has(this)) {
        pattern = this.cachedCanvasPatterns.get(this);
      } else {
        const tmpCanvas = owner.cachedCanvases.getCanvas(
```

Snuffleupagus (approved):
r=me, with two small comments; thank you!

Snuffleupagus:
Hmm, it looks like `makeref` failed in a previous PR, so the "failures" in https://github.com/mozilla/pdf.js/pull/13808#issuecomment-888514458 should be unrelated.

PR #13813, opened 2021-07-28
Remove a couple of closures in the `src/display/api.js` file

Snuffleupagus:

Also, note that this patch-series still leaves the `PDFWorkerClosure` intact. The reason is partly that the code is less straightforward to re-factor in this way, and more importantly that `PDFWorker` is exposed in the *public API* and I really didn't want to accidentally make any internal state visible to the outside. This could possibly be improved upon if/when we've started using static class fields (currently blocked on ESLint support).


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13815, opened 2021-07-28
XFA - Fix font scale factors (bug 1720888)

calixteman:
- All the scale factors in for the substitution font were wrong because of different glyph positions between Liberation and the other ones:
    - regenerate all the factors
  - Text may have polish chars for example and in this case the glyph widths were wrong:
    - treat substitution font as a composite one
    - add a map glyphIndex to unicode for Liberation in order to generate width array for cid font


calixteman:
I always suspected that something was wrong with glyphs widths for substitution font because of scale factors (the rendering always annoyed me) and this patch will fix that, it means a lot of regression on text rendering but overall it's better.

/botio test

src/core/xfa_fonts.js
```diff
@@ -198,11 +223,60 @@ function getXfaFontWidths(name) {
...
   }
-  return baseWidths.map((w, i) => w * factors[i]);
+
+  let currentCode = -2;
+  let currentArray;
+
+  const newWidths = [];
+  for (const [code, i] of baseMapping
+    .map((c, j) => [c, j])
+    .sort(([c1], [c2]) => c1 - c2)) {
```
Snuffleupagus, line 240: This isn't exactly a wonder of readability (at least to me), any chance that it can be re-factored and/or simplified?

marco-c (approved):
I can only see improvements or minor changes in the reftests.

PR #13816, opened 2021-07-29
[api-minor] Remove the separate `enableScripting` option in `BaseViewer`

Snuffleupagus:
Prior to PR #13042, when scripting wasn't really possible to use outside of the full viewer, the `enableScripting` option made sense.
However, at this point in time having to both pass in a `PDFScriptingManager`-instance *and* set the `enableScripting`-boolean when creating a `BaseViewer`-instance feels redundant and (mostly) annoying. Hence this patch, which removes the *separate* boolean and always enables scripting when `scriptingManager` is provided.

The relevant "viewer component" examples are also updated (with a comment), but in such a way that scripting support won't just break when used with the current PDF.js releases.


timvandermeij (approved):

timvandermeij:
This does seem simpler to me too; thanks!

PR #13821, opened 2021-07-29
XFA - Try to figure out what's wrong with xfa reftests

calixteman:


calixteman:
Considering `firefox-xfa_issue13500-page1`: we really have a new intermittent here...

/botio test

Snuffleupagus:
I looks like maybe the bots haven't been updated since https://github.com/mozilla/botio-files-pdfjs/pull/33 landed, which would probably explain why the `botio browsertest` command doesn't work.

Snuffleupagus:
@calixteman Do we still need to keep this PR open?

PR #13822, opened 2021-07-30
Prevent "Uncaught promise" messages in the console when cancelling (some) `ReadableStream`s

Snuffleupagus:
While fixing issue #13794, I noticed that cancelling the `ReadableStream` returned by the `PDFPageProxy.streamTextContent`-method could lead to "Uncaught promise" messages in the console.[1]
Generally speaking, we don't really care about errors when *cancelling* a `ReadableStream` and it thus seems reasonable to simply suppress any output in those cases.

---
[1] Although, after that issue was fixed you'd now need to set the API-option `stopAtErrors = true` to actually trigger this.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1802,7 +1802,11 @@ class PDFPageProxy {
         return;
       }
     }
-    intentState.streamReader.cancel(new AbortException(reason?.message));
+    intentState.streamReader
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #13828, opened 2021-07-31
Introduce the GitHub Advanced Security workflow

timvandermeij:
Supersedes #13737.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,14 +1,24 @@
 name: CI
 on: [push, pull_request]
 jobs:
-  build:
+  test:
+    name: Test
     runs-on: ubuntu-latest
+
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
       - name: Use Node.js 14 LTS
         uses: actions/setup-node@v1
         with:
           node-version: 14.x
-      - run: npm install -g gulp-cli
-      - run: npm install
-      - run: npm test
+
+      - name: Install Gulp
+        run: npm install -g gulp-cli
+
+      - name: Install other dependencies
+        run: npm install
+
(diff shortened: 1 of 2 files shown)


timvandermeij:
Trivial since this is a minor fixup on the original PR, the workflows all pass and this forms the basis of more workflows.

Snuffleupagus:
Is the intention now to basically "replace" the LGTM integration, by manually enabling the same rule-set in this new workflow?

The reason for my question is that it feels a bit unfortunate to have *two separate* systems flagging (mostly) the same things, since we'll have deal with alert suppression twice now (which could be annoying and lead to inconsistencies).

---

Also, should we perhaps go through the list in https://github.com/mozilla/pdf.js/security/code-scanning to decide/agree upon how to classify the listed alerts?
I suppose that I could try to put together a list with a *suggested* resolution for every alert, and then we can discuss them?

timvandermeij:
> Is the intention now to basically "replace" the LGTM integration, by manually enabling the same rule-set in this new workflow?

For now the idea is that GitHub Advanced Security takes care of the security checks and LGTM takes care of the code quality checks. While they both use CodeQL, the ruleset they use is distinct, so they each serve their own purpose. It might indeed be a good idea to replace LGTM later on and change GitHub Advanced Security to _also_ enable the code quality suite, but since LGTM was just integrated it seemed best to defer that to a follow-up and to first get some experience with both tools; basically https://github.com/mozilla/pdf.js/pull/13737#issuecomment-882749094 which I agree with.

I'll file a follow-up issue to, now that the tooling is in place, take a look at the output and decide how to proceed. I'd be perfectly fine with replacing LGTM now and indeed classifying the current list that GitHub Advanced Security shows, especially since the latter is much easier to control given that alerts can be dismissed from the interface instead of needing to have comments in-code (which I don't find ideal).

PR #13830, opened 2021-07-31
Improve the code analysis workflow with quality checks

timvandermeij:
This allows us to get the quality checks that LGTM does into GitHub Advanced Security. Since it not only runs security checks anymore, the workflow is also renamed to CodeQL to make this more explicit (and this matches the documentation better).

@Snuffleupagus Now that we run the security and quality checks, in other words all checks that CodeQL provides, we get more results, mostly the same as LGTM but I think even some new ones (for example https://github.com/timvandermeij/pdf.js/security/code-scanning/103?query=ref%3Arefs%2Fheads%2Fcodeql). Please see https://github.com/timvandermeij/pdf.js/security/code-scanning?query=is%3Aopen+branch%3Acodeql+ for the test run. Once this is merged, we can classify the list of security/quality issues and after that disable LGTM.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,4 +1,4 @@
-name: GitHub Advanced Security
+name: CodeQL
 on: [push, pull_request]
 jobs:
   analyze:
@@ -23,7 +23,7 @@ jobs:
         uses: github/codeql-action/init@v1
         with:
           languages: ${{ matrix.language }}
-          queries: security-extended
+          queries: security-and-quality
 
       - name: Autobuild CodeQL
         uses: github/codeql-action/autobuild@v1


Snuffleupagus:
> mostly the same as LGTM but I think even some new ones (for example https://github.com/timvandermeij/pdf.js/security/code-scanning/103?query=ref%3Arefs%2Fheads%2Fcodeql). Please see https://github.com/timvandermeij/pdf.js/security/code-scanning?query=is%3Aopen+branch%3Acodeql+ for the test run.

I'm not able to access either of those links, they just bring up the standard GitHub "Page not found"-page.

Without having seen the output you're referring to, are the new alerts perhaps coming from the fact that we're explicitly ignoring a couple of queries in the https://github.com/mozilla/pdf.js/blob/master/lgtm.yml file? If so, any chance that we can just disable those in this configuration file as well?

timvandermeij:
I have added you as a collaborator to my fork, so once that invite is accepted those links should work. It's indeed mostly in test files, so we can either choose to ignore them in the configuration file here too, or silence them once by hand in the interface. It might be a bit nicer to exclude it there so all excludes are in a single place instead of in the interface _and_ in the configuration file?

Snuffleupagus:
> I have added you as a collaborator to my fork, so once that invite is accepted those links should hopefully work.

Thank you; I've looked through the results now, so feel free to remove my access again :-)

> It might be a bit nicer to exclude it there so all excludes are in a single place instead of in the interface _and_ in the configuration file?

I suppose that's a point that I cannot argue with, so let's keep it as-is and we can always change that later on!

Snuffleupagus (approved):
Looks good, thank you!

timvandermeij:
Thank you for checking! I have removed the access again.

PR #13831, opened 2021-07-31
Change the `parseQueryString` function to return a `Map` rather than an Object (issue 13829)

Snuffleupagus:
Even though the code as-is *should* be safe, given that we're using an Object with a `null` prototype, it cannot hurt to change this to a Map to prevent any issues (since we're parsing unknown and potentially unsafe data).

Overall I also think that these changes improve the `parseQueryString` call-sites, since we now have a proper way of checking for the existence of a particular key (and don't have to use `in` which stringifies the keys in the Object).

This patch also changes the default, when no `value` exists, from `null` to an empty string since the use of `decodeURIComponent` currently can modify the value in a somewhat surprising way (at least to me).
Note how `decodeURIComponent(null) === "null"` which is unlikely to be what you actually want, whereas `decodeURIComponent("") === ""` which seems much more helpful.

*Edit:* https://github.com/mozilla/pdf.js/pull/13831/checks?check_run_id=3210760893

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -329,35 +329,38 @@ const PDFViewerApplication = {
     if (!hash) {
       return undefined;
     }
-    const hashParams = parseQueryString(hash),
+    const params = parseQueryString(hash),
       waitOn = [];
 
-    if ("disableworker" in hashParams && hashParams.disableworker === "true") {
+    if (params.get("disableworker") === "true") {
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks better and fixes the alert, so let's do this; thanks!

PR #13836, opened 2021-08-01
Fix inefficient regular expressions in `external/cmapscompress/parse.js`

timvandermeij:
The only purpose, according to the README and existing files, is to parse an integer from those lines, so (\d+) is sufficient for that. This avoids potential exponential backtracking as flagged by CodeQL. I have compared the output of the script with and without these changes and the resulting files are the same.


Snuffleupagus (approved):
Looks good, thank you!

PR #13837, opened 2021-08-01
Implement unit tests for the `parseQueryString` utility function

timvandermeij:
Now that these unit tests are in place, we also take the opportunity to slightly modernize the code itself by using a `for ... of` loop.

(I noticed after the `Map` change that this function did not have unit test coverage yet, so I thought it would be good to add that, especially considering the special `decodeURIComponent(null)` case so that the unit tests assert that we don't accidentally return `"null"` anymore.)

### DIFF
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -185,10 +185,9 @@ function watchScroll(viewAreaElement, callback) {
  * @returns {Map}
  */
 function parseQueryString(query) {
-  const parts = query.split("&");
   const params = new Map();
-  for (let i = 0, ii = parts.length; i < ii; ++i) {
-    const param = parts[i].split("="),
+  for (const part of query.split("&")) {
+    const param = part.split("="),
       key = param[0].toLowerCase(),
       value = param.length > 1 ? param[1] : "";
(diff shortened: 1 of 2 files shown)


web/ui_utils.js
```diff
@@ -187,8 +187,8 @@ function watchScroll(viewAreaElement, callback) {
 function parseQueryString(query) {
   const parts = query.split("&");
   const params = new Map();
-  for (let i = 0, ii = parts.length; i < ii; ++i) {
-    const param = parts[i].split("="),
+  for (const part of parts) {
```
Snuffleupagus, line 190: Nit: You could probably just do `for (const part of query.split("&")) {` now, if you wanted to, and remove the separate variable.

Snuffleupagus (approved):
r=me, with passing tests; thank you adding more unit-tests!

timvandermeij:
The last unit test run failed because of a newly added unit test, `TEST-UNEXPECTED-FAIL | checks that `docId`s are unique and increasing | in chrome | Expected 9 to be less than 0`. I have also seen this locally, so I'll make a follow-up issue for that.

PR #13846, opened 2021-08-02
Add a special `gulp xfatest` command, to limit the ref-tests to only XFA-documents (issue 13744)

### LINKED ISSUE #13744: Make it possible to run xfa reftests exclusively
Currently, the bot will run all refests. It'd be nice to add a command to only run xfa reftests, a command to only run !xfa reftests and keep a command to run all.

Snuffleupagus:
The new command is a *variation* of the standard `gulp test` command and will run all unit/font/integration-tests just as normal, while *only* running ref-tests for XFA-documents to speed up development.
Given that we currently have (some) unit-tests for XFA-documents, and that we may also (in the future) want to add integration-tests, it thus makes sense to run all test-suites in my opinion.

*Please note:* Once this patch has landed, I'll submit a follow-up patch to https://github.com/mozilla/botio-files-pdfjs such that we can also run the new command on the bots.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -551,7 +551,7 @@ function getTempFile(prefix, suffix) {
   return filePath;
 }
 
-function createTestSource(testsName, bot) {
+function createTestSource(testsName, { bot = false, xfaOnly = false } = {}) {
   const source = stream.Readable({ objectMode: true });
   source._read = function () {
     console.log();
@@ -561,9 +561,12 @@ function createTestSource(testsName, bot) {
     const args = ["test.js"];
     switch (testsName) {
       case "browser":
-        args.push("--reftest", "--manifestFile=" + PDF_TEST);
-        break;
-      case "browser (no reftest)":
+        if (!bot) {
+          args.push("--reftest");
+        }
+        if (xfaOnly) {
+          args.push("--xfaOnly");
+        }
         args.push("--manifestFile=" + PDF_TEST);
         break;
       case "unit":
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -1592,9 +1595,34 @@ gulp.task(
...
+  })
+);
+
+gulp.task(
+  "xfatest",
+  gulp.series(setTestEnv, "generic", "components", function runXfaTest() {
+    return streamqueue(
+      { objectMode: true },
+      createTestSource("unit"),
+      createTestSource("browser", { xfaOnly: true }),
```
timvandermeij, line 1612: Why are we not running the font tests here while we do on the bots?

test/test.js
```diff
@@ -125,6 +135,23 @@ function parseOptions() {
...
     })
+    .check(argv => {
+      if (
+        +argv.unitTest + argv.fontTest + argv.integration + argv.xfaOnly <=
+        1
+      ) {
+        return true;
+      }
+      throw new Error(
+        "--unitTest, --fontTest, --integration, and -- xfaOnly must not be specified together."
```
timvandermeij, line 146: I think the space before `xfaOnly` needs to be removed here since `--xfaOnly` is the parameter?

gulpfile.js
Snuffleupagus: Because this is *basically* a copy of the pre-existing `gulp test` command, and we don't run the `fonttest`s there either :-)

The reason for this, as far as I understand, is that the `fonttest`s requires a lot more (and manual) setup than the rest. While the other test-suites can be run after simply executing a couple of `npm` commands, as outlined in e.g. the README, the `fonttest`s require that you *manually* install both Python and ttx, and that's probably not entirely necessary for most contributors.

Since having a permanently failing test-suite locally (for occasional contributors) probably doesn't look good, and given how few `fonttest`s we actually have, that's probably why it's only run on the bots by default.

test/test.js
Snuffleupagus: Whoops, thanks for spotting this!

gulpfile.js
timvandermeij: Good point; I didn't see that the normal test command also didn't run it, but the Python/ttx reason indeed makes sense.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #13847, opened 2021-08-03
Remove useless assignment of `availableSpace` in the `src/core/xfa/template.js` file (issue 13829, 13835)

### LINKED ISSUE #13835: Useless assignment to local variable in the `src/core/xfa/template.js` file
The following code is flagged as dead (since it's *replacing* rather than updating the existing Object):
https://github.com/mozilla/pdf.js/blob/4483cba72410027acf46630ebf252ae529704476/src/core/xfa/template.js#L2336-L2339
according to the alert in https://github.com/mozilla/pdf.js/security/code-scanning/103?query=ref%3Arefs%2Fheads%2Fmaster

Looking at the surrounding code, I can see a couple of possibilities here:
 1. This is indeed just dead code, which should be removed.
 2. The code should be moved up, and placed just *above* the previous block instead.
 3. There should be other, more invasive, changes made to this code...

/cc @calixteman Which solution is correct here?
Assuming it's either option 1 or 2, I can submit a patch to fix this.

Snuffleupagus:
Fixes #13835

### DIFF
--- a/src/core/xfa/template.js
+++ b/src/core/xfa/template.js
@@ -2332,12 +2332,6 @@ class ExclGroup extends XFAObject {
     if (!checkDimensions(this, availableSpace)) {
       return HTMLResult.FAILURE;
     }
-
-    availableSpace = {
-      width: this.w === "" ? availableSpace.width : this.w,
-      height: this.h === "" ? availableSpace.height : this.h,
-    };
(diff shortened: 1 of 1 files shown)


calixteman (approved):
Thanks.

PR #13848, opened 2021-08-03
Remove the LGTM configuration and inline disable comments (issue 13829)

Snuffleupagus:
Given that the GitHub Advanced Security workflow now covers everything that LGTM does, but generally faster and with better GitHub-integration, there's no longer much point in also running LGTM separately.
As a follow-up to this patch, we should also disable/remove the LGTM-integration from the PDF.js repository.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,9 +1,0 @@
-path_classifiers:
-  test:
-    - exclude: test
-    - test/resources
-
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13852, opened 2021-08-03
XFA - Page can be too small when printing (bug 1723705)

calixteman:


marco-c:
Can you add a test?

calixteman:
Unfortunately, I think we don't have such tests for now.
I filed a bug: https://github.com/mozilla/pdf.js/issues/13853

marco-c:
> Unfortunately, I think we don't have such tests for now.

What does setting "print: true" in test_manifest.json do then?

calixteman:
Some data are picked up in the annotation storage in order to be used as values for some text fields, checkboxes, ... (depending on the id).

Snuffleupagus:
> What does setting "print: true" in test_manifest.json do then?

It tests only the *rendering* pipeline of the PDF.js library, by using the correct intent and placing Annotation-data on the canvas itself, but doesn't actually cover *complete* printing process as it happens in the browser.

marco-c (approved):

PR #13856, opened 2021-08-03
XFA - Avoid to put something in very small areas

### LINKED ISSUE #13855: XFA - Some fields appear to be outside of a page
[imm5710f(1).pdf](https://github.com/mozilla/pdf.js/files/6924738/imm5710f.1.pdf)

Configuration:
- Operating system and its version: Debian

1. Open the pdf on page 3 
2. See the 2 fields after `DÉTAILS DU TRAVAIL PRÉVU AU CANADA`

image

calixteman:
- it aims to fix #13855.

### DIFF
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -1246,6 +1246,16 @@
        "enableXfa": true,
        "type": "eq"
     },
+    {  "id": "xfa_issue13855",
+       "file": "pdfs/xfa_issue13855.pdf",
+       "md5": "0aaaf62ff979c7837c2c63df44456238",
+       "link": true,
+       "rounds": 1,
+       "firstPage": 3,
+       "lastPage": 3,
+       "enableXfa": true,
+       "type": "eq"
+    },
     {  "id": "xfa_issue13584",
        "file": "pdfs/xfa_issue13584.pdf",
        "md5": "c41af0754dabd81996a0d54db1900beb",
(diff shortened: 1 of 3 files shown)


marco-c:
Did you mean to `git add` something else?

test/test_manifest.json
```diff
@@ -1246,6 +1246,14 @@
...
        "type": "eq"
     },
+    {  "id": "xfa_issue13855",
+       "file": "pdfs/xfa_issue13855.pdf",
+       "md5": "0aaaf62ff979c7837c2c63df44456238",
+       "link": true,
+       "rounds": 1,
+       "enableXfa": true,
+       "type": "eq"
+    },
```
Snuffleupagus, line 1256: Do you need to test all pages here, you would it suffice to limit it to only the affected page(s)?

(One way of keeping the overall runtime of the tests more manageable, is to not test an entire document unless absolutely necessary.)

calixteman: I had the same thought here and I said to myself that maybe it could useful to have all pages to catch future regressions.
In other words, could we consider that xfa stuff is in a good enough state to only need one page ?

Snuffleupagus: > In other words, could we consider that xfa stuff is in a good enough state to only need one page ?

I hope that the answer, at this point in time, would be *yes*.

calixteman: eh eh..., I really hope too and QA people said us that their xp with xfa was pretty good in general.

marco-c (approved):
The intermittents are becoming a bit painful...

PR #13859, opened 2021-08-03
Remove the `isEOF` helper function and slightly re-factor `EOF`

Snuffleupagus:
Given how trivial the `isEOF` function is, we can simply inline the check at the various call-sites and remove the function (which ought to be ever so slightly more efficient as well).
Furthermore, this patch also changes the `EOF` primitive itself to a `Symbol` instead of an Object since that has the nice benefit of making it unclonable (thus preventing *accidentally* trying to send `EOF` from the worker-thread).

### DIFF
--- a/src/core/cmap.js
+++ b/src/core/cmap.js
@@ -20,7 +20,7 @@ import {
   unreachable,
   warn,
 } from "../shared/util.js";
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:

PR #13867, opened 2021-08-04
[api-minor] Re-factor the *internal* renderingIntent, and change the default `intent` value in the `PDFPageProxy.getAnnotations` method

Snuffleupagus:
With the changes made in PR #13746 the *internal* renderingIntent handling became somewhat "messy", since we're now having to do string-matching in various spots in order to handle the "oplist"-intent correctly.
Hence this patch, which implements the idea from PR #13746 to convert the `intent`-strings, used in various API-methods, into an *internal* renderingIntent that's implemented using a bit-field instead. *Please note:* This part of the patch, in itself, does *not* change the public API (but see below).

This patch is tagged `api-minor` for the following reasons:
 1. It changes the *default* value for the `intent` parameter, in the `PDFPageProxy.getAnnotations` method, to "display" in order to be consistent across the API.
 2. In order to get *all* annotations, with the `PDFPageProxy.getAnnotations` method, you now need to explicitly set "any" as the `intent` parameter.
 3. The `PDFPageProxy.getOperatorList` method will now also support the new "any" intent, to allow accessing the operatorList of all annotations (limited to those types that have one).
 4. Finally, for consistency across the API, the `PDFPageProxy.render` method also support the new "any" intent (although I'm not sure how useful that'll be).

Points 1 and 2 above are the significant, and thus breaking, changes in *default* behaviour here. However, unfortunately I cannot see a good way to improve the overall API while also keeping `PDFPageProxy.getAnnotations` unchanged.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -22,6 +22,7 @@ const RenderingIntentFlag = {
   ANY: 0x01,
   DISPLAY: 0x02,
   PRINT: 0x04,
+  ANNOTATION_FORMS: 0x20,
   OPLIST: 0x100,
 };
 
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -18,6 +18,13 @@ import "./compatibility.js";
 const IDENTITY_MATRIX = [1, 0, 0, 1, 0, 0];
 const FONT_IDENTITY_MATRIX = [0.001, 0, 0, 0.001, 0, 0];
 
+const RenderingIntentFlag = {
+  ANY: 0x01,
+  DISPLAY: 0x02,
+  PRINT: 0x04,
+  OPLIST: 0x100,
+};
+
 // Permission flags from Table 22, Section 7.6.3.2 of the PDF specification.
 const PermissionFlag = {
   PRINT: 0x04,
@@ -1033,6 +1040,7 @@ export {
   PasswordResponses,
   PermissionFlag,
   removeNullCharacters,
+  RenderingIntentFlag,
   setVerbosityLevel,
   shadow,
   StreamType,
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -320,14 +320,7 @@ class Page {
     });
   }
 
-  getOperatorList({
-    handler,
-    sink,
-    task,
-    intent,
-    renderInteractiveForms,
-    annotationStorage,
-  }) {
+  getOperatorList({ handler, sink, task, intent, annotationStorage }) {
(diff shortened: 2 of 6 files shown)


src/shared/util.js
```diff
@@ -18,6 +18,15 @@ import "./compatibility.js";
 const IDENTITY_MATRIX = [1, 0, 0, 1, 0, 0];
 const FONT_IDENTITY_MATRIX = [0.001, 0, 0, 0.001, 0, 0];
 
+const RenderingIntentFlag = {
+  ANY: 0x01,
+  DISPLAY: 0x02,
+  PRINT: 0x04,
+  ANNOTATION_STANDARD: 0x10,
```
timvandermeij, line 25: This one is not actually used.

timvandermeij (commented):
I like the changes proposed here, but now that we're changing default API behavior it might be a good time to tag this as `[api-major]` instead and let the next release be version 3.0. This is not only consistent with semantic versioning, but also allows us to get rid of other code that we previously deprecated (in separate patches) since we're doing a major bump. What do you think?

src/shared/util.js
Snuffleupagus: I've been thinking about possible follow-up work, that the new *internal* renderingIntent format would simplify considerably, but let's indeed leave this value out for now.

Snuffleupagus:
> but now that we're changing default API behavior it might be a good time to tag this as `[api-major]` instead and let the next release be version 3.0.

Well, this wouldn't exactly be the first time that we've made backwards incompatible changes in `api-minor` releases :-)
Hence I'm not sure if *this* particular patch is all that much worse than a lot of other things we've changed over the years.

The recent `es5` -> `legacy` renaming is one recent example of something that should probably (technically speaking) not have been done as a minor change, and generally speaking over the years we've done *a lot* of clean-up/re-factoring of API-methods that in one way or another broke backwards compatibility.

All-in-all, if we're thinking about increasing the *major* version then (in my opinion) it feels like there should be some substantial new feature(s) included and not just incremental changes. E.g. XFA-support might be one such thing, however that's still under very active development and I'd really like to not hold back the changes in this PR (since they *might* facilitate things such as issue 7282, assuming we want to fix that).

timvandermeij (approved):

timvandermeij:
Thank you for doing this! I can't really argue with https://github.com/mozilla/pdf.js/pull/13867#issuecomment-893876981 and the default value is now also what most users are interested in anyway, so let's do this.

Issue #7282, opened 2016-05-03
Option to disable showing of PDF annotations


chitgoks:
is there an option for PDF.js not to show annotations embedded in the pdf?

yurydelendik:
There is no such option. However depending on type of annotation some of them placed into operator list, some of them in the separate request.

If you need a workaround, disable getting of annotations at https://github.com/mozilla/pdf.js/blob/master/src/core/document.js#L320

chitgoks:
thanks yury. hope there'll be an option in the next update. this would come in handy.

yurydelendik:
> this would come in handy.

please explain why -- this will help to prioritize the issue.

chitgoks:
ive created an annotation layer on top of every pdf.js page amd i figure that while ive extracted the annotations in the pdf, saves to darabase and shown them in my annotation layer, by not showing them in the pdf.js layer, the resource consumption may lessen

at least that is what i think

messman:
> please explain why -- this will help to prioritize the issue.

The project I'm working on retrieves annotations separately so that they can be edited on top of the page and saved through a web service. For this reason, we don't want PDF.js to render the annotations because we will be showing them on our own.

chitgoks:
just wish to follow up, will this be considered?

please advise.

timvandermeij:
I think I'm fine with this if it can be made a preference like https://github.com/mozilla/pdf.js/blob/master/web/default_preferences.json#L12.

singhkaushlendra:
Is there any update regarding the above issue ?? I also want to disable rendering of annotations by default.

lpillonel:
Also interested in disabling the annotation. Could be in preferences or maybe as a page render param

bdely:
I would also be interested in an option to disable the annotation layer. the one specifically that i would like to not render would be hyperlinks. because for security reasons we do not what the user to be able to click on the links.
currently i am just setting the annotationLayer class to display: none.

jscher2000:
Possibly a special case, but we had a complaint on SuMo that it is distracting to have an annotation appear on **hover** for a highlight annotation when highlighting has been used liberally in a document. https://support.mozilla.org/questions/1246527

vlastimilmaca:
Maybe show popup only if there is any "Contents" or "RichText" ? That's what Acrobat does.

harryMamani:
> > please explain why -- this will help to prioritize the issue.
> 
[...]

PR #13871, opened 2021-08-04
Add ability to remove annotations

mamiller93:
... not sure what to put here. Currently looking to see if there's a template. Will update as soon as I find out.

Also couldn't run `gulp test` locally as my internet is very slow right now. I need to figure that out as well.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -71,6 +71,58 @@ import { XRef } from "./xref.js";
 const DEFAULT_USER_UNIT = 1.0;
 const LETTER_SIZE_MEDIABOX = [0, 0, 612, 792];
 
+function isAnnotationRenderable(
+  annotation,
+  annotationStorage,
+  annotationIntent
+) {
+  return (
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
While I completely understand that this might make sense in a third-party implementation, I'm really sorry to say that I don't think this is a good idea for the PDF.js repository since this isn't e.g. something that viewer would need.

This PR proposes adding a fair amount of both unused *and* untested code, adding a bunch of complexity to a crucial part of the code-base, which we'd then need to maintain and support going forward.
Also, as-is this could lead to all sorts of breakage if the user includes unclonable data in this new option; which would require validation to be added on the main-thread thus further complicating the code.
Furthermore, as-is this will also *subtly* break caching of operatorLists on the main-thread when this new option is being used, which can lead to broken rendering and a lot of surprises.

All-in-all, I'm again sorry to say that this unfortunately doesn't really seem like the kind of feature we'd want to add. (Note also issue #7282, which looks somewhat related here.)

PR #13874, opened 2021-08-05
Update the Annotation `--zoom-factor` CSS variable when `PDFPageView` is used standalone (PR 13868 follow-up)

Snuffleupagus:
Without this patch, when using `PDFPageView` directly[1] this CSS variable won't be updated and consequently things won't work as intended.
This is purposely implemented such that when a `PDFPageView`-instance is part of a viewer, we don't repeatedly set the CSS variable for every single page.

---
[1] See e.g. the "pageviewer" example in the `examples/components/` folder.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -202,7 +202,6 @@ class BaseViewer {
     } else {
       this.renderingQueue = options.renderingQueue;
     }
-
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13875, opened 2021-08-05
Annotation & XFA: Add focus outlines on different fields (bug 1723615, bug 1718528)

### LINKED ISSUE #12223: PDF Fill - TAB-navigation, not highlighting checkboxes
**Affected versions**
* 81.0a1 (2020-08-16)

**Affected platforms**
* Windows 10, macOS 11.0

**Steps to reproduce**
pdfjs.renderInteractiveForms - enabled

1. Launch Firefox;
2. Access a PDF form with checkboxes;
3. Press the TAB key until reaching and selecting checkboxes;

**Expected result**
* checkboxes are marked as being in focus;

**Actual result**
* no indication of checkboxes being in focus

Additional notes:
* S3 as suggested severity.

calixteman:
- set a default tabindex to be sure they'll be taken into account in the TAB cycle (https://bugzilla.mozilla.org/show_bug.cgi?id=1723615).
  - show default outline when fields are focused (it was an a11y bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1718528).

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -33,6 +33,8 @@ import {
 import { AnnotationStorage } from "./annotation_storage.js";
 import { ColorConverters } from "../shared/scripting_utils.js";
 
+const DEFAULT_TAB_INDEX = 1000;
+
 /**
  * @typedef {Object} AnnotationElementParameters
  * @property {Object} data
@@ -722,6 +724,7 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         element.type = "text";
         element.setAttribute("value", textContent);
       }
+      element.tabIndex = DEFAULT_TAB_INDEX;
 
       elementData.userValue = textContent;
       element.setAttribute("id", id);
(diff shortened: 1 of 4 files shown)


src/display/annotation_layer.js
```diff
@@ -33,6 +33,8 @@ import {
 import { AnnotationStorage } from "./annotation_storage.js";
 import { ColorConverters } from "../shared/scripting_utils.js";
 
+const DEFAULT_TAB_INDEX = 100;
```
Snuffleupagus, line 36: Please pick a slightly larger default value, e.g. at least 200, since this is *very* close to the current highest value in https://github.com/mozilla/pdf.js/blob/master/web/viewer.html and any (future) buttons added to the UI *could* easily result in a clash.

src/display/annotation_layer.js
```diff
@@ -33,6 +33,8 @@ import {
 import { AnnotationStorage } from "./annotation_storage.js";
 import { ColorConverters } from "../shared/scripting_utils.js";
 
+const DEFAULT_TAB_INDEX = 1000;
```
marco-c, line 36: Nit: you could define this in a common file to avoid defining it twice

marco-c (approved):

src/display/annotation_layer.js
calixteman: The value is not the same for annotations and for xfa elements to avoid any conflicts (it should be possible to have some annotations and some xfa elements with XFAF).

calixteman:
so painful...
/botio-linux test

marco-c:
Finally no intermittent errors!

PR #13869, opened 2021-08-04
Enable linting in the `test/resources/` folder, and fix the "Remote property injection" warning (issue 13862)

Snuffleupagus:
*Please refer to the individual commit messages for additional details.*

I'm not going to fix the *main* part of issue #13862, which is about updating and modernizing the overall implementation, however these patches will hopefully help with that work by enabling linting of the code.

Issue #13862, opened 2021-08-03
Fix the security warnings related to the reftest analyzer


timvandermeij:
For the reftest analyzer I think the easiest solution would be to take over parts from the upstream one at https://github.com/mozilla/gecko-dev/blob/master/layout/tools/reftest/reftest-analyzer.xhtml since that one already got refactored to get rid of window usage. We shouldn't take over everything since there is also a large block of testing code in there, and we have made our own modifications to our version in https://github.com/mozilla/pdf.js/commits/master/test/resources/reftest-analyzer.js (both functional and code-wise, for example to not use XHTML and put the JS in a separate file), but it might save some time fixing this.

(Subtask of #13829)

Snuffleupagus:
Also, if we're going to be updating/modernizing that code we should probably *start* with enabling ESLint/Prettier for the code since otherwise it's just too easy to introduce bugs during the re-factoring.

timvandermeij:
The security warnings are fixed by the PRs above. This issue is mainly about cherry-picking improvements from upstream and removing dead code.

PR #13879, opened 2021-08-06
Fix the global variable definitions in `test/resources/reftest-analyzer.js` (issue 13862)

Snuffleupagus:
It shouldn't be necessary to assign these variables to the global scope (as far as I can tell), either explicitly with `window` or implicitly with `var`, and this way we don't need to disable the ESLint `no-undef` rule; fixes another small part of issue #13862.

*Please note:* I wasn't going to put additional work into this code after PR #13869, however these changes looked so simple that I figured trying to get rid of the few remaining "Code scanning alerts" wouldn't hurt.
However, this file would still very much benefit from additional clean-up and re-factoring work, since it's quite old and currently contains some dead code (commented out).

### DIFF
--- a/test/resources/reftest-analyzer.js
+++ b/test/resources/reftest-analyzer.js
@@ -28,21 +28,20 @@ the terms of any one of the MPL, the GPL or the LGPL.
 
 Original author: L. David Baron <dbaron@dbaron.org>
 */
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13881, opened 2021-08-06
XFA - Elements under an area must be bound (bug 1723734)

calixteman:
- aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1723734.


calixteman:
Some diffs are expected (some 0 will appear in some text fields), I compared with Acrobat output for the romanian form and it's ok.
For the other one it looks good but the form is dynamic.

/botio test

PR #13883, opened 2021-08-08
Improve caching of Annotations-data, by using a `Map`, in the API

Snuffleupagus:
Rather than caching only the *last* `PDFPageProxy.getAnnotations` call, and having to handle the intent separately, we can instead implement the caching in exactly the same way as done in the `PDFPageProxy.{render, getOperatorList}` methods.


timvandermeij (approved):

timvandermeij:
Thanks; this makes a lot more sense.

PR #13890, opened 2021-08-11
Update some deprecated ESLint rules

Snuffleupagus:
Please see https://eslint.org/docs/rules/#deprecated where the following rules apply to the PDF.js project:
 - [`no-buffer-constructor`](https://eslint.org/docs/rules/no-buffer-constructor), which we can replace with the `unicorn/no-new-buffer` rule; see https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-new-buffer.md
 - [`no-catch-shadow`](https://eslint.org/docs/rules/no-catch-shadow), which was replaced by the `no-shadow` rule (that we're already using).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -46,6 +46,7 @@
     }],
     "unicorn/no-abusive-eslint-disable": "error",
     "unicorn/no-array-push-push": "error",
+    "unicorn/no-new-buffer": "error",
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for updating this!

PR #13892, opened 2021-08-12
Move some validation, in `Dict.merge`, used during merging of sub-dictionaries (PR 13775 follow-up)

Snuffleupagus:
By not adding any additional non-`Dict` entries to the list of candidates for merging of sub-dictionaries, we can very slightly reduce the amount of parsing required by not having to *again* iterate through unmergeable data.


timvandermeij (approved):

timvandermeij:
Thanks!

Let's make new images after the Firefox update:
/botio makeref

PR #13899, opened 2021-08-15
[Regression] Re-factor the *internal* `includeAnnotationStorage` handling, since it's currently subtly wrong

Snuffleupagus:
*This patch is very similar to the recently fixed `renderInteractiveForms`-options, see PR #13867.*
As far as I can tell, this *subtle* bug has existed ever since `AnnotationStorage`-support was first added in PR #12106 (a little over a year ago).

The value of the `includeAnnotationStorage`-option, as passed to the `PDFPageProxy.render` method, will (potentially) affect the size/content of the operatorList that's returned from the worker (for documents with forms).
Given that operatorLists will generally, unless they contain huge images, be cached in the API, repeated `PDFPageProxy.render` calls where the form-data has been changed by the user in between, can thus *wrongly* return a cached operatorList.

In the viewer we're only using the `includeAnnotationStorage`-option when printing, which is probably why this has gone unnoticed for so long. Note that we, for performance reasons, don't cache printing-operatorLists in the API.
However, there's nothing stopping an API-user from using the `includeAnnotationStorage`-option during "normal" rendering, which could thus result in *subtle* (and difficult to understand) rendering bugs.

In order to handle this, we need to know if the `AnnotationStorage`-instance has been updated since the last `PDFPageProxy.render` call. The most "correct" solution would obviously be to create a hash of the `AnnotationStorage` contents, however that would require adding a bunch of code, complexity, and runtime overhead.
Given that operatorList caching in the API doesn't have to be perfect[1], but only have to avoid *false* cache-hits, we can simplify things significantly be only keeping track of the last time that the `AnnotationStorage`-data was modified.

*Please note:* While working on this patch, I also noticed that the `renderInteractiveForms`- and `includeAnnotationStorage`-options in the `PDFPageProxy.render` method are mutually exclusive.[2]
Given that the various Annotation-related options in `PDFPageProxy.render` have been added at different times, this has unfortunately led to the current "messy" situation.[3]

---
[1] Note how we're already not caching operatorLists for pages with *huge* images, in order to save memory, hence there's no guarantee that operatorLists will always be cached.

[2] Setting both to `true` will result in undefined behaviour, since trying to insert `AnnotationStorage`-values into fields that are being excluded from the operatorList-building will obviously not work, which isn't at all clear from the documentation.

[3] My intention is to try and fix this in a follow-up PR, and I've got a WIP patch locally, however it will result in a number of API-observable changes.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1279,15 +1279,15 @@ class PDFPageProxy {
    *   {Array} of the annotation objects.
    */
   getAnnotations({ intent = "display" } = {}) {
-    const renderingIntent = this._transport.getRenderingIntent(intent, {});
+    const intentArgs = this._transport.getRenderingIntent(intent, {});
 
-    let promise = this._annotationPromises.get(renderingIntent);
+    let promise = this._annotationPromises.get(intentArgs.cacheKey);
     if (!promise) {
       promise = this._transport.getAnnotations(
         this._pageIndex,
-        renderingIntent
+        intentArgs.renderingIntent
       );
-      this._annotationPromises.set(renderingIntent, promise);
+      this._annotationPromises.set(intentArgs.cacheKey, promise);
     }
     return promise;
   }
@@ -1335,8 +1335,9 @@ class PDFPageProxy {
       this._stats.time("Overall");
     }
 
-    const renderingIntent = this._transport.getRenderingIntent(intent, {
+    const intentArgs = this._transport.getRenderingIntent(intent, {
       renderForms: renderInteractiveForms === true,
+      includeAnnotationStorage: includeAnnotationStorage === true,
     });
     // If there was a pending destroy, cancel it so no cleanup happens during
     // this call to render.
@@ -1346,10 +1347,10 @@ class PDFPageProxy {
       optionalContentConfigPromise = this._transport.getOptionalContentConfig();
     }
 
-    let intentState = this._intentStates.get(renderingIntent);
+    let intentState = this._intentStates.get(intentArgs.cacheKey);
     if (!intentState) {
       intentState = Object.create(null);
-      this._intentStates.set(renderingIntent, intentState);
+      this._intentStates.set(intentArgs.cacheKey, intentState);
     }
 
     // Ensure that a pending `streamReader` cancel timeout is always aborted.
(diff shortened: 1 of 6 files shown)


src/display/api.js
```diff
@@ -1584,14 +1575,14 @@ class PDFPageProxy {
...
     const waitOn = [];
-    for (const [intent, intentState] of this._intentStates) {
+    for (const intentState of this._intentStates.values()) {
       this._abortOperatorList({
         intentState,
         reason: new Error("Page was destroyed."),
         force: true,
       });
 
-      if (intent & RenderingIntentFlag.OPLIST) {
```
timvandermeij, line 1615: Unless I'm overlooking something, this was the only (and therefore last) place where the `OPLIST` rendering intent bit flag is actually read. If we make this change, can't we now remove it from the enumeration and the `getRenderingIntent` function?

src/shared/util.js
```diff
@@ -22,7 +22,8 @@ const RenderingIntentFlag = {
   ANY: 0x01,
   DISPLAY: 0x02,
   PRINT: 0x04,
-  ANNOTATION_FORMS: 0x20,
+  ANNOTATIONS_FORMS: 0x10,
```
timvandermeij, line 25: Since we have quite a few rendering intents now, it might be a good time to add a small comment to this enumeration stating what the various rendering intents do. For example, the display/print ones are fairly obvious, but the oplist one less so, and the distinction between forms and storage is also perhaps not immediately clear just from their names.

timvandermeij (commented):
Looks good to me with these questions addressed.

src/display/api.js
Snuffleupagus: Nope, it's being used in https://github.com/mozilla/pdf.js/blob/a936509b77ae4f7e49dea57cf2e2a0aacd062516/src/core/operator_list.js#L611 which is why we (unfortunately) still need it :-)

src/shared/util.js
Snuffleupagus: Yeah, I suppose that on its own this isn't terribly helpful.
My question here is if referencing the `WorkerTransport.getRenderingIntent`-method helps explain the idea behind these flags (better than comment would), since I'm a bit worried that trying to explain the flags on their own will be difficult to follow without context (I often struggle to explain things in a clear and concise way).

There's a possible follow-up to this, see a WIP patch at https://github.com/Snuffleupagus/pdf.js/compare/includeAnnotationStorage-fix-caching...Snuffleupagus:AnnotationMode, which may also clarify things further once the new `annotationMode` option is properly documented.

Snuffleupagus: I've tried to add some docs, and I really hope that it makes sense!
To simplify things, as mentioned before, I'm referring to methods/options in the API since looking at the bigger picture will hopefully be clearer.

timvandermeij (approved):

timvandermeij:
Nice work!

PR #13900, opened 2021-08-17
[api-minor] Stop translating logical assignment in non-`legacy` builds (PR 12887 follow-up)

Snuffleupagus:
When we started using logical assignment operators in the PDF.js project, the feature was new enough that browser support was somewhat limited. That should no longer be the case, since logical assignment has now been available for approximately one year.
Hence this patch, which (basically) reverts PR #12887, since using the Babel plugin unconditionally was never intended to be a permanent thing. For browsers/environments without native logical assignment support, users will now have to use a `legacy` build instead.

Please refer to the browser compatibility information on MDN:
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND_assignment#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment#browser_compatibility

Note also the release information for the major browsers:
 - https://wiki.mozilla.org/Release_Management/Calendar
 - https://en.wikipedia.org/wiki/Google_Chrome_version_history
 - https://en.wikipedia.org/wiki/Safari_version_history#Safari_14

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -208,9 +208,6 @@ function createWebpackConfig(
   }
   const babelExcludeRegExp = new RegExp(`(${babelExcludes.join("|")})`);
 
-  // Since logical assignment operators is a fairly new ECMAScript feature,
-  // for now we translate these regardless of the `SKIP_BABEL` value (with the
-  // exception of `MOZCENTRAL`/`TESTING`-builds where this isn't an issue).
   const babelPlugins = [
     "@babel/plugin-transform-modules-commonjs",
     [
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me; thanks!

PR #13904, opened 2021-08-18
Re-factor the LocalTilingPatternCache to cache by Ref rather than Name (PR 12458 follow-up, issue 13780)

### LINKED ISSUE #13780: Tiling pattern cache may incorrectly cache patterns with same name
It's possible for a PDF page to have multiple resource dictionaries which could have patterns (shading or tiling) that have the same names. Our cache currently works off of name, so we could potentially be caching one pattern and retrieving the wrong one if the name is the same, but the resource dictionary has changed.

I'm not aware of any PDFs affected by this issue, but issue6737.pdf has an example of this with radial shadings. There are two patterns `P1` and `P2` which each have their own resource dictionaries. Each of those resource dictionaries has a pattern named `Sh0`, but they are different shadings.

Snuffleupagus:
This way there cannot be any *incorrect* cache hits, since Refs are guaranteed to be unique.
Please note that the reason for caching by Ref rather than doing something along the lines of the `localShadingPatternCache` (which uses a `Map` directly), is that TilingPatterns are streams and those cannot be cached on the `XRef`-instance (this way we avoid unnecessary parsing).

Fixes #13780

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -795,7 +795,6 @@ class PartialEvaluator {
     patternDict,
     operatorList,
     task,
-    cacheKey,
     localTilingPatternCache
   ) {
     // Create an IR of the pattern code.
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks for improving this!

PR #13905, opened 2021-08-18
Re-factor `loadAndEnablePDFBug` and `PDFBug.init`

Snuffleupagus:
The `loadAndEnablePDFBug` helper function, in `web/app.js`, can be simplified a little bit by making it `async`. Furthermore, given how `PDFBug` is being used, we can also (slightly) re-factor `PDFBug.init` such that the `PDFBug.enable`-call is done internally rather than having to handle that manually at the call-site.

(Finally, utilize `await` more in the `loadFakeWorker` helper function.)

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -377,7 +377,7 @@ const PDFViewerApplication = {
       AppOptions.set("fontExtraProperties", true);
 
       const enabled = params.get("pdfbug").split(",");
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13908, opened 2021-08-19
[api-minor] XFA - Support text search in XFA documents.

### LINKED ISSUE #13878: XFA - Searching a text is broken
Since the rendering is done thanks to HTML we should likely use the "normal" firefox search feature (at least for the builtin viewer)

brendandahl:
Moves the logic out of TextLayerBuilder to handle
highlighting matches into a new separate class `TextHighlighter`
that can be used with regular PDFs and XFA PDFs.

To mimic the current find functionality in XFA, two arrays
from the XFA rendering are created to get the text content
and map those to DOM nodes.

Fixes #13878

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -62,6 +62,7 @@ import { MessageHandler } from "../shared/message_handler.js";
 import { Metadata } from "./metadata.js";
 import { OptionalContentConfig } from "./optional_content_config.js";
 import { PDFDataTransportStream } from "./transport_stream.js";
+import { XfaText } from "./xfa_text.js";
 
 const DEFAULT_RANGE_CHUNK_SIZE = 65536; // 2^16 = 65536
 const RENDERING_CANCELLED_TIMEOUT = 100; // ms
@@ -1561,6 +1562,13 @@ class PDFPageProxy {
    *   {@link TextContent} object that represents the page's text content.
    */
   getTextContent(params = {}) {
+    if (this._transport._htmlForXfa) {
+      // TODO: We need to revisit this once the XFA foreground patch lands and
+      // only do this for non-foreground XFA.
+      return this.getXfa().then(xfa => {
+        return XfaText.textContent(xfa);
+      });
+    }
     const readableStream = this.streamTextContent(params);
 
     return new Promise(function (resolve, reject) {
--- a/src/display/xfa_text.js
+++ b/src/display/xfa_text.js
@@ -1,0 +1,79 @@
+/* Copyright 2021 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class XfaText {
+  /**
+   * Walk an XFA tree and create an array of text nodes that is compatible
+   * with a regular PDFs TextContent. Currently, only TextItem.str is supported,
+   * all other fields and styles haven't been implemented.
+   *
+   * @param {Object} xfa - An XFA fake DOM object.
+   *
+   * @returns {TextContent}
+   */
+  static textContent(xfa) {
+    const items = [];
+    const output = {
+      items,
+      styles: Object.create(null),
+    };
+    function walk(node) {
+      if (!node) {
+        return;
+      }
+      let str = null;
+      const name = node.name;
+      if (name === "#text") {
+        str = node.value;
+      } else if (!XfaText.shouldBuildText(name)) {
+        return;
+      } else if (node?.attributes?.textContent) {
+        str = node.attributes.textContent;
+      } else if (node.value) {
+        str = node.value;
+      }
+      if (str !== null) {
+        items.push({
+          str,
+        });
+      }
+      if (!node.children) {
+        return;
+      }
+      for (const child of node.children) {
+        walk(child);
+      }
+    }
+    walk(xfa);
+    return output;
+  }
+
+  /**
+   * @param {string} name - DOM node name. (lower case)
+   *
+   * @returns {boolean} true if the DOM node should have a corresponding text
+   * node.
+   */
+  static shouldBuildText(name) {
+    return !(
+      name === "textarea" ||
+      name === "input" ||
+      name === "option" ||
+      name === "select"
+    );
+  }
+}
+
+export { XfaText };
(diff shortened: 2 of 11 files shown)


web/xfa_layer_builder.css
```diff
@@ -17,6 +17,37 @@
   --unfocused-field-background: url("data:image/svg+xml;charset=UTF-8,<svg width='1px' height='1px' xmlns='http://www.w3.org/2000/svg'><rect width='100%' height='100%' style='fill:rgba(0, 54, 255, 0.13);'/></svg>");
 }
 
+.xfaLayer .highlight {
```
brendandahl, line 20: This attempts to mimic the current highlighting in text_layer_builder.css. Alternatively, I was think we could create a  .textHighlighter class and share the code, but it's currently small enough and different enough I'm not sure it's worth changing.

brendandahl:
Looks like the bots are slow enough that we catch them while still typing. There are 51 matches of "ci", so at least it's correctly wrong. Now how to wait until find is done...

web/text_highlighter.js
```diff
@@ -0,0 +1,287 @@
...
+ * @property {number} pageIndex - The page index.
+ */
+
+/**
+ * TextHighlighter handles highlighting matches from the FindController in
+ * either the text layer or XFA layer depending on the type of document.
+ */
+class TextHighlighter {
+  constructor({ findController, eventBus, pageIndex }) {
+    this.findController = findController;
```
Snuffleupagus, line 29: Can we, for slightly better readability, initialize all of the provided options here instead (with the rest of the properties afterwards)?
```javascript
this.findController = findController;
this.eventBus = eventBus;
this.pageIdx = pageIndex;
```

web/xfa_layer_builder.css
Snuffleupagus: That seems fine for now, and we can always change it in the future (if needed).

web/xfa_layer_builder.css
```diff
@@ -17,6 +17,37 @@
...
+  margin: -1px;
+  padding: 1px;
+  background-color: #efcbed;
```
Snuffleupagus, line 23: In our existing CSS files, we've tried to settle of *one* colour format; please see https://github.com/mozilla/pdf.js/pull/11447.

Hence can you please use the rgba-format here, and below, instead?

web/pdf_page_view.js
```diff
@@ -187,6 +195,16 @@ class PDFPageView {
...
   }
 
+  async _buildXfaTextContentItems(textDivs) {
+    const text = await this.pdfPage.getTextContent();
+    const items = [];
+    for (const item of text.items) {
+      items.push(item.str);
+    }
+    this.textHighlighter.setTextMapping(textDivs, items);
+    this.textHighlighter.enable();
```
Snuffleupagus, line 205: I'm not finding an corresponding `disable` call, for the xfaLayer-case, anywhere!?

Please note that the xfaLayer, similar to the textLayer, will now be removed from the DOM when the page becomes inactive (to reduce memory usage and improve performance).
Hence it seems to me that you'd want to, in the `cancelRendering`-method, ensure that you also *disable* the textHighlighter for the xfaLayer-case to properly clean-up here.

web/text_highlighter.js
```diff
@@ -0,0 +1,287 @@
...
+  setTextMapping(divs, texts) {
+    this.textDivs = divs;
+    this.textContentItemsStr = texts;
+  }
+
+  /**
+   * Start listening for events to update the highlighter and check if there are
+   * any current matches that need be highlighted.
+   */
+  enable() {
```
Snuffleupagus, line 57: Should we throw an Error if `this.enabled` here, or at the very least print a warning and return early, since multiple `enable`-calls (without `disable` between) seem like a bug elsewhere in the code-base?

web/text_highlighter.js
```diff
@@ -0,0 +1,287 @@
...
+      };
+      this.eventBus._on(
+        "updatetextlayermatches",
+        this._onUpdateTextLayerMatches
+      );
+    }
+    this._updateMatches();
+  }
+
+  disable() {
```
Snuffleupagus, line 76: Similar to the comment above, for the `enable`-method, but for the `!this.enabled` case instead.

Snuffleupagus (changes requested):

src/display/api.js
```diff
@@ -1561,6 +1562,13 @@ class PDFPageProxy {
    *   {@link TextContent} object that represents the page's text content.
    */
   getTextContent(params = {}) {
+    if (this._transport._htmlForXfa /* &&  xfa enabled */) {
```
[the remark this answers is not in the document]
brendandahl: It was a reminder to look in to foreground, forgot to remove when I added the comment below.

src/display/xfa_text.js
```diff
@@ -0,0 +1,74 @@
...
+    };
+    function walk(node) {
+      if (!node) {
+        return;
+      }
+      let str = null;
+      const name = node.name;
+      if (name === "#text") {
+        str = node.value;
+      } else if (!XfaText.shouldBuildText(name)) {
```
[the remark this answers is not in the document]
brendandahl: That was failing with lint, but seems to be fine now that I upgraded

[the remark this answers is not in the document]
brendandahl: I take that back, it fails because `this` doesn't exist because we're in the closure.

web/app_options.js
```diff
@@ -312,6 +314,11 @@ if (
     value: "../build/pdf.sandbox.js",
     kind: OptionKind.VIEWER,
   };
+  defaultOptions.enableXfa = {
+    /** @type {boolean} */
+    value: false,
+    kind: OptionKind.API + OptionKind.PREFERENCE,
+  };
```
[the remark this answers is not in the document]
brendandahl: The chrome lint was failing because it finds enableXfa = true. Is there a better way to fix this?

src/display/xfa_text.js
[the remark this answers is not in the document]
Snuffleupagus: Sorry, didn't notice the closure; please disregard my comment :-)

web/app_options.js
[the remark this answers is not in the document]
Snuffleupagus: Sorry, this is my fault since think about all the ways that the `TESTING` define could be used. Please see PR #13911 which should fix this.

web/text_highlighter.js
brendandahl: We currently call disable without enabling since we cancel things out before starting. Since this is safe to call without enabling, I think we should just leave it as is.

Snuffleupagus: Maybe just add a `if (!this.enabled) { return; }` at the top then, since none of the code in this method needs to run if it's already disabled?

Snuffleupagus (approved):
Looks good to me, with one final *optional* comment; thank you!

---

This is kind of how I imagined/hoped that the solution to issue #13908 would look like, with re-using the existing find functionality.
It's especially nice, in my opinion, that you extracted the `TextHighlighter` code into its own class and were able to re-use it for both "normal" and XFA find :-)

PR #13913, opened 2021-08-20
Fix pdf_viewer definitions

michael-yx-wu:
Follow up to https://github.com/mozilla/pdf.js/pull/13588

Current pdf_viewer definitions result in errors like the following when trying to use them in a ts project:

```
[error] TypeScript error node_modules/.pnpm/pdfjs-dist@2.10.377/node_modules/pdfjs-dist/web/pdf_viewer.d.ts:1:15 - error TS2691: An import path cannot end with a '.d.ts' extension. Consider importing 'pdfjs-dist/types/web/pdf_viewer.component.js' instead.

1 export * from "pdfjs-dist/types/web/pdf_viewer.component.d.ts";
```

Import/export statements in typescript should not include file extensions.

### DIFF
--- a/external/dist/legacy/web/pdf_viewer.d.ts
+++ b/external/dist/legacy/web/pdf_viewer.d.ts
@@ -1,1 +1,1 @@
-export * from "pdfjs-dist/types/web/pdf_viewer.component.d.ts";
+export * from "pdfjs-dist/types/web/pdf_viewer.component";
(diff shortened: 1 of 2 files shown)


Snuffleupagus (commented):
Please include the information from https://github.com/mozilla/pdf.js/pull/13913#issue-716412282 in the *commit message* as well.

michael-yx-wu:
Thanks for the review! I've amended the commit message.

Snuffleupagus (approved):
While I don't really know TypeScript, the changes look reasonable enough; thank you for the patch!

PR #13922, opened 2021-08-22
Enable the ESLint `object-shorthand` rule in the `extensions/chromium/`-folder

Snuffleupagus:
Based on the following compatibility information, there can't be any compelling reason to not enable this ESLint rule now: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#browser_compatibility

See also https://eslint.org/docs/rules/object-shorthand

### DIFF
--- a/extensions/chromium/.eslintrc
+++ b/extensions/chromium/.eslintrc
@@ -18,6 +18,5 @@
   "rules": {
     "mozilla/import-globals": "error",
     "no-var": "off",
-    "object-shorthand": "off",
   },
 }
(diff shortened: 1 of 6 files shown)


Snuffleupagus:
gulp lint

timvandermeij (approved):

timvandermeij:
More standardization, nice!

Snuffleupagus:
> More standardization, nice!

Thank you!
I'm guessing that we could probably enable `no-var` as well, but that would likely require a bunch of manual changes as well (hence not something that I'll prioritize working on myself):

 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#browser_compatibility

PR #13923, opened 2021-08-22
[api-minor] Introduce a new `annotationMode`-option, in `PDFPageProxy.{render, getOperatorList}`

### LINKED ISSUE #7282: Option to disable showing of PDF annotations

Snuffleupagus:
*This is a follow-up to PRs #13867 and #13899.*

This patch is tagged `api-minor` for the following reasons:
 - It replaces the `renderInteractiveForms`/`includeAnnotationStorage`-options, in the `PDFPageProxy.render`-method, with the single `annotationMode`-option that controls which annotations are being rendered and how. Note that the old options were mutually exclusive, and setting both to `true` would result in undefined behaviour.

 - For improved consistency in the API, the `annotationMode`-option will also work together with the `PDFPageProxy.getOperatorList`-method.

 - It's now also possible to disable *all* annotation rendering in both the API and the Viewer, since the other changes meant that this could now be supported with a single added line on the worker-thread[1]; fixes #7282.

---
[1] Please note that in order to simplify the overall implementation, we'll purposely only support disabling of *all* annotations and that the option is being shared between the API and the Viewer. For any more "specialized" use-cases, where e.g. only some annotation-types are being rendered and/or the API and Viewer render different sets of annotations, that'll have to be handled in third-party implementations/forks of the PDF.js code-base.

### DIFF
--- a/extensions/chromium/preferences_schema.json
+++ b/extensions/chromium/preferences_schema.json
@@ -144,9 +144,15 @@
       "description": "Whether to prevent the extension from reporting the extension and browser version to the extension developers.",
       "default": false
     },
-    "renderInteractiveForms": {
-      "type": "boolean",
-      "default": true
+    "annotationMode": {
+      "type": "integer",
+      "enum": [
+        0,
+        1,
+        2,
+        3
+      ],
+      "default": 2
     },
     "enableScripting": {
       "type": "boolean",
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -380,7 +380,10 @@ class Page {
     // page's operator list to render them.
     return Promise.all([pageListPromise, this._parsedAnnotations]).then(
       function ([pageOpList, annotations]) {
-        if (annotations.length === 0) {
+        if (
+          annotations.length === 0 ||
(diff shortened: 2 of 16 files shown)


brendandahl:
I think this is still considered a major version change since it removes/changes arguments to the API. A user could not take this version if they had used some different values and seamlessly upgrade. We'll need to either 1) do a major release or 2) convert this to a minor release and leave the old options and mark them as deprecated and have them populate the new arguments.

Snuffleupagus:
> I think this is still considered a major version change since it removes/changes arguments to the API.

Well, this wouldn't exactly be the first time that we've made *strictly speaking* backwards incompatible changes in `api-minor` patches :-)

> convert this to a minor release and leave the old options and mark them as deprecated and have them populate the new arguments.

The issue is then exactly *how* to do that conversion consistently and correctly, since a pretty big issue in the current implementation is that as mentioned above the two old options are unfortunately mutually exclusive!?

brendandahl (approved):

pardoman:
These changes look promising!

Question: How could I use the library to render only the annotations layer?

Thanks

PR #13930, opened 2021-08-23
Fix Viewer API definitions and include in CI

michael-yx-wu:
The Viewer API definitions do not compile because of missing imports and
anonymous objects are typed as `Object`. These issues were not caught
during CI because the test project was not compiling anything from the
Viewer API.

As an example of the first problem:

```
/**
 * @implements MyInterface
 */
export class MyClass {
    ...
}
```

will generate a broken definition that doesn’t import MyInterface:

```
/**
 * @implements MyInterface
 */
export class MyClass implements MyInterface {
    ...
}
```

This can be fixed by adding a typedef jsdoc to specify the import:

```
/** @typedef {import("./otherFile").MyInterface} MyInterface */
```

See https://github.com/jsdoc/jsdoc/issues/1537 and
https://github.com/microsoft/TypeScript/issues/22160 for more details.

As an example of the second problem:

```
/**
 * Gets the size of the specified page, converted from PDF units to inches.
 * @param {Object} An Object containing the properties: {Array} `view`,
 *   {number} `userUnit`, and {number} `rotate`.
 */
function getPageSizeInches({ view, userUnit, rotate }) {
    ...
}
```

generates the broken definition:

```
function getPageSizeInches({ view, userUnit, rotate }: Object) {
    ...
}
```

The jsdoc should specify the type of each nested property:

```
/**
 * Gets the size of the specified page, converted from PDF units to inches.
 * @param {Object} options An object containing the properties: {Array} `view`,
 *   {number} `userUnit`, and {number} `rotate`.
 * @param {number[]} options.view
 * @param {number} options.userUnit
 * @param {number} options.rotate
 */
```

### DIFF
--- a/test/types/main.ts
+++ b/test/types/main.ts
@@ -1,9 +1,12 @@
 import { getDocument } from "pdfjs-dist";
+import { EventBus } from "pdfjs-dist/web/pdf_viewer.component";
 
 class MainTest {
+  eventBus: EventBus;
   task: ReturnType<typeof getDocument> | undefined;
 
   constructor(public file: string) {
+    this.eventBus = new EventBus({});
   }
 
   loadPdf() {
--- a/web/annotation_layer_builder.js
+++ b/web/annotation_layer_builder.js
@@ -13,6 +13,9 @@
  * limitations under the License.
  */
 
+// eslint-disable-next-line max-len
+/** @typedef {import("./interfaces").IPDFAnnotationLayerFactory} IPDFAnnotationLayerFactory */
+
 import { AnnotationLayer } from "pdfjs-lib";
 import { NullL10n } from "./l10n_utils.js";
 import { SimpleLinkService } from "./pdf_link_service.js";
--- a/web/interfaces.js
+++ b/web/interfaces.js
@@ -106,6 +106,11 @@ class IPDFLinkService {
  * @interface
  */
 class IRenderableView {
+  constructor() {
+    /** @type {function | null} */
+    this.resume = null;
+  }
+
   /**
    * @type {string} - Unique ID for rendering queue.
    */
@@ -120,8 +125,6 @@ class IRenderableView {
    * @returns {Promise} Resolved on draw completion.
    */
   draw() {}
-
-  resume() {}
 }
 
 /**
--- a/web/pdf_page_view.js
+++ b/web/pdf_page_view.js
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+/** @typedef {import("./interfaces").IRenderableView} IRenderableView */
+
 import {
   AnnotationMode,
   createPromiseCapability,
--- a/web/pdf_rendering_queue.js
+++ b/web/pdf_rendering_queue.js
@@ -33,6 +33,7 @@ class PDFRenderingQueue {
     this.pdfThumbnailViewer = null;
     this.onIdle = null;
     this.highestPriorityPage = null;
+    /** @type {number} */
     this.idleTimeout = null;
     this.printing = false;
     this.isThumbnailViewEnabled = false;
--- a/web/text_layer_builder.js
+++ b/web/text_layer_builder.js
@@ -13,6 +13,9 @@
  * limitations under the License.
  */
 
+// eslint-disable-next-line max-len
+/** @typedef {import("./interfaces").IPDFTextLayerFactory} IPDFTextLayerFactory */
+
 import { renderTextLayer } from "pdfjs-lib";
 
 const EXPAND_DIVS_TIMEOUT = 300; // ms
--- a/web/ui_utils.js
+++ b/web/ui_utils.js
@@ -285,11 +285,22 @@ function roundToDivide(x, div) {
 }
 
 /**
+ * @typedef {Object} GetPageSizeInchesParameters
+ * @property {number[]} view
+ * @property {number} userUnit
+ * @property {number} rotate
+ */
+
+/**
+ * @typedef {Object} PageSize
+ * @property {number} width - In inches.
+ * @property {number} height - In inches.
+ */
+
+/**
  * Gets the size of the specified page, converted from PDF units to inches.
- * @param {Object} An Object containing the properties: {Array} `view`,
- *   {number} `userUnit`, and {number} `rotate`.
- * @returns {Object} An Object containing the properties: {number} `width`
- *   and {number} `height`, given in inches.
+ * @param {GetPageSizeInchesParameters} params
+ * @returns {PageSize}
  */
 function getPageSizeInches({ view, userUnit, rotate }) {
   const [x1, y1, x2, y2] = view;
--- a/external/dist/legacy/web/pdf_viewer.d.ts
+++ b/external/dist/legacy/web/pdf_viewer.d.ts
@@ -1,1 +1,1 @@
-export * from "pdfjs-dist/types/web/pdf_viewer.component";
+export * from "../../types/web/pdf_viewer.component";
--- a/external/dist/web/pdf_viewer.d.ts
+++ b/external/dist/web/pdf_viewer.d.ts
@@ -1,1 +1,1 @@
-export * from "pdfjs-dist/types/web/pdf_viewer.component";
+export * from "../types/web/pdf_viewer.component";
--- a/test/types/tsconfig.json
+++ b/test/types/tsconfig.json
@@ -17,7 +17,7 @@
     ],
     "paths": {
       "pdfjs-dist": ["../../build/typestest"],
-      "pdfjs-dist/*": ["../../build/typestest/build/*"]
+      "pdfjs-dist/*": ["../../build/typestest/types/*"]
     }
   },
   "files": [
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -139,7 +139,6 @@ function isSameScale(oldScale, newScale) {
 
 /**
  * Simple viewer control to display PDF content/pages.
- * @implements {IRenderableView}
  */
 class BaseViewer {
   /**
--- a/web/pdf_link_service.js
+++ b/web/pdf_link_service.js
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+/** @typedef {import("./interfaces").IPDFLinkService} IPDFLinkService */
+
 import { parseQueryString } from "./ui_utils.js";
 
 /**
--- a/web/struct_tree_layer_builder.js
+++ b/web/struct_tree_layer_builder.js
@@ -13,6 +13,9 @@
  * limitations under the License.
  */
 
+// eslint-disable-next-line max-len
+/** @typedef {import("./interfaces").IPDFStructTreeLayerFactory} IPDFStructTreeLayerFactory */
+
 const PDF_ROLE_TO_HTML_ROLE = {
   // Document level structure types
   Document: null, // There's a "document" role, but it doesn't make sense here.
--- a/web/text_highlighter.js
+++ b/web/text_highlighter.js
@@ -14,7 +14,7 @@
  */
 
 /**
- * @typedef {Object} TextHighlighter
+ * @typedef {Object} TextHighlighterOptions
  * @property {PDFFindController} findController
  * @property {EventBus} eventBus - The application event bus.
  * @property {number} pageIndex - The page index.
@@ -25,6 +25,9 @@
  * either the text layer or XFA layer depending on the type of document.
  */
 class TextHighlighter {
+  /**
+   * @param {TextHighlighterOptions} options
+   */
   constructor({ findController, eventBus, pageIndex }) {
     this.findController = findController;
     this.matches = [];
--- a/web/xfa_layer_builder.js
+++ b/web/xfa_layer_builder.js
@@ -13,12 +13,15 @@
  * limitations under the License.
  */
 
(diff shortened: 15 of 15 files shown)


web/annotation_layer_builder.js
```diff
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+// eslint-disable-next-line max-len
+/** @typedef {import("./interfaces").IPDFAnnotationLayerFactory} IPDFAnnotationLayerFactory */
```
Snuffleupagus, line 17: Here, and elsewhere, you really ought to put a blank line between this sort of comment and the imports below since as-is this is *somewhat* hurting readability.

web/interfaces.js
```diff
@@ -107,14 +107,20 @@ class IPDFLinkService {
  */
 class IPDFHistory {
   /**
-   * @param {Object} params
+   * @param {Object} options
+   * @param {*} options.fingerprint
```
Snuffleupagus, line 111: Why `*` here, since it should be a string!?

web/interfaces.js
```diff
@@ -107,14 +107,20 @@ class IPDFLinkService {
...
+   * @param {boolean} options.resetHistory
+   * @param {boolean} options.updateUrl
```
Snuffleupagus, line 113: Also, both of these should be marked as optional (since they are).

Please see https://github.com/mozilla/pdf.js/blob/bf5a45ce6d0e9d070204859a39844bcdaf69e9aa/web/pdf_history.js#L36-L42 for the expected types (and what's optional).

web/interfaces.js
```diff
@@ -107,14 +107,20 @@ class IPDFLinkService {
...
    */
   initialize({ fingerprint, resetHistory = false, updateUrl = false }) {}
 
   reset() {}
 
   /**
    * @param {Object} params
+   * @param {*} params.namedDest
+   * @param {*} params.explicitDest
+   * @param {number} params.pageNumber
```
Snuffleupagus, line 123: Again, please see https://github.com/mozilla/pdf.js/blob/bf5a45ce6d0e9d070204859a39844bcdaf69e9aa/web/pdf_history.js#L44-L50 for the expected types.

web/pdf_rendering_queue.js
```diff
@@ -33,6 +33,9 @@ class PDFRenderingQueue {
     this.pdfThumbnailViewer = null;
     this.onIdle = null;
     this.highestPriorityPage = null;
+    /**
+     * @type {number}
+     */
```
Snuffleupagus, line 38: Sorry, but why?

This property really *should* be completely internal.

web/ui_utils.js
```diff
@@ -284,12 +284,20 @@ function roundToDivide(x, div) {
   return r === 0 ? x : Math.round(x - r + div);
 }
 
+/**
+ * @typedef {Object} PageSize
+ * @property {number} width
+ * @property {number} height
+ */
```
Snuffleupagus, line 291: Please add a blank line here.

web/ui_utils.js
```diff
@@ -284,12 +284,20 @@ function roundToDivide(x, div) {
   return r === 0 ? x : Math.round(x - r + div);
 }
 
+/**
+ * @typedef {Object} PageSize
+ * @property {number} width
+ * @property {number} height
```
Snuffleupagus, line 290:
```javascript
  * @property {number} width - In inches.
  * @property {number} height - In inches.
```

web/ui_utils.js
```diff
@@ -284,12 +284,20 @@ function roundToDivide(x, div) {
...
- * @param {Object} An Object containing the properties: {Array} `view`,
+ * @param {Object} options An object containing the properties: {Array} `view`,
  *   {number} `userUnit`, and {number} `rotate`.
- * @returns {Object} An Object containing the properties: {number} `width`
- *   and {number} `height`, given in inches.
+ * @param {number[]} options.view
+ * @param {number} options.userUnit
+ * @param {number} options.rotate
+ * @returns {PageSize} An Object containing the properties: {number} `width`
+ * and {number} `height`, given in inches.
```
Snuffleupagus, line 300: Let's introduce a proper `typedef` instead here, i.e. something like:
```javascript
/**
 * @typedef {Object} getPageSizeInchesParameters
 * @property {Array} view
 * @property {number} userUnit
 * @property {number} rotation
 */

/**
 * Gets the size of the specified page, converted from PDF units to inches.
 * 
 * @param {getPageSizeInchesParameters} params
 * @returns {PageSize}
 */
```

Snuffleupagus (changes requested):
I don't really understand the "add CI check task"-part of the commit message, is there something missing in the patch?
Also, it probably wouldn't hurt to actually mention (in the first line of the commit message) that it's the TypeScript definitions for the *viewer* that's being changed here.

web/interfaces.js
michael-yx-wu: Ah thanks, I wasn't sure about the type so I left it as `*`. Will change!

web/interfaces.js
```diff
@@ -149,7 +155,10 @@ class IRenderableView {
    */
   draw() {}
 
-  resume() {}
+  /**
+   * @type {?(() => void)}
+   */
+  get resume() {}
```
Snuffleupagus, line 161: This also looks somewhat weird, since when `resume` is defined it's a regular function and *not* a getter.

michael-yx-wu: Agree, but in the implementing classes (e.g. `PDFPageView`) , it's actually initialized to be `null`

https://github.com/mozilla/pdf.js/blob/bf5a45ce6d0e9d070204859a39844bcdaf69e9aa/web/pdf_page_view.js#L116

and later set to a function

https://github.com/mozilla/pdf.js/blob/bf5a45ce6d0e9d070204859a39844bcdaf69e9aa/web/pdf_page_view.js#L571

which causes the generated types for `PDFPageView` and this interface to disagree unless I type `resume` like this. I felt like it'd be safer for this PR to have the types reflect the current implementation rather than change the implementation to match the desired types.

Snuffleupagus: > which causes the generated types for `PDFPageView` and this interface to disagree unless I type `resume` like this.

But this actually means the interface-definition is now by definition *incorrect*, which really doesn't help someone trying to see how it should be correctly implemented. Can't we just tag the method as follows instead?
```javascript
/**
 * @ignore
 */
resume() {}
```

web/pdf_rendering_queue.js
michael-yx-wu: because of the `setTimeout` call, `tsc` infers this type to be `idleTimeout: NodeJS.Timeout | null;`, resulting in the following compilation error since the types test project isn't building for node:
```
Couldn't compile TypeScript test: build/typestest/types/web/pdf_rendering_queue.d.ts(9,18): error TS2503: Cannot find namespace 'NodeJS'.
```

https://github.com/mozilla/pdf.js/blob/56e7bb626c89282f207b585ebed74c767c68db95/web/pdf_rendering_queue.js#L95-L97

Writing this has made me realize that pdf.js seems to support running server-side and since `NodeJS.setTimeout` returns an object while` window.setTimeout` returns a number, it would be more proper to type this as `ReturnType<typeof setTimeout>`, where `setTimeout`'s type will ultimately depend on your project setup.

web/interfaces.js
michael-yx-wu: `tsc` seems to be ignoring the `@ignore` annotation and generates types anyway, so I either have to set the type to describe the current situation or fix the underlying implementations to match the desired type of `() => void`. I'll make an attempt at a fix since it looks relatively simple

test/types/main.ts
```diff
@@ -1,9 +1,12 @@
 import { getDocument } from "pdfjs-dist";
+import { EventBus } from "pdfjs-dist/web/pdf_viewer.component";
```
michael-yx-wu, line 2: The contents of the `typestest` folder is somewhat different from `pdfjs-dist`. Notably, it's missing a `legacy` folder so it's not possible to check that the `external/dist/legacy/web/pdf_viewer.d.ts` file is doing the right thing during CI. Could be good to do in a future change!

Snuffleupagus: The TypeScript definitions provided can be seen as a best-effort solution, hence we don't want to *further* increase test run-time for `typestest` by testing e.g. the `legacy`-folder (those builds are only provided for older environments anyway).

web/pdf_page_view.js
```diff
@@ -113,7 +114,7 @@ class PDFPageView {
     this.paintTask = null;
     this.paintedViewportMap = new WeakMap();
     this.renderingState = RenderingStates.INITIAL;
-    this.resume = null;
+    this.resume = () => {};
```
Snuffleupagus, line 117: Sorry, but please revert all changes of this kind throughout the patch!

We really don't want to add a, possibly small but nonetheless, overhead to our JavaScript code just for the TypeScript definitions.
Having the TypeScript definitions dictate how we can/cannot write JavaScript code isn't really OK as far as I'm concerned.

web/pdf_history.js
```diff
@@ -33,26 +35,13 @@ const UPDATE_VIEWAREA_TIMEOUT = 1000; // milliseconds
  * @property {EventBus} eventBus - The application event bus.
  */
 
-/**
```
Snuffleupagus, line 36: Please revert all changes in this file, since it really doesn't seem great having to look in a different file to find the docs.

web/pdf_rendering_queue.js
```diff
@@ -33,6 +33,9 @@ class PDFRenderingQueue {
     this.pdfThumbnailViewer = null;
     this.onIdle = null;
     this.highestPriorityPage = null;
+    /**
+     * @type {ReturnType<typeof setTimeout>}
```
Snuffleupagus, line 37: Note that the code in the `web/`-folder is only really intended to run in browsers, and please make this one-line comment instead; i.e. something like `/** @type ... */`

web/text_layer_builder.js
```diff
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+// eslint-disable-next-line max-len
+/** @typedef {import("./interfaces").IPDFTextLayerFactory} IPDFTextLayerFactory */
```
Snuffleupagus, line 17: Missing a blank line here.

web/ui_utils.js
```diff
@@ -284,12 +284,23 @@ function roundToDivide(x, div) {
...
+/**
+ * @typedef {Object} GetPageSizeInchesParameters
+ * @property {number[]} view
+ * @property {number} userUnit
+ * @property {number} rotate
+ */
+
+/**
+ * @typedef {Object} PageSize
+ * @property {number} width In inches.
```
Snuffleupagus, line 296: `* @property {number} width - In inches.` and similar below.

Snuffleupagus (changes requested):

web/pdf_history.js
michael-yx-wu: I was worried about the duplicate definitions diverging over time. Happy to revert though

web/pdf_page_view.js
michael-yx-wu: Sounds good. But this means I'll have to also revert IRenderableView to the below to get the typings to compile:

```
class IRenderableView {
  ...
  /**
   * @type {?(() => void)}
   */
  get resume() {}
}
``` 

Is that okay?

web/pdf_history.js
michael-yx-wu: I suppose the types CI will catch incompatible definitions

web/pdf_page_view.js
Snuffleupagus: No, as was mentioned previously that's also not correct since this isn't a getter.

michael-yx-wu: What about something like this? 
```
class IRenderableView {
  constructor() {
    /** @type {(() => void) | null} */
    this.resume = null;
  }
  ...
}
```

web/pdf_rendering_queue.js
michael-yx-wu: oh, in that case I will replace the `setTimeout` with `window.setTimeout` to disambiguate that we want the browser typings

web/pdf_rendering_queue.js
```diff
@@ -93,7 +93,10 @@ class PDFRenderingQueue {
     }
 
     if (this.onIdle) {
-      this.idleTimeout = setTimeout(this.onIdle.bind(this), CLEANUP_TIMEOUT);
+      this.idleTimeout = window.setTimeout(
+        this.onIdle.bind(this),
+        CLEANUP_TIMEOUT
+      );
```
Snuffleupagus, line 99: Please revert this change, since it'd be a case of the TypeScript definitions dictating how we can/can't write JavaScript code. Adding a one-liner comment with correct `type`-information actually seem preferable here.

Note also that we've got `setTimeout`-calls in various parts of the code-base and having one special case like this doesn't feel right (and I *really* don't want to convert all of them).

Snuffleupagus (changes requested):

timvandermeij (commented):
Looks good with these comments addressed (and some of the remaining ones above). Thank you for improving the types and their testing!

Snuffleupagus (approved):
This looks much better, thank you!
(I'll let Tim take one last look as well, before we merge this.)

timvandermeij (approved):

PR #13932, opened 2021-08-24
Support Optional Content in Image-/XObjects (issue 13931)

### LINKED ISSUE #13931: Can't hide XObject based layer
[444.pdf](https://github.com/mozilla/pdf.js/files/7035877/444.pdf)

Configuration:
- Web browser and its version:  Chorme 92.0.4515.159
- Operating system and its version: Windows10
- PDF.js version: https://mozilla.github.io/pdf.js/web/viewer.html

The attachment is a PDF with a XObject image in a OCG layer. It works ok in the Adobe PDF Reader and Foxit PDF Reader. 

But with pdfjs, when hide the layer, the image is still visible.

The following text is from the specification("PDF32000_2008.pdf"):

 8.11.3.3   Optional Content in XObjects and Annotations

In addition to marked content within content streams, form XObjects and image XObjects (see 8.8, "External Objects")  and  annotations  (see  12.5,  "Annotations")  may  contain  an  OC  entry,  which  shall  be  an  optional  content group or an optional content membership dictionary

Snuffleupagus:
- Support Optional Content in Image-/XObjects (issue 13931)
 
   Currently, in the `PartialEvaluator`, we only support Optional Content in Form-/XObjects. Hence this patch adds support for Image-/XObjects as well, which looks like a simple oversight in PR #12095 since the canvas-implementation already contains the necessary code to support this.

 - Add support for specifying non-default Optional Content in the ref-tests 

Fixes #13931


brendandahl:
Nice, thanks for the tests too!

PR #13933, opened 2021-08-24
Fix saving of XFA checkboxes. (bug 1726381)

brendandahl:
Previously were were always setting the storage value to the on value.

I'd like to add a test for this, but it's proving difficult to intercept the save file. My plan is to add a mozilla central test since we already have some tests for saving acroforms there.


Snuffleupagus (approved):
This looks reasonable as far as I understand, and in my (manual) testing it seems to work nicely; thank you!

timvandermeij (approved):

PR #13941, opened 2021-08-28
Ensure that `PasswordException` is handled correctly in the `wrapReason` function

Snuffleupagus:
While running the unit-tests with some logging statements added to this code, I noticed that `PasswordException` was missing from the list of potential Errors that could be passed to the `wrapReason` function.


timvandermeij (approved):

timvandermeij:
Nice catch!

PR #13942, opened 2021-08-28
Export the XFA/StructTree-layers in the viewer components

Snuffleupagus:
While e.g. the `simpleviewer` and `singlepageviewer` examples work, since they're based on the `BaseViewer`-class, the standalone `pageviewer` example currently doesn't support either XFA- or StructTree-layers. This seems like an obvious oversight, which can be easily addressed simply by exporting the necessary functionality through `pdf_viewer.component.js`, similar to the existing Text/Annotation-layers.

While working on, and testing, these changes I happened to notice a number of smaller things that's also fixed in this patch:

 - Ensure that `XfaLayerBuilder.render` always have a *consistent* return type, to prevent possible run-time failures in `PDFPageView`; PR #13908 follow-up.

 - Change the order of the options in the `XfaLayerBuilder`-constructor to agree with the parameter order in the `DefaultXfaLayerFactory.createXfaLayerBuilder`-method.

 - Add a missing `textHighlighterFactory`-option, in the JSDocs for the `PDFPageView`-class.

 - A couple of small tweaks in the `TextLayerBuilder.render`-method: Re-use an existing Array rather than creating a new one, and replace an `if` with optional chaining instead.

*Please note:* For now XFA-support is currently disabled by default, similar to the regular viewer.

### DIFF
--- a/examples/components/pageviewer.js
+++ b/examples/components/pageviewer.js
@@ -54,11 +54,13 @@ loadingTask.promise.then(function (pdfDocument) {
       scale: SCALE,
       defaultViewport: pdfPage.getViewport({ scale: SCALE }),
       eventBus,
-      // We can enable text/annotations layers, if needed
+      // We can enable text/annotation/xfa/struct-layers, as needed.
       textLayerFactory: new pdfjsViewer.DefaultTextLayerFactory(),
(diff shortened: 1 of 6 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #13943, opened 2021-08-29
Use `async` a bit more in the API

Snuffleupagus:
This patch changes the `PDFDocumentLoadingTask.destroy`-method and the `_fetchDocument`-function to be `async`, which slightly simplifies the relevant code.

Furthermore, remove the catch-handler from the `WorkerTransport.getPageIndex`-method since it's no longer needed. Given that the `MessageHandler` is nowadays wrapping every possible Exception, it's no longer necessary to try and re-wrap the reason here.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #13944, opened 2021-08-29
Re-factor the `setPDFNetworkStreamFactory` usage for the unit-tests (PR 13549 follow-up)

Snuffleupagus:
This should have been part of PR #13549, since we no longer support browsers without native Fetch API and ReadableStream implementations.


timvandermeij (approved):

timvandermeij:

PR #13949, opened 2021-08-30
Only use base encoding if it's populated. (bug 1727053)

brendandahl:
The font dict in this file has an encoding entry, but only specifies a
differences map. The base encoding is empty in this case and shouldn't
be used.


Snuffleupagus (approved):
Looks reasonable to me, thank you!

PR #13955, opened 2021-09-01
Always prefer the post-table for TrueType fonts with (0, x) cmap-tables (issue 13433)

### LINKED ISSUE #13433: Chars missing from a document
Multiple chars are missing from the document below. Nothing is reported in the console.

URL: [pos36.pdf](https://github.com/mozilla/pdf.js/files/6537699/pos36.pdf)

Snuffleupagus:
While I don't know if this is necessarily the "correct" solution, it does fix issue 13433 without breaking any of the existing reference-tests.


timvandermeij (approved):

timvandermeij:
Looks simple enough. Thanks!

PR #13956, opened 2021-09-01
Add app options to limit print range

TcT2k:
This adds application options which allow to limit which pages are
rendered when printing. This is not usefull for the default viewer
but allows applications to programatically print specifc pages
independent of user selection.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1819,6 +1819,14 @@ const PDFViewerApplication = {
       optionalContentConfigPromise,
       this.l10n
     );
+    const printRangeStart = AppOptions.get("printRangeStart");
+    const printRangeEnd = AppOptions.get("printRangeEnd");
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
> This is not usefull for the default viewer
> but allows applications to programatically print specifc pages
> independent of user selection.

Please see http://mozilla.github.io/pdf.js/getting_started/#introduction (emphasis mine):

Hence I'm really sorry to say that this doesn't seem like the kind of functionality that should be added to the default viewer, since these options are (essentially) unused *and* untested as-is. (Also, the implementation does feel somewhat "hacky" in the way it reaches into the active `printService` and thus likely wouldn't even work with e.g. `FirefoxPrintService`.) 
Please keep in mind that it's usually easy to add code, but every feature/option added will incur a maintained cost over time.

Issue #12384, opened 2020-09-16
Remove the unofficial TypeScript definitions from DefinitelyTyped


Snuffleupagus:
Currently there's *unofficial*, likely unmaintained, and out-of-date TypeScript definitions published via https://github.com/DefinitelyTyped/DefinitelyTyped for the `pdfjs-dist` library; see https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/pdfjs-dist

Now that we, from version `2.6.347`, are shipping *official* TypeScript definitions through https://github.com/mozilla/pdfjs-dist, it'd be a good idea to submit a PR to DefinitelyTyped to remove its TypeScript definitions; see https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package

billyjanitsch:
The official TS definitions bundled in 2.6.347, while much appreciated, only include bindings for the top-level API and therefore lack many of the useful types currently exposed by the DefinitelyTyped definitions, such as:

- `PDFPageProxy`
- `PDFDocumentProxy`
- `PDFRenderTask`
- `PDFDocumentLoadingTask`

It's possible to access these types through the official definitions by using complex sequences of type operations, e.g.:

```ts
type PDFDocumentProxyPromise = ReturnType<typeof getDocument>['promise']
type PDFDocumentProxy = PDFDocumentProxyPromise extends Promise<infer U> ? U : never
```

However, this is inconvenient and also unconventional within the TypeScript ecosystem. Type definitions for libraries usually export typings for important objects/classes (without exporting their values).

Would you consider exposing these types in the official type definitions before removing the typings from DefinitelyTyped?

timvandermeij:
I would say so since I think they are all part of the public API; see for example https://github.com/mozilla/pdf.js/blob/master/src/display/api.js#L2966-L2967. The TypeScript definitions are generated from our JSDoc comments because PDF.js itself is not a TypeScript project, so any improvements for the typings are much appreciated from the community who have already helped a lot to get the current typings in place.

Snuffleupagus:
> I would say so since I think they are all part of the public API;

Well, the public API is actually only https://github.com/mozilla/pdf.js/blob/master/src/pdf.js :-)

> see for example https://github.com/mozilla/pdf.js/blob/master/src/display/api.js#L2966-L2967

That was exposed there for the unit-tests specifically, as far as I remember.

[...]

PR #13957, opened 2021-09-01
Export RenderTask type

pardoman:
While using TypeScript, I noticed that the `RenderTask` type is being returned out of `pageProxy.render(renderContext)`, but the package `"pdfjs-dist/types/display/api"` isn't exporting this type.

This PR adds `RenderTask` to the export group.

```ts
/**
 * Sample pseudo-code; I'm not bothering with proper async management in this snippet.
 * Proper async management is left as an exercise to the reader.
 */
import {
  PDFDocumentProxy,
  PDFPageProxy,
  RenderParameters,
  // RenderTask <-- Missing ~ being added in this PR
} from "pdfjs-dist/types/display/api";
import { PageViewport } from "pdfjs-dist/types/display/display_utils";

const pdfProxy: PDFDocumentProxy = getProxySomehow();
const pageProxy: PDFPageProxy = pdfProxy.getPage(1)
const viewport: PageViewport = pageProxy.getViewport({ scale: 1 });

const renderContext: RenderParameters = {
    canvasContext: getContextSomehow(),
    viewport: viewport
  };

// Type RenderTask is missing
const task: RenderTask = pageProxy.render(renderContext);
```

Thank you.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -3364,6 +3364,7 @@ export {
   PDFDocumentProxy,
   PDFPageProxy,
   PDFWorker,
+  RenderTask,
   setPDFNetworkStreamFactory,
   version,
 };


Snuffleupagus:
Thanks for the PR, however we unfortunately cannot accept extending the official API-surface only for the TypeScript definitions; please refer to e.g. https://github.com/mozilla/pdf.js/pull/13930#issuecomment-906201600 and the links provided therein for additional information about an acceptable work-around.

pardoman:
Thanks @Snuffleupagus for taking a look at the PR.

I looked at the proposed workaround in the supplied link, but I think it requires me to build the code myself instead of consuming the types from `pdfjs-dist`.

Anyhow, if anyone else is dealing with this issue in TypeScript, I just provided a stub interface and moved on.
Example:
```ts
  // Pseudo-code again
  interface RenderTask {
    promise: Promise<void> // getter only
  }
  const task: RenderTask = pageProxy.render(renderContext);
  task.then(doSomething);
```

Snuffleupagus:
> I looked at the proposed workaround in the supplied link, but I think it requires me to build the code myself instead of consuming the types from `pdfjs-dist`.

I realize that it maybe wasn't clear, but the point is that we'd accept a patch that implements the idea from e.g. https://github.com/mozilla/pdf.js/issues/12384#issuecomment-695143287 since that should (hopefully) fix the `pdfjs-dist` types without you needing to build them yourself.

pardoman:
@Snuffleupagus got it, thanks again!

pardoman:
Hi @Snuffleupagus 

I tried adding 
```tsx
/**
 * @typedef {import("./display/api").RenderTask} RenderTask
 */
```
into `src/pdf.js` as shown in the linked comment, then run `gulp types`, and checked the content in `build/types/src/pdf.d.ts` and found the following output:
```ts
export type RenderTask = import("./display/api").RenderTask;
import { addLinkAttributes } from "./display/display_utils.js";
import { getFilenameFromUrl } from "./display/display_utils.js";
... // more stuff
```
So far so good!

The problem now is that `type RenderTask` is mapped to `any`, which doesn't really help:
image

And in a way it makes sense to me, given that `RenderTask` is not being exported from `src/display/api.js`.

Maybe I'm missing something? Any further guidance is much appreciated!

Snuffleupagus:
> The problem now is that `type RenderTask` is mapped to `any`, which doesn't really help:

That's indeed unfortunate!
I suppose that as long as the `RenderTask` isn't exported *directly* through `src/pdf.js` it might be OK if it's exported in `src/display/api.js`, however we really don't want to have "unused" exports anywhere in the code-base (it could e.g. be at constant risk of being cleaned-up at any point).

Possibly we could reconcile the points above by adding unit-tests, please let me look into this a little bit more.

Snuffleupagus:
Please check if PR #14013 helps.

pardoman:
> Please check if PR #14013 helps.

I checked this other PR, but it's the same story: For the solution to work, the symbols need to be exported.

PR #13959, opened 2021-09-01
Correctly pad strings when saving an encrypted pdf (bug 1726789)

calixteman:


timvandermeij:
Would it be possible to add a unit test for this since this is core functionality? This particular function can be called with the input that failed before and there is even a helper function that asserts that it can also be decoded again: https://github.com/mozilla/pdf.js/blob/7d9a83012bc8a8f83ae27409e958c5b13af0a209/test/unit/crypto_spec.js#L587

calixteman:
Before the patch, pdf.js, evince, mupdf and qpdf were able to decrypt correctly the string but not foxit, chrome and acrobat.
There is already a test with a string where length is a multiple of 16 (e.g. https://github.com/mozilla/pdf.js/blob/master/test/unit/crypto_spec.js#L836).
So I don't see how to have a test which fails without the patch, any idea will be appreciated.

PR #13961, opened 2021-09-02
Simplify some regular expressions

Snuffleupagus:
There's a fair number of regular expressions througout the code-base which are slightly more verbose than strictly necessary, in particular:
 - We have a lot of regular expressions that use `[0-9]` explicitly, and those can be simplified to use `\d` instead.
 - We have one instance of a regular expression containing a `A-Za-z0-9_` sequence, which can be simplified to use `\w` instead.

### DIFF
--- a/src/core/core_utils.js
+++ b/src/core/core_utils.js
@@ -208,7 +208,7 @@ function isWhiteSpace(ch) {
  * each part of the path.
  */
 function parseXFAPath(path) {
-  const positionPattern = /(.+)\[([0-9]+)\]$/;
+  const positionPattern = /(.+)\[(\d+)\]$/;
   return path.split(".").map(component => {
     const m = component.match(positionPattern);
     if (m) {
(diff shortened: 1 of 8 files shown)


calixteman:
Would it be possible to have some linting on regexs ?

calixteman:
I'm the author of most of the regexs here and at least I can tell you why I did that: in some languages  ̀\d` is a shortcut for `digits` which is not limited to digits 0,...,9 but which is a shortcut for all digits in Unicode. So I prefer to be explicit to avoid to have to remember what `\d` is exactly (in general I try to avoid such shortcuts in regex in whatever language).
Out of curiosity, I searched in firefox code base:
https://searchfox.org/mozilla-central/search?q=%5B0-9%5D&case=true&path=*.js

Anyway I'm fine with this change.

timvandermeij (approved):

timvandermeij:
Thanks! I'm also always a bit careful with `\d` since it indeed has a subtly different meaning in other languages where it also matches Unicode digits, but fortunately for JavaScript there is no such behavior according to MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet

PR #13964, opened 2021-09-03
Use minimized file in pdf.worker.entry.js

hand-dot:
The minimized pdf.worker.js is not used in the production build of pdf.worker.entry.js
It has been fixed to use the minimized file at build time.

This fix will reduce the size of the bundle.

### DIFF
--- a/src/pdf.worker.entry.js
+++ b/src/pdf.worker.entry.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-(typeof window !== "undefined"
-  ? window
-  : {}
-).pdfjsWorker = require("./pdf.worker.js");
+(typeof window !== "undefined" ? window : {}).pdfjsWorker =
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
First of all, I'm not sure that I even understand the added `PDFJSDev`-checks since none of this code matters for anything *except* official builds of the PDF.js library (it's definitely not being used during development).
Secondly, this file is only used together with the [Webpack-example](https://github.com/mozilla/pdf.js/tree/master/examples/webpack) and shouldn't be used elsewhere (since that'd load the `pdf.worker.js` file into the main-thread which effectively disable Workers). The reason to use, or not use, a minified build is really up to the specific user/situation and is probably not something that we should *force* like this; note also the warning in https://github.com/mozilla/pdf.js/tree/master/examples/webpack#minification

All-in-all, I'm sorry to say that this doesn't look like a change that we need/should make here.

PR #13967, opened 2021-09-03
XFA - Overwrite AcroForm dictionary when saving if no datasets in XFA (bug 1720179)

calixteman:
- aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1720179
  - in some pdfs the XFA array in AcroForm dictionary doesn't contain an entry for 'datasets' (which contains saved data), so basically this patch allows to overwrite the AcroForm dictionary with an updated XFA array when doing an incremental update.

In fixing this bug I found another one: https://github.com/mozilla/pdf.js/pull/13966 and with both fix we can save the form in the above bug and opening it in Acrobat.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -130,6 +130,11 @@ class Catalog {
     return shadow(this, "acroForm", acroForm);
   }
 
+  get acroFormRef() {
+    const value = this._catDict.getRaw("AcroForm");
+    return shadow(this, "acroFormRef", isRef(value) ? value : null);
+  }
+
   get metadata() {
     const streamRef = this._catDict.getRaw("Metadata");
     if (!isRef(streamRef)) {
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -573,6 +573,7 @@ class WorkerMessageHandler {
         const promises = [
           pdfManager.onLoadedStream(),
           pdfManager.ensureCatalog("acroForm"),
+          pdfManager.ensureCatalog("acroFormRef"),
           pdfManager.ensureDoc("xref"),
           pdfManager.ensureDoc("startXRef"),
         ];
@@ -597,6 +598,7 @@ class WorkerMessageHandler {
         return Promise.all(promises).then(function ([
           stream,
           acroForm,
+          acroFormRef,
           xref,
           startXRef,
           ...refs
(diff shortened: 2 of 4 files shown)


src/core/catalog.js
```diff
@@ -130,6 +130,15 @@ class Catalog {
...
   }
 
+  get acroFormRef() {
+    let acroFormRef = null;
+    const value = this._catDict.getRaw("AcroForm");
+    if (isRef(value)) {
+      acroFormRef = value;
+    }
+    return shadow(this, "acroFormRef", acroFormRef);
+  }
```
Snuffleupagus, line 140: This could be simplified slightly; how about just:
```javascript
const value = this._catDict.getRaw("AcroForm");
return shadow(this, "acroFormRef", isRef(value) ? value : null);
```

src/core/worker.js
```diff
@@ -621,15 +623,22 @@ class WorkerMessageHandler {
             }
           }
 
-          const xfa = (acroForm instanceof Dict && acroForm.get("XFA")) || [];
+          const xfa = (acroForm instanceof Dict && acroForm.get("XFA")) || null;
           let xfaDatasets = null;
+          let hasDatasets = false;
```
Snuffleupagus, line 628: Do you actually need this separate boolean?
Can't you just check `!!xfaDatasets` in the `src/core/writer.js` code, and thus avoid the need to track closely related state twice?

src/core/writer.js
```diff
@@ -146,10 +146,53 @@ function writeXFADataForAcroform(str, newRefs) {
...
     return;
   }
+
+  if (!hasDatasets) {
+    if (!acroFormRef) {
+      return;
+    }
+
+    // We've a XFA array which doesn't contain a datasets entry.
+    // So we'll update the acroform dictionary to have an XFA containing
```
Snuffleupagus, line 169: Nit: `acroform` -> `AcroForm`.

src/core/writer.js
```diff
@@ -146,10 +146,53 @@ function writeXFADataForAcroform(str, newRefs) {
...
+  xref,
+  xrefInfo,
+}) {
+  if (xref === null) {
     return;
   }
+
+  if (!hasDatasets) {
+    if (!acroFormRef) {
+      return;
```
Snuffleupagus, line 165: Maybe add a `warn` here about saving not working?

src/core/worker.js
calixteman: In case `xfaDatasets` is `null`, I set it to a new reference after having visited the `XFA` array so it's no more null.
So the only way for `xfaDatasets` to be `null` is to not have an array for `XFA`.
So when we have an array we must know if the reference we've has been newly created (and in this case we must update AcroForm) or not.

We could probably get rid of this boolean in moving https://github.com/mozilla/pdf.js/blob/master/src/core/worker.js#L650 in `writer.js` because the afaik the xref for the xref table must be the last one (so it's why I get the new ref for `xfaDatasets` before calling `incrementalUpdate`), but I guess it's better to do that in a follow-up.

PR #13973, opened 2021-09-04
[api-minor] Change `{PDFPageView, PDFThumbnailView}.update` to take a parameter object

Snuffleupagus:
- [api-minor] Change `{PDFPageView, PDFThumbnailView}.update` to take a parameter object
 
   The old `update`-signature started to annoy me back when I added optional content support to the viewer, since we're (often) forced to pass in a bunch of arguments that we don't care about whenever these methods are called.

   This is tagged `api-minor` since `PDFPageView` is being used in the `pageviewer` component example, and it's thus possible that these changes could affect some users; the next commit adds fallback handling for the old format.

 - [GENERIC viewer] Add fallback logic for the old `PDFPageView.update` method signature
    
   This is done separately from the previous commit, to make it easier to revert these changes in the future.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -405,9 +405,9 @@ class BaseViewer {
 
     const pageNumber = this._currentPageNumber;
 
-    for (let i = 0, ii = this._pages.length; i < ii; i++) {
-      const pageView = this._pages[i];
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #13975, opened 2021-09-05

Snuffleupagus:


Snuffleupagus:
Given the `webpack-stream` changes in particular, I've run `gulp mozcentral` and `gulp generic` locally with/without these patches and `diff`ed the output. Apart from updated version/build numbers, there were no actual code changes caused by these patches.

/botio-linux preview

timvandermeij (approved):

timvandermeij:
Thanks for the updates!

PR #13976, opened 2021-09-05
Reduce the size of `TextLayerRenderTask._textDivProperties` in "regular" text-selection mode

Snuffleupagus:
While these changes will obviously not have a significant effect on overall memory usage, it cannot hurt as far as I'm concerned. This patch makes the following changes:
 - Clear out `_textDivProperties` once rendering is done, since those properties are only necessary to keep alive when *enhanced* text-selection is being used.

 - Reduce the size of the `_textDivProperties`-entries by default, since a majority of the properties are only relevant when *enhanced* text-selection is being used.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13977, opened 2021-09-05
[api-minor] Reduce `postMessage` overhead, in `PartialEvaluator.getTextContent`, by sending text chunks in batches (issue 13962)

### LINKED ISSUE #13962: Slow text extraction in large documents due to cross-thread messaging overhead
While looking into an issue with slow text extraction with a large OCR-ed PDF, I saw in Chrome's profiler that there appeared to be a lot of overhead from calls to `postMessage`.

PDF.js extracts a total of ~500,000 text items from the attached 590-page PDF. When extracting text (eg. using the search function) these text items get sent from the worker thread to the main thread in a series of ~200,000 `postMessage` calls. I checked this by adding a `message` event listener to the `Worker` instance. I then tried modifying `src/core/evaluator.js` to comment out the most frequent `enqueueChunk()` call so that it would flush items in larger batches and thus send the same data with fewer cross-thread messages. It seems to make a substantial difference in Chrome.

I'd be happy to send a PR to fix this, but I'm not sure what the best/cleanest way is to resolve this. The smallest change I can think of would be to modify the `next` callback below so that it invokes `enqueueChunk` only once every N calls. Would that break anything? Should the problem be fixed at a higher level?

## Steps to reproduce

Attach (recommended) or Link to PDF file here: [leviathan.pdf](https://github.com/mozilla/pdf.js/files/7100224/leviathan.pdf)

Configuration:
- Web browser and its version: 95.0.4621.4
- Operating system and its version: macOS 11
- PDF.js version: 0a366dda6a174292e6d109df526f835bfabf87a8
- Is a browser extension: No

Open the attached PDF in PDF.js's development server and search for the phrase "for such truth", which occurs near the very end of the document.

On my 2020 Intel MacBook Pro, testing in Chrome 95, PDF.js takes ~15 seconds to find a match.

Now apply the following diff, then reload the viewer and retry the search:

```diff
diff --git a/src/core/evaluator.js b/src/core/evaluator.js
index 90615b228..85d6325a9 100644
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -2669,7 +2669,8 @@ class PartialEvaluator {
 
     return new Promise(function promiseBody(resolve, reject) {
       const next = function (promise) {
-        enqueueChunk();
+        // TESTING
+        // enqueueChunk();
         Promise.all([promise, sink.ready]).then(function () {
           try {
             promiseBody(resolve, reject);
```

With the above diff applied, the search returns the same result but is much faster (~3-4 seconds).

One quirk I noticed is that initial search speed is much slower (~20 seconds) if Chrome's devtools are open when the search happens. This applies with or without the above diff.

Snuffleupagus:
Following the STR in the issue, this patch reduces the number of `PartialEvaluator.getTextContent`-related `postMessage`-calls by approximately 78 percent.[1]
Note that by enforcing a relatively low value when batching text chunks, we should thus improve worst-case scenarios while not negatively affect all `textLayer` building.

While working on these changes I noticed, thanks to our unit-tests, that the implementation of the `appendEOL` function unfortunately means that the number and contents of the textItems could actually be affected by the particular chunking used.
That seems *extremely* unfortunate, since in practice this means that the particular chunking used is thus observable through the API. Obviously that should be a completely internal implementation detail, which is why this patch also modifies `appendEOL` to mitigate that.[2]

Given that this patch adds a *minimum* batch size in `enqueueChunk`, there's obviously nothing preventing it from becoming a lot larger then the limit (depending e.g. on the PDF structure and the CPU load/speed).
While sending more text chunks at once isn't an issue in itself, it could become problematic at the main-thread during `textLayer` building. Note how both the `PartialEvaluator` and `CanvasGraphics` implementations utilize `Date.now()`-checks, to prevent long-running parsing/rendering from "hanging" the respective thread. In the `textLayer` building we don't utilize such a construction[3], and streaming of textContent is thus essentially acting as a *simple* stand-in for that functionality.
Hence why we want to avoid choosing a too large minimum batch size, since that could thus indirectly affect main-thread performance negatively.

Improves and/or fixes #13962

---
[1] While it'd be possible to go even lower, that'd likely require more invasive re-factoring/changes to the `PartialEvaluator.getTextContent`-code to ensure that the batches don't become too large.

[2] This should also, as far as I can tell, explain some of the regressions observed in the "enhance" text-selection tests back in PR #13257.
     Looking closer at the `appendEOL` function it should potentially be changed even more, however that should probably not be done here.

[3] I'd really like to avoid implementing something like that for the `textLayer` building as well, given that it'd require adding a fair bit of complexity.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -107,6 +107,17 @@ const PatternType = {
   SHADING: 2,
 };
 
+// Optionally avoid sending individual, or very few, text chunks to reduce
+// `postMessage` overhead with ReadableStream (see issue 13962).
+//
+// PLEASE NOTE: This value should *not* be too large (it's used as a lower limit
+// in `enqueueChunk`), since that would cause streaming of textContent to become
+// essentially useless in practice by sending all (or most) chunks at once.
+// Also, a too large value would (indirectly) affect the main-thread `textLayer`
+// building negatively by forcing all textContent to be handled at once, which
+// could easily end up hurting *overall* performance (e.g. rendering as well).
+const TEXT_CHUNK_BATCH_SIZE = 10;
+
 const deferred = Promise.resolve();
 
 // Convert PDF blend mode names to HTML5 blend mode names.
@@ -2573,8 +2584,6 @@ class PartialEvaluator {
       if (textContentItem.initialized) {
         textContentItem.hasEOL = true;
         flushTextContentItem();
-      } else if (textContent.items.length > 0) {
-        textContent.items[textContent.items.length - 1].hasEOL = true;
       } else {
         textContent.items.push({
           str: "",
@@ -2656,20 +2665,24 @@ class PartialEvaluator {
       textContentItem.str.length = 0;
     }
 
-    function enqueueChunk() {
+    function enqueueChunk(batch = false) {
       const length = textContent.items.length;
-      if (length > 0) {
-        sink.enqueue(textContent, length);
-        textContent.items = [];
-        textContent.styles = Object.create(null);
+      if (length === 0) {
+        return;
       }
+      if (batch && length < TEXT_CHUNK_BATCH_SIZE) {
+        return;
+      }
+      sink.enqueue(textContent, length);
+      textContent.items = [];
+      textContent.styles = Object.create(null);
     }
 
     const timeSlotManager = new TimeSlotManager();
 
     return new Promise(function promiseBody(resolve, reject) {
       const next = function (promise) {
-        enqueueChunk();
+        enqueueChunk(/* batch = */ true);
         Promise.all([promise, sink.ready]).then(function () {
           try {
             promiseBody(resolve, reject);
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -737,8 +737,6 @@ class WorkerMessageHandler {
 
     handler.on("GetTextContent", function wphExtractText(data, sink) {
       const pageIndex = data.pageIndex;
-      sink.onPull = function (desiredSize) {};
-      sink.onCancel = function (reason) {};
 
       pdfManager.getPage(pageIndex).then(function (page) {
         const task = new WorkerTask("GetTextContent: page " + pageIndex);
--- a/src/shared/message_handler.js
+++ b/src/shared/message_handler.js
@@ -448,7 +448,7 @@ class MessageHandler {
         }
         // Reset desiredSize property of sink on every pull.
         this.streamSinks[streamId].desiredSize = data.desiredSize;
-        const { onPull } = this.streamSinks[data.streamId];
+        const { onPull } = this.streamSinks[streamId];
         new Promise(function (resolve) {
           resolve(onPull && onPull());
         }).then(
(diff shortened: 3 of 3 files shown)


src/core/evaluator.js
```diff
@@ -2656,20 +2654,26 @@ class PartialEvaluator {
...
+    function enqueueChunk(batch = false) {
       const length = textContent.items.length;
-      if (length > 0) {
-        sink.enqueue(textContent, length);
-        textContent.items = [];
-        textContent.styles = Object.create(null);
+      if (length === 0) {
+        return;
+      }
+      if (batch && length < 5) {
```
Snuffleupagus, line 2662: We could likely bump this up to e.g. `10` instead (that'd bring the `postMessage`-reduction to ~75%) without too much problem, but I figured that starting small might not be a bad idea.

src/core/evaluator.js
```diff
@@ -2656,20 +2654,26 @@ class PartialEvaluator {
...
+    function enqueueChunk(batch = false) {
       const length = textContent.items.length;
-      if (length > 0) {
-        sink.enqueue(textContent, length);
-        textContent.items = [];
-        textContent.styles = Object.create(null);
+      if (length === 0) {
+        return;
+      }
+      if (batch && length < 10) {
```
timvandermeij, line 2662: I'd prefer we make this 10 a `TEXT_CHUNK_BATCH_SIZE` constant at the top of the file with a short explanation of the value as a comment there, basically a small extract from the nice commit message that briefly explains why it should not be set too large.

src/core/evaluator.js
```diff
@@ -2656,20 +2654,26 @@ class PartialEvaluator {
...
+        // Avoid sending individual, or very few, text chunks to reduce
+        // `postMessage` overhead (see issue 13962).
+        return;
```
timvandermeij, line 2665: I might be missing something here, but if the last batch to be processed has a size < 10, doesn't that mean we'll never enqueue it at all and basically drop it?

src/core/evaluator.js
Snuffleupagus: That's a *very* fair comment, and something that I should probably have done from the start; I'll try to come up with a reasonable (and not too long-winded comment).

src/core/evaluator.js
Snuffleupagus: Note how at the *end* of the "GetTextContent"-parsing we'll do one final sending of any remaining chunks:
https://github.com/mozilla/pdf.js/blob/168aa5efb12e3c0f6a0a1713cde3717b337c3484/src/core/evaluator.js#L3085-L3101

Furthermore, I very deliberately implemented the batching to be *opt-in* precisely to avoid the problem you're asking about here; so unless I've somehow messed up here we should be good.

---

With the exception of some movement in a couple of "enhance" text-tests, which aren't really my fault (note footnote [2] above), this patch also passes all regular "text" tests locally.

timvandermeij: Yes, the finalization step is the part I missed here, and `enqueueChunk` is indeed called without the `batch` parameter, so we should indeed be good here.

src/core/evaluator.js
Snuffleupagus: I've added a comment; please check that it makes sense and that it clearly conveys why the constant shouldn't be changed lightly!

timvandermeij (approved):

timvandermeij:
Nice work!

Snuffleupagus:
@timvandermeij As always, thanks a lot for reviews :-)

---

*Off-topic: How do you feel about a new release soon, since there's been a fair number of `api-minor` changes recently and also new features such as the just landed `/Interpolate`-support (fixing many old issues)?

PR #13991, opened 2021-09-09
Enable/disable image smoothing based on image interpolate value. (bug 1722191)

### LINKED ISSUE #1344: Blurred images
PDF.js displays blurry plots in PDF documents, e.g.

http://arxiv.org/pdf/1202.5344v1.pdf

When viewed in Chrome or Adobe Acrobat all the plots show sharp-edged squares. When viewed in Preview or PDF.js, all the squares show as blurry pixelated blobs.

### LINKED ISSUE #4706: Blurry/fuzzy embedded images
A fix for this was attempted many times, but still the issue exists.
FF 29.0.

Test PDF: http://rustyx.org/temp/fuzzy-test.pdf

Results

### LINKED ISSUE #5757: [Color rendering] Bug or feature? Colors get blurred (or "smoothened"?) when very small images are scaled up
Here are three different screenshots of PDF viewers (composed into one image) rendering the exact same file:
1. **LEFT:** **_PDF.js**_ (v1.0.1149) running in Chrome/OSX (sorry, I currently don't get PDF.js to work in Firefox/OSX)
2. **CENTER:** **_Adobe Reader**_ (v11.0.10)
3. **RIGHT:** **_Preview.app**_ in OSX 10.9 (v.7.0 -- 826.4)

pdfjs-acroread-previewapp

The PDF source code is written by myself in order to demonstrate a problem I've previously noticed with other (real life) files and provide a _"minimal working example PDF"_. 

**The characteristics of this PDF:**
1. It contains a single image (inline).
2. The image is compose of only 4 pixels with 4 different colors (width x height == 2 x 2).
3. The image is scaled up to 630x630 points on a PDF page sized 640x640 points.

As you can see, only Acrobat renders the 4 pixels as one would expect (IMHO). Or? _(You may disagree with me and regard the way PDF.js renders it as a feature, not a bug...)_

PDF.js deviates from Acrobat's rendering and prefers to show the 4 scaled-up pixels in a smoothened (if you regard it as a feature :-) or heavily blurred (if you regard it as a bug) way.

Preview.app also deviates from Acrobat -- albeit in a different way (you'd think that the image is ridden by heavy JPEG artifacts).

However, there are other open source PDF viewers which all render the file identical to Acrobat's way, with sharp edges between the pixels and uniform color for the four areas:
1. `gs` (Ghostscript
2. `gv` (Ghostscript-based)
3. `MuPDF` (own rendering engine)
4. `Zathura` (Poppler-based)
5. `Evince` (Poppler-based)
6. Chrome's built-in, native PDF viewer (dunno what is its name, maybe _'pdfium'_?)
7. `SumatraPDF` (MuPDF-based)
8. `gv` (Ghostscript-based)

---

<sub>Hmmm... I'm looking for a way to attach my demo PDF, but I can't find it. Maybe I re-write it to use ASCII-only stream for the image data, so I can embed it inline.</sub>

### LINKED ISSUE #8245: PDF Displaying not displaying properly
[document.pdf](https://github.com/mozilla/pdf.js/files/902878/document.pdf)

Configuration:
    Web browser and its version: Firefox 52.0.2 (32-bit)
    Operating system and its version: Windows 10
    PDF.js version: Using version bundled with Firefox

1. View Attached PDF with PDF.js in Firefox
 Similar Issue #8083 

Adobe Reader correctly shows the image and isn't blurry. PDF.js does not. See screenshots for comparison.

screenshot correct

screenshot

The image in the PDF is an Indexed xobject. not sure if this is the issue.

### LINKED ISSUE #9713: Image interpolation incorrect
[image_via_1.3_001.pdf](https://github.com/mozilla/pdf.js/files/1991319/image_via_1.3_001.pdf)

Configuration:
- Web browser and its version: Microsoft Edge 42.17134.1.0 and FF 60.0
- Operating system and its version: W10, 10.0.17134.1 and also on W7 6.1.7601
- PDF.js version: https://mozilla.github.io/pdf.js/web/viewer.html

1. start https://mozilla.github.io/pdf.js/web/viewer.html
2. open file

clipboard01

clipboard02

/Im0 is the lower Image and has /Interpolate true but this isn't done.

File is from
https://storage.googleapis.com/google-code-archive-source/v2/code.google.com/lumi-read/source-archive.zip
and can be found in lumi-read/trunk/PDF_smoketest/pdf/image_via/image_via_1.3_001.pdf

There are more incorrect renderings but I suspect these are related.

brendandahl:
While some of the output looks worse to my eye, this behavior more
closely matches what I see when I open the PDFs in Adobe acrobat.

Fixes #4706
Fixes #9713
Fixes #8245
Fixes #1344
Fixes #5757


brendandahl:
For example:
image
image

brendandahl:
A lot more failures on the bots than locally.

Snuffleupagus (commented):
While I completely agree that this looks good (or better) for most cases, even with the changes in the second commit there appear to be *some* regressions. (Also, please squash the commits before landing this.)

Looking through the test results, and focusing (only) on Firefox, there's a few cases that I do think look worse now when comparing against Adobe Reader:
 - issue3903, page1
 - freeculture, page2
 - wdsg_fitc, page20
 - cmykjpeg, page1
 - issue5644, page1
 - issue7200, page1; the "text" is a lot less readable
 - issue5549, page1
 - issue5475, page1
 - issue5481, all pages

Is there anything that could possibly be done about some of these cases, or are we going to have to accept those changes?

brendandahl:
I'm looking into some heuristics. I was able to fix a few by following [this comment](https://github.com/mozilla/pdf.js/issues/4706#issuecomment-838228855), but there are still a few where acrobat seems to be doing some image smoothing in cases I wouldn't expect.

brendandahl:
This will still need some tweaks. On a high dpi monitor the output now looks very close to acrobat, but on a monitor with a device pixel ratio of 1, pdf.js looks much worse than acrobat. I thought it was maybe do to pdf.js trying to draw images at fractional pixels, but even when I align the image to whole pixel values it still looks worse. I wonder if they're using a different scaling algorithm.

brendandahl:
I think I may have figured it out, instead of comparing the scale to 1 we have to take into account that css pixels per inch are different than pdf pixels per inch. Using that value for deciding when to smooth seems to match acrobats behavior.

src/display/canvas.js
```diff
@@ -37,6 +37,9 @@ const MIN_FONT_SIZE = 16;
 const MAX_FONT_SIZE = 100;
 const MAX_GROUP_SIZE = 4096;
 
+const CSS_PIXELS_PER_INCH = 96.0;
+const PDF_PIXELS_PER_INCH = 72.0;
```
Snuffleupagus, line 41: Given how these values are being used, it appears (as far as I can tell) to be basically the following viewer constant: https://github.com/mozilla/pdf.js/blob/5678c755624a4ab4ff7ebca8620ab1f317568633/web/ui_utils.js#L16

Can we perhaps move the existing `CSS_UNITS` constant into `display_utils.js` (and export it via `pdf.js`) and then re-use here too, instead of essentially duplicating this?

src/display/canvas.js
```diff
@@ -871,6 +874,21 @@ function composeSMask(ctx, smask, layerCtx) {
   ctx.drawImage(mask, 0, 0);
 }
 
+function getImageSmoothingEnabled(transform, interpolate) {
+  const scale = Util.singularValueDecompose2dScale(transform);
+  const actualScale =
+    (window.devicePixelRatio * CSS_PIXELS_PER_INCH) / PDF_PIXELS_PER_INCH;
```
Snuffleupagus, line 880: This is causing the unit-tests to fail, when run in Node.js, since we cannot assume that `window` is defined.
How about simply doing `globalThis.devicePixelRatio || 1` here, since that ought to prevent any issues?

brendandahl:
I need to look into why firefox-issue5549.pdf-page1 is blurry, locally (with lodpi and highdpi) it's getting smoothed in the viewer at 100%. I'm guessing something is different in test harness vs viewer.

brendandahl:
The issue with firefox-issue5549.pdf-page1 was a float rounding error, the scale's were both 1.333... but slightly different.

src/display/canvas.js
Snuffleupagus: Looking at the final patch, I've got one small suggestion for simplifying the new exports a little bit.
However, in the interest of getting this landed sooner rather than later, I'll merge it as-is and file a follow-up PR instead.

Snuffleupagus (approved):
This looks *much* better, and fixes a bunch of old issues which is great; thank you!

PR #13992, opened 2021-09-09
XFA - Remove the checked attribute from the checkbox when unchecked (bug 1729877)

calixteman:
- it aims to fix: https://bugzilla.mozilla.org/show_bug.cgi?id=1729877.

### DIFF
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -967,6 +967,21 @@
        "enableXfa": true,
        "type": "eq"
     },
+    {  "id": "xfa_bug1729877",
+       "file": "pdfs/xfa_bug1729877.pdf",
+       "md5": "1bf5429ea13a090f52edcd49cb0c4a47",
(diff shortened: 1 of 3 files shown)


test/test_manifest.json
```diff
@@ -967,6 +967,21 @@
        "enableXfa": true,
        "type": "eq"
     },
+    {  "id": "xfa_bug1729877",
+       "file": "pdfs/xfa_bug1729877.pdf",
+       "md5": "1bf5429ea13a090f52edcd49cb0c4a47",
+       "link": true,
```
Snuffleupagus, line 973: Are we allowed to include this document in the repo?
 - If no, please update the patch to actually use a *linked* test-case instead.
 - If yes, this line should simply be removed.

marco-c (approved):
r+, with @Snuffleupagus's comment addressed

PR #13993, opened 2021-09-09
Don't create `PDFViewerApplication.pdfHistory` when the browsing history is disabled

Snuffleupagus:
Similar to other viewer components, e.g. the `PDFFindBar` and `PDFPresentationMode`, there's no need to create a `PDFHistory`-instance when it's not going to be used.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #13995, opened 2021-09-09
XFA - Handle $record shorcut in SOM expression (issue #13994)

### LINKED ISSUE #13994: XFA Pdf File unable to load prefilled values
[Form_MGT-14-try-1.pdf](https://github.com/mozilla/pdf.js/files/7138335/Form_MGT-14-try-1.pdf)

Configuration:
- Web browser and its version: Google chrome, Mozilla Nightly
- PDF.js version: 2.11.213 (tested on Demo available on github repo)
- Is a browser extension: -

1.  Fill the pdf, (try to fill email address) or try to open the attached filled pdf in Demo/pdf.js 

Pdf should be filled. (Email address in case of above pdf)
image

Unable to see the prefilled values.
image

None

--

calixteman:


marco-c:
@calixteman is the difference in `firefox-xfa_bug1722030_1-page2` expected?

calixteman:
In fact `Référence de publication :` is correct for the first entry but we should have it for the other entries.

calixteman:
So the first node is correctly bound thanks to this patch but the following ones are not and use their default value.
If I comment: https://github.com/mozilla/pdf.js/blob/d4caa8756029f4748b46398122162975927e0be6/src/core/xfa/bind.js#L595
then all fields have the correct value (`Référence de publication :`).

marco-c (approved):

calixteman:
I'll fix the `consumed` bug in a follow-up.

PR #13996, opened 2021-09-10
Make `verifyManifestFiles` fail for non-linked test-cases with a `"link": true`-entry

Snuffleupagus:
Currently it's possible to accidentally, e.g. by simply copy-and-pasting from an existing test-case, add an unnecessary `"link": true`-entry for locally available PDF files.
This leads to inconsistencies in the manifest file, and doesn't feel like a great developer experience. However we can easily fix it by having `verifyManifestFiles` fail in this situation, and doing so actually turned up a couple of existing cases.

### DIFF
--- a/test/downloadutils.js
+++ b/test/downloadutils.js
@@ -142,6 +142,15 @@ function verifyManifestFiles(manifest, callback) {
       verifyNext();
       return;
     }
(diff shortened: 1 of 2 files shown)


calixteman (approved):
LGTM, thanks for doing this: it's helpful.

PR #13998, opened 2021-09-10
Write boolean value when saving a form (bug 1729971)

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1729971#c4.

### DIFF
--- a/src/core/writer.js
+++ b/src/core/writer.js
@@ -83,10 +83,14 @@ function writeValue(value, buffer, transform) {
     buffer.push(`(${escapeString(value)})`);
   } else if (typeof value === "number") {
     buffer.push(numberToString(value));
+  } else if (typeof value === "boolean") {
+    buffer.push(`${value.toString()}`);
   } else if (isDict(value)) {
     writeDict(value, buffer, transform);
   } else if (isStream(value)) {
     writeStream(value, buffer, transform);
+  } else {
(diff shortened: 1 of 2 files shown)


src/core/writer.js
```diff
@@ -83,10 +83,14 @@ function writeValue(value, buffer, transform) {
     buffer.push(`(${escapeString(value)})`);
   } else if (typeof value === "number") {
     buffer.push(numberToString(value));
+  } else if (typeof value === "boolean") {
+    buffer.push(`${value.toString()}`);
```
Snuffleupagus, line 87: I might be missing something obvious here, but why do you need to wrap this in a string?
Isn't `value.toString()` *already* a string here, in which case just `buffer.push(value.toString());` ought to work.

Snuffleupagus (approved):
r=me with that comment addressed; thank you!

timvandermeij:
Looks good to me too with the comment above addressed.

src/core/writer.js
Snuffleupagus: Also, as a follow-up, we should probably fix the pre-existing cases above to reduce unnecessary confusion when reading this code.

Snuffleupagus:
Actually, let's land this as-is such that it won't *just* miss the upcoming release and we can fix the comment in a follow-up.

PR #14004, opened 2021-09-11
[CRX] enableScripting=false by default in Chrome extension

Rob--W:
The scripting engine is bundled for users who'd like to use the feature, but it is disabled by default.

I just audited the scripting implementation and found that it is generally not unsafe, hence I decided to include it rather than disabling it in gulpfile.


Snuffleupagus (approved):
Thank you!

PR #14005, opened 2021-09-12
[GENERIC viewer] Always show the Download-buttons, to allow saving of forms (issue 13997)

### LINKED ISSUE #13997: Download button disappears after opening a pdf
STR:
  - Go to https://mozilla.github.io/pdf.js/web/viewer.html
  - Click on "Open File" and open a pdf

The `Download` button and the `Current view` one just disappear.
I can reproduce this bug with Firefox nightly, Chrome stable and Epiphany on debian.

Snuffleupagus:
Originally the library/viewer didn't support forms, and the hiding of the Download-buttons (when new documents are opened) didn't really matter all that much. Hence the simplest solution, at the time, was to hide the Download-buttons since we use the URL as a fallback when downloading data in the GENERIC `DownloadManager`.
Nowadays we obviously want to support saving of forms in the GENERIC viewer, regardless of how the document was opened, which could thus *potentially* lead to the fallback download-URL being wrong.

In order to be able to show the Download-buttons unconditionally, this patch slightly re-factors the viewer to track the download-URL *separately* to prevent any issues there.

*Please note:* As mentioned in the issue, the ViewBookmark-buttons are specific to the initial URL when the viewer is first opened. Hence they (still) don't make sense when a new document has been opened, since it's then impossible to obtain a usable link to the *currently* active document.

Fixes #13997

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -251,6 +251,7 @@ const PDFViewerApplication = {
   isViewerEmbedded: window.parent !== window,
   url: "",
   baseUrl: "",
+  _downloadUrl: "",
   externalServices: DefaultExternalServices,
   _boundEvents: Object.create(null),
   documentInfo: null,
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
It looks like the Linux bot is hanging after the recent makeref :-(

/botio-windows preview

timvandermeij (approved):

timvandermeij:
Thanks!

PR #14006, opened 2021-09-12
[src/core/writer.js] Remove unnecessary string-wrapping for boolean values in `writeValue` (PR 13998 follow-up)

Snuffleupagus:


calixteman (approved):
Thanks for the follow-up.

PR #14007, opened 2021-09-12
[src/core/writer.js] Support `null` values in the `writeValue` function

Snuffleupagus:
*This fixes something that I noticed, having recently looked at both the `Lexer.getObj` and `writeValue` code.*

Please note that I unfortunately don't have an example of a form where saving fails without this patch. However, given its overall simplicity and that unit-tests are added, it's hopefully deemed useful to fix this potential issue pro-actively rather than waiting for a bug report.

At this point one might, and rightly so, wonder if there's actually any real-world PDF documents where a `null` value is being used?
Unfortunately the answer is *yes*, and we have a couple of examples in the test-suite (although none of those are related to forms); please see: `issue1015`, `issue2642`, `issue10402`, `issue12823`, `issue13823`, and `pr12564`.

### DIFF
--- a/src/core/writer.js
+++ b/src/core/writer.js
@@ -89,6 +89,8 @@ function writeValue(value, buffer, transform) {
     writeDict(value, buffer, transform);
   } else if (isStream(value)) {
     writeStream(value, buffer, transform);
+  } else if (value === null) {
+    buffer.push("null");
   } else {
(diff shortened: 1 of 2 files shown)


calixteman (approved):
Thanks for doing this.

Issue #13405, opened 2021-05-20
Annotation layer in shadow DOM


cetinsert:
Stating something I have encountered in testing. The fixes mentioned at the end do not cover all cases.

---

# 1
The `document.get*` (`document.query*`) methods used in annotation layer radio button selection switch logic work fine in this configuration.
```html
  <html>
    ⋱
      <#light-ancestor>
        ⋱
          <section class="⋯ radioButton" ⋯>
            <input name=a ⋯>
      <#light-ancestor>
        ⋱
          <section class="⋯ radioButton" ⋯>
            <input name=a ⋯>
```

# 2

They do not work if shadow DOM is involved:

```html
  <html>
    ⋱
      #shadow-root
        ⋱
          <#ancestor>
            <section class="⋯ radioButton" ⋯>
              <input name=a ⋯>
            <section class="⋯ radioButton" ⋯>
              <input name=a ⋯>
``` 

For the above case, an incomplete improvement can be gained by scoping down from `document` to `input` (`event.target`) elements' closest not-parent not-section ancestors (`.parentElement.closest(:not(section))`) or grandparents `.parentElement.parentElement` as the end of this comment demonstrates.

# 3

I have not attacked the more involved case demonstrated below. Any suggestions, web standards / MDN references for the below are welcome and the desired output of this issue.

```html
  <html>
    ⋱
      #shadow-root
        ⋱
          <#ancestor>
            <section class="⋯ radioButton" ⋯>
              <input name=a ⋯>
      ⋮
      #shadow-root
        ⋱
          <#ancestor>
            <section class="⋯ radioButton" ⋯>
              <input name=a ⋯>
``` 

I am thinking of replacing `document` with an object that can query across all related (say of the same PDF file) `#shadow-root`s while maintaining API for `document.get*` (`document.query*`) functions. Is there a standard or anything else that I should be aware of for creating/maintaining such a selectively shadow-piercing `document`-like object?

---

The following 3 calls to `document.getElementsByName` effectively caused **all** clicked/tapped **radio buttons** to **stay checked** because finding the ones that are not the target of the current click/tap to clear would fail.

For the 2. case illustrated above, replacing `document.getElementsByName(⋯)` in

https://github.com/mozilla/pdf.js/blob/17e9cfcd2a86bf6b72faf0ac3ac96fe4224731ba/src/display/annotation_layer.js#L976
[...]

PR #14008, opened 2021-09-12
JS - Don't use document.getElementsByName when updating some annotations

### LINKED ISSUE #14003: Hazard: annotation_layer.js uses document.getElementsByName without verification
`annotation_layer.js` has logic to keep input elements in sync. The logic uses `document.getElementsByName` with [a name derived from the PDF](https://github.com/mozilla/pdf.js/blob/10a1db698047e612e8ae3da66d12a6c1ac5597a3/src/core/annotation.js#L407) to retrieve the set of related elements. This query is not namespaced, so if the viewer ever adds a `name` attribute to a checkbox or radio input in the page, then the PDF file can get it changed (without user interaction via the scripting interface). Fortunately, there doesn't appear to be any input/radio element in the default viewer code. A custom viewer can be affected by this bug, however.

I suggest to replace the `document.getElementsByTagName` calls with a function that checks that the elements are created by the `annotation_layer.js` code before returning them. There are several ways to do so, e.g. by adding a custom attribute to each input element.

## Seemingly generic property assignment
This code on its own looks dangerous because it allows the caller to overwrite any property
https://github.com/mozilla/pdf.js/blob/99e442941c16561e8826a925b164d79f4ad51fbf/src/display/annotation_layer.js#L684-L688
Fortunately, it is only called with a fixed key, so only the `value` property can be updated in response to user input events:
https://github.com/mozilla/pdf.js/blob/99e442941c16561e8826a925b164d79f4ad51fbf/src/display/annotation_layer.js#L732-L740

## Unchecking other checkboxes on checkbox state change
https://github.com/mozilla/pdf.js/blob/99e442941c16561e8826a925b164d79f4ad51fbf/src/display/annotation_layer.js#L979-L990

## Updating the abstract state (storage)
https://github.com/mozilla/pdf.js/blob/99e442941c16561e8826a925b164d79f4ad51fbf/src/display/annotation_layer.js#L1061-L1065

## Allowing scripts to uncheck checkbox/radio inputs
The others require user interaction to have any effect; this one allows scripts embedded in the PDF to run logic without user interaction.
https://github.com/mozilla/pdf.js/blob/99e442941c16561e8826a925b164d79f4ad51fbf/src/display/annotation_layer.js#L1073-L1077

calixteman:
- it aims to fix #14003.
  - in a custom viewer, the use of document.getElementsByName could lead to get some elements which don't correspond to an annotation, so this patch removes it and annotations with the same fieldName are collected when they're created.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -160,6 +160,10 @@ class AnnotationElement {
     this.enableScripting = parameters.enableScripting;
     this.hasJSActions = parameters.hasJSActions;
     this._mouseState = parameters.mouseState;
+    this._annotationsWithSameName = parameters.annotationsWithSameName;
+    this._htmlElement = null;
+
+    this._annotationsWithSameName.push(this);
 
     if (isRenderable) {
       this.container = this._createContainer(ignoreBorder);
@@ -681,18 +685,6 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
     super(parameters, { isRenderable });
   }
 
-  setPropertyOnSiblings(base, key, value, keyInStorage) {
-    const storage = this.annotationStorage;
-    for (const element of document.getElementsByName(base.name)) {
-      if (element !== base) {
-        element[key] = value;
-        const data = Object.create(null);
-        data[keyInStorage] = value;
-        storage.setValue(element.getAttribute("id"), data);
-      }
-    }
-  }
-
   render() {
     const storage = this.annotationStorage;
     const id = this.data.id;
@@ -724,19 +716,23 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         element.type = "text";
         element.setAttribute("value", textContent);
       }
+      this._htmlElement = element;
       element.tabIndex = DEFAULT_TAB_INDEX;
 
       elementData.userValue = textContent;
       element.setAttribute("id", id);
 
       element.addEventListener("input", event => {
-        storage.setValue(id, { value: event.target.value });
-        this.setPropertyOnSiblings(
-          element,
-          "value",
-          event.target.value,
-          "value"
-        );
+        const value = event.target.value;
+        storage.setValue(id, { value });
+        for (const other of this._annotationsWithSameName) {
+          const htmlElement = other._htmlElement;
+          if (htmlElement && htmlElement !== element) {
+            const data = Object.create(null);
+            htmlElement.value = data.value = value;
+            storage.setValue(htmlElement.getAttribute("id"), data);
+          }
+        }
       });
 
       let blurListener = event => {
@@ -973,7 +969,7 @@ class CheckboxWidgetAnnotationElement extends WidgetAnnotationElement {
 
     this.container.className = "buttonWidgetAnnotation checkBox";
 
-    const element = document.createElement("input");
+    const element = (this._htmlElement = document.createElement("input"));
     element.disabled = data.readOnly;
     element.type = "checkbox";
     element.name = this.data.fieldName;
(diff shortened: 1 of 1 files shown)


src/display/annotation_layer.js
```diff
@@ -2049,7 +2049,8 @@ class AnnotationLayer {
    */
   static render(parameters) {
     const sortedAnnotations = [],
-      popupAnnotations = [];
+      popupAnnotations = [],
+      sameName = new Map();
```
Snuffleupagus, line 2053: Could annotations with the same name ever be found on *different* pages, because if so I'm not sure that this will work (since this is limited to just the current page)?

calixteman: That's correct.

calixteman: And I'd say that it was already an issue (I mean before this patch).

Snuffleupagus: My (possibly very stupid) idea when I initially saw the issue was that you maybe track the created `input` elements in e.g. a `WeakSet` defined "globally" in this file, and then just check the elements that we lookup with `document.getElementsByName`  against that.

Based on the following quote from https://github.com/mozilla/pdf.js/issues/14003#issue-993960056, Rob probably has better ideas though:
> I suggest to replace the document.getElementsByTagName calls with a function that checks that the elements are created by the annotation_layer.js code before returning them. There are several ways to do so, e.g. by adding a custom attribute to each input element.

calixteman: It won't avoid the problem with fields with same name across different pages: `document.getElementsByName` will get only the visible elements.

Snuffleupagus: > And I'd say that it was already an issue (I mean before this patch).

I see what you're saying, but it seems to me (I may be wrong of course) that it'd be slightly worse since before you got the Annotations on all active pages whereas with this patch it's *only* the current page.

---

Anyway, it seems that we might need to expand the `AnnotationStorage` to also have a `name -> id` mapping such that the initial value gets set correctly when rendering new pages? If so, that probably shouldn't be too hard to implement (famous last words).

cetinsert:
@calixteman – while eliminating `document.getElementsByName`, can we please consider 

- making the situation at least not worse for shadow DOM (https://github.com/mozilla/pdf.js/issues/13405)
- or even better coming up with a solution that would work equally well for both light DOM and shadow DOM?

---

@Snuffleupagus pre-defense text

> making the situation at least not worse for shadow DOM (https://github.com/mozilla/pdf.js/issues/13405)

I *know* the use cases of Mozilla (with the viewer) but PDF.js as a library is very capable and generic (and able to power a lot more than just the whole document viewer it is bundled with) and `document.getElementsByName` is the only area in my usage that is not compatible with proper per-page encapsulation because of how it breaks form updates in such a setting.

Snuffleupagus:
Re https://github.com/mozilla/pdf.js/pull/14008#issuecomment-917891009, how about *not* moving the "goalpost" here since this PR is specifically about fixing issue #14003 and not something else (e.g. shadow DOM).

src/display/annotation_layer.js
Snuffleupagus: It quite ugly, and possibly not correct in all cases, but the following patch is something that I just threw together quickly in order to try tracking both the `id` and `name` in the AnnotationStorage: https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:AnnotationStorage-name

Snuffleupagus:
Re https://github.com/mozilla/pdf.js/pull/14008#issuecomment-917891009, please let's *not* move the "goalpost" here since this PR is specifically about fixing issue #14003 and not something else (e.g. shadow DOM)?

src/display/annotation_layer.js
Rob--W: When I filed the issue I was thinking of something simple like using a `WeakSet`, tagging the element with a custom attribute, or even just prepending a prefix to the name to have something unique that can't overlap with existing ones. This draft patch looks more complex than that - why?

Snuffleupagus: @calixteman Sorry if I'm massively overstepping here, but I've got a (possibly stupid) idea.

What if we utilized the data returned by the `getFieldObjects`-method in the API, since then we'd be able to both access the fields on all active pages *and* also update the AnnotationStorage for not-yet-rendered pages?

https://github.com/mozilla/pdf.js/compare/master...Snuffleupagus:AnnotationLayer-getElementsByName

calixteman: @Snuffleupagus, no, you aren't and I think that's a very good idea.
This way we don't have to change `AnnotationStorage` implementation because we'll have all the ids for fields with the same name. Would you mind to write the patch since you've almost thought about everything here ? If yes, please close this one.

@Rob--W, my idea was to remove the call to `getElementsByName` and just update the fields with the same name. But as pointed out by Jonas, it was a bad idea and in the meantime it was not so bad because it helped us to find a bug.

Snuffleupagus:
Closing in favour of PR #14023.

PR #14013, opened 2021-09-13
Improve the API unit-tests, and try to expose more API-functionality in the TypeScript definitions

Snuffleupagus:

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -3361,6 +3361,7 @@ export {
   getDocument,
   LoopbackPort,
   PDFDataRangeTransport,
+  PDFDocumentLoadingTask,
   PDFDocumentProxy,
   PDFPageProxy,
   PDFWorker,
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -3365,6 +3365,7 @@ export {
   PDFDocumentProxy,
   PDFPageProxy,
   PDFWorker,
+  RenderTask,
   setPDFNetworkStreamFactory,
   version,
 };
--- a/src/pdf.js
+++ b/src/pdf.js
@@ -14,6 +14,12 @@
  */
 /* eslint-disable sort-exports/sort-exports */
 
+// eslint-disable-next-line max-len
+/** @typedef {import("./display/api").PDFDocumentLoadingTask} PDFDocumentLoadingTask */
+/** @typedef {import("./display/api").PDFDocumentProxy} PDFDocumentProxy */
+/** @typedef {import("./display/api").PDFPageProxy} PDFPageProxy */
+/** @typedef {import("./display/api").RenderTask} RenderTask */
+
 import {
   addLinkAttributes,
   getFilenameFromUrl,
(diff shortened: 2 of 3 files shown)


timvandermeij:
Given https://github.com/mozilla/pdf.js/pull/13957#issuecomment-918603941, does the last commit make sense to include at this point in time?

Snuffleupagus:
> Given [#13957 (comment)](https://github.com/mozilla/pdf.js/pull/13957#issuecomment-918603941), does the last commit make sense to include at this point in time?

I really don't know, since we also have https://github.com/mozilla/pdf.js/issues/12384#issuecomment-695143287 which explicitly recommended this approach.

timvandermeij:
@tamuratak Would you perhaps be able to shed some light on if the third commit in this PR, https://github.com/Snuffleupagus/pdf.js/commit/95057a4e56cd04ecf4248afcc7e82a303d36c5bb to be precise, is working correctly given your suggestion from https://github.com/mozilla/pdf.js/issues/12384#issuecomment-695143287 one the one hand but https://github.com/mozilla/pdf.js/pull/13957#issuecomment-918603941 on the other hand that claims it doesn't have the desired effect?

tamuratak:
The following line does not work as expected.

https://github.com/mozilla/pdf.js/blob/95057a4e56cd04ecf4248afcc7e82a303d36c5bb/src/pdf.js#L21

That is because ` RenderTask` is not exported in `src/display/api.js`. `PDFDocumentProxy` and `PDFPageProxy` are exported in the file.

https://github.com/mozilla/pdf.js/blob/95435ed66f18cd613af45f138eb1b54110a282f2/src/display/api.js#L3140

For #13957, if we remove the `RenderTask` type annotation in the sample code, it works with the type inference of TypeScript.

``` ts
const task: RenderTask = pageProxy.render(renderContext);
```

``` ts
const task = pageProxy.render(renderContext);
```

スクリーンショット 2021-09-16 10 15 54

Snuffleupagus:
> That is because ` RenderTask` is not exported in `src/display/api.js`.

Huh, but it actually is thanks to the previous commit; please see https://github.com/mozilla/pdf.js/pull/14013/files#diff-082d6b37ad01db7ac97cc07c6ddb0dc52040484c5ef91b110b072f50144d9f39

Furthermore, we now have unit-tests that ensure that `RenderTask` is being exported from that file.
@tamuratak Did you test the *entire* PR, or only its last commit?

tamuratak:
> Furthermore, we now have unit-tests that ensure that RenderTask is being exported from that file.
> @tamuratak Did you test the entire PR, or only its last commit?

I am sorry I have missed that. I have confirmed that things work as expected.

Snuffleupagus:
> I am sorry I have missed that. I have confirmed that things work as expected.

No worries; thank you for helping out with testing this!

timvandermeij (approved):

timvandermeij:
Let's do this now that we know it's working; thanks!

pardoman:

PR #14018, opened 2021-09-13
XFA - Don't create images for unsupported mime types

calixteman:


brendandahl:
I didn't look at all the failures, but several of them have blank content types. In those cases to the files do have gif/jpg file extensions, but I wonder if it'd be better to look for a magic number.  A different bug is xfa_issue13633-page1, it has content type of "image/PNG", so we'll probably want to lowercase the string comparison.

PR #14023, opened 2021-09-14
Re-factor `document.getElementsByName` lookups in the AnnotationLayer (issue 14003)

### LINKED ISSUE #14003: Hazard: annotation_layer.js uses document.getElementsByName without verification

Snuffleupagus:
This replaces direct `document.getElementsByName` lookups with a helper method which:
 - Lets the AnnotationLayer use the data returned by the `PDFDocumentProxy.getFieldObjects` API-method, such that we can directly lookup only the necessary DOM elements.
 - Fallback to using `document.getElementsByName` as before, such that e.g. the standalone viewer components still work.

Finally, to fix the problems reported in issue 14003, regardless of the code-path we now also enforce that the DOM elements found were actually created by the AnnotationLayer code.
With these changes we'll thus be able to update form elements on all visible pages just as before, but we'll additionally update the AnnotationStorage for not-yet-rendered elements thus fixing a pre-existing bug.

Fixes #14003


src/display/annotation_layer.js
```diff
@@ -159,6 +161,7 @@ class AnnotationElement {
     this.annotationStorage = parameters.annotationStorage;
     this.enableScripting = parameters.enableScripting;
     this.hasJSActions = parameters.hasJSActions;
+    this._fieldObjects = parameters.fieldObjects;
```
calixteman, line 164: nit: I think that it'd be slightly better to have a `Map` `name => [ids....]` generated from `parameters.fieldObjects` where fields with `page === -1` are removed this way we don't have to check in `_getElementsByName`. Since we needn't other properties than id, it should help to slightly reduce memory use.

calixteman (commented):
Could you add an integration test please ?
Anyway it looks good overall.

src/display/annotation_layer.js
Snuffleupagus: > [...] with `page === -1` are removed this way we don't have to check in `_getElementsByName`.

You're correct that this would slightly reduce the necessary parsing here, however...

> it should help to slightly reduce memory use.

That would actually *increase* overall memory usage, since it'd force us to allocate one `Map` per page whereas currently every page simply gets *the same* Object (since the API implements caching).

---

Do you have any pointers for the integration-test, e.g. more specifically what you'd like to see tested?
(I've not written any integration-tests before, so feeling slightly out of my depth here.)

calixteman: > > [...] with `page === -1` are removed this way we don't have to check in `_getElementsByName`.
> 
> You're correct that this would slightly reduce the necessary parsing here, however...
> 
> > it should help to slightly reduce memory use.
> 
> That would actually _increase_ overall memory usage, since it'd force us to allocate one `Map` per page whereas currently every page simply gets _the same_ Object (since the API implements caching).
> 
I always forget that there is a layer per page...

> Do you have any pointers for the integration-test, e.g. more specifically what you'd like to see tested?
> (I've not written any integration-tests before, so feeling slightly out of my depth here.)
I'd imagine a pdf with on page 1: a text field, a radio and a checkbox and the same on page 2 and page 4 (or more if needed) with the same fields with the same name.
Fill elements on page 1 then scroll on page 2, check that the values are correct (they should have been updated thanks to document.get...) and the same with page 4 (updated through annotationStorage).

This test could help (after reading it, I fixed the current bug but only when the field is updated from JS... :( )
https://github.com/mozilla/pdf.js/blob/master/test/integration/scripting_spec.js#L730
The line here:
https://github.com/mozilla/pdf.js/blob/master/test/integration/scripting_spec.js#L723
open the given pdf and the form is ready when the element with the given id is displayed.
and here you've a way to scroll to the page 4:
https://github.com/mozilla/pdf.js/blob/master/test/integration/scripting_spec.js#L770

calixteman: Maybe we could create the simplified fieldObjects before to pass it to the layer.

calixteman: @Snuffleupagus, if you don't have time to write a test or don't feel like to do it, tell me and I'll write it later today or tomorrow.

Snuffleupagus: > Maybe we could create the simplified fieldObjects before to pass it to the layer.

We could, but I don't want to do that for two reasons:
 - It would make the viewer components more difficult to use standalone, since it'd then not be enough just to call the correct API-method but you'd also need to *manually* transform the data into the required format.
 - It's not inconceivable that we'll want to add some future `fieldObjects` usage in the AnnotationLayer, and if so we'd then need yet *another* external transformation or a breaking change of e.g. the `DefaultAnnotationLayerFactory` class.

All-in-all, I'm going to leave this as-is since the very slight overhead during parsing still ought to be lower than having to use `document.getElementsByName` unconditionally as in the current code.

---

> if you don't have time to write a test or don't feel like to do it, tell me and I'll write it later today or tomorrow.

I'm guessing that it would be *a lot quicker* for you to add a test, and I'd probably won't have time to dive into all of that until the weekend.
So if you don't mind writing the integration-test, I'd be very grateful for your help :-)

calixteman: @Snuffleupagus I'll add a test tomorrow.

src/display/annotation_layer.js
```diff
@@ -1062,26 +1108,27 @@ class RadioButtonWidgetAnnotationElement extends WidgetAnnotationElement {
...
+      const self = this;
       element.addEventListener("updatefromsandbox", jsEvent => {
         const actions = {
           value(event) {
             const checked = pdfButtonValue === event.detail.value;
-            for (const radio of document.getElementsByName(event.target.name)) {
-              const radioId = radio.getAttribute("id");
-              radio.checked = radioId === id && checked;
-              storage.setValue(radioId, { value: radio.checked });
+            for (const radio of self._getElementsByName(event.target.name)) {
```
Rob--W, line 1126: Replace `value(event) {` with `value: event => {` to avoid the need for capturing `self = this`.

src/display/annotation_layer.js
```diff
@@ -363,6 +366,50 @@ class AnnotationElement {
...
   }
 
+  /**
+   * @private
+   * @returns {Array}
+   */
+  _getElementsByName(name, skipId = null) {
+    const fields = [];
+
+    if (this._fieldObjects) {
```
Rob--W, line 376: Why both `_fieldObjects` and DOM-based lookup?

Rob--W (commented):
I started looking but haven't had the chance to take a very close look. A question follows below + small nit about use of `self`.

src/display/annotation_layer.js
Snuffleupagus: As mentioned in https://github.com/mozilla/pdf.js/pull/14023#issue-995915957, that's to ensure that the standalone viewer components won't break.
See https://github.com/mozilla/pdf.js/tree/master/examples/components, and in particular the `pageviewer` example, which uses a simplified AnnotationLayer-initialization without all of the parameters that the default viewer will provide.

calixteman:
@Snuffleupagus here is an integration test: https://github.com/calixteman/pdf.js/commit/2d551ea62e147841679cab821241ed6b32363dbf

Snuffleupagus:
> @Snuffleupagus here is an integration test: [calixteman@2d551ea](https://github.com/calixteman/pdf.js/commit/2d551ea62e147841679cab821241ed6b32363dbf)

Thank you, thank looks fantastic!

calixteman (approved):
LGTM, thank you for doing that.

Snuffleupagus:
@Rob--W Review ping; did you want to take a closer look specifically at the `GetElementsByNameSet`-part of this patch, since that's most relevant to issue #14003?

The rest of the patch fixes an existing short-comings of the implementation, with it not being able to target elements on not-yet-rendered pages, and have been reviewed by Calixte.

src/display/annotation_layer.js
```diff
@@ -728,6 +777,7 @@ class TextWidgetAnnotationElement extends WidgetAnnotationElement {
         element.type = "text";
         element.setAttribute("value", textContent);
       }
+      GetElementsByNameSet.add(element);
```
Rob--W, line 780: Could the `name` assignment be moved from below to this location, so that it is more obvious to which name this element is associated? That would also be more consistent with the other 3 initializations of `GetElementsByNameSet`.
https://github.com/mozilla/pdf.js/blob/6040ae32819c454d8e1724c9d298e94b64e76f67/src/display/annotation_layer.js#L954

web/annotation_layer_builder.js
```diff
@@ -33,6 +33,7 @@ import { SimpleLinkService } from "./pdf_link_service.js";
  * @property {IL10n} l10n - Localization service.
  * @property {boolean} [enableScripting]
  * @property {Promise<boolean>} [hasJSActionsPromise]
+ * @property {Promise<Array<Object> | null} [fieldObjectsPromise]
```
Rob--W, line 36: The type is missing a `>` after `null` (`Promise<` is not closed). Not just here, but everywhere else in this patch.

`fieldObjects` is an object, not an array of objects. A type of `object` is rather unclear. Is it possible to improve the documentation of the public API, so that it is more obvious what its value should be? Even if it just states something like "The document's fieldObjects" at [
https://github.com/mozilla/pdf.js/blob/680f33c31c23c90d01065695e9b278748ad8ccfc/src/core/document.js#L1407-L1435](https://github.com/mozilla/pdf.js/blob/680f33c31c23c90d01065695e9b278748ad8ccfc/src/core/document.js#L1407-L1435)

Rob--W (approved):
r+ with nit. Feel free to merge after addressing the issue, but if you'd like another pair of eyes after making the changes just mention me.

web/annotation_layer_builder.js
Snuffleupagus: > The type is missing a `>` after `null` (`Promise<` is not closed). Not just here, but everywhere else in this patch.

Good eye, that was a copy-and-paste error on my part.

*Regarding the pre-existing type issue:*
I've tried to improve the type in the API to outline the data format a bit better, and copied that across the patch, however I'm not convinced that also linking to the worker-thread code helps all that much (you'd probably need to look at the various `getFieldObject`-methods in `src/core/annotation.js` too).

---

Thanks for the review!

Issue #14015, opened 2021-09-13
PDF.js: Problems with German umlaut “ß” and “ü”


BerndS2020:
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version: Firefox 92.0 (64 bit)
- Operating system and its version: Windows 10
- PDF.js version: 2.11.91
- Is a browser extension: ?

Steps to reproduce the problem:
1. See chapter "What went wrong"

What is the expected behavior? (add screenshot)
Correct display of german umlaute "ü" and "ß"

What went wrong? (add screenshot)
Wrong display of german umlaute "ü" and "ß"
![grafik](https://user-images.githubusercontent.com/90627791/133097571-fd56252b-e23b-4152-9b00-0075968b56d6.png)

Our use case:
1. We create a PDF from our software. We use the font **ArialMT** for our PDF document.
2. The PDF is displayed correctly in PDF viewer tools (Adobe, Sumatra, FreePDF, …) 
3. The only tool in which the PDF is not displayed correctly is the built-in PDF viewer in Firefox.
4. It effects the characters “ß”, “ü”
5. German umlaute “ö” & "ä" works instead.

We have already analyzed the behavior and were able to determine that the internal character code in your PDF is (Identity-H). Could this encoding be the problem?
![grafik](https://user-images.githubusercontent.com/90627791/133097764-c00bc1d3-d684-46b8-af1e-059cb2d473c7.png)

Here is our understanding:
The 'Identity-H' encoding in a PDF means that the bytes in the string specify Glyph IDs: values that identify specific glyphs in a TrueType or OpenType font file. These Glyph IDs are not the same from font to font (e.g. 'A' may map to glyph ID 12 in a Helvetica font file, but to glyph ID 37 in Comic Sans). Because the string directly depends on a specific font, an 'Identity-H' encoding requires that the font file be embedded or subset within the PDF.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension):

Snuffleupagus:
> Attach (recommended) or Link to PDF file here:

Sorry, but this part of the issue template is *required* in order for this (or any other issue) to be actionable!

---

Generally speaking, the only way to *guarantee* that fonts always render correctly is by embedding them into the PDF document.

Snuffleupagus:
Closing as INCOMPLETE for now, we can re-open this is all of the necessary information is provided (see above).

BerndS2020:
[...]

PR #14025, opened 2021-09-15
Improve glyph mapping for non-embedded composite standard fonts with a /CIDToGIDMap (issue 11915)

### LINKED ISSUE #11915: Garbled Display of Fonts
Viewer:  https://mozilla.github.io/pdf.js/web/viewer.html
Chrome on Windows 10
When the attached PDF file is viewed, some of the fonts display as garbled text. 
[Arial, Courier, Times New Roman]
The file views correctly in Adobe Reader.
[fff.pdf](https://github.com/mozilla/pdf.js/files/4652174/fff.pdf)

Snuffleupagus:
*Please note:* All of this feels very handwavy, but at least it passes all tests locally. Hopefully we have enough tests for this part of the font code.

For non-embedded composite standard fonts with an "incomplete" /CIDToGIDMap, we'll now fallback to an *explicitly defined* /ToUnicode map even when that one happens to be an /Identity-H or /Identity-V map.

The `Font.fallbackToSystemFont` method is unfortunately getting more and more special-cases, however that might be unavoidable given all the weird non-embedded fonts found in the wild :-(

Fixes #11915


brendandahl:
Does this fix #14015 too?

Snuffleupagus:
> Does this fix #14015 too?

Yes, it does.

PR #14027, opened 2021-09-15
Annotation - Checkboxes with the same name and export values must be in unison

### LINKED ISSUE #14024: Wrong behaviour with checkboxes sharing a common name
I created a pdf for https://github.com/mozilla/pdf.js/pull/14023:
[same_fields(1).pdf](https://github.com/mozilla/pdf.js/files/7168340/same_fields.1.pdf)
The 3 checkboxes have the same name (`Check Box2`) and have the same values (`Off` and `Yes`).
In pdf.js, these checkboxes behave like radio buttons in the same group (which is wrong) because of:
https://github.com/mozilla/pdf.js/commit/1fcffe80349a98378262c8a16f7b50ff34c3fe82
Since they've the same values they must be synced: if one is checked (resp. unchecked) then all others are checked (resp. unchecked).

calixteman:
- it aims to fix #14024.
  - this patch adds an attribute `acroformExportValue` to the HTML input in order to set the checked attribute in taking into account the exportValue for the checkboxes with the same name.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -976,25 +976,29 @@ class CheckboxWidgetAnnotationElement extends WidgetAnnotationElement {
     const element = document.createElement("input");
     element.disabled = data.readOnly;
     element.type = "checkbox";
-    element.name = this.data.fieldName;
+    element.name = data.fieldName;
     if (value) {
       element.setAttribute("checked", true);
     }
     element.setAttribute("id", id);
+    element.setAttribute("exportValue", data.exportValue);
     element.tabIndex = DEFAULT_TAB_INDEX;
 
     element.addEventListener("change", function (event) {
       const name = event.target.name;
+      const checked = event.target.checked;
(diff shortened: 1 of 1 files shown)


calixteman:
The test I'll write for #14023 will test the current patch.

src/display/annotation_layer.js
```diff
@@ -976,25 +976,29 @@ class CheckboxWidgetAnnotationElement extends WidgetAnnotationElement {
     const element = document.createElement("input");
     element.disabled = data.readOnly;
     element.type = "checkbox";
-    element.name = this.data.fieldName;
+    element.name = data.fieldName;
     if (value) {
       element.setAttribute("checked", true);
     }
     element.setAttribute("id", id);
+    element.setAttribute("acroformExportValue", data.exportValue);
```
Snuffleupagus, line 984: The attribute name feels a bit verbose to me; any reason that just `exportValue` wouldn't work?

Snuffleupagus (approved):

PR #14028, opened 2021-09-15
Use CSS variables for setting the width of the zoom dropdown (PR 11570 follow-up)

Snuffleupagus:
By using CSS variables to set the width of the zoom dropdown, we can simplify both the relevant CSS and JS code. This will not only improve overall maintainability of this code, but should also make it (slightly) easier for third-party users that want to customize the width.

Note in particular that by having the code in `Toolbar._adjustScaleWidth` lookup the values of the CSS variables, we no longer need to worry about keeping hard-coded values up-to-date with the CSS rules.

### DIFF
--- a/web/toolbar.js
+++ b/web/toolbar.js
@@ -23,9 +23,6 @@ import {
 } from "./ui_utils.js";
 
 const PAGE_NUMBER_LOADING_INDICATOR = "visiblePageIsLoading";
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Much better like this; thank you for doing this!

PR #14030, opened 2021-09-15
Update systemjs.config.js

Chandra-Sekhar-Bala:


Snuffleupagus:
Thanks for the PR, however we unfortunately cannot accept this one either.

Before submitting a patch, please make sure that you actually test it; see https://github.com/mozilla/pdf.js/wiki/Contributing
In this case, running `gulp lint` locally would've shown that these changes are breaking the accepted coding-style as enforced by ESLint/Prettier, see also https://github.com/mozilla/pdf.js/wiki/Style-Guide

PR #14035, opened 2021-09-17
Annotation - For checkboxes, get field value from AS (if any) instead of V (bug 1722036)

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1722036.
  - AS and V should share the same value for checkbox: it's at least what the specs say;
  - the pdf in the above bug opens correctly in Acrobat so it likely means that AS is chosen over V.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -2083,6 +2083,13 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       return;
     }
 
+    // See https://bugzilla.mozilla.org/show_bug.cgi?id=1722036.
+    // If we've an AS and a V then take AS.
+    const asValue = this._decodeFormValue(params.dict.get("AS"));
+    if (typeof asValue === "string") {
+      this.data.fieldValue = asValue;
+    }
+
     const exportValues = normalAppearance.getKeys();
     if (!exportValues.includes("Off")) {
       // The /Off appearance is optional.
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -5899,5 +5899,13 @@
        "rounds": 1,
        "lastPage": 1,
        "type": "eq"
+    },
+    {  "id": "bug1722036",
+       "file": "pdfs/bug1722036.pdf",
+       "md5": "bbc7f62ca85609c97069bd3ce2658288",
+       "rounds": 1,
+       "link": true,
+       "forms": true,
+       "type": "eq"
     }
 ]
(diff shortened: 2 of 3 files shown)


src/core/annotation.js
```diff
@@ -2083,6 +2083,12 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       return;
     }
 
+    if (this.appearance) {
+      // See https://bugzilla.mozilla.org/show_bug.cgi?id=1722036
+      // If we've an AS and V then take AS.
+      this.data.fieldValue = this._decodeFormValue(params.dict.get("AS"));
```
Snuffleupagus, line 2089: I'd suggest reducing any *possible* regression risk here, given all the weird things that PDF generators does, by checking that the /AS value makes sense before using it; e.g. something along these lines:
```javascript
      const asValue = this._decodeFormValue(params.dict.get("AS"));
      if (typeof asValue === "string") {
        this.data.fieldValue = asValue;
      }
```

Snuffleupagus (approved):
r=me, with the comments addressed and all tests passing; thank you!

test/test_manifest.json
```diff
@@ -5899,5 +5899,12 @@
        "rounds": 1,
        "lastPage": 1,
        "type": "eq"
+    },
+    {  "id": "bug1722036",
+       "file": "pdfs/bug1722036.pdf",
+       "md5": "bbc7f62ca85609c97069bd3ce2658288",
+       "rounds": 1,
+        "link": true,
```
Snuffleupagus, line 5907: Nit: Too much indentation here.

Snuffleupagus: Also, isn't this test-case passing already without the patch!?

I think you want to add `"forms": true` as well here, right?

calixteman:
It seems that we had a font rendering issue in last makeref:
see firefox-annotation-line-page1 for example: the space between `e` and `i` is a bit weird.

Snuffleupagus:
> It seems that we had a font rendering issue in last makeref:

I don't think it's our "fault", since this looks like (yet another) case where some font rendering changes were made in Firefox itself.
Looking at the pushlog, this was possibly caused by https://bugzilla.mozilla.org/show_bug.cgi?id=1730772 which has since been backed out. I was also seeing bad font rendering locally in Nightly, but with the latest update (< 5 minutes ago) it looks good again.

As far as I'm concerned, I think you can ignore those "regressions" and simply land this PR.

PR #14036, opened 2021-09-17
Annotation - Some checkboxes have an empty N dictionary

### LINKED ISSUE #14021: Checkbox is not checked when it must be
[docOficialPDF.php.pdf](https://github.com/mozilla/pdf.js/files/7159923/docOficialPDF.php.pdf)

Configuration:
- PDF.js version: built-in viewer
- Is a browser extension: no

1. Open the pdf page 2
2. Checkbox "Alta" in section 2 must be checked

It works fine with evince.

calixteman:
- it aims to fix #14021;
  - the N dict is empty here so just create a default one;
  - it implies that the checked checkbox has no appearance so create a default one too in order to print it.
  - in the pdf in the issue, a checked box is not printed because it has no default appearance so we need to guess its appearance from its state.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -1906,7 +1906,7 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     }
   }
 
-  getOperatorList(evaluator, task, renderForms, annotationStorage) {
+  async getOperatorList(evaluator, task, renderForms, annotationStorage) {
     if (this.data.pushButton) {
       return super.getOperatorList(
         evaluator,
@@ -1916,10 +1916,16 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       );
     }
 
+    let value = null;
     if (annotationStorage) {
       const storageEntry = annotationStorage.get(this.data.id);
-      const value = storageEntry && storageEntry.value;
-      if (value === undefined) {
+      value = storageEntry ? storageEntry.value : null;
+    }
+
+    if (value === null) {
+      // Nothing in the annotationStorage.
+      if (this.appearance) {
+        // But we've a default appearance so use it.
         return super.getOperatorList(
           evaluator,
           task,
@@ -1928,35 +1934,33 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
         );
       }
 
-      let appearance;
-      if (value) {
-        appearance = this.checkedAppearance;
+      // There is no default appearance so use the one derived
+      // from the field value.
+      if (this.data.checkBox) {
+        value = this.data.fieldValue === this.data.exportValue;
       } else {
-        appearance = this.uncheckedAppearance;
+        value = this.data.fieldValue === this.data.buttonValue;
       }
-
-      if (appearance) {
-        const savedAppearance = this.appearance;
-        this.appearance = appearance;
-        const operatorList = super.getOperatorList(
-          evaluator,
-          task,
-          renderForms,
-          annotationStorage
-        );
-        this.appearance = savedAppearance;
-        return operatorList;
-      }
-
-      // No appearance
-      return Promise.resolve(new OperatorList());
     }
-    return super.getOperatorList(
-      evaluator,
-      task,
-      renderForms,
-      annotationStorage
-    );
+
+    const appearance = value
+      ? this.checkedAppearance
+      : this.uncheckedAppearance;
+    if (appearance) {
+      const savedAppearance = this.appearance;
+      this.appearance = appearance;
+      const operatorList = super.getOperatorList(
+        evaluator,
+        task,
+        renderForms,
+        annotationStorage
+      );
+      this.appearance = savedAppearance;
+      return operatorList;
+    }
+
+    // No appearance
+    return new OperatorList();
   }
 
   async save(evaluator, task, annotationStorage) {
@@ -1982,7 +1986,7 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       return null;
     }
 
-    const defaultValue = this.data.fieldValue && this.data.fieldValue !== "Off";
+    const defaultValue = this.data.fieldValue === this.data.exportValue;
     if (defaultValue === value) {
       return null;
     }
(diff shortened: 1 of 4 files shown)


src/core/annotation.js
```diff
@@ -1895,47 +1895,45 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       );
     }
 
+    // The annotationStorage can be null if no change occured on the form
+    // but a checkbox could be already checked when it's open.
```
Snuffleupagus, line 1899: Sorry, but I'm finding this second line of the sentence slightly difficult to understand. Could it perhaps be re-worded to improve the clarity somewhat?

src/core/annotation.js
```diff
@@ -1895,47 +1895,45 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
-        );
-        this.appearance = savedAppearance;
-        return operatorList;
-      }
+    let appearance;
+    if (value) {
+      appearance = this.checkedAppearance;
+    } else {
+      appearance = this.uncheckedAppearance;
+    }
```
Snuffleupagus, line 1920: I see that this is pre-existing code, but how about simplifying it a bit when you're already touching it?
```javascript
const appearance = value ? this.checkedAppearance : this.uncheckedAppearance;
```

src/core/annotation.js
```diff
@@ -2147,10 +2194,15 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
-    this.checkedAppearance = normalAppearance.get(this.data.buttonValue);
+    this.checkedAppearance =
+      normalAppearance.get(this.data.buttonValue) || null;
     this.uncheckedAppearance = normalAppearance.get("Off") || null;
 
-    this._streams.push(this.checkedAppearance);
+    if (this.checkedAppearance) {
+      this._streams.push(this.checkedAppearance);
+    } else {
+      this._getDefaultCheckedAppearance(params);
```
Snuffleupagus, line 2204: Given that this is a *radio* button, is it actually correct to use same appearance as for the checkbox-case?

test/test_manifest.json
```diff
@@ -5907,5 +5907,14 @@
...
        "forms": true,
        "type": "eq"
+    },
+    {  "id": "issue14021",
+       "file": "pdfs/issue14021.pdf",
+       "md5": "d18aa84135ce985c70a8f56306ecb95f",
+       "link": true,
+       "form": true,
+       "rounds": 1,
+       "lastPage": 1,
```
Snuffleupagus, line 5917: As far as I can tell, there's actually no form-elements on that page.
I'm guessing that you actually want `firstPage: 2` and `lastPage: 2` here?

test/test_manifest.json
```diff
@@ -5907,5 +5907,14 @@
        "link": true,
        "forms": true,
        "type": "eq"
+    },
+    {  "id": "issue14021",
+       "file": "pdfs/issue14021.pdf",
+       "md5": "d18aa84135ce985c70a8f56306ecb95f",
+       "link": true,
+       "form": true,
```
Snuffleupagus, line 5915: This should be `"forms": true` instead to work as intented; also can you please move this down to below the `type`-line instead since that helps to quickly see exactly what sort of test it is?

Snuffleupagus (commented):
I've left a couple of comments based on a *very quick* look, but I'll need some more time to review this properly (will so during the weekend).

calixteman:
The regression for `issue12294` isn't a regression but an improvement !
I compared preview in firefox and in chrome and we've the same.

src/core/annotation.js
```diff
@@ -2072,6 +2072,36 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
     return newRefs;
   }
 
+  _getDefaultCheckedAppearance(params) {
+    const width = this.data.rect[2] - this.data.rect[0];
+    const height = this.data.rect[3] - this.data.rect[1];
+    const fontSize = Math.min(width, height) * 0.9;
```
Snuffleupagus, line 2078: Where does this magical value come from, and is there any way to handle this without a hard-coded value?
If this is indeed necessary, can we perhaps at least replace it with a constant instead (and add a comment about how the value was chosen)?

src/core/annotation.js
Snuffleupagus: Sorry, but I'd still like an answer to this my previous question here (regarding if it's "correct" to use this appearance also for a *radio* button, since those often look different from checkboxes).

src/core/annotation.js
```diff
@@ -2090,27 +2120,46 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       this.data.fieldValue = asValue;
     }
 
+    let yes = "Yes";
+    if (this.data.fieldValue !== null && this.data.fieldValue !== "Off") {
+      yes = this.data.fieldValue;
+    }
```
Snuffleupagus, line 2126: Optional nit: Maybe ensure that the `yes`-value cannot accidentally be modified below by doing something like:
```javascript
    const yes =
      this.data.fieldValue !== null && this.data.fieldValue !== "Off"
        ? this.data.fieldValue
        : "Yes";
```

src/core/annotation.js
```diff
@@ -2072,6 +2078,63 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
+    let metrics;
+    let char;
+    if (type === "check") {
+      // Char 33 (2713 in unicode)
+      metrics = {
+        width: 0.755 * fontSize,
+        height: 0.705 * fontSize,
+      };
+      char = "\x33";
+    } else {
```
Snuffleupagus, line 2102: Can we please use `} else if (type === "disc") {` here instead, and then add an additional branch below to ensure that the caller explicitly chooses a valid type?
```javascript
} else {
  unreachable(`_getDefaultCheckedAppearance - unsupported type: ${type}`);
}
```

src/core/annotation.js
```diff
@@ -2072,6 +2078,63 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
+
+    // Ratio used to have a mark slightly smaller than the bbox.
+    const FONT_RATIO = 0.8;
+    const fontSize = Math.min(width, height) * FONT_RATIO;
+
+    // Char Metrics
+    // Widths came from widths for ZapfDingbats.
+    // Heights are guessed with Fontforge and FoxitDingbats.pfb.
+    let metrics;
+    let char;
```
Snuffleupagus, line 2094: Nit: Fits on one line, as `let metrics, char;`

test/test_manifest.json
```diff
@@ -5907,5 +5907,36 @@
...
+       "file": "pdfs/issue14021.pdf",
+       "md5": "d18aa84135ce985c70a8f56306ecb95f",
+       "link": true,
+       "rounds": 1,
+       "firstPage": 2,
+       "lastPage": 2,
+       "type": "eq",
+       "forms": true
+    },
+    {  "id": "issue14021-1",
```
Snuffleupagus, line 5921: Let's name it "issue14021-storage" instead, since that seems slightly more helpful.

src/core/annotation.js
```diff
@@ -2072,6 +2078,63 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
...
   }
 
+  _getDefaultCheckedAppearance(params, type) {
+    const width = this.data.rect[2] - this.data.rect[0];
+    const height = this.data.rect[3] - this.data.rect[1];
+    const bbox = [0, 0, width, height];
+
+    // Ratio used to have a mark slightly smaller than the bbox.
+    const FONT_RATIO = 0.8;
+    const fontSize = Math.min(width, height) * FONT_RATIO;
```
Snuffleupagus, line 2088: Thank you, this is helpful when reading the code :-)

Snuffleupagus (approved):
r=me, with the final comments addressed and all tests passing (or improving); thank you!

PR #14038, opened 2021-09-17
JS - Implement few possibilities with app.execMenuItem (bug 1724399)

calixteman:
Can be tested with https://github.com/mozilla/pdf.js/blob/master/test/pdfs/evaljs.pdf
Enter `app.execMenuItem("SaveAs")` in the first field and then click on the button `Execute`.

### DIFF
--- a/src/scripting_api/app.js
+++ b/src/scripting_api/app.js
@@ -479,8 +479,24 @@ class App extends PDFObject {
     /* Not implemented */
   }
 
-  execMenuItem() {
-    /* Not implemented */
+  execMenuItem(item) {
+    switch (item) {
+      case "SaveAs":
+      case "FirstPage":
+      case "LastPage":
+      case "NextPage":
+      case "PrevPage":
+      case "ZoomViewIn":
+      case "ZoomViewOut":
+        this._send({ command: item });
+        break;
+      case "FitPage":
+        this._send({ command: "zoom", value: "page-fit" });
+        break;
+      case "Print":
+        this._send({ command: "print" });
+        break;
+    }
   }
 
   getNthPlugInName() {
--- a/web/pdf_scripting_manager.js
+++ b/web/pdf_scripting_manager.js
@@ -305,6 +305,33 @@ class PDFScriptingManager {
           }
           this._pdfViewer.currentScaleValue = value;
           break;
+        case "SaveAs":
+          this._eventBus.dispatch("save", { source: this });
+          break;
+        case "FirstPage":
+          this._pdfViewer.currentPageNumber = 1;
+          break;
+        case "LastPage":
+          this._pdfViewer.currentPageNumber = this._pdfViewer.pagesCount;
+          break;
+        case "NextPage":
+          this._pdfViewer.nextPage();
+          break;
+        case "PrevPage":
+          this._pdfViewer.previousPage();
+          break;
+        case "ZoomViewIn":
+          if (isInPresentationMode) {
+            return;
+          }
+          this._eventBus.dispatch("zoomin", { source: this });
+          break;
+        case "ZoomViewOut":
+          if (isInPresentationMode) {
+            return;
+          }
+          this._eventBus.dispatch("zoomout", { source: this });
+          break;
       }
       return;
     }


web/pdf_scripting_manager.js
```diff
@@ -305,6 +305,27 @@ class PDFScriptingManager {
...
+          this._eventBus.dispatch("save", { source: this });
+          break;
+        case "FirstPage":
+          this._pdfViewer.currentPageNumber = 1;
+          break;
+        case "LastPage":
+          this._pdfViewer.currentPageNumber = this._pdfViewer.pagesCount;
+          break;
+        case "NextPage":
+          this._pdfViewer.currentPageNumber = Math.min(
```
Snuffleupagus, line 318: Please just use `this._pdfViewer.nextPage()` instead, since that correctly handles things such as e.g. scroll/spread modes.

web/pdf_scripting_manager.js
```diff
@@ -305,6 +305,27 @@ class PDFScriptingManager {
...
+            this._pdfViewer.pagesCount,
+            this._pdfViewer.currentPageNumber + 1
+          );
+          break;
+        case "PrevPage":
+          this._pdfViewer.currentPageNumber = Math.max(
```
Snuffleupagus, line 324: As above, `this._pdfViewer.previousPage()` instead.

marco-c (approved):

web/pdf_scripting_manager.js
```diff
@@ -305,6 +305,27 @@ class PDFScriptingManager {
           }
           this._pdfViewer.currentScaleValue = value;
           break;
+        case "SaveAs":
+          this._eventBus.dispatch("save", { source: this });
+          break;
+        case "FirstPage":
+          this._eventBus.dispatch("firstpage", { source: this });
```
Snuffleupagus, line 312: Events should only be used for things that the `BaseViewer` doesn't *directly* support, and which require the full viewer (such as e.g. saving).

For everything else, please use direct `this._pdfViewer` calls wherever possible to avoid breaking the standalone viewer components unnecessarily; note https://github.com/mozilla/pdf.js/tree/master/examples/components

Hence, for the FirstPage, LastPage, NextPage, and PrevPage event specifically: please revert to the previous version of this PR (with my two comments addressed).

Snuffleupagus (changes requested):

web/pdf_scripting_manager.js
```diff
@@ -305,6 +305,27 @@ class PDFScriptingManager {
...
+        case "LastPage":
+          this._pdfViewer.currentPageNumber = this._pdfViewer.pagesCount;
+          break;
+        case "NextPage":
+          this._pdfViewer.nextPage();
+          break;
+        case "PrevPage":
+          this._pdfViewer.previousPage();
+          break;
+        case "ZoomViewIn":
```
Snuffleupagus, line 323: Similar to the "zoom"-case in the existing code, please add the following check to prevent issues:
```javascript
if (isInPresentationMode) {
  return;
}
```

web/pdf_scripting_manager.js
```diff
@@ -305,6 +305,27 @@ class PDFScriptingManager {
...
+          this._eventBus.dispatch("zoomin", { source: this });
+          break;
+        case "ZoomViewOut":
```
Snuffleupagus, line 326: Same as above.

src/scripting_api/app.js
```diff
@@ -479,8 +479,30 @@ class App extends PDFObject {
...
+    }
+
+    switch (item) {
+      case "FitPage":
+        this._send({ command: "zoom", value: "page-fit" });
+        break;
+      case "Print":
+        this._send({ command: "print" });
+        break;
+    }
```
Snuffleupagus, line 505: Now that the `switch` is added, can't you simply handle all of the cases within that block instead?

```javascript
    switch (item) {
      case "SaveAs":
      case "FirstPage":
      case "LastPage":
      case "NextPage":
      case "PrevPage":
      case "ZoomViewIn":
      case "ZoomViewOut":
        this._send({ command: item });
        break;

      case "FitPage":
        this._send({ command: "zoom", value: "page-fit" });
        break;
      case "Print":
        this._send({ command: "print" });
        break;
    }
```

Snuffleupagus (approved):
r=me with the final comments addressed; thank you!

Really sorry about the back-and-forth here!

PR #14040, opened 2021-09-17
JS - Avoid the Stay/Leave popup when clicking on a button with a JS action

### LINKED ISSUE #14039: When a form is resetted using a JS action, Firefox ask for "Stay... or Leave..."
[0 NIF MODELO 30.pdf](https://github.com/mozilla/pdf.js/files/7187486/0.NIF.MODELO.30.2.pdf)

Configuration:
- Web browser and its version: firefox nightly

1. Open the PDF
2. Click on the button  `Rellenar Formulario` (top right of the first page)
3. There is the pop-up "Stay... or Leave...`.

It's a bit annoying from the user pov and it doesn't make sense (nobody wants to leave the form but just reset it)

It works fine when replacing  `""` by `"#"` here:
https://github.com/mozilla/pdf.js/blob/7082ff9bf876b98fbda118cdbd6267d3c0df6948/src/display/annotation_layer.js#L483

@Snuffleupagus, you changed this few month ago:
https://github.com/mozilla/pdf.js/commit/e2b6d79dee9ffd2aba715d6b3cb8fa143d86153c

do you have some ideas ?

calixteman:
- it aims to fix #14039.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -502,6 +502,10 @@ class LinkAnnotationElement extends AnnotationElement {
         return false;
       };
     }
+
+    if (!link.onclick) {
+      link.onclick = () => false;
+    }
     link.className = "internalLink";
   }
 }


Snuffleupagus (approved):
r=me, since this looks reasonable; thank you!

I'm assuming that you've successfully tested this, such that it a) fixes the issue as reported and b) doesn't break the onmouseup/onmousedown events?

calixteman:
Yes I tested the issue and about the second point it shouldn't impact mouseup and mousedown events since they're triggered before the click one (see https://w3c.github.io/uievents/#events-mouseevent-event-order).

PR #14041, opened 2021-09-17
Support cmaps with only CID characters, when building the ToUnicode-map (issue 9367)

### LINKED ISSUE #9367: pdf rendered completely wrong

Configuration:
- Web browser and its version: Chrome 63.0.3239.132 
- PDF.js version: using your official demo project
- Is a browser extension: No

1.  Just load attached pdf
2.

[E-WG18_105147.pdf](https://github.com/mozilla/pdf.js/files/1634833/E-WG18_105147.pdf)

image

Snuffleupagus:
In this particular case the `CMap`-data that we create contains only numbers, but no strings, which causes `PartialEvaluator.readToUnicode` to create a ToUnicode-map with only empty strings.

*Please note:* This is yet another case where I don't know if it's necessarily the best and most correct solution, but it does fix the referenced issue.

Fixes #9367

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -3455,6 +3455,11 @@ class PartialEvaluator {
           // NOTE: cmap can be a sparse array, so use forEach instead of
           // `for(;;)` to iterate over all keys.
           cmap.forEach(function (charCode, token) {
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me; thank you!

/botio makeref

Issue #13872, opened 2021-08-04
Document information is unreadable in high contrast mode


brendandahl:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1722984

![image](https://user-images.githubusercontent.com/942640/128253205-86da416c-9f3b-42d9-b1d1-7e0a272b22c1.png)

brendandahl:
@utopianknight since you did the original high contrast stuff would you be interested in fixing this too?

Snuffleupagus:
Is this a more general issue with the overlays, i.e. does it affect the PasswordPrompt as well?

utopianknight:
Everything seems to be working fine with system colors (both white and black themes). I think the problem occurs when Firefox does the color overriding. So this is going to take a while. Unless someone else looks into it.

PR #14042, opened 2021-09-17
Fix dialogs with forced colors (bug 1722984)

serdnab:
Fix for #13872.
When forced colors are set, [this](https://github.com/mozilla/pdf.js/commit/a3cc2e63ebdec46deb6a980b45c95d9c61322497#diff-d1047e1376717fbe7c554bb385cd48b2dcbbb135e02884be3db97ad072ba741eL153) variable is set to use the system buttontext color, preventing the use of the color set in the color settings.
This assignment can be removed since the colors of the html elements that use this variable are overwritten by the browser when the forced colors are set.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -150,7 +150,6 @@
 
 @media screen and (forced-colors: active) {
   :root {
-    --main-color: ButtonText;
     --button-hover-color: Highlight;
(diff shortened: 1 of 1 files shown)


brendandahl:
Look good, thank you!

PR #14043, opened 2021-09-18
Re-factor the `EventBus` and `isInAutomation` handling (PR 11655 follow-up)

Snuffleupagus:
Rather than forcing the "regular" `EventBus` to check and handle `isInAutomation` for every `dispatch` call, we can take advantage of subclassing instead.
Hence this PR introduces a new `AutomationEventBus` class, which extends `EventBus`, and is used by the default viewer when `isInAutomation === true`.


timvandermeij (approved):

timvandermeij:

PR #14044, opened 2021-09-18
Annotations - Avoid empty value in text field when storage contains something for it (bug 1719148)

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1719148;
  - JS can set a property for a non-rendered annotation using the annotationStorage but the other unset default properties must be used when the annotation is finally rendered;
  - so this patch just adds the properties already set in the annotationStorage to the default value.

@gijsk shared his pdf with me, I was able to reproduce his issue with a local firefox build and I verified that this patch fixes the issue.

### DIFF
--- a/src/display/annotation_storage.js
+++ b/src/display/annotation_storage.js
@@ -42,7 +42,12 @@ class AnnotationStorage {
    * @returns {Object}
    */
   getValue(key, defaultValue) {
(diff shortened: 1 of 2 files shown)


Snuffleupagus (approved):
r=me, thank you!

Issue #12865, opened 2021-01-15
Weird behaviour of an input in a form


calixteman:
Attach (recommended) or Link to PDF file here:
https://github.com/mozilla/pdf.js/files/2321540/issue6127.pdf

When entering 13 chars in the 3rd text field (i.e. `NUMERO D’IMMATRICULATION`) the first one appears in the first slot and when the input loses the focus everything looks correct.

With focus:
![image](https://user-images.githubusercontent.com/5641725/104729511-d51a7c80-5738-11eb-8d5e-a9e7818adce8.png)

Without:
![image](https://user-images.githubusercontent.com/5641725/104729632-01ce9400-5739-11eb-9d4d-18c75a152330.png)

The issue is reproductible with firefox nightly and locally on master with firefox and chrome (`gulp server`).

somil24:
Sir I would like to work on this issue and also it would be my first contribution so if you could guide me through.

timvandermeij:
Fixed by #13690.

PR #14049, opened 2021-09-18
Annotation - Use border and background colors from MK dictionary

### LINKED ISSUE #13003: PDF rendering problem: Checkboxes have black background

calixteman:
- it aims to fix #13003;
  - set the bg and fg colors as they're in the pdf;
  - put a transparent overlay to help to see the fields.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -231,12 +231,12 @@ class AnnotationFactory {
   }
 }
 
-function getRgbColor(color) {
-  const rgbColor = new Uint8ClampedArray(3);
+function getRgbColor(color, defaultColor = new Uint8ClampedArray(3)) {
   if (!Array.isArray(color)) {
-    return rgbColor;
+    return defaultColor;
   }
 
+  const rgbColor = defaultColor || new Uint8ClampedArray(3);
   switch (color.length) {
     case 0: // Transparent, which we indicate with a null value
       return null;
@@ -254,7 +254,7 @@ function getRgbColor(color) {
       return rgbColor;
 
     default:
-      return rgbColor;
+      return defaultColor;
   }
 }
 
@@ -365,6 +365,7 @@ class Annotation {
     this.setColor(dict.getArray("C"));
     this.setBorderStyle(dict);
     this.setAppearance(dict);
+    this.setBorderAndBackgroundColors(dict.get("MK"));
 
     this._streams = [];
     if (this.appearance) {
@@ -376,6 +377,8 @@ class Annotation {
       annotationFlags: this.flags,
       borderStyle: this.borderStyle,
       color: this.color,
+      backgroundColor: this.backgroundColor,
+      borderColor: this.borderColor,
       contentsObj: this._contents,
       hasAppearance: !!this.appearance,
       id: params.id,
@@ -604,6 +607,23 @@ class Annotation {
   }
 
   /**
+   * Set the color for background and border if any.
+   * The default values are transparent.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {Dict} mk - The MK dictionary
+   */
+  setBorderAndBackgroundColors(mk) {
+    if (mk instanceof Dict) {
+      this.borderColor = getRgbColor(mk.getArray("BC"), null);
+      this.backgroundColor = getRgbColor(mk.getArray("BG"), null);
+    } else {
+      this.borderColor = this.backgroundColor = null;
+    }
+  }
+
+  /**
    * Set the border style (as AnnotationBorderStyle object).
    *
    * @public
@@ -765,6 +785,8 @@ class Annotation {
         id: this.data.id,
         actions: this.data.actions,
         name: this.data.fieldName,
+        strokeColor: this.data.borderColor,
+        fillColor: this.data.backgroundColor,
         type: "",
         kidIds: this.data.kidIds,
         page: this.data.pageIndex,
@@ -1874,6 +1896,8 @@ class TextWidgetAnnotation extends WidgetAnnotation {
       rect: this.data.rect,
       actions: this.data.actions,
       page: this.data.pageIndex,
+      strokeColor: this.data.borderColor,
+      fillColor: this.data.backgroundColor,
       type: "text",
     };
   }
@@ -2304,6 +2328,8 @@ class ButtonWidgetAnnotation extends WidgetAnnotation {
       hidden: this.data.hidden,
       actions: this.data.actions,
       page: this.data.pageIndex,
+      strokeColor: this.data.borderColor,
+      fillColor: this.data.backgroundColor,
       type,
     };
   }
@@ -2385,6 +2411,8 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
       actions: this.data.actions,
       items: this.data.options,
       page: this.data.pageIndex,
+      strokeColor: this.data.borderColor,
+      fillColor: this.data.backgroundColor,
       type,
     };
   }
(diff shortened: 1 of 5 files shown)


web/annotation_layer_builder.css
```diff
@@ -13,6 +13,10 @@
  * limitations under the License.
  */
 
+:root {
+  --unfocused-field-background: url("data:image/svg+xml;charset=UTF-8,<svg width='1px' height='1px' xmlns='http://www.w3.org/2000/svg'><rect width='100%' height='100%' style='fill:rgba(0, 54, 255, 0.13);'/></svg>");
```
Snuffleupagus, line 17: This seem to be using the exact variable name as in https://github.com/mozilla/pdf.js/blob/master/web/xfa_layer_builder.css, could that possibly become an issue?

Given that all CSS files are bundled together into one big file during building, I assume that one of them will end up overriding the other one. If, sometime in the future, these two images are no longer *identical* I suppose that that could lead to some problems.

Hence I'd suggest slightly tweaking the names to:
 - `--annotation-unfocused-field-background` in this file, for the new rule.
 - `--xfa--unfocused-field-background` in web/xfa_layer_builder.css, for the existing rule.

src/core/annotation.js
```diff
@@ -230,12 +230,12 @@ class AnnotationFactory {
...
 
-function getRgbColor(color) {
-  const rgbColor = new Uint8ClampedArray(3);
+function getRgbColor(color, defaultColor = new Uint8ClampedArray(3)) {
   if (!Array.isArray(color)) {
-    return rgbColor;
+    return defaultColor;
   }
 
+  const rgbColor = defaultColor || new Uint8ClampedArray(3);
```
Snuffleupagus, line 238: Unless I'm missing something, I *think* that you can remove `defaultColor` from this line now?

Snuffleupagus:
@timvandermeij Since you reviewed the previous PR, i.e. #13006, mind taking a look at this one when you've got time (so I'm not overlooking anything, given the previous discussion in that PR)?

Snuffleupagus (changes requested):
Looking at the latest round of test results, there's also a couple of *breaking* errors during the reference tests; one example being:
`TEST-UNEXPECTED-FAIL | test failed issue13823 | in firefox | page1 round 1 | render : Error: rasterizeAnnotationLayer: "can't access property 0, color is undefined".`

calixteman:
@Snuffleupagus I forgot to initialize the colors:
https://github.com/mozilla/pdf.js/pull/14049/files#diff-a7e409abd3d3209d32d3adc8457c5e23feb4c06113a2e065b0c6a9322a3a3369R622

timvandermeij:
@Snuffleupagus Thank you for the ping; I'll have a look at this PR today so we can hopefully get it merged soon.

src/core/annotation.js
```diff
@@ -603,6 +606,23 @@ class Annotation {
...
+   * Set the color for background and border if any.
+   * The default values are transparent.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {Dict}  - The MK dictionary
+   */
+  setBorderAndBackgroundColors(mk) {
+    if (mk instanceof Dict) {
+      this.borderColor = getRgbColor(mk.getArray("BC"), null);
```
timvandermeij, line 619: The borders seem thicker now in the reference tests; it looks like the borders are applied twice now, once here and once in existing CSS perhaps?

Moreover, when opening `annotation-text-widget.pdf` from the test suite and clicking the combs field, the border is now extended behind the field (due to the way padding works for the combs fields). Would it maybe be a solution to make the border, just like the background, transparent on focus?

src/core/annotation.js
```diff
@@ -603,6 +606,23 @@ class Annotation {
     this.color = getRgbColor(color);
   }
 
+  /**
+   * Set the color for background and border if any.
+   * The default values are transparent.
+   *
+   * @public
+   * @memberof Annotation
+   * @param {Dict}  - The MK dictionary
```
timvandermeij, line 615: `@param {Dict} mk - The MK dictionary` (the variable name was missing)

timvandermeij (commented):
Looks good to me with these two comments addressed.

src/core/annotation.js
calixteman: The container and the element had a border and so I just removed the one from the element.
About the focus issue on the comb field, I just removed https://github.com/mozilla/pdf.js/blob/93ed4bfa119fe82c1fe1fe9cd7d08b1ae719b49e/web/annotation_layer_builder.css#L144-L152 which seems to be useless now (something has likely been fixed in Firefox) and without any consequence with the pdf in #12865.

timvandermeij (approved):

timvandermeij:
Thank you for improving this!

/botio makeref

timvandermeij:
Euh... what? Let's try that again.

/botio-windows makeref

timvandermeij:
Windows works, now Linux...

/botio-linux makeref

test/test_manifest.json
```diff
@@ -5975,5 +5975,12 @@
            "value": true
          }
        }
+    },
+    {  "id": "issue13003",
+       "file": "pdfs/issue13003.pdf",
```
Snuffleupagus, line 5980: I'm not sure why makeref doesn't fail on the bots, since this file isn't included in anywhere in the diff (neither directly or as a linked test-case). Locally this causes makeref to fail completely for me.

@calixteman Can you please do a follow-up PR adding the missing file?

src/display/annotation_layer.js
```diff
@@ -244,7 +244,8 @@ class AnnotationElement {
           break;
       }
 
-      if (data.color) {
+      const borderColor = data.borderColor || data.color || null;
+      if (borderColor) {
         container.style.borderColor = Util.makeHexColor(
           data.color[0] | 0,
           data.color[1] | 0,
```
Snuffleupagus, line 251: Is this `makeHexColor`-call *supposed* to use `data.color` now, since naively I'd assume that it should actually use `borderColor` instead?
As-is the borderColor is being ignored, which I'm assuming wasn't intended; was this perhaps fallout from addressing one of the review comments?

calixteman: `borderColor` is set from `MK-BC` entry and `color` is set from `C` one.
If I understand the spec correctly, `BC` (see https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=417) is optional and for widget annotations, when `C` (see https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=392) is for icon, popup and link, so only one of the `BC` and `C` should be set.

calixteman: If I'm correct, I should likely add a comment.

PR #14050, opened 2021-09-18
Tweak how fonts with an /Encoding are handled in `adjustToUnicode` (issue 14048, PR 13277 follow-up)

### LINKED ISSUE #14048: Text selection broken with embedded Type1 font
[text-30.pdf](https://github.com/mozilla/pdf.js/files/7190322/text-30.pdf)

Configuration:
- Web browser and its version: Chrome Version 93.0.4577.82 (Official Build) (x86_64)
- Operating system and its version: MacOS Big Sur 11.6
- PDF.js version: Tested and verified the bug exists in v2.9.359 and v2.10.377
- Is a browser extension: No

1. Open the attached file.
2. Try to select the text. It shows "6" instead of the expected "10". This behavior exists on v2.9.359 and v2.10.377, and also as of the latest commit on the master branch (83d3bb43f434b8475cd13f1f9456099677d60acb).

On v2.8.335 the text selected is "10" as expected.

Note that this example was extracted from a larger paper on arXiv: https://arxiv.org/pdf/1106.6241.pdf
On Page 14, Figure 8, the labels on the axes are shown as 10^-3X, but selecting the text "10" shows "6" instead. This doesn't appear to be a problem on other similar figures (e.g. Figure 4 on Page 10 doesn't appear to have this issue).

Snuffleupagus:
Currently we only exclude /Encoding entries that also contains a /Differences array, which is the cause of the text-selection problem in the referenced issue.
In order to address this we'll now also exclude /Encoding entries that contain one of the predefined *named* encodings, and no longer require that it also contains a /Differences array.

*Please note:* This patch cases a small "regression" in the `bug1130815-text` test-case, however this is actually an improvement when compared with Adobe Reader and PDFium (in Google Chrome).

Fixes #14048


Snuffleupagus:
I don't really understand the XFA-related font handling good enough to explain the change in `xfa_issue13668-page10`, however compared with Adobe Reader it's actually a rendering improvement!

timvandermeij (approved):

timvandermeij:
Looks like a good improvement to me; thanks!

/botio makeref

calixteman:
> I don't really understand the XFA-related font handling good enough to explain the change in `xfa_issue13668-page10`, however compared with Adobe Reader it's actually a rendering improvement!

The difference comes from the height of the div which contains the text "Entité d'investissement..... ayant une clientèle local".
Before your patch it was 319px and after it was 307px. So I'd say that the text here is measured thanks to:
https://github.com/mozilla/pdf.js/blob/580bfad6288ca17b90db2dea90914103394d4c1d/src/core/xfa/text.js#L149
and here we use font information to layout text and finally get dimensions for a text container.
Anyway, it's an improvement, so thank you for that.

Issue #7022, opened 2016-02-24
Document zoomed if still scrolling and meta keys pressed.


brendandahl:
STR:
On OSX scroll quickly using the touchpad, then press command.
Expected:
Document is not zoomed
What Happens:
Document is zoomed

We should only zoom if them command key was pressed before the scroll.

timvandermeij:
Fixed by #7097.

PR #7097, opened 2016-03-21
Prevents document zooming if still scrolling and meta keys pressed (Issue #7022)

ManasJayanth:
Disables zooming when the stream of events that don't have meta/ctrl
keys active at the outset, for 100ms

[UPDATE] 1000ms

PR #14051, opened 2021-09-19
fix mouse wheel scroll delay problem

mn297:
Removed the annoying delay when attempting to mouse wheel zoom while viewing a PDF, does not cause other issues. Tested with PDF Viewer extension for chrome https://chrome.google.com/webstore/detail/pdf-viewer/oemmndcbldboiebfnladdacbdfmadadm?hl=en


Snuffleupagus:
Thanks for the PR, however we unfortunately cannot accept this.

> Tested with PDF Viewer extension for chrome

Please note that the Firefox built-in PDF Viewer is the primary development target for this library.
Furthermore, this timeout was added for multiple reasons:
 - Originally to fix issue #7022, see also PR #7097.
 - Then extended to fix [bug 1503412](https://bugzilla.mozilla.org/show_bug.cgi?id=1503412), see #10351.

PR #14052, opened 2021-09-19

Snuffleupagus:
*Please note:* I did try to update the `eslint-plugin-sort-exports` package as well, however that caused *runtime errors* during linting so I'm simply ignoring it for now.

Furthermore, the large number of l10n-strings being removed is (mostly) due to PR #12975 and it apparently takes a while before unused strings are removed in Firefox. I'm guessing that l10n-strings are kept until no official Firefox release uses them any more!?


timvandermeij (approved):

timvandermeij:
Thanks!

PR #14053, opened 2021-09-19
Move the zoomIn/zoomOut functionality into `BaseViewer` (PR 14038 follow-up)

Snuffleupagus:
Given the simplicity of this functionality, we can move it from the default viewer and into the `BaseViewer` class instead. This way, it's possible to support more scripting functionality in the standalone viewer components; please see PR #14038.

Please note that I purposely went with `increaseScale`/`decreaseScale`-method names, rather than using "zoom", to better match the existing `currentScale`/`currentScaleValue` getters/setters that's being used in the `BaseViewer` class.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -26,8 +26,6 @@ import {
   isValidRotation,
   isValidScrollMode,
   isValidSpreadMode,
-  MAX_SCALE,
-  MIN_SCALE,
   noContextMenuHandler,
   normalizeWheelEventDirection,
   parseQueryString,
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
I suppose that it could be argued that scripting support in the standalone viewer components won't ever be *perfect*, and that the zoomIn/zoomOut functionality simply belongs in the default viewer and *not* in the `BaseViewer`.
Hence I'll not definitely mind if we ultimately decide to reject this PR, however I wanted to at least submit it for discussion!

timvandermeij (approved):
I'd say that this is a good change to make. Even though scripting support might not be perfect, this seems like a useful and simple enough addition.

timvandermeij:
Thanks!

PR #14055, opened 2021-09-20
Add `PDF_TO_CSS_UNITS` to the `PixelsPerInch`-structure

Snuffleupagus:
Rather than re-computing this value in a number of different places throughout the code-base[1], we can expose this in the API via the existing `PixelsPerInch`-structure instead.
There's also been feature requests asking for the old `CSS_UNITS` viewer constant to be made accessible, such that it could be used in third-party implementations.

I suppose that it could be argued that it's somewhat confusing to place a unitless property in `PixelsPerInch`, however given that the `PDF_TO_CSS_UNITS`-property is defined strictly in terms of the existing properties this is hopefully deemed reasonable.

---
[1] These include:
 - The viewer, with the `CSS_UNITS` name.
 - The reference-tests.
 - The display-layer, when rendering images; see PR 13991.

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -879,7 +879,7 @@ function getImageSmoothingEnabled(transform, interpolate) {
   scale[0] = Math.fround(scale[0]);
   scale[1] = Math.fround(scale[1]);
   const actualScale = Math.fround(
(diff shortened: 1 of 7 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me; thanks!

PR #14057, opened 2021-09-22
Support CMap-data with only strings, when parsing TrueType composite fonts (bug 920426)

Snuffleupagus:
In the referenced bug, the embedded fonts contain custom CMap-data that only include strings. Note how for embedded composite TrueType fonts we're using the CMap-data when building the glyph mapping, and currently we end up with a completely empty map because the code expects only CID *numbers*.
Furthermore, just fixing the glyph mapping alone isn't sufficient to fully address the bug, since we also need to consider this "special" kind of CMap-data when looking up glyph widths.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=920426


brendandahl:
I think this may be a bug in the cmap parser. Looking at the spec on beginbfchar, it says the second value can be either a "code or name". 
[Here](https://github.com/mozilla/pdf.js/blob/d370a281c450887e65b677520f5395016d182429/src/core/cmap.js#L787) we could instead check if the first char of the string is a `/` if it isn't then convert it to an int.

brendandahl:
Nevermind about checking the `/`. If that's the case the lexer will return a name obj.

Snuffleupagus:
One thing that complicates all of this somewhat is that depending on how/where we're invoking the `CMapFactory` in our existing code, we actually expect the returned data to look subtly different:
 - In the `src/core/fonts.js` file, when accessing the `cMap`-data, the surrounding code always expect to get a CID *number* in return. The situation fixed in this patch should thus be *very rare*, since this is the first case we know of.
 - In the `PartialEvaluator.readToUnicode`-method, please see [this code in particular](https://github.com/mozilla/pdf.js/blob/c3ca78fdf8759d8cf7c8c4f35627c25a64e6bc7b/src/core/evaluator.js#L3443-L3478), we're actually (by default) expecting that the `CMapFactory` will return *string* data. (As mentioned in a comment there, I recently added a code-path to handle essentially the "inverse" of this bug to fix another older issue.)

My initial thought when looking at both this bug, and issue 9367, was indeed that we should perhaps (somehow) change the `CMapFactory` to fix these bugs. However, since I noticed that we have different expectations for the return values depending on how/where we're using that code, it wasn't at all clear (to me) how to reconcile those two cases in the `CMapFactory` directly.

Also, I should probably add that the situation fixed here seems *extremely rare* in practice, since we've seen a total of *two* cases where the existing assumptions regarding numbers/strings for the CMap-data doesn't hold.

@brendandahl Given the above, how should we proceed here?

src/core/fonts.js
```diff
@@ -401,6 +401,18 @@ function buildToFontChar(encoding, glyphsUnicodeMap, differences) {
   return toFontChar;
 }
 
+function convertCidString(charCode, cid, shouldThrow = false) {
+  if (cid.length === 2) {
```
brendandahl, line 405: Should we support length === 1? Seems like that should be valid too.

brendandahl (commented):
Making cmap foreach always call the callback with the same type seems like something we should fix. I'd lean towards an integer and then convert that to a string if we need to for the unicode stuff.

Snuffleupagus:
> I'd lean towards an integer and then convert that to a string if we need to for the unicode stuff.

For the `PartialEvaluator.readToUnicode` case, wouldn't that mean that we'd need to essentially convert the data *twice* then?

Also, looking [at this code](https://github.com/mozilla/pdf.js/blob/c3ca78fdf8759d8cf7c8c4f35627c25a64e6bc7b/src/core/evaluator.js#L3463-L3475), how would you even get the `CMap.forEach`-method to return *one* integer value for the string-case?
I might be missing something obvious here, but I really don't understand how you'd actually implement that.

brendandahl:
Yeah, forgot about how the ToUnicode cMaps work. For regular font ones the charcode should always fit into a 32bit int. What about always returning a string, that conversion should be pretty fast. Or alternatively differrent forEach methods for strings and ints?

Snuffleupagus:
> Or alternatively differrent forEach methods for strings and ints?

I've tried to implement this, which also required some changes to the `lookup`-methods as well, and while it seem to work locally (all tests pass) it does feel a little clunky. @brendandahl Feedback please :-)

brendandahl (approved):
Sounds good. r+ with passing tests

PR #14058, opened 2021-09-22
[api-minor] Change `EventBus.dispatch` to only support *one* data-argument

Snuffleupagus:
This is consistent with how the `EventBus` has *always* been used internally in the viewer, and allows a slight simplification of the relevant code.


timvandermeij (approved):

timvandermeij:
Thank you for simplifying this code!

PR #14063, opened 2021-09-22
[GENERIC viewer] Warn about AppOptions being overridden by Preferences during loading

Snuffleupagus:
Currently any AppOptions set using e.g. the "webviewerloaded" event listener can/will by default be overridden when the Preferences are read.
To avoid that happening the "disablePreferences"-option can be used, however unless it's been explicitly set all non-default AppOptions will be silently ignored. This patch thus attempts to improve the current situation somewhat, for third-party implementations, by logging a warning in the console when this happens.

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -301,13 +301,20 @@ const PDFViewerApplication = {
    */
   async _readPreferences() {
     if (
-      (typeof PDFJSDev === "undefined" ||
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks like a sensible addition to me; thanks!

PR #14065, opened 2021-09-22
[api-minor] Stop exporting, by default, a few additional Font properties (PR 11777 follow-up)

Snuffleupagus:
*This is similar to the "isSymbolicFont"-property, which is no longer exported by default after PR #11777.*

Both "isMonospace" and "isSerifFont" are internal properties, used during font parsing and building of the glyph mapping on the worker-thread.
However both of these properties are completely unused on the main-thread and/or in the API, and accessing them they will now require setting the `fontExtraProperties`-option when calling `getDocument`.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #14066, opened 2021-09-23
Print a special warning message, in the viewer, for XFA Foreground documents

Snuffleupagus:
Currently XFAF documents use the same warning message as in the XFA *disabled* case, which is neither helpful nor correct.


calixteman (approved):
LGTM, thanks for doing that.

PR #14067, opened 2021-09-23
Don't save anything in XFA entry if no XFA! (bug 1732344)

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1732344
  - rename some variables to have a more clear code;
  - and last but no least, add a unit test to test saving.

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -1283,6 +1283,28 @@ describe("api", function () {
       await Promise.all([loadingTask1.destroy(), loadingTask2.destroy()]);
     });
 
+    it("write a value in an annotation, save the pdf and load it", async function () {
+      let loadingTask = getDocument(buildGetDocumentParams("evaljs.pdf"));
+      let pdfDoc = await loadingTask.promise;
+      const value = "Hello World";
+
+      pdfDoc.annotationStorage.setValue("55R", { value });
+
+      const data = await pdfDoc.saveDocument();
+      await loadingTask.destroy();
+
+      loadingTask = getDocument(data);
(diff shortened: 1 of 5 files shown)


src/display/api.js
```diff
@@ -2916,6 +2916,7 @@ class WorkerTransport {
         annotationStorage: this.annotationStorage.serializable,
         filename: this._fullReader?.filename ?? null,
       })
+      .catch(ex => warn(`Error when saving the document: ${ex.message}`))
```
Snuffleupagus, line 2919: Is this a left-over from debugging, or do you actually intend to print a warning in the API-code?

As-is I believe that the `saveDocument´ method will now *always* return a resolved Promise,  which isn't intended if/when failure occurs (for example, it'll break the fallback in the viewer).

In any case, can you please just log a warning [here instead](https://github.com/mozilla/pdf.js/blob/c914e9f0a631b92ed81d1a685b42ae088adefcdc/web/app.js#L1016) using `console.error`.

test/unit/api_spec.js
```diff
@@ -2322,5 +2322,27 @@ describe("api", function () {
         await loadingTask.destroy();
       }
     );
+
+    it("write a value in an annotation, save the pdf and load it", async function () {
```
Snuffleupagus, line 2326: Nit: Sorry, but this unit-test ended up in the "PDFDataRangeTransport"-block which isn't a great fit. Can you please move it to e.g. here instead: https://github.com/mozilla/pdf.js/blob/c914e9f0a631b92ed81d1a685b42ae088adefcdc/test/unit/api_spec.js#L1285

PR #14069, opened 2021-09-24
Mark the `paintJpegXObject` operator as deprecated (PR 11601 follow-up)

Snuffleupagus:
After PR #11601, the `paintJpegXObject` operator is no longer used for anything. While I don't think we can just remove it, and essentially leave a "hole" in the `OPS` structure, we should at least mark it as explicitly unused to aid readability/maintainability of the code.


timvandermeij (approved):

timvandermeij:
Good idea!

PR #14070, opened 2021-09-24
Some small readability improvements in the `MessageHandler` code

Snuffleupagus:
In particular the `_processStreamMessage`-method is a bit cumbersome to read, given the way that the current streamController/streamSink is accessed, which we can improve with a couple of local variables.


timvandermeij (approved):

timvandermeij:
Thank you for doing this!

PR #14074, opened 2021-09-24
[api-minor] Add basic support for RTL text-content in PopupAnnotations (issue 14046)

### LINKED ISSUE #14046: Text annotation in hebrew is rendered in LTR mode
[mytestfile.pdf](https://github.com/mozilla/pdf.js/files/7190033/mytestfile.pdf)

Configuration:

1. Open the pdf  
2. Click on the yellow area

The text layout should be RTL.

The text layout is LTR.

It works correctly in evince and in acrobat.
I added a `dir="rtl"` on the `p` element in using devtools and the rendering is the same as in acrobat.

Snuffleupagus:
In order to implement this, we utilize the existing `bidi` function to infer the text-direction of /T and /Contents entries. While this may not be perfect in cases where one PopupAnnotation mixes LTR and RTL languages, it should work well enough in most cases.
To avoid having to add *two new* properties in lots of annotations, supplementing the existing `title`/`contents`-properties, this patch instead re-factors the existing code such that the properties are replaced by Objects (containing `str` and `dir`).

*Please note:* In order avoid breaking existing third-party implementations, `GENERIC`-builds of the PDF.js library will still provide the old `title`/`contents`-properties on annotations returned by `PDFPageProxy.getAnnotations`.

Fixes #14046


src/display/annotation_layer.js
```diff
@@ -1601,8 +1605,8 @@ class FreeTextAnnotationElement extends AnnotationElement {
   constructor(parameters) {
     const isRenderable = !!(
       parameters.data.hasPopup ||
-      parameters.data.title ||
-      parameters.data.contents
+      parameters.data.titleObj?.str ||
```
calixteman, line 1608: I think the `?` is useless here (`a={b:1}; a.c === a?.c && a.b === a?.b`).

calixteman (approved):
LGTM and thank you for doing this.

src/display/annotation_layer.js
Snuffleupagus: If `titleObj === undefined` would ever happen, trying to access the `str`-property would throw and prevent the rest of the AnnotationLayer from rendering. Hence it seemed *overall safer* to use optional chaining everywhere instead.

calixteman: For me it wasn't possible to have a `titleObj === undefined`, but yes let's do that: it's safer.

PR #14076, opened 2021-09-25
XFA - Add <a> element in button when an url is detected (bug 1716758)

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1716758;
  - some buttons have a JS action with the pattern `app.launchURL(...)` (or similar) so extract when it's possible the url and generate a <a> element with the href equals to the found url;
  - pdf.js already had some code to handle that so this patch slightly refactor that.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -17,6 +17,7 @@ import {
   addDefaultProtocolToUrl,
   collectActions,
   MissingDataException,
+  recoverJsURL,
   toRomanNumerals,
   tryConvertUrlEncoding,
 } from "./core_utils.js";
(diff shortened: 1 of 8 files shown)


src/shared/util.js
```diff
@@ -1030,6 +1030,32 @@ function createObjectURL(data, contentType = "", forceDataSchema = false) {
   return buffer;
 }
 
+function getURL(str) {
```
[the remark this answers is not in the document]
Snuffleupagus: Also, since this function is only being used in the `src/core/` folder, please move it to `src/core/core_utils.js` instead to avoid (needlessly) duplicating this code in both the `pdf.js` and `pdf.worker.js` files.

src/core/catalog.js
```diff
@@ -1400,25 +1401,10 @@ class Catalog {
...
-
-              if (jsUrl[3] === "true" && jsUrl[1] === "app.launchURL") {
-                resultObj.newWindow = true;
-              }
+            const parsed = getURL(stringToPDFString(js));
+            url = parsed.url;
+            resultObj.newWindow = parsed.newWindow;
+            if (url) {
               break;
             }
```
Snuffleupagus, line 1409: With the other suggestions implementation, this could be simplified a bit more:
```javascript
const jsURL = js && recoverJsURL(stringToPDFString(js));
if (jsURL) {
  url = jsURL.url;
  resultObj.newWindow = jsURL.newWindow;
  break;
}
```

test/unit/xfa_tohtml_spec.js
```diff
@@ -588,4 +588,45 @@ describe("XFAFactory", function () {
...
+  </xfa:datasets>
+</xdp:xdp>
+    `;
+    const factory = new XFAFactory({ "xdp:xdp": xml });
+
+    expect(factory.numberPages).toEqual(1);
+
+    const pages = factory.getPages();
+    const a = searchHtmlNode(pages, "name", "a");
+    expect(a.attributes.href).toEqual("https://github.com/mozilla/pdf.js");
```
Snuffleupagus, line 630: Can you please add a check to ensure that the `target` is also correct, since you're using the "app.launchURL(...)" format to specify the URL?

Snuffleupagus (approved):
r=me, looks good to me with the comment addressed; thank you!

test/unit/xfa_tohtml_spec.js
```diff
@@ -17,18 +17,20 @@ import { isNodeJS } from "../../src/shared/is_node.js";
 import { XFAFactory } from "../../src/core/xfa/factory.js";
 
 describe("XFAFactory", function () {
-  function searchHtmlNode(root, name, value, byAttributes = false) {
+  function searchHtmlNode(root, name, value, byAttributes = false, nth = [0]) {
```
Snuffleupagus, line 20: *I might be missing something obvious, but this looks confusing to me:*
Does the new parameter *need* be be an Array, or would a simple value suffice (i.e. `nth = 0`) since that'd be a bit more easy to understand?

calixteman: It's just to pass the argument by "reference": the `nth--` won't change the value of `nth` in the caller.

PR #14077, opened 2021-09-25
Fix issues in driver.js when getting css sheets

calixteman:


calixteman:
In dumping svg, I saw an `undefined` which led me to write this patch.

Snuffleupagus (approved):
r=me, with passning tests; nice catch!

timvandermeij:
Interesting how this seemingly didn't cause noticeable problems before; it looks like it more-or-less worked "accidentally". Anyway, nice find!

PR #14081, opened 2021-09-26
[api-minor] Move the `addDefaultProtocolToUrl`/`tryConvertUrlEncoding` functionality into the `createValidAbsoluteUrl` function

Snuffleupagus:
Having recently worked with, and reviewed patches touching, this code it seemed that it's probably not a bad idea to move that functionality into `createValidAbsoluteUrl` as new options instead.

For the `addDefaultProtocolToUrl` functionality in particular, the existing helper function was not only moved but slightly improved as well. Looking at the code, I realized that there's a small risk that it would incorrectly match a *relative* URL-string too.

With these changes, the `createValidAbsoluteUrl` call-sites in the `src/core/`-code can be simplified a little bit.

*Please note:* This patch may, indirectly, change the format of the `unsafeUrl`-property returned with relevant Annotations and OutlineItems; hence the `api-minor` tag.
However, I'd argue that it's actually more correct this way since the whole purpose of `unsafeUrl` is/was to return the URL data as-is without any parsing done.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -14,14 +14,6 @@
  */
 
 import {
-  addDefaultProtocolToUrl,
-  collectActions,
-  MissingDataException,
-  recoverJsURL,
-  toRomanNumerals,
-  tryConvertUrlEncoding,
-} from "./core_utils.js";
-import {
   clearPrimitiveCaches,
   Dict,
   isDict,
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Good idea; thanks!

PR #14083, opened 2021-09-26
AcroForm: Add support for ResetForm action

### LINKED ISSUE #12721: Add support for ResetForm action
The spec for this action:
https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#G11.2111850
An example of a form using this action:
https://www2.illinois.gov/rev/forms/reg/Documents/reg-1-mr.pdf

calixteman:
- it aims to fix #12721.
  - Thanks to PR #14023, we've now the fieldObjects in the annotation layer so we can easily map fields names on their id if needed.
  - Reset values in the storage, in the JS sandbox and in the visible html elements.


Snuffleupagus (commented):
Despite this being marked as a "draft", I figured it couldn't hurt to add a few comments based on cursory look. The overall approach looks reasonable though :-)

src/display/annotation_layer.js
```diff
@@ -556,6 +565,100 @@ class LinkAnnotationElement extends AnnotationElement {
...
+      if (otherClickAction) {
+        otherClickAction();
+      }
+
+      const {
+        fields: resetFormFields,
+        refs: resetFormRefs,
+        include,
+      } = resetForm;
+      if (resetFormFields.length !== 0 && !this._fieldsObjects) {
```
Snuffleupagus, line 583: Sorry, but I really don't understand what the `resetFormFields.length !== 0` check is doing here!?

Below it's e.g. sufficient for `resetFormRefs.length !== 0` in order to reach line https://github.com/mozilla/pdf.js/pull/14083/files#diff-d7b040092cf6be09a276b785ba03c46e79bc8342cf4ae4fb1bc139dea8513425R600 which will throw when the fieldObjects aren't defined.

Again, why not use a straight `if (!this._fieldObjects) {`-check for the early return?

calixteman:
@Snuffleupagus, when you'll have time, could you review this PR please ?

src/display/annotation_layer.js
```diff
@@ -556,6 +565,99 @@ class LinkAnnotationElement extends AnnotationElement {
...
+      warn(
+        `_bindResetFormAction - "resetForm" action not supported, ` +
+          "ensure that the `fieldObjects` parameter is provided."
+      );
+      return;
+    }
+
+    link.className = "internalLink";
+    const otherClickAction = link.onclick;
+    link.onclick = () => {
```
Snuffleupagus, line 580: Depending on the data, you may or may not have set either the href or className at this point. Furthermore, you also need to ensure that a dummy `onclick` is added in the fallback-case.

All-in-all, can you please make the following changes which I believe should cover all of the cases that could possibly happen here?

```suggestion
    const otherClickAction = link.onclick;
    if (!otherClickAction) {
      link.href = this.linkService.getAnchorUrl("");
    }
    link.className = "internalLink";
      
    if (!this._fieldObjects) {
      warn(
        `_bindResetFormAction - "resetForm" action not supported, ` +
          "ensure that the `fieldObjects` parameter is provided."
      );
      if (!otherClickAction) {
        link.onclick = () => false;
      }
      return;
    }
    
    link.onclick = () => {
```

src/display/annotation_layer.js
```diff
@@ -556,6 +565,99 @@ class LinkAnnotationElement extends AnnotationElement {
...
+        const fieldIds = new Set(resetFormRefs);
+        for (const fieldName of resetFormFields) {
+          const fields = this._fieldObjects[fieldName] || [];
+          for (const { id } of fields) {
+            fieldIds.add(id);
+          }
+        }
+        for (const fields of Object.values(this._fieldObjects)) {
+          for (const field of fields) {
+            if (include === fieldIds.has(field.id)) {
```
Snuffleupagus, line 602: Nit: Let's swap the order in this comparison, since the "constant" value should usually be on the right. (Note that a comparison such as `if (5 === someVar) { ... }` would actually be a linting failure.)

src/display/annotation_layer.js
```diff
@@ -1217,6 +1338,13 @@ class ChoiceWidgetAnnotationElement extends WidgetAnnotationElement {
       }
     }
 
+    selectElement.addEventListener("resetform", event => {
+      const defaultValue = this.data.defaultFieldValue;
+      Array.prototype.forEach.call(selectElement.options, option => {
```
Snuffleupagus, line 1343: Unless I'm mistaken, I believe that you should be able to iterate through the options with a `for...of` loop instead. Assuming that works, let's do that instead since it feels nicer than having to use Array.forEach in this way.

Snuffleupagus (approved):
r=me, with the comments addressed and passing tests; thank you!

PR #14092, opened 2021-09-30
[api-minor] Ensure that various URL-related options are applied in the `xfaLayer` too

Snuffleupagus:
Note how both the annotationLayer and the document outline will apply various URL-related options when creating the link-elements.
For consistency the `xfaLayer`-rendering should obviously use the same options, to ensure that the existing options are indeed applied to all URLs regardless of where they originate.

### DIFF
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -338,7 +338,7 @@ const LinkTarget = {
 
 /**
  * Adds various attributes (href, title, target, rel) to hyperlinks.
- * @param {HTMLLinkElement} link - The link element.
+ * @param {HTMLAnchorElement} link - The link element.
  * @param {ExternalLinkParameters} params
  */
 function addLinkAttributes(link, { url, target, rel, enabled = true } = {}) {
@@ -633,7 +633,6 @@ function getXfaPageViewport(xfaPage, { scale = 1, rotation = 0 }) {
 
 export {
   addLinkAttributes,
-  DEFAULT_LINK_REL,
   deprecated,
   DOMCanvasFactory,
   DOMCMapReaderFactory,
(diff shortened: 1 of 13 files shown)


Snuffleupagus:
Flagging multiple reviewers, since there's basically two parts to this:
 - @calixteman For the XFA changes.
 - @timvandermeij For the viewer changes.

calixteman (approved):
LGTM. Thanks a lot.

src/display/display_utils.js
```diff
@@ -338,7 +338,7 @@ const LinkTarget = {
 
 /**
  * Adds various attributes (href, title, target, rel) to hyperlinks.
- * @param {HTMLLinkElement} link - The link element.
+ * @param {HTMLAnchorElement} link - The link element.
  * @param {ExternalLinkParameters} params
  */
 function addLinkAttributes(link, { url, target, rel, enabled = true } = {}) {
```
Snuffleupagus, line 344: Considering how this helper function is now being used, with the other changes in the PR, we might be able to simply move it into the `web/` folder instead. @timvandermeij What do you think?

web/base_viewer.js
```diff
@@ -1358,6 +1358,7 @@ class BaseViewer {
    * @param {PDFPage} pdfPage
    * @param {AnnotationStorage} [annotationStorage] - Storage for annotation
    *   data in forms.
+   * @property {IPDFLinkService} linkService
```
timvandermeij, line 1361: This property is only used internally and is not part of the input/output and function signature, so I don't think we should mention it here (see https://github.com/mozilla/pdf.js/pull/14092/commits/bb9c905c5d6d11883888f328d288384e764f19fa#diff-f074a785f6c5ac2ecc4eff615f679c23dde5611034f651f85af67daa254c6fe4R137 where it's also not done).

Snuffleupagus: Whoops, this was a case of too much copy-and-pasting; fixed now :-)

timvandermeij (approved):

timvandermeij:
Thanks!

PR #14093, opened 2021-09-30
Update package-lock to lockfile version 2

nikolas:
https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,23053 @@
 {
   "name": "pdf.js",
+  "lockfileVersion": 2,
   "requires": true,
-  "lockfileVersion": 1,
+  "packages": {
+    "": {
+      "name": "pdf.js",
+      "hasInstallScript": true,
+      "license": "Apache-2.0",
+      "devDependencies": {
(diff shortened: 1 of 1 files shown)


Snuffleupagus (changes requested):
How was the updated `package-lock.json` file created/generated, and why *exactly* is this kind of change actually necessary?
Given the huge amount of lines in the diff, I'm not sure that we can accept the patch without a *simple way* of verifying the changes.

nikolas:
The `package-lock.json` was generated with an `npm install` command with npm >v7.0 (I have 7.21.1). You can do this yourself with any recent release of npm. The process I do is:

* `rm -rf node_modules package-lock.json`
* `npm install`

This change isn't necessary yet, but will be as new npm versions are released that aren't compatible with lockfile version 1.

Snuffleupagus:
Given that Node 12 is still "supported" until 2022-04-30, according to https://nodejs.org/en/about/releases/, and that it ships with `npm` version 6 we should probably hold off on these types of changes for the time being.

PR #14095, opened 2021-10-01
Replace a couple of `Array.prototype.forEach`-invocations with `for..of` instead

Snuffleupagus:
Given that `NodeList`s can be iterated using `for..of` we can use that instead, since it's a little bit nicer and easier to read than the `Array.prototype.forEach` format.


calixteman (approved):
Thanks.

PR #14096, opened 2021-10-02
Pre-render *one* additional page when spreadModes are enabled

Snuffleupagus:
*I've had this patch lying around locally for (at least) a year, and going through some old half-finished stuff figured that actually submitting it for review may be useful.*

Please note that we (obviously) don't want to unconditionally pre-render more than one page all the time, since that could very easily lead to overall worse performance in some documents.[1]
However, when spreadModes are enabled it does make sense to attempt to pre-render both of the pages of the next/previous spread.

---
[1] Since it may cause pre-rendering to unnecessarily compete for parsing resources, on the worker-thread, with "regular" rendering.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -1250,10 +1250,16 @@ class BaseViewer {
     const scrollAhead = this._isScrollModeHorizontal
       ? this.scroll.right
       : this.scroll.down;
+    const preRenderExtra =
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #14101, opened 2021-10-05
[Regression] Fix comb fields scrolling when the last character is entered (PR 14049 follow-up)

Snuffleupagus:
Note that PR #14049 removed this, since https://github.com/mozilla/pdf.js/pull/14049#discussion_r716245518 claimed that it's not necessary anymore. Unfortunately that, in my testing on Windows, actually re-introduced exactly the issue described in the comment; more specifically once the *last* character has been entered in the comb-field it's now again incorrectly scrolled (with the first character being invisible) until focus is lost.

This can be tested with e.g. `f1040.pdf`, see page 2, from the test-suite.

### DIFF
--- a/web/annotation_layer_builder.css
+++ b/web/annotation_layer_builder.css
@@ -147,6 +147,16 @@
   padding-right: 0;
 }
 
+.annotationLayer .textWidgetAnnotation input.comb:focus {
+  /*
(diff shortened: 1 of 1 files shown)


calixteman (approved):
Thanks

Nuurali (commented):
<removed>

PR #14106, opened 2021-10-06
Empty name is allowed in ISO 32000

calixteman:
- the exact sentence from the spec:
    "The token SOLIDUS (a slash followed by no regular characters) introduces a unique valid name defined by the empty sequence of characters."
  - so just remove the warning.


timvandermeij (approved):

timvandermeij:
Thank you!

PR #14108, opened 2021-10-07
For Annotations that define a closed area, make all of it toggle the PopupAnnotation (issue 14107)

Snuffleupagus:
For Circle, Square, and Polygon Annotations it's currently only possible to toggle the associated PopupAnnotation by clicking on its border. Depending on the border width, and also the current zoom-level in the viewer, that can make interacting with certain Annotations *practically* impossible (which is the case in issue 14107).
Hence, in order to improve this, change the "fill"-property of the SVG element in the annotationLayer to make the *entire* element part of the click/mouse-over target.

*Please note:* Given that this is a viewer-related issue, there's no simple way to test this as far as I can tell.

Fixes #14107


timvandermeij:
How does Adobe Reader handle this? Does it also make the entire area clickable, and also for the other annotation types you mentioned?

timvandermeij (approved):

timvandermeij:
Looks good!

PR #14112, opened 2021-10-08
Add a new Page scrolling mode (issue 2638, 8952, 10907)

Snuffleupagus:
This implements a new Page scrolling mode, essentially bringing (and extending) the functionality from `PDFSinglePageViewer` into the regular `PDFViewer`-class. Compared to `PDFSinglePageViewer`, which as its name suggests will only display one page at a time, in the `PDFViewer`-implementation this new Page scrolling mode also support spreadModes properly (somewhat similar to e.g. Adobe Reader).

Given the size and scope of these changes, I've tried to focus on implementing the basic functionality. Hence there's room for further clean-up and/or improvements, including e.g. simplifying the CSS/JS related to PresentationMode and implementing easier page-switching with the mouse-wheel/arrow-keys.

Fixes #2638
Fixes #8952
Fixes #10907

PR #14131, opened 2021-10-13
Ensure that pre-rendering works correctly with spreadModes at higher zoom levels

Snuffleupagus:
Having recently worked with this code, in PR #14096 (and indirectly in PR #14112), I happened to notice a pre-existing issue with spreadModes at higher zoom levels.
The `PDFRenderingQueue` code was written back when the viewer only supported "normal" vertical scrolling, and some edge-cases related to spreadModes are thus not perfectly supported. Depending on the zoom level, it's possible that there are "holes" in the currently visible page layout, and those pages will not be pre-rendered as you'd expect.

*Steps to reproduce:*

 0. Open the viewer, e.g. https://mozilla.github.io/pdf.js/web/viewer.html
 1. Enable vertical scrolling.
 2. Enable the ODD spreadMode.
 3. Scroll down, such that both pages 1 and 3 are visible.
 4. Zoom-in until *only* page 1 and 3 are visible.
 5. Open the devtools and, using the DOM Inspector, notice how page 2 is *not* being pre-rendered despite all surrounding pages being rendered.

---

data-loaded

### DIFF
--- a/web/pdf_rendering_queue.js
+++ b/web/pdf_rendering_queue.js
@@ -126,10 +126,24 @@ class PDFRenderingQueue {
         return view;
       }
     }
+    const firstId = visible.first.id,
+      lastId = visible.last.id;
+
+    // All the visible views have rendered; try to handle any "holes" in the
+    // page layout (can happen e.g. with spreadModes at higher zoom levels).
+    if (lastId - firstId > 1) {
+      for (let i = 0, ii = lastId - firstId; i <= ii; i++) {
+        const holeId = scrolledDown ? firstId + i : lastId - i,
+          holeView = views[holeId - 1];
+        if (!this.isViewFinished(holeView)) {
+          return holeView;
+        }
+      }
+    }
 
     // All the visible views have rendered; try to render next/previous page.
     // (IDs start at 1, so no need to add 1 when `scrolledDown === true`.)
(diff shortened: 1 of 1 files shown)


web/pdf_rendering_queue.js
```diff
@@ -126,10 +126,30 @@ class PDFRenderingQueue {
...
+    if (lastId - firstId > 1) {
+      const holeIds = new Set();
+      for (let i = 0, ii = lastId - firstId; i <= ii; i++) {
+        holeIds.add(scrolledDown ? firstId + i : lastId - i);
+      }
+      for (let i = 0; i < numVisible; i++) {
+        holeIds.delete(visibleViews[i].id);
+      }
+      for (const id of holeIds) {
+        const holeView = views[id - 1];
```
Snuffleupagus, line 143: Thinking about this some more, the multiple loops approach doesn't feel all that great.
Another possible approach would be the following, which uses the fact that already rendered pages will be skipped:
```javascript
      for (let i = 0, ii = lastId - firstId; i <= ii; i++) {
        const holeId = scrolledDown ? firstId + i : lastId - i,
          holeView = views[holeId - 1];
        if (holeView && !this.isViewFinished(holeView)) {
          return holeView;
        }
      }
```

timvandermeij (approved):

timvandermeij:
Seems to work just fine; thanks!

PR #14132, opened 2021-10-14
Convert examples/components/simpleviewer.js to await/async

### LINKED ISSUE #14128: Convert examples/learning/helloworld.html to await/async

The [simpleviewer.js](https://github.com/mozilla/pdf.js/blob/master/examples/components/simpleviewer.js) example currently uses the older style of `promise.then` and callbacks. Convert it to use an `async` function and `await` style.

janekotovich:
### Because

Simpleviewer.js uses chain of promises

### This pull request

Convert examples/components/simpleviewer.js to await/async

Closes: #14128

Checklist

Put an x in the boxes that apply

  

- [x] My commit is GPG signed.
- [ ] If applicable, I have modified or added tests which pass locally.
- [ ] I have added necessary documentation (if appropriate).
- [ ] I have verified that my changes render correctly in RTL (if appropriate).

### DIFF
--- a/examples/components/simpleviewer.js
+++ b/examples/components/simpleviewer.js
@@ -88,10 +88,12 @@ const loadingTask = pdfjsLib.getDocument({
   cMapPacked: CMAP_PACKED,
   enableXfa: ENABLE_XFA,
 });
-loadingTask.promise.then(function (pdfDocument) {
+
+(async function () {
   // Document loaded, specifying document for the viewer and
   // the (optional) linkService.
+  const pdfDocument = await loadingTask;
   pdfViewer.setDocument(pdfDocument);
 
   pdfLinkService.setDocument(pdfDocument, null);
-});
+})();


examples/components/simpleviewer.js
```diff
@@ -88,10 +88,12 @@ const loadingTask = pdfjsLib.getDocument({
   cMapPacked: CMAP_PACKED,
   enableXfa: ENABLE_XFA,
 });
-loadingTask.promise.then(function (pdfDocument) {
+
+(async function () {
   // Document loaded, specifying document for the viewer and
   // the (optional) linkService.
+  const pdfDocument = await loadingTask;
```
Snuffleupagus, line 95: Sorry, but this doesn't actually work!

Hence the *correct* change would be `const pdfDocument = await loadingTask.promise;`.

 - Running `gulp server`.
 - Opening http://localhost:8888/examples/components/simpleviewer.html in a browser and check that the example loads and works correctly without e.g. any errors printed in the console.

Snuffleupagus (changes requested):
Besides the inline comment, please remove the word "browserify" from your *commit message* since that's a name of a JavaScript framework (which isn't relevant to this example).

PR #14137, opened 2021-10-14
Convert examples/learning/helloworld64.html to await/async

### LINKED ISSUE #14122: Convert examples/learning/helloworld64.html to await/async

The [helloworld64.html](https://github.com/mozilla/pdf.js/blob/master/examples/learning/helloworld64.html) example currently uses the older style of `promise.then` and callbacks. Convert it to use an `async` function and `await` style.

mikemando:
### **Reference**
Convert examples/learning/helloworld64.html to await/async
### **Issue this pull request solves:**
closes #14122

### DIFF
--- a/examples/learning/helloworld64.html
+++ b/examples/learning/helloworld64.html
@@ -40,36 +40,36 @@
   // Opening PDF by passing its binary data as a string. It is still preferable
   // to use Uint8Array, but string or array-like structure will work too.
   var loadingTask = pdfjsLib.getDocument({ data: pdfData, });
-  loadingTask.promise.then(function(pdf) {
+  (async function() {
+    var pdf = await loadingTask.promise;
     // Fetch the first page.
-    pdf.getPage(1).then(function(page) {
-      var scale = 1.5;
-      var viewport = page.getViewport({ scale: scale, });
-      // Support HiDPI-screens.
-      var outputScale = window.devicePixelRatio || 1;
+    var page = await pdf.getPage(1);
+    var scale = 1.5;
+    var viewport = page.getViewport({ scale: scale, });
+    // Support HiDPI-screens.
+    var outputScale = window.devicePixelRatio || 1;
 
-      // Prepare canvas using PDF page dimensions.
-      var canvas = document.getElementById('the-canvas');
-      var context = canvas.getContext('2d');
+    // Prepare canvas using PDF page dimensions.
+    var canvas = document.getElementById('the-canvas');
+    var context = canvas.getContext('2d');
 
-      canvas.width = Math.floor(viewport.width * outputScale);
-      canvas.height = Math.floor(viewport.height * outputScale);
-      canvas.style.width = Math.floor(viewport.width) + "px";
-      canvas.style.height =  Math.floor(viewport.height) + "px";
+    canvas.width = Math.floor(viewport.width * outputScale);
+    canvas.height = Math.floor(viewport.height * outputScale);
+    canvas.style.width = Math.floor(viewport.width) + "px";
+    canvas.style.height =  Math.floor(viewport.height) + "px";
 
-      var transform = outputScale !== 1
-        ? [outputScale, 0, 0, outputScale, 0, 0]
-        : null;
+    var transform = outputScale !== 1
+      ? [outputScale, 0, 0, outputScale, 0, 0]
+      : null;
 
-      // Render PDF page into canvas context.
-      var renderContext = {
-        canvasContext: context,
-        transform: transform,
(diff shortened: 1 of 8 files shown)


examples/learning/helloworld64.html
```diff
@@ -40,36 +40,35 @@ <h1>'Hello, world!' example</h1>
   // Opening PDF by passing its binary data as a string. It is still preferable
   // to use Uint8Array, but string or array-like structure will work too.
   var loadingTask = pdfjsLib.getDocument({ data: pdfData, });
-  loadingTask.promise.then(function(pdf) {
+  (async function () {
```
SamyCookie, line 43: top level awaits are available since Firefox 89. Should this syntax be preferred ?

examples/learning/helloworld64.html
```diff
@@ -40,36 +40,35 @@ <h1>'Hello, world!' example</h1>
...
-    });
-  });
+    // Render PDF page into canvas context.
+    var renderContext = {
+      canvasContext: context,
+      transform: transform,
+      viewport: viewport,
+    };
+    page.render(renderContext);
+      })();
```
SamyCookie, line 71: unindent

examples/learning/helloworld64.html
```diff
@@ -40,36 +40,35 @@ <h1>'Hello, world!' example</h1>
   // Opening PDF by passing its binary data as a string. It is still preferable
   // to use Uint8Array, but string or array-like structure will work too.
   var loadingTask = pdfjsLib.getDocument({ data: pdfData, });
-  loadingTask.promise.then(function(pdf) {
+  (async function () {
+    var pdf = await loadingTask;
```
Snuffleupagus, line 44: Sorry, but this doesn't actually work!

Hence the *correct* change would be `const pdf = await loadingTask.promise;`.

 - Running `gulp server`.
 - Opening http://localhost:8888/examples/learning/helloworld64.html in a browser and check that the example loads and works correctly without e.g. any errors printed in the console.

examples/learning/helloworld64.html
```diff
@@ -40,36 +40,35 @@ <h1>'Hello, world!' example</h1>
...
+    canvas.width = Math.floor(viewport.width * outputScale);
+    canvas.height = Math.floor(viewport.height * outputScale);
+    canvas.style.width = Math.floor(viewport.width) + "px";
+    canvas.style.height = Math.floor(viewport.height) + "px";
 
-      var transform = outputScale !== 1
-        ? [outputScale, 0, 0, outputScale, 0, 0]
-        : null;
+    var transform =
+        outputScale !== 1 ? [outputScale, 0, 0, outputScale, 0, 0] : null;
```
Snuffleupagus, line 62: Please remove two spaces of indentation on this line.

examples/learning/helloworld64.html
```diff
@@ -40,36 +40,35 @@ <h1>'Hello, world!' example</h1>
...
-        viewport: viewport,
-      };
-      page.render(renderContext);
-    });
-  });
+    // Render PDF page into canvas context.
+    var renderContext = {
+      canvasContext: context,
+      transform: transform,
+      viewport: viewport,
```
Snuffleupagus, line 68: Please use the shorthand notation here instead, i.e.
```javascript
      transform,
      viewport,
```

Snuffleupagus (changes requested):
Besides the inline comments, please improve the *commit message* (and PR title) to say what you're actually changing since it's not entirely clear otherwise without looking at the code; e.g. "Convert examples/learning/helloworld64.html to use async/await".

examples/learning/helloworld64.html
mikemando: @SamyCookie thank you for the review. I would make the required change.

examples/learning/helloworld64.html
```diff
@@ -40,36 +40,36 @@ <h1>'Hello, world!' example</h1>
...
-      };
-      page.render(renderContext);
-    });
-  });
+    var page = await pdf.getPage(1);
+    var scale = 1.5;
+    var viewport = page.getViewport({ scale: scale, });
+    // Support HiDPI-screens.
+    var outputScale = window.devicePixelRatio || 1;
+
```
Snuffleupagus, line 51: Looking at the raw diff, it seems that you've left trailing spaces on this line (and also others below). Please go through the patch and remove all such occurrences (without removing the blank-lines themselves).

Snuffleupagus (changes requested):
Please also squash the commits, see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

PR #14138, opened 2021-10-14
 Convert examples/learning/helloworld.html to await/async 

### LINKED ISSUE #14121: Convert examples/learning/helloworld.html to await/async

The [helloworld.html](https://github.com/mozilla/pdf.js/blob/master/examples/learning/helloworld.html) example currently uses the older style of `promise.then` and callbacks. Convert it to use an `async` function and `await` style.

cathydev:
**Issue this pull request solves:**

closes #14121

### DIFF
--- a/examples/learning/helloworld.html
+++ b/examples/learning/helloworld.html
@@ -17,7 +17,7 @@
   // If absolute URL from the remote server is provided, configure the CORS
   // header on that server.
   //
-  var url = './helloworld.pdf';
+  const url = './helloworld.pdf';
 
   //
   // The workerSrc property shall be specified.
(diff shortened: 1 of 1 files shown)


examples/learning/helloworld.html
```diff
@@ -1,3 +1,4 @@
+
```
Snuffleupagus, line 1: Unrelated change, please revert.

examples/learning/helloworld.html
```diff
@@ -29,42 +30,42 @@ <h1>'Hello, world!' example</h1>
...
-    //
-    // Fetch the first page
-    //
-    pdf.getPage(1).then(function(page) {
-      var scale = 1.5;
-      var viewport = page.getViewport({ scale: scale, });
-      // Support HiDPI-screens.
-      var outputScale = window.devicePixelRatio || 1;
+  (async () => {
+  const pdf = await loadingTask;
```
Snuffleupagus, line 34: Sorry, but this doesn't actually work!

Hence the *correct* change would be `const pdfDocument = await loadingTask.promise;`.

 - Running `gulp server`.
 - Opening http://localhost:8888/examples/learning/helloworld.html in a browser and check that the example loads and works correctly without e.g. any errors printed in the console.

examples/learning/helloworld.html
```diff
@@ -29,42 +30,42 @@ <h1>'Hello, world!' example</h1>
...
-        : null;
+  //
+  // Render PDF page into canvas context
+  //
+  var renderContext = {
+    canvasContext: context,
+    transform,
+    viewport,
+  };
+  page.render(renderContext);
```
Snuffleupagus, line 66: Note that this *entire* block of code should be indented with *two* more spaces.

Snuffleupagus (changes requested):
Besides the inline comments, please improve the *commit message* to say what you're actually changing since it's not entirely clear otherwise without looking at the code; e.g. "Convert examples/learning/helloworld.html to use async/await".

catherinemds:
Thanks, I had a little problem with npm install and that was the reason why of the error but everything is working now

examples/learning/helloworld.html
```diff
@@ -28,43 +28,44 @@ <h1>'Hello, world!' example</h1>
...
+    canvas.style.width = Math.floor(viewport.width) + "px";
+    canvas.style.height = Math.floor(viewport.height) + "px";
 
-      var transform = outputScale !== 1
-        ? [outputScale, 0, 0, outputScale, 0, 0]
-        : null;
+    const transform =
+      outputScale !== 1 
+      ? [outputScale, 0, 0, outputScale, 0, 0] 
+      : null;
```
Snuffleupagus, line 57: As mentioned in https://github.com/mozilla/pdf.js/pull/14138#discussion_r729062437, please ensure that this reads as follows (and without any trailing spaces):
```javascript
const transform = outputScale !== 1
  ? [outputScale, 0, 0, outputScale, 0, 0]
  : null;
```

examples/learning/helloworld.html
```diff
@@ -75,4 +76,4 @@ <h2>JavaScript code:</h2>
       document.getElementById('script').text;
 </script>
 </body>
-</html>
+</html>
```
Snuffleupagus, line 79: Please revert this change, i.e. re-add the line-break at the end here.

package.json
```diff
@@ -69,5 +69,8 @@
     "type": "git",
     "url": "git://github.com/mozilla/pdf.js.git"
   },
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "dependencies": {
+    "pdfjs-dist": "file:build/dist"
+  }
```
Snuffleupagus, line 75: Please revert all changes to the `package-lock.json` and `package.json` files.

examples/learning/helloworld.html
```diff
@@ -28,43 +28,44 @@ <h1>'Hello, world!' example</h1>
   //
   // Asynchronous download PDF
   //
-  var loadingTask = pdfjsLib.getDocument(url);
-  loadingTask.promise.then(function(pdf) {
+  const loadingTask = pdfjsLib.getDocument(url);
+  (async () => {
+    const pdf = await loadingTask.promise;;
```
Snuffleupagus, line 33: One semi-colon too many here.

Snuffleupagus (changes requested):
This looks better now, but there's still a few small things to fix here.
Also, in the commit message the same sentence is repeated more than once; please ensure that the final commit message reads only "Convert examples/learning/helloworld.html to await/async".

Snuffleupagus (approved):
r=me; thanks for the patch!

PR #14141, opened 2021-10-14
Convert examples/text-only/pdf2svg.js to await/async 

adenicole:
This is my first pull request.

### DIFF
--- a/examples/text-only/pdf2svg.js
+++ b/examples/text-only/pdf2svg.js
@@ -48,19 +48,16 @@ function buildSVG(viewport, textContent) {
   return svg;
 }
 
-function pageLoaded() {
+async function pageLoaded() {
   // Loading document and page text content
   const loadingTask = pdfjsLib.getDocument({ url: PDF_PATH });
-  loadingTask.promise.then(function (pdfDocument) {
-    pdfDocument.getPage(PAGE_NUMBER).then(function (page) {
-      const viewport = page.getViewport({ scale: PAGE_SCALE });
-      page.getTextContent().then(function (textContent) {
-        // building SVG and adding that to the DOM
-        const svg = buildSVG(viewport, textContent);
-        document.getElementById("pageContainer").appendChild(svg);
-      });
-    });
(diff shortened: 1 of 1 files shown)


calixteman (changes requested):
You're adding the function `pageLoaded` but it isn't called anywhere so it's useless.

calixteman:
> > You're adding the function `pageLoaded` but it isn't called anywhere so it's useless.
> 
> Isn't the problem actually that the patch adds a _new_ function, rather than simply replacing the old one (as intended)?
> 

Oh! I've been confused because of the github UI.

> Furthermore, was this successfully tested locally? There's also a number of linting failures, please remember to run `gulp lint` and fix all the reported problems. Finally, please squash the commits; see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

Snuffleupagus (changes requested):
As mentioned in https://github.com/mozilla/pdf.js/pull/14141#issuecomment-943489131:
 - Please confirm that you've successfully tested this locally.
 - Please squash the commits, since we don't allow either fixup (or merge) commits to land; see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

Snuffleupagus:
The *second* sentence of the commit message, or at least the "and corrected lint errors"-part, doesn't seem appropriate here since it suggests that there were lint errors in the old code which isn't actually the case.

adenicole:
> The _second_ sentence of the commit message, or at least the "and corrected lint errors"-part, doesn't seem appropriate here since it suggests that there were lint errors in the old code which isn't actually the case.

Oh, my bad! Fixed the commit message. 🙂

Snuffleupagus (approved):
*For future reference:* As outlined in https://github.com/mozilla/pdf.js/wiki/Contributing, you want to use feature branches rather than working directly against the "master" branch (since that makes working on more than thing at a time really difficult and causes e.g. rebasing to become more difficult).

r=me, thank you.

PR #14143, opened 2021-10-14
docs: Fix grammatical error

raghav-wd:
Fixed a small grammatical error in docs.


Snuffleupagus (approved):
*For future reference:* Please update an existing PR by e.g. force-pushing (since this was #14140) rather than submitting a new one, since it's much easier to lose context and overlook prior review comments like this!

r=me; thank you.

PR #14156, opened 2021-10-17
Add support for modern ECMAScript `class` features

Snuffleupagus:
With ESLint 8 we should now finally be able to start using modern `class` features, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

However, while both ESLint and Acorn now support this, it unfortunately turns out that Escodegen (which we use during building) still lack the necessary support. Looking at https://github.com/estools/escodegen there's not been any updates since last year, and there's also open PRs adding support for these new `class` features.

To avoid blocking usage of these `class` features in the PDF.js code-base, in particular *private* fields/methods, this patch thus proposes that we (hopefully temporarily) switch to an `escodegen` fork that has the necessary support; please see https://www.npmjs.com/package/@javascript-obfuscator/escodegen

While I have no reason to doubt the security of the `escodegen` fork, this patch nonetheless pins the version number. Furthermore, I've also diffed the output of the two `.js`-files in this forked package against the original files without finding anything that looks immediately "dangerous".


Snuffleupagus:
```diff
--- C:\Users\Jonas\Git\pdf.js\node_modules\estraverse\estraverse.js
+++ C:\Users\Jonas\Git\pdf.js\node_modules\@javascript-obfuscator\estraverse\estraverse.js
@@ -29,23 +29,11 @@
     'use strict';
 
     var Syntax,
-        isArray,
         VisitorOption,
         VisitorKeys,
-        objectCreate,
-        objectKeys,
         BREAK,
         SKIP,
         REMOVE;
-
-    function ignoreJSHintError() { }
-
-    isArray = Array.isArray;
-    if (!isArray) {
-        isArray = function isArray(array) {
-            return Object.prototype.toString.call(array) === '[object Array]';
-        };
-    }
 
     function deepCopy(obj) {
         var ret = {}, key, val;
@@ -62,17 +50,6 @@
         return ret;
     }
 
-    function shallowCopy(obj) {
-        var ret = {}, key;
-        for (key in obj) {
-            if (obj.hasOwnProperty(key)) {
-                ret[key] = obj[key];
-            }
-        }
-        return ret;
-    }
-    ignoreJSHintError(shallowCopy);
-
     // based on LLVM libc++ upper_bound / lower_bound
     // MIT License
 
@@ -93,52 +70,6 @@
             }
         }
         return i;
-    }
-
-    function lowerBound(array, func) {
-        var diff, len, i, current;
-
-        len = array.length;
-        i = 0;
-
-        while (len) {
-            diff = len >>> 1;
-            current = i + diff;
-            if (func(array[current])) {
-                i = current + 1;
-                len -= diff + 1;
-            } else {
-                len = diff;
-            }
-        }
-        return i;
-    }
-    ignoreJSHintError(lowerBound);
-
-    objectCreate = Object.create || (function () {
-        function F() { }
-
-        return function (o) {
-            F.prototype = o;
-            return new F();
-        };
-    })();
-
-    objectKeys = Object.keys || function (o) {
-        var keys = [], key;
-        for (key in o) {
-            keys.push(key);
-        }
-        return keys;
-    };
-
-    function extend(to, from) {
-        var keys = objectKeys(from), key, i, len;
-        for (i = 0, len = keys.length; i < len; i += 1) {
-            key = keys[i];
-            to[key] = from[key];
-        }
-        return to;
     }
 
     Syntax = {
@@ -153,6 +84,7 @@
         BreakStatement: 'BreakStatement',
         CallExpression: 'CallExpression',
         CatchClause: 'CatchClause',
+        ChainExpression: 'ChainExpression',
         ClassBody: 'ClassBody',
         ClassDeclaration: 'ClassDeclaration',
         ClassExpression: 'ClassExpression',
@@ -177,6 +109,7 @@
         GeneratorExpression: 'GeneratorExpression',  // CAUTION: It's deferred to ES7.
         Identifier: 'Identifier',
         IfStatement: 'IfStatement',
+        ImportExpression: 'ImportExpression',
         ImportDeclaration: 'ImportDeclaration',
         ImportDefaultSpecifier: 'ImportDefaultSpecifier',
         ImportNamespaceSpecifier: 'ImportNamespaceSpecifier',
@@ -191,8 +124,10 @@
         NewExpression: 'NewExpression',
         ObjectExpression: 'ObjectExpression',
         ObjectPattern: 'ObjectPattern',
+        PrivateIdentifier: 'PrivateIdentifier',
         Program: 'Program',
         Property: 'Property',
+        PropertyDefinition: 'PropertyDefinition',
         RestElement: 'RestElement',
         ReturnStatement: 'ReturnStatement',
         SequenceExpression: 'SequenceExpression',
@@ -227,6 +162,7 @@
         BreakStatement: ['label'],
         CallExpression: ['callee', 'arguments'],
         CatchClause: ['param', 'body'],
+        ChainExpression: ['expression'],
         ClassBody: ['body'],
         ClassDeclaration: ['id', 'superClass', 'body'],
         ClassExpression: ['id', 'superClass', 'body'],
@@ -251,6 +187,7 @@
         GeneratorExpression: ['blocks', 'filter', 'body'],  // CAUTION: It's deferred to ES7.
         Identifier: [],
         IfStatement: ['test', 'consequent', 'alternate'],
+        ImportExpression: ['source'],
         ImportDeclaration: ['specifiers', 'source'],
         ImportDefaultSpecifier: ['local'],
         ImportNamespaceSpecifier: ['local'],
@@ -265,8 +202,10 @@
         NewExpression: ['callee', 'arguments'],
         ObjectExpression: ['properties'],
         ObjectPattern: ['properties'],
+        PrivateIdentifier: [],
         Program: ['body'],
         Property: ['key', 'value'],
+        PropertyDefinition: ['key', 'value'],
         RestElement: [ 'argument' ],
         ReturnStatement: ['argument'],
         SequenceExpression: ['expressions'],
@@ -310,7 +249,7 @@
     };
 
     Reference.prototype.remove = function remove() {
-        if (isArray(this.parent)) {
+        if (Array.isArray(this.parent)) {
             this.parent.splice(this.key, 1);
             return true;
         } else {
@@ -334,7 +273,7 @@
         var i, iz, j, jz, result, element;
 
         function addToPath(result, path) {
-            if (isArray(path)) {
+            if (Array.isArray(path)) {
                 for (j = 0, jz = path.length; j < jz; ++j) {
                     result.push(path[j]);
                 }
@@ -434,14 +373,14 @@
         this.__state = null;
         this.__fallback = null;
         if (visitor.fallback === 'iteration') {
-            this.__fallback = objectKeys;
+            this.__fallback = Object.keys;
         } else if (typeof visitor.fallback === 'function') {
             this.__fallback = visitor.fallback;
         }
 
         this.__keys = VisitorKeys;
         if (visitor.keys) {
-            this.__keys = extend(objectCreate(this.__keys), visitor.keys);
+            this.__keys = Object.assign(Object.create(this.__keys), visitor.keys);
         }
     };
 
@@ -454,6 +393,15 @@
 
     function isProperty(nodeType, key) {
         return (nodeType === Syntax.ObjectExpression || nodeType === Syntax.ObjectPattern) && 'properties' === key;
+    }
+  
+    function candidateExistsInLeaveList(leavelist, candidate) {
+        for (var i = leavelist.length - 1; i >= 0; --i) {
+            if (leavelist[i].node === candidate) {
+                return true;
+            }
+        }
+        return false;
     }
 
     Controller.prototype.traverse = function traverse(root, visitor) {
@@ -530,12 +478,17 @@
                         continue;
                     }
 
-                    if (isArray(candidate)) {
+                    if (Array.isArray(candidate)) {
                         current2 = candidate.length;
                         while ((current2 -= 1) >= 0) {
                             if (!candidate[current2]) {
                                 continue;
                             }
+
+                            if (candidateExistsInLeaveList(leavelist, candidate[current2])) {
+                              continue;
+                            }
+
                             if (isProperty(nodeType, candidates[current])) {
                                 element = new Element(candidate[current2], [key, current2], 'Property', null);
                             } else if (isNode(candidate[current2])) {
@@ -546,6 +499,10 @@
                             worklist.push(element);
                         }
                     } else if (isNode(candidate)) {
+                        if (candidateExistsInLeaveList(leavelist, candidate)) {
+                          continue;
+                        }
+
                         worklist.push(new Element(candidate, key, null, null));
                     }
                 }
@@ -684,7 +641,7 @@
                     continue;
                 }
 
-                if (isArray(candidate)) {
+                if (Array.isArray(candidate)) {
                     current2 = candidate.length;
                     while ((current2 -= 1) >= 0) {
                         if (!candidate[current2]) {
@@ -834,7 +791,6 @@
         return tree;
     }
 
-    exports.version = require('./package.json').version;
     exports.Syntax = Syntax;
     exports.traverse = traverse;
     exports.replace = replace;
```

package.json
```diff
@@ -15,6 +15,7 @@
     "dommatrix": "^0.0.6",
     "es-module-shims": "^1.3.0",
     "escodegen": "^2.0.0",
```
timvandermeij, line 17: Shouldn't we remove the old `escodegen`? As far as I can tell the preprocessor is the only place that uses it, unless I missed something of course.

Snuffleupagus: My hope was that the original `escodegen` package will actually be updated (and actively maintained) in the future, such that we don't need to use the fork. Hence I figured that leaving this here would make it easier to spot when/if there's a new release.

If you don't think that's reason enough to keep this, and I can certainly understand that point of view, I can of course remove this package.

timvandermeij (approved):

PR #14159, opened 2021-10-18
Convert examples/node/pdf2svg.js to await/async #14125

### LINKED ISSUE #14125: Convert examples/node/pdf2svg.js to await/async

The [node/pdf2svg.js](https://github.com/mozilla/pdf.js/blob/master/examples/node/pdf2svg.js) example currently uses the older style of `promise.then` and callbacks. Convert it to use an `async` function and `await` style.

stbensonimoh:
Because
`pdf2svg.js` uses chain of promises

### This pull request
Convert examples/components/simpleviewer.js to await/async

Closes: #14125

Checklist

Put an x in the boxes that apply

- [x]  My commit is GPG signed.
- [ ]  If applicable, I have modified or added tests which pass locally.
- [ ]  I have added necessary documentation (if appropriate).
- [ ]  I have verified that my changes render correctly in RTL (if appropriate).

### DIFF
--- a/examples/node/pdf2svg.js
+++ b/examples/node/pdf2svg.js
@@ -40,7 +40,7 @@ try {
 // Dumps svg outputs to a folder called svgdump
 function getFilePathForPage(pageNum) {
   const name = path.basename(pdfPath, path.extname(pdfPath));
-  return path.join(outputDirectory, name + "-" + pageNum + ".svg");
+  return path.join(outputDirectory, `${name}-${pageNum}.svg`);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


examples/node/pdf2svg.js
```diff
@@ -70,73 +70,70 @@ ReadableSVGStream.prototype._read = function () {
...
+    await readableSvgStream.once("error");
+    await writableStream.once("error");
+    await writableStream.once("finish");
+    await readableSvgStream.pipe(writableStream);
+  } catch (err) {
     readableSvgStream = null; // Explicitly null because of v8 bug 6512.
     writableStream.end();
     throw err;
-  });
+  }
```
Snuffleupagus, line 87: The `once` method is Node.js functionality which doesn't return a Promise, hence this unfortunately cannot work; please revert all changes in the function!

examples/node/pdf2svg.js
```diff
@@ -70,73 +70,70 @@ ReadableSVGStream.prototype._read = function () {
...
-    console.log("Number of Pages: " + numPages);
+const loadingTask = async function () {
+  const loadTask = pdfjsLib.getDocument({
+    data,
+    cMapUrl: CMAP_URL,
+    cMapPacked: CMAP_PACKED,
+    fontExtraProperties: true,
+  });
+  const doc = await loadTask.promise;
+  const numPages = await doc.numPages;
```
Snuffleupagus, line 100: This getter doesn't return a Promise, hence please revert this change; see https://github.com/mozilla/pdf.js/blob/bbb64369f1ffa118000ae5b3be1611cd0aec8003/src/display/api.js#L716-L721

examples/node/pdf2svg.js
```diff
@@ -70,73 +70,70 @@ ReadableSVGStream.prototype._read = function () {
...
-    const numPages = doc.numPages;
-    console.log("# Document Loaded");
-    console.log("Number of Pages: " + numPages);
+const loadingTask = async function () {
+  const loadTask = pdfjsLib.getDocument({
+    data,
+    cMapUrl: CMAP_URL,
+    cMapPacked: CMAP_PACKED,
+    fontExtraProperties: true,
+  });
```
Snuffleupagus, line 98: Please just use an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) rather than a named function, i.e.
```javascript
const loadingTask = pdfjsLib.getDocument({
  data,
  cMapUrl: CMAP_URL,
  cMapPacked: CMAP_PACKED,
  fontExtraProperties: true,
});
(async function () {
  try {
    const doc = await loadTask.promise;
    ...
})();
```

examples/node/pdf2svg.js
```diff
@@ -70,73 +70,70 @@ ReadableSVGStream.prototype._read = function () {
...
+  const doc = await loadTask.promise;
+  const numPages = await doc.numPages;
+  console.log("# Document Loaded");
+  console.log("Number of Pages: " + numPages);
+  console.log();
+
+  let finalResult;
+  const loadPage = async function (pageNum) {
+    const page = await doc.getPage(pageNum);
```
Snuffleupagus, line 107: Please get rid of this function, and simply loop over the pages instead; e.g. something like
```javascript
for (let pageNum = 1; i <= numPages; i++) {
  const page = await doc.getPage(pageNum);
  ...
```

examples/node/pdf2svg.js
```diff
@@ -70,73 +70,70 @@ ReadableSVGStream.prototype._read = function () {
...
-            );
-          });
-        });
-      });
-    };
+    const opList = await page.getOperatorList();
+    const svgGfx = new pdfjsLib.SVGGraphics(page.commonObjs, page.objs);
+    svgGfx.embedFonts = true;
+    const svg = await svgGfx.getSVG(opList, viewport);
+    let result;
```
Snuffleupagus, line 117: With the change above, the `result` variable should no longer be needed.

Snuffleupagus (changes requested):
Unfortunately this patch breaks the example, and running it no longer generates any `svg`-files at all. *For future reference:* Please don't submit a pull request unless you've successfully tested it locally.

To test this locally:
 1. Run `gulp dist-install`.
 2. cd into the `examples/node/` directory.
 3. Run `node pdf2svg.js`.
 4. Repeat step 3 after making changes to the example.

When addressing the comments, please keep https://github.com/mozilla/pdf.js/wiki/Squashing-Commits in mind.

examples/node/pdf2svg.js
```diff
@@ -87,56 +87,36 @@ function writeSvgToFile(svgElement, filePath) {
   });
 }
 
-// Will be using promises to load document, pages and misc data instead of
-// callback.
+// Will be using async/await to load document, pages and misc data
```
Snuffleupagus, line 90: Missing a period at the end of this sentence.

examples/node/pdf2svg.js
```diff
@@ -87,56 +87,36 @@ function writeSvgToFile(svgElement, filePath) {
...
-  .then(function (doc) {
-    const numPages = doc.numPages;
-    console.log("# Document Loaded");
-    console.log("Number of Pages: " + numPages);
-    console.log();
+(async function () {
+  const doc = await loadingTask.promise;
+  const numPages = doc.numPages;
+  console.log("# Document Loaded");
+  console.log("Number of Pages: " + numPages);
```
Snuffleupagus, line 101: Here, and elsewhere in this file, let's change these types of strings (that contains parameter values) to template strings instead; in this case it'll become:
```javascript
  console.log(`Number of Pages: ${numPages}`);
```

examples/node/pdf2svg.js
```diff
@@ -87,56 +87,36 @@ function writeSvgToFile(svgElement, filePath) {
...
-        console.log("# Page " + pageNum);
-        const viewport = page.getViewport({ scale: 1.0 });
-        console.log("Size: " + viewport.width + "x" + viewport.height);
-        console.log();
+  for (let pageNum = 1; pageNum <= numPages; pageNum++) {
+    try {
+      const page = await doc.getPage(pageNum);
+      console.log(`# Page ${pageNum}`);
+      const viewport = page.getViewport({ scale: 1.0 });
+      console.log(`Size: ${viewport.width} x ${viewport.height}`);
```
Snuffleupagus, line 109: This is introducing adding additional spaces in the string, why?
The following is equivalent to the old code:
```javascript
console.log(`Size: ${viewport.width}x${viewport.height}`);
```

Snuffleupagus (approved):

examples/node/pdf2svg.js
stbensonimoh: @Snuffleupagus Sorry, I missed this. It has been fixed.

PR #14160, opened 2021-10-18
Fix pattern handling regression in `SVGGraphics` (PR 13770 follow-up)

Snuffleupagus:
While the FAQ clearly lists the SVG back-end as unsupported, see https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#backends, I suppose that small/simple regressions still makes sense to fix.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #14162, opened 2021-10-18
Indent the stepper on save/restore.

brendandahl:
Makes it easier to debug and visualize the graphics
state changes in the debugger.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -1701,6 +1701,7 @@ html[dir="rtl"] #documentPropertiesOverlay .row > * {
 }
 #PDFBug table {
   font-size: 10px;
+  white-space: pre;
 }
 #PDFBug table.showText {
   border-collapse: collapse;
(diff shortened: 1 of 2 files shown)


web/viewer.css
```diff
@@ -1701,6 +1701,8 @@ html[dir="rtl"] #documentPropertiesOverlay .row > * {
 }
 #PDFBug table {
   font-size: 10px;
+  font-family: monospace;
```
Snuffleupagus, line 1704: I'm finding that the use of a monospace font makes it harder to read the text comfortably, at least on Windows.

Do we *really* need to change the font-family here, since it seems to me that the indentation is still noticeable even without that?
Another possible idea, to improve readability of the monospace font, could be to increase the font-size slightly? (Although that would require even more horizontal scrolling during reading.)

timvandermeij (approved):

PR #14166, opened 2021-10-19
Ignore Square/Circle-annnotations with a zero borderWidth when creating a fallback appearance stream (issue 14164)

Snuffleupagus:
Trying to render these Annotation-types, when the borderWidth is `0`, causes a "hairline" border to appear. If these Annotations included an appearance stream, as they are supposed to, this wouldn't have happened and the simplest solution here seem to be to just ignore these particular Annotations.

Fixes #14164


calixteman (approved):
LGTM, thanks for doing that.

PR #14174, opened 2021-10-21
Prevent double-rendering borders for PushButton-annotations (PR 14083 follow-up)

Snuffleupagus:
With ResetForm-action support added in PR #14083, there's a regression in the `issue12716` test-case. More specifically the border around the "Clear Form"-link is now rendered *twice*, once in the canvas via the appearance-stream and once in the annotationLayer via the border-data.
This looks slightly weird, and was most likely not intended, which is why this patch suggests that we ignore the border in the annotationLayer when an appearance-stream exists.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -421,7 +421,7 @@ class AnnotationElement {
 }
 
 class LinkAnnotationElement extends AnnotationElement {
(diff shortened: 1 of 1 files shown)


calixteman (approved):
LGTM, thank you.

PR #14181, opened 2021-10-24
Converted simpleviewer.js to await/async

### LINKED ISSUE #14128: Convert examples/learning/helloworld.html to await/async

PinRathod:
Closes #14128  Converted simpleviewer.js to await/async

### DIFF
--- a/examples/components/simpleviewer.js
+++ b/examples/components/simpleviewer.js
@@ -88,10 +88,11 @@ const loadingTask = pdfjsLib.getDocument({
   cMapPacked: CMAP_PACKED,
   enableXfa: ENABLE_XFA,
 });
-loadingTask.promise.then(function (pdfDocument) {
+(async function () {
+  const pdfDocument = await loadingTask.promise;
   // Document loaded, specifying document for the viewer and
   // the (optional) linkService.
   pdfViewer.setDocument(pdfDocument);
 
   pdfLinkService.setDocument(pdfDocument, null);
-});
+})();


Snuffleupagus (changes requested):
Unfortunately this breaks the example; please make sure that you actually test your patch before opening a pull request.

To test this locally:
1. Run `gulp dist-install`.
2. Run `gulp server`.
3. Open http://localhost:8888/examples/components/simpleviewer.html in a browser.

Furthermore, I don't think that this will pass linting either; please remember to always run `gulp lint` locally and fix the reported errors.
Finally, please don't include *unrelated* changes in patches (i.e. the `package-lock.json` file).

Snuffleupagus (changes requested):
You should *not* be removing the `package-lock.json`/`package.json` files from the repo, just from the patch itself.
Also, please follow https://github.com/mozilla/pdf.js/wiki/Squashing-Commits when addressing review comments.

Snuffleupagus (changes requested):
Your commit message has unrelated content, which doesn't make sense (i.e. the "This reverts commit c001b17." sentence).

Snuffleupagus (approved):
r=me, thanks.

PR #14182, opened 2021-10-24
[api-minor] Support rich content in markup annotation

### LINKED ISSUE #2966: Support for Rich Text annotations

### LINKED ISSUE #13915: Incorrect rendering of annotations

Configuration:
- Web browser and its version: Chrome Version 95.0.4612.0 (Official Build) canary (64-bit)
- Operating system and its version: Windows 10 Home 20H2 19042.1165
- PDF.js version: 2.9.147
- Is a browser extension: Yes

1. Create a pdf with box annotation with formatting s.a. italics, bold, etc
2.

1.Date and time should be set
2.Formatting should be respected (e.g., bold, italics, subscript, superscript)

mouse-over:
pdfjs_OK

clicked:
pdfjs_OK_full

1.Date and time aren't displayed properly
2.Formatting not respected

pdfjs_BAD

calixteman:
- use the xfa parser but in the xhtml namespace.

Fix #2966
Fix #13915


Snuffleupagus (commented):
In the existing `annotation_layer.js` code, we have *a bunch* of checks [along these lines](https://github.com/mozilla/pdf.js/blob/1ab9a6e36e84389775e5f3d6faebdb6b327b8d62/src/display/annotation_layer.js#L678-L679); should all of those perhaps have a `parameters.data.richText`-case added now?

Also, should an explicitly defined `PopupAnnotation` support rich-text; i.e. should we add `richText: this.data.richText,` at end of the parameters listed here? https://github.com/mozilla/pdf.js/blob/1ab9a6e36e84389775e5f3d6faebdb6b327b8d62/src/display/annotation_layer.js#L1578-L1585

Snuffleupagus (changes requested):

Snuffleupagus (changes requested):

src/core/xfa/xhtml.js
```diff
@@ -109,6 +111,7 @@ const StyleMapping = new Map([
 
 const spacesRegExp = /\s+/g;
 const crlfRegExp = /[\r\n]+/g;
+const crlfForRichTextRegExp = /\r\n|\r/g;
```
Snuffleupagus, line 114: Nit: Unless I'm missing something, I *think* that you could just use `/\r\n?/g` instead.

Snuffleupagus (commented):
There also seem to be some existing unit-tests that need to be updated now.

calixteman:
The "regressions" are due to the fact we're using the xhtml now instead of the contents.
For example in `annotation-highlight.pdf`, the xml is:
```xml
RC = "<?xml version=\"1.0\"?><body xmlns=\"http://www.w3.org/1999/xhtml\" xmlns:xfa=\"http://www.xfa.org/schema/xfa-data/1.0/\" xfa:APIVersion=\"Acrobat:11.0.0\" xfa:spec=\"2.0.2\"><p dir=\"ltr\"><span style=\"text-align:left;font-size:12pt;font-style:normal;font-weight:normal;color:#000000;font-family:Arial\">Highlight content</span></p></body>"
```
which specifies a font and a font size.
I wonder if we should strip out the font-family (because Arial for example won't be available everywhere).
Should we use for example Liberation to render all the annotations in order to provide the same rendering for everybody ?

Snuffleupagus:
> I wonder if we should strip out the font-family (because Arial for example won't be available everywhere).

That might be the simplest solution, and would ensure a more consistent handling of all PopupAnnotations regardless if they're using Contents or RichText.

> Should we use for example Liberation to render all the annotations in order to provide the same rendering for everybody ?

Unless absolutely necessary, I'd advise *against* that for overall simplicity since not all PDF.js implementations will be bundling the standard font files and we'd thus need to ensure proper fallback behaviour in that case.

src/core/annotation.js
```diff
@@ -1098,6 +1099,18 @@ class MarkupAnnotation extends Annotation {
...
+    if (dict.has("RC")) {
+      const htmlText = XFAFactory.getRichTextAsHtml(dict.get("RC"));
+      if (
+        htmlText &&
+        (!this.data.contentsObj ||
+          !this.data.contentsObj.str ||
+          this.data.contentsObj.str === htmlText.text)
+      ) {
+        this.data.richText = htmlText.html;
+      }
```
Snuffleupagus, line 1112: Let's improve the API, for a non-viewer consumer, by always returning the `richText` and instead decide on what to use in the display-layer; i.e. let's instead do:
```javascript
      this.data.richText = XFAFactory.getRichTextAsHtml(dict.get("RC"));
```

src/core/annotation.js
```diff
@@ -2545,6 +2558,18 @@ class PopupAnnotation extends Annotation {
...
+    if (parentItem.has("RC")) {
+      const htmlText = XFAFactory.getRichTextAsHtml(parentItem.get("RC"));
+      if (
+        htmlText &&
+        (!this.data.contentsObj ||
+          !this.data.contentsObj.str ||
+          this.data.contentsObj.str === htmlText.text)
+      ) {
+        this.data.richText = htmlText.html;
+      }
```
Snuffleupagus, line 2571: *Same as the comment just above.*

src/core/xfa/html_utils.js
```diff
@@ -606,10 +606,17 @@ function setPara(node, nodeStyle, value) {
 }
 
 function setFontFamily(xfaFont, node, fontFinder, style) {
-  const name = stripQuotes(xfaFont.typeface);
-  const typeface = fontFinder.find(name);
+  if (!fontFinder) {
+    // The font cannot be found in the pdf so use the
+    // default one.
```
Snuffleupagus, line 611: Nit: It looks like this comment fits on one line.

src/core/xfa/factory.js
```diff
@@ -106,6 +113,45 @@ class XFAFactory {
...
+      if (attributes) {
+        if (attributes.class) {
+          attributes.class = attributes.class.filter(
+            attr => !attr.startsWith("xfa")
+          );
+        }
+        attributes.dir = "auto";
+      }
+
+      const text = root[$text]();
```
Snuffleupagus, line 147: Maybe calling this `str` instead, since that'd be slightly more consistent on the API/display-layer side (see comment in the `src/display/annotation_layer.js` file).

src/display/annotation_layer.js
```diff
@@ -1664,8 +1672,17 @@ class PopupElement {
       popup.appendChild(modificationDate);
     }
 
-    const contents = this._formatContents(this.contentsObj);
-    popup.appendChild(contents);
+    if (this.richText) {
+      XfaLayer.render({
+        xfa: this.richText,
```
Snuffleupagus, line 1677: Following the `src/core/annotation.js` changes suggested above:
```javascript
    if (
      this.richText &&
      (!this.contentsObj.str || this.contentsObj.str === this.richText.str)
    ) {
      XfaLayer.render({
        xfa: this.richText.html,
```

Snuffleupagus (commented):
I was just about to land this, when it occurred to me that we can improve the API itself (for a non-viewer use-case); please see the inline comments.

src/display/annotation_layer.js
```diff
@@ -676,7 +678,8 @@ class TextAnnotationElement extends AnnotationElement {
     const isRenderable = !!(
       parameters.data.hasPopup ||
       parameters.data.titleObj?.str ||
-      parameters.data.contentsObj?.str
+      parameters.data.contentsObj?.str ||
+      parameters.data.richText
```
Snuffleupagus, line 682: Given my other comments, I suppose all of these newly added checks should perhaps be `parameters.data.richText?.str` instead?

Snuffleupagus (approved):
r=me, thanks for implementing this!

PR #14189, opened 2021-10-25
[api-minor] Implement securityHandler in the scripting API (bug 1731578)

### LINKED ISSUE #14146: JS - Implement securityHandler in the scripting API
The bug has been reported in bugzilla:
https://bugzilla.mozilla.org/show_bug.cgi?id=1731578

For reference:
https://opensource.adobe.com/dc-acrobat-sdk-docs/acrobatsdk/pdfs/acrobatsdk_jsapiref.pdf#page=256

The value can be found in `Encrypt >> Filter`:
https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=64

It's something we can get here:
https://github.com/mozilla/pdf.js/blob/fc56a781d327422351248f8f894fd6efe771f118/web/generic_scripting.js#L22
pass the value to the sandbox and then return it:
https://github.com/mozilla/pdf.js/blob/fc56a781d327422351248f8f894fd6efe771f118/src/scripting_api/doc.js#L525

janekotovich:
### Because

Missing security handler throws an error on pdf.js

The bug has been reported in bugzilla:
https://bugzilla.mozilla.org/show_bug.cgi?id=1731578

### This pull request

Add Security Handler to pdf.js
Add puppeteer tests

Closes: #14146


Snuffleupagus (changes requested):

Snuffleupagus (commented):
It looks like there's a merge-commit accidentally included here, please remove than one and squash the commits.

---

> However still cannot figure out why after building and everything it gives me error.

Where is the error thrown, in the terminal or in the browser?
Also, can you please provide the *exact* error message here?

calixteman (approved):
LGTM, thank you for doing that.

PR #14201, opened 2021-10-27
Use the correct border-style for Annotations, when a dash array is specified (bug 1219400)

Snuffleupagus:
Even though we cannot use the dash array in the display layer, at least ensure that we use the correct border-style.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1219400


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

/botio makeref

PR #14204, opened 2021-10-28
Remove the `shadowViewer` used with Page scrolling

Snuffleupagus:
The only reason for using a `DocumentFragment` in the first place, originally added in PR #8724, was to prevent errors in the `PDFPageView`-constructor. However, we should be able to simply make its `container`-option *optional* instead, since it's not being used for anything else in the class.

Note that pre-rendering still works correctly in my testing, and given that the `BaseViewer` keeps references to all `PDFPageView`-instances (via its `_pages` Array) it also shouldn't be possible to "lose" any pages/canvases this way.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -148,6 +148,8 @@ function isSameScale(oldScale, newScale) {
  * Simple viewer control to display PDF content/pages.
  */
 class BaseViewer {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Looks good; thanks!

PR #14207, opened 2021-10-29
JS - Avoid a popup to ask for specific version of Acrobat

### LINKED ISSUE #14206: JS - Avoid a popup to ask for specific version of Acrobat
Follow up with issue: https://github.com/mozilla/pdf.js/issues/14146
Merged pull req: https://github.com/mozilla/pdf.js/pull/14189

https://bug1708641.bmoattachments.org/attachment.cgi?id=9219459

- This popup appears because JS is enabled;
- And because the pdf contains some unsupported features (e.g. XFA);

Configuration:
- Firefox Nightly & Chrome
- macOS Big Sur 11.6

1. Run gulp server
2. Open attached pdf

No popup on open

What went wrong?
error

janekotovich:
### Because

Popup keep asking regarding specific version of Acrobat (Spanish and version 5.0 or greater).
That happens because:
- JS in pdf is enabled;
- And the pdf contains some unsupported features (e.g. XFA);

By searching through pdf source code I have found that there is an embedded JS in pdf throws error & alert, if
```
If ( app.formVersion = undefined || app.formVersion < 5.0 ) {
Alert(“These documents are only functional in Spanish versions of Acrobat equal to or greater than 5.0”);
}
```

Regarding documents in https://helpx.adobe.com/livecycle/help/mobile-forms/scripting-support.html for Scripting Support,

Closes: #14206

good

### DIFF
--- a/src/scripting_api/app.js
+++ b/src/scripting_api/app.js
@@ -21,8 +21,8 @@ import { Thermometer } from "./thermometer.js";
 
 const VIEWER_TYPE = "PDF.js";
 const VIEWER_VARIATION = "Full";
-const VIEWER_VERSION = "10.0";
-const FORMS_VERSION = undefined;
+const VIEWER_VERSION = 21.00720099;
+const FORMS_VERSION = 21.00720099;
 
 class App extends PDFObject {
   constructor(data) {


Snuffleupagus:
@calixteman Do we have a good/simple way of testing that an `alert` box isn't displayed on document loading, or can we land this as-is?

/botio integrationtest

calixteman:
> @calixteman Do we have a good/simple way of testing that an `alert` box isn't displayed on document loading, or can we land this as-is?
>

Some actions are ran when the document is open:
https://github.com/mozilla/pdf.js/blob/ec1633c33c8b1d87b1d3fb987ec2ac885123d0bd/src/scripting_api/doc.js#L100-L114 
and `alert` is defined here:
https://github.com/mozilla/pdf.js/blob/ec1633c33c8b1d87b1d3fb987ec2ac885123d0bd/src/scripting_api/app.js#L429
and the `app` object has access to the `doc`.
So it's easy to add a flag in the `doc` in order to know if we are in the "open" process and disable `alert` when this flag is true.

We can disable the popups at loading time but if one contains an information of interest (e.g. "Don't forget to fill whatever before sending the form"), the user will miss it.
I think that a well-designed form shouldn't have any intrusive popup at loading but it should warn before printing, saving, ... that something is missing.

So I think we could land that stuff as it is and I can ask to Product/UX/UI people what they think about disabling all popups at loading. @Snuffleupagus does that work for you ?

Snuffleupagus:
> So I think we could land that stuff as it is and I can ask to Product/UX/UI people what they think about disabling all popups at loading. @Snuffleupagus does that work for you ?

Sure, that sounds good to me!

calixteman (approved):
LGTM, thank you for doing this.

PR #14213, opened 2021-10-31
Tweak the Bidi-detection heuristics for very short RTL strings (issue 11656)

### LINKED ISSUE #11656: Parenthesis with hebrew text not rendering in the correct way
[test.pdf](https://github.com/mozilla/pdf.js/files/4281200/test.pdf)

Configuration:
- Web browser and its version: Google Chrome 80.0.3987.122 
- Operating system and its version: Windows 10 64 bit
- PDF.js version: 2.0.0
- Is a browser extension: No

1. Open hebrew pdf file
2. Set locale to he-IL (#locale=he-IL)
3. Set textLayer to visible (#textLayer=visible)
4. The parenthesis rendering wrong.

like can you see the rendering of the parenthesis wrong 

There is no link

Snuffleupagus:
Very short strings can narrowly miss the existing Bidi-detection threshold, leading to incorrect text-selection and copying behaviour.

In my testing, neither Adobe Reader or PDFium seem to handle copying "correctly" for this document. Hence it's not entirely clear to me that we actually want to fix this, since tweaking these heuristics can *obviously* cause regressions elsewhere (and our test coverage for RTL-text isn't exactly great).


src/core/bidi.js
```diff
@@ -158,15 +158,16 @@ function bidi(str, startLevel = -1, vertical = false) {
...
+  //   unless the string is very short.
   // - If more than 30% chars are rtl then string is primarily rtl
   if (numBidi === 0) {
     isLTR = true;
     return createBidiText(str, isLTR);
   }
 
   if (startLevel === -1) {
-    if (numBidi / strLength < 0.3) {
+    if (numBidi / strLength < 0.3 && strLength > 4) {
```
timvandermeij, line 170: I might be missing something obvious here, but wouldn't this mean that _any_ LTR string of length <= 4 would be considered RTL? Let's say we have the string "foo": `numBidi` is 0, so `numBidi / strLength` = 0 / 3 = 0, and `0 < 0.3` holds. However, `3 > 4` doesn't hold, so we end up in the `else` branch and set it to RTL.

Snuffleupagus: > Let's say we have the string "foo": `numBidi` is 0, so `numBidi / strLength` = 0 / 3 = 0, and `0 < 0.3` holds. However, `3 > 4` doesn't hold, so we end up in the `else` branch and set it to RTL.

Assuming we actually want this patch, I could also add a new unit-test with an LTR-only string to check exactly the situation described above.

timvandermeij (approved):
This seems reasonable, minimizes any risk and it also increases the test coverage for RTL text, so I'm in favor of doing this. r=me with passing tests; thank you!

src/core/bidi.js
Snuffleupagus: @timvandermeij I've added additional unit-tests, please check if those look good; thank you!

PR #14214, opened 2021-10-31
Change annotation border with correct width

### LINKED ISSUE #14203: Annotation borders don't have the correct width
[pdfcomment.pdf](https://github.com/mozilla/pdf.js/files/7428963/pdfcomment.pdf)

Configuration:

1. Have a look on red lines on page 10
2. They're smaller in Firefox compared to how they're rendered in Acrobat

image

image

PinRathod:
CLoses #14203 

Annotation borders don't have the correct width in firefox compared to as rendered in Acrobat. Changes made in annotation border width property.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -884,7 +884,7 @@ class Annotation {
  */
 class AnnotationBorderStyle {
   constructor() {
-    this.width = 1;
+    this.width = 3;
     this.style = AnnotationBorderStyleType.SOLID;
     this.dashArray = [3];
     this.horizontalCornerRadius = 0;
--- a/test/pdfs/issue14203.pdf.link
+++ b/test/pdfs/issue14203.pdf.link
@@ -1,0 +1,1 @@
+https://github.com/mozilla/pdf.js/files/7428963/pdfcomment.pdf
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -6012,4 +6012,13 @@
       "type": "eq",
       "lastPage": 1
     }
+    { "id": "issue14203",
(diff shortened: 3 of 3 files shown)


timvandermeij (changes requested):
Unfortunately this is not a correct fix since the default border width should really be 1. The problem in the issue's PDF file is elsewhere, and you'll find it when examining the annotation data contained within the file using a PDF browser like https://brendandahl.github.io/pdf.js.utils/browser/ and comparing the source code that parses it with the Adobe specification at https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=394&zoom=auto,-215,327.

Snuffleupagus:
Unfortunately this patch is too far from a correct/working solution, and it's also pretty clear that you've not successfully tested it locally (given the lint failure).
Looking at the patch, it seems that you've not fully debugged/understood the underlying problem since (as also mentioned above) just changing the default value is going to break *so many* PDF documents and isn't even correct for the referenced issue.
Furthermore, when making these sort of changes, you *must* follow https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing carefully.
Also, you (once again) haven't created the branch correctly; see https://github.com/mozilla/pdf.js/wiki/Contributing#2-create-feature-branch

Finally, note that when adding a reference test you cannot mix enableXfa and annotations and given that size of the test-case you should use firstPage and lastPage to only the relevant page(s).
And please make sure to write good commit messages, since a one-liner isn't really sufficient in general.

Closing this PR, since it's unfortunately too far from working.

Issue #14130, opened 2021-10-13
XFA subform with occur min=0 and no bound data displaying


JDziurlaj:
Attach (recommended) or Link to PDF file here:
[occurmin0.pdf](https://github.com/mozilla/pdf.js/files/7340277/occurmin0.pdf)

Configuration:
- Web browser and its version: 95.0a1 (2021-10-13) (64-bit)
- Operating system and its version: Windows 10

Steps to reproduce the problem:
1. Open form in pdf.js

What is the expected behavior? (add screenshot)
Form is blank (subform `nomin`  does not display)
What went wrong? (add screenshot)
![image](https://user-images.githubusercontent.com/7104454/137184675-d48027c0-1f17-4faa-bc97-b45814b38e36.png)

subform `nomin` displays, even though it's subform is set to `<occur max="-1" min="0" />`

PR #14216, opened 2021-11-01
Update xfa_layer_builder.css issue(14130)

AbdallahWario:
solves issue #14130 where "no-min you should not see this displays" instead of a blank page;
An  xfa subform will have a class "xfaRich",in case of a blank xfa subform,it lacks inline 'display' property while a non-empty one has an inline css 'display' property.it's external diplay being set to "none" solves the issue since a non empty one will have an inline property which will have precedence over the external hence displays.

### DIFF
--- a/web/xfa_layer_builder.css
+++ b/web/xfa_layer_builder.css
@@ -240,6 +240,7 @@
   white-space: pre-wrap;
   width: 100%;
   height: 100%;
+  display:none;
 }
 
 .xfaImage {


Snuffleupagus:
Sorry, but simply hiding *every* `.xfaRich` element cannot possibly be a correct solution since it'll break *so much more* than it would potentially fix. Furthermore, this patch unfortunately suggests that you've not really debugged/understood *exactly* what's wrong in the referenced issues in particular *and* that you've also not run any tests locally (since even linting fails).

Please make sure that you follow *all* of https://github.com/mozilla/pdf.js/wiki/Contributing carefully, paying special attention to https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing and also https://github.com/mozilla/pdf.js/wiki/Contributing#2-create-feature-branch

Finally, please make sure to write a *good commit message*, since "Update xfa_layer_builder.css" isn't helpful/useful at all. The first line of the commit message should be a summary of the actual changes, and the following lines should provide additional details explaining the problem that's being solved and why the patch is implemented like it is.

All-in-all, closing this PR since it's unfortunately too far from working.

PR #14217, opened 2021-11-01
[Firefox] Handle errors if loading failed before the "supportsRangedLoading" message was sent (bug 1732141)

Snuffleupagus:
This is a follow-up to PR #10675, since there I completely overlooked that we also need to handle the case where a PDF document has *failed* to load when the "supportsRangedLoading" message is sent to the viewer.


brendandahl:
Looks good thanks!

PR #14218, opened 2021-11-02
XFA subform with occur min=0 and no bound data displaying.

### LINKED ISSUE #14130: XFA subform with occur min=0 and no bound data displaying

janekotovich:
### Because

Subform nomin displays, even though it's subform is set to <occur max="-1" min="0" />

### This pull request

Added condition to catch empty forms by check
```
if subform.name === "nomin" && subform.occur.min === 0
```
And remove this empty forms via pushing it to a uselessNode array which later is removed.

Closes: #14130


janekotovich:
I didn't add a test to manifest.json because the one in issue is not a link but attached file.
I ran locally test by adding occurmin0.pdf to pdfs,
on master branch:
- gulp makeref 

on subform_min_0 branch:
- gulp browsertest
- gulp unitest
- gulp lint

All went well but please lemme know if you would prefer me adding test to manifest.json and to test/pdfs.
Thanks!

Snuffleupagus:
> I didn't add a test to manifest.json because pdf in issue is not a link but attached file.

Please note that whenever feasible, we'd like all patches to include tests (since that helps avoid regressions later on).

For those cases when a PDF document is *attached* to an issue, you've got two options for creating the test:
 - Normally you'd treat this just as a *linked* test-case, getting the URL by right-clicking on the link and using the "Copy Link" alternative in the context menu.
 - If the attached PDF document is small/simple, without any personally identifiable information and without anything that looks like copyrighted content, you can actually include the file directly in the PDF.js repository.

---

In this particular case, it looks like option two is actually feasible!
To add a standard, non-linked, test-case you can e.g. follow this procedure: 
 - Download that file, naming it `issue14130.pdf` in this case, and place it in the `test/pdfs/` folder locally.
 - From the root of your local PDF.js repository, run `node test/add_test.js test/pdfs/issue14130.pdf` which will take care of adding an entry to the `test/pdfs/.gitignore` file and also update the `test/test_manifest.json` file.
 - In this case, since you want an XFA-test, you then need to *manually* add a `"enableXfa": true`-line to the entry in the `test/test_manifest.json` file.
 - Finally, you just `git add` the files mentioned above (i.e. `test/pdfs/.gitignore`, `test/pdfs/issue14130.pdf`, and `test/test_manifest.json`) and amend your existing commit.

janekotovich:
@Snuffleupagus  yep no problem at all, will do in a bit 😊 
Thanks, I already knew how to add it, just wasn't sure regarding file attaching / link, so decided to ask here. Will update commit and push alltogether!

janekotovich:
@Snuffleupagus oh sorry, I just noticed you've mentioned amend, but I rebased (as my previous ones). Hope its okay!

Snuffleupagus:
> oh sorry, I just noticed you've mentioned amend, but I rebased (as my previous ones). Hope its okay!

I really don't think it matters *exactly* how you do it, as long as the end result looks correct :-)

src/core/xfa/bind.js
```diff
@@ -643,6 +643,9 @@ class Binder {
                 : dataNode[$namespaceId];
             match = child[$data] = new XmlObject(nsId, child.name);
             if (this.emptyMerge) {
+              if (child.name === "nomin" && child.occur.min === 0) {
```
calixteman, line 646: Could you point out where in the specifications you found that an element called `nomin` is useless ?
Afaik, an element with a child `<occur min="0" ...>` is useless when no data can be bound in this element.

calixteman: Have a look on:
https://github.com/mozilla/pdf.js/blob/6a15973a1b26288f981d971fab1f54bb4c98d8be/src/core/xfa/bind.js#L666
The node is useless when `match` is null (nothing has been found in the data tree which matches with the child from the template tree) and `min <= 0` (i.e. `min===0`).
But because of a bug, with the pdf in the issue, we never hit this path.
So there are several possibilities:
 - we've a `match` when we shouldn't;
 - we've `min > 0` when we shouldn't;
 - we never go there because there is a `continue` or a `return` somewhere.

src/core/xfa/bind.js
```diff
@@ -637,6 +637,9 @@ class Binder {
           if (!match) {
             // We're in matchTemplate mode so create a node in data to reflect
             // what we've in template.
+            if (child.occur.min === 0) {
+              uselessNodes.push(child);
```
calixteman, line 641: - The comment here doesn't apply to the added code so you should move it
- you should add a comment (you can quote the specification or add a link to it) to explain why the node is useless.
- once we know that the child is useless, you can then skip the code after.

janekotovich: 👍🏽

src/core/xfa/bind.js
```diff
@@ -635,6 +635,13 @@ class Binder {
             /* skipConsumed = */ this.emptyMerge
           ).next().value;
           if (!match) {
+            // If there is no match (no data) and occur.min === 0
+            // container is entirely excluded.
+            // https://www.pdfa.org/norm-refs/XFA-3_3.pdf
```
Snuffleupagus, line 640: That document is over 1500 pages long, please link directly to the relevant section instead; I'm guessing that https://www.pdfa.org/norm-refs/XFA-3_3.pdf#G23.1904923 might be correct here?

src/core/xfa/bind.js
```diff
@@ -635,6 +635,13 @@ class Binder {
             /* skipConsumed = */ this.emptyMerge
           ).next().value;
           if (!match) {
+            // If there is no match (no data) and occur.min === 0
+            // container is entirely excluded.
```
Snuffleupagus, line 639: There seem to be a couple of words missing in the middle, how about this instead:
```javascript
            // If there is no match (no data) and `occur.min === 0` then
            // the container is entirely excluded.
```

src/core/xfa/bind.js
calixteman: I think this section is a bit more specific:
https://www.pdfa.org/norm-refs/XFA-3_3.pdf#G12.1428332

src/core/xfa/bind.js
```diff
@@ -635,6 +635,13 @@ class Binder {
             /* skipConsumed = */ this.emptyMerge
           ).next().value;
           if (!match) {
+            // If there is no match (no data) and `occur.min === 0` then
+            // the container is entirely excluded.
+            // https://www.pdfa.org/norm-refs/XFA-3_3.pdf#G12.1428332
+            if (child.occur.min === 0) {
```
calixteman, line 641: I missed that... but `occur` could be null here.
We already get the `min` info here: https://github.com/mozilla/pdf.js/blob/611627f5a1226785a7a952c0fdb6e4e8eba4a2e4/src/core/xfa/bind.js#L560
so you should use this value instead.

calixteman (approved):
LGTM, thank you for doing this.

PR #14219, opened 2021-11-02
Let `getVisibleElements` return a Set containing the visible element `id`s

Snuffleupagus:
Note how in `PDFPageViewBuffer.resize` we're manually iterating through the visible pages in order to build a Set of the visible page `id`s. By instead moving the building of this Set into the `getVisibleElements` helper function, as part of the existing parsing, this code becomes *ever so slightly* more efficient.

Furthermore, more direct access to the visible page `id`s also come in handy in other parts of the viewer as well.
In the `BaseViewer.isPageVisible` method we no longer need to loop through the visible pages, but can instead directly check if the pageNumber is visible.
In the `PDFRenderingQueue.getHighestPriority` method, when checking for "holes" in the page layout, we can also avoid some unnecessary look-ups this way.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -104,22 +104,19 @@ function PDFPageViewBuffer(size) {
       data.shift().destroy();
     }
   };
+
   /**
-   * After calling resize, the size of the buffer will be newSize. The optional
-   * parameter pagesToKeep is, if present, an array of pages to push to the back
-   * of the buffer, delaying their destruction. The size of pagesToKeep has no
-   * impact on the final size of the buffer; if pagesToKeep has length larger
-   * than newSize, some of those pages will be destroyed anyway.
+   * After calling resize, the size of the buffer will be `newSize`.
+   * The optional parameter `idsToKeep` is, if present, a Set of page-ids to
+   * push to the back of the buffer, delaying their destruction. The size of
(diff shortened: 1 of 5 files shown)


web/base_viewer.js
```diff
@@ -1273,16 +1272,14 @@ class BaseViewer {
...
-      return view.id === pageNumber;
-    });
+    return this._getVisiblePages().ids.has(pageNumber);
   }
 
   /**
    * @param {number} pageNumber
    */
   isPageCached(pageNumber) {
-    if (!this.pdfDocument || !this._buffer) {
```
Snuffleupagus, line 1285: *Slightly unrelated to the rest of the patch, but:* Note that `this._buffer` will always be defined whenever `this.pdfDocument` is set.

web/base_viewer.js
```diff
@@ -1296,9 +1293,6 @@ class BaseViewer {
       return false;
     }
     const pageView = this._pages[pageNumber - 1];
-    if (!pageView) {
-      return false;
-    }
```
Snuffleupagus, line 1301: *Slightly unrelated to the rest of the patch, but:* Given the existing validation just above it shouldn't be possible for `pageView` to be undefined here, and in any case passing `undefined` to the `PDFPageViewBuffer.has`-method won't break anything.

timvandermeij (approved):
Looks like a good improvement. r=me once rebased; thanks!

PR #14221, opened 2021-11-02
Use `BaseViewer.previousPage` more in the default viewer (PR 12870 follow-up)

Snuffleupagus:
I missed this one spot in PR #12870, when converting the other cases in the "keydown" event handler. However, given that it only matters in PresentationMode and/or when "page-fit" zooming is enabled, this oversight shouldn't have had any user-observable impact (but we should fix it nonetheless).

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -3057,9 +3057,8 @@ function webViewerKeyDown(evt) {
         ) {
           break;
         }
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #14225, opened 2021-11-03
Avoid doing unnecessary checks, when pre-rendering page layouts with "holes" (PR 14131 follow-up)

Snuffleupagus:
*Sometimes I'll hopefully learn to optimize my code directly when writing it, rather than having to do multiple clean-up passes; sorry about the churn here!*

For most page layouts there won't be any "holes" in the visible pages (or thumbnails), and in those cases it'd obviously be preferable not having to repeat any checks of already rendered pages.
Rather than only checking the "distance" between the first/last pages, we can instead compare the theoretical number of pages (between first/last) with the actually visible number of pages instead. This way, we're able to better detect the "holes"-case and can skip unnecessary parsing in the common case.

### DIFF
--- a/web/pdf_rendering_queue.js
+++ b/web/pdf_rendering_queue.js
@@ -115,13 +115,13 @@ class PDFRenderingQueue {
      * 2. if last scrolled down, the page after the visible pages, or
      *    if last scrolled up, the page before the visible pages
      */
-    const visibleViews = visible.views;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
No worries; thanks for the follow-up!

PR #14228, opened 2021-11-03
Reset path bounding box tracking when starting a new path.

brendandahl:
Starting a new path will wipe out any of the current subpaths in the
current graphics state, so we should reset the min/maxes.

This makes a number of the bounding boxes smaller and reduces the number
of composed pixels. For the smask tests in the corpus, the number of
composed pixesl goes from 19,872,109 to 19,676,905. The difference is much
larger on other PDFs though.


Snuffleupagus (approved):

PR #14229, opened 2021-11-03
Add an easy way to log to the terminal during browser tests.

brendandahl:
On the main thread call `driver.log` and the message will output in the
terminal with the pdf id and the message.

I've been using this a lot when trying to find certain PDFs or logging
stats.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -466,6 +466,24 @@ var Driver = (function DriverClosure() {
       }, this.delay);
     },
 
+    /**
+     * A debugging tool to log to the terminal while tests are running.
+     * XXX: This isn't currently referenced, but it's useful for debugging so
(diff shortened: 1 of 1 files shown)


Snuffleupagus (approved):

timvandermeij:
Unless I'm missing something here, I don't see any other code calling this. If it's intended to only be called manually, I would recommend we introduce a comment that says that to avoid this being deleted in the future during e.g., refactoring, since otherwise it's easy to think that it's dead code and can be removed.

brendandahl:
Yeah, it's not being used currently. I've been using it when debugging. I'll add a comment.

timvandermeij (approved):

timvandermeij:
Looks good now; thanks!

PR #14230, opened 2021-11-03
Create shading patterns the size of the current path. (bug 1722807)

brendandahl:
Previously, when we created a shading pattern canvas we created it
as the same size as the page. This was good for caching if the same
pattern was used over and over again, but when lots of different
shadings are created that caused us to create many full page
canvases.

Instead of creating the full page canvses, instead create the canvas
as the same size as the current path bounding box. This reduces memory
consumption by a lot since most paths are pretty small. Also, in real world
PDFs it's rare for a shading (non shading fill) to be reused over and over again.
Bug 1721949 is an example where the same pattern is reused and it will be slightly
slower than before.

### DIFF
--- a/src/display/canvas.js
+++ b/src/display/canvas.js
@@ -27,7 +27,11 @@ import {
   Util,
   warn,
 } from "../shared/util.js";
-import { getShadingPattern, TilingPattern } from "./pattern_helper.js";
+import {
+  getShadingPattern,
+  PathType,
+  TilingPattern,
+} from "./pattern_helper.js";
 import { PixelsPerInch } from "./display_utils.js";
 
 // <canvas> contexts store most of the state we need natively.
@@ -37,10 +41,6 @@ const MIN_FONT_SIZE = 16;
 // Maximum font size that would be used during canvas fillText operations.
 const MAX_FONT_SIZE = 100;
 const MAX_GROUP_SIZE = 4096;
-
-// This value comes from sampling a few PDFs that re-use patterns, there doesn't
-// seem to be any that benefit from caching more than 2 patterns.
-const MAX_CACHED_CANVAS_PATTERNS = 2;
 
 // Defines the time the `executeOperatorList`-method is going to be executing
 // before it stops and shedules a continue of execution.
@@ -366,46 +366,6 @@ class CachedCanvases {
   }
 }
 
-/**
- * Least recently used cache implemented with a JS Map. JS Map keys are ordered
- * by last insertion.
- */
-class LRUCache {
-  constructor(maxSize = 0) {
-    this._cache = new Map();
-    this._maxSize = maxSize;
-  }
-
-  has(key) {
-    return this._cache.has(key);
-  }
-
-  get(key) {
-    if (this._cache.has(key)) {
-      // Delete and set the value so it's moved to the end of the map iteration.
-      const value = this._cache.get(key);
-      this._cache.delete(key);
-      this._cache.set(key, value);
-    }
-    return this._cache.get(key);
-  }
-
-  set(key, value) {
-    if (this._maxSize <= 0) {
-      return;
-    }
-    if (this._cache.size + 1 > this._maxSize) {
-      // Delete the least recently used.
-      this._cache.delete(this._cache.keys().next().value);
-    }
-    this._cache.set(key, value);
-  }
-
-  clear() {
-    this._cache.clear();
-  }
-}
-
 function compileType3Glyph(imgData) {
   const POINT_TO_PROCESS_LIMIT = 1000;
   const POINT_TYPES = new Uint8Array([
@@ -639,8 +599,23 @@ class CanvasExtraState {
     this.updatePathMinMax(transform, box[2], box[3]);
   }
 
-  getPathBoundingBox() {
-    return [this.minX, this.minY, this.maxX, this.maxY];
+  getPathBoundingBox(pathType = PathType.FILL, transform = null) {
+    const box = [this.minX, this.minY, this.maxX, this.maxY];
+    if (pathType === PathType.STROKE) {
+      if (!transform) {
+        unreachable("Stroke bounding box must include transform.");
+      }
+      // Stroked paths can be outside of the path bounding box by 1/2 the line
+      // width.
+      const scale = Util.singularValueDecompose2dScale(transform);
+      const xStrokePad = (scale[0] * this.lineWidth) / 2;
+      const yStrokePad = (scale[1] * this.lineWidth) / 2;
+      box[0] -= xStrokePad;
+      box[1] -= yStrokePad;
+      box[2] += xStrokePad;
+      box[3] += yStrokePad;
+    }
+    return box;
   }
 
   updateClipFromPath() {
@@ -656,8 +631,11 @@ class CanvasExtraState {
     this.maxY = 0;
   }
 
-  getClippedPathBoundingBox() {
-    return Util.intersect(this.clipBox, this.getPathBoundingBox());
+  getClippedPathBoundingBox(pathType = PathType.FILL, transform = null) {
+    return Util.intersect(
+      this.clipBox,
+      this.getPathBoundingBox(pathType, transform)
+    );
   }
 }
 
@@ -1121,7 +1099,6 @@ class CanvasGraphics {
     this.markedContentStack = [];
     this.optionalContentConfig = optionalContentConfig;
     this.cachedCanvases = new CachedCanvases(this.canvasFactory);
-    this.cachedCanvasPatterns = new LRUCache(MAX_CACHED_CANVAS_PATTERNS);
     this.cachedPatterns = new Map();
     if (canvasCtx) {
       // NOTE: if mozCurrentTransform is polyfilled, then the current state of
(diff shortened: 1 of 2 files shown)


brendandahl:
Some notes from performance testing:
1) bug1722807_page2 is MUCH faster going from 62,066ms to 330ms in Firefox and the several GB memory increase is gone
2) bug1721949 is a big slower, but a pretty small difference

-- Grouped By pdf, stat, browser --
pdf                | stat         | browser | Count | Baseline(ms) | Current(ms) |    +/- |     %  | Result(P<.05)
------------------ | ------------ | ------- | ----- | ------------ | ----------- | ------ | ------ | -------------
bug1721949         | Overall      | chrome  |    50 |          136 |         168 |     32 |  23.27 |        slower
bug1721949         | Overall      | firefox |    50 |          175 |         185 |     10 |   5.87 |              
bug1721949         | Page Request | chrome  |    50 |            2 |           2 |      0 |   4.82 |              
bug1721949         | Page Request | firefox |    50 |            2 |           2 |      0 |  -8.57 |              
bug1721949         | Rendering    | chrome  |    50 |          135 |         166 |     32 |  23.56 |        slower
bug1721949         | Rendering    | firefox |    50 |          172 |         183 |     10 |   6.06 |              
bug1722807_page2   | Overall      | chrome  |     5 |          868 |         308 |   -561 | -64.56 |              
bug1722807_page2   | Overall      | firefox |     5 |        62069 |         332 | -61737 | -99.47 |        faster
bug1722807_page2   | Page Request | chrome  |     5 |            6 |           4 |     -2 | -27.59 |              
bug1722807_page2   | Page Request | firefox |     5 |            3 |           2 |     -1 | -38.46 |              
bug1722807_page2   | Rendering    | chrome  |     5 |          863 |         303 |   -559 | -64.83 |              
bug1722807_page2   | Rendering    | firefox |     5 |        62066 |         330 | -61736 | -99.47 |        faster
issue13325_reduced | Overall      | chrome  |    50 |           16 |          16 |      0 |   0.00 |              
issue13325_reduced | Overall      | firefox |    50 |           24 |          11 |    -13 | -54.85 |        faster
issue13325_reduced | Page Request | chrome  |    50 |            1 |           1 |      0 |   9.26 |              
issue13325_reduced | Page Request | firefox |    50 |            1 |           1 |      0 | -37.04 |              
issue13325_reduced | Rendering    | chrome  |    50 |           15 |          15 |      0 |  -0.81 |              
issue13325_reduced | Rendering    | firefox |    50 |           23 |          10 |    -13 | -55.87 |        faster

src/display/canvas.js
```diff
@@ -639,8 +599,24 @@ class CanvasExtraState {
     this.updatePathMinMax(transform, box[2], box[3]);
   }
 
-  getPathBoundingBox() {
-    return [this.minX, this.minY, this.maxX, this.maxY];
+  getPathBoundingBox(pathType = PathType.FILL, transform = null) {
+    const box = [this.minX, this.minY, this.maxX, this.maxY];
+    if (pathType === PathType.STROKE) {
+      if (!transform) {
+        throw new Error("Stroke bounding box must include transform.");
```
Snuffleupagus, line 606: Perhaps just use `unreachable` instead, rather than throwing manually, since I'm assuming that this code-path shouldn't be hit unless an implementation error was made.

src/display/canvas.js
```diff
@@ -639,8 +599,24 @@ class CanvasExtraState {
...
+      }
+      // Stroked paths can be outside of the path bounding box by 1/2 the line
+      // width.
+      const scale = Util.singularValueDecompose2dScale(transform);
+      const xStrokePad = (scale[0] * this.lineWidth) / 2;
+      const yStrokePad = (scale[1] * this.lineWidth) / 2;
+      // console.log(`increasing bounding box ${xStrokePad} ${yStrokePad}`);
```
Snuffleupagus, line 613: Please remove this left-over debugging statement.

src/display/pattern_helper.js
```diff
@@ -85,42 +90,59 @@ class RadialAxialShadingPattern extends BaseShadingPattern {
...
-        tmpCtx.fill();
+      tmpCtx.fillStyle = this._createGradient(tmpCtx);
+      tmpCtx.fill();
 
-        pattern = ctx.createPattern(tmpCanvas.canvas, "repeat");
-        this.cachedCanvasPatterns.set(this, pattern);
-      }
+      pattern = ctx.createPattern(tmpCanvas.canvas, "no-repeat");
     } else {
       // Don't bother caching gradients, they are quick to rebuild.
```
Snuffleupagus, line 141: Please remove this comment, since I don't think it makes sense now that you've removed the caching above.

src/display/pattern_helper.js
```diff
@@ -85,42 +90,59 @@ class RadialAxialShadingPattern extends BaseShadingPattern {
...
       applyBoundingBox(ctx, this._bbox);
       pattern = this._createGradient(ctx);
     }
-    if (!shadingFill) {
+    if (pathType !== PathType.SHADING) {
```
Snuffleupagus, line 145: Is there any reason to not just place the *contents* of this `if`-block after line https://github.com/mozilla/pdf.js/pull/14230/files#diff-8e7b9866f3c2ff10ae9f003b1ce1c941b0113ae43167ece974c5c391862b8cf6R139 above instead?

brendandahl:
Are these chrome annotation and form failures new? They seem to work fine locally in tests and in the viewer....

Snuffleupagus:
> Are these chrome annotation and form failures new?

I cannot recall having seen them before, unfortunately.

Snuffleupagus:
> Are these chrome annotation and form failures new?

I cannot recall having seen them before, unfortunately.

> They seem to work fine locally in tests and in the viewer....

When testing this patch in Google Chrome locally, I can reproduce the test-failures. However, there's no errors in the console that'd explain the rendering failures.

src/display/pattern_helper.js
```diff
@@ -85,42 +90,59 @@ class RadialAxialShadingPattern extends BaseShadingPattern {
...
 
-        tmpCtx.fillStyle = this._createGradient(tmpCtx);
-        tmpCtx.fill();
+      tmpCtx.fillStyle = this._createGradient(tmpCtx);
+      tmpCtx.fill();
 
-        pattern = ctx.createPattern(tmpCanvas.canvas, "repeat");
-        this.cachedCanvasPatterns.set(this, pattern);
-      }
+      pattern = ctx.createPattern(tmpCanvas.canvas, "no-repeat");
```
Snuffleupagus, line 139: It appears, when testing `ShowText-ShadingPattern` locally in Google Chrome, that reverting this change fixes things!?

brendandahl: I see the issue now for this PDF. Let's fix that first in #14232.

I still have no idea why the annotation/form ones are failing.

brendandahl:
I haven't fixed the annotation/form bug yet. Just pushing so I can try this out on linux/windows...

src/display/pattern_helper.js
```diff
@@ -85,42 +90,53 @@ class RadialAxialShadingPattern extends BaseShadingPattern {
...
-
-        const tmpCtx = tmpCanvas.context;
-        tmpCtx.clearRect(0, 0, tmpCtx.canvas.width, tmpCtx.canvas.height);
-        tmpCtx.beginPath();
-        tmpCtx.rect(0, 0, tmpCtx.canvas.width, tmpCtx.canvas.height);
+    if (pathType === PathType.STROKE || pathType === PathType.FILL) {
+      const ownerBBox = owner.current.getClippedPathBoundingBox(
+        pathType,
+        ctx.mozCurrentTransform
+      );
```
Snuffleupagus, line 99: Looking at the previous test-results, there's also a breaking error which I assume is caused by this line (since this method uses `Util.intersect` which can return `null`).
```
TEST-UNEXPECTED-FAIL | test failed issue6961 | in firefox | page1 round 1 | render : TypeError: can't access property 2, ownerBBox is null
```

brendandahl: Nice catch, I totally missed that! It seems to fix the other issues. However, that PDF is now displaying wrong in a new way (it's wrong as is too). The computed bounding box for the shading is slightly too small. I'm not sure why yet. It's not going to be fun to debug, since that PDF has thousands of paths. :( 

I still don't quite understand why breaking that PDF, breaks other PDFs. We shouldn't be reusing any canvases, but we must not be cleaning up something correctly when an error happens.

Snuffleupagus: Could the problem be that we're re-using the *same* `<canvas>` for every page/document during tests, thus possibly keeping some bad graphics-state alive?

A while back I wanted to see if we could get rid of some intermittent failures by always creating a new `<canvas>` during tests, please see PR #12960; maybe we should see if that'd fixed these problems and potentially land that?

brendandahl: That brings up a good point! The canvas state should be getting reset when we assign to the width/height in driver.js, but in chrome we add our polyfill for mozCurrentTransform and that does NOT get reset. That would explain why this was only an issue in chrome. I'm thinking we should either override the width/height attributes of the canvas to reset the transform stuff or manually wipe out the transform stack and reset the current transform. The former is probably less error prone since we wouldn't have to remember to do it.

brendandahl: As for the bounding box being wrong in issue6961, that is because of #14241.

Snuffleupagus (approved):
r=me, thank you!

timvandermeij:
Awesome work on this and the other PRs; thank you!

PR #14238, opened 2021-11-05
Add a couple of basic unit-tests for `PDFPageViewBuffer`

Snuffleupagus:
The `PDFPageViewBuffer`-code is very important for the correct function of the viewer, but it's currently not tested at all.
While the `PDFPageViewBuffer` is obviously intended to be used with `PDFPageView`-instances, it only accesses a couple of `PDFPageView` properties/methods and consequently it's fairly easy to unit-test this code with dummy-data.

These unit-tests should help improve our confidence in this code, and will also come in handy with other changes that I'm working on (regarding modernizing and re-factoring the `PDFPageViewBuffer`-code).

### DIFF
--- a/test/unit/base_viewer_spec.js
+++ b/test/unit/base_viewer_spec.js
@@ -1,0 +1,162 @@
+/* Copyright 2021 Mozilla Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
It's great to have more unit test for the viewer logic; thank you for doing this!

PR #14240, opened 2021-11-05
XFA - Get each page asynchronously in order to avoid blocking the event loop (#14014)

### LINKED ISSUE #14014: Intermittent bug with XFA test on the bots
I ran `makeref` on the linux bot using X11 forwarding and I managed to get a [profile](https://share.firefox.dev/3hnn0aM) when Firefox hung. For information, I found nothing neither in the console nor in the terminal nor in network panel which helped me to guess what was wrong.
I  added few `console.log` to try to figure out where the hang happened and it was during `toHTML`:
https://github.com/mozilla/pdf.js/blob/064c21d360dd5a8bd0c9cfc6f16d4c135e7d106c/src/core/xfa/factory.js#L43
and so I finally added a `setTimeout(() => console.log(node.constructor.name), 10000)` + a `clearTimeout` around the call to `toHTML`: https://github.com/mozilla/pdf.js/blob/master/src/core/xfa/xfa_object.js#L370 (and the same around the `toHTML` in template.js file).
And I ran `makeref` in a `while true` during 8 hours and no hang... :(

About the profile, from @gabrielesvelto: "Firefox isn't doing nothing but it's waiting for an event in main event loop".

calixteman:


src/core/document.js
```diff
@@ -787,11 +787,11 @@ class PDFDocument {
...
     let num = 0;
     if (this.xfaFactory) {
-      num = this.xfaFactory.numPages;
+      num = this.xfaFactory.getNumPages();
     } else if (this.linearization) {
-      num = this.linearization.numPages;
+      num = Promise.resolve(this.linearization.numPages);
     } else {
-      num = this.catalog.numPages;
+      num = Promise.resolve(this.catalog.numPages);
```
Snuffleupagus, line 794: I don't think that the way you're using Promises here (and elsewhere) in combination with the `PDFManager.ensure...`-method usage, is correct and/or safe in general.
For one thing, the `this.catalog.numPages` getter can throw and I don't think that `Promise.resolve(/* previously thrown error */)` makes a lot of sense.
Furthermore, since the `PDFManager.ensure...`-methods only goes one level "deep" so to speak, I'm not sure that the associated `ensure`-call in `src/core/worker.js` does what it *should* now.

This should be fixable with a bunch more re-factoring, and I can look into that and try to provide more actionable review comments during the weekend; however for now I figured that just pointing these things out makes sense (such that this won't land as-is).

Snuffleupagus (changes requested):

src/core/xfa/factory.js
```diff
@@ -45,9 +46,29 @@ class XFAFactory {
...
+        }
+      };
+      setTimeout(lambda, 0);
+    });
+  }
+
+  async _createPages() {
     try {
-      this.pages = this.form[$toHTML]();
+      this.pages = await this._getPages();
```
Snuffleupagus, line 70: *Given that testing so far seem positive, I'm adding another small comment here.*

We this new method, we now have `getPages`, `_createPages`, and `_getPages` which feels a bit confusing. How about either:
 - Given the small size of both of these methods, how about just inlining the new code directly here instead?
 - Or at least calling the new method e.g. `_createPagesHelper` instead, to reduce some naming-confusion here.

calixteman:
No more test failures for the moment.
I don't know if it'll fix the test issues but it likely doesn't hurt to have this patch.
Once everything is ok, we can merge it and see for its positive effects on CI in the next days.
/botio test

Snuffleupagus (approved):
It certainly seems like this have helped, since there's not been any intermittent failures so far.
This only leaves one thing, trying to explain *why exactly* this actually worked :-)

r=me, since this doesn't really effect anything except the XFA-code now; thank you!

calixteman:
I really hope those intermittents are fixed: crossed fingers...

marco-c:
\o/

PR #14241, opened 2021-11-05
Don't double apply a group xobject's bbox.

brendandahl:
In `beginGroup` we create a new canvas that is the size of the
bounding box and we translate it to the offset. This means we don't need to
also apply the bounding box during `paintFormXObjectBegin`.

This improves mozilla#6961 quite a bit, but it still is missing the indention
in the ruler.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -527,7 +527,11 @@ class PartialEvaluator {
       operatorList.addOp(OPS.beginGroup, [groupOptions]);
     }
 
-    operatorList.addOp(OPS.paintFormXObjectBegin, [matrix, bbox]);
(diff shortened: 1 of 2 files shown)


Snuffleupagus (approved):
Should we perhaps change this to an "eq"-test instead?
https://github.com/mozilla/pdf.js/blob/30bd5f0a397a1235aaaefaada8f145bc35009382/test/test_manifest.json#L5925

r=me, with passing (or improving) tests; thank you!

brendandahl:
> Should we perhaps change this to an "eq"-test instead?

Yes!

brendandahl:
This should fix `transparency_group.pdf`, we weren't actually double applying the matrix, only the bbox.

/botio test

PR #14244, opened 2021-11-06
Prevent mobile devices from interfering with the textLayer-elements (issue 14243)

### LINKED ISSUE #14243: Random textLayer inflation in iOS Safari
*Any PDF with text (used in this demo: [sicp-pocket.pdf](https://github.com/sarabander/sicp-pocket/raw/master/sicp-pocket.pdf))*

Configuration:
- Web browser and its version: *Safari on iOS 14.8*
- Operating system and its version: *iOS 14.8*
- PDF.js version: *2.10.377 (prior versions also)*
- Is a browser extension: *No*

1. *Open PDF in viewer and **put phone in landscape mode***
2. *Tap anywhere in a page with some text in it*

*textLayer should not change. Before tapping:*
IMG_4862

*Some text elements beef up. After tapping:*
IMG_4861

*Affects text selection.*
IMG_4860

*Hosted on localhost.*

---

I don't know if this is a bug with Safari. It happened in previous iOS versions as well.\
Suggestion: this solves the issue:
```css
.textLayer span {
    -webkit-text-size-adjust: none;
}
```

Snuffleupagus:
*This is a tentative patch, since I don't have the necessary hardware to test it.*

See https://developer.mozilla.org/en-US/docs/Web/CSS/text-size-adjust, which is currently ignored in Firefox.
It seems overall safer, and more future-proof, to simply add this to the *entire* `textLayer` rather than its individual elements.

Fixes #14243

### DIFF
--- a/web/text_layer_builder.css
+++ b/web/text_layer_builder.css
@@ -23,6 +23,7 @@
   overflow: hidden;
   opacity: 0.2;
   line-height: 1;
+  text-size-adjust: none;
 }
 
 .textLayer span,


shivaprsd:
@Snuffleupagus Does this automatically add the prefixed version? The unprefixed version is not recognised by iOS Safari.

Snuffleupagus:
> Does this automatically add the prefixed version?

Of course, which is clear if you look at the generated file: http://54.241.84.105:8877/6ca1a9dbdc089cd/web/viewer.css

timvandermeij:
I have tested this with Firefox Nightly on Android and found no difference before/after, which is good because as the bug report mentioned this seems to only affect iOS devices. I don't have such a device to confirm this works. @shivaprsd Could you test the preview link above and see if that fixes the issue?

shivaprsd:
@timvandermeij I can confirm that it fixes the issue in Viewer (legacy). 👍🏼
Should be GTG in the modern viewer also.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #14247, opened 2021-11-06
[api-minor] Render pushbuttons on their own canvas (bug 1737260)

calixteman:
- First step to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1737260;
 - several interactive pdfs use the possibility to hide/show buttons to show different icons;
 - render pushbuttons on their own canvas and then insert it the annotation_layer;
 - update test/driver.js in order to convert canvases for pushbuttons into images.

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -198,7 +198,25 @@ class AnnotationElement {
       page.view[3] - data.rect[3] + page.view[1],
     ]);
 
-    container.style.transform = `matrix(${viewport.transform.join(",")})`;
+    if (data.hasOwnCanvas) {
+      const transform = viewport.transform.slice();
+      const [scaleX, scaleY] = Util.singularValueDecompose2dScale(transform);
+      width = Math.ceil(width * scaleX);
+      height = Math.ceil(height * scaleY);
+      rect[0] *= scaleX;
+      rect[1] *= scaleY;
+      // Reset the scale part of the transform matrix (which must be diagonal
+      // or anti-diagonal) in order to avoid to rescale the canvas.
+      // The canvas for the annotation is correctly scaled when it is drawn
+      // (see `beginAnnotation` in canvas.js).
+      for (let i = 0; i < 4; i++) {
+        transform[i] = Math.sign(transform[i]);
+      }
+      container.style.transform = `matrix(${transform.join(",")})`;
+    } else {
+      container.style.transform = `matrix(${viewport.transform.join(",")})`;
+    }
+
     container.style.transformOrigin = `${-rect[0]}px ${-rect[1]}px`;
 
     if (!ignoreBorder && data.borderStyle.width > 0) {
@@ -258,8 +276,13 @@ class AnnotationElement {
 
     container.style.left = `${rect[0]}px`;
     container.style.top = `${rect[1]}px`;
-    container.style.width = `${width}px`;
-    container.style.height = `${height}px`;
+
+    if (data.hasOwnCanvas) {
+      container.style.width = container.style.height = "auto";
+    } else {
+      container.style.width = `${width}px`;
+      container.style.height = `${height}px`;
+    }
     return container;
   }
 
@@ -2318,10 +2341,12 @@ class AnnotationLayer {
       sortedAnnotations.push(...popupAnnotations);
     }
 
+    const div = parameters.div;
+
     for (const data of sortedAnnotations) {
       const element = AnnotationElementFactory.create({
         data,
-        layer: parameters.div,
+        layer: div,
         page: parameters.page,
         viewport: parameters.viewport,
         linkService: parameters.linkService,
@@ -2343,19 +2368,21 @@ class AnnotationLayer {
         }
         if (Array.isArray(rendered)) {
           for (const renderedElement of rendered) {
-            parameters.div.appendChild(renderedElement);
+            div.appendChild(renderedElement);
           }
         } else {
           if (element instanceof PopupAnnotationElement) {
             // Popup annotation elements should not be on top of other
             // annotation elements to prevent interfering with mouse events.
-            parameters.div.prepend(rendered);
+            div.prepend(rendered);
           } else {
-            parameters.div.appendChild(rendered);
+            div.appendChild(rendered);
           }
         }
       }
     }
+
+    this.#setAnnotationCanvasMap(div, parameters.annotationCanvasMap);
   }
 
   /**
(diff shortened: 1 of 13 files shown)


Snuffleupagus (commented):
*A couple of initial comments, based on a quick look at this patch.*

web/pdf_page_view.js
```diff
@@ -653,35 +664,33 @@ class PDFPageView {
...
-            this.l10n,
-            /* enableScripting = */ null,
-            /* hasJSActionsPromise = */ null,
-            /* mouseState = */ null,
-            /* fieldObjectsPromise = */ null
-          );
-      }
-      this._renderAnnotationLayer();
-    }
-
```
Snuffleupagus, line 684: You shouldn't need to move all of this code, *only* the `this._renderAnnotationLayer();` line as far as I can tell!?

web/pdf_page_view.js
```diff
@@ -653,35 +664,33 @@ class PDFPageView {
...
+                  this.l10n,
+                  /* enableScripting = */ null,
+                  /* hasJSActionsPromise = */ null,
+                  /* mouseState = */ null,
+                  /* fieldObjectsPromise = */ null,
+                  /* annotationCanvasMap */ this._annotationCanvasMap
+                );
+            }
+            this._renderAnnotationLayer();
+          }
```
Snuffleupagus, line 686: Why not simply the following, with the other comment addressed?
```javascript

          if (this.annotationLayer) {
            this._renderAnnotationLayer();
          }
```

web/annotation_layer_builder.js
```diff
@@ -165,7 +170,8 @@ class DefaultAnnotationLayerFactory {
     enableScripting = false,
     hasJSActionsPromise = null,
     mouseState = null,
-    fieldObjectsPromise = null
+    fieldObjectsPromise = null,
+    annotationCanvasMap = null
```
Snuffleupagus, line 174: Please remember to update https://github.com/mozilla/pdf.js/blob/master/web/interfaces.js as well.

src/display/annotation_layer.js
```diff
@@ -2366,18 +2393,70 @@ class AnnotationLayer {
...
+      return;
+    }
+    for (const [id, canvas] of annotationCanvasMap) {
+      const element = div.querySelector(`[data-annotation-id="${id}"]`);
+      if (!element) {
+        continue;
+      }
+
+      const { firstChild } = element;
+      if (firstChild.nodeName === "CANVAS") {
```
Snuffleupagus, line 2453: Does this need to use `toUpperCase()` to avoid (any possible) problems?

brendandahl:
The button in issue12716.pdf is no longer scaled correctly with this patch:

image

src/display/annotation_layer.js
```diff
@@ -2366,18 +2393,70 @@ class AnnotationLayer {
...
         }
       }
     }
-    parameters.div.hidden = false;
+
+    this.setAnnotationCanvasMap(div, annotationCanvasMap);
+    div.hidden = false;
+  }
+
+  static setAnnotationCanvasMap(div, annotationCanvasMap) {
```
Snuffleupagus, line 2442: Let's make this properly private, since is shouldn't be called directly; i.e. `static #setAnnotationCanvasMap` instead.

src/display/annotation_layer.js
```diff
@@ -2366,18 +2393,70 @@ class AnnotationLayer {
...
    */
   static update(parameters) {
-    const transform = `matrix(${parameters.viewport.transform.join(",")})`;
+    const { page, viewport, annotations, annotationCanvasMap, div } =
+      parameters;
+    const transform = viewport.transform;
+    const matrix = `matrix(${transform.join(",")})`;
+
+    let scale, ownMatrix;
+    if (annotations.some(annotation => annotation.hasOwnCanvas)) {
```
Snuffleupagus, line 2402: Rather than having to iterate through the annotations *twice*, given that most pages won't have any canvases, can't you simply calculate this "special" matrix lazily the first time it's need in the loop below instead?

src/display/annotation_layer.js
```diff
@@ -198,7 +198,25 @@ class AnnotationElement {
       page.view[3] - data.rect[3] + page.view[1],
     ]);
 
-    container.style.transform = `matrix(${viewport.transform.join(",")})`;
+    if (data.hasOwnCanvas) {
+      const transform = viewport.transform.slice();
+      const scale = Math.abs(transform[0] || transform[1]);
+      width = Math.ceil(width * scale);
```
brendandahl, line 204: Does it make sense here to use `Util.singularValueDecompose2dScale` so we can handle different scaling in x/y?

calixteman: Is it possible to have a different scaling for x/y ?

src/display/annotation_layer.js
calixteman: Normally there are no problems:
https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName#value

Snuffleupagus (commented):
Another thing that I'm *guessing* doesn't currently work is the `useOnlyCssZoom` mode, when set either explicitly in the appOptions or implicitly by increasing the zoom level (a lot) in the viewer.

Most likely you'll need to modify the width/height of the `annotationCanvas`es in that case, perhaps similar to how the main page-canvas is handled when CSS-zooming is being used.

web/pdf_page_view.js
```diff
@@ -580,6 +583,14 @@ class PDFPageView {
     }
     this.textLayer = textLayer;
 
+    if (
+      this._annotationMode !== AnnotationMode.DISABLE &&
+      this.annotationLayerFactory &&
+      !this._annotationCanvasMap
+    ) {
+      this._annotationCanvasMap = new Map();
+    }
```
Snuffleupagus, line 592: Really sorry about the back-and-forth here, but seeing the latest round of changes it's probably overall simpler and more readable if we just do:
```suggestion
    if (
      this._annotationMode !== AnnotationMode.DISABLE &&
      this.annotationLayerFactory
    ) {
      this._annotationCanvasMap ||= new Map();
      
      this.annotationLayer ||=
        this.annotationLayerFactory.createAnnotationLayerBuilder(
          div,
          pdfPage,
          /* annotationStorage = */ null,
          this.imageResourcesPath,
          this._annotationMode === AnnotationMode.ENABLE_FORMS,
          this.l10n,
          /* enableScripting = */ null,
          /* hasJSActionsPromise = */ null,
          /* mouseState = */ null,
          /* fieldObjectsPromise = */ null,
          /* annotationCanvasMap */ this._annotationCanvasMap
        );
    }
```

web/pdf_page_view.js
Snuffleupagus, line 686: Really sorry about the back-and-forth here, but seeing the latest round of changes it's probably overall simpler and more readable if we just do:
```suggestion
          if (this.annotationLayer) {
            this._renderAnnotationLayer();
          }
```

web/pdf_page_view.js
```diff
@@ -800,10 +809,21 @@ class PDFPageView {
 
     const sfx = approximateFraction(outputScale.sx);
     const sfy = approximateFraction(outputScale.sy);
-    canvas.width = roundToDivide(viewport.width * outputScale.sx, sfx[0]);
-    canvas.height = roundToDivide(viewport.height * outputScale.sy, sfy[0]);
-    canvas.style.width = roundToDivide(viewport.width, sfx[1]) + "px";
-    canvas.style.height = roundToDivide(viewport.height, sfy[1]) + "px";
+
+    const hdpiScales = (w, h) => ({
```
Snuffleupagus, line 813: Following the previous suggestion, perhaps `const canvasDimensions = ...` here and then do `annotationCanvasDimensions: canvasDimensions,` below at line https://github.com/mozilla/pdf.js/pull/14247/files#diff-c48e3561004f5db8f11d5ebab2fd661591222ba911cb4173fbced15f026bac6bR841

quaoaris:
After this merge:
commit 3209c013c411ace4414246a153eac26eea640aba
Merge: e4f97a2a9 33ea817b2
Author: Brendan Dahl <brendan.dahl@gmail.com>
Date:   Tue Nov 16 08:10:40 2021 -0800

The buttons on testcase in https://bugzilla.mozilla.org/show_bug.cgi?id=1737260 are not visible anymore. Tested on Win10 with FF.

PR #14250, opened 2021-11-07
XFA - Encode tag names in UTF-8 when saving (fix #14249)

### LINKED ISSUE #14249: XFA form saved using pdf.js cannot be re-opened with pdf.js - "Warning: XFA Foreground documents are not supported"
[CIE-XFA-work.pdf](https://github.com/mozilla/pdf.js/files/7492888/CIE-XFA-work.pdf)

Configuration:
- Web browser and its version: chromium 90
- Operating system and its version: Debian 11
- PDF.js version: master@891f21fba6db64cd602c1a9a51826d7b9cd06af0
- Is a browser extension: no

1. Open the attached form in pdf.js
2. Fill out a field
3. Save/Download the form
4. Re-open the form

The form should render again, displaying the field with filled in value.

The form only renders the first time. Saving it with at least one field filled and re-opening the saved PDF fails with
`Warning: XFA Foreground documents are not supported`

## Additional info

Saving/Downloading the form without filling a field does not produce the error.
I have also tested other PDFs (e.g. [canadian-xfa-example.pdf](https://github.com/mozilla/pdf.js/files/7492899/canadian-xfa-example.pdf)) that do not exhibit this problem.
From what I can see in the code, it would appear that saving the form somehow turns it from being `pureXfa` to losing that characteristic.
https://github.com/mozilla/pdf.js/blob/c68dc03be685a5f2de5c2e99595f9bc747ffaa34/web/app.js#L1572

calixteman:


Snuffleupagus (approved):
r=me, thank you!

PR #14252, opened 2021-11-08
Correct annotation borders width

### LINKED ISSUE #14203: Annotation borders don't have the correct width

cathydev:
Incrementing value "data.borderStyle.width" in stroke-width property to achieve a wider border and modifying the stroke value to make sure it use data.color as a value instead of "transparent" so the changes to stroke-width are visible.

closes #14203

### DIFF
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -1804,8 +1804,8 @@ class LineAnnotationElement extends AnnotationElement {
     line.setAttribute("y2", data.rect[3] - data.lineCoordinates[3]);
     // Ensure that the 'stroke-width' is always non-zero, since otherwise it
     // won't be possible to open/close the popup (note e.g. issue 11122).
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
Sorry, but this is unfortunately not correct; you cannot just change the borderWidth of every single LineAnnotation.

First of all, I cannot imagine that this change passes all tests locally (there ought to be *loads* of failures); did you perhaps forget to run `gulp test` before submitting the PR?
Secondly, it appears that you've not really debugged/understood the actual underlying error (and what's special about that particular PDF document); being able to explain what's wrong is often as important as the patch itself.

Closing the PR for now, please re-open it once you've come up with a working patch!

PR #14255, opened 2021-11-09
Convert `GrabToPan` to a standard `class`

Snuffleupagus:
- **[web/grab_to_pan.js] Inline the `isLeftMouseReleased` helper function**

   Given the support information listed in the function itself, the [MDN compatibility data](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons#browser_compatibility), and the [currently supported browsers](https://github.com/mozilla/pdf.js/blob/4bb9de4b008dca2cdccf770bc359ae21971debb0/gulpfile.js#L79-L87) in the PDF.js project we should be able to simplify the code by inlining the function instead.

 - **Convert `GrabToPan` to a standard `class`**

   This code is the last piece[1] of the viewer that's not using standard `class`es, and by converting this code we get rid of some now unneeded boilerplate code (slightly reducing the size of the *built* `web/viewer.js` file).
   Note that while this code was originally imported from a separate repository, it was last sync-ed with upstream *five years* ago which is why this re-factoring should be OK as far as I'm concerned (and we've done some other clean-up since then as well).

   ---
   [1] Technically the `web/debugger.js` file is left as well, however that code is first of all not bundled in the *built* `web/viewer.js` file and secondly it's not even loaded by default either.

### DIFF
--- a/web/grab_to_pan.js
+++ b/web/grab_to_pan.js
@@ -14,97 +14,91 @@
  * limitations under the License.
  */
 
-/**
- * Construct a GrabToPan instance for a given HTML element.
- * @param options.element {Element}
- * @param options.ignoreTarget {function} optional. See `ignoreTarget(node)`
- * @param options.onActiveChanged {function(boolean)} optional. Called
- *  when grab-to-pan is (de)activated. The first argument is a boolean that
- *  shows whether grab-to-pan is activated.
- */
-function GrabToPan(options) {
-  this.element = options.element;
-  this.document = options.element.ownerDocument;
-  if (typeof options.ignoreTarget === "function") {
-    this.ignoreTarget = options.ignoreTarget;
-  }
-  this.onActiveChanged = options.onActiveChanged;
+// Class name of element which can be grabbed.
+const CSS_CLASS_GRAB = "grab-to-pan-grab";
 
-  // Bind the contexts to ensure that `this` always points to
-  // the GrabToPan instance.
-  this.activate = this.activate.bind(this);
-  this.deactivate = this.deactivate.bind(this);
-  this.toggle = this.toggle.bind(this);
-  this._onmousedown = this._onmousedown.bind(this);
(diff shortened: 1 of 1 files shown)


web/grab_to_pan.js
```diff
@@ -14,97 +14,91 @@
...
+// Class name of element which can be grabbed.
+const CSS_CLASS_GRAB = "grab-to-pan-grab";
+
+class GrabToPan {
   /**
-   * Class name of element which can be grabbed
+   * Construct a GrabToPan instance for a given HTML element.
+   * @param {Element} options.element
+   * @param {function} [options.ignoreTarget] - See `ignoreTarget(node)`.
+   * @param {function(boolean)} [options.onActiveChanged} - Called when
```
timvandermeij, line 25: `}` should be `]`

web/grab_to_pan.js
```diff
@@ -14,97 +14,91 @@
...
+      this.ignoreTarget = options.ignoreTarget;
+    }
+    this.onActiveChanged = options.onActiveChanged;
+
+    // Bind the contexts to ensure that `this` always points to
+    // the GrabToPan instance.
+    this.activate = this.activate.bind(this);
+    this.deactivate = this.deactivate.bind(this);
+    this.toggle = this.toggle.bind(this);
+    this._onmousedown = this.#onmousedown.bind(this);
```
timvandermeij, line 42: Wouldn't it be better to call this member variable and the associated method to `onMouseDown` instead (same for `onMouseMove` below) to match the casing in the rest of the code?

timvandermeij (approved):
r=me with these two comments addressed; thanks!

web/grab_to_pan.js
Snuffleupagus: Yes, that obviously makes more sense. (To be honest, I forgot to look closely at the names and mostly focused on the class-conversion itself.)

PR #14257, opened 2021-11-09
Ensure that abbreviated dictionary keys always take precedence in inline images (issue 14256)

### LINKED ISSUE #14256: Inline images with both abbreviated and full keys
For reference:
https://www.pdfa.org/safedocs-unearths-pdf-inline-image-issue/

In Firefox nightly:
image
with pdfium (edge):
image
and in Acrobat:
image

Snuffleupagus:
Please refer to:
 - https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#G7.1852045
 - https://www.pdfa.org/safedocs-unearths-pdf-inline-image-issue/
 - https://pdf-issues.pdfa.org/32000-2-2020/clause08.html#H8.9.7

Fixes #14256

### DIFF
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -52,6 +52,18 @@ import { RunLengthStream } from "./run_length_stream.js";
 const MAX_LENGTH_TO_CACHE = 1000;
 const MAX_ADLER32_LENGTH = 5552;
 
+const INLINE_IMAGE_KEYS = {
+  BitsPerComponent: "BPC",
+  ColorSpace: "CS",
+  Decode: "D",
+  DecodeParms: "DP",
+  Filter: "F",
+  Height: "H",
+  ImageMask: "IM",
+  Interpolate: "I",
+  Width: "W",
+};
+
 function computeAdler32(bytes) {
   const bytesLength = bytes.length;
   if (
@@ -495,6 +507,7 @@ class Parser {
     const stream = lexer.stream;
 
     // Parse dictionary.
+    const map = new Map();
     const dict = new Dict(this.xref);
     let dictLength;
     while (!isCmd(this.buf1, "ID") && this.buf1 !== EOF) {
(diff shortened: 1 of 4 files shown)


brendandahl:
Alternatively, should we reverse the get's? e.g. `dict.get("BitsPerComponent", "BPC")` to `dict.get("BPC", "BitsPerComponent")`?

Snuffleupagus:
> Alternatively, should we reverse the get's? e.g. `dict.get("BitsPerComponent", "BPC")` to `dict.get("BPC", "BitsPerComponent")`?

Initially I thought about doing that instead, but it'd have required edits all over the code-base and seems much more error-prone with any future changes since you'd need to ensure that the correct ordering is maintained everywhere.
At least to me, it seems much simpler to just avoid adding the "bad" data to the `Dict` in the first place, especially since all of this is limited to just *inline* images.

src/core/parser.js
```diff
@@ -52,6 +52,19 @@ import { RunLengthStream } from "./run_length_stream.js";
...
 
+const INLINE_IMAGE_KEYS = {
+  BitsPerComponent: "BPC",
+  ColorSpace: "CS",
+  Decode: "D",
+  DecodeParms: "DP",
+  Filter: "F",
+  Height: "H",
+  ImageMask: "IM",
+  Intent: "",
```
calixteman, line 63: nit: it's useless so it can be removed.

brendandahl:
Based of [this comment](https://github.com/pdf-association/pdf-issues/issues/3#issuecomment-753667152) it makes me think we should always do short name first even for other dictionaries. We could enforce this when using .get() with multiple arguments.

brendandahl:
There's also `Table 94 – Additional Abbreviations in an Inline Image Object` in the 2008 spec.

Snuffleupagus:
> Based of [this comment](https://github.com/pdf-association/pdf-issues/issues/3#issuecomment-753667152) it makes me think we should always do short name first even for other dictionaries. We could enforce this when using .get() with multiple arguments.

@brendandahl So I'm guessing that you'd prefer something along these lines instead: https://github.com/mozilla/pdf.js/commit/bfae1a318793d0319261746da78c0ce20f018325?w=1

PR #14258, opened 2021-11-10
Always prefer abbreviated keys, over full ones, when doing any dictionary lookups (issue 14256)

### LINKED ISSUE #14256: Inline images with both abbreviated and full keys

Snuffleupagus:
Note that issue #14256 was specifically about *inline* images, please refer to:
 - https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#G7.1852045
 - https://www.pdfa.org/safedocs-unearths-pdf-inline-image-issue/
 - https://pdf-issues.pdfa.org/32000-2-2020/clause08.html#H8.9.7

However, during review of the initial PR in https://github.com/mozilla/pdf.js/pull/14257#issuecomment-964469710, it was suggested that we instead do this *unconditionally for all* dictionary lookups.
In addition to re-ordering the existing call-sites in the `src/core`-code, and adding non-PRODUCTION/TESTING asserts to catch future errors, for consistency a number of existing `if`/`switch`-blocks were re-factored to also check the abbreviated keys first.

### DIFF
--- a/src/core/colorspace.js
+++ b/src/core/colorspace.js
@@ -379,14 +379,14 @@ class ColorSpace {
     cs = xref.fetchIfRef(cs);
     if (isName(cs)) {
       switch (cs.name) {
-        case "DeviceGray":
         case "G":
+        case "DeviceGray":
           return this.singletons.gray;
-        case "DeviceRGB":
         case "RGB":
+        case "DeviceRGB":
           return this.singletons.rgb;
-        case "DeviceCMYK":
         case "CMYK":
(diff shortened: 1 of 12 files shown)


brendandahl:
/botio makeref

Looks good, thanks!

PR #14260, opened 2021-11-11
Report "pageInfo" telemetry once, rather than for each rendered page

Snuffleupagus:
Reporting telemetry, in Firefox, includes using `JSON.stringify` on the data and then sending an event to the `PdfStreamConverter.jsm`-code.
In that code the event is handled and `JSON.parse` is used to retrieve the data, and in the "pageInfo"-case we'll then proceed to ignore everything except *the first* such event; see https://searchfox.org/mozilla-central/rev/24fac1ad31fb9c6e9c4c767c6a7ff45d226078f3/toolkit/components/pdfjs/content/PdfStreamConverter.jsm#509-514

All-in-all, sending the "pageInfo" telemetry for each rendered page is thus unnecessary and this patch makes the viewer send it only *once* instead.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -554,7 +554,7 @@ class BaseViewer {
       if (evt.cssTransform || this._onePageRenderedCapability.settled) {
         return;
       }
-      this._onePageRenderedCapability.resolve();
+      this._onePageRenderedCapability.resolve({ timestamp: evt.timestamp });
 
       this.eventBus._off("pagerendered", this._onAfterDraw);
       this._onAfterDraw = null;
--- a/web/app.js
+++ b/web/app.js
@@ -1372,7 +1372,12 @@ const PDFViewerApplication = {
       this._initializeAutoPrint(pdfDocument, openActionPromise);
     });
 
-    onePageRendered.then(() => {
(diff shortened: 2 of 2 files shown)


web/base_viewer.js
```diff
@@ -554,7 +554,7 @@ class BaseViewer {
       if (evt.cssTransform || this._onePageRenderedCapability.settled) {
         return;
       }
-      this._onePageRenderedCapability.resolve();
+      this._onePageRenderedCapability.resolve({ timestamp: evt.timestamp });
```
Snuffleupagus, line 557: One thing that I'm wondering is if we could potentially just do the following instead:
```javascript
      this._onePageRenderedCapability.resolve({ timestamp: performance.now() });
```
and then remove the `timestamp: performance.now(),`-lines from the various "pagerendered"-events in the https://github.com/mozilla/pdf.js/blob/master/web/pdf_page_view.js file.

It seems totally unnecessary to do these calls *potentially twice* for each rendered page, when we're not using anything except the very first one anyway.

timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #14262, opened 2021-11-11
Include the /Lang-property, when it exists, in the StructTree-data (issue 14261)

### LINKED ISSUE #14261: Language changes not available to assistive technologies
[test-lang.pdf](https://github.com/mozilla/pdf.js/files/7520359/test-lang.pdf)
The default language in this document is english. But the document contains 2 language changes.

Configuration:
- Web browser and its version: Firefox 93.0
- Operating system and its version: Windows 10 Version 1909 Build 18363.1256
- PDF.js version: from the master branch
- Is a browser extension: no
- Screen reader: NVDA 2021.2
- OS language: english

1. Open the file in PDF.js
2. Read the file with NVDA
3. The text in French is read with an English voice (in Acrobat, the text in French is read with a French voice)

French text should be read with a French voice.

In this PDF document, the content in French is made of two tags, one H1 and one P. These two tags have the property Language set to "French". This info should be available as a `lang` attribute in the DOM set to "French", so that screen readers like NVDA can select the correct language to read this text.
screen shot of a language change in Acrobat Pro

Here the only lang attribute we have in the DOM is the one of the default language. We should find 2 others, one for the title in French and one for the paragraph in French.
Screenshot of the dev tools showing only one alt attribute

/

Snuffleupagus:
*Please note:* This is a tentative patch, since I don't have the necessary a11y-software to actually test it.

### DIFF
--- a/src/core/struct_tree.js
+++ b/src/core/struct_tree.js
@@ -295,6 +295,10 @@ class StructTreePage {
       if (isString(alt)) {
         obj.alt = stringToPDFString(alt);
       }
+      const lang = node.dict.get("Lang");
+      if (isString(lang)) {
+        obj.lang = stringToPDFString(lang);
+      }
 
       for (const kid of node.kids) {
         const kidElement =
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -1836,6 +1836,7 @@ sozialökonomische Gerechtigkeit.`)
         children: [
           {
             role: "Document",
+            lang: "en-US",
             children: [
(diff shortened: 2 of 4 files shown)


Snuffleupagus:
@brendandahl Should we land this as-is for now, since I'm guessing that unless [bug 1741122](https://bugzilla.mozilla.org/show_bug.cgi?id=1741122) is fixed directly in Firefox we'd probably need to make larger (and more invasive) changes to the entire `StructTree` code?

PR #14266, opened 2021-11-12
Don't consider space as real space when there is an extra spacing (bug 931481)

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=931481;
 - real space chars are pushed in the chunk but when there is an extra spacing, the next char position must be compared with the previous one;
 - for example, an extra spacing can cancel a space so visually there are no space.

### DIFF
--- a/test/unit/api_spec.js
+++ b/test/unit/api_spec.js
@@ -1760,6 +1760,34 @@ sources, for full support with Dvips.`)
       await loadingTask.destroy();
     });
 
+    it("gets text content, with negative spaces (bug 931481)", async function () {
+      if (isNodeJS) {
+        pending("Linked test-cases are not supported in Node.js.");
+      }
+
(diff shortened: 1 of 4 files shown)


test/unit/api_spec.js
```diff
@@ -1760,6 +1760,30 @@ sources, for full support with Dvips.`)
       await loadingTask.destroy();
     });
 
+    it("gets text content, with negative spaces (bug 931481)", async function () {
```
Snuffleupagus, line 1763: To fix the GitHub-actions failure, please add the following block at the top of this test:
```javascript
if (isNodeJS) {
  pending("Linked test-cases are not supported in Node.js.");
}
```

calixteman: Thank you, I was exactly looking for that.

Snuffleupagus (approved):
r=me, thank you!

PR #14268, opened 2021-11-12
Remove non-displayable chars from outline title (#14267)

### LINKED ISSUE #14267: Outline title can contain non-displayable chars
See pdf in https://github.com/mozilla/pdf.js/issues/3109

Configuration:
- PDF.js version: builtin viewer

1. Look at outlines

image

image

calixteman:
- it aims to fix #14267;
 - there is nothing about chars in range [0-1F] in the specs but acrobat doesn't display them in any way.


Snuffleupagus (changes requested):
Sorry, but I don't think that we should implement it this way!

First of all, I'm not sure why it's a problem for the API itself to return such strings as-is (and this is potentially a breaking api-change).
Secondly, we're already ignoring *some* characters in the *viewer* instead; see https://github.com/mozilla/pdf.js/blob/7d6d3fc124593d94230f6fbd2f7e16fcd13efb10/web/base_tree_viewer.js#L58-L63
which uses https://github.com/mozilla/pdf.js/blob/7d6d3fc124593d94230f6fbd2f7e16fcd13efb10/src/shared/util.js#L569-L580

Hence my question is if it'd be generally safe to simply update `removeNullCharacters` to skip the `[\x00-\x1F]` range instead, or if that could cause issues at one of its call-sites?
For example what about e.g. https://github.com/mozilla/pdf.js/blob/7d6d3fc124593d94230f6fbd2f7e16fcd13efb10/src/display/display_utils.js#L350, although I suppose that none of the `[\x00-\x1F]`-chars would make sense in a link?

calixteman:
> 
> 
> Sorry, but I don't think that we should implement it this way!
> 
> First of all, I'm not sure why it's a problem for the API itself to return such strings as-is (and this is potentially a breaking api-change).

Yep, you're right: I didn't think about the api.

> , although I suppose that none of the `[\x00-\x1F]`-chars would make sense in a link?

I think it should be possible to have those chars in using some entities.
So I think we can extend `removeNullChar` to range  `[\x00-\x1F]` but we shouldn't use it when creating the href but something based on `encodeURIComponent`. Wdyt ?

Snuffleupagus:
How about limiting the scope of these changes to just the functionality needed for the viewer, since that's basically the only thing we have to fix here as far as I can tell, with e.g. something along these lines (although a better parameter name might not hurt):
```diff
diff --git a/src/shared/util.js b/src/shared/util.js
index ba94fec8b..188e97737 100644
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -567,15 +567,20 @@ class AbortException extends BaseException {
 }
 
 const NullCharactersRegExp = /\x00/g;
+const InvisibleCharactersRegExp = /[\x00-\x1F]/g;
 
 /**
  * @param {string} str
+ * @param {boolean} [removeInvisible]
  */
-function removeNullCharacters(str) {
+function removeNullCharacters(str, removeInvisible = false) {
   if (typeof str !== "string") {
     warn("The argument for removeNullCharacters must be a string.");
     return str;
   }
+  if (removeInvisible) {
+    return str.replace(InvisibleCharactersRegExp, "");
+  }
   return str.replace(NullCharactersRegExp, "");
 }
 
diff --git a/web/base_tree_viewer.js b/web/base_tree_viewer.js
index 6e203aea8..2c1d7c92f 100644
--- a/web/base_tree_viewer.js
+++ b/web/base_tree_viewer.js
@@ -59,7 +59,10 @@ class BaseTreeViewer {
    * @private
    */
   _normalizeTextContent(str) {
-    return removeNullCharacters(str) || /* en dash = */ "\u2013";
+    return (
+      removeNullCharacters(str, /* removeInvisible = */ true) ||
+      /* en dash = */ "\u2013"
+    );
   }
 
   /**
```
That way we'd not change any existing behaviour, and these viewer changes would apply not just to the OutlineView but also the AttachmentsView/LayersView as well.
Furthermore, we could easily test it by adding one (or two) new tests to the pre-existing ones in `test/unit/util_spec.js`.

calixteman:
I had a look on what acrobat and pdfium are doing here.
In pdfium every chars in `[00-1F]` is replaced by a white space (0x20):
https://pdfium.googlesource.com/pdfium/+/refs/heads/main/core/fpdfdoc/cpdf_bookmark.cpp#40
In acrobat it's a bit different:
 - 00: text after 00 is removed
 - 01-08: the char is not replaced
 - 09-1F: the char is replaced by a space (0x20)

Snuffleupagus:
Based on https://github.com/mozilla/pdf.js/pull/14268#issuecomment-968054903, I wonder if we shouldn't perhaps use a slightly modified and "combined" approach here:
 - Leave the `\x00` case alone, in order to avoid introducing any regressions (e.g. with hyper-links).
 - For the `\x01-\x1F` range replace all of them with regular spaces, since I cannot imagine that rendering some of them (as Adobe Reader does) is helpful. This new replacement could then be limited to a new `replaceInvisible`-mode in the existing `removeNullCharacters` function (such that we don't have to make a lot of code-changes).

src/shared/util.js
```diff
@@ -567,15 +567,19 @@ class AbortException extends BaseException {
...
  * @param {string} str
  */
-function removeNullCharacters(str) {
+function removeNullCharacters(str, replaceInvisible = false) {
   if (typeof str !== "string") {
     warn("The argument for removeNullCharacters must be a string.");
     return str;
   }
+  if (replaceInvisible) {
+    return str.replace(InvisibleCharactersRegExp, " ");
```
Snuffleupagus, line 581: My suggestion was to *purposely* leave the `\x00`-case intact, even in the new mode, since there could very well be regressions by changing that.
Not to mentioned that having the `\x00`-char treated differently depending on the value of the new parameter feels surprising.

In the PDF document that originally prompted us to add the `removeNullCharacters`-function, I believe that this will lead to a bunch of trailing spaces, which could also potentially affect formatting of the outline-titles.

test/unit/api_spec.js
```diff
@@ -1102,6 +1102,19 @@ describe("api", function () {
       await loadingTask.destroy();
     });
 
+    it("gets outline with non-displayable chars", async function () {
```
Snuffleupagus, line 1105: Do we need this specific unit-test, given that a more general one is added below?

test/unit/util_spec.js
```diff
@@ -185,6 +185,16 @@ describe("util", function () {
       const str = "string\x00With\x00Null\x00Chars";
       expect(removeNullCharacters(str)).toEqual("stringWithNullChars");
     });
+
+    it("should modify string with non-displayable characters", function () {
+      const str = Array.from(Array(32).keys())
+        .map(x => String.fromCharCode(x))
+        .join("a");
```
Snuffleupagus, line 192: Is this string perhaps missing the last "a" character, in order for it to be consistently formatted? Currently this code gives:
```
\u0000a\u0001a\u0002a\u0003a\u0004a\u0005a\u0006a\u0007a\u0008a	a
a\u000ba\u000ca
a\u000ea\u000fa\u0010a\u0011a\u0012a\u0013a\u0014a\u0015a\u0016a\u0017a\u0018a\u0019a\u001aa\u001ba\u001ca\u001da\u001ea\u001f
```
whereas I believe that you probably want to use e.g. 
```javascript
const str = [...Array(32).keys()]
  .map(x => String.fromCharCode(x) + "a")
  .join("");
```
since that'd give you:
```
\u0000a\u0001a\u0002a\u0003a\u0004a\u0005a\u0006a\u0007a\u0008a	a
a\u000ba\u000ca
a\u000ea\u000fa\u0010a\u0011a\u0012a\u0013a\u0014a\u0015a\u0016a\u0017a\u0018a\u0019a\u001aa\u001ba\u001ca\u001da\u001ea\u001fa
```

Snuffleupagus (changes requested):

test/unit/api_spec.js
calixteman: I let the test to avoid any future api breakages (if the test was here I wouldn't have written the first version of this patch).
But if you think it's useless I'm fine to remove it.

src/shared/util.js
calixteman: Don't you think we should trim the outline title when rendering it ?

test/unit/api_spec.js
Snuffleupagus: Let's just keep it then :-)

test/unit/util_spec.js
```diff
@@ -185,6 +185,23 @@ describe("util", function () {
...
+        .map(x => String.fromCharCode(x) + "a")
+        .join("");
+      // \x00 is replaced by an empty string.
+      const expected =
+        "a" +
+        " "
+          .repeat(31)
+          .split("")
+          .map(x => x + "a")
+          .join("");
```
Snuffleupagus, line 200: Given that it's the expected value, I cannot help wondering if simply spelling it out explicitly wouldn't actually be clearer and easier to read?
```javascript
const expected =
  "a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a";

File: web/base_tree_viewer.js [hunk 5]
```diff
@@ -59,7 +59,11 @@ class BaseTreeViewer {
    * @private
    */
   _normalizeTextContent(str) {
-    return removeNullCharacters(str) || /* en dash = */ "\u2013";
+    // Chars in range [0x00-0x1F] will be replaced with a white space.
```
Snuffleupagus, line 62: The char-range probably needs to be updated here.

Snuffleupagus (approved):
r=me, with the last comments addressed; thank you!

PR #14270, opened 2021-11-13
When parsing corrupt documents without any trailer-dictionary, fallback to the "top"-dictionary (issue 14269)

### LINKED ISSUE #14269: Invalid pdf structure for a signed file
Configuration:
- Chrome: 95.0.4638.69 - 64 bits
- PDF.js version: 2.10.377-dist

[test.pdf](https://github.com/mozilla/pdf.js/files/7529789/test.pdf)

- Is a browser extension: No

1. run the HELLO WORLD github example passing the attached file: https://github.com/mozilla/pdf.js/blob/master/examples/learning/helloworld.html

2. The PDF does NOT load & in Dev tools console we have:
{
    "message": "Invalid PDF structure.",
    "name": "InvalidPDFException"
} 

Show the PDF in the browser like this:
Expected
[teste_assinado.pdf](https://github.com/mozilla/pdf.js/files/7529848/teste_assinado.pdf)

Error Screen 1
Error Screen 2

Snuffleupagus:
There's obviously no guarantee that this will work in general, if the document is sufficiently corrupt, but it should hopefully be better than just throwing `InvalidPDFException` as currently happens.

Please note that, as is often the case with corrupt documents, it's somewhat difficult to know if we're rendering the document "correctly" with this patch[1]. In this case even Adobe Reader cannot open the document, which is always a good sign that it's *really* corrupt, however we're at least able to render *something* with this patch.

---
[1] Whatever "correct" even means when dealing with corrupt PDF documents, where often times different PDF viewers won't agree completely.

### DIFF
--- a/src/core/xref.js
+++ b/src/core/xref.js
@@ -590,6 +590,10 @@ class XRef {
     if (trailerDict) {
       return trailerDict;
     }
+    // No trailer dictionary found, taking the "top"-dictionary (if exists).
+    if (this.topDict) {
+      return this.topDict;
+    }
     // nothing helps
     throw new InvalidPDFException("Invalid PDF structure.");
   }
@@ -680,6 +684,8 @@ class XRef {
         throw e;
       }
       info("(while reading XRef): " + e);
+
+      this.startXRefQueue.shift();
     }
 
     if (recoveryMode) {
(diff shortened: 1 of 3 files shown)


calixteman (approved):
It could avoid an exception so likely it doesn't hurt to have this patch.
Thank you for doing this.

Snuffleupagus:
> It could avoid an exception so likely it doesn't hurt to have this patch.

The worst thing that can happen as a result of the patch is that an Error is thrown later, e.g. during parsing/rendering, but we'll at least have *a chance* of opening a few more corrupt PDF documents this way; thanks for the review!

/makeref

PR #14271, opened 2021-11-13
Parse query string in using URLSearchParams

calixteman:
- I just noticed in reading the code that we parse that stuff when something exists in the web api;
 - see https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -405,14 +405,8 @@ var Driver = (function DriverClosure() {
 
   Driver.prototype = {
     _getQueryStringParameters: function Driver_getQueryStringParameters() {
-      var queryString = window.location.search.substring(1);
-      var values = queryString.split("&");
-      var parameters = {};
-      for (var i = 0, ii = values.length; i < ii; i++) {
-        var value = values[i].split("=");
-        parameters[unescape(value[0])] = unescape(value[1]);
-      }
-      return parameters;
+      const queryString = window.location.search.substring(1);
+      return Object.fromEntries(new URLSearchParams(queryString).entries());
     },
 
     run: function Driver_run() {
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
> I just noticed in reading the code that we parse that stuff when something exists in the web api;

*For reference:* Note that this PDF.js code actually *pre-dates* the existence of the `URLSearchParams` functionality, which is why this code looks the way it does. In particular Safari were quite slow, comparatively, to implement `URLSearchParams` support.

If we're making these changes, shouldn't the following code be updated similar to `web/ui_utils.js` code?
https://github.com/mozilla/pdf.js/blob/85c6dd59cede2ffcfb77a574ddf43c7bf70fa438/test/resources/reftest-analyzer.js#L53-L63

calixteman:
> 
> If we're making these changes, shouldn't the following code be updated similar to the `web/ui_utils.js` code?
> 
> https://github.com/mozilla/pdf.js/blob/85c6dd59cede2ffcfb77a574ddf43c7bf70fa438/test/resources/reftest-analyzer.js#L53-L63

In this code we split on `&` and `;` when `URLSearchParams` handles only `&` as separator, it's why I didn't change it.
Is this `;` here because it has been forgotten or do we really have some urls containing it ?

Snuffleupagus:
> Is this `;` here because it has been forgotten or do we really have some urls containing it ?

As mentioned at the top of that file, that code was originally imported from Firefox and the "special" separator pattern thus came from the original file.
As far as I can find, we only have *a single* call-site that invoke the `reftest-analyzer` code and there we're only passing one hash-parameter; see https://github.com/mozilla/pdf.js/blob/de7cfed9e392eb3bfcb761ccea2c1f09cdf869aa/test/test.js#L231

Hence, as long as all tests and the `reftest-analyzer` itself works, I believe that it's *should* be safe to change that one as well.

Snuffleupagus (approved):

PR #14272, opened 2021-11-14
Use a faster method to determine whether the number is even or odd

kumartul:
Bit masking was slower than directly checking whether the number is odd or even. You can check the speed/time complexity of the algorithm by profiling and timestamping both chunks of code.

### DIFF
--- a/src/core/bidi.js
+++ b/src/core/bidi.js
@@ -74,11 +74,11 @@ const arabicTypes = [
 ];
 
 function isOdd(i) {
-  return (i & 1) !== 0;
+  return i % 2 !== 0;
 }
 
 function isEven(i) {
-  return (i & 1) === 0;
+  return i % 2 === 0;
 }
 
 function findUnequal(arr, start, value) {


Snuffleupagus (changes requested):
> Bit masking was slower than directly checking whether the number is odd or even.

The "directly checking"-part reads a little strangely, since your patch is actually using the [remainder operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder) which is (more or less) essentially the modulo operator.

> You can check the speed/time complexity of the algorithm by profiling and timestamping both chunks of code.

If you're claiming that changes you're making result in faster code, please actually include said measurements in the commit message (and PR description). Also, please keep in mind that micro-benchmarks are usually of quite limited usefulness.

Hence, in order for us to consider this patch, please find a suitable PDF document and then follow https://github.com/mozilla/pdf.js/wiki/Benchmarking-your-changes#using-the-test-runner to provide performance results in the context of an entire PDF document (in this case, you want to use a `text`-test when benchmarking).

---

Finally, please make sure that the commit message actually contains information about what code the patch is modifying (e.g. adding " in `src/core/bidi.js`" to the end of first line would've helped tremendously), since you otherwise have no idea what the patch does without actually reading the code.

Atul-Kumar-Official:
I measured the time taken by both the chunks of code and here is the result:
Bit Masking: 0.0346ms
Modulus Operator: 0.0184ms

I measured the time taken by both the chunks of code 100 times and the above mentioned time is the average time taken

Snuffleupagus (changes requested):
> I measured the time taken by both the chunks of code and here is the result: Bit Masking: 0.0346ms Modulus Operator: 0.0184ms
> 
> I measured the time taken by both the chunks of code 100 times and the above mentioned time is the average time taken

Please note the as mentioned above, see https://github.com/mozilla/pdf.js/pull/14272#pullrequestreview-805447201, micro-benchmarks are usually not all that helpful (and you've not said anything about the *environment*/*browser* used for your test; here we're mostly concerned with performance in Firefox).
Keep in mind that micro-benchmarks can sometimes even be completely misleading, since they may trigger JavaScript engines to optimize away code in a manner that's not representative when the code is run as part of a larger program.

Hence, please follow the benchmarking instructions referenced in https://github.com/mozilla/pdf.js/pull/14272#pullrequestreview-805447201.

MuhammadAmin86:
function user(){
const number = 0;
number = prompt("Enter a number: ");


const result = (number % 2  == 0);
 }

 console.log(`The number is ${result}.`);
[please kindly identify mistake and create a program i am a new learner please help me this code is in javascript]

Atul-Kumar-Official:
This is not a big mistake. You just need to put the console.log statement
in the function's scope so that whenever you call the function and a user
types the number in the prompt box, it gets printed in the console

Snuffleupagus:
Closing as WONTFIX, since without proper benchmarking results we unfortunately can't really evaluate the patch.

PR #14273, opened 2021-11-14

Snuffleupagus:


Snuffleupagus:
Initially I also tried including a Puppeteer update as well, but ran into issues locally with e.g. the unit-tests never closing the browser at the end and thus hanging; looking at https://github.com/puppeteer/puppeteer/releases/tag/v11.0.0 I didn't spot anything immediately obvious and decided to simply leave that part out of the PR.

timvandermeij (approved):

timvandermeij:
Thanks for the update!

PR #14280, opened 2021-11-16
Slightly optimize `spreadMode` toggling with `ScrollMode.PAGE` set (PR 14112 follow-up)

Snuffleupagus:
It shouldn't be necessary to iterate through *all* pages when using a non-default `spreadMode`, since we already know which page(s) should become visible.
This code is a left-over from the initial (local) implementation that resulted in PR #14112, however I forgot to clean-up some things such as e.g. this loop.

Also fixes an outdated comment, see PR #14204 which removed the mentioned data-structure.

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -616,9 +616,8 @@ class BaseViewer {
         }
 
         if (this._scrollMode === ScrollMode.PAGE) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #14287, opened 2021-11-18
Convert the test-suite to use the Fetch API rather than XMLHttpRequest 

Snuffleupagus:
- Convert the test-suite to use the Fetch API rather than XMLHttpRequest 

   The Fetch API should be well supported across modern browsers, see https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility, and allows simplifying the code thanks to all of its method returning Promises.

   This patch also introduces template strings, rather than using manual string concatenation, in a few spots in `test/driver.js`.
   ~~Finally, in `test/driver.js`, we'll now use the PDF.js utility function `createObjectURL` to get rid of the `FileReader` usage.~~

 - Convert the `writeSVG` function, in `test/driver.js`, to return a Promise

   This allows (some) simplification of the call-sites, since we can now use `async`/`await` directly rather than having to manually wrap a lot of code in a Promise.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -56,20 +56,22 @@ function loadStyles(styles) {
   return Promise.all(styles.map(style => style.promise));
 }
 
-function writeSVG(svgElement, ctx, resolve, reject) {
+function writeSVG(svgElement, ctx) {
   // We need to have UTF-8 encoded XML.
   const svg_xml = unescape(
     encodeURIComponent(new XMLSerializer().serializeToString(svgElement))
   );
-  const img = new Image();
-  img.src = "data:image/svg+xml;base64," + btoa(svg_xml);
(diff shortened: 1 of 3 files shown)


brendandahl:
Why'd you close this?

Snuffleupagus:
> Why'd you close this?

Because I was getting additional intermittents locally with the latest version of the patch (e.g. in `textLayer` tests), that I didn't really understand and that didn't really make a lot of sense. Based on that I'm no longer sure if trying to convert everything at once is a good idea, since I really don't want to introduce more intermittent failures in the tests.
Time permitting I'll try to look into this later, but right now I don't think this will work as-is and it does feel somewhat risky (for a small benefit).

PR #14291, opened 2021-11-19
[api-minor] Only use Workers when `postMessage` transfers are supported (PR 11123 follow-up)

Snuffleupagus:
Given that all modern browsers now support `postMessage` transfers, and have for years, it no longer seems necessary for the PDF.js library to support using Workers unless the `postMessage` transfers functionality is available.
This patch is a follow-up to PR #11123, which made it impossible to *manually* disable `postMessage` transfers for performance reasons (since it increases memory usage), which hasn't caused any bug reports as far as I know.[1]

Hence we'll now only support *proper* Worker implementations, with fully working `postMessage` transfers, and fallback to using "fake" Workers otherwise.

---
[1] At the time of that PR we still "supported" IE, which is why this code was left intact.

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -75,16 +75,9 @@ class WorkerMessageHandler {
       }
       testMessageProcessed = true;
 
-      // check if Uint8Array can be sent to worker
-      if (!(data instanceof Uint8Array)) {
-        handler.send("test", null);
-        return;
-      }
-      // making sure postMessage transfers are working
-      const supportTransfers = data[0] === 255;
-      handler.postMessageTransfers = supportTransfers;
-
-      handler.send("test", { supportTransfers });
+      // Ensure that `TypedArray`s can be sent to the worker,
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
I'm pretty much certain that the integration-test failures aren't related to this patch, given that I'm not touching any test-related code this patch, that they *only* fail on Linux, and finally that other recent PRs have shown identical errors.

calixteman:
Oh... these errors look strange.
It seems that something is wrong with Firefox.
Anyway I agree: your patch is very likely not responsible of these issues.

timvandermeij:
Indeed, I have confirmed in https://github.com/mozilla/pdf.js/pull/14233#issuecomment-974324680 that it also happened here, so it's not specific to this patch. Fortunately this is already tracked in #14293.

timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #14292, opened 2021-11-19
Remove empty ruleset css

ArunBohra12:


Snuffleupagus (changes requested):
I'm guessing, based on the (very short) commit message, that you're referring to the `body` here?
Hence, please don't make a whole bunch of other *unrelated* formatting changes here. (If anything, we should definitely enable auto-formatting rather than making manual changes.)

PR #14298, opened 2021-11-23
Center pages vertically in PresentationMode (issue 10906)

### LINKED ISSUE #10906: Pages do not center vertically in the presentation mode
Configuration:
- Web browser and its version: Firefox 67, Chrome 75
- PDF.js version: 2.1.266
- Is a browser extension: no

1. Open a PDF that needs the vertical centering. For example A4 paper displayed on the portrait display, or wide enough PDF page on the landscape orientation display.
2. Enter the presentiation mode.
3. The page is located on the top of the workspace, not in the middle.

Expected vertical align in Adobe Reader XI:

pdfjs-align-expected

Current vertical align in PDF.js:

pdfjs-align-bug

Snuffleupagus:
*This patch can be tested e.g. with the `sizes.pdf` document in the test-suite.*

While this patch isn't necessarily the best solution, e.g. it might be possible to solve this with *only* CSS, it's what I was able to come up with to address an old issue.
The solution here re-uses the `spread`-class in PresentationMode, since that one already takes care of centering pages *vertically*, together with a dummy-page that takes up the entire height of the window.

Finally, some PresentationMode-related CSS-rules are also simplified slightly, since the changes in PR #14112 (using Page-scrolling) allows some clean-up here.

Fixes #10906

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -159,23 +159,13 @@ class PDFPageViewBuffer {
   }
 }
 
-function isSameScale(oldScale, newScale) {
-  if (newScale === oldScale) {
-    return true;
-  }
-  if (Math.abs(newScale - oldScale) < 1e-15) {
-    // Prevent unnecessary re-rendering of all pages when the scale
-    // changes only because of limited numerical precision.
-    return true;
-  }
-  return false;
-}
-
 /**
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Let's do this; thanks!

PR #14299, opened 2021-11-24
Convert examples/components/pageviewer.js to await/async (issue 14127)

### LINKED ISSUE #14127: Convert examples/components/pageviewer.js to await/async

The [pageviewer.js](https://github.com/mozilla/pdf.js/blob/master/examples/components/pageviewer.js) example currently uses the older style of `promise.then` and callbacks. Convert it to use an `async` function and `await` style.

SaiKiranMukka:
resolves #14127

### DIFF
--- a/examples/components/pageviewer.js
+++ b/examples/components/pageviewer.js
@@ -47,28 +47,28 @@ const loadingTask = pdfjsLib.getDocument({
   cMapPacked: CMAP_PACKED,
   enableXfa: ENABLE_XFA,
 });
-loadingTask.promise.then(function (pdfDocument) {
+(async function () {
+  const pdfDocument = await loadingTask.promise;
   // Document loaded, retrieving the page.
-  return pdfDocument.getPage(PAGE_TO_VIEW).then(function (pdfPage) {
-    // Creating the page view with default parameters.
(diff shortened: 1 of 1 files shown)


examples/components/pageviewer.js
```diff
@@ -47,28 +47,29 @@ const loadingTask = pdfjsLib.getDocument({
   cMapPacked: CMAP_PACKED,
   enableXfa: ENABLE_XFA,
 });
-loadingTask.promise.then(function (pdfDocument) {
+
```
Snuffleupagus, line 50: Given that the existing code doesn't have a blank-line here, let's not add one either.

Snuffleupagus (changes requested):
The beginning of the commit message feels a little redundant, with both of the words "refactor" and "convert" used, hence please change it to the following instead (with correct initial capitalization):
```
Convert examples/components/pageviewer.js to await/async (issue 14127)
```

When updating the patch, please keep https://github.com/mozilla/pdf.js/wiki/Squashing-Commits in mind.

examples/components/pageviewer.js
SaiKiranMukka: Updated

Snuffleupagus (approved):
r=me, thanks for the patch!

Issue #14303, opened 2021-11-24
Crashes and timeouts on bug tracker corpus files


tballison:
I recently ran pdf.js via node against our bug tracker corpus files described by Peter Wyatt [here](https://de7.pdfa.org/a-new-stressful-pdf-corpus/) and [here](https://www.pdfa.org/stressful-pdf-corpus-grows/).

The full corpus is available [here](https://corpora.tika.apache.org/base/docs/bug_trackers/), and a prepackaged subset of PDFs is available [here](https://corpora.tika.apache.org/base/packaged/pdfs/pdfs_202011/).

I'm running a slight modification of your `getinfo.js` example: [code](https://github.com/tballison/file-observatory/blob/pdfspelunker/tika-containers/tika-pdfjs/js/my-getinfo.js) with the latest release: pdf.js-2.11.338. The environment is specified via [Docker](https://github.com/tballison/file-observatory/blob/pdfspelunker/tika-containers/tika-pdfjs/Dockerfile).

I should open separate tickets, but I don't want to spam your issue tracker.  

Please let me know how I can help.  Many thanks for pdf.js!

Some of these files are HUGE and some are corrupt (fuzzed).

**ExitValue=1**
~~https://corpora.tika.apache.org/base/docs/bug_trackers/poppler/poppler-91414-0.zip-2.gz-53.pdf~~ ***Fixed by #14312*** 
~~https://corpora.tika.apache.org/base/docs/bug_trackers/poppler/poppler-91414-0.zip-2.gz-54.pdf~~ ***Fixed by #14312*** 
~~https://corpora.tika.apache.org/base/docs/bug_trackers/poppler/poppler-91414-1.zip-2.gz-53.pdf~~ ***Identical to `poppler-91414-0.zip-2.gz-53.pdf`***
~~https://corpora.tika.apache.org/base/docs/bug_trackers/poppler/poppler-91414-1.zip-2.gz-54.pdf~~ ***Identical to `poppler-91414-0.zip-2.gz-54.pdf`***

**ExitValue=134**
~~https://corpora.tika.apache.org/base/docs/bug_trackers/poppler/poppler-67295-0.pdf~~ ***Fixed by #14311***
~~https://corpora.tika.apache.org/base/docs/bug_trackers/poppler/poppler-85140-0.pdf~~ ***Fixed by #14311***
[...]

PR #14304, opened 2021-11-24
Ensure that `ChunkedStream` won't attempt to request data *beyond* the document size (issue 14303)

Snuffleupagus:
This bug was surprisingly difficult to track down, since it didn't just depend on range-requests being used but also on how quickly the document was loaded. To even be able to reproduce this locally, I had to use a very small `rangeChunkSize`-value (note the unit-test).

The cause of this bug is a bogus entry in the XRef-table, causing us to attempt to request data from *beyond* the actual document size and thus getting into an infinite loop.

Fixes *one* of the issues listed in issue #14303, namely the `PDFBOX-4352-0.pdf` document.


Snuffleupagus:
Based on https://github.com/mozilla/pdf.js/pull/14233#issuecomment-978149680 I believe that the Linux failures aren't caused by this patch.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

Issue #14306, opened 2021-11-24
AForm validation rejects valid fields


cincodenada:
## Attach (recommended) or Link to PDF file here:
The PDF itself is semi-sensitive. I'm happy to provide via unsecured channels, just can't quite post it publicly.

## Configuration:
- Web browser and its version: Firefox 94.0.1
- Operating system and its version: Mac OS 11.6
- PDF.js version: Tested with 2.11.298 (in above Firefox) and git master (5e2aec7dd)

## What is the expected behavior? (add screenshot)
I should be able to fill AForm-validated fields with valid values:
- I should be able to type "123-45-6789" into an SSN field
- I should be able to type "867-5309" into a phone field

## What went wrong? (add screenshot)
I got validation errors instead, when trying to insert the dashes. If I try to not insert the dashes, it lets me while editing, but throws an error when leaving the field.
<img width="426" alt="Screen Shot 2021-11-24 at 11 33 53 AM" src="https://user-images.githubusercontent.com/479715/143302917-ee2f6161-616f-4138-b0af-5048fbe735f6.png">
<img width="426" alt="Screen Shot 2021-11-24 at 11 37 13 AM" src="https://user-images.githubusercontent.com/479715/143302928-741ea372-c9a1-4107-9373-94000e0622ce.png">
<img width="622" alt="Screen Shot 2021-11-24 at 11 38 29 AM" src="https://user-images.githubusercontent.com/479715/143303092-5cc0c999-9c15-4a5d-82ea-be4316bfe755.png">
<img width="622" alt="Screen Shot 2021-11-24 at 11 38 44 AM" src="https://user-images.githubusercontent.com/479715/143303099-b5da332a-c30b-4c2f-bcf6-8635ad58a6b3.png">

## Additional info

For zip codes, the while-typing validation doesn't limit length - I can type as many numbers as I want, and am only complained at once I exit the field.

I've dug into this and will open a PR shortly with more details, but there are other even stranger quirks around internal deletions/insertions as well.

PR #14307, opened 2021-11-24
Fix AForm Validation for fields with non-digit characters

cincodenada:
This is a proposed fix for #14306 - I believe the core issue is that `mergeChanges()` tries to use the keyboard event to predict what the value of the field will be after the change is applied, and gets it wrong for changes that don't involve an explicit selection, at least in my environment.

The cause here is that `mergeChanges()` seems to assume that character insertions/appends will come with `selStart` and `selEnd` set, and thus it can just treat them like any other selection replacement. That isn't the case on in my environment, where appends (as well as single-character insertions and deletions) have `selStart` and `selEnd` set to `-1`. This results in the predicted value of the field being a single character: the character being appended.

This ends up not totally breaking things - for validations that will pass if each individual character typed is a valid first character in the field (since that's what it gets interpreted as), things are mostly fine - a ZIP code can be entered just fine, although length is not constrained until the field is exited.

But for fields with hyphens in them, for instance, pdf.js won't allow you to type the hyphen (because "-" by itself is not the beginning of a valid SSN), and then when the field is blurred, the final validation runs, and complains about the missing hyphens.

This PR updates `mergeChanges()` so that if `selStart` and `selEnd` are `-1` it assumes the event is either an append and predicts the result as value + change, or a delete and predicts the result as value.slice(0, -1). This fixes the most common cases, but still leaves room for some very strange behavior if the user inserts or deletes characters in the middle.

Because we don't have `selStart` and `selEnd`, I don't see any way of distinguishing between and append an an insertion, or a deletion at the end of the value vs an internal deletion. In this PR I just assume that changes are happening at the end of the string, but that means we'll be quite wrong about internal deletions/insertions, and if someone goes back and inserts a character we'll predict that it was added at the end of the string.

Fortunately that prediction isn't persisted, and internal insertions/deletions should be rare, so I think it's worth this most-of-the-way fix that should cover the majority of cases. If there is some way to determine _where_ single-character changes are being made, from the event or some other source, that would of course be optimal, but I didn't see one, and I don't know enough about how these events are generated to know if it's something we could add ourselves, or if it's something in some PDF standard somewhere.

### DIFF
--- a/src/scripting_api/aform.js
+++ b/src/scripting_api/aform.js
@@ -564,7 +564,7 @@ class AForm {
           event.change.length +
           event.selStart -
           event.selEnd;
-        if (finalLen >= 8) {
+        if (finalLen > 8 || event.value[0] === "(") {
           formatStr = "(999) 999-9999";
         } else {
           formatStr = "999-9999";
--- a/src/scripting_api/aform.js
+++ b/src/scripting_api/aform.js
@@ -486,15 +486,15 @@ class AForm {
     ]);
 
     function _checkValidity(_value, _cMask) {
-      for (let i = 0, ii = value.length; i < ii; i++) {
-        const mask = _cMask.charAt(i);
-        const char = _value.charAt(i);
-        const checker = checkers.get(mask);
+      for (let i = 0, ii = _value.length; i < ii; i++) {
+        const maskChar = _cMask.charAt(i);
+        const valueChar = _value.charAt(i);
+        const checker = checkers.get(maskChar);
         if (checker) {
-          if (!checker(char)) {
+          if (!checker(valueChar)) {
             return false;
           }
-        } else if (mask !== char) {
+        } else if (maskChar !== valueChar) {
           return false;
         }
       }
--- a/src/scripting_api/event.js
+++ b/src/scripting_api/event.js
@@ -52,19 +52,23 @@ class EventDispatcher {
    * the change is applied, so we can validate against it and cancel the event
    * if need be
    *
-   * TODO: Given the event info we currently have, single-character insertions
-   * and deletes are indistinguishable from appends. For now we just assume
-   * it's an append, the most common case, but this leaves some edge cases. See
-   * issue #14307 for details.
-   *
+   * TODO: Given the event info we currently have, we can't determine where in
+   * the value a single-character insertion/deletion happened.  For now we just
+   * assume they happen at the end of the string, which is by far the most
+   * common case, but this leaves some edge cases, see issue #14307.
    */
   mergeChange(event) {
     let value = event.value;
     if (typeof value !== "string") {
       value = value.toString();
     }
-    // If there was no selection, assume this is an append
+    // If there was no selection, it's a single-character change
     if (event.selStart === -1 && event.selEnd === -1) {
+      if (event.change === "") {
+        // Empty change indicates a deletion
+        return value.slice(0, -1);
+      }
+      // Otherwise, assume it's an append
       return value + event.change;
     }
 
--- a/src/scripting_api/event.js
+++ b/src/scripting_api/event.js
@@ -47,11 +47,28 @@ class EventDispatcher {
     this._document.obj._eventDispatcher = this;
   }
 
+  /*
+   * Take an event object and reconstruct what we think the result will be once
+   * the change is applied, so we can validate against it and cancel the event
+   * if need be
+   *
+   * TODO: Given the event info we currently have, single-character insertions
+   * and deletes are indistinguishable from appends. For now we just assume
+   * it's an append, the most common case, but this leaves some edge cases. See
+   * issue #14307 for details.
+   *
+   */
   mergeChange(event) {
     let value = event.value;
     if (typeof value !== "string") {
       value = value.toString();
     }
+    // If there was no selection, assume this is an append
+    if (event.selStart === -1 && event.selEnd === -1) {
+      return value + event.change;
+    }
+
+    // Otherwise, splice in the change to replace the selection
     const prefix =
       event.selStart >= 0 ? value.substring(0, event.selStart) : "";
     const postfix =
--- a/test/unit/scripting_spec.js
+++ b/test/unit/scripting_spec.js
@@ -361,8 +361,8 @@ describe("Scripting", function () {
         name: "Keystroke",
         willCommit: false,
         change: "o",
-        selStart: 4,
-        selEnd: 4,
+        selStart: -1,
+        selEnd: -1,
       });
 
       expect(send_queue.has(refId)).toEqual(true);
@@ -398,8 +398,8 @@ describe("Scripting", function () {
         name: "Keystroke",
         willCommit: false,
         change: "o",
-        selStart: 4,
-        selEnd: 4,
+        selStart: -1,
+        selEnd: -1,
       });
 
       expect(send_queue.has(refId)).toEqual(true);
@@ -1119,8 +1119,8 @@ describe("Scripting", function () {
           change: "3",
           name: "Keystroke",
           willCommit: false,
-          selStart: 0,
-          selEnd: 0,
+          selStart: -1,
+          selEnd: -1,
         });
         expect(send_queue.has(refId)).toEqual(false);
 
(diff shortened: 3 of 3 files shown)


cincodenada:
I found the contribution guide I had somehow glossed over the first time and ran unit tests, and fixed up some expectations - I plan to circle back and add some more unit tests to cover the additional cases I ran into here as well.

src/scripting_api/event.js
```diff
@@ -47,11 +47,32 @@ class EventDispatcher {
...
+   * the value a single-character insertion/deletion happened.  For now we just
+   * assume they happen at the end of the string, which is by far the most
+   * common case, but this leaves some edge cases, see issue #14307.
+   */
   mergeChange(event) {
     let value = event.value;
     if (typeof value !== "string") {
       value = value.toString();
     }
+    // If there was no selection, it's a single-character change
```
calixteman, line 65: Where did you find this information ?
In the specs (see http://dev.datalogics.com/cookbook/document/AcrobatDC_js_api_reference.pdf#page=381&zoom=auto,-214,751) about `selStart`:
`The starting position of the current text selection during a keystroke event.`
I just tried in Acrobat to add text in textfield and print the changes.
 - I entered: `abcdef`
 - I selected `bcd` and I pressed `g`
and I get the following output:
```
change=a, selStart=0, selEnd=0
change=b, selStart=1, selEnd=1
change=c, selStart=2, selEnd=2
change=d, selStart=3, selEnd=3
change=e, selStart=4, selEnd=4
change=f, selStart=5, selEnd=5
change=g, selStart=1, selEnd=4
```
My understanding is that the selStart is the position of the caret before the change.
When there is no selection, selEnd is equals to selStart else it's the position of the end of the selection.

test/unit/scripting_spec.js
```diff
@@ -361,15 +361,15 @@ describe("Scripting", function () {
         name: "Keystroke",
         willCommit: false,
         change: "o",
-        selStart: 4,
-        selEnd: 4,
+        selStart: -1,
```
calixteman, line 364: All the changes in this file are very likely wrong.

src/scripting_api/aform.js
```diff
@@ -486,15 +486,15 @@ class AForm {
     ]);
 
     function _checkValidity(_value, _cMask) {
-      for (let i = 0, ii = value.length; i < ii; i++) {
-        const mask = _cMask.charAt(i);
-        const char = _value.charAt(i);
-        const checker = checkers.get(mask);
+      for (let i = 0, ii = _value.length; i < ii; i++) {
```
calixteman, line 489: my bad... thank you to point that out.

src/scripting_api/event.js
```diff
@@ -47,11 +47,32 @@ class EventDispatcher {
     this._document.obj._eventDispatcher = this;
   }
 
+  /*
+   * Take an event object and reconstruct what we think the result will be once
+   * the change is applied, so we can validate against it and cancel the event
+   * if need be
```
Snuffleupagus, line 53: Nit: Here, and everywhere in the patch, please make sure to always end all sentences with a period.

Also, please keep https://github.com/mozilla/pdf.js/wiki/Squashing-Commits in mind when updating the patch.

calixteman:
The main issue is here:
https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js#L1072
`elementData.beforeInputSelectionRange` is `null` when it shouldn't.

There is another problem here:
https://github.com/mozilla/pdf.js/blob/master/src/scripting_api/event.js#L29-L30
where we should have `??` instead of `||`.

src/scripting_api/aform.js
```diff
@@ -564,7 +564,7 @@ class AForm {
           event.change.length +
           event.selStart -
           event.selEnd;
-        if (finalLen >= 8) {
+        if (finalLen > 8 || event.value[0] === "(") {
```
calixteman, line 567: I think the rhs of `||` is useless but definitely the lhs is a way better like this.

calixteman:
And I just checked MDN:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event
and `beforeinput` is now available, so we should likely use it now instead of this workaround.

src/scripting_api/event.js
cincodenada: I found it through debugging and seeing what the events actually looked like - like I said (maybe here or on the bug, I don't remember), I wasn't sure if how these events happen was specified in a doc somewhere or not - it looks like it is, which I guess means the bug is that pdf.js isn't populating selStart/selEnd for these events? That would also make sense, and would solve the edge cases as well.

test/unit/scripting_spec.js
cincodenada: These changes may be wrong in that they are against spec, but they are correct in that they are what pdf.js is emitting currently. It sounds like pdf.js's behavior in creating these events is the actual bug though.

src/scripting_api/aform.js
cincodenada: The RHS is necessary because we validate every time the field changes. For final validation you're right that it's unnecessary, but for the validation-during-typing, without the RHS if someone starts typing `(888)` the validation will stop them - for some reason the first character seems to be ignored for validation-during-typing, but the second parenthesis will cause an error:
Screen Shot 2021-11-24 at 3 34 02 PM

This is because `(888)` only has 5 characters, so this code selects the second pattern (`999-9999`), and `(888)` does not match the beginning of the second pattern. So, in that case we want to select the first pattern to match against, which is what the RHS is doing.

cincodenada:
> The main issue is here:
> https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js#L1072
> elementData.beforeInputSelectionRange is null when it shouldn't.

Okay, so yeah sounds like this code will work fine if `selStart`/`selEnd` is always present, which it sounds like is the behavior of Acrobat. So then, the root bug is further up the stack, where we're not passing down `selStart`/`selEnd` for edits without an explicit selection. Are you on the trail of a fix for that, or should I try digging in further?

> There is another problem here:
> https://github.com/mozilla/pdf.js/blob/master/src/scripting_api/event.js#L29-L30
> where we should have ?? instead of ||.

Oh, good catch - I looked at that code and didn't even notice but you're right, we're dropping zeroes there. I wonder if that's connected to the reason the first character isn't validating during typing...

test/unit/scripting_spec.js
cincodenada: So to be clear, I agree these will probably need to end up being reverted, hopefully replaced by some tests (or fixed test) of the generation of events for single-character edits, to validate that they are populating `selStart`/`selEnd` like this further up in the process.

cincodenada:
In summary: I'm happy to continue digging here and move this PR towards integrating beforeInput and getting a fix in higher up where it belongs, but I don't want to do duplicate work, so I can also just revert the workaround bits and this PR can just be the smaller fixes around phone numbers. Either way I'll squash things down once things are settled - do y'all prefer a small number of atomic commits, or one commit per PR?

src/scripting_api/aform.js
calixteman: It makes sense, +1.

calixteman:
I'm fine with a PR to only modify aform.js and it'd be nice if you could add a test to check that format is fine.
It could something similar to:
https://github.com/mozilla/pdf.js/blob/a2c380ccb3fb69ebe436606f3816f7d2851d94d9/test/unit/scripting_spec.js#L791-L828
where an alert is handled and to:
https://github.com/mozilla/pdf.js/blob/a2c380ccb3fb69ebe436606f3816f7d2851d94d9/test/unit/scripting_spec.js#L376-L403
where a text is entered (using selStart & selEnd).
Once that stuff is fixed then I think you could work on a second PR to add beforeinput stuff.

calixteman:
@cincodenada, any progress ?
If you don't have time to work on that stuff, please tell it and I'll fix the issues myself.

calixteman:
I push this PR: https://github.com/mozilla/pdf.js/pull/14429
in order to address issues found by @cincodenada.
@cincodenada I hope you're well and thank you for your help.

PR #14310, opened 2021-11-25
Abort parsing when the XRef /W-array contain bogus entries (issue 14303)

Snuffleupagus:
For this particular PDF document, we have `/W [1 2 166666666666666666666666666]` which obviously makes no sense.

While this patch makes no attempt at actually validating the entries in the /W-array, we'll now simply abort all processing when the end of the PDF document has been reached (thus preventing hanging the browser).
Please note that this patch doesn't enable the PDF document to be loaded/rendered, but at least it fails "correctly" now.

Fixes one of the issues listed in issue #14303, namely the `REDHAT-1531897-0.pdf`document.


calixteman (approved):
LGTM, thx for doing this.

PR #14313, opened 2021-11-26
Change the `_pagePromises` cache, in the worker, from an Array to a Map

Snuffleupagus:
- Change the `_pagePromises` cache, in the worker, from an Array to a Map

   Given that not all pages necessarily are being accessed, or that the pages may be accessed out of order, using a `Map` seems like a more appropriate data-structure here.
   Furthermore, this patch also adds (currently missing) caching for XFA-documents. Loading a couple of such documents in the viewer, with logging added, shows that we're currently re-creating `Page`-instances unnecessarily for XFA-documents.

 - Convert `PDFDocument._getLinearizationPage` to an async method

   This, ever so slightly, simplifies the code and reduces overall indentation.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -1261,7 +1261,7 @@ class PDFDocument {
     ]);
   }
 
-  _getLinearizationPage(pageIndex) {
+  async _getLinearizationPage(pageIndex) {
     const { catalog, linearization } = this;
     if (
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #14295, opened 2021-11-21
Remove the `{BaseViewer, PDFThumbnailViewer}._pagesRequests` caches

Snuffleupagus:
*Please refer to the individual commit messages.*

PR #14314, opened 2021-11-26
[Regression] Prevent errors, during loading, in the viewer for XFA-documents (PR 14295 follow-up)

Snuffleupagus:
In the second commit in PR #14295, I forgot that the pages in XFA-documents don't have references (like in regular PDF documents); sorry about that!


calixteman (approved):
With green CI, r=me.
Thanks.

PR #14320, opened 2021-11-28

Snuffleupagus:


Snuffleupagus:
Given that I don't really know TypeScript, and don't have time to try and figure out why updating it didn't just work, I've settled on simply pinning the version number for now. (We really need TypeScript users to help handle these things...)

/botio test

timvandermeij (approved):

timvandermeij:
Thank you for the updates!

Snuffleupagus:
@timvandermeij Do you perhaps have any idea, or maybe time to look into, why updating Puppeteer to the latest versions fail?

Awhile back I tried to update to version 11, and today I tried version 12 as well, but this causes testing in Firefox to break locally. More specifically, when running e.g. the unit-tests Firefox no longer shuts down at the end of the tests.

I've been looking at https://github.com/puppeteer/puppeteer/releases/, but I'm not able to tell what (if anything) we need to change in our code to be compatible with the new versions.

timvandermeij:
Hm, good question; I don't see anything particularly suspicious in the changelog. I'll try to test it out today.

timvandermeij:
I can now reproduce the issue with Puppeteer 11 on Linux too. This required much more time than I hoped since my system's Node.js version got updated to version 17 and all of sudden Firefox wouldn't start at all anymore. After a lot of debugging and searching I found that the Node.js developers made a breaking change in version 17 that breaks quite a few packages (and people don't seem too happy about it in the upstream issues I'm reading...). Fortunately there is a configuration option to default back to the "old" working behavior, so I'll prepare a PR for that once I also (hopefully) get Puppeteer 11+ working.

Snuffleupagus:
> This required much more effort than I hoped since my system's Node.js version got updated to version 17 and all of sudden Firefox wouldn't start at all anymore.

Oof, that sounds "fun".
Locally I'm still on Node 12, which seem to be *generally* be supported across all(?) `npm` packages, which is why I'm suspecting that *something* must have changed in Puppeteer as well.

PR #14321, opened 2021-11-28
Upgrade to Puppeteer 12

timvandermeij:

@Snuffleupagus Note that the only change that requires an action here is that contributors are now required to run Node.js 15 or higher. I think this is reasonable, but that would mean that you'd have to upgrade your Node.js 12 version (which is 2.5 years old as of now); unfortunately I don't see a way around that. This is because of the fourth commit, about `package-lock.json`, which I had to do since Node.js 15+ doesn't offer a way to produce the old lock file version 1 files anymore, and otherwise I couldn't upgrade Puppeteer. I hope this is OK.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,7 @@
         "postcss": "^8.4.4",
         "postcss-calc": "^8.0.0",
         "prettier": "^2.5.0",
-        "puppeteer": "^10.4.0",
+        "puppeteer": "^12.0.0",
         "rimraf": "^3.0.2",
         "streamqueue": "^1.1.2",
         "stylelint": "^14.1.0",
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
> I think this is reasonable, but that would mean that you'd have to upgrade your Node.js 12 version (which is 2.5 years old as of now); unfortunately I don't see a way around that.

Generally that'd be reasonable, but unfortunately I'm personally still stuck on Windows 7 for the time being (where Node 12 is the highest officially supported version).
Is there really no other way to update Puppeteer, since according to https://nodejs.org/en/about/releases/ Node 12 should still be "supported" for a little while longer?

Snuffleupagus:
> I think this is reasonable, but that would mean that you'd have to upgrade your Node.js 12 version (which is 2.5 years old as of now); unfortunately I don't see a way around that.

Generally that'd be reasonable, but unfortunately I'm personally still stuck on Windows 7 for the time being (where Node 12 is the highest officially supported version).
Is there really no other way to update Puppeteer, since according to https://nodejs.org/en/about/releases/ Node 12 should still be "supported" for a little while longer?

Snuffleupagus (approved):
It looks like some serious detective work and debugging went into this patch series!
Really nice work, and excellent comments and detailed/helpful commit messages all around; thanks a lot for fixing this so quickly!

PR #14322, opened 2021-11-29
Hide specific annotations during render

wilkosz:
### Background
There has been much deliberating over adding functionality to programmatically hide annotations during canvas render. 

### TODO 
- [ ] Hide annotations from render context based on ids

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1384,6 +1384,7 @@ class PDFPageProxy {
     background = null,
     optionalContentConfigPromise = null,
     annotationCanvasMap = null,
+    annotationIdsNotRendered = [],
   }) {
     if (typeof PDFJSDev !== "undefined" && PDFJSDev.test("GENERIC")) {
       if (arguments[0]?.renderInteractiveForms !== undefined) {


Snuffleupagus:
Please note that issue #7282 was closed (since we now support disable rendering of *all* annotations).

Closing as WONTFIX, since this kind of functionality seem like a special-case that has the risk of adding (from the general PDF.js library perspective) too much complexity to the code (which then has to be maintained, supported, and tested).

wilkosz:
@Snuffleupagus Can you please provide information on how to disable specific annotations i.e. for rotated pages we can render highlight and text annotation (dochub.com) but disallow annotation form fields like field types `Tx`, `Btn`, and`Ch`. Is there any way to accomplish this with annotation storage? or other pdfJsPage API methods?

calixteman:
You can use annotation storage:
https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L458-L476
which are used here:
https://github.com/mozilla/pdf.js/blob/c14552874b7464b95adc27a8060a5f5f0b60470b/src/core/document.js#L403

wilkosz:
@calixteman, thanks a million

PR #14324, opened 2021-11-29
Enforce PAGE-scrolling for *very* large/long documents (bug 1588435, PR 11263 follow-up)

Snuffleupagus:
This patch is essentially a continuation of PR #11263, which tried to improve loading/initialization performance of *very* large/long documents.

Note that browsers, in general, don't handle a huge amount of DOM-elements very well, with really poor (e.g. sluggish scrolling) performance once the number gets "large". Furthermore, at least in Firefox, it seems that DOM-elements towards the bottom of a HTML-page can effectively be ignored; for the PDF.js viewer that means that pages at the end of the document can become impossible to access.

Hence, in order to improve things for these *very* large/long documents, this patch will now enforce usage of the (recently added) PAGE-scrolling mode for these documents. As implemented, this will only happen once the number of pages *exceed* 15000 (which is hopefully rare in practice).
While this might feel a bit jarring to users being *forced* to use PAGE-scrolling, it seems all things considered like a better idea to ensure that the entire document actually remains accessible and with (hopefully) more acceptable performance.

Fixes [bug 1588435](https://bugzilla.mozilla.org/show_bug.cgi?id=1588435), to the extent that doing so is possible since the document contains 25560 pages (and is 197 MB large).

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -54,6 +54,11 @@ import { XfaLayerBuilder } from "./xfa_layer_builder.js";
 
 const DEFAULT_CACHE_SIZE = 10;
 
+const PagesCountLimit = {
+  FORCE_SCROLL_MODE_PAGE: 15000,
+  FORCE_LAZY_PAGE_INIT: 7500,
+};
+
 /**
  * @typedef {Object} PDFViewerOptions
  * @property {HTMLDivElement} container - The container for the viewer element.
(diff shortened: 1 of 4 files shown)


brendandahl:
Thanks for looking into this. Another option I was considering is using what many of the infinite scroll libraries do, where you have a fixed number of page divs and as they go out of view when you scroll you recycle them and add them after the visible elements. The scroll bar can become jumpy with this approach, but I think it's also possible to create really big container that is the estimated size of all the pages.

brendandahl:
Of course the above would be a much bigger undertaking.

Snuffleupagus:
> Of course the above would be a much bigger undertaking.

It would, and there'd likely be a bunch of edge-cases that would make all of the much more difficult than it seems at first.

Considering that these *very* large documents hopefully should not be all that common, I just figured that simply utilizing the now existing PAGE-scrolling mode seemed like a quick and easy win (especially compared to the current state of things).

timvandermeij (approved):

timvandermeij:
For such large documents, this seems like a very reasonable approach to me; in any case it's better than the document being inaccessible. If necessary there can always be a follow-up to implement a smarter method, but this should already improve the current situation. Thank you for doing this!

MoYuXianR:
I was thinking that you can achieve dynamic loading rendering, for large pdf documents, do not need to load a one-time loading, each time to load a little bit for the performance of a little bit of the configuration of the use of the user can be liberalized restrictions

PR #14325, opened 2021-11-29
Remove the unused `skipCount` parameter from `Catalog.getPageDict` (PR 14311 follow-up)

Snuffleupagus:
- Remove the unused `skipCount` parameter from `Catalog.getPageDict` (PR 14311 follow-up)

   This was added in PR #14311, but given that I completely missed to update the `PDFDocument.getPage` signature accordingly it's completely unused.
   Given that things work just as fine as-is, let's simply remove that optional parameter for now; sorry about the churn here!

 - Slightly extend the "creates pdf doc from PDF file with bad XRef table" unit-test (PR 14304 follow-up)

   Given that we're able to "render" this document, let's extend the unit-test to actually check that we're able to obtain the operatorList; although given the overall issues in the document it'll be empty.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -1085,7 +1085,7 @@ class Catalog {
     });
   }
 
-  getPageDict(pageIndex, skipCount = false) {
+  getPageDict(pageIndex) {
     const capability = createPromiseCapability();
     const nodesToVisit = [this._catDict.getRaw("Pages")];
     const visitedNodes = new RefSet();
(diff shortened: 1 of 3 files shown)


Snuffleupagus:
@calixteman If you've got time, mind reviewing (or basically rubber-stamping) these changes?
I've got other changes planned, in order to fix a regression I introduced recently, and I'd like to get this one landed first; thank you!

calixteman (approved):
LGTM.

PR #14328, opened 2021-11-30
Update (primarily) the Node.js examples to release page resources

Snuffleupagus:
Given that Node.js doesn't support Workers, general PDF.js performance will be worse when compared to browsers. In an attempt to improve at least memory usage a little bit, update the Node.js examples to release page resources once parsing is done for that page.


timvandermeij (approved):

timvandermeij:
Good idea!

PR #14333, opened 2021-12-01
Handle errors correctly when data lookup fails during /Pages-tree parsing (issue 14303)

Snuffleupagus:
This only applies to severely corrupt documents, where it's possible that the `Parser` throws when we try to access e.g. a /Kids-entry in the /Pages-tree.

Fixes two of the issues listed in issue #14303, namely the `poppler-742-0.pdf...` and `poppler-937-0.pdf...` documents.


Snuffleupagus:
*The commit message was updated slightly, however no actual code changes were made.*

src/core/catalog.js
```diff
@@ -1161,7 +1167,14 @@ class Catalog {
           }
         }
 
-        const kids = currentNode.get("Kids");
+        let kids;
+        try {
+          kids = currentNode.get("Kids");
```
calixteman, line 1172: Can't the other `currentNode.get("Type")` at line 1183 fail too ?

Snuffleupagus: In theory it obviously could, even if we've *thus far* not seen such a document; anyway I've updated the patch as suggested.

calixteman (approved):
LGTM

Issue #8019, opened 2017-02-02
PDF Viewer in Web not displaying content


jclgs:
Incorrect display on PDF file.

Configuration:
- Chrome/Firefox
- Windows/MAC
- PDF.js version: 1.4.20 - 1.6.210

**Bug #1**
The issue is on PDF using Symbol Layer in Illustrator. The issue is only in Chrome MAC/PC. It works well in Safari, Firefox, IE. You can see the good result in picture #1 in Adobe Reader and the bug in picture #2.
The first picture is from a PDF opened in AdobeReader. The second picture is with the PDF.js.

![image](https://cloud.githubusercontent.com/assets/24614035/22557398/6b9f4d52-e938-11e6-8bd4-3530dea02e0e.png)

As soon as we unlink Symbol Layer in Illustrator and generate a new PDF, the second picture with missing information is now printing the text in PDF.js.

Is it possible to get Symbol Layer working in Chrome. It works in IE, Fifefox, Safari but not Chrome, one of the most use browser.

**Bug #2** 
The second issue is related to MASK item in Illustrator. In Firefox, the MASK is not displayed correctly, but that time, in Chrome, it is working well. Look at the orange Circle in Firefox. In Firefox, it should be displayed like Chrome is doing.

![image](https://cloud.githubusercontent.com/assets/24614035/22557640/18c36a4a-e939-11e6-8094-cbfd61d1e87a.png)

If PDF is needed to debug, I can transfer them to you. Is this possible send those in private because they are business content.
Thanks
Jonathan

yurydelendik:
Can you try to disable hardware/GPU  acceleration 2d canvas in Chrome (via chrome://flags), see if it will fix the issue?

yurydelendik:
> If PDF is needed to debug, I can transfer them to you. Is this possible send those in private because they are business content.

Can you recreate them in smaller examples, e.g. by editing sensitive material in Illustrator?

jclgs:
Hello.

For Bug #1 in Chrome, if I disable Acceleration 2d canvas, it solves the issue. The only thing, I cannot ask all my customer to disable that feature. Is there a workaround or fix can be release.

For Bug #2 in Firefox, I will transfer you a example with sensitive data remove. I will ask one of our graphic artist to generate me a new PDF file.

Thanks

yurydelendik:
[...]

PR #14339, opened 2021-12-04
Add a (linked) test-case for issue 8019

Snuffleupagus:
Given that [bug 1336572](https://bugzilla.mozilla.org/show_bug.cgi?id=1336572) was just closed as fixed, thus fixing issue #8019 in Firefox[1], let's add a test-case to enable us to catch any future regressions either in PDF.js or in browsers themselves.

---
[1] It also seems to be working in Google Chrome, although I'm having a slightly difficult time deciphering *exactly* what configurations were affected when looking through issue #8019.

### DIFF
--- a/test/pdfs/issue8019.pdf.link
+++ b/test/pdfs/issue8019.pdf.link
@@ -1,0 +1,1 @@
+https://github.com/mozilla/pdf.js/files/748704/R9450_5_2_jersey.pdf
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

/botio makeref

Issue #14305, opened 2021-11-24
Crashes/timeouts on freshly fuzzed files


tballison:
On #14303, I shared unmodified files from our bug tracker corpus that appeared to cause some problems for pdf.js.  On this issue, I'll attach files that are freshly fuzzed using files from that corpus as seeds.  I avoided files used in #14303 as seeds for these runs  

I haven't done root cause analysis, and I apologize if these are all just triggering the same underlying bug (or that fixed in #14304).  The files are named for their seed file and then the sha-256 of the fuzzed version.

[new-nasties.zip](https://github.com/mozilla/pdf.js/files/7598083/new-nasties.zip)

tballison:
Unrelated sidenote...omg, github accepts or rejects files based solely on their file suffix, and it doesn't recognize tgz?!  Ommm...
[new-nasties.tgz.pdf](https://github.com/mozilla/pdf.js/files/7598095/new-nasties.tgz.pdf)
.

tballison:
[quine.gz](https://github.com/mozilla/pdf.js/files/7598148/quine.gz)

Snuffleupagus:
> new-nasties.zip
> new-nasties.tgz.pdf

All those documents seem to "work" more-or-less about as good as can be expected, when testing with an up-to-date PDF.js version, given that we're dealing with various kinds of corrupt PDF documents.

Let's close this as a duplicate of #14303 for now, since that one likely covers any remaining problems already.

tballison:
I've rebuilt from master, and I'm finding that fuzzing the bugtracker corpus is still yielding some files that cause crashes (exitValue=1) and timeouts (e.g. a 2kb file timing out at 3 minutes).  My goal is not to annoy you.  If you'd like to see the new files, please let me know.

Snuffleupagus:
> If you'd like to see the new files, please let me know where I should hang them (this ticket or #14303 or another).

Adding them to the open issue, i.e. 14303, seems fine assuming they're indeed "new" problems.

---

I've not had time to go all the remaining ones in 14303 in detail, but I believe that there's basically three categories: Bugs already tracked elsewhere, documents which work fine in the viewer (note that Node.js performs worse e.g. on account of no Worker-support), and documents which are so ridiculously huge that it doesn't make much sense to worry about them.

PR #14340, opened 2021-12-04
Handle errors when fetching the raw /Metadata (issue 14305)

Snuffleupagus:
Currently the `Catalog.metadata` getter only handles errors during parsing, however in a *corrupt* PDF document fetching of the raw /Metadata can obviously fail as well.
Without this patch the `PDFDocumentProxy.getMetadata` method, in the API, can thus fail which it *never* should and this will cause the viewer to not initialize all state as expected.

Fixes one of the documents in issue #14305.


timvandermeij (approved):

timvandermeij:
Nice find!

PR #14152, opened 2021-10-16
Fix a `xfaFaxtory` typo in the shadowing in the  `PDFDocument.xfaFactory` getter, and some other clean-up

Snuffleupagus:
*Please refer to the individual commit messages for additional details.*

PR #14341, opened 2021-12-04
Ensure that the `shadow` helper function is passed a valid property (PR 14152 follow-up)

Snuffleupagus:
Trying to shadow a non-existent property is always an implementation mistake, since it leads to the `shadow`-call not having any effect.

In PR #14152 I overlooked the fact that it's fairly easy to enforce this during development/testing, since that can help catch e.g. simple spelling bugs.


calixteman (approved):
Very good idea.

timvandermeij (approved):

timvandermeij:
Thanks!

PR #14344, opened 2021-12-04
Modernize the test driver

timvandermeij:
_The commit messages contain more information about the individual changes. The `?w=1` flag is most likely necessary for reviewing most commits._

Initially I only tried to enable the `no-var` linting rule since our testing code is one of the final places where we exclude that rule, but I quickly found that the driver code could use further modernization. Hence this commit series, where each commit contains a separate refactoring step. Note that there is most likely more opportunity for improvement here, for instance using template strings more often and getting rid of XHR, but I deliberately left that for a follow-up since this series is already quite a big.

This PR makes the driver code 103 lines shorter and, in my opinion at least, easier to understand and work with.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -393,17 +393,12 @@ var rasterizeXfaLayer = (function rasterizeXfaLayerClosure() {
  * @property {HTMLDivElement} end - Container for a completion message.
  */
 
-/**
- * @class
- */
 // eslint-disable-next-line no-unused-vars
-var Driver = (function DriverClosure() {
+class Driver {
   /**
-   * @constructs Driver
    * @param {DriverOptions} options
    */
-  // eslint-disable-next-line no-shadow
-  function Driver(options) {
+  constructor(options) {
     // Configure the global worker options.
     GlobalWorkerOptions.workerSrc = WORKER_SRC;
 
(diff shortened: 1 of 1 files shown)


test/driver.js
```diff
@@ -143,89 +144,79 @@ async function resolveImages(node, silentErrors = false) {
...
+  viewport,
+  textContent,
+  enhanceTextSelection
+) {
   const styles = {
     common: {
       file: "./text_layer_test.css",
       promise: null,
     },
   };
```
Snuffleupagus, line 158: Won't this re-factoring mean that you're now re-initializing the style `promise` for *every* single `rasterizeTextLayer` call, rather than caching it as intended here!?

This probably applies elsewhere below in the file as well.

test/driver.js
```diff
@@ -428,569 +376,550 @@ var Driver = (function DriverClosure() {
     this.canvas = document.createElement("canvas");
   }
 
-  Driver.prototype = {
-    _getQueryStringParameters: function Driver_getQueryStringParameters() {
-      const queryString = window.location.search.substring(1);
-      return Object.fromEntries(new URLSearchParams(queryString).entries());
-    },
+  _getQueryStringParameters() {
```
Snuffleupagus, line 379: Should we perhaps export https://github.com/mozilla/pdf.js/blob/dc455c836e80799fccd8f1741eb8f62e9ec01c37/web/ui_utils.js#L183-L194 in the viewer components and re-use that in this file rather than essentially re-implementing that here?

Snuffleupagus (changes requested):
It's really nice that you're doing this, but before I review it in detail there's one thing I fear is incorrect; please see the inline comment.

test/driver.js
Snuffleupagus: How about e.g. adding a class with static methods, e.g. something like this since that'd allow you to handle the style-caching internally instead:
```javascript
class Rasterize {
  static #textStyles = {...};

  static #annotationStyles = {...};

  static #xfaStyles = {...};

  static async textLayer(...) {}

  static async annotationLayer(...) {}

  static async xfaLayer(...) {}
}
```

Snuffleupagus:
> and getting rid of XHR,

*Just a word of warning:* I actually tried that recently, but abandoned it for the time being, since I ran into some trouble; please see https://github.com/mozilla/pdf.js/pull/14287#issuecomment-974200678

test/driver.js
timvandermeij: Ah, you're right. Somehow I managed to completely overlook this. I'll revisit the patch series today.

timvandermeij: Fixed in the new commit series.

test/driver.js
timvandermeij: For now I skipped that because the series is already quite big and I intended to only touch the driver code here. We can do that in a follow-up though because I don't see other limitations for us to do that.

test/driver.js
```diff
@@ -143,106 +145,75 @@ async function resolveImages(node, silentErrors = false) {
...
+      common: {
+        file: "../web/annotation_layer_builder.css",
+        promise: null,
+      },
+      overrides: {
+        file: "./annotation_layer_builder_overrides.css",
+        promise: null,
+      },
+    };
+    return shadow(Rasterize, "annotationStylePromise", loadStyles(styles));
```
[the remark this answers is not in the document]
Snuffleupagus: Given that you've added proper shadowing, do we still need to track the promises manually here?
Basically the pre-existing code feels overly complicated, and I wonder if we couldn't now do something like
```javascript
const styles = {
  common: "../web/annotation_layer_builder.css",
  overrides: "./annotation_layer_builder_overrides.css",
}
return shadow(this, "annotationStylePromise", loadStyles(styles));
```
with `loadStyles` modified along e.g. these lines:
```javascript
function loadStyles(styles) {
  const promises = [];

  for (const type of styles) {
    const file = styles[type];

    promises.push(new Promise(function (resolve, reject) {
      const xhr = new XMLHttpRequest();
      xhr.open("GET", file);
      xhr.onload = function () {
        resolve(xhr.responseText);
      };
      xhr.onerror = function (e) {
        reject(new Error(`Error fetching style (${file}): ${e}`));
      };
      xhr.send(null);
    });
  }

  return Promise.all(promises);
}
```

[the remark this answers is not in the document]
timvandermeij: Good idea; I have made the changes. The driver file is now exactly 100 lines shorter than before :-)

test/driver.js
```diff
@@ -260,13 +200,12 @@ const rasterizeAnnotationLayer = (function rasterizeAnnotationLayerClosure() {
...
       foreignObject.setAttribute("height", viewport.height + "px");
       const style = document.createElement("style");
-      const stylePromise = loadStyles(styles);
       foreignObject.appendChild(style);
       const div = document.createElement("div");
       div.className = "annotationLayer";
 
       // Rendering annotation layer as HTML.
-      const [common, overrides] = await stylePromise;
+      const [common, overrides] = await Rasterize.annotationStylePromise;
```
Snuffleupagus, line 212: Please note that `static` methods can *directly* call other `static` methods and access static properties.
Hence the more compact `this.annotationStylePromise` format could be used here, and similarly elsewhere in file.

Snuffleupagus (approved):
r=me, with one final comment addressed (everywhere that it applies); thanks a lot for doing this!

I'd suggest running the tests maybe a couple of times, to ensure that there's no new intermittent failures (since the usage of `async` perhaps affect timings ever so slightly).

PR #14345, opened 2021-12-05
Ensure that the viewer handles `BaseViewer` initialization failures

Snuffleupagus:
*This patch can be tested e.g. with the `poppler-85140-0.pdf` document from the test-suite.*

For some sufficiently corrupt documents the `getDocument` call will succeed, but fetching even the very first page fails. Currently we only print error messages (in the console) from the `{BaseViewer, PDFThumbnailViewer}.setDocument` methods, but don't actually provide these errors to allow the viewer to handle them properly.
In practice this means that the GENERIC viewer won't display the `errorWrapper`, and in the MOZCENTRAL viewer the *browser* loading indicator is never hidden (since we never unblock the "load" event).

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -949,22 +949,22 @@ const PDFViewerApplication = {
       pdfDocument => {
         this.load(pdfDocument);
       },
-      exception => {
+      reason => {
         if (loadingTask !== this.pdfLoadingTask) {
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #14347, opened 2021-12-06
Improve caching in `Catalog.getPageDict` (PR 8207 follow-up)

Snuffleupagus:
PR #8207 added caching to improve the performance of `Catalog.getPageDict`, by not having to repeatedly fetch the same data and also reducing the asynchronicity of that method.
However, because of annoying off-by-one errors[1] the caching became less efficient than it could/should be.[2] Note here that the /Pages-tree is zero-indexed, and that e.g. `pageIndex = 5` thus correspond to the *sixth* page of the document.

---
[1] In particular the `currentPageIndex + count < pageIndex` part.

[2] For example, even when loading a relatively small/simple document such as `tracemonkey.pdf` in the viewer, the number of `xref.fetchAsync(currentNode)` calls are reduced from `56` to `44` with this patch.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -1102,7 +1102,7 @@ class Catalog {
         if (isRef(currentNode)) {
           const count = pageKidsCountCache.get(currentNode);
           // Skip nodes where the page can't be.
-          if (count > 0 && currentPageIndex + count < pageIndex) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #14350, opened 2021-12-07
Prevent an infinite loop when parsing corrupt /CCITTFaxDecode data (issue 14305)

Snuffleupagus:
Fixes one of the documents in issue #14305.


timvandermeij (approved):

timvandermeij:
Looks good!

/botio makeref

PR #14354, opened 2021-12-09
Further improve caching in `Catalog.getPageDict`, for `disableAutoFetch` mode (PR 8207 follow-up)

Snuffleupagus:
PR #8207 added caching to improve the performance of `Catalog.getPageDict`, by not having to repeatedly fetch the same data and also reducing the asynchronicity of that method.
However, because of *another* oversight on my part, we're only caching /Page references once we've found the correct page. As long as all pages are loaded *in order* this doesn't really matter (happens by default in the viewer), but when `disableAutoFetch` is used the pages may be fetched in a more random order (this patch reduces the asynchronicity of `Catalog.getPageDict` slightly in that case).

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -1117,13 +1117,14 @@ class Catalog {
 
           xref.fetchAsync(currentNode).then(function (obj) {
             if (isDict(obj, "Page") || (isDict(obj) && !obj.has("Kids"))) {
+              // Cache the Page reference, since it can *greatly* improve
+              // performance by reducing redundant lookups in long documents
+              // where all nodes are found at *one* level of the tree.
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
> Do you think it's possible to have a pdf to add in:
> https://hg.mozilla.org/mozilla-central/file/tip/testing/talos/talos/tests/pdfpaint/pdfpaint.manifest
> where we could see the improvements thanks to your patch ?

Unfortunately I believe that in practice both the improvements here, and in PR #14347, are *most likely* small enough to not really show up clearly in such tests :-(

/botio test

timvandermeij (approved):

timvandermeij:
Thanks!

PR #14355, opened 2021-12-09
Change `WorkerTransport.{pageCache, pagePromises}` from an Array to a Map

Snuffleupagus:
- Change `WorkerTransport.pageCache` from an Array to a Map

   Given that not all pages necessarily are being accessed, or that the pages may be accessed out of order, using a `Map` seems like a more appropriate data-structure here.
   For one thing, this simplifies iteration since we no longer have to worry about/check if `pageCache`-entries are undefined (which will happen for *sparse* `Array`s).

   Of particular note is that we're no longer attempting to "null" the `pageCache`-entry from within the `PDFPageProxy._destroy`-method. Given that *synchronous* JavaScript will always run to completion[1] and that we're looping through all pages in `WorkerTransport.destroy` and immediately clear the cache afterwards, that code did/does not really make a lot of sense (as far as I can tell).

   Finally, also changes the `pageCache` to a *private* property since it's not supposed to be accessed from the "outside".

   ---
   [1] Unless there are errors, of course.

- Change `WorkerTransport.pagePromises` from an Array to a Map

   Given that not all pages necessarily are being accessed, or that the pages may be accessed out of order, using a `Map` seems like a more appropriate data-structure here.

   Finally, also changes the `pagePromises` to a *private* property since it's not supposed to be accessed from the "outside".

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -1666,7 +1666,6 @@ class PDFPageProxy {
    */
   _destroy() {
     this.destroyed = true;
-    this._transport.pageCache[this._pageIndex] = null;
 
     const waitOn = [];
     for (const intentState of this._intentStates.values()) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #14358, opened 2021-12-10
Improve `PDFDocument.checkLastPage`/`Catalog.getAllPageDicts` for documents with corrupt XRef tables (PR 14311, 14335 follow-up)

Snuffleupagus:
- Improve `PDFDocument.checkLastPage` for documents with corrupt XRef tables (PR 14311, 14335 follow-up)

   Rather than trying, and failing, to fetch the entire /Pages-tree for documents with corrupt XRef tables, let's fallback to indexing all objects *before* trying to invoke the `Catalog.getAllPageDicts` method.

 - Update `Catalog.getAllPageDicts` to always propagate the actual Errors (PR 14335 follow-up)

   Rather than "swallowing" the actual Errors, when data fetching fails, ensure that they're always being propagated as intended to the call-site instead.
   Note that we purposely handle `XRefEntryException` specially, to make it possible to fallback to indexing all XRef objects.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -1224,7 +1224,7 @@ class Catalog {
    * Eagerly fetches the entire /Pages-tree; should ONLY be used as a fallback.
    * @returns {Map}
    */
-  getAllPageDicts() {
+  getAllPageDicts(recoveryMode = false) {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for doing this!

PR #14360, opened 2021-12-11

timvandermeij:


Snuffleupagus (approved):
r=me, with passing tests.

timvandermeij:
I'm not really sure why _one_ integration test is failing now and _only_ on Windows, but I can only assume it has something to do the Puppeteer upgrade. In general it's not great that the integration tests already suffer from intermittent failures (even aside from the Linux bot issue that is tracked upstream) :-(

timvandermeij:
I filed an issue for the Puppeteer 13 upgrade and left it out of this PR. Fortunately we could at least upgrade Puppeteer to 12.0.1, the latest 12.x version.

timvandermeij:
In https://github.com/mozilla/pdf.js/labels/1-dependencies we now have an issue/PR for all upgrades we couldn't do yet (the other packages are at the latest version now) and in https://github.com/mozilla/pdf.js/labels/1-test we now have an issue/PR for all test-related issues such as intermittents.

PR #14362, opened 2021-12-11
Support disabling of form editing when `pdfjs.enablePermissions` is set (issue 14356)

### LINKED ISSUE #14356: Firefox PDFViewer Allows Editing of Password Protected PDF Form Files!!!

Configuration:
- Web browser and its version: Firefox 95.0
- PDF.js version: Whatever is shipped with Firefox 95.0
- Is a browser extension:

1. Open file in Adobe Acrobat and get message that file is protected. Cannot change anything without password
2. Open file in Firefox 95.0. Allows free editing of form fields and on top the file can be saved.
3. Open the saved file from Firefox with Adobe Acrobat and you still get the password protected message but the values are changed!

Testfile
[testfile.pdf](https://github.com/mozilla/pdf.js/files/7692008/testfile.pdf)

What went wrong?
Firefox allows editing of protected PDF files!!!

Snuffleupagus:
For encrypted PDF documents without the required permissions set, this patch adds support for disabling of form editing. However, please note that it also requires that the `pdfjs.enablePermissions` preference is set to `true`[1] (since PDF document permissions could be seen as user hostile).

Based on https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#G6.1942134, this condition hopefully makes sense.

Fixes #14356

---
[1] Either manually with `about:config`, or using e.g. a [Group Policy](https://github.com/mozilla/policy-templates).

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -51,7 +51,6 @@ import {
   MissingPDFException,
   OPS,
   PDFWorker,
-  PermissionFlag,
   shadow,
   UnexpectedResponseException,
   UNSUPPORTED_FEATURES,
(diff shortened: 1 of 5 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

zorker:
The PDF has the following given permissions:

grafik

The document can be printed, content can be copied and it is accessible for screen readers.

Not sure what "document without required permission set" means in the description above.
There is a defined set of permissions given.

The enduser must not edit the content of the PDF.

What does the config entry pdfjs.enablePermissions do when manually set to true?

PR #14364, opened 2021-12-11
Only call `PDFDocumentProxy.getPermissions`, in the viewer, when `pdfjs.enablePermissions` is set (PR 14362 follow-up)

Snuffleupagus:
By making this API-call *unconditionally*, we introduce a (slight) delay in the initialization of *all* documents.
That seems quite unfortunate, since `pdfjs.enablePermissions` is off by default, and it thus seem better only do the API-call when actually needed; sorry about this!


timvandermeij (approved):

timvandermeij:
Good point; even though it's a minor delay, it's better to avoid it. Thanks!

Issue #14293, opened 2021-11-19
Failure in integration test in linux bot


calixteman:
The logs
```
  Stack:
    Error: Protocol error (Target.createTarget): can't access property "closing", selectedTab is null _setPositionalAttributes@chrome://browser/content/tabbrowser-tabs.js:1123:11
    _insertTabAtIndex@chrome://browser/content/tabbrowser.js:3286:25
    addTab@chrome://browser/content/tabbrowser.js:2710:16
    addTab@chrome://remote/content/shared/TabManager.jsm:24:31
    createTarget@chrome://remote/content/cdp/domains/parent/Target.jsm:90:28
    execute@chrome://remote/content/cdp/domains/DomainCache.jsm:101:25
    execute@chrome://remote/content/cdp/sessions/Session.jsm:64:25
    onPacket@chrome://remote/content/cdp/CDPConnection.jsm:248:36
    onMessage@chrome://remote/content/server/WebSocketTransport.jsm:89:18
    handleEvent@chrome://remote/content/server/WebSocketTransport.jsm:71:14
    openPromptSync@resource://gre/modules/Prompter.jsm:1057:17
    alert@resource://gre/modules/Prompter.jsm:1329:17
    alert@resource://gre/modules/Prompter.jsm:80:7
    _displayPrintingError@chrome://global/content/printUtils.js:572:21
```
My feeling is that something is wrong with the printing test and an alert is generated which blocks following attempts to get the current selected tab (see https://searchfox.org/mozilla-central/source/toolkit/components/printing/content/printUtils.js#572).

calixteman:
When I wrote the printing test I had to play with some firefox options:
https://github.com/mozilla/pdf.js/blob/c6cb39ef3091fa41dce0be44af639cffeb048d61/test/test.js#L929-L934
maybe one must be changed or we must add a new one.

timvandermeij:
Given that this happened very recently, most likely in an upstream nightly build that we use in the tests, the regression range should be limited to a few days. Maybe we can find out if changes relating to printing were merged upstream, in e.g., Bugzilla or with `mozregression`?

Snuffleupagus:
https://bugzilla.mozilla.org/show_bug.cgi?id=1741698 looks somewhat relevant for us as well, since based on quickly skimming through that one it seems that `print.always_print_silent = true` is now being ignored.

timvandermeij:
[...]

PR #14367, opened 2021-12-12
Disable failing print actions integration test in Firefox

timvandermeij:
Once the upstream bug is fixed it can be enabled again because it's causing way too much noise now. This is tracked in issue #14293. Note that I deliberately added a new block so we can easily remove it later on and because the other block is about another bug.


calixteman (approved):
Thx.

timvandermeij:
Cool, now that it's passing again we can keep a better eye on intermittents.

PR #14372, opened 2021-12-14
Move the /Lang handling into the `BaseViewer` (PR 14114 follow-up)

Snuffleupagus:
In PR #14114 this was only added to the default viewer, which means that in the viewer components the user would need to *manually* implement /Lang handling. This was (obviously) a bad choice, since the viewer components already support e.g. structTrees by default; sorry about overlooking this!

To avoid having to make *two* `getMetadata` API-calls[1] very early during initialization, in the default viewer, the API will now cache its result. This will also come in handy elsewhere in the default viewer, e.g. by reducing parsing when opening the "document properties" dialog.

---
[1] This not only includes a round-trip to the worker-thread, but also having to re-parse the /Metadata-entry when it exists.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -2406,6 +2406,8 @@ class WorkerTransport {
 
   #pagePromises = new Map();
 
+  #metadataPromise = null;
+
   constructor(messageHandler, loadingTask, networkStream, params) {
     this.messageHandler = messageHandler;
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for the follow-up!

PR #14380, opened 2021-12-15
Move the `EventBus`, and related functionality, into its own file

Snuffleupagus:
The size of the `web/ui_utils.js` file has increased over time, as more code has been added to (or moved into) that file. To reduce its size slightly, this patch moves the event-related functionality into a separate file.


timvandermeij (approved):

timvandermeij:
Thanks; I like it!

PR #14387, opened 2021-12-18
Modernize the test utilities

timvandermeij:
_The commit messages contain more information about the individual changes._

This is another step in the direction of modernizing our test infrastructure and enabling the `no-var` ESLint rule in more places. This can be tested locally using the `gulp makeref` command; I would suggest to minimize `test/test_manifest.json` first for a short runtime and temporarily move any existing reference images to another directory. Running `gulp makeref` will prompt at the end and if you answer "n", it will show the other prompt once `gulp makeref` is re-run.


Snuffleupagus (approved):
r=me, since I'm assuming that you've tested this very carefully locally and given that it looks totally reasonable!
Being able to remove a bunch of old custom code is always nice; thank you!

PR #14391, opened 2021-12-22
Handle non-integer Annotation border widths correctly (issue 14203) 

KouWakai:
The border width is expected to be 3 but tracing Class Annotation, I found the width actually comes as float 2.99998 so when the width passed to Number.Isinteger in Class AnnotationBorderStyle:setWidth() the width gets lost.

Number.Isinteger makes sure that the value should be integer.Taking it into account,adding Math.round can solve this issue I think.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -922,7 +922,7 @@ class AnnotationBorderStyle {
       this.width = 0; // This is consistent with the behaviour in Adobe Reader.
       return;
     }
-    if (Number.isInteger(width)) {
+    if (typeof width === "number") {
       if (width > 0) {
         const maxWidth = (rect[2] - rect[0]) / 2;
         const maxHeight = (rect[3] - rect[1]) / 2;
--- a/test/pdfs/issue14203.pdf.link
+++ b/test/pdfs/issue14203.pdf.link
@@ -1,0 +1,1 @@
+https://github.com/mozilla/pdf.js/files/7428963/pdfcomment.pdf
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -4212,6 +4212,15 @@
        "rounds": 1,
        "type": "eq"
     },
+    {  "id": "issue14203",
+       "file": "pdfs/issue14203.pdf",
+       "md5": "26c1c1e7341bc23fb28d805fa516dbbe",
+       "link": true,
+       "rounds": 1,
+       "firstPage": 11,
+       "lastPage": 11,
+       "type": "eq"
+   },
     {  "id": "issue11045",
        "file": "pdfs/issue11045.pdf",
        "md5": "101d4cb649cc006e0f2b14923e8d97d6",
--- a/test/unit/annotation_spec.js
+++ b/test/unit/annotation_spec.js
@@ -414,10 +414,13 @@ describe("annotation", function () {
 
   describe("AnnotationBorderStyle", function () {
     it("should set and get a valid width", function () {
-      const borderStyle = new AnnotationBorderStyle();
-      borderStyle.setWidth(3);
+      const borderStyleInt = new AnnotationBorderStyle();
+      borderStyleInt.setWidth(3);
+      const borderStyleNum = new AnnotationBorderStyle();
+      borderStyleNum.setWidth(2.5);
 
-      expect(borderStyle.width).toEqual(3);
+      expect(borderStyleInt.width).toEqual(3);
+      expect(borderStyleNum.width).toEqual(2.5);
     });
 
     it("should not set and get an invalid width", function () {


KouWakai:
This is my first time Pull request in my life time.I read code contribution guide carefully but I apologize if there are any mistakes in this PR.

src/core/annotation.js
```diff
@@ -922,7 +922,7 @@ class AnnotationBorderStyle {
       this.width = 0; // This is consistent with the behaviour in Adobe Reader.
       return;
     }
-    if (Number.isInteger(width)) {
+    if (Number.isInteger(Math.round(width))) {
```
Snuffleupagus, line 925: I don't believe that this is entirely correct, since neither https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=392 nor https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#G11.2096210 explicitly say that the border width has to be an *integer*.
More likely the existing PDF.js code is actually wrong here, and we should accept any *number* instead; i.e. `typeof width === "number"` here.

Snuffleupagus (changes requested):
In addition to the comment, please write a better *commit message* (and PR title/description) since as-is it's really quite difficult to understand this patch without looking at both the code and the relevant issue. (The first line of the commit message should be a short summary of the changes, and then you should also include additional information such as e.g. links to the specification and a good description of what you're changing and why.)

Furthermore, please add an `eq` test for these changes such that we avoid future regressions.

Also, please make sure that you follow *all of* https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing locally to ensure that the patch passes all existing tests.

src/core/annotation.js
KouWakai: Thank you for the feedback. I'll write a better commit message according to your advice.

It’s exactly as you said,border width can be float as well so I'll change the code to `typeof width === "number"` and add an `eq` test.

timvandermeij: > More likely the existing PDF.js code is actually wrong here

You're right, and this PR is indeed in the right direction. In the documents we had during the implementation and that we saw so far this has always been an integer, but the specification states "number", which we falsely assumed to be "integer". I just had another look and found the following note (note 2) in section 7.3.3 of the specification at https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=22&zoom=auto,-215,391:

> Throughout this standard, the term number refers to an object whose type may be either integer or real.

In short, this PR is fixing the issue in the right place compared to other attempts, and I completely agree with the suggestions from @Snuffleupagus.

KouWakai: Thank you for the review.
>Aside from the eq test, it would also be good to adjust the already existing unit tests to check that float inputs are also handled correctly.

Considering your advice I have checked `test/pdf` and there is [annotation-border-styles.pdf](https://github.com/mozilla/pdf.js/files/7769542/annotation-border-styles.pdf) (attached) in the directory it is pdf with annotation borders and there are lot of blank space.

I thought it would be good to put an annotation with a non-integer border width in it.But I think it would be good to add pdf which name is `annotation-border-width.pdf ` because it fully describes what the pdf stands for so I would add an pdf.

Snuffleupagus: > Considering your advice I have checked `test/pdf` and there is [annotation-border-styles.pdf](https://github.com/mozilla/pdf.js/files/7769542/annotation-border-styles.pdf) (attached) in the directory.It is pdf with annotation borders and there are lot of blank space in it.

Please do *not* update an existing test document, since there's a very real possibility that you could either break the file or possibly no longer test the originally intended thing.
Instead, please add a new *linked* test-case using the PDF document from issue #14203 and a `test_manifest.json` entry looking something like this:
```
{  "id": "issue14203",
   "file": "pdfs/issue14203.pdf",
   "md5": "26c1c1e7341bc23fb28d805fa516dbbe",
   "link": true,
   "rounds": 1,
   "firstPage": 11,
   "lastPage": 11,
   "type": "eq"
},
```

Snuffleupagus:
For the *commit message*, I'd suggest the following wording:

```
Handle non-integer Annotation border widths correctly (issue 14203)

The existing code appears to be wrong, since according to the PDF specification the border width of an Annotation only has to be a number and not specifically an integer. Please see:
 - https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#page=392
 - https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#G11.2096210
 - https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf#G6.1965562
```

test/pdfs/issue14203.pdf.link
```diff
@@ -0,0 +1 @@
+https://github.com/mozilla/pdf.js/files/7428963/pdfcomment.pdf
```
Snuffleupagus, line 1: Nit: Missing a line-break at the end of the line.

KouWakai: InsertFinalNewline of my editor was turned off,I would correct it and push again

Snuffleupagus:
Note that the test results above look good, despite the failures (those tests unfortunately fail intermittently).

https://github.com/mozilla/pdf.js/pull/14391#discussion_r774094303 also asked for a new unit-test as well (since you're new to the code-base I'll make a suggestion to help speed this up), and I'd thus suggest changing https://github.com/mozilla/pdf.js/blob/41dab8e7b6c1e2684d4afabb8f02e40a874d8e85/test/unit/annotation_spec.js#L417-L420
to the following instead:
```javascript
const borderStyleInt = new AnnotationBorderStyle();
borderStyleInt.setWidth(3);
const borderStyleNum = new AnnotationBorderStyle();
borderStyleNum.setWidth(2.5);

expect(borderStyleInt.width).toEqual(3);
expect(borderStyleNum.width).toEqual(2.5);
```

Snuffleupagus (approved):

KouWakai:
Thank you for the suggestion and approved my commit.
I understand where to add unit tests and  learned great example of commit message.

timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #14409, opened 2021-12-29
Improve caching for the `Catalog.getPageIndex` method (PR 13319 follow-up)

Snuffleupagus:
This method is now being used a lot more, compared to when it's added, since it's now used together with scripting as part of the `PDFDocument.fieldObjects` parsing (called during viewer initialization).
For /Page Dictionaries that we've already parsed, the `pageIndex` corresponding to a particular Reference is already known and we're thus able to skip *all* parsing in the `Catalog.getPageIndex` method for those cases.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -1099,7 +1099,8 @@ class Catalog {
       visitedNodes.put(pagesRef);
     }
     const xref = this.xref,
-      pageKidsCountCache = this.pageKidsCountCache;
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

Snuffleupagus:
> Thank you for improving this!

As always, thanks for reviewing/merging!

> Since 2021 ends in a few hours, I'd also like to wish you a happy new year!

Happy new year to you as well :-)

PR #14413, opened 2022-01-02
Drop the beta logic from the Gulpfile/website/`pdfjs.config` file

timvandermeij:
From now on we only make stable releases, so the beta logic should be removed to simplify the code. To test this, you can locally run `gulp wintersmith` to build the website and `gulp publish` to check that the `pdfjs.config` file is adjusted correctly. The website can also be viewed in the browser at http://54.241.84.105:8877/752a0925ca20096/getting_started.


Snuffleupagus (approved):
Looks good to me; thanks for simplifying this!

PR #14418, opened 2022-01-04
Use positive dimensions for text chunks in the text layer (issue #14415)

### LINKED ISSUE #14415: Text layer not well positioned on rotated page by 180

Configuration:
- Using Firefox and the demo viewer

[PDFJS_BUG.pdf](https://github.com/mozilla/pdf.js/files/7803096/PDFJS_BUG.pdf)

1. Open the document
2. Ctrl + A to select all the text
3. Areas selected are off

If we open the document in Chrome or Acrobat we will have right areas displayed on the text
image

What went wrong? 
The text layer is not well rendered.
image

This bug only happened on page rotated by 180

calixteman:
Width/height for text chunks must be positive numbers.


test/pdfs/issue14415.pdf.link
```diff
@@ -0,0 +1 @@
+https://github.com/mozilla/pdf.js/files/7803096/PDFJS_BUG.pdf
```
Snuffleupagus, line 1: This document looks small/simple enough, without any "sensitive" information, hence I think that you could just include directly in the repository instead.

calixteman:
Just adding `Math.abs` on dims was not enough: no space were added between "This" and "agreement" because of the orientation (rotation == 180 and font size is negative).

Snuffleupagus:
There seem to be some regressions in `taro-text-page3` (and `taro-text-enhance-page3`), for all platforms and browsers.

calixteman:
This is because of:
https://github.com/mozilla/pdf.js/pull/14418/files#diff-cf40014d9c7d352f3ae212e7fffe5386decf96a10e53066a4d900c3eb8fc559eR2479
I changed it because it's inconsistent with:
https://github.com/mozilla/pdf.js/blob/568633cf62a9a9f8d37880fda4e015b4ad4c6fc1/src/core/evaluator.js#L2475

So for example, if you look on the fifth column on page 3:
image
before my patch we've a white, a break and a chunk with Japanese glyphes, and after my patch we've a chunk with a white and Japanese glyphes, the position of the span changed because the first char in the span is not the same.

Anyway, this change isn't directly related to what this patch is aimed to fix (I just noticed the error in reading the code and I thought that it doesn't hurt to fix a "typo").
So I'll remove it and I'll fix the issue in an other patch.

PR #14424, opened 2022-01-06
[api-minor] Move `addLinkAttributes`, `LinkTarget`, and `removeNullCharacters` into the viewer (PR 14092 follow-up)

Snuffleupagus:
The purpose of this patch-series is to move some (now) viewer-specific functionality from the API (and worker) and into the viewer itself, in addition to some general `PDFLinkService` clean-up. *Please refer to the individual commit messages for additional details.*


timvandermeij (approved):

timvandermeij:
Nice work!

skjnldsv:
Hi, How can I access `pdfjsLib.LinkTarget.BLANK` now that this PR is perged and released?
We used to do `PDFViewerApplicationOptions.set('externalLinkTarget', pdfjsLib.LinkTarget.BLANK)` which seems like a better way than using integers straight away :)

Thanks!

Issue #14426, opened 2022-01-07
Highlight went wrong when the text font-size less than the minimum font-size of the browser


helianthuswhite:
Attach (recommended) or Link to PDF file here:

Configuration:
- Web browser and its version: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36
- Operating system and its version: Intel Mac OS X 10_15_7
- PDF.js version: the latest
- Is a browser extension: no

Steps to reproduce the problem:
1. open the demo in chrome https://mozilla.github.io/pdf.js/web/viewer.html
2. search any word in the pdf, such as 'JavaScript'
3. zoom out the pdf until the minimum font-size less than 12px, and you will see the problem

What is the expected behavior? (add screenshot)
![image](https://user-images.githubusercontent.com/12878076/148487417-63af7e0d-2057-4cd8-8471-5b7e2bb5c6ee.png)

What went wrong? (add screenshot)
![image](https://user-images.githubusercontent.com/12878076/148487329-b5af4efe-00c5-447c-8c2e-80fba1059278.png)

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): 
https://mozilla.github.io/pdf.js/web/viewer.html

helianthuswhite:
I have created a pr for this #14427 , thank you for feedback.

Snuffleupagus:
> Web browser and its version: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36

Unfortunately it's not *immediately* clear exactly which browser this actually is; I'm assuming you mean Google Chrome here.

>  PDF.js version: the latest

Sorry, but "the latest" isn't helpful; please always provide an actual version *number*.

> zoom out the pdf until the minimum font-size less than 12px, and you will see the problem

I cannot reproduce this (when testing on Windows) in either Mozilla Firefox Nightly `97.0a1` or Google Chrome Beta `97.0.4692.71` *unless* I manually specify a minimum font-size in the browser settings (that part wasn't entirely clear from the STR).

> I have created a pr for this #14427 , thank you for feedback.

[...]

PR #14427, opened 2022-01-07
fix text layer transform when font-size less than the browser's min font-size

helianthuswhite:
#14426

### DIFF
--- a/src/display/display_utils.js
+++ b/src/display/display_utils.js
@@ -631,6 +631,21 @@ function getXfaPageViewport(xfaPage, { scale = 1, rotation = 0 }) {
   });
 }
 
+/**
+ * get minimum font size
+ * @returns the minimum font size of browser
+ */
+function getMinFontSize() {
+  const div = document.createElement("div");
+  div.style.fontSize = "1px";
+  div.style.visibility = "hidden";
+  document.body.appendChild(div);
+  const computedStyle = window.getComputedStyle(div);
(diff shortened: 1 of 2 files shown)


Snuffleupagus (changes requested):
As mentioned in https://github.com/mozilla/pdf.js/issues/14426#issuecomment-1007235945, this feels a little "heavy" as far as implementations go.
Also, when manually testing this code it doesn't seem to work as expected in Firefox (which is the primary development target) since `1` is always returned regardless of the browser setting.

Snuffleupagus:
Looking at this again, the following issues are present:
 - This will add too much runtime overhead, by continually re-computing this for every textLayer element. While that could possibly be solved with caching, there's a couple of other issues as well.
 - Regardless of the minimum font-size set in Mozilla Firefox, this function always seem to return `1`. Given that Firefox is the *primary* development target, we unfortunately cannot accept a patch that doesn't work in Firefox while also adding (general) runtime overhead.
 - The helper function doesn't really support an unset minimum font-size (which is the default in browsers), since it'll always fallback to return `1` even when that's not correct. At very low zoom levels the textLayer may contain text with a font-size `< 1px`, which this patch thus won't handle correctly (and essentially break).

All-in-all, this patch unfortunately seems a bit too far from correct/working; hence let's close this PR and keep the issue open to track the problem.

helianthuswhite:
OK, thank you. I will have a think.

PR #14428, opened 2022-01-07
Use the correct dimension to know if we have to add an EOL in vertical mode

calixteman:


Snuffleupagus (changes requested):
As mentioned in PR #14418 this unfortunately causes a (pretty clear) regression, w.r.t. the positioning of some of the textLayer spans, on page 3 of the `TaroUTR50SortedList112.pdf` document.
Hence, on its own, I don't think that this patch is correct/enough to address this unfortunately.

Also, in the commit message you probably want to replace the word `had` with `add` instead?

calixteman:
> As mentioned in PR #14418 this unfortunately causes a (pretty clear) regression, w.r.t. the positioning of some of the textLayer spans, on page 3 of the `TaroUTR50SortedList112.pdf` document. Hence, on its own, I don't think that this patch is correct/enough to address this unfortunately.
> 
> Also, in the commit message you probably want to replace the word `had` with `add` instead?

The goal of the patch is just to fix an inconsistency.
For horizontal mode we've:
https://github.com/mozilla/pdf.js/blob/290cbc523281cb86fdf0117cf8f2cae521518fb9/src/core/evaluator.js#L2519-L2528
(we use advanceY and height)
when for vertical mode we've:
https://github.com/mozilla/pdf.js/blob/290cbc523281cb86fdf0117cf8f2cae521518fb9/src/core/evaluator.js#L2468-L2479
(we use advanceX, width and height)

Anyway, right now if you copy/paste the last column you'll get an extra EOL between the whitespace and `縦` which is wrong,
and the patch fixes this problem.
In order to fix the position of the text chunk we would need to split the chunk in 2 chunks (one with the whitespace and one with the visible chars) and then we'll get a span for the whitespace and an other one for the visible glyphes (which will be correctly positioned).

Snuffleupagus:
> The goal of the patch is just to fix an inconsistency.

I get that, but as mentioned (and evident by the test results) the patch causes a regression. Given that there's no other documents, that we know of, which are fixed by this patch I really don't think that it's OK to "purposely" introduce a regression here.

> Anyway, right now if you copy/paste the last column you'll get an extra EOL between the whitespace and `縦` which is wrong,
> and the patch fixes this problem.

Sure, but the regression is worse than that (comparatively) small problem as far as I'm concerned. As-is the textLayer no longer agrees with the rendered text, which would make it more difficult for users to actually select the *intended* text.

calixteman:
Agreed and the user selects the text generally to copy/paste it and so the EOL bug... and the circle is closed !

The whitespace on top of the last column is an ideographic space (https://www.compart.com/en/unicode/U+3000).
In the pdf we've two text runs for the last column: 1 for the ideographic space and one for the visible chars.
We ignore latin whitespace when they're in first position of a text run:
https://github.com/mozilla/pdf.js/blob/290cbc523281cb86fdf0117cf8f2cae521518fb9/src/core/evaluator.js#L2589
I suppose we should do the same for any kind of whitespace.

In Chrome, this space is stripped out but it's present in Acrobat.

An other possibility is to detect that the space and the rest aren't really on the same column (not the same X) and in this case push the chunk with space to have a new one for the rest. It'd make sense in horizontal mode too: create a new chunk each time we don't have the same Y.

@Snuffleupagus, what do you think ?

Snuffleupagus:
> We ignore latin whitespace when they're in first position of a text run:
> [...]
> I suppose we should do the same for any kind of whitespace.

I'd guess that'd require changing all of the existing `... = " "` checks, [in this code](https://github.com/mozilla/pdf.js/blob/290cbc523281cb86fdf0117cf8f2cae521518fb9/src/core/evaluator.js#L2588-L2592), into something like `WhitespaceRegexp.test(...)` instead?
If so my only concern with that approach would be regarding the possible performance impact, since regular expression matching is probably slower than string comparisons. Other than that, this sounds like a fairly simple solution.

> An other possibility is to detect that the space and the rest aren't really on the same column (not the same X) and in this case push the chunk with space to have a new one for the rest. It'd make sense in horizontal mode too: create a new chunk each time we don't have the same Y.

It sounds like that approach could potentially help improve other cases as well, although I suppose it could also lead to more textLayer elements being created (for some documents) since we'd break up existing text-runs more than we currently do.

calixteman:
> > We ignore latin whitespace when they're in first position of a text run:
> > [...]
> > I suppose we should do the same for any kind of whitespace.
> 
> I'd guess that'd require changing all of the existing `... = " "` checks, [in this code](https://github.com/mozilla/pdf.js/blob/290cbc523281cb86fdf0117cf8f2cae521518fb9/src/core/evaluator.js#L2588-L2592), into something like `WhitespaceRegexp.test(...)` instead? If so my only concern with that approach would be regarding the possible performance impact, since regular expression matching is probably slower than string comparisons. Other than that, this sounds like a fairly simple solution.
>

According to https://jsbench.me/b1ku0bggvi/1 (found in [stackoverflow](https://stackoverflow.com/questions/1496826/check-if-a-single-character-is-a-whitespace)), the fastest solution is the direct comparison.
In order to avoid to make the check again and again (the `i-1`,  `i+1` thing in the code you mentioned), we could just create and array containing `isWhitespace(glyph.unicode)` for each glyph.
The main advantage of this solution is that we would treat all the whitespaces in the same way which is consistent. So I'd be in favor of this solution.

> > An other possibility is to detect that the space and the rest aren't really on the same column (not the same X) and in this case push the chunk with space to have a new one for the rest. It'd make sense in horizontal mode too: create a new chunk each time we don't have the same Y.
> 
> It sounds like that approach could potentially help improve other cases as well, although I suppose it could also lead to more textLayer elements being created (for some documents) since we'd break up existing text-runs more than we currently do.

I suppose that most of the times all glyphs are on the same line... but who knows.

src/core/evaluator.js
```diff
@@ -2572,6 +2593,8 @@ class PartialEvaluator {
 
       const glyphs = font.charsToGlyphs(chars);
       const scale = textState.fontMatrix[0] * textState.fontSize;
+      const whitespaces = glyphs.map(glyph => isWhitespace(glyph.unicode));
```
Snuffleupagus, line 2596: *One question:* Rather than looping through the glyphs here, would it possibly make sense (and be more efficient) to add a new property (e.g. `isWhitespace` or similar) to the `Glyph`-instances and utilize the new helper function there instead to determine this for each glyph as part of the existing parsing in https://github.com/mozilla/pdf.js/blob/master/src/core/fonts.js#L3153?

That would obviously require extending the `Glyph` class, but given the existence of both a charCache *and* a glyphCache in the font code that should actually help reduce the overall amount of parsing required.

Snuffleupagus (changes requested):
Based on a *quick* look, at least `issue7878` seem to regress a bit with the latest version of the PR.

src/core/evaluator.js
calixteman: Good idea.

calixteman:
> Based on a _quick_ look, at least `issue7878` seem to regress a bit with the latest version of the PR.

The space between `Z` and `a` is a non-breakable space (0xA0) so before the patch it was considered as a "normal" char.
And since the chars in the string are slightly tightened we've a span by char:
https://github.com/mozilla/pdf.js/blob/290cbc523281cb86fdf0117cf8f2cae521518fb9/src/core/evaluator.js#L2524
https://github.com/mozilla/pdf.js/blob/290cbc523281cb86fdf0117cf8f2cae521518fb9/src/core/evaluator.js#L2210

With the patch, the 0xA0 is removed (it's the only one char in its text run) and when we add the `a` we compare its position with the `Z` and there is enough space for a white. Consequently it results in a chunk with `Z a` which is not consistent with the drawn glyphes... I'm sad...
I can cheat in removing the 0xA0 from the whitespace list...

I think this regression is acceptable compared to the "improvements" we have in some other cases: it's always this problem to find a good trade-off between number of chunks and chars positions in the text layer.

For your information, I've a wip locally to replace the html-based text layer by a svg-based one: it'll greatly improve the char positions, but I've few regressions I need to figure out.

src/core/fonts.js
```diff
@@ -212,6 +215,8 @@ class Glyph {
     this.operatorListId = operatorListId;
     this.isSpace = isSpace;
     this.isInFont = isInFont;
+    this.isWhitespace = isWhitespace(unicode);
```
Snuffleupagus, line 218: Given the existing `isSpace` property, could we perhaps call this new one e.g. `isUnicodeSpace` (or something) instead to more clearly distinguish them from each other?

Also, please update the name of the new helper function accordingly to avoid confusion with this already existing utility function:
https://github.com/mozilla/pdf.js/blob/8ac0ccc2277a7c0c85d6fec41c0f3fc3d1a2d232/src/core/core_utils.js#L247-L250

src/core/unicode.js
```diff
@@ -1640,10 +1640,32 @@ function reverseIfRtl(chars) {
...
+    c === " " ||
+    c === "\n" ||
+    c === "\t" ||
+    c === "\r" ||
+    c === "\f" ||
+    c === "\v" ||
+    c === "\u00a0" ||
+    c === "\u1680" ||
+    c === "\u2000" ||
+    c === "\u200a" ||
```
Snuffleupagus, line 1654: Isn't there a bunch of characters missing here, since https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Cheatsheet#character_classes contains the following information:
`[ \f\n\r\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]`

Note in particular the hyphen between `\u2000` and `\u200a`, which indicates a range of characters.

src/core/unicode.js
```diff
@@ -1640,7 +1640,17 @@ function reverseIfRtl(chars) {
   return buf.join("");
 }
 
+const SpecialCharRegExp = new RegExp("^(\\s)|(\\p{Mn})$", "u");
```
Snuffleupagus, line 1643: *Possibly a dumb question, but:* Is there a reason for using `^` and `$` explicitly, rather than simply using the global flag?

calixteman: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#return_value,
`g` mustn't be used in order to get capturing groups.

src/core/fonts.js
```diff
@@ -212,6 +213,10 @@ class Glyph {
     this.operatorListId = operatorListId;
     this.isSpace = isSpace;
     this.isInFont = isInFont;
+
+    const categories = checkCharUnicodeCategory(unicode);
```
Snuffleupagus, line 217: Nit: Perhaps `getCharUnicodeCategory` instead, since that more clearly (to me at least) suggests that the function returns an actual value.

Snuffleupagus: And, for consistency `const category = ...` instead here to not mix singular and plural in the names :-)

Snuffleupagus (approved):
r=me, since this seems like a good change overall by reducing the usage of regular expressions during text-extraction.

PR #14429, opened 2022-01-07
[JS] Fix few errors around AFSpecial_Keystroke

### LINKED ISSUE #14306: AForm validation rejects valid fields

calixteman:
- @cincodenada found some errors which are fixed in this patch;
 - it partially fixes issue #14306;
 - add some tests.

### DIFF
--- a/src/scripting_api/aform.js
+++ b/src/scripting_api/aform.js
@@ -433,11 +433,8 @@ class AForm {
     }
 
     psf = this.AFMakeNumber(psf);
-    if (psf === null) {
-      throw new Error("Invalid psf in AFSpecial_Format");
-    }
 
-    let formatStr = "";
+    let formatStr;
     switch (psf) {
       case 0:
         formatStr = "99999";
@@ -486,7 +483,7 @@ class AForm {
     ]);
 
     function _checkValidity(_value, _cMask) {
-      for (let i = 0, ii = value.length; i < ii; i++) {
+      for (let i = 0, ii = _value.length; i < ii; i++) {
         const mask = _cMask.charAt(i);
         const char = _value.charAt(i);
         const checker = checkers.get(mask);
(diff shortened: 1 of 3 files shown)


src/scripting_api/aform.js
```diff
@@ -564,7 +564,7 @@ class AForm {
           event.change.length +
           event.selStart -
           event.selEnd;
-        if (finalLen >= 8) {
+        if (finalLen > 8 || event.value[0] === "(") {
```
Snuffleupagus, line 567: Assuming that `event.value` is indeed a string, perhaps it'd be a little bit clearer to use `startsWith` for the comparison rather than array notation?

Snuffleupagus (approved):
Looking at the rest of the `AFSpecial_Keystroke` method (to try and understand this code better), I cannot help thinking that the following block is redundant:
https://github.com/mozilla/pdf.js/blob/290cbc523281cb86fdf0117cf8f2cae521518fb9/src/scripting_api/aform.js#L549-L551
given that we're throwing the exact same error in the `default` of the switch just below:
https://github.com/mozilla/pdf.js/blob/290cbc523281cb86fdf0117cf8f2cae521518fb9/src/scripting_api/aform.js#L576-L577

---

r=me, since the patch looks reasonable.

PR #14430, opened 2022-01-08
[JS] Use beforeinput event to trigger a keystroke event in the sandbox

### LINKED ISSUE #14306: AForm validation rejects valid fields

calixteman:
- it aims to fix issue #14307;
 - this event has been added recently in Firefox and we can now use it;
 - fix few bugs in aform.js or in annotation_layer.js;
 - add some integration tests to test keystroke events (see `AFSpecial_Keystroke`);
 - make dispatchEvent in the quickjs sandbox async.


calixteman:
There is definitely something wrong on linux. I wrote the tests under windows...
I'll investigate.

calixteman:
And of course I can't reproduce the issue locally on debian...

calixteman:
I'll try to figure out what's wrong in using a ssh -X on the bot.
I don't really understand how it can happen.

calixteman:
A summary:
 - I ran the tests locally and successfully on windows 10, windows 11, debian, ubuntu in windows 11 through wsl2;
 - I ran them on the linux bot using ssh -X in the ubuntu in windows 11: it was really really slow (the  bot is in US west coast and I connected it through the vpn located in Berlin) but of course it worked.

If anyone has an idea on what's wrong...

calixteman:
So it seems giving back the focus to the input after having dismiss the alert dialog is the right solution.
I'll update the patch tomorrow.

src/pdf.sandbox.external.js
```diff
@@ -157,6 +157,7 @@ class SandboxSupportBase {
         const event = new this.win.CustomEvent("updatefromsandbox", {
           detail: this.importValueFromSandbox(data),
         });
+
```
Snuffleupagus, line 160: Nit: Looks like an unrelated change.

web/generic_scripting.js
```diff
@@ -56,7 +56,7 @@ class GenericScripting {
 
   async dispatchEventInSandbox(event) {
     const sandbox = await this._ready;
-    sandbox.dispatchEvent(event);
+    setTimeout(() => sandbox.dispatchEvent(event), 0);
```
Snuffleupagus, line 59: Was this change necessary to reduce intermittent failures, or just left-over from previous debugging?

calixteman: I added it to reduce intermittent failures.
Locally, I had some issues with chrome integration tests, and the first fix I made was to dispatch the event (in the `beforeinput` listener) using a setTimeout. After thinking about that, in Firefox, the event is dispatched in the sandbox asynchronously, so I thought that it wouldn't hurt to have same behaviour in the tests.

Snuffleupagus: OK, that was just a bit surprising since *naively* I'd have assumed that the previous line (where we await a Promise) already introduces asynchronicity in the event dispatching.

Anyway, if we're adding this additional "delay" to the event dispatching there's *potentially* a slightly larger risk than before that we're now attempting to dispatch an event after destruction have started.
Given all the various `dispatchEventInSandbox` call-sites it's difficult to easily tell what an error could lead to, so how about also changing
https://github.com/mozilla/pdf.js/blob/23b6fde9fc92b827f054e21978e32ce9c13dab82/src/pdf.sandbox.js#L106
into the following to prevent any future errors elsewhere:
```javascript
this.support?.callSandboxFunction("dispatchEvent", event); 
```

Snuffleupagus (approved):
r=me, with one final comment/question addressed; thank you!

calixteman:
When entering `123A` in the second field in the test file, when entering the `A`, I get (in Chrome):
```
BEFOREINPUT 123
dispatchEventInSandbox
<<< alert >>>
VALUE 123 (field value is set from the sandbox)
INPUT 123A (input event which should come after the above BEFOREINPUT)
```
and in Firefox:
```
BEFOREINPUT 123
INPUT 123A
dispatchEventInSandbox
<<< alert >>>
VALUE 123
```

When adding the `setTimeout`, in Chrome, the INPUT event is triggered before the alert.
I think when I release the key for `A` then a beforeinput is triggered (but the `A` is not added yet in the field), ...etc  and once the sandbox stuff is done + alert... then the `A` is finally added to the field and input is triggered. So the `setTimeout` likely helps to let the input stuff be done before triggering an event in the sandbox.

PR #14443, opened 2022-01-12
Prevent run-time errors in `BaseViewer` when it's falling back to `SimpleLinkService` (issue 14442, PR 14295 follow-up)

### LINKED ISSUE #14442: base_viewer.js fails with this.linkService._cachedPageNumber in v2.12.313

Configuration:
Firefox/Chrome, latest version
Linux,Windows,mac
- PDF.js version:
v2.12.313
- Is a browser extension:
no

1. create a custom viewer with 
```
async function getViewer() {
const viewer = await import(/* webpackChunkName: 'pdfviewer' */ 'pdfjs-dist/web/pdf_viewer');
const eventBus = new viewer.EventBus();
return viewer.PDFSinglePageViewer({
        container,
        eventBus,
        useOnlyCssZoom: USE_ONLY_CSS_ZOOM,
        // We can enable text/annotations layers, if needed
        textLayerFactory: new viewer.DefaultTextLayerFactory(),
        annotationLayerFactory: new viewer.DefaultAnnotationLayerFactory(),
        // textLayerMode: TextLayerMode.ENABLE,
        // currently the SVG renderer is not supported
        // and it has problems with telekom documents
        // so it will be disabled until it is feature complete
        // renderer: 'svg'
    });
}
```
2. console prints an error:
```
Unable to get page for page view TypeError: this.linkService._cachedPageNumber is not a function
    ensurePdfPageLoaded_fn base_viewer.js:1502
```
    
Looks like the problem is `IPDFLinkService` which is actually using the default one `this.linkService = options.linkService || new SimpleLinkService();` which does not have this function. of course we can use `PDFLinkService` which will probably work.

Snuffleupagus:
Fixes #14442

### DIFF
--- a/web/base_viewer.js
+++ b/web/base_viewer.js
@@ -1541,7 +1541,7 @@ class BaseViewer {
       if (!pageView.pdfPage) {
         pageView.setPdfPage(pdfPage);
       }
-      if (!this.linkService._cachedPageNumber(pdfPage.ref)) {
+      if (!this.linkService._cachedPageNumber?.(pdfPage.ref)) {
         this.linkService.cachePageRef(pageView.id, pdfPage.ref);
       }
       return pdfPage;


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #14446, opened 2022-01-13
Expose even more API-functionality in the TypeScript definitions (issue 14435, PR 14013 follow-up)

### LINKED ISSUE #14435: Include more type definitions in pdfjs-dist
Hello!

Let me know if i should have opened this issue on the [pdfjs-dist](https://github.com/mozilla/pdfjs-dist) repo but I though here would fit.

I'm using pdfjs-dist and I'm in need of some more types to be exported. I have no idea how the types end up in the dist-repo but would love to open a PR if only someone could point me in the right direction 😃 Since I was using an older version of pdfjs-dist I realize the only type I need that's not in the type defs in the latest release are _PageViewport_ from _[display/display_utils](https://github.com/mozilla/pdf.js/blob/master/src/display/display_utils.js#L153)_.

Snuffleupagus:
While `PageViewport` apparently makes sense in TypeScript environments, given that it's being returned by the `PDFPageProxy.getViewport`-method in the API, we really don't want to extend the *public* API by simply exporting the class directly in `src/pdf.js` since it should never be called/initialized manually.
Hence we follow the same pattern as in PR #14013, and also extend the API unit-tests to ensure that `PDFPageProxy.getViewport` always returns a `PageViewport`-instance as expected.

Fixes #14435

### DIFF
--- a/src/pdf.js
+++ b/src/pdf.js
@@ -18,6 +18,7 @@
 /** @typedef {import("./display/api").PDFDocumentProxy} PDFDocumentProxy */
 /** @typedef {import("./display/api").PDFPageProxy} PDFPageProxy */
 /** @typedef {import("./display/api").RenderTask} RenderTask */
+/** @typedef {import("./display/display_utils").PageViewport} PageViewport */
 
 import {
   AnnotationMode,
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

PR #14450, opened 2022-01-14
Add (basic) UTF-8 support in the `stringToPDFString` helper function (issue 14449)

### LINKED ISSUE #14449: wrong UTF-8 display for PDF-2 document

[Actually downloaded from <https://github.com/pdf-association/pdf20examples/raw/master/pdf20-utf8-test.pdf>](https://github.com/mozilla/pdf.js/files/7871477/pdf20-utf8-test.pdf)

Configuration:
- Web browser and its version: _Firefox 96_.
- Operating system and its version: _Windows 7_.
- PDF.js version: the one that comes with _Firefox 96_.
- Is a browser extension: I think so.

1. Open the file metadata, bookmarks and layers are displayed wrong.
2. The explanation on how this works for PDF-2 seems to be https://www.pdfa.org/understanding-utf-8-in-pdf-2-0/. 

Right character display.

Snuffleupagus:
This patch implements this by looking for the UTF-8 BOM, i.e. `\xEF\xBB\xBF`, in order to determine the encoding.[1]
The actual conversion is done using the `TextDecoder` interface, which should be available in all environments/browsers that we support; please see https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder#browser_compatibility

Fixes #14449

---
[1] Assuming that everything lacking a UTF-16 BOM would have to be UTF-8 encoded really doesn't seem correct.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -957,25 +957,38 @@ const PDFStringTranslateTable = [
 function stringToPDFString(str) {
   const length = str.length,
     strBuf = [];
+  // UTF-16BE BOM
   if (str[0] === "\xFE" && str[1] === "\xFF") {
-    // UTF16BE BOM
     for (let i = 2; i < length; i += 2) {
       strBuf.push(
         String.fromCharCode((str.charCodeAt(i) << 8) | str.charCodeAt(i + 1))
       );
     }
-  } else if (str[0] === "\xFF" && str[1] === "\xFE") {
-    // UTF16LE BOM
+    return strBuf.join("");
+  }
+  // UTF-16LE BOM
+  if (str[0] === "\xFF" && str[1] === "\xFE") {
     for (let i = 2; i < length; i += 2) {
       strBuf.push(
         String.fromCharCode((str.charCodeAt(i + 1) << 8) | str.charCodeAt(i))
       );
     }
-  } else {
-    for (let i = 0; i < length; ++i) {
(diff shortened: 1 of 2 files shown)


calixteman:
According to:
https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/encoding
it should be possible to use it to for utf-16 (le/be) which would be more consistent.

Snuffleupagus:
> According to:
> https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/encoding
> it should be possible to use it to for utf-16 (le/be) which would be more consistent.

Sure, let's try that instead; however I did that change in a *separate* commit since it's not directly related to the issue at hand.

calixteman (approved):
Looks good overall.
I wonder if it could be interesting to add a "fast" path with a single check on `str[0]` (something like `if (str[0] >= "\xEF") { /* check for encoding */ }` in order to avoid 4 checks before reaching the general case (note that I've absolutely no opinion, I'm just wondering).

Snuffleupagus:
> Looks good overall. I wonder if it could be interesting to add a "fast" path with a single check on `str[0]` (something like `if (str[0] >= "\xEF") { /* check for encoding */ }` in order to avoid 4 checks before reaching the general case (note that I've absolutely no opinion, I'm just wondering).

Well, it really cannot hurt so why not :-)

/botio unittest

PR #14453, opened 2022-01-15
Dispatch a "documenterror" event in `PDFViewerApplication._documentError` (issue 14451)

### LINKED ISSUE #14451: Catch Load Errors
When using viewer.js code in an iframe (`web/viewer.html?file=my.pdf"`) and getting a load error like `Error: file origin does not match viewer's` how do I catch this error?  The console reports `app.js:2140 Uncaught (in promise) Error: file origin does not match viewer's`.  I have setup the load listener and am using the `PDFViewerApplication.initializedPromise`.  I was hoping the `PDFViewerApplication.initializedPromise` would reject, but it does not.  I could not find an eventbus event to listen to specifically for load errors either.

I would like to programmatically detect such load errors and dynamically remove the iframe hosting the pdf viewer and instead replace it with my own error messaging.  I can't figure out how to do that.

Configuration:
- Any, but testing on Chrome 97.0.4692.71 64-bit
- Ubuntu 20.04
- PDF.js version:  v2.12.313
- Is a browser extension: No

1.  Purposely try to load a PDF in a way that would cause the `app.js:2140 Uncaught (in promise) Error: file origin does not match viewer's` error.

Snuffleupagus:
*Please note:* This is a tentative patch, since I don't know if this is deemed important enough to fix.

The new event could be seen as a *supplement* to the existing "documentinit" and "documentloaded" events, but for the case when a PDF document fails to load.
To make the "documenterror" event generally useful, it'll include both the localized error message as well as the original reason for the error (when that exists).

This (probably) fixes #14451

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -1057,6 +1057,12 @@ const PDFViewerApplication = {
     this._unblockDocumentLoadEvent();
 
     this._otherError(message, moreInfo);
+
+    this.eventBus.dispatch("documenterror", {
+      source: this,
+      message,
+      reason: moreInfo?.message ?? null,
+    });
   },
 
   /**


timvandermeij (approved):

timvandermeij:
The change is small enough and I can see it being a somewhat useful supplement to `documentloaded`. Thanks!

PR #14454, opened 2022-01-15
Replace some `assert` usage with `unreachable` in the `src/shared/util.js` file

Snuffleupagus:
Inlining the checks should be a *tiny bit* more efficient, since it avoids have to make *unconditional* function calls in these fairly commonly used helper functions.


timvandermeij (approved):

timvandermeij:
This can't hurt indeed; thanks!

PR #14455, opened 2022-01-15
Make the `normalizeWhitespace` handling, in the `PartialEvaluator`, more efficient (PR 14428 follow-up)

Snuffleupagus:
After the changes in PR #14428 we can *directly*, and more efficiently, handle whitespace conversion in `PartialEvaluator.getTextContent` when the `normalizeWhitespace` option is being used.
This way we no longer need a separate helper function for this, and can avoid having to (again) iterate through the text and checking each character. Finally, this also removes the need for using a regular expression on e.g. all non-ASCII text.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -2174,7 +2174,6 @@ class PartialEvaluator {
     resources = resources || Dict.empty;
     stateManager = stateManager || new StateManager(new TextState());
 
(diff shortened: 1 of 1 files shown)


calixteman (approved):
It's a way better like this.

Snuffleupagus:
(Fixing a typo in the commit message, no actual code changes were made.)

/botio test

PR #14456, opened 2022-01-15
Font renderer - get int8 instead of uint8 in composite glyphes (bug 1749563)

### LINKED ISSUE #8754: Superscript Dot misplaced if Font Face disabled
[cover.pdf](https://github.com/mozilla/pdf.js/files/1203336/cover.pdf)

Configuration:
- Web browser and its version: Reproducible on Firefox 54.0.1, Safari 10.1.1
- Operating system and its version: macOS 10.12.5
- PDF.js version: 1.8.609
- Is an extension: No

1. Disable Font Face (via `PDFJS.disableFontFace = true`)
2. Observe positioning of Superscript Dot above characters `i`, `j`

screen shot 2017-08-06 at 19 49 31

screen shot 2017-08-06 at 19 45 17

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1749563;
 - use some helper functions to get (u|i)int** values in buffer: it helps to have a clearer code;
 - in composite glyphes the translations values with a transformations are signed so consequently get some int8 instead of uint8;
 - add few TODOs.

### DIFF
--- a/src/core/font_renderer.js
+++ b/src/core/font_renderer.js
@@ -25,17 +25,32 @@ import { getGlyphsUnicode } from "./glyphlist.js";
 import { StandardEncoding } from "./encodings.js";
 import { Stream } from "./stream.js";
 
-function getLong(data, offset) {
+// TODO: use DataView and its methods.
+
+function getUint32(data, offset) {
   return (
-    (data[offset] << 24) |
-    (data[offset + 1] << 16) |
-    (data[offset + 2] << 8) |
-    data[offset + 3]
+    ((data[offset] << 24) |
+      (data[offset + 1] << 16) |
+      (data[offset + 2] << 8) |
+      data[offset + 3]) >>>
+    0
   );
 }
 
-function getUshort(data, offset) {
+function getUint16(data, offset) {
   return (data[offset] << 8) | data[offset + 1];
+}
+
+function getInt16(data, offset) {
+  return ((data[offset] << 24) | (data[offset + 1] << 16)) >> 16;
+}
+
+function getInt8(data, offset) {
+  return (data[offset] << 24) >> 24;
+}
+
+function getFloat214(data, offset) {
+  return getInt16(data, offset) / 16384;
 }
 
 function getSubroutineBias(subrs) {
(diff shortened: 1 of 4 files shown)


src/core/font_renderer.js
```diff
@@ -25,7 +25,9 @@ import { getGlyphsUnicode } from "./glyphlist.js";
 import { StandardEncoding } from "./encodings.js";
 import { Stream } from "./stream.js";
 
-function getLong(data, offset) {
+// TODO: use DataView and its methods.
+
+function getUint32(data, offset) {
```
Snuffleupagus, line 30: Is this name completely accurate, i.e. shouldn't it be `getInt32` instead?

Please note that we have essentially identical code on the `Stream`-instances, but there we use the `getInt32` name:
https://github.com/mozilla/pdf.js/blob/749a3fa6994a656f2c9427cd932945168c74b943/src/core/base_stream.js#L70-L76
Furthermore, we have a helper function named `readUint32` which actually ensures that the value is in fact unsigned:
https://github.com/mozilla/pdf.js/blob/749a3fa6994a656f2c9427cd932945168c74b943/src/core/core_utils.js#L237-L245

Snuffleupagus (approved):

src/core/font_renderer.js
calixteman: Interesting...
You're right, it's really getInt32 here but in reading for example cmap specs:
https://docs.microsoft.com/en-us/typography/opentype/spec/cmap
and in particular:
https://docs.microsoft.com/en-us/typography/opentype/spec/cmap#encoding-records-and-encodings
we should really have an uint32 in some places (maybe all).
I suppose that in general it doesn't hurt to get an int32 because we don't have font files with several Gb but it's a bug.

calixteman: I updated the code to really have an uint32 instead of an int32.
I checked in parallel the specs to be sure that an uint32 is expected.

src/core/font_renderer.js
```diff
@@ -51,48 +66,49 @@ function getSubroutineBias(subrs) {
...
-        ranges[i].ids[j] = getUshort(data, p + idOffset);
+        ranges[i].ids[j] = getUint16(data, p + idOffset);
         idOffset += 2;
       }
     }
     return ranges;
   } else if (format === 12) {
-    getLong(data, start + offset + 4); // length
-    const groups = getLong(data, start + offset + 12);
+    // getUint32(data, start + offset + 4); // length
```
Snuffleupagus, line 102: If this isn't necessary any more, please don't leave commented-out code in the final patch.

src/core/font_renderer.js
```diff
@@ -25,19 +25,34 @@ import { getGlyphsUnicode } from "./glyphlist.js";
...
+function getInt16(data, offset) {
+  return ((data[offset] << 24) | (data[offset + 1] << 16)) >> 16;
+}
+
+function getInt8(data, offset) {
+  return (data[offset] << 24) >> 24;
+}
+
+function getFloat214(data, offset) {
+  return getInt16(data, offset) / 16384.0;
```
Snuffleupagus, line 53: Is the decimal really necessary here, since I believe that JS engines will just disregard it?

calixteman:
I restarted the process on the win bot.
/botio-windows test

PR #14457, opened 2022-01-16
Implement a unit test for `getCharUnicodeCategory` in `src/core/unicode.js` (PR 14428 follow-up)

timvandermeij:
Given that the other functions in this file are already covered by unit tests, we should also cover this newly added function.


Snuffleupagus (approved):
Looks good to me; thanks for improving the unit-test coverage!

PR #14460, opened 2022-01-16
canvas.js: adding this.current.lineWidth to calculate ctx.lineWidth (Bug 1743245)

quaoaris:
Fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1743245
Bug 1743245: Grided PDF file lines rendered too thick

The lineWidth defined in the pdf was not used to calculate the ctx.lineWidth.
Example pdf: https://bugzilla.mozilla.org/attachment.cgi?id=9255718

(as this is my first PR, please feel free to contact me, if I did a mistake)


Snuffleupagus (changes requested):
Have you successfully run the *entire* test-suite locally with this patch?
See https://github.com/mozilla/pdf.js/wiki/Contributing#4-run-lint-and-testing

Also, all patches affecting rendering must add a test-case to prevent future regressions.
In this particular case, directly attaching the PDF document seems fine (given that it's a reduced document without any personal information). Hence, in this case, simply using https://github.com/mozilla/pdf.js/blob/master/test/add_test.js should suffice.

calixteman:
Out of curiosity, did you git blame the line you're changing to understand why this line is like this ?
I don't say that what you're doing is wrong (maybe I made something wrong myself), but I think you would need to explain why it's better.

quaoaris:
Actually I thought, that the lineWidth should be reflecting the ctx.lineWidth of the pdf file (which is this.current.lineWidth). You put the ctx.lineWidth (in this special case, where the PixelWidth > pdf-lineWidth) to this._combinedScaleFactor. This means that you are assuming a height of 1 (in case of a parallelogram, indicated by a negative PixelWidth), which should be multiplied by this._combinedScaleFactor. I wondered why the value of the pdf file (this.current.lineWidth) is not used in this calculation? So I tried (my first commit) with lineWidth=this._combinedScaleFactor * this.current.lineWidth, which worked fine for the test pdf (of the bug) and many other pdfs as well.

Unfortunately one of the test pdfs showed a massive problem, so I checked the numbers (in my the test pdf) and figured out that a value of lineWidth of about 1 would be best for this setting.. (Actually I got a _combinedScaleFactor of 1.66 and a this.current.lineWidth of 0.5) 
So I calculated for other pdfs as well and finally run the test with the value of "1", which seemed to be fine. I know, that this is not the best way... so if you reject this commit because of this proceeding, it's fine for me.

calixteman:
@quaoaris, it was 1 in the past but because of https://github.com/mozilla/pdf.js/issues/12868 I changed it to be almost `1 * zoomFactor`.
So looking at the pdf in bugzilla bug, I've the feeling that using a `floor` instead of `round` could be a good solution.
Wdyt ?

quaoaris:
@calixteman Thanks for your support. I just changed it as you proposed and tested with the example.pdf as well as with the complete test suite.
What are the next steps to get this PR into the project?

calixteman:
There are few other places with a `round`:
https://github.com/react-pdf-viewer/pdf.js/commit/ec419bfbd77da48a99b8e4619ed53ce078390d6e

test/test_manifest.json
```diff
@@ -2069,6 +2075,12 @@
        "type": "eq",
        "enableAutoFetch": true
     },
+    {  "id": "Bug1743245_example",
+       "file": "pdfs/Bug1743245_example.pdf",
+       "md5": "01eadf118ec05fe02e91db63536d50f6",
+       "rounds": 1,
+       "type": "eq"
+    },
```
Snuffleupagus, line 2083: Please remove this duplicate entry.

test/test_manifest.json
```diff
@@ -1360,6 +1360,12 @@
        "enableXfa": true,
        "type": "eq"
     },
+    {  "id": "Bug1743245_example",
+       "file": "pdfs/Bug1743245_example.pdf",
```
Snuffleupagus, line 1364:
```javascript
    {  "id": "bug1743245",
       "file": "pdfs/bug1743245.pdf",
```
and please re-name the PDF file accordingly.

Also, you're missing an entry in the https://github.com/mozilla/pdf.js/blob/master/test/pdfs/.gitignore file (the `add_test.js` script should've updated that one as well).

Snuffleupagus (changes requested):
Please note that we don't allow merge commits in PRs, so those need to be removed; please refer to https://github.com/mozilla/pdf.js/wiki/Squashing-Commits
Also, please keep in mind that we also don't use separate fixup commits, and that all changes here should be folded into just *one* commit.

Finally, please remember to write a good commit message explaining both what's being changed and why. Also, please make sure that the first line of the commit message is a good (short) description of the changes (having basically a line of JavaScript code in the title is often not that helpful).

quaoaris:
Thanks for your support.. I think the best ist to close this PR and create a new one, with ONLY one commit and a better title and detailled summary etc.

Snuffleupagus:
> I think the best ist to close this PR and create a new one, with ONLY one commit and a better title and detailled summary etc.

In general, please *do not* do this since it loses some context from the discussion in the previous PR. Note that it should *always* be possible to use e.g. interactive rebase to fixup (and remove) unwanted commits.

Also, your new PR still doesn't really address https://github.com/mozilla/pdf.js/pull/14460#discussion_r791089822.

PR #14471, opened 2022-01-19
Update Puppeteer to version `13.1.1`

Snuffleupagus:
Note in particular https://github.com/puppeteer/puppeteer/releases/tag/v13.1.0 which includes an update to Chromium 98, which adds support for `structuredClone` in the browser.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -555,11 +555,19 @@ function createTestSource(testsName, { bot = false, xfaOnly = false } = {}) {
     console.log("### Running " + testsName + " tests");
 
     const PDF_TEST = process.env.PDF_TEST || "test_manifest.json";
+    let forceNoChrome = false;
     const args = ["test.js"];
     switch (testsName) {
       case "browser":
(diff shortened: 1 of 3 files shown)


timvandermeij:
The Chrome on Windows failure here is pretty strange since it seems to work fine up until a certain point where it starts failing every test with `Loading PDF document: Error: Setting up fake worker failed: "Cannot load script at: http://127.0.0.1:60301/build/generic/build/pdf.worker.js"`. It seems strange to me that it's using a fake worker in the first place since it's a real browser?

Snuffleupagus:
> since it seems to work fine up until a certain point

I don't think it actually works all that well since the tests before that run *extremely* slow.
In any case, any objections to simply going ahead with disabling the affected configuration (since it's unrelated to Firefox) to avoid us becoming "stuck" with dependencies that we cannot update?

timvandermeij:
Not really. Let's land it as part of #14392 and close this one.

PR #14474, opened 2022-01-20
Fix legacy build to be ES3 compatible

sithmel:
The legacy build should not include ES6 syntax, but currently the webpack preamble, contains an arrow function.
This change configures webpack to use a preamble that is ES3 compatible, only for the legacy build.


Snuffleupagus:
> The legacy build should not include ES6 syntax, but currently the webpack preamble, contains an arrow function.

Yes, and that should be OK since we no longer support any environments/browsers without e.g. arrow functions; please refer to https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support for the (current) support information.

> This change configures webpack to use a preamble that is ES3 compatible, only for the legacy build.

Sorry, but I really don't think that it makes sense for us to re-introduce support for old and discontinued browsers (since that'd unnecessarily increase the maintenance/support burden).

PR #14479, opened 2022-01-23

Snuffleupagus:


timvandermeij (approved):

timvandermeij:
Thanks! (Unrelated, but that's the first time in a long while I've seen one of the bots pass every single test.)

Snuffleupagus:
> Unrelated, but that's the first time in a long while I've seen one of the bots pass every single test.

Well, that's because it actually didn't :-)
Looking closely at the logs in https://github.com/mozilla/pdf.js/pull/14479#issuecomment-1019457343, focusing on the browser-test, it appears that Firefox didn't even start and somehow that wasn't picked up at all by the test-runner.

PR #14487, opened 2022-01-24
Remove the `out.pdf` file (PR 14430 follow-up)

Snuffleupagus:
It looks like this file was commited accidentally, since it contains *one* page from the default `tracemonkey` document.


calixteman (approved):
Ooops...

PR #14491, opened 2022-01-24
fix for lines (stroke) are rendered too thick  (Bug 1743245)

quaoaris:
This commit fixes Bug 1743245 (Grided PDF file lines rendered too thick) which was created by a fix for  #12868 .
The lineWidth was set to round(1 * this._combinedScaleFactor) when the pixel is drawn as a parallelorgam with a height <1. This fix changes this to floor(1*this._combinedScaleFactor) .

This change shows a visual result comparable to Chrome and Acrobat.
Regarding  #12868 : 3 statements in canvas.js are affected and will change with this commit (stroke and paintChar).
Full tests were successfully


test/test_manifest.json
```diff
@@ -4042,6 +4042,12 @@
        "type": "eq",
        "annotations": true
     },
+    {  "id": "Bug1743245_Example",
+       "file": "pdfs/Bug1743245_Example.pdf",
```
Snuffleupagus, line 4046: https://github.com/mozilla/pdf.js/pull/14460#discussion_r791089822 still applies here, and to the `.gitignore` file.

Snuffleupagus (changes requested):

test/test_manifest.json
quaoaris: I changed the file names to the naming convention, as mentioned.

calixteman (approved):
LGTM. Thank you for doing this.

PR #14495, opened 2022-01-25
web: Read WheelEvent.deltaMode before deltas.

### LINKED ISSUE #14476: Ctrl+mouse wheel zooming zooms in too quickly

Configuration:
- Web browser and its version: firefox 96.0
- Operating system and its version: Arch linux, sway (wayland)
- PDF.js version: Whatever came with the browser
- Is a browser extension: no

1. Open any pdf, can reproduce it on https://mozilla.github.io/pdf.js/web/viewer.html
2. Scroll in with ctrl+mouse wheel using a normal mouse, not a touch pad.

Expected behaviour:
It zooms in with small steps. It can for example be the same steps as ctrl+plus.

Behaviour:
It scrolls in and out way to quickly. It jumps in the steps: 10%, 70%, 150%, 300%, 570%, 1000%.

Notes:
Cannot reproduce on chromium with the web viewer on https://mozilla.github.io/pdf.js/web/viewer.html

emilio:
So that Firefox doesn't switch to pixel mode for compat with other
browsers.

This should fix https://github.com/mozilla/pdf.js/issues/14476, in terms
of restoring the previous behavior.

We probably want to change the pixel-based scrolling code to not scroll
so much (the deltaMode stuff normalizes to +/-1 tick for each wheel
event, perhaps the pixel-based value should do the same).

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -2691,13 +2691,17 @@ function webViewerWheel(evt) {
       return;
     }
 
+    // It is important that we query deltaMode before delta{X,Y}, so that
+    // Firefox doesn't switch to DOM_DELTA_PIXEL mode for compat with other
+    // browsers, see https://bugzilla.mozilla.org/show_bug.cgi?id=1392460.
+    const deltaMode = evt.deltaMode;
+    const delta = normalizeWheelEventDirection(evt);
     const previousScale = pdfViewer.currentScale;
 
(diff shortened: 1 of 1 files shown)


Snuffleupagus (approved):
Looks good to me, thank you for fixing this!

PR #14500, opened 2022-01-26
Take into account all rotations before comparing glyph positions

### LINKED ISSUE #14497: getTextContent Returns wrong text (with spaces removed)
In the attached image below:
1) **H2** is picked up correctly
2) **17,79** is picked up correctly
3) **256 36 30**  is picked up wrong as 2563630 (without spaces)

Before this used to work.  Now I am having trouble getting the correct text.  I tried setting the parameters below on getTextContent to true or false, but no luck.

- 	normalizeWhitespace
- 	disableCombineTextItems
- 	includeMarkedContent

image

The sample is on the attached PDF (Page 2)

image

[e335bb3d-9b84-47eb-8238-a903095668af.pdf](https://github.com/mozilla/pdf.js/files/7938019/e335bb3d-9b84-47eb-8238-a903095668af.pdf)

calixteman:
- it aims to fix #14497;
 - previously, only rotations with an angle 0, 90, 180 or 270 were taken into account;
 - so generalize to any angle but keep the fast path for 0, 90, ... because they're likely more common than anything else.


src/core/evaluator.js
```diff
@@ -2385,6 +2385,12 @@ class PartialEvaluator {
         });
     }
 
+    function applyInverseRotation(x, y, transf) {
+      const scale = Math.hypot(transf[0], transf[1]);
+      const matrix = transf.slice(0, 4).map(c => c / scale);
+      return [matrix[0] * x + matrix[1] * y, matrix[2] * x + matrix[3] * y];
```
Snuffleupagus, line 2391: It seems like this is adding some unnecessary parsing/allocations, since you first create a temporary Array and then iterate through it.
Could you not instead just do e.g. the follow, which ought to be a tiny bit more efficient from both a CPU/memory perspective?
```javascript
return [
  (transf[0] / scale) * x + (transf[1] / scale) * y,
  (transf[2] / scale) * x + (transf[3] / scale) * y
];
```

src/core/evaluator.js
```diff
@@ -2421,7 +2426,19 @@ class PartialEvaluator {
         rotate += currentTransform[1] > 0 ? 90 : 270;
```
Snuffleupagus, line 2426: I'm guessing this line might be connected to the test failures, since it no longer does the right thing with the new default value.

src/core/evaluator.js
```diff
@@ -2421,7 +2426,19 @@ class PartialEvaluator {
...
+        //  - remove the scale factor from the matrix to get a rotation matrix
+        //  - apply the inverse (which is the transposed) to the positions
+        // and we can then compare positions of the glyphes to detect
+        // a whitespace.
+        [posX, posY] = applyInverseRotation(posX, posY, currentTransform);
+        [lastPosX, lastPosY] = applyInverseRotation(
+          lastPosX,
+          lastPosY,
+          textContentItem.prevTransform
+        );
```
Snuffleupagus, line 2440: Can't we just add this as the `default` branch in the `switch` below, since that should be a tiny bit more efficient overall?

test/pdfs/.gitignore
```diff
@@ -506,3 +506,4 @@
 !poppler-90-0-fuzzed.pdf
 !issue14415.pdf
 !issue14307.pdf
+!issue14497.pdf
```
Snuffleupagus, line 509: Nit: Missing a new-line at the end of this file.

Snuffleupagus (changes requested):

Snuffleupagus (approved):
This looks much better now, thank you!

PR #14506, opened 2022-01-27
Update the year in the `license_header` files

Snuffleupagus:
This also includes a couple of files that are included as-is in the `pdfjs-dist` library.


calixteman (approved):
Thank you.

PR #14510, opened 2022-01-27
[api-minor] Annotations - Adjust the font size in text field in considering the total width (bug 1721335)

### LINKED ISSUE #14502: Font size is too big when printing pdfs with input text fields
[Pdfjs bug.pdf](https://github.com/mozilla/pdf.js/files/7952983/Pdfjs.bug.pdf)

Configuration:
- Web browser and its version: Firefox 96.0.2 (64-bit)
- Operating system and its version: MacOS Big Sur
- PDF.js version: Latest v2.12.313
- Is a browser extension: N/A

1. Open a pdf that has an input text field, 
2. Add some content to the input field
3. Try to print it, notice that the font size of the content added to the input fields is much bigger on the print preview, in many cases being cut from the printed version.

All the content added to the input field should be printed in a very similar way to the one being displayed during the edition.

Much bigger font being used during the print preview and final printed version as well, where in many cases the content gets cut.

Pdfjs bug - latest Firefox

calixteman:
- it aims to fix #14502 and bug 1721335;
 - Acrobat and Pdfium do the same;
 - it'll avoid to have truncated data when printed;
 - change the factor to compute font size in using field height: lineHeight = 1.35*fontSize
    - this is the value used by Acrobat.  
 - in order to not have truncated strings on the bottom, add few basic metrics for standard fonts.


calixteman:
I need to take into account how lines are broken.

src/core/standard_fonts.js
```diff
@@ -305,6 +305,93 @@ const getSymbolsFonts = getLookupTableFactory(function (t) {
   t.ZapfDingbats = true;
 });
 
+const getStdFontMetrics = getLookupTableFactory(function (t) {
```
Snuffleupagus, line 308: This honestly doesn't look exactly like data that belong in this file, and you probably want to use https://github.com/mozilla/pdf.js/blob/master/src/core/metrics.js instead.

Also, please let's improve the sorting of this data such that the various font-families have their data grouped together instead. I.e. something like:
```javascript
t["Times-Roman"] = { ... };
t["Times-Bold"] = { ... };
t["Times-BoldItalic"] = { ... };
t["Times-Italic"] = { ... };
t.Helvetica = { ... };
t["Helvetica-Bold"] = { ... };
...
```

src/core/fonts.js
```diff
@@ -1073,6 +1074,14 @@ class Font {
       nonStdFontMap[fontName] && stdFontMap[nonStdFontMap[fontName]]
     );
 
+    if (isStandardFont) {
+      const stdFontMetrics = getStdFontMetrics();
+      const metrics = stdFontMetrics[fontName];
+      this.ascent = metrics.ascent / PDF_GLYPH_SPACE_UNITS;
+      this.descent = metrics.descent / PDF_GLYPH_SPACE_UNITS;
+      this.capHeight = metrics.capHeight / PDF_GLYPH_SPACE_UNITS;
```
Snuffleupagus, line 1082: Is it *generally* appropriate to overwrite possibly existing values, based on the font dictionary, with `NaN` as will happen for some of the fonts (e.g. ZapfDingbats)?

Snuffleupagus (changes requested):
Before I attempt to review this, since it looks a bit complex, can we please run all tests to make sure there's no regressions?

src/core/standard_fonts.js
calixteman: I just copied/pasted the code here:
https://github.com/mozilla/pdf.js/blob/6b9cc24d49583c487d9267c98579fd4cd9db1704/src/core/standard_fonts.js#L23-L38

src/core/fonts.js
calixteman: I think there are no such metrics in pdb files but maybe I'm wrong.
Anyway I got the different values from:
https://github.com/Hopding/standard-fonts/tree/master/font_metrics

calixteman:
> Before I attempt to review this, since it looks a bit complex, can we please run all tests to make sure there's no regressions?

Sure.
Likely there are few regressions from print tests because I changed the way to compute font size for auto-sized fields.
FYI, I created a form with different text fields in acrobat and let it decides the font size, I get field height+font size and ran a linear regression to guess the factor 1.35.
When the text field isn't large enough, the font size is reduced in order to exactly fit the field. I checked how it is done in pdfium:
https://pdfium.googlesource.com/pdfium/+/refs/heads/main/core/fpdfdoc/cpvt_variabletext.cpp#770
and the result isn't the same as in Acrobat.

src/core/fonts.js
Snuffleupagus: > I think there are no such metrics in pdb files but maybe I'm wrong.

Please note that those values (initially) come from the /FontDescriptor dictionary in the PDF document itself (although for embedded font files we then override those after parsing the font data), via https://github.com/mozilla/pdf.js/blob/6b9cc24d49583c487d9267c98579fd4cd9db1704/src/core/fonts.js#L933-L936 and https://github.com/mozilla/pdf.js/blob/6b9cc24d49583c487d9267c98579fd4cd9db1704/src/core/evaluator.js#L4063-L4066

calixteman:
There is a failure with `unittestcli` at:
https://github.com/mozilla/pdf.js/blob/6b9cc24d49583c487d9267c98579fd4cd9db1704/test/unit/api_spec.js#L2013
and it's due to the fact that the font file is something and we don't enter in the then clause:
https://github.com/mozilla/pdf.js/blob/6b9cc24d49583c487d9267c98579fd4cd9db1704/src/core/fonts.js#L960
but with `unittest`, `file` is `null` here.
@Snuffleupagus do you have an idea why the behavior is different ?

Snuffleupagus:
> do you have an idea why the behavior is different ?

Probably because the `useSystemFonts` option has different default values in browsers/Node.js, see https://github.com/mozilla/pdf.js/blob/6b9cc24d49583c487d9267c98579fd4cd9db1704/src/display/api.js#L163-L167
The solution is likely to update the expected values in the unit-test to something like this (and given that these changes are thus API-observable I'm also tagging the PR accordingly):
```javascript
...
ascent: isNodeJS ? NaN : 0.683,
descent: isNodeJS ? NaN : -0.217,
```

Snuffleupagus:
> do you have an idea why the behavior is different ?

Probably because the `useSystemFonts` option has different default values in browsers/Node.js, see https://github.com/mozilla/pdf.js/blob/6b9cc24d49583c487d9267c98579fd4cd9db1704/src/display/api.js#L163-L167
The solution is likely to update the expected values in the unit-test to something like this (and given that these changes are thus API-observable I'm also tagging the PR accordingly):
```javascript
...
ascent: isNodeJS ? NaN : 0.683,
descent: isNodeJS ? NaN : -0.217,
```

src/core/annotation.js
```diff
@@ -1618,34 +1620,84 @@ class WidgetAnnotation extends Annotation {
...
-      // of a single-line field.
-      const FONT_FACTOR = 0.8;
       if (lineCount === -1) {
-        fontSize = roundWithOneDigit(FONT_FACTOR * height);
+        const textWidth = this._getTextWidth(text, font);
+        fontSize = roundWithTwoDigits(
+          Math.min(height / LINE_FACTOR, width / textWidth)
+        );
       } else {
+        const lines = text.split(/\r\n|\r|\n/);
```
Snuffleupagus, line 1651: It probably doesn't matter at all, but I'm guessing that the slightly more compact `/\r\n?|\n/` would also work?

src/core/fonts.js
```diff
@@ -1074,6 +1075,21 @@ class Font {
     );
 
     fontName = stdFontMap[fontName] || nonStdFontMap[fontName] || fontName;
+
+    const stdFontMetrics = getFontBasicMetrics();
```
Snuffleupagus, line 1079: Nit: Given that the name of the helper function was changed, maybe update this variable name accordingly?

Snuffleupagus (approved):
Tentative r=me, since this looks overall reasonable and the changes in the reference tests appear to be improvements.

*Edit:* Does this also fix https://bugzilla.mozilla.org/show_bug.cgi?id=1721335?

calixteman:
@Snuffleupagus do you know what's wrong here ?

PR #14516, opened 2022-01-30
[UI] Avoid to have buttons in hover state after having been clicked (bug 836732)

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=836732;
 - replace :focus by :focus-visible for the buttons in the UI, according to the docs:
   - https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible
   - the button has the focus-visible state when it has been focused with the keyboard

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -559,7 +559,7 @@ html[dir="rtl"] .findbar .splitToolbarButton > .findNext {
 }
 
 .findbar label:hover,
-.findbar input:focus + label {
+.findbar input:focus-visible + label {
   color: var(--toggled-btn-color);
   background-color: var(--button-hover-color);
 }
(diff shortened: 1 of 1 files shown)


Snuffleupagus (changes requested):
Don't you need to update https://github.com/mozilla/pdf.js/blob/7a034706bade015cf2b969524bc6f0058e220eeb/web/viewer.css#L562 as well, in order for all "buttons" on the generic Findbar to work correctly?

Also, looking at https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#browser_compatibility this does seem to be a *somewhat* recent CSS feature. My initial thought would be that we do *not* attempt to polyfill this for now, in order to simplify things, unless we get lots of bug reports about this.

calixteman:
> Don't you need to update
> 
> https://github.com/mozilla/pdf.js/blob/7a034706bade015cf2b969524bc6f0058e220eeb/web/viewer.css#L562
> as well, in order for all "buttons" on the generic Findbar to work correctly?
> 

Yep I missed there were some "buttons" here, thank you. 

> Also, looking at https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#browser_compatibility this does seem to be a _somewhat_ recent CSS feature. My initial thought would be that we do _not_ attempt to polyfill this for now, in order to simplify things, unless we get lots of bug reports about this.
+1

Snuffleupagus (approved):
Looks good to me, and a lot simpler than I thought it'd be to fix that old bug (thanks to this "new" CSS feature); thank you!

PR #14522, opened 2022-02-01
Remove the `xfa_bug1716838` browser-test since it's a duplicate

Snuffleupagus:
The md5 entry perfectly matches the `xfa_bug1717668_1` test-case, which means that we unnecessarily test the same exact document twice.


calixteman (approved):
Thank you.

PR #14527, opened 2022-02-03
[api-minor] Remove the `normalizeWhitespace` option in the `PDFPageProxy.{getTextContent, streamTextContent}` methods (issue 14519, PR 14428 follow-up)

### LINKED ISSUE #14519: The `normalizeWhitespace = false` option now is being (mostly) ignored
After the changes in PR #14428, the [`normalizeWhitespace` option](https://github.com/mozilla/pdf.js/blob/1087b4dffdf5884ad17669fda54c59dd810667d3/src/display/api.js#L1072-L1073) is now in most (or perhaps even all) cases being ignored and treated as if `normalizeWhitespace = true` is passed to the `PDFPageProxy.{getTextContent, streamTextContent}` methods.[1]

With the changes in PR #14428 we're now handling all whitespaces *consistently* in the `PartialEvaluator.getTextContent` method, which together with other (somewhat) recent improvements have fixed many textLayer issues by handling both coalescing of repeated spaces and proper space insertion between text chunks.
Looking briefly at this code, it seems to me that trying to re-instate "proper" `normalizeWhitespace = false` handling would be quite difficult and would likely hurt readability (assuming it's even at all possible now).
Furthermore, note that in both the viewer and the browser-tests we're only ever using `normalizeWhitespace = true` which means that the other code-path isn't really used all that much.

All-in-all, I'd probably suggest that we simply remove the `normalizeWhitespace` option and make the relevant methods always carry out the normalization since that's what e.g. the viewer uses.

Any thoughts/opinions about this?
/cc @timvandermeij 
/cc @calixteman 

---
[1] One example is the `issue6612.pdf` document in the test-suite, added together with the option, where "regular" spaces are now being returned regardless of the `normalizeWhitespace` value.

Snuffleupagus:
With these changes, we'll now *always* replace all whitespaces with standard spaces (0x20). This behaviour is already, since many years, the default in both the viewer and the browser-tests.

Fixes #14519

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -438,7 +438,6 @@ class Page {
   extractTextContent({
     handler,
     task,
-    normalizeWhitespace,
     includeMarkedContent,
     sink,
     combineTextItems,
@@ -469,7 +468,6 @@ class Page {
         stream: contentStream,
         task,
         resources: this.resources,
-        normalizeWhitespace,
         includeMarkedContent,
         combineTextItems,
         sink,
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -2163,7 +2163,6 @@ class PartialEvaluator {
     task,
     resources,
     stateManager = null,
-    normalizeWhitespace = false,
     combineTextItems = false,
     includeMarkedContent = false,
     sink,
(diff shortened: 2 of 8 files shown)


calixteman (approved):
LGTM. Thank you.

PR #14530, opened 2022-02-03
Avoid the `findResultsCount` span taking up (vertical) space when hidden (PR 13261 follow-up)

Snuffleupagus:
When the viewer becomes narrow, the `PDFFindBar` will (forcibly) wrap its elements to prevent it from extending to the full window width.
Currently, after PR #13261, this now leads to the `findResultsCount` span taking up vertical space *unconditionally* when the findbar is wrapped. To avoid a blank space being shown in this case, before searching has begun, place the `findResultsCount` span in a "message" rather than an "options" container.

### DIFF
--- a/web/pdf_find_bar.js
+++ b/web/pdf_find_bar.js
@@ -163,7 +163,6 @@ class PDFFindBar {
     }
     matchCountMsg.then(msg => {
       this.findResultsCount.textContent = msg;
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for fixing this!

PR #14531, opened 2022-02-04
[Docs] Update README.md

MauryaRitesh:
Fixed a grammatical mistake

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Feel free to stop by our [Matrix room](https://chat.mozilla.org/#/room/#pdfjs:mo
 ### Online demo
 
 Please note that the "Modern browsers" version assumes native support for
-features such as e.g. `async`/`await`, `ReadableStream`, optional chaining,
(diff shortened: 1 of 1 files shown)


Snuffleupagus:
Why do you believe that this is a "grammatical mistake", since you've not included any resources to show that this is wrong?
Furthermore, note that the point of this text is to indicate that the list of features is possibly *not* exhaustive.
(Also, a commit message containing nothing but "Update README.md" isn't really sufficient in general.)

All-in-all, I'm sorry to say that this doesn't seem like a necessary change and hence closing this as WONTFIX.

MauryaRitesh:
Before creating the PR, I searched over the Internet about "such as for example" being in the same sentence. I found that we should use only one of them, here is a link:
https://ell.stackexchange.com/questions/141914/can-we-use-such-as-for-example/155199#:~:text=%22Such%20as%20for%20example%22%20is,%22for%20example%22%20is%20redundant.

PR #14532, opened 2022-02-04
[GENERIC viewer] Remove the `moz`-prefixed FullScreen API usage from the viewer

Snuffleupagus:
The unprefixed FullScreen API has been available since Firefox 64, which was [released on 2018-12-11](https://wiki.mozilla.org/Release_Management/Calendar#Past_branch_dates), and has now been included in no less than *three* ESR releases.
Please also see the following MDN compatibility data:
 - https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenEnabled#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenchange_event#browser_compatibility

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -677,9 +677,7 @@ const PDFViewerApplication = {
     return shadow(
       this,
       "supportsFullscreen",
-      document.fullscreenEnabled ||
-        document.mozFullScreenEnabled ||
-        document.webkitFullscreenEnabled
+      document.fullscreenEnabled || document.webkitFullscreenEnabled
     );
   },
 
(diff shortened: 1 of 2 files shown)


Snuffleupagus:
Looking at the compatibility data, it seems that nowadays *only* Safari needs the `webkit`-prefixed methods.
Based on the number of (general) issues we see related to the Safari-browser in particular, I can't help wonder if we *maybe* want to consider removing support for PresentationMode in browsers that lack the unprefixed FullScreen API or if that's (still) too soon?

timvandermeij (approved):

timvandermeij:
Thank you for cleaning this up! Since this one is indeed obsolete but the Webkit one isn't yet, let's keep that one around for the time being.

Issue #14521, opened 2022-02-01
PDF contents turns redacted (blackout) upon zoom-out to thumbnail size.


dhanbhan:
Attach (recommended) or Link to PDF file here:

Sample Document : 
[Sample_blackout.pdf](https://github.com/mozilla/pdf.js/files/7977687/Sample_blackout.pdf)

Configuration:
- Web browser and its version: Chrome, Firefox, MS Edge (latest versions)
- Operating system and its version: Windows, Mac
- PDF.js version: 2.12.313 (latest)
- Is a browser extension: NA

Steps to reproduce the problem:
1. Open attached `Sample_blackout.pdf` in pdfjs viewer
2. Zoom out the document to lesser than 30% or verify the content in thumbnail view
3. You will notice portion of the document content are turned black.

What is the expected behavior? : Document content should not change its style in thumbs view or while zoom out. 

What went wrong? : document content are turned black and even some portions are redacted.

Link to a viewer (if hosted on a site other than mozilla.github.io/pdf.js or as Firefox/Chrome extension): https://mozilla.github.io/pdf.js/web/viewer.html

**Output in pdfjs viewer on zoom out :** 

<img width="241" alt="blackedOutResponse" src="https://user-images.githubusercontent.com/86353329/151952833-bdf359ab-3f55-4afb-afc5-cc4636a87971.png">

Snuffleupagus:
I wonder if this is *effectively* a duplicate of e.g. issue #13211.

Btow01:
[Sample_blackout.pdf](https://github.com/mozilla/pdf.js/files/8004767/Sample_blackout.pdf)


I have attached the file that we were able to recreate the issue with

PR #14534, opened 2022-02-04
Implement reference test support for rendering at different zoom levels (issue 14505)

### LINKED ISSUE #14505: Add tests to check that the rendering is correct at different zoom levels
We could generate some images at different zoom level and use them as reference.

quaoaris:
Add tests to check that the rendering is correct at different zoom levels issue: #14505

This PR adds a scale factor in test_manifest.json. The scale factor is set by default to 1.

Setting to a value >1 results in a larger image for the browsertest.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -430,6 +430,7 @@ class Driver {
       task.pageNum = task.firstPage || 1;
       task.stats = { times: [] };
       task.enableXfa = task.enableXfa === true;
+      task.scale ||= 1;
 
       // Support *linked* test-cases for the other suites, e.g. unit- and
       // integration-tests, without needing to run them as reference-tests.
@@ -601,7 +602,7 @@ class Driver {
         task.pdfDoc.getPage(task.pageNum).then(
           page => {
             const viewport = page.getViewport({
-              scale: PixelsPerInch.PDF_TO_CSS_UNITS,
+              scale: task.scale * PixelsPerInch.PDF_TO_CSS_UNITS,
             });
             this.canvas.width = viewport.width;
             this.canvas.height = viewport.height;
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -5954,6 +5954,14 @@
        "lastPage": 1,
        "type": "eq"
     },
+    {  "id": "issue6127_scale4",
+       "file": "pdfs/issue6127.pdf",
+       "md5": "65ef9572bddef5fddfaf7172f10a0df2",
+       "rounds": 1,
+       "lastPage": 1,
+       "scale": 4,
+       "type": "eq"
+    },
     {  "id": "annotation-polyline-polygon",
        "file": "pdfs/annotation-polyline-polygon.pdf",
        "md5": "e68611602f58c8ca70cc40575ba3b04e",


test/add_test.js
```diff
@@ -65,6 +65,7 @@ calculateMD5(file, (err, md5) => {
     `       "file": "pdfs/${pdf}.pdf",\n` +
     `       "md5": "${md5}",\n` +
     '       "rounds": 1,\n' +
+    '       "scale": 1,\n' +
```
Snuffleupagus, line 68: Please remove this line, since this parameter should be optional and consequently only needs to be set then *not* equal to one (which isn't going to be that common).

By doing this you'll unnecessarily inflate the size of the `test_manifest.json` file, and I'd view this new parameter similar to the exiting (and optional) `firstPage`/`lastPage` parameters that we only set when actually needed.

test/driver.js
```diff
@@ -431,6 +431,9 @@ class Driver {
       task.stats = { times: [] };
       task.enableXfa = task.enableXfa === true;
 
+      // just in case scale is not set, setting to default=1
```
Snuffleupagus, line 434: In this case, the comment doesn't really add all that much value (compared to simply reading the code); hence I'd suggest removing this one.

(For future reference, this type of comments should be a *complete* sentence. That means beginning with an upper-case letter and ending with a period.)

test/driver.js
```diff
@@ -431,6 +431,9 @@ class Driver {
...
+      task.scale = task.scale || 1;
```
Snuffleupagus, line 435: I believe that `task.scale ||= 1;` ought to work, given the browsers that we run tests in.

test/driver.js
```diff
@@ -601,7 +604,7 @@ class Driver {
         task.pdfDoc.getPage(task.pageNum).then(
           page => {
             const viewport = page.getViewport({
-              scale: PixelsPerInch.PDF_TO_CSS_UNITS,
+              scale: PixelsPerInch.PDF_TO_CSS_UNITS * task.scale,
```
Snuffleupagus, line 607: Nit: To be consistent with e.g. the viewer usage, let's re-order this as `task.scale * PixelsPerInch.PDF_TO_CSS_UNITS` instead.

Snuffleupagus (changes requested):
Besides the inline comments, note that as-is this patch doesn't really "do anything" on its own.

Please actually add one (or more) "new" test-cases that actually use the `scale` parameter (with a non-default value) in the browser-tests; this was also mentioned in e.g. https://github.com/mozilla/pdf.js/issues/14505#issuecomment-1023470637.

Snuffleupagus:
Please revert all re-formatting of the `test_manifest.json` file, since those changes make it very difficult to actually review this and it's also completely unrelated to the changes you're making here.

Snuffleupagus:
@calixteman What sort of additional test-cases (and how many) did you have in mind in https://github.com/mozilla/pdf.js/issues/14505#issuecomment-1023470637?

test/test_manifest.json
```diff
@@ -5954,6 +5954,15 @@
        "lastPage": 1,
        "type": "eq"
     },
+	{
```
timvandermeij, line 5957: This entry is misaligned compared to the other entries.

timvandermeij (commented):
I agree with the question above: while this looks good in general, is the intention just to have a test case to cover rendering with a different scale or did you intend to have more tests here, for example with different zoom levels?

Moreover, the commit message should start with a verb, i.e., something like `Implement reference test support for rendering at different zoom levels` should be more descriptive.

calixteman:
And I planned to add a test for #14521.

test/test_manifest.json
```diff
@@ -5954,6 +5954,15 @@
...
     },
+    {
+	   "id": "issue6127_scale4",
+	   "file": "pdfs/issue6127.pdf",
+	   "md5": "65ef9572bddef5fddfaf7172f10a0df2",
+	   "rounds": 1,
+	   "lastPage": 1,
+	   "scale": 4,
+	   "type": "eq"
+    },
```
Snuffleupagus, line 5965: https://github.com/mozilla/pdf.js/pull/14534#discussion_r800173739 unfortunately still applies here, since you've mixed tabs and spaces.

Please note that we *only* use spaces for indentation everywhere in the code-base. It might help to setup the EditorConfig integration for your code-editor to help with formatting, see e.g. https://editorconfig.org/#pre-installed or https://editorconfig.org/#download

---

> Setting to a value >1 results in a larger image for the browsertest.

*Just a FYI:* It'll also be possible to set the `scale` to a value *below* `1`, since one here simply means 100% (e.g. in the viewer).

Snuffleupagus:
Also, it seems that the *commit message* now contains some unnecessary lines that can/should be removed:
 - `formatting test_manifest.json` looks completely unrelated, and should be removed.
 - `eq test at different zoom levels (issue #14505 )` which was the old title of the commit, but can be removed now.

Snuffleupagus:
> I still have problems using Githubdesktop

Personally I'd *very strongly* suggest using the Git command line directly rather that going through some kind of UI since that should give you much greater control over things.
Interactive rebasing with `git rebase -i`, as mentioned e.g. in https://github.com/mozilla/pdf.js/wiki/Squashing-Commits, is usually the best/easiest way to fixup a bad commit series (and it's useful for other things as well).

timvandermeij:
Let's close this PR for now as it is over a year old now and outdated. Thanks.

PR #14545, opened 2022-02-07
Generate test images at different output scales.

brendandahl:
This will default to generating test images at the device pixel
ratio of the machine the tests are created on unless the
test explicitly defines and output scale using the
`outputScale` setting. This makes the test look visually
like they would on the machine they are running on. It
also allows us to test different output scales.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -59,7 +59,7 @@ function loadStyles(styles) {
   return Promise.all(promises);
 }
 
-function writeSVG(svgElement, ctx) {
+function writeSVG(svgElement, ctx, outputScale) {
   // We need to have UTF-8 encoded XML.
   const svg_xml = unescape(
     encodeURIComponent(new XMLSerializer().serializeToString(svgElement))
@@ -102,7 +102,7 @@ function inlineImages(images) {
   return Promise.all(imagePromises);
 }
 
-async function convertCanvasesToImages(annotationCanvasMap) {
+async function convertCanvasesToImages(annotationCanvasMap, outputScale) {
   const results = new Map();
   const promises = [];
   for (const [key, canvas] of annotationCanvasMap) {
@@ -110,7 +110,10 @@ async function convertCanvasesToImages(annotationCanvasMap) {
       new Promise(resolve => {
         canvas.toBlob(blob => {
           const image = document.createElement("img");
-          image.onload = resolve;
+          image.onload = function () {
+            image.style.width = Math.floor(image.width / outputScale) + "px";
+            resolve();
+          };
           results.set(key, image);
           image.src = URL.createObjectURL(blob);
         });
@@ -198,6 +201,7 @@ class Rasterize {
   static async annotationLayer(
     ctx,
     viewport,
+    outputScale,
     annotations,
     annotationCanvasMap,
     page,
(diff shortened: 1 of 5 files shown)


brendandahl:
As for performance locally,
DPR of 1 takes ~500s
DPR of 2 takes 680s

test/driver.js
```diff
@@ -600,13 +605,38 @@ class Driver {
...
+            // unless the test explicitly specifies an output scale.
+            const outputScale = task.outputScale || window.devicePixelRatio;
+            let viewport = page.getViewport({
               scale: PixelsPerInch.PDF_TO_CSS_UNITS,
             });
-            this.canvas.width = viewport.width;
-            this.canvas.height = viewport.height;
+            // Restrict the test from creating a canvas that is too big.
+            const MAX_CANVAS_PIXEL_DIMENSION = 4096;
+            const largetDimension = Math.max(viewport.width, viewport.height);
```
Snuffleupagus, line 616: Spelling-nit: `largetDimension` -> `largestDimension`.

test/driver.js
```diff
@@ -600,13 +605,38 @@ class Driver {
...
+            if (
+              Math.floor(largetDimension * outputScale) >
+              MAX_CANVAS_PIXEL_DIMENSION
+            ) {
+              const rescale = MAX_CANVAS_PIXEL_DIMENSION / largetDimension;
+              viewport = page.getViewport({
```
Snuffleupagus, line 622: While it doesn't matter in this particular case, please note that you generally want to use `viewport.clone({ ... })` instead to ensure that that only the scale changes (while any other possible parameters stay the same).

test/test.js
```diff
@@ -970,7 +974,7 @@ async function startBrowser(browserName, startUrl = "") {
 }
 
 function startBrowsers(initSessionCallback, makeStartUrl = null) {
-  const browserNames = options.noChrome ? ["firefox"] : ["firefox", "chrome"];
```
Snuffleupagus, line 973: This looks like a left-over from debugging, please revert this.

brendandahl:
Should note I expect a few changes because we will now be drawing the PDF's at a smaller size.

Also, we may want to add a scale option the reftest analyzer. Looking at the results from a lowdpi device on a highdpi device makes the images pretty small.

Snuffleupagus:
> Also, we may want to add a scale option the reftest analyzer.

You mean like issue #14505, and PR #14534, or something else?

timvandermeij:
How does this relate to #14534? Does it supersede that one?

Snuffleupagus:
> How does this relate to #14534? Does it supersede that one?

Unless I'm misunderstanding something I don't *believe* that it does, since this PR is essentially just simulating differing `window.devicePixelRatio` values (i.e. normal vs hiDPI screens).
The other PR is basically simulating different zoom levels (in the viewer), hence the two PRs ought to implement complimentary functionality.

brendandahl:
@Snuffleupagus anything else you want me to address?

test/resources/reftest-analyzer.js
```diff
@@ -228,10 +228,15 @@ window.onload = function () {
         });
         continue;
       }
-      match = line.match(/^ {2}IMAGE[^:]*: (.*)$/);
+      match = line.match(/^ {2}IMAGE[^:]*\((\d+)x(\d+)x(\d)+\): (.*)$/);
```
Snuffleupagus, line 231: Is there a typo the last digit group, i.e. did you mean the following instead?
```suggestion
      match = line.match(/^ {2}IMAGE[^:]*\((\d+)x(\d+)x(\d+)\): (.*)$/);
```

Snuffleupagus (approved):
r=me, with the one comment addressed; thank you!

PR #14547, opened 2022-02-08
Update the file used with the `xfa_bug1720182` test-case

Snuffleupagus:
The file used in this test-case is *identical* to, i.e. the md5 entry perfectly matches, the file used with the `xfa_bug1716380` test-case.
While it's obviously fine to use the same PDF document in different reference-tests, note how we e.g. have both `eq` and `text` tests for one document, we should always avoid adding *duplicate* files in the `test/pdfs/` folder.


timvandermeij (approved):

timvandermeij:
Nice find!

PR #14551, opened 2022-02-10
[api-minor] Stop exposing the `createObjectURL` helper function in the API

Snuffleupagus:
- Update two `display_utils` unit-tests to use native functionality rather than the `createObjectURL` helper function

   Given that most of the code-base is already using native functionality, we can update these unit-tests similarily as well.
    - For the `blob:`-URL test, we simply use `URL.createObjectURL(...)` and `Blob` directly instead.
    - For the `data:`-URL test, we simply use `btoa` to do the Base64 encoding and then build the final URL-string.

- [api-minor] Stop exposing the `createObjectURL` helper function in the API

   With recent changes, specifically PR #14515 *and* the previous patch, the `createObjectURL` helper function is now only used with the SVG back-end.
   All other call-sites, throughout the code-base, are now using `URL.createObjectURL(...)` directly and it no longer seems necessary to keep exposing the helper function in the API.
   Finally, the `createObjectURL` helper function is moved into the `src/display/svg.js` file to avoid unnecessarily duplicating this code on both the main- and worker-threads.

### DIFF
--- a/src/display/svg.js
+++ b/src/display/svg.js
@@ -15,7 +15,6 @@
 /* globals __non_webpack_require__ */
 
 import {
-  createObjectURL,
   FONT_IDENTITY_MATRIX,
   IDENTITY_MATRIX,
   ImageKind,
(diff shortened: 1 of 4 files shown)


Snuffleupagus:
I don't really understand these Windows-only regressions, since testing in e.g. PR #14547 didn't fail.
My best guess, based on a recent mozilla-central pushlog, is that this could be connected to https://bugzilla.mozilla.org/show_bug.cgi?id=1587094 since that sounds related and the commit message in https://phabricator.services.mozilla.com/D137584 explicitly mentions a new default behaviour on Windows.

timvandermeij (approved):

timvandermeij:
Looks good; thank you for improving this!

PR #14558, opened 2022-02-12
Remove the UTF-8 fallback, when `TextDecoder` is missing, from the Content-Disposition parser

Snuffleupagus:
Given that `TextDecoder` is now supported by all modern browsers/environments, please see https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder#browser_compatibility, there's no longer any good reason to keep a UTF-8 fallback in the Content-Disposition parser.


timvandermeij (approved):

timvandermeij:

PR #14559, opened 2022-02-12
Revert "Don't block origin-less blob:-URLs in hosted viewer"

Snuffleupagus:
This reverts commit a6aca3cabe581319d93c862cb0b85f71580cacfd, since no version of Internet Explorer is supported any more.


timvandermeij (approved):

timvandermeij:
Thank you for removing the last few left-over IE bits!

Issue #14509, opened 2022-01-27
Re-enable the browsertests in Google Chrome on the Windows bot


Snuffleupagus:
In PR #14392 we disabled the affected configuration, since the Puppeteer update to [version `> 13.1.0`](https://github.com/puppeteer/puppeteer/releases/tag/v13.1.0) caused the browsertests to run *extremely* slowly and then timeout in Google Chrome on the Windows bot.

It's possible that this is either a bug in Google Chrome, in Puppeteer, or a combination of them both. In any case, we should figure out why this happens and report the bug(s) to the appropriate project(s). (Unless it turns out to be fixable directly in our test-suite, of course.)

timvandermeij:
It looks like Puppeteer 13.5.2 did improve things, but not up to the point that we can enable it again. The Windows bot did not fully timeout after 60 minutes anymore, but while completing it still took ~40 minutes and Chrome did eventually timeout in the tests themselves.

Snuffleupagus:
Let's just WONTFIX this, since I don't think that it's really worth *doubling* the runtime of the Windows-bot just to improve testing in Google Chrome; please see PR #16490.

PR #14561, opened 2022-02-13
Update Puppeteer to version `13.3.1`

Snuffleupagus:
This update, via Puppeteer version `13.2.0`, includes Chromium version 99.0.4844.16 as well.


Snuffleupagus:
I hoped that updating this would've helped with issue #14509, but as the first round of testing above shows it didn't make any difference unfortunately.

PR #14563, opened 2022-02-13
[Search] Some matches were incorrectly shifted because of some '-\n'

### LINKED ISSUE #14562: Search results shows incorrect matches
[NHA-20211231-A-018_before.pdf](https://github.com/mozilla/pdf.js/files/8055863/NHA-20211231-A-018_before.pdf)

Configuration:
- Operating system and its version: Windows 11

1. Search for "a"
2. Some highlighter results are wrong

image

After investigations some results are shifted because of a bug with "-\n".

calixteman:
- it aims to fix #14562;
- 'X-\n' were not correctly positioned;
- when X is a diacritic (e.g. in "sä-\n", which is decomposed into "sa¨-\n") we must handle both things:
  - diacritics on the one hand;
  - "-\n" on the other hand.

For information, I checked in the html that all the spans with "highlighted" class contains a "ä" or "Ä" so the positions in the test are correct.

### DIFF
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -6271,5 +6271,12 @@
          "value": "PDF.js PDF.js PDF.js"
         }
       }
+   },
+   { "id": "issue14562",
+      "file": "pdfs/issue14562.pdf",
+      "md5": "565b0a1e46a32e907837506a10d25277",
+      "rounds": 1,
+      "link": true,
+      "type": "other"
    }
 ]
--- a/test/unit/pdf_find_controller_spec.js
+++ b/test/unit/pdf_find_controller_spec.js
@@ -16,6 +16,7 @@
 import { buildGetDocumentParams } from "./test_utils.js";
 import { EventBus } from "../../web/event_utils.js";
 import { getDocument } from "../../src/display/api.js";
+import { isNodeJS } from "../../src/shared/is_node.js";
 import { PDFFindController } from "../../web/pdf_find_controller.js";
 import { SimpleLinkService } from "../../web/pdf_link_service.js";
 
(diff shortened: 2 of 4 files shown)


test/test_manifest.json
```diff
@@ -6264,5 +6264,12 @@
          "value": "PDF.js PDF.js PDF.js"
         }
       }
+   },
+   { "id": "issue14562",
+      "file": "pdfs/issue14562.pdf",
+      "md5": "565b0a1e46a32e907837506a10d25277",
+      "rounds": 1,
+      "link": true,
+      "type": "text"
```
Snuffleupagus, line 6273: Given that none of the changes in this patch show up in a `text` test, let's save time/resources by making this `"type": "other"` instead.

web/pdf_find_controller.js
```diff
@@ -158,18 +158,19 @@ function normalize(text) {
         return replacement;
       }
 
-      if (p2) {
+      if (p3) {
```
Snuffleupagus, line 161: It feels slightly confusing that the `p{n}` cases are now appearing out of order, but I'm assuming that it's actually necessary now!?

test/unit/pdf_find_controller_spec.js
```diff
@@ -554,4 +555,48 @@ describe("pdf_find_controller", function () {
       pageMatchesLength: [[23]],
     });
   });
+
+  it("performs a search in a text containing some -\\n", async function () {
```
Snuffleupagus, line 559: Maybe `"performs a search in a text containing diacritics before -\\n"` to more clearly describe the problem you're fixing here (based on the PR description).

Snuffleupagus (approved):
Seems to work fine when testing manually, and the changes look reasonable (although I cannot claim to understand all details of the normalization code); thank you!

PR #14564, opened 2022-02-13
[api-minor] Don't add in the text content the chars which are out-of-page (bug 1755201)

calixteman:
- it aims to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1755201;
- if the glyph position is not within the view then skip it.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -2167,6 +2167,7 @@ class PartialEvaluator {
     includeMarkedContent = false,
     sink,
     seenStyles = new Set(),
+    viewBox,
   }) {
     // Ensure that `resources`/`stateManager` is correctly initialized,
     // even if the provided parameter is e.g. `null`.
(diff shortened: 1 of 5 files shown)


src/core/evaluator.js
```diff
@@ -2167,6 +2167,7 @@ class PartialEvaluator {
     includeMarkedContent = false,
     sink,
     seenStyles = new Set(),
+    cropBox,
```
Snuffleupagus, line 2170: Are you sure that this is the *correct* thing to use in general, since the thing that determines what we actually render is the `view` getter on the `Page`-instances (if nothing else it feels surprising/inconsistent)?

test/unit/api_spec.js
```diff
@@ -2219,6 +2219,22 @@ Caron Broadcasting, Inc., an Ohio corporation (“Lessee”).`)
...
+        pending("Linked test-cases are not supported in Node.js.");
+      }
+
+      const loadingTask = getDocument(buildGetDocumentParams("bug1755201.pdf"));
+      const pdfDoc = await loadingTask.promise;
+      const pdfPage = await pdfDoc.getPage(6);
+      const { items } = await pdfPage.getTextContent();
+      const text = mergeText(items);
+
+      expect([...text.matchAll(/win aisle/g)].length).toEqual(0);
```
Snuffleupagus, line 2233: Could this not be simplified a bit, perhaps something like this:
```javascript
expect(/win aisle/.test(text)).toEqual(false);
```

Snuffleupagus (changes requested):

test/unit/api_spec.js
Snuffleupagus: Another reason for changing this is that `matchAll` is new enough functionality that it'd require adding a polyfill...

Snuffleupagus (approved):
This seems reasonable to me, thank you!

---

This was tagged `[api-minor]` since it's an observable change of a behaviour that's existed since "forever", and there's always a risk that some users depend on the old behaviour.
While it'd obviously be possible to add an option to toggle this on/off, I'm suggesting that we do *not* do so speculatively since all options have maintenance/testing overhead.

PR #14566, opened 2022-02-14
Miscellaneous small viewer clean-up

Snuffleupagus:
- Remove the `typeof navigator`-checks in the `web/app_options.js` file

   Given that the `Navigator` interface has been available since "forever", please see https://developer.mozilla.org/en-US/docs/Web/API/Navigator#browser_compatibility, it's somewhat difficult to see why these checks are actually necessary since the viewer is only intended for usage in browsers.

   Looking at the history of the code, this functionality was originally placed in the general `src/shared/compatibility.js` file which could thus run in e.g. worker-threads and Node.js environments (where the `Navigator` interface isn't available).

 - Change `PasswordPrompt.close` to an `async` method

   This is consistent with the `open` method, and it actually *ever so slightly* reduces the size of the file.

### DIFF
--- a/web/app_options.js
+++ b/web/app_options.js
@@ -15,12 +15,16 @@
 
 const compatibilityParams = Object.create(null);
 if (typeof PDFJSDev === "undefined" || PDFJSDev.test("GENERIC")) {
-  const userAgent =
-    (typeof navigator !== "undefined" && navigator.userAgent) || "";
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #14569, opened 2022-02-16
Fix canvas state getting out of sync from smasks. (bug 1755507)

### LINKED ISSUE #11328: Images embedded in PDF disappearing
I am using this test link [https://mozilla.github.io/pdf.js/web/viewer.html](https://mozilla.github.io/pdf.js/web/viewer.html)

The test pdf [pdfjs_lost_image_test.pdf](https://github.com/mozilla/pdf.js/files/3840182/pdfjs_lost_image_test.pdf)

pdfjs_2

The image is lost when previewing with pdfjs.

pdfjs_1

A temporary solution is to convert the file itself, convert this pdf to jpg first, and then convert this jpg to pdf.

Is it possible that the font is Chinese traditional? and how can I solve it?

Thanks a million!

### LINKED ISSUE #11937: incorrect position - FireFox browser

Configuration:
- Web browser and its version: FireFox 76.0.1(64bit)
- PDF.js version: 2.3.200
- Is a browser extension: no

1. open pdf.js view.html
2. upload pdf file.
3. it looks different between Chrome and FireFox

chrome

view on firefox
firefox

Attachted pdf
[JRCHIST_wb_s1_1.1_t.pdf](https://github.com/mozilla/pdf.js/files/4685855/JRCHIST_wb_s1_1.1_t.pdf)

it is possible to reproduce on this link

Thanks everyone

### LINKED ISSUE #14297: Pdf.js doesn't fully render the attached PDF page
Attach (recommended) or Link to PDF file here: [RenderIssue.pdf](https://github.com/mozilla/pdf.js/files/7583657/RenderIssue.pdf)

Configuration:
- Web browser and its version: Chrome Version 96.0.4664.45 (Official Build) (64-bit)
- Operating system and its version: Windows 10 Version 10.0.19042

- PDF.js version: v2.12.201 (If this looks incorrect, please take it from the URL)
- Is a browser extension: No, using https://mozilla.github.io/pdf.js/web/viewer.html

1. Visit https://mozilla.github.io/pdf.js/web/viewer.html
2. Drop the attached PDF on to the page - [RenderIssue.pdf](https://github.com/mozilla/pdf.js/files/7583657/RenderIssue.pdf)

What is the expected behavior? It should appear as in Adobe Reader - 
image

What went wrong? Some text is blank - 
image

brendandahl:
Soft masks can be enabled/disabled at anytime and at different
points in the save/restore stack. This can lead to
the amount of save/restores becoming unbalanced across the
two canvases. Instead of save/restoring on the temporary canvas
change it so we only track state on the main (suspended canvas).

I was also getting an out balance stack from patterns, so I've also
fixed that and added a warning that will at least show up on chrome.
It would be nice to add this so Firefox at some point too.

Fixes https://github.com/mozilla/pdf.js/issues/11328, https://github.com/mozilla/pdf.js/issues/14297 and bug 1755507


test/test_manifest.json
```diff
@@ -3685,6 +3685,12 @@
       "rounds": 1,
       "type": "eq"
     },
+    {  "id": "bug1755507_page1_reduced",
```
Snuffleupagus, line 3688: Nit: Maybe just `bug1755507` as the id?

Snuffleupagus (changes requested):
Unfortunately this breaks the `issue6165` test-case.

brendandahl:
Interesting, I don't have that one in my smask tests cases. I think this is also similar to #14297 where the state is getting out of sync between canvases.

src/display/canvas.js
```diff
@@ -191,6 +191,10 @@ function mirrorContextOperations(ctx, destCtx) {
 }
 
 function addContextCurrentTransform(ctx) {
+  if (ctx._transformStack) {
+    // Reset the transform stack.
+    ctx._transformStack = [];
+  }
```
Snuffleupagus, line 197: With these changes, do you still need line 212 below since that one essentially looks like a duplicate now?

src/display/canvas.js
```diff
@@ -1589,21 +1599,37 @@ class CanvasGraphics {
   }
 
   save() {
-    this.ctx.save();
+    if (this.inSMaskMode()) {
+      // SMask mode may be turned on/off causing us to loose graphics state.
```
Snuffleupagus, line 1603: Grammar-nit: I believe that you probably want to change `to loose` -> `to lose`.

src/display/canvas.js
```diff
@@ -1589,21 +1599,37 @@ class CanvasGraphics {
...
+      // Copy the temporary canvas state to the main(suspended) canvas to keep
+      // it in sync.
+      copyCtxState(this.ctx, this.suspendedCtx);
+      // Don't bother calling save on the temporary canvas since state is not
+      // save there.
```
Snuffleupagus, line 1608: Nit: `save` -> `saved`.

src/display/canvas.js
```diff
@@ -1503,8 +1510,12 @@ class CanvasGraphics {
     }
   }
 
-  checkSMaskState() {
-    const inSMaskMode = !!this.suspendedCtx;
+  inSMaskMode() {
```
Snuffleupagus, line 1513: Let's make this one a getter instead.

src/display/canvas.js
```diff
@@ -1503,8 +1510,12 @@ class CanvasGraphics {
...
+    return !!this.suspendedCtx;
+  }
+
+  checkSMaskState(fromRestore = false) {
```
Snuffleupagus, line 1517: I might be wrong, but it appears that the `fromRestore` parameter is completely unused and could thus be removed.

Snuffleupagus (approved):
r=me, with the comments addressed; thank you!

Also, I found the added code comments really helpful when reading the code.

src/display/canvas.js
Snuffleupagus: Given that we have a bunch abstract base-classes and interface definition, where any parameters aren't being used, the currently ESLint is configured to not check function/method definitions.

I can look into fixing that (with manually white-listing the appropriate cases), since I agree that this is less than ideal.

src/display/canvas.js
brendandahl: I don't think so, because then we'd end up attached a _transformStack on canvases in Firefox which don't use this code.

Snuffleupagus: In that case, shouldn't this new block of code simply be removed?
Note that the transform stack is already being forcibly reset further down, in non-Firefox browsers, which then ought to render this newly added code unnecessary: https://github.com/mozilla/pdf.js/blob/d9a3a243533de06a61b94765178c3e4d826c8a43/src/display/canvas.js#L208

PR #14572, opened 2022-02-16
Add a missing string-check in the `_collectJS` helper function

Snuffleupagus:
Unfortunately I don't have a test-case that breaks without this change, however the `stringToPDFString` helper function will fail if anything other than a string is passed to it.
The changes in this patch thus make this code more-or-less identical to that found in the `Catalog.{_collectJavaScript, parseDestDictionary}` methods.

### DIFF
--- a/src/core/core_utils.js
+++ b/src/core/core_utils.js
@@ -22,7 +22,8 @@ import {
   stringToPDFString,
   warn,
 } from "../shared/util.js";
-import { Dict, isName, isRef, isStream, RefSet } from "./primitives.js";
+import { Dict, isName, isRef, RefSet } from "./primitives.js";
+import { BaseStream } from "./base_stream.js";
 
 function getLookupTableFactory(initializer) {
   let lookup;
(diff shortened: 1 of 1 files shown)


src/core/core_utils.js
```diff
@@ -329,15 +330,15 @@ function _collectJS(entry, xref, list, parents) {
...
       let code;
-      if (isStream(js)) {
+      if (js instanceof BaseStream) {
         code = js.getString();
-      } else {
+      } else if (typeof js === "string") {
         code = js;
       }
-      code = stringToPDFString(code);
+      code = code && stringToPDFString(code);
```
calixteman, line 341: nano-nit: use `&&=` to shorten the line

calixteman (approved):
LGTM, thank you.

src/core/core_utils.js
Snuffleupagus: I'd love to do that, but unfortunately we're *still* blocked on [bug 1247687](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687); please see PR #12563.

PR #14575, opened 2022-02-17
Remove the `isStream` helper function

Snuffleupagus:
At this point all the various Stream-classes extends an abstract base-class, hence this helper function is no longer necessary and only adds unnecessary indirection in the code.


timvandermeij (approved):

timvandermeij:
Looks good to me; thank you for simplifying this!

PR #14577, opened 2022-02-18
Remove the `isRef` helper function

Snuffleupagus:
This helper function is not really needed, since it's just a wrapper around a simple `instanceof` check, and it only adds unnecessary indirection in the code.


timvandermeij (approved):
r=me with a rebase since the first commit was already part of another PR; thanks!

PR #14578, opened 2022-02-18
Remove the `backingStorePixelRatio`-part of the `getOutputScale` helper function

Snuffleupagus:
The `CanvasRenderingContext2D.backingStorePixelRatio` property was never standardized, and only Safari set (its prefixed version of) it to anything other than `1`.
Note that e.g. MDN doesn't contain any information about this property, and one of the few sources of information (at this point) is the following post: https://stackoverflow.com/questions/24332639/why-context2d-backingstorepixelratio-deprecated

Hence we can simplify the `getOutputScale` helper function, by removing some dead code, and now it no longer requires any parameters when called.


timvandermeij (approved):

timvandermeij:
Nice cleanup!

PR #14581, opened 2022-02-19
Use the (viewer) `parseQueryString` helper function in the reference tests

Snuffleupagus:
Rather than re-implementing this functionality in the `Driver` class, we can simply re-use the existing `parseQueryString` helper function instead.


timvandermeij (approved):

timvandermeij:
Thanks!

Issue #14308, opened 2021-11-24
PDF unscrollable, 'stuck' at certain pages


lrehmann:
Link to PDF file here: https://manuals.plus/m/bd73fb00d867df453ee8b9b753f761cdf4c3d919f5a5bc956c63d7d87ad45e59.pdf

Configuration:
- Web browser and its version: Firefox 94.0.2 (64-bit), (Also Tried Chrome and Chrome Canary)
- Operating system and its version: OSX Big Sur 11.4 (20F71)
- PDF.js version: whatever ships with 94.0.2 (64-bit) Firefox
- Is a browser extension: No

Steps to reproduce the problem:
1.  Load the PDF, embedded navigation script will force you to page 146
2. Accept the alert dialog
3. PDF will be un-scrollable, stuck at 'Cautions & Warnings' pages  (pg 146,147,148)

What is the expected behavior? (add screenshot)
User should be able to freely scroll all PDF pages

What went wrong? (add screenshot)
View-window gets 'stuck' on a particular page

I suspect there may be some javascript imbedded in this PDF forcing a scroll to a certain page in order to view the 'Cautions & Warnings' page. Though this could be a poorly coded PDF file, the PDF reader shouldn't allow embedded code to break the user interface.

Snuffleupagus:
Oof, that's *really* not good. (I get the same behaviour in Adobe Reader, but still not at all a good UX.)
Looking at the document, all pages contain the following bad JS; @calixteman any ideas for a good/simple work-around?

![pagesAA](https://user-images.githubusercontent.com/2692120/143316208-2ba189c0-b41d-4324-aa0d-bb726930a94f.png)

calixteman:
A possible workaround is to disable javascript execution.
In Firefox, `about:config`, change the option `pdfjs.enableScripting` to false.
The main problem with this file is that there are no action bound in the OK button on the last page.
In Acrobat, once the OK button is pressed, we've access to all the pdf.

After digging, the button itself has no action but there is a Link annotation (id 20) on top of the button.
We only added JS support for fields (our goal was mainly to help with form validation) but not for annotations in general and unfortunately a Link is not a field (when a button is):
https://github.com/mozilla/pdf.js/blob/master/src/core/annotation.js#L2497-L2514

It shouldn't be that hard to implement but it'd require some time...
I'm attaching the pdf here to be sure to have it when I'll have some time to fix this issue.
[...]

PR #14583, opened 2022-02-19
[Annotations] Show buttons even if they've no actions

calixteman:
- it's a regression from PR #14247:
 - before the PR, the button was rendered on the canvas whatever its status was;
 - after the PR, the button image has been moved in an other canvas so when the button is not renderable
   (because it has no actions) then the image is not added the HTML element.
- the buttons in the pdf in bug 1737260 or in the pdf in #14308 were not visible
- make the button always renderable but don't add the link element if it's useless.

### DIFF
--- a/test/test_manifest.json
+++ b/test/test_manifest.json
@@ -6528,5 +6528,13 @@
          "value": "1"
         }
       }
+   },
+   { "id": "bug1737260",
+      "file": "pdfs/bug1737260.pdf",
+      "md5": "8bd4f810d30972764b07ae141a4afbc4",
+      "rounds": 1,
+      "link": true,
+      "type": "eq",
+      "annotations": true
    }
 ]
--- a/src/display/annotation_layer.js
+++ b/src/display/annotation_layer.js
@@ -563,36 +563,29 @@ class AnnotationElement {
 
 class LinkAnnotationElement extends AnnotationElement {
   constructor(parameters, options = null) {
-    const isRenderable = !!(
-      parameters.data.url ||
-      parameters.data.dest ||
-      parameters.data.action ||
-      parameters.data.isTooltipOnly ||
-      parameters.data.resetForm ||
-      (parameters.data.actions &&
-        (parameters.data.actions.Action ||
-          parameters.data.actions["Mouse Up"] ||
-          parameters.data.actions["Mouse Down"]))
-    );
     super(parameters, {
-      isRenderable,
+      isRenderable: true,
       ignoreBorder: !!options?.ignoreBorder,
       createQuadrilaterals: true,
     });
+    this.isTooltipOnly = parameters.data.isTooltipOnly;
   }
 
   render() {
     const { data, linkService } = this;
     const link = document.createElement("a");
+    let isBound = false;
 
     if (data.url) {
       linkService.addLinkAttributes(link, data.url, data.newWindow);
+      isBound = true;
     } else if (data.action) {
       this._bindNamedAction(link, data.action);
+      isBound = true;
     } else if (data.dest) {
       this._bindLink(link, data.dest);
+      isBound = true;
     } else {
-      let hasClickAction = false;
       if (
         data.actions &&
         (data.actions.Action ||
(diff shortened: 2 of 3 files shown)


test/test_manifest.json
```diff
@@ -6290,5 +6290,12 @@
       "rounds": 1,
       "link": true,
       "type": "other"
+   },
+   { "id": "bug1737260",
+      "file": "pdfs/bug1737260.pdf",
+      "md5": "8bd4f810d30972764b07ae141a4afbc4",
+      "rounds": 1,
+      "link": true,
+      "type": "eq"
```
timvandermeij, line 6299: Shouldn't this also have `annotations: true` (see https://github.com/mozilla/pdf.js/blob/571fe0ead8cbdd26c00cc509ba40fe38c4f030bc/test/test_manifest.json#L6174 for example)?

timvandermeij:
Is the change in `firefox-file_url_link-page1` correct? I seem to recall that links to local files were not rendered, but I'm not entirely sure of this.

calixteman:
`isRenderable` has been added in https://github.com/mozilla/pdf.js/pull/8865 in order to fix https://github.com/mozilla/pdf.js/issues/3897.
My feeling is that we should show the annotation whatever the action is but of course we don't have to bind an action in some case.
@timvandermeij, what's your opinion here ?

timvandermeij:
I'm not really sure what's best here. On the one hand we could say that we render the annotation but don't make it clickable since that's closer to what's actually in the PDF file, but on the other hand I can also see why https://github.com/mozilla/pdf.js/pull/8865 made them hidden because it's very likely to confuse the user if you see a link annotation but nothing happens when clicking it. Since we already know `file://` URLs will not work well, it might be nicer to not show them at all to avoid confusion?

Maybe we should compare it to Acrobat and see what that does? I imagine they also have problems with `file://` URLs pointing to non-existent files. Do they not render the annotation then, or just show a warning on click or so?

calixteman:
If you open the pdf in https://github.com/mozilla/pdf.js/issues/14862 in using for example Foxit, you can see a yellow page with a warning:
image

This yellow thing is a Widget button with no action.
When the viewer is Acrobat then it's hidden and visible else.
I think we must display everything we have in the pdf and of course when there are no action we must try to avoid to give the impress that it's a button (e.g. don't bind anything).
It isn't display in the viewer but if you to print the page you can see it.

About `file:///`:
 - in Acrobat:
   image
   and when clicking on `Allow`, it leads to nothing.
 - in Chrome: it tries to open the URL `file://path/to/local/file.pdf` in the browser

Snuffleupagus:
I've not really looked in any detail at the code in this PR, but as far as I understand it these changes will not make the *links* themselves visible/clickable (e.g. for `file://`-URLs) but only make the *border* visible (when one exist)?

While displaying a border around a non-clickable annotation might look a bit strange, and potentially slightly confusing to users, that in itself shouldn't change any actual *behaviour* here.
Furthermore, again assuming that I understand things correctly, we basically "have to" make these changes in order to properly support annotations whose visibility can be changed *dynamically* from scripting.

calixteman:
With `file_url_link.pdf`, the only visible change for the user is to have this green border but nothing.
From my pov, the only drawback would be to display a button with an action we don't support (if any), hence the user could try to click on it without any success. If we don't plan to support the action we should probably warn in the console (or with an alert) else it's a bug.
But if by design a button is displayed whatever the actions are, we should/must render it.

Snuffleupagus:
@timvandermeij Ping; how do think that we should proceed here?

timvandermeij:
Looking at this again, if it's indeed fixing the original issues with the only drawback that we see the green border I think I'm fine with that. Not showing the buttons indeed seems worse than showing them but not having them actually clickable.

Snuffleupagus:
@calixteman Do you have time to rebase this, so that we can run tests again and then land this?

timvandermeij (approved):

timvandermeij:
Thanks!

/botio makeref

PR #14588, opened 2022-02-20
Remove old prefixed CSS rules used with `text` tests

Snuffleupagus:
According to MDN, both the `transform-origin` and `box-sizing` CSS rules are supported in their *unprefixed* versions in modern browsers:

 - https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing#browser_compatibility

### DIFF
--- a/test/text_layer_test.css
+++ b/test/text_layer_test.css
@@ -27,15 +27,9 @@
 .textLayer br {
   position: absolute;
   white-space: pre;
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Nice clean-up!

PR #14594, opened 2022-02-21
Prefer `instanceof ...` rather than calling `isCmd`/`isDict`/`isName` with *one* argument

Snuffleupagus:


calixteman (approved):
I way prefer like this, thanks.

PR #14595, opened 2022-02-22
Update the support information for `structuredClone` (PR 14392 follow-up)

Snuffleupagus:
When the `structuredClone` polyfill was added, the support information in Safari was unclear. Given that an actual version *number* is now available, see below, it seems like a good idea to update the comment accordingly.

https://developer.mozilla.org/en-US/docs/Web/API/structuredClone#browser_compatibility

### DIFF
--- a/src/shared/compatibility.js
+++ b/src/shared/compatibility.js
@@ -78,7 +78,7 @@ if (
     ).ReadableStream;
   })();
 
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #14596, opened 2022-02-22
Remove the `isNum` helper function

Snuffleupagus:
The call-sites are replaced by direct `typeof`-checks instead, which removes unnecessary function calls. Note that in the `src/`-folder we already had more `typeof`-cases than `isNum`-calls.

These changes were *mostly* done using regular expression search-and-replace, with two exceptions:
 - In `Font._charToGlyph` we no longer unconditionally update the `width`, since that seems completely unnecessary.
 - In `PDFDocument.documentInfo`, when parsing custom entries, we now do the `typeof`-check once.

### DIFF
--- a/src/core/catalog.js
+++ b/src/core/catalog.js
@@ -26,7 +26,6 @@ import {
   FormatError,
   info,
   isBool,
-  isNum,
   isString,
   objectSize,
   PermissionFlag,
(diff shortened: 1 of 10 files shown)


calixteman (approved):
LGTM, thank you for simplifying this.

PR #14597, opened 2022-02-22
Ensure that `Dict.set` only accepts string `key`s

Snuffleupagus:
Trying to use a non-string `key` in a `Dict` is not intended, and would basically be an implementation error. Hence we can add a non-PRODUCTION check to enforce this, complementing the existing `value` check added in PR #11672.


timvandermeij (approved):

timvandermeij:
Good idea; let's do this!

PR #14599, opened 2022-02-23
Ensure that `Cmd`/`Name` is only initialized with string arguments

Snuffleupagus:
Trying to use a non-string argument in either a `Cmd` or a `Name` is not intended, and would basically be an implementation error. Hence we can add a non-PRODUCTION check to enforce this, similar to the existing one used e.g. in the `Dict.set` method.


timvandermeij:
Thanks!

PR #14600, opened 2022-02-24
[api-minor] Add validation for the  `PDFDocumentProxy.getPageIndex` method

Snuffleupagus:
Currently we'll happily attempt to send any argument passed to this method over to the worker-thread, without doing any sort of validation.
That could obviously be quite bad, since there's first of all no protection against sending unclonable data. Secondly, it's also possible to pass data that will cause the `Ref.get` call in the worker-thread to fail immediately.

In order to address all of these issues, we'll now properly validate the argument passed to `PDFDocumentProxy.getPageIndex` and when necessary reject already on the main-thread instead.

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -454,8 +454,8 @@ class WorkerMessageHandler {
       });
     });
 
-    handler.on("GetPageIndex", function wphSetupGetPageIndex({ ref }) {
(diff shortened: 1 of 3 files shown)


timvandermeij:
Good idea; thanks!

PR #14602, opened 2022-02-24
 Re-factor the `PDFDocument.documentInfo` method, and remove the `isString` helper function

Snuffleupagus:
- Re-factor the `PDFDocument.documentInfo` method

   This removes the `DocumentInfoValidators` structure, and thus (slightly) simplifies the code overall. With these changes we only have to iterate through, and validate, the actually available Dictionary entries.

 - Remove the `isString` helper function

   The call-sites are replaced by direct `typeof`-checks instead, which removes unnecessary function calls. Note that in the `src/`-folder we already had more `typeof`-cases than `isString`-calls.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -24,7 +24,6 @@ import {
   escapeString,
   getModificationDate,
   isAscii,
-  isString,
   OPS,
   RenderingIntentFlag,
   shadow,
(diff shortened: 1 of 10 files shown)


timvandermeij (approved):
r=me with a rebase; thank you for improving this!

PR #14605, opened 2022-02-25
Simplify the `decodeFontData`/`encodeFontData` font-test helper functions

Snuffleupagus:
We can (and in my opinion should) use the standard `atob`/`btoa` functions, rather than manually re-implementing this functionality for the font-tests.


calixteman (approved):
With green tests, r=me.
Thank you.

PR #14607, opened 2022-02-25
Simplify the `wrapReason` helper function

Snuffleupagus:
All call-sites that use `wrapReason` should be passing a (possibly cloned) `Error` to the helper function, hence we shouldn't need to have a fallback code-path for any other data.
Note that for the `cancel`/`error` methods on Streams, since PR #11115 we've been asserting that the argument is in fact an `Error` as intended.
When calling `wrapReason` from *rejected* Promises, we should also be guaranteed that an `Error` is provided thanks to the ESLint rules `no-throw-literal` and `prefer-promise-reject-errors`.

### DIFF
--- a/src/shared/message_handler.js
+++ b/src/shared/message_handler.js
@@ -21,7 +21,7 @@ import {
   PasswordException,
   UnexpectedResponseException,
   UnknownErrorException,
-  warn,
+  unreachable,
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #14608, opened 2022-02-25
Don't close window from test driver.

brendandahl:
Sometimes I get a "Unable to find target with id XXX closeTarget..." error
when running tests which happens when test.js tries to close all the
open pages. I haven't been able to fully verify since this is intermittent,
but I think this is coming from us closing the window in driver.js and also
trying to close it in test.js.


brendandahl:
Seems to be happening on the bots too. See this [recent makeref](http://54.193.163.58:8877/a4ab951443760d3/output.txt) that failed.

PR #14609, opened 2022-02-25
Improvements to the reftest analyzer.

brendandahl:
- Scroll the selected reference into view (makes it easier to tell which pdf you're looking at)
- Show the keyboard shortcuts (easier for new people)
- Keep the test/ref controls visible (if you scroll you can now tell if you're looking at a test or ref)

### DIFF
--- a/test/resources/reftest-analyzer.html
+++ b/test/resources/reftest-analyzer.html
@@ -92,7 +92,7 @@ Original author: L. David Baron <dbaron@dbaron.org>
     <div id="itemlist">
       <table id="itemtable"></table>
     </div>
-    <div id="images">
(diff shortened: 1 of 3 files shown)


test/resources/reftest-analyzer.html
```diff
@@ -103,73 +103,76 @@ <h1>Reftest analyzer</h1>
       <label>
         <input type="checkbox" id="differences"> Circle differences
       </label>
+      <span id="shortcuts">Shortcuts: n=next p=previous t=toggle</span>
```
Snuffleupagus, line 106: Note that you also have <kbd>d</kbd>, which toggles the "Circle differences" feature.

PR #14610, opened 2022-02-26
[JPEG 2000] Add support for resetContextProbabilities (bug 1731483)

Snuffleupagus:
Given the size of the patch, I figured that submitting it here for testing wouldn't hurt; obviously I'm keeping the original commit author info *intact*!


timvandermeij (approved):
r=me; since this is a patch originating from Bugzilla I'm not sure if we can merge it directly or if we need to notify upstream of that? If not, feel free to merge it.

calixteman (approved):

PR #14611, opened 2022-02-26
Remove the JSDocs "External: Promise"-page, since `Promise`s are now a standard feature

Snuffleupagus:
The "External: Promise"-page in the JSDocs pre-dates the introduction of `Promise`s, as a generally available standard JS feature, by a number of years. Hence it now longer seems necessary, as far as I can tell, to include this "special" page in the documentation.

Also, while unrelated to the rest of the patch, updates the `test/`-folder description in the documentation.


timvandermeij (approved):

timvandermeij:
Thanks!

PR #14612, opened 2022-02-27
Add the `caniuse-lite` package

Snuffleupagus:
Since the Autoprefixer plugin indirectly depends on this, it seems like a good idea to add this as a direct dependency in the PDF.js project to hopefully avoid having to manually update `caniuse-lite` in the future; see https://github.com/browserslist/browserslist#browsers-data-updating

Also, slightly tweaks the Autoprefixer config for GENERIC-builds of the PDF.js library; note that this change doesn't affect the contents of the *built* `web/viewer.js` file.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -81,7 +81,7 @@ const AUTOPREFIXER_CONFIG = {
     "Chrome >= 73",
     "Firefox ESR",
     "Safari >= 12.1",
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Yes, this looks like a good idea to me. Thank you!

PR #14615, opened 2022-02-27
Remove the `-webkit-overflow-scrolling: touch;` CSS rules (PR 4516 follow-up)

Snuffleupagus:
This was added in PR #4516 specifically for Safari on iOS devices, but according to MDN it should no longer be necessary now; see https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-overflow-scrolling#browser_compatibility

According to the MDN compatibility data, this CSS feature:
 - Was never implemented anywhere *except* for Safari on iOS.
 - Was never standardized and thus never existed in an *unprefixed* version.
 - Has now been removed, starting with Safari version 13.

Given that [the FAQ](https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support) already lists Safari as "Mostly" supported, and that the default viewer is written primarily for Mozilla Firefox, it ought to be fine to remove these CSS rules now.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -284,7 +284,6 @@ html[dir="rtl"] #outerContainer.sidebarOpen #sidebarContainer {
   top: 32px;
   bottom: 0;
   overflow: auto;
-  -webkit-overflow-scrolling: touch;
   position: absolute;
   width: 100%;
   background-color: rgba(0, 0, 0, 0.1);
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for cleaning up those prefixed rules!

PR #14620, opened 2022-03-01
Bug 1757527 - Slightly reduce (repeated) message passing overhead when localizing the PDF Viewer

Snuffleupagus:


Snuffleupagus:
Given that https://bugzilla.mozilla.org/show_bug.cgi?id=1757527 has now landed, after being reviewed by both Brendan and Calixte, I'm going to "break protocol" by landing myself to ensure that the GitHub and Bugzilla code-bases agree here.

PR #14624, opened 2022-03-02
Simplify the `fallback`-logic in the default viewer

Snuffleupagus:
After [bug 1705327](https://bugzilla.mozilla.org/show_bug.cgi?id=1705327) the `ChromeActions.fallback`-method is now just a stub[1], since the fallback bar was removed.
Hence there's no good reason, as far as I can tell, to keep this code in the viewer itself when it's completely unused.

---
[1] See https://searchfox.org/mozilla-central/rev/292d17c13daa61016fd082e2337297091d53a015/toolkit/components/pdfjs/content/PdfStreamConverter.jsm#575-581

### DIFF
--- a/web/app.js
+++ b/web/app.js
@@ -149,8 +149,6 @@ class DefaultExternalServices {
 
   static initPassiveLoading(callbacks) {}
 
-  static async fallback(data) {}
-
   static reportTelemetry(data) {}
 
   static createDownloadManager(options) {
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Given the upstream patch I also don't really see a reason to keep this around; thanks!

PR #14625, opened 2022-03-02
Cleanup workflows

tiziodcaio:


Snuffleupagus:
> Removed setup-node cuz is already included in the docker image
> [...]
> It is preinstalled, too!

Can you please provide links to official resources that confirm that these changes are desirable/correct?

Also, please use *proper* English is commit messages (and code-comments); hence replace "cuz" with the correct word.

Finally, please see https://github.com/mozilla/pdf.js/wiki/Squashing-Commits

.github/workflows/ci.yml
```diff
@@ -7,15 +7,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
-
-      - name: Use Node.js 16 LTS
```
Snuffleupagus, line 12: Also, I'm not sure that we want to loose the ability to *manually* specify the exact Node.js that we use for testing!?

tiziodcaio: Yes I agree, maybe keeping setup-node can be the soulution is important. I can find for you the source of the information, but if you are intersted to force a specific nodejs version I cannot change your mind.

timvandermeij:
Initially I was a bit confused about why the `gulp-cli` removal was working, but it turns out our Gulp 4 upgrade fixed that "for free" given that `gulp-cli` is a dependency of `gulp` since version 4.

I do agree that it would be good to be explicit about the Node.js version we're testing against so it doesn't go unnoticed when the upstream image is updated for example. In general we want all our dependencies to be explicit, and Node.js happens to be the only one that we can't enforce easily through e.g., `package.json` like the rest of our dependencies.

Therefore, this PR should be fine with the Node.js part put back and the commits squashed into one with a meaningful commit message.

tiziodcaio:
I will remake soon the PR, let me close for a moment please

PR #14628, opened 2022-03-03
When `stopAtErrors` is set, throw rather than warn when exceeding `maxImageSize` (issue 14626)

### LINKED ISSUE #14626: When image is bigger than maxImageSize, it should throw an error instead of warning
** I dont think the markdown format is relevant thus I removed it**

I was having issue with this
https://github.com/mozilla/pdf.js/issues/6741

I'm thinking to at least show a message to client telling that the image is too big thus it is removed.
but I don't think there is a way to catch a warning thus an error is needed.

https://github.com/mozilla/pdf.js/blob/46840c56502e06888a70854c991bce9d199ca83c/src/core/evaluator.js#L577-L581

To think about it when one of the images are bigger than the stated `maxImageSize`
, it should have alternative to stop all the process on the pdf else it's a kind of meaningless

#### Question
why actually would you limit `maxImageSize`

Snuffleupagus:
The situation described in issue #14626 seems like a fairly special case, and it thus seem reasonable that we simply follow the same pattern as elsewhere in the `PartialEvaluator` when the `stopAtErrors` API-option is being used.

### DIFF
--- a/src/core/evaluator.js
+++ b/src/core/evaluator.js
@@ -576,8 +576,13 @@ class PartialEvaluator {
     }
     const maxImageSize = this.options.maxImageSize;
     if (maxImageSize !== -1 && w * h > maxImageSize) {
-      warn("Image exceeded maximum allowed size and was removed.");
-      return;
+      const msg = "Image exceeded maximum allowed size and was removed.";
+
+      if (this.options.ignoreErrors) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #14629, opened 2022-03-03
Compute the loca table `endOffset`, of the "first" glyph, correctly (issue 14618)

### LINKED ISSUE #14618: Text is not rendered correctly (namely, zeroes are missing)
[unpacked-anonymized.pdf](https://github.com/mozilla/pdf.js/files/8154616/unpacked-anonymized.pdf)

Configuration:
Doesn't work for built-in pdf-preview in, for example, Firefox Developer Edition 98.0b9 (64 bits), but works in, for example, recent Chrome (Version 98.0.4758.109 (Official Build) (x86_64)) or Safari (Version 15.2 (15612.3.6.1.8, 15612)).

macOS Catalina Version 10.15.7

- PDF.js version:
Tested as well with pdf-js "2.12.313" (through "react-pdf") - zeroes would not render in any browser for the pdf in question.

- Is a browser extension:
?

1. Not sure how the original pdf was created, so please see attached the truncated/anonymized example (length and xref are wrong due to me editing the file content using vim after unpacking the file using `qpdf  --qdf`, so `qpdf --check` now gives warnings; _please advise on other ways to anonymise the file if current example is not good enough for some reason_).

Screenshot 2022-02-28 at 17 29 13

Screenshot 2022-02-28 at 17 27 23

?

P.S.:
If this issue seems doable for a new contributor would be willing to help out, preferably, with some initial code pointers to look at.

Snuffleupagus:
When there are *multiple* empty glyphs at the start of the data, ensure that the "first" glyph gets a correct `endOffset` to avoid skipping it during parsing in the `sanitizeGlyph` function.


skivol:
These changes worked fine for 2 original pdf files that had this issue as well 🥇👍

brendandahl:
Looks good, thanks!

/botio makeref

PR #14636, opened 2022-03-06
Replace XMLHttpRequest usage with the Fetch API in `Driver._quit`

Snuffleupagus:
This is another step in what'll hopefully become a series of patches to implement PR #14287 in smaller steps.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -845,9 +845,9 @@ class Driver {
     this.end.textContent = "Tests finished. Close this window!";
 
     // Send the quit request
-    const r = new XMLHttpRequest();
-    r.open("POST", `/tellMeToQuit?browser=${escape(this.browser)}`, false);
-    r.send(null);
+    fetch(`/tellMeToQuit?browser=${escape(this.browser)}`, {
+      method: "POST",
+    });
   }
 
   _info(message) {


test/driver.js
```diff
@@ -845,9 +845,9 @@ class Driver {
     this.end.textContent = "Tests finished. Close this window!";
 
     // Send the quit request
-    const r = new XMLHttpRequest();
-    r.open("POST", `/tellMeToQuit?browser=${escape(this.browser)}`, false);
-    r.send(null);
+    fetch(`/tellMeToQuit?browser=${escape(this.browser)}`, {
```
timvandermeij, line 848: Shouldn't we be `await`ing the returned promise here, and catch any failures? If this fails for some reason, I imagine it could leave the browser hanging.

Snuffleupagus: > Shouldn't we be `await`ing the returned promise here, and catch any failures?

Given that the existing code doesn't bother handling errors, I figured that we didn't need to in the new one either!?

> If this fails for some reason, I imagine it could leave the browser hanging.

Yes, but that's already the case here (and the bots will force-abort after one hour). Note that PR #14608 very recently stopped trying to close the browser from here, in an attempt to reduce intermittent errors.

---

I suppose that we could try to re-introduce closing here *only* if a non-200 status code was observed, but I just don't know how much it'd matter?

timvandermeij (approved):

timvandermeij:
Thanks!

test/driver.js
timvandermeij: I suppose it won't matter too much then. Given that we already didn't handle errors here, I'm fine with keeping it as-is.

PR #14640, opened 2022-03-07
Update TypeScript to version `4.6.2` and work-around stricter type checks

Snuffleupagus:
I'm guessing that we're now running into the class-related improvements mentioned in https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/#target-es2022
To unblock this update, and any future ones, this patch simply tweaks the JSDocs to get `gulp typestest` to run without errors.


timvandermeij (approved):

timvandermeij:
Thank you for the update!

PR #14642, opened 2022-03-07
Replace XMLHttpRequest usage with the Fetch API in `Driver._send`

Snuffleupagus:
This is another part in a series of patches that try to re-implement PR #14287 in smaller steps.

Besides converting `Driver._send` to use the Fetch API, this also changes the method to return a `Promise` to get rid of the callback function.
Please note that I *purposely* try to maintain the existing behaviour of re-sending the data on failure/unexpected response, including how/where the old callback function was invoked.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -19,6 +19,7 @@
 const {
   AnnotationLayer,
   AnnotationMode,
+  createPromiseCapability,
   getDocument,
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks much better this way; thanks!

PR #14648, opened 2022-03-08
Simplify the `PDFDocument` constructor

Snuffleupagus:
Originally the code in the `src/`-folder was shared between the main/worker-threads, and back then it probably made sense that the `PDFDocument` constructor accepted different arguments.
However, for many years we've not been passing anything *except* Streams to `PDFDocument` and we should thus be able to slightly simplify that code. Note that for e.g. unit-tests of this code, using either a `NullStream` or a `StringStream` works just fine.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -18,7 +18,6 @@ import {
   FormatError,
   info,
   InvalidPDFException,
-  isArrayBuffer,
   isArrayEqual,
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Makes sense; thanks!

PR #14651, opened 2022-03-09
Replace XMLHttpRequest usage with the Fetch API in `inlineImages` (in `test/driver.js`)

Snuffleupagus:
This is the final part in a series of patches that try to re-implement PR #14287 in smaller steps.

Besides converting `inlineImages` to use the Fetch API, this patch also combines the `inlineImages` and `resolveImages` functions since they are always used together.

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -78,29 +78,52 @@ function writeSVG(svgElement, ctx) {
   });
 }
 
-function inlineImages(images) {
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thank you for doing this; much better without the indirection!

PR #14654, opened 2022-03-10
Replace XMLHttpRequest usage with the Fetch API in `send` (in `test/unit/testreporter.js`)

Snuffleupagus:
Besides converting the `send` function to use the Fetch API, this patch also changes the method to return a `Promise` to get rid of the callback function. (Although, currently there's no call-site passing in a callback function.)


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #14656, opened 2022-03-10
Move the `isSameOrigin` helper function

Snuffleupagus:
This function is currently placed in the `src/shared/util.js` file, which means that the code is duplicated in both of the *built* `pdf.js` and `pdf.worker.js` files. Furthermore, it only has a single call-site which is also specific to the `GENERIC`-build of the PDF.js library.

Hence this helper function is instead moved into the `src/display/api.js` file, in such a way that it's conditionally defined but still can be unit-tested.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -26,7 +26,6 @@ import {
   info,
   InvalidPDFException,
   isArrayBuffer,
-  isSameOrigin,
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks good!

PR #14658, opened 2022-03-10
Validate the `cMapUrl`/`standardFontDataUrl` parameters in `getDocument`

Snuffleupagus:
These changes make sense for two reasons:
 - Given that the parameters are potentially passed to the worker-thread, depending on the `useWorkerFetch` parameter, we need to prevent errors if the user provides values that aren't clonable.
 - By ensuring that the default values are indeed `null`, we'll trigger main-thread fetching (of CMaps and Standard fonts) as intended in the `PartialEvaluator` and thus potentially provide better Error messages.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -339,6 +339,12 @@ function getDocument(src) {
   if (!Number.isInteger(params.maxImageSize)) {
     params.maxImageSize = -1;
   }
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Makes sense; thanks!

PR #14661, opened 2022-03-11
Update Checkout to version 3

tiziodcaio:
> Inspired from #14625

Checkout was upgraded so it can use node.js 16 by default


timvandermeij (approved):

timvandermeij:
Thank you for updating this!

tiziodcaio:
^^ it wasn't hard

PR #14665, opened 2022-03-12
Convert the `BasePreferences` class to use private fields

Snuffleupagus:
- Convert the `BasePreferences` class to use private fields

   Given that none of these fields were ever intended to be accessed directly from the *outside*, since that will lead to inconsistent/broken state, we can use modern ECMAScript features to ensure that they are indeed private.

 - Revert Preferences to their previous values, when writing to storage failed

   This patch fixes an old inconsistency, when using `BasePreferences.{reset, set}`, where the internal Preference values would be kept even if writing them to storage failed.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #14667, opened 2022-03-13
Enable the "gets fieldObjects" unit-test in Node.js (PR 14409 follow-up)

Snuffleupagus:
Apparently this unit-test works in Node.js now, hence it's *possible* that the reason it didn't work previously is that there were bugs in our old `structuredClone` polyfill.


timvandermeij (approved):

timvandermeij:
Interesting find; thanks!

PR #14669, opened 2022-03-13
Replace `dir`-dependent `padding-left`/`padding-right`/`text-align` with logical properties

Snuffleupagus:
*Please note:* This is small first step in what will, time permitting, become a series of patches to simplify/modernize the viewer CSS.

Rather than having to *manually* specify ltr/rtl-specific padding-values in the CSS, we can use logical padding instead (and similar for text-align).

Note that a number of logical CSS properties are new enough that we'll need to use PostCSS plugins, however the ones in this patch are natively supported in all browsers that we currently support:
 - https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline-start#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline-end#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/CSS/text-align#browser_compatibility

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -904,21 +904,15 @@ html[dir="rtl"] .dropdownToolbarButton::after {
   color: var(--main-color);
   margin: 0;
   padding: 1px 0 2px;
+  padding-inline-start: 4px;
   border: none;
   background-color: var(--dropdown-btn-bg-color);
-}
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
I didn't know about those CSS properties yet, but this is very useful to avoid duplication. Thank you for doing this!

PR #14671, opened 2022-03-13
Replace `dir`-dependent `float` with logical properties

Snuffleupagus:
- Stop polyfilling the CSS `calc` functionality

   At this point in time, all browsers that we support have native support for CSS variables; please see https://developer.mozilla.org/en-US/docs/Web/CSS/calc()#browser_compatibility and https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-support

   We can also remove the hack introduced back in PR https://github.com/mozilla/pdf.js/pull/11567#issuecomment-582166160, which was only necessary to work-around an IE 11 specific bug.

 - Replace `dir`-dependent `float` with logical properties

   *Please note:* This is another step in what will, time permitting, become a series of patches to simplify/modernize the viewer CSS.

   Rather than having to manually specify ltr/rtl-specific float-values in the CSS, we can use logical `inline-start`/`inline-end` instead (and similar for some related left/right occurrences).

   Given that most of these logical CSS properties are fairly new, and that cross-browser support is thus somewhat limited (see below), we rely on PostCSS plugins in order to support this in the GENERIC viewer.

    - https://developer.mozilla.org/en-US/docs/Web/CSS/float#browser_compatibility
    - https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-end#browser_compatibility

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -17,7 +17,6 @@
 "use strict";
 
 const autoprefixer = require("autoprefixer");
-const postcssCalc = require("postcss-calc");
 const postcssDirPseudoClass = require("postcss-dir-pseudo-class");
 const postcssLogical = require("postcss-logical");
 const fs = require("fs");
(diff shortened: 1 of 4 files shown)


timvandermeij (approved):

timvandermeij:
Looks great; thanks!

PR #14675, opened 2022-03-14
Build the `web/viewer.css` file used in the development viewer (i.e. `gulp server`)

Snuffleupagus:
- Build the `web/viewer.css` file used in the development viewer (i.e. `gulp server`)

   To allow using modern CSS features that currently only Mozilla Firefox supports[1], while still enabling development/testing in recent Google Chrome versions, we'll have to start building the `web/viewer.css` file with `gulp server` as well.

   In my testing, building the development CSS (and copying the images) takes *less than* `200 ms` on average which is hopefully an acceptable overhead for this sort of feature.

   ---
   [1] In particular `float`, with `inline-start`/`inline-end` values.

 - Remove the `-webkit-mask-image`/`-webkit-mask-image` CSS rules

   Given that we're now *building* the `web/viewer.css` file used in the development viewer, i.e. with `gulp server`, we no longer need to hard-code these `-webkit`-prefixed rules and can instead let Autoprefixer handle that for us.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1869,6 +1869,24 @@ gulp.task(
   )
 );
 
+gulp.task("dev-css", function createDevCSS() {
+  console.log();
+  console.log("### Building development CSS");
+
+  const defines = builder.merge(DEFINES, { GENERIC: true, TESTING: true });
+  const cssDir = BUILD_DIR + "dev-css/";
+
+  return merge([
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Let's do this; thanks!

PR #8023, opened 2017-02-04
Moves preprocessor stuff to the gulpfile.

yurydelendik:
Moves all preprocessor logic and code from the make.js file and removes intermediate build/pdf.js, build/pdf.worker.js and build/viewer.js files (they were "last build configuration"-processed)

P.S. #7062

PR #14676, opened 2022-03-14
Remove the custom `grab`/`grabbing` cursor image files

Snuffleupagus:
- [gulpfile.js] Remove the `cleanup` parameter in `preprocessCSS` helper function

   Every single call-site has always passed in `true` for this parameter, ever since the function was first added back in PR #8023. Hence the parameter appears to be completely unnecessary, which is why it's removed and the function is updated to *unconditionally* strip out any license headers (in the middle of the file).

 - Remove the custom `grab`/`grabbing` cursor image files

   According to the MDN compatibility data, see https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#browser_compatibility, all browsers that we now support should have these cursors available natively.


timvandermeij (approved):

timvandermeij:
Thanks for the cleanup!

PR #14678, opened 2022-03-15
Convert the `PDFDocumentProperties` class to use private methods

Snuffleupagus:
Given that none of these methods were ever intended to be accessed directly from the outside, we can use modern ECMAScript features to ensure that they are indeed private.

This patch also makes `fieldData` private, to remove the old hack used to prevent it from being modified from the outside.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #14682, opened 2022-03-16
[gulpfile.js] Use the regular `defines` in the `preprocessCSS` function

Snuffleupagus:
Rather than *manually* specifying a "mode", we can simply use the regular `defines` directly instead. To improve consistency, in the `external/builder/builder.js` file, a couple of parameters are also re-named.


timvandermeij (approved):

timvandermeij:
Looks good!

PR #14683, opened 2022-03-16
[src/display/api.js] Simplify the `sendTest` function, used with Worker initialization (PR 14291 follow-up)

Snuffleupagus:
Given that we now only use Workers when `postMessage` transfers are supported, there's really no point in trying to send a "test" message *without* transfers present.
Hence, if `postMessage` transfers are not supported by the browser, we'll now fallback to "fake" Workers immediately instead. The comment about Opera is also removed, since it was originally added back in PR #983 and mentions Opera `11.60` [which was released in 2011](https://en.wikipedia.org/wiki/History_of_the_Opera_web_browser#Version_11).

### DIFF
--- a/src/core/worker.js
+++ b/src/core/worker.js
@@ -75,9 +75,8 @@ class WorkerMessageHandler {
       }
       testMessageProcessed = true;
 
-      // Ensure that `TypedArray`s can be sent to the worker,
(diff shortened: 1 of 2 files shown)


timvandermeij (approved):

timvandermeij:
Thank you!

PR #14688, opened 2022-03-18
Add general iteration support in the `RefSet` and `RefSetCache` classes

Snuffleupagus:
This patch removes the existing `forEach` methods, in favor of making the classes properly iterable instead. Given that the classes are using a `Set` respectively a `Map` internally, implementing this is very easy/efficient and allows us to simplify some existing code.


timvandermeij (approved):

timvandermeij:
Good idea; thanks!

PR #14689, opened 2022-03-19
Slightly simplify the `PDFFindController._extractText` method, and convert all "private" methods into proper ones

Snuffleupagus:
- Slightly simplify the `PDFFindController._extractText` method

   Currently we're resolving the Promises in the `_extractTextPromises` Array with the page-index, despite that not really being necessary since the Promises in the Array are explicitly inserted in the correct order.
   Furthermore, we can replace the standard `for`-loop with a `for...of`-loop which results in ever so slightly more compact code.

 - Convert all "private" methods in `PDFFindController` into proper ones

   Given that none of these methods are/were ever intended to be called manually, we can now enforce this with modern class-features.

### DIFF
--- a/web/pdf_find_controller.js
+++ b/web/pdf_find_controller.js
@@ -260,7 +260,7 @@ class PDFFindController {
     this._linkService = linkService;
     this._eventBus = eventBus;
 
-    this._reset();
+    this.#reset();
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Looks good to me!

PR #14690, opened 2022-03-19
Replace `dir`-dependent `left`/`right` with logical properties

Snuffleupagus:
*Please note:* This is another step in a series of patches to simplify/modernize the viewer CSS.

Rather than having to manually specify ltr/rtl-specific left/right-values in the CSS, we can use logical `inset-inline`/`inset-inline-start`/`inset-inline-end` instead.

Given that most of these logical CSS properties are fairly new, and that cross-browser support is thus somewhat limited (see below), we rely on the previously added PostCSS plugins in order to support this in the GENERIC viewer.

 - https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-start#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-end#browser_compatibility

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -235,21 +235,19 @@ select {
   position: absolute;
   top: 32px;
   bottom: 0;
+  inset-inline-start: calc(-1 * var(--sidebar-width));
   width: var(--sidebar-width);
   visibility: hidden;
   z-index: 100;
   border-top: 1px solid rgba(51, 51, 51, 1);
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
This is really nice; thanks!

PR #14691, opened 2022-03-19
Replace `dir`-dependent `border`-rules with logical properties

Snuffleupagus:
*Please note:* This is the final step in a series of patches to simplify/modernize the viewer CSS, since the remaining `html[dir="rtl"]`-cases cannot be converted.

Rather than having to manually specify ltr/rtl-specific border-values in the CSS, we can use logical `border-inline-end`/`border-start-end-radius`/`border-end-end-radius` instead.

Given that most of these logical CSS properties are fairly new, and that cross-browser support is thus somewhat limited (see below), we rely on the previously added PostCSS plugins in order to support this in the GENERIC viewer.

 - https://developer.mozilla.org/en-US/docs/Web/CSS/border-inline-end#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/CSS/border-start-end-radius#browser_compatibility
 - https://developer.mozilla.org/en-US/docs/Web/CSS/border-end-end-radius#browser_compatibility

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -240,15 +240,10 @@ select {
   visibility: hidden;
   z-index: 100;
   border-top: 1px solid rgba(51, 51, 51, 1);
+  border-inline-end: var(--doorhanger-border-color-whcm);
   transition-property: inset-inline-start;
   transition-duration: var(--sidebar-transition-duration);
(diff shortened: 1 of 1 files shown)


timvandermeij (approved):

timvandermeij:
Thanks!

PR #14692, opened 2022-03-19
[JS] - Parse a date in using the given format first and then try the default date parser

### LINKED ISSUE #14672: Date fields with custom formats are filled with a value but the input display something else
Hi pdfjs team,

Please find attach a really simple pdf file to reproduce the issue.
[Date.pdf](https://github.com/mozilla/pdf.js/files/8242310/Date.pdf)
This pdf contains 4 date fields, One with a default format (dd/mm/yyyy) and the 3 others with custom formats:

    -  dd
    -  mm
    -  yyyy

The expected behavior is that whenever I fill dd field with 16 for instance, then when i focus out that field the value displayed is 16 but I get 01.

This happens for dd field and mm but it's working fine for dd/mm/yyyy and yyyy fields.

The initial values of the pdf are properly displayed, this happens only if I focus one of those fields.

Thank you for your time

Best regards.

calixteman:
- it aims to fix #14672.

### DIFF
--- a/src/scripting_api/util.js
+++ b/src/scripting_api/util.js
@@ -373,6 +373,10 @@ class Util extends PDFObject {
   }
 
   scand(cFormat, cDate) {
+    if (typeof cDate !== "string") {
+      return new Date(cDate);
+    }
+
     if (cDate === "") {
       return new Date();
     }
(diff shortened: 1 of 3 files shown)


src/scripting_api/util.js
```diff
@@ -536,15 +540,15 @@ class Util extends PDFObject {
...
     if (!matches || matches.length !== actions.length + 1) {
       return null;
     }
 
     const data = {
-      year: 0,
+      year: 2000,
       month: 0,
-      day: 0,
+      day: 1,
```
Snuffleupagus, line 551: Are these default values changed to match the Adobe JS specification, or simply to match the implementation in Adobe Reader?

Snuffleupagus (approved):
r=me, with one question; thank you!

src/scripting_api/util.js
calixteman: I opened the pdf from the issue in Acrobat and the UI for the date selection is set to January 2000 for the day field and to the 1st of December 2000 for the month field.

PR #14693, opened 2022-03-19
Remove the remaining `dir`-dependent CSS rules

Snuffleupagus:
After the recent round of patches, I figured that we'd gone as far as possible in replacing `dir`-dependent CSS rules for the viewer.
However, it occurred that me that we could actually use a bit of CSS-trickery to get rid of the remaining ones. More specifically, this was done by defining a CSS variable whose value depends on the document direction and then using that variable together with `calc()` in the affected rules.

*Please note:* I suppose that this could perhaps be seen as a bit too "magical", hence I understand if this patch is ultimately rejected, however this is probably the only simple way to get rid of the remaining `dir`-dependent CSS rules.

### DIFF
--- a/web/viewer.css
+++ b/web/viewer.css
@@ -16,6 +16,7 @@
 @import url(pdf_viewer.css);
 
 :root {
+  --dir-factor: 1;
   --sidebar-width: 200px;
   --sidebar-transition-duration: 200ms;
   --sidebar-transition-timing-function: ease;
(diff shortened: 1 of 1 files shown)


calixteman:
Magic exists to be used ;).

timvandermeij (approved):

timvandermeij:
Thanks!

Snuffleupagus:
@timvandermeij Thanks for reviewing/merging all of the recent CSS patches!

*For reference:* After the recent round of CSS clean-up, for the `gulp mozcentral` target the size of the *built* `viewer.css` file has been reduced by `> 8 %` thanks to using logical properties.

PR #14694, opened 2022-03-20
Removed const warning using comment

EraKin575:
Removed const warning using "jshint esversion:6"


Snuffleupagus:
We're using ESLint in this project, since over five years, hence this patch unfortunately does *not* make sense here!

PR #14699, opened 2022-03-21
Slightly improve validation of (some) parameters in `getDocument`

Snuffleupagus:
There's a couple of `getDocument` parameters that should be numbers, but which are currently not *fully* validated to prevent issues elsewhere in the code-base.
Also, improves validation of the `ownerDocument` parameter since we currently accept more-or-less anything here.

### DIFF
--- a/src/display/api.js
+++ b/src/display/api.js
@@ -316,7 +316,6 @@ function getDocument(src) {
     params[key] = value;
   }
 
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Good idea!

PR #14703, opened 2022-03-21
[text selection] Add the whitespaces present in the pdf in the text chunk

### LINKED ISSUE #14627: Text search is not able to identify text, but working with other pdf viewers
Attach (recommended)
[Issue_Document.pdf](https://github.com/mozilla/pdf.js/files/8176877/Issue_Document.pdf)
 or Link to PDF file here:

Configuration:
- Web browser and its version: Chrome and Version 98.0.4758.102 
- PDF.js version: Using  "ngx-extended-pdf-viewer": "^12.0.0-alpha.2" which a wrapper around Mozilla's PDF viewer
- Is a browser extension: No

1. Open the attached pdf document in 'https://mozilla.github.io/pdf.js/web/viewer.html?'
2. Search for text 'message persists' and 0 results are found
3. Open the attached document with Chrome 
4. Search for text 'message persists' and '1' result is found
5. Open the attached document in Adobe Acrobat Reader
6. Search for text 'message persists' and '1' result is found

Search result for 'message persists' should be 1

> 

**CHROME:**

image

**ADOBE ACROBAT READER**

image

Pdf.js not able to find the text 'message persists'.

**Pdf.js**

image

calixteman:
- it aims to fix issue #14627;
- the basic idea of the recent text refactoring was to only consider the rendered visible whitespaces.
  But sometimes, the heuristics aren't correct and although some whitespaces are in the text stream
  they weren't in the text chunks because they were too small. Hence we added some exceptions, for example,
  we always add a whitespace when it is between two non-whitespace chars but only when in the same Tj.
  So basically, this patch removes the constraint to have the chars in the same Tj
  (in using a circular buffer to save the two last chars) but don't add a space when the visible space is really
  too small (hence `NOT_A_SPACE_FACTOR`).


Snuffleupagus:
There seem to be a regression in `issue6605-page1`, for all platforms/browsers, since most of the spaces appear to be missing now.

src/core/evaluator.js
```diff
@@ -2187,17 +2187,50 @@ class PartialEvaluator {
...
       fontName: null,
       hasEOL: false,
     };
 
+    // Use a cyclic buffer (length === 2) to save the last chars in the
+    // text stream.
+    // It's useful to know when we need to add a whitespace in the
+    // text chunk.
+    const twoLastChars = [" ", " "];
+    let twoLastCharsPos = 0;
```
Snuffleupagus, line 2201: Am I correct in assuming that the reason for using a *cyclic* cache is for overall performance reasons? If so, that's probably worth mentioning in the comment.

The reason for me asking is that this makes the code slightly more complex (when reading it), compared to simply using a combination of e.g. `Array.prototype.push()` and `Array.prototype.shift()` to insert data into the array.

calixteman: Yep, it's a way faster in using this way to do (no need to resize the same array).
I guess there is not so much difference from a user pov, but it's probably better to at least try to minimize the power consumption.
Anyway I updated the comment.

Snuffleupagus: Thanks; I figured that we should fry to avoid anyone coming along in the future and thinking that this code could/should just be simplified.

src/core/evaluator.js
```diff
@@ -2491,7 +2538,17 @@ class PartialEvaluator {
...
+        if (advanceY <= textOrientation * textContentItem.notASpace) {
+          // The real spacing between 2 consecutive chars is thin enough to be
+          // considered a non-space.
+          resetLastChars();
+        }
+
         if (advanceY <= textOrientation * textContentItem.trackingSpaceMin) {
+          if (advanceY <= textContentItem.notASpace) {
+            resetLastChars();
+          }
```
Snuffleupagus, line 2551: There isn't a similar branch below for the `advanceX`-case down below, I'm assuming that's on purpose?

Snuffleupagus (approved):
r=me, with the final comment addressed.

src/core/evaluator.js
calixteman: It's a leftover: it has been replaced by the resetLastChars stuff just above.

PR #14704, opened 2022-03-22
Convert the `OverlayManager` class to use private fields/methods

Snuffleupagus:
- Convert the `OverlayManager` class to use private fields/methods

 - Change `OverlayManager.open` to always error if the overlay is already active

   The old code would allow an overlay to force close *itself*, before immediately re-opening itself, which actually isn't very helpful in practice since that won't re-run any overlay-specific initialization code.
   Given how the overlays are being used this really shouldn't have caused any issues, but it's a bug that we should fix nonetheless.


timvandermeij (approved):

timvandermeij:
Thank you for doing this!

Issue #14649, opened 2022-03-08
[Secondary Toolbar]: After applying high contrast mode the selected controls text is not visible.


seahawksq:
After applying high contrast mode, the selected controls text is not visible.
![image](https://user-images.githubusercontent.com/25859586/157341845-ddc58ce0-edc3-4ea4-8c54-07b69349519c.png)
![image](https://user-images.githubusercontent.com/25859586/157341923-c2741dc7-a0ed-4e55-aa42-eff75d377fd1.png)

Configuration:
- Web browser and its version: Chrome Version 97.0.4692.71 (Official Build) (64-bit) /Edge Version 97.0.1072.69 (Official build) (64-bit)
- Operating system and its version: Operating system and its version: 21H2(OS build 22000.376)
- PDF.js version: 2.13.107
- Is a browser extension: No

Steps to reproduce the problem:
1. Turning on high contrast mode in operating system.
2. Navigate to https://mozilla.github.io/pdf.js/web/viewer.html
3. Open the secondary Toolbar.

What is the expected behavior? (add screenshot)
The selected controls text should be visible.

What went wrong? (add screenshot)
The selected controls text is not visible.

Snuffleupagus:
Given https://github.com/mozilla/pdf.js/issues/14649#issuecomment-1169020381, it really sounds like this should be reported in the Chromium bug-tracker instead.

PR #14710, opened 2022-03-23
Convert the existing overlays to use `<dialog>` elements (issue 14698) 

### LINKED ISSUE #14698: Use <dialog> element for the document properties modal
It has been recently released in Firefox:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
and it's available in Safari too:
https://webkit.org/blog/12445/new-webkit-features-in-safari-15-4/

Snuffleupagus:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "canvas": "^2.9.1",
         "core-js": "^3.21.1",
         "cross-env": "^7.0.3",
+        "dialog-polyfill": "^0.5.6",
         "dommatrix": "^0.0.24",
         "es-module-shims": "1.4.7",
         "eslint": "^8.11.0",
@@ -4349,6 +4350,12 @@
         "enabled": "1.0.x",
         "kuler": "1.0.x"
       }
+    },
+    "node_modules/dialog-polyfill": {
+      "version": "0.5.6",
+      "resolved": "https://registry.npmjs.org/dialog-polyfill/-/dialog-polyfill-0.5.6.tgz",
+      "integrity": "sha512-ZbVDJI9uvxPAKze6z146rmfUZjBqNEwcnFTVamQzXH+svluiV7swmVIGr7miwADgfgt1G2JQIytypM9fbyhX4w==",
+      "dev": true
     },
     "node_modules/dir-glob": {
       "version": "3.0.1",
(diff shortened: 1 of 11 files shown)


calixteman:
I tested in windows 11 in using high contrast themes (I selected "Dusk") and there is potentially a regression.

In Firefox nightly, in using http://54.241.84.105:8877/4c5d97f84b91cdd/web/viewer.html: 
image

In Firefox nightly, with the builtin viewer:
image

In edge, in using http://54.241.84.105:8877/4c5d97f84b91cdd/web/viewer.html: 
image

@Snuffleupagus, do you think it's an issue on our side or something wrong with dialog itself or something else ?
@MReschenberg, what is the correct way to display this dialog ?

Snuffleupagus:
> do you think it's an issue on our side or something wrong with dialog itself or something else ?

Given that I didn't really change the CSS all that much, and that the regular `prefers-color-scheme: dark` appear to work, I'd be inclined to assume that it's actually a *platform issue* with the `<dialog>` element here.

MReschenberg:
We should be using the version that respects user-set colors (so the non-black and white one, if you're running Dusk theme). If this happens on all dialogs, it is probably a platform bug -- do we have something on file for that in bugzilla?

Snuffleupagus:
> I tested in windows 11 in using high contrast themes (I selected "Dusk") and there is potentially a regression.

While I'm currently on an older Windows version, I can reproduce similar problems with *dark* high contrast themes in basic `<dialog>` examples as well; see e.g. https://css-tricks.com/some-hands-on-with-the-html-dialog-element/

timvandermeij:
Given that this is an upstream issue, should we just continue with this or do we need to await a bugfix upstream?

marco-c:
I think we should await for a fix upstream. The first step would be to file a bug on Bugzilla. @Snuffleupagus could you do that?

calixteman:
I filed a bug upstream yesterday:
https://bugzilla.mozilla.org/show_bug.cgi?id=1761611

Snuffleupagus:
https://bugzilla.mozilla.org/show_bug.cgi?id=1761611 has already been fixed in Firefox Nightly, and the HTML standard was updated accordingly as well, hence that should no longer block this PR as far as I can tell :-)

/botio-linux preview

calixteman:
In using `Night sky` theme, I noticed that the buttons have a yellow border line (like the text) but in the dialog it's purple.
And in opening the viewer in edge and in mouse-hovering the button, I get:
image
The printing UI in firefox (and in settings UI), switches bg to yellow and fg to black on mouse-hover, I wonder if we should do the same for the dialog in pdf.js, just to be consistent.

Snuffleupagus:
> And in opening the viewer in edge and in mouse-hovering the button, I get:

That sounds a lot like issue #14649, which honestly feels unrelated to this PR.

> [...] I wonder if we should do the same for the dialog in pdf.js, just to be consistent.

In this PR, I've purposely tried to *only* convert the existing overlays while not changing e.g. the look of things.
Any future improvements, or re-designs, would seem more appropriate for follow-up work and we'd probably need an actual design-spec to follow here.

calixteman (approved):
Looks good to me.
Thank you for doing that.

PR #14715, opened 2022-03-24
Avoid the `textLayer` becoming visible in high contrast mode (issue 13230)

### LINKED ISSUE #13230: Grey background color/text highlight in high contrast mode
Link to PDF file:

Configuration:
- Web browser and its version: Firefox Developer Edition 88.0b9
- Operating system and its version: Windows 10 19042.870
- PDF.js version: Latest version
- Is a browser extension: No

1. Switch to high contrast mode.
2. Open a PDF file.

What is the expected behavior
- Text and background colors are as defined in the file:
Screenshot 2021-04-13 112321

What went wrong?
- Text has a a highlight/grey background in high contrast mode:
Screenshot 2021-04-13 112354

Possible solution:
- Add the following code after line 32 in pdf_viewer.css

```CSS
.pdfViewer > .page > .textLayer {
    forced-color-adjust: none;
}
```

``forced-color-adjust``, however, is not supported in Firefox.

Snuffleupagus:
Unfortunately this CSS property is not yet available in Firefox, see https://developer.mozilla.org/en-US/docs/Web/CSS/forced-color-adjust#browser_compatibility which is tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1591210, however this patch seems to work when testing in Google Chrome.

Given that we really cannot do any more on the PDF.js-side of things, until this CSS feature is actually implemented in Firefox, I figured that submitting this patch cannot hurt in order to get rid of an open issue.

### DIFF
--- a/web/text_layer_builder.css
+++ b/web/text_layer_builder.css
@@ -24,6 +24,7 @@
   opacity: 0.2;
   line-height: 1;
   text-size-adjust: none;
+  forced-color-adjust: none;
 }
 
 .textLayer span,


timvandermeij (approved):

timvandermeij:
Nice find!

PR #14718, opened 2022-03-25
Add test for paintInlineImageXObjectGroup optimization.

brendandahl:
There are currently no PDFs in the test suite that exercise this
code path.


brendandahl:
Never mind issue4926.pdf uses this code path.

PR #14720, opened 2022-03-26
[Annotations] Add support for printing/saving choice list with multiple selections

### LINKED ISSUE #12189: Implement printing/saving support for listboxes
The test file at https://github.com/mozilla/pdf.js/blob/master/test/pdfs/annotation-choice-widget.pdf contains listboxes, which are not rendered correctly when printed right now. These kinds of fields are made by https://github.com/mozilla/pdf.js/blob/master/src/display/annotation_layer.js#L672-L678.

/cc @calixteman

calixteman:
- it aims to fix issue #12189.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -50,6 +50,11 @@ import { StringStream } from "./stream.js";
 import { writeDict } from "./writer.js";
 import { XFAFactory } from "./xfa/factory.js";
 
+// Represent the percentage of the height of a single-line field over
+// the font size.
+// Acrobat seems to use this value.
+const LINE_FACTOR = 1.35;
+
 class AnnotationFactory {
   /**
    * Create an `Annotation` object of the correct type for the given reference
@@ -1405,6 +1410,16 @@ class WidgetAnnotation extends Annotation {
       return null;
     }
 
+    // Value can be an array (with choice list and multiple selections)
+    if (
+      Array.isArray(value) &&
+      Array.isArray(this.data.fieldValue) &&
+      value.length === this.data.fieldValue.length &&
+      value.every((x, i) => x === this.data.fieldValue[i])
+    ) {
+      return null;
+    }
+
     let appearance = await this._getAppearance(
       evaluator,
       task,
@@ -1448,7 +1463,8 @@ class WidgetAnnotation extends Annotation {
       appearance = newTransform.encryptString(appearance);
     }
 
-    dict.set("V", isAscii(value) ? value : stringToUTF16BEString(value));
+    const encoder = val => (isAscii(val) ? val : stringToUTF16BEString(val));
+    dict.set("V", Array.isArray(value) ? value.map(encoder) : encoder(value));
     dict.set("AP", AP);
     dict.set("M", `D:${getModificationDate()}`);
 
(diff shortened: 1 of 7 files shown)


Snuffleupagus:
Looking at the `annotation-choice-widget-print-page1` test results, I'm wondering if the empty "List box, read-only"-part makes sense here? More specifically, since that element ought to be read-only, should we perhaps remove the following part of the manifest such that we actually test printing of the read-only element:
https://github.com/mozilla/pdf.js/blob/db4f3adc5e521d767e0aba6327f7633b926f1523/test/test_manifest.json#L5902-L5904

src/core/annotation.js
```diff
@@ -2483,6 +2494,133 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
...
+    if (!Array.isArray(exportedValue)) {
+      exportedValue = [exportedValue];
+    }
+
+    const defaultPadding = 2;
+    const hPadding = defaultPadding;
+    const totalHeight = this.data.rect[3] - this.data.rect[1];
+    const totalWidth = this.data.rect[2] - this.data.rect[0];
+    const lineCount = this.data.options.length;
+    const valueIndice = [];
```
Snuffleupagus, line 2522: Spelling-nit: `valueIndices`

src/core/annotation.js
```diff
@@ -2483,6 +2494,133 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
...
+    for (let i = 0, ii = this.data.options.length; i < ii; i++) {
```
Snuffleupagus, line 2523: Nit: How about just using `i < lineCount` here, since the length was already determined just above?

src/core/annotation.js
```diff
@@ -2483,6 +2494,133 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
...
+        font,
+        -1
+      );
+    } else {
+      defaultAppearance = this._defaultAppearance;
+    }
+
+    const lineHeight = fontSize * LINE_FACTOR;
+    const vPadding = (lineHeight - fontSize) / 2;
+    const linesByPages = Math.floor(totalHeight / lineHeight);
```
Snuffleupagus, line 2570: This variable name seems a little unclear, at least to me.
It seems that you're actually calculating the number of lines in the Widget, so in that case how about e.g. `numberOfLines` instead?

src/core/annotation.js
```diff
@@ -2483,6 +2494,133 @@ class ChoiceWidgetAnnotation extends WidgetAnnotation {
...
+      // If nothing is selected (valueIndice.length === 0), we render
+      // from the first element.
+      firstIndex = valueIndice.length ? valueIndice[0] : 0;
+    }
+    const end = Math.min(firstIndex + linesByPages + 1, lineCount);
+
+    const buf = [
+      "/Tx BMC q",
+      `1 1 ${totalWidth} ${totalHeight} re W n`,
+      "0.600006 0.756866 0.854904 rg",
```
Snuffleupagus, line 2587: Perhaps add a comment mentioning what RGB value this correspond to, for future reference?

test/test_manifest.json
```diff
@@ -5897,12 +5897,9 @@
            "value": "Dolor"
          },
          "62R": {
-           "value": "Sit"
-         },
-         "63R": {
-           "value": ""
+           "value": ["Sit", "Adipiscing"]
          }
-       }
```
Snuffleupagus, line 5905: Nit: Looks unrelated.

Snuffleupagus (approved):
r=me, with a couple of small comments.

PR #14722, opened 2022-03-28
PDFScriptingManager: Bind mousedown listener with capture=true

### LINKED ISSUE #14721: Hand tool prevents custom JavaScript validation
[pdf_form_js_validation_problem.pdf](https://github.com/mozilla/pdf.js/files/8357577/pdf_form_js_validation_problem.pdf)

Configuration:
- Web browser and its version: doesn't matter
- Operating system and its version: doesn't matter
- PDF.js version: current source code on GitHub; 2.13
- Is a browser extension: no

1. Navigate to https://mozilla.github.io/pdf.js/web/viewer.html
2. Activate the hand tool 
3. open the attached PDF file
4. enter letters into the input field
5. click somewhere outside the input field to trigger validation --> there's no validation. Plus, the input is lost, but it re-appears after entering the input field again.
6. However, leaving the field using the TAB key does trigger validation

image

image

There's a conflict between these lines:
https://github.com/mozilla/pdf.js/blob/20d60d92ba3c701d79a62db2b8ff7899e752925e/web/grab_to_pan.js#L58
https://github.com/mozilla/pdf.js/blob/20d60d92ba3c701d79a62db2b8ff7899e752925e/web/grab_to_pan.js#L71

and this line:
https://github.com/mozilla/pdf.js/blob/869b396011b03d00325f2a5d5c01e58841efc312/web/pdf_scripting_manager.js#L157

After removing the last parameter from the grab_to_pan file (i.e. making it `this.element.removeEventListener("mousedown", this._onMouseDown)`), everything works as expected. Alternatively, I assume you could also add a `, true` to the pdf_scripting_manager line (but I didn't test it yet).

However, I don't know if there are side effects. Digging into the history of the grab-to-pan class reveals the additional parameter was there from day 1 of the external project written by @Rob--W (https://github.com/Rob--W/grab-to-pan.js), so I suspect it's not really necessary.

Rob--W:
PDFScriptingManager uses the `mousedown` and `mouseup` listeners to keep track of whether the mouse pointer is pressed in the `isDown` flag. These listeners were registered to run during the bubbling phase of the event dispatch, which can be interrupted if any of the previous event listeners stopped the event propagation. An example of that is by `GrabToPan` in web/grab_to_pan.js.

Since the mousedown (and mouseup) listeners of PDFScriptingManager are free of side effects, and the intention is to always run them, it makes most sense to register them with the capture flag.

More details at https://github.com/mozilla/pdf.js/issues/14721#issuecomment-1080065362
Fixes #14721

### DIFF
--- a/web/pdf_scripting_manager.js
+++ b/web/pdf_scripting_manager.js
@@ -154,7 +154,7 @@ class PDFScriptingManager {
       this._eventBus._on(name, listener);
     }
     for (const [name, listener] of this._domEvents) {
-      window.addEventListener(name, listener);
+      window.addEventListener(name, listener, true);
     }
 
     try {
@@ -507,7 +507,7 @@ class PDFScriptingManager {
     this._internalEvents.clear();
 
     for (const [name, listener] of this._domEvents) {
-      window.removeEventListener(name, listener);
+      window.removeEventListener(name, listener, true);
     }
     this._domEvents.clear();
 


Snuffleupagus:
What about all of the event listeners attached *directly* in the `AnnotationLayer`-code, as mentioned in https://github.com/mozilla/pdf.js/issues/14721#issuecomment-1079943719, are they not similarly affected?

Rob--W:
> What about all of the event listeners attached *directly* in the `AnnotationLayer`-code, as mentioned in https://github.com/mozilla/pdf.js/issues/14721#issuecomment-1079943719, are they not similarly affected?

These elements are input elements or links, which are already being recognized as targets where panning should not activate (source is linked in my previous comment).

> Also, does this still work even if the handTool is enabled *before* the `PDFScriptingManager`-functionality has been fully initialized, by e.g. setting the `cursorToolOnLoad` preference to `1`?

Yes. The time of registration does not matter because PDFScriptingManager registers the event with`window`, whereas GrabToPan registers the event on the container element. The capturing phase of the event starts at window, then descends the DOM tree via document, html, all the way down to the container element, and (if `event.stopPropagation()` is not callee) then bubbles up again in the reverse order up to window again. Your concern about registration time would be valid indeed if the events were registered on the same target (which they don't), e.g. as in the example at https://stackoverflow.com/q/7398290/#7398447 (bonus: the answer below that post has a visualisation of how event propagation works).

Snuffleupagus (approved):
Let's do this, thank you!

Would it be feasible to add an integration-test for this issue, e.g. in https://github.com/mozilla/pdf.js/blob/master/test/integration/scripting_spec.js, where we first enable the handTool and then ensure that the intended validation runs when the field loses focus?

Rob--W:
Oops. I missed the question.

> Would it be feasible to add an integration-test for this issue, e.g. in https://github.com/mozilla/pdf.js/blob/master/test/integration/scripting_spec.js, where we first enable the handTool and then ensure that the intended validation runs when the field loses focus?

It may be feasible to have a regression test for this issue, by simulating the specific events in a particular order. I don't see any unit tests for the `isDown` logic, i.e. the one that triggers the logic at https://github.com/mozilla/pdf.js/blob/4d4c67dd6e5c826287cd7cdd216435cd651a9301/src/display/annotation_layer.js#L1031, so having some would be very useful. I haven't written tests with that framework, so if you're willing to write tests I would gladly provide the review (to verify correctness of event order and test logic).

PR #14733, opened 2022-03-30
Use `String.prototype.repeat()` in a couple of spots

Snuffleupagus:
Rather than using a temporary Array to manually create repeated strings, we can use `String.prototype.repeat()` instead.
The reason that we didn't use this from the start is most likely because some browsers, notably IE, didn't support this; note https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat#browser_compatibility


calixteman (approved):
Thank you.

PR #14735, opened 2022-03-31
[Annotations] Some annotations can have their values stored in the xfa:datasets

### LINKED ISSUE #14685: PDFJS is not able to show prefilled Acroform PDF with data in xfa:dataset
When Using a PDF which has Type 1 embedded fonts, a prefilled PDF that has been filled with tools like IText or PDFBox cannot be handled as is shown as empty by PDF JS.

If however PDFJS has prefilled the PDF the texts are shown.

Update: After discussion further down, now we know that the PDF is actually a hybrid of Acroform and XFA and PDFJS is not able to prefill Acroform with data in the XFA:dataset, although when saving the PDF, PDFJS actually save them in XFA:dataset also.

[1647183160545.pdf](https://github.com/mozilla/pdf.js/files/8283456/1647183160545.pdf)
Configuration:
- Web browser and its version: Chrome (99)
- PDF.js version:
- Is a browser extension:

1. open the attached file in Acrobat Reader
2. open the attached file in PDFJS

image

image

[1647183160545.pdf](https://github.com/mozilla/pdf.js/files/8283456/1647183160545.pdf)

calixteman:
- it aims to fix #14685;
- add a basic object to get values from the parsed datasets;
- these annotations don't have an appearance so we must create one when printing or saving.

### DIFF
--- a/src/core/annotation.js
+++ b/src/core/annotation.js
@@ -72,14 +72,16 @@ class AnnotationFactory {
   static create(xref, ref, pdfManager, idFactory, collectFields) {
     return Promise.all([
       pdfManager.ensureCatalog("acroForm"),
+      pdfManager.ensureDoc("xfaDatasets"),
       collectFields ? this._getPageIndex(xref, ref, pdfManager) : -1,
-    ]).then(([acroForm, pageIndex]) =>
+    ]).then(([acroForm, xfaDatasets, pageIndex]) =>
       pdfManager.ensure(this, "_create", [
         xref,
         ref,
         pdfManager,
         idFactory,
         acroForm,
+        xfaDatasets,
         collectFields,
         pageIndex,
       ])
@@ -95,6 +97,7 @@ class AnnotationFactory {
     pdfManager,
     idFactory,
     acroForm,
+    xfaDatasets,
     collectFields,
     pageIndex = -1
   ) {
@@ -119,6 +122,7 @@ class AnnotationFactory {
       id,
       pdfManager,
       acroForm: acroForm instanceof Dict ? acroForm : Dict.empty,
+      xfaDatasets,
       collectFields,
       pageIndex,
     };
@@ -1237,7 +1241,7 @@ class WidgetAnnotation extends Annotation {
       );
     }
 
-    const fieldValue = getInheritableProperty({
+    let fieldValue = getInheritableProperty({
       dict,
       key: "V",
       getArray: true,
@@ -1250,6 +1254,15 @@ class WidgetAnnotation extends Annotation {
       getArray: true,
     });
     data.defaultFieldValue = this._decodeFormValue(defaultFieldValue);
+
+    if (fieldValue === undefined && params.xfaDatasets) {
+      // Try to figure out if we have something in the xfa dataset.
+      const path = this._title.str;
+      if (path) {
+        this._hasValueFromXFA = true;
+        data.fieldValue = fieldValue = params.xfaDatasets.getValue(path);
+      }
+    }
 
     // When no "V" entry exists, let the fieldValue fallback to the "DV" entry
     // (fixes issue13823.pdf).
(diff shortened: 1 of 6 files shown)


src/core/annotation.js
```diff
@@ -1251,6 +1255,15 @@ class WidgetAnnotation extends Annotation {
     });
     data.defaultFieldValue = this._decodeFormValue(defaultFieldValue);
 
+    if (fieldValue === undefined && params.xfaDatasets) {
+      // Try to figure out if we have something in the xfa dataset.
+      const path = stringToPDFString(dict.get("T") || "");
```
Snuffleupagus, line 1260: *I don't have time to do a full review right now, but wanted to mention one thing straight away.*

This ought to be available already, via `this._title.str` (note the `setTitle`-call in the `Annotation`- constructor), and you shouldn't need to fetch/parse this again here.

Snuffleupagus, line 1260: This ought to be available already, via `this._title.str` (note the `setTitle`-call in the `Annotation`- constructor), and you shouldn't need to fetch/parse this again here.

src/core/annotation.js
```diff
@@ -1251,6 +1255,15 @@ class WidgetAnnotation extends Annotation {
...
+      if (path) {
+        data.hasValueFromXFA = true;
```
Snuffleupagus, line 1262: Please note that by adding a value to the `data` object you're exposing it in the API as well.

Given that we're only using this in the worker-thread, can we please store (and access) this on the class-instance as `this._hasValueFromXFA = true;` instead?

src/core/dataset_reader.js
```diff
@@ -0,0 +1,70 @@
...
+    this.node = null;
+  }
+
+  onEndElement(name) {
+    const node = super.onEndElement(name);
+    if (node && name === "xfa:datasets") {
+      this.node = node;
+
+      // We don't need anything else, so just kill the parser.
+      throw new Error();
```
Snuffleupagus, line 31: To ensure that we can always get helpful/useful error messages we should always provide *some* kind of message string for all errors, hence please do something like:
```javascript
throw new Error("Aborting DatasetXMLParser.");
```

src/core/document.js
```diff
@@ -820,6 +821,47 @@ class PDFDocument {
     });
   }
 
+  get xfaDatasets() {
```
Snuffleupagus, line 824: I cannot help but noticing the similarities between this new method and the existing `xfaData` one just below. Hence I wonder if it'd perhaps make sense, in a follow-up PR, to re-factor these two method to remove some of the duplication?

src/core/xml_parser.js
```diff
@@ -328,6 +328,10 @@ class SimpleDOMNode {
       .join("");
   }
 
+  get children() {
+    return this.childNodes;
```
Snuffleupagus, line 332: Given the method just below, it seems that `this.childNodes` could *possibly* be undefined in the general case.
Hence it seems like a good idea to prevent run-time errors elsewhere, by doing:
```javascript
return this.childNodes || [];
```

src/core/dataset_reader.js
```diff
@@ -0,0 +1,70 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { parseXFAPath } from "./core_utils.js";
+import { SimpleXMLParser } from "./xml_parser.js";
+
+class DataSetXMLParser extends SimpleXMLParser {
```
Snuffleupagus, line 19: Spelling-nit: Let's call this one `DatasetXMLParser` instead, note the casing, since that'd be consistent with the casing used everywhere else in this patch :-)

Snuffleupagus (changes requested):

src/core/document.js
calixteman: Yes you're right, I'll do a follow-up PR.

Snuffleupagus (approved):
r=me, this looks reasonable; thank you!

PR #14738, opened 2022-04-01
Decode non-ASCII values found in the xfa:datasets (PR 14735 follow-up)

### LINKED ISSUE #14685: PDFJS is not able to show prefilled Acroform PDF with data in xfa:dataset

Snuffleupagus:
*Please note:* This is possibly bad/wrong in general, but I figured that submitting it for review wouldn't hurt.

It seems that even Adobe Reader doesn't handle the non-ASCII characters that appear in some of the fields correctly, however it should be pretty easy to improve things on the PDF.js side.

### DIFF
--- a/src/core/dataset_reader.js
+++ b/src/core/dataset_reader.js
@@ -13,8 +13,18 @@
  * limitations under the License.
  */
 
+import { stringToUTF8String, warn } from "../shared/util.js";
 import { parseXFAPath } from "./core_utils.js";
 import { SimpleXMLParser } from "./xml_parser.js";
+
+function decodeString(str) {
+  try {
+    return stringToUTF8String(str);
+  } catch (ex) {
+    warn(`UTF-8 decoding failed: "${ex}".`);
+    return str;
+  }
+}
 
 class DatasetXMLParser extends SimpleXMLParser {
   constructor(options) {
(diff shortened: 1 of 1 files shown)


calixteman:
Eh eh, you've been faster than me: it was on my todo list.

calixteman (approved):
Nice, thank you.

PR #14739, opened 2022-04-01
Refactor some xfa*** getters in document.js

calixteman:
- it's a follow-up of PR #14735.

### DIFF
--- a/src/core/document.js
+++ b/src/core/document.js
@@ -821,48 +821,7 @@ class PDFDocument {
     });
   }
 
-  get xfaDatasets() {
-    const acroForm = this.catalog.acroForm;
-    if (!acroForm) {
-      return shadow(this, "xfaDatasets", null);
-    }
-
-    const xfa = acroForm.get("XFA");
-    if (xfa instanceof BaseStream && !xfa.isEmpty) {
-      try {
-        const xdp = stringToUTF8String(xfa.getString());
-        return shadow(this, "xfaDatasets", new DatasetReader({ xdp }));
-      } catch (_) {
-        warn("XFA - Invalid utf-8 string.");
-        return shadow(this, "xfaDatasets", null);
-      }
-    }
-
-    if (!Array.isArray(xfa) || xfa.length === 0) {
-      return null;
-    }
-
-    for (let i = 0, ii = xfa.length; i < ii; i += 2) {
-      if (xfa[i] !== "datasets") {
-        continue;
-      }
-      const data = this.xref.fetchIfRef(xfa[i + 1]);
-      if (!(data instanceof BaseStream) || data.isEmpty) {
-        continue;
-      }
-      try {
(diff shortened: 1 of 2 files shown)


src/core/document.js
```diff
@@ -821,76 +821,30 @@ class PDFDocument {
     });
   }
 
-  get xfaDatasets() {
+  get xfaStreams() {
```
Snuffleupagus, line 824: I don't think that we want to use `shadow` in this helper method, since that'll mean caching of Streams and those can be arbitrarily large once decoded (in the general case). Note how we, for that very reason, are purposely *not caching* Streams on the `XRef`-instance.

Also, maybe name it `_xfaStreams` instead to signal that it shouldn't be used directly?

src/core/document.js
```diff
@@ -910,14 +864,52 @@ class PDFDocument {
...
+        return shadow(this, "xfaDatasets", new DatasetReader(data));
+      } catch (_) {
+        warn("XFA - Invalid utf-8 string.");
+        break;
+      }
+    }
+    return shadow(this, "xfaDatasets", null);
+  }
+
+  get xfaData() {
```
Snuffleupagus, line 894: In the existing code we're not using `shadow` for this method, since it's only being accessed *once* from the `xfaFactory` getter and shadowing is being used there instead.

Snuffleupagus (changes requested):

src/core/document.js
calixteman: Yes and it's especially useless because we don't have to call this function so many times.

src/core/document.js
```diff
@@ -821,103 +821,95 @@ class PDFDocument {
...
+    for (const key of ["datasets", "xdp:xdp"]) {
+      const stream = streams[key];
+      if (!stream) {
+        continue;
+      }
       try {
-        entries["xdp:xdp"] = stringToUTF8String(xfa.getString());
-        return entries;
+        const str = stringToUTF8String(stream.getString());
+        const data = Object.fromEntries([[key, str]]);
```
Snuffleupagus, line 884: Nit: Can't you create this object directly, using the `{ [key]: str }` notation, or does that perhaps not work correctly with the development viewer (because of the SystemJS dependency)?

src/core/document.js
```diff
@@ -821,103 +821,95 @@ class PDFDocument {
...
+    const entries = {
+      "xdp:xdp": null,
+      template: null,
+      datasets: null,
+      config: null,
+      connectionSet: null,
+      localeSet: null,
+      stylesheet: null,
+      "/xdp:xdp": null,
+    };
```
Snuffleupagus, line 840: Is there a particular reason for using `null` here rather than empty strings as before, since then you could skip the explicit assign of empty strings in `get xfaData()` below?

Snuffleupagus (approved):
This looks nicer, thank you!

PR #14742, opened 2022-04-02
Replace most loops in `web/debugger.js` with `for...of` loops

Snuffleupagus:
This leads to *slightly* more compact code overall. Also, uses object shorthand notation to remove some now unnecessary `function`-strings.


timvandermeij (approved):

timvandermeij:
Thank you for improving this!

PR #14743, opened 2022-04-02
Use native `import maps` in development mode

Snuffleupagus:
This patch seem to work fine locally now, and `mozregression` points to it being fixed in bug https://bugzilla.mozilla.org/show_bug.cgi?id=1803984 which landed in Firefox 116.

By using the native `import maps` functionality, we can remove a development dependency. Please find the specification at https://wicg.github.io/import-maps/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,6 @@
         "canvas": "^2.11.2",
         "core-js": "^3.32.2",
         "cross-env": "^7.0.3",
-        "es-module-shims": "1.4.7",
         "eslint": "^8.50.0",
         "eslint-config-prettier": "^8.10.0",
         "eslint-plugin-fetch-options": "^0.0.5",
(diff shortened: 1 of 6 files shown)


Snuffleupagus:
Note that https://bugzilla.mozilla.org/show_bug.cgi?id=1688879 has now landed in Firefox, however:
 - The functionality is *disabled* by default, even in Nightly; see `dom.importMaps.enabled` in about:config.
 - When manually enabled, the development viewer fails to load intermittently and the unit/font-test fail consistently; this problem is being tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1762595 (at least I think so).

calixteman:
I'd like to have the possibility to test the `import maps` functionality.
Would it be possible to have a special target in `gulpfile` (e.g. `gulp server-experimental` or something like that) in order to test that stuff and potentially help to find some bugs ?

Snuffleupagus:
> I'd like to have the possibility to test the `import maps` functionality.

*Just in case it's not immediately clear:* Please note that we're *already* using `import maps` in development mode, since quite some time, thanks to the `es-module-shims` polyfill.

The point of this PR is only to remove that polyfill, and start using the native browser functionality. As mentioned in https://github.com/mozilla/pdf.js/pull/14743#issuecomment-1120293329, moving forward with this PR is currently blocked on [bug 1762595](https://bugzilla.mozilla.org/show_bug.cgi?id=1762595) being fixed.

> Would it be possible to have a special target in `gulpfile` (e.g. `gulp server-experimental` or something like that) in order to test that stuff and potentially help to find some bugs ?

Not easily, I don't think, and as mentioned native `import maps` currently don't work (properly) in Firefox anyway.

timvandermeij:
Are there any blockers left here? It looks like the required upstream bugs are closed and `dom.importMaps.enabled` also seems to have been flipped to `true` by default already (at least in my Firefox 113.0.2 instance, and I haven't touched that preference before).

timvandermeij (approved):

timvandermeij:
Nice work; good to see that the upstream issues have been fixed and we can finally get rid of that extra dependency!

PR #14744, opened 2022-04-02
Update GitHub Actions workflow steps/dependencies/translations to the most recent versions

timvandermeij:


package.json
```diff
@@ -43,24 +43,24 @@
...
+    "stylelint": "^14.6.1",
     "stylelint-config-prettier": "^9.0.3",
     "stylelint-prettier": "^2.0.0",
     "systemjs": "^0.21.6",
     "systemjs-plugin-babel": "^0.0.25",
     "terser": "^5.12.1",
     "through2": "^4.0.2",
-    "ttest": "^3.0.0",
-    "typescript": "^4.6.2",
+    "ttest": "^4.0.0",
```
Snuffleupagus, line 57: Given that this is a *major* version increase, have you checked that the `stats/statcmp.js` script still works correctly without any to its code?

timvandermeij: Yes, it does. I'm not really sure why upstream decided to bump the major version here given that the commits between 3.0.0 and 4.0.0 at https://github.com/AndreasMadsen/ttest/commits/master don't seem to contain any API changes...

Snuffleupagus (approved):
I'm guessing that `transactions` in the PR title is a typo, and that it should be `translations` instead?

r=me, thank you!

timvandermeij:
> I'm guessing that transactions in the PR title is a typo, and that it should be translations instead?

Whoops, indeed. Fixed now.

PR #14745, opened 2022-04-03
Convert `web/debugger.js` to a *basic* module

Snuffleupagus:
The various functionality in `web/debugger.js` is currently *indirectly* added to the global scope, since that's how `var` works when it's used outside of any functions/closures.
Given how this functionality is being accessed/used, not just in the viewer but also in the API and textLayer, simply converting the entire file to a module isn't trivial[1]. However, we can at least export the `PDFBug`-part properly and then `import` that dynamically in the viewer.
Also, to improve the code a little bit, we're now *explicitly* exporting the necessary functionality globally.

According to MDN, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#browser_compatibility, all the browsers that we now support have dynamic `imports` implementations.

---
[1] We could probably pass around references to the necessary functionality, but given how this is being used I'm just not sure it's worth the effort. Also, adding *official* support for these viewer-specific debugging tools in the API feels both unnecessary and unfortunate.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -327,6 +327,10 @@ function replaceWebpackRequire() {
   return replace("__webpack_require__", "__w_pdfjs_require__");
 }
 
+function replaceNonWebpackImport() {
+  return replace("__non_webpack_import__", "import");
+}
+
 function replaceJSRootName(amdName, jsName) {
   // Saving old-style JS module name.
   return replace(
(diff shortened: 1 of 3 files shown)


timvandermeij (approved):

timvandermeij:
Nice refactoring; thanks!

calixteman:
This change is making a test failing in m-c:
https://treeherder.mozilla.org/logviewer?job_id=374645040&repo=autoland&lineNumber=2185
It's reproducible locally in m-c in running command:
`./mach mochitest --headless browser_parsable_script.js`

@Snuffleupagus, could you have a look please ?

Snuffleupagus:
Given the name of the failing test, it seems that you'll need to add an entry for the `web/debugger.js` file in this list:
https://searchfox.org/mozilla-central/rev/d34f9713ae128a3138c2b70d8041a535f1049d19/browser/base/content/test/static/browser_parsable_script.js#15

calixteman:
Cool it's an easy fix, thank you.

